PR #48, opened 2017-06-30
network manager: remove duplicated method

paulirish:
see line 86.


paulirish:
(I would have expected some eslint rule to get upset about this, but didn't look into it.)

PR #23, opened 2017-06-19
WIP: use chrome-launcher for handling the browser.

paulirish:
this is the launcher module from lighthouse that we maintain.
https://www.npmjs.com/package/chrome-launcher

it defaults to an open port

addresses #21

PR #55, opened 2017-07-07
Teach browser to dispatch 'stderr' event to report its stderr stream

aslushnikov:
I usually debug chromium with `fprintf` statements. This patch makes it convenient:

```js
let Browser = require('./lib/Browser');
let browser = new Browser({
  // Point to my local chromium build
  executablePath: '/usr/local/home/lushnikov/chromium/out/Release/chrome'
});
// Output chromium stderr and stdout to console
browser.on('stdout', console.log);
browser.on('stderr', console.log);

browser.newPage().then(async page => {
   ...
});
```

### DIFF
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -161,9 +161,15 @@ function waitForChromeResponsive(remoteDebuggingPort, shouldWaitCallback) {
 }
 
 class ProxyStream extends Duplex {
-  _read(size) { }
+  _read() { }
 
+  /**
+   * @param {?} chunk
+   * @param {string} encoding
+   * @param {function()} callback
+   */
   _write(chunk, encoding, callback) {
-    this.push(chunk);
+    this.push(chunk, encoding);
+    callback();
   }
 }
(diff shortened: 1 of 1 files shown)


aslushnikov:
@paulirish what do you think about this? Does it play nice with #23? I initially though about having `browser.stdout` and `browser.stderr` streams which proxy the underlying process streams, but that felt like an overkill.

paulirish:
cc @samccone 

Yeah I think its idiomatic node style to get the stderr/stdout streams. Given this, i think launcher should expose the child_process object.

aslushnikov:
Switched to streams.

The new api would be `browser.stdout.pipe(process.stdout)`:
```js
let Browser = require('./lib/Browser');
let browser = new Browser({
  // Point to my local chromium build
  executablePath: '/usr/local/home/lushnikov/chromium/out/Release/chrome'
});
// Output chromium stderr and stdout to console
browser.stdout.pipe(process.stdout);
browser.stderr.pipe(process.stdout);

browser.newPage().then(async page => {
   ...
});
```

PR #58, opened 2017-07-07
$ and $$ for querySelector and querySelectorAll

JoelEinbinder:


aslushnikov:
Joel, could you give some examples on how is this useful? For example, the
```js
page.$$('h1', h1 => h1.textContent)
```
is the same as
```js
page.evaluate(() => document.querySelectorAll('h1').map(h1 => h1.textContent));
```
It looks shorter, but I'm not sure that's a very common scenario.
@paulirish do you have any examples in mind too?

JoelEinbinder:
The long hand version works, but it is surprisingly difficult to compress into a helper function. This let's you do things like 

```js
let button = page.$.bind(page, 'button');
let click = e => e.click();
button(click);
```

docs/api.md
```diff
@@ -120,6 +122,20 @@ browser.newPage().then(page => {
 ```
 Pages could be closed by `page.close()` method.
 
+#### page.$(querySelector, fun, args)
+
+- `querySelector` <[string]> Query selector to be run on the page
+- `fun` <[function<[Element]>]> Function to be evaluated with first element matching `querySelector`
+- `args` <[Array]<[string]>> Arguments to pass to `fun`
```
paulirish: seems more like a Array[*] ?

docs/api.md
```diff
@@ -120,6 +122,20 @@ browser.newPage().then(page => {
 ```
 Pages could be closed by `page.close()` method.
 
+#### page.$(querySelector, fun, args)
+
+- `querySelector` <[string]> Query selector to be run on the page
```
paulirish: just the string is commonly referred to as `selector`.  "querySelector" is a method.

lib/helper.js
```diff
@@ -52,6 +52,36 @@ class Helper {
     }
     return message;
   }
+
+  /**
+   * @param {!Connection} client
+   * @param {!Object} remoteObject
+   */
+  static async serializeRemoteObjet(client, remoteObject) {
```
paulirish, line 60: typo. missing a `c`

docs/api.md
```diff
@@ -120,6 +122,20 @@ browser.newPage().then(page => {
...
+- `fun` <[function<[Element]>]> Function to be evaluated with first element matching `querySelector`
+- `args` <[Array]<[string]>> Arguments to pass to `fun`
+- returns: <[Promise<[Object]]> Promise which resolves to function return value.
+
+#### page.$$(querySelector, fun, args)
+
+- `querySelector` <[string]> Query selector to be run on the page
+- `fun` <[function<[Element]>]> Function to be evaluted for every element matching `querySelector`.
```
paulirish: "evaluated"

I'd also add "clientside" like.. "Function to be evaluated clientside for every...."

paulirish:
examples

```js
// return array of all heading text
page.$$('h1,h2,h3,h4', el => el.textContent)

// array of all link elements attributes
page.$$('link', el => {
  const obj = {};
  for (const attr of el.attributes) obj[attr.name] = attr.value;
  return obj;
});
```

lib/Page.js
```diff
@@ -561,6 +572,53 @@ class Page extends EventEmitter {
...
   }
+
+  /**
+   * @param {number} nodeId
+   * @param {function(!Element):T} fun
+   * @param {!Array<*>} args
+   * @return {!Promise<T>}
+   */
+  async _callFunctionOnNode(nodeId, fun, ...args) {
+    let { objectId } = (await this._client.send('DOM.resolveNode', { nodeId })).object;
```
aslushnikov: let's impement this in javascript. It will save us **a lot** of roundtrips between puppeteer and page. It will also work much faster.

aslushnikov (changes requested):

PR #59, opened 2017-07-07
Added frame.waitFor(selector)

alexkozy:
Bug: https://github.com/GoogleChrome/puppeteer/issues/42

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -197,6 +197,39 @@ class FrameManager extends EventEmitter {
     });
     return response.result.value;
   }
+
+  /**
+   * @param {string} selector
+   * @param {!Frame} frame
+   * @return {!Promise<undefined>}
+   */
+  async _waitForInFrame(selector, frame) {
+    let code = selector => new Promise((fulfill, reject) => {
+      if (document.querySelector(selector)) {
+        fulfill();
+        return;
+      }
+      new MutationObserver((mutations, observer) => {
+        for (let mutation of mutations) {
+          for (let node of mutation.addedNodes) {
+            if (node.matches(selector)) {
+              observer.disconnect();
+              fulfill();
+              return;
+            }
+          }
+        }
+      }).observe(document.documentElement, {
+        childList: true,
+        subtree: true
+      });
+    });
+    await this._client.send('Runtime.evaluate', {
+      expression: helper.evaluationString(code, selector),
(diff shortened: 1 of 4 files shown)


lib/FrameManager.js
```diff
@@ -197,6 +197,33 @@ class FrameManager extends EventEmitter {
     });
     return response.result.value;
   }
+
+  /**
+   * @param {!Frame} frame
```
paulirish: missing selector arg in jsdoc

lib/FrameManager.js
```diff
@@ -197,6 +197,33 @@ class FrameManager extends EventEmitter {
...
+   * @return {!Promise<undefined>}
+   */
+  async _waitForInFrame(frame, selector) {
```
paulirish: i'd swap the order of these two

alexkozy: done.

lib/FrameManager.js
alexkozy: done.

lib/FrameManager.js
```diff
@@ -197,6 +197,34 @@ class FrameManager extends EventEmitter {
...
   }
+
+  /**
+   * @param {string} selector
+   * @param {!Frame} frame
+   * @return {!Promise<undefined>}
+   */
+  async _waitForInFrame(selector, frame) {
+    let code = selector => new Promise((fulfill, reject) => {
+      new MutationObserver((mutations, observer) => {
```
aslushnikov: let's check if it already exists

lib/FrameManager.js
```diff
@@ -197,6 +197,34 @@ class FrameManager extends EventEmitter {
...
+          }
+        }
+      }).observe(document.documentElement, {
+        childList: true,
+        subtree: true
+      });
+    });
+    await this._client.send('Runtime.evaluate', {
+      expression: helper.evaluationString(code, selector),
+      awaitPromise: true
```
aslushnikov: returnByValue: true

aslushnikov (changes requested):

lib/FrameManager.js
alexkozy: done.

lib/FrameManager.js
alexkozy: done, test added!

Issue #54, opened 2017-07-06
Add API to select files in the file picker


aslushnikov:
There should be a way to set a file for file input:

```html
<input type="file" id="file">
```

JoelEinbinder:
I'll do this.

aslushnikov:
The phantom's way of doing this is [page.uploadFile](http://phantomjs.org/api/webpage/method/upload-file.html) method.

PR #61, opened 2017-07-08
Page.uploadFile

JoelEinbinder:
#54

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -33,7 +33,7 @@
   * [page.setViewportSize(size)](#pagesetviewportsizesize)
   * [page.title()](#pagetitle)
   * [page.type()](#pagetype)
-  * [page.uploadFile(selector, ...files)](#pageuploadfileselector-files)
+  * [page.uploadFile(selector, ...filePaths)](#pageuploadfileselector-filepaths)
   * [page.url()](#pageurl)
   * [page.userAgent()](#pageuseragent)
   * [page.viewportSize()](#pageviewportsize)
@@ -237,9 +237,9 @@ Pages could be closed by `page.close()` method.
 
 #### page.type()
 
-#### page.uploadFile(selector, ...files)
+#### page.uploadFile(selector, ...filePaths)
 - `selector` <[string]> A query selector to a file input
-- `...files` <[string]> Sets the value of the file input these files
+- `...filePaths` <[string]> Sets the value of the file input these paths
 - returns: <[Promise]> Promise which resolves when the value is set.
 
 #### page.url()
--- a/docs/api.md
+++ b/docs/api.md
@@ -33,6 +33,7 @@
   * [page.setViewportSize(size)](#pagesetviewportsizesize)
   * [page.title()](#pagetitle)
   * [page.type()](#pagetype)
+  * [page.uploadFile(selector, ...files)](#pageuploadfileselector-files)
   * [page.url()](#pageurl)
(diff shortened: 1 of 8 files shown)


docs/api.md
```diff
@@ -236,6 +237,11 @@ Pages could be closed by `page.close()` method.
 
 #### page.type()
 
+#### page.uploadFile(selector, ...files)
+- `selector` <[string]> A query selector to a file input
+- `...files` <[string]> Sets the value of the file input these files
```
paulirish: I was definitely expecting this to be taking a file descriptor or buffer.  how about

> Sets the value of the file input these file paths

and i would change the name of the argument to `filePaths`

lib/Page.js
```diff
@@ -569,6 +569,13 @@ class Page extends EventEmitter {
   waitFor(selector) {
     return this.mainFrame().waitFor(selector);
   }
+
+  async uploadFile(selector, ...filePaths) {
+    await this._client.send('DOM.setFileInputFiles', {
+      nodeId: await this._querySelector(selector),
```
aslushnikov: what will happen if this._querySelector fails?

lib/Page.js
```diff
@@ -569,6 +569,13 @@ class Page extends EventEmitter {
   waitFor(selector) {
     return this.mainFrame().waitFor(selector);
   }
+
+  async uploadFile(selector, ...filePaths) {
```
aslushnikov: let's add jsdoc

phantom_shim/WebPage.js
```diff
@@ -333,6 +333,17 @@ class WebPage {
   }
 
   /**
+   * @param {string} selector
+   * @param {string|!Array<string>} files
```
aslushnikov: i'm pretty sure type annotation should be {(string|!Array<string>)}

lib/Page.js
aslushnikov: I don't mind having `...filePaths`, but curious why you chose this over the `array<string>`, which seems to be more natural?

phantom_shim/WebPage.js
JoelEinbinder: But it's an array of strings?

lib/Page.js
JoelEinbinder: Multiple file upload is an edge case, so i wanted to have ```uploadFile('input', 'myfile.txt')``` work.

phantom_shim/WebPage.js
aslushnikov: Right, I meant `{(string|!Array<string>)}`

lib/Page.js
JoelEinbinder: Protocol will throw an error. Same will happen with`.click`.

phantom_shim/WebPage.js
JoelEinbinder: I don't understand, but sure.

lib/Page.js
JoelEinbinder: Done.

lib/Page.js
aslushnikov: Let's handle this separately and throw a more clear error. Currently we get
```
(node:30617) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Protocol error (DOM.getBoxModel): Could not find node with given id undefined
```
which is quite confusing.

Issue #50, opened 2017-07-06
Add tests for all public API methods


aslushnikov:
Every public API method should have a test. At least a sanity test.

aslushnikov:
Proposal:
- where possible, name tests as CLASS.METHOD, e.g. `Page.navigate`

aslushnikov:
@JoelEinbinder: Reopening since we still miss public events coverage.

PR #62, opened 2017-07-08
Add tests for all public methods on Page

JoelEinbinder:
#50

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -285,8 +285,8 @@ class Page extends EventEmitter {
         fitWindow: false
       }),
       this._client.send('Emulation.setVisibleSize', {
-        width: width / zoom,
-        height: height / zoom,
+        width: Math.floor(width / zoom),
+        height: Math.floor(height / zoom),
       })
     ]);
   }
--- a/test/test.js
+++ b/test/test.js
@@ -729,11 +729,7 @@ describe('Puppeteer', function() {
 
   describe('Page.plainText', function() {
     it('should return the page text', SX(async function(){
-      await page.evaluate(function(){
-        let element = document.createElement('div');
-        document.body.appendChild(element);
-        element.textContent = 'the result text';
-      });
+      await page.setContent('<div>the result text</div>');
       expect(await page.plainText()).toBe('the result text');
     }));
   });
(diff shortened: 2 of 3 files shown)


test/test.js
```diff
@@ -678,6 +679,71 @@ describe('Puppeteer', function() {
...
+    it('should print to pdf', SX(async function() {
+      await page.navigate(STATIC_PREFIX + '/grid.html');
+      await page.printToPDF(outputFile);
+      expect(fs.readFileSync(outputFile).byteLength).toBeGreaterThan(0);
+    }));
+  });
+
+  describe('Page.plainText', function() {
+    it('should return the page text', SX(async function(){
+      await page.evaluate(function(){
```
aslushnikov: this could be done
```js
await page.setContent('<div>the result text</div>');
```

test/test.js
```diff
@@ -678,6 +679,71 @@ describe('Puppeteer', function() {
...
+      await page.evaluateOnInitialized(function(){
+        window.injected = 123;
+      });
+      await page.navigate(STATIC_PREFIX + '/tamperable.html');
+      expect(await page.evaluate(() => window.result)).toBe(123);
+    }));
+  });
+
+  describe('Page.printToPDF', function() {
+    let outputFile = __dirname + '/assets/output.pdf';
```
aslushnikov, line 718: i wonder if pdf-generation is platform-dependent (modulo fonts?). Maybe we can just compare pdf to golden?

lib/Page.js
```diff
@@ -285,8 +285,8 @@ class Page extends EventEmitter {
         fitWindow: false
       }),
       this._client.send('Emulation.setVisibleSize', {
-        width: width / zoom,
-        height: height / zoom,
+        width: Math.floor(width / zoom),
```
aslushnikov, line 288: nicee

test/test.js
JoelEinbinder: the pdf files have the creation date and other meta data embedded in them. I was thinking we could render a png of the pdf with puppeteer, but I couldn't get the pdf viewer working in headless mode.

test/test.js
JoelEinbinder: Done.

PR #64, opened 2017-07-08
Fix watchdog timer cleanup

cyrus-and:
Fix a bug in f7cd0048af54d44e0c9b2a4f66eb5136785d57a5 which prevents the cleanup of the watchdog timer, thus keeping the Node.js process alive for `this._maxTime` msec no matter what.


aslushnikov:
Good catch!

PR #67, opened 2017-07-10
Fix docs typo in page.screenshot description

ngryman:
Hey, just a small typo, I think you meant `.png`.


aslushnikov:
Right, it should be `.png`. Could you please sign CLA so that the bot is happy, and we're good to go

paulirish:
Technically the CLA is not necessary for markdown, as it's not "code", but it would be nice to have it signed for the future. :)

ngryman:
I signed it! 🕺

Issue #40, opened 2017-06-27
Add fancy input APIs


paulirish:
```js
.mouseMoved(x, y, options = {})
.mousePressed(x, y, options = {})
.mouseReleased(x, y, options = {})
.tap(x, y, options = {})
.touchmove()
.touchend()
```

JoelEinbinder:
For tapping:
`Input.synthesizeTapGesture` appears to only work on android
`Input.dispatchTouchEvent` doesn't click events like normal tapping would
`Input.emulateTouchFromMouseEvent` fires mouse events, but only after an undefined delay.

The best way to emulate touch right now appears to be setting the device mode to iPhone, which sends touchStart and touchEnd along with every mouse press.

JoelEinbinder:
Closing this in favor of smaller issues

#156 
#157 
#158

PR #69, opened 2017-07-11
Mouse input

JoelEinbinder:
I don't think this is going to be the final API, but it's a start. #40

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -21,6 +21,7 @@
   * [page.httpHeaders()](#pagehttpheaders)
   * [page.injectFile(filePath)](#pageinjectfilefilepath)
   * [page.mainFrame()](#pagemainframe)
+  * [page.mouse()](#pagemouse)
   * [page.navigate(url)](#pagenavigateurl)
   * [page.plainText()](#pageplaintext)
   * [page.printToPDF(filePath[, options])](#pageprinttopdffilepath-options)
@@ -38,6 +39,11 @@
   * [page.userAgent()](#pageuseragent)
   * [page.viewportSize()](#pageviewportsize)
   * [page.waitFor(selector)](#pagewaitforselector)
+- [class: Mouse](#class-mouse)
+  * [mouse.click(button)](#mouseclickbutton)
+  * [mouse.move(x, y)](#mousemovex-y)
+  * [mouse.press(button)](#mousepressbutton)
+  * [mouse.release(button)](#mousereleasebutton)
 - [class: Dialog](#class-dialog)
   * [dialog.accept()](#dialogaccept)
   * [dialog.dismiss()](#dialogdismiss)
@@ -162,6 +168,10 @@ Pages could be closed by `page.close()` method.
 
 #### page.mainFrame()
 
+#### page.mouse()
+
+- returns: <[Mouse]>
+
 #### page.navigate(url)
 
 - `url` <[string]> URL to navigate page to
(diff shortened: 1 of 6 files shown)


aslushnikov:
Can we pass phantom's tests with this?

JoelEinbinder:
Phantom tests modifier keys, which I don't duplicate yet. I am debating between
```js
mouse.press({button: "left", shiftKey: "true"});
```
vs.
```js
keyboard.press("Shift");
mouse.press("left");
keyboard.release("Shift")
```

lib/Input.js
```diff
@@ -0,0 +1,76 @@
+/**
```
aslushnikov, line 1: Shall we call the file "Mouse.js"?

lib/Input.js
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * @param {string=} button
+   */
+  async click(button) {
+    this.press(button);
+    await this.release(button);
+  }
+
+  async press(button = 'left') {
```
aslushnikov, line 53: could you please add jsdoc here

lib/Input.js
```diff
@@ -0,0 +1,76 @@
...
+    await this._client.send('Input.dispatchMouseEvent', {
+      type: 'mousePressed',
+      button,
+      x: this._x,
+      y: this._y,
+      clickCount: 1
+    });
+  }
+
+  async release(button = 'left') {
```
aslushnikov, line 63: and here as well

lib/Input.js
```diff
@@ -0,0 +1,76 @@
...
+      type: 'mouseReleased',
+      button,
+      x: this._x,
+      y: this._y,
+      clickCount: 1
+    });
+  }
+}
+
+module.exports = {
```
aslushnikov, line 74: it looks like you're planning to have other api's here as well. Is there a strong need for it? (e.g. accessing private fields of one-another)

lib/Page.js
```diff
@@ -506,29 +509,32 @@ class Page extends EventEmitter {
...
       nodeId: await this._querySelector(selector)
     })).model.content;
-    let x = Math.round((boxModel[0] + boxModel[4]) / (2 * this._screenDPI));
-    let y = Math.round((boxModel[1] + boxModel[5]) / (2 * this._screenDPI));
-
-    this._client.send('Input.dispatchMouseEvent', {
-      type: 'mouseMoved',
+    let x = Math.round((boxModel[0] + boxModel[4]) / 2);
+    let y = Math.round((boxModel[1] + boxModel[5]) / 2);
+    return {
```
aslushnikov, line 518: nit: let's have this on a single line

```js
return {x, y};
```

lib/Page.js
```diff
@@ -506,29 +509,32 @@ class Page extends EventEmitter {
    * @param {string} selector
    * @param {!Promise}
```
aslushnikov, line 510: let's add a return jsdoc while we're here

test/test.js
```diff
@@ -27,7 +27,7 @@ let EMPTY_PAGE = STATIC_PREFIX + '/empty.html';
 if (process.env.DEBUG_TEST)
   jasmine.DEFAULT_TIMEOUT_INTERVAL = 1000 * 1000 * 1000;
 else
-  jasmine.DEFAULT_TIMEOUT_INTERVAL = 10 * 1000;
+  jasmine.DEFAULT_TIMEOUT_INTERVAL = 100000 * 1000;
```
aslushnikov, line 30: this doesn't look right

lib/Page.js
```diff
@@ -506,29 +509,32 @@ class Page extends EventEmitter {
...
    */
-  async click(selector) {
+  async _positionOfNode(selector) {
```
paulirish, line 512: this technically returning the centroid rather than the position. i think its worth renaming to not confuse when we need the bounding box position.

test/test.js
paulirish, line 30: you can pass a second argument to `it()` as a custom timeout for that specific test, if that's useful.

docs/api.md
```diff
@@ -162,6 +168,10 @@ Pages could be closed by `page.close()` method.
 
 #### page.mainFrame()
 
+#### page.mouse()
```
paulirish, line 171: Should add a note in the docs here to indicate that this is a lower level primitive and for common cases, see `page.click()`..

PR #71, opened 2017-07-11
Disable DOM domain after each operation.

pavelfeldman:


paulirish:
Because I'm curious.. can you explain why this is necessary?

pavelfeldman:
DOM domain was made for inspection, not automation. Backend for this domain is stateful and stores information on the nodes it sent to the client. It also pushes all the nodes that are on the path to any `nodeId` it gives you. And it will also send updates on all the changes to any of those nodes.

In automation, we never want to deal with `nodeId`s, we need simple `backendNodeId` handles to nodes. No paths, no state, no updates, nothing. Just a weak handle to a node so that you could perform operations on it. My plan is to migrate as much of the DOM domain methods to backendNodeIds. In this case you would be able to fetch the box model, focus, populate the node, all without enabling the DOM domain and touching `nodeId`. We might want to do this sooner rather than later...

aslushnikov:
cc @JoelEinbinder

PR #72, opened 2017-07-11
Capture native resolution while emulating.

pavelfeldman:


lib/Browser.js
```diff
@@ -57,7 +57,6 @@ class Browser {
       let chromiumRevision = require('../package.json').puppeteer.chromium_revision;
       let revisionInfo = Downloader.revisionInfo(Downloader.currentPlatform(), chromiumRevision);
       console.assert(revisionInfo, 'Chromium revision is not downloaded. Run npm install');
-      this._chromeExecutable = revisionInfo.executablePath;
```
aslushnikov: this is still needed

lib/EmulatedDevice.js
```diff
@@ -0,0 +1,199 @@
+// Copyright 2014 The Chromium Authors. All rights reserved.
```
aslushnikov: I believe this should go under `/third_party/chromium`

lib/Page.js
```diff
@@ -260,51 +257,45 @@ class Page extends EventEmitter {
   /**
    * @param {string} html
    * @param {!Object=} options
-   * @return {!Promise<!Response>}
```
aslushnikov: we should keep returning the response here

lib/Page.js
```diff
@@ -362,7 +421,7 @@ class Page extends EventEmitter {
       console.assert(Number.isInteger(options.quality), 'Expected options.quality to be an integer');
       console.assert(options.quality >= 0 && options.quality <= 100, 'Expected options.quality to be between 0 and 100 (inclusive), got ' + options.quality);
     }
-    console.assert(!options.clip || !options.fullPage, 'options.clip and options.fullPage are exclusive');
+    console.assert(!options.clip || !options['fullPage'], 'options.clip and options.fullPage are exclusive');
```
aslushnikov: let's be consistent here and use `.fullPage`

lib/Page.js
```diff
@@ -260,51 +257,45 @@ class Page extends EventEmitter {
...
-
+    this._networkManager.on(NetworkManager.Events.Response, onResponse);
+    await this._navigator.navigate();
     this._networkManager.removeListener(NetworkManager.Events.Response, onResponse);
     const response = responses.get(this.mainFrame().url());
     console.assert(response);
+
+    // Await for a single raf rountrip to ensure basic rasterization is complete.
+    await this.evaluate(() => {
+      let callback;
```
aslushnikov:
```js
return new Promise(fulfill => requestAnimationFrame(fulfill));
```

lib/Page.js
```diff
@@ -260,51 +257,45 @@ class Page extends EventEmitter {
...
-    /** @type {!Map<string, !Response>} */
+    const referrer = this._networkManager.httpHeaders()['referer'];
+    this._navigator = new Navigator(this._client, url, referrer, options);
+    return this.reload();
+  }
+
+  /**
+   * @return {!Promise<!Response>}
+   */
+  async reload() {
```
aslushnikov, line 268: can we have reload which just navigates to the same url?
```js
async reload(options) {
    await this.navigate(this.url(), options);
}
```

lib/Page.js
```diff
@@ -509,36 +550,31 @@ class Page extends EventEmitter {
    * @param {!Promise}
    */
   async click(selector) {
-    let center = await this.evaluate(selector => {
```
aslushnikov: this version of the method is better since it doesn't use DOM domain

lib/Page.js
```diff
@@ -257,45 +258,67 @@ class Page extends EventEmitter {
...
    * @return {!Promise}
    */
-  setViewportSize(size) {
-    this._viewportSize = size;
-    return this._resetDeviceEmulation();
+  async setViewport(viewport) {
+    this._viewport = viewport;
+    const needsReload = await EmulationManager.emulateViewport(this._client, viewport);
+    if (needsReload)
+      return this.reload();
```
aslushnikov: await this.reload()

lib/Page.js
```diff
@@ -257,45 +258,67 @@ class Page extends EventEmitter {
...
   }
 
   /**
    * @return {!{width: number, height: number}}
```
aslushnikov: returns Page.Viewport

lib/EmulationManager.js
```diff
@@ -0,0 +1,193 @@
...
+      deviceScaleFactor: device.deviceScaleFactor,
+      isMobile: device.capabilities.includes('mobile'),
+      hasMobile: device.capabilities.includes('touch'),
+      isLandscape: options.landscape || false
+    };
+  }
+
+  /**
+   * @param {string} name
+   * @return {string}
```
aslushnikov: remove options

lib/EmulationManager.js
```diff
@@ -0,0 +1,193 @@
...
+    await Promise.all([
+      client.send('Emulation.setDeviceMetricsOverride', { mobile, width, height, deviceScaleFactor, screenOrientation }),
+      client.send('Emulation.setTouchEmulationEnabled', {
+        enabled: viewport.hasTouch || false,
+        configuration: viewport.isMobile ? 'mobile' : 'desktop'
+      })
+    ]);
+
+    let reloadNeeded = false;
+    if (viewport.hasTouch && !EmulationManager._emulatingTouch) {
```
aslushnikov: drop _emulatingTouch so that it works in multitab world

lib/EmulationManager.js
```diff
@@ -0,0 +1,193 @@
...
+      const source = `(${injectedTouchEventsFunction})()`;
+      EmulationManager._emulatingTouch = await client.send('Runtime.addScriptToEvaluateOnNewDocument', { source });
+      reloadNeeded = true;
+    }
+
+    if (!viewport.hasTouch && EmulationManager._emulatingTouch) {
+      await client.send('Runtime.removeScriptToEvaluateOnNewDocument', EmulationManager._emulatingTouch);
```
aslushnikov: EmulationManager._emulatingTouch = false;

PR #73, opened 2017-07-12
Cleanup _frameIdToExecutionContextId properly

alexkozy:


lib/FrameManager.js
```diff
@@ -110,7 +110,17 @@ class FrameManager extends EventEmitter {
   }
 
   _onExecutionContextDestroyed(executionContextId) {
-    this._frameIdToExecutionContextId.delete(executionContextId);
+    let frameId = this._frameId(executionContextId);
```
aslushnikov: i think we don't need the `_onExecutionContextDestroyed` listener at all.

Instead, let's call the `this._frameIdToExecutionContextId.delete(frameId)` next to every call to the  `this._frames.delete(frameId)` (there are two of them). This will avoid the unfortunate  linear traversal.

alexkozy: done.

PR #75, opened 2017-07-12
$ and $$ as shortcuts for querySelector and querySelectorAll

JoelEinbinder:
This version of the patch uses `document.querySelector` instead of `DOM.querySelector`


docs/api.md
```diff
@@ -319,6 +341,18 @@ Shortcut for [page.mainFrame().waitFor(selector)](#framewaitforselector).
 Dialog's type, could be one of the `alert`, `beforeunload`, `confirm` and `prompt`.
 
 ### class: Frame
+#### frame.$(selector, fun, ...args)
+- `selector` <[string]> Query selector to be run on the page
+- `fun` <[function<[Element]>]> Function to be evaluated with first element matching `selector`
+- `...args` <[Array]<[string]>> Arguments to pass to `fun`
```
aslushnikov: in `page.evaluate`, this is documented with type `<...[string]>`. Not sure which one is better, but let's be consistent

lib/FrameManager.js
```diff
@@ -159,40 +159,27 @@ class FrameManager extends EventEmitter {
    * @return {!Promise<(!Object|undefined)>}
    */
   async _evaluateOnFrame(frame, fun, ...args) {
```
aslushnikov: let's inline _evaluateOnFrame into Frame.evaluate

lib/FrameManager.js
```diff
@@ -159,40 +159,27 @@ class FrameManager extends EventEmitter {
...
+    return this._evaluateStringOnFrame(frame, helper.evaluationString(fun, ...args));
+  }
+
+  /**
+   * @param {!Frame} frame
+   * @param {string} expression
+   * @return {!Promise<(!Object|undefined)>}
+   */
+  async _evaluateStringOnFrame(frame, expression) {
```
aslushnikov: and rename this into `_evaluateOnFrame`

lib/Page.js
```diff
@@ -506,6 +506,39 @@ class Page extends EventEmitter {
 
   /**
    * @param {string} selector
+   * @param {!Promise<!Array<number>>}
+   */
+  async _querySelectorAll(selector) {
```
pavelfeldman: Why do we need these?

lib/FrameManager.js
JoelEinbinder: Done.

lib/FrameManager.js
JoelEinbinder: Done.

docs/api.md

aslushnikov:
Could you please rebaseline atop of master?
Other then that, lgtm!

test/doclint/MDBuilder.js
```diff
@@ -52,7 +52,7 @@ class MDOutline {
     this.errors = [];
     const classHeading = /^class: (\w+)$/;
     const constructorRegex = /^new (\w+)\((.*)\)$/;
-    const methodRegex = /^(\w+)\.(\w+)\((.*)\)$/;
+    const methodRegex = /^(\w+)\.([\w$]+)\((.*)\)$/;
```
aslushnikov: could you please add a doclint test for this

JoelEinbinder:
Rebased and added a test.

lib/FrameManager.js
```diff
@@ -154,45 +154,22 @@ class FrameManager extends EventEmitter {
...
-    }
-    if (!remoteObject.objectId)
-      return remoteObject.value;
-    let response = await this._client.send('Runtime.callFunctionOn', {
-      objectId: remoteObject.objectId,
-      functionDeclaration: 'function() { return this; }',
-      returnByValue: true,
-    });
-    return response.result.value;
+    return helper.serializeRemoteObjet(this._client, remoteObject);
```
aslushnikov: nit: there's a typo in serializeRemoteObje**C**t

It also looks like this helper method is not needed at all and we can keep it inlined.

docs/api.md
```diff
@@ -147,6 +151,24 @@ browser.newPage().then(page => {
 ```
 Pages could be closed by `page.close()` method.
 
+#### page.$(selector, fun, ...args)
+
+- `selector` <[string]> Query selector to be run on the page
```
paulirish: > A selector to be matched in the page

pavelfeldman:
How do I use it?

docs/api.md
```diff
@@ -147,6 +151,24 @@ browser.newPage().then(page => {
...
+- `fun` <[function<[Element]>]> Function to be evaluated with first element matching `selector`
```
paulirish: Generally I see a lot more `fn` or even `func` than `fun` which is basically 🎉 .

Somewhat separately, I think we should start calling out functions that are evaluated clientside.. like `clientsideFunction`

So i'd go with 

```md
 - `fn` <[clientsideFunction<[Element]>]> Function to be evaluated in-page with the matched element`
```

docs/api.md
```diff
@@ -147,6 +151,24 @@ browser.newPage().then(page => {
...
+- `...args` <...[string]> Arguments to pass to `fun`
+- returns: <[Promise<[Object]]> Promise which resolves to function return value.
+
```
paulirish: lets add an example?

```js
const outerhtml = await page.$('#box', e => e.outerHTML);
```

docs/api.md
```diff
@@ -147,6 +151,24 @@ browser.newPage().then(page => {
...
+Shortcut for [page.mainFrame().$(selector, fun, ...args)](#pageselector-fun-args).
+
+#### page.$$(selector, fun, ...args)
+
+- `selector` <[string]> Query selector to be run on the page
```
paulirish: > A selector to be matched in the page

docs/api.md
```diff
@@ -147,6 +151,24 @@ browser.newPage().then(page => {
...
+- `fun` <[function<[Element]>]> Function to be evaluted for every element matching `selector`.
```
paulirish:
```md
- `mapFn` <[clientsideFunction<[Element]>]> Function to be evaluated in-page for every matching element.
```

docs/api.md
```diff
@@ -147,6 +151,24 @@ browser.newPage().then(page => {
...
+- `...args` <...[string]> Arguments to pass to `fun`
+- returns: <[Promise<[Array<[Object]>]>]> Promise which resolves to array of function return values.
+
```
paulirish: Example:

```js
const headings = await page.$$('h1,h2,h3,h4', el => el.innerText);
for (const heading of headings) console.log(heading);
```

paulirish (changes requested):
i have a few suggestions for clarity of the docs.
i expect these to be pretty primary methods, so they can use some love.

aslushnikov:
@JoelEinbinder let's add examples to the documentation!

@pavelfeldman:
To get inner HTML of the page (as requested in https://github.com/GoogleChrome/puppeteer/issues/78):
```js
const html = await page.$('body', body => body.innerHTML);
```
To get a bounding box:
```js
const box = await page.$('#button', button => {
  const box = button.getBoundingClientRect();
  return {x: box.left, y: box.top};
});
```
To get contents of all `h3` headers:
```js
const headers = await page.$$('h3', h => h.textContent);
```
To draw borders around all elements:
```js
await page.$$('*', e => {
  e.style.setProperty('border', '1px solid blue');
});
```

docs/api.md
aslushnikov: Good call. Let's do this in a separate patch throughout the doc. The change would also require js changes to please doclint.

aslushnikov: As discussed offline, named them all `pageFunction` throughout the doc in 
76415e1be537e8d56960ddd985c487a948647cbc

docs/api.md
JoelEinbinder: We are calling it pageFunction now. I'll fix this up in another patch that explains how our functions are executed client side.

docs/api.md

docs/api.md

docs/api.md

docs/api.md

JoelEinbinder:
@paulirish I have a scary red X (requested changes) from you. What do I have to do to get rid of it?

lib/helper.js
```diff
@@ -52,6 +52,36 @@ class Helper {
     }
     return message;
   }
+
+  /**
+   * @param {!Connection} client
+   * @param {!Object} remoteObject
+   */
+  static async serializeRemoteObjet(client, remoteObject) {
```
paulirish: still have this typo. serializeRemoteObjet => serializeRemoteObject

paulirish (approved):
lgtm minus spelling nit.

lib/helper.js
JoelEinbinder: This was merge problem. Whole method doesn't exist anymore.

PR #82, opened 2017-07-14
Addded builtins object

alexkozy:
We can store required methods to this object before any JavaScript execution and use them later, e.g. in case of waitFor we can store querySelector.

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -44,6 +44,16 @@ class Page extends EventEmitter {
     let page = new Page(client, frameManager, networkManager, screenDPI);
     // Initialize default page size.
     await page.setViewportSize({width: 400, height: 300});
+    // Initialize builtins.
+    function createBuiltins() {
+      // eslint-disable-next-line no-console
+      console.debug('driver:Builtins', {
+        querySelector: document.querySelector.bind(document),
+        __proto__: null
+      });
+    }
+    await page.evaluate(createBuiltins);
+    await page.evaluateOnInitialized(createBuiltins);
     return page;
   }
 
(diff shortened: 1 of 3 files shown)


lib/Page.js
```diff
@@ -44,6 +44,16 @@ class Page extends EventEmitter {
     let page = new Page(client, frameManager, networkManager, screenDPI);
     // Initialize default page size.
     await page.setViewportSize({width: 400, height: 300});
+    // Initialize builtins.
+    function createBuiltins() {
+      // eslint-disable-next-line no-console
+      console.debug('driver:Builtins', {
+        querySelector: document.querySelector.bind(document),
```
pavelfeldman, line 51: why not simply return the object with builtins?

lib/Page.js
```diff
@@ -216,6 +226,10 @@ class Page extends EventEmitter {
       }
       return;
     }
+    if (event.type === 'debug' && event.args.length > 1 && event.args[0].value === 'driver:Builtins') {
```
pavelfeldman, line 229: ... then you would not need this.

lib/Page.js
```diff
@@ -216,6 +226,10 @@ class Page extends EventEmitter {
...
+      this._frameManager.setBuiltins(event.executionContextId, event.args[1].objectId);
```
pavelfeldman, line 230: ... and we should do this all in the FrameManager

lib/Page.js
pavelfeldman: also, how is this going to work post cross-process navigation? the only way to resolve this seems to be using addScriptToEvaluateOnNewDocument, which would return values as events. Which is too much complexity.

alexkozy: I'm passing this function to evaluateOnInitialize. And as far as I know console API call is only one way to get any value back from evaluateOnInitialize.
I don't know enough details about cross-process navigation to see an issue. Could you explain please?

lib/Page.js
alexkozy: I need to call evaluateOnInitialize to run builtin-related code in each frame.

PR #87, opened 2017-07-17
Roll chromium to r486981.

pavelfeldman:


lib/Page.js
```diff
@@ -371,48 +360,33 @@ class Page extends EventEmitter {
...
+      const mobile = this._emulatedDevice ? this._emulatedDevice.capabilities.includes(EmulatedDevice.Capability.Mobile) : false;
+      const deviceScaleFactor = this._emulatedDevice ? this._emulatedDevice.deviceScaleFactor : 1;
+      const landscape = this._emulatedDeviceOptions && this._emulatedDeviceOptions['orientation'] === 'landscape';
+      const screenOrientation = landscape ? { angle: 90, type: 'landscapePrimary' } : { angle: 0, type: 'portraitPrimary' };
+      await this._client.send('Emulation.setDeviceMetricsOverride', { mobile, width, height, deviceScaleFactor, screenOrientation });
     }
+
+    let clip = options['clip'];
+    if (clip)
+      clip.scale = clip.scale || 1;
```
aslushnikov: This mutates user's object

lib/Page.js
```diff
@@ -260,19 +256,13 @@ class Page extends EventEmitter {
...
-      this._networkManager.on(NetworkManager.Events.Response, onResponse);
-      await navigator.navigate();
-    } catch (e) {
-      this._networkManager.removeListener(NetworkManager.Events.Response, onResponse);
-      throw e;
-    }
-
+    const navigator = new Navigator(this._client, url, referrer, options);
+    this._networkManager.on(NetworkManager.Events.Response, onResponse);
+    await navigator.navigate();
```
aslushnikov: the `navigator.navigate()` might throw an exception, e.g. in case of malformed URL. The try-catch block is needed to handle the case

lib/Page.js
```diff
@@ -311,6 +285,21 @@ class Page extends EventEmitter {
     return this._viewportSize;
   }
 
+  resetDeviceEmulation() {
```
aslushnikov: can we inline this for now?

lib/Page.js
```diff
@@ -311,6 +285,21 @@ class Page extends EventEmitter {
...
+    const width = this._viewportSize.width;
+    const height = this._viewportSize.height;
+    const deviceScaleFactor = 1;
+    const mobile = false;
+    const screenOrientation = { angle: 0, type: 'portraitPrimary' };
+    const userAgent = '';
+    this._emulatedDevice = null;
+    return Promise.all([
+      this._client.send('Emulation.setDeviceMetricsOverride', { mobile, width, height, deviceScaleFactor, screenOrientation }),
+      this._client.send('Network.setUserAgentOverride', { userAgent }),
```
aslushnikov: do we really need this and the following calls?

PR #90, opened 2017-07-18
frame.waitFor should work with setting innerHTML

JoelEinbinder:


lib/FrameManager.js
```diff
@@ -178,26 +178,25 @@ class FrameManager extends EventEmitter {
...
+      let promise = new Promise(fulfill => {
+        if (document.querySelector(selector)) {
+          fulfill();
+          return;
         }
-      }).observe(document.documentElement, {
-        childList: true,
-        subtree: true
+        new MutationObserver((mutations, observer) => {
+          if (document.querySelector(selector)) {
```
aslushnikov, line 188: Let's have this throttled via setTimeout(100) so that it's not *that* intense.

JoelEinbinder: I thought mutation observer did that for us? Or maybe that was just intersection observer?

aslushnikov: AFAIR mutationObserver is throttled per microtask

JoelEinbinder: A 100 ms delay seems pretty expensive for `waitFor`. Maybe there is a better way?

aslushnikov: Like, 50ms?.. 

Talking seriously, I don't see how we can avoid `document.querySelector`.
> A 100 ms delay seems pretty expensive for waitFor

Do you mean that we risk to miss on an element "blinking" in DOM?

aslushnikov: Let's have this in. The `page.waitFor` method should be reliable so that it's possible to use it for automation, timeout'ing would make the outcome unpredictable. 

If we want to save CPU cycles, we should introduce the `page.waitForClass` / `page.waitForId` methods.

PR #94, opened 2017-07-18
Introduce waitForNavigation.

pavelfeldman:


lib/Navigator.js
```diff
@@ -33,9 +33,6 @@ class Navigator {
     console.assert(this._waitUntil === 'load' || this._waitUntil === 'networkidle', 'Unknown value for options.waitUntil: ' + this._waitUntil);
   }
 
-  /**
```
aslushnikov: it should still return a promise

lib/Page.js
```diff
@@ -266,13 +266,22 @@ class Page extends EventEmitter {
...
     return this.reload();
   }
 
   /**
+   * @param {!Object=} options
+   * @return {!Promise<!Response>}
+   */
+  waitForNavigation(options) {
+    this._navigator = new Navigator(this._client, undefined, undefined, options);
+    return this.reload();
```
aslushnikov: This reload doesn't look right. Why do we force loading while waiting for navigation?
It's also scary that `waitForNavigation` is stateful and will affect subsequent `reload` calls (if any).

PR #96, opened 2017-07-19
page.press and more tests

JoelEinbinder:


aslushnikov (commented):
Thanks, let's remove `keyboard.press()` and `keyboard.type()` to keep only low-level api in keyboard.

PR #98, opened 2017-07-19
Remove keyboard.type and keyboard.press

JoelEinbinder:
This leaves `type` and `press` on the page.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -576,6 +576,10 @@ In case of multiple pages in one browser, each page can have its own viewport si
 - `text` <[string]> A text to type into a focused element.
 - returns: <[Promise]> Promise which resolves when the text has been successfully typed.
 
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -625,17 +615,6 @@ Dispatches a `keypress` and `input` event. This does not send a `keydown` or `ke
 page.keyboard.sendCharacter('嗨');
 ```
 
-#### keyboard.type(text)
-- `text` <[string]> Text to type into the page
-- returns: <[Promise]>
-
-Sends a `keydown`, `keypress`/`input`, and `keyup` event for each character in the text.
```
aslushnikov, line 678: let's have this documentation in page.type()

JoelEinbinder: Done.

PR #101, opened 2017-07-21
Mouse

JoelEinbinder:
#40


lib/Mouse.js
```diff
@@ -0,0 +1,106 @@
...
+    await this.up(options);
+  }
+
+  /**
+   * @param {!Object=} options
+   */
+  async down(options) {
+    if (!options)
+      options = {};
+    this._button = 'left';
```
aslushnikov: we should assign from options here

lib/FrameManager.js
```diff
@@ -406,6 +409,39 @@ class Frame {
...
+
+  /**
+   * @param {string} selector
+   * @param {!Object=} options
+   * @return {!Promise}
+   */
+  async click(selector, options) {
+    await this.hover(selector);
+    await this._frameManager._mouse.press(options);
+    // This is a hack for now, to make clicking less race-prone
```
aslushnikov, line 440: Let's file an upstream bug for the protocol about mouseUp/mouseDown events methods which return too early.

Let's refer to the bug here and explain in more details that mouse.down() and mouse.up() rely on a racy protocol methods

aslushnikov (commented):
Awesome! Let's also have phantom_shim fixed

lib/Mouse.js
JoelEinbinder: Done.

JoelEinbinder:
Pass the phantom tests

Issue #122, opened 2017-07-25
mouse.down() and mouse.up() should wait for raf


aslushnikov:
Currently, only `frame.click()` waits for raf to ensure event delivery.
However, the core issue is lurking inside `mouse.up` / `mouse.down`. We should move the hack down to the methods (and refer to the https://bugs.chromium.org/p/chromium/issues/detail?id=747647)

PR #123, opened 2017-07-25
Wait for raf after all mouse events.

JoelEinbinder:
#122


lib/Mouse.js
```diff
@@ -41,6 +41,8 @@ class Mouse {
       x, y,
       modifiers: this._modifiersMask()
     });
+    // This is a hack for now, to make clicking less race-prone. See crbug.com/747647
+    await this.evaluate(() => new Promise(f => requestAnimationFrame(f)));
```
aslushnikov: is this needed? AFAIU the mouseMove does wait for ack: [input_handler.cc](https://cs.chromium.org/chromium/src/content/browser/devtools/protocol/input_handler.cc?q=input_handler.cc&sq=package:chromium&dr&l=403)

JoelEinbinder: In practice it behaves otherwise, at least as far as I can tell.

aslushnikov:
Exploring the codesearch, I failed to find the place where the mousemove swallows the ACK. It might be that we're facing something else here - let's investigate before landing

lib/Mouse.js
```diff
@@ -69,7 +73,11 @@ class Mouse {
       clickCount: (options.clickCount || 1)
     });
     // This is a hack for now, to make clicking less race-prone. See crbug.com/747647
-    await this.evaluate(() => new Promise(f => requestAnimationFrame(f)));
+    await this._client.send('Runtime.evaluate', {
+      expression: 'new Promise(f => requestAnimationFrame(f))',
```
aslushnikov: let's do rafraf!

PR #128, opened 2017-07-25
Move inner evaluate helpers into frame class.

pavelfeldman:


lib/FrameManager.js
```diff
@@ -108,8 +105,11 @@ class FrameManager extends EventEmitter {
   }
 
   _onExecutionContextCreated(context) {
-    if (context.auxData && context.auxData.isDefault && context.auxData.frameId)
-      this._frameIdToExecutionContextId.set(context.auxData.frameId, context.id);
+    const frameId = context.auxData && context.auxData.isDefault ? context.auxData.frameId : null;
+    const frame = this._frames.get(frameId);
+    if (!frame)
+      return;
```
JoelEinbinder, line 106: Can we get the executionContextCreated before the frameAttached? I vaguely remember that that that could happen.

PR #129, opened 2017-07-25
Simplify keyboard modifiers handling

pavelfeldman:


lib/Input.js
```diff
@@ -45,26 +44,29 @@ class Keyboard {
...
-    };
+  _modifierBit(key) {
+    if (key === 'Alt')
+      return 1;
+    if (key === 'Control')
+      return 2;
+    if (key === 'Meta')
+      return 4;
+    if (key === 'Shift')
+      return 8;
```
JoelEinbinder, line 58: it would make me feel safer to return 0 here.

PR #135, opened 2017-07-25
page.evaluate takes a string in addition to function

### LINKED ISSUE #118: page.evaluate should accept a string as an argument
^

JoelEinbinder:
closes #118

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -354,14 +354,14 @@ Adds a `<script></script>` tag to the page with the desired url. Alternatively, 
 - returns: <[Promise]> Returns promise which resolves when page gets closed.
 
 #### page.evaluate(pageFunction, ...args)
-- `pageFunction` <[function]> Function to be evaluated in browser context
+- `pageFunction` <[function]|[string]> Function to be evaluated in browser context
 - `...args` <...[string]> Arguments to pass to  `pageFunction`
 - returns: <[Promise]<[Object]>> Promise which resolves to function return value
 
 This is a shortcut for [page.mainFrame().evaluate()](#frameevaluatefun-args) method.
 
 #### page.evaluateOnNewDocument(pageFunction, ...args)
-- `pageFunction` <[function]> Function to be evaluated in browser context
+- `pageFunction` <[function]|[string]> Function to be evaluated in browser context
 - `...args` <...[string]> Arguments to pass to `pageFunction`
 - returns: <[Promise]<[Object]>> Promise which resolves to function
 
@@ -825,7 +825,7 @@ Adds a `<script></script>` tag to the frame with the desired url. Alternatively,
 - returns: <[Promise]> Promise which resolves when the element matching `selector` is successfully clicked. Promise gets rejected if there's no element matching `selector`.
 
 #### frame.evaluate(pageFunction, ...args)
-- `pageFunction` <[function]> Function to be evaluated in browser context
+- `pageFunction` <[function]|[string]> Function to be evaluated in browser context
 - `...args` <...[string]> Arguments to pass to  `pageFunction`
 - returns: <[Promise]<[Object]>> Promise which resolves to function return value
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -358,6 +358,26 @@ Adds a `<script></script>` tag to the page with the desired url. Alternatively, 
 - `...args` <...[string]> Arguments to pass to  `pageFunction`
 - returns: <[Promise]<[Object]>> Promise which resolves to function return value
 
+If the function, passed to the `page.evaluate`, returns a [Promise], then `page.evaluate` would wait for the promise to resolve and return it's value.
+
+```js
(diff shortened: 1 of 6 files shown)


docs/api.md
```diff
@@ -363,14 +363,14 @@ Adds a `<script></script>` tag to the page with the desired url. Alternatively,
 - returns: <[Promise]> Returns promise which resolves when page gets closed.
 
 #### page.evaluate(pageFunction, ...args)
-- `pageFunction` <[function]> Function to be evaluated in browser context
+- `pageFunction` <[function]|[string]> Function to be evaluated in browser context
```
aslushnikov, line 357: - let's rename the argument into `pageFunctionOrString`
- let's throw if there are any additional arguments and the first argument is a string

lib/helper.js
```diff
@@ -16,11 +16,13 @@
...
 class Helper {
   /**
-   * @param {function()} fun
+   * @param {function()|string} fun
    * @param {!Array<*>} args
    * @return {string}
    */
   static evaluationString(fun, ...args) {
+    if (typeof fun === 'string')
+      return 'eval(`' + fun.replace(/\\/g, '\\\\').replace(/`/g, '\\`') + '`)';
```
paulirish: i think you might want to use `new Function()` here instead.

eval works with the current execution scope and can access local variables. but new Function cannot.

JoelEinbinder: `new Function()` doesn't return the statement completion value: https://twitter.com/paul_irish/status/818568243350630401

phantom_shim/WebPage.js
```diff
@@ -535,6 +535,8 @@ class WebPage {
    * @param {!Array<!Object>} args
    */
   evaluate(fun, ...args) {
+    if (typeof fun === 'string')
```
JoelEinbinder, line 540: This was working unintentionally before as a side of effect of wrapping things in `Promise.resolve`

docs/api.md
JoelEinbinder: I made it throw. `pageFunctionOrString` looks a big ambiguous. If only took a string, I wouldn't call the paramter `pageString`

lib/helper.js
```diff
@@ -48,6 +53,16 @@ class Helper {
    * @return {!Promise<!Object>}
    */
   static async serializeRemoteObject(client, remoteObject) {
+    if (remoteObject.subtype === 'promise') {
+      let resolvedObject = (await client.send('Runtime.callFunctionOn', {
```
aslushnikov: `Runtime.awaitPromise` would be cleaner

phantom_shim/WebPage.js
aslushnikov: nice, i didn't know that phantom's evaluate accepts string. nit: let's fix jsdoc here

lib/helper.js
```diff
@@ -16,11 +16,16 @@
...
 class Helper {
   /**
-   * @param {function()} fun
+   * @param {function()|string} fun
    * @param {!Array<*>} args
    * @return {string}
    */
   static evaluationString(fun, ...args) {
+    if (typeof fun === 'string') {
+      if (args.length)
```
aslushnikov: nit: let's use `console.assert`
```js
console.assert(!args.length, 'Cannot evaluate a string with arguments');
```

aslushnikov (approved):
Thanks Joel! wdyt about renaming `pageFunction` in `pageFunctionOrString` to make it clear in docs (and in code)?

lib/helper.js
JoelEinbinder: I didn't know we had that. Cool

lib/helper.js
JoelEinbinder: `console.assert` won't throw, I assumed we would want that with everything promise based.

docs/api.md
aslushnikov: ok that's reasonable

lib/helper.js
```diff
@@ -16,11 +16,16 @@
...
+        throw 'Cannot evaluate a string with arguments';
```
Garbee: Wouldn't we want to `throw new Error('message')` instead of just throwing a string? That way any other tooling can easily get the call stack and other details. Or am I missing something about throwing a plain string that still allows the call stack to be retrieved in user-land code?

JoelEinbinder: Nope, that's right. Good catch.

lib/helper.js
aslushnikov: why wouldn't `console.assert` throw?

JoelEinbinder: Apparently node's console.assert throws.

JoelEinbinder:
I switched to `Runtime.awaitPromise`, and fixed up the jsdoc. I added an example of using a string to `page.evaluate` api, and also moved the previous examples from `frame.evaluate` to `page.evaluate`.

PR #139, opened 2017-07-26
Document all options objects wherever they are used

### LINKED ISSUE #114: Options objects should be documented wherever they are used.
In the api, options objects and other parameters should be documented on every method in which they are used, instead of referring you to a different method where they are documented.

JoelEinbinder:
closes #114


aslushnikov:
Thanks!

PR #141, opened 2017-07-26
Trace only string method names

JoelEinbinder:
Method names could be symbol or number, and then startsWith would throw.


aslushnikov:
Where did it hit you?

aslushnikov (approved):

JoelEinbinder:
Nothing yet, it would break on `Symbol.iterator` though.

PR #142, opened 2017-07-26
In page object example

JoelEinbinder:
This example adds a `puppeteer` proxy object into the page.

### DIFF
--- a/examples/inpageobject.js
+++ b/examples/inpageobject.js
@@ -1,0 +1,55 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


Garbee:
While this is super nifty, do you have a use-case for this kind of setup? Flashy but I don't see how it is too useful to developers to include as an example.

JoelEinbinder:
It lets you easily upgrade from a test runner that simply injects javascript into the page. I wrote it for one of my personal projects.

aslushnikov:
Should you rather have this example as a part of a blog post on medium? You would be able to explain the motivation behind it and give a specific practical usecase.

I'd like 'examples' folder to contain practical receipts. We don't have it now, but that might be the ultimate goal.

Garbee:
If it does stay in examples, perhaps a rename to `access-puppeteer-in-page.md` would be better. To more specifically convey the purpose behind the example. `inpageobject.md` is not that descriptive.

I do kind-of agree with this being more of an external blog-post type situation too. I don't feel it would be too common of a thing for developers to do regularly.

JoelEinbinder:
I'm fine with that. I just had some code, figured I would post it.

Issue #106, opened 2017-07-24
Don't use browser Request api, make our own simple one


JoelEinbinder:
@aslushnikov can you fill in why? I just have it in my notes that we should do it.

Garbee:
What exactly would this kind of API look like? I think not using the built in browser systems where available is harmful. As then developers need to learn yet-another-syntax to get the same results. Even though we already have (and are probably using) exactly what they know underneath.

aslushnikov:
> What exactly would this kind of API look like?

We'll get rid of `Headers` object and use something like a map instead. It would be consistent with `page.setHTTPHeaders` methods.

> I think not using the built in browser systems where available is harmful. Even though we already have (and are probably using) exactly what they know underneath.

The API in question is consumed by node.js, not by browser. We don't have [fetch api](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) in node. We have [protocol's network domain](https://chromedevtools.github.io/devtools-protocol/tot/Network/) which we expose via the [puppeteer's request](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-request) and [puppeteer's response](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-response) classes. 

Browser also has [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) and [Response](https://developer.mozilla.org/en-US/docs/Web/API/Response) classes, so we tried to align puppeteer's request and response with the browser counterparts. It turned out that:
- our capabilities are very limited, for example, [puppeteer's request](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-request) implements only 3 our of 11 fields of the [browser's request](https://developer.mozilla.org/en-US/docs/Web/API/Request).
- our usage scenario differs from the browser's one, so we'd like to have a slightly different api, for example [Request.response()](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#requestresponse) method

Since we failed to implement browser api to a reasonable extend, we don't want to commit to it any more.

Garbee:
Ah, thanks for the clarification @aslushnikov. Knowing that it SGTM to make this change.

PR #151, opened 2017-07-26
coverage

### LINKED ISSUE #50: Add tests for all public API methods

JoelEinbinder:
Shows what public methods were not run by the unit tests.
Closes #50

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -13,7 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+let methods = new Set();
+let runMethods = new Set();
+let measuringCoverage = false;
 class Helper {
   /**
    * @param {function()|string} fun
@@ -116,16 +118,19 @@ class Helper {
     let className = classType.prototype.constructor.name;
     className = className.substring(0, 1).toLowerCase() + className.substring(1);
     const debug = require('debug')(`puppeteer:${className}`);
-    if (!debug.enabled)
+    if (!debug.enabled && !measuringCoverage)
       return;
     for (let methodName of Reflect.ownKeys(classType.prototype)) {
       const method = Reflect.get(classType.prototype, methodName);
       if (methodName === 'constructor' || typeof methodName !== 'string' || methodName.startsWith('_') || typeof method !== 'function')
         continue;
+      methods.add(`${className}.${methodName}`);
       Reflect.set(classType.prototype, methodName, function(...args) {
         let argsText = args.map(stringifyArgument).join(', ');
         let callsite = `${className}.${methodName}(${argsText})`;
-        debug(callsite);
+        if (debug.enabled)
+          debug(callsite);
+        runMethods.add(`${className}.${methodName}`);
         return method.call(this, ...args);
       });
     }
@@ -162,6 +167,25 @@ class Helper {
     for (let listener of listeners)
       listener.emitter.removeListener(listener.eventName, listener.handler);
     listeners.splice(0, listeners.length);
+  }
+
+  /**
+   * @return {!Set<string>}
+   */
+  static methodsNotRun() {
+    let out = new Set();
+    for (let method of methods) {
+      if (!runMethods.has(method))
+        out.add(method);
+    }
+    return out;
+  }
+
+  /**
+   * @param {boolean} onOnOff
+   */
+  static toggleCoverage(onOrOff) {
+    measuringCoverage = onOrOff;
   }
 }
 
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -170,15 +170,18 @@ class Helper {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lib/helper.js
```diff
@@ -149,6 +154,25 @@ class Helper {
       listener.emitter.removeListener(listener.eventName, listener.handler);
     listeners.splice(0, listeners.length);
   }
+
+  /**
+   * @return {!Set<string>}
+   */
+  static methodsNotRun() {
```
aslushnikov: let's have a descriptive api for this:
```js
helper.enablePublicAPICoverage(); // enables the coverage data for the helper
helper.getPublicAPICoverage(); // returns the coverage data
```

aslushnikov (commented):
1. Let's force this straight away via a test in the very end of the testsuite. I feel that lint checks which are not enforced are useless

2. We should also trace public events

aslushnikov:
As discussed offline:
- let's pass COVERAGE=true for the `npm test` command. coverage is off for the `npm run unit` command 
- coverage errors should be very clear so that it's not confusing for the developers

JoelEinbinder:
I am trying to write tests for everything that's missing from this, so the coverage test passes. Is there a reason `body.arrayBuffer` exists? It looks redundant with `body.buffer`

What is `frame.name()` supposed to be used for? It seems pretty useless, returning `undefined` for the main frame and then some variant of `frame1` as far as I can tell.

We probably shouldn't have `dialog.accept` nor `dialog.dismiss`  in the api docs if they don't work in headless chrome.

aslushnikov:
> I am trying to write tests for everything that's missing from this, so the coverage test passes. Is there a reason body.arrayBuffer exists? It looks redundant with body.buffer

Nope, that's there for the reasons explained in #106 

> What is frame.name() supposed to be used for? It seems pretty useless, returning undefined for the main frame and then some variant of frame1 as far as I can tell.

That's a frame name, as specified in a tag. I'd keep it since this seems to be a convenient way to address frames.

> We probably shouldn't have dialog.accept nor dialog.dismiss in the api docs if they don't work in headless chrome.

That's a bug in headless, we should fix it.

JoelEinbinder:
> That's a frame name, as specified in a tag. I'd keep it since this seems to be a convenient way to address frames.

Ah, that was created for display purposes for ConsoleContextSelector. It only returns the `name` attribute when the frame is loaded. If the `name` attribute doesn't exist, it returns the `id`. I think having it there will ruin someone's day who tries to use it.

JoelEinbinder:
I made the coverage check mandatory with npm test, and had it generate jasmine tests.

lib/helper.js
```diff
@@ -13,7 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
+let methods = new Set();
+let runMethods = new Set();
+let measuringCoverage = false;
```
aslushnikov: This feels a little hairy to me.
How about a single variable: 
```js
/** @type {?Map<string, boolean>} */
let apiCoverage = null;
```

Enabling coverage would assing the variable to a new Map.

lib/helper.js
```diff
@@ -163,6 +168,28 @@ class Helper {
...
+  static wasPublicAPIMethodCovered(method) {
+    return runMethods.has(method);
+  }
+
+  /**
+   * @return {!Array<string>}
+   */
+  static publicAPIMethods() {
+    return Array.from(methods);
+  }
```
aslushnikov: instead of `publicAPIMethods` and `wasPublicAPIMethodCovered`:

```js
static publicAPICoverage() {
  return apiCoverage;
}
```

test/test.js
```diff
@@ -1345,6 +1348,24 @@ describe('Puppeteer', function() {
   });
 });
 
+if (process.env.COVERAGE) {
```
aslushnikov, line 1351: How about having a single 'coverage' test which reports all the issues:

```js
if (process.env.COVERAGE) {
  describe('COVERAGE', function() {
    it('should cover all public API methods', function() {
      const coverage = helper.publicAPICoverage();
      for (let method of coverage.keys()) {
        if (!coverage.get(method))
          fail('Method not called: ' + method);
      }
    });
  });
}
```

JoelEinbinder: I wanted to have disabled methods show up as disabled in jasmine.

lib/helper.js
JoelEinbinder: Done.

lib/helper.js
JoelEinbinder: Done.

test/test.js
aslushnikov: Ah, I see.

aslushnikov:
Merging this, but we still miss the event tracing.

PR #159, opened 2017-07-27
Simplify frame manager, do not fetch resources upfront.

pavelfeldman:
(this requires branch new chromium that has not baked yet)

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -21,26 +21,15 @@ let helper = require('./helper');
 class FrameManager extends EventEmitter {
   /**
    * @param {!Connection} client
-   * @param {!Mouse} mouse
-   * @return {!Promise<!FrameManager>}
-   */
-  static async create(client, mouse) {
-    let mainFramePayload = await client.send('Page.getResourceTree');
-    return new FrameManager(client, mainFramePayload.frameTree, mouse);
-  }
-
-  /**
-   * @param {!Connection} client
    * @param {!Object} frameTree
    * @param {!Mouse} mouse
    */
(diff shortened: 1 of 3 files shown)


lib/FrameManager.js
```diff
@@ -81,13 +70,34 @@ class FrameManager extends EventEmitter {
...
    */
   _onFrameNavigated(framePayload) {
-    let frame = this._frames.get(framePayload.id);
-    if (!frame) {
-      // Simulate missed "frameAttached" for a main frame navigation to the new backend process.
-      console.assert(!framePayload.parentId, 'Main frame shouldn\'t have parent frame id.');
-      frame = this._mainFrame;
+    const isMainFrame = !framePayload.parentId;
+    let frame = isMainFrame ? this._mainFrame : this._frames.get(framePayload.id);
+    console.assert(isMainFrame || frame, 'We either navigate top level or have old version of the navigated frame');
```
JoelEinbinder, line 75: Can you explain what this assertion is checking?

pavelfeldman: We are getting _onFrameNavigated for top level frames and iframes.

If case we are getting notification about an iframe, we should have it in our map since we've been tracking frames creation and destruction since the beginning of the time. But for the top level frames, we are not getting the frame attached event, so it is fine for map to not have it.

It might be clearer if I put assertion above the ternary operator above.

aslushnikov:
For the record: The upstream has landed as 490209

aslushnikov:
Chromium could be rolled to 490379 to make this work

PR #162, opened 2017-07-28
Fix injectfile test on Windows

JoelEinbinder:
slashes were the wrong direction, so the test failed on windows.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -163,16 +163,16 @@ describe('Puppeteer', function() {
 
   describe('Page.injectFile', function() {
     it('should work', SX(async function() {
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -163,16 +163,16 @@ describe('Puppeteer', function() {
 
   describe('Page.injectFile', function() {
     it('should work', SX(async function() {
-      const helloPath = path.join(__dirname, './assets/injectedfile.js');
```
aslushnikov, line 166: path.join would be cleaner
```js
path.join(__dirname, 'assets', 'injectedfile.js')
```

JoelEinbinder: Done.

PR #171, opened 2017-07-31
Add delay option to input methods

### LINKED ISSUE #156: Add delay option to input methods
For `mouse.click`, delay should specify the time between `mouse.down` and `mouse.up`.

For `keyboard.press` delay should specify the time between `keyboard.down` and `keyboard.up`.

For `keyboard.type` delay should specify both the time between each key is pressed, and the time between the `down` and `up`.

JoelEinbinder:
Closes #156 

This lets you easily simulate a long press of a key or mouse press, or a user typing at a normal speed.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -54,7 +54,7 @@
     + [page.setUserAgent(userAgent)](#pagesetuseragentuseragent)
     + [page.setViewport(viewport)](#pagesetviewportviewport)
     + [page.title()](#pagetitle)
-    + [page.type(text)](#pagetypetext)
+    + [page.type(text, options)](#pagetypetext-options)
     + [page.uploadFile(selector, ...filePaths)](#pageuploadfileselector-filepaths)
     + [page.url()](#pageurl)
     + [page.viewport()](#pageviewport)
@@ -358,6 +358,7 @@ Shortcut for [page.mainFrame().addScriptTag(url)](#frameaddscripttagurl).
 - `options` <[Object]>
   - `button` <[string]> `left`, `right`, or `middle`, defaults to `left`.
   - `clickCount` <[number]> defaults to 1
+  - `delay` <[number]> Time to wait between `mousedown` and `mouseup` in milliseconds. Defaults to 0.
 - returns: <[Promise]> Promise which resolves when the element matching `selector` is successfully clicked. Promise gets rejected if there's no element matching `selector`.
 <!-- gen:stop -->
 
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -826,6 +830,7 @@ Dispatches a `keyup` event.
 - `options` <[Object]>
   - `button` <[string]> `left`, `right`, or `middle`, defaults to `left`.
   - `clickCount` <[number]> defaults to 1
+  - `delay` <[number]> Time to wait between `mousedown` and `mouseup` in milliseconds. Defaults to 0.
```
aslushnikov, line 838: Let's add an example which simulates a user typing in one field, and fills instantly another

test/test.js
```diff
@@ -1098,6 +1098,17 @@ describe('Puppeteer', function() {
           fail(modifiers[modifier] + ' should be false');
       }
     }));
+    it('should type with a delay', SX(async function(){
+      await page.navigate(PREFIX + '/input/textarea.html');
+      await page.focus('textarea');
+      await page.type('Hello World!', {delay: 50});
```
aslushnikov: this test doesn't test typing with a delay, doesn't it? (it would pass without the changes in `lib/`)

test/test.js
```diff
@@ -1098,6 +1098,17 @@ describe('Puppeteer', function() {
...
+      expect(await page.evaluate(() => document.querySelector('textarea').value)).toBe('Hello World!');
+    }));
+    it('should click with a delay', SX(async function(){
+      await page.navigate(PREFIX + '/input/button.html');
```
aslushnikov: looks like this would pass as well

JoelEinbinder: It would, it is just a sanity check. Added a comment and made the delay 1.

docs/api.md
```diff
@@ -358,6 +358,7 @@ Shortcut for [page.mainFrame().addScriptTag(url)](#frameaddscripttagurl).
 - `options` <[Object]>
   - `button` <[string]> `left`, `right`, or `middle`, defaults to `left`.
```
gajus, line 359: What does left, right, middle refer to?

Urigo: @gajus if I understand correctly, it's the mouse buttons - `left button click`, `right button click` and `middle button click`

gajus: I see. I guess it was not obvious since I haven't used a mouse in the last +5 years. :-)

PR #172, opened 2017-07-31
Specify repeat property on repeated keypresses

### LINKED ISSUE #157: Add repeat option to sendCharacter
See https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/repeat

`keyboard.sendCharacter` should be able to specify whether it was a repeated character

`keyboard.press` should get a helper option to send multiple characters

JoelEinbinder:
[Repeat](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/repeat) works a little bit different than I thought it did. When you hold down a key, it sends multiple `keydown` and `input` events, but no `keyup`. This patch automatically sets repeat to the right value.

Closes #157

### DIFF
--- a/lib/Input.js
+++ b/lib/Input.js
@@ -23,15 +23,18 @@ class Keyboard {
   constructor(client) {
     this._client = client;
     this._modifiers = 0;
+    this._pressedKeys = new Set();
   }
 
   /**
    * @param {string} key
-   * @param {{text: (string|undefined)}=} options
+   * @param {{text: (string|undefined)}} options
    * @return {!Promise}
    */
   async down(key, options) {
-    let { text } = options || {};
+    let {text} = options || {};
+    let autoRepeat = this._pressedKeys.has(key);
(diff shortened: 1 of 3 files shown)


lib/Input.js
```diff
@@ -23,23 +23,27 @@ class Keyboard {
...
     this._modifiers |= this._modifierBit(key);
     await this._client.send('Input.dispatchKeyEvent', {
       type: text ? 'keyDown' : 'rawKeyDown',
       modifiers: this._modifiers,
       windowsVirtualKeyCode: codeForKey(key),
       key: key,
       text: text,
-      unmodifiedText: text
+      unmodifiedText: text,
+      autoRepeat
```
aslushnikov, line 46: Let's add to the documentation that one the key is pressed, all the subsequent keyboard.down() of this key would generate autoRepeat events.

JoelEinbinder: Done.

PR #173, opened 2017-07-31
Slow mode

JoelEinbinder:
Throttles the protocol to run tests in slow mode.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -183,7 +183,8 @@ browser.newPage().then(async page => {
 - `options` <[Object]>  Set of configurable options to set on the browser. Can have the following fields:
 	- `headless` <[boolean]> Whether to run chromium in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). Defaults to `true`.
 	- `executablePath` <[string]> Path to a chromium executable to run instead of bundled chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
-	- `args` <[Array]<[string]>> Additional arguments to pass to the chromium instance. List of chromium flags can be found [here](http://peter.sh/experiments/chromium-command-line-switches/).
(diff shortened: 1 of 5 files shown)


paulirish:
Cool.  But... why? :)

Verbose logging isn't good enough?

aslushnikov:
@paulirish one way to debug automation script is to visually follow what it's doing. This comes handy to debug input tests, e.g. try `HEADLESS=false THROTTLE=200 npm run unit -- --filter=input`.

It also makes a cool automation demo :)

@JoelEinbinder `throttle` is an overloaded word; let's use `slowDown` throughout the patch?

JoelEinbinder:
@paulirish I was using it when writing tests with input so I can see what is happening on the page.  Then on the weekend writing my own tests, it was really painful to figure out what was going on without it.

paulirish:
@JoelEinbinder gotcha. sgtm. 

bikeshed: how about we call the option `slowmo`?

docs/api.md
```diff
@@ -183,7 +183,8 @@ browser.newPage().then(async page => {
 - `options` <[Object]>  Set of configurable options to set on the browser. Can have the following fields:
 	- `headless` <[boolean]> Whether to run chromium in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). Defaults to `true`.
 	- `executablePath` <[string]> Path to a chromium executable to run instead of bundled chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
-	- `args` <[Array]<[string]>> Additional arguments to pass to the chromium instance. List of chromium flags can be found [here](http://peter.sh/experiments/chromium-command-line-switches/).
+  - `slowMo` <[number]> Time in milliseconds to wait before responding to protocol messages. This slows down Puppeteer so you can see what is going on.
```
aslushnikov: let's not mention protocol here. Rather something along the lines:

```js
Slows down Puppeteer operations by the specified amount of milliseconds. Useful
for debugging automation scripts visually.
```

PR #174, opened 2017-07-31
Implement CPU profiler wrappers.

a1ph:


test/test.js
```diff
@@ -768,6 +768,40 @@ describe('Puppeteer', function() {
...
+      try {
+        await profiler.startCpuProfile();
+      } catch (e) {
+        error = e;
+      }
+      await profiler.stopCpuProfile();
+      expect(error).toBeTruthy();
+      expect(error.message).toContain('has already been started');
+    }));
+    it('should throw', SX(async function() {
```
JoelEinbinder, line 792: "should throw if not started"

lib/ProfilesManager.js
```diff
@@ -0,0 +1,59 @@
...
+  constructor(client) {
+    this._client = client;
+    this._profiling = false;
+  }
+
+  /**
+   * @param {number=} samplingFrequencyHz
+   * @return {!Promise}
+   */
+  async startCpuProfile(samplingFrequencyHz = 10000) {
```
JoelEinbinder, line 30: Are we going to have more options here? We have been using options objects preemptively.

lib/ProfilesManager.js
```diff
@@ -0,0 +1,59 @@
...
+    await this._client.send('Profiler.start', {});
+  }
+
+  /**
+   * @return {!Promise<!ProfilesManager.CpuProfile>}
+   */
+  async stopCpuProfile() {
+    if (!this._profiling)
+      throw Error(`Cannot stop the CPU profiler. It hasn't been started.`);
+    const profile = await this._client.send('Profiler.stop', {});
```
JoelEinbinder, line 44: `const {profile} = `

examples/cpu-profile.js
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+(async function main() {
+
+const Browser = require('../lib/Browser');
+const browser = new Browser();
+const page = await browser.newPage();
+const profiler = page.profiler();
+
+await profiler.startCpuProfile();
```
paulirish, line 24: warning: naming bikeshed. :)

IMO "Cpu profile" is a legacy term. Looking at it with fresh eyes I don't think it accurately reflects this data. The data is specific to one thread, rather than the whole CPU, but more importantly it's only data from the javascript sampling profiler. 

I'd advocate for `JsProfile` rather than `CpuProfile`.

lib/ProfilesManager.js
```diff
@@ -0,0 +1,59 @@
...
+    this._profiling = false;
+    return profile.profile;
+  }
+}
+
+/** @typedef {!{startTime: number, endTime: number, samples: !Array<number>, timeDeltas: !Array<number>, nodes: !Array<!ProfilesManager.CpuProfileNode>}} */
+ProfilesManager.CpuProfile;
```
paulirish, line 51: Just a reminder that we'll have a lot less flexibility to change this object shape after shipping this.

examples/cpu-profile.js
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+(async function main() {
+
+const Browser = require('../lib/Browser');
+const browser = new Browser();
+const page = await browser.newPage();
+const profiler = page.profiler();
```
paulirish, line 22: is the plan to add more perf stuff on `profiler()`? might make sense to call it `perf` or `performance`

If meant to replicate puppeteer API, this should probably be a property rather than a method. a la `page.keyboard` and `page.mouse`

test/test.js
```diff
@@ -768,6 +768,40 @@ describe('Puppeteer', function() {
     }));
   });
 
+  describe('Profiler.start', () => {
+    it('should work', SX(async function() {
```
paulirish, line 772: can we run some javascript in the page and confirm the profiler picked it up? right now these tests would pass even if samples and nodes are empty arrays.

aslushnikov:
For the record: this got closed because it looks like there're not enough demand to have it

PR #176, opened 2017-08-01
Implement Page.setIgnoreHTTPSErrors

### LINKED ISSUE #84: Allow to ignore certificate errors during navigation
There should be a capability to ignore certificate errors during navigation. See https://github.com/GoogleChrome/puppeteer/issues/66 for proposal and https://github.com/cyrus-and/chrome-remote-interface/issues/183 for original discussion.

For example, we can add `ignoreSSL` navigation option:

```js
page.navigate('https://example.com', { ignoreSSL: true })
```

aslushnikov:
This patch implements Page.setIgnoreHTTPSErrors method.

Fixes #84.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -568,7 +568,7 @@ The extra HTTP headers will be sent with every request the page initiates.
 > **NOTE** page.setExtraHTTPHeaders does not guarantee the order of headers in the outgoing requests.
 
 #### page.setIgnoreHTTPSErrors(enabled)
-- `enabled` <[boolean]> Weather HTTPS errors should be ignored on navigation. Defaults to `false`.
+- `enabled` <[boolean]> Whether HTTPS errors should be ignored on navigation. Defaults to `false`.
 - returns: <[Promise]> Promise which resolves when ignoring HTTPS errors is applied.
 
 When set to `false`, navigating to pages with HTTPS errors fails and chromium shows interstitial page.
--- a/docs/api.md
+++ b/docs/api.md
@@ -49,6 +49,7 @@
     + [page.screenshot([options])](#pagescreenshotoptions)
     + [page.setContent(html)](#pagesetcontenthtml)
     + [page.setExtraHTTPHeaders(headers)](#pagesetextrahttpheadersheaders)
+    + [page.setIgnoreHTTPSErrors(enabled)](#pagesetignorehttpserrorsenabled)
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -566,6 +567,13 @@ The extra HTTP headers will be sent with every request the page initiates.
 
 > **NOTE** page.setExtraHTTPHeaders does not guarantee the order of headers in the outgoing requests.
 
+#### page.setIgnoreHTTPSErrors(enabled)
+- `enabled` <[boolean]> Weather HTTPS errors should be ignored on navigation. Defaults to `false`.
```
JoelEinbinder: Whether

aslushnikov:
The code is ready, but the test which disables ignoring for some reason fails. It looks like once certificate is "trusted" once, it becomes trusted for the lifetime of a browser.

@allada helps to investigate

docs/api.md

lib/NavigatorWatcher.js
```diff
@@ -19,53 +19,60 @@ const helper = require('./helper');
...
-    let loadEventFired = new Promise(fulfill => {
-      this._eventListeners.push(helper.addEventListener(this._client, 'Page.loadEventFired', fulfill));
-    }).then(() => null);
+    this._eventListeners = [];
 
     let watchdog = new Promise(fulfill => this._maximumTimer = setTimeout(fulfill, this._timeout))
         .then(() => 'Navigation Timeout Exceeded: ' + this._timeout + 'ms exceeded');
+    let navigationPromises = [watchdog];
+
+    if (!this._ignoreHTTPSErrors) {
```
JoelEinbinder, line 48: Is this because we still get a certificateError event when telling the protocol to ignore certificate errors?

JoelEinbinder:
We should probably fix the certificate trust thing, but for practical use I imagine I'd turn this on once at startup and forget about it. An option in `Browser` would work for me.

aslushnikov:
@JoelEinbinder totally agree

lib/NavigatorWatcher.js
aslushnikov: We implement ignoring via intercepting certificate errors and handling them with the "continue" command. So yeah, we still get the certificate errors

aslushnikov:
Closing this. I'll send another PR with browser option

Issue #179, opened 2017-08-01
Record a trace


JoelEinbinder:
You should be able to record a trace from puppeteer and view it in timeline.

PR #181, opened 2017-08-02
Tracing

JoelEinbinder:
#179

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -847,7 +847,6 @@ await page.tracing.stop('trace.json');
 #### tracing.start([options])
 - `options` <[Object]>
   - `screenshots` <[boolean]> captures screenshots in the trace.
-  - `clickCount` <[number]> defaults to 1
 - returns: <[Promise]>
 
 #### tracing.stop(path)
--- a/docs/api.md
+++ b/docs/api.md
@@ -54,6 +54,7 @@
     + [page.setUserAgent(userAgent)](#pagesetuseragentuseragent)
     + [page.setViewport(viewport)](#pagesetviewportviewport)
     + [page.title()](#pagetitle)
+    + [page.tracing](#pagetracing)
     + [page.type(text, options)](#pagetypetext-options)
     + [page.uploadFile(selector, ...filePaths)](#pageuploadfileselector-filepaths)
     + [page.url()](#pageurl)
@@ -71,6 +72,9 @@
     + [mouse.down([options])](#mousedownoptions)
     + [mouse.move(x, y)](#mousemovex-y)
     + [mouse.up([options])](#mouseupoptions)
+  * [class: Tracing](#class-tracing)
+    + [tracing.start([options])](#tracingstartoptions)
+    + [tracing.stop(path)](#tracingstoppath)
   * [class: Dialog](#class-dialog)
     + [dialog.accept([promptText])](#dialogacceptprompttext)
     + [dialog.dismiss()](#dialogdismiss)
@@ -638,6 +642,9 @@ In case of multiple pages in one browser, each page can have its own viewport si
 
 Shortcut for [page.mainFrame().title()](#frametitle).
 
+#### page.tracing
+- returns: <[Tracing]>
+
 #### page.type(text, options)
 - `text` <[string]> A text to type into a focused element.
 - `options` <[Object]>
@@ -826,6 +833,26 @@ Dispatches a `mousemove` event.
 - returns: <[Promise]>
 
 Dispatches a `mouseup` event.
+
+### class: Tracing
+
+You can use [`tracing.start`](#tracingstartoptions) and [`tracing.stop`](#tracingstoppath) to create a trace file which can be opened in Chrome DevTools or [timeline viewer](https://chromedevtools.github.io/timeline-viewer/).
+
+```js
+await page.tracing.start();
+await page.navigate('https://www.google.com');
+await page.tracing.stop('trace.json');
+```
+
+#### tracing.start([options])
+- `options` <[Object]>
+  - `screenshots` <[boolean]> captures screenshots in the trace.
+  - `clickCount` <[number]> defaults to 1
+- returns: <[Promise]>
+
+#### tracing.stop(path)
+- `path` <[string]> A path to write the trace file to.
+- returns: <[Promise]>
 
 ### class: Dialog
 
@@ -1194,3 +1221,4 @@ If changed, the request url will be modified in a way that's not observable by p
 [Mouse]: https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-mouse "Mouse"
 [Map]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map "Map"
 [selector]: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors "selector"
+[Tracing]: https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-tracing "Tracing"
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -23,6 +23,7 @@ let Dialog = require('./Dialog');
 let EmulationManager = require('./EmulationManager');
 let FrameManager = require('./FrameManager');
 let {Keyboard, Mouse} = require('./Input');
+let Tracing = require('./Tracing');
 let helper = require('./helper');
 
 class Page extends EventEmitter {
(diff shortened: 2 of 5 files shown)


docs/api.md
```diff
@@ -827,6 +834,26 @@ Dispatches a `mousemove` event.
...
+
+#### tracing.start([options])
+- `options` <[Object]>
+  - `screenshots` <[boolean]> captures screenshots in the trace.
+  - `clickCount` <[number]> defaults to 1
+- returns: <[Promise]>
+
+#### tracing.stop(path)
+- `path` <[string]> A path to write the trace file to.
+- returns: <[Promise]>
```
aslushnikov, line 854: - should return <[Promise<[string]>>
- path is optional

lib/Tracing.js
```diff
@@ -0,0 +1,83 @@
...
+
+  /**
+   * @param {Object=} options
+   */
+  async start(options = {}) {
+    if (this._recording)
+      throw new Error('Cannot start recording trace while already recording trace.');
+
+    const categoriesArray = [
+      '-*', 'devtools.timeline', 'v8.execute', 'disabled-by-default-devtools.timeline',
```
aslushnikov: Where do these categories come from?

docs/api.md
```diff
@@ -827,6 +834,26 @@ Dispatches a `mousemove` event.
...
+```js
+await page.tracing.start();
+await page.navigate('https://www.google.com');
+await page.tracing.stop('trace.json');
+```
+
+#### tracing.start([options])
+- `options` <[Object]>
+  - `screenshots` <[boolean]> captures screenshots in the trace.
+  - `clickCount` <[number]> defaults to 1
```
aslushnikov: what is click count??

lib/Tracing.js
```diff
@@ -0,0 +1,83 @@
...
+      transferMode: 'ReturnAsStream',
+      categories: categoriesArray.join(',')
+    });
+  }
+
+  /**
+   * @param {string} path
+   */
+  async stop(path) {
+    this._recording = false;
```
aslushnikov: let's have it after the trace is fetched from backend

lib/Tracing.js
```diff
@@ -0,0 +1,83 @@
...
+    let fulfill;
+    let contentPromise = new Promise(x => fulfill = x);
+    this._client.once('Tracing.tracingComplete', event => {
+      this._readStream(event.stream, path).then(fulfill);
+    });
+    await this._client.send('Tracing.end', {
```
aslushnikov: let's remove empty multiline arguments object

docs/api.md
JoelEinbinder: That keeps getting added. I've deleted it 3 times..

docs/api.md
JoelEinbinder: Talking with @paulirish , large strings might kill node. So we force it to be written to disk.

lib/Tracing.js
JoelEinbinder: Default devtools trace.

JoelEinbinder:
I enabled JS profiling, and fixed the other concerns.

test/test.js
```diff
@@ -1481,6 +1481,19 @@ describe('Page', function() {
...
+    let outputFile = path.join(__dirname, 'assets', 'trace.json');
+    afterEach(function() {
+      fs.unlinkSync(outputFile);
+    });
+    it('should output a trace', SX(async function() {
+      await page.tracing.start({screenshots: true});
+      await page.navigate(PREFIX + '/grid.html');
+      await page.tracing.stop(outputFile);
+      expect(fs.existsSync(outputFile)).toBe(true);
+    }));
```
aslushnikov, line 1495: let's add a test which starts tracing in parallel in two pages and makes sure we fail

lib/Tracing.js
```diff
@@ -0,0 +1,83 @@
...
+  constructor(client) {
+    this._client = client;
+    this._recording = false;
+  }
+
+  /**
+   * @param {Object=} options
+   */
+  async start(options = {}) {
+    if (this._recording)
```
aslushnikov:
```js
console.assert(!this._recording, 'Cannot start recording trace while already recording trace.')
```

lib/Tracing.js
```diff
@@ -0,0 +1,83 @@
...
+      let response = await this._client.send('IO.read', {handle});
+      eof = response.eof;
+      if (path)
+        fs.writeSync(file, response.data);
+    }
+    fs.closeSync(file);
+    await this._client.send('IO.close', {handle});
+  }
+}
+module.exports = Tracing;
```
aslushnikov: this should also have the "tracePublicAPI" call to helper

test/test.js
JoelEinbinder: Done.

lib/Tracing.js
JoelEinbinder: Done.

lib/Page.js
```diff
@@ -23,6 +23,7 @@ let Dialog = require('./Dialog');
 let EmulationManager = require('./EmulationManager');
 let FrameManager = require('./FrameManager');
 let {Keyboard, Mouse} = require('./Input');
+let Tracing = require('./Tracing');
```
a1ph, line 26: const?

lib/Tracing.js
JoelEinbinder: Done.

test/test.js
```diff
@@ -1481,6 +1481,31 @@ describe('Page', function() {
...
+      await page.tracing.start();
+      let newPage = await browser.newPage();
+      let error = null;
+      try {
+        await newPage.tracing.start();
+      } catch (e) {
+        error = e;
+      }
+      expect(error).toBeTruthy();
+      await page.tracing.stop(outputFile);
```
aslushnikov: Let's close the page so that it's not dangling for the rest of the tests
```js
await newPage.close();
```

aslushnikov (approved):

lib/Tracing.js
```diff
@@ -0,0 +1,83 @@
...
+      categories: categoriesArray.join(',')
+    });
+  }
+
+  /**
+   * @param {string} path
+   */
+  async stop(path) {
+    let fulfill;
+    let contentPromise = new Promise(x => fulfill = x);
```
a1ph, line 56: Why all this promise juggling? Sequential code would look more clear, e.g.:
```js
this._client.send('Tracing.end');
const {stream} = await this._client.once('Tracing.tracingComplete');
await this._readStream(stream, path);
this._recording = false;
```

lib/Tracing.js
```diff
@@ -0,0 +1,83 @@
...
+
+    this._recording = true;
+    await this._client.send('Tracing.start', {
+      transferMode: 'ReturnAsStream',
+      categories: categoriesArray.join(',')
+    });
+  }
+
+  /**
+   * @param {string} path
```
a1ph, line 52: @return {!Promise}

lib/Tracing.js
```diff
@@ -0,0 +1,83 @@
...
+  }
+
+  /**
+   * @param {string} handle
+   * @param {string=} path
+   * @return {!Promise}
+   */
+  async _readStream(handle, path) {
+    let eof = false;
+    let file = fs.openSync(path, 'w');
```
a1ph, line 72: const
Also what if path is undefined?

lib/Tracing.js
```diff
@@ -0,0 +1,83 @@
...
+    while (!eof) {
+      let response = await this._client.send('IO.read', {handle});
```
a1ph, line 74: const

test/test.js
JoelEinbinder: Done.

test/test.js
```diff
@@ -1481,6 +1481,19 @@ describe('Page', function() {
       await Promise.all(pages.map(page => page.close()));
     }));
   });
+
+  describe('Tracing', function() {
+    let outputFile = path.join(__dirname, 'assets', 'trace.json');
```
a1ph, line 1486: const

lib/Page.js
aslushnikov: We need to convert all our `require`'s into consts, let's do it separately

lib/Tracing.js
aslushnikov: Let's address this in a follow-up (I've already merged the PR before I saw your comments)

lib/Tracing.js
aslushnikov: `once` doesn't return a promise

lib/Tracing.js
aslushnikov: path is not optional in `trace.stop()` method

docs/api.md
```diff
@@ -827,6 +834,25 @@ Dispatches a `mousemove` event.
...
+await page.tracing.stop('trace.json');
+```
+
+#### tracing.start([options])
+- `options` <[Object]>
+  - `screenshots` <[boolean]> captures screenshots in the trace.
+- returns: <[Promise]>
+
+#### tracing.stop(path)
+- `path` <[string]> A path to write the trace file to.
```
paulirish, line 853: not options object?  
screenshot takes path inside the object, so this feels a tad inconsistent

paulirish: actually... it makes a bit more sense to specify this path on start() instead.  yah?

Issue #121, opened 2017-07-25
Merge InterceptedRequest with Request


JoelEinbinder:
the `request.continue()` should accept options object with overrides for the request fields.

aslushnikov:
Are the request object, dispatched via the `request` event, and the one passed over to the request interceptor, the same?

- If they are the same, at which point of time should calls `continue` or `abort` stop throwing and start doing meaningful things?
- If they are not the same, then there's no good reason to have a dummy `continue` and `abort` methods for every request.

Can we have `InterceptedRequest` class which has a `.request` field, pointing to the original request, and methods `continue` and `abort`? 
```js
page.setRequestInterceptor(interception => {
  interception.request.url; // interception.request holds the reference to the original request
  interception.continue({method: 'GET'}); // continues the request, overriding `method`
  interception.abort(); // aborts the request
});
```
@pavelfeldman wdyt?

aslushnikov:
As discussed offline with @pavelfeldman:

```js
// enable request interception
await page.setInterceptRequests(true);

page.on('request', request => {
  // request has `continue` method which throws in case of disabled request interception.
  request.continue({method: 'GET'});
});
```

aslushnikov:
Implementation notes:
- there's no `requestId` in the interception, so requests should be matched by payload
- there's no strict order between request interception and `requestWillBeSent` events, so the implementation would need to race and match them

This doesn't sound that great, so we're considering alternatives.

JoelEinbinder:
Can you clarify the different events and what they mean? What are the states a request could be in?
 I asked before but I'm still confused on them.

PR #184, opened 2017-08-02
[WIP] merge Request and InterceptedRequest classes

aslushnikov:
References #121

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -23,11 +23,14 @@ class NetworkManager extends EventEmitter {
   constructor(client) {
     super();
     this._client = client;
-    this._requestInterceptor = null;
     /** @type {!Map<string, !Request>} */
     this._idToRequest = new Map();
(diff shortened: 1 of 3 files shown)


lib/NetworkManager.js
```diff
@@ -160,6 +205,35 @@ class Request {
...
+    });
+  }
+
+  abort() {
+    console.assert(this._interceptionId, 'Request Interception is not enabled!');
+    console.assert(!this._interceptionHandled, 'Request is already handled!');
+    this._interceptionHandled = true;
+    this._client.send('Network.continueInterceptedRequest', {
+      interceptionId: this._interceptionId,
+      errorReason: 'Aborted'
```
dgozman, line 234: Note this aborts the navigation if request is a navigation one. It's safer to use Failed.

aslushnikov:
For the record: this doesn't work nice since we can't match requests with interceptions. We need protocol to help us here.

PR #191, opened 2017-08-02
Ignore package-lock.json

JoelEinbinder:
pacakge-lock.json seems to be buggy now, with different OS giving different results. We have been having trouble keeping it up to date with yarn.lock. It doesn't give us a big win, because it is ignored if you install the package from npm.


aslushnikov (approved):

aslushnikov:
The npm bug that causes our frustration: https://github.com/npm/npm/issues/17749

PR #196, opened 2017-08-03
Add more tests to cover Page.pdf() method

### LINKED ISSUE #166: Try to introduce golden pdf tests
The `page.pdf()` method is poorly tested.
To test it properly, we'd need to actually generate pdf's and to compare them against golden results, like we do with screenshots.

aslushnikov:
This patch adds sanity tests to make sure Page.pdf() produces
sane results.

Fixes #166.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -103,6 +103,19 @@ describe('Page', function() {
     server.reset();
     httpsServer.reset();
     GoldenUtils.addMatchers(jasmine, GOLDEN_DIR, OUTPUT_DIR);
+    jasmine.addMatchers({
+      toEPSEqual: function(util, customEqualityTesters) {
+        return {
+          compare: function(actual, expected, eps) {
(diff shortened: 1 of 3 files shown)


aslushnikov:
@JoelEinbinder doesn't look scary, and doesn't add a lot more to `node_modules/`

test/test.js
```diff
@@ -103,6 +103,19 @@ describe('Page', function() {
     server.reset();
     httpsServer.reset();
     GoldenUtils.addMatchers(jasmine, GOLDEN_DIR, OUTPUT_DIR);
+    jasmine.addMatchers({
+      toEPSEqual: function(util, customEqualityTesters) {
```
JoelEinbinder: let's use toBeCloseTo

JoelEinbinder (approved):

Issue #111, opened 2017-07-25
Element handles


JoelEinbinder:
There has been a lot of talk about element handles/references. Some points to consider:

- Should they reference actual nodeIds, or reference a selector/finder function
- Do we need them?
- What happens if the element leaves DOM? If it gets garbage collected?
- Do we still have `page.click(selector)`?

How they might look:
```js
let button = page.$('button');
await button.click();
```

```js
let button = await page.waitForSelector('button');
await button.click();
```

```js
page.$('button')
    .focus()
    .click()
    .run(b => b.textContent = 'hello');
```

```js
let button = page.$('button')
await page.waitFor('button')
await button.click();
```

```js
let button = page.$('button')
await page.waitFor('button')
await button.click();
```

aslushnikov:
### Suggestion
Let's design element handles similarly to the element handles in devtools console. This would be a
very natural and familiar behavior.

<img width="400" alt="screen shot 2017-07-29 at 11 26 52 pm" src="https://user-images.githubusercontent.com/746130/28751179-ca5585b0-74b5-11e7-9864-c5116d85da8b.png">


Answering questions from this standpoint: 

> - Should they reference actual nodeIds, or reference a selector/finder function

They should reference actual nodes through RemoteObjects.

> Do we need them?

Yes. There are good reasons to have element handles:
- Everybody wants element handles so that they don't have to re-type selectors multiple times
- Handles would allow us to interact with nodes which are only accessible through page's javascript
- Handles will give us a good isolated namespace. Such methods as `Element.screenshot()` would be easily added.

> What happens if the element leaves DOM? If it gets garbage collected?

if element gets detached from DOM, nothing happens. The element is never garbage collected since it is retained through remote object.

> Do we still have page.click(selector)?

Yes, they come handy for small scripts and it would make us closer to other API's.

### API Draft

Let's introduce the `Element` class which is a representation of Page's element.

`Element` instance could be created by one of these methods:
1. Querying the page for the selector

```js
let elem = await page.$('div');
```

2. Returning element from `page.evaluate`:
```js
let elem = await page.evaluate(() => document.body);
```

3. Awaiting element to appear in page:
```js
let elem = await page.waitFor('div');
```

`Element` instance could be destroyed via the `release` method (though I imagine noone would care):
```js
elem.release();
```

`Element` instances would be also destroyed on frame navigation.

API Demo:
```js
let elem = await page.waitFor('input[type=search]');
await elem.focus();
[...]

PR #197, opened 2017-08-03
Kill page.$ and page.$$ functions

aslushnikov:
The current implementation of page.$ and page.$$ conflicts with
the elemnts handle implementation.

This patch removes functions in favor of future implementation.

References #111.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -311,7 +311,7 @@ describe('Page', function() {
 
     it('should work when node is added through innerHTML', SX(async function() {
       await page.navigate(EMPTY_PAGE);
-      let watchdog = page.waitForSelector('h3 div').then(() => added = true);
+      let watchdog = page.waitForSelector('h3 div');
       await page.evaluate(addElement, 'span');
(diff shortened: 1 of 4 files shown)


test/test.js
```diff
@@ -311,13 +311,10 @@ describe('Page', function() {
...
       await page.navigate(EMPTY_PAGE);
-      let frame = page.mainFrame();
-      let added = false;
-      frame.waitForSelector('h3 div').then(() => added = true);
-      expect(added).toBe(false);
-      await frame.evaluate(addElement, 'span');
-      await page.$('span', span => span.innerHTML = '<h3><div></div></h3>');
-      expect(added).toBe(true);
+      let watchdog = page.waitForSelector('h3 div').then(() => added = true);
+      await page.evaluate(addElement, 'span');
```
JoelEinbinder: This doesn't test as much anymore. It will pass if the watchdog resolves immediately.  `added` is also never used.

aslushnikov: Since `waitForSelector` returns a promise, this test is functionally the same.
`added` is a good catch though!

JoelEinbinder (approved):

PR #203, opened 2017-08-04
Fix tests in non-headless mode

aslushnikov:
The chrome-devtools://devtools/bundled/inspector.html is a perfectly
valid url in headful chromium, so we should pick another one for test
to work (and for the navigation inside the test to fail).


JoelEinbinder (approved):
Is there something special about the `chrome-devtools://`? It makes me suspicious of this code.

aslushnikov:
Nah, nothing special for this case

PR #204, opened 2017-08-04
RemoteElement

### LINKED ISSUE #111: Element handles

JoelEinbinder:
What do you think?
closes #111

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -25,16 +25,14 @@
     + [event: 'requestfailed'](#event-requestfailed)
     + [event: 'requestfinished'](#event-requestfinished)
     + [event: 'response'](#event-response)
+    + [page.$(selector)](#pageselector)
     + [page.addScriptTag(url)](#pageaddscripttagurl)
-    + [page.click(selector[, options])](#pageclickselector-options)
     + [page.close()](#pageclose)
     + [page.evaluate(pageFunction, ...args)](#pageevaluatepagefunction-args)
     + [page.evaluateOnNewDocument(pageFunction, ...args)](#pageevaluateonnewdocumentpagefunction-args)
-    + [page.focus(selector)](#pagefocusselector)
     + [page.frames()](#pageframes)
     + [page.goBack(options)](#pagegobackoptions)
     + [page.goForward(options)](#pagegoforwardoptions)
-    + [page.hover(selector)](#pagehoverselector)
     + [page.injectFile(filePath)](#pageinjectfilefilepath)
     + [page.keyboard](#pagekeyboard)
     + [page.mainFrame()](#pagemainframe)
@@ -54,13 +52,20 @@
     + [page.title()](#pagetitle)
     + [page.tracing](#pagetracing)
     + [page.type(text, options)](#pagetypetext-options)
-    + [page.uploadFile(selector, ...filePaths)](#pageuploadfileselector-filepaths)
     + [page.url()](#pageurl)
     + [page.viewport()](#pageviewport)
     + [page.waitFor(selectorOrFunctionOrTimeout[, options])](#pagewaitforselectororfunctionortimeout-options)
     + [page.waitForFunction(pageFunction[, options, ...args])](#pagewaitforfunctionpagefunction-options-args)
     + [page.waitForNavigation(options)](#pagewaitfornavigationoptions)
     + [page.waitForSelector(selector[, options])](#pagewaitforselectorselector-options)
+  * [class: RemoteElement](#class-remoteelement)
+    + [remoteElement.center()](#remoteelementcenter)
+    + [remoteElement.click([options])](#remoteelementclickoptions)
+    + [remoteElement.eval(pageFunction, ...args)](#remoteelementevalpagefunction-args)
+    + [remoteElement.focus()](#remoteelementfocus)
+    + [remoteElement.hover()](#remoteelementhover)
+    + [remoteElement.release()](#remoteelementrelease)
+    + [remoteElement.uploadFile(...filePaths)](#remoteelementuploadfilefilepaths)
   * [class: Keyboard](#class-keyboard)
     + [keyboard.down(key[, options])](#keyboarddownkey-options)
     + [keyboard.sendCharacter(char)](#keyboardsendcharacterchar)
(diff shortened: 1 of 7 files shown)


docs/api.md
```diff
@@ -720,6 +696,40 @@ browser.newPage().then(async page => {
 ```
 Shortcut for [page.mainFrame().waitForSelector(selector[, options])](#framewaitforselectorselector-options).
 
+### class: RemoteElement
```
aslushnikov, line 698: Just `Element` (or we should have `RemotePage`, `RemoteBrowser`, `RemoteFrame`...)

aslushnikov: The documentation doesn't elaborate on some important topics (e.g. lifetime of the element, what if element is detached / not found), but this all could be added when we reach consensus

lib/FrameManager.js
```diff
@@ -326,58 +321,22 @@ class Frame {
...
-   * @return {!Promise}
+   * @return {!Object}
    */
-  async hover(selector) {
-    let {x, y} = await this._centerOfElement(selector);
-    await this._mouse.move(x, y);
-  }
+  async _resolveElement(selector) {
+    let remoteObject = await this._evaluateToRemoteObject(selector => document.querySelector(selector), selector);
+    if (!remoteObject.subtype === 'node')
```
aslushnikov: !==

lib/RemoteElement.js
```diff
@@ -0,0 +1,103 @@
...
+  constructor(client, remoteObject, mouse) {
+    this._client = client;
+    this._remoteObject = remoteObject;
+    this._mouse = mouse;
+  }
+
+  /**
+   * @return {!Promise<{x: number, y: number}>}
+   */
+  center() {
```
aslushnikov: this is not center of the element per se, it's the center of the element rect intersected with viewport. Let's make it private

lib/RemoteElement.js
```diff
@@ -0,0 +1,103 @@
...
+      objectId: (await this._remoteObject).objectId,
+      functionDeclaration: `function() {return (${pageFunction})(${args.map(x => JSON.stringify(x)).concat('this').join(',')}); }`,
+      returnByValue: false
+    });
+    if (exceptionDetails)
+      throw new Error('Evaluation failed: ' + helper.getExceptionMessage(exceptionDetails));
+    return await helper.serializeRemoteObject(this._client, remoteObject);
+  }
+
+  async hover() {
```
aslushnikov, line 73: jsdoc

lib/RemoteElement.js
```diff
@@ -0,0 +1,103 @@
...
+    let {x, y} = await this.center();
+    await this._mouse.move(x, y);
+  }
+
+  async focus() {
```
aslushnikov: jsdoc

lib/RemoteElement.js
```diff
@@ -0,0 +1,103 @@
...
+const path = require('path');
+const helper = require('./helper');
+
+class RemoteElement {
+  /**
+   * @param {!Connection} client
+   * @param {!Promise<?Object>} remoteObject
+   * @param {!Mouse} mouse
+   */
+  constructor(client, remoteObject, mouse) {
```
aslushnikov: - let's call it remoteObjectPromise
- will you pass in keyboard to support type as well?

aslushnikov (commented):
This looks like!
A few important questions:
- How does `page.waitFor` integrate with this?
- How do you plan to get element from js code?

docs/api.md
JoelEinbinder: The reason I went with `RemoteElement` was to avoid conflict with the html type `Element`. It would extend from `RemoteObject`. Nobody sees the word `RemoteElement` except in the docs.

Not done with the docs yet.

JoelEinbinder:
In the future, `helper.serializeRemoteObject` will generate and return a `RemoteElement` if the remoteObject was an element. You will be able to make a `RemoteElement` using `page.evaluate`. `waitForFunction` will return the truthy value that the `pageFunction` returned, which might be a `RemoteElement`. `waitForSelector` is just a special case of `waitForFunction` and will work the same way.

lib/RemoteElement.js
JoelEinbinder: I don't like having jsdoc just for async functions. Every type checker out there knows this returns a promise.

lib/RemoteElement.js
JoelEinbinder: yes, I will pass in keyboard for typing.

lib/FrameManager.js
JoelEinbinder: that is embarrassing

lib/RemoteElement.js
aslushnikov: Sure, let's remove them from everywhere then

docs/api.md
aslushnikov: How about `ElementHandle`?

lib/RemoteElement.js
```diff
@@ -0,0 +1,103 @@
...
+    this._client = client;
+    this._remoteObject = remoteObject;
```
aslushnikov: you want to catch here, so that the following doesn't throw UnhandledPromiseRejection:
```js
page.$('non-existing-element');
```

lib/FrameManager.js
aslushnikov: you also need to release object in this case

lib/RemoteElement.js
JoelEinbinder: I think that would be working as intended? If you try to create a `RemoteElement` and then never use it, there is no place to surface an error involved in it's creation. We could catch and throw our own warning, but I am fine with the default behavior. It is a point in favor of awaitable element handles though.

aslushnikov:
Following the conclusion on https://github.com/GoogleChrome/puppeteer/pull/212, let's have 'orthodox' element handles which don't try to give any sugar.

That said:
- `page.$` should be awaited
- element handles are backed by remote objects
- page shortcuts is a sugar atop of element handles

This is almost this PR, to the extend of `page.$` which should be awaitable.

aslushnikov:
Closing this in favor of `orthodox` element handles in some distant future.

PR #210, opened 2017-08-04
[doc] Remove mention of broken javascript dialogs from api.md

aslushnikov:


docs/api.md
```diff
@@ -822,6 +822,8 @@ await page.tracing.stop();
   - `screenshots` <[boolean]> captures screenshots in the trace.
 - returns: <[Promise]>
 
+Only one tracing could be started simultaneously per browser.
```
JoelEinbinder: Only one trace can be active at a time per browser.

JoelEinbinder (approved):

PR #212, opened 2017-08-04
Speculative: implement chainable API (aka 'nightmare')

aslushnikov:
This is a speculative implementation of chaining. It turns out to be simple: the implementation is generic and is ~20 lines of code.

With this patch, the 'scape google results' script feels light (and it actually works):

```js
(async () => {

const browser = new Browser({headless: false});

// 1. Navigate to google.com
let page = await browser.newPage()
                        .setViewport({width: 800, height: 800})
                        .navigate('https://google.com');

// 2. Type 'godfather' and trigger search.
await page.chain()
          .waitFor('input[title=Search]', {visible: true})
          .type('godfather')
          .click('input[type=submit]');

// 3. Navigate to each result one-by-one and
//    take full page screenshot of each.
for (let i = 0; i < 4; ++i) {
  // Wait for search result link to appear and click it
  const selector = `div.g:nth-child(${i+1}) h3 a`;
  await page.chain()
            .waitFor(selector)
            .click(selector)
            .waitForNavigation({waitUntil: 'networkidle'})
            .screenshot({
              fullPage: true,
              path: `screenshot-${i + 1}.png`
            })
            .goBack();
}

browser.close();

})();
```

For comparison, the current version of the script is:
```js
(async () => {

// 1. Navigate to google.com
await page.navigate('https://google.com');
await page.setViewport({width: 800, height: 800});
const searchField = 'input[title=Search]';
await page.waitFor(searchField, {visible: true});

// 2. Type 'godfather' and trigger search.
await page.focus(searchField);
await page.type('godfather');
await page.click('input[type=submit]');

// 3. Navigate to each result one-by-one and
//    take full page screenshot of each.
for (let i = 0; i < 5; ++i) {
  // Wait for search result link and click it
  const searchResult = `div.g:nth-child(${i+1}) h3 a`;
  await page.waitFor(searchResult);
  await page.click(searchResult);
  // Wait for page to navigate to the search result
  await page.waitForNavigation({waitUntil: 'networkidle'});
  await page.screenshot({
    fullPage: true,
    path: `screenshot-${i + 1}.png`
  });
  // Go back to search results page.
  await page.goBack();
}

browser.close();

})();

```

### DIFF
--- a/lib/Input.js
+++ b/lib/Input.js
@@ -138,11 +138,18 @@ class Mouse {
 
   async _doubleRaf() {
     // This is a hack for now, to make clicking less race-prone. See crbug.com/747647
-    await this._client.send('Runtime.evaluate', {
-      expression: 'new Promise(f => requestAnimationFrame(() => requestAnimationFrame(f)))',
-      awaitPromise: true,
-      returnByValue: true
-    });
+    try { 
+      await Promise.race([
+        this._client.send('Runtime.evaluate', {
+          expression: 'new Promise(f => requestAnimationFrame(() => requestAnimationFrame(f)))',
+          awaitPromise: true,
+          returnByValue: true
+        }),
+        new Promise(x => setTimeout(x, 100))
+      ]);
+    } catch (e) {
+      // Ignore
+    }
   }
 
   /**
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -88,13 +88,17 @@ class Browser {
   /**
    * @return {!Promise<!Page>}
    */
-  async newPage() {
+  async _newPage() {
     await this._ensureChromeIsRunning();
     if (!this._chromeProcess || this._terminated)
       throw new Error('ERROR: this chrome instance is not alive any more!');
     let client = await Connection.create(this._remoteDebuggingPort, this._connectionDelay);
     let page = await Page.create(client, this._ignoreHTTPSErrors, this._screenshotTaskQueue);
     return page;
+  }
+
+  newPage() {
+    return helper.chain(Page, this._newPage());
   }
 
   /**
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -86,6 +86,10 @@ class Page extends EventEmitter {
     client.on('Security.certificateError', event => this._onCertificateError(event));
   }
 
+  chain() {
+    return helper.chain(Page, Promise.resolve(this));
+  }
+
   /**
    * @return {!Frame}
    */
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -224,6 +224,43 @@ class Helper {
   static isNumber(obj) {
     return typeof obj === 'number' || obj instanceof Number;
   }
+
+  static chain(classType, instancePromise) {
+    return new Chain(classType, instancePromise, Promise.resolve());
+  }
 }
+
+class Chain {
+  constructor(classType, instancePromise, actionsChain) {
+    this._instancePromise = instancePromise;
+    this._actionsChain = actionsChain;
+    return new Proxy(this, {
+      get: function(target, property, receiver) {
+        if (property === 'then' || property === 'result' || property === '_instancePromise' || property === '_actionsChain')
+          return Reflect.get(target, property, receiver);
+        let classMethod = Reflect.get(classType.prototype, property);
+        console.assert(classMethod, `FAILED: Class "${classType.prototype.constructor.name}" does not have method "${property}"`);
+        console.assert(typeof classMethod === 'function', `FAILED: ${classType.prototype.constructor.name}.{property} is not a function!`);
+        return function(...args) {
+          let next = target._actionsChain.then(async () => {
+            let instance = await target._instancePromise;
+            return instance[property].call(instance, ...args);
+          });
+          return new Chain(classType, instancePromise, next);
+        }
+      }
+    });
+  }
+
+  async result() {
+    return this._actionsChain;
+  }
+
+  async then(resolve, reject) {
+    await this._actionsChain;
+    return this._instancePromise.then(resolve, reject);
+  }
+}
+
 
 module.exports = Helper;


aslushnikov:
@paulirish @JoelEinbinder

ebidel:
The new example is looking really nice. On first grok, I find it a bit confusing why `browser` doesn't need a `.chain()` to chain the calls but `page` does. Is there a way to make those consistent?

aslushnikov:
@ebidel the difference is the target which methods you chain. For example, in case of `browser.newPage()`, you actually chain the methods of a future page (such as `setViewport` and `navigate`):

This is possible because `browser.newPage` returns <[Chain](#)<[Page](#)>>:
```js
// 1. Navigate to google.com
let page = await browser.newPage() // <-- start a Chain<Page> for the _future_ page
                        .setViewport({width: 800, height: 800}) // call setViewport on a Page
                        .navigate('https://google.com'); // navigate the page
```

In case when you have an instance in hands and want to chain some of its calls, you can start chain by a generic `.chain()` method:
```js
// 2. Type 'godfather' and trigger search.
await page.chain() // <-- start a chain for the _current_ page
          .waitFor('input[title=Search]', {visible: true})
          .type('godfather')
          .click('input[type=submit]');
```

Does it makes things any clearer? 

Hopefully, having the documentation to state return types and a lot of examples will make things more understandable.

JoelEinbinder:
Without assertions, you could implement this in a project-agnostic way. I think it shouldn't be in the core project.

Garbee:
The chainable API looks to help decrease the amount of code developers will need to write by a considerable margin. Especially in larger projects.

Are there any limitations to what can be chained? 

At what point would the a given page chain end?

If someone continues to try and call beyond the chain ending, what would the result be? (I'm expecting call to unexpected/undefined error given what I'm seeing.)

I think as long as the expectations are documented well enough then this is valuable to have internal.

lib/Input.js
```diff
@@ -138,11 +138,18 @@ class Mouse {
 
   async _doubleRaf() {
```
pavelfeldman, line 139: Why do you have double raf?

ebidel:
Yea, I guess my assumption, if `let page` is a Page instance, is that you'd be able pick up from that rather than starting a new chain. Like this:

```javascript
let page = await browser.newPage()
    .setViewport({width: 800, height: 800})
    .navigate('https://google.com');
await page.waitFor('input[title=Search]', {visible: true})
    .type('godfather')
    .click('input[type=submit]');
```

Agreed with @Garbee though. Anytime I see a chain API that works with async code, it raises a ton of questions. Examples will go a long way. Eventually, we could even show the different ways to call the API within the docs (chaining, .then, and async/await examples) so there are no surprises.

Garbee:
The chain system also reflects what [Chromeless](https://github.com/graphcool/chromeless) provides. Which has been garnishing some attention lately.

I'm opening an issue about docs on this stuff. I'd prefer if docs are the main holdup here we have a dedicated meta issue to discuss and make smaller actionable issues from.

aslushnikov:
Summarizing all the concerns about this proposal:
- [this implementation of] chaining feels like a new API surface. We'll be unhappy if we end up with two API styles simultaneously.
- chaining implementations are different from library to library, so it makes learning curve steeper.
- chaining could be easily implemented as a third-party library.

Most importantly, it's unclear whether clients will interact with puppeteer directly or whether puppeteer will power other libraries/tools which wrap it in a nicer api. We'll be happy with either (or both) of the outcomes, but chaining is a clear step towards the former. We should hold on shipping chaining in core unless there's a strong signal for it.

avimar:
How can I implement this myself, as a "third-party library"?

I am using node 6.11.2 without `await` function, and instead of this nice code

```
page.waitFor(fp.random(600,3200)))
      .focus('#ember501'))
      .type('29.95',{delay: 60}))
```

I had to use this kludge
```
.then(p=>page.waitFor(fp.random(600,3200)))
.then(p=>page.focus('#ember501'))
.then(p=>page.type('29.95',{delay: 60}))
```

avimar:
> Most importantly, it's unclear whether clients will interact with puppeteer directly or whether puppeteer will power other libraries/tools which wrap it in a nicer api.

Since it looks like you've _already_ provided all the functionality that `nightmare.js` provides with a similar API (except the chaining), I'd expect people will start using this directly.

(I actually came here because nightmare had issues with an iframe and only a self-proclaimed hack to deal with it. puppeteer had no problem with that use case.)

JoelEinbinder:
If Promises are painful in node6, you can use a generator instead of an async function, `yield` instead of `await`, and wrap it in a function that repeatedly calls the generator. You can also take the code from this PR and add it as something like `Promise.chain`. Except for type checking, nothing in this PR is specific to Puppeteer.

avimar:
Oh, Node8 is going LTS by next month. I guess that's not so long.

Still, it's a lot less lines. I'm not sure how that would work to implement myself. I would add `chain` to the the Promise prototype, then call `Promise.chain('page')` and it would then make subsequent chained commands reference the `page` class?

entrptaher:
@Garbee @aslushnikov,

I tried to do this with my little knowledge without even watching how you guys coded it :D. Now I see what you did there. 

This way you don't have to modify the actual puppeteer, instead it'd be an wrapper. Let's see if this helps anyone and if we can improve this.

```js
// app.js
const Chain = require("pup-chain");

(async() => {

  let urls = ["https://example.com", "https://example.org"];

  for (url of urls) {
    let mainChain = await new Chain(page);

    newurl = await mainChain
      .goto(url)
      .url()
      .run();

    title = await mainChain
      .title()
      .run();

    console.log(`${url} > URL: "${newurl}", Title: "${title}"`);
  }

  browser.close();
})();
```

and here is the wrapper,
``` js
// pup-chain
const ObjProperties = require("simple-property-retriever");
const pWaterfall = require('p-waterfall');

class Chain {
  constructor(initiator) {
    this.initiated = true; // idk, I just put it here to track things. Dead code for now.
    this.chain = []; // the container for all chains
    let addMethods = (ref) => {
      this.ref = ref;
      const properties = ObjProperties.getOwnNonenumerables(ref.__proto__);
      for (const property of properties) {
        // if (property === "constructor")
        //   continue;

        this[property] = function(...args) {
          this.chain.push(() => this.ref[property](...args))
          return this;
        };
      };
      return this.ref;
    }
    addMethods(initiator)
  }

  run() {
    let chain = [...this.chain];
    this.chain = [];
    return pWaterfall(chain)
  }
};

module.exports = Chain;
```

This outputs as below.
```
https://example.com > URL: "https://example.com/", Title: "Example Domain"
https://example.org > URL: "https://example.org/", Title: "Example Domain"
```

## Case

```
newurl = await mainChain
  .goto(url)
  .url()
  .run();
  
title = await mainChain
  .title()
  .run();
```

that compared to,
```
await page.goto(url);
await page.url();
await page.title();
```

Obviously this is not saving the space or characters, nor does it have that much control, instead it looks worse.

But now we have another control, `mainChain.goto(url).DoThis().DoThis().DoThis().DoThis().DoThis().DoThis().DoThis().DoThat().url()` will add the puppeteer methods/calls to the chain till I call the .run().

Also we can impliment something that would save the data from all chained commands.

For my version, if I want any data anytime, I can just add `.run`, Is there any way to omit the .run() in the end? Any help regarding the queue processing will be appreciated. 

### The first example
Here, I rewrote the first example you put in at the beginning. Created the chain, then ran it at the end.

```js
(async() => {
  const Chain = require("pup-chain");

  // 1. Navigate to google.com

  // prepare the chain
  const chain = await new Chain(page);

  // 2. Type 'godfather' and trigger search.
  await chain
    .setViewport({width: 800, height: 800})
    .goto('https://google.com')
    .waitFor('input[title=Search]', {visible: true})
    .type('godfather')
    .click('input[type=submit]');

  // 3. Navigate to each result one-by-one and
  //    take full page screenshot of each.
  for (let i = 0; i < 4; ++i) {
    // Wait for search result link to appear and click it
    const selector = `div.g:nth-child(${i + 1}) h3 a`;
    await chain
      .waitFor(selector)
      .click(selector)
      .waitForNavigation({waitUntil: 'networkidle'})
      .screenshot({
        fullPage: true,
        path: `screenshot-${i + 1}.pngnavigate`
      })
      .goBack();
  }
 
  // RUN IT ALTOGETHER
  await chain.run();
  browser.close();
})();
```

Result:
image

PR #215, opened 2017-08-07
Aborted network requests should use protocol's "Failed" status

aslushnikov:
The "Aborted" status also has a side-effect of aborting navigation
(there will be no error page).

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -770,6 +770,16 @@ describe('Page', function() {
       ]);
       expect(request.headers['foo']).toBe('bar');
     }));
+    it('should fail navigation when aborting main resource', SX(async function() {
+      page.setRequestInterceptor(request => request.abort());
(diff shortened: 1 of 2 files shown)


JoelEinbinder:
Let's add a test

test/test.js
```diff
@@ -770,6 +770,16 @@ describe('Page', function() {
...
     }));
+    it('should fail navigation when aborting main resource', SX(async function() {
+      page.setRequestInterceptor(request => request.abort());
+      let error = null;
+      try {
+        await page.navigate(EMPTY_PAGE);
+      } catch (e) {
+        error = e;
+      }
+      expect(error).toBeTruthy();
```
JoelEinbinder, line 781: check message on the error please

JoelEinbinder (approved):

Issue #178, opened 2017-08-01
examples audit


paulirish:
- [x] add example for saving a PDF
- [x] add example: navigate to google, enter search query, click the submit button
   - maybe do something with results?
   - add pavel's
- [x] screenshot
   - save screenshot to disk. done.
   - get and save fullpage screenshot
   - get and save emulated iphone screenshot
- [x] loadurlwithoutcss
   - update so that it works.
- [x] unrandomize
   - what is this example actually doing? delete?
- [ ] pagecallback
   - needs comments to explain what the example is actually doing.  
- [x] detectsniff
   - at least the setTimeout should be done via an await.
   - would be nice to use a proxy instead of defineGetter, but that's lower priority
   - i'd recommend a function declaration for passing into evaluateOnNewDocument. we use that convention in lighthouse to make it more clear that its' a pageFunction. (colorWheel does this already :)
- [x] waitFor example

ebidel:
For examples, we should start [running them through the linter](https://github.com/GoogleChrome/puppeteer/issues/185) and clean up the code a bit so they're consistent and using es6. e.g. use `const/let` and a top-level async func to remove the mix of `.then()` and async/await. I think that's less common to see. 

For example, this:

```
var browser = new Browser({headless: false});
browser.newPage().then(async page => {
    page.setRequestInterceptor(request => {
        if (request.url.endsWith('.css'))
            request.abort();
        else
            request.continue();
    });
    var success = await page.navigate(address);
    if (!success)
        console.log('Unable to load the address!');
    browser.close();
```

to maybe to something like this:

```
(async () => {

const browser = new Browser({headless: false});
const page = await browser.newPage();
page.setRequestInterceptor(request => {
  request.url.endsWith('.css') ?  request.abort() : request.continue();
});

const success = await page.navigate(address);
if (!success) {
  console.log('Unable to load the address!');
}
browser.close();

})();
```

JoelEinbinder:
Do we want to keep the phantom examples? They feel out of place.

aslushnikov:
I can't come up with a simple yet illustrative example for [page.addBinding](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageaddbindingname-puppeteerfunction) (former inPage callback).

However, documentation has a few examples which might be good enough.
Closing this for now.

P.S. Feel free to PR an example with `page.addBinding` / share an idea for it if any!

PR #216, opened 2017-08-07
Add pdf.js example

aslushnikov:
References #178

### DIFF
--- a/examples/pdf.js
+++ b/examples/pdf.js
@@ -19,8 +19,10 @@
 const {Browser} = require('puppeteer');
 const browser = new Browser();
 
-let page = await browser.newPage();
+const page = await browser.newPage();
 await page.navigate('https://news.ycombinator.com', {waitUntil: 'networkidle'});
+// page.pdf() is currently supported only in headless mode.
+// @see https://bugs.chromium.org/p/chromium/issues/detail?id=753118
 await page.pdf({
   path: 'hn.pdf',
   format: 'A4'
--- a/examples/pdf.js
+++ b/examples/pdf.js
@@ -1,0 +1,31 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


examples/pdf.js
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+(async () => {
+
+const {Browser} = require('puppeteer');
+const browser = new Browser();
+
+let page = await browser.newPage();
```
ebidel: `const page`

examples/pdf.js
```diff
@@ -0,0 +1,31 @@
...
+await page.navigate('https://news.ycombinator.com', {waitUntil: 'networkidle'});
+await page.pdf({
```
ebidel: comment that saving a pdf is only supported in headless?

examples/pdf.js
```diff
@@ -0,0 +1,31 @@
...
+ * limitations under the License.
+ */
+
+(async () => {
+
+const {Browser} = require('puppeteer');
+const browser = new Browser();
+
+let page = await browser.newPage();
+await page.navigate('https://news.ycombinator.com', {waitUntil: 'networkidle'});
```
ebidel: Might be noisy, but but it may be helpful for noobs to see return values when they're reading these examples. Up to you.

```
const response = await page.navigate('https://news.ycombinator.com', {waitUntil: 'networkidle'});
const buffer = await page.pdf({
  path: 'hn.pdf',
  format: 'A4'
});
```

ebidel (approved):
small suggestions

examples/pdf.js
JoelEinbinder: It fails linting if you do that.

ebidel: Figured it would. No biggy.

examples/pdf.js
aslushnikov: Done

examples/pdf.js
aslushnikov: Left as-is to avoid fighting eslint

examples/pdf.js
aslushnikov: Done

PR #218, opened 2017-08-08
Make interception work with redirects

### LINKED ISSUE #217: Request interception doesn't work with redirects
The following script fails for me (since google.com redirects to www.google.com):

```js
(async() => {

await page.setRequestInterceptor(request => request.continue());
await page.navigate('https://google.com');
browser.close();

})();
```

aslushnikov:
This patch:
- changes interception API so that it better aligns with what we'd like to see
  in #121
- fixes the issue with redirect interception

Fixes #217.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -113,7 +113,7 @@
     + [response.url](#responseurl)
   * [class: InterceptedRequest](#class-interceptedrequest)
     + [interceptedRequest.abort()](#interceptedrequestabort)
-    + [interceptedRequest.continue()](#interceptedrequestcontinue)
+    + [interceptedRequest.continue([overwrites])](#interceptedrequestcontinueoverwrites)
     + [interceptedRequest.headers](#interceptedrequestheaders)
     + [interceptedRequest.method](#interceptedrequestmethod)
     + [interceptedRequest.postData](#interceptedrequestpostdata)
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -1112,40 +1112,38 @@ Contains the URL of the response.
...
-[InterceptedRequest] represents an intercepted request, which can be mutated and either continued or aborted. [InterceptedRequest] which is not continued or aborted will be in a 'hanging' state.
+[InterceptedRequest] represents an intercepted request, which can be either continued or aborted. [InterceptedRequest] which is not continued or aborted will be in a 'hanging' state.
 
 #### interceptedRequest.abort()
 
 Aborts request.
 
-#### interceptedRequest.continue()
+#### interceptedRequest.continue([overwrites])
+- `overwrites` <[Object]> Optional request overwrites, which could be one of the following:
```
JoelEinbinder: overrides

lib/NetworkManager.js
```diff
@@ -247,17 +247,23 @@ class InterceptedRequest {
...
+      headers = {};
+      for (let entry of overwrites.headers.entries())
+        headers[entry[0]] = entry[1];
+    }
     this._client.send('Network.continueInterceptedRequest', {
       interceptionId: this._interceptionId,
-      url: this.url,
-      method: this.method,
-      postData: this.postData,
+      url: overwrites.url,
```
JoelEinbinder: Let's make `.url` a getter, as well as other public properties.

test/test.js
```diff
@@ -781,6 +782,12 @@ describe('Page', function() {
       expect(error).toBeTruthy();
       expect(error.message).toContain('Failed to navigate');
     }));
+    it('should work with redirects', SX(async function() {
+      server.setRedirect('/foo.html', '/empty.html');
+      await page.setRequestInterceptor(request => request.continue());
+      let response = await page.navigate(PREFIX + '/foo.html');
+      expect(response.status).toBe(200);
```
JoelEinbinder: Can we check the content here. Do server redirects get reset after each test?

lib/NetworkManager.js
aslushnikov: this will be changed altogether with #121 fixed

test/test.js
aslushnikov: Changed foo.html into non-existing-url.html to make it clear that it doesn't exist.

docs/api.md
```diff
@@ -1112,40 +1112,38 @@ Contains the URL of the response.
...
-#### interceptedRequest.continue()
+#### interceptedRequest.continue([overrides])
+- `overrides` <[Object]> Optional request overwrites, which could be one of the following:
+  - `url` <[string]> If set, the request url will be changed
+  - `method` <[string]> If set changes the request method (e.g. `GET` or `POST`)
+  - `postData` <[string]> If set changes the post data of request
+  - `headers` <[Map]> If set changes the request HTTP headers
 
-Continues request.
+Continues request with optional request overwrites.
```
JoelEinbinder: overrides

JoelEinbinder (approved):

Issue #65, opened 2017-07-08
lint jsdoc types


aslushnikov:
We should lint jsdoc types.

I spent some time trying to make closure compiler work with node.js, but it didn't work due to issues with node module requires.

JoelEinbinder:
I tried using typescript to do type checking. I had some success after removing all of our requires and module exports, and concatenating everything into one big file. There were a lot of errors because our `WebSocket` is different from browser's `WebSocket`, and typescript gets confused by `frame.evaluate` because of the context switch and the way arguments are passed in.

sukrosono:
Did the code base will use typescript after pull number 219 finished ? :fearful:

aslushnikov:
> Did the code base will use typescript after pull number 219 finished ? 😨

@brutalcrozt no, we'll continue using JS with jsdoc annotations.

PR #219, opened 2017-08-08
Proposal: Lint JSDoc with Typescript

JoelEinbinder:
I think I can make this PR a bit cleaner by shuffling around some code, but this is the basic idea.
#65


lib/Browser.js
```diff
@@ -18,8 +18,9 @@ const {Duplex} = require('stream');
 const path = require('path');
 const helper = require('./helper');
 const removeRecursive = require('rimraf').sync;
-const Page = require('./Page');
+const Page = require('./page');
```
aslushnikov: why this change?

package.json
```diff
@@ -12,11 +12,12 @@
     "debug-unit": "DEBUG_TEST=true node --inspect-brk ./node_modules/.bin/jasmine test/test.js",
     "test-phantom": "python third_party/phantomjs/test/run-tests.py",
     "test-doclint": "jasmine utils/doclint/check_public_api/test/test.js && jasmine utils/doclint/preprocessor/test.js",
-    "test": "npm run lint --silent && npm run coverage && npm run test-phantom && npm run test-doclint",
+    "test": "npm run lint --silent && npm run type && npm run coverage && npm run test-phantom && npm run test-doclint",
     "install": "node install.js",
     "lint": "([ \"$CI\" = true ] && eslint --quiet -f codeframe . || eslint .) && npm run doc",
```
aslushnikov, line 17: let's add it to lint as well

utils/doclint/check_public_api/JSBuilder.js
```diff
@@ -81,12 +81,16 @@ class JSOutline {
...
     const args = [];
     for (let param of node.value.params) {
-      if (param.type === 'AssignmentPattern')
+      if (param.type === 'AssignmentPattern' && param.left.type === 'Identifier')
         args.push(new Documentation.Argument(param.left.name));
       else if (param.type === 'RestElement')
         args.push(new Documentation.Argument('...' + param.argument.name));
       else if (param.type === 'Identifier')
         args.push(new Documentation.Argument(param.name));
+      else if (param.type === 'ObjectPattern')
```
aslushnikov, line 90: this is a good change by itself. Can we have it separately?

package.json
```diff
@@ -43,6 +50,7 @@
     "pdfjs-dist": "^1.8.595",
     "pixelmatch": "^4.0.2",
     "pngjs": "^3.2.0",
-    "text-diff": "^1.0.1"
+    "text-diff": "^1.0.1",
+    "typescript": "^2.5.0-dev.20170807"
```
aslushnikov, line 54: can we have something stable here?

lib/helper.js
```diff
@@ -14,16 +14,18 @@
...
 class Helper {
   /**
-   * @param {function()|string} fun
+   * @param {Function|string} fun
    * @param {!Array<*>} args
    * @return {string}
    */
   static evaluationString(fun, ...args) {
-    if (Helper.isString(fun)) {
+    if (!(fun instanceof Function)) {
```
aslushnikov: why this change?

lib/Browser.js
```diff
@@ -18,8 +18,9 @@ const {Duplex} = require('stream');
...
 const childProcess = require('child_process');
+/** @typedef {*} ChildProcess */
```
aslushnikov: what is this?

lib/Dialog.js
```diff
@@ -15,11 +15,12 @@
  */
 
 const helper = require('./helper');
+const Connection = require('./connection'); // eslint-disable-line
```
aslushnikov: let's not do any forward declarations

lib/EmulationManager.js
```diff
@@ -61,7 +63,7 @@ class EmulationManager {
 
     function injectedTouchEventsFunction() {
       const touchEvents = ['ontouchstart', 'ontouchend', 'ontouchmove', 'ontouchcancel'];
-      const recepients = [window.__proto__, document.__proto__];
+      const recepients = [/** @type {*} */(window).__proto__, /** @type {*} */(document).__proto__];
```
aslushnikov, line 64: why type anything here? it doesn't add any information

lib/Page.js
```diff
@@ -25,13 +25,17 @@ const FrameManager = require('./FrameManager');
 const {Keyboard, Mouse} = require('./Input');
 const Tracing = require('./Tracing');
 const helper = require('./helper');
+const Connection = require('./connection'); // eslint-disable-line
+/** @typedef {*} TaskQueue */
```
aslushnikov: TaskQueue is defined below - no need to typedef it. Or is it a forward declaration?

lib/Page.js
```diff
@@ -314,23 +318,23 @@ class Page extends EventEmitter {
...
+   * @return {!Promise<Response|null>}
    */
   async goForward(options) {
     return this._go(+1, options);
   }
 
   /**
    * @param {!Object=} options
-   * @return {!Promise<?Response>}
+   * @return {!Promise<Response|null>}
```
aslushnikov, line 334: can we have question marks? they are so much more readable then "|null"

lib/Page.js
```diff
@@ -651,14 +657,12 @@ let unitToPixels = {
  * @return {(number|undefined)}
  */
 function convertPrintParameterToInches(parameter) {
-  if (typeof parameter === 'undefined')
-    return undefined;
   let pixels;
   if (helper.isNumber(parameter)) {
     // Treat numbers as pixel values to be aligned with phantom's paperSize.
-    pixels = /** @type {number} */ (parameter);
+    pixels = Number(parameter);
```
aslushnikov, line 660: why isn't typecast enough?

lib/Page.js
```diff
@@ -651,14 +657,12 @@ let unitToPixels = {
...
   } else if (helper.isString(parameter)) {
-    let text = parameter;
+    let text = String(parameter);
```
aslushnikov, line 662: can we typecast here?

lib/helper.js
```diff
@@ -181,16 +183,16 @@ class Helper {
   /**
    * @param {!EventEmitter} emitter
    * @param {string} eventName
-   * @param {function(?)} handler
-   * @return {{emitter: !EventEmitter, eventName: string, handler: function(?)}}
+   * @param {function(any)} handler
+   * @return {{emitter: NodeJS.EventEmitter, eventName: string, handler: any}}
```
aslushnikov, line 184: where does NodeJS.EventEmitter type come from?

aslushnikov (commented):
Looks good. I like the absense of "!" a lot!
Can we also:
- keep question marks. Typing "|null" is verbose and not an improvement.
- do not do forward declarations. These look weird

I also wonder how hard is it to author these declarations without VSCode. Some of them look pretty random (like ChildProcess one) - does tsc help to add them?

lib/EmulationManager.js
JoelEinbinder: __proto__ is a chrome only property. It doesn't exist on window as far as typescript is concerned. I can do tsignore instead.

lib/Page.js
JoelEinbinder: TaskQueue is defined in FrameManager, but I don't have a good way to export it at the moment.

lib/Page.js
JoelEinbinder: Unfortunately question marks are unsupported. I'm working on a patch.

lib/Page.js
JoelEinbinder: We shouldn't type things as number when they could be a `Number`. But yeah, typecast should be enough now that I've upgraded to typescript@next.

lib/helper.js
JoelEinbinder: Typescript doesn't know that Helper.isString implies that fun is now a string. instanceof does.

lib/helper.js
JoelEinbinder: EventEmitter is a type defined on the NodeJS namespace. I don't know why it didn't make me use it above. I gave up on this line and moved on.

lib/helper.js
aslushnikov: I'd rather keep the logic straight-forward and typecast.

package.json
JoelEinbinder: Not at the moment. dev gives us support for typecasting.

utils/doclint/check_public_api/JSBuilder.js
JoelEinbinder: Sure!

lib/Page.js
aslushnikov: > We shouldn't type things as number when they could be a Number.

I don't like amending runtime behavior to please linter. Especially in the case of Number / number types which are almost indistinguishable from one another.

lib/Page.js
aslushnikov: I hope there's a way to solve it together with forward declarations

lib/EmulationManager.js
aslushnikov: does typescript actually understand there are two contexts here - node.js and browser? If yes - then how does it do it? If no - does it mean it'd allow `window` in node?

JoelEinbinder: It allows window in node. I can get it not to by messing with the libraries we include, but it works out nicely for this use case.

package.json
aslushnikov: What kind of dev is this - is it tip-of-tree? When will the necessary changes hit stable?

JoelEinbinder: It appears to be a monthly release. So September.

lib/EmulationManager.js
aslushnikov: sure, I think window is fine.
Do you plan to add any externs.js to aid linting? in this case, we can define `window.__proto__` and probably some other things there to reduce "noise" in the source base.

JoelEinbinder:
this patch is currently blocked on https://github.com/Microsoft/TypeScript/issues/17953

lib/EmulationManager.js
```diff
@@ -75,4 +75,14 @@ class EmulationManager {
   }
 }
 
+/**
+ * @typedef {Object} EmulationManager.Viewport
```
nojvek, line 79: Oh man I love what you've done here.

I think I need to get back to my PR in typescript that allows generating .d.ts files from js files when checkJs is enabled.

Would be great to have code completion when writing puppeteer code.

tsconfig.json
```diff
@@ -0,0 +1,11 @@
+{
+    "compilerOptions": {
```
mohsen1, line 2: do you want to enable "declaration" to get `.d.ts` files?

nojvek:
That doesn’t work quite yet. .d.ts files from js will throw error.

I have a PR that I need to finish.

lib/Page.js
plantain-00: TypeScript 2.5 should recognise this now: https://github.com/Microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#type-assertioncast-syntax-in-checkjsts-check-mode

aslushnikov:
Closing in favor of #986 (which is a rebaseline of this one)

tsconfig.json
nojvek: That's still not currently supported.

Issue #206, opened 2017-08-04
page.mouse.click() doesn't work


aslushnikov:
The `page.mouse.click()` might cause navigation, which results in underlying `mouse._doubleRaf` hack to never complete.

For example, consider the following unit test:
```js
it('should click links which cause navigation', SX(async function() {
  await page.setContent(`<a href="${EMPTY_PAGE}">empty.html</a>`);
  // This await should not hang.
  await page.click('a');
}))
```

The core issue is that `Runtime.evaluate({awaitPromise: true})` does not throw when the parent execution context gets destroyed. Instead, it throws after indefinite amount of time, as the promise eventually gets collected.

@ak239 promised to check if this is something fixable in v8.

aslushnikov:
Fixed by https://github.com/GoogleChrome/puppeteer/pull/240

PR #220, opened 2017-08-08
Google search example

JoelEinbinder:
#178

### DIFF
--- a/examples/search.js
+++ b/examples/search.js
@@ -20,10 +20,11 @@ const {Browser} = require('..');
 const browser = new Browser();
 
 const page = await browser.newPage();
-await page.navigate('https://google.com', {waitUntil: 'networkidle'});
+await page.goto('https://google.com', {waitUntil: 'networkidle'});
 // Type our query into the search bar
 await page.type('puppeteer');
-await page.press('Enter');
+
+await page.click('input[type="submit"]');
 
 // Wait for the results to show up
 await page.waitForSelector('h3 a');
--- a/examples/search.js
+++ b/examples/search.js
@@ -16,7 +16,7 @@
 
 (async() => {
 
-const {Browser} = require('..');
+const {Browser} = require('puppeteer');
 const browser = new Browser();
 
 const page = await browser.newPage();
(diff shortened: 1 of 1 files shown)


aslushnikov:
This example is a sub-example of pavel's stress-test, let's have that one instead when the #206 is fixed?

JoelEinbinder:
I find the output of this easier to follow compared to pavel's stress test.

examples/search.js
```diff
@@ -0,0 +1,39 @@
...
+(async() => {
+
+const {Browser} = require('..');
+const browser = new Browser();
+
+const page = await browser.newPage();
+await page.navigate('https://google.com', {waitUntil: 'networkidle'});
+// Type our query into the search bar
+await page.type('puppeteer');
+await page.press('Enter');
```
aslushnikov: this doesn't work for me reliably - sometimes enter is "swallowed" and i'm stuck with the text in the text box.

examples/search.js
```diff
@@ -0,0 +1,39 @@
...
+
+// Wait for the results to show up
+await page.waitForSelector('h3 a');
```
ebidel: FWIW, other test runner examples I've seen look at the title to determine the results page is done.  Could also do that to be familiar: `await page.waitForFunction("document.title == 'puppeteer - Google Search'");`

aslushnikov: The title changes at indeterminate point of time - whereas the `page.waitForSelector('h3 a')` waits for the first search result appearing on the page. Sounds more reliable!

JoelEinbinder:
Sometimes if I run in headful mode, it only gets part of the search results because they stream in. Fixing that would make the example a bit more complicated, so I think the current state of it is fine.

examples/search.js
```diff
@@ -0,0 +1,40 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+(async() => {
+
+const {Browser} = require('..');
```
aslushnikov:
```js
require('puppeteer')
```

aslushnikov (approved):
Interestingly, works smoothly for me now

PR #221, opened 2017-08-08
Remove unrandomize example

JoelEinbinder:
#178


aslushnikov (approved):

aslushnikov:
@JoelEinbinder I just recalled there's a reference to the example from the documetation: https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageevaluateonnewdocumentpagefunction-args

Let's remove it as well

Issue #223, opened 2017-08-08
PhantomJS shim - CasperJS compatibility


kdzwinel:
[CasperJS](http://casperjs.org/) is probably the most popular JS framework for functional testing. Originally, it was supporting only PhantomJS, but today it also supports [SlimerJS](https://slimerjs.org/) (Firefox).  And since it already supports multiple targets it should be possible to add puppeteer as a new one. However, since puppeteer has a phantom shim, it'd be nice if it worked with CasperJS out of the box. I've tried using it as a replacement and found these issues:

1) I haven't found any docs (or examples) on how to use the shim - mentioning in the docs that `phantomjs [script.js]` command equals `node ./phantom_shim/runner.js [script.js]` would probably be enough
2) shim requires `deasync` and `ncp` packages that are in the `devDependencies` section of the `package.json` - they aren't normally installed and shim doesn't work w/o them
3) currently, shim introduces itself as version `0.0.1` of phantomjs - this won't work with Casper as it [checks for the compatible version number](https://github.com/casperjs/casperjs/blob/master/bin/bootstrap.js#L138). I suggest exposing the same version number as the PhantomJS version that is being shimmed.
4) `defaultPageSettings` are missing from the `phanotm` object:
```javascript
var defaultUserAgent = phantom.defaultPageSettings.userAgent
//TypeError: Cannot read property 'userAgent' of undefined
```
5) `WebPage.openUrl` is missing

I got to a point where simple asserts (`1 === 1`) are working but got stuck on that missing [openUrl](https://github.com/ariya/phantomjs/blob/35c0cbb82b614268a09cf4728d0ad19634a7c6de/src/webpage.cpp#L893) method.

Let me know if this is useful and if the aim of the PhantomJS shim is to support such use cases as CasperJS.

JoelEinbinder:
I think we have been misleading with the Phantom shim. The goal of it is just to run the PhantomJS tests, not for actual use out in the wild. We want to run the Phantom tests to prove to ourselves that the Puppeteer API is capable and relevant.

Writing an actual compatibility layer is definitely doable, but it would require quite a bit more work. We cut a lot of corners, for example only supporting a few keys for input: https://github.com/GoogleChrome/puppeteer/blob/master/phantom_shim/WebPage.js#L73

https://github.com/googlechrome/puppeteer#q-what-is-the-phantom-shim

kdzwinel:
[...]

PR #224, opened 2017-08-08
Move phantom_shim into test folder

JoelEinbinder:
`phantom_shim` is part of our test code, but its placement at the top of the repository is very misleading about what we intend it to be used for.

#223


aslushnikov (approved):
lgtm (bots don't pass though)

aslushnikov:
closing this - we opted into removing the shim instead.

PR #225, opened 2017-08-08
User browser endpoint for creating and removing targets.

pavelfeldman:


lib/Connection.js
```diff
@@ -123,6 +133,67 @@ class Connection extends EventEmitter {
...
+      else
+        callback.resolve(object.result);
+    } else {
+      console.assert(!object.id);
+      this.emit(object.method, object.params);
+    }
+  }
+
+  async dispose() {
+    this._connection._sessions.delete(this._sessionId);
```
aslushnikov, line 192: we should also reject all pending messages so that they don't hang

lib/Connection.js
```diff
@@ -123,6 +133,67 @@ class Connection extends EventEmitter {
...
+  /**
+   * @param {string} method
+   * @param {!Object=} params
+   * @return {!Promise<?Object>}
+   */
+  send(method, params = {}) {
+    let id = ++this._lastId;
+    let message = JSON.stringify({id, method, params});
+    this._connection.send('Target.sendMessageToTarget', {sessionId: this._sessionId, message});
+    return new Promise((resolve, reject) => {
```
aslushnikov, line 168: can we just return the connection promise? We won't need the callbacks map in session at all, and we might remove the `Session._onMessage` 

```js
return this._connection.send('Target.sendMessageToTarget', {sessionId: this._sessionId, message});
```

aslushnikov:
Closing in favor of #230

PR #230, opened 2017-08-09
Roll chromium to 492629

aslushnikov:
This patch:

- rolls chromium to 492629
- migrates connection establishing to use browser target. This migration means
  that now we have a single websocket connection to browser (implemented
  in Connection class). A connection to a particular target is
  incapsulated in a new Session class.


JoelEinbinder (approved):
I am excited about eventually being able to list all the tabs.

Lusha: Nice work Pavel!

PR #231, opened 2017-08-10
Introduce page.emulate convenience wrapper

pavelfeldman:


test/test.js
```diff
@@ -1421,6 +1421,11 @@ describe('Page', function() {
       await page.setViewport({width: 100, height: 100});
       expect(await page.evaluate(() => screen.orientation.type)).toBe('portrait-primary');
     }));
+    it('should support emulate shorthand', SX(async function() {
+      await page.emulate({viewport:{width: 123, height: 456}, userAgent:'UA'});
```
aslushnikov: we have a device here for convenience:

```js
await page.emulate(iPhone);
expect(await page.evaluate(() => window.innerWidth)).toBe(375);
expect(await page.evaluate(() => navigator.userAgent)).toContain('Safari');
```

aslushnikov (approved):
thanks!

FYI: linter is unhappy

examples/pdf.js
```diff
@@ -25,7 +25,7 @@ await page.navigate('https://news.ycombinator.com', {waitUntil: 'networkidle'});
 // @see https://bugs.chromium.org/p/chromium/issues/detail?id=753118
 await page.pdf({
   path: 'hn.pdf',
-  format: 'A4'
```
JoelEinbinder, line 28: Why this change?

aslushnikov: I guess the "letter" page format is more widely known then "A4"

PR #233, opened 2017-08-10
Doc polish. Example consistency

ebidel:
Also added some more use case examples.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,9 +15,9 @@ You generally only need to submit a CLA once, so if you've already submitted one
 (even if it was for a different project), you probably don't need to do it
 again.
 
-## Getting the Code
+## Getting setup
 
-1. Clone repository
+1. Clone this repository
 ```bash
 git clone https://github.com/GoogleChrome/puppeteer
 cd puppeteer
@@ -123,22 +123,21 @@ are used to test `phantom_shim`.
 
 Every public API method or event should be called at least once in tests. To ensure this, there's a coverage command which tracks calls to public API and reports back if some methods/events were not called.
 
-- To run coverage:
+Run coverage:
+
 ```
 npm run coverage
 ```
 
 ## Debugging Puppeteer
-Puppeteer uses [DEBUG](https://github.com/visionmedia/debug) module to expose some of it's inner guts under the `puppeteer` namespace.
-Try putting the following script in the `script.js` and running it via `DEBUG=* node script.js`:
+Puppeteer uses [DEBUG](https://github.com/visionmedia/debug) module to expose some of it's inner guts under the `puppeteer` namespace. Try putting the following in a file called `script.js` and running it via `DEBUG=* node script.js`:
 
 ```js
 const {Browser} = require('puppeteer');
 const browser = new Browser();
-browser.newPage().then(async page => {
-  await page.goto('https://example.com');
-  browser.close();
-});
+
+browser.newPage().then(() => page.goto('https://example.com'));
+browser.close();
 ```
 
 Tips-n-tricks:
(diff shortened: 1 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -123,22 +123,21 @@ are used to test `phantom_shim`.
...
 
 ```js
-browser.newPage().then(async page => {
-  await page.goto('https://example.com');
-  browser.close();
-});
+
+browser.newPage().then(() => page.goto('https://example.com'));
```
aslushnikov: the 'browser.newPage()' is not awaited and does not have time to run - the `browser.close` on the next line will close chrome immediately.
(also, the arrow function should accept `page` object as a parameter - or `page is not undefined` will be thrown)

README.md
```diff
@@ -1,93 +1,136 @@
...
+*Puppeteer requires Node version 7.10 or greater*
+
+To use Puppeteer in your project, run:
 ```
 yarn add puppeteer
+# or "npm i puppeteer"
 ```
 
-> **NOTE** Puppeteer bundles a version of Chromium (~90Mb) which it is guaranteed to work with. However, you're free to point Puppeteer to any Chromium executable ([documentation](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#new-browseroptions))
+> **Note**: When you install Puppeteer, it downloads a recent version of Chromium (~71Mb) that is guaranteed to work with the API. However, you can [tell Puppeteer to use any Chromium executable](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#new-browseroptions) installed on the machine.
```
aslushnikov: nit: it's 70mb on mac and 90 on linux, or maybe vice versa :)

README.md
```diff
@@ -1,93 +1,136 @@
...
+(async() => {
+const page = await browser.newPage();
+await page.goto('https://news.ycombinator.com', {waitUntil: 'networkidle'});
+await page.pdf({path: 'hn.pdf', format: 'A4'});
+
+browser.close();
+})();
+```
+
+See [`Page.pdf()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagepdfoptions) for more information creating pdfs.
```
aslushnikov: typo:
```
for more information **about** creating pdfs
```

README.md
```diff
@@ -1,93 +1,136 @@
...
 
 #### Q: What is Puppeteer?
 
-Puppeteer is a light-weight Node module to control headless Chrome using the [DevTools Protocol](https://chromedevtools.github.io/devtools-protocol/).
+Puppeteer is a light-weight Node module to control headless Chrome using the latest version of the [DevTools Protocol](https://chromedevtools.github.io/devtools-protocol/).
 
 #### Q: Which Chromium version does Puppeteer use?
 
-Puppeteer bundles chromium it works best with. As chromium improves over time,   new versions of puppeteer will be released which depend on a newer chromium versions.
+Look for `chromium_revision` in [package.json](https://github.com/GoogleChrome/puppeteer/blob/master/package.json).
```
aslushnikov, line 122: we can auto-insert the current chromium version for the tip-of-tree - does it worth it?

README.md
ebidel: done

README.md
ebidel: done

CONTRIBUTING.md
ebidel: done. moved it to async/await to align with all the other examples.

README.md
ebidel: You mean instead of having it harcoded in package.json in the first place, or do you mean dropping it in the readme? Latter sounds fancy.

ebidel:
PTAL

aslushnikov (approved):

PR #240, opened 2017-08-11
Upgrade chromium to r493673

JoelEinbinder:
Mouse events are no longer racy. Enabling touch no longer converts all mouse events into touches. Promises in destroyed execution contexts are rejected immediately.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -456,7 +456,7 @@ class WaitTask {
 
     // When the page is navigated, the promise is rejected.
     // We will try again in the new execution context.
-    if (error && error.message.startsWith('Protocol error (Runtime.awaitPromise): Execution context was destroyed'))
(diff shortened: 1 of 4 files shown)


lib/FrameManager.js
```diff
@@ -454,6 +454,11 @@ class WaitTask {
     if (!success && !error)
       return;
 
+    // When the page is navigated, the promise is rejected.
+    // We will try again in the new execution context.
+    if (error && error.message.startsWith('Protocol error (Runtime.awaitPromise): Execution context was destroyed'))
```
aslushnikov: good one! let's check for the `Execution context was destroyed` string only: the `Protocol error (Runtime.awaitPromise)` part is generated in Connection.js and no one expects it to be a part of the API

aslushnikov (approved):
nice!

PR #243, opened 2017-08-11
Plumb default prompt value to the dialog

aslushnikov:
This patch:
- plumbs default prompt value to the dialog
- enables one more phantom test


JoelEinbinder:
Can we call it defaultValue or just default? The prompt is the question, and it's not the default question

JoelEinbinder (approved):

PR #245, opened 2017-08-11
Add a test that uses Modernizr to detect touch

JoelEinbinder:
Detecting touch is hard, so I added a test that uses Modernizr's touch detection script to make sure that our emulation will be detected by sites in the wild.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -1429,7 +1429,7 @@ describe('Page', function() {
         return promise;
       }
     }));
-    fit('should be detectable by Modernizr', SX(async function(){
+    it('should be detectable by Modernizr', SX(async function(){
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -1429,6 +1429,13 @@ describe('Page', function() {
         return promise;
       }
     }));
+    fit('should be detectable by Modernizr', SX(async function(){
```
aslushnikov: remove fit

aslushnikov (approved):
shell we remove other tests that check for touches?

JoelEinbinder:
I like the other touch tests because I wrote them and understand how they work. This is just a good sanity test to have.

PR #248, opened 2017-08-11
Implement basic element handles

aslushnikov:
This patch implements basic element handles which a backed with remote objects.

References #111

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -265,7 +265,7 @@ Emitted when a [response] is received.
 
 #### page.$(selector)
 - `selector` <[string]> Selector to query page for
-- returns: <[Promise]<[Element]>> Promise which resolves to Element pointing to the page element.
+- returns: <[Promise]<[ElementHandle]>> Promise which resolves to ElementHandle pointing to the page element.
 
 The method queries page for the selector. If there's no such element on the page, the method will resolve to `null`.
 
@@ -920,7 +920,7 @@ puppeteer.launch().then(async browser => {
 
 #### frame.$(selector)
 - `selector` <[string]> Selector to query page for
-- returns: <[Promise]<[Element]>> Promise which resolves to Element pointing to the page element.
+- returns: <[Promise]<[ElementHandle]>> Promise which resolves to ElementHandle pointing to the page element.
 
 The method queries page for the selector. If there's no such element on the page, the method will resolve to `null`.
 
@@ -1056,11 +1056,20 @@ puppeteer.launch().then(async browser => {
 
 ### class: ElementHandle
 
-ElementHandle holds a reference to the in-page element. Element handles could be thought about as
-a javascript reference. Element handles could be created with the [page.$](#pageselector) function.
+ElementHandle represents an in-page DOM element. ElementHandles could be created with the [page.$](#pageselector) method.
 
-Element handles keep underlying elements from garbage collection unless they are [released](#elementhandlerelease).
-Element handles also get auto-released when their origin frame gets navigated.
+```js
+const puppeteer = require('puppeteer');
+puppeteer.launch().then(async browser => {
+  let page = await browser.newPage();
+  await page.goto('https://google.com');
+  let inputElement = await page.$('input[type=submit]');
+  await inputElement.click();
+  ...
+});
+```
+
+ElementHandle prevents DOM element from garbage collection unless the handle is [released](#elementhandlerelease). ElementHandles are auto-released when their origin frame gets navigated.
 
 #### elementHandle.click([options])
 - `options` <[Object]>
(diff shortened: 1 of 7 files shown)


docs/api.md
```diff
@@ -1225,3 +1275,4 @@ Contains `POST` data for `POST` requests.
 [Map]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map "Map"
 [selector]: https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors "selector"
 [Tracing]: https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-tracing "Tracing"
+[Element]: https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-element "Element"
```
JoelEinbinder: [ElementHandle], and above.

lib/Connection.js
```diff
@@ -167,6 +167,8 @@ class Session extends EventEmitter {
    * @return {!Promise<?Object>}
    */
   async send(method, params = {}) {
+    if (!this._connection)
+      return Promise.reject(new Error(`Protocol error (${method}): Session closed. Most likely the inspected page has been closed.`));
```
JoelEinbinder: nit: inspected page seems like a devtools word.

lib/ElementHandle.js
```diff
@@ -0,0 +1,98 @@
...
+   */
+  async release() {
+    if (this._released)
+      return;
+    this._released = true;
+    await helper.releaseObject(this._client, this._remoteObject);
+  }
+
+  /**
+   * @param {function()|string} pageFunction
```
JoelEinbinder: Don't allow a string if you don't allow a string.

lib/FrameManager.js
```diff
@@ -171,12 +172,34 @@ class Frame {
    * @return {!Promise<(!Object|undefined)>}
    */
   async evaluate(pageFunction, ...args) {
+    let remoteObject = await this._rawEvaluate(pageFunction, ...args);
+    return await helper.serializeRemoteObject(this._client, remoteObject);
```
JoelEinbinder, line 176: Let's make serializeRemoteObject return an ElementHandle if it is a node.

lib/Page.js
```diff
@@ -551,23 +559,32 @@ class Page extends EventEmitter {
    * @return {!Promise}
    */
   async click(selector, options) {
-    await this.mainFrame().click(selector, options);
+    let handle = await this.$(selector);
+    console.assert(handle, 'No node found for selector: ' + selector);
+    await handle.click(options);
+    await handle.release();
```
JoelEinbinder, line 565: Do we need to await the release?

docs/api.md

lib/Connection.js
aslushnikov: Done.

lib/ElementHandle.js
aslushnikov: Done.

lib/Page.js
aslushnikov: Yes, otherwise the following might throw:
```js
await page.click();
page.close();
```

JoelEinbinder: Shouldn't close/navigate mark all handles as released?

lib/FrameManager.js
aslushnikov: Let me do this in a follow-up. This implies the `console` event changes and possibly something else, might be some work to do.
Also, the `page.waitFor` should resolve to elementHandles as well (another follow-up)

lib/Page.js
JoelEinbinder: In helper we just do a try-catch:
https://github.com/GoogleChrome/puppeteer/blob/master/lib/helper.js#L103

aslushnikov: It's the execution context being destroyed that renders all the element handles useless. (we don't release them explicitly for now)

It's nice to await `handle.release` so that there would be no in-flight messages when the connection is getting closed. (Connection rejects all the in-flight messages when it closes).

aslushnikov:
@JoelEinbinder could you please re-review this (I had to force-push)

docs/api.md
```diff
@@ -1055,6 +1054,66 @@ puppeteer.launch().then(async browser => {
...
+#### elementHandle.release()
+- returns: <[Promise]> Promise which resolves when the element handle is successfully released.
+
+The `elementHandle.release` method stops referencing the element handle.
+
+```js
+const {Browser} = require('puppeteer');
+const browser = new Browser();
+browser.newPage().then(async page =>
+  await page.setContent('<div>hello</div>');
```
JoelEinbinder, line 1108: How is this code relevant to `element.release`?

JoelEinbinder (approved):

PR #250, opened 2017-08-12
Respect CHROME_PATH env variable to launch chromium

aslushnikov:
This patch:
- renames Browser.Options.executablePath into Browser.Options.chromePath
- removes the test-related CHROME env variable
- teaches puppeteer to respect CHROME_PATH env variable

The motivation behind this patch:
- we already had the ability to run tests against custom-built chromium
  with CHROME env variable. It turned out, this would come handy for any
  puppeteer scripts
- this aligns puppeteer with [chrome launcher](https://github.com/GoogleChrome/lighthouse/tree/master/chrome-launcher), which uses CHROME_PATH for the same purpose

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -179,9 +179,8 @@ browser.newPage().then(async page => {
 - `options` <[Object]>  Set of configurable options to set on the browser. Can have the following fields:
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
   - `headless` <[boolean]> Whether to run chromium in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). Defaults to `true`.
-  - `executablePath` <[string]> Path to a chromium executable to run instead of bundled chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
-  - `slowMo` <[number]> Slows down Puppeteer operations by the specified amount of milliseconds. Useful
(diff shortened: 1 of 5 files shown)


JoelEinbinder:
Puppeteer is a module, not an application. I don't like it reading my environment variables.

aslushnikov:
Chrome launcher does this already. So If/when we migrate to chrome launcher, we'll get this behavior out-of-the-box.

I also generally like the idea of aligning with similar projects.

aslushnikov:
Closing this - we're afraid that CHROME_PATH is too widely used, and that puppeteer is currently working with a specific chrome versions only

docs/api.md
```diff
@@ -179,9 +179,8 @@ browser.newPage().then(async page => {
 - `options` <[Object]>  Set of configurable options to set on the browser. Can have the following fields:
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
   - `headless` <[boolean]> Whether to run chromium in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). Defaults to `true`.
-  - `executablePath` <[string]> Path to a chromium executable to run instead of bundled chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
```
paulirish, line 182: you should probably mention here that compatibility with various chrome versions is not guaranteed

aslushnikov: https://github.com/GoogleChrome/puppeteer/pull/260

PR #251, opened 2017-08-13
Roll chromium to r493985

aslushnikov:
This patch:
- rolls chromium to r493985
- migrates Browser.version() puppeteer method to use newly added
  Browser.getVersion() protocol method


JoelEinbinder (approved):
Nice!

PR #264, opened 2017-08-15
Introduce Puppeteer.connect method

### LINKED ISSUE #238: Support connecting to a running chromium instance
In certain cases, it would come handy to:
- launch a long-running chromium instance which is [detached from parent node process](https://nodejs.org/api/child_process.html#child_process_options_detached)
- connect to the instance later (from another script)

Launching long-running chromium instance might look like this:
```js
const browserTargetId = Browser.launchDetached({
  // standard Browser options
});
```

Connecting to the detached instance might be:
```js
const browser = await Browser.connect(browserTargetId);
...
```

To think about:
- stderr/stdout streams would not work for `Browser.connect` (unless those are forwarded to a fd?)
- users should do their own profile folder management

aslushnikov:
This patch:
- refactors Connection to use a single remote debugging URL instead of a
  pair of port and browserTargetId
- introduces Puppeteer.connect() method to attach to already running
  browser instance.

Fixes #238.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -136,9 +136,10 @@ puppeteer.launch().then(async browser => {
 });
 ```
 
-#### puppeteer.connect(remoteDebuggingURL[, ignoreHTTPSErrors])
-- `remoteDebuggingURL` <[string]> a remote debugging URL to connect to
-- `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
+#### puppeteer.connect(options)
+- `options` <[Object]>  Set of options to connect to the browser. Can have the following fields:
+  - `remoteDebuggingURL` <[string]> a remote debugging URL to connect to.
+  - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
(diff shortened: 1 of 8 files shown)


docs/api.md
```diff
@@ -134,6 +136,13 @@ puppeteer.launch().then(async browser => {
 });
 ```
 
+#### puppeteer.connect(remoteDebuggingURL[, ignoreHTTPSErrors])
+- `remoteDebuggingURL` <[string]> a remote debugging URL to connect to
+- `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
```
JoelEinbinder: I think this should be an options object. We will probably have more options in the future.

JoelEinbinder (approved):

docs/api.md
```diff
@@ -134,6 +136,14 @@ puppeteer.launch().then(async browser => {
 });
 ```
 
+#### puppeteer.connect(options)
+- `options` <[Object]>  Set of options to connect to the browser. Can have the following fields:
+  - `remoteDebuggingURL` <[string]> a remote debugging URL to connect to.
```
paulirish, line 141: this is a websocket URL right?

the term "remote debugging" isn't really used any longer. I'd go with something like `protocolWsEndpoint`

PR #266, opened 2017-08-15
Emit 'error' event when page crashes

### LINKED ISSUE #262: Puppeteer doesn't tell anything meaningful when the page crashes
In case of a chrome tab crash, the puppeteer doesn't report anything meaningful to the user.
E.g. if the crash happens during the `page.goto` call, then the call will hang until the timeout is hit.

aslushnikov:
This patch starts emitting 'error' event when page crashes.
'error' events have special treatment in node, so page crashes
become observable for users.

Fixes #262.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -14,6 +14,7 @@
   * [class: Page](#class-page)
     + [event: 'console'](#event-console)
     + [event: 'dialog'](#event-dialog)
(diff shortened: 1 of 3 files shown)


JoelEinbinder:
A bit confusing with error vs crash, but seems ok.

JoelEinbinder (approved):

PR #271, opened 2017-08-15
Roll chromium to r494365

alexkozy:
This patch:
- rolls chromium to r494365
- starts using Runtime.evaluate(awaitPromise: true), with new semantic
  we can avoid additional Runtime.awaitPromise call
- stops resolving promises for Console event

### DIFF
--- a/lib/ElementHandle.js
+++ b/lib/ElementHandle.js
@@ -51,7 +51,7 @@ class ElementHandle {
     stringifiedArgs.push(...args.map(x => JSON.stringify(x)));
     let functionDeclaration = `function() { return (${pageFunction})(${stringifiedArgs.join(',')}) }`;
     const objectId = this._remoteObject.objectId;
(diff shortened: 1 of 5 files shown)


paulirish:
What bug does this fix? Just wondering if we should also migrate Lighthouse to this new approach.

alexkozy:
I think lighthouse can be migrated as well. Recently we landed two changes:
- provided correct error message for rejected promises in exceptionDetails,
- aligned awaitPromise:true semantic with await expression semantic.

Second means that (with awaitPromise: true flag we always wrap your evaluation result with native Promise.resolve call):
- there is no "not a promise" protocol error any more,
- we support custom promise implementations like blackbird.js.

aslushnikov (approved):

paulirish:
@ak239 ok thanks. @brendankenny @patrickhulce

brendankenny:
> provided correct error message for rejected promises in exceptionDetails,

we do use `awaitPromise: true`, but this means we may be able to get rid of `Driver.wrapRuntimeEvalErrorInBrowser`. We'll have to investigate

PR #272, opened 2017-08-15
Rename 'remoteDebuggingURL' into 'protocolWSEndpoint'

aslushnikov:
This patch renames remoteDebuggingURL into protocolWSEndpoint

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -11,7 +11,7 @@
   * [class: Browser](#class-browser)
     + [browser.close()](#browserclose)
     + [browser.newPage()](#browsernewpage)
-    + [browser.remoteDebuggingURL()](#browserremotedebuggingurl)
+    + [browser.protocolWSEndpoint()](#browserprotocolwsendpoint)
     + [browser.version()](#browserversion)
   * [class: Page](#class-page)
     + [event: 'console'](#event-console)
--- a/docs/api.md
+++ b/docs/api.md
@@ -11,8 +11,8 @@
   * [class: Browser](#class-browser)
     + [browser.close()](#browserclose)
     + [browser.newPage()](#browsernewpage)
-    + [browser.protocolWSEndpoint()](#browserprotocolwsendpoint)
     + [browser.version()](#browserversion)
+    + [browser.wsEndpoint()](#browserwsendpoint)
   * [class: Page](#class-page)
     + [event: 'console'](#event-console)
     + [event: 'dialog'](#event-dialog)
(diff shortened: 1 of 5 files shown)


examples/block-images.js
```diff
@@ -18,7 +18,9 @@ const puppeteer = require('puppeteer');
 
 (async() => {
 
-const browser = await puppeteer.launch();
+const browser = await puppeteer.launch({
+  executablePath: '/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary'
```
ebidel: did you mean to add this?

lib/Browser.js
```diff
@@ -33,7 +33,7 @@ class Browser {
   /**
    * @return {string}
    */
-  remoteDebuggingURL() {
+  protocolWSEndpoint() {
```
ebidel: should this be a getter instead of a function?

docs/api.md
```diff
@@ -188,7 +188,7 @@ Closes browser with all the pages (if any were opened). The browser object itsel
 #### browser.newPage()
 - returns: <[Promise]<[Page]>> Promise which resolves to a new [Page] object.
 
-#### browser.remoteDebuggingURL()
+#### browser.protocolWSEndpoint()
 - returns: <[string]> A URL that could be used to start debugging this browser instance.
 
 Remote debugging url could be used as an argument to the [puppeteer.connect](#puppeteerconnect).
```
ebidel: Should we add the test.js example here? I think it would be be helpful to see context on why you'd need this.

JoelEinbinder (approved):

examples/block-images.js
paulirish: i don't think so...

lib/Browser.js
paulirish: IIRC its historically devtools style to use methods as the public API for instance properties. 

in puppeteer this is already the case with page.title(), page.url(), etc.

paulirish (approved):
lgtm from me % eric's comments.

lib/Browser.js
ebidel: Now might be a good time to start fresh. _Someone_ once told me using new language features was one of the goals of the project :)

I'll file a bug. Looks like there's enough inconsistency throughout the public API.

examples/block-images.js
aslushnikov: yeah, this is here by accident. Removed.

docs/api.md
aslushnikov: @JoelEinbinder updated docs to make it easier to understand.

The test.js is an artificial example, it doesn't illustrate the usecase. The real usecase is to launch browser in one script and connect to it later from the others. This would be quite a long example, and probably too specific.

aslushnikov:
(Renamed `protocolWSEndpoint` into `browserWSEndpoint` as discussed with @pavelfeldman)

lib/Browser.js
niieani: @paulirish historically that's been the case, but why not simplify with getters now, and start fresh as @ebidel mentioned? Please see my comment here regarding this "forced consistency" breaking some nice patterns: https://github.com/GoogleChrome/puppeteer/pull/1621#issuecomment-357914936

PR #283, opened 2017-08-16
Fix examples/block-images.js

aslushnikov:
This patch removes event listener after the screenshot is taken to avoid
a race condition with `browser.close()` call.

### DIFF
--- a/examples/block-images.js
+++ b/examples/block-images.js
@@ -21,14 +21,18 @@ const puppeteer = require('puppeteer');
 const browser = await puppeteer.launch();
 const page = await browser.newPage();
 await page.setRequestInterceptionEnabled(true);
-page.on('request', request => {
+
+function imageInterceptor(request) {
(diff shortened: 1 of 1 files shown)


JoelEinbinder:
It seems really bad that you have to remove your event listeners before calling `browser.close()`

aslushnikov:
Yeah, this is slightly over-pedantic.

The only alternative I can think of is to not throw when methods are called on a closed browser – which is inconsistent. Do you have any other suggestions?

aslushnikov:
Ah, wait, we shouldn't have delivered the event in the first place. I'll look into it

aslushnikov:
Investigation revealed:
- I was fighting the wrong race condition.
- You're right, there's no need to remove event listeners on `browser.close`.

The exception happens whenever the `request.continue()` is called, and before the response is received we close the browser. This makes us resolve all the in-flight messages with errors.

I don't have a good solution for this right now. Closing this CL since it's pointless.

PR #287, opened 2017-08-16
README: add --save-dev to npm installation

addyosmani:
The equivalent of `yarn add puppeteer` would probably be `npm i --save puppeteer` or  `npm i --save-dev puppeteer` as yarn will also [add](https://yarnpkg.com/lang/en/docs/cli/add/#toc-adding-dependencies) an entry for the package to your package.json file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Most things that you can do manually in the browser can be done using Puppeteer!
 To use Puppeteer in your project, run:
 ```
 yarn add puppeteer
-# or "npm i puppeteer"
+# or "npm i --save-dev puppeteer"
 ```
 
 > **Note**: When you install Puppeteer, it downloads a recent version of Chromium (~71Mb Mac, ~90Mb Linux, ~110Mb Win) that is guaranteed to work with the API. 


paulirish:
i think we'd have `--dev` on the yarn add to keep things consistent?

but regardless i'm not a fan of projects recommending themself as a dev dependency explicitly. Always felt awkward. Plus it looks like the first public project to ship depending on puppeteer uses it as a runtime dep: https://gitlab.com/garbee/gatherer-scraper/blob/master/package.json

anyways.. thank you tho. :)

Garbee:
Yea, it entirely depends on context. Say you're building the Scraper like in my case, you don't exactly expect others to require it (although they could if they wanted.) And since puppeteer is *required* to operate, it is a main dep.

If you're using it just for testing, then it would go under the dev dependencies. Since you wouldn't want the test requirements in production where they aren't used.

Just like anything here, it depends entirely on the context of the thing being developed where it should go in deps. IMO, leaving it as a quick `yarn add puppeteer` is best and not getting into the dev debate. If people know it should go into dev, they'll do it. Otherwise, let it be.

PR #292, opened 2017-08-16
Fix doc on pdf method. Minor addition to PDF path option docs.

Garbee:
Someone asked me whether or not they could get the PDF data without writing to a file. When checking the source to see, I saw the docblock was incorrect. So fixed that. Then added a quick note in the API docs about the path to save being optional. This may help people understand they can simply get just the returned buffer data for the PDF if desired.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -529,7 +529,7 @@ Page is guaranteed to have a main frame which persists during navigations.
 
 #### page.pdf(options)
 - `options` <[Object]> Options object which might have the following properties:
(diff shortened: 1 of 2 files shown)


RobinMalfait:
This is also the case for `screenshot`.

aslushnikov:
Nice! Can we have the same clarification for the `page.screenshot` method? These too are symmetrical

Garbee:
Yup, just looked into that and saw it is optional there as well. Updating now.

Garbee:
🎉 

CI failure on last run was due to some HTTPS thing. No way that could be due to the documentation changes.

aslushnikov (approved):
Thanks @Garbee!

PR #300, opened 2017-08-16
docs: FAQ on setting up debian server to run Chromium

Garbee:
Helps tackle #290 to some degree. Lists the dependencies in a simple to run command. Then shows to turn off the sandbox when running as root with a note to not do so as a normal user for security reasons.

And you can see the render [on my branch](https://github.com/Garbee/puppeteer/tree/faq-debian-server#q-chromium-wont-run-on-my-remote-server-how-do-i-make-it-work). It is using the `details` element to keep the dependency install list/command hidden away unless someone needs it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ yarn add puppeteer
 # or "npm i puppeteer"
 ```
 
-> **Note**: When you install Puppeteer, it downloads a recent version of Chromium (~71Mb Mac, ~90Mb Linux, ~110Mb Win) that is guaranteed to work with the API. 
+> **Note**: When you install Puppeteer, it downloads a recent version of Chromium (~71Mb Mac, ~90Mb Linux, ~110Mb Win) that is guaranteed to work with the API.
 
 ### Usage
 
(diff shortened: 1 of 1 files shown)


aslushnikov:
@Garbee how about adding a `TROUBLESHOOTING.md` that has questions and issue numbers as answers? This would scale pretty good for our needs. Like:


# Troubleshooting

## Q: How do I install on Debian?

Debian installation is discussed in https://github.com/GoogleChrome/puppeteer/issues/290

## Q: How do I ?..

Garbee:
@paulirish got me working on scripting this right now as a postinstall script structure. So what we'll provide is actual checks against installed packages to tell people what they need. As well as checking for sandbox support in the kernel and alerting people to that problem as well.

I'll have an initial PR for that shortly, just figuring out one problem with output before it is ready to be shown off.

Closing this up since it is not going to be nearly as useful as the live post install check.

foolip:
@Garbee were those post install checks added? Came here while trying to understand which dependencies I'd need to install to make Puppeteer launch in a minimal Docker image (with GitHub Actions).

Garbee:
Nope, the team decided to handle it in the `troubleshooting.md` doc. The content is there, not programmed to run automatically on install.

foolip:
Thanks, I see it here: https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md

PR #311, opened 2017-08-16
Initial post install check

Garbee:
This adds a post install script section for linux systems. Attempt to cover the issues from #290 on a more interactive level. Starting with just Debian-based distributions to hone the output and requirements. We can expand later to cover other package managers and distributions. Post install now does the following:

* Checks for linux OS
* If found, execute the main `postinstall` script.
* This script checks for sandbox availability then determines the OS-level script to run to look for dependencies
* The OS scripts have the list of required packages, then loops over them and looks for anything missing. All missing packages are then shown within a script to use to install them easily. So all a developer needs to do is copy and paste the command.
* Chromium is then downloaded if required

<details>
<summary>Example output</summary>

```
~/C/puppeteer (debian-post-install|●4✚2) $ yarn install
yarn install v0.27.5
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning "worker-loader@0.8.1" has unmet peer dependency "webpack@>=0.9 <2 || ^2.1.0-beta || ^2.2.0".
[4/4] Building fresh packages...
$ node install.js
Checking Debian requirements
-----------------------------------------------------
Run the following to get all of the required packages
-----------------------------------------------------
sudo apt install \
gconf-service \
libasound2 \
libatk1.0-0 \
libc6 \
libcairo2 \
libcups2 \
libdbus-1-3 \
libexpat1 \
libfontconfig1 \
libgcc1 \
libgconf-2-4 \
libgdk-pixbuf2.0-0 \
libglib2.0-0 \
libgtk-3-0 \
libnspr4 \
libpango-1.0-0 \
libpangocairo-1.0-0 \
libstdc++6 \
libx11-6 \
libx11-xcb1 \
libxcb1 \
libxcomposite1 \
libxcursor1 \
libxdamage1 \
libxext6 \
libxfixes3 \
libxi6 \
libxrandr2 \
libxrender1 \
libxss1 \
libxtst6 \
ca-certificates \
fonts-liberation \
libappindicator1 \
libnss3 \
lsb-release \
xdg-utils \
wget
-----------------------------------------------------
Downloading Chromium r494755 - 92.1 Mb [====================] 100% 0.0s 
Done in 41.51s.
~/C/puppeteer (debian-post-install|●4) $ 
```

</details>

### DIFF
--- a/install.js
+++ b/install.js
@@ -17,6 +17,15 @@
 const Downloader = require('./utils/ChromiumDownloader');
 const revision = require('./package').puppeteer.chromium_revision;
 const ProgressBar = require('progress');
+const os = require('os');
+const { execSync } = require('child_process');
+const path = require('path');
+const fs = require('fs');
+
+if (os.platform() === 'linux') {
+  let scriptPath = path.join(path.dirname(fs.realpathSync(__filename)), 'utils', 'linux', 'postinstall.sh');
+  process.stdout.write(execSync(scriptPath));
+}
 
 // Do nothing if the revision is already downloaded.
 if (Downloader.revisionInfo(Downloader.currentPlatform(), revision))
--- a/utils/linux/debian-check.sh
+++ b/utils/linux/debian-check.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+packages="gconf-service
+libasound2
+libatk1.0-0
+libc6
+libcairo2
+libcups2
+libdbus-1-3
+libexpat1
+libfontconfig1
+libgcc1
+libgconf-2-4
+libgdk-pixbuf2.0-0
+libglib2.0-0
+libgtk-3-0
+libnspr4
+libpango-1.0-0
+libpangocairo-1.0-0
+libstdc++6
+libx11-6
+libx11-xcb1
+libxcb1
+libxcomposite1
+libxcursor1
+libxdamage1
+libxext6
+libxfixes3
+libxi6
+libxrandr2
+libxrender1
+libxss1
+libxtst6
+ca-certificates
+fonts-liberation
+libappindicator1
+libnss3
+lsb-release
+xdg-utils
+wget"
+
+declare -a neededPackages
+
+for packageName in $packages; do
+  if ! dpkg-query -l "$packageName" > /dev/null 2>&1; then
+    neededPackages[${#neededPackages[@]}]="$packageName"
+  fi
+done
+
+
+neededCount=${#neededPackages[@]}
+
+if [[ $neededCount -gt 0 ]]; then
+  echo "-----------------------------------------------------"
+  echo "Run the following to get all of the required packages"
+  echo "-----------------------------------------------------"
+  echo "sudo apt install \\"
+  for i in "${neededPackages[@]}"; do
+    output="$i"
+
+     if [[ ${neededPackages[@]: -1 } != "$i" ]]; then
+        output+=" \\"
+     fi
+
+    echo "$output"
+  done
+  echo "-----------------------------------------------------"
+fi
+
+exit 0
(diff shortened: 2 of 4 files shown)


aslushnikov:
Thanks @Garbee, that's impressives!

I'm skeptical this approach will work for us in the long run. Over time, distributions will evolve, chrome will change, and users might encounter more and more specific usecases. This script will grow over time, accumulating all kinds of quirks that exist in the world.

More importantly, the script is not tested somehow, so we won't know when it gets outdated. We might end up with a misleading script.

I'd rather go with something like https://github.com/GoogleChrome/puppeteer/pull/315

giodamelio:
Maybe print a message in the postinstall on Linux guiding people to the docs?

Garbee:
> distributions will evolve

This is *Linux* and *Bash*... These basic things, even package names, don't change. The dependencies may get updated from time-to-time, but those are simple to run through and do an update of the list to make it accurate again.

> users might encounter more and more specific usecases

Yes, this is where reactivity is needed when people file bugs about certain OSes. In the case of Debian, this should hit like every Debian system just fine. Since they all inherit from Debian as a core. It is when I look to RPM based distros that things get tricky. But, it's still completely doable in a long-term reliable way.

> More importantly, the script is not tested somehow, so we won't know when it gets outdated. We might end up with a misleading script.

Testing happens by anyone using Linux. They'll open bugs which we can assess. Any script issues here shouldn't stop installation and operation of the package itself. It wouldn't really end up as "misleading" since things don't change that dramatically with Linux.

> I'm skeptical this approach will work for us in the long run.

Skeptical is very good. But, I think the gains of this kind of integration are vastly worth it compared to the approach #315 is taking. In this scripting case, we are being proactive about informing people at install-time what issues they may face and how to solve them. Whereas #315 is a chain-link "please read" situation. As humans are lazy, most won't read or will mis-understand long discussion threads trying to find the right comments in a long list of them. Which leads to more questions and discussion instead of direct actions to resolve the problem.

Scripting to check the environment is the most user-focused path forward. We should at least try it first, then if it proves too unstable we can fallback to a well designed troubleshooting document to help people out.

Garbee:
The namespace check here should be removed. Cloud providers can make this in-accurate by modifying the boot process. Such as Linode. While the distro-default will reside in `/boot` it will actually load the kernel config from somewhere else on bootstrapping.

So while we can have a baseline, that bit is difficult to make generally accurate since any random hosting provider can screw with it.

aslushnikov:
The question here is whether the puppeteer should handle the installation issues as a part of the project codebase.

- If the answer is `yes`, we should go ahead and evolve your script into something that actually installs all these dependencies. We should have CI that makes sure the installation process works on different distributions, and we should run our tests under those distributions.
- if the answer is `no`, we'll have people coming over asking how to use puppeteer on different platforms. We should enable them to find answers easily.

Puppeteer core competency is browser automation, we're not good at dealing with images/platforms/libraries. Yes, we bundle chromium, but that's because we land patches to chromium and want to benefit from them in puppeteer right away. 

I think we should focus on what we're doing good, and let someone else to come up with a `headless-installer` project. It would automatically bring environment up-to-date with chrome headless needs, apt-getting all the missing dependencies, or it could be a carefully maintained docker image. Either way, we'd be happy to point to it from our docs.

Garbee:
We're providing a project that runs Chromium. People expect it to help them understand how to run it as well. Not need to go find yet-another-package to handle those details.

The script can't be made to install deps automatically within a hook script. So it would need to be an extra command people would need to run to install the dependencies for their system. Which is completely doable, but not very proactive. Unless I built the scripts so with a flag they simply alert and when called without it they'll get asked for install authentication needs.

As we are seeing right now with all the support requests, people expect us to help them. This provides a great pro-active way to help them out. So the answer as to whether it should handle the requirements is `yes`. That's the best user experience to provide for people. At least until we can not require a bundled version and tell people to install the Chrome binary distribution.

aslushnikov:
Closing in favor of https://github.com/GoogleChrome/puppeteer/pull/428

PR #314, opened 2017-08-16
Cookies

### LINKED ISSUE #53: Add Cookies API
There should be an API to access cookies.

JoelEinbinder:
Closes #53

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -342,7 +342,7 @@ If there's no element matching `selector`, the method throws an error.
 Gets the full HTML contents of the page, including the doctype.
 
 #### page.cookies(...urls)
-- `...urls` <[string]>
+- `...urls` <...[string]>
 - returns: <[Promise]<[Array]<[Object]>>>
   - `name` <[string]>
   - `value` <[string]>
@@ -357,7 +357,7 @@ If no URLs are specified, this method returns cookies related to the current pag
 If URLs are specified, only cookies for those URLs are returned.
 
 #### page.deleteCookie(...cookies)
-- `...cookies` <[Object]>
+- `...cookies` <...[Object]>
   - `name` <[string]> **required**
   - `url` <[string]>
   - `domain` <[string]>
@@ -680,7 +680,7 @@ Shortcut for [`keyboard.down`](#keyboarddownkey-options) and [`keyboard.up`](#ke
 - returns: <[Promise]>
 
 #### page.setCookie(...cookies)
-- `...cookies` <[Object]>
+- `...cookies` <...[Object]>
   - `name` <[string]> **required**
   - `value` <[string]> **required**
   - `url` <[string]>
--- a/docs/api.md
+++ b/docs/api.md
@@ -353,7 +353,7 @@ Gets the full HTML contents of the page, including the doctype.
   - `secure` <[boolean]>
   - `sameSite` <[string]> `"Strict"` or `"Lax"`.
 
-If no URLs are specified, this method returns cookies related to the current page.
+If no URLs are specified, this method returns cookies for the current page URL.
 If URLs are specified, only cookies for those URLs are returned.
 
 #### page.deleteCookie(...cookies)
--- a/docs/api.md
+++ b/docs/api.md
@@ -34,6 +34,8 @@
     + [page.click(selector[, options])](#pageclickselector-options)
     + [page.close()](#pageclose)
     + [page.content()](#pagecontent)
+    + [page.cookies(...urls)](#pagecookiesurls)
+    + [page.deleteCookie(...cookies)](#pagedeletecookiecookies)
     + [page.emulate(options)](#pageemulateoptions)
     + [page.emulateMedia(mediaType)](#pageemulatemediamediatype)
     + [page.evaluate(pageFunction, ...args)](#pageevaluatepagefunction-args)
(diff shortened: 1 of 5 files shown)


docs/api.md
```diff
@@ -315,6 +318,30 @@ Shortcut for [page.mainFrame().click(selector[, options])](#frameclickselector-o
...
+  - `expires` <[number]> Unix time in seconds.
+  - `httpOnly` <[boolean]>
+  - `secure` <[boolean]>
+  - `sameSite` <[string]> `"Strict"` or `"Lax"`.
+
+If no URLs are specified, this method returns cookies related to the current page.
+If URLs are specified, only cookies for those URLs are returned.
+
+#### page.deleteCookie(...cookies)
+- `...cookies` <[Object]>
```
aslushnikov: here and below: <[Array]<[Object]>>

JoelEinbinder: I don't like putting array when we have a spread operator. The user doesn't pass in an array.

JoelEinbinder:
An update to the cookie protocol is landing in r495262

https://chromium-review.googlesource.com/c/617750
also https://chromium-review.googlesource.com/c/614721 landed recently.

docs/api.md
```diff
@@ -315,6 +318,30 @@ Shortcut for [page.mainFrame().click(selector[, options])](#frameclickselector-o
...
+  - `name` <[string]>
+  - `value` <[string]>
+  - `domain` <[string]>
+  - `path` <[string]>
+  - `expires` <[number]> Unix time in seconds.
+  - `httpOnly` <[boolean]>
+  - `secure` <[boolean]>
+  - `sameSite` <[string]> `"Strict"` or `"Lax"`.
+
+If no URLs are specified, this method returns cookies related to the current page.
```
pavelfeldman: I don't think you should inherit this poor DTP semantics. There is no real 'related to the page', there are cookies matching domains, paths and names. I would suggest to add optional name, domain and path parameters that would match the cookies. Your underlying implementation would be fetching everything using `Network.getAllCookies`.

docs/api.md
```diff
@@ -315,6 +318,30 @@ Shortcut for [page.mainFrame().click(selector[, options])](#frameclickselector-o
...
+If URLs are specified, only cookies for those URLs are returned.
+
+#### page.deleteCookie(...cookies)
+- `...cookies` <...[Object]>
+  - `name` <[string]> **required**
+  - `url` <[string]>
```
pavelfeldman, line 362: Either url or domain should be required for scoping. Also, this method will delete multiple cookies for each pattern, so might be worth calling method `deleteCookies` and parameters `patterns`

docs/api.md
```diff
@@ -611,6 +638,19 @@ Shortcut for [`keyboard.down`](#keyboarddownkey-options) and [`keyboard.up`](#ke
 - `html` <[string]> HTML markup to assign to the page.
 - returns: <[Promise]>
 
+#### page.setCookie(...cookies)
+- `...cookies` <...[Object]>
+  - `name` <[string]> **required**
+  - `value` <[string]> **required**
+  - `url` <[string]>
```
pavelfeldman, line 686: either url or domain required.

JoelEinbinder: It uses the page url if none is specified.

docs/api.md
JoelEinbinder: I'd agree if the method was browser.setCookie. But page.setCookie should act like I set a cookie for the page.

pavelfeldman: I don't think you know what `Network.getCookies` without `urls` does...

docs/api.md
pavelfeldman: Cookies are not associated with the page, they are associated with the document (frame).

docs/api.md
JoelEinbinder: Ok now I don't think I do either.

JoelEinbinder: I want to be able to do
```js
page.setCookie({name, value});
page.deleteCookie({name, value});
```

JoelEinbinder: Disregard that, I read this comment as being on setCookie.

pavelfeldman: (I think your understanding of 'related to the page' is different from the protocol one. I'm fine with your understanding, but it needs to be clarified and implementation needs to be changed. Also, you might want to pull this into the frame and alias in page)

MiniCodeMonkey (commented):
I'm super excited for this feature!

lib/Page.js
```diff
@@ -155,6 +155,47 @@ class Page extends EventEmitter {
...
+      urls: urls.length ? urls : [this.url()]
+    })).cookies;
+  }
+
+  /**
+   * @param {Array<Network.CookieParam>} cookies
+   */
+  async deleteCookie(...cookies) {
+    for (const cookie of cookies) {
+      const item = {};
```
aslushnikov:
```js
const item = Object.assign({}, cookie);
```

test/test.js
```diff
@@ -1920,6 +1920,162 @@ describe('Page', function() {
...
+        value: '1'
+      }, {
+        name: 'cookie2',
+        value: '2'
+      }, {
+        name: 'cookie3',
+        value: '3'
+      });
+      expect(await page.evaluate('document.cookie')).toBe('cookie1=1; cookie2=2; cookie3=3');
+      page.deleteCookie({name: 'cookie2'});
```
aslushnikov: technically, you'd want to `await` it here

lib/Page.js
```diff
@@ -155,6 +155,47 @@ class Page extends EventEmitter {
...
+      Object.assign(item, cookie);
+      const pageURL = this.url();
```
aslushnikov: nit: bring pageURL out of the loop

aslushnikov (approved):
lgtm % nits

PR #315, opened 2017-08-17
Add troubleshooting.md

aslushnikov:


Garbee (changes requested):
Linking off to a big issue thread isn't extremely useful to users. People don't like to read and there is a lot of back and forth within the thread that isn't relevant.

If we decide the install scripting isn't a valid solution for whatever reason, then let's make a well-formed troubleshooting document that provides all the details people need within it rather than linking to a vague resource.

aslushnikov:
@Garbee Linking to the issue thread is not ideal, agreed. Rather then having the instructions in troubleshooting.md, how about writing a well-formed comment in the issue summarizing all the findings, and linking to it from troubleshooting.md? This will point people to a place where they can ask a question, which is important since we're chasing a moving target here.

aslushnikov:
Closing this to cleanup repository branches.

PR #319, opened 2017-08-17
Indent the code in examples

bogas04:

PR #327, opened 2017-08-17
Fix code formatting in readme

freeall:


aslushnikov:
I've just merged #319 which does the same.

Thank you for the contribution though!

PR #332, opened 2017-08-17
Adapt docs regarding headless flag

sericaia:
I just read the [docs](https://developers.google.com/web/updates/2017/04/headless-chrome):

```js
@param {boolean=} headless True (default) launches Chrome in headless mode.
```

It makes more sense to me to have `headless: true` here as `headless: false` may confuse people

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,10 +76,10 @@ See [`Page.pdf()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/ap
 
 **1. Uses Headless mode**
 
-Puppeteer launches Chromium in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). To launch a full version of Chromium, set the ['headless' option](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions) when creating a browser:
+Puppeteer launches Chromium in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). To launch a full version of Chromium, set the ['headless' option](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions) when launching a browser:
 
 ```js
-const browser = await puppeteer.launch({headless: true});
+const browser = await puppeteer.launch({headless: false}); // default is true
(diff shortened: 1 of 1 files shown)


Garbee:
So this is the *example* code in the README. Associated with the following excerpt:

> Puppeteer launches Chromium in headless mode. To launch a full version of Chromium, set the 'headless' option when creating a browser:

So, this example code is instructing people how to run it with the UX so they can see what is going on. Which is *not* the default behavior which is a true flag for headless mode.

Rather than change the example which then makes it not align with the context of usage. Could a simple comment be added at the end of the line `// default headless mode is 'true'`?

README.md
```diff
@@ -79,7 +79,7 @@ See [`Page.pdf()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/ap
 Puppeteer launches Chromium in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). To launch a full version of Chromium, set the ['headless' option](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions) when creating a browser:
```
ebidel: Think we can help clarify with:

`set the ['headless' option](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions) when creating a browser`
->
`use the ['headless: false'](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions) option when launching a browser:`

README.md
```diff
@@ -79,7 +79,7 @@ See [`Page.pdf()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/ap
...
 
 ```js
-const browser = await puppeteer.launch({headless: false});
+const browser = await puppeteer.launch({headless: true});
```
ebidel: `const browser = await puppeteer.launch({headless: false}); // default is true`

ebidel (changes requested):

sericaia:
I agree and did the suggested changes, and CLA was signed.

ebidel (approved):

PR #333, opened 2017-08-17
Use `download` to download and extract Chromium

### LINKED ISSUE #336: What about proxy usage?
Is it possible to setup proxy for request or for user_profile

kevva:
Since `download` is extracting the files in memory we don't have to write the zip to disk before extracting. Also, this makes the code slightly prettier.

Since it handles proxies it should fix https://github.com/GoogleChrome/puppeteer/issues/336.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   "license": "SEE LICENSE IN LICENSE",
   "dependencies": {
     "debug": "^2.6.8",
-    "extract-zip": "^1.6.5",
+    "download": "^6.2.5",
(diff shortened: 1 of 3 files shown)


aslushnikov:
Hey @kevva, thank you for the PR!

The `download` package seems to be quite heavy on dependencies (brings in 65 packages), whereas extract-zip has only 6. It's a high priority for us to keep as lean as possible on dependencies, so we'd rather not add it unless it's absolutely critical.

PR #337, opened 2017-08-17
support downloading through corporate proxies

tardyp:
https module does not support proxy.

We use the request module, which has sam'ish API, and support proxy via standard 'https_proxy' environment variable


aslushnikov:
Hey @tardyp, thank you for the PR!

The `request` module is quite heavy on dependencies, it adds in 47 packages. We'd like to keep our dependencies as lean as possible, so I'd rather have proxy support implemented manually atop of node's `https` module.

tardyp:
Indeed, I did not realize the hell of dependencies on top of ``request``. I would however avoid to reinvent the wheel at maintaining proxy handling code inside this repo. 
So we have to find the right minimalist npm module, which does proxy and streaming.

tardyp:
quick search shows that there is a global-tunnel package which is more minialist, but it looks unmaintain.

request has 4M downloads per week, so I would expect most people already have those deps is their yarn cache.

related link:
https://github.com/nodejs/node/issues/8381
https://github.com/nodejs/node/issues/1490

interesting comment for me is:

> Speaking from prior experience where proxy support was added to an existing product, you start > with HTTP CONNECT but you end up supporting everything from SOCKSv5 to HTTP POST, > basic/digest/and-so-on auth, Kerberos, NTLMv1 and v2, custom CA chains, client certificates, certificate fingerprinting, etc., etc.

That's why there is a ``request`` package which handles this for you.

aslushnikov:
@tardyp thank you for the links, those were really interesting to read.

Here are my findings: 
- According to [this issue](https://github.com/request/request/issues/141), socks proxy is not supported by [request/request](https://github.com/request/request). So I'd assume we're dealing specifically with HTTP proxy here.
- HTTP proxying seems to be doable with node's http module  and https://github.com/Rob--W/proxy-from-env to parse env variables (check out [Rob's example](https://github.com/Rob--W/proxy-from-env#example)).

Let me know if you're interested in implementing this approach. If not, I'd rather go with a graceful failure that would explain how to download needed chromium revision manually.

kevva:
Could also use [`caw`](https://github.com/kevva/caw), which does most of the job automatically. But then again, it also brings some dependencies, although very small ones.

aslushnikov:
Closing in favor of #455

PR #348, opened 2017-08-17
Move uploadFiles onto ElementHandle

JoelEinbinder:
This removes `page.uploadFile` and `frame.uploadFile`


aslushnikov:
We can't land this until 1.0 since it's a breaking change. Let's mirror it on element instead

aslushnikov (approved):

PR #318, opened 2017-08-17
Add Serializable type into the docs

JoelEinbinder:
This patch converts all of our spread operators to be documented as <...Value>. It adds a Serializable type with links to the documentation on JSON.stringify.

PR #349, opened 2017-08-17
Return the success value of page.waitFor

JoelEinbinder:
I will update api.md in this PR after #318 lands.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -332,6 +332,12 @@ describe('Page', function() {
       expect(error).toBeTruthy();
       expect(error.message).toContain('Cannot poll with non-positive interval');
     }));
+    it('should return the success value', SX(async function() {
+      const watchdog = page.waitForFunction(() => window.__FOO);
+      let success = await page.evaluate(() => window.__FOO = 'hit');
+      await watchdog;
+      expect(success).toBe('hit');
+    }));
   });
 
   describe('Frame.waitForSelector', function() {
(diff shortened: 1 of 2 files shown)


aslushnikov:
I'm not sure I follow. Why is this needed?

JoelEinbinder:
`waitForSelector` should return an ElementHandle. `waitForFunction` should return an ElementHandle if the function resolves to an Element. For simplicity and consistency, `waitForFunction` should just return its success value.

aslushnikov:
But in this case, the success value will always be `true`?

JoelEinbinder:
It is always truthy, but it isn't necessarily `true`. Look at the test.

test/test.js
```diff
@@ -332,6 +332,12 @@ describe('Page', function() {
       expect(error).toBeTruthy();
       expect(error.message).toContain('Cannot poll with non-positive interval');
     }));
+    it('should return the success value', SX(async function() {
+      const watchdog = page.waitForFunction(() => window.__FOO);
+      let success = await page.evaluate(() => window.__FOO = 'hit');
+      await watchdog;
+      expect(success).toBe('hit');
```
aslushnikov, line 339: This probably should be
```js
expect(await watchdog).toBe('hit')
```

JoelEinbinder: Yep. Good catch.

aslushnikov:
I see now. Prior to landing this, let's discuss our element handle plan with @pavelfeldman / @dgozman so that we're all on the same page.

aslushnikov:
Do you want this in now or you'll wait for #382 to land first?

aslushnikov:
Closing this in favor of object handles.

PR #351, opened 2017-08-17
Fix flaky test: waitForSelector

JoelEinbinder:
The test was not awaiting the promise before checking if it was resolved, causing a race.


aslushnikov (approved):
Nice!

aslushnikov (approved):

PR #352, opened 2017-08-17
readme: start of useful debugging tips

ebidel:
plan to add more as we collect feedback

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,10 +113,11 @@ Explore the [API documentation](docs/api.md) and [examples](https://github.com/G
     ```
 
 1. Slow it down - the `slowMo` option slows down Puppeteer operations by the
-   specified amount of milliseconds. Another way to help see what's going on:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -102,6 +102,25 @@ Puppeteer creates its own Chromium user profile which it **cleans up on every ru
...
+    ```js
+    const browser = await puppeteer.launch({headless: false});
+    ```
+
+1. Slow it down - the `slowMo` option slows down Puppeteer operations by the
+   specified amount of milliseconds. Another way to help see what's going on:
+
+    ```js
+    const browser = await puppeteer.launch({
+      slowMo: 250 // slow down by 250ms
```
aslushnikov: this probably makes sense only with `headless: false`

aslushnikov (approved):

README.md
ebidel: done

PR #354, opened 2017-08-17
Allow capturing transparent screenshots

pavelfeldman:


lib/Page.js
```diff
@@ -467,7 +467,11 @@ class Page extends EventEmitter {
       await this._client.send('Emulation.setDeviceMetricsOverride', { mobile, width, height, deviceScaleFactor, screenOrientation });
     }
 
+    if (options.noBackground)
+      await this._client.send('Emulation.setDefaultBackgroundColorOverride', { color: { r: 0, g: 0, b: 0, a: 0 } });
     let result = await this._client.send('Page.captureScreenshot', { format, quality: options.quality, clip });
+    if (options.noBackground)
+      await this._client.send('Emulation.setDefaultBackgroundColorOverride', { color: { r: 255, g: 255, b: 255, a: 1 } });
```
JoelEinbinder: Sending no color should go back to the default.

docs/api.md
```diff
@@ -605,6 +605,7 @@ Shortcut for [`keyboard.down`](#keyboarddownkey-options) and [`keyboard.up`](#ke
         - `y` <[number]> y-coordinate of top-left corner of clip area
         - `width` <[number]> width of clipping area
         - `height` <[number]> height of clipping area
+    - `noBackground` <[boolean]> When true, hides default white background and allows capturing screenshots with transparency.
```
JoelEinbinder: Hides default white background and allows capturing screenshots with transparency. Defaults to `false`

examples/screenshot.js
```diff
@@ -23,6 +23,9 @@ const page = await browser.newPage();
 await page.goto('http://example.com');
 await page.screenshot({path: 'example.png'});
 
+await page.setContent('TRANSPARENT BACKGROUND');
+await page.screenshot({path: 'transparent.png', noBackground: true});
```
JoelEinbinder: I am not sure if this warrants a place in the example.

docs/api.md
pavelfeldman: Done

examples/screenshot.js
pavelfeldman: Done.

aslushnikov (approved):

docs/api.md
```diff
@@ -605,6 +605,7 @@ Shortcut for [`keyboard.down`](#keyboarddownkey-options) and [`keyboard.up`](#ke
         - `y` <[number]> y-coordinate of top-left corner of clip area
         - `width` <[number]> width of clipping area
         - `height` <[number]> height of clipping area
+    - `noBackground` <[boolean]> Hides default white background and allows capturing screenshots with transparency. Defaults to `false`.
```
aslushnikov: How about calling this `background` which defaults to `true`, or maybe `transparentBackground` which defaults to false? The double negation in `noBackground: false` feels weird.

pavelfeldman: I prefer parameters that default to `false`, hence the name. `transparentBackground` works for me as well, but I don't think it is particularly better. The parameter controls behavior that is somewhat involved, so it is hard to find a simple name for it. How can background be transparent? Is it when there is noBackground? 😝

PR #355, opened 2017-08-17
Mute audio in headless

JoelEinbinder:
Audio probably shouldn't play in headless. It has no ears or mouth.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -58,7 +58,8 @@ class Launcher {
       chromeArguments.push(
           `--headless`,
           `--disable-gpu`,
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):

tommoor:
Oh man, this just cost me an entire day of trying to figure out why Chrome isn't outputting audio in headless mode :(

**Edit:** For anyone else that comes across this and wants to run Chrome headless WITH audio output you'll want to do something like the following to avoid the mute flag:

```javascript
    headless: false,
    args: [
      '--headless',
      '--disable-gpu'
    ]
  });
```

PR #365, opened 2017-08-18
Determine failed navigation based on Frame.unreachedURL

aslushnikov:
This patch starts using frame.unreachedURL property from
frameNavigated event to determine if the page navigation
was successful.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -130,6 +130,14 @@ class FrameManager extends EventEmitter {
     this._frames.delete(frame._id);
     this.emit(FrameManager.Events.FrameDetached, frame);
   }
+
+  /**
+   * @param {!Frame} frame
+   * @return {boolean}
+   */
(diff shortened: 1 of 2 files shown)


lib/FrameManager.js
```diff
@@ -130,6 +130,14 @@ class FrameManager extends EventEmitter {
     this._frames.delete(frame._id);
     this.emit(FrameManager.Events.FrameDetached, frame);
   }
+
+  /**
+   * @param {!Frame} frame
+   * @return {boolean}
+   */
+  isMainFrameLoadingFailed() {
```
JoelEinbinder, line 138: didMainFrameFailToLoad

JoelEinbinder (approved):

lib/FrameManager.js
aslushnikov: I don't see much difference between the names (and I like the former more since it resembles protocol event names..)

PR #368, opened 2017-08-18
Support anchor navigation

### LINKED ISSUE #257: Hash-only navigation doesn't work
Since the hash-only navigation doesn't cause any network requests and doesn't cause load event, 
the following gets stuck:

```js
(async() => {
  let browser = await puppeteer.launch();
  await page.goto('https://example.com#ohh'); // <== stuck here
  browser.close();
})();

```

aslushnikov:
This patch adds support for anchor navigation for the following methods:
- `page.goto`
- `page.waitForNavigation`
- `page.goForward` / `page.goBack`

Fixes #257

### DIFF
--- a/lib/NavigatorWatcher.js
+++ b/lib/NavigatorWatcher.js
@@ -68,7 +68,18 @@ class NavigatorWatcher {
       navigationPromises.push(networkIdle);
     }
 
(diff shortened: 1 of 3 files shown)


aslushnikov:
Nah, this requires more work. Closing this for now.

PR #369, opened 2017-08-18
readme: tone of logo surface from light source

kensoh:
This is a minor cosmetic update on Puppeteer's logo. The rectangle surface at lower left corner seems to have the wrong shade, base on the scene's light source. Below are the original and the updated ones. Of course, if the designer intentionally wants to make the design asymmetric then this is irrelevant.


ebidel (approved):
Proper light sources are key! Thanks for the update.

paulirish:
:+1:

PR #371, opened 2017-08-18
feat(typings): add TypeScript declaration file

Brooooooklyn:


Brooooooklyn:
no idea why `phantomjs-test` is down, maybe you should retry it `without cache`

JoelEinbinder:
This is interesting. I have a patch that lints puppeteer's JSDoc with TypeScript #219. We are waiting on some features to land in stable TypeScript before switching to it. Hopefully it will allow us to avoid having a hand-written types file.

aslushnikov:
Closing in favor of #219

Brooooooklyn:
this pull request is very different between https://github.com/GoogleChrome/puppeteer/pull/219. It would help everybody who write `TypeScript` with  puppeteer, and give intelligence ablity to them. And the purpose of this pr is not to `lint the sourcecodes`, but to help consumers to avoid type error.

screenshot from 2017-08-19 09-51-01
screenshot from 2017-08-19 09-52-32

rohmanhm:
I agreed with @Brooooooklyn.

Brooooooklyn:
@aslushnikov 
so could you consider about reopen this pull request ? I think maintain the `type declaration` in original repo is a better way rather than maintain it in https://github.com/DefinitelyTyped/DefinitelyTyped

rohmanhm:
@Brooooooklyn I put your typings into a gist https://gist.github.com/rohmanhm/5bf4be5e969250eb6fde558117f4e742
It's helped me a lot. Thanks

index.d.ts
```diff
@@ -0,0 +1,464 @@
...
+  clip?: {
+    x?: number
+    y?: number
+    width?: number
+    height?: number
+  }
+}
+
+export interface Browser {
+  close: void
```
rohmanhm, line 135: This should be a function `close()` not `close`.

aslushnikov:
@Brooooooklyn as a by-product of #219, autocomplete will start working out-of-the box in VSCode.

Brooooooklyn:
uh... typings would give more ability rather than autocomplete, such as smarter `Type inference` and more powerful `type constraint`

for example, if there is `index.d.ts` in this project, `TypeScript` users would have no way to pass `String` to `Page.on` expect it is `string which is defined` in https://github.com/Brooooooklyn/puppeteer/blob/feature/add-typings/index.d.ts#L143

rohmanhm (commented):
Better change to function

JoelEinbinder:
@Brooooooklyn I think I can be a bit fancy with our types and still get your event typing in #219. It should be the same result as far as autocomplete and typescript users are concerned, but it will also allow Puppeteer to lint itself and give us nice autocomplete as we are working on the project.

Lxxyx:
This is awesome. And VSCode JavaScript User can also have good code intellisense feature.

You can see this. [Automatic Type Acquisition](https://code.visualstudio.com/docs/languages/javascript#_automatic-type-acquisition)

image

mohsen1:
I used this type definition file and got a bunch of type mismatches already! Generating types from code makes the most sense

PR #383, opened 2017-08-18
Introduce page.emulateMedia method

### LINKED ISSUE #312: Add an option to generate PDF with the screen CSS media type
It seems the PDF generation activates the print CSS media type. While that makes sense in the majority of use cases, it would be very useful in some other use cases, like [Decktape](https://github.com/astefanutti/decktape) that exports HTML presentations for which it is preferable to retain the screen media type.

aslushnikov:
Emulating `screen` media would force `page.pdf()` to print with `screen`
media.

Fixes #312.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -549,7 +549,7 @@ Page is guaranteed to have a main frame which persists during navigations.
     - `left` <[string]> Left margin, accepts values labeled with units.
 - returns: <[Promise]<[Buffer]>> Promise which resolves with PDF buffer.
 
-`page.pdf()` generates a pdf of the page with `print` css media. To get a pdf with `screen` media, use [page.emulateMedia(mediaType)](#pageemulatemediamediatype).
+`page.pdf()` generates a pdf of the page with `print` css media. To generate a pdf with `screen` media, call [page.emulateMedia('screen')](#pageemulatemediamediatype) before calling `page.pdf()`.
 
 The `width`, `height`, and `margin` options accept values labeled with units. Unlabeled values are treated as pixels.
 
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -544,6 +549,8 @@ Page is guaranteed to have a main frame which persists during navigations.
     - `left` <[string]> Left margin, accepts values labeled with units.
 - returns: <[Promise]<[Buffer]>> Promise which resolves with PDF buffer.
 
+`page.pdf()` generates a pdf of the page with `print` css media. To get a pdf with `screen` media, use [page.emulateMedia(mediaType)](#pageemulatemediamediatype).
```
ebidel: To get a pdf -> To generate a pdf

use page.emulateMedia -> call page.emulateMedia('screen') before calling page.pdf().

ebidel: Alternatively, just show the example:
```js
// Generates a PDF with 'screen' media type.
await page.emulateMedia('screen');
await page.pdf({path: 'page.pdf'});
```

aslushnikov: Did everything!

ebidel: Love it! Thanks.

docs/api.md
```diff
@@ -348,6 +349,10 @@ puppeteer.launch().then(async browser => {
 
 List of all available devices is available in the source code: [DeviceDescriptors.js](https://github.com/GoogleChrome/puppeteer/blob/master/DeviceDescriptors.js).
 
+#### page.emulateMedia(mediaType)
+  - `mediaType` <[string]> Forces a CSS media on page. The only allowed values are `screen`, `media` and `null`. Passing `null` disables media emulation.
```
JoelEinbinder: Pass nothing to disable.

Changes the media type.

JoelEinbinder (approved):

PR #388, opened 2017-08-19
Add support for node v6 (fixes #316)

### LINKED ISSUE #316: Node 6 support
Per the [package.json file](https://github.com/GoogleChrome/puppeteer/blob/master/package.json), it looks like puppeteer only works on Node versions `>=7.10.0`. Would you be willing to maintain this package on the Node 6 release line, since it is the current LTS version?

mgcrea:
Attempt to support NodeJS6 support, so far it looks like to be working (trying locally by linking puppeteer globally on machine with my NodeJS v6.11 ).

Would be great if other with working suites could check that everything is fine.

Had to move `./utils` into `./src/utils` so it can be easily compiled along the source. Since it is required at some point by the lib (in `launcher.js`), I guess it also makes sense to ship it along the lib as a subfolder.


JoelEinbinder:
Doing this is more complicated than just adding the babel plugins because we do some crazy `toString` magic on our functions. I have been working on a potential solution.

aslushnikov:
Thanks @mgcrea for the PR

Indeed, babel is not going to work out of the box since we have parts of code evaluated in  browser, and babel knows nothing about it. @JoelEinbinder did a bunch of prototypes and is currently working on a solution which should address this.

mgcrea:
Indeed it was a bit naive! Many test do fail, but it somehow does work properly for simple screenshots (my use case). I'll use [my fork](https://github.com/mgcrea/puppeteer) waiting for an official solution, great to know this is being worked on, will track #316 for an update. Thanks.

PR #400, opened 2017-08-19
doc: fix some links in api.md

vsemozhetbyt:
There is one more case I do not know how to fix:

In the [`response.text()` description](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#responsetext), type of `return` value misses the reference. Please, let me know if I can add one for the `[text]` link.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -194,7 +194,7 @@ Closes browser with all the pages (if any were opened). The browser object itsel
 - returns: <[string]> Browser websocket url.
 
 Browser websocket endpoint which could be used as an argument to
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):

aslushnikov:
> In the response.text() description, type of return value misses the reference. Please, let me know if I can add one for the [text] link.

ah, thanks. It is supposed to be Promise of string. Fixed in https://github.com/GoogleChrome/puppeteer/commit/ab125f6a698898312d98753afa7fced699b702eb

PR #401, opened 2017-08-19
example 'block images' - make extension check case insensitive

loilo:
Just because images may also have uppercase file extensions.


loilo:
Signed CLA.

aslushnikov (approved):

PR #410, opened 2017-08-20
doc: change ambiguous phrase in api.md

vsemozhetbyt:
I was a bit puzzled about what was passed in and where, but I am not a native speaker and may not be reliable here.

Please, close if the wording was OK.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1129,7 +1129,7 @@ The `elementHandle.dispose` method stops referencing the element handle.
 - returns: <[Promise]<[Serializable]>> Promise which resolves to function return value
 
 If the function, passed to the `elementHandle.evaluate`, returns a [Promise], then `elementHandle.evaluate` would wait for the promise to resolve and return it's value.
-The function will be passed in the element ifself as a first argument.
+The element will be passed as the first argument to `pageFunction`, followed by any `args`.
 
 #### elementHandle.hover()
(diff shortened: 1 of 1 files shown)


aslushnikov:
@ebidel Eric, what would you say? :)

docs/api.md
```diff
@@ -1129,7 +1129,7 @@ The `elementHandle.dispose` method stops referencing the element handle.
 - returns: <[Promise]<[Serializable]>> Promise which resolves to function return value
 
 If the function, passed to the `elementHandle.evaluate`, returns a [Promise], then `elementHandle.evaluate` would wait for the promise to resolve and return it's value.
-The function will be passed in the element ifself as a first argument.
+The element ifself will be passed in the function as a first argument.
```
ebidel: I think it was intended to read as: "The element gets passed to `pageFunction` as a first argument."

However, is this true? Doesn't `pageFunction` get passed `args`, not the element handle? Maybe leftover from a previous time?

aslushnikov: Yes, this is still true: the element itself is passed as a first argument, followed by all the arguments. (note: this is `elementHandle.evaluate`, not `page.evaluate`)

ebidel: Ah ok. Then how about: "The element will be passed as the first argument to `pageFunction`, followed by any `args`."

vsemozhetbyt: Done.

aslushnikov (approved):

Issue #382, opened 2017-08-18
Implement object handles


pavelfeldman:
- Add `page.object(func)` that returns a new `ObjectHandle` type.
- Expose `ObjectHandle.json()` that would return result similar to present `page.evaluate`. That will allow access to the raw object handlers for non-trivial operations.
- Inherit `ElementHandle` from `ObjectHandle` if possible
- Allow object handles to be disposed, expose `evaluateOn` on the objects
- Provide support for object groups (arenas) for bulk dispose

aslushnikov:
The `ObjectHandle` may have a `objectHandle.property('foo')` method that returns value of the property named `foo`. This will address #438.

gajus:
> The ObjectHandle may have a objectHandle.property('foo') method that returns value of the property named foo. This will address #438.

How does one obtain the property value from the JSHandle?

```js
const totalCostElement = await page.$('.review-payment__cost-breakdown strong');

const totalCostElementTextContent = await totalCostElement.getProperty('textContent');

```

This returns `JSHandle` object.

gajus:
Turns out you simply need to use `toString`.

This API is rather odd. Whats the reason for JSHandle abstraction in the first place?

pavelfeldman:
@aslushnikov let's remove toString before it is misused broadly. I don't remember reviewing it btw, I remember renaming old `toString` into `jsonString`, but that was it...

gajus:
> @aslushnikov let's remove toString before it is misused broadly. I don't remember reviewing it btw, I remember renaming old toString into jsonString, but that was it...

Whats the correct way to get the property value?

aslushnikov:
@gajus it's [`JSHandle.jsonValue`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#jshandlejsonvalue)

gajus:
> Returns a JSON representation of the object. The JSON is generated by running JSON.stringify on the object in page and consequent JSON.parse in puppeteer.

This description makes it in no way clear that `jsonValue` returns the value of a property.

aslushnikov:
@gajus `jsonValue` returns a value of the object.

If you want a value of particular object property, you'd need to use both `getProperty()` and `jsonValue()`:

```js
const navigatorHandle = await page.evaluateHandle(() => window.navigator);
const uaHandle = await navigatorHandle.getProperty('navigator');
const ua = await uaHandle.jsonValue();
```

Alternatively, you can use the evaluate method:

[...]

PR #416, opened 2017-08-20
Allow custom function in $ selector

kevin940726:
#### Feature Request

I am working on introducing `puppeteer` to our team recently. The problem is that currently the `page.$()` selector only support css selector, but our stack includes `react` and `css-in-js` which make css selecting next to impossible to perform. I think it will be nice to allow custom function in `page.$()` selector and open up the gate of integrating with multiple frameworks and use cases.

Lets say we have a HTML page looks like
```html
<ul>
  <li>1</li>
  <li>2</li>
</ul>
```

We can select the second `<li>` element with a custom function called `selectSecond`
```js
const selectSecond = selector => document.querySelectorAll(selector)[1];
const secondElement = await page.$(selectSecond, 'ul > li'); // <li>2</li>
```

_The simple example use case above doesn't need custom function at all and can be done with simple css selector, it's just to demonstrate the API._

The implementation is backward-compatible and the API style is similar to `evaluate`, but it still only allows selecting one element since I don't want to over-complicate this PR.

Thanks for the great work! I really enjoy `puppeteer` and definitely will work on it more in the future.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -26,7 +26,7 @@
     + [event: 'requestfailed'](#event-requestfailed)
     + [event: 'requestfinished'](#event-requestfinished)
     + [event: 'response'](#event-response)
-    + [page.$(selector)](#pageselector)
+    + [page.$(selector, ...args)](#pageselector-args)
     + [page.addScriptTag(url)](#pageaddscripttagurl)
     + [page.click(selector[, options])](#pageclickselector-options)
     + [page.close()](#pageclose)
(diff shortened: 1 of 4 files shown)


505studly:
Seriously 

Thank You
Aaron Perkins
Paving Foreman
Armour Pavement Inc
(505) 803-8141

aslushnikov:
Thanks @kevin940726 for the pull request!

We know about this problem and were considering this approach. However, we figured that implementing #382 will cover this and other important usecases we're interested in.

Closing in favor of #382

kevin940726:
@aslushnikov Totally! Thanks for your great effort! Any idea when #382 will be implemented? I mean that will it be in the next major release?

PR #419, opened 2017-08-20
Add .content() method. Fixes #406.

### LINKED ISSUE #406: Request: page.html()
Return the html of the current page. Is there an easy way to do this I'm not seeing?

jeresig:
This adds a new method: `.content()` which is the counterpart of `.setContent()`, retrieving the HTML contents of the page, including the doctype (if there is one).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -30,6 +30,7 @@
     + [page.addScriptTag(url)](#pageaddscripttagurl)
     + [page.click(selector[, options])](#pageclickselector-options)
     + [page.close()](#pageclose)
+    + [page.content()](#pagecontent)
     + [page.emulate(options)](#pageemulateoptions)
     + [page.emulateMedia(mediaType)](#pageemulatemediamediatype)
     + [page.evaluate(pageFunction, ...args)](#pageevaluatepagefunction-args)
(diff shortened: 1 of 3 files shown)


jeresig:
I signed it!

vsemozhetbyt:
Does [`XMLSerializer` API](https://developer.mozilla.org/en-US/docs/Web/API/XMLSerializer) have any restrictions to not use it instead?

vsemozhetbyt:
See also: https://github.com/GoogleChrome/puppeteer/issues/406#issuecomment-323599588

RobertCorey:
Thanks so much! This has been my first time interacting with an open source library and it's been really great being supported.

jeresig:
@vsemozhetbyt Oh, I'm not aware of any! I could give that a try for just the doctype, simplifying the logic a bit. I wanted to avoid using it for the full document as it then became an XML document (which can be pretty different from what was originally intended). I'll update the PR to serialize the doctype.

docs/api.md
```diff
@@ -462,6 +463,11 @@ Shortcut for [page.mainFrame().focus(selector)](#framefocusselector).
 #### page.frames()
 - returns: <[Array]<[Frame]>> An array of all frames attached to the page.
 
+#### page.getContent()
```
aslushnikov: let's go with `page.content()`, we don't have "get" prefixes in other API calls (e.g. [`page.plainText()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageplaintext)).

jeresig: @aslushnikov thank you for the review! I've just updated it to be `.content()` instead of `.getContent()`.

aslushnikov (approved):
Thanks for the PR!

jeresig:
You're very welcome! Thank you for the review!

PR #420, opened 2017-08-20
Remove redundant flag from example.

benjamingr:
Already the default


vsemozhetbyt:
[doc](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagegotourl-options) states:

>  defaults to `load`

aslushnikov:
'networkidle' is not the default. Why remove it?

PR #428, opened 2017-08-21
Add troubleshooting.md

aslushnikov:
This patch adds troubleshooting.md and starts referring to it when chromium
fails to start.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -1,12 +1,12 @@
 # Troubleshooting
 
-## Chrome headless doesn't start
+## Chrome headless doesn't start on Debian (e.g. Ubuntu)
 
 - Make sure all the necessary dependencies are installed
 - The broad discussion on the topic could be found in [#290](https://github.com/GoogleChrome/puppeteer/issues/290)
 
 <details>
-<summary>Debian Dependencies (e.g. Ubuntu)</summary>
+<summary>Debian Dependencies</summary>
 
 ```
 gconf-service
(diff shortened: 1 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -0,0 +1,68 @@
+# Troubleshooting
+
+## Chrome headless doesn't start
+
+- Make sure all the necessary dependencies are installed
+- The broad discussion on the topic could be found in [#290](https://github.com/GoogleChrome/puppeteer/issues/290)
```
ebidel: could -> can

docs/troubleshooting.md
```diff
@@ -0,0 +1,68 @@
+# Troubleshooting
+
+## Chrome headless doesn't start
```
ebidel: "...doesn't start on Debian"?

docs/troubleshooting.md
```diff
@@ -0,0 +1,68 @@
...
+xdg-utils
+wget
+```
+</details>
+
+## Chrome Headless fails due to sandbox issues
+
+- make sure kernel version is up-to-date
+- read about linux sandbox here: https://chromium.googlesource.com/chromium/src/+/master/docs/linux_suid_sandbox_development.md
+- try running without sandbox (**Note: running without sandbox is not recommended due to security reasons!**)
```
ebidel: without sandbox -> without the sandbox

docs/troubleshooting.md
```diff
@@ -0,0 +1,68 @@
...
+```js
+const puppeteer = require('puppeteer');
+(async() => {
+  const browser = await puppeteer.launch({args: ['--no-sandbox', '--disable-setuid-sandbox']});
```
ebidel: no need for the full example, just show the launch line.

ebidel (approved):

docs/troubleshooting.md
aslushnikov: Done.

docs/troubleshooting.md
aslushnikov: Done.

docs/troubleshooting.md
aslushnikov: Done.

PR #429, opened 2017-08-21
Add Page.setElementContent

anefzaoui:
Pretty simple I think but quite handy for interacting with a page's specific content.
r?


aslushnikov:
Thank you for the PR!

We want to keep the `page` interface as lean as possible, so this should go to the `elementHandle` class.

And once you have an element handle, it's really easy to set a content:
```js
elementHandle.evaluate(e => e.innerHTML = "<div>test!</div>");
```

will it work for you?

aslushnikov:
@anefzaoui any thoughts on this?

PR #430, opened 2017-08-21
Fail gracefully when chromium failed to download

aslushnikov:
This patch changes both install.js and Launcher.js to inform how
chromium could be downloaded manually if puppeteer fails to download it.

### DIFF
--- a/install.js
+++ b/install.js
@@ -15,18 +15,30 @@
  */
 
 const Downloader = require('./utils/ChromiumDownloader');
+const platform = Downloader.currentPlatform();
 const revision = require('./package').puppeteer.chromium_revision;
 const ProgressBar = require('progress');
 
+const revisionInfo = Downloader.revisionInfo(platform, revision);
 // Do nothing if the revision is already downloaded.
-if (Downloader.revisionInfo(Downloader.currentPlatform(), revision))
+if (revisionInfo.downloaded)
   return;
 
 let allRevisions = Downloader.downloadedRevisions();
-Downloader.downloadRevision(Downloader.currentPlatform(), revision, onProgress)
+Downloader.downloadRevision(platform, revision, onProgress)
     // Remove previous chromium revisions.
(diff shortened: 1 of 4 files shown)


install.js
```diff
@@ -15,18 +15,30 @@
...
-    .catch(error => console.error('Download failed: ' + error.message));
+    .catch(onError);
+
+function onError(error) {
+  console.error(`ERROR: Failed to download chromium r${revision}!
+- Download chromium manually:
+    ${revisionInfo.url}
+- Extract chromium into ${revisionInfo.folderPath}
+  * Chromium executable should be at ${revisionInfo.executablePath}
+- Alternatively, point puppeteer to the downloaded chromium with the 'executablePath' option:
```
JoelEinbinder: Remove this

lib/Launcher.js
```diff
@@ -72,7 +72,7 @@ class Launcher {
     if (typeof chromeExecutable !== 'string') {
       let chromiumRevision = require('../package.json').puppeteer.chromium_revision;
       let revisionInfo = Downloader.revisionInfo(Downloader.currentPlatform(), chromiumRevision);
-      console.assert(revisionInfo, 'Chromium revision is not downloaded. Run npm install');
+      console.assert(revisionInfo.downloaded, `Chromium revision is not downloaded. Run "npm install", or download manually from ${revisionInfo.url} and extract into ${revisionInfo.folderPath}.`);
```
JoelEinbinder: Let's leave this how it was. If npm install fails, it will tell them. And if chromium didn't download, then other dependencies are probably missing.

JoelEinbinder (approved):

install.js
aslushnikov: Done.

lib/Launcher.js
aslushnikov: Done.

PR #439, opened 2017-08-21
Put chrome profiles into a temporary directory

JoelEinbinder:
I noticed I had a lot of old .dev-profile folders from when I used Ctrl+C to end Puppeteer. I thought about cleaning up on SIGINT, but that would break if users have their own SIGINT handler. So maybe putting the profiles into a temp directory will at least stop them from piling up forever.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -24,7 +24,7 @@ const crypto = require('crypto');
 const helper = require('./helper');
 const ChromiumRevision = require('../package.json').puppeteer.chromium_revision;
 
-const CHROME_PROFILE_PATH = path.resolve(__dirname, '..', '.dev_profile');
+const CHROME_PROFILE_PATH = path.join(os.tmpdir(), 'puppeteer_dev_profile');
 let browserId = 0;
 
 const DEFAULT_ARGS = [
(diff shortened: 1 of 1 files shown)


lib/Launcher.js
```diff
@@ -22,7 +22,7 @@ const Browser = require('./Browser');
 const readline = require('readline');
 const crypto = require('crypto');
 
-const CHROME_PROFILE_PATH = path.resolve(__dirname, '..', '.dev_profile');
+const CHROME_PROFILE_PATH = path.join(os.tmpdir(), 'puppeteer_dev_profile');
```
aslushnikov: does this work on windows? For some reason, chrome launcher has a bunch of heuristics to infer tmp directory

aslushnikov (commented):
The SIGINT is handled in https://github.com/GoogleChrome/puppeteer/pull/434.

> but that would break if users have their own SIGINT handler.

Can you provide an example?

lib/Launcher.js
aslushnikov: It also turned out there's a `fs.mkdtemp` - let's use it in favor of current folder name generation.

aslushnikov (approved):

Issue #433, opened 2017-08-21
No way to associate Frame object with the <iframe> element it represents


twalpole:
While attempting to convert Poltergeist from using PhantomJS to Puppeteer I need to select a `Frame` object (returned from `childFrames`) based on the `<iframe>` element which encapsulates it.  Looking through the Puppeteer code I don't see any link between them, or way to generate a link.  Am I missing something, or is this not possible with the functionality provided by the Dev Tools Protocol?

pavelfeldman:
Which way do you need to go?
(a) You traverse the DOM and given the `<iframe>` node you want to evaluate in its context?
(b) You listen to frame added event and would like to get an `<iframe>` node that contains the frame?
(c) both?

twalpole:
@pavelfeldman I need the first (a)

aslushnikov:
For the record: the upstream patch [has landed](https://chromium.googlesource.com/chromium/src/+/b2949a65405c5548c378e4f382f59e00fee9d5ef). This will work out nicely as #382 is done.

Madd0g:
I'm interested in this as well, with regards to the script injection/evaluation capabilities.

Right now there's `evaluateOnNewDocument` and `exposeFunction`, both work on all frames(?), but is there a way of knowing which frame it's working on?

aslushnikov:
> Right now there's evaluateOnNewDocument and exposeFunction, both work on all frames(?), but is there a way of knowing which frame it's working on?

#382 with object handles, you'll be able to pass in the owner document while calling `exposeFunction`.

elisherer:
@aslushnikov, do you have any progress on this matter? 
And if not, would you like me to assist in any way? 
As I understand, the upstream work have been done and is available on the current chromium version?

EDIT: I was able to construct a naive implementation of finding the containing iframe in the parent frame by doing the following (inside ElementHandle.js):
```js
  /**
   * @return {!Promise<ElementHandle>}
   */
  async ownerFrame() {
    if (!this._frame._parentFrame)
      throw new Error('No parent frame');

    const rootElementHandle = await this._frame.$('*');
    const rootElementDescriptionHandle = await this._client.send('DOM.describeNode', { objectId: rootElementHandle._remoteObject.objectId });

    const parentsIframes = await this._frame._parentFrame.$$('iframe');
    if (!parentsIframes.length)
      throw new Error('No iframe elements found in parent');

    return parentsIframes.find(async parentsIframe => {
[...]

PR #445, opened 2017-08-21
Add get bounding box to ElementHandle for easier element screenshot

elisherer:
~~This will add the ability to capture a screenshot based on a selector (as opposed to using the clip object).~~
~~The implementation is on top of the clip option to make things simple:~~

~~await page.screenshot({~~
  ~~clip: '.box:nth-of-type(43)'~~
~~});~~

~~*The example above is from the unit test (see output image below in the affected files).~~

EDIT: Changed to adding a method: `elementHandle.boundingBox()` which will return the rectangle needed for screenshot's clip option.


ebidel:
FWIW, I posted [a pr](https://github.com/GoogleChrome/puppeteer/pull/452) that shows how to do this in user-code. It's just a few lines but gets even smaller after https://github.com/GoogleChrome/puppeteer/issues/382 lands.

That said, this might be common enough that it's something we should support directly in the API.

aslushnikov:
That's nice. Let's make `elementHandle.screenshot(options)` method instead to keep `Page` lean.

elisherer:
The challenge is that all of the screenshot code is written on the `Page` class (together with a `screenshotTaskQueue`). 
Also, `ElementHandle` does not have a reference to the creating page (maybe it should?)
Is `elementHandle.screenshot(page, options)` acceptable?

elisherer:
I have tried to resolve the cla check by adding an alternative email of mine...

aslushnikov:
@elisherer ah, I see. Let's do the follwing:
- we'll introduce the `elementHandle.ownerFrame()` shortly (there's [some work upstream](https://chromium-review.googlesource.com/c/chromium/src/+/624954) to make it possible)
- then you'll be able to add a `frame.page()` and use `page.screenshot` to drive element screenshots.

elisherer:
@aslushnikov, how about we put the screenshot taking logic into one common place.
I noticed the page is only important for using the fullPage option (usint its viewport), which elements' screenshots don't need.
Also, clipping should not be possible for elements. 
So each of the cases have different "predefined" clip providing options.
In addition, the screenshot task queue is stored on the browser and passed to each of the pages. In the same way it could be passed to each element as a handle for screenshot capturing.
WDYT?

aslushnikov:
@elisherer this approach would work. However, I don't like it because it touches quite a bit of code with no good excuse.

I see the `elementHandle.screenshot` as a convenience method for `page.screenshot`: the feature doesn't add any new capabilities and It's easy to polyfill with current puppeteer API:
```js
// Polyfill for elementHandle.screenshot
async function screenshotElement(page, selector, options)  {
  const handle = await page.$(selector);
  const box = await getElementCoordinates(handle); // roughly return element's bounding box
  options.clip = box;
  return await page.screenshot(options);
}
```
Convenience methods should be implemented atop of core, with as little footprint as possible. This will minimize the maintenance cost.

> In addition, the screenshot task queue is stored on the browser and passed to each of the pages. In the same way it could be passed to each element as a handle for screenshot capturing.

The `screenshotTaskQueue` is a shabby detail of screenshot machinery we have today, it's unfortunate to have it spreading further into subsystems.

test/test.js
```diff
@@ -1116,6 +1116,16 @@ describe('Page', function() {
     }));
   });
 
+  describe('ElementHandle.boundingBox', function() {
+    it('should work', SX(async function() {
+      await page.setViewport({width: 500, height: 500});
+      await page.goto(PREFIX + '/grid.html');
+      const fourtySecondBoxElement = await page.$('.box:nth-of-type(42)');
```
JoelEinbinder: Nit: fortySecondBoxElement

lib/ElementHandle.js
```diff
@@ -103,6 +103,19 @@ class ElementHandle {
     const objectId = this._remoteObject.objectId;
     return this._client.send('DOM.setFileInputFiles', { objectId, files });
   }
+
+  /**
+   * @return {!Object}
+   */
+  async boundingBox() {
+    const boxModel = await this._client.send('DOM.getBoxModel', { objectId: this._remoteObject.objectId });
```
JoelEinbinder: The bounding box of the element could be partially off-screen. The element might also not be scrolled into view. something to consider when taking screenshots. If you look above at `_visibleCenter`, we use `scrollIntoViewIfNeeded` and clip based on the window bounds. Because we are already in JavaScript, we use getBoundingClientRect and avoid introducing `DOM.getBoxModel`.

I don't expect `element.boundingBox()` to scroll the element into view, but I can imagine that it would be intended if boundingBox is used to take screenshots.

test/test.js
elisherer: Needs explaining here... You don't like the long variable name?

lib/ElementHandle.js
elisherer: Would you recommend documenting this consideration?

aslushnikov: Values in `DOM.getBoxModel` are affected by the `box-sizing` css property. AFAIR they also include page scale (@wwwillchen might remember better).

`getBoundingClientRect` might work better for this use case.

> Would you recommend documenting this consideration?

Once this lands, https://github.com/GoogleChrome/puppeteer/pull/452 should work as a good-enough documentation.

test/test.js
```diff
@@ -1116,6 +1116,16 @@ describe('Page', function() {
     }));
   });
 
+  describe('ElementHandle.boundingBox', function() {
+    fit('should work', SX(async function() {
```
aslushnikov: let's do `it` instead of `fit`.

lib/ElementHandle.js
```diff
@@ -103,6 +103,26 @@ class ElementHandle {
...
+        return null;
+      const rect = element.getBoundingClientRect();
+      return {
+        x: rect.left,
+        y: rect.top,
+        width: rect.right - rect.left,
+        height: rect.bottom - rect.top
+      };
+    });
+    if (!box)
```
aslushnikov: let's return nullable `box` without throwing

lib/ElementHandle.js
```diff
@@ -73,7 +73,7 @@ class ElementHandle {
       };
     });
     if (!center)
-      throw new Error('No node found for selector: ' + selector);
+      throw new Error('No node found');
```
aslushnikov: Nice catch!

Could you please change this into a more elaborate message while we're here:
```js
  throw new Error('Element is detached from DOM');
```

lib/ElementHandle.js
```diff
@@ -103,6 +103,26 @@ class ElementHandle {
...
+   */
+  async boundingBox() {
+    const box = await this.evaluate(element => {
+      if (!element.ownerDocument.contains(element))
+        return null;
+      const rect = element.getBoundingClientRect();
+      return {
+        x: rect.left,
+        y: rect.top,
+        width: rect.right - rect.left,
```
aslushnikov: Is there any reason why you don't use the `rect.width` and `rect.height` here?

lib/ElementHandle.js
```diff
@@ -103,6 +103,26 @@ class ElementHandle {
...
   }
+
+  /**
+   * @return {!Promise<Object>}
+   */
+  async boundingBox() {
+    const box = await this.evaluate(element => {
+      if (!element.ownerDocument.contains(element))
+        return null;
+      const rect = element.getBoundingClientRect();
```
aslushnikov: This returns bounding box in viewport, whereas `screenshot` accepts coordinates relative to the main frame. To make this usable with `page.screenshot`, you need to convert coordinates to the window.

That's how we do this in DevTools front-end: [DOMExtension.js](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/devtools/front_end/dom_extension/DOMExtension.js?type=cs&q=AnchorBox+f:front_end&sq=package:chromium&l=494)

Let's add a test that verifies screenshots with clipping, something along the lines:
```js
describe('Page.screenshot', function() {
  ...
  it('should work with elementHandle.boundingBox', SX(async function() {
    await page.goto(PREFIX + '/grid.html');
    await page.evaluate(() => window.scrollBy(0, 100));
    const box = await page.$('div:nth-child(3)');
    const screenshot = await page.screenshot({clip: await box.boundingBox()});
    expect(screenshot).toBeGolden('screenshot-element-bounding-box.png');
  }));
  ...
```

docs/api.md
```diff
@@ -1117,6 +1118,29 @@ puppeteer.launch().then(async browser => {
...
 
+
+#### elementHandle.boundingBox()
+- returns: <[Object]>
+    - x <[number]> the x coordinate of the element in pixels.
+    - y <[number]> the y coordinate of the element in pixels.
+    - width <[number]> the width of the element in pixels.
+    - height <[number]> the height of the element in pixels.
+
+This method returns the bounding box of the element (relative to the main frame).
```
aslushnikov: (relative to the main frame), or `null` if element is detached from dom.

lib/ElementHandle.js
elisherer: I've added the test above, but I ended up using the code from [document-offset](https://github.com/timoxley/offset) for getting the absolute offset.

aslushnikov: Thanks for adding a test!

> but I ended up using the code from document-offset for getting the absolute offset.

In case of nested iframes, this will return the bounding box inside the iframe's document, not the main frame document. (can we have a test for this too?)

elisherer: Since we are running JavaScript we are are bound to CORS rules, accessing parent frame and getting its position might not be possible (it even happens when ran from the file system).
So the DevTools approach might be better to accomplish this. I'm trying to get this to run:
```js
fit('should handle nested frames', SX(async function() {
      await page.goto(PREFIX + '/frames/nested-frames.html');
      const nestedFrame = page.frames()[0].childFrames()[1];
      const elementHandle = await nestedFrame.$('div');
      const box = await elementHandle.boundingBox();
      expect(box).toEqual({ x: 28, y: 253, width: 284, height: 18 });
    }));
```
Currently errors with:
```
  Message:
    Expected $.x = 8 to equal 28.
    Expected $.y = 8 to equal 253.
```

That being said, I'm guessing that `click` and `hover` won't work with iframes either...

elisherer: UPDATE: I tried again with DOM.getBoxModel and the content quad returned is the position of the box relative to the main frame. I will update once both tests are running (scroll on main frame and nested in iframe)

lib/ElementHandle.js
```diff
@@ -103,6 +103,22 @@ class ElementHandle {
     const objectId = this._remoteObject.objectId;
     return this._client.send('DOM.setFileInputFiles', { objectId, files });
   }
+
+  /**
+   * @return {!Promise<Object>}
+   */
+  async boundingBox() {
+    const layoutMetrics = await this._client.send('Page.getLayoutMetrics');
+    const boxModel = await this._client.send('DOM.getBoxModel', { objectId: this._remoteObject.objectId });
```
aslushnikov: Let's not use `DOM.getBoxModel` - it misbehaves with page zoom. It's also an experimental protocol method (and we try to minimize our use of experimental methods so that it would be easier to stabilize them in future).

The logic here is implementable with DOM API, e.g. [like this](https://cs.chromium.org/chromium/src/third_party/WebKit/Source/devtools/front_end/dom_extension/DOMExtension.js?q=DOMExtension&sq=package:chromium&l=494). Do you have any concerns about this approach?

elisherer: Yes, as you pointed out, we need to consider elements inside iframes and this logic you suggest applies only to the frame containing the element. Using a while loop to get the parent frame for offsets might not work because of CORS policy which makes the method inconsistant.
On the other hand using the DevTools API gives us the ability to inspect the window without browser JavaScript constraints.
So it's a matter of deciding if we want to support nested element bounding boxes or not (or we do but only for same domain frames).

aslushnikov: > Using a while loop to get the parent frame for offsets might not work because of CORS policy which makes the method inconsistant.

Yes, you're right.

We're thinking about introducing the `elementHandle.ownerFrame()` method in #433. This would allow you to iterate frame structure avoiding CORS and use bullet-proof `element.getBoundingClientRect` instead of `DOM.getBoxModel`.

If this solves your concerns, I'd like to wait for the `ownerFrame` to land first and then follow-up with this PR.

elisherer: OK, will do.

paulirish:
@elisherer any update? looks like a few people eager for this.

aslushnikov:
@paulirish this is stuck on my implementation of `elementHandle.ownerFrame()` method (#433).

elisherer:
@aslushnikov, It now supports iframes by using upwards traversal using the DOM api (describeNode) and getting the position inside the window using JavaScript.
(I used my own implementation of "ownerFrame", hope it's ok)

elisherer:
why is the cla/google "waiting for status to be reported" is waiting for a couple of days? shouldn't it be instant?

aslushnikov:
> @aslushnikov, It now supports iframes by using upwards traversal using the DOM api (describeNode) and getting the position inside the window using JavaScript.

Thanks @elisherer. I've though about this a bit more and doing multiple hops into page to resolve bounding box seems to be fragile - page might've changed during the process. This would be terrible to debug.

I now tend to agree with you on the matter of using `DOM.getBoxModel` method. I'll look into it shortly to see how it could be fixed.

> why is the cla/google "waiting for status to be reported" is waiting for a couple of days? shouldn't it be instant?

No worries, we can ignore the bot. I believe it's not very reliable.

aslushnikov:
@elisherer: @JoelEinbinder looked into that and it turned out that:
- `getBoxModel` is wrong with non-default page's scale
- we consider page's scale to be always default in pptr

To sum up: we can use `getBoxModel` in pptr. Please, go ahead with this patch!
I'm sorry for making you wait for so long.

Also, @JoelEinbinder has a PR to fix clicking in iframes: https://github.com/GoogleChrome/puppeteer/pull/971
It would be great if you can base your work atop of that one.

elisherer:
@aslushnikov, OK, I'm done bringing back the code of getBoxModel while merging to the new changes in ElementHandle.
I did base my code on @JoelEinbinder PR #971, so it kind of already fixes the problem there with visible center (because I made visible center call boundingBox; no need for 2 separate identical calls).

lib/ElementHandle.js
```diff
@@ -111,6 +100,34 @@ class ElementHandle extends JSHandle {
...
+  /**
+   * @return {!Promise<Object>}
+   */
+  async boundingBox() {
+    const error = await this.executionContext().evaluate(element => {
+      if (!element.ownerDocument.contains(element))
+        return 'Node is detached from document';
+      if (element.nodeType !== HTMLElement.ELEMENT_NODE)
+        return 'Node is not of type HTMLElement';
+      element.scrollIntoViewIfNeeded();
```
aslushnikov: I wouldn't expect `element.boundingBox()` to cause side effects like scrolling page. Can we drop it from here?

elisherer: Isn't it the same side effects as keyboard and touchscreen operations (no expecatations there either)?
Maybe it should be optional?
i.e.
```
element.boundingBox({ scrollIntoView: true })
```
(it will return a different bounding box if scroll was needed)

aslushnikov: > Isn't it the same side effects as keyboard and touchscreen operations (no expecatations there either)?

This only happens for mouse/touch operations. The reason is that we can't click to the off-screen location, it is (somewhat) understandable that element should be on screen to be clicked.

Whereas The `boundingBox` method is a getter, and getters are generally perceived as operations without side-effects.

> (it will return a different bounding box if scroll was needed)

Thanks, I see why you need it. I'd make an internal method `_innerBoundingBox(scrollIntoView)`, that would be used in both `_visibleCenter` and `boundingBox()`:

```js
  async _visibleCenter() {
    const box = this._innerBoundingBox(true /* scrollIntoView */);
    // .. compute center
  }
  async boundingBox() {
    return this._innerBoundingBox(false /* scrollIntoView */);
  }
```

elisherer: I agree about the above, but how do you suggest solve the problem of "off-frame" objects needing to be captured by screenshot.
I would want the behavior of scrolling before calling `boundingBox()`.
`element.scrollIntoViewIfNeeded()` like in `focus()`?

elisherer:
@aslushnikov , as my last commit states, I added a scrollIntoViewIfNeeded method and used is as part of the visualCenter, and in the unit test of taking a screenshot with a framed scrolled, I called it just before calling boundingBox to ensure the box is visible. looks good by me.

aslushnikov:
> as my last commit states, I added a scrollIntoViewIfNeeded method and used is as part of the visualCenter, and in the unit test of taking a screenshot with a framed scrolled, I called it just before calling boundingBox to ensure the box is visible. looks good by me.

@elisherer `element.scrollInfoViewIfNeeded` would not be used for anything other then screenshots. In this case, we can just have the `elementHandle.screenshot`. Conveniently, the `ElementHandle` now has a reference to the page: https://github.com/GoogleChrome/puppeteer/blob/f1aa18af4e713abfb170e1f9f1d3a3d378b32987/lib/ElementHandle.js#L31

Even with `elementHandle.screenshot`, we should keep the `ElementHandle.boundingBox`: it comes handy in multiple places.

aslushnikov (approved):
Great work, thank you.

zhongjiewang:
if the elementhandle in a frame,  screenshot method , does it work ?

PR #447, opened 2017-08-21
cleanup: Use ES6 default params

ebidel:
Also some let -> const action.

### DIFF
--- a/lib/Input.js
+++ b/lib/Input.js
@@ -138,9 +138,7 @@ class Mouse {
   /**
    * @param {!Object=} options
    */
-  async down(options) {
-    if (!options)
-      options = {};
+  async down(options = {}) {
     this._button = (options.button || 'left');
     await this._client.send('Input.dispatchMouseEvent', {
       type: 'mousePressed',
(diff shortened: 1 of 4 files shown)


lib/Input.js
```diff
@@ -127,20 +127,18 @@ class Mouse {
    * @param {number} y
    * @param {!Object=} options
    */
-  async click(x, y, options) {
+  async click(x, y, options = {}) {
     this.move(x, y);
     this.down(options);
-    if (options && options.delay)
+    if (typeof options.delay === 'number')
```
JoelEinbinder, line 133: This is a slight change in functionality, if someone explicitly passes in 0. I'm not sure what the right thing to do is here, but I just thought I'd point it out.

JoelEinbinder (approved):

lib/Input.js
ebidel: Ack. Kinda pointless to pass `0` but I figured we should support it. I can change it back if you want.

ebidel:
rebased!

PR #449, opened 2017-08-21
let -> const. Add eslint rule to enforce

ebidel:
Using `let` where `const` is more appropriate will now lint as an error. The files where corrected using `eslint --fix`

IMO, we should do this but it's up to you all.


aslushnikov (approved):
Let's try!

Issue #448, opened 2017-08-21
`node install.js` fails inside enterprise proxy


phawxby:
```
> puppeteer@0.9.0 install C:\Users\Z269CJBB\Documents\GitHub\xerox-framework\node_modules\puppeteer
> node install.js

Download failed: connect ETIMEDOUT 216.58.213.112:443
```

Looking at the docs it appears that `https` does not natively make use of proxy environmental variables, as a result it fails during the install phase. This appears to be the culprit.
https://github.com/GoogleChrome/puppeteer/blob/master/utils/ChromiumDownloader.js#L187

jsnanigans:
are you sure you have node 7.10 or higher ?

phawxby:
Node 8

Garbee:
Pr #455 aims to address this

PR #451, opened 2017-08-21
Fix for proxy unaware https

phawxby:
Fix for #448

My knowledge on the internals of node modules isn't great so I'm expecting feedback!


phawxby:
I signed it!

phawxby:
It appears I'm not very good at reading, I didn't see #337. I'll look at improving the approach

PR #452, opened 2017-08-21
examples: add screenshot dom element example

ebidel:
Example for how to take a screenshot of an element on the page. We also didn't have any screenshot examples that showed how to use `clip`.

### DIFF
--- a/examples/screenshot-dom-element.js
+++ b/examples/screenshot-dom-element.js
@@ -1,0 +1,72 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


aslushnikov:
Let's wait for #445 to land - it will shrink this example to a few lines

ebidel:
SG. I'll update when that lands.

aslushnikov:
The #445 ended up introducing the `elementHandle.screenshot` method - i think there's no need for a dedicated example any more. A code snippet in documentation would be enough.

ebidel:
This one also showed how to use `clip` (no example for) and add padding to the screenshot. I was going to update it with the new API so we could have more runnable examples on https://try-puppeteer.appspot.com/.

PR #455, opened 2017-08-21
Adding proxy support

phawxby:
Without a disgusting amount of dependencies

Fix for bugs #448 & #337

### DIFF
--- a/utils/ChromiumDownloader.js
+++ b/utils/ChromiumDownloader.js
@@ -228,7 +228,7 @@ function applyAgent(opt)
   {
     let parsedProxy = process.env.http_proxy ? URL.parse(process.env.http_proxy) : URL.parse(process.env.https_proxy);
 
-    if (parsedProxy.protocol === "http:")
+    if (parsedProxy.protocol === 'http:')
       parsedProxy.secureProxy = false;
 
     opt.agent = new ProxyAgent(parsedProxy);
--- a/utils/ChromiumDownloader.js
+++ b/utils/ChromiumDownloader.js
@@ -227,10 +227,10 @@ function requestOptions(url, method = 'GET') {
   const proxyURL = getProxyForUrl(url);
   if (proxyURL) {
     const parsedProxyURL = URL.parse(proxyURL);
-    parsedProxyURL.secureProxy = parsedProxyURL.protocol === 'https:'
+    parsedProxyURL.secureProxy = parsedProxyURL.protocol === 'https:';
 
     result.agent = new ProxyAgent(parsedProxyURL);
   }
 
-  return result; 
+  return result;
 }
--- a/utils/ChromiumDownloader.js
+++ b/utils/ChromiumDownloader.js
@@ -64,10 +64,7 @@ module.exports = {
   canDownloadRevision: function(platform, revision) {
     console.assert(downloadURLs[platform], 'Unknown platform: ' + platform);
 
-    const options = URL.parse(util.format(downloadURLs[platform], revision));
-    options.method = 'HEAD';
-
-    applyAgent(options);
+    const options = requestOptions(util.format(downloadURLs[platform], revision), 'HEAD');
 
     let resolve;
     const promise = new Promise(x => resolve = x);
@@ -189,9 +186,7 @@ function downloadFile(url, destinationPath, progressCallback) {
 
   const promise = new Promise((x, y) => { fulfill = x; reject = y; });
 
-  const options = URL.parse(url);
-  applyAgent(options);
-
+  const options = requestOptions(url);
   const request = https.get(options, response => {
     if (response.statusCode !== 200) {
       const error = new Error(`Download failed: server returned code ${response.statusCode}. URL: ${url}`);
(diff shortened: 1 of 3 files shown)


utils/ChromiumDownloader.js
```diff
@@ -216,3 +221,16 @@ function downloadFile(url, destinationPath, progressCallback) {
 function extractZip(zipPath, folderPath) {
   return new Promise(fulfill => extract(zipPath, {dir: folderPath}, fulfill));
 }
+
+function applyAgent(opt)
+{
+  if (process.env.http_proxy || process.env.https_proxy)
```
aslushnikov: nit: put `{` on the same line

utils/ChromiumDownloader.js
```diff
@@ -216,3 +221,16 @@ function downloadFile(url, destinationPath, progressCallback) {
 function extractZip(zipPath, folderPath) {
   return new Promise(fulfill => extract(zipPath, {dir: folderPath}, fulfill));
 }
+
+function applyAgent(opt)
```
aslushnikov: nit: put `{` on the same line

```js
function applyAgent(opt) {
...
```

utils/ChromiumDownloader.js
```diff
@@ -216,3 +221,16 @@ function downloadFile(url, destinationPath, progressCallback) {
...
+{
+  if (process.env.http_proxy || process.env.https_proxy)
+  {
+    let parsedProxy = process.env.http_proxy ? URL.parse(process.env.http_proxy) : URL.parse(process.env.https_proxy);
```
aslushnikov: A few questions:
1. we should support capitalized env variables as well (and camel-cased, probably). There's also the NO_PROXY env variable, not sure how often it is used. Maybe rely on https://github.com/Rob--W/proxy-from-env to parse environment? 
2. we only do a single request to https url - do we care about `http_proxy` at all?

tardyp: 👍  please do not reinvent another way of parsing the proxy env variables.

phawxby:
Thanks for the feedback @aslushnikov. How about this?

utils/ChromiumDownloader.js
```diff
@@ -217,3 +224,17 @@ function downloadFile(url, destinationPath, progressCallback) {
 function extractZip(zipPath, folderPath) {
   return new Promise(fulfill => extract(zipPath, {dir: folderPath}, fulfill));
 }
+
+function applyAgent(opt) {
+  const proxy_url = getProxyForUrl(opt.href);
+
+  if (proxy_url)
+  {
```
aslushnikov: let's move this to the same line as `if` and call variable `proxyURL`

utils/ChromiumDownloader.js
```diff
@@ -217,3 +224,17 @@ function downloadFile(url, destinationPath, progressCallback) {
 function extractZip(zipPath, folderPath) {
   return new Promise(fulfill => extract(zipPath, {dir: folderPath}, fulfill));
 }
+
+function applyAgent(opt) {
```
aslushnikov: nit: let's call parameter `options`, we tend to avoid abbreviations

utils/ChromiumDownloader.js
```diff
@@ -217,3 +224,17 @@ function downloadFile(url, destinationPath, progressCallback) {
...
+  const proxy_url = getProxyForUrl(opt.href);
+
+  if (proxy_url)
+  {
+    const parsedProxy = URL.parse(proxy_url);
+
+    if (parsedProxy.protocol === 'http:')
+      parsedProxy.secureProxy = false;
```
aslushnikov: AFAIU secureProxy is never true. You rather want:

```js
parsedProxy.secureProxy = parsedProxy.protocol === 'https:';
```

utils/ChromiumDownloader.js
```diff
@@ -217,3 +224,17 @@ function downloadFile(url, destinationPath, progressCallback) {
...
+
+    opt.agent = new ProxyAgent(parsedProxy);
```
aslushnikov: I don't like that we pass result of URL.parse over to the ProxyAgent constructor. I had to check what's the output of `URL.parse` and make sure it doesn't clash with some options of `ProxyAgent`.

Let's make a method `requestOptions` which returns options for the `https.request` module, and will handle proxy inside. Something along the lines: 

```js
function requestOptions(url, method = 'GET') {
  const parsedURL = URL.parse(url);
  const result = {
    method,
    host: parsedURL.host,
    path: parsedURL.pathname
  };
  const proxyURL = getProxyForUrl(url);
  if (proxyURL) {
    const parsedProxyURL = URL.parse(proxyURL);
    result.agent = new ProxyAgent({
      host: parsedProxyURL.host,
      port: parsedProxyURL.port,
      secureProxy: parsedProxyURL.protocol === 'https:'
    });
  }
  return result;
}
```

aslushnikov (commented):
This is great! A few comments before we can have this landed.

utils/ChromiumDownloader.js
phawxby: Constructing a result object seems unnecessary, https.get and https.request accept a URL object and [internally it makes use of urlToOptions](https://github.com/nodejs/node/blob/master/lib/https.js#L234), it makes sense to [leave node to handle the conversion](https://github.com/nodejs/node/blob/master/lib/internal/url.js#L1298) to a options object should anything change in future. 

The same applies to `ProxyAgent`, [it uses url.parse internally too](https://github.com/TooTallNate/node-https-proxy-agent/blob/master/index.js#L27). When I switched it to using the constructed object as you suggested above it broke.

phawxby:
Once Travis does his thing hopefully that'll be it. Side note, linting absolutely explodes when run on windows to the point of being useless which is why I need to check in first. It would be handy if linting worked on Windows.

aslushnikov (approved):
Thanks a lot for the patch!

phawxby:
@aslushnikov do you guys have a release schedule? I'm just wondering when this will be deployed so I can start using it.

aslushnikov:
@phawxby we are still figuring out the process, so no schedule yet. But the next release should happen shortly

Issue #462, opened 2017-08-22
Block scripts


loilo:
Easy question: Is there a reliable way to block execution of any JavaScript on a `Page`?

I've not found anything on this in the docs, but I'm curious if it's possible at all?

JoelEinbinder:
What is your use case for this?

loilo:
I'm fiddling around with scraping webpages using headless Chromium and want to offer some options for blocking scripts, styles, images.

For styles and images, this is mainly to save some bandwidth, so just checking the `Accept` request header suffices for my use case (read: blocking things like inline `<style>` blocks is not my focus there).

However scripts are a bit of another story since they may alter page contents, so I want to try and block them entirely.

Of course, as soon as I disable JavaScript execution, I could just use a simple HTTP-based web scraper for the job, but if there's an option in Chromium then using the same technology stack as for the rest of the scraper would be preferrable.

PR #464, opened 2017-08-22
Add method to disable JavaScript

JoelEinbinder:
#462

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -53,8 +53,8 @@
     + [page.screenshot([options])](#pagescreenshotoptions)
     + [page.setContent(html)](#pagesetcontenthtml)
     + [page.setExtraHTTPHeaders(headers)](#pagesetextrahttpheadersheaders)
+    + [page.setJavaScriptEnabled(enabled)](#pagesetjavascriptenabledenabled)
     + [page.setRequestInterceptionEnabled(value)](#pagesetrequestinterceptionenabledvalue)
-    + [page.setJavaScriptEnabled(enabled)](#pagesetjavascriptEnabledenabled)
     + [page.setUserAgent(userAgent)](#pagesetuseragentuseragent)
     + [page.setViewport(viewport)](#pagesetviewportviewport)
     + [page.title()](#pagetitle)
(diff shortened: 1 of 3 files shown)


aslushnikov:
This might be interesting for @pavelfeldman and @dgozman.

I'm not aware what's the state of `Emulation.setScriptExecutionDisabled` - are we fine with un-experimenting it or thinking about removing.

JoelEinbinder:
I have no strong feelings about whether or not we should add this.

benjamingr:
Why a negaitve? Wouldn't it make more sense to publicly expose an enabling method (like in the Android WebView API)?

Garbee:
This would actually be a fantastic thing to have to test site interaction without JS. Like a site where you progressively enhance features only when JS is available.

JoelEinbinder:
@benjamingr If I recall correctly, WebView has JavaScript disabled by default.

benjamingr:
They do, the question is whether not to align with their API and call it `setJavaScriptEnabled` or something like that.

aslushnikov:
Bikeshedding method name: I like the `page.disableJavaScript(true)` - concise and easily searchable in documentation.

JoelEinbinder:
But then what happens when someone types `page.disableJavaScript()`  :)

aslushnikov:
> But then what happens when someone types page.disableJavaScript() :)

@JoelEinbinder Well...
```
(node:90673) AssertionError [ERR_ASSERTION]: First argument must be of type "boolean"!
    at Console.assert (console.js:164:23)
    at Page.disableJavaScript (/Users/lushnikov/prog/puppeteer/lib/Page.js:505:12)
    at __dirname (/Users/lushnikov/prog/puppeteer/b.js:6:8)
    at <anonymous>

```

pavelfeldman:
I'd rather call it `page.setJavaScriptEnabled(boolean)`. Always disliked `disabled = true` constructs.

benjamingr:
Another bikeshed, in Android and other platforms that use the webview the incorrect `Javascript` capitalization is used. I think we shouldn't use it and should use the compliant `JavaScript` instead.

benjamingr (approved):
Would love to see this land

JoelEinbinder:
Pavel wins the shed!
image

aslushnikov (approved):

loilo:
Awesome!

Issue #298, opened 2017-08-16
EPERM error attempting to run example


tivac:
Trying to run the following code:

```
var puppeteer = require("puppeteer");

puppeteer.launch().then(async browser => {
    let page = await browser.newPage();
    await page.goto('https://google.com');
    await page.screenshot({ path: 'screenshot.png' });
    browser.close();
});
```

I got an error:

`Error: EPERM: operation not permitted, unlink 'C:\Users\tivac\Desktop\puppeteer-testing\node_modules\puppeteer\.dev_profile1\CrashpadMetrics-active.pma'`

In the course of debugging this set `DEBUG=*` to try and understand where things were failing, but after doing that it then worked fine. 😕 

Does the first-run experience not quite work right in non-`DEBUG` modes? I'm not sure, but that's what happened for me!

- `puppeteer@0.9.0`
- `npm@5.3.0`
- `node@8.2.1`
- Windows 10 Version 1607 (OS Build 14393.1593)

aslushnikov:
Speculating: you might've accidentally launched the script twice, and two parallel puppeteers tried to run chrome with the same profile folder - `.dev_profile1`.

We should generate profile folder names randomly so that parallel runs of puppeteer work fine.

tivac:
It's possible! Seems unlikely though as I was launching node via the CLI and not doing anything different beyond originally setting `headless : false` since I wanted to see the browser. It repro'd every time until setting `DEBUG`, at which point everything began working perfectly.

builden:
+1

vinaybedre:
Hi,
I freshly used with only puppeteer, but still got the same error. I used the https://github.com/GoogleChrome/puppeteer/#usage first example script.

toolness:
I've been running into this problem as well, though with an even simpler example script that just launches the browser and then immediately tries to close it:

```js
const puppeteer = require('puppeteer');

(async () => {
  console.log("Launching browser.");

  const browser = await puppeteer.launch();

  console.log("Closing browser.");

  browser.close();
})();
```

I've tried manually deleting the `node_modules/puppeteer/.dev_profile1` directory before running this script, and the error still occurs.

As a workaround, I've found that simply calling `process.exit(0)` when I'm finished works okay--puppeteer seems to be very nice about cleaning up any launched Chromium child processes even in the case of unexpected parent process termination, so that's great!

awayken:
[...]

PR #488, opened 2017-08-23
Kill Chrome quickly on Windows

JoelEinbinder:
This fixed the issues of profiles being left behind and errors deleting profiles still in use on Windows. Tested on Windows 10 in `cmd` and `cygwin`.

#298

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -83,25 +83,31 @@ class Launcher {
 
     // Cleanup as processes exit.
     const listeners = [
-      helper.addEventListener(process, 'exit', killChromeAndCleanup),
-      helper.addEventListener(chromeProcess, 'exit', killChromeAndCleanup),
+      helper.addEventListener(process, 'exit', killChrome),
+      helper.addEventListener(chromeProcess, 'close', cleanupChrome),
     ];
     if (options.handleSIGINT !== false)
-      listeners.push(helper.addEventListener(process, 'SIGINT', killChromeAndCleanup));
+      listeners.push(helper.addEventListener(process, 'SIGINT', killChrome));
     try {
       const connectionDelay = options.slowMo || 0;
       const browserWSEndpoint = await waitForWSEndpoint(chromeProcess, options.timeout || 30 * 1000);
       const connection = await Connection.create(browserWSEndpoint, connectionDelay);
-      return new Browser(connection, !!options.ignoreHTTPSErrors, killChromeAndCleanup);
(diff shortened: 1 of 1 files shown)


lib/Launcher.js
```diff
@@ -83,26 +83,32 @@ class Launcher {
 
     // Cleanup as processes exit.
     const listeners = [
-      helper.addEventListener(process, 'exit', killChromeAndCleanup),
-      helper.addEventListener(chromeProcess, 'exit', killChromeAndCleanup),
+      helper.addEventListener(process, 'exit', killChrome),
+      helper.addEventListener(chromeProcess, 'close', cleanupChrome),
```
JoelEinbinder: `close` is important to wait for all io to be done.

lib/Launcher.js
```diff
@@ -83,26 +83,32 @@ class Launcher {
...
-    function killChromeAndCleanup() {
+    function cleanupChrome() {
       helper.removeEventListeners(listeners);
-      chromeProcess.kill('SIGKILL');
       removeSync(userDataDir);
     }
+
+    function killChrome() {
+      if (process.platform === 'win32')
+        childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T /F`);
```
JoelEinbinder: this is our heavy duty windows kill. Node reports `win32` for both 32 and 64 bit windows.

aslushnikov: Looks like chrome launcher, I like it :)

lib/Launcher.js
```diff
@@ -83,26 +83,32 @@ class Launcher {
 
     // Cleanup as processes exit.
     const listeners = [
-      helper.addEventListener(process, 'exit', killChromeAndCleanup),
-      helper.addEventListener(chromeProcess, 'exit', killChromeAndCleanup),
+      helper.addEventListener(process, 'exit', killChrome),
```
aslushnikov: the `helper.removeEventListeners` is not needed any more. Could you please drop the `listeners` altogether and use the `once` methods?
```js
  chromeProcess.once('close', () => removeSync(userDataDir));
  process.once('exit', killChrome);
  if (options.handleSIGINT !== false)
    process.once('SIGINT', killChrome);
```

aslushnikov (approved):
Awesome!

lib/Launcher.js
JoelEinbinder: once doesn't solve it completely because there are multiple paths into `killChrome`. I added a flag.

PR #501, opened 2017-08-23
readme: remove version from docs

ebidel:
The version in the docs is confusing users b/c it reflects master and not the latest release.

https://github.com/GoogleChrome/puppeteer/issues/492
https://github.com/GoogleChrome/puppeteer/issues/480

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,4 +1,4 @@
-# Puppeteer API (master)
+# Puppeteer API v<!-- GEN:version -->0.9.1-alpha<!-- GEN:stop-->
 
 ##### Table of Contents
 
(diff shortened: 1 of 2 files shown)


JoelEinbinder:
We should bump the version number immediately after we release, not right before we release.

docs/api.md
```diff
@@ -1,4 +1,4 @@
-# Puppeteer API v<!-- GEN:version -->0.9.0<!-- GEN:stop-->
+# Puppeteer API (<!-- GEN:version -->master<!-- GEN:stop-->)
```
aslushnikov: Remove the `<!-- GEN:version --> and <!-- GEN:stop -->` so that preprocessor doesn't mess with the file

ebidel: done

aslushnikov (approved):

PR #503, opened 2017-08-23
Add support for authentification

zqsd:
This adds support for "filling the authentification popup" in puppeteer.

While it's more complex than a simple `page.authentificate('user', 'password');` this is way more flexible because it allows to use different user/password for a different websites or proxies.
Exemple for proxy auth :

```

(async () => {
        args: ['--proxy-server=proxy.server.tld:3128'],
        headless: true,
    });

    await page.setRequestInterceptionEnabled(true);
        if(request.authChallenge) { // you can also check request.authChallenge.source === 'Proxy' for instance
            request.continue({
                authChallengeResponse: {
                    response: 'ProvideCredentials',
                    username: 'yourproxyusername',
                    password: 'yourproxypassword',
                }
            });
        }
        else {
            request.continue();
        }
    });

    await page.goto('https://ip.haxx.es', {waitUntil: 'networkidle'});
    console.log(await page.evaluate(() => {
        return document.body.innerText;
    }));
})();
```

It was tricky because when receiving an authentification, the same request is done again, but since the request was already done once some data was deleted (_requestIdToRequest, _interceptionIdToRequest). I managed a hack that works to keep this data, but I suppose there is a cleaner way so I would recommend that someone with a better knowledge on this base code that me to review and make something more clean.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,8 +23,10 @@
   "dependencies": {
     "debug": "^2.6.8",
     "extract-zip": "^1.6.5",
+    "https-proxy-agent": "^2.1.0",
     "mime": "^1.3.4",
     "progress": "^2.0.0",
+    "proxy-from-env": "^1.0.0",
     "rimraf": "^2.6.1",
     "ws": "^3.0.0"
   },
--- a/utils/ChromiumDownloader.js
+++ b/utils/ChromiumDownloader.js
@@ -22,6 +22,8 @@ const extract = require('extract-zip');
 const util = require('util');
 const URL = require('url');
 const removeRecursive = require('rimraf');
+const ProxyAgent = require('https-proxy-agent');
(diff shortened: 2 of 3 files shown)


zqsd:
signed the CLA

aslushnikov:
Hey @zqsd,

Thank you for the PR!

Could you explain why would you want to auth on a request-per-request basis? Of course, this is 
a more flexible API, but at the same time it's more clunky.

What are the scenarios that cannot be covered with `page.authenticate`?

zqsd:
I don't have this use case but It could be possible in some cases to have a proxy server requiring a user/password and the website requiring another user/password couple.
But yes a .authenticate method would be easier for most cases.

aslushnikov:
@zqsd would you like to implement the `page.authenticate` instead?

PR #504, opened 2017-08-23
[doc] elaborate on `url` parameter in `page.goto`.

### LINKED ISSUE #494: `page.goto` timeouts in headless mode for localhost pages if protocol is not specified
The problem can be reproduced with the following code:
```js
const http = require('http');

const server = http.createServer((req, res) => res.end('Hi'))
    .listen(1337, async () => {
        const browser = await puppeteer.launch({ headless: true });


        try {
            await page.goto('localhost:1337');
        } catch (err) {
            console.log(err);
        }

        console.log('Done');
        browser.close();
        server.close();
    });
```

If you set `headless` to `false` or specify any other domain program will print "Done" and exit, but for `localhost` without protocol it timeouts.  It works well for `http://localhost:1337` though. My wild guess is that `localhost` without a protocol is handled by omnibox or other component that is disabled in headless mode. I guess it's worth at least to be documented.

aslushnikov:
Fixes #494.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -513,7 +513,7 @@ can not go back, resolves to null.
 Navigate to the next page in history.
 
 #### page.goto(url, options)
-- `url` <[string]> URL to navigate page to
+- `url` <[string]> URL to navigate page to. The url should include scheme, e.g. `https://`.
 - `options` <[Object]> Navigation parameters which might have the following properties:
   - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds.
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -513,7 +513,7 @@ can not go back, resolves to null.
 Navigate to the next page in history.
 
 #### page.goto(url, options)
-- `url` <[string]> URL to navigate page to
+- `url` <[string]> URL to navigate page to. The url should include scheme, e.g. `https://`.
```
ebidel, line 516: include a scheme

ebidel (approved):

JoelEinbinder (approved):

PR #510, opened 2017-08-23
Page.setExtraHTTPHeaders - support object and map. Fixes #509

### LINKED ISSUE #509: setExtraHTTPHeaders: support both object and Map as `headers` argument
It would be nice to support passing an object literal into `setExtraHTTPHeaders` as well as a `Map`.

- The protocol already takes an object
- it'll reduce the (`new Map(Object.entries({}))`) boilerplate for users.
- objects are easier to work with.

this:
```js
await page.setExtraHTTPHeaders(new Map(Object.entries({
  'X-Header': 'blah blah' 
})));
```

to:
```js
  'X-Header': 'blah blah' 
});
```

ebidel:


aslushnikov:
Let's drop support for the Map. Also, let's do the same changes in Request.headers and Response.headers so that we're consistent in the API

aslushnikov:
Closing in favor of #567

PR #519, opened 2017-08-24
For your consideration, CircleCI 2.0

phawxby:
I don't really expect this to be accepted but I quickly threw this together to test Puppeteer against Node 7 and 8. And with #484 incoming I can see it being worthwhile testing under different environments.

image


aslushnikov:
So far we're pretty satisfied with Travis, so there's no need to switch.

Thank you for the PR though!

PR #530, opened 2017-08-24
add 'use strict'; to examples

vsemozhetbyt:
I am not sure what else non-module scripts should be addressed. If any, let me know to add them in another commit.

### DIFF
--- a/examples/.eslintrc.js
+++ b/examples/.eslintrc.js
@@ -11,5 +11,6 @@ module.exports = {
      */
     "rules": {
         "indent": [2, 2, { "SwitchCase": 1, "CallExpression": {"arguments": 2}, "MemberExpression": 2, "outerIIFEBody": 0 }],
+        "strict": [2, "global"],
     }
 };
--- a/examples/block-images.js
+++ b/examples/block-images.js
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+'use strict';
+
 const puppeteer = require('puppeteer');
 
 (async() => {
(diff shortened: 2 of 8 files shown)


ebidel (approved):

ebidel:
Should probably enforce this in https://github.com/GoogleChrome/puppeteer/blob/master/examples/.eslintrc.js

test/test.js
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+'use strict';
```
aslushnikov: why do we have it here? do we need to use 'use strict' in all of the source code?

vsemozhetbyt:
@ebidel 
> Should probably enforce this...

Hopefully done in the second commit.

test/test.js
vsemozhetbyt: I think this is the de-facto standard for non-module scripts if I get this right. Should the test check the API with the same semantics as users will use it in scripts?

ebidel: We only need it in non-imported code (tests, example). When users import `puppeteer` those are run as modules and have an implicit 'use strict'.

aslushnikov: > Should the test check the API with the same semantics as users will use it in scripts?

The `use strict` in the beginning of the test ensures that the `test.js` itself conforms to the strict mode; it doesn't imply anything about the puppeteer code. We also have a pretty tight eslint rules that should enforce us into good code practices, so `use strict` in the `test.js` doesn't give us anything.

vsemozhetbyt: Should I exclude this one from the commit?

aslushnikov: @vsemozhetbyt yeah, let's exclude

vsemozhetbyt: Done.

PR #533, opened 2017-08-25
Kill PhantomShim.

aslushnikov:
This patch kills PhantomShim, a lightweight layer that helped to
bootstrap Puppeteer API.

At this point, PhantomShim:
- passes 139 phantomJS tests
- fails 13 tests
- lacks capabilities to run 26

The failings and unsupported tests don't have any value for Puppeteer
API, so it doesn't make sense to maintain PhantomShim any longer.


package.json
```diff
@@ -19,7 +18,6 @@
     "coverage": "COVERAGE=true npm run unit",
     "node6": "node utils/node6-transform/index.js",
     "test-node6": "jasmine utils/node6-transform/test/test.js",
-    "build": "npm run node6",
```
JoelEinbinder: I think this was unintentional removal.

JoelEinbinder:
Github lags when looking through the files changed, but I think you missed the entires in `.gitignore` and `.npmignore`

package.json
aslushnikov: That's a drive-by. Do we need it?

aslushnikov:
Thanks, cleaned up `.npmignore`. There's nothing related to phantom shim in `.gitignore`

package.json
JoelEinbinder: I wanted to future proof, in case we have more build steps.

JoelEinbinder (approved):

PR #543, opened 2017-08-25
Add attribute(key) and property(key) methods to ElementHandle

batiste:
Related to my issue here https://github.com/GoogleChrome/puppeteer/issues/438

I implemented those methods. Those are just simple shortcuts but I think they might be useful to write tests.

```js
    it('should return attribute', SX(async function() {
      await page.setContent('<section id="testAttribute">43543</section>');
      const element = await page.$('section');
      expect(element).toBeTruthy();
      expect(await element.attribute('id')).toBe('testAttribute');
    }));
```


batiste:
I signed it!

aslushnikov:
Nicee, thank you for the PR!

We're currently working on #382, and the feature will add the `properties()` method that would return object handles.

Given this, let's scope this PR only to the `attribute()` method.

batiste:
@aslushnikov very well. I removed the property method.

aslushnikov:
@batiste please run the `npm run doc` command to update table-of-contents in `api.md` and satisfy documentation linter.

batiste:
@aslushnikov done!

aslushnikov (approved):

PR #548, opened 2017-08-25
Fix node6 support for Object.entries and URL object

JoelEinbinder:
`Object.entries` is missing from node6
`url.URL` is missing from node6
The ESTreeWalker was failing on nested awaits
Esprima was failing on return statements outside of functions


utils/node6-transform/test/test.js
```diff
@@ -64,4 +64,10 @@ describe('TransformAsyncFunctions', function() {
     expect(output instanceof Promise).toBe(true);
     output.then(result => expect(result).toBe(123)).then(done);
   });
+  fit('should work with double await', function(done) {
```
aslushnikov: it

aslushnikov (approved):

PR #551, opened 2017-08-25
Test node6 on travis

JoelEinbinder:


utils/node6-transform/index.js
```diff
@@ -19,18 +19,24 @@ const path = require('path');
...
-});
+function copyFolder(source, target, root) {
+  if (fs.existsSync(target))
+    removeRecursive(target);
+  fs.mkdirSync(target);
 
+  fs.readdirSync(source).forEach(file => {
+    const curSource = path.join(source, file);
+    if (fs.lstatSync(curSource).isDirectory()) {
+      copyFolder(curSource, path.join(target, path.basename(curSource)));
```
aslushnikov: isn't `path.basename(curSource) === file`?

utils/node6-transform/index.js
```diff
@@ -19,18 +19,24 @@ const path = require('path');
...
-if (fs.existsSync(outPath))
-  removeRecursive(outPath);
-fs.mkdirSync(outPath);
 
-filePaths.forEach(filePath => {
-  const content = fs.readFileSync(path.join(dirPath, filePath), 'utf8');
-  const output = transformAsyncFunctions(content);
-  fs.writeFileSync(path.resolve(outPath, filePath), output);
-});
+function copyFolder(source, target, root) {
```
aslushnikov: root is not used

utils/node6-transform/index.js
```diff
@@ -19,18 +19,24 @@ const path = require('path');
...
+  if (fs.existsSync(target))
+    removeRecursive(target);
+  fs.mkdirSync(target);
 
+  fs.readdirSync(source).forEach(file => {
+    const curSource = path.join(source, file);
```
aslushnikov: Let's do the path-joining magic in the very beginning so that it's easier to follow what's going on.
```js
const from = path.join(source, file);
const to = path.join(target, file);
...
```

utils/node6-transform/index.js
```diff
@@ -19,18 +19,24 @@ const path = require('path');
...
+    if (fs.lstatSync(curSource).isDirectory()) {
+      copyFolder(curSource, path.join(target, path.basename(curSource)));
+    } else {
```
aslushnikov: let's check for file explicitly. It would not be nice to write over a symlink.

utils/node6-transform/index.js
```diff
@@ -19,18 +19,24 @@ const path = require('path');
...
+      let text = fs.readFileSync(curSource);
+      if (curSource.endsWith('.js'))
+        text = transformAsyncFunctions(text.toString()).replace(/require\('\.\.\/lib\//g, `require('../node6/`);
```
aslushnikov: doesn't copying everything into a `node6/` work? this would spare you from this replace.

I.e.:
```js
removeSync('node6');
copyFolder('lib', 'node6/lib');
copyFolder('test', 'node6/test');
copyFile('index.js', 'node6/index.js');
```

utils/node6-transform/index.js
JoelEinbinder: I don't know what you mean by this one.

utils/node6-transform/index.js
JoelEinbinder: It gets really tricky because we load some utils from lib, and some utils are hard to convert. There are also paths for things like `package.json` and `.local-chromium` which get confusing.

utils/node6-transform/index.js
aslushnikov: I mean:

```js
 if (fs.lstatSync(curSource).isDirectory()) {
  ...
} else if (fs.lstatSync(curSource).isFile()) {
  ...
}
```

aslushnikov (approved):

Issue #443, opened 2017-08-21
Request: browser.currentPage() or similar way to access Pages


karelrochelt:
I was playing around with Puppeteer today and I found out that there is no way (that I know of) to run script like this:
```
const puppeteer = require('puppeteer');

puppeteer.connect({browserWSEndpoint: <endpoint>}).then(async browser => {
  const page = await browser.currentPage();
  // other actions that would manipulate the current page
  browser.close();
});
```

The only way to get a Page instance seems to be browser.newPage() which is rather inconvenient if I want to run several test scripts that all connect to one browser instance.

JoelEinbinder:
This is related to #386

I was thinking we should fire an event when the first tab is created in the browser, or have browser.pages() return all pages.

karelrochelt:
browser.pages() would be great!

PsychoXIVI:
Still there is no function to get current page, or at least check is active?

AgainPsychoX:
To find current page (still no official function, someone can push it):

```js
	browser.currentPage = async function () {
		let targets = await this.targets();
		for (let i = 0, I = targets.length; i < I; ++i) {
			target = targets[i];
			let page = await target.page();
			if (page) {
				return page;
			}
		}
		return null;
	}
```

mkolinichenko:
So, is there a way to access the current page (i.e. page on active browser tab)? 
Right now I'm using something like this:
```javascript
    async function getActivePage(browser, timeout) {
        var start = new Date().getTime();
        while(new Date().getTime() - start < timeout) {
            var pages = await browser.pages();
            var arr = [];
            for (const p of pages) {
                if(await p.evaluate(() => { return document.visibilityState == 'visible' })) {
                    arr.push(p);
                }
            }
            if(arr.length == 1) return arr[0];
        }
        throw "Unable to get active page";
    }
```
But it doesn't seem very robust.

aslushnikov:
@mkolinichenko What's your usecase? IIRC headless considers all pages to be active.

mkolinichenko:
@aslushnikov, thanks for the response. I'm running headfull, trying to combine user interaction with some automation. User switches tabs manually and at some point my app needs to know which tab/page/at least url we're currently on.

aslushnikov:
@mkolinichenko I see, thanks.

We don't have this signal in [DevTools protocol](https://vanilla.aslushnikov.com/) atm. You can file a new issue here, and If there is enough demand we'll add it.

[...]

Issue #386, opened 2017-08-19
Emit new Page objects when new tabs created


notgne2:
Not sure how this would work, or if possible at all with how puppeteer works. A nice feature would be if the browser emit an event with a new `Page` when a new tab is opened by clicking on a `target="_blank"` or through some other means.

michaelfward:
I think I'd like to take a whack at this one. Does anyone have any objections / suggestions? @aslushnikov

aslushnikov:
@michaelfward take a look at Target domain, e.g. `Target.targetCreated` event https://chromedevtools.github.io/devtools-protocol/tot/Target/#event-targetCreated

michaelfward:
@aslushnikov thanks! I should have a PR out for this in the next 24 hours

nwhitmont:
Yes! This would be super useful!

I have a similar question open on Stack Overflow: [Puppeteer: How to handle multiple tabs?](https://stackoverflow.com/questions/45806684/puppeteer-how-to-handle-multiple-tabs)

nwhitmont:
@michaelfward Any update?

ebidel:
@nwhitmont did you see https://github.com/GoogleChrome/puppeteer/pull/554?

michaelfward:
@nwhitmont apologies, I saw @JoelEinbinder 's PR and figured I'd turn to other issues! Does his implementation take care of your use case?

nwhitmont:
@michaelfward Ah ha! That explains it ;-) Thanks for the update.

JasonBoy:
one workaround is in `evaluate` method, get the anchor, and `ele.removeAttribute('target')`, then you click it

pirate:
Not sure how you guys were planning on implementing this, but an API like this would be awesome!

`event:newpage` Gets triggered by things like `window.open()` or links clicked with `target="_blank"`.
```javascript
page.on('newpage', async (new_page) => {
    const url = new_page.url;
    console.log('Browser opened new tab', url);
    const page = await new_page.page();
});
```

If anyone is looking for a workaround, for now I'm using:
```javascript
await page.evaluate(() => {
    window.open = (new_url) => {window.location.href = new_url}
});
page.on('load', (event) => {
    console.log('Opened new url', window.location.toString());
});
```

avimar:
A few questions.

Goal: I'm filling in a form and getting a list of links. To not waste time and fragility on the dynamic page clicking the link and then going back, I'm opening each link in a new window.

[...]

PR #554, opened 2017-08-26
Browser pages

JoelEinbinder:
This adds new events `pageopened` and `pageclosed` to browser.
It also adds `browser.pages()`.
#386 #443

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -296,10 +296,10 @@ Emitted when the url of a target changes.
 
 Emitted when a target is created, for example when a new page is opened by [`window.open`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) or [`browser.newPage`](#browsernewpage).
 
-Emitted when a target is destroyed, for example when a page is closed.
-
 #### event: 'targetdestroyed'
 - <[Target]>
+
+Emitted when a target is destroyed, for example when a page is closed.
 
 #### browser.close()
 - returns: <[Promise]>
@@ -1908,6 +1908,8 @@ Contains the status code of the response (e.g., 200 for a success).
 #### response.url
 - <[string]>
 
+Contains the URL of the response.
+
 ### class: Target
 
 #### target.page()
@@ -1922,8 +1924,6 @@ Identifies what kind of target this is. Can be `"page"`, `"service_worker"`, or 
 
 #### target.url()
 - returns: <[string]>
-
-Contains the URL of the response.
 
 [Array]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array "Array"
 [boolean]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type "Boolean"
--- a/docs/api.md
+++ b/docs/api.md
@@ -13,9 +13,9 @@
   * [puppeteer.executablePath()](#puppeteerexecutablepath)
   * [puppeteer.launch([options])](#puppeteerlaunchoptions)
 - [class: Browser](#class-browser)
+  * [event: 'targetchanged'](#event-targetchanged)
   * [event: 'targetcreated'](#event-targetcreated)
   * [event: 'targetdestroyed'](#event-targetdestroyed)
-  * [event: 'targeturlchanged'](#event-targeturlchanged)
   * [browser.close()](#browserclose)
   * [browser.disconnect()](#browserdisconnect)
   * [browser.newPage()](#browsernewpage)
@@ -286,20 +286,20 @@ puppeteer.launch().then(async browser => {
 });
 ```
 
+#### event: 'targetchanged'
+- <[Target]>
+
+Emitted when the url of a target changes.
+
 #### event: 'targetcreated'
 - <[Target]>
 
 Emitted when a target is created, for example when a new page is opened by [`window.open`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) or [`browser.newPage`](#browsernewpage).
 
-#### event: 'targetdestroyed'
-- <[Target]>
-
 Emitted when a target is destroyed, for example when a page is closed.
 
-#### event: 'targeturlchanged'
+#### event: 'targetdestroyed'
 - <[Target]>
-
-Emitted when the url of a target changes.
 
 #### browser.close()
 - returns: <[Promise]>
--- a/docs/api.md
+++ b/docs/api.md
@@ -180,7 +180,6 @@
   * [response.url](#responseurl)
 - [class: Target](#class-target)
   * [target.page()](#targetpage)
-  * [target.title()](#targettitle)
   * [target.type()](#targettype)
   * [target.url()](#targeturl)
 
@@ -312,7 +311,7 @@ Disconnects Puppeteer from the browser, but leaves the Chromium process running.
 - returns: <[Promise]<[Array]<[Page]>>> Promise which resolves to an array of all open pages.
 
 #### browser.targets()
-- returns: <[Promise]<[Array]<[Page]>>> Promise which resolves to an array of all open pages.
+- returns: <[Array]<[Target]>> An array of all active targets.
 
 #### browser.version()
 - returns: <[Promise]<[string]>> For headless Chromium, this is similar to `HeadlessChrome/61.0.3153.0`. For non-headless, this is similar to `Chrome/61.0.3153.0`.
@@ -1910,16 +1909,13 @@ Contains the status code of the response (e.g., 200 for a success).
 
 If the target is not of type `"page"`, returns `null`.
 
-#### target.title()
-- returns: <[Promise]<[string]>>
-
 #### target.type()
 - returns: <[string]>
 
 Identifies what kind of target this is such as `"page"` or `"service_worker"`.
 
 #### target.url()
-- returns: <[Promise]<[string]>>
+- returns: <[string]>
 
 Contains the URL of the response.
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -15,6 +15,7 @@
 - [class: Browser](#class-browser)
   * [event: 'targetcreated'](#event-targetcreated)
   * [event: 'targetdestroyed'](#event-targetdestroyed)
+  * [event: 'targeturlchanged'](#event-targeturlchanged)
   * [browser.close()](#browserclose)
   * [browser.disconnect()](#browserdisconnect)
   * [browser.newPage()](#browsernewpage)
@@ -294,6 +295,11 @@ Emitted when a target is created, for example when a new page is opened by [`win
 - <[Target]>
 
 Emitted when a target is destroyed, for example when a page is closed.
+
+#### event: 'targeturlchanged'
+- <[Target]>
+
+Emitted when the url of a target changes.
 
 #### browser.close()
 - returns: <[Promise]>
--- a/docs/api.md
+++ b/docs/api.md
@@ -13,9 +13,13 @@
   * [puppeteer.executablePath()](#puppeteerexecutablepath)
   * [puppeteer.launch([options])](#puppeteerlaunchoptions)
 - [class: Browser](#class-browser)
+  * [event: 'targetcreated'](#event-targetcreated)
+  * [event: 'targetdestroyed'](#event-targetdestroyed)
   * [browser.close()](#browserclose)
   * [browser.disconnect()](#browserdisconnect)
   * [browser.newPage()](#browsernewpage)
+  * [browser.pages()](#browserpages)
+  * [browser.targets()](#browsertargets)
   * [browser.version()](#browserversion)
   * [browser.wsEndpoint()](#browserwsendpoint)
 - [class: Page](#class-page)
@@ -174,6 +178,11 @@
   * [response.status](#responsestatus)
   * [response.text()](#responsetext)
   * [response.url](#responseurl)
+- [class: Target](#class-target)
+  * [target.page()](#targetpage)
+  * [target.title()](#targettitle)
+  * [target.type()](#targettype)
+  * [target.url()](#targeturl)
 
 <!-- tocstop -->
 
@@ -277,6 +286,16 @@ puppeteer.launch().then(async browser => {
 });
 ```
 
+#### event: 'targetcreated'
+- <[Target]>
+
+Emitted when a target is created, for example when a new page is opened by [`window.open`](https://developer.mozilla.org/en-US/docs/Web/API/Window/open) or [`browser.newPage`](#browsernewpage).
+
+#### event: 'targetdestroyed'
+- <[Target]>
+
+Emitted when a target is destroyed, for example when a page is closed.
+
 #### browser.close()
 - returns: <[Promise]>
 
@@ -288,6 +307,12 @@ Disconnects Puppeteer from the browser, but leaves the Chromium process running.
 
 #### browser.newPage()
 - returns: <[Promise]<[Page]>> Promise which resolves to a new [Page] object.
+
+#### browser.pages()
+- returns: <[Promise]<[Array]<[Page]>>> Promise which resolves to an array of all open pages.
+
+#### browser.targets()
+- returns: <[Promise]<[Array]<[Page]>>> Promise which resolves to an array of all open pages.
 
 #### browser.version()
 - returns: <[Promise]<[string]>> For headless Chromium, this is similar to `HeadlessChrome/61.0.3153.0`. For non-headless, this is similar to `Chrome/61.0.3153.0`.
@@ -1878,6 +1903,24 @@ Contains the status code of the response (e.g., 200 for a success).
 #### response.url
 - <[string]>
 
+### class: Target
+
+#### target.page()
+- returns: <[Promise]<[Page]>>
+
+If the target is not of type `"page"`, returns `null`.
+
+#### target.title()
+- returns: <[Promise]<[string]>>
+
+#### target.type()
+- returns: <[string]>
+
+Identifies what kind of target this is such as `"page"` or `"service_worker"`.
+
+#### target.url()
+- returns: <[Promise]<[string]>>
+
 Contains the URL of the response.
 
 [Array]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array "Array"
@@ -1911,3 +1954,4 @@ Contains the URL of the response.
 [UIEvent.detail]: https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/detail "UIEvent.detail"
 [Serializable]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#Description "Serializable"
 [Touchscreen]: #class-touchscreen "Touchscreen"
+[Target]: #class-target "Target"
--- a/docs/api.md
+++ b/docs/api.md
@@ -1912,7 +1912,7 @@ If the target is not of type `"page"`, returns `null`.
 #### target.type()
 - returns: <[string]>
 
-Identifies what kind of target this is such as `"page"` or `"service_worker"`.
+Identifies what kind of target this is. Can be `"page"`, `"service_worker"`, or `"other"`.
 
 #### target.url()
 - returns: <[string]>
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -79,7 +79,7 @@ class Browser extends EventEmitter {
 
   /**
    * @param {!Target.TargetInfo} targetInfo
-   * @return {?Target}
+   * @return {!Target}
    */
   _ensureTargetForInfo(targetInfo) {
     if (!this._targets.has(targetInfo.targetId)) {
@@ -87,6 +87,17 @@ class Browser extends EventEmitter {
       this._targets.set(targetInfo.targetId, target);
     }
     return this._targets.get(targetInfo.targetId);
+  }
+
+  /**
+   * @param {string} targetId
+   * @return {!Promise<?Target>}
+   */
+  async _targetForId(targetId) {
+    const target = this._targets.get(targetId);
+    if (!target || !(await target._initializedPromise))
+      return null;
+    return target;
   }
 
   /**
@@ -101,9 +112,8 @@ class Browser extends EventEmitter {
    */
   async newPage() {
     const {targetId} = await this._connection.send('Target.createTarget', {url: 'about:blank'});
-    const {targetInfo} = await this._connection.send('Target.getTargetInfo', {targetId});
-    const target = this._ensureTargetForInfo(targetInfo);
-    await target._initializedPromise;
+    const target = await this._targetForId(targetId);
+    console.assert(target, 'Failed to create target for page');
     const page = await target.page();
     await page.goto('about:blank');
     return page;
@@ -218,7 +228,7 @@ class Target {
     const previousURL = this._targetInfo.url;
     this._targetInfo = targetInfo;
 
-    if (this._targetInfo.type !== 'page' || this._targetInfo.url !== '') {
+    if (!this._wasInitialized && (this._targetInfo.type !== 'page' || this._targetInfo.url !== '')) {
       this._wasInitialized = true;
       this._initialized(true);
       return;
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -176,7 +176,7 @@ class Target {
   constructor(browser, targetInfo) {
     this._browser = browser;
     this._targetInfo = targetInfo;
-    /** @type {!Promise<!Page>} */
+    /** @type {?Promise<!Page>} */
     this._pagePromise = null;
   }
 
@@ -203,11 +203,8 @@ class Target {
    */
   type() {
     const type = this._targetInfo.type;
-    switch (type) {
-      case 'service_worker':
-      case 'page':
-        return type;
-    }
+    if (type === 'page' || type === 'service_worker')
+      return type;
     return 'other';
   }
 
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -39,7 +39,7 @@ class Browser extends EventEmitter {
   }
 
   /**
-   * @param {!Connection} connection
+   * @param {!Puppeteer.Connection} connection
    * @param {boolean} ignoreHTTPSErrors
    * @param {function()=} closeCallback
    */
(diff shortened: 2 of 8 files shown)


lib/Browser.js
```diff
@@ -16,18 +16,69 @@
...
+    const browser =  new Browser(connection, ignoreHTTPSErrors, closeCallback);
+    await connection.send('Target.setDiscoverTargets', {discover: true});
+    return browser;
+  }
+
+  /**
+   * @param {{targetInfo: {targetId: string}}} param0
+   */
+  async _targetCreated({targetInfo}) {
+    if (targetInfo.type !== 'page' && targetInfo.type !== 'other')
```
JoelEinbinder: The newtab pages come in as 'other'. I'm not sure whether this is intentional or not, but it makes them hard to work with.

vsemozhetbyt:
Sorry, it seems the version in the heading of [this doc](https://github.com/GoogleChrome/puppeteer/blob/v0.10.0/docs/api.md) needs to be updated into 0.10.0.

twalpole:
I believe these tests fail in non-headless mode, as soon as `window.open` is run, with "Unhandled promise rejection (rejection id: 1): Error: Protocol error (Network.enable): Target closed."  I think it's due to Chrome shifting tab focus, but not really sure.

twalpole:
Additionally, it appears the main frame is not attached to the page yet when `pageopened` is  fired, limiting the ability to do things like  `injectFile`, etc.  Is this intended?

lib/Browser.js
dgozman: Please file an upstream bug instead of this.

duckywang1:
This PR is awesome!   
I am eager to get something like `browser.pages()`. So I did some tests with [Branch: browser_pages](https://github.com/JoelEinbinder/puppeteer/tree/browser_pages).

What I did:
1. try it with `headless: true`.  
The results were great. I could get pages which comes from `page.goto()` and which created from `a[target="_blank"]` 

2. try it with `headless: false`.  
Error came out. `nodejs.unhandledRejectionError: Protocol error (Network.enable): Target closed`. Just like what @twalpole said.

twalpole:
@duckywang1 When running headless, were you able to actually do anything/interact with the pages created from `a[target="_blank"]` ?

duckywang1:
Apologies. I have no experience of unit test.
Yesterday I didn't try this feature seriously. Today I write an example of my business logic with `browser.pages();` and `headless browser`. It worked!
I can interact with page created from `a[target="_blank"]`. @twalpole 

Now I hope this great feature would be merged, althought it only take effect in `headless browser` temporarily.

Here is my example:  
I got two pages.
1. detail page.
This page provide an HTMLElememt `a[target="_blank"]` and link to download page
2. download page.
This page will provide a resource URL and download it in 2-10 second automatically.

Here is my example code(fake code, but it did work):
```
const browser = initHeadlessBrowser(); // create a Headless browser
const detailPage = browser.newPage();
detailPage.goto(url, '{ waitUntil: 'networkidle' }');
const btn = page.$('.download');
btn.click({ delay: 100 }); // create a download page but we can fetch them with browser.pages();

const allPages = browser.pages();
for (let i = 1; i < allPages.length; i += 1) {
  // allPages[0] is a blank page. We can't get any valid info. If interested, you can open a headless: false browser and see the GUI of Chrominum.
  // compare something with your download page
  // here I confirm it with page.url();
  if (condition) {
    const downloadPage = allPages[i];
    await downloadPage.setRequestInterceptionEnabled(true);
    downloadPage.on('request', request => {
      if (request.url.indexOf('.png') >= 0) {
        // here i use node_modules`download`
        download(request.url).then(streamData => {
          // get streamData and do something fun
        });
      } else {
        request.continue();
      }
    });
    await downloadPage.waitFor(30000); // waiting for loading views
    await downloadPage.close(); // close download page
  }
}

```
By the way, I think it seems a little strange of `how to use the downloadPage` or `lifeCycle of downloadPage`. But it worked and suprised me.

twalpole:
@duckywang1 Ah -- I see you're using `pages` not the 'pageopened' and 'pageclosed' events

nwhitmont:
Looking forward to this!

avimar:
Can someone explain this?
>The newtab pages come in as 'other'

... it seems unnecessary to launch a new browser and _always_ having to open a new tab. Why can't I just use the one that's already open?

dmportella:
Any more work being done on this?

nwhitmont:
@JoelEinbinder What's the latest update on Browser Pages?

schelkun:
I'm waiting for this PR will be accepted. Any problems with this PR?

dmportella:
i am trying to use your changes to capture a popup but it doesnt seem to work.. there are no errors but no pagecreated event is fired

I could share the snippet - could be that i am doing something wrong

JoelEinbinder:
The Chromium can be rolled to r501642 to fix targets appearing as `"other"`.

@dmportella Sharing the snippet would be very useful to me! Thanks.

nwhitmont:
@dgozman @JoelEinbinder @vsemozhetbyt 

Hey guys, can you help us understand what is preventing this from being accepted/merged?

aslushnikov:
@nwhitmont this PR doesn't work well enough. One of the issues is that Puppeteer's FrameManager doesn't know how to attach to a non-empty page.

We're working on it; we'll merge this PR as all the blockers are eliminated.

ghost (approved):

JoelEinbinder:
There is still a race somewhere when running in headful mode. `Target.attachToTarget` returns before the target is usable, and so the first message to it sometimes gets rejected with a target closed error.

nwhitmont:
Thanks for the updates & keep up the great work!

zxsoft:
Hope this PR can be approved in the next release. I need it so badly.

I hold a bunch of page handlers (because of no browser.pages() ), and when a page is crashed, i don't know. So my page handlers are not reliable. And I cannot make it better. because I don't know wether a page is available or not.

trafostream:
How to setRequestInterceptionEnabled on new popup pages (target="_blank") and attach event on request. For example, I need block the download popup pages by url.
Thank you.
Sorry for my english.

I would like something like this:

```
browser.on('pageopened',  pageOpened => {
        pageOpened.setRequestInterceptionEnabled(true);
        pageOpened.on('request', request => {
              request.abort();
        });
});
```

lib/Browser.js
```diff
@@ -30,6 +30,55 @@ class Browser extends EventEmitter {
...
+  }
+
+  /**
+   * @param {!Connection} connection
+   * @param {boolean} ignoreHTTPSErrors
+   * @param {function()=} closeCallback
+   */
+  static async create(connection, ignoreHTTPSErrors, closeCallback) {
+    const browser =  new Browser(connection, ignoreHTTPSErrors, closeCallback);
+    await connection.send('Target.setDiscoverTargets', {discover: true});
```
pavelfeldman, line 48: You should call `Target.getTargets` to read present state and also you should modify it do the same thing `Target.setDiscoverTargets` does. That way you don't have a race between these calls and target creation.

lib/Browser.js
```diff
@@ -30,6 +30,55 @@ class Browser extends EventEmitter {
...
+  }
+
+  /**
+   * @param {string} targetId
+   * @return {Page}
+   */
+  _pageForTargetId(targetId) {
+    if (this._pages.has(targetId))
+      return this._pages.get(targetId);
+    const pagePromise = this._connection.createSession(targetId).then(client => Page.create(client, this._ignoreHTTPSErrors, this._screenshotTaskQueue));
```
pavelfeldman: I don't think proactively attaching to all the targets is necessary. I'd rather expose the API as it is in the protocol - you can list the targets and you can materialize them as pages...

lib/Page.js
```diff
@@ -44,9 +44,9 @@ class Page extends EventEmitter {
     if (ignoreHTTPSErrors)
       await client.send('Security.setOverrideCertificateErrors', {override: true});
     const page = new Page(client, ignoreHTTPSErrors, screenshotTaskQueue);
-    await page.goto('about:blank');
     // Initialize default page size.
     await page.setViewport({width: 800, height: 600});
+    await page._frameManager.fetchResourceTree();
```
pavelfeldman: You are fetching resource tree after initializing Runtime - that way you won't get execution contexts mapped to the frames.

lib/Browser.js
dgozman: Please don't :) Target.setDiscoverTargets should notify about all existing ones through Target.targetCreated. There is no need to use Target.getTargets at all.

pavelfeldman: Well, if you look at [TargetHandler::SetDiscoverTargets](https://cs.chromium.org/chromium/src/content/browser/devtools/protocol/target_handler.cc?q=TargetHandler::SetD&sq=package:chromium&l=187), you would be certain it does not. Except for I now see that your observer has side effects. But that is just mean, you know, calling it observer and making it change the model!

docs/api.md
```diff
@@ -223,6 +242,12 @@ Closes browser with all the pages (if any were opened). The browser object itsel
 #### browser.newPage()
 - returns: <[Promise]<[Page]>> Promise which resolves to a new [Page] object.
 
+#### browser.pages()
+- returns: <[Promise]<[Array]<[Page]>>> Promise which resolves to an array of all open pages.
+
+#### browser.targets()
+- returns: <[Promise]<[Array]<[Page]>>> Promise which resolves to an array of all open pages.
```
aslushnikov: to an array of all open targets

docs/api.md
```diff
@@ -1439,6 +1464,24 @@ Contains the status code of the response (e.g., 200 for a success).
...
 - <[string]>
 
+### class: Target
+
+#### target.page()
+- returns: <[Promise]<[Page]>>
+
+If the target is not of type `"page"`, returns `null`.
+
+#### target.title()
```
aslushnikov: let's drop it?

lib/Browser.js
```diff
@@ -81,3 +147,61 @@ class TaskQueue {
...
+  }
+
+  /**
+   * @return {Promise<string>}
+   */
+  async url() {
+    return (await this._info()).url;
+  }
+
+  type() {
```
aslushnikov: nit: jsdoc

lib/Browser.js
```diff
@@ -81,3 +147,61 @@ class TaskQueue {
...
+    return this._type;
+  }
+}
+helper.tracePublicAPI(Target);
+
+/**
+ * @typedef {Object} Target.targetInfo
```
aslushnikov: this is never used

lib/Page.js
```diff
@@ -36,6 +36,9 @@ class Page extends EventEmitter {
    * @return {!Promise<!Page>}
    */
   static async create(client, ignoreHTTPSErrors, appMode, screenshotTaskQueue) {
+    const page = new Page(client, ignoreHTTPSErrors, screenshotTaskQueue);
+    page._frameManager.fetchResourceTree();
```
aslushnikov: don't you want to await it?

lib/Page.js
```diff
@@ -44,8 +47,6 @@ class Page extends EventEmitter {
     ]);
     if (ignoreHTTPSErrors)
       await client.send('Security.setOverrideCertificateErrors', {override: true});
-    const page = new Page(client, ignoreHTTPSErrors, screenshotTaskQueue);
```
aslushnikov, line 49: why this change?

lib/Browser.js
```diff
@@ -31,6 +31,50 @@ class Browser extends EventEmitter {
...
+      return;
+    this._targets.delete(targetId);
+    this.emit(Browser.Events.TargetDestroyed, target);
+  }
+
+  /**
+   * @param {{targetId: string, type: string}} targetInfo
+   * @return {Target}
+   */
+  _targetForInfo(targetInfo) {
```
aslushnikov: nit: this kind of methods are usually prefixed with `_ensure`: `_ensureTargetForInfo`

lib/Browser.js
```diff
@@ -45,8 +89,24 @@ class Browser extends EventEmitter {
...
-    const client = await this._connection.createSession(targetId);
-    return await Page.create(client, this._ignoreHTTPSErrors, this._appMode, this._screenshotTaskQueue);
+    const {targetInfo} = await this._connection.send('Target.getTargetInfo', {targetId});
+    const page = await this._targetForInfo(targetInfo).page();
+    await page.goto('about:blank');
+    return page;
+  }
+
+  /**
+   * @return {Array<Target>}
```
aslushnikov: For the sake of consistency, please either:
- use `'!'` for non-nullable types in this patch
- remove all exclamation marks throughout the code first

lib/Browser.js
```diff
@@ -45,8 +89,24 @@ class Browser extends EventEmitter {
...
+   */
+  targets() {
+    return Array.from(this._targets.values());
+  }
+
+  /**
+   * @return {Array<Pages>}
```
aslushnikov: Promise type declaration is missing

I also like `async` methods since they make it really obvious that the method should be awaited. Let's make `pages` async!

lib/Browser.js
```diff
@@ -45,8 +89,24 @@ class Browser extends EventEmitter {
...
+   */
+  pages() {
+    return Promise.all(this.targets().map(target => target.page()).filter(page => page));
```
aslushnikov: nit: `page => !!page`

lib/FrameManager.js
```diff
@@ -40,6 +40,20 @@ class FrameManager extends EventEmitter {
...
+    const handleTree = tree => {
+      this._onFrameNavigated(tree.frame);
+      if (!tree.childFrames)
+        return;
+
+      for (const child of tree.childFrames)
+        handleTree(child);
+    };
+
+    const { frameTree } = await this._client.send('Page.getResourceTree');
```
aslushnikov: frameTree should be fetched in `page.create` static method and passed in FrameManager constructor as an argument

lib/Browser.js
```diff
@@ -81,3 +147,61 @@ class TaskQueue {
...
+    this._targetId = targetId;
+    this._type = type;
+    /** @type {Promise<Page>} */
+    this._pagePromise = null;
+  }
+
+  /**
+   * @return {Promise<{targetId: string, type: string, title: string, url: string, attached: boolean, type: string}>}
+   */
+  async _info() {
```
aslushnikov: you should create target with targetInfo and make this sync.
You'll also need to listen to `targetInfoChanged` event.

aslushnikov (commented):
Glad to see this is happenning!

lib/Page.js
JoelEinbinder: The FrameManager needs to be listening for execution contexts before Runtime.enable is called.

lib/Page.js
```diff
@@ -36,34 +36,41 @@ class Page extends EventEmitter {
...
-    await page.goto('about:blank');
     // Initialize default page size.
     if (!appMode)
       await page.setViewport({width: 800, height: 600});
+
+    // The page might still be initializing, if so wait for the first navigation
```
aslushnikov: @JoelEinbinder could you please check with @dgozman? this looks like something that should not be happenning

lib/Browser.js
```diff
@@ -81,3 +155,51 @@ class TaskQueue {
     return result;
   }
 }
+
+class Target {
+  /**
+   * @param {Browser} browser
+   * @param {Target.TargetInfo} targetInfo
```
aslushnikov: nit: '!'

lib/Browser.js
```diff
@@ -31,6 +31,58 @@ class Browser extends EventEmitter {
...
+      return;
+    this._targets.delete(targetId);
+    this.emit(Browser.Events.TargetDestroyed, target);
+  }
+
+  /**
+   * @param {{targetInfo: Target.TargetInfo}} param0
+   */
+  async _targetInfoChanged({targetInfo}) {
+    this._ensureTargetForInfo(targetInfo)._targetInfo = targetInfo;
```
aslushnikov: do we want to emit 'targetChanged' here?

lib/Page.js
JoelEinbinder: "It is good enough" - @dgozman

aslushnikov: I don't understand why we do this here. Doesn't it make more sense in target reporting code then?

lib/Browser.js
JoelEinbinder: Let's add in a later patch.

lib/Page.js
JoelEinbinder: It is accurately reporting whats going on in the browser. it just is inconvenient for users of `Page` so we wait for first navigation.

test/test.js
```diff
@@ -2472,6 +2472,64 @@ describe('Page', function() {
...
+      expect(await closePagePromise).toBe(otherPage);
+
+      allPages = await Promise.all(browser.targets().map(target => target.page()));
+      expect(allPages).toContain(page);
+      expect(allPages).not.toContain(otherPage);
+    }));
+    it('should report when a service worker is created and destroyed', SX(async function() {
+      await page.goto(EMPTY_PAGE);
+      const createdTarget = new Promise(fulfill => browser.once('targetcreated', target => fulfill(target)));
+      const registration = await page.evaluateHandle(() => navigator.serviceWorker.register('sw.js'));
```
aslushnikov, line 2797: what is 'sw.js'? Should you add the file to the patch?

test/test.js
```diff
@@ -2472,6 +2472,64 @@ describe('Page', function() {
 
     }));
   });
+
+  describe('Target', function() {
+    it('Browser.targets should return all of the targets', SX(async function() {
```
aslushnikov, line 2751: this and the following test seem to be doing the same thing

test/test.js
```diff
@@ -2472,6 +2472,64 @@ describe('Page', function() {
...
+      // The pages will be the testing page and the original newtab page
+      const targets = browser.targets();
+      expect(targets.some(target => target.type() === 'page' &&
+        target.url() === 'about:blank')).toBeTruthy('Missing blank page');
+      expect(targets.some(target => target.type() === 'other' &&
+        target.url() === '')).toBeTruthy('Missing browser target');
```
aslushnikov, line 2757: what's the argument in `toBeTruthy`?

lib/Page.js
aslushnikov: is there any test that would help me understand what's going on?

test/test.js
JoelEinbinder: I did, it is just empty.

test/test.js
JoelEinbinder: It is the error message when the condition fails.

aslushnikov:
As discussed offline, these are the last bits remaining:
- [x] move target initialization "hack" to target creation
- [x] report targetchanged event so that URL changes are observable

aslushnikov:
For the record: this is waiting for the [upstream change](https://chromium-review.googlesource.com/c/chromium/src/+/711215) to land.

schelkun:
Is it working with latest code from master?

JoelEinbinder:
Should be good now that #1019 landed.

lib/Browser.js
```diff
@@ -31,6 +31,64 @@ class Browser extends EventEmitter {
...
+
+  /**
+   * @param {{targetInfo: Target.TargetInfo}} event
+   */
+  _targetInfoChanged(event) {
+    const target = this._ensureTargetForInfo(event.targetInfo);
+    const previousURL = target._targetInfo.url;
+    target._targetInfo = event.targetInfo;
+    if (previousURL !== event.targetInfo.url)
+      this.emit(Browser.Events.TargetURLChanged, target);
```
aslushnikov: let's call this `targetchanged`: if we're ever to expose anything else on target (e.g. target's title), it would cover the use case.

lib/Browser.js
```diff
@@ -31,6 +31,64 @@ class Browser extends EventEmitter {
...
+
+  /**
+   * @param {!Target.TargetInfo} targetInfo
+   * @return {!Target}
+   */
+  _ensureTargetForInfo(targetInfo) {
+    if (!this._targets.has(targetInfo.targetId)) {
+      const target = new Target(this, targetInfo);
+      this._targets.set(targetInfo.targetId, target);
+      this.emit(Browser.Events.TargetCreated, target);
```
aslushnikov: let's not emit in-between: exception in one of user's listeners would corrupt our state.

lib/Browser.js
```diff
@@ -80,4 +161,59 @@ class TaskQueue {
...
+
+class Target {
+  /**
+   * @param {Browser} browser
+   * @param {!Target.TargetInfo} targetInfo
+   */
+  constructor(browser, targetInfo) {
+    this._browser = browser;
+    this._targetInfo = targetInfo;
+    /** @type {!Promise<!Page>} */
```
aslushnikov: nit: promise is nullable in this case

lib/Browser.js
```diff
@@ -80,4 +161,59 @@ class TaskQueue {
...
+    this._pagePromise = null;
+  }
+
+  /**
+   * @return {?Promise<!Page>}
```
aslushnikov: let's return `!Promise<?Page>` instead of `?Promise<!Page>`

lib/Browser.js
```diff
@@ -80,4 +161,59 @@ class TaskQueue {
...
+  url() {
+    return this._targetInfo.url;
+  }
+
+  /**
+   * @return {"page"|"service_worker"|"other"}
+   */
+  type() {
+    const type = this._targetInfo.type;
+    switch (type) {
```
aslushnikov: nit: i'd rather have this as an if.
```js
const type = this._targetInfo.type;
if (type === 'page' || type === 'service_worker')
  return type;
return 'other';
```

schelkun:
So now I should use
```
  browser.on('targetcreated',  pageOpened => {
      console.log('1');
  });
```

Right?

lib/Browser.js
JoelEinbinder: Done.

lib/Browser.js
JoelEinbinder: Done.

lib/Browser.js
JoelEinbinder: Done.

lib/Browser.js
JoelEinbinder: Done.

lib/Browser.js
JoelEinbinder: Done.

lib/Browser.js
```diff
@@ -31,6 +31,61 @@ class Browser extends EventEmitter {
...
+    await connection.send('Target.setDiscoverTargets', {discover: true});
+    return browser;
+  }
+
+  /**
+   * @param {{targetInfo: Target.TargetInfo}} event
+   */
+  async _targetCreated(event) {
+    const target = this._ensureTargetForInfo(event.targetInfo);
+    await target._initializedPromise;
```
aslushnikov: this will never resolve if target was killed before initialization.

lib/Browser.js
```diff
@@ -45,8 +111,25 @@ class Browser extends EventEmitter {
    */
   async newPage() {
     const {targetId} = await this._connection.send('Target.createTarget', {url: 'about:blank'});
-    const client = await this._connection.createSession(targetId);
-    return await Page.create(client, this._ignoreHTTPSErrors, this._appMode, this._screenshotTaskQueue);
+    const target = await this._targetForId(targetId);
```
aslushnikov: Let's inline it

```js
const target = this._targets.get(targetId);
console.assert(target, 'Failed to create target for page');
await target._initializedPromise;
```

lib/Browser.js
```diff
@@ -80,4 +170,77 @@ class TaskQueue {
...
+  /**
+   * @param {!Browser} browser
+   * @param {!Target.TargetInfo} targetInfo
+   */
+  constructor(browser, targetInfo) {
+    this._browser = browser;
+    this._targetInfo = targetInfo;
+    /** @type {?Promise<!Page>} */
+    this._pagePromise = null;
+    this._wasInitialized = false;
```
aslushnikov:
```js
this._isInitialized = this._targetInfo.type !== 'page' || this._targetInfo.url !== '');

```

lib/Browser.js
```diff
@@ -80,4 +170,77 @@ class TaskQueue {
...
+    this._initializedPromise = new Promise(fulfill => this._initialized = fulfill);
+
+    this._targetInfoChanged(targetInfo);
```
aslushnikov: this would not be needed

lib/FrameManager.js
```diff
@@ -40,6 +41,20 @@ class FrameManager extends EventEmitter {
...
     this._client.on('Runtime.executionContextCreated', event => this._onExecutionContextCreated(event.context));
+
+    this._handleFrameTree(frameTree);
+  }
+
+  /**
+   * @param {{frame: Object, childFrames: ?Array}} frameTree
+   */
+  _handleFrameTree(frameTree) {
+    this._onFrameNavigated(frameTree.frame);
```
aslushnikov: this shouldn't work? Lets's attach to somthing with nested frames

lib/Browser.js
```diff
@@ -80,4 +170,77 @@ class TaskQueue {
...
+
+  /**
+   * @param {!Target.TargetInfo} targetInfo
+   */
+  _targetInfoChanged(targetInfo) {
+    const previousURL = this._targetInfo.url;
+    this._targetInfo = targetInfo;
+
+    if (!this._wasInitialized && (this._targetInfo.type !== 'page' || this._targetInfo.url !== '')) {
+      this._wasInitialized = true;
```
aslushnikov: s/_wasInitialized/_isInitialized
s/_initialzed/_initializeCallback

lib/Browser.js
```diff
@@ -31,6 +31,61 @@ class Browser extends EventEmitter {
...
+    this._targets.delete(event.targetId);
+    if (await target._initializedPromise)
+      this.emit(Browser.Events.TargetDestroyed, target);
+  }
+
+  /**
+   * @param {{targetInfo: !Target.TargetInfo}} event
+   */
+  _targetInfoChanged(event) {
+    const target = this._ensureTargetForInfo(event.targetInfo);
```
aslushnikov: assert target here; never create

lib/Browser.js
```diff
@@ -31,6 +31,61 @@ class Browser extends EventEmitter {
...
+    const browser =  new Browser(connection, ignoreHTTPSErrors, closeCallback);
+    await connection.send('Target.setDiscoverTargets', {discover: true});
+    return browser;
+  }
+
+  /**
+   * @param {{targetInfo: !Target.TargetInfo}} event
+   */
+  async _targetCreated(event) {
+    const target = this._ensureTargetForInfo(event.targetInfo);
```
aslushnikov: create target here; it should never exist before

aslushnikov (approved):

ebidel:
Wooooooot!

pirate:
Thanks everyone who worked on this, it looks great!  Really appreciate this change, it'll make life a lot easier for us when testing popups!

vivien7512:
so this great feature  is available  from which number version?

nwhitmont:
Sweet!

BlackCapCoder:
Thank you for working on this! I am making a vi-like web automation/scraping language, but it has been on hold for some time because this is absolutely essential.

manideep1111:
can you give any code sample of, how to open multiple tabs using puppeteer?

JoelEinbinder:
@manideep1111 you can use [`browser.newPage`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#browsernewpage) more than once to create more tabs.

manideep1111:
Hi JoelEinbinder Thanks for your reply, I tried with  browser.newPage but my side giving UnhandlePromiseRejection error came .
If possible plea send any example code.

manideep1111:
Thanks @JoelEinbinder for reply, I used browser.newPage my side it's not opening new tab,If you give any code example it's grateful for me.

JoelEinbinder:
@manideep1111 can you file a separate issue? Thanks!

manideep1111:
No separate issue ? If you want see my file I can send to you..Thanks

Issue #411, opened 2017-08-20
Introduce --user-data-dir as part of launcher options


vsemozhetbyt:
Use case: periodically getting some data from a site that requires authorization. A user can login once and then just use the cookies automatically.

kensoh:
Hi @vsemozhetbyt this looks like active work-in-progress here - https://github.com/GoogleChrome/puppeteer/issues/53

vsemozhetbyt:
Thank you. This would be a good detailed API, but I mean something much simpler for a user side: not even thinking about processing cookies.

ORESoftware:
I want to puppeteer to use an existing chrome session and open a tab in that session. Can someone document how? I would be happy to write an article about it on Medium.com or whatever.

Issue #527, opened 2017-08-24
using --user-data-dir=profilepath corrupts settings and remove all extensions


BenHayak:
Launching with args:
```js
	const browser = await puppeteer.launch({
	  headless: false,
	  args: ['--user-data-dir=/pathtoprofile/'],
	});
```
Will override some files and removes all extensions from the chrome profile + alerts for "settings corrupted" at chrome://settings screen.

aslushnikov:
Might be related to #411, and might be related to some headless specifics

aslushnikov:
This should be fixed on tip-of-tree.

aslushnikov:
Ah no, requires more testing.

kaytwo:
I'm still seeing this exact behavior in version 0.13.0 on OSX.

aslushnikov:
@kaytwo can you please file another issue with a step-by-step scenario? It will help us reproduce your issue locally.

kaytwo:
#1404

PR #555, opened 2017-08-26
Allow a custom userDataDir to be specified

JoelEinbinder:
#411


vsemozhetbyt:
Thank you!
Do I get this right?
1. If I set the path to a real user dir from, say, Google Chrome, it will be used properly and will not be corrupted?
2. If I set the path to an empty dir, all the necessary files will be created and will be used properly on next launches?

JoelEinbinder:
@vsemozhetbyt sounds right. I still have to look into what is going on with #527 though.

aslushnikov (approved):
Great, thanks! Let's land this after https://github.com/GoogleChrome/puppeteer/pull/566

PR #559, opened 2017-08-27
Clarify that examples need Node v8

Mohamed3on:
The examples use async/await, but Node v6 doesn't support them, so to use these examples you either need Node 8 or a transpiler (like Babel) or to use promises.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Most things that you can do manually in the browser can be done using Puppeteer!
 
 ### Installation
 
-*Puppeteer requires Node version 6.4.0 or greater*
+> *Note: Puppeteer requires at least Node v6.4.0, but the examples below use async/await which is only supported in Node v8+*
 
 To use Puppeteer in your project, run:
 ```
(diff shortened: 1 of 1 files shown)


Mohamed3on:
CLA signed!

README.md
```diff
@@ -21,7 +21,7 @@ Most things that you can do manually in the browser can be done using Puppeteer!
 
 ### Installation
 
-*Puppeteer requires Node version 6.4.0 or greater*
+> *Note: Puppeteer requires at least Node v6.4.0, but the examples below use async/await which is only supported in Node v8+*
```
vsemozhetbyt: FWIW, it seems to be [v7.6.0+](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V7.md#2017-02-21-version-760-current-italoacasas)

Mohamed3on: Updated, thanks!

aslushnikov (approved):

PR #564, opened 2017-08-27
Add support for request interception patterns

paambaati:
This PR introduces support for request interception patterns, freshly introduced in https://chromium-review.googlesource.com/c/chromium/src/+/602483

I'm not sure which release this commit landed in, as I can't find it yet on https://omahaproxy.appspot.com/

The new test fails for now, but rolling Chromium to the version that includes the commit should make it pass.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -949,6 +949,28 @@ describe('Page', function() {
       expect(requests.length).toBe(1);
       expect(requests[0].url).toBe(EMPTY_PAGE);
     }));
+    it('should work with encoded URLs', SX(async function() {
+      // The requestWillBeSent will report encoded URL, whereas interception will
+      // report URL as-is. @see crbug.com/759388
+      await page.setRequestInterceptionEnabled(true);
+      page.on('request', request => request.continue());
+      const response = await page.goto(PREFIX + '/some nonexisting page');
+      expect(response.status).toBe(404);
+    }));
+    it('should work with encoded URLs - 2', SX(async function() {
+      // The requestWillBeSent will report URL as-is, whereas interception will
+      // report encoded URL for stylesheet. @see crbug.com/759388
+      await page.setRequestInterceptionEnabled(true);
+      const requests = [];
+      page.on('request', request => {
+        request.continue();
+        requests.push(request);
+      });
+      const response = await page.goto(`data:text/html,<link rel="stylesheet" href="${PREFIX}/fonts?helvetica|arial"/>`);
+      expect(response.status).toBe(200);
+      expect(requests.length).toBe(2);
+      expect(requests[1].response().status).toBe(404);
+    }));
   });
 
   describe('Page.Events.Dialog', function() {
--- a/test/test.js
+++ b/test/test.js
@@ -810,6 +810,17 @@ describe('Page', function() {
       const response = await page.goto(EMPTY_PAGE);
       expect(response.ok).toBe(true);
     }));
+    it('should block patterns', SX(async function() {
+      await page.setRequestInterceptionEnabled(true, ['*.css']);
+      page.on('request', request => {
+        request.continue();
+      });
+      let failedRequests = 0;
+      page.on('requestfailed', event => ++failedRequests);
+      const response = await page.goto(PREFIX + '/one-style.html');
+      expect(response.ok).toBe(true);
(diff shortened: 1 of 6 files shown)


paambaati:
@googlebot I signed it!

test/test.js
```diff
@@ -810,6 +810,17 @@ describe('Page', function() {
       const response = await page.goto(EMPTY_PAGE);
       expect(response.ok).toBe(true);
     }));
+    it('should block patterns', SX(async function() {
+      await page.setRequestInterceptionEnabled(true, ['*.css']);
+      page.on('request', request => {
+        request.continue();
```
JoelEinbinder: Shouldn't this be request.abort()?

twalpole:
I'm unclear as to what this option actually does.  Does it mean the `request` event is only emitted when a request to a matching URL happens, or does it mean that request interception is only enabled for requests to matching URLS?  If it's the second, is there a way inside the `request` event handler to determine whether or not interception has been enabled or not (whether or not abort/continue need to be called)?

aslushnikov:
@paambaati thank you for the PR!

Why would you want this? The same behavior is implementable with the request interception API puppeteer has today:

```js
const blockRegexes = [
  /\.jpg$/i,
  /\.jpeg$/i,
  /\.png$/i,
];
  if (blockRegexes.some(regex => regex.test(request.url))
    request.abort();
  else
    request.continue();
});
```

test/test.js
paambaati: D'oh, you're right. Fixed in 72800b7.

paambaati:
@twalpole The `requestIntercepted` event is fired only for request URLs that match the patterns.

paambaati:
@aslushnikov I'd say this is a convenience method. If the protocol comes bundled with it, I'd rather use that instead of the boilerplate from the example.

EDIT: In addition, I believe things should be overall faster as we're not even intercepting them. I got interested after this in the [commit message](https://chromium-review.googlesource.com/c/chromium/src/+/602483) —

> _This should increase the stalled speed a bit._

twalpole:
@paambaati  Are you saying there is a new `requestIntercepted` event, because your test only uses the `request` and `requestFailed` events?  

Normally the `request` event is fired  for every request,  and if interception is enabled you need to call `abort` or `continue` on the `request` object.  If interception is not enabled then you can't call `abort` or `continue` on the`request` object.  I'm guessing what this PR actually does is automatically enable request interception based on the patterns (but not limit which requests the request event is fired for). However if there is no way in the `request` event handler to know whether interception is enabled or not, then you have to check the patterns against the url in the handler to determine whether or not `abort`/`continue` need to be called which seems highly inconvenient and defeats the point of this PR.

test/test.js
paambaati: @JoelEinbinder Also, the test would be more rounded if I have more XHRs in a page and I block specific patterns, while leaving out the rest. Should I create another HTML fixture?

paambaati:
@twalpole `requestIntercepted` is a wire protocol event, I don't think puppeteer exposes it. Sorry for the confusion!

To answer your question, pupeteer's `request` event will be fired for each intercepted request, which includes pattern-matched ones.

> However if there is no way in the request event handler to know whether interception is enabled or not

You're correct, there isn't. See https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#requestabort

paambaati:
@JoelEinbinder @aslushnikov Looks like this landed in [`62.0.3197.0`](https://storage.googleapis.com/chromium-find-releases-static/241.html#241c6b4df3f4c0b93d783ed2b5b8b448e4e2d285).

aslushnikov:
@paambaati The main use case of pattern-blocking is to block all requests of the same type, e.g. all images or all styles. There's a better way to do it using resource type: check out #590 for details.

Codewise, I don't like the suggested approach because of the following questions:
    - When is it ok to call `request.continue()` and when not?
    - How does URL blocking work with redirects?
    - How does URL blocking work with HTTPS errors?
    - How do I know if the request's URL matches the request blocking pattern?

Unless there are very good reasons for this PR and good answers to the questions, I'd rather not land this.

paambaati:
@aslushnikov Fair enough. TBH, I agree with the observation that this doesn't go well with the `request` event and `request.continue()`.

PR #565, opened 2017-08-27
Fix Request Interception

aslushnikov:
It turned out that either `Network.requestIntercepted` or
`Network.requestWillBeSent` occasionally report encoded URL.
Upstream bug: https://crbug.com/759388

As a workaround, this patch starts decoding URL's when generating request hash.

Fixes #558.

PR #566, opened 2017-08-27
Bump version to 0.10.2-alpha

aslushnikov:


JoelEinbinder (approved):

pocketjoso:
@aslushnikov I would like to test these changes, especially #565 - could you publish this version to npm? as under `alpha`, `next` or similar tag..

aslushnikov:
@pocketjoso the suggested way to try tip-of-tree is to install from github: `npm install GoogleChrome/puppeteer`. Do you have any concerns about it?

pocketjoso:
@aslushnikov I use `yarn` and usually it doesn't work well with yarn to install from github. I'll try it..

pocketjoso:
Okay not sure what the difference is, but so far this worked for me with yarn as well, so for now I'm good. 👍

pocketjoso:
#565 btw is an absolute blocker for me to use `puppeteer` in production - do you have any ETA for the next stable release?

aslushnikov:
@pocketjoso should be sometime this week

PR #567, opened 2017-08-27
Use objects instead of maps to describe headers

### LINKED ISSUE #509: setExtraHTTPHeaders: support both object and Map as `headers` argument

### LINKED ISSUE #547: Headers in Response/Request object are case sensitive
I am attempting to get the content-type of HTTP responses.

My code looks approximately like this:

```
page.on("response", response => {
  const contentType = response.headers.get("Content-Type")
})
```

Unfortunately the `response.headers` `Map` is case sensitive, so this only works if the header name is `Content-Type` (as opposed to, say, `content-type`).

I think all headers need to be lower cased before setting them in the map, or there needs to be an alternative API to retrieve headers.

The current API requires code to iterate over all headers just to normalise the case.

aslushnikov:
This patch:
- switches to objects instead of maps for headers (in Request, Response and
  page.setExtraHTTPHeaders)
- converts all header names to lower case 

Fixes #547, fixes #509

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -802,7 +802,7 @@ describe('Page', function() {
       await page.setRequestInterceptionEnabled(true);
       page.on('request', request => {
         expect(request.url).toContain('empty.html');
-        expect(request.headers.has('User-Agent')).toBeTruthy();
+        expect(request.headers['user-agent']).toBeTruthy();
         expect(request.method).toBe('GET');
         expect(request.postData).toBe(undefined);
         request.continue();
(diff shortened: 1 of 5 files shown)


test/test.js
```diff
@@ -839,8 +839,8 @@ describe('Page', function() {
     it('should amend HTTP headers', SX(async function() {
       await page.setRequestInterceptionEnabled(true);
       page.on('request', request => {
-        const headers = new Map(request.headers);
-        headers.set('foo', 'bar');
+        const headers = Object.assign({}, request.headers);
+        headers['FOO'] = 'bar';
         request.continue({ headers });
       });
       await page.goto(EMPTY_PAGE);
```
JoelEinbinder: This test doesn't seem to check that the headers were set correctly.

test/test.js
```diff
@@ -811,12 +811,12 @@ describe('Page', function() {
       expect(response.ok).toBe(true);
     }));
     it('should show custom HTTP headers', SX(async function() {
-      await page.setExtraHTTPHeaders(new Map(Object.entries({
+      await page.setExtraHTTPHeaders({
         foo: 'bar'
```
JoelEinbinder: can we test case sensitivity here?

JoelEinbinder (approved):

PR #580, opened 2017-08-28
[docs] a mistake in allowed values of `page.evaluateMedia`

ls6:
It seems like a simple mistake: allowed `mediaType` is `print` rather than `media`.
Tested on linux Ubuntu: screenshot does show the print version.


ls6:
CLA signed.

aslushnikov (approved):

aslushnikov:
Thanks for the PR!

PR #582, opened 2017-08-28
Fix launcher to not fail in case of empty arguments

aslushnikov:
The regression was introduced in
d5327e6a0fd384cd36bfd0b4a12b27ab362543a7


aslushnikov:
TBR: @JoelEinbinder

PR #590, opened 2017-08-29
Implement Request.resourceType getter

aslushnikov:
This patch introduces Request.resourceType getter that represents resource type as it was perceived by the rendering engine. Possible values are inherited as-is from protocol:
- Document, Stylesheet, Image, Media, Font, Script, TextTrack, XHR, Fetch, EventSource, WebSocket, Manifest, Other

This patch also updates an example to effectively block images.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -117,6 +117,7 @@
     + [request.headers](#requestheaders)
     + [request.method](#requestmethod)
     + [request.postData](#requestpostdata)
+    + [request.resourceType](#requestresourcetype)
     + [request.response()](#requestresponse)
     + [request.url](#requesturl)
   * [class: Response](#class-response)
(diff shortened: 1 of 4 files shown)


JoelEinbinder (approved):
I will feel safer if we have a test that tests resourceType === 'Image'

gajus:
> Document, Stylesheet, Image, Media, Font, Script, TextTrack, XHR, Fetch, EventSource, WebSocket, Manifest, Other

It really bothers me that you are not using all caps or all lowercase convention.

Now every developer (cautious of hard to catch bugs) is going to need to add a safe-guard, such as `request.resourceType.toUpperCase() === 'IMAGE'`.

ebidel:
@gajus those values come from the DT protocol: https://chromedevtools.github.io/devtools-protocol/tot/Page/#type-ResourceType

aslushnikov:
@gajus good point. Sent a PR to fix this: https://github.com/GoogleChrome/puppeteer/pull/990

PR #604, opened 2017-08-29
readme: note on how to skip chromium install

ebidel:
adds note on how to skip install of chromium.

Related to https://github.com/GoogleChrome/puppeteer/issues/603

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ yarn add puppeteer
 # or "npm i puppeteer"
 ```
 
-> **Note**: When you install Puppeteer, it downloads a recent version of Chromium (~71Mb Mac, ~90Mb Linux, ~110Mb Win) that is guaranteed to work with the API.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,7 +29,7 @@ yarn add puppeteer
 # or "npm i puppeteer"
 ```
 
-> **Note**: When you install Puppeteer, it downloads a recent version of Chromium (~71Mb Mac, ~90Mb Linux, ~110Mb Win) that is guaranteed to work with the API.
+> **Note**: When you install Puppeteer, it downloads a recent version of Chromium (~71Mb Mac, ~90Mb Linux, ~110Mb Win) that is guaranteed to work with the API. Although it's not recommend, you can skip the Chromium install with `npm i --ignore-scripts puppeteer`.
```
JoelEinbinder, line 32: recommended

aslushnikov: is this enough? If I want to deploy an app that depends on Puppeteer in its `package.json`, than I can't use the `npm i --ignore-scripts puppeteer`

ebidel:
Upon further investigation it doesn't look like this is enough. I've also
read --ignore-scripts doesn't run the scripts of package dependecies.
Haven't verified that, but it would be no good if that's the case.

aslushnikov:
@ebidel the https://github.com/GoogleChrome/puppeteer/pull/606 should satisfy the need.

PR #606, opened 2017-08-30
Skip download

aslushnikov:
This patch starts skipping chromium download when PUPPETEER_SKIP_CHROMIUM_DOWNLOAD env variable is set.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -142,8 +142,8 @@ Puppeteer is a Node library which provides a high-level API to control Chromium 
 
 Puppeteer looks for certain environment variables to aid its operations:
 
-- `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run chromium.
-- `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled chromium during installation step.
+- `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
+- `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step.
 
 ### class: Puppeteer
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -7,6 +7,7 @@
 <!-- toc -->
 
 - [Puppeteer](#puppeteer)
+  * [Environment Variables](#environment-variables)
   * [class: Puppeteer](#class-puppeteer)
     + [puppeteer.connect(options)](#puppeteerconnectoptions)
     + [puppeteer.launch([options])](#puppeteerlaunchoptions)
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -137,6 +138,12 @@
 
 Puppeteer is a Node library which provides a high-level API to control Chromium over the DevTools Protocol.
 
+### Environment Variables
+
+Puppeteer looks for certain environment variables to aid its operations:
+
+- `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run chromium.
```
ebidel: Chromium

docs/api.md
```diff
@@ -137,6 +138,12 @@
...
+- `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled chromium during installation step.
```
ebidel: same

install.js
```diff
@@ -14,6 +14,11 @@
  * limitations under the License.
  */
 
+if (process.env['PUPPETEER_SKIP_CHROMIUM_DOWNLOAD']) {
```
ebidel: any reason not to use `process.env.PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`?

install.js
```diff
@@ -14,6 +14,11 @@
...
+  console.log('**WARNING** Puppeteer is skipping chromium download because environment contains "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" variable.');
```
ebidel: Chromium

ebidel: Not sure this should say "Warning". Users opt-in to it. How about:

`console.log('**INFO** Skipping Chromium download. "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" environment variable was found.');`

ebidel (changes requested):

ebidel (approved):

JoelEinbinder (approved):

docs/api.md

docs/api.md

install.js
aslushnikov: Property access with string literals is done to trick type linter and avoid the `ERROR: PUPPETEER_SKIP_CHROMIUM_DOWNLOAD is never defined on process.env object`. That's a habbit from devtools codebase.

Done.

install.js
aslushnikov: Done.

sukrosono:
Still related to this, I am new and it take couple of minutes to figure out how to set `process.env` . I made a small changes on the documentation like [so](https://github.com/brutalcrozt/puppeteer/blob/master/docs/api.md#environment-variables), Guys, do you like the changes as a PR?

aslushnikov:
@brutalcrozt instead of adding this example, would you mind sending a PR that links to the [wiki page](https://en.wikipedia.org/wiki/Environment_variable) in the documentation? This would help others who don't know what's an environment variable.

sukrosono:
@aslushnikov [Yes](792), that was kind of response which i expected. Recently my PR are rejected because miss match the expectation, so i like to make a small comment to prevent that. :smile:  , Thanks

serotonyn:
@brutalcrozt I still can't figure out how to do it. Can i ask for your help ?

PR #619, opened 2017-08-30
Throw an error when Chromium download is failed

homaily:
Hi,

When Chromium download is failed (no internet connection for example) **downloadRevision** will not emit an error and therefore **onError** function in the install file will not be triggered and the user will not see the message:
> ERROR: Failed to download chromium ....

I'm using Node v6.11.2

### DIFF
--- a/utils/ChromiumDownloader.js
+++ b/utils/ChromiumDownloader.js
@@ -97,10 +97,12 @@ module.exports = {
       fs.mkdirSync(DOWNLOADS_FOLDER);
     return downloadFile(url, zipPath, progressCallback)
         .then(() => extractZip(zipPath, folderPath))
-        .catch(err => { throw err; })
-        .then(() => {
+        .catch(err => err)
+        .then(err => {
           if (fs.existsSync(zipPath))
             fs.unlinkSync(zipPath);
+          if (err)
+            throw err;
         });
   },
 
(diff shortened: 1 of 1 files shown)


aslushnikov:
This would play nice with #606 

Codewise, we should probably process.exit(1) to terminate the script with non-zero exit code

homaily:
We can do the **process.exit(1)** from the **onError** function

aslushnikov:
@homaily any updates on this? I'm very interested in this PR.

homaily:
@aslushnikov Sorry didn't get your point. Do you want the process.exit in the **onError** on int the place of **throw Error**?

utils/ChromiumDownloader.js
```diff
@@ -97,7 +97,7 @@ module.exports = {
       fs.mkdirSync(DOWNLOADS_FOLDER);
     return downloadFile(url, zipPath, progressCallback)
         .then(() => extractZip(zipPath, folderPath))
-        .catch(err => err)
+        .catch(err => { throw err; })
         .then(() => {
           if (fs.existsSync(zipPath))
```
aslushnikov: Unfortunately, there's no "finally" in promise chains, so I'd fake it like this:
```js
  .catch(err => err)
  .then(err => {
    if (fs.existsSync(zipPath))
      fs.unlinkSync(zipPath);
    if (err)
      throw err;
  });
```
This way we'll throw an error and cleanup zip if any

aslushnikov (commented):
Ah sorry, I got confused. I thought something else is going on here.

utils/ChromiumDownloader.js
homaily: Much better love it. I've updated my code to do that.

aslushnikov (approved):
Thanks!

homaily:
You're welcome Andrey. Appreciated.

PR #626, opened 2017-08-30
Add basic authentication support

zqsd:
This adds `page.authenticate(user, pass)` method to "fill" the authentication "window". Works for either a webpage auth or a proxy auth, and both on http and https.

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -37,6 +37,10 @@ class NetworkManager extends EventEmitter {
     this._requestHashToRequestIds = new Multimap();
     /** @type {!Multimap<string, !Object>} */
     this._requestHashToInterceptions = new Multimap();
+    /** @type {!string} */
+    this._authUsername = undefined;
+    /** @type {!string} */
+    this._authPassword = undefined;
 
     this._client.on('Network.requestWillBeSent', this._onRequestWillBeSent.bind(this));
     this._client.on('Network.requestIntercepted', this._onRequestIntercepted.bind(this));
@@ -81,6 +85,18 @@ class NetworkManager extends EventEmitter {
   }
 
   /**
+   * @param {string} username
+   * @param {string} password
+   */
+  async authenticate(username, password) {
+    if (!this._requestInterceptionEnabled)
+      await this.setRequestInterceptionEnabled(true);
+
+    this._authUser = username;
+    this._authPass = password;
+  }
+
+  /**
    * @param {!Object} event
    */
   _onRequestIntercepted(event) {
@@ -94,6 +110,14 @@ class NetworkManager extends EventEmitter {
       this._handleRequestStart(request._requestId, event.interceptionId, event.redirectUrl, event.request);
       return;
     }
+
+    if (event.authChallenge) {
+      const request = this._interceptionIdToRequest.get(event.interceptionId);
+      console.assert(request, 'INTERNAL ERROR: failed to find request for interception redirect.');
+      this._handleAuthenticate(request, event.authChallenge);
+      return;
+    }
+
     const requestHash = generateRequestHash(event.request);
     this._requestHashToInterceptions.set(requestHash, event);
     this._maybeResolveInterception(requestHash);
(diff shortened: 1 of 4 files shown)


test/test.js
```diff
@@ -750,6 +750,11 @@ describe('Page', function() {
       expect(requests.length).toBe(1);
       expect(requests[0].url).toBe(EMPTY_PAGE);
     }));
+    it('should navigate to URL with authentication set but no challenge', SX(async function() {
```
zqsd, line 753: I've been limited with testing because the SimpleServer class doesn't seem to support authentication.

Khady:
Testing with squid here and it seems to work. Thanks.

lib/NetworkManager.js
```diff
@@ -94,6 +110,14 @@ class NetworkManager extends EventEmitter {
       this._handleRequestStart(request._requestId, event.interceptionId, event.redirectUrl, event.request);
       return;
     }
+
+    if (event.authChallenge) {
```
aslushnikov, line 114: could there be `redirectStatusCode` and `authChallendge`?

lib/NetworkManager.js
```diff
@@ -114,6 +138,29 @@ class NetworkManager extends EventEmitter {
   }
 
   /**
+   * @param {!Request} request
+   * @param {!Object} authentification challenge
+   */
+  _handleAuthenticate(request, authChallenge) {
+    request._interceptionHandled = false;
```
aslushnikov, line 145: let's use `this._client.send` directly to avoid messing with user-visible object.

lib/NetworkManager.js
```diff
@@ -253,6 +300,7 @@ class Request {
       method: overrides.method,
       postData: overrides.postData,
       headers: overrides.headers,
+      authChallengeResponse: overrides.authChallengeResponse,
```
aslushnikov, line 303: let's not expose this to the user-facing api

test/test.js
aslushnikov: The testing is very important.
1. Let's add basic HTTP auth for the SimpleServer:

```js
server.setAuth(route, user, pass)
```
It should be a matter of one HTTP header.
(Make sure to reset the auth at `SimpleServer.reset` so that tests start with clean state.)

2. Let's add at least the following tests:
- Basic `page.authenticate` 
- Check what happens in case of invalid credentials
- `page.authenticate` works with cross-process navigation (look for `CROSS_PROCESS_PREFIX`)
- `page.authenticate` works with redirects
- Make sure `page.authenticate` doesn't allow calling `request.continue()`
- Make sure `page.authenticate` and `page.setRequestInterceptionEnabled(true)` work nice together

lib/Page.js
```diff
@@ -257,6 +257,25 @@ class Page extends EventEmitter {
...
+  async authenticate(username, password) {
+    return this._networkManager.authenticate(username, password);
+  }
+
+  /**
+   * @param {!Object} event
+   */
+  _onRequest(event) {
+    if (this.listenerCount(Page.Events.Request) > 0)
+      this.emit(Page.Events.Request, event);
```
aslushnikov, line 273: I might listen to the events without continuing them. This will hang the loading.

lib/NetworkManager.js
```diff
@@ -81,6 +85,18 @@ class NetworkManager extends EventEmitter {
...
   /**
+   * @param {string} username
+   * @param {string} password
+   */
+  async authenticate(username, password) {
+    if (!this._requestInterceptionEnabled)
+      await this.setRequestInterceptionEnabled(true);
+
+    this._authUser = username;
+    this._authPass = password;
```
aslushnikov, line 96: Let's do `this._authUser` and `this._authPassword`

lib/NetworkManager.js
```diff
@@ -37,6 +37,10 @@ class NetworkManager extends EventEmitter {
     this._requestHashToRequestIds = new Multimap();
     /** @type {!Multimap<string, !Object>} */
     this._requestHashToInterceptions = new Multimap();
+    /** @type {!string} */
+    this._authUsername = undefined;
+    /** @type {!string} */
+    this._authPassword = undefined;
```
aslushnikov, line 43: these are not used

aslushnikov (commented):
Thank you for the PR!

lib/NetworkManager.js
zqsd: The case were the server both sends an authentication challenge and a redirect location header should not happen and is not valid.  
But from what I understand authentication takes over redirect. To check chromium/puppeter's behavior I've forged a fake auth page that presents both auth challenge and redirect headers :
- if http status code is 401 or 407 (auth challenge), then event.redirectStatusCode is undefined
- if any other status code, then event.authChallenge is undefined

I'll still move the block before the redirect test just in case chromium changes their behavior one day.

zqsd:
Thanks for the review, I'll try to fix this according to your comments over week-end.

aslushnikov:
@zqsd any updates on this? If you don't have time, I'd like to implement this myself.

aslushnikov:
@zqsd closing this in favor of #729

PR #632, opened 2017-08-31
Update package.json with Apache License 2.0

glinford:
The license used is Apache License 2.0, so I think it makes sense to update the package.json file with the correct string as a some software use this field to approve or not packages.


glinford:
I signed it!

aslushnikov (approved):

PR #635, opened 2017-08-31
Add 'allCookies' method

MicheleLevorato:
This method returns all the cookies for the current page, even third party ones.
Based on the Chromium command Network.getAllCookies(): https://bugs.chromium.org/p/chromium/issues/detail?id=668932#c15

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -36,6 +36,7 @@
     + [page.close()](#pageclose)
     + [page.content()](#pagecontent)
     + [page.cookies(...urls)](#pagecookiesurls)
+    + [page.allCookies()](#pageallcookies)
     + [page.deleteCookie(...cookies)](#pagedeletecookiecookies)
     + [page.emulate(options)](#pageemulateoptions)
     + [page.emulateMedia(mediaType)](#pageemulatemediamediatype)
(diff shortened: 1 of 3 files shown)


aslushnikov:
Browser is a better place for this API: this should be `browser.cookies()`

MicheleLevorato:
Good point. I'll move and rename it cause it returns all the cookies in the jar. Thanks for the feedback.

aslushnikov:
@MicheleLevorato unfortunately, the Network domain is [not available](https://cs.chromium.org/chromium/src/content/browser/devtools/browser_devtools_agent_host.cc?sq=package:chromium&l=55) on the browser target today, so this would require some chromium work. @JoelEinbinder was planning to do something about it.

MicheleLevorato:
Ouch. Ok, i'll stick locally with my custom code to retrieve third party cookies, looking forward for a stable solution. Thank you, @aslushnikov!

aslushnikov:
@MicheleLevorato thanks for the PR!

aslushnikov:
@MicheleLevorato FYI: filed #645 to track progress on this.

dht:
Will this come with a compatible setCookies?

aslushnikov:
@dht there's a `browser.addCookie` in the suggestion.

Issue #625, opened 2017-08-30
Introduce page.$eval and page.$$eval


aslushnikov:
@paulirish suggested to introduce `page.$eval` and `page.$$eval` that will reduce the need for a lot one-line methods.

E.g. `elementHandle.attribute` would be replaced with a one-liner:

```js
let idAttribute = await page.$eval('div', div => div.id);
```

So the plan is:
- introduce `frame.$eval` / `frame.$$eval` / `page.$eval` / `page.$$eval`
- drop the `elementHandle.attribute`

PR #638, opened 2017-08-31
Implement page.$eval

aslushnikov:
This patch:
- implements page.$eval and frame.$eval
- drops elementHandle.attribute() method in favor of the page.$eval

References #625

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -31,7 +31,7 @@
     + [event: 'response'](#event-response)
     + [page.$(selector)](#pageselector)
     + [page.$$(selector)](#pageselector)
-    + [page.$eval(selector, pageFunction)](#pageevalselector-pagefunction)
+    + [page.$eval(selector, pageFunction[, ...args])](#pageevalselector-pagefunction-args)
     + [page.addScriptTag(url)](#pageaddscripttagurl)
     + [page.click(selector[, options])](#pageclickselector-options)
     + [page.close()](#pageclose)
@@ -95,7 +95,7 @@
   * [class: Frame](#class-frame)
     + [frame.$(selector)](#frameselector)
     + [frame.$$(selector)](#frameselector)
-    + [frame.$eval(selector, pageFunction)](#frameevalselector-pagefunction)
+    + [frame.$eval(selector, pageFunction[, ...args])](#frameevalselector-pagefunction-args)
     + [frame.addScriptTag(url)](#frameaddscripttagurl)
     + [frame.childFrames()](#framechildframes)
     + [frame.evaluate(pageFunction, ...args)](#frameevaluatepagefunction-args)
@@ -324,7 +324,7 @@ The method runs `document.querySelectorAll` within the page. If no elements matc
 
 Shortcut for [page.mainFrame().$$(selector)](#frameselector-1).
 
-#### page.$eval(selector, pageFunction, ...args)
+#### page.$eval(selector, pageFunction[, ...args])
 - `selector` <[string]> Selector to query page for
 - `pageFunction` <[function]> Function to be evaluated in browser context
 - `...args` <...[Serializable]> Arguments to pass to `pageFunction`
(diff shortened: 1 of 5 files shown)


JoelEinbinder:
Can this take `...args` like the old `page.$`?

docs/api.md
```diff
@@ -323,6 +324,17 @@ The method runs `document.querySelectorAll` within the page. If no elements matc
...
+#### page.$eval(selector, pageFunction)
+- `selector` <[string]> Selector to query page for
+- `pageFunction` <[function]> Function to be evaluated in browser context
+- returns: <[Promise]<[Serializable]>> Promise which resolves to the return value of `pageFunction`
+
+This method runs `document.querySelector` within the page and passes it as the first argument to `pageFunction`. If there's no element matching `selector`, the method throws an error.
+
+If `pageFunction` returns a [Promise], then `page.$eval` would wait for the promise to resolve and return it's value.
+
+Shortcut for [page.mainFrame().$eval(selector, pageFunction)](#frameevalselector-pageFunction).
```
vsemozhetbyt: #frameevalselector-pageFunction -> #frameevalselector-pagefunction  (F => f) ?

paulirish:
Could use an  example.. maybe one of these?

```js
const searchValue = await page.$eval('#search', el => el.value);

const preloadHref = await page.$eval('link[rel=preload]', el => el.href);

const html = await page.$eval('.main-container', e => e.outerHTML);
```

docs/api.md
ebidel: Maybe add some useful examples:

```js
const isHidden = await page.$eval('.hidden', el => el.hasAttribute('hidden'));
const val = await page.$eval('input[type="text"]', el => el.value);
const id = await page.$eval('div.item', el => el.dataset.id);
```

aslushnikov:
@JoelEinbinder sure. Do you have an example where arguments would be useful?

docs/api.md

docs/api.md
```diff
@@ -1056,6 +1076,23 @@ The method queries frame for the selector. If there's no such element within the
 
 The method runs `document.querySelectorAll` within the frame. If no elements match the selector, the return value resolve to `[]`.
 
+#### frame.$eval(selector, pageFunction)
```
vsemozhetbyt: \+ `, ...args` ?

docs/api.md
```diff
@@ -31,6 +31,7 @@
     + [event: 'response'](#event-response)
     + [page.$(selector)](#pageselector)
     + [page.$$(selector)](#pageselector)
+    + [page.$eval(selector, pageFunction)](#pageevalselector-pagefunction)
```
vsemozhetbyt: 1. \+ `, ...args` ?
2. #pageevalselector-pagefunction-args ?

docs/api.md
```diff
@@ -94,6 +95,7 @@
   * [class: Frame](#class-frame)
     + [frame.$(selector)](#frameselector)
     + [frame.$$(selector)](#frameselector)
+    + [frame.$eval(selector, pageFunction)](#frameevalselector-pagefunction)
```
vsemozhetbyt: 1. \+ `, ...args` ?
2. #frameevalselector-pagefunction-args ?

lib/FrameManager.js
```diff
@@ -197,6 +197,21 @@ class Frame {
 
   /**
    * @param {string} selector
+   * @param {function()|string} pageFunction
+   * @param {!Array<*>} args
+   * @return {!Promise<?ElementHandle>}
```
JoelEinbinder: This does not return an element handle. It returns !Promise<*>

lib/Page.js
```diff
@@ -148,6 +148,16 @@ class Page extends EventEmitter {
 
   /**
    * @param {string} selector
+   * @param {function()|string} pageFunction
+   * @param {!Array<*>} args
+   * @return {!Promise<?ElementHandle>}
```
JoelEinbinder: Same as above.

lib/FrameManager.js
aslushnikov: Done.

docs/api.md
```diff
@@ -323,6 +324,25 @@ The method runs `document.querySelectorAll` within the page. If no elements matc
 
 Shortcut for [page.mainFrame().$$(selector)](#frameselector-1).
 
+#### page.$eval(selector, pageFunction[, ...args])
+- `selector` <[string]> Selector to query page for
```
JoelEinbinder: [selector]

JoelEinbinder (approved):

Issue #569, opened 2017-08-28
how to fire touch event in page.emulate(iPhone)


zenxds:
I tried mouse.down but not fire touch event，what should i do

```
const browser = await puppeteer.launch()
const page = await browser.newPage()
await page.emulate(iPhone)

await page.mouse.down({ button: 'left' })
```

zenxds:
I tried to enable `Emulation.setEmitTouchEventsForMouse`, touchevent fired but got another issue 

```
await page._client.send('Emulation.setEmitTouchEventsForMouse', {
    enabled: true
  })

await mouse.down({
    button: 'left'
})
```

`mouse.down` will hang up

zenxds:
Thanks a lot. And what i want more is api like `touchScreen.touchStart` `touchScreen.touchMove` `touchScreen.touchEnd` to  emulate drag an element。Could support for it ?

JoelEinbinder:
I'm working on it as part of #158

NanYeYe:
> I tried to enable `Emulation.setEmitTouchEventsForMouse`, touchevent fired but got another issue
> 
> ```
> await page._client.send('Emulation.setEmitTouchEventsForMouse', {
>     enabled: true
>   })
> 
> await mouse.down({
>     button: 'left'
> })
> ```
> 
> `mouse.down` will hang up

@zenxds 请问这个挂起的问题有解吗？

Issue #568, opened 2017-08-28
Can't click mobile version UI (touch event only)


Hello2Jason:
Is there a way to click a mobile UI which only support touch event?
like this web page: https://h5.m.taobao.com/#index
thank you so much!

Issue #158, opened 2017-07-27
Touch input events


JoelEinbinder:
Simulating touch events is difficult.

For tapping:
`Input.synthesizeTapGesture` appears to only work on android
`Input.dispatchTouchEvent` doesn't click events like normal tapping would
`Input.emulateTouchFromMouseEvent` fires mouse events, but only after an undefined delay.

The best way to emulate touch right now appears to be setting the device mode to iPhone, which sends `touchStart` and `touchEnd` along with every mouse press.

LHoin:
Why not just provide API like touchScreen.touchStart/touchMove/touchEnd, so we can do Swipe by combining these methods ?

zhgqthomas:
@LHoin I forked a version that add touch move/down/up in TouchScreen. But it using Emulation.setEmitTouchEventsForMouse which still in EXPERIMENTAL.

Check out this commit: https://github.com/zhgqthomas/puppeteer/commit/9f54be7bd5937c4224e10554b76270ce7357ef8a

RainUUU:
> @LHoin I forked a version that add touch move/down/up in TouchScreen. But it using Emulation.setEmitTouchEventsForMouse which still in EXPERIMENTAL.
> 
> Check out this commit: [zhgqthomas@9f54be7](https://github.com/zhgqthomas/puppeteer/commit/9f54be7bd5937c4224e10554b76270ce7357ef8a)

Will event. isTrusted be true in the browser in this form of simulation?

RainUUU:
> @LHoin I forked a version that add touch move/down/up in TouchScreen. But it using Emulation.setEmitTouchEventsForMouse which still in EXPERIMENTAL.
> 
> Check out this commit: [zhgqthomas@9f54be7](https://github.com/zhgqthomas/puppeteer/commit/9f54be7bd5937c4224e10554b76270ce7357ef8a)

I tried your code, called tap method, and the page scrolled, but the JS in the page could not  listen the touchmove event.

js code:
document.addEventListener("touchmove",function(e){
 	console.log("touchmove:");
    console.log(e);
},false);

zhgqthomas:
@RainUUU I tested the codes. And the `touch start`, `touch move` and `touch end` event all get called. Maybe you can look into the codes again. And the tap method will not trigger the touchmove event

danilogacevic:
@zhgqthomas So, you trigger `touch move` event with `mouse down` + `mouse move` ?

PR #639, opened 2017-08-31
TouchScreen.tap

JoelEinbinder:
This adds a touchScreen input object with a `tap` method, and an equivalent `tap` on elementHandle. In the future touchScreen with have methods like `swipe`, `pinch` and `zoom`. The other methods are waiting on some backend work to make them less racy. 

#568 #569 #158

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -66,6 +66,7 @@
     + [page.setUserAgent(userAgent)](#pagesetuseragentuseragent)
     + [page.setViewport(viewport)](#pagesetviewportviewport)
     + [page.title()](#pagetitle)
+    + [page.touchScreen](#pagetouchscreen)
     + [page.tracing](#pagetracing)
     + [page.type(text, options)](#pagetypetext-options)
     + [page.url()](#pageurl)
@@ -83,6 +84,9 @@
     + [mouse.down([options])](#mousedownoptions)
     + [mouse.move(x, y, [options])](#mousemovex-y-options)
     + [mouse.up([options])](#mouseupoptions)
+  * [class: TouchScreen](#class-touchscreen)
+    + [touchScreen.swipe(x, y, toX, toY)](#touchscreenswipex-y-tox-toy)
+    + [touchScreen.tap(x, y)](#touchscreentapx-y)
   * [class: Tracing](#class-tracing)
     + [tracing.start(options)](#tracingstartoptions)
     + [tracing.stop()](#tracingstop)
@@ -113,6 +117,7 @@
     + [elementHandle.dispose()](#elementhandledispose)
     + [elementHandle.evaluate(pageFunction, ...args)](#elementhandleevaluatepagefunction-args)
     + [elementHandle.hover()](#elementhandlehover)
+    + [elementHandle.tap()](#elementhandletap)
     + [elementHandle.uploadFile(...filePaths)](#elementhandleuploadfilefilepaths)
   * [class: Request](#class-request)
     + [request.abort()](#requestabort)
@@ -783,6 +788,9 @@ In the case of multiple pages in a single browser, each page can have its own vi
 
 Shortcut for [page.mainFrame().title()](#frametitle).
 
+#### page.touchScreen
+- returns: <[TouchScreen]>
+
 #### page.tracing
 - returns: <[Tracing]>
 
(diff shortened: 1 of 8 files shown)


lib/ElementHandle.js
```diff
@@ -112,6 +114,10 @@ class ElementHandle {
     return await this.evaluate((element, key) => element.getAttribute(key), key);
   }
 
+  async tap() {
```
aslushnikov: jsdoc

docs/api.md
```diff
@@ -1237,6 +1251,12 @@ The element will be passed as the first argument to `pageFunction`, followed by
 This method scrolls element into view if needed, and then uses [page.mouse](#pagemouse) to hover over the center of the element.
 If the element is detached from DOM, the method throws an error.
 
+#### elementHandle.tap()
+- returns: <[Promise]> Promise which resolves when the element is successfully tapped. Promise gets rejected if the element is detached from DOM.
+
+This method scrolls element into view if needed, and then uses [touchScreen.tap](#touchscreentapx-y) to tap in the center of the element.
```
aslushnikov: can we explain what's a tap?

docs/api.md
```diff
@@ -81,6 +82,8 @@
     + [mouse.down([options])](#mousedownoptions)
     + [mouse.move(x, y, [options])](#mousemovex-y-options)
     + [mouse.up([options])](#mouseupoptions)
+  * [class: TouchScreen](#class-touchscreen)
```
aslushnikov: can we also have `page.tap()` similarly to `page.click()`?

docs/api.md
```diff
@@ -64,6 +64,7 @@
     + [page.setUserAgent(userAgent)](#pagesetuseragentuseragent)
     + [page.setViewport(viewport)](#pagesetviewportviewport)
     + [page.title()](#pagetitle)
+    + [page.touchScreen](#pagetouchscreen)
```
aslushnikov: the page.touchScreen feels slightly weird. Paul, any naming ideas? @paulirish

test/assets/input/touches.html
```diff
@@ -0,0 +1,35 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Touch test</title>
+  </head>
+  <body>
+    <script src="mouse-helper.js"></script>
+    <button onclick="clicked();">Click target</button>
+    <script>
+      window.result = [];
```
aslushnikov, line 10: don't we have the same asset for mouse? Can we unify the two, simply recording all the events for the button?

test/test.js
```diff
@@ -1497,6 +1497,18 @@ describe('Page', function() {
         [200, 300]
       ]);
     }));
+    it('should tap the button', SX(async function() {
+      await page.goto(PREFIX + '/input/button.html');
```
aslushnikov, line 1513: is is actually ok to send touches to the non-touch viewport?

aslushnikov (commented):
overall looks good!

test/test.js
JoelEinbinder: It seems fine.

docs/api.md
JoelEinbinder: we can do `page.touchscreen`, all lowercase because apparently [touchscreen](https://en.wikipedia.org/wiki/Touchscreen) is one word

docs/api.md
MicheleLevorato: A concise and effective definition: https://api.jquerymobile.com/tap

test/assets/input/touches.html
JoelEinbinder: There is a similar one for keyboard, but it has a textarea.

test/test.js
```diff
@@ -1509,6 +1509,17 @@ describe('Page', function() {
...
+      await page.tap('button');
+      expect(await page.evaluate(() => result)).toBe('Clicked');
+    }));
+    it('should report touches', SX(async function() {
+      await page.goto(PREFIX + '/input/touches.html');
+      const button = await page.$('button');
+      await button.tap();
+      expect(await page.evaluate(() => getResult())).toEqual(['Touchstart: 0', 'Touchend: 0']);
```
aslushnikov, line 1521: Out of curiosity, how's `getResult()` different from `result`? You use one in the first test and other in the second

aslushnikov (approved):
Great patch, thanks.

test/test.js
JoelEinbinder: it clears the result

LLLLLamHo:
I am waiting for swipe function，It may only be used now `evaluate` function in the browser use JavaScript to simulate swipe function.

Genuifx:
@JoelEinbinder thanks for your payout, how about the swipe, pinch and zoom events? when are they will be released?

KartikKumarSahoo:
@JoelEinbinder Is swipe ready yet?

MTTTM:
> @JoelEinbinder Is swipe ready yet?

it looks like no yet!!

PR #641, opened 2017-08-31
Remove promise jsdoc

JoelEinbinder:
Removes `@return {!Promise}` for functions already marked as async.


ebidel:
Should we keep the types for future closure type checking?

JoelEinbinder:
nope, it is implied by async.

aslushnikov (approved):

PR #646, opened 2017-09-01
Add setBlockedURLs

hensansi:
[Network.setBlockedURLs](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-setBlockedURLs)
The interception looks too expensive when used with several workers. Looking up with the timeline/tracing I can see that every request can take initially 100-300ms longer** 

The browser blocker should be more efficient and eventually I will add patterns to the interception since I still need to modify some specific requests. But for this I will create another PR.
https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-setRequestInterceptionEnabled

I never did this kind of documentation so feel free to point me on the right direction.

** I am still trying to check this issue, or if it is even real. If the same image is requested more than once, like on the grid.html the previous implemention looks to be faster. With a lot of different unique resources I think I get a better result with the blockURLs.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -58,6 +58,7 @@
     + [page.press(key[, options])](#pagepresskey-options)
     + [page.reload(options)](#pagereloadoptions)
     + [page.screenshot([options])](#pagescreenshotoptions)
+    + [page.setBlockedURLs(urls)](#pagesetblockedURLs)
     + [page.setContent(html)](#pagesetcontenthtml)
     + [page.setCookie(...cookies)](#pagesetcookiecookies)
     + [page.setExtraHTTPHeaders(headers)](#pagesetextrahttpheadersheaders)
@@ -704,6 +705,10 @@ Shortcut for [`keyboard.down`](#keyboarddownkey-options) and [`keyboard.up`](#ke
         - `height` <[number]> height of clipping area
     - `omitBackground` <[boolean]> Hides default white background and allows capturing screenshots with transparency. Defaults to `false`.
 - returns: <[Promise]<[Buffer]>> Promise which resolves to buffer with captured screenshot
+
+#### page.setBlockedURLs(urls)
+- `urls` <[Array]<[url]>> URL patterns to block. Wildcards ('*') are allowed.
+- returns: <[Promise]>
 
 #### page.setContent(html)
 - `html` <[string]> HTML markup to assign to the page.
--- a/docs/api.md
+++ b/docs/api.md
@@ -58,7 +58,7 @@
     + [page.press(key[, options])](#pagepresskey-options)
     + [page.reload(options)](#pagereloadoptions)
     + [page.screenshot([options])](#pagescreenshotoptions)
-    + [page.setBlockedURLs(urls)](#pagesetblockedURLs)
+    + [page.setBlockedURLs(urls)](#pagesetblockedurlsurls)
     + [page.setContent(html)](#pagesetcontenthtml)
     + [page.setCookie(...cookies)](#pagesetcookiecookies)
     + [page.setExtraHTTPHeaders(headers)](#pagesetextrahttpheadersheaders)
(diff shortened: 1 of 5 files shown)


docs/api.md
```diff
@@ -58,6 +58,7 @@
     + [page.press(key[, options])](#pagepresskey-options)
     + [page.reload(options)](#pagereloadoptions)
     + [page.screenshot([options])](#pagescreenshotoptions)
+    + [page.setBlockedURLs(urls)](#pagesetblockedURLs)
```
vsemozhetbyt: it seems this should be `#pagesetblockedurlsurls`

hensansi: Fixed, next time I will check what is the generated anchor :)

hensansi:
~~There is an issue with failed requests retrying that I am trying to solve. (requestsFailed = 490)~~

I just realized that it will retry for each number on the grid to load the image.

aslushnikov:
Hey @hensansi, thank you for the PR

We used to have `Page.setBlockedURLs` and [removed it](https://github.com/GoogleChrome/puppeteer/commit/5ba6621cde5d0721f017fdb6befde6f8d9fb87a6) because underlying `Network.setBlockedURLs` is deprecated.

To me it's also unclear how blocked urls play with request interception, e.g.:
- is there an ability to intercept before request blocking?
- if interception kicks in before request blocking, how do I know the request will be blocked?

> Looking up with the timeline/tracing I can see that every request can take initially 100-300ms longer**
> ** I am still trying to check this issue, or if it is even real.

Any updates on this? Unless request interception is terribly slow, I wouldn't introduce blocked URLs since it adds up a lot of complexity.

examples/block-images.js
```diff
@@ -22,13 +22,7 @@ const puppeteer = require('puppeteer');
...
 const browser = await puppeteer.launch();
 const page = await browser.newPage();
-await page.setRequestInterceptionEnabled(true);
-page.on('request', request => {
-  if (request.resourceType === 'Image')
-    request.abort();
-  else
-    request.continue();
-});
+await page.setBlockedURLs(['.png', '.svg', '.jpg']);
```
aslushnikov, line 25: the old version is actually much more reliable. This version lacks a lot of image formats, e.g. `jpeg` and `webp`

hensansi: I am aware of that. I hit some issues with the interception, it was hitting broken promises when I started to use more than 3 workers. I will do more profiling today to see if I hit the same issue with the setBlockedURLs.

How do you know that 'Network.setBlockedURLs' is deprecated? I only have the tip of the documentation to rely on and for me it is strange that is deprecated since they added this recently to the inspector/network tool (before you add to activate it through the 'press shift 6 times')

hensansi:
I am going to close the PR since on the timeline the request look like they are waiting longer but after a wrk2 test I got similar results with setBlockedRequests and the interception.

I have one question before closing it, my original quest to bring this functionality was because I am hitting a broken pipe with the interception when I use it with multiple workers/tabs. (I worked with branch on 'Remove promise jsdoc' [42fa3af659] commit)

After testing with the setBlockedURLs I am not seeing this problem, this only happens with the interception. Some pages are rendered too soon, meaning that the body is smaller than it should, the page is being closed and maybe something is still happening on chrome. If I have a global variable so the prerender waits tills is finished/the variable is true I don't see the broken pipe issue. How can I debug this properly?

examples/block-images.js
aslushnikov: > How do you know that 'Network.setBlockedURLs' is deprecated? I only have the tip of the documentation to rely on and for me it is strange that is deprecated since they added this recently to the inspector/network tool (before you add to activate it through the 'press shift 6 times')

@hensansi The `Network.setBlockedURLs` is marked as experimental, which means there's no guarantee the method will exist in the upcoming versions of chrome. However, you can safely use the non-experimental methods and domains - we're committed to maintain them.

There's no deprecation process for experimental methods - so there's no way for you to know that this method would be removed. The reason I know is because I work on Chrome DevTools team.

If you absolutely need URL patterns, we've added them as an argument to the `Network.setRequestInterceptionEnabled` method. It's experimental as well, but there are good chances it'll be stabilized.

aslushnikov:
> I have one question before closing it, my original quest to bring this functionality was because I am hitting a broken pipe with the interception when I use it with multiple workers/tabs.

@hensansi This sounds like a bug with request interception. Please, file an issue with repro steps so that we can repro and fix it.

hensansi:
Thanks I will do that.

Issue #563, opened 2017-08-27
`page.screenshot()` rejects when slowMo >= 1000 and https cert is self-signed even with `ignoreHTTPSErrors`


jessehattabaugh:
I have a `webpack-dev-server` setup that uses https with a self-signed cert. I set `ignoreHTTPSErrors: true` when I create the Browser instance, and the page loads no problem. But when I try to take a screenshot I get an `unhandled promise rejection` error. It's strange because I make the call to `page.screenshot()` in a try/catch, but for some reason the rejection doesn't get caught.

```js
    try {
        const browser = await puppeteer.launch({
            ignoreHTTPSErrors: true,
            slowMo: 1000,
        });
        const page = await browser.newPage();
        await page.goto(serverUrl);
        await page.screenshot('screenshots/homepage.png');
        browser.close();
    } catch (err) {
        console.error(err);
    }
```

```
(node:29232) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Protocol error (Security.handleCertificateError): Target closed.
(node:29232) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:29232) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): AssertionError [ERR_ASSERTION]: false == true
```

aslushnikov:
Looks like the page crashed. I cannot repro this though - would you mind creating a minimalistic repro that I'd be able to run?

Also, running with `headless: false` and `dumpio: true` might give more insights:
```js
const browser = await puppeteer.launch({
  ignoreHTTPSErrors: true,
  dumpio: true,
  headless: false
});
```

jessehattabaugh:
What makes you think that the page crashed? I'm new at this so I don't know how to read the errors yet.

I added `dumpio: true` and I got some new messages

```
DevTools listening on ws://127.0.0.1:51202/devtools/browser/2aff5bce-154c-418a-adf4-52ea6f36960e
[40354:775:0827/153532.983796:ERROR:CONSOLE(1402)] "Uncaught (in promise) Error: Could not resolve node to reveal.", source: chrome-devtools://devtools/bundled/elements/elements_module.js (1402)
2017-08-27 15:35:34.018 Chromium Helper[40374:502869] Couldn't set selectedTextBackgroundColor from default ()
```

[...]

PR #680, opened 2017-09-05
Page._onCertificateError skips error handling when _ignoreHTTPSErrors is false

jessehattabaugh:
Backwards boolean. Happens to the best of us :-)

Fixes issue #563

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -148,7 +148,7 @@ class Page extends EventEmitter {
    * @param {!Object} event
    */
   _onCertificateError(event) {
-    if (!this._ignoreHTTPSErrors)
+    if (this._ignoreHTTPSErrors)
       return;
     this._client.send('Security.handleCertificateError', {
       eventId: event.eventId,


JoelEinbinder:
The original code looks correct to me. If we are set to ignore https errors, then it sends `"continue"` to every https error.

jessehattabaugh:
It's sending an error when I said to ignore errors. How's that right?

jessehattabaugh:
The first time I tested this I swear it fixed the issue, but now all I get is a timeout error

```Error: Navigation Timeout Exceeded: 30000ms exceeded
    at NavigatorWatcher.waitForNavigation (/Users/jesse/Documents/GitHub/puppeteer-issue/node_modules/puppeteer/lib/NavigatorWatcher.js:74:13)
    at <anonymous>```

**Comment by aslushnikov** [comment]
@jessehattabaugh the protocol's [`Security.handleCertificateError`](https://chromedevtools.github.io/devtools-protocol/tot/Security/#method-handleCertificateError) should be called to ignore HTTPS errors, not vice versa.

To get a better idea of what's going on under the hood, try running unit  tests and explore protocol output:
```
DEBUG=*session npm run unit
```

jessehattabaugh:
I see, I misunderstood

PR #682, opened 2017-09-05
Update api.md

pantaluna:
@changes The options param should not be enclosed in an array. Change puppeteer.launch([options]) to puppeteer.launch(options) 

@example let browser = yield puppeteer.launch({
        headless: appConfig.PUPPETEER_IS_HEADLESS_MODE,
        args: ['--allow-sandbox-debugging', '--disable-setuid-sandbox', '--no-sandbox', '--no-zygote'],
        dumpio: true
    });

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -10,7 +10,7 @@
   * [Environment Variables](#environment-variables)
   * [class: Puppeteer](#class-puppeteer)
     + [puppeteer.connect(options)](#puppeteerconnectoptions)
-    + [puppeteer.launch([options])](#puppeteerlaunchoptions)
+    + [puppeteer.launch(options)](#puppeteerlaunchoptions)
   * [class: Browser](#class-browser)
     + [browser.close()](#browserclose)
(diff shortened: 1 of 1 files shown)


pantaluna:
I signed it!
________________________________

vsemozhetbyt:
If I get this right, `[options]` does not mean array, it means that options argument is not mandatory, i.e. a user can just call `puppeteer.launch()`.

pantaluna:
It was confusing to me. I just copied the valid JS syntax [{...}] as is in my project (and did not work as it turned out after some debugging).

vsemozhetbyt:
Well, it seems this is a common practice. See:

`page.$eval(selector, pageFunction[, ...args])`
`page.click(selector[, options])`
`page.press(key[, options])`
`page.screenshot([options])`
and so on...

See also function signatures in any Node.js doc, for example in the [assert.html:](https://nodejs.org/api/assert.html) `assert(value[, message])` etc.

pantaluna:
I see. Never mind then. I got it working for me.

PR #691, opened 2017-09-05
Add Jest test example

trevordmiller:
fyi @cpojer

### DIFF
--- a/examples/jest.js
+++ b/examples/jest.js
@@ -1,0 +1,12 @@
+const puppeteer = require('puppeteer')
+
+test('sets the browser title', async () => {
+  const browser = await puppeteer.launch()
+
+  const page = await browser.newPage()
+  await page.goto('http://localhost:3000')
+  const pageTitle = await page.title()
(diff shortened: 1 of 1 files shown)


trevordmiller:
I signed it!

trevordmiller:
My only problem with this approach is that the debugging options don't launch non-headless Chrome when running through Jest; ie:

```
  headless: false,
  slowMo: 250 // slow down by 250ms
});
```

@cpojer, any idea how to get that working?

aslushnikov:
Hey @trevordmiller,

Why do you want this to be part of puppeteer rather then jest?

trevordmiller:
@aslushnikov I thought it could be helpful to have some examples of using puppeteer for end-to-end tests; but if that's not what the `examples/` directory is for then we can close this :)

JoelEinbinder:
This looks pretty straight forward from a Jest perspective. I'd like to avoid having an example for every different testing library in this project. The debugging story with Jest is interesting. If there is something we can do to make that smoother feel free to file an issue.

PR #692, opened 2017-09-05
support npm proxy configuration for the installation script

### LINKED ISSUE #648: Proxy settings should be taken from npm config before environment variables
Currently when installing behind a proxy to get the chromium downloader to respect the proxy settings you have to set the environment variables for `HTTP_PROXY` and `HTTPS_PROXY`.

For me personally I would expect the proxy settings to be taken from the npm config
e.g. 
```
npm config get https-proxy
npm config get proxy
```
And then to fall back on the environment variables if needed.

aslushnikov:
Fixes #648

### DIFF
--- a/install.js
+++ b/install.js
@@ -30,8 +30,12 @@ if (revisionInfo.downloaded)
   return;
 
 // Override current environment proxy settings with npm configuration, if any.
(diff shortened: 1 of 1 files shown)


JoelEinbinder (approved):

aslushnikov:
For the future reference: this is done similarly to the phantomjs-prebuilt: https://github.com/Medium/phantomjs/blob/a98231b3f689607eb12da565179696476a839534/install.js#L154

PR #693, opened 2017-09-06
Launcher: force kill chrome in case of working with temporary profile.

### LINKED ISSUE #527: using --user-data-dir=profilepath corrupts settings and remove all extensions

aslushnikov:
This patch:
- in case of user-supplied profile directory, starts gracefully closing
  chrome so that it doesn't corrupt profile
- in case of temp profile directory, force-closes chrome and proactively
  tries to remove temp directory to avoid littering

Fixes #527.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -104,10 +104,23 @@ class Launcher {
 
     function killChrome() {
       helper.removeEventListeners(listeners);
-      if (process.platform === 'win32')
-        childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T /F`);
-      else
-        process.kill(-chromeProcess.pid);
+      if (temporaryUserDataDir) {
+        // Force kill chrome.
+        if (process.platform === 'win32')
+          childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T /F`);
+        else
+          process.kill(-chromeProcess.pid, 'SIGKILL');
+        // Attempt to remove temporary profile directory to avoid littering.
+        try {
+          removeSync(temporaryUserDataDir);
+        } catch (e) { }
+      } else {
+        // Terminate chrome gracefully.
+        if (process.platform === 'win32')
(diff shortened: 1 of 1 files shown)


aslushnikov:
@JoelEinbinder could you please check how this behaves on Windows?

s-bach:
Works on Windows with `taskkill /pid ${chromeProcess.pid}` (without `/T`) instead of `taskkill /pid ${chromeProcess.pid} /T`.

lib/Launcher.js
```diff
@@ -104,10 +104,23 @@ class Launcher {
...
-        process.kill(-chromeProcess.pid);
+      if (temporaryUserDataDir) {
+        // Force kill chrome.
+        if (process.platform === 'win32')
+          childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T /F`);
+        else
+          process.kill(-chromeProcess.pid, 'SIGKILL');
+        // Attempt to remove temporary profile directory to avoid littering.
+        try {
+          removeSync(temporaryUserDataDir);
```
JoelEinbinder, line 115: shouldn't the 'close' event handle this? Or we try just in case the close never happens?

lib/Launcher.js
```diff
@@ -104,10 +104,23 @@ class Launcher {
...
+        } catch (e) { }
+      } else {
+        // Terminate chrome gracefully.
+        if (process.platform === 'win32')
+          childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T`);
```
JoelEinbinder, line 120: This didn't work for me on Windows.
```
This process can only be terminated forcefully (with /F option)
```
I'll send a PR with what worked for me.

JoelEinbinder (changes requested):
It doesn't work :(

aslushnikov:
Closing in favor of #700

PR #700, opened 2017-09-06
Launcher: Close gracefully when a userDataDir is specified

### LINKED ISSUE #527: using --user-data-dir=profilepath corrupts settings and remove all extensions

JoelEinbinder:
My version of #693 

In order to make tests work, I needed to add a `'close'` event to `Browser`. This broke the doclint, which didn't like browser having a `'close'` event and a `.close` method. So that is why this patch is bigger than it should be.

Fixes #527

### DIFF
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -22,7 +22,7 @@ class Browser extends EventEmitter {
   /**
    * @param {!Connection} connection
    * @param {boolean} ignoreHTTPSErrors
-   * @param {function()=} closeCallback
+   * @param {function():Promise=} closeCallback
    */
   constructor(connection, ignoreHTTPSErrors, closeCallback) {
     super();
@@ -56,16 +56,11 @@ class Browser extends EventEmitter {
     return version.product;
   }
 
-  close() {
+  async close() {
     this._connection.dispose();
-    this._closeCallback.call(null);
+    await this._closeCallback.call(null);
   }
 }
-
-/** @enum {string} */
-Browser.Events = {
-  Close: 'close'
-};
 
 module.exports = Browser;
 helper.tracePublicAPI(Browser);
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -16,14 +16,16 @@
 
 const {helper} = require('./helper');
 const Page = require('./Page');
+const EventEmitter = require('events');
 
-class Browser {
+class Browser extends EventEmitter {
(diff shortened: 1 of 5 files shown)


lib/Launcher.js
```diff
@@ -96,18 +101,32 @@ class Launcher {
...
-        process.kill(-chromeProcess.pid);
+      if (temporaryUserDataDir) {
+        // Force kill chrome.
+        if (process.platform === 'win32')
+          childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T /F`);
+        else
+          process.kill(-chromeProcess.pid, 'SIGKILL');
+        // Attempt to remove temporary profile directory to avoid littering.
+        try {
+          removeSync(temporaryUserDataDir);
```
JoelEinbinder, line 123: @aslushnikov I copied this from your previous PR. I'm not sure if we need it.

aslushnikov: That's to handle "process.exit " case.

test/test.js
```diff
@@ -120,7 +120,9 @@ describe('Browser', function() {
     const options = Object.assign({userDataDir}, defaultBrowserOptions);
     const browser = await puppeteer.launch(options);
     expect(fs.readdirSync(userDataDir).length).toBeGreaterThan(0);
+    const closePromise = new Promise(fulfill => browser.once('close', fulfill));
```
aslushnikov: let's make `browser.close()` return a promise

lib/Browser.js
```diff
@@ -60,6 +62,11 @@ class Browser {
   }
 }
 
+/** @enum {string} */
+Browser.Events = {
```
aslushnikov: let's kill the event

aslushnikov (commented):
Awesome!

aslushnikov:
@JoelEinbinder any updates on this?

JoelEinbinder:
@aslushnikov I removed the event, `browser.close()` returns a promise.

aslushnikov (approved):
Please, rebaseline so that there are no conflicts

JoelEinbinder:
Rebased

PR #708, opened 2017-09-07
block-images: fix bbc.com link

ebidel:
https://bbc.com doesn't appear to work.

### DIFF
--- a/examples/block-images.js
+++ b/examples/block-images.js
@@ -29,7 +29,7 @@ page.on('request', request => {
   else
     request.continue();
 });
-await page.goto('http://www.bbc.com/');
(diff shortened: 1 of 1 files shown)


JoelEinbinder:
Is this a bug? It works for me in chrome, although it forwards me to the http site.

ebidel:
It's dreadfully slow for me. Caused timeout issues running the sample a couple of times last night. 

What about https://news.google.com instead?

JoelEinbinder:
Sure

ebidel:
PTAL

JoelEinbinder (approved):

PR #721, opened 2017-09-08
Docker example

ebidel:


docs/troubleshooting.md
```diff
@@ -91,3 +91,47 @@ xorg-x11-fonts-misc
...
+
+# Uncomment to skip the chromium download when installing puppeteer. If you do,
+# you'll need to launch puppeteer with:
+#     browser.launch({executablePath: 'google-chrome-unstable'})
+# ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true
+
+# Install puppeteer so it can be required by user code that gets ran in server.js.
+RUN yarn add puppeteer
+
+CMD ["google-chrome-unstable", "--no-sandbox"]
```
aslushnikov, line 121: is this necessary?

docs/troubleshooting.md
```diff
@@ -91,3 +91,47 @@ xorg-x11-fonts-misc
...
+```
+
+Build the container:
+
+```bash
+docker build -t puppeteer-chrome-linux .
+```
+
+Run the container by passing `node -e "<yourscript.js content as a string>` as the command:
```
aslushnikov, line 130: You can also mount the current folder inside the container and run node there.
[Example](https://github.com/aslushnikov/latex-online/blob/master/util/startDevServer.sh#L15)

aslushnikov (commented):
nice! WDYT about publishing this on dockerhub? This would allow others to extend easily

docs/troubleshooting.md
ebidel: Yea, I _think_ you have to have a default CMD or ENTRYPOINT.

docs/troubleshooting.md
ebidel: I thought about that. The problem is if you run this on App Engine (possibly others), there's no way to control the run command. It runs `yarn start` for you. That's why I made something that also works well with a server endpoint: https://github.com/ebidel/try-puppeteer#running-the-container

aslushnikov (approved):

PR #729, opened 2017-09-09
Implement page.authenticate method

### LINKED ISSUE #426: Implement HTTP Authentication API
HTTP Authentication could look like this (similar to [Nightmare.js](https://github.com/segmentio/nightmare#authenticationuser-password)):
```js
await page.authenticate(user, pass);
```

aslushnikov:
This patch implements `page.authenticate` which should cover all
cases of HTTP authentication.

Implementation notes:
1. Since HTTP authentication is based on request interception internally, the implementation
of NetworkManager splits the "requestInterceptionEnabled" flag into two:
    - `userRequestInterceptionEnabled` - `true` if the user has enabled request interception
    - `protocolRequestInterceptionEnabled` - `true` if we've enabled request interception in protocol
2. In case of invalid credentials, in order to escape infinite authentication loop we maintain the `attemptedAuthentications` set that stores all the interception ids we've tried to authenticate.

Fixes #426.

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -32,7 +32,12 @@ class NetworkManager extends EventEmitter {
     /** @type {!Object<string, string>} */
     this._extraHTTPHeaders = {};
 
-    this._requestInterceptionEnabled = false;
+    /** @type {?{username: string, password: string}} */
+    this._credentials = null;
+    /** @type {!Set<string>} */
+    this._attemptedAuthentications = new Set();
+    this._userRequestInterceptionEnabled = false;
+    this._protocolRequestInterceptionEnabled = false;
     /** @type {!Multimap<string, string>} */
     this._requestHashToRequestIds = new Multimap();
     /** @type {!Multimap<string, !Object>} */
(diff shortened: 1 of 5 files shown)


robert-shade:
This works great - even worked for NTLM

lib/NetworkManager.js
```diff
@@ -118,17 +156,19 @@ class NetworkManager extends EventEmitter {
    * @param {!Object} requestPayload
    */
   _handleRequestStart(requestId, interceptionId, url, resourceType, requestPayload) {
-    const request = new Request(this._client, requestId, interceptionId, url, resourceType, requestPayload);
+    const request = new Request(this._client, requestId, interceptionId, this._userRequestInterceptionEnabled, url, resourceType, requestPayload);
     this._requestIdToRequest.set(requestId, request);
     this._interceptionIdToRequest.set(interceptionId, request);
+    if (!this._userRequestInterceptionEnabled && this._protocolRequestInterceptionEnabled)
```
dgozman: Let's move this to _onRequestIntercepted.

dgozman (approved):

lib/NetworkManager.js

Androbin:
@aslushnikov Proxy authentication -> Request interception -> Cache disabled
Is there a way to use an authenticated proxy with an enabled cache?

PR #741, opened 2017-09-11
Carefully manage unhandled rejections for navigation

### LINKED ISSUE #738: Cant catch errors with .goto
### Steps to reproduce

* Puppeteer version: 0.10.2
* Platform / OS version: mac 10.12.6

```

puppeteer.launch({ headless: false }).then(async function(browser) {

	const newPage = await browser.newPage();

	// newPage.goto('http://google.com', { timeout: 1 })
	// 	.catch(function(err) {});

	// newPage.goto('http://google.com', { timeout: 1 })
	// 	.then(function() {}, function(err) {});

	try {
		await newPage.goto('http://google.com', { timeout: 1 });
	} catch (e) {}
});
```

To be able to silence the error.

**What happens instead?**
The error isn't caught.
```
(node:33288) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Navigation Timeout Exceeded: 1ms exceeded
(node:33288) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
(node:33288) PromiseRejectionHandledWarning: Promise rejection was handled asynchronously (rejection id: 1)
```

Listening to `unhandledRejection` on the process returns this:
```
Promise {
  <rejected> Error: Navigation Timeout Exceeded: 1ms exceeded
    at NavigatorWatcher.waitForNavigation (./node_modules/puppeteer/lib/NavigatorWatcher.js:74:13)
    at <anonymous> }
```

aslushnikov:
Currently, navigation watcher throws exception if timeout
is exceeded.

Due to the way it is used in `page.navigate`, the promise
get's rejected before it is awaited, which is considered to
be "unhandled promise rejection".

Fixes #738

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -382,8 +382,10 @@ class Page extends EventEmitter {
       watcher.cancel();
       throw e;
     }
-    await result;
+    const error = await result;
     helper.removeEventListeners([listener]);
+    if (error)
+      throw error;
     if (this._frameManager.isMainFrameLoadingFailed())
       throw new Error('Failed to navigate: ' + url);
     return responses.get(this.mainFrame().url()) || null;
@@ -410,9 +412,10 @@ class Page extends EventEmitter {
 
     const responses = new Map();
     const listener = helper.addEventListener(this._networkManager, NetworkManager.Events.Response, response => responses.set(response.url, response));
-    await watcher.waitForNavigation();
+    const error = await watcher.waitForNavigation();
     helper.removeEventListeners([listener]);
-
+    if (error)
+      throw error;
     return responses.get(this.mainFrame().url()) || null;
   }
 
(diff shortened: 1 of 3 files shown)


lib/Page.js
```diff
@@ -375,8 +375,10 @@ class Page extends EventEmitter {
       watcher.cancel();
       throw e;
     }
-    await result;
+    const error = await result;
```
JoelEinbinder: Just for convention can result be something like navigationResultPromise? It is weird to say error = result.

JoelEinbinder (approved):

lib/Page.js

PR #742, opened 2017-09-11
[api] Do not leave dangling promises when sending messages over protocol

### LINKED ISSUE #627: Protocol error Invalid InterceptionId on cancelled requests
When I tried to implement the "naive" image blocking like the one shown in the [API reference page](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagesetrequestinterceptionenabledvalue) by enabling request interception, at one particular webpage I always encounter this error:

`
(node:73712) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Protocol error (Network.continueInterceptedRequest): Invalid InterceptionId. undefined
`

After some digging I figured out that the error always happen when the Puppeteer intercepting request to one particular URL. And when I open that webpage on non-headless Chrome to debug the all requests being made, it turns out the request which cause the error has "canceled" status.

screen shot 2017-08-31 at 7

What I suspect is when I called `interceptedRequest.continue()`, the underlying Chromium process already "lost" the interception, which cause it to return [this error](https://chromium.googlesource.com/chromium/src.git/+/refs/heads/lkcr/content/browser/devtools/devtools_url_request_interceptor.cc#121).

I tried to ignore that URL and don't call `abort` or `continue`, but it caused the whole page to "hang".

aslushnikov:
It's very bad to have 'unhandled promise rejection' that can't be
handled in user code. These errors will exit node process in a near
future.

This patch avoids 'unhandled promise rejection' while sending protocol
messages.

This patch:
- introduces `puppeteer:error` debug scope and starts using it for all
  swallowed errors.
- makes sure that every `client.send` method is either awaited or its
  errors are handled.
- starts return promises from Request.continue() and Request.abort().
- starts swallow errors from Request.contine() and Request.abort().

The last is the most important part of the patch. Since
`Request.continue()` might try to continue canceled request, we should
disregard the error.

Fixes #627.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1314,6 +1314,7 @@ If request fails at some point, then instead of 'requestfinished' event (and pos
 If request gets a 'redirect' response, the request is successfully finished with the 'requestfinished' event, and a new request is  issued to a redirected url.
 
 #### request.abort()
+- returns: <[Promise]>
 
 Aborts request. To use this, request interception should be enabled with `page.setRequestInterceptionEnabled`.
 Exception is immediately thrown if the request interception is not enabled.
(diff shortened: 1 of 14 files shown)


JoelEinbinder (approved):
We have some similar code related to input that I'll fix.

PR #751, opened 2017-09-11
Fixes #750 - Throw error with stack trace when calling launch with an…

### LINKED ISSUE #750: Throw error with stack trace when launching with an invalid executablePath option
When `puppeteer` is used as a dependency or sub-dependency of another project that exposes the `puppeteer` API, an invalid `executablePath` option should be reported appropriately. Currently, if an invalid `executablePath` is provided to `puppeteer.launch(...)`, it is unclear where the actual error is coming from. It could be from any module. This error should be propagated with a stack trace directly from `puppeteer`. I have a fix for this and I will put up a PR momentarily.

### Steps to reproduce

* Puppeteer version: 0.11.0-alpha
* Platform / OS version: macOS 10.12.6

1. `await puppeteer.launch({ executablePath: 'random-path' })`

An error containing a stack trace clearly describing the issue e.g.

```
Error: Error launching Chromium process with executable path "random-path"
    at fs.stat (/puppeteer/lib/Launcher.js:147:23)
    at FSReqWrap.oncomplete (fs.js:152:21)
```

**What happens instead?**

```
events.js:182
      throw er; // Unhandled 'error' event
      ^

Error: spawn random-path ENOENT
    at exports._errnoException (util.js:1016:11)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:189:19)
    at onErrorNT (internal/child_process.js:366:16)
    at _combinedTickCallback (internal/process/next_tick.js:102:11)
    at process._tickCallback (internal/process/next_tick.js:161:9)
```

austinkelleher:
… invalid executablePath option.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -77,10 +77,26 @@ class Launcher {
       console.assert(revisionInfo.downloaded, `Chromium revision is not downloaded. Run "npm install"`);
       chromeExecutable = revisionInfo.executablePath;
     }
+
     if (Array.isArray(options.args))
       chromeArguments.push(...options.args);
 
     const chromeProcess = childProcess.spawn(chromeExecutable, chromeArguments, {detached: true});
+
+    const chromeProcessErrorListeners = [];
+    let onChromeProcessError;
+    let chromeProcessStartFail = false;
+
+    const chromeProcessErrorPromise = new Promise((resolve, reject) => {
+      onChromeProcessError = err => {
+        const error = new Error(`Error launching Chromium process with executable path "${chromeExecutable}"`);
+        chromeProcessStartFail = true;
+        reject(error);
+      };
+
+      chromeProcessErrorListeners.push(helper.addEventListener(chromeProcess, 'error', onChromeProcessError));
+    });
+
     if (options.dumpio) {
       chromeProcess.stdout.pipe(process.stdout);
       chromeProcess.stderr.pipe(process.stderr);
@@ -101,10 +117,15 @@ class Launcher {
 
     try {
       const connectionDelay = options.slowMo || 0;
-      const browserWSEndpoint = await waitForWSEndpoint(chromeProcess, options.timeout || 30 * 1000);
+      const browserWSEndpointPromise = waitForWSEndpoint(chromeProcess, options.timeout || 30 * 1000);
+
+      const browserWSEndpoint = await Promise.race([chromeProcessErrorPromise, browserWSEndpointPromise]);
+      helper.removeEventListeners(chromeProcessErrorListeners);
+
       const connection = await Connection.create(browserWSEndpoint, connectionDelay);
       return new Browser(connection, !!options.ignoreHTTPSErrors, killChrome);
     } catch (e) {
+      helper.removeEventListeners(chromeProcessErrorListeners);
       killChrome();
       throw e;
     }
@@ -114,7 +135,9 @@ class Launcher {
      */
     function killChrome() {
       helper.removeEventListeners(listeners);
-      if (temporaryUserDataDir) {
+      if (chromeProcessStartFail) {
+        return Promise.resolve();
+      } else if (temporaryUserDataDir) {
         // Force kill chrome.
         if (process.platform === 'win32')
           childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T /F`);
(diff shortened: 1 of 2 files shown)


austinkelleher:
I signed it!

lib/Launcher.js
```diff
@@ -173,4 +176,14 @@ function waitForWSEndpoint(chromeProcess, timeout) {
   });
 }
 
+function validateExecutablePath(chromeExecutable) {
```
aslushnikov: let's add jsdoc:

```js
/**
 * @param {string} chromeExecutable
 * @return {!Promise}
 */
```

test/test.js
```diff
@@ -106,6 +106,12 @@ describe('Browser', function() {
     await neverResolves;
     expect(error.message).toContain('Protocol error');
   }));
+  it('should reject if executable path is invalid', SX(async function() {
+    let waitError = null;
+    const launchOptions = Object.assign({ defaultBrowserOptions }, { executablePath: 'random-invalid-path' });
```
aslushnikov: you want to extend default options with executable path:
```js
const launchOptions = Object.assign({ executablePath: 'random-invalid-path' }, defaultBrowserOptions);
```

aslushnikov (commented):
Good CL, thank you!

test/test.js
austinkelleher: I made a small change, but I don't think we should have `{ executablePath: ... }` as the first argument to `Object.assign(...)` because if `process.env.CHROME` is supplied, then the `executablePath` supplied in `Object.assign(...)` will be overridden causing this test to fail.

```js
const options = Object.assign({}, defaultBrowserOptions, {executablePath: 'random-invalid-path'});
```

https://github.com/GoogleChrome/puppeteer/pull/751/files#diff-c1129c8b045390789fa8ff62f2c6b4a9R111

austinkelleher:
@aslushnikov Thanks for the feedback. I made the changes and squashed.

JoelEinbinder:
The `executablePath` actually takes a command right now. So you can set `executablePath` to `"chrome-unstable"` and it will fail this check but run correctly. Instead I think we should check for an error on `childProcess.spawn`.

austinkelleher:
@JoelEinbinder You're right. I will make the changes soon. Thanks.

austinkelleher:
Hey, so I was looking at attaching an `error` event listener to the spawned process (here: https://github.com/GoogleChrome/puppeteer/blob/master/lib/Launcher.js#L81), but I can't really think of a good way to reject with an error back to the `puppeteer.launch(...)` caller without introducing a timeout. Something like this (untested):

```js
static async launch(options) {
  ...
  const chromeProcess = childProcess.spawn(chromeExecutable, chromeArguments, {detached: true});

  await new Promise((resolve, reject) => {
    let timeout = setTimeout(resolve, 1000);
    chromeProcess.on('error', () => {
      clearTimeout(timeout);
      reject(new Error(`Error launching Chromium process with executable path "${chromeExecutable}"`));
    });
  });
  ...
}
```

I don't really like the idea of introducing a timer, but I'm not sure what the best way of getting around it is.

We could simply just `console.error` inside of the ` chromeProcess.on('error')` handler, but I'm not a huge fan of that either.

Thoughts @aslushnikov @JoelEinbinder?

JoelEinbinder:
You can race `chromeProcess.on('error')` against `waitForWSEndpoint`. If the process sends us a valid websocket endpoint, I think we can safely assume Chrome launched successfully.

austinkelleher:
Thanks @JoelEinbinder. Sounds good. I will have the fix up soon.

aslushnikov:
@austinkelleher any updates on this?

austinkelleher:
Hey, @aslushnikov. I apologize for the delay on PR feedback. Got caught up in some other things. Should be able to get to this later today.

aslushnikov:
@austinkelleher thanks and no rush, I was mostly curious if you still have interest in this PR.

austinkelleher:
Whoops I need to make a few more changes. Hold on.

austinkelleher:
@aslushnikov @JoelEinbinder Thanks again for the feedback. I've addressed them all.

austinkelleher:
@aslushnikov @JoelEinbinder Can you please review my changes? Thanks!

aslushnikov:
@austinkelleher I apologize for the long reply, I'm currently traveling.

Codewise, I think this solution is a little too involved. It should be enough to catch the 'error' event on the chrome process - check out the  https://github.com/GoogleChrome/puppeteer/pull/863.

aslushnikov:
Closing in favor of #863

PR #752, opened 2017-09-11
Being able to define custom chromium download URL

tomasAlabes:
Hi, first thanks for the awesome work with puppeteer!
I need to be able to download chromium from a private repository (like Artifactory) mirroring the chromium repo `https://storage.googleapis.com/chromium-browser-snapshots`.
I added a env variable called `PUPPETEER_DOWNLOAD_CHROMIUM_URL` to specify it. I couldn't find any tests related to downloading chromium or process env variables.
Let me know what you think.
Thanks

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -150,6 +150,7 @@ Puppeteer looks for certain environment variables to aid its operations:
 
 - `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
 - `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step.
+- `PUPPETEER_DOWNLOAD_CHROMIUM_URL` - download bundled Chromium from specified URL.
 
 ### class: Puppeteer
 
--- a/utils/ChromiumDownloader.js
+++ b/utils/ChromiumDownloader.js
@@ -27,12 +27,17 @@ const getProxyForUrl = require('proxy-from-env').getProxyForUrl;
 
 const DOWNLOADS_FOLDER = path.join(__dirname, '..', '.local-chromium');
 
+const downloadURLPrefix = process.env.PUPPETEER_DOWNLOAD_CHROMIUM_URL || 'https://storage.googleapis.com/chromium-browser-snapshots';
 const downloadURLs = {
-  linux: 'https://storage.googleapis.com/chromium-browser-snapshots/Linux_x64/%d/chrome-linux.zip',
-  mac: 'https://storage.googleapis.com/chromium-browser-snapshots/Mac/%d/chrome-mac.zip',
-  win32: 'https://storage.googleapis.com/chromium-browser-snapshots/Win/%d/chrome-win32.zip',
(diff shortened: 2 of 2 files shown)


tomasAlabes:
Signed tha CLA.

JoelEinbinder:
Would `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` and then manually downloading Chromium/Chrome work for you? I'm not familiar Artifactory so I am not sure what advantage this gives.

tomasAlabes:
Yes, I forgot to say that, that's another approach without this. I could download it manually and set that variable to skip it later. This is just to give the option directly in puppeteer.
Artifactory functions as a private mirror for `storage.googleapis.com`, so we don't have to go outside our vpn to get it and don't relay on a public repo to be up.
It's ok if you want to go for the skip option instead of adding this one, I thought it was nice to have it built-in. ✌️

tomasAlabes:
FYI, the code I ended up writing for this is:

```
chromium_download_dir="/tmp/"
chromium_install_dir="/tmp/chrome-linux/"
CUSTOM_URL=http://my.url.com

if [ ! -d "${chromium_install_dir}" ]; then
  echo 'chromium not downloaded, downloading...'
  CHROMIUM_REVISION=`curl -s 'https://raw.githubusercontent.com/GoogleChrome/puppeteer/master/package.json' | grep -oP '(?<="chromium_revision": ")[^"]*'`
  wget -q $CUSTOM_URL/Linux_x64/$CHROMIUM_REVISION/chrome-linux.zip
  echo 'Chrome Downloaded.'
  unzip -q chrome-linux.zip -d $chromium_download_dir
  echo 'Chrome unzipped.'
else
    echo "chromium already downloaded in ${chromium_install_dir}, skipping download"
fi

export CHROMIUM_BIN=/tmp/chrome-linux/chrome
export PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
```
If it helps somebody, or as a way to see how it could save some people's time. :)

aslushnikov:
Hey @tomasAlabes, thanks for the PR!

The `PUPPETEER_DOWNLOAD_CHROMIUM_URL` might not be flexible enough as it only overrides URL prefix. What if I want to download Chromium from a completely different URL?

So I think the script you [came up with](https://github.com/GoogleChrome/puppeteer/pull/752#issuecomment-329326269) is good. The only thing to note is that you parse out the Chromium revision from puppeteer's master:
```bash
https://raw.githubusercontent.com/GoogleChrome/puppeteer/master/package.json
```
This constantly changes and theoretically might not be compatible with the released version of Puppeteer.

To be on the safe side, you can target exact version if you use release number instead of `master` (notice `v0.10.2` in the url):
```bash
https://raw.githubusercontent.com/GoogleChrome/puppeteer/v0.10.2/package.json
```

tomasAlabes:
@aslushnikov thanks for your reply. Yeah, I set it as a prefix thinking on mirroring the google storage, but if for example you want only 1 specific linux chromium release always, then the env variable might not suffice.
Thank you for your input on the version, I didn't think of puppeteer and chrome being unsynced.
🍻

PR #771, opened 2017-09-13
add Puppeteer.executablePath() method

### LINKED ISSUE #745: Feature request: Run browser process as different user
I'd like to be able to run the browser as a different user. That way I can apply a set of outbound firewall rules to it to prevent it from accessing anything in the internal network while still allowing the node process to do that.
The use-case for this lies in `rendertron`-like services that expose some of the browser functionality to the outside world. It could be a security threat if those were able to access internal IPs.

aslushnikov:
This patch adds Puppeteer.executablePath() method to query the path
of bundled chromium.

Fixes #745

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -174,6 +174,9 @@ puppeteer.launch().then(async browser => {
 
 This methods attaches Puppeteer to an existing Chromium instance.
 
+#### puppeteer.executablePath()
+- returns: <[string]> A path where Puppeteer expects to find bundled chromium. Chromium might not exist there if the download was skipped with `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`.
+
 #### puppeteer.launch([options])
 - `options` <[Object]>  Set of configurable options to set on the browser. Can have the following fields:
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -174,6 +174,9 @@ puppeteer.launch().then(async browser => {
 
 This methods attaches Puppeteer to an existing Chromium instance.
 
+#### puppeteer.executablePath()
+- returns: <[string]> A path where Puppeteer expects to find bundled chromium. Chromium might not exist there if the download was skipped with `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`.
```
ebidel: link to the env vars section

ebidel: s/chromium/Chromium

docs/api.md
```diff
@@ -174,6 +174,9 @@ puppeteer.launch().then(async browser => {
 
 This methods attaches Puppeteer to an existing Chromium instance.
 
+#### puppeteer.executablePath()
```
vsemozhetbyt, line 178: This header is missing in the TOC.

JoelEinbinder (approved):
Looks good as long as the other comments are addressed.

Janpot:
I added an alternative solution for https://github.com/GoogleChrome/puppeteer/issues/745
I think it could also work by just providing the executablePath in the environment when spawning the child process. That solution wouldn't require public API change.

docs/api.md
aslushnikov: Done.

docs/api.md

aslushnikov:
@Janpot the `puppeteer.executablePath` is good by itself: i've just seen someone reaching out to ChromiumDownloader to pull the `executablePath`: https://github.com/karma-runner/karma-chrome-launcher#headless-chromium-with-puppeteer

Issue #760, opened 2017-09-12
Translate API support in near future?


ganapativs:
In need of website translate support. Will there be support anytime soon?

ebidel:
Can you elaborate on the request? Chrome's translation service isn't a public API and would need to be supported by the DevTools protocol for it to be exposed in Puppeteer.

There are good 3rd party libs/services like https://cloud.google.com/translate/docs/ that do translation. 

@aslushnikov @JoelEinbinder, on a similar note, [chrome-launcher turns off the service](https://github.com/GoogleChrome/lighthouse/blob/43baee69702d3c85a53e96ab52b0f73d3548af95/chrome-launcher/flags.ts#L10) when we launch Chrome. Should we do that too?

ganapativs:
That makes sense, we wanted a way to convert other language sites into English before taking screenshot. Think we have to use translate API from google cloud. Thanks mate 👍

PR #774, opened 2017-09-13
Launcher: disable extensions and google translate by default

aslushnikov:
This patch teaches `puppeteer.launch` to run chromium with disabled
chrome extensions and google translate services.

References #760


ebidel (approved):

aslushnikov:
@ebidel it turned out I don't really know why we want this in. Why would we want this?

ebidel:
Keep the launched chrome slimmer? Not sure it adds a ton of value for us. Turning these off makes more sense for Lighthouse where we're reusing an existing install of chrome.

aslushnikov:
I slept with this and now I like the patch since it aligns headful/headless with each other.
@JoelEinbinder any concerns?

JoelEinbinder (approved):

cretz:
@aslushnikov , @JoelEinbinder, et al - Sadly creating and merging this has disabled everyone's ability to automate with extensions in non-headless mode. Is there any way to work around this, or are these "default" args more like "permanent unchangeable" args?

aslushnikov:
@cretz @JoelEinbinder is working on a patch to override default arguments. Filed #872 to track progress.

PR #775, opened 2017-09-13
page.waitFor should send ...args to page.waitForFunction

### LINKED ISSUE #770: No argument is passed to waitForFunction/waitFor
### Steps to reproduce

* Puppeteer version: ^0.10.2
* Platform / OS version: Ubuntu 16.04.2 LTS

## Problem 1: arguments passed to waitFor is undefined:
1. Pass any argument to waitFor or waitForFunction.
2. Get/Call that argument inside the function.
3. It's `undefined`.

## Problem 2: `arguments` property inside page.evaluate is undefined.
1. Call `arguments` inside any page.evaluate function.
2. It's undefined and results in TypeError.

```js
(async() => {

  await page.waitForFunction((args) => {
     
    // some tests to show what is happeneing.
    console.log([
      args, // should show args
      arguments, // should not have 0 elements
      Array.isArray(args), // should return True
      Array.isArray([1, 2, 3, 4]) // should return True
    ]);

    return Array.isArray(args);
  }, [1, 2, 3, 4]); // no matter what you pass, it's still undefined;

  //browser.close();
})();
```

Meanwhile, the evaluate method shows `arguments` as undefined while the args provided is working properly.
```js
await page.evaluate((args) => {
    // shows `ReferenceError: arguments is not defined`
    console.log(arguments)

    // runs properly
    console.log([
      args,
      Array.isArray(args),
      Array.isArray([1, 2, 3, 4])
    ]);

    return Array.isArray(args);
  }, [1, 2, 3, 4]);
```

It should behave like `page evaluate` and the arguments should be passed properly.

**What happens instead?**
Arguments are not passed.
Waits are timed out.

Screenshot of console with the code ran thru.
image

JoelEinbinder:
This lets the user pass `...args` into `page.waitFor`. It also clarifies that the docs that `options` is not optional if `...args` are specified.

Fixes #770

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -74,8 +74,8 @@
     + [page.type(text, options)](#pagetypetext-options)
     + [page.url()](#pageurl)
     + [page.viewport()](#pageviewport)
-    + [page.waitFor(selectorOrFunctionOrTimeout[, options])](#pagewaitforselectororfunctionortimeout-options)
-    + [page.waitForFunction(pageFunction[, options, ...args])](#pagewaitforfunctionpagefunction-options-args)
+    + [page.waitFor(selectorOrFunctionOrTimeout[, options[, ...args]])](#pagewaitforselectororfunctionortimeout-options-args)
+    + [page.waitForFunction(pageFunction[, options[, ...args]])](#pagewaitforfunctionpagefunction-options-args)
     + [page.waitForNavigation(options)](#pagewaitfornavigationoptions)
     + [page.waitForSelector(selector[, options])](#pagewaitforselectorselector-options)
   * [class: Keyboard](#class-keyboard)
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -851,9 +852,9 @@ This method behaves differently with respect to the type of the first parameter:
 - if `selectorOrFunctionOrTimeout` is a `number`, than the first argument is treated as a timeout in milliseconds and the method returns a promise which resolves after the timeout
 - otherwise, an exception is thrown
 
-Shortcut for [page.mainFrame().waitFor(selectorOrFunctionOrTimeout[, options])](#framewaitforselectororfunctionortimeout-options).
+Shortcut for [page.mainFrame().waitFor(selectorOrFunctionOrTimeout[, options[, ...args]])](#framewaitforselectororfunctionortimeout-options).
```
vsemozhetbyt: `#framewaitforselectororfunctionortimeout-options` ->
`#framewaitforselectororfunctionortimeout-options-args`

test/test.js
```diff
@@ -524,6 +524,11 @@ describe('Page', function() {
       await page.waitFor({foo: 'bar'}).catch(e => error = e);
       expect(error.message).toContain('Unsupported target type');
     }));
+    it('should wait for predicate with arguments', SX(async function() {
+      const watchdog = page.waitFor((arg1, arg2) => window[arg1] < arg2, {}, 'innerWidth', 100);
+      page.setViewport({width: 10, height: 10});
```
aslushnikov: `page.setViewport` is doing *so much*, let's do a oneliner instead:
```js
it('should wait for predicate with arguments', SX(async function() {
  await page.waitFor((arg1, arg2) => arg1 !== arg2, {}, 1, 2);
}));
```

aslushnikov (approved):

docs/api.md
JoelEinbinder: Nice catch!

PR #779, opened 2017-09-14
Added select method to Page Class

alixaxel:
This PR adds the `Page.select` method to interact with `select` elements on the page.

The accompanying tests and docs demonstrate the basic usage.


docs/api.md
```diff
@@ -720,6 +721,18 @@ Shortcut for [`keyboard.down`](#keyboarddownkey-options) and [`keyboard.up`](#ke
     - `omitBackground` <[boolean]> Hides default white background and allows capturing screenshots with transparency. Defaults to `false`.
 - returns: <[Promise]<[Buffer]>> Promise which resolves to buffer with captured screenshot
 
+#### page.select(selector, value)
+- `selector` <[string]> A [selector] to query page for
+- `value` <[string|string[]]> A string, or array strings in case of `multiple`, of options to select. If the provided value cannot be matched to the `option` value, it tries to match against the `option` label instead.
```
vsemozhetbyt: It seems this should be `<[string]>|<[Array]<[string]>>` according to `args` format [here](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions).

alixaxel: @vsemozhetbyt I wasn't sure and the doc linter didn't complain so I thought was okay. I'll push a fix, thanks.

aslushnikov:
Hey @alixaxel, thank you for the thorough PR.

It looks like the API is similar to nightmare's [`page.select`](https://github.com/segmentio/nightmare#selectselector-option), but it's hard to tell since the doc is not elaborate enough. Is this the case?

lib/Page.js
```diff
@@ -694,6 +694,43 @@ class Page extends EventEmitter {
...
+   * @param {string} selector
+   * @param {string|string[]} value
+   */
+  async select(selector, value) {
+    const handle = await this.$(selector);
+    console.assert(handle, 'No node found for selector: ' + selector);
+
+    await this.evaluate((element, value) => {
+      if (element.nodeName.toLowerCase() !== 'select')
+        return false;
```
aslushnikov: On line 702 you throw if there's no element matching selector.
In order to be consistent, you should also throw if there's an element that is not a `<select>`.

lib/Page.js
```diff
@@ -694,6 +694,43 @@ class Page extends EventEmitter {
...
+
+      if (Array.isArray(value) !== true)
+        value = [value];
+
+      if (element.multiple !== true) {
+        element.value = value[0];
+
+        if (element.value !== value[0])
+          [].some.call(element.options, option => option.selected = value[0] === option.label);
```
aslushnikov: let's convert options to array once to avoid the `Array.prototype.call` workaround:

```js
const options = Array.from(element.options);
// ...
```

alixaxel:
Hi @aslushnikov! Nightmare's [`page.select`](https://github.com/segmentio/nightmare/blob/master/lib/actions.js#L355) only covers the most trivial selection scenario. My implementation is actually highly based on GhostJS's [`fill`](https://github.com/KevinGrandon/ghostjs/blob/aec36d9b69119069053ce8f812b306673420f292/ghostjs-core/src/element.js#L102-L123) method, with a few changes, namely:

- more reliable selector fallback for elements with `multiple` property
- checking for standard `label` attribute of `<option>` (according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/option))

lib/Page.js
```diff
@@ -694,6 +694,43 @@ class Page extends EventEmitter {
...
+  async select(selector, value) {
+    const handle = await this.$(selector);
+    console.assert(handle, 'No node found for selector: ' + selector);
+
+    await this.evaluate((element, value) => {
+      if (element.nodeName.toLowerCase() !== 'select')
+        return false;
+
+      if (Array.isArray(value) !== true)
+        value = [value];
```
aslushnikov: nit: the `value` is always an array; let's call it `values` to make it clear

test/test.js
```diff
@@ -1999,6 +2004,50 @@ describe('Page', function() {
...
+      expect(await page.evaluate(() => result)).toBe('blue');
+    }));
+
+    it('should select single option by label', SX(async function() {
+      await page.goto(PREFIX + '/input/select.html');
+      await page.select('select', 'Brown');
+      expect(await page.evaluate(() => result)).toBe('brown');
+    }));
+
+    it('should select multiple options', SX(async function() {
```
aslushnikov: how is this selecting multiple options?

docs/api.md
```diff
@@ -720,6 +721,18 @@ Shortcut for [`keyboard.down`](#keyboarddownkey-options) and [`keyboard.up`](#ke
     - `omitBackground` <[boolean]> Hides default white background and allows capturing screenshots with transparency. Defaults to `false`.
 - returns: <[Promise]<[Buffer]>> Promise which resolves to buffer with captured screenshot
 
+#### page.select(selector, value)
+- `selector` <[string]> A [selector] to query page for
+- `value` <[string]|[Array]<[string]>> A string, or array strings in case of `multiple`, of options to select. If the provided value cannot be matched to the `option` value, it tries to match against the `option` label instead.
```
aslushnikov: we usually do varargs to support both array and a single value. Check out [`ElementHandle.uploadFile`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#elementhandleuploadfilefilepaths)

aslushnikov: Regarding the matching against value / label, is this an innovation or it's already implemented in some other automation library?

test/test.js
```diff
@@ -1999,6 +2004,50 @@ describe('Page', function() {
     }));
   });
 
+  describe('Page.select', function() {
+    it('should select single option', SX(async function() {
+      await page.goto(PREFIX + '/input/select.html');
+      await page.select('select', '');
```
aslushnikov: what does empty string mean as an argument?

lib/Page.js
alixaxel: This was how I first wrote, but since it can also be provided to the method as a string, I thought `value` was more appropriate in either case - but I'll change it. :+1:

lib/Page.js
```diff
@@ -694,6 +694,40 @@ class Page extends EventEmitter {
...
+
+        if (element.value !== value[0])
+          [].some.call(element.options, option => option.selected = value[0] === option.label);
+      } else {
+        [].forEach.call(element.options, option => {
+          option.selected = value.includes(option.value) || value.includes(option.label);
+        });
+      }
+
+      ['change', 'input'].forEach(trigger => {
```
aslushnikov: nit: let's explicitly emit two events
```js
element.dispatchEvent(new Event('change'));
element.dispatchEvent(new Event('input'));
```

test/test.js
alixaxel: Bad wording, will clarify.

test/test.js
alixaxel: A non-existent option, could be anything (that's not on the options). Will clarify.

docs/api.md
alixaxel: Regarding the matching against value / label, first time I saw it was on CasperJS.
GhostJS also borrows the idea.

lib/Page.js
```diff
@@ -694,6 +694,41 @@ class Page extends EventEmitter {
...
+   * @param {string} selector
+   * @param {!Array<string>} values
+   */
+  async select(selector, ...values) {
+    const handle = await this.$(selector);
+    console.assert(handle, 'No node found for selector: ' + selector);
+
+    await this.evaluate((element, values) => {
+      if (element.nodeName.toLowerCase() !== 'select')
+        throw new Error('Element is not a <select> element.');
```
aslushnikov: in this case, the element handle will never be disposed (since the `handle.dispose` won't be reached).

If you use the `this.$eval`, this should be handled just fine. (and if it doesn't, then we should fix `page.$eval`)

lib/Page.js
```diff
@@ -694,6 +694,41 @@ class Page extends EventEmitter {
...
+
+      const options = Array.from(element.options);
+
+      if (Array.isArray(values) !== true)
```
aslushnikov: this `if` could be removed - `values` are always an array now since you switched to varargs

lib/Page.js
```diff
@@ -694,6 +694,41 @@ class Page extends EventEmitter {
...
+        values = [values];
+
+      if (element.multiple !== true) {
```
aslushnikov: nit: could you please switch this branch with the following one. I mean:

```js
if (element.multiple) {
  // ...
} else {
}
```
it's easier to read assertions rather then negations.

alixaxel:
Hey @aslushnikov, I pushed the fixes you requested - should I make any other changes?

lib/Page.js
```diff
@@ -694,6 +694,41 @@ class Page extends EventEmitter {
...
+        element.value = values[0];
+
+        if (element.value !== values[0])
+          options.some(option => option.selected = values[0] === option.label);
+      } else {
+        options.forEach(option => {
```
aslushnikov: nit: let's use for-of
```js
for (let option of options)
  option.selected = values.includes(option.value) || values.includes(option.label);
```

test/assets/input/select-multiple.html
```diff
@@ -0,0 +1,36 @@
...
+      <option value="violet">Violet</option>
+      <option value="white">White</option>
+      <option value="yellow">Yellow</option>
+    </select>
+    <script>
+      window.result = null;
+
+      let select = document.querySelector('select');
+
+      select.addEventListener('input', () => {
```
aslushnikov: can we also cover the 'change' event? And since the asset is getting heavier, can we reuse single HTML asset to test both multiple and single selection? The `select.html` can always return an array, and the `multiple` attribute could be set from-inside the test:

```js
await page.goto(PREFIX + '/input/select.html');
await page.$eval('select', select => select.setAttribute('multiple', true));
```

lib/Page.js
```diff
@@ -694,6 +694,41 @@ class Page extends EventEmitter {
...
+
+      const options = Array.from(element.options);
+
+      if (Array.isArray(values) !== true)
+        values = [values];
+
+      if (element.multiple !== true) {
+        element.value = values[0];
+
+        if (element.value !== values[0])
```
aslushnikov: I'm concerned about falling back to the value of `label`. Depending on the combination of `value` and `label` attributes on every option, the matching will happen against either `value`, `label` or node's text content. The outcome is hard for me to predict.

Can we start with matching against values only? When/if there's a strong need to match against labels, I'd prefer it to be stated explicitly with options object.

lib/Page.js
```diff
@@ -694,6 +694,41 @@ class Page extends EventEmitter {
...
+          options.some(option => option.selected = values[0] === option.label);
+      } else {
+        options.forEach(option => {
+          option.selected = values.includes(option.value) || values.includes(option.label);
+        });
+      }
+
+      element.dispatchEvent(new Event('change'));
+      element.dispatchEvent(new Event('input'));
```
aslushnikov: do we need to dispatch 'input'? Nightmare gets away [without it](https://github.com/segmentio/nightmare/blob/07b4ba441c9836ab351d091f44e89af9fdf5eb88/lib/actions.js#L355)

aslushnikov (commented):
@alixaxel thank you for working on this and sorry for the long reply! I'm traveling and will be slow for a while.

lib/Page.js
alixaxel: Good to know, I'll fix that.

lib/Page.js
alixaxel: I think so, even listeners could only be listening to `input`, and I think those are also the events that the browser sends on a typical mouse-selection scenario.

test/assets/input/select-multiple.html
alixaxel: Aren't we mixing fixtures with the tests doing it this way? What guarantees do we have that `select.setAttribute('multiple', true)` will actually yield the expected result?

lib/Page.js
alixaxel: @aslushnikov For me this is a important feature, in fact for my specific use case I would specialize it even more. Often times, the option values have little resemblance to the label(s) you're targeting.

Nightmare.js has a convenient [`action`](https://github.com/segmentio/nightmare#extending-nightmare) method that can be used to implement custom behaviors in a global browser scope without repeating code and, AFAIK, puppeteer only has `exposeFunction` on a page scope.

Your `options` suggestion would still be limiting, what are your thoughts on having a `page.select(selector, callback)` signature, where the `callback` is a resolver function that gets the option `value`, `label`, `textContent`, `index` as arguments and has to return `true` to have that option selected, `false` otherwise?

A few scenarios I'm facing that could benefit from this (maybe applies to others too), e.g.:

```html
<select id="country">
	<option value="250">France</option>
	<option value="276">Deutschland</option>
	<option value="620">Portugal</option>
	<option value="724">España</option>
	<option value="840">United States of America</option>
</select>
```

You could select based on value directly:

```javascript
page.select('#country', '724');
```

Or implement advanced "resolvers", for instance ISO 3166-1 numeric to Alpha 2:

```javascript
page.select('#country', 'DE', (target, value, label, options) => {
  let resolved = null;

  switch (value) {
    case 250:
      resolved = 'FR';
    break;

    case 276:
      resolved = 'DE';
    break;

    case 620:
      resolved = 'PT';
    break;

    case 724:
      resolved = 'ES';
    break;

    case 840:
      resolved = 'US';
    break;
  }

  return target === resolved;
});
```

Or even go fuzzy if needed be:

```javascript
page.select('#country', 'Spain', (target, value, label, options) => {
  /**
   * Get all `options`, and find the one with the lowest
   * string distance from either the `value` or the `label`.
   *
   * In this case, "Spain" would be resolved to the value of "España".
   */
});
```

Reading other issues, this could also be helpful to resolve implementation inconsistencies between in libraries and frameworks (ReactJS Virtual DOM was mentioned for instance).

aslushnikov: > Nightmare.js has a convenient action method that can be used to implement custom behaviors in a global browser scope without repeating code and, AFAIK, puppeteer only has exposeFunction on a page scope.

If I understand Nightmare's action correctly, the analogue in puppeteer would be adding new methods to the page prototype:

```js
Page.prototype.size = function(done) {
  return this.evaluate(() => {
    const w = Math.max(document.documentElement.clientWidth, window.innerWidth || 0)
    const h = Math.max(document.documentElement.clientHeight, window.innerHeight || 0)
    return {
      height: h,
      width: w
    }
  });
})
```

It looks like it has nothing to do with `page.exposeFunction`.

> what are your thoughts on having a page.select(selector, callback) signature, where the callback is a resolver function that gets the option value, label, textContent, index as arguments and has to return true to have that option selected, false otherwise?

I think it's not a simple API.

If you can select by values, then you can implement custom resolver with `page.evaluate`. 
For example, selecting all options that contain letter `'e'` in their label:

```js
const values = await page.evaluate(() => {
  const options = Array.from(document.querySelectorAll('select > option'));
  const optionsWithE = options.filter(option => option.label.includes('e'));
  return optionsWithE.map(option => option.value);
  });
});
await page.select('select', ...values);
```

lib/Page.js
```diff
@@ -694,6 +694,41 @@ class Page extends EventEmitter {
...
   /**
+   * @param {string} selector
+   * @param {!Array<string>} values
+   */
+  async select(selector, ...values) {
+    const handle = await this.$(selector);
+    console.assert(handle, 'No node found for selector: ' + selector);
+
+    await this.evaluate((element, values) => {
+      if (element.nodeName.toLowerCase() !== 'select')
```
Garbee: Why call `toLowerCase` here? Checking against `'SELECT'` is just fine and saves extra work from happening.

lib/Page.js
Garbee: `input` is fired on select changes. So it should be fired to best reflect what browsers do.

alixaxel: I'll keep dispatching `change` and `input` events, since that's what browsers do.

lib/Page.js
alixaxel: @Garbee According to https://johnresig.com/blog/nodename-case-sensitivity/:

> 1. The node names of HTML elements are always uppercase, even if they’re explicitly created using lowercase characters. <html> will result in a  .nodeName === "HTML" (see the HTML 5 draft).
> 2. The node names of XML elements are always in the original case, as specified when they’re created. <data> will result in a .nodeName === "data",  <DATA> will result in a .nodeName === "DATA".

I think it's not a big thing to leave `toLowerCase` just for the sake of reliability.

alixaxel:
@aslushnikov Sorry for the delay in pushing all the requested changes, but I've had a very busy week. :100:

Please let me know if anything else is needed.

lib/Page.js
alixaxel: @aslushnikov Extending the `Page` prototype would actually be super useful to me!

But I'm not sure how to go about accessing the `Page` class, it appears that it's not exported?

docs/api.md
```diff
@@ -720,6 +721,19 @@ Shortcut for [`keyboard.down`](#keyboarddownkey-options) and [`keyboard.up`](#ke
...
+- `selector` <[string]> A [selector] to query page for
+- `...values` <...[string]> Values of options to select. If the `<select>` has the `multiple` attribute, all values are considered, otherwise only the first one is taken into account.
+- returns: <[Promise]>
+
+Triggers a `change` and `input` event once all the provided options have been selected.
+If there's no `<select>` element matching `selector`, the method throws an error.
+
+```js
+page.select('select#colors', 'blue'); // single selection
+page.select('select#colors', ['red', 'green', 'blue']); // multiple selections
```
aslushnikov: no need for the array:

```js
page.select('select#colors', 'red', 'green', 'blue');
```

aslushnikov (approved):
Thank you for the CL - this is great!
There's one typo in docs and we can have this merged.

alixaxel:
@aslushnikov Fixed typo. :)

lib/Page.js
aslushnikov: @alixaxel ah right. You can "hack" it with a manual require:

```js
const Page = require('puppeteer/lib/Page');
```

This is not safe of course, but this magic require should not change much over the time.

PR #780, opened 2017-09-14
Added info on how to find the browserWSEndpoint

JakeGinnivan:
It took me a while to figure this out, thought it may help others.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -228,7 +228,7 @@ Closes browser with all the pages (if any were opened). The browser object itsel
 Browser websocket endpoint which could be used as an argument to
 [puppeteer.connect](#puppeteerconnectoptions). The format is `ws://${host}:${port}/browser/<id>`
 
-You can find the `webSocketDebuggerUrl` from `http://${host}:${port}/json/version`. More about the chrome devtools protocol is available at https://chromedevtools.github.io/devtools-protocol, the FAQ is currently available at [devtools-protocol/#55 (FAQ)](https://github.com/ChromeDevTools/devtools-protocol/issues/55) and will be added to the site soon.
+You can find the `webSocketDebuggerUrl` from `http://${host}:${port}/json/version`. More about the chrome devtools protocol is available at [https://chromedevtools.github.io/devtools-protocol](https://chromedevtools.github.io/devtools-protocol), the FAQ is currently available at [devtools-protocol/#55 (FAQ)](https://github.com/ChromeDevTools/devtools-protocol/issues/55) and will be added to the site soon.
 
 ### class: Page
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -226,9 +226,9 @@ Closes browser with all the pages (if any were opened). The browser object itsel
 - returns: <[string]> Browser websocket url.
 
 Browser websocket endpoint which could be used as an argument to
-[puppeteer.connect](#puppeteerconnectoptions). The format is `ws://${host}:${port}/page/<id>`
+[puppeteer.connect](#puppeteerconnectoptions). The format is `ws://${host}:${port}/browser/<id>`
 
-You can find the `webSocketDebuggerUrl` from `http://${host}:${port}/json`.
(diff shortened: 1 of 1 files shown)


JakeGinnivan:
I signed it!

docs/api.md
```diff
@@ -226,7 +226,9 @@ Closes browser with all the pages (if any were opened). The browser object itsel
 - returns: <[string]> Browser websocket url.
 
 Browser websocket endpoint which could be used as an argument to
-[puppeteer.connect](#puppeteerconnectoptions).
+[puppeteer.connect](#puppeteerconnectoptions). The format is `ws://${host}:${port}/page/<id>`
```
aslushnikov: ironically, this is page's debugging URL, not browser's. It might work for now, but might not eventually. 

The Browser WS is dumped by the browser in its stderr. Note: this only happens if the browser is launched with a `remote-debugging-port` flag.

Check out how we extract browser web socket from stderr in the [Launcher.js](https://github.com/GoogleChrome/puppeteer/blob/d7e673645af749bcdf104abe79829c9720c420a0/lib/Launcher.js#L137)

aslushnikov:
Hey @JakeGinnivan, thank you for the PR! Could you please share your user scenario - why would you need a way to get the browser endpoint manually?

JakeGinnivan:
Yeah sure.

We are doing visual regression testing and have mac, windows and linux machines. To remove differences like font rendering etc we are just using the docker container `alpeware/chrome-headless-unstable` and having it run on our build server and all developers machines in the background.

I figured it was the page debugging endpoint but couldn't figure out how to resolve the browsers endpoint.

aslushnikov:
@JakeGinnivan do you want to update this PR with a relevant information to ease the life of future generations? Feel free to close it if you don't.

paulirish:
btw see "How do I access the browser target?" on https://github.com/ChromeDevTools/devtools-protocol/issues/55 ..

i'll be moving that into the real site soon... so that is probably a better place to document the browserWs endpoint details than here.

(So i'd recommend nuking L231 and instead linking up that doc)

JakeGinnivan:
Will get this updated today, appreciate the info!

JakeGinnivan:
@paulirish I have done both for the moment. The redundant info can be removed later if need be but having it directly in this doc I think will be handy for people.

pierre-moire (commented):
Actually the url is /json and not /json/version to get the debugger url

paulirish:
@pierre-moire /json lists page endpoints, the browser endpoint is only available at /json/version.

docs/api.md
```diff
@@ -226,7 +226,9 @@ Closes browser with all the pages (if any were opened). The browser object itsel
 - returns: <[string]> Browser websocket url.
 
 Browser websocket endpoint which could be used as an argument to
-[puppeteer.connect](#puppeteerconnectoptions).
+[puppeteer.connect](#puppeteerconnectoptions). The format is `ws://${host}:${port}/browser/<id>`
+
+You can find the `webSocketDebuggerUrl` from `http://${host}:${port}/json/version`. More about the chrome devtools protocol is available at [https://chromedevtools.github.io/devtools-protocol](https://chromedevtools.github.io/devtools-protocol), the FAQ is currently available at [devtools-protocol/#55 (FAQ)](https://github.com/ChromeDevTools/devtools-protocol/issues/55) and will be added to the site soon.
```
paulirish: you can just update this to point to https://chromedevtools.github.io/devtools-protocol/#faq  

(it will deploy soon if it hasn't already)

JakeGinnivan:
@paulirish updated. 

The FAQ header is missing an anchor so I anchored directly to the interesting header

kachkaev:
Hi guys, a newbie here :–)

Seems like going to `http://localhost:9222/json/version` to figure out the `webSocketDebuggerUrl` did not help in my case. The json being returned is:

```json
{
   "Browser": "Chrome/61.0.3163.100",
   "Protocol-Version": "1.2",
   "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/61.0.3163.100 Safari/537.36",
   "V8-Version": "6.1.534.41",
   "WebKit-Version": "537.36 (@57c9d07b416b5a2ea23d28247300e4af36329bdc)"
}
```

Launching Chrome on macOS with `open '/Applications/Google Chrome.app' --args --remote-debugging-port=9222`.

Ultimately I'm trying to set up Chrome as a docker microservice and talk to it using puppeteer from a Node.js microservice. All works when the browser is in the same container and I launch it with `puppeteer.launch`, but the image becomes not so micro in this case 😄

PR #798, opened 2017-09-15
Added mkdir recursively in `page.screenshot` and `page.pdf`

tkdn:
Now `page.screenshot()` and `page.pdf()` can't receive directory that does not exist as `path` parameter.

`Error: ENOENT: no such file or directory, open`

So, this PR adds mkdir recuresively (if not exists) to these methods and adds some test.
Just me, nobody wants this one?

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -15,6 +15,7 @@
  */
 
 const fs = require('fs');
+const path = require('path');
 const EventEmitter = require('events');
 const mime = require('mime');
 const NetworkManager = require('./NetworkManager');
@@ -585,7 +586,7 @@ class Page extends EventEmitter {
 
     const buffer = new Buffer(result.data, 'base64');
     if (options.path)
-      fs.writeFileSync(options.path, buffer);
+      writeFileSyncWithMkdir(options.path, buffer);
     return buffer;
   }
 
@@ -633,7 +634,7 @@ class Page extends EventEmitter {
     });
     const buffer = new Buffer(result.data, 'base64');
     if (options.path)
-      fs.writeFileSync(options.path, buffer);
+      writeFileSyncWithMkdir(options.path, buffer);
     return buffer;
   }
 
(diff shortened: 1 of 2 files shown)


aslushnikov:
@tkdn I don't think puppeteer should do much outside of its explicit domain of browser automation. I'd rather offload this subdirectory creation to the clients.

Janpot:
imo it's already doing too much by writing to a file in the first place. Returning a buffer should be more than powerful enough.

ebidel:
I agree with @Janpot. If we're not going to add other conveniences to saving the file, the screenshot and pdf methods should probably just return buffers.

aslushnikov:
@Janpot @ebidel `page.screenshot` basic ability to write into a file is justified because it is very convenient. ~~and it allows us to have a streaming implementation under the hood.~~

However, the functionality should stay simple so that it's very cheap to maintain.

**Update:** saving to file doesn't help much with streaming since `page.screenshot` has to return full buffer anyway.

Janpot:
@aslushnikov I thought the devtools protocol sends the screenshot as one big base64 string. Which means the whole thing goes in memory anyway. I don't see what would be the added benefit of using streams under the hood.

aslushnikov:
> I thought the devtools protocol sends the screenshot as one big base64 string.

@Janpot both screenshots and PDF's are currently returned as base64. However, we've already had a few clients with big PDFs, so one day we might enhance devtools protocol with PDF streaming if there's a strong demand.

mediba-takeda:
ok. I understood `puppeteer`'s simple way and should be simple for devtools protocol in future.

Janpot:
Then, doesn't it make more sense to return a stream, rather than write a file? What if I want to postprocess it with transform streams and send as the response of the http request. That means I first have to let it write to some temp file, then open a readstream for it, pipe it to my transform and then to my response. Not to forget I need a bunch of filesystem management to remove all those temp files afterwards.

tkdn:
ok. I understood `puppeteer`'s simple way and should be simple for devtools protocol in future.

aslushnikov:
> doesn't it make more sense to return a stream, rather than write a file?

These things are not exclusive; there might be a method that returns stream and writes into a file:
```js
const stream = page.pdfStream({path: 'article.pdf'});
```

Method `page.pdf` returns buffer today because node streams are harder to work with and because our current pipeline is not "streamified".

Also: it occurred to me that `path` option doesn't help much with streams since we already return full buffer. I corrected [my comment](https://github.com/GoogleChrome/puppeteer/pull/798#issuecomment-330176173).

icssjs:
just copy

PR #801, opened 2017-09-15
Add keyboard methods to elementHandle

### LINKED ISSUE #241: Add keyboard.type
We had `keyboard.type` in our plan for quite some time, but it seems to get lost.

JoelEinbinder:
This adds `elementHandle.focus`, `elementHandle.press` and `elementHandle.type`. It adds an optional selector argument to `page.press` and `page.type` to align them with `page.click`, and adds `keyboard.press` and `keyboard.type`.

Closes #241

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -59,7 +59,6 @@
     + [page.mouse](#pagemouse)
     + [page.pdf(options)](#pagepdfoptions)
     + [page.plainText()](#pageplaintext)
-    + [page.press(key[, options])](#pagepresskey-options)
     + [page.reload(options)](#pagereloadoptions)
     + [page.screenshot([options])](#pagescreenshotoptions)
     + [page.select(selector, ...values)](#pageselectselector-values)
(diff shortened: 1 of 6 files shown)


docs/api.md
```diff
@@ -815,19 +821,20 @@ Shortcut for [page.mainFrame().title()](#frametitle).
 #### page.tracing
 - returns: <[Tracing]>
 
-#### page.type(text, options)
+#### page.type([selector, ]text[, options])
```
aslushnikov: let's make selector unconditional here and in `page.press`. This would align better with the rest of the API. Having selector optional here might create false expectations from other API methods.

We'll break all the beta users, but that's why we're unstable yet.

lib/ElementHandle.js
```diff
@@ -21,15 +21,13 @@ class ElementHandle {
    * @param {!Frame} frame
    * @param {!Connection} client
    * @param {!Object} remoteObject
-   * @param {!Mouse} mouse
-   * @param {!Touchscreen} touchscreen;
+   * @param {!Page} page
    */
-  constructor(frame, client, remoteObject, mouse, touchscreen) {
+  constructor(frame, client, remoteObject, page) {
```
aslushnikov: nit: let's plumb page as a first entry here.

aslushnikov (commented):
The code looks good. I'll take a closer look later

docs/api.md
JoelEinbinder: How do you feel above removing `page.press`

aslushnikov: > How do you feel above removing page.press

@JoelEinbinder i like it.

aslushnikov (approved):

PR #813, opened 2017-09-18
Docs: screenshot/pdf add note on path not existing

ebidel:


docs/api.md
```diff
@@ -637,7 +637,7 @@ Page is guaranteed to have a main frame which persists during navigations.
 
 #### page.pdf(options)
 - `options` <[Object]> Options object which might have the following properties:
-  - `path` <[string]> The file path to save the PDF to. If `path` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd). If no path is provided, the PDF won't be saved to the disk.
+  - `path` <[string]> The file path to save the PDF to. If `path` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd). If no path is provided or it doesn't exist, the PDF won't be saved to the disk.
```
aslushnikov: don't we throw if the folders don't exist?

ebidel: We do. Do you want to be that specific?

adelalsaggaf1: اكيد

aslushnikov: I'd state error explicitly: we do that in multiple other places in the doc.

ebidel: done PTAL

aslushnikov (approved):

docs/api.md
```diff
@@ -637,7 +637,7 @@ Page is guaranteed to have a main frame which persists during navigations.
 
 #### page.pdf(options)
 - `options` <[Object]> Options object which might have the following properties:
-  - `path` <[string]> The file path to save the PDF to. If `path` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd). If no path is provided, the PDF won't be saved to the disk.
+  - `path` <[string]> The file path to save the PDF to. If `path` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd). If no path is provided, the PDF won't be saved to the disk. If the path does not exist, the method throws an error.
```
aslushnikov, line 640: oh wait, this is confusing now: the path technically doesn't exist here:
```js
page.screenshot({path: 'screenshot.png'})
```
Why do we want to do anything about this statement? Was there a confusion somewhere on the bugtracker?

aslushnikov (changes requested):
(requesting changes to override previous "approved")

docs/api.md
ebidel: From https://github.com/GoogleChrome/puppeteer/pull/798

ebidel:
I'm not convinced we need this either.

PR #834, opened 2017-09-20
Add Page.clearDataForOrigin api

xfumihiro:


xfumihiro:
I signed it!

aslushnikov:
Hey @xfumihiro, thanks for the PR!

I like the method, but I don't like that it's per origin (origin is a tricky thing). Maybe we can come up with an API that would not expose the origin, e.g. `page.reload({hard: true})`. Could you please share your usecase?

In case we still go with this approach, we should have this method on Browser rather then on a page.

xfumihiro:
Hi @aslushnikov I'm using it to clear the state of my app (like auth state stored by reduxPersist).
I agree that this could be a Browser method.
Didn't think much about issues of exposing the origin as I'm just mirroring the devtools protocol api. 🤔

xfumihiro:
One thing to be concerned is: will this cause tests to be flaky?

xfumihiro:
@aslushnikov I've switched to use the whiltelist config in reduxPersist instead. (app shouldn't have state for acceptance tests in the first place)
So for now, I don't know if this api have a valid use case. ¯\_(ツ)_/¯ (Feel free to close it)

aslushnikov:
@xfumihiro ok thanks!

PR #837, opened 2017-09-20
document an important `page.setViewport()` use case

vsemozhetbyt:
Refs: https://github.com/GoogleChrome/puppeteer/issues/825

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -795,6 +795,8 @@ puppeteer.launch().then(async browser => {
 
 > **NOTE** in certain cases, setting viewport will reload the page in order to set the `isMobile` or `hasTouch` properties.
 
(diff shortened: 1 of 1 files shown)


Garbee (changes requested):
Given the context of the issue, this isn't something that really needs documenting. The issue is based on a media query on the viewport size. When developers are dealing with these kinds of scripts they should be thinking about that.

I'd drop the blockquote structure and simply add a mention to the existing paragraph about the default viewport size.

> The default viewport size is `800 x 600` pixels. In the case of multiple pages in a single browser, each page can have its own viewport size.

ebidel:
Not sure we need to mention this either. If anything, we may want to re-mention the default viewport is in the troubleshooting guide and suggest folks check that if they're running into unexpected behavior regarding rendering.

PR #838, opened 2017-09-20
Remove --no-sandbox from Docker example

ebidel:
part of https://github.com/GoogleChrome/puppeteer/issues/809

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -103,22 +103,36 @@ Dockerfile:
 ```
 FROM node:8-slim
 
-# Install latest chrome (dev) package.
-# Note: this also installs the necessary libs so we don't need the previous RUN command.
-RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - &&\
-sh -c 'echo "deb http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list' &&\
-apt-get update &&\
-apt-get install -y google-chrome-unstable
+# Install latest chrome dev package.
+# Note: this installs the necessary libs to make the bundled version of Chromium that Pupppeteer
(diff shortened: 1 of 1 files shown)


docs/troubleshooting.md
```diff
@@ -103,22 +103,36 @@ Dockerfile:
...
+    && rm -rf /src/*.deb
 
 # Uncomment to skip the chromium download when installing puppeteer. If you do,
 # you'll need to launch puppeteer with:
 #     browser.launch({executablePath: 'google-chrome-unstable'})
 # ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true
 
-# Install puppeteer so it can be required by user code that gets ran in server.js.
+# Install puppeteer so it's available in the container.
 RUN yarn add puppeteer
```
Garbee, line 125: Shouldn't this be `yarn global add puppeteer`?

ebidel: It doesn't need to be right? `node_modules/puppeteer` will be installed in the root of the container and that will be resolvable by user code.

Garbee: Not from what I'm [seeing in the resolution docs](https://docs.npmjs.com/files/folders). It says it will look in the global install space folder if not found. In this case that install space is [/opt/yarn](https://github.com/nodejs/docker-node/blob/b1d4ff57bc3347818a0977b772d44f8f0753868a/8.5/slim/Dockerfile#L61) which means it will look in `/opt/yarn/lib/node_modules`. Which it would only get there with `yarn global add puppeteer`.

I'm not seeing any mention of checking simply in the root file-system space for a module folder. Could be missing it though or these could not be the full chain instructions.

ebidel: Added `chown -R pptruser:pptruser /node_modules` works

aslushnikov (approved):

PR #853, opened 2017-09-22
Support npm config support for download skipping flag

gordomium:
Add `process.env.npm_config_puppeteer_skip_chromium_download` support with document.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -151,7 +151,6 @@ Puppeteer looks for certain [environment variables](https://en.wikipedia.org/wik
 
 - `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
 - `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step. (`puppeteer_skip_chromium_download` in `.npmrc` as same)
-- `PUPPETEER_DOWNLOADS_FOLDER` - specify an absolute path to hold downloads. (`puppeteer_downloads_folder` in `.npmrc` as same)
 ### class: Puppeteer
 
 Puppeteer module provides a method to launch a Chromium instance.
--- a/docs/api.md
+++ b/docs/api.md
@@ -150,7 +150,8 @@ Puppeteer is a Node library which provides a high-level API to control Chromium 
 Puppeteer looks for certain [environment variables](https://en.wikipedia.org/wiki/Environment_variable) to aid its operations:
 
 - `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
-- `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step.
+- `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step. (`puppeteer_skip_chromium_download` in `.npmrc` as same)
+- `PUPPETEER_DOWNLOADS_FOLDER` - specify an absolute path to hold downloads. (`puppeteer_downloads_folder` in `.npmrc` as same)
 ### class: Puppeteer
 
 Puppeteer module provides a method to launch a Chromium instance.
--- a/install.js
+++ b/install.js
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-if (process.env.PUPPETEER_SKIP_CHROMIUM_DOWNLOAD) {
-  console.log('**INFO** Skipping Chromium download. "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" environment variable was found.');
+if (process.env.npm_config_puppeteer_skip_chromium_download || process.env.PUPPETEER_SKIP_CHROMIUM_DOWNLOAD) {
+  console.log('**INFO** Skipping Chromium download. "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" npm_config/environment variable was found.');
   return;
 }
 
--- a/install.js
+++ b/install.js
@@ -14,8 +14,12 @@
  * limitations under the License.
  */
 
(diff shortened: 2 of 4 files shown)


gordomium:
I signed it!

gordomium:
Again email issue?

ebidel:
I wasn't too familiar with `npm config`. https://docs.npmjs.com/misc/config. Haven't see them used too often over standard env variables. Is the main benefit that you can specify these in an `.npmrc` file? Do projects typically allow for both an env and the `npm_config_` version?

gordomium:
@ebidel 

> Is the main benefit that you can specify these in an `.npmrc` file?

Yes. It can make env settings into project VCS if puppeteer used as a node module. No need to assign env every time.

> Do projects typically allow for both an env and the `npm_config_` version?

I think so. For example: `electron-download` https://github.com/electron-userland/electron-download/blob/master/lib/index.js#L27

gordomium:
I signed it!

gordomium:
I signed it!

utils/ChromiumDownloader.js
```diff
@@ -25,7 +25,10 @@ const removeRecursive = require('rimraf');
 const ProxyAgent = require('https-proxy-agent');
 const getProxyForUrl = require('proxy-from-env').getProxyForUrl;
 
-const DOWNLOADS_FOLDER = path.join(__dirname, '..', '.local-chromium');
+const DOWNLOADS_FOLDER = process.env.npm_config_puppeteer_downloads_folder ||
+                          process.env.PUPPETEER_DOWNLOADS_FOLDER ||
+                          path.join(__dirname, '..', '.local-chromium');
+if (!DOWNLOADS_FOLDER.startsWith('/')) throw new Error('Puppeteer downloads folder must be an absolute path.');
```
aslushnikov: this check will not work on windows

install.js
```diff
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 
-if (process.env.PUPPETEER_SKIP_CHROMIUM_DOWNLOAD) {
-  console.log('**INFO** Skipping Chromium download. "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" environment variable was found.');
+if (process.env.npm_config_puppeteer_skip_chromium_download || process.env.PUPPETEER_SKIP_CHROMIUM_DOWNLOAD) {
```
aslushnikov: electron-download also check for capitalized npm_config env variable. Shall we too?

install.js
```diff
@@ -14,8 +14,8 @@
...
+  console.log('**INFO** Skipping Chromium download. "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" npm_config/environment variable was found.');
```
aslushnikov: let's console log explicitly why we're skipping download. It will save someone time to debug this behavior.

```js
if (process.env.PUPPETEER_SKIP_CHROMIUM_DOWNLOAD) {
  console.log('**INFO** Skipping Chromium download. "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" environment variable was found.');
  return;
}
if (process.env.npm_config_puppeteer_skip_chromium_download) {
  console.log('**INFO** Skipping Chromium download. "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" was set in npm config.');
  return;
}

```

utils/ChromiumDownloader.js
```diff
@@ -25,7 +25,10 @@ const removeRecursive = require('rimraf');
 const ProxyAgent = require('https-proxy-agent');
 const getProxyForUrl = require('proxy-from-env').getProxyForUrl;
 
-const DOWNLOADS_FOLDER = path.join(__dirname, '..', '.local-chromium');
+const DOWNLOADS_FOLDER = process.env.npm_config_puppeteer_downloads_folder ||
```
aslushnikov: the variable `npm_config_puppeteer_download_folder` is defined only when script is run as part of npm step. This is not always the case for `ChromiumDownlaoder`: it is used both from `install.js` (npm script) and `lib/Launcher.js` (not an npm script).

docs/api.md
```diff
@@ -150,7 +150,8 @@ Puppeteer is a Node library which provides a high-level API to control Chromium
 Puppeteer looks for certain [environment variables](https://en.wikipedia.org/wiki/Environment_variable) to aid its operations:
 
 - `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
-- `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step.
+- `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step. (`puppeteer_skip_chromium_download` in `.npmrc` as same)
+- `PUPPETEER_DOWNLOADS_FOLDER` - specify an absolute path to hold downloads. (`puppeteer_downloads_folder` in `.npmrc` as same)
```
aslushnikov: let's put `PUPPETEER_DOWNLOADS_FOLDER` before `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`. Alphabetic order will help to navigate the section.

aslushnikov (commented):
Hey @gordomium, thanks for the PR!

I'm curious why would you need the `PUPPETEER_DOWNLOADS_FOLDER` env variable?

utils/ChromiumDownloader.js
gordomium: Yes. This is expected behavior that specifying `DOWNLOADS_FOLDER` in any time with `npm_config_`. But this change may spark more usage issues, so i will remove this specifying.

install.js
gordomium: You're right.

gordomium:
@aslushnikov 

> I'm curious why would you need the PUPPETEER_DOWNLOADS_FOLDER env variable?

At last, i removed that for less usage issues.

aslushnikov (approved):

aslushnikov:
@gordomium thanks for the patch!

Issue #877, opened 2017-09-26
Adopt plznavigate in puppeteer


aslushnikov:
1. Recentrly, [plznavigate](https://groups.google.com/a/chromium.org/forum/#!topic/chromium-dev/sOs05JfmPFw) was enabled on Chromium Trunk. 
2. plznavigate changed protocol behavior in a few subtle ways that we relied upon in pptr
3. plznavigate causes a protocol bug that blocks our adoption
    - [crbug.com/767683](https://bugs.chromium.org/p/chromium/issues/detail?id=767683)
 

For the reference: a wip of pptr migration over to plznvigate: [aslushnikov/puppeteer:plznavigate](https://github.com/aslushnikov/puppeteer/tree/plznavigate)

The plan is to roll with `--disable-browser-side-navigation` flag passed over to launcher and
remove it as soon as the upstream is fixed.

aslushnikov:
Two blocking upstream bugs:
- [crbug.com/759388](https://crbug.com/759388)
- [crbug.com/767683](https://crbug.com/767683) - landed as [r509486](http://crrev.com/509486)

PR #878, opened 2017-09-26
[roll] Roll chromium to r503964

aslushnikov:
This patch rolls chromium to r503964.

**Note:** since the plznavigate is not supported by puppeteer right now, the
patch also starts passing the `--disable-browser-side-navigation` flag.
This is a temporary work around for us.

References #877.


pavelfeldman:
LGTM

PR #879, opened 2017-09-26
[fix]: Removed synchronous operation with file system.

CheerlessCloud:
I'm completely discouraged by this code. @aslushnikov How could you write such a thing?
I would somehow understand this solution if the code were written on callbacks. But you use the **async
 function**! And inside the asynchronous function you are happy to block the EventLoop. What for?

If someone tells me where to put this "promisify", I'll move it without problems. But still I recommended somehow to solve this problem in the near future.

The remaining cases use synchronous writing and reading are not as critical, but it is advisable to fix them too: [1](https://github.com/GoogleChrome/puppeteer/blob/master/lib/Launcher.js#L62), [2](https://github.com/GoogleChrome/puppeteer/blob/master/lib/Tracing.js#L69).


CheerlessCloud:
I signed it!

Garbee (approved):
LGTM

lib/Page.js
```diff
@@ -26,6 +26,12 @@ const {Keyboard, Mouse, Touchscreen} = require('./Input');
 const Tracing = require('./Tracing');
 const {helper, debugError} = require('./helper');
 
+function writeFileAsync(...options) {
```
ebidel: @aslushnikov , should this be added in helper? I could use it in https://github.com/GoogleChrome/puppeteer/pull/881.

CheerlessCloud: As it seems to me, in a certain file/folder "utils" or "helpers" it is necessary to make the function promisify (as [here](https://github.com/CheerlessCloud/puppeteer/blob/master/utils/doclint/SourceFactory.js#L150)) and use it to create wrappers of the fs methods or some other in local module scope (as [here](https://github.com/CheerlessCloud/puppeteer/blob/master/utils/doclint/SourceFactory.js#L20)). Or we can add some external dependency, such as Bluebird or [pify](https://github.com/sindresorhus/pify).
IMHO, I, as a person external to the project, can not make such decisions. And I do not want to rewrite the code, so let's work out solution and I will implement it in this PR. This is almost my first PR to open source project, so I may not know some rules, sorry =)

aslushnikov: @CheerlessCloud let's add both `promisify` and `writeFileAsync` into helper.

aslushnikov:
Hey @CheerlessCloud, thank you for the pull request!

> I'm completely discouraged by this code. @aslushnikov How could you write such a thing?
I would somehow understand this solution if the code were written on callbacks. But you use the async
function! And inside the asynchronous function you are happy to block the EventLoop. What for?

`writeFileSync` is bad. Thanks to your passion, the software is getting better!

> The remaining cases use synchronous writing and reading are not as critical, but it is advisable to fix them too: 1, 2.

I'd be happy to review PRs.

CheerlessCloud:
Do not you think that the file "helper" for 300+ lines is antipattern? What about separate his to files, single file for one method? Yes, this does not completely solve the problem, but it will make it a little easier.

aslushnikov:
> Do not you think that the file "helper" for 300+ lines is antipattern? 

I think its fine.

lib/Launcher.js
```diff
@@ -122,7 +127,7 @@ class Launcher {
           process.kill(-chromeProcess.pid, 'SIGKILL');
         // Attempt to remove temporary profile directory to avoid littering.
         try {
-          removeSync(temporaryUserDataDir);
```
aslushnikov: This should stay sync: handlers for process `exit` are expected to complete synchronously.

lib/Tracing.js
```diff
@@ -13,8 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-const fs = require('fs');
 const {helper} = require('./helper');
+const fs = require('fs');
+
+fs.openAsync = helper.promisify(fs.open);
```
aslushnikov: let's not spoil the globals

```js
const openAsync = helper.promisify(fs.open);
const writeAsync = helper.promisify(fs.write);
const closeAsync = helper.promisify(fs.close);
```

lib/Launcher.js
```diff
@@ -15,16 +15,18 @@
  */
 const os = require('os');
 const path = require('path');
-const removeSync = require('rimraf').sync;
+const {helper} = require('./helper');
+const rimrafAsync = helper.promisify(require('rimraf'));
```
aslushnikov: nit: let's name this `removeAsync`: the `rimraf` doesn't say much of what the function is doing.

lib/Launcher.js
```diff
@@ -15,16 +15,18 @@
...
 const childProcess = require('child_process');
 const Downloader = require('../utils/ChromiumDownloader');
 const Connection = require('./Connection');
 const Browser = require('./Browser');
 const readline = require('readline');
 const fs = require('fs');
-const {helper} = require('./helper');
 const ChromiumRevision = require('../package.json').puppeteer.chromium_revision;
 
+fs.mkdtempAsync = helper.promisify(fs.mkdtemp);
```
aslushnikov: let's not spoil the global:

```js
const mkdtempAsync = helper.promisify(fs.mkdtemp);
```

lib/helper.js
```diff
@@ -225,6 +226,86 @@ class Helper {
...
+
+  /**
+   * @description Add to object for each method promisified analog by ${methodName}Async
+   * @static
+   * @param {Object} object
+   * @example
+   *  const fs = Helper.promisifyAll(require('fs'));
+   *  fs.writeFileAsync(...); // -> Promise
+   */
+  static promisifyAll(object) {
```
aslushnikov: we don't need this method

lib/helper.js
```diff
@@ -225,6 +226,86 @@ class Helper {
   static isNumber(obj) {
     return typeof obj === 'number' || obj instanceof Number;
   }
+
+  /**
```
aslushnikov: let's not fallback to `util` - the `helper.promisify` implementation is simple enough for our use case. It's easier to read and maintain a single code path.

```js
/**
 * @param {function(?)} nodeFunction
 * @return {function(?):!Promise<?>}
 */
static promisify(nodeFunction) {
  /**
   * @param {!Array<?>} options
   * @return {!Promise<?>}
   */
  return function(...options) {
    return new Promise(function(fulfill, reject) {
      options.push(callback);
      nodeFunction.call(null, ...options);
      function callback(err, result) {
        if (err)
          reject(err);
        else
          fulfill(result);
      }
    });
  };
}
```

lib/helper.js
```diff
@@ -225,6 +226,86 @@ class Helper {
...
+   * @static
```
aslushnikov: we don't use `@static`, `@throws` and `@example` declarations. They are harmless, but I'd prefer them to be removed for consistency with the rest of the code.

lib/Launcher.js
```diff
@@ -90,8 +92,11 @@ class Launcher {
       chromeProcess.once('close', () => {
         // Cleanup as processes exit.
         if (temporaryUserDataDir)
-          removeSync(temporaryUserDataDir);
-        fulfill();
+          rimrafAsync(temporaryUserDataDir)
```
aslushnikov: nit: braces are required around multi-line branches:

```js
if (temporaryUserDataDir) {
  rimrafAsync(..)
     ..
     ....
} else {
  fulfill();
}
```

lib/Launcher.js
```diff
@@ -90,8 +92,11 @@ class Launcher {
...
+            .then(() => fulfill())
+            .catch(err => console.error(err));
```
aslushnikov: let's reject the `waitForChromeToClose` promise

lib/helper.js
CheerlessCloud: By "not fallback to util" you mean using util.promisify.custom symbol?

lib/helper.js
CheerlessCloud: Well, I can remove it, but it may be needed in the future.

lib/helper.js
aslushnikov: yes, and also the `util.promisify` method.

I'd just stick with the simple shim we already have somewhere.

lib/helper.js
aslushnikov: Thanks, we'll add it if/when the need arises.

lib/Launcher.js
CheerlessCloud: In a few cases we don't catch errors from this promise, so we can get an unhandledRejection.

petrgrishin:
👍

lib/Launcher.js
```diff
@@ -86,12 +89,16 @@ class Launcher {
       chromeProcess.stderr.pipe(process.stderr);
     }
 
-    const waitForChromeToClose = new Promise(fulfill => {
+    const waitForChromeToClose = new Promise((fulfill, reject) => {
```
graingert, line 113: Don't mix `new Promise`, `x.then` and `async/await`. Convert callback apis into promises ASAP, then switch back to the highest abstraction.

```js
const waitForCloseEvent = () => new Promise(resolve => chromeProcess.once('close', resolve);
const waitForChromeToClose = (async () => {
  await waitForCloseEvent();
  if (temporaryUserDateDir) {
    try {
      await removeFolderAsync(temporaryUserDataDir);
    } catch(e) {
       console.error(e); // are you sure, what's wrong with letting it throw?
    }
  }
})();

File: lib/Page.js [hunk 12]
```diff
@@ -585,7 +587,7 @@ class Page extends EventEmitter {
 
     const buffer = new Buffer(result.data, 'base64');
     if (options.path)
-      fs.writeFileSync(options.path, buffer);
+      await writeFileAsync(options.path, buffer);
```
graingert, line 660: Why is there even an option to write the file here? Either puppeteer should return the actual stream or users should just write the buffer themselves.

lib/helper.js
```diff
@@ -225,6 +225,43 @@ class Helper {
   static isNumber(obj) {
     return typeof obj === 'number' || obj instanceof Number;
   }
+
+  /**
+   * @param {function(?)} nodeFunction
+   * @return {function(?):!Promise<?>}
+   */
+  static promisify(nodeFunction) {
```
graingert: why not use @ljharb's wonderful https://www.npmjs.com/package/util.promisify

lib/Launcher.js
```diff
@@ -86,12 +89,16 @@ class Launcher {
...
       chromeProcess.once('close', () => {
         // Cleanup as processes exit.
-        if (temporaryUserDataDir)
-          removeSync(temporaryUserDataDir);
-        fulfill();
+        if (temporaryUserDataDir) {
+          removeFolderAsync(temporaryUserDataDir)
+              .then(() => fulfill())
+              .catch(err => console.error(err));
```
graingert, line 119: why not just let this throw, like removeSync would have done.

graingert (changes requested):

lib/Launcher.js
aslushnikov, line 113: nit: `reject` is not used

aslushnikov: @graingert I don't see much benefits in the approach you suggested, nor it is easier for me to read. I'd say the current code is just good enough.

lib/Page.js
aslushnikov: @graingert two reasons:
- it comes handy from user standpoint: the majority of `page.screenshot` usages have `path` option in it.
- it's cheap from the project standpoint.

lib/helper.js
aslushnikov: @graingert no need to pull in a dependency for 30 lines of code.

lib/helper.js
```diff
@@ -225,6 +225,43 @@ class Helper {
...
+    if (typeof nodeFunction !== 'function')
+      throw new TypeError('A function is expected');
+
+    // Because native util.promisify differently work with arguments
```
aslushnikov: nit: let's drop the comment

lib/helper.js
```diff
@@ -225,6 +225,43 @@ class Helper {
...
+          else
+            // Otherwise return array of arguments
+            return resolve(result);
+        }
+        args.push(callback);
+        nodeFunction.call(null, ...args);
+      });
+    }
+
+    promisified.__isPromisified__ = true;
```
aslushnikov: this is never used - let's drop it.

lib/helper.js
```diff
@@ -225,6 +225,43 @@ class Helper {
...
+     * @param {!Array<?>} args
+     * @return {!Promise<?>}
+     */
+    function promisified(...args) {
+      return new Promise((resolve, reject) => {
+        function callback(err, ...result) {
+          if (err)
+            return reject(err);
+
+          // If callback invoked with single argument, we return in in Promise directly
```
aslushnikov: nit: let's drop the comment, the code is self explanatory

lib/helper.js
```diff
@@ -225,6 +225,43 @@ class Helper {
   static isNumber(obj) {
     return typeof obj === 'number' || obj instanceof Number;
   }
+
+  /**
+   * @param {function(?)} nodeFunction
+   * @return {function(?):!Promise<?>}
+   */
+  static promisify(nodeFunction) {
+    if (typeof nodeFunction !== 'function')
```
aslushnikov: Let's drop the check - the API is for internal use only.

lib/Launcher.js
aslushnikov: @graingert to avoid `unhandledPromiseRejection`.

lib/Launcher.js
```diff
@@ -86,12 +89,16 @@ class Launcher {
...
 
-    const waitForChromeToClose = new Promise(fulfill => {
+    const waitForChromeToClose = new Promise((fulfill, reject) => {
       chromeProcess.once('close', () => {
         // Cleanup as processes exit.
-        if (temporaryUserDataDir)
-          removeSync(temporaryUserDataDir);
-        fulfill();
+        if (temporaryUserDataDir) {
+          removeFolderAsync(temporaryUserDataDir)
```
aslushnikov, line 117: @CheerlessCloud I've seen your comment that the promise is not not catched in a few places, thanks for the clarification.

let's fulfill regardless of the temprorary data directory cleanup, and use `debugError` instead of `console.error`:

```js
if (temporaryUserDataDir) {
  removeFolderAsync(temporaryUserDataDir)
    .catch(debugError)
    .then(fulfill);
} else {
  fulfill();
}
```

aslushnikov (commented):
Thank you @CheerlessCloud for working on this. I left a few minor comments before we can merge this.

lib/Launcher.js
graingert: @aslushnikov I disagree, new Promise should not be mixed with x.then.

lib/Launcher.js
graingert: I want to be able to try/catch the error that removing the temporaryUserDataDir throws. You should await it and throw.

graingert (changes requested):

lib/Launcher.js
graingert: presumably people are going to `await launch();` and thus catch any errors

lib/helper.js
ljharb: you can use https://npmjs.com/util.promisify if you want a compatible shim.

lib/helper.js
ljharb: when it's reliable and faster in modern node, it's always better to have more deps :-)

lib/Launcher.js
aslushnikov: How's `Page.pdf` related to this code?

lib/Launcher.js
aslushnikov: @graingert what are you going to do with the exception? The `temporaryUserDataDir` is internal to launcher's implementation.

lib/Launcher.js
aslushnikov: @graingert I might be misunderstanding something. Lets split this discussion out of this thread: would you mind sending a separate PR that does changes that you find important and explains why are they necessary?

aslushnikov:
@CheerlessCloud any updates on this? I'm very interested in this going in.

CheerlessCloud:
@aslushnikov Sorry, I've been busy the last few days. I'll try to finish it in the near future)

lib/Launcher.js
graingert: @aslushnikov it's the API that calls the download

graingert: @aslushnikov ah you're right. Edited comment

aslushnikov:
Hey @CheerlessCloud!

I see you're pressed on time. I'd like to add a few commits to this PR to address all the comments and merge this in. Let me know if this would work for you.

Issue #478, opened 2017-08-22
startScreencast feature?


ghost:
Would be great to screencast the page.

Garbee:
Why screencast a headless display? If you want to see it, simply don't start it in headless mode.

ghost:
Good point. In other words if there is a need to screencast page - one have to use Chromium without headless mode.

On the other hand puppeteer supports non-headless mode too. Would be great that Puppetier would support this: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-startScreencast

nigeon:
+1

aslushnikov:
@Mykolas-Molis how do you want to consume it? Do you want to save a screencast as a video file?

michaelshiel:
I've implemented this for myself over here: [https://github.com/michaelshiel/puppeteer](https://github.com/michaelshiel/puppeteer) (Page emits an event for each frame) but I'm not sure what kind of tests would be appropriate, or how to describe/structure them, any suggestions?

sul4bh:
@aslushnikov I would like to have this feature implemented too. My use case is recording the screen as a video.

paulirish:
+1 to video capture start/stop/save API

sul4bh:
Not sure where to ask this question but here it goes:
What does the document here(https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-startScreencast) mean by a frame? What is the maximum frame-rate that I might be able to get by using `startScreencast()`?

I experimented with `startScreencast()` but I was only able to get about 5fps (meaning 5 png images/frames in 1 second). On headless mode I got only 1 fps. This is a very low fps for video.

michaelshiel:
I think it depends on how much the page is changing, in my experience it will stop sending frames if there are no changes on the page. I tried it on a youtube video, and I was getting at least 20 frames per second.

sul4bh:
@michaelshiel you are right. I played a 25fps youtube video and got an average of 25 "frames" per second. ~~EDIT: Can't verify if the frames are generated only when the content of the screen changes.~~ EDIT: Verified

Also I got really high fps today. Not sure why it was low yesterday. Probably Chrome Canary got updated or something...

Also, it would be nice to have a `Viewport` option for `startScreencast()`. (not sure where to request for a feature)

KeizerDev:
Check this issue: https://github.com/ChromeDevTools/devtools-protocol/issues/17

imolorhe:
[...]

PR #881, opened 2017-09-26
Screencast API

ebidel:
Fixes https://github.com/GoogleChrome/puppeteer/issues/478.

Want to get some feedback on this before finishing tests and api docs.

The shape of the API is similar to Tracing. This PR introduces `page.screencast`:

```js
page.screencast.on('frame', frame => {
  ...
});

await page.screencast.start();
await page.goto(...);
// ... other stuff ...
const frames = await page.screencast.stop();

// user can save frames themselves
frames.forEach((frame, i) => {
  fs.writeFileSync(`frame_${i}.png`, new Buffer(frame, 'base64'));
});
```

Passing a `path` creates a .webm video file when you call `stop()`. The solution uses **no dependencies**, only web platform APIs! The downside is that capturing a 'recording' is real-time, meaning longer videos will take more time to complete.

```js
await page.screencast.start({path: 'video.webm'});
```

### DIFF
--- a/lib/Screencast.js
+++ b/lib/Screencast.js
@@ -39,7 +39,7 @@ async function createVideo(frames, fps) {
   canvas.style.width = `${WIDTH / 2}px`;
   canvas.style.height = `${HEIGHT / 2}px`;
 
-  // Create video stream from canvas. Since we're not passing an fsp, it will
+  // Create video stream from canvas. Since we're not passing an fps, it will
   // match how fast the canvas is being drawn to.
   const stream = canvas.captureStream();
   const recorder = new MediaRecorder(stream, {mimeType: 'video/webm'});
--- a/lib/Screencast.js
+++ b/lib/Screencast.js
@@ -36,8 +36,6 @@ async function createVideo(frames, fps) {
   canvas.style.display = 'none';
   canvas.width = WIDTH;
   canvas.height = HEIGHT;
-  canvas.style.width = `${WIDTH / 2}px`;
-  canvas.style.height = `${HEIGHT / 2}px`;
 
   // Create video stream from canvas. Since we're not passing an fps, it will
   // match how fast the canvas is being drawn to.
--- a/lib/Screencast.js
+++ b/lib/Screencast.js
@@ -1,0 +1,158 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+const fs = require('fs');
+const path = require('path');
+const EventEmitter = require('events');
+const {helper, debugError} = require('./helper');
+
+
+class Screencast extends EventEmitter {
+  /**
+   * @param {!Session} client
+   */
+  constructor(client) {
+    super();
+    this._client = client;
+    /** @type {!Array<!Object>} */
+    this._frames = [];
+    /** @type {string} */
+    this._format = 'png';
+    /** @type {boolean} */
+    this._recording = false;
+
+    this._client.on('Page.screencastFrame', event => this._onScreencastFrame(event));
+  }
+
+  /**
+   * @return {!Array<!Screencast.Frame>}
+   */
+  frames() {
+    return this._frames;
+  }
+
+  /**
+   * Format of each frame.
+   * @return {!string}
+   */
+  format() {
+    return this._format;
+  }
+
+  /**
+   * @param {!Object} event
+   */
+  async _onScreencastFrame(event) {
+    console.assert(typeof event.sessionId === 'number',
+        'Expected screencast frame to have sessionId');
+    await this._client.send(
+        'Page.screencastFrameAck', {sessionId: event.sessionId}).catch(debugError);
+    const buffer = new Buffer(event.data, 'base64');
+    this.frames().push(buffer);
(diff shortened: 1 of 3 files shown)


ebidel:
There also appears to be bugs in the DTP screen api:
- Navigating to another page while a screencast is running changes the viewport. Therefore, causing screenshots to be different sizes.
- DPR always defaults to 1.
- It would be nice if there were an option to specify FPS.

e.g. Using `page.setViewport({width: 1000, height: 800, deviceScaleFactor: 2});` and then navigating to a new page:

DPR, width, height
screen shot 2017-09-26 at 9 08 56 am

lib/Screencast.js
```diff
@@ -0,0 +1,236 @@
...
+
+  // Use last image as width so we know the set viewport is correct.
+  const WIDTH = imageBitmaps[imageBitmaps.length - 1].width;
+  const HEIGHT = imageBitmaps[imageBitmaps.length - 1].height;
+
+  const canvas = document.createElement('canvas');
+  canvas.style.display = 'none';
+  canvas.width = WIDTH;
+  canvas.height = HEIGHT;
+  canvas.style.width = `${WIDTH / 2}px`;
```
ebidel: Typically you scale a canvas according to the DPR of the device so the images are clear. IOW, if the screencast api captured images at 1000x800 and DPR was 2, you'd want to create a 500x400 canvas. I think we can probably remove this since 1.) we're not accounting for DPR here and 2.) The DTP screen API doesn't respect DPR.

graingert:
can we get a stream api?

```js
const express = require('express');
const app = express();

async function screencastSteps(page) {
  await page.goto(...);
  // ... other stuff ...
}

app.get('/goofy', function(req, res) {
  res.setHeader('Content-Type', 'video/ogg');
  page.screencast.start(screencastSteps).pipe(res);
});

app.listen(9999);
```

aslushnikov:
I especially love the "no dependencies" part, impressive.

Can you write frames iteratively so that you don't accumulate them all in the memory?

ebidel:
For producing image frames, users can do: 

```js
const frames = [];
page.screencast.on('frame', frame => {
  frames.push(frame.buffer);
});
```

For the video chunks, it looks like `MediaRecorder.start()` accepts a time slice param:

> If a timeslice property was passed into the MediaRecorder.start() method that started media capture, a dataavailable event is fired every timeslice milliseconds. That means that each blob will have a specific time duration (except the last blob, which might be shorter, since it would be whatever is left over since the last event). So if the method call looked like this — recorder.start(1000); — the dataavailable event would fire after each second of media capture, and our event handler would be called every second with a blob of media data that's one second long. You can use timeslice alongside MediaRecorder.stop() and MediaRecorder.requestData() to produce multiple same-length blobs plus other shorter blobs as well.

We can send back video chunks at 1s intervals or something, but I'm not sure how useful individual small chunks would be in practice. Users would still need to put them together. That said, it's just a one liner (`new Blob([chunks], {type: 'video/webm'})`) and might be nice to have for larger videos.

https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/ondataavailable

Thoughts?

graingert:
@ebidel using an async iterator would allow users to request frames whenever they need them (it's a pull based async collection)

ebidel:
@graingert async iterators are still a ways off. http://node.green/#ESNEXT-candidate--stage-3--Asynchronous-Iterators-for-await-of-loops

RobertoUa:
Is it possible to support fixed FPS in Puppeteer? Or should it be first implemented in DevTools Protocol?

ebidel:
@RobertoUa DT would need to support that. Once this PR lands, you could also write a script that encodes the frames at the desired rate, using ffmpeg or something.

ebidel:
@aslushnikov anything else we want to change before I add tests and document? 

> Can you write frames iteratively so that you don't accumulate them all in the memory?

Were you referring to image frames or video chunks? For the former, we could write frames as they happen but I think that would mean changing the API to look like this:

```
await page.screencast.start({
  path: './' // optional. If specified, writes each frame image to the directory. Required if the `file` arg is specified.
  file: './path/to/video.webm', // optional. Creates the video file when `.stop()` is called.
});`
```

or make users save frames themselves and call an extra method to produce the video:

```
const frames = [];
page.screencast.on('frame', frame => {
  frames.push(frame.buffer);
});

await page.screencast.start();
// ... later ...
await page.screencast.stop(); // wouldn't return an array of frames :(
// ... later ...
const videoFileBuffer = await page.screencast.video(frames, {path: 'screencast.webm'});
```

...not as nice IMO but wouldn't store any frames in mem.

graingert:
@ebidel how about https://github.com/GoogleChrome/puppeteer/pull/881#issuecomment-332975524

graingert:
Then you don't need a "start/stop"

ebidel:
@graingert looks neat, but that's a different shape than the rest of the codebase/API. We don't use streams anywhere else AFAIK. In the future, I'm imaging we could add other APIs like `page.screencast.pause()` to start/stop screencast from different parts of a script. Not necessarily tie the screencasting to a list of actions in a callback.

julien-c:
I would also like to get video chunks iteratively (to stream them to ffmpeg for transcoding, for instance)

julien-c:
Also wanted to add that this is _really_ exciting. As someone who was generating frames one by one in Phantom.js to create videos, there are so many things that this is going to enable. Great job @ebidel 🙏

lib/Screencast.js
```diff
@@ -0,0 +1,234 @@
...
+    /** @type {!Array<!Object>} */
+    this._frames = [];
+    /** @type {string} */
+    this._format = 'png';
+    /** @type {string} */
+    this._path = '';
+    /** @type {boolean} */
+    this._recording = false;
+
+    this._client.on('Page.screencastFrame', event => this._onScreencastFrame(event));
```
aslushnikov, line 110: we don't want to use this api for video streaming; it gonna be slow and low-quality. We'd need to introduce a nicer API for the protocol first to have this feature in pptr.

aslushnikov:
Closing this in favor of upstream implementation: https://bugs.chromium.org/p/chromium/issues/detail?id=781117

azohra:
:( seems like the upstream stalled out almost right after it started. Is anyone able to shed light of if this feature is still being pursued? Thanks for all the hard work! #478 and this were a very interesting read.

PR #887, opened 2017-09-26
Allow options.timeout to be `0` to disable timeout

### LINKED ISSUE #782: Constantly getting Navigation Timeout Error
### Steps to reproduce

* Puppeteer version: 0.10.2
* Platform / OS version: Ubuntu / node v8.5.0
* URLs (if applicable): https://www.883jia.com.sg, https://www.power98.com.sg/

```

(async() => {
  page.on('console', (...args) => {
    for (let i = 0; i < args.length; ++i)
    console.log(`${i}: ${args[i]}`);
  });
  page.on('error', err => {
    console.log(err)
  })
  await page.goto('https://www.883jia.com.sg/');
  console.log(page.url());
  browser.close();
})()
```

Successful logging of page.url() then closure of the instance.

**What happens instead?**

```
0: JQMIGRATE: Migrate is installed, version 1.4.1
Error: Navigation Timeout Exceeded: 30000ms exceeded
```

Not really sure what happened, pretty new to this and all, do let me know what other information is needed. Would suspect that it is due to the JQMIGRATE message from the website though.

sukrosono:
Intended to close #782 , sorry if this one look really premature.

As the [first point](https://github.com/GoogleChrome/puppeteer/blob/master/CONTRIBUTING.md#writing-tests) 

> every feature should be accompanied by a test
are we count this one count as feature? if yes is this correct way to test it?

Clarification:
the puppeteer launch [options](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions) is documented accept `0` to disable timeout. Is #782 only match `goto` options?

i already test with `fit` with 100 sec, and it pass.

on my windows machine test 1 spec always fail and that is:
> Page Network Events Page.Events.Response should not report body unless request is finished

so i just rely on travis

documentation: 
should i also change `page.goBack(options)` , `page.goForward(options)`  ? let me know if i miss some

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -594,7 +594,7 @@ Navigate to the next page in history.
 #### page.goto(url, options)
 - `url` <[string]> URL to navigate page to. The url should include scheme, e.g. `https://`.
 - `options` <[Object]> Navigation parameters which might have the following properties:
-  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds.
+  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds, pass `0` to disable timeout.
   - `waitUntil` <[string]> When to consider navigation succeeded, defaults to `load`. Could be either:
     - `load` - consider navigation to be finished when the `load` event is fired.
     - `networkidle` - consider navigation to be finished when the network activity stays "idle" for at least `networkIdleTimeout` ms.
--- a/docs/api.md
+++ b/docs/api.md
@@ -567,7 +567,7 @@ If there's no element matching `selector`, the method throws an error.
 
 #### page.goBack(options)
 - `options` <[Object]> Navigation parameters which might have the following properties:
-  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds.
+  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds, pass `0` to disable timeout.
   - `waitUntil` <[string]> When to consider a navigation finished, defaults to `load`. Could be either:
     - `load` - consider navigation to be finished when the `load` event is fired.
     - `networkidle` - consider navigation to be finished when the network activity stays "idle" for at least `networkIdleTimeout` ms.
@@ -580,7 +580,7 @@ Navigate to the previous page in history.
 
 #### page.goForward(options)
 - `options` <[Object]> Navigation parameters which might have the following properties:
-  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds.
+  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds, pass `0` to disable timeout.
   - `waitUntil` <[string]> When to consider navigation succeeded, defaults to `load`. Could be either:
     - `load` - consider navigation to be finished when the `load` event is fired.
     - `networkidle` - consider navigation to be finished when the network activity stays "idle" for at least `networkIdleTimeout` ms.
@@ -705,7 +705,7 @@ Shortcut for [`keyboard.down`](#keyboarddownkey-options) and [`keyboard.up`](#ke
 
 #### page.reload(options)
 - `options` <[Object]> Navigation parameters which might have the following properties:
-  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds.
+  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds, pass `0` to disable timeout.
   - `waitUntil` <[string]> When to consider navigation succeeded, defaults to `load`. Could be either:
     - `load` - consider navigation to be finished when the `load` event is fired.
     - `networkidle` - consider navigation to be finished when the network activity stays "idle" for at least `networkIdleTimeout` ms.
@@ -900,7 +900,7 @@ Shortcut for [page.mainFrame().waitForFunction(pageFunction[, options[, ...args]
 
 #### page.waitForNavigation(options)
 - `options` <[Object]> Navigation parameters which might have the following properties:
-  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds.
+  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds, pass `0` to disable timeout.
   - `waitUntil` <[string]> When to consider navigation succeeded, defaults to `load`. Could be either:
     - `load` - consider navigation to be finished when the `load` event is fired.
     - `networkidle` - consider navigation to be finished when the network activity stays "idle" for at least `networkIdleTimeout` ms.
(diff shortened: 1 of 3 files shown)


lib/NavigatorWatcher.js
```diff
@@ -40,8 +40,10 @@ class NavigatorWatcher {
 
     this._eventListeners = [];
 
-    const watchdog = new Promise(fulfill => this._maximumTimer = setTimeout(fulfill, this._timeout))
-        .then(() => 'Navigation Timeout Exceeded: ' + this._timeout + 'ms exceeded');
+    const watchdog = new Promise(fulfill => {
```
aslushnikov: this is a little tricky: there should be no watchdog if there's no timeout timer.

I'd rather do:
```js
const navigationPromises = [];
if (this._timeout) {
  const watchdog = new Promise(fulfill => {
    // ...
  });
  navigationPromises.push(watchdog);
}
```

test/test.js
```diff
@@ -619,6 +619,16 @@ describe('Page', function() {
       expect(error.message).toContain('Navigation Timeout Exceeded: 1ms');
       process.removeListener('unhandledRejection', unhandledRejectionHandler);
     }));
+    it('should work when option.timeout set to 0', SX(async function() {
```
aslushnikov: this test will run for quite some time - which is unfortunate.
Developer's time is precious! :)

I'd rather make sure that `timeout: 0` doesn't set a zero timeout watchdog:

```js
it('should disable timeout when its set to 0', SX(async function() {
  let error = null;
  await page.goto(PREFIX + '/grid.html', {timeout: 0}).then(e => error = e);
  expect(error).toBe(null);
}));
```

docs/api.md
```diff
@@ -594,7 +594,7 @@ Navigate to the next page in history.
 #### page.goto(url, options)
 - `url` <[string]> URL to navigate page to. The url should include scheme, e.g. `https://`.
 - `options` <[Object]> Navigation parameters which might have the following properties:
-  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds.
+  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds, pass `0` to disable timeout.
```
aslushnikov, line 597: There's a bunch of other methods that rely on NavigationWatcher, so you should add this clarification there as well:

```js
page.goForward()
page.boBack();
page.reload();
page.waitForNavigation()
```

aslushnikov (commented):
yay, thanks for the pull request!

test/test.js
```diff
@@ -619,6 +619,14 @@ describe('Page', function() {
       expect(error.message).toContain('Navigation Timeout Exceeded: 1ms');
       process.removeListener('unhandledRejection', unhandledRejectionHandler);
     }));
+    it('should disable timeout when its set to 0', SX(async function() {
+      server.setRoute('/empty.html', (req, res) => {});
+      let error = null;
+      page.goto(PREFIX + '/empty.html', {timeout: 0}).catch(e => error = e);
+      // const sleep = e => new Promise(ff => setTimeout(ff, e));
+      // await sleep(90000);
```
sukrosono: > Developer's time is precious! :)

agree, sorry but leave it as comment. Perhaps someone want to inspect this.

lib/NavigatorWatcher.js
```diff
@@ -40,9 +40,12 @@ class NavigatorWatcher {
...
 
-    const watchdog = new Promise(fulfill => this._maximumTimer = setTimeout(fulfill, this._timeout))
-        .then(() => 'Navigation Timeout Exceeded: ' + this._timeout + 'ms exceeded');
-    const navigationPromises = [watchdog];
+    const navigationPromises = [];
+    if (this._timeout) {
+      const watchdog = new Promise(fulfill => this._maximumTimer = setTimeout(fulfill, this._timeout))
+          .then(() => 'Navigation Timeout Exceeded: ' + this._timeout + 'ms exceeded');
+      navigationPromises.push(watchdog);
+    }
```
sukrosono, line 48: this is as the @aslushnikov suggestion, it's simple and clear code. But it think it still have problem if user pass timeout 0 as a string. I'd like add more conditional check, or we just rely on the documentation?

sukrosono:
## Clarification:
the puppeteer launch [options](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions) is documented accept `0` to disable timeout.

my question:
-  is it same the page options and puppeteer options?
-  is it completely different thing? what's the relation?

as the puppeteer options which first documented as accept 0 to disable timeout. I can't see it's implementation, i hope just i who miss it.  Also when i look up `Puppeteer`, `Browser` and `Launcher` class, that class not importing `NavigationWatcher`. Sorry dude if i just to noisy :smile:

lib/NavigatorWatcher.js
aslushnikov: @brutalcrozt yes, we should rely on documentation in this case and expect user to pass in a number. There's also a check that timeout is always a number: https://github.com/GoogleChrome/puppeteer/blob/c46c41d89caefd1e1951055fb041521a2bcc933b/lib/NavigatorWatcher.js#L28

test/test.js
```diff
@@ -619,6 +619,14 @@ describe('Page', function() {
       expect(error.message).toContain('Navigation Timeout Exceeded: 1ms');
       process.removeListener('unhandledRejection', unhandledRejectionHandler);
     }));
+    it('should disable timeout when its set to 0', SX(async function() {
+      server.setRoute('/empty.html', (req, res) => {});
+      let error = null;
+      page.goto(PREFIX + '/empty.html', {timeout: 0}).catch(e => error = e);
```
aslushnikov: `await` is missing: 

```js
await page.goto(PREFIX + '/empty.html', {timeout: 0}).catch(e => error = e);
```

test/test.js
aslushnikov: Please, drop these comments: they are not useful for other puppeteer developers.

test/test.js
```diff
@@ -619,6 +619,14 @@ describe('Page', function() {
...
+      // const sleep = e => new Promise(ff => setTimeout(ff, e));
+      // await sleep(90000);
+      expect(error).toBe(null);
+    }));// 100000
```
aslushnikov: nit: please drop the comment here as well

aslushnikov (commented):
thanks for working on this! Left a few more comments before we can merge this.

aslushnikov:
> is it same the page options and puppeteer options?
> is it completely different thing? what's the relation?

These are different timeouts. One is a navigation timeout, another one is a browser launcher timeout.

test/test.js
sukrosono: ok sir

sukrosono:
https://github.com/GoogleChrome/puppeteer/blob/c46c41d89caefd1e1951055fb041521a2bcc933b/lib/Launcher.js#L177

Ok this what i looking for, i just miss it. Thanks for the review, thank for your cooperation :smile:

aslushnikov (approved):

PR #888, opened 2017-09-26
[WIP] Adding support for Experimental Page.setDownloadBehavior

### LINKED ISSUE #299: Question: How do I get puppeteer to download a file?
Question: How do I get puppeteer to download a file or make additional http requests and save the response?

ericduran:
- Work done on top https://github.com/GoogleChrome/puppeteer/pull/878
 - Fixes #299 
 - [ ] Needs test

I don't think the params are great right now, since it's just a plan passthrough of `(behavior, downloadPath)`.

Opening it up for feedback, I need this feature asap (which is why I'm using this hack). But I rather get feedback and put in a proper fix.

I'll spend sometime looking at the rest of puppeteer and see how other similar features are set up.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -63,6 +63,7 @@
     + [page.select(selector, ...values)](#pageselectselector-values)
     + [page.setContent(html)](#pagesetcontenthtml)
     + [page.setCookie(...cookies)](#pagesetcookiecookies)
+    + [page.setDownloadBehavior(behavior, downloadPath)](#pagesetdownloadbehaviorbehavior-downloadpath)
     + [page.setExtraHTTPHeaders(headers)](#pagesetextrahttpheadersheaders)
     + [page.setJavaScriptEnabled(enabled)](#pagesetjavascriptenabledenabled)
     + [page.setRequestInterceptionEnabled(value)](#pagesetrequestinterceptionenabledvalue)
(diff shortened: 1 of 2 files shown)


aslushnikov:
@ericduran Hi Eric, first of all - thank you for the PR.

We don't want to land the API in this form:
- it lacks the download notification, so it's unclear when the download is complete
- since the browser might be running somewhere in other universe (and connection could be established with the `puppeteer.connect`), it wouldn't be possible to access the downloaded file. There's an upstream work to plumb a byte stream of the download over the protocol, so we're waiting for it.

>  I need this feature asap (which is why I'm using this hack)

You can hack it locally for yourself: the `page._client` is accessible.

ericduran:
@aslushnikov makes perfect sense. This is meh at best ;-) closing this.

Quick question: Do you have a quick link to the chromium issue re: the byte stream over the protocol? Just so I can keep tabs on it.

ericduran:
@aslushnikov I want to follow it to see if it treats PDF the same. Since currently PDF are treated differently depending on the pdfViewer plugin (but that's a whole other thing)

aslushnikov:
@ericduran the upstream bug about byte stream for the pdf: https://bugs.chromium.org/p/chromium/issues/detail?id=748956

arliber:
Hi @ericduran - could you please elaborate on the workaround for setting the download path?

I tried `await page._client.send('Page.setDownloadBehavior', { behavior: "", downloadPath: "~/Repos/ad-network/scraper/downloads" })` but that didn't work :(

Thank you

paulirish:
@LeonLiber just a guess: i imagine the homedir expansion isn't legit. You'll have to pass it an absolute path.

ericduran:
☝️ That's probably it. I'm using absolute paths. Also if that doesn't work make sure you're running the latest version of Puppeteer and Chromium. I believe `npm install puppeteer` will update chromium. You need at least rev `503964` which is the latest as on this repo.

sohailalam2:
I have been searching for this solution, until @kensoh mentioned it to me. 👍 
Awesome its working for me now :)

@LeonLiber - @paulirish is probably correct. Also relative path works for me :)

```js
await page._client.send('Page.setDownloadBehavior', {behavior: 'allow', downloadPath: './'})
```

PR #906, opened 2017-09-28
Introduce appMode which does not override viewport and other browser p…

pavelfeldman:
…properties

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -155,7 +155,7 @@ class Launcher {
    * @return {!Promise<!Browser>}
    */
   static async connect(options = {}) {
-    const connection = await Connection.create(browserWSEndpoint);
+    const connection = await Connection.create(options.browserWSEndpoint);
     return new Browser(connection, options);
   }
 }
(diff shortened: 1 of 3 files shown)


lib/Launcher.js
```diff
@@ -144,12 +151,12 @@ class Launcher {
   }
 
   /**
-   * @param {string} options
+   * @param {!Object=} options
    * @return {!Promise<!Browser>}
    */
-  static async connect({browserWSEndpoint, ignoreHTTPSErrors = false}) {
+  static async connect(options = {}) {
     const connection = await Connection.create(browserWSEndpoint);
```
aslushnikov: to fix tests:
```js
const connection = await Connection.create(options.browserWSEndpoint);
```

lib/Launcher.js
```diff
@@ -144,12 +151,12 @@ class Launcher {
...
-    return new Browser(connection, !!ignoreHTTPSErrors);
+    return new Browser(connection, options);
```
aslushnikov: to fix tests:
```js
const connection = await Connection.create(options.browserWSEndpoint);
```

aslushnikov (approved):

PR #909, opened 2017-09-28
feat(Console): Introduce ConsoleMessage type

### LINKED ISSUE #744: Feature request: console level
Page offers a console event but it is not possible to know the level of the message (log, error...).

    page.on('console', function (...args) {
        // No way to know if this is an error
    })

Would it be possible to implement a way to know the level? I don't know what would be the cleanest way, either one more event per level:

    page.on('consoleError', ...)

or a new type of event:

    page.on('consoleWithType', function ({type, args}) ...

Basically here: https://github.com/GoogleChrome/puppeteer/blob/master/lib/Page.js#L308 the event type is totally ignored

aslushnikov:
This patch introduces ConsoleMessage type and starts dispatching
it for the `'console'` event.

BREAKING CHANGE: this breaks the api of the `'console'` event.

Fixes #744.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -99,6 +99,10 @@
     + [dialog.dismiss()](#dialogdismiss)
     + [dialog.message()](#dialogmessage)
     + [dialog.type](#dialogtype)
+  * [class: ConsoleMessage](#class-consolemessage)
+    + [consoleMessage.args](#consolemessageargs)
+    + [consoleMessage.text](#consolemessagetext)
+    + [consoleMessage.type](#consolemessagetype)
   * [class: Frame](#class-frame)
     + [frame.$(selector)](#frameselector)
     + [frame.$$(selector)](#frameselector)
@@ -249,7 +253,7 @@ puppeteer.launch().then(async browser => {
 ```
 
 #### event: 'console'
-- <[string]>
+- <[ConsoleMessage]>
 
 Emitted when JavaScript within the page calls one of console API methods, e.g. `console.log` or `console.dir`. Also emitted if the page throws an error or a warning.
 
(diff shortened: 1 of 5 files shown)


lib/Page.js
```diff
@@ -313,7 +313,10 @@ class Page extends EventEmitter {
       return;
     }
     const values = await Promise.all(event.args.map(arg => helper.serializeRemoteObject(this._client, arg)));
-    this.emit(Page.Events.Console, ...values);
+    const text = values.length && typeof values[0] === 'string' ? values[0] : null;
```
JoelEinbinder: Let's make text all the `values.join(" ")`

docs/api.md
```diff
@@ -1093,6 +1097,21 @@ puppeteer.launch().then(async browser => {
 
 Dialog's type, could be one of the `alert`, `beforeunload`, `confirm` and `prompt`.
 
+### class: ConsoleMessage
```
JoelEinbinder, line 1100: Inline this into the console event?

JoelEinbinder (approved):

docs/api.md
aslushnikov: I think it's fine like this

vsemozhetbyt:
It seems these fragments also need updating:

https://github.com/GoogleChrome/puppeteer/blame/0164b7cc4e60451665fafc2e8ccdb7855237ed29/docs/api.md#L517

https://github.com/GoogleChrome/puppeteer/blame/0164b7cc4e60451665fafc2e8ccdb7855237ed29/docs/api.md#L539

https://github.com/GoogleChrome/puppeteer/blame/ccf8fb168166ee12f42586e608d3557ff48edc27/README.md#L156

PR #910, opened 2017-09-29
fix MaxListenersExceededWarning

mizdra:
A EventListener is not removed when `Protocol error (Page.navigate)` is catched. This causes `MaxListenersExceededWarning`.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -743,16 +743,14 @@ describe('Page', function() {
       process.removeListener('warning', warningHandler);
       expect(warning).toBe(null);
     }));
-    it('should remove event listeners when page is closed', SX(async function() {
-      const closedPage = await browser.newPage();
-      await closedPage.close();
-
-      process.once('warning', warning => expect(warning).toBe(null));
-      for (let i = 0; i < 20; ++i) {
-        try {
-          await closedPage.goto(EMPTY_PAGE);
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -743,6 +743,17 @@ describe('Page', function() {
       process.removeListener('warning', warningHandler);
       expect(warning).toBe(null);
     }));
+    it('should remove event listeners when page is closed', SX(async function() {
+      const closedPage = await browser.newPage();
+      await closedPage.close();
+
+      process.once('warning', warning => expect(warning).toBe(null));
```
aslushnikov: this event listener should be removed after the test.

note: You can just copy the previous test and use 'asdf' instead of EMPTY_PAGE as url:

```js
    it('should not leak listeners during bad navigation', SX(async function() {
      let warning = null;
      const warningHandler = w => warning = w;
      process.on('warning', warningHandler);
      for (let i = 0; i < 20; ++i)
        await page.goto('asdf').catch(e => {/* swallow navigation error */});
      process.removeListener('warning', warningHandler);
      expect(warning).toBe(null);
    }));
```

aslushnikov (approved):
Awesome, thank you for the PR.

lgtm % test comment

test/test.js
mizdra: Oh, I fixed it.

Thank you 😃

PR #912, opened 2017-09-29
[feat] Allow environment variables definition when launching chromium

BenoitZugmeyer:
I'd like to launch Chromium with a specific timezone for my tests.  On Linux and macOS, I just need to define the `TZ` environment variable to the wanted timezone. So I added a new `puppeteer.launch()` option to allow defining the environment variables of the forked process.

I read the contributing rules, but I failed to write a good unit test for Windows, because Chromium seems to [ignore the TZ environment variable](https://bugs.chromium.org/p/chromium/issues/detail?id=125614) on this platform. If anyone have an idea for a more cross-platform unit test, I'd be happy to amend my PR. For now, the test is simply disabled on Windows.

FWIW, a similar issue is open at the [chrome-launcher project](https://github.com/GoogleChrome/chrome-launcher/issues/29).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -190,6 +190,7 @@ This methods attaches Puppeteer to an existing Chromium instance.
   - `timeout` <[number]> Maximum time in milliseconds to wait for the Chrome instance to start. Defaults to `30000` (30 seconds). Pass `0` to disable timeout.
   - `dumpio` <[boolean]> Whether to pipe browser process stdout and stderr into `process.stdout` and `process.stderr`. Defaults to `false`.
   - `userDataDir` <[string]> Path to a [User Data Directory](https://chromium.googlesource.com/chromium/src/+/master/docs/user_data_dir.md).
+  - `env` <[Object]> Specify environment variables that will be visible to Chromium. Defaults to `process.env`.
 - returns: <[Promise]<[Browser]>> Promise which resolves to browser instance.
 
 The method launches a browser instance with given arguments. The browser will be closed when the parent node.js process is closed.
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -82,7 +82,14 @@ class Launcher {
     if (Array.isArray(options.args))
       chromeArguments.push(...options.args);
 
-    const chromeProcess = childProcess.spawn(chromeExecutable, chromeArguments, {detached: true});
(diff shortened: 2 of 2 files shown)


BenoitZugmeyer:
I signed it!

aslushnikov:
Hi @BenoitZugmeyer, thank you for the pull request.

The code looks legit. However, I wonder if setting up `process.env` prior to launching browser would work for your use case?

```js
const savedEnv = Object.assign({}, process.env);

process.env.TZ = 'America/Noronha';

Object.assign(process.env, savedEnv); // Restore env
```

BenoitZugmeyer:
It is indeed possible to mutate the `process.env`, but I find it inconvenient for one major reason: In your code sample, since you are restoring the environment after waiting for `puppeteer.launch` to finish, the env is altered for any concurrent task going on. A better (but more inconvenent) solution would be:
```js
const savedEnv = Object.assign({}, process.env);

process.env.TZ = 'America/Noronha';
const browserPromise = puppeteer.launch();
Object.assign(process.env, savedEnv); // Restore env

const browser = await browserPromise;
```
This is possible because `puppeteer.launch` is calling `child_process.spawn` synchronously.

Also, in my opinion, documenting the possibility to define the chromium environment variables this way is clearer than letting the user guess how to do it by mutating `process.env`.

test/test.js
```diff
@@ -121,6 +121,19 @@ describe('Puppeteer', function() {
       expect(fs.readdirSync(userDataDir).length).toBeGreaterThan(0);
       rm(userDataDir);
     }));
+
+    // Disabled on Windows until we find a good use case on this platform. The
+    // TZ environment variable is currently ignored, see
+    // https://bugs.chromium.org/p/chromium/issues/detail?id=125614
+    (process.platform === 'win32' ? xit : it)('env option', SX(async function() {
```
aslushnikov: Lets not add this test - it adds more debt.
The patch is small and mostly does plumbing, so we can tolerate landing this without a test.

BenoitZugmeyer: Alright, test removed! (commit amended)

aslushnikov (approved):

chux0519:
Why setting the time zone to "Asia/Shanghai" does not work, and eventually I use 'utc-8' instead？

Issue #645, opened 2017-09-01
Browser-level cookies API


aslushnikov:
We need a way to fetch, add and remove cookies from browser:
```js
const cookies = await browser.cookies();
await browser.deleteCookie(cookies[0])
await browser.addCookie(cookies[0]);
```

When the #85 is implemented, the cookies will belong to the browser context. This will nicely explain cookie scoping: different pages from the same browser context share cookies, whereas pages from different browser contexts don't.

And once we have cookies on browser / browserContext, we can remove page's cookies api.

OlegWock:
Any updates about this feature? 

Will be very helpful

VaporFan:
bumppp

ai0:
beautiful feature, bump

tolmasky:
This is a key feature we’d like to have, is it only a matter of creating the node interface of does it require work in chrome proper as well? If the former, I’d be happy to help.

Thanks,

Francisco

aslushnikov:
> This is a key feature we’d like to have

@tolmasky Would you mind sharing your usecase?

> is it only a matter of creating the node interface of does it require work in chrome proper as well? 

This does require Chromium patch. Cookies are stored per browser context, so we need a way to query browser context for cookies without any page.

tolmasky:
Sure, we use Puppeteer for testing, and many actions require log in. If we go through the manual steps of logging in it makes the tests way slower. So instead, we just fetch(/login), grab the cookie from that, and implant it into the browser. Right now, we're just putting it into a random page (which I guess then propagates it to everything else in the context?). I suppose this is fine -- its just very confusing whether a page is logged in, or a context is logged in. If we instead were just doing `await login(context)` it would be a lot clearer what's going on.

devanteOne:
Found a work around. https://stackoverflow.com/questions/46631333/how-to-recreate-a-page-with-all-of-the-cookies

jeroenvisser101:
We use another workaround that might help others here: https://gist.github.com/jeroenvisser101/636030fe66ea929b63a33f5cb3a711ad

It uses the [`Network.setCookies`](https://chromedevtools.github.io/devtools-protocol/tot/Network#method-setCookies) and [`Network.getAllCookies`](https://chromedevtools.github.io/devtools-protocol/tot/Network#method-getAllCookies) CDP commands.

28141932:
这个的确需要，因为page层面的cookie管理不能处理第三方的cookie。必须要browser层面才可以处理。
[...]

PR #931, opened 2017-10-02
Get all cookies method

schelkun:
Page does not have `getAllCookies` method. Just added it.

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -195,6 +195,13 @@ class Page extends EventEmitter {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


schelkun:
I signed it!

aslushnikov:
@schelkun thanks for the pr! However, this is a dupe of https://github.com/GoogleChrome/puppeteer/pull/635

This api should be on a browser object, but we're lacking the necessary functionality in chromium.
Please read #635 for details and track #645 for progress.

PR #932, opened 2017-10-02
chore(doclint): move doclint tests expectations into test folders

aslushnikov:
This makes for much easier work with doclint tests.


aslushnikov:
TBR: @JoelEinbinder

PR #934, opened 2017-10-02
chore(doclint): add basic tests for documentation parsers

aslushnikov:
This patch adds basic tests to verify javascript and markdown
documentation parsers.


aslushnikov:
TBR: @JoelEinbinder

PR #935, opened 2017-10-02
chore(doclint): support classes inheritance

aslushnikov:
This patch:
- gives meaningful names to doclint tests
- supports classes inheritance in documentation linter. When class A
  extends class B, all methods of class B are added to documentation of
  class A.

This is a prerequisite for Object Handles: ElementHandle will be
extending ObjectHandle.

References #382


utils/doclint/check_public_api/JSBuilder.js
```diff
@@ -141,18 +146,46 @@ class JSOutline {
...
  */
 module.exports = async function(sources) {
   const classes = [];
   const errors = [];
+  const inheritance = new Map();
   for (const source of sources) {
     const outline = new JSOutline(source.text());
     classes.push(...outline.classes);
     errors.push(...outline.errors);
+    for (let entry of outline.inheritance)
```
JoelEinbinder: Why are we cloning the map?

utils/doclint/check_public_api/JSBuilder.js
```diff
@@ -141,18 +146,46 @@ class JSOutline {
...
 /**
+ * @param {!Array<!Documentation.Class>} classes
+ * @param {!Map<string, string>} inheritance
+ * @return {!Array<!Documentation.Class>}
+ */
+function recreateClassesWithInheritance(classes, inheritance) {
+  const classesByName = new Map(classes.map(cls => [cls.name, cls]));
+  return classes.map(cls => {
+    const membersMap = new Map();
+    for (let wp = cls; wp; wp = classesByName.get(inheritance.get(wp.name))) {
```
JoelEinbinder, line 157: What does `wp` stand for?

aslushnikov: work pointer

utils/doclint/check_public_api/JSBuilder.js
aslushnikov: that's because super classes are not necessarily defined in the same file

JoelEinbinder (approved):

Issue #933, opened 2017-10-02
Clipped screenshot above 10,000px fails


paulirish:
I'm experimenting with trying to get screenshots of the fullpage past the 16k pixel limit (crbug.com/770769)... however some basics dont' seem to work.

Below my second screenshot, full2.png, doesn't extend past the  16384 limit, but regardless, the screenshot returned is all white pixels. (The first screenshot succeeds just fine)

```js
'use strict';
const puppeteer = require('puppeteer');
const devices = require('puppeteer/DeviceDescriptors');

(async() => {
const browser = await puppeteer.launch();
const page = await browser.newPage();
await page.emulate(devices['iPhone 6']);
await page.goto('http://news.bbc.co.uk');

await page.screenshot({path: 'full.png', clip: {
	x: 0, y: 0,
	width: 375, height: 10000
}});
await page.screenshot({path: 'full2.png', clip: {
	x: 0, y: 10000,
	width: 375, height: 6000
}});

await browser.close();
})();

```

paulirish:
Nevermind, the DPR of 2 was throwing me off. This technique is possible.

With a DPR of 2, you'll want heights of 8192 maximum.
 
Here's a working approach:

```js
await page.screenshot({path: 'full.png', clip: {
	x: 0, y: 0,
	width: 375, height: 8000
}});
await page.screenshot({path: 'full2.png', clip: {
	x: 0, y: 8000,
	width: 375, height: 8000
}});
```

-----

**_EDIT:_** I've packaged this technique  into a simple example: https://github.com/GoogleChrome/puppeteer/pull/937

Issue #926, opened 2017-10-01
Erratic screenshots (rectangular area not 'painted'). Potential bug. Only happens with headless=true


geertbrit:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 0.11.0 (also on 0.10.x)
* Platform / OS version: Macbook Pro Retina early 2015 / OS X Yosemite 10.10.5
* URLs (if applicable): http://bbc.com

**What steps will reproduce the problem?**

Running the following snippet wil reproduce the problem. NOTE: the problem is erratic; sometimes it will exhibit the correct behavior. 

```
const puppeteer = require('puppeteer');

(async() => {

  const browser = await puppeteer.launch({ headless: true });

  const page = await browser.newPage();
  await page.goto("http://bbc.com");
  await page.evaluate(() => window.scrollBy(0, document.body.clientHeight));
  // await page.setViewport({ width: 1280, height: 1024 }); //uncommenting seems to expose the erratic behavior more often
  await page.screenshot({ fullPage: true, path: 'screenshot.png' });

  await browser.close();

})();


```
**What is the expected result?**

Doing a fullpage screenshot should result in a complete screenshot of the page without rectangular areas that are not 'painted'.

**What happens instead?**

There's an erratic issue with the produced screenshot, clearly visible at the bottom of the screenshot (the footer of bbc.com homepage). It appears as if chrome doesn't do a complete paint, which results in a rectangular area to be entirely white. 

Note: this only happens when using `{headless:true}`. Doing `{headless:false}` produces the correct result. 

I included 2 screenshots, please check the bottom of the screenshots for the clear difference. To be clear: I'm NOT concerned with images that not have been loaded. Also: using `{headless: true}` produces different results each time. In about 10% of cases I'm getting a correct result. The other 90% differ from a screenshot as shown (a rectangular area missing from the paint) to the footer not displayed at all. 

I have not seen this with any other page and I've checked quite a few. Looking at the bbc.com source, I can't see a clear reason why this specific one is failing.

**Screenshot showing the CORRECT behavior (`{headless:true: false}`)**

![bbc headless false correct](https://user-images.githubusercontent.com/393259/31056757-65d1b65a-a6d7-11e7-933c-b4f3f80d53fd.png)

**Screenshot showing the ERRATIC behavior (`{headless:true}`)**

[...]

Issue #703, opened 2017-09-06
Full page screenshot is not using viewport size for `vw` and `vh` CSS.


yuliyan:
Using `fullPage: true` option with [Page.screenshot](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagescreenshotoptions) seems to be messing `vw` and `vh` CSS units, even though page size is set with [Page.setViewport](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagesetviewportviewport).

aslushnikov:
@yuliyan could you please share with us a reproduction scenario, including script, page you try to screenshot and actual/expected outcomes?

This would save us a ton of time figuring what's going on.

yuliyan:
An example page that's using `vh` CSS units - http://jsbin.com/fawerivecu/edit?html,css,output.

**Script:**
```Javascript
const puppeteer = require('puppeteer');

(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  await page.goto('http://output.jsbin.com/fawerivecu');
  await page.setViewport({
  	width: 1920,
  	height: 1080
  });
  await page.screenshot({ path: 'example.png', fullPage: true });

  browser.close();
})();
```

**Expected Result:**
![expected_result](https://user-images.githubusercontent.com/2722412/30196887-9eb06188-946c-11e7-95d5-a46ba6f6e5ea.png)

**Actual Result:**
![actual_result](https://user-images.githubusercontent.com/2722412/30196890-a84bd6a0-946c-11e7-8021-967e415f1ac0.png)

aslushnikov:
Thanks, that's a bug with full page screenshots. Filed upstream: [crbug.com/763421](https://bugs.chromium.org/p/chromium/issues/detail?id=763421)

gtroshin:
Hey!
Any progress here? :) Maybe somebody knows the workaround?

tcjr:
The only workaround I have been able to get working is to change this line
https://github.com/GoogleChrome/puppeteer/blob/master/lib/Page.js#L713
from this:

```
const width = Math.ceil(metrics.contentSize.width);
```

to this:

```
const width = this._viewport.width;
```

The current code is resizing the height correctly because you want a full-height screenshot, but it's also resizing the screenshot's width to the size of the content, ignoring the viewport settings.  I think the width should maintain the desired viewport size.

schnerd:
[...]

Issue #359, opened 2017-08-17
Chrome seems to have a hard limit when taking screenshots of long pages


guilhermehn:
I'm using puppeteer (previously was using pure CDP) to take screenshots from all the endpoints of my application and some pages that contain legal info and EULA are really tall. Like really tall:

![image](https://user-images.githubusercontent.com/685012/29429858-7f967c66-8368-11e7-82e9-ae412cd48c0e.png)

And the resulting screenshots are only 16384px high. When debugging the headless instance with devTools the page seems to be fully rendered.

Here's the minimal snippet:

``` js
const browser = await puppeteer.launch();
const page = await browser.newPage();
await page.goto(url, {
	waitUntil: 'load'
});
await page.setViewport({
	width: size, // 320, 768, 1024, 1280
	height: 600
});
const screenshot = await page.screenshot({
	fullPage: true
});
browser.close();
fs.writeFileSync('./results/screenshot.png', screenshot);
```

Some of the tall pages screenshots result in a completely white screenshot. If I set the viewport using the `page.evaluate` method, it seems to work, but the height limit of 16384px still holds true.

``` js
const browser = await puppeteer.launch();
const page = await browser.newPage();
await page.goto(url, {
	waitUntil: 'load'
});
await page.setViewport({
	width: size, // 320, 768, 1024, 1280
	height: 600
});
await page.setViewport({
	width: size,
	height: await page.evaluate(() => document.body.clientHeight)
});
const screenshot = await page.screenshot();
browser.close();
fs.writeFileSync('./results/screenshot.png', screenshot);
```

But regardless of the technique used, the screenshots' height are all capped to 16384px.

paulirish:
Currently the maximum height is hardcoded in Chromium to that height. 
More information: https://groups.google.com/a/chromium.org/d/topic/headless-dev/DqaAEXyzvR0/discussion

timhaines:
@paulirish It would be very nice if Chrome handled the stitching together of tall pages for you.  :)

djones:
I have the same issue here. Here's another highly relevant thread https://bugs.chromium.org/p/chromium/issues/detail?id=638281

Two thoughts:

[...]

PR #937, opened 2017-10-02
Add example for screenshotting long pages (over 16,000px)

### LINKED ISSUE #929: Screenshots of long pages get cut
### Steps to reproduce

* Puppeteer version: 0.11.0
* Platform / OS version: macOS Sierra v. 10.12.6
* URLs (if applicable): http://www.spiegel.de/

```javascript
const browser = await Puppeteer.launch();
await page.goto('http://www.spiegel.de');
await page.setViewport({
  width: 1280,
  height: 1000,
  deviceScaleFactor: 1
});
const buffer = await page.screenshot({fullPage: true});
```

screenshot_01

**What happens instead?**

screenshot_00

**Note**

It looks like screenshots higher than 16374 pixels get cut.

paulirish:
Figured it'd be useful to have an example for the screenshot issues around.

Chromium issue: https://crbug.com/770769

This is a workaround that creates a few screenshots. It doesn't stitch them together though folks have had success using [sharp](https://github.com/lovell/sharp) for that.  

(Apologies for using a private API ;)

Fixes #929, #926, #933, #359 

----------

2020 update: https://github.com/morteza-fsh/puppeteer-full-page-screenshot handles this and also the [vw/vh bug](https://github.com/puppeteer/puppeteer/issues/703). seems pretty good.

### DIFF
--- a/examples/screenshots-longpage.js
+++ b/examples/screenshots-longpage.js
@@ -38,7 +38,7 @@ for (let ypos = 0; ypos < contentSize.height; ypos += maxScreenshotHeight) {
       x: 0,
       y: ypos,
       width: contentSize.width,
-      height: height
+      height
     }
   });
 }
--- a/examples/screenshots-longpage.js
+++ b/examples/screenshots-longpage.js
@@ -28,7 +28,7 @@ await page.goto('http://www.spiegel.de');
 
 const {contentSize} = await page._client.send('Page.getLayoutMetrics');
 const dpr = page.viewport().deviceScaleFactor || 1;
-const maxScreenshotHeight = 16000 / dpr; // Hardcoded max texture size of 16,384 (crbug.com/770769)
+const maxScreenshotHeight = Math.floor(16 * 1024 / dpr); // Hardcoded max texture size of 16,384 (crbug.com/770769)
 
 for (let ypos = 0; ypos < contentSize.height; ypos += maxScreenshotHeight) {
   const height = Math.min(contentSize.height - ypos, maxScreenshotHeight);
--- a/examples/screenshots-longpage.js
+++ b/examples/screenshots-longpage.js
@@ -27,7 +27,7 @@ await page.emulate(devices['iPhone 6']);
 await page.goto('http://www.spiegel.de');
 
 const {contentSize} = await page._client.send('Page.getLayoutMetrics');
-const dpr = page._viewport.deviceScaleFactor || 1;
+const dpr = page.viewport().deviceScaleFactor || 1;
 const maxScreenshotHeight = 16000 / dpr; // Hardcoded max texture size of 16,384 (crbug.com/770769)
 
 for (let ypos = 0; ypos < contentSize.height; ypos += maxScreenshotHeight) {
--- a/examples/screenshots-longpage.js
+++ b/examples/screenshots-longpage.js
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+
+const puppeteer = require('puppeteer');
+const devices = require('puppeteer/DeviceDescriptors');
+
+(async() => {
+const browser = await puppeteer.launch();
+const page = await browser.newPage();
+
+await page.emulate(devices['iPhone 6']);
+await page.goto('http://www.spiegel.de');
+
+const {contentSize} = await page._client.send('Page.getLayoutMetrics');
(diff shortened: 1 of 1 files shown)


robhope:
Thank you @paulirish - I haven't had a chance to test this yet but wanted to quickly share this Google Chrome plugin that I've been testing in past week (maybe 20+ times) that splits up long screenshots quite well:
https://mrcoles.com/full-page-screen-capture-chrome-extension/
https://github.com/mrcoles/full-page-screen-capture-chrome-extension
Maybe of some use as the code is Open Source too.

examples/screenshots-longpage.js
```diff
@@ -0,0 +1,47 @@
...
+const devices = require('puppeteer/DeviceDescriptors');
+
+(async() => {
+const browser = await puppeteer.launch();
+const page = await browser.newPage();
+
+await page.emulate(devices['iPhone 6']);
+await page.goto('http://www.spiegel.de');
+
+const {contentSize} = await page._client.send('Page.getLayoutMetrics');
```
ebidel, line 29: We might want to land https://github.com/GoogleChrome/puppeteer/issues/31 first, so we're not showing an example that uses private APIs.

examples/screenshots-longpage.js
```diff
@@ -0,0 +1,47 @@
...
+const dpr = page._viewport.deviceScaleFactor || 1;
```
ebidel: `const dpr = page.viewport().deviceScaleFactor || 1;`

examples/screenshots-longpage.js
```diff
@@ -0,0 +1,47 @@
...
+const maxScreenshotHeight = 16000 / dpr; // Hardcoded max texture size of 16,384 (crbug.com/770769)
```
ebidel: is 16000 to be conservative? Why not the full 16,384?

examples/screenshots-longpage.js
```diff
@@ -0,0 +1,47 @@
...
+
+for (let ypos = 0; ypos < contentSize.height; ypos += maxScreenshotHeight) {
+  const height = Math.min(contentSize.height - ypos, maxScreenshotHeight);
+  await page.screenshot({
+    path: `screenshot-@${ypos}px.png`,
+    clip: {
+      x: 0,
+      y: ypos,
+      width: contentSize.width,
+      height: height
```
ebidel: `height: height` -> `height`

examples/screenshots-longpage.js
paulirish: Perhaps.. but as the reporter of the oldest bug in this repo... I'm a little dubious that's happening soon. :)

aslushnikov:
@paulirish i think we should fix this in puppeteer / chromium. Its very surprising that full-page screenshot does not take a full page screenshot.

tomreid76:
Just an FYI...I tried this approach, as well as iterating over a grid of captures, on a project I am currently working on without much success. It was not so foreign to me as I have had to do the same thing with Phantom in order to keep memory consumption down on AWS Lambda.

In my use case I was taking screenshots of pages around 4000x4000. Each page contained between 6-15 full size original jpg images from digital cameras arranged on a page with a large png with transparent cutouts laid on top of the whole thing. I observed that the more images there were combined with the visual complexity of the overlay png, played a part in successful capture, or where the capture would fail silently with unrasterized blocks.

Even with striping or gridding, instead of having a half rendered image, I had a bunch of half rendered stripes or half rendered boxes respectively. Even when taking the stripes down to 10px and the grid down to 50px boxes, it was still problematic. Granted, my case is more extreme, but I just wanted to point out that the above approach still has limits when you get into things that are a bit more than capturing mobile web pages.

That said, puppeteer project looks super promising and was a delight to work with in comparison to alternatives, so needless to say I was very sad to have to abandon it (for the time being). Hoping to see this issue in Chromium resolved so I can pick it back up again in the future. Very exciting work.

aslushnikov:
Closing this since we not gonna land it.

joelgriffith:
Regarding @aslushnikov's comment: is there a bug we can track in Chromium?

examples/screenshots-longpage.js
paulirish: 2018 update: it happened. yay. `createCDPSession()`

hbakhtiyor:
@paulirish not work for https://bonobos.com/shop/tops, not fully capture screenshot, only parts

AutoSponge:
@tomreid76, @hbakhtiyor I'm late to this, but I think your problem is specific to message size limits. I noticed when my screenshots got too big, the message was truncated. 

I use a similar technique as above "stitching" multiple viewports together but I also crop inside ~80% of the view when it's not the first screen. This helps me avoid seeing a fixed header as "bands" multiple times in the final screenshot composite. This also makes sure my messages are small enough to get under the gate (which I think is ~5mb). If I'm right, the only other thing I can think of doing is to detect the image size before serialization and send it in chunks if it exceeds the threshold.

pirate:
Tangentially related, I'm posting here because I bet many people here would like to see this feature as well.

I've opened Chromium feature request to add a `--full-page` CLI flag to chrome headless: https://bugs.chromium.org/p/chromium/issues/detail?id=854013.  Please comment there as well if you'd like to see that feature added.

The intention is for `--full-page` to have identical full-page screenshot behavior to what's implemented in puppeteer for `fullPage: true`.

examples/screenshots-longpage.js
paulirish: https://github.com/morteza-fsh/puppeteer-full-page-screenshot handles this and also the [vw/vh bug](https://github.com/puppeteer/puppeteer/issues/703). seems pretty good.

PR #938, opened 2017-10-02
chore(doclint): exclude constructors by default

aslushnikov:
Our API does not expose any classes; thus all the constructors
should be excluded from the API.


JoelEinbinder (approved):
yay!

PR #939, opened 2017-10-02
feat(Page): Support Page.getMetrics and metrics event.

a1ph:
Provides access to the current page performance metrics.
Allows to push page metrics from the page JavaScript with `console.timeStamp()`

Fixes https://github.com/GoogleChrome/puppeteer/issues/309

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -315,8 +315,10 @@ class Page extends EventEmitter {
    */
   _buildMetricsObject(metrics) {
     const result = {};
-    for (const metric of metrics || [])
-      result[metric.name] = metric.value;
+    for (const metric of metrics || []) {
+      if (Page._supportedMetrics.has(metric.name))
+        result[metric.name] = metric.value;
+    }
     return result;
   }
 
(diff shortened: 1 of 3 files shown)


aslushnikov:
Thanks Alexei!

Let's list the metric names explicitly so that it's easier to consume.

test/test.js
```diff
@@ -587,6 +587,39 @@ describe('Page', function() {
...
+    }));
+    it('metrics event fired on console.timeStamp', SX(async function() {
+      const metricsPromise = new Promise(fulfill => page.once('metrics', fulfill));
+      await page.evaluate(() => console.timeStamp('test42'));
+      const metrics = await metricsPromise;
+      expect(metrics.title).toBe('test42');
+      checkMetrics(metrics.metrics);
+    }));
+    function checkMetrics(metrics) {
+      expect(metrics['Timestamp']).toBeGreaterThan(0);
```
aslushnikov: can we also check that there are no other fields beyond these? this would allow us to update docs when something new is added

a1ph: There are other fields coming over the protocol. But we don't want to document/expose those in puppeteer.

lib/Page.js
```diff
@@ -290,6 +292,35 @@ class Page extends EventEmitter {
...
+  }
+
+  /**
+   * @param {?Array<!{name: string, value: number}>} metrics
+   * @return {!Object}
+   */
+  _buildMetricsObject(metrics) {
+    const result = {};
+    for (const metric of metrics || [])
+      result[metric.name] = metric.value;
```
aslushnikov: since we don't want to expose certain numbers, can we white-list properties that we copy (instead of copying everything)?

a1ph: Done

aslushnikov (approved):

PR #942, opened 2017-10-02
Node6: Remove parentheses around the body of async arrow functions

JoelEinbinder:
Single line arrow functions can have parenthesis wrapping their body, but multi-line arrow functions cannot.


aslushnikov (approved):
lgtm

did you hit this somewhere?

JoelEinbinder:
In the browser.pages patch.

PR #944, opened 2017-10-03
[fix]: handleSIGINT should trigger process.exit()

Cap32:
If launch option `handleSIGINT` not `false`, `Ctrl+C` should close process.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -106,7 +106,7 @@ class Launcher {
 
     const listeners = [ helper.addEventListener(process, 'exit', killChrome) ];
     if (options.handleSIGINT !== false)
-      listeners.push(helper.addEventListener(process, 'SIGINT', killChrome));
+      listeners.push(helper.addEventListener(process, 'SIGINT', process.exit));
 
     try {
       const connectionDelay = options.slowMo || 0;


Cap32:
I signed it!

aslushnikov:
it looks like with this patch both `handleSIGINT: true` and `handleSIGINT: false` do the same thing.

Cap32:
No. Chrome process will not catch “exit” event on Ctrl+C if `handleSIGINT: false`.

aslushnikov:
> No. Chrome process will not catch “exit” event on Ctrl+C if handleSIGINT: false.

That's right. In this case the chrome process should be closed in the process.exit handler: [Launcher.js:114](https://github.com/GoogleChrome/puppeteer/blob/59bcc2ee567171d87c06a0f37f699d7827d13fa7/lib/Launcher.js#L114).

Cap32:
Sorry, I doesn't mean the chrome process, I mean the node process.

Without this PR, if `handleSIGINT !== false`, the chrome process will be closed on Ctrl+C, but since the node process added a `SIGINT` event listener, so the node process will no longer be closed on Ctrl+C, unless adding something like `process.on('SIGINT', () => process.exit())` manually.

IMHO, for the most cases, both node process and chrome process should be closed on Ctrl-C. For edge cases, developers should use `handleSIGINT: false`.

aslushnikov:
> Without this PR, if handleSIGINT !== false, the chrome process will be closed on Ctrl+C, but since the node process added a SIGINT event listener, so the node process will no longer be closed on Ctrl+C, unless adding something like process.on('SIGINT', () => process.exit()) manually.

@Cap32 Alternatively to adding the ` process.on('SIGINT', () => process.exit())`, you can launch pptr with `handleSIGINT: false`. Do you have any specific user scenario that prevents you from doing this? Could you please describe it as an issue so that we can continue discussion over there?

I'll close this PR since the suggested change makes `handleSIGINT: true` behave the same as `handleSIGINT: false`.

PR #953, opened 2017-10-05
feat(Launcher): add devtools option

vsemozhetbyt:
Fixes: https://github.com/GoogleChrome/puppeteer/issues/864 (see https://github.com/GoogleChrome/puppeteer/issues/864#issuecomment-334318996)
Refs: https://peter.sh/experiments/chromium-command-line-switches/#auto-open-devtools-for-tabs

Unfortunately, I do not know if this option needs a test and how it should be tested if needed. Please, let me know what code should I add to the tests,

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -72,6 +72,10 @@ class Launcher {
 
       chromeArguments.push(`--user-data-dir=${options.userDataDir || temporaryUserDataDir}`);
     }
+    if (options.devtools === true) {
+      chromeArguments.push('--auto-open-devtools-for-tabs');
+      options.headless = false;
+    }
     if (typeof options.headless !== 'boolean' || options.headless) {
       chromeArguments.push(
           '--headless',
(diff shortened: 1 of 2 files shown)


paulirish:
Personally I think this is better solved with manually adding that flag like:

```js
puppeteer.launch({args: ['--auto-open-devtools-for-tabs']})
```

aslushnikov:
@paulirish The bare-bone flag is a minimalistic approach yet not very convenient.

`devtools` option is nice because:
- it's hard to memorize the chrome flag
- the `devtools` option costs nothing to maintain
- it's a feature parity with nightmare

JoelEinbinder (approved):
I like it.

lib/Launcher.js
```diff
@@ -80,6 +80,8 @@ class Launcher {
           '--mute-audio'
       );
     }
+    if (options.headless === false && options.devtools === true)
+      chromeArguments.push('--auto-open-devtools-for-tabs');
```
paulirish: @vsemozhetbyt we discussed and folks are comfortable with this. (including me. :)

 can you also set `options.headless = false` automatically if `devtools` is true?

vsemozhetbyt: I've amended. Is it OK now?

PR #958, opened 2017-10-05
[feat] add environment variable to overwrite host part of url

### LINKED ISSUE #951: Add possibility to configure download mirror
Currently the URL "https://storage.googleapis.com/chromium-browser-snapshots/..." is hardcoded.
Since we would like to use this package behind a firewall which is blocking downloads, we need to be able to configure the download mirror.

An option like the ENV variable in the phantomjs wrapper package: "PHANTOMJS_CDNURL" would be useful for us.

r3dDoX:
fixes #951

### DIFF
--- a/utils/ChromiumDownloader.js
+++ b/utils/ChromiumDownloader.js
@@ -26,12 +26,13 @@ const ProxyAgent = require('https-proxy-agent');
 const getProxyForUrl = require('proxy-from-env').getProxyForUrl;
 
 const DOWNLOADS_FOLDER = path.join(__dirname, '..', '.local-chromium');
+const DEFAULT_DOWNLOAD_HOST = 'https://storage.googleapis.com';
 
 const downloadURLs = {
-  linux: 'https://storage.googleapis.com/chromium-browser-snapshots/Linux_x64/%d/chrome-linux.zip',
-  mac: 'https://storage.googleapis.com/chromium-browser-snapshots/Mac/%d/chrome-mac.zip',
-  win32: 'https://storage.googleapis.com/chromium-browser-snapshots/Win/%d/chrome-win32.zip',
-  win64: 'https://storage.googleapis.com/chromium-browser-snapshots/Win_x64/%d/chrome-win32.zip',
+  linux: '%s/chromium-browser-snapshots/Linux_x64/%d/chrome-linux.zip',
+  mac: '%s/chromium-browser-snapshots/Mac/%d/chrome-mac.zip',
+  win32: '%s/chromium-browser-snapshots/Win/%d/chrome-win32.zip',
+  win64: '%s/chromium-browser-snapshots/Win_x64/%d/chrome-win32.zip',
 };
 
 module.exports = {
@@ -64,7 +65,7 @@ module.exports = {
   canDownloadRevision: function(platform, revision) {
     console.assert(downloadURLs[platform], 'Unknown platform: ' + platform);
 
-    const options = requestOptions(util.format(downloadURLs[platform], revision), 'HEAD');
+    const options = requestOptions(util.format(downloadURLs[platform], DEFAULT_DOWNLOAD_HOST, revision), 'HEAD');
 
     let resolve;
     const promise = new Promise(x => resolve = x);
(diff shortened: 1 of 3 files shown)


r3dDoX:
I signed it!

utils/ChromiumDownloader.js
```diff
@@ -27,11 +27,15 @@ const getProxyForUrl = require('proxy-from-env').getProxyForUrl;
 
 const DOWNLOADS_FOLDER = path.join(__dirname, '..', '.local-chromium');
 
+const DOWNLOAD_HOST = process.env.PUPPETEER_DOWNLOAD_HOST
```
aslushnikov: It's not reliable to access env in ChromiumDownloader: it could be run in multiple of different processes and scripts:
- utils/check_availability.js
- install.js
- lib/Browser.js

For example, reading .npm_config_puppeteer_download_host doesn't make sense anywhere but in install.js.

Instead, host should be passed in to the `canDownloadRevision`/`downloadRevision` methods.

r3dDoX: So when I pass the host into these two methods I'd make the checks for them on the calling side. I don't see the spawning of new processes so I would have the exact same environment variables as the ChromiumDownloader has. Isn't it more maintainable to have the check in the central script? Or am I missing that it could be called from different spawned processes?

Since I'm defaulting the values, there is no harm in checking the config var for other scripts too. If it's not set, it would just be undefined and fall back to the hardcoded hostname. Or is there a potential error I don' see?

aslushnikov: > I don't see the spawning of new processes so I would have the exact same environment variables as the ChromiumDownloader has. Isn't it more maintainable to have the check in the central script? 

Reading env variables is required only in `install.js`. Maintenance-wise, I don't see much difference. 

>  Or am I missing that it could be called from different spawned processes?

ChromiumDownloader is `require`'d from different scripts, each of which could be run under different environments.

> Since I'm defaulting the values, there is no harm in checking the config var for other scripts too. If it's not set, it would just be undefined and fall back to the hardcoded hostname. Or is there a potential error I don' see?

No error here. I don't like reading env in ChromiumDownloader.js since it violates a [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege).

r3dDoX: I do see your points. I was too focused that this var would have to be passed in by many callees that I overlooked it only being needed by install.js. I will extract this ENV var handling 👍

r3dDoX:
PR is updated.

I'm not entirely happy with the concatenation of host and path. What do you think?

utils/ChromiumDownloader.js
```diff
@@ -149,13 +152,22 @@ module.exports = {
       executablePath,
       folderPath,
       downloaded: fs.existsSync(folderPath),
-      url: util.format(downloadURLs[platform], revision)
+      url: util.format(getDownloadUrl(platform), revision)
```
aslushnikov: you can remove the `url` field: it's never used (and will be misleading with this change)

utils/check_availability.js
```diff
@@ -106,8 +106,9 @@ async function checkRangeAvailability(fromRevision, toRevision) {
  */
 async function checkAndDrawRevisionAvailability(table, name, revision) {
   const promises = [];
+  const DOWNLOAD_HOST = process.env.PUPPETEER_DOWNLOAD_HOST || process.env.npm_config_puppeteer_download_host;
```
aslushnikov: I wouldn't do anything with this script: it's used solely for the purpose of rolling chromium dependency to the appropriate revision.

utils/ChromiumDownloader.js
```diff
@@ -82,11 +84,12 @@ module.exports = {
...
    * @param {string} platform
    * @param {string} revision
+   * @param {string} [downloadHost]
    * @param {?function(number, number)} progressCallback
    * @return {!Promise}
    */
-  downloadRevision: function(platform, revision, progressCallback) {
-    let url = downloadURLs[platform];
+  downloadRevision: function(platform, revision, downloadHost, progressCallback) {
+    let url = getDownloadUrl(platform, downloadHost);
```
aslushnikov: We already do formatting of download urls with `util.format`. I'd prefer this to be done in the same way, e.g.:

```js
const downloadURLs = {
  // Notice the %s in the very beginning
  linux: '%s/chromium-browser-snapshots/Linux_x64/%d/chrome-linux.zip',
  // ...
}
```

And later, every `util.format` will pass in HOST:

```js
url = util.format(url, host, revision);
```

r3dDoX: How did I miss that, yeah it's clear how to do that.

utils/ChromiumDownloader.js
r3dDoX: ok 👍

utils/check_availability.js
r3dDoX: okay, I thought for the sake of completeness. Will remove it.

r3dDoX:
Ok, next try :)

aslushnikov:
Code's good! Could you please update docs/api.md with a new env variable, and it's good to go.

r3dDoX:
Ok, done :+1:

aslushnikov (approved):
Thanks!

PR #961, opened 2017-10-05
use deprecated initEvent on change and input events

ChristianDavis:
Without the deprecated initEvent method, my use case fails to recognize the 'change' or 'input' event for select elements.

Adding it doesn't cause the tests to fail and it causes my use case to pass. 

I swapped some nightmare scripts over to puppeteer, and on healthcare.gov it wouldn't register input or change on a select. The nightmare implementation uses the deprecated syntax, but I just added a call to initEvent on the new event syntax and that solved it. 

I tried coming up with a replica scenario to add to a test, but I couldn't replicate outside the scraping target's website. I'm guessing it's some manual checking done in older JS libraries that might be missing something. I'll revisit adding a new test. 

adding @alixaxel because he implemented the select function

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -722,14 +722,8 @@ class Page extends EventEmitter {
       } else {
         element.value = values.shift();
       }
-      // init events are deprecated but still have an effect
-      const inputEvent = new Event('input');
-      inputEvent.initEvent('input', true, true);
-      element.dispatchEvent(inputEvent);
-
-      const changeEvent = new Event('change');
-      changeEvent.initEvent('change', true, true);
-      element.dispatchEvent(changeEvent);
+      element.dispatchEvent(new Event('input', {'bubbles': true, 'cancelable': true}));
+      element.dispatchEvent(new Event('change', {'bubbles': true, 'cancelable': true}));
     }, values);
   }
 
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -722,8 +722,8 @@ class Page extends EventEmitter {
       } else {
         element.value = values.shift();
       }
-      element.dispatchEvent(new Event('input', {'bubbles': true, 'cancelable': true}));
-      element.dispatchEvent(new Event('change', {'bubbles': true, 'cancelable': true}));
+      element.dispatchEvent(new Event('input', { 'bubbles': true }));
+      element.dispatchEvent(new Event('change', { 'bubbles': true }));
     }, values);
   }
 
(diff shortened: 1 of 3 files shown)


ChristianDavis:
I signed it!

aslushnikov:
@ChristianDavis that's interesting. Could you please add a test case? It would help to understand why this is necessary.

lib/Page.js
```diff
@@ -722,9 +722,14 @@ class Page extends EventEmitter {
       } else {
         element.value = values.shift();
       }
-
-      element.dispatchEvent(new Event('change'));
```
paulirish, line 726: does this work instead of using `initEvent`?

```js
new Event("change", {"bubbles":true, "cancelable":true});
```

(they default to false). It does probably make sense for them to be true here. Good call.

ChristianDavis: man, it sure did. I had the option syntax wrong.

ChristianDavis:
removed changes and just added {"bubbles":true, "cancelable":true} per @paulirish suggestion

lib/Page.js
alixaxel: Nice catch @ChristianDavis!

ChristianDavis:
I signed it!

ChristianDavis:
I think I screwed up the history here, any way to get the CLA to pass?

gmlewis:
I apologize for the CLA issue.  Resolved.

lib/Page.js
```diff
@@ -722,9 +722,8 @@ class Page extends EventEmitter {
...
-      element.dispatchEvent(new Event('input'));
+      element.dispatchEvent(new Event('input', {'bubbles': true, 'cancelable': true}));
+      element.dispatchEvent(new Event('change', {'bubbles': true, 'cancelable': true}));
```
aslushnikov: both 'input' and 'change' events are not cancelable:
- https://developer.mozilla.org/en-US/docs/Web/Events/input
- https://developer.mozilla.org/en-US/docs/Web/Events/change

aslushnikov:
Please, add a test to the `Page.select` testsuite. It should be easy to do similarly to one of the [existing tests](https://github.com/GoogleChrome/puppeteer/blob/master/test/test.js#L2082-L2087).

There's also a [test/assets/input/select.html](https://github.com/GoogleChrome/puppeteer/blob/master/test/assets/input/select.html), you can add add a listener to the body and pull it from the test.

ChristianDavis:
Cool, thanks for the suggestion. Dropped the 'cancelable' option because it was ignored

aslushnikov (approved):
Thanks!

Issue #549, opened 2017-08-25
Add inverse of waitForSelector visible: true (waitForNotSelector?)


quicksnap:
It would be very useful to test when a selector is no longer visible. For instance, I'm waiting for a certain container not to have a `loading` class and then assert things.

Would you accept a PR for this? It wouldn't be tough to implement. If so, what should the API look like?

As it stands now, it's necessary to do some dance with `waitForFunction`.

aslushnikov:
> As it stands now, it's necessary to do some dance with waitForFunction.

If I understand correctly, this behavior is easily implementable with `waitForFunction`:

```js
// wait for ".box" to disappear
await page.waitForFunction(() => !document.querySelector('.box'), {polling: 'mutation'});
```

quicksnap:
@aslushnikov Yeah, I did something similar. I didn't even need to specify the polling option and it worked just fine.

Still, seems like it would be a very useful utility. If it's worth adding to the API, just needs bikeshedding.

aslushnikov:
Closing this for now. We'll add this to the API if there are many clients who demand this and who are not satisfied with the one-liner for some reason.

aslushnikov:
Closed with PR #967

PR #967, opened 2017-10-06
feat(page): waitForSelector hidden option

ChristianDavis:
Waits for a selector to be present and hidden, opposite behavior of the option visible.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -401,22 +401,26 @@ class Frame {
   waitForSelector(selector, options = {}) {
     const timeout = options.timeout || 30000;
     const waitForVisible = !!options.visible;
-    const polling = waitForVisible ? 'raf' : 'mutation';
-    return this.waitForFunction(predicate, {timeout, polling}, selector, waitForVisible);
+    const waitForHidden = !!options.hidden;
+    const polling = waitForVisible || waitForHidden ? 'raf' : 'mutation';
+    return this.waitForFunction(predicate, {timeout, polling}, selector, waitForVisible, waitForHidden);
 
     /**
      * @param {string} selector
      * @param {boolean} waitForVisible
+     * @param {boolean} waitForHidden
      * @return {boolean}
      */
-    function predicate(selector, waitForVisible) {
+    function predicate(selector, waitForVisible, waitForHidden) {
       const node = document.querySelector(selector);
       if (!node)
         return false;
-      if (!waitForVisible)
+      if (!waitForVisible && !waitForHidden)
         return true;
       const style = window.getComputedStyle(node);
-      return style && style.display !== 'none' && style.visibility !== 'hidden';
+      if (waitForVisible)
+        return style && style.display !== 'none' && style.visibility !== 'hidden';
+      return style && (style.display === 'none' || style.visibility === 'hidden');
     }
   }
 
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -417,17 +417,10 @@ class Frame {
         return waitForHidden;
       if (!waitForVisible && !waitForHidden)
         return true;
-      let ancestorsVisible = true;
-      let parent = node.parentElement;
-      while (parent && ancestorsVisible){
-        const parentStyle = window.getComputedStyle(parent);
-        ancestorsVisible = parentStyle && parentStyle.display !== 'none';
-        parent = parent.parentElement;
-      }
       const style = window.getComputedStyle(node);
       if (waitForVisible)
-        return style && style.display !== 'none' && style.visibility !== 'hidden' && ancestorsVisible;
-      return style && (style.display === 'none' || style.visibility === 'hidden' || !ancestorsVisible);
+        return style && style.display !== 'none' && style.visibility !== 'hidden';
+      return style && (style.display === 'none' || style.visibility === 'hidden');
     }
   }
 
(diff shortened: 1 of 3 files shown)


JoelEinbinder:
I'm curious as to what the use case for this is.

ChristianDavis:
I'm using it for a screen where a modal is covering a button, click doesn't work till the modal is hidden

ebidel:
It's really hard to determine proper visibility of an element. For example, these checks fail if an ancestor node has `display: none`. The node will still report itself as "visible". 

I almost wish `waitForSelector` didn't have a `visible` option to begin with. There's too many gotchas around visibility of dom nodes.

ChristianDavis:
@ebidel I agree, it's tricky. And it doesn't take opacity into account, but I haven't seen any implementations that consider opacity. But, I disagree that it should be left out. Developer time being the most valuable asset, visibility is an easy way to time events on a page that might be complicated otherwise. 

I added reverse and hidden options around the visible feature, both options visible and hidden climb the DOM to check for a hidden ancestor. #549

trevordmiller:
I'm using `await page.waitFor(() => !document.querySelector("{selector}"));` for this now, but would appreciate some sort of reverse option on `waitForSelector`.

aslushnikov:
Hey @ChristianDavis, thanks for the PR.

I like the `hidden` option:
```js
await page.waitForSelector('.dialog', {hidden: true});
```

However, I'd love it to be a full opposite of `visible` option, i.e. to wait for element to be either absent from DOM or having `display: none` or `visibility: hidden`.

I don't like the `reverse` option. It's confusing if there's both `visible: true` and `reverse: true`: is it a `hidden: true` now?

I'd like to scope this PR to only add a `hidden` option. Let me know how you feel about this.

ChristianDavis:
@aslushnikov yeah, looks confusing. Someone had pointed out that webdriverio has an implementation { reverse: true }, but I should just combine the functionality into hidden

ChristianDavis:
{hidden: true} checks for NOT present, or hidden, or ancestors hidden

aslushnikov:
@ChristianDavis This PR now seems to be doing two things:
1. Adding 'hidden' option which is opposite to the 'visible'
2. Changing 'visible' and 'hidden' options to traverse the hierarchy to check for visibility.

Could you please split the PR into two separate ones? This will make for a nicer history and smaller diffs. Also, I feel like we've agreed on 'hidden' option, but there might be some questions on the hierarchy traversal.

ChristianDavis:
@ebidel didn't think hidden OR visible makes sense without the ancestor check, and that's probably true. But I could split it up

ebidel:
My point was really that checking "visibility" is hard to get right and we should consider removing the option.  But may people like the convenience and it's pretty nice for the simpler cases.

I'm a fan of making the existing checks more reliable. So traversing the DOM to check ancestors makes sense to me. To save some work, you could also cut out early in the case of `element.offsetParent === null`. That'd let you know if the element (or any ancestor) is `display: none`.

Splitting these up into separate PRs sgtm.

ChristianDavis:
feature is now just the hidden option, waits for not present or hidden

lib/FrameManager.js
```diff
@@ -401,22 +401,26 @@ class Frame {
...
-        return false;
-      if (!waitForVisible)
+        return waitForHidden;
+      if (!waitForVisible && !waitForHidden)
         return true;
       const style = window.getComputedStyle(node);
-      return style && style.display !== 'none' && style.visibility !== 'hidden';
+      if (waitForVisible)
+        return style && style.display !== 'none' && style.visibility !== 'hidden';
+      return style && (style.display === 'none' || style.visibility === 'hidden');
```
aslushnikov: Technically the `!style` makes more sense here.

How about:
```js
const style = window.getComputedStyle(node);
const isVisible = style && style.display !== 'none' && style.visibility !== 'hidden';
return waitForVisible === isVisible;
```

test/test.js
```diff
@@ -498,12 +498,40 @@ describe('Page', function() {
...
       await page.evaluate(() => document.querySelector('div').style.removeProperty('display'));
       expect(divFound).toBe(false);
       await page.evaluate(() => document.querySelector('div').style.removeProperty('visibility'));
       expect(await waitForSelector).toBe(true);
+      expect(divFound).toBe(true);
+    }));
+    it('hidden should wait for visibility: hidden', SX(async function() {
+      let divHidden = false;
+      await page.setContent(`<div style='display: block;'></div>`);
+      expect(divHidden).toBe(false);
```
aslushnikov: nit - here and in other tests: please drop this check, it's a noop

ChristianDavis: the line on 512?

lib/FrameManager.js
ChristianDavis: looks better
``` javascript
return ( waitForVisible === isVisible || waitForHidden === !isVisible);
```

aslushnikov (approved):

aslushnikov:
@ChristianDavis thanks!

trevordmiller:
@aslushnikov When will this be released to npm?

vsemozhetbyt:
Is this valid for `frame.waitForSelector(selector[, options])`? If so, the doc addition seems missing.

ChristianDavis:
@vsemozhetbyt You're right, I didn't notice that. Page.waitfor is actually just abstracting the frame implementation. Maybe the options for the frame.waitFor could be linked, or vice versa?

vsemozhetbyt:
I am not sure, but it seems all the other descriptions are copied when similar.

aslushnikov:
@ChristianDavis we copy descriptions: it's less clicks for users to navigate documentation.

PR #969, opened 2017-10-06
help me to $XPath()  

tamtc84:
I saw puppeteer_xpath v0.1.0. 
If i use function $XPath is:
`ret = document.evaluate(selector, document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue.textContent;`
=> this will return one text result => it ok.
and i want to get multi result then i change function $XPath to:
`ret = document.evaluate(selector, document, null, XPathResult.UNORDERED_NODE_SNAPSHOT_TYPE, null);`
=> this will return {} 
=> remoteObject.snapshotLength always emty

Please help me to get multi result.
Thank you very much.
----------------------
I test in console of firefox:
`var selector = "//div[@class='classname']";
var result = document.evaluate(selector, document, null, XPathResult.UNORDERED_NODE_SNAPSHOT_TYPE, null);
for (let i=0, length=result.snapshotLength; i<length; ++i) {
    var tmp = result.snapshotItem(i);
    console.log(tmp.innerText);
  }`
=> it ok.

### DIFF
--- a/examples/xpath.js
+++ b/examples/xpath.js
@@ -5,12 +5,21 @@ const puppeteer = require('../../puppeteer');
   const page = await browser.newPage();
   await page.goto('https://www.realestate.com.au/buy', {waitUntil: 'networkidle'});
 
-  const xpath = "/html/body/div[1]/header/div[2]/a/img/@src"
-  console.log(await page.waitForXpath(xpath))
+  var xpath = "/html/body/div[3]/main/section/div/div/div/article[2]/div/p[2]/@class"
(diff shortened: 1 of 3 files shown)


aslushnikov:
I don't think I can speak for puppeteer-xpath project that you use.
However, please check out the https://github.com/GoogleChrome/puppeteer/issues/537#issuecomment-334918553 for an example of using `xpath` function with Puppeteer 0.12.0-alpha. It might come handy for your use case.

PR #943, opened 2017-10-03
feat(ObjectHandles): Implement Object Handles

aslushnikov:
This patch:
- introduces ExecutionContext class that incapsulates javascript
  execution context. Examples of execution contexts are workers and
  frames
- introduces ObjectHandle that holds a references to the javascript
  object in ExecutionContext
- inherits ElementHandle from ObjectHandle

Fixes #382.

Issue #618, opened 2017-08-30
Clicking on an element in an iframe fails


krlwlfrt:
`page.click(selector)` does not click an element in a frame of the page.

```
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): AssertionError [ERR_ASSERTION]: No node found for selector: SELECTOR
```

How is it possible to click the element? Getting the frame that contains the element and doing `frame.click(selector)` does not work.

```
UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: frame.click is not a function
```

paulirish:
I looked into this and it turns out there's a bug you didn't yet run into that would have stopped you dead in your tracks. :) I'll add the details to this ticket since it basically breaks clicking correctly inside a frame.

repro script:

```js
const puppeteer = require('puppeteer');
(async() => {
const browser = await puppeteer.launch({headless: false, slowMo: 250});
const page = await browser.newPage();
await page.goto('https://www.w3schools.com/TAgs/tryit.asp?filename=tryhtml_button_test');

const frames = await page.frames();
const tryItFrame = frames.find(f => f.name() === 'iframeResult');
const framedButton = await tryItFrame.$('button');

const outerHTML = await framedButton.evaluate(e => e.parentNode.outerHTML);
console.log('the outerhtml: ', outerHTML); // to verify we're talking about the right element.. (we are.)

framedButton.click();
})();
```

Basically on this page we want to click the "Click me!" button. However a click is dispatched on top of the banner ad at the top of the page causing a new tab to open.

![image](https://user-images.githubusercontent.com/39191/29886462-00bf1108-8d6f-11e7-9a00-6afc2147ce1f.png)


It appears that when `elhandle.click()` is called, pptr gets the boundingclientrect and then dispatches an input event to the coordinates.

**However this doesnt account for an iframe that's not at 0,0 of the page's viewport.** Puppeteer needs to handle the viewport position of the iframe.

Here's the debug log showing this back and forth:

![image](https://user-images.githubusercontent.com/39191/29886418-d566d540-8d6e-11e7-9cb3-dec39265c85c.png)

(FWIW i gathered this log with `env DEBUG="*" env DEBUG_COLORS=true node examples/click-button-in-frame.js 2>&1 | grep -v Network`)

aslushnikov:
Ah, indeed! #445 will solve this

SMJ93:
[...]

PR #971, opened 2017-10-06
fix(input): clicking an element should take into account frame position

### LINKED ISSUE #684: Clicking element inside iFrame does not give any response
I am clicking an element inside an iframe which should give me a different frame/view, that is move into the view represented by that element. If i click the element manually,the url does not change. So, I guess only the frame is changed. The element is being clicked properly since there are no errors for that. However nothing happens when it is clicked and returns error when I do `page.waitForNavigation()`.
Here is the debug log listening for mouse and keyboard clicks.
puppteer
Also made sure proper element is clicked referring to #618. Here is portion of code that handles clicking the element inside iframe : 
```javascript
var frames = await page.frames();
var myframe = frames.find(f => f.name() === 'mygurukuliframe_submenu'); 
const attselector = '#imgid_12';
const attel = await myframe.$(attselector);
const outerHTML = await attel.evaluate(e => e.parentNode.outerHTML);
console.log('the outerhtml: ', outerHTML);
await attel.click({button:'middle'});
await page.waitForNavigation();
await page.screenshot({path : 'example.png'});
browser.close();
```

Please mention if any further info is needed and help if it can be resolved.

JoelEinbinder:
fixes #684

### DIFF
--- a/lib/ElementHandle.js
+++ b/lib/ElementHandle.js
@@ -44,22 +44,22 @@ class ElementHandle extends JSHandle {
    * @return {!Promise<{x: number, y: number}>}
    */
   async _visibleCenter() {
-    const {center, error} = await this.executionContext().evaluate(element => {
+    const error = await this.executionContext().evaluate(element => {
       if (!element.ownerDocument.contains(element))
-        return {center: null, error: 'Node is detached from document'};
+        return 'Node is detached from document';
       if (element.nodeType !== HTMLElement.ELEMENT_NODE)
-        return {center: null, error: 'Node is not of type HTMLElement'};
+        return 'Node is not of type HTMLElement';
       element.scrollIntoViewIfNeeded();
-      const rect = element.getBoundingClientRect();
-      const center = {
-        x: (Math.max(rect.left, 0) + Math.min(rect.right, window.innerWidth)) / 2,
-        y: (Math.max(rect.top, 0) + Math.min(rect.bottom, window.innerHeight)) / 2
-      };
-      return {center, error: null};
     }, this);
     if (error)
       throw new Error(error);
-    return center;
+    const {model} = await this._client.send('DOM.getBoxModel', {
+      objectId: this._remoteObject.objectId
+    });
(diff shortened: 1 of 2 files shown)


aslushnikov:
The `.getBoxModel` returns something different for me.  It probably should be fixed/enhanced first.

JoelEinbinder:
It returns different numbers when scale is involved, which we currently don't handle in Puppeteer.

veb:
Sorry if I'm not supposed to comment here, but I pulled this fix to try this snippet (posted by @paulirish in #618) and it *does* click if I comment out the `outerHTML` stuff.

```
(async() => {
const browser = await puppeteer.launch({headless: false, slowMo: 250});
await page.goto('https://www.w3schools.com/TAgs/tryit.asp?filename=tryhtml_button_test');

const frames = await page.frames();
const tryItFrame = frames.find(f => f.name() === 'iframeResult');
const framedButton = await tryItFrame.$('button');

const outerHTML = await framedButton.evaluate(e => e.parentNode.outerHTML);
console.log('the outerhtml: ', outerHTML); // to verify we're talking about the right element.. (we are.)

framedButton.click();
})();
```

How is the evaluate stuff supposed to work in this context?

In another script, I tried to select an input to type in some data but I keep getting `framedInput.type` is not a function:

```
    const framedInput = await iframe.$('input#fm-login-id');
    framedInput.click(usernameSelector);
    framedInput.type(username);
```

Am I being an idiot?!

JoelEinbinder:
@veb Paul's code was slightly out of date. `elementHandle.evaluate` was dropped. `elementHandle.type` is coming shortly in #801.

Edit: #943 does not add `elementHandle.evaluate`, I got confused.

aslushnikov (approved):

veb:
Thanks @JoelEinbinder I've tested those changes on #801 before it was merged and they work perfectly. I'm just waiting for this one to be merged and then selecting and typing into iframes will be worked just as intended 👍

PR #973, opened 2017-10-06
Allow using content shell instead of full chromium.

pavelfeldman:


utils/ChromiumDownloader.js
```diff
@@ -25,15 +25,24 @@ const removeRecursive = require('rimraf');
 const ProxyAgent = require('https-proxy-agent');
 const getProxyForUrl = require('proxy-from-env').getProxyForUrl;
 
-const DOWNLOADS_FOLDER = path.join(__dirname, '..', '.local-chromium');
+const DOWNLOADS_FOLDER = path.join(__dirname, '..', process.env.CONTENT_SHELL ? '.local-shell' : '.local-chromium');
```
aslushnikov, line 28: chromium downloader might be run in a different environment; the process.env.CONTENT_SHELL might not be defined there.

Can we plumb "product" as a third parameter to the downloader methods?

paambaati:
@pavelfeldman @aslushnikov Just curious, what advantages or features does content shell have over Chromium?

aslushnikov:
> Just curious, what advantages or features does content shell have over Chromium?

@paambaati it might be slimmer in size. Beware though: DevTools protocol is not fully functional there, and there's no intent to invest into it and no guarantees on what will be working in future.

So use at your own risk.

Closing in favor of #1987.

PR #975, opened 2017-10-06
feat(console): dispatch JSHandles as console arguments

### LINKED ISSUE #324: evaluate - strange execute sequence
scripts:

```js
(async () => {
  page.on('console', console.log);
    console.log('1. abc');
    console.log('2. ', this);
    console.log('3. ', this.document.title);
    console.log('4. ', this.window);
  });
  browser.close();
})();
```

output:
```bash
➜  headless node index.js
1. abc
3.  Google
2.  Window
4.  Window
```

aslushnikov:
This patch starts dispatching JSHandle instances as console arguments.
In doing so, we no longer do hops to the page to serialize page arguments, which
removes races between events.

BREAKING CHANGE: this changes the API of the ConsoleMessage.

Fixes #324.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1161,7 +1161,7 @@ Dialog's type, could be one of the `alert`, `beforeunload`, `confirm` and `promp
 [ConsoleMessage] objects are dispatched by page via the ['console'](#event-console) event.
 
 #### consoleMessage.args
-- <[Array]<[string]>>
+- <[Array]<[JSHandle]>>
 
 #### consoleMessage.text
 - <[string]>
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -1161,7 +1161,7 @@ Dialog's type, could be one of the `alert`, `beforeunload`, `confirm` and `promp
 [ConsoleMessage] objects are dispatched by page via the ['console'](#event-console) event.
 
 #### consoleMessage.args
-- <[Array]<[string]>>
+- <[Array]<[JSHandle]>>
```
vsemozhetbyt, line 1164: FWIW, this link is somehow broken here and in some other places.

aslushnikov: @vsemozhetbyt thanks, fixed in https://github.com/GoogleChrome/puppeteer/commit/3214bb73b05ea281bf49341e4e81cb8dc489a864

JoelEinbinder (approved):

PR #976, opened 2017-10-07
fix(doc): fix nits after JSHandles PR

vsemozhetbyt:
~~I am a bit confused about [`elementHandle.asElement()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#elementhandleaselement). What is its purpose and difference from [`jsHandle.asElement()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#jshandleaselement)? Should it stay?~~ (see https://github.com/GoogleChrome/puppeteer/pull/976#issuecomment-334936114 below)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -548,7 +548,7 @@ console.log(await resultHandle.jsonValue());
 await resultHandle.dispose();
 ```
 
-Shortcut for [page.mainFrame().executionContext().evaluateHandle(pageFunction, ...args)](#frameobjectpagefunction-args).
+Shortcut for [page.mainFrame().executionContext().evaluateHandle(pageFunction, ...args)](#executioncontextevaluatehandlepagefunction-args).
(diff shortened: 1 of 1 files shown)


Garbee:
The docs look fine for those two methods in question. `jsHandle` [returns](https://github.com/GoogleChrome/puppeteer/blob/0d0f9b79843da030013da02606ce3ea723175a1d/lib/ExecutionContext.js#L168) `null` since it is the base class. Once `elementHandle` extends it, it [returns](https://github.com/GoogleChrome/puppeteer/blob/0af0d7dba5e2248149211a36cc06c348b3cfc1c8/lib/ElementHandle.js#L40) the elementHandle object since it is an instance of jsHandle as the docs describe.

aslushnikov (approved):
nice!

PR #983, opened 2017-10-09
fix: return undefined when Page.evaluate encounters circular JSON

JoelEinbinder:
The new behavior of `page.evaluate` to convert to JSON makes it difficult to use if you don't care about the return value of your function. Consider 
```js
await page.evaluate(() => window.open('about:blank'));
await page.evaluate(() => $('body').hide());
```

I don't want to be paranoid about whether or not my evaluations unintentionally return circular json.


aslushnikov (approved):

aslushnikov:
Let's update documentation

PR #986, opened 2017-10-10
Use Typescript to lint JSDoc annotations

JoelEinbinder:
A re-upload of #219 from my new repository.

#65


lib/NetworkManager.js
```diff
@@ -116,7 +116,7 @@ class NetworkManager extends EventEmitter {
         response = 'ProvideCredentials';
         this._attemptedAuthentications.add(event.interceptionId);
       }
-      const {username, password} = this._credentials || {};
+      const {username, password} = this._credentials || {username: undefined, password: undefined};
```
aslushnikov, line 119: is this necessary now?

npm-debug.log
```diff
@@ -0,0 +1,48 @@
+0 info it worked if it ends with ok
```
bootstraponline: I don't think we want to check in `npm-debug.log`?

aslushnikov:
Awwwwsome.

Could you please minimize the diff (there are a lot of unnecessary changes like "!" added/removed). And it's good to go.

Also, npm-debug is not needed.

npm-debug.log
plantain-00: `*.log` should be gitignored

lib/NetworkManager.js
JoelEinbinder: Yes. I think it will help more than hurt.

aslushnikov (approved):

lib/Multimap.js
```diff
@@ -14,19 +14,11 @@
...
- */
 class Multimap {
   constructor() {
-    /** @type {!Map<K, !Set<!V>>} */
     this._map = new Map();
   }
 
-  /**
-   * @param {K} key
-   * @param {V} value
```
paulirish, line 28: @JoelEinbinder just curious... TSC didn't need these annotations? And how did you know they were unnecessary ?

JoelEinbinder: It’s a bug in typescript. https://github.com/Microsoft/TypeScript/issues/18135

stereobooster:
As soon as annotations are "compatible" with typescript it should be easy to generate TS definitions from JSDoc with something like this:
- https://github.com/brettle/jsdoc-to-typescript-declaration
- https://www.npmjs.com/package/tsd-jsdoc

PR #990, opened 2017-10-10
fix(Request): convert resourceType to all small-caps

aslushnikov:
This patch moves resourceType to be all small-caps. This aligns
with our convention that all string constants should be smallcaps.

BREAKING CHANGE: this patch changes the constants of the
request.resourceType to be all small-caps.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1696,7 +1696,7 @@ Contains the request's post body, if any.
 - <[string]>
 
 Contains the request's resource type as it was perceived by the rendering engine.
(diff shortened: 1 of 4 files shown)


vsemozhetbyt:
See also [examples/block-images.js#L27](https://github.com/GoogleChrome/puppeteer/blob/aa58f25bc1c7f8cf72c70c453f036bb873349164/examples/block-images.js#L27)

JoelEinbinder (approved):

Issue #918, opened 2017-09-30
userDataDir + headless  = browser and script (node) hang


vsemozhetbyt:
* Puppeteer version: v0.12.0-alpha
* Platform / OS version: Windows 7 x64

Create a test script and an empty folder `test-profile-dir`:
```js
'use strict';

const puppeteer = require('puppeteer');

(async function main() {
  try {
    const browser = await puppeteer.launch({
      headless: true,
      userDataDir: 'test-profile-dir',
    });
    await browser.close();
  } catch (err) {
    console.error(err);
  }
})();
```
Browser and script (node) hang (do not exit). After `Ctrl+C`, the script (node) exits, but some Chromium processes remain.

It seems [`taskkill /pid`](https://github.com/GoogleChrome/puppeteer/pull/876) does not close headless process without `/f` flag (`WM_CLOSE` message is not effective without the window?) — but with `/f` flag the process [does not exit gracefully](https://github.com/GoogleChrome/puppeteer/issues/839). Stalemate(

vsemozhetbyt:
It seems [`taskkill /pid`](https://github.com/GoogleChrome/puppeteer/pull/876/files) does not close headless process without `/f` flag — but with `/f` flag the process does not exit gracefully. Stalemate(

vsemozhetbyt:
Some more notes:

1. `response.request().headers` does not contains cookies in both `headless: false` and `headless: true` modes.

2. `console.log(await page.cookies('https://twitter.com/'));` contains many cookies in the `headless: false` mode. In the `headless: true` mode it gives an empty array `[]`.

PR #992, opened 2017-10-10
feat(Dev): Allow development under Windows platform

elisherer:
Allow development under Windows platform:

### Problems resolved
* Setting environment variables using `cross-env` since Windows requires the SET command.
* Calling Jasmine in the script `debug-unit` using jasmine's JavaScript binary instead of shell.
* Add `/test/test-user-data-dir*` to .gitignore since temporary user data directories, in case of test fails, remains in the test directory.

### Problems which didn't resolve
* Two unit tests of using a custom user data directory hangs when trying to close the browser. Currently patched by crossing those when in Windows platform.
Output:
```
Failures:
1) Puppeteer Puppeteer.launch userDataDir option
  Message:
    Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
  Stack:
    Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
        at ontimeout (timers.js:469:11)
        at tryOnTimeout (timers.js:304:5)
        at Timer.listOnTimeout (timers.js:264:5)

2) Puppeteer Puppeteer.launch userDataDir argument
  Message:
    Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
  Stack:
    Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
        at ontimeout (timers.js:469:11)
        at tryOnTimeout (timers.js:304:5)
        at Timer.listOnTimeout (timers.js:264:5)
```


package.json
```diff
@@ -9,13 +9,13 @@
   },
   "scripts": {
     "unit": "jasmine test/test.js",
-    "debug-unit": "DEBUG_TEST=true node --inspect-brk ./node_modules/.bin/jasmine test/test.js",
```
aslushnikov, line 12: out of curiosity, why changing `.bin/jasmine` to `jasmine/bin/jasmine.js`?

test/test.js
```diff
@@ -102,7 +103,8 @@ describe('Puppeteer', function() {
       await neverResolves;
       expect(error.message).toContain('Protocol error');
     }));
-    it('userDataDir option', SX(async function() {
+    // Windows has issues running Chromium using a custom user data dir. It hangs when closing the browser.
```
aslushnikov: This seems to be the https://github.com/GoogleChrome/puppeteer/issues/918

Could you please refer to the issue in the comment so that we keep track of it?

test/test.js
```diff
@@ -111,7 +113,7 @@ describe('Puppeteer', function() {
       expect(fs.readdirSync(userDataDir).length).toBeGreaterThan(0);
       rm(userDataDir);
     }));
-    it('userDataDir argument', SX(async function() {
+    (windows ? xit : it)('userDataDir argument', SX(async function() {
```
aslushnikov: let's refer to #918 here as well

package.json
elisherer: I had an exception:
```
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
          ^^^^^^^
SyntaxError: missing ) after argument list
```

This is where I found the solution https://github.com/gotwarlost/istanbul#usage-on-windows

test/test.js
elisherer: Done

test/test.js
elisherer: Done

aslushnikov (approved):
Thanks, these are good changes. I wonder though why @JoelEinbinder didn't experience any of these during his time on Windows machine.

JoelEinbinder (approved):
Looks good. I hit these problems but was just living with the pain.

Issue #949, opened 2017-10-04
Replace `injectFile` to `injectScript` and add `injectCSS`


SamVerschueren:
When browsing trough the docs and using the API, I noticed that Puppeteer has an `injectFile` method. The name might be confusing, because you might think it works for a CSS file as well. I think it would be better to rename this method to `injectScript` and also add a `injectCSS` method.

Just an idea though :).

vsemozhetbyt:
CSS injecting seems to be possible when https://github.com/GoogleChrome/puppeteer/pull/947 is landed.

aslushnikov:
@vsemozhetbyt `injectFile` injects from file system, whereas `addStyleTag` adds a url. I'd rather unify the two methods since it's confusing.

Something like:

```js
page.addScriptTag({
  url: '...', // either URL or PATH could be specified
  path: '..',
});
```

SamVerschueren:
`addScriptTag` and `addStyleTag` currently already work with a local file path, that's what the tests do as well. Not sure if it's a good idea though, that's probably the reason why `injectFile` exists.

If we would go with your solution, an object with `url` and `path`, it might be nice to have a `content` option as well.

```js
page.addScriptTag({
  url: '...',
  path: '..',
  content: `document.write('Hello World')`
});
```

SamVerschueren:
@aslushnikov Let me know what the preferred path forward is for this. Happy to do a PR to fix this.

aslushnikov:
> @aslushnikov Let me know what the preferred path forward is for this. Happy to do a PR to fix this.

@SamVerschueren Your suggestion with url/path/content options sounds good to me:

```js
page.addScriptTag({
  url: '...',
  path: '..',
  content: `document.write('Hello World')`
});
```

if you have time, I'd be happy to review a PR. (note: please make sure to use sourceURL for the 'content' option of the addScriptTag/addStyleTag methods).

If you're short on time, I'd rather do it myself: this is one of the bugs that we'd like to have closed soon.

SamVerschueren:
I'll try to work on it today.

Can you elaborate on
> make sure to use sourceURL for the 'content' option of the addScriptTag/addStyleTag methods

Or do you mean 
> make sure to use sourceURL for the `path` option

aslushnikov:
>Or do you mean
>
>>make sure to use sourceURL for the path option

That's right, I meant `path` option.

SamVerschueren:
PR send!

PR #996, opened 2017-10-10
feat(Frame): Support options in `addScriptTag` and `addStyleTag`

SamVerschueren:
This PR implements #949.

- Deprecated `injectFile` as it was confusing with the `addScriptTag`
- Accept an `options` object in `addScriptTag` which supports properties `url`, `path` and `content`.
- Accept an `options` object in `addStyleTag` which supports properties `url`, `path` and `content`.

// @aslushnikov

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -425,7 +425,7 @@ Adds a `<script>` tag into the page with the desired url or content.
 
 Shortcut for [page.mainFrame().addScriptTag(options)](#frameaddscripttagoptions).
 
-#### page.addStyleTag(url)
+#### page.addStyleTag(options)
 - `options` <[Object]>
   - `url` <[string]> Url of the `<link>` tag.
   - `path` <[string]> Path to the CSS file to be injected into frame. If `path` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
--- a/docs/api.md
+++ b/docs/api.md
@@ -35,8 +35,8 @@
     + [page.$$(selector)](#pageselector)
     + [page.$$eval(selector, pageFunction[, ...args])](#pageevalselector-pagefunction-args)
     + [page.$eval(selector, pageFunction[, ...args])](#pageevalselector-pagefunction-args)
-    + [page.addScriptTag(url)](#pageaddscripttagurl)
-    + [page.addStyleTag(url)](#pageaddstyletagurl)
+    + [page.addScriptTag(options)](#pageaddscripttagoptions)
+    + [page.addStyleTag(options)](#pageaddstyletagoptions)
     + [page.authenticate(credentials)](#pageauthenticatecredentials)
     + [page.click(selector[, options])](#pageclickselector-options)
     + [page.close()](#pageclose)
@@ -56,7 +56,6 @@
     + [page.goForward(options)](#pagegoforwardoptions)
     + [page.goto(url, options)](#pagegotourl-options)
     + [page.hover(selector)](#pagehoverselector)
-    + [page.injectFile(filePath)](#pageinjectfilefilepath)
     + [page.keyboard](#pagekeyboard)
     + [page.mainFrame()](#pagemainframe)
     + [page.mouse](#pagemouse)
@@ -113,12 +112,11 @@
     + [frame.$$(selector)](#frameselector)
     + [frame.$$eval(selector, pageFunction[, ...args])](#frameevalselector-pagefunction-args)
     + [frame.$eval(selector, pageFunction[, ...args])](#frameevalselector-pagefunction-args)
-    + [frame.addScriptTag(url)](#frameaddscripttagurl)
-    + [frame.addStyleTag(url)](#frameaddstyletagurl)
+    + [frame.addScriptTag(options)](#frameaddscripttagoptions)
+    + [frame.addStyleTag(options)](#frameaddstyletagoptions)
     + [frame.childFrames()](#framechildframes)
     + [frame.evaluate(pageFunction, ...args)](#frameevaluatepagefunction-args)
     + [frame.executionContext()](#frameexecutioncontext)
-    + [frame.injectFile(filePath)](#frameinjectfilefilepath)
     + [frame.isDetached()](#frameisdetached)
     + [frame.name()](#framename)
     + [frame.parentFrame()](#frameparentframe)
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -389,21 +387,27 @@ const html = await page.$eval('.main-container', e => e.outerHTML);
...
+  - `content` <[string]> Raw JavaScript content to be injected into frame.
+- returns: <[Promise]> which resolves when the script's onload fires or when the script content was injected into frame.
 
-Adds a `<script>` tag into the page with the desired url. Alternatively, a local JavaScript file can be injected via [`page.injectFile`](#pageinjectfilefilepath) method.
+Adds a `<script>` tag into the page with the desired url or content.
 
-Shortcut for [page.mainFrame().addScriptTag(url)](#frameaddscripttagurl).
+Shortcut for [page.mainFrame().addScriptTag(options)](#frameaddscripttagoptions).
 
 #### page.addStyleTag(url)
```
vsemozhetbyt: page.addStyleTag(options)

SamVerschueren: Good catch!

SamVerschueren:
If you want I could add backwards compatibility. Something like

```js
if (typeof options === 'string') {
     options = {url: options};
}
```

I'd rather only support the object though. But not my call to make :).

vsemozhetbyt:
Not sure, but maybe it would be more consistent if all the 'frame' instances are replaced by 'page' or 'main frame' in the first two descriptions.

SamVerschueren:
I agree. I actually just used the same wording as what was there before. I think it would be more clear if we just wrote `page` instead of frame when explaining `page.addScriptTag` and `page.addStyleTag`.

lib/FrameManager.js
```diff
@@ -291,55 +291,93 @@ class Frame {
...
-      fs.readFile(filePath, 'utf8', (err, data) => {
-        if (err) return reject(err);
-        resolve(data);
+  async addScriptTag(options) {
+    if (typeof options.url === 'string')
+      return this.evaluate(addScriptUrl, options.url);
+
+    if (typeof options.path === 'string') {
+      let contents = await new Promise((resolve, reject) => {
+        fs.readFile(options.path, 'utf8', (err, data) => {
```
aslushnikov: Let me merge in #879 so that we can reuse the readFileAsync here.

SamVerschueren: Good idea!!

test/test.js
```diff
@@ -417,21 +417,6 @@ describe('Page', function() {
...
   });
 
-  describe('Page.injectFile', function() {
-    it('should work', SX(async function() {
-      const helloPath = path.join(__dirname, 'assets', 'injectedfile.js');
-      await page.injectFile(helloPath);
-      const result = await page.evaluate(() => __injected);
-      expect(result).toBe(42);
-    }));
-    it('should include sourcemap', SX(async function() {
```
aslushnikov, line 435: let's keep a test that checks for sourcemaps

lib/FrameManager.js
aslushnikov: @SamVerschueren ok it's in. Please, make a readFileAsync with `helper.promisify` and use it here.

lib/FrameManager.js
```diff
@@ -291,55 +291,93 @@ class Frame {
...
+          if (err) return reject(err);
+          resolve(data);
+        });
+      });
+
+      return this.evaluate(addStyleContent, contents);
+    }
+
+    if (typeof options.content === 'string')
+      return this.evaluate(addStyleContent, options.content);
```
aslushnikov: let's throw if neither of the options was given (would help with migration)

lib/FrameManager.js
```diff
@@ -291,55 +291,93 @@ class Frame {
...
-  /**
-   * @param {string} url
-   */
-  async addScriptTag(url) {
-    return this.evaluate(addScriptTag, url);
+      return this.evaluate(addScriptContent, contents);
+    }
+
+    if (typeof options.content === 'string')
+      return this.evaluate(addScriptContent, options.content);
```
aslushnikov: Ditto: let's throw if neither of the options was given (would help with migration)

test/test.js
SamVerschueren: Missed that one, good catch!

SamVerschueren:
Processed the feedback. Let me know if I can do anything else.

lib/FrameManager.js
```diff
@@ -303,55 +305,87 @@ class Frame {
...
+   * @return {!Promise}
    */
-  async addStyleTag(url) {
-    return this.evaluate(addStyleTag, url);
+  async addStyleTag(options) {
+    if (typeof options.url === 'string')
+      return this.evaluate(addStyleUrl, options.url);
+
+    if (typeof options.path === 'string') {
+      const contents = await readFileAsync(options.path, 'utf8');
```
aslushnikov: Let's add a sourceURL here as well. You won't be able to write a test for it, but debugging would benefit from it.

aslushnikov (commented):
Thanks @SamVerschueren. One last nit before we can merge it in.

lib/FrameManager.js
SamVerschueren: Good point! Added it, as well as a test which grabs the content of the style tag and checks if the path is present.

lib/FrameManager.js
```diff
@@ -303,55 +305,88 @@ class Frame {
...
    */
-  async addStyleTag(url) {
-    return this.evaluate(addStyleTag, url);
+  async addStyleTag(options) {
+    if (typeof options.url === 'string')
+      return this.evaluate(addStyleUrl, options.url);
+
+    if (typeof options.path === 'string') {
+      let contents = await readFileAsync(options.path, 'utf8');
+      contents += '//# sourceURL=' + options.path.replace(/\n/g, '');
```
aslushnikov: @SamVerschueren There's no line comment in css; css sourceurl/sourcemap is put in a block comment:
```js
contents += '/*# sourceURL=' + options.path.replace(/\n/g, '') + '*/';
```

SamVerschueren: Oops, too spoiled using SCSS I guess :).

aslushnikov (approved):

aslushnikov:
Thanks! This is really nice.

SamVerschueren:
Sweet! Thanks for guiding me through this and merging it in @aslushnikov :beers:!

PR #998, opened 2017-10-10
Add test for #750 - Throw error with stack trace when launching with …

austinkelleher:
…an invalid executablePath option.

When https://github.com/GoogleChrome/puppeteer/pull/863 was opened as opposed to my original PR (https://github.com/GoogleChrome/puppeteer/pull/751), the test was not carried over.

/cc @aslushnikov @JoelEinbinder


aslushnikov (approved):
Thanks for the test

PR #1001, opened 2017-10-10
fix(Page.getMetrics): do not report a few metrics

aslushnikov:
A few of the reported metrics in the Page.getMetrics are already
reported with lifecycle events.

This patch excludes these metrics for now.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -663,9 +663,6 @@ If there's no element matching `selector`, the method throws an error.
   - `TaskDuration` <[number]> Combined duration of all tasks performed by the browser.
   - `JSHeapUsedSize` <[number]> Used JavaScript heap size.
   - `JSHeapTotalSize` <[number]> Total JavaScript heap size.
-  - `FirstMeaningfulPaint` <[number]> Timestamp of the first meaningful paint event.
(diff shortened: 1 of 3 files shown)


a1ph (approved):

Everettss:
Oh god. Why did you @aslushnikov remove this metrics?
Can you point how to now get `FirstMeaningfulPaint` `DomContentLoaded` `NavigationStart` times without:
```javascript
await page._client.send('Performance.enable');
const metrics = await page._client.send('Performance.getMetrics');
```
?
It is very confusing that `page.getMetrics()` doesn't report basic metrics (!), and we are forced to search them in some "lifecycle events" (whatever they are).

aslushnikov:
>Oh god. Why did you @aslushnikov remove this metrics?
> Can you point how to now get FirstMeaningfulPaint DomContentLoaded NavigationStart times

@Everettss you can get `DOMContentLoaded` and `NavigationStart` from `window.performance` DOM api. But, we'll bring them back!

There is too much movement in the `First*Paint` metrics, so we're hesitant exposing them at this point. What are your usecases for them?

inikulin:
@aslushnikov 

>  There is too much movement in the First*Paint metrics, so we're hesitant exposing them at this point. What are your usecases for them?

Like... measuring performance?

PR #1005, opened 2017-10-11
feat(Page): introduce Page.queryObjects

### LINKED ISSUE #304: Add memory heap snapshot API
# Problem
I see there is an API for starting and stoping a timeline recording `await page.tracing.start({path: 'trace.json'});` but I would like to request the ability to take a memory heap snapshot and the ability  to write expectations on the snapshot data.

My use case is roughly:
 - load a page
 - load a component & do stuff
 - remove the component
 - force a garbage collection (would love a method for this as well!)
 - take a heap snapshot
 - Verify that the component is not retained in memory somehow

This would be amazing to be able to have automated headless memory leak regression testing!

Maybe something like:
```
var heapsnap = await page.heap.snapshot('heap1.snapshot');
expect(!heapsnap.contains('SomeClassName'))
```

Bonus points for an API for the memory allocation timeline that you could query for the number of objects of a type added / removed during the timeline.

aslushnikov:
This patch introduces `Page.queryObjects` and
`ExecutionContext.queryObjects` methods to query JavaScript heap
for objects with a certain prototype.

Fixes #304.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -330,6 +330,30 @@ describe('Page', function() {
     }));
   });
 
+  describe('ExecutionContext.queryObjects', function() {
+    it('should work', SX(async function() {
+      // Instantiate an object
+      await page.evaluate(() => window.map = new Map());
+      const prototypeHandle = await page.evaluateHandle(() => Map.prototype);
+      const objectsHandle = await page.queryObjects(prototypeHandle);
+      const count = await page.evaluate(objects => objects.length, objectsHandle);
+      expect(count).toBe(1);
+    }));
+    it('should fail for disposed handles', SX(async function() {
+      const prototypeHandle = await page.evaluateHandle(() => HTMLBodyElement.prototype);
+      await prototypeHandle.dispose();
+      let error = null;
+      await page.queryObjects(prototypeHandle).catch(e => error = e);
+      expect(error.message).toBe('Prototype JSHandle is disposed!');
+    }));
+    it('should fail primitive values as prototypes', SX(async function() {
+      const prototypeHandle = await page.evaluateHandle(() => 42);
+      let error = null;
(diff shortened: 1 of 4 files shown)


test/test.js
```diff
@@ -330,6 +330,30 @@ describe('Page', function() {
     }));
   });
 
+  describe('ExecutionContext.queryObjects', function() {
+    it('should work', SX(async function() {
+      // Instantiate an object
+      await page.evaluate(() => window.map = new Map());
+      const prototypeHandle = await page.evaluateHandle(() => Map.prototype);
+      const objectsHandle = await page.queryObjects(prototypeHandle);
+      const count = await page.evaluate(objects => objects.length, objectsHandle);
```
JoelEinbinder: To make me feel safer, can we actually check that the object here is the map and is usable.

JoelEinbinder:
This is really exciting. It feels a bit clumsy to use, but maybe that is ok. I was hoping for something like `page.queryObjects(string)` -> `Array<JSHandle>`, but I can easily write that helper myself.

test/test.js

aslushnikov:
> This is really exciting. It feels a bit clumsy to use, but maybe that is ok. I was hoping for something like page.queryObjects(string) -> Array<JSHandle>, but I can easily write that helper myself.

Disposing an array of JSHandles is not very nice and talkative from protocol's point of view:

```js
await Promise.all(handles.map(handle => handle.dispose()));
```

Most of the time you'll be satisfied with inspecting the objects with `page.evaluate`, thus this approach:

```js
const objectPrototype = await page.evaluateHandle(() => Object.prototype);
const objectsHandle = await page.queryObjects(objectPrototype);
const count = await page.evaluate(objects => objects.length, objectsHandle);
```

alexkozy (approved):
Nice!

PR #1006, opened 2017-10-11
feat(Page): introduce Page.$$eval method

### LINKED ISSUE #625: Introduce page.$eval and page.$$eval

aslushnikov:
This patch adds a Page.$$eval method that runs document.querySelectorAll
and passes resulting array to the page function.

Fixes #625.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -240,6 +240,20 @@ class Frame {
 
   /**
    * @param {string} selector
+   * @param {Function|string} pageFunction
+   * @param {!Array<*>} args
+   * @return {!Promise<(!Object|undefined)>}
+   */
+  async $$eval(selector, pageFunction, ...args) {
+    const arrayHandle = await this._context.evaluateHandle(selector => Array.from(document.querySelectorAll(selector)), selector);
+    args = [arrayHandle].concat(args);
+    const result = await this.evaluate(pageFunction, ...args);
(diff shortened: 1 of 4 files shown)


lib/FrameManager.js
```diff
@@ -240,6 +240,20 @@ class Frame {
 
   /**
    * @param {string} selector
+   * @param {Function|string} pageFunction
+   * @param {!Array<*>} args
+   * @return {!Promise<(!Object|undefined)>}
+   */
+  async $$eval(selector, pageFunction, ...args) {
+    const arrayHandle = await this._context.evaluateHandle(selector => Array.from(document.querySelectorAll(selector)), selector);
+    args = [arrayHandle].concat(args);
```
JoelEinbinder: nit: Can avoid reassigning to args
`await this.evaluate(pageFunction, arrayHandle, ...args)`

JoelEinbinder (approved):

JoelEinbinder (approved):

Issue #723, opened 2017-09-08
Keyboard API is confusing 


aslushnikov:
Keyboard API is very powerful but confusing.

Questions/clarifications:
- [x] Why's there no `keyboard.press()`?
- [x]  Explain the difference between `keyboard.press('a')` and `keyboard.press('a', {text: 'a'})`. Also, should there be a difference?
- [x] Explain why `await keyboard.down('Shift'); await keyboard.type('hello');` **does not** produce 'HELLO'.

thetutlage:
Any chance to get some updates on it?

aslushnikov:
@thetutlage no real updates so far; but @JoelEinbinder has something in works for this.

thetutlage:
Oh that’s nice. Thanks for sharing

aslushnikov:
For this issue to be closed: 
- add API example
- explain why await keyboard.down('Shift'); await keyboard.type('hello'); does not produce 'HELLO'

christophby:
any news on this?

PR #1016, opened 2017-10-12
feat: make keyboard.down generate input event by default

JoelEinbinder:
Breaking change:
Except for in really strange circumstances, pressing the key `a` will generate an `input` event with the text `'a'`. This patch makes it the default behavior in puppeteer.

#723


aslushnikov (approved):
Clarification:

This patch addresses bullet n.2 of #723:

> Explain the difference between keyboard.press('a') and keyboard.press('a', {text: 'a'}). Also, should there be a difference?

With this patch, there's no longer any difference.

PR #1017, opened 2017-10-12
fix: account for margin, padding, and border in element.boxModel

### LINKED ISSUE #1010: ElementHandle screenshot does not count padding (or add options to count the padding/margin)
### Steps to reproduce

* Puppeteer version: 0.12.0-alpha
* Platform / OS version:
* URLs (if applicable): http://example.com

Take screenshot using ElementHandler.

```js
await page.goto("http://example.com");
const title = await page.$('h1');
const screenshot = await title.screenshot({path: "justTitle.png"});
```

Take screenshot of the highlighted area as from chrome. When I mean highlighted, I mean the margin/padding.

image

**What happens instead?**
justtitle

JoelEinbinder:
fixes #1010


aslushnikov (approved):
lgtm

let's make a rotated test case then!

Issue #989, opened 2017-10-10
No way to close a browser instance that was opened with connect()


ebidel:
How does one close a browser instance that was opened by another script? Assuming you don't have access to that process.

Repro:
1. Run `node connect.js &`
2. Run `wsURL=<ws url printed to console in step 1> node connect.js`

```js
const puppeteer = require('puppeteer');

const browserWSEndpoint = process.env.wsURL || null;

(async () => {
  if (!browserWSEndpoint) {
    const browser = await puppeteer.launch();
    console.log(browser.wsEndpoint());
    return;
  }

  const browser2 = await puppeteer.connect({browserWSEndpoint});
  const page = await browser2.newPage();
  await page.goto('https://example.com');

  await browser2.close(); // Expected this to close the instance. I
})();
```

Thought about navigating to `about:quit` but that doesn't work: `Error: Failed to navigate: about:quit`

aslushnikov:
On a first glance, this could be addressed with the `browser.close()` protocol method.

On a second thought though, the `browser.connect` method doesn't own the process, and thus shouldn't be able to close it. The service that started the browser should do the clients bookkeeping and manage browser instance lifetime.

What's your usecase?

ebidel:
Protocol method sounds good. Making some examples for CDS, tried this, and it didn't work as expected.

 > The service that started the browser should do the clients bookkeeping and manage browser instance lifetime.

It's not always trivial to get back to the original service. One use case might be a long running chrome instance in cloud vm that you want to shut down after a period of time (cleanup / memory consumption issues). Your script or handler might be different than the one that launched Chrome.

Basically, if you can `connect()` to a browser, you should be able to close that browser. This makes `connect()` and `launch()` return the same thing. All methods continue to do the right thing.

aslushnikov:
The `browser.close()` in its current form does make sense. Having both `browser.disconnect()` and `browser.close()` is confusing.

I like the `puppeteer.terminate(wsEndpoint)`. This would be doable with protocol method.

ebidel:
Or what about allowing an opt param:`browser.close(wsEndpoint)`? That could even call `puppeteer.terminate` internally. My thinking is that it's simpler to have one API call to kill chrome.

PR #1028, opened 2017-10-12
fix: close chrome gracefully with Browser.close

### LINKED ISSUE #921: userDataDir + headless = lost authorization
* Puppeteer version: v0.12.0-alpha
* URLs (if applicable): any that needs authorization

1. Create a test script and an empty folder `test-profile-dir`:
```js
'use strict';

(async function main() {
  try {
      headless: false,
      userDataDir: 'test-profile-dir',
    });

    await page.goto('https://twitter.com/');

    console.log(await page.evaluate(() => document.title));
    console.log(await page.evaluate(() => document.cookie));

    // await browser.close();
  } catch (err) {
    console.error(err);
  }
})();
```
You will see something like this:
```console
Twitter. It's what's happening.

personalization_id=...; guest_id=...; ct0=...
```
Sign in into Twitter and close the browser.

2. Run the same script second time (`await browser.close();` can be uncommented). You will see something like this:
```console
Twitter

personalization_id=...; guest_id=...; ct0=...; ads_prefs=...; remember_checked_on=...;
twid=...; tip_nightmode=...; _ga=...; _gid=...; dnt=...; lang=...
```
3. Run the same script, but with `headless: true,` The output is the same as before authorization:
```console
Twitter. It's what's happening.

personalization_id=...; guest_id=...; ct0=...
```

I have tried various sites, all of them seem to lose authorization in headless mode.

Some more notes:

* `response.request().headers` does not contain cookies in both `headless: false` and `headless: true` modes.

* `console.log(await page.cookies('https://twitter.com/'));` contains many cookies in the `headless: false` mode. In the `headless: true` mode it gives an empty array `[]`.

JoelEinbinder:
`browser.close()` now closes the browser, regardless of whether or not it was created with `puppeteer.launch()` or `puppeteer.connect()`
#989

`browser.disconnect()` is a new method which disconnects Puppeteer from the browser, but does not close the Chromium process.

Should fix #921 and #918

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -260,14 +260,31 @@ puppeteer.launch().then(async browser => {
 });
 ```
 
+An example of disconnecting from and reconnecting to a [Browser]:
+```js
+const puppeteer = require('puppeteer');
+
+puppeteer.launch().then(async browser => {
+  // Store the endpoint to be able to reconnect to Chromium
+  const browserWSEndpoint = browser.wsEndpoint();
+  // Disconnect puppeteer from Chromium
+  browser.disconnect();
+
+  // Use the endpoint to reestablish a connection
+  const browser2 = await puppeteer.connect({browserWSEndpoint});
+  // Close Chromium
+  await browser2.close();
+});
+```
+
 #### browser.close()
 - returns: <[Promise]>
 
-Closes browser with all the pages (if any were opened). The browser object itself is considered to be disposed and can not be used anymore.
+Closes Chromium and all of its pages (if any were opened). The browser object itself is considered disposed and cannot be used anymore.
 
 #### browser.disconnect()
 
-Disconnects Puppeteer from the browser, but leaves the Chromium process running.
+Disconnects Puppeteer from the browser, but leaves the Chromium process running. After calling `disconnect`, the browser object is considered disposed and cannot be used anymore.
 
 #### browser.newPage()
 - returns: <[Promise]<[Page]>> Promise which resolves to a new [Page] object.
--- a/docs/api.md
+++ b/docs/api.md
@@ -14,6 +14,7 @@
   * [puppeteer.launch([options])](#puppeteerlaunchoptions)
 - [class: Browser](#class-browser)
   * [browser.close()](#browserclose)
+  * [browser.disconnect()](#browserdisconnect)
   * [browser.newPage()](#browsernewpage)
   * [browser.version()](#browserversion)
   * [browser.wsEndpoint()](#browserwsendpoint)
@@ -263,6 +264,10 @@ puppeteer.launch().then(async browser => {
 - returns: <[Promise]>
 
 Closes browser with all the pages (if any were opened). The browser object itself is considered to be disposed and can not be used anymore.
+
+#### browser.disconnect()
+
+Disconnects Puppeteer from the browser, but leaves the Chromium process running.
 
 #### browser.newPage()
 - returns: <[Promise]<[Page]>> Promise which resolves to a new [Page] object.
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -126,12 +126,13 @@ class Launcher {
     const listeners = [ helper.addEventListener(process, 'exit', killChrome) ];
     if (options.handleSIGINT !== false)
       listeners.push(helper.addEventListener(process, 'SIGINT', killChrome));
-
+    /** @type {?Connection} */
+    let connection = null;
     try {
       const connectionDelay = options.slowMo || 0;
       const browserWSEndpoint = await waitForWSEndpoint(chromeProcess, options.timeout || 30 * 1000);
-      const connection = await Connection.create(browserWSEndpoint, connectionDelay);
-      return new Browser(connection, options, killChrome);
+      connection = await Connection.create(browserWSEndpoint, connectionDelay);
+      return new Browser(connection, options, killChrome.bind(null, connection));
     } catch (e) {
       killChrome();
       throw e;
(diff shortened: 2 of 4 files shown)


JoelEinbinder:
@vsemozhetbyt @elisherer does this work with your Windows setups? Mine is pretty bizarre.

ebidel:
Very cool. Thanks for the quick fix on this one!

lib/Launcher.js
```diff
@@ -142,20 +142,12 @@ class Launcher {
...
-          // Terminate chrome gracefully.
-          if (process.platform === 'win32')
-            childProcess.execSync(`taskkill /pid ${chromeProcess.pid}`);
-          else
-            process.kill(-chromeProcess.pid, 'SIGTERM');
-        }
+      if (chromeProcess.pid && temporaryUserDataDir) {
+        // Force kill chrome.
+        if (process.platform === 'win32')
+          childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T /F`);
```
aslushnikov: this is racing with protocol's browser.close. What happens if browser has already been closed? Could it be that we might terminate something else (in case this pid got reused)?

JoelEinbinder: Outdated, I check !chromeProcess.killed && chromeProcess.connected

vsemozhetbyt:
1. This seems to fix https://github.com/GoogleChrome/puppeteer/issues/918, but Ctrl+C seems not to work:

```js
'use strict';

(async function main() {
  try {
      headless: true,
      userDataDir: 'test-profile-dir',
    });
    // await browser.close();
  } catch (err) {
    console.error(err);
  }
})();
```

If I type Ctrl+C once, nothing happens. If I type Ctrl+C second time, node exits, but Chromium remains.

vsemozhetbyt:
2. This DOES NOT fix https://github.com/GoogleChrome/puppeteer/issues/921: headless Chromium still does not preserve authorization (does not send cookies?).

fhmd4k:
to fix #918:  in the `lib/Launcher.js` find this spawn chrome line:

`const chromeProcess = childProcess.spawn(chromeExecutable, chromeArguments, { detached: false });`
set `detached: false`

fhmd4k:
#921:
In the headless mode, chromium can't save storage and cookie. But in the normal mode, it can.
I think it's chromium bug.

You can test it without puppeteer. Run chromium with the specified arguments`chromiumPath --user-data-dir=path --headless -- "https://www.google.com"`

JoelEinbinder:
It appears to be working with localStorage but not with cookies. I'll look into it further.

lib/Launcher.js
aslushnikov: This does reduce probability but doesn't fix the race.

lib/Launcher.js
```diff
@@ -142,20 +142,12 @@ class Launcher {
...
-          // Terminate chrome gracefully.
-          if (process.platform === 'win32')
-            childProcess.execSync(`taskkill /pid ${chromeProcess.pid}`);
-          else
-            process.kill(-chromeProcess.pid, 'SIGTERM');
-        }
+      if (chromeProcess.pid && temporaryUserDataDir && !chromeProcess.killed && chromeProcess.connected) {
+        // Force kill chrome.
+        if (process.platform === 'win32')
+          childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T /F`);
```
aslushnikov: For the record: this still is racy

sukrosono:
I will wait this one, can we expect [Browser.close](https://chromedevtools.github.io/devtools-protocol/tot/Browser/#method-close) to return boolean? Currently i check it just return empty object.

lib/Launcher.js
JoelEinbinder: Should be not racy anymore.

JoelEinbinder:
@brutalcrozt Browser.close tries to close the browser, but it doesn't make any guarantees about when or if that will happen.

lib/Launcher.js
```diff
@@ -126,12 +126,13 @@ class Launcher {
...
-
+    /** @type {?Connection} */
+    let connection = null;
     try {
       const connectionDelay = options.slowMo || 0;
       const browserWSEndpoint = await waitForWSEndpoint(chromeProcess, options.timeout || 30 * 1000);
-      const connection = await Connection.create(browserWSEndpoint, connectionDelay);
-      return new Browser(connection, options, killChrome);
+      connection = await Connection.create(browserWSEndpoint, connectionDelay);
+      return new Browser(connection, options, killChrome.bind(null, connection));
```
aslushnikov: no need to bind - you don't use it.

docs/api.md
```diff
@@ -264,6 +265,10 @@ puppeteer.launch().then(async browser => {
 
 Closes browser with all the pages (if any were opened). The browser object itself is considered to be disposed and can not be used anymore.
 
+#### browser.disconnect()
+
+Disconnects Puppeteer from the browser, but leaves the Chromium process running.
```
aslushnikov: ... This is needed to free up resources allocated for browser connection.

(and also a small example of puppeteer.connect + browser.disconnect)

test/test.js
```diff
@@ -170,14 +163,28 @@ describe('Puppeteer', function() {
     }));
   });
   describe('Puppeteer.connect', function() {
-    it('should work', SX(async function() {
+    fit('should be able to connect multiple times to the same browser', SX(async function() {
```
aslushnikov: `s/fit/it/`

aslushnikov (approved):

sukrosono:
I subscribe to this PR, i was thinking that i will get notified when it got merged. But it wasn't :confused:

ronalb:
ok I'll subscribe also

PR #1029, opened 2017-10-13
WaitFor options consider ancestors

ChristianDavis:
@ebidel you were right, visibility is rabbit hole. But, I think this is a simple version of the visibility check. jQuery uses this check for visibility: 
``` javascript
!!( elem.offsetWidth || elem.offsetHeight || elem.getClientRects().length );
```
Their solution considers visible to be if the element takes space on the page. They don't check visibility or opacity attributes because those don't change the layout. This PR uses the jquery definition of visibility and adds a check to the visibility attribute, both check all ancestors.

I left the visibility check in there because it was already there, but I'm impartial.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -928,8 +928,8 @@ Shortcut for [page.mainFrame().waitForFunction(pageFunction[, options[, ...args]
 #### page.waitForSelector(selector[, options])
 - `selector` <[string]> A [selector] of an element to wait for,
 - `options` <[Object]> Optional waiting parameters
-  - `visible` <[boolean]> wait for element to be present in DOM and to be visible, i.e. to not have `display: none` or `visibility: hidden` CSS properties or have an ancestor with those properties. Defaults to `false`.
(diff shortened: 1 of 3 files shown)


ChristianDavis:
Eh... is this a typescript error? an HTMLElement inherits from Element...

docs/api.md
```diff
@@ -929,6 +929,7 @@ Shortcut for [page.mainFrame().waitForFunction(pageFunction[, options[, ...args]
 - `selector` <[string]> A [selector] of an element to wait for,
 - `options` <[Object]> Optional waiting parameters
   - `visible` <[boolean]> wait for element to be present in DOM and to be visible, i.e. to not have `display: none` or `visibility: hidden` CSS properties. Defaults to `false`.
+  - `hidden` <[boolean]> wait for element to not be found in the DOM or to be hidden, i.e. have `display: none` or `visibility: hidden` CSS properties. Defaults to `false`.
```
vsemozhetbyt, line 932: It seems the same should be added in `frame.waitForSelector(selector[, options])` description.

ChristianDavis: that's part of  PR #967 that's still open. I should have wait to open this one. I make bad decisions late at night, sorry about that

PR #1032, opened 2017-10-13
feat(Page): implement page.setOfflineMode

### LINKED ISSUE #63: Emulate offline
(taken from https://github.com/gr2m/headless-chrome-test/issues/1)

Puppeteer should have an offline mode to test PWA's. For example:

```js
page.setOfflineMode(true);
```

Emulation of other network conditions might be useful as well, but we can wait with them until there are good scenarios.

aslushnikov:
This patch implements page.setOfflineMode method.

Fixes #63.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -328,6 +328,25 @@ describe('Page', function() {
     }));
   });
 
+  describe('Page.setOfflineMode', function() {
+    it('should work', SX(async function() {
+      await page.setOfflineMode(true);
+      let error = null;
+      await page.goto(EMPTY_PAGE).catch(e => error = e);
+      expect(error).toBeTruthy();
+      await page.setOfflineMode(false);
+      const response = await page.reload();
+      expect(response.status).toBe(200);
+    }));
+    it('should emulate navigator.onLine', SX(async function() {
+      expect(await page.evaluate(() => window.navigator.onLine)).toBe(true);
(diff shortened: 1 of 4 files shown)


test/test.js
```diff
@@ -328,6 +328,25 @@ describe('Page', function() {
     }));
   });
 
+  describe('Page.setOfflineMode', function() {
+    it('should work', SX(async function() {
+      await page.setOfflineMode(true);
+      let error = null;
+      await page.goto(EMPTY_PAGE).catch(e => error = e);
+      expect(error).toBeTruthy();
```
JoelEinbinder, line 336: Can we check error message here? Also what is the error here? Timeout?

aslushnikov: @JoelEinbinder it's currently "Failed to navigate!", since we don't report errors for now. It won't be a timeout though: it gonna be ERR_INTERNET_DISCONNECTED

JoelEinbinder: Ok good for now I guess. Would be nice to report a more descriptive error here. Also how do I automate the dino game? :D

JoelEinbinder (approved):

test/test.js
aslushnikov: > Would be nice to report a more descriptive error here

Yes, hopefully it would be easier to do with plznavigate enabled.

PR #1035, opened 2017-10-14
📝 Keyboard.press to  page.press

lxp521125:
Keyboard.press to  page.press
up the doc


lxp521125:
the api has many vension. I saw it .

PR #1037, opened 2017-10-15
copy hidden option

ChristianDavis:
hidden option should be in page.waitForSelector as well as frame.waitForSelector


aslushnikov (approved):
Thanks!

PR #1038, opened 2017-10-15
fix: page.close() throws an error if the connection is closed

ChristianDavis:
Page.close throws an ambiguous error if the connection is already closed 
```javascript
await page.close();
await page.close();
```
I think the assumption of the function is that the connection is open, so it should throw. But, often close can be called many times without consequences. Node http server for instance does nothing when close is called on a closed server.


lodenrogue (approved):

lib/Connection.js
```diff
@@ -201,6 +201,8 @@ class Session extends EventEmitter {
   }
 
   async dispose() {
+    if (!this._connection)
```
aslushnikov: Let's `console.assert` instead.

ChristianDavis: instead of the error? that sounds reasonable

aslushnikov (approved):

PR #1048, opened 2017-10-16
Update README.md

nexus6psoftwaredeveloper:


aslushnikov:
@nexus6psoftwaredeveloper Why do anything? I like it as it is now.

PR #1055, opened 2017-10-17
Add a failing test for storing cookies

JoelEinbinder:


aslushnikov (approved):

vsemozhetbyt:
FWIW, the issue is not only storing but also reading (and sending), [In this examples](https://github.com/GoogleChrome/puppeteer/issues/921), cookies are stored with `headless: false` mode but are not required by `headless: true` mode.

aslushnikov:
@vsemozhetbyt the headless and headful modes maintain a different directory structure; we'll probably need to highlight this somewhere in the docs/api.

vsemozhetbyt:
Well, this can be a problem. How can users log in to sites to use them headlessly? Should they replace this directory structure manually?

aslushnikov:
> Well, this can be a problem. 

Yes, this is very uncomfortable. Filed upstream [crbug.com/775703](https://crbug.com/775703)

fatmalama:
that crbug was closed as fixed yet it doesn't seem like it's maintaining the same structure
what happened? regression?

PR #1063, opened 2017-10-17
feat(chromium): roll chromium to r509368

JoelEinbinder:
This roll gets us https://crrev.com/509347 which will save cookies to disk.


JoelEinbinder:
Looks like the cookies tests might be a bit flaky.

aslushnikov (approved):

Issue #1020, opened 2017-10-12
Request Interception Enhancements


aslushnikov:
Request interception should allow:
- abort request with different errors
- fake the response: headers, status, body.

sukrosono:
That was nice feature to have, can someone tell me what M1 stand for?

aslushnikov:
@brutalcrozt M1 is the next major milestone.

entrptaher:
```
const puppeteer = require('puppeteer');

puppeteer.launch().then(async browser => {
  const page = await browser.newPage({headless: false});
  await page.setRequestInterceptionEnabled(true);
  page.on('request', interceptedRequest => {
    if (interceptedRequest.url.endsWith('.png') || interceptedRequest.url.endsWith('.jpg'))
      interceptedRequest.abort();
    else
      interceptedRequest.continue();
  });
  await page.goto('https://example.com');
  await page.screenshot('test.png');
  // await browser.close();
});
```

This hangs the node process on windows.
Nothing happens if I use setRequestInterceptionEnabled.

aslushnikov:
@entrptaher Doesn't hang for me when I uncomment the `await browser.close()`. Anyhow, it has nothing to do with this issue; please, file another one.

This, btw, is fixed.

entrptaher:
@aslushnikov I did create an issue, https://github.com/GoogleChrome/puppeteer/issues/1110

PR #1064, opened 2017-10-17
feat(interception): Implement request.mockResponse method

aslushnikov:
This patch implements a new `Request.mockResponse` method. This
allows users to fulfill the intercepted request with a hand-crafted
response if they wish so.

References #1020.


lib/NetworkManager.js
```diff
@@ -327,6 +327,46 @@ class Request {
     });
   }
 
+  async mockResponse(response = {}) {
+    // DataURL's are not interceptable. In this case, do nothing.
+    if (this.url.startsWith('data:'))
```
JoelEinbinder: Should we throw here instead of failing silently?

aslushnikov: Good call.

aslushnikov: Left it as-is for the following code to work:

```js
page.on('request', request => request.respond({
  status: 404,
  body: Buffer.from('<div>Not Found!</div>')
});
```

lib/NetworkManager.js
```diff
@@ -341,6 +341,49 @@ class Request {
...
+    const statusCode = response.status || 200;
+    const statusText = statusTexts[statusCode] || '';
+    responseLines.push(`HTTP/1.1 ${statusCode} ${statusText}`);
+    const headers = {};
+    if (response.headers) {
+      for (const header of Object.keys(response.headers))
+        headers[header.toLowerCase()] = response.headers[header];
+    }
+    if (response.body.length && !('content-length' in headers))
+      headers['content-length'] = response.body.length;
```
paambaati: @aslushnikov Do you want to perhaps use `Buffer.byteLength(body)` instead of `body.length`?

aslushnikov: @paambaati this is reasonable, thanks.

test/test.js
```diff
@@ -1339,6 +1339,44 @@ describe('Page', function() {
...
+    it('should allow mocking binary responses', SX(async function() {
+      await page.setRequestInterceptionEnabled(true);
+      const imageBuffer = fs.readFileSync(path.join(__dirname, 'assets', 'pptr.png'));
+      page.on('request', request => {
+        request.respond({
+          contentType: 'image/png',
+          body: imageBuffer
+        });
+      });
+      await page.goto(EMPTY_PAGE);
```
JoelEinbinder: I am confused what this test is doing. You are loading the image, and intercepting it with the same exact image? And won't the initial navigation also be overridden as the image?

I would expect there to be two images. And then bad.png is overridden with good.png.

lib/NetworkManager.js
```diff
@@ -341,6 +341,49 @@ class Request {
   }
 
   /**
+   * @param {!Object} response
+   */
+  async respond(response = {}) {
+    // DataURL's are not interceptable. In this case, do nothing.
```
JoelEinbinder: Let's put a note in the API.

docs/api.md
```diff
@@ -1897,6 +1899,32 @@ Contains the request's post body, if any.
...
+
+An example of fulfilling all requests with 404 responses:
+
+```js
+await page.setRequestInterceptionEnabled(true);
+page.on('request', request => {
+  request.respond({
+    status: 404,
+    contentType: 'text/plain',
+    body: 'Not Found!'
```
JoelEinbinder, line 1927: Body is a buffer, but the example uses a string. Should the type on body be Buffer|string?

docs/api.md
```diff
@@ -1897,6 +1899,32 @@ Contains the request's post body, if any.
...
+#### request.respond(response)
+- `response` <[Object]> Response that will fulfill this request
+  - `status` <[number]> Response status code, defaults to `200`.
+  - `headers` <[Object]> Optional response headers
+  - `contentType` <[string]> If set, equals to setting `Content-Type` response header
+  - `body` <[Buffer]> Optional response body
+- returns: <[Promise]>
+
+Fulfills request with given response. To use this, request interception should
+be enabled with `page.setRequestInterceptionEnabled`. Exception is immediately
```
JoelEinbinder: An exception is thrown if request interception is not enabled.

test/test.js
```diff
@@ -1339,6 +1339,44 @@ describe('Page', function() {
...
+      await page.evaluate(() => {
+        const img = document.createElement('img');
+        img.src = '/pptr.png';
+        document.body.appendChild(img);
+        return new Promise(fulfill => img.onload = fulfill);
+      });
+      const img = await page.$('img');
+      expect(await img.screenshot()).toBeGolden('mock-binary-response.png');
+    }));
+  });
```
JoelEinbinder: Add a test to throw if interception is not enabled or if trying to respond twice to the same request?

lib/NetworkManager.js
```diff
@@ -341,6 +341,49 @@ class Request {
   }
 
   /**
+   * @param {!Object} response
```
JoelEinbinder: I'd feel safer if the options were typed here, considering we are getting a buffer.

docs/api.md
aslushnikov: Done

docs/api.md
aslushnikov: Done.

lib/NetworkManager.js

test/test.js
aslushnikov: This is a funny one :) Fixed.

test/test.js

lib/NetworkManager.js

docs/api.md
```diff
@@ -994,6 +996,9 @@ puppeteer.launch().then(async browser => {
 });
 ```
 
+> **NOTE** Request interception doesn't work with data URLs. Calling `abort`,
+> `continue` or `respond` on requests for data URLs is a noop.
```
JoelEinbinder, line 1000: no-op

JoelEinbinder (approved):

docs/api.md
aslushnikov: Found "noop" in google, it's apparently a thing as well :)

PR #1065, opened 2017-10-17
feat(Request): expose request failure reason

### LINKED ISSUE #901: Provide reason for failed requests
The 'requestfailed' event is similar to PhantomJS' onResourceError event, but in Puppeteer the event handler receives only a Request object. There should be a way for a requestfailed handler to determine what the cause of the failure was - for example, a timeout vs an unresolvable hostname.

aslushnikov:
Similarly to the `request.response()` method, this patch adds
`request.failure()` method that returns error details for the failed
requests.

Fixes #901.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1798,7 +1798,7 @@ Exception is immediately thrown if the request interception is not enabled.
 
 #### request.failure()
 - returns: <[Object]> Object describing request failure, if any
-  - `errorText` <[string]> Human-readable error message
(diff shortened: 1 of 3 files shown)


JoelEinbinder (approved):
Can we add an example error message to the docs?

ali-habibzadeh:
```
const onRequestFailedHandler = (request) => {
        sets.failures.push(request);
        console.log('failure');
        console.log(request.response());
        console.log(request.failure());
    };
```

TypeError: request.failure is not a function

ali-habibzadeh:
When will this available on GoogleChrome/puppeteer ?

PR #1066, opened 2017-10-17
fix(examples): add missing argument

sashiyama:
I tried to do it as follows, but I encountered an error.

```
$ node examples/search.js
(node:95394) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: Cannot read property 'type' of null
(node:95394) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

page.type requires selector of argument.


aslushnikov (approved):
Thanks!

Issue #320, opened 2017-08-17
Is it possible to start and capture code coverage data?


craftand:
Is there a way to enable code coverage and access the data it generates?

https://developers.google.com/web/updates/2017/04/devtools-release-notes#coverage

aslushnikov:
Not yet. There's a #31 which would allow this with [devtools protocol](https://github.com/ChromeDevTools/devtools-protocol).

joelgriffith:
DevTools protocol exposes a range of coverage data, which isn't immediately useful from a high level (What percent of the code am I using?). In Navalia, I had to reduce this array of arranges and build out the metrics myself https://github.com/joelgriffith/navalia/blob/master/src/Chrome.ts#L977. Note this is for JS coverage as CSS coverage has a different API and small semantic differences.

You also have to inform Chrome that you want to collect coverage prior to any sort of navigating/interaction. So I think there's two things involved if we want to make this a first-class API:

- Have an API for toggling coverage capturing.
- An API for generating some higher-level data for coverage info.

craftand:
@joelgriffith I did a small poc for last night and was straight forward to add it in as a _coverage_ object. The data returned is a lot to go through and organize. I was just looking at it from a JS coverage not looking at CSS atm. 

Since the coverage relies on the Profiler being active I would guess it may make sense to expose a Profiler object and start coverage from that to round out an api and make it nicer to use. If I can get some time I'd like to work on it and submit a PR.

abhishek1nair:
@craftand Are you still working on it? It would be great if this feature was available. I'm currently using puppeteer for automation test cases and coverage data would be essential to define completeness.

craftand:
Yes, @abhisek1nair, I will be working on it

joelgriffith:
PST: https://github.com/joelgriffith/navalia/blob/master/src/Chrome.ts#L918 FWIW

ldd:
so, I am just confused and I'd love to have some clarification.

At some point in the future, I'd love to get code coverage for my tests. I am unable to use Istanbul, or something similar, and I guess I am not the only one.

So far, it seems that [navalia](https://github.com/joelgriffith/navalia) supports my use case, but it is (allegedly) being [deprecated](https://github.com/joelgriffith/navalia/issues/65) in favour of puppeteer). 

Puppeteer is awesome, but, as per bug #31, the raw devtools protocol is not being exposed in the API.

[...]

PR #1067, opened 2017-10-17
feat(coverage): Support Node.js coverage instrumentation.

mikeal:
This patch detects coverage and inserts proxy objects to handle coverage on instrumented evaluate calls.

This patch works great, I used a bunch of puppeteer tests I have locally to verify, but I'm not entirely sure how to write a unit test for puppeteer itself for it. 

Additionally, the typescript linter is failing on globals detection and I have no idea how to suppress them.

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -25,6 +25,7 @@ const {FrameManager} = require('./FrameManager');
 const {Keyboard, Mouse, Touchscreen} = require('./Input');
 const Tracing = require('./Tracing');
 const {helper, debugError} = require('./helper');
+const {createHash} = require('crypto');
 
 const writeFileAsync = helper.promisify(fs.writeFile);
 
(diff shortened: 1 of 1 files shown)


mikeal:
I'll explain a bit more about why this patch is necessary so that people don't have to read this entire thread https://github.com/GoogleChrome/puppeteer/issues/1054

`page.evaluate(() => {})` will .`toString()` the function and eval it in the page's context.

When coverage is enabled in Node.js it hooks into require and instruments every line of code. This means that the `.toString()` value has *partial* instrumentation and will fail with `cv_randomid` is undefined errors.

This patch detects coverage and inserts proxy objects for the instrumentation objects that communicate back to the Node.js process which lines have been called.

mikeal:
I've got a tool, along with tests that enable coverage, https://github.com/mikeal/cappadonna , using this branch.

bcoe:
this is awesome, let me know if there's anything I can do to help see this over the finish line. I'm excited to add documentation for this to Istanbul, perhaps we can also call this out in puppeteer's docs somewhere?

aslushnikov:
Hey @mikeal, thanks for the effort of fixing this.

There's a bunch of missing pieces, like evaluations in child frames and page navigations, but these are all fixable. There's also a little bit of race conditions related to async nature exposed functions. I, however, have a stronger concern about this whole approach.

The logic in this PR is very much istanbul-related. Puppeteer, however, shouldn't depend on istanbul, nor should it have any code that relies on internal istanbul machinery, such as `__coverage__` global.

I see that this PR solves two problems:
1. It makes istanbul to work with puppeteer
2. It teaches istanbul to compute code coverage inside browser

The (2) should be solved separately: there's a superior approach for the browser code coverage (see #320). 

The core issue with (1) is that istanbul's code preprocessing yields side-effects: the function's toString method return different values with and without coverage.

A simple fix would be to change istanbul to override `Function.prototype.toString` with something like this:
```js
(() => {
  const originalToString = Function.prototype.toString;
  Function.prototype.toString = () => {
    const code = originalToString.call(this);
    return code.replace(/\bcov_[^;]+/g, '');
  }
})();
```

mikeal:
A few things:

* Istanbul is really the defacto-standard for JS based coverage. There isn't much else that is in wide use.
* It's a fairly standard practice among test tools and frameworks in JS to dynamically enable coverage features based on `__coverage__`, you can find code in `tap` for instance that does just that.

This all said, I can just as easily stick this code in `cappadonna` instead of `puppeteer`. I figured I should send it as a PR so that you can stop getting strange issue reports when people enable coverage :)

> A simple fix would be to change istanbul to override Function.prototype.toString with something like this:

If that happened then we wouldn't be able to hack in support the way we just did ;)

It's great that you have a way to collect better coverage statistics in Chrome but for most practical use cases we're going to need a unified coverage report across Node.js and browser code. Tap jumps through a lot of hoops to capture child processes and whatnot without any user configuration or intervention.

In the end a Chrome-centric solution just isn't very useful if we have to manually integrate it into the rest of an ecosystem that already "just works."

PR #1080, opened 2017-10-18
feat(Request): allow aborting intercepted requests with custom reasons

aslushnikov:
This patch adds optional parameter to the `request.abort()` method
that specifies abort reason.

References #1020.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -162,7 +162,7 @@
   * [elementHandle.type(text[, options])](#elementhandletypetext-options)
   * [elementHandle.uploadFile(...filePaths)](#elementhandleuploadfilefilepaths)
 - [class: Request](#class-request)
-  * [request.abort()](#requestabort)
+  * [request.abort([errorCode])](#requestaborterrorcode)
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -1178,6 +1178,17 @@ describe('Page', function() {
       expect(response.ok).toBe(true);
```
JoelEinbinder, line 1178: Can we check the error text in this test? just to be safe

JoelEinbinder (approved):

docs/api.md
```diff
@@ -1831,7 +1831,22 @@ If request fails at some point, then instead of 'requestfinished' event (and pos
 
 If request gets a 'redirect' response, the request is successfully finished with the 'requestfinished' event, and a new request is  issued to a redirected url.
 
-#### request.abort()
+#### request.abort(errorCode)
```
JoelEinbinder: [errorCode], because it is optional

test/test.js
aslushnikov: Done.

docs/api.md

PR #1095, opened 2017-10-19
Doc(example) update search.js

sukrosono:
This patch using new API `Keyboard class`.

### DIFF
--- a/examples/search.js
+++ b/examples/search.js
@@ -23,10 +23,9 @@ const puppeteer = require('puppeteer');
 const browser = await puppeteer.launch();
 const page = await browser.newPage();
 await page.goto('https://google.com', {waitUntil: 'networkidle'});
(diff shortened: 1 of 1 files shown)


examples/search.js
```diff
@@ -23,10 +23,9 @@ const puppeteer = require('puppeteer');
 const browser = await puppeteer.launch();
 const page = await browser.newPage();
 await page.goto('https://google.com', {waitUntil: 'networkidle'});
-// Type our query into the search bar
-await page.type('input[name=q]', 'puppeteer');
 
 await page.click('input[type="submit"]');
+await page.keyboard.type('puppeteer');
```
aslushnikov, line 28: Why typing after clicking the "submit" button?

aslushnikov (commented):
I'm not convinced this is better than what we have today.

PR #1098, opened 2017-10-19
fix(page.evaluate): jsonValue should ignore toJSON property

### LINKED ISSUE #1003: Strange breaking change in page.evaluate()
```js
'use strict';

(async function main() {
  try {

    await page.goto('https://example.org/');
    const result1 = await page.evaluate(() => []);
    console.log(typeof result1);

    await page.goto('https://developer.apple.com/library/content/documentation/Darwin/Conceptual/FSEvents_ProgGuide/Introduction/Introduction.html');
    const result2 = await page.evaluate(() => []);
    console.log(typeof result2);

    await browser.close();
  } catch (err) {
    console.error(err);
  }
})();
```
puppeteer v0.11.0:
```console
object
object
```
puppeteer v0.12.0-alpha (current master)
```console
object
string
```

aslushnikov:
Currently, JSHandle.jsonValue() is implemented as in-page JSON.stringify
call and consequent JSON.parse in node. This approach proved to be
unfortunate for automation purposes: if page author overrode the
Object.prototype.toJSON method, then it's harder for puppeteer to
interact with the page.

This patch switches JSHandle.jsonValue to use protocol serialization
that ignores toJSON property. THis also changes the `page.evaluate`
behavior since it is based on JSHandle.jsonValue().

Fixes #1003.

BREAKING CHANGE:
`page.evaluate` no longer calls toJSON when generating return value.
For the old behavior, do JSON.parse/JSON.stringify manually:

```js
const json = JSON.parse(await page.evaluate(() => JSON.stringify(obj)));
```


docs/api.md
```diff
@@ -1677,7 +1677,10 @@ Fetches a single property from the referenced object.
 #### jsHandle.jsonValue()
 - returns: <[Promise]<[Object]>>
 
-Returns a JSON representation of the object. The JSON is generated by running [`JSON.stringify`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) on the object in page and consequent [`JSON.parse`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) in puppeteer.
+Returns a JSON representation of the object. If the method has a
```
vsemozhetbyt: If the _object_ has a `toJSON` _method_?

JoelEinbinder (approved):

Issue #681, opened 2017-09-05
Having trouble getting typing/pressing to work


Madd0g:
I'm not sure what I'm doing wrong, I even tried to add logging in `up()` in `lib/Input.js` to see if my commands end up in there (they do).

I tried using `page.keyboard.down/page.keyboard.up` and `page.press`, while having an input/textarea element focused. But nothing got typed in the input.

I used something like: `page.press("keyA")`

Is there a special trick for getting these to work? Anything I can do to debug?

Thanks

JoelEinbinder:
did you try `page.type("some text")`?

Madd0g:
I saw it, but haven't tried that or `sendCharacter` because I'm working from keycodes.

I tried it now and it _does work_.

Madd0g:
why close? `page.type()` is not what I need.

JoelEinbinder:
`page.type()` sends the proper keycodes. If you aren't trying to do something fancy, `page.type("a")` is probably what you want. See https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-keyboard

`"keyA"` is not a valid key, you want just `"A"`.

Madd0g:
I actually am trying to do something fancy, like holding keys :)
I thought I was supposed to send the `"code"` you get from a keyboard event (MetaLeft, keyA, keyS).

I tried sending just `"Meta"` and `"a"` instead, tried with `press/up/down` and none of them work for me. When I accidentally passed null in there, I got an error, so supposedly the values I'm sending are getting somewhere without error.

aslushnikov:
@Madd0g if you want to down "A", do

```js
page.keyboard.down('A');
```

If you want to down "meta", do
```js
page.keyboard.down('Meta');
```

Same for `up`/`press`. Is this what you need?
We also have a lot of [tests](https://github.com/GoogleChrome/puppeteer/blob/51a507b2361bed942a9cc1aa384c9350cc834286/test/test.js#L1292) to check that keyboard works as expected. Tests run many different scenarios, you might find some of them useful.

Also, the full list of keys could be found on MDN: [key values](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values).

Madd0g:
Thank you for your reply.

From reading the tests, it seems I have misunderstood `press(key, [options])`, I thought it was a combination of down/up (down holds a key and up releases it which causes a character to appear), but it only actually types if `text` is passed as part of the options?

[...]

Issue #777, opened 2017-09-14
Keyboard emulation is not perfect


aslushnikov:
Keyboard emulation is not perfect
- [x] [`event.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) is not reported
- [x] [`event.location`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/location) is not reported

aslushnikov:
For the record: event.location is merged upstream as [crrev.com/509990](https://chromium-review.googlesource.com/c/chromium/src/+/714280)

aslushnikov:
This is fixed now.

PR #1116, opened 2017-10-20
feat(keyboard): Accept codes

JoelEinbinder:
This patch lets key names be [code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) in addition to [key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key). When specifying a code, the proper text is generated assuming a standard US keyboard layout. e.g `Digit5` -> `"5"` or `"%"` depending on `Shift`. 

[location](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/location) is now specified.

Using unknown key names now throws an error.

Typing newlines now correctly presses enter.


lib/Input.js
```diff
@@ -30,23 +40,25 @@ class Keyboard {
...
-  async down(key, options = {text: undefined}) {
-    let { text } = options;
-    // If the key is a single character, and no modifiers are pressed except shift, a keypress should be generated by default.
-    if (text === undefined)
-      text = (key.length === 1 && !(this._modifiers & ~8)) ? key : '';
-    const autoRepeat = this._pressedKeys.has(key);
-    this._pressedKeys.add(key);
-    this._modifiers |= this._modifierBit(key);
+  async down(key, options = { text: undefined }) {
+    const description = this._keyDescriptionForString(key);
```
aslushnikov, line 44: this should throw if there's no description for the key

lib/Input.js
```diff
@@ -67,17 +79,71 @@ class Keyboard {
...
+    return description;
+  }
+
+  /**
    * @param {string} key
    */
   async up(key) {
-    this._modifiers &= ~this._modifierBit(key);
-    this._pressedKeys.delete(key);
+    const description = this._keyDescriptionForString(key);
```
aslushnikov, line 133: this throws if there's no key description as well

aslushnikov (approved):

lib/Input.js
JoelEinbinder: Done.

lib/Input.js
JoelEinbinder: Done.

aslushnikov:
let's at least reference bugs this is about.

JoelEinbinder:
This should help/fix
#681 
#777
#723

PR #1120, opened 2017-10-21
Enable pipe connection while in the appMode.

pavelfeldman:


aslushnikov:
Closed in favor of #2032: the same code with proper typing + tests

PR #1136, opened 2017-10-23
Add the ability to end navigation using a console message

IAmJulianAcosta:
This PR allows puppeteer to set a special log message that will determine when the navigation is done. Is particularly useful when you need to load a page that takes a long time to render and network activity has ended long time before page is ready.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -764,8 +764,10 @@ If there's no element matching `selector`, the method throws an error.
   - `waitUntil` <[string]> When to consider a navigation finished, defaults to `load`. Can be either:
     - `load` - consider navigation to be finished when the `load` event is fired.
     - `networkidle` - consider navigation to be finished when the network activity stays "idle" for at least `networkIdleTimeout` ms.
+    - `consolemessage` - consider navigation to be finished when a specific message is received on console, defined by `consoleMessage`.
   - `networkIdleInflight` <[number]> Maximum amount of inflight requests which are considered "idle". Takes effect only with `waitUntil: 'networkidle'` parameter.
   - `networkIdleTimeout` <[number]> A timeout to wait before completing navigation. Takes effect only with `waitUntil: 'networkidle'` parameter.
+  - `consoleMessage` <[string]> Console message that will end navigation. Takes effect only with `waitUntil: 'consolemessage'` parameter. Default value: `'puppeteerClosePage'`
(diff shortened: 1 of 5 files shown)


IAmJulianAcosta:
I signed it!

aslushnikov:
@IAmJulianAcosta What's your usecase? If you need a way for a page to notify puppeteer that it has loaded, you can can do this already with `exposeFunction`.

IAmJulianAcosta:
@aslushnikov I'm rendering a heavy 3D application, that will take several seconds to render on screen after data was loaded and `onLoad` element, so `goTo` ends before I wanted.

I checked `exposeFunction` and I can use it to close the browser, but this could be a nice functionality, so we can use `async`/`await` and don't have to wrap `browser.close()` to keep it simple

aslushnikov:
@IAmJulianAcosta Let me suggest alternative approach.

In your tests, you currently emit the 'customConsoleMessage' log. Instead, I suggest you to use exposed function, e.g. like this:

```js
async function main() {
  setTimeout (() => {
    if (window.pptrNotifyLoaded)
      window.pptrNotifyLoaded();
  }, 5000);
}
```

In puppeteer itself, you can set up the function and use it later like this:

```js
// Setup a notifyLoaded callback
await page.exposeFunction('pptrNotifyLoaded', () => page.emit('myPageHasLoaded'));

// Navigate and wait for the event to fire
await Promise.all([
  page.goto('https://example.com'),
  new Promise(fulfill => page.once('myPageHasLoaded'));
]);
```

This is rough, but this approach could be evolved into a few helper functions that would prepare the page and navigate it onto your website.

IAmJulianAcosta:
@aslushnikov The example you provided works and is very useful, I'm leaving you the decision to accept or decline this PR, in my opinion is something that might be useful for other people, but there is a workaround as you know.

aslushnikov:
@IAmJulianAcosta there's even a better way: you can use `console.timeStamp("load")` in your application and subscribe to `'metrics'` event in puppeteer:

The page code:
```js
async function main() {
  setTimeout (() => {
    console.timeStamp('load');
  }, 5000);
}
```

And in your application:
```js
page.goto('https://example.com');
page.on('metrics', (name, metrics) => {
  if (name !== 'load')
    return;
  // ... do the rest
});
```

The `console.timeStamp` is nice because it is also helpful for timeline. Additionally, there's no nasty if `if (window.pptrNotifyLoaded)`.

Regarding the PR, I'll close it for now. If some approach of notifying controller script becomes widely accepted, we'll include it in puppeteer by default.

Issue #728, opened 2017-09-09
Page.setContent should wait for resources to be loaded


aslushnikov:
(as mentioned in #486 and other places)

We need a way to wait for page to load all the resources after the `page.setContent`.

The [lifecycle events](https://aslushnikov.github.io/vanilla-protocol-viewer/#Page.lifecycleEvent) might help help.

aslushnikov:
Meanwhile. a good workaround for `page.setContent` that waits for all the resources to load:

```js
await page.goto(`data:text/html,${html}`, { waitUntil: 'networkidle0' });
```

cmazakas:
It's amazing that you posted the workaround. I'm running exactly into this issue, attempting to use `puppeteer` as a PDF generating service from HTML.

Thank you for filing a formal issue.

murilozilli:
I'm having trouble with this too on version 0.13-alpha with browser configs as:
```
"waitUntil": "networkidle2",
"timeout": 60000
```

HanXHX:
I confirm the same trouble.

Padam87:
with the latest release this hangs too:

```js
await page.goto(`data:text/html,${html}`, { waitUntil: 'networkidle' });
```

wait for `networkidle0` instead

```js
await page.goto(`data:text/html,${html}`, { waitUntil: 'networkidle0' });
```

HanXHX:
Hey @aslushnikov 

You said in #1312 to wait for https://chromium-review.googlesource.com/c/chromium/src/+/747805
The patch is merged...

Everything is OK to solve this issue?

Cheers!

aslushnikov:
@HanXHX this requires more work upstream: in order to reuse lifecycle events, `page.setContent` should initiate a navigation, which in turn should be plumbed through browser-side navigation aka "plznavigate".

knownasilya:
The workaround doesn't seem to work for me. Using 1.0.

cmazakas:
For reference,
```
      const loaded = page.waitForNavigation({
        waitUntil: 'load',
      });

      await page.setContent(html);
      await loaded;
```
seems to work for me.

knownasilya:
My use case is that background-color css isn't loading. I'll give the above a try as well.

knownasilya:
@LeonineKing1199  I got navigation exceeded timeout using the above method.

cmazakas:
Darn, it worked for me. Sorry it didn't for you :P

[...]

PR #1141, opened 2017-10-23
refactor: migrate NavigatorWatcher to lifecycle events

aslushnikov:
This patch:
- migrates navigation watcher to use protocol-issued lifecycle events.
- removes `networkIdleTimeout` and `networkIdleInflight` options for
  `page.goto` method
- adds a new `networkidle0` value to the waitUntil option of navigation
  methods

References #728.

BREAKING CHANGE:

As an implication of this new approach, the `networkIdleTimeout` and
`networkIdleInflight` options are no longer supported. Interested
clients should implement the behavior themselves using the `request` and
`response` events.

### DIFF
--- a/lib/NavigatorWatcher.js
+++ b/lib/NavigatorWatcher.js
@@ -26,13 +26,14 @@ class NavigatorWatcher {
   constructor(client, frameId, ignoreHTTPSErrors, options = {}) {
     console.assert(options.networkIdleTimeout === undefined, 'ERROR: networkIdleTimeout option is no longer supported.');
     console.assert(options.networkIdleInflight === undefined, 'ERROR: networkIdleInflight option is no longer supported.');
+    console.assert(options.waitUntil !== 'networkidle', 'ERROR: "networkidle" option is no longer supported. Use "networkidle2" instead');
     this._client = client;
     this._frameId = frameId;
     this._ignoreHTTPSErrors = ignoreHTTPSErrors;
     this._timeout = typeof options['timeout'] === 'number' ? options['timeout'] : 30000;
     const waitUntil = typeof options['waitUntil'] === 'string' ? options['waitUntil'] : 'load';
-    const possibleWaitValues = new Set(Object.values(protocolLifecycleToPuppeteer));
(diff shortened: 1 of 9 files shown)


lib/NavigatorWatcher.js
```diff
@@ -19,25 +19,27 @@ const {helper} = require('./helper');
...
     this._client = client;
+    this._frameId = frameId;
     this._ignoreHTTPSErrors = ignoreHTTPSErrors;
     this._timeout = typeof options['timeout'] === 'number' ? options['timeout'] : 30000;
-    this._idleTime = typeof options['networkIdleTimeout'] === 'number' ? options['networkIdleTimeout'] : 1000;
-    this._idleInflight = typeof options['networkIdleInflight'] === 'number' ? options['networkIdleInflight'] : 2;
-    this._waitUntil = typeof options['waitUntil'] === 'string' ? options['waitUntil'] : 'load';
-    console.assert(this._waitUntil === 'load' || this._waitUntil === 'networkidle', 'Unknown value for options.waitUntil: ' + this._waitUntil);
+    const waitUntil = typeof options['waitUntil'] === 'string' ? options['waitUntil'] : 'load';
+    const possibleWaitValues = new Set(Object.values(protocolLifecycleToPuppeteer));
```
JoelEinbinder: We don't have Object.values in node6.

test/test.js
```diff
@@ -899,6 +907,11 @@ describe('Page', function() {
       await page.goto(HTTPS_PREFIX + '/empty.html').catch(e => error = e);
       expect(error.message).toContain('SSL Certificate error');
     }));
+    it('should throw if networkidle is passed as an option', SX(async function() {
+      let error = null;
+      await page.goto(EMPTY_PAGE, {waitUntil: 'networkidle'}).catch(err => error = err);
+      expect(error.message).toContain('"networkidle" option is no longer supported');
```
JoelEinbinder, line 913: You made network idle allowed.

lib/NavigatorWatcher.js
aslushnikov: Thanks, done.

test/test.js
aslushnikov: Made it disallowed again :)

JoelEinbinder (approved):

PR #1147, opened 2017-10-24
feat(Page.goto): support multiple events in waitUntil option

### LINKED ISSUE #805: Waiting for both `load` and `networkIdle`
Some pages will fire load event before network goes idle, some after. I'd like to wait for both events to have happened before considering a page 'navigated'. Currently looking at the `watchdog` code it seems like it's either one or the other. The way to do it right now is:

```js
const timeout = 30000;
const networkIdleTimeout = 500;
await Promise.all([
  page.waitForNavigation({ timeout, waitUntil: 'load' }),
  page.waitForNavigation({ timeout, waitUntil: 'networkidle', networkIdleTimeout }),
  page.goto(url)
]);
```

But it seems to me that this is a functionality that deserves a more straightforward out-of-the-box solution. for instance:

```js
page.goto(url, { 
  timeout: 3000,
  waitUntil: [ 'load', 'networkidle' ],
  networkIdleTimeout: 500
});
```

Or assume that people that wait for `'networkidle'` also always want to wait for `'load'`.

PS: I might be wrong about this, but after digging through the code, it seems to me that this function doesn't take the actual state into account. For instance it won't work if the load event has already fired or if the network is already idle.

aslushnikov:
This patch adds support to multiple events that could be passed inside
navigation methods:
- Page.goto
- Page.waitForNavigation
- Page.goForward
- Page.goBack
- Page.reload

Fixes #805

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -761,7 +761,7 @@ If there's no element matching `selector`, the method throws an error.
 #### page.goBack(options)
 - `options` <[Object]> Navigation parameters which might have the following properties:
   - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds, pass `0` to disable timeout.
-  - `waitUntil` <[string]|[Array]<[string]>> When to consider navigation succeeded, defaults to `load`. The option allows to pass in an array of events to wait for. Events Can be either:
+  - `waitUntil` <[string]|[Array]<[string]>> When to consider navigation succeeded, defaults to `load`. Given an array of event strings, navigation is considered to be successful after all events have been fired. Events Can be either:
     - `load` - consider navigation to be finished when the `load` event is fired.
     - `domcontentloaded` - consider navigation to be finished when the `DOMContentLoaded` event is fired.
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -761,7 +761,7 @@ If there's no element matching `selector`, the method throws an error.
 #### page.goBack(options)
 - `options` <[Object]> Navigation parameters which might have the following properties:
   - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds, pass `0` to disable timeout.
-  - `waitUntil` <[string]> When to consider a navigation finished, defaults to `load`. Can be either:
+  - `waitUntil` <[string]|[Array]<[string]>> When to consider navigation succeeded, defaults to `load`. The option allows to pass in an array of events to wait for. Events Can be either:
```
JoelEinbinder: Given an array of event strings, navigation is considered to be successful after all events have been fired.

JoelEinbinder (approved):

PR #1148, opened 2017-10-24
requstWillBeSent not called for certain requests

durackpl:
Hi Everyone,

I came across this behavior: requstWillBeSent is not called and request object is not created for certain request. 

It seems that this issue is addressed already in _onLoadingFinished/onLoadingFailed function in NetworkManager.js.  However, a similar issue in _onRequestWillBeSent was not addressed.

Here a few lines patch fixing this.

Cheers,
D.


durackpl:
I've signed it

lib/NetworkManager.js
```diff
@@ -203,6 +203,14 @@ class NetworkManager extends EventEmitter {
     }
     if (event.redirectResponse) {
       const request = this._requestIdToRequest.get(event.requestId);
+
+      // Added (copy similar lines from below): For certain requestIds
+      // we never receive requestWillBeSent event.  @see
+      // https://crbug.com/750469
+      if (!request) {
```
aslushnikov, line 210: When and why does this happen? I tried a few scenarios, but neither worked for me. Network is complicated, I'm very interested in a test for this.

aslushnikov: For the record: one of my test attempts https://github.com/aslushnikov/puppeteer/commit/3146445bfd2deb2dabd9b3433e243af08dd33a0e

durackpl: The basic scenario when it happens: (1) click a link which opens a popup; (2) catch popup with 'targetcreated'; (3) click a link on popup; (4) another popup comes up and is caught again through 'targetcreated'; (5) while this second popup is getting loaded, a redirect is received/followed and that is when it happens.

I cannot disclose/inspect internals of the actual webapp where this is observed and I failed to reproduce similar behaviour with simple examples.

Feel free to guide me and suggest further tests I can try on the actual webapp to identify the problem closer.

aslushnikov: @durackpl do you have any interstitial page in the second pop-up?

durackpl: As a matter of fact, the first popup seems "interstitial".  It is the first time I see this term "interstitial", and according to wikipedia, the first popup fits "interstitial" definition :)

aslushnikov:
All of a sudden, this has been solved in https://github.com/GoogleChrome/puppeteer/pull/1420.

Thank you for the PR, though!

PR #1161, opened 2017-10-25
Add new buildpack url that adds additional fonts

jonknapp:
The jontewks buildpack worked great for running Chrome headless, but it did not include support for Chinese characters and the PDF rendering I was attempting was not usable. This fix brings in a few custom fonts to allow PDF generation to utilize the fonts.

The idea came from https://github.com/dscout/wkhtmltopdf-buildpack/pull/14/files


jonknapp:
I signed it!

aslushnikov (approved):

PR #1171, opened 2017-10-26
When printing pdf, default to printing the background

jbryson3:
Embarrassingly, I lost half a day to debugging why my backgrounds weren't appearing when printing pdfs.

Since chromium's --print-to-pdf feature defaults to printing the backgrounds, it would be good to match that default.

https://github.com/GoogleChrome/puppeteer/issues/301
https://bugs.chromium.org/p/chromium/issues/detail?id=771634


aslushnikov:
@jbryson3 On headful chrome the `printBackground` option defaults to `false`. Given that we already have a lot of clients for `page.pdf()` and that there's no clear winner, I'd keep everything as-is.

jbryson3:
I would think that there are more users choosing puppeteer for headless mode, and thus, would align with their defaults. But you're right, Chromium itself has different defaults based on mode. Thanks for your consideration @aslushnikov

jbryson3:
I signed it!

PR #1172, opened 2017-10-26
test: remove irrelevant test

aslushnikov:
Since the migration on lifecycle events, the websockets are no longer
taken into account to detect navigation success with networkidle events.


JoelEinbinder (commented):
Why wasn’t this failing? Are we losing functionality?

aslushnikov:
@JoelEinbinder: this is because websockets are no longer accounted for in networkidle events.

PR #1178, opened 2017-10-26
 feat: support Chrome 63

aslushnikov:
Puppeteer needs to be mostly functional with stable chrome. The first
stable chrome that will be supported will be Chrome 63.

This patch:
- adds backwards compatibility with Chrome 63 (currently, it's only the
  renamed request interception protocol call)
- whitelists a few tests to be ran only on Chrome 64+
- adds a new `unit-chrome-63` command that runs unit tests against
  Chrome M63
- adds a travis job to run tests against Chrome 63.

Over time, when Chrome 63 will be promoted to beta, we'll switch to
support Chrome Beta, and eventually Chrome Stable.


package.json
```diff
@@ -9,6 +9,7 @@
   },
   "scripts": {
     "unit": "jasmine test/test.js",
+    "unit-chrome-63": "node ./utils/downloader.js --download 508578 && CHROME=$(node ./utils/downloader.js --path 508578) npm run unit",
```
JoelEinbinder, line 12: Use cross env for windows support

JoelEinbinder (approved):

aslushnikov:
Closing this since we missed the chrome 63 target for now.

Bnaya:
Any idea when its going to be back?

aslushnikov:
@Bnaya we're aiming for M64 now.

PR #1181, opened 2017-10-27
update search example

andrew-apicello:
The search bar was not fully loaded and therefore "puppeteer" could not be entered into the submit field. 

await page.waitFor('input[name=q]');

was added to ensure this element is loaded before attempting to populate it.


andrew-apicello:
I signed it!

aslushnikov (approved):
Thanks, this is reasonable.

PR #1194, opened 2017-10-27
Fixes #1139 - Add ElementHandle.select method.

### LINKED ISSUE #1139: Missing ElementHandle.select; impossible (?) to make selection in a frame

austinkelleher:


aslushnikov:
@austinkelleher Thanks for the PR.

I'd rather have `Frame.select` then `ElementHandle.select`.
Also, there's in-flight #1099; let's land that first so that there are no conflicts.

alixaxel:
Isn't it redundant to have `selector` argument in `elementHandle.select` method?

```javascript
page.$('select#colors').select('select#colors', 'blue');
```

aslushnikov:
Closing in favor of  #1243

Issue #1195, opened 2017-10-27
Test `Page.goto should fail when navigating to bad SSL` is flaky


JoelEinbinder:
An error is always reported, but sometimes it reports `Failed to navigate: https://localhost:8908/empty.html` instead of `SSL Certificate error`

PR #1196, opened 2017-10-27
test: disable flaky ssl test

JoelEinbinder:
The test `Page.goto should fail when navigating to bad SSL` is flaky. See #1195


aslushnikov:
Closing in favor of #1237

PR #1197, opened 2017-10-27
Fix README.md runnable code to respect current API.

bo01ean:
Puppeteer README.md has options as second argument to page.goto, however, page.goto does not allow options.


bo01ean:
I signed it!

vsemozhetbyt:
[It seems it allows.](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagegotourl-options)

bo01ean:
I had an earlier node (7.1) which didn't have this capability for some reason. I upgraded and it works now :)

PR #1198, opened 2017-10-28
Fixes #1139 - Add Frame.select method 

ghost:
Added Frame.select method for issue #1139

PR #1200, opened 2017-10-28
Fixes #1139 frame.select method implemented

### LINKED ISSUE #1139: Missing ElementHandle.select; impossible (?) to make selection in a frame

suvratjain1995:


aslushnikov:
Closing this in favor of #1198

PR #1202, opened 2017-10-28
add evaluateOnNewDocument examples

### LINKED ISSUE #1201: Preloading a Script
This might not be an issue, more like a reference to how to mimic preload feature from nightmare/electron. 

If this looks good maybe we can include this in docs.
If this needs improvement, then please suggest.

If we want to change some default page variables before the page loads, we got no way of doing this in puppeteer. If we do, I don't see it in any docs or code.

For example, I might want to change navigator object properties and mock tests.

```js
// overwrite the `languages` property to use a custom getter
Object.defineProperty(navigator, 'languages', {
  get: function() {
    return ['en-US', 'en'];
  },
});

// overwrite the `plugins` property to use a custom getter
Object.defineProperty(navigator, 'plugins', {
  get: function() {
    // this just needs to have `length > 0`, but we could mock the plugins too
    return [1, 2, 3, 4, 5];
  },
});
```

It's easily doable with nightmare,

```js
var Nightmare = require('nightmare'),
  nightmare = Nightmare({
  webPreferences: {
        preload: 'alertMessage.js' // <-- this will get executed first, so we can change whatever object we want
      }
});
```

Note this is not addScriptTag or exposeFunction. exposeFunction will add a function, addScriptTag will do something on load. But, a preload script will load first, and make whatever changes we need to change.

I am testing this function right now, `evaluateOnNewDocument`. But defining a file which will load before every page load seems much more convenient.

Right now, I can do this.
```js
function preload() {
  // overwrite the `languages` property to use a custom getter
  Object.defineProperty(navigator, "languages", {
    get: function() {
      return ["en-US", "en", "bn", "github", "idontknowlanguage"];
    }
  });

  // overwrite the `plugins` property to use a custom getter
  Object.defineProperty(navigator, "plugins", {
    get: function() {
      // this just needs to have `length > 0`, but we could mock the plugins too
      return [1, 2, 3, 4, 5];
    }
  });
}

await page.evaluateOnNewDocument(preload);
```
Or like this in seperate file.
```js
// preload.js
module.exports = function(){
    // overwrite the `languages` property to use a custom getter
    Object.defineProperty(navigator, "languages", {
      get: function() {
        return ["en-US", "en", "bn", "c", "js"];
      }
    });

    // overwrite the `plugins` property to use a custom getter
    Object.defineProperty(navigator, "plugins", {
      get: function() {
        // this just needs to have `length > 0`, but we could mock the plugins too
        return [1, 2, 3, 4, 5];
      }
    });
}

// in out puppeteer
await page.evaluateOnNewDocument(require("./preload"));
```
Maybe we can include some preload scripts/list of files to preload in the `browser.newPage()`
for example, 

```js
browser.newPage({
 preload: "this.js"
})

// or something like this which will execute in order or async
browser.newPage({
 preload: ["this.js"],
 preloadAsync: true
})
```

entrptaher:
Fixes #1201

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -670,23 +670,22 @@ Adds a function which would be invoked in one of the following scenarios:
 
 The function is invoked after the document was created but before any of its scripts were run. This is useful to amend JavaScript environment, e.g. to seed `Math.random`.
 
-An example of using preloading some functions from an external file would be following,
+An example of overriding the navigator.languages property before the page loads:
 
 ```js
 // preload.js
-module.exports = function(){
-    console.log("Adding some fake languages to navigator property");
-    
+module.exports = function() {
     // overwrite the `languages` property to use a custom getter
     Object.defineProperty(navigator, "languages", {
       get: function() {
         return ["en-US", "en", "c", "js"];
-      }
+      };
     });
-}
+};
 
-// on the puppeteer script, assuming the preload.js file is in same folder of our script
-await page.evaluateOnNewDocument(require("./preload"));
+// In your puppeteer script, assuming the preload.js file is in same folder of our script
+const preloadFile = fs.readFileSync('filepath.js', 'utf8');
+await page.evaluateOnNewDocument(preloadFile);
 ```
 
 #### page.exposeFunction(name, puppeteerFunction)
--- a/docs/api.md
+++ b/docs/api.md
@@ -674,14 +674,13 @@ An example of overriding the navigator.languages property before the page loads:
 
 ```js
 // preload.js
-module.exports = function() {
-    // overwrite the `languages` property to use a custom getter
-    Object.defineProperty(navigator, "languages", {
-      get: function() {
-        return ["en-US", "en", "c", "js"];
-      };
-    });
-};
+
+// overwrite the `languages` property to use a custom getter
+Object.defineProperty(navigator, "languages", {
+  get: function() {
+    return ["en-US", "en", "c", "js"];
+  };
+});
 
 // In your puppeteer script, assuming the preload.js file is in same folder of our script
 const preloadFile = fs.readFileSync('./preload.js', 'utf8');
--- a/docs/api.md
+++ b/docs/api.md
@@ -678,7 +678,7 @@ An example of overriding the navigator.languages property before the page loads:
 // overwrite the `languages` property to use a custom getter
 Object.defineProperty(navigator, "languages", {
   get: function() {
-    return ["en-US", "en", "c", "js"];
+    return ["en-US", "en", "bn"];
   };
 });
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -684,7 +684,7 @@ module.exports = function() {
 };
 
 // In your puppeteer script, assuming the preload.js file is in same folder of our script
-const preloadFile = fs.readFileSync('filepath.js', 'utf8');
+const preloadFile = fs.readFileSync('preload.js', 'utf8');
 await page.evaluateOnNewDocument(preloadFile);
 ```
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -684,7 +684,7 @@ module.exports = function() {
 };
 
 // In your puppeteer script, assuming the preload.js file is in same folder of our script
-const preloadFile = fs.readFileSync('preload.js', 'utf8');
+const preloadFile = fs.readFileSync('./preload.js', 'utf8');
 await page.evaluateOnNewDocument(preloadFile);
 ```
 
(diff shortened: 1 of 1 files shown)


entrptaher:
I signed it!

docs/api.md
```diff
@@ -670,6 +670,25 @@ Adds a function which would be invoked in one of the following scenarios:
 
 The function is invoked after the document was created but before any of its scripts were run. This is useful to amend JavaScript environment, e.g. to seed `Math.random`.
 
+An example of using preloading some functions from an external file would be following,
```
ebidel: "An example of overriding the `navigator.languages` property before the page loads:"

docs/api.md
```diff
@@ -670,6 +670,25 @@ Adds a function which would be invoked in one of the following scenarios:
...
+
+```js
+// preload.js
+module.exports = function(){
```
ebidel: space: `function() {`

docs/api.md
```diff
@@ -670,6 +670,25 @@ Adds a function which would be invoked in one of the following scenarios:
...
+    console.log("Adding some fake languages to navigator property");
+    
+    // overwrite the `languages` property to use a custom getter
+    Object.defineProperty(navigator, "languages", {
+      get: function() {
+        return ["en-US", "en", "c", "js"];
+      }
+    });
+}
```
ebidel: `};`

docs/api.md
```diff
@@ -670,6 +670,25 @@ Adds a function which would be invoked in one of the following scenarios:
 
 The function is invoked after the document was created but before any of its scripts were run. This is useful to amend JavaScript environment, e.g. to seed `Math.random`.
 
+An example of using preloading some functions from an external file would be following,
+
+```js
+// preload.js
+module.exports = function(){
+    console.log("Adding some fake languages to navigator property");
```
ebidel: IMO, the `console.log` doesn't add much value. Please remove.

docs/api.md
```diff
@@ -670,6 +670,25 @@ Adds a function which would be invoked in one of the following scenarios:
...
+    
+    // overwrite the `languages` property to use a custom getter
+    Object.defineProperty(navigator, "languages", {
+      get: function() {
+        return ["en-US", "en", "c", "js"];
+      }
+    });
+}
+
+// on the puppeteer script, assuming the preload.js file is in same folder of our script
```
ebidel: just: `// In your puppeteer script.`

docs/api.md
```diff
@@ -670,6 +670,25 @@ Adds a function which would be invoked in one of the following scenarios:
...
+await page.evaluateOnNewDocument(require("./preload"));
```
ebidel: `"./preload.js"`

ebidel (changes requested):
Just some nits and comment updates to make things more clear.

docs/api.md
aslushnikov: I'd drop the `require` part: it adds unnecessary indirection.
It also assumes commonJS modules, which is opinionated.

Other then that, great example!

entrptaher: Is there any other easy way to preload an external file?

aslushnikov: @entrptaher sure, you can use `fs` for this: `fs.readFileSync('filepath.js', 'utf8')`. Eventually the es6 imports would work as well.

entrptaher: oh, that, I was wondering if you meant not to use any external file.
Let me fix these styles and things when I get time.

aslushnikov: @entrptaher I suggested not to bother with an external file since it has nothing to do with the API itself.

aslushnikov:
@entrptaher any updates on this?

docs/api.md
```diff
@@ -670,6 +670,24 @@ Adds a function which would be invoked in one of the following scenarios:
 
 The function is invoked after the document was created but before any of its scripts were run. This is useful to amend JavaScript environment, e.g. to seed `Math.random`.
 
+An example of overriding the navigator.languages property before the page loads:
+
+```js
+// preload.js
+module.exports = function() {
```
aslushnikov: if you use `readFile` method, then you shouldn't use `module.exports` part. You also don't need to define a wrapper function - the `readFile` returns a string.

So this file will simply contain the following:

```js
// preload.js  {
  get: function() {
    return ["en-US", "en", "c", "js"];
  };
});
```

aslushnikov (approved):

aslushnikov:
@entrptaher thank you.

@ebidel how do you like this PR?

JoelEinbinder (approved):
I like it

ebidel (approved):

entrptaher:
@aslushnikov do I have to do anything else? or it will merged later?

aslushnikov:
@entrptaher do I get it right that you signed the CLA, but the googlebot is stuck?

entrptaher:
@aslushnikov Seems so. I am sure I signed it.
Could it be that I used the web version to make pull request and thus it's stuck?

ebidel:
@aslushnikov this is ready to merge

PR #1226, opened 2017-10-31
Explain how to run examples

ithinkihaveacat:
The examples in the examples directory are not directly runnable--`node example.js` just fails on `require('puppeteer')`. Maybe there's a better way of doing this, but this worked for me.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,3 +1,12 @@
+# Running the examples
+
+(The follow assumes you have checked out the complete puppeteer repository, and installed puppeteer's dependencies via `yarn`.)
+
+```sh
+# creates example.png
+$ env NODE_PATH=../.. node screenshot.js
+```
+
 # Other resources
 
 > Other useful tools, articles, and projects that use Puppeteer.
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,10 +1,9 @@
 # Running the examples
 
-(The follow assumes you have checked out the complete puppeteer repository, and installed puppeteer's dependencies via `yarn`.)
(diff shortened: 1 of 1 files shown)


examples/README.md
```diff
@@ -1,3 +1,12 @@
+# Running the examples
+
+(The follow assumes you have checked out the complete puppeteer repository, and installed puppeteer's dependencies via `yarn`.)
```
ebidel: How about this:

Assuming you have a check out of this repo and have run `yarn` (or `npm i`) to install the dependencies, the examples can be run from the root folder like so:

```sh
NODE_PATH=../ node examples/search.js
```

ebidel (changes requested):

examples/README.md
ithinkihaveacat: Sure, changed! (Small tweaks to wording.)

aslushnikov (approved):

examples/README.md
```diff
@@ -1,3 +1,11 @@
+# Running the examples
+
+Assuming you have a checked out the puppeteer repo and run yarn (or `npm i`) to install the dependencies, the examples can be run from the root folder like so:
```
ebidel: Can you use this exact sentence (some grammatical fixes):

Assuming you have a checkout of the Puppeteer repo and have run `yarn` (or `npm i`) to install the dependencies, the examples can be run from the root folder like so:

ebidel (changes requested):

examples/README.md
ithinkihaveacat: Done.

ebidel (approved):

PR #1230, opened 2017-10-31
Add example of callPhantom callback usage

jbryson3:
I am migrating a codebase from phantomjs to puppeteer. One of the functions that we used from the phantomjs codebase was `window.callPhantom` which triggered a phantomjs render when the page itself deemed it ready to do so.
http://phantomjs.org/api/webpage/handler/on-callback.html

it wasn't immediately clear to me how to accomplish this with the puppeteer api. But this example is what I came up with. Perhaps, this would help provide users coming from phantomjs, some guidance when porting over this specific function - or page callbacks in general.

### DIFF
--- a/examples/call-phantom.js
+++ b/examples/call-phantom.js
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


aslushnikov:
We already have an example for the `exposeFunction` method: https://github.com/GoogleChrome/puppeteer/blob/ce8a952044ef318baf3d9ff25157e41571293de5/examples/custom-event.js#L26-L30

I'd appreciate if you can come up with a blog post / article that would sum up your experience migrating phantomJS codebase over to puppeteer.

jbryson3:
I must've missed that example, thanks for pointing me in that direction. 

I'll keep that request in mind as I'm migrating.

PR #1233, opened 2017-10-31
docs(contributing): define semantic commit message rules

aslushnikov:
Semantic commit messages drastically simplify preparation of releases.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,7 +47,28 @@ npm run lint
 
 ## Commit Messages
 
-Commit messages should follow the Semantic Commit Messages format.
+Commit messages should follow the Semantic Commit Messages format:
+
+```
+label(namespace): title
+
+description
+
+footer
+```
+
+1. *label* is one of the following:
+    - `fix` - puppeteer bug fixes
+    - `feat` - puppeteer features
+    - `docs` - changes to docs, e.g. `docs(api.md): ..` to change documentation
+    - `test` - changes to puppeteer tests infrastructure
+    - `style` - puppeteer code style: spaces/alignment/wrapping etc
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -45,6 +45,36 @@ To run code linter, use:
 npm run lint
 ```
 
+## Commit Messages
+
+Commit messages should follow the Semantic Commit Messages format.
```
ebidel: is there a link for "Semantic Commit Messages" that we can link to?

CONTRIBUTING.md
```diff
@@ -45,6 +45,36 @@ To run code linter, use:
...
+Example:
+
+```
+fix(Page): fix page.pizza method
+
+This patch fixes page.pizza so that it works with iframes.
+
+Fixes #123, Fixes #234
+
+BREAKING CHANGE: page.pizza now delivers pizza at home by default.
```
ebidel, line 82: awesome feature!

CONTRIBUTING.md
```diff
@@ -45,6 +45,36 @@ To run code linter, use:
...
+To deliver to a different location, use "deliver" option:
+  `page.pizza({deliver: 'work'})`.
+```
+1. *label* is one of the following:
```
ebidel: It's not immediately obvious where each of these items maps to in your example. Is there a way to show an abstract example along with the concrete pizze one?

Like:
```
label(namespace): description

body

footer
```

ebidel (approved):

CONTRIBUTING.md
aslushnikov: No good link that I can refer to

CONTRIBUTING.md
aslushnikov: indeed; done, should be better now

PR #1237, opened 2017-11-01
fix(Navigation): do not race with security error for navigation

### LINKED ISSUE #1163: Test "Puppeteer.launch should support ignoreHTTPSErrors option" is flaky
It looks like the test started to flake after #1141.
The possible reason might be the lifecycle events that we catch from the initial target navigation.

### LINKED ISSUE #1195: Test `Page.goto should fail when navigating to bad SSL` is flaky

aslushnikov:
Currently, NavigationWatcher listens to lifecycle events from Page
domain and security events from Security domain.

However, the events are dispatched from different processes in browser:
- Page's lifecycle events are dispatched from renderer process
- Security events are dispatched from browser process

This makes for the undefined order between events and results in
NavigationWatcher reporting different failuer messages, based on
the event order.

This patch stops relying on security errors in navigation watcher and
instead switches to request failure codes for the main resource.

Fixes #1195, Fixes #1163

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -173,10 +173,10 @@ class FrameManager extends EventEmitter {
   }
 
   /**
-   * @return {string}
+   * @return {?string}
    */
   unreachableMainFrameURL() {
-    return this._mainFrame._unreachableURL;
+    return this._mainFrame._unreachableURL || null;
   }
 }
 
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -173,10 +173,10 @@ class FrameManager extends EventEmitter {
   }
 
   /**
-   * @return {boolean}
+   * @return {string}
    */
-  isMainFrameLoadingFailed() {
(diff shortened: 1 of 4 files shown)


aslushnikov:
@Janpot this should also address your comment here  https://github.com/GoogleChrome/puppeteer/issues/901#issuecomment-333005738

lib/FrameManager.js
```diff
@@ -173,10 +173,10 @@ class FrameManager extends EventEmitter {
   }
 
   /**
-   * @return {boolean}
+   * @return {string}
```
JoelEinbinder: ?string

JoelEinbinder (approved):

lib/FrameManager.js
aslushnikov: Donel

PR #1239, opened 2017-11-01
fix: support PlzNavigate in puppeteer.

### LINKED ISSUE #877: Adopt plznavigate in puppeteer

aslushnikov:
This patch migrates puppeteer to support PlzNavigate chromium
project.

As a consequence of this patch, we no longer wait for both
`requestWillBeSent` and `requestIntercepted` events to happen. This should
resolve a ton of request interception bugs that "hanged" the loading.

Fixes #877.

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -16,7 +16,6 @@
 const EventEmitter = require('events');
 const {helper, debugError} = require('./helper');
 const Multimap = require('./Multimap');
-const url = require('url');
 
 class NetworkManager extends EventEmitter {
   /**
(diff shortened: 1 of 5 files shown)


lib/NetworkManager.js
```diff
@@ -179,30 +181,39 @@ class NetworkManager extends EventEmitter {
   }
 
   /**
-   * @param {string} requestId
+   * @param {?string} requestId
    * @param {string} interceptionId
```
JoelEinbinder: ?string

JoelEinbinder (approved):

lib/NetworkManager.js

PR #1243, opened 2017-11-01
Page.select Calling mainframe.select() Fixes #1139 

### LINKED ISSUE #1139: Missing ElementHandle.select; impossible (?) to make selection in a frame

suvratjain1995:


aslushnikov:
@suvratjain1995 are you the author of #1198? If yes, should the #1198 be closed in favor of this PR?

suvratjain1995:
@aslushnikov yes I am author that ID was having some issue regarding cla and all

docs/api.md
```diff
@@ -1507,6 +1508,19 @@ If the name is empty, returns the id attribute instead.
 #### frame.parentFrame()
 - returns: <[Frame]> Returns parent frame, if any. Detached frames and main frames return `null`.
 
+#### frame.select(selector, ...values)
+- `selector` <[string]> A [selector] to query page for
```
aslushnikov: nit: s/page/frame

test/test.js
```diff
@@ -901,7 +901,7 @@ describe('Page', function() {
       await page.goto('asdfasdf').catch(e => error = e);
       expect(error.message).toContain('Cannot navigate to invalid URL');
     }));
-    it('should fail when navigating to bad SSL', SX(async function() {
+    xit('should fail when navigating to bad SSL', SX(async function() {
```
aslushnikov: not need to skip this any more

lib/Page.js
```diff
@@ -782,17 +782,7 @@ class Page extends EventEmitter {
...
-        throw new Error('Element is not a <select> element.');
-      const options = Array.from(element.options);
-      element.value = undefined;
-      for (const option of options)
-        option.selected = values.includes(option.value);
-      element.dispatchEvent(new Event('input', { 'bubbles': true }));
-      element.dispatchEvent(new Event('change', { 'bubbles': true }));
-      return options.filter(option => option.selected).map(option => option.value);
-    }, values);
+    return this.mainFrame().select(selector, ...values);
```
aslushnikov, line 785: can you please state in `page.select` documentation that it's a shortcut for `page.mainFrame().select`.

E.g. see how it's done for [page.$](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageselector)

aslushnikov (commented):
Overall looks good, just a few nits

docs/api.md
suvratjain1995: @aslushnikov I couldnt get you here.

aslushnikov: @suvratjain1995 sorry, I meant that "frame" should be used instead of "page".

lib/FrameManager.js
```diff
@@ -407,6 +407,25 @@ class Frame {
   }
 
   /**
+  *@param {string} selector
+  *@param {!Array<string>} values
```
aslushnikov: 1. there should be spaces between "*" and "@"
2. this should also have a @return annotation:

```js
/**
  ...
  * @return {!Promise<!Array<string>>}
 */
```

suvratjain1995:
@aslushnikov worked out those .

aslushnikov (approved):
Awesome, thanks!

PR #1251, opened 2017-11-02
docs(api.md): fix return value of frame.select()

vsemozhetbyt:
It seems this should be in sync with [`page.select()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pageselectselector-values).


aslushnikov (approved):
Thanks!

Issue #1244, opened 2017-11-01
Add page bringToFront method


schelkun:
We need to add `bringToFront` method to page. Why it doesn't exist in puppeteer?
I can create PR, but I'm not really sure about comment and docs. Is docs auto generated or not?

```js
  /**
   * @return {!Promise<?Response>}
   */
  async bringToFront() {
    await this._client.send('Page.bringToFront');
  }
```

aslushnikov:
> We need to add bringToFront method to page. Why it doesn't exist in puppeteer?

@schelkun there's been no need to. Can you please share your user scenario?

satanch:
@aslushnikov for example, I will open 3 pages (tabs), how can I activate second or first tab?
By using `Page.bringToFront`. It's working for me. Any other way to do tihs?

schelkun:
@aslushnikov for example, I will open 3 pages (tabs), how can I activate second or first tab?
By using `Page.bringToFront`. It's working for me. Any other way to do this?

aslushnikov:
>  for example, I will open 3 pages (tabs), how can I activate second or first tab?

@schelkun that's a good scenario :)

> I can create PR, but I'm not really sure about comment and docs. Is docs auto generated or not?

- Doc is not autogenerated, but  linted
- Comments are [closure annotations](https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler).

I'll be happy to review the pr that adds `page.bringToFront` method. Please, make sure to read [CONTRIBUTING.md](https://github.com/GoogleChrome/puppeteer/blob/master/CONTRIBUTING.md) if not yet - it has a lot of useful information.

PR #1252, opened 2017-11-02
Issue #1244 - Page bringToFront method

schelkun:
According to: https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-bringToFront
Accept nothing and returns nothing. Please check, and advice some changes.

More information at https://github.com/GoogleChrome/puppeteer/issues/1244

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -519,6 +519,10 @@ Provide credentials for [http authentication](https://developer.mozilla.org/en-U
 
 To disable authentication, pass `null`.
 
+#### page.bringToFront()
+
+Brings page to front (activates tab).
+
 #### page.click(selector[, options])
 - `selector` <[string]> A [selector] to search for element to click. If there are multiple elements satisfying the selector, the first will be clicked.
 - `options` <[Object]>
--- a/docs/api.md
+++ b/docs/api.md
@@ -522,6 +522,8 @@ To disable authentication, pass `null`.
 
 #### page.bringToFront()
 
+- returns: <[Promise]>
+
 Brings page to front (activates tab).
 
 #### page.click(selector[, options])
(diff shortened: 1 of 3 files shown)


aslushnikov:
Hey @schelkun, thanks for the PR.

SInce this is a headful-only feature and we currently don't have headful tests, can you please exclude the method from coverage checker? You'll need to initialize `disabled` set differently:

```js
const disabled = new Set(['page.bringToFront']);
```

https://github.com/GoogleChrome/puppeteer/blob/f5bb333cd0c42306f47e7767e5cb8b78af557c27/test/test.js#L3045-L3058

schelkun:
@aslushnikov  what about headless? How page focus working on new page opened?
What about blur and focus events?

aslushnikov:
> @aslushnikov what about headless? How page focus working on new page opened?
What about blur and focus events?

I'd expect the "bringToFront" to have no effect on headless. Let me know if the focus/blur/visibilityState behaves weirdly.

lib/Page.js
```diff
@@ -544,6 +544,13 @@ class Page extends EventEmitter {
   }
 
   /**
+  * @return {!Promise<?Response>}
```
aslushnikov: Please remove the "@return" statement - it's not needed since the function doesn't return anything. (technically, the function returns a promise, but there's an agreement to not declare empty promises)

schelkun:
@aslushnikov done

docs/api.md
```diff
@@ -519,6 +520,10 @@ Provide credentials for [http authentication](https://developer.mozilla.org/en-U
 
 To disable authentication, pass `null`.
 
+#### page.bringToFront()
```
aslushnikov, line 523: here return value needs to be declared:

```
#### page.bringToFront()

- returns: <[Promise]>

Brings page to front (activates tab).
```

You can check that your CL passes all lints/checks by running `npm test` locally.

aslushnikov (approved):
Thanks!

PR #1258, opened 2017-11-02
feat(Frame) addStyle returns element handle

### LINKED ISSUE #1179: Add an option to set data attributes on elements added by the API
It would be useful to be able to set [data attributes](https://html.spec.whatwg.org/multipage/dom.html#custom-data-attribute) on `script` and `style` elements added by puppeteer.

For instance:

```js
await page.addScriptTag({
  url: 'https://example.com/script.js',
  attributes: { foo: bar },
});
```
would add the following tag:

```html
<script type="text/javascript" src="https://example.com/script.js" data-foo="bar">
```

For instance, I inject a script that (for a reason) selects all the `script` elements in my pages, but then it selects itself (and its dependencies). I'd like to be able to filter out all the elements added by puppeteer. Having a custom `data-*` attribute would allow me to do that. I hope it makes sense.

xg-wang:
fix #1179 
@aslushnikov Also added some tests. If this looks good, I can do same changes to the `addScript`.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -323,86 +323,92 @@ class Frame {
 
   /**
    * @param {Object} options
-   * @return {!Promise}
+   * @return {!Promise<!ElementHandle>}
    */
   async addScriptTag(options) {
     if (typeof options.url === 'string')
-      return this.evaluate(addScriptUrl, options.url);
+      return this._context.evaluateHandle(addScriptUrl, options.url);
 
     if (typeof options.path === 'string') {
       let contents = await readFileAsync(options.path, 'utf8');
       contents += '//# sourceURL=' + options.path.replace(/\n/g, '');
 
-      return this.evaluate(addScriptContent, contents);
+      return this._context.evaluateHandle(addScriptContent, contents);
     }
 
     if (typeof options.content === 'string')
-      return this.evaluate(addScriptContent, options.content);
+      return this._context.evaluateHandle(addScriptContent, options.content);
 
     throw new Error('Provide an object with a `url`, `path` or `content` property');
 
     /**
      * @param {string} url
+     * @return {!Promise<!HTMLElement>}
      */
-    function addScriptUrl(url) {
+    async function addScriptUrl(url) {
       const script = document.createElement('script');
       script.src = url;
-      const promise = new Promise(x => script.onload = x);
       document.head.appendChild(script);
-      return promise;
+      await new Promise(x => script.onload = x);
+      return script;
     }
 
     /**
      * @param {string} content
+     * @return {!HTMLElement}
      */
     function addScriptContent(content) {
       const script = document.createElement('script');
       script.type = 'text/javascript';
       script.text = content;
       document.head.appendChild(script);
+      return script;
     }
   }
 
   /**
    * @param {Object} options
-   * @return {!Promise}
+   * @return {!Promise<!ElementHandle>}
    */
   async addStyleTag(options) {
     if (typeof options.url === 'string')
-      return this.evaluate(addStyleUrl, options.url);
+      return await this._context.evaluateHandle(addStyleUrl, options.url);
 
     if (typeof options.path === 'string') {
       let contents = await readFileAsync(options.path, 'utf8');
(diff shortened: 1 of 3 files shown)


xg-wang:
Signed.

Best,
Xingan

lib/FrameManager.js
```diff
@@ -390,7 +390,7 @@ class Frame {
       const link = document.createElement('link');
       link.rel = 'stylesheet';
       link.href = url;
-      const promise = new Promise(x => link.onload = x);
+      const promise = new Promise(res => link.onload = () => res(link));
```
aslushnikov: arrow function in arrow function is a little hard to read.

I think this could be written simpler:
```js
async function addStyleUrl(url) {
  const link = document.createElement('link');
  link.rel = 'stylesheet';
  link.href = url;
  document.head.appendChild(link);
  await new Promise(x => link.onload = x);
  return link;
}
```

lib/FrameManager.js
```diff
@@ -403,6 +403,7 @@ class Frame {
       style.type = 'text/css';
       style.appendChild(document.createTextNode(content));
       document.head.appendChild(style);
+      return Promise.resolve(style);
```
aslushnikov: no need to `Promise.resolve`, just return:
```js
return style;
```

test/test.js
```diff
@@ -2416,6 +2416,37 @@ describe('Page', function() {
       await page.addStyleTag({ content: 'body { background-color: green; }' });
       expect(await page.evaluate(`window.getComputedStyle(document.querySelector('body')).getPropertyValue('background-color')`)).toBe('rgb(0, 128, 0)');
     }));
+
+    it('should returning element handle work with a url', SX(async function() {
```
aslushnikov: These are a lot of tests for a not-involved code.

Can you please make sure that `await page.addStyleTag` in three already-existing tests return `ElementHandle`? This would be enough for the change.

aslushnikov (commented):
Hey @xg-wang,

Thanks for the PR. Can we have the same for the `addScriptTag` method right away so that they are consistent with each other?

test/test.js
xg-wang: Thanks for the review, I removed these and just check `.asElement()`. Also updated the api doc.

test/test.js
```diff
@@ -2355,21 +2355,27 @@ describe('Page', function() {
 
     it('should work with a url', SX(async function() {
       await page.goto(EMPTY_PAGE);
-      await page.addScriptTag({ url: '/injectedfile.js' });
+      const scriptHandle = await page.addScriptTag({ url: '/injectedfile.js' });
+      expect(scriptHandle.asElement()).not.toBeNull();
       expect(await page.evaluate(() => __injected)).toBe(42);
+      scriptHandle.dispose();
```
aslushnikov: (here and in other tests): This should be awaited so that this doesn't race with the page been closed after the test.

However, I wouldn't bother disposing these handles: since page is closed after every test, all handles are automatically released.

test/test.js
```diff
@@ -2392,29 +2398,34 @@ describe('Page', function() {
...
     }));
 
     it('should include sourcemap when path is provided', SX(async function() {
       await page.goto(EMPTY_PAGE);
-      await page.addStyleTag({ path: path.join(__dirname, 'assets/injectedstyle.css') });
-      const styleHandle = await page.$('style');
+      const styleHandle = await page.addStyleTag({ path: path.join(__dirname, 'assets/injectedstyle.css') });
       const styleContent = await page.evaluate(style => style.innerHTML, styleHandle);
       expect(styleContent).toContain(path.join('assets', 'injectedstyle.css'));
       styleHandle.dispose();
```
aslushnikov: while we're here, can you please drop this?

test/test.js
```diff
@@ -2408,12 +2413,12 @@ describe('Page', function() {
       const styleHandle = await page.$('style');
       const styleContent = await page.evaluate(style => style.innerHTML, styleHandle);
       expect(styleContent).toContain(path.join('assets', 'injectedstyle.css'));
-      styleHandle.dispose();
```
xg-wang, line 2411: @aslushnikov dropped this

aslushnikov (approved):
Awesome, thanks

PR #1269, opened 2017-11-03
docs(api.md): fix return values of page.add*Tag()

vsemozhetbyt:
It seems this should be in sync with [`frame.addScriptTag()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#frameaddscripttagoptions) and [`frame.addStyleTag()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#frameaddstyletagoptions).

Follow-up of https://github.com/GoogleChrome/puppeteer/pull/1258.


aslushnikov (approved):
Thanks, I missed this in that review.

PR #1280, opened 2017-11-04
fix(Page.select): assert all values are of type "string".

### LINKED ISSUE #1276: page.select() should error if it can't find an option with requested value
### Steps to reproduce

* Puppeteer version: 0.13-alpha
* Platform / OS version: macos

html:
```html
<html><body><select><option value="12"/></select></body</html>
```
puppeteer:
```js
page.select('select', 12)
```

Either select the option with the value that is string-equal to the second arg (0.12 behavior) OR throw an error when page.select cannot find a matching option.  The latter solution seems more generally useful.

**What happens instead?**
Select is blanked, no error is thrown

aslushnikov:
This patch starts asserting that all values are of type "string".

The alternative approach to cast values to strings
might yield a hard-to-debug errors.

Fixes #1276.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -2737,7 +2737,6 @@ describe('Page', function() {
       expect(await page.evaluate(() => result.onInput)).toEqual(['blue']);
       expect(await page.evaluate(() => result.onChange)).toEqual(['blue']);
     }));
-
     it('should select multiple options', SX(async function() {
       await page.goto(PREFIX + '/input/select.html');
       await page.evaluate(() => makeMultiple());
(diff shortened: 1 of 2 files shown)


JoelEinbinder:
I think we should coerce to strings. That's what `<select>` does.

aslushnikov:
Implicit typecasting is unpredictable at times.

We don't save much in terms of usability, so I prefer a strict API that will simplify debugging story.

JoelEinbinder (approved):

test/test.js
```diff
@@ -2737,69 +2737,69 @@ describe('Page', function() {
...
       const result = await page.select('select','42','blue','black','magenta');
       expect(result.length).toEqual(1);
     }));
-
     it('should return [] on no values',SX(async function() {
       await page.goto(PREFIX + '/input/select.html');
       const result = await page.select('select');
       expect(result).toEqual([]);
     }));
-
```
JoelEinbinder, line 2787: What happened to my newlines?

aslushnikov: We don't have newlines between "it" tests, do we?

PR #1285, opened 2017-11-06
feat(Page,ElementHandle,Input): add callback when typing each char

### LINKED ISSUE #1272: feature: page.keyboard.type support callback function
keyboard.type(text, options)  => keyboard.type(text, options, callback)

because i need to do something for each character in the text

daviscai:
feat(Page, ElementHandle,Input): add callback when typing each char

This patch support callback function when typing each char in text

Fixes #1272

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -83,7 +83,7 @@
   * [page.title()](#pagetitle)
   * [page.touchscreen](#pagetouchscreen)
   * [page.tracing](#pagetracing)
-  * [page.type(selector, text[, options])](#pagetypeselector-text-options)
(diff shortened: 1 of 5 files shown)


daviscai:
I signed it!

daviscai:
I had update email

aslushnikov:
Hey @daviscai,

Is there a reason why the following wouldn't work for you?
```js
async function typeWithCallback(text, callback) {
  for (let i = 0; i < text.length; ++i) {
    await page.keyboard.type(text[i]);
    callback();
  }
}
```

PR #1287, opened 2017-11-06
feat(Frame) reject error for addScriptTag/addStyleTag

### LINKED ISSUE #1221: calling addScriptTag returns Promise that is not rejected in case of an error
### Steps to reproduce

Puppeteer version: 0.12.0
Platform / OS version: macOS El Capitan

If loading external script fails (wrong url, internet is down ...), if you call:

await page.addScriptTag({url: "someurl"});

nothing will happen and execution of a code will be stuck on this line.

Promise should be rejected because of an error.

**What happens instead?**

Promise is not rejected because in FrameManager.js "onerror" callback is not registered.

xg-wang:
fix #1221

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -326,8 +326,14 @@ class Frame {
    * @return {!Promise<!ElementHandle>}
    */
   async addScriptTag(options) {
-    if (typeof options.url === 'string')
-      return this._context.evaluateHandle(addScriptUrl, options.url);
+    if (typeof options.url === 'string') {
+      const url = options.url;
+      try {
+        return await this._context.evaluateHandle(addScriptUrl, url);
+      } catch (error) {
+        throw new Error(`Loading script from ${url} failed`);
+      }
+    }
 
     if (typeof options.path === 'string') {
       let contents = await readFileAsync(options.path, 'utf8');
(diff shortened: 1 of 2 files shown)


xg-wang:
Reject onerror as discussed in the issue.
Add by path has already been taken care by the `readFileAsync`, only add by url is considered here.

lib/FrameManager.js
```diff
@@ -326,8 +326,15 @@ class Frame {
    * @return {!Promise<!ElementHandle>}
    */
   async addScriptTag(options) {
-    if (typeof options.url === 'string')
-      return this._context.evaluateHandle(addScriptUrl, options.url);
+    if (typeof options.url === 'string') {
+      let handle;
+      try {
+        handle = await this._context.evaluateHandle(addScriptUrl, options.url);
```
aslushnikov: a few nits:
1. there should be no need to extract a "handle" variable
2. I'd extract the "url" into a separate variable since `options.url` might've changed by client code at the moment of throwing error.

```js
const url = options.url;
try {
  return await this._context.evaluateHandle(addScriptUrl, url);
} catch (error) {
  throw new Error(`Loading script from ${url} failed`);
}
```

xg-wang: Thanks! This makes sense.

aslushnikov (approved):
Thanks!

PR #1295, opened 2017-11-06
fix: make exposeFunction work for frames

### LINKED ISSUE #690: Expose function doesn't add to frames the first time.
It looks like exposeFunction is only running in the mainframe until a reload happens.

JoelEinbinder:
Fixes #690

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -289,8 +289,9 @@ class Page extends EventEmitter {
     this._pageBindings[name] = puppeteerFunction;
 
     const expression = helper.evaluationString(addPageBinding, name);
-    await this._client.send('Page.addScriptToEvaluateOnNewDocument', { source: expression });
-    await this._client.send('Runtime.evaluate', { expression, returnByValue: true });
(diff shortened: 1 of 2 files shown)


lib/Page.js
```diff
@@ -289,8 +289,9 @@ class Page extends EventEmitter {
     this._pageBindings[name] = puppeteerFunction;
 
     const expression = helper.evaluationString(addPageBinding, name);
-    await this._client.send('Page.addScriptToEvaluateOnNewDocument', { source: expression });
-    await this._client.send('Runtime.evaluate', { expression, returnByValue: true });
+    await this._client.send('Page.addScriptToEvaluateOnNewDocument', {source: expression});
+    for (const frame of this.frames())
+      await frame.evaluate(expression).catch(debugError);
```
aslushnikov: nit: Promise.all it?

aslushnikov (approved):

lib/Page.js
JoelEinbinder: Done.

aslushnikov (approved):
Thanks!

PR #1301, opened 2017-11-07
Added assertion to page.type()

slaskis:
To make it match the other event shortcuts but mostly because I got confused by the ` TypeError: Cannot read property 'type' of null` exception i kept getting.


slaskis:
I signed it!

aslushnikov (approved):

aslushnikov:
Thanks!

PR #1304, opened 2017-11-07
Test screenshot of element that needs to be scrolled into view

geecu:
Taking a screenshot of an element that needs to be scrolled into view doesn't work

The resulting screenshot is empty. 

Unfortunately, this pull request has only a (failing) test case, not a fix. I couldn't figure out how to fix this, maybe the problem comes from the DevTools protocol... Weirdly, taking a screenshot of the entire viewport, without the clip option seems correct, with the desired element scrolled into view...


cdaringe:
debugged this. see https://github.com/GoogleChrome/puppeteer/issues/1315

geecu:
I signed it!

geecu:
fixed by #1315

PR #1308, opened 2017-11-07
Load crx example tip

ebidel:


examples/README.md
```diff
@@ -6,6 +6,22 @@ Assuming you have a checkout of the Puppeteer repo and have run yarn (or npm i)
...
+# Tips & Tricks
+
+### Load a Chrome extension
+
+By default, Puppeteer disables extensions when launching Chrome. You can load a specific
+extension using: 
+
+```js
+const browser = await puppeteer.launch({
+  args: [
```
JoelEinbinder: You also need to be in headful mode.

ebidel: done

aslushnikov (approved):
awesome!

Issue #1294, opened 2017-11-06
Broken Click after upgrading to 0.12.0


divyamvn:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version:0.12.0
* Platform / OS version: Mac OS Sierra
* URLs (if applicable): Testing on LocalHost

**What steps will reproduce the problem?**
My existing working test started failing after upgrading to 0.12.0.
I was looking into similar issues with click. However, this seems different to me and this is the very first Step in my test. Is this because of the 'href' property in my css selector?
First step after launching browser, is to click on a main menu item and then on sub menu item, code for which is provided below.
This was very well working before upgrade.
This is NOT the ONLY place, where Click suddenly started failing. I am getting the "Error: Protocol error (DOM.getBoxModel): Could not compute box model. undefined" for various other buttons.

_Please include code that reproduces the issue._

await page.click("div.menus ul.main a[href='#abc']");
await page.waitForSelector( "div.menus ul.sub a[href='#abc/childOfabc']", { visible: true } );
//I added the below focus just to verify its working and my selector is being recognized    
await page.focus("div.menus ul.sub a[href='#abc/childOfabc']");
await page.click("div.menus ul.sub a[href='#abc/childOfabc']");

**What is the expected result?**
 Expected is that Click works normally.

**What happens instead?**
Nothing happens. After default time out the test fails. Below is the error message.
(node:52818) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: waiting failed: timeout 30000ms exceeded
(node:52818) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

JoelEinbinder:
The main problem here should be fixed by #1299, which properly handles retina display with DOM.getBoxModel. But I am very curious about the error `Error: Protocol error (DOM.getBoxModel): Could not compute box model. undefined`. Do you have a repro case for that? Thanks!

aslushnikov:
@divyamvn @patrickeakin can you guys please share a repro case where you experience the `DOM.getBoxModel` error?

```
Error: Protocol error (DOM.getBoxModel): Could not compute box model. undefined
```

rafagsiqueira:
I can confirm this also happens for puppeteer@0.13.0-alpha.

divyamvn:
Thank you for the response.
[...]

PR #1309, opened 2017-11-07
test: more element click tests and better error messages

JoelEinbinder:
Adds a bunch more tests for Element.click, and throw new `'Node is not visible'` error message.

### DIFF
--- a/lib/ElementHandle.js
+++ b/lib/ElementHandle.js
@@ -15,7 +15,7 @@
  */
 const path = require('path');
 const {JSHandle} = require('./ExecutionContext');
-const {helper} = require('./helper');
+const {helper, debugError} = require('./helper');
 
 class ElementHandle extends JSHandle {
   /**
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -2559,6 +2579,15 @@ describe('Page', function() {
       expect(await page.evaluate(() => window.innerWidth)).toBe(375);
       expect(await page.evaluate(() => navigator.userAgent)).toContain('Safari');
     }));
+    it('should support clicking', SX(async function() {
+      await page.goto(PREFIX + '/mobile.html');
```
aslushnikov: why is this navigation needed?

lib/ElementHandle.js
```diff
@@ -119,13 +119,14 @@ class ElementHandle extends JSHandle {
...
+    const result = await this._client.send('DOM.getBoxModel', {
       objectId: this._remoteObject.objectId
-    });
-    if (!model)
-      throw new Error('Node is detached from document');
+    }).catch(error => void debugError(error));
 
-    const quad = model.border;
+    if (!result)
+      throw new Error('Node is not visible');
```
aslushnikov: i'd return null in this case and throw in the `click` itself.

aslushnikov:
Is this a fix to the #1294?

test/test.js
JoelEinbinder: It's not, mis-copy.

lib/ElementHandle.js
JoelEinbinder: Done.

aslushnikov (approved):
Great,thanks!

PR #1311, opened 2017-11-07
docs: document osx keyboard shortcuts not working

### LINKED ISSUE #776: Keyboard shortcuts don't work on Mac
The following script doesn't select text:

```js
(async () => {
  // Launch headful so that we see what's going on
  // Add textarea
  await page.setContent('<textarea>hello</textarea>');
  await page.focus('textarea');
  // Try to select all text in textarea
  await page.keyboard.down('Meta');
  await page.keyboard.press('a');
  await page.keyboard.up('Meta');
  // By now, text should be selected in the text area.
  // It doesn't on Mac.
})();
```

First reported in #761.

JoelEinbinder:
Fixes #776

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1193,6 +1193,8 @@ await page.keyboard.press('KeyA');
 await page.keyboard.up('Shift');
 ```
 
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):
thanks!

Issue #672, opened 2017-09-04
Failed to download Chromium r497674! Set "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" env variable to skip download.


ystop:
mac 10.12.5
node v8.4.0
npm 5.3.0
puppeteer  0.10.2

when l " npm install", l get this:

ERROR: Failed to download Chromium r497674! Set "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" env variable to skip download.
npm WARN backendbrowser@1.0.0 No repository field.
npm WARN backendbrowser@1.0.0 No license field.

npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! puppeteer@0.10.2 install: `node install.js`
npm ERR! Exit status 1
npm ERR!
npm ERR! Failed at the puppeteer@0.10.2 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/xxxx/.npm/_logs/2017-09-04T02_50_51_507Z-debug.log

JoelEinbinder:
what happens if you run 
```bash
wget https://storage.googleapis.com/chromium-browser-snapshots/Mac/497674/chrome-mac.zip
```
Maybe you are behind a proxy or there is some other reason that the download is failing.

ystop:
@JoelEinbinder Thanks. Due to network, l do nothing, but  it's all right now.

mcrowder65:
Is there an alternative to this? I am trying to use this for pdf generation but our servers are pretty locked down to only using npm... is there an alternative download directly on npm?

mcrowder65:
oh i found that my http_proxy and https_proxy's weren't set on jenkins :)

Mr-LiuDC:
Hi bro. I also got this error on Windows 7. I have set the proxy host，and sure it's correct.

Failed to download Chromium r497674 ! Set "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" env variable to skip download.

Is there any suggestions？

Thanks.
@JoelEinbinder @ystop

ystop:
@Mr-LiuDC Due to network!Trust me. I'm Chinese, too.

renmoqiqi:
我也是个原因

henryatman:
@JoelEinbinder , what should i do if i donwloaded this zip by manual，where its should be?

[...]

PR #1314, opened 2017-11-08
support for npm no_proxy

tomasAlabes:
Support npm `no_proxy` configuration for the installation script, similar to #692 and helps with #672.

Having `.yarnrc` with:
```
proxy "my-proxy:80"
no_proxy "my_private_url"
```

What is happening a lot to us is that we have our proxy defined and we connect somewhere inside our vpn to get chromium, so it tries to go to the private location ("my_private_url" in the example) using the proxy, which makes it fail with:
```
[ERROR] ERROR: Failed to download Chromium r503964! Set “PUPPETEER_SKIP_CHROMIUM_DOWNLOAD” env variable to skip download.
[ERROR] { Error: write EPROTO 140521875134272:error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol:../deps/openssl/openssl/ssl/s23_clnt.c:797:
```
One workaround is setting the `no_proxy` env variable to the same `no_proxy` property in the `.yarnrc` but I think the chromium installer should be the one aware of that property.

Thanks.

### DIFF
--- a/install.js
+++ b/install.js
@@ -36,10 +36,14 @@ if (revisionInfo.downloaded)
 // Override current environment proxy settings with npm configuration, if any.
 const NPM_HTTPS_PROXY = process.env.npm_config_https_proxy || process.env.npm_config_proxy;
 const NPM_HTTP_PROXY = process.env.npm_config_http_proxy || process.env.npm_config_proxy;
+const NPM_NO_PROXY = process.env.npm_config_no_proxy;
+
 if (NPM_HTTPS_PROXY)
   process.env.HTTPS_PROXY = NPM_HTTPS_PROXY;
 if (NPM_HTTP_PROXY)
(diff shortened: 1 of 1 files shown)


aslushnikov (commented):
Thanks! Can you please also add the "NO_PROXY" env variable to `docs/api.md`?

tomasAlabes:
It's already there. It was mentioned but it wasn't used in the code.
```
Puppeteer looks for certain environment variables to aid its operations. These variables can either be set in the environment or in the npm config.

* HTTP_PROXY, HTTPS_PROXY, NO_PROXY - defines HTTP proxy settings that are used to download and run Chromium.
```

aslushnikov (approved):
Ah indeed, thanks!

Issue #1315, opened 2017-11-08
elementHandle.screenshot clip object uses wrong y value


cdaringe:
### problem

calls to `elementHandle.screenshot` use the incorrect y value, resulting in invalid images

### Steps to reproduce

- scroll an element into view
- get an element handle
- call elementHandle.screenshot
- observe chrome scroll to the top of the doc, at y: 0, and capture content from there
  - we use the boundingBox object as a input to `clip`, but when scrolled into view, `boundingBox.y` is ~0 (or a small value), and `Page.screenShot` seems to expect an absolute `y` value, from top of doc

example!  https://www.youtube.com/edit?video_id=TcGYq5d3n4c

puppeteer calls:
```
// Page.js
const result = await this._client.send('Page.captureScreenshot', { format, quality: options.quality, clip })
```

**What is the expected result?**

- a screenshot of the requested element

**What happens instead?**

- a screenshot of the correct dimensions, but at the top of my document, not over my actual element.
- or, in other words, a screenshot of the correct dimensions in the viewport, BUT, the viewport has moved

cdaringe:
ah ha!  see: https://chromedevtools.github.io/devtools-protocol/tot/Page/#type-VisualViewport

confirmed, pup's clip values need to be patched

cdaringe:
... https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot 

^^ would be handy too :)

cdaringe:
rather than being a productive citizen and landing a patch here (which i should :(), i'll at least share the hack fix:

```js
    // get the full height of the document
    var height = await page.evaluate(function () {
      var body = document.body
      var html = document.documentElement
      var height = Math.max(body.scrollHeight, body.offsetHeight, html.clientHeight, html.scrollHeight, html.offsetHeight)
      return height
    })
    // set the viewport to be the full height of the document
    await page.setViewport({
      width, // whatever it may be
      height
    })
```

^^ this _only works_ if you're always starting from the top of the document to begin with.

cdaringe:
working on a PR now

cdaringe:
see #1323

cdaringe:
see #1323

waqarorakzai:
My problem was i was setting the Viewport after goTo(), when change it it worked for me perfectly

```
`const puppeteer = require('puppeteer');

async function run(url) {
    let browser = await puppeteer.launch({ headless: true });
    let page = await browser.newPage();
[...]

PR #1323, opened 2017-11-08
Fix/element screenshot

cdaringe:
# problem statement

- element screenshot capture doesn't work when elements are outside of the original viewport

# solution

- use absolute coordinates

# discussion

- ~https://chromedevtools.github.io/devtools-protocol/tot/Page/#type-Viewport is ambiguous about whether `x` and `y` are relative/absolute to what reference~
- https://github.com/ChromeDevTools/devtools-protocol/issues/67

closes #1304 & #1315


cdaringe:
I signed it!

JoelEinbinder:
This looks good, but I am debating whether or not this is a bug in the protocol, and that `Page.captureScreenshot` should speak the same language as `getBoundingClientRect`

cdaringe:
mmm, interesting.  fair. 

- `Page.captureScreenshot` in the protocol expects a `Viewport` object.  this seems correct.  it locates and defines a rectangle relative to the document.
- `getBoundingClientRect` is paired to the corresponding browser method, and returns a [DOMRect](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect), which also seems correct for its use case.  it's x & y are relative to the current viewport.

the only thing i'm seeing as a viable change is expanding the `Viewport` protocol to be relative or absolute positioning aware.  however, i dont think they could use the exact same format w/out breaking backwards compatibility--one is relative, one is absolute.  thoughts?

aslushnikov (approved):
> This looks good, but I am debating whether or not this is a bug in the protocol, and that Page.captureScreenshot should speak the same language as getBoundingClientRect


We might want to change it in protocol, but it is not critical for now.

Thanks for the PR!

PR #1336, opened 2017-11-09
Set blocked urls

skysteve:
Implement support for `setBlockedURLs` as per https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-setBlockedURLs

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -72,6 +72,7 @@
   * [page.reload(options)](#pagereloadoptions)
   * [page.screenshot([options])](#pagescreenshotoptions)
   * [page.select(selector, ...values)](#pageselectselector-values)
+  * [page.setBlockedURLs(blockedUrls)](#pagesetblockedurlsblockedurls)
   * [page.setContent(html)](#pagesetcontenthtml)
   * [page.setCookie(...cookies)](#pagesetcookiecookies)
   * [page.setExtraHTTPHeaders(headers)](#pagesetextrahttpheadersheaders)
@@ -953,6 +954,10 @@ page.select('select#colors', 'red', 'green', 'blue'); // multiple selections
 ```
 
 Shortcut for [page.mainFrame.select()](#frameselectselector-values)
+
+#### page.setBlockedURLs(blockedUrls)
+- `blockedUrls` <[Array]<[string]>> Urls to block, may also be a [match pattern](https://developer.chrome.com/apps/match_patterns).
+- returns: <[Promise]>
 
 #### page.setContent(html)
 - `html` <[string]> HTML markup to assign to the page.
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -30,6 +30,7 @@ class NetworkManager extends EventEmitter {
     this._interceptionIdToRequest = new Map();
     /** @type {!Object<string, string>} */
     this._extraHTTPHeaders = {};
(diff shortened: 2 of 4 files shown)


skysteve:
I signed it!

docs/api.md
```diff
@@ -954,6 +955,10 @@ page.select('select#colors', 'red', 'green', 'blue'); // multiple selections
 
 Shortcut for [page.mainFrame.select()](#frameselectselector-values)
 
+#### page.setBlockedURLs(blockedUrls)
+- `blockedUrls` <[Array]<[string]>> Urls to block, may also be a [match pattern](https://developer.chrome.com/apps/match_patterns).
```
ebidel, line 959: https://developer.chrome.com/apps/match_patterns is 404ing for me. Is there a better link?

vsemozhetbyt: Seems working for me.

<details>
<summary>200:</summary>




</details><br>

skysteve: Hey @ebidel as @vsemozhetbyt says, it works for me too 😕

ebidel: Must have been my shuttle wifi. It's working for me too :)

lib/NetworkManager.js
```diff
@@ -110,6 +111,21 @@ class NetworkManager extends EventEmitter {
...
+  async setBlockedURLs(blockedUrls) {
+    this._blockedUrls = [];
+
+    console.assert(Array.isArray(blockedUrls), `Expected blocked urls to be an array of Strings, but "${typeof blockedUrls}" is found.`);
+
+    blockedUrls.forEach(url => {
+      console.assert(helper.isString(url), `Expected url "${url}" to be a String, but "${typeof url}" is found`);
+      this._blockedUrls.push(url);
+    });
+    await this._client.send('Network.setBlockedURLs', { urls: this._blockedUrls });
```
aslushnikov, line 126: The `Network.setBlockedURLs` is deprecated; please don't use it.

There are interception patterns as a part of `setRequestInterceptionEnabled`. However, we already have request interception implemented so it should cover the usecase.

Why would you want this feature? Is it out of convenience?

avimar: Seems we already have this, e.g. https://github.com/GoogleChrome/puppeteer/blob/master/examples/block-images.js

If you would make it that you can attach a handler to the `browser` element and that somehow attaches it to all sub-pages, so it intercepts every single request, that would be great. (Trying to add via `browser.on('targetcreated',handler)` seems to miss the first few network events.)

skysteve: Apologies, I thought the `experimental` flag on the docs meant it was new rather than deprecated.

the TLRD; of why I want/need the feature is to block third party scripts when testing things as they slow things down. It would be nice to just set some blocked urls rather than have to intercept every request and do it manually.

aslushnikov: > Apologies, I thought the experimental flag on the docs meant it was new rather than deprecated.

The flag means we're experimenting with the method support. Usually, experimental methods either get promoted to non-experimental (and we commit to LTS support), or get removed (like in case of `setBlockedURLs`).

> It would be nice to just set some blocked urls rather than have to intercept every request and do it manually.

in this case, we can introduce a helper method `setBlockedURLs` that uses request interception under the hood. Let me know if you're interested in implementing this.

PR #1337, opened 2017-11-09
feat: Set default executablePath

### LINKED ISSUE #1335: Environment Variable & npm config for executablePath
You can't currently tell other packages to use a bundled chrome version. This could be implemented in the packages, for example [penthouse](https://github.com/pocketjoso/) (See https://github.com/pocketjoso/penthouse/pull/177 for discussions).

However it seems that it would be beneficial to implement this change here, as all consumers would benefit for it, and even use the same bundle version, unless something else was specified.

I'm thinking `PUPPETEER_EXECUTABLE_PATH` in the environment or in the npm config like the rest of the variables.

Here is the current code with no support for this: https://github.com/GoogleChrome/puppeteer/blob/f5bb333cd0c42306f47e7767e5cb8b78af557c27/lib/Launcher.js#L89-L94

Saturate:
Allows to set the default executablePath as the
`PUPPETEER_EXECUTABLE_PATH`  Environment Variable. Or with a npm config.

This is best used when you want to use a bundled chromium, in a third
party package where puppeteer is a dependency.

Fixes #1335

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -215,7 +215,9 @@ Puppeteer looks for certain [environment variables](https://en.wikipedia.org/wik
 
 - `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
 - `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step.
-- `PUPPETEER_DOWNLOAD_HOST` - overwrite host part of URL that is used to download Chromium
+- `PUPPETEER_DOWNLOAD_HOST` - overwrite host part of URL that is used to download Chromium.
+- `PUPPETEER_EXECUTABLE_PATH` - change the default executablePath, useful when using bundled Chromium, in a package that's depended on Puppeteer.
+
 
 ### class: Puppeteer
 
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -179,8 +179,12 @@ class Launcher {
    * @return {string}
    */
   static executablePath() {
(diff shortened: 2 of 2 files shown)


aslushnikov:
@Saturate This PR slipped by me, I apologize for the long delay.

I don't like having global env variables as a part of the puppeteer API. The reason is that it's not uncommon in npm world to have multiple versions of the same library in the dependency tree.

This would come problematic in case of `PUPPETEER_EXECUTABLE_PATH`: referenced chrome might work fine with new puppeteer, but might fail miserably with some old versions of the library.

Saturate:
It's fine, that happens :-)

I get your point, however it does have the effect that all peerDependencies that use puppeteer will have to implement something like this, if they want to use both the `SKIP_CHROMIUM_DOWNLOAD` and set a custom executable path and I guess these settings go hand in hand.

This become an issue if you are behind a corporate proxy, and want to use a bundled chromium with a tool like `penthouse` to generate Critical CSS.

We could just implement it in `penthouse` , but my first feeling was that this was better to implement it here as everyone could benefit from it.

You can just close this PR, and I'll create one for `penthouse` where I need this feature right now.

https://github.com/pocketjoso/penthouse/pull/177

aslushnikov:
> I get your point, however it does have the effect that all peerDependencies that use puppeteer will have to implement something like this, if they want to use both the SKIP_CHROMIUM_DOWNLOAD and set a custom executable path and I guess these settings go hand in hand.

Yeah, unfortunately env variables don't play nice with dependency tree.

> You can just close this PR, and I'll create one for penthouse where I need this feature right now.

Closing.

PR #1354, opened 2017-11-10
fix: waitForSelector "visible" option should check boundingClientRect

JoelEinbinder:
This makes `"visible"` work recursively. `visibility: hidden` is inherited, and any element inside of something with `display: none` will have no width/height. This will give a false negative on an element with `display: block; position: absolute; top:0; left:0; right:0; bottom:0; overflow: visible` and child elements.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -752,7 +752,7 @@ describe('Page', function() {
     it('should wait for visible', SX(async function() {
       let divFound = false;
       const waitForSelector = page.waitForSelector('div', {visible: true}).then(() => divFound = true);
-      await page.setContent(`<div style='display: none; visibility: hidden;'></div>`);
+      await page.setContent(`<div style='display: none; visibility: hidden;'>1</div>`);
       expect(divFound).toBe(false);
       await page.evaluate(() => document.querySelector('div').style.removeProperty('display'));
       expect(divFound).toBe(false);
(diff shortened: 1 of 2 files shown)


test/test.js
```diff
@@ -752,34 +752,50 @@ describe('Page', function() {
...
+      await page.evaluate(() => document.querySelector('div').style.removeProperty('visibility'));
+      expect(await waitForSelector).toBe(true);
+      expect(divVisible).toBe(true);
+    }));
     it('hidden should wait for visibility: hidden', SX(async function() {
       let divHidden = false;
-      await page.setContent(`<div style='display: block;'></div>`);
+      await page.setContent(`<div style='display: block;'>hi</div>`);
       const waitForSelector = page.waitForSelector('div', {hidden: true}).then(() => divHidden = true);
+      await page.evaluate('1'); // do a round trip
```
JoelEinbinder: These tests would always pass because they checked `divHidden` in the same microtask as it was assigned.

test/test.js
```diff
@@ -752,18 +752,31 @@ describe('Page', function() {
...
+      expect(divVisible).toBe(false);
+      await page.evaluate(() => document.querySelector('div').style.removeProperty('visibility'));
+      expect(await waitForSelector).toBe(true);
+      expect(divVisible).toBe(true);
+    }));
     it('hidden should wait for visibility: hidden', SX(async function() {
       let divHidden = false;
       await page.setContent(`<div style='display: block;'></div>`);
       const waitForSelector = page.waitForSelector('div', {hidden: true}).then(() => divHidden = true);
+      await page.evaluate('1'); // do a round trip
```
JoelEinbinder: These tests would always pass because they checked divHidden in the same microtask as it was assigned.

aslushnikov:
```js
expect(await page.waitForSelector('div')).toBe(true);
```

aslushnikov (approved):

PR #1364, opened 2017-11-12
docs(api.md): fix return signatures that contains null

seantanly:
Ref: https://github.com/GoogleChrome/puppeteer/issues/1358#issuecomment-343688600

Here's the PR to update the docs on master, hopefully this goes into 0.13.0 before the release.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -444,7 +444,7 @@ Emitted when a [response] is received.
 
 #### page.$(selector)
 - `selector` <[string]> A [selector] to query page for
-- returns: <[Promise]<[ElementHandle]>>
+- returns: <[Promise]<?[ElementHandle]>>
 
 The method runs `document.querySelector` within the page. If no element matches the selector, the return value resolve to `null`.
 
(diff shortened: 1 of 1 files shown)


seantanly:
I signed it!

docs/api.md
```diff
@@ -444,7 +444,7 @@ Emitted when a [response] is received.
 
 #### page.$(selector)
 - `selector` <[string]> A [selector] to query page for
-- returns: <[Promise]<[ElementHandle]>>
+- returns: <[Promise]<[ElementHandle] | `null`>>
```
aslushnikov: here and everywhere: let's use question mark to define nullable types.
E.g.:

```
returns: <[Promise]<?[ElementHandle]>>
```

This would align with jsdoc notation we currently use.

seantanly: Thanks for pointing that out. I have wrongly taken JSDocs notation to be the same as typescript's, in which optional type is denoted with `?` (`undefined`) and `null` value has to be explicitly stated. Apparently, in JSDocs `?` means nullable types.

I have updated the PR to question marks as required.

aslushnikov (approved):
Thank you.

PR #1372, opened 2017-11-13
Support both HTTP & HTTPS

### LINKED ISSUE #1371: Support for both HTTPS and HTTP in downloader
It's currently possible to overwrite the `PUPPETEER_DOWNLOAD_HOST`, but only with another HTTPS url, as the HTTPS module is used. While it's nice to have HTTPS be default, for people behind a corporate proxy it might not be possible to use HTTPS, or very hard to get it up and running.

It would be neat if it detected the required module from the URL, and just used that.

`https://storage.googleapis.com` would use the [HTTPS module](https://nodejs.org/api/https.html)
`http://custom-internal-mirror` would use the [HTTP module](https://nodejs.org/api/http.html)

Saturate:
Right now the downloader does not support a http mirror, this can
complicate things when you want to use a custom internal mirror.

HTTPS might be the best option to use, but for a internal server this
might not the optimal solution as it brings forward awhole other issue
of selfsigned certs and configuration.

This fixes #1371


utils/ChromiumDownloader.js
```diff
@@ -37,6 +37,21 @@ const downloadURLs = {
   win64: '%s/chromium-browser-snapshots/Win_x64/%d/chrome-win32.zip',
 };
 
+const protocolSelector = (function() {
+  const protocols = {
+    'http:': require('http'),
+    'https:': require('https'),
```
graingert: and http2 if available

graingert: https is already required at [line 18](https://github.com/GoogleChrome/puppeteer/pull/1372/files#diff-86a9034561c2112f5ab45ccf6a6f8415R18)

Saturate: Remove https from line 18. As of now it's a bit low level to detect support for HTTP/2 for this we would need in my mind a library that handled it all for us, one with unit tests so that we don't have to test and maintain that part here.

This could be done, but I guess it's a little overkill for a install script that downloads one file.

utils/ChromiumDownloader.js
```diff
@@ -193,7 +207,7 @@ function downloadFile(url, destinationPath, progressCallback) {
   const promise = new Promise((x, y) => { fulfill = x; reject = y; });
 
   const options = requestOptions(url);
-  const request = https.get(options, response => {
+  const request = protocolSelector(url).get(options, response => {
```
aslushnikov: Let's rename `requestOptions(url, method)` into `httpRequest(url, method)` and handle the http/https selection inside.

So here instead of getting options and then initiating a request, we'll have just

```js
const request = httpRequest(url, 'GET', response => {
  // ..
});
```

Saturate: That's a nice solution, now that I see it. I'll update the code!

Saturate: I digged a bit down in the code, and it seems that you are handling proxies with the following modules: `proxy-from-env` & `https-proxy-agent`.

This is needed for proxy support, but my suggestion would just be to replace it all with [request](https://www.npmjs.com/package/request), this has all the features while being simple and would greatly simplify the code needed. This would handle https/http/http2, proxies ect. It would also allow for the use of promises and Async/Await.

How would you feel about this?

aslushnikov: @Saturate `request` has a huge dependency tree. We intentionally avoid using it since the functionality is only needed for a single file download.

graingert: How about http://npm.anvaka.com/#/view/2d/axios

Saturate: Didn't really dig much into `request` but, I made a mistake thinking it was good as `npm` used it. I did try to implement the download with it, and it works fine with all proxy settings ect. However I can see that it's huge for the little things that we would use it for.

As far as I can see the `axios` packages does not support proxies via env vars, so we won't get that for free, however it does seem to figure out switching between HTTP and HTTPS. I was thinking https://www.npmjs.com/package/needle but they seem pretty similar, however `needle` has very few dependencies, but it's a lot bigger.

I could implement it with either pretty easily. Keeping the following dependencies for handling proxies:
```
```

Would a `axios` version be welcome?

On a side note, I found that the `node .\check_availability.js` script right now does not handle proxies, and thus does not work for me. I'll rewrite that as well.

Note: Also took at look at https://github.com/sindresorhus/got but it seems almost as bad as request.

Saturate: I'll just do what @aslushnikov proposed 5 days ago, and I'll make sure that it supports proxies, and reads the proxies the same way NPM does, this would be the minimal change and we would not have to add any new dependencies.

aslushnikov: > Would a axios version be welcome?

@Saturate we've spent quite a bit of time figuring the `https-proxy-agent` and `proxy-from-env` dependencies a while ago. We're pretty happy with the results so far, let's keep them as-is and avoid adding anything new (unless it's absolutely crucial)

Saturate:
@aslushnikov I implemented the change requested in https://github.com/GoogleChrome/puppeteer/pull/1372#discussion_r151234062

The builds are broken by an `socket hang up` error, but it works on all of my machines.
```
{ Error: socket hang up
    at createHangUpError (_http_client.js:302:15)
    at TLSSocket.socketOnEnd (_http_client.js:394:23)
    at emitNone (events.js:91:20)
    at TLSSocket.emit (events.js:188:7)
    at endReadableNT (_stream_readable.js:975:12)
    at _combinedTickCallback (internal/process/next_tick.js:80:11)
    at process._tickCallback (internal/process/next_tick.js:104:9) code: 'ECONNRESET' }
error Command failed with exit code 1.
```

utils/ChromiumDownloader.js
```diff
@@ -226,19 +223,22 @@ function extractZip(zipPath, folderPath) {
...
     const parsedProxyURL = URL.parse(proxyURL);
     parsedProxyURL.secureProxy = parsedProxyURL.protocol === 'https:';
 
-    result.agent = new ProxyAgent(parsedProxyURL);
+    options.agent = new ProxyAgent(parsedProxyURL);
   }
 
-  return result;
+  const driver = options.protocol === 'https:' ? require('https') : require('http');
+  const request = driver.request(options, response);
```
aslushnikov: @JoelEinbinder tsc is [unhappy](https://travis-ci.org/GoogleChrome/puppeteer/jobs/311095356) here: 

> utils/ChromiumDownloader.js(241,19): error TS2349: Cannot invoke an expression whose type lacks a call signature. Type '((options: string | URL | RequestOptions, callback?: (res: IncomingMessage) => void) => ClientReq...' has no compatible call signatures.

Any ideas how to please it?

aslushnikov (commented):
@Saturate there was a missing `request.end()` call; I added a few commits to fix the issue, I hope you don't mind.

However, tsc is unhappy now.

utils/ChromiumDownloader.js
Saturate: @aslushnikov the one with the least changes would be:

```
  const driver = options.protocol === 'https:' ? 'https' : 'http';
  const request = require(driver).request(options, response);
```

TypeScript like this!

Saturate:
Resolved the conflict with master, now it's all working.

aslushnikov (approved):
Awesome, thanks!

PR #1383, opened 2017-11-14
feat(errors): nicer stack traces on protocol errors

JoelEinbinder:
This rewrites protocol errors to have the stack trace of the call site.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -42,7 +42,7 @@ class Connection extends EventEmitter {
     super();
     this._url = url;
     this._lastId = 0;
-    /** @type {!Map<number, {resolve: function, reject: function, method: string}>}*/
+    /** @type {!Map<number, {resolve: function, reject: function, error: !Error, method: string}>}*/
     this._callbacks = new Map();
     this._delay = delay;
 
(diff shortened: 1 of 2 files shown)


aslushnikov:
@ak239 Alexey, that's our own version of stack tagging. What do you think of it?

JoelEinbinder:
Made things a bit simpler and added a test.

alexkozy (commented):
Looks good to me. We do not have anything in JavaScript runtime for async stacks yet.

lib/Connection.js
```diff
@@ -71,7 +71,7 @@ class Connection extends EventEmitter {
     debugProtocol('SEND ► ' + message);
     this._ws.send(message);
     return new Promise((resolve, reject) => {
-      this._callbacks.set(id, {resolve, reject, method});
+      this._callbacks.set(id, {resolve, reject: rewriteError.bind(null, new Error(), reject), method});
```
aslushnikov: instead of binding in the `rewriteError` function, let's just save the error as a part of the already-stored `resolve` and `reject` methods.

aslushnikov:
@JoelEinbinder do you want to land this?

lib/Connection.js
JoelEinbinder: Done.

JoelEinbinder:
Sure

JoelEinbinder:
ptal

aslushnikov (approved):

PR #1389, opened 2017-11-14
chore: bump version to 1.0.0rc

aslushnikov:


JoelEinbinder (approved):

joelgriffith:
Dumb question: does this also entail puppeteer landing/packing a stable version of Chrome as well?

aslushnikov:
@joelgriffith no, this step is to stabilize PPTR api.

alixaxel:
@aslushnikov Are there any plans to export the `Page` class from PPTR in version 1.0.0? Right now, if I want to add methods to the `Page` prototype, I need to run the same `lib`/`es6` switch logic that is present in `index.js`. Would be great if `Page` and other first-class classes were exported as non-defaults.

aslushnikov:
@alixaxel we currently have no plans to expose inner classes.

For your usecase, would it work if you just add methods to both node6 and regular versions of page?

```js
function extendPuppeteer() {
  extendPage(require('lib/Page'));
  extendPage(require('node6/Page'));
}

function extendPage(pageClass) {
  pageClass.prototype.foo = function() {
    // ...
  }
}
```

alixaxel:
@aslushnikov Yeah, exporting would be more elegant but something like that will do yes.

aslushnikov:
> @aslushnikov Yeah, exporting would be more elegant 

@alixaxel Would be. The reason we're reluctant on exporting classes is to limit API abuse: without classes there's no daunting desire to run constructors with hacked-in arguments.

PR #1398, opened 2017-11-16
feat(Browser): add remoteDebuggingPort method

Khady:
It is sometimes useful to have the remote debugging port which is used by the browser. To get some debugging information, or to use other tools like lighthouse.

This PR adds a method to the `Browser` class returning the port based on the wsEndpoint. The same thing could be achieved on the user side, but I feel it is a convenient helper method to have.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -21,6 +21,7 @@
   * [browser.disconnect()](#browserdisconnect)
   * [browser.newPage()](#browsernewpage)
   * [browser.pages()](#browserpages)
+  * [browser.remoteDebuggingPort()](#browserremotedebuggingport)
   * [browser.targets()](#browsertargets)
   * [browser.version()](#browserversion)
   * [browser.wsEndpoint()](#browserwsendpoint)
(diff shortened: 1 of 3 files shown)


Khady:
I signed it.

Khady:
The CI error doesn't seem to be related to my PR or I'm missing something?

aslushnikov:
@Khady Can you please elaborate more on the usecase? Why would you need a remote debugging port separately from ws endpoint?

aslushnikov:
> The CI error doesn't seem to be related to my PR or I'm missing something?

You're good, we have a few tests that flake occasionally.

Khady:
My default use case is to use lighthouse from javascript as documented there: https://github.com/GoogleChrome/lighthouse/tree/master/docs#using-programmatically

Corresponding code using the port: https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/index.js#L38

```javascript
function launchChromeAndRunLighthouse(url, flags = {}, config = null) {
  return chromeLauncher.launch().then(chrome => {
    flags.port = chrome.port;
    return lighthouse(url, flags, config).then(results =>
      chrome.kill().then(() => results));
  });
}
```

aslushnikov:
@Khady one reason I don't like exposing port is because it will prevent us from switching ws for the pipe connection in future.

Ideally, the pptr/LH interop should happen over any type of connection.

Khady:
Ok. I can survive without this PR for now. It is possible to extract the
information from `Browser.wsEndpoint`. So probably it's better to close
this PR and see how it goes in the future.

PR #1411, opened 2017-11-17
docs(api.md): add note for page.setCookie method

ghost:
Knowing that `name` and `value` are required is nice, but this doesn't tell us everything we need to know.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -973,6 +973,8 @@ Shortcut for [page.mainFrame.select()](#frameselectselector-values)
   - `sameSite` <[string]> `"Strict"` or `"Lax"`.
 - returns: <[Promise]>
 
+> **NOTE** For each cookie object, at least one of `url` and `domain` must be specified.
+
 #### page.setExtraHTTPHeaders(headers)
 - `headers` <[Object]> An object containing additional http headers to be sent with every request. All header values must be strings.
 - returns: <[Promise]>


docs/api.md
```diff
@@ -973,6 +973,8 @@ Shortcut for [page.mainFrame.select()](#frameselectselector-values)
   - `sameSite` <[string]> `"Strict"` or `"Lax"`.
 - returns: <[Promise]>
 
+> **NOTE** For each cookie object, at least one of `url` and `domain` must be specified.
```
aslushnikov, line 976: @JoelEinbinder why both `domain` and `url` are needed? Can we fix this in protocol?

JoelEinbinder:
Thanks for the PR!
By default, the current url of the page is set: See https://github.com/GoogleChrome/puppeteer/blob/master/lib/Page.js#L260

ghost:
Oh, I didn't know that (and well, I'll document it too).
But do you know what happens when you use `page.setCookie` *before* `page.goto`?
...

😬

JoelEinbinder:
It fails because `about:blank` can't have cookies. We should have a better error message in Puppeteer when you try to use setCookies on `about:blank` with no `url`. Are you interested in sending in a PR for that?

ghost:
I could do that, yeah. In another PR?
There should still be a note for that in the doc, right?

aslushnikov:
@JoelEinbinder shall we merge this?

JoelEinbinder:
Closing this for now. Feel free to send another PR for the `about:blank` note.

PR #1415, opened 2017-11-18
fix(Frame): postpone evaluations until execution context gets created

### LINKED ISSUE #827: frame.title() throws protocol error when used with framenavigated
Repro: 

```js

page.on('framenavigated', async frame => {
  console.log(await frame.title());
  browser.close();  
});

```

Produces:

```
Error: Protocol error (Runtime.evaluate): Cannot find context with specified id undefined
```

aslushnikov:
In Blink, frames don't necesserily have execution context all the time.
DevTools Protocol precisely reports this situation, which results in
Puppeteer's frame.executionContext() being null occasionally.

However, from puppeteer point of view every frame will have at least a
default executions context, sooner or later:
- frame's execution context might be created naturally to run frame's
  javascript
- if frame has no javascript, devtools protocol will issue execution
  context creation

This patch builds up on this assumption and makes `frame.executionContext()`
to be a promise.
As a result, all the evaluations await for the execution context to be created first.

Fixes #827, #1325 

BREAKING CHANGE: this patch changes `frame.executionContext()` method to return a promise.
To migrate onto a new behavior, await the context first before using it.

### DIFF
--- a/lib/ExecutionContext.js
+++ b/lib/ExecutionContext.js
@@ -110,6 +110,48 @@ class ExecutionContext {
   }
 }
 
+class PendingContext {
+  constructor() {
+    this._invocations = [];
+  }
+
+  /**
+   * @return {!Object}
+   */
+  createProxy() {
+    const invocations = this._invocations;
+    return new Proxy({}, {
+      get: function(target, propertyName) {
+        const property = Reflect.get(ExecutionContext.prototype, propertyName);
+        if (property && typeof property === 'function')
+          return recordInvocation.bind(null, propertyName);
+        return property;
+      }
+    });
+
+    function recordInvocation(methodName, ...args) {
+      const invocation = { methodName, args };
+      const promise = new Promise((resolve, reject) => {
+        invocation.resolve = resolve;
+        invocation.reject = reject;
+      });
+      invocations.push(invocation);
+      return promise;
+    }
+  }
+
+  /**
+   * @param {!ExecutionContext} context
+   */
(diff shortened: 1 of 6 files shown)


lib/ExecutionContext.js
```diff
@@ -19,12 +19,16 @@ const {helper} = require('./helper');
...
+   * @param {!Object} contextPayload
    * @param {function(*):!JSHandle} objectHandleFactory
    */
-  constructor(client, contextId, objectHandleFactory) {
+  constructor(client, contextPayload, objectHandleFactory) {
     this._client = client;
-    this._contextId = contextId;
+    this._contextId = contextPayload.id;
+
+    const auxData = contextPayload.auxData || {};
```
dgozman: .. || {isDefault: true}

utils/doclint/check_public_api/index.js
```diff
@@ -28,7 +28,6 @@ const EXCLUDE_CLASSES = new Set([
   'Multimap',
   'NavigatorWatcher',
   'NetworkManager',
-  'ProxyStream',
```
dgozman, line 31: Unrelated?

dgozman (approved):

lib/ExecutionContext.js
aslushnikov: Thanks, that's a nice one.

utils/doclint/check_public_api/index.js
aslushnikov: That's a drive-by

PR #1416, opened 2017-11-18
Updated block-images.js to reflect updated setRequestInterception

icarusmiles:
The /examples didn't have the adjustment from:

`await page.setRequestInterception(true);` 
to
`await page.setRequestInterceptionEnabled(true);`


icarusmiles:
I signed it!

icarusmiles:
It appears I was wrong, my mistake.

Issue #1300, opened 2017-11-07
Tracing: allow adding custom tracing categories


aslushnikov:
(Based on discussion in https://github.com/GoogleChrome/puppeteer/issues/854#issuecomment-342247715)

There should be a way to run tracing with custom categories.

@paulirish suggested having a "sampling" option with optional "hires" parameter.

```js
page.tracing.start({
  path: 'trace.json',
  sampling: true // could be also |false| and |'hires'|
});
```

As a valid alternative, I like letting users specify custom categories:

```js
page.tracing.start({
  path: 'trace.json',
  // Use these categories instead of the puppeteer defaults
  categories: ['disabled-by-default-v8.cpu_profiler.hires', /* ... other categories.. */ ]
});
```

slohmes:
Would implementation for this just involve adding something like
`    if (options.categories)
        categoriesArray.push(...options.categories);
`
to the start method in Tracing.js?

aslushnikov:
@slohmes yes. However, I'd rather use passed-in categories as a substitution for the default ones.

```js
const categories = options.categories || defaultCategories;
```

slohmes:
@aslushnikov Cool. I opened a PR, do you mind taking a look?https://github.com/GoogleChrome/puppeteer/pull/1439

aslushnikov:
For this to be closed: 
- a `jssampling` option should be added to enable js sampling.
The option should add a `disabled-by-default-v8.cpu_profiler.hires` category to the default sampling categories.

vaibhavtripathi:
@aslushnikov @paulirish I setup tracing as:
`page.tracing.start({path: 'tab.trace', screenshots: false, catgories: ['devtools.timeline']});`
I am still getting events from other categories. Is this a bug? Can this issue be repoened?

TomasHubelbauer:
I'm seeing the same issue. Even though `categories` is set, events from other categories still come. According to the PR, this I don't think should be the case. Has anything changed?

mathiasbynens:
@vaibhavtripathi There's a typo in your script (`catgories` instead of `categories`). If this issue still persists in the latest version of Puppeteer, please file a new bug with an example script that reproduces the problem.

levesquejf:
The `categories` parameter of `Tracing.start` is now deprecated. Probably need to use `traceConfig` instead (src: https://chromedevtools.github.io/devtools-protocol/tot/Tracing/#type-TraceConfig).

[...]

Issue #854, opened 2017-09-22
profiling javascript on webpage 


datasunny:
Hi all,
Is there a way to profile target page's javascript in puppeteer? say which functions are called and time spend for each function call?

Thanks.

ebidel:
The devtools protocol has these capabilities, but they're not exposed in puppeteer yet.
https://chromedevtools.github.io/devtools-protocol/tot/Profiler/.

Once https://github.com/GoogleChrome/puppeteer/issues/31 lands, you could access the protocol APIs directly. Puppeteer could also have a higher level API someday.

paulirish:
The page.tracing.start/stop methods can capture a trace. https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-tracing

with `disabled-by-default-v8.cpu_profiler` turned on you can get JS sampling profiler data in the trace. this category is [already enabled](https://github.com/GoogleChrome/puppeteer/blob/master/lib/Tracing.js) in puppeteer's settings..

(optionally also capturing `disabled-by-default-v8.cpu_profiler.hires`  will get higher resolution)

datasunny:
@paulirish Could you explain how to turn on 'disabled-by-default-v8.cpu_profiler' and 'disabled-by-default-v8.cpu_profiler.hires' in puppeteer? Thanks.

paulirish:
> Could you explain how to turn on 'disabled-by-default-v8.cpu_profiler' and 'disabled-by-default-v8.cpu_profiler.hires' in puppeteer?

Currently you cannot. https://github.com/GoogleChrome/puppeteer/blob/5e154dc835fcacdec6e680a695630352dae959dc/lib/Tracing.js#L36-L55

perhaps there should be a new option added like: `page.tracing.start({sampling:true})`.

paulins:
(file); await this._client.send

2017-11-06 21:15 GMT+01:00 Paul Irish <notifications@github.com>:

aslushnikov:
@datasunny @paulirish filed https://github.com/GoogleChrome/puppeteer/issues/1300 for this, looks like a good-first-bug

PR #1439, opened 2017-11-21
feat(Tracing): allow custom tracing categories

slohmes:
References #1300, #854

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1306,6 +1306,7 @@ await page.tracing.stop();
 - `options` <[Object]>
   - `path` <[string]> A path to write the trace file to. **required**
   - `screenshots` <[boolean]> captures screenshots in the trace.
+  - `categories` <[Array]<[string]>> specify custom categories to use instead of default. 
 - returns: <[Promise]>
 
 Only one trace can be active at a time per browser.
--- a/lib/Tracing.js
+++ b/lib/Tracing.js
@@ -37,12 +37,13 @@ class Tracing {
     console.assert(!this._recording, 'Cannot start recording trace while already recording trace.');
     console.assert(options.path, 'Must specify a path to write trace file to.');
 
-    const categoriesArray = [
+    const defaultCategories = [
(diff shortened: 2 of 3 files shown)


docs/api.md
```diff
@@ -1306,6 +1306,7 @@ await page.tracing.stop();
 - `options` <[Object]>
   - `path` <[string]> A path to write the trace file to. **required**
   - `screenshots` <[boolean]> captures screenshots in the trace.
+  - `categories` <[Array]<[string]>> specify custom categories to use instead of default.
```
aslushnikov, line 1309: is there any good resource to link to about tracing categories?

aslushnikov (approved):
thanks! cc perf gurus @a1ph @pavelfeldman

docs/api.md
paulirish: There isn't. Various tools just have a list hardcoded most of the time. You normally find what categories you want by reading devtools/lighthouse/webpagetest source. :)

paulirish (commented):
I'm very comfortable with this solution, though I'll offer another take:

1. keep this `categories` option for anyone who wants to take over and control it compeltely
2. add a `jssampling` option to `tracing.start()` for the basic usecase. This will enable hi-res js sampling profiler, a la the `screenshots` option.

slohmes:
@paulirish Sure, that sounds fine to me. For the sampling option, should there be any categories included besides the default categories and `disabled-by-default-v8.cpu_profiler.hires`?

paulirish:
@slohmes no that'd be all.   but you should hold off until we hear back on the API from @aslushnikov or @pavelfeldman

aslushnikov:
@slohmes `jssamping` lgtm

aslushnikov:
This looks good to merge. We can add `jssampling` in a separate PR to close #1300.

TomasHubelbauer:
FYI this doesn't seem to work the way it was intended anymore, see #1300. Setting `categories` explicitly to only a single value (in my case, `disabled-by-default-v8.cpu_profiler`), I still receive events from other (all?) categories.

PR #1445, opened 2017-11-21
feat(Page): implement page.connection method

### LINKED ISSUE #31: Expose raw protocol in the API
While using puppeteer there are still occasions where I want to use the raw protocol.
I'd certainly want to do this against the Page, and potentially the browser as well.

I want to send methods and get responses. Additionally, I want to listen for specific events.

Not sure of the right API and if it needs to be transactional as to not mixup state.

trentmwillis:
This patch implements a new page.connection method which exposes an interface for working with the raw protocol.

Fixes #31

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -48,6 +48,7 @@
   * [page.bringToFront()](#pagebringtofront)
   * [page.click(selector[, options])](#pageclickselector-options)
   * [page.close()](#pageclose)
+  * [page.connection()](#pageconnection)
   * [page.content()](#pagecontent)
   * [page.cookies(...urls)](#pagecookiesurls)
   * [page.deleteCookie(...cookies)](#pagedeletecookiecookies)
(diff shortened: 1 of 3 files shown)


aslushnikov:
Hey @trentmwillis, thanks for the PR.

We don't want to expose raw connection in this form: using raw protocol might easily conflict with the puppeteer internal state, so the two clients of the raw protocol (the user and the puppeteer) will fight each other.

We're exploring alternative approaches to support this scenario.

trentmwillis:
The concern is understandable, but just thinking aloud here: with coming support for multiple, concurrent clients on the protocol, couldn't this happen anyway? Maybe a warning could be given when using the raw protocol about potential collisions? Or maybe disallow certain domains/methods that Puppeteer relies on?

In any case, let me know if I can be of use or iterate on this. It is a feature that I think will unlock a lot more use cases for Puppeteer without having to continue increasing API surface area :)

aslushnikov:
> The concern is understandable, but just thinking aloud here: with coming support for multiple, concurrent clients on the protocol, couldn't this happen anyway? Maybe a warning could be given when using the raw protocol about potential collisions? Or maybe disallow certain domains/methods that Puppeteer relies on?

You're right, this can happen as well. For this reason, it might worth it to handle all the issues on the protocol level rather then in pptr.

> In any case, let me know if I can be of use or iterate on this. It is a feature that I think will unlock a lot more use cases for Puppeteer without having to continue increasing API surface area :)

Thanks for the help :) For now, I'll close the PR since we're not ready to follow this path. If you need the raw protocol somewhere, you can "hack" it with `page._client`.

trentmwillis:
Sounds good. I wanted to avoid using a private property, but since it seems to be the recommended workaround, I'll take that approach until something else comes along :)

PR #1447, opened 2017-11-21
feat(Frame): implement frame.content and frame.setContent methods

### LINKED ISSUE #754: [Question] Frame html content
Hi all,
Not sure if it's right place to ask question, how can I get frame content (HTML before and after JS rendering). I see page.content() method, but not such thing for frame. Thanks for your time!

trentmwillis:
This refactors the page.content and page.setContent methods to be defined on the Frame class. This allows access from the Page still but also on all frames.

Fixes #754

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -125,12 +125,14 @@
   * [frame.addScriptTag(options)](#frameaddscripttagoptions)
   * [frame.addStyleTag(options)](#frameaddstyletagoptions)
   * [frame.childFrames()](#framechildframes)
+  * [frame.content()](#framecontent)
   * [frame.evaluate(pageFunction, ...args)](#frameevaluatepagefunction-args)
   * [frame.executionContext()](#frameexecutioncontext)
   * [frame.isDetached()](#frameisdetached)
   * [frame.name()](#framename)
   * [frame.parentFrame()](#frameparentframe)
   * [frame.select(selector, ...values)](#frameselectselector-values)
+  * [frame.setContent(html)](#framesetcontenthtml)
   * [frame.title()](#frametitle)
   * [frame.url()](#frameurl)
   * [frame.waitFor(selectorOrFunctionOrTimeout[, options[, ...args]])](#framewaitforselectororfunctionortimeout-options-args)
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -628,6 +628,31 @@ describe('Page', function() {
     }));
   });
 
+  describe('Frame.setContent', function() {
```
aslushnikov: Are these tests exactly the same as the ones for the page.setContent?

In this case, I'd drop them to avoid repetition and save some time on test running.

aslushnikov (approved):
Thanks!

test/test.js
trentmwillis: Yes. My thinking is that since both the Frame and Page APIs are publicly exposed, it made sense to test them both. However, I could merge the tests together to reduce the number of individual tests that run.

Let me know what you think :)

aslushnikov: > Yes. My thinking is that since both the Frame and Page APIs are publicly exposed, it made sense to test them both.

This is reasonable. We have a [simple coverage](https://github.com/GoogleChrome/puppeteer/blob/master/CONTRIBUTING.md#public-api-coverage) for the public API to enforce this.

> However, I could merge the tests together to reduce the number of individual tests that run.

Thanks, I appreciate it!

trentmwillis:
I've updated the tests. Let me know if any other changes are needed.

test/test.js
```diff
@@ -2381,25 +2381,41 @@ describe('Page', function() {
...
-  describe('Page.setContent', function() {
-    const expectedOutput = '<html><head></head><body><div>hello</div></body></html>';
+  describe('Page/Frame.setContent', function() {
+    const pageExpectedOutput = '<html><head></head><body><div>hello</div></body></html>';
+    const frameExpectedOutput = '<html><head></head><body><div>world</div></body></html>';
     it('should work', SX(async function() {
       await page.setContent('<div>hello</div>');
-      const result = await page.content();
-      expect(result).toBe(expectedOutput);
+      const pageResult = await page.content();
```
aslushnikov: Ah sorry, I must've been unclear. I meant that there should be no need to change tests at all since `page.setContent` is just a shortcut for the `frame.setContent` (so it already tests the same codepaths).

trentmwillis: Okay, I'll defer to your judgment. My reasoning was `page.setContent` using `frame.setContent` is an implementation detail and thus without an explicit call to `frame.setContent` there is no guarantee its public API is actually being used and tested.

aslushnikov: Do you think that the api coverage linter gives insufficient guarantees? The coverage is runtime-based and makes sure that tests call every public method at least once.

trentmwillis: I don't know much about how it works, which is why I said I'd defer to you :) I've never worked with a project that has an API coverage mechanism like that. Seems reasonable though!

PR #1518, opened 2017-12-02
fix(examples): press enter instead of clicking search button

### LINKED ISSUE #1517: Page.click does not work as expected in slowMo mode
I'm opening google.com and type 'test' in the input box. Then click the submit button. Using slowMo:250 this leaves enough time for the autocomplete dropdown box to show up. The Page.click event then goes to the list element which is located at the position of the submit button. Therefor a different search result will show up.

This applies to both headless and non-headless mode.
```
  const browser = await puppeteer.launch({slowMo: 250});
  //const browser = await puppeteer.launch({headless: false, slowMo: 250});
  await page.type('input[name=q]', 'test');
  await page.click('input[type=submit]');
  await page.screenshot({path: 'google.png'});
  await browser.close();
```

JoelEinbinder:
Fixes #1517

### DIFF
--- a/examples/search.js
+++ b/examples/search.js
@@ -16,7 +16,7 @@
 
 'use strict';
 
-const puppeteer = require('..');
+const puppeteer = require('puppeteer');
 
 (async() => {
 
--- a/examples/search.js
+++ b/examples/search.js
@@ -16,7 +16,7 @@
 
 'use strict';
 
-const puppeteer = require('puppeteer');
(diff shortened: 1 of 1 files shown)


aslushnikov:
@JoelEinbinder does it work reliably for 100 consecutive runs? I have a vague memory that google page was tricky.

JoelEinbinder:
It looks like we are removing the search example.

PR #1519, opened 2017-12-02
test: drop jasmine test runner

aslushnikov:
This patch introduces a tiny test runner to run puppeteer tests.
The test runner is self-container and allows parallel test execution.
It will also allow us to split tests into multiple files if necessary.

Comparing to the jasmine, the testrunner supports parallel execution, properly
handles "unhandled promise rejection" event and signals.

Comparing to ava/jest, the testrunner doesn't run multiple node processes,
which makes it simpler but sufficient for our goals.


test/test.js
```diff
@@ -65,23 +58,18 @@ else
...
-  specStarted: result => specName = result.fullName
-});
+const timeout = process.env.DEBUG_TEST || slowMo ?  0 : 10 * 1000;
+
+const {TestRunner, Reporter, Matchers}  = require('../utils/testrunner/');
+const runner = new TestRunner({timeout});
+const reporter = new Reporter(runner);
+
+const {describe, xdescribe, fdescribe} = runner;
+const {it, fit, xit} = runner;
```
JoelEinbinder: we should probably have an eslint-disable here for the unused variable rule.

test/test.js
```diff
@@ -65,23 +58,18 @@ else
...
 
-// Hack to get the currently-running spec name.
-let specName = null;
-jasmine.getEnv().addReporter({
-  specStarted: result => specName = result.fullName
-});
+const timeout = process.env.DEBUG_TEST || slowMo ?  0 : 10 * 1000;
+
+const {TestRunner, Reporter, Matchers}  = require('../utils/testrunner/');
+const runner = new TestRunner({timeout});
```
JoelEinbinder, line 64: The main benefit of the test runner is that it runs tests in parallel right? This patch looks fine, but I don't think we should land it until we are sure that the tests run correctly in parallel on the bots.

JoelEinbinder (approved):

test/test.js
JoelEinbinder: Let's not run tests in parallel.

PR #1526, opened 2017-12-03
Close #1492, fix screenshot mimetype definition when type and path are provided

### LINKED ISSUE #1492: page.screenshot.type not respected
When taking a screenshot, even though the type is specified as "png", puppeteer uses the file extension of the output file to determine the mime type.

### Environment

- Windows 7
- Puppeteer NPM version 0.13.0
- Node version 7.10.0

### Steps to reproduce

Create a basic screenshot script such as this (example.js):
```
(async () => {
	await page.goto("https://google.co.uk/", {waitUntil: "load"});
	await page.screenshot({
		path: "C:/img.tmp",
		type: "png"
	});
	await browser.close();
})();
```

Run the script on the command line:
`$ node example.js`

The program should take a PNG screenshot of https://google.co.uk/ and save it to "C:/img.tmp".

**What happens instead?**
`(node:65264) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id:2): Assertion error: Unsupported screenshot mime type: application/octet-stream`

It seems to be trying to determine the output type from the file extension regardless of the type setting. This may prevent usage of output targets like /dev/stdout.

MrMitch:
This fixes https://github.com/GoogleChrome/puppeteer/issues/1492 by only trying to define the mimetype from the `options.path` value if the `options.type` is not provided.

This is useful, for example, in situation where `options.path` and `options.type` are provided and `options.path` exists but is an empty file, with no extension. With the current code, the mimetype deduced is `application/octet-stream` and this raises an error. With this fix, the `options.type` value is used, removing the need to deduce the mimetype from the file path.

Now the screenshot's mime-type is defined using : 

1. `options.type` if it is provided, whether or not `options.path` is provided (**this is what this PR introduces**)
2. `options.path` if `options.type` is not provided, by inferring the mime-type using `mime.lookup` (_this was not changed_)
3. `png` as a default value if neither `options.type` nor `options.path` is provided (_this was not changed_)

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -614,19 +614,18 @@ class Page extends EventEmitter {
    */
   async screenshot(options = {}) {
     let screenshotType = null;
-    // try to define screenshotType from options.path if options.type isn't provided
-    if (!options.type) {
-      console.assert(options.path, 'Expected options.path or options.type to define screenshot type');
+    // options.type takes precedence over inferring the type from options.path
+    // because it may be a 0-length file with no extension created beforehand (i.e. as a temp file).
+    if (options.type) {
+      console.assert(options.type === 'png' || options.type === 'jpeg', 'Unknown options.type value: ' + options.type);
+      screenshotType = options.type;
+    } else if (options.path) {
       const mimeType = mime.lookup(options.path);
       if (mimeType === 'image/png')
         screenshotType = 'png';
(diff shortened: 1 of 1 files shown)


MrMitch:
@googlebot I signed it!

aslushnikov (approved):

aslushnikov:
Thank you!

Issue #594, opened 2017-08-29
EventEmitter memory leak detected. 11 exit listeners added


selfrefactor:
This is the full error message:
`MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 exit listeners added. Use emitter.setMaxListeners() to increase limit`

I run Node v8.1.2 with tip-of-tree Puppeteer

The code is written with `Promise.all`, but even if I run `work()` sequentially the same error appears.

```
const puppeteer = require('puppeteer')
const R = require('rambda')

const resolution = {
  x : 1920,
  y : 1080,
}

const args = [
  '--disable-gpu',
  `--window-size=${ resolution.x },${ resolution.y }`,
  '--no-sandbox',
]

const work = async () => {
  const browser = await puppeteer.launch({
    headless     : true,
    handleSIGINT : false,
    args         : args,
  })
  const page = await browser.newPage()

  const url = 'https://ilearnsmarter.com/learning-meme'

  await page.setViewport({
    width  : resolution.x,
    height : resolution.y,
  })

  await page.goto(url, { waitUntil : 'networkidle' })
  await browser.close()
}

const fn = async () => {
  const promised = R.range(0, 12).map(() => work())
  await Promise.all(promised)

  console.log('DONE')
}

fn()
```

selfrefactor:
One possible solution is using child process:

```
// workChild.js
const puppeteer = require('puppeteer')

const resolution = {
  x : 1920,
  y : 1080,
}

const args = [
  '--disable-gpu',
  `--window-size=${ resolution.x },${ resolution.y }`,
  '--no-sandbox',
]

const work = async () => {
  const browser = await puppeteer.launch({
    headless     : true,
    handleSIGINT : false,
    args         : args,
  })
  const page = await browser.newPage()

  const url = 'https://ilearnsmarter.com/learning-meme'

  await page.setViewport({
    width  : resolution.x,
    height : resolution.y,
  })

  await page.goto(url, { waitUntil : 'networkidle' })

  const result = await page.content()
  await browser.close()

  return result.length
}

work()
  .then(result => {
    process.send(result)
    process.exit()
  })
```

```
const { fork } = require("child_process")
const R = require('rambda')

const work = () => new Promise(resolve=>{
  const child = fork("workChild.js", [], {
    cwd : __dirname,
    env : process.env,
  })

  child.on("message", result => {
    return resolve(result)
  })
})

const fn = async () => {
[...]

PR #1528, opened 2017-12-04
Unlimited browser instances and ^c fix.

dusty-commits:
fix(Launcher): Fix limit on browser instances and needing to issue ^c twice to exit. 

This fixes 2 quirks.

1. If creating 11 instances of a Browser (i.e. puppeteer.launch()), we will receive the memory leak warning found in #594. This solution is much safer than setting the maxListeners on `process` and allows unlimited browser instances.

2. This also closes the node process when hitting ^c. In the past, we had to do this command twice to actually exit our program. This ensures it exits with the correct exit code after only issuing the command once. 

[Further explanation](https://github.com/GoogleChrome/puppeteer/issues/1481)


aslushnikov:
Hey @vhesener, thank you for the PR!

We had this implemented once and for some reason didn't land it. @JoelEinbinder do you remember why?

lib/Launcher.js
```diff
@@ -190,6 +191,31 @@ class Launcher {
...
+    const processListeners = [ helper.addEventListener(process, 'exit', exitHandler.bind(this)) ];
+    processListeners.push(helper.addEventListener(process, 'SIGINT', sigHandler('SIGINT', 2).bind(this)));
+    processListeners.push(helper.addEventListener(process, 'SIGTERM', sigHandler('SIGTERM', 15).bind(this)));
+    processListeners.push(helper.addEventListener(process, 'SIGHUP', sigHandler('SIGHUP', 1).bind(this)));
+
+    function sigHandler(eventName, exitCode) {
+      return function() {
+        helper.removeEventListeners(processListeners);
+        this.emit(eventName);
+        process.exit(exitCode);
```
aslushnikov, line 207: What if client code has a custom SIGINT handler that doesn't want to process.exit right away?

dusty-commits: Fair questions. 

1. Isn't this why we allow them to turn it off via [handleSIGINT](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions)?
2. Can they not use [prependListener](https://nodejs.org/api/events.html#events_emitter_prependlistener_eventname_listener) if they want both?

JoelEinbinder:
I think in all of our use cases, the browser was the only thing keeping node alive and so closing it caused node to exit. So I never noticed that it wasn't closing node.

However we should wait for the `browser.close()` command to finish in Chrome before exiting in node. These are theoretically graceful exits.

dusty-commits:
Interesting. Ok I think I understand more. I have a couple follow up questions:

The SIGINT event always seemed to call [forceKillChrome](https://github.com/vhesener/puppeteer/blob/80eeabb83d3570760d5ebfe90295fc95da6bf073/lib/Launcher.js#L160). It makes sense to try to gracefully close the browser instance, but then we should probably try to call [killChrome](https://github.com/vhesener/puppeteer/blob/80eeabb83d3570760d5ebfe90295fc95da6bf073/lib/Launcher.js#L149) instead, right? I'm sure there is a reason to call forceKillChrome here instead but I'm not sure of the history of this logic.

Does this have anything to do with the fact that SIGINT handlers do something unique on non-Windows systems:
[Signal Events](https://nodejs.org/api/process.html#process_signal_events)
>  If one of these signals has a listener installed, its default behavior will be removed (Node.js will no longer exit)

aslushnikov:
> The SIGINT event always seemed to call forceKillChrome. It makes sense to try to gracefully close the browser instance, but then we should probably try to call killChrome instead, right? I'm sure there is a reason to call forceKillChrome here instead but I'm not sure of the history of this logic.

The reason is that we want to guarantee a shutdown, otherwise chrome instances will be orphaned and will never be closed. The graceful shutdown doesn't guarantee closing in certain scenarios, so we force kill browser.

> Does this have anything to do with the fact that SIGINT handlers do something unique on non-Windows systems:

The SIGINT is not sent on windows; we only have a simulated SIGHUP on win.

dusty-commits:
Ok, so since SIGTERM and SIGHUP need to gracefully close via a Promise, these probably aren't suited well for an EventEmitter, agree? I'm not entirely sure the graceful browser closing is happening in the current logic, but I'm interested in helping solve that properly if needed too. 

What do you suggest? I can think of a few ways to go, but let me know if you have any other suggestions:

- Close the PR and maybe attack the issue at a later time.
- Stick with current PR which will still send the Browser.close message but not wait for the response when calling process.exit() (possibly ungraceful, but might be similar to current logic).
- Remove the process.exit() call. This would resolve the first quirk of having 11 instances, but the second quirk of hitting ^c twice for SIGINT would remain.
- Only perform process.exit() call for SIGINT. This will allow the other events to attempt a graceful Browser.close.

aslushnikov:
As you suggested, let's close this PR. It might not be needed to launch multiple browsers once browser contexts are introduced.

For now, I think it's perfectly fine to do `setMaxListeners` on `process`, as node.js platform suggests.

Khady:
> As you suggested, let's close this PR. It might not be needed to launch multiple browsers once browser contexts are introduced.
> 

Even for performance reasons? Like it could be more interesting to have multiple browsers with one tab rather than one browser with multiple tab to be sure the tab won't compete for ressources?

aslushnikov:
> Even for performance reasons? Like it could be more interesting to have multiple browsers with one tab rather than one browser with multiple tab to be sure the tab won't compete for ressources?

@Khady browser contexts are very cheap to instantiate; way cheaper then launching a new browser instance. Regarding the tabs fighting for resources, multiple tabs inside single browser still should be way cheaper and performant then multiple browsers with single tab.

PR #1531, opened 2017-12-04
test: migrate test.js to support concurrent test execution

aslushnikov:
This patch migrates tests so that they can be run concurrently.
- By default, tests still run in one thread.
- To run tests in 4 parallel threads, run `node test/test.js -j 4` or `npm run unit -- -j 4`
- Environment variable `PPTR_PARALLEL_TESTS` could be set to override default parallelization

Every test gets passed in a `state`. State is set up in the `beforeAll` and `beforeEach` callbacks,
and should be teared down in the `afterAll` and `afterEach` callbacks.

By default, state has a `parallelIndex` variable initialized that defines the thread index that runs the execution.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -242,20 +242,20 @@ describe('Puppeteer', function() {
 });
 
 describe('Page', function() {
-  beforeAll(async state  => {
+  beforeAll(async state => {
     state.browser = await puppeteer.launch(defaultBrowserOptions);
   });
 
-  afterAll(async state  => {
+  afterAll(async state => {
     await state.browser.close();
     state.browser = null;
   });
 
-  beforeEach(async state  => {
+  beforeEach(async state => {
     state.page = await state.browser.newPage();
   });
 
-  afterEach(async state  => {
+  afterEach(async state => {
     await state.page.close();
     state.page = null;
   });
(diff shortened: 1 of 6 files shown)


test/test.js
```diff
@@ -208,63 +213,59 @@ describe('Puppeteer', function() {
...
-      const restoredPage = pages.find(page => page.url() === PREFIX + '/frames/nested-frames.html');
+      const restoredPage = pages.find(page => page.url() === urls.PREFIX + '/frames/nested-frames.html');
       expect(FrameUtils.dumpFrames(restoredPage.mainFrame())).toBeGolden('reconnect-nested-frames.txt');
       expect(await restoredPage.evaluate(() => 7 * 8)).toBe(56);
       await browser.close();
     }));
   });
   describe('Puppeteer.executablePath', function() {
-    it('should work', SX(async function() {
+    it('should work', SX(async({urls}) => {
```
JoelEinbinder: Let's remove SX in this patch.

test/test.js
```diff
@@ -72,65 +67,73 @@ const {expect} = new Matchers({
...
+  rm(OUTPUT_DIR);
+
+beforeAll(async state  => {
   const assetsPath = path.join(__dirname, 'assets');
-  server = await SimpleServer.create(assetsPath, PORT);
-  httpsServer = await SimpleServer.createHTTPS(assetsPath, HTTPS_PORT);
-  if (fs.existsSync(OUTPUT_DIR))
-    rm(OUTPUT_DIR);
-}));
+  const port = 8907 + state.parallel * 2;
```
JoelEinbinder: index

JoelEinbinder (approved):

PR #1554, opened 2017-12-08
refactor: move ChromiumDownloader under lib/

aslushnikov:
This patch:
- renames `ChromiumDownloader` into `Downloader` (this is in
  preparation for different products download)
- moves `Downloader` from `utils/` to `lib/`. This unifies all of the
  production-critical code in the `lib/` and would make our node6
  transpilation easier.

Drive-by: make `Downloader` a regular class.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -23,7 +23,6 @@ const {Browser} = require('./Browser');
 const readline = require('readline');
 const fs = require('fs');
 const {helper} = require('./helper');
-// @ts-ignore
 const ChromiumRevision = Downloader.defaultRevision();
 
 const mkdtempAsync = helper.promisify(fs.mkdtemp);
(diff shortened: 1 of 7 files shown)


lib/Launcher.js
```diff
@@ -17,14 +17,14 @@ const os = require('os');
...
 const removeFolder = require('rimraf');
 const childProcess = require('child_process');
-const Downloader = require('../utils/ChromiumDownloader');
+const Downloader = require('./Downloader');
 const {Connection} = require('./Connection');
 const {Browser} = require('./Browser');
 const readline = require('readline');
 const fs = require('fs');
 const {helper} = require('./helper');
 // @ts-ignore
```
JoelEinbinder: remove ts-ignore

test/test.js
```diff
@@ -20,7 +20,7 @@ const {helper} = require('../lib/helper');
 if (process.env.COVERAGE)
   helper.recordPublicAPICoverage();
 console.log('Testing on Node', process.version);
-const puppeteer = require('..');
```
JoelEinbinder: Let's leave it how it was.

JoelEinbinder (approved):

PR #1562, opened 2017-12-08
feat(install): build node6 support when installing from github

aslushnikov:
This patch starts building node6 support if puppeteer is installed directly from
github and is run under Node 6.


JoelEinbinder:
We should be able to remove `yarn run build` from the .travis.yml. That will cover this with tests.

JoelEinbinder (approved):

PR #1564, opened 2017-12-09
chore: remove search.js example

### LINKED ISSUE #1557: Example: search.js should demo something other than google.com

aslushnikov:
The example is violating the TOS of google.com

Fix #1557


JoelEinbinder (approved):

ebidel:
LGTM. I was thinking we'd just replace the demo with something similar that shows grabbing page content. This example is also nice b/c it shows a couple of the APIs together (keyboard input and click).

ebidel (approved):

aslushnikov:
@ebidel yeah we'll probably make another one with our own protocol viewer

schelkun:
What about DuckDuckGo?

ebidel:
https://chromedevtools.github.io/devtools-protocol/ is the protocol viewer @aslushnikov is referring to. I think it's probably safer to use that for an example than a site that our team does not maintain. It has an input with search results...so it would be very similar to the original demo.

PR #1565, opened 2017-12-09
Add `opacity: 0` to `waitForSelector`’s hidden element check

benschwarz:
There are many, many ways someone can hide an element from view, but it seems odd that the only stylistic check done here is on the `visibility` property. 

This change adds checks for `display: none` and `opacity: 0;`. We're very likely to see these methods of hiding something from view in plenty of codebases. 

Ideally, this code would follow whatever it is that selenium does for 'visibility checks', but I'd rather not start bike shedding that on this PR.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -577,7 +577,7 @@ class Frame {
       if (!waitForVisible && !waitForHidden)
         return true;
       const style = window.getComputedStyle(node);
-      const isVisible = style && style.visibility !== 'hidden' && hasVisibleBoundingBox();
+      const isVisible = style && style.visibility !== 'hidden' && style.display !== 'none' && style.opacity !== '0' && hasVisibleBoundingBox();
       return (waitForVisible === isVisible || waitForHidden === !isVisible);
 
       /**
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -577,7 +577,7 @@ class Frame {
       if (!waitForVisible && !waitForHidden)
         return true;
       const style = window.getComputedStyle(node);
-      const isVisible = style && style.visibility !== 'hidden' && style.display !== 'none' && style.opacity !== '0' && hasVisibleBoundingBox();
+      const isVisible = style && style.visibility !== 'hidden' && style.opacity !== '0' && hasVisibleBoundingBox();
       return (waitForVisible === isVisible || waitForHidden === !isVisible);
 
       /**
--- a/test/test.js
+++ b/test/test.js
@@ -809,6 +809,16 @@ describe('Page', function() {
       expect(await waitForSelector).toBe(true);
       expect(divHidden).toBe(true);
     }));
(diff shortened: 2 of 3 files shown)


lib/FrameManager.js
```diff
@@ -577,7 +577,7 @@ class Frame {
       if (!waitForVisible && !waitForHidden)
         return true;
       const style = window.getComputedStyle(node);
-      const isVisible = style && style.visibility !== 'hidden' && hasVisibleBoundingBox();
+      const isVisible = style && style.visibility !== 'hidden' && style.display !== 'none' && style.opacity !== '0' && hasVisibleBoundingBox();
```
ebidel: ` style.display !== 'none'` should be covered by the `hasVisibleBoundingBox()` check. If the element (or any ancestor) is `display: none` then the bounding rect will be 0 width/height.

benschwarz:
@ebidel I guess just `opacity: 0`, then?

ebidel:
Yea

benschwarz:
Sweet, I'll add a test to cover. Anything else required to merge?

ebidel:
Docs too. https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md

test/test.js
```diff
@@ -809,6 +809,16 @@ describe('Page', function() {
       expect(await waitForSelector).toBe(true);
       expect(divHidden).toBe(true);
     }));
+    it('hidden should wait for opacity: 0', SX(async function() {
+      let divHidden = false;
+      await page.setContent(`<div style='opacity: 1;'></div>`);
+      const waitForSelector = page.waitForSelector('div', {hidden: true}).then(() => divHidden = true);
+      await page.waitForSelector('div'); // do a round trip
+      expect(divHidden).toBe(false);
+      await page.evaluate(() => document.querySelector('div').style.setProperty('display', '0'));
```
ebidel, line 818: `document.querySelector('div').style.opacity = '0';`

test/test.js
```diff
@@ -809,6 +809,16 @@ describe('Page', function() {
       expect(await waitForSelector).toBe(true);
       expect(divHidden).toBe(true);
     }));
+    it('hidden should wait for opacity: 0', SX(async function() {
```
aslushnikov, line 812: Tests have slightly changed; the `SX` function is no longer needed, and the `page` is passed to the test function itself.
```js
it('hidden should wait for opacity: 0', async function({page}) {
  // ...
});
```

aslushnikov (commented):
@benschwarz: do you still want to land this CL?

test/test.js
benschwarz: Thanks for the heads up.

benschwarz:
@aslushnikov Yep, I've just had a busy week and didn't get to come back to this

aslushnikov:
Closing this for now. @benschwarz please feel free to send this again as you have time to get back to this!

PR #1566, opened 2017-12-10
Throw exception when set cookie in 'about:blank' page

jihchi:
# DEPRECATED
Please refer to #1567 

---

Original from https://github.com/GoogleChrome/puppeteer/pull/1411#issuecomment-345423707

When you set cookie in 'about:blank' page, you'll get following exception:

```
AssertionError [ERR_ASSERTION]: Blank page can not have cookie "<cookie_name>"
```


jihchi:
I signed it!

PR #1571, opened 2017-12-11
fix(ElementHandle): throw when trying to focus non-focusable element

### LINKED ISSUE #1486: page.type with disabled input and focus on a different input
I encountered some unexpected behavior when I was using page.type on a disabled input field, while the focus was on a different field.

**Example:**
Tested with v0.13.0 on Ubuntu 16.04 and node v0.8.4.

```html
<html>
<body>
	<input type="text" id="username" value="john" disabled="disabled" />
	<input type="text" id="password" autofocus />
</body>
</html>
```

```javascript
'use strict';

(async () => {
  let username, password;

    headless: false
  });
  await page.goto('file:///tmp/test.html');

  await page.type('#username', 'john.doe');
  await page.type('#password', '123456');

  username = await page.evaluate(() => {
    return document.querySelector('#username').value;
  });
  console.log(username);

  password = await page.evaluate(() => {
    return document.querySelector('#password').value;
  });
  console.log(password);

  await browser.close();
})();
```

**Expected Result:**
I expected that the username remains unchanged and the password contains _123456_. Or maybe an Error because I'm trying to type into a disabled input field.

**Actual Result:**
Instead the username remains unchanged and the password contains both the username and the password.

Right now I avoid this issue by checking the disabled attribute before using page.type, which isn't a bad idea anyway.

PS: Thank you for Puppeteer!

aslushnikov:
This patch starts throwing errors when trying to focus non-focusable
element.

Fixes #1486

### DIFF
--- a/lib/ElementHandle.js
+++ b/lib/ElementHandle.js
@@ -96,7 +96,11 @@ class ElementHandle extends JSHandle {
   }
 
   async focus() {
-    await this.executionContext().evaluate(element => element.focus(), this);
+    await this.executionContext().evaluate(element => {
+      element.focus();
+      if (element.ownerDocument.activeElement !== element)
+        throw new Error('Failed to focus: element is not focusable!');
+    }, this);
   }
 
   /**
--- a/test/test.js
+++ b/test/test.js
@@ -1998,6 +1998,13 @@ describe('Page', function() {
       await textarea.press('a', {text: 'y'});
       expect(await page.evaluate(() => document.querySelector('textarea').value)).toBe('f');
     }));
+    it('should not focus non-focusable elements', SX(async function() {
+      await page.setContent('<input disabled></input>');
+      const input = await page.$('input');
+      let error = null;
+      await input.focus().catch(e => error = e);
+      expect(error.message).toContain('not focusable');
+    }));
     it('should send a character with sendCharacter', SX(async function() {
       await page.goto(PREFIX + '/input/textarea.html');
       await page.focus('textarea');


lib/ElementHandle.js
```diff
@@ -96,7 +96,11 @@ class ElementHandle extends JSHandle {
   }
 
   async focus() {
-    await this.executionContext().evaluate(element => element.focus(), this);
+    await this.executionContext().evaluate(element => {
+      element.focus();
+      if (element.ownerDocument.activeElement !== element)
```
JoelEinbinder, line 101: `element.focus()` calls the event listeners, which may redirect focus to some other element.

aslushnikov: Indeed, thank you

JoelEinbinder: Maybe you can add an event listener for focus, and see if a focus event was fired after calling .focus.

JoelEinbinder: Adding an event listener to element.ownerDocument with useCapture true only fails if there is a previous event listener on the document that calls stopImmediatePropagation and doesn't call preventDefault.

Or we can modify [`DOM.focus`](https://aslushnikov.github.io/vanilla-protocol-viewer/#DOM.focus) to return true if the focus was applied successfully.

aslushnikov: > Or we can modify DOM.focus to return true if the focus was applied successfully.

I like this

aslushnikov:
Closing this for now since the protocol approach would be ideal.

Issue #291, opened 2017-08-16
Video Playback Fails


benjamingr:
Video playback does not work with puppeteer.

Sample code:

```js
const browser = await puppeteer.launch({headless: false}); // reproduces with headless: true too
const page = await browser.newPage();
await page.goto('https://video-dev.github.io/hls.js/demo/');
```

Expected result: big buck bunny plays at `https://video-dev.github.io/hls.js/demo/`
Result: no video playback with the error: 

> your Browser does not support MediaSourceExtension / MP4 mediasource toggle playback controls

Garbee:
I think is is because the MP4 codec is under a paid license fee. Which means it isn't distributed with Chromium but only within Google Chrome.

aslushnikov:
mp4 is missing in chromium: https://www.chromium.org/audio-video

Garbee:
https://www.quirksmode.org/html5/tests/video.html  as well to test. MP4 fails, but OGG Theora and WebM both play fine with Chromium.

benjamingr:
Thanks for the amazingly fast response!

I'll resolve this by using Chrome instead as suggested in https://github.com/GoogleChrome/puppeteer/issues/288

Garbee:
You'd need to be on Chrome Canary to get accurate enough runnings at this point. Puppeteer relies on some bleeding-edge changes on Chromium internally. Which have yet to get out to even Dev Channel. So while Chrome may work, it isn't currently guaranteed to work.

benjamingr:
Thanks, that's probably something you're going to want to document.

Garbee:
It's noted under the [installation section](https://github.com/GoogleChrome/puppeteer#installation) of the README.

PR #1572, opened 2017-12-11
docs(README): add note on features not supported by Chromium

ithinkihaveacat:
Developers are frequently surprised that Puppeteer does not support video (see e.g. https://github.com/GoogleChrome/puppeteer/issues/291 and [other issues mentioning "video"](https://github.com/GoogleChrome/puppeteer/issues?utf8=%E2%9C%93&q=is%3Aissue+video), many of which are related).

The documentation does explain how to use Chrome instead of Chromium, but I couldn't find anywhere that links this "fix" to the problem of audio/video not working (the only reason to do this?).


README.md
```diff
@@ -219,3 +219,10 @@ other frameworks could adopt Puppeteer as their foundational layer.
...
 
 We've reached out to a number of these projects to see if there are opportunities for collaboration, and we're happy to do what we can to help.
+
+#### Q: What features does Puppeteer not support?
+
+You may find that Puppeteer does not behave as expected when controlling pages that incorporate audio and video. (For example, [video playback/screenshots is likely to fail](https://github.com/GoogleChrome/puppeteer/issues/291).) There are two reasons for this:
+
+* Puppeteer is bundled with Chromium--not Chrome--and so by default, it inherits all of [Chromium's video-related limitations](https://www.chromium.org/audio-video). This means that Puppeteer does not understand the MP4 container format, and also does not support MP3, AAC or H.264. (However, it is possible to force Puppeteer to use a separately-installed version Chrome instead of Chromium via the [`executablePath` option to `puppeteer.launch`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions). You should only use this configuration if you need Chrome's audio/video support.)
```
ebidel: Thanks for adding this. mp3/mp4 support was added to chromium 2 days ago. https://bugs.chromium.org/p/chromium/issues/detail?id=746579

We should probably leave them out since we'll pick them up the next time we rev chromium.

ebidel (changes requested):

README.md
ithinkihaveacat: Sure, that's good news! Updated to remove mention of MP3/MP4.

README.md
```diff
@@ -219,3 +219,10 @@ other frameworks could adopt Puppeteer as their foundational layer.
 The past few months have brought [several new libraries for automating headless Chrome](https://medium.com/@kensoh/chromeless-chrominator-chromy-navalia-lambdium-ghostjs-autogcd-ef34bcd26907). As the team authoring the underlying DevTools Protocol, we're excited to witness and support this flourishing ecosystem.
 
 We've reached out to a number of these projects to see if there are opportunities for collaboration, and we're happy to do what we can to help.
+
+#### Q: What features does Puppeteer not support?
+
+You may find that Puppeteer does not behave as expected when controlling pages that incorporate audio and video. (For example, [video playback/screenshots is likely to fail](https://github.com/GoogleChrome/puppeteer/issues/291).) There are two reasons for this:
```
ebidel, line 225: Sorry to be picky. I think we can ditch the link now b/c it's discussing lack of mp3/4 support, which is not true anymore :)

README.md
```diff
@@ -219,3 +219,10 @@ other frameworks could adopt Puppeteer as their foundational layer.
...
+
+* Puppeteer is bundled with Chromium--not Chrome--and so by default, it inherits all of [Chromium's media-related limitations](https://www.chromium.org/audio-video). This means that Puppeteer does not support AAC or H.264. (However, it is possible to force Puppeteer to use a separately-installed version Chrome instead of Chromium via the [`executablePath` option to `puppeteer.launch`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions). You should only use this configuration if you need Chrome's audio/video support.)
```
ebidel: Puppeteer does not support AAC or H.264 -> Puppeteer does not support licensed formats such as AAC or H.264.

ebidel: if you need Chrome's audio/video support. -> if you need an official release of Chrome that supports these media formats.

ebidel (changes requested):

aslushnikov:
@ithinkihaveacat do you want to proceed with this?

README.md
ithinkihaveacat: Is MP4 supposed to work now with GoogleChrome/master? #291 now fails in a slightly different way (no error message), but the video still doesn't play.

ithinkihaveacat:
@aslushnikov Yes, sorry, I just just waiting to see what forms of video Chromium/Puppeteer supports with the latest update, but I'm not sure there's any change--is there supposed to be?

aslushnikov:
@ithinkihaveacat I don't think anything has changed.

README.md
ithinkihaveacat: Changed in 5b0d01878eab40d59ff252a837195087322b9ed1.

ithinkihaveacat: Changed in b4ee320b517d1fee3baba3bbeeb4003a81eb5c36.

ithinkihaveacat:
@aslushnikov Ok, I think this is ready now.

@ebidel wanted the bit that mentions video playback being likely to fail removed, but in my tests I don't think Puppeteer's behaviour has changed much here, so I've left it in.

aslushnikov (approved):

README.md
aslushnikov: Indeed, looks like mp4 is still not supported by chromium atm.

PR #1574, opened 2017-12-11
docs(README.md): Update docs to more clearly differentiate between Ch…

### LINKED ISSUE #1516: Clarify whether this uses Chrome or Chromium
### Steps to reproduce

* Puppeteer version: 0.13
* Platform / OS version: Mac OSX 10.12.6
* URLs (if applicable):

The documentation is rather confusing - it is hard to tell if you are using Chromium or Chrome (or both?)

For example:
"Puppeteer is a Node library which provides a high-level API to control headless Chrome over the DevTools Protocol. It can also be configured to use full (non-headless) Chrome."

then
" When you install Puppeteer, it downloads a recent version of Chromium (~71Mb Mac, ~90Mb Linux, ~110Mb Win) that is guaranteed to work with the API"

then
"Puppeteer launches Chromium in headless mode. To launch a full version of Chromium, set the 'headless' option when launching a browser:"

then
"By default, Puppeteer downloads and uses a specific version of Chromium so its API is guaranteed to work out of the box. To use Puppeteer with a different version of Chrome, pass in the executable's path when creating a Browser instance:"

Within the course of two sentences, it changes from Chromium to Chrome.

Another example:
"Puppeteer works only with Chrome. However, many teams only run unit tests with a single browser (e.g. PhantomJS). In non-testing use cases, Puppeteer provides a powerful but simple API because it's only targeting one browser that enables you to rapidly develop automation scripts.

Puppeteer uses the latest versions of Chromium."

Does Puppeteer use Chrome, or Chromium?

Please clarify. I think the docs should also be updated to be consistent in the use of 'chrome' vs 'chromium'

FWIW, when I install it on a mac, it installs Chromium into .local_chromium.

karloskalcium:
…romium and Chrome

This patch clarifies the documentation to more clearly note that Puppeteer works with both Chromium (by default)
and Chrome.

Fixes #1516

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ###### [API](docs/api.md) | [FAQ](#faq) | [Contributing](https://github.com/GoogleChrome/puppeteer/blob/master/CONTRIBUTING.md)
 
-> Puppeteer is a Node library which provides a high-level API to control [headless](https://developers.google.com/web/updates/2017/04/headless-chrome) Chrome over the [DevTools Protocol](https://chromedevtools.github.io/devtools-protocol/). It can also be configured to use full (non-headless) Chrome.
+> Puppeteer is a Node library which provides a high-level API to control [headless](https://developers.google.com/web/updates/2017/04/headless-chrome) Chrome or Chromium over the [DevTools Protocol](https://chromedevtools.github.io/devtools-protocol/). It can also be configured to use full (non-headless) Chrome or Chromium.
 
 ###### What can I do?
 
@@ -112,7 +112,7 @@ const browser = await puppeteer.launch({headless: false}); // default is true
 **2. Runs a bundled version of Chromium**
 
 By default, Puppeteer downloads and uses a specific version of Chromium so its API
-is guaranteed to work out of the box. To use Puppeteer with a different version of Chrome,
+is guaranteed to work out of the box. To use Puppeteer with a different version of Chrome or Chromium,
 pass in the executable's path when creating a `Browser` instance:
 
 ```js
@@ -120,6 +120,9 @@ const browser = await puppeteer.launch({executablePath: '/path/to/Chrome'});
 ```
 
 See [`Puppeteer.launch()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions) for more information.
+
+See [`this article`](https://www.howtogeek.com/202825/what%E2%80%99s-the-difference-between-chromium-and-chrome/) for a description 
+of the differences between Chromium and Chrome. [`This article`](https://chromium.googlesource.com/chromium/src/+/lkcr/docs/chromium_browser_vs_google_chrome.md) describes some differences for Linux users.
 
 **3. Creates a fresh user profile**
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -121,6 +121,9 @@ const browser = await puppeteer.launch({executablePath: '/path/to/Chrome'});
 
 See [`Puppeteer.launch()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions) for more information.
 
+See [`this document`](https://chromium.googlesource.com/chromium/src/+/lkcr/docs/chromium_browser_vs_google_chrome.md) for a description
```
ebidel: https://www.howtogeek.com/202825/what%E2%80%99s-the-difference-between-chromium-and-chrome/ is a bit newer and isn't specific to Linux.

docs/api.md
```diff
@@ -12,6 +12,7 @@
   * [puppeteer.connect(options)](#puppeteerconnectoptions)
   * [puppeteer.executablePath()](#puppeteerexecutablepath)
   * [puppeteer.launch([options])](#puppeteerlaunchoptions)
+  * [Chromium vs Chrome](#chromium-vs-chrome)
```
ebidel: I wonder if we need this section in the API docs? It's duplicated in README.

docs/api.md
```diff
@@ -249,13 +250,13 @@ This methods attaches Puppeteer to an existing Chromium instance.
...
+  - `executablePath` <[string]> Path to a Chromium or Chrome executable to run instead of bundled Chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
   - `slowMo` <[number]> Slows down Puppeteer operations by the specified amount of milliseconds. Useful so that you can see what is going on.
   - `args` <[Array]<[string]>> Additional arguments to pass to the Chromium instance. List of Chromium flags can be found [here](http://peter.sh/experiments/chromium-command-line-switches/).
-  - `handleSIGINT` <[boolean]> Close chrome process on Ctrl-C. Defaults to `true`.
-  - `handleSIGTERM` <[boolean]> Close chrome process on SIGTERM. Defaults to `true`.
-  - `handleSIGHUP` <[boolean]> Close chrome process on SIGHUP. Defaults to `true`.
-  - `timeout` <[number]> Maximum time in milliseconds to wait for the Chrome instance to start. Defaults to `30000` (30 seconds). Pass `0` to disable timeout.
+  - `handleSIGINT` <[boolean]> Close Chromium process on Ctrl-C. Defaults to `true`.
```
ebidel: Chromium -> "browser" for these?

ebidel (changes requested):

README.md
karloskalcium: Fair enough, but my thought was, chromium.googlesource.com is an official google page, as opposed to howtogeek, which is maintained by someone random. We could link both perhaps?

docs/api.md
karloskalcium: I added it because there was already a note about chrome vs chrominum, so I just made it into a section and added a little bit more exposition.

docs/api.md
karloskalcium: sure that seems reasonable.

docs/api.md
ebidel: Ack

README.md
ebidel: I tend to agree with you, but that page is very specific to Linux. It might confuse people.

karloskalcium:
Made some changes as discussed, please take another look.

ebidel (approved):

PR #1581, opened 2017-12-11
feat(Browser): introduce browser.process() variable

### LINKED ISSUE #1539: Feature request: Give access to created processes
It would be great if the library gave access to the browser process it creates. I need access to it and I'm thinking of either forking the repository or creating the browser process myself.

aslushnikov:
This patch adds a `browser.process()` getter to expose the child process
with running browser.

Fixes #1539.

### DIFF
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -22,12 +22,14 @@ class Browser extends EventEmitter {
   /**
    * @param {!Puppeteer.Connection} connection
    * @param {!Object=} options
+   * @param {?Puppeteer.ChildProcess} process
    * @param {(function():Promise)=} closeCallback
    */
-  constructor(connection, options = {}, closeCallback) {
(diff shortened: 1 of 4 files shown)


lib/Browser.js
```diff
@@ -22,12 +22,14 @@ class Browser extends EventEmitter {
   /**
    * @param {!Puppeteer.Connection} connection
    * @param {!Object=} options
+   * @param {?Puppeteer.ChildProcess} process
```
JoelEinbinder, line 25: Shouldn’t this be NodeJS.ChildProcess?

JoelEinbinder (approved):

lib/Browser.js
aslushnikov: For some reason, that's what we have in externs

JoelEinbinder: Ok I trust whatever I was doing to make it compile then

PR #1587, opened 2017-12-12
chore: remove "SX" function from test/test.js

aslushnikov:
The new testrunner supports async tests.


aslushnikov:
Closed in favor of #1531

PR #1601, opened 2017-12-14
feat(Response): add fromDiskCache/fromServiceWorker info

### LINKED ISSUE #1551: Add Response.fromDiskCache/fromServiceWorker for better offline PWA testing
DTP [supports](https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-Response) a `fromDiskCache` and `fromServiceWorker` on the Response object.  It would be useful to expose these props on `Response` so users don't have to hack something together:

```js
page._client.on('Network.responseReceived', result => {
  if (!resp.fromServiceWorker) {
    console.log('Not cached by sw:', result.response.url);
  }
});
```

**Use cases**
1. allow authors to verify a sw is caching the expected assets.
2. allow authors to identify which requests are _not_ being cached by sw. Catch regressions on their offline experience.

I hit number 2 recently. Thought I was caching something that I wasn't:

screen shot 2017-12-07 at 3 40 46 pm

---
I originally approached this similar to the Lighthouse solution: load the PWA, wait for service worker to cache assets, go offline, reload the page and listen for responses that are `!== 200`. 

```js
await page.goto('https://www.chromestatus.com/features', {waitUntil: 'networkidle0'});

await page.evaluate('navigator.serviceWorker.ready'); // wait for sw to become active

await page.setOfflineMode(true); // go offline

page.on('response', resp => {
  if (resp.status !== 200) {
    ...
  }
});

await page.reload({waitUntil: 'networkidle0'});
```

However, this **doesn't work** in Puppeteer. Using Puppeteer, many of the responses will come from the disk cache after the page reload. So, it looks like the responses work offline but they don't!  In Lighthouse, we [disable the disk cache](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-core/gather/driver.js#L951-L958) to avoid this issue.

Another approach I thought of is to inject code into the page and examine what's in `window.caches`. But that's 1000% overkill.

...much easier to make`fromDiskCache` and `fromServiceWorker` available 👍.

ebidel:
Fixes #1551

I tried to use `one-style.html` for the test but it would require additions to sw.js and the test server doesn't support caching AFAIK.


ebidel:
I'm not able to run all the tests b/c of https://github.com/GoogleChrome/puppeteer/issues/1600. Any idea why Travis is unhappy?

docs/api.md
```diff
@@ -182,6 +182,8 @@
   * [request.url](#requesturl)
 - [class: Response](#class-response)
   * [response.buffer()](#responsebuffer)
+  * [response.fromDiskCache](#responsefromdiskcache)
```
aslushnikov: If I understand it correctly, every response could be served either from network, disk cache or service worker. So it's never the case that both `fromDiskCache` and `fromServiceWorker` are `true`. Is this the case? 

If so, I'd rather do a `response.servedFrom()` method that returns either `network`, `diskcache` or `serviceworker`.

amprew: They can both be true at the same time. If you check a request against `https://googlechrome.github.io/samples/service-worker/basic/`, it shows both `fromDiskCache` and `fromServiceWorker` as true. From what I understand `fromServiceWorker` is when the request is served from the `fetch` event inside the web worker. But fromDiskCache will also return true if you return from the Cache API.

Hope I haven't misunderstood your question.

test/test.js
```diff
@@ -3301,6 +3303,24 @@ describe('Page', function() {
       }
     });
   });
+
+  describe('Response', function() {
+    it('specifies when response comes from cache and service worker', async({page, server}) => {
```
andrewwwcollins: Looks like `server` in `{page, server}` isn't being used :)

test/test.js
```diff
@@ -3301,6 +3303,24 @@ describe('Page', function() {
...
+      let responses = [];
+      page.on('response', response => {
+        if (response.url.includes('.js'))
+          responses.push(response);
+      });
+      await page.goto('https://www.chromestatus.com/features', {waitUntil: 'networkidle2'});
+      expect(responses[0].fromDiskCache).toBe(false);
+      expect(responses[0].fromServiceWorker).toBe(false);
+      responses = [];
+      response = await page.reload({waitUntil: 'networkidle2'});
```
amprew: looks like `response` is being assigned to the scope but not used

docs/api.md
ebidel: It looks like they're different. Running against chromestatus shows all 4 permutations can happen (urls have been truncated):

screen shot 2017-12-16 at 8 43 22 am

That makes sense. The disk cache is different than the sw cache storage api.

test/test.js
ebidel: done

test/test.js
ebidel: done. weird linting didnt catch this.

test/test.js
```diff
@@ -3301,6 +3303,30 @@ describe('Page', function() {
       }
     });
   });
+
+  describe('Response', function() {
+    it('specifies when response comes from cache and service worker', async() => {
```
aslushnikov: page and browser are passed into the test:

```js
async ({page, browser}) => {
  // ..
});
```

No need to create another one; reusing browser speeds up tests significantly

test/test.js
```diff
@@ -3301,6 +3303,30 @@ describe('Page', function() {
...
+      const newPage = await browser.newPage();
+
+      // Use known site that uses server caching and service worker caching.
+      let responses = [];
+      newPage.on('response', response => {
+        if (response.url.includes('.js'))
+          responses.push(response);
+      });
+
+      await newPage.goto('https://www.chromestatus.com/features', {waitUntil: 'networkidle2'});
```
aslushnikov: please don't use external URLs; we aim for the hermetic builds.

docs/api.md
```diff
@@ -2065,6 +2067,16 @@ Contains the URL of the request.
 #### response.buffer()
 - returns: <Promise<[Buffer]>> Promise which resolves to a buffer with response body.
 
+#### response.fromDiskCache
```
aslushnikov: let's make them functions to unify our API

test/test.js
ebidel: I was doing that but it was causing other tests to fail: https://travis-ci.org/GoogleChrome/puppeteer/jobs/317396566#L556

test/test.js
ebidel: > I tried to use one-style.html for the test but it would require additions to sw.js and the test server doesn't support caching AFAIK.

What do you suggest for the server part?

docs/api.md
ebidel: done

test/test.js
ebidel: I've added caching to the test server. Still seeing a lot of unrelated errors that I could use your help on. Not familiar enough with the worker and custom test server.

ebidel:
Rebased. Still need guidance on what do about https://github.com/GoogleChrome/puppeteer/pull/1601#discussion_r157564662. The added caching to the server causes unrelated tests to fail. 

Instead, we could go back to a well-known URL that has caching + sw. https://googlechrome.github.io/samples/service-worker/basic/index.html is our official service worker demo and won't be changing.

ebidel:
I can also switch to `<link rel="stylesheet" href="https://developers.google.com/web/styles/wf-root.css">` in sw-fetch.html if you don't mind a page that uses an external resource. That makes all tests pass :)

ebidel:
rebased

aslushnikov:
@ebidel let me introduce caching to the simple server so that this can be merged in.

aslushnikov:
@ebidel the patch doesn't handle memory caching, and it's quite hard to predict if the response will be memory-cached or disk-cached. As far as I know the behavior is not spec'ed in chrome.

I'd expose `response.fromCache()` instead of `response.fromDiskCache()` that will report `true` if the response is served either from memory or disk cache. wdyt? I have a draft that implements this.

ebidel:
I don't mind changing the name but doesn't the protocol differentiate the two? The name and docs from [`Response.fromDiskCache`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-Response) suggests that it's just disk. Or is it a deceiving name?

aslushnikov:
@ebidel the protocol differentiates between the two; memory-cached responses are reported with `Network.requestServedFromCache`. It's hard to tell when chrome will use memory cache and when disk cache, as far as i know the behavior is not spec'ed.

ebidel:
That must be how the frontend shows each:

screen shot 2018-01-05 at 11 01 15 am

Does DT prioritize showing "from memory cache" if the resource is in both? Should we do the same in PPTR or add `response.fromMemoryCache()` somehow?

ebidel:
What about this:

```js
const responsesFromMemCache = [];
page._client.on('Network.requestServedFromCache', result => {
  responsesFromMemCache.push(result.requestId);
});

page._client.on('Network.responseReceived', result => {
  const resp = result.response;
  resp.fromMemoryCache = responsesFromMemCache.includes(result.requestId);
  console.log(resp.url, resp.fromMemoryCache, resp.fromDiskCache, resp.fromServiceWorker);
});
```

Screenshot is from testing on https://stackoverflow.com/questions/13140318/check-whether-network-response-is-coming-from-server-or-chrome-cache:

screen shot 2018-01-05 at 11 28 01 am

Want to add `response.fromMemoryCache()` or would you prefer `response.fromCache()`, which covers both cases?

aslushnikov:
> Does DT prioritize showing "from memory cache" if the resource is in both?

It does. However, since response could be served either from memory or disk cache, a single `response.fromCache()` makes sense to me. Later on, if there's a need, we can start returning a string instead of boolean that clarifies cache source.

ebidel:
Cool. I've changed the name.

`response.fromCache()` only considers disk cache. LMK if you want me to add the plumbing in https://github.com/GoogleChrome/puppeteer/pull/1601#issuecomment-355646010 to also consider the mem cache.

Also removed my [`Cache-Control` addition](https://github.com/GoogleChrome/puppeteer/pull/1601/commits/c701fe1eebe6cfab3dfd520f562320c59f49bfc9#diff-38e89beafd599cb655dc914a04f642feL201) to SimpleServer.js. So Travis will continue to fail until [you add caching to it](https://github.com/GoogleChrome/puppeteer/pull/1601#issuecomment-355409613).

aslushnikov:
> response.fromCache() only considers disk cache. LMK if you want me to add the plumbing in #1601 (comment) to also consider the mem cache.

Yeah, let's add it - otherwise it's quite hard to test the `fromCache` feature: it's hard to predict when chrome uses memcache and when it falls back to disk cache.

> Also removed my Cache-Control addition to SimpleServer.js. So Travis will continue to fail until you add caching to it.

I'll upload implementation shortly

lib/NetworkManager.js
```diff
@@ -302,6 +312,12 @@ class Request {
     this._headers = {};
     for (const key of Object.keys(payload.headers))
       this._headers[key.toLowerCase()] = payload.headers[key];
+
+    this._fromMemCache = false;
+  }
+
+  _setFromMemoryCache(fromMemCache) {
```
aslushnikov: nit: let's drop the method and reach into request right away

ebidel:
LMK if that looks ok.

lib/NetworkManager.js
```diff
@@ -165,6 +166,15 @@ class NetworkManager extends EventEmitter {
     }
   }
 
+  /**
+   * @param {!Object} event
+   */
+  _onRequestServedFromCache(event) {
+    const request = this._requestIdToRequest.get(event.requestId);
+    if (request)
+      request._setFromMemoryCache(true);
```
aslushnikov:
```js
request._fromMemoryCache = true;
```

lib/NetworkManager.js
```diff
@@ -302,6 +312,12 @@ class Request {
     this._headers = {};
     for (const key of Object.keys(payload.headers))
       this._headers[key.toLowerCase()] = payload.headers[key];
+
+    this._fromMemCache = false;
```
aslushnikov: nit: let's rename into `this._fromMemoryCache`

lib/NetworkManager.js
ebidel: done

lib/NetworkManager.js
ebidel: done

lib/NetworkManager.js
ebidel: done

ebidel:
@aslushnikov ready for your additions :p

aslushnikov:
I wasn't able to push in this branch, so closing in favor of #1971.

PR #1605, opened 2017-12-15
test: nicer matchers output

aslushnikov:
This patch improves matchers to provide more information once
some of them fails.

### DIFF
--- a/utils/testrunner/Matchers.js
+++ b/utils/testrunner/Matchers.js
@@ -96,7 +96,7 @@ const DefaultMatchers = {
   },
 
   toEqual: function(value, other, message) {
-    message = message || `Object::${value.constructor.name} ≈ Object::${other.constructor.name}`;
+    message = message || `${JSON.stringify(value)} ≈ ${JSON.stringify(other)}`;
     return { pass: JSON.stringify(value) === JSON.stringify(other), message };
   },
 
--- a/utils/testrunner/Matchers.js
+++ b/utils/testrunner/Matchers.js
@@ -43,7 +43,7 @@ class Expect {
 
     function applyMatcher(matcherName, matcher, inverse, value, ...args) {
       const result = matcher.call(null, value, ...args);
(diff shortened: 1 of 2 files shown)


utils/testrunner/Matchers.js
```diff
@@ -43,50 +43,60 @@ class Expect {
...
       const result = matcher.call(null, value, ...args);
-      const message = result.message || `expect.${matcherName} failed`;
+      const message = `expect.${matcherName} failed` + (result.message ? `: ${result.message}` : '');
       console.assert(result.pass !== inverse, message);
     }
   }
 }
 
 const DefaultMatchers = {
   toBe: function(value, other, message) {
```
JoelEinbinder, line 53: If you want to be fancy
function (value, other, message=`${value} == ${other}`)

utils/testrunner/Matchers.js
```diff
@@ -43,50 +43,60 @@ class Expect {
...
     return { pass: value === null, message };
   },
 
   toContain: function(value, other, message) {
+    message = message || `${value} ⊇ ${other}`;
     return { pass: value.includes(other), message };
   },
 
   toEqual: function(value, other, message) {
+    message = message || `Object::${value.constructor.name} ≈ Object::${other.constructor.name}`;
```
JoelEinbinder: Jasmine outputs json here. We use toEqual with plane objects, so the constructor name doesn't help much.

utils/testrunner/Matchers.js
aslushnikov: nice! I'll try it out some other time

JoelEinbinder (approved):

utils/testrunner/Matchers.js
aslushnikov: I remember huge JSON.stringifed output of `Request` objects in our tests from time to time which I hated, but it seems to be never a case for `toEqual` matcher. Switched to `JSON.stringify` here.

PR #1606, opened 2017-12-15
chore(testrunner): support first-class test debugging

aslushnikov:
This patch teaches testrunner to override both timeout and
parallel execution option if there's attached inspector.


JoelEinbinder (commented):
Philosophically I’m not sure if it’s a good idea to change the way tests run while debugging

JoelEinbinder (approved):

Issue #1556, opened 2017-12-08
Add page.disableCache(boolean)


ebidel:
Related to https://github.com/GoogleChrome/puppeteer/issues/1551, it would be useful to have an API call to disable the browser's disk cache. 

Use cases:
- A/B test server caching/config without having to relaunch chrome or use `userDataDir` to do multiple runs.
- Use a single browser instance to test the first load of every view of a SPA. Need to bypass the cache to ensure requests are always fresh for shared resources.

Ref: [`Network.setCacheDisabled`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-setCacheDisabled)

aslushnikov:
Nice usecases, thanks

This should be easy to implement, similarly to the `setJavascriptEnabled`:

https://github.com/GoogleChrome/puppeteer/blob/a164524c72443920e0f008d62c5362c08a7794a0/lib/Page.js#L564-L566

vonGameTheory:
Yes please.   Is there any approved way of disabling it currently?  The only thing I could come up with was using these at launch:

--media-cache-size=1 
--disk-cache-size=1

 (On linux it was suggested to set the cache directory to /dev/null, but on Windows?)

ppatel221:
Hey, im part of an Open Source Development class at Seneca College and im looking to contribute to an open source project. Can i help with this feature?

ebidel:
@ppatel221 feel free. We're not going to turn down a PR :)

amprew:
I have added this in my open PR ^ (#1609). Although it does depend on your changed in #1601, @ebidel

amprew:
Would be great to get a review on this and with new methods implemented https://github.com/GoogleChrome/puppeteer/pull/1609, thanks!

PR #1609, opened 2017-12-15
feat(Page): add `setCacheEnabled(enabled)` to Page object

amprew:
Fixes: #1556 

Implements `Page.disableCache(disabled)` method.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -86,6 +86,7 @@
   * [page.reload(options)](#pagereloadoptions)
   * [page.screenshot([options])](#pagescreenshotoptions)
   * [page.select(selector, ...values)](#pageselectselector-values)
+  * [page.setCacheEnabled(enabled)](#pagesetcacheenabledenabled)
   * [page.setContent(html)](#pagesetcontenthtml)
   * [page.setCookie(...cookies)](#pagesetcookiecookies)
   * [page.setDefaultNavigationTimeout(timeout)](#pagesetdefaultnavigationtimeouttimeout)
@@ -1149,6 +1150,12 @@ page.select('select#colors', 'red', 'green', 'blue'); // multiple selections
 ```
 
 Shortcut for [page.mainFrame().select()](#frameselectselector-values)
+
+#### page.setCacheEnabled(enabled)
+- `enabled` <[boolean]> sets the `enabled` state of the cache.
+- returns: <[Promise]>
+
+Toggles ignoring cache for each request based on the enabled state. By default, caching is enabled.
 
 #### page.setContent(html)
 - `html` <[string]> HTML markup to assign to the page.
--- a/test/test.js
+++ b/test/test.js
@@ -3050,6 +3050,21 @@ describe('Page', function() {
     });
   });
 
+  describe('Page.setCacheEnabled', function() {
+    it('should enable or disable the cache based on the state passed', async({page, server}) => {
+      const responses = new Map();
+      page.on('response', r => responses.set(r.url().split('/').pop(), r));
+
+      await page.goto(server.PREFIX + '/cached/one-style.html', {waitUntil: 'networkidle2'});
+      await page.reload({waitUntil: 'networkidle2'});
+      expect(responses.get('one-style.css').fromCache()).toBe(true);
+
+      await page.setCacheEnabled(false);
+      await page.reload({waitUntil: 'networkidle2'});
+      expect(responses.get('one-style.css').fromCache()).toBe(false);
+    });
+  });
+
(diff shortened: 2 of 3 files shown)


docs/api.md
```diff
@@ -593,6 +594,12 @@ If URLs are specified, only cookies for those URLs are returned.
   - `secure` <[boolean]>
 - returns: <[Promise]>
 
+#### page.disableCache(disabled)
```
aslushnikov: Let's name this `page.setCacheEnabled` to be aligned with `page.setJavascriptEnabled` method that we already have.

amprew: I have made these changes 👍

docs/api.md
```diff
@@ -990,6 +991,12 @@ page.select('select#colors', 'red', 'green', 'blue'); // multiple selections
 
 Shortcut for [page.mainFrame.select()](#frameselectselector-values)
 
+#### page.setCacheEnabled(enabled)
+- `enabled` <[boolean]> sets the `enabled` state of the cache.
+- returns: <[Promise]>
+
+Toggles ignoring cache for each request based on the enabled state.
```
aslushnikov: Can you please add:

> By default, caching is enabled.

test/test.js
```diff
@@ -2751,6 +2753,23 @@ describe('Page', function() {
...
 
+  describe('Page.setCacheEnabled', function() {
+    it('should enable or disable the cache based on the state passed', async({page}) => {
+      let responses = [];
+      page.on('response', response => {
+        if (response.url.includes('.js'))
+          responses.push(response);
+      });
+      await page.setCacheEnabled(false);
+      await page.goto('https://www.chromestatus.com/features', {waitUntil: 'networkidle2'});
```
aslushnikov: Can you please use one of the `test/assets/cached/` assets intead of `chromestatus.com` website? We aim for the hermetic tests.

aslushnikov (commented):
@andrewwwcollins The `fromCache/fromServiceWorker` methods are just landed: #1971. Are you interesteed in landing this PR? It would be nice have.

amprew:
@aslushnikov - ~yes I can make these changes this evening.~

I have made these changes now.

test/test.js
```diff
@@ -3050,6 +3050,23 @@ describe('Page', function() {
...
 
+  describe('Page.setCacheEnabled', function() {
+    it('should enable or disable the cache based on the state passed', async({page, server}) => {
+      let responses = [];
+      page.on('response', response => {
+        if (response.url().includes('.css'))
+          responses.push(response);
+      });
+      await page.setCacheEnabled(false);
+      await page.goto(server.PREFIX + '/cached/one-style.html', {waitUntil: 'networkidle2'});
```
aslushnikov: two nits:
1. Let's start from making sure it's cached, and then busting caching with this method
2. Can we load-and-reload initially to make sure the stylesheet is cached?

amprew: I have made these changes now. I needed to reset the `responses` array twice. What do you think?

aslushnikov (approved):

test/test.js
aslushnikov: ok, I sent a comment but for some reason it didn't get posted to github. 
Let me re-type it here:

> I needed to reset the responses array twice. What do you think?

You can avoid this by subscribing after the `page.goto`. You can also save a few lines by storing resources in the map:

```js
const responses = new Map();
page.on('response', r => responses.set(r.url().split('/').pop(), r));

await page.goto(server.PREFIX + '/cached/one-style.html');
await page.reload();
expect(responses.get('one-style.css').fromCache()).toBe(true);

await page.setCacheEnabled(false);
await page.reload();
expect(responses.get('one-style.css').fromCache()).toBe(true);
```

amprew:
@aslushnikov - is the appveyor continuous-integration required for this PR? I have had it passing before it just seems a bit flakey. I can rerun them if you think necessary.

test/test.js
amprew: Ah nice, didn't think of subscribing to the event later. Could be an issue when the responses map gets larger because of no filtering due to more resources being added. But I like this for the current solution.

amprew: Just a note for anyone else who goes to look at the above code the second expectation should be `toBe(false)`.

amprew: Have made those changes now.

aslushnikov:
> @aslushnikov - is the appveyor continuous-integration required for this PR? I have had it passing before it just seems a bit flakey. I can rerun them if you think necessary.

@andrewwwcollins appveyour is very flaky for us; this has nothing to do with your change.

Thank you for you time.

Issue #1229, opened 2017-10-31
Page with CSP breaks puppeteer


aslushnikov:
A page with CSP will break some of puppeteer methods, such as `page.addScriptTag`. At the very least, we should fail gracefully.

It would be ideal to have a way to disable CSP altogether.

Reported in: #1091

rakeshpai:
+1 for ability to disable CSP.

Use case: I'm using puppeteer as a post-build crawler to generate a static dump of my SPA. When crawling, I'd like to inline critical CSS, and add a CSP directive for the CSS text hash. While I'm at it, I want to change the API server from `localhost:port` to `api.example.com`. However, if my page already had a CSP meta tag in it (which is good to have during development so I can catch CSP violations early), puppeteer starts getting funky. Would love it if I can disable CSP altogether when crawling.

sradu:
+1 for ability to disable CSP.

Thank you.

yujiosaka:
In order to disable CSP, it's not a complete solution but adding `nonce-<base64-value>` to script tag will relax the security blocks. It only works for the site where the CSP headers or meta tags have the nonce part. It doesn't work for other sites, but in this way we don't have to change the default behavior or disable security blocks.

I'm working on that idea, but it will be probably difficult to completely disable CSP with current Chrome DevTool Protocol because I don't find APIs to intercept/modify response headers to do that yet.

niieani:
Here's my workaround:

```ts
  import fetch from 'node-fetch'

  const requestInterceptor = async (request) => {
    const url = request.url()
    const requestHeaders = request.headers()
    const acceptHeader = requestHeaders.accept || ''
    if (url.includes('some-website-with-csp.com') && (acceptHeader.includes('text/html'))) {
      const cookiesList = await page.cookies(url)
      const cookies = cookiesList.map(cookie => `${cookie.name}=${cookie.value}`).join('; ')
      delete requestHeaders['x-devtools-emulate-network-conditions-client-id']
      if (requestHeaders.Cookie) {
        requestHeaders.cookie = requestHeaders.Cookie
        delete requestHeaders.Cookie
      }
      const theseHeaders = Object.assign({'cookie': cookies}, requestHeaders, {'accept-language': 'en-US,en'})

      const init = {
        body: request.postData(),
        headers: theseHeaders,
        method: request.method(),
        follow: 20,
      }
      const result = await fetch(
        url,
        init,
      )
      const resultHeaders = {}
[...]

PR #1611, opened 2017-12-16
fix(Page): Prevent silently failing when addScriptTag and addStyleTag violate Content Security Policy (#1229)

yujiosaka:
Both `page.addScriptTag()` and `page.addStyleTag()` silently fail when the requested page is protected by Content Security Policy.

Although it would be a nice feature to have an option to disable it, I think the default behavior should be failing gracefully.

I don't like my implementation of adding a global variable and an invisible element in the page, so I would like to hear better ideas and solutions.

Fixes: https://github.com/GoogleChrome/puppeteer/issues/1229

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -457,9 +457,9 @@ class Frame {
      * @return {!Promise<!ElementHandle>}
      */
     async function evaluateScriptContent(content) {
-      content += `\nwindow.pptrCsp = true; //# Detect Content Security Policy`;
+      content += '\nwindow.pptrCsp = true; //# Detect Content Security Policy';
       const script = await context.evaluateHandle(addScriptContent, content);
-      const cspInactive = await context.evaluate(`window.pptrCsp`);
+      const cspInactive = await context.evaluate(() => window['pptrCsp']);
       if (!cspInactive) throw new Error('Refused to execute inline script because it violates the Content Security Policy');
       return script;
     }
@@ -536,9 +536,11 @@ class Frame {
      */
     async function evaluateStyleContent(content) {
       await context.evaluate(addCSPElement);
-      content += `\n#pptr-csp { font-family: csp-inactive; } /*# Detect Content Security Policy */`;
+      content += '\n#pptr-csp { font-family: csp-inactive; } /*# Detect Content Security Policy */';
       const style = await context.evaluateHandle(addStyleContent, content);
-      const fontFamily = await context.evaluate("window.getComputedStyle(document.getElementById('pptr-csp'), null).fontFamily");
+      const fontFamily = await context.evaluate(() => (
+        window.getComputedStyle(window.document.getElementById('pptr-csp'), null).fontFamily
+      ));
       const cspInactive = (fontFamily === 'csp-inactive');
       if (!cspInactive) throw new Error('Refused to execute inline style because it violates the Content Security Policy');
       return style;
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -404,10 +404,10 @@ class Frame {
    * @return {!Promise<!ElementHandle>}
    */
   async addScriptTag(options) {
+    const context = await this._contextPromise;
     if (typeof options.url === 'string') {
       const url = options.url;
       try {
-        const context = await this._contextPromise;
         return await context.evaluateHandle(addScriptUrl, url);
       } catch (error) {
         throw new Error(`Loading script from ${url} failed`);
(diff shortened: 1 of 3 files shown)


dios-david:
Hi @yujiosaka,
I like your idea, have you tried to catch the error message by listening on Page events?
https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-page

yujiosaka:
@dios-david 
Unfortunately, Puppeteer has no clue when the inline scripts/styles fail by CSP. It doesn't catch either `error` nor `requesterror` events (It's understandable because it's not crashing the page or sending requests at all). I failed to mention that silent-fail on CSP occurs only when adding scripts/styles inline. When they are added as external files that violate CSP, Puppeteer already catches the errors.

lib/FrameManager.js
```diff
@@ -454,17 +451,29 @@ class Frame {
...
+
+    /**
+     * @param {string} content
+     * @return {!Promise<!ElementHandle>}
+     */
+    async function evaluateScriptContent(content) {
+      content += '\nwindow.pptrCsp = true; //# Detect Content Security Policy';
+      const script = await context.evaluateHandle(addScriptContent, content);
+      const cspInactive = await context.evaluate(() => window['pptrCsp']);
+      if (!cspInactive) throw new Error('Refused to execute inline script because it violates the Content Security Policy');
```
JoelEinbinder, line 463: We should set `pptrCsp` to false to avoid confusing ourselves the second time through this function. I think it would also be safter is `pptrCsp` was replaced by a random string or a static counter.

lib/FrameManager.js
```diff
@@ -512,6 +518,33 @@ class Frame {
...
+     * @return {!Promise<!ElementHandle>}
+     */
+    async function evaluateStyleContent(content) {
+      await context.evaluate(addCSPElement);
+      content += '\n#pptr-csp { font-family: csp-inactive; } /*# Detect Content Security Policy */';
+      const style = await context.evaluateHandle(addStyleContent, content);
+      const fontFamily = await context.evaluate(() => (
+        window.getComputedStyle(window.document.getElementById('pptr-csp'), null).fontFamily
+      ));
+      const cspInactive = (fontFamily === 'csp-inactive');
```
JoelEinbinder, line 544: We should clean up the csp element here, and also have a unique ID for it.

test/server/SimpleServer.js
```diff
@@ -99,6 +101,10 @@ class SimpleServer {
     this._auths.set(path, {username, password});
   }
 
+  setCSP() {
```
JoelEinbinder, line 104: I'd like this to take a boolean: `setCSP(true)`. Otherwise I will use setCSP(false) and become very confused as to why it doesn't work 😀

JoelEinbinder (commented):
This is an interesting patch. I don't know enough about CSP at the moment. Would it be possible to detect the CSP header on the main response, and then set a flag to disable addScriptTag?

lib/FrameManager.js
dios-david: +1, I was thinking of a unique identifier as well

yujiosaka:
@JoelEinbinder 

> Would it be possible to detect the CSP header on the main response, and then set a flag to disable addScriptTag?

It's a very good point :)
I was first thinking that it would be difficult because the code has to deal with both HTTP headers and meta tags, and it would be easy to make mistakes because the [CSP Level 3](https://www.w3.org/TR/CSP3/) is still a working draft and [Chrome's implementation](https://developer.chrome.com/extensions/contentSecurityPolicy) is still ongoing.

However, I would be nice to give a try for that, so I'll make two PRs in the following steps:

1. Improve this PR by following @JoelEinbinder and @dios-david 's comments
2. Make another PR which catches error by CSP headers and meta tags.

Then, hopefully you can choose one of the two PRs.

(Also, I can hopefully relax the CSP block, but it should be another PR)

paulirish:
Here's an alternative approach.. This is me logging the protocol traffic while adding a style tag to the DOM which has CSP enabled:

image



Before the `Runtime.evaluate` resolves, a `Log.entryAdded` event is fired with `source:"security"` and `level:"error"`. (You can see the event is received before 317 returns. It's deterministic). 

So if this log entry is added during this time, take the message, and reject with it.

yujiosaka:
@paulirish Your approach sounds elegant. I should try that first!

test/server/SimpleServer.js
yujiosaka: Thanks! I applied your comment [here](https://github.com/GoogleChrome/puppeteer/pull/1624/files#diff-38e89beafd599cb655dc914a04f642feR104)!

aslushnikov:
Closing this in favor of #1624

PR #1613, opened 2017-12-16
test(BrokenTest): Fix broken Tracing unit tests

dios-david:
This patch fixes broken Tracing tests which are failing since the `state.parallel` -> `state.parallelIndex` modification of #1531


aslushnikov (approved):
Thank you!

PR #1614, opened 2017-12-16
feat(Tracing): add support for tracing without saving to file

dios-david:
This patch adds support for using tracing without necessarily saving the result to a file. This allows consumers to process the tracing result without any unnecessary file operations.

### DIFF
--- a/lib/Tracing.js
+++ b/lib/Tracing.js
@@ -27,15 +27,24 @@ class Tracing {
   constructor(client) {
     this._client = client;
     this._recording = false;
+    this._result = '';
     this._path = '';
+  }
+
+  /**
+   * @return {string} The trace result.
+   */
+  getResult() {
+    if (this._recording)
+      throw new Error('Tracing is still in progress!');
+    return this._result;
   }
 
   /**
    * @param {!Object} options
    */
-  async start(options) {
+  async start(options = {}) {
     console.assert(!this._recording, 'Cannot start recording trace while already recording trace.');
-    console.assert(options.path, 'Must specify a path to write trace file to.');
 
     const defaultCategories = [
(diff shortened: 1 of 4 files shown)


lib/Tracing.js
```diff
@@ -69,20 +82,26 @@ class Tracing {
...
+  async _readResultFromStream(handle) {
     let eof = false;
-    const file = await openAsync(path, 'w');
+    this._result = '';
     while (!eof) {
       const response = await this._client.send('IO.read', {handle});
       eof = response.eof;
-      if (path)
-        await writeAsync(file, response.data);
+      this._result += response.data;
```
aslushnikov, line 92: This loads the whole trace into the node.js heap - do you really want to do this? Some traces could be huge and won't fit in the memory.

dios-david: You're right, that's what I was thinking the other day..
Maybe the best would be just to expose the stream, or convert it to a buffer and expose that.
What do you think?

paulirish: fwiw in node <= 8, the string length maximum was 256MB. you hit that pretty regularly with big complex traces.  in node 9, it's 1GB, which is usually fine for traces.  https://stackoverflow.com/a/47781288/89484

so until puppeteer gives up support for node 8, we wouldnt be able to build a string like this. 

So yes, stream/buffer are other worthwhile options.

aslushnikov: > So yes, stream/buffer are other worthwhile options.

@paulirish @dios-david how would you process a trace that is saved into a buffer? You won't be able to `json.parse` it.

aslushnikov:
I don't think we should land this PR, given the memory concerns. Closing for now; feel free to resubmit if there are any good alternatives.

PR #1615, opened 2017-12-16
test(UnitTest): Adding support to run tests in concurrent mode based on number of CPUs

dios-david:
This patch adds support for defining thread count in concurrent unit test execution by using `MAX` as value. This will result in running tests based on number of CPUs dynamically.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -16,6 +16,7 @@
 const fs = require('fs');
 const rm = require('rimraf').sync;
 const path = require('path');
+const os = require('os');
 const {helper} = require('../lib/helper');
 if (process.env.COVERAGE)
   helper.recordPublicAPICoverage();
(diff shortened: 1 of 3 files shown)


test/test.js
```diff
@@ -54,12 +55,23 @@ const defaultBrowserOptions = {
...
+  const parallelArgIndex = process.argv.indexOf('-j');
+  let value;
+
+  if (process.env.PPTR_PARALLEL_TESTS)
+    value = process.env.PPTR_PARALLEL_TESTS;
+  if (parallelArgIndex !== -1)
+    value = process.argv[parallelArgIndex + 1];
+
+  if (value === 'MAX')
+    return os.cpus().length;
```
aslushnikov, line 68: We experimented with `os.cpus()` some time ago, but it didn't work out nicely: some of our workstations report up to 48 cpus, and running *this* parallel doesn't speed up things. We decided to introduce the `PPTR_PARALLEL_TESTS` env variable so that every machine could be configured nicely. 

Doesn't this approach work for you?

dios-david: WOW 48 sounds really massive, was it a NASA machine? :)

Anyway, this solution is backward compatible, so it still supports defining a desired thread count using the `PPTR_PARALLEL_TESTS` env variable, or using a CLI argument. This change just adds support for using `MAX` as value, which will be replaced with `os.cpus().length`. We can rename `MAX` to anything else if you wish (eg. `AVAILABLE_CPUS`).

As I noticed we still run our tests (eg. in Travis) using only 1 thread. Do you plan to modify it to run tests concurrently?

aslushnikov: > WOW 48 sounds really massive, was it a NASA machine? :)

It takes some power to build chromium :) 

> Anyway, this solution is backward compatible, so it still supports defining a desired thread count using the PPTR_PARALLEL_TESTS env variable, or using a CLI argument. This change just adds support for using MAX as value, which will be replaced with os.cpus().length.

Yeah I see. I ask because I'd avoid landing any code that won't be utilized. I wonder if you can use the env variable for your workflow like we do?

> As I noticed we still run our tests (eg. in Travis) using only 1 thread. Do you plan to modify it to run tests concurrently?

The main goal of test parallelisation was to cut on the local development time (we run tests dozens of times every day).
Running in parallel on trybots would sacrifice build reproducibility - I wonder if the speedup gains are significant enough to do this, at least for now.

dios-david: > The main goal of test parallelisation was to cut on the local development time (we run tests dozens of times every day).

Yes, my intention was just to provide a convenient way for developers to run tests locally.
I know that we can set thread count with env var or with cli arg, but running `npm run unit-concurrent` is just way more simple :)

aslushnikov:
@dios-david I'll close this for now; some of the most active pptr developers are pretty happy with `npm run unit -- -j 4`. Let's wait until the frustration builds up.

PR #1617, opened 2017-12-18
feat(Page.select) Throwing an error when select option is not found

### LINKED ISSUE #1384: Should get an error when select option is not found
### Steps to reproduce

* Puppeteer version: 0.13.0
* Platform / OS version: OSX 10.12.6
* URLs (if applicable): not available

HTML:
```html
<select id="my-select">
  <option value="first-option">First Option</option>
</select>
```

```js
await page.select('select#my-select', 'First Option');
```

I would expect an error that Puppeteer was unable to select the option specified since it seems to be looking at the `value` attribute and not the innerHTML. 

**What happens instead?**
The code executes without warning or error

dios-david:
This patch adds support for throwing an error when passing an option to `Page.select()` which is not found in the given select.

Fixes #1384

BREAKING CHANGE: `Page.select()` now throws an error if an invalid option was passed.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -527,6 +527,12 @@ class Frame {
         throw new Error('Element is not a <select> element.');
 
       const options = Array.from(element.options);
+      const optionValues = options.map(option => option.value);
+      const unknownValues = values.filter(value => !optionValues.includes(value));
+
+      if (unknownValues.length)
+        throw new Error(`Option(s) not found: ${unknownValues.join(', ')}`);
+
       element.value = undefined;
       for (const option of options)
         option.selected = values.includes(option.value);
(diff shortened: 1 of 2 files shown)


aslushnikov:
For the reference: this has been discussed before #1099 

@alixaxel @AlexChung1995 what do you guys think about this?

AlexChung1995:
Happy Holidays everyone,

We had a pretty extensive conversation about this. I think we found it difficult to find a solution that throws an error and was general enough of an API. Returning all the selected options provides more information than throwing an error also so I vote to stick with the current code.

aslushnikov:
Closing this for now; we can revisit this later if there are many unhappy clients of the current API.

PR #1618, opened 2017-12-18
feat(ElementHandle): Full model for boundingBox

schelkun:
According to issue https://github.com/GoogleChrome/puppeteer/issues/1357

Now user can specify box model. If model is not specified - `border` will be used.
Method will be more flexible with full model: https://chromedevtools.github.io/devtools-protocol/tot/DOM/#type-BoxModel


schelkun:
Naming is incorrect as I think.
Model should be changed to box. Any comments?

aslushnikov:
> Naming is incorrect as I think.
> Model should be changed to box. Any comments?

I agree. I'd have the `elementHandle.boxModel()` that returns all of the boxes right away.
```js
const boxModel = await elementHandle.boxModel();
boxModel.content; // box with {x, y, width, height}
boxModel.margin; // the same here and others
boxModel.border;
boxModel.padding;
```

schelkun:
@aslushnikov in this case we should rewrite all code that use `boundingBox` method. My idea is optional parameter, so method should be working the same way if no args specified.

docs/api.md
```diff
@@ -1829,10 +1829,28 @@ The method runs `element.querySelectorAll` within the page. If no elements match
...
+    - x <[number]> the x coordinate of the box in pixels.
+    - y <[number]> the y coordinate of the box in pixels.
+    - width <[number]> the width of the box in pixels.
+    - height <[number]> the height of the box in pixels.
+  - padding <[Object]>
+    - x <[number]> the x coordinate of the box in pixels.
+    - y <[number]> the y coordinate of the box in pixels.
+    - width <[number]> the width of the box in pixels.
+    - height <[number]> the height of the box in pixels.
+  - width <[number]> the width of the element in pixels.
```
aslushnikov: How is this different from content width?

lib/ElementHandle.js
```diff
@@ -118,7 +118,7 @@ class ElementHandle extends JSHandle {
   }
 
   /**
-   * @return {!Promise<?{x: number, y: number, width: number, height: number}>}
+   * @return {!Promise<?{border: object, content: object, margin: object, padding: object, width: number, height: number}>}
    */
   async boundingBox() {
```
aslushnikov: Can we rename this into `boxModel`? The `boundingBox` was named to match the `getBoundingClientRect` semantics; the new behavior no longer matches it.

lib/ElementHandle.js
```diff
@@ -128,13 +128,25 @@ class ElementHandle extends JSHandle {
     if (!result)
       return null;
 
-    const quad = result.model.border;
-    const x = Math.min(quad[0], quad[2], quad[4], quad[6]);
-    const y = Math.min(quad[1], quad[3], quad[5], quad[7]);
-    const width = Math.max(quad[0], quad[2], quad[4], quad[6]) - x;
-    const height = Math.max(quad[1], quad[3], quad[5], quad[7]) - y;
+    const calculatePos = function(quad)  {
```
aslushnikov: nit: `calculateBox` could be a better name

schelkun:
Oops, conflict. What's wrong? :thinking:

aslushnikov (approved):

aslushnikov (changes requested):
Excuse me my confusion.

I just realized that for a rotated element the `boxModel` should be rotated as well. The `boundingBox`, on the other hand, is axis-aligned.

For most purposes, the `boundingBox` is easier and nicer to work with; we should keep it.

The `boxModel` should return quads; but it's unclear  how convenient it is.

schelkun:
So, this PR was made for changing original `boundingBox`.
This PR is closed and new PR will created with new method.

PR #1620, opened 2017-12-19
feat: add page.xpath

### LINKED ISSUE #537: Handle XPath selectors
Is it planned/wanted to implement XPath selectors?

This can be done quite easily, but I don't know if it's a choice to not handle them?

I've implemented two methods in my own code to handle them :

```javascript
async waitForXpath(selector, options = { polling: 'mutation' }) {
  return this.waitForFunction(selector => {
    return null !== document.evaluate(selector, document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue;
  }, options, selector);
}

async $XPath(selector) {
  const remoteObject = await this._rawEvaluate(selector => {
    return document.evaluate(selector, document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue;
  }, selector);

  if (remoteObject.subtype === 'node')
    return new ElementHandle(this._client, remoteObject, this._mouse);

  await utils.releaseObject(this._client, remoteObject);
  return null;
}
```

I can make a PR if you're interested.

JoelEinbinder:
Adds
* page.xpath
* frame.xpath
* elementHandle.xpath

fixes #537

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -283,13 +283,29 @@ class Frame {
    * @return {!Promise<?ElementHandle>}
    */
   async $(selector) {
+    const document = await this._document();
+    const value = await document.$(selector);
+    await document.dispose();
+    return value;
+  }
+
+  /**
+   * @return {!Promise<!ElementHandle>}
+   */
+  async _document() {
     const context = await this._contextPromise;
-    const handle = await context.evaluateHandle(selector => document.querySelector(selector), selector);
-    const element = handle.asElement();
-    if (element)
-      return element;
-    await handle.dispose();
-    return null;
+    return await context.evaluateHandle(() => document);
+  }
+
+  /**
+   * @param {string} expression
+   * @return {!Promise<?ElementHandle>}
+   */
+  async xpath(expression) {
+    const document = await this._document();
+    const value = await document.xpath(expression);
+    await document.dispose();
+    return value;
   }
 
   /**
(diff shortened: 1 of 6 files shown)


lib/FrameManager.js
```diff
@@ -283,13 +283,29 @@ class Frame {
...
+    const document = await this._document();
+    const value = await document.$(selector);
+    await document.dispose();
+    return value;
+  }
+
+  /**
+   * @return {!Promise<!ElementHandle>}
+   */
+  async _document() {
```
aslushnikov, line 298: don't you want to cache this? Otherwise, everything becomes two-hops instead of a one hop.

paulirish:
I'm not an XPATH user so my opinion matters little, but the inclusion of `elementHandle.xpath()` stands out to me. Might want to see that element-rooted xpath is desired before committing to it in the API. It wasn't mentioned in #537 so it's hard to say..

edit: i made a poll: https://github.com/GoogleChrome/puppeteer/issues/537#issuecomment-352605428

JoelEinbinder:
@paulirish I wanted to move all the query selector code to elementHandle anyway. It comes mostly for free. And `elementHandle.xpath('..')` to get the parent feels pretty nice.

lib/FrameManager.js
```diff
@@ -408,7 +423,7 @@ class Frame {
       const url = options.url;
       try {
         const context = await this._contextPromise;
-        return await context.evaluateHandle(addScriptUrl, url);
+        return (await context.evaluateHandle(addScriptUrl, url)).asElement();
```
JoelEinbinder, line 426: I added jsdoc for _contextPromise and lots of errors showed up.

lib/FrameManager.js
JoelEinbinder: Done.

aslushnikov (approved):
Thank you!

brikou:
I think we should also have `page.$x` as we have `page.$` and `page.$$`, WDYT?

JoelEinbinder:
@brikou just as an alias or with different functionality? I think things could quickly get very cryptic.

brikou:
I don't think it is cryptic... current `page.$$`, and `page.$` exists and are meant to work as in browser (`$$`, `$`), the same should go for `$x` (and so `page.$x`).

image

JoelEinbinder:
I didn't realize we had `$x` in DevTools. Learn something new every day.

aslushnikov:
wow, that nice. I'd keep this aligned with devtools command-line api.

@JoelEinbinder wdyt about renaming `page.xpath` into `page.$x`?

PR #1621, opened 2017-12-19
fix: convert all getters to methods

### LINKED ISSUE #280: API consistency: property getters vs methods
Conversation from https://github.com/GoogleChrome/puppeteer/pull/272#discussion_r133332351. Some properties of the public API are exposed as methods:

- `browser.version()`
- `page.protocolWSEndpoint()`, `page.url()`, `page.viewport()`
- `frame.title()`, `frame.url()`, `frame.name()`
- ...

Others are properties:

- `page.mouse`, `page.tracing`
- `dialog.type`
- `request.url`, `request.url`
- `response.url`
- ...

Q: Is there a reason all of these aren't ES5 getters? At the very least, we should be consistent across the API surface.

aslushnikov:
The patch converts all the getters in the codebase into the methods.
For example, the `request.url` getter becomes the `request.url()`
method.

This is done in order to unify the API and make it more predictable.
The general rule for all further changes would be:
- there are no getters/fields exposed in the api
- the only exceptions are "namespaces", e.g. `page.keyboard`

Fixes #280.

BREAKING CHANGE:
This patch ditches getters and replaces them with methods throughout
the API. The following methods were added instead of the fields:
- dialog.type()
- consoleMessage.args()
- consoleMessage.text()
- consoleMessage.type()
- request.headers()
- request.method()
- request.postData()
- request.resourceType()
- request.url()
- response.headers()
- response.ok()
- response.status()
- response.url()

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -113,11 +113,11 @@
   * [dialog.defaultValue()](#dialogdefaultvalue)
   * [dialog.dismiss()](#dialogdismiss)
   * [dialog.message()](#dialogmessage)
-  * [dialog.type](#dialogtype)
+  * [dialog.type()](#dialogtype)
 - [class: ConsoleMessage](#class-consolemessage)
-  * [consoleMessage.args](#consolemessageargs)
-  * [consoleMessage.text](#consolemessagetext)
-  * [consoleMessage.type](#consolemessagetype)
+  * [consoleMessage.args()](#consolemessageargs)
+  * [consoleMessage.text()](#consolemessagetext)
+  * [consoleMessage.type()](#consolemessagetype)
 - [class: Frame](#class-frame)
   * [frame.$(selector)](#frameselector)
   * [frame.$$(selector)](#frameselector)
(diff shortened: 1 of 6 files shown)


JoelEinbinder:
Do any examples need to be changed?

JoelEinbinder (approved):

zyrolasting:
@JoelEinbinder @aslushnikov  There's a [missed spot in the docs](https://github.com/GoogleChrome/puppeteer/blame/master/docs/api.md#L407). I just got bit by this.

Happy to fix, but do I use this issue for the PR or a new one?

JoelEinbinder:
Good catch. Referencing issue #280 would be great. Thanks!

zyrolasting:
You got it.

niieani:
I was pretty sad to learn this change had happened. I used to be able to destructure contents of a `request` and do clean and concise things like:

```ts
page.on('response', ({url, headers, method}) => log({level: 'info', url, headers, method}))
```

Sadly, it is not possible to be as concise anymore, not even if trying to invoke the methods, since they are not bound to the `Response`'s class instance. Couldn't we have unified the API _towards_ getters, rather than _away_ from them (as was originally requested in the #280 issue by @ebidel)?

The same code now has to be much longer, taking more mental effort to read:

```ts
page.on('response', (response) => log({level: 'info', url: response.url(), headers: response.headers(), method: response.method()}))
```

In my opinion, it would be better to have an API where all the synchronously available information is available as getters/properties, while all the asynchronously available information is available as methods. **That** would be real _consistency_, and not just _consistency for consistency's sake_.

aslushnikov:
@niieani destructuring is a nice usecase we didn't think about. I'm sorry we broke it.

> In my opinion, it would be better to have an API where all the synchronously available information is available as getters/properties, while all the asynchronously available information is available as methods. That would be real consistency, and not just consistency for consistency's sake.

For now, v1.0 is already out and we'll stick with the current approach.

For future major releases, I'm not sure if migrating to the suggested approach worth the effort for both us and out clients: the pattern you've mentioned here didn't show up much on our bugtracker.

PR #1623, opened 2017-12-19
feat(Launcher): allow the default arguments to be overridden

### LINKED ISSUE #872: Add an API to override default launcher arguments
There are certain usecases where it is needed to override default puppeteer launcher flags.

There should be a way to do that.

Suggestion:
```js
// Get default launcher flags.
const myArgs = puppeteer.defaultArgs();
// Add/remove arguments as needed.
myArgs = myArgs.filter(argsFilter).concat(secretArgs);
// Launch with exactly these arguments.
  disableDefaultArgs: true, // do not use default args to launch puppeteer
  args: myArgs
});
```

JoelEinbinder:
Closes #872

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -257,7 +257,7 @@ This methods attaches Puppeteer to an existing Chromium instance.
   - `executablePath` <[string]> Path to a Chromium or Chrome executable to run instead of bundled Chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
   - `slowMo` <[number]> Slows down Puppeteer operations by the specified amount of milliseconds. Useful so that you can see what is going on.
   - `args` <[Array]<[string]>> Additional arguments to pass to the browser instance. List of Chromium flags can be found [here](http://peter.sh/experiments/chromium-command-line-switches/).
-  - `disableDefaultArgs` <[boolean]> Do not use [`puppeteer.defaultArgs()`](#puppeteerdefaultargs). Defaults to `false`.
+  - `ignoreDefaultArgs` <[boolean]> Do not use [`puppeteer.defaultArgs()`](#puppeteerdefaultargs). Dangerous option; use with care. Defaults to `false`.
   - `handleSIGINT` <[boolean]> Close browser process on Ctrl-C. Defaults to `true`.
   - `handleSIGTERM` <[boolean]> Close browser process on SIGTERM. Defaults to `true`.
   - `handleSIGHUP` <[boolean]> Close browser process on SIGHUP. Defaults to `true`.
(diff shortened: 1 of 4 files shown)


aslushnikov (approved):
@paulirish @ebidel for bikeshedding
otherwise lgtm!

paulirish:
We bikeshedded it quite recently: https://github.com/GoogleChrome/chrome-launcher/issues/65

Our [winning proposal](https://github.com/GoogleChrome/chrome-launcher/issues/65#issuecomment-349161118) (in puppeteer-speak) is:

```js
const args = puppeteer.defaultArgs(); // Map of all the flags it uses by default

flags.delete('--disable-extensions');
flags.delete('--mute-audio');
flags.set('--user-data-dir', somepath);

// pass in map to override all args
puppeteer.launch({args});
```

```js
// no regressions for adding additional args (when args is provided an array)
puppeteer.launch({args: ['--show-paint-rects']});
```

JoelEinbinder:
Paul and I talked in person and agreed that the use cases are sufficiently different to warrant different design.

docs/api.md
```diff
@@ -253,6 +257,7 @@ This methods attaches Puppeteer to an existing Chromium instance.
   - `executablePath` <[string]> Path to a Chromium or Chrome executable to run instead of bundled Chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
   - `slowMo` <[number]> Slows down Puppeteer operations by the specified amount of milliseconds. Useful so that you can see what is going on.
   - `args` <[Array]<[string]>> Additional arguments to pass to the browser instance. List of Chromium flags can be found [here](http://peter.sh/experiments/chromium-command-line-switches/).
+  - `disableDefaultArgs` <[boolean]> Do not use [`puppeteer.defaultArgs()`](#puppeteerdefaultargs). Defaults to `false`.
```
paulirish: feels like you want to say more here like....

> Do not use any of the default puppeteer args for Chrome (see `puppeteer.defaultArgs()`](#puppeteerdefaultargs)). Dangerous option; use with care. Defaults to `false`.

lib/Launcher.js
```diff
@@ -61,10 +61,13 @@ class Launcher {
   static async launch(options) {
     options = Object.assign({}, options || {});
     let temporaryUserDataDir = null;
-    const chromeArguments = [].concat(DEFAULT_ARGS);
+    const chromeArguments = [];
+    if (!options.disableDefaultArgs)
```
paulirish: in chrome-launcher we're calling it `ignoreDefaultArgs`. i do think the `ignore` verb is more appropriate, especially when we're talking about a collection of things, and not a singular item.

docs/api.md

lib/Launcher.js
JoelEinbinder: Done.

paulirish (approved):

aslushnikov (approved):

PR #1624, opened 2017-12-19
fix(Page): Fail on Content Security Policy by detecting Log.entryAdded

### LINKED ISSUE #1229: Page with CSP breaks puppeteer

yujiosaka:
A better solution to detect CSP errors than https://github.com/GoogleChrome/puppeteer/pull/1611

Since @paulirish 's [idea](https://github.com/GoogleChrome/puppeteer/pull/1611#issuecomment-352618025) is much better than mine, so I implemented it
If it looks ok, I'll close https://github.com/GoogleChrome/puppeteer/pull/1611

Also, thanks @JoelEinbinder @dios-david for comments!

Fixes: https://github.com/GoogleChrome/puppeteer/issues/1229

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -36,6 +36,8 @@ class FrameManager extends EventEmitter {
     this._frames = new Map();
     /** @type {!Map<string, !ExecutionContext>} */
     this._contextIdToContext = new Map();
+    /** @type {!Error} */
+    this._securityError = null;
 
     this._client.on('Page.frameAttached', event => this._onFrameAttached(event.frameId, event.parentFrameId));
     this._client.on('Page.frameNavigated', event => this._onFrameNavigated(event.frame));
@@ -44,6 +46,7 @@ class FrameManager extends EventEmitter {
     this._client.on('Runtime.executionContextDestroyed', event => this._onExecutionContextDestroyed(event.executionContextId));
     this._client.on('Runtime.executionContextsCleared', event => this._onExecutionContextsCleared());
     this._client.on('Page.lifecycleEvent', event => this._onLifecycleEvent(event));
+    this._client.on('Log.entryAdded', event => this._onEntryAdded(event.entry));
 
     this._handleFrameTree(frameTree);
   }
@@ -105,7 +108,7 @@ class FrameManager extends EventEmitter {
       return;
     console.assert(parentFrameId);
     const parentFrame = this._frames.get(parentFrameId);
-    const frame = new Frame(this._client, this._page, parentFrame, frameId);
+    const frame = new Frame(this._client, this._page, parentFrame, frameId, this);
     this._frames.set(frame._id, frame);
     this.emit(FrameManager.Events.FrameAttached, frame);
   }
@@ -132,7 +135,7 @@ class FrameManager extends EventEmitter {
         frame._id = framePayload.id;
       } else {
         // Initial main frame navigation.
-        frame = new Frame(this._client, this._page, null, framePayload.id);
+        frame = new Frame(this._client, this._page, null, framePayload.id, this);
       }
       this._frames.set(framePayload.id, frame);
       this._mainFrame = frame;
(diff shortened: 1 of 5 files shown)


lib/FrameManager.js
```diff
@@ -637,6 +629,20 @@ class Frame {
...
+   * @param {function(string)} pageFunction
+   * @param {!string} arg
+   * @return {!Promise<!ElementHandle>}
+   */
+  async _evaluateHandle(pageFunction, arg) {
+    const context = await this._contextPromise;
+    this._page._securityError = null;
+    const element = await context.evaluateHandle(pageFunction, arg);
+    if (this._page._securityError)
+      throw this._page._securityError;
```
aslushnikov: this `securityError` is not necessarily generated by the evaluation above.
In order to link one to another, you'd need to label evaluation with unique sourceURL and match
log entries by this sourceURL.

Do you want to proceed with this? Feel free to ask me any questions.
Let me know if you don't have time to deal with this: I'd like to have this landed and I'll take this over.

yujiosaka: @aslushnikov
Thanks! Let me handle this for a day or two because I can learn a lot from you in this way. I'll ask you if it's difficult :)

yujiosaka:
@aslushnikov 

I need your help.

> In order to link one to another, you'd need to label evaluation with unique sourceURL and match

By `sourceURL` you mean `//# sourceURL=` comment on the inserted script?
If so, I cannot use this for the following two reasons:

1. when adding a script directly with content like `addScriptTag("alert(1);")`, the `sourceURL` isn't there.
2. `Log.entryAdded` does not inform the content. It's something like this:
```
```

yujiosaka:
@aslushnikov 

I need your help.

> In order to link one to another, you'd need to label evaluation with unique sourceURL and match

By `sourceURL` you mean `//# sourceURL=` comment on the inserted script?
If so, I cannot use that for the following two reasons:

1. When adding a script tag directly with content like `Page.addScriptTag("alert(1);")`, the `sourceURL` isn't there.
2. `Log.entryAdded` does not inform the added content. Instead, it tells me something like this:
```
{ source: 'security',
  level: 'error',
  text: 'Refused to apply inline style because it violates the following Content Security Policy directive: "default-src \'self\'". Either the \'unsafe-inline\' keyword, a hash (\'sha256-F602irze5WpgKxQ4WBmbH1lheK/4Ejv/Vaxh9ajWedQ=\'), or a nonce (\'nonce-...\') is required to enable inline execution. Note also that \'style-src\' was not explicitly set, so \'default-src\' is used as a fallback.\n',
  timestamp: 1514628626813.12,
  lineNumber: 5,
  stackTrace: { [ { functionName: 'addStyleContent',
       scriptId: '21',
       url: '',
       lineNumber: 5,
       columnNumber: 20 } ] }
```

One idea I came up with is to use `functionName` in the stack trace.
My implementation goes like this:

```
  async addStyleTag(options) {

    ...

    function addStyleContent(content, hash) {
      Object.defineProperty(addStyleContent, 'name', {value: `addStyleContent_${hash}`}); // Set unique function name
      const style = document.createElement('style');
      style.type = 'text/css';
      document.head.appendChild(style);
      return style;
    }
  }

  ...

  async _evaluateHandle(pageFunction, content) {
    const context = await this._contextPromise;
    this._page._securityError = null;
    const hash = helper.sha256(content); // TODO: implement
    const element = await context.evaluateHandle(pageFunction, content, hash);
    if (this._page._cspLog && hash === this._page._cspLog.hash)
      throw new Error(this._page._cspLog.message);
    return element;
  }
```

It works perfectly fine, but it's a bit too hack-ish.
Is there a better way, or am I missing something?

aslushnikov:
@yujiosaka I apologize for the long response.

> By sourceURL you mean //# sourceURL= comment on the inserted script?

Yes, exactly. I'd expect `logEntry.url` to be set to the script sourceURL  value.

> One idea I came up with is to use functionName in the stack trace.

This is a nice idea; the only catch here is to generate a unique'ish function name. I'd still go with `sourceURL` approach if it works since it's the usual way we do these things.

yujiosaka:
@aslushnikov

This is the best I could think of. https://github.com/GoogleChrome/puppeteer/pull/1624/commits/18893ad6a80dae4d2480cbc6387bbcc3bb11ecba

I couldn't find a way to set `logEntry.url` except by magic like modifying the function name.
Instead, I simply compared the text from the log.

Please take a look.
This is not what you exactly recommended, so you can close this PR if there is a better way ;)

aslushnikov:
Closing this in favor of #2320.

PR #1631, opened 2017-12-20
feat: add device iphonex and iphonex landscape

x-cold:


x-cold:
I signed it!

DeviceDescriptors.js
```diff
@@ -327,6 +327,30 @@ module.exports = [
       'isLandscape': true
     }
   },
+  {
+    'name': 'iPhone X',
+    'userAgent': 'Mozilla/5.0 (iPhone; CPU iPhone OS 10_3 like Mac OS X) AppleWebKit/602.1.50 (KHTML, like Gecko) CriOS/56.0.2924.75 Mobile/14E5239e Safari/602.1',
+    'viewport': {
+      'width': 375,
+      'height': 734,
```
aslushnikov: for some reason we have different height in devtools emulation: https://cs.chromium.org/chromium/src/third_party/WebKit/Source/devtools/front_end/emulated_devices/module.json?l=175-209

x-cold: We consider the safe area for Iphone X as following image:

image

Actually, chrome emulation usees the viewport `812*375` for Iphone X, I will modify the values and bring into correspondence with it.

aslushnikov (approved):

PR #1633, opened 2017-12-20
test: run chrome with --disable-dev-shm-usage

aslushnikov:
This might speedup tests on CI.


ebidel (approved):

paambaati:
@aslushnikov @ebidel Would having this as a default flag help? Can you guys expand a little bit on scenarios where this is beneficial?

ebidel:
@paambaati that seems like a great suggestion. Want to file an issue?

Issue #1579, opened 2017-12-11
Attribute network requests to frames


aslushnikov:
We should attribute network requests to frames so that we know where requests are coming from.

```js
const frame = request.frame();
```

This would simplify network traffic analysis.

Open question: it might make sense to expose execution contexts instead to attribute requests coming from workers.

yujiosaka:
+1

Use case: I was writing a patch to disable/relax CSP by reading a part from response headers, but it was hard to follow frame -> request -> response. It would be a nice feature to easily access the property like that.

PR #1646, opened 2017-12-21
 Attribute network requests to frames (#1579)

yujiosaka:
I think it's not only convenient for requests to have frame properties, but it's consistent with Chrome DevTool Protocol because [Network.requestWillBeSent](https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-requestWillBeSent) event has frameId property.

One problem I faced is that the frameId is optional, so the frame cannot always be accessed from request. I noticed that the property is empty for offline mode and when navigating to bad SSL pages.

Fixes: https://github.com/GoogleChrome/puppeteer/issues/1579

@aslushnikov

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -177,6 +177,7 @@
   * [request.abort([errorCode])](#requestaborterrorcode)
   * [request.continue([overrides])](#requestcontinueoverrides)
   * [request.failure()](#requestfailure)
+  * [request.frame()](#requestframe)
   * [request.headers()](#requestheaders)
   * [request.method()](#requestmethod)
   * [request.postData()](#requestpostdata)
@@ -2042,6 +2043,9 @@ page.on('requestfailed', request => {
   console.log(request.url + ' ' + request.failure().errorText);
 });
 ```
+
+#### request.frame()
+- returns: <?[Frame]> A matching [Frame] object, or `null` for offline mode or navigating to error pages.
 
 #### request.headers()
 - returns: <[Object]> An object with HTTP headers associated with the request. All header names are lower-case.
--- a/docs/api.md
+++ b/docs/api.md
@@ -2045,7 +2045,7 @@ page.on('requestfailed', request => {
 ```
 
 #### request.frame()
-- returns: <?[Frame]> A matching [Frame] object, or `null` for offline mode or navigating to error pages.
+- returns: <?[Frame]> A matching [Frame] object, or `null` if navigating to error pages.
 
 #### request.headers()
 - returns: <[Object]> An object with HTTP headers associated with the request. All header names are lower-case.
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -88,6 +88,14 @@ class FrameManager extends EventEmitter {
   }
 
   /**
+   * @param {!string} frameId
+   * @return {!Frame}
+   */
+  frame(frameId) {
+    return this._frames.get(frameId);
+  }
+
+  /**
    * @param {string} frameId
    * @param {?string} parentFrameId
    * @return {?Frame}
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -92,7 +92,7 @@ class FrameManager extends EventEmitter {
    * @return {?Frame}
    */
   frame(frameId) {
-    return this._frames.get(frameId);
+    return this._frames.get(frameId) || null;
   }
 
   /**
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -89,7 +89,7 @@ class FrameManager extends EventEmitter {
 
   /**
    * @param {!string} frameId
-   * @return {!Frame}
+   * @return {?Frame}
    */
   frame(frameId) {
     return this._frames.get(frameId);
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -89,10 +89,10 @@ class FrameManager extends EventEmitter {
 
   /**
    * @param {!string} frameId
-   * @return {!Frame}
+   * @return {?Frame}
    */
   frame(frameId) {
-    return this._frames.get(frameId);
+    return this._frames.get(frameId) || null;
   }
 
   /**
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -89,10 +89,10 @@ class FrameManager extends EventEmitter {
 
   /**
    * @param {!string} frameId
-   * @return {?Frame}
+   * @return {!Frame}
    */
   frame(frameId) {
-    return this._frames.get(frameId) || null;
+    return this._frames.get(frameId);
   }
 
   /**
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -285,7 +285,7 @@ class Request {
    * @param {string} url
    * @param {string} resourceType
    * @param {!Object} payload
-   * @param {Puppeteer.Frame} frame
+   * @param {?Puppeteer.Frame} frame
    */
   constructor(client, requestId, interceptionId, allowInterception, url, resourceType, payload, frame) {
     this._client = client;
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -19,11 +19,13 @@ const Multimap = require('./Multimap');
 
 class NetworkManager extends EventEmitter {
   /**
-   * @param {Puppeteer.Session} client
+   * @param {!Puppeteer.Session} client
+   * @param {!Puppeteer.FrameManager} frameManager
    */
-  constructor(client) {
+  constructor(client, frameManager) {
     super();
     this._client = client;
+    this._frameManager = frameManager;
     /** @type {!Map<string, !Request>} */
     this._requestIdToRequest = new Map();
     /** @type {!Map<string, !Request>} */
@@ -189,7 +191,8 @@ class NetworkManager extends EventEmitter {
    * @param {string} frameId
    */
   _handleRequestStart(requestId, interceptionId, url, resourceType, requestPayload, frameId) {
-    const request = new Request(this._client, requestId, interceptionId, this._userRequestInterceptionEnabled, url, resourceType, requestPayload, frameId);
+    const frame = this._frameManager._frames.get(frameId);
+    const request = new Request(this._client, requestId, interceptionId, this._userRequestInterceptionEnabled, url, resourceType, requestPayload, frame);
     if (requestId)
       this._requestIdToRequest.set(requestId, request);
     if (interceptionId)
@@ -282,9 +285,9 @@ class Request {
    * @param {string} url
    * @param {string} resourceType
    * @param {!Object} payload
-   * @param {string} frameId
+   * @param {!Puppeteer.Frame} frame
    */
-  constructor(client, requestId, interceptionId, allowInterception, url, resourceType, payload, frameId) {
+  constructor(client, requestId, interceptionId, allowInterception, url, resourceType, payload, frame) {
     this._client = client;
     this._requestId = requestId;
     this._interceptionId = interceptionId;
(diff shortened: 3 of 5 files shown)


lib/Page.js
```diff
@@ -98,6 +100,16 @@ class Page extends EventEmitter {
     client.on('Performance.metrics', event => this._emitMetrics(event));
   }
 
+  /**
+   * @param {!string} eventName
+   * @param {!Object} event
+   */
+  _handleRequest(eventName, event) {
```
aslushnikov: let's pass FrameManager over to the NetworkManager. This will make it easier to attribute requests with frames.

So in page constructor you'll have:

```js
this._frameManager = new FrameManager(client, frameTree, this);
this._networkManager = new NetworkManager(client, this._frameManager);  
```

and this code will move to the NetworkManager.

test/test.js
```diff
@@ -2530,6 +2533,7 @@ describe('Page', function() {
       expect(failedRequests[0].response()).toBe(null);
       expect(failedRequests[0].resourceType()).toBe('stylesheet');
       expect(failedRequests[0].failure().errorText).toBe('net::ERR_FAILED');
+      expect(failedRequests[0].frame()).toBeTruthy();
```
aslushnikov, line 2536: Lets's also check somewhere that `request.frame()` returns reasonable frame for requests from subframes.
It would be nice to check this with and without request interception enabled.

docs/api.md
```diff
@@ -2043,6 +2044,9 @@ page.on('requestfailed', request => {
 });
 ```
 
+#### request.frame()
+- returns: <?[Frame]> A matching [Frame] object, or `null` for offline mode or navigating to error pages.
```
aslushnikov: I wonder why this returns `null` for offline mode, this shouldn't be the case. Maybe that's because the request events are send *before* the frames are reported to puppeteer?

aslushnikov (commented):
Thanks for taking this!

I have a few comments, but overall it looks nice!

docs/api.md
yujiosaka: I'll investigate deeper on this. I'll ask your help if it's difficult for me :)

lib/Page.js
yujiosaka: https://github.com/GoogleChrome/puppeteer/pull/1646/commits/da9a0b06f984c0c219a04f21e280c8056085604d

lib/NetworkManager.js
```diff
@@ -19,11 +19,13 @@ const Multimap = require('./Multimap');
 
 class NetworkManager extends EventEmitter {
   /**
-   * @param {Puppeteer.Session} client
+   * @param {!Puppeteer.Session} client
```
yujiosaka, line 22: Session is always passed.

lib/NetworkManager.js
```diff
@@ -186,9 +188,11 @@ class NetworkManager extends EventEmitter {
    * @param {string} url
    * @param {string} resourceType
    * @param {!Object} requestPayload
+   * @param {string} frameId
    */
-  _handleRequestStart(requestId, interceptionId, url, resourceType, requestPayload) {
-    const request = new Request(this._client, requestId, interceptionId, this._userRequestInterceptionEnabled, url, resourceType, requestPayload);
+  _handleRequestStart(requestId, interceptionId, url, resourceType, requestPayload, frameId) {
+    const frame = this._frameManager._frames.get(frameId);
```
aslushnikov: nit: can you please add a `FrameManager.frame(frameId)` method that would fetch frame for the given frameId. This is to avoid reaching into frameManager internals.

test/test.js
yujiosaka: Done!
https://github.com/GoogleChrome/puppeteer/pull/1646/commits/77e7dd1e36b19eaeafe7668d2ed34dae6b662274

test/test.js
```diff
@@ -1580,7 +1582,9 @@ describe('Page', function() {
       page.on('request', request => requests.push(request));
       await page.goto(server.EMPTY_PAGE);
       expect(requests.length).toBe(1);
-      expect(requests[0].url()).toContain('empty.html');
```
yujiosaka: Let's make it consistent with other tests.

test/test.js
aslushnikov: Thank you. Can you please also add a test with a request from non-mainframe?

lib/NetworkManager.js
yujiosaka: @aslushnikov 
How about adding a underscore prefix and making a `FrameManager._frame(frameId)` method?
When you make a method without underscore, it should be documented on API docs.
It's not strictly a private method I know, but puppeteer's users shouldn't have to know about `frameId` in my opinion.
Or else, maybe expose `frame.id()` method so that users have more control on frames.

yujiosaka: Ooops, never mind! `FrameManager` wasn't on the doc from the first place. I'll do that very quick!

yujiosaka: Done!
https://github.com/GoogleChrome/puppeteer/pull/1646/commits/82215f41627fd14f619b1ce5c33c91e67c1971a3

test/test.js
yujiosaka: Done!
https://github.com/GoogleChrome/puppeteer/pull/1646/commits/acf2c0942f541b243d293de6d196abe60512d965

test/test.js
```diff
@@ -29,6 +29,7 @@ const iPhoneLandscape = DeviceDescriptors['iPhone 6 landscape'];
 
 const SimpleServer = require('./server/SimpleServer');
 const GoldenUtils = require('./golden-utils');
+const FrameUtils = require('./frame-utils');
```
yujiosaka, line 32: I didn't find a reason no to require the util only once on the top.

lib/FrameManager.js
```diff
@@ -87,6 +87,14 @@ class FrameManager extends EventEmitter {
     return Array.from(this._frames.values());
   }
 
+  /**
+   * @param {!string} frameId
+   * @return {!Frame}
+   */
+  frame(frameId) {
+    return this._frames.get(frameId);
```
aslushnikov:
```js
 return this._frames.get(frameId) || null;
```

lib/FrameManager.js
```diff
@@ -87,6 +87,14 @@ class FrameManager extends EventEmitter {
     return Array.from(this._frames.values());
   }
 
+  /**
+   * @param {!string} frameId
+   * @return {!Frame}
```
aslushnikov: there might be no frame for the frameId

```js
/**
 * @param {!string} frameId
 * @return {?Frame}
 */
```

lib/NetworkManager.js
```diff
@@ -281,8 +285,9 @@ class Request {
    * @param {string} url
    * @param {string} resourceType
    * @param {!Object} payload
+   * @param {!Puppeteer.Frame} frame
```
aslushnikov: Frame is nullable here as well

test/test.js
aslushnikov: this is good, thanks

test/test.js
```diff
@@ -1575,17 +1573,29 @@ describe('Page', function() {
...
+    });
+
+    it('should fire on child frames', async({page, server}) => {
+      await page.goto(server.EMPTY_PAGE);
+      const requests = [];
+      page.on('request', request => requests.push(request));
+      await FrameUtils.attachFrame(page, 'frame1', server.EMPTY_PAGE);
+      expect(requests.length).toBe(1);
+      expect(requests[0].url()).toBe(server.EMPTY_PAGE);
+      expect(requests[0].frame() === page.mainFrame()).toBe(false);
```
aslushnikov: nit: let's compare this to the `page.frames()[1]`

lib/FrameManager.js
yujiosaka: nice catch!
https://github.com/GoogleChrome/puppeteer/pull/1646/commits/2ea00a341f7867a57ececf5aaec22f260fcf6217

lib/NetworkManager.js
yujiosaka: nice catch! thanks!
https://github.com/GoogleChrome/puppeteer/pull/1646/commits/31c31da50c65f8ced1cd7b49dc1e3b37bfb0bcf2

lib/FrameManager.js
yujiosaka: Thank you very much!
https://github.com/GoogleChrome/puppeteer/pull/1646/commits/6a3de8f93550fa179efc10cf58458981f9c66c36

test/test.js
yujiosaka: How about this?
https://github.com/GoogleChrome/puppeteer/pull/1646/commits/d8a2974c9496f6cfbd9e222f6c9c407ac9378c04

lib/NetworkManager.js
```diff
@@ -281,8 +285,9 @@ class Request {
    * @param {string} url
    * @param {string} resourceType
    * @param {!Object} payload
+   * @param {Puppeteer.Frame} frame
```
aslushnikov: nit: question mark is missing for the frame jsdoc annotation

test/test.js
```diff
@@ -1579,13 +1577,18 @@ describe('Page', function() {
...
       await page.goto(server.EMPTY_PAGE);
-      expect(requests.length).toBe(1);
-      expect(requests[0].url()).toContain('empty.html');
+      await FrameUtils.attachFrame(page, 'frame1', server.EMPTY_PAGE);
+      expect(requests.length).toBe(2);
+      expect(requests[0].url()).toBe(server.EMPTY_PAGE);
+      expect(requests[0].frame() === page.mainFrame()).toBe(true);
+      expect(requests[0].frame().url()).toBe(server.EMPTY_PAGE);
+      expect(requests[1].url()).toBe(server.EMPTY_PAGE);
+      expect(requests[1].frame() === page.mainFrame()).toBe(false);
```
aslushnikov: nit: can you please compare with the frames[1] instead?
```js
expect(requests[1].frame() === page.frames()[1]).toBe(true);
```

yujiosaka: Done! Thanks!
https://github.com/GoogleChrome/puppeteer/pull/1646/commits/47c93c516f5f0c4e7927f5d878272d0f4e0a61b7

lib/NetworkManager.js
yujiosaka: Did it!
https://github.com/GoogleChrome/puppeteer/pull/1646/commits/577117b362e311fcc7e471cb82b8c11ac21c575e

yujiosaka:
@aslushnikov 

I started thinking that [this annotation](https://github.com/GoogleChrome/puppeteer/pull/1646/files#diff-16f3b9739b97830927de83812ac92d78R91) is probably wrong because `frameId` was absent on offline mode and bad SSL pages when I looked into it.
But as you pointed out, I should look deeper into why and when the frame returns `null`, and fix it when I become 100% confident on this.

yujiosaka:
@aslushnikov 

> I wonder why this returns null for offline mode, this shouldn't be the case.

The fact turned out to be simple.
When it's offline mode, frameId is `null` for the following reasons:

1. request correctly fails in offline mode
2. frameId is `null` for navigating to error pages in general

When frameId is returned, I can always find a frame corresponding to the id.
Therefore, I fixed the doc, code and annotation [here](https://github.com/GoogleChrome/puppeteer/pull/1646/commits/83c0148cd09bf31af3661a5bed08f31ea932161d)

lib/FrameManager.js
aslushnikov: I'm not sure why, but this doesn't seem to be addressed yet (though I see your commit that addresses the comment)

aslushnikov (commented):
@yujiosaka looks like some of the comments are left un-addressed for some reason. Could you please take a look?

yujiosaka:
@aslushnikov 

Sorry I took back some code in [this commit](https://github.com/GoogleChrome/puppeteer/pull/1646/commits/83c0148cd09bf31af3661a5bed08f31ea932161d), but it was a mistake.

I fixed the code [here](https://github.com/GoogleChrome/puppeteer/pull/1646/commits/ba0fed3d38ce90ea32f955c40ca8d923b6208307), so can you take a look?

A test is failing, but I don't think it's related to this fix.

aslushnikov (approved):
Thank you!

PR #1652, opened 2017-12-24
CLI support on install.js

Kikobeats:
Hello,

In this PR I added the functionality for being possible to download a chromium version specifying `platform` and `revision`:

```bash
$ node install.js --platform=linux 
```

This is helpful when you need to provision production machines that normally are a different architecture than a development environment.

### DIFF
--- a/install.js
+++ b/install.js
@@ -13,6 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+const ProgressBar = require('progress');
+const mri = require('mri');
 
 buildNode6IfNecessary();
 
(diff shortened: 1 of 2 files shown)


Kikobeats:
I signed it!

aslushnikov:
Hey @Kikobeats,

Will it work for you if the [Downloader](https://github.com/GoogleChrome/puppeteer/blob/master/lib/Downloader.js) class would be exposed through pptr api? This will allow you to write any scripts to download chromium.

If this suites you well, I'd rather proceed with this approach instead of your PR for the following reasons:
- I'd like to keep `install.js` as simple as possible
- exposed `Downloader` allows for great flexibility

aslushnikov:
Closing in favor of https://github.com/GoogleChrome/puppeteer/issues/1748. Please upvote the issue to affect our work prioritization.

PR #1658, opened 2017-12-26
fix(NavigationWatcher):correct closure annotation

yujiosaka:
`_createTimeoutPromise()` always returns a promise.

It changed [here](https://github.com/GoogleChrome/puppeteer/pull/1435).


JoelEinbinder (approved):
Thanks!

PR #1661, opened 2017-12-26
Add nativeVirtualKeyCode to key event

minodisk:
I added [`nativeVirtualKeyCode`](https://chromedevtools.github.io/devtools-protocol/tot/Input/#method-dispatchKeyEvent) as described in https://github.com/GoogleChrome/puppeteer/issues/776#issuecomment-329589760 and confirmed that shortcuts work properly under certain conditions on OSX.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1252,7 +1252,13 @@ await page.keyboard.press('KeyA');
 await page.keyboard.up('Shift');
 ```
 
-> **NOTE** On MacOS, keyboard shortcuts like `⌘ A` -> Select All do not work. See [#1313](https://github.com/GoogleChrome/puppeteer/issues/1313)
+> **NOTE** On MacOS, keyboard shortcuts like `⌘ A` -> There are some conditions to work properly. See [#776](https://github.com/GoogleChrome/puppeteer/issues/776#issuecomment-329589760) and the code below:
+```js
+const browser = await puppeteer.launch({headless: false}); // headless: false is required
+const page = await browser.newPage();
+await page.bringToFront();                                 // required
+await page.keyboard.press('Meta', {text: 'a'});
+```
 
 #### keyboard.down(key[, options])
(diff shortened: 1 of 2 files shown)


JoelEinbinder:
The problem with adding `nativeVirtualKeyCode` is that its too good. It will sometimes perform actions outside of Puppeteer, which can be dangerous.

JoelEinbinder:
I tried this again, and I wasn't able to get actions to happen outside of Puppeteer so I may have misremembered that. There is a scary "thunk" sound if Puppeteer's Chromium isn't the active window, but that's it.

You are using the Windows keycode, but Mac has a different set of keycodes that would need to be passed in as nativeVirtualKeyCode. It happens to work anyway for most things, because Mac more often reads the text instead of the keycode, but that shouldn't be relied upon.

This works better than it use to now that `bringToFront` actually focuses the web contents, but I still don't think it is worth adding. It needs a very specific set of conditions to work right: headful Chrome, Chrome window focused, web contents focused, and keyboard commands are not overridden. And the cost of it is increasing the difference between headless and headful modes, maintaining a list of all mac keycodes, and adding OS detection. I am particularly worried about the the effects something like `Cmd+Q` could have when trying to debug an application in headful mode. Especially if the user remaps their shortcuts.

Closing for now.

minodisk:
Thank you for your review and trying. I could understand that there are many things to think and I learned a lot. :thinking:

I was completely satisfied with what you wrote, but I think it's painful to have no way to call shortcuts with macOS. How about adding an option to transparently pass `nativeVirtualKeyCode` like below?

```js
  async down(key, options = { text: undefined, nativeVirtualKeyCode: undefined }) {
    ...
    await this._client.send('Input.dispatchKeyEvent', {
      ...
      nativeVirtualKeyCode: nativeVirtualKeyCode,
      ...
    });
  }
```

I think Puppeteer will not have to take responsibility for what happens, but what do you think?

PR #1674, opened 2017-12-28
test: do not leave dangling listener on shared browser object

aslushnikov:
This resulted in quite a few failures on travis.


aslushnikov:
@JoelEinbinder FYI

Issue #1659, opened 2017-12-26
Possible race condition when simulating double click - dedicated API method needed


canonic-epicure:
It seems currently the only way to simulate double click is to call `mouse.down()/up/down/up". 

Every call will wait until previous one completes. And if the remote method call is performed using the network (like websocket) there's a possibility of arbitrary delay between the calls. 

If such delay will exceed the double click threshold - browser won't fire the "dblclick" event, only 2 regular clicks.

This is a potential race condition.

aslushnikov:
@JoelEinbinder is there any way to simulate `dblclick` event? Will the [mouse.click](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#mouseclickx-y-options) with `clickCount` argument work?

JoelEinbinder:
Yep! `dblclick` is fired by `clickCount`, not by the timeout. That way we can send clicks as fast as possible without triggering double click, or send double clicks with arbitrary delay in between each click.

```js
 const button = await page.$('button');
 await button.click();
 await new Promise(x => setTimeout(x, 1000));
 await button.click({ clickCount: 2 });
```

canonic-epicure:
Good to know. This detail (setting `clickCount` for simulating double click) should be mentioned in docs for "click" method then..

stoffeastrom:
This doesn't work for us. Using `ng-dblclick` in an old angular `1.5.8`. Tried setting different `delay` for the `click` without success. The only way I could get it to work was doing:

```
await page.goto('foo-url', { waitUntil: 'networkidle0' });
const from = await page.$('.foo-item');
const fromBox = await from.boundingBox();
const fromX = fromBox.x + fromBox.width / 2;
const fromY = fromBox.y + fromBox.height / 2;
await page.evaluate((e, clientX, clientY) => {
return e.dispatchEvent(new MouseEvent('dblclick', { bubbles: true, cancelable: true, clientX, clientY }));
    }, from, fromX, fromY);
await page.waitFor('.foo-something');
```

JoelEinbinder:
@stoffeastrom can you share the puppeteer code that didn't work for you? The delay has no effect on whether or not a `dblclick` is sent. That is entirely decided by the `clickCount` field.

stoffeastrom:
Tried getting the element and explicitly call click on that
Tried using the page to click

```
const foo = await page.$('.foo-item');
await = foo.click({ clickCount: 2 });
```
```
[...]

PR #1675, opened 2017-12-28
test: add dblclick input test

JoelEinbinder:
#1659 brings up an interesting case about the `dblclick` event. We behave properly here, but were missing a test.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -2169,6 +2169,20 @@ describe('Page', function() {
       await page.click('#button-80');
       expect(await page.evaluate(() => document.querySelector('#button-80').textContent)).toBe('clicked');
     });
+    it('should double click the button', async({page, server}) => {
+      await page.goto(server.PREFIX + '/input/button.html');
+      await page.evaluate(() => {
+        window.double = false;
+        const button = document.querySelector('button');
+        button.addEventListener('dblclick', event => {
+          window.double = true;
+        });
+      });
+      const button = await page.$('button');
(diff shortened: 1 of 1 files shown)


test/test.js
```diff
@@ -2169,6 +2169,20 @@ describe('Page', function() {
...
+      await page.goto(server.PREFIX + '/input/button.html');
+      await page.evaluate(() => {
+        window.double = false;
+        const button = document.querySelector('button');
+        button.addEventListener('dblclick', event => {
+          window.double = true;
+        });
+      });
+      const button = await page.$('button');
+      await button.click();
```
aslushnikov: is this needed?

JoelEinbinder: no

test/test.js
```diff
@@ -2169,6 +2169,19 @@ describe('Page', function() {
...
+      await page.goto(server.PREFIX + '/input/button.html');
+      await page.evaluate(() => {
+        window.double = false;
+        const button = document.querySelector('button');
+        button.addEventListener('dblclick', event => {
+          window.double = true;
+        });
+      });
+      const button = await page.$('button');
+      await button.click({ clickCount: 2 });
```
aslushnikov, line 2182: another question: will it also generate a click event?

aslushnikov (approved):

test/test.js
JoelEinbinder: Yes. Added to the test.

Issue #1683, opened 2017-12-28
It would be useful to have Node.js version to the issue template


yujiosaka:
The current Issue template asks the following questions.

* Puppeteer version:
* Platform / OS version:
* URLs (if applicable):

I noticed some folks adds Node.js version to this list, and I found it very useful.
It would be nice have it in the template.

PR #1684, opened 2017-12-28
feat(doc): Add Node.js version to Issue template(#1683)

yujiosaka:
Just a quick idea, which you can trash if you don't like :)

Fixes: https://github.com/GoogleChrome/puppeteer/issues/1683


aslushnikov (approved):
I like it, thanks!

PR #1692, opened 2017-12-29
Add the ability to specify padding for elementHandle.screenshot()

Glennvd:
Proposal to add an easy option to add padding to the elementHandle.screenshot() function. I'm not quite sure yet how the api interface should look, but this is a first working proposal.

### DIFF
--- a/lib/ElementHandle.js
+++ b/lib/ElementHandle.js
@@ -152,6 +152,38 @@ class ElementHandle extends JSHandle {
     const clip = Object.assign({}, boundingBox);
     clip.x += pageX;
     clip.y += pageY;
+
+    if (options.padding) {
+      if(options.padding.constructor !== Array || options.padding.length > 4) {
+          throw Error('Invalid padding parameter (should be [top, right, bottom, left])')
+      }
+      switch (options.padding.length){
+          case 1:
+              clip.x -= options.padding[0];
+              clip.y -= options.padding[0];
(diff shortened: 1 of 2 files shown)


Glennvd:
I signed it (covered under corporate agreement)

lib/ElementHandle.js
```diff
@@ -152,6 +152,37 @@ class ElementHandle extends JSHandle {
...
+
+    if (options.padding) {
+      if (options.padding.constructor !== Array || options.padding.length > 4)
+          throw Error('Invalid padding parameter (should be [top, right, bottom, left])');     
+      switch (options.padding.length){
+        case 1:
+          clip.x -= options.padding[0];
+          clip.y -= options.padding[0];
+          clip.width += options.padding[0];
+          clip.height += options.padding[0];
```
JoelEinbinder: Shouldn't this be:
`clip.height += options.padding[0] + options.padding[0]`?
A 2x2 square with a padding of 3 should be an 8x8 square.

JoelEinbinder:
Thanks for the PR! What is your use case for this? It seems like it could be done currently by using `element.boundingBox()` and `page.screenshot()`. I am not sure if it is worth it to add a shortcut to `element.screenshot()` for this.

aslushnikov:
This looks easily polyfillable to me. Please feel free to object if there are good reasons why this should be a part of the api.

PR #1702, opened 2018-01-01
Update block-images.js

vpotseluyko:
Maybe you have changed api, or..

### DIFF
--- a/examples/block-images.js
+++ b/examples/block-images.js
@@ -24,7 +24,7 @@ const browser = await puppeteer.launch();
 const page = await browser.newPage();
 await page.setRequestInterception(true);
 page.on('request', request => {
-  if (request.resourceType() === 'image')
+  if (request.resourceType === 'image')
     request.abort();
   else
     request.continue();


vsemozhetbyt:
See https://github.com/GoogleChrome/puppeteer/commit/b73737302a5dbfc5cd67f37be8460e53ab4f417f

vpotseluyko:
So, is my commit is correct. :)

vsemozhetbyt:
@vpotseluyko I think the example reflects the tip of tree (master) state, it will work properly if a user upgrades to the v1.0.0-rc and it is also in sync with the current https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md

vpotseluyko:
Oh, understood. Extremely sorry.

vsemozhetbyt:
@vpotseluyko Do not be discouraged, it is a common thing in this repository, docs and examples always outrun a bit the last npm version, so many people get confused)

PR #1710, opened 2018-01-03
feat(Page): add Page.wsEndpoint, for monitoring / debugging.

pwnall:
I'm using Puppeteer for parallel scraping in a personal project. I have one master connected to 25 Chrome instances running on separate worker machines, each of which might have multiple tabs open.

I've been using a hack that pokes into the private fields of Page and Session to get the equivalent of this PR, for monitoring purposes. I have a scheduler that keeps track of available pages, and assigns pages to tasks. When a task errors out, having the /devtools/page URL allows me to connect directly to the tab and see what happened.

I'm submitting this PR hoping that others will find the feature useful, and that I won't have to maintain my hack :smile:

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -96,6 +96,7 @@
   * [page.waitForFunction(pageFunction[, options[, ...args]])](#pagewaitforfunctionpagefunction-options-args)
   * [page.waitForNavigation(options)](#pagewaitfornavigationoptions)
   * [page.waitForSelector(selector[, options])](#pagewaitforselectorselector-options)
+  * [page.wsEndpoint()](#pagewsendpoint)
 - [class: Keyboard](#class-keyboard)
   * [keyboard.down(key[, options])](#keyboarddownkey-options)
   * [keyboard.press(key[, options])](#keyboardpresskey-options)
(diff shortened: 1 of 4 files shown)


aslushnikov:
Hey @pwnall, thanks for the PR, and sorry for the long response.

We finally have a reasonable solution for the #31, so you'll get a raw devtools protocol. Will it be good enough for you?

pwnall:
@aslushnikov Thank you for getting back to me!

How would the raw devtools be exposed? If this entails exposing the Connection and targetId, I'm perfectly fine with munging the URL on my own.

To clarify, I'm interested in this information from a monitoring + debugging standpoint. I have a node.js server that connects to a cluster of VMs running Chrome and uses them in a complex scraping job with millions of tasks that get assigned to individual tabs. The server exposes a status page with information about each puppeteer.Page that includes the number of tasks, the start / end of last scraping task, and any errors. If a Page is wedged (last task has been going on for a long time, or the number of tasks is unusually low), I want a URL that I can connect to and diagnose.

Ideally, I'd connect to the relevant Chrome tab directly, without having to write extra logic in my server. I was exposing the page's WebSocket URL because it can be used for that purpose. Getting equivalent information from the API would be fine. Getting a connection (without an URL) that I can speak raw devtools to would make my life a bit more difficult.

Sorry if this was obvious. I explained because I think that the other use cases in #31 are fine with a pipe that speaks devtools. Thank you very much for reading!

aslushnikov:
@pwnall thank you Victor for the detailed explanation.

As per offline discussion, I'll close this for now. Hopefully, hopefully the newly introduced `target.createCDPSession()` should help.

PR #1711, opened 2018-01-03
docs(API): add more explicit examples for page.evaluate

ghost:
I'm pushing this PR to add more explicit examples in the API documentation of how can one use ```page.evaluate``` with parameters.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -663,18 +663,20 @@ List of all available devices is available in the source code: [DeviceDescriptor
 If the function, passed to the `page.evaluate`, returns a [Promise], then `page.evaluate` would wait for the promise to resolve and return its value.
 
 If the function passed into `page.evaluate` returns a non-[Serializable] value, then `page.evaluate` resolves to `undefined`.
+Passing arguments to ```pageFunction```.
 
 ```js
-const result = await page.evaluate(() => {
-  return Promise.resolve(8 * 7);
-});
+const result = await page.evaluate(x => {
+  return Promise.resolve(8 * x);
+}, 7);
 console.log(result); // prints "56"
 ```
 
-A string can also be passed in instead of a function.
+A template string can also be passed in instead of a function.
 
 ```js
-console.log(await page.evaluate('1 + 2')); // prints "3"
+const x = 10;
+console.log(await page.evaluate(`1 + ${x}`)); // prints "11"
 ```
 
 [ElementHandle] instances can be passed as arguments to the `page.evaluate`:
--- a/docs/api.md
+++ b/docs/api.md
@@ -672,9 +672,10 @@ const result = await page.evaluate(x => {
 console.log(result); // prints "56"
 ```
 
-A template string can also be passed in instead of a function.
+A string can also be passed in instead of a function.
 
 ```js
+console.log(await page.evaluate('1 + 2')); // prints "3"
 const x = 10;
 console.log(await page.evaluate(`1 + ${x}`)); // prints "11"
 ```


docs/api.md
```diff
@@ -671,12 +671,68 @@ const result = await page.evaluate(() => {
...
+const age  = 29;
+const location = 'Paris';
+
+const information = await page.evaluate((name, age, location) => {
+  return {
+    nameValue: name,
+    ageValue: age,
+    locationValue: location
+  };
+}, {name, age, location});
```
aslushnikov: don't you want to pass these arguments one-by-one, e.g.

```js
const information = await page.evaluate((name, age, location) => {
  return {nameValue: name, ageValue: age, locationValue: location};
}, name, age, location);
```

This will print it correctly.

docs/api.md
```diff
@@ -671,12 +671,68 @@ const result = await page.evaluate(() => {
...
+
+console.log(information); // prints: "{ nameValue: { name: 'Thomas', age: 29, location: 'Paris' }, ageValue: null, locationValue: null }"
+```
+
+Passing pageFunction to page.evaluate as parameter.
+```js
+function manipulateDOM(){
```
aslushnikov: you can pass arguments here exactly in the same fashion: 
```js
function manipulateDOM(arg) {
  return 'puppeteer was here and received ' + arg;
}

const information = await page.evaluate(manipulateDOM, 42);
console.log(information);

File: docs/api.md (hunk 1 above)
Comment (ghost) [reply docs/api.md]: Commit amended.

File: docs/api.md (hunk 2 above)
Comment (ghost) [reply docs/api.md]: @aslushnikov this is exactly what I was missing. I've amended the commit. Thank you!

**Comment by ghost** [comment]
@aslushnikov can you please review this? Let me know if anything else is needed.

File: docs/api.md [hunk 3]
```diff
@@ -671,12 +671,51 @@ const result = await page.evaluate(() => {
 console.log(result); // prints "56"
 ```
 
+Passing arguments to pageFunction.
+
+```js
+const name = 'Thomas';
```
aslushnikov: Let's amend the example on line 668 instead so that it accepts an argument:
```js
const result = await page.evaluate(x => {
  return Promise.resolve(8 * x);
}, 7);
console.log(result); // prints "56"
```
 
This should cover this and the following examples.

docs/api.md
```diff
@@ -671,12 +671,51 @@ const result = await page.evaluate(() => {
...
+
+const information = await page.evaluate(manipulateDOM, 42);
+
+console.log(information); // prints: "puppeteer was here and received 42";
+```
+
 A string can also be passed in instead of a function.
 
 ```js
 console.log(await page.evaluate('1 + 2')); // prints "3"
```
aslushnikov: let's incorporate template string example here and demonstrate string interpolation:
```js
const x = 10;
console.log(await page.evaluate(`1 + ${x}`)); // prints "11"
```

docs/api.md
```diff
@@ -671,12 +671,51 @@ const result = await page.evaluate(() => {
...
+
+Passing both ```pageFunction``` and ```...args``` to ```page.evaluate``` as parameters.
+
+```js
+function manipulateDOM(arg){
+  // manipulate DOM logic
+  return `puppeteer was here and received ${arg}`;
+}
+
+const information = await page.evaluate(manipulateDOM, 42);
```
aslushnikov: This is more a javascript example then a puppeteer one: as in other examples, we get a function as the first argument and arguments as the rest. I'd remove it.

aslushnikov (commented):
@octasimo I agree that examples are a little lacking, but I'd like to keep documentation concise so that it's fast to look through.

I left a few comments that suggest to re-organize current examples so that your usecases are covered. Let me know what you think.

ghost:
Works for me. I've pushed the amended commit.

aslushnikov (approved):
@octasimo thanks. I did a small changes to the second example, hope you don't mind. This is ready to be merged as soon as the CI bots cycle.

PR #1716, opened 2018-01-04
Browser user agent

yujiosaka:
- Add `browser.userAgent()`
- cache `Browser.getVersion` response

Fixes: https://github.com/GoogleChrome/puppeteer/issues/1706

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -364,9 +364,9 @@ Disconnects Puppeteer from the browser, but leaves the Chromium process running.
 - returns: <[Array]<[Target]>> An array of all active targets.
 
 #### browser.userAgent()
-- returns: <[Promise]<[string]>> Promise which resolves to the browser's user agent. For headless Chromium, this includes string like `HeadlessChrome/61.0.3153.0`.
+- returns: <[Promise]<[string]>> Promise which resolves to the browser's original user agent. For headless Chromium, this includes a string like `HeadlessChrome/61.0.3153.0`.
 
-> **NOTE** browser.userAgent() does not always match the actual user agent if it's overridden by `page.setUserAgent(userAgent)` or `page.emulate(options)`.
+> **NOTE** browser.userAgent() does not always match the actual user agent if it's overridden by page.setUserAgent(userAgent) or page.emulate(options).
 
 #### browser.version()
 - returns: <[Promise]<[string]>> For headless Chromium, this is similar to `HeadlessChrome/61.0.3153.0`. For non-headless, this is similar to `Chrome/61.0.3153.0`.
--- a/docs/api.md
+++ b/docs/api.md
@@ -366,7 +366,7 @@ Disconnects Puppeteer from the browser, but leaves the Chromium process running.
 #### browser.userAgent()
 - returns: <[Promise]<[string]>> Promise which resolves to the browser's original user agent. For headless Chromium, this includes a string like `HeadlessChrome/61.0.3153.0`.
 
-> **NOTE** browser.userAgent() does not always match the actual user agent if it's overridden by page.setUserAgent(userAgent) or page.emulate(options).
+> **NOTE** Pages can override browser user agent with [page.setUserAgent](#pagesetuseragentuseragent)
 
 #### browser.version()
 - returns: <[Promise]<[string]>> For headless Chromium, this is similar to `HeadlessChrome/61.0.3153.0`. For non-headless, this is similar to `Chrome/61.0.3153.0`.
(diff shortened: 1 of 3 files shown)


yujiosaka:
@aslushnikov 

Thanks for your comment! I added `browser.userAgent()` method as you suggested.

docs/api.md
```diff
@@ -362,6 +363,11 @@ Disconnects Puppeteer from the browser, but leaves the Chromium process running.
 #### browser.targets()
 - returns: <[Array]<[Target]>> An array of all active targets.
 
+#### browser.userAgent()
+- returns: <[Promise]<[string]>> Promise which resolves to the browser's original user agent. For headless Chromium, this includes a string like `HeadlessChrome/61.0.3153.0`.
+
+> **NOTE** browser.userAgent() does not always match the actual user agent if it's overridden by page.setUserAgent(userAgent) or page.emulate(options).
```
yujiosaka: Maybe we can create `page.userAgent()` to return the actual user agent. But in that case the user must have enough control on which user agent to send, so I don't find necessity to do that.

lib/Browser.js
```diff
@@ -131,8 +132,18 @@ class Browser extends EventEmitter {
    * @return {!Promise<string>}
    */
   async version() {
-    const version = await this._connection.send('Browser.getVersion');
-    return version.product;
+    if (!this._version)
```
aslushnikov: nit: since this is not a hot path or anything I'd avoid caching to keep the program state small.

test/test.js
```diff
@@ -272,6 +272,14 @@ describe('Page', function() {
     });
   });
 
+  describe('Browser.userAgent', function() {
+    it('should return whether we are in headless', async({browser}) => {
+      const userAgent = await browser.userAgent();
+      expect(userAgent.length).toBeGreaterThan(0);
+      expect(userAgent.includes('Headless')).toBe(headless);
```
aslushnikov: nit: I'd check that it contains 'webkit' as well (so that the test differs from `browser.getVersion()`)

docs/api.md
aslushnikov: Your comment makes me think that you don't need the browser's user agent, but instead you're interested in the page's user agent. Is this the case?

yujiosaka: Not exactly.

In my case, I need to get browser's user agent even before opening a new page.

I'm writing a crawler based on Puppeteer.
The basic logic is here:

1. Access the target sites' robots.txt
2. Check robots.txt to see if the user agent is allowed
3. If the user agent is not allowed, stop here and don't go to the next step.
4. Open a new page by `browser.newPage()`
5. Start scraping the page

The code is [here](https://github.com/yujiosaka/headless-chrome-crawler/blob/master/lib/hccrawler.js#L302).

test/test.js
yujiosaka: Done
https://github.com/GoogleChrome/puppeteer/pull/1716/commits/28f2626ae79f59060bee1852727ece3ebc4aad12

lib/Browser.js
yujiosaka: https://github.com/GoogleChrome/puppeteer/pull/1716/commits/a48d759035777c22a509c4b8a61c29b20d6b6874

test/test.js
```diff
@@ -272,6 +272,14 @@ describe('Page', function() {
     });
   });
 
+  describe('Browser.userAgent', function() {
+    it('should includes WebKit', async({browser}) => {
+      const userAgent = await browser.userAgent();
+      expect(userAgent.length).toBeGreaterThan(0);
+      expect(userAgent.includes('WebKit')).toBe(headless);
```
aslushnikov: `WebKit` is always a part of UA, so this should always be true

```js
expect(userAgent).toContain('WebKit');
```

docs/api.md
aslushnikov: Thanks, I understand now.

nit: let's change the "NOTE" text to the following:

> **NOTE** Pages can override browser user agent with [`page.setUserAgent`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagesetuseragentuseragent)

aslushnikov (approved):

test/test.js
yujiosaka: Done
https://github.com/GoogleChrome/puppeteer/pull/1716/commits/5565ed6c76b1926fedd031df06c9e1f824b3f5d7

Thank you for catching my mistake :)

docs/api.md
yujiosaka: Done
https://github.com/GoogleChrome/puppeteer/pull/1716/commits/5565ed6c76b1926fedd031df06c9e1f824b3f5d7

aslushnikov:
Thank you @yujiosaka for working on this. I've changed a few nits in your PR, I hope you don't mind. 

I'll merge this as soon as CI passes.

yujiosaka:
@aslushnikov

Thank you so much for taking time reviewing my PR. I really appreciate it :)

PR #1721, opened 2018-01-04
test: verify serialization of "undefined" fields

### LINKED ISSUE #1510: page.evaluate incorrectly serializes undefined as null
### Steps to reproduce

* Puppeteer version: 0.13.0
* Platform / OS version: macOS / 10.12.6
* URLs (if applicable):

`console.log( await page.evaluate('var x = {a: undefined}; x') );`

`{}`

**What happens instead?**
`{a:null}`

Current workaround is to `JSON.stringify` result before returning it from the browser context.

Returning `null` instead of `undefined` was breaking [html-sketchapp](https://github.com/brainly/html-sketchapp/issues/29).

aslushnikov:
Fixes #1510.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -389,6 +389,9 @@ describe('Page', function() {
       const result = await page.evaluate((a, b) => Object.is(a, undefined) && Object.is(b, 'foo'), undefined, 'foo');
       expect(result).toBe(true);
     });
+    it('should properly serialize null fields', async ({page}) => {
+      expect(await page.evaluate(() => ({a: undefined}))).toEqual({});
+    });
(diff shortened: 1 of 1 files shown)


JoelEinbinder (commented):
How does toEqual work? Isn’t it json stringify?

aslushnikov:
It is a JSON.stringify under the hood, but it will yield different results for the `{a: null}` and `{a: undefined}`, so I think we're good here.

JoelEinbinder:
`expect("a" in result).toBe(false)` seems more robust but its not a big deal either way.

JoelEinbinder (approved):

PR #1722, opened 2018-01-04
fix: don't show max listeners warning when running tests in parallel

JoelEinbinder:


test/test.js
```diff
@@ -60,8 +60,9 @@ if (process.env.PPTR_PARALLEL_TESTS)
 const parallelArgIndex = process.argv.indexOf('-j');
 if (parallelArgIndex !== -1)
   parallel = parseInt(process.argv[parallelArgIndex + 1], 10);
+require('events').defaultMaxListeners *= parallel;
```
aslushnikov, line 63: this will kill certain tests check for the warning, but i guess that's not a big deal for the parallel mode

aslushnikov (approved):

PR #1723, opened 2018-01-04
fix: scrub port from coverage golden files

JoelEinbinder:
These tests were flaky in parallel.


aslushnikov (approved):
Thanks, I've just seen this as well

PR #1734, opened 2018-01-05
fix: setRequestInterception and setExtraHTTPHeaders not working together (#1729)

### LINKED ISSUE #1729: page.goto returns a promise to null when intercepting request and modifying referer together
### Steps to reproduce

* Puppeteer version: 0.13.0 (but also occurs at puppeteer@next )
* Platform / OS version: Mac OS
* URLs (if applicable): *
* Node.js version: 8.9.1

1. It works
```js
const URL = 'http://example.com/';
const REFERER = 'http://example.net/';
(async() => {
    // await page.setRequestInterception(true);
    // page.on('request', interceptedRequest => {
    //  interceptedRequest.continue();
    // });
    await page.setExtraHTTPHeaders({ 'referer': REFERER });
    const response = await page.goto(URL);
    try {
      const text = await response.text();
      console.log(text); // goes here
    } catch(error) {
      console.error(error);
    }
    await page.close();
    await browser.close();
})();
```

2. It works
```js
const URL = 'http://example.com/';
const REFERER = 'http://example.net/';
(async() => {
    page.on('request', interceptedRequest => {
     interceptedRequest.continue();
    });
    // await page.setExtraHTTPHeaders({ 'referer': REFERER });
    const response = await page.goto(URL);
    try {
      const text = await response.text();
      console.log(text); // goes here
    } catch(error) {
      console.error(error);
    }
    await page.close();
    await browser.close();
})();
```

3. but it catches the error
```js
const URL = 'http://example.com/';
const REFERER = 'http://example.net/';
(async() => {
    page.on('request', interceptedRequest => {
     interceptedRequest.continue();
    });
    await page.setExtraHTTPHeaders({ 'referer': REFERER });
    const response = await page.goto(URL);
    try {
      const text = await response.text();
      console.log(text);
    } catch(error) {
      console.error(error); // goes here
    }
    await page.close();
    await browser.close();
})();
```

The most strange thing is that if I modify other headers than referer, it works.

4. it works
```js
await page.setExtraHTTPHeaders({ 'otherthanreferrer': REFERER });
```

Either way, the screenshots are taken ok, so it's not like failing to navigating to the next page.

`page.goto()` always returns a promise to response

**What happens instead?**

`page.goto()` returns a promise to `null`

```
TypeError: Cannot read property 'text' of null
```

yujiosaka:
This is just a fix to the problem that `page.goto` resolves to `null`.
Setting referer does not work properly, but that should be discussed [here](https://github.com/GoogleChrome/puppeteer/issues/469).

Fixes: https://github.com/GoogleChrome/puppeteer/issues/1729

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -569,16 +569,14 @@ function generateRequestHash(request) {
   };
 
   if (!normalizedURL.startsWith('data:')) {
-    const headers = {};
-    for (const key of Object.keys(request.headers))
-      // Need to consider uncapitalized headers. See https://github.com/GoogleChrome/puppeteer/issues/1729
-      headers[key.toLowerCase()] = request.headers[key];
-    const headerKeys = Object.keys(headers);
-    headerKeys.sort();
-    for (const key of headerKeys) {
-      if (key === 'accept' || key === 'referer' || key === 'x-devtools-emulate-network-conditions-client-id')
+    const headers = Object.keys(request.headers);
+    headers.sort();
+    for (let header of headers) {
+      const headerValue = request.headers[header];
+      header = header.toLowerCase();
+      if (header === 'accept' || header === 'referer' || header === 'x-devtools-emulate-network-conditions-client-id')
         continue;
-      hash.headers[key] = headers[key];
+      hash.headers[header] = headerValue;
     }
   }
   return JSON.stringify(hash);
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -569,12 +569,16 @@ function generateRequestHash(request) {
   };
 
   if (!normalizedURL.startsWith('data:')) {
-    const headers = Object.keys(request.headers);
-    headers.sort();
-    for (const header of headers) {
-      if (header === 'Accept' || header === 'Referer' || header === 'X-DevTools-Emulate-Network-Conditions-Client-Id')
+    const headers = {};
+    for (const key of Object.keys(request.headers))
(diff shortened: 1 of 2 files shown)


lib/NetworkManager.js
```diff
@@ -569,12 +569,16 @@ function generateRequestHash(request) {
   };
 
   if (!normalizedURL.startsWith('data:')) {
-    const headers = Object.keys(request.headers);
-    headers.sort();
-    for (const header of headers) {
```
aslushnikov: can we please minimize the change? I believe you can add a `header.toLowerCase()` as the first line in the loop:
```js
for (let header of headers) {
  header = header.toLowerCase();
  if (key === 'accept' || key === 'referer' ...)
}
```

test/test.js
```diff
@@ -1321,6 +1321,16 @@ describe('Page', function() {
       const response = await page.goto(server.EMPTY_PAGE);
       expect(response.ok()).toBe(true);
     });
+    it('should works with customizing referer headers', async({page, server}) => {
+      await page.setExtraHTTPHeaders({ 'referer': server.EMPTY_PAGE });
```
aslushnikov, line 1325: this is a good test, thank you.

aslushnikov (commented):
Thank you @yujiosaka for taking time to work on this.

aslushnikov (approved):
@yujiosaka I've slightly changed your code; I hope you don't mind. Thank you for fixing this!

yujiosaka:
@aslushnikov thanks!

PR #1738, opened 2018-01-08
docs(api.md): add missing parentheses after mainFrame

mkamakura:
Add missing parentheses after mainFrame.


aslushnikov (approved):
Thanks

PR #1753, opened 2018-01-09
chore(Page): Add missing return comments in addScriptTag and addStyleTag

kblok:
Added some missing docs based on `mainFrame`.

It's also my first tiny contribution so I also want to see how your workflow works.

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -269,6 +269,7 @@ class Page extends EventEmitter {
 
   /**
    * @param {Object} options
+   * @return {!Promise<!ElementHandle>}
    */
   async addScriptTag(options) {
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):
Thanks!

lib/Page.js
```diff
@@ -269,13 +269,15 @@ class Page extends EventEmitter {
 
   /**
    * @param {Object} options
+   * @return {!Promise<!ElementHandle>}
```
aslushnikov: This should probably be `Puppeteer.ElementHandle`, otherwise tsc is not happy.

You can try running tsc locally with
```
npm run tsc
```

PR #1759, opened 2018-01-10
fix: recommend yarn install than npm install in the error message

yujiosaka:
Small fix.

I think the error message should recommend "yarn install" than "npm install".


aslushnikov (approved):
Sure, sg.

PR #1770, opened 2018-01-11
feat: expose raw devtools protocol connection

### LINKED ISSUE #31: Expose raw protocol in the API

aslushnikov:
This patch introduces `target.createCDPSession` method that
allows directly communicating with the target over the
Chrome DevTools Protocol.

Fixes #31.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2225,6 +2225,7 @@ to send messages.
 - `params` <[Object]> Optional method parameters
 - returns: <[Promise]<Object>>
 
+
 ### class: Coverage
 
 Coverage gathers information about parts of JavaScript and CSS that were used by the page.
--- a/docs/api.md
+++ b/docs/api.md
@@ -2223,7 +2223,7 @@ to send messages.
 #### cdpSession.send(method[, params])
 - `method` <[string]> protocol method name
 - `params` <[Object]> Optional method parameters
-- returns: <[Promise]<Object>>
+- returns: <[Promise]<[Object]>>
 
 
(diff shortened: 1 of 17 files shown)


docs/api.md
```diff
@@ -2166,6 +2174,11 @@ Contains the URL of the response.
 
 ### class: Target
 
+#### target.createCDPSession()
+- returns: <[CDPSession]>
```
JoelEinbinder: Promise to session? And the link doesn't work.

docs/api.md
```diff
@@ -2166,6 +2174,11 @@ Contains the URL of the response.
...
+
+Creates a *C*hrome *D*evtools *P*rotocol session attached to the target.
```
JoelEinbinder: Let's not have the letters in italics.

docs/api.md
```diff
@@ -2192,6 +2192,7 @@ Identifies what kind of target this is. Can be `"page"`, `"service_worker"`, or
 #### target.url()
 - returns: <[string]>
 
+
```
JoelEinbinder, line 2195: This newline is in the wrong place.

aslushnikov: Added another one.

JoelEinbinder (approved):

PR #1780, opened 2018-01-11
docs(troubleshooting): add not on using ldd to find missing deps

ebidel:
R: @aslushnikov


aslushnikov (approved):
This is a really good tip!

PR #1787, opened 2018-01-12
feat: elHandle:screenshot captures full element

### LINKED ISSUE #1779: Screenshot element larger than viewport
### Steps to reproduce

* Puppeteer version: 0.13.0
* Platform / OS version: OSX
* Node.js version: 8.9.0

```js
// compare with .to-screenshot size

await page.setContent(`
  something above
  <style>div.spacer {
    border: 2px solid blue;
    background: red;
    height: 600px;
  }
  div.to-screenshot {
    border: 2px solid blue;
    background: green;
    width: 50px;
    height: 1000px;
    margin-left: 1000px;
  }
  </style>
  <div class="spacer"></div>
  <div class="to-screenshot"></div>
  <div class="spacer"></div>
`);
const elementHandle = await page.$('div.to-screenshot');
const screenshot = await elementHandle.screenshot();
```

---

I'm using `elementHandle.screenshot()` which is working, but if the element is larger than the viewport, it won't capture all of it. This make sense if you look at the source and see that it's using `page.screenshot` with a `clip`, and you know that clip doesn't work with `fullPage`, but it's not intuitive, and doesn't seem like a good default behavior.

Would it be possible to get `elementHandle.screenshot` to capture the whole element? It already tries to ignore scroll position, but viewport size limits it.

I tried making it use `Emulation.setDeviceMetricsOverride`, but then clip stopped working. Guessing this is why `clip`+`fullPage` errors.

On the project I need this for, I ended up doing:

```js
const bounds = await el.boundingBox();
const initial = Object.assign({}, page.viewport());

await page.setViewport({
  width: Math.max(Math.ceil(bounds.width), initial.width),
  height: Math.max(Math.ceil(bounds.height), initial.height),
});

const imageRes = await el.screenshot({ type: 'jpeg', quality: 90 });
await page.setViewport(initial);
return imageRes;
```

Of course, this would impact media queries and 'resize' events, so not high enough quality to be included in puppeteer's source, I think.

I don't know what a good solution would be, but I use element screenshots a lot so I'm happy to implement any solutions you come up with. Thanks for the great tool!

brigand:
Modifies `ElementHande::screenshot` to capture the full element, even if
it's larger than the current viewport, or it falls off screen. Uses
`window.scrollTo` instead of `scrollIntoViewIfNeeded` because it must be
as much inside the viewport as possible. Resets the viewport size and scroll position after
it's finished.

Getting a random test failure (ran it 100 times, failed 10 times). It's calling a function I created, but I don't know what's causing the issue.

```
Failures:
1) Page Request.respond should allow mocking binary responses
  Message:
    Node is not visible @ test.js:1564:5
  Stack:
    Error: Node is not visible
        at ElementHandle._assertBoundingBox (/Users/fb/github/brigand/puppeteer/lib/ElementHandle.js:160:13)
        at <anonymous>
```

Other than that, it seems to be working properly.

I think the scroll logic is the main place for improvement. It doesn't deal well with non-window scroll parents at the moment. Could add `scrollIntoViewIfNeeded` back in, but I'd want to revert the changes after taking the screenshot. Not sure if that's introducing too much extra code.

---

Fixes #1779  
cc @JoelEinbinder


brigand:
Signed CLA

lib/ElementHandle.js
```diff
@@ -58,9 +58,7 @@ class ElementHandle extends JSHandle {
    */
   async _visibleCenter() {
     await this._scrollIntoViewIfNeeded();
-    const box = await this.boundingBox();
-    if (!box)
-      throw new Error('Node is not visible');
+    const box = await this._assertBoundingBox();
```
brigand, line 61: Extracted this out as it's used in several places.

lib/ElementHandle.js
```diff
@@ -121,9 +119,25 @@ class ElementHandle extends JSHandle {
...
-    }).catch(error => void debugError(error));
+    let result = null;
+    try {
+      result = await this._client.send('DOM.getBoxModel', {
+        objectId: this._remoteObject.objectId
+      });
+    } catch (protoError) {
+      debugError(protoError);
+
+      const noElementError = await this.executionContext().evaluate(element => {
```
brigand: Check is in the catch to avoid slowing down the success path.

lib/ElementHandle.js
```diff
@@ -137,24 +151,72 @@ class ElementHandle extends JSHandle {
...
+    let needsViewportReset = false;
+
+    let boundingBox = await this._assertBoundingBox();
+
+    const viewport = Object.assign({}, this._page.viewport());
+
+    const newViewport = {
+      width: Math.max(viewport.width, Math.ceil(boundingBox.width)),
+      height: Math.max(viewport.height, Math.ceil(boundingBox.height)),
+    };
```
brigand: Never making it smaller in either dimension seems good; avoids a resize entirely in some common cases.

lib/ElementHandle.js
```diff
@@ -137,24 +151,72 @@ class ElementHandle extends JSHandle {
...
+
+    const initialScroll = await this.executionContext().evaluate(function() {
+      return { scrollX: window.scrollX, scrollY: window.scrollY };
+    });
```
brigand: So we can reset the scroll; fewer side effects from running the screenshot function.

test/test.js
```diff
@@ -1906,6 +1906,44 @@ describe('Page', function() {
...
+          border: 2px solid blue;
+          background: rgba(0, 0, 0, 0.5);
+          width: 600px;
+          height: 200.5px;
+          margin-left: 50px;
+          transform: scaleY(1.2);
+        }
+        </style>
+        <div class="spacer"></div>
+        <div class="to-screenshot">intentional red in left corners</div>
```
JoelEinbinder: Text will render differently on different platforms, which causes the golden file test to fail on the bots. If you remove the text it should be fine.

lib/ElementHandle.js
```diff
@@ -137,24 +151,72 @@ class ElementHandle extends JSHandle {
...
+
+      needsViewportReset = true;
+    }
+
+    await this.executionContext().evaluate(function(el) {
+      const box = el.getBoundingClientRect();
+
+      window.scrollTo(
+          box.left + window.scrollX,
+          box.top + window.scrollY
```
brigand: Was going to put screenshots of `scrollIntoViewIfNeeded` compared to this, but looking at just the current one with a pink shadow inside body, we're actually screenshotting outside body with this code and test. Seems better given the context.


screenshot-element-larger-than-viewport

lib/ElementHandle.js
```diff
@@ -137,24 +151,72 @@ class ElementHandle extends JSHandle {
...
-    return await this._page.screenshot(Object.assign({}, {
+
+    const imageData = await this._page.screenshot(Object.assign({}, {
       clip
     }, options));
+
+    if (needsViewportReset)
+      await this._page._emulationManager.emulateViewport(this._client, viewport);
+
+    await this.executionContext().evaluate(function(x, y) {
```
JoelEinbinder: I wouldn't worry about putting the scroll position back. I'd favor minimal interactions with the page over cleaning up after ourselves.

brigand: Fixed

lib/ElementHandle.js
```diff
@@ -137,24 +151,72 @@ class ElementHandle extends JSHandle {
...
+    if (newViewport.width !== viewport.width || newViewport.height !== viewport.height) {
+      await this._page._emulationManager.emulateViewport(this._client, newViewport);
+
+      needsViewportReset = true;
+    }
+
+    await this.executionContext().evaluate(function(el) {
+      const box = el.getBoundingClientRect();
+
+      window.scrollTo(
```
JoelEinbinder: I'm not understanding how scrollIntoViewIfNeeded fails here. I can see how you might need to call it again after resizing the viewport.

brigand: This is with `await this._scrollIntoViewIfNeeded();`. Part of the element and its right border get cut off.

screenshot-element-larger-than-viewport

JoelEinbinder: It looks like the viewport isn’t wide enough though? The left side of of the image is flush against the left of the viewport.

brigand: There's a 50px left margin. The only difference between the full and right-cropped version is the scroll logic. The viewport size is exactly the same. It's not actually right cropped, more like clipped. There's some white space on the right side of the image. I think that part of the element is outside `<body>`.

Could probably get it working with `scrollIntoViewIfNeeded` if we set the viewport width minimum to be `bounds.left + bounds.width`. That has the cost of increasing our chance of resizing the browser, and gives an extra thing that can be hit by media queries and percentages.

brigand:
Anything I should be doing to move this forward?

lib/ElementHandle.js
JoelEinbinder: Sorry for the delay. I understand now!

You can use code like this instead of `window.scrollTo`:
```js
await this.executionContext().evaluate(el => {
  el.scrollIntoView({
    behavior: 'instant',
    block: 'center',
    inline: 'nearest'
  });
}, this);
```

`window.scrollTo` doesn't work because the element that needs to scroll might not be the window.

lib/ElementHandle.js
```diff
@@ -137,24 +151,64 @@ class ElementHandle extends JSHandle {
...
+          box.left + window.scrollX,
+          box.top + window.scrollY
+      );
+    }, this);
+
+    boundingBox = await this._assertBoundingBox();
```
JoelEinbinder: reseting the bounding box and the above scrolling code can go inside the same `if` where we set the new viewport. That way we don't do send extra messages into chrome for elements that fit on screen.

test/test.js
```diff
@@ -1906,6 +1906,44 @@ describe('Page', function() {
...
+          width: 52px;
+        }
+        div.to-screenshot {
+          border: 2px solid blue;
+          background: rgba(0, 0, 0, 0.5);
+          width: 600px;
+          height: 200.5px;
+          margin-left: 50px;
+          transform: scaleY(1.2);
+        }
```
JoelEinbinder, line 2109: This will show scrollbars in headful mode. To hide them, add:
```css
::-webkit-scrollbar {
  display: none;
}
```

lib/ElementHandle.js
```diff
@@ -137,24 +151,64 @@ class ElementHandle extends JSHandle {
...
    */
   async screenshot(options = {}) {
-    await this._scrollIntoViewIfNeeded();
+    let needsViewportReset = false;
+
+    let boundingBox = await this._assertBoundingBox();
+
+    const viewport = Object.assign({}, this._page.viewport());
+
+    const newViewport = {
```
JoelEinbinder: Its a nitpicky thing, but can you build this rect inside the if statement below it. Like
```js
if (boundingBox.width > viewport.width || boundingBox.height > viewport.height) {
   const newViewport = ...
```
It makes it clearer at a glance that we don't always set a new viewport

lib/ElementHandle.js
```diff
@@ -137,24 +151,64 @@ class ElementHandle extends JSHandle {
...
    * @param {!Object=} options
    * @returns {!Promise<Object>}
    */
   async screenshot(options = {}) {
-    await this._scrollIntoViewIfNeeded();
+    let needsViewportReset = false;
+
+    let boundingBox = await this._assertBoundingBox();
+
+    const viewport = Object.assign({}, this._page.viewport());
```
JoelEinbinder: I don't think you need `Object.assign` here.

JoelEinbinder:
Sorry for the delay. I had a few comments but overall it looks great!

lib/ElementHandle.js
brigand: I tried it out, but the scroll logic often needs to occur even if we don't resize the viewport.

With the currently pushed code:

screenshot-element-off-screen

With scroll conditional on resizing viewport (where no viewport resize occurs)

screenshot-element-off-screen

In this example, the viewport is 500x500. The top of the element we're screenshotting is at y: 400. Without the scroll, it's only screenshotting at max the first 500x500 of the page.


Full test case for completeness
```js
    it('should capture full element when off screen', async({page, server}) => {
      // compare with .to-screenshot size

      await page.setContent(`
        something above
        <style>div.spacer {
          border: 2px solid red;
          background: red;
          height: 400px;
          width: 50px;
        }
        div.to-screenshot {
          border: 2px solid blue;
          background: rgba(0, 0, 0, 0.5);
          width: 400px;
          height: 200.5px;
          margin-left: 50px;
          transform: scaleY(1.2);
        }
        ::-webkit-scrollbar {
          display: none;
        }
        </style>
        <div class="spacer"></div>
        <div class="to-screenshot"></div>
        <div class="spacer"></div>
      `);

      await page.evaluate(function() {
        window.scrollTo(0, 0);
      });

      const elementHandle = await page.$('div.to-screenshot');
      const screenshot = await elementHandle.screenshot();
      expect(screenshot).toBeGolden('screenshot-element-off-screen.png');
    });
```

test/test.js
brigand: Fixed

lib/ElementHandle.js
brigand: I put it there to be safe around page._viewport being updated. Removed.

lib/ElementHandle.js
JoelEinbinder: That makes sense. I didn't realize you weren't scrolling at the top. Trading an extra boundingClientRect for a scroll is a good trade. Leave things how they are.

lib/ElementHandle.js
brigand: Tried that. Test failed with this output:

screenshot-element-larger-than-viewport

You're not going to like this, but it seems to be working with this code, tested against two scroll parent divs plus window.

```js
    await this.executionContext().evaluate(function(el) {
      const parents = [];
      for (let parent = el.parentNode; parent; parent = parent.parentNode) {
        if (parent.scrollHeight !== parent.offsetHeight)
          parents.unshift(parent);
      }

      for (const parent of parents) {
        parent.scrollTo(0, 0);

        const parentBounds = parent.getBoundingClientRect();
        const elBounds = el.getBoundingClientRect();

        const x = elBounds.left - parentBounds.left;
        const y = elBounds.top - parentBounds.top;

        parent.scrollTo(x, y);
      }

      el.scrollIntoView({
        behavior: 'instant',
        block: 'center',
        inline: 'nearest',
      });
    }, this);
```

brigand: This could be optimized, but I'll do that if you don't give a hard **no** 😄

brigand:
Ready for another round of review.

lib/ElementHandle.js
JoelEinbinder: I'd like to avoid it. I'll give it a shot later. `scrollIntoView` passed the test for me when I tried it. What OS are you on?

brigand: OSX. The test is `should screenshot element with scroll container`.

brigand:
Had to remove the `scrollIntoView` again because it was causing a test failure on 'should capture full element when larger than viewport'. 

```diff
-      el.scrollIntoView({
-        behavior: 'instant',
-        block: 'center',
-        inline: 'nearest',
-      });
+      const bounds2 = el.getBoundingClientRect();
+
+      window.scrollTo(
+          bounds2.left + window.scrollX,
+          bounds2.top + window.scrollY,
+      );
```

Had output (right edge missing):

screenshot-element-larger-than-viewport

aslushnikov:
@JoelEinbinder is anything blocking this? Let me know if there's any help needed.

brigand:
I'd still like to get this in, so just let me know what I need to do.

JoelEinbinder:
The tests all pass for me with `scrollIntoView`. See https://github.com/JoelEinbinder/puppeteer/tree/full_element_screenshot . I'd really like to avoid the `scrollTo` function. It won't work for content inside frames, and I would bet there are other edge cases we haven't thought of. Do you have a scenario/test where `scrollIntoView` doesn't work but the recursive `scrollTo` does?

brigand:
Oh, I was using `'inline': 'nearest'`, and before that it was `scrollIntoViewIfNeeded`. Your new code seems to be working perfectly, thanks!

lib/ElementHandle.js
```diff
@@ -137,24 +151,58 @@ class ElementHandle extends JSHandle {
...
+    let boundingBox = await this._assertBoundingBox();
+
+    const viewport = this._page.viewport();
+
+    if (boundingBox.width > viewport.width || boundingBox.height > viewport.height) {
+      const newViewport = {
+        width: Math.max(viewport.width, Math.ceil(boundingBox.width)),
+        height: Math.max(viewport.height, Math.ceil(boundingBox.height)),
+      };
+      await this._page._emulationManager.emulateViewport(this._client, newViewport);
```
JoelEinbinder: We try not to use underscore methods from other files. Can you switch this to page.setViewport?

brigand: Yeah, I wasn't sure about this one. For reference:

```js
  async setViewport(viewport) {
    const needsReload = await this._emulationManager.emulateViewport(this._client, viewport);
    this._viewport = viewport;
    if (needsReload)
      await this.reload();
  }
```

As I understand it, we won't ever cause a reload here, and the `this._viewport` won't actually change unless another operation is running concurrently with `elHandle.screenshot`.

In theory it should be fine, but I'm getting a consistently failing test after making the change.

```
Failures:
1) Page Request.respond should allow mocking binary responses
  Message:
    Node is not visible @ test.js:1564:5
  Stack:
    Error: Node is not visible
        at ElementHandle._assertBoundingBox (/Users/fb/github/brigand/puppeteer/lib/ElementHandle.js:160:13)
        at <anonymous>
```

brigand: I'm guessing the `tsc` error in CI is because I'm accessing `this._page._emulationManager`.

```
> puppeteer@1.0.0-post tsc /home/travis/build/GoogleChrome/puppeteer
> tsc -p .
lib/ElementHandle.js(181,13): error TS2554: Expected 1 arguments, but got 2.
lib/ElementHandle.js(203,13): error TS2554: Expected 1 arguments, but got 2.
```

Pointing to

```js
await this._page._emulationManager.emulateViewport(this._client, newViewport);
```

JoelEinbinder: The tsc error is because you need to rebase.

You should use setViewport, but the reload brings up a good point. You need to copy the other parameters besides width/height into the new viewport. I recommend Object.assign.

brigand:
Uh I seem to have messed up the rebase.

brigand:
I think I fixed the rebase. Got a random test failure (ran it once: error, ran two more times: no error). Seems unrelated to this PR. I'm on OSX, btw.

```
Failures:
1) Page Frame.waitForSelector Page.waitForSelector is shortcut for main frame (test.js:801:5)
  Message:
    expect.toBe failed: false == true
  Stack:
    AssertionError [ERR_ASSERTION]: expect.toBe failed: false == true
        at applyMatcher (/Users/fb/github/brigand/puppeteer/utils/testrunner/Matchers.js:47:15)
     ⇨  at it (/Users/fb/github/brigand/puppeteer/test/test.js:810:21)
        at <anonymous>
```

lib/ElementHandle.js
```diff
@@ -121,9 +119,25 @@ class ElementHandle extends JSHandle {
...
+
+      const noElementError = await this.executionContext().evaluate(element => {
+        if (!element.isConnected)
+          return 'Node is detached from document';
+        if (element.nodeType !== Node.ELEMENT_NODE)
+          return 'Node is not of type HTMLElement';
+        return null;
+      }, this);
+
+      if (noElementError)
```
JoelEinbinder: I think you wanted this code in assertBoundingBox? We intentionally have bounding box throw no errors, and return null for invisible nodes.

JoelEinbinder:
Looks good!

lib/ElementHandle.js
brigand: Fixed

lib/ElementHandle.js
```diff
@@ -137,24 +140,67 @@ class ElementHandle extends JSHandle {
...
 
+  /**
+   * @return {!Promise<?{x: number, y: number, width: number, height: number}>}
+   */
+  async _assertBoundingBox() {
+    const boundingBox = await this.boundingBox();
+    if (boundingBox)
+      return boundingBox;
+
+    const noElementError = await this.executionContext().evaluate(element => {
```
aslushnikov: I'd save a roundtrip and throw a generic error: `Node is either not visible or not an HTMLElement`.

lib/ElementHandle.js
```diff
@@ -137,24 +140,67 @@ class ElementHandle extends JSHandle {
...
    */
   async screenshot(options = {}) {
-    await this._scrollIntoViewIfNeeded();
+    let needsViewportReset = false;
+
+    let boundingBox = await this._assertBoundingBox();
+
+    const viewport = this._page.viewport();
+
+    if (boundingBox.width > viewport.width || boundingBox.height > viewport.height) {
```
aslushnikov, line 161: can we move this functionality into `page.screenshot` so that it works with clips larger then viewport? Seems reasonable to me.

JoelEinbinder: that might be a bit tricky, because resizing the viewport will likely move the bounding box to an unknown location.

lib/ElementHandle.js
brigand: Fixed. I agree that we're not losing much useful information in this case.

lib/ElementHandle.js
brigand: I think solving that in `page.screenshot` should be a separate PR, but resizing the viewport in `page.screenshot` seems like it would do bad things more often than `elementHandle.screenshot` does, since at the page level, many things are based on the viewport size. If possible, that should be solved in chromium/protocol.

I'd be good with scrolling in `page.screenshot` if the clip area requires it, but I'd like to get this PR merged first.

brigand:
Rebased, and ready for review. I think it's in a good state.

lib/ElementHandle.js
```diff
@@ -121,9 +119,14 @@ class ElementHandle extends JSHandle {
    * @return {!Promise<?{x: number, y: number, width: number, height: number}>}
    */
   async boundingBox() {
-    const result = await this._client.send('DOM.getBoxModel', {
-      objectId: this._remoteObject.objectId
-    }).catch(error => void debugError(error));
+    let result = null;
```
aslushnikov: I'm not sure I understand why this has changed. Is it necessary?

brigand: I think it ended up at this after a few rounds of changes, almost full circle. Will complete the circle.

brigand: Fixed

JoelEinbinder (approved):

brigand:
Checked the appveyor build. I was getting random failures on "Page Request.respond should allow mocking binary responses" also.

zhongjiewang:
hello, I want to  screenshot  an  elHandle  in  a frame , always failed , Can you give me some advice?

PR #1795, opened 2018-01-12
feat(Page): add monotonic timestamp to Request and Response events

joeyvandijk:

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -188,6 +188,7 @@
   * [request.resourceType()](#requestresourcetype)
   * [request.respond(response)](#requestrespondresponse)
   * [request.response()](#requestresponse)
+  * [request.timestamp()](#requesttimestamp)
   * [request.url()](#requesturl)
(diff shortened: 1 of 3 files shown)


joeyvandijk:
I signed it!

joeyvandijk:
See https://github.com/GoogleChrome/puppeteer/pull/1794 for having the issue that I am running Google for Work, but using contact@xxx for Github and mail@xxx as default user in Google for Work. I am the same person, but I cannot apply that fact anywhere, so I cannot add this code without above problems.

joeyvandijk:
I have changed my mail account and received an update.

rafaelugolini:
What's the status of this PR? This feature would be very handy 👍

joeyvandijk:
@rafaelugolini when I have time I want to make another PR so the cla/google step will succeed, while I've now fixed the e-mail adresses. I will close this and want to create this week another PR.

PR #1801, opened 2018-01-14
docs(api.md): fix missing await in examples

futpib:
`frame.executionContext()` returns a Promise, so running examples without `await` results in exceptions


futpib:
I signed it!

ebidel (approved):

PR #1806, opened 2018-01-15
feat(prefs): add ability to launch browser with desired preferences

skyiea:
The ability to start browser with desired [preferences](https://src.chromium.org/viewvc/chrome/trunk/src/chrome/common/pref_names.cc?view=markup).
[Related issue](https://github.com/GoogleChrome/puppeteer/issues/1693#issuecomment-356181997)

The easier way to kickoff was to allow to set `prefs` for the cases when there is no preconfigured `userDataDir` / `--user-data-dir` and to see the feedback from the maintainers.
What's done:
- [x] implementation
- [x] write tests
- [x] update documentation

What could be done as a separate feature:
- allow setting `prefs` in conjunction with `userDataDir` / `--user-data-dir`
- add ability to set ["local state" prefs](https://src.chromium.org/viewvc/chrome/trunk/src/chrome/common/pref_names.cc?view=markup#l1255)

@aslushnikov WDYT?

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -26,6 +26,8 @@ const {helper} = require('./helper');
 const ChromiumRevision = Downloader.defaultRevision();
 
 const mkdtempAsync = helper.promisify(fs.mkdtemp);
+const mkdirAsync = helper.promisify(fs.mkdir);
+const writeFileAsync = helper.promisify(fs.writeFile);
 const removeFolderAsync = helper.promisify(removeFolder);
 
 const CHROME_PROFILE_PATH = path.join(os.tmpdir(), 'puppeteer_dev_profile-');
@@ -75,6 +77,13 @@ class Launcher {
         temporaryUserDataDir = await mkdtempAsync(CHROME_PROFILE_PATH);
 
       chromeArguments.push(`--user-data-dir=${options.userDataDir || temporaryUserDataDir}`);
+    }
+    if (options.prefs) {
+      console.assert(!options.headless, `Custom preferences can't be used in headless mode`);
+      console.assert(temporaryUserDataDir, `Custom preferences can't be used with custom userDataDir`);
+      const profileDefaultDir = temporaryUserDataDir + '/Default';
+      await mkdirAsync(profileDefaultDir);
+      await writeFileAsync(profileDefaultDir + '/Preferences', JSON.stringify(options.prefs));
     }
     if (options.devtools === true) {
       chromeArguments.push('--auto-open-devtools-for-tabs');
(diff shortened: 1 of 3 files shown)


skyiea:
I signed it!

skyiea:
I signed it!

aslushnikov:
@skyiea This looks good! Looking forward to landing this.

skyiea:
@aslushnikov Added tests and updated docs.

lib/Launcher.js
```diff
@@ -76,6 +78,13 @@ class Launcher {
 
       chromeArguments.push(`--user-data-dir=${options.userDataDir || temporaryUserDataDir}`);
     }
+    if (options.prefs) {
+      console.assert(!options.headless, `Custom preferences can't be used in headless mode`);
```
aslushnikov, line 82: That's a bummer. I didn't know preferences are not supported by headless browser. We try hard not to land API that works only in headful browser.

@skyiea can you please share what particular preferences are you interested in? We might be able to plumb them for you through DevTools protocol so that it works in both headful and headless modes.

skyiea: I was surprised too that it doesn't work in headless browser, given that headless supports custom `userDataDir` and, say, `localStorage`/`cookies`. Might be some of you guys have some insight why `Preferences` file is being ignored in headless?
Personally for me it makes no difference, because I'm using puppeteer to test Chrome extension, which means that I don't have a chance to use headless mode anyway (BTW, any insight why so?). Thus I guess many of the users are already using puppeteer in headful mode.

That's not only about my particular case, I guess. As I've said, there is a lot (https://src.chromium.org/viewvc/chrome/trunk/src/chrome/common/pref_names.cc?view=markup) of additional settings that are available via preferences. For me it allows to set `prefs.download.default_directory` (`setDownloadBehavior ` doesn't work for me, I guess because I'm downloading via `chrome.downloads` API), and in the future I'd like to be able to trigger some experiments (via `prefs.experiments_enabled`).

skyiea:
@aslushnikov Any updates on this?

aslushnikov:
@skyiea I apologize for the long delay.

> @aslushnikov Any updates on this?

It turns out preferences are part of `chrome/` layer in Chromium, not `content/`; I wasn't aware of this.

Landing this PR will introduce more inconsistency in headless vs headful modes, which I very much want to avoid. I think you can implement this functionality as a utility method for your needs, or even publish as NPM module. Will this work for you? 

I'm sorry for misleading you on this feature.

aslushnikov:
Closing this as per previous comment.

berstend:
In case anyone else who needs this ends up here, I've published a small wrapper to enable custom preferences in puppeteer: https://github.com/berstend/puppeteer-extra

Heavily inspired by @skyiea's PR, thanks! 👍

PR #1817, opened 2018-01-17
fix(NetworkManager): add initiator property to Request object

### LINKED ISSUE #1395: Please add the initiator information
Requests, that are fired after the basic request is done, are initiated by some source - f.e. a script or something. Since chrome does provide this information in the network tab, I wonder if you can give it via puppeteer?

Example:
image

joshgoodwin:
this change allows you to check the initiator of a request.

Fixes #1395 

note: does not add property in time for it to be available within interceptedRequest event.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -183,6 +183,7 @@
   * [request.failure()](#requestfailure)
   * [request.frame()](#requestframe)
   * [request.headers()](#requestheaders)
+  * [request.initiator()](#requestinitiator)
   * [request.method()](#requestmethod)
   * [request.postData()](#requestpostdata)
   * [request.resourceType()](#requestresourcetype)
@@ -2098,6 +2099,9 @@ page.on('requestfailed', request => {
 
 #### request.headers()
 - returns: <[Object]> An object with HTTP headers associated with the request. All header names are lower-case.
+
+#### request.initiator()
+- returns: <?[Object]> An object with request initiator data.
 
 #### request.method()
 - returns: <[string]> Request's method (GET, POST, etc.)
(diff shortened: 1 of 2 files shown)


joshgoodwin:
test comment to see if googlebot likes my email now.

joshgoodwin:
try again...

aslushnikov:
Hey @joshgoodwin, Thank you for the PR.

I'd like the request interception to have initiator before exposing it in the pptr API. Do you want to contribute the change to chromium first? There's a related chromium bug in the issue discussion.

joshgoodwin:
@aslushnikov Just want to make sure I'm clear ... if I find a hack that makes initiator available for an intercepted request in the request event (instead of just afterward as is the case with this patch), would you all not accept it? I'm not a huge fan of having to keep patching NetworkManager.js as part of my release process, so I'm willing to put some more work in if it'd actually have a chance of getting merged.

sebastianneubert (commented):
simple MR
- added initiator information incl. docs

Michaelknigth:
Check it out
<https://drive.google.com/file/d/1zup0tSiSnupOVGpvqwKTCt2kBQFPm1Jf/view?usp=drivesdk>

aslushnikov:
> Just want to make sure I'm clear ... if I find a hack that makes initiator available for an intercepted request in the request event (instead of just afterward as is the case with this patch), would you all not accept it? I'm not a huge fan of having to keep patching NetworkManager.js as part of my release process, so I'm willing to put some more work in if it'd actually have a chance of getting merged.

@joshgoodwin I'm not sure hack is possible. @caseq is re-factoring our network domain on the chromium-side, there are plans to complete refactoring by the end of this quarter. As the refactoring completes, we'll have intitiator awailable for interceptions.

joshgoodwin:
@aslushnikov Cool, glad to hear about the refactor. I'll just stick with my current hack (this patch) until the refactor hits. My change is simple enough that the merge is easy compared to my old hack that worked in the 'request' event. I might give an hour this weekend to seeing if I can put together a hack that works with the current NetworkManager.js just to see if it's possible. 

Thanks, I'll close this patch out.

PR #1833, opened 2018-01-17
feat(page): Add Network.getAllCookies, which returns all the cookies in the cookie jar.

codersbrew:


codersbrew:
Good to go bot overlord!

aslushnikov:
Hey @codersbrew, thank you for the PR.

This has been suggested a few times already, e.g. in #635. Long story short: we'd like the `browser.allCookies()` instead of `page.allCookies()`, and this requires chromium contributions.

Are you interested in working on chromium to support this?

aslushnikov:
Looks like there's no interest in upstream contributions. Closing this for now.

alixaxel:
@aslushnikov Seems like a useful addition, and I would be willing to contribute upstream but is there a contributors guideline or something? Not exactly sure how to begin...

aslushnikov:
@alixaxel here are a few useful pointers:
- the general chromium contribution guide is here: http://dev.chromium.org/developers/contributing-code
- design docs and cheat sheets on chromium subsystems: http://dev.chromium.org/developers
- some information on browser protocol could be found here: https://github.com/ChromeDevTools/devtools-frontend/wiki

The goal is to add a `browser.getCookies()` protocol command that will fetch all the cookies.

neilwashere:
This seems like a lost opportunity for community engagement. Whilst I understand the domain boundary for a `browser` is preferable to a `page` where obtaining all cookies is concerned, this PR does, nonetheless, mirror the exact feature as it appears in chrome. As this is actually a useful feature and consistent with upstream, why not just take it? Following on from the other conversations it seems that `page.cookies()` will also become deprecated when cookies are browser scoped. Nothing lost, only gained. Just random thoughts from someone else who has implemented getAllCookies in lieu of nothing here.

PR #1845, opened 2018-01-18
docs(troubleshooting): fix a link

vsemozhetbyt:
It seems without the full URL this link is not rendered correctly.


ebidel (approved):
Thanks!

PR #1864, opened 2018-01-20
feat: add slowMo option to Launcher.connect

atabel:


lib/Launcher.js
```diff
@@ -196,7 +196,8 @@ class Launcher {
    * @return {!Promise<!Browser>}
    */
   static async connect(options = {}) {
-    const connection = await Connection.create(options.browserWSEndpoint);
+    const connectionDelay = options.slowMo || 0;
+    const connection = await Connection.create(options.browserWSEndpoint, connectionDelay);
```
JoelEinbinder, line 200: You can pass options.slowMo in directly without || 0. This log will be handled by `Connection.create`

JoelEinbinder (approved):
Thanks!

lib/Launcher.js
atabel: You're right, I've just followed the same pattern in https://github.com/GoogleChrome/puppeteer/pull/1864/files/bf40675d4b311487493fe2a47e1c224736d254d7#diff-cd756d74c4f724a8d0ebe5f47fff175aR139

PR #1867, opened 2018-01-21
Add more explanation for page.evaluateHandle()

yelisaveta:


yelisaveta:
I signed it!

docs/api.md
```diff
@@ -695,22 +695,21 @@ List of all available devices is available in the source code: [DeviceDescriptor
 #### page.evaluate(pageFunction, ...args)
 - `pageFunction` <[function]|[string]> Function to be evaluated in the page context
 - `...args` <...[Serializable]|[JSHandle]> Arguments to pass to `pageFunction`
-- returns: <[Promise]<[Serializable]>> Resolves to the return value of `pageFunction`
+- returns: <[Promise]<[Serializable]>> Promise which resolves to the return value of `pageFunction` as a javascript-object
```
ebidel: JavaScript object.

examples/element-html.js
```diff
@@ -0,0 +1,46 @@
...
+
+  // page.evaluate() evaluates pageFunction in page context and returns javascript-object
+  const jsObject = await page.evaluate(pageFunction, elementHandle);
+
+  // page.evaluateHandler() evaluates pageFunction in page context and returns in-page object (JSHandle)
+  const jsHandle = await page.evaluateHandle(pageFunction, elementHandle);
+
+  // to show the difference between two methods
+  console.log(`jsObject is NOT equal to jsHandle: ${jsObject !== jsHandle}`); // prints true
+  console.log(`jsObject is equal to jsHandle.jsonValue(): ${jsObject === (await jsHandle.jsonValue())}`); // prints true
```
ebidel: Not sure if this is the best result to show int the demo, but I can't think of anything better.

ebidel (changes requested):

examples/element-html.js
yelisaveta: Alternatively, this example might just be part of the docs?

ebidel: Yea. I'm also thinking we should just have these snippets and explanation as part of the docs. It's a less compelling standalone demo.

yelisaveta:
@ebidel please take a look now: I've removed the example file and moved it to the api.md.

docs/api.md
```diff
@@ -729,19 +728,26 @@ Shortcut for [page.mainFrame().evaluate(pageFunction, ...args)](#frameevaluatepa
 #### page.evaluateHandle(pageFunction, ...args)
 - `pageFunction` <[function]|[string]> Function to be evaluated in the page context
 - `...args` <...[Serializable]|[JSHandle]> Arguments to pass to `pageFunction`
-- returns: <[Promise]<[JSHandle]>> Resolves to the return value of `pageFunction`
+- returns: <[Promise]<[JSHandle]>> Promise which resolves to the return value of `pageFunction` as in-page object (JSHandle)
 
-If the function, passed to the `page.evaluateHandle`, returns a [Promise], then `page.evaluateHandle` would wait for the promise to resolve and return its value.
+If the function passed to the `page.evaluateHandle` returns a [Promise], then `page.evaluateHandle` would wait for the promise to resolve and return its value.
 
+The only difference between `page.evaluate` and `page.evaluateHandler` is that `page.evaluate` returns JSON representation of the object and `page.evaluateHandler` returns in-page object (JSHandle):
```
ebidel: evaluateHandler -> evaluateHandle on both of these.

docs/api.md
```diff
@@ -729,19 +728,26 @@ Shortcut for [page.mainFrame().evaluate(pageFunction, ...args)](#frameevaluatepa
...
 ```js
-const aWindowHandle = await page.evaluateHandle(() => Promise.resolve(window));
-aWindowHandle; // Handle for the window object.
-```
+// returns JavaScript object
```
ebidel: "returns a JSON object" to be consistent with your description.

docs/api.md
ebidel: returns JSON representation -> returns a JSON representation

ebidel (changes requested):

yelisaveta:
@ebidel I've updated the PR, please take a look. Thanks!

yelisaveta:
Hi @ebidel @JoelEinbinder this PR is ready to be reviewed again, very quick one. Thanks.

docs/api.md
```diff
@@ -729,19 +728,26 @@ Shortcut for [page.mainFrame().evaluate(pageFunction, ...args)](#frameevaluatepa
 #### page.evaluateHandle(pageFunction, ...args)
 - `pageFunction` <[function]|[string]> Function to be evaluated in the page context
 - `...args` <...[Serializable]|[JSHandle]> Arguments to pass to `pageFunction`
-- returns: <[Promise]<[JSHandle]>> Resolves to the return value of `pageFunction`
+- returns: <[Promise]<[JSHandle]>> Promise which resolves to the return value of `pageFunction` as in-page object (JSHandle)
```
aslushnikov, line 826: There are many `evaluateHandle` calls: `page.evaluateHandle`, `frame.evaluateHandle` and `elementHandle.evaluateHandle`. We try to keep documentation consistent between them all; so all the changes has to be mirrored there as well.

docs/api.md
```diff
@@ -729,19 +728,26 @@ Shortcut for [page.mainFrame().evaluate(pageFunction, ...args)](#frameevaluatepa
...
+// returns a JSON object
+const jsObject = await page.evaluate(pageFunction, elementHandle);
 
-A string can also be passed in instead of a function.
+// returns in-page object (JSHandle)
+const jsHandle = await page.evaluateHandle(pageFunction, elementHandle);
+
+// both lines below print true
+console.log(`jsObject is NOT equal to jsHandle: ${jsObject !== jsHandle}`);
+console.log(`jsObject is equal to jsHandle.jsonValue(): ${jsObject === (await jsHandle.jsonValue())}`);
```
aslushnikov: I don't think this is actually the case: triple-equals checks for identity equality, which is not the case.

docs/api.md
```diff
@@ -729,19 +728,26 @@ Shortcut for [page.mainFrame().evaluate(pageFunction, ...args)](#frameevaluatepa
...
-```
+// returns a JSON object
+const jsObject = await page.evaluate(pageFunction, elementHandle);
 
-A string can also be passed in instead of a function.
+// returns in-page object (JSHandle)
+const jsHandle = await page.evaluateHandle(pageFunction, elementHandle);
+
+// both lines below print true
+console.log(`jsObject is NOT equal to jsHandle: ${jsObject !== jsHandle}`);
```
aslushnikov: I'd drop this `console.log` – even two consecutive `page.evaluate` will result in strict equality failing as well:
```js
const jsObject1 = await page.evaluate(pageFunction, elementHandle);
const jsObject2 = await page.evaluate(pageFunction, elementHandle);
console.log(jsObject1 === jsObject2); // false, unless pageFunction returns primitives
```

aslushnikov (commented):
Thanks for working on this.

docs/api.md
yelisaveta: But identity equality (aka strict equality) means that type and values are the same. In this case it shows that jsObject is the same type and value as jsHandle.jsonValue().

docs/api.md
yelisaveta: Good point, agree.

docs/api.md
yelisaveta: Cool, I will update there as well.

docs/api.md
aslushnikov: @yelisaveta in case of non-primitives, it means that two objects are the same by identity, not by value.

e.g.:
```js
console.log({} === {}); // false
let a = {};
let b = a;
let c = a;
console.log(b === c); // true
```

yelisaveta: Sorry for a delay - I've now understood why I used strict equality check - initially I've created an example to demonstrate the difference between these two methods, see here: https://github.com/GoogleChrome/puppeteer/pull/1867/commits/53ae96405076cd21c7b7183646237a9436e3e167

Examples were using `pageFunction` which returned `element.innterHTML` - which is a String. That's why strict equality worked correctly for me. 

I see that having these snippets in the docs without the `pageFunction` definition looks confusing.

docs/api.md
yelisaveta: Same as my comment above - it was working for me since I was using the `pageFunction` that returns `element.innterHTML`.

yelisaveta:
@aslushnikov PR has been updated, please take a look. Thank you.

aslushnikov (approved):

PR #1799, opened 2018-01-14
docs: Spelling and Markdown Consistency

andiamantoukos-fadv:
- Adding missing language tags to markdown code blocks.
- Switching `npm` to `yarn` in README to be consistent with what repo is using locally/with travis.

PR #1878, opened 2018-01-22
chore: switch to npm from yarn

JoelEinbinder:
npm is ok


aslushnikov (approved):
Thanks Joel!

.travis.yml
```diff
@@ -6,28 +6,24 @@ addons:
...
 cache:
-  yarn: true
   directories:
     - node_modules
 # allow headful tests
 before_install:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
-install:
-  - yarn install
```
paulirish, line 17: curious: why isn't the `npm install` step necessary any longer?

.gitignore
```diff
@@ -8,4 +8,5 @@
 *.pyc
 .vscode
 package-lock.json
```
paulirish, line 10: dont' you want to check in the package-lock.json now?

.travis.yml
JoelEinbinder: It was never necessary, but I finally read the docs.

.gitignore
JoelEinbinder: Nope. With developers using separate package managers, the locks would often be ignored. Checking in the lock files proved more of a headache than security it gives over our dev-dependencies.

aslushnikov: @paulirish package-lock.json is ignored on the `npm install puppeteer` step; on the trybots, we'd like to install the same way our users do.

paulirish: k sg!

ebidel (approved):
Seems fine. Why do we not like yarn though?

aslushnikov:
> Why do we not like yarn though?

@ebidel `yarn` is great, but there's some confusion on npm vs yarn usage in the docs: #1799.

We picked NPM since that's what dev team uses on a daily basis.

PR #1886, opened 2018-01-23
fix: fixed select() behavior to be on par with docs

alixaxel:
This PR fixes lost functionality that is no longer on-par with the documentation for `Page.select`, namely:

> `...values` <...string> Values of options to select. If the `<select>` has the `multiple` attribute, all values are considered, **otherwise only the first one is taken into account**.

I've also added an accompanying test for this use case.


aslushnikov (approved):
Thank you

PR #1928, opened 2018-01-30
Fix: Remove key codes from _pressedKeys Set after keyboard.up() is called

travch:
This is a one line change to resolve an issue where once a key has been utilized in keyboard.down(), the repeat property on keydown events using the same key will always be set to true. See https://github.com/GoogleChrome/puppeteer/issues/1901 for full details.

keyboard.down() and keyboard.up() both use the _pressedKeys Set, however keyboard.down() adds and searches for the key code, whereas keyboard.up() attempts to delete based on the key rather than the key code.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -2301,6 +2301,18 @@ describe('Page', function() {
       await page.keyboard.type('Hello World!');
       expect(await page.evaluate(() => textarea.value)).toBe('He Wrd!');
     });
+    it('should remove keys from _pressedKeys after keyboard.up()', async({page, server}) => {
+      await page.goto(server.PREFIX + '/input/textarea.html');
+      await page.focus('textarea');
+      await page.keyboard.down('w');
+      expect(await page.keyboard._pressedKeys.size).toEqual(1);
+      await page.keyboard.down('1');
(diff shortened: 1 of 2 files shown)


aslushnikov:
@travch thank you for the PR. Can you please add a test? If you have any questions on tests, feel free to ask.

travch:
@aslushnikov test has been added. Please let me know if you need anything specific for the test. Thanks for the awesome library.

test/test.js
```diff
@@ -2301,6 +2301,18 @@ describe('Page', function() {
       await page.keyboard.type('Hello World!');
       expect(await page.evaluate(() => textarea.value)).toBe('He Wrd!');
     });
+    it('should remove keys from _pressedKeys after keyboard.up()', async({page, server}) => {
+      await page.goto(server.PREFIX + '/input/textarea.html');
+      await page.focus('textarea');
+      await page.keyboard.down('w');
+      expect(await page.keyboard._pressedKeys.size).toEqual(1);
```
aslushnikov, line 2308: @travch It's really unfortunate to access private field here. Can we come up with a test that uses public API only?

JoelEinbinder (approved):
Let's merge this and fix the test later. This is a very good fix. Thanks!

PR #1933, opened 2018-01-31
docs(api.md): Fixing two minor anchor issues

cedric25:
- Menu item "page.$$(selector)" now pointing to the right anchor: "pageselector-1" instead of "pageselector"
- Within "page.$x(expression)", shortcut link now pointing to "framexexpression" instead of "frameexpression"

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -43,7 +43,7 @@
   * [event: 'requestfinished'](#event-requestfinished)
   * [event: 'response'](#event-response)
   * [page.$(selector)](#pageselector)
-  * [page.$$(selector)](#pageselector)
+  * [page.$$(selector)](#pageselector-1)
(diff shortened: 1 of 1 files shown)


cedric25:
I signed it!

docs/api.md
```diff
@@ -43,7 +43,7 @@
   * [event: 'requestfinished'](#event-requestfinished)
   * [event: 'response'](#event-response)
   * [page.$(selector)](#pageselector)
-  * [page.$$(selector)](#pageselector)
+  * [page.$$(selector)](#pageselector-1)
```
aslushnikov, line 46: unfortunately, the table-of-contents is generated, and there's a bug in a generator we use. 

So manual changes won't work.

cedric25: Alright, no problem, thanks for your answer :)

PR #1946, opened 2018-02-01
Correcting Chrome dependency size

oliviernt:
Whenever I install puppeteer I get a node_modules/puppeteers folder that is 282Mb (instead of the advertised ~90Mb).

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ npm i --save puppeteer
 # or "yarn add puppeteer"
 ```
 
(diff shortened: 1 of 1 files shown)


oliviernt:
Signed the CLA...

vsemozhetbyt:
FWIW, `.local-chromium` folder is 280 MB on Windows 7 x64.

oliviernt:
I wonder where the Readme values come from in that case... I have tried to install different versions but never came close to the ~90Mb

aslushnikov (approved):
Thank you. The numbers were about the download size rather then resulting `.local-chromium` folder.
Indeed, this was confusing.

Issue #1910, opened 2018-01-28
Behavior of setViewport() inconsistent in Headful


alixaxel:
**Environment:**

* Puppeteer: 1.0.0
* Platform / OS: Ubuntu 16.04 LTS (1920x1080@1x)
* Node.js: 8.9.3
* Chromium: 65.0.3312.0 (Developer Build) (64-bit)

**What steps will reproduce the problem?**

Running the following code several times.

```javascript
const puppeteer = require('puppeteer');

(async () => {
  const browser = await puppeteer.launch({
    args: [
      '--start-maximized',
    ],
    headless: false,
  });

  const page = await browser.newPage();

  /**
   * This should make Chromium infer current screen resolution (only in headless = false).
   */
  await page.setViewport({
    width: 0,
    height: 0,
  });

  await page.goto('http://matanich.com/test/viewport-width', {
    waitUntil: [
      'domcontentloaded',
      'load',
    ],
  });

  let result = await page.evaluate(
    () => {
      return window.innerWidth;
    }
  );

  console.log(`Detected window.innerWidth to be ${result}.`);

  await browser.close();
})();
```

**What is the expected result?**

I should get the save `window.innerWidth` all the time.

**What happens instead?**

Seems that _sometimes_ `setViewport()` fails silently and fallbacks to the [default viewport size](https://github.com/GoogleChrome/puppeteer/blob/cb684ebbc4de271c7f2187f4e52c0b3a83cc0122/lib/Page.js#L58) of 800x600.

```
$ node setViewportTest.js
Detected window.innerWidth to be 1855.

$ node setViewportTest.js
Detected window.innerWidth to be 1855.

$ node setViewportTest.js
Detected window.innerWidth to be 800.

$ node setViewportTest.js
Detected window.innerWidth to be 1855.

$ node setViewportTest.js
Detected window.innerWidth to be 1855.

$ node setViewportTest.js
Detected window.innerWidth to be 1855.

$ node setViewportTest.js
Detected window.innerWidth to be 800.
```

alixaxel:
Also experimented adding a `await page.waitFor(500);` after `setViewport()` and `goto()`.

Same results. :confused:

aizuyan:
```sh
➜  testChromium node t.js
Detected window.innerWidth to be 800.
➜  testChromium node t.js
Detected window.innerWidth to be 800.
➜  testChromium node t.js
Detected window.innerWidth to be 800.
➜  testChromium node t.js
[...]

PR #1954, opened 2018-02-02
docs(api): document appMode

ebidel:
I've found `appMode` useful when using pptr with lighthouse. 

Background:
Both Lighthouse and Puppeteer emulate a viewport when you run the tools, but they use different sizes. This causes lighthouse's perf audits to fail because screenshots end up changing sizes during an audit. During an audit session, Lighthouse reloads the page several times.

example  - inject a stylesheet into the page before each lighthouse runs.

```js
const {URL} = require('url');
const lighthouse = require('lighthouse');

(async() => {

const url = 'https://www.chromestatus.com/features';

  appMode: true, // sets headless: true and does not set a viewport size. Let ligthhouse do that.
});

browser.on('targetchanged', async target => {
  const page = await target.page();
  if (page && page.url() === url) {
  await page.addStyleTag({content: 'body {color: red}'});
  }
});

const lhr = await lighthouse(url, {
  port: (new URL(browser.wsEndpoint())).port,
  output: 'json',
  logLevel: 'info',
  // disableDeviceEmulation: true,
});

console.log(`Lighthouse score: ${lhr.score}`);

await browser.close();

})();
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -288,6 +288,7 @@ This methods attaches Puppeteer to an existing Chromium instance.
   - `userDataDir` <[string]> Path to a [User Data Directory](https://chromium.googlesource.com/chromium/src/+/master/docs/user_data_dir.md).
   - `env` <[Object]> Specify environment variables that will be visible to browser. Defaults to `process.env`.
   - `devtools` <[boolean]> Whether to auto-open DevTools panel for each tab. If this option is `true`, the `headless` option will be set `false`.
+  - `appMode` <[boolean]> If this option is set to `true`, the browser opens in headful mode and Puppeteer does not set a default viewport of 800x600 for new pages. Overrides the `headless` option if both options are used together.
 - returns: <[Promise]<[Browser]>> Promise which resolves to browser instance.
 
 The method launches a browser instance with given arguments. The browser will be closed when the parent node.js process is closed.


alixaxel:
I had to find out about this the hard way.

It would be useful to be able to not set a default viewport size *and* still be able to have `headless` setting. Related #1910.

aslushnikov:
@alixaxel @ebidel would the `page.setViewport(null)` to disable viewport emulation help you?

ebidel:
I think one of the issues we see is that that users have to remember to override the viewport in the first place. It's a source of confusion and easy to forget. It would also be nice to not have to make an API call for every page.

aslushnikov:
> I think one of the issues we see is that that users have to remember to override the viewport in the first place. It's a source of confusion and easy to forget. It would also be nice to not have to make an API call for every page.

@ebidel I think this would always be the case: If we don't override the viewport by default, it's size will be unpredictable.

alixaxel:
@aslushnikov For me, the appMode is confusing because it does too many things:

https://github.com/GoogleChrome/puppeteer/blob/e998ac93250cc7279043831328b6f4f49bc244bb/lib/Launcher.js#L68-L71

https://github.com/GoogleChrome/puppeteer/blob/e998ac93250cc7279043831328b6f4f49bc244bb/lib/Launcher.js#L83-L90

And then `Page` will set the viewport:

https://github.com/GoogleChrome/puppeteer/blob/e998ac93250cc7279043831328b6f4f49bc244bb/lib/Page.js#L57-L58

In summary, setting `appMode: true` has the effect of:

- turning off `headless` mode
- ignoring the option of `ignoreDefaultArgs`
- not setting the viewport size to the default 800x600
- not using `AUTOMATION_ARGS`:

https://github.com/GoogleChrome/puppeteer/blob/e998ac93250cc7279043831328b6f4f49bc244bb/lib/Launcher.js#L50-L54

In my case, I just want the viewport (in headful) to always be set to the resolution that I expect.

@ebidel In your example you have:

```
appMode: true, // sets headless: true and does not set a viewport size. Let ligthhouse do that.
```

Which doesn't agree with the docs you wrote (correct version):

```
- `appMode` <[boolean]> If this option is set to `true`, the browser opens in headful mode and Puppeteer does not set a default viewport of 800x600 for new pages. Overrides the `headless` option if both options are used together.
```

aslushnikov:
> @aslushnikov For me, the appMode is confusing because it does too many things

@alixaxel that's true. `appMode` is purely experimental and is a subject for removal; that's the reason it's not documented yet (and shouldn't be).

I'm not sure I understand if the `page.setViewport(null)` would help you. Is it enough?

alixaxel:
@aslushnikov I don't know what's the behavior/idea of `page.setViewport(null)` TBH. Basically, it would be helpful to find a solution for the inconsistency in #1910, which (I assume) will also affect Lighthouse.

If `page.setViewport(null)` would do the trick, great!

ebidel:
> @ebidel I think this would always be the case: If we don't override the viewport by default, it's size will be unpredictable.

Guess I'm unclear why we set a viewport. Chrome seems to be fine without setting one viewport:

```js
(async () => {
  const browser = await puppeteer.launch({appMode: true});
  await page.goto('https://www.facebook.com/');
  await page.screenshot({path: 'screenshot.png'});
  await browser.close();
})();
```

That produces a screenshot where the page/viewport take up the full window dimensions. That feels like the right behavior over pptr setting one for users. It's also more like the rest of the API (less magic, you opt-in to things).

Does chrome not launch with the same dimensions across platforms?

JoelEinbinder:
It doesn't launch with the same dimensions across headful/headless. I can't imagine its the same across platform.

ebidel:
Can we make a decision on this? 

TLDR: we need some way to prevent pptr from setting a default viewport. It would be nice if that was a flag and not an explicit API call. The default 800x600 is a source of gotcha/bugs for us.

BTW, does `page.setViewport(null)` currently work?

aslushnikov:
> Can we make a decision on this?

`appMode` is a highly experimental feature that serves as our playground; it's subject to change every other day and shouldn't be relied upon.

> The default 800x600 is a source of gotcha/bugs for us.

I'm not sure I follow this. Would it be better if there are random numbers instead?

> BTW, does page.setViewport(null) currently work?

It doesn't.

Closing this in favor of  https://github.com/GoogleChrome/puppeteer/issues/1910

ebidel:
> Would it be better if there are random numbers instead?

No, it would be better if pptr didn't set a default viewport at all for users. [Seems to work just fine](https://github.com/GoogleChrome/puppeteer/pull/1954#issuecomment-363229670).  People run into issues with responsiveness: e.g. site not matching selectors, page layout being different for mobile, etc. It's usually b/c they don't realize we're setting a small viewport at launch.

paulirish:
> No, it would be better if pptr didn't set a default viewport at all for users. Seems to work just fine. People run into issues with responsiveness: e.g. site not matching selectors, page layout being different for mobile, etc. It's usually b/c they don't realize we're setting a small viewport at launch.

We discussed this more offline. Many sites are clipped when screenshotted by puppeteer in a user's first hello world script. 

There's a proposal for changing the default viewport width from it's current 800x600. A fresh profile headful Chrome launches at 1200px wide, so we think that's a good new width. (height TBD, not as big a deal)   This will require a major version bump, so it should go onto a list of things to change at the 2.0. :)

PR #1964, opened 2018-02-04
Allow Timeouts of 0

### LINKED ISSUE #1960: waitForSelector should accept 0 for timeout option
Currently the `waitForSelector` is evaluating `timeout` like this: https://github.com/GoogleChrome/puppeteer/blob/189deb84c0ecf43f405961814aad840c1107ee85/lib/FrameManager.js#L640

For me, this is obscure behavior. I ran into the case where I had already waited for other preconditions that ensure the DOM is in the right state and I need to check if _either one of **two** possible elements_ exist on the page, immediately. It's a fair assumption that `timeout = 0` would mean immediately, but because of the fallback, it ends up spending 30 seconds looking for an element that is not there and won't be there.

alixaxel:
Closes #1960 .

It also addresses the timeout in `Puppeteer.launch()`, as [per the docs](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteerlaunchoptions):

> `timeout` <number> Maximum time in milliseconds to wait for the browser instance to start.
> Defaults to `30000` (30 seconds). **Pass `0` to disable timeout.**

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -618,7 +618,7 @@ class Frame {
    * @return {!Promise}
    */
   waitForFunction(pageFunction, options = {}, ...args) {
-    const timeout = (options.timeout != null) ? options.timeout : 30000;
+    const timeout = helper.isNumber(options.timeout) ? options.timeout : 30000;
     const polling = options.polling || 'raf';
     return new WaitTask(this, pageFunction, polling, timeout, ...args).promise;
   }
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -618,7 +618,7 @@ class Frame {
    * @return {!Promise}
    */
   waitForFunction(pageFunction, options = {}, ...args) {
-    const timeout = options.timeout || 30000;
+    const timeout = (options.timeout != null) ? options.timeout : 30000;
     const polling = options.polling || 'raf';
     return new WaitTask(this, pageFunction, polling, timeout, ...args).promise;
   }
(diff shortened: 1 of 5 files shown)


alixaxel:
Seems like AppVeyor is not playing nice with the `should work with a rotated element` unit test.

I've increased the test limit, but the problem remains, and it's not related to this PR: #1954 also fails.

lib/FrameManager.js
```diff
@@ -618,7 +618,7 @@ class Frame {
    * @return {!Promise}
    */
   waitForFunction(pageFunction, options = {}, ...args) {
-    const timeout = options.timeout || 30000;
+    const timeout = (options.timeout != null) ? options.timeout : 30000;
```
aslushnikov: nit: Let's use `helper.isNumber` instead.
```js
const timeout = helper.isNumber(options.timeout) ? options.timeout : 30000;
```

test/test.js
```diff
@@ -54,7 +54,7 @@ const defaultBrowserOptions = {
   args: ['--no-sandbox', '--disable-dev-shm-usage']
 };
 
-const timeout = slowMo ?  0 : 10 * 1000;
+const timeout = slowMo ? 0 : 15 * 1000;
```
aslushnikov: unfortunately this doesn't help; for some reason appveyour/travis sometimes are very short on resources and take forever to run tests.

Let's split timeout changse out of the PR and discuss separately.

aslushnikov (commented):
@alixaxel awesome, thank you for the fix

lib/FrameManager.js
alixaxel: @aslushnikov Great, do you want me to rollback the eslint config to `== null` too?

test/test.js
alixaxel: Yeah, on my local it was also timing out, after increasing it to 15secs it started working.

Must be the Chromium revision.

lib/FrameManager.js
aslushnikov: > Great, do you want me to rollback the eslint config to == null too?

@alixaxel yes, I like keeping it as strict as possible. Thanks!

aslushnikov (approved):
Thanks for taking time.

Issue #1454, opened 2017-11-22
Properly handle target crashes


aslushnikov:
Crashed targets should be handled gracefully:
1. All pending commands should be rejected. This includes protocol messages and watchdogs, such as NavigatorWatcher / WaitTask
2. All subsequent commands to the crashed target should reject right away

Neither (1) or (2) happen with Puppeteer v0.13.0.

schelkun:
Why page is crashing? Some Chromium bug?

aslushnikov:
@schelkun might be many reasons: chromium bug, OOM, memory corruption etc.

bahattincinic:
@aslushnikov When will you fix this issue?

We got the same problem. Zombie chrome processes have been growing increasingly.  We can't kill them. AWS Lambda has a maximum process limit. (max 1024 process) That's why we reach the lambda limits

**For example;**
```
483 1 3.3 1.6 1226196 65408 ? Ssl 22:07 0:05 /var/lang/bin/node --max-old-space-size=870 --max-semi-space-size=54 --max-executable-size=109 --expose-gc /var/runtime/node_modules/awslambda/index.js
483 22 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 73 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 119 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 166 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 214 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 262 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 307 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 353 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 398 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 446 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 493 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 538 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
[...]

PR #1967, opened 2018-02-05
Handling WebSocket errors (fix to potential crashes)

GregRos:
Puppeteer doesn't handle WebSocket errors correctly. This pull request is meant to fix that.

Things puppeteer doesn't handle:

1. Errors that happen in the middle/end of the WS connection, by subscribing to the `ws.on('error')` event.
2. Less important, WS protocol errors that happen when calling `send`.

Not handling protocol errors of type (1) results in an error that cannot be handled and crashes the application, such as `ECONNRESET`. This is what happened to me.

Not handling errors of type (2) may cause weird behavior in functions that work with promises, since the `Connection.send` function can throw an exception before it returns a promise.

If this pull request is relevant, I'll elaborate on the changes.

### DIFF
--- a/test/puppeteer.spec.js
+++ b/test/puppeteer.spec.js
@@ -206,7 +206,7 @@ module.exports.addTests = function({testRunner, expect, PROJECT_ROOT, defaultBro
         const dumpioTextToLog = 'MAGIC_DUMPIO_TEST';
         let dumpioData = '';
         const {spawn} = require('child_process');
-        const options = Object.assign({dumpio: true}, defaultBrowserOptions);
+        const options = Object.assign({}, defaultBrowserOptions, {dumpio: true});
         const res = spawn('node',
             [path.join(__dirname, 'fixtures', 'dumpio.js'), PROJECT_ROOT, JSON.stringify(options), server.EMPTY_PAGE, dumpioTextToLog]);
         res.stderr.on('data', data => dumpioData += data.toString('utf8'));
--- a/test/test.js
+++ b/test/test.js
@@ -47,6 +47,7 @@ const defaultBrowserOptions = {
   executablePath,
   slowMo,
   headless,
+  dumpio: (process.env.DUMPIO || 'false').trim().toLowerCase() === 'true',
   args: ['--no-sandbox']
 };
 const browserWithExtensionOptions = {
@@ -120,11 +121,22 @@ describe('Page', function() {
     state.browser = null;
   });
 
-  beforeEach(async state => {
+  beforeEach(async(state, test) => {
     state.page = await state.browser.newPage();
+    const rl = require('readline').createInterface({input: state.browser.process().stderr});
+    test.output = '';
+    rl.on('line', onLine);
+    state.tearDown = () => {
+      rl.removeListener('line', onLine);
+      rl.close();
+    };
+    function onLine(line) {
+      test.output += line + '\n';
+    }
   });
 
   afterEach(async state => {
+    state.tearDown();
     await state.page.close();
     state.page = null;
   });
--- a/utils/testrunner/Reporter.js
+++ b/utils/testrunner/Reporter.js
@@ -83,6 +83,10 @@ class Reporter {
             console.log(stack.join('\n'));
           }
         }
+        if (test.output) {
+          console.log('  Output:');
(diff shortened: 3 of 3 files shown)


GregRos:
Didn't know about the CLA thing. I'll talk to my employers.

lib/Connection.js
```diff
@@ -70,8 +80,8 @@ class Connection extends EventEmitter {
     const id = ++this._lastId;
     const message = JSON.stringify({id, method, params});
     debugProtocol('SEND ► ' + message);
-    this._ws.send(message);
     return new Promise((resolve, reject) => {
+      this._ws.send(message);
```
aslushnikov: I don't think this change affects anything. Am I missing something?

lib/Browser.js
```diff
@@ -165,7 +173,8 @@ Browser.Events = {
   TargetCreated: 'targetcreated',
   TargetDestroyed: 'targetdestroyed',
   TargetChanged: 'targetchanged',
-  Disconnected: 'disconnected'
+  Disconnected: 'disconnected',
+  ConnectionError: 'connectionerror'
```
aslushnikov: can we issue "disconnected" event for this?

lib/Browser.js
```diff
@@ -38,11 +38,19 @@ class Browser extends EventEmitter {
...
     });
+    this._connection.on('error', this._onConnectionError.bind(this));
     this._connection.on('Target.targetCreated', this._targetCreated.bind(this));
     this._connection.on('Target.targetDestroyed', this._targetDestroyed.bind(this));
     this._connection.on('Target.targetInfoChanged', this._targetInfoChanged.bind(this));
   }
 
+  _onConnectionError(err) {
+    // we're not going to call the closeCallback if we closed due to an error
+    if (this.listenerCount(Browser.Events.ConnectionError) === 0) throw err;
```
aslushnikov: I'm not sure I follow this: what kind of situation does it handle?

GregRos: Well, the default logic of the `EventTarget` is that when there are no subscribers to an event called `error`, an error is thrown. I wanted to replicate that behavior while still allowing users to handle the event.

GregRos: And not call it error.

lib/Browser.js
GregRos: Well, my intention was to allow users to handle an error that would otherwise crash the application, and notify them such an error occurred. With `disconnected` you lose that information.

lib/Connection.js
GregRos: `_ws.send` can throw an exception, but the function returns a promise. If an exception is thrown, `.catch` would not work. This way the promise is returned and rejects if `send` throws an exception.

lib/Browser.js
aslushnikov: I see. Can we consider connection errors as failures and dispatch "disconnected" event instead? From the clients point of view, this situation is equal to the browser connection becoming non-functional.

GregRos: Alright, that makes sense.

I just realized that there also needs to be some code to clean up the browser process if the WS connection breaks.

aslushnikov: > I just realized that there also needs to be some code to clean up the browser process if the WS connection breaks.

I believe all the cleanup should happen on the page level, see #1454. But it can be addressed separately.

pbarbiero:
We have been experiencing this in production, however it looks like we resolved it by locking the `ws` dependency to `v3.3.2`. 

It looks like this pull should address this issue without having to pin the dependency in our package.json (using yarn), any chance this can get followed up on?

lib/Connection.js
```diff
@@ -121,6 +131,10 @@ class Connection extends EventEmitter {
       this._closeCallback = null;
     }
     this._ws.removeAllListeners();
+
+    // swallow all WS errors
+    this._ws.on('error', () => {
```
aslushnikov: I recently debugged a recurring travis failure to this; landed this as https://github.com/GoogleChrome/puppeteer/commit/80078d9526528c0ee7ed745b75595c7274a907f4

lems111:
@aslushnikov @GregRos 
Is this change still being worked on? I am still getting the "Page crashed" error in the latest release

PR #1970, opened 2018-02-05
feat(Frame): added click(), focus(), hover(), tap() and type()

alixaxel:
This PR delegates the methods originally implemented in `Page` class to the `Frame` class:

- `click()`
- `focus()`
- `hover()`
- `tap()`
- `type()`


aslushnikov (approved):
@alixaxel Thank you!

PR #1971, opened 2018-02-05
feat(Response): add Response.fromCache / Response.fromServiceWorker

### LINKED ISSUE #1551: Add Response.fromDiskCache/fromServiceWorker for better offline PWA testing

aslushnikov:
This patch:
- introduces `test/assets/cached` folder and teaches server to cache
  all the assets from the folder
- introduces `test/assets/serviceworkers` folder that stores all the
  service workers and makes them register with unique URL prefix
- introduces `Response.fromCache()` and `Response.fromServiceWorker()`
  methods

Fixes #1551.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -80,14 +80,18 @@ if (fs.existsSync(OUTPUT_DIR))
 
 beforeAll(async state  => {
   const assetsPath = path.join(__dirname, 'assets');
+  const cachedPath = path.join(__dirname, 'assets', 'cached');
+
   const port = 8907 + state.parallelIndex * 2;
   state.server = await SimpleServer.create(assetsPath, port);
+  state.server.enableHTTPCache(cachedPath);
   state.server.PREFIX = `http://localhost:${port}`;
   state.server.CROSS_PROCESS_PREFIX = `http://127.0.0.1:${port}`;
   state.server.EMPTY_PAGE = `http://localhost:${port}/empty.html`;
 
   const httpsPort = port + 1;
   state.httpsServer = await SimpleServer.createHTTPS(assetsPath, httpsPort);
+  state.httpsServer.enableHTTPCache(cachedPath);
   state.httpsServer.PREFIX = `https://localhost:${httpsPort}`;
   state.httpsServer.CROSS_PROCESS_PREFIX = `https://127.0.0.1:${httpsPort}`;
   state.httpsServer.EMPTY_PAGE = `https://localhost:${httpsPort}/empty.html`;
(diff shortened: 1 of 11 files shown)


test/test.js
```diff
@@ -2663,8 +2667,40 @@ describe('Page', function() {
...
+      const responses = new Map();
+      page.on('response', r => responses.set(r.url().split('/').pop(), r));
+
+      // Load and re-load to make sure it's cached.
+      await page.goto(server.PREFIX + '/cached/one-style.html');
+      await page.reload();
+
+      expect(responses.size).toBe(2);
+      expect(responses.get('one-style.html').status()).toBe(304);
+      expect(responses.get('one-style.html').fromCache()).toBe(false);
```
JoelEinbinder, line 2685: Can we have a 200 code where it wasn’t cached?

JoelEinbinder (approved):

test/test.js
aslushnikov: Do you mean instead of `304` we currently have?
This will make the test less isolated: if there's some other test that loads the same URL, it will be cached.

For this reason the test explicitly reloads the page to trigger caching.

JoelEinbinder: In addition to. Have a css file that has cache busting.

aslushnikov: The cache-busting example is covered in the other place: see line 2670. Is it enough?

JoelEinbinder: Ah I missed that. All good

PR #1972, opened 2018-02-06
feat(ExecutionContext): introduce ExecutionContext.frame()

aslushnikov:
This patch introduces ExecutionContext.frame() that returns Frame
associated with this Execution Context.

This allows to associate console messages with the originating frame,
if any.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -167,6 +167,7 @@
 - [class: ExecutionContext](#class-executioncontext)
   * [executionContext.evaluate(pageFunction, ...args)](#executioncontextevaluatepagefunction-args)
   * [executionContext.evaluateHandle(pageFunction, ...args)](#executioncontextevaluatehandlepagefunction-args)
+  * [executionContext.frame()](#executioncontextframe)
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -1961,6 +1962,11 @@ await aHandle.dispose();
 await resultHandle.dispose();
 ```
 
+#### executionContext.frame()
+- returns: <?[Frame]> Frame associated with this execution context.
+
+> **NOTE** Not every execution context is associated with a frame.
```
JoelEinbinder: It looks like workers and content scripts have null frames? The note here is not helpful, it doesn’t explain why.

JoelEinbinder:
Why do we need to associate console messages with frames?

aslushnikov:
@JoelEinbinder IIRC, someone wanted to wait for the frame to load relying on the `console.log` it emitted. I found this patch locally; someone on the bugtracker wanted this sometime ago.

docs/api.md
aslushnikov: @JoelEinbinder added a better explanation.

JoelEinbinder (approved):

PR #1973, opened 2018-02-06
docs(CONTRIBUTING): update contributing.md

aslushnikov:


CONTRIBUTING.md
```diff
@@ -45,6 +45,14 @@ To run code linter, use:
 npm run lint
 ```
 
+## API guidelines
+
+When authoring new API methods, consider the following:
+- expose as little as needed. When in doubt, don't expose
```
JoelEinbinder: expose as little information as needed. When in doubt, don’t expose new information.

CONTRIBUTING.md
```diff
@@ -118,9 +126,9 @@ Despite being named 'unit', these are integration tests, making sure public API
 ```
 npm run unit
 ```
-- To filter tests by name:
+- To run tests in parallel, use `-j` flag:
 ```
-npm run unit -- --filter=waitFor
```
JoelEinbinder, line 123: Can we still filter?

aslushnikov: We can't. Though it's trivial to implement if we ever need it back.

CONTRIBUTING.md
aslushnikov: Done.

CONTRIBUTING.md
```diff
@@ -45,6 +45,14 @@ To run code linter, use:
 npm run lint
 ```
 
+## API guidelines
+
+When authoring new API methods, consider the following:
+- expose as little information as needed. When in doubt, don’t expose new information
+- methods are used in favor of getters/setters
+  - the only exception is namespaces, e.g. `page.keyboard` and `page.coverage`
+- all string literals must be small case. This includes event names and option values
```
ebidel, line 54: There have been questions in slack about adding new sugar apis (`page.exist()`). It might be nice to call out a bullet that discusses our philosophy on adding new apis.

aslushnikov: @ebidel good point; added a note

JoelEinbinder (approved):

ebidel (approved):

Issue #85, opened 2017-07-17
Support browser contexts to launch different sessions


aslushnikov:
Support browser contexts (Target.createBrowserContext) so to avoid launching multiple instances if one wants a pristine session. See proposal https://github.com/GoogleChrome/puppeteer/issues/66 and original discussion at https://github.com/cyrus-and/chrome-remote-interface/issues/118.

A viable user scenario might be testing several users logged in simultaneously into the service.
We might expose browser contexts as a string literal option to the browser.newPage:
```js
browser.newPage(); // creates a new page in a default browser context
browser.newPage({ context: 'default' }); // same as previous call
browser.newPage({ context: 'another-context' }); // creates a page in another browser context
```

miles-collier:
When this is implemented which way should be utilized? 
1. Having a constant chrome instance spawned all the time that the `newPages` connect to?
2. Business as usual, launching the chrome as usual however with the `newPages` context would it auto hijack an existing chrome instance if it sees it? 

I guess it will become more clear as it develops, just seems a bit confusing to wrap my head around right now. Especially with how internally contexts are going to be handled.

ks07:
This would be very useful for me - currently I'm forced to start multiple chrome instances if I'm wanting to have multiple simultaneous sessions on the same site.

Is there any way to workaround this currently? Or do I need to revert to using chrome-remote-interface to access context creation?

jamesmartinduffy:
+1, this feature would be crucial to me as well. Right now simulating 5+ simultaneous users is very memory intensive, since each user needs to launch its own chrome instance.

Would be useful to be able to use an unlimited number of browser contexts to simulate the desired number of users while using only one browser instance.

ks07:
In case anybody is still stuck waiting on this, it's a surprisingly easy change to hack it working in the very simple case where you want a new context for every page. All that is required is adding a call to `Target.createBrowserContext` at the top of `Browser.newPage`, and passing the returned `contextID` into the call to `Target.createTarget` (viewing how this is done at cyrus-and/chrome-remote-interface#118 might make this clearer).

Note that depending on what you're doing, you will probably want to add some support to delete these browser contexts when pages are closed - but if your Puppeteer instance is short-lived then this might not matter.

aslushnikov:
@ks07 FYI: browser contexts are **not supported** in headful mode (which is the reason we wait on implementing this in pptr).

icarusmiles:
[...]

PR #1978, opened 2018-02-06
feat: add Browser.newIsolatedPage

CodyRay:
One use case for #85 is being able to open up many isolated pages. In our case, we needed each page to have its own auth cookie. I got the impression that the puppeteer team was holding off implementing something like this until it was available in headful chrome, but we found this to be very helpful even with that caveat so I figured I would give a PR a shot.

### DIFF
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -113,6 +113,24 @@ class Browser extends EventEmitter {
   }
 
   /**
+   * @return {!Promise<!Page>}
+   */
+  async newIsolatedPage() {
+    const {browserContextId} = await this._connection.send('Target.createBrowserContext');
+    const {targetId} = await this._connection.send('Target.createTarget', {url: 'about:blank', browserContextId});
+    this._connection.on('Target.targetDestroyed', async function disposeBrowserContext(event) {
+      if (event.targetId === targetId) {
+        await this._connection.send('Target.disposeBrowserContext', {browserContextId});
+        this._connection.removeListener('Target.targetDestroyed', disposeBrowserContext);
+      }
+    }.bind(this));
+    const target = await this._targets.get(targetId);
+    console.assert(await target._initializedPromise, 'Failed to create target for page');
+    const page = await target.page();
+    return page;
+  }
+
+  /**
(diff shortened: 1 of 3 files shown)


aslushnikov:
Thank you for the nice patch with tests.

>  I got the impression that the puppeteer team was holding off implementing something like this until it was available in headful chrome

@haroldhues that's right. We already suffer from `page.pdf` being headless-only and breaking our debugging story; I don't want to worsen the situation.

I think you still able to implement this functionality atop of `target.createCDPSession()`; will it work for you?

CodyRay:
@aslushnikov Ah, so you are saying I could probably do more or less the same thing with the `CDPSession` session that I do in my commit here -- but without the need to patch puppeteer? If so, it looks like the only way to get a new session is to already have a page created? How would I go about creating a `CDPSession` with a new target?

Thanks!

lib/Browser.js
```diff
@@ -112,6 +112,24 @@ class Browser extends EventEmitter {
     return page;
   }
 
+  /**
+   * @return {!Promise<!Page>}
+   */
+  async newIsolatedPage() {
+    const {browserContextId} = await this._connection.send('Target.createBrowserContext');
+    const {targetId} = await this._connection.send('Target.createTarget', {url: 'about:blank', browserContextId});
+    this._connection.on('Target.targetDestroyed', async function disposeBrowserContext(event) {
```
alixaxel, line 121: Elegant solution for what it's worth.

aslushnikov:
> @aslushnikov Ah, so you are saying I could probably do more or less the same thing with the CDPSession session that I do in my commit here -- but without the need to patch puppeteer? If so, it looks like the only way to get a new session is to already have a page created? How would I go about creating a CDPSession with a new target?

You're right; let me introduce the `browser.createCDPSession()` so that you can connect to the browser target.

aslushnikov:
@haroldhues the https://github.com/GoogleChrome/puppeteer/pull/2036 exposes the "browser" target you can connect to. You'll be able to implement this patch atop of it then.

PR #1983, opened 2018-02-06
feat: Introduce BrowserFetcher class

### LINKED ISSUE #1748: Expose Downloader
[Downloader](https://github.com/GoogleChrome/puppeteer/blob/7c9104e1bc118ad227c7c4ff81025d6107317f8a/lib/Downloader.js) provides capabilities to download chromium from the upstream buckets. Exposing the API will drastically simplify devops and simplify everyone's life.

This has been requested a few times; one example is https://github.com/GoogleChrome/puppeteer/pull/1652.

aslushnikov:
This patch introduces `BrowserFetcher` class that manages
downloaded versions of products.

This patch:
- shapes Downloader API to be minimal yet usable for our needs. This
  includes removing such methods as `Downloader.supportedPlatforms` and
  `Downloader.defaultRevision`.
- makes most of the fs-related methods in Downloader async. The only
  exception is the `Downloader.revisionInfo`: it has stay sync due to the
  `pptr.executablePath()` method being sync.
- updates `install.js` and `utils/check_availability.js` to use new API
- finally, renames `Downloader` into `BrowserFetcher`

Fixes #1748.


lib/Downloader.js
```diff
@@ -34,70 +35,51 @@ const downloadURLs = {
...
+  let fulfill = null;
+  const promise = new Promise(x => fulfill = x);
+  fs.access(filePath, err => fulfill(!err));
+  return promise;
+}
 
+class Downloader {
   /**
-   * @return {!Array<string>}
+   * @param {!Object=} options
```
JoelEinbinder: can we jsdoc these options? It's a bit hard to read otherwise.

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+> **NOTE** Downloader class is not designed to work concurrently with other instances of Downloader class.
+
+#### downloader.canDownloadRevision(revision)
+- `revision` <[string]> a revision to check availability.
+- returns: <[Promise]<[boolean]>>  returns `true` if the revision could be downloaded from the host.
+
+The method initiates a HEAD request to check if revision is available.
+
+#### downloader.downloadRevision(revision[, progressCallback])
+- `revision` <[string]> a revision to check availability.
```
JoelEinbinder: a revision to download.

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
 > See [`this article`](https://www.howtogeek.com/202825/what%E2%80%99s-the-difference-between-chromium-and-chrome/) for a description
   of the differences between Chromium and Chrome. [`This article`](https://chromium.googlesource.com/chromium/src/+/lkcr/docs/chromium_browser_vs_google_chrome.md) describes some differences for Linux users.
+  
+### class: Downloader
+
+A Downloader is manages downloaded chromium instances.
+
+Downloader is created per platform and operates revisions: strings representing 6-digit numbers, e.g. `"533271"`. Revisions could be checked with [omahaproxy.appspot.com](http://omahaproxy.appspot.com/).
+
+> **NOTE** Downloader class is not designed to work concurrently with other instances of Downloader class.
```
JoelEinbinder: This makes me want to have it as a namespace with configurable options.

aslushnikov: This constrained can be removed with a lock file if we need to. I'd keep it as implementation detail without hardcoding into the API: it shouldn't yield much confusion.

docs/api.md
aslushnikov: Done

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+
+#### downloader.platform()
+- returns: <[string]> Downloader target platform. Returns one of the `mac`, `linux`, `win32` or `win64`.
+
+#### downloader.removeRevision(revision)
+- `revision` <[string]> a revision to remove. The method will throw if the revision is not downloaded.
+- returns: <[Promise]> Resolves when the revision is downloaded
+
+#### downloader.revisionInfo(revision)
+- `revision` <[string]> a revision to get info.
```
JoelEinbinder: to get info for.

aslushnikov: Done

lib/Downloader.js
```diff
@@ -34,70 +35,51 @@ const downloadURLs = {
...
+    this._downloadHost = options.host || DEFAULT_DOWNLOAD_HOST;
+    this._platform = options.platform || '';
+    if (!this._platform) {
+      const platform = os.platform();
+      if (platform === 'darwin')
+        this._platform = 'mac';
+      else if (platform === 'linux')
+        this._platform = 'linux';
+      else if (platform === 'win32')
+        this._platform = os.arch() === 'x64' ? 'win64' : 'win32';
```
JoelEinbinder, line 64: console.assert(this._platform, 'Unsupported platform: ' + os.platform());

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+- returns: <[Promise]> Resolves when the revision is downloaded
+
+#### downloader.revisionInfo(revision)
+- `revision` <[string]> a revision to get info for.
+- returns: <[Object]>
+  - `revision` <[string]> the revision the info was created from
+  - `folderPath` <[string]> path to the extracted revision folder
+  - `executablePath` <[string]> path to the revision executable
+  - `url` <[string]> URL this revision can be downloaded from
+  - `downloaded` <[boolean]> weather the revision is downloaded or not
```
JoelEinbinder: whether

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
 >
 > See [`this article`](https://www.howtogeek.com/202825/what%E2%80%99s-the-difference-between-chromium-and-chrome/) for a description
   of the differences between Chromium and Chrome. [`This article`](https://chromium.googlesource.com/chromium/src/+/lkcr/docs/chromium_browser_vs_google_chrome.md) describes some differences for Linux users.
+  
+### class: Downloader
+
+A Downloader is manages downloaded chromium instances.
```
JoelEinbinder: A Downloader can download and manage different versions of Chromium.

lib/Downloader.js
aslushnikov: Done.

docs/api.md

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+
+Downloader is created per platform and operates revisions: strings representing 6-digit numbers, e.g. `"533271"`. Revisions could be checked with [omahaproxy.appspot.com](http://omahaproxy.appspot.com/).
```
JoelEinbinder: The Downloader operates on revision strings that specify a precise version of Chromium, e.g. `"533271"`. Revision strings can be obtained from [omahaproxy.appspot.com](http://omahaproxy.appspot.com/).

docs/api.md
JoelEinbinder: The Downloader class ... of the Downloader class

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+- `revision` <[string]> a revision to download.
+- `progressCallback` <[function]> A function that will be called with two arguments:
+  - `downloadedBytes` <[number]> how many bytes are downloaded
+  - `totalBytes` <[number]> how many bytes is the total download
+- returns: <[Promise]> Resolves when the revision is downloaded and extracted
+
+The method initiates a GET request to download revision from the host. 
+
+#### downloader.downloadedRevisions()
+- returns: <[Promise]<[Array]<[string]>>> a list of all revisions downloaded for the given platform
```
JoelEinbinder: storedRevisions

A list of all revisions previously downloaded for the current platform.

docs/api.md

docs/api.md

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+#### downloader.canDownloadRevision(revision)
+- `revision` <[string]> a revision to check availability.
+- returns: <[Promise]<[boolean]>>  returns `true` if the revision could be downloaded from the host.
+
+The method initiates a HEAD request to check if revision is available.
+
+#### downloader.downloadRevision(revision[, progressCallback])
+- `revision` <[string]> a revision to download.
+- `progressCallback` <[function]> A function that will be called with two arguments:
+  - `downloadedBytes` <[number]> how many bytes are downloaded
```
JoelEinbinder: have been downloaded

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+  - `totalBytes` <[number]> how many bytes is the total download
```
JoelEinbinder: How large is the total download.

docs/api.md
```diff
@@ -273,6 +281,13 @@ puppeteer.launch().then(async browser => {
 
 This methods attaches Puppeteer to an existing Chromium instance.
 
+#### puppeteer.createDownloader([options])
+- `options` <[Object]>
+  - `host` <[string]> A download host to be used. Defaults to `https://storage.googleapis.com`.
+  - `path` <[string]> A path for the downloads folder. Defaults `<root>/.local-chromium`, where `<root>` is puppeteer's package root.
+  - `platform` <[string]> Specify binary type. Possible values are: `mac`, `win32`, `win64`, `linux`. Defaults to the current platform.
```
JoelEinbinder: Drop: Specify binary type.

docs/api.md
```diff
@@ -273,6 +281,13 @@ puppeteer.launch().then(async browser => {
 
 This methods attaches Puppeteer to an existing Chromium instance.
 
+#### puppeteer.createDownloader([options])
+- `options` <[Object]>
+  - `host` <[string]> A download host to be used. Defaults to `https://storage.googleapis.com`.
+  - `path` <[string]> A path for the downloads folder. Defaults `<root>/.local-chromium`, where `<root>` is puppeteer's package root.
```
JoelEinbinder: Defaults to

docs/api.md
aslushnikov: Done

docs/api.md
aslushnikov: Done

docs/api.md
aslushnikov: Done

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+
+The Downloader operates on revision strings that specify a precise version of Chromium, e.g. "533271". Revision strings can be obtained from [omahaproxy.appspot.com](http://omahaproxy.appspot.com/).
+
+> **NOTE** Downloader class is not designed to work concurrently with other instances of Downloader class.
+
+#### downloader.canDownloadRevision(revision)
+- `revision` <[string]> a revision to check availability.
+- returns: <[Promise]<[boolean]>>  returns `true` if the revision could be downloaded from the host.
+
+The method initiates a HEAD request to check if revision is available.
```
JoelEinbinder: the revision

docs/api.md
aslushnikov: Done

docs/api.md
aslushnikov: Done

docs/api.md
aslushnikov: Done

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+
+#### downloader.downloadRevision(revision[, progressCallback])
+- `revision` <[string]> a revision to download.
+- `progressCallback` <[function]> A function that will be called with two arguments:
```
JoelEinbinder: <[function]([number], [number])>

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+  - `downloadedBytes` <[number]> how many bytes are downloaded
+  - `totalBytes` <[number]> how many bytes is the total download
+- returns: <[Promise]> Resolves when the revision is downloaded and extracted
+
+The method initiates a GET request to download revision from the host.
```
JoelEinbinder: This method ... the revision

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+
+#### downloader.downloadedRevisions()
+- returns: <[Promise]<[Array]<[string]>>> a list of all revisions downloaded for the given platform
+
+#### downloader.platform()
+- returns: <[string]> Downloader target platform. Returns one of the `mac`, `linux`, `win32` or `win64`.
+
+#### downloader.removeRevision(revision)
+- `revision` <[string]> a revision to remove. The method will throw if the revision is not downloaded.
+- returns: <[Promise]> Resolves when the revision is downloaded
```
JoelEinbinder: has been downloaded.

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+The method initiates a GET request to download revision from the host. 
+
+#### downloader.downloadedRevisions()
+- returns: <[Promise]<[Array]<[string]>>> a list of all revisions downloaded for the given platform
+
+#### downloader.platform()
+- returns: <[string]> Downloader target platform. Returns one of the `mac`, `linux`, `win32` or `win64`.
+
+#### downloader.removeRevision(revision)
+- `revision` <[string]> a revision to remove. The method will throw if the revision is not downloaded.
```
JoelEinbinder: has not been downloaded.

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+  - `totalBytes` <[number]> how many bytes is the total download
+- returns: <[Promise]> Resolves when the revision is downloaded and extracted
+
+The method initiates a GET request to download revision from the host. 
+
+#### downloader.downloadedRevisions()
+- returns: <[Promise]<[Array]<[string]>>> a list of all revisions downloaded for the given platform
+
+#### downloader.platform()
+- returns: <[string]> Downloader target platform. Returns one of the `mac`, `linux`, `win32` or `win64`.
```
JoelEinbinder: Returns one of ~the~ ...

docs/api.md
```diff
@@ -307,6 +322,48 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+- returns: <[Promise]> Resolves when the revision is downloaded
+
+#### downloader.revisionInfo(revision)
+- `revision` <[string]> a revision to get info for.
+- returns: <[Object]>
+  - `revision` <[string]> the revision the info was created from
+  - `folderPath` <[string]> path to the extracted revision folder
+  - `executablePath` <[string]> path to the revision executable
+  - `url` <[string]> URL this revision can be downloaded from
+  - `downloaded` <[boolean]> wether the revision is downloaded or not
```
JoelEinbinder: has been downloaded or not

install.js
```diff
@@ -25,14 +25,14 @@ if (process.env.NPM_CONFIG_PUPPETEER_SKIP_CHROMIUM_DOWNLOAD || process.env.npm_c
   return;
 }
 
-const Downloader = require('./lib/Downloader');
-const downloader = Downloader.createDefault();
+const downloadHost = process.env.PUPPETEER_DOWNLOAD_HOST || process.env.npm_config_puppeteer_download_host;
+
+const puppeteer = require('./index');
```
JoelEinbinder, line 30: Does this work on node 6???

JoelEinbinder (approved):
@paulirish can you take a look at the api? Thanks.

docs/api.md
```diff
@@ -273,6 +281,13 @@ puppeteer.launch().then(async browser => {
 
 This methods attaches Puppeteer to an existing Chromium instance.
 
+#### puppeteer.createDownloader([options])
+- `options` <[Object]>
+  - `host` <[string]> A download host to be used. Defaults to `https://storage.googleapis.com`.
+  - `path` <[string]> A path for the downloads folder. Defaults to `<root>/.local-chromium`, where `<root>` is puppeteer's package root.
+  - `platform` <[string]> Possible values are: `mac`, `win32`, `win64`, `linux`. Defaults to the current platform.
+- returns: <[Downloader]>
```
vsemozhetbyt: Missing bottom reference?

docs/api.md

docs/api.md

docs/api.md

docs/api.md

docs/api.md

docs/api.md

install.js
aslushnikov: Of course! the `index.js` is the one that checks for async/await support and picks the right puppeteer.

docs/api.md

test/test.js
```diff
@@ -113,6 +114,31 @@ afterAll(async({server, httpsServer}) => {
...
+      server.setRoute(revisionInfo.url.substring(server.PREFIX.length), (req, res) => {
+        server.serveFile(req, res, '/chromium-linux.zip');
+      });
+
+      expect(downloader.platform()).toBe('linux');
+      expect(await downloader.canDownload('100000')).toBe(false);
+      expect(await downloader.canDownload('123456')).toBe(true);
+
+      await downloader.download('123456');
+      expect(await readFileAsync(revisionInfo.executablePath, 'utf8')).toBe('LINUX BINARY\n');
```
paulirish: the executablePath is defined before the binary is downloaded?

aslushnikov: yes, the revisionInfo could be called for both downloaded and non-downloaded revisions.

docs/api.md
```diff
@@ -307,6 +322,49 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+- returns: <[Promise]> Resolves when the revision has been removed.
+
+#### downloader.revisionInfo(revision)
+- `revision` <[string]> a revision to get info for.
+- returns: <[Object]>
+  - `revision` <[string]> the revision the info was created from
+  - `folderPath` <[string]> path to the extracted revision folder
+  - `executablePath` <[string]> path to the revision executable
+  - `url` <[string]> URL this revision can be downloaded from
+  - `downloaded` <[boolean]> wether the revision has been downloaded or not
```
paulirish: whether

docs/api.md
```diff
@@ -307,6 +322,49 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
...
+- returns: <[Promise]<[boolean]>>  returns `true` if the revision could be downloaded from the host.
+
+The method initiates a HEAD request to check if the revision is available.
+
+#### downloader.download(revision[, progressCallback])
+- `revision` <[string]> a revision to download.
+- `progressCallback` <[function]([number], [number])> A function that will be called with two arguments:
+  - `downloadedBytes` <[number]> how many bytes have been downloaded
+  - `totalBytes` <[number]> how large is the total download.
+- returns: <[Promise]> Resolves when the revision is downloaded and extracted
```
paulirish: should it return revisionInfo? 
seems optional but kinda nice.

paulirish (commented):
This is the expected use, right?

```js
// I want to use a different revision than the one in package.json...

const downloader = puppeteer.createDownloader();
await downloader.download('621459');
const revisionInfo = downloader.revisionInfo('621459');

await puppeteer.launch({executablePath: revisionInfo.executablePath})
```

docs/api.md
paulirish: if i download and then remove a rev, what will revInfo.downloaded tell me?  does this property really mean "currently on disk" ?  maybe `available`?

also if i do any of these actions in separate downloader instances (assuming same host/path/platform), does it matter?

aslushnikov: > if i download and then remove a rev, what will revInfo.downloaded tell me? does this property really mean "currently on disk" ? 

Yes, that's 'currently on disk'.

> also if i do any of these actions in separate downloader instances (assuming same host/path/platform), does it matter?

There are no guarantees in this case. If two downloaders try to download the same revision simultaneously, they will fight with each other and fail in some weird way.

There's a note about this in the description of downloader class.

aslushnikov: Done

docs/api.md
aslushnikov: I like returning revisionInfo, done.

aslushnikov:
> This is the expected use, right?

@paulirish yeah. Also I'd expect different scripts for downloading and running, like we have `install.js`.

Also, with the `download` method returning revisionInfo, your example becomes a little nicer.

install.js
JoelEinbinder: What happens if you checkout from the GitHub and run npm install? Do we built before downloading somewhere?

aslushnikov: @JoelEinbinder yes, we build in the very first line of the script

docs/api.md
```diff
@@ -307,6 +322,54 @@ If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://
 >
 > See [`this article`](https://www.howtogeek.com/202825/what%E2%80%99s-the-difference-between-chromium-and-chrome/) for a description
   of the differences between Chromium and Chrome. [`This article`](https://chromium.googlesource.com/chromium/src/+/lkcr/docs/chromium_browser_vs_google_chrome.md) describes some differences for Linux users.
+  
+### class: Downloader
+
+A Downloader can download and manage different versions of Chromium.
```
ebidel: ChromiumDownloader? At first I thought this PR was for handing file downloads, but I see that's it for downloading Chromium. I wonder if the name should be more explicit?

`createDownloader` -> `createChromiumDownloader` too. Thoughts?

aslushnikov: @ebidel in a follow-up, I'll teach it to download content_shell builds, so it's not solely about chromium.

PR #2003, opened 2018-02-11
Add AWS Lambda solution

Kikobeats:
Hello, 

I added my solution to used puppeteer into an AWS Lambda, using a npm dependency with last chrome compatible version, compressed and auto-decompress on the fly.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -250,6 +250,5 @@ There's also another [simple guide](https://timleland.com/headless-chrome-on-her
 AWS Lambda [limits](https://docs.aws.amazon.com/lambda/latest/dg/limits.html) deployment package sizes to ~50MB. This presents challenges for running headless Chrome (and therefore Puppeteer) on Lambda. The community has put together a few resources that work around the issues:
 
 - https://github.com/adieuadieu/serverless-chrome/blob/master/docs/chrome.md (tracks the latest Chromium snapshots)
-- https://github.com/sambaiz/puppeteer-lambda-starter-kit (uses an old version of Chromium)
 - https://github.com/universalbasket/aws-lambda-chrome
 - https://github.com/Kikobeats/aws-lambda-chrome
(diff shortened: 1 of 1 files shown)


ebidel:
How is this one different than the resources we currently Iist? And version of chromium does it bundle?

Kikobeats:
Hey,

I added the chrome version on the repository. The current version is `66.0.3343.0`.

I found my solution very different from the current proposals:

**https://github.com/adieuadieu/serverless-chrome**
- It ships different chrome version in the npm package, so the size is heavy.
- It uses `chrome-launcher`, so actually it is adding another dependency non related with puppeteer.

**https://github.com/sambaiz/puppeteer-lambda-starter-kit**
- unmaintained, not ready for Puppeteer 1.x
- although the chrome version is ok, it doesn't ship a minimal code for be easy connect with puppeteer.

**https://github.com/universalbasket/aws-lambda-chrome**
- multiple bundles, so heavy.
- not initialization script.

ebidel:
Thanks for the explanation. 

I think we should get rid of https://github.com/sambaiz/puppeteer-lambda-starter-kit in that case. Can you update this PR to remove it?

Kikobeats:
done!

aslushnikov (approved):

smoll:
Thanks for creating this and adding it to the docs, however, it's not immediately obvious how to use `puppeteer` directly in the docs for https://github.com/Kikobeats/aws-lambda-chrome

Can you provide a code example/snippet explaining how to do so? It would be good for users who already have a few puppeteer scripts written and would like to start running them in AWS Lambda @Kikobeats

Kikobeats:
I added a more detailed example days ago:
https://github.com/Kikobeats/aws-lambda-chrome/blob/master/README.md

but yea I suppose we can reduce friction, we can extend the readme as part of https://github.com/Kikobeats/aws-lambda-chrome/issues/3

tomgallagher:
I've been having a look at all the various solutions for using Puppeteer on Lambda. Is there now a Puppeteer-officially-approved way that will be supported going forward?

PR #2015, opened 2018-02-12
feat(Page): add timestamp to Request and Response events

joeyvandijk:

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -327,6 +327,7 @@ class Request {
     this._headers = {};
     this._frame = frame;
     this._monotonicTimeStart = monotonicTimestamp;
+    this._monotonicTimeEnd = null;
     for (const key of Object.keys(payload.headers))
       this._headers[key.toLowerCase()] = payload.headers[key];
 
@@ -388,7 +389,7 @@ class Request {
   monotonicTimeStart() {
     return this._monotonicTimeStart;
   }
-  
+
   /**
    * Timestamp for when a request finished or failed.
    * @return {?number}
@@ -583,7 +584,7 @@ class Response {
   monotonicTimeEnd() {
     return this._monotonicTimeEnd;
   }
-  
+
   /**
    * @return {?SecurityDetails}
    */
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -153,8 +153,8 @@ class NetworkManager extends EventEmitter {
     if (event.redirectUrl) {
       const request = this._interceptionIdToRequest.get(event.interceptionId);
       if (request) {
-        this._handleRequestRedirect(request, event.responseStatusCode, event.responseHeaders, false /* fromDiskCache */, false /* fromServiceWorker */);
-        this._handleRequestStart(request._requestId, event.interceptionId, event.redirectUrl, event.resourceType, event.request, event.frameId);
+        this._handleRequestRedirect(request, event.responseStatusCode, event.responseHeaders, false /* fromDiskCache */, false /* fromServiceWorker */, null /* timestamp */);
+        this._handleRequestStart(request._requestId, event.interceptionId, event.redirectUrl, event.resourceType, event.request, event.frameId, null);
       }
       return;
     }
@@ -162,10 +162,10 @@ class NetworkManager extends EventEmitter {
     const requestId = this._requestHashToRequestIds.firstValue(requestHash);
     if (requestId) {
       this._requestHashToRequestIds.delete(requestHash, requestId);
-      this._handleRequestStart(requestId, event.interceptionId, event.request.url, event.resourceType, event.request, event.frameId);
+      this._handleRequestStart(requestId, event.interceptionId, event.request.url, event.resourceType, event.request, event.frameId, event.timestamp);
     } else {
       this._requestHashToInterceptionIds.set(requestHash, event.interceptionId);
-      this._handleRequestStart(null, event.interceptionId, event.request.url, event.resourceType, event.request, event.frameId);
+      this._handleRequestStart(null, event.interceptionId, event.request.url, event.resourceType, event.request, event.frameId, null);
     }
   }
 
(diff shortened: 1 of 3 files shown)


lib/NetworkManager.js
```diff
@@ -268,6 +268,8 @@ class NetworkManager extends EventEmitter {
     if (!request)
       return;
     request._completePromiseFulfill.call(null);
+    // save loading finished timestamp and not loading started timestamp
+    request._timestamp = event.timestamp;
```
aslushnikov: shall we have `request.timestamp()` as the start time, and `response.timestamp()` as the loading-finished time? This seems more logical to me.

aslushnikov (commented):
Thanks for the PR. This is the first use of MonotonicTime in our API - how do you use it? Do you pair it with tracing?

joeyvandijk:
The use-case is to align the received list of Network requests with a timestamp to identify in which order the came in. I'd like to have all the timing information Chromium provides through the Devtools protocol like https://developers.google.com/web/tools/chrome-devtools/network-performance/reference#requests but could not find it, so for now the timestamp is good.

lib/NetworkManager.js
joeyvandijk: I partly understand but how do we keep the timestamps that a request failed or finished?

Request.timestampFinished()
Request.timestampFailed()

perhaps on Request object? So when requests fail I still know the performance hit it got?

aslushnikov: > I partly understand but how do we keep the timestamps that a request failed or finished?

Good point. I like your suggestions; I'd prefix the time with `"monotonic"` so that it's less confusing.

```js
request.monotonicTimeStart() // when the request started
request.monotonicTimeEnd() // when the request failed / finished
```

cc @paulirish for API suggestions

joeyvandijk: I will wait a day or 2 on @paulirish before changing the code. I do not want to add extra metadata, but `request.hasFailed()` could be handy? Or just stick with `.monotonicxxx`?

aslushnikov: Request failed state is dispatched with events; do you want the `request.hasFailed()` for convenience?

joeyvandijk: @aslushnikov perhaps it is confusing for developers, but it was just a logical thought to show no `response` or `response.timestamp` can be expected if that request has failed.

joeyvandijk:
@aslushnikov I have updated the PR with the changes you proposed. The failing Travis build is by an error which I assume I did not create: `Error: read ECONNRESET` and only on Node.js 7. How can I retrigger the build or is this a bug?

rafaelugolini:
@aslushnikov any news on this?

joeyvandijk:
@aslushnikov could you have a check on this and perhaps initiate a review? Just curious when I can use this or what needs to change?

aslushnikov:
@joeyvandijk Sorry for the long response. I'm on vacation and didn't find a way to set status on github.

Regarding the API shape, I had some time to think and I ended up loving the `timing` namespace.
Here's the suggestion:

```js
request.timing.start() // monotonic time of request creation in renderer
request.timing.issued() // monotonic time of request been sent out to the server 
request.timing.headersReceived() // monotonic time of response headers received
request.timing.finished() // monotonic time of when we got all the response in
request.timing.failed() // monotonic time of when we got the failure
```

Some of the timings are available in [`Network.ResourceTiming`](https://aslushnikov.github.io/vanilla-protocol-viewer/#Network.ResourceTiming) object, some are dispatched with network events (e.g. [`loadingFinished`](https://aslushnikov.github.io/vanilla-protocol-viewer/#Network.loadingFinished) / [`loadingFailed`](https://aslushnikov.github.io/vanilla-protocol-viewer/#Network.loadingFailed) events).

Let me know what you think.

@paulirish: let me know if there are any concerns with this.

ebidel:
Why not `request.timing.metrics()` to get and obj of everything? We can add new properties in later and keep the API smaller. Also more consistency with `page.metrics()` and other web timing apis.

rafaelugolini:
@joeyvandijk still interested in doing it? If not do you mind if I take over? Really need this feature :)

joeyvandijk:
Go forward and implement further while for my use-case I will also get all
timings on client-side with window.performance. 👍

aslushnikov:
> Why not request.timing.metrics() to get and obj of everything? We can add new properties in later and keep the API smaller. Also more consistency with page.metrics() and other web timing apis.

Let's do `request.metrics()` then, to be even more consistent with `page.metrics()`

paulirish:
Some thoughts on naming..

`timing` typically means time durations, rather than timestamps. So the earlier proposal would be awkwardly mixing these. 

Timestamps are a little easier to maintain forward compatibility for, as diff duration spans become interesting over time. (appcache, service worker, etc). And if you're only representing timings, it becomes unclear what their sequence is, and how they add up. We've observed this challenge when implementing HAR.

### HAR spec
The [HAR spec](https://github.com/ahmadnassri/har-spec/blob/master/versions/1.3.md#timings) reports timings rather than timestamps. Their names: `blocked dns connect send wait receive ssl`.

### Resource timing spec
The [resource timing spec](https://www.w3.org/TR/resource-timing-2/#performanceresourcetiming) reports timestamps. Their names: ` workerStart redirectStart redirectEnd fetchStart domainLookupStart domainLookupEnd connectStart connectEnd secureConnectionStart requestStart responseStart responseEnd`


### chrome devtools


### @aslushnikov's earlier proposal:
```js
request.timing.start() // monotonic time of request creation in renderer
request.timing.issued() // monotonic time of request been sent out to the server 
request.timing.headersReceived() // monotonic time of response headers received
request.timing.finished() // monotonic time of when we got all the response in
request.timing.failed() // monotonic time of when we got the failure
```

---------------

So I think the proposal is a good start. But a few notes:

1. DevTools uses `started` to indicate the request was issued, so changing "started" in here is probably confusing.
1. DevTools considers the timestamp of failure to be the end time of the request
1. The proposal doesn't offer any timing details within the following: 
   - DNS lookup 
   - TCP handshake
   - TLS handshake 
   - service worker handling 
   - redirects 
   - proxy negotiation

That said, I think it's fine to punt including those details until later.

### Here's a revised proposal:

```js
request.timing.enqueued() // monotonic time of request creation in renderer
request.timing.sent() // monotonic time of when the (final post-redirect) request was sent out to the server. This is also post-DNS/TCP/TLS.
request.timing.headersReceived() // monotonic time of response headers received
request.timing.finished() // monotonic time of when the response completed (or when the request failed)
```

joeyvandijk:
This is costing too much time and while I really appreciate the input/feedback, it provides me coding the approach 3 different times. I did not mind the first approach, understood the 2nd approach and approve the 3rd approach, but what if I finish? While it be generic enough? If someone will follow-up, please do so @rafaelugolini but for my use-case I can also use window.performance. 

Perhaps in the (near) future I will need this approach and will implement as described by @paulirish . Sorry and thanks for the feedback, but this is not helping my use-case.

PR #2016, opened 2018-02-13
fix(test): don't race against the MutationObserver in waitForSelector tests

JoelEinbinder:


aslushnikov (approved):
@JoelEinbinder Nice catch!

PR #2032, opened 2018-02-14
feat(appMode): support pipes for appMode

aslushnikov:
This patch enhances `appMode` to use pipes as a transport to
the browser instance instead of websocket.

### DIFF
--- a/lib/Pipe.js
+++ b/lib/Pipe.js
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+const {helper} = require('./helper');
 const EventEmitter = require('events');
 
 class Pipe extends EventEmitter {
@@ -24,7 +25,9 @@ class Pipe extends EventEmitter {
     super();
     this._pipeWrite = pipeWrite;
     this._pendingMessage = '';
-    pipeRead.on('data', buffer => this._dispatch(buffer));
+    this._eventListeners = [
+      helper.addEventListener(pipeRead, 'data', buffer => this._dispatch(buffer))
+    ];
   }
 
   /**
@@ -57,7 +60,10 @@ class Pipe extends EventEmitter {
     this._pendingMessage = buffer.toString(undefined, start);
   }
 
-  close() { }
+  close() {
+    this._pipeWrite = null;
+    helper.removeEventListeners(this._eventListeners);
+  }
 }
 
 module.exports = Pipe;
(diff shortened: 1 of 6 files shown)


lib/externs.d.ts
```diff
@@ -25,5 +25,9 @@ export class JSHandle extends RealJSHandle {}
 export class ExecutionContext extends RealExecutionContext {}
 export class Page extends RealPage {}
 
+export interface ConnectionTransport extends NodeJS.EventEmitter {
+  send(any);
```
JoelEinbinder: buffer/String or something.

lib/Pipe.js
```diff
@@ -0,0 +1,63 @@
...
+    end = buffer.indexOf('\n', start);
+    while (end !== -1) {
+      this.emit('message', buffer.toString(undefined, start, end));
+      start = end + 1;
+      end = buffer.indexOf('\n', start);
+    }
+    this._pendingMessage = buffer.toString(undefined, start);
+  }
+
+  close() { }
```
JoelEinbinder: destroy streams

JoelEinbinder (approved):

lib/Pipe.js
aslushnikov: We don't own streams; I don't think we should destroy them here.

JoelEinbinder: Reasonable. Should we stop listening to them at least?

JoelEinbinder: Ah you did. Cool

paambaati:
@aslushnikov How does socket mode compare with app mode? What are the differences between the two?

schelkun:
Pipe? Why not IPC?

aslushnikov:
> @aslushnikov How does socket mode compare with app mode? What are the differences between the two?

We initially thought pipes would be a faster transport then websocket, but it turns out websockets a crazy fast and there are no speed gains.

The only difference is that pipe doesn't open an HTTP port that's accessible to everyone on the machine.

> Pipe? Why not IPC?

I'm not sure I follow the question: both pipes and HTTP are transports for IPC.


Please keep in mind: `appMode` is not part of PPTR API, it's not documented and serves as a playground for our experiments. It could be changed or removed any time in future.

schelkun:
@aslushnikov I was talking about this:
https://nodejs.org/api/child_process.html#child_process_subprocess_send_message_sendhandle_options_callback

aslushnikov:
@schelkun ah I see. This would require chromium to support `node.js`-compatible serialization-deserialization, making the chromium part of the IPC node-specific. Pipes, on the other hand, are agnostic to the platform.

PR #2038, opened 2018-02-16
Improved XPath Detection

alixaxel:
This patch extends the XPath detection from the basic prefix of `//` to the following prefix axis:

- `/` Root
- `//` Anywhere
- `./` Relative

It also detects the following [XPath node set functions](https://www.w3.org/TR/xpath/#section-Node-Set-Functions):

- `count(`
- `id(`
- `last(`
- `name(`
- `position(`

This is important because an XPath can look something like:

```
id("user-links")
count(//*)
```

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -631,11 +631,11 @@ class Frame {
    * @return {!Promise}
    */
   waitFor(selectorOrFunctionOrTimeout, options = {}, ...args) {
-    const xPathPattern = '//';
+    const xPathPattern = /^(?:[.][/]|[/][/]?|(?:count|id|last|name|position)[(])/i;
 
     if (helper.isString(selectorOrFunctionOrTimeout)) {
       const string = /** @type {string} */ (selectorOrFunctionOrTimeout);
-      if (string.startsWith(xPathPattern))
+      if (xPathPattern.test(string))
         return this.waitForXPath(string, options);
       return this.waitForSelector(string, options);
     }
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -631,7 +631,7 @@ class Frame {
    * @return {!Promise}
    */
   waitFor(selectorOrFunctionOrTimeout, options = {}, ...args) {
-    const xPathPattern = /^(?:[.][/]|[/][/]?|(?:count|id|last|name|position)[(])/i;
+    const xPathPattern = /^[(]?(?:[.][/]|[/][/]?|(?:count|id|last|name|position)[(])/i;
 
     if (helper.isString(selectorOrFunctionOrTimeout)) {
       const string = /** @type {string} */ (selectorOrFunctionOrTimeout);
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -631,7 +631,7 @@ class Frame {
    * @return {!Promise}
    */
   waitFor(selectorOrFunctionOrTimeout, options = {}, ...args) {
-    const xPathPattern = /^[(]?(?:[.][/]|[/][/]?|(?:count|id|last|name|position)[(])/i;
+    const xPathPattern = /^[(]?(?:[.][/]|[/][/]?|id[(])/i;
 
     if (helper.isString(selectorOrFunctionOrTimeout)) {
       const string = /** @type {string} */ (selectorOrFunctionOrTimeout);
--- a/test/test.js
+++ b/test/test.js
@@ -1044,11 +1044,25 @@ describe('Page', function() {
       await waitFor;
       expect(found).toBe(true);
     });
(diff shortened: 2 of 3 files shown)


lib/FrameManager.js
```diff
@@ -631,11 +631,11 @@ class Frame {
    * @return {!Promise}
    */
   waitFor(selectorOrFunctionOrTimeout, options = {}, ...args) {
-    const xPathPattern = '//';
+    const xPathPattern = /^(?:[.][/]|[/][/]?|(?:count|id|last|name|position)[(])/i;
```
aslushnikov: these will break behavior for css selectors with class names "count" or others.

alixaxel:
@aslushnikov If the pattern for the XPath functions need to match a `(` after the function name too, so it won't clash with CSS. Besides `count {something}` would only be usable if you had a `count` tag since it also needs to start with that string.

lib/FrameManager.js
alixaxel: Here are some samples: https://regex101.com/r/OQXuZ9/1

alixaxel: The pattern for the XPath functions needs to match a `(` after the function name too, so it won't clash with CSS. Besides `count {something}` would only be usable if you had a `count` tag since the string also needs to start with it and contain a parentheses after.

alixaxel: Made a small modification to also handle XPath groups, like `(//h1)[1]/text()`:

- https://regex101.com/r/OQXuZ9/2.

aslushnikov: > The pattern for the XPath functions needs to match a ( after the function name too, so it won't clash with CSS.

Ah, indeed. What about functional CSS selectors, e.g. `:not()` and `:nth-child()`? These require parentheses, and any of them could be added in future to CSS.

I'm concerned we'll end up with unpredictable behavior in certain cases.

alixaxel: You'd need to have a CSS selector that *starts with* the function name and is *followed by* `(`.

It would be a clear break of style if a selector like that would ever be added to the CSS standard.

Even if `:count()` or something similar is eventually introduced, it won't get matched.

aslushnikov: @alixaxel sg. Can we have some tests that illustrate all the heuristics with this detection?

alixaxel: @aslushnikov Sure, I'll write some.

alixaxel:
@aslushnikov Wrote some tests and removed all XPath functions except `id` because they were either returning `number` or `string` and not a `node-set` and thus couldn't be converted to a `ElementHandle` within PPTR. Also, because `waitFor` (unlike `Frame.$x`) is always using `document` as the `contextNode`, it's not possible to use relative XPaths of the form of `./{...}` but I still left the pattern for correctness sake.

test/test.js
```diff
@@ -1044,11 +1044,25 @@ describe('Page', function() {
...
       expect(found).toBe(true);
     });
-    it('should not allow you to select an element with single slash xpath', async({page, server}) => {
+    it('should wait for an absolute xpath', async({page, server}) => {
+      let found = false;
       await page.setContent(`<div>some text</div>`);
-      let error = null;
-      await page.waitFor('/html/body/div').catch(e => error = e);
-      expect(error).toBeTruthy();
+      await page.waitFor('/html/body/div').then(() => found = true);
```
aslushnikov, line 1050: Will the following be treated as CSS selector or XPath selector?

```js
await page.waitFor('/*comment*/html');
```

alixaxel: XPath.

But are those CSS comments something we wish to support in determent of plain XPaths?

alixaxel: We could have a negative lookahead in the pattern: https://regex101.com/r/OQXuZ9/5

```
(?![/][*])^[(]?(?:[.][/]|[/][/]?|id[(])
```

But the problem then would be that XPaths like `/*/div` would not be considered.

aslushnikov: > But are those CSS comments something we wish to support in determent of plain XPaths?

It would be nice to have a good xpath support. However, CSS comments don't look like an exotic scenario to me, so we shouldn't break it until the next major release.

alixaxel: IMO, I don't know why anyone would wait for a CSS selector with a comment, but YMMV. 

Still, do you think we could introduce the negative lookahead to deal with these comments?
It would still be an improvement XPath-wise and it wouldn't break the comment scenario.

aslushnikov: @alixaxel I don't think the negative lookahead will cover all the usecases.

Let's not complicate the heuristics here for now. We can revisit it later if there's a demand.

aslushnikov:
Closing this for now; let's revisit later if there's more demand.

PR #2056, opened 2018-02-19
Virtual time for Puppeteer

ADClarke:
This is a prototype for integrating Virtual Time with Puppeteer.  Virtual Time tries to make the browser deterministic (by controlling the renderer process's view of time, which is paused while there are pending network fetches) and to reduce the CPU load (not wall time) for renders by eliminating unnecessary compositing and JS execution. 

The general idea is the page should be (effectively) paused while it's being inspected, and after navigating or otherwise interacted with, some virtual time can be granted to let the page settle.  I've hooked this up to page.goto, page.waitFor() and page.screenshot.   

I've added a deterministic object to the launch parameters (which is used to turn virtual time on):  e.g.
   const browser = await puppeteer.launch({deterministic: {date: new Date()}});
   ...

The date controls what's returned by Date.now etc... in the browser.  Unless otherwise specified by deterministic:animationBeginFrameInterval, rAF now runs every 100 virtual milliseconds rather than at 60 fps.  RequestAnimationFrame no longer runs when virtual time is paused.

NB VirtualTime is an experimental chromium feature, and Chrome 66.0.3347.1 or newer is required for all of the flags to work.


lib/VirtualTimeController.js
```diff
@@ -0,0 +1,353 @@
...
+   */
+  virtualTimeStarted(virtualTimeOffset) {}
+
+  /**
+   * Called when virtual time stops because either all tasks have been canceled or one of them voted
+   * to stop.
+   *
+   * @param {number} virtualTimeOffset The time in virtual milliseconds since the renderer was created
+   */
+  virtualTimeSopped(virtualTimeOffset) {}
```
betasheet: virtualTimeStopped

lib/FrameManager.js
```diff
@@ -786,6 +790,7 @@ class WaitTask {
     });
     // Since page navigation requires us to re-install the pageScript, we should track
     // timeout on our end.
+    this._stopVirtualBeginframesCallback = null;
```
betasheet: _stopVirtualBeginFramesCallback

betasheet (commented):
Some nits. Otherwise this looks good, but probably needs some more testing / evaluation :)

One thing I'm wondering is if it's enough for puppeteer to apply these settings globally, or if we would really want a per-BrowserContext (theoretically possible?) or per-WebContents (more difficult) setting.

ADClarke:
I'm not really sure how this code review tool works, but I've fixed the two nits.

Re settings, currently whether or not to use virtual time is a global setting because --initial-virtual-time is a browser context setting.  It would be great if we could do that per WebContents but that's not currently possible.  See HeadlessContentBrowserClient::AppendExtraCommandLineSwitches.

lib/Page.js
```diff
@@ -492,19 +525,26 @@ class Page extends EventEmitter {
...
-    ]);
-    if (!error)
-      error = await navigationPromise;
-    watcher.cancel();
-    helper.removeEventListeners(eventListeners);
-    if (error)
-      throw error;
+    if (this._virtualTimeController) {
+      const timeout = typeof options.timeout === 'number' ? options.timeout : this._defaultNavigationVirtualTimeout;
+      const budgetExiredPromise = this._virtualTimeController.grantVirtualTimeBudget(timeout, VirtualTimeController.StartPolicy.WAIT_FOR_NAVIGATION);
```
aslushnikov: Can we have this as a part of the `NavigatorWatcher`? This way the `page.waitForNavigation()` api will work out-of-the box.

lib/VirtualTimeController.js
```diff
@@ -0,0 +1,353 @@
...
+          resolve();
+        }
+      }
+    }();
+  });
+};
+
+
+module.exports = VirtualTimeController;
+helper.tracePublicAPI(VirtualTimeController);
```
aslushnikov: `VirtualTimeController` is not exposed to the users, isn't it? In this case, the `helper.tracePublicAPI()` call is not needed.

lib/CompositorController.js
```diff
@@ -0,0 +1,401 @@
...
+      // TODO(alexclarke): this should really be promise rejection.
+      const callback = this._screenshotCapturedCallback;
+      this._screenshotCapturedCallback = null;
+      callback();
+    }
+  }
+}
+
+module.exports = CompositorController;
+helper.tracePublicAPI(CompositorController);
```
aslushnikov: This line is not needed since `CompositorController` is not exposed in the API

lib/Launcher.js
```diff
@@ -82,6 +82,25 @@ class Launcher {
       chromeArguments.push('--auto-open-devtools-for-tabs');
       options.headless = false;
     }
+    if (options.deterministic) {
+      if (options.deterministic.date)
+        chromeArguments.push('--initial-virtual-time=' + options.deterministic.date.getTime() / 1000);
+      options.deterministic.beginFrameControl = options.deterministic.beginFrameControl || true;
+      if (options.deterministic.beginFrameControl) {
```
aslushnikov: Can we get away with just two flags:
- `determenistic` that will assume `beginFrameControl` and determenistic fetch
- `determenisticTime` that will set the initial virtual time and will work only together with `determenistic` option

lib/Launcher.js
```diff
@@ -82,6 +82,25 @@ class Launcher {
       chromeArguments.push('--auto-open-devtools-for-tabs');
       options.headless = false;
     }
+    if (options.deterministic) {
```
aslushnikov, line 85: We should pass the same options to the `puppeteer.connect`: a common usecase is to launch browser in a container and connect to it remotely with `puppeteer.connect`.

It would be way better though if all these special flags could be enabled with protocol. Is this possible?

lib/Page.js
```diff
@@ -492,19 +525,26 @@ class Page extends EventEmitter {
...
-      navigationPromise,
-    ]);
-    if (!error)
-      error = await navigationPromise;
-    watcher.cancel();
-    helper.removeEventListeners(eventListeners);
-    if (error)
-      throw error;
+    if (this._virtualTimeController) {
+      const timeout = typeof options.timeout === 'number' ? options.timeout : this._defaultNavigationVirtualTimeout;
```
aslushnikov: can we use `defaultNavigationTimeout` instead of `defaultNavigationVirtualTimeout` here? This will keep API lean.

aslushnikov (commented):
I played with the patch, and it worked well with css animations on `https://aslushnikov.com`, but failed with gifs. I'd expect screenshots `a.jpg` and `b.jpg` to be the same, but they're different:

```js

(async() => {
    deterministic: {
      date: new Date('Jan 01, 2000')
    }
  });
  await page.goto('https://i.imgur.com/2OxhTSW.gif', {
    waitUntil: 'networkidle0'
  });
  await page.screenshot({path: 'a.jpg'});
  await page.reload({
    waitUntil: 'networkidle0'
  });
  await page.screenshot({path: 'b.jpg'});
  await browser.close();
})();
```

**EDIT:** I misspelled a `deterministic` flag; now the script times out for me.

lib/CompositorController.js
ADClarke: Done

lib/VirtualTimeController.js
ADClarke: Done

lib/Launcher.js
ADClarke: Done - apart from a couple of caveats:

1. It doesn't work with --deterministic-fetch.  I suppose in theory  HeadlessURLRequestContextGetter::GetURLRequestContext could clear |url_request_context_| and regenerate the net::URLRequestContext but I presume that's not safe...

2. If a Pupeteer script does Puppeteer.connect _AND_ uses Browser.pages() to try and render with an existing page then this won't work.  It should work fine with new pages however.

lib/Launcher.js
ADClarke: Done. I think overriding the virtual FPS with  options.deterministic.animationBeginFrameInterval can be pretty useful too.

lib/Page.js
ADClarke: 30 virtual seconds is way overkill so I guess I'll change _defaultNavigationTimeout's defalt based on whether or not we're using virtual time.

lib/Page.js
ADClarke: Done.

ADClarke (commented):
Using a local build of chrome (ToT + https://chromium-review.googlesource.com/c/chromium/src/+/937402) and your script (modified to launch my local build) the two gifs are the same.

lib/Launcher.js
aslushnikov: These are limitations that would be quite hard to debug; any chance we can assert for them?

aslushnikov:
> I think overriding the virtual FPS with options.deterministic.animationBeginFrameInterval can be pretty useful too.

I don't think I understand usecases atm, but maybe I'll figure if I play with the virtual time more.

> Using a local build of chrome (ToT + https://chromium-review.googlesource.com/c/chromium/src/+/937402) and your script (modified to launch my local build) the two gifs are the same.

Thanks; I'll be able to check once I get back from vacation on March, 10.

lib/Launcher.js
ADClarke: I've sent https://chromium-review.googlesource.com/c/chromium/src/+/939472 to add Browser.getCommandLine which lets throw an error if the expected command line flags are not set.

I've also added a flag into Puppeteer.Target to track if the target was created before or after the new Browser.deterministicMode has been set.  This is checked in Target.page() and if the browser was created in deterministic mode and we're trying to access an existing page then it will throw an error.

ADClarke:
Hey Andrey I've updated the pull request to work with the latest ToT chromium. It passes the CQ now:)

I'd like to write some tests for this feature, but considering it needs ToT chromium I'm expecting those would fail on the CQ because that (I assume) downloads a release build of chrome. It will be some time before our chromium patches make it to a release build.  What do you think we should do here?

aslushnikov:
> Hey Andrey I've updated the pull request to work with the latest ToT chromium. It passes the CQ now:)

@ADClarke nice! I'm happy the issues are resolved.

> I'd like to write some tests for this feature, but considering it needs ToT chromium I'm expecting those would fail on the CQ because that (I assume) downloads a release build of chrome.

We depend on a particular revision of chromium that we [roll from time to time](https://github.com/GoogleChrome/puppeteer/commit/28f88e8ef907a3216c13dc8deb7e042bd350ce10), see "puppeteer" field in our [package.json](https://github.com/GoogleChrome/puppeteer/blob/master/package.json).

We're releasing 1.2.0 tomorrow; we'll be able to roll chromium to the current latest after that, your code will work just fine.

ADClarke:
Ah interesting.  I added some code to check the version of chrome when using puppeter.connect, am I duplicating anything here?

ADClarke:
I've added some tests, PTAL :)

test/puppeteer.spec.js
```diff
@@ -255,6 +255,101 @@ module.exports.addTests = function(runner, expect, defaultBrowserOptions, puppet
         expect(fs.existsSync(executablePath)).toBe(true);
       });
     });
+    describe('DeterministicMode', function() {
+      it('should let us override the date', async({server}) => {
+        const browser = await puppeteer.launch({deterministic: {date: new Date('Jan 1, 1980')} });
```
aslushnikov: @ADClarke you should re-use `defaultBrowserOptions` so that this launches well on Ci's

ADClarke:
Thanks that fixes it.  I did find another bug with begin frame timestamps however.  We'll need to wait for https://chromium-review.googlesource.com/#/c/chromium/src/+/972985 to land and be used by puppeteer before we can proceed however.

aslushnikov:
Closing this for now since @pavelfeldman is re-working the protocol for virtual time.

lucemia:
https://chromium-review.googlesource.com/c/chromium/src/+/1056405 
It looks like the re-working is done, are we able to use virtual time with puppeteer now?

aslushnikov:
@lucemia you can use it already with raw protocol, using [`createCDPSession`](https://pptr.dev/#?product=Puppeteer&version=v1.8.0&show=api-targetcreatecdpsession). 

The reason we haven't exposed this yet is because there's no good debugging story ATM: virtual time doesn't work in headful.

PR #2066, opened 2018-02-20
feat: workers

JoelEinbinder:


docs/api.md
```diff
@@ -2586,6 +2593,14 @@ console.log(`Bytes used: ${usedBytes / totalBytes * 100}%`);
 > **NOTE** JavaScript Coverage doesn't include anonymous scripts. However, scripts with sourceURLs are
 reported.
 
+### class: Worker
+
+#### worker.executionContext()
```
ebidel, line 2598: One sentence description would be nice. I assume this covers service workers too? 

"Worker provides an API for interacting with the web and service workers".

ebidel, line 2598: Can you add this snippet:

```js
const worker = page.workers()[0];
const executionContext = await worker.executionContext();
await executionContext.evaluate(() => console.log('hello from worker!'));
```

JoelEinbinder: I’m holding off on writing the docs until everything has been sufficiently bikeshedded. I’m not sure how to handle service workers yet.

aslushnikov:
I'd expect page subtargets API to be the same as browser targets api, e.g.:

```js
page.targets() // returns list of all page subtargets without connecting to them
page.workers() // connects to all the page workers
```

aslushnikov:
I'll close this for now since it looks like there's much more work to be done in the protocol before we can safely implement workers.

PR #2071, opened 2018-02-21
fix(Launcher): dumpio bug

### LINKED ISSUE #2046: Using the dumpio setting throws an error in v1.1.0
### Steps to reproduce

* Puppeteer version: 1.1.0
* Platform / OS version: Ubuntu 16.04
* Node.js version: 6.13.0

Running this code:
```

puppeteer.launch({
  dumpio: true
}).then(function (browser) {
  // irrelevant code here
});
```

This should not cause an error.

**What happens instead?**
There is a thrown error:
```
(node: 23455) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: input.on is not a function
```
And the program exits.

**More notes**
Version 1.0.0 seems not to have this issue.

yanivefraim:
fixes #2046

### DIFF
--- a/test/fixtures/dumpio.js
+++ b/test/fixtures/dumpio.js
@@ -1,9 +1,9 @@
 (async() => {
-  const [, , puppeteerRoot, options, emptyPage] = process.argv;
+  const [, , puppeteerRoot, options, emptyPage, dumpioTextToLog] = process.argv;
   const browser = await require(puppeteerRoot).launch(JSON.parse(options));
   const page = await browser.newPage();
   await page.goto(emptyPage);
-  await page.evaluate(() => console.log('test'));
+  await page.evaluate(_dumpioTextToLog => console.log(_dumpioTextToLog), dumpioTextToLog);
   await page.close();
   await browser.close();
 })();
--- a/test/fixtures/dumpio.js
+++ b/test/fixtures/dumpio.js
@@ -1,0 +1,9 @@
+(async() => {
+  const [, , puppeteerRoot, options, emptyPage] = process.argv;
+  const browser = await require(puppeteerRoot).launch(JSON.parse(options));
+  const page = await browser.newPage();
+  await page.goto(emptyPage);
+  await page.evaluate(() => console.log('test'));
+  await page.close();
+  await browser.close();
+})();
--- a/test/test.js
+++ b/test/test.js
@@ -271,14 +271,14 @@ describe('Puppeteer', function() {
       const args = puppeteer.defaultArgs();
       expect(args).toContain('--no-first-run');
     });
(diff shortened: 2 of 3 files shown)


yanivefraim:
I would like to use polling with a tiny util called [eventually](https://github.com/wix/wix-eventually), in order to replace this [part](https://github.com/GoogleChrome/puppeteer/pull/2071/files#diff-c1129c8b045390789fa8ff62f2c6b4a9R280) (it will be much cleaner). Is it ok to install it as a devDependency?

test/fixtures/dumpio.js
```diff
@@ -0,0 +1,9 @@
+(async() => {
+  const [, , puppeteerRoot, options, emptyPage] = process.argv;
+  const browser = await require(puppeteerRoot).launch(JSON.parse(options));
+  const page = await browser.newPage();
+  await page.goto(emptyPage);
+  await page.evaluate(() => console.log('test'));
```
aslushnikov: nit: let's log `"MAGIC_DUMPIO_TEST"` to be distinct

test/test.js
```diff
@@ -271,6 +271,19 @@ describe('Puppeteer', function() {
...
+      const path = require('path');
+      const options = Object.assign({dumpio: true}, defaultBrowserOptions);
+      const res = spawn('node',
+          [path.resolve('test', 'fixtures', 'dumpio.js'), PROJECT_ROOT, JSON.stringify(options), server.EMPTY_PAGE]);
+      const response = await new Promise(resolve => res.stderr.on('data', data => {
+        if (data.toString('utf8').includes('test'))
+          resolve(data.toString('utf8'));
+      }));
+
+      expect(response).toContain('test');
```
aslushnikov: you should await for the node script to terminate

aslushnikov (approved):
Impressive test, thank you

aslushnikov:
> I would like to use polling with a tiny util called eventually, in order to replace this part (it will be much cleaner). Is it ok to install it as a devDependency?

Can we instead wait for the script to finish and check it's stderr after that?

test/fixtures/dumpio.js
yanivefraim: Sure, done...

test/test.js
yanivefraim: Done

yanivefraim:
> > I would like to use polling with a tiny util called eventually, in order to replace this part (it will be much cleaner). Is it ok to install it as a devDependency?
> 
> Can we instead wait for the script to finish and check it's stderr after that?

Yes, sure. It was too late at night for me (: 
Done (thanks)

PR #2073, opened 2018-02-22
fix(evaluate): throw error when page reloads during page.evaluate.

### LINKED ISSUE #2021: page.evaluate unpredictable when it triggers a reload
### Steps to reproduce

* Puppeteer version: 1.0.0
* Platform / OS version: Mac OS 10.13.3
* URLs (if applicable): 
* Node.js version: 9.5.0

I'm aware that this is nonsense code 🙂 , this is just to replicate the unpredictable behavior of `.evaluate`. Following up on #2020.

1. execute
```js

(async () => {
  try {
    await page.goto('https://www.google.com/');

    const result = await page.evaluate(function () {
      location.reload();
        setTimeout(() => resolve({}), 0);
      });
    });

    console.log('result', result);
  } catch (err) {
    console.log('error', err.message);
  } finally {
    await browser.close();
  }
})();
```

it prints
```
error <Some error here>
```
or 
```
result {}
```

**What happens instead?**

it prints
```
result undefined
```

Which seems like it introduces a third state to the result of `.evaluate`. Strangely enough, when you change the object by a string or number it does print the value.

```js
const result = await page.evaluate(function () {
  location.reload();
    setTimeout(() => resolve(1), 0);
  });
});
// result 1
```

Increasing the timeout on the promise results in an error 👍 

```js
const result = await page.evaluate(function () {
  location.reload();
    setTimeout(() => resolve({}), 1000);
  });
});
// error Protocol error (Runtime.callFunctionOn): Execution context was destroyed. undefined
```

Unless you return a primitive

```js
const result = await page.evaluate(function () {
  location.reload();
    setTimeout(() => resolve(1), 1000);
  });
});
// result 1
```

aslushnikov:
Fixes #2021.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -484,6 +484,15 @@ describe('Page', function() {
       const result = await page.evaluate(() => window);
       expect(result).toBe(undefined);
     });
+    it('should fail for circular object', async({page, server}) => {
+      const result = await page.evaluate(() => {
+        const a = {};
+        const b = {a};
+        a.b = b;
+        return a;
+      });
+      expect(result).toBe(undefined);
+    });
     it('should accept a string', async({page, server}) => {
       const result = await page.evaluate('1 + 2');
       expect(result).toBe(3);
--- a/test/test.js
+++ b/test/test.js
@@ -415,7 +415,7 @@ describe('Page', function() {
     });
     it('should throw when evaluation triggers reload', async({page, server}) => {
       let error = null;
-      const result = await page.evaluate(() => {
+      await page.evaluate(() => {
         location.reload();
         return new Promise(resolve => {
           setTimeout(() => resolve(1), 0);
--- a/test/test.js
+++ b/test/test.js
@@ -413,6 +413,16 @@ describe('Page', function() {
       const result = await page.evaluate(() => 7 * 3);
       expect(result).toBe(21);
     });
+    it('should throw when evaluation triggers reload', async({page, server}) => {
+      let error = null;
+      const result = await page.evaluate(() => {
+        location.reload();
+        return new Promise(resolve => {
+          setTimeout(() => resolve(1), 0);
+        });
+      }).catch(e => error = e);
+      expect(error.message).toContain('Protocol error');
+    });
     it('should await promise', async({page, server}) => {
       const result = await page.evaluate(() => Promise.resolve(8 * 7));
       expect(result).toBe(56);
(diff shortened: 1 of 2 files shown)


JoelEinbinder (approved):

test/test.js
```diff
@@ -474,6 +484,15 @@ describe('Page', function() {
...
       expect(result).toBe(undefined);
     });
+    it('should fail for circular object', async({page, server}) => {
+      const result = await page.evaluate(() => {
+        const a = {};
+        const b = {a};
+        a.b = b;
+        return a;
+      });
+      expect(result).toBe(undefined);
```
Janpot, line 494: @aslushnikov Is there any reason on why this should return `undefined`? I think throwing an error would be more appropriate. Think of the consumer of this API. They will not realise their result is circular, or it might happen only in some occasions. Or maybe sometimes `undefined` is a legit output of their function. How will they be able to distinguish these cases from a bug? How will they be notified of bugs in their code?

aslushnikov: @Janpot the motivation behind this was to let people use `page.evaluate` safely without caring on the return value. Consider the following simple script:

```js
await page.evaluate(() => window.foo = window);
```

Technically, the `page.evaluate` should throw since the expression returns non-serializable value,
but most of the time its unintentional.

> How will they be able to distinguish these cases from a bug? How will they be notified of bugs in their code?

I filed https://github.com/GoogleChrome/puppeteer/issues/2079 with a solution that will hopefully address your concerns and satisfy our usecase.

Janpot: @aslushnikov How did you arrive at this conclusion: _but most of the time its unintentional._?
In my experience it's exactly the opposite, I've never had to evaluate and not care about the result.

aslushnikov: > How did you arrive at this conclusion: but most of the time its unintentional.?

@Janpot once we started throwing errors, we instantly had a bunch of feedback on both bugtracker and internally about the unwanted errors caused by one-liners in `page.evaluate`.

Janpot: Ok, I don't really agree with API design like this but I see you want to be sort of pragmatic, and I don't have arguments against that that would actually change your mind 🙂. I'll settle for
```js
async function evaluate (page, ...args) {
  const result = await page.evaluate(...args);
  if (result === puppeteer.UNSERIALIZABLE_VALUE) {
    throw new Error('Unserializable value');
  }
  retun result;
}
```
then.

PR #2081, opened 2018-02-23
fix(Launcher): do not detach child process on windows

### LINKED ISSUE #2053: Allow Chromium process "detached" argument to be configurable?
Is there any special reason why Chromium process is spawned with "detached:true" argument? Can it be configurable? I have a situation on windows platform when i am creating child processes which are launching puppeteer instances. When i kill child processes, Chromium instances remain (on windows signal SIGTERM is not supported ... ). Is this a preferable solution or something else?

aslushnikov:
Fixes #2053.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -107,7 +107,10 @@ class Launcher {
         chromeExecutable,
         chromeArguments,
         {
-          detached: true,
+          // On non-windows platforms, `detached: false` makes child process a leader of a new
+          // process group, making it possible to kill child process tree with `.kill(-pid)` command.
+          // @see https://nodejs.org/api/child_process.html#child_process_options_detached
+          detached: process.platform !== 'win32',
           env: options.env || process.env,
           stdio
         }


aslushnikov:
@alixaxel @JoelEinbinder can you guys please check if this works for you on win? Process management is not a well-tested area of ours.

alixaxel:
@aslushnikov I'm not a windows user.

aslushnikov:
@alixaxel oops, sorry for the noize.
cc @vsemozhetbyt

vsemozhetbyt:
@aslushnikov Can you provide some code I could test on Windows to be sure this fix the issue?

aslushnikov:
@vsemozhetbyt Try running the following script and then terminate running node.js program differently:
- via taskmanager
- via Ctrl-C
- via closing the terminal

In all the cases, the spawned chromium instance should be closed. 

```js
var pptr = require('puppeteer');
(async() => {
  const browser = await pptr.launch();
  console.log('Launched!');
})();
```

vsemozhetbyt:
@aslushnikov I can confirm that with this patch all three terminations of the running node.js program also close the spawned chromium instance on Windows 7 x64 with the `cmd.exe` shell.

aslushnikov:
@vsemozhetbyt: Thank you.

@JoelEinbinder, let's merge this?

aslushnikov:
@vsemozhetbyt: Thank you.

@JoelEinbinder, let's merge this?

JoelEinbinder:
I still get some errors when closing things on windows, but it actually closes things now so I'd say this is better.

JoelEinbinder (approved):

PR #2090, opened 2018-02-23
feat(FrameManager): add type=module to addScriptTag

### LINKED ISSUE #2078: addScriptTag type=module
It would be awesome if puppeteer supported type=module for addScriptTag.

yanivefraim:
fixes #2078

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1708,7 +1708,7 @@ The method evaluates the XPath expression.
   - `url` <[string]> URL of a script to be added.
   - `path` <[string]> Path to the JavaScript file to be injected into frame. If `path` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
   - `content` <[string]> Raw JavaScript content to be injected into frame.
-  - `type` <[string]> Script type. To be used with `url`. Use 'module' in order to load a Javascript ES6 module. See [script](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) for more details.
+  - `type` <[string]> Script type. Use 'module' in order to load a Javascript ES6 module. See [script](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) for more details.
 - returns: <[Promise]<[ElementHandle]>> which resolves to the added tag when the script's onload fires or when the script content was injected into frame.
 
 Adds a `<script>` tag into the page with the desired url or content.
--- a/docs/api.md
+++ b/docs/api.md
@@ -655,8 +655,10 @@ Shortcut for [page.mainFrame().$x(expression)](#frameexpression)
 #### page.addScriptTag(options)
 - `options` <[Object]>
   - `url` <[string]> URL of a script to be added.
-  - `path` <[string]> Path to the JavaScript file to be injected into frame. If `path` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
+  - `path` <[string]> Path to the JavaScript file to be injected
+  into frame. If `path` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
   - `content` <[string]> Raw JavaScript content to be injected into frame.
+  - `type` <[string]> Script type. Use 'module' in order to load a Javascript ES6 module. See [script](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) for more details
 - returns: <[Promise]<[ElementHandle]>> which resolves to the added tag when the script's onload fires or when the script content was injected into frame.
 
 Adds a `<script>` tag into the page with the desired url or content.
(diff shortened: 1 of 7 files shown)


test/test.js
```diff
@@ -2922,6 +2922,14 @@ describe('Page', function() {
       expect(await page.evaluate(() => __injected)).toBe(42);
     });
 
+    it('should work with a url and type=module', async({page, server}) => {
+      await page.goto(server.EMPTY_PAGE);
+      const scriptHandle = await page.addScriptTag({ url: '/injectedfile.js', type: 'module' });
```
aslushnikov: can we check that `import` statement works in the injected url? e.g. via importing some script that yields `console.log`.

lib/FrameManager.js
```diff
@@ -464,9 +464,11 @@ class Frame {
      * @param {string} url
      * @return {!Promise<!HTMLElement>}
      */
-    async function addScriptUrl(url) {
+    async function addScriptUrl(url, type) {
```
ebidel: `type` needs a docstring.

lib/FrameManager.js
```diff
@@ -464,9 +464,11 @@ class Frame {
...
       const script = document.createElement('script');
       script.src = url;
+      if (type === 'module')
```
ebidel: should we pass through any type if the user passes one in? e.g. allow someone to inject a HB template: `<script type="text/x-handlebars-template"`.

test/test.js
yanivefraim: Yes, I started with that. But then eslint was failing because you must put import at the top (so I couldn't use it inside my test, in an evaluate statement). 
And then I thought: I don't want to test `type="module"` itself, I assume it will works, as long as I add type="module" to my script. WDYT?

lib/FrameManager.js
yanivefraim: Yes, you are right. Fixed. Thanks!

lib/FrameManager.js
yanivefraim: fixed. Thanks!

docs/api.md
```diff
@@ -657,6 +657,7 @@ Shortcut for [page.mainFrame().$x(expression)](#frameexpression)
   - `url` <[string]> URL of a script to be added.
   - `path` <[string]> Path to the JavaScript file to be injected into frame. If `path` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
   - `content` <[string]> Raw JavaScript content to be injected into frame.
+  - `type` <[string]> Script type. To be used with `url`. Use 'module' in order to load a Javascript ES6 module. See [script](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) for more details
```
vsemozhetbyt: A nit: missing period after "more details".

docs/api.md
```diff
@@ -1707,6 +1708,7 @@ The method evaluates the XPath expression.
...
```
vsemozhetbyt: A nit: missing period after "more details".

yanivefraim: fixed. Thanks

docs/api.md
yanivefraim: fixed, thanks!

test/test.js
yanivefraim: One more point - user can add more kind of types (I fixed that). So this feature is not only about `type=module` (:

aslushnikov: > And then I thought: I don't want to test type="module" itself, I assume it will works, as long as I add type="module" to my script. WDYT?

Well, there's a claim in #2078 that `type="module"` doesn't work this way, so a test would be a good sanity check.

> But then eslint was failing because you must put import at the top (so I couldn't use it inside my test, in an evaluate statement).

Ah, you can create a test asset `assets/es6import.js` that uses modules. This should make eslint happy.

yanivefraim: yes, so after playing with `type="module"`, it seems that es6 modules with `type="module"` and chromium doesn't work. I wrote a test which failed for not recognizing es6 syntax. I then tried running [this](https://stackoverflow.com/a/45745913/916450) code snippet on Canary / Chromium and got the same error.

aslushnikov: @yanivefraim that's weird. I'm pretty sure chromium supports es6 modules: https://caniuse.com/#feat=es6-module

Do you have time to debug this further?

ebidel: Modules definitely work. I'm using them in several projects:)

@yanivefraim can you share the test case? We can take over this PR if you don't have time. There use case is a neat one (sharing code with node). I can see other cases like universal js with pptr running on a server and injecting modules into a page at serve time.

yanivefraim: @aslushnikov @ebidel  - I will love to give it a try... (didn't have much time last two weeks but now I will find time. if this is ok with you)

ebidel: Yea go for it. This is all you if you want it :)

yanivefraim: @ebidel @aslushnikov - so apparently It was my mistake: I forgot to add `.js` in my import (this is what happens when working on PRs at 2am...)
Fixed my test to actually be an e2e and test that modules are working.

yanivefraim:
** @aslushnikov @ebidel - Test was fixed: e2e test module imports

docs/api.md
```diff
@@ -1707,6 +1708,7 @@ The method evaluates the XPath expression.
   - `url` <[string]> URL of a script to be added.
   - `path` <[string]> Path to the JavaScript file to be injected into frame. If `path` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
   - `content` <[string]> Raw JavaScript content to be injected into frame.
+  - `type` <[string]> Script type. To be used with `url`. Use 'module' in order to load a Javascript ES6 module. See [script](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) for more details.
```
aslushnikov: @yanivefraim shouldn't it also work with `path` and `content` options?

yanivefraim: @aslushnikov - you are right! added support for `path` and `content`. Thanks!

yanivefraim:
@aslushnikov - added support for path and content. Thanks!

aslushnikov (approved):

PR #2091, opened 2018-02-23
test: fix Matcher.toEqual to not depend on key insertion order

aslushnikov:
Objects `{foo: 1, bar: 2}` and `{bar: 2, foo: 1}` should be considered
equal.


JoelEinbinder (commented):
Did you hit this somewhere? Key order is detectable, so maybe we should leave it as is.

JoelEinbinder (approved):

PR #2094, opened 2018-02-24
feat(ElementHandle): implement ElementHandle.contentFrame()

### LINKED ISSUE #433: No way to associate Frame object with the <iframe> element it represents

aslushnikov:
This patch adds ElementHandle.contentFrame() method that allows
to get a reference to the Frame owned by the iframe.

Fixes #433.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -187,6 +187,7 @@
   * [elementHandle.asElement()](#elementhandleaselement)
   * [elementHandle.boundingBox()](#elementhandleboundingbox)
   * [elementHandle.click([options])](#elementhandleclickoptions)
+  * [elementHandle.contentFrame()](#elementhandlecontentframe)
   * [elementHandle.dispose()](#elementhandledispose)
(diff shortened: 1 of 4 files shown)


JoelEinbinder (approved):

lems111:
guys - it's difficult to express over words here. It's such a coincidence that last night, I was up until late hours, trying to find this functionality until I found the conversation and pull request for it. In a similar fashion to a Christmas day, I woke this morning to find this merged to master and I was able to get latest in my app. I can confirm that this, indeed, is the functionality that I needed to cut my time in half for scraping data from a site. I no longer have to take an extra step to visit a new page; instead, I can look at data in the embedded iframe that I wasn't able to access without this. Thanks everyone!

PR #2096, opened 2018-02-24
docs(api.md): add link to the v1.1.1

aslushnikov:


JoelEinbinder (approved):
Maybe we should say **is expected to be release** instead of being so sure about the future.

PR #2102, opened 2018-02-26
fix(Launcher): killing a process can throw

thorn0:
... if the target pid does not exist.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -175,10 +175,14 @@ class Launcher {
       helper.removeEventListeners(listeners);
       if (chromeProcess.pid && !chromeProcess.killed && !chromeClosed) {
         // Force kill chrome.
-        if (process.platform === 'win32')
-          childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T /F`);
-        else
-          process.kill(-chromeProcess.pid, 'SIGKILL');
+        try {
+          if (process.platform === 'win32')
+            childProcess.execSync(`taskkill /pid ${chromeProcess.pid} /T /F`);
+          else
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):

aslushnikov:
@thorn0 can you please make `npm run lint` happy? I think you're missing space after `catch`.

thorn0:
@aslushnikov done

aslushnikov:
I was initially positive about this change, but now I'm scared we might be killing some other process. Looks like we're covering the actual problem instead of solving it properly. 

@JoelEinbinder, what do you think about this patch?

JoelEinbinder:
> I was initially positive about this change, but now I'm scared we might be killing some other process. Looks like we're covering the actual problem instead of solving it properly.

That seems like a valid concern. Do we get a disconnected event synchronously on our websocket? Maybe we can use that to avoid killing someone else's process with the same id. It is weird that we don't get a `'close'` event on the process in time.

thorn0:
At least on Windows, the exception occurs only when there is no process with the specified PID. That's the situation the proposed change fixes. I easily reproduce it killing Chromium manually.

If, as you're afraid, it kills some other process with the same reused PID, it won't throw. So I don't see how this change "covers the actual problem".

aslushnikov:
> At least on Windows, the exception occurs only when there is no process with the specified PID. That's the situation the proposed change fixes. I easily reproduce it killing Chromium manually.

> If, as you're afraid, it kills some other process with the same reused PID, it won't throw. So I don't see how this change "covers the actual problem".

I'm afraid we might be killing some other process, if the pid of the chrome process gets re-used after its manual termination. However, since this is highly unlikely and there's no easy way around, I'll merge this in.

PR #2106, opened 2018-02-27
fix: force layout in bounding box

JoelEinbinder:
Some tests were flaky because `DOM.getBoxModel` doesn't force a layout. Maybe it should on the backend?


aslushnikov:
@JoelEinbinder That’s a nice catch. Yes, getboxmodel has to force layout on the backend. Let’s fix upstream

aslushnikov:
@JoelEinbinder is this indeed the case and requires fixing in the protocol?

JoelEinbinder:
Fixed upstream

PR #2109, opened 2018-02-27
fix(test): add missing test for making sure chrome is closed

JoelEinbinder:
This should give some peace of mind that Puppeteer isn't leaving zombie Chromium processes everywhere as we make changes to the closing code.

### DIFF
--- a/test/fixtures/closeme.js
+++ b/test/fixtures/closeme.js
@@ -1,0 +1,5 @@
+(async() => {
+  const [, , puppeteerRoot, options] = process.argv;
+  const browser = await require(puppeteerRoot).launch(JSON.parse(options));
+  console.log(browser.wsEndpoint());
+})();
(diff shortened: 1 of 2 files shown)


aslushnikov (commented):
I’m not sure I understand how this test relates to zombies. I’d expect some `ps aux | grep Chromium` or similar

JoelEinbinder:
It's not a comprehensive test. It just makes sure that Chromium shuts down when node is killed. I avoid `ps aux` by have a second outside connection to the Chromium and listening for it to disconnect.

aslushnikov:
> I avoid ps aux by have a second outside connection to the Chromium and listening for it to disconnect.

I think there's already a test that does this (the one for the `browser.disconnected` event to be delivered). Do you still think this test adds value?

JoelEinbinder:
This doesn't test that we can close Chromium, but tests that Chromium closes when its parent node process is killed. I think its valuable by itself, and also as a template for other closing scenarios.

aslushnikov (approved):
I see; thanks for explanation.

Issue #1361, opened 2017-11-11
 Any way to check iframe loaded successfully??


fran0254:
**yes work**
```javascript
await page.waitFor(2000); 
const frame = await page.frames().find(f => f.name() === 'ifrw');
const button = await frame.$('#selector');
button.click(); 
```
**don´t work**
```javascript
// Was trying to select a iframe before it is actually loaded.
// Is possible check before, or wait for it to load ??
const frame = await page.frames().find(f => f.name() === 'ifrw');
const button = await frame.$('#selector');
button.click(); 
```

ebidel:
If you're using `page.goto()`, in that waits for the window `load` event to fire. So in theory, any frames on the page should also be loaded by that point in time.

If that's not working, you can use `waitForSelector`:

```js
const frame = await page.frames().find(f => f.name() === 'ifrw');
await frame.waitForSelector('#selector');
const button = await frame.$('#selector');
await button.click();
```

Note: this will be easier with https://github.com/GoogleChrome/puppeteer/issues/1007, tracks `waitForSelector` returning `Promise<ElementHandle>` instead of `Promise<undefined>`.

fran0254:
I'm not using page.goto(), the frame is in a modal window within the same page. The way you indicated me has not worked for me

aslushnikov:
@fran0254 there are frame events that notify about frame structure changes:

- [`frameattached`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#event-frameattached)
- [`framenavigated`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#event-framenavigated)
- [`framedetached`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#event-framedetached)

You can wait until there's a desired frame attached with `frameattached` event.
Let me know if it doesn't work for you.

fran0254:
@aslushnikov  Could you explain how I can check these three events that you mention to me?

**attached the frame**

``` html
Frame {
  _client: 
   Session {
     domain: null,
     _events: 
      { 'Page.frameAttached': [Function],
        'Page.frameNavigated': [Function],
        'Page.frameDetached': [Function],
        'Runtime.executionContextCreated': [Function],
        'Network.requestWillBeSent': [Function: bound _onRequestWillBeSent],
        'Network.requestIntercepted': [Function: bound _onRequestIntercepted],
        'Network.responseReceived': [Function: bound _onResponseReceived],
[...]

PR #2110, opened 2018-02-27
 docs(api.md): iframe load event

### LINKED ISSUE #1413: Docs: clarify how to know when a frame is loaded.
https://github.com/GoogleChrome/puppeteer/issues/1361 asked how to determine if a frame is loaded. 

@aslushnikov posted a [suggestion](https://github.com/GoogleChrome/puppeteer/issues/1361#issuecomment-343748051), but it's unclear if that addresses "frame is loaded". 

We should clarify in the docs whether `frameattached` or `framenavigated` are guaranteed to happen after `window.onload`. e.g. where do those events fall in relation to the frames `load` event and when using pptr's `waitUntil` timings?

yanivefraim:
Fixes #1413 

After playing around with frames, I figured out that execution order is:
1. `frameattached`
2. `framenavigated`
3.  Frame's `window.onLoad`

This means that the two steps are needed:
1. Wait for Frame to be attached
2. Wait for Frame to be loaded

I added a section in `Frame` docs: `Waiting for a frame to be loaded`, explaining this concept with the example from #1361.

I think that we might need a better built-in solution for that,  something like: `page.waitForFrameAttached(frameName)`

WDYT?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1637,29 +1637,29 @@ Waiting for a frame to be loaded is done in two steps:
 1. Wait for Frame to be attached - needed because looking for a specific frame on a page (using [`page.frames()`](#pageframes)) might be too early, before the frame is actually attached.
 2. Wait for Frame to be loaded - needed because both `frameattached` and `framenavigated` events will happen before frame's `window.onLoad` event.
 
-  An example for such implementation:
+    An example for such implementation:
 
-  ```js
-  function waitForFrame(page) {
-    let fulfill;
-    const promise = new Promise(x => fulfill = x);
-    checkFrame();
-    return promise;
+    ```js
+    function waitForFrame(page) {
+      let fulfill;
+      const promise = new Promise(x => fulfill = x);
+      checkFrame();
+      return promise;
 
-    function checkFrame() {
-      const frame = page.frames().find(f => f.name() === 'ifrw');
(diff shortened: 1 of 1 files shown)


yanivefraim:
Closing in favor of a `page.waitForFrameAttached(frameName)` PR

gonzaloserrano:
Hi, tried to use `page.waitForFrameAttached(1)` but got `TypeError: page.waitForFrameAttached is not a function`. Any hints?

yanivefraim:
Hi... there is no such method. I suggested to add such a method, but did not do it, yet (not sure if this is something we want to have)

sean-hill:
@yanivefraim I'm facing an issue where this would be helpful. I have a retarded page I have to write tests against that has a DOM structure like this:

image

If I call `page.childFrames()` I get a list of 5 frames, and if for each of those frames I call `frame.name()` immediately after the page loads, they all return empty strings. However if I wait for an arbitrary amount of time until all the frames have loaded, I get back the frame names for each child. So it'd be nice to have some function to wait for a frame to be loaded. 

```js
const childFrames = await page.childFrames()
await Promise.all(childFrames.map(frame => frame.waitForLoad()))
const childFrameNames = childFrames.map(frame => frame.name)
```
Just a suggestion.

PR #2121, opened 2018-02-28
 feat(Browser): add url param to newPage, default value is 'about:blank'

Rhymen:


Rhymen:
I signed it!

JoelEinbinder:
Thanks for the PR. Unfortunately this is a more complicated change that it might appear to be at first. Determining when a page has finished navigating to a URL is a difficult question. We have a bunch of [options](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagegotourl-options) in all of our navigation methods to deal with this.

Currently in your patch, the initial navigation is not awaited. This means that the page could be in a volatile state when it is returned from newPage. If we add the navigation options, a new problem arises. The page exists, but the user wont be notified of its existence until after the first navigation.

Leaving things as is, with `about:blank` as the default page, keeps our API and implementation simple.

aslushnikov:
Yes, adding custom url to the "newPage" will require adding all the navigation options, e.g. `waitUntil`. I'd rather hold off from that; the `page.newPage()`, followed by `page.goto()` seems good enough.

PR #2122, opened 2018-02-28
Add's page.removeListener to Page class examples

joelgriffith:
Just so folks are clear that you can un-subscribe from page events

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -508,6 +508,17 @@ This example logs a message for a single page `load` event:
 page.once('load', () => console.log('Page loaded!'));
 ```
 
+To unsubscribe from events use the `removeListener` method:
+
+```js
+function onLoad() {
+  console.log('Page loaded!');
+}
+page.on('load', onLoad);
+// Sometime later...
+page.removeListener('load', onLoad);
+```
+
 #### event: 'console'
 - <[ConsoleMessage]>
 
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -508,6 +508,17 @@ This example logs a message for a single page `load` event:
 page.once('load', () => console.log('Page loaded!'));
 ```
 
+To unsubscribe from events use the `removeListener` method:
+
+```js
+function onLoad() {
+  console.log('Page loaded!');
+}
+page.on('load', onLoad);
```
ebidel: This is fine, but it might be more clear if we use an invent that fires more than once. `page.on('console'...`?

docs/api.md
```diff
@@ -501,13 +501,24 @@ puppeteer.launch().then(async browser => {
...
 
 This example logs a message for a single page `load` event:
 ```js
 page.once('load', () => console.log('Page loaded!'));
 ```
 
+To unsubscribe from events use the `removeListener` method:
+
+```js
+function logRequest() {
```
ebidel: Include the `request` param. Sorry to nit pic.

ebidel (approved):

docs/api.md
joelgriffith: No worries, would rather be more thorough

PR #2133, opened 2018-03-02
How to execute example js in the README.md

if-kyle:
Hello,

I am very new to node/npm and it took me a bit to figure out how execute/run the examples in the README.  Likely obvious to those that run node all day, but hopefully this saves some time to other newbies.


ebidel (approved):
I'm ok with adding this extra info if it's helpful for new comers. I think it's come up a few times already as people come into the project from a non-Node world :)

aslushnikov (approved):

PR #2152, opened 2018-03-07
fix: response.ok should be true for file:// urls

### LINKED ISSUE #2116: page.goto with file:/// URI returns non-ok response
### Steps to reproduce

* Puppeteer version: 1.1.0
* Platform / OS version: Windows 10
* URLs (if applicable):
* Node.js version: 9.6.1

```js

let src = `file://${process.cwd()}/examples/basic.built.html`; // can be a file or http url

(async () => {
  const response = await page.goto(src);
  console.log(response.ok(), response.status());
  console.log(await response.text());
  browser.close();
})();
```

`response.ok()` should be `true`.

**What happens instead?**

`response.ok()` is `false` even when it gets a valid response from file system.

JoelEinbinder:
I tried to add a test, but we don't have a good way to load `file://` urls cross platform. I also found out that our type annotations for request and response were wrong.

Fixes #2116

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -538,7 +538,7 @@ class Response {
    * @return {boolean}
    */
   ok() {
-    return this._status >= 200 && this._status <= 299;
+    return this._status === 0 || (this._status >= 200 && this._status <= 299);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


aslushnikov (approved):

aslushnikov:
> I tried to add a test, but we don't have a good way to load file://

@JoelEinbinder  Out of curiosity, what was the issue? The `path -> url` translation doesn't look very involved.

JoelEinbinder:
The translation was a bit complicated to do right because its different for windows/posix and there could be special characters in the path to the puppeteer directory.

PR #2178, opened 2018-03-12
Cirrus CI for Windows Builds

fkorotkov:
Cirrus CI recently added support for Windows Docker containers. This change configures Cirrus CI so Puppeteer CI builds for Windows platform are executed in a controlled dockerized environment. 

Cirrus CI has a slightly faster build start time than the current CI thanks to light Docker containers. Build times itself are also comparable with the current VM-based CI. Another advantage could be in future to unify CI builds for Linux and Windows platforms since Cirrus CI supports both of them. 

Here is [a Cirrus CI build for this change](https://cirrus-ci.com/build/5645256259272704): 

image

I was testing Windows Docker support on this repository and decided to share the results. Feel free to close the PR if not interested in a better CI. Didn't want to waste my work 😅

In case you will merge the PR. Don't forget to [install Cirrus CI application from Github Marketplace](https://github.com/apps/cirrus-ci) as described in [Quick Start guide](https://cirrus-ci.org/guide/quick-start/).

PR #2181, opened 2018-03-12
Cirrus CI

fkorotkov:
As opposite to #2178, this change configures Cirrus CI to run Puppeteer in both Windows and Linux containers.

Here is a link to a [corresponding Cirrus CI build](https://cirrus-ci.com/build/5079208896233472):

image

As you can see there are 4 auto-generate Docker Builder tasks to build custom Linux and Windows Docker containers. And 4 tasks to actually run tests in the cached containers. If interested read more about Cirrus CI's container builds in [this blog post](https://medium.com/cirruslabs/introducing-container-builder-for-cirrus-ci-80b9234f007).

### DIFF
--- a/.ci/node7/Dockerfile.linux
+++ b/.ci/node7/Dockerfile.linux
@@ -1,0 +1,9 @@
+FROM node:7.10.1
+
+RUN apt-get update && \
+    apt-get -y install xvfb gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 \
+      libdbus-1-3 libexpat1 libfontconfig1 libgcc1 libgconf-2-4 libgdk-pixbuf2.0-0 libglib2.0-0 \
+      libgtk-3-0 libnspr4 libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 \
+      libxcomposite1 libxcursor1 libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 \
+      libxtst6 ca-certificates fonts-liberation libappindicator1 libnss3 lsb-release xdg-utils wget && \
+    rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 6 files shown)


aslushnikov:
@fkorotkov interesting! Are there any re-tries on Cirrus CI? I wonder how come our builds always flake on travis/appveyor and are green on Cirrus.

fkorotkov:
@aslushnikov there are no re-tries on Cirrus CI. I've seen some tests failing with timeouts during testing. Maybe just once of twice.

The only possible explanation I can come up with is that the environment that tests are executed is more performant. Travis executes builds on AWS VMs that run not the latest Docker. AppVeyour uses not the latest Windows VMs on Azure.

Cirrus on the other hand uses Google Cloud and specifically Kubernetes Engine to run Linux-based container builds. Kubernetes nodes that run containers are using the latest Docker and by the way Chromium OS 😅 For Windows containers, at the moment, Cirrus CI uses Compute Engine to run the latest Windows Server optimized for Docker container. We are planning to move Windows container to GKE once GKE supports Windows.

So since Cirrus CI builds are executed on Google Cloud Platform, maybe you can tell me why tests are more stable? 🤣

.ci/node7/Dockerfile.linux
```diff
@@ -0,0 +1,9 @@
+FROM node:7.10.1
+
+RUN apt-get update && \
```
aslushnikov, line 3: Nice! This means we now have a docker image that's a part of CI that is guaranteed to work with pptr.
This will be very helpful for everybody who runs puppeteer in a docker

cc @ebidel

aslushnikov (approved):

.ci/node7/Dockerfile.linux
fkorotkov: If you are interested we can add [Container Builder](https://medium.com/cirruslabs/introducing-container-builder-for-cirrus-ci-80b9234f007) later on to publish such official images to Docker Hub the [same way we do it for Flutter](https://github.com/cirruslabs/docker-images-flutter).

aslushnikov: @fkorotkov this would be great. Let's give cirrus some time and see how it goes. If it proves to be reliable, we'll be happy to publish puppeteer-compatible docker images.

PR #2194, opened 2018-03-14
feat(Network): implement request.redirectChain() api

### LINKED ISSUE #1816: Expose redirect chain for requests
The `request.redirectChain()` should return all the requests that are a part of the redirect chain for the `request`.

```js
const response = await page.goto('http://google.com');
const requests = response.request().redirectChain();
for (const request of requests)
  console.log(request.url());
```

This was requested a few times, e.g. https://github.com/GoogleChrome/puppeteer/issues/1708.

aslushnikov:
The request.redirectChain() returns all the requests issued to fetch
the resource.

Fixes #1816.

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -154,7 +154,7 @@ class NetworkManager extends EventEmitter {
       const request = this._interceptionIdToRequest.get(event.interceptionId);
       if (request) {
         this._handleRequestRedirect(request, event.responseStatusCode, event.responseHeaders, false /* fromDiskCache */, false /* fromServiceWorker */, null /* securityDetails */);
-        this._handleRequestStart(request._requestId, event.interceptionId, event.redirectUrl, event.resourceType, event.request, event.frameId);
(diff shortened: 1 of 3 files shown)


JoelEinbinder:
`Response.fromServiceWorker` failed on cirrus. Is that related to this patch?

lib/NetworkManager.js
```diff
@@ -377,6 +384,13 @@ class Request {
     return this._frame;
   }
 
+  /**
+   * @return {!Array<!Request>}
+   */
+  redirectChain() {
+    return this._redirectChain;
```
JoelEinbinder: It looks like we are returning a live array that might change after the user receives it. Slice here?

aslushnikov:
@JoelEinbinder i don't think so. I think I fix flakiness in https://github.com/GoogleChrome/puppeteer/pull/2195 (btw, ptal)

aslushnikov:
@JoelEinbinder addressed comments; ptal.

JoelEinbinder (approved):

PR #2198, opened 2018-03-15
fix(node6): fix one line await arrow functions

JoelEinbinder:
When the start of the function body was `await`, the async function transformer could break.

### DIFF
--- a/utils/node6-transform/TransformAsyncFunctions.js
+++ b/utils/node6-transform/TransformAsyncFunctions.js
@@ -52,18 +52,26 @@ const asyncToGenerator = fn => {
  * @return {string}
  */
 function transformAsyncFunctions(text) {
+  /**
+   * @type {!Array<{from: number, to: number, replacement: string}>}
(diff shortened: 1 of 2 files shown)


utils/node6-transform/TransformAsyncFunctions.js
```diff
@@ -63,7 +66,15 @@ function transformAsyncFunctions(text) {
...
   walker.walk(ast);
 
-  edits.sort((a, b) => b.from - a.from);
+  edits.sort((a, b) => {
+    const diff = b.from - a.from;
+    if (diff) return diff;
+    if (a.isOpening !== b.isOpening)
+      return a.isOpening ? 1 : -1;
+    if (a.isOpening)
+      return edits.indexOf(a) - edits.indexOf(b);
```
aslushnikov: this makes the sort O(N^2*logN) - do you really want this?

aslushnikov (approved):

PR #2201, opened 2018-03-15
test: another attempt to fix sw test

aslushnikov:


test/assets/serviceworkers/fetch/sw.html
```diff
@@ -1,4 +1,5 @@
 <link rel="stylesheet" href="./style.css">
 <script>
   window.registrationPromise = navigator.serviceWorker.register('sw.js');
```
JoelEinbinder, line 3: There is a second test that uses this pattern, and we dont need window.registrationPromise anymore.

aslushnikov: It's needed to unregister!

JoelEinbinder (approved):

PR #2218, opened 2018-03-17
Health: Break 'page.spec.js' to smaller files

yanivefraim:
Started breaking page.spec.js to some small files, by subject.


test/browser.spec.js
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+module.exports.addTests = function({
+  describe, xdescribe, fdescribe, it, fit, xit, beforeAll, beforeEach, afterAll, afterEach
```
aslushnikov: nit: let's pass testrunner here and extract the DSL below. I think it'll look better.

test/coverage.spec.js
```diff
@@ -0,0 +1,37 @@
...
+ * limitations under the License.
+ */
+
+const {helper} = require('../lib/helper');
+
+module.exports.addTests = function({
+  describe, xdescribe, fdescribe, it, fit, xit, beforeAll, beforeEach, afterAll, afterEach
+}, expect, defaultBrowserOptions, puppeteer, PROJECT_ROOT) {
+
+  if (process.env.COVERAGE) {
```
aslushnikov: I'd move this into the `test.js`; it's an infrastructure thing and tightly depends on `helper.recordPublicAPICoverage()` call.

test/page.spec.js
```diff
@@ -46,58 +46,29 @@ module.exports.addTests = function(runner, expect, defaultBrowserOptions, puppet
...
+
+    testFiles
+        .map(file => path.join(__dirname, file))
+        .forEach(file =>
+          require(file).addTests(
+              {
+                describe, xdescribe, fdescribe, it, fit, xit, beforeAll, beforeEach, afterAll, afterEach
+              },
+              expect,
+              defaultBrowserOptions,
```
aslushnikov: I think `defaultBrowserOptions` and `PROJECT_ROOT` are not needed for page subtests? Not sure what's the use of `puppeteer` as well.

test/test-driver.js
```diff
@@ -0,0 +1,69 @@
...
+ * limitations under the License.
+ */
+
+/**
+   * @param {!EventEmitter} emitter
+   * @param {string} eventName
+   * @param {number=} eventCount
+   * @return {!Promise<!Object>}
+   */
+module.exports.waitForEvents = function(emitter, eventName, eventCount = 1) {
```
aslushnikov: let's call this file `utils.js` and merge with `frame-utils.js`.

test/browser.spec.js
yanivefraim: done (:

test/coverage.spec.js
yanivefraim: done

test/page.spec.js
yanivefraim: we do use some of them. I removed unnecessary params...

test/test-driver.js
yanivefraim: done. Side note: I like the name `driver` better than `utils`  😃

yanivefraim:
@aslushnikov - thanks for the review! 

All comments fixed. `puppeteer`, `PROJECT_ROOT` and `defaultBrowserOptions` were still needed in some of the tests. I changed `addTests` params to be an object, so we can consume only what is needed.

BTW - `page.spec.js` is now ~2400 lines, instead of ~4100 in old `test.js`. It is still way too much, but we are in the right direction, I think...

test/utils.js
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2017 Google Inc. All rights reserved.
+ * Copyright 2018 Google Inc. All rights reserved.
```
aslushnikov: why's the change?

test/test-driver.js
aslushnikov: Thanks. `driver`, and especially `test-driver`, raises very strong associations in my head with `webdriver` and `chromedriver`

aslushnikov:
> All comments fixed. puppeteer, PROJECT_ROOT and defaultBrowserOptions were still needed in some of the tests. I changed addTests params to be an object, so we can consume only what is needed.

@yanivefraim These look nice

> BTW - page.spec.js is now ~2400 lines, instead of ~4100 in old test.js. It is still way too much, but we are in the right direction, I think...

Definitely.

test/utils.js
yanivefraim: fixed back to 2017

test/test-driver.js
yanivefraim: This is a long discussion 😄 

We don't like the name `utils`, it reminds us of a place you put stuff that you don't know what to do with. On the other hand, we use drivers to put common test stuff + encapsulate test logic / selectors. This way our tests are much more readable (they are written in "english") + they are not aware of selectors / dom implementation. If we change dom implementation, we will change selectors only in one place (and not in all tests). WDYT?

aslushnikov (approved):

PR #2246, opened 2018-03-22
Proposal: flexible chromium revisions

Glennvd:
Often we want to test changes on new chromium versions. At the moment we would have to switch from the default downloader to specifying our own executable and downloading that etc. 

Exposing this to an environment variable would make it easy to test against different chromium versions in CI environments.

### DIFF
--- a/install.js
+++ b/install.js
@@ -30,7 +30,7 @@ const downloadHost = process.env.PUPPETEER_DOWNLOAD_HOST || process.env.npm_conf
 const puppeteer = require('./index');
 const browserFetcher = puppeteer.createBrowserFetcher({ host: downloadHost });
 
-const revision = require('./package.json').puppeteer.chromium_revision;
+const revision = process.env.PUPPETEER_CHROMIUM_REVISION || require('./package.json').puppeteer.chromium_revision;
 const revisionInfo = browserFetcher.revisionInfo(revision);
 
 // Do nothing if the revision is already downloaded.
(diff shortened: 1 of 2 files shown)


Glennvd:
CLA is signed under Social Significance LTD

Glennvd:
CLA is signed under Social Significance LTD

install.js
```diff
@@ -30,7 +30,7 @@ const downloadHost = process.env.PUPPETEER_DOWNLOAD_HOST || process.env.npm_conf
 const puppeteer = require('./index');
 const browserFetcher = puppeteer.createBrowserFetcher({ host: downloadHost });
 
-const revision = require('./package.json').puppeteer.chromium_revision;
+const revision = process.env.PUPPETEER_CHROMIUM_REVISION || require('./package.json').puppeteer.chromium_revision;
```
ebidel, line 33: Note, we do have a `BrowserFetcher` API that serves this purpose:
https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#class-browserfetcher

But this one liner is a nicer than writing some code. @aslushnikov wdyt?

We'd need to also document this env var in https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#environment-variables.

Glennvd: I've added the documentation as well FYI

aslushnikov (approved):

PR #2256, opened 2018-03-25
feat(elementHandle): add boxModel method

### LINKED ISSUE #1357: Feature: improve boundingBox method at ElementHandle
Current version of `boundingBox` is terribly coded as I think 😄 :
https://github.com/GoogleChrome/puppeteer/blob/master/lib/ElementHandle.js#L123

Look at DevTools method:
https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-getBoxModel

DevTools method returns `BoxModel`:
https://chromedevtools.github.io/devtools-protocol/tot/DOM/#type-BoxModel

`BoxModel` contains:
* content
* padding
* border
* margin
* width
* height

1) Why we are calculating width and height? It's already calculated and returned by DevTools, but we have:
```js
const width = Math.max(quad[0], quad[2], quad[4], quad[6]) - x;
const height = Math.max(quad[1], quad[3], quad[5], quad[7]) - y;
```
2) Puppeteer use only border position:
```js
const quad = result.model.border;
```
What If I want to get content position? Or padding position or margin position? 😢 

My advice is smth like that:
```js
async boundingBox() {
  const result = await this._client.send('DOM.getBoxModel', {
    objectId: this._remoteObject.objectId
  }).catch(error => void debugError(error));

  if (!result)
    return null;

  const borderQuad = result.model.border;

  return {
    border: {
      x: Math.min(borderQuad[0], borderQuad[2], borderQuad[4], borderQuad[6]),
      y: Math.min(borderQuad[1], borderQuad[3], borderQuad[5], borderQuad[7]),
      width: Math.max(borderQuad[0], borderQuad[2], borderQuad[4], borderQuad[6]) - x,
      height: Math.max(borderQuad[1], borderQuad[3], borderQuad[5], borderQuad[7]) - y
    },
    // same for padding and margin
    width: result.model.width,
    height: result.model.height
  };
}
```

yanivefraim:
Fixes #1357 - expose element's boxModel. 

I wasn't sure how exactly do we want to represent `boxmodel` data, so I just represented it "as is": an object containing [quads](https://chromedevtools.github.io/devtools-protocol/tot/DOM#type-Quad), exactly as it is represented in [chromedevtools](https://chromedevtools.github.io/devtools-protocol/tot/DOM#type-BoxModel)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2206,7 +2206,7 @@ This method returns the bounding box of the element (relative to the main frame)
   - width <[number]> Element's width.
   - height <[number]> Element's height.
 
-This method returns boxes of the element, or `null` if the element is not visible. Boxes are represented as an array of quad vertices, x immediately followed by y for each point, points clock-wise. See [getBoxModel](https://chromedevtools.github.io/devtools-protocol/tot/DOM#method-getBoxModel) for more details
+This method returns boxes of the element, or `null` if the element is not visible. Boxes are represented as an array of quad vertices, x immediately followed by y for each point, points clock-wise. See [getBoxModel](https://chromedevtools.github.io/devtools-protocol/tot/DOM#method-getBoxModel) for more details.
 
 #### elementHandle.click([options])
 - `options` <[Object]>
--- a/docs/api.md
+++ b/docs/api.md
@@ -2199,14 +2199,14 @@ This method returns the bounding box of the element (relative to the main frame)
 
 #### elementHandle.boxModel()
 - returns: <[Promise]<?[Object]>>
-  - content <[Array]<[number]>> Content box, represented as a [quad](https://chromedevtools.github.io/devtools-protocol/tot/DOM#type-Quad).
-  - padding <[Array]<[number]>> Padding box, represented as a [quad](https://chromedevtools.github.io/devtools-protocol/tot/DOM#type-Quad).
-  - border <[Array]<[number]>> Border box, represented as a [quad](https://chromedevtools.github.io/devtools-protocol/tot/DOM#type-Quad).
-  - margin <[Array]<[number]>> Margin box, represented as a [quad](https://chromedevtools.github.io/devtools-protocol/tot/DOM#type-Quad).
+  - content <[Array]<[Object]>> Content box, represented as an array of {x, y} points.
+  - padding <[Array]<[Object]>> Padding box, represented as an array of {x, y} points.
+  - border <[Array]<[Object]>> Border box, represented as an array of {x, y} points.
+  - margin <[Array]<[Object]>> Margin box, represented as an array of {x, y} points.
   - width <[number]> Element's width.
   - height <[number]> Element's height.
 
-This method returns boxes of the element, or `null` if the element is not visible. Boxes are represented as an array of quad vertices, x immediately followed by y for each point, points clock-wise. See [getBoxModel](https://chromedevtools.github.io/devtools-protocol/tot/DOM#method-getBoxModel) for more details.
+This method returns boxes of the element, or `null` if the element is not visible. Boxes are represented as an array of quad objects, {x, y} for each point, points clock-wise. See [getBoxModel](https://chromedevtools.github.io/devtools-protocol/tot/DOM#method-getBoxModel) for more details.
 
 #### elementHandle.click([options])
 - `options` <[Object]>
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -2191,6 +2192,17 @@ The method evaluates the XPath expression relative to the elementHandle. If ther
 
 This method returns the bounding box of the element (relative to the main frame), or `null` if the element is not visible.
 
+#### elementHandle.boxModel()
+- returns: <[Promise]<?[Object]>>
+  - content <[Array]<[number]>> Content box, represented as a [quad](https://chromedevtools.github.io/devtools-protocol/tot/DOM#type-Quad).
```
aslushnikov: Let's have an array of objects with`{x, y}` - that will be a more friendly format to work with.

test/elementhandle.spec.js
```diff
@@ -52,6 +52,28 @@ module.exports.addTests = function({testRunner, expect}) {
...
   });
 
+  describe('ElementHandle.boxModel', function() {
+    it('should work', async({page, server}) => {
+      await page.setViewport({width: 500, height: 500});
+      await page.setContent('<div style="width: 100px; height: 100px; margin: 5px; padding: 4px; border: 1px solid black;top: 0; left: 0; position: absolute;"></div>');
+      const elementHandle = await page.$('div');
+      const box = await elementHandle.boxModel();
+
+      expect(box.content).toEqual([10, 10, 110, 10, 110, 110, 10, 110]);
```
aslushnikov: In what coordinates are the returned numbers? Can we have a test with nested frames to make the coordinate system clear?

lib/ElementHandle.js
```diff
@@ -149,6 +156,18 @@ class ElementHandle extends JSHandle {
     return {x, y, width, height};
   }
 
+  /**
+   * @return {!Promise<?object>}
+   */
+  async boxModel() {
+    const result = await this._getBoxModel();
+
+    if (!result)
```
aslushnikov, line 178: nit: to my taste, i'd replace lines 164-168 with a one-liner:

```js
return result ? result.model : null;
```

yanivefraim: this is not relevant after my last change (moving to {x, y} representation)

test/elementhandle.spec.js
yanivefraim: done, 10x!

docs/api.md
yanivefraim: done

lib/ElementHandle.js
```diff
@@ -149,6 +156,30 @@ class ElementHandle extends JSHandle {
...
+      return null;
+
+    return Object.keys(result.model).reduce((prev, key) => {
```
aslushnikov: I'm sure this works, but I'm having hard times following the logic.
I'd rather have something more straight-forward here that outlines the shape of the returned object.

This will make it clear what we return (and also improves code grep'pability!)

```js
return {
  content: fromProtocolQuad(result.content),
  padding: fromProtocolQuad(result.padding),
  // ...
}

function fromProtocolQuad(protocolQuad) {
  // ...
}
```

yanivefraim: done, 10x!

yanivefraim:
@aslushnikov - fixed everything. 10x!

test/elementhandle.spec.js
```diff
@@ -52,6 +52,33 @@ module.exports.addTests = function({testRunner, expect}) {
     });
   });
 
+  describe('ElementHandle.boxModel', function() {
+    it('should work', async({page, server}) => {
+      const leftTop = {x: 28, y: 260};
```
aslushnikov, line 57: I can't validate these numbers from just looking at the test. However, I played with the API and the functionality works as expected.

Let's land this as-is, I'll re-write the test in a follow-up.

aslushnikov (approved):
Thanks!

test/elementhandle.spec.js
yanivefraim: @aslushnikov - agree. Maybe I can add back the previous test (in addition to this one). It had a div with specific dimensions (so you can actually tell that the sizes are good. WDYT?

aslushnikov: @yanivefraim no worries, I came up with something I've just uploaded while playing with your patch. See https://github.com/GoogleChrome/puppeteer/pull/2287

schelkun:
Nice! Thx!

PR #2257, opened 2018-03-25
Health: break page.spec.js to smaller files

yanivefraim:
Break some large chunks into smaller files. This change will remove ~600 lines from `page.spec.js`


test/target.spec.js
```diff
@@ -0,0 +1,134 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
```
aslushnikov: here and in the other created files: 2018.

aslushnikov (approved):
Love this.

PR #2287, opened 2018-03-29
test(ElementHandle.boxModel): remake test to make it more readable

aslushnikov:
This patch:
- re-makes test for EH.boxModel to make it more readable
- slightly changes wording in the `docs/api.md`.

References #1357.

### DIFF
--- a/test/elementhandle.spec.js
+++ b/test/elementhandle.spec.js
@@ -54,22 +54,58 @@ module.exports.addTests = function({testRunner, expect}) {
 
   describe('ElementHandle.boxModel', function() {
     it('should work', async({page, server}) => {
-      const leftTop = {x: 28, y: 260};
-      const rightTop = {x: 292, y: 260};
-      const rightBottom = {x: 292, y: 278};
-      const leftBottom = {x: 28, y: 278};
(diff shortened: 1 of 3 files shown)


aslushnikov:
cc @yanivefraim

test/elementhandle.spec.js
```diff
@@ -54,22 +54,58 @@ module.exports.addTests = function({testRunner, expect}) {
...
+        `;
+        return div;
+      })).asElement();
+
+      // Step 3: query div's boxModel and assert box values.
+      const box = await divHandle.boxModel();
+      expect(box.width).toBe(6);
+      expect(box.height).toBe(7);
+      expect(box.margin[0]).toEqual({
+        x: 1 + 4, // frame.left + div.left
```
yanivefraim, line 94: nice, really liked it! 
Maybe `1 + 4` can be replaced with variables (`frameLeft + divLeft`), so no need for comments here, wdyt? (:

aslushnikov: @yanivefraim started there, and had to pull them from pptr context into the browser context. This felt like too much of a hassle!

PR #2289, opened 2018-03-29
fix(Page): teach Page.waitFor* function to work with strict CSP

aslushnikov:
This patch removes function creation from WaitTask, making the
following methods work with strict CSP policy:
- frame.waitForFunction
- frame.waitForSelector
- frame.waitForXPath

Page counterparts of the methods work as well.

References #1229.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -810,7 +810,11 @@ class WaitTask {
     let success = null;
     let error = null;
     try {
-      success = await (await this._frame.executionContext()).evaluateHandle(waitForPredicatePageFunction, this._predicateBody, this._polling, this._timeout, ...this._args);
+      // Do surgery over evaluation string to insert predicate body and make WaitTask work with strict CSP.
+      let evalFunctionText = waitForPredicatePageFunction.toString();
+      evalFunctionText = evalFunctionText.replace('// _PUPPETEER_PREDICATE_BODY_', this._predicateBody);
+      const evalFunction = eval('var x = ' + evalFunctionText + ';x');
+      success = await (await this._frame.executionContext()).evaluateHandle(evalFunction, this._polling, this._timeout, ...this._args);
     } catch (e) {
       error = e;
     }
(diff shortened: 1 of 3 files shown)


sean-hill (approved):

simoneb:
This doesn't seem to work when `waitForFunction` is provided arguments to be passed to the function, i.e.

```
page.waitForFunction(arg => !!arg, {}, 'some argument')
```

aslushnikov:
@simoneb we have a test for this, it passes:

https://github.com/GoogleChrome/puppeteer/blob/dde45faaebd4232ef953622a8e1e07e544cf64a8/test/frame.spec.js#L139-L147

simoneb:
Doesn't for me in a real use case. Also, I see that the test is passing a DOM elementi as an argument, I never got anything except primitive values to work in this way (i.e. passing a regex wouldn't work, a string would), so I'm wondering if this test is running in a sort of special context

lib/FrameManager.js
```diff
@@ -810,7 +810,11 @@ class WaitTask {
     let success = null;
     let error = null;
     try {
-      success = await (await this._frame.executionContext()).evaluateHandle(waitForPredicatePageFunction, this._predicateBody, this._polling, this._timeout, ...this._args);
+      // Do surgery over evaluation string to insert predicate body and make WaitTask work with strict CSP.
+      let evalFunctionText = waitForPredicatePageFunction.toString();
+      evalFunctionText = evalFunctionText.replace('// _PUPPETEER_PREDICATE_BODY_', this._predicateBody);
+      const evalFunction = eval('var x = ' + evalFunctionText + ';x');
```
JoelEinbinder, line 816: just to make things a bit more straightforward:
```js
const evalFunction = eval('(' + evalFunctionText + ')');
```

JoelEinbinder: Something like this I think will fix the scoped variables from waitForPredicateFunction polluting the user code.
```js
function thing() {
(**waitForPredicateFunction**).call(null, function(..args){**predicate_body**}, ...arguments)
}
```

aslushnikov:
Closing in favor of fixing this upstream in V8: https://chromium-review.googlesource.com/c/v8/v8/+/994070

sean-hill:
Nooooo. I wanted this PR, haha :)

aslushnikov:
@sean-hill we rolled the upstream fix into the puppeteer,  puppeteer@next should work with waitFor* functions.

PR #2296, opened 2018-04-01
feat: add target.targetId() method

CodyRay:
This is the last piece that I needed to be able to support the use case from #1978

With this change, I can create 'incognito pages' using the (public) puppeteer api. [gist](https://gist.github.com/haroldhues/645a7d85d6686dab7f3fe81492d1e608)

There are also `TargetInfo`'s `title` and `attached` properties that aren't exposed. I considered exposing those as well, but I didn't have a use case for them, so I didn't. However, I am happy to do that as well if maintainers think that is a good idea?

PS: I have been unable to get the all the tests to pass on Windows. Is that a known issue?

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -236,6 +236,7 @@
 - [class: Target](#class-target)
   * [target.createCDPSession()](#targetcreatecdpsession)
   * [target.page()](#targetpage)
+  * [target.targetId()](#targettargetid)
   * [target.type()](#targettype)
   * [target.url()](#targeturl)
 - [class: CDPSession](#class-cdpsession)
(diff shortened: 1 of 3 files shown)


JoelEinbinder:
I am generally against exposing DevTools metadata from Puppeteer APIs.

Creating your own targets and then connecting to them with Puppeteer will work, but it feels just as sketchy as using `_targetId` or calling `new Target` yourself.

aslushnikov:
@haroldhues I don't like exposing `targetId`: once the #85 is done, you won't need them any more. 
In general, we don't like exposing DevTools protocol entities in the pptr api to keep them as much separated as possible. This way we can effectively hide underlying browser complexity from the end-users. 

I'd suggest quietly reaching for the `_targetId` instead for now - it is private, but should work fine for your needs.

PR #2319, opened 2018-04-06
test: add test on Page.evaluateOnNewDocument with CSP

aslushnikov:
This patch adds a test that Page.evaluateOnNewDocument works
with CSP: there's been some concerns on the bugtracker before.

References #1229

### DIFF
--- a/test/assets/csp.html
+++ b/test/assets/csp.html
@@ -1,0 +1,1 @@
+<meta http-equiv="Content-Security-Policy" content="default-src 'self'">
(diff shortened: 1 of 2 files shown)


JoelEinbinder (commented):
Can we do this with the header. Maybe the eval happens before the meta tag somehow?

JoelEinbinder (approved):
Csp.html is unused

aslushnikov:
> Csp.html is unused

Comes handy; I'll use it in a follow-up

PR #2320, opened 2018-04-06
feat(Page): Page.addScriptTag should throw when blocked by CSP

aslushnikov:
This patch teaches Page.addScriptTag and Page.addStyleTag to throw
an error when blocked by CSP.

References #1229.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -543,7 +543,7 @@ class Frame {
 
     /**
      * @param {string} content
-     * @return {!HTMLElement}
+     * @return {!Promise<!HTMLElement>}
      */
     function addStyleContent(content) {
       const style = document.createElement('style');
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -470,12 +470,12 @@ class Frame {
       script.src = url;
       if (type)
         script.type = type;
-      document.head.appendChild(script);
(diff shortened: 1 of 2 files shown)


lib/FrameManager.js
```diff
@@ -470,12 +470,12 @@ class Frame {
...
-      document.head.appendChild(script);
-      await new Promise((res, rej) => {
-        script.onload = res;
+      const promise = new Promise((res, rej) => {
+        script.onload = res.bind(null, script);
         script.onerror = rej;
       });
-      return script;
+      document.head.appendChild(script);
+      return promise;
```
JoelEinbinder: nit: 
```js
await promise;
return script;
```
Makes this a bit easier to read. I missed that we were returning the script at first.

lib/FrameManager.js
```diff
@@ -529,24 +533,28 @@ class Frame {
...
     /**
      * @param {string} content
-     * @return {!HTMLElement}
+     * @return {!Promise<!HTMLElement>}
      */
     function addStyleContent(content) {
       const style = document.createElement('style');
       style.type = 'text/css';
       style.appendChild(document.createTextNode(content));
+      const promise = new Promise((res, rej) => {
```
JoelEinbinder: same comment as above.

JoelEinbinder (approved):

PR #2324, opened 2018-04-06
feat(Page): introduce Page.setBypassCSP

### LINKED ISSUE #1229: Page with CSP breaks puppeteer

aslushnikov:
This patch introduces `page.setBypassCSP` method that allows clients
to ignore Content-Security-Policy for a given page.

Fixes #1229.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -90,6 +90,7 @@
   * [page.reload(options)](#pagereloadoptions)
   * [page.screenshot([options])](#pagescreenshotoptions)
   * [page.select(selector, ...values)](#pageselectselector-values)
+  * [page.setBypassCSP(enabled)](#pagesetbypasscspenabled)
   * [page.setCacheEnabled(enabled)](#pagesetcacheenabledenabled)
   * [page.setContent(html)](#pagesetcontenthtml)
   * [page.setCookie(...cookies)](#pagesetcookiecookies)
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -1176,6 +1177,15 @@ page.select('select#colors', 'red', 'green', 'blue'); // multiple selections
 
 Shortcut for [page.mainFrame().select()](#frameselectselector-values)
 
+#### page.setBypassCSP(enabled)
+- `enabled` <[boolean]> sets by-passing of page's Content-Security-Policy.
+- returns: <[Promise]>
+
+Toggles by-passing page's Content-Security-Policy.
```
JoelEinbinder: Bypassing I think

JoelEinbinder (approved):

TomasHubelbauer:
Is there a way to tell when this will land? It's not a part of any milestone and I don't think there is a roadmap for this yet, right? So probably *Subscribe* to this issue and that's it?

aslushnikov:
@TomasHubelbauer you can find the next release date in the very top of [`docs/api.md`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md).

We'll release this on April 12 in version 1.3.0.

PR #2328, opened 2018-04-07
chore(types): type check the rest of the protocol

JoelEinbinder:
This adds events and commands into `protocol.d.ts`

### DIFF
--- a/lib/EmulationManager.js
+++ b/lib/EmulationManager.js
@@ -33,13 +33,13 @@ class EmulationManager {
     const width = viewport.width;
     const height = viewport.height;
     const deviceScaleFactor = viewport.deviceScaleFactor || 1;
+    /** @type {Protocol.Emulation.ScreenOrientation} */
     const screenOrientation = viewport.isLandscape ? { angle: 90, type: 'landscapePrimary' } : { angle: 0, type: 'portraitPrimary' };
 
     await Promise.all([
       this._client.send('Emulation.setDeviceMetricsOverride', { mobile, width, height, deviceScaleFactor, screenOrientation }),
       this._client.send('Emulation.setTouchEmulationEnabled', {
-        enabled: viewport.hasTouch || false,
-        configuration: viewport.isMobile ? 'mobile' : 'desktop'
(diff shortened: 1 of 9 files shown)


lib/EmulationManager.js
```diff
@@ -33,13 +33,13 @@ class EmulationManager {
...
     const height = viewport.height;
     const deviceScaleFactor = viewport.deviceScaleFactor || 1;
+    /** @type {Protocol.Emulation.ScreenOrientation} */
     const screenOrientation = viewport.isLandscape ? { angle: 90, type: 'landscapePrimary' } : { angle: 0, type: 'portraitPrimary' };
 
     await Promise.all([
       this._client.send('Emulation.setDeviceMetricsOverride', { mobile, width, height, deviceScaleFactor, screenOrientation }),
       this._client.send('Emulation.setTouchEmulationEnabled', {
-        enabled: viewport.hasTouch || false,
-        configuration: viewport.isMobile ? 'mobile' : 'desktop'
```
aslushnikov, line 42: why is `configuration` removed?

lib/externs.d.ts
```diff
@@ -12,7 +12,10 @@ import * as child_process from 'child_process';
 declare global {
   module Puppeteer {
     export class Connection extends RealConnection {}
-    export class CDPSession extends RealCDPSession {}
+    export class CDPSession extends RealCDPSession {
+      on<T extends keyof Protocol.Events>(event: T, listener: (arg: Protocol.Events[T]) => void): this;
```
aslushnikov, line 16: this looks like black magic

lib/EmulationManager.js
JoelEinbinder: There is no such option on `Emulation.setTouchEmulationEnabled`. It is only on `Page.setTouchEmulationEnabled`.

lib/externs.d.ts
JoelEinbinder: It is!

aslushnikov (approved):

PR #2335, opened 2018-04-09
feat(ElementHandle): accessible name

JoelEinbinder:
This relies on an upstream [patch](https://chromium-review.googlesource.com/c/chromium/src/+/1000713), but I expect bikeshedding on the API so I will do the roll separately.

`elementHandle.accessibleName()` returns the computed accessible name for the node. This is not the entirety of what would be read by a screen reader for this element, that depends on the role and value of this node. But the accessible name is one of the hardest properties to compute because it can't be read directly from the aria attributes.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -185,6 +185,7 @@
   * [elementHandle.$(selector)](#elementhandleselector)
   * [elementHandle.$$(selector)](#elementhandleselector)
   * [elementHandle.$x(expression)](#elementhandlexexpression)
+  * [elementHandle.accessibleName()](#elementhandleaccessiblename)
   * [elementHandle.asElement()](#elementhandleaselement)
   * [elementHandle.boundingBox()](#elementhandleboundingbox)
   * [elementHandle.boxModel()](#elementhandleboxmodel)
@@ -2204,8 +2205,14 @@ The method runs `element.querySelectorAll` within the page. If no elements match
 
 The method evaluates the XPath expression relative to the elementHandle. If there's no such element, the method will resolve to `null`.
 
+#### elementHandle.accessibleName()
+- returns: <[Promise]<?[string]>>
+
+This method computed the acessible name of the element that would exposed to a screen reader.
+Might return `null` if the element is ignored for accessibility, e.g. a hidden element or a simple container element.
+
 #### elementHandle.asElement()
-- returns: <[elementhandle]>
+- returns: <[ElementHandle]>
 
 #### elementHandle.boundingBox()
 - returns: <[Promise]<?[Object]>>
--- a/docs/api.md
+++ b/docs/api.md
@@ -185,7 +185,7 @@
   * [elementHandle.$(selector)](#elementhandleselector)
   * [elementHandle.$$(selector)](#elementhandleselector)
   * [elementHandle.$x(expression)](#elementhandlexexpression)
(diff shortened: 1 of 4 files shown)


aslushnikov:
How about a separate namespace for AX features, e.g. `page.accessibility.*`, where we'll put all the ax-related functionality?

Another idea: prefix all ax-related methods with `ax`.

JoelEinbinder:
A separate namespace on the page might be a little verbose. Every command is going to have an element passed into it. There is also some powerful information available through the accessibility API that we might want to expose on its own. For example whether or not a `<select>` is current expanded.

cc: @alice

alice:
"ax" is not a standard prefix for accessibility - just a WebKit weirdness (which funnily enough Apple's WebKit no longer even uses). "a11y" is more common, but still confuses people. There's no good solution here unfortunately (cf. https://github.com/WICG/aom/blob/gh-pages/explainer.md#appendix-accessiblenode-naming)

I would say that very few of the semantic properties are straightforward to compute from ARIA, as usually there are default or implicit values for things, or rules for whether the ARIA semantics can apply in the given context.

It would be good to be able to get more computed properties than just the name. Would it be possible to have an API similar to the wire API which fetches the "accessible node", but possibly whitelists a limited set of properties?

JoelEinbinder:
I was playing around with having `elementHandle.a11yNode()` return something with an interface like
```ts
class A11yNode {
    elementHandle(): Promise<ElementHandle>;
    parent(): A11yNode ;
    children(): A11yNode[];
// Properties:  busy, disabled, hidden, hiddenRoot, invalid, keyshortcuts, roledescription, live, atomic, relevant, root, autocomplete, haspopup, level, multiselectable, orientation, multiline, readonly, required, valuemin, valuemax, valuetext, checked, expanded, modal, pressed, selected, activedescendant, controls, describedby, details, errormessage, flowto, labelledby, owns
}
```
Using the API became cumbersome though, because the lifetime of an A11yNode is a single layout, so I have to constantly re-query what look like persistent references.

alice:
What if it was something like `page.accessibility.snapshot(elementHandle)`? Wouldn't make it less cumbersome, but at least would communicate clearly that it wasn't a live object.

docs/api.md
```diff
@@ -2200,8 +2201,14 @@ The method runs `element.querySelectorAll` within the page. If no elements match
 
 The method evaluates the XPath expression relative to the elementHandle. If there's no such element, the method will resolve to `null`.
 
+#### elementHandle.accessibleName()
+- returns: <[Promise]<?[string]>>
+
+This method computed the acessible name of the element that would exposed to a screen reader.
```
vsemozhetbyt: computed -> computes?
acessible -> accessible?
would exposed -> would be exposed?

JoelEinbinder:
I pushed an experiment where I return all properties on the axnode that aren't nodes or references. I like the look of the objects it returns, but I am not sure what everything means enough to write the API docs.

There is some strangeness with the wire API that I would like to even out if possible. Like sometimes `false` is a boolean, but other times `'false'` is a string if it could be `'mixed'`.

aslushnikov:
@JoelEinbinder do we have any plan to eventually land this?

JoelEinbinder:
yes

JoelEinbinder:
@robdodson

JoelEinbinder:
Closing this in favor of something event based.

PR #2336, opened 2018-04-09
test: refactor utils.waitForEvents into utils.waitEvent

aslushnikov:
This simplifies waiting for events in tests.

### DIFF
--- a/test/page.spec.js
+++ b/test/page.spec.js
@@ -377,7 +377,7 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions, p
     describe('Page.Events.DOMContentLoaded', function() {
       it('should fire when expected', async({page, server}) => {
         page.goto('about:blank');
-        await waitEvent(page, 'domcontentloaded', 1);
+        await waitEvent(page, 'domcontentloaded');
       });
     });
 
--- a/test/page.spec.js
+++ b/test/page.spec.js
@@ -16,7 +16,7 @@
 const fs = require('fs');
 const path = require('path');
 const utils = require('./utils');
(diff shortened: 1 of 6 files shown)


test/page.spec.js
```diff
@@ -380,7 +377,7 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions, p
     describe('Page.Events.DOMContentLoaded', function() {
       it('should fire when expected', async({page, server}) => {
         page.goto('about:blank');
-        await waitForEvents(page, 'domcontentloaded', 1);
+        await waitEvent(page, 'domcontentloaded', 1);
```
JoelEinbinder: extra 1 here

test/page.spec.js
```diff
@@ -340,20 +340,17 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions, p
...
-            console.time('calling console.time');
-            console.timeEnd('calling console.time');
-            console.trace('calling console.trace');
-            console.dir('calling console.dir');
-            console.warn('calling console.warn');
-            console.error('calling console.error');
-            console.log(Promise.resolve('should not wait until resolved!'));
-          }),
-          // Wait for 5 events to hit - console.time is not reported
-          waitForEvents(page, 'console', 5)
```
JoelEinbinder, line 355: This works because console logs wait for the ack in JavaScript?

JoelEinbinder (approved):

test/page.spec.js
aslushnikov: Done.

test/page.spec.js
aslushnikov: That's because protocol is serial: console.log's will be posted before the evaluation response.

PR #2338, opened 2018-04-10
fix(Page): support anchor navigation

### LINKED ISSUE #257: Hash-only navigation doesn't work

aslushnikov:
This patch fixes puppeteer navigation primitives to work with
same-document navigation.

In browser world, same-document navigation happens when document's URL is changed,
but document instance is not re-created. Some common scenarios
for same-document navigation are:
- History API
- anchor navigation

With this patch:
- pptr starts dispatching `framenavigated` event when frame's URL gets
changed due to same-document navigation
- `page.waitForNavigation` now works with same-document navigation
- `page.goBack()` and `page.goForward()` are handled correctly.

Fixes #257.

### DIFF
--- a/lib/NavigatorWatcher.js
+++ b/lib/NavigatorWatcher.js
@@ -43,8 +43,10 @@ class NavigatorWatcher {
     this._frame = frame;
     this._initialLoaderId = frame._loaderId;
     this._timeout = timeout;
+    this._hasSameDocumentNavigation = false;
     this._eventListeners = [
       helper.addEventListener(this._frameManager, FrameManager.Events.LifecycleEvent, this._checkLifecycleComplete.bind(this)),
+      helper.addEventListener(this._frameManager, FrameManager.Events.FrameNavigatedWithinDocument, this._navigatedWithingDocument.bind(this)),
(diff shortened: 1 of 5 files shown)


lib/NavigatorWatcher.js
```diff
@@ -43,8 +43,10 @@ class NavigatorWatcher {
     this._frame = frame;
     this._initialLoaderId = frame._loaderId;
     this._timeout = timeout;
+    this._hasSameDocumentNavigation = false;
     this._eventListeners = [
       helper.addEventListener(this._frameManager, FrameManager.Events.LifecycleEvent, this._checkLifecycleComplete.bind(this)),
+      helper.addEventListener(this._frameManager, FrameManager.Events.FrameNavigatedWithinDocument, this._navigatedWithingDocument.bind(this)),
```
JoelEinbinder: WithinDocument

JoelEinbinder (approved):

lib/NavigatorWatcher.js
aslushnikov: Done.

Issue #2167, opened 2018-03-08
Scripts are not evalutated in header and footer HTML templates for PDFs.


chrisjpatty:
### Steps to reproduce

* Puppeteer version: 1.1.1
* Platform / OS version: Windows 10, version 1709, build 16299.248
* Node.js version: v7.9.0

**What steps will reproduce the problem?**

First, I setup a basic test page like this:

```jsx
const browser = await puppeteer.launch({ignoreHTTPSErrors: true})
const page = await browser.newPage()
await page.setContent(`
<html>
  <head>
    <title>Test Page</title>
    <style>
      body{
        background: yellow;
        min-height: 6000px;
      }
    </style>
  </head>
  <body>
    I'm an HTML page.
  </body>
</html>
`)
const pdf = await page.pdf({
  format: 'Letter',
  printBackground: true,
  headerTemplate: headerTemplate,
  displayHeaderFooter: true,
  margin: {
    top: '1in',
    right: '0in',
    bottom: '1in',
    left: '0in'
  }
})
await browser.close()
```

The `headerTemplate` looks like this:
```jsx
const headerTemplate = `
<div style="font-size: 10px; display: flex; flex-direction: row; justify-content: space-between; width: 100%" id='template'>
  <div class='pageNumber' id='num' style="font-size: 10px;">Page number</div>
  <div class='date' style="font-size: 10px;"></div>
  <div class='title' style="font-size: 10px;"></div>
  <div class='url' style="font-size: 10px;"></div>
  <div class='totalPages' style="font-size: 10px;"></div>
  <script>
    var pageNum = document.getElementById("num");
    pageNum.remove()
    var template = document.getElementById("template")
    template.style.background = 'red';
  </script>
</div>`
```

**What is the expected result?**
The docs don't mention whether or not script tags are supported in the header and footer templates. I would expect then that the code above would render a header with a red background and remove the page number node. (The code above was just for testing purposes).

**What happens instead?**
The script appears not to evaluate. Instead the resulting PDF has no background, and a page number div.

If evaluating scripts in these templates is not supported, might I suggest updating the PDF section of the docs to state as much. Thanks for all the hard work the Puppeteer team has put into this project!

jmas:
I also need this feature, because I do not see a normal way to make different headers for different pages. For example on first page (intro page) is one header and on the internal pages is different.
If not possible run scripts in header/footer templates I have another proposal that can be solved via CSS:
```html
[...]

PR #2341, opened 2018-04-10
docs(api.md): mention header/footer template limitations in `page.pdf`.

aslushnikov:
References #2167.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1117,6 +1117,10 @@ The `format` options are:
 - `A5`: 5.83in x 8.27in
 - `A6`: 4.13in x 5.83in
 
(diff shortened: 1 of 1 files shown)


JoelEinbinder (approved):

RayAnti2:
I found puppeteer is a good tool for converting from html to pdf.
But the footerTemplate must be the same on all pages?
Now I need, for example, pages' footer like this:

(page 1)
pageNumber of totalNumber---------------------------vocabulary list

(page 2)
vocabulary list-------------------------------------pageNumber of totalNumber

(page 3)
pageNumber of totalNumber---------------------------vocabulary list

I prepare 2 Templates for odd and even num pages, i don't need every pages with different footers, but just these two Templates.
What should i do ?

PR #2344, opened 2018-04-10
feat(npm): support depending on un-published code

intellix:
Just noticed that a commit just landed to support waitForNavigation on same-document navigations and it's really important for our project.

To use it, it looks like we'll have to wait for a publish. The npm prepare script was added in 5.0.0 and allows you to depend on a git project directly as it'll install it's devDependencies and run the prepare script as though it was published

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "unit-node6": "node node6/test/test.js",
     "tsc": "node utils/protocol-types-generator && tsc -p .",
     "prepublishOnly": "npm run build",
+    "prepare": "npm run build",
     "apply-next-version": "node utils/apply_next_version.js"
   },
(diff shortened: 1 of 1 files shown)


aslushnikov:
@intellix we publish tip-of-tree puppeteer to npm as `puppeteer@next`. Will it work for you?

Also, note: the next pptr version will be published on April, 12.

intellix:
Sounds good to me, but with the above merged, we can do:
```
"puppeteer": "github:GoogleChrome/puppeteer#master"
```

I'm happy with @next though and April 12th is cool :)

aslushnikov:
> I'm happy with @next though and April 12th is cool :)

Sweet, closing this then.

Issue #2262, opened 2018-03-26
Problem with PUPPETEER_SKIP_CHROMIUM_DOWNLOAD


Tmin10:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.2.0
* Platform / OS version: windows 10 x64
* URLs (if applicable): -
* Node.js version: 8.1.3 / NPM: 4.1.2

**What steps will reproduce the problem?**

_Please include code that reproduces the issue._

1. Add `"PUPPETEER_SKIP_CHROMIUM_DOWNLOAD": true` into package.json config section.
2. Add `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true` to local .npmrc file.
3. Run `npm config set PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true` to put property in global .npmrc file.
4. Run `npm i puppeteer` to install puppeteer.

**What is the expected result?**
Puppeteer won't download chromium.

**What happens instead?**
Puppeteer downloaded chromium. (-_-)

In [install.js](https://github.com/GoogleChrome/puppeteer/blob/master/install.js#L19) we can see that puppeteer tries to read `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`, `NPM_CONFIG_PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` and `npm_config_puppeteer_skip_chromium_download` environment variables and didn't read any config. 

How it could work? Who should have to fill this variables if I would like to use only npm config without global envirements?

jeffguorg:
try this:
```bash
env PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true npm i puppeteer
```
or just set it in your shell and run npm
```bash
export PUPPETEER_SKIP_CHROMIUM_DOWNLOAD=true
npm i puppeteer
```
i'm not experienced in node.js, but if i'm right, process.env only reads environment variables from the parent of the process, which is your shell.

alixaxel:
@Tmin10 Check the solution on #2270.

kwkbtr:
I tried this by adding the following to `package.json` and it did not work:

```json
"config": {
  "puppeteer_skip_chromium_download": true
}
```

According to [npm documentation](https://docs.npmjs.com/files/package.json#config), the variable specified in `package.json` is prefixed by `npm_package_config_`, not `npm_config_`.
[...]

PR #2345, opened 2018-04-10
feat: Support configuration parameters in package.json (#2262)

kwkbtr:
Background: https://github.com/GoogleChrome/puppeteer/issues/2262#issuecomment-378122692
This patch adds support for config variables in `package.json` in the same way as config variables in `.npmrc`.
See npm documentation: https://docs.npmjs.com/files/package.json#config

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -268,7 +268,7 @@ The Puppeteer API is hierarchical and mirrors the browser structure. On the foll
 ### Environment Variables
 
 Puppeteer looks for certain [environment variables](https://en.wikipedia.org/wiki/Environment_variable) to aid its operations.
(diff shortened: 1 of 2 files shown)


JoelEinbinder:
Can you provide a quick example on how to use this? It looks cool.

aslushnikov:
Is package.json config inherited by sub-sub-modules?

JoelEinbinder:
From my tests, it looks like `config` in package.json only affects the root package, and is not visible in its dependencies. That means that to use these config variables, you'd have to edit Puppeteer's `package.json`.

aslushnikov:
@JoelEinbinder sounds like this is not useful for our users then.

Issue #1896, opened 2018-01-24
Access to body content of a 3XX response 


phackt:
Hello, i would like to access to the response.text() having a response status code in the 3XX range. 

The fact is that the response.text() Promise in this case will stay in a Pending state. Example for a request on ```https://google.com```, i would like to output:

```html
<HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>302 Moved</TITLE></HEAD><BODY>
<H1>302 Moved</H1>
The document has moved
<A HREF="https://www.google.fr/?gfe_rd=cr&amp;dcr=0&amp;ei=jMJoWrHVI-_t8we8_67ICg">here</A>.
</BODY></HTML>
```

A part of the response:
```
...
_contentPromise: Promise { <pending> },
 _status: 302,
_url: 'https://google.com/',
...
```

but the following code will never trigger (and in my case will also exit the main process, that's the most annoying part):
```
response.text().then(
    function(result) {
        console.log('result: ' + result)
});
```

Do you have any idea of how i can access to the body content of this response (request is flagged as resource type ```document```).

Thanks in advance,

aslushnikov:
@phackt Indeed, I found a bug in pptr regarding this. But there's still a fundamental issue in devtools protocol: we won't be able to fetch content of a redirect response. It will be even harder to do so in future with the network layer refactoring in chrome.

What's your usecase? I don't think browser users ever get to see the redirect response body.

PR #2352, opened 2018-04-11
fix(Network): explicitly throw on content request for redirect response

aslushnikov:
DevTools protocol doesn't support returning body of redirect responses.
We should explicitly throw in this case.

References #1896.

### DIFF
--- a/test/network.spec.js
+++ b/test/network.spec.js
@@ -103,7 +103,7 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(redirected.status()).toBe(302);
       let error = null;
       await redirected.text().catch(e => error = e);
-      expect(error).toBeTruthy();
+      expect(error.message).toContain('Response body is unavailable for redirect responses');
     });
     it('Page.Events.Response should not report body unless request is finished', async({page, server}) => {
       await page.goto(server.EMPTY_PAGE);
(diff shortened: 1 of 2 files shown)


test/network.spec.js
```diff
@@ -94,6 +94,17 @@ module.exports.addTests = function({testRunner, expect}) {
...
     });
+    it('Page.Events.Response should throw when requesting body of redirected response', async({page, server}) => {
+      server.setRedirect('/foo.html', '/empty.html');
+      const response = await page.goto(server.PREFIX + '/foo.html');
+      const redirectChain = response.request().redirectChain();
+      expect(redirectChain.length).toBe(1);
+      const redirected = redirectChain[0].response();
+      expect(redirected.status()).toBe(302);
+      let error = null;
+      await redirected.text().catch(e => error = e);
```
JoelEinbinder, line 105: for next time `let error = await redirected.text().catch(e => e)`

test/network.spec.js
```diff
@@ -94,6 +94,17 @@ module.exports.addTests = function({testRunner, expect}) {
...
+      expect(error).toBeTruthy();
```
JoelEinbinder: .message

JoelEinbinder (approved):

PR #2354, opened 2018-04-11
docs(api.md): add explanation to page.waitForNavigation

### LINKED ISSUE #1860: Docs missing for page.waitForNavigation()
The docs for [`page.waitForNavigation()`](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagewaitfornavigationoptions) includes only an API outline. All other methods have at least one paragraph stating what the function does and why you might want to use it, but this one doesn't.

I can guess from the name that it waits for navigation, but it's not clear when it might be necessary to use this, since other methods like `.goto()` seem to do this anyway.

aslushnikov:
Fixes #1860

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1404,6 +1404,19 @@ Shortcut for [page.mainFrame().waitForFunction(pageFunction[, options[, ...args]
     - `networkidle2` - consider navigation to be finished when there are no more than 2 network connections for at least `500` ms.
 - returns: <[Promise]<[Response]>> Promise which resolves to the main resource response. In case of multiple redirects, the navigation will resolve with the response of the last redirect.
 
+The method waits for the navigation to happen in future. Navigation is an event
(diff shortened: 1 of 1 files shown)


JoelEinbinder:
How does this sound?

This resolves when the page navigates to a new URL or reloads. It is useful for when you run code which will indirectly cause the page to navigate. Consider this example:

```js
const navigationPromise = page.waitForNavigation();
await page.click('a.my-link'); // Clicking the link will indirectly cause a navigation
await navigationPromise; // The navigationPromise resolves after navigation has finished 
```

**NOTE** Usage of the [History API](https://developer.mozilla.org/en-US/docs/Web/API/History_API) to change the URL is considered a navigation.

aslushnikov:
@JoelEinbinder can you please take a look?

JoelEinbinder (approved):

PR #2360, opened 2018-04-11
feat(Tracing): return a buffer from tracing.stop

### LINKED ISSUE #2148: Tracing: stop() should return Promise<Buffer>
It would be convenient if `page.tracing.stop()` returned `Promise<Buffer>` so users could act ton the trace JSON without saving a file to disk. This would also be consistent with the screenshot API.

yanivefraim:
Fixes #2148

### DIFF
--- a/lib/Tracing.js
+++ b/lib/Tracing.js
@@ -74,14 +74,17 @@ class Tracing {
   async _readStream(handle, path) {
     let eof = false;
     const file = await openAsync(path, 'w');
+    const bufs = [];
     while (!eof) {
       const response = await this._client.send('IO.read', {handle});
       eof = response.eof;
+      bufs.push(new Buffer(response.data));
       if (path)
         await writeAsync(file, response.data);
     }
     await closeAsync(file);
     await this._client.send('IO.close', {handle});
+    return Buffer.concat(bufs);
   }
 }
 helper.tracePublicAPI(Tracing);
--- a/lib/Tracing.js
+++ b/lib/Tracing.js
@@ -35,7 +35,6 @@ class Tracing {
    */
   async start(options) {
     console.assert(!this._recording, 'Cannot start recording trace while already recording trace.');
-    console.assert(options.path, 'Must specify a path to write trace file to.');
 
     const defaultCategories = [
       '-*', 'devtools.timeline', 'v8.execute', 'disabled-by-default-devtools.timeline',
@@ -73,7 +72,9 @@ class Tracing {
    */
   async _readStream(handle, path) {
     let eof = false;
-    const file = await openAsync(path, 'w');
+    let file;
+    if (path)
+      file = await openAsync(path, 'w');
     const bufs = [];
     while (!eof) {
       const response = await this._client.send('IO.read', {handle});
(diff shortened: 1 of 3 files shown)


test/tracing.spec.js
```diff
@@ -52,5 +52,12 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(error).toBeTruthy();
       await page.tracing.stop();
     });
+    it('should return a buffer', async({page, server, outputFile}) => {
+      await page.tracing.start({screenshots: true, path: outputFile});
```
aslushnikov, line 58: can we drop a necessity for the path as a part of this patch? This way we'll align with `page.screenshot` and `page.pdf`.

yanivefraim: @aslushnikov - sure, done.

lib/Tracing.js
```diff
@@ -73,15 +72,21 @@ class Tracing {
...
       eof = response.eof;
+      bufs.push(new Buffer(response.data));
       if (path)
         await writeAsync(file, response.data);
     }
-    await closeAsync(file);
+    if (path)
+      await closeAsync(file);
     await this._client.send('IO.close', {handle});
+    return Buffer.concat(bufs);
```
aslushnikov: @yanivefraim I'm recalling now that back in the days we didn't implement buffers here because buffers are constrained to 1Gb in size.

It's unlikely someone has such a huge trace, but it's nice to know what will happen in this case. 
If this line throws an error, let's catch it and return `null` to not compromise existing users.

yanivefraim: @aslushnikov - sure, done. 10x!

test/tracing.spec.js
```diff
@@ -23,34 +23,60 @@ module.exports.addTests = function({testRunner, expect}) {
...
-    });
-    it('should throw if tracing on two pages', async({page, server, browser, outputFile}) => {
-      await page.tracing.start({path: outputFile});
-      const newPage = await browser.newPage();
-      let error = null;
-      await newPage.tracing.start({path: outputFile}).catch(e => error = e);
-      await newPage.close();
-      expect(error).toBeTruthy();
-      await page.tracing.stop();
+    it('should suppert a buffer without a path', async({page, server}) => {
```
JoelEinbinder: support

test/tracing.spec.js
```diff
@@ -23,34 +23,60 @@ module.exports.addTests = function({testRunner, expect}) {
   const {beforeAll, beforeEach, afterAll, afterEach} = testRunner;
 
   describe('Tracing', function() {
-    beforeEach(function(state) {
-      state.outputFile = path.join(__dirname, 'assets', `trace-${state.parallelIndex}.json`);
-    });
-    afterEach(function(state) {
-      fs.unlinkSync(state.outputFile);
-      state.outputFile = null;
+    describe('with a path', () => {
```
JoelEinbinder: I don't think this extra describe `with a path` is necessary. It complicates the git blame on these tests.

lib/Tracing.js
JoelEinbinder: Just to confirm, I tested it and a large buffer throws `Uncaught RangeError: Invalid typed array length`.

test/tracing.spec.js
yanivefraim: fixed, 10x!

test/tracing.spec.js
yanivefraim: fixed. 10x

aslushnikov (approved):

Issue #1183, opened 2017-10-27
window inner size not equal to viewport size


ghost:
**Tell us about your environment:**

* Puppeteer version: 0.13.0-alpha
* Platform / OS version: macos high sierra 10.13
* URLs (if applicable):

**What steps will reproduce the problem?**

_Please include code that reproduces the issue._

```javascript
// start.js

const puppeteer = require('puppeteer')

const run = async () => {
  const browser = await puppeteer.launch({
    headless: false,
    slowMo: 250,
    args: [
      '--disable-infobars',
    ],
  });

  const page = await browser.newPage();

  await page.goto('https://www.google.com/');

  await page.waitFor(60000)
  await browser.close();
}

run()
```

1. `node start.js`

**What is the expected result?**

1. window inner size and viewport should be equal, all the default 800px x 600px

**What happens instead?**

1. window inner size is larger than the viewport size

![google-puppeteer.png](https://i.loli.net/2017/10/27/59f2f0f3b461e.png)

entrptaher:
To add more salt into it, opening devtools will make the viewport lose the predefined size. For example,
![image](https://user-images.githubusercontent.com/8284972/32138788-178bc6cc-bc5b-11e7-8050-34543d4a2705.png)

zicongxie:
It has a param can be reset the window viewports in launch method.

ghost:
@zicongxie `await page.setViewport({ width: 600, height: 400 })` only change the viewport size, but the window size will not changed to 600px x 400px too.

ghost:
@entrptaher I noticed that too.

aslushnikov:
@chanjsq what's your usecase? Why would you need to change the window size?

entrptaher:
For me, say I will want to test responsiveness.

ghost:
@aslushnikov When setting the viewport size larger than the default window size, we also need to set the window size with the same value to see the whole page width, it's convenient if the window size can change with viewport size accordingly.

mike-aungsan:
Have the same issue on Windows 8.

aslushnikov:
@entrptaher isn't viewport enough for responsiveness?
@chanjsq so this is mostly for debugging purposes?

ghost:
@aslushnikov yes, smart can be better

z-vr:
can i make the viewport equal to the size of the window by default, and adjustable while resizing? i don't want the viewport which is not equal to the window size :( +1 on this one

[...]

PR #2364, opened 2018-04-12
feat(Page): allow disabling viewport emulation

### LINKED ISSUE #2358: No way to determine native DPR of device
The following code doesn't allow someone to use the device's native DPR by setting `await page.setViewport({deviceScaleFactor: 0,...})`:

https://github.com/GoogleChrome/puppeteer/blob/master/lib/Page.js#L729
https://github.com/GoogleChrome/puppeteer/blob/294f33b75c5a7c87d447816218438f56c2d56f0e/lib/EmulationManager.js

That makes it impossible to determine the real DPR of the device:

```js
await page.evaluate('window.devicePixelRatio'); 
```

aslushnikov:
This patch:
- teaches page.setViewpot() to accept `null` to disable
viewport emulation.
- allows passing `0` as viewport's deviceScaleFactor to use machine-defined
deviceScaleFactor.

References #1183.
Fixes #2358.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1275,7 +1275,7 @@ puppeteer.launch().then(async browser => {
 - returns: <[Promise]> Promise which resolves when the user agent is set.
 
 #### page.setViewport(viewport)
-- `viewport` <[Object]>
+- `viewport` <?[Object]> Disables viewport emulation if `null` is passed.
   - `width` <[number]> page width in pixels.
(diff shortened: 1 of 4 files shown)


aslushnikov:
`Emulation.clearDeviceMetricsOverride` turned out to be buggy; we'd need to wait for the [upstream fix](https://chromium-review.googlesource.com/c/chromium/src/+/1011121) to land before landing this.

PR #2376, opened 2018-04-14
feat(Response): add response.remoteAddress method

### LINKED ISSUE #2171: Can I get Remote Address with puppteer？
Can I get Remote Address with puppteer？

szdc:
This patch exposes a remoteAddress property which is a formatted string containing the IP address and port of the remote server

Closes #2171 

I wasn't sure if I should expose the remote address and port separately (rather than a formatted version), so please let me know if that's a better idea.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -223,6 +223,7 @@
   * [response.headers()](#responseheaders)
   * [response.json()](#responsejson)
   * [response.ok()](#responseok)
+  * [response.remoteAddress()](#responseremoteaddress)
   * [response.request()](#responserequest)
   * [response.securityDetails()](#responsesecuritydetails)
   * [response.status()](#responsestatus)
@@ -2512,6 +2513,11 @@ This method will throw if the response body is not parsable via `JSON.parse`.
 - returns: <[boolean]>
 
 Contains a boolean stating whether the response was successful (status in the range 200-299) or not.
+
+#### response.remoteAddress()
+- returns: <[Object]>
+  - `ip` <[string]> the IP address of the remote server
+  - `port` <[string]> the port used to connect to the remote server
 
 #### response.request()
 - returns: <[Request]> A matching [Request] object.
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -154,7 +154,7 @@ class NetworkManager extends EventEmitter {
     if (event.redirectUrl) {
       const request = this._interceptionIdToRequest.get(event.interceptionId);
       if (request) {
-        this._handleRequestRedirect(request, event.responseStatusCode, event.responseHeaders, false /* fromDiskCache */, false /* fromServiceWorker */, null /* securityDetails */);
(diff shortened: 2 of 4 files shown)


szdc:
Thoughts on how I tackle the IPv4 loopback address in the Travis build?

lib/NetworkManager.js
```diff
@@ -188,7 +188,7 @@ class NetworkManager extends EventEmitter {
    * @param {?Object} securityDetails
    */
   _handleRequestRedirect(request, redirectStatus, redirectHeaders, fromDiskCache, fromServiceWorker, securityDetails) {
-    const response = new Response(this._client, request, redirectStatus, redirectHeaders, fromDiskCache, fromServiceWorker, securityDetails);
+    const response = new Response(this._client, request, redirectStatus, redirectHeaders, fromDiskCache, fromServiceWorker, securityDetails, null, null);
```
aslushnikov: let's plumb the remoteIDAddress and remotePort to the redirect responses (also let's cover this with a test)

examples/search.js
```diff
@@ -27,8 +27,6 @@ const puppeteer = require('puppeteer');
   const browser = await puppeteer.launch();
   const page = await browser.newPage();
 
-  await page.goto('https://developers.google.com/web/');
```
aslushnikov: why this change?

docs/api.md
```diff
@@ -2513,6 +2514,9 @@ This method will throw if the response body is not parsable via `JSON.parse`.
 
 Contains a boolean stating whether the response was successful (status in the range 200-299) or not.
 
+#### response.remoteAddress()
+- returns: <[string]> The formatted address and port of the remote server.
```
aslushnikov: I'd return IP address and port separately: it's a little error-prone to parse out port number with IPv6 addresses.

For example, this method can return an object `{ip, port}`

test/network.spec.js
```diff
@@ -56,6 +56,7 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(responses[0].fromCache()).toBe(false);
       expect(responses[0].fromServiceWorker()).toBe(false);
       expect(responses[0].request()).toBeTruthy();
+      expect(responses[0].remoteAddress()).toBe(`[::1]:${server.PORT}`);
```
aslushnikov: I'd try navigating to IP explicitly in line 51 to make this less flaky across environments (different setups might resolve
`localhost` differently)

```js
await page.goto(`http://[::1]:${server.PORT}/empty.html`);
```

examples/search.js
szdc: Accidental; sorry.

lib/NetworkManager.js
szdc: Good point - I'll update with a test later today.

test/network.spec.js
szdc: We'll still have trouble on IPv4-only machines as they won't be able to handle `[::1]`.  If we use `127.0.0.1` instead, Chrome still returns `[::1]` on IPv6-capable machines.

aslushnikov:
Close in favor of #3192.

PR #2400, opened 2018-04-17
Feat(Page): Add 'Log' domain entries

### LINKED ISSUE #1939: Uncaught browser console error message
* Puppeteer version: 1.0.0-post, chromium_revision 526987
* Node.js version: 9.4.0

```js
'use strict';

(async function main() {
  try {
    const browser = await puppeteer.launch({ headless: false });

    page.on('console', (msg) => { console.log(msg.text()); });
    page.on('pageerror', (exceptionMessage) => { console.log(exceptionMessage); });

    const result = await page.evaluate(async () => {
      try {
        const status = await new Promise((resolve, reject) => {
          const xhr = new XMLHttpRequest();
          xhr.onreadystatechange = () => {
            if (xhr.readyState === 4) {
              if (xhr.status === 200) resolve(`good status: ${xhr.status}`);
              else reject(`bad status: ${xhr.status}`);
            }
          };
          xhr.onerror = () => { reject('xhr error'); };
          xhr.open('GET', 'https://example.org/');
          xhr.send();
        });
        return status;
      } catch (err) {
        return err;
      }
    });
    console.log(result);
  } catch (err) {
    console.error(err);
  }
})();
```

CORS error message is transferred to puppeteer

**What happens instead?**

Only `bad status: 0` in the Node.js console, while this error message in the browser console:

err

Is there a way to catch this error message, while `page.on('console')` and `page.on('pageerror')` seem to not work here?

yanivefraim:
~~WIP (I'm working on docs)~~

Fixes #1939 

**Question**: Do we want to return a [LogEntry](https://vanilla.aslushnikov.com/#Log.LogEntry) object, similar to `ConsoleMessage`? (I guess the answer is yes, just making sure)

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -128,7 +128,10 @@ class Page extends EventEmitter {
 
   _onLogEntryAdded(event) {
     const {level, text, args} = event.entry;
-    this.emit(Page.Events.Console, new ConsoleMessage(level, text, args));
+    if (args)
+      args.map(arg => helper.releaseObject(this._client, arg));
+
+    this.emit(Page.Events.Console, new ConsoleMessage(level, text));
   }
 
   /**
@@ -1054,7 +1057,7 @@ class ConsoleMessage {
    * @param {string} text
    * @param {!Array<*>} args
    */
-  constructor(type, text, args) {
+  constructor(type, text, args = []) {
     this._type = type;
     this._text = text;
     this._args = args;
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -103,7 +103,7 @@ class Page extends EventEmitter {
     client.on('Security.certificateError', event => this._onCertificateError(event));
     client.on('Inspector.targetCrashed', event => this._onTargetCrashed());
     client.on('Performance.metrics', event => this._emitMetrics(event));
-
+    client.on('Log.entryAdded', event => this.emit(Page.Events.LogEntryAdded, event));
     this._target._isClosedPromise.then(() => this.emit(Page.Events.Close));
   }
 
@@ -114,6 +114,9 @@ class Page extends EventEmitter {
     return this._target;
   }
 
+  enableLogEntries() {
+    return this._client.send('Log.enable');
+  }
   /**
    * @return {!Puppeteer.Browser}
    */
@@ -1002,6 +1005,7 @@ Page.Events = {
   FrameNavigated: 'framenavigated',
   Load: 'load',
   Metrics: 'metrics',
+  LogEntryAdded: 'Log.entryAdded'
 };
 
 /**
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -1007,7 +1007,7 @@ Page.Events = {
   FrameDetached: 'framedetached',
   FrameNavigated: 'framenavigated',
   Load: 'load',
-  Metrics: 'metrics'
+  Metrics: 'metrics',
 };
 
 /**
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -50,6 +50,7 @@ class Page extends EventEmitter {
       client.send('Runtime.enable', {}),
       client.send('Security.enable', {}),
       client.send('Performance.enable', {}),
+      client.send('Log.enable', {}),
     ]);
     if (ignoreHTTPSErrors)
       await client.send('Security.setOverrideCertificateErrors', {override: true});
@@ -103,7 +104,7 @@ class Page extends EventEmitter {
     client.on('Security.certificateError', event => this._onCertificateError(event));
     client.on('Inspector.targetCrashed', event => this._onTargetCrashed());
     client.on('Performance.metrics', event => this._emitMetrics(event));
-    client.on('Log.entryAdded', event => this.emit(Page.Events.LogEntryAdded, event));
+    client.on('Log.entryAdded', event => this._onLogEntryAdded(event));
     this._target._isClosedPromise.then(() => this.emit(Page.Events.Close));
   }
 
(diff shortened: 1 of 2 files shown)


aslushnikov:
> Question: Do we want to return a LogEntry object, similar to ConsoleMessage? (I guess the answer is yes, just making sure)

@yanivefraim from user perspective, these all are "console" messages, so I'd rather re-use the "console" event together with "ConsoleMessage" object.

Thanks for checking in.

yanivefraim:
> @yanivefraim from user perspective, these all are "console" messages, so I'd rather re-use the "console" event together with "ConsoleMessage" object.

@aslushnikov - agree, I will changed my implementation.

One more question - when using `LogEntry`, we first have to [enable it](https://vanilla.aslushnikov.com/#Log.enable)
As I see it, there are three options:
1. Enable it by default. This is my [current implementation](https://github.com/GoogleChrome/puppeteer/pull/2400/files#diff-b1397100d3965ca1373667a75e2c5f94R53)
2. Ask the user to enable it, if `console` events are needed (I don't like this option)
3. Have a solution which will enable Logentry only when user did registered to `console` event (not sure if this is possible. I was thinking of using ES6 Proxy for that, but it might make the code too complex). WDYT?


** and another comment - as far as I understand my `args` are not working as expected, yet. I will look for a test case for it

yanivefraim:
I couldn't find a use case where we have `args` on `LogEntry`. Any idea? (I want to add such a case for my test)
cc @aslushnikov

test/page.spec.js
```diff
@@ -340,6 +340,18 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
     });
   });
 
+  describe('Page.Events.Log.entryAdded', function() {
```
aslushnikov: let's move this to other console tests

test/page.spec.js
```diff
@@ -340,6 +340,18 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+    it('should trigger correct Log', async({page, server}) => {
+      await page.goto('about:blank');
+      let message;
+      page.on('console', event => message = event);
+      page.evaluate(async() => fetch('https://example.org/').catch(e => {}));
```
aslushnikov: nit: let's use `server.EMPTY_PAGE`instead of `example.com`

lib/Page.js
```diff
@@ -125,6 +126,11 @@ class Page extends EventEmitter {
     this.emit('error', new Error('Page crashed!'));
   }
 
+  _onLogEntryAdded(event) {
+    const {level, text, args} = event.entry;
+    this.emit(Page.Events.Console, new ConsoleMessage(level, text, args));
```
aslushnikov: args are [`RemoteObjects`](https://vanilla.aslushnikov.com/#Log.LogEntry) here; you'd need to convert these to JSHandles first.

It would be ideal if you can reuse all the logic that's already in [`page._onConsoleAPI`](https://github.com/GoogleChrome/puppeteer/blob/0820d48f804ce924af85fe160ea20a210f734046/lib/Page.js#L421-L448) code.

aslushnikov (commented):
> I couldn't find a use case where we have args on LogEntry. Any idea? (I want to add such a case for my test)

Looks like there's only one client who reports args - [form autofill](https://cs.chromium.org/chromium/src/components/autofill/content/renderer/page_form_analyser_logger.cc?type=cs&sq=package:chromium&l=45). We don't have autofill component in headless, so there's no good way to have a test here. However, if you manage to re-use the `_onConsoleAPI` logic, we'll be fine.

P.S. also check out [cs.chromium.org](https://cs.chromium.org): queries like `-f:out f:inspector entryAdded` can give valuable insights on inner workings of devtools protocol.

lib/Page.js
yanivefraim: @aslushnikov - I tried reusing `page._onConsoleAPI` but I found that it does not make much sense:
 `_onConsoleAPI` initialize `ConsoleMessage`'s text from args while `Log.entryAdded` take the text from `event.text` (when usually there are no args at all). My question is if we do need args for `Log.entryAdded`? I can refactor `_onConsoleAPI` to be compatible for both cases (taking text from args or from event.text), but because I don't have a test case for `Log.entryAdded` with args, I cannot really tell if this use case is working.

WDYT?

test/page.spec.js
yanivefraim: @aslushnikov  - done

test/page.spec.js
yanivefraim: Done

lib/Page.js
aslushnikov: @yanivefraim I see. Let's just disregard `entryAdded` arguments. You can release them like console does.

https://github.com/GoogleChrome/puppeteer/blob/f797f8c307ff5380c750d3851c0b3d61a5299df5/lib/Page.js#L434-L437

I also figured there's no `executionContextId` in the `Log.entryAdded`, so you won't be able to convert these args to JSHandles anyway.

yanivefraim: @aslushnikov - thanks - fixed (I hope that I understood you correct)

BTW, I noticed `event.args.map(arg => helper.releaseObject(this._client, arg));` (in [here](https://github.com/GoogleChrome/puppeteer/blob/f797f8c307ff5380c750d3851c0b3d61a5299df5/lib/Page.js#L434-L437)) does not wait for promise to be resolved. I assume it is because we do not care about the response? (just making sure)

aslushnikov: > BTW, I noticed event.args.map(arg => helper.releaseObject(this._client, arg)); (in here) does not wait for promise to be resolved. I assume it is because we do not care about the response? (just making sure)

That's right

aslushnikov (approved):

Issue #2079, opened 2018-02-22
v2.0 Breaking Changes


aslushnikov:
These are the breaking changes we'd like to have for v2.0:
- [ ] ~~set initial viewport size to 1280 x 1024 - https://github.com/GoogleChrome/puppeteer/issues/1910~~ (not doing this for 2.0 - doesn't seem worth the hassle)
- [ ] make `browser.disconnect` async: sockets don't close instantly.
- [ ] make `page.evaluate` to throw for the non-serializable values (e.g. objects with circular references) #2864
- [ ] refactor `response.securityDetails()` into `response.certificate()` and actually include certificate data (DEPRECATE OLD ONE)
- [ ] ~~consider switching to pipes instead of websocket connection by default~~ (not doing this - not worth the hassle)
- [x] consider dropping node6 support (https://github.com/GoogleChrome/puppeteer/pull/5045)
- [ ] consider renaming `waitForSelector`'s "visible" option into "interactable" since we also have a real visibility with IntersectionObserver. https://github.com/GoogleChrome/puppeteer/pull/2673 
- [ ] consider passing calling frame as the first argument to the function exposed via `page.exposeFunction`.  #2126  (BETTER DO A `frame.exposeFunction` to expose per-frame)
- [x] use single "defaultTimeout" option to setup default timeout for *all* methods. The option should support `0` to help with debugging. #3158 
- [x] move to flatten protocol. https://github.com/GoogleChrome/puppeteer/pull/3524
- [ ] move `page.tracing` to `browser.tracing`; pass a page to capture screenshots of as a method argument. https://github.com/GoogleChrome/puppeteer/issues/3639 (DEPRECATE THE OLD API)
- [x] make `page.screenshot` clip elements to the viewport per upstream Chromium changes. This matches the clipping behavior of elements in inner scrollers, i.e. document and overflow scroll clipping now work the same.
- [x] rename `page.emulateMedia` to `page.emulateMediaType` (DEPRECATE THE OLD API). This is technically not a breaking chance since we'll keep an alias for now. #5012

yanivefraim:
Can I please take:

> make page.evaluate return puppeteer.UNSERIALIZABLE_VALUE for the non-serializable values (e.g. objects with circular references)

and 

> set initial viewport size to 1280 x 1024

?

aslushnikov:
> Can I please take:

[...]

PR #2425, opened 2018-04-20
 feature(Page): evaluate should return a special message in case of non serialisable returned value

### LINKED ISSUE #2420: feature request: throw an error in case of page.evaluate returning non-Serializable value
follow-up for https://github.com/GoogleChrome/puppeteer/issues/2418:

Today, `page.evaluate` returns `undefined` in case of a non-Serializable returned object (see [docs](https://github.com/GoogleChrome/puppeteer/blob/v1.3.0/docs/api.md#pageevaluatepagefunction-args)). It could be nice to have an error for such case (it is not intuitive to have undefined returned here, IMO)

yanivefraim:
This PR is a followup for the following discussion: https://github.com/GoogleChrome/puppeteer/issues/2418#issuecomment-383156756

I wasn't sure which message do we want for each scenario, and how exactly do we want to implement the entire feature (if we want it at all...). So it is a start - we can discuss it from here.

it will fix #2420

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -209,6 +209,10 @@ class Helper {
     apiCoverage = new Map();
   }
 
+  static getNonSerializableMessage() {
+    return 'Non serializable object returned from evaluate';
+  }
+
   /**
    * @param {!Object} obj
    * @return {boolean}
--- a/docs/api.md
+++ b/docs/api.md
@@ -825,7 +825,7 @@ List of all available devices is available in the source code: [DeviceDescriptor
 
 If the function passed to the `page.evaluate` returns a [Promise], then `page.evaluate` would wait for the promise to resolve and return its value.
 
-If the function passed to the `page.evaluate` returns a non-[Serializable] value, then `page.evaluate` resolves to `undefined`.
+If the function passed to the `page.evaluate` returns a non-[Serializable] value, then `page.evaluate` resolves to `Non serializable object returned from evaluate` message.
 
 Passing arguments to `pageFunction`:
 ```js
@@ -1802,7 +1802,7 @@ Gets the full HTML contents of the frame, including the doctype.
 
 If the function passed to the `frame.evaluate` returns a [Promise], then `frame.evaluate` would wait for the promise to resolve and return its value.
 
-If the function passed to the `frame.evaluate` returns a non-[Serializable] value, then `frame.evaluate` resolves to `undefined`.
+If the function passed to the `frame.evaluate` returns a non-[Serializable] value, then `frame.evaluate` resolves to `Non serializable object returned from evaluate` message.
 
 ```js
 const result = await frame.evaluate(() => {
--- a/lib/ExecutionContext.js
+++ b/lib/ExecutionContext.js
@@ -46,9 +46,9 @@ class ExecutionContext {
     const handle = await this.evaluateHandle(pageFunction, ...args);
     const result = await handle.jsonValue().catch(error => {
       if (error.message.includes('Object reference chain is too long'))
-        return;
+        return helper.getNonSerializableMessage();
       if (error.message.includes('Object couldn\'t be returned by value'))
-        return;
+        return helper.getNonSerializableMessage();
       throw error;
     });
     await handle.dispose();
--- a/test/page.spec.js
+++ b/test/page.spec.js
@@ -16,6 +16,7 @@
 const fs = require('fs');
 const path = require('path');
 const utils = require('./utils');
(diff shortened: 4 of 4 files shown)


benjamingr (changes requested):
I don't think returning a predefined string is a good way to do this - puppeteer should probably `reject` on that.

yanivefraim:
> I don't think returning a predefined string is a good way to do this - puppeteer should probably reject on that.

@benjamingr  - yes, we already had this discussion [here](https://github.com/GoogleChrome/puppeteer/issues/2418#issuecomment-383156756). I also agree an error is better here

lib/helper.js
```diff
@@ -209,6 +209,10 @@ class Helper {
     apiCoverage = new Map();
   }
 
+  static getNonSerializableMessage() {
+    return 'Non serializable object returned from evaluate';
```
aslushnikov, line 213: instead of returning a string, this should be a [Symbol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol) exposed on the `puppeteer` namespace. This way you'll be able to do the following:

```js
const result = await page.evaluate(() => window);
if (result === puppeteer.NON_SERIALIZABLE_OBJECT) {
  // ...
}
```

aslushnikov (commented):
> I don't think returning a predefined string is a good way to do this - puppeteer should probably reject on that.

@benjamingr We were trying to reject before and got a pushback from our clients. Users shouldn't be afraid to evaluate any expression, regardless of its return value.

What would be the usecase to justify this change?
- There's always a `page.evaluateHandle` that returns a handle for the object, if needed.
- The "serializability" of the object could be checked easily, trying to `JSON.stringify` it manually.

cc @Janpot  who had thoughts on this topic as well.

Janpot:
> who had thoughts on this topic as well.

w.r.t. API design, I have to agree with @benjamingr. But @aslushnikov explained to me that the current approach was a pragmatic choice as many users seem to want to do things  like:
```js
await page.evaluate(() => window.foo = window);
```
Personally, I have no idea what's the use-case behind that code, but I think we could educate people to instead do:
```js
await page.evaluate(() => {
  window.foo = window;
});
```
Maybe the error message can include a link to a documentation page on github that explains this pattern? 
I know my team has ran into bugs caused by the current behavior more than once. All in all, this approach solves it for us and if `puppeteer` devs want to take the pragmatic approach, I have no strong objections against that. We can always wrap the `evaluate` call by one that takes a, for us, more sensible approach.

aslushnikov:
> I know my team has ran into bugs caused by the current behavior more than once. 

@Janpot can you please give a few examples?

benjamingr:
>  the current approach was a pragmatic choice as many users seem to want to do things like:

That's an interesting point. In general I think seralization could use some love.

Janpot:
>can you please give a few examples?

We were running an external library in `.evaluate` that under some circumstances returned a nested object that had a cyclic reference somewhere.

```js
await page.evaluate(() => {
  const obj = {};
  obj.a = obj;
  return obj;
});
// undefined
```

I think it's very reasonable to assume `.evaluate` either returns a result or throws an error. But there's a class of errors that are magically encoded in a third state, being `undefined`. There is also the case that `undefined` might be an expected result of `.evaluate` and then there is no way to differentiate between the legit `undefined` and the one that denotes an error.

And then there was also this issue https://github.com/GoogleChrome/puppeteer/issues/2021.

I think APIs should be intuitive and having magical third states is not intuitive I think.

benjamingr:
@Janpot just so we're clear - there is no hard limitation meaning puppeteer can't deal with returning recursive references, object methods, complex values (like buffers) or other things - only the fact the code for it hasn't been written in

I suggest we fix that instead of the error messages

Janpot:
I am familiar with the limitations  🙂. I only proposed the error message as a compromise since there seems to be concern users won't understand why their code is broken.
If it were my choice, the `evaluate` API would either, resolve with the expected result, or reject when it can't, including unserializable values.

aslushnikov:
> I think it's very reasonable to assume .evaluate either returns a result or throws an error. But there's a class of errors that are magically encoded in a third state, being undefined. There is also the case that undefined might be an expected result of .evaluate and then there is no way to differentiate between the legit undefined and the one that denotes an error.

This ambiguity in evaluation results is indeed uncomfortable; throwing an error would be more logical here.

Postponing this until 2.0 since this will be a breaking change for our clients.
Please, follow #2079 for the future updates.

PR #2428, opened 2018-04-22
fix(Launcher): exit the process after killing chrome on SIGINT

ranyitz:
👋 Hey

This is a fix for this issue https://github.com/facebook/jest/issues/5748

In the issue, `puppeteer.launch` method is called from jest `globalSetup` script which [is running on the same process as jest](https://github.com/facebook/jest/blob/6a77ee37ec2d46ece7e9cfd0f891d11c113cc4c4/packages/jest-cli/src/run_jest.js#L242).

Puppeteer catches `SIGINT` in order to kill chrome, but it also replaces the default behaviour of `SIGINT`, [which is to exit with status code 130](https://nodejs.org/dist/latest-v9.x/docs/api/all.html#process_signal_events). In jest's case, the process does not exit, which keeps the process hanging in a weird way.

This PR makes sure that the process exits after killing chrome.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -143,7 +143,7 @@ class Launcher {
 
     const listeners = [ helper.addEventListener(process, 'exit', killChrome) ];
     if (options.handleSIGINT !== false)
-      listeners.push(helper.addEventListener(process, 'SIGINT', killChrome));
+      listeners.push(helper.addEventListener(process, 'SIGINT', () => { killChrome(); process.exit(130); }));
     if (options.handleSIGTERM !== false)
       listeners.push(helper.addEventListener(process, 'SIGTERM', gracefullyCloseChrome));
     if (options.handleSIGHUP !== false)


ranyitz:
I signed it!

aslushnikov:
> Puppeteer catches SIGINT in order to kill chrome, but it also replaces the default behaviour of SIGINT, which is to exit with status code 130. In jest's case, the process does not exit, which keeps the process hanging in a weird way.

@ranyitz why doesn't jest launch pptr with `handleSIGINT: false`? This way it can handle signal on its side in any way it wants.

ranyitz:
@aslushnikov Because when `handleSIGINT: false` is configured, chromium stays open after a `ctrl+c` in the middle of the tests. The launcher usually kills it on SIGINT.

aslushnikov:
@ranyitz right, if you do `handleSIGINT: false`, then you'll need to close the browser manually via `browser.close()`.

ranyitz:
@aslushnikov Thanks! This is definitely working as expected. :ok_hand:

The only thing that I have in mind is that all the users that want to setup pptr with jest need to configure it explicitly in their setup script.

```js
const browser = await puppeteer.launch({handleSIGINT: false})
process.on('SIGINT', () => { browser.close(); process.exit(130); })
```

It feels to me like it's generic enough to be in pptr, so the users would not need to configure it, is there any reason not to do it there?

JoelEinbinder (approved):

aslushnikov:
> It feels to me like it's generic enough to be in pptr, so the users would not need to configure it, is there any reason not to do it there?

@ranyitz thanks, I see now. This is reasonable.

itaysabato:
I feel I should mention this was a breaking change for me - my code already had a handler for `SIGINT` that prevented the process from exiting until some cleanup was made, but once I upgraded puppeteer from `1.3.0` to `1.6.0` it started to exit immediately.

It took me a long while to track down the issue to puppeteer and eventually to this change - but `handleSIGINT: false` solved it.

While this behavior may make sense for a process that only runs puppeteer, if you assume (as you should IMO) that the parent process is doing other stuff as well, you should never call `process.exit`.

LLAP 🖖

aslushnikov:
> I feel I should mention this was a breaking change for me

@itaysabato thanks for writing us. I'm sorry you spent a lot of time figuring this out.

> While this behavior may make sense for a process that only runs puppeteer, if you assume (as you should IMO) that the parent process is doing other stuff as well, you should never call process.exit.

That's fair.  @ranyitz: there's now https://github.com/smooth-code/jest-puppeteer that helps dealing with Puppeteer in jest. Do you think we can now revert the change and contribute to jest-puppeteer instead?

ranyitz:
@itaysabato I'm also sorry that you've spent a lot of time because of this change.

@aslushnikov using pptr with jest may be a common use case, but we'll need to implement this into other tests runners as well.

> 'SIGINT' have default handlers on non-Windows platforms that reset the terminal mode before exiting with code 128 + signal number. If one of these signals has a listener installed, its default behavior will be removed (Node.js will no longer exit).

(https://nodejs.org/api/process.html#process_signal_events)

If pptr would not set up a SIGINT listener, the process would exit after `ctrl+c`, but we have to, in order to kill the chrome instance.

I think that it makes sense to preserve the default behavior. If someone wants to configure custom behavior on SIGINT event by himself he can use `{handleSIGINT: false}`.
_____

The problematic part here is that people that configure a SIGINT handler, use to the fact that they can control the exit event themselves (and we've changed that) but users that not configure SIGINT use to the fact that the process exits after SIGINT.

We can't win them all but we should go with what most users would want. I think that there are fewer people that set up SIGINT custom handlers and that they could easily use `{handleSIGINT: false}`.

We can add a note about the fact that we're configuring a SIGINT handler in pptr in the main README, so it will be easier to find out about it.

itaysabato:
Just an idea:

Is it possible to check for existing `SIGINT` listeners so that if there are any you defer to them instead of existing? If so it seems like best of both worlds to me.

If you can check this at the time of the signal then you could even spot listeners who subscribed after you.

aslushnikov:
> @aslushnikov using pptr with jest may be a common use case, but we'll need to implement this into other tests runners as well.

@ranyitz ah right. Thanks for brushing up my memory on this topic. 

> Is it possible to check for existing SIGINT listeners so that if there are any you defer to them instead of existing? If so it seems like best of both worlds to me.

@itaysabato This might work, but sounds quite magical to me from the API standpoint. Do you know if any node.js project employs this strategy? I wonder what's their milage. All the projects I'm aware of are using the `handeSIGINT: false` approach.

PR #2437, opened 2018-04-24
Fix code highlighting / formatting in README

piperchester:


ebidel:
We actually need this to stay as is. We consume this .md file on developers.google.com/web and adding the gh flavored syntax highlighting messes that up. See https://github.com/GoogleChrome/puppeteer/commit/9b06ed28ae72ee3aa6de1f8b534f51347320b4a6#diff-04c6e90faac2675aa89e2176d2eec7d8

PR #2451, opened 2018-04-26
test: verify file url interception works as expected

### LINKED ISSUE #1506: page content incorrectly embedded when using file protocol
Hi again,
As usual, thanks a lot for maintaining puppeteer - it's super useful. 👍 

Seems to me that there was a regression in `0.13` 🐛  when using the combination of:
- `file:///` protocol
- `setRequestInterception: true`

### Steps to reproduce

* Puppeteer version: `0.13.0`
* Platform / OS version: `MacOS High Sierra`, same on `windows`
* URLs (if applicable):
Any url over the `file` protocol seems to have this problem.

```js
(async() => {
      // waitUntil: 'networkidle2', // see if this performs better than default for timeouts
      ignoreHTTPSErrors: true,
      args: ['--disable-setuid-sandbox', '--no-sandbox']
    })

// CRITICAL: _only_ with setRequestInterception = true do we get this problem:
// the real html is escaped and embedded in a template like below.
// <html><head></head><body><pre style="word-wrap: break-word; white-space: pre-wrap;">
// <REAL CONTENT HERE>
// </pre></body></html>
page.on('request', interceptedRequest => {
  console.log(interceptedRequest.url)
  interceptedRequest.continue()
})

// ANY local html file will do
await page.goto('file:///Users/pocketjoso/penthouse/test/static-server/yeoman.html')

const content = await page.content()
console.log('content: ' + content)

browser.close();
})();

```
The html is parsed in the same way regardless of file/http protocol, and regardless of whether `setRequestInterception` is true or false. 

**What happens instead?**
```
content: <html><head></head><body><pre style="word-wrap: break-word; white-space: pre-wrap;">&lt;!doctype html&gt;
&lt;html class="no-js"&gt;
    &lt;head&gt;
        &lt;meta charset="utf-8"&gt;
        &lt;title&gt;critical css test&lt;/title&gt;
        &lt;meta name="description" content=""&gt;
        &lt;meta name="viewport" content="width=device-width, initial-scale=1"&gt;
        &lt;!-- Place favicon.ico and apple-touch-icon.png in the root directory --&gt;
... etc
```
Or in a picture (the browser renders the markup as text, rather than as html):
bjor6laez-after

aslushnikov:
This was fixed upstream: https://crrev.com/550319
Rolled into pptr: https://github.com/GoogleChrome/puppeteer/pull/2393

Fixes #1506.

### DIFF
--- a/test/network.spec.js
+++ b/test/network.spec.js
@@ -459,6 +459,18 @@ module.exports.addTests = function({testRunner, expect}) {
       await page.goto(server.EMPTY_PAGE);
       expect(error.message).toContain('Request Interception is not enabled');
     });
+    it('should work with file URLs', async({page, server}) => {
+      await page.setRequestInterception(true);
+      const urls = new Set();
+      page.on('request', request => {
+        urls.add(request.url().split('/').pop());
+        request.continue();
+      });
+      await page.goto('file://' + path.join(__dirname, 'assets', 'one-style.html'));
+      expect(urls.size).toBe(2);
+      expect(urls.has('one-style.html')).toBe(true);
+      expect(urls.has('one-style.css')).toBe(true);
+    });
   });
 
   describe('Request.respond', function() {
--- a/test/network.spec.js
+++ b/test/network.spec.js
@@ -466,7 +466,7 @@ module.exports.addTests = function({testRunner, expect}) {
         urls.add(request.url().split('/').pop());
         request.continue();
       });
-      await page.goto('file://' + path.join(__dirname, 'assets', 'one-style.html'));
+      await page.goto(pathToFileURL(path.join(__dirname, 'assets', 'one-style.html')));
       expect(urls.size).toBe(2);
       expect(urls.has('one-style.html')).toBe(true);
       expect(urls.has('one-style.css')).toBe(true);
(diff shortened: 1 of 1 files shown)


test/network.spec.js
```diff
@@ -459,6 +459,18 @@ module.exports.addTests = function({testRunner, expect}) {
...
       expect(error.message).toContain('Request Interception is not enabled');
     });
+    it('should work with file URLs', async({page, server}) => {
+      await page.setRequestInterception(true);
+      const urls = new Set();
+      page.on('request', request => {
+        urls.add(request.url().split('/').pop());
+        request.continue();
+      });
+      await page.goto('file://' + path.join(__dirname, 'assets', 'one-style.html'));
```
JoelEinbinder: Path.join will have the wrong way slashes on windows

aslushnikov: I wonder how did it pass on appveyor.

JoelEinbinder: I think Chrome auto corrects slashes. A bit iffy to rely on that though.

aslushnikov: @JoelEinbinder agreed. Translated the path to URL on our end.

JoelEinbinder (approved):

PR #2456, opened 2018-04-26
feat: add mouse.wheel method

jiladahe1997:
add the mouseWheel event method . This method could scroll the page to a specific coordinate .

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -126,6 +126,7 @@
   * [mouse.down([options])](#mousedownoptions)
   * [mouse.move(x, y, [options])](#mousemovex-y-options)
   * [mouse.up([options])](#mouseupoptions)
+  * [mouse.wheel(deltaX, deltaY)](#mousewheeldeltax-deltay)
 - [class: Touchscreen](#class-touchscreen)
   * [touchscreen.tap(x, y)](#touchscreentapx-y)
 - [class: Tracing](#class-tracing)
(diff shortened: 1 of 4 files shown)


docs/api.md
```diff
@@ -1589,6 +1590,13 @@ Dispatches a `mousemove` event.
 
 Dispatches a `mouseup` event.
 
+#### mouse.wheel(deltaX, deltaY)
+- `deltaX` <[number]>
+- `deltaY` <[number]>
+- returns: <[Promise]>
+
+Dispatches a `mousewheel` event. Please note that it will take a very short time to scroll the page, so maybe you need use Page.waitFor(). The method Page.screenshot will scroll the page to 0, 0.
```
vsemozhetbyt, line 1598: Not sure, but maybe:
`` Page.waitFor() `` -> `` `page.waitFor()` ``
`` Page.screenshot `` -> `` `page.screenshot()` ``

aslushnikov:
@JoelEinbinder can you please take a look?

JoelEinbinder (commented):
I don’t think we should add this until we properly wait for the scrolling on the back end

aslushnikov:
@JoelEinbinder do you have an upstream bug we can refer to here?

aslushnikov:
DevTools protocol method for dispatching mouse wheel events is broken now: it needs to wait for the scrolling to finish before returning.

This has to be fixed upstream.

PR #2468, opened 2018-04-28
fix(Page): fix page.goto to return Response when page pushes new state

### LINKED ISSUE #2411: Page.goto returns null for some URLs since version 1.3.0
### Steps to reproduce

* Puppeteer version: 1.3.0
* Platform / OS version: Mac OSX
* Node.js version: v7.10.1 / v9.8.0
* URLs (if applicable):
  * https://remarkjs.com/
  * http://lab.hakim.se/reveal-js
  * https://leaverou.github.io/csss/
  * http://artificer.jboss.org/slides/general/opensource-getting-involved.html
  * http://joshbode.github.io/remark/ansi.html

```

(async () => {
  response = await page.goto('https://remarkjs.com/');

  console.log('response:', response)

  await browser.close();
})();
```

For the URLs listed above.

`response` should not be null.

**What happens instead?**

`response` is null. Note that it works with Puppeteer version 1.2.0.

aslushnikov:
Fixes #2411.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -774,6 +774,8 @@ class Frame {
    */
   _navigated(framePayload) {
     this._name = framePayload.name;
+    // TODO(lushnikov): remove this once requestInterception has loaderId exposed.
+    this._navigationURL = framePayload.url;
     this._url = framePayload.url;
   }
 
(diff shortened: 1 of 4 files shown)


JoelEinbinder (approved):
Makes sense.

ntzm:
Does this also fix #2479 @aslushnikov?

aslushnikov:
@ntzm it doesn't.

ntzm:
No worries, thanks for all your hard work!

PR #2475, opened 2018-04-29
feat(Connection): Enable timeout to wait for commands

### LINKED ISSUE #2473: Enable timeout to wait for commands
Im using a single instance of chromium and run on it lots of tests over time and concurrently using incognito tabs. After sometime, the calls to the web socket starts to hang and there is no way of knowing that, causing my flow to get stuck without a way out.
There is a need to implement a timeout mechanism for the requests as they happen in an async matter.
My suggestion would be to add In the Connection class a timeout after performing the send, which will reject the promise after the timeout...
Currently my solution is to perform a regular web socket new connection to chromium to make sure it is still responsive but that doesn't help if the tab itself gets stuck...

regevbr:
Fixes #2473

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -301,6 +301,7 @@ puppeteer.launch().then(async browser => {
   - `browserWSEndpoint` <[string]> a [browser websocket endpoint](#browserwsendpoint) to connect to.
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
   - `slowMo` <[number]> Slows down Puppeteer operations by the specified amount of milliseconds. Useful so that you can see what is going on.
+  - `commandTimeout` <[number]> Maximum time in milliseconds to wait for command response. Defaults to `0`. Pass `0` to disable timeout.
 - returns: <[Promise]<[Browser]>>
 
 This methods attaches Puppeteer to an existing Chromium instance.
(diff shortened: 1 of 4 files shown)


regevbr:
I signed it!

regevbr:
@JoelEinbinder or @aslushnikov  can you please code review this?

aslushnikov:
@regevbr I'm not sure I follow: why does websocket hang for you? 

WebSocket should never hang, but this PR makes it look like this kind of behavior is legitimate, even exposing an API bit to control it.

regevbr:
@aslushnikov I don't know why it hangs and it is very hard to investigate it but the fact of the matter is that it happens, and once I started using my forked version, it stopped happening...
Also, I'm not syaing that the web socket is hanging, im saying that after you send a message, the response never comes back, causing the promise to never be settled...

aslushnikov:
>  I don't know why it hangs and it is very hard to investigate it but the fact of the matter is that it happens, and once I started using my forked version, it stopped happening...
Also, I'm not syaing that the web socket is hanging, im saying that after you send a message, the response never comes back, causing the promise to never be settled...

I see, thanks for the response.
This approach is a no-go for us: it *hides* the problem rather then fixes it. It could also be a problem in your environment.

Let's continue the discussion in the bug.

regevbr:
@aslushnikov The issue seems to come from the chromium side and not from this library... How do you suggest handling non responsive tabs?

aslushnikov:
@regevbr I'd start with buliding a reliable repro case and then recording a [trace](https://www.chromium.org/developers/how-tos/trace-event-profiling-tool).

PR #2478, opened 2018-04-30
feat(Page): add option to run 'beforeunload' when closing the page

### LINKED ISSUE #2386: onbeforeunload not called when page.close() called
### Steps to reproduce

* Puppeteer version: 1.2
* Platform / OS version: Mac 10.13.2
* URLs (if applicable):
* Node.js version: 8.11.1

I need to release some resources when page.close called.  Code below called when I close the tab manually.  But in puppet, after I invoke page.close(). It seems not get called at all.

```
window.onbeforeunload = function() {
  $log.debug('MainController onbeforeunload');
  rtc.disconnect();
};
```

**What happens instead?**

aslushnikov:
Today, `page.close()` method doesn't run page's beforeunload listeners.
This way users can be sure that `page.close()` actually closes the
page.

This patch adds an optional `runBeforeUnload` option to the
`page.close()` method that would run beforeunload listeners. Note:
running beforeunload handlers might cancel page closing.

Fixes #2386.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -69,7 +69,7 @@
   * [page.bringToFront()](#pagebringtofront)
   * [page.browser()](#pagebrowser)
   * [page.click(selector[, options])](#pageclickselector-options)
-  * [page.close()](#pageclose)
+  * [page.close(options)](#pagecloseoptions)
   * [page.content()](#pagecontent)
   * [page.cookies(...urls)](#pagecookiesurls)
   * [page.coverage](#pagecoverage)
@@ -747,8 +747,17 @@ const [response] = await Promise.all([
 
 Shortcut for [page.mainFrame().click(selector[, options])](#frameclickselector-options).
 
-#### page.close()
+#### page.close(options)
+- `options` <[Object]>
+  - `runBeforeUnload` <[boolean]> Defaults to `false`. Whether to run the
+    [before unload](https://developer.mozilla.org/en-US/docs/Web/Events/beforeunload)
+    page handlers.
 - returns: <[Promise]>
+
+By default, `page.close()` **does not** run beforeunload handlers.
+
+> **NOTE** if `runBeforeUnload` is passed as true, a `beforeunload` dialog might be summoned
+> and should be handled manually via page's ['dialog'](#event-dialog) event.
 
 #### page.content()
 - returns: <[Promise]<[String]>>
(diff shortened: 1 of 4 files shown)


lib/Page.js
```diff
@@ -823,10 +823,18 @@ class Page extends EventEmitter {
     return this.mainFrame().title();
   }
 
-  async close() {
+  /**
+   * @param {!{runBeforeUnload: boolean}} options
```
JoelEinbinder: add a ?, so runBeforeUnload?: boolean, to indicate an optional property.

docs/api.md
```diff
@@ -748,8 +748,15 @@ const [response] = await Promise.all([
 Shortcut for [page.mainFrame().click(selector[, options])](#frameclickselector-options).
 
 #### page.close()
+- `options` <[Object]>
+  - `runBeforeUnload` <[boolean]> Defaults to `false`. Whether to run the before unload page handlers.
```
JoelEinbinder: I feel like this is a bit confusing, because before Unload handlers might still not run if there was no user interaction.

docs/api.md
```diff
@@ -748,8 +748,15 @@ const [response] = await Promise.all([
...
 - returns: <[Promise]>
 
+By default, `page.close()` **does not** run beforeunload handlers.
+
+> **NOTE** if `runBeforeUnload` is passed as true, `beforeunload` dialog might be summoned
```
JoelEinbinder: ...true, a ‘beforeunload’ dialog...

docs/api.md
JoelEinbinder: Can we get a link to mdn on beforeunload here?

docs/api.md

docs/api.md
aslushnikov: > I feel like this is a bit confusing, because before Unload handlers might still not run if there was no user interaction.

I agree. However, there are many different heuristics in place, and they keep changing. I don't think that has anything to do with pptr or our automation API - that's how WebPlatform rolls.

> Can we get a link to mdn on beforeunload here?

Done.

lib/Page.js
aslushnikov: I did the `(boolean|undefined)` to be consistent with other parts of our codebase regarding this.

aslushnikov:
@JoelEinbinder all done; please take another look.

JoelEinbinder (approved):

Issue #2490, opened 2018-05-02
PUPPETEER_CHROMIUM_REVISION not used when launching chromium


benperiton:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.3.0
* Platform / OS version: OSX 10.11.6
* URLs (if applicable):
* Node.js version: 10.0.0

**What steps will reproduce the problem?**

_Please include code that reproduces the issue._

1. export PUPPETEER_CHROMIUM_REVISION=526987 && npm install

**What is the expected result?**
When running puppeteer it should use the version specified in the ENV variable

**What happens instead?**
it tries to use the version specified in the package.json

aslushnikov:
Closing this given the comment in the relevant pull request: https://github.com/GoogleChrome/puppeteer/pull/2491#issuecomment-386152557

PR #2491, opened 2018-05-02
fix(Launcher): PUPPETEER_CHROMIUM_REVISION not used when launching chromium #2490

benperiton:
fix(Launcher): PUPPETEER_CHROMIUM_REVISION not used when launching chromium #2490

Check for PUPPETEER_CHROMIUM_REVISION when launching the browser, can be used during installation

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -23,7 +23,7 @@ const Browser = require('./Browser');
 const readline = require('readline');
 const fs = require('fs');
 const {helper, debugError} = require('./helper');
-const ChromiumRevision = require(path.join(helper.projectRoot(), 'package.json')).puppeteer.chromium_revision;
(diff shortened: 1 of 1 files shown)


benperiton:
I signed it!

aslushnikov:
We refrain from using env variables in the `lib/` code as much as possible. `env` doesn't scope well: if there's a few pptr's in the dependency tree of your project, you'll infect them all. Of course, a couple of pptr's in the dependency tree is a bit of a fantasy today, but will be quite real once we have `puppeteer-core` package  (see https://github.com/GoogleChrome/puppeteer/issues/1809).

It's not much overhead to explicitly specify needed revision:
```js
const browserFetcher = puppeteer.createBrowserFetcher();
const revision = browserFetcher.revisionInfo(process.env.PUPPETEER_CHROMIUM_REVISION);
const browser = await puppeteer.launch({executablePath: revisionInfo.executablePath})
```

Issue #2439, opened 2018-04-24
Target: pass through "background_page" types


ebidel:
Related to https://github.com/GoogleChrome/puppeteer/issues/2438.

The lighthouse chrome extension tests [check](https://github.com/GoogleChrome/puppeteer/blob/master/lib/Target.js#L60) a private `target._targetInfo` property to verify the target is the LH background page:

```js
const targets = await browser.targets();
const extensionTarget = targets.find(({_targetInfo}) => {
  return _targetInfo.title === 'Lighthouse' && _targetInfo.type === 'background_page';
});
```

We should pass through this type through so `target.type()` works as expected. We currently mark it as "other":

https://github.com/GoogleChrome/puppeteer/blob/e9dc3ae74fc01a99d1d5b66ffe716b4de4aa39e8/lib/Target.js#L58-L63

aslushnikov:
This is fixed by https://github.com/GoogleChrome/puppeteer/pull/2493.

PR #2493, opened 2018-05-02
feat(Target):  Pass through "background_page" types (#2439)

Runaider:


Runaider:
I signed it!

ebidel (approved):

test/target.spec.js
```diff
@@ -36,6 +36,12 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(allPages).toContain(page);
       expect(allPages[0]).not.toBe(allPages[1]);
     });
+    it('should allow background_page target type to pass through', async({browser}) => {
+      const targets = browser.targets();
+      targets[0]._targetInfo.type = 'background_page';
```
aslushnikov: let's be real: can you please create a minimalistic chrome extension under `/test/assets/simple-extension` and load it in the headful browser?

Runaider: Yes, will do it later today

docs/api.md
```diff
@@ -2595,7 +2595,7 @@ If the target is not of type `"page"`, returns `null`.
 #### target.type()
 - returns: <[string]>
 
-Identifies what kind of target this is. Can be `"page"`, `"service_worker"`, `"browser"` or `"other"`.
```
JoelEinbinder, line 2598: As this list grows, we should probably provide some explaining links as to what these types are: [`"background_page"`](https://developer.chrome.com/extensions/background_pages)

JoelEinbinder:
Thanks for the patch, looks good!

test/target.spec.js
Runaider: I have updated the test to use a headfull browser with a bare extension. I'm not sure about starting a new browser instance in a spec that already accepts a headless browser instance, Should I leave the test where it is or move it to another spec?

aslushnikov (approved):

PR #2503, opened 2018-05-04
fix: allow user gesture restricted code to be run in page.evaluate

### LINKED ISSUE #2502: Page.evaluate should bypass user gestures
`await page.evaluate('document.documentElement.webkitRequestFullscreen()')` currently does not work because full screen requires a user gesture. Running the same snippet in the DevTools console does work. 

Presumably devtools bypasses user gestures by adding [`userGesture: true`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime#method-evaluate). We can do the same for our `Runtime.evaluate` and `Runtime.callFunctionOn` calls.

JoelEinbinder:
Fixes #2502

### DIFF
--- a/lib/ExecutionContext.js
+++ b/lib/ExecutionContext.js
@@ -75,7 +75,8 @@ class ExecutionContext {
       executionContextId: this._contextId,
       arguments: args.map(convertArgument.bind(this)),
       returnByValue: false,
-      awaitPromise: true
+      awaitPromise: true,
+      userGesture: true
     });
     if (exceptionDetails)
       throw new Error('Evaluation failed: ' + helper.getExceptionMessage(exceptionDetails));
--- a/lib/ExecutionContext.js
+++ b/lib/ExecutionContext.js
@@ -64,7 +64,13 @@ class ExecutionContext {
     if (helper.isString(pageFunction)) {
       const contextId = this._contextId;
       const expression = /** @type {string} */ (pageFunction);
-      const { exceptionDetails, result: remoteObject } = await this._client.send('Runtime.evaluate', { expression, contextId, returnByValue: false, awaitPromise: true});
+      const { exceptionDetails, result: remoteObject } = await this._client.send('Runtime.evaluate', {
+        expression,
+        contextId,
+        returnByValue: false,
+        awaitPromise: true,
+        userGesture: true
+      });
       if (exceptionDetails)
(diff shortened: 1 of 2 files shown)


test/page.spec.js
```diff
@@ -196,6 +196,13 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
       const isFive = await page.evaluate(e => Object.is(e, 5), aHandle);
       expect(isFive).toBeTruthy();
     });
+    it('should simulate a user gesture', async({page, server}) => {
+      await page.evaluate(() => {
+        const audio = document.createElement('audio');
+        audio.src = 'data:audio/wav;base64,UklGRiQAAABXQVZFZm10IBAAAAABAAEARKwAAIhYAQACABAAZGF0YQAAAAA=';
+        return audio.play();
```
JoelEinbinder: This returns a promise, which throws if the audio fails to play.

lib/ExecutionContext.js
```diff
@@ -75,7 +75,8 @@ class ExecutionContext {
       executionContextId: this._contextId,
       arguments: args.map(convertArgument.bind(this)),
       returnByValue: false,
-      awaitPromise: true
+      awaitPromise: true,
+      userGesture: true
```
ebidel, line 85: Do we need this up on 67 too for the `Runtime.evaluate` call?

test/page.spec.js
ebidel: I'd add that comment. It's not obvious unless you know about our autoplay policy.

test/page.spec.js
```diff
@@ -196,6 +196,13 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
       const isFive = await page.evaluate(e => Object.is(e, 5), aHandle);
       expect(isFive).toBeTruthy();
     });
+    it('should simulate a user gesture', async({page, server}) => {
+      await page.evaluate(() => {
```
ebidel: add a test of for the string case: page.evaluate('...')

lib/ExecutionContext.js
JoelEinbinder: Good catch. done.

test/page.spec.js
JoelEinbinder: done.

test/page.spec.js
JoelEinbinder: done

ebidel (approved):

PR #2552, opened 2018-05-16
docs(README.md): add badge to display install size

styfle:
This adds a badge to the `README.md` file to display the npm install size

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 # Puppeteer
 
 <!-- [START badges] -->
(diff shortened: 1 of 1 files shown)


aslushnikov:
Install sizes are different on different platforms - which one is this? We should either have all of them or neither.

styfle:
Good point! 👍

Currently, Package Phobia runs on Zeit Now which appears to be Linux.

I tested on Mac OS just now and the size was roughly 186 MB so that is much smaller than the size reported by Package Phobia (286 MB).

I'll have to think about how to make Package Phobia run multiple Archs and how that will affect the GUI 🤔

Issue #2350, opened 2018-04-11
Expose Page Workers


aslushnikov:
We need to develop an API to access WebWorkers and ServiceWorkers.
Workers:
- have execution context, so `worker.evaluate()` should work
- can issue requests, so `worker.on('request')` should be available as well

proposal:
```js
const workers = await page.workers(); // return all the workers 
workers[0].type(); // return either 'webworker' or 'serviceworker'
```

PR #2560, opened 2018-05-16
feat: workers

JoelEinbinder:
This adds `page.workers()`, and two events `workercreated` and `workerdestroyed`. It also forwards logs from a worker into the page `console` event.

Only dedicated workers are supported for now, ServiceWorkers will probably work differently because they aren't necessarily associated with a single page.

#2350

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1620,6 +1620,18 @@ This method returns all of the dedicated [WebWorkers](https://developer.mozilla.
 
 ### class: Worker
 
+The Worker class represents a [WebWorker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API).
+The events `workercreated` and `workerdestroyed` are emitted on the page object to signal the worker lifecycle.
+
+```js
+page.on('workercreated', worker => console.log('Worker created: ' + worker.url()));
+page.on('workerdestroyed', worker => console.log('Worker destroyed: ' + worker.url()));
+
+console.log('Current workers:');
+for (const worker of page.workers())
+  console.log('  ' + worker.url());
+```
+
 #### worker.executionContext()
 - returns: <[Promise]<[ExecutionContext]>>
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -67,6 +67,8 @@
   * [event: 'requestfailed'](#event-requestfailed)
   * [event: 'requestfinished'](#event-requestfinished)
   * [event: 'response'](#event-response)
+  * [event: 'workercreated'](#event-workercreated)
+  * [event: 'workerdestroyed'](#event-workerdestroyed)
   * [page.$(selector)](#pageselector)
   * [page.$$(selector)](#pageselector)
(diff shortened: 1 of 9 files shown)


docs/api.md
```diff
@@ -1596,6 +1612,20 @@ puppeteer.launch().then(async browser => {
...
 Shortcut for [page.mainFrame().waitForXPath(xpath[, options])](#framewaitforxpathxpath-options).
 
+#### page.workers()
+- returns: <[Array]<[Worker]>>
+This method returns all of the dedicated [WebWorkers](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) associated with the page.
+
+> **NOTE** This does not contain ServiceWorkers
+
+### class: Worker
+
```
aslushnikov, line 1622: Worker represents a dedicated instance of [WebWorker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API). Page gets notified about worker lifecycle with `workercreated` and `workerdestroyed` events.

```js
page.on('workercreated', worker => console.log('Worker created: ' + worker.url()));
page.on('workerdestroyed', worker => console.log('Worker destroyed: ' + worker.url()));

console.log('Current workers:');
for (const worker of page.workers())
  console.log('  ' + worker.url());
```

lib/Connection.js
```diff
@@ -214,6 +216,16 @@ class CDPSession extends EventEmitter {
...
+    } else if (object.method === 'Target.receivedMessageFromTarget') {
+      const session = this._sessions.get(object.params.sessionId);
+      if (session)
+        session._onMessage(object.params.message);
+    } else if (object.method === 'Target.detachedFromTarget') {
+      const session = this._sessions.get(object.params.sessionId);
+      if (session)
+        session._onClosed();
+      this._sessions.delete(object.params.sessionId);
+      this.emit(object.method, object.params);
```
aslushnikov: why this emit?

lib/Page.js
```diff
@@ -87,6 +89,30 @@ class Page extends EventEmitter {
...
+    this._workers = new Map();
+    client.on('Target.attachedToTarget', event => {
+      if (event.targetInfo.type !== 'worker') {
+        // If we don't detach from service workers, they will never die.
+        client.send('Target.detachFromTarget', {
+          sessionId: event.sessionId
+        });
+        return;
+      }
+      const session = client._createSession(event.targetInfo.targetId, event.sessionId);
```
aslushnikov, line 102: should this be a part of `CDPSession`?

lib/Connection.js
JoelEinbinder: We still want to send the event to people interacting with the raw protocol. But I see that we need it above in recievedMessageFromTarget as well.

lib/Page.js
JoelEinbinder: I don't think so. If people are already interacting with raw protocol, they can handle sub-targets without our help.

aslushnikov (approved):

PR #2563, opened 2018-05-17
fix(page): respect timeout 0 in page.waitForFunction

### LINKED ISSUE #2540: waitForPredicatePageFunction doesn't respect timeout equal to 0
### Steps to reproduce

* Puppeteer version: 1.4.0
* Platform / OS version: macOS
* URLs (if applicable): -
* Node.js version: v8.9.1

I know that I could use `page.waitForSelector` instead of following the snippet, but this snippet helps me right now to show the problem.

``` typescript
const querySelector = () => {
    return `document.querySelector('${selector}') !== null`;
};
const result = await this.page.waitForFunction(querySelector(), {
    polling: 50,
    timeout: 0,
});
```

I would expect that the following function would wait till the element with the given selector is present and then resolves.

**What happens instead?**
The method `waitForPredicatePageFunction` in the `WaitTask` class does not respect timeout of 0, which should disable the timeout.
So the function resolves immediately without running even once the predicate function.

See [here](https://github.com/GoogleChrome/puppeteer/blob/3b03ff65c7b369debfcaa67a8673fba105a2d20d/lib/FrameManager.js#L914)

The referenced line needs to be changed from
``` javascript
setTimeout(() => timedOut = true, timeout);

```
to
``` javascript
if (timeout) {
    setTimeout(() => timedOut = true, timeout);
}
```

aslushnikov:
The in-page task should not set timeout when timeout is 0.

Fixes #2540.

### DIFF
--- a/test/frame.spec.js
+++ b/test/frame.spec.js
@@ -163,8 +163,7 @@ module.exports.addTests = function({testRunner, expect}) {
         window.__counter = (window.__counter || 0) + 1;
         return window.__injected;
       }, {timeout: 0, polling: 10});
-      await page.waitFor(100);
-      expect(await page.evaluate(() => window.__counter)).not.toBe(0);
+      await page.waitForFunction(() => window.__counter > 10);
       await page.evaluate(() => window.__injected = true);
       await watchdog;
     });
--- a/test/frame.spec.js
+++ b/test/frame.spec.js
@@ -159,10 +159,14 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(error.message).toContain('waiting for function failed: timeout');
     });
     it('should disable timeout when its set to 0', async({page}) => {
-      let error = null;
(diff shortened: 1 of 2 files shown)


test/frame.spec.js
```diff
@@ -159,10 +159,14 @@ module.exports.addTests = function({testRunner, expect}) {
...
     it('should disable timeout when its set to 0', async({page}) => {
-      let error = null;
-      const res = await page.waitForFunction(() => new Promise(res => setTimeout(() => res(42), 100)), {timeout: 0}).catch(e => error = e);
-      expect(error).toBe(null);
-      expect(await res.jsonValue()).toBe(42);
+      const watchdog = page.waitForFunction(() => {
+        window.__counter = (window.__counter || 0) + 1;
+        return window.__injected;
+      }, {timeout: 0, polling: 10});
+      await page.waitFor(100);
```
JoelEinbinder: have another wait for function that waits for counter.

JoelEinbinder (approved):

test/frame.spec.js
aslushnikov: Done.

PR #2572, opened 2018-05-18
feat(page): don't report clientside error with a node stack attached

paulirish:
If you're listening to errors reported via the `'pageerror'` event, it's very odd for it to have a callstack from node appended to the clientside callstack.

Repro:

<details>

```js
(async function() {

  page.on('pageerror', error => console.error('pageerror recieved', error));

  await page.goto('https://output.jsbin.com/yuxaxeh/quiet');
  await browser.close();
})();
```

</details>

### Before
image

### After
image

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -424,7 +424,9 @@ class Page extends EventEmitter {
    */
   _handleException(exceptionDetails) {
     const message = helper.getExceptionMessage(exceptionDetails);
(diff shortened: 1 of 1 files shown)


JoelEinbinder (approved):
This is fine, but for extra credit you can get the message and stack trace separately out of the exceptionDetails and use that to construct an Error with the proper stack trace.

aslushnikov (approved):
I think this is good, let's have it in.

PR #2582, opened 2018-05-21
fix: wait for execution context after navigation

JoelEinbinder:
This is blocked on https://chromium-review.googlesource.com/c/chromium/src/+/1066978

Waits for navigations to resolve before returning the frame's execution context.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -291,7 +291,6 @@ class Frame {
     /** @type {?Promise<!ExecutionContext>} */
     this._contextPromise = null;
     this._contextResolveCallback = null;
-    this._rawContextPromise = null;
     this._setDefaultContext(null);
 
     this._navigationPromise = null;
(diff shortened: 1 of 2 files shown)


lib/FrameManager.js
```diff
@@ -282,16 +305,32 @@ class Frame {
       this._parentFrame._childFrames.add(this);
   }
 
+  _navigationScheduled() {
+    if (!this._defaultContext)
```
aslushnikov: this misses the case when:
- frame schedules navigation
- a runtime context is coming afterwards

In this case, the `_navigationScheduled` doesn't have any effect.

lib/FrameManager.js
```diff
@@ -47,6 +47,17 @@ class FrameManager extends EventEmitter {
     this._client.on('Runtime.executionContextsCleared', event => this._onExecutionContextsCleared());
     this._client.on('Page.lifecycleEvent', event => this._onLifecycleEvent(event));
 
+    this._client.on('Page.frameScheduledNavigation', event => {
+      console.log('sched', event);
+      const frame = this._frames.get(event.frameId);
```
aslushnikov: let's ignore the event if the "delay" is more then 0.

JoelEinbinder:
Taking a different approach to this.

PR #2586, opened 2018-05-22
feat(page): allow screenshot to return a base64 string

### LINKED ISSUE #2566: Allow base64 result in screenshot
When grabbing a screenshot CDP will return base64 which is then encoded into a binary buffer by puppeteer. https://github.com/GoogleChrome/puppeteer/blob/master/lib/Page.js#L761

This behaviour is favourable for most use cases, but I have a use case that benefits from skipping that stage and keeping the returned value as base64. Is it possible to add an option to skip encoding and return a base64 string as the result?

yotamlaufer:
Fixes #2566 

Add `encoding` to the options passed to screenshot. When set to `base64` will return a base64 string instead of a buffer.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1315,7 +1315,8 @@ Shortcut for [page.mainFrame().executionContext().queryObjects(prototypeHandle)]
     - `width` <[number]> width of clipping area
     - `height` <[number]> height of clipping area
   - `omitBackground` <[boolean]> Hides default white background and allows capturing screenshots with transparency. Defaults to `false`.
-- returns: <[Promise]<[Buffer]>> Promise which resolves to buffer with captured screenshot
+  - `encoding` <[string]> The encoding of the image, can be either `base64` or `binary`. Defaults to `binary`.
+- returns: <[Promise]<[Buffer|String]>> Promise which resolves to buffer or a base64 string (depending on the value of `encoding`) with captured screenshot.
 
 > **NOTE** Screenshots take at least 1/6 second on OS X. See https://crbug.com/741689 for discussion.
 
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -723,7 +723,7 @@ class Page extends EventEmitter {
 
   /**
    * @param {!Object=} options
-   * @return {!Promise<!Buffer>}
+   * @return {!Promise<!Buffer|!String>}
    */
   async screenshot(options = {}) {
     let screenshotType = null;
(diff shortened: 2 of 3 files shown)


yotamlaufer:
I signed it!

JoelEinbinder:
The CLA bot is still unhappy with you for some reason. Additionally, can you elaborate a bit on why you want the base64 version of the image? It seems reasonable, but I just want to have some context.

yotamlaufer:
@JoelEinbinder any ideas on what do with the CLA bot? I tried signing again but it says I've already signed. I also tried closing and opening this PR, but doesn't seem to help much.

Context: when Puppeteer is used on Heroku worker nodes and managed by a Redis job queue I found better performance when the value saved is a String (vs Buffer). Honestly, this might be because of the library that I am using, ([bull](https://github.com/OptimalBits/bull)), which seems to convert the Buffer into a ByteArray - but that library works very well for me and is quite popular. Given these circumstances, I think it makes sense to enable this feature for performance reasons.

docs/api.md
```diff
@@ -1315,7 +1315,8 @@ Shortcut for [page.mainFrame().executionContext().queryObjects(prototypeHandle)]
     - `width` <[number]> width of clipping area
     - `height` <[number]> height of clipping area
   - `omitBackground` <[boolean]> Hides default white background and allows capturing screenshots with transparency. Defaults to `false`.
-- returns: <[Promise]<[Buffer]>> Promise which resolves to buffer with captured screenshot
+  - `encoding` <[string]> The encoding of the image, can be either `base64` or `binary`. Defaults to `binary`.
```
aslushnikov, line 1318: In context of images and screenshots, `encoding` usually means JPG vs PNG, so I'd rather name name this  `base64` and make it boolean.

aslushnikov:
@yotamlaufer did you sign CLA as a company? If yes, than can you please tell me the company name so that I can check manually?

yotamlaufer:
@aslushnikov the company name is  'Lost My Name LTD trading as Wonderbly'

docs/api.md
yotamlaufer: @aslushnikov

1. I usually try to avoid boolean values when describing formats to support future compatibility with new values. 
2. `encoding` I believe that the correct term, even in image context, is `compression` or `format`, where `encoding` describes how the compression mechanism works. In this particular use case `encoding` with a `base64` value is the same property/value selection used in NodeJS for Buffers and Streams and I believe is a more familiar term.

If you still think that the property should be changed I don't mind, but I don't think it will be the best choice.

yotamlaufer:
@aslushnikov I received the complete CLA yesterday via email.  Thanks for looking into this.

docs/api.md
aslushnikov: That's reasonable, thank you for argumentation.

aslushnikov (approved):

PR #2589, opened 2018-05-22
feat(elementhandle): add elementHandle.$$eval method

vasyl-pahut:
Fixes https://github.com/GoogleChrome/puppeteer/issues/2401

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2400,20 +2400,13 @@ If `pageFunction` returns a [Promise], then `frame.$$eval` would wait for the pr
 Examples:
 ```html
 <div class="feed">
-  <div class="tweet">
-    <div class="likes">100</div>
-    <div class="retweets">10</div>
-  </div>
-  <div class="tweet">
-    <div class="likes">10</div>
-    <div class="retweets">1</div>
-  </div>
+  <div class="tweet">Hello!</div>
+  <div class="tweet">Hi!</div>
 </div>
 ```
 ```js
 const feedHandle = await page.$('.feed');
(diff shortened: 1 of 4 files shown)


vasyl-pahut:
I signed it!

docs/api.md
```diff
@@ -2386,6 +2387,35 @@ The method runs `element.querySelector` within the page. If no element matches t
...
+<div class="feed">
+  <div class="tweet">
+    <div class="likes">100</div>
+    <div class="retweets">10</div>
+  </div>
+  <div class="tweet">
+    <div class="likes">10</div>
+    <div class="retweets">1</div>
+  </div>
+</div>
```
yanivefraim: I would change the example to be simpler, with only one parent div + two children

lib/ElementHandle.js
```diff
@@ -295,6 +295,25 @@ class ElementHandle extends JSHandle {
     return result;
   }
 
+  /**
+   * @param {string} selector
+   * @param {Function|String} pageFunction
+   * @param {!Array<*>} args
+   * @return {!Promise<(!Object|undefined)>}
+   */
+  async $$eval(selector, pageFunction, ...args) {
```
aslushnikov, line 304: Let's re-use this implementation in `frame.$$eval` method similarly to how it's done for `farme.$eval`:

https://github.com/GoogleChrome/puppeteer/blob/17bc6515deec8229afb73edc3869f497e6c06e10/lib/FrameManager.js#L368-L372

aslushnikov (approved):
Thanks!

PR #2590, opened 2018-05-23
fix: wait for execution context after navigation

JoelEinbinder:
Waits for navigations to resolve before returning the frame's execution context.

This PR currently does not work with navigations triggered by clicking on an anchor tag. That will require an upstream patch which is in progress. But this code is done, and once that upstream patch lands we can just enable the tests here.

Unlike the previous PR, this one does not rely on a new protocol event `frameCanceledNavigation`. Instead, it simulates that event by listening to the network events.

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -256,6 +256,10 @@ class NetworkManager extends EventEmitter {
    * @param {!Protocol.Network.responseReceivedPayload} event
    */
   _onResponseReceived(event) {
+    if (event.type === 'Document' && (event.response.status === 204 || event.response.status === 205)) {
+      const frame = this._frameManager.frame(event.frameId);
+      frame._navigationCanceled();
+    }
     const request = this._requestIdToRequest.get(event.requestId);
     // FileUpload sends a response without a matching request.
     if (!request)
@@ -291,6 +295,8 @@ class NetworkManager extends EventEmitter {
     // @see https://crbug.com/750469
     if (!request)
       return;
+    if (event.type === 'Document' && request._frame)
+      request._frame._navigationCanceled();
     request._failureText = event.errorText;
     const response = request.response();
     if (response)
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -155,7 +155,7 @@ class NetworkManager extends EventEmitter {
       const request = this._interceptionIdToRequest.get(event.interceptionId);
       if (request) {
         this._handleRequestRedirect(request, event.responseStatusCode, event.responseHeaders, false /* fromDiskCache */, false /* fromServiceWorker */, null /* securityDetails */);
-        this._handleRequestStart(request._requestId, event.interceptionId, event.redirectUrl, event.resourceType, event.request, event.frameId, request._redirectChain);
+        this._handleRequestStart(request._requestId, event.interceptionId, event.redirectUrl, event.resourceType, event.request, event.frameId, request._redirectChain, request._isNavigationRequest);
       }
       return;
     }
@@ -163,10 +163,10 @@ class NetworkManager extends EventEmitter {
     const requestId = this._requestHashToRequestIds.firstValue(requestHash);
     if (requestId) {
       this._requestHashToRequestIds.delete(requestHash, requestId);
-      this._handleRequestStart(requestId, event.interceptionId, event.request.url, event.resourceType, event.request, event.frameId, []);
+      this._handleRequestStart(requestId, event.interceptionId, event.request.url, event.resourceType, event.request, event.frameId, [], event.isNavigationRequest);
     } else {
       this._requestHashToInterceptionIds.set(requestHash, event.interceptionId);
-      this._handleRequestStart(null, event.interceptionId, event.request.url, event.resourceType, event.request, event.frameId, []);
+      this._handleRequestStart(null, event.interceptionId, event.request.url, event.resourceType, event.request, event.frameId, [], event.isNavigationRequest);
     }
   }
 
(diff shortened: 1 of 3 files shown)


lib/NetworkManager.js
```diff
@@ -249,13 +250,18 @@ class NetworkManager extends EventEmitter {
...
     }
-    this._handleRequestStart(event.requestId, null, event.request.url, event.type, event.request, event.frameId, redirectChain);
+    this._handleRequestStart(event.requestId, null, event.request.url, event.type, event.request, event.frameId, redirectChain, event.loaderId === event.requestId);
   }
 
   /**
    * @param {!Protocol.Network.responseReceivedPayload} event
    */
   _onResponseReceived(event) {
+    if (event.loaderId === event.requestId && (event.response.status === 204 || event.response.status === 205)) {
```
aslushnikov, line 260: there's asymmetry in the way you handle the `_onResponseReceived` and `_onLoadingFailed`. Is this intentional? 

I'd expect you can move this code at the line 269 and use the `request` and `response` objects here, e.g.:

```js
if (request._isNavigationRequest && request._frame && (response.status() === 204 || response.status() === 205))
  request._frame._navigationCanceled();
```

test/page.spec.js
```diff
@@ -1617,4 +1617,183 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+      expect(await page.evaluate(() => document.location.href)).toBe(server.EMPTY_PAGE);
+    });
+    xit('should not wait for a-click that was canceled by the renderer', async({page, server}) => {
+      server.setRoute('/nextpage', async(req, res) => {
+        // stall
+      });
+      await page.goto(server.EMPTY_PAGE);
+      await page.setContent('<a href="/nextpage">click here</a>');
+      await page.evaluate(() => document.querySelector('a').addEventListener('click', e => setTimeout(() => document.open())));
+      await page.click('a');
```
aslushnikov, line 1672: why's this click canceled by the renderer?

test/page.spec.js
```diff
@@ -1617,4 +1617,183 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+    });
+    it('navigation after a renderer-canceled navigation to the same page', async({page, server}) => {
+      await page.goto(server.EMPTY_PAGE);
+      await page.evaluate(() => {
+        document.location.href = 'grid.html';
+        document.location.href = 'empty.html';
+      });
+      expect(await page.evaluate(() => document.location.href)).toBe(server.EMPTY_PAGE);
+    });
+    it('fail to naviagte', async({page, server}) => {
```
aslushnikov, line 1706: nit: typo in "navigate"

test/page.spec.js
```diff
@@ -1617,4 +1617,183 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+    it('cancel a navigation after it commits and has an execution context', async({page, server}) => {
+      let cleanup;
+      const promise = new Promise(x => cleanup = x);
+      await page.goto(server.EMPTY_PAGE);
+      server.setRoute('/nextpage', async(req, res) => {
+        await evalPromise;
+        res.write('<body><div id="token">token</div>');
+        await page.waitForSelector('#token');
+        await page._client.send('Page.stopLoading');
+        cleanup();
```
aslushnikov, line 1726: can we use `req.abort()` instead? the `Page.stopLoading` is scary.

test/page.spec.js
```diff
@@ -1617,4 +1617,183 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+      });
+      const evalPromise = page.evaluate(() => {
+        document.location.href = '/nextpage';
+      });
+      await evalPromise;
```
aslushnikov, line 1731: this can be inlined

test/page.spec.js
```diff
@@ -1617,4 +1617,183 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+      expect(await page.evaluate(() => document.location.href)).toBe(server.PREFIX + '/nextpage');
+      await promise;
```
aslushnikov, line 1733: I've no clue what's the state of the page when this promise resolves. This is a little worrisome since your expects before and after the promise are the same. (so if this promise is a noop nothing would change).

Is there possibly any nicer way to get to the state you want? (also I'm not sure what's the state we're waiting for)

test/page.spec.js
```diff
@@ -1617,4 +1617,183 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+        res.statusCode = 204;
+        res.end();
+      });
+      await page.evaluate(() => {
+        document.location.href = '/nextpage';
+      });
+      expect(await page.evaluate(() => document.location.href)).toBe(server.EMPTY_PAGE);
+    });
+    it('cancel a navigation after it commits and has an execution context', async({page, server}) => {
+      let cleanup;
```
aslushnikov, line 1718: nit: a nicer name maybe for both the callback and the promise? The test is quite involved so naming will help to read through.

test/page.spec.js
```diff
@@ -1617,4 +1617,183 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+      });
+      await evalPromise;
+      expect(await page.evaluate(() => document.location.href)).toBe(server.PREFIX + '/nextpage');
+      await promise;
+      expect(await page.evaluate(() => document.location.href)).toBe(server.PREFIX + '/nextpage');
+    });
+    it('cancel a navigation before it commits', async({page, server}) => {
+      await page.goto(server.EMPTY_PAGE);
+      server.setRoute('/nextpage', async(req, res) => {
+        await page._client.send('Page.stopLoading');
```
aslushnikov, line 1739: `req.abort()` should do here.

test/page.spec.js
```diff
@@ -1617,4 +1617,183 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+      await page.evaluate(() => {
+        document.location.href = '/nextpage';
+      });
+      expect(await page.evaluate(() => document.location.href)).toBe(server.EMPTY_PAGE);
+    });
+    it('cancel a navigation by the renderer after it hits the browser', async({page, server}) => {
+      await page.goto(server.EMPTY_PAGE);
+      server.setRoute('/nextpage', async(req, res) => {
+        await evalPromise;
+        await page.keyboard.down('1');
```
aslushnikov, line 1750: can you just evaluate `document.open`?

aslushnikov (commented):
Awesome test coverage.

test/page.spec.js
JoelEinbinder: For some reason I can't evaluate in the page while waiting on the initial request, even if I save the possibly stale execution context myself.

test/page.spec.js
JoelEinbinder: Are those going to do the same thing? I want to emulate the user pressing the stop button, which might happen in headful mode.

test/page.spec.js
JoelEinbinder: The link starts loading immediately, and then document.open aborts the loading after a timeout.

lib/NetworkManager.js
JoelEinbinder: It's that way for historical reasons. I'll change it.

JoelEinbinder:
This ended up not being feasible. It relies too much on undocumented behavior in Chromium that is likely to change, and causes the Puppeteer to closely mirror the internals of Blink.

PR #2594, opened 2018-05-23
feat(launcher): add default url to launched browser

### LINKED ISSUE #2402: Different initial pages in different modes
* Puppeteer version: v1.3.0-post
* Node.js version: 10.0.0 v8-canary

```js
(async () => {
  const browser = await require('puppeteer').launch();
  console.log((await browser.pages()).map(page => page.url()));
  await browser.close();
})();
```
```js
[ 'about:blank' ]
```
```js
(async () => {
  const browser = await require('puppeteer').launch({ headless: false });
  console.log((await browser.pages()).map(page => page.url()));
  await browser.close();
})();
```
```js
[ 'chrome-search://local-ntp/local-ntp.html' ]
```

Would it be more consistent, safe and performance-wise optimal to load `about:blank` initially in both modes?

yanivefraim:
fixes #2402

### DIFF
--- a/test/puppeteer.spec.js
+++ b/test/puppeteer.spec.js
@@ -264,6 +264,18 @@ module.exports.addTests = function({testRunner, expect, PROJECT_ROOT, defaultBro
         await page.close();
         await browser.close();
       });
+      it('should have default url when launching browser', async function() {
+        const browser = await puppeteer.launch();
+        const pages = (await browser.pages()).map(page => page.url());
+        expect(pages).toEqual(['about:blank']);
+        await browser.close();
+      });
+      it('should have default url when launching browser with headless:false', async function() {
+        const browser = await puppeteer.launch({ headless: false });
+        const pages = (await browser.pages()).map(page => page.url());
+        expect(pages).toEqual(['about:blank']);
+        await browser.close();
+      });
     });
     describe('Puppeteer.connect', function() {
       it('should be able to connect multiple times to the same browser', async({server}) => {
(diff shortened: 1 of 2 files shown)


JoelEinbinder (approved):
Thanks!

test/puppeteer.spec.js
```diff
@@ -264,6 +264,18 @@ module.exports.addTests = function({testRunner, expect, PROJECT_ROOT, defaultBro
         await page.close();
         await browser.close();
       });
+      it('should have default url when launching browser', async function() {
+        const browser = await puppeteer.launch();
```
JoelEinbinder: Please use the default browsing arguments here. See other tests that launch chrome.

yanivefraim:
@JoelEinbinder - thanks for approving. My tests are broken on Linux, I'm about to push a fix.

test/puppeteer.spec.js
yanivefraim: yes, I figured this out. pushed a fix. 10x!

JoelEinbinder (changes requested):
Actually, this might burn people that were using the arguments to launch Chromium to a default page. Now they will get two tabs, and their tab will no longer be the first entry in browser.pages(). I think we  have to do something fancy and only append this argument if no other page was specified in the arguments.

yanivefraim:
@JoelEinbinder - I understand. Good catch! 

So, I will have to look for an argument without `--` at start maybe (not sure this is a good indicator, I'll have to look at the docs), and in case it is present, I'll skip default `about:blank`? WDYT?

yanivefraim:
After looking at [chromium-command-line-switches](https://peter.sh/experiments/chromium-command-line-switches/), I think that the solution I suggested might work... (although it is kind of ugly)

JoelEinbinder:
Yes, anything starting with '-' should be a URL in the chrome command line args.

yanivefraim:
@JoelEinbinder - added a fix for existing url in launch args. 10x!

JoelEinbinder (approved):

PR #2595, opened 2018-05-23
docs: update page.$$eval and frame.$$eval with 'Array.from'

yanivefraim:
Docs about `page.$$eval` and `frame.$$eval` are not accurate and might be confusing. `document.querySelectorAll` returns `NodeList`, while `frame.$$eval` is actually doing `Array.from(querySelectorAll(selector))`, which actually returns an array. 

This makes things this possible:

`await page.$$eval('div', divs => divs.map...)` 

This patch fixes docs to mention that $$eval is actually performing:

`Array.from(querySelectorAll(selector))`

Which will let the user understand that the element he receives is an array, and not a NodeList.


aslushnikov (approved):
That's a good point, thanks!

PR #2596, opened 2018-05-23
fix: worker execution context flakiness

JoelEinbinder:
Some of the worker tests were failing on the bots. After investigating, I found one race in the test, and one race upstream in Chromium which I filed upstream as https://crbug.com/846099. But I added a small hack here as a temporary workaround.


aslushnikov (approved):
Let's have this in while upstream issue is investigated.

PR #2600, opened 2018-05-24
feat(target): add support for target.page for 'backgroud_page'

### LINKED ISSUE #2438: Add target.evaluate and target.evaluateHandle
Related to https://github.com/GoogleChrome/puppeteer/issues/1215.

Lighthouse has [started testing](https://github.com/GoogleChrome/lighthouse/blob/master/lighthouse-extension/test/extension-test.js#L63-L79) its chrome extension using pptr. They currently use`createCDPSession` to inject code into the extension and run methods:

```js
    const targets = await browser.targets();
    const extensionTarget = targets.find(({_targetInfo}) => {
      return _targetInfo.title === 'Lighthouse' && _targetInfo.type === 'background_page';
    });

    if (!extensionTarget) {
      return await browser.close();
    }

    const client = await extensionTarget.createCDPSession();
    const lighthouseResult = await client.send('Runtime.evaluate', {
      expression: `runLighthouseInExtension({
          restoreCleanState: true,
        }, ${JSON.stringify(lighthouseCategories)})`,
      awaitPromise: true,
      returnByValue: true,
    });
```

It would be nice to make this use case easier:

```js
const lighthouseResult = target.evaluate(cats => {
  return runLighthouseInExtension({restoreCleanState: true}, cats);
}, lighthouseCategories);
```

yanivefraim:
Fixes #2438


lib/Target.js
```diff
@@ -38,7 +38,7 @@ class Target {
    * @return {!Promise<?Page>}
    */
   async page() {
-    if (this._targetInfo.type === 'page' && !this._pagePromise) {
+    if ((this._targetInfo.type === 'page' || this._targetInfo.type === 'background_page') && !this._pagePromise) {
```
aslushnikov, line 41: can you please update `target.page()` documentation and mention that it now works with `'background_page'` type?

Also: can you please add the `'background_page'` type to  `target.type()` documentation?

test/target.spec.js
```diff
@@ -45,6 +45,15 @@ module.exports.addTests = function({testRunner, expect, puppeteer, browserWithEx
       await browserWithExtension.close();
       expect(backgroundPageTarget).toBeTruthy();
     });
+    it('target.page() should return a background_page', async({browser}) => {
+      const browserWithExtension = await puppeteer.launch(browserWithExtensionOptions);
+      const targets = await browserWithExtension.targets();
+      const backgroundPageTarget = targets.find(target => target.type() === 'background_page');
+      const page = await backgroundPageTarget.page();
+      expect(page).not.toBeNull();
+      await page.close();
```
aslushnikov: Can you please try evaluating something inside the page? It should be working, but I'd like to be sure.

Also: it's a little crazy to do `page.close` - iirc background page lifetime is controlled automatically. Let's drop the line?

aslushnikov (approved):

test/target.spec.js
yanivefraim: done. 10x!

lib/Target.js
yanivefraim: `target.page()` docs were updated. 

Regarding `target.type()` - it was already there :)

test/target.spec.js
yanivefraim: @aslushnikov - after rebasing master I have the [following test](https://github.com/GoogleChrome/puppeteer/blob/master/test/puppeteer.spec.js#L273) failing. The reason is that we now have a background page as well as the default `[about:blank]`, which I just added. 

Is this an indicator that we just broke compatibility here?

yanivefraim:
@aslushnikov - after rebasing master I have the [following test](https://github.com/GoogleChrome/puppeteer/blob/master/test/puppeteer.spec.js#L273) failing (pushed it red to my branch intentionally, so you could see the error). 

The reason is that we now have a background page as well as the default `[about:blank]`, which I just added. 

I am wondering if this is an indicator that we just broke backward compatibility here (by automatically adding something like `chrome-extension://kmendfapggjehodndflmmgagdbamhnfd/_generated_background_page.html` page to everyone)?

aslushnikov:
@yanivefraim good point.

It looks like the right thing to do here would be to have `target.page()` working for both background pages and regular pages, whereas `browser.pages()` should return only regular pages.

The motivation behind this is that `browser.pages()` method is perceived as browser visible tabs.

yanivefraim:
> It looks like the right thing to do here would be to have target.page() working for both background pages and regular pages, whereas browser.pages() should return only regular pages.

Done, the test is now green again :)
10x!

yanivefraim:
Something went wrong here, after rebasing (had conflicts and now the test we discussed [here](https://github.com/GoogleChrome/puppeteer/pull/2600#issuecomment-392290756) is failing again). I will take a look later today and fix it.

yanivefraim:
> Something went wrong here, after rebasing (had conflicts and now the test we discussed here is failing again). I will take a look later today and fix it.

For some reason rebase removed my last fix (no conflicts, just removed my code)... 

Added back the fix for this:
>It looks like the right thing to do here would be to have target.page() working for both background pages and regular pages, whereas browser.pages() should return only regular pages.

@aslushnikov - I believe the PR is ready again :)

docs/api.md
```diff
@@ -523,7 +523,7 @@ Disconnects Puppeteer from the browser, but leaves the Chromium process running.
 Promise which resolves to a new [Page] object. The [Page] is created in a default browser context.
 
 #### browser.pages()
-- returns: <[Promise]<[Array]<[Page]>>> Promise which resolves to an array of all open pages.
+- returns: <[Promise]<[Array]<[Page]>>> Promise which resolves to an array of all open pages. Non visible pages, such as `"background_page"` and `"service_worker"`, will not be listed here. You can find them using [target.page()](#targetpage).
```
JoelEinbinder: nit: service_worker isn't a page, so I think it isn't worth being mentioned here.

lib/Browser.js
```diff
@@ -183,7 +183,12 @@ class Browser extends EventEmitter {
    * @return {!Promise<!Array<!Puppeteer.Page>>}
    */
   async pages() {
-    const pages = await Promise.all(this.targets().map(target => target.page()));
+    const nonVisiblePageTypes = ['service_worker', 'background_page'];
```
JoelEinbinder: I think this can be done simpler as something like
```js
return Promise.all(this.targets().filter(target => target.type === 'page').map(target => target.page())));
```

yanivefraim: @JoelEinbinder - done, 10x!

docs/api.md
yanivefraim: done. 10x

aslushnikov (approved):

PR #2618, opened 2018-05-27
fix: Change networkidle2 to networkidle0

### LINKED ISSUE #2587: Make function listenFor async in example custom-event
### Steps to reproduce

* Puppeteer version:1.x.0
* Platform / OS version:Ubuntu 16.04.3 LTS / Windows 10
* URLs (if applicable):https://github.com/GoogleChrome/puppeteer/blob/master/examples/custom-event.js
* Node.js version:8.x.x

[Here is the code](https://github.com/GoogleChrome/puppeteer/blob/master/examples/custom-event.js)

I think ```app-ready fired {}```  message should be displayed on the terminal 

**What happens instead?**
In Ubuntu sometimes  ```app-ready fired {}```  message is displayed on the terminal and sometimes not ,in windows  ```app-ready fired {}```  message is displayed when `headless:false`,but it is not displaying when `headless:true`.

I solved the problem by making `listenFor` function **async** in windows ,but in ubuntu there is some inconsistency with displaying message on console.

Is the solution valid , if not then  why isn't function `listenFor` is **async** when you are **awaiting** it.

Please correct me if i  didn't understood this example correctly.

YogeshSinghChouhan:
Switch example to networkidle0 to solve inconsistency of console messages 
Fixes #2587

### DIFF
--- a/examples/custom-event.js
+++ b/examples/custom-event.js
@@ -42,7 +42,7 @@ const puppeteer = require('puppeteer');
 
   await listenFor('app-ready'); // Listen for "app-ready" custom event on page load.
 
-  await page.goto('https://www.chromestatus.com/features', {waitUntil: 'networkidle2'});
+  await page.goto('https://www.chromestatus.com/features', {waitUntil: 'networkidle0'});
 
   await browser.close();
 })();


YogeshSinghChouhan:
I signed it!

aslushnikov:
This might make the race less bad, but it doesn't fix it.
Instead, we should await the event from the page, and then close the browser.

Feel free to send a separate PR for this.

YogeshSinghChouhan:
@aslushnikov  i didn't understand your point ,i think we are awaiting the event and afterwards browser is closing. Please correct me , if i understood it wrong

aslushnikov:
@YogeshSinghChouhan The solution that waits for `networkidle0` might mostly work but gives no guarantees: how's network idleness related to `app-ready` event?

The inconsistency happens because we sometimes close the browser before the `app-ready` event is dispatched. To fix the race condition the right way, we should wait for the `app-ready` event explicitly and then close the browser.

On a second thought, it doesn't hurt to land this CL since it makes things better without much effort, so I'll just merge it in.

PR #2623, opened 2018-05-29
chore(docs): remove old and non relevant function usage

yanivefraim:


CONTRIBUTING.md
```diff
@@ -150,21 +150,21 @@ npm run unit -- -j 4
 ```js
   ...
   // Using "fit" to run specific test
-  fit('should work', SX(async function() {
+  fit('should work', async function() {
```
JoelEinbinder: can you add the server and page variables required now? `async function({server, page})` ... `server.EMPTY_PAGE`

yanivefraim: sure, done. 10x!

JoelEinbinder (approved):
Thanks!

PR #2631, opened 2018-05-30
fix(page): migrate exposeFunction from console.debug to Runtime.installBinding

alexkozy:
New way is faster and cleaner.


alexkozy:
It is demo PR, it shows how new [binding API](https://chromium-review.googlesource.com/c/chromium/src/+/1079431) may be used instead of console.debug.

It requires at least rolling Chromium with Runtime.installBinding before merging.

aslushnikov:
This is awesome.

@kozy: pptr's chromium version already includes your binding patches to V8; this should all work once you rebaseline this PR on tip-of-tree.

alexkozy:
Andrey, I believe that this PR is ready! :)

aslushnikov (approved):

Issue #1132, opened 2017-10-23
How to stop puppeteer follow redirects


ali-habibzadeh:
Currently it seems the default behaviour of puppeteer is to follow redirects and return the DOM at the end of the chain.
How can this be changed when need to stop the behaviour and make the .goto to stop after the first redirect and simply return the html from that first 301 page for example?

ali-habibzadeh:
Use the `await page.setRequestInterceptionEnabled(true);` when within the interceptor function if the initial response is 3xx, call the request.respond() method to break the chain and discontinue the redirect chain.

GuilloOme:
Same problem here…
I agree, the `page.goto()` should return the 30x response in the promise. 
It could be a params to pass to it upon calling it, like `followRedirect: false`.

ali-habibzadeh:
So I have a solution now but I agree with @GuilloOme that followRedirect: false should be a .goto option prop.

```
let docRedirected = false;

const onResponseHandler = (response) => {
    docRedirected = redirects.isRedirect(response);
};
const onRequestHandler = (request) => {
    if (docRedirected) { return request.abort(); } //Aborts all subsequent requests
    reqRejection.shouldReject(request, options) ? request.abort() : request.continue();
};
page.on('request', onRequestHandler);
page.on('response', onResponseHandler);
```

then in redirect I have:

```
const redirectStatuses = [301, 302, 303, 307, 308];

exports.isRedirect = (response) => {
    return redirectStatuses.includes(response.status)
        && response.request().resourceType === "document";
};
```

GuilloOme:
Thanks a lot @ali-habibzadeh for this workaround!
Still, it would be better to have this `docRedirected` state in puppeteer.

ali-habibzadeh:
definitely agree. using let with side effect like i have done is not ideal code. `followRedirect: false` as an option for` .goto` would be much better.

GuilloOme:
@ali-habibzadeh, your solution is good enough if you are sure that it will not be any concurrent request. In my context, it could be possible to have multiple concurrent request ; so, to avoid blocking the wrong request, I store the redirect response and compare the "location" url with the given url.

Here is my workaround:

```javascript
let lastRedirectResponse = undefined;
page.setRequestInterceptionEnabled(true);

page.on('response', response => {
    // if this response is a redirect
    if ([301, 302, 303, 307, 308].includes(response.status) 
            && response.request().resourceType === 'document') {
        lastRedirectResponse = response;
[...]

PR #2633, opened 2018-05-31
feat(network): introduce Request.isNavigationRequest() method

### LINKED ISSUE #2627: How to set header in puppeter just main page request not all request
* Puppeteer version: 1.4
* Platform / OS version: ubuntu
* URLs (if applicable): any
* Node.js version: 8.x

I want to just apply header in just opened page in puppeter not all inside page headers

  page.setExtraHTTPHeaders({
     'X-Just-Must-Be-Request-In-Main-Request': '1',
  })

will apply all pages request headers all css/js/images and etc...

i want to set just main request headers not all of pages requests.

no one cares:
https://stackoverflow.com/questions/50568054/how-to-set-header-in-puppeter-just-main-page-request-not-all-request

aslushnikov:
This patch introduces `Request.isNavigationRequest()` method.

Fixes #2627, #1132.

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -155,7 +155,7 @@ class NetworkManager extends EventEmitter {
       const request = this._interceptionIdToRequest.get(event.interceptionId);
       if (request) {
         this._handleRequestRedirect(request, event.responseStatusCode, event.responseHeaders, false /* fromDiskCache */, false /* fromServiceWorker */, null /* securityDetails */);
-        this._handleRequestStart(request._requestId, event.interceptionId, event.redirectUrl, event.resourceType, event.request, event.frameId, request._redirectChain);
+        this._handleRequestStart(request._requestId, event.interceptionId, event.redirectUrl, event.isNavigationRequest, event.resourceType, event.request, event.frameId, request._redirectChain);
       }
       return;
     }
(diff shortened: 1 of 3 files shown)


test/network.spec.js
```diff
@@ -197,6 +197,17 @@ module.exports.addTests = function({testRunner, expect}) {
...
+    it('request.isNavigationRequest() should work', async({page, server}) => {
+      const requests = [];
+      page.on('request', request => requests.push(request));
+      server.setRedirect('/rrredirect', '/frames/one-frame.html');
+      await page.goto(server.PREFIX + '/rrredirect');
+      expect(requests[0].isNavigationRequest()).toBe(true);
+      expect(requests[1].isNavigationRequest()).toBe(true);
+      expect(requests[2].isNavigationRequest()).toBe(true);
+      expect(requests[3].isNavigationRequest()).toBe(false);
+      expect(requests[4].isNavigationRequest()).toBe(false);
```
JoelEinbinder: How do we know the order of these? Looks flaky.

lib/NetworkManager.js
```diff
@@ -249,7 +250,8 @@ class NetworkManager extends EventEmitter {
         redirectChain = request._redirectChain;
       }
     }
-    this._handleRequestStart(event.requestId, null, event.request.url, event.type, event.request, event.frameId, redirectChain);
+    const isNavigationRequest = event.requestId === event.loaderId && event.type === 'Document';
```
JoelEinbinder, line 253: The requestId can’t equal the loaderId and it not be a document request right? What happens when navigating to an image?

aslushnikov: The `request.resourceType()` is inferred by the request initiator rather then URL extension; so yes, it should always be `document` as far as I know. 

I'll add a test for navigation to the image to prove the point.

test/network.spec.js
aslushnikov: It's very deterministic:

- `requests[0]` is the initial request for the main document
- `requests[1]` is the redirected request for the main document
- `requests[2]` is the only request from the main document - the iframe main resource
- `requests[3]` and `requests[4]` are both subresources of the iframe.

JoelEinbinder (approved):

PR #2637, opened 2018-05-31
feat(tracing): introduce `jssampling` option for tracing.start

### LINKED ISSUE #1300: Tracing: allow adding custom tracing categories

aslushnikov:
Fixes #1300.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1805,6 +1805,7 @@ await page.tracing.stop();
 - `options` <[Object]>
   - `path` <[string]> A path to write the trace file to.
   - `screenshots` <[boolean]> captures screenshots in the trace.
+  - `jssampling` <[boolean]> enable js sampling. Defaults to `false`.
   - `categories` <[Array]<[string]>> specify custom categories to use instead of default.
 - returns: <[Promise]>
 
--- a/lib/Tracing.js
+++ b/lib/Tracing.js
@@ -46,6 +46,8 @@ class Tracing {
 
     if (options.screenshots)
       categoriesArray.push('disabled-by-default-devtools.screenshot');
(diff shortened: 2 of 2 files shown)


docs/api.md
```diff
@@ -1805,6 +1805,7 @@ await page.tracing.stop();
 - `options` <[Object]>
   - `path` <[string]> A path to write the trace file to.
   - `screenshots` <[boolean]> captures screenshots in the trace.
+  - `jssampling` <[boolean]> enable js sampling. Defaults to `false`.
```
JoelEinbinder, line 1808: What does this do?

paulirish: ive always thought calling this "sampling" is really unhelpful to end users. same with "js samples", though that'd be more accurate.

those names do kinda help set expectations for people who will consume with this data, as working w/ sampling data is a bit more complicated than working with tracing data. but that's a whole 'nother thing.  

other potential names:

* `callstacks`
* `jsCallstacks`
* `javascript`

aslushnikov:
Closing in favor of #2702.

PR #2641, opened 2018-05-31
feat(targets): add target.opener()

Janpot:
feat(targets): add target.opener()

This adds a `.opener` property to a target so that its origin can be tracked.
For now returns `null` when there's no `openerId`.

Fixes https://github.com/GoogleChrome/puppeteer/issues/1830

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -261,6 +261,7 @@
   * [target.browser()](#targetbrowser)
   * [target.browserContext()](#targetbrowsercontext)
   * [target.createCDPSession()](#targetcreatecdpsession)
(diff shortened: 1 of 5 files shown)


docs/api.md
```diff
@@ -2810,6 +2811,11 @@ The browser context the target belongs to.
 
 Creates a Chrome Devtools Protocol session attached to the target.
 
+#### target.opener()
+- returns: <[?Target]>
```
vsemozhetbyt: It seems this should be `<?[Target]>`, otherwise it is not rendered as a link.

aslushnikov (approved):
Thanks!

PR #2646, opened 2018-05-31
chore: stop using console.assert everywhere

### LINKED ISSUE #2547: Annoying "Assertion failed" logs even when error is handled
### Steps to reproduce

* Puppeteer version: 1.3.0
* Node.js version: 10.0.0

1. Have a test that runs `page.click('.selector-that-does-not-exist')`

It should throw an Error that can be caught and not log anything to `console`

**What happens instead?**

`Assertion failed: No node found for selector:` is logged.

This is due to the use of `console.assert()` in the code: https://github.com/GoogleChrome/puppeteer/blob/3b03ff65c7b369debfcaa67a8673fba105a2d20d/lib/FrameManager.js#L594

In my case, I have to click a button that may or may not be there, so I would like to wrap the call to `click()` in a `try/catch` and ignore the `No node found` error. But even though the error is caught, this line is still logged. The logline doesn't expose any more information than the error. But it spams the test log with red herrings for tests that are passing:

```
      file tree
Assertion failed: No node found for selector: .toast__close-button
        ✓ does navigation on file click
Assertion failed: No node found for selector: .toast__close-button
        ✓ expands directory on row click (no navigation)
Assertion failed: No node found for selector: .toast__close-button
        ✓ does navigation on directory row click
Assertion failed: No node found for selector: .toast__close-button
```

Error handling should be in control of the caller of `click()`, which may chose to log the error (which any test framework would do for you if it is not caught).

aslushnikov:
Since Node 10, `console.assert` no longer throws an AssertionError.
(This is generally good since it aligns Node.js with Browsers.)

This patch migrates all usages of `console.assert` in our codebase.
- All the `lib/` and testing code is migrated onto a handmade `assert`
function. This is to make Puppeteer transpilation / bundling easier.
- All the tooling is switched to use Node's `assert` module.

Fixes #2547.

### DIFF
--- a/utils/testrunner/TestRunner.js
+++ b/utils/testrunner/TestRunner.js
@@ -316,7 +316,7 @@ class TestRunner extends EventEmitter {
   }
 
   _addHook(hookName, callback) {
-    console.assert(this._currentSuite[hookName] === null, `Only one ${hookName} hook available per suite`);
+    assert(this._currentSuite[hookName] === null, `Only one ${hookName} hook available per suite`);
     const hook = new UserCallback(callback, this._timeout);
     this._currentSuite[hookName] = hook;
   }
@@ -387,6 +387,11 @@ class TestRunner extends EventEmitter {
   _didFinishTest(test) {
     this.emit('testfinished', test);
   }
+}
+
+function assert(value, message) {
+  if (!value)
+    throw new Error(message);
 }
 
 TestRunner.Events = {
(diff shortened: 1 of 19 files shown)


utils/testrunner/TestRunner.js
```diff
@@ -389,6 +389,11 @@ class TestRunner extends EventEmitter {
   }
 }
 
+function assert(value, message) {
```
JoelEinbinder: Nit: jsdoc

JoelEinbinder (approved):

utils/testrunner/TestRunner.js
aslushnikov: Done.

PR #2652, opened 2018-06-01
feat(launcher): disable crash reporting by default

aslushnikov:
This patch disables crash reporting since it's not needed for
automation purposes.

It also deals some troubles for us since crashpad is a separate
process on Windows which has a larger lifetime than chromium.
This, in turn, prevents us from cleaning up profile directory.


JoelEinbinder (approved):
Exciting!

PR #2658, opened 2018-06-01
fix(browser): ensure first page is created when browser is launched

aslushnikov:
It's impossible to launch chromium without initial page.
This patch makes sure that `puppeteer.launch()` always returns a browser
with at least one page user can connect to.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -166,21 +166,30 @@ class Launcher {
       const ignoreHTTPSErrors = !!options.ignoreHTTPSErrors;
       const setDefaultViewport = !options.appMode;
       const browser = await Browser.create(connection, [], ignoreHTTPSErrors, setDefaultViewport, chromeProcess, gracefullyCloseChrome);
-      // Wait for initial page target to be created.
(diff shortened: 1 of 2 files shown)


lib/Launcher.js
```diff
@@ -165,7 +165,19 @@ class Launcher {
       }
       const ignoreHTTPSErrors = !!options.ignoreHTTPSErrors;
       const setDefaultViewport = !options.appMode;
-      return Browser.create(connection, [], ignoreHTTPSErrors, setDefaultViewport, chromeProcess, gracefullyCloseChrome);
+      const browser = await Browser.create(connection, [], ignoreHTTPSErrors, setDefaultViewport, chromeProcess, gracefullyCloseChrome);
+      // Wait for initial page target to be created.
+      await new Promise(resolve => {
+        waitForInitialPage();
+
+        function waitForInitialPage() {
```
JoelEinbinder: This is too clever.

aslushnikov: My bad.

JoelEinbinder (approved):

Issue #2548, opened 2018-05-15
Support OOPIF


aslushnikov:
[Out-of-process iframes](https://www.chromium.org/developers/design-documents/oop-iframes) are here now and have to be supported.

For the following HTML served from `localhost`, pptr fails to report the frame:

```html
<h1>Hello world</h1>
<script>
  const frame = document.createElement('iframe')
  frame.setAttribute('name', 'bob')
  frame.setAttribute('src', 'https://google.com')
  document.body.appendChild(frame)
</script>
```

Original example: https://github.com/achingbrain/puppeteer-frames
Reported here: https://github.com/GoogleChrome/puppeteer/issues/1140#issuecomment-389054750

achingbrain:
It doesn't even need any JavaScript on the page, plain HTML also fails in the same way:

```html
<!DOCTYPE html>
<html>
  <head>
    <title>Where is my iframe?</title>
  </head>
  <body>
    <h1>Hello world</h1>
    <iframe name="bob" src="https://google.com"></iframe>
  </body>
</html>
```

aslushnikov:
While we're figuring this out, a workaround is to run with `--disable-features=site-per-process` flag:

```js
const browser = await puppeteer.launch({                                                                                                                                                                                                                                                            
  args: ['--disable-features=site-per-process'],                                                                                                                                                                                                                                                    
  headless: false                                                                                                                                                                                                                                                                                   
}) ;
```

tomholub:
Unfortunately `--disable-features=site-per-process` does not help for extension frames on standard domains. Still missing. Tested with bundle from #2506

aslushnikov:
@tomholub indeed, thanks for pointing out. It turned out extension isolation is hard-coded in [`chrome/browser/extensions/chrome_content_browser_client_extensions_part.cc`](https://cs.chromium.org/chromium/src/chrome/browser/extensions/chrome_content_browser_client_extensions_part.cc?type=cs&g=0&l=366-385).

aslushnikov:
For the record: puppeteer v1.5.0 disables out-of-process iframes by default (see #2661). This, however, doesn't apply to extension iframes - these still operate out-of-process.

Swordyjohn:
Is there a time frame for OOPIF support?

aslushnikov:
@Swordyjohn no good eta; is this pressing you in any way?

Swordyjohn:
[...]

PR #2661, opened 2018-06-01
fix: disable OOPIF by default

aslushnikov:
This patch disables OOPIF by default.

**NOTE**: this is a temporary bandaid for the time we're crafting
the full-fledged support for site isolation over DevTools protocol.

References #2548.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -38,6 +38,8 @@ const DEFAULT_ARGS = [
   '--disable-default-apps',
   '--disable-dev-shm-usage',
   '--disable-extensions',
+  //TODO Support OOOPIF. @see https://github.com/GoogleChrome/puppeteer/issues/2548
+  '--disable-features=site-per-process',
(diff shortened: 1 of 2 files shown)


JoelEinbinder (approved):

SilurianYang:
I did it! When the desired effect is achieved, remove the intercept

```
    page.on('request', async function jsRequest(res) {
        if (/modules\/cube-loader\/cube-loader\.min\.js(?=\?.*)/.test(res.url())) {
            const jsCont = getExtendsJs('src/extendsJS/cube-loader.min.js');
            await res.respond({
                status: 200,
                contentType: 'application/javascript; charset=utf-8',
                body: jsCont
            });
            page.removeListener('request',jsRequest);
            await page.setRequestInterception(false)
            return false
        }
        res.continue();
    });
```

o0101:
Is this still needed? 🤔

PR #2662, opened 2018-06-01
chore(ci): switch Windows containers

fkorotkov:
Cirrus CI got some optimizations for containers based of `microsoft/windowsservercore:latest`.

Now startup time for windows builds is around 1:30 seconds instead of around 4 minutes.

to: @aslushnikov


aslushnikov (approved):
Thanks!

PR #2666, opened 2018-06-02
chore(ci): temporary disable cirrus win builds

aslushnikov:
Win bots all fail on Cirrus; disable them until the issue is resolved.

We still have win coverage with AppVeyour.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -3,6 +3,11 @@ env:
 
 task:
   matrix:
+    - name: node6 (windows)
+      allow_failures: true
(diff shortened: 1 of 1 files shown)


aslushnikov:
cc @fkorotkov

.cirrus.yml
```diff
@@ -3,10 +3,6 @@ env:
 
 task:
   matrix:
-    - name: node6 (windows)
-      windows_container:
```
fkorotkov: you can try just to add `allow_failures: true`

Issue #2629, opened 2018-05-30
Method to show if element is displayed


b-ponomarenko:
Does Puppeteer has API which help detect if element is visible (or displayed), analogue to Selenium (Webdriver) `.isDisplayed` method?

I found method [waitForSelector](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagewaitforselectorselector-options) method which has options `visible`, but I think if DOM node has `display: block;` and some parent component has style `overflow: hidden` and my element outside the parent this `waitForSelector` returns wrong

ebidel:
Overflow "visibility" gets tricky. Using [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) would be more appropriate for this case. You can inject a snippet of IO code into the page and get the results using `page.evaluate`.

aslushnikov:
> Using Intersection Observer would be more appropriate for this case.

@ebidel nice idea, we should use Intersection Observer for the `waitForSelector` implementation.

> Does Puppeteer has API which help detect if element is visible (or displayed), analogue to Selenium (Webdriver) .isDisplayed method?

Not yet, but it makes perfect sense to have `ElementHandle.isVisible()` that is aligned with `waitForSelector`'s implementation of visibility.

yanivefraim:
@b-ponomarenko - can you help with a test case where `waitForSelector` with `visible` does not work for you?

b-ponomarenko:
@yanivefraim yes, simple case is carousel, for example from all carousel items I want choose only visible, but waitForSelector with visible option returns all items.

b-ponomarenko:
I try to inject script with IO:
‘’’js
 const isDisplayed = await page.evaluate(
        () => new Promise(resolve => {
          const callback = ([entry]) => resolve(entry.isIntersecting);
          const observer = new IntersectionObserver(callback, { root: null, rootMargin: '0px', threshold: 1.0 });

          observer.observe(document.querySelector('.ui-trading-main__product-title'));
          setTimeout(() => observer.unobserve(document.querySelector('.ui-trading-main__product-title')), 0)
        })
    );
‘’’
but have error stack trace
‘’’

Error: Protocol error (Runtime.callFunctionOn): Promise was collected undefined
    at Promise (/Users/b.ponomarenko/Desktop/puppeteer/lib/Connection.js:202:56)
    at new Promise (<anonymous>)
[...]

PR #2673, opened 2018-06-04
feat(elementhandle): introduce elementHandle.isIntersectingViewport() method.

b-ponomarenko:
Add API to define is element visible in viewport via Intersection Observer

References #2629


b-ponomarenko:
I signed it!

b-ponomarenko:
Can I rerun `node6 (windows)`?

b-ponomarenko:
@aslushnikov please check

docs/api.md
```diff
@@ -219,6 +219,7 @@
   * [elementHandle.getProperties()](#elementhandlegetproperties)
   * [elementHandle.getProperty(propertyName)](#elementhandlegetpropertypropertyname)
   * [elementHandle.hover()](#elementhandlehover)
+  * [elementHandle.isDisplayed()](#elementhandleisdisplayed)
```
JoelEinbinder: nit: I like isVisible over isDisplayed, because we already use the word visible in waitForSelector.

aslushnikov: `elementHandle.isVisible()` is better since it aligns with `waitForSelector`'s "visible" option.

lib/ElementHandle.js
```diff
@@ -341,6 +341,24 @@ class ElementHandle extends JSHandle {
...
+   * @returns {!Promise<boolean>}
+   */
+  isDisplayed() {
+    return this.executionContext().evaluate(
+        node => new Promise(resolve => {
+          const callback = entries => {
+            resolve(entries[0].isIntersecting);
+            observer.unobserve(node);
+          };
+          const observer = new IntersectionObserver(callback, { root: null, rootMargin: '0px', threshold: 1.0 });
```
aslushnikov: so when will the `callback` be called? Is it always called initially?

test/elementhandle.spec.js
```diff
@@ -184,6 +184,17 @@ module.exports.addTests = function({testRunner, expect}) {
...
 
+  describe('ElementHandle.isDisplayed', function() {
+    it('should return false if element is not visible in viewport', async({page, server}) => {
+      await page.goto(server.PREFIX + '/hidden-and-visible.html');
+      await page.setViewport({ width: 1000, height: 400 });
+
+      const box = await page.$('.box');
+
+      expect(await box.isDisplayed()).toBe(false);
+    });
```
aslushnikov: can we also add a test that checks for the truthy value?

test/assets/hidden-and-visible.html
```diff
@@ -0,0 +1,23 @@
+<html>
```
aslushnikov: let's use `page.setContent` instead of a separate asset.

aslushnikov (commented):
Thanks, this looks promising

As part of this PR, let's change `waitForSelector` implementation to use intersection observer; this way we won't have inconsistencies like this:

```js
const e = await page.waitForSelector('div', {visible: true});
await e.isVisible(); // false
```

JoelEinbinder:
If we change waitForSelector to use intersection observer, won’t that cause elements below the fold to be considered invisible? That’s a pretty major change, and I’m not sure if it makes that feature more or less useful. “Visible” vs “would visible if you scroll down” is hard to define. And it conflicts with the original use case for this feature, which was to report elements ina carousel as invisible even though they could be scrolled to horizontally.

aslushnikov:
> won’t that cause elements below the fold to be considered invisible?

@JoelEinbinder These will be considered invisible. This is a very good point though, we might need to postpone this till 2.0. 

> That’s a pretty major change, and I’m not sure if it makes that feature more or less useful.

Usefulness is debatable, but API sanity will improve:
- Our current `isVisible` is more like `isClickable`, it's quite quirky with a huge grey area of what is considered "visible".
- Visibility defined by intersection observer is aligned with a natural perception of visibility, it's easy to explain and understand.

docs/api.md
b-ponomarenko: Ok, I will rename to `isVisible`

lib/ElementHandle.js
b-ponomarenko: Yes, it will be called initially and then unobserve from the node

test/elementhandle.spec.js
b-ponomarenko: Yes, I'll add this test case soon

b-ponomarenko:
@aslushnikov change implementation of `_waitForSelectorOrXPath`, but in some cases [this test](https://github.com/GoogleChrome/puppeteer/blob/master/test/frame.spec.js#L287) fails (when I run this test in single mode almost always test passed), what can be wrong?

**upd**
no longer relevant, understood the problem and fix

b-ponomarenko:
@aslushnikov please check PR

aslushnikov:
> @aslushnikov please check PR

@b-ponomarenko We discussed this PR with @JoelEinbinder one more time.

Summary:
- there are two types of visibility: visible-in-viewport (this implementation) and displayed-in-page (currently used for `visible` option in `waitForSelector`)
- both make perfect sense from API standpoint: visible-in-viewport comes handy to detect what user sees, and displayed-in-page is more like "interactable", e.g. something that could be clicked.
- changing `waitForSelector`'s `visible` semantics from displayed-in-page to visible-in-viewport is a **breaking change**; we can't afford it yet.

So the way forward with this:
- disregard my suggestion on using IntersectionObserver for `waitForSelector`. (Sorry for misguiding you)
- introduce a new distinct method on `ElementHandle`, e.g. `ElementHandle.isVisibleInViewport`, that uses Intersection Observer.

As a part of [2.0 effort](https://github.com/GoogleChrome/puppeteer/issues/2079), we might want to do a better job separating "interactivity" vs "visibility", e.g. renaming `visible` into `interactable`. I've added a bullet.

b-ponomarenko:
@aslushnikov I canceled using Intersection Observer in `_waitForSelectorOrXPath`. What's next? Really, maybe rename `visible` property to `interactable` or some other property name, and move current implementation via Intersection Observer to `visible` property?

aslushnikov (approved):
We discussed this with @JoelEinbinder and renamed into `elementHandle.isIntersectingViewport()`. Thanks for the PR!

justnpT:
I wouldn't be so sure that inputElement.isIntersectingViewport() is true only when an element is visible for human eyes, I have a case where element didn't yet appear as visible for eyes and
`await element.isIntersectingViewport()`
returns true

aslushnikov:
>  I have a case where element didn't yet appear as visible for eyes and
await element.isIntersectingViewport()
returns true

@justnpT interesting! Can you please file a new issue with a repro for your usecase?

Issue #2410, opened 2018-04-18
Inconsistent text rendering in headless mode


Mins:
**EDIT:** The fix is to add `--font-render-hinting=none` to the launch args. E.g.
```
  var browser = await puppeteer.launch({
    headless: true,
    args: ['--font-render-hinting=none']
  });
```

**Original Comment:**
Font spacing seems to be inconsistent between headless and non headless mode.
This is likely a Chromium bug for Puppeteer versions 1.2.0 and above.

### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.2
* Platform / OS version: Linux
* URLs (if applicable):
* Node.js version: 8.6.0

**What steps will reproduce the problem?**

1. screenshot.js
```
'use strict';

const puppeteer = require('puppeteer');

(async() => {
  const browser = await puppeteer.launch({ headless: true }); // toggle to false
  const page = await browser.newPage();
  await page.goto('file:///tmp/test.html');
  await page.waitFor(5000);
  await page.screenshot({path: '/tmp/screenshot.png'});
  await browser.close();
})();
```
2. test.html
```
<html>
  <head>
    <link rel="stylesheet"
          href="https://fonts.googleapis.com/css?family=Lato">
    <style>
      body {
        font-family: 'Lato', serif;
        font-size: 15px;
      }
      div.a {
        line-height: 0.5;
      }
    </style>
  </head>
  <body>
    <div class="a">
      <div>aaaaa..............................................................................|111</div><br>
      <div>qwertyasdfzxcvyuiohjklbnm................................|222</div><br>
      <div>longlonglonglonglonglonglonglongshorty......|333</div>
    </div>
  </body>
</html>
```
3. `node /tmp/screenshot.js`, then repeat with `headless: false` 

**What is the expected result?**
Text is correctly aligned and looks the same as opening the HTML in browser. `headless: false`

![headless_false](https://user-images.githubusercontent.com/1380839/38925691-f36ba690-4332-11e8-96b4-a1cc0e784e60.png)

**What happens instead?**
Text is misaligned with `headless: true`

![headless_true](https://user-images.githubusercontent.com/1380839/38925690-f33bcfce-4332-11e8-9a19-0ef463f032e9.png)

yrshaikh:
Could you try once by setting the [viewport](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagesetviewportviewport)

Mins:
[...]

PR #2693, opened 2018-06-06
feat(launcher): explicitly set font-render-hinting to medium

cfellin1:
In response to #2410.

Snapshot of properly aligned text:

text-alignment

Looking for feedback on snapshot test which
1) is probably not in the right place within the testing suite
2) is failing for me on Linux, with the snapshot being different from the expectation, but not visibly so.


test/page.spec.js
```diff
@@ -1659,4 +1659,13 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
       expect(page.browser()).toBe(browser);
     });
   });
+
+  describe('Page', function() {
+    fit('should render text consistently', async function({ page, server }) {
```
JoelEinbinder: s/fit/it

lib/Launcher.js
```diff
@@ -45,6 +45,7 @@ const DEFAULT_ARGS = [
   '--disable-prompt-on-repost',
   '--disable-sync',
   '--disable-translate',
+  '--font-render-hinting=medium',
```
JoelEinbinder, line 48: If this is really better, we should probably set it as the default in headless chromium and not pass it in with Puppeteer. @caseq ptal.

test/assets/text-alignment.html
```diff
@@ -0,0 +1,22 @@
...
+    font-size: 15px;
+  }
+  div.a {
+    line-height: 0.5;
+  }
+  </style>
+</head>
+<body>
+  <div class="a">
+    <div>aaaaa..............................................................................|111</div><br>
```
JoelEinbinder, line 17: We don't have consistent font rendering across operating systems, so this test is going to be tricky to make work. I am not an expert on fonts, but I imagine there is more than just the hinting that will make this test not pixel perfect.

cfellin1: Yeah, that's consistent with my experiences thus far. How would you recommend testing the change?

cfellin1: (assuming this is even the approach we decide on)

JoelEinbinder: We could launch the page and headless and headful, and check that the screenshots are the same. Look at the headful.spec.js for examples of that.

aslushnikov:
@cfellin1 @JoelEinbinder I looked more into this. We need to fix chromium to make headful respect the `--font-render-hinting` flag. Otherwise there's no guarantee font hinting will be the same between headful and headless.

aslushnikov:
Closing this until upstream is fixed.

PR #2697, opened 2018-06-07
fix: emit all arguments from worker console logs

### LINKED ISSUE #2640: console message from Webworker receives only first argument
When using `page.on('console'...` on a console from a WebWorker, only first argument is printed.

In the following example, only 'a' will be logged:

```js
(async () => {
  page.on('console', message => console.log(message.text()));
  await page.setContent(`<!DOCTYPE html>
  <script id="worker1" type="javascript/worker">
    console.log('a', 'b'); //only 'a' will be logged
  </script>
  <script>
    const blob = new Blob([
      document.querySelector('#worker1').textContent
    ], { type: "text/javascript" })

    new Worker(window.URL.createObjectURL(blob));
  </script>`);
})()
```

JoelEinbinder:
Log.entryAdded doesn't report all the arguments from console logs. This PR switches to use Runtime.consoleAPICalled.

fixes #2640

### DIFF
--- a/lib/Worker.js
+++ b/lib/Worker.js
@@ -21,23 +21,24 @@ class Worker extends EventEmitter {
   /**
    * @param {Puppeteer.CDPSession} client
    * @param {string} url
-   * @param {function(!Protocol.Log.entryAddedPayload)} logEntryAdded
+   * @param {function(!string, !Array<!JSHandle>)} consoleAPICalled
    */
-  constructor(client, url, logEntryAdded) {
+  constructor(client, url, consoleAPICalled) {
     super();
     this._client = client;
     this._url = url;
     this._executionContextPromise = new Promise(x => this._executionContextCallback = x);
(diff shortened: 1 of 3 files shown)


lib/Worker.js
```diff
@@ -21,23 +21,24 @@ class Worker extends EventEmitter {
   /**
    * @param {Puppeteer.CDPSession} client
    * @param {string} url
-   * @param {function(!Protocol.Log.entryAddedPayload)} logEntryAdded
+   * @param {function(!string, !Array<!JSHandle>)} consoleAPICalled
```
ebidel, line 24: `!Array<!Puppeteer.JSHandle>`?

JoelEinbinder: We haven't been using the Puppeteer. namespace when the type is included directly in the file.

ebidel: Ack

ebidel (approved):

PR #2702, opened 2018-06-08
feat(tracing): enable high resolution JavaScript sampling

JoelEinbinder:
I've been told that this will make our JavaScript tracing 💯x more accurate, at minimal performance loss. Let's turn it on for everyone always.

alternative to #2637


aslushnikov (approved):
This will result in ~3% perf overhead - not that much, given how useful it is.

Got a blessing for this from @paulirish and @a1ph.

PR #2710, opened 2018-06-10
fix(frame): set execution context null if it's not created

### LINKED ISSUE #2709: In mixed-content page page.exposeFunction is stuck indefinitely
### Steps to reproduce

* Puppeteer version: 1.5.0
* Platform / OS version: Mac OS 10.13.3
* URLs (if applicable): https://www.bennish.net/mixed-content.html
* Node.js version: 10.1.0

1. Access to a mixed-content page.
2. expose a function by `page.exposeFunction`.
3. the execution is stuck indefinitely.

```js
(async() => {

  await page.goto('https://www.bennish.net/mixed-content.html');
  await page.exposeFunction('compute', function(a, b) {
    return Promise.resolve(a * b);
  });
  const result = await page.evaluate(async function() {
    return await compute(3, 5);
  });
  console.log(result);
  await browser.close();
})();
```

Expose function successfully or at lease throws an error.

**What happens instead?**

The execution is is stuck indefinitely.

yujiosaka:
Fixes: https://github.com/GoogleChrome/puppeteer/issues/2709

In mixed-content page like [this one](https://www.bennish.net/mixed-content.html), frame stops loading before execution context is created.
This behavior causes `exposeFunction` to be stuck indefinitely because `contextPromise` will never be never resolved.

There is no way setting proper context if it's not created yet, so I think it's better to set `null`.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -68,6 +68,7 @@ class FrameManager extends EventEmitter {
     const frame = this._frames.get(frameId);
     if (!frame)
       return;
+    frame._setContext(null);
     frame._onLoadingStopped();
     this.emit(FrameManager.Events.LifecycleEvent, frame);
   }
@@ -185,7 +186,7 @@ class FrameManager extends EventEmitter {
     const context = new ExecutionContext(this._client, contextPayload, this.createJSHandle.bind(this, contextPayload.id), frame);
     this._contextIdToContext.set(contextPayload.id, context);
     if (frame)
-      frame._setDefaultContext(context);
+      frame._setContext(context);
   }
 
   /**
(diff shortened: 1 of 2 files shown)


JoelEinbinder:
Interesting. Ideally, Puppeteer should force the creation of an execution context. We already do this for pages like `about:blank` that normally wouldn't have an execution context.

It looks like DevTools has the same bug on the page. No execution context is shown for the iframe. But I can force an execution context to be created by evaluating `document.querySelector('iframe').contentWindow`. I'll look into this a bit more and then possibly file an upstream bug on Chromium.

yujiosaka:
@JoelEinbinder 
Thanks! If it can be fixed upstream, that would be perfect!
Please feel free to close this PR as far as the referred issue stays open.

aslushnikov:
Closing in favor of upstream fix: https://chromium-review.googlesource.com/c/chromium/src/+/1102154

PR #2715, opened 2018-06-11
fix(helpers): support thrown strings in getExceptionMessage

thorn0:
Of course, strings aren't proper errors, but scripts sometimes do that, and Puppeteer loses such error messages.

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -61,7 +61,7 @@ class Helper {
    */
   static getExceptionMessage(exceptionDetails) {
     if (exceptionDetails.exception)
-      return exceptionDetails.exception.description;
+      return exceptionDetails.exception.description || exceptionDetails.exception.value;
     let message = exceptionDetails.text;
     if (exceptionDetails.stackTrace) {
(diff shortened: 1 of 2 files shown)


lib/helper.js
```diff
@@ -60,8 +60,12 @@ class Helper {
    * @return {string}
    */
   static getExceptionMessage(exceptionDetails) {
-    if (exceptionDetails.exception)
+    if (exceptionDetails.exception) {
+      if (exceptionDetails.exception.type === 'string') {
```
aslushnikov: let's return `description` if there's any, and fallback to `value` otherwise.
This way we'll support throwing numbers as well

aslushnikov (commented):
Nice patch.

Let's add a test to `Page.evaluate` testsuite so that we don't regress this in future.

https://github.com/GoogleChrome/puppeteer/blob/147f98d1f0e02221895573f2b0958125d9a7ddb4/test/page.spec.js#L75-L89

thorn0:
@aslushnikov Done. Please have a look. Travis failed for some reason, but it seems unrelated.

aslushnikov (approved):

PR #2717, opened 2018-06-11
feat(worker): Allow Workers to emit their network events

chad3814:
small patch that hooks `Worker`s into the `NetworkManager` so that they can `emit` network events


chad3814:
I signed it!

JoelEinbinder:
Thanks for the PR!

Maybe we should have these events fired on the page, like we do for frames. Wdyt @aslushnikov ?

Also needs a test.

chad3814:
yeah I was sure how to make some tests, I'm not familiar with how the coverage tests work, there don't appear to be specific tests for the `Page` versions, they just look like they get exercised in the process of other tests. Something similar could be done in `Worker` but where do those tests get their worker script? I can change it to open another remote resource.

chad3814:
Okay wrote tests :)

chad3814:
The travis build had a timeout on a different test I didn't touch and AppVeyor succeeded on that test...

aslushnikov:
@chad3814 we played with this and found critical issues with nested targets in DevTools protocol:
- events are not fired when request interception is enabled in the main target
- network throttling doesn't quite work with subtargets

We'll have to fix these in Chromium as a part of #2548. There's no ETA for OOPIFs since it's unclear what else is broken in puppeteer with multiple subtargets.

I'll close this PR for now since there's no way we can move forward here without upstream being fixed.
I filed https://github.com/GoogleChrome/puppeteer/issues/2781 to track the feature.

Thank you for the PR!

PR #2736, opened 2018-06-14
fix(tests): make tests work on non-English locales

thorn0:
`Date.prototype.toString` may return non-ASCII characters, which aren't accepted by `setHeader`.

E.g., on Russian locale, it might look like this:
```
> new Date().toString()
'Thu Jun 14 2018 13:11:50 GMT+0300 (Финляндия (лето))'
```


aslushnikov (approved):
Awesome, thanks for the PR

PR #2739, opened 2018-06-14
fix(page): fix race condition in WaitTask

### LINKED ISSUE #1325: When would an error "Cannot find context with specified id undefined" happen?
I am getting the following error:

```js
Error: Protocol error (Runtime.callFunctionOn): Cannot find context with specified id undefined
    at Session._onMessage (/srv/node_modules/puppeteer-edge/lib/Connection.js:205:25)
    at Connection._onMessage (/srv/node_modules/puppeteer-edge/lib/Connection.js:105:19)
    at emitOne (events.js:115:13)
    at WebSocket.emit (events.js:210:7)
    at Receiver._receiver.onmessage (/srv/node_modules/ws/lib/WebSocket.js:143:47)
    at Receiver.dataMessage (/srv/node_modules/ws/lib/Receiver.js:389:14)
    at Receiver.getData (/srv/node_modules/ws/lib/Receiver.js:330:12)
    at Receiver.startLoop (/srv/node_modules/ws/lib/Receiver.js:165:16)
    at Receiver.add (/srv/node_modules/ws/lib/Receiver.js:139:10)
    at Socket._ultron.on (/srv/node_modules/ws/lib/WebSocket.js:139:22)
    at emitOne (events.js:115:13)
    at Socket.emit (events.js:210:7)
    at addChunk (_stream_readable.js:266:12)
    at readableAddChunk (_stream_readable.js:253:11)
    at Socket.Readable.push (_stream_readable.js:211:10)
    at TCP.onread (net.js:585:20)

```

It is a large codebase and it is unclear whats triggering this error.

Any guides?

On that note, there needs to be a better way to throw error. Without knowing the origin of the error in the code it is impossible to trace down these exceptions.

aslushnikov:
This patch eliminates a common race condition with WaitTask, that
happens when predicate function gets resolved right before the execution
context gets destroyed.
This situation results in a "Cannot find context with specified id undefined"
exception.

Credits go to @jakub300 for his wonderful [investigation](https://github.com/GoogleChrome/puppeteer/issues/1325#issuecomment-395472092).

Fixes #1325.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -869,7 +869,9 @@ class WaitTask {
     }
 
     // Ignore timeouts in pageScript - we track timeouts ourselves.
-    if (!error && await this._frame.evaluate(s => !s, success)) {
+    // If the frame's execution context has already changed, `frame.evaluate` will
+    // throw an error - ignore this predicate run altogether.
+    if (!error && await this._frame.evaluate(s => !s, success).catch(e => true)) {
       await success.dispose();
       return;
     }
--- a/test/frame.spec.js
+++ b/test/frame.spec.js
@@ -84,6 +84,14 @@ module.exports.addTests = function({testRunner, expect}) {
       await page.evaluate(() => window.__FOO = 1);
       await watchdog;
     });
+    it('should work when resolved right before execution context disposal', async({page, server}) => {
+      await page.evaluateOnNewDocument(() => window.__RELOADED = true);
+      await page.waitForFunction(() => {
+        if (!window.__RELOADED)
+          window.location.reload();
+        return true;
+      });
+    });
     it('should poll on interval', async({page, server}) => {
       let success = false;
       const startTime = Date.now();


JoelEinbinder (approved):
Good catch!

7starsone:
Hello, I tried to edit `.../node_modules/puppeteer/lib/FrameManager.js` in a project.... but I get the same error despite of that. Maybe is that not enough and do I need to do something else? Or is this fix really not definitive? Thanks

aslushnikov:
@7starsone try installing `puppeteer@next` and see if it helps you.

7starsone:
Hello, if it helps I'm using this https://github.com/nesk/puphpeteer but the owner banned me from comments because of **Cannot find context with specified id undefined** 😆 because not about his work and he stated that I didn't provide all the details but I did it, a code example, the logger, the used software versions and so on... and still had some issues with crashing script (not about my php code but the try catch not working and sometimes even if the URLs are from the same website and the page structure is the same...) like this, then "...`Unexpected token { in JSON at position`..." and "`listen EADDRINUSE`". Here the entire discussion https://github.com/nesk/puphpeteer/issues/7

Thank you very much

aslushnikov:
@7starsone Looks like you have multiple issues going on.
- The "unexpected token" comes from the syntax error in javascript code
- The "listen EADDRINUSE" comes from trying to use occupied HTTP port.

These all have nothing to do with "Cannot find context with specified id", which usually happens because page navigates away while you're trying to `page.evaluate` in it.

If you have a puppeteer script that reproduces the problem with current puppeteer tip-of-tree and which I can run locally, please file a bug and we'll try to see what's going on.

7starsone:
@aslushnikov  "_error in javascript code_" what javascript code? Something in the project from puppeteer or that one coming from the page content when goto URL? 
The error on log is related to the line 230 of ProcessSupervisor.php where he (?) does a `checkProcessStatus();` but it doesn't seem to be related to javascript...maybe you could take a look at that code.
Then, "`These all have nothing to do with`" that was clear, I just listed what issues were detected...that cause a crash and not only an exception to handle. Indeed, I have not a javascript code because I code in php in this case and the procedure often stops because of those **external** errors. Thanks

p.s.: but the funny thing is also that he himself on code comments of that file writes..."...`if the process crash while executing the instruction`" so already foresees a crash, which should not be there because in my opinion the goto function should just retrieve the page content after `'waitUntil' => 'networkidle0'` and if it can't, if the URL is invalid, a 404 or no content after the timeout...it should not crash because of this reason but just handle an exception, so you could retry at another time or investigate about that specific bad/strange URL. So, the process crash, what's the real reason about it? A mistery.

7starsone:
Hello, please, can you explain how to change the used port? (on the /lib folder? where?) The problem is that on documentation of nesk/puphpeteer this is not explained and not even here. Unfortunately, `listen EADDRINUSE` is really annoying and slows down everything because even if I launch a script one time a minute in case it's not running, this condition may  persist for many minutes...and also, still not clear where  "...`Unexpected token { in JSON at position`..." comes from 😉 
thank you

p.s.: on CentOS 7 I did 
`setcap 'cap_net_bind_service=+ep' /usr/bin/node`
and at the moment, it seems to fix the `listen EADDRINUSE`, at least
Update: I'm afraid not, it's not fixed, the error is always there, again and nowhere is explained how to manage puppeteer/puphpeteer in this case and when I wrote issue about this I was asked for a log. Provided the log (even if someone should know how to fix this irrespective of a log...) anyway, no solution and no explanation, on the contrary an unjustified ban from comments instead 😄

PR #2740, opened 2018-06-14
feat(element-handle): remove throw in case of empty elementHandle

### LINKED ISSUE #2708: $$eval call causes error if no elements found
* Puppeteer version: 1.5.0
* Platform / OS version: Win 10.0.16299 Build 16299
* Node.js version: 8.7.0

1. Load any page. Could be even blank page;
2. Call $$eval() with any non-existing selector

var browser = await puppeteer.launch()
var page = await browser.newPage()
console.log(await page.$$eval('.notaclass', els => els.length))

0 is shown in console output;

**What happens instead?**
Error: failed to find element matching selector ".notaclass"

**Note:** This is only reproducible with v1.5.0 (1.4.0 works as expected)

yanivefraim:
Fixes #2708 

`elementHandle.$$eval` introduced a breaking change in `page.$$eval`: In case of an empty selector, it throws an error. I'm not sure what is the best solution for that, so I added this PR, mainly for discussion.

Possible solutions:

1. Remove the check for $$eval empty result from elementHandle.$$eval (which will also be a breaking change for this feature?). We can fix this later on V2.0.0.
2. Remove the usage of elementHandle.$$eval in [frameManager.$$eval](https://github.com/GoogleChrome/puppeteer/commit/1e07925e26783d586ab63a6eb88063ca51fb5647#diff-16f3b9739b97830927de83812ac92d78). This will undo the breaking change, leaving elementHandle.$$eval with the test for empty result.
3. Do nothing...

### DIFF
--- a/test/elementhandle.spec.js
+++ b/test/elementhandle.spec.js
@@ -336,6 +336,15 @@ module.exports.addTests = function({testRunner, expect}) {
       const content = await elementHandle.$$eval('.a', nodes => nodes.map(n => n.innerText));
       expect(content).toEqual(['a1-child-div', 'a2-child-div']);
     });
+
+    it('should not throw in case of missing selector', async({page, server}) => {
+      const htmlContent = '<div class="a">not-a-child-div</div><div id="myId"></div>';
+      await page.setContent(htmlContent);
+      const elementHandle = await page.$('#myId');
+      const nodesLength = await elementHandle.$$eval('.a', nodes => nodes.length);
+      expect(nodesLength).toBe(0);
+    });
+
   });
 
   describe('ElementHandle.$$', function() {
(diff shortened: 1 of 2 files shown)


test/elementhandle.spec.js
```diff
@@ -336,14 +336,6 @@ module.exports.addTests = function({testRunner, expect}) {
       const content = await elementHandle.$$eval('.a', nodes => nodes.map(n => n.innerText));
       expect(content).toEqual(['a1-child-div', 'a2-child-div']);
     });
-
-    it('should throw in case of missing selector', async({page, server}) => {
```
aslushnikov: let's keep the test and make sure it doesn't throw in case of missing selector

JoelEinbinder (approved):

aslushnikov (approved):
Thanks @yanivefraim, this is a nice patch. 

P.S. I  see you pinged me in the issue. Sorry for not responding: didn't have time to go through GH notifications since I returned from vacation yesterday.

test/elementhandle.spec.js
yanivefraim: @aslushnikov - done, 10x!

PR #2742, opened 2018-06-14
feat: nicer protocol error messages

JoelEinbinder:
This stops appending `undefined` to our protocol messages unnecessarily.

It also rewrites `Cannot find execution context id` to `Execution context was destroyed, most likely because of a navigation.` when it occurs from a Puppeteer ExecutionContext. The error message is left alone if it occurs via a CDPSession.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -108,7 +108,7 @@ class Connection extends EventEmitter {
       if (callback) {
         this._callbacks.delete(object.id);
         if (object.error)
-          callback.reject(rewriteError(callback.error, `Protocol error (${callback.method}): ${object.error.message} ${object.error.data}`));
+          callback.reject(createProtocolError(callback.error, callback.method, object));
         else
           callback.resolve(object.result);
       }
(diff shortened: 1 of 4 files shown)


aslushnikov (approved):

sejoker:
@JoelEinbinder I found this PR by searching "'Execution context was destroyed, most likely because of a navigation." Any hint on how to identify a root cause of such error will be greatly appreciated.

carlfost:
> @JoelEinbinder I found this PR by searching "'Execution context was destroyed, most likely because of a navigation." Any hint on how to identify a root cause of such error will be greatly appreciated.

Did you get an answer to your question? I am having the same issue.

JoelEinbinder:
This error is caused when you try to evaluate some JavaScript on a page, but the page has unexpectedly navigated causing all JavaScript execution to be interrupted.

This is usually caused by not awaiting page.goto or related commands that navigate the browser. Or the page navigates itself via JavaScript, clicking on a link, etc.

carlfost:
I noticed that if I do these same codes outside of the For Next loop, it
works, but failed within the loop.

PR #2743, opened 2018-06-14
feat(chromium): roll Chromium to r567388

aslushnikov:
This roll includes:
- https://crrev.com/567104 - DevTools: introduce Target.exposeDevToolsProtocol() method

The patch includes a drive-by fix to DevToolsAgentHostImpl that
eliminats chromium crashes in certain cases.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "node": ">=6.4.0"
   },
   "puppeteer": {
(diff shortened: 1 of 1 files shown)


JoelEinbinder:
Where’s the drive by fix?

JoelEinbinder (approved):

PR #2749, opened 2018-06-15
feat(launcher): add disable gpu option

### LINKED ISSUE #1260: GPU is always disabled when running in headless mode.
I'm trying to use puppeteer to interact with a webpage that requires WebGL support. When puppeteer launches Chromium in headless mode, the `--disable-gpu` flag is always used. Based on https://bugs.chromium.org/p/chromium/issues/detail?id=737678#c1, that flag is not necessary on Linux or MacOS anymore.

### Steps to reproduce

* Puppeteer version: 0.12.0
* Platform / OS version: macOS 10.12.6

1. Execute the below script (used node 8.9.0 locally).

```javascript

(async () => {
    // headless: false,
  });

  // Test for webgl support
  // e.g. https://developer.mozilla.org/en-US/docs/Learn/WebGL/By_example/Detect_WebGL
  const webgl = await page.evaluate(() => {
    const canvas = document.createElement('canvas');
    const gl = canvas.getContext('webgl');
    const expGl = canvas.getContext('experimental-webgl');

    return {
      gl: gl && gl instanceof WebGLRenderingContext,
      expGl: expGl && expGl instanceof WebGLRenderingContext,
    };
  });

  console.log('WebGL Support:', webgl);

  await browser.close();
})();
```

`WebGL Support: { gl: true, expGl: true }`

**What happens instead?**
`WebGL Support: { gl: null, expGl: null }`

abelvarga:
Fixes #1260

The idea is to be able to run test in `headless: true` utilising GPU.  For backwards compatibility default behaviour has not changed. To have a symmetric interface, one will be able to disable GPU in `headless: false` too.

We need this capability to be able to test our integration with `mapbox-gl-js`, which would not run on software render.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -347,6 +347,7 @@ This methods attaches Puppeteer to an existing Chromium instance.
 - `options` <[Object]>  Set of configurable options to set on the browser. Can have the following fields:
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
   - `headless` <[boolean]> Whether to run browser in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). Defaults to `true` unless the `devtools` option is `true`.
+  - `disableGPU` <[boolean]> Whether to disable GPU. Defaults to `false` unless `headless` option is set to `true`. Set to `false` will force usage of GPU in headless mode.
   - `executablePath` <[string]> Path to a Chromium or Chrome executable to run instead of the bundled Chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
   - `slowMo` <[number]> Slows down Puppeteer operations by the specified amount of milliseconds. Useful so that you can see what is going on.
   - `args` <[Array]<[string]>> Additional arguments to pass to the browser instance. The list of Chromium flags can be found [here](http://peter.sh/experiments/chromium-command-line-switches/).
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -97,7 +97,8 @@ class Launcher {
 
       if (options.disableGPU !== false)
         chromeArguments.push('--disable-gpu');
-    } else if (options.disableGPU === false) {
+
+    } else if (options.disableGPU === true) {
       chromeArguments.push('--disable-gpu');
     }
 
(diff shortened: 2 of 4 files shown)


abelvarga:
I signed it!

abelvarga:
Any ideas why the AppVeyour build fails for me?

aslushnikov:
> Any ideas why the AppVeyour build fails for me?

Hey @abelvarga, appveyor fails due to #2778. I'll address it separately.

Regarding the PR itself - you can enable GPU by filtering out the `--disable-gpu` from [`puppeteer.defaultArgs`](https://pptr.dev/#?product=Puppeteer&version=v1.5.0&show=api-puppeteerdefaultargs) and passing `ignoreDefaultArgs` to [`puppeteer.launch`](https://pptr.dev/#?product=Puppeteer&version=v1.5.0&show=api-puppeteerlaunchoptions):

```js
const args = puppeteer.defaultArgs().filter(arg => arg !== '--disable-gpu');
  ignoreDefaultArgs: true,
  args
});
```

Does it solve your issue, or am I missing something?

abelvarga:
Hi @aslushnikov,

sorry for getting back this late.

I tried that as my first attempt before looking under the hood, but it's not in `defaultArgs`. It is set by the launch script if `headless: true`. I found no way to circumvent that.

JoelEinbinder:
I think we should fix `Emulation.setDefaultBackgroundColorOverride` and any other problems, and then just enable the GPU, instead of baking this bug into the puppeteer API.

The `ignoreDefaultArgs` flag looks additionally broken. `puppeteer.defaultArgs` needs to take an options object. But as a workaround you can do `{headless: false, args: ['--headless']}` to have headless mode with the GPU enabled.

abelvarga:
Hey @JoelEinbinder,

thanks for your suggestion, I'll try that one out today and get back.

But on a general note, I think it's beneficial to have a symmetric API for CLI args and options. By which I mean every CLI argument should have a JS equivalent.

And after all bugs are fixed, it would still have purpose.

abelvarga:
@JoelEinbinder, I can confirm that it works!

But I still think it would be beneficial to have an option for disabling GPU. After fixing the bugs you mentioned, the extra conditions could be removed, along with the tests for those.

aslushnikov:
> But I still think it would be beneficial to have an option for disabling GPU. After fixing the bugs you mentioned, the extra conditions could be removed, along with the tests for those.

We don't want to add every possible flag to launch options. Instead we should fix the `defaultArgs()`. Closing this in favor of #2873.

PR #2771, opened 2018-06-19
fix(request): extend error reasons based on missing values from spec

Mickael-van-der-Beek:
As the title indicated, adds two allowed error codes that are specified in the Chrome Developer Tools Protocol specifications but missing from the Puppeteer enum:

https://chromedevtools.github.io/devtools-protocol/tot/Network#type-ErrorReason

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -506,6 +506,8 @@ const errorReasons = {
   'aborted': 'Aborted',
   'accessdenied': 'AccessDenied',
   'addressunreachable': 'AddressUnreachable',
+  'blockedbyclient': 'BlockedByClient',
+  'blockedbyresponse': 'BlockedByResponse',
   'connectionaborted': 'ConnectionAborted',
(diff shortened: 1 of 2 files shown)


Mickael-van-der-Beek:
@googlebot Signed it.

lib/NetworkManager.js
```diff
@@ -506,6 +506,8 @@ const errorReasons = {
   'aborted': 'Aborted',
   'accessdenied': 'AccessDenied',
   'addressunreachable': 'AddressUnreachable',
+  'blockedbyclient': 'BlockedByClient',
```
aslushnikov, line 509: can you please update documentation for the `request.abort` method in `docs/api.md`:

https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#requestaborterrorcode

Mickael-van-der-Beek: @aslushnikov Sure. Done.

aslushnikov (approved):
Awesome, thanks!

PR #2774, opened 2018-06-19
chore(ci): Configure OSX builds on Cirrus CI

fkorotkov:
Cirrus CI recently started supporting Mac OS builds. This PR adds a CI task to test Puppeteer on MacOS.

It uses Node 8 since only [Node 6 and Node 8 are LTS](https://github.com/nodejs/Release#release-schedule) so `brew` only has receipts for those two TLS versions besides the latest one.


aslushnikov (approved):
Awesome!

PR #2776, opened 2018-06-20
feat(page): created waitForRequest and waitForResponse methods

### LINKED ISSUE #2362: [Feature] Add page.waitForRequest() and page.waitForResponse()
Hey! 

After using Puppeteer for a little while now in a number of different scenarios (SSR, automated testing) a common piece of functionality that often came up as needed, or at very minimum a nice to have, was the ability to wait for a specific network request to be made, or alternatively for a specific network response to come back. As a result I propose the addition of the following two functions.

```javascript
page.waitForRequest( match[, options])
```

- `match` \<string\> A regular expression to match against the request URL
- `options` \<Object\> Optional parameters
  - `method` \<string|Array\> Either a single HTTP request method (GET, POST, etc) as a string or an array of values `["GET", "POST"]`. Defaults to all HTTP request methods.
  - `timeout` \<number\> Maximum time to wait in milliseconds

And additionally

```javascript
page.waitForResponse( match[, options])
```
With a similar set of options. I'd probably also additionally add support for filtering by response code.

**An example**

```javascript
await page.waitForRequest( 'https?://example.com/(users|companies)', { method: 'POST' });
```

Happy to submit a PR if you all think it's a worthwhile addition!

Thanks!

matthewshirley:
Resolves #2362

Introduced waitForRequest and waitForResposne methods with the ability to handle a URL as a String or RegExp. My implementation is heavily based on the waitForNavigation method with a new file NetworkWatcher that works exactly like NavigatorWatcher.


matthewshirley:
I signed it!

docs/api.md
```diff
@@ -1569,6 +1571,29 @@ await navigationPromise; // The navigationPromise resolves after navigation has
 
 **NOTE** Usage of the [History API](https://developer.mozilla.org/en-US/docs/Web/API/History_API) to change the URL is considered a navigation.
 
+#### page.waitForRequest(url, options)
+- `url` <[function]|[string]> String or RegExp of a URL of the request to wait for.
```
vsemozhetbyt: Should this be `` <[RegExp]|[string]> `` with a new bottom reference added?

docs/api.md
```diff
@@ -1569,6 +1571,29 @@ await navigationPromise; // The navigationPromise resolves after navigation has
...
+  - `method` <[string]> A HTTP method to match on a request. 
+- returns: <[Promise]<[Request]>> Promise which resolves to the matched request.
+
+```js
+const request = await page.waitForRequest(new RegExp('http://localhost/(method|post)'));
+return request.url();
+```
+
+#### page.waitForResponse(url, options)
+- `url` <[function]|[string]> String or RegExp of a URL of the response to wait for.
```
vsemozhetbyt: Should this be `` <[RegExp]|[string]> ``?

docs/api.md
```diff
@@ -1569,6 +1571,29 @@ await navigationPromise; // The navigationPromise resolves after navigation has
 
 **NOTE** Usage of the [History API](https://developer.mozilla.org/en-US/docs/Web/API/History_API) to change the URL is considered a navigation.
 
+#### page.waitForRequest(url, options)
+- `url` <[function]|[string]> String or RegExp of a URL of the request to wait for.
+- `options` <[Object]> Optional waiting parameters
+  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds, pass `0` to disable timeout.
+  - `method` <[string]> A HTTP method to match on a request.
```
vsemozhetbyt: A HTTP -> An HTTP?

docs/api.md
```diff
@@ -1569,6 +1571,29 @@ await navigationPromise; // The navigationPromise resolves after navigation has
...
+```js
+const request = await page.waitForRequest(new RegExp('http://localhost/(method|post)'));
+return request.url();
+```
+
+#### page.waitForResponse(url, options)
+- `url` <[function]|[string]> String or RegExp of a URL of the response to wait for.
+- `options` <[Object]> Optional waiting parameters
+  - `timeout` <[number]> Maximum navigation time in milliseconds, defaults to 30 seconds, pass `0` to disable timeout.
+- returns: <[Promise]<[Response]>> Promise which resolves to the matched request.
```
vsemozhetbyt: request -> response?

docs/api.md
```diff
@@ -1569,6 +1571,29 @@ await navigationPromise; // The navigationPromise resolves after navigation has
...
+
+```js
+const request = await page.waitForResponse('http://localhost/getData');
```
vsemozhetbyt: request -> response?

docs/api.md
```diff
@@ -1569,6 +1571,29 @@ await navigationPromise; // The navigationPromise resolves after navigation has
...
+return request.ok();
```
vsemozhetbyt: request -> response?

matthewshirley:
Thanks, @vsemozhetbyt for the feedback. I've made changes to the reviewed areas.

I've updated the test case which hopefully passes AppVeyor this time around. (AppVeyor is failing for something unrelated).

matthewshirley:
Rebased my fork and AppVeyor is still failing after PR #2779 was merged. Is the failing test related to my changes?

lib/Page.js
```diff
@@ -88,6 +89,7 @@ class Page extends EventEmitter {
     this._ignoreHTTPSErrors = ignoreHTTPSErrors;
     this._coverage = new Coverage(client);
     this._defaultNavigationTimeout = 30000;
+    this._defaultNetworkTimeout = 30000;
```
aslushnikov: let's drop this for now

lib/Page.js
```diff
@@ -632,6 +634,62 @@ class Page extends EventEmitter {
...
 
+  /**
+   * @param {!Object=} options
+   * @return {!Promise<!Puppeteer.Request>}
+   */
+  async waitForRequest(url, options = {}) {
+    let matchedRequest;
+
+    const timeout = typeof options.timeout === 'number' ? options.timeout : this._defaultNetworkTimeout;
+    const watcher = new NetworkWatcher(this._networkManager, 'request', url, timeout, options);
```
aslushnikov: @matthewshirley I see the `NetworkWatcher` is modeled after `NavigationWatcher`. However, I believe this design is not needed in this case. Let me suggest an alternative approach.

Consider there's a function `waitForEvent` that returns a promise, such as:
- promise is fulfilled if `emitter` emits `event` that matches `predicate`
- promise is rejected if such an event is not emitted during `timeout` milliseconds

A draft implementation might look like this: 
```js
function waitForEvent(emitter, eventName, predicate, timeout) {
  let resolveCallback, rejectCallback;
  const promise = new Promise((resolve, reject) => {
    resolveCallback = resolve;
    rejectCallback = reject;
  });
  const listener = helper.addEventListener(emitter, eventName, event => {
    if (!predicate(event))
      return;
    cleanup();
    resolveCallback(event);
  });
  const timeoutId = setTimeout(() => {
    cleanup();
    rejectCallback(new Error('Timeout exceeded while waiting on request'));
  }, timeout);

  return promise;

  function cleanup() {
    helper.removeEventListeners([listener]);
    clearTimeout(timeoutId);
  }
}
```

Given this function, both `waitForRequest` and `waitForResponse` would be implementable like this:
```js
async waitForRequest(url, options = {}) {
  const timeout = typeof options.timeout === 'number' ? options.timeout : 30000;
  return waitForEvent(this._newtorkManager, NetworkManager.Events.Request, request => {
    return url === request.url() || (url instanceof RegExp && url.test(request.url()));
  }, timeout);
}

async waitForResponse(url, options = {}) {
  const timeout = typeof options.timeout === 'number' ? options.timeout : 30000;
  return waitForEvent(this._newtorkManager, NetworkManager.Events.Response, response => {
    return url === response.url() || (url instanceof RegExp && url.test(response.url()));
  }, timeout);
}
```

aslushnikov:
> Rebased my fork and AppVeyor is still failing after PR #2779 was merged. Is the failing test related to my changes?

These are not related to your change; we unfortunately have some flakiness in the testsuite.

lib/Page.js
matthewshirley: Oh, awesome. Reviewing back my original method was a bit too much. That's a solid draft implementation and I couldn't find much to change. I've put this into the helper file.

I've reviewed the original issue raised and added the method option to both waitForRequest and waitForResponse and added statusCode to waitForResponse. Both can take an array.

Thanks!

lib/helper.js
```diff
@@ -240,6 +240,40 @@ class Helper {
...
+    });
+
+    const listener = Helper.addEventListener(emitter, eventName, event => {
+      if (!predicate(event)) return;
+
+      cleanup();
+      resolveCallback(event);
+    });
+
+    const eventTimeout = setTimeout(() => {
```
aslushnikov: We should teach this to support timeout `0`, otherwise this diverges with docs

lib/helper.js
```diff
@@ -240,6 +240,40 @@ class Helper {
...
+  static waitForEvent(emitter, eventName, predicate, timeout) {
+    let resolveCallback, rejectCallback;
+
+    const promise = new Promise((resolve, reject) => {
+      resolveCallback = resolve;
+      rejectCallback = reject;
+    });
+
+    const listener = Helper.addEventListener(emitter, eventName, event => {
+      if (!predicate(event)) return;
```
aslushnikov: style: let's have `return` on a new line
```js
if (!predicate(event))
  return;
```

lib/Page.js
```diff
@@ -632,6 +632,54 @@ class Page extends EventEmitter {
     return responses.get(this.mainFrame().url()) || null;
   }
 
+  /**
+   * @param {!Object=} options
+   * @return {!Promise<!Puppeteer.Request>}
+   */
+  async waitForRequest(url, options = {}) {
+    let method;
+
```
aslushnikov: nit: let's drop the newline

lib/Page.js
```diff
@@ -632,6 +632,54 @@ class Page extends EventEmitter {
...
+    if (Array.isArray(options.method))
+      method = options.method;
+    else if (typeof method === 'string')
+      method = [options.method];
+
```
aslushnikov: nit: let's drop the newline

lib/Page.js
```diff
@@ -632,6 +632,54 @@ class Page extends EventEmitter {
...
+
+    const timeout = typeof options.timeout === 'number' ? options.timeout : 30000;
+
+    return helper.waitForEvent(this._networkManager, NetworkManager.Events.Request, request => {
+      if (url === request.url() || (url instanceof RegExp && url.test(request.url())))
+        return method ? method.includes(request._method) : true;
+
```
aslushnikov: nit: let's drop the newline

lib/Page.js
```diff
@@ -632,6 +632,54 @@ class Page extends EventEmitter {
...
+    let method;
+
+    if (Array.isArray(options.method))
+      method = options.method;
+    else if (typeof method === 'string')
+      method = [options.method];
+
+
+    const timeout = typeof options.timeout === 'number' ? options.timeout : 30000;
+
```
aslushnikov: nit: let's drop the newline

lib/Page.js
```diff
@@ -632,6 +632,54 @@ class Page extends EventEmitter {
...
+    return helper.waitForEvent(this._networkManager, NetworkManager.Events.Request, request => {
+      if (url === request.url() || (url instanceof RegExp && url.test(request.url())))
+        return method ? method.includes(request._method) : true;
```
aslushnikov: nit: let's use public `request.method()`

test/page.spec.js
```diff
@@ -781,6 +781,126 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
     });
   });
 
+  describe('Page.waitForRequest', function() {
+    it('should work', async({page, server}) => {
+      await page.goto(server.EMPTY_PAGE);
+
+      server.setRoute('/getRequest', (req, res) => {
```
aslushnikov: there's no need to do `setRoute` for server: it's fine to fetch non-existing resources (and get 404)

test/page.spec.js
```diff
@@ -781,6 +781,126 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+        res.statusCode = 200;
+        res.end();
+      });
+
+      page.evaluate(() => {
```
aslushnikov: In order to avoid race condition between evaluation completion and page closing in the end of the test, we should `await` all our promises that interact with the page.

With the comment above, I'd write this test in the following way:

```js
it('should work', async({page, server}) => {
  const [request] = await Promise.all([
    page.waitForRequest(server.PREFIX + '/404'),
    page.evaluate(() => fetch('/404', {method: 'GET'}))
  ]);
  expect(request.method()).toBe('GET');
  expect(request.url()).toBe(server.PREFIX + '/404');
});
```

test/page.spec.js
```diff
@@ -781,6 +781,126 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
...
+      const response = await page.waitForResponse(`${server.PREFIX}/testResponse`);
+
+      expect(response.ok()).toBe(true);
+      expect(response.url()).toBe(`${server.PREFIX}/testResponse`);
+    });
+
+    it('should work with regex', async({page, server}) => {
+      await page.goto(server.EMPTY_PAGE);
+
+      server.setRoute('/testResponse', (req, res) => serverResponse = () => res.end('Hello'));
```
aslushnikov: you can avoid setting routes if you just query for some of the resources in the `test/assets/` folder

docs/api.md
```diff
@@ -1569,6 +1571,31 @@ await navigationPromise; // The navigationPromise resolves after navigation has
 
 **NOTE** Usage of the [History API](https://developer.mozilla.org/en-US/docs/Web/API/History_API) to change the URL is considered a navigation.
 
+#### page.waitForRequest(url, options)
+- `url` <[string]|[RegExp]> A URL to wait for.
+- `options` <[Object]> Optional waiting parameters
+  - `timeout` <[number]> Maximum wait time in milliseconds, defaults to 30 seconds, pass `0` to disable the timeout.
```
aslushnikov: let's add a test that validates timeout `0`.
The test should do a 50ms delay before responding to a request; this way we'll demonstrate that timeout 0 is not handled literally.

matthewshirley: Awesome. Thanks for the feedback. I've made some changes and added timeout = 0 test case.

lib/Page.js
```diff
@@ -632,6 +632,47 @@ class Page extends EventEmitter {
...
 
+  /**
+   * @param {!Object=} options
+   * @return {!Promise<!Puppeteer.Request>}
+   */
+  async waitForRequest(url, options = {}) {
+    let method;
+    if (Array.isArray(options.method))
+      method = options.method;
+    else if (typeof method === 'string')
```
aslushnikov: I think there's a typo, you want to run `typeof` against `options.method` instead of `method`:
```js
else if (typeof options.method === 'string')
```

It's unfortunate tests didn't catch this. Can we update them?

lib/Page.js
```diff
@@ -632,6 +632,47 @@ class Page extends EventEmitter {
...
+  /**
+   * @param {!Object=} options
+   * @return {!Promise<!Puppeteer.Response>}
+   */
+  async waitForResponse(url, options = {}) {
+    let method;
+    let statusCode;
+    if (Array.isArray(options.method))
+      method = options.method;
+    else if (typeof method === 'string')
```
aslushnikov: ditto: use `typeof options.method` instead of `typeof method`

Let's test this as well

lib/Page.js
```diff
@@ -632,6 +632,47 @@ class Page extends EventEmitter {
...
+      method = [options.method];
+    if (Array.isArray(options.statusCode))
+      statusCode = options.statusCode;
+    else if (typeof method === 'string')
```
aslushnikov: Same here:
```js
else if (typeof options.statusCode === 'string')
```

A test?

lib/Page.js
aslushnikov: Here and in other places in this patch: let's use `helper.isString(options.method)` instead of `typeof options.method === 'string'`.

aslushnikov (commented):
@matthewshirley: Sorry for this taking so long: summer!

There are a few bugs in the patch - please, see inline comments.

While reviewing this, I didn't much like the `statusCode` and `method` options - these look like a long-tail features to me. We might end up with many more like this over time - imagine someone would like to wait for request based on its certificate.

Instead, I'd suggest supporting predicate as the first parameter:
```js
page.waitForRequest(urlOrPredicate, options);
```

This way we'll end up with a powerful API that supports all the usecases. For example, if one wants to wait for a specific URL and status code:

```js
await page.waitForRequst(request => {
  return /example/.test(request.url()) && request.status() === 302;
}, {timeout: 10000});
```

matthewshirley:
@aslushnikov and done - it looks much better. Thanks!

aslushnikov (approved):
Thanks!

Issue #2778, opened 2018-06-20
Disable crash reporting by default


aslushnikov:
Win CI usually fails to cleanup userDataDir folder in some tests. Example: [appveyour build](https://ci.appveyor.com/project/aslushnikov/puppeteer/build/1.0.1672/job/j2bllv92su3uajye)

It turns out that on Windows:
- crash pad  is launched in a separate process
- crash pad writes a few files into userDataDir
- crash pad process doesn’t necessarily shutdown when we close browser.

As a result, we fail to remove temporary user data directory.

We need to disable crash reporting by default. Somehow chrome headless ignores the `--disable-breakpad` flag and launches crashpad anyway; this requires investigation and fixing upstream.

Sunbreak:
Does `--disable-breakpad` work on Apple Silicon?

alamothe:
bump

alejandrod-f:
I can confirm that `--disable-breakpad` on ubuntu has the same problem. Is a major issue as looks like all works for creating PDFs but there is an error

Joaquin-Boilet:
Tengo exactamente le mismo problema al iniciar multiples ventanas desde xdge , url

sebastianhelbig:
Any progress on this?

shunkica:
I am running with args `"--single-process", "--disable-crash-reporter", "--disable-breakpad"` but it starts two `chrome_crashpad_handler` processes anyway.

overtube:
have the same issue too that is a binary file that launched from /usr/lib, and also if remove that file, browser that won't start anymore, also flags cannot work too, also settings for disable that behavior won't found too.

gregg-cbs:
Same issue here, every time we close the browser and spawn a new one using puppeteer api it spawns a new chrome_crashpad_handler. Explains why our memory is creeping:
![image](https://github.com/puppeteer/puppeteer/assets/57114335/94ed8a04-1134-4fae-a1e1-383e3f8c260a)

autopulated:
The  `--disable-crash-reporter` flag appears to only disable crash reporting, it doesn't do anything to disable the crashpad crash handling library (and subprocesses). `--disable-crashpad-for-testing` is a switch that does do that for the base chrome process, but this flag does not get propagated to child processes so it cannot stop them starting their own crashpad_handler. In combination with `--single-process` this might solve your problem.

[...]

PR #2779, opened 2018-06-20
test: fix appveyour flakiness

aslushnikov:
Crashpad sometimes prevents us from removing a folder in tests.

References #2778.


aslushnikov:
TBR: @JoelEinbinder

PR #2795, opened 2018-06-23
feat(worker): implement pageerror event from a worker

### LINKED ISSUE #2761: Web Worker errors are not collected by `page.on('pageerror'...`
When an error occures on a Web Worker, it is not collected by `page.on('pageerror'...` event. 

For example:

```js
(async () => {
  page.on('pageerror', message => console.log(message));

  await page.setContent(`<!DOCTYPE html>
  <script id="worker1" type="javascript/worker">
    throw new Error('this is my error');
  </script>
  <script>
    const blob = new Blob([
      document.querySelector('#worker1').textContent
    ], { type: "text/javascript" })

    new Worker(window.URL.createObjectURL(blob));
  </script>`);
})()
```

yanivefraim:
This will fix #2761

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -102,7 +102,7 @@ class Page extends EventEmitter {
         return;
       }
       const session = client._createSession(event.targetInfo.type, event.sessionId);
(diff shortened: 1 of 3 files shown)


aslushnikov (approved):
Thanks!

PR #2796, opened 2018-06-24
feat(coverage): add an option to collect coverage of anonymous scripts

### LINKED ISSUE #2777: Can't collect coverage of anonymous scripts
Please, add option to enable covering of anonymous scripts.

https://github.com/GoogleChrome/puppeteer/blob/dcae6bcff49150bec2060667fef7bd7ab71db042/lib/Coverage.js#L111-L114

Change it to somethink like that:
``` 
    if (!event.url && this.ignoreAnonymousScripts)
      return;
```
where `this.ignoreAnonymousScripts` defines as `true` by default to do not break current behavior.

pomerantsev:
Resolves #2777


JoelEinbinder (approved):
Looks great! Thanks

pomerantsev:
Great!
On a separate note, shouldn't the meaning of "anonymous scripts" be explained somewhere in the docs? Or is it just me? Googling around didn't give me an answer, and only after looking at the tests did I realize that it's the scripts injected via `page.evaluate`.

JoelEinbinder:
Anonymous scripts are anything that doesn’t have a url. So puppeteer’s page.evaluate, javascript’s eval, or ‘new Function’ will create them. They show up in DevTools as something like ‘VM 38’.

pomerantsev:
Thanks! It would be helpful to have that in the docs. I can add it in a separate PR.
@JoelEinbinder do you know if there's a definition of "anonymous scripts" anywhere, or is it just puppeteer-speak?

aslushnikov:
@pomerantsev that's a devtools-speak :) Please, feel free to add this to the docs!

test/coverage.spec.js
```diff
@@ -38,13 +38,20 @@ module.exports.addTests = function({testRunner, expect}) {
...
       await page.coverage.startJSCoverage();
       await page.goto(server.EMPTY_PAGE);
       await page.evaluate(() => console.log(1));
       const coverage = await page.coverage.stopJSCoverage();
       expect(coverage.length).toBe(0);
     });
+    it('shouldn\'t ignore anonymous scripts if ignoreAnonymousScripts is false', async function({page, server}) {
+      await page.coverage.startJSCoverage({ignoreAnonymousScripts: false});
+      await page.goto(server.EMPTY_PAGE);
+      await page.evaluate(() => console.log(1));
```
aslushnikov: That's a nice test since it exploits an issue! We don't want puppeteer evaluations to be surfaced in coverage: they are meaningless for coverage clients and might be quite noisy.

To workaround this, we should add `// sourceURL=puppeteer_evaluation_script.js` to the end of all our evaluations and filter them out in coverage.

pomerantsev: So that `ignoreAnonymousScripts` only adds `eval` and `new Function` scripts to coverage, but not the puppeteer-injected ones, correct?

aslushnikov: > So that ignoreAnonymousScripts only adds eval and new Function scripts to coverage, but not the puppeteer-injected ones, correct?

@pomerantsev yes, exactly

pomerantsev: Cool, I'll work on it.

pomerantsev: I'm almost done implementing the above locally, but there's one unrelated test that got broken in the process: https://github.com/GoogleChrome/puppeteer/blob/d6741eca083cf20a2debd78ed43e7c257b217b0e/test/CDPSession.spec.js#L43-L57

`event.url` now has the value of the new `sourceURL` that we are appending to every injected script, and not `foo.js`.

@aslushnikov is there an obvious way to fix the broken test? I'm happy to find a solution myself, it'll just take me some time to figure out what's being tested in the first place.

pomerantsev: Never mind, I think I figured it out. Please review the update when you get a chance.

lib/ExecutionContext.js
```diff
@@ -61,9 +61,11 @@ class ExecutionContext {
    * @return {!Promise<!JSHandle>}
    */
   async evaluateHandle(pageFunction, ...args) {
+    const suffix = '//# sourceURL=__puppeteer_evaluation_script__';
+
     if (helper.isString(pageFunction)) {
       const contextId = this._contextId;
-      const expression = /** @type {string} */ (pageFunction);
+      const expression = /** @type {string} */ (pageFunction + '\n' + suffix);
```
aslushnikov: Let's append suffix only if there's no sourceURL magic comment in `pageFunction`. This way you won't need to change the test that evaluates a script with a source URL.

You can check for sourceURL comment quite reliably without parsing `pageFunction`'s AST; take a look on how it's done in DevTools front-end for CSS:
- [third_party/blink/renderer/devtools/front_end/sdk/CSSModel.js:77](https://cs.chromium.org/chromium/src/third_party/blink/renderer/devtools/front_end/sdk/CSSModel.js?q=sourceURLRegex+f:front_end&dr=CSs&l=77-92)

You can also get the DevTools regex for sourceURL comment for js:
- [third_party/blink/renderer/devtools/front_end/sdk/Script.js:264](https://cs.chromium.org/chromium/src/third_party/blink/renderer/devtools/front_end/sdk/Script.js?q=sourceURLRegex+f:front_end&l=264&dr=C)

Let me know if DevTools code is confusing to read!

test/CDPSession.spec.js
```diff
@@ -50,10 +50,10 @@ module.exports.addTests = function({testRunner, expect}) {
       // generate a script in page and wait for the event.
       const [event] = await Promise.all([
         waitEvent(client, 'Debugger.scriptParsed'),
-        page.evaluate('//# sourceURL=foo.js')
```
aslushnikov: let's keep this hidden feature: it helps a lot with chromium debugging.

docs/api.md
```diff
@@ -2933,8 +2933,11 @@ _To output coverage in a form consumable by [Istanbul](https://github.com/istanb
 #### coverage.startJSCoverage(options)
 - `options` <[Object]>  Set of configurable options for coverage
   - `resetOnNavigation` <[boolean]> Whether to reset coverage on every navigation. Defaults to `true`.
+  - `ignoreAnonymousScripts` <[boolean]> Whether anonymous scripts generated by the page should be ignored. Defaults to `true`.
 - returns: <[Promise]> Promise that resolves when coverage is started
 
+> **NOTE** Anonymous scripts are ones that don't have an associated url. These are scripts that are either dynamically created on the page via `eval` or `new Function`, or ones injected by puppeteer using `page.evaluate`. `ignoreAnonymousScripts: false` means that the page-generated ones (`eval` and `new Function`) will be included in the coverage. Puppeteer-injected ones will always be ignored.
```
aslushnikov: I'd slightly strip this down:

> **NOTE** Anonymous scripts are ones that don't have an associated url. These are scripts that are dynamically created on the page using `eval` or `new Function`. If `reportAnonymousScript` is set to `true`, anonymous scripts will have a URL starting with `debugger://VM`.

docs/api.md
```diff
@@ -2933,8 +2933,11 @@ _To output coverage in a form consumable by [Istanbul](https://github.com/istanb
 #### coverage.startJSCoverage(options)
 - `options` <[Object]>  Set of configurable options for coverage
   - `resetOnNavigation` <[boolean]> Whether to reset coverage on every navigation. Defaults to `true`.
+  - `ignoreAnonymousScripts` <[boolean]> Whether anonymous scripts generated by the page should be ignored. Defaults to `true`.
```
aslushnikov: nit: let's call this `reportAnonymousScripts`and default it to `false`.

lib/Coverage.js
```diff
@@ -109,8 +110,11 @@ class JSCoverage {
    * @param {!Protocol.Debugger.scriptParsedPayload} event
    */
   async _onScriptParsed(event) {
-    // Ignore anonymous scripts
-    if (!event.url)
+    // Ignore puppeteer-injected scripts
+    if (event.url === '__puppeteer_evaluation_script__')
```
aslushnikov: let's extract `__puppeteer_evaluation_script__` into a constant and put it into `ExecutionContext.js`

lib/ExecutionContext.js
```diff
@@ -76,8 +78,23 @@ class ExecutionContext {
...
+    const functionDeclarationInBraces = originalFunctionDeclaration[originalFunctionDeclaration.length - 1] === '}' ?
+      originalFunctionDeclaration :
+      (() => {
+        const indexAfterArrow = originalFunctionDeclaration.indexOf('=>') + 2;
+        return originalFunctionDeclaration.slice(0, indexAfterArrow) + '{ return ' + originalFunctionDeclaration.slice(indexAfterArrow) + '}';
+      })();
+    const declarationWithSourceURL = functionDeclarationInBraces.slice(0, functionDeclarationInBraces.length - 1) + '\n' + suffix + '\n}';
     const { exceptionDetails, result: remoteObject } = await this._client.send('Runtime.callFunctionOn', {
-      functionDeclaration: pageFunction.toString(),
+      functionDeclaration: declarationWithSourceURL,
```
aslushnikov: instead of computing `declarationWithSourceURL` in this fancy way, you [should be able](https://cs.chromium.org/chromium/src/v8/src/inspector/v8-runtime-agent-impl.cc?type=cs&q=callFunctionOn&sq=package:chromium&g=0&l=146) to just pass sourceURL with a new line in the end:

```js
  // ...
  fuctionDeclaration: pageFunction.toString() + '\n//# sourceURL=somename.js\n',
  // ...
```
Trailing `\n` should do the trick.

lib/ExecutionContext.js
pomerantsev: I hadn't really looked at the DevTools source before, but the JS bit doesn't look confusing.

docs/api.md
pomerantsev: Sorry, but I didn't understand what `debugger://VM` stands for, so I put `__puppeteer_evaluation_script__` there instead - let me know if that needs fixing.

pomerantsev:
@aslushnikov hopefully I addressed all of your comments in my latest commit.

aslushnikov (approved):
@pomerantsev thank you for working on this!

PR #2806, opened 2018-06-27
fix(page): teach page.click() to click partially offscreen buttons

### LINKED ISSUE #2804: click does not fire.

* Puppeteer version: 1.5.0
* Platform / OS version: mac OSX 10.13.5
* URLs (if applicable):
* Node.js version: 8.11.3

I tried the following code. but click event does not fire.

    headless: false,
    slowMo: 450
  }); 
  await page.goto('https://www.yahoo.co.jp/')
  await page.click('.Util-textBold > a')
  await page.type('#username', 'test_user')
  await page.click('#btnNext')

// also I tried
// await page.click('button[type="submit"]')

click a '#btnNext'

**What happens instead?**
nothing happens.

aslushnikov:
Originally, we use `Element.scrollIntoViewIfNeeded` to make sure
button is on screen before trying to click it.

However, `Element.scrollIntoViewIfNeeded` doesn't work in certain
scenarios, e.g. when element is partially visible and horizontal
scrolling is required to make it fully visible.

This patch polyfills `element.scrollIntoViewIfNeeded` using
IntersectionObserver and `Element.scrollIntoView`.

Fixes #2804.

### DIFF
--- a/lib/ElementHandle.js
+++ b/lib/ElementHandle.js
@@ -67,7 +67,7 @@ class ElementHandle extends JSHandle {
         });
         observer.observe(element);
       });
-      if (Math.abs(visibleRatio - 1.0) > 1e-3)
+      if (visibleRatio !== 1.0)
         element.scrollIntoView({block: 'center', inline: 'center', behavior: 'instant'});
       return false;
     }, this);
--- a/lib/ElementHandle.js
+++ b/lib/ElementHandle.js
@@ -55,12 +55,20 @@ class ElementHandle extends JSHandle {
   }
 
   async _scrollIntoViewIfNeeded() {
-    const error = await this.executionContext().evaluate(element => {
+    const error = await this.executionContext().evaluate(async element => {
       if (!element.isConnected)
         return 'Node is detached from document';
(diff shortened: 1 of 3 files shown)


lib/ElementHandle.js
```diff
@@ -55,12 +55,20 @@ class ElementHandle extends JSHandle {
...
         return 'Node is not of type HTMLElement';
-      element.scrollIntoViewIfNeeded();
+      const visibleRatio = await new Promise(resolve => {
+        const observer = new IntersectionObserver(entries => {
+          resolve(entries[0].intersectionRatio);
+          observer.disconnect();
+        });
+        observer.observe(element);
+      });
+      if (Math.abs(visibleRatio - 1.0) > 1e-3)
```
JoelEinbinder: 1e-3 is scary. Lets just compare to 1

JoelEinbinder (approved):

transitive-bullshit:
Just wanted to say that it's an amazing feeling to run into an issue, track it down, and find out it's already been solved!

Keep up the great work, guys!! @aslushnikov @JoelEinbinder 💯

PR #2812, opened 2018-06-28
feat: extension execution contexts

JoelEinbinder:


docs/api.md
```diff
@@ -2314,6 +2335,11 @@ await resultHandle.dispose();
 
 > **NOTE** Not every execution context is associated with a frame. For example, workers and extensions have execution contexts that are not associated with frames.
 
+#### executionContext.name()
+- returns: <[string]> A human readable name for the execution context.
+
+Most execution contexts have an empty string for a name. Execution contexts created by extensions will have the title of the extension as thier name.
```
vsemozhetbyt, line 2341: thier -> their :)

docs/api.md
```diff
@@ -1116,6 +1131,9 @@ puppeteer.launch().then(async browser => {
 
 ```
 
+#### page.extensionContexts()
```
aslushnikov, line 1134: I don't like restricting us to extensions. Execution contexts can exist and might be handy in non-chromium embedders, e.g. in headless. 

From this point of view, the ideal API would be like this:
```js
// ExecutionContext lifecycle
page.on('executioncontextcreated');
page.on('executioncontextdestroyed');

// Querying  contexts
const contexts = frame.executionContexts();
const context = frame.defaultExecutionContext();

// Inspecting context
context.name(); // returns extension name, if any
context.isDefault(); // returns `true` if this is a default execution context
```

The only downside here is that we already have `frame.executionContext()` instead of `frame.defaultExecutionContext()`. I think we can keep it like this for now, and rename it later on in the API cleanup for v2.0.

JoelEinbinder: An api that contains all execution contexts is reasonable, but id still want helper methods specifically for extension contexts.

I’m wary of releasing an api that will today be effectively only extension contexts, but in the future might contain some more exotic contexts.

aslushnikov: > An api that contains all execution contexts is reasonable, but id still want helper methods specifically for extension contexts.

I'd move slowly into this direction. For the first-class extension support, we'll need to have an `Extension` class - there you might have methods like `extension.contentScript(frame)` and others.

aslushnikov:
Closing this in favor of https://github.com/GoogleChrome/puppeteer/pull/2812#discussion_r198923093

PR #2814, opened 2018-06-28
feat(chromium): roll Chromium to r571040

### LINKED ISSUE #1247: SVG element causes DOM.getBoxModel error
I've been having issues trying to hover an svg element using `elementHandle.hover()`, now using `page.evaluate(element => element.getBoundingClientRect(), elementHandle)` works so I'm not sure why it throws an error for `.hover()`.

Here's a minimal example that causes it:

[puppeteer-svg-hover-test.zip](https://github.com/GoogleChrome/puppeteer/files/1436011/puppeteer-svg-hover-test.zip)

aslushnikov:
This roll includes:
- https://crrev.com/570566 - DevTools: teach DOM.getBoxModel to work with SVG nodes

Fixes #1247.

### DIFF
--- a/test/assets/frames/frame.html
+++ b/test/assets/frames/frame.html
@@ -1,3 +1,8 @@
 <link rel='stylesheet' href='./style.css'>
 <script src='./script.js' type='text/javascript'></script>
+<style>
+div {
+  line-height: 18px;
+}
+</style>
 <div>Hi, I'm frame</div>
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "node": ">=6.4.0"
   },
   "puppeteer": {
-    "chromium_revision": "568432"
+    "chromium_revision": "571040"
   },
   "scripts": {
     "unit": "node test/test.js",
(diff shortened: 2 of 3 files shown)


test/assets/frames/frame.html
```diff
@@ -1,3 +1,8 @@
 <link rel='stylesheet' href='./style.css'>
 <script src='./script.js' type='text/javascript'></script>
+<style>
+div {
```
aslushnikov, line 4: Looks like font resolution has changed in chromium, and this HTML now has different default font on travis's VM.

JoelEinbinder: Interesting. Might be worth a PSA in the release notes.

JoelEinbinder (approved):

test/assets/frames/frame.html
aslushnikov: Interesting thought.

Font stack is already highly unpredictable and volatile, so I don't think this is worth mentioning.

Issue #2685, opened 2018-06-06
Page.pdf - Colors rendering


gary149:
When I run my page in headless mode every colors is rendered right.
But when I get the pdf with page.pdf() with no options some colors (a few in fact) are wrong - too bright or too dark - Is there any reason for this?

update: I solved this with this css (not sure why it wasn't working without it)

```css
html {
  -webkit-print-color-adjust: exact;
}
```

kaveet:
Thanks for sharing your solution--that wasn't a fun problem to debug.

The color changes appear to be part of the browser's strategy for reducing print cost and achieving more accurate colors on paper. It's part of the browser's printing functionality, which is why the colors only change when using `page.pdf()` as opposed to opening the page in a browser. This isn't really a problem with Puppeteer but really should be somewhere in its documentation.

You can read more about this here: [https://drafts.csswg.org/css-color-4/#color-adjust](https://drafts.csswg.org/css-color-4/#color-adjust)

aslushnikov:
>  This isn't really a problem with Puppeteer but really should be somewhere in its documentation.

@kaveet I'd be happy to review a pull request with this.

kaveet:
@aslushnikov Ready for your review, thanks!

IDCOLL:
> When I run my page in headless mode every colors is rendered right.
> But when I get the pdf with page.pdf() with no options some colors (a few in fact) are wrong - too bright or too dark - Is there any reason for this?
> 
> update: I solved this with this css (not sure why it wasn't working without it)
> 
> ```css
> html {
>   -webkit-print-color-adjust: exact;
> }
> ```
Hi,
Where do you pass the -webkit-print-color-adjust :exact property as listed in this thread?

In the puppeteer documentation it simply states:

> Use the -webkit-print-color-adjust property to force rendering of exact colors.

Do I pass it in the page.pdf() function? If so how would it look?
Any help would be appreciated.

kaveet:
@IDCOLL the property referenced is a CSS one—you’ll need to apply it from within your HTML source (inline or via a linked stylesheet). The easy solution is to set print color rendering to ‘exact’ for the entire page:

```html
<style>
  html {
    -webkit-print-color-adjust: exact;
  }
</style>
```

Hopefully that clears things up.

demobin8:
thx, save my time

ix-xerri:
[...]

PR #2821, opened 2018-06-29
docs(api.md): add note on page.pdf() color rendering behavior

kaveet:
Adds guidance for producing accurate colors in PDF output. page.pdf() can produce unexpected document colors unless forced to render exact colors.

References #2685


kaveet:
I signed it!

aslushnikov (approved):
Thanks!

PR #2822, opened 2018-06-29
fix(page.click): teach puppeteer click wrapped links

### LINKED ISSUE #2798: Possible bug in click using visible center
### Steps to reproduce

* Puppeteer version:1.5.0
* Platform / OS version:Ubuntu
* URLs (if applicable):https://www.handelsregister.de/rp_web/mask.do?Typ=e
* Node.js version:v8.10

https://stackoverflow.com/questions/51011466/puppeteer-select-link
Basically these https://bitbucket.org/ytNeskews/puppeteer-attempt/src/master/src/index.js

1.
2.
3.

The expectation is that page.click('p a') should trigger a click. While it appears there is no network request triggered

**What happens instead?**
Click happens but not on the link. Further debugging suggested that the visible center coordinates seem off. Once I changed the viewport of the page to 1366,768, the click successfully generated a navigation

Perhaps something to do with elements staggered across the page (in multiple lines if you will).

aslushnikov:
This patch teaches Puppeteer to click elements that are
part of inline layout and that wrap on multiple lines.

Fixes #2798.

### DIFF
--- a/test/assets/wrappedlink.html
+++ b/test/assets/wrappedlink.html
@@ -1,0 +1,29 @@
+<style>
+:root {
+  font-family: monospace;
+}
+
+body {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+div {
+  width: 10ch;
+  word-wrap: break-word;
+  border: 1px solid blue;
+  transform: rotate(33deg);
+  line-height: 8ch;
+  padding: 2ch;
+}
+
+a {
+  margin-left: 7ch;
+}
+</style>
+<div>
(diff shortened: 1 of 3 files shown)


test/assets/wrappedlink.html
```diff
@@ -0,0 +1,29 @@
...
+}
+
+body {
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
+
+div {
+  width: 10ch;
```
JoelEinbinder, line 13: Impressive CSS!

JoelEinbinder (approved):

PR #2824, opened 2018-06-29
fix: typing emoji

### LINKED ISSUE #1096: page.type has Emoji issues (UTF-8)
Two-folded issue. Some Emojis aren't typed correctly (others seem to work, randomly). And newlines are also ignored (so is `keyboard.press('Enter')`). The solution for newline treatment is to split the text by "\n" and then do line by line, with a `page.type("#target", String.fromCharCode(13));` at the end of it.

### Steps to reproduce

* Puppeteer version: 0.13.0-alpha
* Platform / OS version: OSX, also Debian 9

An empty HTML file, only has the following inside (it also doesn't work with the `<meta charset='utf-8'>` tag added, no difference):

`<textarea id="theTarget"></textarea>`

The code that causes the problem:

`var sample = Buffer.from("8J+RuSBUb2t5byBzdHJlZXQKSmFwYW4g8J+Hr/Cfh7Ug", 'base64').toString('utf8');
await page.type("#theTarget", sample);`

Even without the Buffer and decoding, if you just use the following, it won't work:
`await page.type("#theTarget", "👹 Tokyo street");`

👹 Tokyo street
Japan 🇯🇵

**What happens instead?**

 Tokyo streetJapan 

JoelEinbinder:
This changes sendCharacter to use document.execCommand instead of sending a `'char'` event from the protocol. This is more aligned with how input would come in from emoji keyboards, and removes the 3ish byte limit on characters that can be sent which prevented larger emoji from being rendered correctly.

Emoji will still fail to type correctly if typing them into an iframe that is in shadow dom.

fixes #1096

### DIFF
--- a/lib/Input.js
+++ b/lib/Input.js
@@ -148,13 +148,7 @@ class Keyboard {
    * @param {string} char
    */
   async sendCharacter(char) {
-    await this._client.send('Input.dispatchKeyEvent', {
-      type: 'char',
-      modifiers: this._modifiers,
-      text: char,
-      key: char,
-      unmodifiedText: char
-    });
+    await this._client.send('Input.insertText', {text: char});
   }
 
   /**
--- a/test/input.spec.js
+++ b/test/input.spec.js
@@ -472,6 +472,19 @@ module.exports.addTests = function({testRunner, expect, DeviceDescriptors}) {
       error = await page.keyboard.press('😊').catch(e => e);
       expect(error && error.message).toBe('Unknown key: "😊"');
     });
+    it('should type emoji', async({page, server}) => {
+      await page.goto(server.PREFIX + '/input/textarea.html');
+      await page.type('textarea', '👹 Tokyo street Japan 🇯🇵');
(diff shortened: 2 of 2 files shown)


lib/Input.js
```diff
@@ -148,13 +150,31 @@ class Keyboard {
...
   async sendCharacter(char) {
-    await this._client.send('Input.dispatchKeyEvent', {
-      type: 'char',
-      modifiers: this._modifiers,
-      text: char,
-      key: char,
-      unmodifiedText: char
-    });
+    // find the focused frame
+    let frame = this._page.mainFrame();
```
aslushnikov: shell we do this only for emoji characters to limit the scope of the change?

h2non:
Is this ready to be merged anytime soon?

aslushnikov:
@h2non there's a promising upstream patch that should address this: https://chromium-review.googlesource.com/c/chromium/src/+/1128398

aslushnikov:
@JoelEinbinder let's rebaseline this beauty and land it!

aslushnikov (approved):
awesome.

Issue #2671, opened 2018-06-04
Feature request: create isolated world


furstenheim:
I'd like to create an isolated context to execute some js in the same website I'm visiting without polluting the main context. It is possible right now, but it requires to access a lot of internal methods:

      const mainFrame = page.mainFrame()

      const isolatedWorldInfo = await page._client.send('Page.createIsolatedWorld', {frameId: mainFrame._id, worldName: 'new-isolated-world'})
      const executionContextId = isolatedWorldInfo.executionContextId
      const JsHandleFactory = page._frameManager.createJSHandle.bind(page._frameManager, executionContextId)

      const executionContext = new ExecutionContext(page._client, {id: executionContextId}, JsHandleFactory)
      await executionContext.evaluate(..)

It would be nice if puppeteer exposed this as a method. Something like: page.createNewIsolatedContext

If necessary I could write up the code and the tests

aslushnikov:
@furstenheim this is interesting. Can you please share your usecase?

furstenheim:
@aslushnikov I'm using it webscraper-headless https://github.com/geoblink/web-scraper-chrome-extension
Basically, it was an existing Chrome extension used for scraping. In the normal browser it will execute a bunch of code accessing the DOM to scrape the website. In the browser this is safe because there is a different context for the extension. For example, there cannot be collisions with jquery since it won't be globally defined.

To scrape on the server what I'm doing is opening a new context as in the snippet and injecting the code of the extension there.

https://github.com/geoblink/web-scraper-chrome-extension/blob/headless-mode/extension/scripts/ChromeHeadlessBrowser.js#L47

aslushnikov:
@furstenheim this looks very reasonable; once #2812 lands we'll add this one as well.

aslushnikov:
> If necessary I could write up the code and the tests

I'd be happy to review a PR! Please make sure to add reference to https://developer.chrome.com/extensions/content_scripts when explaining isolated worlds in documentation.

aslushnikov:
(cross-posting from https://github.com/GoogleChrome/puppeteer/pull/2829#issuecomment-418479798)

I've played with this for a while. A few observations:
- execution contexts per se are not very handy. Once I have an isolated world, I want to be able to run querySelector and all the ElementHandle goodness. THis means we need to have an `IsolatedWorld` class in our API.
- once I have querySelector and everything, I want `isolatedWorld.waitForSelector` to be working just fine.
[...]

Issue #609, opened 2017-08-30
page.waitForSelector doesn't work on baidu.com


aslushnikov:
The following script fails with baidu.com:
```js
const puppeteer = require('puppeteer');
(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  await page.goto('https://baidu.com');
  await page.waitForSelector('div');
  browser.close();
})();
```

The reason is that Baidu nullifies MutationObserver that is required for `waitForSelector`.

aslushnikov:
As reported in #1668, `page.click(selector)` doesn't work if page redefines `window.Node`

nylen:
Suggest renaming this issue to something more general, like: "Puppeteer methods break if pages redefine built-in global objects"

Is this sort of thing even possible to work around at the library level?  I did this in my code, which seems to work ok:

```js
await this.page.evaluate( () => {
    if ( ! window.Node ) {
        window.Node = {};
    }
    if ( ! Node.ELEMENT_NODE ) {
        Node.ELEMENT_NODE = 1;
    }
} );
```

klichukb:
Any updates on this issue?

PR #2829, opened 2018-06-30
feat: createNewIsolatedContext

furstenheim:
Implements feature described in #2671

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -83,6 +83,7 @@
   * [page.content()](#pagecontent)
   * [page.cookies(...urls)](#pagecookiesurls)
   * [page.coverage](#pagecoverage)
+  * [page.createNewIsolatedContext(contextIdentifier)](#pagecreatenewisolatedcontextcontextidentifier)
   * [page.deleteCookie(...cookies)](#pagedeletecookiecookies)
   * [page.emulate(options)](#pageemulateoptions)
   * [page.emulateMedia(mediaType)](#pageemulatemediamediatype)
@@ -921,6 +922,22 @@ If URLs are specified, only cookies for those URLs are returned.
 #### page.coverage
 
 - returns: <[Coverage]>
+
+
+#### page.createNewIsolatedContext(contextIdentifier)
+- `contextIdentifier`<[string]> Identifier to assign to the new isolated context
+- returns: <[Promise]<[ExecutionContext]>> Promise which resolves to the new isolatedContext
+
+Creates a new isolated context for the page. Context are able to access DOM but have an independent [JavaScript environment](https://developer.chrome.com/extensions/content_scripts).
+This allows executing code without conflicts with the current website.
+
+```js
+
+const context = await page.createNewIsolatedContext('myIsolatedWorld')
+await context.evaluate(function () {
+    // DOM is shared with page but JS Environment is not
+})
+```
 
 #### page.deleteCookie(...cookies)
 - `...cookies` <...[Object]>
--- a/docs/api.md
+++ b/docs/api.md
@@ -926,17 +926,16 @@ If URLs are specified, only cookies for those URLs are returned.
 
 #### page.createNewIsolatedContext(contextIdentifier)
 - `contextIdentifier`<[string]> Identifier to assign to the new isolated context
-- returns: <[Promise]<[ExecutionContext]>> Promise which resolves to the new isolatedContext
+- returns: <[Promise]<[ExecutionContext]>> Promise which resolves to the new isolated context.
 
-Creates a new isolated context for the page. Context are able to access DOM but have an independent [JavaScript environment](https://developer.chrome.com/extensions/content_scripts).
+Creates a new isolated context for the page. Contexts are able to access DOM but have an independent [JavaScript environment](https://developer.chrome.com/extensions/content_scripts).
 This allows executing code without conflicts with the current website.
 
 ```js
-
-const context = await page.createNewIsolatedContext('myIsolatedWorld')
+const context = await page.createNewIsolatedContext('myIsolatedWorld');
 await context.evaluate(function () {
     // DOM is shared with page but JS Environment is not
-})
+});
 ```
 
 #### page.deleteCookie(...cookies)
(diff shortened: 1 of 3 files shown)


furstenheim:
I've signed the CLA

docs/api.md
```diff
@@ -922,6 +923,22 @@ If URLs are specified, only cookies for those URLs are returned.
 
 - returns: <[Coverage]>
 
+
+#### page.createNewIsolatedContext(contextIdentifier)
+- `contextIdentifier`<[string]> Identifier to assign to the new isolated context
+- returns: <[Promise]<[ExecutionContext]>> Promise which resolves to the new isolatedContext
```
vsemozhetbyt: `` isolatedContext `` -> `` `isolatedContext` `` or `` isolated context ``?

docs/api.md
```diff
@@ -922,6 +923,22 @@ If URLs are specified, only cookies for those URLs are returned.
...
+
+Creates a new isolated context for the page. Context are able to access DOM but have an independent [JavaScript environment](https://developer.chrome.com/extensions/content_scripts).
```
vsemozhetbyt: `Context are` -> `Contexts are` or `Context is`?

docs/api.md
```diff
@@ -922,6 +923,22 @@ If URLs are specified, only cookies for those URLs are returned.
...
+This allows executing code without conflicts with the current website.
+
+```js
+
```
vsemozhetbyt: Unneeded empty line?

docs/api.md
```diff
@@ -922,6 +923,22 @@ If URLs are specified, only cookies for those URLs are returned.
...
+const context = await page.createNewIsolatedContext('myIsolatedWorld')
```
vsemozhetbyt: Missing semicolon.

docs/api.md
```diff
@@ -922,6 +923,22 @@ If URLs are specified, only cookies for those URLs are returned.
...
+await context.evaluate(function () {
+    // DOM is shared with page but JS Environment is not
+})
```
vsemozhetbyt: Missing semicolon.

vsemozhetbyt (commented):
Some nits)

furstenheim:
@vsemozhetbyt all comments addressed

lib/Page.js
```diff
@@ -269,6 +270,26 @@ class Page extends EventEmitter {
     return context.evaluateHandle(pageFunction, ...args);
   }
 
+  /**
+   * @param {string} contextIdentifier
+   * @returns {!Promise.<!Puppeteer.ExecutionContext>}
+   */
+  async createNewIsolatedContext(contextIdentifier) {
```
JoelEinbinder, line 277: Does `contextIdentifier` have any significance for puppeteer scripts? It looks like this parameter is unnecessary.

test/page.spec.js
```diff
@@ -264,6 +264,46 @@ module.exports.addTests = function({testRunner, expect, puppeteer, DeviceDescrip
     });
   });
 
+  describe('Page.createNewIsolatedContext', function() {
+    it('should work', async({page, server}) => {
+      await page.evaluate(function() {
+        window.constantVariable = new Proxy({}, {
```
JoelEinbinder, line 270: I think this test can be much simpler. Instead of messing around with proxies, just something like:
```js
await page.evaluate(() => window.dontFindThis = true);
expect(await isolatedContext.evaluate(() => window.dontFindThis)).toBe(undefined);
```

lib/Page.js
```diff
@@ -269,6 +270,26 @@ class Page extends EventEmitter {
...
+    const mainFrame = this.mainFrame();
+    const isolatedWorldInfo = await this._client.send(
+        'Page.createIsolatedWorld',
+        {frameId: mainFrame._id, worldName: contextIdentifier}
```
JoelEinbinder, line 281: This looks like it should be a frame-level method. Then the page-level method would just do `this.mainFrame().createNewIsolatedContext()`

JoelEinbinder:
Thanks for the pull request!

lib/Page.js
aslushnikov: @JoelEinbinder: it'll play nice with your "frame.executionContexts()" pull request: #2812 

@furstenheim: nit - I'd name this method `createIsolatedWorld(name)`.

JoelEinbinder: The name should at least be optional in that case.

furstenheim: @JoelEinbinder should name be created as a random string then?

lib/Page.js
furstenheim: Happy to do it, but I'm not sure how. To create a context I need a JSHandleFactory which I can only get from a FrameManager. That I can access from Page but not from Frame. Any idea how to deal with that?

I refer to:

`
    const JsHandleFactory = this._frameManager.createJSHandle.bind(this._frameManager, executionContextId);
`

Which is a required parameter to create a new execution context.

lib/Page.js
aslushnikov: @furstenheim I think name is not necessary for the protocol method.

lib/Page.js
aslushnikov: @furstenheim you'll need to await the `ExecutionContext` with this id; I'd do this after the  #2812 lands.

aslushnikov:
I've played with this for a while. A few observations:
- execution contexts per se are not very handy. Once I have an isolated world, I want to be able to run querySelector and all the ElementHandle goodness.
- once I have querySelector and everything, I want `isolatedWorld.waitForSelector` to be working just fine.
- the `isolatedWorld.waitForSelector` should work just fine if my frame navigates away. It means that worlds should be recreated after navigation.

In order to have this, we need to support "persistent" isolated worlds on the protocol level. Such isolated worlds:
- are re-created once the frame navigates somewhere
- support `exposeFunction` and `evaluateOnNewDocument` methods 

This will make a complete story around isolated worlds. With this level of isolation, things like #609 will be automatically addressed from the user-land.

Closing this; once we have upstream work done, we can introduce the isolated worlds to our API.

PR #2838, opened 2018-07-03
fix: wait for the chromeProcess to exit, not close

JoelEinbinder:
When closing Chromium with the `Browser.close` protocol command, the 'exit' event can sometimes fire 15 seconds before the close event. This is causing our user-data-dir tests to timeout on the bots.


JoelEinbinder:
@ak239 does this look right?

alexkozy (approved):
It looks better to me.

aslushnikov (approved):
nicee!

PR #2866, opened 2018-07-10
feat(timeout): closing browser terminates page.goto

cfellin1:
https://github.com/GoogleChrome/puppeteer/issues/2721

The changes to `test/test.js` are my attempt to deal with the fact that the unit test closes the browser mid-run. I recognize that they are probably too extensive and am open to other options.


aslushnikov:
Closing in favor of #3245.

PR #2881, opened 2018-07-12
feat: introduce basic Error classes

### LINKED ISSUE #1694: Conditional catch for errors
<sup>Issue template didn't really apply.</sup>

I can't find anything in the api docs page about error handling. The main type of error that's encountered is a timeout. The issue is that there's no documented way to distinguish the type of error received in the `catch` after awaiting a promise that rejects.

The issue is that good code (most of the time) needs to do conditional catches to avoid silencing unexpected errors. There doesn't seem to be a good, stable way to do this currently.

I tried inspecting some of the errors. For these tests, the output is logging `error` and then logging `{...error}`.

```js
await page.click('doesntexist');

// Result
{ AssertionError [ERR_ASSERTION]: No node found for selector: doesntexist
    at Page.click (/private/tmp/temp-puppet-7695/node_modules/puppeteer/lib/Page.js:770:13)
    at <anonymous>
  generatedMessage: false,
  name: 'AssertionError [ERR_ASSERTION]',
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '==' }
{ generatedMessage: false,
  name: 'AssertionError [ERR_ASSERTION]',
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '==' }
```

Gives a generic `AssertionError` with no information about the cause, except if you want to parse the error message, which seems very unstable as it can change without much notice.

The next issue is `waitFor`:

```js
await page.waitFor('doesntexist');

// Result
Error: waiting failed: timeout 30000ms exceeded
    at Timeout.WaitTask._timeoutTimer.setTimeout (/private/tmp/temp-puppet-7695/node_modules/puppeteer/lib/FrameManager.js:593:58)
    at ontimeout (timers.js:475:11)
    at tryOnTimeout (timers.js:310:5)
    at Timer.listOnTimeout (timers.js:270:5)
{}
```

Again, the only way I see to extract information is the error message.

---

There are a few potential solutions here:

1. document the error messages and give an example regex to check them. This requires no code changes, but makes it clear that the error messages changing would be a breaking change. 
2. don't change the errors, but give official utility functions like `puppeteer.isTimeoutError(error)`. Very little code changes, but the api will be stable across versions. Better than 1.
3. make each error either an Error subclass, or give it properties e.g. `type: 'PuppeteerTimeout', selector: 'doesntexist'`. This is about tied with 2, since it gives a readable way to check the error type.

Thoughts? Sorry if this has come up before, but I couldn't find anything relevant.

aslushnikov:
This patch introduces 3 error classes that are used throughout
the code to report errors:
- `InvalidNodeError` - thrown whenever Puppeteer didn't like the node it
  was asked to operate upon
- `ProtocolError` - thrown whenever DevTools protocol returns an error
- `TimeoutError` - thrown whenever an operation was aborted due to
  timeout.

Fixes #1694

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -16,6 +16,7 @@
   * [puppeteer.connect(options)](#puppeteerconnectoptions)
   * [puppeteer.createBrowserFetcher([options])](#puppeteercreatebrowserfetcheroptions)
   * [puppeteer.defaultArgs()](#puppeteerdefaultargs)
+  * [puppeteer.errors](#puppeteererrors)
   * [puppeteer.executablePath()](#puppeteerexecutablepath)
   * [puppeteer.launch([options])](#puppeteerlaunchoptions)
 - [class: BrowserFetcher](#class-browserfetcher)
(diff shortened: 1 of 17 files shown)


eliucs:
Any update on this issue?

aslushnikov:
Closed in favor of #3056

PR #2887, opened 2018-07-13
feat(chromium): roll Chromium to r574897

JoelEinbinder:
This contains https://chromium-review.googlesource.com/c/chromium/src/+/1135939 which fixes tracing over the remote debugging pipe.

### DIFF
--- a/lib/Pipe.js
+++ b/lib/Pipe.js
@@ -35,14 +35,14 @@ class Pipe extends EventEmitter {
    */
   send(message) {
     this._pipeWrite.write(message);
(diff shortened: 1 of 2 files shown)


aslushnikov (approved):

aslushnikov:
@JoelEinbinder Shall we add a test for this?

JoelEinbinder:
The right thing to do is to run all of our tests with pipe=true and pipe=false. But that's quite a pain. So I think we can just have faith that Chromium can parse JSON properly, even if that faith may be misplaced.

PR #2890, opened 2018-07-15
docs(api.md): CamelCase to ElementHandle class in asElement method

MaxCilauro:
Fixes typo in ElementHandle.asElement method, the return value was written without upper camelcase and since it returns a class, it should.


MaxCilauro:
I've signed it

aslushnikov (approved):

PR #2893, opened 2018-07-16
docs(api.md): expand on testing Chrome Extensions

### LINKED ISSUE #2823: docs: Puppeteeer with Chrome Extensions
Puppeteer is organically getting better and better support for extensions.

We should expand in our docs on some common tasks to work with extensions and the limitations we have:
- use targets with `background_page` type to connect to background pages
- (once #2812 lands) explain how to access content scripts for a particular frame
- note that Chrome Headless doesn't support extensions; headful mode should be used.

pomerantsev:
Resolves #2823 

Took a stab at expanding docs on working with extensions.

@aslushnikov let me know if this aligns with what you had in mind.


aslushnikov (approved):
Thanks! This a very good start!

PR #2904, opened 2018-07-18
Update README.md typo

killa-kyle:
Under the create a PDF example the 'waitUntil' option gives the error 

AssertionError [ERR_ASSERTION]: Unknown value for options.waitUntil: networkidle2

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ const puppeteer = require('puppeteer');
 (async () => {
   const browser = await puppeteer.launch();
   const page = await browser.newPage();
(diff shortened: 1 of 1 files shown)


killa-kyle:
I signed the CLA

aslushnikov:
@killa-kyle I think you're using old puppeteer version. We used to have `networkidle` and `networkalmostidle` options, but they got renamed into `networkidle0` and `networkidle2`. Puppeteer v1.6.0 doesn't support `newtorkidle` argument.

PR #2907, opened 2018-07-18
fix(page): page.click should work with disabled javascript

### LINKED ISSUE #2898: not working "page.click" if "setJavaScriptEnabled" is "false"
* Puppeteer version: 1.6
* Platform / OS version: windows 10
* URLs (if applicable): all !
* Node.js version: v8.11.3

Hi guys!
I can not click when I set "JavascriptEnabled" to "false"!
This problem is solved only by changing "false" to "true"!
Is this a problem of the "puppeteer" is?

**What happens instead?**
When "click" does not work, no output will not happen and is not even an error

**Edit**:
This problem exists only in v1.6.0
I tested in v1.5.0 and worked fine

aslushnikov:
Unfortunately, disabling javascript in page prevents any microtasks
to be executed even from puppeteer-originating javascript. As a
result, the IntersectionObserver hack we use to conditionally
scroll into view doesn't work.

To workaround this, we start always scrolling before clicking if
page's javascript is disabled.

Fixes #2898

### DIFF
--- a/lib/ElementHandle.js
+++ b/lib/ElementHandle.js
@@ -55,11 +55,16 @@ class ElementHandle extends JSHandle {
   }
 
   async _scrollIntoViewIfNeeded() {
-    const error = await this.executionContext().evaluate(async element => {
+    const error = await this.executionContext().evaluate(async(element, pageJavascriptEnabled) => {
       if (!element.isConnected)
         return 'Node is detached from document';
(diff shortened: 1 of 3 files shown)


JoelEinbinder (approved):
This works. JavaScript might be disabled via multiclient or reconnecting to a page. I don’t think those are important cases though.

Issue #2854, opened 2018-07-08
Feature Request: clearer UX for waitForSelector inverse.


andyfleming:
I'm familiar with the `await page.waitForSelector('.example', {hidden: true});` that came about, but the syntax is not that intuitive and the error message can be misleading.

```
Error: waiting for selector ".example" failed: timeout 30000ms exceeded
```

It would be good to have an explicit API and error messaging.

### page.waitForSelectorRemoval

```js
await page.waitForSelectorRemoval('.example')
```

```
Error: waiting for selector ".example" to be removed failed: timeout 30000ms exceeded
```

### page.waitForSelectorToBeHidden

```js
await page.waitForSelectorToBeHidden('.example')
```

```
Error: waiting for selector ".example" to be hidden failed: timeout 30000ms exceeded
```
---------

## Additional Details

It seems like there has been enough activity around this to warrant the consideration of these additions.

#### Related Issues:
#549 — Add inverse of waitForSelector visible: true (waitForNotSelector?)
#705 — Feature request for page.waitForSelector() reverse
#967 — feat(page): waitForSelector hidden option

#### Proposed labels:
<kbd>feature</kbd> <kbd>good first issue</kbd>

aslushnikov:
> It seems like there has been enough activity around this to warrant the consideration of these additions.

@andyfleming The 2 issues you're referring to were fixed by the PR #967 that you also refer. They are all about the 'hidden' option in the `waitForSelector` method - I don't think this has anything to do with this proposal.

> It would be good to have an explicit API and error messaging.

I don't think these API changes make much difference and have any justification, so I'd keep things as they are today.

Thanks for writing us!

andyfleming:
Thanks for taking a look at this @aslushnikov!

I've opened a simpler PR that improves the error message without making any changes to the actual API. Take a look and let me know what you think.

https://github.com/GoogleChrome/puppeteer/pull/2911

PR #2911, opened 2018-07-19
feat(frame-manager): improve wait for to be hidden error message

andyfleming:
Improves error message for when a `waitForSelector` call uses the hidden: true option.

Rather than just saying

```
waiting for selector "div" failed: timeout
```

it will say

```
waiting for selector "div" to be hidden failed: timeout
```

(when appropriate).

Addresses #2854, replaces PR #2909


andyfleming:
I signed it!

JoelEinbinder (approved):
Looks good. Thanks!

Issue #1963, opened 2018-02-04
Missing options preferCSSPageSize and ignoreInvalidPageRanges in Page.pdf()


N-X-N:
Chrome's DevTools version 1.3 (targets Chrome 64) accepts these options for [Page.printToPDF](https://chromedevtools.github.io/devtools-protocol/1-3/Page#method-printToPDF) :
- ignoreInvalidPageRanges
- preferCSSPageSize

I found no corresponding options in Page.pdf().

CaelanStewart:
Are there any plans to add this option to Puppeteer?

I'm not familiar with Puppeteer's internals, but could this option be of help to [this issue](https://github.com/GoogleChrome/puppeteer/issues/2278)?

PR #2928, opened 2018-07-20
feat(page): add 'preferCSSPageSize' to page.pdf options

fchasen:
Adds an option to set the `preferCSSPageSize` boolean when sending `Page.printToPDF`

Per #1963

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1277,6 +1277,7 @@ Page is guaranteed to have a main frame which persists during navigations.
     - `right` <[string]> Right margin, accepts values labeled with units.
     - `bottom` <[string]> Bottom margin, accepts values labeled with units.
     - `left` <[string]> Left margin, accepts values labeled with units.
+  - `preferCSSPageSize` <[boolean]> Prefer page size as defined by css. Defaults to `false`.
 - returns: <[Promise]<[Buffer]>> Promise which resolves with PDF buffer.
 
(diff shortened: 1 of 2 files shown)


aslushnikov:
@fchasen  What does this option do? Is this the "fit to page" checkbox I see in the print dialog?

fchasen:
If `false` it will scale to fit like that checkbox, but if it is `true` it will use any page sizing declared in CSS (`@page { size: A5 }`) over what is declared in the Puppeteer width and height.

Per https://chromedevtools.github.io/devtools-protocol/1-3/Page#method-printToPDF 

> Whether or not to prefer page size as defined by css. Defaults to false, in which case the content will be scaled to fit the paper size.

With this we are able to set a default width and height but let the content override it as needed.

I'm also hopeful that it might help with the rounding issues in the output size of the PDF a bit by letting us define the unit conversions to inches in the document.

aslushnikov:
@fchasen thanks for the explanation. Can you please add this to the documentation?

aslushnikov (approved):
thanks!

insist-git:
image
table page break ?

insist-git:
How to solve it? How to solve the problem of table Pagination？

insist-git:
image

How to solve it? How to solve the problem of table Pagination？

54huige:
> image
> 
> How to solve it? How to solve the problem of table Pagination？

估计没好办法了

Issue #2925, opened 2018-07-20
puppeteer@next is not autoupdated


aslushnikov:
[Recent eslint hack](https://eslint.org/blog/2018/07/postmortem-for-malicious-package-publishes) caused all NPM keys to be invalidated. This broke our puppeteer@next self-deployment through travis-ci.

aslushnikov:
Ok, this didn't work.

aslushnikov:
no luck!

aslushnikov:
It works!

ntzm:
I think this is still a little broken, the latest commit d54c7edeae44800dabb87e52ae96180a16a149a1 isn't on next

aslushnikov:
@ntzm crazy. Indeed.

PR #2937, opened 2018-07-23
chore(ci): use specific npm version

aslushnikov:
References #2925.


JoelEinbinder:
This didn't turn out to be needed. Let's just upgrade our node version and leave npm alone.

PR #2938, opened 2018-07-23
chore(ci): move all CI from Node7 to Node8

aslushnikov:
Node 8 is the current LTS. We should use it by default on all
our CIs.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,14 +1,14 @@
 environment:
   matrix:
     - nodejs_version: "6.12.3"
-    - nodejs_version: "7.10.1"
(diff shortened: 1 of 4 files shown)


JoelEinbinder:
Let's do node 10 and node 6.

aslushnikov:
> Let's do node 10 and node 6.

With pleasure. Closing in favor of https://github.com/GoogleChrome/puppeteer/pull/2941

aslushnikov:
Resurrecting this since node 10 is not quite ready yet

JoelEinbinder (approved):

PR #2942, opened 2018-07-24
fix(launcher): always add `about:blank` to default arguments

aslushnikov:
Chrome Headless used to open about:blank by default; however, this
was recently changed.

We should open starting page no matter what to keep the environment
predictable.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -97,7 +97,7 @@ class Launcher {
       if (os.platform() === 'win32')
         chromeArguments.push('--disable-gpu');
     }
-    if (!options.ignoreDefaultArgs && Array.isArray(options.args) && options.args.every(arg => arg.startsWith('-')))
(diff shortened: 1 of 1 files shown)


lib/Launcher.js
```diff
@@ -97,7 +97,7 @@ class Launcher {
       if (os.platform() === 'win32')
         chromeArguments.push('--disable-gpu');
     }
-    if (!options.ignoreDefaultArgs && Array.isArray(options.args) && options.args.every(arg => arg.startsWith('-')))
```
JoelEinbinder, line 100: This is a slight functionality change. There is now no way to get the newtab page in headful chrome as the startup tab. That is probably fine though.

JoelEinbinder (approved):

lib/Launcher.js
aslushnikov: >  There is now no way to get the newtab page in headful chrome as the startup tab. 

I think you can navigate to `chrome://newtab`

PR #2944, opened 2018-07-24
feat(launcher): use Chromium revision in npm config in launcher

kwkbtr:
Since Chromium revision specified in npm config was not referred from launcher, it was not possible to launch (or getting the executable path by `executablePath()`) the downloaded Chromium.


aslushnikov:
Closing in favor of #3190: it does the same, but aligns with out vision regarding puppeteer-core and puppeteer.

PR #2950, opened 2018-07-24
feat: allow options to be passed into puppeteer.defaultArgs

JoelEinbinder:
Default args didn't take launch options, so were missing some args like `--headless` and `--mute-audio` when in headless mode.

Replaces the experimental option `appMode` with an experimental option `setsDefaultViewport`.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -332,19 +332,7 @@ function waitForWSEndpoint(chromeProcess, timeout) {
 
 /**
  * @typedef {Object} LaunchOptions
-<<<<<<< HEAD
- * @property {boolean=} headless
  * @property {string=} executablePath
- * @property {!Array<string>=} args
-=======
-<<<<<<< HEAD
- * @property {boolean=} ignoreHTTPSErrors
- * @property {string=} executablePath
- * @property {number=} slowMo
-=======
- * @property {string=} executablePath
->>>>>>> fcaf8be10a8b3d7d218f55db3426849f0a9ce0d2
->>>>>>> 1
  * @property {boolean=} ignoreDefaultArgs
  * @property {boolean=} handleSIGINT
  * @property {boolean=} handleSIGTERM
@@ -353,12 +341,6 @@ function waitForWSEndpoint(chromeProcess, timeout) {
  * @property {boolean=} dumpio
  * @property {!Object<string, string | undefined>=} env
  * @property {boolean=} pipe
-<<<<<<< HEAD
-=======
-<<<<<<< HEAD
- * @property {boolean=} setDefaultViewport
-=======
->>>>>>> 1
  */
 
 /**
@@ -366,10 +348,6 @@ function waitForWSEndpoint(chromeProcess, timeout) {
  * @property {boolean=} ignoreHTTPSErrors
  * @property {(?Puppeteer.Viewport)=} defaultViewport
  * @property {number=} slowMo
-<<<<<<< HEAD
-=======
->>>>>>> fcaf8be10a8b3d7d218f55db3426849f0a9ce0d2
->>>>>>> 1
  */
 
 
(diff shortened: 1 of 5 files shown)


lib/Launcher.js
```diff
@@ -164,7 +131,7 @@ class Launcher {
         connection = Connection.createForPipe(/** @type {!NodeJS.WritableStream} */(chromeProcess.stdio[3]), /** @type {!NodeJS.ReadableStream} */ (chromeProcess.stdio[4]), connectionDelay);
       }
       const ignoreHTTPSErrors = !!options.ignoreHTTPSErrors;
-      const setDefaultViewport = !options.appMode;
+      const {setDefaultViewport = true} = options;
```
aslushnikov: People will appreciate this; let's release this to public api separately.

lib/Launcher.js
```diff
@@ -231,10 +198,32 @@ class Launcher {
   }
 
   /**
+   * @param {!ChromeArgOptions=} options
    * @return {!Array<string>}
    */
-  static defaultArgs() {
-    return DEFAULT_ARGS.concat(AUTOMATION_ARGS);
+  static defaultArgs(options = {}) {
+    const userArgs = options.args || [];
```
aslushnikov: we had `Array.isArray` check before for user args

aslushnikov (approved):

lib/Launcher.js
JoelEinbinder: If someone passes a truthy but non array value, we should throw an error instead of silently moving on.

aslushnikov: I'm all in for an assert

test/puppeteer.spec.js
```diff
@@ -193,13 +194,18 @@ module.exports.addTests = function({testRunner, expect, PROJECT_ROOT, defaultBro
         await browser.close();
       });
       it('should work with no default arguments', async() => {
+        // We need to set a user-data-dir otherwise this test will fail on the windows bots
+        const userDataDir = await mkdtempAsync(TMP_FOLDER);
```
aslushnikov: why this change?

lib/Launcher.js
```diff
@@ -48,62 +48,50 @@ const DEFAULT_ARGS = [
...
-    if (!options.args || !options.args.some(arg => arg.startsWith('--user-data-dir'))) {
-      if (!options.userDataDir)
-        temporaryUserDataDir = await mkdtempAsync(CHROME_PROFILE_PATH);
+    const chromeArguments = !ignoreDefaultArgs ? this.defaultArgs(options) : args;
+    let temporaryUserDataDir = null;
 
-      chromeArguments.push(`--user-data-dir=${options.userDataDir || temporaryUserDataDir}`);
+    if (!chromeArguments.some(argument => argument.startsWith('--remote-debugging-')))
+      chromeArguments.push(pipe ? '--remote-debugging-pipe' : '--remote-debugging-port=0');
+    if (!ignoreDefaultArgs && !chromeArguments.some(arg => arg.startsWith('--user-data-dir'))) {
```
aslushnikov: I don't think we want `ignoreDefaultArgs` to disable temp userDataDir management. We didn't do this before.

test/puppeteer.spec.js
JoelEinbinder: Removed.

lib/Launcher.js
JoelEinbinder: Fair.

lib/Launcher.js
```diff
@@ -316,28 +322,54 @@ function waitForWSEndpoint(chromeProcess, timeout) {
...
+ * @typedef {Object} ChromeArgOptions
+ * @property {boolean=} headless
+ * @property {Array<string>=} args
+ * @property {string=} userDataDir
+ * @property {boolean=} devtools
+ */
+
 /**
  * @typedef {Object} LaunchOptions
+<<<<<<< HEAD
```
ebidel: uh, what happened here?

JoelEinbinder: earthquake

aslushnikov (approved):

PR #2954, opened 2018-07-25
fix(networkmanager): Remove header 'cookie' from request hash

DonoA:
This fixes an issue in which a response may not be recorded or trigger an event because of differing cookie headers from the request.

I didn't see anywhere to add testing for this functionality, however if you would like tests added, just point me in the right direction.


DonoA:
I signed it!

aslushnikov (approved):
Thanks. I'll add a test for this in a follow-up.

PR #2956, opened 2018-07-26
feat(page): make waitForNavigation work for subframes

### LINKED ISSUE #2918: Support waitForNavigation for frames
### Steps to reproduce

* Puppeteer version: 1.6.0
* Platform / OS version: Linux henrique-Latitude-5480 4.15.0-23-generic #25-Ubuntu SMP Wed May 23 18:02:16 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
* URLs (if applicable):  http://home.ubalt.edu/abento/frames/myframe.html
* Node.js version: v8.11.3

```js
(async () => {

const puppeteer = await require('puppeteer');

const browser = await puppeteer.launch({headless:false,/* slowMo: 50*/});
const incognitoContext = await browser.createIncognitoBrowserContext();
const page = await incognitoContext.newPage();
page.on('dialog', async dialog => {
    await console.log('MESSAGE "%s": "%s"',dialog.type(),dialog.message());
    await dialog.dismiss();
});
await page.setCacheEnabled(false);

await page.goto('http://home.ubalt.edu/abento/frames/myframe.html', {waitUntil: ['load','domcontentloaded','networkidle0']});

const leftFrame = await page.frames().filter( el => el.name() == 'left')[0]
const rightFrame = await page.frames().filter( el => el.name() == 'right')[0]
await console.log('URLS: MAIN FRAME: %s ; RIGHT FRAME: %s ; LEFT FRAME: %s ', page.url(), rightFrame.url(), leftFrame.url());
const navigationPromise = page.waitForNavigation();
await leftFrame.click('a[href="option2.html"]');
await console.log('URLS: MAIN FRAME: %s ; RIGHT FRAME: %s ; LEFT FRAME: %s ', page.url(), rightFrame.url(), leftFrame.url());
await page.waitFor(5000);
await console.log('URLS: MAIN FRAME: %s ; RIGHT FRAME: %s ; LEFT FRAME: %s ', page.url(), rightFrame.url(), leftFrame.url()); // DISPLAY OTHER URL FOR rightFrame
await navigationPromise;
await console.log('URLS: MAIN FRAME: %s ; RIGHT FRAME: %s ; LEFT FRAME: %s ', page.url(), rightFrame.url(), leftFrame.url()); // NOT DISPLAYED, navigationPromise ERROR
await browser.close();
    
})();
```

I was expecting navigationPromise to be resolved, but it doesn't.

**What happens instead?**

navigationPromise isn't resolved. Is it the expected behavior? Is it possible to add a frame.waitForNavigation() function?

Screenshot:

image

Thanks

PS: I've also tried with the code below, using waitForNavigation as suggested in frame.click() documentation:

```js
"use strict";

(async () => {

const puppeteer = await require('puppeteer');

const browser = await puppeteer.launch({headless:false,/* slowMo: 50*/});
const incognitoContext = await browser.createIncognitoBrowserContext();
const page = await incognitoContext.newPage();
page.on('dialog', async dialog => {
    await console.log('MESSAGE "%s": "%s"',dialog.type(),dialog.message());
    await dialog.dismiss();
});
await page.setCacheEnabled(false);

await page.goto('http://home.ubalt.edu/abento/frames/myframe.html', {waitUntil: ['load','domcontentloaded','networkidle0']});

const leftFrame = await page.frames().filter( el => el.name() == 'left')[0]
const rightFrame = await page.frames().filter( el => el.name() == 'right')[0]

await Promise.all([
    page.waitForNavigation(),
    leftFrame.click('a[href="option2.html"]')
]);

})();
```

pomerantsev:
Resolves #2918


JoelEinbinder:
Code looks good. As far as the api is concerned, I was expecting `frame.waitForNavigation()` instead of `page.waitForNavigation({frame})`. Then `page.waitForNavigation` would be a shortcut for `page.mainFrame().waitForNavigation()` similar to other methods on `page`. Wdyt?

pomerantsev:
Agreed. Please see the above update. If it looks good API-wise, I can add the tests.

pomerantsev:
Actually, a couple seemingly unrelated tests are now broken. Working on getting them fixed.

pomerantsev:
Turned out that with `waitForNavigation` inside `Frame`, `Frame` had to have a reference to page's `NetworkManager`, not create a new one. 
In order to prevent circular dependencies, I changed `FrameManager` and `NetworkManager` so they have a reference to the instance of `Page` that created them.
All existing tests are fixed now.

docs/api.md
```diff
@@ -1158,6 +1161,9 @@ If there's no element matching `selector`, the method throws an error.
 
 Shortcut for [page.mainFrame().focus(selector)](#framefocusselector).
 
+#### page.frameManager()
```
JoelEinbinder: Let's not expose the frameManager and networkManager in the public API.

pomerantsev: If these methods are not exposed, we have to call private methods from outside of the class, which breaks the encapsulation principle of OOP. I'm fine with it though - you're more familiar with the codebase, so you probably know how good the rest of the code fares in this respect.
I decided to ditch getter methods altogether since we're accessing private stuff anyway, and just stick to the properties.

pomerantsev:
I moved tests from Page to Frame and slightly modified them. I believe that the feature is complete.

lib/FrameManager.js
```diff
@@ -719,6 +722,20 @@ class Frame {
     return new WaitTask(this, pageFunction, 'function', polling, timeout, ...args).promise;
   }
 
+  async waitForNavigation(options = {}) {
+    const timeout = typeof options.timeout === 'number' ? options.timeout : 30000;
+    const NavigatorWatcher = require('./NavigatorWatcher');
```
JoelEinbinder: Can you move this require up top with the others? Keeps things organized.

lib/FrameManager.js
```diff
@@ -719,6 +722,20 @@ class Frame {
     return new WaitTask(this, pageFunction, 'function', polling, timeout, ...args).promise;
   }
 
+  async waitForNavigation(options = {}) {
+    const timeout = typeof options.timeout === 'number' ? options.timeout : 30000;
```
JoelEinbinder: This should use the default timeout from the page. You can just do ‘this._page._defaultTimeout.

docs/api.md
```diff
@@ -2304,6 +2305,27 @@ const selector = '.foo';
 await page.waitForFunction(selector => !!document.querySelector(selector), {}, selector);
 ```
 
+#### frame.waitForNavigation(options)
```
JoelEinbinder, line 2310: Can you add a message to page.waitForNavigation saying that it’s a shortcut to this method?

lib/NetworkManager.js
```diff
@@ -20,12 +20,12 @@ const Multimap = require('./Multimap');
 class NetworkManager extends EventEmitter {
   /**
    * @param {!Puppeteer.CDPSession} client
-   * @param {!Puppeteer.FrameManager} frameManager
+   * @param {!Puppeteer.Page} page
    */
-  constructor(client, frameManager) {
+  constructor(client, page) {
```
JoelEinbinder, line 25: I don’t think you need to pass the page into the network manager anymore

JoelEinbinder (commented):
Just a few small changes left to make. Thanks!

lib/FrameManager.js
pomerantsev: This will require some refactoring, because inside `NavigatorWatcher`, we require `FrameManager`. So one of them is going to be undefined inside the other if we move this require to the top.
I'm open to suggestions.

lib/NetworkManager.js
pomerantsev: With the refactoring that I've done, we do.
Both FrameManager and NetworkManager are just wrappers for some Page functionality. So instead of passing them into each other and resorting to ugly hacks to ensure none of them is `undefined`, I'm simply passing a reference to the Page that generated them both.
And yes, FrameManager is still used here in [`_handleRequestStart`](https://github.com/GoogleChrome/puppeteer/blob/bbd1543ebd6148112d1e0d133530053775587b1d/lib/NetworkManager.js#L244).

docs/api.md
pomerantsev: Done: https://github.com/GoogleChrome/puppeteer/blob/89469192c71d8edbe7dba9cc76ff1907f98f8c3c/docs/api.md#pagewaitfornavigationoptions

aslushnikov:
This should be a part of a frame navigation story, e.g. both `page.goto` and `page.waitForNavigation` should be moved to the frame level.

In order to do this, NavigatorWatcher should be able to gracefully handle frame detachments - this currently never happens, but is a viable scenario for subframes.

While reading your patch I found a few bugs in our implementation of `page.waitForNavigation` - let me handle them.

pomerantsev:
Cool, I'm happy to help if it still makes sense.

aslushnikov:
> Cool, I'm happy to help if it still makes sense.

Thanks! The refactoring naturally concluded in #3276, introducing the navigation API for frames.

PR #2989, opened 2018-07-31
feat(chromium): roll Chromium to r579032

aslushnikov:
This roll includes:
- https://crrev.com/577366 - DevTools: report redirect responses only if response interception is enabled
- https://crrev.com/577212 - DevTools: intercept requests resulting from redirects
- https://crrev.com/578934 - DevTools: Add a protocol method to insertText

Interception Logic in DevTools protocol has changed regarding redirects;
this patch migrates interceptions to dispatch "request" events based on
requestWillBeSent event.

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -28,8 +28,8 @@ class NetworkManager extends EventEmitter {
     this._frameManager = frameManager;
     /** @type {!Map<string, !Request>} */
     this._requestIdToRequest = new Map();
-    /** @type {!Map<string, !Request>} */
-    this._interceptionIdToRequest = new Map();
+    /** @type {!Map<string, !Protocol.Network.requestWillBeSentPayload>} */
+    this._requestIdToRequestWillBeSentEvent = new Map();
     /** @type {!Object<string, string>} */
     this._extraHTTPHeaders = {};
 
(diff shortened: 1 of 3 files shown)


lib/NetworkManager.js
```diff
@@ -151,25 +170,36 @@ class NetworkManager extends EventEmitter {
...
 
+  /**
+   * @param {!Protocol.Network.requestWillBeSentPayload} event
+   * @param {?string} interceptionId
+   */
+  _onRequest(event, interceptionId) {
+    let redirectChain = [];
+    if (event.redirectResponse) {
+      const request = this._requestIdToRequest.get(event.requestId);
+      // If we connect late to the target, we could have missed the requestWillBeSent event.
```
JoelEinbinder, line 193: This comment is suspicious now.

lib/NetworkManager.js
```diff
@@ -192,15 +222,14 @@ class NetworkManager extends EventEmitter {
     request._response = response;
     request._redirectChain.push(request);
     response._bodyLoadedPromiseFulfill.call(null, new Error('Response body is unavailable for redirect responses'));
-    this._requestIdToRequest.delete(request._requestId);
-    this._interceptionIdToRequest.delete(request._interceptionId);
     this._attemptedAuthentications.delete(request._interceptionId);
+    this._requestIdToRequest.delete(request._requestId);
```
JoelEinbinder: nit: leave above for a nicer diff.

lib/NetworkManager.js
aslushnikov: It's still valid: when we attach to the page after the initial request but before the redirect.

lib/NetworkManager.js

JoelEinbinder (approved):

PR #2990, opened 2018-07-31
test: do not handle sigint for all the browsers launched in tests

aslushnikov:
Puppeteer's default handler issues "process.exit(130)", causing
test runner to shut too early.

As a result, test runner doesn't show run summary.


JoelEinbinder (approved):
Looks good assuming it works with multiple threads enabled.

PR #3001, opened 2018-07-31
fix(page): Use _pageBindings as a Map

### LINKED ISSUE #3000: Page._pageBindings is declared as `Map<string,Function>` but used as JS object
### Steps to reproduce

1. Turn on noImplicitAny in tsconfig.json and open Page.js.
2. Note error on `this._pageBindings[name]`: Element implicitly has an 'any' type because 'Map<string, Function>` has no index signature.

* Puppeteer version: checked out from master

Based on its declared type, `this._pageBindings[name]` should variously be `this._pageBindings.get/has/set` instead. This is what leads to the compile error. Technically treating a Map as a plain old JS object *works*, but it's a waste of a Map.

sandersn:
In accordance with its declared type, `Map<string, Function>`.
Currently, it is used as a plain old JS object. The compiler marks this
usage as an error when `noImplicitAny: true`. This change switches to
use the appropriate Map methods `has/get/set`.

Fixes #3000

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -384,9 +384,9 @@ class Page extends EventEmitter {
    * @param {function(?)} puppeteerFunction
    */
   async exposeFunction(name, puppeteerFunction) {
-    if (this._pageBindings[name])
+    if (this._pageBindings.has(name))
       throw new Error(`Failed to add page binding with name ${name}: window['${name}'] already exists!`);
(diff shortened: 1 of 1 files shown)


sandersn:
I missed the rule that says scopes are lowercase. How do you prefer that I fix the commit title?

aslushnikov (approved):
Thanks!

PR #3003, opened 2018-07-31
feat(browsercontext): add BrowserContext.pages() method

aslushnikov:


lib/Browser.js
```diff
@@ -250,6 +250,18 @@ class BrowserContext extends EventEmitter {
     return this._browser.targets().filter(target => target.browserContext() === this);
   }
 
+  /**
+   * @return {!Promise<!Array<!Puppeteer.Page>>}
+   */
+  async pages() {
+    const pages = await Promise.all(
```
JoelEinbinder, line 254: Let's use this method to build browser.pages().

JoelEinbinder (approved):

lib/Browser.js
aslushnikov: Done.

PR #3004, opened 2018-07-31
test: use new browser context for every page test

aslushnikov:


JoelEinbinder:
What does this win us?

aslushnikov:
> What does this win us?

1. this drastically reduces chances of tests collapsing when running parallel
2. dogfoods browsercontexts the way we want them to be used
3. simplifies the dance around service workers / cookies setting up and tiering down .

JoelEinbinder:
How does it effect tests running in parallel? Weren't parallel tests always in their own browser?

aslushnikov:
> How does it effect tests running in parallel? Weren't parallel tests always in their own browser?

Ah, you're right. Cross-out the first bullet then

test/test.js
```diff
@@ -112,41 +112,49 @@ describe('Page', function() {
...
-
-  afterEach(async state => {
-    state.tearDown();
-    await state.page.close();
-    state.page = null;
+  describe('Page', function() {
+    beforeEach(async(state, test) => {
+      state.context = await state.browser.createIncognitoBrowserContext();
+      state.page = await state.context.newPage();
+      const rl = require('readline').createInterface({input: state.browser.process().stderr});
```
JoelEinbinder: Can you move the stderr code into its own beforeEach in the `Browser` block? Otherwise the browser context tests lose out on stderr recording.

aslushnikov:
Accidentally closed this; re-opened as #3010.

PR #3005, opened 2018-07-31
feat: specify the default viewport

JoelEinbinder:
Removes the undocumented `appMode`, adds a public option `defaultViewport`

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -359,6 +359,13 @@ puppeteer.launch().then(async browser => {
 - `options` <[Object]>
   - `browserWSEndpoint` <[string]> a [browser websocket endpoint](#browserwsendpoint) to connect to.
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
+  - `defaultViewport` <?[Object]> . Defaults to an 800x600 viewport. `null` disables the default viewport.
+    - `width` <[number]> page width in pixels.
+    - `height` <[number]> page height in pixels.
+    - `deviceScaleFactor` <[number]> Specify device scale factor (can be thought of as dpr). Defaults to `1`.
+    - `isMobile` <[boolean]> Whether the `meta viewport` tag is taken into account. Defaults to `false`.
+    - `hasTouch`<[boolean]> Specifies if viewport supports touch events. Defaults to `false`
+    - `isLandscape` <[boolean]> Specifies if viewport is in landscape mode. Defaults to `false`.
   - `slowMo` <[number]> Slows down Puppeteer operations by the specified amount of milliseconds. Useful so that you can see what is going on.
 - returns: <[Promise]<[Browser]>>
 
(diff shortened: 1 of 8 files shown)


docs/api.md
```diff
@@ -359,6 +359,13 @@ puppeteer.launch().then(async browser => {
 - `options` <[Object]>
   - `browserWSEndpoint` <[string]> a [browser websocket endpoint](#browserwsendpoint) to connect to.
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
+  - `defaultViewport` <?[Object]> . Defaults to an 800x600 viewport. `null` disables the default viewport.
```
vsemozhetbyt: Nit: unneeded spaces and period in `<?[Object]> . Defaults`?

docs/api.md
```diff
@@ -383,6 +390,13 @@ This methods attaches Puppeteer to an existing Chromium instance.
   - `headless` <[boolean]> Whether to run browser in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). Defaults to `true` unless the `devtools` option is `true`.
   - `executablePath` <[string]> Path to a Chromium or Chrome executable to run instead of the bundled Chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
   - `slowMo` <[number]> Slows down Puppeteer operations by the specified amount of milliseconds. Useful so that you can see what is going on.
+  - `defaultViewport` <?[Object]> . Defaults to an 800x600 viewport. `null` disables the default viewport.
```
vsemozhetbyt: Ditto.

docs/api.md
JoelEinbinder: Oops, accidentally deleted the sentence that was there. Good catch!

aslushnikov (approved):

lib/Launcher.js
```diff
@@ -155,17 +150,19 @@ class Launcher {
     /** @type {?Connection} */
     let connection = null;
     try {
-      const connectionDelay = options.slowMo || 0;
+      const {
+        ignoreHTTPSErrors = false,
+        defaultViewport = {width: 800, height: 600},
+        slowMo = 0
+      } = options;
```
Palgie, line 157: Will this not cause timeout to be `null` below and always default to 30000?

Palgie: Getting issues with third parties using puppeteer where users are reporting not being able to change timeout anymore.

Palgie:
@aslushnikov @vsemozhetbyt

lib/Launcher.js
JoelEinbinder: The code for setting the timeout is two lines below this, and remains unchanged. Can you file an issue? I don't think this patch is the cause.

cbravo: @Palgie this code above is simply pulling values OUT of the options object using es6 destructuring . So it is not the cause of the timeout value not being set as you described [here](https://github.com/chrisvfritz/prerender-spa-plugin/issues/222)

PR #3012, opened 2018-08-01
test: attempt to make CSS coverage tests less flaky on win

aslushnikov:
This patch starts using networkidle0 event for all navigations in
CSS coverage tests.


aslushnikov:
@JoelEinbinder didn't like this.

Issue #3013, opened 2018-08-01
puppeteer executablePath doesn't seem to be honored (on windows)


jonmadison:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version:
    "puppeteer": "^1.4.0",
    (also tried 1.6.x) 

* Platform / OS version: WIndows 10
* URLs (if applicable):
* Node.js version: 10.5.0

**What steps will reproduce the problem?**

1. in code, set executablePath to an existing chromium browser as an option
(`browserpath` set earlier as "C:\\Users\\jonmadi\\Documents\\chrome-win32\\chrome.exe")
```
  let browserOptions = {
                          executablePath: browserpath,
                          headless: headless,
                          slowMo: speed
                       }

  log.debug(`browserOptions: ${JSON.stringify(browserOptions)}`)
  log.debug(`user: ${JSON.stringify(user)}`)
  //step 1 launch browser
  process.env['PUPPETEER_SKIP_CHROMIUM_DOWNLOAD'] = true
  const browser = await puppeteer.launch(browserOptions);
```

I've also tried setting PUPPETEER_SKIP_CHROMIUM_DOWNLOAD = true on the command line
and for giggles, "false".

2. run code, puppeteer launch executes

**What is the expected result?**
puppeteer would launch and begin to take screenshots

**What happens instead?**
get error that chromium is not installed.

log output:
``` browserOptions: {"executablePath":["C:\\Users\\jonmadi\\Documents\\chrome-win32\\chrome.exe"],"headless":false,"slowMo":0}
{ AssertionError [ERR_ASSERTION]: Chromium revision is not downloaded. Run "npm install" or "yarn install"
    at Console.assert (console.js:191:23)
    at Function.launch (E:\Automation\src\NodePageScreenshotTool\node_modules\puppeteer\lib\Launcher.js:100:15)
    at <anonymous>
  generatedMessage: false,
  name: 'AssertionError [ERR_ASSERTION]',
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '==' }
```

interestingly enough, it runs on mac, but the excutablePath() function *after* launching does not reflect the custom set browser, but the one inside node_modules.

aslushnikov:
@jonmadison how do you define the `browserpath` variable? It indeed won't work if you create it as a String object rather than string literal, e.g. `new String('/usr/var/...')`

jonmadison:
[...]

PR #3014, opened 2018-08-02
fix(launcher): support String objects passed as executablePath option

aslushnikov:
References #3013.


aslushnikov:
This was addressed by #2950.

PR #3023, opened 2018-08-06
feat(launcher): add additional options to dumpio

### LINKED ISSUE #1292: More flexible dumpio
Enabling `dumpio` is quite useful as I'm experiencing some instability with headless chrome, but it can be very noisy.

It would be good if we could filter chrome's output (e.g. to filter out uninteresting messages based on tags like CONSOLE:), as well as redirect it elsewhere (e.g. to a file, or a buffer).

ghost:
Launcher now supports dumpOptions with the following properties:
  - file: redirect dumpio to a file
  - ignore: removes output based on user-specified criteria

Fixes #1292


aslushnikov:
Let's not land this; we've never had an urge to have this, and there's also a `browser.process()` exposed that allows you to pick into stdout/stderr manually.

PR #3042, opened 2018-08-07
docs(api.md): clarification note on request interception

### LINKED ISSUE #3040: Requests don't finish when request interception is enabled but there is no handler to call `request.continue`
### Steps to reproduce

* Puppeteer version: 1.6.2 
* Platform / OS version: OSX
* URLs (if applicable):
* Node.js version: 10.8

This also reproduces on the puppeteer playground at https://try-puppeteer.appspot.com/

```

try {
  await page.goto('https://example.com', { timeout: 5000 });
  console.log("Done!");
} catch(e) {
  console.log(e);
}

await browser.close();
```

The request finishes.

**What happens instead?**
The request stalls indefinitely. A simple handler like the following fixes it:

```
page.on('request', (req) => req.continue());
```

I encountered this when I enabled request interception and only used a `response` handler. Of course it's not a big deal to add a small request handler, so if this isn't easy to fix, then it would be great to get a line in the documentation mentioning that once request interception is enabled, a handler must call `req.continue()` for requests to finish.

aslushnikov:
Fixes #3040.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1486,7 +1486,9 @@ The extra HTTP headers will be sent with every request the page initiates.
 Activating request interception enables `request.abort`, `request.continue` and
 `request.respond` methods.  This provides the capability to modify network requests that are made by a page.
 
+Once request interception is enabled, every request will stall unless it's continued, responded or aborted.
 An example of a naïve request interceptor that aborts all image requests:
+
 ```js
 const puppeteer = require('puppeteer');
 


JoelEinbinder (approved):

docs/api.md
```diff
@@ -1486,7 +1486,9 @@ The extra HTTP headers will be sent with every request the page initiates.
 Activating request interception enables `request.abort`, `request.continue` and
 `request.respond` methods.  This provides the capability to modify network requests that are made by a page.
 
+Once request interception is enabled, every request will stall unless it's continued, responded or aborted.
 An example of a naïve request interceptor that aborts all image requests:
+
```
JoelEinbinder, line 1491: should this newline be one line higher?

aslushnikov: Nah, I thought it's nice to illustrate my addition with an example :)

PR #3047, opened 2018-08-08
feat: prepare for publishing puppeteer-core

aslushnikov:


JoelEinbinder:
Did we want to call it pptr-core?

aslushnikov:
We had plans to rename Puppeteer into pptr, but I don't see it happening any time soon.

JoelEinbinder (approved):

PR #3048, opened 2018-08-08
feat: expose frame's execution contexts

aslushnikov:
This patch exposes frame's execution contexts, making it possible
to debug extension's content scripts.

This is a resurrected #2812.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -58,6 +58,8 @@ Next Release: **Aug 9, 2018**
   * [event: 'dialog'](#event-dialog)
   * [event: 'domcontentloaded'](#event-domcontentloaded)
   * [event: 'error'](#event-error)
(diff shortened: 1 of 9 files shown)


JoelEinbinder:
I think we should either have
`frame.createIsolatedWord` and `frame.isolatedWorlds` or `frame.createExecutionContext` and `frame.executionContexts`.

aslushnikov:
I don't think aligning one with another gives us much here, given the narrow API surface and limited usecases.

Let's use DevTools conventions instead of inventing a new vocabulary. They're well-thought, time-proof, and protocol-aligned.
- `frame.executionContexts()` to query
- `frame.createIsolatedWorld()` to create new

JoelEinbinder (approved):

PR #3049, opened 2018-08-08
feat(launcher): ignoreDefaultArgs to accept array of options

aslushnikov:
If `ignoreDefaultArgs` is given an array of options, than
these options will be excluded from the default command-line
flags.

### DIFF
--- a/test/puppeteer.spec.js
+++ b/test/puppeteer.spec.js
@@ -201,6 +201,19 @@ module.exports.addTests = function({testRunner, expect, PROJECT_ROOT, defaultBro
         await page.close();
         await browser.close();
       });
+      it('should filter out ignored default arguments', async() => {
+        // Make sure we launch with `--enable-automation` by default.
+        const defaultArgs = puppeteer.defaultArgs();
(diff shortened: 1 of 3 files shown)


test/puppeteer.spec.js
```diff
@@ -201,6 +201,19 @@ module.exports.addTests = function({testRunner, expect, PROJECT_ROOT, defaultBro
         await page.close();
         await browser.close();
       });
+      it('should filter out ignored default arguments', async() => {
+        // Make sure we launch with `--enable-automation` by default.
+        const defaultArgs = puppeteer.defaultArgs();
+        const browser = await puppeteer.launch(Object.assign({}, defaultBrowserOptions, {
+          // Ignore first and third default argument.
+          ignoreDefaultArgs: [ defaultArgs[0], defaultArgs[2] ],
```
JoelEinbinder, line 209: this test seems fragile if the order of arguments changes.

JoelEinbinder (approved):

test/puppeteer.spec.js
aslushnikov: why does the order matter?

PR #3056, opened 2018-08-09
feat: introduce puppeteer/Errors

### LINKED ISSUE #1694: Conditional catch for errors

aslushnikov:
This patch adds a new require, `puppeteer/Errors`, that
holds all the Puppeteer-specific error classes.

Currently, the only custom error class we use is `TimeoutError`. We'll
expand in future with `CrashError` and some others.

Fixes #1694.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -12,6 +12,7 @@ Next Release: **Aug 9, 2018**
 <!-- GEN:toc -->
 - [Overview](#overview)
 - [Environment Variables](#environment-variables)
+- [Error handling](#error-handling)
 - [Working with Chrome Extensions](#working-with-chrome-extensions)
 - [class: Puppeteer](#class-puppeteer)
   * [puppeteer.connect(options)](#puppeteerconnectoptions)
@@ -285,6 +286,7 @@ Next Release: **Aug 9, 2018**
   * [coverage.startJSCoverage(options)](#coveragestartjscoverageoptions)
   * [coverage.stopCSSCoverage()](#coveragestopcsscoverage)
   * [coverage.stopJSCoverage()](#coveragestopjscoverage)
+- [class: TimeoutError](#class-timeouterror)
 <!-- GEN:stop -->
 
 ### Overview
@@ -315,6 +317,34 @@ If puppeteer doesn't find them in environment, lowercased variant of these varia
 - `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step.
 - `PUPPETEER_DOWNLOAD_HOST` - overwrite host part of URL that is used to download Chromium
 - `PUPPETEER_CHROMIUM_REVISION` - specify a certain version of chrome you'd like puppeteer to use during the installation step.
+
+### Error handling
+
+Often times, async methods might throw an error, signaling about their inability
+to fulfill request. For example, [page.waitForSelector(selector[, options])](#pagewaitforselectorselector-options)
+might fail if selector doesn't appear during the given timeframe.
+
+For certain types of errors Puppeteer uses specific error classes.
+These classes are available through the `require('puppeteer/Errors')`.
+
+List of supported classes:
+- [`TimeoutError`](#class-timeouterror)
+
+An example of handling timeout error:
+```js
+const {TimeoutError} = require('puppeteer/Errors');
+
+// ...
+
+try {
+  await page.waitForSelector('.foo');
+} catch (e) {
+  if (e instanceof TimeoutError) {
+    // Do something if this is a timeout.
+  }
+}
+```
+
 
 ### Working with Chrome Extensions
 
@@ -3156,6 +3186,14 @@ _To output coverage in a form consumable by [Istanbul](https://github.com/istanb
 
 > **NOTE** JavaScript Coverage doesn't include anonymous scripts by default. However, scripts with sourceURLs are
 reported.
+
+### class: TimeoutError
+
+* extends: [Error]
+
+TimeoutError is emitted whenever certain operations are terminated due to timeout, e.g. [page.waitForSelector(selector[, options])](#pagewaitforselectorselector-options) or [puppeteer.launch([options])](#puppeteerlaunchoptions).
+
+
 
 [Array]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array "Array"
 [boolean]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures#Boolean_type "Boolean"
--- a/lib/Errors.js
+++ b/lib/Errors.js
@@ -1,0 +1,29 @@
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class CustomError extends Error {
+  constructor(message) {
+    super(message);
+    this.name = this.constructor.name;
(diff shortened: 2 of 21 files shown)


utils/doclint/check_public_api/index.js
```diff
@@ -42,6 +43,10 @@ const EXCLUDE_METHODS = new Set([
   'JSHandle.toString',
 ]);
 
+const EXCLUDE_PROPERTIES = new Set([
```
JoelEinbinder: Methods are properties. I think you can just rename EXCLUDE_METHODS to EXCLUDE_PROPERTIES and add `TimeoutError.name` into it. They can't conflict.

JoelEinbinder (approved):

utils/doclint/check_public_api/index.js
aslushnikov: Done.

docs/api.md
```diff
@@ -316,6 +318,34 @@ If puppeteer doesn't find them in environment, lowercased variant of these varia
...
+An example of handling timeout error:
+```js
+const {TimeoutError} = require('puppeteer/Errors');
+
+// ...
+
+try {
+  await page.waitForSelector('.foo');
+} catch (e) {
+  if (e instanceof TimeoutError) {
```
felixfbecker, line 342: Imo this is not a good suggestion. `instanceof` is an antipattern. Assert against interfaces, not implementations. Better would be `if (e.name === 'TimeoutError')`

lib/Errors.js
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+class CustomError extends Error {
+  constructor(message) {
+    super(message);
+    this.name = this.constructor.name;
```
felixfbecker, line 20: This will break if the code gets minified in some way. It is better to assign a string constant in each of the subclasses.

aslushnikov: > This will break if the code gets minified in some way. It is better to assign a string constant in each of the subclasses.

We don't assume code minification/transpilation usecases, and we already fail miserably with some tools like babel. Does it hurt you? If yes, please file an issue so that your concerns are tracked.

docs/api.md
aslushnikov: > Imo this is not a good suggestion. instanceof is an antipattern. Assert against interfaces, not implementations. Better would be if (e.name === 'TimeoutError')

I'm not sure I follow. Can you please expand on this?

felixfbecker: When using `instanceof`, you check if an object is an instance of a concrete class implementation. It is better to check if the object adheres to the interface you expect, e.g. if it has the needed properties or in this case, if the error has a certain error code. This is the approach taken by NodeJS errors (you cannot even check with `instanceof` as the classes are not exposed, just check `error.code`) and also browser APIs (for example, `AbortError` is not even a class, but you always have to check `err.name === 'AbortError'`).

Imagine you had two versions of Puppeteer in your tree, because some library depends on it, but another library or your root package does too, but a different version (this might have happened silently through an update). Now when you import `TimeoutError` from Puppeteer, the class will actually be a different class, and all your `instanceof` checks will fail - even though it clearly is a `TimeoutError`, and says so in `error.name`. Relying on a certain layout in node_modules in brittle.

This is why it's best practice to avoid `instanceof` and check error names/codes instead.

aslushnikov: > Imagine you had two versions of Puppeteer in your tree, because some library depends on it, but another library or your root package does too, but a different version (this might have happened silently through an update).

Let me re-iterate if understand correctly. I'm developing project `myproj` that depends on `Puppeteer` and also has a dependency on `FancyLib` that happens to use Puppeteer as well. In addition, in order to make things extreme (and prevent tree flattening), let's consider these Puppeteers to be of different versions.

```
myproj
|
|--FancyLib
|      |--Puppeteer v1.7.0
|
|--Puppeteer v1.8.0
```

In this setup, checking errors emitted by FancyLib to be Puppeteer errors is a logical error since they're not part of FancyLib API:

```js
import {TimeoutError} from 'puppeteer/Errors';
import {FancyLib} from 'fancylib';
try {
  await FancyLib.doStuff();
} catch(e) {
  // Logic error: FancyLib has nothing to do with Puppeteer.
  if (e instanceof TimeoutError) {
  }
}
```

Instead, if FancyLib wants to provide error handling as a part of its api, it should provide explicit errors for its clients to catch:
```js
import {FancyLib, TimeoutError} from 'fancylib';
try {
  await FancyLib.doStuff();
} catch(e) {
  // Correct: FancyLib exposes TimeoutError as part of its API
  if (e instanceof TimeoutError) {
  }
}
```

So when using puppeteer library directly, its errors are part of its api, so its safe to instanceof against them. Using event names as an alternative to classes is possible, but we'd have to come up with a very distinct error names so that there are less chances for them to clash with other libraries. 

To sum up, both approaches are viable, but I'd prefer to keep the doc as-is.

felixfbecker: > we'd have to come up with a very distinct error names so that there are less chances for them to clash with other libraries.

This is the point - you _want_ to have them clash. What else could a `TimeoutError` be other than an error because of a timeout? I.e. as a user, you don't care if it's a timeout error from library FooBarLib, you only care if it is a timeout error. Just like I don't care if it is a FooBarLibAbortError, I only care if it is an AbortError.

Yes, both approaches work usually. I've just seen people being tripped up too often because an update changed their node_modules and suddenly their `instanceof` silently failed that worked before, so I believe it's better to spread the practice of checking against names.

If you don't want to change it that's fine, it was just a suggestion.

aslushnikov: > This is the point - you want to have them clash. What else could a TimeoutError be other than an error because of a timeout?

That's fair for the timeout error, but I disagree in general. For example, the `CrashError` we're about to introduce for target crashes might mean very different things from library to library, with different recovery strategies.

IMO since node.js platform doesn't define a set of common errors with a well-defined semantics, we shouldn't invent one as well.

> If you don't want to change it that's fine, it was just a suggestion.

Thank you for the suggestion!

felixfbecker: I can definitely see your point.

PR #3069, opened 2018-08-10
Null check `args`

brian-gates:
Fixes https://github.com/GoogleChrome/puppeteer/issues/3068, i.e. `puppeteer.launch({ args: null })` resulting in error.


brian-gates:
I signed it!

aslushnikov:
Thanks for the PR. I think throwing is fine in this case.

PR #3078, opened 2018-08-11
docs(readme): add a way that install puppeteer faster

suhaotian:


aslushnikov:
How's this different from [puppeteer-core](https://github.com/GoogleChrome/puppeteer#puppeteer-core)?

PR #3088, opened 2018-08-15
Add Puppeteer Recorder extension to docs

tnolet:


aslushnikov (approved):

ebidel:
This extension doesn't work all that well for. I'm not sure we should really be highlighting it just yet.

aslushnikov:
> This extension doesn't work all that well for. I'm not sure we should really be highlighting it just yet.

Fair.

PR #3099, opened 2018-08-16
refactor: drop jshandle factory from ExecutionContext

aslushnikov:
This patch:
- merges `ElementHandle` into `ExecutionContext` (for simplicity; there's no good reason to have them in separate files).
- removes the necessity to pass handle factory when creating `ExecutionContext`

This makes it easier to create execution contexts out of payloads.


package.json
```diff
@@ -8,7 +8,7 @@
     "node": ">=6.4.0"
   },
   "puppeteer": {
-    "chromium_revision": "579032"
```
JoelEinbinder: what

lib/Worker.js
```diff
@@ -33,7 +33,7 @@ class Worker extends EventEmitter {
     let jsHandleFactory;
     this._client.once('Runtime.executionContextCreated', async event => {
       jsHandleFactory = remoteObject => new JSHandle(executionContext, client, remoteObject);
```
JoelEinbinder, line 35: remove jsHandleFactory

lib/Page.js
```diff
@@ -27,6 +27,7 @@ const Tracing = require('./Tracing');
 const {helper, debugError, assert} = require('./helper');
 const {Coverage} = require('./Coverage');
 const {Worker} = require('./Worker');
+const {createJSHandle} = require('./ExecutionContext');
```
JoelEinbinder, line 30: JSHandleFromProtocol

lib/FrameManager.js
```diff
@@ -259,17 +254,19 @@ FrameManager.Events = {
...
 class Frame {
   /**
+   * @param {!FrameManager} frameManager
    * @param {!Puppeteer.CDPSession} client
    * @param {?Frame} parentFrame
    * @param {string} frameId
    */
-  constructor(client, parentFrame, frameId) {
+  constructor(frameManager, client, parentFrame, frameId) {
+    this._frameManager = frameManager;
```
JoelEinbinder, line 263: page

JoelEinbinder (approved):

package.json
aslushnikov: rebaselined.

lib/FrameManager.js
aslushnikov: done

aslushnikov: Not done; we need framemanager in element handle to resolve content frames

Issue #1215, opened 2017-10-30
Support extensions execution contexts


Banou26:
The [documentation](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#overview)  show that Frames can be associated with extensions but there's no documentation on how to register an extension & run a script under one of the extensions environments (option page/background page/page action/ect...).

It would be awesome to run automated tests for web-extensions developpers.

ebidel:
Can you provide an example of what you're trying to achieve?

Banou26:
Something like that https://github.com/Banou26/PuppeteerWebExt
It's basically being able to register extensions & interact with it

ebidel:
Ah Interesting! What's the benefit of using an extension instead of using puppeteer directly to modify/work with the page? I'd imagine most things you want to use the crx web extension API for, you can do in puppeteer's API. 

BTW, there's a [flag for loading chrome with an extension](https://peter.sh/experiments/chromium-command-line-switches/#load-extension). Not sure we need a new api (`browser.registerExtension`).

I'll let @aslushnikov speak to the feasibility of working with an extension like this. I'm not familiar enough with where the limitations would be.

Note that puppeteer [disables extensions by default](https://github.com/GoogleChrome/puppeteer/blob/ab9b34cd5dad54e170be916016a4e9b4097231ed/lib/Launcher.js#L41), so we'd need some way to flip that off.

Banou26:
There's no benefits of using an extension over puppeteer directly.

It's just that there's almost no tools for web extension developers and it's a real hassle to test extensions in general, so i though a tool like puppeteer would be just awesome to run tests on web extensions and know when things break.

Yes, there's no real need of implementing a new api to register extensions if there's a flag, but is there a way to get their id (to access the options page (chrome-extension://id) via a puppeteer page) ?

aslushnikov:
@Banou26 the documentation you pointed to also says:

> On the following diagram, faded entities are not currently represented in Puppeteer.

So this is not currently supported by puppeteer.

> Yes, there's no real need of implementing a new api to register extensions if there's a flag, but is there a way to get their id (to access the options page (chrome-extension://id) via a puppeteer page) ?

DevTools protocol [has this capability](https://aslushnikov.github.io/vanilla-protocol-viewer/#Runtime.ExecutionContextDescription), but we haven't exposed it through pptr api yet.

What exactly do you need from PPTR api?

Banou26:
[...]

PR #3100, opened 2018-08-16
refactor(waittask): prepare WaitTask for IsolatedWorlds

aslushnikov:
`WaitTask` will belong to `IsolatedWorld` rather then `Frame`.
This patch decouples `WaitTask` from `Frame`.

References #1215


JoelEinbinder:
How can `WaitTask` belong to an isolated world? Worlds don't survive navigation. If I start a WaitTask in an isolated world, and the frame the world belongs to navigates, what execution context will the waittask be restarted in, if any?

PR #3108, opened 2018-08-20
feat(framemanager): click handler random option on multiple elements

june07:
If multiple elements are returned via the selector, then instead of only allowing the first selector, allow a random one to be chosen.


JoelEinbinder:
What's the rationale for this?

aslushnikov:
I don't think this has enough reasons to be landed.

PR #3128, opened 2018-08-23
feat(page): Added 'referer' as a parameter in options of Page.goto()

### LINKED ISSUE #3090: Referrer should be parameter, not extra header
If you look at this line, we trying to get referrer from `extraHTTPHeaders`:
https://github.com/GoogleChrome/puppeteer/blob/master/lib/Page.js#L565

So, at this code example I forced to clear extra headers for proper resources loading:
```js
  referer: 'https://example.com/'
});

await page.goto('https://github.com/');

await page.setExtraHTTPHeaders({});
```

This is really bad way to set referrer, because **even in CDP** referrer is specified as a parameter, not header. This is not even funny!
https://github.com/GoogleChrome/puppeteer/blob/master/lib/Page.js#L601

While I'm trying to clear headers after navigation with empty object, page will load resources (like CSS and scripts, images) with a **wrong** referrer!

Try to set extra header `referer` on your page and load image, and you will see amazing wrong referer:

ref

So wee need smth like this:
* referrer as a optional parameter at `goto`, in this case method will ignore headers `referer` check
* referrer as a parameter only (best way)

Now I'm forced to rewrite `goto` method a little, can we fix it?

aliahmad490:
If referer is passed to the options object its value will be used as the referer instead of the value set by Page.setExtraHTTPHeaders().

Fixes #3090

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1268,6 +1268,7 @@ Navigate to the next page in history.
     - `domcontentloaded` - consider navigation to be finished when the `DOMContentLoaded` event is fired.
     - `networkidle0` - consider navigation to be finished when there are no more than 0 network connections for at least `500` ms.
     - `networkidle2` - consider navigation to be finished when there are no more than 2 network connections for at least `500` ms.
+  - `referer` <[string]> Referer header value. If provided it will take prefrence over the referer header value set by [page.setExtraHTTPHeaders()](#pagesetextrahttpheadersheaders).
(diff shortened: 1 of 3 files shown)


aliahmad490:
Can anyone review this please?

aslushnikov (approved):

aslushnikov:
@aliahmad490 thank you for the PR!

Issue #3149, opened 2018-08-28
checking if browser is still open and working


Hretic:
im trying to open multiple tabs in a single browser instance  , after they're done i close the tabs and then re-opening those tabs every x seconds ... but i want to keep the browser itself open so i dont have to login in each tab on every loop

so the browser remains open , tabs open and close 

here is my simplified code  , pleas ignore syntax erros 

    var  global_browser = false ;
    async function init_puppeteer( settings ) {
    
    
    
        if(global_browser === false )
            global_browser = await puppeteer.launch({headless: false  , args:['--no-sandbox']});
    
        for(var i = 0  ; i < settings.length ; i++ )
        {
            var setting = settings[i];
    		open_tab($setting);
    	}
    }
    
     
     
     async function open_tab( setting ){
        const page    = await global_browser.newPage();
    	// do stuff 
    	page.close();	
    }
    
    setInterval(function (){
        init_puppeteer(settings)
    }, 50000 );

here is the problem , sometimes browser crashes or it get closed for whatever reason but the `global_browser` remains an object /instance of puppeteer  ... of curse it wont work when i try open a tab with that instance  , and i get something like

    (node:5720) UnhandledPromiseRejectionWarning: Error: WebSocket is not open: readyState 3 (CLOSED)

here is my question , how can i check and make sure i have a working/open instance of puppeteer in `global_browser` ? so i can create a new instance and replace it if the previous one is closed

ris58h:
Why just not to handle that error? Even if there is/was such method to check the browser's state, the browser can crash right after this check.

Hretic:
the whole process repeats in a setInterval loop so if it crashes right after check i'll re-open it in the next loop (if i can figure out how to check the browser) ..... but its also a good idea to check for that specific error , that would be my  plan B ..... the thing is crashes might result on different error(s) ... i got this error by closing the browser myself so i think its more reliable to check the state of browser

ShragaUser:
Hi! 
we can check to see if browser is open and running by accessing the browsers ChildProcess instance and see if the process itself is running on the computer via the exposed pid. 

also we could add logic to chrome process 'close' event to emit an event  or update Browser's new property such as "ísOpen".

@Hretic - for now as a workaround you can add an event listener on the local browser instance _process to let you know you need to "restart" the browser in your code. 

any thoughts?

ShragaUser:
@Hretic  - something of the sorts:
`    global_browser._process.once('close',()=>{
[...]

PR #3171, opened 2018-09-01
Feature: see if chrome process is currently open

ShragaUser:
Exposing a function on browser instance to see if Chrome process is currently open.

This feature addresses ISSUE #3149 

This is actually my first time contributing to an open source program.
so i would love your help with anything here. 

thank you! 👍


ShragaUser:
@googlebot  I signed it!

PR #3180, opened 2018-09-03
feat: add page.clear method

### LINKED ISSUE #3094: Feature request: Add 'clear' option to the .type() function
Function concerned : ``.type`` - [api here](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#pagetypeselector-text-options)

Basically, add a boolean ``clear`` into the options so we specify if we want to clear the field before typing.

`page.type('#myinput', 'toto', {clear: true})`

This would clear any existing input into ``#myinput`` then it would type ``toto``.

yanivefraim:
1. I wasn't sure about implementation (if to use `setSelection` + backspace or just to clear the value). Setting `element.value = ''` was the easiest... (:
2. Didn't handle case of non input type selector (what do we want to do?)

Fixed #3094

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -84,6 +84,7 @@ Next Release: **Sep 6, 2018**
   * [page.authenticate(credentials)](#pageauthenticatecredentials)
   * [page.bringToFront()](#pagebringtofront)
   * [page.browser()](#pagebrowser)
+  * [page.clear(selector)](#pageclearselector)
   * [page.click(selector[, options])](#pageclickselector-options)
   * [page.close(options)](#pagecloseoptions)
   * [page.content()](#pagecontent)
@@ -1017,6 +1018,10 @@ Brings page to front (activates tab).
 - returns: <[Browser]>
 
 Get the browser the page belongs to.
+
+#### page.clear(selector)
+- `selector` <[string]> A [selector] of an element to clear. If there are multiple elements satisfying the selector, the first will be used.
+- returns: <[Promise]>
 
 #### page.click(selector[, options])
 - `selector` <[string]> A [selector] to search for element to click. If there are multiple elements satisfying the selector, the first will be clicked.
--- a/lib/Input.js
+++ b/lib/Input.js
@@ -159,8 +159,6 @@ class Keyboard {
     let delay = 0;
     if (options && options.delay)
       delay = options.delay;
-    if (!text)
-      await this.press(text, {delay});
     for (const char of text) {
       if (keyDefinitions[char])
         await this.press(char, {delay});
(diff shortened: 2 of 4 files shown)


yanivefraim:
In case we want to use `setSelectionRange`. Do we want something like:

```js
async clear(selector) {
    await this.focus(selector);
    await this.$eval(selector, el => el.setSelectionRange(0, el.value.length));
    return this.keyboard.press('Backspace');
}
```

?

aslushnikov:
@JoelEinbinder, can you please take a look? I remember you had thoughts regarding this.

JoelEinbinder:
I don't like this approach. Its a one liner in the current API, so I don't really see the need. This method doesn't handle all the special cases: contenteditable input, `defaultValue` being set, something listening for `change` or `input`, or a page rerouting focus.

I'd rather have `page.$eval(selector, el => el.value = '')` in my project, because it holds no secrets about what it is or isn't going to do.

yanivefraim:
@JoelEinbinder - I agree

aslushnikov:
@yanivefraim I'll close this for now.

Niceplace:
@JoelEinbinder @aslushnikov I have interesting behavior right now with your proposed solution and I'd like to have your advice on it.

I have a helper function that helps us type text with puppeteer and as of today i've noticed that the following behavior happens (Puppeteer 1.11.0) :

- Part of the text is typed
- The field is cleared after a few characters have been typed
- The rest of the text is typed

Ultimately, this causes a test to fail because the input wasn't filled properly. Could this be because the act of clearing the field with `page.$eval` doesn't actually return a promise ? I'm basing myself on the [documentation](https://pptr.dev/#?product=Puppeteer&version=v1.11.0&show=api-pageevalselector-pagefunction-args-1) available on `pptr.dev` regarding `page.$eval` which states that : "If pageFunction returns a Promise, then page.$eval would wait for the promise to resolve and return its value."

Here's the code for my helper function, we are using this with jest and `page` is a global that refers to a Puppeteer `Page` object. The other two arguments are the selector to type text into and the text to type.

```
const waitAndFill = async (
  selector,
  text,
  pageInstance = page,
) => {
  try {
    const selectorTagName = await pageInstance.$eval(selector, el => el.tagName);
    if (selectorTagName.toLowerCase().trim() !== 'input') {
      debug(
        `[ERROR-waitAndFill]: ${selector} of type ${selectorTagName} is not of type input. Will throw`,
      );
      throw new Error(
        `Cannot fill element ${selector} of type ${selectorTagName}.
        \r\nExpected type 'input'.`,
      );
    }
    // eslint-disable-next-line no-param-reassign,no-return-assign
    await pageInstance.$eval(selector, el => (el.value = ''));
    await pageInstance.type(selector, text);
  } catch (error) {
    console.error(error, `Error while trying to wait and fill input ${selector} with text ${text}`);
    throw error;
  }
};
```

Issue #2479, opened 2018-04-30
page.goto returns null periodically


ntzm:
### Steps to reproduce

Similar/same as: #1391

**Tell us about your environment:**

* Puppeteer version: 1.3.0
* Platform / OS version: Mac
* URLs (if applicable): https://www.microsoft.com/en-gb/store/d/xbox-one-s-1tb-console-playerunknowns-battlegrounds-bundle/908z9jn5cnh2/gz4w?cid=msft_web_collection
* Node.js version: 9.11.1

**What steps will reproduce the problem?**

```js
const browser = await puppeteer.launch();

const page = await browser.newPage();

for (var i = 0; i < 15; i++) {
  const r = await page.goto('https://www.microsoft.com/en-gb/store/d/xbox-one-s-1tb-console-playerunknowns-battlegrounds-bundle/908z9jn5cnh2/gz4w?cid=msft_web_collection', { waitUntil: 'domcontentloaded' });

  console.log(r.ok());
}

await browser.close();
```

**What is the expected result?**

`true` outputted 15 times

**What happens instead?**

Sometimes `true` 15 times, sometimes `TypeError: Cannot read property 'ok' of null`

It works as expected on 1.2.0, but fails on 1.3.0

aslushnikov:
@ntzm Thanks for filing this separately, I can easily repro this.

You can workaround using `waitUntil: "networkidle0"` command while this is being fixed.

ntzm:
Thanks!

peterbe:
Here's an even simpler way to reproduce it:
```javascript
const puppeteer = require("puppeteer");
(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();

  // const Url = "https://www.peterbe.com/#anything";  // FAILS
  const Url = "https://www.peterbe.com#anything";  // ALSO FAILS
  // const Url = "https://www.peterbe.com";  // WORKS
  const r1 = await page.goto(Url, { waitUntil: "domcontentloaded" });
  if (r1 === null) console.log("Response is null!");
  else console.log("Response OK?", r1.ok());

  const r2 = await page.goto(Url, { waitUntil: "domcontentloaded" });
  if (r2 === null) console.log("Response is null!");
  else console.log("Response OK?", r2.ok());

  await browser.close();
})();
```
Output is:
```
Response OK? true
Response is null!
```
Every time. 

Note that removing the anchor part of the `Url` variable value "solves" the problem. 

[...]

PR #3198, opened 2018-09-04
fix(page): Page.goto should properly handle historyAPI in beforeunload

### LINKED ISSUE #2764: Some URLs sequence makes next responses to be null
* Puppeteer version: 1.5.0-post (chromium_revision 567388)
* URLs (if applicable): see in the example
* Node.js version: 10.4.1

I have a script that validates links. Sometimes, after getting some URL, all next responses become `null`.

I've managed to reduce the issue to this case:

```js
'use strict';

(async function main() {
  try {

    let response = await page.goto('https://twitter.com/');
    console.log(response ? response.status() : response);
    response = await page.goto('https://github.com/nodejs/node');
    console.log(response ? response.status() : response);
    response = await page.goto('https://github.com/GoogleChrome/puppeteer/');
    console.log(response ? response.status() : response);

    await browser.close();
  } catch (err) {
    console.error(err);
  }
})();
```

```
200
200
200
```

**What happens instead?**
```
200
null
null
```

aslushnikov:
Protocol's `page.navigate` doesn't report loaderId if the command resulted in a same-document
navigation. In this case we should anticipate same-document navigation and otherwise we should
wait for new document navigation.

Fixes #2764.

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -602,7 +602,7 @@ class Page extends EventEmitter {
         ensureNewDocumentNavigation ? watcher.newDocumentNavigationPromise() : watcher.sameDocumentNavigationPromise(),
       ]);
     }
-    watcher.cancel();
+    watcher.dispose();
     helper.removeEventListeners(eventListeners);
     if (error)
       throw error;
@@ -654,6 +654,7 @@ class Page extends EventEmitter {
       watcher.sameDocumentNavigationPromise(),
       watcher.newDocumentNavigationPromise()
     ]);
+    watcher.dispose();
     helper.removeEventListeners([listener]);
     if (error)
       throw error;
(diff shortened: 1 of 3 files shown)


lib/Page.js
```diff
@@ -644,7 +649,11 @@ class Page extends EventEmitter {
 
     const responses = new Map();
     const listener = helper.addEventListener(this._networkManager, NetworkManager.Events.Response, response => responses.set(response.url(), response));
-    const error = await watcher.navigationPromise();
+    const error = await Promise.race([
+      watcher.timeoutPromise(),
```
JoelEinbinder, line 653: why await the timeout promise here? Won't the same and new document navigations be called with the timeout?

aslushnikov: no, I split them out (see NavigationWatcher)

JoelEinbinder (approved):

ntzm:
It's me again :D Will this fix #2479?

aslushnikov:
@ntzm ah sorry, I still can repro this on tip-of-tree. Let me look into this; i forgot what was the root cause.

ntzm:
No worries, sorry to keep bothering you :D

PR #3201, opened 2018-09-05
docs: Added cutoff values for scale option in page.pdf function

jsfeldman:
I thought it would be useful to add the lowest and highest values for the scale option to the page.pdf function call in the api documentation. The library throws an error and doesn't generate a pdf if you go beyond those values. I didn't manually test it, the values are based on the warning Chrome gives you if you try to set the scale value outside of that range (divided by 100), so let me know if they're wrong.

It may be useful to change the behavior in the future so that if the value is set beyond that range puppeteer sets the value to the closest value within the range and throws a warning.

Let me know what you think!


jsfeldman:
I signed it!

aslushnikov (approved):
Thanks!

Issue #2119, opened 2018-02-28
Feature request: run puppeteer in the browser


Janpot:
Just trying to feel the water here, but it seems to me that apart from downloading chrome and launching a browser, puppeteer isn't really doing anything that can't be done in a browser. I'm thinking `puppeteer.connect()` in a webpage. Would there be any interest in supporting this? Am I overlooking any barriers that are in the way of achieving this? I can probably make some time to look into it.

aslushnikov:
@Janpot this definitely sounds interesting. I don't think it makes sense to have it as a part of this repository - it deserves a separate project.

Do you have any success with this?

Janpot:
I tinkered a bit with it a month ago. I think it's feasible to create a build of puppeteer that runs in the browser but I haven't picked it back up. I think a separate project would only complicate things as the goal would be to not add code at all, just refactor here and there and add a build target.

aslushnikov:
> I think a separate project would only complicate things as the goal would be to not add code at all, just refactor here and there and add a build target.

This would be ideal. @JoelEinbinder had a prototype as well and refactored pptr codebase to simplify things there. However, iirc he still needed to mock `fs` and some other node modules we use occasionally.

Janpot:
@aslushnikov Ok, so I did a quick and very very dirty test again, building `puppeteer` with `browserify`. To make it work:
- I used `exposify` to shim `require('ws')` with `WebSocket`, and `'./BrowserFetcher'` with `null`
- I changed `ChromiumRevision` in `Launcher.js` and set it to `null`
- I removed the second argument in `new WebSocket(url)` in`Connection.js`
- I shimmed `WebSocket` with 
  ```js
    WebSocket.prototype.on = function (eventName, handler) {
      WebSocket.prototype.addEventListener.call(this, eventName, ({data}) => handler(data));
    }
  ```
Then with running browserless locally (`docker run -p 3000:3000 browserless/chrome`), I was able to make 
```js
const puppeteer = require('puppeteer');
puppeteer.connect({ browserWSEndpoint: 'ws://localhost:3000' })
  .then(async browser => {
    const page = await browser.newPage();
    await page.goto('https://example.com');
    console.log(await page.content())
  });
```
work in the browser.

So to list the main problems:
1. `helpers.projectRoot` in `ChromiumRevision`  breaks the browser
2. `WebSocket` seems to be used in a browser incompatible way
3. `BrowserFetcher` needs to be excluded

[...]

PR #3208, opened 2018-09-06
refactor: avoid dynamic requires in lib/ folder

aslushnikov:
This patch removes all dynamic requires in Puppeteer. This should
make it much simpler to bundle puppeteer/puppeteer-core packages.

We used dynamic requires in a few places in lib/:
- BrowserFetcher was choosing between `http` and `https` based on some
  runtime value. This was easy to fix with explicit `require`.
- BrowserFetcher and Launcher needed to know project root to store
  chromium revisions and to read package name and chromium revision from
  package.json. (projectRoot value would be different in node6).
  Instead of doing a backwards logic to infer these
  variables, we now pass them directly from `//index.js`.

With this patch, I was able to bundle Puppeteer using browserify and
the following config in `package.json`:

```json
  "browser": {
    "./lib/BrowserFetcher.js": false,
    "ws": "./lib/BrowserWebSocket",
    "fs": false,
    "child_process": false,
    "rimraf": false,
    "readline": false
  }
```

(where `lib/BrowserWebSocket.js` is a courtesy of @Janpot from
https://github.com/GoogleChrome/puppeteer/pull/2374/)

And command:

```sh
$ browserify -r puppeteer:./index.js > ppweb.js
```

References #2119

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,7 +22,9 @@ try {
 }
 
 // If node does not support async await, use the compiled version.
-if (asyncawait)
-  module.exports = require('./lib/Puppeteer');
-else
-  module.exports = require('./node6/lib/Puppeteer');
+const Puppeteer = asyncawait ? require('./lib/Puppeteer') : require('./node6/lib/Puppeteer');
+const packageJson = require('./package.json');
+const preferredRevision = packageJson.puppeteer.chromium_revision;
+const isPuppeteerCore = packageJson.name === 'puppeteer-core';
+
+module.exports = new Puppeteer(__dirname, preferredRevision, isPuppeteerCore);
--- a/lib/BrowserFetcher.js
+++ b/lib/BrowserFetcher.js
@@ -49,10 +49,11 @@ function existsAsync(filePath) {
 
 class BrowserFetcher {
   /**
+   * @param {string} projectRoot
    * @param {!BrowserFetcher.Options=} options
    */
-  constructor(options = {}) {
-    this._downloadsFolder = options.path || path.join(helper.projectRoot(), '.local-chromium');
+  constructor(projectRoot, options = {}) {
+    this._downloadsFolder = options.path || path.join(projectRoot, '.local-chromium');
     this._downloadHost = options.host || DEFAULT_DOWNLOAD_HOST;
     this._platform = options.platform || '';
     if (!this._platform) {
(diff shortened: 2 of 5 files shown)


Janpot:
@aslushnikov Just want to mention that `ws` supports both browser EventTarget and node EventEmitter interfaces simultaneously. With minimal refactoring, I think it can be made so the `BrowserWebSocket` replacement isn't even necessary.

Edit:
nice work btw, 👍 

Edit2:
[Here](https://github.com/websockets/ws/blob/master/package.json#L19) you can see `ws` is already browserifyable, `puppeteer` is just calling it in a  way that is only compatible in node.

Edit3:
this should do it
https://github.com/GoogleChrome/puppeteer/blob/3d7ae2a2591a55e4f94c67d62a22f7103698cba5/lib/Connection.js#L32
```js
new WebSocket(url, null, { perMessageDeflate: false });
ws.addEventListener('open', () => resolve(new Connection(url, ws, delay)));
ws.addEventListener('error', ({ error }) => reject(error));
```

https://github.com/GoogleChrome/puppeteer/blob/3d7ae2a2591a55e4f94c67d62a22f7103698cba5/lib/Connection.js#L62
```js
this._transport.addEventListener('message', ({ data }) => this._onMessage(data));
this._transport.addEventListener('close', this._onClose.bind(this));
```

https://github.com/GoogleChrome/puppeteer/blob/3d7ae2a2591a55e4f94c67d62a22f7103698cba5/lib/Connection.js#L138
```js
this._transport.addEventListener('error', () => {});
```
package.json:
```json
  "browser": {
    "./lib/BrowserFetcher.js": false,
    "fs": false,
    "child_process": false,
    "rimraf": false,
    "readline": false
  }
```

aslushnikov:
> Here you can see ws is already browserifyable, puppeteer is just calling it in a way that is only compatible in node.

@Janpot Ah, how do I call it the right way?

Janpot:
add a second `null` parameter to the constructor, but keep passing the options in the third one
refactor the `.on` to `.addEventListener` (the signature is a little different)
I edited my previous answer.

Janpot:
🤔 Or second parameter has to be empty array: https://github.com/GoogleChrome/puppeteer/pull/2374/commits/d68b4f65c7de8e8cafb4b863ccc69ac4d8d3f82d#diff-0c1fe9dd78cc7f29823c09b251c8aae0R32
sorry, I'm doing this from memory in between work 🙂

aslushnikov:
@Janpot ah, unfortunately we have a Pipe transport that doesn't support `addEventListener`. I'll play with this separately.

Janpot:
Ah yes, indeed, that makes sense. I guess in that case you could probably go the other way around.

```js
const WebSocket = require('ws');

class WebSocketTransport extends EventEmitter {
  constructor(url) {
    super();
    this._ws = new WebSocket(url, [], { perMessageDeflate: false });
    this._ws.addEventListener('open', event => this.emit('open'));
    this._ws.addEventListener('close', event => this.emit('close'));
    this._ws.addEventListener('error', event => this.emit('error', event.error));
    this._ws.addEventListener('message', event => this.emit('message', event.data));
  }
   send(message) {
    this._ws.send(message);
  }
   close() {
    this._ws.close();
  }
}

// ...

  static async createForWebSocket(url, delay = 0) {
      const ws = new WebSocketTransport(url);
      ws.on('open', () => resolve(new Connection(url, ws, delay)));
      ws.on('error', reject);
    });
  }
```

JoelEinbinder (approved):

Issue #3204, opened 2018-09-05
docs: mention require('puppeteer-core')


ebidel:
https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#puppeteer-vs-puppeteer-core should mention that pptr needs to be imported with a different require statement:

```js
const puppeteer = require('puppeteer-core');
```

PR #3213, opened 2018-09-07
docs: add an include statement for puppeteer-core in api.md

silent-lad:
I did this so as to resolve issue #3204 and to make doc better .


silent-lad:
I signed it!

docs/api.md
```diff
@@ -335,6 +335,12 @@ To sum up, the only differences between `puppeteer-core` and `puppeteer` are:
 
 In most cases, you'll be fine using the `puppeteer` package.
 
+But if you still need `puppeteer-core` remember to change the *include* line :-
```
aslushnikov: let's move this below the "however" paragraph bith bullets: 

When using `puppetee-core`, remember to change the *include* line:
```js
const puppeteer = require('puppeteer-core');
```

silent-lad: Doc changed as suggested by you . @aslushnikov

aslushnikov (approved):
Thanks!

PR #3236, opened 2018-09-13
feat(chromium): roll Chromium to r590951

aslushnikov:
Just a healthy roll.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "node": ">=6.4.0"
   },
   "puppeteer": {
-    "chromium_revision": "588429"
+    "chromium_revision": "590951"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


JoelEinbinder (changes requested):
Worker behavior is different in this release. We should be a bit careful

aslushnikov:
@JoelEinbinder in what release and what's different?

aslushnikov:
For the record (as per discussion with @JoelEinbinder):

> @dgozman made workers not interrupt JavaScript on runtime evaluate; we might be able to remove some worker hacks.

Waiting for Joel to validate that our workers are reliable enough to roll this.

JoelEinbinder (approved):
Looks good

PR #3239, opened 2018-09-13
chore: make sure Puppeteer bundling works

aslushnikov:
This patch:
- adds "browser" field to the package.json with default
  bundling options.
- introduces "bundle" and "unit-bundle" commands to
  create bundle and test bundle
- starts running bundle tests on Travis Node 8 bots

Fixes #2374.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,5 +1,6 @@
 test/assets/modernizr.js
 third_party/*
+utils/browser/puppeteer-web.js
 utils/doclint/check_public_api/test/
 utils/testrunner/examples/
 node6/*
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ package-lock.json
 yarn.lock
 /node6
 /lib/protocol.d.ts
+/utils/browser/puppeteer-web.js
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ script:
   - 'if [ "$NODE8" = "true" ]; then npm run lint; fi'
   - 'if [ "$NODE8" = "true" ]; then npm run coverage; fi'
   - 'if [ "$NODE8" = "true" ]; then npm run test-doclint; fi'
+  - 'if [ "$NODE8" = "true" ]; then npm run bundle; fi'
+  - 'if [ "$NODE8" = "true" ]; then npm run unit-bundle; fi'
   - 'if [ "$NODE6" = "true" ]; then npm run unit-node6; fi'
 jobs:
   include:
--- a/package.json
+++ b/package.json
@@ -24,7 +24,9 @@
     "unit-node6": "node node6/test/test.js",
     "tsc": "tsc -p .",
     "prepublishOnly": "npm run build",
-    "apply-next-version": "node utils/apply_next_version.js"
+    "apply-next-version": "node utils/apply_next_version.js",
+    "bundle": "npx browserify -r ./index.js:puppeteer -o utils/browser/puppeteer-web.js",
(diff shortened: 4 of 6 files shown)


JoelEinbinder (approved):

brandonros:
```
    browserWSEndpoint: '<another-browser-ws-endpont>'
  });
```

if a user has Chrome installed on their computer and loads a page that server`<script src='./puppeteer-web.js'></script>` in said instance of Chrome, is their a command line argument they could pass that would expose their own browser to be controlled via websocket?

aslushnikov:
@brandonros you can start with --remote-debugging-port=0, but you'll still need some way to expose the WS address into the page.

Can you share your scenario? Maybe we can help somehow.

brandonros:
https://github.com/GoogleChrome/puppeteer/issues/2119#issuecomment-427219356

I have clients that would like to have me control their web browser.

I can tell them:

1. "run Chrome with this command line argument"
1. could navigate them to a page that, since they enabled it and opted in, retrieved their WebSocket control address
1. Ran some variation of:

```
<script type="text/javascript" src="puppeteer-web.js"></script>
<script type="text/javascript">

(async () => {
    browserWSEndpoint: theirBrowserWSEndpoint // the exact same browser WS endpoint that just loaded this page
  });

  .. a whole bunch of operations to perform to data scraping, etc.
})()
</script>
```

I see this as the best way to scale Puppeteer. $40 DigitalOcean instances offer terrible performance, but they might have 20 computers in their office they'd gladly run a tab on (provided their tab could follow along to Puppeteer instructions)

brandonros:
Update: so, yeah. This works, and is insanely awesome. 

```
.local-chromium/mac-594312/chrome-mac/Chromium.app/Contents/MacOS/Chromium \
--disable-background-networking \
--disable-background-timer-throttling \
--disable-breakpad \
--disable-client-side-phishing-detection \
--disable-default-apps \
--disable-dev-shm-usage \
--disable-extensions \
--disable-features=site-per-process \
--disable-hang-monitor \
--disable-popup-blocking \
--disable-prompt-on-repost \
--disable-sync \
--disable-translate \
--metrics-recording-only \
--no-first-run \
--safebrowsing-disable-auto-update \
--enable-automation \
--password-store=basic \
--use-mock-keychain \
--remote-debugging-port=0
```

will spit `DevTools listening on ws://127.0.0.1:57683/devtools/browser/foo` into the console

From there,

```
    browserWSEndpoint: 'ws://127.0.0.1:57683/devtools/browser/foo'
});
```

works. I'm pretty mind blown. This is pretty awesome. Is there any way to automate having the user to run the browser with the `--remote-debugging-port=0` command then copy + pasting the exposed output?

entrptaher:
Wasn't this already possible with browserify/webpacking the `puppeteer-core`? :D

Janpot:
@entrptaher not really, `browser` field was missing in `package.json` and a few node.js specific things were still present, even in `puppeteer-core`.

@brandonros I think with minimal changes, `puppeteer` can be packed up in an extension and use the [`chrome.debugger`](https://developer.chrome.com/extensions/debugger) API. You'd probably have to create a [Transport](https://github.com/GoogleChrome/puppeteer/blob/07febb637c78cd59e22a15166f816d838a36e614/lib/WebSocketTransport.js#L21) for it and provide it [here](https://github.com/GoogleChrome/puppeteer/blob/cd54ce388eec426034ae7b7ee5058e54b5ac5da8/lib/Launcher.js#L165).

entrptaher:
Soon people will ask if `screenshot()` and file related api works on this version, since it's not noted.

PR #3254, opened 2018-09-17
Problem with non-HTTP requests.

SzymonLisowiec:
If I enabled request interception, event Page.Events.Request not returns requests with url started with "xyz://" (using in mobile applications for the webview).

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -136,6 +136,7 @@ class NetworkManager extends EventEmitter {
         this._onRequest(event, interceptionId);
         this._requestHashToInterceptionIds.delete(requestHash, interceptionId);
       } else {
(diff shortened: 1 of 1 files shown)


lib/NetworkManager.js
```diff
@@ -136,6 +136,7 @@ class NetworkManager extends EventEmitter {
         this._onRequest(event, interceptionId);
         this._requestHashToInterceptionIds.delete(requestHash, interceptionId);
       } else {
+        this._onRequest(event, null);
```
aslushnikov, line 139: this will double-fire the event. That's probably the reason while bots failed.

aslushnikov:
I don't understand the problem - let's start with filing issue first with a comprehensive explanation.

PR #3258, opened 2018-09-17
refactor: unify response tracking in page.goto and waitForNavigation

aslushnikov:
This patch unifies logic in response tracking in page.goto and
page.waitForNavigation.

As a drive-by, we now make sure that we return the right response
for the right frame. This will come handy for future frame navigation
API.

References #2918

### DIFF
--- a/lib/NavigatorWatcher.js
+++ b/lib/NavigatorWatcher.js
@@ -16,6 +16,7 @@
 
 const {helper, assert} = require('./helper');
 const {FrameManager} = require('./FrameManager');
+const {NetworkManager} = require('./NetworkManager');
 const {TimeoutError} = require('./Errors');
 const {Connection} = require('./Connection');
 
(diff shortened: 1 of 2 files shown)


lib/NavigatorWatcher.js
```diff
@@ -68,6 +74,23 @@ class NavigatorWatcher {
     });
   }
 
+  /**
+   * @param {!Puppeteer.Request} request
+   */
+  _onRequest(request) {
+    if (request.frame() !== this._frame || !request.isNavigationRequest())
+      return;
+    this._navigationRequests.set(request.url(), request);
```
JoelEinbinder: we do we care about the url here? Shouldn't just the most recent navigation request be the only one that matters?

aslushnikov: Fair.

JoelEinbinder (approved):

PR #3262, opened 2018-09-17
feat: async stacks for all "async" public methods

aslushnikov:
This patch traces all public async methods and wraps them
in a helper method that tags the sync stack trace.

Later on, if the method call throws an exception, we add
a captured stack trace to the original stack trace with the "--ASYNC--"
heading.

An example of a stack trace:

```
Error: net::ERR_ABORTED at http://localhost:8907/empty.html
    at navigate (/Users/lushnikov/prog/puppeteer/lib/Page.js:622:37)
  -- ASYNC --
    at Page.<anonymous> (/Users/lushnikov/prog/puppeteer/lib/helper.js:147:27)
    at fit (/Users/lushnikov/prog/puppeteer/test/page.spec.js:546:18)
```

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -138,11 +138,9 @@ class Helper {
   static tracePublicAPI(classType, publicName) {
     for (const methodName of Reflect.ownKeys(classType.prototype)) {
       const method = Reflect.get(classType.prototype, methodName);
-      if (methodName === 'constructor' || typeof methodName !== 'string' || methodName.startsWith('_') || typeof method !== 'function')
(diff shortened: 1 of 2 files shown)


JoelEinbinder:
This seems like the wrong level to solve this problem. Are there no existing solutions? Or at least some prior art?

JoelEinbinder (approved):

felixfbecker:
I thought this was now supported by V8 for async functions natively?

aslushnikov:
@felixfbecker ATM it's supported only with debugger turned on.

noggff:
اكتمال

PR #3265, opened 2018-09-17
feat(launcher): add experimental "transport" option to pptr.connect

aslushnikov:
This patch:
- adds experimental "transport" option to pptr.connect
- uses "transport" option to make sure Puppeteer-Web works with
  Target.exposeDevToolsProtocol

Drive-by: add `browser.target()` to access browser target.

### DIFF
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -186,6 +186,13 @@ class Browser extends EventEmitter {
   }
 
   /**
+   * @return {!Target}
+   */
+  target() {
+    return this.targets().find(target => target.type() === 'browser');
+  }
+
+  /**
    * @return {!Promise<!Array<!Puppeteer.Page>>}
    */
   async pages() {
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -270,7 +270,7 @@ class Launcher {
   }
 
   /**
-   * @param {!(BrowserOptions & {browserWSEndpoint: string})} options
+   * @param {!(BrowserOptions & {browserWSEndpoint: string, transport: ?Puppeteer.ConnectionTransport})} options
    * @return {!Promise<!Browser>}
    */
   async connect(options) {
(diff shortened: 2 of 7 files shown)


lib/Browser.js
```diff
@@ -185,6 +185,13 @@ class Browser extends EventEmitter {
     return Array.from(this._targets.values()).filter(target => target._isInitialized);
   }
 
+  /**
+   * @return {!Target}
+   */
+  target() {
+    return this.targets().find(target => target.type() === 'browser');
```
JoelEinbinder, line 192: this should be a promise right? The Browser can exist before it is notified of its own browser target.

lib/Launcher.js
```diff
@@ -270,18 +270,18 @@ class Launcher {
...
     const {
       browserWSEndpoint,
       ignoreHTTPSErrors = false,
       defaultViewport = {width: 800, height: 600},
+      transport = null,
       slowMo = 0,
     } = options;
-    const transport = await WebSocketTransport.create(browserWSEndpoint);
-    const connection = new Connection(browserWSEndpoint, transport, slowMo);
+    const connection = new Connection(browserWSEndpoint, transport || (await WebSocketTransport.create(browserWSEndpoint)), slowMo);
```
JoelEinbinder: TODO(einbinder): node6 transpiler broke here.

aslushnikov: fixed this one.

lib/Browser.js
aslushnikov: no, we await the setDiscoverTargets call when instantiating browser, so there's always a browser target.

lib/Launcher.js
```diff
@@ -270,17 +270,17 @@ class Launcher {
   }
 
   /**
-   * @param {!(BrowserOptions & {browserWSEndpoint: string})} options
+   * @param {!(BrowserOptions & {browserWSEndpoint: string, transport: ?Puppeteer.ConnectionTransport})} options
```
JoelEinbinder: nit: `transport?: !Puppeteer.ConnectionTransport` is a bit more precise. It isn't nullable, its only optional.

utils/ESTreeWalker.js
```diff
@@ -80,6 +80,7 @@ ESTreeWalker._walkOrder = {
   'ArrayExpression': ['elements'],
   'ArrowFunctionExpression': ['params', 'body'],
   'AssignmentExpression': ['left', 'right'],
+  'AssignmentPattern': ['left', 'right'],
```
JoelEinbinder, line 83: Thanks!

JoelEinbinder (approved):

lib/Launcher.js
aslushnikov: Done.

PR #3266, opened 2018-09-17
refactor: move navigation management to FrameManager

aslushnikov:
This patch:
- moves implementation of page.goto and page.waitForNavigation
  into FrameManager. The defaultNavigationTimeout gets moved to
  FrameManager as well.
- moves NavigatorWatcher into FrameManager to avoid circular dependency


lib/Page.js
```diff
@@ -79,16 +78,15 @@ class Page extends EventEmitter {
     this._keyboard = new Keyboard(client);
     this._mouse = new Mouse(client, this._keyboard);
     this._touchscreen = new Touchscreen(client, this._keyboard);
-    /** @type {!FrameManager} */
```
JoelEinbinder: Can you leave this annotation? I'm pretty sure that it is needed because typescript gets confused when we pass `this` as a parameter to an object in our own constructor.

JoelEinbinder (approved):

PR #3306, opened 2018-09-25
fix(page): full page screenshot when defaultViewport is null

### LINKED ISSUE #3104: Full page screenshot fails when defaultViewport is null
### Steps to reproduce

* Puppeteer version: 1.7.0
* Platform / OS version: Windows 7
* URLs (if applicable):
* Node.js version: v9.10.1

```

(async () => {
    const browser = await puppeteer.launch({ headless: false, defaultViewport: null });
    await page.goto('https://www.google.com/');
    await page.screenshot({ path: 'example.png', fullPage: true });
    await browser.close();
})();
```

No error and screenshot is saved.

**What happens instead?**
`UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'isMobile' of null`

lusarz:
fixes #3104

### DIFF
--- a/test/puppeteer.spec.js
+++ b/test/puppeteer.spec.js
@@ -279,6 +279,19 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions}) 
         expect(page.viewport()).toBe(null);
         await browser.close();
       });
+      it('should take fullPage screenshots when defaultViewport is null', async({server}) => {
+        const options = Object.assign({}, defaultBrowserOptions, {
+          defaultViewport: null
+        });
+        const browser = await puppeteer.launch(options);
+        const page = await browser.newPage();
+        await page.goto(server.PREFIX + '/grid.html');
+        const screenshot = await page.screenshot({
+          fullPage: true
+        });
+        expect(screenshot).toBeInstanceOf(Buffer);
+        await browser.close();
+      });
     });
     describe('Puppeteer.connect', function() {
       it('should be able to connect multiple times to the same browser', async({server}) => {
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -796,19 +796,20 @@ class Page extends EventEmitter {
       clip.scale = 1;
 
     if (options.fullPage) {
-      assert(this._viewport, 'fullPage screenshots do not work without first setting viewport.');
       const metrics = await this._client.send('Page.getLayoutMetrics');
       const width = Math.ceil(metrics.contentSize.width);
       const height = Math.ceil(metrics.contentSize.height);
 
       // Overwrite clip for full page at all times.
       clip = { x: 0, y: 0, width, height, scale: 1 };
-      const mobile = this._viewport.isMobile || false;
-      const deviceScaleFactor = this._viewport.deviceScaleFactor || 1;
-      const landscape = this._viewport.isLandscape || false;
+      const {
+        isMobile = false,
+        deviceScaleFactor = 1,
+        isLandscape = false
+      } = this._viewport || {};
       /** @type {!Protocol.Emulation.ScreenOrientation} */
-      const screenOrientation = landscape ? { angle: 90, type: 'landscapePrimary' } : { angle: 0, type: 'portraitPrimary' };
-      await this._client.send('Emulation.setDeviceMetricsOverride', { mobile, width, height, deviceScaleFactor, screenOrientation });
+      const screenOrientation = isLandscape ? { angle: 90, type: 'landscapePrimary' } : { angle: 0, type: 'portraitPrimary' };
+      await this._client.send('Emulation.setDeviceMetricsOverride', { mobile: isMobile, width, height, deviceScaleFactor, screenOrientation });
     }
     const shouldSetDefaultBackground = options.omitBackground && format === 'png';
     if (shouldSetDefaultBackground)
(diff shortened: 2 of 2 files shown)


aslushnikov:
@lusarz can you please add a test?

lusarz:
@aslushnikov I think I need some help to write test for that. I don't know how to emulate page._viewport to be null in unit tests.

I've tried to create similar test like here:
https://github.com/GoogleChrome/puppeteer/blob/5acf953104a09d7d688d76eadced6a4eba75d5d1/test/page.spec.js#L1648-L1655

but without `await page.setViewport({width: 500, height: 500}); ` line. Unfortunately `page._viewport` is not null in that case. 

I can't also use `await page.setViewport(null)`, because `page.setViewport` method requires `viewport` parameter to be not null.

aslushnikov:
> don't know how to emulate page._viewport to be null in unit tests.

@lusarz you'll need to launch a browser with the `defaultViewport: null` option. We have a bunch of tests that launch a browser, check out `test/puppeteer.spec.js`.

test/page.spec.js
```diff
@@ -1653,6 +1654,15 @@ module.exports.addTests = function({testRunner, expect, headless}) {
       });
       expect(screenshot).toBeGolden('screenshot-grid-fullpage.png');
     });
+    it('should take fullPage screenshots when defaultViewport is null', async({server}) => {
+      const browser = await puppeteer.launch({ args: ['--no-sandbox'], defaultViewport: null });
```
lusarz: Would it be better to pass `defaultBrowserOptions` here:
https://github.com/GoogleChrome/puppeteer/blob/85aca8e1a56b5698f447ffd55b148482bf8db7ae/test/test.js#L157

and use it like:

    const browser = await puppeteer.launch(Object.assign({}, defaultBrowserOptions, { defaultViewport: null }));

?

test/page.spec.js
```diff
@@ -1653,6 +1654,15 @@ module.exports.addTests = function({testRunner, expect, headless}) {
...
+      const page = await browser.newPage();
+      await page.goto(server.PREFIX + '/grid.html');
+      const screenshot = await page.screenshot({
+        fullPage: true
+      });
+      expect(screenshot).toBeGolden('screenshot-grid-fullpage-2.png');
```
lusarz: Size of screenshot is different when `defaultViewport` is null - it's why I've created additional fixture. Is that okay ?

lusarz (commented):
@aslushnikov I have some doubts about tests I've created - I've put them as comments to specific lines

test/page.spec.js
```diff
@@ -1653,6 +1654,15 @@ module.exports.addTests = function({testRunner, expect, headless}) {
       });
       expect(screenshot).toBeGolden('screenshot-grid-fullpage.png');
     });
+    it('should take fullPage screenshots when defaultViewport is null', async({server}) => {
```
aslushnikov: Let's move this test to `puppeteer.spec.js`. Tests in `page.spec.js` are getting a pre-launched browser with a page; tests in `puppeteer.spec.js` only get pre-launched HTTP server. (if curious, you can see this setup done in `test/test.js`).

test/page.spec.js
aslushnikov: yep; you'll get defaultBrowesrOptions once you move the test to `puppeteer.spec.js`

test/page.spec.js
aslushnikov: Yes, perfect.

lusarz:
@aslushnikov ready

test/puppeteer.spec.js
```diff
@@ -279,6 +279,19 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions})
...
+        const options = Object.assign({}, defaultBrowserOptions, {
+          defaultViewport: null
+        });
+        const browser = await puppeteer.launch(options);
+        const page = await browser.newPage();
+        await page.goto(server.PREFIX + '/grid.html');
+        const screenshot = await page.screenshot({
+          fullPage: true
+        });
+        expect(screenshot).toBeGolden('screenshot-grid-fullpage-2.png');
```
aslushnikov: on a second though, why's screenshot size different here? I thought it shouldn't depend on viewport size at all.

lusarz: Yeah, I was surprised as well ...

lusarz: Original `screenshot-grid-fullpage.png` fixture has 500x1000 dimensions, probably because:
https://github.com/GoogleChrome/puppeteer/blob/5acf953104a09d7d688d76eadced6a4eba75d5d1/test/page.spec.js#L1649

When `viewport` is null, it has default width set to 800 - don't know why ...

aslushnikov: ah right, headless [defaults to 800x600](https://cs.chromium.org/chromium/src/headless/public/headless_browser.cc?type=cs&q=800+f:headless&sq=package:chromium&g=0&l=24).

However, this is vague on headful. Let's just make sure the screenshot doesn't hang here instead of comparing to golden; this might help us to avoid troubles in future when we start running tests in headful as well.

lusarz: Done

aslushnikov (approved):

Issue #3327, opened 2018-10-01
Puppeteer page.select got error: Cannot read property 'indexOf' of undefined


jennycai0807:
When I use page.select function, it got an error:
TypeError: Cannot read property 'indexOf' of undefined

My code is:
await page.select('#providerSelector', '101');

Then I use another way:
let selectElem = await page.waitForSelector("#providerSelector");
let optionElem = await page.waitForSelector("#providerSelector > option:nth-child(3)");
await page.evaluate((optionElem, selectElem) => {
optionElem.selected = true;
const event = new Event('change', {bubbles: true});
event.simulated=true;
selectElem.dispatchEvent(event);
}, optionElem, selectElem);

But still got the same error.
Is it a bug of Puppeteer?
Note: My Chromium Version 69.0.3452.0 (Developer Build) (64-bit)

If it's a puppeteer bug? How can I upgrade my puppeteer version?

lusarz:
Hello @jennycai0807, I cannot reproduce this issue. I've run this code:

    const browser = await puppeteer.launch();
    const page = await browser.newPage();
    page.setViewport({ width: 80, height: 40 });
    await page.setContent('<!doctype html><html><body><select id="providerSelector"><option value="99">99</option><option value="100">100</option><option value="101">101</option></select></body></html>');
    
    await page.select('#providerSelector', '101');
    await page.screenshot({path: 'select.png'});
    
    await browser.close();

on https://try-puppeteer.appspot.com/ and I got expected result:

![output](https://user-images.githubusercontent.com/10059264/46291679-66c54680-c58f-11e8-8468-edaa11307adb.png)

Could you provide html or link to page you're testing on ?

jennycai0807:
Here is the html:
```
<div>
<div class="form form-group" data-test-automation-id="selector">
<label class="control-label" for="Selector" data-test-automation-id="label">Current service provider</label>
<span class="select">
<select name="select" class="form-control" id="selector" data-test-automation-id="select">
<option value="">Please select</option>
<option value="100" data-test-automation-id="choose100">Test1</option>
<option value="101" data-test-automation-id="choose101">Test2</option>
<option value="102" data-test-automation-id="choose102">Test3</option>
</select>
</span>
</div>
</div>
```

The default value is "Please select", and the default value can't be selected. So when I tried page.type function, it didn't work.
Note: When I select other option from real UI, e.g. Test2, the 1st option "Please select" will add disabled=""

[...]

PR #3346, opened 2018-10-04
test: add failing test for page.select

aslushnikov:
References #3327.


JoelEinbinder (approved):
We should make page.select better.

PR #3365, opened 2018-10-08
feat(page): exposed additional info for ConsoleMessage object

### LINKED ISSUE #3029: Feature request: expose URL and line number for ConsoleMessage object
[LogEntry](https://chromedevtools.github.io/devtools-protocol/tot/Log#type-LogEntry) object provides url and lineNumber

```
url: string - URL of the resource if known.
lineNumber: integer - Line number in the resource.
```

can we also expose it?

Related source https://github.com/GoogleChrome/puppeteer/blob/25632133e2683c613c97d9088968c48e3c07a086/lib/Page.js#L167-L173

tompere:
This patch add additional method to ConsoleMessage object sent on console event. Currently, added URL and line number to additionalInfo.

Fixes #3029

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -178,6 +178,7 @@
   * [dialog.type()](#dialogtype)
 - [class: ConsoleMessage](#class-consolemessage)
   * [consoleMessage.args()](#consolemessageargs)
+  * [consoleMessage.location()](#consolemessagelocation)
   * [consoleMessage.text()](#consolemessagetext)
   * [consoleMessage.type()](#consolemessagetype)
 - [class: Frame](#class-frame)
@@ -2170,6 +2171,12 @@ puppeteer.launch().then(async browser => {
 
 #### consoleMessage.args()
 - returns: <[Array]<[JSHandle]>>
+
+#### consoleMessage.location()
+- returns: <[Object]>
+  - `url` <[string]> URL of the resource if known
+  - `lineNumber` <[number]> line number in the resource 
+  - `columnNumber` <[number]> line number in the resource 
 
 #### consoleMessage.text()
 - returns: <[string]>
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -495,10 +495,12 @@ class Page extends EventEmitter {
   async _onConsoleAPI(event) {
     const context = this._frameManager.executionContextById(event.executionContextId);
     const values = event.args.map(arg => createJSHandle(context, arg));
-    let location;
+    const location = {};
     if (event.stackTrace && event.stackTrace.callFrames) {
       const {url, lineNumber, columnNumber} = event.stackTrace.callFrames[0];
-      location = {url, lineNumber, columnNumber};
+      if (url) location.url = url;
+      if (lineNumber) location.lineNumber = lineNumber;
+      if (columnNumber) location.columnNumber = columnNumber;
     }
     this._addConsoleMessage(event.type, values, location);
   }
@@ -521,6 +523,7 @@ class Page extends EventEmitter {
   /**
    * @param {string} type
    * @param {!Array<!Puppeteer.JSHandle>} args
+   * @param {ConsoleMessage.Location} location
    */
   _addConsoleMessage(type, args, location) {
     if (!this.listenerCount(Page.Events.Console)) {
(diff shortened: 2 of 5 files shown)


lib/Page.js
```diff
@@ -179,11 +179,11 @@ class Page extends EventEmitter {
    * @param {!Protocol.Log.entryAddedPayload} event
    */
   _onLogEntryAdded(event) {
```
JoelEinbinder, line 181: This is not the only entry point for console messages. There is also Runtime.consoleAPICalled. It has line number and url contained in the stack trace.

docs/api.md
```diff
@@ -2168,6 +2169,11 @@ puppeteer.launch().then(async browser => {
 
 [ConsoleMessage] objects are dispatched by page via the ['console'](#event-console) event.
 
+#### consoleMessage.additionalInfo()
```
JoelEinbinder: Why not just put the url and line number on the console message itself?

tompere: you mean extending the api? same as `type` and `args`?

JoelEinbinder: yes

lib/Page.js
tompere: can you please elaborate?
I'm looking into [`Runtime.consoleAPICalled`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime#event-consoleAPICalled) and I don't see such details, nor a reference to [`ExceptionDetails`](https://chromedevtools.github.io/devtools-protocol/tot/Runtime#type-ExceptionDetails)

docs/api.md
tompere: as I commented in the [origin issue](https://github.com/GoogleChrome/puppeteer/issues/3029#issuecomment-427738035):
> This is under the assumption that additionalInfo is a subset of LogEntry optional params, thus future requests for additional data can easily added

makes sense?

tompere: also the fact the lineNumber and url are optionals, made me feel less comfortable in adding a dedicated and explicit API (as it will return `undefined` many times...)

lib/Page.js
JoelEinbinder: It is contained in the callframes of the stacktrace: https://chromedevtools.github.io/devtools-protocol/tot/Runtime#type-CallFrame

docs/api.md
JoelEinbinder: Having `additionalInfo` raises some difficult questions. Which values are considered core, and which are considered additional? Can additional info be null, or will it be an empty object? Even if we have reasoning, its not obvious to a user of the API. They would have to check the documentation every time.

That being said, it might make sense to group line, column, and url under a `location` object if they are often all null or all present.

lib/Page.js
tompere: done

docs/api.md
tompere: done. `location` it is.

lib/Page.js
```diff
@@ -495,7 +495,12 @@ class Page extends EventEmitter {
   async _onConsoleAPI(event) {
     const context = this._frameManager.executionContextById(event.executionContextId);
     const values = event.args.map(arg => createJSHandle(context, arg));
-    this._addConsoleMessage(event.type, values);
+    let location;
+    if (event.stackTrace && event.stackTrace.callFrames) {
+      const {url, lineNumber, columnNumber} = event.stackTrace.callFrames[0];
+      location = {url, lineNumber, columnNumber};
```
JoelEinbinder: nit: don't unconditionally add the properties if they might be undefined. Otherwise a user doing `"url" in location` will be misled.

lib/Page.js
```diff
@@ -517,7 +522,7 @@ class Page extends EventEmitter {
    * @param {string} type
    * @param {!Array<!Puppeteer.JSHandle>} args
    */
-  _addConsoleMessage(type, args) {
+  _addConsoleMessage(type, args, location) {
```
JoelEinbinder, line 528: jsdoc

lib/Page.js
tompere: just a clarification about how defensive puppeteer code should be:

Can i rely on devtools protocol API for optional/mandatory return values?
e.g.
* [Runtime#type-CallFrame](https://chromedevtools.github.io/devtools-protocol/tot/Runtime#type-CallFrame) always return `{url, lineNumber, columnNumber}`.
* [Runtime#event-consoleAPICalled](https://chromedevtools.github.io/devtools-protocol/tot/Runtime#event-consoleAPICalled) might not return `StackTrace`. But if returned, it will always return `Array{CallFrame}`.

JoelEinbinder: You can rely on the values that the protocol will provide, but there might be extra values alongside those values. So you should build your own object instead of serving the raw protocol object.

lib/Page.js
tompere: done

lib/Page.js
tompere: done

aslushnikov (approved):
In it goes!

tompere:
👏

PR #3376, opened 2018-10-08
docs(examples): render custom HTML to PDF

guumaster:
add example to create an HTML page in memory, render it with puppeteer and save it as a pdf file.

### DIFF
--- a/examples/custom-html-pdf.js
+++ b/examples/custom-html-pdf.js
@@ -1,0 +1,57 @@
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+/**
+ * Render your custom HTML and save it to PDF file
+ */
+
+'use strict';
+
+const puppeteer = require('puppeteer');
+
+const YOUR_HTML = `<!doctype html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Example HTML</title>
+  <!-- you can link any local/external file -->
+  <link rel="stylesheet" href="css/styles.css">
+</head>
+<body>
+  <!-- also you can run any script here -->
+  <script src="js/scripts.js"></script>
+  <article>
+    <h1>Google Chrome</h1>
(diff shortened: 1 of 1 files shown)


guumaster:
`I signed it!`

examples/custom-html-pdf.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
```
JoelEinbinder: 2018

examples/custom-html-pdf.js
```diff
@@ -0,0 +1,57 @@
...
+
+const puppeteer = require('puppeteer');
+
+const YOUR_HTML = `<!doctype html>
+<html>
+<head>
+  <meta charset="utf-8">
+  <title>Example HTML</title>
+  <!-- you can link any local/external file -->
+  <link rel="stylesheet" href="css/styles.css">
```
JoelEinbinder, line 32: You can't link to an external file if using a data url.

examples/custom-html-pdf.js
```diff
@@ -0,0 +1,57 @@
...
+</body>
+</html>
+`;
+
+(async() => {
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+
+  // set your content and wait until render is finished
+  await page.goto(`data:text/html,${YOUR_HTML}`, { waitUntil: 'networkidle2' });
```
JoelEinbinder: why wait for networkidle2?

JoelEinbinder (commented):
Can you elaborate on a scenario where this would be useful for someone?

examples/custom-html-pdf.js
guumaster: I have a working example that loads JS and CSS and Fonts from the network based on this code.

examples/custom-html-pdf.js
guumaster: updated.

examples/custom-html-pdf.js
guumaster: updated to `networkidle0`. I've read the docs wrong.

guumaster:
> Can you elaborate on a scenario where this would be useful for someone?

If you have a server that renders HTML and you want to generate pdf from any page, you can use this code to do it instead of navigating to an internal page with `goto()`.

I've tried with `setContent()/content()` but even if you use `await page.content()` it wont show immediately. So after a while of going through the issues and docs I got this example working.

I thought it would be useful for somebody else.

Thanks for reviewing it!

aslushnikov:
@guumaster The "goto hack" is a consequence of #728; I'd rather have it fixed than spreading the hack further.

#728 requires contributing upstream to Chromium and DevTools protocol; let me know if you want to give it a shot.

guumaster:
Probably I can't contribute to #728 to Chromium.

This example may not be too important. But is not such a hack either, it is something that you can normally do with a browser, and `puppeteer` is meant to control a browser, isn't it?  Maybe someone that needs this can find the code in this issue useful.

Thanks for your time anyway.

PS: You think you can merge it if I get the example working with `setContent()/content()` ?

PR #3386, opened 2018-10-10
DOC():Put DEV Parameter in install commands

alvarocjunq:
Normally, Puppeteer is used for development =P

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,8 @@ Give it a spin: https://try-puppeteer.appspot.com/
 To use Puppeteer in your project, run:
 
 ```bash
-npm i puppeteer
-# or "yarn add puppeteer"
+npm i puppeteer --save-dev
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -33,8 +33,8 @@ Give it a spin: https://try-puppeteer.appspot.com/
 To use Puppeteer in your project, run:
 
 ```bash
-npm i puppeteer
-# or "yarn add puppeteer"
+npm i puppeteer --save-dev
+# or "yarn add puppeteer" --dev
```
ntzm, line 37: should be inside the quotes

JoelEinbinder:
If someone knows what `--dev` does, they can decide whether or not to use it themselves. If they don't, this could lead to problems for them. Puppeteer isn't exclusively used as a dev dependency.

PR #3409, opened 2018-10-14
feat(Websocket): isOpen func add

maddhruv:


JoelEinbinder:
`WebSocketTransport` is not part of Puppeteer's public API. There is no reason to add unused methods to it.

PR #3430, opened 2018-10-19
fix(launcher): ensureInitialPage awaits forever

rbochenski:
There might be a situation when event was sent before listener attached and initialPageCallback is never called.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -183,32 +183,32 @@ class Launcher {
       if (browser.targets().find(target => target.type() === 'page'))
         return;
 
-        let initialPageCallback;
-        let initialCallbackDone = false;
+      let initialPageCallback;
+      let initialCallbackDone = false;
 
-        const initialPagePromise = new Promise(resolve => initialPageCallback = resolve);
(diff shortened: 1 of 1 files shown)


JoelEinbinder:
I want to fix this bug on the Chromium side rather than introducing a timeout.

rbochenski:
> I want to fix this bug on the Chromium side rather than introducing a timeout.

I would be more than happy to see problem solved on Chromium. How long it will take?
If few weeks/months it might be good to use this PR in the meantime.

aslushnikov:
> There might be a situation when event was sent before listener attached and initialPageCallback is never called.

@rbochenski Can you please clarify how's this possible? Is there any repro I can try locally to see this happenning?

rbochenski:
> > There might be a situation when event was sent before listener attached and initialPageCallback is never called.
> 
> @rbochenski Can you please clarify how's this possible? Is there any repro I can try locally to see this happenning?

I will prepare something for you.

aslushnikov:
@rbochenski I'll close this for now; feel free to send another PR once you're ready.

PR #3435, opened 2018-10-21
fix(page.exportfunction): remove the name clash, avoid internal chrom…

pavelfeldman:
Avoid the race condition between Page.addScriptToEvaluate and Runtime.addBinding.

This reproduces reliably for me, but I could not implement the test.

### DIFF
--- a/test/page.spec.js
+++ b/test/page.spec.js
@@ -1051,6 +1051,18 @@ module.exports.addTests = function({testRunner, expect, headless}) {
       });
       expect(result).toBe(15);
     });
+    it('multiple bindings should work', async({page, server}) => {
+      await page.exposeFunction('multiply', function(a, b) {
+        return a * b;
+      });
+      await page.exposeFunction('divide', function(a, b) {
+        return a / b;
+      });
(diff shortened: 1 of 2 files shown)


test/page.spec.js
```diff
@@ -1051,6 +1051,18 @@ module.exports.addTests = function({testRunner, expect, headless}) {
...
     });
+    it('multiple bindings should work', async({page, server}) => {
+      await page.exposeFunction('multiply', function(a, b) {
+        return a * b;
+      });
+      await page.exposeFunction('divide', function(a, b) {
+        return a / b;
+      });
+      const result = await page.evaluate(async function() {
+        return await divide(await multiply(5, 3), 5);
```
JoelEinbinder, line 1062: If `divide` and `multiply` were mixed up, this would still return 3.

JoelEinbinder:
I can't repro this unless I don't await the exposeFunction calls.

JoelEinbinder:
The test is good to have, but I don't believe that there is a bug in Puppeteer. `addScriptToEvaluateOnNewDocument` should always fire after `addBinding`.

PR #3449, opened 2018-10-26
fix: support configurations in package.json

jxltom:
Fixes https://github.com/GoogleChrome/puppeteer/issues/2262#issuecomment-378122692

This PR support set ```puppeteer_skip_chromium_download```, ```puppeteer_download_host``` and ```puppeteer_chromium_revision``` in ```config``` of ```package.json```


jxltom:
I signed it!

aslushnikov:
@jxltom so now when we have [puppeteer-core](https://www.npmjs.com/package/puppeteer-core) - why would you want to have the config support?

jxltom:
Hi @aslushnikov It is because I'm not using puppeteer directly. I'm using storybook's addon snapshots which requires puppeteer rather than puppeteer-core. But I don't need chromium, so it would be nice to disable downnloading chromium in package.json.

aslushnikov (approved):

PR #3480, opened 2018-11-01
fix(page.evaluate): better function serialization

### LINKED ISSUE #1665: async function inside evaluate fails
#### Code
```
    console.log('1')
})
```
#### Error
```
(node:66796) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Evaluation failed: ReferenceError: fn is not defined
    at <anonymous>:1:26
```

aslushnikov:
References #1665

### DIFF
--- a/lib/ExecutionContext.js
+++ b/lib/ExecutionContext.js
@@ -94,8 +94,23 @@ class ExecutionContext {
     if (typeof pageFunction !== 'function')
       throw new Error('The following is not a function: ' + pageFunction);
 
+    let functionText = pageFunction.toString();
+    try {
+      eval(functionText);
+    } catch (e1) {
+      // This means we have a function shorthand. Try another
+      // time prefixing 'function '.
+      functionText = 'function ' + functionText;
+      try {
+        eval(functionText);
+      } catch (e2) {
+        // We tried hard to serialize, but there's a weird beast here.
+        throw new Error('Passed function is not well-serializable!');
+      }
+    }
+
     const { exceptionDetails, result: remoteObject } = await this._client.send('Runtime.callFunctionOn', {
-      functionDeclaration: pageFunction.toString() + '\n' + suffix + '\n',
+      functionDeclaration: functionText + '\n' + suffix + '\n',
       executionContextId: this._contextId,
       arguments: args.map(convertArgument.bind(this)),
       returnByValue: false,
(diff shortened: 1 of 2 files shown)


lib/ExecutionContext.js
```diff
@@ -94,8 +94,23 @@ class ExecutionContext {
     if (typeof pageFunction !== 'function')
       throw new Error('The following is not a function: ' + pageFunction);
 
+    let functionText = pageFunction.toString();
+    try {
+      eval(functionText);
+    } catch (e1) {
+      // This means we have a function shorthand. Try another
+      // time prefixing 'function '.
+      functionText = 'function ' + functionText;
```
JoelEinbinder: What about async static generators?

lib/ExecutionContext.js
```diff
@@ -94,8 +94,23 @@ class ExecutionContext {
     if (typeof pageFunction !== 'function')
       throw new Error('The following is not a function: ' + pageFunction);
 
+    let functionText = pageFunction.toString();
+    try {
+      eval(functionText);
```
JoelEinbinder: I'd be worried about unintended side effects here. Consider: `function functionText(){return 5}`.

JoelEinbinder:
Can we also add a note about `toString`ing the pageFunctions? This has come up many times before.

lib/ExecutionContext.js
aslushnikov: async done; the rest is too exotic for now!

lib/ExecutionContext.js
aslushnikov: good catch; switched to `new Function` instead

lib/ExecutionContext.js
```diff
@@ -94,8 +94,26 @@ class ExecutionContext {
     if (typeof pageFunction !== 'function')
       throw new Error('The following is not a function: ' + pageFunction);
 
+    let functionText = pageFunction.toString();
+    try {
+      new Function('(' + functionText + ')');
+    } catch (e1) {
+      // This means we have a function shorthand. Try another
```
JoelEinbinder:
```suggestion
      // This means we might have a function shorthand. Try another
```

JoelEinbinder (approved):

PR #3492, opened 2018-11-03
test(cookies): Remove duplicated test

kblok:
We have the same test on line 130 https://github.com/GoogleChrome/puppeteer/blob/v1.10.0/test/cookies.spec.js#L130


aslushnikov (approved):
Thanks!

PR #3509, opened 2018-11-07
fix: Inclusion of target platform configuration on installation script

ralphg6:
Included was the ability to inform with the destination platform where the process will be deployed.

For cases where the build process of package to be deployed runs on a different platform from which it will be run.


ralphg6:
Fixes #3510

satanninja:
( Overlay - values / successful interation / proceed to end biometric / sustain values. )

aslushnikov:
Closing; see discussion in https://github.com/GoogleChrome/puppeteer/issues/3510.

PR #3511, opened 2018-11-08
chore: add utils/bisect.js to bisect chromium revisions

aslushnikov:
This patch adds a new utility - `utils/bisect.js` - that accepts
a range of Chromium revisions and a pptr script and bisects the
range to figure when the script breaks.

The Puppeteer Script, given to the tool, should be exiting
with non-zero code to signify malfunctioning.

Example:
```
$ node utils/bisect.js --good 577361 --bad 599821 a.js
```

### DIFF
--- a/utils/bisect.js
+++ b/utils/bisect.js
@@ -1,0 +1,194 @@
+#!/usr/bin/env node
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


aslushnikov:
Screenshot of me bisecting https://github.com/GoogleChrome/puppeteer/issues/3463: 
image

Where the `a.js` is:
```js
const puppeteer = require('.');

(async() => {

  let success = true;
    const url = 'data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU5ErkJggg=='
    const response = await window.fetch(url);
    const blob = await response.blob();
    const blobURL = URL.createObjectURL(blob);
    // Throws here:
    const response2 = await window.fetch(blobURL);
  }).catch(e => success = false);
  await browser.close();
  process.exit(success ? 0 : 1);
})();
```

JoelEinbinder (approved):
Add it to one of the contributing docs?

aslushnikov:
@JoelEinbinder let me add it later on - turns out I need bisecting again

PR #3524, opened 2018-11-09
refactor: move to flatten protocol

aslushnikov:
DevTools protocol is dropping nested targets and switching to
flatten protocol. This patch adopts the new scheme.

Once this change lands, tip-of-tree Puppeteer will be incompatible
with Chromium below 72.0.3606.0. We'll need to adopt multiple pptr versions approach for [carlo](https://github.com/GoogleChromeLabs/carlo) to make it work nicely.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -94,13 +94,11 @@ class Connection extends EventEmitter {
       const sessionId = object.params.sessionId;
       const session = new CDPSession(this, object.params.targetInfo.type, sessionId);
       this._sessions.set(sessionId, session);
-      this.emit(Connection.Events.SessionAttached, session);
     } else if (object.method === 'Target.detachedFromTarget') {
       const session = this._sessions.get(object.params.sessionId);
       if (session) {
         session._onClosed();
         this._sessions.delete(object.params.sessionId);
-        this.emit(Connection.Events.SessionDetached, session);
       }
     }
     if (object.sessionId) {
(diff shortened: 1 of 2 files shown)


lib/Connection.js
```diff
@@ -65,15 +68,20 @@ class Connection extends EventEmitter {
...
-    const message = JSON.stringify({id, method, params});
-    debugProtocol('SEND ► ' + message);
-    this._transport.send(message);
+    const id = this._rawSend({method, params});
     return new Promise((resolve, reject) => {
       this._callbacks.set(id, {resolve, reject, error: new Error(), method});
     });
   }
 
+  _rawSend(message) {
```
JoelEinbinder, line 77: jsdoc.

lib/Connection.js
```diff
@@ -65,15 +68,20 @@ class Connection extends EventEmitter {
...
+    const id = ++this._lastId;
+    message = JSON.stringify(Object.assign({}, message, {id}));
```
JoelEinbinder, line 79: Can we not re-use the message variable here? It becomes a bit tricky to follow.

JoelEinbinder (approved):

JoelEinbinder:
I still don't think we should land this until flat hits stable.

JoelEinbinder (changes requested):
undoing my +1

aslushnikov:
We'll come back to this with 2.0 release.

PR #3525, opened 2018-11-09
fix(types): fix parameters where they disagree with the api

JoelEinbinder:


aslushnikov (approved):

docs/api.md
```diff
@@ -2966,10 +2967,18 @@ This method returns the bounding box of the element (relative to the main frame)
 
 #### elementHandle.boxModel()
 - returns: <[Promise]<?[Object]>>
-  - content <[Array]<[Object]>> Content box, represented as an array of {x, y} points.
-  - padding <[Array]<[Object]>> Padding box, represented as an array of {x, y} points.
-  - border <[Array]<[Object]>> Border box, represented as an array of {x, y} points.
-  - margin <[Array]<[Object]>> Margin box, represented as an array of {x, y} points.
+  - content <[Array]<[Object]>> Content box
```
vsemozhetbyt: Nit: maybe it is worth to unify periods in: `Content box` etc vs `Element's width.`  / `Element's height.`

PR #3527, opened 2018-11-09
test: fix workers test

aslushnikov:


test/worker.spec.js
```diff
@@ -15,7 +15,7 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(await worker.evaluate(() => self.workerFunction())).toBe('worker function result');
 
       await page.goto(server.EMPTY_PAGE);
-      expect(page.workers()).toEqual([]);
+      expect(page.workers().length).toBe(1);
```
JoelEinbinder: This doesn't make sense to me. Isn't it the opposite of what the test is testing for? EMPTY_PAGE shouldn't have any workers.

aslushnikov: Ah, I meant 0.

test/worker.spec.js
```diff
@@ -15,7 +15,7 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(await worker.evaluate(() => self.workerFunction())).toBe('worker function result');
 
       await page.goto(server.EMPTY_PAGE);
-      expect(page.workers()).toEqual([]);
+      expect(page.workers().length).toBe(0);
```
JoelEinbinder, line 18: I don't get it. Whats the change?

aslushnikov: when this test fails, the output is cryptic since Worker object is not nicely serializable.

JoelEinbinder: Fair

JoelEinbinder (approved):

PR #3530, opened 2018-11-12
test: setup sandbox on linux

aslushnikov:


test/test.js
```diff
@@ -47,7 +47,7 @@ const defaultBrowserOptions = {
   slowMo,
   headless,
   dumpio: (process.env.DUMPIO || 'false').trim().toLowerCase() === 'true',
-  args: ['--no-sandbox']
+  args: []
```
JoelEinbinder: nit: can just drop this line.

JoelEinbinder (approved):

JoelEinbinder:
woo!

test/test.js

PR #3536, opened 2018-11-13
chore(types): lint using private properties

JoelEinbinder:
This adds a new linting step that forbids using variables prefixed by `_` outside the file they were declared in.

Sometimes we used `_` not to mean file-private, but to mean not part of our public api. Some of these  usages became static methods on the class, and some became prefixed with `$_` to mean internal. Most current usages of `$_` are a bit ugly, and I can refactor them away in follow up patches.


JoelEinbinder:
@brendankenny

Bambam71 (approved):

aslushnikov:
Ok closing this for now - we'll get this back if/when we need it again

PR #3539, opened 2018-11-13
test(accessibility): Fix duplicated test

kblok:
[plain text field without role should not have content](https://github.com/GoogleChrome/puppeteer/blob/v1.10.0/test/accessibility.spec.js#L150) and [plain text field with tabindex and without role should not have content](https://github.com/GoogleChrome/puppeteer/blob/v1.10.0/test/accessibility.spec.js#L150) are duplicated.

By looking at the rest of the tests I bet [plain text field with tabindex and without role should not have content](https://github.com/GoogleChrome/puppeteer/blob/v1.10.0/test/accessibility.spec.js#L150) needs a `tabIndex` in the HTML

### DIFF
--- a/test/accessibility.spec.js
+++ b/test/accessibility.spec.js
@@ -158,7 +158,7 @@ module.exports.addTests = function({testRunner, expect}) {
       });
       it('plain text field with tabindex and without role should not have content', async function({page}) {
         await page.setContent(`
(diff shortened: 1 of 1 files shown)


aslushnikov:
@JoelEinbinder, can you please take a look?

aslushnikov (approved):
Joel Approves.

PR #3550, opened 2018-11-15
fix(page): dispatch errors into page

### LINKED ISSUE #3549: Functions bound with page.exposeFunction() produce unhandled promise rejections
Hello everyone,

this may or may not be an issue, we're using `page.exposeFunction()` to expose certain APIs to the browser context and from our observations, any error thrown within the APIs turns into an unhandled promise rejection. This is due to the implementation of the `.onBindingCalled()` handler in `Page`. 

See https://github.com/GoogleChrome/puppeteer/blob/c185eeef6119141ebc07f07d46ae0dcb16a555f3/lib/Page.js#L514

From my observation, exceptions thrown within the `_pageBindings` function call are not handled in any way and since the `.onBindingCalled()` function is used as a callback, the resulting rejected promises cannot be handled.

Is this a bug or intended behavior?

### Steps to reproduce

* Puppeteer version: 1.9.0
* Platform / OS version: MacOS 10.14 locally; node:8-slim (Debian Jessie) remotely;
* Node.js version: 8.12.0

The very minimal reproduction code looks like this:
```js

puppeteer.launch()
.then(async browser => {
  await page.exposeFunction('func', async () => {
    throw new Error('err');
  });
    const content = await window.func();
  });
  await browser.close();
}).catch(err => console.log('does not get called'));
```
The same issue would also manifest in the `page.exposeFunction()` examples, where a rejection of the `readFile` `Promise` would produce an unhandled rejection.
```js
const fs = require('fs');

  page.on('console', msg => console.log(msg.text()));
  await page.exposeFunction('readfile', async filePath => {
      fs.readFile(filePath, 'utf8', (err, text) => {
        if (err)
          reject(err);
        else
          resolve(text);
      });
    });
  });
    // use window.readfile to read contents of a file
    const content = await window.readfile('/etc/hosts');
    console.log(content);
  });
  await browser.close();
});
```

I would expect the exceptions thrown from the bound functions to be manageable in some way. Either by allowing a handler to be attached to the `page.exposeFunction()` call or by backpropagating the errors back to the Browser context so that the `Promise` there gets rejected with  a representation of the exception thrown in Node context.

We implemented those two options manually in our code.

**What happens instead?**
An unhandled promise rejection is produced.

Thank you for your time.

aslushnikov:
Errors thrown on the node side of the `page.exposeFunction` callback
should be dispatched into the page.

Fixes #3549

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -428,7 +428,7 @@ class Page extends EventEmitter {
         }
         const seq = (me['lastSeq'] || 0) + 1;
         me['lastSeq'] = seq;
-        const promise = new Promise(fulfill => callbacks.set(seq, fulfill));
+        const promise = new Promise((resolve, reject) => callbacks.set(seq, {resolve, reject}));
         binding(JSON.stringify({name: bindingName, seq, args}));
         return promise;
       };
@@ -511,12 +511,29 @@ class Page extends EventEmitter {
    */
   async _onBindingCalled(event) {
     const {name, seq, args} = JSON.parse(event.payload);
-    const result = await this._pageBindings.get(name)(...args);
-    const expression = helper.evaluationString(deliverResult, name, seq, result);
+    let result = null;
+    let error = null;
+    try {
+      result = await this._pageBindings.get(name)(...args);
+    } catch (e) {
+      error = e;
+    }
+    let expression = null;
+    if (error)
+      expression = helper.evaluationString(deliverError, name, seq, error.message, error.stack)
+    else
+      expression = helper.evaluationString(deliverResult, name, seq, result);
     this._client.send('Runtime.evaluate', { expression, contextId: event.executionContextId }).catch(debugError);
 
     function deliverResult(name, seq, result) {
-      window[name]['callbacks'].get(seq)(result);
+      window[name]['callbacks'].get(seq).resolve(result);
+      window[name]['callbacks'].delete(seq);
+    }
+
+    function deliverError(name, seq, message, stack) {
+      let error = new Error(message);
+      error.stack = stack;
+      window[name]['callbacks'].get(seq).reject(error);
       window[name]['callbacks'].delete(seq);
     }
   }
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -511,18 +511,16 @@ class Page extends EventEmitter {
    */
   async _onBindingCalled(event) {
     const {name, seq, args} = JSON.parse(event.payload);
-    let result = null;
-    let error = null;
-    try {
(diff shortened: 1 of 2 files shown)


lib/Page.js
```diff
@@ -511,12 +511,29 @@ class Page extends EventEmitter {
...
     this._client.send('Runtime.evaluate', { expression, contextId: event.executionContextId }).catch(debugError);
 
     function deliverResult(name, seq, result) {
-      window[name]['callbacks'].get(seq)(result);
+      window[name]['callbacks'].get(seq).resolve(result);
+      window[name]['callbacks'].delete(seq);
+    }
+
+    function deliverError(name, seq, message, stack) {
+      let error = new Error(message);
```
JoelEinbinder: Const

lib/Page.js
```diff
@@ -511,12 +511,29 @@ class Page extends EventEmitter {
...
-    const expression = helper.evaluationString(deliverResult, name, seq, result);
+    let result = null;
+    let error = null;
+    try {
+      result = await this._pageBindings.get(name)(...args);
+    } catch (e) {
+      error = e;
+    }
+    let expression = null;
+    if (error)
```
JoelEinbinder: Things can throw null. Instead of checking if there is an error, do the logic in the catch.

lib/Page.js
```diff
@@ -511,12 +511,32 @@ class Page extends EventEmitter {
...
+    }
     this._client.send('Runtime.evaluate', { expression, contextId: event.executionContextId }).catch(debugError);
 
     function deliverResult(name, seq, result) {
-      window[name]['callbacks'].get(seq)(result);
+      window[name]['callbacks'].get(seq).resolve(result);
+      window[name]['callbacks'].delete(seq);
+    }
+
+    function deliverError(name, seq, message, stack) {
```
JoelEinbinder, line 542: nit: JSDoc

lib/Page.js

JoelEinbinder (approved):

lib/Page.js
aslushnikov: Fair. Creepy, but fair.

JoelEinbinder:
oh, we should probably pass the name as well. For TimeoutErrors

lib/Page.js
```diff
@@ -428,7 +428,7 @@ class Page extends EventEmitter {
         }
         const seq = (me['lastSeq'] || 0) + 1;
         me['lastSeq'] = seq;
-        const promise = new Promise(fulfill => callbacks.set(seq, fulfill));
```
satanninja, line 431: Ok

PR #3557, opened 2018-11-16
feat(page): support wait options for `page.setContent`

### LINKED ISSUE #728: Page.setContent should wait for resources to be loaded

aslushnikov:
This patch teaches `page.setContent` to await resources in
the new document.

**NOTE**: This patch changes behavior: currently, `page.setContent`
awaits the `"domcontentloaded"` event; with this patch, we can now await
other lifecycle events, and switched default to the `"load"` event.

The change is justified since current behavior made `page.setContent`
unusable for its main designated usecases, pushing our client
to use [dataURL workaround](https://github.com/GoogleChrome/puppeteer/issues/728#issuecomment-334301491).

Fixes #728

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -69,12 +69,14 @@ class FrameManager extends EventEmitter {
    * @return {!Promise<?Puppeteer.Response>}
    */
   async navigateFrame(frame, url, options = {}) {
+    assertNoLegacyNavigationOptions(options);
     const {
       referer = this._networkManager.extraHTTPHeaders()['referer'],
+      waitUntil = ['load'],
       timeout = this._defaultNavigationTimeout,
     } = options;
 
-    const watcher = new NavigatorWatcher(this._client, this, this._networkManager, frame, timeout, options);
+    const watcher = new LifecycleWatcher(this, frame, waitUntil, timeout);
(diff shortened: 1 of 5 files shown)


lib/FrameManager.js
```diff
@@ -525,13 +529,28 @@ class Frame {
...
    */
-  async setContent(html) {
+  async setContent(html, options = {}) {
+    const {
+      waitUntil = ['load'],
+      timeout = 30000,
+    } = options;
+    // We rely upon the fact that document.open() will reset frame lifecycle with "init"
+    // lifecycle event. @see https://crrev.com/608658
     await this.evaluate(html => {
```
JoelEinbinder, line 541: By the time this evaluate has finished, couldn't we have received all of the life cycle events? Then the lifecyclePromise will timeout?

aslushnikov: ah, I see what you mean. Done.

JoelEinbinder (approved):

PR #3558, opened 2018-11-16
feat(launcher): add browserUrl option to puppeteer.connect

### LINKED ISSUE #3537: Feature proposal: Ability to pass browserUrl to puppeteer.connect
### Description
Add support for new `browserUrl` property in `options` object passed to `puppeteer.connect(options)`. It would accept url in form of `http://{host}:{port}` which should point to the existing Chromium instance. For example:

```
puppeteer.connect({
    browserUrl: http://localhost:9222
})
```

If this option is passed to `puppeteer.connect`, then  `browserWSEndpoint` is retrieved from `http://{host}:{port}/json/version` and passed `new Connection()`. The rest of the flow remains unchanged.

If you're OK with the proposal I could write up the code and the tests

### Background

Currently to attach Puppeteer to an existing Chrome instance, one must provide  url of browser websocket endpoint. For example:

```
puppeteer.connect({
    browserWSEndpoint: ws://localhost:9222/devtools/browser/b0b8a4fb-bb17-4359-9533-a8d9f3908bd8
})
```

However, as many issues raised here show (#2790, #1645, #940) `browserWSEndpoint` is often not known to the code calling `puppeteer.connect`, for example because Chrome was lunched in a docker container. 

In such a case, a common and repeating scenario is to:
1. make HTTP request to `http://{host}:{port}/json/version`
2. parse the response JSON
3. get `browserWSEndpoint` from and use it in `puppeteer.connect`

The steps above could be done by puppeteer resulting in less code in the user-land.

lukaszfiszer:
Fixes #3537

I will add documentation if the code is ok.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -441,7 +441,8 @@ puppeteer.launch().then(async browser => {
 
 #### puppeteer.connect(options)
 - `options` <[Object]>
-  - `browserWSEndpoint` <[string]> a [browser websocket endpoint](#browserwsendpoint) to connect to.
+  - `browserWSEndpoint` <?[string]> a [browser websocket endpoint](#browserwsendpoint) to connect to.
+  - `browserUrl` <?[string]> a browser url to connect to, in format `http://${host}:${port}`. Use interchangeably with `browserWSEndpoint` to let Puppeteer fetch it from [metadata endpoint](https://chromedevtools.github.io/devtools-protocol/#how-do-i-access-the-browser-target).
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
   - `defaultViewport` <?[Object]> Sets a consistent viewport for each page. Defaults to an 800x600 viewport. `null` disables the default viewport.
     - `width` <[number]> page width in pixels.
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -15,6 +15,8 @@
  */
 const os = require('os');
 const path = require('path');
+const http = require('http');
+const URL = require('url');
 const removeFolder = require('rimraf');
 const childProcess = require('child_process');
 const BrowserFetcher = require('./BrowserFetcher');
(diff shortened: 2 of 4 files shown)


lib/Puppeteer.js
```diff
@@ -37,7 +37,7 @@ module.exports = class {
   }
 
   /**
-   * @param {!(Launcher.BrowserOptions & {browserWSEndpoint: string, transport?: !Puppeteer.ConnectionTransport})} options
+   * @param {!(Launcher.BrowserOptions & {browserWSEndpoint: string, browserUrl: string, transport?: !Puppeteer.ConnectionTransport})} options
```
JoelEinbinder: `browserUrl?:`

The question mark makes it optional.

lib/Launcher.js
```diff
@@ -373,6 +390,48 @@ function waitForWSEndpoint(chromeProcess, timeout, preferredRevision) {
...
+  const endpointUrl = URL.resolve(browserUrl, '/json/version');
+  const requestOptions = Object.assign(URL.parse(endpointUrl), { method: 'GET' });
+  const promise = new Promise((res, rej) => { resolve = res; reject = rej; });
+
+  function onGetWsEndpointDone(data) {
+    try {
+      const {webSocketDebuggerUrl} = JSON.parse(data);
+      resolve(webSocketDebuggerUrl);
+    } catch (e) {
+      onGetWsEndpointError(e);
```
JoelEinbinder: I don't think we need special error forwarding here.

lib/Puppeteer.js
lukaszfiszer: Should we make both `browserWSEndpoint` and `browserUrl` optional then? You should pass one or another, but not both.

lib/Launcher.js
lukaszfiszer: This is to display the same user-friendly error message both in case of network failure and JSON parsing failure. Example:

```
Error: Failed to fetch browser webSocket url from http://127.0.0.1:21222/: SyntaxError: Unexpected token < in JSON at position 0
    at onGetWsEndpointError (/Users/lf/Projects/puppeteer/lib/Launcher.js:413:12)
    at onGetWsEndpointDone (/Users/lf/Projects/puppeteer/lib/Launcher.js:408:7)
    at IncomingMessage.res.on (/Users/lf/Projects/puppeteer/lib/Launcher.js:426:25)
    at emitNone (events.js:111:20)
    at IncomingMessage.emit (events.js:208:7)
    at endReadableNT (_stream_readable.js:1064:12)
    at _combinedTickCallback (internal/process/next_tick.js:139:11)
    at process._tickCallback (internal/process/next_tick.js:181:9)
```

lukaszfiszer:
Added docs, made `browserWSEndpoint` and `browserUrl` optional, refactor event handler function names in `getWSEndpoint`. @JoelEinbinder @aslushnikov could you take a look?

docs/api.md
```diff
@@ -3533,4 +3534,4 @@ TimeoutError is emitted whenever certain operations are terminated due to timeou
 [Worker]: #class-worker "Worker"
 [Accessibility]: #class-accessibility "Accessibility"
 [AXNode]: #accessibilitysnapshotoptions "AXNode"
-[ConnectionTransport]: ../lib/WebSocketTransport.js "ConnectionTransport"
\ No newline at end of file
+[ConnectionTransport]: ../lib/WebSocketTransport.js "ConnectionTransport"
```
lukaszfiszer, line 3537: a final new line was added at the end of this file, conforming with `.editorconfig`

aslushnikov (approved):
In it goes!

Issue #1995, opened 2018-02-08
Issues with Parallel pages html content download


kushaljshah:
It does not seem that there is parallel download happening when I am downloading a batch of urls. The url downloading is  completing way late than it does when download individually. 

I am using a 2 core ec2-machine with 4gb RAM. All system resources are pretty stable in a batch size of 30 urls. The connection time out used here is 30 seconds. Their are urls which complete as late as 15 s, 20 secs, 23 secs which complete in about 3-7 seconds when run individually. Is not parallel IO supposed to happen here. What am I missing?

```
let startDate new Date();
let finishDate;
const puppeteer = require('puppeteer');

await puppeteer.launch().then(async browser => {
  const promises=[];
  urlArr.forEach(function(url){
  promises.push(browser.newPage().then(async page => {
          try{
               await page.goto(url.startsWith('http') ? url : `http://${url}`, {
                 waitUntil: ['load'],timeout:connectionTimeOut
               });

              const html = await page.content();
              finishDate = new Date();
              console.log("html download complete for " + url + " - " + finishDate.getTime() + " - " + (finishDate.getTime() - startDate.getTime())/1000);
            }catch(err){
              console.log(err)
            }
       }))
  });
  await Promise.all(promises)
  await browser.close();
})
```

nammaianh:
How about using cluster?

kushal-tracxn:
@nammaianh The issue I am referring to is about how much parallel operations can be done on a single node. Irrespective of, if it is part of a cluster or not.

Garbee:
Standard networking applies and the browser needs to delegate and prioritize resource downloads. You're most likely simply saturating the network.

kushaljshah:
@nammaianh The issue I am referring to is about how much parallel operations can be done asynchronously, on a single node. Irrespective of, if it is part of a cluster or not.

@Garbee All resource including CPU, RAM, Network are stable. The instance type which I am using has a network bandwidth of 2000 MBPS. Have performed the similar exercise for a batch of 5 domains as well, getting the same results, so network is surely not the culprit here. 

What is observed here is that all 'goto' operations for all urls get started  simultaneously but its completion and 'download' operation occur sequentially. There is some blocking operation happening which is resulting in this sequential competition. 

Sample output for batch size of 5 and their respective completion time stamps:
Start goto : url1 - 1518433045085
Start goto : url2 - 1518433045099
Start goto : url3 - 1518433045107
Start goto : url4 - 1518433045108
[...]

PR #3560, opened 2018-11-17
fix(page): navigating 11 pages simultaneously should not throw warning

aslushnikov:
NavigatorWatcher subscribes to Connection to get a `Disconnected` event,
causing us to hit the default max of 10 listeners constraint.

Technically we don't leak anything here and can safely bump
the maxListenersCount to Infinity.

However, we conveniently have `CDPSession`, and
can re-dispatch the event on it and keep the safety check in place.

References #1995

### DIFF
--- a/test/page.spec.js
+++ b/test/page.spec.js
@@ -752,6 +752,18 @@ module.exports.addTests = function({testRunner, expect, headless}) {
       process.removeListener('warning', warningHandler);
       expect(warning).toBe(null);
     });
+    it('should not leak listeners during navigation of 11 pages', async({page, context, server}) => {
+      let warning = null;
+      const warningHandler = w => warning = w;
+      process.on('warning', warningHandler);
+      await Promise.all([...Array(20)].map(async() => {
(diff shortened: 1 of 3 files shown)


test/page.spec.js
```diff
@@ -752,6 +752,18 @@ module.exports.addTests = function({testRunner, expect, headless}) {
...
+    it('should not leak listeners during navigation of 11 pages', async({page, context, server}) => {
+      let warning = null;
+      const warningHandler = w => warning = w;
+      process.on('warning', warningHandler);
+      await Promise.all([...Array(20)].map(async() => {
+        const page = await context.newPage();
+        await page.goto(server.EMPTY_PAGE);
+        await page.close();
+      }));
+      process.removeListener('warning', warningHandler);
```
JoelEinbinder, line 764: This test looks like it might fail in parallel if some other test creates a warning.

aslushnikov: :'(

JoelEinbinder (approved):

PR #3561, opened 2018-11-17
docs(api.md): Update setViewport doc

### LINKED ISSUE #3517: Page#setViewport required options
```js
	await page.setViewport({width: WIDTH});
```

with no `height`, puppeteer rejects with the error message:

```
Protocol error (Emulation.setDeviceMetricsOverride): Invalid parameters width: integer value expected
```

However, the documentation https://pptr.dev/#?product=Puppeteer&version=v1.10.0&show=api-pagesetviewportviewport does not indicate that both height and width are required.

IMHO setting just the width should change the width while leaving the height untouched (default being the current width/height).  However, if the current behavior is desirable, the docs should be updated to reflect that.  Other functions like page#setCookie explicitly mark fields as required in bold, and the same thing should probably happen here.

Nazanin1369:
Add required to height and width properties in setViewport function

fixes #3517

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1705,8 +1705,8 @@ puppeteer.launch().then(async browser => {
 
 #### page.setViewport(viewport)
 - `viewport` <[Object]>
-  - `width` <[number]> page width in pixels.
-  - `height` <[number]> page height in pixels.
+  - `width` <[number]> page width in pixels. **required**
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):

aslushnikov:
@Nazanin1369 thanks!

PR #3565, opened 2018-11-18
fix(api.md): mark optional arguments in signatures

vsemozhetbyt:
Some `[]` marks were missing.


aslushnikov (approved):
@vsemozhetbyt awesome, thank you! We'll probably lint this one day in future.

PR #3591, opened 2018-11-22
feat(executioncontext): warn on nested js handle

### LINKED ISSUE #3562: Better error message for TypeError: Converting circular structure to JSON
I was trying to use `page.evaluateHandle` and kept getting an error: `TypeError: Converting circular structure to JSON`. Everything I was sending was either Serializable or a JSHandle. Eventually, between reading the source and stepping through in the debugger, I discovered my error. I was passing a JSHandle _inside_ an object as an argument: `{ scope: JSHandle }`.

Would you be open to a patch that caught and re-raised that error, but also logged a warning about "Nesting JSHandles inside objects is not supported"? I would be happy to write it if you think this would be a worthwhile change.

bgschiller:
ExecutionContext.evaluateHandle accepts arguments that are either
serializable, or JSHandles. A potential confusion is that it *does not*
accept arguments that *contain* JSHandles.

This patch adds a log message warning when it encounters that situation.

Fixes #3562

### DIFF
--- a/lib/ExecutionContext.js
+++ b/lib/ExecutionContext.js
@@ -122,7 +122,8 @@ class ExecutionContext {
         userGesture: true
       });
     } catch (err) {
-      err.message += ' Are you passing a nested JSHandle?';
+      if (err instanceof TypeError && err.message === 'Converting circular structure to JSON')
+        err.message += ' Are you passing a nested JSHandle?';
       throw err;
     }
     const { exceptionDetails, result: remoteObject } = await funcCallResultP.catch(rewriteError);
--- a/lib/ExecutionContext.js
+++ b/lib/ExecutionContext.js
@@ -121,10 +121,9 @@ class ExecutionContext {
         awaitPromise: true,
         userGesture: true
       });
-    } catch (e) {
-      if (e instanceof TypeError && e.message === 'Converting circular structure to JSON')
-        warnOnNestedJSHandle(args);
-      throw e;
+    } catch (err) {
+      err.message += ' Are you passing a nested JSHandle?';
+      throw err;
     }
     const { exceptionDetails, result: remoteObject } = await funcCallResultP.catch(rewriteError);
     if (exceptionDetails)
@@ -168,28 +167,6 @@ class ExecutionContext {
       if (error.message.endsWith('Cannot find context with specified id'))
         throw new Error('Execution context was destroyed, most likely because of a navigation.');
       throw error;
-    }
-    /**
-     * @param {!Array<*>} args
-     */
-    function warnOnNestedJSHandle(args) {
-      const alreadyTraversed = [];
-      function walk(obj, level) {
-        if (alreadyTraversed.includes(obj)) return;
-
-        if (obj instanceof JSHandle) {
-          if (level >= 1)
-            console.warn('Found JSHandle nested within an argument. JSHandles may only be passed as top-level arguments');
-        } else if (Array.isArray(obj)) {
-          alreadyTraversed.push(obj);
-          obj.forEach(elem => walk(elem, level + 1));
-        } else if (typeof obj === typeof {}) {
-          alreadyTraversed.push(obj);
-          Object.keys(obj).forEach(k => walk(obj[k], level + 1));
-        }
-      }
-
-      args.forEach(arg => walk(arg, 0));
     }
 
   }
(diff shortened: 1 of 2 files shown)


SimonSchick:
I don't know if that's a good thing to do, if anything this should re-throw as another error, emitting warnings on incorrect usage that *should* result in errors doesn't make much sense.

I think it would be sufficient to alter the error message eg. `err.message += ' Are you passing a nested JSHandle?'`

aslushnikov:
I like @SimonSchick suggestion - that's the right amount of code/effort for the problem we're having.

bgschiller:
Sounds good, I'll make that change and push an update

lib/ExecutionContext.js
```diff
@@ -111,15 +111,21 @@ class ExecutionContext {
...
+      funcCallResultP = this._client.send('Runtime.callFunctionOn', {
+        functionDeclaration: functionText + '\n' + suffix + '\n',
+        executionContextId: this._contextId,
+        arguments: args.map(convertArgument.bind(this)),
+        returnByValue: false,
+        awaitPromise: true,
+        userGesture: true
+      });
+    } catch (err) {
+      err.message += ' Are you passing a nested JSHandle?';
```
SimonSchick: You *do* want to check if it's a TypeError still.

SimonSchick (approved):
This is better but I'm still unsure if this sort of validation is the right way to do at all.
There are probably hundreds possible ways to pass bad data into puppeteer and make it give obscure errors, that's (to some degree) due to the nature of JS.

If you want actual type safety, you want to go with something like TypeScript.

lib/ExecutionContext.js
bgschiller: Ah, right. I'll add that back.

bgschiller:
@SimonSchick, I'd absolutely prefer to lean on the type system for this. I actually *am* using Typescript, but I don't get an error in this case. Looking at `puppeteer/index.d.ts`, I see the following definition:

```ts
evaluateHandle(
  fn: EvaluateFn,
  ...args: any[]
): Promise<JSHandle>;
```

Even though args is documented as being `Serializable | JSHandle`, the type definitions don't include that information.

I'm pretty new to Typescript, but I'm not aware of a way to specify that arguments be serializable. Maybe you can point me in the right direction? Is there a built-in type I'm missing, or would it involve a definition something like this?

```ts
type Serializable = number | boolean | string | { [key: string]: Serializable } | Serializable[];
```

lib/ExecutionContext.js
```diff
@@ -111,15 +111,22 @@ class ExecutionContext {
...
+      funcCallResultP = this._client.send('Runtime.callFunctionOn', {
+        functionDeclaration: functionText + '\n' + suffix + '\n',
+        executionContextId: this._contextId,
+        arguments: args.map(convertArgument.bind(this)),
+        returnByValue: false,
+        awaitPromise: true,
+        userGesture: true
+      });
+    } catch (err) {
+      if (err instanceof TypeError && err.message === 'Converting circular structure to JSON')
```
aslushnikov, line 125: this error happens due to `convertArgument` call. Let's scope it down to this:

```js
try {
  args = args.map(convertArgument.bind(this));
} catch (err) {
  if (err instanceof TypeError && err.message === 'Converting circular structure to JSON')
    err.message += ' Are you passing a nested JSHandle?';
  throw err;
}
```

lib/ExecutionContext.js
```diff
@@ -162,6 +169,7 @@ class ExecutionContext {
         throw new Error('Execution context was destroyed, most likely because of a navigation.');
       throw error;
     }
+
```
aslushnikov: nit: stray line

aslushnikov (commented):
Thanks; can you please also add a test?

lib/ExecutionContext.js
bgschiller: Unfortunately, the error actually happens after convertArgument, in `CDPSession.send`. It looks like convertArgument prepares the args for serialization, but the call to JSON.stringify is later on.

bgschiller:
As far as I can tell, test failures are either unrelated (`1) Browser Page Workers Page.workers (worker.spec.js:8:5)`) or due to linting in files I haven't edited. Do I need to rebase/merge a more recent commit?

Also, I learned how to write a "Serializable" type, so I'm intending to submit a PR to DefinitelyTyped as well.

aslushnikov (approved):
Awesome, thanks!

Issue #2634, opened 2018-05-31
Support ServiceWorkers


aslushnikov:
We need an API to connect and manage ServiceWorkers.

At the very least, we need to:
- have a way to shut down service worker (requested at https://github.com/GoogleChrome/puppeteer/issues/1396)
- have a way to see service worker traffic (requested at https://github.com/GoogleChrome/puppeteer/issues/2617)
- support code coverage for service workers (requested at https://github.com/GoogleChrome/puppeteer/issues/2092)

dsych:
I wrote a [post](https://medium.com/@gogsmak/interacting-with-service-workers-using-puppeteer-583e5fc4a40b) that might provide a temporary workaround for people who want to interact with SW.

aadityataparia:
For people looking for a way to get sw coverage using istanbul:

I was able to get coverage of a istanbul instrumented `sw.js` file using message event:
```js
// sw.js
self.addEventListener('message', async (e) => {
  if (e.data === 'coverage') {
    // istanbul coverage is saved as __coverage__ variable
    e.ports[0].postMessage(self.__coverage__);
  }
});
```

```js
// index.js
function send_message_to_sw(msg) {
  return new Promise(function(resolve, reject) {
    // Create a Message Channel
    var msg_chan = new MessageChannel();

    // Handler for recieving message reply from service worker
    msg_chan.port1.onmessage = function(event) {
      if (event.data.error) {
        reject(event.data.error);
      } else {
        resolve(event.data);
      }
    };

    // Send message to service worker along with port for reply
    navigator.serviceWorker.controller.postMessage(msg, [msg_chan.port2]);
  });
};
```
And you can get coverage by `await send_message_to_sw('coverage')`

Meir017:
related to https://github.com/microsoft/playwright/issues/1090

PR #3607, opened 2018-11-29
feat(serviceworkers): add basic support for Service Workers

ksm2:
After figuring out how Service Workers are being handeld by Chrome DevTools, I created this PR which adds basic support for Service Workers. It allows users to get all Service Workers running in the browser, receive all requests which the Service Worker performs (including his own `sw.js` request), and listen for ServiceWorkerRegistration's `scopeURL` and `status`.

Example:
```js

function logRequest(interceptedRequest) {
  console.log('A request was made by the ServiceWorker:', interceptedRequest.url());
}

  page.on('serviceworkercreated', serviceWorker => {
    serviceWorker.on('request', logRequest);
  });
  
  await page.goto('https://service-worker-example.org/');
  await browser.close();
});
```

Unfortunately, I was not sure how to test this feature and would appreciate help with it. Also, I did not implement code coverage or stopping.

Addresses #2634.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -87,6 +87,9 @@
   * [event: 'response'](#event-response)
   * [event: 'serviceworkercreated'](#event-serviceworkercreated)
   * [event: 'serviceworkerdestroyed'](#event-serviceworkerdestroyed)
+  * [event: 'serviceworkerregistrationcreated'](#event-serviceworkerregistrationcreated)
+  * [event: 'serviceworkerregistrationdeleted'](#event-serviceworkerregistrationdeleted)
+  * [event: 'serviceworkerstatuschanged'](#event-serviceworkerstatuschanged)
   * [event: 'workercreated'](#event-workercreated)
   * [event: 'workerdestroyed'](#event-workerdestroyed)
   * [page.$(selector)](#pageselector)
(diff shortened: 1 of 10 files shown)


aslushnikov:
Hey @ksm2, thanks for the PR.

The main issue we're having with Service Workers is that we don't know yet the user scenarios. Your PR exposes *a lot* of APIs - is this really needed?

Can we start small, starting from the most painful testing usecases that will bring the most value to our users and expand further from there?

ksm2:
Hey @aslushnikov, thanks for your reply! Sure, I just wanted to include into my branch everything I found out using the Chrome DevTools Protocol how Chrome actually handles Service Workers. From our use case, the following would be most important:

- Inspecting requests the Service Worker executes (including its own `sw.js` request)
- Getting events when a Service Worker is getting active
- Retrieve a Service Worker's scope

At Baqend, we have a Service Worker which is its own product. We use Puppeteer to detect whether our users have installed the Service Worker correctly, so we could need the features mentioned above. I guess my PR is not adding a lot of APIs on top, because the way Service Workers work in Chrome is very complicated. If you have any ideas with which things to start, I am open to any changes!

aslushnikov:
> I guess my PR is not adding a lot of APIs on top, because the way Service Workers work in Chrome is very complicated. 

@ksm2 Oh yes, ServiceWorkers are crazy complicated - I totally agree here. Many scenarios, though, don't require dealing with the whole surface of Service Workers API. For example, "kill all service workers on this domain so that I can do a cold load" thing doesn't require understanding much of their API. (though this particular usecase is addressed with browser contexts so doesn't require SW-related stuff)

Rather than bringing the whole complexity of service workers world into PPTR API, I'd love to see the subset that addresses the most critical scenarios in testing world. This way we can not only keep the scope reasonable, but also help our clients use PPTR properly.

The #2634 has some ideas related to service worker functionality (e.g. coverage and network traffic), though we're not sure how critical/necessary they are to the community. 

Closing this for now in favor of the discussion in #2634.

PR #3653, opened 2018-12-11
feat(chromium): roll Chromium to r615489

aslushnikov:
This roll includes:
- https://crrev.com/611230 - Headless: enable disk cache for default profile
- https://crrev.com/613297 - headless: support --disk-cache-dir flag


JoelEinbinder (approved):
Will this have any impact on performance or the behavior of network requests?

PR #3654, opened 2018-12-11
feat(browser): introduce Browser.waitForNewTarget

aslushnikov:
This adds `browser.waitForNewTarget` and
`browserContext.waitForNewTarget` that wait for target to be
opened in future rather then search the one that currently
exists.


aslushnikov:
We didn't like it - closing.

PR #3669, opened 2018-12-13
docs(api.md): better documentation for PUPPETEER_DOWNLOAD_HOST

### LINKED ISSUE #3635: PUPPETEER_DOWNLOAD_HOST poorly named & documented
### Steps to reproduce

* Puppeteer version: 1.10.0
* Platform / OS version: Ubuntu 18.04
* URLs (if applicable):
* Node.js version:

PUPPETEER_DOWNLOAD_HOST is not a hostname, rather it is a URL prefix.  Hostnames do not include protocol, nor do they indicate path prefix.  This name makes the variable less discoverable for users who need to override protocol and / or path.

The documentation:
https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#environment-variables
"overwrite host part of URL that is used to download Chromium"

Again, this makes it seem like the PUPPETEER_DOWNLOAD_HOST cannot override the protocol or path prefix, when in fact, it can.  It may also lead to an incorrect value being supplied, e.g. "my-mirror.com" rather than "http://my-mirror.com".

At one point in the discussion of issue #951, PUPPETEER_DOWNLOAD_CHROMIUM_URL_PREFIX was considered as the name.  This is verbose, but far more descriptive of how the variable actually functions.

N/A

N/A

**What happens instead?**
N/A

aslushnikov:
Fix #3635.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -364,7 +364,7 @@ If Puppeteer doesn't find them in the environment during the installation step, 
 
 - `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
 - `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step.
-- `PUPPETEER_DOWNLOAD_HOST` - overwrite URL prefix that is used to download Chromium. Note: this includes protocol and might even include path prefix.
+- `PUPPETEER_DOWNLOAD_HOST` - overwrite URL prefix that is used to download Chromium. Note: this includes protocol and might even include path prefix. Defaults to `https://storage.googleapis.com`.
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -364,7 +364,7 @@ If Puppeteer doesn't find them in the environment during the installation step,
 
 - `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
 - `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step.
-- `PUPPETEER_DOWNLOAD_HOST` - overwrite host part of URL that is used to download Chromium
+- `PUPPETEER_DOWNLOAD_HOST` - overwrite URL prefix that is used to download Chromium. Note: this includes protocol and might even include path prefix.
```
JoelEinbinder: This is still unclear. Just say what it defaults to, or add an example.

JoelEinbinder (approved):

PR #3676, opened 2018-12-14
fix(browser): Close browser targets before closing browser itself

### LINKED ISSUE #3673: browser.close hangs with beforeunload dialog
Puppeteer: v1.11.0

In headful mode, `browser.close` doesn't force-close browser, stalling the process in case of beforeunload dialogs. This is a little surprising to me - I'd expect beforeunloads to be ignored by default.

```js

(async() => {
    headless: false,
    userDataDir: 'ppp',
  });
  await page.goto('https://www.w3schools.com/jsref/tryit.asp?filename=tryjsref_onbeforeunload');
  const frame = page.frames().find(frame => frame.name() === 'iframeResult');
  await Promise.all([
    // make sure beforeunload triggers once
    new Promise(resolve => page.once('dialog', resolve)).then(dialog => dialog.dismiss()),
    frame.click('a'),
  ]);
  // This will hang forever
  console.log('Exiting...');
  await browser.close();
  console.log('Done.');
})();
```

gidztech:
This PR fixes #3673 by adding logic to [browser.close](https://github.com/GoogleChrome/puppeteer/blob/v1.11.0/docs/api.md#browserclose) to close the browser targets first before the browser. This seems to ignore the beforeunload dialog on the page and succesfully closes the browser.

### DIFF
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -252,6 +252,10 @@ class Browser extends EventEmitter {
   }
 
   async close() {
+    await Promise.all(this.targets().map(async target => {
+      await this._connection.send('Target.closeTarget', { targetId: target._targetId });
+    }));
+
     await this._closeCallback.call(null);
     this.disconnect();
   }
--- a/test/browser.spec.js
+++ b/test/browser.spec.js
@@ -15,8 +15,9 @@
  */
 const utils = require('./utils');
 const puppeteer = utils.requireRoot('index');
+const {waitEvent} = utils;
 
-module.exports.addTests = function({testRunner, expect, headless}) {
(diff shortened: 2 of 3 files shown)


aslushnikov (commented):
Thanks for the PR. Solving this on puppeteer-side is racy: targets might spawn new targets with beforeunload dialogs and so on.

I've submitted a CL to fix this on the protocol-side: https://chromium-review.googlesource.com/c/chromium/src/+/1381513

aslushnikov:
Closing this PR in favor of upstream fix.

gidztech:
Ah yes that makes sense.

PR #3694, opened 2018-12-20
feat: introduce target.title()

brendankenny:
Talked to @JoelEinbinder about this and he suggested the change.

When looking at `background_page` targets of extensions, there's not really any way to tell which extension is the one you want without knowing the hash (the url is just something like `chrome-extension://kmendfapggjehodndflmmgagdbamhnfd/_generated_background_page.html`), which is a problem if you're dynamically loading the extension for testing.

The problem is compounded since `--disable-extensions-except` doesn't actually load *only* the extension specified. Currently, at least the 2FA `CryptoTokenExtension` is always loaded (at least on my machine) in addition to the extension path passed in.

`_targetInfo.title` looks like it's a stable part of the protocol and I can't see any other way to get this string, so hopefully this change seems reasonable.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -296,6 +296,7 @@
   * [target.createCDPSession()](#targetcreatecdpsession)
   * [target.opener()](#targetopener)
   * [target.page()](#targetpage)
+  * [target.title()](#targettitle)
   * [target.type()](#targettype)
   * [target.url()](#targeturl)
 - [class: CDPSession](#class-cdpsession)
@@ -3412,6 +3413,9 @@ Get the target that opened this target. Top-level targets return `null`.
 - returns: <[Promise]<?[Page]>>
 
 If the target is not of type `"page"` or `"background_page"`, returns `null`.
+
+#### target.title()
+- returns: <[string]> Returns target's title.
 
 #### target.type()
 - returns: <"page"|"background_page"|"service_worker"|"other"|"browser">
--- a/test/headful.spec.js
+++ b/test/headful.spec.js
@@ -76,6 +76,14 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions}) 
       expect(await page.evaluate(() => 2 * 3)).toBe(6);
       await browserWithExtension.close();
     });
+    it('target.title() should return extension title', async({}) => {
+      const browserWithExtension = await puppeteer.launch(extensionOptions);
+      const backgroundPageTarget = await waitForBackgroundPageTarget(browserWithExtension);
+      const backgroundPageTitle = backgroundPageTarget.title();
+      expect(backgroundPageTitle === 'Simple extension' ||
+        backgroundPageTitle === 'CryptoTokenExtension').toBeTruthy();
+      await browserWithExtension.close();
+    });
     it('should have default url when launching browser', async function() {
       const browser = await puppeteer.launch(extensionOptions);
       const pages = (await browser.pages()).map(page => page.url());
--- a/lib/Target.js
+++ b/lib/Target.js
@@ -75,6 +75,13 @@ class Target {
   }
 
   /**
+   * @return {string}
+   */
+  title() {
+    return this._targetInfo.title;
+  }
+
+  /**
    * @return {!Puppeteer.Browser}
    */
   browser() {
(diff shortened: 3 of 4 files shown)


test/headful.spec.js
```diff
@@ -76,6 +76,14 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions})
       expect(await page.evaluate(() => 2 * 3)).toBe(6);
       await browserWithExtension.close();
     });
+    it('target.title() should return extension title', async({}) => {
+      const browserWithExtension = await puppeteer.launch(extensionOptions);
+      const backgroundPageTarget = await waitForBackgroundPageTarget(browserWithExtension);
+      const backgroundPageTitle = backgroundPageTarget.title();
+      expect(backgroundPageTitle === 'Simple extension' ||
+        backgroundPageTitle === 'CryptoTokenExtension').toBeTruthy();
```
JoelEinbinder, line 84: I’d rather not have the crypto token leak into our test. How about just getting a page, setting the title, then doing page.target.title?

docs/api.md
```diff
@@ -3413,6 +3414,9 @@ Get the target that opened this target. Top-level targets return `null`.
 
 If the target is not of type `"page"` or `"background_page"`, returns `null`.
 
+#### target.title()
+- returns: <[string]> Returns target's title.
```
JoelEinbinder, line 3418: Remove the description, it looks redundant.

JoelEinbinder (approved):

test/headful.spec.js
brendankenny, line 84: At least on my machine, these tests actually load two extensions, `CryptoTokenExtension`, and `Simple extension`, and  `CryptoTokenExtension` looks like it's the one actually always passed back by `waitForBackgroundPageTarget` (which makes sense since it's presumably built in and loads first). Depending on when the extensions load relative to the test, after the `await`, `browser.targets()` sometimes has `Simple extension` and sometimes doesn't.

None of these tests seem to be dependent on exactly which extensions' `background_page` is being returned, so that seems fine, but it seemed like a bad idea to depend on `CryptoTokenExtension` existing forever (and appearing in this order) for this test. I could also modify `waitForBackgroundPageTarget` to be more "wait for a *particular* background target" instead of checking both titles if that's better.

brendankenny: > How about just getting a page, setting the title, then doing page.target.title?

No problem, but is the `target.spec.js` test sufficient without a test here then?

brendankenny:
We've decided this is too extension-specific since the target titles for basically any other type of target are useless. Closing :)

AmySteam:
@brendankenny The Puppeteer documentation [for Chrome Extensions](https://pptr.dev/#?product=Puppeteer&version=v1.18.1&show=api-working-with-chrome-extensions) is currently incorrect because of the issue you described:

> The problem is compounded since --disable-extensions-except doesn't actually load only the extension specified. Currently, at least the 2FA CryptoTokenExtension is always loaded (at least on my machine) in addition to the extension path passed in.

As you noted, Crypto is always loaded first and `pptr.launch` resolves before the second extension loads. It took me a day to figure out what was happening, and [I'm not the only one](https://github.com/GoogleChrome/puppeteer/issues/4300). It's pretty confusing to think you're testing your own extension when the page belongs to another one. 😉

We're putting together an issue for this, but I wanted to reach out first here. I know Chrome Extensions are a low priority, but we should at least fix the docs so people new to Puppeteer don't lose faith.

aslushnikov:
> As you noted, Crypto is always loaded first and pptr.launch resolves before the second extension loads. 

@AmySteam whoa, I missed this! Thanks for bringing this up!

> I know Chrome Extensions are a low priority, but we should at least fix the docs so people new to Puppeteer don't lose faith.

Definitely. You can pass the `--disable-component-extensions-with-background-pages` chromium flag to disable CryptoToken extension; in fact, we should do it by default in Puppeteer. Submitted https://github.com/GoogleChrome/puppeteer/pull/4704.

AmySteam:
@aslushnikov Thanks for your quick reply! I guess if you do it by default, the current documentation would work! 😃

I've been using `Target#waitForTarget` to wait for my extension to load, and adding `--disable-component-extensions-with-background-pages` gets the my extension without having to use `Target#_targetInfo.title`.

```javascript
const browser = await pptr.launch({
  headless: false,
  args: [
    '--disable-component-extensions-with-background-pages',
    `--disable-extensions-except=${pathToExtension}`,
    `--load-extension=${pathToExtension}`,
  ],
})

const backgroundTarget = await browser.waitForTarget(
  (t) => t.type() === 'background_page',
)

const backgroundPage = await backgroundTarget.page()
```

PR #3743, opened 2019-01-10
docs(troubleshooting): fix docker example

neilsh:
* removing libgconf-2-4 install since no longer needed according to https://bugs.chromium.org/p/chromium/issues/detail?id=795759#c7
* wget is already included in `node:8-slim` image, so removed lines related to install/cleanup
* node 8 has EOL this year, so incremented to node:10-slim
* use "docker run --init" if available (available in docker-engine >= 1.13.0)
* make dumb-init optional
* combine permission changes and 'npm install' of puppeteer into same line to reduce image size by few hundred MB
* overall image size reduction: 1.21GB -> 865MB

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -186,7 +186,7 @@ To fix, you'll need to install the missing dependencies and the
 latest Chromium package in your Dockerfile:
 
 ```Dockerfile
-FROM node:8-slim
+FROM node:10-slim
 
 # Install latest chrome dev package and fonts to support major charsets (Chinese, Japanese, Arabic, Hebrew, Thai and a few others)
 # Note: this installs the necessary libs to make the bundled version of Chromium that Puppeteer
@@ -199,9 +199,11 @@ RUN wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key
     && rm -rf /var/lib/apt/lists/* \
     && rm -rf /src/*.deb
 
-# It's a good idea to use dumb-init to help prevent zombie chrome processes.
-ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.2.0_amd64 /usr/local/bin/dumb-init
-RUN chmod +x /usr/local/bin/dumb-init
+# If running Docker >= 1.13.0 use docker run's --init arg to reap zombie processes, otherwise
+# uncomment the following lines to have `dumb-init` as PID 1
+# ADD https://github.com/Yelp/dumb-init/releases/download/v1.2.0/dumb-init_1.2.0_amd64 /usr/local/bin/dumb-init
+# RUN chmod +x /usr/local/bin/dumb-init
+# ENTRYPOINT ["dumb-init", "--"]
 
 # Uncomment to skip the chromium download when installing puppeteer. If you do,
 # you'll need to launch puppeteer with:
@@ -210,8 +212,8 @@ RUN chmod +x /usr/local/bin/dumb-init
 
 # Install puppeteer so it's available in the container.
 RUN npm i puppeteer \
-# Add user so we don't need --no-sandbox.
-# same layer as npm install to keep re-chowned files from using up several hundred MBs more space
+    # Add user so we don't need --no-sandbox.
+    # same layer as npm install to keep re-chowned files from using up several hundred MBs more space
     && groupadd -r pptruser && useradd -r -g pptruser -G audio,video pptruser \
     && mkdir -p /home/pptruser/Downloads \
     && chown -R pptruser:pptruser /home/pptruser \
@@ -220,7 +222,6 @@ RUN npm i puppeteer \
 # Run everything after as non-privileged user.
 USER pptruser
 
-ENTRYPOINT ["dumb-init", "--"]
 CMD ["google-chrome-unstable"]
 ```
 
@@ -233,7 +234,7 @@ docker build -t puppeteer-chrome-linux .
 Run the container by passing `node -e "<yourscript.js content as a string>` as the command:
 
 ```bash
- docker run -i --rm --cap-add=SYS_ADMIN \
+ docker run -i --init --rm --cap-add=SYS_ADMIN \
    --name puppeteer-chrome puppeteer-chrome-linux \
    node -e "`cat yourscript.js`"
 ```
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -191,17 +191,17 @@ FROM node:8-slim
 # See https://crbug.com/795759
 RUN apt-get update && apt-get install -yq libgconf-2-4
 
-# Install latest chrome dev package and fonts to support major charsets (Chinese, Japanese, Arabic, Hebrew, Thai and a few others)
+# Install latest chrome package and fonts to support major charsets (Chinese, Japanese, Arabic, Hebrew, Thai and a few others)
 # Note: this installs the necessary libs to make the bundled version of Chromium that Puppeteer
 # installs, work.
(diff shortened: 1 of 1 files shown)


jamsea (approved):
Looks good to me!

danielrussellLA (approved):

dorachan2000 (approved):
lgtm

docs/troubleshooting.md
```diff
@@ -191,17 +191,17 @@ FROM node:8-slim
...
-# Install latest chrome dev package and fonts to support major charsets (Chinese, Japanese, Arabic, Hebrew, Thai and a few others)
+# Install latest chrome package and fonts to support major charsets (Chinese, Japanese, Arabic, Hebrew, Thai and a few others)
 # Note: this installs the necessary libs to make the bundled version of Chromium that Puppeteer
 # installs, work.
 RUN apt-get update && apt-get install -y wget --no-install-recommends \
     && wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | apt-key add - \
     && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google.list' \
     && apt-get update \
-    && apt-get install -y google-chrome-unstable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst ttf-freefont \
+    && apt-get install -y google-chrome-stable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst ttf-freefont \
```
aslushnikov: Latest Puppeteer is usually ahead of Chrome release schedule, so I'd stick with unstable version here.

Alternatively, we should install both `google-chrome-stable` and `puppeteer@chrome-stable`.

neilsh: cool, I'll make that change

docs/troubleshooting.md
```diff
@@ -214,10 +209,10 @@ RUN chmod +x /usr/local/bin/dumb-init
 # ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true
 
 # Install puppeteer so it's available in the container.
-RUN npm i puppeteer
-
+RUN npm i puppeteer \
 # Add user so we don't need --no-sandbox.
-RUN groupadd -r pptruser && useradd -r -g pptruser -G audio,video pptruser \
+# same layer as npm install to keep re-chowned files from using up several hundred MBs more space
```
aslushnikov: I'm surprised this change has any effect on the resulting image. Can you please explain why? (or maybe you point me towards some documentation regarding this..)

ebidel: https://medium.com/@jessgreb01/digging-into-docker-layers-c22f948ed612 is good for explaining layers.

ebidel: @neilsh what's the savings in image size with these changes?

docs/troubleshooting.md
ebidel: I think we should stick with unstable in case it has new deps that is needed by the chromium that pptr installs.

ebidel:
Should we take this opportunity to remove the dumb-init lines and switch over the example run commands to using `--init`? That would save more space. We may want to note the version of docker that added that flag in the docs.

We could also show a dockerfile example that uses puppeteer-core since we're installing google-chrome-unstable anyways. However, that may not be guarnteed to work if the version of puppeteer needs a newer build than what google-chrome-unstable. But...the image would be MBs smaller :)

docs/troubleshooting.md
neilsh: The main culprit is the `chown -R pptruser:pptruser /node_modules` line. If  it's on a separate docker operation from the `npm install` it essentially creates a completely separate copy of the chrome binary in the node_modules. Just that `chown` layer adds an extra 323MB to the image. 

https://medium.com/@lmakarov/the-backlash-of-chmod-chown-mv-in-your-dockerfile-f12fe08c0b55

> Updating ownership on a file with chown effectively results in duplicating that file and storing its new copy in a new layer. The original copy is still there in the underlying layer. The same applies to chmod and even mv commands. When using them in your Dokerfile, you may end up with an inflated image, before you realize what’s going on.

@ebidel overall image size change I saw was 1.21GB -> 868MB

ebidel: Wow, that's insane. Good article!

neilsh:
@aslushnikov @ebidel I incorporated your feedback, let me know if you have any other suggestions!

edit: I think I'm going to go ahead and try taking a crack this weekend at @ebidel 's suggestion of using puppeteer-core and reduce the image size by another few hundred MBs. If we use the `puppeteer-core@chrome-stable` npm tag, I think we should be able to produce a stable image.

aslushnikov (approved):

neilsh:
After a little experimentation I confirmed the following Dockerfile produces a 494MB image with puppeteer-core@1.9.0, although it needs to be launched with `puppeteer.launch({executablePath: '/usr/bin/chromium'})`

Main differences from what I have above:
* install debian stretch's chromium71 instead of google-chrome-unstable
* use puppeteer-core@1.9.0 instead of puppeteer, which is compatible with chromium71 (on a related note, I wonder if it would make sense to start tagging puppeteer npm versions with the corresponding chrome version)

```
FROM node:10-slim

RUN apt-get update \
    && apt-get install -y chromium fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst ttf-freefont \
      --no-install-recommends \
    && rm -rf /var/lib/apt/lists/* \
    && rm -rf /src/*.deb

# node:10-slim has chromium 71, which corresponds to puppeteer@1.9.0
# puppeteer-core must be launched from JS with `puppeteer.launch({executablePath: '/usr/bin/chromium'});`
RUN npm i puppeteer-core@1.9.0 \
    && groupadd -r pptruser && useradd -r -g pptruser -G audio,video pptruser \
    && mkdir -p /home/pptruser/Downloads \
    && chown -R pptruser:pptruser /home/pptruser \
    && chown -R pptruser:pptruser /node_modules

# Run everything after as non-privileged user.
USER pptruser
CMD ["chromium"]
```

argyleink (approved):
💘

ebidel:
@neilsh Two things:

1. There are tags for chromium versions now. It's not called out well, but you can find it at https://github.com/GoogleChrome/puppeteer#q-why-doesnt-puppeteer-vxxx-work-with-chromium-vyyy. e.g. `puppeteer-core@chrome-71`. Let's switch to using that. We should also remind folks to use `const puppeteer = require('puppeteer-core');` in their code.
2. I think we should also add another minimal Dockerfile example that shows how to use the latest version of Puppeteer and does not require users to modify their `.launch()` command:

```
FROM node:10-slim

RUN apt-get update \
    && apt-get install -y google-chrome-unstable fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-kacst ttf-freefont \
      --no-install-recommends \
    && rm -rf /var/lib/apt/lists/* \
    && rm -rf /src/*.deb

# Installs the latest version of Puppeteer. Note: this (by default) will also install another build of Chromium on the system. If care about image size, we recommend using the other Dockerfile example.
RUN npm i puppeteer \
    && groupadd -r pptruser && useradd -r -g pptruser -G audio,video pptruser \
    && mkdir -p /home/pptruser/Downloads \
    && chown -R pptruser:pptruser /home/pptruser \
    && chown -R pptruser:pptruser /node_modules

# Run everything after as non-privileged user.
USER pptruser
CMD ["google-chrome-unstable"]
```

Can you make those updates to in the PR?

aslushnikov:
Merged this as-is; we can further iterate on in follow-ups.

PR #3758, opened 2019-01-12
chore(ci): attempt to fix travis

aslushnikov:


JoelEinbinder:
It doesn't look like it worked

PR #3770, opened 2019-01-15
fix(frames): make sure evaluation does not hang in detached iframes

### LINKED ISSUE #3261: `frame.evaluate` freezes / hangs on detached frames.
* Puppeteer version: tested with 1.6.2 and 1.8.0
* Platform / OS version: ubuntu 16.04
* Node.js version: 8.9.1

Repro script:
```typescript

await page.goto('https://example.com')
await page.evaluate(`
  const iframe = document.createElement('iframe')
  iframe.src = 'https://www.example.com'
  document.body.appendChild(iframe)
  setTimeout(() => document.body.removeChild(iframe), 1000)
`)

// The frame does not appear immediately.
await new Promise((resolve) => setTimeout(resolve, 100))

const iframe = page.frames()[1]
console.log(iframe.isDetached())  // -> false
console.log(await iframe.evaluate(() => window.location.href))

await new Promise((resolve) => setTimeout(resolve, 1500))

console.log(iframe.isDetached())  // -> true
// Uncomment this line to make the script hangs here.
// console.log(await iframe.evaluate(() => window.location.href))

await browser.close()
```

If you un-comment the last `.evaluate`, the code hangs whereas it should not. I guess it should raise an error like `DetachedFrameEvalutionError` or something similar to allow to catch it. Note that you can check manually `frame.isDetached()` to avoid the call to `evaluate` but IMHO:
1. if you're manipulating many frames in many places, it's much more convient to have one single `try-catch` to catch all the detached frame errors (and give up with this frame) than to put the check around each call to `evaluate`.
2. more importantly, it's probably subject to some race conditions.

Thanks!

aslushnikov:
Fix #3261

### DIFF
--- a/test/frame.spec.js
+++ b/test/frame.spec.js
@@ -60,7 +60,7 @@ module.exports.addTests = function({testRunner, expect}) {
       await utils.detachFrame(page, 'frame1');
       let error = null;
       await frame1.evaluate(() => 7 * 8).catch(e => error = e);
-      console.log(error);
+      expect(error.message).toContain('Cannot evaluate in detached frame');
     });
   });
 
--- a/test/frame.spec.js
+++ b/test/frame.spec.js
@@ -55,12 +55,15 @@ module.exports.addTests = function({testRunner, expect}) {
       const windowHandle = await mainFrame.evaluateHandle(() => window);
       expect(windowHandle).toBeTruthy();
     });
+  });
+
+  describe('Frame.evaluate', function() {
     it('should throw for detached frames', async({page, server}) => {
       const frame1 = await utils.attachFrame(page, 'frame1', server.EMPTY_PAGE);
       await utils.detachFrame(page, 'frame1');
       let error = null;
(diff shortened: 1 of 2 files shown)


test/frame.spec.js
```diff
@@ -55,6 +55,13 @@ module.exports.addTests = function({testRunner, expect}) {
       const windowHandle = await mainFrame.evaluateHandle(() => window);
       expect(windowHandle).toBeTruthy();
     });
+    it('should throw for detached frames', async({page, server}) => {
+      const frame1 = await utils.attachFrame(page, 'frame1', server.EMPTY_PAGE);
+      await utils.detachFrame(page, 'frame1');
+      let error = null;
+      await frame1.evaluate(() => 7 * 8).catch(e => error = e);
+      console.log(error);
```
JoelEinbinder: Log

aslushnikov: Fixed

JoelEinbinder (approved):

PR #3782, opened 2019-01-15
chore(testrunner): sort workers numerically when terminating

aslushnikov:
See screenshot for the workers state:

image


aslushnikov:
btw 133 tests in 1 second :fire:

JoelEinbinder (approved):

PR #3803, opened 2019-01-18
fix(page): include event context on 'load' and 'domcontentloaded'

mbeasley:
For both the 'load' and 'domcontentloaded' events, we were just emitting the event without any context about the event. The [event from dev tools](https://chromedevtools.github.io/devtools-protocol/tot/Page#event-loadEventFired) includes the timestamp which is fairly useful for any timing measurements.  This change adds those events to the emitted context.

On a side note, it's a little odd that the `load` event doesn't include any context about the frame (like [lifecycle events](https://chromedevtools.github.io/devtools-protocol/tot/Page#event-lifecycleEvent) do), but I think that's more an issue for dev tools.


mbeasley:
Waiting on the CLA side of this.

mbeasley:
Closing so that I can resubmit under our CLA.

PR #3809, opened 2019-01-19
fix(page): use secondary DOMWorld to drive page.select()

### LINKED ISSUE #3327: Puppeteer page.select got error: Cannot read property 'indexOf' of undefined

aslushnikov:
This patch starts creating secondary DOMWorld for every connected
page and switches `page.select()` to run inside the secondary world.

Drive-by: move `DOMWorld.waitFor` into Frame: in future, the shortcut will expand into calls in different DOMWorlds, e.g. `frame.waitForFunction` should be always  run in main world, whereas `frame.waitForSelector` in secondary.

Fix #3327.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -17,9 +17,11 @@
 const EventEmitter = require('events');
 const {helper, assert} = require('./helper');
 const {Events} = require('./Events');
-const {ExecutionContext} = require('./ExecutionContext');
+const {ExecutionContext, EVALUATION_SCRIPT_URL} = require('./ExecutionContext');
 const {LifecycleWatcher} = require('./LifecycleWatcher');
 const {DOMWorld} = require('./DOMWorld');
+
+const UTILITY_WORLD_NAME = '__puppeteer_utility_world__';
 
 class FrameManager extends EventEmitter {
   /**
@@ -38,6 +40,8 @@ class FrameManager extends EventEmitter {
     this._frames = new Map();
     /** @type {!Map<number, !ExecutionContext>} */
     this._contextIdToContext = new Map();
+    /** @type {!Set<string>} */
+    this._isolatedWorlds = new Set();
 
     this._client.on('Page.frameAttached', event => this._onFrameAttached(event.frameId, event.parentFrameId));
     this._client.on('Page.frameNavigated', event => this._onFrameNavigated(event.frame));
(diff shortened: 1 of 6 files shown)


lib/FrameManager.js
```diff
@@ -17,10 +17,12 @@
 const EventEmitter = require('events');
 const {helper, assert} = require('./helper');
 const {Events} = require('./Events');
-const {ExecutionContext} = require('./ExecutionContext');
+const {ExecutionContext, EVALUATION_SCRIPT_URL} = require('./ExecutionContext');
 const {LifecycleWatcher} = require('./LifecycleWatcher');
 const {DOMWorld} = require('./DOMWorld');
 
+const UTILITY_WORLD_NAME = '__puppeteer_utility_world__';
+
```
JoelEinbinder, line 25: Let’s cover connecting two puppeteers to a browser with a test

lib/Page.js
```diff
@@ -50,7 +50,7 @@ class Page extends EventEmitter {
       client.send('Target.setAutoAttach', {autoAttach: true, waitForDebuggerOnStart: false}),
       client.send('Page.setLifecycleEventsEnabled', { enabled: true }),
       client.send('Network.enable', {}),
-      client.send('Runtime.enable', {}),
+      client.send('Runtime.enable', {}).then(() => page._frameManager.ensureSecondaryDOMWorld()),
```
JoelEinbinder, line 53: why does this have to happen after runtime.enable?

lib/FrameManager.js
aslushnikov: we have it already - https://github.com/GoogleChrome/puppeteer/blob/c09835fd70e144cb705eb5ae8b169e49725532bd/test/puppeteer.spec.js#L305-L316

It caused me troubles.

lib/Page.js
aslushnikov: All execution contexts will be reported before Runtime.enable returns. If there's the one with helper isolated world, it'll be reported.

JoelEinbinder (approved):

PR #3826, opened 2019-01-22
test: Add test for waitForFunction with cross-process navigation

LinusU:
Added a test case for running `waitForFunction` during navigations. There is a similar test case for `waitForSelector`, but I didn't see any for `waitForFunction` that also has `goto` in it.


aslushnikov (approved):

LinusU:
Thanks for the quick pull ❤️

aslushnikov:
Thanks for helping!

PR #3827, opened 2019-01-22
refactor: move to flatten protocol

aslushnikov:
DevTools protocol is dropping nested targets and switching to
flatten protocol. This patch adopts the new scheme.

Once this change lands, tip-of-tree Puppeteer will be incompatible
with Chromium below 72.0.3606.0. Chromium 72 goes stable on [Jan, 29](https://www.chromestatus.com/features/schedule) - the same time we release the
next version of Puppeteer, so this change won't hurt those clients who try using
tip-of-tree Puppeteer with stable chrome. 

For the record: the previous attempt to land this was https://github.com/GoogleChrome/puppeteer/pull/3524.

### DIFF
--- a/lib/Connection.js
+++ b/lib/Connection.js
@@ -75,6 +75,10 @@ class Connection extends EventEmitter {
     });
   }
 
+  /**
+   * @param {*} message
+   * @return {number}
+   */
   _rawSend(message) {
     const id = ++this._lastId;
     message = JSON.stringify(Object.assign({}, message, {id}));
@@ -106,9 +110,7 @@ class Connection extends EventEmitter {
       const session = this._sessions.get(object.sessionId);
       if (session)
         session._onMessage(object);
-      return;
-    }
-    if (object.id) {
+    } else if (object.id) {
       const callback = this._callbacks.get(object.id);
       // Callbacks could be all rejected if someone has called `.dispose()`.
       if (callback) {
@@ -118,9 +120,9 @@ class Connection extends EventEmitter {
         else
           callback.resolve(object.result);
       }
-      return;
+    } else {
+      this.emit(object.method, object.params);
     }
-    this.emit(object.method, object.params);
   }
 
   _onClose() {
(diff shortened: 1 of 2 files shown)


JoelEinbinder:
I don't think it affects any users, but it's worth noting that the CDPSession public api cannot be used with the flattened protocol at the moment. We only accept method/params in CDPSession.send

lib/Connection.js
```diff
@@ -66,15 +69,20 @@ class Connection extends EventEmitter {
...
-    const message = JSON.stringify({id, method, params});
-    debugProtocol('SEND ► ' + message);
-    this._transport.send(message);
+    const id = this._rawSend({method, params});
     return new Promise((resolve, reject) => {
       this._callbacks.set(id, {resolve, reject, error: new Error(), method});
     });
   }
 
+  _rawSend(message) {
```
JoelEinbinder, line 82: I'd expect rawSend to take a string, or just stringify my object and send it raw. Not do id fiddling. Maybe call this _sendWithId or something?

lib/Connection.js
```diff
@@ -83,6 +91,23 @@ class Connection extends EventEmitter {
...
+      if (session) {
+        session._onClosed();
+        this._sessions.delete(object.params.sessionId);
+      }
+    }
+    if (object.sessionId) {
+      const session = this._sessions.get(object.sessionId);
+      if (session)
+        session._onMessage(object);
+      return;
```
JoelEinbinder: nit: I'd prefer else ifs here. Thats just me though.

lib/Connection.js
```diff
@@ -168,28 +176,16 @@ class CDPSession extends EventEmitter {
...
-    });
+    const id = this._connection._rawSend({sessionId: this._sessionId, method, params});
     return new Promise((resolve, reject) => {
       this._callbacks.set(id, {resolve, reject, error: new Error(), method});
     });
   }
 
   /**
-   * @param {string} message
+   * @param {*} object
```
JoelEinbinder: {id?: number, method: string, params: Object}

JoelEinbinder (approved):

aslushnikov:
>  the CDPSession public api cannot be used with the flattened protocol at the moment.

@JoelEinbinder what does it mean?

Issue #1434, opened 2017-11-20
expose timing information


aslushnikov:
Exposing timing information would greatly benefit performance scenarios.
We shoud:
1. add `DOMContentLoaded` and `NavigationStart` metrics to the `Page.metrics()` struct
2. expose timing information for requests/response objects

There's currently a suspicion that [1] is not aligned with [2] in the protocol. We, however, should pick a single time scale and use it throughout the api.

yanivefraim:
@aslushnikov - can I help here? (:

aslushnikov:
@yanivefraim please go ahead!

yanivefraim:
> There's currently a suspicion that [1] is not aligned with [2] in the protocol. We, however, should pick a single time scale and use it throughout the api.

@aslushnikov - just want to make sure that I understand this correct:

> 1. add DOMContentLoaded and NavigationStart metrics to the Page.metrics() struct

This is section is about:

Options a:  expose those two events from protocol to metrics struct
Options b: use performance.timing for that

> 2. expose timing information for requests/response objects

take timing for each req/res against `NavigationStart` from [1]

Did I get it right?

aslushnikov:
Hey @yanivefraim:
- for the (1), the `DOMContentLoaded` and `NavigationStart` events are filtered out from metrics event: https://github.com/GoogleChrome/puppeteer/blob/1c2adf61e971577a6bc96621fcb78ee88a9a9ff8/lib/Page.js#L983-L998 The main reason for that is the uncertainty on the time scale that's used for the values.
- (2) is best summarized by Paul here: https://github.com/GoogleChrome/puppeteer/pull/2015#issuecomment-377588552

PR #3841, opened 2019-01-25
fix(page): include event context on 'load' and 'domcontentloaded'

mbeasley:
For both the 'load' and 'domcontentloaded' events, we were just emitting the event without any context about the event. The [event from dev tools](https://chromedevtools.github.io/devtools-protocol/tot/Page#event-loadEventFired) includes the timestamp which is fairly useful for any timing measurements.  This change adds those events to the emitted context.

On a side note, it's a little odd that the `load` event doesn't include any context about the frame (like [lifecycle events](https://chromedevtools.github.io/devtools-protocol/tot/Page#event-lifecycleEvent) do), but I think that's more an issue for dev tools.

*Note: this was originally submitted in #3803 but resubmitted to be in alignment with our CLA*

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -129,8 +129,8 @@ class Page extends EventEmitter {
     this._networkManager.on(Events.NetworkManager.RequestFailed, event => this.emit(Events.Page.RequestFailed, event));
     this._networkManager.on(Events.NetworkManager.RequestFinished, event => this.emit(Events.Page.RequestFinished, event));
 
-    client.on('Page.domContentEventFired', event => this.emit(Events.Page.DOMContentLoaded));
-    client.on('Page.loadEventFired', event => this.emit(Events.Page.Load));
+    client.on('Page.domContentEventFired', event => this.emit(Events.Page.DOMContentLoaded, event));
+    client.on('Page.loadEventFired', event => this.emit(Events.Page.Load, event));
     client.on('Runtime.consoleAPICalled', event => this._onConsoleAPI(event));
     client.on('Runtime.bindingCalled', event => this._onBindingCalled(event));
     client.on('Page.javascriptDialogOpening', event => this._onDialog(event));


mbeasley:
I signed it!

mbeasley:
@aslushnikov I apologize if this is the wrong way to go about this, but is there anything additional needed from me to help get this over the line? I wasn't sure if tying this to an issue would be more helpful. Let me know if there's something more I should be doing. Thank you!

aslushnikov:
> I apologize if this is the wrong way to go about this, but is there anything additional needed from me to help get this over the line? I wasn't sure if tying this to an issue would be more helpful. Let me know if there's something more I should be doing. Thank you!

@mbeasley sorry, it took me a while to crawl through all notifications to get here. You did everything nicely!

I have a few concerns:
- we try not to expose protocol events as-is - instead, we expose only the necessary parts. In your case, you're interested in a timestamp, so I'd fire event with an explicitly created new object that has a "timestsamp" field that's picked from CDP event
- the timestamps in  `load` and `domcontentloaded` events are in something called `MonotonicTime`. Are you sure this is what you want?

mbeasley:
@aslushnikov regarding the event (just to make sure I'm understanding this correctly), would the idea be to simply just add a handler that emits a custom event similar to what is done for things like `_emitMetrics`? If so, so that's pretty straightforward. 

Definitely looking to use `MonotonicTime`, as that seems to be consistent with what is used on other events; while arbitrary, it's been reliably consistent for measurements between events.

aslushnikov:
> egarding the event (just to make sure I'm understanding this correctly), would the idea be to simply just add a handler that emits a custom event similar to what is done for things like _emitMetrics? If so, so that's pretty straightforward.

@mbeasley exactly

> Definitely looking to use MonotonicTime, as that seems to be consistent with what is used on other events; while arbitrary, it's been reliably consistent for measurements between events.

This seems to be related to #1434. Why don't you use the `window.performance.timing`? It should satisfy all your needs.

aslushnikov:
Closing this; let us know if `window.performance.timing` doesn't work for you.

PR #3856, opened 2019-01-28
fix(page): execute frame.waitFor{Selector,XPath} in secondary world

### LINKED ISSUE #609: page.waitForSelector doesn't work on baidu.com

aslushnikov:
This patch starts executing frame.waitForSelector and frame.waitForXPath
in secondary world. As a result, websites that mutate page global
context (e.g. removing global MutationObserver) don't break Puppeteer's
behavior.

Fixes #609

### DIFF
--- a/lib/ExecutionContext.js
+++ b/lib/ExecutionContext.js
@@ -175,8 +175,8 @@ class ExecutionContext {
   }
 
   /**
-   * @param {!ElementHandle} elementHandle
-   * @return {Promise<ElementHandle>}
+   * @param {Puppeteer.ElementHandle} elementHandle
+   * @return {Promise<Puppeteer.ElementHandle>}
    */
   async _adoptElementHandle(elementHandle) {
     assert(elementHandle.executionContext() !== this, 'Cannot adopt handle that already belongs to this execution context');
@@ -187,7 +187,7 @@ class ExecutionContext {
     const {object} = await this._client.send('DOM.resolveNode', {
       backendNodeId: nodeInfo.node.backendNodeId,
     });
-    return createJSHandle(this, object);
+    return /** @type {Puppeteer.ElementHandle}*/(createJSHandle(this, object));
   }
 }
 
--- a/lib/ExecutionContext.js
+++ b/lib/ExecutionContext.js
@@ -173,6 +173,22 @@ class ExecutionContext {
     });
     return createJSHandle(this, response.objects);
   }
+
+  /**
+   * @param {!ElementHandle} elementHandle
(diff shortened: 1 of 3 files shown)


test/waittask.spec.js
```diff
@@ -247,15 +254,6 @@ module.exports.addTests = function({testRunner, expect, product}) {
       expect(eHandle.executionContext().frame()).toBe(frame2);
     });
 
-    it('should throw if evaluation failed', async({page, server}) => {
```
JoelEinbinder, line 250: Does this imply that we can never throw now?

aslushnikov: yep, only for timeouts

JoelEinbinder (approved):

lib/ExecutionContext.js
```diff
@@ -173,6 +173,22 @@ class ExecutionContext {
     });
     return createJSHandle(this, response.objects);
   }
+
+  /**
+   * @param {Puppeteer.ElementHandle} elementHandle
+   * @return {Promise<Puppeteer.ElementHandle>}
+   */
+  async _adoptElementHandle(elementHandle) {
+    assert(elementHandle.executionContext() !== this, 'Cannot adopt handle that already belongs to this execution context');
```
JoelEinbinder, line 182: Fine for now. But in the case of a future public version we should probably `return elementHandle` here instead of asserting.

aslushnikov: Public API will be either on DOMWorld or on ElementHandle, so we'll have an opportunity to think about it

lib/ExecutionContext.js
```diff
@@ -173,6 +173,22 @@ class ExecutionContext {
...
+    assert(this._world, 'Cannot adopt handle without DOMWorld');
+    const nodeInfo = await this._client.send('DOM.describeNode', {
+      objectId: elementHandle._remoteObject.objectId,
+    });
+    const {object} = await this._client.send('DOM.resolveNode', {
+      backendNodeId: nodeInfo.node.backendNodeId,
```
JoelEinbinder, line 188: how does this work without taking a worldName?

aslushnikov: Follow-up: https://github.com/GoogleChrome/puppeteer/pull/3857

PR #3872, opened 2019-02-01
feat(firefox): implement browser contexts

aslushnikov:
BrowserContexts are necessary to run Puppeteer tests against Puppeteer-Firefox


aslushnikov:
TBR @JoelEinbinder

Issue #3889, opened 2019-02-02
firefox: merge Puppeteer-Firefox tests with Puppeteer tests


aslushnikov:
When working on prototype of Puppeteer-Firefox, we were copying many Puppeteer tests, adding some others and changing a few.

This issue is to track progress on unifying tests for Puppeteer and Puppeteer-Firefox.
The roadmap:
- bring Puppeteer-Firefox unique tests to Puppeteer
- skip failing tests and enable Puppeteer-Firefox on CI

For the record: There's a helper script [`//test/difftests.js`](https://gist.github.com/aslushnikov/31244444cf920305d53c06339f08a217) (not committed - dirty but useful) that diffs tests between Puppeteer and Puppeteer-Firefox.

PR #3907, opened 2019-02-04
test(firefox): support Puppeteer-Firefox specific env variables

aslushnikov:
This patch:
- adds support to `FFOX` env variable for Puppeteer testsuite
- install Firefox preferences when running tests with custom firefox
  executable

References #3889


JoelEinbinder (approved):

JoelEinbinder:
Lint is unhappy

PR #3925, opened 2019-02-06
feat(chromium): roll Chromium to r629336

aslushnikov:
Just a healthy roll.


JoelEinbinder:
We are waiting on a v8 coverage fix iirc.

PR #3928, opened 2019-02-06
test(network.spec.js): fix weird string concatenation

Meir017:
clean-up some weird string concatenation in a test


aslushnikov:
This is an important string concatenation - otherwise parser fails

See also: https://stackoverflow.com/questions/236073/why-split-the-script-tag-when-writing-it-with-document-write

Meir017:
image

PR #3937, opened 2019-02-07
test(firefox): enable more tests on Puppeteer-Firefox

aslushnikov:
References #3889

### DIFF
--- a/test/keyboard.spec.js
+++ b/test/keyboard.spec.js
@@ -136,7 +136,7 @@ module.exports.addTests = function({testRunner, expect, FFOX}) {
             'Keyup: ~ Backquote 192 [Shift]'].join('\n'));
       await keyboard.up('Shift');
     });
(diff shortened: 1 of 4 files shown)


test/keyboard.spec.js
```diff
@@ -146,7 +146,7 @@ module.exports.addTests = function({testRunner, expect, FFOX}) {
           if (event.key === 'l')
             event.preventDefault();
           if (event.key === 'o')
-            Promise.resolve().then(() => event.preventDefault());
```
JoelEinbinder, line 149: firefox doesn't let you preventDefault after a promise resolve?!

JoelEinbinder (approved):

test/keyboard.spec.js
aslushnikov: apparently, yeah

PR #3963, opened 2019-02-09
fix(firefox): add ExecutionContext to api

aslushnikov:


JoelEinbinder (approved):
Are we returning it anywhere yet?

aslushnikov:
It's used for the "Is Puppeteer-Firefox Ready" website to compare APIs.

PR #3988, opened 2019-02-13
feat(firefox): basic support for Network

aslushnikov:
This patch introduces basic Request and Response events for
page. It also teaches navigation methods, e.g. `page.goto` to return
navigation response.


JoelEinbinder (approved):
I assume we are going to add JSDoc later?

aslushnikov:
> I assume we are going to add JSDoc later?

Yeah; there's a lot of movement atm as I'm getting to know gecko network stack.

PR #3989, opened 2019-02-13
feat(firefox): Page.waitForRequest/Page.waitForResponse

aslushnikov:
Drive-by: refactor `Request.frame()` tests into a separate test suite.


aslushnikov:
@JoelEinbinder - I now filter out favicon requests (firefox reports them just fine). There's a headless bug regarding favicon not reported, so might be useful eventually.

PR #4002, opened 2019-02-14
feat(firefox): support consoleMessage.location()

aslushnikov:


JoelEinbinder (commented):
Why is the column different?

aslushnikov:
> Why is the column different?

Chrome reports different location wrt Firefox (Firefox's nicer).

test/page.spec.js
```diff
@@ -355,7 +355,7 @@ module.exports.addTests = function({testRunner, expect, headless, Errors, Device
       expect(message.location()).toEqual({
         url: server.PREFIX + '/consolelog.html',
         lineNumber: 7,
-        columnNumber: 14,
+        columnNumber: CHROME ? 14 : 6,
```
JoelEinbinder: can we add a comment here.
|console.log
vs
console.|log

JoelEinbinder (approved):

PR #4003, opened 2019-02-14
feat(firefox): support running beforeunload hooks when closing

aslushnikov:


test/page.spec.js
```diff
@@ -48,19 +48,29 @@ module.exports.addTests = function({testRunner, expect, headless, Errors, Device
...
-      newPage.close({ runBeforeUnload: true });
+      const pageClosingPromise = newPage.close({ runBeforeUnload: true });
       const dialog = await waitEvent(newPage, 'dialog');
       expect(dialog.type()).toBe('beforeunload');
-      expect(dialog.defaultValue()).toBe('');
-      expect(dialog.message()).toBe('');
-      dialog.accept();
-      await waitEvent(newPage, 'close');
+      if (CHROME) {
+        expect(dialog.defaultValue()).toBe('');
```
JoelEinbinder: Are these not implemented in Firefox?

aslushnikov: the `message()` is intentionally empty in Chrome for beforeunload - to avoid scam.

Updated the with the actual ffox text.

PR #4006, opened 2019-02-14
fix(firefox): enable a bunch of evaluation tests

aslushnikov:


experimental/puppeteer-firefox/lib/ExecutionContext.js
```diff
@@ -21,9 +21,34 @@ class ExecutionContext {
       });
       return createHandle(this, payload.result, payload.exceptionDetails);
     }
+    if (typeof pageFunction !== 'function')
+      throw new Error('The following is not a function: ' + pageFunction);
```
JoelEinbinder: Say what we tried to do please. "Cannot evaluate".

The error:
"The following is not a function: 56" looks pretty unhelpful.

JoelEinbinder (approved):

experimental/puppeteer-firefox/lib/ExecutionContext.js
aslushnikov: sure, reworded both here and in Puppeteer

PR #4014, opened 2019-02-14
fix: Respect the loglevel for logging

stphnrdmr:
This adds support for disabling console.log when the loglevel asks for it.

### DIFF
--- a/install.js
+++ b/install.js
@@ -21,15 +21,15 @@ if (require('./package.json').name === 'puppeteer-core')
 buildNode6IfNecessary();
 
 if (process.env.PUPPETEER_SKIP_CHROMIUM_DOWNLOAD) {
-  console.log('**INFO** Skipping Chromium download. "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" environment variable was found.');
+  log('**INFO** Skipping Chromium download. "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" environment variable was found.');
   return;
 }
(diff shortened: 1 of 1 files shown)


stphnrdmr:
I think it could be argued that disabling this during `CI` is not universally loved. Respecting the log level seems like a good idea to me.

install.js
```diff
@@ -145,3 +145,17 @@ function generateProtocolTypesIfNecessary(updated) {
     return;
   return require('./utils/protocol-types-generator');
 }
+
+function log(toBeLogged) {
+  function isTrue(value) {
+    return !!value && value !== '0' && value !== 'false';
+  }
+
+  const envDisable = isTrue(process.env.CI);
```
JoelEinbinder: Is it standard practice to disable logs when CI is set? I find the logs useful at least on our CI.

stphnrdmr: I agree with you and I’ll remove that part.

stphnrdmr:
@JoelEinbinder Do you agree with the latest changes? I don't know why CI is failing right now.

install.js
```diff
@@ -145,3 +145,12 @@ function generateProtocolTypesIfNecessary(updated) {
     return;
   return require('./utils/protocol-types-generator');
 }
+
+function log(toBeLogged) {
```
JoelEinbinder: nit: I'd rename log to logPolitely or to be a little easier to read what this does.

JoelEinbinder (approved):

install.js
stphnrdmr: I like logPolitely, I’ll use that.

PR #4016, opened 2019-02-14
feat(executioncontext): support bigints transferring

vsemozhetbyt:
Refs: https://chromedevtools.github.io/devtools-protocol/tot/Runtime#type-UnserializableValue

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -66,6 +66,8 @@ class Helper {
   static valueFromRemoteObject(remoteObject) {
     assert(!remoteObject.objectId, 'Cannot extract value when objectId is given');
     if (remoteObject.unserializableValue) {
+      if (remoteObject.type === 'bigint' && typeof BigInt !== 'undefined')
+        return BigInt(remoteObject.unserializableValue.replace('n', ''));
       switch (remoteObject.unserializableValue) {
         case '-0':
           return -0;
--- a/docs/api.md
+++ b/docs/api.md
@@ -1242,7 +1242,7 @@ List of all available devices is available in the source code: [DeviceDescriptor
 
 If the function passed to the `page.evaluate` returns a [Promise], then `page.evaluate` would wait for the promise to resolve and return its value.
 
-If the function passed to the `page.evaluate` returns a non-[Serializable] value, then `page.evaluate` resolves to `undefined`.
+If the function passed to the `page.evaluate` returns a non-[Serializable] value, then `page.evaluate` resolves to `undefined`. DevTools Protocol also supports transferring some additional values that are not serializable by `JSON`: `-0`, `NaN`, `Infinity`, `-Infinity`, and bigint literals.
 
 Passing arguments to `pageFunction`:
 ```js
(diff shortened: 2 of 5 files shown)


lib/helper.js
```diff
@@ -66,6 +66,8 @@ class Helper {
   static valueFromRemoteObject(remoteObject) {
     assert(!remoteObject.objectId, 'Cannot extract value when objectId is given');
     if (remoteObject.unserializableValue) {
+      if (remoteObject.type === 'bigint')
+        return eval(remoteObject.unserializableValue);
```
JoelEinbinder: This is very scary. We shouldn't execute any code from the browser in the node process.

JoelEinbinder: You can do BigInt("str") instead.

JoelEinbinder:
Looks cool! Can we add a test?

lib/helper.js
vsemozhetbyt: Yeah, I was not sure about that, But there is a small issue: protocol sends the value like `'42n'` and `BigInt()` can parse only `BigInt('42')`, so we will need some more string preprocessing (trim the `'n'`). And my concern here was a bit alleviated by the check for `remoteObject.type`: if it is `bigint` it seems rather safe value for `eval` — unless the protocol can be hacked?

JoelEinbinder: I would consider all strings coming over the protocol to be unsafe. Otherwise an exploit in chrome’s content process would be a remote execution exploit in puppeteer.

JoelEinbinder: I would consider all strings coming over the protocol to be unsafe. Otherwise an exploit in chrome’s content process would be a remote execution exploit in puppeteer.

vsemozhetbyt:
> Can we add a test?

I was not able to find any tests for `unserializableValue`s to add a new case, so it seems a new test should be added. However, I am not aware of test tools and infrastructure in `puppeteer`, so this will take some learning time. If anybody can quickly push a test to the branch, please do. Otherwise, I will try to add one later.

lib/helper.js
vsemozhetbyt: Addressed.

vsemozhetbyt: Oops. It seems this need more care for the old Node.js versions and ESlint because of missed `BigInt` global. Sorry, will try to fix.

vsemozhetbyt: Addressed. Sorry, I do not know TypeScript. What should be added to pass TS checks?

vsemozhetbyt:
Tests hopefully added. TS checks still are to be fixed.

vsemozhetbyt:
I've ventured to up `"target"` to `"ESNext"` in `tsconfig.json`. Let me know if this is overkill.

vsemozhetbyt:
Is there anything else I should improve?

vsemozhetbyt:
@aslushnikov @JoelEinbinder Should I improve something else?

aslushnikov (approved):
Oh this is awesome. Thanks a lot!

Mafia-777:
38qopLNuY5QZ9PRf2HA6Ubso4zFiqFS6hv
IMG_2555

PR #4017, opened 2019-02-15
fix: check if async error has a stack

JoelEinbinder:
If a non-error object or an error without a stack was thrown, the async stack trace could would fail.


aslushnikov:
Can you please add the same thing to puppeteer-firefox's helper?

aslushnikov (approved):

Issue #2278, opened 2018-03-28
Headless PDF printing inconsistent page width and height


Jarrydcrawford:
I'm currently encountering some potential rounding issues with the PDF page size where I have to use an additional few 10ths of a millimetre on the width and height for the pages, the `@page`s declaration in my CSS, and also the CSS height of my `section`s to render (mostly) correctly.

Even after adding the extra mm to cover the issue, the width then increases by about 1-2mm on either side of the page, only when Puppeteer is used to generate the PDF, but not when printed as PDF directly from my local Chrome (65.0.3325.181).

The removal of the additional mm breaks the layout in both local Chrome and puppeteer rendered PDFs.

Possibly similar issue to #666.

**Tell us about your environment:**

* Puppeteer version: 1.2.0
* Platform / OS version: MacOS 10.13.3
* URLs (if applicable): https://gist.github.com/Jarrydcrawford/9b64fc289e58d990aa4a8990331538e4
* Node.js version: 8.10.0

**What steps will reproduce the problem?**

```js
const puppeteer = require('puppeteer');
const fs = require('fs');

const DELTA = 0.27; // closest additional mm to not break layout

(async () => {
  let browser;
  try {
    browser = await puppeteer.launch({
      headless: true,
      args: [
        '--no-sandbox',
        '--disable-setuid-sandbox',
        '--disable-gpu',
        '--hide-scrollbars',
        '--disable-web-security',
      ],
    });
    const page = await browser.newPage();
    await page.goto(
      `data:text/html,${fs.readFileSync('index.html').toString('utf8')}`,
    );
    await page.pdf({
      path: './index.pdf',
      displayHeaderFooter: false,
      printBackground: true,
      pageRanges: '1-2',
      height: 139 + DELTA + 3 + 3 + 'mm', // Additional page margins of 3mm
      width: 550 + DELTA + 3 + 3 + 'mm', // Additional page margins of 3mm
      margin: {
        top: 0,
        right: 0,
        bottom: 0,
        left: 0,
      },
    });
    await browser.close();
  } catch (err) {
    if (browser) {
      await browser.close();
    }
    throw err;
  }
})();
```

1. `node index.js`.
2. Open PDF.
3. Open HTML file directly in Chrome browser.
4. Print as PDF.
5. Compare (may have to zoom in to see).

**What is the expected result?**

PDF should not have any margin/whitespace on the page edges.

**What happens instead?**

There is a horizontal page gap on either side of each page of about 1mm.
[...]

PR #4021, opened 2019-02-15
fix(page): add screen dpi property to page.pdf method

### LINKED ISSUE #666: PDF creation pixel sizes causes layout to change
Hi,

When creating PDFs from a responsive web page, lets say the page has a css rule to show a narrower version below `1024` pixels screen width and we supply `width: "1024px"` to `pdf` function. The rule should not apply on this case however it does and the layout changes in generated pdf. 

Only when the width parameter is **4/3** of the actual width of viewport, then the pdf has exact same layout of given viewport width. This seems to be something related to an inches / points confusion somewhere (96 pixels per inch, 72 pixels per point. 96/72 = 4/3)

So, when I provide `1365px` as pdf width I get the exact representation of the page at `1024px` viewport width. Anything below `1365px`, page acts like the viewport is below `1024px` and applies the appropriate responsive css rules.

I'm not sure if this is an upstream bug or something else though. Margins are set to 0 btw. 

Any ideas?

### LINKED ISSUE #3098: Pdf width is smaller than it should be
### Steps to reproduce
There seems to be an issue with the way the function page.pdf interprets the width of the page. Eventually using the design of the media query below the current one and I can't seem to know what's going on here.

* Puppeteer version: 1.5.0
* Platform / OS version: Mac
* URLs (if applicable): http://stephencaver.com/
* Node.js version: 8.11.3

To reproduce the issue, you can try a page.pdf with this page http://stephencaver.com/, with a width of 902 and an height of 2000 (height does not matter here).
You'll see that the design display a two rows website demo instead of three, which is the rule for the media-query below 900 pixels.

```
import fs from 'fs';
import express from 'express';
import bodyParser from 'body-parser';
import { join } from 'path';
import mkdirp from 'mkdirp';

const port = '8800';

/**
 * Creates and returns a configured express server.
 */

function createExpressServer() {
  let app = express();
  app.use(bodyParser.json({ limit: '50mb' }));
  app.use(bodyParser.urlencoded({ limit: '50mb', extended: true }));

  const wait = (timer) => new Promise(resolve => setTimeout(() => resolve(), timer));

  app.post('/createPDF/', async (req, res) => {
    const startTime = Date.now();
    console.log("Received request to render page: %s", req.body.url);

    // Launch Puppeteer
    const uuid = uuidv4();

    await page.emulateMedia('screen');

    // Go to the page and wait for all the connection on the page to be resolved
    await page.goto(`${req.body.url}`, {waitUntil: 'networkidle0'});

    // Send the response
    switch(req.body.format) {
      case 'pdf':
        await page.pdf({
          path: join(__dirname, PDF_EVIDENCE_FD, `${uuid}.pdf`),
          height: pageHeight,
          width: pageWidth,
          printBackground: true
        });

        res.status(200).sendFile(`${uuid}.pdf`);
        break;

      default:
        res.status(501).send({ error: 'error'});
        break;
    }
  });

  app.on("listening", function() {
    console.log("Listening on port %d", app.port);
  });
  return app;
}

let appInstance = createExpressServer();

appInstance.listen(port);
```

A three row demo section.

**What happens instead?**
A one row below section.

ggevgg:
This patch adds the dpi property to the page.pdf method to fix pdf output resolution

Fixes #3098, Fixes #666

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1514,6 +1514,7 @@ Page is guaranteed to have a main frame which persists during navigations.
     - `bottom` <[string]|[number]> Bottom margin, accepts values labeled with units.
     - `left` <[string]|[number]> Left margin, accepts values labeled with units.
   - `preferCSSPageSize` <[boolean]> Give any CSS `@page` size declared in the page priority over what is declared in `width` and `height` or `format` options. Defaults to `false`, which will scale the content to fit the paper size.
+  - `dpi` <[number]> Screen dpi. Defaults to 96.
 - returns: <[Promise]<[Buffer]>> Promise which resolves with PDF buffer.
 
 > **NOTE** Generating a pdf is currently only supported in Chrome headless.
--- a/docs/api.md
+++ b/docs/api.md
@@ -1514,7 +1514,7 @@ Page is guaranteed to have a main frame which persists during navigations.
     - `bottom` <[string]|[number]> Bottom margin, accepts values labeled with units.
     - `left` <[string]|[number]> Left margin, accepts values labeled with units.
   - `preferCSSPageSize` <[boolean]> Give any CSS `@page` size declared in the page priority over what is declared in `width` and `height` or `format` options. Defaults to `false`, which will scale the content to fit the paper size.
-  - `dpi` <[number]> Screen dpi. Defaults to 96.
+  - `dpi` <[number]> Screen DPI. Defaults to 96.
 - returns: <[Promise]<[Buffer]>> Promise which resolves with PDF buffer.
 
 > **NOTE** Generating a pdf is currently only supported in Chrome headless.
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -936,7 +936,8 @@ class Page extends EventEmitter {
       pageRanges = '',
       preferCSSPageSize = false,
       margin = {},
-      path = null
+      path = null,
+      dpi = 96
     } = options;
 
     let paperWidth = 8.5;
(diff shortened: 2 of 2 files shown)


docs/api.md
```diff
@@ -1514,6 +1514,7 @@ Page is guaranteed to have a main frame which persists during navigations.
     - `bottom` <[string]|[number]> Bottom margin, accepts values labeled with units.
     - `left` <[string]|[number]> Left margin, accepts values labeled with units.
   - `preferCSSPageSize` <[boolean]> Give any CSS `@page` size declared in the page priority over what is declared in `width` and `height` or `format` options. Defaults to `false`, which will scale the content to fit the paper size.
+  - `dpi` <[number]> Screen dpi. Defaults to 96.
```
vsemozhetbyt: Maybe:
```suggestion
  - `dpi` <[number]> Screen DPI. Defaults to 96.
```

vsemozhetbyt:
Could these scary messages on approved suggestions be somehow addressed by the bot creators?

JoelEinbinder:
Code looks reasonable, but I am not that familiar with the topic. How do I know what dpi to pass into this function? If it is supposed to match something in the website/chrome, we should be querying for it instead.

ggevgg:
> Code looks reasonable, but I am not that familiar with the topic. How do I know what dpi to pass into this function? If it is supposed to match something in the website/chrome, we should be querying for it instead.

In most cases, Chrome correctly detects dpi, and you do not need to change the default dpi. But when the PDF resolution is not as expected, we need the ability to change this property.

JoelEinbinder:
It looks like I can get the dpi for a web page via a script like this:
```js
const dpi = await page.evaluate(() => {
  const div = document.createElement('div');
  div.style.position = 'absolute';
  div.style.top = 0;
  div.style.left = 0;
  div.style.width = '1in';
  div.style.height = '1in';
  document.body.appendChild(div);
  const height = div.offsetHeight;
  div.remove();
  return height;
});
```

Should the DPI value that the user passes into this method always match the result of this function?

ggevgg:
This approach has some minuses:
1. it isn't quite clean - it affects original DOM
2. our new DIV element can be affected by styles of original DOM and we can receive incorrect result

We have to provide a manual correction of the coefficient for the application flexibility. In some cases we need to use particular dpi and it may be different from the system one
Also we have checked this approach in our container. Result is the same as default value (equals 96), but isn't fit to us (manual dpi equals 72)

kblok:
Just my two cents here.
I'm trying to answer [this question on StackOverflow](https://stackoverflow.com/questions/54784768/how-to-generate-pdf-of-exactly-specified-part-of-the-page-with-puppeter). I pulled this branch and used @JoelEinbinder code to determine my DPI. It was 96.

It seems that this solution isn't pixel perfect.
image

clickyclick:
This wont fix the two bugs linked since the bugs occur even if you pass in inches.

aslushnikov:
@ggevgg Spring cleanup! Sorry for this hanging for so long.

I don't think this code fixes the related bugs. Most of the pixel issues we have in PDF printing are related to the inherent unit translation on the chromium side; the fix should be on the chromium side as well.

> In most cases, Chrome correctly detects dpi, and you do not need to change the default dpi. But when the PDF resolution is not as expected, we need the ability to change this property.

I don't actually understand this. What are the cases?

cliedelt:
> I don't think this code fixes the related bugs. Most of the pixel issues we have in PDF printing are related to the inherent unit translation on the chromium side; the fix should be on the chromium side as well.

are you able to fix this? @ggevgg @aslushnikov  its one of the highest rated bugs 😞 #2278

aslushnikov:
> are you able to fix this?

@orgertot not yet.

Closing PR since I don't think it addresses the issue.

PR #4048, opened 2019-02-21
test: use require('puppeteer') instead of utils.requireRoot

aslushnikov:
Add a dev-dependency "puppeteer" that points to "self" to use
`require('puppeteer')` instead of a `utils.requireRoot`.

NOTE: This also eliminates the need for `NODE_PATH` when running
Puppeteer examples.


aslushnikov:
This approach not gonna work with the way we handle node6 for now. We'll get back to it later once node6 is off the table.

PR #4095, opened 2019-03-01
fix(chromium): bypass service workers when request interception is on

### LINKED ISSUE #4041: Requests are not intercepted when cached by Service Worker
Not sure if this is by design but requests that were previously cached by service worker are not intercepted after `page.setRequestInterception(true)`.

Simplified code to reproduce:
```js
const hostname = 'http://localhost:';
const {port, proc} = await startServer(); // local implementation
  args: [
    '--no-sandbox',
    '--disable-setuid-sandbox',
    '--enable-features=NetworkService',
  ],
  ignoreHTTPSErrors: true,
});

await page.goto(`${hostname}${port}`);
await page.evaluate('navigator.serviceWorker.ready');

await page.setRequestInterception(true); //toggle this
page.on('request', req => {
   console.log('requesting', req.url());
   req.continue().catch(e => e /* not intercepting */);
});

await page.reload({waitUntil: 'domcontentloaded'});

await browser.close();
```

1. Set `await page.setRequestInterception(true)`
2. Run test, only URLs not previously cached by service worker are logged
3. Set `await page.setRequestInterception(false)` (or remove Service Worker registration)
4. Run test, all URLs are logged

Are requests intercepted only when requests are sent through to the network, or is this a bug?

aslushnikov:
Puppeteer currently relies on invariant that once request interception
is enabled in DevTools protocol, every `Network.requestWillBeSent` event
is accompanied with a corresponding `Network.requestIntercepted`.

However, if a request is fulfilled by ServiceWorker, protocol will not
emit the `Network.requestIntercepted` event, and the invariant breaks.
As a result, Puppeteer never dispatches a `request` event.

The proper solution would be connecting to all the service workers and
adding an instrumentation to designate that request was fulfilled from
worker.

The compromised solution is to bypass service workers for all network
requests - and that's what we do here.

Fix #4041

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1758,6 +1758,7 @@ puppeteer.launch().then(async browser => {
 ```
 
 > **NOTE** Enabling request interception disables page caching.
+> **NOTE** Enabling request interception starts bypassing ServiceWorkers for network.
 
 #### page.setUserAgent(userAgent)
 - `userAgent` <[string]> Specific user agent to use in this page
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -128,6 +128,7 @@ class NetworkManager extends EventEmitter {
     const patterns = enabled ? [{urlPattern: '*'}] : [];
     await Promise.all([
       this._client.send('Network.setCacheDisabled', {cacheDisabled: enabled}),
+      this._client.send('Network.setBypassServiceWorker', {bypass: enabled}),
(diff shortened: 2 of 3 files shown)


JoelEinbinder:
Fails the lint. And seems controversial. Won't this effect the functionality of some pages?

aslushnikov:
Not landing since it's controversial.

PR #4123, opened 2019-03-05
test: fix fixtures test when run with env DUMPIO=1

aslushnikov:
The DUMPIO env variable is propagated to a spawned process
and results in unfortunate stdout.


JoelEinbinder (approved):

JoelEinbinder:
Looks like you need to bump the eslint version or use Object.assign.

Issue #4163, opened 2019-03-13
Unable to get content of iframe inside shadow dom


dmshkv:
Hey, developers!
I'm trying to select elements of iframe, that is inside of shadow root of Polymer component, but no result. Have to mention that the page is the opened popup from the main app.

So I'm taking the popup's page like this and trying to iterate its 
```
const pages = await browser.pages();
const popup = pages[pages.length - 1]; // last opened page, verified by url
const frames = popup.frames(); // this returns the wrong number of iframes inside the popup's page
frames.map(async (f) => {
    const itsUrl = await f.url();
    console.log(itsUrl);
}); // and there is no iframe from shadow root of polymer component
```

Also one of the used approaches
```
const pages = await browser.pages();
const popup = pages[pages.length - 1];
const shadowRoot = await page.evaluateHandle("document.querySelector('my-polymer-component').shadowRoot");
const some = await shadowRoot.$("body"); // this returns null
```

Can you please help in finding the proper solution for this?

dmshkv:
Hey, @aslushnikov! Would you be so kind to take a look at the issue?

aslushnikov:
hey @dmshkv! Frames are reported just fine from-inside shadow DOM. One thing I can think of is that the page hasn't loaded yet, so the iframe is not inserted.

It's hard to say what's going on here for sure - since there's not enough information for me to reproduce this locally. I'd try waiting for the [`frameattached`](https://pptr.dev/#?product=Puppeteer&version=v1.13.0&show=api-event-frameattached) event. 

If you don't see the frame reported by Puppeteer no matter what - than it's likely that you have an OOPIF and than this is a dupe of #2548.

dmshkv:
Thanks, Andrey!

PR #4167, opened 2019-03-15
test: make sure frames are reported from-inside shadow DOM

aslushnikov:
References #4163.

### DIFF
--- a/test/frame.spec.js
+++ b/test/frame.spec.js
@@ -162,6 +162,17 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(detachedFrames.length).toBe(4);
       expect(navigatedFrames.length).toBe(1);
     });
+    it('should report frame from-inside shadow DOM', async({page, server}) => {
+      await page.goto(server.PREFIX + '/shadow.html');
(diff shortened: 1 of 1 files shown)


test/frame.spec.js
```diff
@@ -162,6 +162,17 @@ module.exports.addTests = function({testRunner, expect}) {
...
     });
+    it('should report frame from-inside shadow DOM', async({page, server}) => {
+      await page.goto(server.PREFIX + '/shadow.html');
+      await page.evaluate(async url => {
+        const frame = document.createElement('iframe');
+        frame.src = url;
+        document.body.shadowRoot.appendChild(frame);
+        await new Promise(x => frame.onload = x);
+      }, server.EMPTY_PAGE);
+      expect(page.frames().length).toBe(2);
```
JoelEinbinder, line 173: I think this might be flaky if you don’t still wait for the frameattached event

aslushnikov: shouldn't be - we await the "load" event on the iframe which happens way after the frameattached protocol event.

JoelEinbinder (approved):

PR #4225, opened 2019-03-29
feat: cookies-if no args, returns all browser cookies for current URL

andromeda-galaxy:
As described in:
https://chromedevtools.github.io/devtools-protocol/tot/Network#method-getCookies
Network.getCookies, returns all browser cookies for the current page.
Cookies method return just current URL cookies if no args set.  It's better that cookies method behave like getCookies, so if no args passed to it, it returns all browser cookies related to the current page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 <img src="https://user-images.githubusercontent.com/10379601/29446482-04f7036a-841f-11e7-9872-91d1fc2ea683.png" height="200" align="right">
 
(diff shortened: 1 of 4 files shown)


andromeda-galaxy:
> I signed it!

aslushnikov:
This has been discussed a while ago; we're waiting on #645.

Once upstream work is done, we'll provide all the methods necessary to fetching cookies per browser context.

Issue #4162, opened 2019-03-13
Failed to connect Puppeteer to existing Firefox instance run by web-ext


denar90:
Hi, 
At @treosh we are working on analyzing extensions (ext) impact on real user performance. 
We try to install extensions for Firefox (FF) and handle browser with Puppeteer (PP) to do measurements and other stuff, but struggling with some sort of issues.

**Tell us about your environment:**

* Puppeteer version: `"puppeteer-firefox": "^0.5.0"`

* Platform / OS version: macOS Mojave 10.14.3 

* Node.js version: v11.1.0

**What steps will reproduce the problem?**

There are several cases we tried and all of them based on using [web-ext](https://github.com/mozilla/web-ext) library for installing ext for FF.

----

### Case 1: Connect FF instance run by webExt to PP

**Code simple**:

```
    webExt.util.logger.consoleStream.makeVerbose();

    extensionRunners = await webExt.cmd.run({
      sourceDir: extPath,
      // comment if connect to default FF
      firefox: PP_FF.executablePath(),
    }, {
      // These are non CLI related options for each function.
      // You need to specify this one so that your NodeJS application
      // can continue running after web-ext is finished.
      shouldExitProgram: false,
    })

    const extensionRunner = extensionRunners.extensionRunners[0]

    const browserWSEndpoint = `ws://127.0.0.1:${extensionRunner.remoteFirefox.client.client.client.remotePort}`
    browser = await PP_FF.connect({
      browserWSEndpoint,
    })
```

**Error**:

```
Error: Parse Error
  at Socket.socketOnData (_http_client.js:449:20)
  at Socket.emit (events.js:182:13)
  at addChunk (_stream_readable.js:283:12)
  at readableAddChunk (_stream_readable.js:264:11)
  at Socket.Readable.push (_stream_readable.js:219:10)
  at TCP.onStreamRead (internal/stream_base_commons.js:122:17) bytesParsed: 0, code: 'HPE_INVALID_CONSTANT' } 
```
> Fairly saing it crashes in PP helper [line 112](https://github.com/GoogleChrome/puppeteer/blob/v1.13.0/lib/helper.js#L112), but after commenting it - error HPE_INVALID_CONSTANT can be seen.

---

### Case 2: Connect webExt to FF instance run by PP
> install web ext from built fork https://github.com/denar90/web-ext-firefox-port instead of web-ext original

**Code simple**:

```
    browser = await PP_FF.launch({
      headless: false,
    })

[...]

Issue #3660, opened 2018-12-12
Firefox: Loading extension ignored


denar90:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.11.0
* Platform / OS version: MacOS
* Node.js version:v9.5.0

**What steps will reproduce the problem?**

Using example from [docs](https://github.com/GoogleChrome/puppeteer/blob/v1.10.0/docs/api.md#working-with-chrome-extensions) and any extension from store, just downloading and unpack it for FF and Chrome.

**What is the expected result?**

If put breakpoint on `await browser.close();` and look at Chrome browser, installation process can be seen. 

**What happens instead?**

In FF case, just empty page no matter how long to wait.

I suppose FF has different mechanism for addons and installing them. Probably it's just docs issue and missed some flag or something bigger and related to security stuff...

aslushnikov:
> Using example from docs and any extension from store, just downloading and unpack it for FF and Chrome.

@denar90 The documentation at [`//docs/api.md`](https://github.com/GoogleChrome/puppeteer/blob/v1.10.0/docs/api.md#working-with-chrome-extensions) covers Chromium Puppeteer. we'll come up with a similar write-up for Puppeteer-Firefox later on as the project matures.

denar90:
Thanks :)

PR #4228, opened 2019-04-01
docs(puppeteer-firefox/readme.md): adds add-ons

denar90:
Adds documentation for Add-ons installation for Firefox

#3660 #4162


denar90:
@aslushnikov is it ok that temporary users will have to use my fork instead of web-ext@3.1? There's a lot of work for 3.1 and I'm not sure that it will be finished soon.

denar90:
We agreed with Andrey that better to wait for web-ext@3.1 to not confuse anyone.

PR #4234, opened 2019-04-03
docs: add table-of-contents to troubleshooting

aslushnikov:
Drive-by: teach our table-of-contents generator to ignore comments
inside fenced blocks and to de-linkify titles.

### DIFF
--- a/utils/doclint/preprocessor/test.js
+++ b/utils/doclint/preprocessor/test.js
@@ -156,6 +156,42 @@ describe('runCommands', function() {
         #### page.$
         #### page.$$`);
     });
+    it('should work with code blocks', () => {
+      const source = new Source('doc.md', `<!-- gen:toc -->XXX<!-- gen:stop -->
+        ### class: page
+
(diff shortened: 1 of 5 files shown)


utils/doclint/preprocessor/test.js
```diff
@@ -156,6 +156,42 @@ describe('runCommands', function() {
         #### page.$
         #### page.$$`);
     });
+    it('should work with code blocks', () => {
+      const source = new Source('doc.md', `<!-- gen:toc -->XXX<!-- gen:stop -->
+        ### class: page
```
JoelEinbinder, line 161: nit: remove tabs here and below.

JoelEinbinder (approved):

utils/doclint/preprocessor/test.js
aslushnikov: that's aligned with other tests :-) if you really want, I'll follow up for them all!

PR #4240, opened 2019-04-04
feat(chromium): roll Chromium to r647475

aslushnikov:
This roll includes:
- https://crrev.com/1549884 - DevTools: disable interception for
  data: URLs

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "node": ">=6.4.0"
   },
   "puppeteer": {
-    "chromium_revision": "641577"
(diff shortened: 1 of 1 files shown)


JoelEinbinder (approved):

JoelEinbinder:
Did we have docs about this that need to change?

aslushnikov:
@JoelEinbinder no, that's the way it is atm. There was an unfortunate regression in the protocol that prevented the roll; @caseq's patch fixed it.

PR #4248, opened 2019-04-05
fix: Use new requestId for interception request matching

### LINKED ISSUE #3471: [umbrella] Fix Request Interception
Umbrella bug for all request interception issues in Puppeteer.
This mostly depends on upstream Chromium bug: https://crbug.com/899386

#### Description

DevTools Protocol emits two events for every intercepted request:
- [`Network.requestWillBeSent(requestId)`](https://vanilla.aslushnikov.com/#Network.requestWillBeSent)
- [`Network.requestIntercepted(interceptionId)`](https://vanilla.aslushnikov.com/#Network.requestIntercepted)

Puppeteer extracts different pieces of information from these events. However, since `interceptionId` and `requestId` aren't matching, Puppeteer uses [heuristics](https://github.com/GoogleChrome/puppeteer/blob/9800b2c3c2d31379badb9d4dab1cc8fc95aee48a/lib/NetworkManager.js#L640-L672) to match these events together. This is necessary to provide a consistent API to our clients.

Unfortunately, these heuristics aren't reliable and break under different corner cases: #1664, #2848, #3030, #3369 and so on.

#### Mitigation

We weren't able to fix this upstream up until today due to the network stack refactoring in Chromium. Now, with Network Service getting gradually adopted, we can fix this upstream.

The plan is:
1. Enable Network Service by default to Puppeteer users
2. Fix network instrumentation in DevTools protocol for the Network Service and migrate onto it. (see https://crbug.com/899386).

josepharhar:
Fix #3471

### DIFF
--- a/lib/NetworkManager.js
+++ b/lib/NetworkManager.js
@@ -16,7 +16,6 @@
 const EventEmitter = require('events');
 const {helper, assert, debugError} = require('./helper');
 const {Events} = require('./Events');
-const Multimap = require('./Multimap');
 
 class NetworkManager extends EventEmitter {
   /**
@@ -41,10 +40,8 @@ class NetworkManager extends EventEmitter {
     this._attemptedAuthentications = new Set();
     this._userRequestInterceptionEnabled = false;
     this._protocolRequestInterceptionEnabled = false;
-    /** @type {!Multimap<string, string>} */
-    this._requestHashToRequestIds = new Multimap();
-    /** @type {!Multimap<string, string>} */
-    this._requestHashToInterceptionIds = new Multimap();
+    /** @type {!Map<string, string>} */
+    this._requestIdToInterceptionId = new Map();
 
     this._client.on('Network.requestWillBeSent', this._onRequestWillBeSent.bind(this));
     this._client.on('Network.requestIntercepted', this._onRequestIntercepted.bind(this));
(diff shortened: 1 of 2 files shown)


lib/NetworkManager.js
```diff
@@ -178,15 +174,13 @@ class NetworkManager extends EventEmitter {
...
     }
 
-    const requestHash = generateRequestHash(event.request);
-    const requestId = this._requestHashToRequestIds.firstValue(requestHash);
-    if (requestId) {
+    const requestId = event.requestId;
+    if (requestId && this._requestIdToRequestWillBeSentEvent.has(requestId)) {
       const requestWillBeSentEvent = this._requestIdToRequestWillBeSentEvent.get(requestId);
       this._onRequest(requestWillBeSentEvent, event.interceptionId);
-      this._requestHashToRequestIds.delete(requestHash, requestId);
```
aslushnikov, line 186:
```js
this._requestIdToInterceptionId.delete(requestId);
```

lib/NetworkManager.js
```diff
@@ -178,15 +174,13 @@ class NetworkManager extends EventEmitter {
       }).catch(debugError);
     }
 
-    const requestHash = generateRequestHash(event.request);
-    const requestId = this._requestHashToRequestIds.firstValue(requestHash);
-    if (requestId) {
+    const requestId = event.requestId;
+    if (requestId && this._requestIdToRequestWillBeSentEvent.has(requestId)) {
```
aslushnikov, line 178: how can there be no `requestId`?

josepharhar: Browser initiated requests which we don't have instrumentation for, like the opensearch thing, don't have a devtools request id in the mojo request struct and therefore don't get a requestId in the requestIntercepted event

lib/NetworkManager.js
josepharhar: The requestId gets deleted out of this._requestIdToInterceptionId in _onRequestWillBeSent. If I put another .delete here it would never do anything because this branch is for when we already had a requestWillBeSent and _requestIdToInterceptionId is not used.

aslushnikov (approved):

PR #4260, opened 2019-04-09
fix: cache disabling should stick when toggling request interception

aslushnikov:
This patch:
- refactors `NetworkManager`/`FrameManager` so that they enable all the
relevant domains themselves. This is a preparation for OOPIF support and
migration onto fetch domain.
- stops enabling Security domain: it saves quite some traffic over
websocket since it no longer sends annoying "SecurityStateChanged" events.
Instead, use `Security.setIgnoreCertificateErrors` method.
- consolidates network cache state in network manager. This even fixes a
bug with caching and request interception interop.

### DIFF
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -20,6 +20,7 @@ const {Events} = require('./Events');
 const {ExecutionContext, EVALUATION_SCRIPT_URL} = require('./ExecutionContext');
 const {LifecycleWatcher} = require('./LifecycleWatcher');
 const {DOMWorld} = require('./DOMWorld');
+const {NetworkManager} = require('./NetworkManager');
 
 const UTILITY_WORLD_NAME = '__puppeteer_utility_world__';
 
@@ -27,14 +28,15 @@ class FrameManager extends EventEmitter {
   /**
    * @param {!Puppeteer.CDPSession} client
    * @param {!Puppeteer.Page} page
-   * @param {!Puppeteer.NetworkManager} networkManager
+   * @param {boolean} ignoreHTTPSErrors
    * @param {!Puppeteer.TimeoutSettings} timeoutSettings
    */
-  constructor(client, page, networkManager, timeoutSettings) {
+  constructor(client, page, ignoreHTTPSErrors, timeoutSettings) {
     super();
     this._client = client;
     this._page = page;
-    this._networkManager = networkManager;
+    this._networkManager = new NetworkManager(client, ignoreHTTPSErrors);
+    this._networkManager.setFrameManager(this);
     this._timeoutSettings = timeoutSettings;
     /** @type {!Map<string, !Frame>} */
     this._frames = new Map();
@@ -63,7 +65,15 @@ class FrameManager extends EventEmitter {
     await Promise.all([
       this._client.send('Page.setLifecycleEventsEnabled', { enabled: true }),
       this._client.send('Runtime.enable', {}).then(() => this._ensureIsolatedWorld(UTILITY_WORLD_NAME)),
+      this._networkManager.initialize(),
     ]);
+  }
+
+  /**
+   * @return {!NetworkManager}
+   */
+  networkManager() {
+    return this._networkManager;
   }
 
   /**
--- a/lib/FrameManager.js
+++ b/lib/FrameManager.js
@@ -26,12 +26,11 @@ const UTILITY_WORLD_NAME = '__puppeteer_utility_world__';
 class FrameManager extends EventEmitter {
   /**
    * @param {!Puppeteer.CDPSession} client
-   * @param {!Protocol.Page.FrameTree} frameTree
    * @param {!Puppeteer.Page} page
    * @param {!Puppeteer.NetworkManager} networkManager
    * @param {!Puppeteer.TimeoutSettings} timeoutSettings
    */
-  constructor(client, frameTree, page, networkManager, timeoutSettings) {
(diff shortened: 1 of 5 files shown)


lib/Page.js
```diff
@@ -43,36 +43,26 @@ class Page extends EventEmitter {
...
-      client.send('Page.setLifecycleEventsEnabled', { enabled: true }),
-      client.send('Network.enable', {}),
-      client.send('Runtime.enable', {}).then(() => page._frameManager.ensureSecondaryDOMWorld()),
-      client.send('Security.enable', {}),
       client.send('Performance.enable', {}),
       client.send('Log.enable', {}),
     ]);
-    if (ignoreHTTPSErrors)
-      await client.send('Security.setOverrideCertificateErrors', {override: true});
-    // Initialize default page size.
```
JoelEinbinder, line 61: The comment got killed! I guess that’s ok

lib/Page.js
```diff
@@ -43,36 +43,26 @@ class Page extends EventEmitter {
    * @return {!Promise<!Page>}
    */
   static async create(client, target, ignoreHTTPSErrors, defaultViewport, screenshotTaskQueue) {
-
-    await client.send('Page.enable');
-    const {frameTree} = await client.send('Page.getFrameTree');
-    const page = new Page(client, target, frameTree, ignoreHTTPSErrors, screenshotTaskQueue);
+    const page = new Page(client, target, ignoreHTTPSErrors, screenshotTaskQueue);
     await Promise.all([
+      page._networkManager.initialize(),
```
JoelEinbinder: I have to think a lot about this. The frame manager can initialize before the network manager, and it has a network manager that it doesn’t know is uninitialized. I believe that it works, but it’s hard to reason about.

lib/FrameManager.js
```diff
@@ -53,7 +52,18 @@ class FrameManager extends EventEmitter {
...
     this._client.on('Runtime.executionContextsCleared', event => this._onExecutionContextsCleared());
     this._client.on('Page.lifecycleEvent', event => this._onLifecycleEvent(event));
+  }
+
+  async initialize() {
+    const [,{frameTree}] = await Promise.all([
+      this._client.send('Page.enable'),
+      this._client.send('Page.getFrameTree'),
+    ]);
     this._handleFrameTree(frameTree);
```
JoelEinbinder, line 64: We can get frame events in between page Page.getFrameTree and handleFrameTree. What happens to them?

lib/Page.js
aslushnikov: > The frame manager can initialize before the network manager, and it has a network manager that it doesn’t know is uninitialized. I believe that it works, but it’s hard to reason about.

The following observations should help with reasoning:
- all state that depends on events is encapsulated inside the related manager.
- `page` instance is not exposed to a userland at this point - so all missed events is a natural race.

Also, there's no semantic change here: the initialization sequence maintains the same contracts as before. So if there's some confidence in the tip-of-tree code, it should be inheritable here.

lib/FrameManager.js
aslushnikov: > We can get frame events in between page Page.getFrameTree and handleFrameTree. What happens to them?

This is the unfortunate consequence of the current protocol design. The race is already present in the tip of tree; this patch shortens the gap a bit.

lib/LifecycleWatcher.js
```diff
@@ -48,7 +47,7 @@ class LifecycleWatcher {
       helper.addEventListener(this._frameManager, Events.FrameManager.LifecycleEvent, this._checkLifecycleComplete.bind(this)),
       helper.addEventListener(this._frameManager, Events.FrameManager.FrameNavigatedWithinDocument, this._navigatedWithinDocument.bind(this)),
       helper.addEventListener(this._frameManager, Events.FrameManager.FrameDetached, this._onFrameDetached.bind(this)),
-      helper.addEventListener(this._networkManager, Events.NetworkManager.Request, this._onRequest.bind(this)),
+      helper.addEventListener(this._frameManager.networkManager(), Events.NetworkManager.Request, this._onRequest.bind(this)),
```
JoelEinbinder, line 50: looks way better

lib/Page.js
```diff
@@ -280,18 +265,6 @@ class Page extends EventEmitter {
     this._timeoutSettings.setDefaultTimeout(timeout);
   }
 
-  /**
-   * @param {!Protocol.Security.certificateErrorPayload} event
-   */
-  _onCertificateError(event) {
-    if (!this._ignoreHTTPSErrors)
-      return;
-    this._client.send('Security.handleCertificateError', {
```
JoelEinbinder: Where did this go?

lib/Page.js
```diff
@@ -449,21 +422,21 @@ class Page extends EventEmitter {
...
   async authenticate(credentials) {
-    return this._networkManager.authenticate(credentials);
+    return this._frameManager.networkManager().authenticate(credentials);
   }
 
   /**
    * @param {!Object<string, string>} headers
    */
   async setExtraHTTPHeaders(headers) {
-    return this._networkManager.setExtraHTTPHeaders(headers);
```
JoelEinbinder, line 459: How do you feel about adding `this._networkManager = this._frameMangaer.networkManager()`?

JoelEinbinder (approved):

lib/Page.js
aslushnikov: Didn't do it to stress the ownership

Issue #2242, opened 2018-03-22
Chrome remote debugging expects host header


grantstephens:
This is a slightly weird one and I don't think it is strictly speaking a puppeteer bug but thought I'd raise it here anyway.
Since version 1.2, when trying to connect to chrome via the remote debugging interface a host header must be specified.
This is not a problem in puppeteer per se, but we use python to get the ws address and so the python requests line now has to include `headers={'Host':''}` otherwise chrome returns an error: `Host header is specified and is not an IP address or localhost.`
I assume that puppeteer is setting this header in the background somewhere or chrome doesn't care anymore once the request comes in via the websocket. I also could not find anything documented about this change, but I'm assuming that is has something to do with the .dev domain changes to HTTPS.

gios:
I was testing this bug with previous puppeteer version and had the same error. Maybe this bug is related to chrome because I used v67 chrome. With v66 chrome everything is OK.

ermyril:
Got the same error with a Chromium 67.0.3383.0 while trying to get a docker-container to work with my laravel application, `curl {host with chrome-headless}` gives me that

Will try to get version 66 by now

aslushnikov:
@grantstephens thanks for reaching to us.

This was recently introduced in https://chromium-review.googlesource.com/c/chromium/src/+/952522. The error is issued if the host header is set to non-localhost and non-IP address.

What's your python library? It looks like it sets a non-trivial host header, so you have to override it with empty value.

cc @pavelfeldman

grantstephens:
I'm using requests, which AFAIK does not send a host header by default, but I may be wrong. Either way- couldn't find any changelog for it suddenly happening (From the chrome side) so thought I'd just put it here in case anybody else was having a similar issue.
Maybe adding something in the documentation about it?

hbakhtiyor:
and so what's the workaround of the issue?

bean5:
I started getting this when I upgraded from version 65 to 66.

bean5:
"The error is issued if the host header is set to non-localhost and non-IP address." Does this imply that the host header should use an IP address? Which header? Is this an error in a websites configuration or an error in puppeteer?

cyrus-and:
@aslushnikov @pavelfeldman the problem of setting the `host` header to the empty value is that it will break the `webSocketDebuggerUrl` field for the `/json/list` and `/json/new` endpoints:

```console
$ curl -H host: localhost:9222/json/list                                                       
[ {
[...]

PR #4268, opened 2019-04-10
fix(launcher): support https for browserURL endpoint

sbekrin:
Hey there 👋

We'd like to deploy Puppeteer to internal Kubernetes (which serves everything via https, at least in our setup), but apperently `browserURL` is limited to http requests at the moment. I wonder if there's any particular reason behind it?

I'm not sure how to setup https case for testing, but happy to work on it with some guidance.

Thanks!

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -16,6 +16,7 @@
 const os = require('os');
 const path = require('path');
 const http = require('http');
+const https = require('https');
 const URL = require('url');
 const removeFolder = require('rimraf');
 const childProcess = require('child_process');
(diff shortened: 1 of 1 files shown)


aslushnikov:
> which serves everything via https, at least in our setup

How's this possible? Do I understand correctly that k8s launches a chrome instance with a remote debugging exposed on a particular port, and then this port is re-exposed with K8s over https?

sbekrin:
@aslushnikov ah sorry, actually https is enforced on a different level for our whole infra. But yeah, in general we deploy a container with a puppeteer w/ remote debugging enabled to k8s. There's minor hijack required to compensate #2242, but overall it works as expected :)

aslushnikov (approved):

sbekrin:
@aslushnikov thanks!

PR #4269, opened 2019-04-10
fix(domworld): Allow setting re-setting of context

joelgriffith:
Allows re-setting of context inside of the DOMWorld utility, as having multiple debuggers attached can cause this value to get re-sent.

### DIFF
--- a/experimental/puppeteer-firefox/lib/DOMWorld.js
+++ b/experimental/puppeteer-firefox/lib/DOMWorld.js
@@ -17,6 +17,7 @@ class DOMWorld {
     /** @type {!Set<!WaitTask>} */
     this._waitTasks = new Set();
     this._detached = false;
+    this._context = null;
   }
 
   frame() {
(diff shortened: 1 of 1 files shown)


aslushnikov:
>  as having multiple debuggers attached can cause this value to get re-sent.

@joelgriffith wait how's this?

joelgriffith:
@aslushnikov see https://github.com/GoogleChrome/puppeteer/issues/4197#issuecomment-481793410. It's kind of an edgecase, but I'm sure it'll appear once more and more tooling start to use puppeteer...

joelgriffith:
@aslushnikov What are your thoughts on this? Happy to take a different direction in the fix, otherwise I'm not sure how far back this issue goes.

JoelEinbinder:
This is fixing a puppeteer-firefox bug?

aslushnikov:
@joelgriffith thank you for the PR and the repro! I'll close this in favor of #4276

PR #4270, opened 2019-04-11
fix: Allow PUPPETEER_EXECUTABLE_PATH to be set in npmrc and package.json

travis-w:
Hello,

This PR sets fixes the PUPPETEER_EXECUTABLE_PATH environment variable so that it can be set in the .npmrc or package.json like the documentation says. Other env variables are set up this way, this one just seems to have been missed.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -297,7 +297,7 @@ class Launcher {
     const browserFetcher = new BrowserFetcher(this._projectRoot);
     // puppeteer-core doesn't take into account PUPPETEER_* env variables.
     if (!this._isPuppeteerCore) {
-      const executablePath = process.env['PUPPETEER_EXECUTABLE_PATH'];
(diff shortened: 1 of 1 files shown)


aslushnikov:
@travis-w the `npm_config_` env variables are exposed only in install scripts. Install scripts don't run puppeteer's launch - why do you need this?

travis-w:
@aslushnikov I believe they are exposed in all npm scripts. So anytime someone starts a program that uses puppeteer via an `npm run` command, these will be exposed.

We have a project where we have to point to a local chrome executable for puppeteer (because of a corporate proxy), so whenever the project is run, the PUPPETEER_EXECUTABLE_PATH needs to be set. It can be easier/cleaner to set these config variables in the .npmrc or package.json.

Plus in the [documentation](https://github.com/GoogleChrome/puppeteer/blob/v1.14.0/docs/api.md#environment-variables), it says:

> If Puppeteer doesn't find them in the environment during the installation step, a lowercased variant of these variables will be used from the npm config.

aslushnikov (approved):

PR #4276, opened 2019-04-11
fix: gracefully handle multiple contexts for secondary DOM World

### LINKED ISSUE #4197: TypeError: Cannot read property 'call' of null
This gets tricky since I can't really provide you the best steps to reproduce.
We have a flow that opens in a popup. Once you click the final "Done" button, it dismisses the popup. In the latest versions of Puppeteer (1.12.0+) this seems to also close the page that opened the popup (This should not be happening. This page should redirect.)

This issue does NOT occur on Puppeteer 1.10.0. (I suspect the issue may actually lie in Chromium, but figured I'd start here.)

* Puppeteer version: 1.13.0
* Platform / OS version: macOS 10.14.2
* Node.js version: 10.15.0

I can't really provide a code snippet, but essentially we perform a `page.click(doneButton);`, the click is successful, then after a second or two both the popup, and the original page are torn down (the entire browser, actually) and this error is thrown:

```
TypeError: Cannot read property 'call' of null

      at DOMWorld._setContext (../../node_modules/puppeteer/lib/DOMWorld.js:61:36)
      at FrameManager._onExecutionContextCreated (../../node_modules/puppeteer/lib/FrameManager.js:303:13)
      at CDPSession.FrameManager._client.on.event (../../node_modules/puppeteer/lib/FrameManager.js:52:70)
      at CDPSession._onMessage (../../node_modules/puppeteer/lib/Connection.js:200:12)
      at Connection._onMessage (../../node_modules/puppeteer/lib/Connection.js:112:17)
      at WebSocketTransport._ws.addEventListener.event (../../node_modules/puppeteer/lib/WebSocketTransport.js:41:24)
      at WebSocket.onMessage (../../node_modules/ws/lib/event-target.js:120:16)
      at Receiver.receiverOnMessage (../../node_modules/ws/lib/websocket.js:741:20)
```

I'm aware this is probably not enough to go on, so if there are additional logs that I can grab (and you can provide me with instructions on how to grab them) I'd be happy to provide some more logging.

Original page should remain and redirect.

**What happens instead?**
Original page is destroyed and the entire browser instance is closed.

aslushnikov:
In case of multiple sessions to the same target, there's a race between
sessions to create a secondary isolated world. As a result, we might
end up having 2 execution contexts created for the needs of the
secondary isolated world.

This patch starts handling this race gracefully: instead of crashing,
we can use either of the execution contexts and ignore the rest.

Notably, the same race condition might happen if page reloads itself
in-between the calls to `page.addEvaluateOnNewDocument` and
`page.createIsolatedWorld`.

Fixes #4197.

### DIFF
--- a/lib/DOMWorld.js
+++ b/lib/DOMWorld.js
@@ -70,6 +70,13 @@ class DOMWorld {
     }
   }
 
+  /**
+   * @return {boolean}
+   */
+  _hasContext() {
+    return !this._contextResolveCallback;
+  }
+
   _detach() {
     this._detached = true;
     for (const waitTask of this._waitTasks)
--- a/test/launcher.spec.js
+++ b/test/launcher.spec.js
@@ -316,6 +316,17 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions, p
         expect(await restoredPage.evaluate(() => 7 * 8)).toBe(56);
         await browser.close();
       });
+      // @see https://github.com/GoogleChrome/puppeteer/issues/4197#issuecomment-481793410
+      it('should be able to connect to the same page simultaneously', async({server}) => {
+        const browserOne = await puppeteer.launch();
+        const browserTwo = await puppeteer.connect({ browserWSEndpoint: browserOne.wsEndpoint() });
+        await Promise.all([
+          new Promise(x => browserOne.once('targetcreated', target => x(target.page()))),
+          browserTwo.newPage(),
+        ]);
+        await browserOne.close();
+      });
+
     });
     describe('Puppeteer.executablePath', function() {
       it('should work', async({server}) => {
--- a/test/launcher.spec.js
+++ b/test/launcher.spec.js
@@ -320,10 +320,12 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions, p
       it('should be able to connect to the same page simultaneously', async({server}) => {
         const browserOne = await puppeteer.launch();
         const browserTwo = await puppeteer.connect({ browserWSEndpoint: browserOne.wsEndpoint() });
-        await Promise.all([
+        const [page1, page2] = await Promise.all([
           new Promise(x => browserOne.once('targetcreated', target => x(target.page()))),
           browserTwo.newPage(),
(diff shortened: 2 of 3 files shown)


lib/DOMWorld.js
```diff
@@ -70,6 +70,13 @@ class DOMWorld {
     }
   }
 
+  /**
+   * @return {boolean}
+   */
+  _hasContext() {
```
JoelEinbinder, line 76: Public?

test/launcher.spec.js
```diff
@@ -316,6 +316,17 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions, p
...
       });
+      // @see https://github.com/GoogleChrome/puppeteer/issues/4197#issuecomment-481793410
+      it('should be able to connect to the same page simultaneously', async({server}) => {
+        const browserOne = await puppeteer.launch();
+        const browserTwo = await puppeteer.connect({ browserWSEndpoint: browserOne.wsEndpoint() });
+        await Promise.all([
+          new Promise(x => browserOne.once('targetcreated', target => x(target.page()))),
+          browserTwo.newPage(),
+        ]);
+        await browserOne.close();
```
JoelEinbinder, line 329: Can we make this test stronger? Query something with each page

JoelEinbinder (approved):

lib/DOMWorld.js
aslushnikov: i'd keep it private - no need to be public.

JoelEinbinder: But isn’t the class private?

JoelEinbinder: But isn’t the class private?

test/launcher.spec.js
aslushnikov: Done.

lib/DOMWorld.js
aslushnikov: There are plans to eventually make it public as part of #2671. For this reason the `_setContext` was made private.

JoelEinbinder (approved):

Issue #4281, opened 2019-04-13
Puppeteer consider visible element as not visible


Kalliser:
Hello I have a similar issue than the one you previously corrected in ticket #4274 but this one is still present even with last version of master branch of puppeteer

## Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: master branch ( 8f8e77492dd91a00d53965b998b4199bcc5e29a3 )
* Platform / OS version: Debian (Dockerized in node:10-stretch-slim)
* URLs (if applicable): https://www.edistat.com/
* Node.js version: v10.15.1

**What steps will reproduce the problem?**

Open the website
Click on `.list-group .list-group-item:nth-of-type(1) .gotodetail`
Wait for the url `https://www.edistat.com/livre_tarifs.php?ean=9782702165485`

_Please include code that reproduces the issue._
Note 1 : this code sample works fine until 1.5.0 of puppeteer and below

```
const browser = await puppeteer.launch();

const page = await browser.newPage();
await page.goto('https://www.edistat.com/');

await Promise.all([
  page.click('.list-group .list-group-item:nth-of-type(1) .gotodetail'),
  page.waitForNavigation({waitUntil: ['networkidle0', 'load', 'domcontentloaded'], timeout: 30000})
]);
console.log(await page.url());
await page.screenshot({path: 'screenshot.png', fullPage: true});

await browser.close();

```

**What is the expected result?**
The page `https://www.edistat.com/livre_tarifs.php?ean=9782702165485` should be opened when clicking on `.list-group .list-group-item:nth-of-type(1) .gotodetail`

**What happens instead?**
Browser is stuck on homepage with the message 

> Error running your code. Error: Node is either not visible or not an HTMLElement

aslushnikov:
@Kalliser nice, thank you for reporting this. This is upstream bug, filed as https://crbug.com/952562

Upstream fix is out for a review: https://chromium-review.googlesource.com/c/chromium/src/+/1564937

aslushnikov:
@Kalliser the fix has landed and is published at `puppeteer@next`.

Kalliser:
@aslushnikov Thank you very much I confirm you that it now working on my side.

I stay aware if another bug appear.

PR #4282, opened 2019-04-13
test: add failing test for clicking certain elements

aslushnikov:
References #4281


JoelEinbinder (approved):
Can you remove the Unicode/::before or explain why they are needed?

PR #4292, opened 2019-04-16
feat(pki): use npm ca file for install

boliangt:
Port pki install from phantomjs-prebuilt
https://github.com/Medium/phantomjs/blob/2.1.14/install.js#L175

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -441,7 +441,7 @@ This methods attaches Puppeteer to an existing Chromium instance.
   - `host` <[string]> A download host to be used. Defaults to `https://storage.googleapis.com`.
   - `path` <[string]> A path for the downloads folder. Defaults to `<root>/.local-chromium`, where `<root>` is puppeteer's package root.
   - `platform` <[string]> Possible values are: `mac`, `win32`, `win64`, `linux`. Defaults to the current platform.
-  - `ca` <?string|Buffer|Array<string|Buffer>> Certificate Authority.
+  - `ca` <?[string]|[Buffer]|Array<[string]|[Buffer]>> Certificate Authority.
 - returns: <[BrowserFetcher]>
 
 #### puppeteer.defaultArgs([options])
--- a/docs/api.md
+++ b/docs/api.md
@@ -441,6 +441,7 @@ This methods attaches Puppeteer to an existing Chromium instance.
   - `host` <[string]> A download host to be used. Defaults to `https://storage.googleapis.com`.
   - `path` <[string]> A path for the downloads folder. Defaults to `<root>/.local-chromium`, where `<root>` is puppeteer's package root.
   - `platform` <[string]> Possible values are: `mac`, `win32`, `win64`, `linux`. Defaults to the current platform.
+  - `ca` <?string|Buffer|Array<string|Buffer>> Certificate Authority.
 - returns: <[BrowserFetcher]>
 
 #### puppeteer.defaultArgs([options])
--- a/install.js
+++ b/install.js
@@ -36,7 +36,23 @@ if (process.env.NPM_PACKAGE_CONFIG_PUPPETEER_SKIP_CHROMIUM_DOWNLOAD || process.e
 const downloadHost = process.env.PUPPETEER_DOWNLOAD_HOST || process.env.npm_config_puppeteer_download_host || process.env.npm_package_config_puppeteer_download_host;
 
 const puppeteer = require('./index');
-const browserFetcher = puppeteer.createBrowserFetcher({ host: downloadHost });
+
+// Override current environment proxy settings with npm configuration, if any.
+const NPM_HTTPS_PROXY = process.env.npm_config_https_proxy || process.env.npm_config_proxy;
+const NPM_HTTP_PROXY = process.env.npm_config_http_proxy || process.env.npm_config_proxy;
+const NPM_NO_PROXY = process.env.npm_config_no_proxy;
+const NPM_CONFIG_CAFILE = process.env.npm_config_cafile;
+
+if (NPM_HTTPS_PROXY)
+  process.env.HTTPS_PROXY = NPM_HTTPS_PROXY;
+if (NPM_HTTP_PROXY)
+  process.env.HTTP_PROXY = NPM_HTTP_PROXY;
+if (NPM_NO_PROXY)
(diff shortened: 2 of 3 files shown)


docs/api.md
```diff
@@ -465,6 +465,7 @@ This methods attaches Puppeteer to an existing Chromium instance.
   - `host` <[string]> A download host to be used. Defaults to `https://storage.googleapis.com`.
   - `path` <[string]> A path for the downloads folder. Defaults to `<root>/.local-chromium`, where `<root>` is puppeteer's package root.
   - `platform` <[string]> Possible values are: `mac`, `win32`, `win64`, `linux`. Defaults to the current platform.
+  - `ca` <?string|Buffer|Array<string|Buffer>> Certificate Authority.
```
vsemozhetbyt: To linkify:
```suggestion
  - `ca` <?[string]|[Buffer]|Array<[string]|[Buffer]>> Certificate Authority.
```

aslushnikov:
@boliangt can you please sign CLA to make bot happy?

lib/BrowserFetcher.js
```diff
@@ -287,6 +288,8 @@ function httpRequest(url, method, response) {
     else
       response(res);
   };
+  if (opt.ca && options.protocol === 'https:')
+    options.ca = opt.ca;
```
aslushnikov, line 292: I don't see node.js API to accept a "ca" option here: https://nodejs.org/api/https.html#https_https_request_options_callback

Is this a thing?

boliangt: Well, Node API states : 
> The following additional options from tls.connect() are also accepted: ca, cert, ciphers, clientCertEngine, crl, dhparam, ecdhCurve, honorCipherOrder, key, passphrase, pfx, rejectUnauthorized, secureOptions, secureProtocol, servername, sessionIdContext.

The example provided does not use "ca" option, but i don't think it's relevant since it's an "tls.connect()" option

boliangt:
I signed it!

boliangt:
@aslushnikov I signed it, but it seems there's still something missing.

install.js
```diff
@@ -61,6 +50,21 @@ if (NPM_HTTP_PROXY)
 if (NPM_NO_PROXY)
   process.env.NO_PROXY = NPM_NO_PROXY;
 
+const NPM_CONFIG_CA = getCertificateAuthority(NPM_CONFIG_CAFILE);
+const fetcherOptions = NPM_CONFIG_CA ? { host: downloadHost, ca: NPM_CONFIG_CA } : { host: downloadHost };
+const browserFetcher = puppeteer.createBrowserFetcher(fetcherOptions);
+
+const revision = process.env.PUPPETEER_CHROMIUM_REVISION || process.env.npm_config_puppeteer_chromium_revision || process.env.npm_package_config_puppeteer_chromium_revision
```
aslushnikov, line 57: why is this code moved to here?

install.js
```diff
@@ -154,3 +158,23 @@ function logPolitely(toBeLogged) {
     console.log(toBeLogged);
 }
 
+function getCertificateAuthority(configCaFile) {
+  let ca;
```
aslushnikov, line 162: can you please add a comment pointing that this code is ported from phantomjs install script?

lib/BrowserFetcher.js
```diff
@@ -299,6 +302,7 @@ function httpRequest(url, method, response) {
  * @property {string=} platform
  * @property {string=} path
  * @property {string=} host
+ * @property {?(string|Buffer|Array<string|Buffer>)=} ca
```
aslushnikov, line 305: how can it be a buffer?

install.js
```diff
@@ -154,3 +158,23 @@ function logPolitely(toBeLogged) {
...
+  if (configCaFile) {
```
aslushnikov, line 163: let's merge these two `if` statements; or better - use fast-return:

```js
function getCertificateAuthority() {
  if (NPM_CONFIG_CA)
    return NPM_CONFIG_CA;
  if (!NPM_CONFIG_CAFILE)
    return null;
  try {
    const fs = require('fs');
    const ca = fs.readFileSync(NPM_CONFIG_CAFILE, { encoding: 'utf8' })
                           .split(/\n(?=-----BEGIN CERTIFICATE-----)/g);
    // Comments at the beginning of the file result in the first
    // item not containing a certificate - in this case the
    // download will fail
    if (ca.length > 0 && !/-----BEGIN CERTIFICATE-----/.test(ca[0]))
      ca = ca.shift();
    return ca;
  } catch (e) {
    console.error('Could not read cafile', NPM_CONFIG_CAFILE, e);
  }
}
```

install.js
```diff
@@ -61,6 +50,21 @@ if (NPM_HTTP_PROXY)
 if (NPM_NO_PROXY)
   process.env.NO_PROXY = NPM_NO_PROXY;
 
+const NPM_CONFIG_CA = getCertificateAuthority(NPM_CONFIG_CAFILE);
```
aslushnikov, line 53:
```js
const ca = getCertificateAuthority();
```

install.js
```diff
@@ -36,23 +36,12 @@ if (process.env.NPM_PACKAGE_CONFIG_PUPPETEER_SKIP_CHROMIUM_DOWNLOAD || process.e
...
-if (revisionInfo.local) {
-  generateProtocolTypesIfNecessary(false /* updated */);
-  return;
-}
 
 // Override current environment proxy settings with npm configuration, if any.
 const NPM_HTTPS_PROXY = process.env.npm_config_https_proxy || process.env.npm_config_proxy;
 const NPM_HTTP_PROXY = process.env.npm_config_http_proxy || process.env.npm_config_proxy;
 const NPM_NO_PROXY = process.env.npm_config_no_proxy;
+const NPM_CONFIG_CAFILE = process.env.npm_config_cafile;
```
aslushnikov, line 44: let's also define `NPM_CONFIG_CA`:

```js
const NPM_CONFIG_CA = process.env.npm_config_ca;
```

aslushnikov (commented):
Left some inline comments.

In general, why would you want to have this logic? When is it requires to use custom CA?

aslushnikov:
> In general, why would you want to have this logic? When is it requires to use custom CA?

Since we still don't know when and why is this needed - closing this.
We'll revisit when we understand the purpose and usecases for the code.

PR #4314, opened 2019-04-19
test: Make intervention test Edge friendly

kblok:
All tests are green when running against [Microsoft Edge Insider](https://www.microsoftedgeinsider.com/en-us/), except this test.

Edge returns `https://to-be-replaced.invalid/feature/5718547946799104` instead of `https://www.chromestatus.com/feature/5718547946799104`

I think that checking for `feature/5718547946799104` is a small test and will do the job for both browsers.

```
$env:CHROME = "C:\Program Files (x86)\Microsoft\Edge Dev\Application\msedge.exe"; node
.\test.js
```


aslushnikov (approved):
Thanks!

Issue #3641, opened 2018-12-07
Accessibility: expose element reference


AutoSponge:
### Steps to reproduce

While using the puppeteer api `page.accessibility.snapshot()` the AXNode model leaves no trace of the DOM implementation. The underlying calls to [CDP](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility#method-getFullAXTree) return `nodeId` (and `backendDOMNodeId`).

I suggest one or more of the following (will update if more ideas come in comments):
1. Give AXNode a new method 'asElement' which will create an `ElementHandle` from the stored nodeId or remoteObject reference.
1. Expose the nodeId or remoteObjectId on the AXNode (so it can be passed to other CDP calls).
1. Expose AXNode so CDP calls to `Accessibility.getPartialAXTree` can use this code to normalize the result the way `snapshot()` does.

* Puppeteer version: 1.10+
* Platform / OS version: any
* URLs (if applicable):
* Node.js version: latest

_Please include code that reproduces the issue._

```js
// ... open google.com
const tree = await this.page.accessibility.snapshot();
console.log(tree);
```

Yields: 

```js
{ role: 'WebArea',
  name: 'Google',
  children:
   [ { role: 'button', name: 'Dismiss' },
     { role: 'link', name: 'About' },
     { role: 'link', name: 'Store' },
     { role: 'link', name: 'Gmail' },
     { role: 'link', name: 'Images' },
     { role: 'button', name: 'Google apps' },
     { role: 'link', name: 'Sign in' },
     { role: 'combobox',
       name: 'Search',
       description: 'Search',
       focused: true,
       autocomplete: 'both' },
     { role: 'button', name: 'Search by voice' },
     { role: 'button',
       name: 'Google Search',
       description: 'Google Search' },
     { role: 'button',
       name: 'I\'m Feeling Lucky',
       description: 'I\'m Feeling Lucky' },
     { role: 'img', name: 'Google' },
     { role: 'link', name: 'Privacy' },
     { role: 'link', name: 'Terms' },
     { role: 'combobox', name: 'Settings', haspopup: 'menu' },
     { role: 'link', name: 'Advertising' },
     { role: 'link', name: 'Business' } ] }
```

But more useful would be:

```js
{ role: 'WebArea',
  name: 'Google',
  nodeId: 1,
  children:
   [ { role: 'button', name: 'Dismiss', nodeId: 4 },
     { role: 'link', name: 'About', nodeId: 6 },
[...]

PR #4318, opened 2019-04-20
feat: element option in page.accessibility.snapshot()

JoelEinbinder:
Going from  AXNode -> ElementHandle is turning out to be controversial. This patch instead adds a way to go from ElementHandle -> AXNode. If the API looks good, I'll add it into Firefox as well.
#3641

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2090,6 +2090,7 @@ Most of the accessibility tree gets filtered out when converting from Blink AX T
 #### accessibility.snapshot([options])
 - `options` <[Object]>
   - `interestingOnly` <[boolean]> Prune uninteresting nodes from the tree. Defaults to `true`.
+  - `element` <[ElementHandle]> The root DOM element for the snapshot. Defaults to the whole page.
 - returns: <[Promise]<[Object]>> An [AXNode] object with the following properties:
   - `role` <[string]> The [role](https://www.w3.org/TR/wai-aria/#usage_intro).
   - `name` <[string]> A human readable name for the node.
--- a/docs/api.md
+++ b/docs/api.md
@@ -2090,7 +2090,7 @@ Most of the accessibility tree gets filtered out when converting from Blink AX T
 #### accessibility.snapshot([options])
 - `options` <[Object]>
   - `interestingOnly` <[boolean]> Prune uninteresting nodes from the tree. Defaults to `true`.
-  - `element` <[ElementHandle]> The root DOM element for the snapshot. Defaults to the whole page.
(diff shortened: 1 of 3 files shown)


aslushnikov:
High-level questions:
- Is there any particular reason why this is not under `page.accessibility` namespace?
- How does it actually help to solve the AXNode -> ElementHandle dilemma?

JoelEinbinder:
As we discussed offline, I changed the API to be part of page.accessibility.snapshot. We can look to adding something on elementHandle later. Same functionality. Pass an element to capture a snapshot rooted at that element.

test/accessibility.spec.js
```diff
@@ -307,6 +307,58 @@ module.exports.addTests = function({testRunner, expect, FFOX}) {
...
+            [ { role: 'menuitem', name: 'First Item' },
+              { role: 'menuitem', name: 'Second Item' },
+              { role: 'menuitem', name: 'Third Item' } ]
+          });
+        });
+        it('should return null when the element is no longer in DOM', async({page}) => {
+          await page.setContent(`<button>My Button</button>`);
+          const button = await page.$('button');
+          await page.$eval('button', button => button.remove());
+          expect(await await page.accessibility.snapshot({element: button})).toEqual(null);
```
aslushnikov: double await?!

test/accessibility.spec.js
```diff
@@ -307,6 +307,58 @@ module.exports.addTests = function({testRunner, expect, FFOX}) {
...
+            <div role="menuitem">Third Item</div>
+          </div>`);
+
+          const button = await page.$('button');
+          expect(await await page.accessibility.snapshot({element: button})).toEqual({
+            role: 'button',
+            name: 'My Button'
+          });
+
+          const input = await page.$('input');
```
aslushnikov, line 324: can we split this into 3 tests?

test/accessibility.spec.js
JoelEinbinder: Done

test/accessibility.spec.js
JoelEinbinder: Done

aslushnikov (approved):

docs/api.md
```diff
@@ -2090,6 +2090,7 @@ Most of the accessibility tree gets filtered out when converting from Blink AX T
 #### accessibility.snapshot([options])
 - `options` <[Object]>
   - `interestingOnly` <[boolean]> Prune uninteresting nodes from the tree. Defaults to `true`.
+  - `element` <[ElementHandle]> The root DOM element for the snapshot. Defaults to the whole page.
```
aslushnikov: On a second though - can we call this `root`?

PR #4319, opened 2019-04-20
docs(api.md): add notes on race condition in page.setViewport

### LINKED ISSUE #2755: blurry canvas if deviceScaleFactor > 1
### Steps to reproduce

* Puppeteer version: 1.5.0
* Platform / OS version: Mac OS X 10.11.6
* URLs (if applicable):
* Node.js version: 9.4.0

1. Go to https://try-puppeteer.appspot.com/. Use the following code:

```

await page.goto('https://www.google.de/maps/@52.5197103,13.4068538,16.11z');

await page.setViewport({
    width: 800,
    height: 600,
    deviceScaleFactor: 3
});

await page.waitFor(5000);

await page.screenshot({path: 'screenshot.png'});

await browser.close();
```
2. Check screenshot and compare it against what you can see while browsing https://openlayersbook.github.io/ch06-styling-vector-layers/example-03.html on a retina display device in a real Chrome.

Sharp canvas.

**What happens instead?**
Blurry canvas.

aslushnikov:
Fixes #2755.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1236,7 +1236,12 @@ Emulates given device metrics and user agent. This method is a shortcut for call
 - [page.setViewport(viewport)](#pagesetviewportviewport)
 
 To aid emulation, puppeteer provides a list of device descriptors which can be obtained via the [`puppeteer.devices`](#puppeteerdevices).
-Below is an example of emulating an iPhone 6 in puppeteer:
+
+To avoid race condition with page's logic that sniffs for viewport and user
+agent properties, it is recommended to call `page.emulate` *before* navigating
+page to the target URL.
+
+Example of emulating an iPhone 6 in puppeteer:
 
 ```js
 const puppeteer = require('puppeteer');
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -1799,6 +1804,20 @@ puppeteer.launch().then(async browser => {
 
 In the case of multiple pages in a single browser, each page can have its own viewport size.
 
+To avoid race condition with page's logic that sniffs for viewport properties,
```
JoelEinbinder: This sounds like inside baseball. Can we phrase it in a way that will make more sense to puppeteer users? Is this a bug or intended behavior? What happens if I emulate after navigating? If its a bug where is the link?

This also implies that I can't use setViewport to simulate a resize event. Is that true?

JoelEinbinder: page.setViewport will resize the page. A lot of websites don't expect phones to change size, so you should set the viewport before navigating to the page.

docs/api.md
```diff
@@ -1236,7 +1236,8 @@ Emulates given device metrics and user agent. This method is a shortcut for call
 - [page.setViewport(viewport)](#pagesetviewportviewport)
 
 To aid emulation, puppeteer provides a list of device descriptors which can be obtained via the [`puppeteer.devices`](#puppeteerdevices).
-Below is an example of emulating an iPhone 6 in puppeteer:
+
+`page.emulate` will resize the page. A lot of websites don't expect phones to change size, so you should emulate before navigating to the page.
```
JoelEinbinder, line 1240: This is even easier. The user agent needs to be sent before the first navigation.

JoelEinbinder (approved):

docs/api.md
aslushnikov: Fair! Keeping it as-is doesn't hurt.

PR #4327, opened 2019-04-23
fix: update troubleshooting.md

sstopkin:
Removed unused `rm -rf` call. This directory doesn't exist (and same for all root docker images).


sstopkin:
I signed CLA

aslushnikov (approved):
Thanks!

PR #4345, opened 2019-04-25
docs(readme): Document debugging with node --inspect-brk

melissachang:
To me, this is the most important type of debugging. It's what I think of when I think of debugging my test. I was happy to find the instructions [buried here](https://github.com/GoogleChrome/puppeteer/issues/398#issuecomment-323551586).


aslushnikov (approved):

aslushnikov:
@melissachang Thank you for the PR!

PR #4349, opened 2019-04-26
fix(docs): screen readers aren't the only kind of assistive technology

JoelEinbinder:
The docs incorrectly implied that screen readers were the only assistive technology to do filtering on the tree.
@alice


aslushnikov (approved):

alice:
Nice fix 👍

Issue #4334, opened 2019-04-24
exposeFunction fails to serialize arguments correctly when toJSON is overridden


mnmkng:
Hello,

today I encountered an interesting issue where a specific website would cause `page.exposeFunction()` to stop working properly. The issue lies in arguments being incorrectly serialized or deserialized. See the code examples below.

### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.14.0
* Platform / OS version: MacOS, Linux
* URL: http://erasmus-databaze.naep.cz/modules/erasmus/index.php?page=1
* Node.js version: LTS/Dubnium

**What steps will reproduce the problem?**
```js
const puppeteer = require('puppeteer');

(async function main() {
    const browser = await puppeteer.launch({ headless: true });
    const page = await browser.newPage();
    await page.exposeFunction('myFunction', async (...args) => {
        console.log(...args);
        for (arg of args) {
            console.log(typeof arg);
        }
    });
    await page.goto('http://erasmus-databaze.naep.cz/modules/erasmus/index.php?page=1');
    await page.evaluate(async () => {
        await myFunction({ x: 1 })
    });

    await browser.close();
})();
```

**What is the expected result?**
```
{ x: 1 }
object
```

**What happens instead?**
```
[ { " x " :   1 } ]
string
string
string
string
string
string
string
string
string
string
```

**Further findings**
It seems that some environment change in the website itself causes the parameters to be incorrectly serialized. It can be seen in the `event.data` variable in `WebSocketTransport`.

```
{"method":"Runtime.bindingCalled","params":{"name":"myFunction","payload":"{\"name\":\"myFunction\",\"seq\":1,\"args\":\"[{\\\"x\\\": 1}]\"}","executionContextId":3}, "sessionId": "D74A031C993D759D5EA29C38C002400F"}
```
The `args` are a `string`, but it should be an `Array`.

When the code above is called for `http://example.com` or mostly any other website, it runs fine.

**Conclusion**
I assume that the problem is caused by the website itself, but I'd expect Puppeteer not to rely on the website's environment to do serialization. I'll try to dig more into how serialization is done in Puppeteer the meantime.

Thank you for your time.

mnmkng:
Ok, quick update. It's most probably due to the `Object.toJSON` function that's in the page's environment. **Deobfuscated** using `jsnice.org`

```js
/**
 * @param {!Object} b
 * @return {?}
 */
function K(b) {
  /** @type {string} */
[...]

PR #4352, opened 2019-04-26
test(page): add failing test for complex objects in exposeFunction

kblok:
related to #4334

I have this fix locally:

```ts
async _onBindingCalled(event) {
    let {name, seq, args} = JSON.parse(event.payload);
    let expression = null;

   //THIS IF
    if (helper.isString(args))
      args = JSON.parse(args);

    try {
      const result = await this._pageBindings.get(name)(...args);
      expression = helper.evaluationString(deliverResult, name, seq, result);
    } catch (error) {
      if (error instanceof Error)
        expression = helper.evaluationString(deliverError, name, seq, error.message, error.stack);
      else
        expression = helper.evaluationString(deliverErrorValue, name, seq, error);
    }
    this._client.send('Runtime.evaluate', { expression, contextId: event.executionContextId }).catch(debugError);

    /**
     * @param {string} name
     * @param {number} seq
     * @param {*} result
     */
    function deliverResult(name, seq, result) {
      window[name]['callbacks'].get(seq).resolve(result);
      window[name]['callbacks'].delete(seq);
    }

    /**
     * @param {string} name
     * @param {number} seq
     * @param {string} message
     * @param {string} stack
     */
    function deliverError(name, seq, message, stack) {
      const error = new Error(message);
      error.stack = stack;
      window[name]['callbacks'].get(seq).reject(error);
      window[name]['callbacks'].delete(seq);
    }

    /**
     * @param {string} name
     * @param {number} seq
     * @param {*} value
     */
    function deliverErrorValue(name, seq, value) {
      window[name]['callbacks'].get(seq).reject(value);
      window[name]['callbacks'].delete(seq);
    }
  }
```

But it could also be fixed on Chromium.

### DIFF
--- a/test/page.spec.js
+++ b/test/page.spec.js
@@ -656,6 +656,16 @@ module.exports.addTests = function({testRunner, expect, headless, puppeteer, CHR
       });
       expect(result).toBe(15);
     });
+
+    xit('should work with complex objects', async({page, server}) => {
+      await page.exposeFunction('complexObject', function(a, b) {
+        return {x: a.x + b.x};
+      });
+      const result = await page.evaluate(async function() {
+        return complexObject({x: 5}, {x: 2});
+      });
+      expect(result.x).toBe(7);
+    });
   });
 
   describe('Page.Events.PageError', function() {


aslushnikov:
@kblok wait, your test works for me just fine. What am I doing wrong?

Submitted https://github.com/GoogleChrome/puppeteer/pull/4411 - let's see how it works on different platforms

aslushnikov:
@kblok ah ok, the #4334 happens because of overriden `toJSON` - which is not part of the test.

I'll close this for now since the test is actually landed to pptr repo.

PR #4359, opened 2019-04-29
docs(troubleshooting.md): Pin Alpine Chromium version

Cherry:
* Pins Alpine Chromium version to prevent updates from causing issues

When using these instructions today, I found that the Chromium version in the latest Alpine edge was 73, which caused errors with Puppeteer. Pinning to the latest 72 version in Alpine registry resolved the issue, and should prevent others from running into it in the future.


Cherry:
I signed it!

aslushnikov (approved):
Thank you @Cherry for the PR!

(Sorry for the long response - we've been caught at the I/O preparations.)

PR #4277, opened 2019-04-12
fix(page): intersect content quads with viewport

aslushnikov:
In certain cases inline element children might be positioned
outside of viewport.

In this case, we should intersect all content quads with viewport
before we pick one to click into.

Fixes #4274.

PR #4369, opened 2019-04-30
fix(page): Simplify _clickablePoint function

kblok:
Based on #4277

We have two loops in `_clickablePoint`. One to convert an `Array<number>` to `Array<{x: number, y: number}>` and another run to call `_intersectQuadWithViewport`.
I think that having all that in on run will make the code a little easier to read and also little faster.


aslushnikov:
> I think that having all that in on run will make the code a little easier to read

@kblok Nah, I like it more the way it is today 🤷‍♂️

> and also little faster.

If it hurts your perf somewhere - I'm open to the change. But there should not be any observable perf gains from this change.

kblok:
Closing then 😂

PR #4371, opened 2019-04-30
fix: waitForSelector with visible:true causes timeout

### LINKED ISSUE #4356: waitForSelector with visible:true not returning the first visible element, causes timeout.
### Steps to reproduce

* Puppeteer version: 0.14.0
* Platform / OS version: MacOS (Dockerized in node:10)
* URLs (if applicable): https://j4q389wzv3.codesandbox.io/
* Node.js version: v10.15.3

Open Website OR use this file:
```
<html>
  <head>
    <meta charset="UTF-8" />
  </head>

  <body>
    <input style="display: none" type="number" class="my-input-class" />
    <input type="number" class="my-input-class" />
  </body>
</html>
```

```

await page.goto('https://j4q389wzv3.codesandbox.io/');

try {
  const visibleInput = await page.waitForSelector('.my-input-class', {visible: true, timeout: 1000 })
  console.log('Found visible Element')
} catch (e) {
  console.log('Could NOT find a visible element ', e.message)
}

const inputs = await page.$$('.my-input-class')
console.log(`Found ${inputs.length} inputs`)

await browser.close()
```

There are 2 elements matching the CSS Selector on the page.  the first one is hidden, the second one is visible.  The `page.waitForSelector` with `{visible: true}` should have found and returned the visible element on the page.

**What happens instead?**
It Times Out since there was another HIDDEN element matching the CSS selector higher up in the DOM structure, causing it to wait until timeout even though a visible element exists on the page.

razorman8669:
This patch fixes an issue where waitForSelector with visible:true would cause a timeout should there be multiple elements matching the selector, but the ones higher up on the DOM are hidden.  It now returns the first visible element it finds.

fixes #4356

### DIFF
--- a/lib/DOMWorld.js
+++ b/lib/DOMWorld.js
@@ -516,7 +516,7 @@ class DOMWorld {
 
       // We need to loop over all matching nodes, and test each one, returning the first successful one... or null
       if (isXPath) {
-        const nodeIterator = document.evaluate(selectorOrXPath, document, null, XPathResult.ORDERED_NODE_ITERATOR_TYPE, null)
+        const nodeIterator = document.evaluate(selectorOrXPath, document, null, XPathResult.ORDERED_NODE_ITERATOR_TYPE, null);
         let node = nodeIterator.iterateNext();
         while (node) {
           if (testNode(node))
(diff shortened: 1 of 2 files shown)


razorman8669:
line endings caused all sorts of issues with OFL.txt on macOSX.  closing and redoing this PR cause I need to edit my commits to include the correct emails, and rebasing fails because of damn line endings.

PR #4373, opened 2019-04-30
fix: waitForSelector with visible:true causes timeout

### LINKED ISSUE #4356: waitForSelector with visible:true not returning the first visible element, causes timeout.

razorman8669:
This patch fixes an issue where waitForSelector with visible:true would cause a timeout should there be multiple elements matching the selector, but the ones higher up on the DOM are hidden. It now returns the first visible element it finds.

fixes #4356

### DIFF
--- a/experimental/puppeteer-firefox/lib/DOMWorld.js
+++ b/experimental/puppeteer-firefox/lib/DOMWorld.js
@@ -409,24 +409,52 @@ class DOMWorld {
      * @return {?Node|boolean}
      */
     function predicate(selectorOrXPath, isXPath, waitForVisible, waitForHidden) {
-      const node = isXPath
-        ? document.evaluate(selectorOrXPath, document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue
-        : document.querySelector(selectorOrXPath);
-      if (!node)
-        return waitForHidden;
-      if (!waitForVisible && !waitForHidden)
-        return node;
-      const element = /** @type {Element} */ (node.nodeType === Node.TEXT_NODE ? node.parentElement : node);
+      if (!waitForVisible) {
+        const node = isXPath
+          ? document.evaluate(selectorOrXPath, document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue
+          : document.querySelector(selectorOrXPath);
 
-      const style = window.getComputedStyle(element);
-      const isVisible = style && style.visibility !== 'hidden' && hasVisibleBoundingBox();
-      const success = (waitForVisible === isVisible || waitForHidden === !isVisible);
-      return success ? node : null;
+        if (!node)
+          return waitForHidden;
+        if (!waitForHidden)
+          return node;
+      }
+
+      // We need to loop over all matching nodes, and test each one, returning the first successful one... or null
+      if (isXPath) {
+        const nodeIterator = document.evaluate(selectorOrXPath, document, null, XPathResult.ORDERED_NODE_ITERATOR_TYPE, null);
+        let node = nodeIterator.iterateNext();
+        while (node) {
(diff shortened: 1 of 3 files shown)


razorman8669:
@JoelEinbinder anything else I need to do with this PR to get a reviewer assigned?

aslushnikov:
@razorman8669 I don't think this is the best solution for the problem; please, see [my comment](https://github.com/GoogleChrome/puppeteer/issues/4356#issuecomment-497906422).

I'll close this for now; we can always re-open if this proves to be optimal.

experimental/puppeteer-firefox/lib/DOMWorld.js
```diff
@@ -409,24 +409,52 @@ class DOMWorld {
...
-        ? document.evaluate(selectorOrXPath, document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue
-        : document.querySelector(selectorOrXPath);
-      if (!node)
-        return waitForHidden;
-      if (!waitForVisible && !waitForHidden)
-        return node;
-      const element = /** @type {Element} */ (node.nodeType === Node.TEXT_NODE ? node.parentElement : node);
-
-      const style = window.getComputedStyle(element);
-      const isVisible = style && style.visibility !== 'hidden' && hasVisibleBoundingBox();
```
razorman8669, line 422: @aslushnikov you're very quick to close this out... but if you look at the code here, you'll realize that it originally already uses the so called `magic` to determine visibility (as you can see by `const isVisible = style && style.visibility !== 'hidden' && hasVisibleBoundingBox();`).

My fix merely just loops over all elements on page instead of stopping at the **first** element to determine `visibility`...

could you take a closer look at this before closing it out. thx!

aslushnikov: > already uses the so called magic to determine visibility (as you can see by const isVisible = style && style.visibility !== 'hidden' && hasVisibleBoundingBox();).

@razorman8669  Sure, but this magic is scoped to determining visibility and doesn't affect element selection.

Everywhere throughout our API we rely on CSS Selectors to exclusively determine the element we act upon. We never try to be clever and "find another element" if something goes wrong with the one user gives us. And we shouldn't: users can always provide a more specific css selector if they need to.

The approach when we use a combination of a css selector and options to determine the target element doesn't play nice with the rest of our API and our conventions.

razorman8669: aha, I see what you're saying now and that makes sense.  more specific CSS selector it is!  thanks for clarifying your feedback.

PR #4388, opened 2019-05-05
fix(tracing): start without options

aslushnikov:


lib/Tracing.js
```diff
@@ -33,7 +33,7 @@ class Tracing {
   /**
    * @param {!{path: string, screenshots?: boolean, categories?: !Array<string>}} options
```
JoelEinbinder: Path says it’s required here. Is it?

aslushnikov: It's not. fixed.

JoelEinbinder (approved):

PR #4397, opened 2019-05-06
feat(workers): create workers from service workers and shared workers

JoelEinbinder:
This allows users to easily evaluate javascript inside service workers and shared workers by creating a `Worker` object for them.

### DIFF
--- a/lib/Target.js
+++ b/lib/Target.js
@@ -84,19 +84,16 @@ class Target {
     if (this._targetInfo.type !== 'service_worker' && this._targetInfo.type !== 'shared_worker')
       return null;
     if (!this._workerPromise) {
-      this._workerPromise = (async() => {
-        // top level workers have a fake page wrapping the actual worker
-        const client = await this._sessionFactory();
-        const sessionId = new Promise(resolve => {
-          client.once('Target.attachedToTarget', event => {
-            resolve(event.sessionId);
-          });
-        });
-        await client.send('Target.setAutoAttach', {autoAttach: true, waitForDebuggerOnStart: false, flatten: true});
-        const session = Connection.fromSession(client).session(await sessionId);
-        // TODO(einbinder): Make workers send thier conosle logs
-        return new Worker(session, this._targetInfo.url, () => void 0, () => void 0);
-      })();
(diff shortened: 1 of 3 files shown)


test/target.spec.js
```diff
@@ -83,6 +83,23 @@ module.exports.addTests = function({testRunner, expect, puppeteer}) {
       await page.evaluate(() => window.registrationPromise.then(registration => registration.unregister()));
       expect(await destroyedTarget).toBe(await createdTarget);
     });
+    it_fails_ffox('should create a worker from a service worker', async({page, server, context}) => {
+      await page.goto(server.EMPTY_PAGE);
```
aslushnikov: I guess this is not needed?

lib/Target.js
```diff
@@ -73,6 +77,30 @@ class Target {
...
   }
 
+  /**
+   * @return {!Promise<?Worker>}
+   */
+  async worker() {
+    if (this._targetInfo.type !== 'service_worker' && this._targetInfo.type !== 'shared_worker')
+      return null;
+    if (!this._workerPromise) {
+      this._workerPromise = (async() => {
```
aslushnikov: it took me a while to understand your trick 🧐

can we make it similarly to what we do with `_pagePromise`?

```js
this._workerPromise = this._sessionFactory().then(async client => {
  // Top level workers have a fake page wrapping the actual worker.
  const [targetAttached] = await Promise.all([
    new Promise(x => client.once('Target.attachedToTarget', x)),
    client.send('Target.setAutoAttach', {autoAttach: true, waitForDebuggerOnStart: false, flatten: true}),
  ]);
  const session = Connection.fromSession(client).session(targetAttached.sessionId);
  // TODO(einbinder): Make workers send their console logs.
  return new Worker(session, this._targetInfo.url, () => {} /* consoleAPICalled */, () => {} /* exceptionThrown */);
});
```

test/target.spec.js
JoelEinbinder: Done

lib/Target.js
JoelEinbinder: Done

aslushnikov (approved):
❤️

PR #4402, opened 2019-05-08
docs(firefox): Adding bug reference & prototype clarification to juggler

digitarald:
Given that the readme is the current source of truth for most users that are interested in Juggler, this tries to further clarify the experimental nature of the project and the follow up work happening on the Firefox side.

Does that reflect your assumptions as well, @aslushnikov, or should I tweak it?


aslushnikov (approved):
Sure, works for me 👍

PR #4415, opened 2019-05-10
feat(launcher): add 'detached' option to launch

CirroStorm:
This option causes launch to run the browser in a detached state and return null. You can use 'connect' to use the launched instance.

To connect to it you need to know the websocket endpoint so this is best used with the --remote-debugging-port option so that you can query the browser to get the endpoint.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -532,7 +532,8 @@ try {
   - `env` <[Object]> Specify environment variables that will be visible to the browser. Defaults to `process.env`.
   - `devtools` <[boolean]> Whether to auto-open a DevTools panel for each tab. If this option is `true`, the `headless` option will be set `false`.
   - `pipe` <[boolean]> Connects to the browser over a pipe instead of a WebSocket. Defaults to `false`.
-- returns: <[Promise]<[Browser]>> Promise which resolves to browser instance.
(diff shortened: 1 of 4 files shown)


aslushnikov:
@CirroStorm so why do you want this? Is there a related issue?

CirroStorm:
@aslushnikov , it has to do with the way I use puppeteer.

I use it for E2E testing of web apps (driven by mocha). With puppeteer the way it currently is when the test is finished and mocha exits the browser is automatically closed. 

By adding this option I can have the browser left open when the test finishes. This has two benefits:
1) If the test failed I can see the last state of the page, and use chrome's DevTools to look for error messages on the console, examine the DOM, etc. 
2) When I next run a test it will connect to the already open browser and re-use it, which is a little bit faster than spawning a new browser.

JoelEinbinder:
I think there are sufficient alternative ways to achieve this without adding it to puppeteer itself. You can use another node process to launch Chromium, you can launch Chrome manually, or leave your test runner running while you inspect the page.

As-is, this patch makes it harder to predict what the result of `puppeteer.launch` will be. It also makes it really easy for people to accidentally spawn zombie headless chromes.

aslushnikov:
I agree with Joel; let's not land this for now.

PR #4432, opened 2019-05-15
test: add SVG clicking test

aslushnikov:
Someone had concerns that pptr's clicking doesn't work with SVG.

Make sure Puppeteer's clicking does.


JoelEinbinder (approved):

JoelEinbinder:
A good test. But maybe they meant an svg with an exotic shape. Like a pretzel svg or something. We might try to click in the wrong place.

aslushnikov:
> Like a pretzel svg or something. We might try to click in the wrong place.

@JoelEinbinder Good call! And it turns out it doesn't.

PR #4433, opened 2019-05-15
fix: Page.setContent working with unicode strings

### LINKED ISSUE #4424: puppeteer from html to pdf arabic fonts are not showing
delete all npm modules and packagelock.json then do npm install arabic text is showing junk values.
Screenshot 2019-05-14 at 4 35 49 PM

semoal:
This pull request fixes: https://github.com/GoogleChrome/puppeteer/issues/4424 and this https://github.com/GoogleChrome/puppeteer/issues/4425

Explanation:

> The "Unicode Problem" Since DOMStrings are 16-bit-encoded strings, in most browsers calling window.btoa on a Unicode string will cause a Character Out Of Range exception if a character exceeds the range of a 8-bit byte (0x00~0xFF). There are two possible methods to solve this problem:
the first one is to escape the whole string (with UTF-8, see encodeURIComponent) and then encode it;
the second one is to convert the UTF-16 DOMString to an UTF-8 array of characters and then encode it.

Now accents, Cz_Cz characters, emojis, and all of this unicode stuff are working correctly.


semoal:
I signed it! 👍🏻

semoal:
I signed it!

semoal:
I signed it!!

experimental/puppeteer-firefox/lib/DOMWorld.js
```diff
@@ -134,7 +134,7 @@ class DOMWorld {
   async setContent(html) {
     await this.evaluate(base64html => {
       document.open();
-      document.write(atob(base64html));
+      document.write(decodeURIComponent(atob(base64html).split('').map(c => '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2)).join('')));
```
aslushnikov, line 137: @semoal well, this is impressive.

I have hard time understanding what this does. Can you please explain this to me?

semoal: Hi! basically, DOMString are 16-bit encoded strings, so when you call atob/btoa this methods create a exception when they exceeds the range of 8-bit byte (0x00~0xFF).
Mozilla MDN docs spoke a lot about this situation: https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding

Normally ASCII characters like ™ if this is not handled, they got converted to something extrange like: â¢, because exceed the range of bytes allowed.

aslushnikov: > Hi! basically, DOMString are 16-bit encoded strings, so when you call atob/btoa this methods create a exception when they exceeds the range of 8-bit byte (0x00~0xFF).

@semoal I don't think `atob` actually throws an exception (at least it does not for me). I also don't think your solution corresponds to any of the solutions, listed on the [MDN page](https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding#The_Unicode_Problem). 

So how do you actually fix this - from the high-level standpoint? Why don't you change the base64 encoding part?

aslushnikov: Ah ok, I think I understand now.

Can't you drop the `'00'` together with `slice(-2)`?

The following works for me:

```js
document.write(decodeURIComponent(atob(base64html).split('').map(c => '%' + c.charCodeAt(0).toString(16)).join('')));
```

semoal: Yes your solution works too mate!

semoal: I've did a playground for testing online mate, https://repl.it/@semoal/AtobTest

You can see here all the solutions we have provided

aslushnikov (approved):
@semoal Great, I understand the whole thing now. Thank you for the patch!

experimental/puppeteer-firefox/lib/DOMWorld.js
aslushnikov: @semoal the `'00'` is actually important for things like a "newline" - I added a test for this.

PelicanMaster:
I ran into this issue, and I looked for a solution too.

I wonder if the base64 part is really relevant ? Could someone explain why we have to encode the original string to base64 just to decode it right after ? 

Anyway, great solution, thanks for that.

Issue #4379, opened 2019-05-01
Respond with multiple headers in request interception


jake314159:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.15.0
* Platform / OS version: Linux Ubuntu 18.04
* Node.js version: 10.9.0

**What steps will reproduce the problem?**

Execute the following code which sets up puppeteer with request interception enabled. It attempts to respond to the request with multiple set-cookie headers (but it applies to all headers). I believe this was supported in the previous version.

```javascript
const puppeteer = require('puppeteer');
const browser = await puppeteer.launch();
const page = await browser.newPage();
page.on("request", function(r) {
  r.respond({
    status: 200,
    contentType: "text/plain",
    body: "[BODY]",
    headers: {
      "set-cookie": ["a=1", "b=2"]
    }
  });
});
await page.setRequestInterception(true);
await page.goto('https://example.com');
console.log("Success");
```

**What is the expected result?**

It should respond with multiple Set-Cookie headers ie.

```
< HTTP/1.1 200 
< content-type: text/plain
< set-content: a=1
< set-content: b=2
[BODY]
```

**What happens instead?**

```page.goto()``` will not return for 30s, at which point it'll throw a "```TimeoutError: Navigation Timeout Exceeded: 30000ms exceeded```" exception.

jake314159:
I believe it was introduced by this PR: https://github.com/GoogleChrome/puppeteer/pull/4265

lebreRafael:
I was having the same problem with the following header: `'Access-Control-Allow-Credentials': true`

This bug was introduced in the last version 1.15

Seems that the problem is that we wasn't defining the value of our headers as text. I was defining a boolean and you was defining an array.

For me the solution was simple. Just using `'true'` instead of `true`. For you I think that the best you can do is to create a single text separating your values with `;` so the header will be `"set-cookie": "a=1;b=2"`

But I think this is a bug because a wrong value in the headers should throw an error instead of never respond the request.

jasonswearingen:
@lebreRafael I have the same issue.  ***you can not set multiple cookies*** as you suggest.  (joining with "; ")  doing so results puppeteer only seeing the first cookie.

aslushnikov:
> It should respond with multiple Set-Cookie headers ie.

@jake314159  I don't think this has ever worked. The v1.14.0 was simply `JSON.stringify`'ing header values - so there was an impression that things work.

[...]

PR #4436, opened 2019-05-16
fix: Stringify response headers for intercepted requests

jake314159:
Stringifying the headers was the behaviour before v1.15

Relates to the discussion in #4379


aslushnikov:
@jake314159 Thanks! Can you please also update documentation to mention this behavior?

aslushnikov (approved):
Awesome, thanks!

PR #4440, opened 2019-05-16
chore(examples): Add wait for custom event

mflodin:
This patch adds an example that waits for a custom event to trigger

### DIFF
--- a/examples/await-custom-event.js
+++ b/examples/await-custom-event.js
@@ -1,0 +1,56 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


mflodin:
While the guidelines mention that one should not use too many comments, I added some in places where I found the original example _custom-event.js_ a bit confusing. I hope that is ok in an example. I also added some more logging to the console to make it clear in what order different parts of the code run.

aslushnikov:
Hey @mflodin, thanks for the PR.

We already have [custom-event.js](https://github.com/GoogleChrome/puppeteer/blob/master/examples/custom-event.js) example. Can you please elaborate why you want another one? Can we instead enhance the example we already have?

mflodin:
The custom-event.js example doesn't actually await the custom event. It will just log that event whenever it happens, but puppeteer will not wait for it to trigger. I guess that has it's uses too if you want to log multiple events, so I think both examples might be valid. But perhaps I should add some comments to the other example about how it behaves, and what the differences between them are? And maybe add some more logging to the other example so one can see that the order stuff happens might not be what one expects.

aslushnikov:
@mflodin Hmm. Honestly, your example is pretty involved - I don't think it's a very common usecase.

In general, we try to have just a few examples to illustrate some big concepts rather than specific usecases. Some specific scenarios are better covered in blog posts.

I'll go ahead and close this PR. Thank you for your time though!

PR #4445, opened 2019-05-19
docs(troubleshooting.md): Fixed alpine install example

aos:
First of all, thank you for your wonderful documentation on installing for Docker/Alpine. I was able to dockerize my simple script using parts of the example Dockerfile you gave. It was missing one thing though:

We `chown /app` directory but I got the error that it doesn't exist yet. I added it in the `mkdir` command and that fixed the problem.


aos:
I signed the CLA

aslushnikov (approved):
Awesome, thanks!

PR #4455, opened 2019-05-21
feat(chromium): roll Chromium to r661432

### LINKED ISSUE #4443: page.evaluate hangs on Args key has unicode char
### Steps to reproduce
Calling await page.evaluate() hangs on Args key has unicode char

* Puppeteer version: 1.16.0
* Platform / OS version: windows10 1809
* URLs (if applicable):
* Node.js version: 8.16.0

```js

(async () => {
    try {
        console.log("wait evaluate");
        await page.evaluate(() => { }, {"中文字符": 1 })
    } finally {
        browser.close();
    }
})()
```

 `await page.evaluate(() => { }, {"中文字符": 1 })` should resolve quickly !

**What happens instead?**
Script hangs on line 8

aslushnikov:
This roll includes:
- https://crrev.com/661411 - Update V8 to version 7.6.215.

The new V8 has an updated inspector-protocol dependency that correctly
parses UTF-8 strings

Fixes #4443.

### DIFF
--- a/lib/DOMWorld.js
+++ b/lib/DOMWorld.js
@@ -203,11 +203,11 @@ class DOMWorld {
     } = options;
     // We rely upon the fact that document.open() will reset frame lifecycle with "init"
     // lifecycle event. @see https://crrev.com/608658
-    await this.evaluate(base64html => {
+    await this.evaluate(html => {
       document.open();
-      document.write(decodeURIComponent(atob(base64html).split('').map(c => '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2)).join('')));
+      document.write(html);
       document.close();
-    }, Buffer.from(html).toString('base64'));
+    }, html);
     const watcher = new LifecycleWatcher(this._frameManager, this._frame, waitUntil, timeout);
     const error = await Promise.race([
       watcher.timeoutOrTerminationPromise(),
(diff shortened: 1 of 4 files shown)


lib/DOMWorld.js
```diff
@@ -203,11 +203,11 @@ class DOMWorld {
     } = options;
     // We rely upon the fact that document.open() will reset frame lifecycle with "init"
     // lifecycle event. @see https://crrev.com/608658
-    await this.evaluate(base64html => {
+    await this.evaluate(html => {
       document.open();
-      document.write(decodeURIComponent(atob(base64html).split('').map(c => '%' + ('00' + c.charCodeAt(0).toString(16)).slice(-2)).join('')));
+      document.write(html);
       document.close();
-    }, Buffer.from(html).toString('base64'));
```
JoelEinbinder, line 210: Why does this change? Or rather why was it there to begin with?

aslushnikov: Turns out this is another problem in protocol's encoder.

Patch is out for review: https://chromium-review.googlesource.com/c/deps/inspector_protocol/+/1621848

Once it lands and protocol rolls in both V8 and Chromium, we should re-roll.

Closing this PR until that.

aslushnikov:
Closing this - we'll need to pick the https://chromium-review.googlesource.com/c/deps/inspector_protocol/+/1621848 as well

PR #4459, opened 2019-05-21
Styled js code blocks in README where missing

amerikan:
- just added back ticks with js language identifier were it was missing

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,22 +195,28 @@ Puppeteer creates its own Chromium user profile which it **cleans up on every ru
    displaying. Instead of launching in headless mode, launch a full version of
    the browser using  `headless: false`:
 
+    ```js
         const browser = await puppeteer.launch({headless: false});
+    ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -195,22 +195,28 @@ Puppeteer creates its own Chromium user profile which it **cleans up on every ru
    displaying. Instead of launching in headless mode, launch a full version of
    the browser using  `headless: false`:
 
+    ```js
         const browser = await puppeteer.launch({headless: false});
```
vsemozhetbyt: You can also remove the indentation now, like this:
```suggestion
     const browser = await puppeteer.launch({headless: false});
```

aslushnikov:
@amerikan thanks for the PR! IIRC we have to keep these like this so that it imports well on developers.google.com

@ebidel - do I get it right?

amerikan:
@aslushnikov didn't know, feel free to close is such the case.

ebidel:
Yep, IIRC. Can't do GH style markdown for codeblocks for the import to work :(

PR #4461, opened 2019-05-22
docs: clarify when requestfailed event is emitted

mattzeunert:
Clarify that `requestfailed` is only emitted if there's no response, and not if the the server returns an error.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1028,7 +1028,7 @@ In order to intercept and mutate requests, see `page.setRequestInterception`.
 #### event: 'requestfailed'
 - <[Request]>
 
-Emitted when a request fails, for example by timing out.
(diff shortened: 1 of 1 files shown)


IO137:
What you need to do is download the CA Certificates package for whatever OS you're on, under Linux (well Debian/Ubuntu) it'll probably be something like

apt-get install ca-certificates
then

git config --global http.sslVerify true
git config --global http.sslCAinfo /etc/ssl/certs/ca-certificates.crt
although your path to your certificate file might be different depending on OS version.

This should get it working.

aslushnikov:
Thanks for the PR! I think I have some better wording; closing in favor of https://github.com/GoogleChrome/puppeteer/pull/4507

PR #4478, opened 2019-05-26
fix(helper): fix Helper.installAsyncStackHooks method

adrielcodeco:
Using Error.captureStackTrace instead of new Error () to avoid UnhandledPromiseRejectionWarning

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -107,7 +107,8 @@ class Helper {
       if (methodName === 'constructor' || typeof methodName !== 'string' || methodName.startsWith('_') || typeof method !== 'function' || method.constructor.name !== 'AsyncFunction')
         continue;
       Reflect.set(classType.prototype, methodName, function(...args) {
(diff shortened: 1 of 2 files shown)


lib/helper.js
```diff
@@ -107,7 +107,8 @@ class Helper {
       if (methodName === 'constructor' || typeof methodName !== 'string' || methodName.startsWith('_') || typeof method !== 'function' || method.constructor.name !== 'AsyncFunction')
         continue;
       Reflect.set(classType.prototype, methodName, function(...args) {
-        const syncStack = new Error();
+        const syncStack = {};
+        Error.captureStackTrace(syncStack);
```
aslushnikov, line 111: can you please add the same in puppeteer-firefox?

aslushnikov: nvm I hopped onto this patch

aslushnikov (approved):

PR #4486, opened 2019-05-28
fix(install): use --no-sandbox to generate protocol types

### LINKED ISSUE #4485: Cannot `yarn "install` inside container because of protocal-types-generator
### Steps to reproduce

* Puppeteer version: master (084cf021195dbe125d26496796f590a4300fb844)
* Platform / OS version: Docker container running `node:8`
* Node.js version: 8.16

1. `git clone https://github.com/GoogleChrome/puppeteer.git`
2. `yarn` (`npm` fails the same way)

Puppeteer is installed.

**What happens instead?**
Installation fails when running the install script `node install.js` because the protocol-types-generation tries to launch chrome without the `--no-sandbox` argument.

Here is the output:
```
root@494abee07b90:/app# yarn
yarn install v1.15.2
info No lockfile found.
[1/4] Resolving packages...
warning parcel > htmlnano > uncss > jsdom > left-pad@1.3.0: use String.prototype.padStart()
[2/4] Fetching packages...
info fsevents@1.2.9: The platform "linux" is incompatible with this module.
info "fsevents@1.2.9" is an optional dependency and failed compatibility check. Excluding it from installation.
[3/4] Linking dependencies...
[4/4] Building fresh packages...
[-/4] ⢀ waiting...
[4/4] ⢀ parcel
[3/4] ⢀ puppeteer
error /app/node_modules/puppeteer: Command failed.
Exit code: 1
Command: node install.js
Arguments:
Directory: /app/node_modules/puppeteer
Output:
Chromium downloaded to /app/node_modules/puppeteer/.local-chromium/linux-641577
ERROR: Failed to download Chromium r641577! Set "PUPPETEER_SKIP_CHROMIUM_DOWNLOAD" env variable to skip download.
Error: Failed to launch chrome!
[0528/160658.820235:ERROR:zygote_host_impl_linux.cc(89)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.


TROUBLESHOOTING: https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md

    at onClose (/app/node_modules/puppeteer/lib/Launcher.js:342:14)
    at Interface.helper.addEventListener (/app/node_modules/puppeteer/lib/Launcher.js:331:50)
    at emitNone (events.js:111:20)
    at Interface.emit (events.js:208:7)
    at Interface.close (readline.js:368:8)
    at Socket.onend (readline.js:147:10)
    at emitNone (events.js:111:20)
    at Socket.emit (events.js:208:7)
    at endReadableNT (_stream_readable.js:1064:12)
```

The "Fail to launch Chrome" error is misleading as the `.catch` handler ([here](https://github.com/GoogleChrome/puppeteer/blob/master/install.js#L67) also catches error from function invocation in the `onSuccess` handler.
The error is actually thrown by `onSuccess > generateProtocolTypesIfNecessary > utils/protocol-types-generator/index.js`; because it launches chrome without the `--no-sandbox` argument and this fails within a container because of root shenanigans.

note: I've also tried using the [recommended Dockerfile](https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker), but it yields the same results.

pyrho:
fixes #4485

### DIFF
--- a/utils/protocol-types-generator/index.js
+++ b/utils/protocol-types-generator/index.js
@@ -4,6 +4,7 @@ const puppeteer = require('../..');
 module.exports = puppeteer.launch({
   pipe: false,
   executablePath: process.env.CHROME,
+  args: [ '--no-sandbox' ],
 }).then(async browser => {
   const origin = browser.wsEndpoint().match(/ws:\/\/([0-9A-Za-z:\.]*)\//)[1];
   const page = await browser.newPage();


pyrho:
I signed it!

utils/protocol-types-generator/index.js
```diff
@@ -4,6 +4,7 @@ const puppeteer = require('../..');
 module.exports = puppeteer.launch({
   pipe: false,
   executablePath: process.env.CHROME,
+  args: [ '--no-sandbox' ],
```
aslushnikov, line 7: ok, there are more into this: there's not only protocol types, but also our linter, and all of our tests. You'll probably want to add this flag in all these places.

We had `--no-sandbox` everywhere some time ago. But since we figured the proper sandboxing setup and documented it in our troubleshooting guide, we dropped them all to actually validate that things work.

Adding `--no-sandbox` makes our CI story weaker - let's not do it.

aslushnikov:
Closing since this is not the direction we want to go here.

utils/protocol-types-generator/index.js
pyrho: Ok, so maybe the actual fix is to not generate the protocol types when installing puppeteer via npm/yarn ?

PR #4491, opened 2019-05-29
Update troubleshooting docs' Alpine to Chromium 73 and Puppeteer 1.12.2

r4j4h:
According to the compatibility pages these versions are the most recent supported ones


r4j4h:
Hmm, used the inline Github editor to make the commit. Looks like I may need to clone down with git to fix the message. 🤦‍♀

r4j4h:
Uh oh, linter did not see the push. I will just close this and re-open. 🤷‍♀

PR #4498, opened 2019-05-30
Add swipe action to Touchscreen

jerrydeng:
I happen to have a use case for swipe action and just follow your tap approach to make this method for Touchscreen.


JoelEinbinder:
Thanks for the patch. Unfortunately we can’t land this. Chrome is lacking a way to wait for the side effects of the swipe gesture to complete. Mainly scrolling. The compositor/input teams in chrome have been working on it. But it’s a surprisingly hard problem.

PR #4504, opened 2019-05-31
style: Update README.md

jeongukjae:
* highlights code blocks in lists
* use headings instead of emphasis
* add new lines before or after code blocks and lists
* remove duplicated new lines

---

I think it is better to use headings instead of emphasis like line 154 in `README.md`. But in some cases like line 9, cause these lines don't follow levels of headings, it is also good to use emphasis. I guess headings was used to style those lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,12 @@
 
 <img src="https://user-images.githubusercontent.com/10379601/29446482-04f7036a-841f-11e7-9872-91d1fc2ea683.png" height="200" align="right">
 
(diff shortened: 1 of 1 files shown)


aslushnikov:
> highlights code blocks in lists

@jeongukjae These are not highlighted because our README.md is imported to developers.google.com, and there are some integration troubles with having these blocks highlighted.

Overall, I think the style is a matter of taste. Currently, things look fine to me - so let's keep things as-is.

Thank you for the PR though!

PR #4508, opened 2019-06-01
feat: report iframe target types

aslushnikov:
This is necessary so that we (and our users) can diagnose OOPIFs
and OOPIF-related issues.

Add a test to make sure that `--site-per-process` chromium flag actually forces iframes to be out-of-process.

### DIFF
--- a/lib/Target.js
+++ b/lib/Target.js
@@ -110,7 +110,7 @@ class Target {
    */
   type() {
     const type = this._targetInfo.type;
-    if (type === 'page' || type === 'background_page' || type === 'service_worker' || type === 'shared_worker' || type === 'browser')
+    if (type === 'page' || type === 'background_page' || type === 'service_worker' || type === 'shared_worker' || type === 'browser' || type === 'iframe')
       return type;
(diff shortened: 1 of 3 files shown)


lib/Target.js
```diff
@@ -106,11 +106,11 @@ class Target {
   }
 
   /**
-   * @return {"page"|"background_page"|"service_worker"|"shared_worker"|"other"|"browser"}
+   * @return {"page"|"background_page"|"service_worker"|"shared_worker"|"other"|"browser"|"iframe"}
```
JoelEinbinder, line 109: How bad does it look if we don't add this?
I don't think any puppeteer user should ever connect to an oopif directly. Can we just use _targetInfo.type in the test?

aslushnikov: Sure! Closing this in favor of https://github.com/GoogleChrome/puppeteer/pull/4516

aslushnikov:
Closing in favor of https://github.com/GoogleChrome/puppeteer/pull/4516

PR #4528, opened 2019-06-05
docs(readme.md): add missing code highlighting

VadimDez:


aslushnikov:
@VadimDez  thank you Vadim for the PR!

The lack of highlight is actually intentional - otherwise auto-export to https://developers.google.com won't work. I'll have to close this without merging.

PR #4532, opened 2019-06-06
feat(launcher): Resolve remote debugging address

hason:
I use remote chromium in docker container and puppeteer in another container. I use option --remote-debugging-address=chromium (docker compose service name). It is necessary to resolve this address.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -18,6 +18,7 @@ const path = require('path');
 const http = require('http');
 const https = require('https');
 const URL = require('url');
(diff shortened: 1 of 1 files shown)


JoelEinbinder (commented):
I’m not in favor of transforming the users chrome arguments. Why can’t you do the transformation yourself before launching puppeteer?

aslushnikov:
> Why can’t you do the transformation yourself before launching puppeteer?

Closing this since we don't understand the necessity.

Issue #4536, opened 2019-06-06
Dangling promises / unhandled rejection due to missing await keywords in mouse.click


pqvst:
The `page.mouse.click` helper function lacks `await` on `this.move` and `this.up` which leads to dangling promises and unhandled rejections that are not possible to catch on the caller site if these calls fail (e.g. if the page is closed).

https://github.com/GoogleChrome/puppeteer/blob/master/lib/Input.js#L227-L228

### Steps to reproduce

```
page.close();
try {
  await page.click(1, 2);
} catch (err) {
  // catches this.up() but not this.move() and this.down()
  // unhandled rejection will occur despite try/catch...
}
```

PR #4541, opened 2019-06-07
fix(page): fix missing awaits in mouse.click #4536

pqvst:
Lack of await causes dangling promises, which lead to unhandled rejections that cannot be caught if these calls fail (e.g. if the page is closed). See issue #4536 for details.

### DIFF
--- a/lib/Input.js
+++ b/lib/Input.js
@@ -224,11 +224,20 @@ class Mouse {
    */
   async click(x, y, options = {}) {
     const {delay = null} = options;
-    await this.move(x, y);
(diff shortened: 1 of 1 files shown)


pqvst:
I signed it!

aslushnikov (approved):
We sit with @JoelEinbinder and ended up with this. 🤷‍♂️

aslushnikov:
I'll follow-up with a test and a firefox-fix.

edit: follow-up is https://github.com/GoogleChrome/puppeteer/pull/4561

PR #4547, opened 2019-06-08
chore: highlights syntax in README

sobolevn:


aslushnikov:
@sobolevn thank you for the PR!

Unfortunately, these are not highlighted intentionally so that our export to developers.google.com works as expected.

PR #4551, opened 2019-06-08
fix(eval): be able to throw protocol like errors

JoelEinbinder:
sorry

### DIFF
--- a/test/evaluation.spec.js
+++ b/test/evaluation.spec.js
@@ -182,6 +182,12 @@ module.exports.addTests = function({testRunner, expect}) {
       });
       expect(result).toBe(undefined);
     });
+    it('should be able to throw a tricky error', async({page}) => {
+      const error = await page.evaluate(() => {
+        throw new Error('Object reference chain is too long');
+      }).catch(e => e);
+      expect(error.message).toContain('Object reference chain is too long');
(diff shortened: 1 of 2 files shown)


test/evaluation.spec.js
```diff
@@ -182,6 +182,12 @@ module.exports.addTests = function({testRunner, expect}) {
       });
       expect(result).toBe(undefined);
     });
+    it('should be able to throw a tricky error', async({page}) => {
+      const error = await page.evaluate(() => {
+        throw new Error('Object reference chain is too long');
```
aslushnikov: Let's get the error text dynamically - this way it'll be a cross-browser test.

```js
    it('should be able to throw a tricky error', async({page, server}) => {
      const windowHandle = await page.evaluateHandle(() => window);
      const errorText = await windowHandle.jsonValue().catch(e => e.message);
      const error = await page.evaluate(errorText => {
        throw new Error(errorText);
      }, errorText).catch(e => e);
      expect(error.message).toBe(errorText);
    });
```

aslushnikov (approved):

aslushnikov:
Interestingly enough, this test won't work in Firefox 🤷‍♂️

PR #4553, opened 2019-06-09
docs(api.md): clarify frame/page.goto()'s handling of HTTP error stat…

michaeloryl:
…uses

The documentation for frame.goto() and page.goto() were updated to make
it clear that the method will not throw an error if the HTTP requests
results in any valid HTTP status code being returned by the remote
server.


aslushnikov (approved):

aslushnikov:
Thanks!

PR #4567, opened 2019-06-11
fix(network): align HTTP status codes with IANA

aslushnikov:
Aline HTTP status texts with https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml


JoelEinbinder:
They are missing some that we have, like I’m a teapot. I feel like we are entering a scary world.

aslushnikov:
> They are missing some that we have, like I’m a teapot. I feel like we are entering a scary world.

These are added

lib/NetworkManager.js
```diff
@@ -773,9 +776,12 @@ const STATUS_TEXTS = {
   '422': 'Unprocessable Entity',
   '423': 'Locked',
   '424': 'Failed Dependency',
+  '425': 'Too Early',
   '426': 'Upgrade Required',
+  '427': 'Unassigned',
   '428': 'Precondition Required',
   '429': 'Too Many Requests',
+  '430': 'Unassigned',
```
JoelEinbinder: unassigned just means unassigned, its not a specific code.

JoelEinbinder (approved):

lib/NetworkManager.js
aslushnikov: indeed!

PR #4569, opened 2019-06-11
fix(page): fix A2 and A3 format size

### LINKED ISSUE #4568: A3 paper size is incorrect
I was trying to generate PDF using this library. However, I noticed the size of A3 format is incorrect. The correct size should be **297mm** x **420mm**, but the generated PDF size is **297mm** x **419mm**.

### Steps to reproduce
Call `pdf` function with param `{ format: 'A3' }`.

* Puppeteer version: 1.17.0
* Platform / OS version: Mac OS High Sierra / Linux
* URLs (if applicable): 
* Node.js version: 10.x

The A3 format size should be **297mm** x **420mm**.

sapics:
This PR fixes #4568.

16.5 inch ~ 419.1mm
16.54 inch ~ 420.1mm


aslushnikov (approved):
Fair, thank you!

PR #4571, opened 2019-06-11
fix(connection): bump websocket max received message size to 256Mb

### LINKED ISSUE #4543: Cannot retrieve content of pages that are >100MB.
Puppeteer >=1.11.0 cannot retrieve the content of HTML pages that are >100MB.

In Puppeteer 1.11.0, the `ws` dependency was bumped to ^6.1.0 (See: https://github.com/GoogleChrome/puppeteer/commit/d3f50ea425af38516f671a22e9d42363e7d681c5). However, `ws` introduced a breaking change in 6.0.0 by adding a `maxPayload` option that capped WebSocket message sizes to 100MB by default (See: https://github.com/websockets/ws/pull/1402) and Puppeteer relies on the `ws` defaults (See: https://github.com/GoogleChrome/puppeteer/blob/9c4b6d06e214946e38999b9325c7d10152a1cf69/lib/WebSocketTransport.js#L28)

### Suggested Fix

Increasing the `maxPayload` option in `ws` allows us to circumvent this issue. I would suggest allowing users to customize `ws` options in Puppeteer to set a custom `maxPayload` size. Thanks!

### Steps to reproduce

**Environment:**

* Puppeteer version: >=1.11.0

**Steps to reproduce**

1. Create a `test.html` file >100MB in size. Quick example in Python:
```py
contents = 'a' * 100 * 1024 * 1024
styling = '<style type="text/css">body{display:none}</style>'
with open('test.html', 'w') as file:
    file.write(
        f'<!DOCTYPE html><html><head>{styling}</head><body>{contents}</body></html>'
    )
```
(Note: display set to none in test file purely to speed up loading of page. It can be omitted, but please note Puppeteer will take longer to load the page when trying to reproduce this issue).

2.  Open the large `test.html` file and try to retrieve page content (Add in path to your `test.html` file in below script):
```js

(async () => {
    const browser = await puppeteer.launch({ args: ["--no-sandbox"] });
    await page.goto("file://{path to test.html}", { timeout: 100000 });
    content = await page.content();
    await page.close();
    browser.close();
})();
```

Page content is retrievable without error.

**What happens instead?**

We see the following traceback:
```
(node:86223) UnhandledPromiseRejectionWarning: Error: Protocol error (Runtime.callFunctionOn): Target closed.
    at Promise (~/node_modules/puppeteer/lib/Connection.js:183:56)
    at new Promise (<anonymous>)
    at CDPSession.send (~/node_modules/puppeteer/lib/Connection.js:182:12)
    at ExecutionContext.evaluateHandle (~/node_modules/puppeteer/lib/ExecutionContext.js:106:44)
    at ExecutionContext.<anonymous> (~/node_modules/puppeteer/lib/helper.js:111:23)
    at ExecutionContext.evaluate (~/node_modules/puppeteer/lib/ExecutionContext.js:48:31)
    at ExecutionContext.<anonymous> (~/node_modules/puppeteer/lib/helper.js:111:23)
    at DOMWorld.evaluate (~/node_modules/puppeteer/lib/DOMWorld.js:112:20)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
  -- ASYNC --
    at Frame.<anonymous> (~/node_modules/puppeteer/lib/helper.js:110:27)
    at Page.content (~/node_modules/puppeteer/lib/Page.js:612:49)
    at Page.<anonymous> (~/node_modules/puppeteer/lib/helper.js:111:23)
    at ~/test.js:7:26
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
```

aslushnikov:
This is the max message size that DevTools can emit over the DevTools
protocol: https://cs.chromium.org/chromium/src/content/browser/devtools/devtools_http_handler.cc?type=cs&q=kSendBufferSizeForDevTools&sq=package:chromium&g=0&l=83

Test is failing on firefox since Firefox crashes when allocating 100Mb string.

Fix #4543

### DIFF
--- a/lib/WebSocketTransport.js
+++ b/lib/WebSocketTransport.js
@@ -25,7 +25,10 @@ class WebSocketTransport {
    */
   static create(url) {
     return new Promise((resolve, reject) => {
-      const ws = new WebSocket(url, [], { perMessageDeflate: false });
+      const ws = new WebSocket(url, [], {
+        perMessageDeflate: false,
+        maxPayload: 256 * 1024 * 1024, // 256Mb
+      });
       ws.addEventListener('open', () => resolve(new WebSocketTransport(ws)));
       ws.addEventListener('error', reject);
     });
--- a/test/evaluation.spec.js
+++ b/test/evaluation.spec.js
@@ -247,6 +247,10 @@ module.exports.addTests = function({testRunner, expect}) {
       });
       expect(result).toEqual([42]);
     });
+    it_fails_ffox('should transfer 100Mb of data from page to node.js', async({page, server}) => {
+      const a = await page.evaluate(() => Array(100 * 1024 * 1024 + 1).join('a'));
+      expect(a.length).toBe(100 * 1024 * 1024);
+    });
   });
 
   describe('Page.evaluateOnNewDocument', function() {


lib/WebSocketTransport.js
```diff
@@ -25,7 +25,10 @@ class WebSocketTransport {
    */
   static create(url) {
     return new Promise((resolve, reject) => {
-      const ws = new WebSocket(url, [], { perMessageDeflate: false });
+      const ws = new WebSocket(url, [], {
+        perMessageDeflate: false,
+        maxPayload: 256 * 1024 * 1024, // 256Mb
```
JoelEinbinder, line 30: why is there a max payload at all? What would happen if there was no max and someone tried to send 8 gb?

JoelEinbinder (approved):

lib/WebSocketTransport.js
aslushnikov: I guess to protect against process crashes.

PR #4587, opened 2019-06-14
feat(page): move page.pdf to protocol streams

### LINKED ISSUE #4563: Too large write data is pending: size=725124416, max_buffer_size=268435456
### Steps to reproduce

* Puppeteer version: 1.2.0
* Platform / OS version: Windows Server 2008
* URLs (if applicable):
* Node.js version: 8.9.1

When printing web pages that contains hundreds of images.
Even tried the "pipe" mode but no avail. If this is a limitation, is there a way to find the size of the write data as it is obviously printing the following statement on the console window

**[0609/223146.069:ERROR:http_connection.cc(112)] Too large write data is pending: size=725124416, max_buffer_size=268435456**

Tried **Network.dataReceived** option to read the data lengths to get to the overall size but the data size does not match. In my case, I got the data as 521 MB but the write data says 691 MB. 

Attached a picture of the **Network** tab on the devTools. It says 521 MB resources which is correctly matching my reading. However, it says **252 MB** transferred. How do I get this reading? Either way, how do I get the write data size?
Screen Shot 2019-06-10 at 9 01 44 AM

Successfully generate the PDF irrespective of the size of the write data. Else, way to know the size of the write data so that we can document the known limitations.

**What happens instead?**
Fails to generate the PDF when the page contains of hundreds of images. In other words, when the write data is too large.

aslushnikov:
This lets transferring massive PDF files over the protocol.

Fix #4563

### DIFF
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -226,6 +226,7 @@ class Helper {
    * @param {!Puppeteer.CDPSession} client
    * @param {string} handle
    * @param {?string} path
+   * @return {!Buffer}
    */
   static async readProtocolStream(client, handle, path) {
     let eof = false;
--- a/lib/helper.js
+++ b/lib/helper.js
@@ -14,8 +14,8 @@
  * limitations under the License.
  */
 const {TimeoutError} = require('./Errors');
-
 const debugError = require('debug')(`puppeteer:error`);
+const fs = require('fs');
 
 class Helper {
   /**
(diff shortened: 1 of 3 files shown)


lib/helper.js
```diff
@@ -221,8 +221,42 @@ class Helper {
       clearTimeout(timeoutTimer);
     }
   }
+
+  /**
+   * @param {!Puppeteer.CDPSession} client
+   * @param {string} handle
+   * @param {?string} path
```
JoelEinbinder, line 228: `@return`

JoelEinbinder (approved):
this will break pdfs against stable for awhile. Not a problem, but something people will complain about anyway.

nobleach:
>Not a problem, but something people will complain about anyway.

I imagine folks will complain if the API changes and it's not documented anywhere.

Am I correct to assume that `page.pdf()` no longer returns a `<Promise<Buffer>>`?

PR #4591, opened 2019-06-15
feat(chromium): roll Chromium to r669486

aslushnikov:
This roll includes:
- https://crrev.com/c/1589672/ - Include retired instruction delta in TraceEvents

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "node": ">=6.4.0"
   },
   "puppeteer": {
-    "chromium_revision": "668716"
+    "chromium_revision": "669486"
   },
   "scripts": {
     "unit": "node test/test.js",


JoelEinbinder (approved):
I’m not sure what that is. Are we going to turn it on?

aslushnikov:
+cc @ai @mbalabash


This rolls in a new Chromium into Puppeteer that supports retired instruction counter on **Linux** (other platforms are not supported yet).

More context: https://twitter.com/aslushnikov/status/1121816765518897152

Once this lands, you'll be able to record a trace with RTI like this:
```js
(async () => {
  const browser = await puppeter.launch({
    // --no-sandbox is needed so that renderer process has access to RTI
    args: [ '--enable-thread-instruction-count', '--no-sandbox'],
  });
  await page.tracing.start({path: './trace.json'});
  // run scenario
  // ...
  await page.tracing.stop();
  await browser.close();
}();
```

The resulting trace has a new `'tidelta'` parameter that contains RTI.

@JoelEinbinder: I hope this answers your question too!

aslushnikov:
Thank you @acomminos for your hard work exposing RTI and landing patches to Chromium! 🎉

PR #4653, opened 2019-07-01
feat(page): introduce page.waitForFileChooser

### LINKED ISSUE #2946: Handling File Input Dialogs
Is there any way to handle File Input Dialogs?

When I click an element on the page, it opens up the upload file dialog.

I can not use `elementHandle.uploadFile` as it is not an `input[type="file"]`. Plus, #1376 will not help as, dragging and dropping in the file upload field does not work, instead, dragging and dropping opens up the file in the browser tab.

I am using Windows, so there won't be the problem where, file dialog behavior is different across operating systems.

So, can you provide some example code on how would I go about doing that? Thanks!

aslushnikov:
This patch introduces a `page.waitForFileChooser()` method
that adds a watchdog to wait for file chooser dialogs.

This lets Puppeteer users to capture file chooser requests
and fulfill/cancel them if necessary:

```js
const [fileChooser] = await Promise.all([
  // await for upcoming filechooser
  page.waitForFileChooser(),
  // some button that triggers file picking.
  page.click('#upload-file-button'),
]);
// Pretend as if user has selected '/tmp/foo.pdf'.
await fileChooser.accept('/tmp/foo.pdf');
```

> **NOTE** In non-headless Chromium, waiting for file chooser results in the native file picker dialog **not showing up**.

Fix #2946

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -152,6 +152,7 @@
   * [page.url()](#pageurl)
   * [page.viewport()](#pageviewport)
   * [page.waitFor(selectorOrFunctionOrTimeout[, options[, ...args]])](#pagewaitforselectororfunctionortimeout-options-args)
+  * [page.waitForFileChooser([options])](#pagewaitforfilechooseroptions)
   * [page.waitForFunction(pageFunction[, options[, ...args]])](#pagewaitforfunctionpagefunction-options-args)
   * [page.waitForNavigation([options])](#pagewaitfornavigationoptions)
   * [page.waitForRequest(urlOrPredicate[, options])](#pagewaitforrequesturlorpredicate-options)
@@ -182,6 +183,10 @@
 - [class: Tracing](#class-tracing)
   * [tracing.start([options])](#tracingstartoptions)
   * [tracing.stop()](#tracingstop)
+- [class: FileChooser](#class-filechooser)
+  * [fileChooser.accept(filePaths)](#filechooseracceptfilepaths)
+  * [fileChooser.cancel()](#filechoosercancel)
+  * [fileChooser.isMultiple()](#filechooserismultiple)
 - [class: Dialog](#class-dialog)
   * [dialog.accept([promptText])](#dialogacceptprompttext)
   * [dialog.defaultValue()](#dialogdefaultvalue)
@@ -1727,6 +1732,7 @@ This setting will change the default maximum time for the following methods and 
 - [page.reload([options])](#pagereloadoptions)
 - [page.setContent(html[, options])](#pagesetcontenthtml-options)
 - [page.waitFor(selectorOrFunctionOrTimeout[, options[, ...args]])](#pagewaitforselectororfunctionortimeout-options-args)
+- [page.waitForFileChooser([options])](#pagewaitforfilechooseroptions)
 - [page.waitForFunction(pageFunction[, options[, ...args]])](#pagewaitforfunctionpagefunction-options-args)
 - [page.waitForNavigation([options])](#pagewaitfornavigationoptions)
 - [page.waitForRequest(urlOrPredicate[, options])](#pagewaitforrequesturlorpredicate-options)
@@ -1913,6 +1919,28 @@ await page.waitFor(selector => !!document.querySelector(selector), {}, selector)
 ```
 
 Shortcut for [page.mainFrame().waitFor(selectorOrFunctionOrTimeout[, options[, ...args]])](#framewaitforselectororfunctionortimeout-options-args).
+
+#### page.waitForFileChooser([options])
+- `options` <[Object]> Optional waiting parameters
+  - `timeout` <[number]> Maximum wait time in milliseconds, defaults to 30 seconds, pass `0` to disable the timeout. The default value can be changed by using the [page.setDefaultTimeout(timeout)](#pagesetdefaulttimeouttimeout) method.
+- returns: <[Promise]<[FileChooser]>> A promise that resolves after a page requests a file picker.
+
+> **NOTE** In non-headless Chromium, this method results in the native file picker dialog **not showing up** for the user.
+
+This method is typically coupled with an action that triggers file choosing.
+The following example clicks a button that issues a file chooser, and then
+responds with `/tmp/myfile.pdf` as if a user has selected this file.
+
+```js
+const [fileChooser] = await Promise.all([
+  page.waitForFileChooser(),
+  page.click('#upload-file-button'), // some button that triggers file selection
+]);
+await fileChooser.accept(['/tmp/myfile.pdf']);
+```
+
+> **NOTE** This must be called *before* the file chooser is launched. It will not return a currently active file chooser.
+
 
 #### page.waitForFunction(pageFunction[, options[, ...args]])
 - `pageFunction` <[function]|[string]> Function to be evaluated in browser context
@@ -2350,6 +2378,37 @@ Only one trace can be active at a time per browser.
 
 #### tracing.stop()
 - returns: <[Promise]<[Buffer]>> Promise which resolves to buffer with trace data.
+
+### class: FileChooser
+
+[FileChooser] objects are returned via the ['page.waitForFileChooser'](#pagewaitforfilechooseroptions) method.
+
+File choosers let you react to the page requesting for a file.
+
+An example of using [FileChooser]:
+
+```js
+const [fileChooser] = await Promise.all([
+  page.waitForFileChooser(),
+  page.click('#upload-file-button'), // some button that triggers file selection
+]);
+await fileChooser.accept(['/tmp/myfile.pdf']);
+```
+
+> **NOTE** In browsers, only one file chooser can be opened at a time.
+> All file choosers must be accepted or canceled. Not doing so will prevent subsequent file choosers from appearing.
+
+#### fileChooser.accept(filePaths)
+- `filePaths` <[Array]<[string]>> Accept the file chooser request with given paths. If some of the  `filePaths` are relative paths, then they are resolved relative to the [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
+- returns: <[Promise]>
+
+#### fileChooser.cancel()
+- returns: <[Promise]>
+
+Closes the file chooser without selecting any files.
+
+#### fileChooser.isMultiple()
+- returns: <[boolean]> Whether file chooser allow for [multiple](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#attr-multiple) file selection.
 
 ### class: Dialog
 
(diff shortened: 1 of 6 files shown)


JoelEinbinder:
> This patch introduces a `page.introudceFileChooser()` method
> that implements interception of file chooser dialogs.

s/introduce/intercept

docs/api.md
```diff
@@ -2351,6 +2375,34 @@ Only one trace can be active at a time per browser.
...
+- `filePaths` <[Array]<[string]>> Accept file picker request with given paths. If some of the  `filePaths` are relative paths, then they are resolved relative to the [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
+- returns: <[Promise]>
+
+#### fileChooser.cancel()
+- returns: <[Promise]>
+
+Cancel file chooser requests. As a result, a follow-up file choosers might be shown.
+
+#### fileChooser.isMultiple()
+- returns: <[boolean]> Whether file chooser allow for multiple file selection.
```
JoelEinbinder: add: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#attr-multiple

docs/api.md
```diff
@@ -2351,6 +2375,34 @@ Only one trace can be active at a time per browser.
...
+```
+
+#### fileChooser.accept(filePaths)
+- `filePaths` <[Array]<[string]>> Accept file picker request with given paths. If some of the  `filePaths` are relative paths, then they are resolved relative to the [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
+- returns: <[Promise]>
+
+#### fileChooser.cancel()
+- returns: <[Promise]>
+
+Cancel file chooser requests. As a result, a follow-up file choosers might be shown.
```
JoelEinbinder: What does this mean? When should I call this?

docs/api.md
```diff
@@ -2351,6 +2375,34 @@ Only one trace can be active at a time per browser.
...
+```js
+const [fileChooser] = await Promise.all([
+  page.interceptFileChooser(),
+  page.click('#upload-file-button'), // some button that triggers file selection
+]);
+await fileChooser.accept(['/tmp/myfile.pdf']);
+```
+
+#### fileChooser.accept(filePaths)
+- `filePaths` <[Array]<[string]>> Accept file picker request with given paths. If some of the  `filePaths` are relative paths, then they are resolved relative to the [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
```
JoelEinbinder: Accept the file chooser request.

docs/api.md
```diff
@@ -2351,6 +2375,34 @@ Only one trace can be active at a time per browser.
 #### tracing.stop()
 - returns: <[Promise]<[Buffer]>> Promise which resolves to buffer with trace data.
 
+### class: FileChooser
+
+[FileChooser] objects are dispatched by page via the ['page.interceptFileChooser'](#page-interceptfilechooser) method.
```
JoelEinbinder: The word dispatch is misleading, because we don't have events for this at the moment. They are returned.

docs/api.md
```diff
@@ -1492,6 +1497,25 @@ If there's no element matching `selector`, the method throws an error.
 
 Shortcut for [page.mainFrame().hover(selector)](#framehoverselector).
 
+#### page.interceptFileChooser()
```
JoelEinbinder: interceptNextFileChooser? I think the name to signify that it only intercepts one.

lib/Page.js
```diff
@@ -125,12 +128,42 @@ class Page extends EventEmitter {
...
     client.on('Performance.metrics', event => this._emitMetrics(event));
     client.on('Log.entryAdded', event => this._onLogEntryAdded(event));
+    client.on('Page.fileChooserOpened', event => this._onFileChooser(event));
     this._target._isClosedPromise.then(() => {
       this.emit(Events.Page.Close);
       this._closed = true;
     });
   }
 
+  _onFileChooser(event) {
```
JoelEinbinder, line 148: jsdoc

lib/Page.js
```diff
@@ -115,6 +117,7 @@ class Page extends EventEmitter {
     networkManager.on(Events.NetworkManager.Response, event => this.emit(Events.Page.Response, event));
     networkManager.on(Events.NetworkManager.RequestFailed, event => this.emit(Events.Page.RequestFailed, event));
     networkManager.on(Events.NetworkManager.RequestFinished, event => this.emit(Events.Page.RequestFinished, event));
+    this._fileChooserInterceptors = new Set();
```
JoelEinbinder, line 115: Should this be a set? Does it make sense to have multiple file chooser interceptors? Maybe we should throw?

lib/Page.js
```diff
@@ -1258,5 +1291,48 @@ class ConsoleMessage {
...
+  isMultiple() {
+    return this._multiple;
+  }
+
+  /**
+   * @param {!Array<string>} filePaths
+   * @return {!Promise}
+   */
+  async accept(filePaths) {
+    assert(!this._handled, 'Cannot accept dialog which is already handled!');
```
JoelEinbinder: s/dialog/FileChooser here and below

lib/helper.js
```diff
@@ -214,7 +214,7 @@ class Helper {
     let reject;
     const timeoutError = new TimeoutError(`waiting for ${taskName} failed: timeout ${timeout}ms exceeded`);
     const timeoutPromise = new Promise((resolve, x) => reject = x);
-    const timeoutTimer = setTimeout(() => reject(timeoutError), timeout);
+    const timeoutTimer = timeout ? setTimeout(() => reject(timeoutError), timeout) : 0;
```
JoelEinbinder: how did this come up? A quick glance says it will change the behavior of some public facing methods.

test/input.spec.js
```diff
@@ -35,4 +37,150 @@ module.exports.addTests = function({testRunner, expect}) {
...
+    it('should work for "multiple"', async({page, server}) => {
+      await page.setContent(`<input multiple type=file>`);
+      const [chooser] = await Promise.all([
+        page.interceptFileChooser(),
+        page.click('input'),
+      ]);
+      expect(chooser.isMultiple()).toBe(true);
+    });
+    it('should work for "webkitdirectory"', async({page, server}) => {
+      await page.setContent(`<input multiple type=file>`);
```
JoelEinbinder: wheres webkitdirectory?

test/input.spec.js
```diff
@@ -35,4 +37,150 @@ module.exports.addTests = function({testRunner, expect}) {
...
+          const el = document.createElement('input');
+          el.type = 'file';
+          el.click();
+        }, 50))
+      ]);
+      expect(chooser).toBeTruthy();
+    });
+  });
+
+  describe('FileChooser.accept', function() {
```
JoelEinbinder: Let's have a test with multiple interceptors waiting for a single file chooser.

test/input.spec.js
```diff
@@ -35,4 +37,150 @@ module.exports.addTests = function({testRunner, expect}) {
...
+        }),
+      ]);
+      expect(chooser).toBeTruthy();
+    });
+    it('should respect timeout', async({page, server}) => {
+      let error = null;
+      await page.interceptFileChooser({timeout: 1}).catch(e => error =e);
+      expect(error).toBeInstanceOf(puppeteer.errors.TimeoutError);
+    });
+    it('should respect default timeout', async({page, server}) => {
```
JoelEinbinder: This doesnt test what it says to test if our test timeout > the otherwise default timeout.

test/input.spec.js
```diff
@@ -35,4 +37,150 @@ module.exports.addTests = function({testRunner, expect}) {
...
+      await chooser.accept([
+        path.relative(process.cwd(), __dirname + '/assets/file-to-upload.txt'),
+        path.relative(process.cwd(), __dirname + '/assets/pptr.png'),
+      ]).catch(e => error = e);
+      expect(error).not.toBe(null);
+    });
+  });
+
+  describe('FileChooser.cancel', function() {
+    it('should cancel dialog', async({page, server}) => {
```
JoelEinbinder, line 168: add a double-cancel test

docs/api.md

docs/api.md
aslushnikov: whenever you want to cancel file chooser dialog. How do you wish this worded?

docs/api.md

docs/api.md

docs/api.md
aslushnikov: `interceptFileChooser` is already clear enough - no need to make it more verbose.

lib/Page.js
aslushnikov: Done.

lib/Page.js
aslushnikov: If we were to have an event - this would behave as a set. This is done for consistency with other methods, e.g. `page.waitForRequest`.

lib/Page.js

lib/helper.js
aslushnikov: The method has never been used before (uh, tsc, where are you...)

test/input.spec.js
aslushnikov: Done.

test/input.spec.js
aslushnikov: Done.

test/input.spec.js
aslushnikov: Done.

test/input.spec.js
aslushnikov: Done.

docs/api.md
```diff
@@ -2351,6 +2375,34 @@ Only one trace can be active at a time per browser.
 #### tracing.stop()
 - returns: <[Promise]<[Buffer]>> Promise which resolves to buffer with trace data.
 
+### class: FileChooser
+
+[FileChooser] objects are returned by page via the ['page.interceptFileChooser'](#page-interceptfilechooser) method.
```
vsemozhetbyt:
```suggestion
[FileChooser] objects are returned by page via the ['page.interceptFileChooser'](#pageinterceptfilechooseroptions) method.
```

docs/api.md
JoelEinbinder: Closes the file chooser without selecting any files.
Somewhere:
> Note: All file choosers must be accepted or canceled. Not doing so will prevent subsequent file choosers from appearing.

Whats the difference between accept([]) and cancel()? Maybe we can merge them into one method, like `fileChooser.respond`?

lib/Page.js
JoelEinbinder: calling `waitForRequest` doesn't have side effects though. If multiple people are waiting for a file chooser, who calls cancel/accept?

Your explanation should go in the docs at least.

JoelEinbinder:
Offline you mentioned favoring `waitForFileChooser`. This isn't blocked on my review right?

docs/api.md
aslushnikov: > Whats the difference between accept([]) and cancel()? Maybe we can merge them into one method, like fileChooser.respond?

The `fileChooser.cancel()` is a user cancelling file chooser, which is a vendor-defined behavior. In Chrome, this will also reset selected filed; it will not however do so in Firefox.

The `fileChooser.accept([])` is a vendor-agnostic way to reset input value.

docs/api.md
aslushnikov: Classic! Thanks!

docs/api.md
```diff
@@ -1914,6 +1920,26 @@ await page.waitFor(selector => !!document.querySelector(selector), {}, selector)
...
+
+```js
+const [fileChooser] = await Promise.all([
+  page.waitForFileChooser(),
+  page.click('#upload-file-button'), // some button that triggers file selection
+]);
+await fileChooser.accept(['/tmp/myfile.pdf']);
+```
+
+> **NOTE** In non-headless Chromium, this method results in the native file picker dialog **not showing up** to the user.
```
JoelEinbinder: s/to the user/for the user

docs/api.md
```diff
@@ -2351,6 +2377,36 @@ Only one trace can be active at a time per browser.
 #### tracing.stop()
 - returns: <[Promise]<[Buffer]>> Promise which resolves to buffer with trace data.
 
+### class: FileChooser
+
+[FileChooser] objects are returned by page via the ['page.waitForFileChooser'](#pageinterceptfilechooseroptions) method.
```
JoelEinbinder: ~~by page~~

docs/api.md
```diff
@@ -1914,6 +1920,26 @@ await page.waitFor(selector => !!document.querySelector(selector), {}, selector)
 
 Shortcut for [page.mainFrame().waitFor(selectorOrFunctionOrTimeout[, options[, ...args]])](#framewaitforselectororfunctionortimeout-options-args).
 
+#### page.waitForFileChooser([options])
```
JoelEinbinder, line 1923: waitForNextFileChooser

If there is a currently active file chooser, I'd expect it to be returned by `waitForFileChooser`.

docs/api.md
```diff
@@ -2351,6 +2377,36 @@ Only one trace can be active at a time per browser.
...
+> **NOTE** All file choosers must be accepted or canceled. Not doing so will prevent subsequent file choosers from appearing.
+
+#### fileChooser.accept(filePaths)
+- `filePaths` <[Array]<[string]>> Accept the file chooser request with given paths. If some of the  `filePaths` are relative paths, then they are resolved relative to the [current working directory](https://nodejs.org/api/process.html#process_process_cwd).
+- returns: <[Promise]>
+
+#### fileChooser.cancel()
+- returns: <[Promise]>
+
+Closes the file chooser without selecting any files.
```
JoelEinbinder, line 2408: note: only one file chooser can be open at a time.

lib/Page.js
```diff
@@ -125,12 +123,59 @@ class Page extends EventEmitter {
...
     });
   }
 
+  async _initialize() {
+    await Promise.all([
+      this._frameManager.initialize(),
+      this._client.send('Target.setAutoAttach', {autoAttach: true, waitForDebuggerOnStart: false, flatten: true}),
+      this._client.send('Performance.enable', {}),
+      this._client.send('Log.enable', {}),
+      this._client.send('Page.setInterceptFileChooserDialog', {enabled: true}).catch(e => {
```
JoelEinbinder, line 139: this can fail?! needs a comment explaining.

lib/Page.js
```diff
@@ -125,12 +123,59 @@ class Page extends EventEmitter {
...
+      interceptor.call(null, fileChooser);
+  }
+
+  /**
+   * @param {!{timeout?: number}=} options
+   * @return !Promise<!FileChooser>}
+   */
+  async waitForFileChooser(options = {}) {
+    if (this._fileChooserInterceptionIsDisabled)
+      throw new Error('File chooser handling does not work with multiple connections to the same page');
```
JoelEinbinder, line 166: this note needs to be in the docs.

aslushnikov: > this note needs to be in the docs.

I don't think this is worth mentioning. Or is there any legit usecase to connect to automation agents to the same page?

JoelEinbinder: I think that if we support multi client, it gets a note. But I don't feel strongly. `waitForNextFileChooser` is blocking the lgtm

lib/Page.js
aslushnikov: We have tests for this.

docs/api.md
JoelEinbinder: At least add

**NOTE** This must be called BEFORE the file chooser is launched. It will not return a currently active file chooser.

JoelEinbinder: And if in the future we got the ability to return the currently active file picker, would we do it with this same API?

aslushnikov: > And if in the future we got the ability to return the currently active file picker, would we do it with this same API?

I think technically there's no difference between getting the currently active file picker and the one that's opening - so I guess yes.

aslushnikov: > At least add

> NOTE This must be called BEFORE the file chooser is launched. It will not return a currently active file chooser.

Done.

lib/Page.js
aslushnikov: For the record: we agreed upon `waitForFileChooser` in offline discussion to be consistent with other APIs we have - such as `waitForRequest`, `waitForResponse` and `waitForNavigation`.

docs/api.md
aslushnikov: Incorporated this in the `FileCHooser` class note when we talk about handling all of the file choosers.

docs/api.md
```diff
@@ -2351,6 +2379,37 @@ Only one trace can be active at a time per browser.
 #### tracing.stop()
 - returns: <[Promise]<[Buffer]>> Promise which resolves to buffer with trace data.
 
+### class: FileChooser
+
+[FileChooser] objects are returned via the ['page.waitForFileChooser'](#pageinterceptfilechooseroptions) method.
```
vsemozhetbyt:
```suggestion
[FileChooser] objects are returned via the ['page.waitForFileChooser'](#pagewaitforfilechooseroptions) method.
```

JoelEinbinder (approved):

OmarKhattab:
Not sure why but for me when I am running headless filechooser works.
When I run headless set to false it gives error 
`TimeoutError: waiting for waiting for file chooser failed: timeout 30000ms exceeded`

aslushnikov:
@OmarKhattab please file a new issue with a good repro steps so that we can debug locally.

solartes:
@OmarKhattab Did you fix it?

OmarKhattab:
nah @solartes

solartes:
@OmarKhattab Try this 
    await page.waitForSelector('input[type=file]');    
    const input = await page.$('input[type=file]');
    await input.uploadFile(filePath);

PR #4654, opened 2019-07-01
chore: drop Node6 support

aslushnikov:
Node6 was end-of-life'ed in the end of April, 2019, with AWS Lambda
prohibiting updaets to the node6 runtime since June 30, 2019.

This makes it quite safe for us to remove the Node 6 support from the
repository.

From now on, all subsequent releases will require Node 8.9.0+.


JoelEinbinder:
I think this should be a major change

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "index.js",
   "repository": "github:GoogleChrome/puppeteer",
   "engines": {
-    "node": ">=6.4.0"
+    "node": ">=8.9.0"
```
JoelEinbinder, line 8: As a practical concern, is there anything from 7 that would break us?

utils/node6-transform/TransformAsyncFunctions.js
```diff
@@ -1,156 +0,0 @@
...
-    }
-    return step('next');
-  });
-};
-
-/**
- * @param {string} text
- * @return {string}
- */
-function transformAsyncFunctions(text) {
```
JoelEinbinder, line 54: RIP

package.json
aslushnikov: Not sure. I picked 8 here since it's an LTS release with a clear lifetime.

JoelEinbinder: Ok. Turns out this only produces a warning.

JoelEinbinder:
The test: Puppeteer-Web should work over web socket
looks like its failing.

README.md
```diff
@@ -57,7 +57,8 @@ See [puppeteer vs puppeteer-core](https://github.com/GoogleChrome/puppeteer/blob
 
 ### Usage
 
-Note: Puppeteer requires at least Node v6.4.0, but the examples below use async/await which is only supported in Node v7.6.0 or greater.
+Note: Prior to v1.18.1, Puppeteer required at least Node v6.4.0. All subsequent versions rely on
+Node 8.9.0+. All examples below use async/await which is only supported in Node v7.6.0 or greater.
```
mathiasbynens, line 61: Let's codify the supported Node.js versions over time as well. We could say something like:

```
Puppeteer follows the latest [maintenance LTS](https://github.com/nodejs/Release#release-schedule) version of Node.
```

mathiasbynens:
Note that since Node.js v8+ supports `async`/`await`, we can now drop [the `asyncawait` feature testing](https://github.com/GoogleChrome/puppeteer/search?q="new+Function%28%27async+function+test%28%29%7Bawait+1%7D%27%29%3B") present in various places. :+1:

install.js
```diff
@@ -109,22 +107,6 @@ function toMegabytes(bytes) {
...
-    return;
-  // if async/await is supported, then node6 is not needed.
-  if (supportsAsyncAwait())
-    return;
-  // Re-build node6/ folder.
-  logPolitely('Building Puppeteer for Node 6');
-  require(path.join(__dirname, 'utils', 'node6-transform'));
-}
-
 function supportsAsyncAwait() {
```
mathiasbynens: This can be removed as well.

mathiasbynens: Amended your patch accordingly :+1:

mathiasbynens:
Moved to https://github.com/GoogleChrome/puppeteer/pull/5045.

PR #4698, opened 2019-07-12
fix(tests): jshandle test fix

yury-s:
Check message prefix rather than strict equality when detecting circular JSON error. The message format has changed recently which broke the condition.


aslushnikov:
> The message format has changed recently which broke the condition.

In Node 12 I guess. I'll add Node 10 and Node 12 bots to catch this kind of regressions

aslushnikov (approved):

PR #4699, opened 2019-07-13
chore(ci): add Node10 and Node12 to Cirrus

aslushnikov:


utils/testserver/index.js
```diff
@@ -46,6 +46,7 @@ class TestServer {
     const server = new TestServer(dirPath, port, {
       key: fs.readFileSync(path.join(__dirname, 'key.pem')),
       cert: fs.readFileSync(path.join(__dirname, 'cert.pem')),
+      secureProtocol: 'TLSv1_2_method',
```
JoelEinbinder, line 49: what's this?

JoelEinbinder (approved):

utils/testserver/index.js
aslushnikov: Node 12 uses TLS 1.3 by default; we have a few tests that check for the TLS version used. 

This forces TLS to always be 1.2

PR #4700, opened 2019-07-13
fix(tests): Get TLS version from Node instead of hardcoding it

yury-s:
This will make the tests resilient to open ssl updates in node.

### DIFF
--- a/test/launcher.spec.js
+++ b/test/launcher.spec.js
@@ -287,12 +287,14 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions, p
 
         const browser = await puppeteer.connect({browserWSEndpoint, ignoreHTTPSErrors: true});
         const page = await browser.newPage();
+        const requestEvent = httpsServer.waitForRequest('/empty.html');
         let error = null;
         const response = await page.goto(httpsServer.EMPTY_PAGE).catch(e => error = e);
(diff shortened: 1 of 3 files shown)


test/launcher.spec.js
```diff
@@ -287,12 +287,14 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions, p
 
         const browser = await puppeteer.connect({browserWSEndpoint, ignoreHTTPSErrors: true});
         const page = await browser.newPage();
+        const requestEvent = httpsServer.waitForRequest('/empty.html');
         let error = null;
         const response = await page.goto(httpsServer.EMPTY_PAGE).catch(e => error = e);
```
aslushnikov:
```js
const [serverRequest, response] = await Promise.all([
  httpsServer.waitForRequest('/empty.html'),
  page.goto(httpsServer.EMPTY_PAGE).catch(e => error = e),
]);
```

yury-s: Done, thanks!

yury-s:
> Once you've signed (or fixed any issues), please reply here (e.g. `I signed it!`) and we'll verify it.

Looks like my previous signature has expire or something. I've just signed the CLA again, let's see if that works.

aslushnikov (approved):
Thanks!

PR #4704, opened 2019-07-14
fix(launcher): disable component extensions

### LINKED ISSUE #4300: `backgroundpage.on('console')` not working
### Steps to reproduce
* Puppeteer version: 1.14.0
* Platform / OS version: macOS 10.14.3
* Node.js version: 11.6.0

1. `git clone https://github.com/pyrho/puppeteer-bug.git`
2. `yarn`
3. `node example.js`

The repo for this issue: https://github.com/pyrho/puppeteer-bug 

Messages (`console.log`)from the background page are caught by the `backgroundPage.on('console')` handler. 
`bg: PORT CONNECTED" should print in the node console.

**What happens instead?**
Messages are not caught by the handler.

When manually inspecting the background page, we can see that the message is printed in the console tab:
Screenshot 2019-04-17 at 16 09 00

Thank you for your time and this awesome project :)

aslushnikov:
Chrome has a set of component extensions - e.g. CryptoTokenExtension
that helps with 2FA.

These extensions are loaded regardless of the `--disable-extensions`
flag we already pass. To disable these extensions, we need to pass additional
`--disable-component-extensions-with-background-pages` flag.

Fix #4300

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -42,6 +42,7 @@ const DEFAULT_ARGS = [
   '--disable-backgrounding-occluded-windows',
   '--disable-breakpad',
   '--disable-client-side-phishing-detection',
+  '--disable-component-extensions-with-background-pages',
   '--disable-default-apps',
   '--disable-dev-shm-usage',
   '--disable-extensions',
--- a/test/assets/simple-extension/index.js
+++ b/test/assets/simple-extension/index.js
@@ -1,1 +1,2 @@
 // Mock script for background extension
+window.MAGIC = 42;
--- a/test/headful.spec.js
+++ b/test/headful.spec.js
@@ -73,6 +73,7 @@ module.exports.addTests = function({testRunner, expect, puppeteer, defaultBrowse
       const backgroundPageTarget = await waitForBackgroundPageTarget(browserWithExtension);
       const page = await backgroundPageTarget.page();
       expect(await page.evaluate(() => 2 * 3)).toBe(6);
+      expect(await page.evaluate(() => window.MAGIC)).toBe(42);
       await browserWithExtension.close();
     });
     it('should have default url when launching browser', async function() {


aslushnikov:
I signed it!

JoelEinbinder:
Why?

aslushnikov:
@JoelEinbinder context: https://github.com/GoogleChrome/puppeteer/pull/3694#issuecomment-511242140

Long story short: these extensions add background pages, so it becomes impossible to wait for *the* background page of my extension.

JoelEinbinder:
Presumably these extensions do something. Disabling them because it makes our API easier to use sounds to me like you are "fixing" this problem at the wrong layer.

aslushnikov:
> Presumably these extensions do something.

@JoelEinbinder They don't do anything important from the WebPlatform point of view. They do provide some enhanced UserAgent experience, but we already compromise it in many ways in favor of a pleasant automation, e.g. disabling keychains, phishing protection, translation e.t.c.

JoelEinbinder (approved):

JoelEinbinder:
Let's add a test

aslushnikov:
> Let's add a test

@JoelEinbinder I don't know any reliable way to check that this extension doesn't load; any suggestion?

JoelEinbinder:
Have a test with our own extension. Verify that there are no other
extensions once that one loads?

aslushnikov:
> Have a test with our own extension. Verify that there are no other
extensions once that one loads?

This won't work reliably - the order of background page loading is not guaranteed. There's no event I'm aware of that we can subscribe to that happens when all extension background pages are loaded.

AmySteam:
I've been using `Target#waitForTarget` to wait for my extension to load, and adding `--disable-component-extensions-with-background-pages` gets my extension without having to use `Target#_targetInfo.title`.

```javascript
const browser = await pptr.launch({
  headless: false,
  args: [
    '--disable-component-extensions-with-background-pages',
    `--disable-extensions-except=${pathToExtension}`,
    `--load-extension=${pathToExtension}`,
  ],
})

const backgroundTarget = await browser.waitForTarget(
  (t) => t.type() === 'background_page',
)

const backgroundPage = await backgroundTarget.page()
```

JoelEinbinder:
> > Have a test with our own extension. Verify that there are no other
> > extensions once that one loads?
> 
> This won't work reliably - the order of background page loading is not guaranteed. There's no event I'm aware of that we can subscribe to that happens when all extension background pages are loaded.

I'm fine with a test that succeeds reliably but might give false positives if it starts failing.

We are implicitly telling users to rely on the fact that no extensions will be in the browser besides the ones they added. Having a test codifies this as the intended behavior.

JoelEinbinder (approved):

Issue #4283, opened 2019-04-13
Support Electron and other content embedders


aslushnikov:
## What's a "content embedder"?

Chromium codebase is organized into layers:
- `//blink` - HTML rendering engine
- `//content` - an API for browser implementers (think of it as a library that you'd use if you were to write your own browser). Things like process model are handled here; uses `//blink`.
- `//chrome` - Chromium implementation (uses `//content`).

Content-embedders are all the products that are based on `//content` layer. Chromium, Chrome Headless, ChromeCast are all different `//content` embedders.

Electron is a `//content` embedder as well.

## DevTools Protocol and content embedders

Majority of DevTools API is implemented in `//blink` and `//content`. However, certain methods are supposed to be implemented by embedders. DevTools team makes sure all the methods needed for Puppeteer operation are supported by both Chromium and Headless.

Electron is, however, missing out.

## What exactly goes wrong?

- Since version v1.5.0, when Puppeteer connects to a browser it requests all existing browser
contexts with the `Target.getBrowserContexts()` protocol method.
- Method `Target.getBrowserContexts()` is supposed to be implemented by content embedders.
- Electron doesn't implement the method, so Puppeteer connect to electron fails (#3793).

## What can be done?

**Option 1: Defer to Embedders.**

Ideally, `//content`-embedders implement all necessary DevTools protocol methods on their end. This way they'll guarantee 100% compatibility with Puppeteer API. This, however, is a major undertaking for embedders; not sure if there's any interest in this.

**Option 2: support clicking/typing/evaluation**

Alternatively, we can aim for a "good-enough" compatibility with `//content` embedders. For example, we can make sure that things like clicking, navigation, typing and evaluation work fine with electron app (and add these electron tests to our CI). These should be enough to drive many default testing and automation scenarios.

# 🚀  Vote! 

- If there's an interest in **Option 1**, please file a bug to the repository with the content embedder you'd like to be supported and ask them to implement the `Target.getBrowserContexts()` method. Plz cross-post the link here.
- If there's an interest in **Option 2**, please  👍 this issue and share your usecase.

aslushnikov:
**Option 2** seems to be good-enough for me to address the e2e tests of electron apps.

JoelEinbinder:
Do we have a list of all the protocol methods that are embedder-implemented which puppeteer uses?

aslushnikov:
[...]

PR #4720, opened 2019-07-17
Fix connection issue to VSCode instance

caslt:
fix(Launcher): fix Launcher.connect method

This patch fixes Launcher.connect  so that it works with existing Visual Studio code instance.

The is an issue when connecting to the existing Visual Studio Code instance.

Visual Studio Code seems not supporting the protocol command _Target.getBrowserContexts_

So when trying to connect to the Visual Studio Code instance by these steps, it will failed

1. Open a Visual Studio Code instance, code . --remote-debugging-port=9222
2. Get the WSEndPoint in http://localhost:9222/json/version
3. Connect to this WSEndPoint by puppeteer.connect({browserWSEndpoint:'...'})

The issue raises from 
const {browserContextIds} = await connection.send('Target.getBrowserContexts');
in Launcher.js

I take a look the value when VSCode is lanuched with puppeteer.Lanuch, and it is null. So null is set for this parameter, and the Browser.js need to updated to check if contextIds is not null when trying to loop the contextIds.

With this fix, I could connect to an existing Visual Studio Code instance.

### DIFF
--- a/lib/Browser.js
+++ b/lib/Browser.js
@@ -55,8 +55,11 @@ class Browser extends EventEmitter {
     this._defaultContext = new BrowserContext(this._connection, this, null);
     /** @type {Map<string, BrowserContext>} */
     this._contexts = new Map();
-    for (const contextId of contextIds)
+    if (contextIds!=null) {
+      for (const contextId of contextIds)
       this._contexts.set(contextId, new BrowserContext(this._connection, this, contextId));
+    }
+    
 
     /** @type {Map<string, Target>} */
     this._targets = new Map();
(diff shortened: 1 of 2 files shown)


aslushnikov:
@caslt we have #4283 to support various electron applications.

We explicitly avoid this kind of workarounds. I believe you can just use an older pptr version, e.g. v1.4.0, to drive VSCode instance. 

Thank you for the PR though! Let's continue discussion in the issue.

caslt:
@aslushnikov 
So actually I should ask the Visual Studio Code team to support the protocol command Target.getBrowserContexts, right?

aslushnikov:
Rather electron team; I believe they have a bunch of issues already

caslt:
OK. Thanks for the information.

caslt:
I find the related issue in electron https://github.com/electron/electron/issues/17776

PR #4721, opened 2019-07-17
feat(testrunner): async suite descriptions

aslushnikov:
This patch teaches TestRunner to support async suite
descriptions. This is needed to require tests using ES6 dynamic
imports:

```js
const t = new TestRunner();

await t.describe('tests', async () => {
  (await import('./some.spec.js')).addTests(t);
  (await import('./other.spec.js')).addTests(t);
});
```

### DIFF
--- a/utils/testrunner/TestRunner.js
+++ b/utils/testrunner/TestRunner.js
@@ -319,12 +319,14 @@ class TestRunner extends EventEmitter {
     this._hasFocusedTestsOrSuites = this._hasFocusedTestsOrSuites || mode === TestMode.Focus;
   }
 
-  _addSuite(mode, comment, name, callback) {
+  async _addSuite(mode, comment, name, callback) {
     const oldSuite = this._currentSuite;
     const suite = new Suite(this._currentSuite, name, mode, comment);
     this._currentSuite.children.push(suite);
     this._currentSuite = suite;
-    callback();
+    const result = callback();
+    if (result && (typeof result.then === 'function'))
+      await result;
     this._currentSuite = oldSuite;
     this._hasFocusedTestsOrSuites = this._hasFocusedTestsOrSuites || mode === TestMode.Focus;
   }


utils/testrunner/TestRunner.js
```diff
@@ -319,12 +319,14 @@ class TestRunner extends EventEmitter {
...
 
-  _addSuite(mode, comment, name, callback) {
+  async _addSuite(mode, comment, name, callback) {
     const oldSuite = this._currentSuite;
     const suite = new Suite(this._currentSuite, name, mode, comment);
     this._currentSuite.children.push(suite);
     this._currentSuite = suite;
-    callback();
+    const result = callback();
+    if (result && (typeof result.then === 'function'))
```
JoelEinbinder, line 328: This feels hacky. I get why you do it, but it means that your entire test suite needs to properly to await everything if its async, leaving you open to flakiness bugs in your test runner itself!

I expected to see a while loop waiting for all test suites here, until there are no more test suites to wait for. This will be way safer for users.

aslushnikov: > I expected to see a while loop waiting for all test suites here, until there are no more test suites to wait for. This will be way safer for users.

Since testrunner DSL (it, xit, fit, beforeEach etc) is stack-based, i don't think this suggestion works.

>  it means that your entire test suite needs to properly to await everything if its async, leaving you open to flakiness bugs in your test runner itself!

That's a fair concern.

The way to go is to restrict parallel declaration of 2 async test suites, so that the following throws:

```js
testRunner.describe('uno', async() => {});
// The following line should throw since previous async testsuite is pending.
testRunner.describe('dos', async() => {});
```

Also, various testRunner APIs should throw if there's a pending async testsuite declaration:

```js
testRunner.describe('uno', async() => {});
// The following should throw since previous testsuite is async and is pending.
testRunner.tests();
testRunner.failedTests();
testRunner.passedTests();
testRunner.skippedTests();
testRunner.hasFocusedTestsOrSuites();
```

I'm all-in for building this safety system in a follow-up, together with a tests for the testrunner.

JoelEinbinder (approved):

PR #4723, opened 2019-07-17
feat(testrunner): async suite descriptions

JoelEinbinder:
Alternative to #4721

Does this work?

### DIFF
--- a/utils/testrunner/TestRunner.js
+++ b/utils/testrunner/TestRunner.js
@@ -306,6 +306,8 @@ class TestRunner extends EventEmitter {
     this.beforeEach = this._addHook.bind(this, 'beforeEach');
     this.afterAll = this._addHook.bind(this, 'afterAll');
     this.afterEach = this._addHook.bind(this, 'afterEach');
+
+    this._currentChildSuiteDescriptors = [];
   }
 
   addTestDSL(dslName, mode, comment) {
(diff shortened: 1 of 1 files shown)


aslushnikov:
> Does this work?

@JoelEinbinder This mangles test order, basically top-sorting them. Consider running this snippet: https://gist.github.com/aslushnikov/93a8fa25655063187c64767969c33ae8

You'd expect tests to run sequentially, but in fact they run as 1, 4, 2, 3.

The suggestion is very fascinating though, thank you for bringing it up! 👍

Comparing to #4721:
1. I **`don't like`** it because this is not how other test runners work. It makes test description asynchronous in a unique way - you don't know in which context and which point of time they are executed. 
2. I **`don't like`** it because it's a conceptual change driven by a rare use case that affects all users.
3. I **`don't like`** it because debugging experience will suffer (can't step over spec definitions, stack traces are not nice)
4. I **`like`** it because it drops the need for top-level await for descriptions
5. I **`like`** it because it makes it harder to forget to await test suite, like in #4721 

So far I still lean towards #4721 primarily due to the ratio of simplicity / effort / importance, but I'm open to be convinced otherwise.

P.S. There's an interesting discussion on async testsuites in jest https://github.com/facebook/jest/issues/2235

aslushnikov:
> So far I still lean towards #4721 primarily due to the ratio of simplicity / effort / importance, but I'm open to be convinced otherwise.

...convinced otherwise, especially if the importance of this proves to be significant

JoelEinbinder:
> > Does this work?
> 
> @JoelEinbinder This mangles test order, basically top-sorting them. Consider running this snippet: https://gist.github.com/aslushnikov/93a8fa25655063187c64767969c33ae8
> 
> You'd expect tests to run sequentially, but in fact they run as 1, 4, 2, 3.

My bad, I didn't realize that addSuite and addTest were separate.

JoelEinbinder:
Closing, this was just a discussion prop.

PR #4727, opened 2019-07-18
fix(launcher): make dumpio and pipe options work together

yury-s:
Don't ignore stdout and stderr when using pipe for remote debugging and dumpio is true. In that case puppeteer process connects to the stdout/stderr streams of the child process and it will not hang.

### DIFF
--- a/lib/Launcher.js
+++ b/lib/Launcher.js
@@ -123,7 +123,14 @@ class Launcher {
 
     const usePipe = chromeArguments.includes('--remote-debugging-pipe');
     /** @type {!Array<"ignore"|"pipe">} */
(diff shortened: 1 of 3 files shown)


aslushnikov:
Let's add a test! E.g. similar to https://github.com/GoogleChrome/puppeteer/blob/3982a603cc0db32bf05cd0960d24679a5c39313b/test/fixtures.spec.js#L25-L41

yury-s:
> Let's add a test!

@aslushnikov Done. I was originally inspired by testing approach from 14e69d18 :)

aslushnikov (approved):
Thanks!

PR #4769, opened 2019-07-29
fix(tests): background page tests should wait for the page

JoelEinbinder:
It tests `waitForBackgroundPageTarget(browser)` would fail if it ever had to wait for the target. Luckily today we have `browser.waitForTarget` which makes life easier.

### DIFF
--- a/test/headful.spec.js
+++ b/test/headful.spec.js
@@ -25,20 +25,6 @@ const mkdtempAsync = helper.promisify(fs.mkdtemp);
 
 const TMP_FOLDER = path.join(os.tmpdir(), 'pptr_tmp_folder-');
 
-function waitForBackgroundPageTarget(browser) {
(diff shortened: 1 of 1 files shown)


test/headful.spec.js
```diff
@@ -25,20 +25,6 @@ const mkdtempAsync = helper.promisify(fs.mkdtemp);
...
 
-function waitForBackgroundPageTarget(browser) {
-  const target = browser.targets().find(target => target.type() === 'background_page');
-  if (target)
-    return Promise.resolve(target);
-  return new Promise(resolve => {
-    browser.on('targetcreated', function listener(target) {
-      if (target.type() !== 'background_page')
-        return;
-      browser.removeListener(listener);
```
JoelEinbinder, line 36: This should have been `browser.removeListener('targetcreated', listener)`

aslushnikov (approved):

PR #4772, opened 2019-07-29
fix(test): make sure selection is not empty before running copy command

yury-s:
If current selection is empty document.execCommand('copy') may return false in some browsers (e.g. WebKit based ones).

### DIFF
--- a/test/evaluation.spec.js
+++ b/test/evaluation.spec.js
@@ -226,9 +226,11 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(error.message).toContain('JSHandles can be evaluated only in the context they were created');
     });
     it('should simulate a user gesture', async({page, server}) => {
-      await page.evaluate(() => document.body.appendChild(document.createTextNode('test')));
-      await page.evaluate(() => document.execCommand('selectAll'));
-      const result = await page.evaluate(() => document.execCommand('copy'));
+      const result = await page.evaluate(() => {
+        document.body.appendChild(document.createTextNode('test'));
+        document.execCommand('selectAll');
+        return document.execCommand('copy');
+      });
       expect(result).toBe(true);
     });
     it('should throw a nice error after a navigation', async({page, server}) => {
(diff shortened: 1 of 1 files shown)


test/evaluation.spec.js
```diff
@@ -226,8 +226,11 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(error.message).toContain('JSHandles can be evaluated only in the context they were created');
     });
     it('should simulate a user gesture', async({page, server}) => {
-      const result = await page.evaluate(() => document.execCommand('copy'));
-      expect(result).toBe(true);
+      await page.evaluate(() => document.body.appendChild(document.createTextNode('test')));
+      const selectResult = await page.evaluate(() => document.execCommand('selectAll'));
+      expect(selectResult).toBe(true);
```
JoelEinbinder: Firefox is unhappy about this. Yay! Browsers!

Let's just do this with window.getSelection to not mess around more with execCommand than we have to, or just don't assert that the selectAll worked.

yury-s: Updated expectations to check only result of 'copy'.

MedRedha (approved):

test/evaluation.spec.js
```diff
@@ -226,6 +226,8 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(error.message).toContain('JSHandles can be evaluated only in the context they were created');
     });
     it('should simulate a user gesture', async({page, server}) => {
+      await page.evaluate(() => document.body.appendChild(document.createTextNode('test')));
```
JoelEinbinder: Nit: you can do these all in one evaluate, save some hops.

JoelEinbinder (approved):

test/evaluation.spec.js
yury-s: Good point, done. Await and arrow functions make the code so much more readable that it's easy to forget that there are roundtrips over the protocol on each evaluate invocation :-)

PR #4778, opened 2019-07-31
test: add page.bringToFront test

kblok:
I think we can test `page.bringToFront` in headful.

### DIFF
--- a/test/headful.spec.js
+++ b/test/headful.spec.js
@@ -130,5 +130,26 @@ module.exports.addTests = function({testRunner, expect, puppeteer, defaultBrowse
       await browser.close();
     });
   });
+
+  describe('Page.bringToFront', function() {
+    it('should work', async() => {
+      const browser = await puppeteer.launch(headfulOptions);
+      const page = await browser.newPage();
+
+      expect(await page.evaluate(() => document.visibilityState)).toBe('visible');
+
+      const newPage = await browser.newPage();
+      expect(await page.evaluate(() => document.visibilityState)).toBe('hidden');
+      expect(await newPage.evaluate(() => document.visibilityState)).toBe('visible');
+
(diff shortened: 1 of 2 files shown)


test/headful.spec.js
```diff
@@ -130,5 +130,26 @@ module.exports.addTests = function({testRunner, expect, puppeteer, defaultBrowse
       await browser.close();
     });
   });
+
+  describe('Page.bringToFront', function() {
```
JoelEinbinder, line 134: This is cool, and looks like it works. I am concerned that it is more testing for the presence of a tab strip than for bring-to-front to be actually working properly. Browsers don't need to have a tab strip that behaves this way, with only one web contents visible at a time, nor do they need to make the page created by browser.newPage() visible. But maybe that's fine, and we can cross that bridge when we get to it. @aslushnikov wdyt?

aslushnikov: > Browsers don't need to have a tab strip that behaves this way

Yeah, but I'd speculate that browsers are unlikely to change here in the foreseeable future.

> nor do they need to make the page created by browser.newPage() visible

I don't really think we've relied on this anywhere so far. Let's not use this assumption in the test:

```js

await page1.bringToFront();
expect(await page1.evaluate(() => document.visibilityState)).toBe('visible');
expect(await page2.evaluate(() => document.visibilityState)).toBe('hidden');
await page2.bringToFront();
expect(await page1.evaluate(() => document.visibilityState)).toBe('hidden');
expect(await page2.evaluate(() => document.visibilityState)).toBe('visible');
```

> But maybe that's fine, and we can cross that bridge when we get to it

Yeah I like it! Let's land this!

kblok: Done!

aslushnikov (approved):

PR #4781, opened 2019-07-31
chore: introduce initial version of flakiness dashboard

aslushnikov:
This patch introduces a dashboard that records test results and
uploads them to https://github.com/aslushnikov/puppeteer-flakiness-dashboard

Since many bots might push results in parallel, each bot pushes
results to its own git branch.

FlakinessDashboard also generates a simple README.md with a flakiness
summary. If this proves to be not enough, we can build a website that
fetches flakiness data and renders it nicely.

### DIFF
--- a/test/test.js
+++ b/test/test.js
@@ -111,4 +111,50 @@ new Reporter(testRunner, {
   projectFolder: utils.projectRoot(),
   showSlowTests: process.env.CI ? 5 : 0,
 });
-testRunner.run();
+
+const flakinessDashboard = initializeFlakinessDashboardIfNeeded(testRunner);
+testRunner.run().then(() => {
+  if (flakinessDashboard)
+    flakinessDashboard.uploadAndCleanup();
+});
+
+
+function initializeFlakinessDashboardIfNeeded(testRunner) {
+  // whitelist Cirrus for now.
+  if (!process.env.CIRRUS_CI)
+    return null;
+  // FLAKINESS_DASHBOARD_PASSWORD is encrypted. Cirrus DOES NOT inject enctrypted
+  // variables if PR's are sent from users without write permissions to the repo.
+  //
+  // This makes sure we are running on Cirrus CI master branch, not a PR.
+  if (!process.env.FLAKINESS_DASHBOARD_PASSWORD || process.env.CIRRUS_BASE_SHA)
+    return null;
+  const {FlakinessDashboard} = require('../utils/flakiness-dashboard');
+  const sha = process.env.CIRRUS_CHANGE_IN_REPO;
+  const dashboard = new FlakinessDashboard({
+    dashboardName: process.env.CIRRUS_TASK_NAME,
+    build: {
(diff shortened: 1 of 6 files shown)


test/test.js
```diff
@@ -105,10 +105,52 @@ if (process.env.CI && testRunner.hasFocusedTestsOrSuites()) {
...
+  const {FlakinessDashboard} = require('../utils/flakiness-dashboard');
+  const sha = process.env.CIRRUS_CHANGE_IN_REPO;
+  const dashboard = await FlakinessDashboard.initialize({
+    dashboardName: process.env.CIRRUS_TASK_NAME,
+    build: {
+      url: `https://cirrus-ci.com/build/${process.env.CIRRUS_BUILD_ID}`,
+      name: sha.substring(0, 8),
+    },
+    dashboardRepo: {
+      url: 'https://github.com/aslushnikov/puppeteer-flakiness-dashboard.git',
```
JoelEinbinder: puppeteer org?

JoelEinbinder:
> Since many bots might push results in parallel, each bot pushes
> results to its own git branch.

Who guarantees that even a single bot won't run two instances in parallel?

aslushnikov:
> Who guarantees that even a single bot won't run two instances in parallel?

Added retries with cooldown in case there are several instances of the same bot working on different revisions in parallel. This, however, should not happen in our setup since CIs limit free accounts to one bot instance.

MedRedha (approved):

test/test.js
```diff
@@ -111,4 +111,67 @@ new Reporter(testRunner, {
   projectFolder: utils.projectRoot(),
   showSlowTests: process.env.CI ? 5 : 0,
 });
+
+// Unconditionally generate testIDs for all tests and verify they don't clash.
+// This will add |test.testId| for every test.
+generateTestIDs(testRunner);
```
JoelEinbinder: nit: can we put this into its own file. I want to be able to easily remove the flakiness dashboard code if we decide its not worth it.

utils/flakiness-dashboard/FlakinessDashboard.js
```diff
@@ -0,0 +1,276 @@
...
+
+const RED_COLOR = '\x1b[31m';
+const GREEN_COLOR = '\x1b[32m';
+const YELLOW_COLOR = '\x1b[33m';
+const RESET_COLOR = '\x1b[0m';
+
+class FlakinessDashboard {
+  /**
+   * @param {{url: string, branch: string, username: string, password: string}} dashboardRepo
+   * @param {{url: string, name: string}} checkout
```
JoelEinbinder: what are all these other params here?

JoelEinbinder (approved):

utils/flakiness-dashboard/FlakinessDashboard.js
aslushnikov: Indeed

test/test.js
aslushnikov: Moved to `//test/utils.js` and consolidated in a single function call.

PR #4786, opened 2019-08-01
fix(page): fix page.screenshot method

himaozi:
This patch fixes page.screenshot. so that it throws correct error message when options.clip.height===0


aslushnikov (approved):
Thanks!

PR #4797, opened 2019-08-02
test(network): be able to remove headers using request.continue

### LINKED ISSUE #4743: Documentation example for removing header doesn't remove header
The API.md page has the following example:
```
  // Override headers
  const headers = Object.assign({}, request.headers(), {
    foo: 'bar', // set "foo" header
    origin: undefined, // remove "origin" header
  });
  request.continue({headers});
});
```
When the example is executed as follows:
```
req.continue({headers:{'pragma': undefined}});
```

The actual headers look like:
```
Connection: keep-alive
Pragma: undefined
Cache-Control: no-cache
Referer: http://www.google.com/
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/77.0.3835.0 Safari/537.36
```

Notice that `Pragma` is still there, merely its value replaced.

kblok:
closes #4743 

I think this was a regression caused here https://github.com/GoogleChrome/puppeteer/pull/4265/files#diff-d2ac7cb061b0c51644d0a5d6140e3a32R446

### DIFF
--- a/experimental/puppeteer-firefox/lib/NetworkManager.js
+++ b/experimental/puppeteer-firefox/lib/NetworkManager.js
@@ -158,7 +158,7 @@ class Request {
     } = overrides;
     await this._session.send('Network.resumeSuspendedRequest', {
       requestId: this._id,
-      headers: headers ? Object.entries(headers).map(([name, value]) => ({name, value})) : undefined,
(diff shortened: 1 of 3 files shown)


aslushnikov (approved):
awesome!

MedRedha (commented):
Love it! 🎉

PR #4808, opened 2019-08-06
chore(flakiness): update flakiness format

aslushnikov:
This patch:
- updates Flakiness Dashboard format to define version per-build
  and to pass COMMIT information
- drops the README.md generation - we'll move on to a designated flakiness
  dashboard viewer


aslushnikov:
NOTE: Prior to landing this commit, I'll migrate all the current flakiness data to the format.

Edit: data updated in flakiness repo.

test/test.js
```diff
@@ -111,6 +111,8 @@ new Reporter(testRunner, {
   showSlowTests: process.env.CI ? 5 : 0,
 });
 
-utils.initializeFlakinessDashboardIfNeeded(testRunner);
-testRunner.run();
+(async() => {
```
JoelEinbinder, line 114: nit: can you move this to the top of this file, like line 20. Its easier to reason about this file if the whole thing is async, rather than just one step at the end.

test/utils.js
```diff
@@ -181,18 +181,22 @@ const utils = module.exports = {
...
     if (!process.env.FLAKINESS_DASHBOARD_PASSWORD || process.env.CIRRUS_BASE_SHA)
       return;
-    const sha = process.env.FLAKINESS_DASHBOARD_BUILD_SHA;
+    const {sha, timestamp} = await FlakinessDashboard.getCommitDetails(__dirname, 'HEAD');
     const dashboard = new FlakinessDashboard({
-      dashboardName: process.env.FLAKINESS_DASHBOARD_NAME,
+      commit: {
+        sha,
+        timestamp,
+        url: `https://github.com/GoogleChrome/puppeteer/commit/${sha}`,
```
JoelEinbinder, line 189: is baking in the repo URL necessary here? The commit hash should be enough.

JoelEinbinder (approved):

test/utils.js
aslushnikov: Don't want to rely on Github URL schemes too much in flakiness viewer. This comes for free since this is per-build only (and there are not many builds)

test/test.js
aslushnikov: I tried that initially, but didn't like extra indentation. Does this bother you strongly?

test/utils.js
JoelEinbinder: What happens if we ever move the repo?

JoelEinbinder:
I think it bothers me more than the indentation bothers you, but I don’t
feel super strongly about it.

aslushnikov:
I'll land this now and will discuss the reason for your disturbance offline

PR #4816, opened 2019-08-07
fix(page): Resolve file chooser using a single promise

kblok:
Can't we resolve all the `waitForFileChooser` calls using a single promise instead of creating a `Set` of promises that will be resolved at the same time with the same output?

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -112,7 +112,8 @@ class Page extends EventEmitter {
     networkManager.on(Events.NetworkManager.RequestFailed, event => this.emit(Events.Page.RequestFailed, event));
     networkManager.on(Events.NetworkManager.RequestFinished, event => this.emit(Events.Page.RequestFinished, event));
     this._fileChooserInterceptionIsDisabled = false;
-    this._fileChooserInterceptors = new Set();
+    /** @type {?Promise<FileChooser>} */
+    this._fileChooserPromise = null;
 
     client.on('Page.domContentEventFired', event => this.emit(Events.Page.DOMContentLoaded));
     client.on('Page.loadEventFired', event => this.emit(Events.Page.Load));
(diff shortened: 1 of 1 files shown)


lib/Page.js
```diff
@@ -167,13 +166,11 @@ class Page extends EventEmitter {
     const {
       timeout = this._timeoutSettings.timeout(),
     } = options;
-    let callback;
-    const promise = new Promise(x => callback = x);
-    this._fileChooserInterceptors.add(callback);
-    return helper.waitWithTimeout(promise, 'waiting for file chooser', timeout).catch(e => {
-      this._fileChooserInterceptors.delete(callback);
```
aslushnikov, line 174: At first I looked at this patch and was "Ah, totally! Reusing a single promise is *the* way to go here!"

But then I noticed [line #174](https://github.com/GoogleChrome/puppeteer/pull/4816/files#diff-b1397100d3965ca1373667a75e2c5f94L174) that cleans up fileChooserInterceptors which is important and that didn't go to a new patch.

Consider the following testcase:
```js
it('should kill timedout watchdogs', async({page}) => {
  await page.setContent(`<input multiple type=file>`);
  await page.waitForFileChooser({timeout: 1}).catch(e => {});
  // This should show the file chooser dialog to the user.
  await page.click();
  // We don't have a way to do this check unfortunately 🤷‍♂️
  expect(page).toShowFileChooserDialogToTheUser(); 
});
```

Currently, the filechooser watchdog will die and all the subsequent file chooser requests will be ["fallbacked"](https://github.com/GoogleChrome/puppeteer/pull/4816/files#diff-b1397100d3965ca1373667a75e2c5f94L150). In a new patch, we don't clean up the cached promise here. As a result, subsequent file chooser events coming from DevTools protocol won't "fallback".

To restore similar behavior, you would need to count the number of active file chooser watchdogs - and clean up the cached promise once there are none any more. It looks to me that this will boil down to something very similar to the `this._fileChooserInterceptors` dance we do here.

Too bad we don't have a test to cover this :'(

kblok: @aslushnikov we are not cleaning up the promise here, but we are doing that on [line 156]( https://github.com/GoogleChrome/puppeteer/pull/4816/files#diff-b1397100d3965ca1373667a75e2c5f94R156).

`_onFileChooser` would resolve the promise, if any, and set that back to null. After that, we would fallback again.

Am I missing something?

kblok: I got your point. you are right :(

PR #4818, opened 2019-08-07
chore(flakiness): limit max builds to 100

aslushnikov:
This was missing in #4808


JoelEinbinder (commented):
Why? I’m missing context.

aslushnikov:
That's a good practice to limit storage. 100 is the number of builds used in Chromium Flakiness Dashboard.

JoelEinbinder (approved):

PR #4836, opened 2019-08-13
feat(frame): introduce frame.ownerElement

### LINKED ISSUE #4709: [Feature Request] Get ElementHandle for corresponding Frame
We need a nicer way to get `ElementHandle<HTMLIframeElement>` given an `Frame`.
Currently, there is no API for this.
Related: #2889

```js
// We can get Frame from ElementHandle:

const iframe = await page.$('iframe'); // iframe is ElementHandle
const frame = await iframe.contentFrame(); // frame is Frame

// ...
// How do we get ElementHandle out of Frame now?
// We need something like iframeElement method:

const iframe = await frame.iframeElement();
```

kblok:
closes #4709 (I found this issue with a few ❤️)

The `asElement` name is used so it's consistent with [JSHandle.asElement](https://github.com/GoogleChrome/puppeteer/blob/v1.19.0/docs/api.md#jshandleaselement). 
[DOM.getFrameOwner](https://chromedevtools.github.io/devtools-protocol/tot/DOM#method-getFrameOwner) is marked as experimental. I don't know whether we should use it or not.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -206,6 +206,7 @@
   * [frame.$x(expression)](#framexexpression)
   * [frame.addScriptTag(options)](#frameaddscripttagoptions)
   * [frame.addStyleTag(options)](#frameaddstyletagoptions)
+  * [frame.asElement()](#frameaselement)
   * [frame.childFrames()](#framechildframes)
   * [frame.click(selector[, options])](#frameclickselector-options)
   * [frame.content()](#framecontent)
@@ -229,7 +230,6 @@
   * [frame.waitForNavigation([options])](#framewaitfornavigationoptions)
   * [frame.waitForSelector(selector[, options])](#framewaitforselectorselector-options)
   * [frame.waitForXPath(xpath[, options])](#framewaitforxpathxpath-options)
-  * [frame.asElement()](#frameaselement)
 - [class: ExecutionContext](#class-executioncontext)
   * [executionContext.evaluate(pageFunction[, ...args])](#executioncontextevaluatepagefunction-args)
   * [executionContext.evaluateHandle(pageFunction[, ...args])](#executioncontextevaluatehandlepagefunction-args)
@@ -2572,6 +2572,9 @@ Adds a `<script>` tag into the page with the desired url or content.
 
 Adds a `<link rel="stylesheet">` tag into the page with the desired url or a `<style type="text/css">` tag with the content.
 
+#### frame.asElement()
+- returns: <[Promise]<?[ElementHandle]>> Resolves to the element handle representing frame node. The method will return null for the main frame.
+
 #### frame.childFrames()
 - returns: <[Array]<[Frame]>>
 
@@ -2912,9 +2915,6 @@ puppeteer.launch().then(async browser => {
   await browser.close();
 });
 ```
-
-#### frame.asElement()
-- returns: <[Promise]<?[ElementHandle]>> Resolves to the element handle representing frame node. The method will return null for the main frame.
 
 ### class: ExecutionContext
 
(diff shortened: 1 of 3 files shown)


test/frame.spec.js
```diff
@@ -208,4 +208,25 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(frame1).not.toBe(frame2);
     });
   });
+
+  describe_fails_ffox('Frame.asElementHandle', function() {
```
aslushnikov: nit: Frame.asElement

docs/api.md
```diff
@@ -2912,6 +2913,9 @@ puppeteer.launch().then(async browser => {
 });
 ```
 
+#### frame.asElement()
```
JoelEinbinder: JSHandle.asElement is a conditional cast. Could be replaced by isElement : boolean and nothing except some ergonomics would change.

Your method is very different. It doesn't cast a frame to a different type, it returns an entirely new handle that happens to be associated with the frame.

We should think a lot about what to call this method. `parentElement`, `frameElement` or `elementHandle` come to mind. But I don't really like any of them. The API should be clear about whether is the `<iframe>` element in the parent document, or the `document.documentElement` of the inner contents of the frame.

test/frame.spec.js
```diff
@@ -208,4 +208,25 @@ module.exports.addTests = function({testRunner, expect}) {
...
+    it('should work', async({page,server}) => {
+      await page.goto(server.EMPTY_PAGE);
+      await utils.attachFrame(page, 'frame1', server.EMPTY_PAGE);
+      const elementHandle = await page.$('#frame1');
+      const frame = await elementHandle.contentFrame();
+      const elementFromFrame = await frame.asElement();
+
+      expect(elementFromFrame._protocolValue).toBe(elementHandle._protocolValue);
```
aslushnikov: I'd do the following here to not rely on privates

```js
const frame = await utils.attachFrame(page, 'frame1', server.EMPTY_PAGE);
const handle = await page.$('#frame1');
expect(await page.evaluate((f1, f2) => f1 === f2, handle, await frame.asElement())).toBe(true);
```

docs/api.md
kblok: How about `getFrameOwner`? 
We would honor the CDP method. But I don't know if it's a friendly name for the user.

aslushnikov: Linter is not happy because of the ordering of this method wrt others

JoelEinbinder: getFrameOwner makes me think its going to return the parent frame.

aslushnikov: hm, I like the `asElement` here - I don't think users of JSHandle.asElement() care much what happens under the hood (and waether we actually want them to care about it).

JoelEinbinder: Another argument: `JSHandle.asElement` return synchronously, where as this returns a promise.

Wdyt about `frame.element()`?

kblok: I like `frame.element`, so we are consistent with `element.contentFrame`. It's not `element.asContentFrame` there.

What do you think @aslushnikov?

aslushnikov: I'll sleep a night and see if there are any second thoughts about it, but it seems good to me at the first glance!

docs/api.md
```diff
@@ -229,6 +229,7 @@
   * [frame.waitForNavigation([options])](#framewaitfornavigationoptions)
   * [frame.waitForSelector(selector[, options])](#framewaitforselectorselector-options)
   * [frame.waitForXPath(xpath[, options])](#framewaitforxpathxpath-options)
+  * [frame.asElement()](#frameaselement)
```
vsemozhetbyt: Nit: it seems the other sections are sorted alphabetically.

kblok:
This is ready for another run :)

docs/api.md
aslushnikov: @pavelfeldman suggests `frame.ownerElement()`. Let's roll!

kblok: @aslushnikov done!

kblok:
@aslushnikov do we need anything else on this one?
It’s hacktoberfest time! 😂

tszyszko:
Why was this closed ? Seems to work as a solution

Issue #4651, opened 2019-07-01
Error details for UnhandledPromiseRejectionWarning


strarsis:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.17.0
* Platform / OS version: WSL (Bash on Windows)
* Node.js version: v12.4.0

**What steps will reproduce the problem?**
```js
const clickAndWait = function(container, target, btn) {
    return container.evaluate((target, btnNext) => {
        return new Promise((resolve, reject) => {
            let observer = new MutationObserver(function() {
                observer.disconnect(); // use the mutation observer only once
                resolve();
            });
            observer.observe(target, { childList: true, subtree: true, });

            // some reasonable timeout, in case loading got stuck
            setTimeout(reject.bind(this, 'timeout'), 5000);

            // puppeteer click() doesn't work, must use native click() in eval
            btn.click();
        });
    }, target, btn);
};

await ph.clickAndWait(iframe, someContainer, btnNext);
````

**What is the expected result?**
The function evaluates, `MutationObserver` is set up, the button is clicked, `MutationObserver` resolves the Promise and `await` proceeds.

**What happens instead?**
Error:
````
(node:4335) UnhandledPromiseRejectionWarning: Error: Evaluation failed: [object Object]
    at ExecutionContext._evaluateInternal (/d/DATA/WDH/Webseite/workspaces/Ari/www/application/noch-frei/crawl/puppeteer-launch/node_modules/puppeteer/lib/ExecutionContext.js:122:13)

(node:4335) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:4335) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
````
How to catch the Promise exception and log the `Object` so I can read the error messages or at least find out what failed during the evaluation?

Also, this helper method worked in the past, after changing some of 
the code and moving its invocations, this error occurs.

strarsis:
So it was an issue with the variable name, `btn` instead `btnNext`.

But the actual issue still holds: How can I get the error message to know it was `btnNext`.

aslushnikov:
Indeed, we do horrible job showing some errors 🤦‍♂️

For example, the following doesn't tell me why exactly evaluation has failed:

```js
[...]

PR #4838, opened 2019-08-13
fix: added a test for errors inside promise

yury-s:
Also bump Chromium revision past 686227 which contained v8 roll with the actual fix.

References #4651

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "node": ">=6.4.0"
   },
   "puppeteer": {
-    "chromium_revision": "686378"
+    "chromium_revision": "681777"
   },
   "scripts": {
     "unit": "node test/test.js",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "node": ">=6.4.0"
   },
   "puppeteer": {
-    "chromium_revision": "681777"
+    "chromium_revision": "686378"
   },
   "scripts": {
     "unit": "node test/test.js",
--- a/test/evaluation.spec.js
+++ b/test/evaluation.spec.js
@@ -256,11 +256,11 @@ module.exports.addTests = function({testRunner, expect}) {
       expect(a.length).toBe(100 * 1024 * 1024);
     });
     it('should throw error with detailed information on exception inside promise ', async({page, server}) => {
-      let error = 'did not catch error';
+      let error = null;
       await page.evaluate(() => new Promise(() => {
-        does_not_exist.click();
+        throw new Error('Error in promise');
       })).catch(e => error = e);
-      expect(error.message).toContain('does_not_exist');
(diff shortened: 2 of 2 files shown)


yury-s:
It seems that the new release will break a bunch of tests, after inspecting CSP changes with Andrey I skipped one of them. There are also crashes in the browser that cause worker tests to timeout:

Failures:
1) Chromium Browser Page Target should create a worker from a service worker (target.spec.js:86:5)
  Message:
    Timeout Exceeded 10000ms

2) Chromium Browser Page Target should create a worker from a shared worker (target.spec.js:93:5)
  Message:
    Timeout Exceeded 10000ms

package.json
```diff
@@ -8,7 +8,7 @@
     "node": ">=6.4.0"
   },
   "puppeteer": {
-    "chromium_revision": "681777"
+    "chromium_revision": "686378"
```
JoelEinbinder: we generally do rolls as their own commits, as they can lots of unforeseen side effects.

test/evaluation.spec.js
```diff
@@ -255,6 +255,13 @@ module.exports.addTests = function({testRunner, expect}) {
       const a = await page.evaluate(() => Array(100 * 1024 * 1024 + 1).join('a'));
       expect(a.length).toBe(100 * 1024 * 1024);
     });
+    it('should throw error with detailed information on exception inside promise ', async({page, server}) => {
+      let error = 'did not catch error';
+      await page.evaluate(() => new Promise(() => {
+        does_not_exist.click();
```
JoelEinbinder: i think this test would be easier to read if you just throw an error here

test/evaluation.spec.js
```diff
@@ -255,6 +255,13 @@ module.exports.addTests = function({testRunner, expect}) {
       const a = await page.evaluate(() => Array(100 * 1024 * 1024 + 1).join('a'));
       expect(a.length).toBe(100 * 1024 * 1024);
     });
+    it('should throw error with detailed information on exception inside promise ', async({page, server}) => {
+      let error = 'did not catch error';
```
JoelEinbinder: nit: `let error = null` to be consistent with our other tests that look at errors

test/evaluation.spec.js
yury-s: Done.

test/evaluation.spec.js
yury-s: I find error messages like 
'Cannot read property 'message' of null'
less informative than
'expect.toContain failed: did not catch error ⊇ Error in promise'

but I agree that the approach should be consistent across the codebase, so changing it to null, we can update the tests later.

yury-s:
This is now blocked on #4841

aslushnikov (approved):

PR #4844, opened 2019-08-14
feat(chromium): roll Chromium to r682225

aslushnikov:
This roll includes:
- https://crrev.com/681997 - Turn on default SiteInstance by default.

The SiteInstance by default was breaking "devtools: true" option, so
there's a new feature we disable now by default.

This keeps pressuring us towards OOPIF support since that's an
inevitable future.

### DIFF
--- a/test/headful.spec.js
+++ b/test/headful.spec.js
@@ -86,7 +86,8 @@ module.exports.addTests = function({testRunner, expect, puppeteer, defaultBrowse
       await rmAsync(userDataDir).catch(e => {});
       expect(cookie).toBe('foo=true');
     });
-    it('OOPIF: should report google.com frame', async({server}) => {
+    // TODO: Support OOOPIF. @see https://github.com/GoogleChrome/puppeteer/issues/2548
+    xit('OOPIF: should report google.com frame', async({server}) => {
       // https://google.com is isolated by default in Chromium embedder.
       const browser = await puppeteer.launch(headfulOptions);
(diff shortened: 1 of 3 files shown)


JoelEinbinder:
Can we instead fix devtools so that it supports site per process = off? I really don't think this is a good solution. Opening devtools (via devtools:true) shouldn't observably change how your puppeteer script runs.

JoelEinbinder:
Is the headless team still using site-per-process=off? If so, I imagine other people will want it off for performance reasons. Maybe this fix should happen on the devtools side, so that devtools doesnt crash with this flag?

test/headful.spec.js
```diff
@@ -116,6 +117,13 @@ module.exports.addTests = function({testRunner, expect, puppeteer, defaultBrowse
       await page.click('body');
       await browser.close();
     });
+    it('should open devtools when "devtools: true" option is given', async({server}) => {
```
JoelEinbinder, line 120: firefox won't like this

aslushnikov:
> If so, I imagine other people will want it off for performance reasons.

I don't think there are any embedders that will use this in the long run - this mode is going away.

test/headful.spec.js
aslushnikov: Why? We run headful tests on CHROME only

JoelEinbinder (approved):
Description is out of date

Issue #4470, opened 2019-05-24
puppeteer-firefox: ERROR: failed to call method 'Browser.deleteCookies'


brandon-bethke-timu:
I receive the following error when I try to run the following example using puppeteer-firefox version 0.5.0:
```
const pptrFirefox = require('puppeteer-firefox');
(async () => {
  const browser = await pptrFirefox.launch({headless: false});
  const page = await browser.newPage();
  await page.goto('https://www.google.com');
  await page.setCookie({name: 'mycookie', value: 'myvalue'});
})();

```

```
$ node sample.js 
(node:92226) UnhandledPromiseRejectionWarning: Error: Protocol error (Browser.deleteCookies): ERROR: failed to call method 'Browser.deleteCookies' with parameters {
  "browserContextId": null,
  "cookies": [
    {
      "url": "https://www.google.com/",
      "name": "mycookie"
    }
  ]
}
Expected "<root>.browserContextId" to be |string|; found |object| `null` instead. _dispatch@chrome://juggler/content/protocol/Dispatcher.js:100:15

    at node_modules/puppeteer-firefox/lib/Connection.js:74:56
    at new Promise (<anonymous>)
    at Connection.send (node_modules/puppeteer-firefox/lib/Connection.js:73:12)
    at Page.deleteCookie (node_modules/puppeteer-firefox/lib/Page.js:115:22)
    at Page.<anonymous> (node_modules/puppeteer-firefox/lib/helper.js:32:23)
    at Page.setCookie (node_modules/puppeteer-firefox/lib/Page.js:135:16)
    at Page.<anonymous> (node_modules/puppeteer-firefox/lib/helper.js:32:23)
    at sample.js:6:14

```

aslushnikov:
@brandon-bethke-neudesic thank you for taking time to file this!

michaelcypher:
I'm seeing the exact same issue for `Browser.getCookies` for puppeteer-firefox version 0.5.0.

```js
const pptrFirefox = require('puppeteer-firefox');
(async () => {
  const browser = await pptrFirefox.launch({headless: false});
  const page = await browser.newPage();
  await page.goto('https://www.google.com');
  const cookies = await page.cookies(page.url());
})();
```

```
node:36006) UnhandledPromiseRejectionWarning: Error: Protocol error (Browser.getCookies): ERROR: failed to call method 'Browser.getCookies' with parameters {
  "browserContextId": null,
  "urls": [
    "https://www.google.com/"
[...]

PR #4850, opened 2019-08-15
fix(firefox): fix cookies in default browser context

aslushnikov:
This patch adds tests and fixes the nodejs part of the problem.
The issue will be fixed once we roll a new version of Firefox.

References #4470


JoelEinbinder (approved):

limonte:
When we can expect the release with this fix?

PR #4851, opened 2019-08-15
test: fix tests to work on node6

aslushnikov:
Our magical node6 transpiler can't handle object spreads :(

Drive-by: use "includes" instead of "startsWith" for devtools URL
since I remember that it used to be "chrome-devtools://", and somehow
I saw it as "devtools://" somewhere.

### DIFF
--- a/test/headful.spec.js
+++ b/test/headful.spec.js
@@ -118,11 +118,11 @@ module.exports.addTests = function({testRunner, expect, puppeteer, defaultBrowse
       await browser.close();
     });
     it('should open devtools when "devtools: true" option is given', async({server}) => {
-      const browser = await puppeteer.launch({...headfulOptions, devtools: true});
+      const browser = await puppeteer.launch(Object.assign({devtools: true}, headfulOptions));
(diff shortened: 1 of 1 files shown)


test/headful.spec.js
```diff
@@ -118,11 +118,11 @@ module.exports.addTests = function({testRunner, expect, puppeteer, defaultBrowse
       await browser.close();
     });
     it('should open devtools when "devtools: true" option is given', async({server}) => {
-      const browser = await puppeteer.launch({...headfulOptions, devtools: true});
+      const browser = await puppeteer.launch(Object.assign({devtools: true}, headfulOptions));
       const context = await browser.createIncognitoBrowserContext();
       await Promise.all([
         context.newPage(),
-        context.waitForTarget(target => target.url().startsWith('devtools://')),
+        context.waitForTarget(target => target.url().includes('devtools://')),
```
JoelEinbinder, line 125: It’s always devtools:// now. We changed it because edge was allergic to the word chrome.

JoelEinbinder (approved):

Issue #4733, opened 2019-07-20
page.waitForRequest() and .waitForResponse() not aborted by browser.close()


davecardwell:
If you call `browser.close()` while `page.waitForFunction()`, `page.waitForNavigation()`, etc. are pending then the `Promise` is immediately rejected.

`page.waitForRequest()` and `page.waitForResponse()`, however, hang around until they time out.

My use case is that I want to use `Promise.race()` on a `page.waitForSelector()` and `page.waitForResponse()` to trigger when either an element is made visible in the `page`, or the `page` is redirected to a specific URL. If `waitForSelector` is resolved first I call `browser.close()` and expect my process to then exit, but it waits until the timeout for `waitForResponse` has passed.

### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.18.1
* Platform / OS version: macOS 10.14.5
* URLs (if applicable): N/A
* Node.js version: 10.15.3

**What steps will reproduce the problem?**

```js
const puppeteer = require("puppeteer");

(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();

  page.setDefaultTimeout(5000);

  const tests = [
    ["waitForFunction", "false"],
    ["waitForNavigation"],
    ["waitForSelector", "#test"],
    ["waitForXPath", "/test"],
    ["waitForRequest", "https://example.com/"],
    ["waitForResponse", "https://example.com/"]
  ];

  let seconds = 0;
  const interval = setInterval(
    () => console.log(`${++seconds} Mississippi`),
    1000
  );

  await Promise.all([
    ...tests.map(([method, ...args]) =>
      page[method](...args).then(
        () => console.log(`${method} resolved`),
        () => console.log(`${method} rejected`)
      )
    ),

    browser.close()
  ]);

  clearInterval(interval);
})();
```

**What is the expected result?**

```
waitForFunction rejected
waitForNavigation rejected
waitForSelector rejected
waitForXPath rejected
waitForRequest rejected
waitForResponse rejected
```

**What happens instead?**

```
waitForFunction rejected
waitForNavigation rejected
waitForSelector rejected
waitForXPath rejected
1 Mississippi
2 Mississippi
3 Mississippi
4 Mississippi
waitForRequest rejected
waitForResponse rejected
```

davecardwell:
@yury-s @aslushnikov Thank you!

alamothe:
[...]

PR #4865, opened 2019-08-19
fix: abort page.waitForRequest/Response when page closes

yury-s:
We'd like to pass an abortion signal inside Helper.waitForEvent in order to interrupt it when browser/page closes. Several approaches have been considered:

1. Pass CDPSession instance as a another parameter to the helper method and listen to Disconnected event on it. It would introduce undesired dependency on the session object.
2. Listen to the CDPSession closure at the call sites (e.g. waitForRequest) and pass an abortion promise which would be fulfilled when such event is fired. The listeners would have to be removed from the session on successful completion of waitForEvent so we'd have to pass some kind of DisposablePromise which would be disposed during cleanup. Such parameter looked somewhat hairy.
3. Create DisconnectPromise on CDPSession. One potential risk with that is all chained promises would hang around until the event is fired which might inadvertently cause memory leaks. On the other hand, adding such promise to Promise.race will remove dependency as soon as the race is finished. So this is the approach we're taking with one tweak: the promise is created locally inside Page. 

Ideally the disconnectPromise would throw when the session is closed but it may lead to uncaught promise errors if all chained promises are resolved, to avoid that the promise is resolved with an Error and Helper.waitForEvent throws it later.

References #4733


lib/helper.js
```diff
@@ -176,9 +176,10 @@ class Helper {
    * @param {!NodeJS.EventEmitter} emitter
    * @param {(string|symbol)} eventName
    * @param {function} predicate
+   * @param {!Promise<!Error>} abortPromise
```
JoelEinbinder, line 180: not you, but this method is missing a JSDoc for timeout. We should really lint for that.

yury-s: Done. Added a type annotation for this.

aslushnikov (approved):
Great, thank you!

PR #4877, opened 2019-08-24
Fix missing package in Docker instructions

dotdoom:
`ttf-freefont` is gone -- looks like it's [replaced](https://packages.debian.org/search?keywords=ttf-freefont) by `fonts-freefont-ttf`, which is [available since oldoldstable](https://packages.debian.org/search?keywords=fonts-freefont-ttf).


aslushnikov (approved):
Thanks!

Issue #3829, opened 2019-01-23
[win] browser.close() blocks forever when combining puppeteer and readline.


ZumiKua:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.11.0
* Platform / OS version: Windows 10 1803
* Node.js version: 10.14.2

**What steps will reproduce the problem?**

````javascript
const puppeteer = require('puppeteer');
const readline = require("readline");
const rl = readline.createInterface({
    input: process.stdin,
    output: process.stdout
});
async function main() {
    browser = await puppeteer.launch();
    rl.close();
    await browser.close();  
    console.log("this line will not be executed.");
}
main();
````

1. run above program, you will find program blocks at `await browser.close();` and `console.log("...")` will never be reached.
2. move rl.close() after `await browser.close();` solves this problem.
3. According to [this answer](https://stackoverflow.com/a/54318611/2661043), this issue also disappears if terminal option of readline.createInterface() is set to false.

**What is the expected result?**
Program exits normally. And "this line will not be executed." should be printed.

**What happens instead?**
Program hangs, no output.

**Comment by vsemozhetbyt** [comment]
Can reproduce on Windows 7 x64.

When I run this code in the Chome inspector step by step, the last message appears in the inspector console (but not in the system shell) and then I cannot run any expression in the inspector console as if the ENTER is intercepted and buffered: when I disconnect the inspector, the system prompt appears several times as if all ENTERs are released.

**Comment by FremyCompany** [comment]
I have a similar problem when I combine puppeteer with EdgeJS and create a thread in C# and call `browser.close()` before killing the C# thread.

I therefore suspects that somehow browser.close() is confused by threads created by other modules.

**Comment by 2kindsofcs** [comment]
puppeteer version 1.19.0 still has same issue. I found out `_closeCallback` function is the problem.
```javascript
  async close() {
    await this._closeCallback.call(null); // this statement is a bottleneck 
    this.disconnect();
  }
```
when `this._closeCallback` is declared such as `this._closeCallback = function() {}`, it seemed the code runs well. but when I tried several times more, sometimes it failed.

PR #4897, opened 2019-09-02
Use arrow function for default this._closeCallback #3829

2kindsofcs:
`browser.close()` is blocked when used with `readline` in a specific way. 
this issue can be fixed if default value of `this._closeCallback` is assigned with an anonymous function by the arrow function, rather than function constructor.


2kindsofcs:
@googlebot I signed it!

2kindsofcs:
I checked out that the same problem still remaining even after using arrow function. it seems depending on environment the issue might look solved, though it is not solved at all.

Issue #1893, opened 2018-01-24
Document the behavior when the same header is present multiple times


Khady:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: puppeteer@1.0.0 and puppeteer@1.0.0-next.1516669982697
* Platform / OS version: debian
* URLs (if applicable): https://www.google.com.sg
* Node.js version: 8.9.4

**What steps will reproduce the problem?**

Some pages return multiple time the same header when requested. `set-cookie` for example.

When requesting the same page using puppeteer, there is only one entry in the resulting headers object. And the value is a string, not an array. Both values for the header `set-cookie` are present in this string, separated by a `\n`.

It would be nice to document this properly. Or even better, to change the type of the headers object to something like this in typescript:

```typescript
interface headers {
  [header: string]: Array<{ key: string, value: string | boolean | number }>
}
```

Which is the same format used in some parts of aws:
https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-event-structure.html

> headers (read/write)
> 
> The headers in the request. The keys in the headers object are
> lowercase versions of the header names in the HTTP request. These
> lowercase keys give you case-insensitive access to the header
> values. Each header (for example, headers["accept"] or
> headers["host"]) is an array of key-value pairs, where key is the
> case-sensitive name of the header as it appears in an HTTP request and
> value is a header value. The number of elements in the array is the
> number of times that a header is repeated in an HTTP request.

Or a type like the one used in axios:

```typescript
interface headers {
  [header: string]: Array<string | boolean | number>
}
```

Also, I didn't test the behavior in request interception. I don't know if it is possible to give a header with a newline in the middle to `request.respond`. But it would be more natural to have a list rather than to rely on a trick like this.

Probably need to test the merge behavior of two headers with the same key except the case in request interception too.

```
$ curl -Is "https://www.google.com.sg" | grep set-cookie
set-cookie: 1P_JAR=2018-01-24-10; expires=Fri, 23-Feb-2018 10:14:01 GMT; path=/; domain=.google.com.sg
[...]

PR #4915, opened 2019-09-07
fix(network): handle newline delimited headers in request interception

adrianlyjak:
The current behavior of `\n` in headers seems to, somewhere down the line, screw up chrome's idea of whether a request is complete. Page navigation will hang until it times out.

Since the existing behavior for returned header models on request and response objects is to join same named headers with a line break (`\n`) as a `Record<string, string>`. Seems like at the very least `request.respond` and `request.continue` should accept `\n`s without crashing. `\n` is an illegal character in http headers, so it's safe to assume they mean the header value is a list. 

This change allow multiple headers to be passed through in `request.respond` and partially fixes `request.continue`

Partial fix for #1893 (related to  #4379)

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -3494,7 +3494,7 @@ ResourceType will be one of the following: `document`, `stylesheet`, `image`, `m
 #### request.respond(response)
 - `response` <[Object]> Response that will fulfill this request
   - `status` <[number]> Response status code, defaults to `200`.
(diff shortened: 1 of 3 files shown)


adrianlyjak:
@googlebot I signed it!

adrianlyjak:
@aslushnikov I'd appreciate your feedback on this

PR #4929, opened 2019-09-11
test: mark Page.close test as failing in Firefox

mathiasbynens:
This should fix the CI build (#4930).

### DIFF
--- a/test/page.spec.js
+++ b/test/page.spec.js
@@ -77,7 +77,7 @@ module.exports.addTests = function({testRunner, expect, headless, puppeteer, CHR
       await newPage.close();
       expect(newPage.isClosed()).toBe(true);
     });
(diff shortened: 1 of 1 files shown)


aslushnikov (approved):
i think this is fixed in a new version of juggler. Shall we roll it as well?

mathiasbynens:
SGTM. I could update https://github.com/GoogleChrome/puppeteer/blob/a5f03ce1c85c2630739f2a894a303c68632aa250/experimental/puppeteer-firefox/package.json#L12 myself but I would need access to the upload bucket to make sure a corresponding download exists. @aslushnikov could you give `mathiasb@google.com` access?

paulirish:
ayo guys. just a driveby, but did this bucket access get sorted out?

mathiasbynens:
Not yet

PR #4934, opened 2019-09-11
chore: avoid unnecessary zero-delays in input code

yury-s:
If delay is 0 we'll now skip the wait step.

### DIFF
--- a/lib/Input.js
+++ b/lib/Input.js
@@ -156,16 +156,15 @@ class Keyboard {
    * @param {{delay: (number|undefined)}=} options
    */
   async type(text, options) {
-    let delay = 0;
-    if (options && options.delay)
-      delay = options.delay;
+    const delay = (options && options.delay) || null;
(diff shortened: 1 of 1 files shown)


mathiasbynens:
`setTimeout(fn, 0)` is clamped to ~4 ms, so there is a difference between “waiting for 0 ms” and “not waiting at all”. This would be an observable and potentially breaking change.

yury-s:
> `setTimeout(fn, 0)` is clamped to ~4 ms, so there is a difference between “waiting for 0 ms” and “not waiting at all”. This would be an observable and potentially breaking change.

This delay is inside a method on client side (inside puppeteer). It'd be wrong to rely on any effects of it in the browser as it's not far from the length of transport delays in the protocol. At the same time current implementations of type() and press() with no delay will have slightly different behavior.

JoelEinbinder (approved):

PR #4941, opened 2019-09-14
docs(troubleshooting.md): Update travis examples

ScriptSmith:
For the last two days, travis builds using trusty have been failing to install google chrome when configured to do so with:

```yaml
addons:
  chrome: stable
```
https://stackoverflow.com/questions/57903415/travis-ci-chrome-62-instead-of-77
https://github.com/travis-ci/travis-ci/issues/9361

This will probably be fixed in time, but it prompted me to update to the latest LTS distribution (bionic) in my project.

The troubleshooting guide still only gives an example using trusty, so I've updated it with instructions that reflect how a configuration with bionic works.

It's the same except xvfb is started as a service, and libnss3 isn't installed

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -191,7 +191,27 @@ Tips-n-tricks:
   proper sandboxing
 - [xvfb](https://en.wikipedia.org/wiki/Xvfb) should be launched in order to run Chromium in non-headless mode (e.g. to test Chrome Extensions)
 
-To sum up, your `.travis.yml` might look like this:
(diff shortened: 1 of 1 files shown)


jschfflr:
@ScriptSmith Thanks for your contribution! Could you verify that this change is still relevant and if yes, please resolve the merge conflict? Happy to get this landed :)

jschfflr:
Closing this as there haven't been any updates on this for a while. Feel free to reopen if you feel like this is still relevant!

PR #4949, opened 2019-09-17
feat(api): add page.emulateTimezone(timezoneId)

mathiasbynens:
This enables dynamically changing the timezone of the page.

Blocked on a Chromium roll to v79.0.3923.0 (r699550) or later, though.


docs/api.md
```diff
@@ -1279,6 +1280,10 @@ List of all available devices is available in the source code: [DeviceDescriptor
 - `mediaType` <?[string]> Changes the CSS media type of the page. The only allowed values are `'screen'`, `'print'` and `null`. Passing `null` disables media emulation.
 - returns: <[Promise]>
 
+#### page.emulateTimezone(timezoneId)
+- `timezoneId` <?[string]> Changes the timezone of the page. Passing `null` disables timezone emulation.
```
paulirish: link to some canonical list of timezoneIds?

test/emulation.spec.js
```diff
@@ -114,4 +114,20 @@ module.exports.addTests = function({testRunner, expect, puppeteer}) {
...
+      await page.emulateTimezone('Pacific/Honolulu');
+      expect(await page.evaluate(() => new Date(1479579154987).toString())).toBe('Sat Nov 19 2016 08:12:34 GMT-1000 (Hawaii-Aleutian Standard Time)');
+
+      await page.emulateTimezone('America/Argentina/Buenos_Aires');
+      expect(await page.evaluate(() => new Date(1479579154987).toString())).toBe('Sat Nov 19 2016 15:12:34 GMT-0300 (Argentina Standard Time)');
+
+      await page.emulateTimezone('Europe/Berlin');
+      expect(await page.evaluate(() => new Date(1479579154987).toString())).toBe('Sat Nov 19 2016 19:12:34 GMT+0100 (Central European Standard Time)');
+    });
+  });
```
paulirish, line 185: maybe a test for passing an invalid tzId and verifying the rejection msg?

mathiasbynens: I agree rejecting on invalid inputs would be useful. However, the CDP API doesn't signal whether the time zone is valid or not --- it just silently ignores invalid inputs.



Any ideas? We could hardcode a list of recognized time zones in Puppeteer and reject on unknown inputs (similar to what we do for e.g. `page.emulateMedia`), but that would mean we'd have to maintain the list.

docs/api.md
mathiasbynens: Good idea. I'll add a link to https://cs.chromium.org/chromium/src/third_party/icu/source/data/misc/metaZones.txt?rcl=faee8bc70570192d82d2978a71e2a615788597d1.

test/emulation.spec.js
mathiasbynens: We could also change the CDP to return some kind of signal if the input wasn't recognized, but that seems to deviate from other `set*Override` APIs.

mathiasbynens: CDP has now been updated accordingly: https://chromium-review.googlesource.com/c/chromium/src/+/1822557 Once this change rolls into Puppeteer's `chromium_revision` we can make use of the negative signal to detect the failure case.

mathiasbynens: Patch updated accordingly. It's blocked on a Chromium roll to v79.0.3923.0 (r699550) or later, though.

crobinson42:
@mathiasbynens I'm curious to know what's needed or if this will make it into puppeteer? thx!

mathiasbynens:
@crobinson42 It's currently blocked on a Chromium roll. You can track the progress of that here: #5058

It sounds like you have a use case for this feature? Please let us know what you plan on doing with it :)

lib/Page.js
```diff
@@ -808,6 +808,19 @@ class Page extends EventEmitter {
...
   }
 
+  /**
+   * @param {?string} timezoneId
+   */
+  async emulateTimezone(timezoneId) {
+    try {
+      await this._client.send('Emulation.setTimezoneOverride', {timezoneId: timezoneId || ''});
+    } catch (exception) {
+      if (exception.message.includes('Invalid timezone'))
```
szuend, line 834: nit: Braces for multi-line if.

test/emulation.spec.js
```diff
@@ -114,4 +114,29 @@ module.exports.addTests = function({testRunner, expect, puppeteer}) {
...
+      expect(await page.evaluate(() => date.toString())).toBe('Sat Nov 19 2016 08:12:34 GMT-1000 (Hawaii-Aleutian Standard Time)');
+
+      await page.emulateTimezone('America/Buenos_Aires');
+      expect(await page.evaluate(() => date.toString())).toBe('Sat Nov 19 2016 15:12:34 GMT-0300 (Argentina Standard Time)');
+
+      await page.emulateTimezone('Europe/Berlin');
+      expect(await page.evaluate(() => date.toString())).toBe('Sat Nov 19 2016 19:12:34 GMT+0100 (Central European Standard Time)');
+    });
+
+    fit('should throw for invalid timezone IDs', async({page, server}) => {
```
szuend:
```suggestion
    it('should throw for invalid timezone IDs', async({page, server}) => {
```

szuend (approved):
Thanks!

lib/Page.js
mathiasbynens: The current puppeteer code style doesn't allow braces around single-line `if` bodies.

PR #4993, opened 2019-10-01
feat: add custom debugging port option

tanhauhau:
I found out that the only way to customise the debugging port for now is to pass in the `--remote-debugging-port` as `args`:

```
puppeteer.laucnh({ args: ['--remote-debugging-port=12345'] })
```


lib/Launcher.js
```diff
@@ -249,6 +250,8 @@ class Launcher {
           '--mute-audio'
       );
     }
+    if (debuggingPort)
```
jschfflr: In the documentation, you stated that passing `0` for the `debuggingPort` will assign a random port but I think this if clause will prevent that from happening.

jschfflr (changes requested):
@tanhauhau Thanks for your contribution! 
Could you update this patch to resolve the merge conflicts and make sure to add test coverage for this feature? `chromiumonly.spec.ts` looks like a promising starting point but ideally, Firefox would also be covered.

jschfflr:
@googlebot I consent.

jschfflr (approved):

PR #5010, opened 2019-10-06
docs(api.md): clarified js example with type method

QuantumArchive:
This simply clarifies the use of typing as an async method that should be awaited


QuantumArchive:
@googlebot I signed it!

mathiasbynens (approved):

Issue #4906, opened 2019-09-04
feat: emulate prefers-color-scheme 


manucorporat:
It would be interesting to simulate different values of `prefers-color-scheme`, so developers could write visual tests for light and dark mode.

Possible API:
```
page.emulate({
  viewport: {
     colorScheme: 'dark'
  }
});
```

There is an article of how this could be achieved: https://www.bram.us/2019/09/04/take-both-light-and-dark-mode-screenshots-with-puppeteer/

But seems rather hacky and unperformant, since it needs to boot up two completely different chrome instances.

tomayac:
Heya, @tomayac here, author of the "hacky" and "unperformant" solution https://github.com/tomayac/dark-mode-screenshot. I would very much +1 a *native* way in Puppeteer, as typically the flags that my approach uses at the moment tend to be rather ephemeral. 

Different than the proposal above, I would, however, vouch for a more generic approach that allows for testing all current and potential future [User Preference Media Features](https://drafts.csswg.org/mediaqueries-5/#mf-user-preferences), like `prefers-reduced-motion`, `prefers-contrast` etc.

My idea would be to have a field where one can shove the different media features à la

```js
{
  …
  userPreferences: {
    "prefers-color-scheme": 'dark',
    "prefers-reduced-motion": 'reduce',
  }
  …
}
```

I'm not that familiar with the Puppeteer API, so would leave it to the authors to decide where such a field would live best.

manucorporat:
@tomayac sorry for the adjectives haha it was very inspiring though! 

Should not "userPreferences" be an object?
```
userPreferences: {
 'prefers-color-scheme': 'dark',
 'prefers-reduced-motion': true
}
```

i like so much the API! 

At Stencil as part of our e2e API, we have a way to run the tests under different simulated devices, it would be great to configure this user preferences as well!

tomayac:
> [S]orry for the [adjectives]

No offense taken, no worries. 

> Should not "userPreferences" be an object?

🙈 Of course, updated my comment. Thanks!

mathiasbynens:
Puppeteer already has [a `page.emulateMedia` API](https://github.com/GoogleChrome/puppeteer/blob/v1.19.0/docs/api.md#pageemulatemediamediatype), although it currently only supports `'screen'` and `'print'`. One idea would be to extend it to support newer user preference media types as Chrome adds support for them:

```js
await page.emulateMedia('prefers-color-scheme: dark');
[...]

PR #5012, opened 2019-10-07
feat(api): add page.emulateMedia{Type,Features}

mathiasbynens:
**Note: this PR is just to illustrate what the API _could_ look like. The goal is to get feedback and iterate on the approach.**

CDP supports emulating CSS media features (e.g. `prefers-color-scheme`, `prefers-reduced-motion`) in addition to media types (e.g. `screen`, `print`) [as of Chromium v79.0.3929.0+](https://chromium-review.googlesource.com/c/chromium/src/+/1821608).

An open question is, how do we expose this functionality as a Puppeteer API? IMHO, there's two main options:

1. Provide two separate APIs: one for media types (`page.emulateMediaType(type)`), and one for media features: `page.emulateMediaFeatures(features)`
2. Provide a single overloaded API that supports media types and media features: `page.emulateMedia(typeOrFeatures)`

This PR implements the first option, which feels cleanest to me. Note that in this WIP PR, `page.emulateMedia(mediaType)` has been renamed to `page.emulateMediaType(type)` (a breaking change!) to make the distinction between the two APIs more clear. We could of course decide to keep the old API for the time being as an alias.

Thoughts?

Ref. #4906.

Blocked on a Chromium roll to v79.0.3929.0 (r701361).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -108,7 +108,8 @@
   * [page.coverage](#pagecoverage)
   * [page.deleteCookie(...cookies)](#pagedeletecookiecookies)
   * [page.emulate(options)](#pageemulateoptions)
-  * [page.emulateMedia(mediaType)](#pageemulatemediamediatype)
+  * [page.emulateMediaFeatures(features)](#pageemulatemediafeaturesfeatures)
+  * [page.emulateMediaType(type)](#pageemulatemediatypetype)
   * [page.evaluate(pageFunction[, ...args])](#pageevaluatepagefunction-args)
   * [page.evaluateHandle(pageFunction[, ...args])](#pageevaluatehandlepagefunction-args)
   * [page.evaluateOnNewDocument(pageFunction[, ...args])](#pageevaluateonnewdocumentpagefunction-args)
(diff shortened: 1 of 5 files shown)


lib/Page.js
```diff
@@ -800,11 +800,27 @@ class Page extends EventEmitter {
   }
 
   /**
-   * @param {?string} mediaType
+   * @param {?string} type
    */
-  async emulateMedia(mediaType) {
```
bmeurer, line 806: Shouldn't we retain `emulateMedia` and mark it deprecated?

mathiasbynens: That's part of what I'd like to discuss here. I think we should definitely retain it for v1. Then again, the upcoming v2 gives us a potential "clean break" opportunity.

bmeurer: Sounds good. I like the separation of the two APIs reflected explicitly in puppeteer. :+1:

mathiasbynens:
Added the old `page.emulateMedia` back as an alias.

docs/api.md
```diff
@@ -1276,10 +1278,78 @@ puppeteer.launch().then(async browser => {
 
 List of all available devices is available in the source code: [DeviceDescriptors.js](https://github.com/GoogleChrome/puppeteer/blob/master/lib/DeviceDescriptors.js).
 
-#### page.emulateMedia(mediaType)
-- `mediaType` <?[string]> Changes the CSS media type of the page. The only allowed values are `'screen'`, `'print'` and `null`. Passing `null` disables media emulation.
+#### page.emulateMedia(type)
+- `type` <?[string]> Changes the CSS media type of the page. The only allowed values are `'screen'`, `'print'` and `null`. Passing `null` disables CSS media emulation.
 - returns: <[Promise]>
 
+**Note:** This method is deprecated, and only kept around as an alias for backwards compatibility. Use [`page.emulateMediaType(type)`](#pageemulatemediamediatypetype) instead.
```
szuend: nit: This should probably be #pageemulatemediatypetype (without the extra media)

lib/Page.js
```diff
@@ -801,11 +801,27 @@ class Page extends EventEmitter {
...
+  /**
+   * @param {?Array<MediaFeature>} features
+   */
+  async emulateMediaFeatures(features) {
+    if (features === null)
+      await this._client.send('Emulation.setEmulatedMedia', {features: null});
+    if (Array.isArray(features)) {
+      features.every(mediaFeature => {
+        const name = mediaFeature.name;
+        assert(/^prefers-(?:color-scheme|reduced-motion)$/.test(name), 'Unsupported media feature: ' + name);
```
szuend, line 820: super-nit: I had to lookup what "?:" does. Do we really care if the match is captured?

lib/Page.js
szuend: Is there a tracking issue documenting that we should remove `emulateMedia` for 2.0?

szuend (approved):
Thanks! lgtm

lib/Page.js
mathiasbynens: > Do we really care if the match is captured?

There's some overhead to capturing groups of the form `(pattern)`. In this case we don't need to capture at all, so we use a non-capturing group of the form `(?:pattern)` instead.

Another approach here would be to maintain a set of supported media features, i.e.

```js
const SUPPORTED_MEDIA_FEATURES = new Set([
  'prefers-color-scheme',
  'prefers-reduced-motion',
]);
```

...and then check that instead of using a regular expression. This initially felt like overkill to me, because we currently only support two CSS media features, but maybe it's more readable / maintainable going forward. WDYT?

docs/api.md
mathiasbynens: GoodGood catchcatch!

lib/Page.js
szuend: Like I said, super-nit. I am fine with the regex as is. It is just some cognitive overhead for a tiny performance gain.

mathiasbynens:
> Is there a tracking issue documenting that we should ~~remove~~ deprecate emulateMedia for 2.0?

#2079

docs/api.md
```diff
@@ -1276,10 +1278,78 @@ puppeteer.launch().then(async browser => {
...
+await page.evaluate(() => matchMedia('(prefers-reduced-motion: reduce)').matches));
+// → true
+await page.evaluate(() => matchMedia('(prefers-color-scheme: no-preference)').matches));
+// → false
+
+await page.emulateMediaFeatures([{ name: 'prefers-reduced-motion', value: 'reduce' }]);
+await page.evaluate(() => matchMedia('(prefers-reduced-motion: reduce)').matches));
+// → true
+await page.evaluate(() => matchMedia('(prefers-color-scheme: no-preference)').matches));
+// → false
```
vsemozhetbyt, line 1312: It seems these blocks are duplicate.

mathiasbynens: @vsemozhetbyt Wanna submit a patch?

vsemozhetbyt: Hope it's right: https://github.com/GoogleChrome/puppeteer/pull/5072

PR #5015, opened 2019-10-07
docs(api): add chromium version for puppeteer 1.20

kblok:
Added version according to the release https://github.com/GoogleChrome/puppeteer/releases/tag/v1.20.0

PR #5023, opened 2019-10-10
docs(contributing): mention manual docs/api updates

mathiasbynens:
Ref. #5015.


kblok:
My two cents here. I commented this on #5015 

This was being done as part of step 4 ´ Source Code: mark post-release´.

image

It's not it should be in that way.  It's just how it was being done in the past.

mathiasbynens:
Thanks for pointing that out, @kblok. Doing it the way this patch proposes means the README on npm will also have the new version included, which seems strictly better.

PR #5029, opened 2019-10-11
feat(page): introduce newPage in background

### LINKED ISSUE #5025: Is there a target.createTarget mapping from CDP to puppeteer
I see that I can create `browser.newPage()` but it doesn't the `background` flag that I am looking for. I want to create a new page which opens in the background, essentially replicating what a `CMD+Click` does to open a new window but in the background.

Is there a mapping to `target.createTarget` in puppeteer that I can use? I can directly use CDP commands but I am looking for a way to do that from puppeteer.

kblok:
closes #5025 

This flag didn´t work for me in headless mode, so I added the test on the headful spec.
I could also have added the [newWindow](https://chromedevtools.github.io/devtools-protocol/tot/Target#method-createTarget) flag, but I don´t know what that could be tested.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -718,7 +718,7 @@ Indicates that the browser is connected.
 
 #### browser.newPage([options])
 - `options` <[Object]>
-  - `background` <[boolean]> Defaults to `false`. Whether to run open the new page in background or not.
+  - `background` <[boolean]> Defaults to `false`. Whether to run open the new page in the background or not.
 - returns: <[Promise]<[Page]>>
 
 Promise which resolves to a new [Page] object. The [Page] is created in a default browser context.
@@ -849,7 +849,7 @@ The default browser context is the only non-incognito browser context.
 
 #### browserContext.newPage([options])
 - `options` <[Object]>
-  - `background` <[boolean]> Defaults to `false`. Whether to run open the new page in background or not.
+  - `background` <[boolean]> Defaults to `false`. Whether to run open the new page in the background or not.
 - returns: <[Promise]<[Page]>>
 
 Creates a new page in the browser context.
--- a/docs/api.md
+++ b/docs/api.md
@@ -47,7 +47,7 @@
   * [browser.defaultBrowserContext()](#browserdefaultbrowsercontext)
   * [browser.disconnect()](#browserdisconnect)
   * [browser.isConnected()](#browserisconnected)
-  * [browser.newPage()](#browsernewpage)
+  * [browser.newPage([options])](#browsernewpageoptions)
   * [browser.pages()](#browserpages)
   * [browser.process()](#browserprocess)
   * [browser.target()](#browsertarget)
@@ -64,7 +64,7 @@
   * [browserContext.clearPermissionOverrides()](#browsercontextclearpermissionoverrides)
   * [browserContext.close()](#browsercontextclose)
   * [browserContext.isIncognito()](#browsercontextisincognito)
-  * [browserContext.newPage()](#browsercontextnewpage)
+  * [browserContext.newPage([options])](#browsercontextnewpageoptions)
   * [browserContext.overridePermissions(origin, permissions)](#browsercontextoverridepermissionsorigin-permissions)
   * [browserContext.pages()](#browsercontextpages)
   * [browserContext.targets()](#browsercontexttargets)
@@ -716,7 +716,9 @@ Disconnects Puppeteer from the browser, but leaves the Chromium process running.
 
 Indicates that the browser is connected.
 
-#### browser.newPage()
+#### browser.newPage([options])
+- `options` <[Object]>
+  - `background` <[boolean]> Defaults to `false`. Whether to run open the new page in background or not.
 - returns: <[Promise]<[Page]>>
 
 Promise which resolves to a new [Page] object. The [Page] is created in a default browser context.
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -716,7 +716,9 @@ Disconnects Puppeteer from the browser, but leaves the Chromium process running.
 
 Indicates that the browser is connected.
 
-#### browser.newPage()
+#### browser.newPage([options])
+- `options` <[Object]>
+  - `background` <[boolean]> Defaults to `false`. Whether to run open the new page in background or not.
```
```suggestion
  - `background` <[boolean]> Defaults to `false`. Whether to run open the new page in the background or not.
```

docs/api.md
```diff
@@ -845,7 +847,9 @@ The default browser context is the only non-incognito browser context.
 
 > **NOTE** the default browser context cannot be closed.
 
-#### browserContext.newPage()
+#### browserContext.newPage([options])
+- `options` <[Object]>
+  - `background` <[boolean]> Defaults to `false`. Whether to run open the new page in background or not.
```
```suggestion
  - `background` <[boolean]> Defaults to `false`. Whether to run open the new page in the background or not.
```

test/headful.spec.js
```diff
@@ -147,5 +147,19 @@ module.exports.addTests = function({testRunner, expect, puppeteer, defaultBrowse
       await browser.close();
     });
   });
+
+  describe('Page.newPage', function() {
+    it_fails_ffox('newPage in background', async() => {
```
mathiasbynens:
```suggestion
    it_fails_ffox('newPage in background', async () => {
```

mathiasbynens (approved):

mathiasbynens:
I do not have access to edit your PR, so please apply my suggestions. Thanks!

kblok:
@mathiasbynens done!

JoelEinbinder:
What does it mean to open the page in the background? What if its the only page? Can I still move it to the foreground? Why do I have to decide that the page is "background" when I create it, and can't dynamically change it from background to foreground?

kblok:
@JoelEinbinder 

>What does it mean to open the page in the background? 

We are trying to emulate the control/cmd+click behavior on a link.
e.g. You go to a page, you grab all the links from that page and you want to start opening new tabs without losing the focus on the main tab.

>What if its the only page? 

The argument is ignored. This is CDP behavior.

>Can I still move it to the foreground? Why do I have to decide that the page is "background" when I create it, and can't dynamically change it from background to foreground?

Yes, you can call `await page.bringToFront();`

JoelEinbinder:
> @JoelEinbinder
> 
> > What does it mean to open the page in the background?
> 
> We are trying to emulate the control/cmd+click behavior on a link.
> e.g. You go to a page, you grab all the links from that page and you want to start opening new tabs without losing the focus on the main tab.

Why don't you just send a control+click then?

> 
> > What if its the only page?
> 
> The argument is ignored. This is CDP behavior.
> 

CDP behavior can be however we decide it should behave. I don't like how unpredictable this API would be.

> > Can I still move it to the foreground? Why do I have to decide that the page is "background" when I create it, and can't dynamically change it from background to foreground?
> 
> Yes, you can call `await page.bringToFront();`

This feels like an emulated viewport thing to me. How about we add "visibility" to the viewport options?

kblok:
@VikramTiwari do you want to share your use case scenario here?

kblok:
>CDP behavior can be however we decide it should behave.

That's good for me to know @JoelEinbinder. I thought that Puppeteer should eventually support all the API CDP already supports.
Don't you think this might be confusing for some users? "Why I can do this using CDP and in Puppeteer is not considered useful?". Again, I'm not saying it's wrong, it's good to know the approach you have.

JoelEinbinder:
> > CDP behavior can be however we decide it should behave.
> 
> That's good for me to know @JoelEinbinder. I thought that Puppeteer should eventually support all the API CDP already supports.
> Don't you think this might also be confusing for some users, "Why I can do this using CDP and in Puppeteer is not considered useful?". Again, I'm not saying it's wrong, it's good to know, the approach you have.

Yeah, my opinion on it might not be canonical, but I feel that Puppeteer should stand alone as a great API for controlling the web.

The vast majority of the work for Puppeteer happens upstream in Chromium and Firefox, working on protocols. In Chromium, we share our protocol with a lot of different products that are under the DevTools umbrella, so sometimes we get features for free. But in practice, it is extremely rare that a CDP method will translate into a Puppeteer API without any upstream work.

You make a good point though. Someone thought this was a good enough API for CDP, so why do we have a different standard for what goes into Puppeteer? The answer is mainly due to stability. When adding new APIs to Puppeteer, we are committing to support them forever for every browser. When things are added to CDP, they are free to expose internal Chromium concepts. Even when the API stays the same, CDP methods often do different things in different versions of Chromium.

When you are a CDP client such as Puppeteer or DevTools, it is a bit like you are writing code within Chromium itself. You get the full power of everything Chromium has to offer, but in exchange you have to keep up with the constant refactoring and behavior changes that occur in a project that size. CDP is very powerful. That's why we expose it through the `createCDPSession` method. But it should only be for the brave.

Puppeteer's goal is to swallow the turbulence of the browser protocol and present a nice stable API. That's why there are APIs in CDP that are not in Puppeteer.

kblok:
@JoelEinbinder feel free to close the PR if you consider that appropriate.

VikramTiwari:
> Why don't you just send a control+click then?


Hey @JoelEinbinder
The aim is to run some automation task on a browser that is already open. This allows us to bypass 2-factor and other auth (because user is already logged in). The task can be triggered by an external process which would cause puppeteer to open a new tab in the browser and start execution.

The aim is to run this automation task in a background tab. But when we open a new tab using `browser.newPage()`, it always opens the tab in foreground. I can make consecutive tabs go in background by opening them using CMD + Click but there is no way to start this first tab in background.

Regarding your comment about why it should or should not be included: I fully agree that there is a lot of burden on puppeteer to maintain any features and options that it takes from CDP, biggest of which is the changes between different versions. Though I would assume that the end result of how a new page opens (in background or foreground) will remain pretty stable. To me it feels like we are exposing something that is already pretty stable in terms of end-result and not just API.

PR #5068, opened 2019-10-23
chore: use map instead of plain object

mathiasbynens:


alfredwesterveld (commented):
Curious why you changed this?

mathiasbynens:
Map is the more appropriate data structure here. We don't need to serialize, so we don't need to keep this a plain object.

PR #5121, opened 2019-11-03
style(page): Refactored code to remove redundancy

yashLadha:
There were multiple if block in the logic that can be ironed
out into a single statement, which is much more robust and readable.

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -892,10 +892,8 @@ class Page extends EventEmitter {
       screenshotType = options.type;
     } else if (options.path) {
       const mimeType = mime.getType(options.path);
-      if (mimeType === 'image/png')
-        screenshotType = 'png';
-      else if (mimeType === 'image/jpeg')
-        screenshotType = 'jpeg';
+      const mimeMatch = mimeType.match(/^image\/(png|jpeg)$/);
+      screenshotType = Array.isArray(mimeMatch) ? mimeMatch[1] || null : null;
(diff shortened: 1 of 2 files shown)


yashLadha:
There is some flakiness around `accessibility.spec` which is probably targeted in this pr: https://github.com/GoogleChrome/puppeteer/pull/5116

yashLadha:
@aslushnikov  Can you please review the PR?

test/accessibility.spec.js
```diff
@@ -42,6 +42,7 @@ module.exports.addTests = function({testRunner, expect, FFOX}) {
         </select>
       </body>`);
 
+      await page.focus('[placeholder="Empty input"]');
```
mathiasbynens, line 45: Please remove these unrelated changes from this patch

lib/Page.js
```diff
@@ -892,10 +892,8 @@ class Page extends EventEmitter {
       screenshotType = options.type;
     } else if (options.path) {
       const mimeType = mime.getType(options.path);
-      if (mimeType === 'image/png')
-        screenshotType = 'png';
-      else if (mimeType === 'image/jpeg')
-        screenshotType = 'jpeg';
+      const mimeMatch = mimeType.match(/^image\/(png|jpeg)$/);
+      screenshotType = Array.isArray(mimeMatch) ? mimeMatch[1] || null : null;
```
mathiasbynens, line 896: This is less readable than the original code.

PR #5137, opened 2019-11-08
feat(launcher): Provide option to run puppeteer with different browsers

mjzffr:
This is an initial refactoring of Launcher to enable run-time browser selection in Puppeteer, starting with Firefox. On our side, we're tracking this and related work at https://bugzilla.mozilla.org/show_bug.cgi?id=1590467

/cc @whimboo @andreastt @AutomatedTester

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -389,7 +389,7 @@ If Puppeteer doesn't find them in the environment during the installation step, 
 - `PUPPETEER_DOWNLOAD_HOST` - overwrite URL prefix that is used to download Chromium. Note: this includes protocol and might even include path prefix. Defaults to `https://storage.googleapis.com`.
 - `PUPPETEER_CHROMIUM_REVISION` - specify a certain version of Chromium you'd like Puppeteer to use. See [puppeteer.launch([options])](#puppeteerlaunchoptions) on how executable path is inferred. **BEWARE**: Puppeteer is only [guaranteed to work](https://github.com/GoogleChrome/puppeteer/#q-why-doesnt-puppeteer-vxxx-work-with-chromium-vyyy) with the bundled Chromium, use at your own risk.
 - `PUPPETEER_EXECUTABLE_PATH` - specify an executable path to be used in `puppeteer.launch`. See [puppeteer.launch([options])](#puppeteerlaunchoptions) on how the executable path is inferred. **BEWARE**: Puppeteer is only [guaranteed to work](https://github.com/GoogleChrome/puppeteer/#q-why-doesnt-puppeteer-vxxx-work-with-chromium-vyyy) with the bundled Chromium, use at your own risk.
-- `PUPPETEER_PRODUCT` - specify which browser you'd like Puppeteer to use. Must be one of `chrome` or `firefox`. This is exposed in [`puppeteer.product`](#puppeteerproduct)
+- `PUPPETEER_PRODUCT` - specify which browser you'd like Puppeteer to use. Must be one of `chrome` or `firefox`. Setting `product` programmatically in [puppeteer.launch([options])](#puppeteerlaunchoptions) supercedes this environment variable. The product is exposed in [`puppeteer.product`](#puppeteerproduct)
 
 > **NOTE** PUPPETEER_* env variables are not accounted for in the [`puppeteer-core`](https://www.npmjs.com/package/puppeteer-core) package.
 
@@ -527,6 +527,7 @@ try {
 
 #### puppeteer.launch([options])
 - `options` <[Object]>  Set of configurable options to set on the browser. Can have the following fields:
+  - `product` <[string]> Which browser to launch. At this time, this is either `chrome` or `firefox`. See also `PUPPETEER_PRODUCT`.
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
   - `headless` <[boolean]> Whether to run browser in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). Defaults to `true` unless the `devtools` option is `true`.
   - `executablePath` <[string]> Path to a browser executable to run instead of the bundled Chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd). **BEWARE**: Puppeteer is only [guaranteed to work](https://github.com/GoogleChrome/puppeteer/#q-why-doesnt-puppeteer-vxxx-work-with-chromium-vyyy) with the bundled Chromium, use at your own risk.
@@ -549,7 +550,7 @@ try {
   - `env` <[Object]> Specify environment variables that will be visible to the browser. Defaults to `process.env`.
   - `devtools` <[boolean]> Whether to auto-open a DevTools panel for each tab. If this option is `true`, the `headless` option will be set `false`.
   - `pipe` <[boolean]> Connects to the browser over a pipe instead of a WebSocket. Defaults to `false`.
-  - `extraPrefs` <[Object]> Additional [preferences](https://developer.mozilla.org/en-US/docs/Mozilla/Preferences/Preference_reference) that can be passed to Firefox (see `PUPPETEER_PRODUCT`)
+  - `extraPrefsFirefox` <[Object]> Additional [preferences](https://developer.mozilla.org/en-US/docs/Mozilla/Preferences/Preference_reference) that can be passed to Firefox (see `PUPPETEER_PRODUCT`)
 - returns: <[Promise]<[Browser]>> Promise which resolves to browser instance.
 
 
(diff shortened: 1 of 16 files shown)


mjzffr:
@mathiasbynens Hi! This is ready for review. My CLA is in the works.

I looked at the results of the checks and it seems the failures are known to be intermittent... Let me know.

mjzffr:
@googlebot I signed it!

mathiasbynens:
Yay, thanks for sending this over! Just as a heads-up, I won’t be able to review this as soon as I’d like, since I’m traveling for ChromeDevSummit and BlinkOn this week. I’ll get back to y’all as soon as I can.

test/puppeteer.spec.js
```diff
@@ -28,31 +28,40 @@ module.exports.addTests = ({testRunner, product, puppeteerPath}) => {
...
+  }, extraLaunchOptions);
+
 
   if (defaultBrowserOptions.executablePath) {
     console.warn(`${YELLOW_COLOR}WARN: running ${product} tests with ${defaultBrowserOptions.executablePath}${RESET_COLOR}`);
   } else {
-    // Make sure the `npm install` was run after the chromium roll.
-    if (!fs.existsSync(puppeteer.executablePath()))
-      throw new Error(`Browser is not downloaded. Run 'npm install' and try to re-run tests`);
+    const path = puppeteer.executablePath();
```
mathiasbynens: Let’s use a different identifier here so as not to shadow `path` (so that it `path` remains `require('path')` throughout this entire file). How about `executablePath`?

lib/helper.js
```diff
@@ -137,7 +137,7 @@ class Helper {
   static removeEventListeners(listeners) {
     for (const listener of listeners)
       listener.emitter.removeListener(listener.eventName, listener.handler);
-    listeners.splice(0, listeners.length);
+    listeners.length = 0;
```
mathiasbynens, line 140: Ooh, drive-by simplification! I dig it 👍🏻

test/puppeteer.spec.js
```diff
@@ -28,31 +28,40 @@ module.exports.addTests = ({testRunner, product, puppeteerPath}) => {
...
+  const JUGGLER = product === 'Juggler';
 
   const puppeteer = require(puppeteerPath);
 
   const headless = (process.env.HEADLESS || 'true').trim().toLowerCase() === 'true';
   const slowMo = parseInt((process.env.SLOW_MO || '0').trim(), 10);
+  let extraLaunchOptions = {};
+  try {
+    extraLaunchOptions = JSON.parse(process.env.EXTRA_LAUNCH_OPTIONS || '{}');
+  } catch (e) {
```
mathiasbynens: nit: `e` → `error` or `exception`

lib/Puppeteer.js
```diff
@@ -18,6 +18,8 @@ const BrowserFetcher = require('./BrowserFetcher');
 const Errors = require('./Errors');
 const DeviceDescriptors = require('./DeviceDescriptors');
 
+const PRODUCT = process.env.PUPPETEER_PRODUCT || process.env.npm_config_puppeteer_product || process.env.npm_package_config_puppeteer_product;
```
mathiasbynens: Let’s move these `env` variables elsewhere — `puppeteer-core` must remain agnostic, and only `lib/Launcher.js` and `install.js` should consider them. `lib/Launcher.js` seems like a good fit. WDYT?

lib/Launcher.js
```diff
@@ -294,49 +342,193 @@ class Launcher {
...
+    return firefoxArguments;
   }
 
+  /**
+   * @param {!Object=} extraPrefs
+   * @return {!Promise<string>}
+   */
+  async _createProfile(extraPrefs) {
+    const profilePath = await mkdtempAsync(path.join(os.tmpdir(), 'puppeteer_dev_firefox_profile-'));
+    const prefs_js = [];
```
mathiasbynens: coding style nit: `prefs_js` → `prefsJs`

same for `user_js`

lib/Launcher.js
```diff
@@ -294,49 +342,193 @@ class Launcher {
...
+ * @param {!Puppeteer.ChildProcess} browserProcess
  * @param {number} timeout
  * @param {string} preferredRevision
  * @return {!Promise<string>}
  */
-function waitForWSEndpoint(chromeProcess, timeout, preferredRevision) {
+function waitForWSEndpoint(browserProcess, timeout, preferredRevision) {
   return new Promise((resolve, reject) => {
-    const rl = readline.createInterface({ input: chromeProcess.stderr });
+    const rl = readline.createInterface({ input: browserProcess.stderr});
```
mathiasbynens: nit: space before `}`

mathiasbynens (changes requested):
I left some initial comments and questions. I want to do a more thorough second pass after BlinkOn, but figured this should get the conversation started.

Thanks again for your work on this!

lib/Puppeteer.js
mjzffr: Yes, there may be a better place for this.

The main motivation here is to maintain some sort of privileged state about which browser we're talking to. This is both for Puppeteer API consumers and the Puppeteer implementation itself, since both internally and externally some browser-specific code may be needed from time to time.

The same holds true even in the case of `puppeteer-core`: the library doesn't download any browser binary for us, but which browser is it ultimately talking to? Or am I mistaken?

Given that, do you have any suggestions?

mjzffr: ^ @mathiasbynens 

Mentioning you in case this question got lost in a sea of notifications. :)

docs/api.md
```diff
@@ -547,6 +549,7 @@ try {
   - `env` <[Object]> Specify environment variables that will be visible to the browser. Defaults to `process.env`.
   - `devtools` <[boolean]> Whether to auto-open a DevTools panel for each tab. If this option is `true`, the `headless` option will be set `false`.
   - `pipe` <[boolean]> Connects to the browser over a pipe instead of a WebSocket. Defaults to `false`.
+  - `extraPrefs` <[Object]> Additional [preferences](https://developer.mozilla.org/en-US/docs/Mozilla/Preferences/Preference_reference) that can be passed to Firefox (see `PUPPETEER_PRODUCT`)
```
mathiasbynens: If we're gonna add a Firefox-specific option, it should probably have Firefox in its name, no?

lib/Puppeteer.js
mathiasbynens: The problem is that `puppeteer-core` might be run outside of Node.js, where `process.*` is not available. In the Launcher, we wrap any usage of `process.env` in `if (!this._isPuppeteerCore) { ... }` for that reason.

lib/Launcher.js
```diff
@@ -412,6 +604,52 @@ function getWSEndpoint(browserURL) {
...
+}
+
+/**
+ * @param {string} product
+ * @param {string} projectRoot
+ * @param {string} preferredRevision
+ * @param {boolean} isPuppeteerCore
+ * @return {ChromeLauncher|FirefoxLauncher}
+ */
+function Launcher(product, projectRoot, preferredRevision, isPuppeteerCore) {
```
mathiasbynens: Let's rename this `launch` now that it's no longer a class

lib/Puppeteer.js
mathiasbynens: Maybe I'm misunderstanding. What problems do you see with keeping the logic local to `lib/Launcher.js`?

test/puppeteer.spec.js
mjzffr: Done.

docs/api.md
mjzffr: Sure. Done.

lib/Puppeteer.js
mjzffr: I think I've addressed both our concerns with the latest change. :) The `PUPPETEER_PRODUCT` stuff is only used if we're not running `puppeteer-core`, and I changed Puppeteer to lazily load an appropriate Launcher based on a `product` option, if available, falling back on environment variables.

lib/Launcher.js
mjzffr: I'm happy to, but it implies a load of type annotation changes (e.g. `Launcher.BrowserOptions`), and I'm not sure that's what you want. 

I left it as `Launcher`, but do let me know if you still want `launcher`.

mathiasbynens (approved):

hanselfmu (approved):
LGTM.

rodoabad:
Have the documents been updated on how to use this new feature? i.e. do you still need `puppeteer-firefox`?

chrisLovesCode:
> Have the documents been updated on how to use this new feature? i.e. do you still need `puppeteer-firefox`?

`puppeteer.launch({product: 'firefox'})`
but it's classified as experimental.
Have a look at the FAQ here: https://github.com/puppeteer/puppeteer#faq

rodoabad:
@chrisLovesCode it doesn't say if `puppeteer` downloads firefox like how it downloads chromium or not. If we still need `puppeteer-firefox` then that's fine. Just wish the documentation was clear on that apart from explaining the new API.

chrisLovesCode:
okay , i got your point! ..yes i agree, would be nice to get some informations, currently i  have not tried out to use firefox, but would like to do it soon, since i will write e2e tests with puppeteer

mjzffr:
@rodoabad puppeteer doesn't download Firefox. That's tracked in https://github.com/puppeteer/puppeteer/issues/5151 

Totally agree that this should be clearer from the docs; I expect that will get better as progress is made on Firefox support.

Issue #3083, opened 2018-08-13
feature: expose page.on('networkidle0') and page.on('networkidle2') events


aslushnikov:
This has been requested quite a few times on the bugtracker already.

As a workaround, network idleness [can be emulated](https://github.com/GoogleChrome/puppeteer/issues/1353#issuecomment-356561654) on the puppeteer-side.

paulirish:
A few notes for anyone else impatient for these first-class events.

The _**can be emulated**_ link above is useful if you need different idleness parameters (# of requests, ms of time).

Alternatively, if you want a basic `page.on('networkidle0', fn)` event, a hacky solution is to drop down to the protocol:

```js
page._client.on('Page.lifecycleEvent', e => {
	// networkidle0 => 'networkIdle', networkidle2 => 'networkAlmostIdle'
    if (e.name === 'networkIdle') fn()
});
```
Standard disclaimers about using the protocol directly (_this may break unexpectedly_) apply.

aslushnikov:
@paulirish  the lifecycle events are [*per frame*](https://vanilla.aslushnikov.com/#Page.lifecycleEvent), so the hacky way doesn't really work

rob4629:
Any updates on this?

tjenkinson:
I opened [this PR](https://github.com/GoogleChrome/puppeteer/pull/5140) which should help with this.

sudarsunperu:
@aslushnikov I'm still looking for `waitForNetworkIdle` for an SPA application. Do we have to use the network idle watcher from https://github.com/puppeteer/puppeteer/issues/1353#issuecomment-356561654 ? are there any plans to get this native in Puppeteer?

PR #5140, opened 2019-11-09
feat(api): implement `Page.waitForNetworkIdle()`

tjenkinson:
Which will wait for there to be no network requests in progress during the `idleTime` before resolving.

One case where this is useful is if you want to take a screenshot after adding a css class that results in more assets being downloaded (background images etc).

refs #3083

## TODO

- [x] The tests currently use a public url, because for some reason when using "/digits/*.png" the `RequestFinished`/`RequestFailed` events aren't fired, and the request isn't removed from `_requestIdToRequest`, even though the `fetch()` call resolves. Any idea why? Is this a chromium bug?
- [ ] Is firefox support needed too?


docs/api.md
```diff
@@ -2089,6 +2090,17 @@ const [response] = await Promise.all([
...
 Shortcut for [page.mainFrame().waitForNavigation(options)](#framewaitfornavigationoptions).
 
+#### page.waitForNetworkIdle([options])
+- `options` <[Object]> Optional waiting parameters
+  - `timeout` <[number]> Maximum wait time in milliseconds, defaults to 30 seconds, pass `0` to disable the timeout. The default value can be changed by using the [page.setDefaultTimeout(timeout)](#pagesetdefaulttimeouttimeout) method.
+  - `idleTime` <[number]> How long to wait for no network requests in milliseconds, defaults to 500 milliseconds.
+- returns: <[Promise]<void>> Promise which resolves when network is idle.
+
+```js
+page.evaluate(() => fetch('some-url'));
```
abratnap, line 2857: Why do we need to use `fetch` here? 
Can't this work with `page.goto` like below?

```
[response] = await Promise.all([
            await page.goto(url, {'timeout': timeOut, 'waitUntil': 'load']}),
            await page.waitForNetworkIdle({idleTime:1000, timeout:30000})
 ]);
```

I did some testing with some URLs and it's not working as expected.

tjenkinson: I was trying to come up with a simple example. One case though is you update the css for something that results in downloading an image, and then you want to wait for the image to be downloaded before you continue and take a screenshot 

What wasn’t working as expected?

stevezau:
any status on this?

davispuh:
I'm also interested in this. Super useful feature when want to do some action after everything has finished loading.

davispuh:
Just found this https://github.com/jtassin/pending-xhr-puppeteer looks exactly what's needed.

alvis:
I'm trying to implement something similar and also realise that sometimes the `RequestFinished`/`RequestFailed` events aren't fired. It seems like to be that some `document` type requests get cancelled before it's fulfilled and in this case no event is emitted.

jcayzac (approved):

jschfflr (changes requested):
@tjenkinson Thanks for your contribution! Could you resolve the merge conflicts and verify if local urls might be working now? Happy to get this landed :)

jschfflr:
Thanks for updating the commit!
It does indeed look like a memory leak. But calling `_forgetRequest` in `_onResponseReceived` might cause problems because there's also the `ResponseReceivedExtraInfo` event that might arrive before or after `ResponseReceived`.
Can you file an issue for that and assign it to me as there's currently some work being done to improve support for `ResponseReceivedExtraInfo` to unblock rolling the latest Chromium.

jschfflr (approved):

jschfflr:
@tjenkinson I just saw this test flake on the mac bot: 
https://github.com/puppeteer/puppeteer/runs/3585119647?check_suite_focus=true

Could you take a look into why that happened and address the issue? Thanks!

jschfflr:
Yeah I think the bots are not very powerful so timing might be problematic in some places. But it looks like the fix solved it :)

PR #5167, opened 2019-11-15
feat: Improve error reporting on AArch64

### LINKED ISSUE #5147: Architecture check to use local chromium binary for aarch64
### Steps to reproduce

* Puppeteer version: 2.0.0
* Platform / OS version: Ubuntu:18.04
* URLs (if applicable): None 
* Node.js version: v8.16.0

1. Git clone the puppeteer package on your environment.
2. Install and test the puppeteer package on your environment.
3. Use command " npm install-test " for the same.

I am ready with a PR which provides the following features:

- To provide an architecture check for aarch64 to check if chromium binary is present or not in the local environment.

- If the chromium binary is present then the by default downloading of the chrome binary at location .local-chromium should not happen and the already present binary in the local environment at /usr/bin/chromium-browser should be used.

- If the chromium binary is not present in the environment then the error message should come on the console that the user needs to install chromium binary manually.

**What happens instead?**
The chromium binary which gets downloaded with puppeteer doesn't support aarch64 hence puppeteer doesn't work on aarch64 environment.

**Note:** If you agree with the above changes I can propose a PR with the above mentioned changes, this PR will help the aarch64 users to use puppeteer and work without any change for the other users. Also if you have any further suggestions too please do share I will work on including them as well in the PR.

ossdev07:
This patch provides architecture check for aarch64 to use local chromium binary and works as usual with other architectures.

Fixes #5147

### DIFF
--- a/src/BrowserFetcher.ts
+++ b/src/BrowserFetcher.ts
@@ -101,6 +101,16 @@ function downloadURL(
   return url;
 }
 
+function handleArm64() {
+  fs.stat('/usr/bin/chromium-browser', function (err, stats) {
+    if (stats === undefined) {
+      console.error(`The chromium binary is not available for arm64: `);
+      console.error(`If you are on Ubuntu, you can install with: `);
+      console.error(`\n apt-get install chromium-browser\n`);
+      throw new Error();
+    }
+  });
+}
 const readdirAsync = helper.promisify(fs.readdir.bind(fs));
 const mkdirAsync = helper.promisify(fs.mkdir.bind(fs));
 const unlinkAsync = helper.promisify(fs.unlink.bind(fs));
@@ -219,15 +229,20 @@ export class BrowserFetcher {
     if (await existsAsync(outputPath)) return this.revisionInfo(revision);
     if (!(await existsAsync(this._downloadsFolder)))
       await mkdirAsync(this._downloadsFolder);
-    try {
-      await downloadFile(url, archivePath, progressCallback);
-      await install(archivePath, outputPath);
-    } finally {
-      if (await existsAsync(archivePath)) await unlinkAsync(archivePath);
+    if (os.arch() === 'arm64') {
+      await handleArm64();
+      return;
+    } else {
+      try {
+        await downloadFile(url, archivePath, progressCallback);
+        await install(archivePath, outputPath);
+      } finally {
+        if (await existsAsync(archivePath)) await unlinkAsync(archivePath);
+      }
+      const revisionInfo = this.revisionInfo(revision);
+      if (revisionInfo) await chmodAsync(revisionInfo.executablePath, 0o755);
+      return revisionInfo;
     }
-    const revisionInfo = this.revisionInfo(revision);
-    if (revisionInfo) await chmodAsync(revisionInfo.executablePath, 0o755);
-    return revisionInfo;
   }
 
   async localRevisions(): Promise<string[]> {
--- a/src/BrowserFetcher.ts
+++ b/src/BrowserFetcher.ts
@@ -232,16 +232,16 @@ export class BrowserFetcher {
     if (os.arch() === 'arm64') {
       await handleArm64();
       return;
-    } else {
-      try {
-        await downloadFile(url, archivePath, progressCallback);
-        await install(archivePath, outputPath);
-      } finally {
-        if (await existsAsync(archivePath)) await unlinkAsync(archivePath);
-      }
-      const revisionInfo = this.revisionInfo(revision);
-      if (revisionInfo) await chmodAsync(revisionInfo.executablePath, 0o755);
-      return revisionInfo;
+    }
+    try {
+      await downloadFile(url, archivePath, progressCallback);
+      await install(archivePath, outputPath);
+    } finally {
+      if (await existsAsync(archivePath)) await unlinkAsync(archivePath);
+    }
+    const revisionInfo = this.revisionInfo(revision);
+    if (revisionInfo) await chmodAsync(revisionInfo.executablePath, 0o755);
+    return revisionInfo;
     }
   }
 
--- a/src/BrowserFetcher.ts
+++ b/src/BrowserFetcher.ts
@@ -230,7 +230,7 @@ export class BrowserFetcher {
     if (!(await existsAsync(this._downloadsFolder)))
       await mkdirAsync(this._downloadsFolder);
     if (os.arch() === 'arm64') {
-      await handleArm64();
+      handleArm64();
       return;
     }
     try {
--- a/src/BrowserFetcher.ts
+++ b/src/BrowserFetcher.ts
@@ -242,7 +242,6 @@ export class BrowserFetcher {
     const revisionInfo = this.revisionInfo(revision);
     if (revisionInfo) await chmodAsync(revisionInfo.executablePath, 0o755);
     return revisionInfo;
-    }
   }
 
   async localRevisions(): Promise<string[]> {
--- a/experimental/puppeteer-firefox/lib/BrowserFetcher.js
+++ b/experimental/puppeteer-firefox/lib/BrowserFetcher.js
@@ -158,14 +158,32 @@ class BrowserFetcher {
     if (!(await existsAsync(this._downloadsFolder)))
       await mkdirAsync(this._downloadsFolder);
     try {
-      await downloadFile(url, zipPath, progressCallback);
-      await extractZip(zipPath, folderPath);
+      if (os.arch() === 'arm64'){
+        fs.stat('/usr/bin/chromium-browser', function(err, stats) {
+          if (stats === undefined){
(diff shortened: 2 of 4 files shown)


ossdev07:
Ping @mathiasbynens, Please can you have a look at this PR. I will be happy to update the patch if you have any suggestions.

lib/BrowserFetcher.js
```diff
@@ -137,14 +137,32 @@ class BrowserFetcher {
     if (!(await existsAsync(this._downloadsFolder)))
       await mkdirAsync(this._downloadsFolder);
     try {
-      await downloadFile(url, zipPath, progressCallback);
-      await extractZip(zipPath, folderPath);
+      if (os.arch() === 'arm64'){
+        fs.stat('/usr/bin/chromium-browser', function(err, stats) {
+          if (stats === undefined){
+            console.error('Chromium Binary is not available for aarch64, Download it manually.');
+            const result = require('child_process').execSync('cat /etc/lsb-release | grep -in "DISTRIB_ID" | cut -f 2 -s -d =').toString().trim();
```
rhenwood-arm: /etc/lsb-release is not part of my Ubuntu default OS. I've looked around and it isn't obvious how you would reliability test for the OS distro name in a standard way. It may be simpler to leave out this check, and just have the console.error lines below:

"If you are on Ubuntu, you can install with:"...

rhenwood-arm (changes requested):
I have tested this patch on my aarch64 server and it works as expected.

I suggest a change to the title as this patch doesn't strictly speaking add support for AArch64 - more like it improves error reporting on AArch64. So I suggest a title like:

feat: Improve error reporting on AArch64.

lib/BrowserFetcher.js
ossdev07: @rhenwood-arm As per your suggestion, I have modified the title and removed the check for ubuntu.

rhenwood-arm:
Thanks @ossdev07 : this looks good to me. I would appreciate feedback from the community.

ossdev07:
Ping @christian-bromann @jackfranklin I have rebased the PR again, Please share your thoughts on this. Please do revert if anything else is required to be done from my end.

jackfranklin:
@ossdev07 thanks for this PR! The changes seem OK to me but I'd like @mathiasbynens to have a look.

src/BrowserFetcher.ts
```diff
@@ -220,14 +220,29 @@ export class BrowserFetcher {
...
-      await install(archivePath, outputPath);
+      if (os.arch() === 'arm64') {
+        fs.stat('/usr/bin/chromium-browser', function (err, stats) {
+          if (stats === undefined) {
+            console.error(` The chromium binary is not available for arm64: `);
+            console.error(` If you are on Ubuntu,You can install with: `);
+            console.error(` \n apt-get install chromium-browser\n`);
+            throw new Error();
+          }
+        });
```
jackfranklin: For the rest of this file we use `async/await` so I'd rather use that here.

```js
const statAsync = helper.promisify(fs.stat)

const stats = await statAsync('...')
```

src/BrowserFetcher.ts
```diff
@@ -220,14 +220,29 @@ export class BrowserFetcher {
...
       await mkdirAsync(this._downloadsFolder);
     try {
-      await downloadFile(url, archivePath, progressCallback);
-      await install(archivePath, outputPath);
+      if (os.arch() === 'arm64') {
+        fs.stat('/usr/bin/chromium-browser', function (err, stats) {
+          if (stats === undefined) {
+            console.error(` The chromium binary is not available for arm64: `);
+            console.error(` If you are on Ubuntu,You can install with: `);
+            console.error(` \n apt-get install chromium-browser\n`);
```
jackfranklin: please could we tidy these up a bit? They all have a space at the beginning that I think can go.

src/BrowserFetcher.ts
```diff
@@ -220,14 +220,29 @@ export class BrowserFetcher {
     if (!(await existsAsync(this._downloadsFolder)))
       await mkdirAsync(this._downloadsFolder);
     try {
-      await downloadFile(url, archivePath, progressCallback);
-      await install(archivePath, outputPath);
+      if (os.arch() === 'arm64') {
+        fs.stat('/usr/bin/chromium-browser', function (err, stats) {
+          if (stats === undefined) {
+            console.error(` The chromium binary is not available for arm64: `);
+            console.error(` If you are on Ubuntu,You can install with: `);
```
jackfranklin: `If you are on Ubuntu, you can install with: `

src/BrowserFetcher.ts
```diff
@@ -220,14 +220,29 @@ export class BrowserFetcher {
...
+        if (await existsAsync(archivePath)) await unlinkAsync(archivePath);
     }
     const revisionInfo = this.revisionInfo(revision);
-    if (revisionInfo) await chmodAsync(revisionInfo.executablePath, 0o755);
-    return revisionInfo;
+    if (revisionInfo) {
+      if (os.arch() !== 'arm64')
+        await chmodAsync(revisionInfo.executablePath, 0o755);
+      return revisionInfo;
+    }
```
jackfranklin, line 235: I'm a bit concerned about all the different conditionals that now exist around the `os.arch()` check. I wonder if there's a nicer way to just completely bypass any of this if we're on `arm64` ?

ossdev07:
@jackfranklin I have incorporated the changes you suggested, please do revert if something else is required to be done.

jackfranklin:
Thanks - have asked @mathiasbynens to have a look.

src/BrowserFetcher.ts
ossdev07: @jackfranklin I will be happy to modify the code more please do revert and provide your suggestions on to what do you exactly want to replace ```os.arch()``` with.

ossdev07:
Ping @mathiasbynens could you please have a look at this PR.

src/BrowserFetcher.ts
mathiasbynens: Let's indeed keep the arm64-specific logic in a single place. Something like this:

```js
if (os.arch() === 'arm64') {
  await handleArm64();
  return;
}
// existing code for non-arm64 goes here...
```

ossdev07: @mathiasbynens incorporated the changes you suggested, please do revert if something else is required to be done.

src/BrowserFetcher.ts
```diff
@@ -219,15 +229,20 @@ export class BrowserFetcher {
...
+    } else {
+      try {
+        await downloadFile(url, archivePath, progressCallback);
+        await install(archivePath, outputPath);
+      } finally {
+        if (await existsAsync(archivePath)) await unlinkAsync(archivePath);
+      }
+      const revisionInfo = this.revisionInfo(revision);
+      if (revisionInfo) await chmodAsync(revisionInfo.executablePath, 0o755);
+      return revisionInfo;
```
mathiasbynens: No `else` after `return`:

```suggestion
    if (os.arch() === 'arm64') {
      await handleArm64();
      return;
    }
    try {
      await downloadFile(url, archivePath, progressCallback);
      await install(archivePath, outputPath);
    } finally {
      if (await existsAsync(archivePath)) await unlinkAsync(archivePath);
    }
    const revisionInfo = this.revisionInfo(revision);
    if (revisionInfo) await chmodAsync(revisionInfo.executablePath, 0o755);
    return revisionInfo;
```

src/BrowserFetcher.ts
```diff
@@ -219,6 +229,10 @@ export class BrowserFetcher {
     if (await existsAsync(outputPath)) return this.revisionInfo(revision);
     if (!(await existsAsync(this._downloadsFolder)))
       await mkdirAsync(this._downloadsFolder);
+    if (os.arch() === 'arm64') {
+      await handleArm64();
```
mathiasbynens: No need to `await` since it doesn't return a promise:

```suggestion
      handleArm64();
```

src/BrowserFetcher.ts
```diff
@@ -228,6 +242,7 @@ export class BrowserFetcher {
     const revisionInfo = this.revisionInfo(revision);
     if (revisionInfo) await chmodAsync(revisionInfo.executablePath, 0o755);
     return revisionInfo;
+    }
```
mathiasbynens:
```suggestion
```

mathiasbynens (approved):

mathiasbynens:
Jack, does this LGTY too now?

ossdev07:
Ping @jackfranklin could you share your thoughts on this.

jackfranklin (approved):

jackfranklin:
So sorry for the delay, thanks for the contribution!

ossdev07:
Thanks @jackfranklin @mathiasbynens for accepting this PR.

Issue #5168, opened 2019-11-15
Improve typescript accuracy


connorjclark:
The intended way to use puppeteer + typescript today is by installing [`@types/puppeteer`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/puppeteer/index.d.ts). However, there are some gaps in these types:

* (big) `CDPSession` is ignorant of the protocol commands / events and return types. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9f57df89f30bacee1d46dfbf40a08132b4b059ef/types/puppeteer/index.d.ts#L2151
* (nit) `Page.emulate(emulateOptions)` - types claim that both `viewport` and `userAgent` are optional, but actually both are required https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9f57df89f30bacee1d46dfbf40a08132b4b059ef/types/puppeteer/index.d.ts#L519
* I use very little of the API surface, there are likely more.

These discrepancies can be avoided if the type definitions are generated from the source directly.

# Approach 1: Generate declaration files from JS

This seemed like the ideal approach - generate a `.d.ts` for every module and publish on npm. Users could use the types like this:

```js
const Browser = require('puppeteer/lib/Browser');

interface MyInterface {
  browser: Browser;
}
```

Ultimately, this didn't work, I think due to bugs within TS. Here's what would need to be done.

## Upgrade to TS 3.7

3.7 added support of `--declaration` with `--allowJs`. https://github.com/microsoft/TypeScript/issues/7546#issuecomment-535699638

## tsc errors for `events`

This pattern trips up TS when generated the declaration file:

```js
const EventEmitter = require('events');

class Blah extends EventEmitter {...}
```

The error manifests like this:

![image](https://user-images.githubusercontent.com/4071474/68933495-9e6a6000-0749-11ea-8514-8443973b8b55.png)

Using `const EventEmitter = require('events').EventEmitter` fixes this. Yeah, that works - it was [originally the only way](https://stackoverflow.com/questions/52524715/in-node-js-why-was-eventemitter-included-in-events-as-a-circular-reference) to import `EventEmitter`, and has stuck around. TS doesn't error b/c `@types/node` references a non-exported name for the default export, but for this named export it doesn't.

But, all modules that import a puppeteer class that extends `EventEmitter` still get the same error:

```
[...]

PR #5174, opened 2019-11-16
fix(types): publish protocol types to npm

connorjclark:
see #5168

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -252,7 +252,7 @@ Releasing to npm consists of the following phases:
     1. On your local machine, pull from [upstream](https://github.com/GoogleChrome/puppeteer) and make sure the last commit is the one just merged.
     2. Run `git status` and make sure there are no untracked files.
         - **WHY**: this is to avoid adding unnecessary files to the npm package.
-    3. Run `yarn` to make sure the latest `lib/protocol.d.ts` is generated.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -252,8 +252,9 @@ Releasing to npm consists of the following phases:
     1. On your local machine, pull from [upstream](https://github.com/GoogleChrome/puppeteer) and make sure the last commit is the one just merged.
     2. Run `git status` and make sure there are no untracked files.
         - **WHY**: this is to avoid adding unnecessary files to the npm package.
-    3. Run [`npx pkgfiles`](https://www.npmjs.com/package/pkgfiles) to make sure you don't publish anything unnecessary.
-    4. Run `npm publish`. This publishes the `puppeteer` package.
+    3. Run `yarn` to make sure the latest `lib/protocol.d.ts` is generated.
```
mathiasbynens: Is there an easy way to do this that does not require installing`yarn`?

connorjclark: This should work for `npm install` too. It's done in an install hook.

Could replace with `npm run install / yarn install`.

mathiasbynens: Let’s do that then — it’s more consistent with the other steps:

```suggestion
    3. Run `npm install` to make sure the latest `lib/protocol.d.ts` is generated.
```

mathiasbynens (approved):

mathiasbynens:
Thanks!

PR #5184, opened 2019-11-21
fix(frame): Fix Frame.waitFor's XPath pattern detection

xprudhomme:
Up to now, only strings starting with '`//`' are considered as to XPath selectors. 

Unfortunately, this is too restricting and leads to errors especially when asking the page to wait for XPath selectors not starting with '`//`'. 

If the XPath is something like '`(//div)[4]/@itemprop`' then it is passed to the waitForSelector method and eventually document.querySelector, which in turn throws an error because it does not recognize it as a valid CSS selector.

Because of this, I had to implement something like this in my code:

```javascript
await isXPathSelector(selector)
            ? page.waitForXPath(selector)
            : page.waitForSelector(selector);
```

This fix allows the use of valid XPath selectors starting with parenthesis and to be recognized as such by the Frame's waitFor method.


mathiasbynens:
This needs tests.

xprudhomme:
It's my first PR here. Never done any test for this before.
Any help appreciated for implementing required tests.
Thanks :)

mathiasbynens:
Look for the existing `waitForXPath` tests and add some examples that failed without your patch, but now work.

xprudhomme:
Ok thanks.

I've changed the xPathPattern to allow for the use of XPath starting with not only "//" but also with an opening parenthesis.

In my last commit, I've also added a new test accordingly.

jschfflr (approved):
@xprudhomme Thanks for your contribution! Please resolve the merge conflicts and I'm happy to get this landed!

jschfflr:
@googlebot I consent.

PR #5196, opened 2019-11-26
fix: prepare jsHandle.uploadFile for CDP Page.handleFileChooser removal

mathiasbynens:
https://chromium-review.googlesource.com/c/chromium/src/+/1935410 removes `Page.handleFileChooser` from the CDP. This patch prepares `jsHandle.uploadFile` for that change while preserving existing semantics as much as possible.

### DIFF
--- a/lib/JSHandle.js
+++ b/lib/JSHandle.js
@@ -14,11 +14,7 @@
  * limitations under the License.
  */
 
-const fs = require('fs');
-const path = require('path');
-
 const {helper, assert, debugError} = require('./helper');
-const readFileAsync = helper.promisify(fs.readFile);
 
 function createJSHandle(context, remoteObject) {
   const frame = context.frame();
@@ -305,8 +301,8 @@ class ElementHandle extends JSHandle {
         if (option.selected && !element.multiple)
           break;
       }
-      element.dispatchEvent(new Event('input', { 'bubbles': true }));
-      element.dispatchEvent(new Event('change', { 'bubbles': true }));
+      element.dispatchEvent(new Event('input', { bubbles: true }));
+      element.dispatchEvent(new Event('change', { bubbles: true }));
       return options.filter(option => option.selected).map(option => option.value);
     }, values);
   }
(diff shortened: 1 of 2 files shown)


lib/JSHandle.js
```diff
@@ -312,9 +315,21 @@ class ElementHandle extends JSHandle {
    * @param {!Array<string>} filePaths
    */
   async uploadFile(...filePaths) {
-    const files = filePaths.map(filePath => path.resolve(filePath));
-    const objectId = this._remoteObject.objectId;
-    await this._client.send('DOM.setFileInputFiles', { objectId, files });
+    const promises = filePaths.map(filePath => readFileAsync(filePath, 'utf8'));
+    const buffers = await Promise.all(promises);
+    const fileContents = buffers.map(buffer => buffer.toString());
```
pavelfeldman: `buffer.toString('base64')` to support binary data.

lib/JSHandle.js
```diff
@@ -312,9 +315,21 @@ class ElementHandle extends JSHandle {
...
+    filePaths = filePaths.map(filePath => path.basename(filePath));
+    await this.evaluateHandle((element, filePaths, fileContents) => {
+      const dt = new DataTransfer();
+      for (let i = 0; i < filePaths.length; i++) {
+        const name = filePaths[i];
+        const content = fileContents[i];
+        const file = new File([content], name);
```
pavelfeldman:
```
const result = await fetch(`data:${files[i].mimeType};base64,${files[i].content}`);
const file = new File([await result.blob()], files[i].name);
```

mathiasbynens: Where are you getting `mimeType` from? Are you suggesting we introduce a dependency on `mime-types`, or am I missing some trick?

hanselfmu (commented):
LGTM

hanselfmu (approved):
LGTM

lib/JSHandle.js
```diff
@@ -312,9 +311,35 @@ class ElementHandle extends JSHandle {
...
+    }
+    await this.evaluateHandle(async(element, files) => {
+      const dt = new DataTransfer();
+      for (const item of files) {
+        const response = await fetch(`data:${item.mimeType};base64,${item.content}`);
+        const file = new File([await response.blob()], item.name);
+        dt.items.add(file);
+      }
+      element.files = dt.files;
+      element.dispatchEvent(new Event('input', { bubbles: true }));
```
JoelEinbinder, line 341: Is there any way we can avoid the fake event here?

lib/JSHandle.js
```diff
@@ -312,9 +311,35 @@ class ElementHandle extends JSHandle {
    * @param {!Array<string>} filePaths
    */
   async uploadFile(...filePaths) {
-    const files = filePaths.map(filePath => path.resolve(filePath));
-    const objectId = this._remoteObject.objectId;
-    await this._client.send('DOM.setFileInputFiles', { objectId, files });
```
JoelEinbinder: This method hasn't been removed. I'm curious as to why we go down to JavaScript here.

PR #5200, opened 2019-11-27
docs(api.md): Adding new example to page.$$eval

ffeu:
IMHO it's unclear that the second parameter actually receives the array instead of being executed for each item. Adding this example to maybe make that clear.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1113,6 +1113,11 @@ Examples:
 const divsCounts = await page.$$eval('div', divs => divs.length);
 ```
 
+```js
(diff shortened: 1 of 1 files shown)


mathiasbynens:
> IMHO it's unclear that the second parameter actually receives the array instead of being executed for each item.

Arguably, the `length` example already makes that clear -- or at least, that the callback is being executed on something with a `length` property.

> Adding this example to maybe make that clear.

The new example makes it clear that it's in fact a real array and not a NodeList, so I'll accept this PR for that reason :) Thanks!

mathiasbynens (approved):

PR #5224, opened 2019-12-04
Require https-proxy-agent only when actually needed

ouraios:
`https-proxy-agent` patches core Node methods in `https` for whatever reason, and that breaks unrelated code. In our app we don't use proxy, and the mere fact that `puppeteeer` requires the proxy package, it breaks other code.

See TooTallNate/node-agent-base#35 and sindresorhus/got#951

### DIFF
--- a/lib/BrowserFetcher.js
+++ b/lib/BrowserFetcher.js
@@ -23,8 +23,6 @@ const URL = require('url');
 const {helper, assert} = require('./helper');
 const removeRecursive = require('rimraf');
 // @ts-ignore
-const ProxyAgent = require('https-proxy-agent');
-// @ts-ignore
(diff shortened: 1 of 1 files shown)


ouraios:
@googlebot I signed it!

ouraios:
Hmm it seems like the failing tests are not due to my changes apparently, can someone look at it ?

mathiasbynens:
Thanks for putting together this patch! As you implied, this approach just moves the problem further down the line instead of fully solving it, so I'm closing this in favor of #5243 which does fully solve it! Hope that's okay :)

ouraios:
No problem at all, as long as it gets fixed i'm fine with whatever fix is used 😉

PR #5233, opened 2019-12-07
feat: port to strict typescript; convert to esm

### LINKED ISSUE #5168: Improve typescript accuracy

AviVahl:
closes #5168

This PR ports puppeteer's `lib` to strict TypeScript with ESM syntax. I've created/used a custom script to inline the jsdocs types and then continued manually fixing.

Notes:
- `.d.ts` files are automatically generated. :)
- all tests pass without any changes (no `*.spec.js` file has been changed).
- updated type tests to include all tests of `@types/puppeteer`. They pass. Not sure how to handle `ExpectType` comments (even though correct types are inferred in all cases it is used).
- compatible with user typescript projects not using `esModuleInterop` as well. tested in `"test-types"` script.
- esm migration required some `index.ts` tweaking. It now works/behaves the same as master's commonjs version (from a user's perspective), both in typed and non-typed modes.
- While some fields are prefixed with an underscore (signalling they are private), they were accessed in other classes. These are now marked as `/* @internal */` and `public`, and the `"stripInternal"` field of `tsconfig.json` has been turned on, so they don't get emitted into `.d.ts`
- protocol types are now consumed from official [devtools-protocol](https://github.com/ChromeDevTools/devtools-protocol) npm package.
- ported `fetch-devices` script to typescript (inside the `/scripts` folder).

Pain points:
- when cloning the repo itself and running `npm i` / `yarn`, the `"install"` script executes and downloads local chromium. This poses an issue when the source code is typescript-based, as the script depends on the not-yet-built `BrowserFetcher` class. I've adjusted `install.js` to check if `'./lib'` exists. If it doesn't, it will require ``ts-node/register`` and use '`./src'` directly.
- The generated `.d.ts` files require typescript@3.6+.


AviVahl:
juggler tests appear to be flaky (reproduced locally with master)

SimonSchick:
Not a maintainer but looking at the diff gives me vertigo due to the enourmous amount of changes.
Also was there any effort to stay somewhat compatible with current typings, this has been an issue before when the team tried to ship their typings vs. DTs.

AviVahl:
@SimonSchick yeah, I realize it's big, but most new lines are due to `protocol.ts` being inlined. GitHub's diff is probably not the best tool to review the changes...

There's a lot of work being done (still WIP) to get DefinitelyTyped's tests into `./test/types/*` and ensure they pass. I use `@types/puppeteer` myself in many projects and want/need this compatibility. I'm also trying to get as many jsdocs in as I can. Super useful for auto-completions.

There were several cases where the DT types had `Promise<Buffer>`, while the actual code returns `Promise<Buffer | null>`. This was caught by the `.d.ts` being auto-generated, which is a good thing.

SimonSchick:
I fully expected a TS port to uncover some incorrect types and those are totally cool :)

AviVahl:
last commit made PR diff stats saner :)
also cleared the protocol types concerns by consuming them from official source.

tomrav:
Is anyone planning on looking at this PR? I feel like it would make future contributions much easier.

AviVahl:
Closing due to lack of response.

PR #5289, opened 2020-01-08
[WIP] feat(chromium): roll Chromium to r722269

mathiasbynens:
This corresponds to Chromium 80.0.3987.0.

This roll includes:

- Implement support for the new ARIA `generic` role
  https://chromium-review.googlesource.com/c/chromium/src/+/1872305
- Expose button's children to accessibility tree
  https://chromium-review.googlesource.com/c/chromium/src/+/1845810
- Remove `Page.handleFileChooser` from CDP
  https://chromium-review.googlesource.com/c/chromium/src/+/1935410

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -15,7 +15,6 @@
  */
 
 const fs = require('fs');
-const path = require('path');
 const EventEmitter = require('events');
 const mime = require('mime');
 const {Events} = require('./Events');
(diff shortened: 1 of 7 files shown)


lib/Page.js
```diff
@@ -1386,9 +1379,6 @@ class FileChooser {
   async cancel() {
```
hanselfmu, line 1379: I suppose we left this method "async" to be consistent with accept() and be backward-compatible?

hanselfmu (approved):
LGTM; looks like a clean solution to me 😃

lib/Page.js
mathiasbynens: Indeed, [`FileChooser.cancel`](https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#filechoosercancel) is part of the public API, and should continue to return a promise (or at least a thenable) since code might depend on it.

PR #5291, opened 2020-01-09
Update node image in Dockerfile doc 

boubkerbribri:
Problem with node:10-slim since today

image


boubkerbribri:
@googlebot I signed it!

hanselfmu:
@boubkerbribri I think this is the same issue as addressed in https://github.com/puppeteer/puppeteer/pull/5400. We would like to keep the `:slim` version of the Node.js image because the normal version adds more than just `wget` and `gnupg`. I'm closing this one, but thanks for noticing this.

PR #5295, opened 2020-01-09
Add durableStorage to allowed permissions

mushishi78:
Please tell me if I've got the wrong end of the stick, I'm totally operating under guess work. I'd like to be able to use indexeddb. I've found "durableStorage" [here](https://chromium-review.googlesource.com/c/chromium/src/+/1185877/4/content/browser/devtools/protocol/browser_handler.cc#120).

I can also see it in the [dart puppeteer docs](https://pub.dev/documentation/puppeteer/latest/protocol_browser/PermissionType-class.html) (along with a bunch of others).

So I was hoping it was as simple as just adding to this list 🤞

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1094,7 +1094,7 @@ Creates a new page in the browser context.
   - `'clipboard-read'`
   - `'clipboard-write'`
   - `'payment-handler'`
-  - `'durable-storage'`
+  - `'persistent-storage'`
 - returns: <[Promise]>
 
 ```js
--- a/docs/api.md
+++ b/docs/api.md
@@ -1094,6 +1094,7 @@ Creates a new page in the browser context.
   - `'clipboard-read'`
   - `'clipboard-write'`
   - `'payment-handler'`
+  - `'durable-storage'`
 - returns: <[Promise]>
 
 ```js
--- a/src/common/Browser.ts
+++ b/src/common/Browser.ts
@@ -56,7 +56,7 @@ const WEB_PERMISSION_TO_PROTOCOL_PERMISSION = new Map<
   ['clipboard-read', 'clipboardReadWrite'],
   ['clipboard-write', 'clipboardReadWrite'],
   ['payment-handler', 'paymentHandler'],
-  ['durable-storage', 'durableStorage'],
+  ['persistent-storage', 'durableStorage'],
   ['idle-detection', 'idleDetection'],
   // chrome-specific permissions we have.
   ['midi-sysex', 'midiSysex'],
@@ -80,7 +80,7 @@ export type Permission =
   | 'clipboard-read'
   | 'clipboard-write'
   | 'payment-handler'
-  | 'durable-storage'
+  | 'persistent-storage'
   | 'idle-detection'
   | 'midi-sysex';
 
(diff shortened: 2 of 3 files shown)


mushishi78:
@googlebot I signed it!

jschfflr (changes requested):
@mushishi78 Thanks for your contribution! Could you add the new permission to the documentation and maybe also find a way to add a test for this? Happy to get this landed :)

mushishi78:
@jschfflr added to docs and test. It's a bit low effort copy and paste, but possibly good enough? 🤞

mushishi78:
@jschfflr And now rebased on main

jschfflr (approved):
Looks good to me, thanks!

jschfflr:
> @jschfflr added to docs and test. It's a bit low effort copy and paste, but possibly good enough? 🤞

Looks good enough to me, thanks!

mushishi78:
Seems like the test failed on CI 😞

jschfflr:
It looks like `getPermission` is using `navigator.permissions.query({ name }).then((result) => result.state)` internally and the durable storage thing is not exposed by that. So instead, you should probably test if access to indexeddb is granted with the permission enabled.

jschfflr:
@mushishi78 I just tried to write a test for this usecase:
```js
itFailsFirefox('should grant durable-storage', async () => {
      const { page, server, context } = getTestState();

      function checkIfIndexedDBIsAvailable() {
        const request = window.indexedDB.open("MyTestDatabase", 1);
          request.onsuccess = () => resolve(true);
          request.onerror = (err) => reject(err);
        });
      }

      await context.clearPermissionOverrides();
      expect(await page.evaluate(checkIfIndexedDBIsAvailable)).toBe(false);
      await context.overridePermissions(server.EMPTY_PAGE, ['durable-storage']);
      expect(await page.evaluate(checkIfIndexedDBIsAvailable)).toBe(true);
      await context.clearPermissionOverrides();
    });
```

but it looks like the `durable-storage` permission is not needed to create the indexedDB instance. Could you verify that this unit test actually tests the problem that you ran into?

mushishi78:
Thanks @jschfflr. The problem for me is only when not in headless if that helps. Whenever I debug my tests I have to manually press accept on the permission.

jschfflr:
@mushishi78 Could you share the code that you are running? Or a minimal reproduction?
I tried running `HEADLESS=0 npm run unit -- -g "should grant durable-storage"` but it didn't ask for permission either :/

mushishi78:
That's frustrating. I'm going on holiday until next monday, so won't have time to until sometime next week unfortuneately. Thanks for your help.

jschfflr:
I took a look at this change again and the underlying permissions. It looks like `durableStorage` maps to `persistent-storage` in the permission api. I therefore changed this patch to use `persistent-storage` instead of `durable-storage`. Let me know what you think!

Regarding the `indexedDB` test - the permission is probably only needed if the indexeddb reaches a certain since which is not the case for the test.

> That's frustrating. I'm going on holiday until next monday, so won't have time to until sometime next week unfortuneately. Thanks for your help.

Have a great time off! Let me know if you're happy with the change as it is now and I'll get it merged :)

jschfflr:
@googlebot I consent.

PR #5344, opened 2020-01-27
docs(examples): Update api.md

odykyi:


mathiasbynens:
Thanks for the patch! However, the suggested phrasing is not accurate. These are HTML `class` attribute values, which indeed can be styled using CSS or selected using JavaScript.

PR #5358, opened 2020-01-29
Fix: Add mimeType to the File instance uploadFile()

heikkipora:
Not having the File.type property present breaks fileUpload functionality for implementations that rely on the file mime type being set by the browser.

Was broken in release 2.1.0 by commit https://github.com/puppeteer/puppeteer/commit/6091a34a360b0b8aec4d67f6190f131d7e8e3300

### DIFF
--- a/lib/JSHandle.js
+++ b/lib/JSHandle.js
@@ -336,7 +336,7 @@ class ElementHandle extends JSHandle {
       const dt = new DataTransfer();
       for (const item of files) {
         const response = await fetch(`data:${item.mimeType};base64,${item.content}`);
-        const file = new File([await response.blob()], item.name);
+        const file = new File([await response.blob()], item.name, {type: item.mimeType});
         dt.items.add(file);
       }
       element.files = dt.files;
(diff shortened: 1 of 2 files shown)


mathiasbynens:
Could you please add a test that fails without this patch but succeeds with it? Thanks.

heikkipora:
@googlebot I signed it!

heikkipora:
Updated input.spec.js to cover the new type property and linked the breaking commit to PR description

reedom:
Hi, this patch is what I was about to write by myself to solve an issue on my project.
It looks got paused over a month.
Who's responsible to proceed it?
And/or is there anything I can do for that?

renfeng:
@mathiasbynens Could you help the people out?

mathiasbynens:
@paullewis, could you PTAL in light of #5654?

paullewis:
This is fixed as of 532ae573d29063ffd081ed9ee340d71b25320dec

SimonSchick:
Are you sure this is fixed?
I'm having an issue where when calling `uploadFile` seems to add files to the upload with no mime type.

image

This does not occur when using the browser manually.

paullewis:
Hi Simon, which version is this? The fix isn't in a release version yet, so to check you'd need to `npm install https://github.com/puppeteer/puppeteer` to get what's on master.

SimonSchick:
That's probably why, I assumed the fix was in yesterdays 3.0.0 release 🙂 
It's always confusing when an issue is marked as `closed` but the fix isn't technically released yet 😛

PR #5400, opened 2020-02-08
Bugfix: Docker build failure

zawataki:
The recommended Dockerfile uses `node:10-slim` image as a base,
but the base image does not contain `wget` command anymore.
(About the reason, see https://github.com/nodejs/docker-node/issues/1185)
So fixed the problem.


hanselfmu (approved):
LGTM

zawataki:
Thank you for your review.

PR #5389, opened 2020-02-05
Emit change event on uploadFile calls

phihag:
In puppeteer 1.20.0, the following code emitted an alert:
```
const puppeteer = require('puppeteer');

(async () => {
	const browser = await puppeteer.launch({args: ['--no-sandbox'], headless: false});
	const page = (await browser.pages())[0];

	await page.setContent(`
		<input type=file>
		<script>
		document.querySelector('input').addEventListener('change', () => {
			alert('Uploaded a file');
		});
		</script>
		`);
	const input = await page.waitForSelector('input[type="file"]');
	await input.uploadFile(__filename);
})();
```

This is broken in puppeteer 2.0.0 and 2.1.0.
A bisection yielded 6091a34a360b0b8aec4d67f6190f131d7e8e3300 as the commit that broke this.

Emitting the change event as well seems to fix the problem.

PR #5446, opened 2020-02-24
Dispatch event so listeners would see changed file

### LINKED ISSUE #5420: 2.1.0 and 2.1.1 uploadFile failing
I have this code: (via codeceptjs)
```
    async attachFile(locator, pathToFile) {
      const file = path.join(global.codecept_dir, pathToFile);

      if (!fileExists(file)) {
        throw new Error(`File at ${file} can not be found on local system`);
      }
      const els = await findFields.call(this, locator);
      assertElementExists(els, 'Field');
      await els[0].uploadFile(file);
      return this._waitForAction();
    }
```

In in 2.0.0 this works but in 2.1.0 and 2.1.1 this no longer works.

yurynix:
Closes https://github.com/puppeteer/puppeteer/issues/5420

In a similar way it's done at:
https://github.com/puppeteer/puppeteer/blob/master/lib/JSHandle.js#L305

We should notify listeners the input is changed.

### DIFF
--- a/lib/JSHandle.js
+++ b/lib/JSHandle.js
@@ -341,6 +341,7 @@ class ElementHandle extends JSHandle {
       }
       element.files = dt.files;
       element.dispatchEvent(new Event('input', { bubbles: true }));
+      element.dispatchEvent(new Event('change', { bubbles: true }));
     }, files);
   }
 


yurynix:
@googlebot I signed it!

gziolo:
Just noting that there is a similar patch approved: https://github.com/puppeteer/puppeteer/pull/5389. I hope it's merged and a new version published to npm sooner than later. At the moment, it prevents upgrade to the latest version for projects that have media upload in their specs.

mathiasbynens:
Thanks! Closing in favor of #5389, which was merged.

PR #5365, opened 2020-01-30
chore: drop Node.js v8 support

hanselfmu:
Update Node.js versions to 10.18.1

Fixes #5186

PR #5467, opened 2020-02-28
(feat) Add option to fetch Firefox Nightly

### LINKED ISSUE #5151: (Laucher, BrowserFetcher) Download appropriate Firefox version; determine executable path
(This depends on https://github.com/GoogleChrome/puppeteer/pull/5137)

When users ask Puppeteer to run Firefox, they should have the option to make Puppeteer fetch an appropriate Firefox version.

mjzffr:
Add Firefox support to BrowserFetcher and the install script.
By default, the latest Firefox Nightly is downloaded
directly from archive.mozilla.org (dmg, tar.bz2 and zip)

This also required changes that impact `puppeteer.launch()`
and `puppeteer.executablePath()`

Fixes #5151


README.md
```diff
@@ -37,13 +37,13 @@ npm i puppeteer
 # or "yarn add puppeteer"
 ```
 
-Note: When you install Puppeteer, it downloads a recent version of Chromium (~170MB Mac, ~282MB Linux, ~280MB Win) that is guaranteed to work with the API. To skip the download, see [Environment variables](https://github.com/puppeteer/puppeteer/blob/v2.1.1/docs/api.md#environment-variables).
+Note: When you install Puppeteer, it downloads a recent version of Chromium (~170MB Mac, ~282MB Linux, ~280MB Win) that is guaranteed to work with the API. To skip the download, or to download a different browser, see [Environment variables](https://github.com/puppeteer/puppeteer/blob/v2.1.1/docs/api.md#environment-variables).
```
mathiasbynens, line 40: For the record, we probably don’t want to do this _right now_, but in the future we could work towards `npm install puppeteer` fetching both Chromium and Firefox binaries by default. I like your approach of taking one step at a time 👍🏻

docs/api.md
```diff
@@ -572,17 +575,19 @@ const browser = await puppeteer.launch({
 > See [`this article`](https://www.howtogeek.com/202825/what%E2%80%99s-the-difference-between-chromium-and-chrome/) for a description of the differences between Chromium and Chrome. [`This article`](https://chromium.googlesource.com/chromium/src/+/lkgr/docs/chromium_browser_vs_google_chrome.md) describes some differences for Linux users.
 
 #### puppeteer.product
-- returns: <[string]> returns the name of the browser that is under automation ("chrome" or "firefox")
+- returns: <[string]> returns the name of the browser that is under automation (`chrome` or `firefox`)
```
```suggestion
- returns: <[string]> returns the name of the browser that is under automation (`"chrome"` or `"firefox"`)
```

mathiasbynens:
I’ll take a closer look during the week, but just wanted to drop a note saying I’m **super excited** about this!

docs/api.md
```diff
@@ -391,7 +393,7 @@ If Puppeteer doesn't find them in the environment during the installation step,
...
 - `PUPPETEER_EXECUTABLE_PATH` - specify an executable path to be used in `puppeteer.launch`. See [puppeteer.launch([options])](#puppeteerlaunchoptions) on how the executable path is inferred. **BEWARE**: Puppeteer is only [guaranteed to work](https://github.com/puppeteer/puppeteer/#q-why-doesnt-puppeteer-vxxx-work-with-chromium-vyyy) with the bundled Chromium, use at your own risk.
-- `PUPPETEER_PRODUCT` - specify which browser you'd like Puppeteer to use. Must be one of `chrome` or `firefox`. Setting `product` programmatically in [puppeteer.launch([options])](#puppeteerlaunchoptions) supercedes this environment variable. The product is exposed in [`puppeteer.product`](#puppeteerproduct)
+- `PUPPETEER_PRODUCT` - specify which browser you'd like Puppeteer to use. Must be one of `chrome` or `firefox`. This can also be used during installation to fetch the recommended browser binary. Setting `product` programmatically in [puppeteer.launch([options])](#puppeteerlaunchoptions) supercedes this environment variable. The product is exposed in [`puppeteer.product`](#puppeteerproduct)
```
mathiasbynens: Let's fix the pre-existing typo while we're here: 

```suggestion
- `PUPPETEER_PRODUCT` - specify which browser you'd like Puppeteer to use. Must be one of `chrome` or `firefox`. This can also be used during installation to fetch the recommended browser binary. Setting `product` programmatically in [puppeteer.launch([options])](#puppeteerlaunchoptions) supersedes this environment variable. The product is exposed in [`puppeteer.product`](#puppeteerproduct)
```

docs/api.md
```diff
@@ -633,6 +644,10 @@ The method initiates a GET request to download the revision from the host.
   - `executablePath` <[string]> path to the revision executable
   - `url` <[string]> URL this revision can be downloaded from
   - `local` <[boolean]> whether the revision is locally available on disk
+  - `product` <[string]> one of `chrome` or `firefox`
+
+> **NOTE** Many BrowserFetcher methods, like `remove` and `revisionInfo`
+> are affected the choice of `product`. See [puppeteer.createBrowserFetcher([options])](#puppeteercreatebrowserfetcheroptions).
```
```suggestion
> **NOTE** Many BrowserFetcher methods, like `remove` and `revisionInfo`
> are affected by the choice of `product`. See [puppeteer.createBrowserFetcher([options])](#puppeteercreatebrowserfetcheroptions).
```

install.js
```diff
@@ -20,105 +20,147 @@
...
+};
+
+function download() {
+  const downloadHost = process.env.PUPPETEER_DOWNLOAD_HOST || process.env.npm_config_puppeteer_download_host || process.env.npm_package_config_puppeteer_download_host;
+  const puppeteer = require('./index');
+  const product = process.env.PUPPETEER_PRODUCT || process.env.npm_config_puppeteer_product || process.env.npm_package_config_puppeteer_product || 'chrome';
+  const browserFetcher = puppeteer.createBrowserFetcher({ product, host: downloadHost });
+  let revisionPromise;
+  if (product === 'chrome') {
+    revisionPromise = Promise.resolve(process.env.PUPPETEER_CHROMIUM_REVISION || process.env.npm_config_puppeteer_chromium_revision || process.env.npm_package_config_puppeteer_chromium_revision
```
mathiasbynens: This is getting a bit much, so let's move it to its own function, maybe `async function getRevision`, so we can do something like:

```js
const revision = await getRevision();
fetchBinary(revision);
```

install.js
```diff
@@ -20,105 +20,147 @@
...
+      logPolitely(`Requesting latest Firefox Nightly version from ${host}`);
+      https.get(host + '/', r => {
+        r.on('data', chunk => {
+          data += chunk;
+        });
+        r.on('end', parseVersion);
+      }).on('error', e => reject(e));
+
+      function parseVersion() {
+        const regex = /firefox\-(?<version>\d\d)\..*/gm;
```
mathiasbynens, line 148: Loving the use of named capture groups here!

This will fail on Node.js v8.x.x so we'd want to land #https://github.com/puppeteer/puppeteer/pull/5365 first. @hanselfmu could you please take another look?

install.js
```diff
@@ -20,105 +20,147 @@
...
+        const matches = data.matchAll(regex);
```
mathiasbynens: Unfortunately `matchAll` is not available in Node.js v10, so we'll have to go old-school on this one

test/launcher.spec.js
```diff
@@ -31,30 +31,65 @@ module.exports.addTests = function({testRunner, expect, defaultBrowserOptions, p
...
+        expect(revisionInfo.local).toBe(false);
+        expect(browserFetcher.platform()).toBe('linux');
+        expect(browserFetcher.product()).toBe('firefox');
+        expect(await browserFetcher.canDownload('100000')).toBe(false);
+        expect(await browserFetcher.canDownload(expectedVersion)).toBe(true);
+
+        revisionInfo = await browserFetcher.download(expectedVersion);
+        expect(revisionInfo.local).toBe(true);
+        expect(await readFileAsync(revisionInfo.executablePath, 'utf8')).toBe('FIREFOX LINUX BINARY\n');
+        const expectedPermissions = os.platform() === 'win32' ? 0666 : 0755;
```
mathiasbynens: I realize this is just matching the existing convention throughout this file, but since this is new code: let's not use legacy octal literals, since they are invalid in strict mode / modules (which we might want to move to at some point). We can use explicit octal integer literals instead:

```suggestion
        const expectedPermissions = os.platform() === 'win32' ? 0o666 : 0o755;
```

mathiasbynens (changes requested):
Generally looks good, just some questions and suggestions. @hanselfmu could you PTAL as well, especially w.r.t. dropping Node.js v8?

install.js
hanselfmu: The PR to drop Node.js v8.x support is labeled as a breaking change, which means landing it will be in Puppeteer v3. I think we have two options now:
1. land both of them in v3;
2. temporarily use Node.js-8-compatible code for now, and when v3 lands, update the legacy code. 
@mjzffr what do you think? Do you have other ideas?

mathiasbynens: IMHO major version bumps are no big deal. We shouldn't hold back on publishing v3.

mjzffr:
I'm fine with swapping in node-8-compatible code if you want to delay v3.
Based on Mathias' reply, I'm assuming you'll move forward with v3 soon, but
if that's not the case, just ping me and I'll update the PR. I think I've
addressed all the other review feedback.

install.js
hanselfmu: Sounds good. I'll publish v3 now and include this PR.

hanselfmu (approved):
LGTM

install.js
nylen: @hanselfmu @mathiasbynens What happened with this? This PR doesn't seem to be available in a release version yet.

PR #5546, opened 2020-03-27
Update the latest Chromium version on alpine:edge

jo23sh:
The latest Chromium version on alpine edge is 80 - see https://pkgs.alpinelinux.org/package/edge/community/x86_64/chromium


jo23sh:
@googlebot I signed it!

jo23sh:
This has become obsolete 😄

PR #5556, opened 2020-03-30
Upgrade to TypeScript 3.5

jackfranklin:
TS 3.5 got much stricter on writing changes to objects with varied types
[1] so we have to do a bit of typecasting work to convince TS about the
types of keys and values that we are setting.

Longer term we should think about a better data structure that avoids us
having to jump through some hoops but for now I think this is a
reasonable step to get us onto 3.5.

Same story regarding bindings on `window`, the easiest fix is to cast
window to any for the code that adds to it. I'm sure we can come up with
a more type-safe way of doing this in the future.

[1]: https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#fixes-to-unsound-writes-to-indexed-access-types

### DIFF
--- a/lib/Accessibility.js
+++ b/lib/Accessibility.js
@@ -329,7 +329,9 @@ class AXNode {
       role: this._role
     };
 
-    /** @type {!Array<keyof SerializedAXNode>} */
+    /** @enum {'name'|'value'|'description'|'keyshortcuts'|'roledescription'|'valuetext'} */
+    let UserStringProperties; // eslint-disable-line no-unused-vars
+    /** @type {!Array<UserStringProperties>} */
     const userStringProperties = [
       'name',
       'value',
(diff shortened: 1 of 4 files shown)


lib/Accessibility.js
```diff
@@ -329,7 +329,9 @@ class AXNode {
       role: this._role
     };
 
-    /** @type {!Array<keyof SerializedAXNode>} */
+    /** @enum {'name'|'value'|'description'|'keyshortcuts'|'roledescription'|'valuetext'} */
+    let UserStringProperties; // eslint-disable-line no-unused-vars
```
mathiasbynens, line 333: Is there a way to do this without the need for the `let` and the `eslint` comment? (I'm no TypeScript expert...)

mathiasbynens (commented):
Nit: please change the commit message to follow https://github.com/puppeteer/puppeteer/blob/master/CONTRIBUTING.md#commit-messages

@TimvdLippe Do you wanna take a look?

package.json
```diff
@@ -61,7 +61,7 @@
     "pixelmatch": "^4.0.2",
     "pngjs": "^3.3.3",
     "text-diff": "^1.0.1",
-    "typescript": "3.2.2"
+    "typescript": "^3.5.3"
```
TimvdLippe: You probably want a specific version here, as CI is running 3.8.3 instead.

lib/Accessibility.js
TimvdLippe: I am not sure if this actually works, but it is worth a try:

```js
/** @enum {string} */
const UserStringProperties = {
  Name: 'name',
  Value: 'value',
  ...
}
```

Then you can do `Object.values(UserStringProperties)` instead of the `userStringProperties` array down below.

TimvdLippe (approved):
Suggestion for the enum definitions and locking down the TS version. Other than that LGTM.

lib/Accessibility.js
jackfranklin: I tried this but it unfortunately doesn't work, because when you do `for foo in Object.values(UserStringProperties)` TypeScript only knows that `foo` is a `string`; it loses the fact that it was actually one of `UserStringProperties`, which causes problems. So I think this is the nicest way for now :(

jackfranklin:
@mathiasbynens I've rewritten the commit message 👍

mathiasbynens (approved):

mathiasbynens:
Alright, let's do it! Thanks everyone :)

PR #5559, opened 2020-03-31
chore: Upgrade TypeScript to 3.6

jackfranklin:
Continues the work to get up to TS 3.8 (latest release at time of
writing).

This version of TS introduced built in definitions for web workers that
include an `interface Worker` so TS gets confused when it sees us
reference a `Worker`. I have renamed the imports to `PuppeteerWorker` as
I couldn't figure out a way to tell TS to not load in the worker types;
longer term we might consider renaming `Worker` to `PuppeteerWorker` (or
an alternative) but that would be a breaking change that we don't need
right now.

The other fix is to solve an issue where the `ws` library doesn't
implement a `dispatchEvent` method - which is not a problem - but TS
expects it to. I've used `@ts-ignore` on that error because it's
absolutely fine and not something we rely on.

### DIFF
--- a/lib/Page.js
+++ b/lib/Page.js
@@ -26,7 +26,7 @@ const {Keyboard, Mouse, Touchscreen} = require('./Input');
 const Tracing = require('./Tracing');
 const {helper, debugError, assert} = require('./helper');
 const {Coverage} = require('./Coverage');
-const {Worker} = require('./Worker');
+const {Worker: PuppeteerWorker} = require('./Worker');
 const {createJSHandle} = require('./JSHandle');
 const {Accessibility} = require('./Accessibility');
 const {TimeoutSettings} = require('./TimeoutSettings');
@@ -79,7 +79,7 @@ class Page extends EventEmitter {
 
     this._screenshotTaskQueue = screenshotTaskQueue;
 
-    /** @type {!Map<string, Worker>} */
+    /** @type {!Map<string, PuppeteerWorker>} */
     this._workers = new Map();
     client.on('Target.attachedToTarget', event => {
       if (event.targetInfo.type !== 'worker') {
@@ -90,7 +90,7 @@ class Page extends EventEmitter {
         return;
       }
       const session = Connection.fromSession(client).session(event.sessionId);
-      const worker = new Worker(session, event.targetInfo.url, this._addConsoleMessage.bind(this), this._handleException.bind(this));
+      const worker = new PuppeteerWorker(session, event.targetInfo.url, this._addConsoleMessage.bind(this), this._handleException.bind(this));
       this._workers.set(event.sessionId, worker);
       this.emit(Events.Page.WorkerCreated, worker);
     });
@@ -274,7 +274,7 @@ class Page extends EventEmitter {
   }
 
   /**
-   * @return {!Array<!Worker>}
+   * @return {!Array<!PuppeteerWorker>}
    */
   workers() {
     return Array.from(this._workers.values());
--- a/lib/WebSocketTransport.js
+++ b/lib/WebSocketTransport.js
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-const WebSocket = require('ws');
+const NodeWebSocket = require('ws');
 
 /**
  * @implements {!Puppeteer.ConnectionTransport}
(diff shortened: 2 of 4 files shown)


jackfranklin:
@TimvdLippe happy to take any thoughts you have on TypeScript and if there's better workarounds.

My approach here is I'd rather do the minimum reasonable fixes to get TS happy so we can get up to 3.8, and then solve these problems "properly" when we migrate the source code to `.ts`.

tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "noEmit": true,
     "allowJs": true,
     "checkJs": true,
-    "target": "ESNext"
+    "target": "ESNext",
```
mathiasbynens: This makes it invalid JSON. Maybe TypeScript can handle that for its config files but then it's weird that the file is called `.json`. Let's get rid of this since I imagine this'll error in most editors. WDYT?

mathiasbynens (changes requested):

tsconfig.json
jackfranklin: Yeah I think TS is happy with this but agreed, this was an accidental by-product of some debugging anyway :)

lib/WebSocketTransport.js
```diff
@@ -29,6 +29,12 @@ class WebSocketTransport {
         perMessageDeflate: false,
         maxPayload: 256 * 1024 * 1024, // 256Mb
       });
+
+      /* error that WebSocket is not assignable to type WebSocket
+      * due to a misisng dispatchEvent() method which the ws library
+      * does not implement and we do not need
```
mathiasbynens, line 35:
```suggestion
      /* WebSocket is not assignable to type WebSocket due to a
       * missing dispatchEvent() method which the ws library
      * does not implement and we do not need.
```

lib/WebSocketTransport.js
```diff
@@ -29,6 +29,12 @@ class WebSocketTransport {
         perMessageDeflate: false,
         maxPayload: 256 * 1024 * 1024, // 256Mb
       });
+
+      /* WebSocket is not assignable to type WebSocket due to a
+       * missing dispatchEvent() method which the ws library
+      * does not implement and we do not need.
```
mathiasbynens:
```suggestion
      /* WebSocket is not assignable to type WebSocket due to a
       * missing dispatchEvent() method which the ws library
       * does not implement and we do not need.
```

mathiasbynens (approved):
LGTM but I'd like for @TimvdLippe to take a look as well before landing.

lib/Page.js
```diff
@@ -26,7 +26,7 @@ const {Keyboard, Mouse, Touchscreen} = require('./Input');
 const Tracing = require('./Tracing');
 const {helper, debugError, assert} = require('./helper');
 const {Coverage} = require('./Coverage');
-const {Worker} = require('./Worker');
+const {Worker: PuppeteerWorker} = require('./Worker');
```
TimvdLippe, line 29: @mathiasbynens Can the export of `Worker` be renamed to `PuppeteerWorker` in `Worker.js` or does that constitute a breaking change?

lib/WebSocketTransport.js
```diff
@@ -29,6 +29,12 @@ class WebSocketTransport {
         perMessageDeflate: false,
         maxPayload: 256 * 1024 * 1024, // 256Mb
       });
+
+      /* WebSocket is not assignable to type WebSocket due to a
+       * missing dispatchEvent() method which the ws library
```
TimvdLippe: Can we file a bug at either the `ws` library or its types to get that fixed?

TimvdLippe (approved):
If the export cant be renamed, this LGTM!

lib/Page.js
jackfranklin: I think that would be a breaking change if we support people who have done:

```js
const { Worker } = require('puppeteer/lib/Worker')
```

I'm not sure how common that is - @mathiasbynens WDYT?

lib/WebSocketTransport.js
jackfranklin: It's fixed in ws: https://github.com/websockets/ws/commit/cec1e1cf44f2aeadbe7aa271c46e00cee245a135

But upgrading `ws` in this project didn't help. I then thought about upgrading `@types/ws` and that didn't help either. So I _think_ the problem is in `@types/ws` but I'm not 100% clear.

lib/Page.js
mathiasbynens: It's a potentially breaking change that we can avoid in this case, so let's just stick to the more conservative change for now.

mathiasbynens (approved):

PR #5562, opened 2020-03-31
chore: Upgrade to TypeScript 3.7

jackfranklin:
TypeScript seems to struggle to understand `Promise.all` when the items
in the array return different types. If we were authoring in TS we could
fix this with TS generics (`Promise.all<OurTypeHere>(...)`) but for now
we can typecast the result. We'll fix this properly when we author in
TS.

### DIFF
--- a/lib/Coverage.js
+++ b/lib/Coverage.js
@@ -137,7 +137,7 @@ class JSCoverage {
   async stop() {
     assert(this._enabled, 'JSCoverage is not enabled');
     this._enabled = false;
-    const [profileResponse] = await Promise.all([
+    const result = await Promise.all([
       this._client.send('Profiler.takePreciseCoverage'),
       this._client.send('Profiler.stopPreciseCoverage'),
       this._client.send('Profiler.disable'),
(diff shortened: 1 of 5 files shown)


mathiasbynens (approved):

jackfranklin:
Don't merge this yet - the typecheck works but the TS API must have changed slightly which is causing some doc errors:

```
DocLint Failures:
  1) [MarkDown] Method Accessibility.snapshot() options.root ElementHandle != ElementHandle<>
  2) [MarkDown] Method ExecutionContext.queryObjects() prototypeHandle JSHandle != JSHandle<>
  3) [MarkDown] Method Page.queryObjects() prototypeHandle JSHandle != JSHandle<>
  4) [MarkDown] Method Request.respond() response.body string|Buffer != string|Buffer<>
```

EDIT: this is fixed by a33e743

mathiasbynens:
> We'll fix this properly when we author in TS.

Are we keeping track of these notes somewhere? If not, let's add a `// TODO`.

jackfranklin:
> Are we keeping track of these notes somewhere? If not, let's add a // TODO.

I think the `/** @type` is how we keep track of these; they are all effectively shortcuts in order to get us onto the right TS version. As we then convert files one by one we can't help but come back to them. So I don't think we need anything extra.

mathiasbynens (approved):

PR #5566, opened 2020-04-01
chore: Upgrade TypeScript to 3.8

jackfranklin:
Easy upgrade this one! No changes that affect us so this change simply
bumps TS in `package.json`.

Also found the `test-types` command and bumped that too.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "tsc": "tsc --version && tsc -p .",
     "apply-next-version": "node utils/apply_next_version.js",
     "bundle": "npx browserify -r ./index.js:puppeteer -o utils/browser/puppeteer-web.js",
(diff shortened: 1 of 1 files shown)


mathiasbynens (approved):

package.json
```diff
@@ -24,7 +24,7 @@
     "tsc": "tsc --version && tsc -p .",
     "apply-next-version": "node utils/apply_next_version.js",
     "bundle": "npx browserify -r ./index.js:puppeteer -o utils/browser/puppeteer-web.js",
-    "test-types": "node utils/doclint/generate_types && npx -p typescript@3.2 tsc -p utils/doclint/generate_types/test/",
+    "test-types": "node utils/doclint/generate_types && npx -p typescript@3.8 tsc -p utils/doclint/generate_types/test/",
```
mathiasbynens: Good catch. Shall we just drop the `@3.8` here, so that it uses the version down below in `package.json`?

jackfranklin: 💯

PR #5568, opened 2020-04-01
chore: Move source files into src/ and emit to lib/ with TypeScript

jackfranklin:
This updates our `tsconfig.json` so it emits our JavaScript files as
well as type checking them. We compile into `./out` which we then ship
in our npm package.

Because the `lib/` directory is exclusively JS files, this change is a
no-op in terms of code functionality but is the first step towards being
able to replace `lib/X.js` with `lib/X.ts` in a way that allows us to
migrate incrementally.

The `out` directory is gitignored, and the `lib` directory is
npmignored. On `npm publish` we will now run `npm run tsc` in order to
generate the outputted code.

Any imports that import from `lib/X` now import from `out/X`.

BREAKING CHANGE: if users are importing Puppeteer modules, e.g.:

```js
const PuppeteerWorker = require('puppeteer/lib/Worker');
```

This path has now changed to:

```js
const PuppeteerWorker = require('puppeteer/out/Worker');
```

** EDIT ** the above is outdated; the below is what we now do:

This updates our `tsconfig.json` so it emits our JavaScript files as
well as type checking them. We compile into `./lib` which we then ship
in our npm package. The source code has moved from `./lib` into `./src`.

Because the `src/` directory is exclusively JS files, this change is a
no-op in terms of code functionality but is the first step towards being
able to replace `src/X.js` with `src/X.ts` in a way that allows us to
migrate incrementally.

The `lib` directory is gitignored, and the `src` directory is
npmignored. On `npm publish` we will now run `npm run tsc` in order to
generate the outputted code.


jackfranklin:
@mathiasbynens @TimvdLippe would love your thoughts on this please.

The goal here is to have TypeScript not only type-check but emit our JavaScript, in preparation for us to start replacing files with `*.ts` ones.

This PR is more a starting point & one solution that works rather than definitely the best one - I think there's a few ways we could do this and I'm happy to chat about it. Can always jump on a call :)

jackfranklin:
CI is failing for a legitimate reason: `node install.js` needs `out/helper.js`, but that can't exist until TypeScript has run. (It depends on it because `node install.js` requires the main Puppeteer module)

_However_ for TypeScript to run it needs `protocol.d.ts`, which is installed by `node install.js`. So I think we need to either:

* Update the install script to not depend on `out/helper`
* Update the install script to depend on `lib/helper.js` (works until we want to TypeScriptify that file)
* Update the install script to grab protocol.d.ts, then run TypeScript, then run the install script. This seems wasteful given this runs every time anyone installs the package from npm.

EDIT: thinking about it, this isn't a problem for an end user installing the package because `out/` will already exist. So this is a problem for CI runs and installing the package to work on locally.

TimvdLippe:
GitHub search is not really working for me, but this search seemed the most sane: https://github.com/search?q=%22puppeteer%2Flib%22&type=Code That search isn't immediately showing any actual usage of direct files in puppeteer.

Emitting with `tsc` seems 👍 One option you also have is to move `lib/` into `src/` and then still write to `lib/`. That might not be a breaking change after all?

I have no knowledge about the install scripts and how they all tie together. I will leave that to @mathiasbynens

mathiasbynens:
We can prevent the breaking change in this case, so let's do it. I like Tim's suggestion of moving `lib` to `src` and then having `lib` be the generated output. I also appreciate you sending this smaller patch first, so that it's easier to review the actual changes 👍

jackfranklin:
Great thinking - I'll update this PR to move `lib/X` into `src/X` and generate into `lib`.

package.json
```diff
@@ -16,16 +16,19 @@
...
-    "test": "npm run lint --silent && npm run coverage && npm run test-doclint && npm run test-types && node utils/testrunner/test/test.js",
+    "test": "npm run tsc && npm run lint --silent && npm run coverage && npm run test-doclint && npm run test-types && node utils/testrunner/test/test.js",
+    "prepublishOnly": "npm run tsc",
+    "dev-install": "npm run tsc && node install.js",
     "install": "node install.js",
     "lint": "([ \"$CI\" = true ] && eslint --quiet -f codeframe . || eslint .) && npm run tsc && npm run doc",
     "doc": "node utils/doclint/cli.js",
     "coverage": "cross-env COVERAGE=true npm run unit",
-    "tsc": "tsc --version && tsc -p .",
+    "tsc": "tsc -p . && cp src/protocol.d.ts lib/ && cp src/externs.d.ts lib/",
```
mathiasbynens: Why remove the `tsc --version`?

utils/protocol-types-generator/index.js
```diff
@@ -72,12 +72,12 @@ declare global {
 
 export default Protocol;
 `;
-  const outputPath = path.join(__dirname, '..', '..', 'lib', 'protocol.d.ts');
+  const outputPath = path.join(__dirname, '..', '..', 'src', 'protocol.d.ts');
   require('fs').writeFileSync(outputPath, output);
   console.log(`Wrote protocol.d.ts for ${version} to ${path.relative(process.cwd(), outputPath)}`);
+  console.log(`You should commit the changes.`);
```
mathiasbynens, line 78: Nice touch!

mathiasbynens (approved):

package.json
jackfranklin: I accidentally left it in from debugging, but thinking about it it's probably useful so I'll add it back.

PR #5573, opened 2020-04-02
Automatic deploy screenshot Puppeteer

czuli:


czuli:
@googlebot I signed it!

jschfflr:
@mathiasbynens Not sure if we have a policy for endorsing third party services like this. Could you take a look and resolve accordingly?

mathiasbynens:
Yeah, we don’t generally do this, sorry.

PR #5575, opened 2020-04-02
chore: Skip failing cookie tests in Firefox

jackfranklin:
They fail because cookies in Firefox return a `sameSite` key which the
tests don't expect.

This is a solution that at least gets the Travis FF build (hopefully!)
green again. Longer term it'd be great to allow the assertion to change
based on the browser, rather than skip these tests entirely.


mathiasbynens:
@mjzffr Could you PTAL?

mjzffr (approved):
Thanks for looping me in, lgtm

mathiasbynens (approved):

PR #5576, opened 2020-04-02
Fix link to source files in documentation

tinovyatkin:
Look like it was overlooked while moving source files.


jackfranklin:
Ah thank you for this! Good spot 👍

mathiasbynens (approved):

PR #5581, opened 2020-04-03
chore: Use expect for assertions

jackfranklin:
Rather than use our own custom expect library, we can use expect from
npm [1], which has an API almost identical to the one Puppeteer has, but
with more options, better diffing and is used by many in the community
as it's the default assertions library that comes with Jest.

It's also thoroughly documented [2].

[1]: https://www.npmjs.com/package/expect
[2]: https://jestjs.io/docs/en/expect


test/emulation.spec.js
```diff
@@ -99,7 +99,12 @@ module.exports.addTests = function({testRunner, expect, puppeteer}) {
 
   describe('Page.emulateMedia', function() {
     it('should be an alias for Page.emulateMediaType', async({page, server}) => {
-      expect(page.emulateMedia).toEqual(page.emulateMediaType);
+      /* they aren't referentially equivalent
+       * so instead we check the bodies are the same
+       */
+      const emulateMediaStr = page.emulateMedia.toString();
+      const emulateMediaTypeStr = page.emulateMediaType.toString();
+      expect(emulateMediaStr).toEqual(emulateMediaTypeStr);
```
jackfranklin: @mathiasbynens I don't really understand this one; they are both methods on `Page.prototype` so I would expect them to be equivalent - but maybe I'm missing something obvious?

mathiasbynens: Did the TypeScriptification cause a change in the (now generated) code? They're definitely supposed to be aliases: https://github.com/puppeteer/puppeteer/blob/f43468452a4397d72d5f64ad75bf41f362b7d4df/src/Page.js#L1149-L1150

mathiasbynens: Like, does TypeScript translate that into something unexpected somehow?

jackfranklin: Nope - TypeScript outputs that line exactly :/

Is it something to do with how the test framework provides `page` ? I'm not sure.

package.json
```diff
@@ -58,6 +58,7 @@
     "cross-env": "^5.0.5",
     "eslint": "^6.8.0",
     "esprima": "^4.0.0",
+    "expect": "25.2.7",
```
mathiasbynens: Why is this an exact version and not a `^` version range like the others?

jackfranklin: Primarily because we got caught out running `typescript: '^3.6'` and CI installed a different version than I was running locally. Therefore I'd rather pin things so CI has the same versions.

Alternatively we could commit the `package-lock.json` and that way everything would run the same version?

test/emulation.spec.js
jackfranklin: This is to do with the `installAsyncHooks` code that creates new methods. Fixed by doing the legacy binding in `index.js`.

mathiasbynens: As discussed offline. Thanks for this drive-by fix!

mathiasbynens (approved):

package.json
mathiasbynens: This is already a problem (or non-problem, depending on your POV) for all the other dependencies as well. I'd prefer not deviating from the way we version other dependencies. Doing things inconsistently is confusing. Let's stick with`^` for now; when an `expect` update comes out that doesn't respect semver properly and breaks something, CI will tell us about it. WDYT?

jackfranklin: 👍 done!

PR #5583, opened 2020-04-03
Trim CI down whilst we get it stable

jackfranklin:
Our CI build has been incredibly flakey across all three of our current
CIs:

* Appveyor: Chromium + Windows
* Travis: Firefox + Linux, Chromium + Linux
* Cirrus: Chromium + Linux, Chromium + Mac

Legitimate issues and errors have been missed because it's expected that
the CI is red and therefore it's not seen as an issue when a PR's build
fails.

We should have a build that covers the full combo of browsers and
operating systems but it's more important to have a consistent, reliable
green build where failures are genuine. So this commit strips our CI
back to Chromium on Linux on Travis, and nothing more. Once this is
stable we will expand out into more operating systems and bring back
Firefox, too.


jackfranklin:
@mathiasbynens this is now passing on Travis. Please could you disable CirrusCI + Appveyor for now as discussed? Let's get CI stable on one browser and one OS and then we can iterate from there up to a full comprehensive suite where things are hopefully stable.

mathiasbynens:
> Please could you disable CirrusCI + Appveyor for now as discussed? 

Done.

PR #5584, opened 2020-04-03
chore: Fix Page.emulateMedia doclint failure

jackfranklin:


utils/doclint/check_public_api/index.js
```diff
@@ -176,8 +183,12 @@ function compareDocumentations(actual, expected) {
     const actualMethods = Array.from(actualClass.methods.keys()).sort();
     const expectedMethods = Array.from(expectedClass.methods.keys()).sort();
     const methodDiff = diff(actualMethods, expectedMethods);
-    for (const methodName of methodDiff.extra)
+    for (const methodName of methodDiff.extra) {
+      const expectedMissing = expectedMissingMethods.find(method => method.className === className && method.methodName === methodName);
```
mathiasbynens: Let's make `expectedMissingMethods` a `Map<String, Set>` to make this more efficient (`O(2)` i.e. `O(1)` instead of `O(n)`).

```js
const expectedMissingMethods = new Map([
  ['Page', new Set('emulateMedia')],
]);
```

mathiasbynens (changes requested):

utils/doclint/check_public_api/index.js
jackfranklin: 👍

utils/doclint/check_public_api/index.js
```diff
@@ -176,8 +183,12 @@ function compareDocumentations(actual, expected) {
     const actualMethods = Array.from(actualClass.methods.keys()).sort();
     const expectedMethods = Array.from(expectedClass.methods.keys()).sort();
     const methodDiff = diff(actualMethods, expectedMethods);
-    for (const methodName of methodDiff.extra)
+    for (const methodName of methodDiff.extra) {
+      const missingMethodsForClass = expectedNonExistingMethods.get(className);
+      const methodNameIsExpectedMissing = missingMethodsForClass ? missingMethodsForClass.has(methodName) : false;
+      if (methodNameIsExpectedMissing) continue;
```
mathiasbynens: How about something like:

```suggestion
      const missingMethodsForClass = expectedNonExistingMethods.get(className);
      if (!missingMethodsForClass) continue;
      if (missingMethodsForClass.has(methodName)) continue;
```

...or...

```
      const missingMethodsForClass = expectedNonExistingMethods.get(className);
      if (!missingMethodsForClass || missingMethodsForClass.has(methodName)) continue;
```

I think the former is most readable, though. WDYT?

PR #5595, opened 2020-04-06
feat(TypeScript): Move DeviceDescriptors to TS

jackfranklin:
This commit moves `src/DeviceDescriptors` to be authored in TypeScript.
This file was chosen due to its simplicity so that we can focus on
getting a mixed JS/TS codebase playing nicely before migrating the more
complex files.

The file itself was a bit odd: although the array of devices was
exported via `module.exports` that was never referenced by any
consumers; each device was also exported via `module.exports[name] =
device` and that is how it's consumed. The Puppeteer docs suggest using
it like so:

```js
puppeteer.devices['iPhone 6']
```

So instead of exporting the array and then setting a bunch of properties
on that, we instead define the array and export an object of keys where
each key is a device.

Rather than export an object I'd much rather export a Map, but that
would be a breaking change and I'm keen to avoid those for the time
being.

Note that we have to use special TypeScript specific syntax for the
export that enables it to work in a CommonJS codebase [1] but again I'd
rather this than move to ESM at this time. TypeScript still outputs
CommonJS into `lib/` as you would expect.

[1]: https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

### DIFF
--- a/src/DeviceDescriptors.ts
+++ b/src/DeviceDescriptors.ts
@@ -14,7 +14,20 @@
  * limitations under the License.
  */
 
-module.exports = [
+interface Device {
+  name: string;
+  userAgent: string;
+  viewport: {
+    width: number;
+    height: number;
+    deviceScaleFactor: number;
+    isMobile: boolean;
+    hasTouch: boolean;
+    isLandscape: boolean;
+  };
+};
+
+const devices: Device[] = [
   {
     'name': 'Blackberry PlayBook',
     'userAgent': 'Mozilla/5.0 (PlayBook; U; RIM Tablet OS 2.1.0; en-US) AppleWebKit/536.2+ (KHTML like Gecko) Version/7.2.1.0 Safari/536.2+',
@@ -868,5 +881,15 @@ module.exports = [
     }
   }
 ];
-for (const device of module.exports)
-  module.exports[device.name] = device;
+
+type DevicesMap = {
+  [name: string]: Device;
+};
+
+const devicesMap: DevicesMap = {};
+
+for (const device of devices) {
+  devicesMap[device.name] = device;
+}
+
+export = devicesMap;
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,8 @@
     "checkJs": true,
     "outDir": "./lib",
     "target": "ESNext",
-    "moduleResolution": "node"
+    "moduleResolution": "node",
+    "module": "CommonJS"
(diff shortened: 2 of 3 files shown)


jackfranklin:
@mathiasbynens see what you think - I don't want to ship this anytime soon but thought I'd see what the changes looked like.

AviVahl:
I've made a PR a while ago which got ignored:
https://github.com/puppeteer/puppeteer/pull/5233

After seeing the work being merged these days, you may find some of the those changes useful.

mathiasbynens:
No longer exporting the array is also a breaking change. I think that's fine for the reasons you point out, but we should just be aware of it.

mathiasbynens (approved):

jackfranklin:
@AviVahl I did see that indeed, thanks! It will definitely be a useful reference as we try to incrementally migrate over 👍

src/DeviceDescriptors.ts
```diff
@@ -868,5 +881,15 @@ module.exports = [
...
   }
 ];
-for (const device of module.exports)
-  module.exports[device.name] = device;
+
+type DevicesMap = {
+  [name: string]: Device
+}
+
+const devicesMap: DevicesMap = {}
```
mathiasbynens:
```suggestion
const devicesMap: DevicesMap = {};
```

src/DeviceDescriptors.ts
```diff
@@ -14,7 +14,20 @@
...
+interface Device {
+  name: string;
+  userAgent: string;
+  viewport: {
+    width: number;
+    height: number;
+    deviceScaleFactor: number;
+    isMobile: boolean;
+    hasTouch: boolean;
+    isLandscape: boolean
```
mathiasbynens:
```suggestion
    isLandscape: boolean;
```

right?

mathiasbynens:
Note to self: I'll rephrase the commit message upon merging in light of the following:

> So instead of exporting the array and then setting a bunch of properties
> on that, we instead define the array and export an object of keys where
> each key is a device.

This is technically a breaking change (although for the reasons you provide it seems fine in this case).

> Rather than export an object I'd much rather export a Map, but that
> would be a breaking change and I'm keen to avoid those for the time
> being.

Given the above, it would be "a _larger_ breaking change".

mathiasbynens:
doclint seems to be timing out on one of the bots. Could you please take a look?

PR #5600, opened 2020-04-07
Migrate unit tests to Mocha

jackfranklin:
@mathiasbynens this is a proof of concept of how we can move to Mocha by replacing the custom test runner with a few small helpers.

I have migrated cookies.spec.js and all the tests pass. 

I'm keen to plough ahead and start migrating more test files, but wanted to raise this PR to check:

* do you have any comments on how the new cookie tests look? Before I make similar changes across all files!
* Using the `cookies.mocha.spec.js` renaming we can run one test at a time rather than migrate in bulk. Would you rather one PR for one test file or I just do them all in bulk? The changes are mechanical mostly so I think one PR might be OK.


script.js
```diff
@@ -0,0 +1,22 @@
...
+const newContents = contents.replace(regex, (match, group1) => {
+  console.log('here', group1);
+
+  const replacementLines = `async() => {
+    const { ${group1} } = getTestState();
+  `;
+  return replacementLines;
+});
+
+fs.writeFileSync(filePath, newContents, { encoding: 'utf8'});
```
jackfranklin: whoops, ignore this, it's a quick script I wrote to automate some of the migration. Will not make it into the actual codebase.

test/mocha-utils.js
```diff
@@ -0,0 +1,56 @@
...
+  server.PREFIX = `http://localhost:${port}`;
+  server.CROSS_PROCESS_PREFIX = `http://127.0.0.1:${port}`;
+  server.EMPTY_PAGE = `http://localhost:${port}/empty.html`;
+  // const httpsPort = port + 1;
+  // httpsServer = await TestServer.createHTTPS(assetsPath, httpsPort);
+  // httpsServer.enableHTTPCache(cachedPath);
+  // httpsServer.PORT = httpsPort;
+  // httpsServer.PREFIX = `https://localhost:${httpsPort}`;
+  // httpsServer.CROSS_PROCESS_PREFIX = `https://127.0.0.1:${httpsPort}`;
+  // httpsServer.EMPTY_PAGE = `https://localhost:${httpsPort}/empty.html`;
```
jackfranklin: this stuff will be enabled again - just haven't yet migrated a test that needs it.

This setup is lifted right from our custom test runner.

test/mocha-utils.js
```diff
@@ -0,0 +1,56 @@
...
+    state.server.reset();
+    state.context = await state.browser.createIncognitoBrowserContext();
+    state.page = await state.context.newPage();
+  });
+
+  after(async() => {
+    await state.browser.close();
+    await state.server.stop();
+  });
+}
```
jackfranklin: all these before/beforeEach are almost exactly what's in the existing testrunner

test/mocha-utils.js
```diff
@@ -0,0 +1,56 @@
...
+
+  return server;
+};
+
+exports.getTestState = () => state;
+
+const state = {};
+
+// purposefully global
+if (process.argv.some(part => part.includes('mocha'))) {
```
jackfranklin: this is to stop our test runner trying to execute this file. Only a temporary thing whilst we're mid-migration.

.mocharc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  file: ['./test/mocha-utils.js'],
+  spec: 'test/*.mocha.spec.js',
+  reporter: 'dot',
+}
```
mathiasbynens: match existing coding style

```suggestion
};
```

script.js
```diff
@@ -0,0 +1,22 @@
+const fs = require('fs');
+const path = require('path');
+
+const file = process.argv.slice(2)[0];
```
mathiasbynens: no need for `.slice()` if you're just reading an indexed element anyway

script.js
```diff
@@ -0,0 +1,22 @@
...
+
+const filePath = path.resolve(path.join(__dirname, file));
```
mathiasbynens: no need for `path.resolve` here, since `__dirname` is absolute and so the result of `path.join` will be as well

script.js
mathiasbynens: ack

test/mocha-utils.js
mathiasbynens: ack

mathiasbynens:
I love the way the tests look, thanks for taking this on Jack! I can't wait to drop the custom test runner code. To answer your question, I'd prefer a single large PR in this case since the changes are mostly mechanical.

mathiasbynens (approved):
Rubberstamp LGTM as discussed offline

test/mocha-utils.js
```diff
@@ -0,0 +1,155 @@
...
+  if (!fs.existsSync(executablePath))
+    throw new Error(`Browser is not downloaded at ${executablePath}. Run 'npm install' and try to re-run tests`);
+}
+
+const setupGoldenAssertions = () => {
+  const suffix = product.toLowerCase();
+  const GOLDEN_DIR = path.join(__dirname, 'golden-' + suffix);
+  const OUTPUT_DIR = path.join(__dirname, 'output-' + suffix);
+  if (fs.existsSync(OUTPUT_DIR))
+    rm(OUTPUT_DIR);
```
mathiasbynens, line 76: Just hit this: `rm()` is undefined here.

Issue #5034, opened 2019-10-13
FreeBSD is not supported


ip75:
v1.20.0
puppeteer/lib/BrowserFetcher.js:
...

```

class BrowserFetcher {
  /**
   * @param {string} projectRoot
   * @param {!BrowserFetcher.Options=} options
   */
  constructor(projectRoot, options = {}) {
    this._downloadsFolder = options.path || path.join(projectRoot, '.local-chromium');
    this._downloadHost = options.host || DEFAULT_DOWNLOAD_HOST;
    this._platform = options.platform || '';
    if (!this._platform) {
      const platform = os.platform();
      if (platform === 'darwin')
        this._platform = 'mac';
      else if (platform === 'linux')
        this._platform = 'linux';
      else if (platform === 'win32')
        this._platform = os.arch() === 'x64' ? 'win64' : 'win32';
      assert(this._platform, 'Unsupported platform: ' + os.platform());
    }
    assert(supportedPlatforms.includes(this._platform), 'Unsupported platform: ' + this._platform);
  }


```
freebsd is not supported

tarkhil:
puppeteer is dependent on headless chromium which is not available for FreeBSD, so it won't work anyway. 

Well, you can try to install headless browsers in linux emulator than patch BrowserFetcher, I will try it as soon as I'll have time.

Don't expect headless chromium ported to FreeBSD soon, all resources are distracted to more pressing issues - renaming master to main, Code of Conduct, inclusive naming and all that Extremely  Urgent And Important Stuff.

ip75:
Chromium exists in standard ports collection of FreeBSD.
/usr/ports/www/chromium/

[ /usr/ports/www/chromium ] # cat pkg-descr
Chromium is an open-source browser project that aims to build a safer,
faster, and more stable way for all users to experience the web.

The Chromium website contains design documents, architecture overviews,
testing information, and more to help you learn to build and work with
the Chromium source code.

`--headless `
headless mode available since 2017

tarkhil:
As far as I've checked (right now), headless build is not supported.

root@over:/usr/ports/www/chromium # grep -i headless *
Makefile.tests:         headless_unittests \
pkg-plist:%%DATADIR%%/headless_lib.pak
pkg-plist:%%DATADIR%%/resources/inspector/cm_headless/cm_headless.js

It only builds with everything on board ((

And my attempt to build headless remotely controlled firefox resulted in 
X-required uncontrollable one. I'd love to have spare time to 
[...]

PR #5601, opened 2020-04-07
Fix for #5034

plato79:
This fixes the errors when installing, although environment variable to skip download still should be used like this:
`setenv PUPPETEER_SKIP_DOWNLOAD true`


plato79:
@googlebot I signed it!

plato79:
Could somebody check this? This fixes https://github.com/puppeteer/puppeteer/issues/5034 . Although it stucks for something not related with this it seems.

0mp:
It would be great if we could have it merged. @plato79, do you still have time to resolve the conflicts perhaps?

jschfflr:
@mathiasbynens Not sure if supporting `freebsd` is something we want to officially commit to doing - could you take a look and resolve this accordingly?

0mp:
> @mathiasbynens Not sure if supporting `freebsd` is something we want to officially commit to doing - could you take a look and resolve this accordingly?

Thanks for working on this PR. It's much appreciated. 😃

src/BrowserFetcher.js
```diff
@@ -222,6 +228,8 @@ class BrowserFetcher {
         executablePath = path.join(folderPath, archiveName(this._product, this._platform, revision), 'Chromium.app', 'Contents', 'MacOS', 'Chromium');
       else if (this._platform === 'linux')
         executablePath = path.join(folderPath, archiveName(this._product, this._platform, revision), 'chrome');
+      else if (this._platform === 'freebsd')
```
mathiasbynens, line 231: Can we merge this with the `else if (this._platform === 'linux')` above?

src/BrowserFetcher.js
```diff
@@ -67,6 +69,8 @@ function archiveName(product, platform, revision) {
       return 'chrome-linux';
     if (platform === 'mac')
       return 'chrome-mac';
+    if (platform === 'freebsd')
```
mathiasbynens, line 72: Can we merge this with the `if (platform === 'linux')` above?

src/BrowserFetcher.js
```diff
@@ -231,6 +239,8 @@ class BrowserFetcher {
         executablePath = path.join(folderPath, 'Firefox Nightly.app', 'Contents', 'MacOS', 'firefox');
       else if (this._platform === 'linux')
         executablePath = path.join(folderPath, 'firefox', 'firefox');
+      else if (this._platform === 'freebsd')
```
mathiasbynens, line 242: Merge this with the check above

src/BrowserFetcher.js
```diff
@@ -32,12 +32,14 @@ const getProxyForUrl = require('proxy-from-env').getProxyForUrl;
 const downloadURLs = {
   chrome: {
     linux: '%s/chromium-browser-snapshots/Linux_x64/%d/%s.zip',
+    freebsd: '%s/chromium-browser-snapshots/Linux_x64/%d/%s.zip',
```
mathiasbynens, line 35: This and the Firefox URL below are the same ones as for Linux? In that case let's simplify the patch by removing this line and line 42 and treating `'freebsd'` as equivalent to `'linux'`.

That way this change also doesn't impact the public `puppeteer.createBrowserFetcher` API.

mathiasbynens (changes requested):
I believe this patch affects the public `puppeteer.createBrowserFetcher` API and its `platform` option but it doesn't document the changes.
However, it's better not to change this API and stick to the current list of user-exposed `platform`s (`"linux"|"mac"|"win32"|"win64"`) since those are the ones we test and provide support for.

It looks like this patch is just using the Linux URLs and codepaths for `freebsd` anyway — so can we just treat `freebsd` as an alias to `linux`? That would avoid these issues.

plato79:
> I believe this patch affects the public `puppeteer.createBrowserFetcher` API and its `platform` option but it doesn't document the changes.
> However, it's better not to change this API and stick to the current list of user-exposed `platform`s (`"linux"|"mac"|"win32"|"win64"`) since those are the ones we test and provide support for.
> 
> It looks like this patch is just using the Linux URLs and codepaths for `freebsd` anyway — so can we just treat `freebsd` as an alias to `linux`? That would avoid these issues.

The thing is puppeteer checks for the system and if it's freebsd it just doesn't work at all.. The links doesn't matter because chromium should be installed manually anyway on freebsd ( there is no official support ). You could also use dummy download links or no download links at all.. So, as I said in the first post you need to use `setenv PUPPETEER_SKIP_DOWNLOAD true` first to avoid downloading part.

jschfflr:
Not sure if I understand you correctly: For FreeBSD, the normal linux build of Chromium doesn't work. That's why you `setenv PUPPETEER_SKIP_DOWNLOAD true` to prevent Puppeteer from downloading something and you instead install it manually. In which situation would the `BrowserFetcher` be used then?

plato79:
> Not sure if I understand you correctly: For FreeBSD, the normal linux build of Chromium doesn't work. That's why you `setenv PUPPETEER_SKIP_DOWNLOAD true` to prevent Puppeteer from downloading something and you instead install it manually. In which situation would the `BrowserFetcher` be used then?

Yes, normal linux build doesn't work but there are custom FreeBSD builds of chromium exists which you need to compile yourself.. The thing is puppeteer still searches for freebsd in BrowserFetcher if the system is FreeBSD. This is just a dummy selection to prevent runtime errors.

jschfflr:
@plato79 My expectation would be that we don't ever get into a situation where we'd run `BrowserFetcher` if `PUPPETEER_SKIP_DOWNLOAD` is set. Could you maybe share a quick snippet to reproduce the problem?

jschfflr:
@plato79 Gentle ping :)

jschfflr:
@plato79 I'm going to close this for now because it's not actionable for me but please feel free to reopen if you found out why the browser fetcher is still invoked even though `PUPPETEER_SKIP_DOWNLOAD ` was set!

PR #5591, opened 2020-04-06
chore: Fix missed `src/` vs `lib/` documentation

jackfranklin:

PR #5603, opened 2020-04-08
chore: Fix missed `src/` vs `lib/` documentation

Gbahdeyboh:
Fixes #5591 more

### DIFF
--- a/DeviceDescriptors.js
+++ b/DeviceDescriptors.js
@@ -14,4 +14,4 @@
  * limitations under the License.
  */
 
-module.exports = require('./lib/DeviceDescriptors');
+module.exports = require('./src/DeviceDescriptors');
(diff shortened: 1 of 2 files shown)


mathiasbynens (approved):
cc @jackfranklin

mathiasbynens:
LGTM but this seems blocked on https://github.com/puppeteer/puppeteer/pull/5595, right?

jackfranklin:
These imports are deliberately left as `lib/` - we need them to import from the compiled output. There should be a comment explaining that to be fair - it's non obvious. I still don't fully understand why we have these special files that just re-export from `lib/` but until I do I'm keen to leave these be.

@Gbahdeyboh appreciate your help but in this case we need to leave these be unfortunately.

PR #5607, opened 2020-04-08
fixed api emulateMedia the right way

savigny030:
now doing it the right way, i corrected all occurring things. sorry for incomplete patch before. :v:

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1323,7 +1323,7 @@ List of all available devices is available in the source code: [DeviceDescriptor
 - `type` <?[string]> Changes the CSS media type of the page. The only allowed values are `'screen'`, `'print'` and `null`. Passing `null` disables CSS media emulation.
 - returns: <[Promise]>
 
-**Note:** This method is deprecated, and there is no backwards compatibility anymore. Use [`page.emulateMediaType(type)`](#pageemulatemediatypetype) instead.
+**Note:** This method is deprecated, and only kept around as an alias for backwards compatibility. Use [`page.emulateMediaType(type)`](#pageemulatemediatypetype) instead.
 
 #### page.emulateMediaFeatures(features)
 - `features` <?[Array]<[Object]>> Given an array of media feature objects, emulates CSS media features on the page. Each media feature object must have the following properties:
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -1323,7 +1323,7 @@ List of all available devices is available in the source code: [DeviceDescriptor
 - `type` <?[string]> Changes the CSS media type of the page. The only allowed values are `'screen'`, `'print'` and `null`. Passing `null` disables CSS media emulation.
 - returns: <[Promise]>
 
-**Note:** This method is deprecated, and only kept around as an alias for backwards compatibility. Use [`page.emulateMediaType(type)`](#pageemulatemediatypetype) instead.
+**Note:** This method is deprecated, and there is no backwards compatibility anymore. Use [`page.emulateMediaType(type)`](#pageemulatemediatypetype) instead.
```
mathiasbynens: What makes you say that? It's definitely still supposed to be around as an alias.

Which Puppeteer version are you using and what are you trying to do exactly? cc @jackfranklin

savigny030: i use ts and it says `Property 'emulateMedia' does not exist on type 'Page'.`  It's maybe typescript related. Puppeteer is at `puppeteer@2.1.1`.

savigny030: then it's a problem here [@types/puppeteer](https://www.npmjs.com/package/@types/puppeteer)...

jackfranklin: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/puppeteer/index.d.ts does not include `emulateMedia` on `Page`. 

Given that `2.1.1` is a couple of months old I don't think this anything to do with our latest changes.

This is another +1 to migrating the Puppeteer source code to TypeScript so we can generate our own definitions.

savigny030: okay i reverted this line, but there was still smth wrong in page.pdf which i corrected.

savigny030: Mathias do you work for google in germany? if so how is it? It was always a dream. I'm still a young student, maybe that dream comes true one day... :)

jackfranklin: Nice, thanks!

mathiasbynens (approved):

PR #5616, opened 2020-04-09
chore: Migrate remaining tests to Mocha

jackfranklin:
This commit updates all the non-Puppeteer unit tests to run using Mocha
and then deletes the custom testrunner framework from this repository.
The documentation has also been updated.


jackfranklin:
Note that the only changes to the spec files are the renaming and mechanical updates to migrate to Mocha. All the assertions are untouched :)

mathiasbynens (approved):

Issue #5625, opened 2020-04-11
Get the sanList array of the SSL Certificate


mtoma:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 2.1.1 
* Platform / OS version: openSUSE Tumbleweed 20200123
* URLs (if applicable):
* Node.js version: v12.16.1

**What steps will reproduce the problem?**

Navigate to any https URL

**What is the expected result?**

Be able to access the sanList array of the Developer Tools security details object

**What happens instead?**

The SecurityDetails class doesn't expose this parameter:
```
SecurityDetails {
  _subjectName: 'domain.com',
  _issuer: "Let's Encrypt Authority X3",
  _validFrom: 1578909379,
  _validTo: 1586685379,
  _protocol: 'TLS 1.2'
}
```

PR #5626, opened 2020-04-11
Fixes issue #5625 - Get the sanList array of the SSL Certificate

mtoma:
Fixes issue #5625 - Get the sanList array of the SSL Certificate


mtoma:
I signed the CLA, but it seems undetected.

PR #5628, opened 2020-04-11
feat: add securityDetails.subjectAlternativeNames()

mtoma:
Fixes issue #5625 - Get the sanList array of the SSL Certificate


test/ignorehttpserrors.spec.js
```diff
@@ -63,6 +63,7 @@ describeFailsFirefox('ignoreHTTPSErrors', function() {
       expect(securityDetails.subjectName()).toBe('puppeteer-tests');
       expect(securityDetails.validFrom()).toBe(1550084863);
       expect(securityDetails.validTo()).toBe(33086084863);
+      expect(securityDetails.sanList()).toEqual([]);
```
mathiasbynens: Could you add a test with an actual SAN list?

You can use `https://1000-sans.badssl.com/` for the test, and check the length of the resulting array, as well as one particular element (so that the tests show the expected format of a single SAN object).

docs/api.md
```diff
@@ -3727,6 +3728,9 @@ Contains the URL of the response.
 #### securityDetails.protocol()
 - returns: <[string]> String with the security protocol, eg. "TLS 1.2".
 
+#### securityDetails.sanList()
```
mathiasbynens: Let's rename to `subjectAlternativeNames` to more closely match `subjectName`.

mathiasbynens (changes requested):
Left some suggestions. I'm open to this change but would like to double-check with @whimboo and @mjzffr that this is something Firefox can expose as well.

mjzffr:
Yes, Firefox can expose this.

test/ignorehttpserrors.spec.js
mtoma: Shouldn't this be included in the local test webserver certificate? Seems risky to me to have the whole test chain depend on network presence & internet presence.

mathiasbynens: Yes, adding some SANs to the local cert would be ideal.

mathiasbynens: @mtoma Do you want to give this a go?

mtoma: @mathiasbynens Yes, it's the plan. I'm just catastrophically busy these days. Any help or guidance with generating the local certificate wit a san list would be greatly appreciated.
Maybe someone knows how the certificate was generated in the first place? openssl command line probably?

mathiasbynens: I have no idea how the existing cert was generated, but it seems we could use something like https://github.com/FiloSottile/mkcert.

docs/api.md
mtoma: Done

test/ignorehttpserrors.spec.js
mtoma: Done. For future the command line I used to generate the certificate is: 
```
openssl req -x509 -nodes -subj "/CN=puppeteer-tests" -newkey rsa:2048 -keyout key.pem -out cert.pem -addext "subjectAltName=DNS:www.puppeteer-tests.tld,DNS:www.puppeteer-tests-1.tld" -days 3650
```

test/ignorehttpserrors.spec.js
```diff
@@ -61,8 +61,9 @@ describeFailsFirefox('ignoreHTTPSErrors', function() {
       const protocol = serverRequest.socket.getProtocol().replace('v', ' ');
       expect(securityDetails.protocol()).toBe(protocol);
       expect(securityDetails.subjectName()).toBe('puppeteer-tests');
-      expect(securityDetails.validFrom()).toBe(1550084863);
-      expect(securityDetails.validTo()).toBe(33086084863);
+      expect(securityDetails.validFrom()).toBe(1589283259);
+      expect(securityDetails.validTo()).toBe(1904643259);
+      expect(securityDetails.subjectAlternativeNames()).toEqual(['www.puppeteer-tests.tld', 'www.puppeteer-tests-1.tld']);
```
mathiasbynens: Let’s use `.test` as the TLD per https://tools.ietf.org/html/rfc2606#section-2.

mathiasbynens (approved):
Looking good! One last request w.r.t. the test domain names.

mathiasbynens (approved):

mathiasbynens:
Would you mind rebasing? `src/NetworkManager.js` has moved here: https://github.com/puppeteer/puppeteer/blob/master/src/NetworkManager.ts

After the rebase this looks good to go. Thanks!

Issue #5356, opened 2020-01-28
2.1.0 Windows regression: process stays open due to open sockets


AviVahl:
Project: https://github.com/wixplosives/mocha-pup

When [upgrading to 2.1.0](https://github.com/wixplosives/mocha-pup/pull/1), tests started failing (timeouts) on Windows CI. Containers got stalled.

My tests execute a CLI which:
- runs webpack
- starts an http server
- launches puppeteer and opens a page.
- checks stuff in the page.
- closes the browser.

Debugging revealed the CLI process never shuts down, as there are two sockets left open:
https://github.com/wixplosives/mocha-pup/runs/413725670

Further investigation suggested it has to do with the browser connection.

Switching to `pipe: true` made it work. :o

Best guess: the WebSocket connection stays open and never gets destroyed.
Suspected PR: https://github.com/puppeteer/puppeteer/pull/5137 (it refactored `Launcher.js`, which has the connection setup)

### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 2.1.0
* Platform / OS version: Happens only on a CI which runs Windows Server 2019.
* URLs (if applicable): https://github.com/wixplosives/mocha-pup/pull/1
* Node.js version: 12.14.1

**What steps will reproduce the problem?**
I wish it was that easy. :(
I couldn't reproduce the issue locally. I use Fedora 31, and only had Windows 10 at my office (where issue didn't reproduce). Issue happened consistently in every Windows CI run.

_Please include code that reproduces the issue._
Apologies again. I'll try to create a smaller reproducible test-case.
(maybe a script that opens and closes puppeteer 10 times)

**What is the expected result?**
After calling browser.close(), no socket should stay open.

**What happens instead?**
Sockets get left behind, process never exits.

AviVahl:
Still trying to get a smaller reproduction ready.
Meanwhile, I merged to `master`, with `pipe: true`, and opened a separate PR to see the issue:
https://github.com/wixplosives/mocha-pup/pull/2

nemchik:
I believe I may be affected by a similar issue. I have a process in place that works perfectly fine with 2.0.0 but when upgrading to 2.1.0 or 2.1.1 I see intermittent failures with various things I'm doing.

I confirmed with my project that the only change was the version of puppeteer. Do you think it would be possible to confirm that downgrading puppeteer to 2.0.0 resolves your issue?

P.S. I'm not a contributor to this project, just another user 😄

[...]

PR #5637, opened 2020-04-14
chore(ci): re-enable tests on Windows

jackfranklin:


AviVahl:
EDIT: irrelevant. already fixed by `puppeteer@3.0.0`. 

You should note there was a Windows-specific regression right before you started working on this repo... https://github.com/puppeteer/puppeteer/issues/5356

So if you see stalled test runs, it might be related. From own experience, it reproduces more consistently on GitHub Actions, but also happens on Travis.

jackfranklin:
@AviVahl thanks for the heads up! I'll see if I can dig into that later.

jackfranklin:
@mjzffr hey, we are seeing one test fail consistently on Windows at the moment and wondered if you'd mind having a look as we think it relates to some changes in the browser fetcher (we're not 100% though).

Here you can see a test run: https://travis-ci.com/github/puppeteer/puppeteer/jobs/320395754

And the issue is:

```
  1) Launcher specs
       Puppeteer
         Puppeteer.launch
           should be able to launch different products:
     Error: Could not find browser revision latest. Run "npm install" or "yarn install" to download a browser binary.
      at FirefoxLauncher.launch (lib\Launcher.js:336:23)
```

It passes just fine on Linux CI, but consistently fails on Windows. We'd like to resolve this so we can add Windows back to CI. We could skip this test but it'd be great if we could figure a solution :)

test/network.spec.js
```diff
@@ -189,15 +190,15 @@ describe('network', function() {
...
+      expect(await response.text()).toBe(`{"foo": "bar"}${os.EOL}`);
     });
     it('should return uncompressed text', async() => {
       const { page, server } = getTestState();
 
       server.enableGzip('/simple.json');
       const response = await page.goto(server.PREFIX + '/simple.json');
       expect(response.headers()['content-encoding']).toBe('gzip');
-      expect(await response.text()).toBe('{"foo": "bar"}\n');
+      expect(await response.text()).toEqual(`{"foo": "bar"}${os.EOL}`);
```
mathiasbynens: `toBe`

test/network.spec.js
```diff
@@ -189,15 +190,15 @@ describe('network', function() {
       const { page, server } = getTestState();
 
       const response = await page.goto(server.PREFIX + '/simple.json');
-      expect(await response.text()).toBe('{"foo": "bar"}\n');
+      expect(await response.text()).toBe(`{"foo": "bar"}${os.EOL}`);
```
mathiasbynens: Why do we need the `os.EOL`?

The `simple.json` file is checked in to our repository and uses LF for line endings. Our `.gitattributes` enforces LF line endings on checkout as well. How could we still end up with CRLF in server responses?

.travis.yml
```diff
@@ -13,6 +13,18 @@ cache:
...
+    - os: "windows"
+      node_js: "10.19.0"
+      env:
+        - CHROMIUM=true
+      before_install:
+        # populate .local-firefox for launcher tests
+        - PUPPETEER_PRODUCT=firefox npm install
+      script:
+        - ls .local-chromium
+        - ls .local-firefox
```
mathiasbynens:
```suggestion
        - ls .local-chromium .local-firefox
```

test/network.spec.js
jackfranklin: Hmm I'm not sure - I presumed this was maybe due to OS level differences when the server fetches the JSON file but I'm not sure if that makes sense or not thinking about it.

mathiasbynens: Without this change, does the test fail?

jackfranklin: Yes it does.  We'll `trimEnd()` it to avoid any line ending faff 👍

test/headful.spec.js
```diff
@@ -85,26 +85,33 @@ describeChromeOnly('headful tests', function() {
...
-      await rmAsync(userDataDir).catch(e => {});
-      expect(cookie).toBe('foo=true');
-    });
+    itFailsWindowsUntilDate(
+        /* We have deferred fixing this test on Windows in favour of
+         * getting all other Windows tests running on CI. Putting this
+         * date in to force us to come back and debug properly in the
+         * future.
+         */
+        new Date(2020, 5, 1),
```
mathiasbynens: Readability (to avoid the off-by-one-month gotcha):

```suggestion
        new Date('2020-06-01'),
```

test/launcher.spec.js
```diff
@@ -367,21 +368,28 @@ describe('Launcher specs', function() {
...
-            expect(userAgent).toContain('Chrome');
-          else
-            expect(userAgent).toContain('Firefox');
-        }
+      /* We think there's a bug in the FF Windows launcher, or some
+       * combo of that plus it running on CI, but we're deferring fixing
+       * this so we can get Windows CI stable and then dig into this
+       * properly with help from the Mozilla folks.
+       */
+      itFailsWindowsUntilDate(new Date(2020, 5, 1), 'should be able to launch Firefox', async() => {
```
mathiasbynens: same here

```suggestion
      itFailsWindowsUntilDate(new Date('2020-06-01'), 'should be able to launch Firefox', async() => {
```

test/mocha-utils.js
```diff
@@ -92,6 +93,16 @@ global.itFailsFirefox = (...args) => {
     return it(...args);
 };
 
+global.itFailsWindowsUntilDate = (date, ...args) => {
+  const today = new Date();
```
mathiasbynens: Since we only need the timestamp to compare:

```suggestion
  const today = Date.now;
```

Since this is pretty short we don't really need the `today` variable (better: `now`?) anymore.

mathiasbynens (approved):
LGTM with some final nitpicks

test/network.spec.js
```diff
@@ -15,6 +15,7 @@
  */
 
 const fs = require('fs');
+const os = require('os');
```
mathiasbynens: We don't need this anymore:

```suggestion

```

test/mocha-utils.js
mathiasbynens: We could make this function accept a string for the `date` parameter, since all callers do `new Date(string)` anyway.

mathiasbynens: i.e. `itFailsWindowsUntilDate('2020-06-01', blah)`

jackfranklin: May do that in a follow up PR - I'm also tempted to tidy up the test helpers a bit. Right now we have `itFailsFirefox` (with no date) and `itFailsWindowsUntilDate`. I'd quite like a `itFails({ product: 'firefox', date: '... })` or similar API rather than bespoke methods every time

PR #5644, opened 2020-04-15
feat(chromium): roll Chromium to r737027

hanselfmu:
This corresponds to Chromium ~~80.0.3987.0~~ 81.0.4044.0.

This roll includes:

- [DevTools] Add Cookie Priority support to CDP
https://chromium-review.googlesource.com/c/chromium/src/+/1959029

- Reject cookies with empty names and values
https://chromium-review.googlesource.com/c/chromium/src/+/1982549

### DIFF
--- a/test/cookies.spec.js
+++ b/test/cookies.spec.js
@@ -41,13 +41,14 @@ describe('Cookie specs', () => {
         size: 16,
         httpOnly: false,
         secure: false,
-        session: true
+        session: true,
+        priority: 'Medium',
       }]);
     });
     it('should properly report httpOnly cookie', async() => {
       const {page, server} = getTestState();
       server.setRoute('/empty.html', (req, res) => {
-        res.setHeader('Set-Cookie', ';HttpOnly; Path=/');
+        res.setHeader('Set-Cookie', 'a=b; HttpOnly; Path=/');
         res.end();
       });
       await page.goto(server.EMPTY_PAGE);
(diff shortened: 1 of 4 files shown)


test/cookies.spec.js
```diff
@@ -96,7 +97,8 @@ describe('Cookie specs', () => {
           size: 12,
           httpOnly: false,
           secure: false,
-          session: true
+          session: true,
+          priority: 'Medium',
```
mathiasbynens: Since `priority` is not cross-browser yet (I think?), I think we should not expose this property yet. @hanselfmu Could you change `page.cookies()` to remove the `priority` property for now? We can re-enable it once support is more widespread. cc @mjzffr

Vorobeyko (approved):

test/cookies.spec.js
mathiasbynens: FYI @mjzffr, once this one lands, we plan on cutting a new release.

mjzffr: That's right, cookie priority is Chrome-only as far a I know. Thanks for the heads-up.

hanselfmu: I've removed the 'priority' field from cookies for now.

mathiasbynens (approved):

mathiasbynens:
Hmm, looks like r737027 corresponds to version 81.0.4044.0. I guess that was a typo. I'll amend the commit message.

PR #5655, opened 2020-04-16
fix(JSHandle): Fixes file upload

paullewis:


src/JSHandle.js
```diff
@@ -15,6 +15,7 @@
  */
 
 const {helper, assert, debugError} = require('./helper');
+const path = require('path');
```
mathiasbynens: Nit: since this is only used in `uploadFile`, let's move the `require` there. That way, scripts that don't use `uploadFile` don't need to pay the cost.

mathiasbynens (approved):
LGTM % nit

src/JSHandle.js
```diff
@@ -311,38 +312,11 @@ class ElementHandle extends JSHandle {
    * @param {!Array<string>} filePaths
    */
   async uploadFile(...filePaths) {
-    const isMultiple = await this.evaluate(element => element.multiple);
-    assert(filePaths.length <= 1 || isMultiple, 'Multiple file uploads only work with <input type=file multiple>');
-    // These imports are only needed for `uploadFile`, so keep them
-    // scoped here to avoid paying the cost unnecessarily.
```
mathiasbynens, line 317: Ah, yeah, this comment. :)

src/JSHandle.js
paullewis: Done.

src/JSHandle.js
paullewis: Done. Added the comment for posterity.

paullewis:
PTAL

src/JSHandle.js
```diff
@@ -313,36 +313,29 @@ class ElementHandle extends JSHandle {
...
-      };
-      files.push(file);
+    const files = filePaths.map(filePath => path.resolve(filePath));
+    const { objectId } = this._remoteObject;
+    const { node } = await this._client.send('DOM.describeNode', { objectId });
+    const { backendNodeId } = node;
+
+    // The zero-length array is a special case, it seems that DOM.setFileInputFiles does
+    // not actually update the files in that case, so the solution is to eval the element
+    // value to an empty string directly.
```
mathiasbynens: Woah, weird -- good catch! Instead of `value`, could we set `files`?

paullewis: Done with `DataTransfer` as per offline discussion.

PR #5666, opened 2020-04-17
docs: Change the Alpine Docker example to use env instead of launch option

neet:
This PR changes the documentation for the way to specify `executablePath` on Alpine Docker.

Formerly, the doc used options of `puppeteer.launch`, but since Puppeteer has an environment variable for it, I think it would be better to use it instead of specifying it every time.

I've already verified this can be built through the latest version of Docker. The Docker image can be found at [DockerHub](https://hub.docker.com/repository/docker/neetshin/pptr-alpine) and the source code it at [neetbox/pptr-alpine-docker](https://github.com/neetbox/pptr-alpine-docker).

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -340,7 +340,8 @@ RUN apk add --no-cache \
 ...
 
 # Tell Puppeteer to skip installing Chrome. We'll be using the installed package.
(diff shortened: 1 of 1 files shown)


neet:
@googlebot I signed it!

mathiasbynens (approved):

PR #5682, opened 2020-04-19
Fix: support async functions as an argument for waitForFunction

### LINKED ISSUE #4045: waitForFunction only called once
The quick summary is that whether I use `raf`, or set `polling` to a number, `waitForFunction` only ever runs once.

### Steps to reproduce

* Puppeteer version: 1.12.2
* Platform / OS version: macOS Mojave 10.14.3
* URLs (if applicable): example.com
* Node.js version: v8.15.0

The following puppeteer script is saved as `pup.js`:

```js

  const [page] = await browser.pages()

  let result = false

  page.exposeFunction('getResult', () => {
    console.log('here', result)

    const thisResult = result

    result = true

    return thisResult
  })

  await page.goto('http://example.com')
  await page.waitForFunction('getResult()')
  await page.close()
  await browser.close()
})
```

My command to run it is `node pup.js`.

I expect the following output:

```
here false
here true
```

and a clean exit.

**What happens instead?**

I get the following:

```
here false
(node:64382) UnhandledPromiseRejectionWarning: TimeoutError: waiting for function failed: timeout 30000ms exceeded
    at new WaitTask (/Users/jstaab/Desktop/cc/client/node_modules/puppeteer/lib/DOMWorld.js:554:28)
    at DOMWorld.waitForFunction (/Users/jstaab/Desktop/cc/client/node_modules/puppeteer/lib/DOMWorld.js:459:12)
    at Frame.waitForFunction (/Users/jstaab/Desktop/cc/client/node_modules/puppeteer/lib/FrameManager.js:636:28)
    at Page.waitForFunction (/Users/jstaab/Desktop/cc/client/node_modules/puppeteer/lib/Page.js:1098:29)
    at puppeteer.launch.then (/Users/jstaab/Desktop/cc/client/pup.js:19:14)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
(node:64382) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:64382) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

And it fails to exit (because I never get to `page.close()`).

EslamHiko:
This PR is made to fix : https://github.com/puppeteer/puppeteer/issues/4045

After investigating I found out that `predicate.apply(null, args)` returns an unresolved promise in case predicate is `AsyncFunction` which causes puppeteer to hang.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2070,7 +2070,7 @@ await fileChooser.accept(['/tmp/myfile.pdf']);
 
 
 #### page.waitForFunction(pageFunction[, options[, ...args]])
-- `pageFunction` <[function]|[string]> Function to be evaluated in browser context
+- `pageFunction` <[function]|[AsyncFunction]|[string]> Function to be evaluated in browser context
 - `options` <[Object]> Optional waiting parameters
   - `polling` <[string]|[number]> An interval at which the `pageFunction` is executed, defaults to `raf`. If `polling` is a number, then it is treated as an interval in milliseconds at which the function would be executed. If `polling` is a string, then it can be one of the following values:
     - `raf` - to constantly execute `pageFunction` in `requestAnimationFrame` callback. This is the tightest polling mode which is suitable to observe styling changes.
--- a/docs/api.md
+++ b/docs/api.md
@@ -2070,7 +2070,7 @@ await fileChooser.accept(['/tmp/myfile.pdf']);
 
 
 #### page.waitForFunction(pageFunction[, options[, ...args]])
-- `pageFunction` <[function]|[AsyncFunction]|[string]> Function to be evaluated in browser context
+- `pageFunction` <[function]|[string]> Function to be evaluated in browser context
 - `options` <[Object]> Optional waiting parameters
   - `polling` <[string]|[number]> An interval at which the `pageFunction` is executed, defaults to `raf`. If `polling` is a number, then it is treated as an interval in milliseconds at which the function would be executed. If `polling` is a string, then it can be one of the following values:
     - `raf` - to constantly execute `pageFunction` in `requestAnimationFrame` callback. This is the tightest polling mode which is suitable to observe styling changes.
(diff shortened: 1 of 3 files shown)


EslamHiko:
@googlebot I signed it!

mathiasbynens:
This looks good, but

- please update the docs accordingly
- this unfortunately requires a rebase now :(

Would you be up for rebasing?

mathiasbynens (approved):

mathiasbynens (changes requested):
marking as “changes requested” per the above

docs/api.md
```diff
@@ -2097,7 +2097,7 @@ To pass arguments from node.js to the predicate of `page.waitForFunction` functi
 
 ```js
 const selector = '.foo';
-await page.waitForFunction(selector => !!document.querySelector(selector), {}, selector);
+await page.waitForFunction(async selector => !!document.querySelector(selector), {}, selector);
```
mathiasbynens: This example does not need `async`, so let's leave it as-is and instead add an additional example that does something asynchronously.

EslamHiko: @mathiasbynens what do you think of this example : 

`page.waitForFunction` also works with asynchronous promises:

```js
const username = 'github-username';
await page.waitForFunction(async username => {
  let githubResponse = await fetch(`https://api.github.com/users/${username}`);
  let githubUser = await githubResponse.json();
  // show the avatar
  let img = document.createElement('img');
  img.src = githubUser.avatar_url;
  // wait 3 seconds
  await new Promise((resolve, reject) => setTimeout(resolve, 3000));
  img.remove();
}, {}, username);
```

mathiasbynens: Looks good, but let's use `const` instead of `let` to match the Puppeteer coding style.

EslamHiko: @mathiasbynens Done 😄

mathiasbynens (approved):

PR #5692, opened 2020-04-20
chore: migrate `src/pipetransport` to typescript

jackfranklin:
Hit one bump in the fact that I want to share an interface across files.
TypeScript only lets you import/export these if you're using ESM, not
CommonJS. So the two options are:

- Migrate to ESM on a per file basis as we do this migration. This won't
affect the output as we output as CommonJS.
- Create a global `types.d.ts` file that we'll use and then migrate to
ESM after.

Right now I've gone for the second option in order to not introduce more
changes in one go. But if we end up finding we have lots of
interfaces/types/etc that we want modules to expose, we might decide
slowly introducing ESM might be a better way forwards.

### DIFF
--- a/src/Launcher.js
+++ b/src/Launcher.js
@@ -29,7 +29,7 @@ const {helper, assert, debugError} = require('./helper');
 const debugLauncher = require('debug')(`puppeteer:launcher`);
 const {TimeoutError} = require('./Errors');
 const WebSocketTransport = require('./WebSocketTransport');
-const PipeTransport = require('./PipeTransport');
+const {PipeTransport} = require('./PipeTransport');
 
 const mkdtempAsync = helper.promisify(fs.mkdtemp);
 const removeFolderAsync = helper.promisify(removeFolder);
--- a/src/types.d.ts
+++ b/src/types.d.ts
@@ -15,8 +15,8 @@
  */
 
 /* These types exist here until we migrate over to ESM where we can
- *import / export them properly from modules - TS doesn't support
- *exposing interfaces in CommonJS land.
+ * import / export them properly from modules - TS doesn't support
+ * exposing interfaces in CommonJS land.
  */
 interface PuppeteerEventListener {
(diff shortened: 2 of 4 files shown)


src/Launcher.js
```diff
@@ -29,7 +29,7 @@ const {helper, assert, debugError} = require('./helper');
 const debugLauncher = require('debug')(`puppeteer:launcher`);
 const {TimeoutError} = require('./Errors');
 const WebSocketTransport = require('./WebSocketTransport');
-const PipeTransport = require('./PipeTransport');
+const {PipeTransport} = require('./PipeTransport');
```
jackfranklin, line 32: did this change because the majority of files do this and I think named imports in ESM are nicer than defaults so this aligns more closely with where I see the codebase going.

src/types.d.ts
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/* These types exist here until we migrate over to ESM where we can
+ *import / export them properly from modules - TS doesn't support
+ *exposing interfaces in CommonJS land.
```
mathiasbynens:
```suggestion
 * import / export them properly from modules - TS doesn't support
 * exposing interfaces in CommonJS land.
```

src/Launcher.js
mathiasbynens: +1 named exports 4 lyfe

mathiasbynens (approved):
I would have been fine with option 1 too, FWIW. Happy to go with your recommendation though!

jackfranklin:
@mathiasbynens 👍 It might be that we realise soon option 1 is better in which case let's just swap. One isn't more work than the other really so I'll just see how we go

PR #5694, opened 2020-04-20
chore: migrate `src/Connection` to TypeScript

jackfranklin:
This commit migrates `src/Connection` to TypeScript. It also changes its
exports to be ESM because TypeScript's support for exporting values to
use as types via CommonJS is poor (by design) and so rather than battle
that it made more sense to migrate the file to ESM.

The good news is that TypeScript is still outputting to `lib/` as
CommonJS, so the fact that we author in ESM is actually not a breaking
change at all.

So going forwards we will:

* migrate TS files to use ESM for importing and exporting
* continue to output to `lib/` as CommonJS
* continue to use CommonJS requires when in a `src/*.js` file

I'd also like to split `Connection.ts` into two; I think the
`CDPSession` class belongs in its own file, but I will do that in
another PR to avoid this one becoming bigger than it already is.

I also turned off `@typescript-eslint/no-use-before-define` as I don't
think it was adding value and Puppeteer's codebase seems to have a style
of declaring helper functions at the bottom which is fine by me.

### DIFF
--- a/src/Connection.ts
+++ b/src/Connection.ts
@@ -13,38 +13,51 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-const {assert} = require('./helper');
-const {Events} = require('./Events');
-const debugProtocol = require('debug')('puppeteer:protocol');
-const EventEmitter = require('events');
+import helper = require('./helper');
+const {assert} = helper;
 
-class Connection extends EventEmitter {
+import EventsModule = require('./Events');
+const {Events} = EventsModule;
+
+import debug = require('debug');
+const debugProtocol = debug('puppeteer:protocol');
+
+import EventEmitter = require('events');
+
+interface ConnectionCallback {
+    resolve: Function;
+    reject: Function;
+    error: Error;
+    method: string;
+
+}
+
+export class Connection extends EventEmitter {
+  _url: string;
+  _transport: Puppeteer.ConnectionTransport;
+  _delay: number;
+  _lastId = 0;
+  _sessions: Map<string, CDPSession> = new Map();
+  _closed = false;
+
+  _callbacks: Map<number, ConnectionCallback> = new Map();
+
   /**
    * @param {string} url
    * @param {!Puppeteer.ConnectionTransport} transport
    * @param {number=} delay
    */
-  constructor(url, transport, delay = 0) {
+  constructor(url: string, transport: Puppeteer.ConnectionTransport, delay = 0) {
     super();
     this._url = url;
-    this._lastId = 0;
-    /** @type {!Map<number, {resolve: function, reject: function, error: !Error, method: string}>}*/
-    this._callbacks = new Map();
     this._delay = delay;
 
     this._transport = transport;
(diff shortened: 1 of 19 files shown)


mathiasbynens (approved):

mathiasbynens:
Yay for modules!

utils/doclint/check_public_api/index.js
```diff
@@ -247,21 +247,61 @@ function compareDocumentations(actual, expected) {
...
+  }
+
+
   /**
    * @param {string} source
    * @param {!Documentation.Type} actual
    * @param {!Documentation.Type} expected
    */
   function checkType(source, actual, expected) {
+
```
mathiasbynens:
```suggestion
```

mathiasbynens (approved):
Still LGTM

src/Connection.ts
```diff
@@ -149,45 +147,49 @@ class Connection extends EventEmitter {
...
-  send(method, params = {}) {
+  send<T extends keyof Protocol.CommandParameters>(method: T, params?: Protocol.CommandParameters[T]): Promise<Protocol.CommandReturnValues[T]> {
     if (!this._connection)
       return Promise.reject(new Error(`Protocol error (${method}): Session closed. Most likely the ${this._targetType} has been closed.`));
-    const id = this._connection._rawSend({sessionId: this._sessionId, method, params});
+
+    const id = this._connection._rawSend({
+      sessionId: this._sessionId,
+      method,
+      params: params || {}
```
jackfranklin, line 188: hey @mjzffr I think there might be a Firefox bug here with the fact that it's expecting params to be present. Without this `|| {}` some Firefox specs fail like so:

```
 1) Page.click
       should not hang with touch-enabled viewports:
     Error: Protocol error (Page.reload): (destructured parameter) is undefined reload@chrome://remote/content/domains/content/Page.jsm:135:1
execute@chrome://remote/content/domains/DomainCache.jsm:99:25
receiveMessage@chrome://remote/content/sessions/ContentProcessSession.jsm:69:45
MessageListener.receiveMessage*ContentProcessSession@chrome://remote/content/sessions/ContentProcessSession.jsm:27:25
@chrome://remote/content/sessions/frame-script.js:12:1
      at Promise (lib/Connection.js:147:63)
      at new Promise (<anonymous>)
      at CDPSession.send (lib/Connection.js:146:16)
      at Page.reload (lib/Page.js:630:26)
      at Page.<anonymous> (lib/helper.js:82:27)
      at Page.setViewport (lib/Page.js:776:24)
    -- ASYNC --
      at Page.<anonymous> (lib/helper.js:81:19)
      at Context.it (test/click.spec.js:207:16)
```

However Chrome is fine. I think that maybe FF is expecting params where it shouldn't? This is an easy workaround for now but might be worth looking into. I also might have misunderstood the issue so sorry if that's the case. Also pinging @mathiasbynens for any thoughts.

mjzffr: @jackfranklin Thanks, I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1631570

PR #5698, opened 2020-04-21
Launcher.js: don't exit process on SIGINT

AndrewSB:
This caused a bug where other signal handlers for a node process don't receive 
their listener callbacks on a SIGINT. With this change, other signal listeners 
that were added before puppeteer's will also fire

not sure if this breaks something else inside puppeteer, but this fixed the issue I was running into in my tree

cc @JoelEinbinder

### DIFF
--- a/src/Launcher.js
+++ b/src/Launcher.js
@@ -30,7 +30,6 @@ const debugLauncher = require('debug')(`puppeteer:launcher`);
 const {TimeoutError} = require('./Errors');
 const {WebSocketTransport} = require('./WebSocketTransport');
 const {PipeTransport} = require('./PipeTransport');
-
 const mkdtempAsync = helper.promisify(fs.mkdtemp);
 const removeFolderAsync = helper.promisify(removeFolder);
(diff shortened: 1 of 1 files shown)


AndrewSB:
Actually, I stand corrected - this doesn't work - it calls my SIGINT handler, but then doesn't exit the node process... I'm not sure what the right thing to do here is.

Any ideas? Is there documentation around this issue? I couldn't find any previous work around this

JoelEinbinder:
Launch with handleSIGINT: false and close the browser/exit in your own sigint handler. One handler per node process keeps things simple.

AndrewSB:
in my handler, how do I instruct the browser to close? the public API of 

```js
await page.close()
await browser.close()
```

or something else?

jackfranklin:
@AndrewSB yes, `browser.close()` should do it. I'm going to close this PR based on your comment that this PR doesn't work as intended but please shout if there's more to dig into here or if there's a Puppeteer bug you're hitting. Thanks!

PR #5700, opened 2020-04-21
chore: enforce spacing around braces

jackfranklin:
The codebase was incredibly inconsistent with the use of spacing around
curly braces, e.g.:

```
// this?
const a = {b: 1}
// or?
const a = { b: 1 }
```

This extended into import statements also. I don't have a strong
preference for each one but I feel strongly that we should be consistent
so this PR enforces the rule.


mathiasbynens (approved):

mathiasbynens:
The Google JS style guide says "no spaces" but I'm fine with either option. Agreed that what matters is that we have consistency, not what the particular style is.

jackfranklin:
> The Google JS style guide says "no spaces"

Thinking about it, that's what Clang (and therefore DevTools) does, so I'll swap and do that one.

PR #5705, opened 2020-04-21
chore: migrate src/ExecutionContext

jackfranklin:
I spent a while trying to decide on the best course of action for
typing the `evaluate` function.

Ideally I wanted to use generics so that as a user you could type
something like:

```
handle.evaluate<HTMLElement, number, boolean>((node, x) => true, 5)
```

And have TypeScript know the arguments of `node` and `x` based on those
generics. But I hit two problems with that:

* you have to have n overloads of `evaluate` to cope for as many number
of arguments as you can be bothered too (e.g. we'd need an overload for
1 arg, 2 args, 3 args, etc)
* I decided it's actually confusing because you don't know as a user
what those generics actually map to.

So in the end I went with one generic which is the return type of the
function:

```
handle.evaluate<boolean>((node, x) => true, 5)
```

And `node` and `x` get typed as `any` which means you can tell TS
yourself:

```
handle.evaluate<boolean>((node: HTMLElement, x:  number) => true, 5)
```

I'd like to find a way to force that the arguments after the function do
match the arguments you've given (in the above example, TS would moan if
I swapped that `5` for `"foo"`), but I tried a few things and to be
honest the complexity of the types wasn't worth it, I don't think.

I'm very open to tweaking these but I'd rather ship this and tweak going
forwards rather than spend hours now tweaking. Once we ship these
typedefs and get feedback from the community I'm sure we can improve
them.

### DIFF
--- a/src/JSHandle.ts
+++ b/src/JSHandle.ts
@@ -1,4 +1,4 @@
-/**
+**
  * Copyright 2019 Google Inc. All rights reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
@@ -15,6 +15,7 @@
  */
 
 import {helper, assert, debugError} from './helper';
+import {ExecutionContext} from './ExecutionContext';
 import {CDPSession} from './Connection';
 
 interface BoxModel {
@@ -26,7 +27,7 @@ interface BoxModel {
   height: number;
 }
 
-export function createJSHandle(context: Puppeteer.ExecutionContext, remoteObject: Protocol.Runtime.RemoteObject): JSHandle {
+export function createJSHandle(context: ExecutionContext, remoteObject: Protocol.Runtime.RemoteObject): JSHandle {
   const frame = context.frame();
   if (remoteObject.subtype === 'node' && frame) {
     const frameManager = frame._frameManager;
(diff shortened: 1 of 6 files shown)


src/JSHandle.ts
```diff
@@ -151,14 +152,14 @@ export class ElementHandle extends JSHandle {
...
   async _scrollIntoViewIfNeeded(): Promise<void> {
-    const error = await this.evaluate<string | false>(async(element, pageJavascriptEnabled) => {
+    const error = await this.evaluate<Promise<string | false>>(async(element: HTMLElement, pageJavascriptEnabled: boolean) => {
       if (!element.isConnected)
         return 'Node is detached from document';
       if (element.nodeType !== Node.ELEMENT_NODE)
         return 'Node is not of type HTMLElement';
       // force-scroll if page's javascript is disabled.
       if (!pageJavascriptEnabled) {
-        element.scrollIntoView({block: 'center', inline: 'center', behavior: 'instant'});
```
jackfranklin, line 161: This is interesting: TS shouts here because I think `instant` was going to be in the spec but now isn't? So I've ditched this in favour of `auto` which is the default. @mathiasbynens WDYT? I'm not sure if Chrome does still support instant but if it's going/gone from the spec presumably we'll stop supporting it one day.

jackfranklin:
I also had a look at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/puppeteer/index.d.ts which do some smart things to get good type support for `$eval` and the like and I think we can use that as inspiration to vastly improve upon the types in this PR. @mathiasbynens I'd like to merge this as the first step and then follow up with a PR that improves the types a bunch if that's OK with you? Else this PR will grow a lot.

@mathiasbynens do we support taking a string argument as the function for `evaluate()` ? Some of the JSDoc comments say we do but I can't find an example so I'm wondering if that's correct or not. If we didn't have to type `pageFunction` as `Function | string` that would help with better TypeScript types.

src/JSHandle.ts
mathiasbynens: Woah, this is a breaking change. We shouldn't work around TypeScript bugs by changing observable behavior in Puppeteer. Chrome still supports `behavior: 'instant'`. Can we make TypeScript ignore this particular line?

jackfranklin: @mathiasbynens this isn't a TypeScript bug, but fine, yes we can.

mathiasbynens: (It’s arguably a bug: it doesn’t match real-world browser behavior. In general, when the spec and implementations disagree, then implementations take precedence, IMHO. This is a separate discussion though :])

mathiasbynens:
@jackfranklin We [support it](https://github.com/puppeteer/puppeteer/blob/6760b9225e52d2722805ba5b8f7f76816da7c15c/src/ExecutionContext.js#L69-L88) but I don’t think the API docs capture that (and we probably shouldn't recommend that pattern).

src/JSHandle.ts
jackfranklin: keyword there is "arguably" ;) 😂

jackfranklin:
> We support it but I don’t think the API docs capture that (and we probably shouldn't recommend that pattern).

Great - I'll continue to support it in TS land. Maybe one to consider deprecating in future releases.

mathiasbynens (approved):

PR #5714, opened 2020-04-22
chore: add test for npm package installing correctly

jackfranklin:
This command packs up the module and installs it again to check we're
correctly bundling everything we need to allow users to do a fresh
install.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ addons:
       # This is required to run new chrome on old trusty
       - libnss3
       # Needed for realpath which we use in npm run test-install
-      - coreutils
+      - realpath
 notifications:
   email: false
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ addons:
     packages:
       # This is required to run new chrome on old trusty
       - libnss3
+      # Needed for realpath which we use in npm run test-install
+      - coreutils
 notifications:
   email: false
 cache:
@@ -33,7 +35,9 @@ jobs:
         - "sh -e /etc/init.d/xvfb start"
         # populate .local-firefox for launcher tests
         - PUPPETEER_PRODUCT=firefox npm install
-      script: ./travis/chromium.sh
+      script:
+        - npm run test-install
+        - travis/chromium.sh
     - node_js: "10.19.0"
       dist: trusty
       env:
--- a/package.json
+++ b/package.json
@@ -30,14 +30,16 @@
     "bundle": "npm run tsc && npx browserify -r ./index.js:puppeteer -o utils/browser/puppeteer-web.js",
     "test-types": "node utils/doclint/generate_types && tsc --version && tsc -p utils/doclint/generate_types/test/",
     "unit-bundle": "mocha --config mocha-config/browser-bundle-tests.js",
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -30,14 +30,16 @@
     "bundle": "npm run tsc && npx browserify -r ./index.js:puppeteer -o utils/browser/puppeteer-web.js",
     "test-types": "node utils/doclint/generate_types && tsc --version && tsc -p utils/doclint/generate_types/test/",
     "unit-bundle": "mocha --config mocha-config/browser-bundle-tests.js",
-    "update-protocol-d-ts": "node utils/protocol-types-generator"
+    "update-protocol-d-ts": "node utils/protocol-types-generator",
+    "test-install": "./scripts/test-install.sh"
```
mathiasbynens:
```suggestion
    "test-install": "scripts/test-install.sh"
```

scripts/test-install.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env sh
+
+# Pack the module into a tarball
+npm pack
+tarball=$(realpath puppeteer-*.tgz)
```
mathiasbynens:
```suggestion
tarball="$(realpath puppeteer-*.tgz)"
```

scripts/test-install.sh
```diff
@@ -0,0 +1,13 @@
...
+cd "$(mktemp -d)"
+npm init --yes
```
mathiasbynens: Do we need this?

jackfranklin: Added this because otherwise you get npm warning you about non existing `package.json` which at first glance looks a bit like an error, but it's not. I think silencing those errors is preferable.

scripts/test-install.sh
```diff
@@ -0,0 +1,13 @@
...
+# Check we can install from the tarball.
+# This emulates installing from npm and ensures that:
+# 1. we publish the right files in the `files` list from package.json
+# 2. The install script works and correctly exits without errors
+# 3. Requiring Puppeteer from Node works.
+npm install "${tarball}" && node --eval="require('puppeteer')" && rm "${tarball}";
```
mathiasbynens: nit: either end each line with a semicolon (which I weakly prefer because it enables copying and pasting as a one-liner) or don't use semicolons at all

Also:

```suggestion
npm install "${tarball}" && node --eval="require('puppeteer')"
rm "${tarball}"
```

scripts/test-install.sh
mathiasbynens: If that's the goal, then let's just use `--loglevel silent` for `npm install`?

.travis.yml
```diff
@@ -33,7 +33,9 @@ jobs:
         - "sh -e /etc/init.d/xvfb start"
         # populate .local-firefox for launcher tests
         - PUPPETEER_PRODUCT=firefox npm install
-      script: ./travis/chromium.sh
+      script:
+        - npm run test-install
+        - ./travis/chromium.sh
```
mathiasbynens: while we're here...

```suggestion
        - travis/chromium.sh
```

mathiasbynens (approved):

Issue #5720, opened 2020-04-23
Unable to override clipboard permissions


VikramTiwari:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 3.0.0
* Platform / OS version: OSX
* URLs (if applicable):
* Node.js version: 12.16.1

**What steps will reproduce the problem?**
Try to override clipboard permissions. They won't work.

```
const context = browser.defaultBrowserContext();
context.overridePermissions('https://example.com', ['clipboard-read']);
```

Seems like this is happening because of change in underlying CDP.
In code the clipboard permissions are being converted to an older version while the updated value is supposed to be: `clipboardReadWrite` ([from CDP](https://chromedevtools.github.io/devtools-protocol/tot/Browser/#type-PermissionType))

Current code: https://github.com/puppeteer/puppeteer/blob/master/src/Browser.js#L347-L348

VikramTiwari:
@mathiasbynens I have created a quick PR. Not sure if you also want to expose `clipboardSanitizedWrite` as an option.

PR #5721, opened 2020-04-23
Update clipboard read write permissions with CDP

VikramTiwari:
For https://github.com/puppeteer/puppeteer/issues/5720


mathiasbynens:
Thanks for the patch! For the record, this was changed upstream in https://chromium-review.googlesource.com/c/chromium/src/+/1895749.

mathiasbynens (approved):

PR #5731, opened 2020-04-24
Add page.pdfReadable, a way to stream the pdf without disk I/O

bmisiak:
Hello.

I had to render some big PDFs. Unfortunately, puppeteer's public API exposes only `Page.pdf()`, which always returns a concatenated `Buffer`, even though internally it deals with a stream of data. Buffer.concat seems to be copying each chunk before concatenating them, so processing my sample 60MB PDF required at least 120MB of node's memory for a single PDF. Saving to a temporary file won't help because the method always creates the buffer. This can be avoided by exposing the underlying stream as `Readable`.

I wanted to add a parameter to `Page.pdf()`, but its return type would have to change to `Promise<Buffer | Readable>` or, ideally, its signature changed to `pdf<B extends boolean>(..., returnReadable: B): Promise<B extends true ? Readable : Buffer>`, but both would be a breaking change for TypeScript consumers. Therefore, I settled on adding a new method, `pdfReadable`, which allows us to retain 100% backwards compatibility.

🤗

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1519,6 +1519,21 @@ If URLs are specified, only cookies for those URLs are returned.
 
 > **NOTE** This method is identical to [page.pdf](#pagepdfoptions), except it returns the PDF as a readable stream of binary data. If you are generating very large PDFs, it may be useful to use a stream to avoid high memory usage. This version will ignore the `path` option.
 
+```js
+const puppeteer = require('puppeteer');
+
+(async () => {
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+
+  // Stream a PDF into a file
+  const pdfStream = await page.createPDFStream();
+  const writeStream = fs.createWriteStream('test.pdf');
+  pdfStream.pipe(writeStream);
+  await browser.close();
+})();
+```
+
 #### page.deleteCookie(...cookies)
 
 - `...cookies` <...[Object]>
--- a/docs/api.md
+++ b/docs/api.md
@@ -162,6 +162,7 @@
   * [page.metrics()](#pagemetrics)
   * [page.mouse](#pagemouse)
   * [page.pdf([options])](#pagepdfoptions)
+  * [page.pdfReadable([options])](#pagepdfreadableoptions)
   * [page.queryObjects(prototypeHandle)](#pagequeryobjectsprototypehandle)
   * [page.reload([options])](#pagereloadoptions)
   * [page.screenshot([options])](#pagescreenshotoptions)
@@ -2050,6 +2051,33 @@ The `format` options are:
 >
 > 1. Script tags inside templates are not evaluated.
 > 2. Page styles are not visible inside templates.
+
+#### page.pdfReadable([options])
+- `options` <[Object]> Options object which might have the following properties:
+  - `scale` <[number]> Scale of the webpage rendering. Defaults to `1`. Scale amount must be between 0.1 and 2.
+  - `displayHeaderFooter` <[boolean]> Display header and footer. Defaults to `false`.
+  - `headerTemplate` <[string]> HTML template for the print header. Should be valid HTML markup with following classes used to inject printing values into them:
+    - `date` formatted print date
+    - `title` document title
+    - `url` document location
+    - `pageNumber` current page number
+    - `totalPages` total pages in the document
+  - `footerTemplate` <[string]> HTML template for the print footer. Should use the same format as the `headerTemplate`.
+  - `printBackground` <[boolean]> Print background graphics. Defaults to `false`.
+  - `landscape` <[boolean]> Paper orientation. Defaults to `false`.
+  - `pageRanges` <[string]> Paper ranges to print, e.g., '1-5, 8, 11-13'. Defaults to the empty string, which means print all pages.
+  - `format` <[string]> Paper format. If set, takes priority over `width` or `height` options. Defaults to 'Letter'.
+  - `width` <[string]|[number]> Paper width, accepts values labeled with units.
+  - `height` <[string]|[number]> Paper height, accepts values labeled with units.
+  - `margin` <[Object]> Paper margins, defaults to none.
+    - `top` <[string]|[number]> Top margin, accepts values labeled with units.
+    - `right` <[string]|[number]> Right margin, accepts values labeled with units.
+    - `bottom` <[string]|[number]> Bottom margin, accepts values labeled with units.
+    - `left` <[string]|[number]> Left margin, accepts values labeled with units.
+  - `preferCSSPageSize` <[boolean]> Give any CSS `@page` size declared in the page priority over what is declared in `width` and `height` or `format` options. Defaults to `false`, which will scale the content to fit the paper size.
+- returns: <[Promise]<[Readable]>> Promise which resolves with a Readable stream
+
+This method is identical to [page.pdf](#pagepdfoptions), except it returns the PDF as a readable stream of binary data. If you are generating very large PDFs, it may be useful to use a stream to avoid high memory usage.
 
 #### page.queryObjects(prototypeHandle)
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -136,6 +136,7 @@
   * [page.content()](#pagecontent)
   * [page.cookies([...urls])](#pagecookiesurls)
   * [page.coverage](#pagecoverage)
+  * [page.createPDFStream([options])](#pagecreatepdfstreamoptions)
   * [page.deleteCookie(...cookies)](#pagedeletecookiecookies)
   * [page.emulate(options)](#pageemulateoptions)
   * [page.emulateIdleState(overrides)](#pageemulateidlestateoverrides)
@@ -162,7 +163,6 @@
   * [page.metrics()](#pagemetrics)
   * [page.mouse](#pagemouse)
   * [page.pdf([options])](#pagepdfoptions)
-  * [page.pdfReadable([options])](#pagepdfreadableoptions)
   * [page.queryObjects(prototypeHandle)](#pagequeryobjectsprototypehandle)
   * [page.reload([options])](#pagereloadoptions)
   * [page.screenshot([options])](#pagescreenshotoptions)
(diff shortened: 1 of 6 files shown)


jackfranklin:
> but both would be a breaking change for TypeScript consumers.

I think we could consider a breaking change to the TS types - as long as we don't break the default behaviour of `page.pdf()` - which is doable. 

Puppeteer is being migrated to TS but we don't ship our own types yet, so we could make this change before we start to ship our own types.

I'll defer on the implementation and desire to add this to the library to @mathiasbynens but just wanted to say I think I'd prefer one method rather than two that are nearly identical.

bmisiak:
Great. If backwards compatibility is not a concern, I agree - I'd rather have it as new param on `Page.pdf`. Let me know @mathiasbynens.

mathiasbynens:
cc @janscheffler who had some thoughts on streaming APIs in Puppeteer recently. Jan, does this align with what you had in mind for downloads?

mathiasbynens:
cc @jschfflr who had some thoughts on streaming APIs in Puppeteer recently. Jan, does this align with what you had in mind for downloads?

jschfflr:
I also think that wrapping the underlying IO domain in a readable stream is the way to go. 
But I also think there is value in not only exposing the readable stream but also writing to a file directly or returning the information as a single buffer. Node.js for example has `fs.readFile` and `fs.createReadStream` for that. 
And I think, all methods that deal with an IO stream internally should follow the same conventions. WDYT?

bmisiak:
For sure, sounds good.

I already created `readableProtocolStream`:
https://github.com/bmisiak/puppeteer/blob/85a41501302944b631352a40f3840abc32c21065/src/helper.ts#L196-L215

Then it is just a matter of, within `Page.pdf()`, interpreting the params and consuming the stream if we need to save to file or output a `Buffer`. An elegant solution for saving to a file at `path` is:
```
await utils.pipeline(
	readableProtocolStream(...),
	fs.createWriteStream(path)
)
```
and outputting a `Buffer`:
```
let chunks = []
for await (const chunk of readableProtocolStream(...)) {
	chunks.push(chunk)
}
return Buffer.concat(chunks)
```
... which can be a part of the pipeline as a generator.

To remain backwards compatible we will want to always return a `Buffer` unless a readable stream was explicitly requested.

I'll update the PR to have a single `Page.pdf()` returning `Readable` when requested and `Buffer` otherwise. Thanks!

bmisiak:
Updated. `Page.pdf` now takes an optional `asReadable: boolean` and returns `Promise<Buffer|Readable>`. Please let me know if you have any suggestions w.r.t. naming etc.

Edit:
Actually, I have an idea which is going to work better with TypeScript, allowing the compiler to narrow down the output type automatically:
image

Essentially, we'd add an `output` param which could take `string | PDFOutput.Buffer | PDFOutput.ReadableStream`. It would default to the old `path`s value, or if it is absent, `PDFOutput.Buffer`.

As a result:
* When `output` is `PDFOutput.ReadableStream`, return type is Readable. `path` can not be a string in this scenario.
* Otherwise, return type is Buffer. If `output` or `path` are a `string`, treat it as path and write to a file.

This won't even break TypeScript users, which is great.

jschfflr:
Thanks for already implementing that. 
I think we should not only look at how page.pdf handles this but also take a look at the bigger picture and get a common understanding on how methods working with IO streams should work in general. At the moment, I see two different approaches:

**Based on a parameter**
```
const buffer = await page.pdf({output: puppeteer.Output.Buffer}); // returns a Buffer
const stream = page.pdf({output: puppeteer.Output.Stream}); // returns a ReadableStream
page.pdf({output: puppeteer.Output.File, path: '...'}); // writes to a file specified by path

const buffer = await page.tracing.start({output: puppeteer.Output.Buffer}); // returns a Buffer
const stream = page.tracing.start({output: puppeteer.Output.Stream}); // returns a ReadableStream
page.tracing.start({output: puppeteer.Output.File, path: '...'}); // writes to a file specified by path
```

**Based on method name**
```
const buffer = await page.pdfAsBuffer(); 
const stream = page.pdfAsStream();
page.pdfToFile('...'); 

const buffer = await page.tracing.startAsBuffer();
const stream = page.tracing.startAsStream();
page.tracing.startToFile('...');
```

The exact naming is of course up for discussion. Let me know what you think!
@mathiasbynens Do you have an opinion on that?

mathiasbynens:
I weakly prefer explicit method names (e.g. having both `pdf()` + `pdfAsBuffer()`) over an additional option.

Another possibility is to do something similar to the `fetch()` API: change `page.pdf()` and `page.tracing.start()` to resolve with an instance/object on which the user can call `toString()`, `.toBuffer()`, `toReadable()` (again, naming TBD). I think this would be ideal. Note that we _could_ even do this as a non-breaking change: `page.pdf()` currently always returns a `Promise<Buffer>`, so we could install the `toReadable()` on that object. Returning a dedicated non-Buffer instance seems cleaner though.

Existing usage:

```js
// Write the file to disk, I don't care about the return value:
await page.pdf({ path: '/path/to/save/the.pdf' });
// Give me the PDF as a buffer:
const buffer = await page.pdf();
```

With the above proposal, this would become:

```js
// Write the file to disk, I don't care about the return value:
await page.pdf({ path: '/path/to/save/the.pdf' });
{ // Give me the PDF as a buffer:
  const pdf = await page.pdf();
  const buffer = await pdf.toBuffer();
}
{  // Give me the PDF as a ReadableStream:
  const pdf = await page.pdf();
  const readable = await pdf.toReadable();
}
```

I'm still not sure whether I prefer that or just separate `pdf`/`pdfAsBuffer`/`pdfAsReadable` methods. What do you think, @bmisiak @jschfflr?

jschfflr:
I like your idea about the fetch approach. The underlying implementation does rely on the CDP IO domain. So maybe those functions could return an instance of CDPStream which then has methods like toBuffer or toFile. 
In that case, CDPStream could already implement Readable or have a toStream method. 
Having CDPStream implement Readable would probably be more straightforward but in theory, CDPStreams also allow non sequential reads that would not be representable with a Readable.

So what do you think about the following solution: 

``` TypeScript
type StreamHandle = string

class CDPStream {
  public readonly handle: StreamHandle
  constructor(handle: StreamHandle) { ... }
  
  // return Buffer or null in case of eof
  read(offset: number, size: number): Promise<Buffer|null> { ... }
  close(): Promise<void> { ... }

  toReadable(): Readable { ... }
  toBuffer(): Promise<Buffer> { ... }
  // Writes data a file, promise is resolved when all data has been written
  toFile(file: string): Promise<void> { ... }
}

```

and then

``` TypeScript

const stream = await page.pdf( ... );
const buffer = await stream.toBuffer();

// This should fail because stream is already converted to Buffer
const readable = await buffer.toStream();

```

The only thing I don't like about this solution is that `stream` can now refer to either the CDP one or the Node.js one. 
Any thoughts on this approach?

mathiasbynens:
@bmisiak what do you think of @jschfflr's comments?

shmilyoo:
My PDF generation service is taking up a lot of memory, mostly buffer/cache.
I hope you can provide a method to return the pdf stream.

micahthomas:
@bmisiak @jschfflr 

Wouldn't that break the functionality for existing users?

I think the `asReadable` option and preserving the current API & Types, and just extending it is great. I don't think the benefit of holding this for CDPStream, and breaking the API is worth it.
(you mention non-sequential reads, is this something that's been requested?)

I would love to use this feature in our current application, and hope to see this move forward.

summera:
@bmisiak @jschfflr any chance this will be moving forward? It would be a great addition.

bmisiak:
Thanks for the pig. I got busy with other work, but would love to get this over the finish line, as I am happy to report we have been successfully using this feature at my current company since I posted the PR.

@mathiasbynens 
> `page.pdf()` currently always returns a `Promise<Buffer>`, so we could install the toReadable() on that object.

I am afraid this would defeat the purpose. If we concatenate into a `Buffer` in `pdf()`, getting a `Readable` from it will be pointless as the entirety of the PDF will already be in memory. The point of having the readable stream instead is to avoid having to allocate enough memory for an entire PDF, and limit memory consumption to a small single chunk at a time, consuming them piece by piece.

This proposal:
```js
  // Give me the PDF as a ReadableStream:
  const pdf = await page.pdf();
  const readable = await pdf.toReadable();
```
Is desirable, but it would break backwards compatibility, as `page.pdf()` could no longer return a concatenated Buffer like it does now, if the goal of reducing memory consumption was to be achieved. It would instead have to return a structure which would only contain a reference to the stream. Then, a `toBuffer` could consume the stream and turn it into a `Buffer`, and `toFile` could consume it by writing to a file.

Because it breaks backwards compat, it doesn't seem like a great idea.
We can easily add `page.pdfAsReadable()` instead of an option in `page.pdf()` though!

jschfflr:
I'm more than happy to go ahead with this but I'd prefer to not change the return type based on an option and instead use different method names. WDYT about adding a new method called `Page.createPDFStream` (in alignment with the `fs.createWriteStream` from Node.js) and have it return that stream?
This way, the functionality will also be easily discoverable by new users.

bmisiak:
Perfect! I've had to use the fork for a year now so I would be happy to get
this updated to `createPDFStream` and merged. I'll update the PR soon.

Am Mo., 12. Apr. 2021 um 01:45 Uhr schrieb Jan Scheffler <
***@***.***>:

ghost:
@bmisiak how's it going? Thanks for this great pr! 🎉

jschfflr:
@bmisiak Gentle ping to see if there any updates on this?

Vijei:
@bmisiak, will you be able to merge this PR anytime soon? waiting for it. Thank you so much.

jschfflr:
@googlebot I consent.

jschfflr:
@bmisiak I took the liberty to rebase this pull request and to include the proposed changes to get this landed.

OrKoN (approved):

docs/api.md
```diff
@@ -1489,6 +1490,35 @@ If URLs are specified, only cookies for those URLs are returned.
...
+  - `margin` <[Object]> Paper margins, defaults to none.
+    - `top` <[string]|[number]> Top margin, accepts values labeled with units.
+    - `right` <[string]|[number]> Right margin, accepts values labeled with units.
+    - `bottom` <[string]|[number]> Bottom margin, accepts values labeled with units.
+    - `left` <[string]|[number]> Left margin, accepts values labeled with units.
+  - `preferCSSPageSize` <[boolean]> Give any CSS `@page` size declared in the page priority over what is declared in `width` and `height` or `format` options. Defaults to `false`, which will scale the content to fit the paper size.
+  - `omitBackground` <[boolean]> Hides default white background and allows capturing screenshots with transparency. Defaults to `false`.
+- returns: <[Promise]<[Readable]>> Promise which resolves with a Node.js stream for the PDF file.
+
+> **NOTE** This method is identical to [page.pdf](#pagepdfoptions), except it returns the PDF as a readable stream of binary data. If you are generating very large PDFs, it may be useful to use a stream to avoid high memory usage. This version will ignore the `path` option.
```
mathiasbynens, line 1520: Could we add a code example?

src/common/Page.ts
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import type { Readable } from 'stream';
```
mathiasbynens, line 17: Does this impact agnostification? @jackfranklin

src/common/helper.ts
```diff
@@ -318,35 +320,56 @@ async function readProtocolStream(
...
     return resultBuffer;
   }
 }
 
+async function getReadableFromProtocolStream(
+  client: CDPSession,
+  handle: string
+): Promise<Readable> {
+  if (!isNode) {
+    throw new Error('Cannot create a stream outside of Node.js environment.');
```
mathiasbynens, line 352: In the future we can solve this with https://github.com/nodejs/node/pull/39062. Perhaps add a TODO with a link?

src/common/Page.ts
jschfflr: I think this should be fine because the `import type` will not be included in the output of the ts compiler. But @jackfranklin  might know more about that :)

jackfranklin: Yup, `type` is OK :)

jschfflr: Awesome, thanks! Do we have something in the try bots to check if puppeteer-common is indeed agnostic to the platform?

Vijei:
@jschfflr , Thank you so much!. could you pls let me know which version it is available

Vijei:
@jschfflr @mathiasbynens  - Any tentative date when this feature will be available

jschfflr:
@Vijei Sorry for the delay, I had hoped to include some other features in the release too.

jschfflr:
@Vijei The change has landed with the latest version.

Vijei:
@jschfflr Great! Thank you so much.

davidchai717:
@jschfflr You are doing God's work. thank you so much!!!

jschfflr:
Happy to be helpful!

Vijei:
@jschfflr , Hi am able to create large pdf files ~21MB in my local. but not able to create the same in docker. I tried both options. file around 7MB is created in docker. 
1) page.pdf
2) page.createpdfstream.
I dont see any errors as well. any pointers?

jschfflr:
Can you share your code and describe what happens when you try to run it in docker?

Vijei:
Hi @jschfflr , thanks for the reply.

I didn't see any error and it just hangs. same working without docker. Actually am trying to write around 60k records getting 500 records per batch from backend. we will get all the records, create the html and use setcontent method and page.pdf/page.createPDFStream for pdf creation.

Also one more question
I just wonder do we have anything in puppeteer similar to pdfkit (https://pdfkit.org/docs/getting_started.html) which writes content to file/response for every fetch instead of getting all records, then create html and use setcontent method.

jschfflr:
@Vijei I'm not sure I understand your use case correctly. Could you share a quick example that reproduces the problem?
I'd be happy to look into it to get this issue resolved :)

Vijei:
Here is the code
```js
// before this I will get the records from backend 500/batch. total 60k records and frame the pdfHtml
    headless: true,
    executablePath: process.env.CHROME_BIN || null,
    args: ['--no-sandbox', '--headless', '--disable-gpu', '--disable-dev-shm-usage']
});
const pdfPage = await browser.newPage();
await pdfPage.setDefaultNavigationTimeout(0);
await pdfPage.setContent(pdfHtml);
const pdfStream = await pdfPage.createPDFStream(CONSTANTS.PDF_OPTIONS);
const fileName = 'test.pdf';
response.set({
    "Access-Control-Allow-Origin": "*",
    'Content-Type': 'application/pdf',
    'Content-disposition': 'attachment; filename='+fileName,
});
pdfStream.pipe(response);
```

jschfflr:
@Vijei Thanks for sharing your code! Unfortunately, it looks like it's only a small part of your actual code so I can't use it to reproduce your issue. I would really love to fix the problem (as long as it is within puppeteer) but without a minimal reproduction, there's nothing I can do at this point. 
Can you confirm that both `page.pdf` and `page.createPDFStream` are working without docker for you and only start failing when you try to use them in a docker container?

Vijei:
@jschfflr , yes both works fine without docker.

Also one more question
I just wonder do we have anything in puppeteer similar to pdfkit (https://pdfkit.org/docs/getting_started.html) which writes content to file/response for every fetch instead of getting all records, then create html and use setcontent method.

Issue #5719, opened 2020-04-22
Puppeteer not working with node 14


DRoet:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version:3.0.1
* Platform / OS version:docker (FROM:node:14.0.0-buster)
* URLs (if applicable):
* Node.js version: 14.0.0

**What steps will reproduce the problem?**

_Please include code that reproduces the issue._

1. Upgrade node image from 13.13.0 -> 14.0.0
2. Run same codebase (`puppeteer.launch()`)

**What is the expected result?**
code runs without issues, just like on 13.13.0

**What happens instead?**
`Error: Failed to launch the browser process! spawn /opt/app/node_modules/puppeteer/.local-chromium/linux-737027/chrome-linux/chrome ENOENT`

munrocket:
@DRoet only with node 14? Try `3.0.1-post`

munrocket:
Try `3.0.1-post`

DRoet:
> @DRoet only with node 14?

yes I just rebuild my images using `12.16.2` and that version also still works

DRoet:
> Try `3.0.1-post`

how do I install this? doesn't seem to be a valid tag on the registry

jsoverson:
It looks like it's a problem with extraction. A temporary workaround is to extract the downloaded chromium manually:

```shell
$ mv node_modules/puppeteer/.local-chromium/mac-737027 node_modules/puppeteer/.local-chromium/mac-737027.bak 
$ unzip node_modules/puppeteer/.local-chromium/chrome-mac.zip -d node_modules/puppeteer/.local-chromium/mac-737027
```

nstepien:
Likely culprit: https://github.com/maxogden/extract-zip/issues/94

jackfranklin:
Thanks for the debugging all; looks like `extract-zip` is indeed the culprit here. It would be nice if Puppeteer gave a more helpful message than `/opt/app/node_modules/puppeteer/.local-chromium/linux-737027/chrome-linux/chrome ENOENT` so I'll try to look into that.

We'll see if a fix lands in extract-zip and do some digging.

@DRoet you can install via the master branch on GitHub to get the very latest.

jackfranklin:
It also looks like the suspected bug might have been in Node: https://github.com/nodejs/node/pull/32968.

In which case we may have to sit tight for a Node patch release. In the mean time I'll see if the error message from Puppeteer can be nicer.

[...]

PR #5732, opened 2020-04-24
chore: log useful error for Node v14 breakage

jackfranklin:
There is currently a bug with extract-zip and Node v14.0.0 that
causes extractZip to silently fail:
https://github.com/puppeteer/puppeteer/issues/5719

Rather than silenty fail if the user is on Node 14 we instead
detect that and throw an error directing the user to that bug. The
rejection message below is surfaced to the user in the command line.

The issue seems to be in streams never resolving so we wrap the call in
a timeout and give it 100ms to resolve before deciding on an error. If
the user is on Node < 14 we maintain the behaviour we had before this
patch.

Here's how this change impacts the output on Node 14 and Node 10:

## Node 10:

```
npm run tsc && rm -r .local-* && node install

> puppeteer@3.0.1-post tsc /Users/jacktfranklin/src/puppeteer
> tsc --version && tsc -p . && cp src/protocol.d.ts lib/ && cp src/externs.d.ts lib/

Version 3.8.3
Downloading Chromium r737027 - 118.4 Mb [====================] 100% 0.0s
Chromium (737027) downloaded to /Users/jacktfranklin/src/puppeteer/.local-chromium/mac-737027
```

## Node 14 without this patch:

```
npm run tsc && rm -r .local-* && node install

> puppeteer@3.0.1-post tsc /Users/jacktfranklin/src/puppeteer
> tsc --version && tsc -p . && cp src/protocol.d.ts lib/ && cp src/externs.d.ts lib/

Version 3.8.3
Downloading Chromium r737027 - 118.4 Mb [====================] 100% 0.0s
```

Note that whilst it doesn't error, it doesn't complete the install. We
don't get the success message that we saw above in the Node 10 install.

## Node 14 with this patch:

```npm run tsc && rm -r .local-* && node install

> puppeteer@3.0.1-post tsc /Users/jacktfranklin/src/puppeteer
> tsc --version && tsc -p . && cp src/protocol.d.ts lib/ && cp src/externs.d.ts lib/

Version 3.8.3
Downloading Chromium r737027 - 118.4 Mb [====================] 100% 0.0s
ERROR: Failed to set up Chromium r737027! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
Puppeteer currently does not work on Node v14 due to an upstream bug. Please see https://github.com/puppeteer/puppeteer/issues/5719 for details.
```

The explicit message should save users a good amount of debugging time.

### DIFF
--- a/src/BrowserFetcher.ts
+++ b/src/BrowserFetcher.ts
@@ -321,10 +321,43 @@ function install(archivePath: string, folderPath: string): Promise<unknown> {
 }
 
 async function extractZip(zipPath: string, folderPath: string): Promise<void> {
-  try {
-    await extract(zipPath, {dir: folderPath});
-  } catch (error) {
-    return error;
+  const nodeVersion = process.version;
+
+  /* There is currently a bug with extract-zip and Node v14.0.0 that
+   * causes extractZip to silently fail:
+   * https://github.com/puppeteer/puppeteer/issues/5719
+   *
+   * Rather than silenty fail if the user is on Node 14 we instead
+   * detect that and throw an error directing the user to that bug. The
+   * rejection message below is surfaced to the user in the command
+   * line.
+   *
+   * The issue seems to be in streams never resolving so we wrap the
+   * call in a timeout and give it 100ms to resolve before deciding on
+   * an error.
+   *
+   * If the user is on Node < 14 we maintain the behaviour we had before
+   * this patch.
+   */
+  if (nodeVersion === 'v14.0.0') {
+    await new Promise((resolve, reject) => {
+      let extractedResolved = false;
+      setTimeout(async() => {
+        await extract(zipPath, {dir: folderPath});
+        extractedResolved = true;
+      }, 10);
+
+      if (extractedResolved)
+        resolve();
+      else
+        reject(`Puppeteer currently does not work on Node v14 due to an upstream bug. Please see: https://github.com/puppeteer/puppeteer/issues/5719 for details.`);
+    });
+  } else {
+    try {
+      await extract(zipPath, {dir: folderPath});
+    } catch (error) {
+      return error;
+    }
   }
 }
 
--- a/src/BrowserFetcher.ts
+++ b/src/BrowserFetcher.ts
@@ -339,19 +339,21 @@ async function extractZip(zipPath: string, folderPath: string): Promise<void> {
    * If the user is on Node < 14 we maintain the behaviour we had before
    * this patch.
    */
-  if (nodeVersion === 'v14.0.0') {
-    await new Promise((resolve, reject) => {
-      let extractedResolved = false;
-      setTimeout(async() => {
-        await extract(zipPath, {dir: folderPath});
-        extractedResolved = true;
-      }, 10);
+  if (nodeVersion.startsWith('v14.')) {
+    let timeoutReject;
+    const timeoutPromise = new Promise((resolve, reject) => { timeoutReject = reject; });
 
-      if (extractedResolved)
-        resolve();
-      else
-        reject(`Puppeteer currently does not work on Node v14 due to an upstream bug. Please see: https://github.com/puppeteer/puppeteer/issues/5719 for details.`);
-    });
+    const timeoutToken = setTimeout(() => {
+      const error = new Error(`Puppeteer currently does not work on Node v14 due to an upstream bug. Please see: https://github.com/puppeteer/puppeteer/issues/5719 for details.`);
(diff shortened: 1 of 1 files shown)


src/BrowserFetcher.ts
```diff
@@ -321,10 +321,43 @@ function install(archivePath: string, folderPath: string): Promise<unknown> {
...
+   * line.
+   *
+   * The issue seems to be in streams never resolving so we wrap the
+   * call in a timeout and give it 100ms to resolve before deciding on
+   * an error.
+   *
+   * If the user is on Node < 14 we maintain the behaviour we had before
+   * this patch.
+   */
+  if (nodeVersion === 'v14.0.0') {
```
mathiasbynens: `nodeVersion.startsWith('v14.')`?

mathiasbynens: Or do you think this is something Node.js will ship a patch for?

jackfranklin: I was hopeful that Node will ship a patch for it (given https://github.com/nodejs/node/pull/32968).

But it could be safer to default to starts with `v14.` for now and then we can just ship a new release when the problem is solved, so let's do that 👍

src/BrowserFetcher.ts
```diff
@@ -321,10 +321,43 @@ function install(archivePath: string, folderPath: string): Promise<unknown> {
...
+      if (extractedResolved)
+        resolve();
+      else
+        reject(`Puppeteer currently does not work on Node v14 due to an upstream bug. Please see: https://github.com/puppeteer/puppeteer/issues/5719 for details.`);
+    });
+  } else {
+    try {
+      await extract(zipPath, {dir: folderPath});
+    } catch (error) {
+      return error;
```
dr-js, line 361: Why not throw the error here but return it?
I think this will mask error from `extract`

jackfranklin: Yup, this is a problem, but this code path is the codepath we already ship and changing this unfortunately doesn't solve the problem we're seeing (the problem is that extract-zip is failing silently).

I have this on my list to fix as you're right, it silences errors, but this code has been like this for a long time so I'll leave it for now.

src/BrowserFetcher.ts
```diff
@@ -321,10 +321,43 @@ function install(archivePath: string, folderPath: string): Promise<unknown> {
...
+      setTimeout(async() => {
+        await extract(zipPath, {dir: folderPath});
+        extractedResolved = true;
+      }, 10);
+
+      if (extractedResolved)
+        resolve();
+      else
+        reject(`Puppeteer currently does not work on Node v14 due to an upstream bug. Please see: https://github.com/puppeteer/puppeteer/issues/5719 for details.`);
+    });
```
dr-js: Current code will always reject.
I think the code should be something like:
```suggestion
    let timeoutReject;
    const timeoutPromise = new Promise((resolve, reject) => { timeoutReject = reject })
    const timeoutToken = setTimeout(
      () => timeoutReject(new Error(`Puppeteer currently does not work on Node v14 due to an upstream bug. Please see: https://github.com/puppeteer/puppeteer/issues/5719 for details.`)),
      10 * 1000 // wait 10 second for 200MB zip to finish
    );
    await Promise.race([
      extract(zipPath, {dir: folderPath}),
      timeoutPromise
    ]);
    clearTimeout(timeoutToken);
```

jackfranklin: ah nice catch, thanks 👍

munrocket:
Yes this is much better, but IMHO puppeteer should work like an axe in any OS and nodejs.

Don’t know what’s happened after 2.2.1 but in future better to avoid such bugs.
For example in my repo every contributor should make a screenshot after PR, I don’t want help to everybody how to install packages with npm :D

mathiasbynens (approved):

src/BrowserFetcher.ts
```diff
@@ -321,10 +321,45 @@ function install(archivePath: string, folderPath: string): Promise<unknown> {
...
+   * causes extractZip to silently fail:
+   * https://github.com/puppeteer/puppeteer/issues/5719
+   *
+   * Rather than silenty fail if the user is on Node 14 we instead
+   * detect that and throw an error directing the user to that bug. The
+   * rejection message below is surfaced to the user in the command
+   * line.
+   *
+   * The issue seems to be in streams never resolving so we wrap the
+   * call in a timeout and give it 100ms to resolve before deciding on
```
mathiasbynens: 100ms or 10 seconds?

PR #5733, opened 2020-04-24
fix: make uploadFile throw for non-existent files

paullewis:


src/JSHandle.ts
```diff
@@ -291,7 +291,24 @@ export class ElementHandle extends JSHandle {
...
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const {promisify} = require('util');
+    const access = promisify(fs.access);
+
+    // Locate all files and confirm that they exist.
+    const files = await Promise.all(filePaths.map(async filePath => {
+      const resolvedPath: string = path.resolve(filePath);
+      try {
+        await access(resolvedPath);
+      } catch (e) {
```
mathiasbynens: nit: e -> error

src/JSHandle.ts
```diff
@@ -291,7 +291,24 @@ export class ElementHandle extends JSHandle {
     // the cost unnecessarily.
     // eslint-disable-next-line @typescript-eslint/no-var-requires
     const path = require('path');
-    const files = filePaths.map(filePath => path.resolve(filePath));
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const fs = require('fs');
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+    const {promisify} = require('util');
+    const access = promisify(fs.access);
```
mathiasbynens, line 298: Should we use `fs.constants.R_OK` to ensure the file doesn't just exist but also that it is readable? WDYT?

mathiasbynens (changes requested):
LGTM % question + nit

src/JSHandle.ts
paullewis: Done

src/JSHandle.ts
paullewis: Done

Issue #5746, opened 2020-04-26
Can the explicit locator strategy be supported?


mokeyish:
<img width="475" alt="图片" src="https://user-images.githubusercontent.com/16131917/80323134-193cb880-885c-11ea-9060-cbb85b6cd541.png">

mokeyish:
**Here is my temporary implementation**


<img width="915" alt="图片" src="https://user-images.githubusercontent.com/16131917/80323392-ad5b4f80-885d-11ea-961d-f7e34a35bcd7.png">

PR #5747, opened 2020-04-27
feat(puppeteer): impl DOMWorld.$ for `xpath:` selector

mokeyish:
#5746 

**explicit locator strategy**

图片

### DIFF
--- a/src/DOMWorld.js
+++ b/src/DOMWorld.js
@@ -129,8 +129,8 @@ class DOMWorld {
    */
   async $(selector) {
     const document = await this._document();
-    if (selector.startsWith('xpath:')) {
-      const handles = await document.$x(selector);
+    if (selector && selector.startsWith('xpath:')) {
+      const handles = await document.$x(selector.substring(6));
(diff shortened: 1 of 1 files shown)


mokeyish:
@googlebot I signed it!

mathiasbynens (changes requested):
This would need documentation as well. 

Either way, we're currently investigating adding this functionality more generically, so that folks can register custom query handlers and then opt-in to them for any part of the API that accepts a selector. It seems like opting in to XPath (as opposed to using [`page.$x`](https://github.com/puppeteer/puppeteer/blob/master/docs/api.md#pagexexpression)) would be a good fit.

mathiasbynens:
cc @paullewis

mathiasbynens:
We decided to go with custom query handlers instead. The implementation is currently experimental, and you can read about it here: https://github.com/puppeteer/puppeteer/releases/tag/v3.0.3 It should enable you to build what you had in mind! Please give it a try and let us know if you run into any issues.

PR #5748, opened 2020-04-27
chore: use Node's promisify function

jackfranklin:
Node's promisify function and the TS types for it give much better type
understanding when wrapping a function in `promisify`. This change means
we don't maintain our own implementation and our own (sub-par) types and
rather lean on the tested and thorough @types/node version instead.


jackfranklin:
This will pass once #5750 is merged + this PR updated.

mathiasbynens (approved):
Nice cleanup!

PR #5750, opened 2020-04-27
chore: remove puppeteer-web

jackfranklin:
We don't support it and v3 shipped without including puppeteer-web in
the browser. People are welcome to manually use Browserify to try to get
Puppeteer running in a browser but it ultimately isn't our primary focus
right now.

Getting puppeteer-core able to run in a browser is something we'll be
looking at in the future so we'll revisit this soon.


mathiasbynens (approved):

entrptaher:
I tried to implement it back, but the reasons to remove this feature is pretty obvious, lots of filesystem related calls are inside latest puppeteer which is a pain to support.

If anyone is interested, here is my version, https://github.com/entrptaher/puppeteer-web

PR #5752, opened 2020-04-27
chore: mark version v3.0.2

jackfranklin:


mathiasbynens (approved):

mathiasbynens:
That CI failure is interesting...

```
Coverage check failed: not all API methods called. See above ouptut for list of missing methods.
accessibility.snapshot
```

jackfranklin:
I think what happened is two tests errored for some unknown reason (annoyingly the output seems to have been lost) and then the coverage failure happened as a result. Seeing if a rebuild solves it.

Issue #5405, opened 2020-02-10
Support for custom selector engines (Querying nested shadow roots)


Georgegriff:
**What is this?**
Other tools offer the ability to provide a custom `engine` selecting elements in the DOM. 

(example from playwright)
```
  await playwright.selectors.register(selectorEngine, { name: 'shadow' })
  await page.waitForSelector('shadow=#no-downloads span', {timeout: 3000}
```

I've seen https://github.com/puppeteer/puppeteer/issues/382 but i'm not sure this offers such an easy mechanism thats really simple for end users. I'm happy to be wrong on this, i couldn't find any examples.

Playwright offers this. https://github.com/microsoft/playwright/blob/master/docs/api.md#selectorsregisterenginefunction-args

In Selenium world its things like this: https://chercher.tech/java/custom-locators-selenium-webdriver

**Why is this useful?**
Traditionally these custom locators would be used to provide the ability to select elements via XPATH or JQuery selectors.

**Why do i want this?**
I maintain:  https://github.com/Georgegriff/query-selector-shadow-dom which allows users to write css selectors that automatically pierce web component shadow roots and it was trivial to add support in `Playwright` to use my library as a selector engine.
Like so:

```
const { selectorEngine } = require("query-selector-shadow-dom/plugins/playwright");
const playwright = require('playwright')
  await playwright.selectors.register(selectorEngine, { name: 'shadow' })

  const browser = await playwright.chromium.launch({ headless: false})
  const context = await browser.newContext({ viewport: null })
  const page = await context.newPage()

  await page.goto('chrome://downloads')

  await page.waitForSelector('shadow=#no-downloads span', {timeout: 3000})
  await new Promise(resolve => setTimeout(resolve, 3000))   

  await page.close()
  await context.close()
  await browser.close()
```
 Registering this engine allows users to use `click` `waitForSelector` and thing that accepts a selector to use my library to automatically pierce shadow roots.

**How is my engine implemented in playwright?**

Playwright defines this interface: https://github.com/microsoft/playwright/blob/master/docs/api.md#selectorsregisterenginefunction-args which accepts a Function/String
They will take your function and pass into into the browser context and handle the rest for you so you can use the engine for `click` etc.

My library implements this interface: https://github.com/Georgegriff/query-selector-shadow-dom/blob/master/plugins/playwright/index.js (It does this a little strangely using string because i need to inject my library into the function scope)

paullewis:
[...]

Issue #4171, opened 2019-03-15
Better Support for Shadow DOM


aslushnikov:
Since `Copy JS Path` has landed in Chrome DevTools, we now have a way to address nodes inside shadow DOM.

Details: https://github.com/GoogleChrome/puppeteer/issues/858#issuecomment-438540596

This, however, doesn't simplify life much when it comes to sugar methods, such as `page.click`, `page.select` and others.

It'd be nice to have a way to address nodes in shadow DOM in these methods.

Suggestions:
1. Teach `page.click` and others to accept an array of selectors. ([source](https://github.com/GoogleChrome/puppeteer/issues/858#issuecomment-466186707)). 

```js
await page.select(['#container', 'select#foo'], 'value');
```

2. Teach `page.click` and others to treat jsPaths as selectors. 

```js
await page.select("document.querySelector('body > toolbar-component > toolbar-section.left')", 'value');
```

cc @JoelEinbinder

felixfbecker:
Another thing to consider that in an application where every (web)component is a shadow root, this would basically be equivalent to an _exact_ step by step DOM path instead of a "selector". This means you would have to update _all_ your e2e tests _every time_ you change _anything_ about the DOM structure, which would be extremely annoying (no matter if it supports JS path or arrays).

Maybe another better option would be to have an option `{ penetrateShadowRoot: true }` to the functions that would recursively dig into shadow roots.

Jamesernator:
Could the `>>>` selector be reintroduced but only allowed via the devtools protocol?

felixfbecker:
Btw, how do XPath queries behave with regards to shadow roots?

Jamesernator:
Also with the new introduction of `::part` it would make sense if `page.click()` and friends accepted an option to select a specific `::part`.

JoelEinbinder:
```js
await page.click(() => window.mySuperCoolElement)
```

elf-pavlik:
Maybe `page.waitForSelector(selector[, options])` could also take JS Path?
How one supposed currently to wait for selectors in shadow dom?

Georgegriff:
I feel like most of solutions to shadow Dom in puppeteer force the tests/tester to have to know far too much about the structure of the application. 

Having page.click work with jsPath would make a library I wrote a bit easier to work with. (Below)

The purpose of the library is to allow you to just write a normal querySelector and it will automatically find the selector whether it's in a shadow root or deeply nested shadow roots.

[...]

PR #5753, opened 2020-04-27
feat: add support for string-based custom queries

paullewis:
Relevant issues: #5747, #5405, #4171.

### DIFF
--- a/src/DOMWorld.ts
+++ b/src/DOMWorld.ts
@@ -384,7 +384,7 @@ export class DOMWorld {
     } = options;
     const polling = waitForVisible || waitForHidden ? 'raf' : 'mutation';
     const title = `${isXPath ? 'XPath' : 'selector'} "${selectorOrXPath}"${waitForHidden ? ' to be hidden' : ''}`;
-    const { updatedSelector, queryHandler} = getQueryHandlerAndSelector(selectorOrXPath, (element, selector) => document.querySelector(selector));
+    const {updatedSelector, queryHandler} = getQueryHandlerAndSelector(selectorOrXPath, (element, selector) => document.querySelector(selector));
     const waitTask = new WaitTask(this, predicate, queryHandler, title, polling, timeout, updatedSelector, isXPath, waitForVisible, waitForHidden);
     const handle = await waitTask.promise;
     if (!handle.asElement()) {
@@ -444,11 +444,18 @@ class WaitTask {
     else
       throw new Error('Unknown polling options: ' + polling);
 
-    const wrapper =`
-    return (function wrapper(args) {
-      ${ predicateQueryHandlerBody ? `const predicateQueryHandler = ${predicateQueryHandlerBody};` : ''}
-      return (${predicateBody})(...args);
-    })(args)`;
+    function getPredicateBody(predicateBody: Function | string, predicateQueryHandlerBody: Function | string) {
+      if (helper.isString(predicateBody))
+        return `return (${predicateBody});`;
+      if (predicateQueryHandlerBody) {
+        return `
+          return (function wrapper(args) {
+            const predicateQueryHandler = ${predicateQueryHandlerBody};
+            return (${predicateBody})(...args);
+          })(args);`;
+      }
+      return `return (${predicateBody})(...args);`;
+    }
 
     this._domWorld = domWorld;
     this._polling = polling;
--- a/src/DOMWorld.ts
+++ b/src/DOMWorld.ts
@@ -369,7 +369,7 @@ export class DOMWorld {
       polling = 'raf',
       timeout = this._timeoutSettings.timeout(),
     } = options;
-    return new WaitTask(this, pageFunction, '', 'function', polling, timeout, ...args).promise;
+    return new WaitTask(this, pageFunction, undefined, 'function', polling, timeout, ...args).promise;
   }
 
   async title(): Promise<string> {
@@ -436,7 +436,7 @@ class WaitTask {
   _timeoutTimer?: NodeJS.Timeout;
   _terminated = false;
 
-  constructor(domWorld: DOMWorld, predicateBody: Function | string, predicateQueryHandlerBody: Function | string = '', title: string, polling: string | number, timeout: number, ...args: unknown[]) {
+  constructor(domWorld: DOMWorld, predicateBody: Function | string, predicateQueryHandlerBody: Function | string | undefined, title: string, polling: string | number, timeout: number, ...args: unknown[]) {
     if (helper.isString(polling))
       assert(polling === 'raf' || polling === 'mutation', 'Unknown polling option: ' + polling);
     else if (helper.isNumber(polling))
--- a/src/DOMWorld.ts
+++ b/src/DOMWorld.ts
@@ -453,7 +453,7 @@ class WaitTask {
     this._domWorld = domWorld;
     this._polling = polling;
     this._timeout = timeout;
-    this._predicateBody = helper.isString(predicateBody) ? 'return (' + predicateBody + ')' : wrapper;
+    this._predicateBody = getPredicateBody(predicateBody, predicateQueryHandlerBody);
     this._args = args;
     this._runCount = 0;
     domWorld._waitTasks.add(this);
--- a/src/DOMWorld.ts
+++ b/src/DOMWorld.ts
@@ -23,6 +23,11 @@ import {ExecutionContext} from './ExecutionContext';
 import {TimeoutSettings} from './TimeoutSettings';
 import {MouseButtonInput} from './Input';
 import {FrameManager, Frame} from './FrameManager';
+import {getQueryHandlerAndSelector, QueryHandler} from './QueryHandler';
+
+// This predicateQueryHandler is declared here so that TypeScript knows about it
+// when it is used in the predicate function below.
+declare const predicateQueryHandler: QueryHandler;
 
 const readFileAsync = helper.promisify(fs.readFile);
 
@@ -364,7 +369,7 @@ export class DOMWorld {
       polling = 'raf',
       timeout = this._timeoutSettings.timeout(),
     } = options;
-    return new WaitTask(this, pageFunction, 'function', polling, timeout, ...args).promise;
+    return new WaitTask(this, pageFunction, '', 'function', polling, timeout, ...args).promise;
   }
 
   async title(): Promise<string> {
@@ -379,7 +384,8 @@ export class DOMWorld {
     } = options;
     const polling = waitForVisible || waitForHidden ? 'raf' : 'mutation';
     const title = `${isXPath ? 'XPath' : 'selector'} "${selectorOrXPath}"${waitForHidden ? ' to be hidden' : ''}`;
-    const waitTask = new WaitTask(this, predicate, title, polling, timeout, selectorOrXPath, isXPath, waitForVisible, waitForHidden);
+    const { updatedSelector, queryHandler} = getQueryHandlerAndSelector(selectorOrXPath, (element, selector) => document.querySelector(selector));
+    const waitTask = new WaitTask(this, predicate, queryHandler, title, polling, timeout, updatedSelector, isXPath, waitForVisible, waitForHidden);
     const handle = await waitTask.promise;
     if (!handle.asElement()) {
       await handle.dispose();
@@ -397,7 +403,7 @@ export class DOMWorld {
     function predicate(selectorOrXPath: string, isXPath: boolean, waitForVisible: boolean, waitForHidden: boolean): Node | null | boolean {
       const node = isXPath
         ? document.evaluate(selectorOrXPath, document, null, XPathResult.FIRST_ORDERED_NODE_TYPE, null).singleNodeValue
-        : document.querySelector(selectorOrXPath);
+        : predicateQueryHandler ? predicateQueryHandler(document, selectorOrXPath) as Element : document.querySelector(selectorOrXPath);
       if (!node)
         return waitForHidden;
       if (!waitForVisible && !waitForHidden)
@@ -430,7 +436,7 @@ class WaitTask {
   _timeoutTimer?: NodeJS.Timeout;
   _terminated = false;
 
-  constructor(domWorld: DOMWorld, predicateBody: Function | string, title: string, polling: string | number, timeout: number, ...args: unknown[]) {
+  constructor(domWorld: DOMWorld, predicateBody: Function | string, predicateQueryHandlerBody: Function | string = '', title: string, polling: string | number, timeout: number, ...args: unknown[]) {
     if (helper.isString(polling))
       assert(polling === 'raf' || polling === 'mutation', 'Unknown polling option: ' + polling);
     else if (helper.isNumber(polling))
@@ -438,10 +444,16 @@ class WaitTask {
     else
       throw new Error('Unknown polling options: ' + polling);
 
+    const wrapper =`
+    return (function wrapper(args) {
+      ${ predicateQueryHandlerBody ? `const predicateQueryHandler = ${predicateQueryHandlerBody};` : ''}
+      return (${predicateBody})(...args);
+    })(args)`;
+
     this._domWorld = domWorld;
     this._polling = polling;
     this._timeout = timeout;
-    this._predicateBody = helper.isString(predicateBody) ? 'return (' + predicateBody + ')' : 'return (' + predicateBody + ')(...args)';
+    this._predicateBody = helper.isString(predicateBody) ? 'return (' + predicateBody + ')' : wrapper;
     this._args = args;
     this._runCount = 0;
     domWorld._waitTasks.add(this);
--- a/src/JSHandle.ts
+++ b/src/JSHandle.ts
@@ -19,7 +19,7 @@ import {ExecutionContext} from './ExecutionContext';
 import {CDPSession} from './Connection';
 import {KeyInput} from './USKeyboardLayout';
 import {FrameManager, Frame} from './FrameManager';
-import {customQueryHandlers} from './QueryHandler';
+import {customQueryHandlers, QueryHandler} from './QueryHandler';
 
 const SEARCH_SEPARATOR = '/';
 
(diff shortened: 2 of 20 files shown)


src/JSHandle.ts
```diff
@@ -426,8 +429,18 @@ export class ElementHandle extends JSHandle {
   }
 
   async $(selector: string): Promise<ElementHandle | null> {
-    const handle = await this.evaluateHandle(
-        (element, selector) => element.querySelector(selector),
+    let handler = (element: Element, selector: string): ElementHandle => element.querySelector(selector) as unknown as ElementHandle;
+    if (selector.indexOf(SEARCH_SEPARATOR) !== -1) {
+      const [name, updatedSelector] = selector.split(SEARCH_SEPARATOR);
```
mathiasbynens: If you're gonna use this to create a custom XPath-based query handler, you'll probably use `/` quite often in your selectors. With this implementation, `updatedSelector` only contains everything between the first `/` and the second `/`. We could instead use the result of `indexOf` with `slice` to get the entire selector.

src/Puppeteer.js
```diff
@@ -144,4 +146,23 @@ module.exports = class {
   createBrowserFetcher(options) {
     return new BrowserFetcher(this._projectRoot, options);
   }
+
+  /**
+   * @param {string} name
+   * @param {!Function} queryFunction
+   */
+  registerCustomQueryFunction(name, queryFunction) {
```
mathiasbynens: bikeshed: `Function` sounds a little generic. How about `Handler`?

test/elementhandle.spec.js
```diff
@@ -248,4 +248,25 @@ describe('ElementHandle specs', function() {
...
     });
   });
+
+  describe('Custom queries', function() {
+    it('should register and unregister', async() => {
+      const {page, puppeteer} = getTestState();
+      await page.setContent('<div id="not-foo"></div><div id="foo"></div>');
+
+      // Register.
+      puppeteer.registerCustomQueryFunction('get-by-id', (element, selector) => document.querySelector(`[id="${selector}"]`));
```
mathiasbynens: I think it'd be nice to limit query handler names to `/^[a-zA-Z]+$/`, matching built-in JavaScript APIs. We could validate this upon registration. WDYT? In this case, it'd be `getById` instead of `get-by-id`.

src/QueryFunction.ts
```diff
@@ -0,0 +1,47 @@
...
+import {ElementHandle} from './JSHandle';
+
+export interface QueryFunction {
+  (element: Element, selector: string): (ElementHandle | null) | ElementHandle[];
+}
+
+const _customQueryFunctions = new Map<string, QueryFunction>();
+
+export function registerCustomQueryFunction(name: string, queryFunction: Function): void {
+  if (_customQueryFunctions.get(name))
```
mathiasbynens: TODO: validate the name here. At a minimum, we should ensure it doesn't contain `SEARCH_SEPARATOR`. But IMHO we should be even more restrictive and only allow `/^[a-zA-Z]+$/`.

src/JSHandle.ts
```diff
@@ -442,10 +455,18 @@ export class ElementHandle extends JSHandle {
    * @return {!Promise<!Array<!ElementHandle>>}
    */
   async $$(selector: string): Promise<ElementHandle[]> {
-    const arrayHandle = await this.evaluateHandle(
-        (element, selector) => element.querySelectorAll(selector),
-        selector
-    );
+    let handler = (element: Element, selector: string): ElementHandle[] => element.querySelectorAll(selector) as unknown as ElementHandle[];
+    if (selector.indexOf(SEARCH_SEPARATOR) !== -1) {
+      const [name, updatedSelector] = selector.split(SEARCH_SEPARATOR);
```
mathiasbynens: Same here.

src/JSHandle.ts
```diff
@@ -442,10 +455,18 @@ export class ElementHandle extends JSHandle {
    * @return {!Promise<!Array<!ElementHandle>>}
    */
   async $$(selector: string): Promise<ElementHandle[]> {
-    const arrayHandle = await this.evaluateHandle(
-        (element, selector) => element.querySelectorAll(selector),
-        selector
-    );
+    let handler = (element: Element, selector: string): ElementHandle[] => element.querySelectorAll(selector) as unknown as ElementHandle[];
+    if (selector.indexOf(SEARCH_SEPARATOR) !== -1) {
```
mathiasbynens: Can we extract this common `indexOf` + `split` logic?

test/elementhandle.spec.js
```diff
@@ -248,4 +248,25 @@ describe('ElementHandle specs', function() {
...
+      const element = await page.$('get-by-id/foo');
+      expect(await page.evaluate(element => element.id, element)).toBe('foo');
+
+      // Unregister.
+      puppeteer.unregisterCustomQueryFunction('get-by-id');
+      try {
+        await page.$('get-by-id/foo');
+        expect.fail('Custom query function not set - throw expected');
+      } catch (e) {
```
mathiasbynens: nit: `error`

mathiasbynens (changes requested):
Looking great so far! Left some thoughts, PTAL.

src/JSHandle.ts
mokeyish: > If you're gonna use this to create a custom XPath-based query handler, you'll probably use `/` quite often in your selectors. With this implementation, `updatedSelector` only contains everything between the first `/` and the second `/`. We could instead use the result of `indexOf` with `slice` to get the entire selector.

What about this?
图片

test/elementhandle.spec.js
mokeyish: I think  using `id:elementId` will be better than `get-by-id/elementId`. 

图片

src/JSHandle.ts
mathiasbynens: With this patch, you could implement everything in that table yourself, without being limited to what's in the table!

test/elementhandle.spec.js
mathiasbynens: When registering your own query handler you could name it `id` and then it would be `id/elementId`, which seems close enough?

The reason we cannot use `:` is because it has special meaning in CSS, so selectors like `p:first` become ambiguous.

test/elementhandle.spec.js
jackfranklin: we could consider adding this eslint plugin/rule: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/master/docs/rules/catch-error-name.md to enforce this

src/QueryFunction.ts
```diff
@@ -0,0 +1,47 @@
...
+ * limitations under the License.
+ */
+
+import {ElementHandle} from './JSHandle';
+
+export interface QueryFunction {
+  (element: Element, selector: string): (ElementHandle | null) | ElementHandle[];
+}
+
+const _customQueryFunctions = new Map<string, QueryFunction>();
```
jackfranklin: nit: could we lose the `_` here? I think the fact it's not exported is enough.

src/QueryFunction.ts
jackfranklin: 👍 and then let's add some unit tests to check the behaviour

test/elementhandle.spec.js
mathiasbynens: let's do it!

src/QueryFunction.ts
paullewis: Sure

src/QueryFunction.ts
paullewis: It chucks a lint error: `Individual declarations in merged declaration 'customQueryFunctions' must be all exported or all local.`

src/JSHandle.ts
paullewis: Done.

src/JSHandle.ts
paullewis: Done.

src/Puppeteer.js
paullewis: I was aware that this file was likely going over to TS at some point soon, and therefore would get a "proper" signature at that point. AFAICT you can't easily specify JSDoc for function params, but maybe I missed something.

src/QueryFunction.ts
paullewis: It throws a lint error :(

src/QueryFunction.ts
paullewis: Added validation, but our current setup isn't geared up for unit tests particularly. I have a test that asserts it throws with invalid names.

src/Puppeteer.js
jackfranklin: 👍 on this - let's leave this and be much more explicit and define nicer types when we migrate this file to TS

test/elementhandle.spec.js
jackfranklin: Will do in a follow up PR :)

paullewis:
I _think_ that's all the feedback addressed. PTAL.

(@mathiasbynens I will namespace this under experimental as discussed when everything else is okay)

src/Puppeteer.js
mathiasbynens: I was talking about the method name (`registerCustomQueryFunction`), not the types

paullewis: Well that's changed now, too! 🥳

src/JSHandle.ts
mathiasbynens: Instead of just `_getQueryFunctionNameAndSelector`, could we extract the entire block? It's shared between `$$` and `$$eval` below.

```js
    let handler = (element: Element, selector: string): ElementHandle => element.querySelector(selector) as unknown as ElementHandle;
    if (selector.indexOf(SEARCH_SEPARATOR) !== -1) {
      const {name, updatedSelector} = this._getQueryFunctionNameAndSelector(selector);
      const queryFunction = customQueryHandlers().get(name);
      if (!queryFunction)
        throw new Error(`$ query set to use "${name}", but no query function of that name was found`);

      handler = queryFunction as typeof handler;
      selector = updatedSelector;
    }
```

Could be something like:

```js
const {handler, selector} = getHandlerAndSelector();
```

(with possibly better names)

WDYT?

src/QueryHandler.ts
```diff
@@ -0,0 +1,51 @@
...
+  (element: Element, selector: string): (ElementHandle | null) | ElementHandle[];
+}
+
+const _customQueryHandlers = new Map<string, QueryHandler>();
+
+export function registerCustomQueryHandler(name: string, queryFunction: Function): void {
+  if (_customQueryHandlers.get(name))
+    throw new Error(`A custom query function named "${name}" already exists`);
+
+  const isValidName = /^[a-zA-Z\-]+$/.test(name);
```
mathiasbynens: nit: no need to escape `-` as last char in character class

```suggestion
  const isValidName = _/^[a-zA-Z-]+$/_.test(name);
```

but as I said earlier I'd prefer simplifying this to `[a-zA-Z]`, so that query name handlers are (roughly) similar to JS API function names

src/JSHandle.ts
```diff
@@ -467,10 +497,18 @@ export class ElementHandle extends JSHandle {
...
-    const arrayHandle = await this.evaluateHandle(
-        (element, selector) => Array.from(element.querySelectorAll(selector)),
-        selector
-    );
+    let handler = (element: Element, selector: string): JSHandle => Array.from(element.querySelectorAll(selector)) as unknown as JSHandle;
+    if (selector.indexOf(SEARCH_SEPARATOR) !== -1) {
+      const {name, updatedSelector} = this._getQueryFunctionNameAndSelector(selector);
+      const queryFunction = customQueryHandlers().get(name);
+      if (!queryFunction)
+        throw new Error(`$$ query set to use "${name}", but no query function of that name was found`);
```
mathiasbynens: `$$` -> `$$eval`, but we don't need to be this specific here. If we abstract this into a shared helper between `$$eval` and `$$` as suggested above this will automatically be resolved

src/Puppeteer.js
```diff
@@ -34,6 +35,7 @@ module.exports = class {
     this._isPuppeteerCore = isPuppeteerCore;
     // track changes to Launcher configuration via options or environment variables
     this.__productName = productName;
+    this._customQueryFunctions = new Map();
```
mathiasbynens: One last thing, let's do a global rename for "queryFunction" to "queryHandler" so that the naming is internally consistent

paullewis: Actually, that Map isn't necessary. Removed.

src/QueryHandler.ts
paullewis: Done

src/QueryHandler.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import {ElementHandle, JSHandle} from './JSHandle';
+
+export interface QueryHandler {
+  (element: Element, selector: string): (ElementHandle | null) | ElementHandle[] | JSHandle;
+}
+
+const _customQueryHandlers = new Map<string, QueryHandler>();
+
+export function registerCustomQueryHandler(name: string, queryFunction: Function): void {
```
mathiasbynens:
```suggestion
export function registerCustomQueryHandler(name: string, handler: Function): void {
```

...and similar below

src/JSHandle.ts
```diff
@@ -442,10 +464,11 @@ export class ElementHandle extends JSHandle {
    * @return {!Promise<!Array<!ElementHandle>>}
    */
   async $$(selector: string): Promise<ElementHandle[]> {
-    const arrayHandle = await this.evaluateHandle(
-        (element, selector) => element.querySelectorAll(selector),
-        selector
-    );
+    const defaultHandler = (element: Element, selector: string): ElementHandle[] => element.querySelectorAll(selector) as unknown as ElementHandle[];
+    const {updatedSelector, queryHandler} = this._getQueryHandlerAndSelector(selector, defaultHandler,
+        '$$ query set to use "QUERY_HANDLER_NAME", but no query function of that name was found');
```
mathiasbynens: Looks like the error message is always the same. We could either parameterize the function name (`$`, `$$`, or `$$eval`) or just omit it. I think omitting seems fine:

    'Query set to use "QUERY_HANDLER_NAME", but no query function of that name was found'

That way we no longer need the `errorMessage` parameter.

test/elementhandle.spec.js
```diff
@@ -248,4 +248,53 @@ describe('ElementHandle specs', function() {
...
+      const classNames = await Promise.all(elements.map(async element => await page.evaluate(element => element.className, element)));
+      puppeteer.__experimental_unregisterCustomQueryHandler('getbyclass');
+
+      expect(classNames).toStrictEqual(['foo', 'foo baz']);
+    });
+    it('should eval correctly', async() => {
+      const {page, puppeteer} = getTestState();
+      await page.setContent('<div id="not-foo"></div><div class="foo">Foo1</div><div class="foo baz">Foo2</div>');
+      puppeteer.__experimental_registerCustomQueryHandler('getbyclass', (element, selector) => document.querySelectorAll(`.${selector}`));
+      const elements = await page.$$eval('getbyclass/foo', divs => divs.length);
```
mathiasbynens: nit: let's make these `getByClass` etc. to match built-in JS API naming

src/JSHandle.ts
paullewis: Done

src/JSHandle.ts
```diff
@@ -425,11 +428,29 @@ export class ElementHandle extends JSHandle {
...
+      { updatedSelector: string; queryHandler: QueryHandler} {
+    if (selector.indexOf(SEARCH_SEPARATOR) === -1)
+      return {updatedSelector: selector, queryHandler: defaultQueryHandler};
+
+    const index = selector.indexOf(SEARCH_SEPARATOR);
+    const name = selector.slice(0, index);
+    const updatedSelector = selector.slice(index + SEARCH_SEPARATOR.length);
+    const queryHandler = customQueryHandlers().get(name);
+    if (!queryHandler)
+      throw new Error('Query set to use "QUERY_HANDLER_NAME", but no query handler of that name was found'.replace('QUERY_HANDLER_NAME', name));
```
mathiasbynens:
```suggestion
      throw new Error(`Query set to use "${name}", but no query handler of that name was found`);
```

mathiasbynens (approved):

paullewis:
Okay, these are the call sites. I'll go through and confirm that each one lands in the correct place to have custom query support.

 - [X] page.$(selector)
 - [X] page.$$(selector)
 - [X] page.$$eval(selector, pageFunction[, ...args])
 - [X] page.$eval(selector, pageFunction[, ...args])
 - [X] page.click(selector[, options])
 - [X] page.focus(selector)
 - [X] page.hover(selector)
 - [X] page.select(selector, ...values)
 - [X] page.tap(selector)
 - [X] page.type(selector, text[, options])
 - [X] page.waitFor(selectorOrFunctionOrTimeout[, options[, ...args]])
 - [X] page.waitForSelector(selector[, options])
 - [X] frame.$(selector)
 - [X] frame.$$(selector)
 - [X] frame.$$eval(selector, pageFunction[, ...args])
 - [X] frame.$eval(selector, pageFunction[, ...args])
 - [X] frame.click(selector[, options])
 - [X] frame.focus(selector)
 - [X] frame.hover(selector)
 - [X] frame.select(selector, ...values)
 - [X] frame.tap(selector)
 - [X] frame.type(selector, text[, options])
 - [X] frame.waitFor(selectorOrFunctionOrTimeout[, options[, ...args]])
 - [X] frame.waitForSelector(selector[, options])
 - [X] elementHandle.$(selector)
 - [X] elementHandle.$$(selector)
 - [X] elementHandle.$$eval(selector, pageFunction[, ...args])
 - [X] elementHandle.$eval(selector, pageFunction[, ...args])

paullewis:
Had to force push because of all sorts of conflicts that I presumably made worse during rebase :(

Anyway, PTAL

src/DOMWorld.ts
```diff
@@ -430,18 +436,24 @@ class WaitTask {
...
+    return (function wrapper(args) {
+      ${ predicateQueryHandlerBody ? `const predicateQueryHandler = ${predicateQueryHandlerBody};` : ''}
+      return (${predicateBody})(...args);
+    })(args)`;
+
     this._domWorld = domWorld;
     this._polling = polling;
     this._timeout = timeout;
-    this._predicateBody = helper.isString(predicateBody) ? 'return (' + predicateBody + ')' : 'return (' + predicateBody + ')(...args)';
+    this._predicateBody = helper.isString(predicateBody) ? 'return (' + predicateBody + ')' : wrapper;
```
mathiasbynens: This is getting complicated enough to move to its own helper like

```suggestion
    this._predicateBody = getPredicateBody(predicateBody, predicateQueryHandlerBody);
```

which then does

```js
if (helper.isString(predicateBody)
  return `return (${predicateBody});`;
if (predicateQueryHandlerBody) {
  return `
    return (function wrapper(args) {
      const predicateQueryHandler = ${predicateQueryHandlerBody};
      return (${predicateBody})(...args);
    })(args);`;
}
return `return (${predicateBody})(...args);`;
```

src/QueryHandler.ts
```diff
@@ -0,0 +1,74 @@
...
+  const hasCustomQueryHandler = /^[a-zA-Z]+\//.test(selector);
+  if (!hasCustomQueryHandler)
+    return {updatedSelector: selector, queryHandler: defaultQueryHandler};
+
+  const index = selector.indexOf('/');
+  const name = selector.slice(0, index);
+  const updatedSelector = selector.slice(index + 1);
+  const queryHandler = customQueryHandlers().get(name);
+  if (!queryHandler)
+    throw new Error('Query set to use "QUERY_HANDLER_NAME", but no query handler of that name was found'.replace('QUERY_HANDLER_NAME', name));
```
mathiasbynens:
```suggestion
    throw new Error(`Query set to use "${name}", but no query handler of that name was found`);
```

mathiasbynens (approved):
LGTM! Super excited about this :fire::fire::fire:

Left some suggestions for cleanup/polish, but I'm fine with doing those as follow-up changes in a separate PR if you prefer.

paullewis:
Changed CLA to yes because all commits were by people who have signed the CLA.

paullewis:
Changed CLA to yes (again) because all commits were by people who have signed the CLA.

PR #5760, opened 2020-04-28
chore: update protocol.d.ts

jackfranklin:
The result of `npm run update-protocol-d-ts`

### DIFF
--- a/src/protocol.d.ts
+++ b/src/protocol.d.ts
@@ -797,7 +797,7 @@ events afterwards if enabled and recording.
              */
             windowState?: WindowState;
         }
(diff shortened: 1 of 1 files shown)


mathiasbynens (approved):

mathiasbynens:
Should we enforce this on prepublish (or pre-commit, with git hooks)?

jackfranklin:
> Should we enforce this on prepublish (or pre-commit, with git hooks)?

Yeah, I was just thinking similar. Maybe a CI task that runs the generator and compares ? And errors if they are different? 

My thinking of it being on CI is then people don't pay the cost on every commit. WDYT?

jackfranklin:
@mathiasbynens I'm going to merge this as it's blocking my work on a TS migration but let me know what your thoughts are on how best to enforce this.

mathiasbynens:
> My thinking of it being on CI is then people don't pay the cost on every commit. WDYT?

SGTM.

PR #5761, opened 2020-04-28
chore: migrate src/Browser to TS

jackfranklin:
This was an interestng migration and hit upon some tweaks elsewhere in
the codebase where we can gain stricter typing now more code is ported
to TS.


mathiasbynens (approved):
Rubberstamp LGTM

src/Browser.ts
```diff
@@ -0,0 +1,312 @@
...
+    return !!this._id;
+  }
+
+  async overridePermissions(origin: string, permissions: Protocol.Browser.PermissionType[]): Promise<void> {
+    const webPermissionToProtocol = new Map<string, Protocol.Browser.PermissionType>([
+      ['geolocation', 'geolocation'],
+      ['midi', 'midi'],
+      ['notifications', 'notifications'],
+      // TODO: push isn't a valid type?
+      // ['push', 'push'],
```
jackfranklin, line 272: @mathiasbynens this is interesting to me. The protocol doesn't list `push` as a valid permission type so I'm not quite sure if:

* the protocol is wrong
* Puppeteer is wrong and calling overridePermissions with `push` was effectively a no-op.

I think I'm lacking context here so I can't really make a call. We can keep the code and loosen the types to `string[]` rather than `PermissionType[]` but if this is highlighting a bug it's worth exploring.

src/Connection.ts
```diff
@@ -72,7 +72,7 @@ export class Connection extends EventEmitter {
     return this._url;
   }
 
-  send<ReturnType extends {}>(method: string, params = {}): Promise<ReturnType> {
+  send<T extends keyof Protocol.CommandParameters>(method: T, params?: Protocol.CommandParameters[T]): Promise<Protocol.CommandReturnValues[T]> {
```
jackfranklin, line 75: this is an improvement on the types that I missed previously that was highlighted when migrating src/Browser

src/TaskQueue.ts
```diff
@@ -19,7 +19,7 @@
  * to make this into TaskQueue<T> and let the caller tell us what types
  * the promise in the queue should return.
  */
-class TaskQueue {
+export class TaskQueue {
```
jackfranklin, line 22: this change allows the file to be imported into other TS files nicely

src/Browser.ts
mathiasbynens: I lack the context here as well. Let's stick with the existing behavior for now, and we can resolve this separately later. Thanks for adding the TODO.

src/Connection.ts
mathiasbynens: neat!

mathiasbynens (approved):

PR #5762, opened 2020-04-28
chore: enforce src/protocol.d.ts is in sync

jackfranklin:
On CI we run `npm run compare-protocol-d-ts` which checks that the file
on disk is up to date with the protocol we fetch from the browser.

### DIFF
--- a/utils/protocol-types-generator/index.js
+++ b/utils/protocol-types-generator/index.js
@@ -98,7 +98,7 @@ const cli = async() => {
       console.log(`Success: ${relativeProtocolOutputPath} is up to date.`);
     } else {
       console.log(`Error: ${relativeProtocolOutputPath} is out of date.`);
-      console.log('You should run \`npm run update-protocol-d-ts\` and commit the changes.');
+      console.log('You should run `npm run update-protocol-d-ts` and commit the changes.');
       process.exit(1);
     }
 
--- a/utils/protocol-types-generator/index.js
+++ b/utils/protocol-types-generator/index.js
@@ -1,7 +1,8 @@
 // @ts-check
 const path = require('path');
 const puppeteer = require('../..');
-module.exports = puppeteer.launch({
+
+const fetchAndGenerateProtocolDefinitions = () => puppeteer.launch({
   pipe: false,
   executablePath: process.env.BINARY,
 }).then(async browser => {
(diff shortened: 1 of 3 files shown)


utils/protocol-types-generator/index.js
```diff
@@ -72,12 +73,47 @@ declare global {
...
 `;
-  const outputPath = path.join(__dirname, '..', '..', 'src', 'protocol.d.ts');
-  require('fs').writeFileSync(outputPath, output);
-  console.log(`Wrote protocol.d.ts for ${version} to ${path.relative(process.cwd(), outputPath)}`);
-  console.log(`You should commit the changes.`);
+
+  return {output, version};
 });
 
+const protocolOutputPath = path.join(__dirname, '..', '..', 'src', 'protocol.d.ts');
```
mathiasbynens, line 80: Readability:

```suggestion
const protocolOutputPath = path.join(__dirname, '../../src/protocol.d.ts');
```

utils/protocol-types-generator/index.js
```diff
@@ -72,12 +73,47 @@ declare global {
...
+  if (scriptToRun === 'update') {
+    writeOutputToDisk(await fetchAndGenerateProtocolDefinitions());
+  } else if (scriptToRun === 'compare') {
+    const {output} = await fetchAndGenerateProtocolDefinitions();
+    const outputOnDisk = require('fs').readFileSync(protocolOutputPath, {encoding: 'utf8'});
+    if (output === outputOnDisk) {
+      console.log(`Success: ${relativeProtocolOutputPath} is up to date.`);
+    } else {
+      console.log(`Error: ${relativeProtocolOutputPath} is out of date.`);
+      console.log('You should run \`npm run update-protocol-d-ts\` and commit the changes.');
```
mathiasbynens: unnecessary escaping:

```suggestion
      console.log('You should run `npm run update-protocol-d-ts` and commit the changes.');
```

mathiasbynens (approved):
LGTM % nits

utils/protocol-types-generator/index.js
jackfranklin: I'd rather leave this as is so it works should anyone run this on Windows.

mathiasbynens: [`path.join` normalizes](https://nodejs.org/api/path.html#path_path_join_paths), so it would still work on Windows. (I would not have suggested this change otherwise.)

PR #5764, opened 2020-04-28
chore: migrate src/DOMWorld to TypeScript

jackfranklin:


src/DOMWorld.ts
```diff
@@ -615,20 +506,13 @@ class WaitTask {
     this._cleanup();
   }
 
-  _cleanup() {
+  _cleanup(): void {
     clearTimeout(this._timeoutTimer);
     this._domWorld._waitTasks.delete(this);
-    this._runningTask = null;
```
jackfranklin, line 621: this was never used or read anywhere across the entire codebase so I removed this line

mathiasbynens (approved):

src/DOMWorld.ts
mathiasbynens: good catch!

PR #5765, opened 2020-04-28
chore: Enforce array type styles in TypeScript

jackfranklin:
This change enforces how we type arrays, e.g. choosing between:

* `string[]`
* `Array<string>`

I've gone for the `array-simple` option [1] which enforces that:

* primitive types and type references use `X[]`
* complex types use `Array<X>`

For example, we'd type an array of strings as `string[]`, but an array
of a union type as `Array<SomeUnionType>`.

[1]: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/array-type.md

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -120,7 +120,10 @@ module.exports = {
                 "@typescript-eslint/no-empty-function": 0,
                 "@typescript-eslint/no-use-before-define": 0,
                 // We know it's bad and use it very sparingly but it's needed :(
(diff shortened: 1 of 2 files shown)


mathiasbynens (approved):
I quite like the `Array<>` notation (even for primitive types and type references), but don't feel strongly.

jackfranklin:
I don't feel strongly either - the one I picked follows the Google TS style guide, but happy to change it if you feel strongly. It's easily changed in the future too (it's fixable by eslint)

PR #5779, opened 2020-04-30
chore: update how we track coverage during unit tests

jackfranklin:
The old method of tracking coverage was causing issues. If a test failed
on CI, that test's failure would be lost because the test failing would
in turn cause the coverage to fail, but the `process.exit(1)` in the
coverage code caused Mocha to not output anything useful.

Instead the coverage checker now:

* tracks the coverage in memory in a Map (this hasn't changed)
* after all tests, writes that to disk in test/coverage.json (which is
gitignored)
* we then run a single Mocha test that asserts every method was called.

This means if the test run fails, the build will fail and give the error
about that test run, and that output won't be lost when the coverage
then fails too.

### DIFF
--- a/test/assert-coverage-test.js
+++ b/test/assert-coverage-test.js
@@ -1,0 +1,24 @@
+const {describe, it} = require('mocha');
+const {getCoverageResults} = require('./coverage-utils');
+const expect = require('expect');
+
+describe('API coverage test', () => {
+  it('calls every method', () => {
+    if (!process.env.COVERAGE) return;
+
+    const coverageMap = getCoverageResults();
+    const missingMethods = [];
+    for (const method of coverageMap.keys()) {
+      if (!coverageMap.get(method))
+        missingMethods.push(method);
+    }
+    if (missingMethods.length) {
+      console.error('\nCoverage check failed: not all API methods called. See above ouptut for list of missing methods.');
+      console.error(Array.from(missingMethods).join('\n'));
+    }
+
+    // We know this will fail because we checked above
+    // but we need the actual test to fail.
+    expect(missingMethods.length).toEqual(0);
+  });
+});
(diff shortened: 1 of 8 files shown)


test/assert-coverage-test.js
```diff
@@ -0,0 +1,24 @@
...
+    if (!process.env.COVERAGE) return;
+
+    const coverageMap = getCoverageResults();
+    const missingMethods = [];
+    for (const method of coverageMap.keys()) {
+      if (!coverageMap.get(method))
+        missingMethods.push(method);
+    }
+    if (missingMethods.length) {
+      console.error('\nCoverage check failed: not all API methods called. See above ouptut for list of missing methods.');
```
mathiasbynens: typo:

```suggestion
      console.error('\nCoverage check failed: not all API methods called. See above output for list of missing methods.');
```

test/assert-coverage-test.js
```diff
@@ -0,0 +1,24 @@
...
+
+    const coverageMap = getCoverageResults();
+    const missingMethods = [];
+    for (const method of coverageMap.keys()) {
+      if (!coverageMap.get(method))
+        missingMethods.push(method);
+    }
+    if (missingMethods.length) {
+      console.error('\nCoverage check failed: not all API methods called. See above output for list of missing methods.');
+      console.error(Array.from(missingMethods).join('\n'));
```
mathiasbynens: I don't understand why we need `Array.from` here. How about this:

```suggestion
      console.error(missingMethods.join('\n'));
```

lmk if I'm missing something

mathiasbynens (approved):
LGTM % comments

test/assert-coverage-test.js
jackfranklin: No nor me - this was left over from pre-me working on it. Nice spot!

jackfranklin:
Added the CLA label as the two commiters are myself + Mathias :). I think the rebase I did against master threw the googlebot off.

PR #5789, opened 2020-04-30
chore: migrate src/Puppeteer to TypeScript

jackfranklin:


index.js
```diff
@@ -25,7 +25,7 @@ for (const className in api) {
 // Expose alias for deprecated method.
 Page.prototype.emulateMedia = Page.prototype.emulateMediaType;
 
-const Puppeteer = require('./lib/Puppeteer');
```
jackfranklin, line 28: I changed lib/Puppeteer to export as a named export which is the pattern the vast majority of files use in the codebase. It also plays nicer with TypeScript and outputting as CommonJS.

This does mean that if someone has `const Puppeteer = require('puppeteer/lib/puppeteer')` it would break, but I don't think we should consider that a breaking change (`src/` should be considered internal + we could change its structure/files at any point).

@mathiasbynens WDYT?

jackfranklin: found 3 matches on GitHub for this pattern: https://github.com/search?q=%22require%28%27puppeteer%2Flib%2FPuppeteer%27%29%22&type=Code but they don't look like active codebases.

Plus the change only requires adding braces around the require. We can call that out in the release notes.

mathiasbynens (approved):

index.js
mathiasbynens: Agreed we should be able to change this.

PR #5805, opened 2020-05-04
chore: improve readability for path strings

mathiasbynens:
This addresses the feedback from https://github.com/puppeteer/puppeteer/pull/5762#discussion_r416554314.

### DIFF
--- a/utils/protocol-types-generator/index.js
+++ b/utils/protocol-types-generator/index.js
@@ -77,7 +77,7 @@ export default Protocol;
   return {output, version};
 });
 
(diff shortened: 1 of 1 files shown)


jackfranklin:
I don't actually agree with this change - I prefer using `join` consistently rather than mixing both - but I don't feel strongly enough to block it. Just sharing my 2c :)

jackfranklin (approved):

mathiasbynens:
Ack. `join` normalizes the resulting path to allow for exactly this kind of usage. IMHO we should make use of it, since the hardcoded path part is now readable and copy-pastable, which seems strictly better than that not being the case.

PR #5809, opened 2020-05-05
chore: migrate src/Page.js to TypeScript

jackfranklin:
The final one! This is a huge file and needs to be split up and tidied,
but for now I've left all the definitions in place and converted types
accordingly.

There's some additional tidying we can do now every `src` file is TS,
but I'll leave that for another PR to avoid this one getting any bigger.

### DIFF
--- a/src/Dialog.ts
+++ b/src/Dialog.ts
@@ -17,14 +17,17 @@
 import {assert} from './helper';
 import {CDPSession} from './Connection';
 
-enum DialogType {
+/* TODO(jacktfranklin): protocol.d.ts defines this
+ * so let's ditch this and avoid the duplication
+ */
+export enum DialogType {
   Alert = 'alert',
   BeforeUnload = 'beforeunload',
   Confirm = 'confirm',
   Prompt = 'prompt'
 }
 
-class Dialog {
+export class Dialog {
   static Type = DialogType;
 
   private _client: CDPSession;
@@ -69,5 +72,3 @@ class Dialog {
     });
   }
 }
-
-export = {Dialog};
--- a/src/Page.ts
+++ b/src/Page.ts
@@ -146,7 +146,7 @@ export class Page extends EventEmitter {
   _pageBindings = new Map<string, Function>();
   _coverage: Coverage;
   _javascriptEnabled = true;
-  _viewport?: Puppeteer.Viewport;
+  _viewport: Puppeteer.Viewport | null;
   _screenshotTaskQueue: TaskQueue;
   _workers = new Map<string, PuppeteerWorker>();
   // TODO: improve this typedef - it's a function that takes a file chooser or something?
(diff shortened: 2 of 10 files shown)


src/Dialog.ts
```diff
@@ -17,14 +17,17 @@
 import {assert} from './helper';
 import {CDPSession} from './Connection';
 
-enum DialogType {
+/* TODO(jacktfranklin): protocol.d.ts defines this
```
jackfranklin, line 20: follow up PR

src/externs.d.ts
```diff
@@ -1,14 +1,7 @@
...
   module Puppeteer {
-    export class Page extends RealPage { }
-    /* TODO(jacktfranklin@): once DOMWorld, Page, and FrameManager are in TS
-     * we can remove this and instead use the type defined in LifeCycleWatcher
-     */
-    export type PuppeteerLifeCycleEvent = 'load' | 'domcontentloaded' | 'networkidle0' | 'networkidle2';
-
     export interface ConnectionTransport {
       send(string);
       close();
```
jackfranklin, line 7: in a follow up PR I will remove this file and find a new home for the interfaces we need where they are not exposed globally

utils/doclint/check_public_api/index.js
```diff
@@ -359,6 +359,46 @@ function compareDocumentations(actual, expected) {
...
+        expectedName: 'ScreenshotOptions'
+      }],
+      ['Method Page.setContent() options', {
+        actualName: 'Object',
+        expectedName: 'WaitForOptions'
+      }],
+      ['Method Page.setCookie() ...cookies', {
+        actualName: '...Object',
+        expectedName: '...CookieParam'
+      }],
```
jackfranklin, line 401: now we're in TS land we can start diving properly into documentation tools!

src/Page.ts
```diff
@@ -14,92 +14,160 @@
...
+  _frameManager: FrameManager;
+  _emulationManager: EmulationManager;
+  _tracing: Tracing;
+  _pageBindings = new Map<string, Function>();
+  _coverage: Coverage;
+  _javascriptEnabled = true;
+  _viewport?: Puppeteer.Viewport;
+  _screenshotTaskQueue: TaskQueue;
+  _workers = new Map<string, PuppeteerWorker>();
+  // TODO: improve this typedef - it's a function that takes a file chooser or something?
```
jackfranklin, line 152: will do in a follow up PR

src/Dialog.ts
mathiasbynens: ack

utils/doclint/check_public_api/JSBuilder.js
```diff
@@ -204,6 +204,7 @@ function checkSources(sources) {
 
     for (const [name, member] of symbol.members || []) {
 
+
```
mathiasbynens: This seems accidental?

```suggestion
```

utils/doclint/check_public_api/index.js
mathiasbynens: \o/ so good!

mathiasbynens (approved):
lgtm

PR #5816, opened 2020-05-06
fix: remove node@14 specific extract timeout

### LINKED ISSUE #5719: Puppeteer not working with node 14

AviVahl:
closes #5719

Original stream.pipeline issue has been fixed in Node 14.1.0 and above.

Meanwhile, the workaround itself caused timout failures in many CI Node 14 runs, as reported in #5719.

Removed it, as it's no longer needed, and actually blocks consumers from adding CI testing on Node 14.

### DIFF
--- a/src/BrowserFetcher.ts
+++ b/src/BrowserFetcher.ts
@@ -22,7 +22,7 @@ import * as childProcess from 'child_process';
 import * as https from 'https';
 import * as http from 'http';
 
-import * as extract from 'extract-zip';
+import * as extractZip from 'extract-zip';
 import * as debug from 'debug';
 import * as removeRecursive from 'rimraf';
 import * as URL from 'url';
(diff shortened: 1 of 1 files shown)


jackfranklin (approved):
Thank you!

PR #5821, opened 2020-05-07
chore: add Node.js 12 & 14 to CI

jackfranklin:


jackfranklin:
@mathiasbynens I think we should think about if we want to stick with our current setup of running primarily on Node 10 or if we should bring the baseline up. I have no strong prefs as long as we cover all versions we care about :)

mathiasbynens (approved):

mathiasbynens:
@jackfranklin I feel the same way. It's important that we keep running the general tests on LTS, but it seems like a good idea to move up the base line to, say, the latest stable version of Node.js for the extra checks.

PR #5825, opened 2020-05-07
Add Prettier to the codebase.

jackfranklin:
@mathiasbynens as discussed. Recommend going by commit as the last commit is just the re-format, the first two do the setup. Happy to change any of the config.

I also took the opportunity to tidy the ESLint file and remove rules that Prettier will effectively enforce.


mathiasbynens (approved):
Improved and automated formatting (finally we can have spaces between `async` and `()` again!) and a simpler eslint as well. Big fan of this patch! :) Thanks Jack

PR #5832, opened 2020-05-07
chore: private-ise src/Accessibility.ts

jackfranklin:
Now the first pass of migrating to TypeScript is complete I'm going
through the src files one by one to tidy up the public/private
interfaces.

Puppeteer used an underscore convention to denote privacy but violates
this in some places; we should be strict with TypeScript's `public` and
`private` keywords instead.

This means we'll get nice TS errors if you try to refer to a private
method/variable, and means when we swap to generating our TS docs the
tooling knows what method(s) are public and therefore need to be
documented vs private internals that don't.

### DIFF
--- a/src/Accessibility.ts
+++ b/src/Accessibility.ts
@@ -51,13 +51,13 @@ interface SerializedAXNode {
 }
 
 export class Accessibility {
-  _client: CDPSession;
+  private _client: CDPSession;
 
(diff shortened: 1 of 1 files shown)


mathiasbynens (approved):

mathiasbynens:
Rubberstamp LGTM

PR #5850, opened 2020-05-12
chore: extract `BrowserRunner` into its own module

jackfranklin:
`src/Launcher.ts` is large and hard to work in. It has multiple objects
defined in it:

* ChromeLauncher
* FirefoxLauncher
* BrowserRunner
* Launcher

This change moves BrowserRunner into its own module. More refactorings
like this will follow but this is the first step.

### DIFF
--- a/src/launcher/BrowserRunner.ts
+++ b/src/launcher/BrowserRunner.ts
@@ -1,0 +1,241 @@
+/**
+ * Copyright 2020 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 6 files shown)


src/launcher/BrowserRunner.ts
```diff
@@ -0,0 +1,241 @@
...
+import { Connection } from '../Connection';
+import { WebSocketTransport } from '../WebSocketTransport';
+import { PipeTransport } from '../PipeTransport';
+import * as readline from 'readline';
+import { TimeoutError } from '../Errors';
+
+const removeFolderAsync = helper.promisify(removeFolder);
+const debugLauncher = debug('puppeteer:launcher');
+
+export class BrowserRunner {
```
jackfranklin, line 32: the functionality and contents of this class has not changed, other than the `private` keyword on the instance properties.

mathiasbynens (approved):
Nice! This smells like the beginning of #5851 :)

jackfranklin:
@mathiasbynens you read my mind!

PR #5856, opened 2020-05-13
chore: extract `ConsoleMessage` and `FileChooser` into its own module

christian-bromann:
To support @jackfranklin's recent efforts in splitting up classes into their own modules here is a patch to do this for `ConsoleMessage` and `FileChooser`. It also starts to extract types into a single module and define them as types instead of interfaces.


src/types.ts
```diff
@@ -0,0 +1,97 @@
...
+export type PuppeteerLifeCycleEvent =
+  | 'load'
+  | 'domcontentloaded'
+  | 'networkidle0'
+  | 'networkidle2';
+export type ProtocolLifeCycleEvent =
+  | 'load'
+  | 'DOMContentLoaded'
+  | 'networkIdle'
+  | 'networkAlmostIdle';
```
jackfranklin: @christian-bromann do you have a reason for types over interfaces? I tend to lean towards interfaces but don't have a hugely strong opinion. I think we should be consistent though.

I'd also rather not create a generic `types.ts` file but instead keep these types in the modules that make sense to own them. E.g. `PuppeteerLifeCycleEvent` sits nicely inside the `LifecycleWatcher`.

There are definitely some that should be in a generic `types.ts` file but I think for the ones that have a natural home inside another module we could leave them there for now.

jackfranklin (commented):
Nice, thanks for helping out! :)

I'd like to think about interfaces vs types and pick one and probably put some of the types back into the modules that own them (see the other comment) but once that's done I think this LGTM!

src/types.ts
christian-bromann: > do you have a reason for types over interfaces? I tend to lean towards interfaces but don't have a hugely strong opinion. I think we should be consistent though.

Not really. I thought if we would want to collect some generic types in an own module it would make sense to actually define them as types. Let's keep things as they are for now and let me revert this specific change.

christian-bromann:
Can we rerun `3173.3 Unit tests [with coverage]: Linux/Chromium`, seems like a flake.

jackfranklin:
@christian-bromann it's not a flake, it's because the references to these classes in `src/api.ts` are now invalid. Would you mind updating those please?

jackfranklin (approved):

PR #5858, opened 2020-05-13
chore: extract `SecurityDetails` into its own module

christian-bromann:
Continuing with the effort of #5856 splitting classes into their own modules.

### DIFF
--- a/src/SecurityDetails.ts
+++ b/src/SecurityDetails.ts
@@ -15,11 +15,11 @@
  */
 
 export class SecurityDetails {
-  _subjectName: string;
-  _issuer: string;
-  _validFrom: number;
-  _validTo: number;
-  _protocol: string;
+  private _subjectName: string;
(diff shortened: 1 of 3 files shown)


src/SecurityDetails.ts
```diff
@@ -0,0 +1,51 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export class SecurityDetails {
+  _subjectName: string;
+  _issuer: string;
+  _validFrom: number;
+  _validTo: number;
+  _protocol: string;
```
jackfranklin: while we're in this area, could we try putting `private` on these and see if TypeScript is happy? Just a nice quick win whilst we're here :)

jackfranklin (approved):
Thank you so much, this is great! One thought RE privatising properties.

PR #5473, opened 2020-03-04
Don't use expect within Promises (#5466)

whimboo:
This PR will partly fix issue #5466.

If a call to expect fails within a Promise it will not
be resolved, and causing the test to crash.

The patch aligns the code similar to what is used by all
the other tests.

Issue #5466, opened 2020-02-28
Some tests with Firefox cause a crash of the test runner and an abort of the test run


whimboo:
### Steps to reproduce
Run all the Puppeteer unit tests with Firefox, and make sure that `should allow accepting prompts` in dialog.spec.js is enabled. When all the unit tests are run, it fails with this crash message:

```
## CRASHED ##
Message:
  UNHANDLED PROMISE REJECTION
Stack:
  Error: expect.toBe failed:  == yes.
      at applyMatcher (/Users/henrik/code/puppeteer/utils/testrunner/Matchers.js:48:15)
      at Page.<anonymous> (/Users/henrik/code/puppeteer/test/dialog.spec.js:35:39)
      at Page.emit (events.js:196:13)
      at Page._onDialog (/Users/henrik/code/puppeteer/lib/Page.js:641:10)
      at CDPSession.<anonymous> (/Users/henrik/code/puppeteer/lib/Page.js:120:61)
      at CDPSession.emit (events.js:196:13)
      at CDPSession._onMessage (/Users/henrik/code/puppeteer/lib/Connection.js:200:12)
      at Connection._onMessage (/Users/henrik/code/puppeteer/lib/Connection.js:112:17)
      at WebSocket.<anonymous> (/Users/henrik/code/puppeteer/lib/WebSocketTransport.js:44:24)
      at WebSocket.onMessage (/Users/henrik/code/puppeteer/node_modules/ws/lib/event-target.js:120:16)
WORKERS STATE
  0: [TERMINATED] Firefox Browser Page Page.Events.Dialog should allow accepting prompts (dialog.spec.js:32:5)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
```

When I run this test alone I only get a test failure with the navigation timeout. But no crash.

**Tell us about your environment:**

* Puppeteer version: 2.0.0-post
* Platform / OS version: MacOS 10.15
* URLs (if applicable):
* Node.js version: 10.19.0

**What steps will reproduce the problem?**
1. Run npm install
2. Edit dialog.spec.js and enable `should allow accepting prompts` for Firefox
3. Run PUPPETEER_EXECUTABLE_PATH=firefox npm run funit

**What is the expected result?**
After the failure logged the unit test run should continue.

**What happens instead?**
Due to the crash failure the whole test run is immediately aborted.

whimboo:
[...]

PR #5859, opened 2020-05-13
chore: Fix async dialog specs when they fail

jackfranklin:
When this test was failing, it would cause no future tests to run. This
was because the `expect` call within the `page.on` callback would throw
an error, and that would trigger a unhandled promise rejection that
caused the test framework to stop.

The fundamental issue here is making `expect` calls within callbacks.
They are brittle due to the fact that they throw, and the test framework
won't catch it, but also because you have no guarantee that they will
run. If the callback is never executed you dont' know about it.

Although it's slightly more code, using a stub is the way to do this.
Not only can we assert that the stub was called, we can make synchronous
`expect` calls that Mocha will pick up properly if they fail.

Before this change, running the tests (and making it fail on purpose)
would cause all test execution to stop:

```
> puppeteer@3.0.4-post unit /Users/jacktfranklin/src/puppeteer
> mocha --config mocha-config/puppeteer-unit-tests.js

  .(node:69580) UnhandledPromiseRejectionWarning: Error: expect(received).toBe(expected) // Object.is equality

Expected: "yes."
Received: ""
    at Page.<anonymous> (/Users/jacktfranklin/src/puppeteer/test/dialog.spec.js:42:37)
    [snip]
(node:69580) UnhandledPromiseRejectionWarning: Unhandled promise rejection ... [snip]
```

But with this change, the rest of the tests run:

```
> puppeteer@3.0.4-post unit /Users/jacktfranklin/src/puppeteer
> mocha --config mocha-config/puppeteer-unit-tests.js

  Page.Events.Dialog
    ✓ should fire
    1) should allow accepting prompts
    ✓ should dismiss the prompt

  2 passing (2s)
  1 failing

  1) Page.Events.Dialog
       should allow accepting prompts:
     Error: expect(received).toBe(expected) // Object.is equality

Expected: "yes."
Received: ""
      at Context.<anonymous> (test/dialog.spec.js:53:35)
      at processTicksAndRejections (internal/process/task_queues.js:94:5)
```

This is much better because one failing test now doesn't stop the rest
of the test suite.

This probably isn't the only instance of this in the codebase so I
propose as we encounter them we fix them usng this commit as the
template.

### DIFF
--- a/test/dialog.spec.js
+++ b/test/dialog.spec.js
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 const expect = require('expect');
+const sinon = require('sinon');
+
 const {
   getTestState,
   setupTestPageAndContextHooks,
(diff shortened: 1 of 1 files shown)


jackfranklin:
@whimboo I believe this is the root cause of #5466. Let me know what you think.

whimboo:
> This probably isn't the only instance of this in the codebase so I
> propose as we encounter them we fix them usng this commit as the
> template.

See also the second half of my patch on PR #5473, which fixes another test in `navigation.spec.js`.

jackfranklin:
@whimboo are you happy with this being merged?

whimboo:
I think that's more a decision between you and @mathiasbynens.

If we go that way I assume all other instances will also have to be updated in the same way?

jackfranklin:
> If we go that way I assume all other instances will also have to be updated in the same way?

Yes, we can do that incrementally as we find them.

@mathiasbynens let me know if you're happy with this. Pretty confident it solves the issues.

mathiasbynens (approved):

mathiasbynens:
Alright, let's do it.

PR #5860, opened 2020-05-13
chore: upgrade TypeScript to 3.9

jackfranklin:
Sticking to a specific version of TS rather than a ^ version so we know
everyone is on the exact same version. Think that's preferable for such
an important dependency.


mathiasbynens (approved):
Sounds reasonable

PR #5861, opened 2020-05-13
chore: extract `Request` and `Response`  into its own module

christian-bromann:
Continuing with the effort of #5856 splitting classes into their own modules.


jackfranklin (approved):
Nice, thank you!

src/NetworkManager.ts
```diff
@@ -274,8 +275,7 @@ export class NetworkManager extends EventEmitter {
     const response = new Response(this._client, request, responsePayload);
     request._response = response;
     request._redirectChain.push(request);
-    response._bodyLoadedPromiseFulfill.call(
-      null,
+    response._resolveBody(
```
jackfranklin, line 278: @christian-bromann is this change deliberate? Looks like we lost a call to `._bodyLoadedPromiseFulfill`.

christian-bromann: @jackfranklin when making `_bodyLoadedPromiseFulfill` a private field we can't access it from `NetworkManager` anymore. I created a method in the Response class to allow access to it.

```js
  _resolveBody(err: Error | null): void {
    return this._bodyLoadedPromiseFulfill(err);
  }
```

Alternatively I could just keep `_bodyLoadedPromiseFulfill` as non private field.

jackfranklin: Ah got you, thanks for explaining. As a general rule I think we should just do the easy privatisation work and leave any others as non private when initially splitting classes out (just to keep changes per PR down) but in this case this all looks fine to me and CI is green so no need to undo any of it 👍

PR #5862, opened 2020-05-13
chore: force Mocha to exit on CI

jackfranklin:
We have some bug where some test runs will pass but then stall as Mocha
doesn't exit cleanly. This is proving very hard to track down (I've yet
to replicate it or find the test that causes it) and it doesn't happen
consistently.

The `exit` flag forces Mocha to hard exit. This will help with CI
stability. The flag only gets set on CI runs.

### DIFF
--- a/mocha-config/base.js
+++ b/mocha-config/base.js
@@ -1,3 +1,4 @@
 module.exports = {
   reporter: 'dot',
+  exit: !!process.env.CI,
 };


mathiasbynens (approved):

AviVahl:
Please note, this flag is _really_ dangerous. It obscures real issues and allows _other_ "failing to cleanup" issues to slip in while turned on.

`jsdom` had it on for a while (removed it now), and it allowed a change where `requestAnimationFrame` caused a stale process. They didn't know anything about it, as the flag caused mocha to always exit.

@jackfranklin @mathiasbynens 

Writing this, as I myself [have been bitten](https://github.com/puppeteer/puppeteer/issues/5356) by never-closing processes on `puppeteer@2.1.0-2.1.1`.

mathiasbynens:
Thanks for sharing those insights @AviVahl, that's good to know.

Ideally we'd find the cause for the flaky stalling but as Jack pointed out this is proving difficult. Until we find the cause and fix it, we're torn between a flaky CI (without this PR) or a less-flaky CI that might hide issues. :/ I'm not sure what the best way is to proceed here. WDYT, @jackfranklin?

jackfranklin:
Fully aware this isn't good, but at this point it's a trade off of flakey CI that have we have to kick vs hiding a potential issue.

I'm lost in terms of debugging this further. I'd like to see if this does make our CI less flakey and if so it might be a trade-off worth taking for a bit. If it doesn't, we'll ditch it.

We don't run this flag locally so hopefully we don't introduce any consistent issues via this locally as people won't have this flag on.

AviVahl:
Using [wtfnode](https://github.com/myndzi/wtfnode), we saw some hanging websockets in our setup when we debugged this.

When checking the current code, I found a something odd:
https://github.com/puppeteer/puppeteer/blob/v3.0.4/src/WebSocketTransport.ts#L57
I would expect that line to say:
```ts
return new Promise(resolve => this._ws.close(resolve))
```
and whoever calling `close()` to await the returned promise.

WDYT? could that be a/the reason for the stale process?

ref: https://github.com/websockets/ws/blob/7.3.0/doc/ws.md#serverclosecallback

PR #5887, opened 2020-05-18
Fix for incorrect catch syntax

kbreed:
This was causing npm install to fail.


mathiasbynens:
You're using an unsupported Node.js version. See https://github.com/puppeteer/puppeteer#usage.

PR #5898, opened 2020-05-20
chore: Error for bad Node versions

jackfranklin:
We define the Node version in the `engines` field in `package.json`
however this is only used and mentioned to the user when installing the
package as a dependency [1]. Node/npm is quite noisy when installing
these so it's very easily missed.

We now use the semver package to check our version in the install script
and bail if it's not at least what's in our `engines` field. This fix
isn't perfect because if we use syntax that isn't supported in old
versions the user will get a syntax error rather than our nice error,
but it should help moving forwards as we push Node versions up. e.g.
when we were to drop Node 10 (for the avoidance of doubt, not happening
anytime soon at all) this would help users be aware and know to upgrade.

[1]: https://docs.npmjs.com/files/package.json#engines

### DIFF
--- a/install.js
+++ b/install.js
@@ -26,6 +26,19 @@
 
 const compileTypeScriptIfRequired = require('./typescript-if-required');
 
+const checkNodeVersionSupported = () => {
+  const minRequiredVersion = require('./package.json').engines.node;
+  const semver = require('semver');
+
+  if (!semver.satisfies(process.version, minRequiredVersion)) {
+    console.error(`Error: Node version ${process.version} is not supported.`);
+    console.error(
+      `You must be running at least ${minRequiredVersion} to use Puppeteer.`
+    );
+    process.exit(1);
+  }
+};
+
 const firefoxVersions =
   'https://product-details.mozilla.org/1.0/firefox_versions.json';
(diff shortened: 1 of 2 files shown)


mathiasbynens:
I'm not convinced this change pays for itself / is worth the extra dependency. The supported Node.js version is already defined in `engines` as you point out, and our general policy on supported Node.js versions is [well-documented](https://github.com/puppeteer/puppeteer#usage).

Generally it's a good idea to provide helpful error messages, but in this case I don't think it's worth introducing a new dependency.

WDYT?

jackfranklin:
Hmmmmm yeah fair point. I might make a PR to make the expected version slightly clearer and easier to spot on the README.

mathiasbynens:
> I might make a PR to make the expected version slightly clearer and easier to spot on the README.

Sounds good. If we do this we should also update https://github.com/puppeteer/puppeteer/blob/master/CONTRIBUTING.md#for-project-maintainers to include an explicit step updating the Node.js version in the README for each release (since it would otherwise be forgotten at times). Alternatively, we could point to the `package.json#engines` from the README instead of listing a specific version number.

PR #5901, opened 2020-05-20
feat(api): add page.emulateVisionDeficiency(type)

mathiasbynens:
~~To make it possible for contributors to check this out locally, I've temporarily checked in a random Chromium roll for a binary that works on macOS. Before merging, we'd separate this out to a separate commit which rolls Chromium + updates `src/protocol.d.ts`.~~ This has now happened as part of #5879.


src/Page.ts
```diff
@@ -919,6 +919,29 @@ export class Page extends EventEmitter {
...
+      'protanopia',
+      'tritanopia',
+    ]);
+    try {
+      assert(
+        !type || visionDeficiencies.has(type),
+        `Unsupported vision deficiency: ${type}`
+      );
+      await this._client.send('Emulation.setEmulatedVisionDeficiency', {
+        // @ts-ignore: What's the proper way of dealing with this? */
```
mathiasbynens: @jackfranklin Any ideas here?

```
src/Page.ts:938:9 - error TS2322: Type 'string' is not assignable to type '"none" | "achromatopsia" | "blurredVision" | "deuteranopia" | "protanopia" | "tritanopia"'.

938         type: type || 'none',
            ~~~~

  src/protocol.d.ts:4661:13
    4661             type: "none"|"achromatopsia"|"blurredVision"|"deuteranopia"|"protanopia"|"tritanopia";
                     ~~~~
    The expected type comes from property 'type' which is declared here on type 'setEmulatedVisionDeficiencyParameters'
```

jackfranklin: I think in the function parameter you don't want to use `type?: string` but `type?: "none" | "achromatopsia" | ...`

mathiasbynens: Gotcha, thanks!

OrKoN (approved):
LGTM

test/emulation.spec.js
```diff
@@ -351,4 +351,65 @@ describe('Emulation', () => {
...
+        const screenshot = await page.screenshot();
+        expect(screenshot).toBeGolden('vision-deficiency-tritanopia.png');
+      }
+
+      {
+        await page.emulateVisionDeficiency('none');
+        const screenshot = await page.screenshot();
+        expect(screenshot).toBeGolden('screenshot-sanity.png');
+      }
+    });
```
jackfranklin, line 403: nit: I'd personally have one individual `it` for each test rather than the blocks (I think it leads to nicer errors if a test fails) but take it or leave it

src/Page.ts
```diff
@@ -920,6 +920,36 @@ export class Page extends EventEmitter {
...
+      | 'tritanopia'
+  ): Promise<void> {
+    const visionDeficiencies = new Set([
+      'none',
+      'achromatopsia',
+      'blurredVision',
+      'deuteranopia',
+      'protanopia',
+      'tritanopia',
+    ]);
```
jackfranklin: I think we could define the vision deficiency type as an enum here and take advantage of their runtime behaviour to avoid duplicating this list, maybe something like this: https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time

So we define the enum:

```
enum VisionDeficiency {
  none,
  achromatopsia,
  ... 
} 
```

And then create a set using it because it will actual exist at runtime:  

```
const visionDefs = new Set(Object.keys(VisionDeficiency))
```

Or similar. It might not be exactly this but it will clear up some duplication here. Does `protocol.d.s` generate any types for this set that we can use?

test/emulation.spec.js
mathiasbynens: I see what you're saying. I think the reason it's done this way for existing tests is because we'd otherwise have to duplicate the `page.setViewport` + `page.goto` setup for each subtest.

jackfranklin (approved):

PR #5902, opened 2020-05-20
[Do not merge] debugging a certification test failure locally that does not reproduce on CI

jackfranklin:


jackfranklin:
Figured it! It's a version of OS X issue locally vs travis. PR that solves the issue incoming.

PR #5903, opened 2020-05-20
chore: update Travis to run latest macOS and fix HTTPS test

jackfranklin:


test/navigation.spec.js
```diff
@@ -146,6 +146,10 @@ describe('navigation', function () {
         expect(error.message).toContain('Cannot navigate to invalid URL');
       else expect(error.message).toContain('Invalid url');
     });
+
+    /* If you are running this on Mac OSX that's pre Catalina this will fail locally.
```
mathiasbynens: Nit:

```suggestion
    /* If you are running this on pre-Catalina versions of macOS this will fail locally.
```

mathiasbynens (approved):

Issue #1695, opened 2017-12-30
Allow Chromium extension url pages and webviews to be recognized as regular pages


gpetrov:
Currently if you open pages with urls like:  chrome-extension://fjjllfpmlaejicjkpknbdicd...
Those get:

```
_targetInfo:
   { targetId: '(769AF8566CD3629647D0BC2EFBD10E1)',
     type: 'app',
```

so the type is **app** and target.page() doesn't work to obtain the target page, due to this check:

```javascript
if (this._targetInfo.type === 'page' && !this._pagePromise) {
```
at https://github.com/GoogleChrome/puppeteer/blob/master/lib/Browser.js#L194

The same is also valid for pages within a webview
```
 _targetInfo:
   { targetId: '(28682BC78076959FC22B67763A8251DD)',
     type: 'webview',
```

so the check should be changed to 
```javascript
if ((this._targetInfo.type === 'page' || this._targetInfo.type === 'app' || this._targetInfo.type === 'webview') && !this._pagePromise) {
```

aslushnikov:
We're conservative in exposing target types since not all parts of API might work with certain types.

For the webview case, do you connect puppeteer to the chromium on android?

gpetrov:
@aslushnikov  I'm trying to use Puppeteer with https://nwjs.io/ it will be really useful there for automated UI testing.
It works all really great with the changes above! 

The other thing that was missing the Page.getFrameTree and Page.setLifecycleEventsEnabled as described in https://github.com/GoogleChrome/puppeteer/issues/1681

But this was just because NWjs was one build behind and it will soon be at the latest Chrome 64 as it gets released.

You might want to add a simple check if Page.getFrameTree and Page.setLifecycleEventsEnabled are available before usage so it runs fine on Chrome 63 as well ...

aslushnikov:
@gpetrov I'll close this for now since it doesn't look like there's a lot of interest from the community.

Feel free to send us a PR for this if/when you have time.

GEMI:
Could you please fix this?

PR #5905, opened 2020-05-21
feat: recognize webviews as regular pages

musou1500:
this PR adds new `webview` target type and make `target.page()` method return valid `Page` instance for webviews.
currently, webview is recognized as `other` type so `target.page()` returns `null`.

related to #1695 . but this PR does not adds `app` type.

### DIFF
--- a/src/Target.ts
+++ b/src/Target.ts
@@ -81,7 +81,8 @@ export class Target {
   async page(): Promise<Page | null> {
     if (
       (this._targetInfo.type === 'page' ||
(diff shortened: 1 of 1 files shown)


jackfranklin:
@mathiasbynens please could you take a look?

mathiasbynens (approved):

lightnessking:
Apparently this was merged in may 2020 and release 5.5.0 was in november 2020. However, page() still returns null if target is not "page"` or `"background_page". How can I get this to work?

PR #5907, opened 2020-05-21
Use Mitt as the Event Emitter

jackfranklin:
**Important** this PR is a breaking change.  

Swaps out our dependency on Node's event emitter for Mitt, which is agnostic of environment.

Still to be done: 

- [x] wait for https://github.com/developit/mitt/pull/99 to ship and upgrade Mitt
- [x] update docs to make it clear that we don't extend Node's EventEmitter

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -335,6 +335,13 @@
   * [coverage.stopCSSCoverage()](#coveragestopcsscoverage)
   * [coverage.stopJSCoverage()](#coveragestopjscoverage)
 - [class: TimeoutError](#class-timeouterror)
+- [class: EventEmitter](#class-eventemitter)
+  * [eventEmitter.emit(event, [eventData])](#eventemitteremitevent-eventdata)
+  * [eventEmitter.listenerCount(event)](#eventemitterlistenercountevent)
+  * [eventEmitter.off(event, handler)](#eventemitteroffevent-handler)
+  * [eventEmitter.on(event, handler)](#eventemitteronevent-handler)
+  * [eventEmitter.once(event, handler)](#eventemitteronceevent-handler)
+  * [eventEmitter.removeListener(event, handler)](#eventemitterremovelistenerevent-handler)
 <!-- GEN:stop -->
 
 ### Overview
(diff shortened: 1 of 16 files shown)


mathiasbynens:
Excited to see this! Looking good so far. Feel free to ping me for review once you no longer consider this WIP 👍

tsconfig.json
```diff
@@ -5,7 +5,8 @@
     "outDir": "./lib",
     "target": "ESNext",
     "moduleResolution": "node",
-    "module": "CommonJS"
+    "module": "CommonJS",
+    "esModuleInterop":true
```
jackfranklin: This change makes it easier to consume libraries from npm - without this we had to do some ugly jumps in code to satisfy TS + ensure CJS compatibility in our output. Didn't feel worth fighting.

docs/api.md
```diff
@@ -3935,6 +3942,35 @@ reported.
...
+
+#### eventEmitter.once(event, handler)
+- `event` <[string]|[symbol]> the event to add the handler to.
+- `handler` <[Function]> the event listener that will be added.
+
+#### eventEmitter.removeListener(event, handler)
+- `event` <[string]|[symbol]> the event to remove the handler from.
+- `handler` <[Function]> the event listener that will be removed.
+
+This method is identical to `off` and maintained for compatibility with Node's EventEmitter. We recommend using `off` by default.
```
jackfranklin, line 3973: @mathiasbynens I'm tempted to not even document this method given we want people to use `off`. WDYT? We only need this as Node streams which we use in Puppeteer use `removeListener` so for now we need our emitter to keep that same interface. 

Or we just don't use `off` and instead use `removeListener`?

utils/doclint/generate_types/index.js
```diff
@@ -19,7 +19,6 @@ const objectDefinitions = [];
     path.join(__dirname, '..', '..').length
   )}
 import { ChildProcess } from 'child_process';
-import { EventEmitter } from 'events';
```
jackfranklin: this is a quick fix to get this branch passing CI. On doing so I realised this file is not needed so https://github.com/puppeteer/puppeteer/pull/5932 removes it completely.

docs/api.md
mathiasbynens: Hmm, I like the way you've done it in this PR. Describe what's there explicitly but point folks towards a specific preferred solution. Let's keep this as-is?

src/EventEmitter.ts
```diff
@@ -0,0 +1,61 @@
+import mitt, { Emitter, EventType, Handler } from 'mitt';
+
+export interface CommonEventEmitter {
+  on(event: EventType, handler: Handler): void;
+  off(event: EventType, handler: Handler): void;
+  /* To maintain parity with the built in NodeJS event emitter which uses removeListener
+   * rather than `off`.
+   * if you're implementing new code you should use off
```
mathiasbynens:
```suggestion
   * If you're implementing new code you should use `off`.
```

tsconfig.json
mathiasbynens:
```suggestion
    "esModuleInterop": true
```

mathiasbynens: ack

mathiasbynens (approved):

PR #5934, opened 2020-05-28
chore: rename Request class to HTTPRequest

jackfranklin:
It conflicts with an inbuilt TypeScript `Request` type so can cause
confusion when in TS land. Note: `Response.ts` and `Worker.ts` also
suffer from this; PRs to rename them are incoming.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -281,20 +281,20 @@
   * [elementHandle.toString()](#elementhandletostring)
   * [elementHandle.type(text[, options])](#elementhandletypetext-options)
   * [elementHandle.uploadFile(...filePaths)](#elementhandleuploadfilefilepaths)
-- [class: Request](#class-request)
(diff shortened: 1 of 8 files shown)


mathiasbynens (approved):

jackfranklin:
@mathiasbynens worth noting again this is a sort-of-not-really breaking change. Given we don't expose the constructor for `Request`, it's not easily done, but people could be requiring it from `puppeteer/lib/Request` which will break, but I think we agreed that we don't support people reaching into the internal file structure and as such we're not going to treat things like renames as breaking changes. I can't imagine anyone is relying on the name of this class being `Request`

mathiasbynens:
Ack. I don't consider `require`ing implementation details such as `puppeteer/lib/Request` directly something we should support, and am okay with not labeling this as a breaking change.

PR #5941, opened 2020-05-29
chore: rename Worker to WebWorker

jackfranklin:


mathiasbynens (approved):
This is actually a much better name, since conceptually "workers" includes web workers and service workers, but this class does not. 👍

PR #5954, opened 2020-06-01
chore: Defer Windows FF test for longer

jackfranklin:
We deferred this initially because our Windows CI built wasn't stable
and so debugging this was hard. It's now much more stable so let's push
this back a month but at the same time I'll reach out to the Moz folks
as it should be easier to debug reliably now CI is stable on Windows.


jackfranklin:
Here's the issue about this test: https://github.com/puppeteer/puppeteer/issues/5955

paullewis (approved):

PR #5964, opened 2020-06-02
Add test configuration options for running tests against multiple products

mjzffr:
This is in support of Mozilla's internal sync to Puppeteer 3.1 at https://bugzilla.mozilla.org/show_bug.cgi?id=1632710

Notably add `itChromeOnly`, `itRegularInstallOnly`

Also unskip more Firefox tests in Puppeteer CI.

### DIFF
--- a/src/BrowserFetcher.ts
+++ b/src/BrowserFetcher.ts
@@ -415,9 +415,6 @@ function extractTar(tarPath: string, folderPath: string): Promise<unknown> {
     tarStream.on('error', reject);
     tarStream.on('finish', fulfill);
     const readStream = fs.createReadStream(tarPath);
-    readStream.on('data', () => {
-      process.stdout.write('\rExtracting...');
-    });
     readStream.pipe(bzip()).pipe(tarStream);
   });
 }
--- a/test/navigation.spec.js
+++ b/test/navigation.spec.js
@@ -57,15 +57,12 @@ describe('navigation', function () {
       const response = await page.goto('about:blank');
       expect(response).toBe(null);
     });
-    itFailsFirefox(
-      'should return response when page changes its URL after load',
-      async () => {
-        const { page, server } = getTestState();
+    it('should return response when page changes its URL after load', async () => {
+      const { page, server } = getTestState();
 
-        const response = await page.goto(server.PREFIX + '/historyapi.html');
-        expect(response.status()).toBe(200);
-      }
-    );
+      const response = await page.goto(server.PREFIX + '/historyapi.html');
+      expect(response.status()).toBe(200);
+    });
     it('should work with subframes return 204', async () => {
       const { page, server } = getTestState();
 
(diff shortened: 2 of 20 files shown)


mathiasbynens (approved):
Excited about the many unskipped tests! LGTM, but @jackfranklin PTAL as well.

src/BrowserFetcher.ts
```diff
@@ -415,9 +415,6 @@ function extractTar(tarPath: string, folderPath: string): Promise<unknown> {
     tarStream.on('error', reject);
     tarStream.on('finish', fulfill);
     const readStream = fs.createReadStream(tarPath);
-    readStream.on('data', () => {
-      process.stdout.write('\rExtracting...');
-    });
```
jackfranklin, line 420: I've been meaning to get rid of this for so long, thank you!

jackfranklin (approved):
So exciting to see many more tests not skipped!

mathiasbynens:
This is failing on CI on macOS with the following:

```
  1) navigation
       Page.goto
         should fail when navigating to bad SSL after redirects:
     Error: expect(received).toContain(expected) // indexOf
Expected substring: "net::ERR_CERT_AUTHORITY_INVALID"
Received string:    "net::ERR_CERT_INVALID at https://localhost:8908/redirect/1.html"
      at Context.it (test/navigation.spec.js:183:31)
```

@jackfranklin does Travis somehow get https://github.com/puppeteer/puppeteer/blob/6cfe142af1cc8188d8011fff00e8b37d480934c7/test/navigation.spec.js#L151-L159 wrong? Any ideas what's going on here?

To resolve this, instead of the current check, we could just check for the presence of both the substrings `'ERR_CERT_'` and `'_INVALID'`. WDYT?

test/navigation.spec.js
```diff
@@ -179,22 +170,19 @@ describe('navigation', function () {
...
+      const { page, server, httpsServer, isChrome } = getTestState();
+
+      server.setRedirect('/redirect/1.html', '/redirect/2.html');
+      server.setRedirect('/redirect/2.html', '/empty.html');
+      let error = null;
+      await page
+        .goto(httpsServer.PREFIX + '/redirect/1.html')
+        .catch((error_) => (error = error_));
+      if (isChrome)
+        expect(error.message).toContain('net::ERR_CERT_AUTHORITY_INVALID');
```
jackfranklin: @mjzffr I think you need to change the assertion here back to what it was - the `EXPECTED_SSL_CERT_MESSAGE` constant. See the comment further up this file about some slight oddness with Chrome + Mac versions that Travis hits vs what we get locally. Did you need to change this to get the test passing for you? There may well be a better way of dealing of this.

mjzffr: This change may just have been the result of a bad merge on my end.

jackfranklin:
@mjzffr this looks good except you left the `describe.only` in - once that's gone I think this should be green and we can ship it.

jackfranklin:
I think a webhook didn't fire here because the build is green on Travis but GH doesn't know about it...going to prod it to rebuild.

PR #5967, opened 2020-06-03
[WIP] chore: Introduce API Extractor and start generating documentation

jackfranklin:
This PR introduces [API Extractor](https://api-extractor.com/) to the codebase and generates new Markdown documentation into the `new-doc` folder. _This documentation is incomplete and not meant to be used!_ 

This PR doesn't include the documentation in the `new-doc` folder in order to make this PR reviewable. If this lands, I will follow it up with another PR that generates the docs.

Once this PR lands we can start slowly documenting our code (by converting the `api.md` docs into TSDoc comments) and once we think the new docs are ready we can make them the primary documentation and remove the old documentation.

@mathiasbynens we could exclude `new-docs` from Git for now, but I think it's useful to see them rendered in pull requests when we begin work on improving them so I've left them in for now. LMK what you think.


src/Browser.ts
```diff
@@ -22,7 +22,7 @@ import Protocol from './protocol';
 import { Connection } from './Connection';
 import { Page } from './Page';
 import { ChildProcess } from 'child_process';
-import type { Viewport } from './PuppeteerViewport';
+import { Viewport } from './PuppeteerViewport';
```
jackfranklin, line 25: API Extractor doesn't work fully with `import type` yet so we should use regular imports. This isn't really an issue for us anyway.

src/Puppeteer.ts
```diff
@@ -14,21 +14,31 @@
...
+import { DevicesMap } from './/DeviceDescriptors';
 import { Browser } from './Browser';
-import * as QueryHandler from './QueryHandler';
+import {
+  registerCustomQueryHandler,
+  unregisterCustomQueryHandler,
+  customQueryHandlers,
+  clearQueryHandlers,
+  QueryHandler,
+} from './QueryHandler';
```
jackfranklin, line 36: APIExtractor also doesn't support `import * as` imports just yet, but it [is being worked on](https://github.com/microsoft/rushstack/pull/1796). I felt this change was fine to make to unblock our progress.

src/QueryHandler.ts
```diff
@@ -74,11 +74,3 @@ export function getQueryHandlerAndSelector(
...
   };
 }
-
-module.exports = {
-  registerCustomQueryHandler,
-  unregisterCustomQueryHandler,
-  customQueryHandlers,
-  getQueryHandlerAndSelector,
-  clearQueryHandlers,
-};
```
jackfranklin, line 84: didn't need these as well as the ESM exports

tsconfig.json
```diff
@@ -5,7 +5,9 @@
     "outDir": "./lib",
     "target": "ESNext",
     "moduleResolution": "node",
-    "module": "CommonJS"
+    "module": "CommonJS",
+    "declaration": true,
+    "declarationMap": true
```
jackfranklin, line 10: API Extractor works off the `.d.ts` files so we need to generate them now.

src/Puppeteer.ts
mathiasbynens: ack

tsconfig.json
mathiasbynens: ack

mathiasbynens (approved):
LGTM, and checking in `new-docs` seems fine.

PR #5968, opened 2020-06-03
Error Handling When Checking Version of Firefox

david-dest01:
On line `175` of `install.js` there's a missing error within the `try... catch` statement, resulting in a syntax error when installing.

### DIFF
--- a/install.js
+++ b/install.js
@@ -172,7 +172,7 @@ async function download() {
             try {
               const versions = JSON.parse(data);
               return resolve(versions.FIREFOX_NIGHTLY);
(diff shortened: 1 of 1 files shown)


david-dest01:
@googlebot I signed it!

MrObe13 (approved):

mathiasbynens:
This is not missing. 

You're using an old, [unsupported](https://github.com/puppeteer/puppeteer#usage) Node.js version that lacks support for [optional `catch` binding](https://v8.dev/features/optional-catch-binding).

PR #5979, opened 2020-06-08
chore: migrate away from Node's EventEmitter

jackfranklin:
This PR re-introduces Mitt (that was released in v3.2 and reverted in v3.3).

The majority of EventEmitter methods are re-implemented in Puppeteer and we will release this as a major version accompanied by release notes that document exactly which methods have been removed to help users migrate.


docs/api.md
```diff
@@ -3935,6 +3942,35 @@ reported.
...
+
+#### eventEmitter.once(event, handler)
+- `event` <[string]|[symbol]> the event to add the handler to.
+- `handler` <[Function]> the event listener that will be added.
+
+#### eventEmitter.removeListener(event, handler)
+- `event` <[string]|[symbol]> the event to remove the handler from.
+- `handler` <[Function]> the event listener that will be removed.
+
+This method is identical to `off` and maintained for compatibility with Node's EventEmitter. We recommend using `off` by default.
```
jackfranklin, line 4033: these docs are not up to date and need to be written again.

mathiasbynens (approved):
LGTM % the TODO you outlined

test/page.spec.js
```diff
@@ -130,6 +131,25 @@ describe('Page', function () {
...
+      page.off('response', handler);
+      await page.goto(server.EMPTY_PAGE);
+      // Still one because we removed the handler.
+      expect(handler.callCount).toBe(1);
+      page.on('response', handler);
+      await page.goto(server.EMPTY_PAGE);
+      // Two now because we added the handler back.
+      expect(handler.callCount).toBe(2);
+    });
+  });
```
jackfranklin, line 155: a user reported a bug in 3.2 where this code didn't work. I don't think it was a bug in Mitt, but I wanted to add a test that explicitly listened to an event, stopped, and then started again just to make sure. It's a useful test to have IMO.

PR #5981, opened 2020-06-08
feat(new-docs): migrate Dialog documentation to TSDoc

jackfranklin:


src/Dialog.ts
```diff
@@ -16,29 +16,42 @@
 
 import { assert } from './helper';
 import { CDPSession } from './Connection';
+import Protocol from './protocol';
 
-/* TODO(jacktfranklin): protocol.d.ts defines this
- * so let's ditch this and avoid the duplication
+/**
+ * Dialog instances are dispatched by `Page` via the `dialog` event.
```
jackfranklin: once we have documented `Page` and the events we can turn these into links to those docs.

src/Page.ts
```diff
@@ -738,11 +738,16 @@ export class Page extends EventEmitter {
 
   _onDialog(event: Protocol.Page.javascriptDialogOpeningPayload): void {
     let dialogType = null;
-    if (event.type === 'alert') dialogType = Dialog.Type.Alert;
```
jackfranklin, line 741: this `Dialog.Type` only existed for this method and was never documented. So I removed it to save having it (and therefore having to document it).

mathiasbynens (approved):

PR #5982, opened 2020-06-08
chore: clean `lib` when running tsc

jackfranklin:
I've had misleading type errors due to left over builds or tests
behaving oddly. We should just strip the entire folder out before
building again so there's no left over artefacts that could cause
issues.

The `|| true` in the command is so if `rm` errors because the folder
doesn't exist, it doesn't exit with an error code.


package.json
```diff
@@ -27,7 +27,8 @@
     "eslint-fix": "eslint --ext js --ext ts --fix .",
     "lint": "npm run eslint && npm run tsc && npm run doc",
     "doc": "node utils/doclint/cli.js",
-    "tsc": "tsc --version && tsc -p . && cp src/protocol.d.ts lib/",
+    "clean-lib": "rm -r lib || true",
```
mathiasbynens:
```suggestion
    "clean-lib": "rm -rf lib",
```

mathiasbynens (approved):
LGTM

> The `|| true` in the command is so if `rm` errors because the folder
> doesn't exist, it doesn't exit with an error code.

Seems simpler to use `-f`. WDYT?

PR #5993, opened 2020-06-10
feat(api): add Page.isJavaScriptEnabled + mark properties as private

jackfranklin:
Just one was used externally and I wrapped that up in a method. I think
it's a useful method to provide (I can imagine wanting to know if JS is
enabled on a page) so I think there's no harm here (I'd rather that then
have JSHandle reach into a private variable).


mathiasbynens (approved):
LGTM. Please mark this as `feat(api):` in the commit message.

PR #5994, opened 2020-06-10
chore: ensure new-docs are up to date

jackfranklin:
This adds a Travis check that if we re-generate the docs we get the
exact output (by checking if the Git tree is dirty).

We do the dirty check by using `git status --porcelain`, seeing how many
lines that outputs, and using that as the exit code (taking only the
first 255 lines to avoid invalid exit codes). `--porcelain` makes the
output be empty if the repo is not dirty in anyway which translates into
an exit code of 0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "compare-protocol-d-ts": "node utils/protocol-types-generator compare",
     "test-install": "scripts/test-install.sh",
     "generate-docs": "npm run tsc && api-extractor run --local --verbose && api-documenter markdown -i temp -o new-docs",
-    "ensure-new-docs-up-to-date": "npm run generate-docs && exit (git status --porcelain | head -255 | wc -l)"
+    "ensure-new-docs-up-to-date": "npm run generate-docs && exit `git status --porcelain | head -255 | wc -l`"
   },
   "files": [
     "lib/",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -33,7 +33,8 @@
     "update-protocol-d-ts": "node utils/protocol-types-generator update",
     "compare-protocol-d-ts": "node utils/protocol-types-generator compare",
     "test-install": "scripts/test-install.sh",
-    "generate-docs": "npm run tsc && api-extractor run --local --verbose && api-documenter markdown -i temp -o new-docs"
+    "generate-docs": "npm run tsc && api-extractor run --local --verbose && api-documenter markdown -i temp -o new-docs",
+    "ensure-new-docs-up-to-date": "npm run generate-docs && exit `git status --porcelain | head -255 | wc -l`"
```
mathiasbynens, line 37:
```suggestion
    "ensure-new-docs-up-to-date": "npm run generate-docs && git diff-index --quiet HEAD"
```

mathiasbynens (approved):
`git diff-index --quiet HEAD` seems simpler, so let’s use that? LGTM otherwise 👍

package.json
jackfranklin: This would be nicer, but doesn't error if there's an untracked file in the repo. e.g.:

```
// start with clean repo
touch foo.txt
git diff-index --quiet HEAD exits with 0 
```

PR #5995, opened 2020-06-10
docs(new): add TSDoc comments to the Frame class

kaycebasques:
@jackfranklin || @mathiasbynens please let me know if I'm on the right track

### DIFF
--- a/src/FrameManager.ts
+++ b/src/FrameManager.ts
@@ -360,6 +360,80 @@ export class FrameManager extends EventEmitter {
   }
 }
 
+/**
+ * At every point of time, page exposes its current frame tree via the [page.mainFrame()](#pagemainframe) and [frame.childFrames()](#framechildframes) methods.
+ *
+ * [Frame] object's lifecycle is controlled by three events, dispatched on the page object:
+ * - ['frameattached'](#event-frameattached) - fired when the frame gets attached to the page. A Frame can be attached to the page only once.
+ * - ['framenavigated'](#event-framenavigated) - fired when the frame commits navigation to a different URL.
+ * - ['framedetached'](#event-framedetached) - fired when the frame gets detached from the page.  A Frame can be detached from the page only once.
+ *
+ * @example
+ * An example of dumping frame tree:
+ *
+ * ```js
+ * const puppeteer = require('puppeteer');
+ *
+ * (async () => {
+ *   const browser = await puppeteer.launch();
+ *   const page = await browser.newPage();
+ *   await page.goto('https://www.google.com/chrome/browser/canary.html');
+ *   dumpFrameTree(page.mainFrame(), '');
+ *   await browser.close();
+ *
+ *   function dumpFrameTree(frame, indent) {
+ *     console.log(indent + frame.url());
+ *     for (const child of frame.childFrames()) {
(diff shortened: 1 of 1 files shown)


src/FrameManager.ts
```diff
@@ -441,6 +515,13 @@ export class Frame {
     return this._mainWorld.evaluate<ReturnType>(pageFunction, ...args);
   }
 
+  /**
+   * Queries the frame for the CSS selector. If there's no such element within the frame, the method resolves to `null`.
```
kaycebasques, line 519: Format `null` as inline code here?

src/FrameManager.ts
```diff
@@ -360,6 +360,80 @@ export class FrameManager extends EventEmitter {
   }
 }
 
+/**
+ * At every point of time, page exposes its current frame tree via the [page.mainFrame()](#pagemainframe) and [frame.childFrames()](#framechildframes) methods.
```
kaycebasques, line 364: Usage of `the` is a bit inconsistent. Here `page` is described without `the` in front of it. But on line 367 it's `the page`.

jackfranklin: you'll need to swap the links for the `@link` syntax that TSDoc uses: https://api-extractor.com/pages/tsdoc/tag_link/

src/FrameManager.ts
```diff
@@ -360,6 +360,80 @@ export class FrameManager extends EventEmitter {
...
+ *
+ * [Frame] object's lifecycle is controlled by three events, dispatched on the page object:
+ * - ['frameattached'](#event-frameattached) - fired when the frame gets attached to the page. A Frame can be attached to the page only once.
+ * - ['framenavigated'](#event-framenavigated) - fired when the frame commits navigation to a different URL.
+ * - ['framedetached'](#event-framedetached) - fired when the frame gets detached from the page.  A Frame can be detached from the page only once.
```
jackfranklin, line 369: Once these are documented on the `Page` class we can link to them but we won't be able to yet - so maybe just make these:

. `framedetached` - fires when the ...

src/FrameManager.ts
```diff
@@ -360,6 +360,80 @@ export class FrameManager extends EventEmitter {
...
+ * @example
+ * An example of getting text from an iframe element:
+ *
+ * ```js
+ *   const frame = page.frames().find(frame => frame.name() === 'myframe');
+ *   const text = await frame.$eval('.selector', element => element.textContent);
+ *   console.log(text);
+ * ```
+ *
+ * @public
```
jackfranklin, line 433: this looks great - I think you'll want to use `@remarks` to split a summary from the full description: https://api-extractor.com/pages/tsdoc/tag_remarks/

src/FrameManager.ts
jackfranklin: 👍

jackfranklin:
@kaycebasques definitely on the right track :) Thank you!

jackfranklin:
@kaycebasques hey, do you have any time soon to finish this off? Doesn't matter at all, I'm very happy to do the final tidying up and merge it, but don't want to if you're planning to :)

jackfranklin:
#6188 picks this up and adds docs for all the frame methods, building off this start. Thanks @kaycebasques :)

Issue #1366, opened 2017-11-12
Feature Request: To have drag and drop feature or mouse hold


shmdhussain:
Feature Request: To have drag and drop feature:

Is there any way to drag one element and drop in target specified, I checked Mouse class, It seems it support basic operations, is there any plan to support drag and drop

even hold the mouse on the element to select and move it and stop mouse hold to drop in the target seems fine for drag and drop.

**Tell us about your environment:**

* Puppeteer version: 0.12
* Platform / OS version:Mac 10.10.5
* URLs (if applicable):NA

aslushnikov:
@shmdhussain did you try using `mouse` to emulate drag-and-drop? I mean, doing precisely

```js
// Do drag-and-drop
mouse.down(..);
mouse.move(..);
mouse.up(..);
```

shmdhussain:
@aslushnikov 

I could click on the element like

`await page.click('#searchResultsSidebar li.search-result-item');`

then How could I use Mouse class in that element, to move it after clicking? . 

Thanks in advance for any help. I am newbie to this puppeeter,

My implementation of drag and drop is baed on jQuery UI. I hope it is based on mouse move

aslushnikov:
@shmdhussain you can do something like this:

```js
const e = await page.$('#searchResultsSidebar');
const box = await e.boundingBox();
await page.mouse.move(box.x + box.width / 2, box.y + boy.height / 2);
await page.mouse.down();
await page.mouse.move(100, 200); // move to (100, 200) coordinates
await page.mouse.up();
```

alexdong:
Thanks @aslushnikov for the code. 

What tripped me over was the `await` in `await e.boundingBox();`. The official documentation for [elementHandle.boudingBox](https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md#elementhandleboundingbox) seems to forget to mention that the `boundingBox` returns a `Promise`.

aslushnikov:
@alexdong thanks, addressed in https://github.com/GoogleChrome/puppeteer/pull/1437

shmdhussain:
@aslushnikov thanks for your help. now as you said it is working. thanks a lot for your time in helping me. since no time for me ,  not able to implement what your have suggested.

Now I did what you mentioned , it is working like a charm. :)

ghost:
@aslushnikov  .. I tried the solution but its not working.
Here is my code :  
[...]

Issue #423, opened 2017-08-20
dragging


tcurdt:
I need to drag something on the screen. I was trying like this:
```
  const mouse = page.mouse
  await mouse.move(x,y)
  await mouse.down()
  await mouse.move(x+400, y)
  await mouse.up()
```
The `down` and `up` work (if I move the mouse between the `down` and `up` events manually the map is being dragged) but the call to `move` between the `down` and `up` seems to get ignored. 

Or what am I missing?

Here is a testcase where the images `1.png` and `2.png` should not look the same:
```
const puppeteer = require('puppeteer');

(async () => {
  const browser = await puppeteer.launch({
    headless: true,
  })

  const page = await browser.newPage()
  await page.setUserAgent('Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/60.0.3112.90 Safari/537.36')
  await page.setViewport({width: 1200, height: 800, deviceScaleFactor: 1});

  await page.goto('https://www.google.de/maps/@52.4889552,13.4707445,1555m/data=!3m1!1e3')
  await page.waitForNavigation({waitUntil: 'networkidle'})

  await page.screenshot({path: '1.png', fullPage: true})

  const innerWidth = await page.evaluate(_ => {return window.innerWidth})
  const innerHeight = await page.evaluate(_ => {return window.innerHeight})

  const mouse = page.mouse
  await mouse.move(innerWidth/2, innerHeight/2)
  await mouse.down()
  await mouse.move(innerWidth/2+200, innerHeight/2)
  await mouse.up()
  await page.waitForNavigation({waitUntil: 'networkidle'})

  await page.screenshot({path: '2.png', fullPage: true})

  browser.close()
})()
```

Garbee:
Duplicate of #343

tcurdt:
Sorry, didn't see it as it's closed.
But also makes me wonder why it is closed :)

aslushnikov:
@JoelEinbinder Joel, could you please take a loot into this?

3zzy:
+1 would be really useful.

aslushnikov:
Looks like google maps are fancy and discard mouse movements with huge delta.

We've added the "steps" option to the `mouse.move` method so that you can send intermediate events. 
The following script works fine for me:
```js
const puppeteer = require('puppeteer');

(async () => {
  const browser = await puppeteer.launch({
    headless: true,
  })

[...]

Issue #343, opened 2017-08-17
Question: possible to simulate drag and drop?


patrick-webs:
Can puppeteer perform drag and drop operations similar to the capabilities provided by [WebDriver/Selenium's Actions class](https://stackoverflow.com/a/14222645/631628)?

JoelEinbinder:
It depends how faithfully you want it to work. You can use `page.mouse` to do a series of down -> move -> up. Drop events will be buggy where they cross into some OS behavior. It is on my list of things to fix.

patrick-webs:
thanks :)

nov-liluo:
I meet the same problem，I use page.mouse move to canvas, but the canvas.ondrop() event did't work.

TNT-Likely:
@JoelEinbinder down->move->up not work for me

TNT-Likely:
![image](https://user-images.githubusercontent.com/9278923/47691847-96b74680-dc2e-11e8-9a43-64e757bdfae4.png)

PR #6000, opened 2020-06-11
Add documentation to Mouse class for copy-pasting text

### LINKED ISSUE #3252: The clipboard contains nothing when use command + c(copy)
### Steps to reproduce

* Puppeteer version: 1.5.0
* Platform / OS version: Mac
* URLs (if applicable):
* Node.js version: 10.7.0

1. select the text in the webpage.
2.  type the following codes:
```javascript
await page.keyboard.down('Meta');
await page.keyboard.down('C');
await page.keyboard.up('Meta');
await page.keyboard.up('C');
```
3.  type the following codes:
```javascript
await page.keyboard.down('Meta');
await page.keyboard.down('V');
await page.keyboard.up('Meta');
await page.keyboard.up('V');
```

In order to test whether our users can copy and paste successfully in our app 

**What happens instead?**
the clipboard contains nothing, which means my " command+c " operation failed.

### LINKED ISSUE #4954: Copy element(licenskey) from one page and add it to another
### Steps to reproduce

* Puppeteer version: 1.9.0
* Platform / OS version: MacOS
* URLs (if applicable):
* Node.js version: 12.10

I working on a project, where i generate a licens key and then create an account using that licens key.. But i happens on 2 different pages. 

Because of workguidelines i cannot share the urls or the code, but i hoped you would be able to help me anyways.

I need to copy the element, where the licens key is generated and insert it on the next page. My puppeteer script is working perfect, but the only part im missing is the "remembering" og copy/paste part of the licens key.

**What happens instead?**

Nothing is happening. I cant even define the value?

TimvdLippe:
There are numerous issues on the Puppeteer issue tracker
about selecting and copy-pasting of text (#3252, #4954,
 #423, #1366, #343) as well as several Stackoverflow questions:

- https://stackoverflow.com/questions/57101467/how-do-you-paste-text-using-puppeteer
- https://stackoverflow.com/questions/49131516/how-to-copy-text-from-browser-clipboard-using-puppeteer-in-nodejs
- https://stackoverflow.com/questions/60158746/how-do-i-access-the-contents-of-the-clipboard-from-within-a-headless-puppeteer-t
- https://stackoverflow.com/questions/56306153/domexception-on-calling-navigator-clipboard-readtext

I myself have spent a considerable amount of time trying to figure
out the combination of required APIs. Therefore, I propose we
include this commonly-asked question in the main documentation.

Fixes #3252
Fixes #4954

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -185,6 +185,7 @@
   * [keyboard.type(text[, options])](#keyboardtypetext-options)
   * [keyboard.up(key)](#keyboardupkey)
 - [class: Mouse](#class-mouse)
+  * [Selecting text and moving the mouse](#selecting-text-and-moving-the-mouse)
   * [mouse.click(x, y[, options])](#mouseclickx-y-options)
   * [mouse.down([options])](#mousedownoptions)
   * [mouse.move(x, y[, options])](#mousemovex-y-options)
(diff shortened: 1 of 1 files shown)


jackfranklin (approved):

docs/api.md
```diff
@@ -2487,6 +2488,45 @@ await page.mouse.move(0, 0);
 await page.mouse.up();
 ```
 
+#### Selecting text and moving the mouse
```
mathiasbynens, line 2491: I haven’t checked, but it could be that pptr.dev doesn’t deal with this header nicely (it might expect that each header corresponds with a method). In that case we could simply remove this header, making the content part of the `Mouse` class description.

mathiasbynens (approved):

docs/api.md
TimvdLippe: Good question and I wouldn't know. Let's see if it breaks when we publish and if yes, your solution sounds 👍

mathiasbynens: Looks like it indeed shows up as a namespace: https://pptr.dev/#?product=Puppeteer&version=master&show=api-class-mouse We could update pptr.dev’s handling of the Markdown but since we’re reworking the docs anyway let’s just remove the header for now: 4205ff79d93c1d27b9cadde6f56c19be04397abd

PR #6001, opened 2020-06-11
docs(new): Start documenting the `Page` class

jackfranklin:
This PR starts exploring the Page class and how to best document it. It explores how best to document events in the system, and I think pulling them out into an `enum` is the best solution here. It lets us end up with a page of docs that explicitly lists all the events the page class can ever emit.


mathiasbynens (approved):

mathiasbynens:
Nit for the commit message: `feat(new-docs):` → `docs(new):`. (`feat` is for new user-facing features/APIs.)

PR #6007, opened 2020-06-12
chore: move `index.js` into `src`

jackfranklin:
This is the first part of a series of pull requests that will slowly
make it possible to initialise Puppeteer for either a Node environment
or a web one. By creating the `initialisePuppeteer` function we can
inject dependencies in for the given version of Puppeteer (e.g. inject a
different debug library for Node vs the web).

This PR moves the initialisation into `src` and calls into it from the
root `index.js`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,7 @@
 
 const { initializePuppeteer } = require('./lib/index');
 
-const puppeteer = initialisePuppeteer({
+const puppeteer = initializePuppeteer({
   packageJson: require('./package.json'),
   rootDirectory: __dirname,
 });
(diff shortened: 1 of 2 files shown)


mathiasbynens (approved):

index.js
```diff
@@ -14,40 +14,11 @@
...
  */
 
-const { helper } = require('./lib/helper');
-const api = require('./lib/api');
-const { Page } = require('./lib/Page');
-for (const className in api) {
-  if (typeof api[className] === 'function')
-    helper.installAsyncStackHooks(api[className]);
-}
+const { initialisePuppeteer } = require('./lib/index');
```
mathiasbynens: We use American English spelling throughout the code base:

```suggestion
const { initializePuppeteer } = require('./lib/index');
```

PR #6011, opened 2020-06-12
Fix childprocess killing when the parent process SIGINTs

### LINKED ISSUE #1825: Zombie Process problem.
Hello,

Recently we talked about this problem in the issues #1823 and #1791.

**Environment:**
- Puppeteer Version: 1.0.
- Chrome Version: 64.0.3282.71 (https://github.com/adieuadieu/serverless-chrome/releases/tag/v1.0.0-34)
- Platform / OS version: AWS Lambda
- Node.js version: 6.10 (https://aws.amazon.com/about-aws/whats-new/2017/03/aws-lambda-supports-node-js-6-10/)

Use Case:
===================
We are using puppeteer on AWS Lambda. We take a screenshot of given HTML template and upload it to S3 and use this image for future requests
It handles over 100 million requests each month. That's why every process should be atomic and immutable. (AWS Lambda has a disk and process limit.)

**Example Code:**

```javascript
  args: ['--disable-gpu', '--no-sandbox', '--single-process', 
             '--disable-web-security', '--disable-dev-profile']
});
await page.goto('https://s3bucket.com/markup/a.html');
const response = await page.screenshot({{ type: 'jpeg', quality: 95 }});
browser.close();
```

Problem
===================
When we are using example code, we got disk error from AWS Lambda. 

**Example `/tmp` folder:**
```
2018-01-12T14:55:38.553Z    a6ef3454-f7a8-11e7-be0f-17f405d5a180    start stdout: total 226084
drwx------ 3 sbx_user1067 479 4096 Jan 12 14:55 .
drwxr-xr-x 21 root root 4096 Jan 12 10:53 ..
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:33 core.headless-chromi.129
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:15 core.headless-chromi.131
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:49 core.headless-chromi.135
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:52 core.headless-chromi.137
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:50 core.headless-chromi.138
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:51 core.headless-chromi.14
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:49 core.headless-chromi.15
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:36 core.headless-chromi.169
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:15 core.headless-chromi.174
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:52 core.headless-chromi.178
-rw------- 1 sbx_user1067 479 15126528 Jan 12 14:50 core.headless-chromi.180
drwx------ 3 sbx_user1067 479 4096 Jan 12 14:14 .pki
``` 
When we investigated these files, we understood that it is a `core dump`. We removed these files after the process completed.

When we monitored process list, we saw `zombie processes` Zombie chrome processes have been growing increasingly. We can't kill them.  AWS Lambda has a maximum process limit. (max 1024 process) That's why we reach the lambda limits.

```
483 1 3.3 1.6 1226196 65408 ? Ssl 22:07 0:05 /var/lang/bin/node --max-old-space-size=870 --max-semi-space-size=54 --max-executable-size=109 --expose-gc /var/runtime/node_modules/awslambda/index.js
483 22 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 73 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 119 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 166 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 214 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 262 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 307 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 353 0.0 0.0 0 0 ? Z 22:07 0:00 [headless-chromi] <defunct>
483 1915 0.0 0.0 0 0 ? Z 22:09 0:00 [sh] <defunct>
```

We couldn't use `dump-init` on lambda. Because lambda already has an `init` system.

How did we fix it? (very hacky method)
===================
We used `browser.disconnect()` instead of `browser.close()`. We manualy managed chrome processes such as `kill`.

**Example Code:**

```javascript
browser.on('disconnected', () => {
    console.log('sleeping 100ms'); //  sleep to eliminate race condition  
    setTimeout(function(){
    console.log(`Browser Disconnected... Process Id: ${process}`);
    child_process.exec(`kill -9 ${process}`, (error, stdout, stderr) => {
        if (error) {
        console.log(`Process Kill Error: ${error}`)
        }
        console.log(`Process Kill Success. stdout: ${stdout} stderr:${stderr}`);
    });
}, 100);
```

Firstly we didn't use this method. We only killed the process after browser disconnect. We got the following error:
```
Error: read ECONNRESET at exports._errnoException (util.js:1018:11) at TCP.onread (net.js:568:26)
``` 

I think it looks like a puppeteer process management problem. When we used this method, we didn't receive any puppeteer related errors. How can we fix it?

Thanks.

### LINKED ISSUE #4333: Timed out after 30000 ms while trying to connect to Chrome!
### Steps to reproduce

* Puppeteer version: 1.14.0
* Platform / OS version: Windows 10 Pro / 17134.112
* Node.js version: 10.15.3
* Npm version : 6.4.1

```js
const main = async url => {
  let browser;
  let results;
  let page;
  let handle;

  // Setup Puppeteer
    args: [
      '--no-sandbox',
      '--headless',
      '--disable-gpu',
      '--window-size=1920x1080',
      '--shm-size=1gb'
    ]
  });

  // Get new page
  page = await browser.newPage();

  // To handle a promise rejection
  process.on('unhandledRejection', (reason, p) => {
    console.error('Unhandled Rejection at: Promise', p, 'reason:', reason);
    browser.close();
  });

  await page.goto(url, {
    timeout: 60000 // 10 minutes
  });

  // Destroy the handle & return results.
  await handle.dispose();

  await browser.close();
  return results;
};

main(url).then(result => {
  // stuff
});
```

No errors.

**What happens instead?**

```
Timed out after 30000 ms while trying to connect to Chrome! 
The only Chrome revision guaranteed to work is r641577
```

### LINKED ISSUE #4796: TimeoutError: Timed out after 30000 ms while trying to connect to Chrome!
Hello

I try to implement/deploy puppeteer on Google Cloud but I have some issues. What I don't understand, it worked well 2 days ago.
Since then, I have this error: 

TimeoutError: Timed out after 30000 ms while trying to connect to Chrome! The only Chrome revision guaranteed to work is r674921 at Timeout.onTimeout 

I ran this command and everything seems fine : DEBUG=* node app.js

Version puppeteer : ^1.19.0
Version Node : v8.11.3

And here the code: 

`async function main_screenshot(project_id) {
  try {
    const browser =  await puppeteer.launch({
        headless: true,
        args: ["--window-size=1440,1000", "--no-sandbox", "--disable-setuid-sandbox", "--disable-gpu"]
      });
    // const browser = await puppeteer.launch({dumpio: true});
    page.setUserAgent('Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.77 Safari/537.36')
    page.setViewport({ width: 1440, height: 721 })
    console.log('entering in screenshot zone')
    page.setDefaultNavigationTimeout(30000)
    await page.goto('https://app.slack.com/client/TLW4JE7AA/CLTFQT9GC')
    await page.waitFor(3000)
    await page.type('#email', process.env.SLACK_EMAIL)
    await page.type('#password', process.env.SLACK_PWD)
    await page.waitFor(3000)
    await page.click('#signin_btn')
    console.log('credentials done')
    await page.waitFor(3000)
    console.log('so?')
    await page.goto('https://app.slack.com/client/TLW4JE7AA/CLTFQT9GC')
    await page.waitFor(3000)
    console.log('connected to the right environment')
    const div = await page.$$('.c-message.c-message--light')
    var nuwwmber_loop= 0
    for (var i = div.length ; i > 0; i--) {
      var text = await (await div[i - 1].getProperty('textContent')).jsonValue();
      
      if(text.includes(project_id)) {
        number_loop = i
        break
      }
    }
    // const text = await (await div.getProperty('textContent')).jsonValue();
    
    
     const shotResult = await div[number_loop-1].screenshot();
    console.log(shotResult)
    cloudinaryOptions = {
      public_id: `${project_id}`
    }

    const results = await uploadToCloudinary(shotResult, cloudinaryOptions)
    console.log(results)
    browser.close()
    return results
  } catch (e) {
    console.log(e)
  }
}`

### LINKED ISSUE #4963: Error generating PDF { TimeoutError: Timed out after 30000 ms while trying to connect to Chrome!
Hi i am using the puppeteer package in my project, it works great when i run it on my localhost but once i deployed it (on google cloud functions) i get the following error:

_Error generating PDF { TimeoutError: Timed out after 30000 ms while trying to connect to Chrome! The only Chrome revision guaranteed to work is r674921 at Timeout.onTimeout (/srv/node_modules/puppeteer/lib/Launcher.js:360:14) at ontimeout (timers.js:498:11) at tryOnTimeout (timers.js:323:5) at Timer.listOnTimeout (timers.js:290:5) name: 'TimeoutError' }_

it happend at the following line in my code:
which is inside the printPdf async function

any suggestions? please help

### LINKED ISSUE #5729: Spawning browser with detached = false
Is it possible to always launch the child process with `detached = false`?

https://github.com/puppeteer/puppeteer/blob/8509f4660efc57e265b66db7e6d35bfe5a217f3d/src/Launcher.js#L77-L88

In some instances, puppeteer browsers linger even after the parent process is closed/killed.

This can be reproduced by using a worker thread.

`p1.js`:
```js
const { Worker, isMainThread } = require('worker_threads')

if (isMainThread) {
  new Worker(__filename)
  console.log('main thread')
} else {
  require('./p2')
}
```

`p2.js`:
```js

console.log('worker thread')

puppeteer.launch({
  headless: false,
  executablePath: '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome',
})
```

Run `node p1`, once the browser opens, hit <kbd>Ctrl+C</kbd>, you'll notice the node process has exited but the browser still lingers.

I'm working around this by sending a cleanup signal from the main thread to the worker thread and manually closing the puppeteer instance in the worker thread. This isn't ideal and has been error prone.

This might help solve #1825

TimvdLippe:
If you `ctrl + c` the Puppeteer parent process, we would sometimes
not properly handle killing of the child processes. This would then
leave child processes behind, with running Chromium instances. This
in turn could block Puppeteer from launching again and results in
cryptic errors.

Instead of using the generic `process.kill` with the process id
(which for some reason is negative the pid, which I don't get),
we can kill the child process directly by calling `proc.kill`.

Additionally, we have to remove the eventlisteners as the last
invocation of a Node emitter callback. If we remove the listener
before the full callback is run, it would prevent further execution.
E.g. if the first call is the removal of the event listener, then the
cleanup later in the callback would not happen.

You can test this locally by running the unittests (`npm run unit`)
and terminating the process prematurely (`ctrl + c`). Before this
change, the process would terminate with exit code 0. With this
change, it now properly terminates with exit code 130 (as defined
by the `BrowserRunner.ts` in the `SIGINT` callback).

Lastly, I added a more helpful error message in the event that
Puppeteer is not able to clean up the Node process. In theory,
this should now no longer happen. However, if it does, we now
explain what is going on and where users can report their bug.

After this change, whenever you run the unit tests or anything
related to Puppeteer, you should have zero browser processes
lingering. E.g. `ps aux | grep Chromium` should show no
results. This therefore prevents the need for continuous killing
of the browser processes when working on Puppeteer.

Fixes #5729
Fixes #4796
Fixes #4963
Fixes #4333
Fixes #1825


TimvdLippe:
This is an attempt to fix the pesky "`ctrl + C` and now my Puppeteer is broken" issue. Pushing this to Travis to see if it is happy. I will need to do more testing locally verify it is actually fixed. Would be great if you could test this too!

TimvdLippe:
Seems like there is only 1 test failing. That seems hopeful. There are some failures on Windows though. But I suppose we can use the same `proc.kill` there.

jackfranklin:
The test failures look like a simple error message has changed too which is great.

Could we write a test that demonstrates this new behaviour?

mocha-config/base.js
```diff
@@ -1,4 +1,6 @@
 module.exports = {
   reporter: 'dot',
+  // Allow `console.logs` to show up during test execution
```
mathiasbynens: nit:

```suggestion
  // Allow `console.log`s to show up during test execution
```

src/launcher/BrowserRunner.ts
```diff
@@ -131,24 +134,32 @@ export class BrowserRunner {
...
+
+    // If the process failed to launch (for example if the browser executable path
+    // is invalid), then the process does not get a pid assigned. A call to
+    // `proc.kill` would error, as the `pid` to-be-killed can not be found.
+    if (this.proc && this.proc.pid && !this.proc.killed) {
+      try {
+        this.proc.kill('SIGKILL');
+      } catch (error) {
+        throw new Error(`${PROCESS_ERROR_EXPLANATION}
+Error cause: ${error.stack}`);
```
mathiasbynens: nit to maintain indentation:

```suggestion
        throw new Error(`${PROCESS_ERROR_EXPLANATION}\nError cause: ${error.stack}`);
```

mathiasbynens (approved):
LGTM! Thanks for looking into this 🎉

PR #5105, opened 2019-10-31
feat: allow overwrite the path for the downloads folder

fengmk2:
e.g.:

```bash
PUPPETEER_DOWNLOAD_PATH=/store/chromiums npm install
```

PR #6014, opened 2020-06-12
Support configuring the browser binary download path

lcabral37:
By adding support for a variable PUPPETEER_DOWNLOAD_PATH it is possible
to support downloading the browser binaries into a folder outside the
node_modules this will in a way allow to preserve previous downloaded
binaries and not required to download then again

This PR is very similar to https://github.com/puppeteer/puppeteer/pull/5105 but with the addition that the puppeteer runner also makes use of the PUPPETEER_DOWNLOAD_PATH without additional configuration


jackfranklin:
@lcabral37 thank you for the PR, appreciate it.  

My worry with this change is that it's becoming a bit complex to understand which download variables do what and how many you may or may not need to set. I'm wondering if there's a nicer way of doing this or we could take an opportunity to tidy up the ENV vars we support in a breaking change release.

I'd also be interested to know the situation that means you need to install the binaries outside of `node_modules` - most CI systems will cache this I think? 

@mathiasbynens what do you think?

lcabral37:
Thanks for the quick feedback. 
I certainly do not have an overview as the maintainers of the project regarding its increasing complexity use of  environment variables but I  can  at least explain the reason behind this PR.
This was a result from a hackday in my organization that was a reflection of the issues I've notice our systems and colleagues dependency on downloading the chromium binaries. 
We build our product with `npm ci`, (whether we are in our development environment or in our build CI service) which will result in purging the `node_modules/` folder so it results in downloading the binaries every time we run this command (which might occur also very often in a developers daily workflow, when changing branches, starting a new task, etc.)  
This behavior has resulted in slowing down the build process. At some point in time, we even had  3 different version dependencies  which would result in requiring to download 3 different chromium binaries( I know ! I know! insane but it has been fixed now). 
I've also seen  issues  with storage.googleapis blocking our IPs (due rate limit), effectively stopping our build servers from building.
Another scenario I found out recently is that some  colleagues are permanently blocked unless they are working within the VPN...  storage.googleapis has permanently blocked their country (a quiet country in Asia)

Anyhow this is the short story of the pain I've seen with having to download the chromium binaries and I understand that this might not event be the best resolution for many of  theses issue (using a mirror host would be better in some of these scenarios for sure) but I honestly think this  approach would also be a major gain in this architecture. I've researched a bit through the documentation and could not find any combination that would mach this workflow. And it would make it possible to persist the same configuration for the download path  while upgrading puppeteer  (it would just download and run the newer revision without additional changes)

To answer your question  regarding CI systems, I can be sure but our experience is that we use `npm ci`  which will build the project with a clean slate  while respecting the package-lock. npm  itself will maintain a local cache of all npm modules downloaded so that running `npm ci` is very fast but it will still run every package `postinstall`  script and in puppeteer case, will perform the download. I mean... this is what we do. I definitely can't say for sure this is the best workflow.

Having bothered you with a 5m read already, I can finish with  saying that if you can identify any changes that could make this PR more future prof or at least less complex, I'm willing to make it so.

jackfranklin:
@lcabral37 thanks for explaining your situation, that makes sense. I can see the benefit of being able to manage where the binary is put yourselves.

@mathiasbynens what do you think to this as a new feature? I think we could rework the ENV vars potentially but at a high level do you think this is a feature worth adding?

lcabral37:
Any status update on this @jackfranklin @mathiasbynens?
Is this worth of adding in a coming release or  just not interesting enough now?

lcabral37:
Still no status update on this @jackfranklin @mathiasbynens?

mathiasbynens:
@lcabral37 We're planning to work on this soon. Please don't ping issues repeatedly asking for updates. If there is an update, you'll see it here.

mathiasbynens (approved):

mathiasbynens:
Longer-term we'll want to change the default download dir to something like `~/.puppeteer`. The ENV var could then be used to restore the current functionality of downloading it to a local subdirectory.

I wonder if we should make this change all at once (to reduce confusion), or land this as-is first, and then change the default later. WDYT?

OrKoN (approved):

OrKoN:
I think this can be landed as-is first because it does not block changing the default in the future.

mathiasbynens:
> I think this can be landed as-is first because it does not block changing the default in the future.

To clarify, my worry is that people will learn about the new env var, and by the time they start to use it, we'll have changed its meaning again. We could avoid this confusion by ensuring the two changes (this + a future patch that changes the default) make it into the same release, but we'd have to decide soonish.

OrKoN:
> > I think this can be landed as-is first because it does not block changing the default in the future.
> 
> To clarify, my worry is that people will learn about the new env var, and by the time they start to use it, we'll have changed its meaning again. We could avoid this confusion by ensuring the two changes (this + a future patch that changes the default) make it into the same release, but we'd have to decide soonish.

By changing the meaning of the env var, do you mean the change of the default value? I think in this case the meaning is still the same: your custom download path would be used. Btw, I am not sure how much effort is it to change the default to something like `~/.puppeteer`?

jackfranklin:
I think 99% of people won't have a strong opinion or care about where Puppeteer downloads things, so I think we can land this and then make a new change in the future and clearly document it as breaking. So I'm happy to ship this as is.

jackfranklin (approved):

hanselfmu (approved):
I'm a little hesitant around default-downloading something to a user's home directory. I think we can ship it as-is, and for me if I don't like where it currently downloads, I could just set the ENV.

I also find this feature quite useful, for example, when I need to switch back and forth my Chromium versions for unit-testing purposes.

johanbay:
I definitely think it makes sense to allow configuring the download path. I see the point in downloading to the home folder to enable sharing binaries across projects or (as here) including the binaries in a CI image, but as a user I would be slightly annoyed if `~/.puppeteer` was the default. Maybe that is just me though, but I would prefer if I had to explicitly enable downloading large binaries to a hidden folder in my home directory 🙂

mathiasbynens:
> By changing the meaning of the env var, do you mean the change of the default value? I think in this case the meaning is still the same: your custom download path would be used.

When the var is set, that path is used, sure. The difference we'd get when changing the default is when it's not set — the "meaning" changes in the sense that you now need to use the var in different cases.

Sounds like there's appetite for not putting large binaries in the home folder by default. Thanks for chiming in, everyone! Let's go with this as-is then.

mathiasbynens:
@lcabral37 Would you be up for rebasing this patch?

lcabral37:
> @lcabral37 Would you be up for rebasing this patch?

Most certainly. 
Rebase done with minimal manual changes to move introduced logic from `install.js` to `src/install.js`

mathiasbynens:
Thanks @lcabral37! I think we can land this soon — I'll just wait for @jschfflr to chime in on the above discussion as well.

jschfflr:
Just my 2 cents: I think, having puppeteer download files to `./.puppeteer` instead of `node_modules` by default would be a better solution because it would minimise the amount of downloads a user would have to do by default. To make sure, people are aware of if it, I'd maybe print a line like: `Downloading chromium to ~/.puppeteer` or `Found an already downloaded chromium version in ~/.puppetter`.
By having it as the default, tools using puppeteer via npx like for example https://github.com/puppeteer/recorder would also benefit from this.

mathiasbynens:
@lcabral37 Looks like `npm run lint` is unhappy, could you please take a look?

mjzffr:
I suggest removing the changes to the `puppeteer-firefox` directory, since it's deprecated and not tested.

lcabral37:
> @lcabral37 Looks like npm run lint is unhappy, could you please take a look?
Sorry for that... when the lint tool is unhappy I'm unhappy :( 

@mjzffr if you insist I can remove it, but for the time being I would prefer to leave it there for consistency. Test wise, I agree it should be tested if it is to be there. so I've performed a quick & dirty manual test to validate firefox downloads to the specified folder (if specified) and that it runs form the specified (if specified)

@jschfflr It would certainly resolve the problem I have but after thinking for a bit I  would dislike this approach as a user as I see that npm modules should be self contained in its ecosystem (node_modules, cache folders, etc ) and not leak out artifacts through the system... Unless I specify otherwise (not trying to defend my PR in this manner), I would find it very annoying to find a `~/.puppetter` or other npm  artifacts (imagine several npm packages start to implement this approach)  that I might not even have installed directly but is instead a dependency of another package. 

Thank you all for your patience

mathiasbynens (approved):

PR #6028, opened 2020-06-16
chore: create new debug module

jackfranklin:


src/Debug.ts
```diff
@@ -0,0 +1,44 @@
...
+ * @example
+ * ```
+ * const log = debug('Page');
+ *
+ * log('new page created')
+ * // logs "Page: new page created"
+ * ```
+ */
+export const debug = (prefix: string): ((...args: unknown[]) => void) => {
+  const isNodeEnv = typeof document === 'undefined';
```
mathiasbynens: nit: Since the result should not change we can do it just once, so let's hoist this check out of the `debug` export.

mathiasbynens (approved):
LGTM % comment

PR #6029, opened 2020-06-16
docs(new): add TSDoc comments to `WebWorker`

AVGP:


src/WebWorker.ts
```diff
@@ -30,12 +30,33 @@ type ExceptionThrownCallback = (
 ) => void;
 type JSHandleFactory = (obj: Protocol.Runtime.RemoteObject) => JSHandle;
 
+/**
+ * The WebWorker class represents a [WebWorker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API).
```
jackfranklin: these need to use the TSDoc @link syntax: https://api-extractor.com/pages/tsdoc/tag_link/

src/WebWorker.ts
```diff
@@ -30,12 +30,33 @@ type ExceptionThrownCallback = (
...
+ * The events `workercreated` and `workerdestroyed` are emitted on the page object to signal the worker lifecycle.
```
jackfranklin, line 37: before this line you probably want to use a `@remarks` to split the summary from the examples, etc: https://api-extractor.com/pages/tsdoc/tag_remarks/

src/WebWorker.ts
```diff
@@ -76,14 +97,31 @@ export class WebWorker extends EventEmitter {
     );
   }
 
+  /**
+   * Returns the URL for the script this web worker is running.
```
jackfranklin: I think in this case we could lose the description in favour of just the `@returns ...` line rather than duplicate them both

jackfranklin (commented):
@AVGP thank you so much for this! Few minor tweaks then I think this is good to go 👍

AVGP:
@jackfranklin Thanks for the review! Slowly getting the hang of this :)

src/WebWorker.ts
```diff
@@ -30,12 +30,35 @@ type ExceptionThrownCallback = (
 ) => void;
 type JSHandleFactory = (obj: Protocol.Runtime.RemoteObject) => JSHandle;
 
+/**
+ * The WebWorker class represents a {@link https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API | WebWorker}.
+ * The events `workercreated` and `workerdestroyed` are emitted on the page object to signal the worker lifecycle.
+ *
+ * @remarks
```
jackfranklin, line 36: nitpick: I'd move the `@remarks` before the sentence about the events. The summary section (the bit before the `@remarks`) should be one short sentence at most.

jackfranklin:
@AVGP one small comment then I think we can ship this!

You need to also run `npm run generate-docs` and commit the diff.

AVGP:
@jackfranklin done :)

jackfranklin (approved):

PR #6032, opened 2020-06-17
docs(new): add TSDoc comments to Puppeteer

AVGP:


src/Puppeteer.ts
```diff
@@ -37,6 +37,23 @@ import {
 
 /**
  * The main Puppeteer class
+ * Puppeteer module provides a method to launch a Chromium instance.
```
jackfranklin: we also support Firefox, so maybe I'd just say "browser instance." here :)

src/Puppeteer.ts
```diff
@@ -47,6 +64,9 @@ export class Puppeteer {
   __productName: string;
   _lazyLauncher: ProductLauncher;
```
jackfranklin, line 65: could we try putting `private` on these fields? It might cause TS errors in which case we can leave them for now and I'll tackle them in a later PR...

src/Puppeteer.ts
```diff
@@ -122,26 +187,86 @@ export class Puppeteer {
...
+   * (async () => {
+   *   const browser = await puppeteer.launch();
+   *   const page = await browser.newPage();
+   *   await page.emulate(iPhone);
+   *   await page.goto('https://www.google.com');
+   *   // other actions...
+   *   await browser.close();
+   * })();
+   * ```
+   * **NOTE** The old way (Puppeteer versions <= v1.14.0) devices can be obtained with `require('puppeteer/DeviceDescriptors')`.
```
jackfranklin: We actually shouldn't say this (I know this was in the old docs and I'll remove it from there) because we shouldn't support importing anything other than `puppeteer` (e.g. our folder structure is subject to change) - so could we lose this line ?

src/Puppeteer.ts
```diff
@@ -122,26 +187,86 @@ export class Puppeteer {
...
+   *
+   * @returns a list of devices to be used with `page.emulate(options)`. Actual list of devices can be found in {@link https://github.com/puppeteer/puppeteer/blob/main/src/DeviceDescriptors.ts | src/DeviceDescriptors.js}.
```
jackfranklin: nit: the `.js` at the end of this sentence should be `.ts`

src/Puppeteer.ts
```diff
@@ -122,26 +187,86 @@ export class Puppeteer {
...
+   * try {
+   *   await page.waitForSelector('.foo');
+   * } catch (e) {
+   *   if (e instanceof puppeteer.errors.TimeoutError) {
+   *     // Do something if this is a timeout.
+   *   }
+   * }
+   * ```
+   *
+   * **NOTE** The old way (Puppeteer versions <= v1.14.0) errors can be obtained with `require('puppeteer/Errors')`.
```
jackfranklin: again, let's lose this note

jackfranklin (commented):
@AVGP this is looking good! Few tiny tweaks and then we can ship this I think. Thank you!

jackfranklin:
Looks like you also need to run Prettier to fix up the formatting (that's what is failing CI). `npm run eslint-fix` will do it.

AVGP:
Done, including the tslint-fix

jackfranklin:
@AVGP if you wouldn't mind dealing with the conflicts, then we can ship this! Or if you'd like me to do the conflicts work and ship it, also happy to. Thanks!

AVGP:
@jackfranklin Done. Just needs a review :)

jackfranklin (approved):

PR #6034, opened 2020-06-17
chore: add dependency chart generation as script

jackfranklin:


jackfranklin:
This is a small PR that adds the script that I use to generate the dependency chart. This is useful as we work to restructure and make Puppeteer agnostic of environments.

jackfranklin:
dependency-chart

Here's what it looks like right now.

mathiasbynens (approved):

PR #6042, opened 2020-06-18
chore: create `common` directory

jackfranklin:
These files will be used by both the web and node versions of Puppeteer.
Another name for this might be "core" but I don't want to cause
confusion with the puppeteer-core package that we publish at the moment.


jackfranklin:
@hanselfmu let me know if you have any questions. This is a mostly mechanical PR that moves files around; no functionality has changed here.

hanselfmu (approved):
LGTM

PR #6043, opened 2020-06-18
chore: remove top level errors and DeviceDescriptors

jackfranklin:
As far as I can tell these became irrelevant as of v1.15 which added
`puppeteer.errors` and `puppeteer.devices [1]. This is a breaking change
but one that's easily mitigated. We've said that we don't consider
changes to our folder/file structure a breaking change, but we can't
really do that if we have these two top level files that we've
documented.

[1]: https://github.com/puppeteer/puppeteer/commit/e3abb0aa32e38660abffad22d6446473be4b13cd#diff-522b24108d7446af4c59873472a90444


jackfranklin:
@mathiasbynens WDYT with regards to this being a breaking change? My hunch is that we should play it better safe than sorry here.

mathiasbynens:
@jackfranklin Version numbers are free. When in doubt, let's just bump another major version. 👍

mathiasbynens (approved):

PR #6049, opened 2020-06-19
chore: move `index.js` into `src` and TS-ify

jackfranklin:
It was causing some infra issues when trying to migrate tests to
TypeScript (that's WIP in another branch that I'll have up soon). It's
unusual to have the entire src in TS except for the main file, which
then reaches into the compiled `lib` directory for the files it needs.

Much better is to move the entry point into TypeScript itself and update
the `main` entry in our `package.json` to point to the compiled output.

This also has the advantange of hooking up all the TS type defs that we
are shipping and will make that process easier too, along with making it
easier to port our tests to TypeScript.


src/initialize.ts
```diff
@@ -0,0 +1,72 @@
...
+    isPuppeteerCore,
+    product
+  );
+
+  // The introspection in `Helper.installAsyncStackHooks` references `Puppeteer._launcher`
+  // before the Puppeteer ctor is called, such that an invalid Launcher is selected at import,
+  // so we reset it.
+  puppeteer._lazyLauncher = undefined;
+  return puppeteer;
+};
```
jackfranklin, line 72: Note: the body of this module is the same, it was renamed from `index.ts` to `initialize.ts`. Nothing in here has changed functionality-wise.

mathiasbynens (approved):

PR #6062, opened 2020-06-22
docs(new): migrate TimeoutError to TSDoc

OrKoN:


OrKoN:
@jackfranklin it looks like the tests fail? Did I do something wrong here?

OrKoN:
It looks like CI is using api-documenter 7.8.13  and my local version after npm install is 7.8.12

jackfranklin:
> It looks like CI is using api-documenter 7.8.13 and my local version after npm install is 7.8.12

Ah, we should probably hard pin this dependency to make sure there are no output differences locally vs CI. I'll do that in a PR now and that should resolve this.

jackfranklin:
https://github.com/puppeteer/puppeteer/pull/6063 should enable this to go green!

jackfranklin (approved):

PR #6063, opened 2020-06-22
chore: pin TSDoc related dependencies

jackfranklin:
Without the API-* dependencies pinned different versions may be
installed on local machines vs CI. One of the checks we do is to check
that the checked in docs matches what is generated on CI. Therefore we
need to ensure devs locally run the exact version that CI runs such that
they generate the same output. So in this case we pin to a particular
version of the dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,8 +59,8 @@
     "ws": "^7.2.3"
   },
   "devDependencies": {
-    "@microsoft/api-documenter": "^7.8.8",
-    "@microsoft/api-extractor": "^7.8.8",
+    "@microsoft/api-documenter": "7.8.14",
+    "@microsoft/api-extractor": "7.8.12",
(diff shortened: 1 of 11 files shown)


OrKoN (approved):

package.json
```diff
@@ -59,8 +59,8 @@
     "ws": "^7.2.3"
   },
   "devDependencies": {
-    "@microsoft/api-documenter": "^7.8.8",
-    "@microsoft/api-extractor": "^7.8.8",
+    "@microsoft/api-documenter": "7.8.14",
```
OrKoN, line 62: should package-lock.json get updated too?

jackfranklin: it's gitignored so no. I'm not sure why it's gitignored but going to leave it be for now.

PR #6065, opened 2020-06-22
docs(new): migrate ConsoleMessage docs to TSDoc

paullewis:


paullewis:
Includes a fix on the console message type to be more restrictive than just `string`, since according to the current docs there's a fixed number of types it can be.

src/common/ConsoleMessage.ts
```diff
@@ -16,20 +16,55 @@
...
+// so it is switched off just for that block.
+/* eslint-disable prettier/prettier */
+
+/**
+ * The supported types for console messages.
+ */
+export type ConsoleMessageType = 'log' | 'debug' | 'info' | 'error' | 'warning' |
+    'dir' | 'dirxml' | 'table' | 'trace' | 'clear' | 'startGroup' |
+    'startGroupCollapsed' | 'endGroup' | 'assert' | 'profile' |
+    'profileEnd' | 'count' | 'timeEnd' | 'verbose';
```
jackfranklin, line 49: does this type already exist in `src/protocol.d.ts` ? If so could we import it to avoid the duplication? It might not (there's ongoing work to get us onto the devtools-protocol npm package) and in that case this is fine

src/common/ConsoleMessage.ts
```diff
@@ -16,20 +16,55 @@
...
   lineNumber?: number;
+
+  /**
+   * 0-based column number in the resource if known or `undefined` otherwise.
+   */
   columnNumber?: number;
 }
 
+// Prettier seems to struggle with the ConsoleMessageType declaration
+// so it is switched off just for that block.
```
jackfranklin, line 40: what do you mean by struggle? https://prettier.io/playground/#N4Igxg9gdgLgprEAuEcAeAHCAnGACGATwzjwGFoBnCAGzgFk5LKBDAczgBVjSBePAOQ0IbAXgA+ggCZwARgFdREwQEsoAMwhjJAuNmw5tggO4tsUNUvEAdKHnvSV2IwKlO0AWxouYLWXR9sFjA4FzA6MxdKX1wAcQN5DG1bB0Fosxh4iESKGhoWDEo4KRcEKSzElxZmPRgXDAN1FQCJFIcBBogmugBRKBLlAUh5WB8VDzg+gZ0ANz1ZCCKBAG5bEAAaEAgMGBUqZFAzA2MABTMESmQQFhpTQkvN2SCwAGs4GABlDGDLZBhseRwTYACxgXgA6sCVPBKN8Qh8LtCVDNoYQrmBmBsQGoirgTkE2B4WMh1DcipsAFaUNAAIWeb0+LAmABk1HASWSgSAqWgPpY6ABFeQQeAcmjkkDfbC4q6yPxwGhYhpqGDglRSGDA5AADgADJtOkVwUEMFcGkw9HMsQBHYXwfHbS4oaoAWigcGKxSx2DgtqccHx7CJYolRQ8Kj+AK5lH5cCFIvZSFJ4q5vlkao1WqQACZNv8WM1LBQPMGUEwAKxY+RFTh+J3JiUzQEASX6CE+YGwKh2AEF+h8iHQQ3AAL4joA seems OK

jackfranklin (approved):
LGTM with nits :)

src/common/ConsoleMessage.ts
paullewis: Yeah I don't know. As per our offline conversation, it seems unhappy in the context of the file itself. Maybe a version difference?

src/common/ConsoleMessage.ts
paullewis: It turns out no. While there is something similar in the protocol, because it's autogenerated we can't get at the values.

src/common/ConsoleMessage.ts
jackfranklin: Yeah maybe! Let's leave this be for now. Not worth blocking.

PR #6068, opened 2020-06-22
docs(new): add TSDoc comments to Keyboard

TimvdLippe:


new-docs/puppeteer.keyboard.md
```diff
@@ -4,33 +4,58 @@
 
 ## Keyboard class
 
+Keyboard provides an api for managing a virtual keyboard. The high level api is , which takes raw characters and generates proper keydown, keypress/input, and keyup events on your page.
```
jackfranklin, line 7: hmmm, seems like API Extractor is missing a link here from the docs ("high level api is , which ..."). Not sure quite why that is.

src/common/Input.ts
```diff
@@ -22,15 +22,81 @@ type KeyDescription = Required<
   Pick<KeyDefinition, 'keyCode' | 'key' | 'text' | 'code' | 'location'>
 >;
 
+/**
+ * Keyboard provides an api for managing a virtual keyboard.
+ * The high level api is {@link Keyboard.type | Keyboard.type},
```
jackfranklin, line 27: seems like it's getting tripped up on this link and not outputting it.

FYI if the thing after the | is the same as the thing before it, you don't need to provide it. `{@link Keyboard.type}` is the same.

jackfranklin (commented):
This looks good bar that odd issue where API Extractor isn't finding the link to `Keyboard.type`. I'll grab the branch down later and have a play.

It'd be good if we could tidy up `{@link X | |}` to just `{@link X}`, thanks :)

PR #6073, opened 2020-06-22
docs(new): migrate ElementHandle to TSDoc

OrKoN:


src/common/JSHandle.ts
```diff
@@ -71,6 +102,17 @@ export class JSHandle {
     return this._context;
   }
 
+  /**
+   * This method passes this handle as the first argument to `pageFunction`.
+   * If `pageFunction` returns a [Promise], then `handle.evaluate` would wait
```
jackfranklin: I think we can lose the `[]` around `Promise`

jackfranklin (approved):

PR #6078, opened 2020-06-23
docs(new): add TSDoc comments to BrowserFetcher

AVGP:


new-docs/puppeteer.browser.md
```diff
@@ -9,7 +9,6 @@
 ```typescript
 export declare class Browser extends EventEmitter 
 ```
-<b>Extends:</b> [EventEmitter](./puppeteer.eventemitter.md)
```
jackfranklin: @AVGP I think you're running an old version of the doc generator - could you `npm install` again to get up to date? Then this diff goes away :)

src/node/BrowserFetcher.ts
```diff
@@ -138,6 +137,27 @@ interface BrowserFetcherRevisionInfo {
   product: string;
 }
 /**
+ * BrowserFetcher can download and manage different versions of Chromium and Firefox.
+ *
+ * @remarks
+ * BrowserFetcher operates on revision strings that specify a precise version of Chromium, e.g. `"533271"`. Revision strings can be obtained from [omahaproxy.appspot.com](http://omahaproxy.appspot.com/).
```
jackfranklin: this needs to be a `@link`

src/node/BrowserFetcher.ts
```diff
@@ -178,18 +201,35 @@ export class BrowserFetcher {
...
+  /**
+   * @returns One of `mac`, `linux`, `win32` or `win64`.
+   */
   platform(): string {
     return this._platform;
   }
 
+  /**
+   * @returns One of `chrome` or `firefox`.
+   */
```
jackfranklin, line 219: we actually could update this return type to use the `Product` type (defined at the top of the file)

src/node/BrowserFetcher.ts
```diff
@@ -178,18 +201,35 @@ export class BrowserFetcher {
     else assert(this._platform, 'Unsupported platform: ' + os.platform());
   }
 
+  /**
+   * @returns One of `mac`, `linux`, `win32` or `win64`.
+   */
```
jackfranklin, line 212: we actually could update this return type to use the `Platform` type (defined at the top of the file)

jackfranklin (commented):
Couple of minor changes but otherwise this is great!

new-docs/puppeteer.browser.md
mathiasbynens: These issues (contributors using older dep versions) are popping up regularly now. I wonder if we should add Git hooks that re-run `npm install` after `git pull`ing any changes to `package.json`, like we do for v8.dev? https://github.com/v8/v8.dev/blob/dc477272d0aff653ffe5a7acb53246af8f3e7c7b/package.json#L24-L30

AVGP: +1 to this. I don't run `npm i` on every pull and I doubt most people do.

mathiasbynens (approved):

PR #6079, opened 2020-06-23
chore: avoid disabling prettier

jackfranklin:


jackfranklin:
@paullewis PTAL. This worked fine for me with no extra changes. I think locally you might be on an old Prettier? Please make sure your editor is using the one from the `node_modules` of this package.

mathiasbynens (approved):

PR #6081, opened 2020-06-23
chore: fix ESLint violation and lessen rule strictness

jackfranklin:


.eslintrc.js
```diff
@@ -100,6 +100,11 @@ module.exports = {
                 "@typescript-eslint/semi": 2,
                 "@typescript-eslint/no-empty-function": 0,
                 "@typescript-eslint/no-use-before-define": 0,
+                // We have to use any on some types so the warning isn't valuable.
```
paullewis, line 103: :(

test/emulation.spec.ts
```diff
@@ -67,7 +67,7 @@ describe('Emulation', () => {
       function dispatchTouch() {
         let fulfill;
         const promise = new Promise((x) => (fulfill = x));
-        window.ontouchstart = function (e) {
+        window.ontouchstart = () => {
```
paullewis, line 70: But why though?

paullewis (approved):
LGTM with curiosity question

test/emulation.spec.ts
jackfranklin: the majority of the codebase + tests use arrow functions - this one stood out to me so I just swapped it out

.eslintrc.js
jackfranklin: Yeah :( Long term we'll use the TS compiler strictness options to get rid of this but these warnings are adding no value.

PR #6085, opened 2020-06-23
docs(new): migrate HTTPResponse docs to TSDoc

hanselfmu:


src/common/HTTPResponse.ts
```diff
@@ -1,29 +1,37 @@
...
  *
  * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
```
jackfranklin: Did you rewrap this bit? Could we revert this diff just so the license is consistent in all the files? :)

src/common/HTTPResponse.ts
```diff
@@ -67,38 +78,70 @@ export class HTTPResponse {
...
+   */
   url(): string {
     return this._url;
   }
 
+  /**
+   * @returns True if the response was successful (status in the range 200-299).
+   */
   ok(): boolean {
+    // TODO: document === 0 case?
```
jackfranklin, line 108: @mathiasbynens do you have any context to this case? I'm not sure why we'd have this.

src/common/HTTPResponse.ts
```diff
@@ -1,29 +1,37 @@
...
+/**
+ * @public
+ */
+export interface RemoteAddress {
   ip: string;
   port: number;
 }
 
+/**
+ * The HTTPResponse class represents responses which are received by page.
```
jackfranklin: Maybe we could link to the page class here? `responses which are received by instances of the {@link Page} class`

jackfranklin (commented):
Minor comments, looks great on the whole, thank you!

src/common/HTTPResponse.ts
mathiasbynens: I don't know, but I would _guess_ this is used for errors before we can even get an HTTP response + status code, e.g. certificate issues. Maybe @sigurdschneider knows?

src/common/HTTPResponse.ts
hanselfmu: Yes I did. Let me change this part back. Enough for the laziness of cmd+A and alt+Q I guess :P

src/common/HTTPResponse.ts
hanselfmu: Yes and I like the minor tweak of the grammar.

jackfranklin (approved):
This looks good to me ! If we're not sure on the status code 0 case, happy to merge as is.

jackfranklin:
I'm going to merge this, and we can make another PR to address the status = 0 case.

PR #6086, opened 2020-06-24
docs(new): Adds TSDoc to Mouse class

AVGP:


src/common/Input.ts
```diff
@@ -153,21 +153,84 @@ interface MouseOptions {
...
+ * await page.mouse.move(0, 0);
+ * await page.mouse.up();
+ * ```
+ *
+ * **Note**: The mouse events trigger synthetic `MouseEvent`s.
+ * This means that it does not fully replicate the functionality of what a normal user
+ * would be able to do with their mouse.
+ *
+ * For example, dragging and selecting text is not possible using `page.mouse`.
+ * Instead, you can use the [`DocumentOrShadowRoot.getSelection()`](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/getSelection) functionality implemented in the platform.
```
jackfranklin: this needs to be a `@link`

jackfranklin (approved):
One small change then this looks great!

AVGP:
> One small change then this looks great!

Done!

PR #6095, opened 2020-06-24
chore(agnostic): ship CJS and ESM builds

jackfranklin:
For our work to enable Puppeteer in other environments (e.g. a browser) we need to ship an ESM build. This commit changes our config to ship to `lib/cjs` and `lib/esm` accordingly. The majority of our code stays the same, with one small fix for the CJS build to ensure that we ship a version that lets you `require('puppeteer')` rather than have to `require('puppeteer').default`. We do this with the `cjs-entry.js` which is what the `main` field in our `package.json` points to.

We also swap to `read-pkg-up` to find the `package.json` file. This is because the folder structure of `lib/` does not match `src/` now we ship to `cjs` and `esm`, so you cannot rely on exact paths. This module works up from the file to find the nearest `package.json` so it will always find Puppeteer's `package.json`.

Note that we *do not* point any users to the ESM build. We happen to ship those files so people who know about them can get at them but it's not expected (nor will we actively support) that people will rely on them. The CommonJS build is considered our main build.

We may make breaking changes to the structure of the ESM build which we will do without requiring new major versions. For example the ESM build currently ships all files that the CJS build does, but given we are working on the ESM build being able to run in the browser this may change over time.

Long term once the Node versions catch up we can ditch CJS and ship exclusively ESM but we are not there yet.

### DIFF
--- a/src/common/Puppeteer.ts
+++ b/src/common/Puppeteer.ts
@@ -23,6 +23,7 @@ import { ProductLauncher } from '../node/Launcher';
 import { BrowserFetcher, BrowserFetcherOptions } from '../node/BrowserFetcher';
 import { puppeteerErrors, PuppeteerErrors } from './Errors';
 import { ConnectionTransport } from './ConnectionTransport';
+import readPkgUp from 'read-pkg-up';
 
 import { devicesMap } from './DeviceDescriptors';
 import { DevicesMap } from './DeviceDescriptors';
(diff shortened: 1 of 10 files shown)


jackfranklin:
@TimvdLippe PTAL also - you have more TS config experience than me so I'm open to different ways of solving this!

TimvdLippe (approved):
TSConfig looks fine. Haven't reviewed the other stuff.

src/common/Puppeteer.ts
```diff
@@ -23,6 +23,7 @@ import { ProductLauncher } from '../node/Launcher';
 import { BrowserFetcher, BrowserFetcherOptions } from '../node/BrowserFetcher';
 import { puppeteerErrors, PuppeteerErrors } from './Errors';
 import { ConnectionTransport } from './ConnectionTransport';
+import readPkgUp from 'read-pkg-up';
```
paullewis, line 26: Does this work in the browser as well as node?

jackfranklin: No. This is on my list to figure out. The old solution wouldn't have worked either.

I think it only needs to use the `package.json` in order to know which browser revision to download which is not needed in the browser; the work here is to try to prise that code out into a node specific function.

mathiasbynens (approved):

PR #6096, opened 2020-06-24
feat(types): improve typing of `.evaluate()`

jackfranklin:
This is the start of the work to take the types from the
`@types/puppeteer` repository and port them into our repo so we can ship
our built-in types out the box.

This change types the `evaluate` function properly. It takes a generic
type which is the type of the function you're passing, and the arguments
and the return that you get back from the `evaluate` call are typed
correctly.


src/common/EvalTypes.ts
```diff
@@ -0,0 +1,37 @@
...
+  | string
+  | boolean
+  | null
+  | JSONArray
+  | JSONObject;
+export type JSONArray = Serializable[];
+export interface JSONObject {
+  [key: string]: Serializable;
+}
+export type SerializableOrJSHandle = Serializable | JSHandle;
```
jackfranklin, line 58: these are taken directly from the DefinitelyTyped repository. Do we need to acknowledge that / credit back?

jackfranklin:
@marvinhagemeister would you mind having a look if you have time, please? There's much more to do with the types but this is the first step. I'm open to any suggestions or changes to the types you'd like to make.

jackfranklin:
(The build will fail until the docs are updated but I'm going to do that just before merging to keep the diff clean in the interim)

marvinhagemeister (approved):
The changes look good to me 👍 

It would be sure nice to not remove credits. Probably hard to do on a file by file basis. Maybe a small section on the README that acknowledges that the types are taken from DT and mentions the authors, is a way to do that globally?

jackfranklin:
> It would be sure nice to not remove credits. Probably hard to do on a file by file bases. Maybe a small section on the README that acknowledges that the types are taken from DT and mentions the authors is a way to do that globally?

I like this idea - very keen to acknowledge all the work that has gone in thus far. @mathiasbynens wdyt?

mathiasbynens:
> > It would be sure nice to not remove credits. Probably hard to do on a file by file bases. Maybe a small section on the README that acknowledges that the types are taken from DT and mentions the authors is a way to do that globally?
> 
> I like this idea - very keen to acknowledge all the work that has gone in thus far. @mathiasbynens wdyt?

👍  Sounds good.

mathiasbynens (approved):

src/common/EvalTypes.ts
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * @public
+ */
+export type JSONArray = Serializable[];
+
+/**
+ * @public
+ */
+export interface JSONObject {
+  [key: string]: Serializable;
```
felixfbecker, line 52: I know these come from DT, but I honestly don't think these are good typings. This attempts to enforce that the passed argument must be deep-serializable, but that is impossible to express in TypeScript: https://github.com/microsoft/TypeScript/issues/1897
What this line here declares is not "accept any object where all values have this type" (that is inherently impossible with structural typing), but "accept only objects with an index signature of this type". This means that if you pass an object typed with an _interface_ it could be perfectly serializable but TypeScript will still fail with "Type is missing index signature", which makes these types less ergonomic too use for little benefit.

src/common/FrameManager.ts
```diff
@@ -454,16 +455,16 @@ export class Frame {
 
   async $eval<ReturnType extends any>(
     selector: string,
-    pageFunction: Function | string,
-    ...args: unknown[]
+    pageFunction: EvaluateFn | string,
+    ...args: SerializableOrJSHandle[]
```
felixfbecker, line 459: What would be really helpful here (more than trying to constrain the serializable types) would be to make the arguments and return type type-safe, like this:
```ts
$eval<A extends any[], R>(pageFunction: (...args: A) => R, ...args: A): Promise<R>
```

PR #6097, opened 2020-06-24
Remove hard coded `?hl=en` from docs

petele:
The hard coded `?hl=en` is breaking deployments on WebFu


paulirish (approved):

paulirish (approved):

paulirish:
yup. it's best practice that `?hl` shouldn't be in links.

PR #6102, opened 2020-06-25
docs(new): migrate JSHandle docs to TSDoc

mathiasbynens:


src/common/JSHandle.ts
```diff
@@ -80,6 +77,18 @@ export function createJSHandle(
 }
 
 /**
+ * Represents an in-page JavaScript object. JSHandles can be created with the {@link Page.evaluateHandle | page.evaluateHandle} method.
+ *
+ * @example
+ * ```js
+ * const windowHandle = await page.evaluateHandle(() => window);
+ * // …
```
jackfranklin: I think we could lose this `// ...` line.

jackfranklin (approved):

jackfranklin:
@mathiasbynens looks like the docs are not up to date, you may need to `npm install` to get the latest version of the api-* dependencies that we use, then generate the docs again.

PR #6104, opened 2020-06-25
docs(new): migrate Page.ts to TSDoc (part 0 / 2)

hanselfmu:
I'm breaking down the entire migration into three parts. This is the first part.


src/common/Page.ts
```diff
@@ -396,11 +453,18 @@ export class Page extends EventEmitter {
...
 
-  async setGeolocation(options: {
-    longitude: number;
-    latitude: number;
-    accuracy?: number;
-  }): Promise<void> {
+  /**
+   * Sets the page's geolocation.
+   *
+   * * @remarks Consider using {@link BrowserContext.overridePermissions} to
```
jackfranklin: You've got an extra `*` here before `@remarks`

jackfranklin (approved):
Tiny nitpick, otherwise this looks great! Thank you :)

src/common/Page.ts
hanselfmu: Thanks for noticing! Removed it.

PR #6106, opened 2020-06-25
docs(new): Add TSDoc to Coverage class

hashseed:


src/common/Coverage.ts
```diff
@@ -36,6 +91,17 @@ export class Coverage {
     this._cssCoverage = new CSSCoverage(client);
   }
 
+  /**
+   * @param options - Set of configurable options for coverage.
+   * @param options.resetOnNavigation - Whether to reset coverage on every navigation. @defaultValue true.
```
hashseed: This doesn't seem to be the right syntax. How do we document a bag of options?

src/common/Coverage.ts
```diff
@@ -21,12 +21,67 @@ import { CDPSession } from './Connection';
 
 import { EVALUATION_SCRIPT_URL } from './ExecutionContext';
 
+/**
+ * The CoverageEntry class represents one entry of the coverage report.
+- `url` <[string]> StyleSheet URL
+- `text` <[string]> StyleSheet content
+- `ranges` <[Array]<[Object]>> StyleSheet ranges that were used. Ranges are sorted and non-overlapping.
+  - `start` <[number]> A start offset in text, inclusive
+  - `end` <[number]> An end offset in text, exclusive
```
jackfranklin: there's no need to document these here - the comments you've added below will do the trick.

What you do need is to mark this as `@public` and also `export` the interface too.

src/common/Coverage.ts
```diff
@@ -21,12 +21,67 @@ import { CDPSession } from './Connection';
...
+   * The covered range as start and end positions.
+   */
   ranges: Array<{ start: number; end: number }>;
 }
 
+/**
+ * The Coverage class provides methods to gathers information about parts of
+ * JavaScript and CSS that were used by the page.
+ * 
+ * @example
```
jackfranklin, line 77: I think the `@remark` should go before the example so the summary (the stuff pre-remark) is nice and concise.

src/common/Coverage.ts
jackfranklin: The best fix is to extract an interface for these options so you can document them properly. Else TSDoc doesn't handle it very well :(

hashseed: Done.

src/common/Coverage.ts
hashseed: Done.

src/common/Coverage.ts
hashseed: Done.

hashseed:
Please re-review :)

jackfranklin:
@hashseed LGTM! I think after you merged in `main` you need to re-generate the docs and then CI will be happy.

jackfranklin (approved):

PR #6109, opened 2020-06-25
chore: Don't store revisions in `package.json`

jackfranklin:
It's quite messy to have to require the `package.json` file in multiple
places purely to find out what revision of a given browser we want to
use. We can also achieve better type safety by placing it in an actual
source file.

This commit makes that change and also tidies up our reliance on
`package.json` within the source code generally; we now only use it to
find the location of the Puppeteer root such that we know where to
install downloaded browsers to.

To avoid using `package.json` to parse the name of the module, we also
now explicitly have an entry point for the Puppeteer module and the
Puppeter Core module. This will make it easier in the future to ship
less code as part of core (e.g. core never needs to download a browser,
so why ship that code?).

The test install script has also been updated to ensure that
puppeteer-core can be installed correctly too.


OrKoN (approved):

mathiasbynens:
LGTM! Thanks.

The next time we roll Chromium, we should update the example PR in `CONTRIBUTING.md`, since it'll now affect a different file.

> (e.g. core never needs to download a browser, so why ship that code?).

core also doesn't need the info in `revision.ts` :P

mathiasbynens (approved):

PR #6129, opened 2020-06-30
chore: fix Firefox install checker

jackfranklin:
I think when the FF version changes this check breaks - let's just make
it check that a version installed rather than the specific version else
this will happen on a continuing basis.


mathiasbynens (approved):

mathiasbynens:
cc @mjzffr

jackfranklin:
@mjzffr LMK if there's a better way to do this - I'm going to ship this now to get the build green again.

mjzffr:
@jackfranklin How about `ls $TMPDIR/node_modules/puppeteer/.local-firefox/linux-*/firefox/firefox`?

jackfranklin:
@mjzffr sounds good!

Issue #6124, opened 2020-06-30
Tracking doc: shipping built-in type definitions


jackfranklin:
Now the codebase is in TypeScript we can ship type definitions built-in to the library and deprecate the 3rd party DefinitelyTyped version. To be clear, we are very grateful to those who have worked hard on providing types, but if we can ship them built-in that's a better experience for all.

This proposal is also related to [#6118](https://github.com/puppeteer/puppeteer/issues/6118) as shipping new docs will make it easier to use our TS defs as the docs will document them all.

The steps to doing this are roughly like so:

- [ ] Port over the most valuable type defs from DefinitelyTyped. The main functions that we have inferior types for now are `$eval`, `$$eval` and similar.
- [x] Use API Extractor to rollup all our types into one large `puppeteer.d.ts` file that we can ship as the typings. This will enable people to import types directly from `puppeteer` rather than `import {TypeX} from 'puppeteer/lib/...'` which we do not want to support (we do not consider internal file structure changes as breaking and would like to keep it that way).
- [x] Update our `package.json` to point the `types` property to our `puppeteer.d.ts` file (as documented in the [TS handbook](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)).
- [ ] Ship a major version release with the types in.
- [ ] Gather feedback (I expect there will be bugs in our type-defs, or issues to resolve) and iterate in some minor releases with type improvements.

_If you're reading and would like to get involved, that's great and very welcome, but please ping me first to ensure we're not wasting effort by doing the same work!_

christian-bromann:
@jackfranklin what would be the best way to use built in types right now already? While I can import the `d.ts` typings directly I struggle with that in certain cases, e.g. when using `evaluate`. To reproduce:

- `mkdir tmp && cd tmp && npm i typescript puppeteer-core @types/puppeteer-core`
- create this file:
    ```ts
    import puppeteer from 'puppeteer-core'
    import PuppeteerNode from 'puppeteer-core/lib/cjs/puppeteer/node'

    const p: typeof PuppeteerNode = puppeteer as unknown as typeof PuppeteerNode;

    (async () => {
      const browser = await p.launch();
      const page = await browser.newPage();
      await page.goto('https://example.com');

      // Get the "viewport" of the page, as reported by the page.
      const someNumber = await page.evaluate<(a: number, b: number) => number>(
        (a: number, b: number): number => a * b,
      1, 2)

      console.log('Dimensions:', someNumber);

[...]

PR #6130, opened 2020-06-30
feat(types): add (and fix) `evaluateHandle` types

jackfranklin:
This change started as a small change to pull types from DefinitelyTyped over to
Puppeteer for the `evaluateHandle` function but instead ended up also fixing
what looks to be a long standing issue with our existing documentation.
`evaluateHandle` can in fact return an `ElementHandle` rather than a `JSElement`
if you return something that is an HTML element:

```
const button = page.evaluateHandle(() => document.querySelector('button'));
// this is an ElementHandle, not a JSHandle
```

Therefore I've updated the original docs and added a large explanation to the
TSDoc for `page.evaluateHandle`.

In TypeScript land we'll assume the function will return a `JSHandle` but you
can tell TS otherwise via the generic argument, which can only be `JSHandle`
(the default) or `ElementHandle`:

```
const button = page.evaluateHandle<ElementHandle>(() => document.querySelector('button'));
```

This is part of #6124


jackfranklin:
@hanselfmu heads up that this touches `Page.ts` - just in case you have any ongoing TSDoc work in that file.

test/elementhandle.spec.ts
```diff
@@ -182,17 +182,11 @@ describe('ElementHandle specs', function () {
...
         // @ts-expect-error button is expected to be in the page's scope.
         () => button
       );
-      // TODO (@jackfranklin): TS types are off here. evaluateHandle returns a
-      // JSHandle but that doesn't have a click() method. In this case it seems
-      // to return an ElementHandle. I'm not sure if the tests are wrong here
-      // and should use evaluate<ElementHandle> or if the type of evaluateHandle
-      // should change to enable the user to tell us they are expecting an
-      // ElementHandle rather than the default JSHandle.
-      await (buttonHandle as ElementHandle).click();
```
jackfranklin, line 195: it's great that this PR enables this big TODO to go!

mathiasbynens:
Thanks for doing this 👍 The docs you put together are much better!

Let’s be precise when describing this change: `ElementHandle` extends `JSHandle`, so it’s not “not a `JSHandle`”.

docs/api.md
```diff
@@ -2298,12 +2306,14 @@ Shortcut for [(await worker.executionContext()).evaluate(pageFunction, ...args)]
...
 - `pageFunction` <[function]|[string]> Function to be evaluated in the page context
 - `...args` <...[Serializable]|[JSHandle]> Arguments to pass to `pageFunction`
-- returns: <[Promise]<[JSHandle]>> Promise which resolves to the return value of `pageFunction` as in-page object (JSHandle)
+- returns: <[Promise]<[JSHandle]|[ElementHandle]>> Promise which resolves to the return value of `pageFunction` as an in-page object.
 
 The only difference between `worker.evaluate` and `worker.evaluateHandle` is that `worker.evaluateHandle` returns in-page object (JSHandle).
 
 If the function passed to the `worker.evaluateHandle` returns a [Promise], then `worker.evaluateHandle` would wait for the promise to resolve and return its value.
 
+If the function returns an element, the returned handle will be of type [ElementHandle].
```
mathiasbynens: Stoyan once taught me this trick, sharing in case you like it too: https://twitter.com/mathias/status/839160485324947456

If so, `will be` → `is` throughout the doc

jackfranklin: Nice, updated.

jackfranklin:
I've updated the commit message to clarify that ElementHandle extends JSHandle. So the original docs were technically correct, if not a little misleading / lacking clarity.

mathiasbynens (approved):

PR #6138, opened 2020-07-02
chore(docs): reduce warnings when generating docs

jackfranklin:
This is a bunch of small miscellaneous fixes that reduce the amount of
warnings logged when generating our new docs. The long term goal is to
get this list down to 0 warnings, but I'll do it in multiple PRs.


api-extractor.json
```diff
@@ -27,6 +27,9 @@
     },
 
     "extractorMessageReporting": {
+      "ae-internal-missing-underscore": {
+        "logLevel": "none"
```
jackfranklin, line 31: this has been disabled because the docs tool wants anything marked as `@internal` to be prefixed with an underscore. We don't follow this convention (nor do I really want to) so I've disabled this.

src/common/Input.ts
```diff
@@ -33,7 +33,7 @@ type KeyDescription = Required<
  * {@link Keyboard.up}, and {@link Keyboard.sendCharacter}
  * to manually fire events as if they were generated from a real keyboard.
  *
- * On MacOS, keyboard shortcuts like `⌘ A` -> Select All do not work.
+ * On MacOS, keyboard shortcuts like `⌘ A` -\> Select All do not work.
```
jackfranklin, line 36: we need to escape HTML chars in TSDoc comments

src/common/Input.ts
```diff
@@ -275,11 +275,16 @@ export class Keyboard {
   }
 }
 
-type MouseButton = 'none' | 'left' | 'right' | 'middle';
-export type MouseButtonInput = Exclude<MouseButton, 'none'>;
+/**
+ * @public
+ */
+export type MouseButton = 'left' | 'right' | 'middle';
```
jackfranklin, line 281: just simplified this type to make it easier to use and document

OrKoN (approved):

PR #6139, opened 2020-07-02
feat(types): add types for `page.$$eval`

jackfranklin:


jackfranklin:
@hanselfmu PTAL, thanks :)

src/common/Page.ts
```diff
@@ -795,31 +794,70 @@ export class Page extends EventEmitter {
...
+   * // or if you want to be more explicit, provide it as the generic type.
+   * const allInputValues = await page.$$eval<string[]>(
+   *  'input', (elements: HTMLInputElement[]) => elements.map(e => e.textContent)
+   * );
+   * ```
+   *
+   * @param selector the
    * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors | selector}
    * to query for
-   * @param pageFunction - the function to be evaluated in the page context.
```
hanselfmu, line 811: Will API Extractor report a warning here if we remove the `-` after `pageFunction`?

hanselfmu (approved):
LGTM % the `@param -` thing, but it shouldn't block this from landing.

src/common/Page.ts
jackfranklin: yes, it needs there to be a `-` between the param name and the description

Issue #6118, opened 2020-06-29
Tracking doc: migrating to TSDoc


jackfranklin:
We have recently had a week long push where many team members contributed documentation to the codebase in the form of [TSDoc] comments that are used to generate the documentation in the [new-docs] folder.

This issue is to track the remainder of the work done to migrate our documentation fully to being powered by TSDoc.

- [ ] Fix any API Extractor warnings - when running `npm run generate-docs` API extractor spits quite a few warnings out. We should fix these warnings (or silence any we are OK with)
- [ ] Ensure all files are documented fully. Check none were missed. For files that we deliberately don't document (e.g. because they are internal only), we should mark them as such (with an `@internal` so we don't accidentally generate docs for them)
- [ ] Decide how to document and track events that various classes emit.
- [ ] Figure out what to do with pptr.dev. Currently it parses `docs/api.md` to generate the interface for browsing the documentation. We have some options here:
  -  we could provide a custom generator to the API Documenter to generate HTML
  -  we could generate HTML from the markdown and ship that, and figure out how to layer on the search functionality.
  -  we ditch pptr.dev and point to GitHub only. Either permanently, or at least in the interim.
- [ ] Remove `docs/api.md` and all the tooling (`utils/doclint`) around it

# Update: 08-07-20

Chatted with @mathiasbynens and we have come up with a plan:

- [ ] Go through the TSDoc output and tidy up / fix any obvious issues. Make sure all docs from `api.md` are ported.
- [ ] Update `docs/api.md` to point users to `new-docs/puppeteer.puppeteer.md`.
- [ ] Update `docs/api.md` to contain links to old documentation on GitHub, so users on version <5 can still find the docs they need.
- [ ] For now, update pptr.dev to redirect users to GitHub to view the new docs.
- [ ] Open an issue asking users for feedback on the docs, and iterate as needed (I fully expect to find issues; we've moved a lot of documentation around)
- [ ] Spike into generating pptr.dev as a static website.

[TSDoc]: https://github.com/microsoft/tsdoc
[new-docs]: https://github.com/puppeteer/puppeteer/tree/main/new-docs

jasonswearingen:
Hi @jackfranklin is there a page/readme somewhere explaining the new doc workflow?  specifically, I'm currious as to the workflow for turning the tsdocs into a user facing doc site.  (github readmes?)

if this is already implemented somewhere (such as when running ```npm run generate-docs```)  just elt me know and I'll walk myself through that.

thanks

[...]

PR #6141, opened 2020-07-02
feat: add wheel method to Mouse class

### LINKED ISSUE #1679: [Feature request] Mouse wheel API calls
Sounds a bit old in our age of touch interfaces, but would be great to have native mouse wheel simulation (to use with existing code).

christian-bromann:
fixes #1679

This patch adds a wheel command to the [`Mouse`](https://pptr.dev/#?product=Puppeteer&version=v4.0.1&show=api-class-mouse) class. It behaves similar to other mouse commands like `up` or `down` and therefor requires a `mouse.move(x, y)` for targeting the element properly.

An example script to test this feature:

```js
const puppeteer = require('./');

function sleep (ms = 3000) {
  return new Promise((resolve) => setTimeout(resolve, ms))
}

(async () => {
    headless: false,
  });

  page.on('request', async (request) => {
    console.log(await request.headers());
  });

  await page.goto('https://mdn.mozillademos.org/en-US/docs/Web/API/Element/wheel_event$samples/Scaling_an_element_via_the_wheel?revision=1587366');
  await page.mouse.wheel({ y: 100 });
  await sleep(1000);

  const elem = await page.$('div');
  const boundingBox = await elem.boundingBox();
  console.log(boundingBox); // returns: { x: 342.5, y: 242.5, width: 115, height: 115 }

  await page.mouse.move(
    boundingBox.x + boundingBox.width / 2,
    boundingBox.y + boundingBox.height / 2
  );

  await page.mouse.wheel({ deltaY: -100 });
  console.log(await elem.boundingBox()); // returns: { x: 285, y: 185, width: 230, height: 230 }

  await sleep();
  await browser.close();
})();
```


test/assets/input/wheel.html
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <style type="text/css">
```
mathiasbynens:
```suggestion
    <style>
```

mathiasbynens: https://mathiasbynens.be/notes/html5-levels#level-1

src/common/Input.ts
```diff
@@ -440,6 +445,38 @@ export class Mouse {
...
+   * await page.goto('https://mdn.mozillademos.org/en-US/docs/Web/API/Element/wheel_event$samples/Scaling_an_element_via_the_wheel?revision=1587366');
+   *
+   * const elem = await page.$('div');
+   * const boundingBox = await elem.boundingBox();
+   * await page.mouse.move(
+   *   boundingBox.x + boundingBox.width / 2,
+   *   boundingBox.y + boundingBox.height / 2
+   * );
+   *
+   * await page.mouse.wheel({ y: -100 })
```
mathiasbynens: `deltaY`?

mathiasbynens (changes requested):
Thanks for taking this on, Christian! Looking good so far.

Please document this in `docs/api.md` as well.

src/common/Input.ts
```diff
@@ -283,6 +283,11 @@ interface MouseOptions {
   clickCount?: number;
 }
 
+interface MouseWheelOptions {
```
jackfranklin: you'll need an `export` on this (so that when we generate the `puppeteer.d.ts` file it gets included)

christian-bromann: I assume we want to export `MouseOptions` then too.

christian-bromann:
@mathiasbynens @jackfranklin thanks for the comments. I addressed them and squashed all commits into one. Please let me know if anything is missing.

jackfranklin (approved):
Thanks for this! Sorry you have to write the docs in two places; hopefully we can finish the migration in the not-too-distant future.

christian-bromann:
> hopefully we can finish the migration in the not-too-distant future.

Sounds good! Let me know if there is anything I can do to help.

christian-bromann:
@jackfranklin I need some help understanding why the docs test are failing:

```
DocLint Failures:
  1) [MarkDown] Method Mouse.wheel() options Object != MouseWheelOptions
  2) [MarkDown] Method Mouse.wheel() options has unexpected property 
```

They seem to scan the `api.md` file and complain about my change there:

```diff
+ #### mouse.wheel([options])
+ - `options` <[Object]>
+   - `deltaX` X delta in CSS pixels for mouse wheel event (default: 0).
+   - `deltaY` Y delta in CSS pixels for mouse wheel event (default: 0).
+ - returns: <[Promise]>
+ 
+ Dispatches a `mousewheel` event.
```

However I don't see any other TS interface options being used anywhere else in that document, e.g. for `mouse.up()`:

```md
#### mouse.up([options])
- `options` <[Object]>
  - `button` <"left"|"right"|"middle"> Defaults to `left`.
  - `clickCount` <[number]> defaults to 1. See [UIEvent.detail].
- returns: <[Promise]>

Dispatches a `mouseup` event.
```

Any advice?

jackfranklin:
@christian-bromann ah, this is because the old system that's used to validate the docs doesn't understand the TypeScript source code as well as it should. Given we are migrating away from it I've not invested time into fixing it.

So what we do is we basically tell the old doc tool to ignore it, because it doesn't understand. You can see examples of this here: https://github.com/puppeteer/puppeteer/blob/main/utils/doclint/check_public_api/index.js#L746

This is not good, and another reason we're keen to get #6118 done!

OrKoN (approved):

docs/api.md
```diff
@@ -2579,6 +2580,14 @@ Dispatches a `mousemove` event.
 
 Dispatches a `mouseup` event.
 
+#### mouse.wheel([options])
+- `options` <[Object]>
+  - `deltaX` X delta in CSS pixels for mouse wheel event (default: 0).
+  - `deltaY` Y delta in CSS pixels for mouse wheel event (default: 0).
```
mathiasbynens: Please include an explanation of what positive and negative means in this context.

docs/api.md
```diff
@@ -2579,6 +2580,14 @@ Dispatches a `mousemove` event.
...
+- returns: <[Promise]>
+
+Dispatches a `mousewheel` event.
```
mathiasbynens, line 2589: Let's include the code example here as well

hanselfmu (approved):

mathiasbynens (approved):

PR #6145, opened 2020-07-02
chore(docs): fix more documentation warnings

jackfranklin:


mathiasbynens (approved):

src/common/FrameManager.ts
```diff
@@ -234,7 +237,7 @@ export class FrameManager extends EventEmitter {
     if (this._frames.has(frameId)) return;
     assert(parentFrameId);
     const parentFrame = this._frames.get(parentFrameId);
-    const frame = new Frame(this, this._client, parentFrame, frameId);
```
jackfranklin, line 237: turns out that we instantiate the `Frame` class with a CDP session, but we never use it.

mathiasbynens: 🤯

PR #6146, opened 2020-07-02
chore(docs): document HTTPRequest with TSDoc

jackfranklin:


src/common/HTTPRequest.ts
```diff
@@ -66,44 +143,120 @@ export class HTTPRequest {
...
+   * @returns one of the following: `document`, `stylesheet`, `image`, `media`,
+   * `font`, `script`, `texttrack`, `xhr`, `fetch`, `eventsource`, `websocket`,
+   * `manifest`, `other`.
+   */
   resourceType(): string {
+    // TODO (@jackfranklin): protocol.d.ts has a type for this, but all the
+    // string values are uppercase. The Puppeteer docs explicitly say the
+    // potential values are all lower case, and the constructor takes the event
+    // type and calls toLowerCase() on it, so we can't reuse the type from the
+    // protocol.d.ts. Why do we lower case?
```
jackfranklin, line 166: @mathiasbynens I don't suppose you have any insight here?

src/common/HTTPRequest.ts
```diff
@@ -66,44 +143,120 @@ export class HTTPRequest {
...
     return this._redirectChain.slice();
   }
 
   /**
-   * @returns {?{errorText: string}}
+   * Access information about the request's failure.
+   *
+   * @remarks
+   *
+   * Example of logging all failed requests:
```
hanselfmu, line 250: Do we want to separate these in a `@example` tag?

hanselfmu (approved):
LGTM % one small nit, and it can go either way. Thanks Jack!

src/common/HTTPRequest.ts
jackfranklin: Yes, I think we should! Good catch, thanks :)

src/common/HTTPRequest.ts
mathiasbynens: I suppose for parity with other Puppeteer APIs? We don't really use uppercase string values anywhere.

PR #6169, opened 2020-07-07
chore: rename `createJSHandle` to `createHandle`

jackfranklin:
The name was misleading as the function will create an `ElementHandle`
should the remote object its given be a DOM Node.

For clarity, this is a function that's only used internally, and this is
therefore not a breaking change. It's not part of the public API.

### DIFF
--- a/src/common/ExecutionContext.ts
+++ b/src/common/ExecutionContext.ts
@@ -16,7 +16,7 @@
 
 import { assert } from './assert';
 import { helper } from './helper';
-import { createJSHandle, JSHandle, ElementHandle } from './JSHandle';
+import { createHandle, JSHandle, ElementHandle } from './JSHandle';
 import { CDPSession } from './Connection';
 import { DOMWorld } from './DOMWorld';
 import { Frame } from './FrameManager';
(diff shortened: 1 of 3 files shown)


mathiasbynens:
Hmm... It's not really misleading IMHO since `ElementHandle` extends `JSHandle`. In fact, in a way `createHandle` is more confusing to me, since we have `objectHandle`s in the codebase too.

jackfranklin:
@mathiasbynens IMO this is a better name because it aligns with the `evaluateHandle` method. Whilst we do have `objectHandles` they are largely internal and not really exposed to end users.

I think this is a better name (motivated by losing time to not realising that it can return an `ElementHandle`) but if you strongly disagree let's leave it; I don't think it's worth much time debating on.

@OrKoN @hanselfmu what do you think?

hanselfmu:
I'd be inclined to say to align `createJSHandle` with `evaluateHandle`, how about go the other way around and change `evaluateHandle` to `evaluateJSHandle`? I know this doesn't solve the problem that by the function name it's not clear that it can return `ElementHandle`, but I'd agree with Mathias that since `ElementHandle` extends from `JSHandle`, it's sort of assumed.

jackfranklin:
Hmmm, the only thing with `evaluateJSHandle` is that's a breaking change that I'm not sure is worth it.

Let's leave this for now - if we're not all aligned this doesn't feel worth debating :)

PR #6172, opened 2020-07-07
chore: use devtools-protocol package

jackfranklin:
Rather than maintain our own protocol we can instead use the devtools-protocol package and pin it to the version of Chromium that Puppeteer is shipping with.

The only changes are naming changes between the bespoke protocol that Puppeteer created and the devtools-protocol one.


jackfranklin:
This isn't ready for merging yet; I'd like to understand how the revision numbers of devtools-protocol work - there isn't an exact version for the current revision of Chromium that we support, for example.

utils/doclint/check_public_api/index.js
```diff
@@ -807,6 +814,16 @@ function compareDocumentations(actual, expected) {
...
     if (namingMismatchIsExpected) return;
+
+    /* Some methods cause errors in the property checks for an unknown reason
+     * so we support a list of methods whose parameters are not checked.
+     */
+    const skipPropertyChecksOnMethods = new Set([
+      'Method Page.deleteCookie() ...cookies',
+      'Method Page.setCookie() ...cookies',
+    ]);
+    if (skipPropertyChecksOnMethods.has(source)) return;
```
jackfranklin, line 851: I have no idea why this change suddenly causes this; I've manually checked that the new types are identical to the old types in all but name.

At this point the doclint tooling is a mystery to me and I'd rather hack around it as we're quite close to removing it.

package.json
```diff
@@ -47,6 +45,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "debug": "^4.1.0",
+    "devtools-protocol": "0.0.758979",
```
jackfranklin: need to confirm which revision number we're using

scripts/ensure-correct-devtools-protocol-package.ts
```diff
@@ -0,0 +1,83 @@
...
+    Current devtools-protocol version in package.json: ${currentProtocolPackageInstalledVersion}
+    Expected devtools-protocol version:                ${bestRevisionFromNpm}`);
+
+  process.exit(1);
+}
+
+console.log(
+  `Correct devtools-protocol version found (${bestRevisionFromNpm}).`
+);
+process.exit(0);
```
jackfranklin, line 83: @mathiasbynens particularly interested in getting some more eyes on this

scripts/ensure-correct-devtools-protocol-package.ts
```diff
@@ -0,0 +1,83 @@
...
+
+import packageJson from '../package.json';
+
+const currentProtocolPackageInstalledVersion =
+  packageJson.dependencies['devtools-protocol'];
+
+/**
+ * Ensure that the devtools-protocol version is pinned.
+ */
+if (!currentProtocolPackageInstalledVersion[0].match(/[0-9]/)) {
```
mathiasbynens: Nit: we don't need the match result, so let's use `RegExp#test`

```suggestion
if (/^[^0-9]/.test(currentProtocolPackageInstalledVersion) {
```

(`[^0-9]` could be `\D` but I actually like spelling it out more explicitly)

scripts/ensure-correct-devtools-protocol-package.ts
```diff
@@ -0,0 +1,83 @@
...
+const output = execSync(command, {
+  encoding: 'utf8',
+});
+
+const bestRevisionFromNpm = output.split(' ')[1].replace(/'|\n/g, '');
+
+if (currentProtocolPackageInstalledVersion !== bestRevisionFromNpm) {
+  console.log(`ERROR: bad devtools-protocol revision detected:
+
+    Current Pptr Chromium revision: ${PUPPETEER_REVISIONS.chromium}
```
mathiasbynens:
```suggestion
    Current Puppeteer Chromium revision: ${PUPPETEER_REVISIONS.chromium}
```

utils/doclint/check_public_api/index.js
mathiasbynens: ack, sg

scripts/ensure-correct-devtools-protocol-package.ts
jackfranklin: ah nice

mathiasbynens (approved):

PR #6184, opened 2020-07-08
chore: remove coverage dependency on Events.ts

jackfranklin:
Another step towards removing Events.ts. This commit updates the
coverage utils to instead parse each module and check for the presence
of an `XEmittedEvents` object which indicates that module has events
that we need to check coverage on.

I also refactored how we store the list of what we consider public API
to make it easier to use in coverage and then in the
`installAsyncStackHelper` that gets run in `initialize.ts`.


src/public-api.ts
```diff
@@ -0,0 +1,66 @@
...
+]);
+
+export const classesForAsyncHelpersInstall = new Set();
+
+for (const [moduleName, modulePath] of publicPuppeteerClasses.entries()) {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const module = require(modulePath);
+  const classForAsyncHelpersInstall = module[moduleName];
+  classesForAsyncHelpersInstall.add(classForAsyncHelpersInstall);
+}
```
jackfranklin, line 66: @OrKoN we're actually going to remove the async error helpers (see team chat - it predates https://v8.dev/blog/fast-async#improved-developer-experience and Node/browsers handle it better now) so don't pay too much attention to this. Once this PR lands another will immediately remove this :)

jackfranklin:
Realised I can make this much tidier if we remove the async error stack helper first, and then do this work. Will close this.

PR #6187, opened 2020-07-09
chore: document browser install breakage on Node 14.0.0

jackfranklin:


docs/troubleshooting.md
```diff
@@ -130,6 +131,18 @@ yum update nss -y
...
+## Chrome is downloaded but fails to launch on Node 14
+
+If you get an error that looks like this when trying to launch Chromium:
+
+```
+(node:15505) UnhandledPromiseRejectionWarning: Error: Failed to launch the browser process!
+spawn /Users/.../node_modules/puppeteer/.local-chromium/mac-756035/chrome-mac/Chromium.app/Contents/MacOS/Chromium ENOENT
+```
+
+This means that the browser was downloaded but failed to be extracted correctly. The most common cause is a bug in Node 14.0.0 which broke `extract-zip`, the module Puppeteer uses to extract browser downloads into the right place. The bug was fixed in Node 14.1.0, so please make sure you're ruunning that version or higher. Alternatively, if you cannot upgrade, you could downgrade to Node 12, but we recommend upgrading when possible.
```
mathiasbynens:
```suggestion
This means that the browser was downloaded but failed to be extracted correctly. The most common cause is a bug in Node.js v14.0.0 which broke `extract-zip`, the module Puppeteer uses to extract browser downloads into the right place. The bug was fixed in Node.js v14.1.0, so please make sure you're running that version or higher. Alternatively, if you cannot upgrade, you could downgrade to Node.js v12, but we recommend upgrading when possible.
```

docs/troubleshooting.md
```diff
@@ -130,6 +131,18 @@ yum update nss -y
 - [#379](https://github.com/puppeteer/puppeteer/issues/379) - Alpine troubleshooting <br/>
 </details>
 
+## Chrome is downloaded but fails to launch on Node 14
```
mathiasbynens:
```suggestion
## Chrome is downloaded but fails to launch on Node.js v14
```

mathiasbynens (approved):

jackfranklin:
@jschfflr PTAL :)

jschfflr:
Thanks for adding this!

Issue #6125, opened 2020-06-30
Tracking doc: creating environment-agnostic Puppeteer


jackfranklin:
In #5750 we removed the puppeteer-web bundle as it wasn't properly supported and we didn't want to continue shipping it. However we think it's important to ship a web bundle that could be used in a browser and this issue will track work in that area. There is no current timescale set for this work.

The current focus for this work is to go through Puppeteer's core and pull out the parts that explicitly depend on Node. Most of these are for features we wouldn't support in a web based Puppeteer (e.g. downloading a browser, or launching one!) so we need to work to split the codebase up into those parts. Currently in `src` we have `common` and `node` but there is code in `common` that has a dependency on Node which should not be the case.

Running `npm run generate-dependency-graph` will generate a diagram of dependencies and it's this that we need to work on. Once that is done the work then becomes:

1. Ship an ESM build of Puppeteer. As of #6095 this is landed, but we don't point anyone to it. The web build should ship as ESM, not CJS. Once that version is released we can dog-food it. (Note to anyone reading: we are not supporting the ESM version as a first class citizen yet, so please do not start to use it. It may break at any time). Long term once we only support Nodes that support ESM we can ship ESM and remove CJS.
2. Create web based tests for the Puppeteer web version that mimic the Node tests we have now.
3. Decide on how to publish the web version - under `puppeteer` but as `import X from 'puppeteer/web'` ? Or a separate npm package? Or just as `puppeteer` and let bundlers decide which to use?
4. What does this mean for `puppeteer-core` ? Is that useful in its own right?
5. Create separate documentation for the web based version.

Mohammed-Fayaz:
@jackfranklin is this task complete? Not clear if this effort to create environment agnostic puppeteer has been abandoned or is complete from this issue

Issue #2136, opened 2018-03-03
puppeteer format coverage is missing some information


bcoe:
Some of the folks working on Istanbul recently released the tool [puppeteer-to-istanbul](https://github.com/istanbuljs/puppeteer-to-istanbul), which converts from puppeteer coverage format to Istanbul format, allowing Istanbul reports to be run.

The tool works, but there does seem to be some information missing in puppeteer style output (this is discussed in #1768). Specifically:

* callCount is not reported  (it seems like there should probably be a flag to enable callCount).
* I noticed block level coverage is working for `if` statements, but not binary statements, e.g., `const foo = 99 || 103`.

**What steps will reproduce the problem?**

1. collect coverage for a file that executes a line multiple times (see: https://github.com/bcoe/puppeteer-to-istanbul-example, for setting up the test case). 
2. collect coverage for a file that has a binary operation, e.g., `const foo = 99 || 100`.

**What is the expected result?**

* call counts should be tracked.
* partial block level coverage should be collected (`[const foo = 99 ]|| 100`).

**What happens instead?**

* No callCount information is reported.
* the entire line `const foo = 99 || 100` is missing coverage.

----

It would be awesome to get all the fancy pants work that @schuay has done around block-level coverage outputting in reports collected through puppeteer.

CC: @aslushnikov, @shannon, @JoelEinbinder,  @vkoves and @rjaltman

LarsDenBakker:
@jackfranklin would it be possible to expose the raw v8 test coverage data? There are some tools which expect this format. 

We could add a flag to the `stop` function to specify the format, or we could introduce a separate method. What do you think?

TimvdLippe:
For additional context, I found this reference in the code that Lars is referring to: https://github.com/modernweb-dev/web/pull/174/files#diff-8365ecf94906ff774beccd6509eec588R88

LarsDenBakker:
Indeed. I'm using https://github.com/istanbuljs/v8-to-istanbul/ which expects the raw v8 data format.

PR #6200, opened 2020-07-10
chore: add test runner for testing pptr in a browser

LarsDenBakker:
@jackfranklin @TimvdLippe 
As discussed, this is a basic example setup for web test runner. I added an example test using mocha and expect, similar to the node tests.

Reference: https://github.com/puppeteer/puppeteer/issues/6125

The config is an es module (requires latest node 12 or 14) but it can be commonjs as well. 

### Things to be done:

**Make the tests actually work**
~~Ideally we can reuse the existing unit tests, or at least a subset of them~~

**TS or JS tests**
~~The node tests are written in TS and run in mocha with ts-node. I followed a similar approach of using esbuild to compile TS to JS on the fly. We could also run the compiled JS in the browser.~~

**Handle `expect` errors**
~~I created a small UMD wrapper to make `expect` work in the browser. However the error stack traces are handled properly by web test runner (I've only been testing with chai until now). Something to look into on WTR side.~~

**Remove caniuse dependency**
~~caniuse has a license which isn't whitelisted, and it's actually problematic in general. I thought I had gotten rid of it, but it turns out for test coverage we still need to use babel. I want to get rid of this dependency and use v8 test coverage information. I got it working with playwright, but for puppeteer we need something like https://github.com/puppeteer/puppeteer/issues/2136.~~

**puppeteer dependency**
web test runner uses puppeteer-core to control the browser when running tests. so we have puppeteer-ception. I'm not sure that would be a problem here?

The test runner is brand new, so I'm sure we'll find some rough edges.

### DIFF
--- a/test-browser/debug.spec.js
+++ b/test-browser/debug.spec.js
@@ -1,0 +1,65 @@
+/**
+ * Copyright 2020 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { debug } from '../lib/esm/puppeteer/common/Debug.js';
+import expect from '../node_modules/expect/build-es5/index.js';
+
+describe('debug', () => {
+  let originalLog;
+  let logs;
+  beforeEach(() => {
+    originalLog = console.log;
+    logs = [];
+    console.log = (...args) => {
+      logs.push(args);
+    };
+  });
+
+  afterEach(() => {
+    console.log = originalLog;
+  });
+
+  it('should return a function', async () => {
+    expect(debug('foo')).toBeInstanceOf(Function);
+  });
+
+  it('does not log to the console if __PUPPETEER_DEBUG global is not set', async () => {
+    const debugFn = debug('foo');
+    debugFn('lorem', 'ipsum');
+
+    expect(logs.length).toEqual(0);
+  });
+
+  it('logs to the console if __PUPPETEER_DEBUG global is set to *', async () => {
+    globalThis.__PUPPETEER_DEBUG = '*';
+    const debugFn = debug('foo');
+    debugFn('lorem', 'ipsum');
+
+    expect(logs.length).toEqual(1);
+    expect(logs).toEqual([['foo:', 'lorem', 'ipsum']]);
+  });
+
+  it('logs only messages matching the __PUPPETEER_DEBUG prefix', async () => {
+    globalThis.__PUPPETEER_DEBUG = 'foo';
+    const debugFoo = debug('foo');
+    const debugBar = debug('bar');
+    debugFoo('a');
+    debugBar('b');
+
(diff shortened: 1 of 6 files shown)


TimvdLippe (commented):
For my own understanding, this PR serves as an example as to how your `web-test-runner` *uses* Puppeteer to run tests in a browser, right? Or do you want it to be used *as part of* the test suite for Puppeteer itself?

If the former, it probably makes more sense to put this in a folder like `examples/` or something. But I am not sure what the best format is and where best to place it.

LarsDenBakker:
@TimvdLippe the goal here is to use WTR to run test the puppeteer browser distribution.

TimvdLippe:
Ah I see. Could you please reference #6125 in that case?

LarsDenBakker:
Ah I didn't know about that. I added it now.

jackfranklin:
@LarsDenBakker thanks for putting this together! We should see if we can get it running on Travis to see what happens but I think the ESM build isn't truly functioning yet, we need to add file extensions (#6202) and also figure out what to do with Mitt #6203).

I'd also be quite keen not to compile via the test runner, but instead run the tests directly against the output in `lib/esm`. That will let us test the same ESM build we're shipping to users. WDYT?

LarsDenBakker:
@jackfranklin yeah that makes sense, we can run the tests on the .js files. We don't do source maps in test reporting yet, but we'll pick that up very soon.

Regarding mitt, the test runner has `--node-resolve` flag to resolve imports serverside. But that doesn't help after publishing, of course :)

jackfranklin:
Yeah, I'd rather not rely on the `--node-resolve` flag. 

I think a good step here would be to get a test written for the changes to our debug module in #6210 - that's a self contained function that we can run and test in the browser. Probably a good candidate for the very first test as it's tiny and doesn't rely on any big Puppeteer objects.

LarsDenBakker:
@jackfranklin I removed the node resolve and TS compilation, and added a test for the current debug module.

I kept the UMD -> ESM transform for the `expect` library since they don't ship an es module. You could also use it from the window, though I'm not sure how well that works for IDE code completion.

jackfranklin:
@LarsDenBakker thanks for your efforts here, sorry for my slow reply.

Looks like CI has a problem with the new spec file:

> error: Parsing error: ImportDeclaration should appear when the mode is ES6 and in the module context at test-browser/debug.spec.js:

But this is looking good to go. Could you also try adding this as a step to `.travis.yml` so we can check CI is happy and stable?

Once we get this all green I'd be happy to ship this. @mathiasbynens what say you? :)

LarsDenBakker:
No worries. I updated the linting configuration to lint the browser tests in a browser context. I also updated the travis configuration, but it doesn't seem to affect the CI. Does it always pull from master? Or do I need to update something else as well?

LarsDenBakker:
Looks like it's picking up the changes now, not sure why it wasn't before. I updated the tests to match the changes in the debug module on master. All green now 👍

jackfranklin (approved):

test-browser/debug.spec.js
```diff
@@ -0,0 +1,65 @@
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
```
mathiasbynens: 2018?

web-test-runner.config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
```
mathiasbynens, line 17: Could you please add the license header here too? We need it for all files

test-browser/debug.spec.js
LarsDenBakker: Updated

web-test-runner.config.js
LarsDenBakker: Added

mathiasbynens (approved):

mathiasbynens:
Thanks, Lars!

PR #6202, opened 2020-07-10
chore: enforce file extensions on imports

jackfranklin:


.eslintrc.js
```diff
@@ -88,7 +89,8 @@ module.exports = {
 
         "no-restricted-imports": ["error", {
             patterns: ["*Events"],
-        }]
+        }],
+        "import/extensions": ["error", "ignorePackages"]
```
jackfranklin, line 93: we need to figure out how packages work for a browser ESM build - but that's a separate issue to this one. I'll have a PR/issue up describing our potential approach later :)

paullewis (approved):

scripts/ensure-correct-devtools-protocol-package.ts
```diff
@@ -33,10 +33,10 @@
  * find the one closest to our Chromium revision.
  */
 
-import { PUPPETEER_REVISIONS } from '../src/revisions';
+import { PUPPETEER_REVISIONS } from '../src/revisions.js';
 import { execSync } from 'child_process';
 
-import packageJson from '../package.json';
+import packageJson from '../package.json.js';
```
TimvdLippe: I think this needs to remain `package.json`

test/jshandle.spec.ts
```diff
@@ -14,13 +14,13 @@
  * limitations under the License.
  */
 
-import expect from 'expect';
+import expect from 'expect.js';
```
TimvdLippe: This should remain `'expect'` as it uses the Node resolution algorithm.

test/launcher.spec.ts
```diff
@@ -16,18 +16,18 @@
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
-import sinon from 'sinon';
+import sinon from 'sinon.js';
```
TimvdLippe: Same here.

TimvdLippe (changes requested):

Issue #6203, opened 2020-07-10
Vendor common dependencies into the repository


jackfranklin:
We are currently working on shipping a browser build. #6202 fixes one of the issues @paullewis found; we need file extensions on our imports to match the ESM spec.

The next challenge is the import of dependencies that we use in any environments. The only dependency we have currently that's common to all environments is Mitt. Right now our bundled code ships as `import mitt from 'mitt'` but a browser has no idea what `mitt` refers to and how to resolve it.

[importMaps](https://github.com/WICG/import-maps) may provide a solution in the future but they are not at a point to be relied on yet.

I'm proposing that we vendor any common dependencies into this repository. This means we can import them as `import mitt from './common/third_party/mitt/mitt.es.js` (exact path TBC) and that will work in the browser and Node equally well.

The downside is we have to manage these dependencies but I think our list of dependencies will be very small; the vast majority of Pptr's dependencies are for extracting files from zips/DMGs/etc and I think as a rule we'll be very considered about adding new common dependencies in the future. If needed we can introduce some scripts to make managing them and updating them easier (like we have in DevTools-Frontend).

jackfranklin:
@mathiasbynens @paullewis @OrKoN @hanselfmu PTAL :)

paullewis:
Do you expect this to only apply to mitt, or other dependencies?

I don't know of any other way of doing this besides rolling in a `third_party` or `lib` folder, and certainly if the number of deps is low, and it allows the codebase to be fully agnostic I'm personally very supportive.

hanselfmu:
It looks like Mitt does not have any dependencies, and seems like a single build can work in both the browser and Node, so this solution seems quite promising, at least for this interim period.

And I agree that we should be careful about adding other common dependencies, because not all the libraries out there are as self-contained as Mitt 😃

jackfranklin:
> Do you expect this to only apply to mitt, or other dependencies?

It would apply to any common dependencies - that is, ones that are used in both the browser and Node version of Puppeteer.

> It looks like Mitt does not have any dependencies, and seems like a single build can work in both the browser and Node, so this solution seems quite promising, at least for this interim period.

This is a good point, if Mitt had 5 dependencies, we'd have a bigger challenge on our hands. I hope we can avoid this and keep our dependencies down to a minimum. In our case Mitt being tiny really helps.

As @paullewis says I can't think of another way to do this; I'll put together a proof of concept to see how hard/easy/painful/etc it is.

[...]

PR #6209, opened 2020-07-13
chore: vendor Mitt & update project structure

jackfranklin:
As discussed in #6203 we need to vendor our common dependencies in so
that when we ship an ESM build all imports point to file paths and do
not rely on Node resolution (e.g. a browser does not understand `import
mitt from 'mitt'`).


paullewis:
The standard pattern is to put it in a root folder under `third-party` or `lib`, and I think we should do that here rather than embedding it in the file structure. That way it is very clear that it has been vendored in without digging.

jackfranklin:
> The standard pattern is to put it in a root folder under third-party or lib, and I think we should do that here rather than embedding it in the file structure. That way it is very clear that it has been vendored in without digging.

I normally would agree but I want to make it clear that we only vendor things in that are needed in all environments. E.g. we're not vendoring in any node-only dependencies. 

WDYT?

mathiasbynens:
> I normally would agree but I want to make it clear that we only vendor things in that are needed in all environments. E.g. we're not vendoring in any node-only dependencies.

Could we achieve that by adding `vendor/README.md` stating as much?

jackfranklin:
I just realised it will complicate our setup much more if we have code outside of `src` - because we generate that directly into `lib`, but if code is outside of that structure the `lib` directory will change.

Therefore, how do people feel about having `src/vendor`, with a `README` as @mathiasbynens suggested explaining the setup?

We could have vendor at the top level, but it will mean a fair bit of tweaking of the build system that doesn't feel worth the effort, to be honest with you.

@mathiasbynens @paullewis LMK what you think

mathiasbynens:
I'd be alright with `src/vendor` or `src/third_party`, possibly with a symlink to it at the actual root for extra discoverability. Paul, would that address your concerns?

> We could have vendor at the top level, but it will mean a fair bit of tweaking of the build system that doesn't feel worth the effort, to be honest with you.

With `vendor`/`third_party` at the top-level + a symlink perhaps the build system wouldn't need any tweaks?

paullewis (approved):
LGTM

jackfranklin:
@mathiasbynens we ended up chatting it through and going for the top level vendor directory, which I do think is probably cleaner in the long run. We've had to do some tweaking of the configuration but actually it's probably a bit cleaner now, and has some more clear divides between all the various ways we use TS (docs, tests, source code).

mathiasbynens:
Nice! I'm happy when Paul's happy. Thanks for resolving this 👍

mathiasbynens (approved):

mathiasbynens:
(note: please change the commit message accordingly while squash+merging)

Issue #5786, opened 2020-04-30
Add first level logging capability


christian-bromann:
I would like to propose a logging capability that allows Puppeteer user to be informed about events happening in Puppeteer and Chrome DevTools. Currently you have to [overwrite](https://github.com/webdriverio/webdriverio/blob/master/packages/devtools/src/utils.js#L304-L335) the `debug` package to access this. It would be great if this would be more accessible. My goals are:

- ability to receive log messages from Puppeteer for my own logging purposes
- ability to listen to all CDP communication
- __extra__ (not required): ability to recreate Puppeteer code based on command calls on its primitives

## Use Cases

1) As a WebdriverIO maintainer that [uses](https://webdriver.io/blog/2019/09/16/devtools.html) Puppeteer for Chromium based automation I would like to incorporate Puppeteer logs into the logs that are generated by WebdriverIO without having to monkey patch the debug package. This would fail if the user has the same version of the debug package installed (see webdriverio/webdriverio#5338).

2) As a reporting plugin maintainer it would be very beneficial if all CDP communication can be obtained in a single place. With that I can attach that reporting tool during initialisation and doesn't need to work with Puppeteer primitives to register unnecessary event listeners.

3) As a cloud provider that runs Puppeteer on its platform it would be ideal if Puppeteer commands can be be retraced from the test execution. For example if someone calls `wait page.keyboard.type('Hello World!')` a bunch of `Input.dispatchKeyEvent` commands are being send over to Chrome. For user not familiar with the Chrome DevTools protocol it would be much more insightful if Puppeteer commands can be logged instead of CDP events. However this is not a requirement for me as I am not sure if this should be done Puppeteer or some other instrumentation tool.

---

If you all could give me some direction I would be happy to contribute. I was looking into [Playwrights](https://github.com/microsoft/playwright/blob/master/docs/api.md#class-logger) approach and found it very useful. Maybe it could be just ported to Puppeteer? What do you all think?

jackfranklin:
I like the idea of launching and providing some `logger: (name, severity, message) => {}` and letting users do with that what they will. That seems like the best way to enable everyone to do exactly what they want with the logs.

We could also default to what we do right now if this isn't provided, so it's not a breaking change.

@mathiasbynens WDYT?

mathiasbynens:
I'm in the same boat. Definitely open to this, but it would be easier to discuss if there was a concrete API proposal or a patch.

christian-bromann:
not stale

[...]

PR #6210, opened 2020-07-13
fix: configure debug logging in browser

jackfranklin:
The Node debug library uses the `DEBUG` environment variable to
configure what gets logged. Our browser version just logs everything;
this commit changes it to look for `window.__PUPPETEER_DEBUG` and matches the behaviour accordingly:

* If the value is not set, nothing is logged.
* If the value is set to `*` everything is logged.
* If the value is set to a string `foo`, messages with that prefix are
  logged.
* If the value is set to a string ending in `*`, e.g. `foo*`, messages
  with prefixes that start with `foo` are logged.

### DIFF
--- a/src/common/Debug.ts
+++ b/src/common/Debug.ts
@@ -19,12 +19,30 @@ import { isNode } from '../environment.js';
 /**
  * A debug function that can be used in any environment.
  *
+ * @remarks
+ *
  * If used in Node, it falls back to the
  * {@link https://www.npmjs.com/package/debug | debug module}. In the browser it
  * uses `console.log`.
  *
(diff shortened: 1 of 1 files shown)


paullewis:
Do you want any tests for this?

paullewis (approved):

jackfranklin:
@paullewis , yes but we can't yet until we have browser test frameworks set up. Which is on the agnostification list of work to be done.

christian-bromann:
@jackfranklin would you mind giving feedback on #5786 I would love for Puppeteer to provide even more insights about what it is doing.

PR #6213, opened 2020-07-13
chore: generate docs for the protocol

jackfranklin:
The motivation here is that in our code we often say a function takes an argument that is of type `Protocol.X`.

We then have some options for how we document this well:

* Link the user to the devtools protocol viewer
* Duplicate the possible values in the TSDoc comment
* Leave the user to figure this out.

None of these are great options, so instead what we can do is tell API Extractor to generate API docs for the `devtools-protocol` package. This means if we have `function x(someArg: Protocol.Y)` the docs will generate a link on `Protocol.X` that the user can click to be taken to documentation for that part of the protocol.

We can improve the situation further with some small tweaks to the devtools protocol generator to make the outputted comments TSDoc compliant (I will be doing that today) but this PR just dumps in the generated documentation (there's quite a lot!.


mathiasbynens (approved):
I carefully reviewed every single li—

Rubberstamp LGTM

PR #6218, opened 2020-07-14
feat: change QueryHandler to contain QueryOne and QueryAll methods

johanbay:
Extends #5753 with support for registering both `$` and `$$` handlers under the same selection prefix, i.e `myCustomQueryHandlerPrefix/<selector>`.


src/common/JSHandle.ts
```diff
@@ -890,15 +887,17 @@ export class ElementHandle<
...
-      defaultHandler
+      selector
     );
-
+    const queryHandlerToArrayStr = `(element, selector) =>
+      Array.from((${queryHandler.queryAll})(element, selector))`;
+    /* The `eval` ensures that the query handler is inlined
+     * before sending the program text to the browser.
+     */
+    const queryHandlerToArray = eval(queryHandlerToArrayStr);
```
mathiasbynens: I'm wondering if we could do something like:

```js
const queryHandlerToArray = Function(
  'element',
  'selector',
  `return Array.from((${queryHandler.queryAll})(element, selector))`);
```

Or, even more simply, we could pass the string (`queryHandlerToArrayStr`) directly to `evaluateHandle`. WDYT?

src/common/DOMWorld.ts
```diff
@@ -33,7 +33,7 @@ import { isNode } from '../environment';
 
 // This predicateQueryHandler is declared here so that TypeScript knows about it
 // when it is used in the predicate function below.
-declare const predicateQueryHandler: QueryHandler;
+declare const predicateQueryHandler: Function;
```
paullewis: We seem to have a little loss here by switching to `Function`. Can we define the signature of the query handler like we did before?

src/common/JSHandle.ts
paullewis: +1

paullewis (commented):
Generally LGTM, but I'll await Mathias' approval.

src/common/DOMWorld.ts
johanbay: Ah, yes. We should still be able to something like this here:
```
declare const predicateQueryHandler: (
  element: Element | Document,
  selector: string
) => Element | Element[] | NodeListOf<Element>;
```

src/common/JSHandle.ts
johanbay: I actually tried both. `new Function` gives a type error since an `EvaluateHandleFn` is expected and casting seems to fail (at least things start getting undefined, but I haven't debugged that further though).
Passing the string directly to `evaluateHandle` seems like the correct way forward to me, but the pipeline for evaluating something passed as string is apparently quite different? As far as I can tell, `evaluateHandle` ignores the `args` if `pageFunction` is a string.

jackfranklin: @johanbay the typescript definitions are a work in progress - so please shout if they cause you problems. There could well be a problem with the types, not with your code. Happy to help with any of this :)

johanbay: @jackfranklin thanks! I was wrong on `Function('element', 'selector', <handlerBody>)` not working here, it actually does. Only problem is that the type `EvaluateHandleFn` clashes with the type `Function`. Do you see any reason not to change `(...args: unknown[]) => unknown` to `Function` in `EvaluateHandleFn` ([link to file](https://github.com/puppeteer/puppeteer/blob/fb806109a09e82b51734615a958dbd1518e0c5e4/src/common/EvalTypes.ts#L40))?

test/elementhandle.spec.ts
```diff
@@ -380,10 +379,9 @@ describe('ElementHandle specs', function () {
     });
     it('should wait correctly with waitFor', async () => {
       const { page, puppeteer } = getTestState();
-      puppeteer.__experimental_registerCustomQueryHandler(
-        'getByClass',
-        (element, selector) => element.querySelector(`.${selector}`)
-      );
+      puppeteer.__experimental_registerCustomQueryHandler('getByClass', {
+        queryOne: (element, selector) => element.querySelector(`.${selector}`),
+      });
```
mathiasbynens, line 395: Let's add some tests for a handler that supports both `queryOne` and `queryAll` (which I suspect will be the common case)

src/common/JSHandle.ts
```diff
@@ -791,15 +791,12 @@ export class ElementHandle<
...
   async $$(selector: string): Promise<ElementHandle[]> {
-    const defaultHandler = (element: Element, selector: string) =>
-      element.querySelectorAll(selector);
     const { updatedSelector, queryHandler } = getQueryHandlerAndSelector(
-      selector,
-      defaultHandler
+      selector
     );
 
     const arrayHandle = await this.evaluateHandle(
```
hanselfmu: Just a quick suggestion: I know the variable name `arrayHandle` is from the old code, but we can consider changing it to something like `handles` or `handleList`, to be more standard IMO.

src/common/JSHandle.ts
```diff
@@ -908,8 +906,8 @@ export class ElementHandle<
       ) => ReturnType | Promise<ReturnType>
     >(pageFunction, ...args);
     await arrayHandle.dispose();
-    /* This as exists for the same reason as the `as` in $eval above.
-     * See the comment there for a ful explanation.
+    /* This `as` exists for the same reason as the `as` in $eval above.
+     * See the comment there for a full explanation.
```
hanselfmu, line 912: Thanks for addressing these as well!

src/common/JSHandle.ts
johanbay: fixed

test/elementhandle.spec.ts
johanbay: Good idea! I have added a couple in d83b8b3.

OrKoN (approved):

src/common/JSHandle.ts
```diff
@@ -849,7 +846,7 @@ export class ElementHandle<
     await elementHandle.dispose();
 
     /**
-     * This as is a little unfortunate but helps TS understand the behavour of
+     * This `as` is a little unfortunate but helps TS understand the behaviour of
```
mathiasbynens: While we're at it:

```suggestion
     * This `as` is a little unfortunate but helps TS understand the behavior of
```

The source code is supposed to be `en-US`, not `en-GB` (sorry @jackfranklin!)

test/elementhandle.spec.ts
```diff
@@ -394,5 +392,44 @@ describe('ElementHandle specs', function () {
 
       expect(element).toBeDefined();
     });
+    it('should work when both queryOne and queryAll is registered', async () => {
```
mathiasbynens:
```suggestion
    it('should work when both queryOne and queryAll are registered', async () => {
```

mathiasbynens (approved):

src/common/DOMWorld.ts
```diff
@@ -26,7 +26,7 @@ import { ExecutionContext } from './ExecutionContext.js';
 import { TimeoutSettings } from './TimeoutSettings.js';
 import { MouseButton } from './Input.js';
 import { FrameManager, Frame } from './FrameManager.js';
-import { getQueryHandlerAndSelector, QueryHandler } from './QueryHandler.js';
+import { getQueryHandlerAndSelector } from './QueryHandler';
```
mathiasbynens:
```suggestion
import { getQueryHandlerAndSelector } from './QueryHandler.js';
```

to fix the CI issue

hanselfmu (approved):
LGTM, thanks Johan!

src/common/EvalTypes.ts
```diff
@@ -37,7 +37,7 @@ export type EvaluateFnReturnType<T extends EvaluateFn> = T extends (
 /**
  * @public
  */
-export type EvaluateHandleFn = string | ((...args: unknown[]) => unknown);
+export type EvaluateHandleFn = string | Function;
```
jackfranklin: it's a shame that we have to make this more generic; what happens without this change? We might be able to `// @ts-expect-error` elsewhere for now (whilst this is experimental)

src/common/JSHandle.ts
```diff
@@ -892,15 +889,16 @@ export class ElementHandle<
...
-      Array.from(element.querySelectorAll(selector));
     const { updatedSelector, queryHandler } = getQueryHandlerAndSelector(
-      selector,
-      defaultHandler
+      selector
+    );
+    const queryHandlerToArray = Function(
+      'element',
+      'selector',
+      `return Array.from((${queryHandler.queryAll})(element, selector));`
```
jackfranklin, line 898: for my understanding, why does this have to be defined like this and not as an actual function definition?

src/common/QueryHandler.ts
```diff
@@ -34,7 +35,7 @@ export function registerCustomQueryHandler(
   if (!isValidName)
     throw new Error(`Custom query handler names may only contain [a-zA-Z]`);
 
-  _customQueryHandlers.set(name, handler as QueryHandler);
```
jackfranklin, line 37: nice work getting rid of this `as` :)

test/elementhandle.spec.ts
```diff
@@ -405,5 +403,44 @@ describe('ElementHandle specs', function () {
...
+        'getByClass/foo',
+        (div) => div.textContent
+      );
+      expect(txtContent).toBe('text');
+
+      const txtContents = await page.$$eval('getByClass/foo', (divs) =>
+        divs.map((d) => d.textContent).join('')
+      );
+      expect(txtContents).toBe('textcontent');
+    });
```
jackfranklin, line 444: nice tests!

jackfranklin (approved):
This LGTM with one question about the change of type to `Function` and if we can avoid it.

src/common/EvalTypes.ts
jackfranklin: (or we can use `as ...` if needed)

src/common/JSHandle.ts
mathiasbynens: We need some way of inlining `String(queryHandler.queryAll)` into the function body. This PR started with an `eval`-based approach but then we moved towards `Function` which is slightly more minimal.

johanbay:
**Summary of changes:**
The experimental API for registering custom query handlers is slightly changed such that one registers query handlers containing two strategies: one for querying a single element and one for querying all matching elements.

The api for custom query handlers is unchanged:
```
__experimental_registerCustomQueryHandler(name: string, queryHandler: QueryHandler): void;
__experimental_unregisterCustomQueryHandler(name: string): void;
__experimental_customQueryHandlers(): Map<string, QueryHandler>;
__experimental_clearQueryHandlers(): void;
```
but `QueryHandler` is now defined as
```
export interface QueryHandler {
  queryOne?: (element: Element | Document, selector: string) => Element | null;
  queryAll?: (element: Element | Document, selector: string) => Element[] | NodeListOf<Element>;
}
```

src/common/EvalTypes.ts
johanbay: Fair point. I'll revert to `((...args: unknown[]) => unknown)`

mathiasbynens:
Discussed with Paul offline. He's fine with this, so not blocking on his review+ bit.

PR #6236, opened 2020-07-17
docs(troubleshooting): install libXss in Dockerfile

envoked:
The latest version of chromium requires libXss which isn't installed in the official node docker images.

```
Failed to launch the browser process!
/app/node_modules/puppeteer/.local-chromium/linux-756035/chrome-linux/chrome: error while loading shared libraries: libXss.so.1: cannot open shared object file: No such file or directory



TROUBLESHOOTING: https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md
```
This updates the `Dockerfile` example to also install `libxss1` which solves the issue. It also bumps the base docker image to the latest node LTS (12).

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -256,7 +256,7 @@ To fix, you'll need to install the missing dependencies and the
 latest Chromium package in your Dockerfile:
 
 ```Dockerfile
-FROM node:10-slim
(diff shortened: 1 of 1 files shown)


jackfranklin (approved):
Thank you!

PR #6263, opened 2020-07-23
param size to 'IO.read' solve problem firefox PDF

logycware:
problem  related to the size of the generated PDF with firefox

doc for IO.read: https://chromedevtools.github.io/devtools-protocol/tot/IO/


logycware:
> @googlebot I signed it!

Issue #5671, opened 2020-04-17
v3.0 - firefox pdf fails when printBackground set to true


kumar-muthu:
### Steps to reproduce

* Puppeteer version: 3.0
* Platform / OS version: macOs Mojave
* Node.js version: 12.16.1

**What steps will reproduce the problem?**

1. test.html

```html
<!doctype html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport"
          content="width=800, height=800">
    <style>
        .bg {
            background-image: url('https://images.unsplash.com/photo-1586973699006-2a096c90f847?ixlib=rb-1.2.1&ixid=eyJhcHBfaWQiOjEyMDd9&auto=format&fit=crop&w=2734&q=80');
            height: 800px;
            width: 800px;
            background-size: cover;
        }
    </style>
</head>
<body>
<div class="bg"></div>
</body>
</html>
```

2. js code
```javascript
const puppeteer = require('puppeteer');
const firefoxOptions = {
    product: 'firefox',
    extraPrefsFirefox: {
        // Enable additional Firefox logging from its protocol implementation
         'remote.log.level': 'Trace',
    },
    // Make browser logs visible
    dumpio: true,
};

(async() => {
    const browser = await puppeteer.launch(firefoxOptions);
    const page = await browser.newPage();
    await page.pdf({
        printBackground: true, // when set to false, there is no issue
        path: 'hn.pdf'});
    await browser.close();
})();
```

**What is the expected result?**
Convert to pdf without issues

**What happens instead?**
It results in the following error
```bash
1587125621499   RemoteAgent     TRACE   (connection {c236c872-af3d-7f40-8239-c03ac2d5072c})-> {"sessionId":1,"method":"Page.printToPDF","params":{"transferMode":"ReturnAsStream","landscape":false,"displayHeaderFooter":false,"headerTemplate":"","footerTemplate":"","printBackground":true,"scale":1,"paperWidth":8.5,"paperHeight":11,"marginTop":0,"marginBottom":0,"marginLeft":0,"marginRight":0,"pageRanges":"","preferCSSPageSize":false},"id":18}
1587125622620   RemoteAgent     TRACE   <-(connection {c236c872-af3d-7f40-8239-c03ac2d5072c}) {"sessionId":1,"id":18,"result":{"data":null,"stream":"18c71222-e870-4f4d-b68b-11f51869596a"}}
[...]

PR #6264, opened 2020-07-23
param size to 'IO.read' solve problem firefox PDF 

logycware:
problem related to the size of the generated PDF with firefox

doc for IO.read: https://chromedevtools.github.io/devtools-protocol/tot/IO/

### DIFF
--- a/src/common/helper.ts
+++ b/src/common/helper.ts
@@ -211,7 +211,10 @@ async function readProtocolStream(
   if (path) file = await openAsync(path, 'w');
   const bufs = [];
   while (!eof) {
-    const response = await client.send('IO.read', { handle: handle, size: 102400 });
+    const response = await client.send('IO.read', { 
+      handle: handle, 
+      size: 102400 
+    });
     eof = response.eof;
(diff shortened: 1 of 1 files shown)


jackfranklin:
@mjzffr @whimboo would one of you mind taking a look? :)

mjzffr:
@logycware Could you provide more context? What issue does this solve? Why 102400? Could you add a Puppeteer unit test that captures the issue?

The IO.read implementation in Firefox sets a [default chunk size](https://searchfox.org/mozilla-central/rev/227f22acef5c4865503bde9f835452bf38332c8e/remote/domains/parent/IO.jsm#99)

mjzffr:
@logycware Oops, I missed #5671, but my other questions still stand. Thanks :)

mjzffr:
I did a bit of digging and found [the source of the problem](https://bugzilla.mozilla.org/show_bug.cgi?id=1632722#c2). We'll push a fix to Firefox and I'll comment on #5671 when that lands so you can try it out.

Thanks for bringing this to my attention. :)

PR #6268, opened 2020-07-23
feat: page.waitForTimeout

### LINKED ISSUE #6214: Proposal: deprecate `waitFor` and add `waitForTimeout`
Puppeteer's API currently has a few `waitForX` functions:

* `waitForSelector`
* `waitForFunction`
* `waitForXPath`

(And others that are less relevant to this proposal).

We also ship `waitFor`, which has the following docs:

```
This method behaves differently with respect to the type of the first parameter:

* if selectorOrFunctionOrTimeout is a string, then the first argument is treated as a selector or xpath, depending on whether or not it starts with '//', and the method is a shortcut for page.waitForSelector or page.waitForXPath
* if selectorOrFunctionOrTimeout is a function, then the first argument is treated as a predicate to wait for and the method is a shortcut for page.waitForFunction().
* if selectorOrFunctionOrTimeout is a number, then the first argument is treated as a timeout in milliseconds and the method returns a promise which resolves after the timeout
* otherwise, an exception is thrown
```

So if you call `waitFor(someFunc)`, you're just calling `waitForFunction`. Similarly, `waitFor('.foo')` calls `waitForSelector`, and `waitFor('//div')` calls `waitForXPath`. The only unique behaviour `waitFor` provides is calling it with a time: `waitFor(1000)` does exactly what you expect.

Another downside of this method is that it's harder to type safely with TypeScript (you can do it via overloads, but it's uneccessarily complex, especially when all the standalone `waitForX` functions are already typed).

I think it's confusing that most of the API for waiting is explicitly named, e.g. `waitForSelector`, yet the only way to wait for a timeout is to call `waitFor`.

I'd propose a nicer, more consistent API would:

1. Add `waitForTimeout`. This is consistent with the naming of the other wait for methods, which are all named `waitForX`.
2. Remove `waitFor`.

Rather than do this as one breaking change we can do this change in two parts:

1. Ship a version with `waitForTimeout`. Deprecate `waitFor` and log a message when it's used, but maintain its current behaviour.
2. After a reasonable amount of time, ship a release that removes `waitFor`. The migration path should be straightforward, users will have to swap `waitFor` for the correct `waitForX` method. We'll have a `waitForX` for every behaviour of the original `waitFor`, so this won't be more than a renaming of some method calls.

jackfranklin:
This PR adds `page.waitForTimeout` (and subsequently, `frame.waitForTimeout`) and deprecates the `page.waitFor` and `frame.waitFor` methods.

None of the functionality that `waitFor` provided is lost; it's now all available via the more explicit `waitForX` functions. This PR doesn't remove `waitFor`, all its functionality is still available. I'd suggest we ship this as a minor release, give it a while for users to adjust and then remove `waitFor` in the next reasonable breaking release.

Fixes #6214.


docs/api.md
```diff
@@ -2235,6 +2244,26 @@ const puppeteer = require('puppeteer');
...
+
+Pauses script execution for the given number of seconds before continuing:
+
+```js
+const puppeteer = require('puppeteer');
+
+(async () => {
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+  let currentURL;
```
OrKoN: why is currentURL needed here?

test/waittask.spec.ts
```diff
@@ -28,6 +29,12 @@ describe('waittask specs', function () {
   setupTestPageAndContextHooks();
 
   describe('Page.waitFor', function () {
+    /* This method is deprecated but we don't want the warnings showing up in
+     * tests. Until we remove this method we still want to ensure we don't break
+     * it.
+     */
+    beforeEach(() => sinon.stub(console, 'warn').callsFake(() => {}));
```
OrKoN, line 36: when is the stub restored?

OrKoN (approved):

docs/api.md
jackfranklin: Ah good spot - was a copy paste from another example

test/waittask.spec.ts
jackfranklin: Automatically after each test: https://github.com/puppeteer/puppeteer/blob/main/test/mocha-utils.ts#L273

docs/api.md
```diff
@@ -3148,6 +3184,27 @@ const puppeteer = require('puppeteer');
...
```
vsemozhetbyt: Ditto)

hanselfmu (approved):

mathiasbynens (approved):

PR #6269, opened 2020-07-23
chore: test Connection class in the browser

jackfranklin:
This commit creates a test that runs in the browser and ensures that we
can instantiate a pptr Connection with a fake transport and send/receive
messages.

Longer term I'd like to have tests that actually connect to a browser
and send CDP messages but that needs some changes to web-test-runner
which I'm going to speak to them about - the DevTools URL needs to be
exposed some how for us to use and connect to.

### DIFF
--- a/test-browser/connection.spec.js
+++ b/test-browser/connection.spec.js
@@ -1,0 +1,61 @@
+/**
+ * Copyright 2020 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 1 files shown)


test-browser/connection.spec.js
```diff
@@ -0,0 +1,61 @@
...
+  close() {}
+}
+
+describe('Connection', () => {
+  it('can be created in the browser and send/receive messages', async () => {
+    let receivedOutput = '';
+    const connection = new Connection('fake-url', new EchoTransport());
+
+    /**
+     * 1 here is the ID because pptr increments a counter from 0 for each
```
paullewis: nit: That took me a couple of reads, just because you started with the number 1, I think!

paullewis (approved):

test-browser/connection.spec.js
jackfranklin: ah yeah this could be worded more clearly, thanks! I'll tweak.

PR #6274, opened 2020-07-24
fix: broken socks proxy for binary downloads

### LINKED ISSUE #4401: Chromium skip install without error log when terminal behind socks5 proxy on Mac.
### Steps to reproduce

* Puppeteer version: **1.15.0**
* Platform / OS version: **MacOS 10.14.4** 
* URLs (if applicable): 
* Node.js version: **10.15.3**

1. export all_proxy=socks5://127.0.0.1:1080/
2. npm -g i puppeteer

```bash
➜ npm -g i puppeteer

> puppeteer@1.15.0 install /Users/Like/.nvm/versions/node/v10.15.3/lib/node_modules/puppeteer
> node install.js

Downloading Chromium r650583 - 87.7 Mb [====================] 100% 0.0s
Chromium downloaded to /Users/Like/.nvm/versions/node/v10.15.3/lib/node_modules/puppeteer/.local-chromium/mac-650583
+ puppeteer@1.15.0
added 43 packages from 22 contributors in 19.111s
```

**What happens instead?**
```bash
➜ npm -g i puppeteer

> puppeteer@1.15.0 install /Users/Like/.nvm/versions/node/v10.15.3/lib/node_modules/puppeteer
> node install.js

+ puppeteer@1.15.0
added 43 packages from 22 contributors in 6.571s
```
`.local-chromiun` under puppetter package is empty.

### LINKED ISSUE #7597: [Bug]: Chromium download over http proxy fails
### Bug description

1. Have a proxy server setup, and set HTTP_PROXY and HTTPS_PROXY environment variables to point to it, both as http:// urls.
2. Install puppeteer
3. Download of Chromium during install fails with ERR_INVALID_PROTOCOL

It appears that this is happening because we are proxying HTTPS over an HTTP proxy.

### Puppeteer version

5.5.0

### Node.js version

16.9.1

### npm version

7.21.1 (but we use yarn, version 1.22.5)

Linux

### Relevant log output

```shell
2021-09-23T19:25:57.805Z puppeteer:fetcher Downloading binary from https://storage.googleapis.com/chromium-browser-snapshots/Linux_x64/818858/chrome-linux.zip
ERROR: Failed to set up Chromium r818858! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
TypeError [ERR_INVALID_PROTOCOL]: Protocol "https:" not supported. Expected "http:"
    at new NodeError (node:internal/errors:371:5)
    at new ClientRequest (node:_http_client:158:11)
    at Object.request (node:https:353:10)
    at httpRequest (<REDACTED>/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:488:17)
    at downloadFile (<REDACTED>/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:357:21)
    at BrowserFetcher.download (<REDACTED>/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:239:19)
    at async downloadBrowser (<REDACTED>/node_modules/puppeteer/lib/cjs/puppeteer/node/install.js:48:5) {
  code: 'ERR_INVALID_PROTOCOL'
}
```

spacelatte:
Hi there,

This patch allows BrowserFetcher module to use socks proxies defined in `ALL_PROXY` environment variable or `proxy=socks5://...` line in `~/.npmrc` (or related configuration)

> Fixes: #4401

Normally with unsupported proxy configuration install fails silently (yields empty .local-chromium directory). It still is btw :)
To solve this I can add `default:` case to throw an error as a fallback...

Ps. Btw I'm not TS dev, sometimes writing some JS in free time... Feedback appreciated :)


docs/api.md
```diff
@@ -406,7 +406,7 @@ You will then need to call [`puppeteer.connect([options])`](#puppeteerconnectopt
 Puppeteer looks for certain [environment variables](https://en.wikipedia.org/wiki/Environment_variable) to aid its operations.
 If Puppeteer doesn't find them in the environment during the installation step, a lowercased variant of these variables will be used from the [npm config](https://docs.npmjs.com/cli/config).
 
-- `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
+- `HTTP_PROXY`, `HTTPS_PROXY`, `ALL_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
```
spacelatte, line 414: To show that's a valid variable to be useful. I first tried `SOCKS_PROXY` variable but that is not a standard variable.

experimental/puppeteer-firefox/install.js
```diff
@@ -35,12 +35,15 @@ if (revisionInfo.local)
...
 const NPM_HTTPS_PROXY = process.env.npm_config_https_proxy || process.env.npm_config_proxy;
 const NPM_HTTP_PROXY = process.env.npm_config_http_proxy || process.env.npm_config_proxy;
+const NPM_ALL_PROXY = process.env.npm_config_all_proxy || process.env.npm_config_proxy;
 const NPM_NO_PROXY = process.env.npm_config_no_proxy;
 
 if (NPM_HTTPS_PROXY)
   process.env.HTTPS_PROXY = NPM_HTTPS_PROXY;
 if (NPM_HTTP_PROXY)
   process.env.HTTP_PROXY = NPM_HTTP_PROXY;
+if (NPM_ALL_PROXY)
```
spacelatte, line 46: I've added this to just to follow the pattern. :)

package.json
```diff
@@ -49,6 +49,7 @@
     "devtools-protocol": "0.0.781568",
     "extract-zip": "^2.0.0",
     "https-proxy-agent": "^4.0.0",
+    "socks-proxy-agent": "^5.0.0",
```
spacelatte, line 52: I actually am not sure what to do here Found a few modules.

1. https://www.npmjs.com/package/socks-proxy-agent Similar to current Https module
2. https://www.npmjs.com/package/proxy-agent Seems versatile can be used instead of 2 separate modules.
3. https://www.npmjs.com/package/simple-proxy-agent Low # of downloads
4. https://www.npmjs.com/package/socks-proxy-agent-ts Even lower # of downloads

I've used 1st but 2nd seems better suited (PAC support et.al.) So directly pass logic to that module entirely.
Note: 1, 2, and 4 are from the same developer/publisher.

src/install.ts
```diff
@@ -79,10 +79,13 @@ export async function downloadBrowser() {
       process.env.npm_config_https_proxy || process.env.npm_config_proxy;
     const NPM_HTTP_PROXY =
       process.env.npm_config_http_proxy || process.env.npm_config_proxy;
+    const NPM_ALL_PROXY =
+      process.env.npm_config_all_proxy || process.env.npm_config_proxy;
     const NPM_NO_PROXY = process.env.npm_config_no_proxy;
 
     if (NPM_HTTPS_PROXY) process.env.HTTPS_PROXY = NPM_HTTPS_PROXY;
     if (NPM_HTTP_PROXY) process.env.HTTP_PROXY = NPM_HTTP_PROXY;
+    if (NPM_ALL_PROXY) process.env.ALL_PROXY = NPM_ALL_PROXY;
```
spacelatte, line 90: Again following the pattern :)

src/node/BrowserFetcher.ts
```diff
@@ -27,7 +27,8 @@ import { debug } from '../common/Debug.js';
 import { promisify } from 'util';
 import removeRecursive from 'rimraf';
 import * as URL from 'url';
-import ProxyAgent from 'https-proxy-agent';
+import HttpsProxyAgent from 'https-proxy-agent';
+import { SocksProxyAgent, SocksProxyAgentOptions } from 'socks-proxy-agent';
```
spacelatte, line 31: Renamed import name since there are different types of proxies now. Also Socks module is having a different structure than Https one.

src/node/BrowserFetcher.ts
```diff
@@ -561,7 +562,7 @@ function httpRequest(
 
   type Options = Partial<URL.UrlWithStringQuery> & {
     method?: string;
-    agent?: ProxyAgent;
+    agent?: HttpsProxyAgent | SocksProxyAgent | any;
```
spacelatte, line 565: Docs explicitly said no `any` :) But I couldn't found a better way to handle those different types at once. Some pointers really appreciated :)

src/node/BrowserFetcher.ts
```diff
@@ -572,23 +573,32 @@ function httpRequest(
...
-      } as ProxyAgent.HttpsProxyAgentOptions;
-
-      options.agent = new ProxyAgent(proxyOptions);
-      options.rejectUnauthorized = false;
+    const parsedProxyURL = URL.parse(proxyURL);
+    const proxyOptions = {
+      ...parsedProxyURL,
+      secureProxy: false,
+    };
+    switch (true) {
```
spacelatte, line 581: IMHO not to best way handling those (may be anti-pattern too)

Idk if TS had such thing:
```
let options = when {
  proxyURL.match(...) -> new SocksProxyAgent(...)
  url.startsWith(...) -> new HttpsProxyAgent(...)
}
```

src/node/BrowserFetcher.ts
```diff
@@ -572,23 +573,32 @@ function httpRequest(
...
+      case proxyURL.match(/^socks(4|5|5h)?:/) !== null:
```
spacelatte, line 582: I firstly added multiple OR (`||`) statemens here but seemed ugly. So I've replaced it with regexp thinking not much performance hit since its only done once at the initial install phase.

Edit: Also tried not to do over-engineering (eg: `/^socks(4|5[h]?)?:` since its also longer and messy.)

spacelatte:
@googlebot I signed it!

spacelatte:
@googlebot I fixed it.

spacelatte:
Hi @mathiasbynens & @jackfranklin ,

Sorry to bother you. Is this looking good? Can you give some hints/pointers if it's not :) 

Thanks!

jackfranklin:
Hey @pvtmert thanks for the PR and I'm sorry for the lack of response. 

This PR looks OK to me but I'm a bit worried now about the amount of different configurations you can apply for proxies, given we have `HTTP_PROXY`, `HTTPS_PROXY` and now potentially `ALL_PROXY`. I'm not sure if those are the clearest names and I wonder if we could think about that.

I'm also not familiar with SOCKS Proxy at all but what you've done seems reasonable. My only thought is how common it is (it may be extremely - as I say I'm unaware of it!) and if it's worth supporting in Puppeteer but given this is a relatively small change maybe that's fine.

@mathiasbynens wdyt?

jschfflr (changes requested):
@pvtmert Thanks for your contribution! Could you explain in which situation someone would be using a SOCKS proxy over a HTTPS proxy for example? 
And regarding the naming (as @jackfranklin pointed out): Maybe calling the env variable `SOCKS_PROXY` would be more helpful to users. 
Let me know what you think! Happy to get this landed :)

spacelatte:
Hi, it's been a while & I probably need to rebase from `main`  :)

> Could you explain in which situation someone would be using a SOCKS proxy over a HTTPS proxy for example?

Yeah, for instance when you do the _dynamic forwarding_ through SSH _tunnel_. Eg: via `ssh -D 1080` it opens `localhost:1080` as SOCKS5 proxy.

User does not need to setup a dedicated proxy server such as apache2 or polipo...

Edit: For the use-case part, the company I worked for at the time had intranet. Packages etc mirrored inside using internal registry. Since both the registry and the CI servers didn't had access, we've been able to gateway them through another "jumphost". Which is basically a SSH server plus using dynamic-forwarding.

> And regarding the naming (as @jackfranklin pointed out): Maybe calling the env variable `SOCKS_PROXY` would be more helpful to users.

Yeah, can be. ALL_PROXY uses the URI scheme to see what kind of proxy it needs to use. For instance: `socks5://localhost:1080` or `http://localhost:3128`...

Plus, ALL_PROXY is the standard ENV variable exported by GNOME, KDE & various other desktop environments...

docs/api.md
jschfflr: Ah ok, I see. Didn't know about the standard :)

jschfflr: So maybe the text should also be updated - currently it says ...defines HTTP proxy settings...

experimental/puppeteer-firefox/install.js
jschfflr: LGTM!

package.json
jschfflr: Given that `proxy-agent` already implements all the different protocols and resolving ENV variables it sounds like the best option to me.

src/node/BrowserFetcher.ts
jschfflr: I think in this case `| any` is not really needed.

src/node/BrowserFetcher.ts
jschfflr: switch/case could be used for that but I think going with a classic if/elseif is the better way here to keep the code simple.

jschfflr (changes requested):
I feel like using the `proxy-agent` package that you recommended might be the best option to go forward because it already implements everything we could need here, right?

jschfflr:
> Hi, it's been a while & I probably need to rebase from `main` :)
> 
> > Could you explain in which situation someone would be using a SOCKS proxy over a HTTPS proxy for example?
> 
> Yeah, for instance when you do the _dynamic forwarding_ through SSH _tunnel_. Eg: via `ssh -D 1080` it opens `localhost:1080` as SOCKS5 proxy.
> 
> User does not need to setup a dedicated proxy server such as apache2 or polipo...
> 
> Edit: For the use-case part, the company I worked for at the time had intranet. Packages etc mirrored inside using internal registry. Since both the registry and the CI servers didn't had access, we've been able to gateway them through another "jumphost". Which is basically a SSH server plus using dynamic-forwarding.
> 
> > And regarding the naming (as @jackfranklin pointed out): Maybe calling the env variable `SOCKS_PROXY` would be more helpful to users.
> 
> Yeah, can be. ALL_PROXY uses the URI scheme to see what kind of proxy it needs to use. For instance: `socks5://localhost:1080` or `http://localhost:3128`...
> 
> Plus, ALL_PROXY is the standard ENV variable exported by GNOME, KDE & various other desktop environments...

Thanks for your explanations! I didn't know much about proxies before so this was definitely helpful :)

jschfflr:
Just for reference: https://github.com/puppeteer/puppeteer/pull/6577

jschfflr:
@pvtmert Gentle ping to see if there are any updates on this or if you want me to take a look at this :)

OrKoN (changes requested):
Thanks for the PR. I think we need to get it updated and resolve the previous discussions.

Issue #6255, opened 2020-07-21
ConsoleMessage type and text not interoperable between Chrome and Firefox


foolip:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 5.2.0
* Platform / OS version: macOS 10.15.5
* Node.js version: v10.16.3

**What steps will reproduce the problem?**

_Please include code that reproduces the issue._

```js
'use strict';

const puppeteer = require('puppeteer');

async function main() {
  for (const product of ['chrome', 'firefox']) {
    const browser = await puppeteer.launch({product});
    const page = await browser.newPage();
    await page.goto('about:blank');
    const msgPromise = new Promise((resolve) => {
      page.on('console', resolve);
    });
    await page.evaluate('console.log("message")');
    const msg = await msgPromise;
    console.log({
      product,
      type: msg.type(),
      text: msg.text(),
      args: msg.args(),
    });
    await browser.close();
  }
}

main();
```

Output from running the above is:
```
{ product: 'chrome',
  type: 'log',
  text: 'message',
  args:
   [ JSHandle {
       _disposed: false,
       _context: [ExecutionContext],
       _client: [CDPSession],
       _remoteObject: [Object] } ] }
{ product: 'firefox',
  type: 'verbose',
  text: [ 'message' ],
  args: [] }
```

**What is the expected result?**

The `msg.type()` and `msg.text()` should be the same for Chrome and Firefox, to make it possible to use console messages without switching on which product is under test.

**What happens instead?**

- Chrome: `msg.type()` is "log" and `msg.text()` is "message".
- Firefox: `msg.type()` is "verbose" and `msg.text()` is an array with a single item "message"

`msg.args()` is also different, but I don't know what it should be, and am not using it myself.

christian-bromann:
The [`location`](https://pptr.dev/#?product=Puppeteer&version=v5.2.1&show=api-consolemessagelocation) method also shows different return values. Firefox is missing `columnNumber` here. I added a PR to increase the test coverage around this area a bit.

jackfranklin:
@mjzffr this might be a FF issue? I'm not 100% - should we also consider Puppeteer itself translating Chrome/FF consoles into an interface that is consistent regardless of browser?

mjzffr:
There are a number of issues here. tl;dr is that Firefox doesn't implement all the CDP events yet + there may be some discrepancies between Blink and Gecko.

In the Chrome case, the sample script is reporting data based on

```json
[...]

PR #6282, opened 2020-07-27
test: assert location value of ConsoleMessage

christian-bromann:
Issue #6255 shows differences in the `ConsoleMessage` object between Chromium and FF. These differences are already covered in the unit (I guess they are actually e2e tests 😉) tests. This small patch enhances the coverage to include the `location` method as it seems they differ between the browsers too.


hanselfmu (approved):
LGTM, thanks 👍

PR #6289, opened 2020-07-29
Ship Puppeteer TypeScript definitions

jackfranklin:
This PR:

* uses API-Extractor to generate a large .d.ts file with all the types in. This file is great, but it's missing the default export (e.g. the definition that when you import/require Puppeteer you get an instance of the `Puppeteer` class).
* Therefore, once the file has been generated, we add the required lines to declare the default export to the typings file.
* This _should_ mean that importing from `'puppeteer'` will work in TS land, as all the types will flow in from `puppeteer.d.ts`, as linked to via the `typings` field in our `package.json`.

### DIFF
--- a/api-extractor.json
+++ b/api-extractor.json
@@ -12,7 +12,8 @@
   },
 
   "dtsRollup": {
-    "enabled": false
+    "enabled": true,
+    "untrimmedFilePath": "<projectFolder>/lib/typings/puppeteer.d.ts"
   },
 
   "tsdocMetadata": {
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "5.2.1-post",
   "description": "A high-level API to control headless Chrome over the DevTools Protocol",
   "main": "./cjs-entry.js",
+  "typings": "./lib/typings/puppeteer.d.ts",
   "repository": "github:puppeteer/puppeteer",
   "engines": {
     "node": ">=10.18.1"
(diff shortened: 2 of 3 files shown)


jackfranklin:
@jsoverson @marvinhagemeister I feel like you two are more versed in shipping TS types than me; is this the right approach / will this result in the right thing?  

I'm going to try creating a new project and install Puppeteer from this branch but if either of you (or anyone else reading!) has any pointers that'd be great - if you have time, of course :)

jsoverson:
@jackfranklin TBH I don't know the best practices for large projects like this. I've gotten by just by re-exporting what I want form the index/main file so tsc generates everything automatically.

I've only used api-extractor for generating documentation and testing inline examples (see [shift-refactor](https://github.com/jsoverson/shift-refactor)), not generating types.

You may want to poke the TS maintainers to get some advice. I haven't really dealt with namespaces and that seems like something that should come into play here.

marvinhagemeister:
I share similar experiences as @jsoverson . I've only ever used `tsc` to generate types by pointing it to the main entry. With `allowJs` enabled it will also extract JSDoc types. From an interoperability  perspective TS doesn't care if there are lots of little `.d.ts` files or one giant one. The `types/typings` field in `package.json` only needs to point at the main entry file. The many files approach is definitely easier to generate and to maintain for sure.

adamdbradley:
Having one dts bundle is nice, but really doesn't matter much in the end. For simplicity I'd recommend to:

1) set the a `declarationDir` output directory in the [ts config compiler options](https://www.typescriptlang.org/docs/handbook/compiler-options.html) (and gitignore that types directory so the build files don't end up in the repo). This way can also serve as a way you're only shipping one copy of the dts files, rather than the same dts files for both esm and cjs. Directory could be named anything, but usually `types`.
2) Point the `types` property in the package.json to the generated declarationDir, such as `"types": "types/index.d.ts"`
3) Add `types` to the `files` array in package.json so that directory gets published to npm too.
4) In `src/index.ts`, I think you'll want to manually export each of the types you'd like to export, and using the `export type {...}` will help ensure it's only exporting the types and not any actual code. This way the generated `index.d.ts` serves as the same entry as `index.js`.

Hope this helps.

jackfranklin:
Replacing this PR with https://github.com/puppeteer/puppeteer/pull/6745.

@adamdbradley I tried putting it all into one folder but hit upon problems with TypeScript picking it up. It also may be useful to be able to ship separate CJS / ESM typedefs (or it may not) so I think we'll just ship the .d.ts files inline with the source code. Please take a look at the above PR 👍

PR #6296, opened 2020-08-01
docs(troubleshooting.md): updated commands

prashantghimire:
prior command installs stable version of chrome (RUN command), but after that, the docs is referring to unstable version.
added correct commands for troubleshooting in linux environment.


prashantghimire:
@googlebot I signed it!

mathiasbynens (approved):

PR #6301, opened 2020-08-03
docs(examples/search) change searching for searchbox to .devsite class

Talendran:
I have recently begun learning coding and have started looking through pupeteer when the search example failed to pull results from the search. I changed the search example to instead of looking for '#searchbox input' to search for the class '.devsite-search-field' since searchbox does not exist in the code of the website


Talendran:
@googlebot I signed it!

jackfranklin:
Thanks for the PR! What do you think about using `page.type('[aria-label="Search box"]')` ? That might be less brittle than using a class.

jschfflr (approved):
This change looks good to me! Thanks!

Issue #6217, opened 2020-07-14
Shadow Dom support


rameshrc:
Is there any support for finding elements in the Shadow Dom? Our application uses lit-html components. Please provide some examples.

Rainase:
> Is there any support for finding elements in the Shadow Dom? Our application uses lit-html components. Please provide some examples.

Try this ;) 
`const shadowButton = page.$('pierce/.my-btn-style')`

PR #6314, opened 2020-08-06
Proposal: fetch elements within shadow root with `shadow$` and `shadow$$`

christian-bromann:
There are various issues on missing support for fetching elements within the shadow DOM of elements. Some related issues are:

- https://github.com/puppeteer/puppeteer/issues/6217
- https://github.com/puppeteer/puppeteer/issues/5405
- https://github.com/puppeteer/puppeteer/issues/4171

Also for Puppeteer embedder projects like WebdriverIO:

- https://github.com/webdriverio/webdriverio/issues/4484

This PR proposes `shadow$` and `shadow$$` to allow query an element through element shadow roots. Given the following DOM structure:

```html
<html>
<body>
  <div id="elem">
    #shadow-root
    <customA></customA>
    <customA>
      #shadow-root
      <subCustomB>
        #shadow-root
        <div class="container"></div>
      </subCustomB>
    </customA>
  </div>
</body>
</html>
```

Given someone would fetch the root of an shadow element:

```js
const elem = await page.$('#elem')
```

The following queries would give following results:

- ```js
  elem.shadow$(['customA', 'subCustomB', '.container'])
  ```
  would return `null` because we would always pick the first element being found
- ```js
  const customAs = await elem.shadow$$(['customA'])
  return customAs[1].shadow$(['subCustomB', '.container'])
  ```
  would return `div.container`

One idea could be to always use `Element.querySelectorAll` and iterate through all nodes. This could be potentially a very expensive operation though and would only make sense for `shadow$$`.

WDYAT?

### DIFF
--- a/src/common/JSHandle.ts
+++ b/src/common/JSHandle.ts
@@ -789,6 +789,26 @@ export class ElementHandle<
   }
 
   /**
+   * Runs `element.shadowRoot.querySelector` within the page. If no element matches the selector,
+   * the return value resolves to `null`.
+   */
+  async $shadow(selector: string[]): Promise<ElementHandle | null> {
+    const { updatedSelector, queryHandler } = getQueryHandlerAndSelector(
+      selector.join(', ')
+    );
+
+    const handle = await this.evaluateHandle(
+      queryHandler.shadowQueryOne,
+      updatedSelector
+    );
+
+    const element = handle.asElement();
+    if (element) return element;
+    await handle.dispose();
+    return null;
+  }
+
+  /**
    * Runs `element.querySelectorAll` within the page. If no elements match the selector,
    * the return value resolves to `[]`.
    */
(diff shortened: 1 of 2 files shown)


LarsDenBakker:
👍 I think this is a great step for testing web components. Always picking the first instance found is fine here I think.

One concern I have with this is that it makes tests very brittle, and tightly coupled to private APIs of web components. You have the specify the entire path of all the elements. When you build your whole application out of web components, you have a lot of implementation details in your tests. 

In traditional webpages, people are used to being able to select elements by just using some landmarks such as ids, element names, form references or data attributes. With shadow dom this used by possible with the `/deep/` selector but this was removed from the spec.

I think something like this is necessary to have, but it doesn't fully address all the use cases. So I wouldn't close all the related issues. There should also be a way to query elements deeply, such as implemented by libraries like https://github.com/Georgegriff/query-selector-shadow-dom

jackfranklin:
Pinging @mathiasbynens @johanbay @paullewis on this as they've worked on the query handler parts more than me 👍

mathiasbynens:
Thanks for kicking off this discussion, @christian-bromann! As a general note, building on top of the existing custom query handler infrastructure makes a lot of sense 👍 

This PR lets you select across a single shadow root by explicitly referencing it. That's indeed better than anything we support out of the box right now. 

But, I think where we want to go eventually is a little different: as a user, I want to either select using the default magic (`querySelector{All}`), or select across _all_ shadow roots (without having to reference any of them in particular). This patch only lets me pierce a single shadow root, by explicitly referencing it.

WDYT?

LarsDenBakker:
What can we do to move forward with this?

christian-bromann:
@LarsDenBakker I haven't been able to follow up on this but will take some time this week to address @mathiasbynens comment.

jackfranklin:
@mathiasbynens is this PR still useful? Or is the work than @johanbay has been doing now going to enable this more easily?

johanbay:
With the changes to the query handler infrastructure, we can do something like this https://github.com/puppeteer/puppeteer/pull/6509 to support querying elements inside shadow roots.
So in the examples above one would do `page.$('pierce/.container')`.
Does that work for you @LarsDenBakker and @christian-bromann?

LarsDenBakker:
Yep that works for me.

christian-bromann:
Sounds good to me too? Shall we close this PR then?

johanbay:
Yes it sounds like we can close this one and merge https://github.com/puppeteer/puppeteer/pull/6509. Thanks!

Issue #5673, opened 2020-04-17
Re-enable launching Firefox on Windows CI test


jackfranklin:
@mjzffr in #5637 we had to disable this test on the Windows CI run because it would timeout.

```js
      it('should be able to launch Firefox', async() => {
        const {puppeteer} = getTestState();
        const browser = await puppeteer.launch({product: 'firefox'});
        const userAgent = await browser.userAgent();
        await browser.close();
        expect(userAgent).toContain('Firefox');
      });
```

I'm not quite sure why but it'd be great to have a look to see if we can enable this test on Windows again. Would you mind having a look when you get a chance? Thank you :)

mathiasbynens:
cc @whimboo

hlawuleka:
Either this or `Could not find browser revision latest. Run "npm install" or "yarn install" to download a browser binary.`

Yet there's a line in the docs that suggests that all you need to do is set `product: 'firefox'`.

mjzffr:
I'll def take a look when I have the chance.

whimboo:
So didn't it find a suitable download of Firefox Nightly? Or which line of the test actually times out?

mjzffr:
From a user perspective, `Browser.close` is broken for Firefox on Windows.

I believe `BrowserRunner` needs a special case for Firefox on Windows to record the actual browser process to kill, which is a child of the initial process we run. 

For future reference, here are debug logs. 

```
2020-08-06T19:50:52.299Z puppeteer:launcher Calling C:\Users\mf\dev\puppeteer\.local-firefox\win64-81.0a1\firefox\firefox.exe --no-remote --foreground --wait-for-browser --headless about:blank --remote-debugging-port=0 --profile C:\Users\mf\AppData\Local\Temp\puppeteer_dev_firefox_profile-Ij7Ruz
*** You are running in headless mode.
DevTools listening on ws://localhost:49962/devtools/browser/f6f6da26-da96-4145-98af-21aab53ef000
2020-08-06T19:51:03.660Z puppeteer:protocol:SEND {"method":"Target.setDiscoverTargets","params":{"discover":true},"id":1}
1596743463673   RemoteAgent     TRACE   (connection {dafcaf62-af68-434b-bab1-fcc99fd65dc0})-> {"method":"Target.setDiscoverTargets","params":{"discover":true},"id":1}
[...]

PR #6315, opened 2020-08-06
fix(FirefoxLauncher): use wait-for-process option on Windows

mjzffr:
Without `--wait-for-process`, `puppeteer.launch` returns before Firefox is ready on Windows, so this should be part of the default Firefox arguments on that platform.

On Windows, firefox.exe starts a "launcher process" that is responsible for configuring and starting the "real" browser process. Passing `--wait-for-process` makes that launcher process wait for the entire life of the browser process, and will also propagate the browser process's exit code as its own.

Note that the test highlighted in https://github.com/puppeteer/puppeteer/issues/5673 is still expected to fail because `BrowserRunner` doesn't track the right process on Windows, which breaks its `close` callback.

### DIFF
--- a/src/node/Launcher.ts
+++ b/src/node/Launcher.ts
@@ -467,6 +467,9 @@ class FirefoxLauncher implements ProductLauncher {
 
   defaultArgs(options: ChromeArgOptions = {}): string[] {
     const firefoxArguments = ['--no-remote', '--foreground'];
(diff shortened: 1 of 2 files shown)


jackfranklin (approved):
Nice fix, thanks!

mathiasbynens (approved):

PR #6322, opened 2020-08-07
Adding a quick copy paste to required dependencies

LucasAsafe:
Me and the team were loosing some time putting this into a apt install command, so i thought it would be better having the instalation command already bellow the dependencies not only for debian but the other distros.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -84,6 +84,10 @@ lsb-release
 wget
 xdg-utils
 ```
+
+```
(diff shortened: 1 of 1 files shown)


jackfranklin:
I like the idea of being able to copy and paste but not the fact that we now have two duplicate lists and they will get out of sync. What if we make the list just the copy paste command, but break it onto multiple lines?

```
sudo apt-get install 
  wget \
  ...
```

WDYT? Thanks for the PR 👍

LucasAsafe:
> I like the idea of being able to copy and paste but not the fact that we now have two duplicate lists and they will get out of sync. What if we make the list just the copy paste command, but break it onto multiple lines?
> 
> ```
> sudo apt-get install 
>   wget \
>   ...
> ```
> 
> WDYT? Thanks for the PR 👍

Liked it, gonna make it for all distros later and open a new PR

PR #6334, opened 2020-08-12
fix: revise interesting/uninteresting classification for AXNodes

johanbay:
Minor changes to the logic for determining if an AXNode is interesting.
We might want to consider skipping this logic altogether and simply rely on CDP's `ignored` property on [AXNodes](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/#type-AXNode)

### DIFF
--- a/src/common/Accessibility.ts
+++ b/src/common/Accessibility.ts
@@ -96,12 +96,12 @@ export interface SerializedAXNode {
  */
 export interface SnapshotOptions {
   /**
-   * Prune unintersting nodes from the tree.
+   * Prune uninteresting nodes from the tree.
    * @defaultValue true
    */
   interestingOnly?: boolean;
   /**
-   * Prune unintersting nodes from the tree.
+   * Root node to get the accessibility tree for
    * @defaultValue The root node of the entire page.
    */
   root?: ElementHandle;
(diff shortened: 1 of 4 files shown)


mathiasbynens (approved):

mathiasbynens:
> We might want to consider skipping this logic altogether and simply rely on CDP's `ignored` property on [AXNodes](https://chromedevtools.github.io/devtools-protocol/tot/Accessibility/#type-AXNode)

This sounds good. Until then, I like this smaller, incremental change, so LGTM.

src/common/Accessibility.ts
```diff
@@ -264,11 +266,7 @@ class AXNode {
   private _isPlainTextField(): boolean {
     if (this._richlyEditable) return false;
     if (this._editable) return true;
-    return (
-      this._role === 'textbox' ||
-      this._role === 'ComboBox' ||
```
hanselfmu, line 269: Out of curiosity, what was this 'ComboBox' role and why is it not relevant anymore?

src/common/Accessibility.ts
```diff
@@ -363,6 +362,7 @@ class AXNode {
   public isInteresting(insideControl: boolean): boolean {
     const role = this._role;
     if (role === 'Ignored' || this._hidden) return false;
+    if (this._ignored) return false;
```
hanselfmu: nit: maybe we could also combine this condition check with the check above, since the above check also contains a check for "being ignored or not". WDYT?

hanselfmu (approved):
LGTM, thanks!

src/common/Accessibility.ts
johanbay: I'm definitely not very knowledgeable in the area, but from searching around I concluded that there is no such role in the spec. There is a `combobox`, but I'm not sure we want to categorize those as plain text fields? At least our tests don't want that. I think we should remove the `ComboBox` for now as it is confusing.

src/common/Accessibility.ts
johanbay: yeah let's do that :)

Issue #6293, opened 2020-07-30
Getting pageBinding.get is not a function when I exposed imported module. 


comcomhanb:
* Puppeteer version:v1.18.1 
* Platform / OS version: mac 10.13.6
* URLs (if applicable):
* Node.js version: v10.15.0.

**What steps will reproduce the problem?**

With this code, I am getting the exact same error with https://github.com/puppeteer/puppeteer/issues/5586, "VM1571:2 Uncaught (in promise) TypeError: this._pageBindings.get(...) is not a function".

var finder = require('@medv/finder') 

        await page.exposeFunction('finder', finder);
        await page.evaluateOnNewDocument(()=> {    
            document.addEventListener("click", function callback(e){
                const selecotr = finder(e.target);
                console.log(selecotr)
            } , true /* capture */);
            
        });

**What is the expected result?**
Should able to call "finder." I imported the finder module but I think it fails to bind it. 

**What happens instead?**
 
throw ""VM1571:2 Uncaught (in promise) TypeError: this._pageBindings.get(...) is not a function" error

joshgrime:
t

Issue #5586, opened 2020-04-04
Imported functions not working


ErezShahaf:
**Tell us about your environment:**

* Puppeteer version: The latest version
* Platform / OS version: Visual studio code windows 10
* Node.js version: The latest version

**What steps will reproduce the problem?**

**Importing a long JS function from a different file:**

**And then trying to use that function on the page.**
```
await page.exposeFunction("startTheShow",startTheShow);
await page.evaluate(async function() {
   await startTheShow('info','hi','banana');
})
```

**What is the expected result?**
The imported function will run on the page

**What happens instead?**
```
Error: Evaluation failed: TypeError: this._pageBindings.get(...) is not a function
```

Do you guys have an idea how I can fix this?

ErezShahaf:
Hi, I realized that I first have to expose the function like this:
```
await page.exposeFunction("startTheShow",startTheShow);
await page.evaluate(async function() {
   await startTheShow('info','hi','banana');
})
```

But it will return this error:
`Error: Evaluation failed: TypeError: this._pageBindings.get(...) is not a function`
any idea?

ErezShahaf:
I have also tried this syntax : 
```
await page.exposeFunction("startTheShow",async function() {await startTheShow;});
await page.evaluate(async () => {
  await startTheShow('hi','hi','hi');
```


It does not throw an error any more, but it will not run the function on the page

And also this syntax: 

```
await page.exposeFunction("startTheShow",startTheShow);
await page.evaluate(() => {
  startTheShow('hi','hi','hi');
```
which brings the same outcome, no error but nothing executes on the page

jackfranklin:
Hey @ErezShahaf , are you running the latest published Puppeteer from npm or from this GitHub repo directly?

jackfranklin:
Going to close this, but please open a new issue if you're still seeing this issue.

comcomhanb:
I am using the latest published Puppeteer which is v1.18.1 and node v10.15.0. 
With this code, I am getting the exact same error, "VM1571:2 Uncaught (in promise) TypeError: this._pageBindings.get(...) is not a function". 

            await page.exposeFunction('finder', finder);
            await page.evaluateOnNewDocument(()=> {    
                document.addEventListener("click", function callback(e){
                    const selecotr = finder(e.target);
                    console.log(selecotr)
                } , true /* capture */);
                
            });

Catsvilles:
Can confirm same error, got no idea how to fix this

[...]

PR #6365, opened 2020-08-27
fix(page): fallback to default in exposeFunction when using imported module 

joshgrime:
I came across the same issue as [#5586](https://github.com/puppeteer/puppeteer/issues/5586) and [#6293](https://github.com/puppeteer/puppeteer/issues/6293) and decided to take a look.

The problem appears to be that if you pass `page.exposeFunction` an imported module which uses `export default`, then `this._pageBindings.get(name)` will return an object rather than a function as expected.

**myModule.js**
```
function parseData(options){
    console.log(options);
}

export default parseData;
```

**App.js**
```
var plugin = await import('./myModule.js');
var options = {/*     */};
await page.exposeFunction('myPlugin', plugin);
var data = await page.evaluate(async options => {
                var payload = await myPlugin(options);
                return payload;
 }, options);
```

**puppeteer/common/Page.js**
```
async _onBindingCalled(event) {
        const { name, seq, args } = JSON.parse(event.payload);
        let expression = null;
        try {
            const result = await this._pageBindings.get(name)(...args);
...
}
```

Inspecting `const result = await this._pageBindings.get(name)`, it is actually an object with a default property:

```
console.log(result);
// [Module] { default: [Function: parseData] }
```

I've added the following fallback to `exposeFunction`, which will use the 'default' if there's a function defined for it:

```
if (typeof puppeteerFunction !== 'function' &&
      typeof puppeteerFunction['default'] === 'function') {
      puppeteerFunction = puppeteerFunction['default'];
 }
```

**Alternative fix**

An alternative way to get around this with no changes is, if you are using a default export, to make sure you pass the default function rather than the whole module to `page.exposeFunction`. For example:

**App.js**
```
//From:
await page.exposeFunction('myPlugin', plugin);
//To:
await page.exposeFunction('myPlugin', plugin.default);

```

For me it's preferable to include the fallback. I think that this behaviour is more in-line with what you expect from default exported modules.


joshgrime:
@googlebot I signed it!

jackfranklin:
@joshgrime thanks for your PR and sorry about the delay in responding.

I'm 50/50 on this one to be honest - I'm not sure if actually asking users to pass `foo.default` in this case is more reasonable rather than hide it on the Puppeteer side. But I can also see the reasoning to have Puppeteer do it.

@mathiasbynens would be good to have your opinion here, WDYT?

joshgrime:
@jackfranklin thanks for the response. I understand the deliberation; I went through it a bit myself, hence adding the quick fix at the end. My thinking was that since I couldn't forsee any significant downsides to having the fallback option, and not having it could lead to errors/confusion as in the issues raised, it was worth adding it in.

Really interested to hear of any negative side-effects or reasoning I haven't considered, especially as this is my first contribution

src/common/Page.ts
```diff
@@ -1052,6 +1052,14 @@ export class Page extends EventEmitter {
       throw new Error(
         `Failed to add page binding with name ${name}: window['${name}'] already exists!`
       );
+
+    if (
+      typeof puppeteerFunction !== 'function' &&
+      typeof puppeteerFunction['default'] === 'function'
+    ) {
+      puppeteerFunction = puppeteerFunction['default'];
```
mathiasbynens:
```suggestion
      typeof puppeteerFunction.default === 'function'
    ) {
      puppeteerFunction = puppeteerFunction.default;
```

mathiasbynens (dismissed):
Thanks for the patch and the detailed description!

I'm ok with letting Puppeteer handle it, as this patch does. The current patch LGTM % a suggestion. Would you mind adding a test for this new behavior as well?

With those resolved, we can merge this.

src/common/Page.ts
joshgrime: Hey @mathiasbynens , thanks for your comments and suggestions.

I've added the test and is passing for me. Had to use expect error next line as it's expecting a function rather than an object:

```
    it('should fallback to default export when passed a module object', async () => {
      var moduleObject = {
        default: function(a, b) {
                  return a * b;
                }
              };
      // @ts-expect-error
      await page.exposeFunction('compute', moduleObject);
      const result = await page.evaluate(async function () {
        return await globalThis.compute(9, 4);
      });
      expect(result).toBe(36);
    });
```



The changes to dot syntax in Page.ts is throwing an error for me though, please let me know if you prefer to ignore the line or keep the [ ] syntax.

```
src/common/Page.ts:1058:32 - error TS2339: Property 'default' does not exist on type 'never'.

1058       typeof puppeteerFunction.default === 'function'
```

Quentin-M:
Howdy, any movement here? Thanks!

jschfflr:
@joshgrime I updated the change to address typing problems and added documentation. Let me know if this still looks good to you and I'm happy to get this landed!

jschfflr:
@googlebot I consent.

jschfflr (approved):

joshgrime:
Looks good to me @jschfflr, thanks for picking this up

jschfflr:
Happy to help. Let's wait for @mathiasbynens review to get this landed :)

PR #6374, opened 2020-08-31
Typo: Substitute AudioListener for AudioListner

guest271314:


jackfranklin:
Thanks for your PR, but this actually comes from the devtools-protocol package, which in turn comes from the protocol PDL file in Chromium. I've opened a CL (what Chromium calls a PR!) to fix it: https://crrev.com/c/2401043.

PR #6398, opened 2020-09-07
fix: merge frame url and frame url fragment for getting full url

bmlyr:
This Pr : 

- Retrieve full url by adding url fragment to url when trying to call url function from a frame.


bmlyr:
@googlebot I signed it!

bryanmlyr:
@googlebot I fixed it.

bmlyr:
@googlebot I fixed it.

mathiasbynens:
Thanks for the patch. Two requests:

- In the commit message, please explain what problem this is fixing.
- Please add a test

bmlyr:
> Thanks for the patch. Two requests:
> 
> * In the commit message, please explain what problem this is fixing.
> * Please add a test

Done 😄

test/assets/frames/one-frame-url-fragment.html
```diff
@@ -0,0 +1 @@
+<iframe src='./frame.html&test=fragment'></iframe>
```
mathiasbynens:
```suggestion
<iframe src='./frame.html&test=fragment'></iframe>

```

mathiasbynens (approved):

mathiasbynens:
@mjzffr FYI, the CI has been unhappy lately, with Firefox launch failing. Does this ring any bells?

mathiasbynens:
Ah, this is still #5673. Will post there

test/assets/frames/one-frame-url-fragment.html
campersau, line 1: Shouldn't this be `./frame.html?test=fragment` or `./frame.html#fragment` instead?
According to https://github.com/puppeteer/puppeteer/blob/main/new-docs/puppeteer.protocol.page.frame.urlfragment.md the fragment is `#fragment` and `?param=value` is often referred to as query.

mathiasbynens: @campersau That's a good point. Happy to accept a PR that changes these names!

PR #6399, opened 2020-09-08
chore: update documentation on rolling chromium

johanbay:


CONTRIBUTING.md
```diff
@@ -269,6 +269,18 @@ See [Debugging Tips](README.md#debugging-tips) in the readme.
 
 # For Project Maintainers
 
+## Rolling new Chromium version
+
+The following steps are needed to update the Chromium version.
+
+1. Find a suitable Chromium revision
+   Not all revisions has builds for all platforms, so we need to find one that does.
```
mathiasbynens:
```suggestion
   Not all revisions have builds for all platforms, so we need to find one that does.
```

CONTRIBUTING.md
```diff
@@ -269,6 +269,18 @@ See [Debugging Tips](README.md#debugging-tips) in the readme.
...
+   To do so, run `utils/check_availability.js -rb` to find the latest suitable beta Chromium revision (see `utils/check_availability.js -help` for more options).
+2. Update `src/revisions.ts` with the found revision number.
+3. Run `npm run ensure-correct-devtools-protocol-revision`.
+   If it fails, update `package.json` with the expected `devtools-protoocol` version.
+4. Run `npm run tsc` and `npm install` and ensure that all tests pass.
```
mathiasbynens:
```suggestion
1. Update `src/revisions.ts` with the found revision number.
1. Run `npm run ensure-correct-devtools-protocol-revision`.
   If it fails, update `package.json` with the expected `devtools-protocol` version.
1. Run `npm run tsc` and `npm install` and ensure that all tests pass. If a test fails, bisect the upstream cause of the failure, and either update the test expectations accordingly (if it was an intended change) or work around the changes in Puppeteer (if it’s not desirable to change Puppeteer’s observable behavior).
```

mathiasbynens (changes requested):

mathiasbynens:
Thanks for kicking this off! Left some suggestions, but LGTM with those.

mathiasbynens:
cc @sadym-chromium

CONTRIBUTING.md
mathiasbynens: We could also add a section on bisecting (I’d be happy to do this as a follow-up to this PR). TL;DR would be:

> Given a Node.js script that calls `process.exit(1)` for bad revisions, run this from the puppeteer repository’s root directory:
> 
> ```sh
> node utils/bisect.js --good 686378 --bad 706915 script.js
> ```

mathiasbynens (approved):

hanselfmu (approved):

CONTRIBUTING.md
johanbay: I think that would be great! I don't think it needs to be extensive either, but it would be nice if all the `utils/` scripts are mentioned in the documentation.

CONTRIBUTING.md
```diff
@@ -269,6 +269,18 @@ See [Debugging Tips](README.md#debugging-tips) in the readme.
...
+
+The following steps are needed to update the Chromium version.
+
+1. Find a suitable Chromium revision
+   Not all revisions have builds for all platforms, so we need to find one that does.
+   To do so, run `utils/check_availability.js -rb` to find the latest suitable beta Chromium revision (see `utils/check_availability.js -help` for more options).
+2. Update `src/revisions.ts` with the found revision number.
+3. Run `npm run ensure-correct-devtools-protocol-revision`.
+   If it fails, update `package.json` with the expected `devtools-protocol` version.
+4. Run `npm run tsc` and `npm install` and ensure that all tests pass. If a test fails, bisect the upstream cause of the failure, and either update the test expectations accordingly (if it was an intended change) or work around the changes in Puppeteer (if it’s not desirable to change Puppeteer’s observable behavior).
```
mathiasbynens: I had changed the numbering on purpose — in Markdown, it's better to always number with `1.` so that you can later update the lists without having to worry about updating all the other numbers (smaller diffs, easier maintenance). Please reconsider?

johanbay: Oh, I did not know of this magic :) I'll fix. Thanks!

mathiasbynens (approved):

PR #6400, opened 2020-09-08
chore: add channel parameter in check_availability roll mode

hanselfmu:
This PR adds an optional parameter in `check_availability.js -r` mode (roll mode) to specify the channel (stable | dev | beta) to check the latest revision from.

### DIFF
--- a/utils/check_availability.js
+++ b/utils/check_availability.js
@@ -56,16 +56,20 @@ class Table {
 const helpMessage = `
 This script checks availability of prebuilt Chromium snapshots.
 
-Usage: node check_availability.js [<options>] [<browser version(s)>]
+Usage
 
-options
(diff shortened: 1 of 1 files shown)


utils/check_availability.js
```diff
@@ -97,7 +104,7 @@ function main() {
       case 'f':
         break;
       case 'r':
-        checkRollCandidate();
+        checkRollCandidate(args[1]);
```
mathiasbynens, line 107: Let's assert that `args[1]` is one of the supported values here, and error out otherwise

mathiasbynens:
```js
const supportedChannels = new Map(['stable', 'beta', 'dev']);
```

hanselfmu:
Closing this in favor of #6399 .

PR #6402, opened 2020-09-08
doc(contributing): add info on bisecting upstream changes

mathiasbynens:


CONTRIBUTING.md
```diff
@@ -281,6 +281,14 @@ The following steps are needed to update the Chromium version.
    If it fails, update `package.json` with the expected `devtools-protocol` version.
 1. Run `npm run tsc` and `npm install` and ensure that all tests pass. If a test fails, bisect the upstream cause of the failure, and either update the test expectations accordingly (if it was an intended change) or work around the changes in Puppeteer (if it’s not desirable to change Puppeteer’s observable behavior).
```
johanbay: Should we maybe add a link to the new section here?
```suggestion
1. Run `npm run tsc` and `npm install` and ensure that all tests pass. If a test fails, [bisect](#bisecting-upstream-changes) the upstream cause of the failure, and either update the test expectations accordingly (if it was an intended change) or work around the changes in Puppeteer (if it’s not desirable to change Puppeteer’s observable behavior).
```

johanbay (approved):
Looks good to me 🙂

CONTRIBUTING.md
mathiasbynens: Good call!

PR #6410, opened 2020-09-10
feat(page): emulate idle state

sadym-chromium:
* https://bugs.chromium.org/p/chromium/issues/detail?id=1090802
* added `page.emulateIdleState(...)` allowing emulate or remove emulation of the idle state;
* added test `emulate idle` -> `remove emulation` -> `emulate idle` -> `remove emulation`;
* added launch argument `--enable-blink-features=IdleDetection` to turn IdleDetection on.


src/common/Page.ts
```diff
@@ -1443,6 +1443,41 @@ export class Page extends EventEmitter {
...
+   *
+   * @example
+   * ```js
+   * await page.setIdleOverride({isUserActive: true, isScreenUnlocked: false});
+   * ```
+   *
+   * @param isUserActive Mock isUserActive
+   * @param isScreenUnlocked Mock isScreenUnlocked
+   */
+  async setIdleOverride(
```
mathiasbynens: What's the rationale for the current name? How would you feel about something like `emulateIdleState` (seems closer to existing Puppeteer APIs)? As an end user, this seems nice:

```js
page.emulateIdleState({ userActive: true, screenUnlocked: true });
```

src/common/Page.ts
```diff
@@ -1443,6 +1443,41 @@ export class Page extends EventEmitter {
...
+   * Clears Idle state overrides.
+   *
+   * @example
+   * ```js
+   * await page.setIdleOverride({isUserActive: true, isScreenUnlocked: false});
+   * // do some checks here
+   * await page.clearIdleOverride();
+   * ```
+   */
+  async clearIdleOverride(): Promise<void> {
```
mathiasbynens: For other emulation APIs we tend to not have a separate `clear` API in Puppeteer, and instead we clear when the same method is called with an empty string or `null`/`undefined` or similar, depending on the situation. In this case, we could go with `page.emulateIdleState()` (i.e. `undefined` clears all overrides). WDYT?

test/assets/idle-detector.html
```diff
@@ -0,0 +1,23 @@
...
+    });
+    idleDetector.addEventListener('change', () => {
+      const userState = idleDetector.userState;
+      const screenState = idleDetector.screenState;
+      setState(`Idle state: ${userState}, ${screenState}.`);
+    });
+    idleDetector.start();
+  }
+  main();
+</script>
```
mathiasbynens: nit: trailing newline (you can probably configure your editor to automatically take care of this)

test/idle_override.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+  after(async () => {
+    const { page } = getTestState();
+    await page.clearIdleOverride();
+  });
+
+  async function getIdleState() {
+    const { page } = getTestState();
+
+    const stateElement = await page.$('#state');
+    return await page.evaluate((e: HTMLElement) => {
```
mathiasbynens: nit: e -> element

src/node/Launcher.ts
```diff
@@ -190,6 +190,7 @@ class ChromeLauncher implements ProductLauncher {
       '--enable-automation',
       '--password-store=basic',
       '--use-mock-keychain',
+      '--enable-blink-features=IdleDetection',
```
mathiasbynens, line 195: Let's add a TODO here reminding us to remove it once the feature ships by default.

mathiasbynens (dismissed):
The implementation LGTM. I left some comments mostly wanting to discuss the Puppeteer-level API. There is no need to match the CDP API directly — we should try to provide the most user-friendly API possible (which in most cases is somewhat different from the CDP one).

src/common/Page.ts
OrKoN: I'd also suggest to use the object instead of two boolean arguments.

sadym-chromium: done

src/common/Page.ts
sadym-chromium: done

test/assets/idle-detector.html
sadym-chromium: done

src/node/Launcher.ts
sadym-chromium: done

test/idle_override.spec.ts
sadym-chromium: done

OrKoN (approved):

src/node/Launcher.ts
```diff
@@ -190,6 +190,9 @@ class ChromeLauncher implements ProductLauncher {
...
+      // TODO sadym: remove '--enable-blink-features=IdleDetection'
+      // after IdleDetection is turned on by default
```
mathiasbynens: nit: TODOs go before the line they're commenting on

```suggestion
      // TODO(sadym): remove '--enable-blink-features=IdleDetection'
      // once IdleDetection is turned on by default.
      '--enable-blink-features=IdleDetection',
```

test/idle_override.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+    await page.emulateIdleState({
+      isUserActive: false,
+      isScreenUnlocked: true,
+    });
+    await verifyState('Idle state: idle, unlocked.');
+
+    await page.emulateIdleState();
+    // Remove Idle emulation and verify IdleDetector is in initial state.
+    await verifyState(initialState);
```
mathiasbynens, line 92: Let's also test this in between the other tests, so that we test what happens after disabling the emulation and then re-enabling it. You can just copy-paste the two lines.

mathiasbynens (approved):
LGTM % some final requests

test/idle_override.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+      isUserActive: false,
+      isScreenUnlocked: true,
+    });
+    await verifyState('Idle state: idle, unlocked.');
+
+    // Remove Idle emulation and verify IdleDetector is in initial state.
+    await page.emulateIdleState();
+    await verifyState(initialState);
+
+    // Emulation idle state again after removing emulation
```
mathiasbynens:
```suggestion
    // Emulate idle state again after removing emulation.
```

src/node/Launcher.ts
```diff
@@ -190,6 +190,9 @@ class ChromeLauncher implements ProductLauncher {
       '--enable-automation',
       '--password-store=basic',
       '--use-mock-keychain',
+      // TODO sadym: remove '--enable-blink-features=IdleDetection'
+      // after IdleDetection is turned on by default
+      '--enable-blink-features=IdleDetection',
```
mathiasbynens:
```suggestion
      // TODO(sadym): remove '--enable-blink-features=IdleDetection'
      // once IdleDetection is turned on by default.
      '--enable-blink-features=IdleDetection',
```

test/idle_override.spec.ts
sadym-chromium: done

Issue #6307, opened 2020-08-05
Proposal: Query by accessible name


johanbay:
## Motivation
Currently Puppeteer supports querying the DOM using CSS selectors. While this provides a versatile way to select DOM elements by for example id, tag, or class-attributes, these properties are often subject to change or even generated at build-time.
Alternatively, we could support querying the accessibility tree for accessible names.
This could provide the following benefits:
- Make selectors in test scripts more resilient to source code changes.
- Make test scripts more readable since the accessible names should carry semantic context.
- Motivate good practices for assigning accessibility properties to elements.

## Current state of affairs
Querying by accessible names is currently possible by implementing a custom query handler to traverse the DOM and check accessible names along the way. This is what [puppeteer/recorder](https://github.com/puppeteer/recorder) currently does. Since there is no direct way to get the computed accessible name and role from the browser-side, one has to compute those on demand which is not a trivial task.
One can instead access and traverse the accessibility tree through Puppeteer where the computed accessible names and roles are available. The catch then is that Puppeteer only gives access to a static snapshot of the accessibility tree. 
To support interaction, we need a mapping to live DOM nodes.

## Proposal
Extend Puppeteer to support querying the accessibility tree with support for interaction.
The steps for achieving this could be:
- [x] Implement CDP-level support for querying the accessibility tree.
Landed in https://github.com/chromium/chromium/commit/d557b1cd92b3c47dae01aaaddd7c6be0ceb6cb94
- [x] Restructure such that query handlers can query on the Puppeteer-level (as opposed to querying in the page context) 
Done in #6437.
- [x] Implement atlernative query handler for a11y-based querying
Done in #6459  and #6472.

WickyNilliams:
Hey, sorry to jump on a closed issue - i'm wondering if there are any docs available for this feature? The [introductory blog post](https://developers.google.com/web/updates/2020/11/puppetaria) mostly explains about the internal implementation details, and doesn't explain the feature set. 

Love the idea here, I'm eager to put it to use in my test suite!

mathiasbynens:
That's a good point. While we do have docs on [how to register custom query handlers](https://github.com/puppeteer/puppeteer/blob/v7.0.1/docs/api.md#puppeteerregistercustomqueryhandlername-queryhandler), we don't really document the built-in ones. Let’s use a new issue for this: #6822

mathiasbynens:
Let’s use a new issue for this: #6822

WickyNilliams:
[...]

PR #6426, opened 2020-09-17
feat(a11y-query): split waitFor logic for selectors and xpath

johanbay:
The logic for `waitForXPath` and `waitForSelector` is currently very tightly coupled. This PR tries to untangle that relationship. 
This is the first step towards introducing built-in query handlers that are not executed in the page context (#6307).

### DIFF
--- a/src/common/DOMWorld.ts
+++ b/src/common/DOMWorld.ts
@@ -37,12 +37,17 @@ import {
 } from './EvalTypes.js';
 import { isNode } from '../environment.js';
 
-// This predicateQueryHandler is declared here so that TypeScript knows about it
-// when it is used in the predicate function below.
+// predicateQueryHandler and checkWaitForOptions are declared here so that
+// TypeScript knows about them when used in the predicate function below.
 declare const predicateQueryHandler: (
   element: Element | Document,
   selector: string
 ) => Element | Element[] | NodeListOf<Element>;
+declare const checkWaitForOptions: (
+  node: Node,
+  waitForVisible: boolean,
+  waitForHidden: boolean
+) => Element | null | boolean;
 
 /**
  * @public
@@ -457,48 +462,9 @@ export class DOMWorld {
     await handle.dispose();
   }
 
-  waitForSelector(
+  async waitForSelector(
     selector: string,
     options: WaitForSelectorOptions
-  ): Promise<ElementHandle | null> {
-    return this._waitForSelectorOrXPath(selector, false, options);
-  }
-
-  waitForXPath(
-    xpath: string,
-    options: WaitForSelectorOptions
-  ): Promise<ElementHandle | null> {
-    return this._waitForSelectorOrXPath(xpath, true, options);
-  }
-
-  waitForFunction(
-    pageFunction: Function | string,
-    options: { polling?: string | number; timeout?: number } = {},
-    ...args: SerializableOrJSHandle[]
-  ): Promise<JSHandle> {
-    const {
-      polling = 'raf',
-      timeout = this._timeoutSettings.timeout(),
-    } = options;
-    return new WaitTask(
-      this,
-      pageFunction,
-      undefined,
-      'function',
-      polling,
-      timeout,
-      ...args
-    ).promise;
-  }
-
-  async title(): Promise<string> {
-    return this.evaluate(() => document.title);
-  }
-
-  private async _waitForSelectorOrXPath(
-    selectorOrXPath: string,
-    isXPath: boolean,
-    options: WaitForSelectorOptions = {}
   ): Promise<ElementHandle | null> {
     const {
       visible: waitForVisible = false,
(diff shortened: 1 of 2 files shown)


NuruddeenSalis (commented):
Avoid using null, any number will better

OrKoN (approved):
LGTM but I'd let someone else to review as well

src/common/DOMWorld.ts
```diff
@@ -568,6 +587,29 @@ export class DOMWorld {
...
+      timeout = this._timeoutSettings.timeout(),
+    } = options;
+    return new WaitTask(
+      this,
+      pageFunction,
+      'function',
+      polling,
+      timeout,
+      ...args
+    ).promise;
```
jackfranklin, line 609: one day (not in this PR!) I'm going to refactor `WaitTask` to take an options bag, it's so hard to figure out what all of these are for!

jackfranklin (approved):
LGTM - are there any additional tests we can add whilst we're tidying up this area? And could you do a cursory check through the unit tests just to make sure that this function is well tested?

src/common/DOMWorld.ts
johanbay: I agree. At least we get rid of the `predicateQueryHandlerBody` argument with this change 🙂

johanbay:
> LGTM - are there any additional tests we can add whilst we're tidying up this area? And could you do a cursory check through the unit tests just to make sure that this function is well tested?

I think this is actually rather well-tested. [test/waittask.spec.ts](https://github.com/puppeteer/puppeteer/blob/main/test/waittask.spec.ts) tests both `waitForSelector` and `waitForXPath` quite intensively for things such as that it re-inserts the waitFor on navigations, handles various combinations of the `waitForOptions`, handles mutliple frames and error handling when frames are detached.

hanselfmu (approved):

src/common/DOMWorld.ts
```diff
@@ -37,12 +37,17 @@ import {
 } from './EvalTypes.js';
 import { isNode } from '../environment.js';
 
-// This predicateQueryHandler is declared here so that TypeScript knows about it
-// when it is used in the predicate function below.
+// predicateQueryHandler and checkWaitForOptions is declared here so that
```
mathiasbynens:
```suggestion
// predicateQueryHandler and checkWaitForOptions are declared here so that
```

src/common/DOMWorld.ts
```diff
@@ -37,12 +37,17 @@ import {
...
+// TypeScript knows about it when it is used in the predicate function below.
```
mathiasbynens:
```suggestion
// TypeScript knows about them when used in the predicate function below.
```

src/common/DOMWorld.ts
```diff
@@ -457,70 +462,82 @@ export class DOMWorld {
...
-  ): Promise<JSHandle> {
     const {
-      polling = 'raf',
+      visible: waitForVisible = false,
+      hidden: waitForHidden = false,
       timeout = this._timeoutSettings.timeout(),
     } = options;
-    return new WaitTask(
+    const polling = waitForVisible || waitForHidden ? 'raf' : 'mutation';
+    const title = `selector "${selector}"${
```
mathiasbynens: Let's wrap the select in backticks instead of double quotes, since it's more likely to contain double quotes (e.g. attribute selectors) than backticks:

```suggestion
    const title = `selector \`${selector}\`${
```

src/common/DOMWorld.ts
```diff
@@ -457,70 +462,82 @@ export class DOMWorld {
...
-  }
-
-  async title(): Promise<string> {
-    return this.evaluate(() => document.title);
+      updatedSelector,
+      waitForVisible,
+      waitForHidden
+    );
+    const handle = await waitTask.promise;
+    if (!handle.asElement()) {
```
mathiasbynens:
```js
const someBetterNameHere = handle.asElement();
if (!someBetterNameHere) { ... }
return someBetterNameHere;
```

src/common/DOMWorld.ts
```diff
@@ -457,70 +462,82 @@ export class DOMWorld {
...
       timeout = this._timeoutSettings.timeout(),
     } = options;
     const polling = waitForVisible || waitForHidden ? 'raf' : 'mutation';
-    const title = `${isXPath ? 'XPath' : 'selector'} "${selectorOrXPath}"${
-      waitForHidden ? ' to be hidden' : ''
-    }`;
-    const { updatedSelector, queryHandler } = getQueryHandlerAndSelector(
-      selectorOrXPath
-    );
+    const title = `XPath "${xpath}"${waitForHidden ? ' to be hidden' : ''}`;
```
mathiasbynens: Same here with the backticks, WDYT?

mathiasbynens (changes requested):

src/common/DOMWorld.ts
johanbay: Good idea!

src/common/DOMWorld.ts
johanbay: Good idea, done.

mathiasbynens (approved):

PR #6437, opened 2020-09-21
feat(a11y-query): differentiate between user-defined and internal query handlers

johanbay:
This PR changes the internal representation of query handlers to contain Puppeteer-level code instead of page functions.
The interface `CustomQueryHandler` is introduced for user-defined query handlers. When a `CustomQueryHandler` is registered using  `registerCustomQueryHandler` a corresponding Puppeteer-level handler is created through `makeQueryHandler` by wrapping the page functions as appropriate.
The internal query handlers (defined by the interface `QueryHandler`) contain two new functions: `waitFor` and `queryAllArray`.
- `waitFor` allows page-based handlers to make use of the `WaitTask`-backed implementation in `DOMWorld`, whereas purely Puppeteer-based handlers can define an alternative approach instead.
- `queryAllArray` is similar to `queryAll` but with a slightly different interface; it returns a `JSHandle` to an array with the results as opposed to an array of `ElementHandle`. It is used by `$$eval`. 

After this change, we can introduce built-in query handlers that are not executed in the page context (#6307).

### DIFF
--- a/src/common/JSHandle.ts
+++ b/src/common/JSHandle.ts
@@ -777,15 +777,7 @@ export class ElementHandle<
     const { updatedSelector, queryHandler } = getQueryHandlerAndSelector(
       selector
     );
-
-    const handle = await this.evaluateHandle(
-      queryHandler.queryOne,
-      updatedSelector
-    );
-    const element = handle.asElement();
-    if (element) return element;
-    await handle.dispose();
-    return null;
+    return queryHandler.queryOne(this, updatedSelector);
   }
 
   /**
@@ -796,19 +788,7 @@ export class ElementHandle<
     const { updatedSelector, queryHandler } = getQueryHandlerAndSelector(
       selector
     );
-
-    const handles = await this.evaluateHandle(
-      queryHandler.queryAll,
-      updatedSelector
-    );
-    const properties = await handles.getProperties();
-    await handles.dispose();
-    const result = [];
-    for (const property of properties.values()) {
-      const elementHandle = property.asElement();
-      if (elementHandle) result.push(elementHandle);
-    }
-    return result;
+    return queryHandler.queryAll(this, updatedSelector);
   }
 
   /**
@@ -892,15 +872,7 @@ export class ElementHandle<
     const { updatedSelector, queryHandler } = getQueryHandlerAndSelector(
       selector
     );
-    const queryHandlerToArray = Function(
-      'element',
-      'selector',
-      `return Array.from((${queryHandler.queryAll})(element, selector));`
-    ) as (...args: unknown[]) => unknown;
-    const arrayHandle = await this.evaluateHandle(
-      queryHandlerToArray,
-      updatedSelector
-    );
+    const arrayHandle = await queryHandler.queryAllArray(this, updatedSelector);
     const result = await arrayHandle.evaluate<
       (
         elements: Element[],
--- a/src/common/QueryHandler.ts
+++ b/src/common/QueryHandler.ts
@@ -14,7 +14,33 @@
  * limitations under the License.
  */
 
-export interface QueryHandler {
+import { WaitForSelectorOptions, DOMWorld } from './DOMWorld.js';
+import { ElementHandle, JSHandle } from './JSHandle.js';
+
+/**
+ * @internal
+ */
+interface InternalQueryHandler {
+  queryOne?: (
+    element: ElementHandle,
+    selector: string
+  ) => Promise<ElementHandle | null>;
+  waitFor?: (
+    domWorld: DOMWorld,
+    selector: string,
+    options: WaitForSelectorOptions
+  ) => Promise<ElementHandle | null>;
+  queryAll?: (
+    element: ElementHandle,
+    selector: string
+  ) => Promise<ElementHandle[]>;
+  queryAllArray?: (
+    element: ElementHandle,
+    selector: string
+  ) => Promise<JSHandle>;
+}
+
+export interface CustomQueryHandler {
   queryOne?: (element: Element | Document, selector: string) => Element | null;
   queryAll?: (
     element: Element | Document,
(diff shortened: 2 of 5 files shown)


src/common/QueryHandler.ts
```diff
@@ -14,62 +14,137 @@
...
+export interface CustomQueryHandler {
   queryOne?: (element: Element | Document, selector: string) => Element | null;
   queryAll?: (
     element: Element | Document,
     selector: string
   ) => Element[] | NodeListOf<Element>;
 }
 
-const _customQueryHandlers = new Map<string, QueryHandler>();
+function mkQueryHandler(handler: CustomQueryHandler): QueryHandler {
```
jackfranklin: is `mk` here short for `make` ? Could we name it that? Might as well be explicit for the sake of some extra chars.

src/common/QueryHandler.ts
```diff
@@ -14,62 +14,137 @@
...
  */
 
-export interface QueryHandler {
+import { WaitForSelectorOptions, DOMWorld } from './DOMWorld.js';
+import { ElementHandle, JSHandle } from './JSHandle.js';
+
+/**
+ * @internal
+ */
+interface QueryHandler {
```
jackfranklin: what do you think about calling this type `InternalQueryHandler` to really make it clear that it's not user facing?

src/common/QueryHandler.ts
```diff
@@ -14,62 +14,137 @@
...
+      );
+      const arrayHandle = await resultHandle.evaluateHandle(
+        (res: Element[] | NodeListOf<Element>) => Array.from(res)
+      );
+      return arrayHandle;
+    };
+  }
+
+  return internalHandler;
+}
```
jackfranklin, line 94: are these all tested in existing unit tests? If not, is it worth adding more? This feels like a fair chunk of code that should have some tests backing it.

jackfranklin:
> queryAllArray is similar to queryAll but with a slightly different interface; it returns a JSHandle to an array with the results as opposed to an array of ElementHandle. It is used by $$eval.

Just to check, this isn't a breaking change? As in, to users calling `$$eval`, the return type hasn't changed? I don't think so but want to call it out as we've been bitten by subtle breaking changes in the past.

src/common/QueryHandler.ts
```diff
@@ -14,62 +14,137 @@
...
-  _customQueryHandlers.delete(name);
+  if (_queryHandlers.has(name)) {
+    _queryHandlers.delete(name);
+  }
 }
 
-export function customQueryHandlers(): Map<string, QueryHandler> {
-  return _customQueryHandlers;
+export function customQueryHandlerNames(): string[] {
+  return Array.from(_queryHandlers.keys());
```
mathiasbynens: This is debatable, but I weakly prefer array spread syntax here:

```js
[..._queryHandlers.keys()]
```

WDYT?

src/common/JSHandle.ts
```diff
@@ -892,15 +872,7 @@ export class ElementHandle<
...
-      `return Array.from((${queryHandler.queryAll})(element, selector));`
-    ) as (...args: unknown[]) => unknown;
-    const arrayHandle = await this.evaluateHandle(
-      queryHandlerToArray,
-      updatedSelector
-    );
+    const arrayHandle = await queryHandler.queryAllArray(this, updatedSelector);
     const result = await arrayHandle.evaluate<
       (
         elements: Element[],
```
mathiasbynens, line 878: This is a comment on **all** the changes in `src/common/JSHandle.ts`: wow, this is so much nicer! Awesome refactoring.

mathiasbynens (approved):
Looking great so far! Can we add some additional tests for the different scenarios you mention (e.g. `queryAllArray`), or are they sufficiently covered by the existing tests?

src/common/QueryHandler.ts
johanbay: Sure! considering that they are internal, I suppose we can always change our minds about this as well.

src/common/QueryHandler.ts
johanbay: It is. I'll rename to `make` :)

johanbay:
> Just to check, this isn't a breaking change? As in, to users calling `$$eval`, the return type hasn't changed? I don't think so but want to call it out as we've been bitten by subtle breaking changes in the past.

It doesn't change the return type, and I generally don't think it's a breakning change. 
I'll try to highligh the difference in the code to make the actual change clear.
Before, we would do this (in `$$eval`) to get a JSHandle for the result array:
```
const queryHandlerToArray = Function(
  "element",
  "selector",
  `return Array.from((${handler.queryAll})(element, selector));`
) as (...args: unknown[]) => unknown;
const arrayHandle = await this.evaluateHandle(
  queryHandlerToArray,
  updatedSelector
);
```
and now (in `queryAllArray`) we instead do
```
const resultHandle = await element.evaluateHandle(handler.queryAll, selector);
const arrayHandle = await resultHandle.evaluateHandle(
  (res: Element[] | NodeListOf<Element>) => Array.from(res)
);
```
So we simplify the code a bit and get rid of a `Function` invocation. The price is that we do two CDP calls through `evaluateHandle` instead of one. If you prefer, we can also just use the previous approach in `queryAllArray` instead?

src/common/QueryHandler.ts
johanbay: I keep forgetting that this syntax exist. We'll do that instead :)

src/common/QueryHandler.ts
johanbay: They are tested through all the `$`, `$$`, `$eval`, and `$$eval` tests in `queryselector.spec.ts` (and `elementhandle.spec.ts`). The coverage seems quite good, but I'll add a few more for `$$eval` to make sure we cover every codepath here.
We primarily test the default handler, so I'll add a few tests for a custom handler that returns `Element[]` (as opposed to the default which returns `NodeListOf<Element>`).

test/queryselector.spec.ts
```diff
@@ -312,4 +359,100 @@ describe('querySelector', function () {
       expect(second).toEqual([]);
     });
   });
+
```
johanbay, line 365: The tests are copied from above. What do you think about this approach? We could avoid this by setting up a loop and try each test for both handlers, but I feel like that will hurt readability of the tests.
We could also cut down the number of tests for this alternative interface.

test/queryselector.spec.ts
```diff
@@ -312,4 +359,100 @@ describe('querySelector', function () {
...
+  // this is the same tests for `$$eval` and `$$` as above, but with a handler
```
mathiasbynens: nit: comments should be sentences

```suggestion
  // This is the same tests for `$$eval` and `$$` as above, but with a handler
```

test/queryselector.spec.ts
mathiasbynens: Not a fan of `for` loops around tests — they make it harder to debug. I like the current approach. Consider adding a comment to the tests at the top to indicate that they’re copied below? (You already added a comment in the new place.)

mathiasbynens (approved):

test/queryselector.spec.ts
```diff
@@ -312,4 +359,100 @@ describe('querySelector', function () {
...
+  // that returns an array instead of a list of nodes.
+  describe('Custom query handler QueryAll can return Element[]', function () {
```
mathiasbynens: The string passed to `describe()` is generally the name of the API that's being tested

johanbay: The public facing API that we are testing here is `registerCustomQueryHandler`, but that seems quite bit more general than what we are actually testing. I'll put `describe('QueryAll',...)` instead. WDYT?

test/queryselector.spec.ts
johanbay: Done.

test/queryselector.spec.ts
johanbay: I agree. I'll add a comment above as well. Thanks!

test/queryselector.spec.ts
mathiasbynens: Yeah, names of internal APIs are fine if they are what's really being tested. Your suggestion SGTM!

test/queryselector.spec.ts
```diff
@@ -312,4 +362,100 @@ describe('querySelector', function () {
...
+          ) + Number(div.textContent),
+        divHandle
+      );
+      expect(sum).toBe(8);
+    });
+    it('$$eval should handle many elements', async () => {
+      const { page } = getTestState();
+      await page.evaluate(
+        `
+        for (var i = 0; i <= 1000; i++) {
```
mathiasbynens, line 448:
```suggestion
        for (let i = 0; i <= 1000; i++) {
```

test/queryselector.spec.ts
```diff
@@ -77,6 +81,52 @@ describe('querySelector', function () {
...
+          ) + Number(div.textContent),
+        divHandle
+      );
+      expect(sum).toBe(8);
+    });
+    it('should handle many elements', async () => {
+      const { page } = getTestState();
+      await page.evaluate(
+        `
+        for (var i = 0; i <= 1000; i++) {
```
mathiasbynens, line 118:
```suggestion
        for (let i = 0; i <= 1000; i++) {
```

mathiasbynens (commented):
Still LGTM 👍

test/queryselector.spec.ts
johanbay: Great! Done.

jackfranklin:
Nice work, agree with Mathias that the refactoring makes this much nicer. LGTM 👍

PR #6438, opened 2020-09-22
feat(permissions): Add idle-detection permission

sadym-chromium:
Added `idle-detection` permission needed to test IdleDetector:
* [CL 2417779 idle-detection: Implement requestPermission() method](https://chromium-review.googlesource.com/c/chromium/src/+/2417779)


johanbay (approved):
Looks good to me!

PR #6441, opened 2020-09-23
fix: add mime dependency back to fix build

sadym-chromium:
* Added back dependency `"mime": "^2.0.3",` to fix build;
* Previous builds passed because of the node package caching.


mathiasbynens (approved):
LGTM, but please mention in the commit message that it's now a `devDependency` instead of a `dependency`

PR #6445, opened 2020-09-25
feat(console): expose stack trace for console messages

hashseed:


src/common/ConsoleMessage.ts
```diff
@@ -110,6 +110,13 @@ export class ConsoleMessage {
    * @returns The location of the console message.
    */
   location(): ConsoleMessageLocation {
-    return this._location;
+    return this._stackTraceLocations[0] ?? {};
```
mathiasbynens: We [support](https://github.com/puppeteer/puppeteer#usage) the latest maintenance Node.js LTS, so we cannot use nullish coalescing syntax in the compiled JS yet. Let's double-check that the generated output works in Node.js v10. cc @jackfranklin

mathiasbynens (approved):

src/common/ConsoleMessage.ts
jackfranklin: If CI passes we should be good as that compiles and runs on Node 10. I think our ts-config is set up to compile this down correctly but we will see!

jackfranklin (commented):
I think you'll also need to do `npm run generate-docs` to regenerate the TSDoc output (which is the new docs we are going to migrate to in time).

src/common/ConsoleMessage.ts
hashseed: the build artifacts in `lib/` still has `??`. Let me rephrase this in outdated JS.

jackfranklin: Looks like TypeScript is not compiling it away. I can look into fixing it or I'd rewrite to use a ternary (or similar) to avoid this for now, to be honest.

hashseed:
I ran `npm run generate-docs` and replaced `??` with a ternary!

src/common/ConsoleMessage.ts
mathiasbynens: Longer-term it might make sense to fix our tsconfig to align with the maintenance Node.js LTS version, so we don't have to worry about this in the future.

mathiasbynens (approved):

PR #6453, opened 2020-09-29
feat(a11y-query): add aria query handler

johanbay:
This PR adds an alternative built-in handler `aria/` that allows querying by accessible name and role.
This draft PR is mostly for reference to showcase the full implementation.
I will split this into 3 PRs:
- [ ] Roll Chromium revision 809590.
- [ ] Introduce aria handler with queryOne and queryAll support (but no waitFor).
- [ ] Add waitFor support to the aria query handler.

### DIFF
--- a/src/common/AriaQueryHandler.ts
+++ b/src/common/AriaQueryHandler.ts
@@ -1,0 +1,127 @@
+/**
+ * Copyright 2020 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { InternalQueryHandler } from './QueryHandler.js';
+import { ElementHandle } from './JSHandle.js';
+import { Protocol } from 'devtools-protocol';
+import { CDPSession } from './Connection.js';
+import { DOMWorld } from './DOMWorld.js';
+
+async function queryAXTree(
+  client: CDPSession,
+  element: ElementHandle,
+  accessibleName?: string,
+  role?: string
+): Promise<Protocol.Accessibility.AXNode[]> {
+  // @ts-ignore
+  const { nodes } = await client.send(
+    // @ts-ignore
+    'Accessibility.queryAXTree',
+    {
+      objectId: element._remoteObject.objectId,
+      accessibleName,
+      role,
+    }
+  );
+  // @ts-ignore
+  const filteredNodes: Protocol.Accessibility.AXNode[] = nodes.filter(
+    (node: Protocol.Accessibility.AXNode) => node.role.value !== 'text'
+  );
+  return filteredNodes;
+}
+
+/*
+ * The aria selectors are on the form '<computed name>&<computed role>'.
+ * The following examples showcase how the syntax works wrt. querying:
+ * - 'title&heading' queries for elements with computed name 'title' and role 'heading'.
+ * - '&img' queries for elements with role 'img' and any name.
+ * - 'label' queries for elements with name 'label' and any role.
+ */
+function parseAriaSelector(selector: string): { name?: string; role?: string } {
+  const s = selector.split('&');
+  const name = s[0] || undefined;
+  const role = s.length > 1 ? s[1] : undefined;
+  return { name, role };
+}
+
+const queryOne = async (element: ElementHandle, selector: string) => {
+  const exeCtx = element.executionContext();
+  const { name, role } = parseAriaSelector(selector);
+  const res = await queryAXTree(exeCtx._client, element, name, role);
+  if (res.length < 1) {
+    return null;
+  }
+  const handle = await exeCtx._adoptBackendNodeId(res[0].backendDOMNodeId);
+  return handle;
+};
+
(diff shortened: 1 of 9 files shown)


src/common/AriaQueryHandler.ts
```diff
@@ -0,0 +1,133 @@
...
+ */
+
+import { InternalQueryHandler } from './QueryHandler.js';
+import { ExecutionContext } from './ExecutionContext.js';
+import { ElementHandle } from './JSHandle.js';
+import { Protocol } from 'devtools-protocol';
+import { CDPSession } from './Connection.js';
+import { DOMWorld } from './DOMWorld.js';
+
+function backendNodeIdtoHandle(
```
mathiasbynens: We could get rid of this and just call `_adoptBackendNodeId` directly — my gut feeling is that having the extra helper doesn't seem to really make things more readable. WDYT?

src/common/AriaQueryHandler.ts
```diff
@@ -0,0 +1,133 @@
...
+    }
+  );
+  // @ts-ignore
+  const filteredNodes: Protocol.Accessibility.AXNode[] = nodes.filter(
+    (node: Protocol.Accessibility.AXNode) => node.role.value !== 'text'
+  );
+  return filteredNodes;
+}
+
+function parseAriaSelector(selector: string): { name?: string; role?: string } {
```
mathiasbynens, line 53: To make this easier to understand quickly, let's leave a comment giving one (or a few) example input(s).

src/common/helper.ts
```diff
@@ -182,6 +182,115 @@ function evaluationString(fun: Function | string, ...args: unknown[]): string {
...
+  function addPageBinding(type: string, bindingName: string): void {
+    /* Cast window to any here as we're about to add properties to it
+     * via win[bindingName] which TypeScript doesn't like.
+     */
+    const win = window as any;
+    const binding = win[bindingName];
+
+    win[bindingName] = (...args: unknown[]): Promise<unknown> => {
+      const me = window[bindingName];
+      let callbacks = me['callbacks'];
```
mathiasbynens: In general, use dot notation except when you need dynamic property names

`me['callbacks']` -> `me.callbacks`

(and so on elsewhere in this file)

test/ariaqueryhandler.spec.ts
```diff
@@ -0,0 +1,526 @@
...
+
+      await page.goto(server.EMPTY_PAGE);
+      const frame = page.mainFrame();
+      const watchdog = frame.waitForSelector('aria/&heading');
+      await frame.evaluate(addElement, 'br');
+      await frame.evaluate(addElement, 'h1');
+      const eHandle = await watchdog;
+      const tagName = await eHandle
+        .getProperty('tagName')
+        .then((e) => e.jsonValue());
```
mathiasbynens: nit: avoid single-letter variable names where possible (I know the old test set a bad example here)

`e` -> `element`?

test/ariaqueryhandler.spec.ts
```diff
@@ -0,0 +1,526 @@
...
+          h1.textContent = 'text content';
+          h1.id = 'node13';
+          shadowRoot.appendChild(h1);
+          document.documentElement.appendChild(div);
+          </script>
+          <!-- Elements inside shadow dom should be found -->
+          <img id="node20" src="" alt="Accessible Name">
+          <input id="node21" type="submit" value="Accessible Name">
+          <label id="node22" for="node23">Accessible Name</label>
+            <input id="node23">
```
mathiasbynens:
```suggestion
          <input id="node23">
```

mathiasbynens (commented):
Did a quick first pass already and left some comments. Will do a more thorough review tomorrow.

One thing that's missing from this PR is documentation. This is not yet technically required as long as custom query handlers are still experimental, but we'll need it when we ship them (which we hope to do soon after merging this). Just something to keep in mind.

Thanks for powering through here, very excited to see things coming together 👍

src/common/AriaQueryHandler.ts
johanbay: ah yes, I agree. I think this helper was initially `AXNode -> ElementHandle`.

src/common/AriaQueryHandler.ts
johanbay: Will do. We should also consider if this is the selector syntax we want.

src/common/AriaQueryHandler.ts
```diff
@@ -0,0 +1,133 @@
...
+}
+
+async function queryAXTree(
+  client: CDPSession,
+  element: ElementHandle,
+  accessibleName?: string,
+  role?: string
+): Promise<Protocol.Accessibility.AXNode[]> {
+  // @ts-ignore
+  const { nodes } = await client.send(
```
jackfranklin, line 30: why the ts-ignore? Is that because the protocol doesn't yet know about `Accessibility.queryAXTree` ?

src/common/AriaQueryHandler.ts
```diff
@@ -0,0 +1,133 @@
...
+    // @ts-ignore
+    'Accessibility.queryAXTree',
+    {
+      objectId: element._remoteObject.objectId,
+      accessibleName,
+      role,
+    }
+  );
+  // @ts-ignore
+  const filteredNodes: Protocol.Accessibility.AXNode[] = nodes.filter(
```
jackfranklin, line 40: same Q RE ts-ignore :)

src/common/AriaQueryHandler.ts
johanbay: yes exactly. Once we roll Chromium r809590 (or later) we can land this pr without the ts-ignores.

src/common/helper.ts
johanbay: Thanks, done.

test/ariaqueryhandler.spec.ts
johanbay: Done.

PR #6458, opened 2020-09-30
feat(chromium) roll Chromium to r809590

johanbay:
This corresponds to Chromium 87.0.4272.0
This roll includes:
- Reland "idle-detection: Implement requestPermission() method" ([crrev.com/c/2417779](crrev.com/c/2417779))


src/common/Browser.ts
```diff
@@ -650,7 +650,7 @@ export class BrowserContext extends EventEmitter {
    */
   async overridePermissions(
     origin: string,
-    permissions: Protocol.Browser.PermissionType[]
```
johanbay, line 653: I think this was a bug in the typings. With this type, we can only override with permissions where `web` and `protocol` permission names coincide?

mathiasbynens (approved):

PR #6459, opened 2020-09-30
feat(a11y-query): add aria query handler (without waitFor support)

johanbay:
This PR adds a new built-in handler for querying by accessible name and role (https://github.com/puppeteer/puppeteer/issues/6307).
Support for `waitForSelector` will be added in a follow-up PR.
This is part 2 of https://github.com/puppeteer/puppeteer/pull/6453 and this PR is therefore blocked on https://github.com/puppeteer/puppeteer/pull/6458.

### DIFF
--- a/src/common/AriaQueryHandler.ts
+++ b/src/common/AriaQueryHandler.ts
@@ -1,0 +1,122 @@
+/**
+ * Copyright 2020 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { InternalQueryHandler } from './QueryHandler.js';
+import { ElementHandle, JSHandle } from './JSHandle.js';
+import { Protocol } from 'devtools-protocol';
+import { CDPSession } from './Connection.js';
+
+async function queryAXTree(
+  client: CDPSession,
+  element: ElementHandle,
+  accessibleName?: string,
+  role?: string
+): Promise<Protocol.Accessibility.AXNode[]> {
+  const { nodes } = await client.send('Accessibility.queryAXTree', {
+    objectId: element._remoteObject.objectId,
+    accessibleName,
+    role,
+  });
+  const filteredNodes: Protocol.Accessibility.AXNode[] = nodes.filter(
+    (node: Protocol.Accessibility.AXNode) => node.role.value !== 'text'
+  );
+  return filteredNodes;
+}
+
+/*
+ * The selectors consist of an accessible name to query for and optionally
+ * further aria attributes on the form `[<attribute>=<value>]`.
+ * Currently, we only support the `name` and `role` attribute.
+ * The following examples showcase how the syntax works wrt. querying:
+ * - 'title[role="heading"]' queries for elements with name 'title' and role 'heading'.
+ * - '[role="img"]' queries for elements with role 'img' and any name.
+ * - 'label' queries for elements with name 'label' and any role.
+ * - '[name=""][role="button"]' queries for elements with no name and role 'button'.
+ */
+type ariaQueryOption = { name?: string; role?: string };
+function parseAriaSelector(selector: string): ariaQueryOption {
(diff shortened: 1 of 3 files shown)


src/common/AriaQueryHandler.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+const queryOne = async (element: ElementHandle, selector: string) => {
+  const exeCtx = element.executionContext();
+  const { name, role } = parseAriaSelector(selector);
+  const res = await queryAXTree(exeCtx._client, element, name, role);
+  if (res.length < 1) {
+    return null;
+  }
+  const handle = await exeCtx._adoptBackendNodeId(res[0].backendDOMNodeId);
```
OrKoN: I think you could just return without an intermediate handle variable?

src/common/AriaQueryHandler.ts
```diff
@@ -0,0 +1,92 @@
...
+  return handle;
+};
+
+const queryAll = async (element: ElementHandle, selector: string) => {
+  const exeCtx = element.executionContext();
+  const { name, role } = parseAriaSelector(selector);
+  const res = await queryAXTree(exeCtx._client, element, name, role);
+  const resHandles = Promise.all(
```
OrKoN: Same here

src/common/AriaQueryHandler.ts
```diff
@@ -0,0 +1,92 @@
...
+ * - 'label' queries for elements with name 'label' and any role.
+ */
+function parseAriaSelector(selector: string): { name?: string; role?: string } {
+  const s = selector.split('&');
+  const name = s[0] || undefined;
+  const role = s.length > 1 ? s[1] : undefined;
+  return { name, role };
+}
+
+const queryOne = async (element: ElementHandle, selector: string) => {
```
OrKoN: Could we explicitly define return types in functions here?

src/common/AriaQueryHandler.ts
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * @internal
+ */
+export const ariaHandler: InternalQueryHandler = {
+  queryOne,
+  waitFor: () => {
+    throw new Error('waitForSelector is not supported for aria selectors');
+  },
+  queryAll,
+  queryAllArray: async (element, selector) => {
```
OrKoN: Could we extract the function to the top level like the rest?

OrKoN (approved):

src/common/AriaQueryHandler.ts
johanbay: Thanks! done.

src/common/AriaQueryHandler.ts
johanbay: Done.

src/common/AriaQueryHandler.ts
johanbay: Thanks! done.

src/common/AriaQueryHandler.ts
johanbay: Done. I extracted the `waitFor` placeholder as well.

src/common/AriaQueryHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+ * The following examples showcase how the syntax works wrt. querying:
+ * - 'title[role="heading"]' queries for elements with name 'title' and role 'heading'.
+ * - '[role="img"]' queries for elements with role 'img' and any name.
+ * - 'label' queries for elements with name 'label' and any role.
+ * - '[name=""][role="button"]' queries for elements with no name and role 'button'.
+ */
+type ariaQueryOption = { name?: string; role?: string };
+function parseAriaSelector(selector: string): ariaQueryOption {
+  const normalize = (value: string): string => value.replace(/ +/g, ' ').trim();
+  const knownAttributes = ['name', 'role'];
```
mathiasbynens: Conceptually, this should be a `Set`. You can then use `Set#has` on line 59 (instead of `Array#includes`).

src/common/AriaQueryHandler.ts
```diff
@@ -0,0 +1,122 @@
...
+};
+
+const queryAllArray = async (
+  element: ElementHandle,
+  selector: string
+): Promise<JSHandle> => {
+  const elementHandles = await queryAll(element, selector);
+  const exeCtx = element.executionContext();
+  const jsHandle = exeCtx.evaluateHandle(
+    (...elements) => Array.from(elements),
```
mathiasbynens: Since rest parameters always produce arrays, do we need the `Array.from` here?

mathiasbynens (approved):
LGTM % nits

src/common/AriaQueryHandler.ts
johanbay: Good point. Thanks!

src/common/AriaQueryHandler.ts
johanbay: No that looks weird :) Done.

PR #6467, opened 2020-10-03
feat: add iPhone 11 models to DeviceDescriptors

yomed:
Hi, I noticed the latest iPhone models were missing. I got the `width`, `height`, `deviceScaleFactor` from this site: https://yesviz.com/devices.php

### DIFF
--- a/src/common/DeviceDescriptors.ts
+++ b/src/common/DeviceDescriptors.ts
@@ -549,6 +549,84 @@ const devices: Device[] = [
     },
   },
   {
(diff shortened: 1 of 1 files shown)


yomed:
@googlebot I signed it!

whalderman (commented):
I don't know if the `userAgent`s need to be updated to the latest supported version. I know they aren't terribly important, so if up-to-date user agents are not necessary, this'll work fine.

mathiasbynens (approved):

PR #6472, opened 2020-10-05
feat(a11y-query): extend aria handler with waitFor

### LINKED ISSUE #6307: Proposal: Query by accessible name

johanbay:
This PR adds `waitFor` to the built-in aria handler (#6307).
This is part 3 of https://github.com/puppeteer/puppeteer/pull/6453.

### DIFF
--- a/src/common/AriaQueryHandler.ts
+++ b/src/common/AriaQueryHandler.ts
@@ -18,6 +18,7 @@ import { InternalQueryHandler } from './QueryHandler.js';
 import { ElementHandle, JSHandle } from './JSHandle.js';
 import { Protocol } from 'devtools-protocol';
 import { CDPSession } from './Connection.js';
+import { DOMWorld, WaitForSelectorOptions } from './DOMWorld.js';
 
 async function queryAXTree(
   client: CDPSession,
@@ -82,8 +83,17 @@ const queryOne = async (
   return exeCtx._adoptBackendNodeId(res[0].backendDOMNodeId);
 };
 
-const waitFor = () => {
-  throw new Error('waitForSelector is not supported for aria selectors');
+const waitFor = async (
+  domWorld: DOMWorld,
+  selector: string,
+  options: WaitForSelectorOptions
+) => {
+  await addHandlerToWorld(domWorld);
+  return domWorld.waitForSelectorInPage(
+    (_: Element, selector: string) => globalThis.ariaQuerySelector(selector),
+    selector,
+    options
+  );
 };
 
 const queryAll = async (
@@ -110,6 +120,14 @@ const queryAllArray = async (
   );
   return jsHandle;
 };
+
+async function addHandlerToWorld(domWorld: DOMWorld) {
+  await domWorld.addBinding('ariaQuerySelector', async (selector: string) => {
+    const document = await domWorld._document();
+    const element = await queryOne(document, selector);
+    return element;
+  });
+}
 
 /**
  * @internal
--- a/src/common/DOMWorld.ts
+++ b/src/common/DOMWorld.ts
@@ -15,7 +15,7 @@
  */
 
 import { assert } from './assert.js';
-import { helper } from './helper.js';
+import { helper, debugError } from './helper.js';
 import {
   LifecycleWatcher,
   PuppeteerLifeCycleEvent,
@@ -36,13 +36,14 @@ import {
   UnwrapPromiseLike,
 } from './EvalTypes.js';
 import { isNode } from '../environment.js';
+import { Protocol } from 'devtools-protocol';
 
 // predicateQueryHandler and checkWaitForOptions are declared here so that
 // TypeScript knows about them when used in the predicate function below.
 declare const predicateQueryHandler: (
   element: Element | Document,
   selector: string
-) => Element | Element[] | NodeListOf<Element>;
+) => Promise<Element | Element[] | NodeListOf<Element>>;
 declare const checkWaitForOptions: (
   node: Node,
   waitForVisible: boolean,
@@ -76,6 +77,11 @@ export class DOMWorld {
    */
   _waitTasks = new Set<WaitTask>();
 
+  // Contains mapping from functions that should be bound to Puppeteer functions.
+  private _boundFunctions = new Map<string, Function>();
+  // Set of bindings that have been registered in the current context.
+  private _ctxBindings = new Set<string>();
+
   constructor(
     frameManager: FrameManager,
     frame: Frame,
@@ -85,16 +91,23 @@ export class DOMWorld {
     this._frame = frame;
     this._timeoutSettings = timeoutSettings;
     this._setContext(null);
+    frameManager._client.on('Runtime.bindingCalled', (event) =>
+      this._onBindingCalled(event)
+    );
   }
 
   frame(): Frame {
     return this._frame;
   }
 
-  _setContext(context?: ExecutionContext): void {
+  async _setContext(context?: ExecutionContext): Promise<void> {
     if (context) {
       this._contextResolveCallback.call(null, context);
       this._contextResolveCallback = null;
+      this._ctxBindings.clear();
+      for (const name of this._boundFunctions.keys()) {
+        await this.addBindingToContext(name);
+      }
       for (const waitTask of this._waitTasks) waitTask.rerun();
     } else {
       this._documentPromise = null;
@@ -119,7 +132,7 @@ export class DOMWorld {
   executionContext(): Promise<ExecutionContext> {
     if (this._detached)
       throw new Error(
-        `Execution Context is not available in detached frame "${this._frame.url()}" (are you trying to evaluate?)`
+        `Execution context is not available in detached frame "${this._frame.url()}" (are you trying to evaluate?)`
       );
     return this._contextPromise;
   }
@@ -438,7 +451,6 @@ export class DOMWorld {
 
   async tap(selector: string): Promise<void> {
     const handle = await this.$(selector);
-    assert(handle, 'No node found for selector: ' + selector);
     await handle.tap();
     await handle.dispose();
   }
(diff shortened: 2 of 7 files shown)


src/common/AriaQueryHandler.ts
```diff
@@ -111,6 +121,29 @@ const queryAllArray = async (
   return jsHandle;
 };
 
+// If multiple waitFor are set up asynchronously, we need to wait for the first
+// one to set up the binding in the page before running the others.
+let settingUpBinding = null;
```
johanbay: Is this a reasonable way to implement a semaphore-like waiting logic between asynchrounous operations?

src/common/AriaQueryHandler.ts
```diff
@@ -111,6 +121,29 @@ const queryAllArray = async (
...
+  }
+  if (!domWorld.hasBinding('ariaQuerySelector')) {
+    let done: () => void | null = null;
+    settingUpBinding = new Promise((resolve) => {
+      done = resolve;
+    });
+    await domWorld.addBinding('ariaQuerySelector', async (selector: string) => {
+      const document = await domWorld._document();
+      const element = await queryOne(document, selector);
+      return element;
```
mathiasbynens: This is complicated and a little magical to me, so apologies if I misunderstood, but: would something like this work here?

```js
hasBinding = true;
```

instead of `settingUpBinding` and `done`

src/common/DOMWorld.ts
```diff
@@ -464,6 +473,82 @@ export class DOMWorld {
...
+      // exection context may have been destroyed.
+      // In both caes, the promises above are rejected with a protocol error.
+      // We can safely ignores these, as the WaitTask is re-installed in
+      // the next execution context if needed.
+      if (error.message.includes('Protocol error')) return;
+      debugError(error);
+    }
+    function deliverResult(name: string, seq: number, result: unknown): void {
+      window[name].callbacks.get(seq).resolve(result);
+      window[name].callbacks.delete(seq);
```
mathiasbynens: `window` -> `globalThis` like elsewhere in the PR? That way we support worker contexts as well. Even if those don't need to be supported in this code path, it's more consistent with other code in this PR

mathiasbynens (commented):
Not done reviewing, but some initial comments before I sign off

src/common/AriaQueryHandler.ts
johanbay: I'm not sure how that would work. Suppose we do the following
```addHandlerToWorld(...); await addHandlerToWorld(...);```
The first call starts adding the binding, but at some point it is blocked while waiting for the CDP response. Now, the second call can either decide that the binding has already been added (which it hasn't yet) or itself start adding the binding as well.
The code here tries to fix this situation by having the first call set up a promise in `settingUpBinding` which it resolves when it is done adding the binding by calling `done()`. The second call sees that the other call is setting up the binding and blocks until the `settingUpBinding` promise is resolved.

Is there a simpler (or more canonical) way to achieve this?

Comment (mathiasbynens) [reply src/common/AriaQueryHandler.ts]: I see — we don't just need to know whether or not the binding has been set up, we want to know whether it's already _in the process of being set up_.

I know this is captured in your code comment above but somehow I didn't understand the full intent behind this code until just now — thanks for the explanation!

File: src/common/DOMWorld.ts [hunk 4]
```diff
@@ -464,6 +473,82 @@ export class DOMWorld {
...
+    }
+  }
+
+  /**
+   * @internal
+   */
+  async addBinding(name: string, puppeteerFunction: Function): Promise<void> {
+    if (this._ctxBindings.has(name))
+      throw new Error(
+        `Failed to add page binding with name ${name}: window['${name}'] already exists!`
```
mathiasbynens: Let’s make sure this produces a correct message for arbitrary string values for `name`s (e.g. if `name` includes a `'`):

```suggestion
        `Failed to add page binding with name ${name}: window[${JSON.stringify(name)}] already exists!`
```

Similar to the other comment, we could use `globalThis` instead of `window` — WDYT?

src/common/AriaQueryHandler.ts
```diff
@@ -111,6 +121,29 @@ const queryAllArray = async (
...
+async function addHandlerToWorld(domWorld: DOMWorld) {
+  if (settingUpBinding) {
+    await settingUpBinding;
+  }
+  if (!domWorld.hasBinding('ariaQuerySelector')) {
+    let done: () => void | null = null;
+    settingUpBinding = new Promise((resolve) => {
+      done = resolve;
+    });
+    await domWorld.addBinding('ariaQuerySelector', async (selector: string) => {
```
OrKoN: Can you do `settingUpBinding  = domWorld.addBinding(...)` and then `await settingUpBinding`? I think it might help avoid having the extra promise?

src/common/DOMWorld.ts
```diff
@@ -715,23 +764,34 @@ class WaitTask {
       await success.dispose();
       return;
     }
+    // When frame is detached the task should have been terminated by the DOMWorld.
+    // This can fail if we were adding this task while the frame was detached,
+    // so we terminate here instead.
+    if (error) {
+      if (
+        error.message.includes(
+          'Execution Context is not available in detached frame'
```
mathiasbynens: nit:

```suggestion
          'Execution context is not available in detached frame'
```

src/common/AriaQueryHandler.ts
OrKoN: So bindings are set up per world? What would happen if we do `addHandlerToWorld(world1); addHandlerToWorld(world2)`? Is it possible to add bindings to two different worlds concurrently?

src/common/helper.ts
```diff
@@ -177,6 +177,115 @@ function evaluationString(fun: Function | string, ...args: unknown[]): string {
   return `(${fun})(${args.map(serializeArgument).join(',')})`;
 }
 
+function pageBindingInitString(type: string, name: string): string {
+  return evaluationString(addPageBinding, type, name);
```
mathiasbynens: I don't like relying on function hoisting — WDYT about moving the `return` to be the last line in this function body?

test/ariaqueryhandler.spec.ts
```diff
@@ -173,6 +174,305 @@ describeChromeOnly('AriaQueryHandler', () => {
...
+      await page.setContent(`<div aria-label='zombo'>anything</div>`);
+      expect(
+        await page.evaluate(
+          (x: HTMLElement) => x.textContent,
+          await waitForSelector
+        )
+      ).toBe('anything');
+    });
+
+    // TODO: Figure out what the stack trace should look like for custom handlers
```
mathiasbynens: What does it look like currently? Can you make a recommendation?

src/common/AriaQueryHandler.ts
johanbay: Ah, good point! In practice, bindings are set up per execution context, but the world is responsible for re-binding when the execution context changes. It's fine to add bindings in two different worlds concurrently (it should happen when doing `waitForSelector` in two different frames), so maybe this code actually synchronizes a bit too much. I'll try pushing it into the worlds instead.

src/common/DOMWorld.ts
johanbay: Done. Also updated the actual error.

src/common/DOMWorld.ts
johanbay: Good idea, done.

src/common/helper.ts
johanbay: Done.

src/common/AriaQueryHandler.ts
johanbay: Done, but moved into DOMWorld instead.

test/ariaqueryhandler.spec.ts
johanbay: This is the current stack trace:
```
"TimeoutError: waiting for selector `zombo` failed: timeout 10ms exceeded
    at new WaitTask (.../puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:494:34)
    at DOMWorld.waitForSelectorInPage (.../puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:427:26)
    at Object.waitFor (.../puppeteer/lib/cjs/puppeteer/common/AriaQueryHandler.js:55:21)
    at process._tickCallback (internal/process/next_tick.js:68:7)"
```
It would be nice if `ariaqueryhandler.spec.ts` was part of the trace since that is the actual call site for `waitForSelector`.
The fact that the selector is `zombo` instead of `aria/zombo` is maybe a bit unfortunate as well.

src/common/AriaQueryHandler.ts
johanbay: Moved to `DOMWorld.ts` to allow asyncronously setting up bindings in multiple worlds.

src/common/DOMWorld.ts
johanbay: Done.

test/ariaqueryhandler.spec.ts
johanbay: It seems like we can't do much better than this since waitFor is async.

src/common/DOMWorld.ts
```diff
@@ -464,6 +476,86 @@ export class DOMWorld {
...
+  /**
+   * @internal
+   */
+  async addBindingToContext(name: string) {
+    // Previous operation added the binding so we are done.
+    if (this._ctxBindings.has(name)) return;
+    // Wait for other operation to finish
+    if (this._settingUpBinding) {
+      await this._settingUpBinding;
+      return this.addBindingToContext(name);
```
OrKoN, line 491: Would it set up the binding for the same world again?

johanbay: No it shouldn't. The first check `this._ctxBindings.has(name)` should be true if the binding was added by the current operation, so it will return immediately.
The idea behind the recursive call is that if multiple tasks are blocked on `await this._settingUpBinding;`, then one of them will get to do the recursive call and assign a new promise to `_settingUpBinding`. When the other tasks gets to do the recursive call they are therefore blocked here again.

OrKoN (approved):

mathiasbynens (approved):

PR #6476, opened 2020-10-07
docs(queryhandler): add custom query handler docs

johanbay:


src/common/Puppeteer.ts
```diff
@@ -284,7 +284,18 @@ export class Puppeteer {
...
   /**
-   * @internal
+   * Registers a {@link CustomQueryHandler | custom query handler}. After
+   * registration, the handler can be used everywhere where a selector is
+   * expected by prepending the selection string with `<name>/`. The name is
+   * only allowed to consist of lower- and upper case latin letters.
+   * @example
+   * ```
+   * puppeteer.registerCustomQueryHandler('text', { ... });
+   * const aHandle = await page.$('text/...');
```
mathiasbynens: Let’s use `…` to differentiate from `...` which is JS syntax in some contexts.

```suggestion
   * puppeteer.registerCustomQueryHandler('text', { … });
   * const aHandle = await page.$('text/…');
```

mathiasbynens (approved):
LGTM!

When we ship custom query handlers (remove the `experimental` status) we can essentially copy this to the `docs/api.md` as well.

src/common/Puppeteer.ts
johanbay: Ah yes that sounds like a good idea. Thanks!

PR #6477, opened 2020-10-07
chore(agnostify): Create Node and Web initializer.

jackfranklin:
This PR splits `initialize.ts` into two files, one for web, and one for
Node. The Node initializer requires much more information as it needs to
know which browser(s) to download and where to store them, whereas the
web one does not.

A future PR that I'm working on will tidy up `src/common/Puppeteer.ts`
(as it contains a lot of Node specific logic around downloading and
installing browsers), but this change enables us to stop the browser
bundle attempting to use the `pkg-dir` dependency, which wouldn't work
within a browser, as well as keeping the size of the PRs down and
avoiding one mammoth PR.

### DIFF
--- a/src/initialize-web.ts
+++ b/src/initialize-web.ts
@@ -1,0 +1,34 @@
+/**
+ * Copyright 2020 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { Puppeteer } from './common/Puppeteer.js';
+
+export const initializePuppeteerWeb = (packageName: string): Puppeteer => {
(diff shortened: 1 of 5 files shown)


src/initialize-web.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const initializePuppeteerWeb = (packageName: string): Puppeteer => {
+  const isPuppeteerCore = packageName === 'puppeteer-core';
+
+  // puppeteer-core ignores environment variables
+  return new Puppeteer(
+    // Product root directory is undefined as we're not concerned about
+    // downloading and installing browsers in the web environment
+    undefined,
+    // Preferred revision is undefined as we will use the browser we are running in.
```
mathiasbynens: Ah, this gives me another opportunity to share my favorite tech writing tip ever: 

> Avoid “will”.

http://blog.stoyanstefanov.com/technical-writing-checklist/#:~:text=%22will%22

It’s so simple, but it really works — it's almost always better to remove it. Just wanted to share in case you like it too :)

src/initialize-web.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import { Puppeteer } from './common/Puppeteer.js';
+
+export const initializePuppeteerWeb = (packageName: string): Puppeteer => {
+  const isPuppeteerCore = packageName === 'puppeteer-core';
+
+  // puppeteer-core ignores environment variables
+  return new Puppeteer(
+    // Product root directory is undefined as we're not concerned about
+    // downloading and installing browsers in the web environment
```
mathiasbynens: nit: Comments should read like sentences.

```suggestion
    // downloading and installing browsers in the web environment.
```

mathiasbynens (approved):

PR #6484, opened 2020-10-08
chore(agnostification): split up launcher class

jackfranklin:
The `Launcher` class was serving two purposes:

1. Launch browsers
2. Connect to browsers

Number 1) only needs to be done in Node land, but 2) is agnostic; in a
browser version of Puppeteer we'll need the ability to connect over a
websocket to send commands back and forth.

As part of the agnostification work we needed to split the `Launcher` up
so that the connection part can be made agnostic. Additionally, I
removed dependencies on `https`, `http` and `URL` from Node, instead
leaning on fetch (via `node-fetch` if in Node land) and the browser
`URL` API (which was added to Node in Node 10).


jackfranklin:
There's more work to be done here as the `Puppeteer` class remains littered with Node specific concerns, but I'm breaking it down into small steps to make it easier to work through. PTAL :)

src/common/BrowserConnector.ts
```diff
@@ -0,0 +1,109 @@
...
+      method: 'GET',
+    });
+    if (!result.ok) {
+      throw new Error(`HTTP ${result.statusText}`);
+    }
+    const data = await result.json();
+    return data.webSocketDebuggerUrl;
+  } catch (error) {
+    error.message =
+      `Failed to fetch browser webSocket url from ${endpointURL}: ` +
```
mathiasbynens:
```suggestion
      `Failed to fetch browser webSocket URL from ${endpointURL}: ` +
```

src/common/Puppeteer.ts
```diff
@@ -141,11 +140,12 @@ export class Puppeteer {
...
       browserURL?: string;
       transport?: ConnectionTransport;
-      product?: string;
+      product?: Product;
     }
   ): Promise<Browser> {
     if (options.product) this._productName = options.product;
-    return this._launcher.connect(options);
+    return connectToBrowser(options);
+    // return this._launcher.connect(options);
```
mathiasbynens:
```suggestion
```

src/node/Launcher.ts
```diff
@@ -392,63 +324,6 @@ class FirefoxLauncher implements ProductLauncher {
...
-    );
-    return Browser.create(
-      connection,
-      browserContextIds,
-      ignoreHTTPSErrors,
-      defaultViewport,
-      null,
-      () => connection.send('Browser.close').catch(debugError)
-    );
-  }
```
jackfranklin, line 450: The `FirefoxLauncher` class and the `ChromeLauncher` class both defined the `connect` method, but it was identical on both. Hence now there being just a single function to do the connecting.

src/common/fetch.ts
```diff
@@ -0,0 +1,22 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { isNode } from '../environment.js';
+
+/* Use the global version if we're in the browser, else load the node-fetch module */
```
mathiasbynens: nit: end comments with a period

mathiasbynens (approved):

OrKoN (approved):

mathiasbynens (approved):

PR #6485, opened 2020-10-08
test: use sourcemaps in unit tests

psmarshall:
Introduce the source-map-support package and require it for mocha
running unit tests.

Turn on the sourceMap option for tsconfig.base.json so that the
sourceMappingURL= line is emitted in the generated files.

Before:
Screen Shot 2020-10-08 at 1 10 45 pm

After:
Screen Shot 2020-10-08 at 11 58 04 am


jackfranklin (approved):

package.json
```diff
@@ -53,6 +53,7 @@
     "progress": "^2.0.1",
     "proxy-from-env": "^1.0.0",
     "rimraf": "^3.0.2",
+    "source-map-support": "^0.5.19",
```
mathiasbynens: Since Puppeteer users don’t need this dependency, can we move this to `devDependencies` instead?

tsconfig.base.json
```diff
@@ -8,6 +8,7 @@
     "module": "ESNext",
     "declaration": true,
     "declarationMap": true,
-    "resolveJsonModule": true
+    "resolveJsonModule": true,
+    "sourceMap": true,
```
mathiasbynens, line 12: It triggers me that this is `.json` but actually supports trailing commas 🤯

mathiasbynens (changes requested):

package.json
psmarshall: I thought I did this but npm disagrees. Who's to say who is right and who is wrong...

tsconfig.base.json
psmarshall: 😨

mathiasbynens (approved):

PR #6504, opened 2020-10-12
chore(agnostification): split up root Puppeteer class

jackfranklin:
The `Puppeteer` class had two concerns:

* connect to an existing browser
* launch a new browser

The first of those concerns is needed in all environments, but the
second is only needed in Node.
https://github.com/puppeteer/puppeteer/pull/6484 landing enabled us to
pull the `Puppeteer` class apart into two:

1. `Puppeteer` which hosts the behaviour for connecting to existing
   browsers.
2. `PuppeteerNode`, which extends `Puppeteer` and also adds the ability
   to launch a new browser.

This is a non-breaking change, because Node users will still get an
instance of a class with all the methods they expect, but it'll be a
`PuppeteerNode` rather than `Puppeteer`. I don't expect this to cause
people any issues.

We also now have new files that are effectively the entry points for
Puppeteer:

* `node.ts`: the main entry point for Puppeteer on Node.
* `web.ts`: the main entry point for Puppeteer on the web.
* `node-puppeteer-core.ts`: for those using puppeteer-core (which only
  exists in Node, not on the web).


jackfranklin:
Puppeteer Node vs Browser entry points (1)

I drew this diagram as it might help visualise the entry points for each environment.

mathiasbynens (approved):
Loved the diagram!

PR #6509, opened 2020-10-13
feat(queryhandler): add built-in pierce handler

### LINKED ISSUE #6217: Shadow Dom support

johanbay:
Adds a handler 'pierce' that pierces shadow roots while querying.
Closes #6217.

### DIFF
--- a/src/common/QueryHandler.ts
+++ b/src/common/QueryHandler.ts
@@ -112,7 +112,60 @@ const _defaultHandler = makeQueryHandler({
     element.querySelectorAll(selector),
 });
 
-const _builtInHandlers = new Map([['aria', ariaHandler]]);
+const pierceHandler = makeQueryHandler({
+  queryOne: (element, selector) => {
+    let found: Element | null = null;
+    const search = (root: Element | ShadowRoot) => {
+      const iter = document.createTreeWalker(root, NodeFilter.SHOW_ELEMENT);
+      do {
+        const currentNode = iter.currentNode as HTMLElement;
+        if (currentNode.shadowRoot) {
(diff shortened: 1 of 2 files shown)


src/common/QueryHandler.ts
```diff
@@ -112,7 +112,57 @@ const _defaultHandler = makeQueryHandler({
...
+    };
+    if (element instanceof Document) {
+      element = element.documentElement;
+    }
+    collect(element);
+    return result;
+  },
+});
+
+const _builtInHandlers = new Map([['aria', ariaHandler], ['pierce', pierceHandler]]);
```
mathiasbynens: Now that there are multiple entries, let's split this across multiple lines for readability:

```suggestion
const _builtInHandlers = new Map([
  ['aria', ariaHandler],
  ['pierce', pierceHandler],
]);
```

(unless Prettier likes to put it all on one line for some reason?)

mathiasbynens (approved):

php4fan:
Why is this still undocumented??

j-wsl:
is it possible to get this with xpaths?

gitressa:
I was also looking for this and a simple example, and found one by @Remco Vaes in https://stackoverflow.com/a/76967459:

> This issue is solved with [P selectors in Puppetteer](https://pptr.dev/guides/query-selectors/#-and--combinators).
> 
> In this case you could use `page.click('>>> #my-button')`. The `>>>` Allow you to select stuff in shadowRoots.
> 
> I know this issue is old, but I needed this and this issue still pops up on searches.

PR #6515, opened 2020-10-15
chore(agnostification): common/helper.ts

jackfranklin:
The `readProtocolStream` method uses `fs` only if you want to write to a
file. So we gate at the start of the function and ensure that if we got
given a path we are not in a Node environment.

We then use `import('fs')` to lazily load fs. We have multiple calls but
they will be cached so I don't believe that it's going to be a perf hit.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,6 +59,8 @@ jobs:
 
     - node_js: "12.16.3"
       name: 'Browser tests: Linux/Chromium'
+      addons:
+        chrome: stable
       env:
         - CHROMIUM=true
       script:
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "@types/ws": "^7.2.4",
     "@typescript-eslint/eslint-plugin": "^4.4.0",
     "@typescript-eslint/parser": "^4.4.0",
-    "@web/test-runner": "^0.8.4",
+    "@web/test-runner": "^0.9.2",
     "commonmark": "^0.28.1",
     "cross-env": "^7.0.2",
     "dependency-cruiser": "^9.7.0",
--- a/src/common/helper.ts
+++ b/src/common/helper.ts
@@ -15,11 +15,11 @@
  */
 import { TimeoutError } from './Errors.js';
 import { debug } from './Debug.js';
-import * as fs from 'fs';
 import { CDPSession } from './Connection.js';
 import { Protocol } from 'devtools-protocol';
 import { CommonEventEmitter } from './EventEmitter.js';
 import { assert } from './assert.js';
+import { isNode } from '../environment.js';
 
 export const debugError = debug('puppeteer:error');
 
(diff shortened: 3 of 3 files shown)


src/common/helper.ts
```diff
@@ -309,9 +309,14 @@ async function readProtocolStream(
   handle: string,
   path?: string
 ): Promise<Buffer> {
+  if (!isNode && path) {
+    throw new Error('Cannot write to a path outside of Node.js environment.');
+  }
+
   let eof = false;
-  let fileHandle: fs.promises.FileHandle;
+  let fileHandle: import('fs').promises.FileHandle;
```
mathiasbynens, line 319: How is `.promises` already available synchronously? Does this not need `await`?

jackfranklin: No, because this is a type, so TS understands what we mean here. It's just so we can still type the `fileHandle` variable, but without an `import * as fs...` at the top of the file.

src/common/helper.ts
```diff
@@ -309,9 +309,14 @@ async function readProtocolStream(
   handle: string,
   path?: string
 ): Promise<Buffer> {
+  if (!isNode && path) {
+    throw new Error('Cannot write to a path outside of Node.js environment.');
+  }
+
```
mathiasbynens, line 315: Can't we do `let fs;` and then `fs = ...` inside one of the `if`s here? (It seems fine to rely on Node.js' caching regardless, but if we can avoid it and simplify the code at the same time...)

src/common/helper.ts
jackfranklin: (this line doesn't make it into the compiled output)

src/common/helper.ts
mathiasbynens: I meant `let fs` — updated comment to clarify

jackfranklin: 👍 PTAL!

.travis.yml
```diff
@@ -59,6 +59,8 @@ jobs:
 
     - node_js: "12.16.3"
       name: 'Browser tests: Linux/Chromium'
+      addons:
+        chrome: stable
```
jackfranklin, line 63: needed to ensure Travis has an up to date chrome that web test runner can use

package.json
```diff
@@ -74,7 +74,7 @@
     "@types/ws": "^7.2.4",
     "@typescript-eslint/eslint-plugin": "^4.4.0",
     "@typescript-eslint/parser": "^4.4.0",
-    "@web/test-runner": "^0.8.4",
+    "@web/test-runner": "^0.9.2",
```
jackfranklin, line 77: the old version started timing out on CI...upgrading to this fixes it. Not entirely sure why!

.travis.yml
LarsDenBakker: We could also point it to the bundled chromium which puppeteer already brings. Let me know if I can help here :)

src/common/helper.ts
mathiasbynens: Oh yeah, I totally misread it as run-time code. Got it!

.travis.yml
jackfranklin: @LarsDenBakker thanks! I'm quite keen to see if it's stable on a regular Chrome, but if not we'll reach for that and I expect I'll ping you for help :D

mathiasbynens (approved):

PR #6520, opened 2020-10-16
chore(agnostification): agnostify web socket connections

jackfranklin:
This PR updates the socket transport code to swap between a Node web socket transport or a web one based on `isNode`. It also adds unit tests to the browser tests that show we can connect in a browser.

### DIFF
--- a/src/common/BrowserConnector.ts
+++ b/src/common/BrowserConnector.ts
@@ -19,9 +19,9 @@ import { Browser } from './Browser.js';
 import { assert } from './assert.js';
 import { debugError } from '../common/helper.js';
 import { Connection } from './Connection.js';
-import { WebSocketTransport } from './WebSocketTransport.js';
 import { getFetch } from './fetch.js';
 import { Viewport } from './PuppeteerViewport.js';
+import { isNode } from '../environment.js';
 
 /**
  * Generic browser options that can be passed when launching any browser.
@@ -32,6 +32,13 @@ export interface BrowserOptions {
   defaultViewport?: Viewport;
   slowMo?: number;
 }
+
+const getWebSocketTransportClass = async () => {
+  return isNode
+    ? (await import('../node/NodeWebSocketTransport.js')).NodeWebSocketTransport
+    : (await import('./BrowserWebSocketTransport.js'))
+        .BrowserWebSocketTransport;
+};
 
 /**
  * Users should never call this directly; it's called when calling
@@ -44,7 +51,7 @@ export const connectToBrowser = async (
     browserURL?: string;
     transport?: ConnectionTransport;
   }
-) => {
+): Promise<Browser> => {
   const {
     browserWSEndpoint,
     browserURL,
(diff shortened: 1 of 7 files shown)


src/common/BrowserConnector.ts
```diff
@@ -33,6 +33,13 @@ export interface BrowserOptions {
   slowMo?: number;
 }
 
+const getWebSocketTransportClass = async () => {
+  return isNode
+    ? (await import('../node/NodeWebSocketTransport.js')).NodeWebSocketTransport
+    : (await import('./BrowserWebSocketTransport.js'))
```
paullewis, line 39: Does there need to be a subdirectory for browser-centric code?

src/common/BrowserWebSocketTransport.ts
```diff
@@ -0,0 +1,55 @@
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
```
paullewis: 2020

src/common/BrowserWebSocketTransport.ts
```diff
@@ -0,0 +1,55 @@
...
+      const ws = new WebSocket(url);
+
+      ws.addEventListener('open', () =>
+        resolve(new BrowserWebSocketTransport(ws))
+      );
+      ws.addEventListener('error', reject);
+    });
+  }
+
+  _ws: WebSocket;
```
paullewis: I was wondering, is _ws consistent with the rest of the codebase, or should this be private?

src/node/BrowserRunner.ts
```diff
@@ -22,7 +22,7 @@ import { assert } from '../common/assert.js';
 import { helper, debugError } from '../common/helper.js';
 import { LaunchOptions } from './LaunchOptions.js';
 import { Connection } from '../common/Connection.js';
-import { WebSocketTransport } from '../common/WebSocketTransport.js';
+import { NodeWebSocketTransport as WebSocketTransport } from '../node/NodeWebSocketTransport.js';
```
paullewis, line 25: Do you want to rename the import? Might be better to refactor it to use the new name.

paullewis (approved):
LGTM with nits.

mathiasbynens (approved):

src/common/BrowserConnector.ts
jackfranklin: Yeah, I think probably there does. But then it'll get a bit messy with node/browser/common where common will sometimes conditionally load in things. But maybe that's fine? 

I might leave this for a follow up PR - let me know if you feel strongly to fix this now 👍

src/common/BrowserWebSocketTransport.ts
jackfranklin: Good catch

src/common/BrowserConnector.ts
paullewis: I don't personally think it's a blocker if you're planning to do it anyway.

src/common/BrowserWebSocketTransport.ts
paullewis: On that, should it be `ws` or `_ws`? I don't have strong opinions besides it being consistent.

src/node/BrowserRunner.ts
jackfranklin: Yeah, I think so. Will rework in follow-up.

PR #6524, opened 2020-10-19
chore(agnostification): create `web` for web only code

jackfranklin:
This change makes a `web` folder to house any code that is just for the
browser and won't work in a Node environment.

We have three folders now to divide code up nicely:

- `node`
- `web`
- `common`.


jackfranklin:
One thing I'd welcome opinions on: the two `WebSocketTransport` classes are almost identical bar the slight difference in how we create the socket (Node's ws lib vs browser) and in the types (one has `private ws: NodeWebSocket` and the other `private ws: WebSocket`).

That said, I cannot find a nice way to avoid the duplication in the code that doesn't 1) make it far more complex, 2) lose some type safety. That's why I've left the duplication as I think it's actually best (or least worst) compared to forcing an abstraction.

mathiasbynens:
How did you arrive at `web` and not `browser`? The latter seems more precise but I might be missing some context.

jackfranklin:
> How did you arrive at web and not browser? The latter seems more precise but I might be missing some context.

None really beyond I think we've talked about it as `puppeteer-web` and I named the entrypoint `web.js`. I don't have a strong feeling either way on which we should go but we should be consistent.

jschfflr:
@googlebot I consent.

jschfflr:
@jackfranklin I updated the pr - let me know if you think it's still relevant.
There's one change in the Node WebSocket that I'm not sure we can also include in the web version:
https://github.com/puppeteer/puppeteer/blob/cb4470a6d9b0a7f73836458bb3d5779eb85ac5f2/src/node/NodeWebSocketTransport.ts#L29

Do you know if that would be possible?

jschfflr:
Closing this for now as discussed offline with @jackfranklin

PR #6538, opened 2020-10-22
fix: ignore spurious bindingCalled events

johanbay:


src/common/DOMWorld.ts
```diff
@@ -537,7 +537,15 @@ export class DOMWorld {
   private async _onBindingCalled(
     event: Protocol.Runtime.BindingCalledEvent
   ): Promise<void> {
-    const { type, name, seq, args } = JSON.parse(event.payload);
+    let payload: { type: string; name: string; seq: number; args: unknown[] };
+    try {
+      payload = JSON.parse(event.payload);
+    } catch {
+      // The binding was either called by something in the page or it was
+      // called before our wrapper was initialized
```
jackfranklin: just for my understanding: what might cause this? Just some code running before we set up our binding ?

johanbay: Exactly. I can't make out how it happens, but very rarely we'll call the `ariaQueryHandler` binding before the [pageBinding](https://github.com/puppeteer/puppeteer/blob/main/src/common/helper.ts#L180) wrapper is set up.

jackfranklin (approved):

src/common/DOMWorld.ts
mathiasbynens: nitty nit

```suggestion
      // called before our wrapper was initialized.
```

src/common/Page.ts
```diff
@@ -1132,7 +1132,15 @@ export class Page extends EventEmitter {
   private async _onBindingCalled(
     event: Protocol.Runtime.BindingCalledEvent
   ): Promise<void> {
-    const { type, name, seq, args } = JSON.parse(event.payload);
+    let payload: { type: string; name: string; seq: number; args: unknown[] };
+    try {
+      payload = JSON.parse(event.payload);
+    } catch {
+      // The binding was either called by something in the page or it was
+      // called before our wrapper was initialized
```
mathiasbynens:
```suggestion
      // called before our wrapper was initialized.
```

mathiasbynens (approved):

PR #6549, opened 2020-10-26
fix(helper): access promises from raw 'fs' module

### LINKED ISSUE #6548: Compatible issue with node@10.13 - node@10.17
### Steps to reproduce

* Puppeteer version: `5.4.0`
* Platform / OS version: `macOS 10.15.17
* URLs (if applicable):
* Node.js version: `10.15.0`

`fs.promises` is experimental in `node@10.13 ~ node@10.17`, so the result of `Object.getOwnPropertyDescriptor(require('fs'), 'promises').enumerable` is `false`, which means https://github.com/puppeteer/puppeteer/blob/main/src/common/helper.ts#L322 this line would throw `Can not read property open from undefined`

Brooooooklyn:
Close https://github.com/puppeteer/puppeteer/issues/6548

### DIFF
--- a/src/common/helper.ts
+++ b/src/common/helper.ts
@@ -319,7 +319,7 @@ async function readProtocolStream(
   let fileHandle: import('fs').promises.FileHandle;
 
   if (path && fs) {
-    fileHandle = await fs.promises.open(path, 'w');
+    fileHandle = await fs.default.promises.open(path, 'w');
   }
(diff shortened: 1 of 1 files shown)


jackfranklin:
@Brooooooklyn thanks for your help here. I took your work and built on it a little and a fix will land in #6550 so I'm closing this.

PR #6550, opened 2020-10-26
chore: run unit tests on node 10.15 + fix fs.promises access

### LINKED ISSUE #6548: Compatible issue with node@10.13 - node@10.17

jackfranklin:
We saw in https://github.com/puppeteer/puppeteer/issues/6548 that the
`fs.promises` module was experimental in Node <10.17 and as such we
introduced issues for users on 10.15.

Until we can drop Node v10 (it's EOL is 30-04-20201
https://github.com/nodejs/Release#release-schedule) we should run our
tests on an old Node 10 to avoid regressing in this area.

Closes #6548 .

### DIFF
--- a/src/common/helper.ts
+++ b/src/common/helper.ts
@@ -313,7 +313,7 @@ async function readProtocolStream(
     throw new Error('Cannot write to a path outside of Node.js environment.');
   }
 
-  const fs = isNode ? await import('fs') : null;
+  const fs = isNode ? await importFSModule() : null;
 
   let eof = false;
   let fileHandle: import('fs').promises.FileHandle;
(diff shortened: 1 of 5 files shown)


jackfranklin:
Note that I'm expecting the 10.15 CI run to fail, but hoping that I can then patch https://github.com/puppeteer/puppeteer/pull/6549 in and that will fix it.

mathiasbynens (approved):

jackfranklin:
@mathiasbynens PTAL at the latest commit. I took the work in #6549 but extracted it into a helper function that we can use in the mean time. That, plus running tests on Node 10.15, should ensure we don't hit this again.

src/common/helper.ts
```diff
@@ -344,6 +344,31 @@ async function readProtocolStream(
...
+async function importFSModule(): Promise<typeof import('fs')> {
+  if (!isNode) {
+    throw new Error('Cannot load the fs module API outside of Node.');
+  }
+
+  const fs = await import('fs');
+  if (fs.promises) {
+    return fs;
+  } else {
+    return fs.default;
```
mathiasbynens: nit: no else after return (we should probably enable the eslint rule for it)

mathiasbynens (approved):

Issue #6482, opened 2020-10-08
Simplify & automate the release process


mathiasbynens:
- [x] Step 0.5: automate `npm publish` when pushing new Git tags (https://github.com/puppeteer/puppeteer/pull/6536; [example run](https://github.com/puppeteer/puppeteer/runs/1315478407))
- [x] Step 1: enforce Conventional Commits (https://github.com/puppeteer/puppeteer/pull/6483)
- [x] Step 2: add a changelog (#6556)
- [x] Step 3: automate version bumps (incl. changelog, docs, etc.) (#6627 + #6628)

**Note:** Although step 0.5 is not part of the longer-term plan, it’s a relatively small change that already reduces the amount of manual steps in the release process significantly in the mean time.

mathiasbynens:
The release process is now mostly automated 🎉 

Before:
![](https://user-images.githubusercontent.com/81942/108219681-41bcd880-7136-11eb-84bd-7a14d834eab6.png)

After:
![](https://user-images.githubusercontent.com/81942/108219727-48e3e680-7136-11eb-88fb-451bd89c7b21.png)

There’s still a few manual steps (creating and merging PRs), but I believe we’ve reached diminishing returns here. We _could_ take this even further, by automating the _entire_ release process to the point where we can cut a release just by merging an automatically-created/updated PR, but that likely is more work than we’d save. So for now, I'm happily closing this 👍

PR #6556, opened 2020-10-29
chore: automate versioning + changelog

mathiasbynens:
As long as we follow Conventional Commits for our commit messages (which is now enforced since #6483), we can automate the maintenance of this new changelog, which enables us to later automate the release process altogether.

This patch also eliminates the versioning decision-making process by automating it based on our commit messages.

Issue: #6482


package.json
```diff
@@ -105,5 +107,11 @@
     "hooks": {
       "commit-msg": "commitlint --env HUSKY_GIT_PARAMS"
     }
+  },
+  "standard-version": {
+    "skip": {
+      "commit": true,
+      "tag": true
```
mathiasbynens, line 114: Note: the eventual goal is for standard-version to handle creating the release commits + tags as well, but this requires some more follow-up work.

johanbay (approved):
Yay! we are getting close to having this fully automated!

PR #6557, opened 2020-10-29
docs(api): expand registerCustomQueryHandler example

mathiasbynens:
This patch makes it easier to see exactly how to use the registerCustomQueryHandler API without having to follow the trail of breadcrumbs throughout the docs.


johanbay (approved):
Good idea! Thanks for fixing the formatting and typos as well!

PR #6562, opened 2020-10-30
fix(domworld): fix missing binding for waittasks

johanbay:
This patch fixes a flake where a `WaitTask` tries to use a binding that has not been set up.
The issue was that from the time we added the binding to we ran the waittask, the execution context could have changed due to e.g navigation.
This patch moves the responsibility for adding the binding from the `AriaQueryHandler`/`DOMWorld` to the `WaitTask` itself. The task now asks the `DOMWorld` to add the binding in a specific execution context and then execute the `waitForPredicate` function in the same execution context. This way, we know that the execution context must have the binding before we evaluate in it.

### DIFF
--- a/src/common/DOMWorld.ts
+++ b/src/common/DOMWorld.ts
@@ -62,6 +62,14 @@ export interface WaitForSelectorOptions {
 /**
  * @internal
  */
+export interface PageBinding {
+  name: string;
+  pptrFunction: Function;
+}
+
+/**
+ * @internal
+ */
 export class DOMWorld {
   private _frameManager: FrameManager;
   private _frame: Frame;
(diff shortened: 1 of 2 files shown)


src/common/DOMWorld.ts
```diff
@@ -482,19 +489,23 @@ export class DOMWorld {
...
    * @internal
    */
-  async addBindingToContext(name: string) {
+  async addBindingToContext(
+    context: ExecutionContext,
+    name: string
+  ): Promise<void> {
     // Previous operation added the binding so we are done.
-    if (this._ctxBindings.has(name)) return;
+    if (this._ctxBindings.has(name + context._contextId)) {
```
johanbay: Instead of just keeping track of the names we have set up bindings for, we remember `binding + contextId`.
At least theoretically, I think we could otherwise end up in a situation where we added the binding just as the context was changed. If so, we could end up putting the name in `_ctxBindings` even though it isn't added in the new context.

mathiasbynens: Currently, we get the same key for:

```
name = 'abc'
contextId = 'def'
```

and 

```
name = 'ab'
contextId = 'cdef'
```

...which seems potentially bad. I think we should we do something like

```
`${name}_${context._contextId}`
```

That is, always separate `name` and `contextId` by _something_ (not necessarily `_` — ideally some character that cannot occur in `contextId`s) to avoid collisions between overlapping characters.

mathiasbynens (changes requested):

src/common/DOMWorld.ts
johanbay: Good idea! I've added a `_` separator. The `contextId`s are numbers, so `_` should do.

mathiasbynens (approved):

jackfranklin (approved):

PR #6574, opened 2020-11-03
chore: mark launcher spec as failing on Windows

jackfranklin:
We've pushed this date back a lot, let's make it permanent until we have
more time to investigate.


jackfranklin:
@mjzffr just FYI, I've removed the date limit on the test as we've pushed it back a bunch of times. Seems worth permanently disabling it until we've more time to properly dig into it 👍

mathiasbynens (approved):

PR #5008, opened 2019-10-05
Support promises for page.waitForResponse predicate

josalmi:
Fixes #4323.

PR #6624, opened 2020-11-25
feat: support promise as return value for page.waitForResponse predicate

### LINKED ISSUE #4323: Can waitForResponse take a promise instead of function or url?

* Puppeteer version: 1.14
* Platform / OS version:  Windows
* Node.js version: 10

try to pass a promise to `await page.waitForResponse(response => condition)`  instead of `urlOrPredicate`.

expect async function to work

**What happens instead?**
no waiting as promise gets eval to truthy. 

response object has text method which is a promise to use that i need callback to be promise as well.. but if i make callback to be promise it breaks the waitForResponse function

mickykebe:
Fix #4323 

Adaptation of #5008 with example code added to the docs.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2255,6 +2255,7 @@ return firstRequest.url();
 ```js
 const firstResponse = await page.waitForResponse('https://example.com/resource');
 const finalResponse = await page.waitForResponse(response => response.url() === 'https://example.com' && response.status() === 200);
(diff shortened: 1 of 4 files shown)


mathiasbynens:
Let's add

```
Co-authored-by: Joni <josalmi@google.com>
```

to the commit message, to credit the author of #5008 (@josalmi).

mathiasbynens (approved):

PR #6627, opened 2020-11-26
chore: automate version bumps

mathiasbynens:
This patch removes another manual step from the release process.

Issue: #6482

### DIFF
--- a/utils/doclint/preprocessor/index.js
+++ b/utils/doclint/preprocessor/index.js
@@ -16,6 +16,8 @@
 
 const Message = require('../Message');
 
+const IS_RELEASE = Boolean(process.env.IS_RELEASE);
+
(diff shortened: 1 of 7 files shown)


OrKoN (approved):

utils/doclint/preprocessor/index.js
```diff
@@ -34,9 +36,6 @@ module.exports.ensureReleasedAPILinks = function (sources, version) {
 };
 
 module.exports.runCommands = function (sources, version) {
-  // Release version is everything that doesn't include "-".
-  const isReleaseVersion = !version.includes('-');
```
mathiasbynens, line 38: Hmm, actually I think this would now cause problems on CI. We need to keep this check and combine it with the new environment variable. Will fix.

mathiasbynens: Done

PR #6646, opened 2020-12-03
bidiClient

foolip:
- Add first version of WebDriver BiDi server (#3)
- Add Python code to connect to the BiDi server and say hello


foolip:
Oops, this was supposed to be a PR against https://github.com/sadym-chromium/WebDriverBiDiServerExperiments. `gh pr create -f` realized that repo was a fork and did this instead.

PR #6650, opened 2020-12-04
Test changes as result of Mozilla's sync up with v5.5.0

mjzffr:
We just did a sync of our Puppeteer copy in mozilla-central and there are some test changes that came up as a result.

### DIFF
--- a/test/emulation.spec.ts
+++ b/test/emulation.spec.ts
@@ -258,7 +258,7 @@ describe('Emulation', () => {
     it('should work', async () => {
       const { page } = getTestState();
 
-      page.evaluate(() => {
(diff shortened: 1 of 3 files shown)


test/emulation.spec.ts
```diff
@@ -261,7 +258,7 @@ describe('Emulation', () => {
     it('should work', async () => {
       const { page } = getTestState();
 
-      page.evaluate(() => {
+      await page.evaluate(() => {
```
mathiasbynens, line 261: Oof, good catch

mathiasbynens (approved):
Nice!

PR #6692, opened 2020-12-22
fix(launcher): rename TranslateUI to Translate to match Chrome

paulirish:
same as https://github.com/GoogleChrome/chrome-launcher/pull/225

> TranslateUI isn't a thing as of several months ago ([chromium-review.googlesource.com/c/chromium/src/+/2404484](https://chromium-review.googlesource.com/c/chromium/src/+/2404484)).


mathiasbynens (approved):

mathiasbynens:
Thanks! Do you need this in a release urgently?

PR #6703, opened 2020-12-27
Fixed misspelling in v5.5.0 text and link in API docs

ownchoice:


mathiasbynens:
Good catch! The same change needs to be made in `versions.js`. Could you please go ahead and do that as well?

jschfflr:
This problem has already been resolved in https://github.com/puppeteer/puppeteer/pull/6778

PR #6731, opened 2021-01-06
chore(agnostification): inject dependecies(ws, fs, path etc) for common in entry(node, web) initialization.

webee:
related to #6125
by add a Config module which contains the node specific packages(ws, fs etc), let the common depend on it(dependeny injection), we can remove isNode check in common code, as well as dynamic imports.
every entry(node, web) needs initialize the Config before export the Puppeteer object.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,8 @@
 {
-  "name": "puppeteer-core",
+  "name": "puppeteer",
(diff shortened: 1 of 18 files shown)


webee:
@googlebot I signed it!

PR #6732, opened 2021-01-06
fix(common): ensure target exists

Kikobeats:
Hello,

This PR handles the situation where you are trying to create a new page in a crashed browser.

Stack trace related: https://sentry.io/share/issue/5cc5151149fd4bb18126dd470311f4ed/

I'm not sure if the `target` check should be done as well at:

- https://github.com/puppeteer/puppeteer/blob/87fe99a54ae5b9e7eff541b9ce794adc2d746e4a/src/common/Browser.ts#L297
- https://github.com/puppeteer/puppeteer/blob/87fe99a54ae5b9e7eff541b9ce794adc2d746e4a/src/common/Browser.ts#L308


mathiasbynens:
Could you add a test?

Kikobeats:
@mathiasbynens yes, although I'm not familiar with puppeteer testing scaffolding.

Maybe can you point me where is the current test for `target._initializedPromise`, then I can extend it?

mathiasbynens:
https://github.com/puppeteer/puppeteer/blob/main/test/target.spec.ts seems like a reasonable place to add the new test. You can probably use `chrome://crash` in the test to trigger a crash. Make sure the test fails without your change, and succeeds with it.

Kikobeats:
@mathiasbynens I tried to add the test but it was pretty hard for to me keep the project running on my local :(

mathiasbynens:
@jschfflr Can you help @Kikobeats?

jschfflr:
Yeah sure, happy to help! 
What do you mean by:
> @mathiasbynens I tried to add the test but it was pretty hard for to me keep the project running on my local :(

Could you elaborate on the challenges you are facing?

Kikobeats:
@jschfflr essentially still I need to add the test to finally land this

jschfflr:
Let me know if I can help with the tests!

Kikobeats:
I tried to write a test for reproducing the issue:

```js
const { page, browser } = getTestState();
page.goto('chrome://crash').catch(() => {});
await waitEvent(page, 'error');
process.kill(browser.process().pid, 'SIGKILL')
console.log('creating page')
const newPage = await browser.newPage()
```

but even in the worst scenario looks like the internal `target._initializedPromise` is present.

Also looks like the original error traced expired, so I'm going to close for now and revisiting it in the future if needed.

jschfflr:
@Kikobeats Thanks for updating the bug. Feel free to open it again if the problem comes up again!

Issue #5334, opened 2020-01-22
UnhandledPromiseRejectionWarning: Error: Request is already handled!


giaphiep:
I trying use puppeteer to crawler but I have a problem when I call request to "Kick out login!". someone can help me!
thanks

`const puppeteer = require("puppeteer");
( async () => {
    const browser = await puppeteer.launch()
    const page = await browser.newPage()
    await page.setRequestInterception(true)
    await page.setUserAgent(process.env.USER_AGENT)

    page.on('request', interceptedRequest => {
        interceptedRequest.continue({
        method: 'POST',
        postData: 'account=' + process.env.EMAIL + '&password=' + process.env.PASSWORD,
        headers: {
            ...interceptedRequest.headers(),
            'Content-Type': 'application/x-www-form-urlencoded',
            'Cookie': process.env.COOKIE_FAKE
            }
        })
    })
    const response = await page.goto(process.env.LOGIN_URL)
    const body = await response.json()
    const token = body.result.token
    // return
    if (body.result.statusCode !== 200) {
        return
    } else if (body.result.type === 'login') {
        console.log('Login success!')
        // do something
    }
    else if (body.result.type === 'overLogin') {
        console.log('Kick out login!')
        page.on('request', interceptedRequest => {
            interceptedRequest.continue({
            method: 'POST',
            postData: 'account=' + process.env.EMAIL + '&token=' + token,
            headers: {
                ...interceptedRequest.headers(),
                'Cookie': process.env.COOKIE_FAKE
                }
            })
        })

        const response = await page.goto(process.env.LOGOUT_URL)
        let body = await response.json()

        if (body.result.statusCode !== 200) {
            return
        } else if (body.result.type === 'login') {
            console.log('Login success!')
            // do something
        } else {
            return
        }
    }
})()`

darkguy2008:
Same here, still waiting for a fix...

mesoar:
Same here, still waiting for a fix......

luka-dev:
I have the same issue

davidoskay:
Same

delmotte:
Same, any update on this ?

mclimb:
I have the same issue as well. I uninstalled all of my puppeteer-related packages and then reinstalled each and then it worked again for a short time. Then it stopped working again... ugh

NealBozeman:
The package puppeteer-extra-plugin-block-resources is also failing with the same error. I can't find any way to block resources using the API directly.

[...]

Issue #4524, opened 2019-06-05
when multi request listeners added, request.continue() works in a confusing way


xiyuan-fengyu:
following code shows my question:
```typescript
import {launch, Request} from "puppeteer";

(async () => {
    const browser = await launch({
        headless: false,
        devtools: true
    });
    const page = await browser.newPage();
    await page.setViewport({width: 1920, height: 1080});
    await page.setRequestInterception(true);
    page.on("request", async (req: Request) => {
        await req.continue();
    });
    page.on("request", async (req: Request) => {
        await req.respond({
            status: 200,
            body: "test"
        });
    });
    await page.goto("https://www.baidu.com/");
    console.log();
})();
```
req.respond won't work in the second request listener and just throw an error.  
If multi request listeners added, how to decide whether request.continue should be called is confusing.  
  
In my way, I solve it as following  
```typescript
import {launch, Request} from "puppeteer";

(async () => {
    const browser = await launch({
        headless: false,
        devtools: true
    });
    const page = await browser.newPage();
    await page.setViewport({width: 1920, height: 1080});
    await page.setRequestInterception(true);

    page.on("request", (req: Request) => {
        const oldContinue = req.continue;
        let continueNum = 0;
        // override request.continue
        req.continue = (override?: any) => {
            if (override) {
                return oldContinue.call(req, override);
            }
            else {
                continueNum++;
                if (continueNum == page.listeners("request").length) {
                    return oldContinue.call(req);
                }
            }
        };
        return req.continue();
    });

    page.on("request", async (req: Request) => {
        await req.continue();
    });
    page.on("request", async (req: Request) => {
        await req.respond({
            status: 200,
            body: "test"
        });
    });
    await page.goto("https://www.baidu.com/");
    console.log();
})();
```

Is there a more elegant way to solve it?

aslushnikov:
@xiyuan-fengyu in general, you should have only one place in code where you decide on request handling. Why do have multiple request event handlers?

xiyuan-fengyu:
In my crawler framework，I need to supply some method to help user to disable image load, or proxy request for each page, or record request, and so on, all those need request interception.   
[PuppeteerUtil.setImgLoad PuppeteerUtil.useProxy](https://github.com/xiyuan-fengyu/ppspider/blob/master/src/puppeteer/PuppeteerUtil.ts)  
[...]

Issue #3853, opened 2019-01-28
Feature request: possibility to handle `page.on('request',...)` in several places


yanivefraim:
Use case:

I am developing a testkit which does `page.on('request', ...)` in order to intercept specific requests. 
The problem: A user can write another interceptor which handles specific requests and calls `request.continue()` for all the rest (as this is mandatory for requests that we do not handle).

As a result, only the first request handler "wins" (the second handler will receive "request is already handled!" error.)

A possible solution: 
Today we must have `request.continue()` for every request that we do not want to handle. What if puppeteer will do it by default, for requests that it finds as "unhandled". In this case, each handler will only take care of its own interceptors, and will do nothing about the rest of requests.

aslushnikov:
> What if puppeteer will do it by default, for requests that it finds as "unhandled".

@yanivefraim With the currently-exposed API, it's unclear when the request is "unhandled". Consider an artificial example of "network throttling" using request interception:

```js
page.on('request', request => {
  setTimeout(request.continue.bind(request), 1000);
});
```

Since Puppeteer gives away control over the request to the client, client might continue them anytime in future as they require. They might even drop the request on the floor without continuing to simulate server stalling. So from Puppeteer's stand point, we can't auto-continue requests since it will break many valid usecases.

However, it looks like your testkit introduces certain conventions over request interception, e.g. "request should be either synchronously aborted, or it will be auto-continued" (i might got it wrong). In this case, you can auto-continue all requests in a subsequent microtask:

```js
page.on('request', request => Promise.resolve().then(() => request.continue()).catch(e => {}));
```

Let me know if I understand you correctly and if it helps.

yanivefraim:
@aslushnikov Thanks for the fast response. We will take it further from here.

b5l:
Puppeteer could implement a public API to check whether a request is already handled or not. There is already the `_interceptionHandled` property on the `Request` class, we would essentially just need a getter for that. Any thoughts? @aslushnikov

NinoSkopac:
Aslushnikov's "workaround" (i don't know what it is) works.

```js
page.on('request', request => {
  console.log(request.method(), request.url()); // Basically, do your work here
  
  return Promise.resolve().then(() => request.continue()).catch(e => {});
})
```

@aslushnikov Could you please explain why this works?

whalderman:
> Aslushnikov's "workaround" (i don't know what it is) works.
> 
> ```js
> page.on('request', request => {
[...]

Issue #2687, opened 2018-06-06
AssertionError when intercepting requests on same url


lsbyerley:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.4.0
* Platform / OS version: MacOS High Sierra
* URLs (if applicable): any
* Node.js version: 8.9.4

**What steps will reproduce the problem?**
I have a pool of puppeteer browsers running (a minimum of 2). I recently added a request interceptor to filter out ad urls. When performing multiple requests to the same url, usually happens on the 3rd time, an AssertionError is generated for all filtered requests: `AssertionError [ERR_ASSERTION]: Request is already handled!`. The strack trace points to the `interceptedRequest.continue()` line of my `page.on('request', (interceptedRequest) => {})` function in the api route below. Further on in the strack trace points to `node_modules/puppeteer/lib/NetworkManager.js:220:10` and `node_modules/puppeteer/lib/NetworkManager.js:169:12` while also showing an unhandled promise rejection error

_Please include code that reproduces the issue._

// api route using the pool

    const browserPagePool = require("../utils/browserPagePool");
    
    router.get("/api/test", async (req, res, next) => {
      try {
    
        const page = await browserPagePool.acquire();

        await page.setRequestInterception(true);
        page.on('request', (interceptedRequest) => {
          if ( isAdUrl(interceptedRequest.url()) ) {
            interceptedRequest.abort();
          } else {
            interceptedRequest.continue();
          }
        });
    
        await page.goto(url, { waitUntil: "networkidle0", timeout: 10000 });
        const content = await page.content()
        await browserPagePool.release(page); 
  
        return res.status(200).json(content);
    
      } catch (err) {
        console.error(err);
      }
    });

// browserPagePool.js

    const genericPool = require("generic-pool");
    const puppeteer = require("puppeteer");
    const browserPromise = async () => await puppeteer.launch({ headless: true });
    
    const factory = {
      create: async function() {
        const browser = await browserPromise();
        return browser.newPage();
      },
      destroy: function(puppeteer) {
        return puppeteer.close();
      }
    };
    
    const browserPagePool = genericPool.createPool(factory, {
      max: 10,
      min: 2,
      maxWaitingClients: 50
    });
    
    module.exports = browserPagePool;

Should I somehow be clearing out previous requests from an already used puppeteer pool, or checking if a request is already completed in the interceptedRequest function, or something else? I'm not sure how to gracefully handle this.

[...]

PR #6733, opened 2021-01-06
feat: Cooperative request intercepts

benallfree:
Hello, I wanted to open a discussion via this PR about a solution for cooperative request intercepts. I needed this for a recent project and thought it might help to share my fork.

If this PR seems too aggressive to be a core change, my other thought is https://github.com/puppeteer/puppeteer/pull/6734, which is just the modification to allow Puppeteer to accept pluggable `NetworkManager` and `HTTPResponse` classes. Then I can maintain a separate package with these more extensive changes.

# The Problems

## Puppeteer's core expects only one `page.on('request')` handler

Puppeteer anticipates only one request intercept handler (`page.on('request', ...)`) to perform `continue()`, `abort()`, or `respond()`. As the community has pointed out in multiple issues, there are often cases where multiple request handlers make sense. For example, `puppeteer-extra` attempts a plugin architecture. The current core design makes it impossible for multiple request handlers to cooperatively determine what to do with a request.

Here is a sample of issues for background:

https://github.com/berstend/puppeteer-extra/issues/364
https://github.com/berstend/puppeteer-extra/issues/156
https://github.com/smooth-code/jest-puppeteer/issues/308
https://github.com/puppeteer/puppeteer/issues/5334
https://github.com/puppeteer/puppeteer/issues/3853#issuecomment-458193921
https://github.com/puppeteer/puppeteer/issues/2687
https://github.com/puppeteer/puppeteer/issues/4524

## Puppeteer's request intercept handler does not wait for async handlers to finish

Request intercept handling currently does not support async intercept handlers. The underlying `mitt` event system does not await async events, and `NetworkManager` just fires the events and moves on:

```typescript
    this.emit(NetworkManagerEmittedEvents.Request, request);
```

# The solutions

The solution this PR proposes a cooperative interception strategy while also making `NetworkManager` wait for async handlers to be fulfilled before finalizing the request interception.

One caveat to a cooperative interception strategy is deciding which directive should 'win'. I decided upon `abort > respond > continue`

```typescript

// This handler will 'win' because it asks for an abort(). The request will be aborted.
page.on('request', req=> {
  req.abort()
});

// Had there not been an abort(), this handler would 'win' because a respond() is better than a continue()
page.on('request', req=> {
  req.respond({...})
});

// This is the lowest priority. The request will be continued only if no abort() or respond()
page.on('request', req=> {
  req.continue() // Not even necessary, NetworkManger will fall through to continue() by default
});

// Async example. NetworkManager will not fulfill the request until this deferred operation has been completed
page.on('request', req=> {
  req.defer( async ()=> {
      // do something async like a database lookup
  })
});
```

# Possible breaking changes

`HTTPRequest.abort()`, `HTTPRequest.continue()` and `HTTPRequest.respond()` are no longer return promises. Their behavior has changed because they no longer await resolution of the underlying the [CDP Fetch events](https://chromedevtools.github.io/devtools-protocol/tot/Fetch/). 

I don't think they should anymore, because they are merely cooperative 'recommendations' about what to do. We could make these still return a promise, but then we'd have to decide what to do if the request interception ultimately resolved differently. For example, if you did `await continue()` but the request actually aborted, what should happen? Probably an exception? I think it is cleaner to make these return nothing, and if needed, introduce a separate `HTTPRequest.fulfilled()=>Promise<ResolutionType>` signal.

### DIFF
--- a/src/common/HTTPRequest.ts
+++ b/src/common/HTTPRequest.ts
@@ -114,6 +114,13 @@ export class HTTPRequest {
   private _postData?: string;
   private _headers: Record<string, string> = {};
   private _frame: Frame;
+  private _interceptionRequestOverrides: ContinueRequestOverrides;
+  private _interceptionResponse: ResponseForRequest;
+  private _abortErrorReason: ErrorReasons;
+  private _shouldContinue: boolean;
+  private _shouldRespond: boolean;
+  private _shouldAbort: boolean;
+  private _deferredRequestHandlers: DeferredRequestHandler[];
 
   /**
    * @internal
@@ -124,7 +131,8 @@ export class HTTPRequest {
     interceptionId: string,
     allowInterception: boolean,
     event: Protocol.Network.RequestWillBeSentEvent,
-    redirectChain: HTTPRequest[]
+    redirectChain: HTTPRequest[],
+    deferredRequestHandlersRef: DeferredRequestHandler[]
   ) {
     this._client = client;
     this._requestId = event.requestId;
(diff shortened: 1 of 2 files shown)


benallfree:
@googlebot I signed it!

benallfree:
Note: moved to https://github.com/puppeteer/puppeteer/pull/6735

Issue #6696, opened 2020-12-25
Request interception causes web fonts pending forever


mstr-pezhang:
### Steps to reproduce

* Puppeteer version: 5.5.0
* Platform / OS version: Windows 10, Ubuntu 18, MacOS Big Sur
* URLs (if applicable): 
* Node.js version: 12.19.0

**What steps will reproduce the problem?**

Example Script: https://github.com/mstr-pezhang/puppeteer-issue/blob/36cddcc721ec0cf228f213f72dce2addcd5e223d/test/webfont.spec.ts

1. Create a font-face like https://github.com/mstr-pezhang/puppeteer-issue/blob/36cddcc721ec0cf228f213f72dce2addcd5e223d/static/global.css
2. In the webpage, do not directly use the web font, but generate a DOM node to use it. https://github.com/mstr-pezhang/puppeteer-issue/blob/36cddcc721ec0cf228f213f72dce2addcd5e223d/static/add-text.js
3. Enable request interception. For each request, simply call `request.continue();`.
3. Open DevTools and go to Network tab.
4. See that the font cannot be correctly rendered because the URL is pending forever.

**What is the expected result?**
The web font can be successfully loaded and rendered.

**What happens instead?**
The default system font is used instead, while the web font is pending forever.
https://github.com/mstr-pezhang/puppeteer-issue/actions/runs/443832997

I'm not sure if it is a Chromium issue, but there are similar issues like #4208 and #5003

mstr-pezhang:
A [workaround](https://github.com/mstr-pezhang/puppeteer-issue/commit/e9bcd73e25f79d3e12f921e1b81cea591d73829b) is to enable request interception before loading the web page.
More requests will be tracked in this way, but the web font can be loaded.

dobesv:
In our project we are setting request interception on as right after calling `newPage()` and before opening any pages but we are still having this problem.

benallfree:
@mstr-pezhang Can you please verify that #7802 fixes this issue?

slawiko:
@benallfree probably that could help:

I had similar issue with Font request pending for more than 30sec (I had timeout 35s). After upgrade on 13.0.0 this issue is completely gone

mstr-pezhang:
> @mstr-pezhang Can you please verify that #7802 fixes this issue?

Thanks!
Tested puppeteer v12.01, v13.0.1 and v13.2.0.
Now the issue is gone.

rudolfbyker:
[...]

Issue #6567, opened 2020-11-01
Puppeteer Extensions (request handlers, custom methods, ...)


XavierGeerinck:
Hi Everyone!

I'm currently loving puppeteer and am using it daily! However while working with it - as with any product - it seems that my codebase is getting larger and larger making it quite hard to keep a correct overview 😊

Therefore I started using an abstraction layer on top of puppeteer called [`puppeteer-extra`](https://github.com/berstend/puppeteer-extra/) but am starting to think more and more that this should be in the official puppeteer repo for obvious extensibility reasons.

The most crucial parts that required extensibility for me while working with puppeteer are:
* Request handlers -> I would love to create multiple request handlers that each do their own thing (e.g. blocking resources, waiting for XHR calls in a custom way, ...)
* Custom methods -> I would love to add custom methods that allow me to for create simplified methods (e.g. clickAndWait, ...)

Is this something that was answered or thought about already? Would it make sense to start integrating extension possibilities within puppeteer? I'd love to hear your thoughts around this!

jackfranklin:
Thanks for this issue and sorry for the lack of replies. I'll leave request handlers to #6735 as that's an ongoing discussion.

RE custom methods, it's not something we've really thought about to be honest. On projects we have that use Puppeteer (we write DevTools e2e tests with Pptr, for example) we have a lot of custom helpers that take the `page` object in:

```
function clickAndWait5Seconds(page: Page) {
  ...
}
```

Whilst that isn't quite as nice as defining custom methods, it's a nice way to build up a library of helpers and means we don't have to add the ability to define custom methods into pptr itself. It also means a developer jumping between two projects that use Pptr won't have to remember which methods are custom vs baked into the library itself.

PR #6735, opened 2021-01-06
Cooperative request intercepts

benallfree:
Hello, I wanted share my fork with the Puppeteer team to see if there is anything here you'd like to bring into the core.

This PR introduces request intercept handlers that play well with others. I needed this for a recent project and thought it might help to share my fork because it seems that there is a lot of community confusion around the way request intercepts are handled.

# The Problem: Puppeteer's core expects only one `page.on('request')` handler to call `abort/respond/continue`

Puppeteer anticipates only one request intercept handler (`page.on('request', ...)`) to perform `continue()`, `abort()`, or `respond()`. There are valid use cases where multiple request handlers may call `abort/respond/continue`, unaware of each other. For example, `puppeteer-extra` attempts a plugin architecture, yet enabling any plugin that utilizes request intercepts (such as ad blocker) means that I cannot further alter request interceptions because calling `abort/respond/continue` a second time will throw an error. 

The current core design makes it impossible for multiple request handlers to cooperatively determine what to do with a request.

Here is a sample of issues for background:

https://github.com/berstend/puppeteer-extra/issues/364
https://github.com/berstend/puppeteer-extra/issues/156
https://github.com/smooth-code/jest-puppeteer/issues/308
https://github.com/puppeteer/puppeteer/issues/5334
https://github.com/puppeteer/puppeteer/issues/3853#issuecomment-458193921
https://github.com/puppeteer/puppeteer/issues/2687
https://github.com/puppeteer/puppeteer/issues/4524

# The solution: Cooperative Request Interception mode

This PR proposes a cooperative interception strategy which allows multiple request intercept handlers to call `abort/respond/continue` in any order and multiplicity, while also making `NetworkManager` wait for async handlers to be fulfilled before finalizing the request interception.

```typescript

// 2nd param enables Cooperative mode rather than Legacy mode
// In Cooperative mode, abort/respond/continue can be called multiple times by multiple handlers.
// The winning action is decided after all handlers have resolved.
page.setRequestInterception(true, true) 

// If any handler calls abort(), the request will be aborted.
// All handlers will still run, but ultimately the request will be aborted because at least one handler
// requested an abort.
page.on('request', req=> {
  req.abort()
});

// Had no handler called abort(), but at least one handler called respond(),
// then the request will be responded.
page.on('request', req=> {
  req.respond({...})
});

// This is the lowest priority. continue() is the default action and no handler even needs to call it explicitly
// unless request info needs to be changed. A request will be continued only if no abort() or respond()
// has been called by any handler.
// Cooperative mode is more forgiving because no request will hang since continue() is the default action.
page.on('request', req=> {
  req.continue() // Not even necessary, NetworkManger will fall through to continue() by default
});
```

# Breaking changes

None. Legacy intercept mode is still the default. A new "Cooperative Intercept Mode" is introduced by a second `Page.setRequestInterception` parameter.

# Todo

- [ ] Document cooperative mode
- [ ] Add unit tests
- [ ] Fix failing tests/regressions


whalderman:
There's no reason to pull these breaking changes when a valid workaround already exists, as you referenced ( #3853 ).
This is a solvable issue within the plugin community.

benallfree:
@whalderman Thanks for looking at this!

#3853 doesn't solve the problems I mentioned above, it only defers one handler's intercept resolution until all synchronous handlers have run. But you might be right that the plugin community could agree to install ONE handler that does the same thing my PR proposes. Let me take a shot at that and get back to you.

benallfree:
I have published a community package (https://www.npmjs.com/package/enchant-puppeteer) that monkey-patches `NetworkManager` and `HTTPRequest` to cooperatively process multiple request handlers with asynchronous executions. 

IMHO this ought to be a core feature. It can be done with no breaking changes and would significantly improve interception capabilities.

I would like to hear an official thumbs-up/down from a Puppeteer maintainer before closing this issue. @whalderman forgive me if you are maintainer, I didn't see your name in the contributor list.

XavierGeerinck:
I +1 on @benallfree his comment, this is a crucial feature that I have mentioned as well already in #6567

jackfranklin:
@benallfree thanks for your efforts here and for kicking off the discussion.

I've read through your PR and I _think_ I follow, but what I'm struggling with in my head is fully understanding a situation where the limitation of having one request handler causes problems. Is there a small code example that you could share? I just don't think that's landed in my head yet so I'm struggling to form an opinion and would like to make sure I understand fully - thanks.

benallfree:
@jackfranklin Thanks for writing back and for giving this feature honest consideration 👍 

Please consider this use case of two request handlers needing to work cooperatively: 

The package @cliqz/adblocker contains a Puppeteer binding that intercepts Puppeteer requests and conditionally blocks them by calling `req.abort()` if they appear to be related to ads. Otherwise, it calls `req.continue()`. https://github.com/cliqz-oss/adblocker/blob/master/packages/adblocker-puppeteer/adblocker.ts#L276

AdBlocker makes for a nice plug-in, except using it precludes any other request intercept handlers from further manipulating the request. Once `abort()` or `continue()` are called by any handler, calling them again has no effect and in fact throws an error. 

Suppose I wanted to use AdBlocker, but I also wanted to further intercept requests not blocked by AdBlocker and return an alternate response stored in Redis. This is my real-world use case and how I found this problem.

```typescript
page.setInterception(true)

// Initialize ad blocker
// Internally, this registers a handler with page.on('request') and calls a req.abort() or req.continue()
initializeAdBocker(page) 

// Now I would like to add my own caching intercept that that either calls respond()
// or continue(). The problem is, by the time my handler runs, AdBlocker has already
// sealed our fate. There is no way to call respond() from here.
//
// Moreover, because this is an async function, the continue() will not
// be called until the next event loop. Any synchronous handler will "win" this race.
//
// By adding a second `true` to Page.setInterception(true,true), we can enable
// Cooperative Intercept Mode and this use case will work with no other changes.
page.on('request', async req=>{
   const cachedResponse = await lookUpCachedResponseInRedis(req.url())
   if(cachedResponse) {
      req.respond(cachedResponse)
      return
   }
   req.continue()
})

await page.goto('...')
```

The modifications I propose in this PR create no breaking changes. Legacy behavior is preserved by default, and the new "Cooperative Intercept Mode" can be enabled for those who want to use it. In cooperative mode, it possible for any number of handlers to call `abort()`, `respond()`, and `continue()` without throwing errors. After all the handlers have run and all asynchronous handlers have resolved, it will decide which action wins: if any handler called `abort()`, the response will be aborted. If no handler called `abort()` but any handler called `respond()`, the request will be responded. And then the default is that the request will be continued.

jackfranklin:
@benallfree thank you, that's made it super clear and I think I understand the problem now.

I like the fact we could land such a change without needing to introduce breaking changes to the API. 

@mathiasbynens what do you think about this?

benallfree:
@jackfranklin Just checking back, thank you 👍

mathiasbynens:
@jschfflr and @sadym-chromium, could you PTAL?

jschfflr:
I also see a lot of value in this approach to handling request interception. Especially in the context of third party code that might be intercepting requests too. 
But have you also considered the possibility to run the handlers in sequence instead of in parallel? I'm thinking of something similar to how express handles network requests. The advantage would be that it's more explicit why a certain request was handled in a certain way. Otherwise a developer might for example be confused why a request they chose to modify via `response` was aborted. 
Let me know what you think!

benallfree:
@jschfflr Thanks for the feedback! 

I have considered the same things, great discussion.

Synchronous handlers run in series because the `Mitt` event library [fires them in the order they were subscribed](https://github.com/puppeteer/puppeteer/blob/713e990531bf2b898066a5ffca50f0b130a4ad4f/vendor/mitt/src/index.ts#L80). But if those handlers perform asynchronous operations, Mitt does not await them. I could take a shot at modifying Mitt to accept async handlers and await them in order. I agree that more determinism could be valuable. 

On the other hand, there could be significant performance penalties associated with serializing async handlers. We also don't currently have a way of prioritizing or introspecting `page.on('request')` handlers so it would be hard to determine the order in which handlers are registered and executed anyway. (Especially if they are registered via plugins or even transient dependencies). But to your point, in theory two ill-behaved async handlers could find themselves in occasional race conditions that are difficult to debug if they are allowed to execute in parallel.

Well-written handlers shouldn't assume anything about what other handlers might be doing, or how state may have changed during an `await`. I added the `shouldAbort()`, `shouldRespond()`, and `shouldContinue()` helpers address this for interested handlers. If it's important that my handler behaves differently based on what may have happened already, I would check these states and react accordingly. Likewise, I can get the existing continuation or response data and modify it via `continueRequestOverrides()` or `respondForRequest()`. 

Overall, I'm leaning toward allowing parallel, non-deterministic execution, but the express community may have gotten this one right. Let me know what you think! 👍

benallfree:
@jschfflr Any further thoughts? I've been thinking more about your idea of deterministic execution, I might have an idea. Do you think it's worthwhile pursuing?

jschfflr:
I've been thinking a lot about this problem because I'm not sure how to handle the situation best. But I came up with the following requirements:
- Multiple call sites should be able to use request interception without causing race conditions with each other
- When calling `abort` or `respond` after the request has already been aborted or responded to, the call should fail. This makes it a lot easier to reason about the behaviour of the program.
- Requests should not be continued automatically (see https://github.com/puppeteer/puppeteer/issues/3853#issuecomment-458193921)

Having the requests be handled in a deterministic order sounds like the best first step in that direction. From the top of my head we could maybe work with an api that looks like the following:

```js

// Adding the first interception handler will automatically enable interception
// Request handlers will be run in the order they have been registered
// request.continue() will call the next request handler
// request.abort() and request.respond() will prevent further request handlers 
// from being called
page.addRequestInterceptionHandler((request) => {
   // Do something
  request.abort()
})

page.addRequestInterceptionHandler((request) => {
   // Do something
  request.respond()
})

  // Deprecate abort and respond in the on handler
  // But keep current behaviour for now
  // Event handlers will be triggered after the interception handlers
  // even if abort or respond have already been called to make sure
  // this is backwards compatible
})


await page.goto('...')
```

Let me know what you think!

XavierGeerinck:
Why was this closed @sadym-chromium ? I've been looking forward to this one 😊

benallfree:
@sadym-chromium Was this closed in error?

sadym-chromium:
I'm terribly sorry! I closed it by mistake.

benallfree:
@jschfflr Thank you for the awesome thoughtfulness and feedback, I appreciate a more seasoned perspective! It will be a big win for the community if we can create an effective solution in the core.

Taking your points one by one:

### Multiple call sites should be able to use request interception without causing race conditions with each other

Agreed. We already keep an array of handler promises, we just need to make sure they execute serially. I'll add this.

### When calling abort or respond after the request has already been aborted or responded to, the call should fail. This makes it a lot easier to reason about the behaviour of the program.

If I am understanding correctly, it seems like you want to allow `abort/respond/continue` to be called just once. This crux of this PR is proposing exactly the opposite, for good reasons.

Example:

1. Cache busting plugin calls `continue` with an adjusted `max-age` header
2. Proxy plugin calls `continue` to add a custom header
2. Caching plugin calls `respond` if a cached copy from Redis is available, `continue` if not
3. Logging plugin calls `continue`, making no changes
4. Ad Blocker plugin calls `abort`, but only if it's an ad
5. Malware blocker calls `abort` or `respond`

This PR proposes that all 6 plugins get to run, and the fate of the intercept is not known until they have all finished. If AdBlocker or MalBlocker (or both) call `abort`,  the request is ultimately aborted because `abort` always wins.

If they are all written properly, they don't need to have knowledge of each other and order of execution doesn't matter. That is the essence of cooperative mode, and it cannot be accomplished if `abort/respond/continue` are only allowed to be called once. 

All handlers are allowed to suggest how to resolve an interception, including multiple alterations to `respond` and `continue` data. The essence of the use case is precisely that we don't know what other handlers have been installed by 3rd party/community plugins, or what they may have already done.

Even if all that were not true, forcing the user to check these conditions before taking its own action creates unnecessarily verbose logic:

```ts
if(!req.shouldAbort() && !req.shouldRespond() && !req.shouldContinue()) req.continue()
```

vs

```ts
req.continue() // Here's my handler's vote, let's see what everyone else thinks!
```

### Requests should not be continued automatically (see #3853 (comment))

Good for legacy mode, bad for cooperative mode. In cooperative mode, intentionally hanging a request can be accomplished by simply not resolving the intercept handler promise. Cooperative mode awaits all handlers to resolve their promise wrappers before continuing.

Perhaps this is just a matter of personal opinion, but I think it is disconcerting that requests must be intentionally continued while in intercept mode. It feels more graceful to "fall through" to continuation if the handlers don't do anything else. 

Thank you again for your feedback, looking forward to pinning down requirements!

benallfree:
@jschfflr Checking back on latest thoughts :)

jschfflr:
Hey, sorry it took awhile for me to get back to this. 
I have no strong opinion on doing it one way or the other and I'd be happy to just give the cooperative request interception a try. What do you think about adding this as an experimental feature and see how it's going to be adopted?
Maybe we can change the event name for the cooperative mode to something like `cooperative_request`? 
That way we have a clear separation between the two modes for now and someone registering a `request` listener (maybe even before the cooperative mode is enabled) does not get unexpected behaviour.

sadym-chromium:
+1 for `cooperative_request`

benallfree:
@jschfflr @sadym-chromium Interesting idea, I’ll play around with it. Thanks for the feedback!

benallfree:
@jschfflr @sadym-chromium Ok guys you've convinced me the new event name is a good idea :)

Here's how it works now:

```typescript

;(async () => {
  console.log

  page.setRequestInterception(true)

  const myHandler = (name) => (req) => {
    return new Promise((resolve) => {
      setTimeout(() => {
        console.log(`${name} handled!`)
        req.continue()
        resolve()
      }, Math.random() * 1000)
    })
  }
  /**
   * This will only get called if `cooperative_request` is supported.
   */
  page.on('cooperative_request', myHandler('coop money eat'))
  page.on('cooperative_request', myHandler('coop money sleep'))
  page.on('cooperative_request', myHandler('coop money in that order'))

  /**
   * This will throw an exception in any version of Puppeteer that supports cooperative mode
   * if and only if a handler is listening for `cooperative_request`. It must do this because
   * `request` and `cooperative_request` are mutually exclusive handling modes. This behavior
   * will encourage code maintainers to update their code to use `cooperative_request` in
   * order to play nicely with others using cooperative intercepts.
   */
  page.on('request', myHandler('legacy'))

  await page.goto('https://example.com')
  await page.screenshot({ path: 'example.png' })

  await browser.close()
})()
```

Rules:
* Cooperative handlers are queued and executed in the order they were registered. Each handler's promise is resolved before the next one is executed. 
* During a request, `cooperative_request` is fired first. If any handler responds, the `request` legacy flow is automatically canceled. Any attempts to use it will result in a runtime error. By throwing an exception, we help people understand that these modes are mutually exclusive and will also help to track down and update 3rd party code.

What do you think?

benallfree:
@XavierGeerinck please take a look, i think this makes sense

XavierGeerinck:
> @XavierGeerinck please take a look, i think this makes sense

This is amazing work! Glad to see someone picking this up and working so hard on resolving it!!

My opinion here is that I find it strange to utilize a new event name. Although I do understand the reasoning behind this and making it "experimental", but I however think this should be a core feature in Puppeteer that is there from the start.

When now requests are not handled but multiple intercepts are being done, Puppeteer at its core will crash. By resolving this on the `request` event handler, this will be solved for more codebases already. 

If we name this `cooperative_requests` we are adding a feature and relying on people to update towards this new event handler, causing a potential fork in community adoption (people having `request` and people having `cooperative_requests`)

Seeing the above, my preference then goes to `request` handling this properly as it should 😊 but just my 2 cents of course, would love to see all input on this!

benallfree:
@XavierGeerinck Thank you for articulating that better than I could. I have the identical concern and completely agree that it would be better to keep it the way I had it before. 

The original PR required an explicit flag to enable cooperative mode. I don’t think users would be anything other than pleasantly surprised when the behavior of request intercepts changed due to that explicit setting. And since they explicitly enabled cooperative mode, I think they would also know to consider that choice during any troubleshooting. 

I would love to revert this latest approach for the reasons you described.  I think it’s harmful for adoption. But if it’s the only way to get something in the core, then I’m inclined to accept it.

To be fair to @jschfflr and @sadym-chromium, I can understand the their concern about creating a vector for unexpected behavior to occur, resulting in spurious bug reports when request interception in a third-party package stops behaving like it should. From what I have seen, that would be a pretty uncommon occurrence since I’ve been very careful to preserve every aspect of backward compatibility.

src/common/HTTPRequest.ts
```diff
@@ -296,10 +432,26 @@ export class HTTPRequest {
     // Request interception is not supported for data: urls.
     if (this._url.startsWith('data:')) return;
     assert(this._allowInterception, 'Request Interception is not enabled!');
-    assert(!this._interceptionHandled, 'Request is already handled!');
-    const { url, method, postData, headers } = overrides;
-    this._interceptionHandled = true;
+    assert(
+      !this._interceptionHandled,
+      'Request is already handled! Did you mean to use cooperative_request?'
```
jschfflr: Maybe change this to mention that cooperative_request is an event one can subscribe to.

src/common/HTTPRequest.ts
```diff
@@ -150,6 +166,126 @@ export class HTTPRequest {
     return this._url;
   }
 
+  hasCooperativeInterceptHandlers(): boolean {
+    return this._pendingCooperativeInterceptionHandlers.length > 0;
+  }
+
+  setAllowCooperativeRequestInterceptionMode(isEnabled: boolean): boolean {
```
jschfflr: Is there a specific reason for returning the boolean again? Otherwise, I'd change this to void.

jschfflr (changes requested):
Apart from the nits, please also include tests for the new functionality!

jschfflr:
@benallfree Thanks for your understanding. The change now looks good to me with some small nits and a request to add tests for the new functionality. 
I will be more than happy to make this the default functionality in the future but I'd like to keep it somewhat separate for now to see how people are adopting it and to be able to make changes without breaking any assumptions.

benallfree:
@jschfflr I'm very happy! :) Yes, I'll add tests and better docs now that we have an approved plan. Thank you!

jschfflr:
@benallfree just a gentle ping to see if there's been some progress on this cl :)

benallfree:
@jschfflr Just merged the latest from upstream `main` and am working on adding tests.

Note: upstream `main` is currently failing locally with 

```
1) Page
       Page.Events.Console
         should have location when fetch fails:
     Error: expect(received).toContain(expected) // indexOf

Expected substring: "ERR_NAME_NOT_RESOLVED"
Received string:    "Access to fetch at 'http://wat/' from origin 'http://localhost:8907' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled."
      at Context.<anonymous> (test/page.spec.ts:577:30)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:94:5)
```

benallfree:
@jschfflr Looking at the docs, I have a couple questions:

1. Am I supposed to modify `docs/api.md` manually?
2. I am thinking more about adding a `Page.setCooperativeRequestInterception(true)` to match `Page.setRequestInterception()`. As I think about it more, I feel like it might be better to require explicit enabling of cooperative mode and maybe even make it mutually exclusive. The code is currently backward compatible and will run normal request intercepts after running cooperative intercepts, but I don't love this idea anymore. I think making them completely separate might be best, even to the extent of subclassing an `HTTPCooperativeRequest`. This would allow us to use the same `request` event name, but the behavior would be governed by `Page.setXXXRequestInterception()`. I like this idea because then it would allow users to test their code simply by replacing `setRequestInterception(true)` with `setCooperativeRequestInterception(true)`. On the other hand, maybe that idea is going in circles. Using a separate `cooperative_request` event name is safer because it provide better isolation particularly where 3rd party and transient dependencies might still be relying upon `request` and the old behavior. Thoughts?

Examples:

```typescript
page.setCooperativeRequestInterception(true);
page.setRequestInterception(true); // Throws an Error: use legacy request interception mode or cooperative mode, but not both
```

```typescript
page.setRequestInterception(true); 
page.setCooperativeRequestInterception(true); // Throws an Error: use legacy request interception mode or cooperative mode, but not both
```

```typescript
// OPTION A
page.setCooperativeRequestInterception(true);
page.on('request', ...) // This code will no longer run because the 'request' event doesn't fire
page.on('cooperative_request', ...) // Only this code will run

// OPTION B
page.setCooperativeRequestInterception(true);
page.on('request', ...) // This code will run in cooperative mode
page.on('cooperative_request', ...) // This event name does not exist, we reuse the 'request' event name
```

jschfflr:
Thanks for the update! I'd love to get this over the finish line soon :)
Not sure why the test fails on main - I'll investigate this.

> 1. Am I supposed to modify `docs/api.md` manually?

Yes, you modify the file manually and have `npm run doc` check it afterwards. 
You probably already checked this, so just for reference: [Writing Documentation](https://github.com/puppeteer/puppeteer/blob/main/CONTRIBUTING.md#writing-documentation)

> 2. I am thinking more about adding a `Page.setCooperativeRequestInterception(true)` to match `Page.setRequestInterception()`. 

As I mentioned before, I'm also a fan of keeping the two separate for now. So adding `Page.setCooperativeRequestInterception(true)` and using `cooperative_request` (Option A) sounds good to me!
When we see a wider adoption of the new behaviour, we can always adapt the API in the future.

jschfflr:
Unfortunately, I can't reproduce the failing test on main. 
If it persists, feel free to file a new issue for this with further details!

benallfree:
@jschfflr I've been working more on this and have what I think might be a fantastic idea. @XavierGeerinck @berstend please tell me what you think too.

`page.on('request', ...)` is fired every time a request comes in, regardless of whether interception is enabled. `page.setRequestInterception(true)` just allows `abort/respond/continue` to be called, and the request will be held (frozen) until one of those is called. So [Option A above](https://github.com/puppeteer/puppeteer/pull/6735#issuecomment-826924219) will not work. 

But here's something that might work really well while preserving legacy behavior:

```typescript
page.setRequestInterception(true)
page.on('request', (request)=>{
   // Legacy behavior. The first handler to call any of these 'wins' 
   // and other handlers will error out if these are called a second time.
   request.abort() 
   request.respond()
   request.continue()

   // Cooperation. Any handler may call these in any order, any number of times, to suggest
   // an intercept resolution. Async handlers are await'd before deciding what to do.
   // The highest priority resolution wins. In the event of a tie, abort > respond > continue.
   request.requestAbort(priority=10)
   request.requestRespond(priority=10)
   request.requestContinue(priority=10)
});
```

This feels really clean to me. If any handler uses the legacy abort/respond/continue, everything will continue to work as it always has. But if all handlers use the new actions, then interception will behave cooperatively. 

I love this solution because we can keep the `request` event name and maintainers can upgrade to the new actions with time. The hope would be that these new methods become the recommended/preferred approach and the old methods can be deprecated.

benallfree:
Ok @jschfflr I'm pretty excited about this solution, including passing tests. It's super elegant, I think. Maybe I'm missing something big though, please review the code and see if you like this approach.

My thinking behind the `priority` flag is that end users can use this to override intercept behavior buried in 3rd party interceptions. Authors of 3rd party libraries should always use the default priority (0) or a configurable priority number so that end users can adjust their intercept handlers accordingly.

```typescript
page.on('request', (request) => 
   request.requestContinue({})
);
page.on('request', (request) => 
   request.requestAbort('aborted') // This will succeed because abort > continue when priorities are identical
);
page.on('requestfailed', (request) => 
   console.log('hello world')
);
```


```typescript
page.on('request', (request) => 
   request.requestContinue({},1) // This will succeed because priority 1 > priority 0
);
page.on('request', (request) => 
   request.requestAbort('aborted') // Default priority 0 
);
page.on('requestfinished', (request) => 
   console.log('hello world')
);
```

benallfree:
@jschfflr I copied the whole interception test spec and converted it to use the new experimental API just to prove everything is backward-compatible. So the legacy API spec passes (unchanged), and the same spec also passes using the new experimental API instead. That means, in most cases, maintainers should be able to update to the `respondXXX()` API without any other changes to their code.

I also added a test to cover cooperative behavior nature of the new API.

jschfflr:
@benallfree Thanks for all the effort you are putting into this! I appreciate it a lot!
The new solutions looks good to me but I'd like to check back with the team before committing to it - just to make sure we don't miss any details here.

jschfflr:
@benallfree The team is happy with the new approach! I'll review the pull request from a technical point of view tomorrow :)

benallfree:
@jschfflr Wow very cool, a fresh review is very welcome. I know the API docs need updating, but I'm also curious what you think about the method naming convention (`requestXXX`), additional testing, etc.

docs/api.md
```diff
@@ -1354,6 +1355,12 @@ Shortcut for [page.mainFrame().$x(expression)](#framexexpression)
 
 - returns: <[Accessibility]>
 
+#### page.abortErrorReason()
```
jschfflr: Why do we need this method on `page`?

src/common/HTTPRequest.ts
```diff
@@ -160,6 +178,118 @@ export class HTTPRequest {
...
+  /**
+   * @returns `true` if `respond` is currently the highest priority intercept resolution
+   */
+  private shouldRespond(): boolean {
+    if (!this._allowInterception) return false;
+    if (this._interceptionHandled) return false;
+    if (!this._respondRequested) return false;
+    if (
+      this._continueRequested &&
+      this._continuePriority > this._respondPriority
```
jschfflr: All other comparisons use `>=` but this one uses `>` - is there a specific reason for that?

src/common/HTTPRequest.ts
```diff
@@ -459,6 +717,9 @@ const errorReasons: Record<ErrorCode, Protocol.Network.ErrorReason> = {
   failed: 'Failed',
 } as const;
 
+export type ActionResult = 'continue' | 'abort' | 'respond';
+export const DEFAULT_INTERCEPT_PRIORITY = 0;
```
jschfflr: Earlier, we talked about this being 10 - what's the reason you changed this to 0?

jschfflr (commented):
I already left some comments but unfortunately didn't get to test the code hands on yet.

docs/api.md
benallfree: Doc error :D

src/common/HTTPRequest.ts
benallfree: Continue us the lowest priority. In the event of a tie, respond should win.

If abortPriority <= respondPriority, abort

If respondPriority <= continuePrioirty, respond

src/common/HTTPRequest.ts
benallfree: I can't think of a use case where someone would intentionally want a priority lower than the default, but I suppose it could happen.

jschfflr: Sounds good!

jschfflr:
So I checked out the cl and played around with it a bit and came across the following scenario:
```

    console.log('continue', request.url().slice(0, 30));
    timers.setTimeout(() => {
      request.continue();
    }, 500);
  })

    console.log('requestAbort', request.url().slice(0, 30));
    request.requestAbort()
  })
  ```
When assuming that the first request handler is something that a user is doing in their code and the second one is something that happens in a plugin - this might still be confusing to users :/
Any thoughts?

**Comment by benallfree** [comment]
@jschfflr hmm not sure I follow. In your example, It seems like the request would be aborted either way, whether using abort() or requestAbort().

Were you trying to come up with an example where a plug-in changing from abort() to requestAbort() would change the outcome?

**Comment by jschfflr** [comment]
@benallfree Yeah, I tried to check the code for situations in which it behaved in an unexpected way. But you're right - the problem in this case is not with the change to `requestAbort`...

**Comment by benallfree** [comment]
@jschfflr Here's an example:

```ts
// Old plugin code
page.on('request', req => {
   if(req.url().match(/doubleclick/) req.abort();
});
```

```ts
// New plugin code
page.on('request', req => {
   if(req.url().match(/doubleclick/) req.requestAbort();
});
```

```ts
// User code
page.on('request', req => {
   // With the old plugin code, the request would have been aborted.
   // With the new plugin code, the legacy continue() forces a continue
   // even though an abort was requested.
   // Just by updating the plugin, the behavior will have changed. But that's
   // because the plugin maintainer actually changed their code, it wasn't
   // just that the Puppeteer version updated (what we were trying to avoid).
   req.continue()
});
```

jschfflr:
Thanks for the example - this scenario looks good to me!

benallfree:
@jschfflr Have everyone's concerns been addressed? Green light to proceed with documentation and cleanup?

jschfflr:
Yes, I think all concerns are addressed for now and we can start with documentation and cleanup!

jschfflr:
@benallfree Hey, did you already have time to clean up and work on the documentation?

benallfree:
@jschfflr I’ll update this PR within a week :) The project using this ended, so I’m continuing work in my spare cycles.

jschfflr:
Yeah sure, no worries! I just wanted to make sure that we're bringing this over the finish line :)

benallfree:
Hey @jschfflr I apologize for the delay but I started documenting and then realized that think I found an even simpler API to express our goals. 

Can you please let me know what you and the team think of https://github.com/puppeteer/puppeteer/pull/6735/commits/2b8088eb35ed19b7bfa22f7de751418d4d8e67e8? I propose that we add a parameter to existing methods rather than introduce new methods:

1. Add a `priority` parameter to `abort`, `continue`, and `respond`, default `0`.
1. Lower priorities win, with 0 meaning execute immediately (legacy behavior)
2. In the event of a priority tie, `abort` > `respond` > `continue`

See `interceptResolution()` for logic.

So it would look like this:

```ts
// Final outcome: immediate abort()
page.setRequestInterception(true);
page.on('request', request=>{
   request.abort('failed'); // Legacy behavior: would execute immediately
});
page.on('request', request=>{
   console.log(request.interceptResolution()); // ['already-handled'], meaning a legacy resolution has taken place
});
```

```ts
// Final outcome: immediate continue()
page.setRequestInterception(true);
page.on('request', request=>{
   request.abort('failed', 10); // Cooperative behavior: would execute at priority 10
});
page.on('request', request=>{
   console.log(request.interceptResolution()); // ['abort', 10], meaning an abort @ 10 is the current winning resolution
   request.continue({}); // Legacy behavior: would execute immediately
});
```

```ts
// Final outcome: cooperative continue()
page.setRequestInterception(true);
page.on('request', request=>{
   request.abort('failed', 10); // Cooperative behavior: aborts at priority 10
});
page.on('request', request=>{
   request.continue(this.continueRequestOverrides(), 5);  // Cooperative behavior: continues at priority 5
});
page.on('request', request=>{
   console.log(request.interceptResolution()); // ['continue', 5], because continue @ 5 > abort @ 10
});
```

```ts
// Final outcome: cooperative respond()
page.setRequestInterception(true);
page.on('request', request=>{
   request.abort('failed', 10); // Cooperative behavior: aborts at priority 10
});
page.on('request', request=>{
   request.continue(this.continueRequestOverrides(), 5);   // Cooperative behavior: continues at priority 5
});
page.on('request', request=>{
   request.respond(this.responseForRequest(), 5);   // Cooperative behavior: responds at priority 5
});
page.on('request', request=>{
   console.log(request.interceptResolution()); // ['respond', 5], because respond @ 5 > continue @ 5 > abort @ 10
});
```

Pros:
* Minimal API and code changes
* Preserves legacy behavior - if any handler is still using the legacy system, it prevails
* The new `interceptResolution()` method gives users a way to query the status of the current intercept. For example, the `already-handled` status can be used to detect unexpected resolutions.
* Requires explicit adoption of cooperative behavior. Later, this can be transitioned to default behavior simply by making the `priority` param default to `10` instead of `0`. We could even make an advisory note in the docs that 10 is the future planned default so users can code with that in mind. We could also recommend that plugin authors add a plugin config param so the priority they use internally can be set by the user.

Cons:
* Needing to pass a magic `priority > 0` parameter to enable cooperative mode is less obvious than new methods named `requestXXX()`. But since the parameter signatures would be identical anyway, it made sense to just add the `priority` param to the old method names.
* Any downside to a new param vs a new method from an API compat/maintenance point of view?

src/common/HTTPRequest.ts
```diff
@@ -160,6 +178,126 @@ export class HTTPRequest {
...
+    return this._abortErrorReason;
+  }
+
+  /**
+   * @returns An array of the current intercept resolution strategy and priority
+   * `[strategy,priority]`. Strategy is one of: `abort`, `respond`, `continue`,
+   *  `disabled`, `none`, or `already-handled`.
+   */
+  private interceptResolution(): [
+    'abort' | 'respond' | 'continue' | 'disabled' | 'none' | 'alreay-handled',
```
jschfflr: Could you create a new type for this?

src/common/HTTPRequest.ts
```diff
@@ -124,6 +124,16 @@ export class HTTPRequest {
   private _postData?: string;
   private _headers: Record<string, string> = {};
   private _frame: Frame;
+  private _continueRequestOverrides: ContinueRequestOverrides;
+  private _responseForRequest: Partial<ResponseForRequest>;
+  private _abortErrorReason: Protocol.Network.ErrorReason;
+  private _continueRequested: boolean;
```
jschfflr: WDYT about changing this to _currentStrategy and _currentPriority instead?

src/common/HTTPRequest.ts
```diff
@@ -462,6 +648,10 @@ const errorReasons: Record<ErrorCode, Protocol.Network.ErrorReason> = {
   failed: 'Failed',
 } as const;
 
+export type ActionResult = 'continue' | 'abort' | 'respond';
+export const DEFAULT_INTERCEPT_PRIORITY = 10;
+export const LEGACYY_DEFAULT_INTERCEPT_PRIORITY = 0; // Immediate
```
jschfflr: Nit: typo

src/common/HTTPRequest.ts
```diff
@@ -160,6 +178,126 @@ export class HTTPRequest {
...
+    return ['none'];
+  }
+
+  /**
+   * Adds an async request handler to the processing queue.
+   * Deferred handlers are not guaranteed to execute in any particular order,
+   * but they are guarnateed to resolve before the request interception
+   * is finalized.
+   */
+  enqueueInterceptAction(pendingHandler: () => void | PromiseLike<any>): void {
```
jschfflr: Could this also be changed to unknown?

src/common/HTTPRequest.ts
```diff
@@ -295,12 +433,29 @@ export class HTTPRequest {
    * ```
    *
    * @param overrides - optional overrides to apply to the request.
+   * @param priority - (optional, default 0). Priority 0 will execute immediately.
+   * Priorities > 0 will be resolved using coopeative handling rules.
    */
-  async continue(overrides: ContinueRequestOverrides = {}): Promise<void> {
+  async continue(
+    overrides: ContinueRequestOverrides = {},
+    priority = LEGACYY_DEFAULT_INTERCEPT_PRIORITY
```
jschfflr: WDYT about typing this as `priority?: number` and checking for `undefined` instead of `LEGACYY_DEFAULT_INTERCEPT_PRIORITY`

jschfflr (commented):
I like your idea as it introduces less changes to the API surface. 
Let me know what you think about my comments!

jschfflr:
@benallfree Did you already have time to look at my comments?

src/common/HTTPRequest.ts
benallfree: I like this idea.

src/common/HTTPRequest.ts
benallfree: Will do 👍

src/common/HTTPRequest.ts
benallfree: Sounds okay, I'll try it and see how it goes.

jschfflr:
@benallfree Just a gentle ping to see if there are any updates on this?

benallfree:
Apologies @jschfflr i thought I was waiting on you but I see it’s my turn to make changes. I’ll get on it!

jschfflr:
No worries, looking forward to get this landed :)

benallfree:
@jschfflr Please check out the latest and let me know what you think! If you like it what you see in the code, I'll update the tests again.

src/common/HTTPRequest.ts
```diff
@@ -295,12 +382,42 @@ export class HTTPRequest {
...
+      return;
+    }
+    if (priority === this._currentPriority) {
+      if (
+        this._currentStrategy === 'abort' ||
+        this._currentStrategy === 'respond'
+      ) {
+        return;
+      }
+      this._currentPriority = priority;
```
jschfflr: In this case, `priority` is already equal to `this._currentPriority` because of the `===` comparison earlier.

src/common/HTTPRequest.ts
```diff
@@ -295,12 +382,42 @@ export class HTTPRequest {
    * ```
    *
    * @param overrides - optional overrides to apply to the request.
+   * @param priority - (optional, default 0). Priority 0 will execute immediately.
```
jschfflr: In the current implementation, `0` is not implemented as an immediate execution (only `undefined`)

jschfflr (commented):
LGTM % some minor comments

benallfree:
@jschfflr Good catches, fixed :)

Docs & tests have not been updated, just focused on implementation for now.

jschfflr:
Thanks! I think this looks good from a technical point of view now :)
Let's get testing and docs ready and maybe we can even include this in the release that I'm going to cut later today!

benallfree:
@jschfflr Ok I'll see how far I can get! Take a look at api.md, I just pushed some changes.

docs/api.md
```diff
@@ -2352,6 +2352,184 @@ const puppeteer = require('puppeteer');
...
 
+##### Cooperative Intercept Mode and Legacy Intercept Mode
+
+`request.respond`, `request.abort`, and `request.continue` can accept an optional `priority` to activate Cooperative Intercept Mode. In Cooperative Mode, all intercept handlers are guaraneed to run and the interception will be resolved to the highest priority resolution. Here are the rules of Cooperative Mode:
+
+- Async handlers will finish before intercept resolution is finalized
+- The highest priority interception resolution "wins", ie, the interception will ultimately be aborted/responded/continued according to which resolution was given the highest priority.
+- In the event of a tie, `abort` > `respond` > `continue`
+
+For standardization, when specifying a Cooperative Mode priority use `DEFAULT_INTERCEPT_RESOLUTION_PRIORITY` (currently 10) unless you have a clear reason to use a higher priority. This will gracefully prefer `respond` over `continue` and `abort` over `respond`. If you do intentionally want to use a higher priority, `DEFAULT_INTERCEPT_RESOLUTION_PRIORITY + n` is the recommended approach. For example, `continue({}, DEFAULT_INTERCEPT_RESOLUTION_PRIORITY + 4)` rather than `continue({}, 14)`.
```
jschfflr: I'm wondering if there is a way to get around carrying the `DEFAULT_INTERCEPT_RESOLUTION_PRIORITY` constant around all the time. I feel like this makes for a rather bumpy experience. WDYT about changing the default value to 0 and allowing both negative and positive priorities or instead fixing it to some given value like 10?

docs/api.md
```diff
@@ -2344,6 +2352,184 @@ const puppeteer = require('puppeteer');
...
+    interceptedRequest.abort('failed', DEFAULT_INTERCEPT_RESOLUTION_PRIORITY);
+  else
+    interceptedRequest.continue(
+      interceptedRequest.continueRequestOverrides(),
+      DEFAULT_INTERCEPT_RESOLUTION_PRIORITY
+    );
+});
+```
+
+With those simiple upgrades, your handler will use Cooperative Mode instead.
```
jschfflr: simple

jschfflr:
@benallfree Thanks for taking the time to write this extensive documentation - I hope this will help package maintainers adopt the new mode easily :)

benallfree:
@jschfflr I noticed that https://www.npmjs.com/package/@types/puppeteer exists as a separate package, this one will need to be updated. How do we coordinate that?

benallfree:
@jschfflr unit tests updated... i think we're getting close!

benallfree:
@jschfflr api.md finalized

Ok I think everything is finalized. If more changes are needed, we may need to wait for the next release cycle because I'm pretty busy today.

jschfflr:
> @jschfflr I noticed that https://www.npmjs.com/package/@types/puppeteer exists as a separate package, this one will need to be updated. How do we coordinate that?

I think, we are now shipping our own type definitions so I'm not sure if that package is still relevant.
@mathiasbynens might know more about that. 

I'll check out your updates :)

jschfflr (approved):
LGTM now but I'd like for @mathiasbynens to also take a look!

jschfflr:
@googlebot I consent.

jschfflr:
@benallfree I fixed some minor nits myself, I hope that's ok for you!

benallfree:
> @benallfree I fixed some minor nits myself, I hope that's ok for you!

Whatever gets it shipped is okay with me! 😂

docs/api.md
```diff
@@ -328,22 +330,21 @@
...
-  * [httpRequest.continue([overrides])](#httprequestcontinueoverrides)
+  * [httpRequest.continue([overrides], [priority])](#httprequestcontinueoverrides-priority)
   * [httpRequest.continueRequestOverrides()](#httprequestcontinuerequestoverrides)
-  * [httpRequest.enqueueInterceptAction(handler)](#httprequestnequeueinterceptactionhandler)
+  * [httpRequest.enqueueInterceptAction(pendingHandler)](#httprequestenqueueinterceptactionpendinghandler)
   * [httpRequest.failure()](#httprequestfailure)
   * [httpRequest.finalizeInterceptions()](#httprequestfinalizeinterceptions)
   * [httpRequest.frame()](#httprequestframe)
   * [httpRequest.headers()](#httprequestheaders)
-  * [httpRequest.interceptResolution()](#httprequestinterceptresolution)
```
benallfree: Did you mean to remove this? It’s an important method for package authors.

jschfflr:
Great :)

It looks like we still need tests for the following three methods:
```
HTTPRequest.continueRequestOverrides
HTTPRequest.responseForRequest
HTTPRequest.abortErrorReason
```

Could you add them (I'm in a meeting now unfortunately)

benallfree:
```
HTTPRequest.continueRequestOverrides
HTTPRequest.responseForRequest
HTTPRequest.abortErrorReason
```

These are used indirectly in the new test harness, except for `abortErrorReason`. I'm out of time today, I'll see if I can get these added ASAP. IMHO I'd say don't hold up the release since 2/3 of them are used in testing already.

jschfflr:
> These are used indirectly in the new test harness, except for `abortErrorReason`. I'm out of time today, I'll see if I can get these added ASAP. IMHO I'd say don't hold up the release since 2/3 of them are used in testing already.

Unfortunately, the bots require us to have test coverage for all public methods. 
I added tests for them - let's see if they are passing now.

docs/api.md
```diff
@@ -2344,6 +2353,175 @@ const puppeteer = require('puppeteer');
 })();
 ```
 
+##### Cooperative Intercept Mode and Legacy Intercept Mode
+
+`request.respond`, `request.abort`, and `request.continue` can accept an optional `priority` to activate Cooperative Intercept Mode. In Cooperative Mode, all intercept handlers are guaraneed to run and the interception will be resolved to the highest priority resolution. Here are the rules of Cooperative Mode:
```
```suggestion
`request.respond`, `request.abort`, and `request.continue` can accept an optional `priority` to activate Cooperative Intercept Mode. In Cooperative Mode, all intercept handlers are guaranteed to run and the interception is resolved to the highest-priority resolution. Here are the rules of Cooperative Mode:
```

docs/api.md
```diff
@@ -2344,6 +2353,175 @@ const puppeteer = require('puppeteer');
...
+
+- Async handlers will finish before intercept resolution is finalized
+- The highest priority interception resolution "wins", ie, the interception will ultimately be aborted/responded/continued according to which resolution was given the highest priority.
+- In the event of a tie, `abort` > `respond` > `continue`
+
+For standardization, when specifying a Cooperative Mode priority use `0` unless you have a clear reason to use a higher priority. This will gracefully prefer `respond` over `continue` and `abort` over `respond`. If you do intentionally want to use a different priority, higher priorities win over lower priorities. Negative priorities are allowed. For example, `continue({}, 4)` would win over `continue({}, -2)`.
+
+To preserve backward compatibility, any handler resolving the intercept without specifying `priority` (Legacy Mode) will cause immediate resolution. For Cooperative Mode to work, all resolutions must use a `priority`.
```
```suggestion
- Async handlers finish before intercept resolution is finalized.
- The highest priority interception resolution "wins", i.e. the interception is ultimately aborted/responded/continued according to which resolution was given the highest priority.
- In the event of a tie, `abort` > `respond` > `continue`.

For standardization, when specifying a Cooperative Mode priority use `0` unless you have a clear reason to use a higher priority. This gracefully prefers `respond` over `continue` and `abort` over `respond`. If you do intentionally want to use a different priority, higher priorities win over lower priorities. Negative priorities are allowed. For example, `continue({}, 4)` would win over `continue({}, -2)`.

To preserve backward compatibility, any handler resolving the intercept without specifying `priority` (Legacy Mode) causes immediate resolution. For Cooperative Mode to work, all resolutions must use a `priority`.
```

docs/api.md
```diff
@@ -2344,6 +2353,175 @@ const puppeteer = require('puppeteer');
...
+});
+page.on('request', (request) => {
+  request.respond(request.responseForRequest(), 15); // Cooperative behavior: responds at priority 15
+});
+page.on('request', (request) => {
+  console.log(request.interceptResolution()); // ['respond', 15], because respond @ 15 > continue @ 15 > abort @ 10
+});
+```
+
+##### Upgrading to Cooperative Mode for Package Maintainers
```
mathiasbynens: We use sentence case for titles:

```suggestion
##### Upgrading to Cooperative Mode for package maintainers
```

("Cooperative Mode" is consistently capitalized even in sentences)

docs/api.md
```diff
@@ -2344,6 +2353,175 @@ const puppeteer = require('puppeteer');
...
+    interceptedRequest.abort('failed', 0);
+  else
+    interceptedRequest.continue(
+      interceptedRequest.continueRequestOverrides(),
+      0
+    );
+});
+```
+
+With those simple upgrades, your handler will use Cooperative Mode instead.
```
```suggestion
With those simple upgrades, your handler now uses Cooperative Mode instead.
```

docs/api.md
```diff
@@ -2344,6 +2353,175 @@ const puppeteer = require('puppeteer');
...
+    interceptedRequest.abort('failed', _strategy.abortPriority);
+  else
+    interceptedRequest.continue(
+      interceptedRequest.continueRequestOverrides(),
+      _strategy.continuePriority
+    );
+});
+```
+
+The above solution will ensure that your package continues to work as expected with other legacy handlers while also allowing the user to adjust the importance of your package in the resolution chain when Cooperative Mode is being used.
```
```suggestion
The above solution ensures that your package continues to work as expected with other legacy handlers while also allowing the user to adjust the importance of your package in the resolution chain when Cooperative Mode is being used.
```

docs/api.md
```diff
@@ -4432,18 +4610,26 @@ If request gets a 'redirect' response, the request is successfully finished with
   - `namenotresolved` - The host name could not be resolved.
   - `timedout` - An operation timed out.
   - `failed` - A generic failure occurred.
+- `priority` <[number]> - Optional intercept abort priority. If provided, intercept will be resolved using coopeative handling rules. Otherwise, intercept will be resovled immediately.
```
mathiasbynens, line 4674:
```suggestion
- `priority` <[number]> - Optional intercept abort priority. If provided, intercept is resolved using cooperative handling rules. Otherwise, intercept is resolved immediately.
```

docs/api.md
```diff
@@ -4432,18 +4610,26 @@ If request gets a 'redirect' response, the request is successfully finished with
...
+Returns the most recent reason for aborting set by the previous call to abort() in Cooperative Mode.
+
+#### httpRequest.continue([overrides], [priority])
 
 - `overrides` <[Object]> Optional request overwrites, which can be one of the following:
   - `url` <[string]> If set changes the request URL. This is not a redirect. The request will be silently forwarded to the new URL. For example, the address bar will show the original URL.
   - `method` <[string]> If set changes the request method (e.g. `GET` or `POST`).
   - `postData` <[string]> If set changes the post data of request.
   - `headers` <[Object]> If set changes the request HTTP headers. Header values will be converted to a string.
+- `priority` <[number]> - Optional intercept abort priority. If provided, intercept will be resolved using coopeative handling rules. Otherwise, intercept will be resovled immediately.
```
mathiasbynens, line 4693:
```suggestion
- `priority` <[number]> - Optional intercept abort priority. If provided, intercept is resolved using cooperative handling rules. Otherwise, intercept is resolved immediately.
```

docs/api.md
```diff
@@ -4461,6 +4647,22 @@ page.on('request', (request) => {
 });
 ```
 
+#### httpRequest.continueRequestOverrides()
+
+- returns: <[Object]> Optional request overwrites, which can be one of the following:
+  - `url` <[string]> If set changes the request URL. This is not a redirect. The request will be silently forwarded to the new URL. For example, the address bar will show the original URL.
```
mathiasbynens, line 4714:
```suggestion
  - `url` <[string]> Changes the request URL if set. This is not a redirect — the request is silently forwarded to the new URL. For example, the address bar shows the original URL.
```

docs/api.md
```diff
@@ -4461,6 +4647,22 @@ page.on('request', (request) => {
...
+  - `method` <[string]> If set changes the request method (e.g. `GET` or `POST`).
+  - `postData` <[string]> If set changes the post data of request.
+  - `headers` <[Object]> If set changes the request HTTP headers. Header values will be converted to a string.
```
mathiasbynens, line 4717:
```suggestion
  - `headers` <[Object]> If set changes the request HTTP headers. Header values are converted to a string.
```

docs/api.md
```diff
@@ -4536,13 +4744,14 @@ console.log(chain.length); // 0
...
 
-#### httpRequest.respond(response)
+#### httpRequest.respond(response, [priority])
 
 - `response` <[Object]> Response that will fulfill this request
   - `status` <[number]> Response status code, defaults to `200`.
   - `headers` <[Object]> Optional response headers. Header values will be converted to a string.
   - `contentType` <[string]> If set, equals to setting `Content-Type` response header
   - `body` <[string]|[Buffer]> Optional response body
+- `priority` <[number]> - Optional intercept abort priority. If provided, intercept will be resolved using coopeative handling rules. Otherwise, intercept will be resovled immediately.
```
mathiasbynens, line 4815:
```suggestion
- `priority` <[number]> - Optional intercept abort priority. If provided, intercept is resolved using cooperative handling rules. Otherwise, intercept is resolved immediately.
```

src/common/HTTPRequest.ts
```diff
@@ -160,6 +170,83 @@ export class HTTPRequest {
...
+   * if the interception is allowed to respond (ie, `abort()` is not
+   * called).
+   */
+  responseForRequest(): Partial<ResponseForRequest> {
+    assert(this._allowInterception, 'Request Interception is not enabled!');
+    return this._responseForRequest;
+  }
+
+  /**
+   * @returns the most recent reason for aborting the request
```
mathiasbynens, line 193:
```suggestion
   * @returns The most recent reason for aborting the request.
```

src/common/HTTPRequest.ts
```diff
@@ -160,6 +170,83 @@ export class HTTPRequest {
...
+   */
+  continueRequestOverrides(): ContinueRequestOverrides {
+    assert(this._allowInterception, 'Request Interception is not enabled!');
+    return this._continueRequestOverrides;
+  }
+
+  /**
+   * @returns the `ResponseForRequest` that will be used
+   * if the interception is allowed to respond (ie, `abort()` is not
+   * called).
```
mathiasbynens:
```suggestion
   * @returns The `ResponseForRequest` that gets used if the
   * interception is allowed to respond (ie, `abort()` is not called).
```

src/common/HTTPRequest.ts
```diff
@@ -295,12 +382,45 @@ export class HTTPRequest {
    * ```
    *
    * @param overrides - optional overrides to apply to the request.
+   * @param priority - If provided, intercept will be resolved using
+   * coopeative handling rules. Otherwise, intercept will be resovled
+   * immediately.
```
mathiasbynens:
```suggestion
   * @param priority - If provided, intercept is resolved using
   * cooperative handling rules. Otherwise, intercept is resolved
   * immediately.
```

src/common/HTTPRequest.ts
```diff
@@ -351,12 +471,39 @@ export class HTTPRequest {
    * Calling `request.respond` for a dataURL request is a noop.
    *
    * @param response - the response to fulfill the request with.
+   * @param priority - If provided, intercept will be resolved using
+   * coopeative handling rules. Otherwise, intercept will be resovled
+   * immediately.
```
mathiasbynens:
```suggestion
   * @param priority - If provided, intercept is resolved using
   * cooperative handling rules. Otherwise, intercept is resolved
   * immediately.
```

src/common/HTTPRequest.ts
```diff
@@ -403,14 +550,37 @@ export class HTTPRequest {
    * throw an exception immediately.
    *
    * @param errorCode - optional error code to provide.
+   * @param priority - If provided, intercept will be resolved using
+   * coopeative handling rules. Otherwise, intercept will be resovled
+   * immediately.
```
mathiasbynens:
```suggestion
   * @param priority - If provided, intercept is resolved using
   * cooperative handling rules. Otherwise, intercept is resolved
   * immediately.
```

src/common/NetworkManager.ts
```diff
@@ -376,6 +376,11 @@ export class NetworkManager extends EventEmitter {
     );
     this._requestIdToRequest.set(event.requestId, request);
     this.emit(NetworkManagerEmittedEvents.Request, request);
+    request.finalizeInterceptions().catch((error) => {
+      // This should never happen, but catch just in case
```
mathiasbynens: comments are sentences

```suggestion
      // This should never happen, but catch just in case.
```

OrKoN (approved):

mathiasbynens (approved):
Overall LGTM. W.r.t. my suggestions, I mostly focused on the docs — there's some typos that are repeated multiple times (copy-paste?)

Please wait for Alex and Maksim's reviews before merging. Thanks!

benallfree:
@mathiasbynens Thanks for your edits.

src/common/HTTPRequest.ts
```diff
@@ -160,6 +170,82 @@ export class HTTPRequest {
...
+   * Awaits pending interception handlers and then decides how to fulfill
+   * the request interception.
+   */
+  async finalizeInterceptions(): Promise<void> {
+    await this._interceptActions.reduce(
+      (promiseChain, interceptAction) =>
+        promiseChain.then(interceptAction).catch((error) => {
+          // This is here so cooperative handlers that fail do not stop other handlers
+          // from running
+          console.error(error);
```
jschfflr:
```suggestion
```

src/common/NetworkManager.ts
```diff
@@ -382,6 +382,11 @@ export class NetworkManager extends EventEmitter {
     );
     this._requestIdToRequest.set(event.requestId, request);
     this.emit(NetworkManagerEmittedEvents.Request, request);
+    request.finalizeInterceptions().catch((error) => {
+      // This should never happen, but catch just in case.
+      console.error(error);
```
jschfflr:
```suggestion
```

jschfflr:
@googlebot I fixed it.

jschfflr:
@googlebot I fixed it.

src/common/HTTPRequest.ts
```diff
@@ -160,6 +170,81 @@ export class HTTPRequest {
     return this._url;
   }
 
+  /**
+   * @returns the `ContinueRequestOverrides` that will be used
+   * if the interception is allowed to continue (ie, `abort()` and
+   * `respond()` aren't called).
+   */
+  continueRequestOverrides(): ContinueRequestOverrides {
+    assert(this._allowInterception, 'Request Interception is not enabled!');
```
sadym-chromium, line 179: Maybe add a link to doc or some more instructions? I imaging myself seeing this message and trying to understand "And how can I fix this?"

Same for the other similar asserts.

src/common/HTTPRequest.ts
```diff
@@ -160,6 +170,81 @@ export class HTTPRequest {
...
+      (promiseChain, interceptAction) =>
+        promiseChain.then(interceptAction).catch((error) => {
+          // This is here so cooperative handlers that fail do not stop other handlers
+          // from running
+          debugError(error);
+        }),
+      Promise.resolve()
+    );
+    const [resolution] = this.interceptResolution();
+    switch (resolution) {
```
sadym-chromium, line 238: Maybe add a debug message of how the interceptions were resolved with which priority?

test/requestinterception-experimental.spec.ts
```diff
@@ -0,0 +1,854 @@
...
+        expect(request.resourceType()).toBe('document');
+        expect(request.frame() === page.mainFrame()).toBe(true);
+        expect(request.frame().url()).toBe('about:blank');
+        request.continue({}, 0);
+      });
+      const response = await page.goto(server.EMPTY_PAGE);
+      expect(response.ok()).toBe(true);
+      expect(response.remoteAddress().port).toBe(server.PORT);
+    });
+    it('should work when POST is redirected with 302', async () => {
```
sadym-chromium, line 114: I'm having trouble trying to understand how this test checks if redirection and 302 works. What is the assertion? Could you please add some code comments?

test/requestinterception-experimental.spec.ts
```diff
@@ -0,0 +1,854 @@
...
+
+      server.setRedirect('/rrredirect', '/empty.html');
+      await page.setRequestInterception(true);
+      page.on('request', (request) => {
+        const headers = Object.assign({}, request.headers(), {
+          foo: 'bar',
+        });
+        request.continue({ headers }, 0);
+
+        expect(request.continueRequestOverrides()).toEqual({ headers });
```
sadym-chromium, line 143: shouldn't you verify this code is called at all?

sadym-chromium (approved):
LGTM with couple of minor comments

src/common/HTTPRequest.ts
benallfree: Yes, I think many in the community would be grateful for a link. It's a rather ambiguous error message, along with the `already handled` error.

src/common/HTTPRequest.ts
benallfree: I assume not a `console.log`, what were you imagining for logging?

sadym-chromium: I meant something like `const debugProtocolSend = debug('puppeteer:protocol:SEND ►');`,but now I'm quite sceptical on introducing a new log instance.

Please disregard my previous comment.

sadym-chromium (approved):

src/common/HTTPRequest.ts
jschfflr: Let's ignore this issue for this cl as it's not introducing this message but only picking it up.
I'll file an issue to improve error messages in the future though.

test/requestinterception-experimental.spec.ts
jschfflr: I think these tests are just copied from the original `requestinterception.spec.ts` and I added a pull request to add a comment there to explain this tests origin. Suggesting the same comment here:

```suggestion
    // @see https://github.com/puppeteer/puppeteer/pull/3105
    it('should work when POST is redirected with 302', async () => {
```

test/requestinterception-experimental.spec.ts
jschfflr: I think this tests primary intention is to check that the navigation does not time out.
Adding a suggestion with a comment.

test/requestinterception-experimental.spec.ts
```diff
@@ -0,0 +1,854 @@
...
+      });
+      await page.goto(server.PREFIX + '/rrredirect');
```
jschfflr, line 146:
```suggestion
      // Make sure that the goto does not time out.
      await page.goto(server.PREFIX + '/rrredirect');
```

jschfflr:
@benallfree Whohooo, this change is finally merged 🎉 🎉 🎉

benallfree:
Fantastic!!! 🎉🎉🥳

jschfflr:
Thank you so much for your contribution! 
I'm very much looking forward to seeing this being adopted :)

Androbin:
I believe this PR introduced a regression regarding issue #7038. The logic breaks when the handler needs to be called twice on the same request in the following event order: `_onRequestPaused`, `_onRequestWillBeSent`, `_onRequestPaused` (see https://crbug.com/1196004). The issue was fixed by #7060 but it requires calling the handler more than once for the same request. Please take a look at the [documentation](https://github.com/puppeteer/puppeteer/blob/d01aa6c84a1e41f15ffed3a8d36ad26a404a7187/src/common/NetworkManager.ts#L74) that was added on the matter.

benallfree:
@Androbin I’m the main author of this PR, can you please provide a little more explanation? Do you mean that this feature introduces a regression even when it is not enabled?

benallfree:
Tagging

https://github.com/product-os/jellyfish/pull/7068 
https://github.com/puppeteer/puppeteer/issues/7475 
https://github.com/puppeteer/puppeteer/issues/6696

PR #6737, opened 2021-01-07
chore(agnostification): eliminate common's dependent on isNode check.

webee:
related to #6125
by add a Config module which contains the node specific packages(ws, fs etc), let the common depend on it(dependeny injection), we can remove isNode check in common code, as well as dynamic imports.
every entry(node, web) needs initialize the Config before export the Puppeteer object.
this way, we may support some kind of support for filesystem in the browser(normal webpage or chrome extension).


mathiasbynens:
Jack, PTAL

src/common/BrowserConnector.ts
```diff
@@ -14,14 +14,13 @@
  * limitations under the License.
  */
 
+import Config from './Config.js';
```
jackfranklin: I really like this PR, thank you! 

How do you feel about changing `Config` to `Environment` ? I think that might read slightly nicer.

src/common/Config.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2020 Google Inc. All rights reserved.
+ * Copyright 2017 Google Inc. All rights reserved.
```
jackfranklin: this should stay as 2020

src/common/Config.ts
```diff
@@ -14,9 +14,20 @@
  * limitations under the License.
  */
 
-import { isNode } from '../environment.js';
+interface Config {
+  WebSocketTransportClass: any;
```
jackfranklin: We have `ConnectionTransport` as a defined interface which we could use here rather than `any`

src/common/Config.ts
```diff
@@ -14,9 +14,20 @@
...
+  fetch: any;
+  fs: any;
+  path: any;
```
jackfranklin: I'm not sure I love defining these as any. 

I think we could do `typeof import('fetch') | null` or something like that?

src/common/DOMWorld.ts
```diff
@@ -292,12 +292,12 @@ export class DOMWorld {
     }
 
     if (path !== null) {
-      if (!isNode) {
+      const fs = Config.fs;
+      if (!fs) {
```
jackfranklin: The problem with this change is that one of the things we have planned is that we will bundle a web version of Puppeteer that uses a tool like Rollup to minify and eliminate dead code.

Our plan was to have a bundle step where Rollup can eliminate any `if (isNode)` code branches as we'll tell it that `isNode` is false.

I wonder if we should change this to be `if (!Environment.isNode)` because then we can eliminate these branches at bundling time for our web version.

src/common/Debug.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2020 Google Inc. All rights reserved.
+ * Copyright 2017 Google Inc. All rights reserved.
```
jackfranklin: date shouldn't change

src/common/Debug.ts
```diff
@@ -14,70 +14,17 @@
...
-export const debug = (prefix: string): ((...args: unknown[]) => void) => {
-  if (isNode) {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    return require('debug')(prefix);
-  }
-
-  return (...logArgs: unknown[]): void => {
-    const debugLevel = globalThis.__PUPPETEER_DEBUG as string;
-    if (!debugLevel) return;
+const _debugs = new Map();
```
jackfranklin: I think this could do with some comments around it - took me a while to realise you're storing debugs for a given prefix.

src/node/fs.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
```
jackfranklin: 2021

src/node/fs.ts
```diff
@@ -0,0 +1,32 @@
...
+ * fs.promises is experimental, and therefore not marked as enumerable. That
+ * means when TypeScript compiles an `import('fs')`, its helper doesn't spot the
+ * promises declaration and therefore on Node <10.17 you get an error as
+ * fs.promises is undefined in compiled TypeScript land.
+ *
+ * See https://github.com/puppeteer/puppeteer/issues/6548 for more details.
+ *
+ * Once Node 10 is no longer supported (April 2021) we can remove this and use
+ * `(await import('fs')).promises`.
+ */
```
jackfranklin: I don't think this comment is now needed

jackfranklin: Or rather: it's out of date, because it talks about loading in the promises part of the library but the code below makes no mention of promises.

jackfranklin (commented):
Thanks for this PR! I have a few comments and considerations with regards to future web optimisations we can make & eliminating Node code from our web bundle all together, but I like this direction and idea 👍

src/common/Config.ts
webee: the WebSocketTransportClass is just the Transport factory which should have a static create() method that return a ConnectionTransport instance.
so I need to define a factory interface for WebSocketTransportClass.

src/common/Config.ts
webee: you're right, I should define these as explicit clear types.
I'm not so familiar with typescript, I'll look into it.

src/common/DOMWorld.ts
webee: I agree with you about using `if(!Environment.xxx)` to check for dead code, but I don't think `isNode` is the good choice, because we can config the `fs` module for the web if needed(that's more flexible), so we can't say it's always dead code,  it is determined by `fs` itself not `isNode`.
If `if (!Environment.fs)` works in bundle tools, I think it's better.

src/common/Debug.ts
webee: you're right, I'll give it some comments.
Because I can't use `Config.debug` in the top level of a module, it's not initialized at this time, so I define this `debug` method to define real debug logger lazily.

src/node/fs.ts
webee: I'm not sure if this `src/node/fs.ts` is necessary, or I can just import it in the `initialize-node.ts`?

mathiasbynens:
Jack, could you PTAL and see if your feedback was addressed? Thanks!

src/common/Environment.ts
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CreateConnectionTransport } from './ConnectionTransport.js';
+import fetch from 'node-fetch';
```
jackfranklin: I think we should do this inline like we do for the other types:

```
fetch?: typeof import('node-fetch') | null
```

Is there a reason you didn't do this - just checking I'm not missing something 👍

jackfranklin:
This is looking really good, thanks! I have one final though and then I think we can try to get this in.

src/common/Environment.ts
webee: No, I think the method you mentioned is more reasonable, I'll fix it.

webee: Yes, `typeof import('node-fetch')` is not the fetch function type.
Finally, I found the following method is ok:
```typescript
  fetch?: typeof import('node-fetch').default | null
```
I've made a fix commit.

webee: and, anyway, the types of `node-fetch.fetch` and browser `globalThis.fetch` are not compatible,
so I choose to disable typescript check in `initialize-web.ts`.

PR #6739, opened 2021-01-08
chore: add GitHub Action workflow for main CI checks

mathiasbynens:
This runs all the checks our current Travis setup does, except it only does it on Ubuntu/Linux for now. I’d like to address other OSes in separate patches. PTAL

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,21 +41,27 @@ jobs:
       - name: Run unit tests
         env:
           CHROMIUM: true
-        run: |
-          npm run unit
+        uses: GabrielBB/xvfb-action@v1
+        with:
+          run: |
+            npm run unit
 
       - name: Run unit tests with coverage
         env:
           CHROMIUM: true
-        run: |
-          npm run unit-with-coverage
-          npm run assert-unit-coverage
+        uses: GabrielBB/xvfb-action@v1
+        with:
+          run: |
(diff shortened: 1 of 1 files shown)


mathiasbynens:
Jack, given your experience setting up our current CI, PTAL.

.github/workflows/main.yml
```diff
@@ -0,0 +1,79 @@
...
+        run: |
+          npm run lint
+          npm run generate-docs
+          npm run ensure-correct-devtools-protocol-revision
+
+      - name: Run unit tests
+        env:
+          CHROMIUM: true
+        run: |
+          xvfb-run --auto-servernum npm run unit
```
jackfranklin, line 53: for my understanding: why do we need xvfb-run now?

jackfranklin (approved):

.github/workflows/main.yml
mathiasbynens: GitHub Actions workflows run in a headless environment. We need xvfb on Travis CI as well: https://github.com/puppeteer/puppeteer/blob/9a8479a52a7d8b51690b0732b2a10816cd1b8aef/.travis.yml#L2

mathiasbynens: Example error log without xvfb: https://github.com/puppeteer/puppeteer/runs/1668553136#step:6:1050

PR #6754, opened 2021-01-16
chore: remove coverage dependency on Events.ts

giakhangdouble:
Another step towards removing Events.ts. This commit updates the
coverage utils to instead parse each module and check for the presence
of an `XEmittedEvents` object which indicates that module has events
that we need to check coverage on.

I also refactored how we store the list of what we consider public API
to make it easier to use in coverage and then in the
`installAsyncStackHelper` that gets run in `initialize.ts`.src/public-api.ts


mathiasbynens:
@giakhangdouble, it looks like this conflicts with our main branch. Could you rebase please?

Jack, could you PTAL?

giakhangdouble:
😭😭😭

PR #6759, opened 2021-01-18
feat: add page.emulateNetworkConditions

jschfflr:
@mathiasbynens Please take a look!

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -42,6 +42,7 @@
   * [puppeteer.errors](#puppeteererrors)
   * [puppeteer.executablePath()](#puppeteerexecutablepath)
   * [puppeteer.launch([options])](#puppeteerlaunchoptions)
+  * [puppeteer.networkConditions](#puppeteernetworkconditions)
   * [puppeteer.product](#puppeteerproduct)
   * [puppeteer.registerCustomQueryHandler(name, queryHandler)](#puppeteerregistercustomqueryhandlername-queryhandler)
   * [puppeteer.unregisterCustomQueryHandler(name)](#puppeteerunregistercustomqueryhandlername)
@@ -128,6 +129,7 @@
   * [page.emulateIdleState(overrides)](#pageemulateidlestateoverrides)
   * [page.emulateMediaFeatures(features)](#pageemulatemediafeaturesfeatures)
   * [page.emulateMediaType(type)](#pageemulatemediatypetype)
+  * [page.emulateNetworkConditions(networkConditions)](#pageemulatenetworkconditionsnetworkconditions)
   * [page.emulateTimezone(timezoneId)](#pageemulatetimezonetimezoneid)
   * [page.emulateVisionDeficiency(type)](#pageemulatevisiondeficiencytype)
   * [page.evaluate(pageFunction[, ...args])](#pageevaluatepagefunction-args)
@@ -609,6 +611,26 @@ const browser = await puppeteer.launch({
 > In [puppeteer.launch([options])](#puppeteerlaunchoptions) above, any mention of Chromium also applies to Chrome.
 >
 > See [`this article`](https://www.howtogeek.com/202825/what%E2%80%99s-the-difference-between-chromium-and-chrome/) for a description of the differences between Chromium and Chrome. [`This article`](https://chromium.googlesource.com/chromium/src/+/lkgr/docs/chromium_browser_vs_google_chrome.md) describes some differences for Linux users.
+
+#### puppeteer.networkConditions
+- returns: <[Object]>
+
+Returns a list of network conditions to be used with [`page.emulateNetworkConditions(networkConditions)`](#pageemulatenetworkconditionsnetworkconditions). Actual list of
+conditions can be found in [`src/common/NetworkConditions.ts`](https://github.com/puppeteer/puppeteer/blob/main/src/common/NetworkConditions.ts).
+
+```js
+const puppeteer = require('puppeteer');
+const slow3G = puppeteer.networkConditions['Slow 3G'];
+
+(async () => {
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+  await page.emulateNetworkConditions(slow3G);
+  await page.goto('https://www.google.com');
+  // other actions...
+  await browser.close();
+})();
+```
 
 #### puppeteer.product
 - returns: <[string]> returns the name of the browser that is under automation (`"chrome"` or `"firefox"`)
@@ -1439,6 +1461,29 @@ await page.evaluate(() => matchMedia('screen').matches);
 // → true
 await page.evaluate(() => matchMedia('print').matches);
 // → false
+```
+
+#### page.emulateNetworkConditions(networkConditions) 
+- `networkConditions` <?[Object]> Passing `null` disables network condition emulation.
+  - `download` <[number]> Download speed (bytes/s), `-1` to disable
+  - `upload` <[number]> Upload speed (bytes/s), `-1` to disable
+  - `latency` <[number]> Latency (ms), `0` to disable
+- returns: <[Promise]>
+
+> **NOTE** This does not affect WebSockets and WebRTC PeerConnections (see https://crbug.com/563644)
+
+```js
+const puppeteer = require('puppeteer');
+const slow3G = puppeteer.networkConditions['Slow 3G'];
+
+(async () => {
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+  await page.emulateNetworkConditions(slow3G);
+  await page.goto('https://www.google.com');
+  // other actions...
+  await browser.close();
+})();
 ```
 
 #### page.emulateTimezone(timezoneId)
--- a/src/common/NetworkConditions.ts
+++ b/src/common/NetworkConditions.ts
@@ -1,0 +1,32 @@
+/**
+ * Copyright 2021 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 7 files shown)


docs/api.md
```diff
@@ -610,6 +612,26 @@ const browser = await puppeteer.launch({
 >
 > See [`this article`](https://www.howtogeek.com/202825/what%E2%80%99s-the-difference-between-chromium-and-chrome/) for a description of the differences between Chromium and Chrome. [`This article`](https://chromium.googlesource.com/chromium/src/+/lkgr/docs/chromium_browser_vs_google_chrome.md) describes some differences for Linux users.
 
+#### puppeteer.networkConditions
+- returns: <[Object]>
+
+Returns a list of network conditions to be used with [`page.emulateNetworkConditions(networkConditions)`](#pageemulatenetworkConditions). Actual list of
+devices can be found in [`src/common/NetworkConditions.ts`](https://github.com/puppeteer/puppeteer/blob/main/src/common/NetworkConditions.ts).
```
mathiasbynens: `s/devices/conditions/`?

docs/api.md
```diff
@@ -1441,6 +1463,25 @@ await page.evaluate(() => matchMedia('print').matches);
 // → false
 ```
 
+#### page.emulateNetworkConditions(networkConditions) 
+- `networkConditions` <[NetworkConditions]> Network conditions to emulate.
+- returns: <[Promise]>
+
+Note: This will not affect WebSockets and WebRTC PeerConnections (https://crbug.com/563644)
```
```suggestion
Note: This does not affect WebSockets and WebRTC PeerConnections (https://crbug.com/563644)

```

src/common/NetworkManager.ts
```diff
@@ -130,14 +161,21 @@ export class NetworkManager extends EventEmitter {
...
-    this._offline = value;
+    await this.emulateNetworkConditions(
+      value ? networkConditionsMap['Offline'] : networkConditionsMap['Online']
+    );
+  }
+
+  async emulateNetworkConditions(
+    networkConditions: NetworkConditions
+  ): Promise<void> {
+    if (this._emulatedNetworkConditions === networkConditions) return;
```
mathiasbynens: Note: this can never be `===` when people call `page.emulateNetworkConditions({})` with an object literal (even when it contains exactly the same values). But perhaps that's ok? We expect the most common usage to be 

```js
const slow3G = …;
await page.emulateNetworkConditions(slow3G);
```

src/common/NetworkConditions.ts
```diff
@@ -0,0 +1,48 @@
+/**
+ * Copyright 2020 Google Inc. All rights reserved.
```
mathiasbynens:
```suggestion
 * Copyright 2021 Google Inc. All rights reserved.
```

mathiasbynens:
@sadym-chromium PTAL as well

docs/api.md
```diff
@@ -1441,6 +1463,29 @@ await page.evaluate(() => matchMedia('print').matches);
 // → false
 ```
 
+#### page.emulateNetworkConditions(networkConditions) 
+- `networkConditions` <[Object]> Network conditions to emulate.
+  - `download` <[number]> Download speed (bytes/s), `-1` to disable
+  - `upload` <[number]> Upload speed (bytes/s), `-1` to disable
+  - `latency` <[number]> Latency (ms), `0` to disable
+  - `connectionType` <?["none"| "cellular2g"| "cellular3g"| "cellular4g"| "bluetooth"| "ethernet"| "wifi"| "wimax"| "other"] Optional connection type
```
sadym-chromium: Could you please add a bit more clarification for the params here? Are params `download`, `upload` and `latency` optional in case of having `connectionType`? Or why the `connectionType` needed at all?

src/common/NetworkConditions.ts
```diff
@@ -0,0 +1,48 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { NetworkConditions } from './NetworkManager.js';
+
+export type NetworkConditionsMap = { [name: string]: NetworkConditions };
+
+export const networkConditionsMap: NetworkConditionsMap = {
```
sadym-chromium: Shouldn't the list cover all the options from the list `"none"| "cellular2g"| "cellular3g"| "cellular4g"| "bluetooth"| "ethernet"| "wifi"| "wimax"` maybe omitting `other`?

src/common/NetworkManager.ts
```diff
@@ -130,14 +161,21 @@ export class NetworkManager extends EventEmitter {
   }
 
   async setOfflineMode(value: boolean): Promise<void> {
-    if (this._offline === value) return;
-    this._offline = value;
+    await this.emulateNetworkConditions(
+      value ? networkConditionsMap['Offline'] : networkConditionsMap['Online']
```
sadym-chromium: Calling `setOfflineMode(true); setOfflineMode(true);` removes conditions previously set by the command `emulateNetworkConditions`, which can be an unexpected side-effect.

src/common/Page.ts
```diff
@@ -693,6 +697,17 @@ export class Page extends EventEmitter {
     return this._frameManager.networkManager().setOfflineMode(enabled);
   }
 
+  /**
+   * @param enabled - When `true`, enables offline mode for the page.
```
sadym-chromium: NIT: JSDOC artefact.

docs/api.md
jschfflr: This is unrelated to the predefined network conditions in NetworkConditions.ts but only changes what `navigator.connection.type` returns. But this seems to be unsupported on desktop anyways so maybe it's best to just omit it. WDYT?

src/common/NetworkConditions.ts
jschfflr: Please take a look at my other comment in regards to the relationship between the list here and the `connectionType`.

src/common/NetworkManager.ts
jschfflr: This is a good point! I'll change the two apis to not interfere with each other.

docs/api.md
sadym-chromium: Omitting seems reasonable for me.
Another option is to rename it to something like `connectionTypeDescription` and to explicitly describe what it intended to do.

UPD: having `networkConditionsMap: NetworkConditionsMap` with quite close naming and very different meaning, I'd vote for omitting to prevent misunderstanding of the API.

src/common/Page.ts
sadym-chromium: Actual param is `networkConditions: NetworkConditions`, and not `@param enabled`

docs/api.md
```diff
@@ -610,6 +612,26 @@ const browser = await puppeteer.launch({
 >
 > See [`this article`](https://www.howtogeek.com/202825/what%E2%80%99s-the-difference-between-chromium-and-chrome/) for a description of the differences between Chromium and Chrome. [`This article`](https://chromium.googlesource.com/chromium/src/+/lkgr/docs/chromium_browser_vs_google_chrome.md) describes some differences for Linux users.
 
+#### puppeteer.networkConditions
```
sadym-chromium, line 615: NIT: naming. We use `networkConditions` as a single instance and as a list of conditions. Can we rename this method to something like `networkConditionsList`?

Another option is to use `networkCondition` as a single instance, but it goes against the CDP naming.

mathiasbynens: It‘s not a method, but rather an object. Given that this is analogous to [`puppeteer.devices`](https://github.com/puppeteer/puppeteer/blob/v5.5.0/docs/api.md#puppeteerdevices), I think we should stick to the current name.

jschfflr: Sounds good to me!

docs/api.md
jschfflr: WDYT about the new naming?

src/common/Puppeteer.ts
```diff
@@ -125,6 +129,31 @@ export class Puppeteer {
     return puppeteerErrors;
   }
 
+  /**
+   * @remarks
+   * Returns a list of network conditions to be used with `page.emulateNetworkConditions(networkConditions)`. Actual list of devices can be found in {@link https://github.com/puppeteer/puppeteer/blob/main/src/common/NetworkConditions.ts | src/common/NetworkConditions.ts}.
```
sadym-chromium: NIT: `devices `
```suggestion
   * Returns a list of network conditions to be used with `page.emulateNetworkConditions(networkConditions)`. Actual list of predefined conditions can be found in {@link https://github.com/puppeteer/puppeteer/blob/main/src/common/NetworkConditions.ts | src/common/NetworkConditions.ts}.
```

mathiasbynens (approved):
Looking good! Thanks

docs/api.md
```diff
@@ -1441,6 +1463,29 @@ await page.evaluate(() => matchMedia('print').matches);
 // → false
 ```
 
+#### page.emulateNetworkConditions(networkConditions)
```
talldan, line 1466: @jschfflr Thanks for working on this feature. As someone looking to switch to using this API to emulate offline testing, I wasn't sure how to transition from `Network.emulateNetworkConditions` (which accepted an `offline` property) to this API. Is calling `page.setOfflineMode(true)` all that's needed?

Unfortunately `setOfflineMode` docs also don't go into detail.

It might be worth mentioning how this should work. If you can clarify I'm happy to update the docs.

jschfflr: Let me take a look at this - I'll get back to you in a sec :)

jschfflr: Yes, `page.setOfflineMode(true)` is all you need. I'll update the documentation accordingly.

jschfflr: Let me know if that helps or if I should further clarify their relationship: https://github.com/puppeteer/puppeteer/pull/7446

jschfflr: I'll close this comment to move the discussion to the pull request.

PR #6775, opened 2021-01-22
fix(domworld): fix waitfor bindings (#6766)

### LINKED ISSUE #6766: Aria queries combined with waitFor* throws unexpected error after page navigation
### Steps to reproduce

* Puppeteer version: v5.5.0
* Platform / OS version: macOS v10.15.7
* URLs (if applicable): https://github.com/kevin940726/puppeteer-aria-query-bug
* Node.js version: 14.15.1

Aria queries companied with `waitFor` will fail to query any elements after page navigation.

```js
await page.waitForSelector('aria/Button[role="button"]'); // This is fine

await page.reload(); // Or any other page navigation

await page.waitForSelector('aria/Button[role="button"]'); // Cannot find the element anymore 💀 
```

I've put together a minimal repro [repo](https://github.com/kevin940726/puppeteer-aria-query-bug) to better demo the issue.

It should still successfully get the elements even after page navigation.

**What happens instead?**
It throws timeout error.

jschfflr:
Please take a look! This should close #6766

### DIFF
--- a/src/common/DOMWorld.ts
+++ b/src/common/DOMWorld.ts
@@ -512,9 +512,12 @@ export class DOMWorld {
     const bind = async (name: string) => {
       const expression = helper.pageBindingInitString('internal', name);
       try {
+        // TODO: In theory, it would be enough to call this just once
         await context._client.send('Runtime.addBinding', {
           name,
-          executionContextId: context._contextId,
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore The protocol defintion is not up to date.
+          executionContextName: context._contextName,
         });
         await context.evaluate(expression);
       } catch (error) {
--- a/src/common/DOMWorld.ts
+++ b/src/common/DOMWorld.ts
@@ -516,7 +516,7 @@ export class DOMWorld {
         await context._client.send('Runtime.addBinding', {
           name,
           // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-          // @ts-ignore The protocol defintion is not up to date.
+          // @ts-ignore The protocol definition is not up to date.
           executionContextName: context._contextName,
         });
         await context.evaluate(expression);
--- a/src/common/ExecutionContext.ts
+++ b/src/common/ExecutionContext.ts
@@ -56,6 +56,10 @@ export class ExecutionContext {
    * @internal
    */
   _contextId: number;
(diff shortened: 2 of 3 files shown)


src/common/DOMWorld.ts
```diff
@@ -512,9 +512,12 @@ export class DOMWorld {
     const bind = async (name: string) => {
       const expression = helper.pageBindingInitString('internal', name);
       try {
+        // TODO: In theory, it would be enough to call this just once
         await context._client.send('Runtime.addBinding', {
           name,
-          executionContextId: context._contextId,
+          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+          // @ts-ignore The protocol defintion is not up to date.
```
johanbay:
```suggestion
          // @ts-ignore The protocol definition is not up to date.
```

src/common/DOMWorld.ts
```diff
@@ -512,9 +512,12 @@ export class DOMWorld {
     const bind = async (name: string) => {
       const expression = helper.pageBindingInitString('internal', name);
       try {
+        // TODO: In theory, it would be enough to call this just once
```
johanbay, line 515: Would the binding also survive page navigations if we only call it once?

johanbay (approved):
Nice detective work @jschfflr! I was not even aware that contexts have both a name and an id. Do you know whether or not it is a bug that names are added to `m_activeBindings` when using `executionContextId` but not when using `executionContextName`?

mathiasbynens:
Thanks for taking a look, @johanbay! 👋 

LGTM with Johan's questions answered

mathiasbynens (approved):

src/common/DOMWorld.ts
jschfflr: The binding is added to `m_activeBindings` in both cases and stays there until the binding is removed by calling the `removeBinding` cdp command. And whenever a new `ExecutionContext` is created, it automatically restores the bindings. This works except for the case when the binding is only exposed to a single execution context identified by an id because it will not be able to restore it.

Adding the binding by name will therefore restore it after navigations. I also added a test for that case to make sure that we don't regress.

johanbay: Ah, I see. It sounds like addressing contexts by name instead of id is quite appropriate for this use case then. Thanks!

Issue #6726, opened 2021-01-05
Migrate CI to GitHub Actions


mathiasbynens:
Travis has effectively rendered their free plan useless, so we'll have to migrate to something else. Builds are currently stalling, and PRs cannot be merged without admin overrides (and without knowing whether the patch is safe to land!). Even before Travis made this change, builds have been unreliable.

GitHub Actions seems like a decent alternative.

For reference, here's our current Travis config: https://github.com/puppeteer/puppeteer/blob/cac540be3ab8799a1d77b0951b16bc22ea1c2adb/.travis.yml

mathiasbynens:
#6739 landed the main CI checks for Linux. What's left to do is to set things up so these can run on Windows and macOS as well.

mathiasbynens:
Now that Travis can no longer hang our CI status checks, we can [enable auto-merge on pull requests](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/automatically-merging-a-pull-request) 🥳  I’ve gone ahead and done this. PRs can now be merged automatically when all required reviews are met and status checks are passed. No more waiting on CI before manually merging!

mathiasbynens:
I’ve revoked Travis CI’s access from both this repository and the `puppeteer` GitHub org. I manually deleted its webhooks as well.

PR #6782, opened 2021-01-26
chore(ci): use GitHub Actions to run unit tests on Windows and macOS

mathiasbynens:
Similar to our earlier Travis CI setup, we continue to run exhaustive checks on Linux, while also verifying the build + unit tests still work on other platforms.

Issue: #6726

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -135,9 +135,11 @@ jobs:
         run: |
           # Test platform-specific browser binary fetching for both
           # Chromium and Firefox.
-          PUPPETEER_PRODUCT=firefox npm install
+          $env:PUPPETEER_PRODUCT='firefox'
           npm install
-          ls .local-chromium .local-firefox
+          Remove-Item Env:\PUPPETEER_PRODUCT
+          npm install
+          Get-ChildItem -Path .local-chromium,.local-firefox
 
       - name: Build
         run: |
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,8 @@ on:
       - main
 
 jobs:
-  checks:
+  linux-main-checks:
+    # https://github.com/actions/virtual-environments#available-environments
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -77,3 +78,75 @@ jobs:
...
+          # https://github.com/nodejs/Release#release-schedule
+          node-version: 10
+
+      - name: Install dependencies
+        run: |
+          # Test platform-specific browser binary fetching for both
+          # Chromium and Firefox.
+          $env:PUPPETEER_PRODUCT='firefox'
+          npm install
+          Remove-Item Env:\PUPPETEER_PRODUCT
```
jschfflr, line 140: On the other platforms you're only running `npm install` once - is there a reason why windows is treated differently here?

jschfflr (approved):
LGTM % question.
Is there also maybe a way to define the common steps just once?

.github/workflows/main.yml
mathiasbynens: The intention is to run `npm install` twice (once for Firefox, once for Chromium), and AFAICT that’s what we‘re doing on all platforms — see lines 36-37 for Linux, 103-104 for macOS. What makes you say otherwise?

mathiasbynens:
> Is there also maybe a way to define the common steps just once?

I initially tried doing that with [`needs:`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) but it got a little unwieldy pretty fast. IMHO this is a case where it's more readable/maintainable to have a small amount of duplication.

.github/workflows/main.yml
jschfflr: Ah sorry, I missed the fact that the other platforms allow putting the `npm install` in the same line as  `PUPPETEER_PRODUCT=firefox`!

jschfflr:
Sounds good, thanks!

mathiasbynens:
Maksim, could you also PTAL?

sadym-chromium:
> > Is there also maybe a way to define the common steps just once?
> 
> I initially tried doing that with [`needs:`](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idneeds) but it got a little unwieldy pretty fast. IMHO this is a case where it's more readable/maintainable to have a small amount of duplication.

I have the same concern about having lots of copy-pasted code. But if the `needs` implementation is even more bulky, let's stick to this one for now.

sadym-chromium:
Do we need to add those jobs somewhere, or they all of them run automatically in workflows?

mathiasbynens:
> Do we need to add those jobs somewhere, or they all of them run automatically in workflows?

They're all run automatically.

.github/workflows/main.yml
```diff
@@ -77,3 +78,75 @@ jobs:
...
+          # https://github.com/nodejs/Release#release-schedule
+          node-version: 10
+
+      - name: Install dependencies
+        run: |
+          # Test platform-specific browser binary fetching for both
+          # Chromium and Firefox.
+          PUPPETEER_PRODUCT=firefox npm install
+          npm install
+          ls .local-chromium .local-firefox
```
sadym-chromium, line 105: Is it for debugging reasons? Do we need it?

.github/workflows/main.yml
```diff
@@ -77,3 +78,75 @@ jobs:
...
+        with:
+          # Test only the oldest maintenance LTS Node.js version.
+          # https://github.com/nodejs/Release#release-schedule
+          node-version: 10
+
+      - name: Install dependencies
+        run: |
+          # Test platform-specific browser binary fetching for both
+          # Chromium and Firefox.
+          PUPPETEER_PRODUCT=firefox npm install
```
sadym-chromium, line 103: Either add `Run unit tests on Firefox` or don't install Firefox. Same for Windows.

.github/workflows/main.yml
mathiasbynens: Our previous Travis CI setup did the same thing. I left it in on purpose, as it improves the log output (which helps to post-mortem-debug CI failures).

.github/workflows/main.yml
mathiasbynens: For macOS and Windows, we don't run exhaustive checks on purpose: testing everything on all platforms would take too much time. This line tests that the platform-specific binary fetching works properly on macOS (see the comment above), which is useful independently of the unit tests.

We intentionally don't run Firefox unit tests _on all platforms_ (instead, we only do so on Linux, as part of our exhaustive checks), but we do check whether `npm install` (which downloads the platform-specific binary) works.

sadym-chromium (approved):
LGTM

PR #6805, opened 2021-02-02
feat: use `captureBeyondViewport` in `Page.captureScreenshot`

sadym-chromium:
BREAKING CHANGE:
- `page.screenshot` makes a screenshot with the clip dimensions, not cutting it by the ViewPort size.


mathiasbynens:
> `page.screenshot` makes a screenshot with the clip dimensions, not cutting it by the ViewPort size.

Are you saying this restores the pre-v6.0.0 behavior?

test/screenshot.spec.ts
```diff
@@ -50,26 +50,8 @@ describe('Screenshots', function () {
...
-          height: 100,
-        },
-      });
-      expect(screenshot).toBeGolden('screenshot-offscreen-clip.png');
-    });
-    // TODO: remove after the screenshot is fixed.
-    // https://crbug.com/1173457
     itFailsFirefox(
-      'should clip elements to the viewport size with artefacts',
+      'should get screenshot bigger then the viewport',
```
mathiasbynens:
```suggestion
      'should get screenshot bigger than the viewport',
```

test/screenshot.spec.ts
```diff
@@ -50,26 +50,8 @@ describe('Screenshots', function () {
       });
       expect(screenshot).toBeGolden('screenshot-clip-rect.png');
     });
-    // TODO: enable after the screenshot is fixed.
```
mathiasbynens, line 53: Why is this test being removed?

sadym-chromium:
@mathiasbynens 
> Are you saying this restores the pre-v6.0.0 behavior?

* Before 90, the screenshot is cut based on the ViewPort position and size. 
* In 90 without `captureBeyondViewport`, the screenshot is cut based on the ViewPort size and clip position with some unwilling artefacts out of the ViewPort.
* In 90 with `captureBeyondViewport` the screenshot is cut based on the clip position and size.

From this point, using `captureBeyondViewport` is still a breaking change, but it makes the screenshots work properly.

test/screenshot.spec.ts
sadym-chromium: The bug https://crbug.com/1173457 was closed with the motivation: if you need a screenshot bigger than the ViewPort, you have to use the flag `captureBeyondViewport `.

This change uses the flag, screenshot works in the new way, but in the way can easily be called "correctly", the same way Firefox does it. No further `TODO` needed.

It used to be 2 similar tests with the different expectations:
* skipped `should clip elements to the viewport size without artefacts`
* `should clip elements to the viewport size with artefacts`

I removed one of them and adjusted the expectations. It covers the issue point and the breaking change, verifying the screenshot works correctly.

mathiasbynens (approved):

test/screenshot.spec.ts
mathiasbynens: Ack, thanks!

mahnunchik:
Why `captureBeyondViewport` option doesn't documented? https://github.com/puppeteer/puppeteer/blob/v8.0.0/docs/api.md#pagescreenshotoptions

mahnunchik:
@sadym-chromium could you please make `captureBeyondViewport` configurable. To be able to not use this patch https://github.com/puppeteer/puppeteer/issues/7043#issuecomment-812068853

sadym-chromium:
PR #7063 should address the issue.

jschfflr (commented):
LGTM!

PR #6815, opened 2021-02-04
fix(launcher): output correct error message for browser

jackfranklin:
When running `npm run release` today I got this error logged:

```
Error: Could not find browser revision 848005. Run "PUPPETEER_PRODUCT=firefox npm install" or "PUPPETEER_PRODUCT=firefox yarn install" to download a supported Firefox browser binary.
    at ChromeLauncher.launch (/Users/jacktfranklin/src/puppeteer/lib/cjs/puppeteer/node/Launcher.js:80:27)
```

The error is only partially correct; I did have the browser revision
missing, but I needed the Chromium browser, not Firefox. It turns out
the logic in `Launcher.ts` didn't take this into account; it mistakenly
had been hardcoded to always log out the error as if the Firefox binary
was missing.

This PR updates the message depending on the browser:

Chrome error:
> Error: Could not find expected browser (chrome) locally. Run npm
> install or yarn install to download the correct Chromium revision
> (848005).

Firefox error:
> Error: Could not find expected browser (firefox) locally. Run
> "PUPPETEER_PRODUCT=firefox npm install" or "PUPPETEER_PRODUCT=firefox
> yarn install" to download a supported Firefox browser binary.

### DIFF
--- a/src/node/Launcher.ts
+++ b/src/node/Launcher.ts
@@ -629,8 +629,8 @@ function resolveExecutablePath(
   }
   const revisionInfo = browserFetcher.revisionInfo(launcher._preferredRevision);
 
-  const firefoxHelp = `Run "PUPPETEER_PRODUCT=firefox npm install" or "PUPPETEER_PRODUCT=firefox yarn install" to download a supported Firefox browser binary.`;
-  const chromeHelp = `Run npm install or yarn install to download the correct Chromium revision (${launcher._preferredRevision}).`;
+  const firefoxHelp = `Run \`PUPPETEER_PRODUCT=firefox npm install\` to download a supported Firefox browser binary.`;
+  const chromeHelp = `Run \`npm install\` to download the correct Chromium revision (${launcher._preferredRevision}).`;
   const missingText = !revisionInfo.local
     ? `Could not find expected browser (${launcher.product}) locally. ${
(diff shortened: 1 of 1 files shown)


src/node/Launcher.ts
```diff
@@ -627,8 +628,13 @@ function resolveExecutablePath(
     }
   }
   const revisionInfo = browserFetcher.revisionInfo(launcher._preferredRevision);
+
+  const firefoxHelp = `Run "PUPPETEER_PRODUCT=firefox npm install" or "PUPPETEER_PRODUCT=firefox yarn install" to download a supported Firefox browser binary.`;
+  const chromeHelp = `Run npm install or yarn install to download the correct Chromium revision (${launcher._preferredRevision}).`;
```
mathiasbynens: Now that we're here, I'd suggest dropping the `yarn`-specific command. Those who prefer `yarn` know how to use it, so I don't think this adds anything, and removing it simplifies the error message. WDYT?

```suggestion
  const firefoxHelp = `Run \`PUPPETEER_PRODUCT=firefox npm install\` to download a supported Firefox browser binary.`;
  const chromeHelp = `Run \`npm install\` to download the correct Chromium revision (${launcher._preferredRevision}).`;
```

mathiasbynens (approved):

SyedMuzamilM:
Well I am still getting the same error

PR #6817, opened 2021-02-04
chore: update proxy-from-env

jschfflr:
Also-By: Arjen Korevaar <a.korevaar@mephix.com>


jschfflr:
This is no longer needed.

Issue #6819, opened 2021-02-04
Devtools resize viewport dimensions overlay included in screenshots


marvinhagemeister:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 7.0.0
* Platform / OS version: MacOS Big Sur
* URLs (if applicable): -
* Node.js version: 14.13.0

**What steps will reproduce the problem?**

```js
const puppeteer = require('puppeteer');

(async () => {
    const browser = await puppeteer.launch({
        headless: false,
        devtools: true,
    });
    const page = await browser.newPage();
    await page.goto('https://example.com');

    await page.screenshot({ path: 'example.png', fullPage: true });

    await browser.close();
})();
```

1. Run the above script
2. open `example.png`

**What is the expected result?**

![example](https://user-images.githubusercontent.com/1062408/106923681-a166bd00-670e-11eb-90b4-58870f8aac79.png)

**What happens instead?**

Notice the viewport dimensions in the top right corner of the screenshot. I expect them not to be there. They are seemingly injected by the devtools somehow. They are not present when the devtools panel isn't open. Doesn't seem to be injected into the pages HTML either, otherwise I'd be able to hide it via CSS or something.

![image](https://user-images.githubusercontent.com/1062408/106923749-b0e60600-670e-11eb-8642-b5f6e061a86b.png)

mathiasbynens:
Maksim, could you PTAL? Looks like the overlay (cc @OrKoN) shows up.

OrKoN:
The viewport size overlay is shown when you resize the window and have DevTools opened. So I wonder if it's a regression issue? Maybe the capturing of a screenshot causes the window to resize now.

marvinhagemeister:
Can confirm that this issue is not present in `5.x` or `6.x`. It starts to appear with version `7.0.0`.

OrKoN:
@sadym-chromium PTAL then. Also, when a screenshot is taken via DevTools, DevTools disables all overlays so perhaps Puppeteers implementation should do the same?

sadym-chromium:
On it

sadym-chromium:
Confirmed. I was able to reproduce and to add an automated test for it. Looking into solving the issue.

sadym-chromium:
It looks like the issue is with having 2 Overlay agents controlling Overlays. One from the DevTools session, another from the Puppeteer session. The question is should the Overlay be a global or should we stick to the current behaviour having an overlay manager per CDP session.

[...]

Issue #6823, opened 2021-02-05
Respect screen `deviceScaleFactor` on the headful fullscreen screenshot


sadym-chromium:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: `v7.0.1`
* Platform / OS version: Linux, X-Cinnamon
* Node.js version: v14.9.0

**What steps will reproduce the problem?**

0. Having Linux with DeviceScaleRation `1.25`
1. Set page ViewPort 500x500.
2. Open `example.com`
3. Make a fullscreen screenshot.
4. Verify the screenshot size is 625px X 625px, with the area outside of the rectangle 500px X 500px filled with the background.

```
const puppeteer = require('puppeteer');

(async () => {
    const browser = await puppeteer.launch({
        headless: false,
    });
    const page = await browser.newPage();
    await page.setViewport({ width: 500, height: 500});
    await page.goto('https://example.com');

    await page.screenshot({ path: 'example.png', fullPage: true });

    await browser.close();
})();
```

**What is the expected result?**
`example.png` has a size of 500px x 500px.

**What happens instead?**
`example.png` has a size of 625px x 625px. Not scaled up, but with added background field.

Relevant change in the DevTools: https://crrev.com/c/2637622

sadym-chromium:
* https://crbug.com/1175209
* https://crrev.com/c/2679434

PR #6824, opened 2021-02-05
fix(screenshot): don't show viewport size overlay

sadym-chromium:
* Fix the issue #6819 `Devtools resize viewport dimensions overlay included in screenshots`
* Hide size overlay before taking screenshot.
* Tests (unstable on the environments with the screen scale factor !=1 due to #6823).


test/headful.spec.ts
```diff
@@ -62,6 +68,41 @@ describeChromeOnly('headful tests', function () {
   });
 
   describe('HEADFUL', function () {
+    itFailsFirefox(
+      'headfull with devtools should take fullPage screenshot',
```
mathiasbynens:
```suggestion
      'headful with DevTools should take fullPage screenshot',
```

mathiasbynens (approved):
This is still marked as draft, yet you requested review already — which is it? Happy to take a look :)

src/common/Page.ts
```diff
@@ -1675,6 +1675,11 @@ export class Page extends EventEmitter {
     });
     let clip = options.clip ? processClip(options.clip) : undefined;
 
+    // Not hiding the size overlay will cause unwilling artefacts on the screenshot.
+    await this._client.send('Overlay.setShowViewportSizeOnResize', {
```
OrKoN: Probably we need to set it back to the original value afterwards? Btw, have you found why the viewport is being resized when you take a screenshot?

mathiasbynens:
Please try optimizing `test/golden-chromium/headful-screenshot-grid-fullpage.png` if you haven't already

src/common/Page.ts
mathiasbynens: > Probably we need to set it back to the original value afterwards?

I initially thought this too, but then couldn't immediately think of any reason a Puppeteer script would need the overlays to render. Are you thinking of headful scenarios where the Puppeteer scripts does some set up on the page and then hands it over to the user for some manual interaction?

sadym-chromium:
> This is still marked as draft, yet you requested review already — which is it? Happy to take a look :)

It's not ready yet. I added reviewers too early, sorry.

src/common/Page.ts
OrKoN: Hm, I am not sure. What is the use case for running puppeteer with devtools: true? can you manipulate DevTools window as well? A hypothetical scenario of a test: resizing a window & evaluating that viewport overlay is rendered. Also, for DevTools it'd be probably helpful to capture screenshots with overlays in e2e tests at some point.

sadym-chromium: DevTools doesn't set it back, and there is no method to get the initial value.

sadym-chromium:
> Please try optimizing `test/golden-chromium/headful-screenshot-grid-fullpage.png` if you haven't already

It was optimised, but at the end there is no need in the file at all.

src/common/Page.ts
OrKoN: I see. So do you know why the page is resized when taking a screenshot?

marvinhagemeister: > Hm, I am not sure. What is the use case for running puppeteer with devtools: true? 

We've build a complete test runner around puppeteer over at https://github.com/boxine/pentf/ . To make test inspection as easy as possible for our developers we have a `-d, --debug` flag that opens up the browser in headful mode with devtools already opened, with persistent logging enabled and a bunch of other stuff. They love it because it saves them quite a few clicks. This file (especially the `newPage` function) with the puppeteer enhancements might be worth a look: https://github.com/boxine/pentf/blob/master/src/browser_utils.js

I'm using the same runner to test the devtools extensions for Preact (frontend framework) https://github.com/preactjs/preact-devtools/ . Currently I'm faking the extension inside an iframe, but the long term goal is to test the extension pages inside the devtools panel directly. This would properly test it like the user would experience it and would allow me to verify things like "is my extension tab showing up?", syncing extension settings, etc.

jschfflr:
@sadym-chromium just a gentle ping to see if this is still relevant. If not, feel free to close this. Otherwise, let's get this landed :)

sadym-chromium:
As described [in the comment](https://github.com/puppeteer/puppeteer/issues/6819#issuecomment-776700311), the suggested solution wouldn't work. Closing.

PR #6826, opened 2021-02-06
chore: cleanup remaining Travis references

theDavidBarton:
As it was dropped completely in [v6.0.0](v5.5.0...v6.0.0) the remaining references can be cleaned up about Travis by now:

- **Troubleshooting:** I've reword the present to past tense about _Travis usage on Puppeteer_ and replaced the now broken `.travis.yml` link to a historical one (from [v5.5.0](https://github.com/puppeteer/puppeteer/releases/tag/v5.5.0)). I've found it more useful like this than completely removing the reference.
- **`.npmignore`:** Cleaned up the `.npmignore` from `experimental/puppeteer-firefox`.
- **`doclint/cli.js`:** I also removed the Travis-related condition in _log reminder about tsc (in case of DocLint fails locally)_ introduced with #5652, replaced it with GitHub Actions environment variable.

Follow up to: #6726

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -194,7 +194,7 @@ export CHROME_DEVEL_SANDBOX=/usr/local/sbin/chrome-devel-sandbox
 
 ## Running Puppeteer on Travis CI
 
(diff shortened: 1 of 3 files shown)


mathiasbynens (approved):

mathiasbynens:
Thanks!

PR #6827, opened 2021-02-07
fix type definitions not working when using require()

k-yle:
The type definitions are wrong and don't work at all. 

If you try to use puppeteer v7 in a javascript file using `const puppeteer = require('puppeteer')`, typescript will tell you that it's wrong and you should be doing `const puppeteer = require('puppeteer').default` which is rubbish.

To confirm, enter this into your terminal:
```sh
node -p "require('puppeteer').default"
```

and note how it prints `undefined`. 

Therefore the type definitions are wrong and should be using the `export =` syntax.

### DIFF
--- a/cjs-entry.d.ts
+++ b/cjs-entry.d.ts
@@ -1,3 +1,3 @@
 declare const _default: typeof import('./lib/cjs/puppeteer/node').default;
 
-export default _default;
+export = _default;


k-yle:
@googlebot I signed it!

thomassuckow:
~~@k-yle Do you have the ability to test this also with import? I'm wondering if this combined with an addition to `puppeteer/node.ts` to `export * from "./api-docs-entry"` would expose all the types as well.~~

Would prefer #6837 over this PR.

k-yle:
closing, #6837 fixed this

PR #6836, opened 2021-02-08
fix(domworld): reset bindings when context changes (#6766)

jschfflr:


mathiasbynens (approved):

src/common/DOMWorld.ts
```diff
@@ -115,6 +115,7 @@ export class DOMWorld {
 
   async _setContext(context?: ExecutionContext): Promise<void> {
     if (context) {
+      this._ctxBindings = new Set();
```
mathiasbynens, line 118:
```suggestion
      this._ctxBindings.clear();
```

mathiasbynens (commented):
Ugh I failed to submit this review comment. Jan is already tackling this nit as a follow-up patch

test/ariaqueryhandler.spec.ts
```diff
@@ -195,6 +195,22 @@ describeChromeOnly('AriaQueryHandler', () => {
       await page.waitForSelector('aria/[role="button"]');
     });
 
+    it('should persist query handler bindings across navigations', async () => {
+      const { page, server } = getTestState();
+
+      // Reset page but make sure that execution context ids start with 1.
```
sadym-chromium, line 201: This comment confuses me. How do you assert the ids?

jschfflr: I don't know how how the execution context ids are generated. But it seems like navigating within one origin results in incrementing ids while navigating across origins resets the ids to 1 again. That's why I added the two navigations there.
The previous test just checked by reloading the page which did not reset the ids and therefor did not uncover this edge case.

PR #6855, opened 2021-02-10
fix: wider compat TS types and CI checks to ensure correct type defs

jackfranklin:
This PR improves our TS types further to make sure they are usable in a
TS environment where ES Modules are the target output. Our use of
`export =` is problematic this environment as TypeScript does not allow
`export =` to be used and it errors.

The fix for the type issues to avoid `export =` is to instead define the
functions that you gain access to when you import Puppeteer as top level
functions in our `types.d.ts` file. We can do this by declaring them
explicitly in `src/node.ts`. These are then rolled into `lib/types.d.ts`
at build time. The downside to this is that we have to keep those
declarations in sync with the Puppeteer API; should we add a new method
to the `Puppeteer` class, we must add it to the `nodes.ts` declarations.
However, this could easily be automated by a small script that walks the
AST and generates these. I will do that in a follow-up PR, but I
consider this low risk given how rarely the very top level API of
Puppeteer changes. The nice thing about this approach is we no longer
need our script that hacks on changes to `lib/types.d.ts`.

To avoid yet more releases to fix issues in one particular TS
environment, this PR also includes a suite of example setups that we
test on each CI run. Each sample folder contains `good.ts`, which should
have no TS errors, and `bad.ts`, which should have some errors. The test
first packs Puppeteer into a tar, and then installs it from that tar
into each project. This should replicate how the published package
behaves when it is installed. We then check that we get no errors on
`good.ts`, and the expected errors on `bad.ts`.

We have a variety of test projects that cover both TS and JS source
code, and CJS and ESM imports and outputs.


.github/workflows/main.yml
```diff
@@ -46,6 +46,10 @@ jobs:
           npm run generate-docs
           npm run ensure-correct-devtools-protocol-revision
 
+      - name: Run check that types.d.ts file is valid and detects errors.
+        run: |
+          npm run test-types-file
```
mathiasbynens, line 48: Let's move this to the existing “Run code checks” section

.github/workflows/publish-on-tag.yml
```diff
@@ -15,6 +15,8 @@ jobs:
         run: npm install
       - name: Build
         run: npm run build
+      - name: Test types.d.ts file is correct
+        run: npm run test-types-file
```
mathiasbynens: There should be no need for this, since PRs should only be merged and releases only created once `main.yml` checks have already passed. This is why we don't run any other checks here as well. This will only slow down releases unnecessarily so let's drop it

scripts/test-ts-definition-files.ts
```diff
@@ -0,0 +1,180 @@
...
+  console.log('\nTesting:', folder);
+  const projectLocation = path.join(
+    process.cwd(),
+    PROJECT_FOLDERS_ROOT,
+    folder
+  );
+  const tarLocation = path.relative(projectLocation, tarPath);
+  console.log('===> Installing Puppeteer from tar file');
+  execSync(
+    `cd ${projectLocation} && PUPPETEER_SKIP_DOWNLOAD=1 npm install --no-package-lock ${tarLocation}`,
```
mathiasbynens: What happens if `projectLocation` contains, say, a space? Does this need escaping?

scripts/test-ts-definition-files.ts
```diff
@@ -0,0 +1,180 @@
...
+  }
+  expectedErrors.forEach((expected) => {
+    if (!expectedErrorsThatHaveOccurred.has(expected)) {
+      console.error(
+        `${projectLocation} expected error that was not thrown: ${expected}`
+      );
+      process.exit(1);
+    }
+  });
+  console.log('===> ✅ Type checked correctly.');
```
mathiasbynens:
```suggestion
  console.log('===> ✅ Type-checked correctly.');
```

mathiasbynens (changes requested):

scripts/test-ts-definition-files.ts
jackfranklin: 👍 have added a `replace` to escape spaces

scripts/test-ts-definition-files.ts
```diff
@@ -83,48 +83,57 @@ function packPuppeteer() {
 const tar = packPuppeteer();
 const tarPath = path.join(process.cwd(), tar);
 
+function cdIntoProjectAndRunCommand(projectLocation: string, command: string) {
+  return spawnSync(
+    'sh',
+    ['-c', [`cd "${projectLocation}"`, command].join('; ')],
```
mathiasbynens: We're essentially moving the problem to this one "command" string that we pass to `sh -c` now. But we're still hardcoding the arguments within that command string ourselves, so this still doesn't escape correctly.

Taking a step back, I don't think we need this custom logic at all. Have you considered using `spawnSync(command, { cwd: projectLocation })` instead of creating this helper?

mathiasbynens (changes requested):

mathiasbynens (approved):

PR #6859, opened 2021-02-10
fix: change kill to signal the whole process group to terminate 

petuomin:
In the [PR](https://github.com/puppeteer/puppeteer/pull/6011) that improves the `ctrl+c`, the signal handling may have been accidentally changed to kill just the process instead of the process group.

We are using puppeteer to launch a bash script that starts xvfb and then chrome. Since the bash script receives the SIGKILL, it has no change to do cleanup, leaving a dangling xvfb. Signaling the process group fixes the issue.

Note: I'm not familiar with Windows. The code in this PR is from BrowserRunner.ts before PR6011 was applied. It seems like it should be right based on this: https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/taskkill

Note2: The nodejs documentation states: "Sending signal 0 can be used as a platform independent way to test for the existence of a process." However the documentation does not state whether it will throw an error with code `ESRCH` on Windows.

### DIFF
--- a/src/node/BrowserRunner.ts
+++ b/src/node/BrowserRunner.ts
@@ -186,7 +186,7 @@ export class BrowserRunner {
     if (this.proc && this.proc.pid && pidExists(this.proc.pid)) {
       try {
         if (process.platform === 'win32') {
-          childProcess.execSync(`taskkill /pid ${this.proc.pid} /T /F`);
+          childProcess.exec(`taskkill /pid ${this.proc.pid} /T /F`, () => {});
         } else {
           // on linux the process group can be killed with the group id prefixed with
           // a minus sign. The process group id is the group leader's pid.
--- a/src/node/BrowserRunner.ts
+++ b/src/node/BrowserRunner.ts
@@ -183,9 +183,16 @@ export class BrowserRunner {
     // If the process failed to launch (for example if the browser executable path
     // is invalid), then the process does not get a pid assigned. A call to
     // `proc.kill` would error, as the `pid` to-be-killed can not be found.
-    if (this.proc && this.proc.pid && !this.proc.killed) {
+    if (this.proc && this.proc.pid && pidExists(this.proc.pid)) {
       try {
-        this.proc.kill('SIGKILL');
+        if (process.platform === 'win32') {
+          childProcess.execSync(`taskkill /pid ${this.proc.pid} /T /F`);
+        } else {
+          // on linux the process group can be killed with the group id prefixed with
+          // a minus sign. The process group id is the group leader's pid.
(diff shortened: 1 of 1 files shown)


petuomin:
Ah, the CI is kind enough to provide some answers regarding the windows platform.

mathiasbynens:
@TimvdLippe PTAL since this refers to your earlier patch

TimvdLippe (commented):
Ah that's unfortunate, sorry for causing unexpected regressions. At this point, I think we would benefit from having several tests that we can run to ensure that numerous process configurations and launches are behaving as expected.

Sadly I don't have the cycles myself to work on that, but @petuomin would you in a position to do so?

Ideally, we have a set of tests that can verify that launching Puppeteer as the main process, in bash scripts and on multiple platforms and that all processes are cleaned up appropriately. I am not aware of prior art in this area and I especially don't really know how you would be able to obtain process information on platforms such as Windows. Let me know if you are up for the challenge and I would happily assist wherever possible!

petuomin:
@googlebot I signed it!

petuomin:
No worries, we were able to figure out a workaround for us for now. 

I'll try to find some time to work on this. It might take a while though as I, too, will have to learn some windows.

jschfflr (approved):
This change looks good to me! As it's basically restoring previous (also untested) behaviour, I would be happy to land this for now and figure out a better way to test this in a follow up change. Let me know what you think @TimvdLippe

jschfflr:
@googlebot I signed it!

jschfflr:
@petuomin It looks like the @googlebot couldn't verify that you signed the cla yet. Could maybe check if you did and post another `@googlebot I signed it!`. Thanks!

jschfflr:
@petuomin Also, it looks like your change doesn't quite work on windows yet - could you also look into that?

TimvdLippe:
@jschfflr I am happy with relanding, but I don't think the current fix works quite yet. Unfortunately I don't have the resources to check what is going on, but I have no objections to this PR 😄

jschfflr:
@googlebot I signed it!

jschfflr:
@petuomin I changed the implementation to ignore `taskkill` failures as it looks like all processes will be successfully cleaned on windows anyways. 
Please make sure to get the CLA signed and I'm happy to get this landed!

jschfflr:
@googlebot I consent!

jschfflr:
@petuomin Could you make sure to sign the CLA with the email address you used to create the commit? 
It looks like the google bot didn't catch that one yet :/

OrKoN (approved):

PR #6860, opened 2021-02-10
chore: improve TS automated type tests

jackfranklin:
This PR:

1. Makes sure we remove and freshly install Puppeteer before testing our
   type defs, to avoid running on stale files.
2. Makes the tests run off `puppeteer.tgz` to avoid having version
   numbers in the file name and therefore having to update it when we
   bump versions.

### DIFF
--- a/scripts/test-ts-definition-files.ts
+++ b/scripts/test-ts-definition-files.ts
@@ -1,6 +1,7 @@
 import { spawnSync } from 'child_process';
 import { version } from '../package.json';
 import path from 'path';
+import fs from 'fs';
 const PROJECT_FOLDERS_ROOT = 'test-ts-types';
 const EXPECTED_ERRORS = new Map<string, string[]>([
   [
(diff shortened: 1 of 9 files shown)


scripts/test-ts-definition-files.ts
```diff
@@ -112,12 +122,19 @@ function testProject(folder: string) {
   );
 
   const tarLocation = path.relative(projectLocation, tarPath);
-  console.log('===> Installing Puppeteer from tar file');
+  console.log('===> Clearing left over node_modules to ensure clean slate');
+  // We don't care if this errors because if it did it's most likely because
+  // there was no node_modules folder, which is fine.
+  spawnSync('rm', ['-r', path.join(projectLocation, 'node_modules')], {
```
mathiasbynens: We could also use `fs.rmSync` with the `recursive` option, which seems less indirect. WDYT?

Q: What's `encoding` doing here?

mathiasbynens (approved):
Nice! Just a suggestion and a question

scripts/test-ts-definition-files.ts
jackfranklin: Ah nice idea 👍 

`encoding` was left over from refactor, as we don't need the output we don't need to care about it

PR #6861, opened 2021-02-10
chore: Re-enable Firefox unit tests on Linux with WebRender disabled

jackfranklin:
Trying to debug some of the Firefox test failures we're seeing in:

* https://github.com/puppeteer/puppeteer/pull/6856/
* https://github.com/puppeteer/puppeteer/pull/6857

What's odd is that the failures are always in the `beforeEach`, so it's hard to know exactly what's causing the failures, but they are consistently within this block of tests.

PR #6864, opened 2021-02-11
chore: temporarily disable FF tests on CI

jackfranklin:
The tests are regularly flaking (see #6861 for some investigation). In
the mean time it's blocking us landing and releasing, so we'll
temporarily skip FF tests for now.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -60,11 +60,12 @@ jobs:
           xvfb-run --auto-servernum npm run unit-with-coverage
           xvfb-run --auto-servernum npm run assert-unit-coverage
 
-      - name: Run unit tests on Firefox
-        env:
-          FIREFOX: true
-        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -60,11 +60,12 @@ jobs:
           xvfb-run --auto-servernum npm run unit-with-coverage
           xvfb-run --auto-servernum npm run assert-unit-coverage
 
-      - name: Run unit tests on Firefox
-        env:
-          FIREFOX: true
-        run: |
-          xvfb-run --auto-servernum npm run funit
+      # Temporarily disabled due to FF Nightly flakes.
```
mathiasbynens: Fun fact: the official shorthand is not “FF” but rather “Fx”. https://mths.be/fx

```suggestion
      # Temporarily disabled due to Firefox Nightly flakes.
```

mathiasbynens (approved):

PR #6865, opened 2021-02-11
fix: jsonValue() type is generic

jackfranklin:
During the migration to TS we changed `jsonValue` so it returned
`<Record<string, unknown>>`. This is only true if all the JSON values it
returns are objects; but it could return an array, a string, a number,
etc. Therefore we make the type generic, setting the default to
`unknown`, so the user has control over the type.


jackfranklin:
Test failures look like FF, which #6864 should resolve for now

mathiasbynens (approved):

PR #6868, opened 2021-02-11
fix: add favicon test asset

mathiasbynens:
The favicon was taken from https://github.com/mathiasbynens/small.

Issue: https://bugzilla.mozilla.org/show_bug.cgi?id=1682076


whimboo:
Thanks a lot for adding the favicon! Please note that at least the above mentioned test will now pass in Firefox. I would expect that others might also pass. So `itFailsFirefox` would have to be removed too.

test/assets/offscreenbuttons.html
```diff
@@ -29,8 +30,9 @@
 <button id=btn9>9</button>
 <button id=btn10>10</button>
 <script>
-window.addEventListener('DOMContentLoaded', () => {
```
mathiasbynens, line 32: I wasn't around when this test was created, but I imagine it was written this way assuming the favicon warning would be logged before `DOMContentLoaded`. Now that we have a favicon, we no longer need to try and avoid the error message.

mathiasbynens:
@whimboo Done! Please mark review+ through GitHub's UI once everything looks okay.

whimboo (approved):
Looks fine from my side, but as I'm not a proper reviewer for this repo you might want to seek someone else too.

jackfranklin (approved):

PR #6870, opened 2021-02-11
chore: upgrade @web/test-runner and increase timeout

jackfranklin:
We've seen the odd timeout on the bots; increasing this in the hope that
it helps stop the flakes.

### DIFF
--- a/web-test-runner.config.js
+++ b/web-test-runner.config.js
@@ -17,6 +17,7 @@ const { chromeLauncher } = require('@web/test-runner-chrome');
 
 module.exports = {
   files: ['test-browser/**/*.spec.js'],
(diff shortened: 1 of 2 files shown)


mathiasbynens (approved):

web-test-runner.config.js
```diff
@@ -17,6 +17,7 @@ const { chromeLauncher } = require('@web/test-runner-chrome');
 
 module.exports = {
   files: ['test-browser/**/*.spec.js'],
+  browserStartTimeout: 60 * 1000,
```
mathiasbynens, line 20: Once we’re Node.js 12+ only, we could use `60_000`!

PR #6882, opened 2021-02-12
fix: explicit HTTPRequest.resourceType type defs

### LINKED ISSUE #6854: puppeteer.HTTPRequest resourceType() returns string instead of a resource type enum
### Steps to reproduce

* Puppeteer version: 7.0.4
* Platform / OS version: Mac
* URLs (if applicable): n/a
* Node.js version: v10

I want to ask about difference while migrating from `@types/puppeteer` to the included definitions (thank you btw!)

In the past (with DefinitelyTyped), request resource types were defined ("document", "xhr", etc.):

```
page.on("request", request => {
   const resourceType = request.resourceType(); // puppeteer.ResourceType
});
```

Now, it returns `string`. I see there is a `puppeteer.Protocol.Network.ResourceType` (although those are not lowercased). 

Is there a type we can use for this instead of `string`? Same for `request.method()` which also now returns `string` instead of the old DefinitelyTyped `puppeteer.HttpMethod`.

Thank you!

jackfranklin:
Fixes #6854.

### DIFF
--- a/src/common/HTTPRequest.ts
+++ b/src/common/HTTPRequest.ts
@@ -46,6 +46,13 @@ export interface ResponseForRequest {
 }
 
 /**
+ * Resource types for HTTPRequests as perceived by the rendering engine.
+ *
+ * @public
+ */
+export type ResourceType = Lowercase<Protocol.Network.ResourceType>;
+
+/**
  *
  * Represents an HTTP request sent by a page.
  * @remarks
@@ -108,7 +115,7 @@ export class HTTPRequest {
   private _allowInterception: boolean;
   private _interceptionHandled = false;
   private _url: string;
-  private _resourceType: string;
+  private _resourceType: ResourceType;
 
   private _method: string;
   private _postData?: string;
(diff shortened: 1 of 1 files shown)


src/common/HTTPRequest.ts
```diff
@@ -45,6 +45,26 @@ export interface ResponseForRequest {
   body: string | Buffer;
 }
 
+/**
+ * Resource types for HTTPRequests as perceived by the rendering engine.
+ *
+ * @public
+ */
+export type ResourceType =
```
jschfflr: In my local browser_protocol.pdl file, there are some more `ResourceType`s:
- SignedExchange
- Ping
- CSPViolationReport
- Preflight

Is there a specific reason why they are not included here?

jackfranklin:
@jschfflr please take another look - I discovered the `Lowercase` util so we don't have to duplicate the list :)

jschfflr (approved):
LGTM

PR #6888, opened 2021-02-15
fix: improve TS types for launching browsers

jackfranklin:
This commit tidies up the quite confusing state of all the various types
required to launch a browser. As we saw when upgrading DevTools
(https://source.chromium.org/chromium/chromium/src/+/master:third_party/devtools-frontend/src/test/conductor/hooks.ts;l=77),
we had to define our launch options variable like so:

```ts
const opts: puppeteer.LaunchOptions & puppeteer.ChromeArgOptions & puppeteer.BrowserOptions = {
  ...
};
```

This commit fixes that by introducing `AllNodeLaunchOptions`, which is
defined as the intersection of all the above types.

Additionally, the types defined as `ChromeArgOptions` are actually used
for launching both Chrome and Firefox, so I have renamed them to
`BrowserArgOptions`, and therefore this change is breaking because
anyone using `ChromeArgOptions` in their types will need to switch.

BREAKING CHANGE: renamed type `ChromeArgOptions` to `BrowserArgOptions`

### DIFF
--- a/src/node/LaunchOptions.ts
+++ b/src/node/LaunchOptions.ts
@@ -13,16 +13,37 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+import { BrowserConnectOptions } from '../common/BrowserConnector.js';
+import { Product } from '../common/Product.js';
+
 /**
  * Launcher options that only apply to Chrome.
  *
  * @public
  */
-export interface ChromeArgOptions {
+export interface BrowserLaunchArgumentOptions {
+  /**
+   * Whether to run the browser in headless mode.
+   * @defaultValue true
+   */
   headless?: boolean;
-  args?: string[];
+  /**
+   * Path to a user data directory.
+   * {@link https://chromium.googlesource.com/chromium/src/+/master/docs/user_data_dir.md | see the Chromium docs}
+   * for more info.
+   */
   userDataDir?: string;
+  /**
+   * Whether to auto-open a DevTools panel for each tab. If this is set to
+   * `true`, then `headless` will be set to `false` automatically.
+   * @defaultValue `false`
+   */
   devtools?: boolean;
+  /**
+   * Additional command line arguments to pass to the browser instance.
+   */
+  args?: string[];
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/node/LaunchOptions.ts
```diff
@@ -13,30 +13,110 @@
...
+   */
+  extraPrefsFirefox?: Record<string, unknown>;
 }
+
+/**
+ * Utility type exposed to enable users to define options that can be passed to
+ * `puppeteer.launch` without having to list the set of all types.
+ * @public
+ */
+export type AllNodeLaunchOptions = BrowserArgOptions &
```
jackfranklin: I'm trying to find better names here because `BrowserArgOptions` and `BrowserOptions` are very similar.

I'm wondering about:

* `BrowserArgOptions` => `BrowserLaunchArgumentOptions` 
* `BrowserOptions` => `BrowserConnectOptions`

To make it clearer that `BrowserOptions` can be used when connecting to an existing browser (it's what `puppeteer.connect` takes) and `BrowserArgOptions` is only when launching browsers. 

@mathiasbynens wdyt? If we have to make a breaking change for this we might as well try to make all the names useful.

mathiasbynens: `BrowserLaunchArgumentOptions` + `BrowserConnectOptions` sgtm

mathiasbynens (approved):

mathiasbynens:
@sadym-chromium Please also take a look

src/node/LaunchOptions.ts
sadym-chromium: Agree with Mathias that `BrowserLaunchArgumentOptions ` + `BrowserConnectOptions` looks reasonable.

`LaunchOptions` and `AllNodeLaunchOptions` have confusing names as well.

jackfranklin: Yeah these are confusing :( My hope is that users don't have to worry about the "sub types" and can just use the main type. I'd like to do a wider clean up of these types but they are used in a few places across the codebase and API so I'd rather keep the scope down for now and try to improve the names.

sadym-chromium: SGTM

PR #6891, opened 2021-02-16
fix: type page event listeners correctly

### LINKED ISSUE #6854: puppeteer.HTTPRequest resourceType() returns string instead of a resource type enum

jackfranklin:
This PR fixes the fact that currently if you have:

```ts

})
```

Then `request` will be typed as `any`. We can fix this by defining an
interface of event name => callback argument type, and looking that up
when you call `page.on`.

Fixes #6854.

### DIFF
--- a/src/common/HTTPRequest.ts
+++ b/src/common/HTTPRequest.ts
@@ -40,7 +40,10 @@ export interface ContinueRequestOverrides {
  */
 export interface ResponseForRequest {
   status: number;
-  headers: Record<string, string>;
+  /**
+   * Optional response headers. All values are converted to strings.
+   */
+  headers: Record<string, unknown>;
   contentType: string;
   body: string | Buffer;
 }
@@ -346,7 +349,7 @@ export class HTTPRequest {
    *
    * @param response - the response to fulfill the request with.
    */
-  async respond(response: ResponseForRequest): Promise<void> {
+  async respond(response: Partial<ResponseForRequest>): Promise<void> {
     // Mocking responses for dataURL requests is not currently supported.
     if (this._url.startsWith('data:')) return;
     assert(this._allowInterception, 'Request Interception is not enabled!');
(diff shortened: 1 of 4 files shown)


mathiasbynens (approved):

src/common/HTTPRequest.ts
```diff
@@ -40,7 +40,10 @@ export interface ContinueRequestOverrides {
  */
 export interface ResponseForRequest {
   status: number;
-  headers: Record<string, string>;
+  /**
+   * Optional response headers. All values are converted to strings.
+   */
+  headers: Record<string, any>;
```
jschfflr: Why did you change this to any?

src/common/Page.ts
```diff
@@ -329,6 +329,34 @@ export const enum PageEmittedEvents {
...
+  load: never;
+  metrics: { title: string; metrics: Metrics };
+  pageerror: Error;
+  popup: Page;
+  request: HTTPRequest;
+  response: HTTPResponse;
+  requestfailed: HTTPRequest;
+  requestfinished: HTTPRequest;
+  workercreated: WebWorker;
+  workerdestroyed: WebWorker;
```
jschfflr, line 358: What about adding a [event: string]: any at the bottom here to make sure we don't break places that use other event names? Or is that intended?

src/common/HTTPRequest.ts
jackfranklin: Good point, not sure. I think we stringify them anyway so `unknown` is better here.

src/common/Page.ts
jackfranklin: I'd rather it fail explicitly because we're missing a type.

PR #6895, opened 2021-02-17
Enable unit tests for Firefox on Windows

whimboo:
This will eventually fix issue #5673.

First I would like to see if tests might work again. If not further debugging is necessary.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -155,3 +155,10 @@ jobs:
           CHROMIUM: true
         run: |
           npm run unit
+
+      - name: Run unit tests on Firefox
+        env:
+          FIREFOX: true
+          MOZ_WEBRENDER: 0
+        run: |
+          npm run funit
--- a/src/node/BrowserRunner.ts
+++ b/src/node/BrowserRunner.ts
@@ -24,6 +24,7 @@ import { LaunchOptions } from './LaunchOptions.js';
 import { Connection } from '../common/Connection.js';
 import { NodeWebSocketTransport as WebSocketTransport } from '../node/NodeWebSocketTransport.js';
 import { PipeTransport } from './PipeTransport.js';
+import { Product } from '../common/Product.js';
 import * as readline from 'readline';
 import { TimeoutError } from '../common/Errors.js';
 import { promisify } from 'util';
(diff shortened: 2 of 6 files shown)


whimboo:
It's still busted so I'll have to take a look at https://bugzilla.mozilla.org/show_bug.cgi?id=1656962 first. We do not experience similar issues for other test jobs in our CI.

whimboo:
This PR is ready to be reviewed now. I hope the way that I used is fine for you.

.github/workflows/main.yml
```diff
@@ -155,3 +155,11 @@ jobs:
           CHROMIUM: true
         run: |
           npm run unit
+
+      - name: Run unit tests on Firefox
+        env:
+          FIREFOX: true
+        run: |
+          $env:PUPPETEER_PRODUCT='firefox'
+          $env:MOZ_WEBRENDER='0'
```
mathiasbynens: These can be added to to `env` above. (We only use `$env:` for environment variables which we only need for one (but not all) command within the `run` block)

whimboo: Oh, that is actually a left over from testing. I corrected that.

.github/workflows/main.yml
```diff
@@ -155,3 +155,10 @@ jobs:
           CHROMIUM: true
         run: |
           npm run unit
+
+      - name: Run unit tests on Firefox
+        env:
+          FIREFOX: true
+          MOZ_WEBRENDER: 0
+        run: |
+          npm run funit
```
whimboo, line 164: Lets see if that works. Last time the check on Windows failed because the environment variable for `PUPPETEER_PRODUCT` could not be set.

whimboo: I think that I found the problem and pushed a follow-up commit for it.

mathiasbynens (approved):

src/node/BrowserRunner.ts
```diff
@@ -36,6 +36,7 @@ Please check your open processes and ensure that the browser processes that Pupp
 If you think this is a bug, please report it on the Puppeteer issue tracker.`;
 
 export class BrowserRunner {
+  private _product: string;
```
jschfflr: We already have a type for this in common/Product.js. Can you use that one instead of string?

jschfflr (approved):
LGTM % nit about types

src/node/BrowserRunner.ts
whimboo: Oh, I totally overlooked that `Launcher.product` already returns that type. Will absolute fix that.

whimboo:
@jschfflr I updated the patch and all tests are passing. Would you mind merging the PR?

jschfflr (approved):

PR #6896, opened 2021-02-17
chore: run unit tests for Firefox on MacOS

whimboo:
I noticed that the unit tests for Firefox are only run on Linux, but not on MacOS. To get out of experimental state it would be good to have these running as well.

### DIFF
--- a/test/keyboard.spec.ts
+++ b/test/keyboard.spec.ts
@@ -127,7 +127,7 @@ describe('Keyboard', function () {
       await page.evaluate(() => document.querySelector('textarea').value)
     ).toBe('嗨a');
   });
-  it('should report shiftKey', async () => {
+  itFailsFirefox('should report shiftKey', async () => {
     const { page, server } = getTestState();
 
     await page.goto(server.PREFIX + '/input/keyboard.html');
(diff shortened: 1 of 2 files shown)


whimboo:
@jschfflr any idea why the job for Mac isn't running? 

Update: I had an invalid configuration. Now it works.

test/keyboard.spec.ts
```diff
@@ -127,7 +127,7 @@ describe('Keyboard', function () {
       await page.evaluate(() => document.querySelector('textarea').value)
     ).toBe('嗨a');
   });
-  it('should report shiftKey', async () => {
+  itFailsFirefox('should report shiftKey', async () => {
```
whimboo, line 130: Is there a way to only mark it failing on MacOS?

whimboo:
The failures for `linux-main-checks` seem to be unrelated to my changes.

test/keyboard.spec.ts
mathiasbynens: Not built-in, no — we haven’t needed it until now. You could add another `itFailsFirefoxOnMacOS` helper based on `process.platform === 'darwin'`

whimboo: I'm ok with leaving it as is for now. Note that it is also marked as `pass` or `fail` on our side. We can re-enable when the underlying problem is fixed.

jschfflr (approved):

PR #6924, opened 2021-02-25
Add instructions for headless GPU acceleration

### LINKED ISSUE #6083: disable-gpu-compositing always present when using headless mode
**TL;DR:**

When using `--headless` flag puppeteer always also sets (implicitly) `--disable-gpu-compositing` `--allow-pre-commit-input` flags

Browser: HeadlessChrome/83.0.4103.106
OS: Linux 4.13.0-46-generic

Don't know if this is actual bug, but didn't find anywhere on the Internet info about why that happens or how to remove this flag on headless.

--------------------------------------------------------------------------------------------------------------

**Use case:**
Im using puppeteer to generate PNG and PDF files from html source containing svg element.

Problem occurs when svg contains huge images (e.g. png image of size 12.000x15.000px)

When using window mode whole process of generating pdf/png takes about 45 seconds (not great, not terrible), but by adding just one option `--headless` time increases to over 30 minutes(!).

After hours of investigation, tweaking and adjusting settings it came out that with `--headless` flag puppeteer also sets (implicitly) `--disable-gpu-compositing` `--allow-pre-commit-input` flags, and due to this compositing is "_Software only. Hardware acceleration disabled_". I found no way to turn this off (means to enable gpu compositioning).

net:
Resolves: #6083


net:
@googlebot I signed it!

jschfflr (approved):
This change looks good to me!

jschfflr:
@googlebot I consent.

PR #6935, opened 2021-02-27
test(firefox): ignore false positive

kblok:
In Firefox a `Protocol error` is thrown but because the evaluate command was failing (error below) not because the promises were being rejected.

`'Protocol error (Runtime.callFunctionOn): can't access property "sendAsyncMessage", this.mm is null executeInChild/<@chrome://remote/content/cdp/sessions/TabSession.jsm:78:7\nexecuteInChild@chrome://remote/content/cdp/sessions/TabSession.jsm:73:12\nexecute@chrome://remote/content/cdp/sessions/TabSession.jsm:69:17\nonPacket@chrome://remote/content/cdp/Connection.jsm:227:36\nonMessage@chrome://remote/content/server/WebSocketTransport.jsm:85:18\nhandleEvent@chrome://remote/content/server/WebSocketTransport.jsm:67:14\n'`

### DIFF
--- a/test/page.spec.ts
+++ b/test/page.spec.ts
@@ -37,13 +37,17 @@ describe('Page', function () {
       const { context } = getTestState();
 
       const newPage = await context.newPage();
+      const evaluatePromise = newPage
+        .evaluate(() => new Promise(() => {}))
+        .catch((error_) => (error = error_));
+
+      await newPage.evaluate(() =>
+        console.log('make sure that the previous eval got to the browser')
+      );
+
       let error = null;
-      await Promise.all([
-        newPage
-          .evaluate(() => new Promise(() => {}))
-          .catch((error_) => (error = error_)),
-        newPage.close(),
-      ]);
+      await Promise.all([evaluatePromise, newPage.close()]);
+
       expect(error.message).toContain('Protocol error');
     });
     it('should not be visible in browser.pages', async () => {
--- a/test/page.spec.ts
+++ b/test/page.spec.ts
@@ -35,6 +35,7 @@ describe('Page', function () {
   describe('Page.close', function () {
     it('should reject all promises when page is closed', async () => {
       const { context } = getTestState();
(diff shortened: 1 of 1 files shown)


test/page.spec.ts
```diff
@@ -37,13 +37,17 @@ describe('Page', function () {
       const { context } = getTestState();
 
       const newPage = await context.newPage();
+      const evaluatePromise = newPage
+        .evaluate(() => new Promise(() => {}))
+        .catch((error_) => (error = error_));
```
whimboo, line 43: `error` is defined in line 48 so this will access an undefined variable.

test/page.spec.ts
```diff
@@ -37,13 +37,17 @@ describe('Page', function () {
...
+      await newPage.evaluate(() =>
+        console.log('make sure that the previous eval got to the browser')
+      );
+
       let error = null;
-      await Promise.all([
-        newPage
-          .evaluate(() => new Promise(() => {}))
-          .catch((error_) => (error = error_)),
-        newPage.close(),
```
whimboo, line 45: I would be more in favor of getting this fixed in Firefox.

Since when is it actually failing? Is it triggered by a recent change in Puppeteer? I would be interested in a larger trace log. Maybe you can give some steps to reproduce? Thanks.

kblok: I’m implementing Firefox support on PuppeteerSharp (https://github.com/hardkoded/puppeteer-sharp/pull/1640)

I found that this test was passing here but failing in pptr#.
While debugging this test here, you can do the same in the main branch,
I found that the test was passing, but the “protocol error” I was getting wasn't due to the “target closed” error generated “gracefully” by pptr, but in fact, due to the error I mentioned in the description.

If that can be fixed upstream, perfect :)

test/page.spec.ts
kblok: fixed

PR #6937, opened 2021-02-27
fix(filechooser): cancel is sync

kblok:
Cancel doesn't have async code.
This should be considered a breaking change.


mathiasbynens:
Thanks for the patch!

> This should be considered a breaking change.

Could you please edit the commit message accordingly? See https://www.conventionalcommits.org/en/v1.0.0/#summary for details.

mathiasbynens (approved):

kblok:
How about now @mathiasbynens?

mathiasbynens:
Thanks!

PR #6949, opened 2021-03-02
Clarify error message when node is not clickable

pauladams8:
I've clarified the error message when the browser returns no quads for an element. The previous error message was confusing because the element in question was both an HTMLElement and visible, but not clickable because there was an overlay on the screen with a greater z-index.﻿


pauladams8:
@googlebot I signed it!

jschfflr (approved):
This change looks good to me!

jschfflr:
Thanks for your contribution @pauladams8!

jschfflr:
@googlebot I consent.

Issue #2905, opened 2018-07-18
Request interception and caching compatibility


ObviouslyGreen:
According to this issue, https://github.com/GoogleChrome/puppeteer/pull/1154, caching was disabled during request interception to fix unexpected behavior. It didn't mention in the ticket what didn't work properly, specifics, etc.

Are there any plans on getting caching to work with request interception, or updates on the refactor to fix the chrome limitation that's mentioned in the issue?

aslushnikov:
> According to this issue, #1154, caching was disabled during request interception to fix unexpected behavior. It didn't mention in the ticket what didn't work properly, specifics, etc.

@ObviouslyGreen Let me expand on this. There are a few layers of caching: renderer-based in-memory cache and browser-level cache. Neither of them hits network. Now, request interception is in fact doing *network* interception in Chromium, so it's not getting involved with caching is enabled.

> Are there any plans on getting caching to work with request interception, or updates on the refactor to fix the chrome limitation that's mentioned in the issue?

First of all, the network service refactoring is on it's way in Chromium, but I don't have any ETAs on it. (I personally can't wait for it to happen).

Next, once Network Service is in place, it will be easier for us to teach request interception to work with caching enabled. However, we don't have any plans to do this since there is not enough interest that would justify the effort.

Hope this helps.

JohnBruce:
Consider this me expressing my interest in caching being enabled when request interception is enabled.  Consider the following use case:

While using a proxy, you want caching enabled to save bandwidth cost.  However, you might also want to abort some requests that caching might not save you bandwidth on.  For example, all kinds of images, or one problem I've personally run into a few times, some third party script added via a tag manager or something like it that has outages on its end and prevents load events on the page from ever firing (or delays them beyond navigation timeouts).  

Fwiw, at least as recently as version 1.9.0 allows for caching with requestInterception enabled, so I can use that for a while.  It would be nice not to have to choose between newer versions and features though. 🤷‍♂️ 

Cheers

jadell:
I would also like to see caching enabled with request interception, for the same reasons @JohnBruce mentions. Downgrading is not an option for us at the moment. Please consider this as me expressing my interest in justifying the effort, @aslushnikov

jadell:
[...]

PR #6962, opened 2021-03-06
fix(network): don't disable cache for auth challenge

Androbin:
#729 introduced `_updateProtocolRequestInterception`
Objective: Call `Network.setRequestInterceptionEnabled` for auth challenge or if requested by user

#1154 introduced `Network.setCacheDisabled`
Objective: Ensure custom user response handling is always called

#4260 introduced `_updateProtocolCacheDisabled`
Objective: Call `Network.setCacheDisabled` if requested by user

The cache and request interception work on the network level, while we want the custom user response handling to work independent of the network, which is why we're currently forced to disable the cache in some cases.

As long as there is no custom user response handling, it should be unnecessary to disable the cache for the auth challenge, since it only needs to be called in case of an actual network request anyway.

Related issue: #2905

### DIFF
--- a/src/common/NetworkManager.ts
+++ b/src/common/NetworkManager.ts
@@ -217,7 +217,7 @@ export class NetworkManager extends EventEmitter {
   async _updateProtocolCacheDisabled(): Promise<void> {
     await this._client.send('Network.setCacheDisabled', {
       cacheDisabled:
-        this._userCacheDisabled || this._protocolRequestInterceptionEnabled,
+        this._userCacheDisabled || this._userRequestInterceptionEnabled,
     });
   }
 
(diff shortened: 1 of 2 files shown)


mathiasbynens:
@OrKoN could you PTAL?

OrKoN (commented):
The change LGTM. It seems there are no current tests for this logic. Would it be possible to add the tests so that we capture the right expectations and no one changes the logic back if it does not work for some use cases?

Androbin:
@OrKoN I added a unit test covering this logic.

OrKoN (approved):
Thanks!

Androbin:
How about now @mathiasbynens?

Androbin:
Related: #6996, #7038, and #7060

Issue #6906, opened 2021-02-18
Support .omitBackground as option to .pdf()


rentalhost:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: m85 (remote) / m90 7.1.0 (local)
* Platform / OS version: browserless.io (remote) / Windows 10 x64 (local)
* Node.js version: 14.15.1 (local) and 15.9.0 (local)

**What steps will reproduce the problem?**

`.pdf()` cannot generate a transparent output to be used as watermark.

I have uploaded a .zip containing all tests with a valid output and an invalid output (via puppeteer).

My suggestion is allow `omitBackground` property (like `.screenshot()`) with the `.pdf()` method. So outputs could be used as watermark.

It was discussed here: https://github.com/puppeteer/puppeteer/issues/2545 and a workaround here https://github.com/puppeteer/puppeteer/issues/2545#issuecomment-397590569. It works, but is a hacky way and could stop works at anytime. So a definitive solution is better via `options`.

**What is the expected result?**

Generated PDF output be optionally transparent.

**What happens instead?**

PDF have a opaque background.

**Test**

1. Run `npm i puppeteer`
2. Download pdfbox to use `OverlayPDF` tool.
3. Following example doc, run `node example.js` to generate a PDF via puppeteer. You could set `background-color: transparent`, but it will not help.
4. Run `java -jar pdfbox.jar OverlayPDF puppeteer-output.pdf original.pdf output.pdf` (`original.pdf` is just any PDF document to be overlayed, the "original document" to be watermarked with the puppeteer PDF output)

The zip attached contains all files to test.

Expected output:

![output-valid](https://user-images.githubusercontent.com/421851/108387844-c2d6a700-71ec-11eb-9cf6-ffbb78f411cb.png)
[puppeteer-test.zip](https://github.com/puppeteer/puppeteer/files/6004295/puppeteer-test.zip)

gavin310:
+1. Supporting transparent backgrounds should definitely be a feature without having to use hacks.

d-buckner:
I've added this functionality in the PR above (forgot to update this issue before). It's included in [v9.0.0](https://pptr.dev/#?product=Puppeteer&version=v9.0.0&show=api-pagepdfoptions).

gavin310:
[...]

Issue #2545, opened 2018-05-14
Print to PDF with transparent background


armellarcier:
This is a feature request.

I'm currently developing a small pdf generator tool and I really need custom headers and footers on multipage documents. 

So I print (page.pdf()) the document with margins but no header/footer.

Then I print the header and the footer

Third, I use an other js library (hummus) to put the header and footer on top of every page of the multipage doc.

BUT! the printed header and footer have a white background. This seems ok to not support transparent backgrounds for printing but hey, here is a use case. ;)

Thanks for the great work.

### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.3.0
* Platform / OS version: Mac OSX 10.12.6
* Node.js version: 9.5.0

yonishostak:
+1, need this too

UNDEADgtr:
+1, need this too

genevien:
+1, me too

anttiviljami:
Here's a working proof-of-concept for printing pdf with transparent background:

```
await page.emulateMedia('screen');
await page._emulationManager._client.send(
  'Emulation.setDefaultBackgroundColorOverride',
  { color: { r: 0, g: 0, b: 0, a: 0 } }
);
await page.pdf({ path: 'transparent.pdf', printBackground: true });
```

Have a go :)

UNDEADgtr:
Thanks, @anttiviljami ! It works for me!

armellarcier:
Thanks a lot @anttiviljami ! Our pdf webservice is up and running thx to U!

abdelilah:
There's a slight problem with that solution when enabling displayHeaderFooter:

```
await page.emulateMedia('screen');
await page._emulationManager._client.send(
  'Emulation.setDefaultBackgroundColorOverride',
  { color: { r: 0, g: 0, b: 0, a: 0 } }
);
await page.pdf({
    path: 'transparent.pdf',
    printBackground: true,
    displayHeaderFooter: true // White background comes back after setting this
});
```

kblok:
@aslushnikov any plans to implement this with a PDF option?

aslushnikov:
> any plans to implement this with a PDF option?

@kblok no, we don't have any plans - folks seem to be happy with the workaround.

septatrix:
The workaround seems to work though it took me some time to find it. Would it be possible to state this in the documentation (next to the background option) or add as a separate option (which probably could just to the above)?

septatrix:
[...]

PR #6981, opened 2021-03-13
feat(page): add omitBackground option for page.pdf method

d-buckner:
Adds `omitBackground` option for Page.pdf to allow for transparent backgrounds. This option is available in the screenshot api, this change unifies this for both `.screenshot` and `.pdf`.

Related issues: #6906  #2545


d-buckner:
@googlebot I signed it!

jschfflr (changes requested):
Hi d-rowe, thanks for your contribution! 
Your changes look good to me but it would be great if you could add a test for them.
I'm not sure how stable the pdf output is so maybe you can try testing it with golden snapshots (take a look at the screenshot tests to see how) and if that doesn't work we can find another way to at least have basic test coverage.
Thanks!

d-buckner:
Thanks. Looks like the tests are only setup to compare pngs. Any suggestions for alternative ways to get some basic test coverage?

jschfflr:
Well if that's not currently set up to handle PDFs, I think we can skip it for now. 
But maybe you can create a tracking bug for this on this repository so that we can take a look at it in the future.
Please also link to this pull request from that new issue. Thanks!

jschfflr (approved):

PR #6942, opened 2021-03-01
fix(typescript): allow defaultViewport to be 'null'

christian-bromann:
Fixes: #6885

This aligns with how `defaultViewport` can be applied to a [`Page`](https://github.com/puppeteer/puppeteer/blob/main/src/common/Page.ts#L391) or [`Target`](https://github.com/puppeteer/puppeteer/blob/main/src/common/Target.ts#L69).

PR #6992, opened 2021-03-15
fix: defaultViewport can be null in BrowserConnectOptions

zoltan-boros:
### Steps to reproduce

**Description:**

As per API `defaultViewport` field in launch options ( `puppeteer.lauch( { defaultViewport: ... } )` [link](https://pptr.dev/#?product=Puppeteer&version=v8.0.0&show=api-puppeteerlaunchoptions) ) can have the three below values:
* `Viewport` object - The specified viewport will be applied.
* `undefined` (i.e. `defaultViewport` field is absent) - Default viewport will be applied.
* `null` - No viewport will be applied. Useful in headful mode, because page content will accomodate to browser extent.

* Puppeteer version: 8.0.0
* Platform / OS version: N/A
* URLs (if applicable): https://github.com/puppeteer/puppeteer/blob/0b5969dea9048dc34a4d2e109e96c42e8928182c/src/common/BrowserConnector.ts#L40
* Node.js version: N/A

1. Create a TypeScript project that launches a Chromium instance using `puppeteer.launch(options)` function.
2. Among options specify `defaultViewport` field as `null`.

```javascript
puppeteer.lauch({
  headless: false,
  defaultViewport: null,
});
```

Code is accepted as valid, no type error raises.

**What happens instead?**

Type error raises, which tells that `null` is not valid value for `defaultViewport`.

### DIFF
--- a/src/common/BrowserConnector.ts
+++ b/src/common/BrowserConnector.ts
@@ -37,7 +37,7 @@ export interface BrowserConnectOptions {
   /**
    * Sets the viewport for each page.
    */
-  defaultViewport?: Viewport;
+  defaultViewport?: Viewport | null;
   /**
    * Slows down Puppeteer operations by the specified amount of milliseconds to
    * aid debugging.


zoltan-boros:
@googlebot I signed it!

sadym-chromium:
Could you please add more details? Ideally - open an issue justifying the need of the change.
What is the use-case when the `defaultViewport` is null?

zoltan-boros:
@sadym-chromium, thank you for the advice.
I added details.

L0tso:
@sadym-chromium Another open pull request with linked issues https://github.com/puppeteer/puppeteer/pull/6942
Also, according to spec `defaultViewport <?Object> Sets a consistent viewport for each page. Defaults to an 800x600 viewport. null disables the default viewport. null disables the default viewport. `

sadym-chromium (approved):

zoltan-boros:
@sadym-chromium thank you for your review and approval.
I realized that #6942 is the proper fix of the same issue.
I would close this issue in favour of #6942. Could you review that one, please?

zoltan-boros:
Closed in favour of #6942

Issue #6976, opened 2021-03-11
Firefox - Cookie unit tests fail for asserts on non-standard properties (eg sameParty) 


juliandescottes:
Puppeteer recently started to assert new cookie properties in unit tests:
- `sourcePort` & `sourceScheme` in https://github.com/puppeteer/puppeteer/pull/6927
- `sameParty` in https://github.com/puppeteer/puppeteer/pull/6797

They seem non-standard. At least for `sameParty` I could find a reference clearly stating that Firefox would not support it: https://www.chromestatus.com/feature/5280634094223360 
For sourcePort & sourceScheme I didn't find any resource.

Firefox currently has no way of supporting such properties, so all the tests asserting those values will fail on Firefox when we update Puppeteer:
- https://github.com/puppeteer/puppeteer/blob/35e66c43040a41d7e47578044688f368f366c5bb/test/cookies.spec.ts#L34
- https://github.com/puppeteer/puppeteer/blob/35e66c43040a41d7e47578044688f368f366c5bb/test/cookies.spec.ts#L90
- https://github.com/puppeteer/puppeteer/blob/35e66c43040a41d7e47578044688f368f366c5bb/test/defaultbrowsercontext.spec.ts#L27
(only mentioning the tests which used to pass before asserting the new properties)

I realize that in puppeteer those tests were already running as `itFailsFirefox` before, so the regression was not detected. But they were actually passing and we are running them successfully in the Firefox CI today

Firefox will probably never be able to pass the assert for those properties, but it would be great to be able to still run the tests, maybe we could check the non-standard properties only when running in Chrome?

(corresponding Bugzilla reference: https://bugzilla.mozilla.org/show_bug.cgi?id=1694506)

whimboo:
@jschfflr and @mathiasbynens could we get those extra checks put behind an if condition for Chrome?

juliandescottes:
FWIW, I tried a simple modification to skip `sourcePort`, `sourceScheme` and `sameParty` from the asserts on cookies when `isChrome` is false and this fixes the tests for Firefox, so it's really just about those three properties ([changeset](https://hg.mozilla.org/try/rev/e8da1a77883362c631bc2c3ab622502290d62dbd), [ci job](https://treeherder.mozilla.org/jobs?repo=try&revision=427da6f8fe1cf60e21598dab9026341c54fc040e)).

Mostly extracted the logic to a small helper to assert cookies.
```javascript
[...]

PR #6994, opened 2021-03-15
chore: skip non-standard cookie asserts on Firefox

juliandescottes:
Fix for https://github.com/puppeteer/puppeteer/issues/6976

This change adds a dedicated assert helper for cookies which will only check the non-standard properties `sameParty`, `sourcePort` and `sourceScheme` when running against chrome.

All cookie asserts are replaced to use this helper.

3 tests marked as failing with Firefox are enabled: 
- Page.cookies should get a cookie
- Page.cookies should get multiple cookies
- DefaultBrowserContext page.cookies() should work

### DIFF
--- a/test/mocha-utils.ts
+++ b/test/mocha-utils.ts
@@ -28,6 +28,7 @@ import { Page } from '../lib/cjs/puppeteer/common/Page.js';
 import { PuppeteerNode } from '../lib/cjs/puppeteer/node/Puppeteer.js';
 import utils from './utils.js';
 import rimraf from 'rimraf';
+import expect from 'expect';
 
 import { trackCoverage } from './coverage-utils.js';
 
@@ -276,4 +277,26 @@ export const mochaHooks = {
   afterEach: () => {
     sinon.restore();
   },
+};
+
+export const expectCookieEquals = (cookies, expectedCookies) => {
+  const { isChrome } = getTestState();
+  if (!isChrome) {
+    // Only keep standard properties when testing on a browser other than Chrome.
+    expectedCookies = expectedCookies.map((cookie) => {
+      return {
+        domain: cookie.domain,
+        expires: cookie.expires,
+        httpOnly: cookie.httpOnly,
+        name: cookie.name,
+        path: cookie.path,
+        secure: cookie.secure,
+        session: cookie.session,
+        size: cookie.size,
+        value: cookie.value,
+      };
+    });
+  }
+
+  expect(cookies).toEqual(expectedCookies);
 };
--- a/test/cookies.spec.ts
+++ b/test/cookies.spec.ts
@@ -15,6 +15,7 @@
  */
 import expect from 'expect';
 import {
+  expectCookieEquals,
   getTestState,
   setupTestBrowserHooks,
   setupTestPageAndContextHooks,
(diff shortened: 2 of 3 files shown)


test/mocha-utils.ts
```diff
@@ -277,3 +278,18 @@ export const mochaHooks = {
     sinon.restore();
   },
 };
+
+export const expectCookieEquals = (cookies, expectedCookies) => {
+  const { isChrome } = getTestState();
+  if (!isChrome) {
+    for (const expectedCookie of expectedCookies) {
```
jschfflr: What do you think about changing this to check to only checking the standard properties instead of removing the non standard ones?

jschfflr (changes requested):

juliandescottes:
@googlebot I signed it!

test/mocha-utils.ts
juliandescottes: Thanks for the review! Tried to keep it simple and just copied over all the standard/shared values to a new object. Let me know if you had something else in mind.

juliandescottes:
A test failed at https://github.com/puppeteer/puppeteer/pull/6994/checks?check_run_id=2114890609
Looks like the failing tests are:
```
  Workers
    1) "before all" hook for "Page.workers"
    2) "after all" hook for "should report errors"
```

Any chance this could be an intermittent failure or is it really something my patch broke?

whimboo:
I triggered a re-run of all the jobs. Lets see...

whimboo:
All the checks passed this time.

jschfflr:
Yes, this is what I had in mind. Now, you should also be able to remove the `isChrome` branch.

juliandescottes:
> Yes, this is what I had in mind. Now, you should also be able to remove the `isChrome` branch.

Sorry, I don't have a `isChrome` branch in the current patch. Are you referring to the change in test/mocha-utils.ts or to something else?

test/mocha-utils.ts
```diff
@@ -277,3 +278,25 @@ export const mochaHooks = {
     sinon.restore();
   },
 };
+
+export const expectCookieEquals = (cookies, expectedCookies) => {
+  const { isChrome } = getTestState();
+  if (!isChrome) {
```
jschfflr, line 284: I was talking about this one. 
If I understand the problem space correctly, we don't need it anymore.

juliandescottes: Oh you mean you don't want to assert the properties even in chrome? That's fine I just assumed we preferred to keep checking those properties on the browsers which supported them. Will update!

juliandescottes: To be clear, I need to reverse the logic in my current patch. 

If we want to assert only standard properties on all browsers, then I'll have to remove `sameParty`, `sourcePort` and `sourceScheme` from all the tests. And in the helper, instead of filtering out properties from `expectedCookies`, I will extract them from `cookies`. Because Chrome will keep providing cookie objects with extra information.

jschfflr (approved):

juliandescottes:
Thanks for merging @jschfflr ! 
I was ok with updating the patch, and would have done so later today. I hope you're fine with the version you merged. If not, I'm really happy to provide another patch.

jschfflr:
Sorry for the late reply - I was OOO. No, I'm more than happy with the current version. Thanks!

PR #6995, opened 2021-03-15
chore: improve "Page.close should run beforeunload if asked for" unittest for Firefox

whimboo:
@jackfranklin, @mathiasbynens, or @jschfflr could we please get this reviewed soon? It's failing in our own CI now due to a change in the message string.

### DIFF
--- a/test/page.spec.ts
+++ b/test/page.spec.ts
@@ -67,10 +67,7 @@ describe('Page', function () {
       expect(dialog.type()).toBe('beforeunload');
       expect(dialog.defaultValue()).toBe('');
       if (isChrome) expect(dialog.message()).toBe('');
-      else
-        expect(dialog.message()).toBe(
-          'This page is asking you to confirm that you want to leave - data you have entered may not be saved.'
-        );
+      else expect(dialog.message()).toBeTruthy();
       await dialog.accept();
       await pageClosingPromise;
     });


test/page.spec.ts
```diff
@@ -67,10 +67,7 @@ describe('Page', function () {
       expect(dialog.type()).toBe('beforeunload');
       expect(dialog.defaultValue()).toBe('');
       if (isChrome) expect(dialog.message()).toBe('');
-      else
-        expect(dialog.message()).toBe(
-          'This page is asking you to confirm that you want to leave - data you have entered may not be saved.'
-        );
+      else expect(dialog.message()).toBeTruthy();
```
whimboo, line 70: Not sure if this is the right check. Please let me know if it's not and how to best test for any string.

whimboo:
Also I wonder why CDP returns an empty string for the dialog's message in Chrome.

test/page.spec.ts
jschfflr: This looks good to me!

jschfflr (approved):

jschfflr:
Not sure what you mean by:
> Also I wonder why CDP returns an empty string for the dialog's message in Chrome.

Is there a difference between Chrome and Chromium?

whimboo:
> > Also I wonder why CDP returns an empty string for the dialog's message in Chrome.
> 
> Is there a difference between Chrome and Chromium?

No, but see that same test but for Chrome:

> if (isChrome) expect(dialog.message()).toBe('');

Why isn't the message from the `beforeunload` prompt returned, or better why it is empty?

whimboo:
@jschfflr specifically this is related to https://chromedevtools.github.io/devtools-protocol/tot/Page/#event-javascriptDialogOpening:

> Message that will be displayed by the dialog.

And when I try Chrome with a `beforeunload` handler set outside of Puppeteer there is clearly a message displayed.

Issue #5176, opened 2019-11-17
Discussion: Enable request interception and caching at the same time


liweixi100:
Caching is intentionally disabled with page.setRequestInterception(true):
https://github.com/GoogleChrome/puppeteer/blob/d17708ba1fc8411941bfd07f633aaf2fb5cb671e/lib/NetworkManager.js#L154

As a result, all requests sent by Puppeteer carry the headers `Cache-Control: no-cache; Pragma: no-cache`, which disallows any external proxies to serve cached responses.

But our application leverages a caching proxy to save bandwidth and improve performance. It would be very important for us to have the caching enabled.

The only I way I found that would make request interception and caching work at the same time is to hack NetworkManager.js and remove `|| this._protocolRequestInterceptionEnabled ` at L154. (I think that was also the old behavior until the change introduced by #1154). 

My questions are:
1. Is there a better way to solve this dilemma?
2. Exactly which parts of the interception feature would be broken if we were to do the hack? If it were only a few edge cases, we may want to choose to live with them. 

@aslushnikov

kblok:
Duplicated of #4620

liweixi100:
I did tests with or without this hack for over 600 URLs and found something very shocking: 
with `|| this._protocolRequestInterceptionEnabled` removed, it ended up with a timeout in 70% of the cases while waiting for the `load` event. 

So to answer partially of question 2 --- the hack simply will not work in the majority of the cases. 

Is there any way to make request interception and caching work at same time? 

Header overrides obviously did not work as many (including myself) have tried: #3403 #4055

28141932:
我也需要

Androbin:
@liweixi100 Would you mind sharing the details of your tests? Ideally, we would have a unit test covering this.

Androbin:
The issue described by @liweixi100 is described in #7038 and solved by #7060.

liweixi100:
This is a great enhancement. Thank you @Androbin @starrify

PR #6996, opened 2021-03-16
Request interception and caching compatibility

### LINKED ISSUE #2905: Request interception and caching compatibility

Androbin:
See #2905

This change is fully backwards compatible as it defaults to the current behavior.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -166,7 +166,7 @@
   * [page.setGeolocation(options)](#pagesetgeolocationoptions)
   * [page.setJavaScriptEnabled(enabled)](#pagesetjavascriptenabledenabled)
   * [page.setOfflineMode(enabled)](#pagesetofflinemodeenabled)
-  * [page.setRequestInterception(value)](#pagesetrequestinterceptionvalue)
+  * [page.setRequestInterception(value[, cacheSafe])](#pagesetrequestinterceptionvalue-cachesafe)
   * [page.setUserAgent(userAgent)](#pagesetuseragentuseragent)
   * [page.setViewport(viewport)](#pagesetviewportviewport)
   * [page.tap(selector)](#pagetapselector)
@@ -2039,8 +2039,9 @@ await page.setGeolocation({latitude: 59.95, longitude: 30.31667});
 - `enabled` <[boolean]> When `true`, enables offline mode for the page.
 - returns: <[Promise]>
 
-#### page.setRequestInterception(value)
+#### page.setRequestInterception(value[, cacheSafe])
 - `value` <[boolean]> Whether to enable request interception.
+- `cacheSafe` <[boolean]> Whether to trust browser caching. If set to false, enabling request interception disables page caching. Defaults to false.
 - returns: <[Promise]>
 
 Activating request interception enables `request.abort`, `request.continue` and
--- a/src/common/NetworkManager.ts
+++ b/src/common/NetworkManager.ts
@@ -190,7 +190,10 @@ export class NetworkManager extends EventEmitter {
     await this._updateProtocolCacheDisabled();
   }
 
-  async setRequestInterception(value: boolean, cacheSafe: boolean = false): Promise<void> {
+  async setRequestInterception(
+    value: boolean,
+    cacheSafe = false
+  ): Promise<void> {
     this._userRequestInterceptionEnabled = value;
     this._userRequestInterceptionCacheSafe = cacheSafe;
     await this._updateProtocolRequestInterception();
@@ -219,7 +222,9 @@ export class NetworkManager extends EventEmitter {
   async _updateProtocolCacheDisabled(): Promise<void> {
     await this._client.send('Network.setCacheDisabled', {
       cacheDisabled:
-        this._userCacheDisabled || (this._userRequestInterceptionEnabled && !this._userRequestInterceptionCacheSafe),
+        this._userCacheDisabled ||
+        (this._userRequestInterceptionEnabled &&
+          !this._userRequestInterceptionCacheSafe),
     });
   }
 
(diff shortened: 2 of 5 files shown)


Androbin:
This is not a perfect solution, as outlined in https://github.com/puppeteer/puppeteer/issues/2905#issuecomment-406128495. It should cover most people's use cases though, while we are waiting for Chromium changes to land.

Androbin:
@OrKoN could you PTAL?

OrKoN:
Jan, could you PTAL since you looked into the request interception recently?

jschfflr (approved):
LGTM, thanks!

jschfflr:
One thing though: Could you add a unit test to cover this use case?

jschfflr (changes requested):
Ah damn, I'm confused by this UI.
Could you add a test that covers this new behaviour?

Androbin:
When we call `Network.setCacheDisabled({cacheDisabled: false})`, certain events are not fired, including `Page.Events.Response`. To get around this, the test is listening to the newly added `Page.Events.RequestServedFromCache` instead.

jschfflr (approved):

jribbens:
Can I suggest that the documentation update should give the merest clue as to why you might or might not want caching to be disabled? If I'm understanding it correctly, the only problem is that if caching is not disabled, then requests served from the cache will not be intercepted - which may well not be a problem at all in many applications.

starrify:
I'm under the impression that caching is purposely disabled when configuring request interception because otherwise unexpected issues would happen. This is also stated by the original author @aslushnikov of the relevant code in #1154 and #4260.

It is relatively easy to just configure request interception and having caching enabled at the same time. I have tried that some time ago and have got the similar results (rendering some pages never finish due to lack of the `load` event in page) as described in this comment: https://github.com/puppeteer/puppeteer/issues/5176#issuecomment-554815586 . (sorry I'm unable to share further details on my previous tests)

Unfortunately, this PR does not seem to resolve the earlier issue as it merely provides the means for users to keep caching enabled while doing request interception.

I have performed further tests, and it seems that Puppeteer at 8695759 (revision that has this PR merged) would fail to render many webpages, one example of which is https://github.com/, while having `page.setRequestInterception(true, true);`. I have created issue #7038 to share further details.

I'm afraid that this PR may need to be excluded from the next Puppeteer release unless properly fixed, as the current version merely reintroduced the issue that was previously got rid of in #1154 and #4260.

Androbin:
@starrify I am fully aware that `page.setRequestInterception(true, true);` might misbehave for some use cases.
However there are people with use cases where this doesn't break and they really need this flag.
It's opt-in, so there is no harm in keeping the flag. We can however document it as experimental.

starrify:
Right, I agree that people with specific use cases may be interested in this, though I much doubt the coverage as there are many real-world web pages that would break on this, according to my previous tests and what's mentioned here: https://github.com/puppeteer/puppeteer/issues/5176#issuecomment-554815586

<details>
<summary>
@Androbin: Could you help quickly verify whether it is true that Puppeteer at revision 8695759 (with this PR merged) would fail to render https://github.com/ ? (click me for a sample test script)
</summary>

```javascript
(async () => {
    // Needed only when inside a Docker container
    args: ['--no-sandbox'],
  });
  const enableCaching = process.env.TEST_ENABLE_CACHING == 'true';
  await page.setRequestInterception(true, enableCaching);
    request.continue();
  });
  const url = process.env.TEST_URL || 'https://github.com/';
  try {
    await page.goto(url);
  } catch (err) {
    console.error(err);
  }
  await browser.close();
})();
```
</details>

Sorry but I'm against the very idea of have it documented _as experimental_, as I would like to see from the documents _clear warnings that this feature is known to break the rendering of many real world web pages_ (if my other issue report doesn't get proved wrong).

Androbin:
```js
await page.setRequestInterception(true, false);
page.on('request', (request) => request.continue());
await page.goto('https://github.com/', { waitUntil: 'load' });
```
PASSED

```js
await page.setRequestInterception(true, true);
page.on('request', (request) => request.continue());
await page.goto('https://github.com/', { waitUntil: 'domcontentloaded' });
```
PASSED

```js
await page.setRequestInterception(true, true);
page.on('request', (request) => request.continue());
await page.goto('https://github.com/', { waitUntil: 'load' });
```
FAILED

```js
await page.setRequestInterception(true, true);
  // alternative: request.resourceType() == "font"
  if (request.url().endsWith(".woff")) {
    request.abort();
  } else {
    request.continue();
  }
});
await page.goto('https://github.com/', { waitUntil: 'load' });
```
PASSED

Androbin:
I would like to note that blocking requests for fonts and images is a common use case of the people who requested this flag.

Androbin:
@starrify If you provide me with a list of sites that have this issue, I could compile a list of failing resources, we could identify what's common between them, and fix the issue.

Androbin:
I have done some debugging and it looks like the events generated by Chromium are in a different order for fonts than for other resources like stylesheets.

Stylesheet request:
- `onRequestWillBeSent`
- `onRequestPaused` (calls `request.continue()`)
- `onRequest`
- `onResponseReceived`
- `onLoadingFinished`

Font request (cache disabled):
- `onRequestPaused`
- `onRequestWillBeSent` (calls `request.continue()`)
- `onRequest`
- `onResponseReceived`
- `onLoadingFinished`

Font request (cache enabled):
- `onRequestPaused`
- `onRequestWillBeSent` (calls `request.continue()`)
- `onRequest`
- `onRequestPaused` (does not call `request.continue()`)
- request pending indefinitely

Font request (cache enabled + fix):
- `onRequestPaused`
- `onRequestWillBeSent` (calls `request.continue()`)
- `onRequest`
- `onRequestPaused` (calls `request.continue()`)
- `onRequest`
- `onResponseReceived`
- `onLoadingFinished`

jschfflr:
Could you also check a scenario with redirects and how they change the order of events?

PR #7015, opened 2021-03-22
Update troubleshooting.md

conradkirschner:
On Linux npm complains with an error code that no package.json was found, then it fails the whole command (&&)


conradkirschner:
@googlebot I signed it!

jackfranklin (approved):
LGTM, thanks!

jschfflr (approved):
The changes look good to me. But we should consider to move things into a different folder than `/`.

PR #7022, opened 2021-03-24
Change

akshat01112001:


.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
jackfranklin: Do you know what caused this change?

package.json
```diff
@@ -109,6 +110,7 @@
     "standard-version": "^9.0.0",
     "text-diff": "^1.0.1",
     "ts-node": "^9.0.0",
+    "typedoc": "^0.20.32",
```
jackfranklin: typedoc is different to TSDoc (confusingly!) so I don't think we want this. Why did you add it?

src/common/Browser.ts
```diff
@@ -412,7 +412,7 @@ export class Browser extends EventEmitter {
       url: 'about:blank',
       browserContextId: contextId || undefined,
     });
-    const target = await this._targets.get(targetId);
+    const target = this._targets.get(targetId);
```
jackfranklin, line 418: Nice spot!

src/common/Connection.ts
```diff
@@ -67,8 +67,8 @@ export class Connection extends EventEmitter {
   }
 
   /**
-   * @param {string} sessionId
-   * @returns {?CDPSession}
+   * @param sessionId - Gets Session ID
+   * @returns - Gets CDPSession
```
jackfranklin: You don't need a dash after `@returns` - see https://tsdoc.org/pages/tags/returns/

So this should be

```
@returns the current CDP session if it exists

File: src/common/Connection.ts [hunk 5]
```diff
@@ -167,8 +167,8 @@ export class Connection extends EventEmitter {
   }
 
   /**
-   * @param {Protocol.Target.TargetInfo} targetInfo
-   * @returns {!Promise<!CDPSession>}
+   * @param targetInfo - Gets targetInfo
+   * @returns sessionID
```
jackfranklin, line 179: try and write these like a sentence, so this might be:

```
@returns the CDP session that is created
```

src/common/Connection.ts
```diff
@@ -167,8 +167,8 @@ export class Connection extends EventEmitter {
   }
 
   /**
-   * @param {Protocol.Target.TargetInfo} targetInfo
-   * @returns {!Promise<!CDPSession>}
+   * @param targetInfo - Gets targetInfo
```
jackfranklin: targetInfo here is a param the user provides, so "Gets targetInfo" doesn't quite fit. If you're not sure what to document for this param, just leave it as:

```
@param targetInfo
```

src/common/EventEmitter.ts
```diff
@@ -33,7 +33,7 @@ export interface CommonEventEmitter {
  * {@link EventEmitter.off | off} to bind
  * and unbind to event listeners.
  *
- * @public
+ * @internal
```
jackfranklin: why did you change this? We do want to document this class so I don't believe this should be internal.

src/common/ExecutionContext.ts
```diff
@@ -33,7 +33,7 @@ const SOURCE_URL_REGEX = /^[\040\t]*\/\/[@#] sourceURL=\s*(\S*?)\s*$/m;
  *   {@link https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe |
  *   frame } has "default" execution context that is always created after frame is
  *   attached to DOM. This context is returned by the
- *   {@link frame.executionContext()} method.
+ *   {@link frame.executionContext()| method} method.
```
jackfranklin: this will case the rendered text to read `method method` - the bit after the `|` is the text used. Why did you make this change?

src/common/FileChooser.ts
```diff
@@ -33,7 +33,8 @@ import { assert } from './assert.js';
  * ```
  * **NOTE** In browsers, only one file chooser can be opened at a time.
  * All file choosers must be accepted or canceled. Not doing so will prevent
- * subsequent file choosers from appearing.
+ * subsequent file choosers from app/earing.
+ * @internal
```
jackfranklin: I think this should actually be `@public` as we want it documented to the user.

src/common/Page.ts
```diff
@@ -283,7 +283,7 @@ export const enum PageEmittedEvents {
    * Emitted when a page issues a request and contains a {@link HTTPRequest}.
    *
    * @remarks
-   * The object is readonly. See {@Page.setRequestInterception} for intercepting
+   * The object is readonly. See /{@Page setRequestInterception \} for intercepting
```
jackfranklin: why did you add the slashes here?

jackfranklin (commented):
Thanks for your PR! I've left some comments with some questions, if you wouldn't mind taking another look 👍

akshat01112001:
I have updated the changes. Have a look.

Cheers,
Akshat

src/common/Page.ts
```diff
@@ -283,7 +283,7 @@ export const enum PageEmittedEvents {
    * Emitted when a page issues a request and contains a {@link HTTPRequest}.
    *
    * @remarks
-   * The object is readonly. See {@Page.setRequestInterception} for intercepting
+   * The object is readonly. See {@Page setRequestInterception} for intercepting
```
jackfranklin: We still need the dot here - this should be `@Page.setRequestInterception` so TSDoc links to the right method

src/common/ExecutionContext.ts
jackfranklin: @akshat01112001 please could you undo this change?

akshat01112001: I have undone this change

jackfranklin (approved):

PR #7037, opened 2021-03-30
fix(aria): fix parsing of ARIA selectors

OrKoN:
- adds a g flag so that the replacer is called for every match.
- fixes the reported error message and adds a test.


mathiasbynens (approved):
LGTM, but the linter seems unhappy

PR #7052, opened 2021-04-04
Make respond Header names case sensitive

lolyinseo:
According RFC 2616 Header names are not case sensitive but produces an error on some sites


jschfflr:
This change looks good to me from a technical point of view. But I'd consider it a breaking change for most users. 
@mathiasbynens Let me know if you think we should land this as is or maybe include both original and lowercase versions for a transition period.

mathiasbynens:
@lolyinseo Could you please elaborate on what breakage the current behavior is causing exactly? Header names are case-insensitive and IIRC we aim to lowercase them before using them in Puppeteer, matching e.g. https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagesetextrahttpheadersheaders.

jschfflr:
@mathiasbynens I think this is about transforming the http headers in responses to lowercase too.

jschfflr:
> @mathiasbynens I think this is about transforming the http headers in responses to lowercase too.

Ah wait, I got it the other way around. This removes lowercasing the headers. That's probably not what we want. 
@lolyinseo Could you provide an example where the existing behaviour does not work as expected?

lolyinseo:
> @lolyinseo Could you please elaborate on what breakage the current behavior is causing exactly? Header names are case-insensitive and IIRC we aim to lowercase them before using them in Puppeteer, matching e.g. https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagesetextrahttpheadersheaders.

Yes you are right. Header names are case-insensitive and IIRC... But I have already encountered a problem several times when on the server side (apparently) the check is performed on the exact occurrence of the header. + I don't understand why transforming the http headers to lowercase too, it doesn't solve any problems, but it potentially creates them. It doesn't make sense.

jschfflr:
As this is only touching the client side, I feel like we should stick to the spec but I'd be happy to accept a pull request that doesn't change the casing for requests that go out.

OrKoN:
So to summarise the discussion: Puppeteer works according to the spec and some backends out there are not following the spec? And it sounds like we don't want to change the case to avoid breaking changes? I am closing this PR in this case. Feel free to re-open if I misunderstood the discussion.

PR #7058, opened 2021-04-05
docs: warning messages

AymanAzzam:
I solved the (ae-missing-release-tag) warning by adding public tag.  there are a lot of warning:
1. (tsdoc-param-tag-missing-hyphen)
2. (ae-forgotten-export)
3. (tsdoc-escape-right-brace)
4. (tsdoc-escape-greater-than)
5. (tsdoc-characters-after-inline-tag)
5. (tsdoc-reference-missing-dot)
6. (tsdoc-malformed-inline-tag)
7. (tsdoc-param-tag-with-invalid-type)

I tried in number 2 but it seems like nothing is forgotten ! should I silence this warning ?
what about the other warnings, should i silence them too or try to solve them first?
@jackfranklin

### DIFF
--- a/src/common/Browser.ts
+++ b/src/common/Browser.ts
@@ -550,7 +550,9 @@ export class Browser extends EventEmitter {
     return this._connection.send('Browser.getVersion');
   }
 }
-
+/**
+ * @public
+ */
 export const enum BrowserContextEmittedEvents {
   /**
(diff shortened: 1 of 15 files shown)


AymanAzzam:
> @googlebot I signed it!

@googlebot I signed it!

PR #7060, opened 2021-04-06
fix(requestinterception): fix font loading issue

### LINKED ISSUE #7038: Request interception with caching (rev 8695759) fails when loading stylesheet-initiated fonts (?)
## 0. Overview

**TL;DR:**
- The recent fix of request interception and caching in revision 8695759 (#6996) does not work properly in some conditions.
- The error may be stably reproduced.
- While I'm not 100% sure before understanding the relevant Chromium implement details, this seems to be caused by CSS-initiated font resources

## 1. Test Configuration

My tests are conducted with:
- GNU/Linux Debian 10 (buster)
- `Node.js 14.16.0` and `npm 7.7.6`
- Puppeteer 8695759 (also its parent bf60a30 for comparison)
- Chromium r856583 (the bundled version from Puppeteer)

Here is a `Dockerfile` that may duplicate the exact test environment:
```Dockerfile
FROM node:14.16.0-buster
# Just for the dependencies.
RUN apt-get update && apt-get install -y chromium
# Puppeteer installation from git requires npm>=7
# See https://github.com/puppeteer/puppeteer/issues/7034
RUN npm install -g npm@7.7.6
WORKDIR /app
ARG PPTR_RVSN=8695759
RUN npm install git+ssh://git@github.com/puppeteer/puppeteer.git#${PPTR_RVSN}
ADD test.js .
```

Here is the sample JavaScript code I used to trigger the issue:
```javascript
(async () => {
    // Needed only when inside a Docker container
    args: ['--no-sandbox'],
  });
  const enableCaching = process.env.TEST_ENABLE_CACHING == 'true';
  await page.setRequestInterception(true, enableCaching);
    request.continue();
  });
  const url = process.env.TEST_URL || 'https://github.com/';
  try {
    await page.goto(url);
  } catch (err) {
    console.error(err);
  }
  await browser.close();
})();
```
This script accesses https://github.com/ by default, while in practice I observed the same issue from several other frequently visited websites.

## 2. Tests and Results

From the above configurations one may set up the test environment and launch the quick test.
Here is an example of triggering the issue:
```
$ docker build -t tmptest-pptr -f Dockerfile --build-arg PPTR_RVSN=8695759 . &> /dev/null
$ docker run --entrypoint=/bin/bash --env TEST_ENABLE_CACHING=true tmptest-pptr -c "node test.js"
TimeoutError: Navigation timeout of 30000 ms exceeded
    at /app/node_modules/puppeteer/lib/cjs/puppeteer/common/LifecycleWatcher.js:106:111
```

Here are the respective results per test:

| Puppeteer ↓ \ caching → | enabled | disabled |
|-------------------------|---------|----------|
| 8695759                 | error   | no error |
| bf60a30 (previous one)  | N/A     | no error |

Therefore the implement of 8695759 seems not fully functional, at least not for rendering https://github.com/ as tested.

## 3. Further Analysis

Thanks to Puppeteer's debugging ability, I tried to dump the CDP communication of the failed test via the below command:
```
$ docker run --entrypoint=/bin/bash --env TEST_ENABLE_CACHING=true --env DEBUG="puppeteer:*" tmptest-pptr -c "node test.js" &> 8695759_caching_on.log
```

In case that may save you some time, I've uploaded the recorded `8695759_caching_on.log` file here: https://gist.github.com/starrify/e1f5ad68c358b932890933be3fb2b736

By comparing the URLs in the log file, it is observed that there are four `.woff` resources for fonts, all initiated by the stylesheets, that had the `Network.requestWillBeSent` event but not a `Network.responseReceived`:
```
$ tmp_filter() { grep -Po "(?<=protocol:RECV ◀ ).*" | jq 'select(.method=="'$1'")|.params|(.response // .request)|.url' | sort; }
$ diff <(cat 8695759_caching_on.log | tmp_filter Network.requestWillBeSent) <(cat 8695759_caching_on.log | tmp_filter Network.responseReceived)
37,40d36
< "https://github.githubassets.com/static/fonts/alliance/Alliance-No-1-Bold.woff"
< "https://github.githubassets.com/static/fonts/alliance/Alliance-No-1-ExtraBold.woff"
< "https://github.githubassets.com/static/fonts/alliance/Alliance-No-1-Regular.woff"
< "https://github.githubassets.com/static/fonts/alliance/Alliance-No-1-SemiBold.woff"
```

These four font resources are apparently not cached since the Docker container has a cold start. Also that's confirmed from the extracted CDP log file (e.g. no `Network.requestServedFromCache` event observed).

Taking one resource `Alliance-No-1-SemiBold.woff` as an example for further inspection, I've got these from the log file:
```
$ cat 8695759_caching_on.log | grep "Alliance-No-1-SemiBold.woff" | grep -Po "(?<=protocol:RECV ◀ ).*" | jq '.params|(.requestId//.networkId)' -r
interception-job-28.0
51.164
interception-job-38.0

$ cat 8695759_caching_on.log | grep -P '\b(51\.164|interception-job-28\.0|interception-job-38\.0)\b'
2021-03-30T22:25:48.770Z puppeteer:protocol:RECV ◀ {"method":"Fetch.requestPaused","params":{"requestId":"interception-job-28.0","request":{"url":"https://github.githubassets.com/static/fonts/alliance/Alliance-No-1-SemiBold.woff","method":"GET","headers":{"sec-ch-ua":"","Origin":"https://github.com","sec-ch-ua-mobile":"?0","User-Agent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/90.0.4427.0 Safari/537.36","Accept":"*/*","Referer":"https://github.githubassets.com/assets/site-7c8868efa7b2b508d6f89c0e94f94864.css"},"initialPriority":"VeryHigh","referrerPolicy":"strict-origin-when-cross-origin"},"frameId":"0979EF16C3B9EFE96B9BD45E76E24204","resourceType":"Font","networkId":"51.164"},"sessionId":"E5290B3CCB2B7FB6607E6186D7644353"}
2021-03-30T22:25:48.935Z puppeteer:protocol:RECV ◀ {"method":"Network.requestWillBeSent","params":{"requestId":"51.164","loaderId":"5423DC2D96468E21561099D7F1FA7D90","documentURL":"https://github.com/","request":{"url":"https://github.githubassets.com/static/fonts/alliance/Alliance-No-1-SemiBold.woff","method":"GET","headers":{"sec-ch-ua":"","Referer":"https://github.githubassets.com/assets/site-7c8868efa7b2b508d6f89c0e94f94864.css","Origin":"https://github.com","sec-ch-ua-mobile":"?0","User-Agent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/90.0.4427.0 Safari/537.36"},"mixedContentType":"none","initialPriority":"VeryHigh","referrerPolicy":"strict-origin-when-cross-origin"},"timestamp":781832.362451,"wallTime":1617143148.769689,"initiator":{"type":"parser","url":"https://github.githubassets.com/assets/site-7c8868efa7b2b508d6f89c0e94f94864.css"},"type":"Font","frameId":"0979EF16C3B9EFE96B9BD45E76E24204","hasUserGesture":false},"sessionId":"E5290B3CCB2B7FB6607E6186D7644353"}
2021-03-30T22:25:48.935Z puppeteer:protocol:SEND ► {"sessionId":"E5290B3CCB2B7FB6607E6186D7644353","method":"Fetch.continueRequest","params":{"requestId":"interception-job-28.0"},"id":46}
2021-03-30T22:25:48.966Z puppeteer:protocol:RECV ◀ {"method":"Fetch.requestPaused","params":{"requestId":"interception-job-38.0","request":{"url":"https://github.githubassets.com/static/fonts/alliance/Alliance-No-1-SemiBold.woff","method":"GET","headers":{"sec-ch-ua":"","Origin":"https://github.com","sec-ch-ua-mobile":"?0","User-Agent":"Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/90.0.4427.0 Safari/537.36","Accept":"*/*","Referer":"https://github.githubassets.com/assets/site-7c8868efa7b2b508d6f89c0e94f94864.css"},"initialPriority":"VeryHigh","referrerPolicy":"strict-origin-when-cross-origin"},"frameId":"0979EF16C3B9EFE96B9BD45E76E24204","resourceType":"Font","networkId":"51.164"},"sessionId":"E5290B3CCB2B7FB6607E6186D7644353"}
```

This suggests something interesting: for the same request (as identified by the `networkId` field), Chrome actually issued **twice** the `Fetch.requestPaused` event for some reason I do not yet know.

While Puppeteer only manages to resume the first pause, such requests are therefore blocked forever.

In addition, since such resources are initiated in the stylesheets, having the requests pending indefinitely would block the browser from issuing the `Page.loadEventFired` event, therefore Puppeteer would also keep blocking on the `page.goto` line.

Androbin:
Fixes #7038 @starrify

See https://github.com/puppeteer/puppeteer/pull/6996#issuecomment-811546501 and https://github.com/puppeteer/puppeteer/pull/6996#issuecomment-813797393 for context

### DIFF
--- a/src/common/NetworkManager.ts
+++ b/src/common/NetworkManager.ts
@@ -69,11 +69,17 @@ export class NetworkManager extends EventEmitter {
   _client: CDPSession;
   _ignoreHTTPSErrors: boolean;
   _frameManager: FrameManager;
-  _requestIdToRequest = new Map<string, HTTPRequest>();
+
   _requestIdToRequestWillBeSentEvent = new Map<
     string,
     Protocol.Network.RequestWillBeSentEvent
   >();
+  _requestIdToRequestPausedEvent = new Map<
+    string,
+    Protocol.Fetch.RequestPausedEvent
+  >();
+  _requestIdToRequest = new Map<string, HTTPRequest>();
+
   _extraHTTPHeaders: Record<string, string> = {};
   _credentials?: Credentials = null;
   _attemptedAuthentications = new Set<string>();
@@ -81,7 +87,6 @@ export class NetworkManager extends EventEmitter {
   _userRequestInterceptionCacheSafe = false;
   _protocolRequestInterceptionEnabled = false;
   _userCacheDisabled = false;
-  _requestIdToInterceptionId = new Map<string, string>();
   _emulatedNetworkConditions: InternalNetworkConditions = {
     offline: false,
     upload: -1,
@@ -222,12 +227,17 @@ export class NetworkManager extends EventEmitter {
     }
   }
 
+  _cacheDisabled(): boolean {
+    return (
+      this._userCacheDisabled ||
+      (this._userRequestInterceptionEnabled &&
+        !this._userRequestInterceptionCacheSafe)
+    );
+  }
+
   async _updateProtocolCacheDisabled(): Promise<void> {
     await this._client.send('Network.setCacheDisabled', {
-      cacheDisabled:
-        this._userCacheDisabled ||
-        (this._userRequestInterceptionEnabled &&
-          !this._userRequestInterceptionCacheSafe),
+      cacheDisabled: this._cacheDisabled(),
     });
   }
 
@@ -238,13 +248,18 @@ export class NetworkManager extends EventEmitter {
       !event.request.url.startsWith('data:')
     ) {
       const requestId = event.requestId;
-      const interceptionId = this._requestIdToInterceptionId.get(requestId);
-      if (interceptionId) {
+      const requestPausedEvent = this._requestIdToRequestPausedEvent.get(
+        requestId
+      );
+
+      if (requestPausedEvent) {
+        const interceptionId = requestPausedEvent.requestId;
         this._onRequest(event, interceptionId);
-        this._requestIdToInterceptionId.delete(requestId);
+        this._requestIdToRequestPausedEvent.delete(requestId);
       } else {
-        this._requestIdToRequestWillBeSentEvent.set(event.requestId, event);
+        this._requestIdToRequestWillBeSentEvent.set(requestId, event);
       }
+
       return;
     }
     this._onRequest(event, null);
@@ -288,14 +303,20 @@ export class NetworkManager extends EventEmitter {
 
     const requestId = event.networkId;
     const interceptionId = event.requestId;
-    if (requestId && this._requestIdToRequestWillBeSentEvent.has(requestId)) {
-      const requestWillBeSentEvent = this._requestIdToRequestWillBeSentEvent.get(
-        requestId
-      );
+
+    if (!requestId) {
+      return;
+    }
+
+    const requestWillBeSentEvent = this._requestIdToRequestWillBeSentEvent.get(
+      requestId
+    );
+
+    if (requestWillBeSentEvent) {
       this._onRequest(requestWillBeSentEvent, interceptionId);
       this._requestIdToRequestWillBeSentEvent.delete(requestId);
     } else {
-      this._requestIdToInterceptionId.set(requestId, interceptionId);
+      this._requestIdToRequestPausedEvent.set(requestId, event);
     }
   }
 
@@ -346,8 +367,7 @@ export class NetworkManager extends EventEmitter {
     response._resolveBody(
       new Error('Response body is unavailable for redirect responses')
     );
-    this._requestIdToRequest.delete(request._requestId);
-    this._attemptedAuthentications.delete(request._interceptionId);
+    this._forgetRequest(request);
     this.emit(NetworkManagerEmittedEvents.Response, response);
     this.emit(NetworkManagerEmittedEvents.RequestFinished, request);
   }
@@ -361,6 +381,14 @@ export class NetworkManager extends EventEmitter {
     this.emit(NetworkManagerEmittedEvents.Response, response);
   }
 
+  _forgetRequest(request: HTTPRequest): void {
+    const requestId = request._requestId;
+    const interceptionId = request._interceptionId;
+
+    this._requestIdToRequest.delete(requestId);
+    this._attemptedAuthentications.delete(interceptionId);
+  }
+
   _onLoadingFinished(event: Protocol.Network.LoadingFinishedEvent): void {
     const request = this._requestIdToRequest.get(event.requestId);
     // For certain requestIds we never receive requestWillBeSent event.
@@ -370,8 +398,7 @@ export class NetworkManager extends EventEmitter {
     // Under certain conditions we never get the Network.responseReceived
     // event from protocol. @see https://crbug.com/883475
     if (request.response()) request.response()._resolveBody(null);
-    this._requestIdToRequest.delete(request._requestId);
-    this._attemptedAuthentications.delete(request._interceptionId);
+    this._forgetRequest(request);
     this.emit(NetworkManagerEmittedEvents.RequestFinished, request);
   }
 
@@ -383,8 +410,7 @@ export class NetworkManager extends EventEmitter {
     request._failureText = event.errorText;
     const response = request.response();
     if (response) response._resolveBody(null);
-    this._requestIdToRequest.delete(request._requestId);
-    this._attemptedAuthentications.delete(request._interceptionId);
+    this._forgetRequest(request);
     this.emit(NetworkManagerEmittedEvents.RequestFailed, request);
   }
 }
--- a/src/common/NetworkManager.ts
+++ b/src/common/NetworkManager.ts
@@ -70,6 +70,25 @@ export class NetworkManager extends EventEmitter {
   _ignoreHTTPSErrors: boolean;
   _frameManager: FrameManager;
 
+  /*
+   * There are four possible orders of events:
+   *  A. `_onRequestWillBeSent`
+   *  B. `_onRequestWillBeSent`, `_onRequestPaused`
+   *  C. `_onRequestPaused`, `_onRequestWillBeSent`
+   *  D. `_onRequestPaused`, `_onRequestWillBeSent`, `_onRequestPaused`
+   *     (see crbug.com/1196004)
+   *
+   * For `_onRequest` we need the event from `_onRequestWillBeSent` and
+   * optionally the `interceptionId` from `_onRequestPaused`.
+   *
+   * If request interception is disabled, call `_onRequest` once per call to
+   * `_onRequestWillBeSent`.
+   * If request interception is enabled, call `_onRequest` once per call to
+   * `_onRequestPaused` (once per `interceptionId`).
+   *
+   * Events are stored to allow for subsequent events to call `_onRequest`.
+   * Note that redirect requests have the same `requestId` (!).
+   */
   _requestIdToRequestWillBeSentEvent = new Map<
     string,
     Protocol.Network.RequestWillBeSentEvent
@@ -252,12 +271,12 @@ export class NetworkManager extends EventEmitter {
         requestId
       );
 
+      this._requestIdToRequestWillBeSentEvent.set(requestId, event);
+
       if (requestPausedEvent) {
         const interceptionId = requestPausedEvent.requestId;
         this._onRequest(event, interceptionId);
         this._requestIdToRequestPausedEvent.delete(requestId);
-      } else {
-        this._requestIdToRequestWillBeSentEvent.set(requestId, event);
       }
 
       return;
@@ -308,9 +327,19 @@ export class NetworkManager extends EventEmitter {
       return;
     }
 
-    const requestWillBeSentEvent = this._requestIdToRequestWillBeSentEvent.get(
+    let requestWillBeSentEvent = this._requestIdToRequestWillBeSentEvent.get(
       requestId
     );
+
+    // redirect requests have the same `requestId`,
+    if (
+      requestWillBeSentEvent &&
+      (requestWillBeSentEvent.request.url !== event.request.url ||
+        requestWillBeSentEvent.request.method !== event.request.method)
+    ) {
+      this._requestIdToRequestWillBeSentEvent.delete(requestId);
+      requestWillBeSentEvent = null;
+    }
 
     if (requestWillBeSentEvent) {
       this._onRequest(requestWillBeSentEvent, interceptionId);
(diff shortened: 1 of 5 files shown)


Androbin:
CI fails at "should work with redirects for subresources" but cannot reproduce locally, might be flaky

starrify:
Thanks @Androbin for the updates!

A quick question please: are you sure that the test case you added does trigger the very issue as earlier reported issue in #7038 ? Asking because I recall having the highly similar tests as well and the very test case just won't cause anything wrong.

It shall be easy to confirm, though: Just keep the test case and drop the newly added fixes to see whether the issue is triggered.

Or maybe you could check your fix towards https://github.com/ because this page stably reproduces the issue, at least for me.

I have indeed got some further checks regarding this very issue and have it reported here: https://bugs.chromium.org/p/chromium/issues/detail?id=1196004
In the issue report here I stated the very conditions that I think could trigger this issue, together with several minimal test cases I created.

Androbin:
I have tested every combination of { without the fix, with the fix } and { https://github.com, minimal example included in PR }

src/common/NetworkManager.ts
```diff
@@ -74,6 +74,10 @@ export class NetworkManager extends EventEmitter {
     string,
     Protocol.Network.RequestWillBeSentEvent
   >();
+  _requestIdToRequestPausedEvent = new Map<
```
OrKoN, line 107: Could we add a comment explaining how this works? I.e., which events under which circumstances are expected to be stored here.

src/common/NetworkManager.ts
```diff
@@ -242,8 +246,14 @@ export class NetworkManager extends EventEmitter {
...
         this._requestIdToInterceptionId.delete(requestId);
-      } else {
+      }
+      if (
+        !interceptionId ||
+        (this._requestIdToRequestPausedEvent.has(requestId) &&
+          event.type === 'Font')
+      ) {
         this._requestIdToRequestWillBeSentEvent.set(event.requestId, event);
+        this._requestIdToRequestPausedEvent.delete(requestId);
```
OrKoN, line 290: Is it possible that this map will grow indefinitely? E.g., is _onRequestWillBeSent called if a request is aborted?

Androbin: See https://github.com/puppeteer/puppeteer/pull/6996#issuecomment-813797393

Variant A (stylesheets etc.)
- `_onRequestWillBeSent`
- `_onRequestPaused`

Variant B (fonts, cache disabled)
- `_onRequestPaused`
- `_onRequestWillBeSent`

Variant C (fonts, cache enabled)
- `_onRequestPaused`
- `_onRequestWillBeSent`
- `_onRequestPaused` <-- addressed by this PR

The branch of `_onRequestPaused` in question only runs if it was called before `_onRequestWillBeSent`, so only for variant B and the first call in variant C.
The subsequent call to `_onRequestWillBeSent` will consume `_requestIdToRequestPausedEvent`.
It will also consume `_requestIdToInterceptionId` ~~but only if `event.type === 'Font'`~~.

Androbin: We could remove the entry in `_onLoadingFinished` and `_onLoadingFailed`, one of them should be called in any case.

src/common/NetworkManager.ts
Androbin: See https://github.com/puppeteer/puppeteer/pull/7060#discussion_r608784254

- If `_onRequestWillBeSent` runs before `_onRequestPaused`, the event will be stored in `_requestIdToRequestWillBeSentEvent`.
- If `_onRequestPaused` runs before `_onRequestWillBeSent`, the event will be stored in `_requestIdToRequestPausedEvent`.

Androbin: I'll add a comment for both explaining that the event order is variable and that we need both for `_onRequest`, so we store whichever event arrives first.

Androbin:
Turns out that for redirect requests, we receive multiple calls to to `_onRequestWillBeSent` with the same `requestId` and the current changeset might lead to `_onRequestPaused` calling `_onRequest` with a previous `_onRequestWillBeSent` event. I'm already working on a solution.

Androbin:
The `interceptionId`s seem to provide a clue. For fonts, they might be `interception-job-28.0` and `interception-job-35.0`, while for redirect requests, they might be `interception-job-28.0` and `interception-job-28.1`.

Androbin:
In `_onRequestPaused`, we can compare the `interceptionId` to the previously stored one (if there is one).
If it's a redirect, we just drop the call to `_onRequest`.

src/common/NetworkManager.ts
```diff
@@ -238,13 +267,18 @@ export class NetworkManager extends EventEmitter {
       !event.request.url.startsWith('data:')
     ) {
       const requestId = event.requestId;
-      const interceptionId = this._requestIdToInterceptionId.get(requestId);
-      if (interceptionId) {
+      const requestPausedEvent = this._requestIdToRequestPausedEvent.get(
+        requestId
+      );
+
+      this._requestIdToRequestWillBeSentEvent.set(requestId, event);
```
Androbin, line 285: In theory, the condition for this line could be stricter:
`!requestPausedEvent || (!this._cacheDisabled() && event.type === 'Font')`
But we don't know for sure if that would cover all cases.

src/common/NetworkManager.ts
```diff
@@ -238,13 +267,18 @@ export class NetworkManager extends EventEmitter {
       !event.request.url.startsWith('data:')
     ) {
       const requestId = event.requestId;
-      const interceptionId = this._requestIdToInterceptionId.get(requestId);
-      if (interceptionId) {
+      const requestPausedEvent = this._requestIdToRequestPausedEvent.get(
```
OrKoN, line 281: should we swap the lines 270 & 274?

OrKoN:
Jan, could you please take a look as well?

src/common/NetworkManager.ts
Androbin: What is your reasoning?

OrKoN: My reasoning (more of a nit) is that requestPausedEvent is not used until the line 276.

OrKoN:
In general, it looks good to me but let's wait for a review for Jan since he should know more about the request interception.

jschfflr:
Yes, sure! 
This patch looks good to me too from a technical side. But I'm not sure if this is the best way to fix the problem as it seems like this might be better fixed on the chromium side to get a more coherent flow of events. 
Do you know about some documentation that specify in which order we can expect the events to arrive?

Androbin:
The documentation for `Fetch.requestPaused` suggests that it's supposed to fire after `Network.requestWillBeSent`.
Experiments show that with the cache disabled, this holds true, but enabled, the order of those events is reversed.
And in the case of fonts, `Fetch.requestPaused` fires both before AND after `Network.requestWillBeSent`.
The current code anticipates the order to be reversed but it does not anticipate multiple `Fetch.requestPaused` events.

Androbin:
There are three commits in this PR: testing, refactoring, fixing
The fixing commit consists of storing the `requestWillBeSent` event unconditionally and a fix for redirects.
For redirect requests, Chromium sends multiple `requestWillBeSent` events with the same `requestId`.
That means if we just check the `requestId`, we might now pair `requestPaused` with the wrong one.
A redirect might change either the `url`, the `method` or both, so we compare the values for both events.

src/common/NetworkManager.ts
```diff
@@ -252,12 +271,12 @@ export class NetworkManager extends EventEmitter {
         requestId
       );
 
+      this._requestIdToRequestWillBeSentEvent.set(requestId, event);
+
       if (requestPausedEvent) {
         const interceptionId = requestPausedEvent.requestId;
         this._onRequest(event, interceptionId);
         this._requestIdToRequestPausedEvent.delete(requestId);
-      } else {
```
Androbin, line 245: @jschfflr Dropping this condition is the only workaround introduced, the rest is just refactoring.

src/common/NetworkManager.ts
```diff
@@ -252,12 +271,12 @@ export class NetworkManager extends EventEmitter {
...
-        this._requestIdToRequestWillBeSentEvent.set(requestId, event);
```
Androbin: @jschfflr Making this line run unconditionally is the only "workaround" introduced, the rest is just refactoring.

Androbin: And the value is cleared anyway. I really don't see why we wouldn't do this regardless of Chomium changes.

Androbin:
The real issue is that the user interception code is called twice (once per `requestPaused` event).
This PR does not attempt to work around that, that's why we still need a fix to land in Chromium.

src/common/NetworkManager.ts
```diff
@@ -69,19 +69,43 @@ export class NetworkManager extends EventEmitter {
   _client: CDPSession;
   _ignoreHTTPSErrors: boolean;
   _frameManager: FrameManager;
-  _requestIdToRequest = new Map<string, HTTPRequest>();
+
+  /*
+   * There are four possible orders of events:
+   *  A. `_onRequestWillBeSent`
+   *  B. `_onRequestWillBeSent`, `_onRequestPaused`
+   *  C. `_onRequestPaused`, `_onRequestWillBeSent`
```
OrKoN, line 77: when does the C sequence happen? So A is when there is no interception and B is the normal interception flow?

Androbin: Yes, that is correct. The difference between B and C is undocumented.
Seems to depend on whether the cache is enabled and whether the cache is hit or missed: https://github.com/puppeteer/puppeteer/issues/7038#issuecomment-817412431

Androbin: This module has always been explicitly anticipating both B and C but it previously failed for D.
D is trivially fixed by preserving the entry in `_requestIdToRequestWillBeSentEvent` until `_onLoadingFinished` (and adding a guard clause for redirect requests as they have the same `requestId`): https://github.com/puppeteer/puppeteer/pull/7060/commits/5f183e68e691acc7b8115f7c09345921214e4a72#r613342144

jschfflr (approved):
This change looks good to me from a technical standpoint.
But could you add some documentation around the expected order of events for redirects?

Androbin:
@OrKoN Here's a remark from upstream: https://bugs.chromium.org/p/chromium/issues/detail?id=1196004#c10

> There's no guaranteed order between `Fetch.requestPaused` and `Network.requestWillBeSent`, these are inherently racy because they come from different processes.

Androbin:
CI has a flaky test, same test failing on `main`

OrKoN:
@jschfflr Have you looked into this Firefox test failure already?

jschfflr:
@OrKoN Yes, I'm investigating in https://github.com/puppeteer/puppeteer/pull/7182

OrKoN (approved):

PR #7064, opened 2021-04-06
test: fix Browser.pages should return all of the pages

kblok:
If `expect(allPages.length).toBe(1);`, `allPages[1]` will be undefined, therefore, that `expect` makes no sense.


jackfranklin (approved):
Thanks!

PR #7066, opened 2021-04-07
Fixed warning for export of "MediaFeature" from Page.ts

KinetiCore-Labs:
warning in line 1510 on Page.ts, 
I exported "MediaFeature" at line 135 in "Page.ts".


jschfflr:
Thanks for your contribution! 
But it looks like the problem has already been addressed in https://github.com/puppeteer/puppeteer/pull/7071.

PR #7067, opened 2021-04-07
Fixed warning of line 300 in Page.ts

tasneemkoushar:
On running "npm run generate-docs",  getting warning "The TSDoc tag "@see" is not supported by this tool" in line 300 on "Page.ts"
As @see is not supported by TSDoc so I changed it to {@link}


jackfranklin:
@tasneemkoushar thanks for this PR ! I think the issue is that your git commit messages aren't passing the check:

```
> puppeteer@8.0.0-post commitlint /home/runner/work/puppeteer/puppeteer
> commitlint --from=HEAD~1

⧗   input: Removed @see tag from line 300 in "Page.ts"
✖   subject may not be empty [subject-empty]
✖   type may not be empty [type-empty]

✖   found 2 problems, 0 warnings
ⓘ   Get help: https://github.com/conventional-changelog/commitlint/#what-is-commitlint
```

If you change the last commit message to be something like:

```
fix: remove invalid @see documentation reference
```

Then I think commitlint should be happy.

jackfranklin (approved):

tasneemkoushar:
@jackfranklin Sir you have approved my changes, but I can see some checks were not successful, most probably because of .eslintrc
code style. Will my code gets merged with the present state or do I need to fix these checks?

jackfranklin:
@tasneemkoushar I think we can close this PR now as you've made more since that have solved the issues :)

PR #7072, opened 2021-04-08
fix(types): allow evaluate functions to take a readonly array as an argument

### LINKED ISSUE #6953: page.evaluate* functions typings don't accept readonly arrays

* Puppeteer version: 8.0.0
* Platform / OS version: macOS 11.2
* Node.js version: 12.18.1
* TypeScript version: 3.8.3

```

const readonlyArray: readonly string[] = ['a']
page.evaluate(() => {}, readonlyArray)
```

No error

**What happens instead?**
Error at the last line:
```
Argument of type 'readonly string[]' is not assignable to parameter of type 'SerializableOrJSHandle'.
  Type 'readonly string[]' is not assignable to type 'JSONObject'.
    Index signature is missing in type 'readonly string[]'.ts(2345)
```
Screen Shot 2021-03-04 at 13 35 53

n10v:
Fixes #6953


jackfranklin:
Thanks for this change (and sorry for the slow reply).

I'm curious, if we made the type `JSONArray = readonly Serializable[]`, would that work? I think a mutable array can become a readonly, but not the other way round. So if so we can avoid the union type and just define it with `readonly`.

n10v:
Hi @jackfranklin,
yes, [it seems so](https://www.typescriptlang.org/play?ssl=9&ssc=1&pln=10&pc=1#code/C4TwDgpgBAyhBOBLAhgG0QL2QI1dAvAFBRQA+UAdgK4C22CxZUAzsEhQOaPnYD2veZBW6UqqVCIBSMAPIA5AILx4yEFNlyZ2AFYQAxsADchQqEhRp8pSpBR8UeBGQATXhVS24SNJhx4A2gC6xoSIFMAIAGbIetCWmjr6wFAA3oz+ANYQIABcLGxhHIF5XijoWLgQxgC+JnpurFAQAG5oVMgRdlAAFMjwHCUIZb6VAJR2AHypUPUUzAIQAHSovBy9-ePVIbONji5uHtaqeXuu7rbUdAhBXf4AjAA0AEwPAMzBhC1tHRDdpwcgI4gUbbBrJU4AdyQESBeUu9HgN3s92eb0Cn1aqHaET+TmcUMQMOUqhBJiAA) 👍 
I will change the PR later today.

n10v:
Updated. Please take a look :)

jschfflr (approved):
This change looks good to me, thanks!

PR #7080, opened 2021-04-12
feat(page): emit the event after removing the Worker

### LINKED ISSUE #7079: Page: The WorkerDestroyed event should be emitted after deleting the Worker
I was going through the code to understand how puppeteer works and found this code fragment:

```ts
client.on('Target.detachedFromTarget', (event) => {
  const worker = this._workers.get(event.sessionId);
  if (!worker) return;
  this.emit(PageEmittedEvents.WorkerDestroyed, worker);
  this._workers.delete(event.sessionId);
});
```
> Copied from [Page.ts, L507-L512](https://github.com/puppeteer/puppeteer/blob/main/src/common/Page.ts#L507-#L512)

Here, the worker is being deleted only after the event has been sent. The order should be reverse.

pallavJha:
- Fixes #7079

### DIFF
--- a/src/common/Page.ts
+++ b/src/common/Page.ts
@@ -507,8 +507,8 @@ export class Page extends EventEmitter {
     client.on('Target.detachedFromTarget', (event) => {
       const worker = this._workers.get(event.sessionId);
       if (!worker) return;
-      this.emit(PageEmittedEvents.WorkerDestroyed, worker);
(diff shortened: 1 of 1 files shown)


pallavJha:
@googlebot  I signed it!

jschfflr (approved):

Issue #6462, opened 2020-10-01
Page.click() does not work with an input selector


pauloacosta:
Ps.: View the issue #3347 - The problem continues.

Steps to reproduce
Tell us about your environment:

Puppeteer version: 5.3.0
Platform / OS version: Windows 10
URLs (if applicable):
Node.js version: v12.18.3
What steps will reproduce the problem?

I am trying to use page.click() to click at a Button . If I use the selector button[type='button'] in developer console it works. However, puppeteer keeps throwing an error that no node was found.

This issue is the same problem related at: https://github.com/puppeteer/puppeteer/issues/3347

I am using a workaround:

        selector = 'button[type="button"]';
        await youtube.page.waitForSelector (selector);
        await youtube.page.evaluate((selector) => document.querySelector(selector).click(), selector);

gidztech:
I just ran into this and for me it was because the element wasn't in view. According to https://github.com/puppeteer/puppeteer/issues/2190#issuecomment-380254352, Puppeteer will attempt to scroll the element into view before clicking, but it can fail sometimes, like if you have custom scrolling behavior.

I was able to get it working for me with:

```
await page.evaluate(selector => document.querySelector(selector).scrollIntoView(), [mySelector]);
await page.click(mySelector);
```

I think Puppeteer should be throwing an error if it fails to get the element in view itself, as part of the `click` function, because right now it thinks it's successfully clicked the element but it didn't. This is misleading. 

I know it errors if you click an element that has no height.

Issue #3347, opened 2018-10-04
Page.click() does not work with an input selector


mehtaanirudh:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.8.0
* Platform / OS version: Windows 10
* URLs (if applicable):
* Node.js version: 8.11.4

**What steps will reproduce the problem?**

I am trying to use page.click() to go to the username filed in Instagram. If I use the selector input[name='username'] in developer console it works. However, puppeteer keeps throwing an error that no node was found

_Please include code that reproduces the issue._

```
const puppeteer = require('puppeteer');

(async () => {
	 try {
	const Browser = await puppeteer.launch({headless:false,
										defaultViewport:{width:600,height:800},
										waitUntil: ['load','domcontentloaded','networkidle0','networkidle2'],
						                timeout :0
						            	});
	const page=await Browser.newPage();
		
    await page.goto('https://www.instagram.com/accounts/login/?source=auth_switcher');
    
    await page.click('input[name="username"]');

    await page.keyboard.type('NewUser');
	await Browser.close();
	
	console.log("Iam done!");
	

	} catch(e) {console.log('main program error:' + e);
	}
	
})();
```

**What is the expected result?**

I should be able to type my username

**What happens instead?**

Puppeteer can't find the user name field

russelldmatt:
I'm running into the same issue.  Strangely, if I use page.evaluate, it does work.  For example:

```
let selector = 'input[name="username"]';

// does not work
await page.click(selector); 

// does work
await page.evaluate((selector) => document.querySelector(selector).click(), selector); 
```

I'd be curious if the same workaround works for you.

mehtaanirudh:
> I'm running into the same issue. Strangely, if I use page.evaluate, it does work. For example:
> 
> ```
> let selector = 'input[name="username"]';
> 
> // does not work
> await page.click(selector); 
> 
> // does work
> await page.evaluate((selector) => document.querySelector(selector).click(), selector); 
> ```
> I'd be curious if the same workaround works for you.

Thanks. I think it was something to do with the form not uploading before the click event firing. Which is weird because in my launch function I explicitly asks it to wait until the page has been loaded. Nevertheless,  I was able to fix it by adding some random delay:

 await page.goto('https://www.instagram.com/accounts/login/?source=auth_switcher');
    await utilities.addDelay(6000+Math.random()*832+ Math.floor(Math.random()*2312));
    await page.click('input[name="username"]');

[...]

PR #7097, opened 2021-04-17
fix(page): fix Mouse.click method

ptommasi:
page.click method relies on Mouse.click method for execution. Mouse.click triggers the move, down and up methods in parallel waiting for all of them to finish, when they should be called sequentially instead.

Issues: #6462 (related), #3347 (left a comment in it).

### DIFF
--- a/src/common/Input.ts
+++ b/src/common/Input.ts
@@ -409,15 +409,14 @@ export class Mouse {
   ): Promise<void> {
     const { delay = null } = options;
     if (delay !== null) {
-      await Promise.all([this.move(x, y), this.down(options)]);
+      await this.move(x, y);
+      await this.down(options);
       await new Promise((f) => setTimeout(f, delay));
       await this.up(options);
     } else {
-      await Promise.all([
-        this.move(x, y),
-        this.down(options),
-        this.up(options),
-      ]);
+      await this.move(x, y);
+      await this.down(options);
+      await this.up(options);
     }
   }
 


mathiasbynens (approved):

mathiasbynens:
Good catch!

ptommasi:
Branch deleted for clean up, big thanks for your work! :)

ondrejvelisek:
It seems to me this fix could increas probability to unexpected buggy behaviour. Kind of race condition. Since click is not an atomic operation. There could be another external mouse.move call between mouse.move and mouse.down. In such case puppeteer would click on a unexpected and potentialy dangerous position. For example (but not exclusivley) in headful mode where user moves mouse over the screen manually. Am I wrong?

ptommasi:
> For example (but not exclusivley) in headful mode where user moves mouse over the screen manually. Am I wrong?

I think you are indeed correct, but out of the two implementations (Promise.all vs sequential await) the latter is more predictable and much safer, in my opinion. Using the Promise.all implementation would not ensure that mouse down will happen before mouse up, and will not ensure that either of them will happen after the move.

As a consequence, page.click might be ignored (e.g. sequence ends up being up, move, down), it could cause a drag (e.g. down, move, up) or any other combination which is not move -> down -> up would cause a problem. In the issues I saw, people were complaining about the unpredictability of page.click, and Promise.all would definitely explain this type of behaviour.

My two cents is that is better to have an implementation that works most of the time and that is observable rather than an unpredictable one which could or could not work (as the issues demonstrated).

ondrejvelisek:
Agree. It is definitely better this way.

Anyway it is fair to say and post it here for the future.

To fix it Chrome would need to implement some kind of atomic click operation or a way to send multiple sequential operations together within one call. Which what I know it is not able. Until then i'm afraid Puppeteer has no way to fix it.

ptommasi:
To my understanding, _page.click_ reason to exist is to simulate user behaviour by marking the click event as trusted (based on [this comment](https://github.com/microsoft/playwright/issues/1808#issuecomment-614849382)). Single soft clicks can be still sent to the browser using _page.evaluate_, which is indeed what I still do at the moment (since I don't know exactly when the fix will be pushed in the deployed version, and the website I'm scraping doesn't seem to make a difference between trusted and untrusted events). For your use case, I suppose you could easily create a util function based on _page.$eval(selector, (element) => element.click())_ (or _page.evaluate_).

Disclaimer: I'm not a front-end developer, I'm not a puppeteer developer, I'm barely a developer, I used puppeteer in a hobbyist fashion for one of my pet project, so just keep in mind that my comments are based on very little experience on the matter! 😣

PR #7099, opened 2021-04-18
feat(launcher): fix installation error on Apple M1 chips

### LINKED ISSUE #6622: Installation fails on Apple Silicon / M1
### Steps to reproduce

* Puppeteer version: 5.5.0
* Platform / OS version: macOS Big Sur 11.0.1 (Apple Silicon)
* Node.js version: 15.2.1

1. Install `puppeteer` using `yarn` or `npm`

`puppeteer` gets installed

**What happens instead?**
Installation fails.

```
error path-to-project/node_modules/puppeteer: Command failed.
Exit code: 1
Command: node install.js
Arguments:
Directory: path-to-project/node_modules/puppeteer
Output:
The chromium binary is not available for arm64:
If you are on Ubuntu, you can install with:

 apt-get install chromium-browser

path-to-project/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:112
            throw new Error();
            ^
```

marvinhagemeister:
The previous logic assumed that an arm64 arch is only available in Linux. WIth Apple's arm64 M1 Chip this assumption isn't true anymore.

Currently there are no official MacOS arm64 chromium builds available, but we can make use of the excellent Rosetta feature in MacOS that allows us to run x86 binaries on M1.

Once official native MacOS arm64 chromium builds are available we should switch to those.

Fixes #6622

### DIFF
--- a/src/node/BrowserFetcher.ts
+++ b/src/node/BrowserFetcher.ts
@@ -294,8 +294,8 @@ export class BrowserFetcher {
     if (!(await existsAsync(this._downloadsFolder)))
       await mkdirAsync(this._downloadsFolder);
 
-    // Use intel x86 build on Apple M1 until native MacOS arm64
-    // chromium builds are available
+    // Use Intel x86 builds on Apple M1 until native macOS arm64
+    // Chromium builds are available.
     if (os.platform() !== 'darwin' && os.arch() === 'arm64') {
       handleArm64();
       return;
--- a/src/node/BrowserFetcher.ts
+++ b/src/node/BrowserFetcher.ts
@@ -293,7 +293,10 @@ export class BrowserFetcher {
     if (await existsAsync(outputPath)) return this.revisionInfo(revision);
     if (!(await existsAsync(this._downloadsFolder)))
       await mkdirAsync(this._downloadsFolder);
-    if (os.arch() === 'arm64') {
+
+    // Use intel x86 build on Apple M1 until native MacOS arm64
+    // chromium builds are available
+    if (os.platform() !== 'darwin' && os.arch() === 'arm64') {
       handleArm64();
       return;
     }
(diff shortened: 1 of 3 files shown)


mathiasbynens (approved):

src/node/BrowserFetcher.ts
```diff
@@ -293,7 +293,10 @@ export class BrowserFetcher {
     if (await existsAsync(outputPath)) return this.revisionInfo(revision);
     if (!(await existsAsync(this._downloadsFolder)))
       await mkdirAsync(this._downloadsFolder);
-    if (os.arch() === 'arm64') {
+
+    // Use intel x86 build on Apple M1 until native MacOS arm64
+    // chromium builds are available
```
mathiasbynens:
```suggestion
    // Use Intel x86 builds on Apple M1 until native macOS arm64
    // Chromium builds are available.
```

src/node/Launcher.ts
```diff
@@ -105,7 +105,9 @@ class ChromeLauncher implements ProductLauncher {
 
     let chromeExecutable = executablePath;
     if (!executablePath) {
-      if (os.arch() === 'arm64') {
+      // Use intel x86 build on Apple M1 until native MacOS arm64
+      // chromium builds are available
```
mathiasbynens:
```suggestion
    // Use Intel x86 builds on Apple M1 until native macOS arm64
    // Chromium builds are available.
```

src/node/install.ts
```diff
@@ -90,7 +90,9 @@ export async function downloadBrowser() {
     if (NPM_NO_PROXY) process.env.NO_PROXY = NPM_NO_PROXY;
 
     function onSuccess(localRevisions: string[]): void {
-      if (os.arch() !== 'arm64') {
+      // Use intel x86 build on Apple M1 until native MacOS arm64
+      // chromium builds are available
```
mathiasbynens:
```suggestion
    // Use Intel x86 builds on Apple M1 until native macOS arm64
    // Chromium builds are available.
```

l198881 (changes requested):
.

PR #7150, opened 2021-04-24
Add drag-and-drop support.

### LINKED ISSUE #1376: Support Drag-and-Drop
We support `ElementHandle.uploadFile` for `<input type=file>` but we don't support emulating dragging in files onto an arbitrary element.

danpark-salesforce:
This PR adds drag-and-drop support, leveraging new additions to the CDP Input domain (Input.setInterceptDrags, Input.dispatchDragEvent, and Input.dragIntercepted).                                            

Fixes #1376

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2988,10 +2988,10 @@ Dispatches a `mousedown` event.
 
 #### mouse.drag(start, target)
 
-- `start` <[Object]>
+- `start` <[Object]> the position to start dragging from
   - `x` <[number]> x coordinate
   - `y` <[number]> y coordinate
-- `target` <[Object]>
+- `target` <[Object]> the position to drag to
   - `x` <[number]> x coordinate
   - `y` <[number]> y coordinate
 - returns: <[Promise<[DragData]>]>
@@ -4142,7 +4142,7 @@ This method creates and captures a drag event from the element.
 
 #### elementHandle.dragAndDrop(target[, options])
 
-- `target` <[Object]>
+- `target` <[ElementHandle]>
 - `options` <[Object]>
   - `delay` <[number]> how long to delay before dropping onto the target element
 - returns: <[Promise]>
--- a/docs/api.md
+++ b/docs/api.md
@@ -155,6 +155,7 @@
   * [page.goto(url[, options])](#pagegotourl-options)
   * [page.hover(selector)](#pagehoverselector)
   * [page.isClosed()](#pageisclosed)
+  * [page.isDragInterceptionEnabled](#pageisdraginterceptionenabled)
   * [page.isJavaScriptEnabled()](#pageisjavascriptenabled)
   * [page.keyboard](#pagekeyboard)
   * [page.mainFrame()](#pagemainframe)
@@ -171,6 +172,7 @@
   * [page.setCookie(...cookies)](#pagesetcookiecookies)
   * [page.setDefaultNavigationTimeout(timeout)](#pagesetdefaultnavigationtimeouttimeout)
   * [page.setDefaultTimeout(timeout)](#pagesetdefaulttimeouttimeout)
+  * [page.setDragInterception(enabled)](#pagesetdraginterceptionenabled)
   * [page.setExtraHTTPHeaders(headers)](#pagesetextrahttpheadersheaders)
   * [page.setGeolocation(options)](#pagesetgeolocationoptions)
   * [page.setJavaScriptEnabled(enabled)](#pagesetjavascriptenabledenabled)
@@ -215,6 +217,10 @@
   * [mouse.click(x, y[, options])](#mouseclickx-y-options)
   * [mouse.down([options])](#mousedownoptions)
   * [mouse.drag(start, target)](#mousedragstart-target)
+  * [mouse.dragAndDrop(start, target[, options])](#mousedraganddropstart-target-options)
+  * [mouse.dragEnter(target, data)](#mousedragentertarget-data)
+  * [mouse.dragOver(target, data)](#mousedragovertarget-data)
+  * [mouse.drop(target, data)](#mousedroptarget-data)
   * [mouse.move(x, y[, options])](#mousemovex-y-options)
   * [mouse.up([options])](#mouseupoptions)
   * [mouse.wheel([options])](#mousewheeloptions)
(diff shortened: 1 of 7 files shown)


danpark-salesforce:
Hello! I've been trying to test drag-and-drop at work using a variety of E2E tools like Cypress and Webdriver. Unfortunately like everyone else I kept running into a frustrating issue where the thing getting dragged would just stick to my mouse cursor instead of getting dropped at a given location.

Luckily, it turns out that Chromium recently added some new hooks for dragging and I'm happy to report they work great! 

I know this PR is a bit light on the tests, but I wanted to make sure this was going in the right direction before I spent more time. Looking forward to your feedback, thanks!

test/drag-and-drop.spec.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
```
mathiasbynens:
```suggestion
 * Copyright 2021 Google Inc. All rights reserved.
```

mathiasbynens:
@jschfflr Could you PTAL (and consider the record/replay perspective)?

jschfflr (changes requested):
First of all, the general direction of this change looks good to me. Thanks for taking the time to create this patch!
I just have some small questions:
I'm not sure why you chose to create a new CDP Session for the drag events - could you elaborate on that or just use the already existing one?
Also, there is the optional `modifiers` flag - could you make sure that the current set of modifiers from the keyboard is respected?
For the return value of the drag: WDYT about returning the DragData directly (instead of the event)?

danpark-salesforce:
Thanks @jschfflr for taking a look! Updated the PR based on your feedback.

RE: `I'm not sure why you chose to create a new CDP Session for the drag events - could you elaborate on that..?`

This was for isolating the drag interception in case something else was also listening for drag events on the existing client. If this isn't a concern, I can use the existing client.

On a different note, the `drop()` method is currently dispatching three drag events in sequence (dragEnter, dragOver, and drop). I was thinking it might be better to move the dragEnter and dragOver into separate methods. What are your thoughts?

jschfflr:
Thanks for updating the patch!
I'd prefer to reuse the existing client but you have a good point about the isolation. 
WDYT about approaching this in a similar way as the network interception by exposing a `page.setDragInterception(true)`?

If you feel like the drop events would also be useful on their own, feel free to also expose them. 
But I'd prefer to have a single `drop` method because developers might not know about the specific event order for drag'n'drop events.
Also: WDYT about exposing a single `dragAndDrop` function that takes two selectors and drags one of them into the other?

danpark-salesforce:
Updated and moved the drag interception to the Page class; thanks for pointing me to that.

>developers might not know about the specific event order for drag'n'drop events.

Makes total sense, will keep the drop method as is. ~I'll also add methods for the other drag events in case things like dragover effects need to be tested.~ (edit: I think adding a delay parameter to the `drop()` would be simpler).

>Also: WDYT about exposing a single dragAndDrop function that takes two selectors and drags one of them into the other?

I think a `dragAndDrop` function is a good idea. One issue I do see with using selectors is it won't work with web components / shadow DOM. What are your thoughts on passing either selectors or element handles? Also, if we go down this route, we could add the `dragAndDrop` method directly to the Input as well.

jschfflr:
Awesome, thanks! 
I think, having individual methods for each event could still be valuable if a user only wants to test one of them.
But the delay parameter sounds super helpful too. Could you add both?

We do have the `pierce/` query handler to deal with selectors that use the shadow dom. But I agree with you that using element handles is a better solution. Let's use them instead of the selectors.

jschfflr:
@danparksf Hi, did you already have time to look into the changes we discussed? 
I'd love to get this landed soon :)

danpark-salesforce:
Hi @jschfflr, thanks for the feedback! Been meaning to look into updating this PR and finally have some time to work on them. Should have an update later today, thanks!

danpark-salesforce:
@jschfflr haven't added the `dragAndDrop()` yet since I wanted to check something with you. In order to drag to a target element, we need to get the x, y coordinates of the target element. At the moment there is a method `_clickablePoint()` that can do that for us. Are you fine with having a public method to expose it?

jschfflr:
@danparksf Yes, feel free to mark this function as public!

jschfflr:
Maybe you can also change the return type and expose a public `Point` type similar to `BoundingBox`

danpark-salesforce:
Thanks for the tips, PR updated!

jschfflr:
@danparksf Not sure if you are already aware of this, but you can use `npm run lint` locally to find all the lint errors that will also occur on the bots later.

danpark-salesforce:
Thanks, should be passing the lint check now.

jschfflr:
@danparksf Thanks! The linter is happy now. Could you also look into the failing unit tests? You can run them locally with `npm run unit`

danpark-salesforce:
@jschfflr looks like you fixed the tests.. thanks for doing that!

docs/api.md
```diff
@@ -2960,6 +2986,62 @@ Shortcut for [`mouse.move`](#mousemovex-y-options), [`mouse.down`](#mousedownopt
...
+
+- `start` <[Object]>
+  - `x` <[number]> x coordinate
+  - `y` <[number]> y coordinate
+- `target` <[Object]>
+  - `x` <[number]> x coordinate
+  - `y` <[number]> y coordinate
+- returns: <[Promise<[DragData]>]>
+
+This method creates and captures a dragevent from a given point.
```
jschfflr, line 2997: Can you explain what target is used for in this context?

docs/api.md
```diff
@@ -4045,6 +4131,45 @@ If the element is detached from DOM, the method throws an error.
...
+- `target` <[Object]>
+  - `x` <[number]> x coordinate
+  - `y` <[number]> y coordinate
+- returns: <[Promise<[DragData]>]>
+
+This method creates and captures a drag event from the element.
+
+#### elementHandle.dragAndDrop(target[, options])
+
+- `target` <[Object]>
```
jschfflr: Can you change this to `ElementHandle`?

jschfflr (commented):
Looks great, thanks!

jschfflr (approved):
LGTM, thanks! Let's get this landed! 🎉

jschfflr:
You can fix the lint errors by running `npm run eslint-fix`.
And maybe you can change the tests back to `it` and use `describeChromeOnly` for the whole block instead.

jschfflr:
@danparksf Sorry for the confusion - github didn't show me the commits I pushed and therefore I thought it didn't work. 
I hope it's ok for you that I fixed the eslint errors. If yes, I'm happy to land the change!

danpark-salesforce:
@jschfflr no problem, thanks for the fix!

danpark-salesforce:
@googlebot I consent.

jschfflr:
@googlebot I consent.

jschfflr:
@danparksf Awesome, let's get this landed then :)

jschfflr:
@danparksf Congrats on getting this landed!

danpark-salesforce:
Thanks @jschfflr for your help!

src/common/Page.ts
```diff
@@ -748,6 +749,10 @@ export class Page extends EventEmitter {
     return this._accessibility;
   }
 
+  get isDragInterceptionEnabled(): boolean {
```
mathiasbynens, line 752: It's surprising/unexpected that this is a getter. Other `page.is*` APIs are methods. Patch: https://github.com/puppeteer/puppeteer/pull/7419

test/drag-and-drop.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+
+import expect from 'expect';
+import {
+  getTestState,
+  setupTestPageAndContextHooks,
+  setupTestBrowserHooks,
+  describeChromeOnly,
+} from './mocha-utils'; // eslint-disable-line import/extensions
+
+describeChromeOnly('Input.drag', function () {
```
mathiasbynens, line 25: `page.isDragInterceptionEnabled` isn’t tested.

jschfflr: Good catch, I'll add a test for it. Don't we have presubmit checks in place for this?

src/common/Page.ts
jschfflr: I looks like this has already been addressed in https://github.com/puppeteer/puppeteer/pull/7419

test/drag-and-drop.spec.ts
jschfflr: I looks like this has already been addressed in https://github.com/puppeteer/puppeteer/pull/7419

PR #7168, opened 2021-04-26
fix: change `rm -rf` to `rimraf`

jschfflr:
Currently, `npm clean-lib` fails on windows with `cmd` because it does not now about `rm`.
This change uses the already installed `rimraf` to do the job instead.


mathiasbynens:
Alex, could you PTAL?

OrKoN (approved):

PR #7192, opened 2021-04-30
feat: add option to filter targets

### LINKED ISSUE #7191: Add a way to ignore certain targets
As puppeteer enables some domains by default, it should have a way to ignore certain targets that should not be interfered with.

jschfflr:
This change closes  #7191 by introducing a callback that can be used to decide if puppeteer should connect to a given target.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -624,7 +624,7 @@ try {
   - `devtools` <[boolean]> Whether to auto-open a DevTools panel for each tab. If this option is `true`, the `headless` option will be set `false`.
   - `pipe` <[boolean]> Connects to the browser over a pipe instead of a WebSocket. Defaults to `false`.
   - `extraPrefsFirefox` <[Object]> Additional [preferences](https://developer.mozilla.org/en-US/docs/Mozilla/Preferences/Preference_reference) that can be passed to Firefox (see `PUPPETEER_PRODUCT`)
-  - `targetFilter` <?[function]\([Protocol.Target.TargetInfo]\):[Promise]<[boolean]>|[boolean]> Use this function to decide if Puppeteer should connect to the given target. By default, puppeteer will connect to all available targets.
+  - `targetFilter` <?[function]\([Protocol.Target.TargetInfo]\):[Promise]<[boolean]>|[boolean]> Use this function to decide if Puppeteer should connect to the given target. If a `targetFilter` is provided, Puppeteer only connects to targets for which `targetFilter` returns `true`. By default, Puppeteer connects to all available targets.
 - returns: <[Promise]<[Browser]>> Promise which resolves to browser instance.
 
 You can use `ignoreDefaultArgs` to filter out `--mute-audio` from default arguments:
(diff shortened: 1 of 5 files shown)


src/common/Browser.ts
```diff
@@ -330,6 +342,11 @@ export class Browser extends EventEmitter {
         ? this._contexts.get(browserContextId)
         : this._defaultContext;
 
+    const shouldIgnoreTarget = await this._targetFilterCallback(targetInfo);
```
mathiasbynens: If we're gonna use `filter` in the name, I think we should follow the JS language precedent of `Array.prototype.filter`, which works the other way around: if the callback returns `true`, it means the item is preserved. WDYT?

docs/api.md
```diff
@@ -516,6 +516,7 @@ Clears all registered handlers.
   - `slowMo` <[number]> Slows down Puppeteer operations by the specified amount of milliseconds. Useful so that you can see what is going on.
   - `transport` <[ConnectionTransport]> **Experimental** Specify a custom transport object for Puppeteer to use.
   - `product` <[string]> Possible values are: `chrome`, `firefox`. Defaults to `chrome`.
+  - `targetFilter` <?[function]\([Protocol.Target.TargetInfo]\):[Promise]<[boolean]>|[boolean]> Use this function to decide if Puppeteer should connect to the given target. By default, puppeteer will connect to all available targets.
```
mathiasbynens: Please clarify what the callback should return to keep/remove an item. Maybe something like:

```suggestion
  - `targetFilter` <?[function]\([Protocol.Target.TargetInfo]\):[Promise]<[boolean]>|[boolean]> Use this function to decide if Puppeteer should connect to the given target. If a `targetFilter` is provided, Puppeteer only connects to targets for which `targetFilter` returns `true`. By default, Puppeteer connects to all available targets.
```

(this phrasing assumes the logic is changed per my comment below)

mathiasbynens (changes requested):

src/common/Browser.ts
OrKoN: perhaps it'd be better to call it `shouldAttachToTarget` or smth like this?

jschfflr: Yes, that makes a lot of sense! I'll update the patch :)

mathiasbynens (approved):

OrKoN (approved):

stoffel007 (approved):

peterbe:
I have no idea how this PR works or what the changes mean, but in the release change log for version 9.1.0 this PR is the only one in that release. 
Ever since upgrading to puppeteer 9.1.0 we started seeing strange errors from `jest`. 
Our project has a bunch of `*.test.js` files and only a couple of them have anything to do with puppeteer. There are a couple that use `jest-puppeteer`. 

Unscientifically, I found that our test suite stopped being so busted when we downgraded back to puppeteer 9.0.0. 
Not sure who or how it helps. 

[Here's my issue about it](https://github.com/mdn/yari/issues/3694)
[Here's my PR to hack around it](https://github.com/mdn/yari/pull/3696)

jschfflr:
Hi @peterbe Thanks for bringing this to my attention!
I'll try to reproduce the problem and see what this might be about.

jschfflr:
@peterbe Thanks for pinging - I think this is related to the fact that the filter can be async. I'll verify that now...

peterbe:
What's a "filter" :)

jschfflr:
By default, puppeteer connects to all available targets (=pages and workers etc.). We wanted to give developers a way to control which ones to connect because puppeteer enables some CDP domains by default which might interfere with what that target is doing. 
Unfortunately, I allowed the filter to be async and awaited its result. That led to a situation in which the `targetInfoChanged` event was processed before the target was created because it was still waiting for the return of the `targetFilter`.

jribbens:
Yeah I have just raised issue #7204 which turns out to be due to this.

Issue #6753, opened 2021-01-15
Ship Puppeteer as JavaScript modules


mathiasbynens:
As Jack points out in https://twitter.com/Jack_Franklin/status/1349780548848390144, once [even the oldest maintenance LTS version of Node.js](https://github.com/puppeteer/puppeteer#usage) supports JavaScript modules, we can ship Puppeteer as pure JavaScript modules.

🥳 

https://2ality.com/2021/06/typescript-esm-nodejs.html

mathiasbynens:
We might want to wait for https://github.com/microsoft/TypeScript/pull/45884 to make it into a TypeScript release.

mathiasbynens:
As a first step, of course we should ship modules (in a way that Node.js can directly consume) alongside CommonJS. Then in the future, we can consider moving to modules-only.

OrKoN:
> As a first step, of course we should ship modules (in a way that Node.js can directly consume) alongside CommonJS. Then in the future, we can consider moving to modules-only.

any suggestions on how we do that?

mathiasbynens:
The link in my top post has some info.

mathiasbynens:
Also https://2ality.com/2019/10/hybrid-npm-packages.html

mathiasbynens:
Also https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77

PR #7200, opened 2021-05-03
chore: drop support for Node.js 10

mathiasbynens:
This is a prerequisite for shipping Puppeteer as JavaScript modules (#6753).

@jackfranklin I noticed that our tsconfig has `target` (and `module`) set at `ESNext`. Should we hardcode that at whatever our currently-supported Node.js versions support (in this case, that would be `ES2019`)?


jackfranklin:
Regarding the tsconfig target - yeah I reckon we should target whatever the lowest Node we support supports? WDYT?

Also not sure why there are so many Prettier related failures on CI; those shouldn't have made it through onto main :/

mathiasbynens:
Apparently `ES2019` is not a valid option for `module` specifically, so I went with the next best thing, which is `ES2015`.

jackfranklin (approved):

PR #7210, opened 2021-05-05
feat: add more Android models to DeviceDescriptors

mrienstra:
Adds device descriptions for:
* Galaxy S8
* Galaxy S9+
* Galaxy Tab S4
* Pixel 3
* Pixel 4

These devices are regarded as worthy of targeting by [BrowserStack](https://www.browserstack.com/test-on-the-right-mobile-devices).

Sources (both have identical data for these 5 devices):
1. https://github.com/aerokube/moon-deploy/blob/master/moon-local.yaml#L199
2. https://www.danhendricks.com/2018/04/adding-iphone-galaxy-chrome-mobile-emulated-devices/#heading_device_data


jackfranklin:
@mathiasbynens wdyt?

mathiasbynens:
I think the intention is to keep this in sync with the emulated devices in DevTools: https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/panels/emulation/EmulatedDevices.ts;l=637-1368;drc=229a54f7ea0133738b1a6d395507dc80e4f18de8 I'm not opposed to landing this, but the way to do it would be to suggest the change as a DevTools patch before landing it here. Would you be up for that?

mrienstra:
@mathiasbynens, sounds good, I took a crack at it: https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/2919612

jschfflr (approved):
This change has now landed in DevTools.

mrienstra:
Thank you @jschfflr, for your help here and also in DevTools, very much appreciated! 🙏

jschfflr:
@mrienstra Sure, happy to help!

Issue #6915, opened 2021-02-21
setRequestInterception disables caching. I need it not to do that.


jribbens:
The documentation says that `page.setRequestInterception` disables "page caching", without explaining what "page caching" is or why it is being disabled. Firstly, it looks like it actually means it disables all caching. Secondly, I need it not to do that. Why does it happen? There is no obvious connection between the two things. If I were to do:
```javascript
  await page.setRequestInterception(true)
  page.on('request', request => request.continue())
  await page._client.send('Network.setCacheDisabled', { cacheDisabled: false })
```
would anything bad happen, and if so, what?

whalderman:
On why that happens: https://github.com/puppeteer/puppeteer/issues/2905#issuecomment-406128495
Also see #6996

jribbens:
I think they're saying that the only problem is that if caching is not disabled, then requests served from the cache will not be intercepted? If so then it would be very much worth saying so in the documentation - I think for my application that wouldn't be a problem at all (if I didn't want to block or alter the request the first time it was issued, I'm unlikely to want to do so on subsequent occasions).

Androbin:
#7060 was just merged, it solves the issue that originally led to the decision to disable the cache

jribbens:
> #7060 was just merged, it solves the issue that originally led to the decision to disable the cache

Excellent, hopefully there will now be a patch to allow request interception with cache enabled :-)

Androbin:
@jribbens See #6996, and #7217

jribbens:
This issue was fixed in Puppeteer version 9.

Issue #4620, opened 2019-06-24
Validate setCacheEnable and setRequestInterception usage


kblok:
I got a question on Puppeteer-Sharp saying something like:

>I want to enable cache, to cache images, but I need to intercept my xhr request. When I set `setRequestInterception`, cache stop working.

So I ended up reading some changes made on #4260. Here [_updateProtocolCacheDisabled](https://github.com/GoogleChrome/puppeteer/pull/4260/files#diff-d2ac7cb061b0c51644d0a5d6140e3a32R148) to be more specific.

So, I have two questions regarding this piece of code:

```js
cacheDisabled: this._userCacheDisabled || this._protocolRequestInterceptionEnabled
````

1. Do we need to disable the cache there, when the user explicitly opted to enable it?
2. If so, as those are two explicit actions made by the user (`setCacheEnable` and `setRequestInterception`), Shouldn't we assert there saying something like `Cannot user setCacheEnable and setRequestInterception at the same time`?

**Comment by jonasgroendahl** [comment]
Any solution to this? Or is it just always better to have the cache option enabled to lower bandwidth?

**Comment by liweixi100** [comment]
Our test showed: when interception and caching were enabled at the same time, many websites will fail to load: https://github.com/puppeteer/puppeteer/issues/5176#issuecomment-554815586 

If the website of your interest happens to work, good for you. 

But I guess there was a reason why the API was coded that way, based on our tests at least.

**Comment by Androbin** [comment]
The issue described by @liweixi100 is described in #7038 and solved by #7060.

**Comment by kblok** [comment]
@Androbin, did #7060 solve this? can I close it?

**Comment by Androbin** [comment]
@kblok #7060 solved the issue underlying the need the disable the cache.
If we implement this fix in PuppeteerSharp, we can simplify write:

```
_cacheDisabled(): boolean {
  return this._userCacheDisabled;
}
```

Androbin:
See #1765

Androbin:
See https://github.com/hardkoded/puppeteer-sharp/pull/1765

Issue #4055, opened 2019-02-22
page.setRequestInterception(true) adds cache-control and pragma headers that cannot be removed


outlabeau:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 1.12.2
* Platform / OS version: Linux / Mac
* Node.js version: 10.13.0
* Browsers: both Chrome and Chromium

**What steps will reproduce the problem?**
Setting page.setRequestInterception(true) causes the **cache-control** and **pragma** headers to be set to **no-cache**.  While this kind of makes sense (the documentation states that request interception disables caching), we should be able to manually remove those request headers by setting them to **undefined**, as per the [documentation](https://pptr.dev/#?product=Puppeteer&version=v1.12.2&show=api-requestcontinueoverrides).

_Please include code that reproduces the issue._
```
await page.setRequestInterception(true);
page.on('request', request => {
  // Override headers
  const headers = Object.assign({}, request.headers(), {
    'cache-control': undefined,  // remove "cache-control" header
    pragma: undefined            // remove "pragma" header
  });
  request.continue({headers});
});
```

**What is the expected result?**
Request headers should not include **cache-control** and **pragma**.

**What happens instead?**
Request headers have **cache-control** and **pragma** headers set to **no-cache**.

fastjien:
me too, have you solved it?

fastjien:
@aslushnikov

larry80808:
I have the same problem. Setting cache-control and pragma to undefined do not work to remove these headers when request interception is turned on.

outlabeau:
I don't know if this is a Chromium issue or a Puppeteer issue, but seems like it should be fixed.

outlabeau:
I just posted a $50 bounty for this on [BountyHub](https://bountyhub.co/bounty/5d5da0140dac4e230c8ccf5d).

PR #7215, opened 2021-05-07
feat(requestinterception): don't disable cache by default

Androbin:
Reverts #1154 @aslushnikov disabling the cache by default.

People don't expect this behavior, see tickets like: #2905, #3403. #4055, #4620, #5176, #6915, ...

There were two reasons for this change:
- Some sites did not fire the `load` event with caching enabled.
    - analysed in #7038 @starrify 
    - fixed in #7060
- We cannot intercept requests that are answered from the cache.
    - Most people are willing to accept cached resources not being intercepted.

This flag has been introduced in #6996.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2219,7 +2219,7 @@ await page.setGeolocation({ latitude: 59.95, longitude: 30.31667 });
 #### page.setRequestInterception(value[, cacheSafe])
 
 - `value` <[boolean]> Whether to enable request interception.
-- `cacheSafe` <[boolean]> Whether to trust browser caching. If set to false, enabling request interception disables page caching. Defaults to false.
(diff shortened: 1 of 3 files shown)


Androbin:
@OrKoN @jschfflr PTAL

Androbin:
I personally prefer #7217

jschfflr:
@Androbin Could you elaborate what you want to achieve here?

Androbin:
There was bug that forced us to implicitly disable the cache when enabling interception.
This flag was added to allow people to override that behavior while keeping it the default behavior.
Now the bug in question is fixed and it is safe to change the default behavior.

Androbin:
People expect interception and caching to be independent and that's how it was before #1154.

jschfflr:
Closing this in favor of https://github.com/puppeteer/puppeteer/pull/7215

PR #7217, opened 2021-05-07
feat(requestinterception): remove cacheSafe flag

Androbin:
Reverts #6996 

#1154 @aslushnikov disabled the cache by default.

There were two reasons for this change:
- Some sites did not fire the `load` event with caching enabled.
    - analysed in #7038 @starrify 
    - fixed in #7060
- We cannot intercept requests that are answered from the cache.
    - Use `setCacheDisabled` instead.

### DIFF
--- a/src/common/NetworkManager.ts
+++ b/src/common/NetworkManager.ts
@@ -114,7 +114,6 @@ export class NetworkManager extends EventEmitter {
   _credentials?: Credentials = null;
   _attemptedAuthentications = new Set<string>();
   _userRequestInterceptionEnabled = false;
-  _userRequestInterceptionCacheSafe = false;
   _protocolRequestInterceptionEnabled = false;
   _userCacheDisabled = false;
   _emulatedNetworkConditions: InternalNetworkConditions = {
@@ -228,12 +227,8 @@ export class NetworkManager extends EventEmitter {
     await this._updateProtocolCacheDisabled();
   }
 
-  async setRequestInterception(
-    value: boolean,
-    cacheSafe = false
-  ): Promise<void> {
+  async setRequestInterception(value: boolean): Promise<void> {
     this._userRequestInterceptionEnabled = value;
-    this._userRequestInterceptionCacheSafe = cacheSafe;
     await this._updateProtocolRequestInterception();
   }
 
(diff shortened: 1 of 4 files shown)


Androbin:
@OrKoN @jschfflr PTAL

jschfflr:
Could you elaborate why you chose to remove this feature again?

Androbin:
There was a bug that forced us to implicitly disable the cache when enabling interception.
Some people were not affected by this bug and wanted to keep the cache enabled.
This flag was added to allow people to override that behavior while keeping it the default behavior.
Now the bug in question is fixed and there is no point in implicitly disabling the cache anymore.
This PR removes the logic around disabling the cache implicitly, including the workaround flag.
It is still possible to disable the cache explicitly using `setCacheDisabled`.

src/common/NetworkManager.ts
```diff
@@ -258,11 +253,7 @@ export class NetworkManager extends EventEmitter {
   }
 
   _cacheDisabled(): boolean {
-    return (
-      this._userCacheDisabled ||
-      (this._userRequestInterceptionEnabled &&
-        !this._userRequestInterceptionCacheSafe)
-    );
+    return this._userCacheDisabled;
```
Androbin, line 256: This is the core change of the PR.

Androbin: People expect interception to be independent of caching and that's how it was before #1154.

jschfflr:
Awesome, thanks for the explanation!

jschfflr (approved):
LGTM, thanks!

jschfflr:
Just to make sure I understand this correctly: This cl replaces https://github.com/puppeteer/puppeteer/pull/7215, right?

Androbin:
> Just to make sure I understand this correctly: This cl replaces #7215, right?

Yes, that is correct.

jschfflr:
Can you update this change? I think the Node.js v10 support was recently dropped and it seems like the check is still waiting for it...

Tyler-Murphy:
[`api.md` says](https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#pagesetrequestinterceptionvalue):

> NOTE Enabling request interception disables page caching.

Does that need to be removed? I'm not sure, because I might have misunderstood the effect of this change.

Androbin:
@Tyler-Murphy You are right, thanks for spotting this, I have created #7304 to update the docs.

jschfflr:
@Tyler-Murphy Thanks for reaching out!

PR #7224, opened 2021-05-10
docs: add lowercasing of HTTP headers to documentation

TimvdLippe:
I was adding extra HTTP headers that were case-sensitive and my
server wasn't seeing the corresponding headers. Turns out that
Puppeteer lowercases the headers, which was not written in the
documentation.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2189,6 +2189,8 @@ This setting will change the default maximum time for the following methods and 
 
 The extra HTTP headers will be sent with every request the page initiates.
 
+> **NOTE** All headers are lowercased, since HTTP headers are case insensitive.
+
 > **NOTE** page.setExtraHTTPHeaders does not guarantee the order of headers in the outgoing requests.
 
 #### page.setGeolocation(options)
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -2189,6 +2189,8 @@ This setting will change the default maximum time for the following methods and
 
 The extra HTTP headers will be sent with every request the page initiates.
 
+> **NOTE** All headers are lowercased, since HTTP headers are case insensitive.
```
mathiasbynens: This doesn't logically follow, so perhaps rephrase to:

```suggestion
> **NOTE** All HTTP header names are lowercased. (HTTP headers are case-insensitive, so this shouldn’t impact your server code.)
```

Since this is a small change I'll go ahead and apply this 👍

mathiasbynens:
Thanks! For the record, this happens here: https://github.com/puppeteer/puppeteer/blob/25ad7402e9c66736e1c5198f1ebcb7392dd3f8e1/src/common/NetworkManager.ts#L181

PR #7232, opened 2021-05-12
chore: upgrade and pin prettier dependencies

jackfranklin:
We're seeing odd failures with Prettier on some CI branches; my hunch
is that they are installing different versions of the package and
therefore getting formatting conflicts. This PR updates them all and
pins them to specific versions - something we should probably consider
generally, or remove our `package-lock.json` from the gitignore.


mathiasbynens (approved):

mathiasbynens:
I'd like to avoid checking in `package-lock.json` if we can. Pinning dependencies at the `package.json` level like in this PR SGTM. Thanks for looking into this, Jack!

PR #7238, opened 2021-05-13
chore: enforce pinned dependencies

jackfranklin:
Because we don't check our `package-lock.json` in, we can end up with
different versions installed locally vs CI, or even two devs having
different versions. Let's pin and enforce we pin every version to
avoid this.


jschfflr (approved):
Can you get husky to also run this check on `precommit` maybe?

jackfranklin:
👍  done!

PR #7243, opened 2021-05-16
build: allow to customize tmpdir

Kikobeats:
Hello,

This change allows customizing the directory for temporary files, setting up the current behavior as default.

I'm particularly interested in this change because I want to use a memory filesystem (`memfs`) to speed up a bit the puppeteer process.


Kikobeats:
hum checks are failing for a non related thing.

mathiasbynens (changes requested):
The option is named `tmpdir` but it should be `tmpDir` to match the casing convention.

If we want to do this then introducing an [environment variable](https://github.com/puppeteer/puppeteer/blob/v9.1.1/docs/api.md#environment-variables) like `PUPPETEER_TMP_DIR` seems like a cleaner option.

The patch would also need to include a test and the relevant changes to documentation. WDYT?

Kikobeats:
ops, fixed the camel case naming 🙂

I'm not sure about `PUPPETEER_TMP_DIR` because it's limiting the possibility to use a function where the path can change over time, rather than a static string.

mathiasbynens:
> I'm not sure about `PUPPETEER_TMP_DIR` because it's limiting the possibility to use a function where the path can change over time, rather than a static string.

If the path changes over time and your code knows about it, can't you have it update the `PUPPETEER_TMP_DIR` env var accordingly?

Kikobeats:
you're right; I'm assuming `os.tmpdir()` will return a different path every time, but it looks like constant value:

Screen Shot 2021-05-17 at 11 06 05

let me do the code changes for a new environment variable 🙂

Kikobeats:
@mathiasbynens done at https://github.com/puppeteer/puppeteer/pull/7243/commits/083ea5ff7e8f6a51b16db9699e237864961c9eaa

Do you think it could be moved into a common, so it just could be declared once?

Kikobeats:
@mathiasbynens Can I do something to move this PR forward? 🙂

mathiasbynens:
The diff seems to include lots of unrelated changes. Could you please check?

Kikobeats:
@mathiasbynens updated 👍

src/node/Launcher.ts
```diff
@@ -29,8 +29,11 @@ import {
   BrowserLaunchArgumentOptions,
   PuppeteerNodeLaunchOptions,
 } from './LaunchOptions.js';
+
 import { Product } from '../common/Product.js';
 
+const tmpDir = () => process.env.PUPPETEER_TMP_DIR || os.tmpdir()
```
mathiasbynens: Could you please fix the lint errors?

Kikobeats:
@mathiasbynens all green now 🙂

experimental/puppeteer-firefox/lib/Launcher.js
```diff
@@ -30,7 +30,9 @@ const WebSocketTransport = require('./WebSocketTransport');
 const mkdtempAsync = util.promisify(fs.mkdtemp);
 const removeFolderAsync = util.promisify(removeFolder);
 
-const FIREFOX_PROFILE_PATH = path.join(os.tmpdir(), 'puppeteer_firefox_profile-');
+const tmpDir = () => process.env.PUPPETEER_TMP_DIR || os.tmpdir();
+
+const FIREFOX_PROFILE_PATH = path.join(tmpDir(),'puppeteer_firefox_profile-');
```
mathiasbynens:
```suggestion
const FIREFOX_PROFILE_PATH = path.join(tmpDir(), 'puppeteer_firefox_profile-');
```

docs/api.md
```diff
@@ -454,6 +454,7 @@ If Puppeteer doesn't find them in the environment during the installation step,
 
 - `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` - defines HTTP proxy settings that are used to download and run Chromium.
 - `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` - do not download bundled Chromium during installation step.
+- `PUPPETEER_TMP_DIR` - defines the directory to be used by Puppeteer for creating temporary files. It's [os.tmpdir](https://nodejs.org/api/os.html#os_os_tmpdir) by default.
```
```suggestion
- `PUPPETEER_TMP_DIR` - defines the directory to be used by Puppeteer for creating temporary files. Defaults to [`os.tmpdir()`](https://nodejs.org/api/os.html#os_os_tmpdir).
```

Kikobeats:
@mathiasbynens I think now can be merged

jschfflr (changes requested):
Could you maybe move the implementation of `const tmpDir = () => process.env.PUPPETEER_TMP_DIR || os.tmpdir();` into a separate file like `helpers.ts` or something similar to reduce code duplication?

Kikobeats:
> Could you maybe move the implementation of `const tmpDir = () => process.env.PUPPETEER_TMP_DIR || os.tmpdir();` into a separate file like `helpers.ts` or something similar to reduce code duplication?

I want to do that, `experimental/puppeteer-firefox/lib/Launcher.js` is CJS and `src/node/Launcher.ts` is ESM.

I guess I can't just create a file that can be loaded on both parts; am I wrong?

jschfflr:
> > Could you maybe move the implementation of `const tmpDir = () => process.env.PUPPETEER_TMP_DIR || os.tmpdir();` into a separate file like `helpers.ts` or something similar to reduce code duplication?
> 
> I want to do that, `experimental/puppeteer-firefox/lib/Launcher.js` is CJS and `src/node/Launcher.ts` is ESM.
> 
> I guess I can't just create a file that can be loaded on both parts; am I wrong?

Ah sorry, I didn't see that one is for Firefox and one in src. In this case, I think this change looks good!

jschfflr (approved):

Kikobeats:
@jschfflr @mathiasbynens any chance to merge this? it's already approved and tests pass on green 🙂

mathiasbynens (approved):

PR #7266, opened 2021-05-25
Added comments for some of the methods

tasneemkoushar:


src/common/JSHandle.ts
```diff
@@ -259,12 +259,13 @@ export class JSHandle {
   }
 
   /**
-   * Returns either `null` or the object handle itself, if the object handle is
-   * an instance of {@link ElementHandle}.
+   * @returns return either `null` or the object handle itself, if the object
```
jackfranklin: here you can just do:

```
@returns either...
```

Rather than:

```
@returns return either...
```

jackfranklin (commented):
Looks good with one small change, please!  Thanks :)

jackfranklin (approved):

PR #7324, opened 2021-06-11
chore: synchronize Bug 1710839: update comment about firefox preference remote.enabled

juliandescottes:
…ce remote.enabled

This is a minor comment we updated as we are in the process of deprecating this preference. Proposing to merge this upstream, as it will make our sync back to mozilla-central easier.


whimboo (approved):
Syncs the comment to what we have in downstream. The preference is no longer used in recent Firefox builds.

Issue #6563, opened 2020-10-30
Error: EPERM: operation not permitted


garosi:
[Error: EPERM: operation not permitted, unlink 'C:\Users\user\AppData\Local\Temp\puppeteer_dev_chrome_profile-LixSno\CrashpadMetrics-active.pma'] {
  errno: -4048,
  code: 'EPERM',
  syscall: 'unlink',}]

### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 5.4.1
* Platform / OS version: Windows 10
* URLs (if applicable):
* Node.js version : v14.13.0

**What steps will reproduce the problem?**
await browser.close()
I tried await page.close() but problem not solving

Somberor:
Happens to me too .

Jatz:
This also happens for me.

danielshiloah:
Happens to me too, took me all the way back to 4.0.1 for it to stop happening.
It also leaves the chromium instance open in the background.

gaspromov:
Happens to me too.

TheRealMal:
### Same

nikbeznosikov:
Same

denisbrodbeck:
While this is annoying behaviour, there is a workaround: close *all* pages before closing the browser

```js
// close all pages, fix perm issues on windows 10 (https://github.com/puppeteer/puppeteer/issues/298)
let pages = await browser.pages();
await Promise.all(pages.map(page =>page.close()));
await browser.close();
```

feijuncheng:
Happens to me too.

rebelinnovator:
Same

imbhaskarn:
This happened to me too but I got rid of this using this 
`await page.close();
 await browser.close()
`

michael-lloyd-morris:
The above solutions work, but we also need to make sure this tear down process takes place on SIGINT, SIGUSR1 and SIGUSR2

```
// Make sure Chromium is told to shut down if we are interrupted!!
  process.on("SIGINT", async () => {
    await this.tearDown();
  });
  process.on("SIGUSR1", async () => {
    await this.tearDown();
  });
  process.on("SIGUSR2", async () => {
    await this.tearDown();
  });
```
this.tearDown in the above contains the code @denisbrodbeck gave above in addition to other tear down operations specific to my app.

talles-coder:
[...]

PR #7338, opened 2021-06-18
fix(browser-runner): reject promise on error

andresrondon:
This patch adds a reject callback to the _processClosing promise and executes it if it catches an error on removeFolderAsync(...).

Worth noting that I found this while experiencing this issue: [original](https://github.com/puppeteer/puppeteer/issues/298), [fairly-recent duplicate](https://github.com/puppeteer/puppeteer/issues/6563).

This change DOES NOT attempt to fix the actual bug that prevents from removing CrashpadMetrics-active.pma temp file, but it solves the most annoying part which is letting the process hanging. This change should allow the client to catch the error in their own implementation and handle it adequately based on their needs.


mathiasbynens (commented):
Could you please

- add a test so that we don’t regress this behavior?
- update the documentation

andresrondon:
@mathiasbynens for sure. I will add that as soon as I have some free time. What part specifically from the doc should I update? Thanks.

mathiasbynens:
> What part specifically from the doc should I update?

`docs/api.md`. Just a brief note on how you might observe this new behavior as a user would be great 👍

badluck13:
I would really love to see this in the next version!

andresrondon:
Sorry for the delay. Been very busy (with a project using puppeteer hah). Will get on this this evening hopefully.

andresrondon:
Hi folks! Firstly, I added the required documentation in `docs/api.md`.

On the unit tests, after spending some hours, I think there's no practical way to unit test this without doing some other changes. Here's what happens:

This can only be reproduced if for some reason the temp folder cannot be deleted (for example: a different process is blocking a file in it, the folder no longer exists or insufficient permissions). Knowing this, the only practical way I can think of is changing the value of the `_tempDirectory` property used in line 99 of BrowserRunner.ts, which I can't do since it's private. I also thought of removing the folder before calling `browser.close()`, but again, I cannot do that because I can't know the randomly-generated name of the temp folder, and no way to access that private property I mentioned. Finally, I even thought of the following: 

```js
      it('should reject promise on error with deleting temp folder', async () => {
        const { puppeteer, defaultBrowserOptions } = getTestState();

        const browser = await puppeteer.launch(defaultBrowserOptions); // the temp folder gets created
        await browser.close(); // the temp folder gets removed and connection is disposed

        let error = undefined;

        try {
          await browser.close(); // will attempt to delete non-existent temp folder
        } catch (err) {
          error = err
        }

        expect(error).toBeDefined();
      });
```

but this wouldn't work either because the attempt to delete the folder occurs in a `once` event, meaning that the second time I call `browser.close()` it won't execute the callback.

If you can think of a way to unit test this change or if you agree to make the value of `[browser-runner]._tempDirectory` somehow accessible from a `browser` object, please let me know and I would be happy to make the necessary changes. Thanks.

@mathiasbynens

docs/api.md
```diff
@@ -882,6 +882,8 @@ a single instance of [BrowserContext].
 
 Closes Chromium and all of its pages (if any were opened). The [Browser] object itself is considered to be disposed and cannot be used anymore.
 
+During the process of closing the browser, it will attempt to delete the temp folder created exclusively for this browser instance. If this fails (either because a file in the temp folder is locked by another process or because of insufficient permissions) it will log the error in the console and throw back the exception. This implies that: a) the folder and/or its content will not be fully deleted; and b) the connection with the browser will not be properly disposed (see [browser.disconnect()](#browserdisconnect)).
```
```suggestion
During the process of closing the browser, Puppeteer attempts to delete the temp folder created exclusively for this browser instance. If this fails (either because a file in the temp folder is locked by another process or because of insufficient permissions) an error is logged. This implies that: a) the folder and/or its content is not fully deleted; and b) the connection with the browser is not properly disposed (see [browser.disconnect()](#browserdisconnect)).
```

mathiasbynens (approved):

alamothe:
I would like to report that, while this fix works (I prefer resolved vs rejected) and doesn't hang the browser, there is still a lot of instability caused by trying to remove the temporary directory. Our production service on Windows crashes ever so often due to that. Right now the way we patched this is to schedule (`setTimeout`) deleting temp folder after a minute instead of immediately during close. Will report if this fully resolves the problem.

PR #7343, opened 2021-06-21
feat: add page.emulateCPUThrottling

jschfflr:


jschfflr:
@OrKoN Do you have a suggestion for how to test this feature in a better way?

docs/api.md
```diff
@@ -1535,6 +1536,27 @@ const iPhone = puppeteer.devices['iPhone 6'];
 
 List of all available devices is available in the source code: [src/common/DeviceDescriptors.ts](https://github.com/puppeteer/puppeteer/blob/main/src/common/DeviceDescriptors.ts).
 
+#### page.emulateCPUThrottling(rate)
+
+- `rate` <?[number]> Rate at which the CPU will be throttled (2x, 3x, ...). Passing `null` disables cpu throttling.
```
mathiasbynens: Based on this description, would `factor` be a clearer name?

jschfflr: I don't have strong feelings about the naming. `rate` is used in the CDP definition - that's why I chose it here.
In Chromium, this gets converted into a percentage (2 -> 200%, 2.5 -> 250%) so we could also change the argument to be a percentage instead. LMKWYT!

src/common/Page.ts
```diff
@@ -1587,6 +1587,16 @@ export class Page extends EventEmitter {
     });
   }
 
+  async emulateCPUThrottling(rate: number | null): Promise<void> {
+    assert(
+      rate === null || rate >= 1,
+      'Throttling rate should be greater or equal to 1'
+    );
+    await this._client.send('Emulation.setCPUThrottlingRate', {
```
OrKoN, line 1595: Does it work without the emulation domain enabled?

test/emulation.spec.ts
```diff
@@ -408,4 +408,34 @@ describe('Emulation', () => {
       await page.emulateNetworkConditions(null);
     });
   });
+
+  describeFailsFirefox('Page.emulateCPUThrottling', function () {
```
OrKoN, line 412: This is potentially very flaky so I don't think we should really test the effect. Let's just test that a command is received by the backend successfully.

src/common/Page.ts
OrKoN: how do we handle errors?

test/emulation.spec.ts
```diff
@@ -408,4 +408,34 @@ describe('Emulation', () => {
...
+            i++;
+          }
+        });
+
+        const end = new Date();
+        return +end - +start;
+      }
+
+      const baseline = await measure();
+      await page.emulateCPUThrottling(100);
```
OrKoN, line 416: does it need to be reset in a finally block? if the test fails, I guess the throttling might apply to the next tests.

src/common/Page.ts
jschfflr: The emulation domain does not have `enable`/`disable` commands and the test works. So I think we are good here.
About the error: The [CDP command](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/inspector_emulation_agent.cc;l=323;bpv=1;bpt=1?q=setCPUThrottlingRate%20-out%2F) only fails if it's called on something else than a page. But since this is added to the `Page` class, we should be good to go. Let me know what you think!

test/emulation.spec.ts
jschfflr: SGTM!

test/emulation.spec.ts
jschfflr: Other tests for emulation features also don't reset afterwards. I think it's because the test runner will always create a new page for every test.

OrKoN (approved):

docs/api.md
mathiasbynens: I don't feel strongly either but I do think that `factor` is clearer/more self-documenting. Let's not do the percentages for now — it seems confusing IMHO.

mathiasbynens (approved):
LGTM

docs/api.md
jschfflr: I renamed it to factor for now :)

docs/api.md
```diff
@@ -1535,6 +1536,27 @@ const iPhone = puppeteer.devices['iPhone 6'];
 
 List of all available devices is available in the source code: [src/common/DeviceDescriptors.ts](https://github.com/puppeteer/puppeteer/blob/main/src/common/DeviceDescriptors.ts).
 
+#### page.emulateCPUThrottling(rate)
```
OrKoN: needs updating here too?

jschfflr: Yeah, you're absolutely right. Forgot to save the changes so git did get them -.-

PR #7349, opened 2021-06-22
feat: add support for web vitals

jschfflr:


docs/api.md
```diff
@@ -198,6 +200,7 @@
   * [page.waitForSelector(selector[, options])](#pagewaitforselectorselector-options)
   * [page.waitForTimeout(milliseconds)](#pagewaitfortimeoutmilliseconds)
   * [page.waitForXPath(xpath[, options])](#pagewaitforxpathxpath-options)
+  * [page.webvitals](#pagewebvitals)
```
mathiasbynens, line 204: This casing is inconsistent with the other API names used. To fix, use `webVitals`

docs/api.md
```diff
@@ -137,6 +137,7 @@
   * [page.cookies([...urls])](#pagecookiesurls)
   * [page.coverage](#pagecoverage)
   * [page.deleteCookie(...cookies)](#pagedeletecookiecookies)
+  * [page.disableWebVitalsReporting()](#pagedisablewebvitalsreporting)
```
mathiasbynens, line 141: This introduces a new pattern into Puppeteer for `.enableX()` and `.disableX()`. What’s the rationale behind this? (Is there a design doc?) Have we considered matching the `X.start()`/`X.stop()` precedent set by [the `Tracing` class](https://github.com/puppeteer/puppeteer/blob/v10.0.0/docs/api.md#class-tracing) instead?

docs/api.md
```diff
@@ -1498,6 +1501,12 @@ If URLs are specified, only cookies for those URLs are returned.
   - `path` <[string]>
 - returns: <[Promise]>
 
+#### page.disableWebVitalsReporting()
+
+- returns: <[Promise]>
+
+Disables the reporting of Web Vitals after a previous call to [page.enableWebVitalsReporting(enabledMetrics)](#pageenablewebvitalsreportingenabledmetrics).
```
mathiasbynens, line 1553:
```suggestion
Disables the reporting of Web Vitals after a previous call to [`page.enableWebVitalsReporting(enabledMetrics)`](#pageenablewebvitalsreportingenabledmetrics).
```

docs/api.md
```diff
@@ -1697,6 +1706,30 @@ const puppeteer = require('puppeteer');
...
+
+```js
+const puppeteer = require('puppeteer');
+
+(async () => {
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+  page.webvitals.on('LCP', ({value}) => {
+    console.log('LCP:', value);
+  })
```
mathiasbynens, line 1769: Instead of just `value`, this is an opportunity to show what the complete object looks like. Are there other properties than `value`? Also please double-check the code style (e.g. missing semicolon, periods after comments, etc.)

test/page.spec.ts
```diff
@@ -1745,4 +1745,21 @@ describe('Page', function () {
...
   });
+
+  describe('WebVitals', () => {
+    it('should report all relevant metrics', async () => {
+      const { page, server } = getTestState();
+
+      const events = [];
+      await page.webvitals.on('*', (metric) => {
+        events.push(metric);
+      });
```
mathiasbynens, line 1770: Based on the example above, I was expecting `metric` to be an object with a `value` property, but the assertion on line 1762 suggests that it’s just a string?

docs/api.md
```diff
@@ -1697,6 +1706,30 @@ const puppeteer = require('puppeteer');
...
+  await page.enableWebVitalsReporting();
+  await page.goto('https://v8.dev/blog/10-years');
+  // Some metrics will only be reported when navigating away from a page
+  await page.goto('about:blank');
+  await browser.close();
+})();
```
mathiasbynens, line 1775: This snippet calls `enable` but not `disable`, is that intended?

mathiasbynens (commented):
This feels like an important change which we should probably discuss in a design doc first!

Nevertheless I’ve already left some comments here.

jschfflr:
> This feels like an important change which we should probably discuss in a design doc first!
> 
> Nevertheless I’ve already left some comments here.

First of all, thanks for your feedback - I'll address it on the design doc.
I just wanted to prototype this to learn about the technical feasibility that might influence the design. 
This pull request was not meant to be in a landable shape yet but github has no way of making this private unfortunately :/

mathiasbynens:
> This pull request was not meant to be in a landable shape yet but github has no way of making this private unfortunately :/

I think the closest thing might be: https://docs.github.com/en/github/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-stage-of-a-pull-request#converting-a-pull-request-to-a-draft

jschfflr:
Awesome, thanks! I wasn't aware of that feature.

test/page.spec.ts
jschfflr: That is because of the '*' event listener. In that case, the event name is the first parameter. 
The object from the example would be the second argument.

docs/api.md
jschfflr: Following the pattern that we already established with the tracing is a great idea!

src/common/Page.ts
```diff
@@ -605,6 +612,62 @@ export class Page extends EventEmitter {
...
+    enabledMetrics?: string[]
+  ): Promise<void> {
+    if (!isNode) {
+      throw new Error(
+        'Cannot enable Web Vitals reporting outside of Node.js environment.'
+      );
+    }
+
+    const fs = await helper.importFSModule();
+    const webVitalsScript = await fs.promises.readFile(
```
jschfflr: @jackfranklin Do you have an idea how we could achieve this in a not Node.js specific way?

OrKoN:
@mathiasbynens do you think we should revisit this feature? I am closing it for now.

PR #7351, opened 2021-06-23
fix: remove redundant await while fetching target

yashLadha:
Removed the redundant await from the block while fetching the target.
Since the value of map is of instance `Target` which itself is not an
async resource.


mathiasbynens:
There’s some async stuff happening in the constructor, but nothing that `await` has any effect on AFAICT. There’s also no `then` method on the `Target` class (which would implicitly be called by the `await`). So it does seem like this `await` is indeed redundant. Good catch!

mathiasbynens (approved):

Issue #7350, opened 2021-06-22
Add support for the TTI Lifecycle Event


jschfflr:
Chromium [recently added](https://chromium-review.googlesource.com/c/chromium/src/+/2891445) a new livecycle event called TTI.
Puppeteer should also expose it.

mathiasbynens:
I'd love to see a holistic solution that allows us to expose TTI and other similar metrics in the future!

jschfflr:
I think this would be especially useful in the context of the `waitUntil` parameters to `frame.goto` for example.
We already have `load`, `domcontentloaded`, `networkidle0`, `networkidle2` and `tti` could be a great addition here.

Lightning00Blade:
Closing per comment on the PR - https://github.com/puppeteer/puppeteer/pull/7377#issuecomment-1040013711

PR #7377, opened 2021-06-24
feat: add 'interactive' to lifecycle events

jschfflr:
This has to wait for a new Chromium roll before it can be landed I think but maybe you can already take a look at it.
Closing #7350


OrKoN (approved):

mathiasbynens:
> This has to wait for a new Chromium roll before it can be landed

There’s a label for that(TM)!

jschfflr:
> > This has to wait for a new Chromium roll before it can be landed
> 
> There’s a label for that(TM)!

Ah nice, I didn't know about that :)

OrKoN:
The `interactive` lifecycle event is not firing for all pages and therefore might be never reached. Let's close this PR for now unless we have a good use case for this event.

PR #7389, opened 2021-06-29
feat: add channel parameter for puppeteer.launch

### LINKED ISSUE #7340: Feature Request: 'channel' parameter like Playwright

* Puppeteer version: 10.0.0
* Platform / OS version: macOS
* Node.js version: 14.15.1

---

#### Summary

When we want to use existing browser (Chrome or Edge), we have to specify `executablePath`.
I know it is very simple, however it would be more useful if it is specified like `chrome` or `chrome-canary` as Playwright does, because executablePath is almost not changing. Currently macOS users always have to input `"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"` or `"/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary"`.

#### Background

In our organization, Puppeteer is used for enjoying the browser automation for all members . Most Windows users says 'How to use already installed MS Edge?' and copy-paste the Edge's path, however some members fails to copy it ... :(
It would reduce the mistake to specify the browser name with `channel` (as Playwright) or something.

#### ref

* https://playwright.dev/docs/browsers#google-chrome--microsoft-edge
* https://github.com/microsoft/playwright/blob/ea4eebeb2d1f5984e9dee504aa7501349ad762cc/src/server/chromium/findChromiumChannel.ts#L62

YusukeIwaki:
Add 'channel' parameter for Puppeteer.launch

Resolves #7340

Google Chrome is often used for using puppeteer as a browser automation tool, 
We have to specify executablePath in such case. However we can't share the code among Windows/macOS users, because executablePath is different. 

microsoft/playwright already introduced the `channel` parameter for solving this problem.
Let's provide the similar functionality for Puppeteer users.

```js

puppeteer.launch({ channel: 'chrome-canary', headless: false }).then(async (browser) => {
    await page.goto('https://github.com/puppeteer');
    await browser.close();
});
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -454,7 +454,7 @@ When using `puppeteer-core`, remember to change the _include_ line:
 const puppeteer = require('puppeteer-core');
 ```
 
-You will then need to call [`puppeteer.connect([options])`](#puppeteerconnectoptions) or [`puppeteer.launch([options])`](#puppeteerlaunchoptions) with an explicit `executablePath` option.
+You will then need to call [`puppeteer.connect([options])`](#puppeteerconnectoptions) or [`puppeteer.launch([options])`](#puppeteerlaunchoptions) with an explicit `executablePath` or `channel` option.
 
 ### Environment Variables
 
@@ -627,6 +627,7 @@ try {
   - `product` <[string]> Which browser to launch. At this time, this is either `chrome` or `firefox`. See also `PUPPETEER_PRODUCT`.
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
   - `headless` <[boolean]> Whether to run browser in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). Defaults to `true` unless the `devtools` option is `true`.
+  - `channel` <[string]> When specified, Puppeteer will search for the locally installed release channel of Google Chrome and use it to launch. Available values are `chrome`, `chrome-beta`, `chrome-canary`, `chrome-dev`. When channel is specified, `executablePath` cannot be specified.
   - `executablePath` <[string]> Path to a browser executable to run instead of the bundled Chromium. If `executablePath` is a relative path, then it is resolved relative to [current working directory](https://nodejs.org/api/process.html#process_process_cwd). **BEWARE**: Puppeteer is only [guaranteed to work](https://github.com/puppeteer/puppeteer/#q-why-doesnt-puppeteer-vxxx-work-with-chromium-vyyy) with the bundled Chromium, use at your own risk.
   - `slowMo` <[number]> Slows down Puppeteer operations by the specified amount of milliseconds. Useful so that you can see what is going on.
   - `defaultViewport` <?[Object]> Sets a consistent viewport for each page. Defaults to an 800x600 viewport. `null` disables the default viewport.
@@ -660,7 +661,7 @@ const browser = await puppeteer.launch({
 });
 ```
 
-> **NOTE** Puppeteer can also be used to control the Chrome browser, but it works best with the version of Chromium it is bundled with. There is no guarantee it will work with any other version. Use `executablePath` option with extreme caution.
+> **NOTE** Puppeteer can also be used to control the Chrome browser, but it works best with the version of Chromium it is bundled with. There is no guarantee it will work with any other version. Use `executablePath` or `channel` option with extreme caution.
 >
 > If Google Chrome (rather than Chromium) is preferred, a [Chrome Canary](https://www.google.com/chrome/browser/canary.html) or [Dev Channel](https://www.chromium.org/getting-involved/dev-channel) build is suggested.
 >
--- a/src/node/Launcher.ts
+++ b/src/node/Launcher.ts
@@ -17,6 +17,7 @@ import * as os from 'os';
 import * as path from 'path';
 import * as fs from 'fs';
 
+import { assert } from '../common/assert.js';
 import { BrowserFetcher } from './BrowserFetcher.js';
 import { Browser } from '../common/Browser.js';
 import { BrowserRunner } from './BrowserRunner.js';
@@ -109,6 +110,12 @@ class ChromeLauncher implements ProductLauncher {
     let chromeExecutable = executablePath;
 
     if (channel) {
+      // executablePath is detected by channel, so it should not be specified by user.
+      assert(
+        !executablePath,
+        '`executablePath` must not be specified when `channel` is given.'
+      );
+
       chromeExecutable = executablePathForChannel(channel);
     } else if (!executablePath) {
       // Use Intel x86 builds on Apple M1 until native macOS arm64
--- a/src/node/Launcher.ts
+++ b/src/node/Launcher.ts
@@ -598,43 +598,75 @@ class FirefoxLauncher implements ProductLauncher {
 function executablePathForChannel(channel: ChromeReleaseChannel): string {
   const platform = os.platform();
 
+  let chromePath: string | undefined;
   switch (platform) {
     case 'win32':
       switch (channel) {
         case 'chrome':
-          return `${process.env.PROGRAMFILES}\\Google\\Chrome\\Application\\chrome.exe`;
+          chromePath = `${process.env.PROGRAMFILES}\\Google\\Chrome\\Application\\chrome.exe`;
+          break;
         case 'chrome-beta':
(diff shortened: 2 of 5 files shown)


YusukeIwaki:
@googlebot I signed it!

test/launcher.spec.ts
```diff
@@ -268,7 +268,7 @@ describe('Launcher specs', function () {
         expect(await page2.evaluate(() => localStorage.hey)).toBe('hello');
         await browser2.close();
         // This might throw. See https://github.com/puppeteer/puppeteer/issues/2778
-        await rmAsync(userDataDir).catch(() => {});
+        await rmAsync(userDataDir).catch(() => { });
```
jschfflr: It looks like there are a bunch of drive by changes, can you run `npm run eslint-fix`?

src/node/Launcher.ts
```diff
@@ -587,6 +595,35 @@ class FirefoxLauncher implements ProductLauncher {
   }
 }
 
+function executablePathForChannel(channel: ChromeReleaseChannel): string {
+  const platform = os.platform();
```
jschfflr, line 607: WDYT about checking if the path that we assumed here exists and if not throw an error specific to this situation?

jschfflr (commented):
The change looks good to me, thanks for your contribution!
I left some comments and maybe we should also think about whether `PUPPETEER_EXECUTABLE_PATH ` should take precedence over the channel or not.
Let me know what you think!

test/launcher.spec.ts
YusukeIwaki: Oh, my mistake, I will fix it. Thank you.

src/node/Launcher.ts
YusukeIwaki: The check should be added here, showing an error like 'Browser is not installed'.

YusukeIwaki:
Thank you for your kind review. I considered again about the priority of channel and executablePath.

* [1] executablePath  should be ignored when channel is specified
* [2] executablePath should be respected even when channel is specified

[1] is actually adopted in Playwright. This is really simple. I think [1] is better.

The use-case of [2] is like below:
When some users changed the install path of Chrome,  `channel: chrome` fails to detect Chrome. Without modifying the script source code, users can set the install path with `PUPPETEER_EXECUTABLE_PATH=/path/to/chrome`

However in [2], users can also specify MS Edge as executablePath even when `channel: chrome`.
If the script developer intends to the script is assumed to work with Chrome, PUPPETEER_EXECUTABLE_PATH=/path/to/msedge  is just a source of worry.

This is just my observation but I didn't see the people who change the Chrome install path, and I thought [2] is useful for only a few people. IMO, simplicity is better than considering the edge case.
So I suggedt the spec [1] executablePath  should be ignored when channel is specified

src/node/Launcher.ts
```diff
@@ -587,6 +596,78 @@ class FirefoxLauncher implements ProductLauncher {
...
+    throw new Error(
+      `Unable to detect browser executable path for '${channel}' on ${platform}.`
+    );
+  }
+
+  // Check if Chrome exists and is accessible.
+  try {
+    fs.accessSync(chromePath);
+  } catch (error) {
+    throw new Error(`Chrome is not installed: ${channel}`);
```
jschfflr:
```suggestion
    throw new Error(`Could not find Google Chrome executable for channel '${channel}' at '${chromePath}'.`);
```

jschfflr (changes requested):
Looks good to me from a technical point of view now.
But please also update the documentation in `docs/api.md` to explain the change.

jschfflr:
Regarding the priority: Let's make them mutually exclusive.
If one of them is specified, the other one can not be set - that way, people don't run into unexpected behaviour.
Can you add an `assert` for that?

YusukeIwaki:
> Regarding the priority: Let's make them mutually exclusive.
> If one of them is specified, the other one can not be set - that way, people don't run into unexpected behaviour.

That's a good idea. I will add assertion for it.

docs/api.md
```diff
@@ -620,6 +620,7 @@ try {
   - `product` <[string]> Which browser to launch. At this time, this is either `chrome` or `firefox`. See also `PUPPETEER_PRODUCT`.
   - `ignoreHTTPSErrors` <[boolean]> Whether to ignore HTTPS errors during navigation. Defaults to `false`.
   - `headless` <[boolean]> Whether to run browser in [headless mode](https://developers.google.com/web/updates/2017/04/headless-chrome). Defaults to `true` unless the `devtools` option is `true`.
+  - `channel` <[string]> Release channel of Google Chrome. Available values are `chrome`, `chrome-beta`, `chrome-canary`, `chrome-dev`. When channel is specified, executablePath cannot be specified.
```
jschfflr:
```suggestion
  - `channel` <[string]> When specified, Puppeteer will search for the locally installed release channel of Google Chrome and use it to launch. Available values are `chrome`, `chrome-beta`, `chrome-canary`, `chrome-dev`. When channel is specified, `executablePath` cannot be specified.
```

src/node/Launcher.ts
```diff
@@ -105,7 +108,16 @@ class ChromeLauncher implements ProductLauncher {
     }
 
     let chromeExecutable = executablePath;
-    if (!executablePath) {
+
+    if (channel) {
+      // executablePath is detected by channel, so it should not be specified by user.
+      assert(
+        !executablePath,
```
jschfflr, line 115:
```suggestion
        !(executablePath ||
      process.env.PUPPETEER_EXECUTABLE_PATH ||
      process.env.npm_config_puppeteer_executable_path ||
      process.env.npm_package_config_puppeteer_executable_path),
```

YusukeIwaki: Oh, thank you for your suggestion. I understood this assertion is just for checking invalid and ambiguous usage of the parameters (both executablePath and channel are specified), and I thought each environment-based variables are just ignored when channel is specified.

Also I think warning(or info logging) is enough to notify users to know PUPPETEER_EXECUTABLE_PATH or other variables are ignored.
I don't think it is invalid to specify channel when PUPPETEER_EXECUTABLE_PATH is set as a default executable path.
How do you think?

jschfflr: Yes, while I think it would be super helpful to make sure that the parameters can not be used in an invalid and ambiguous way I feel like the same should also apply to the environment variables. 
Otherwise, people might be setting them and then wonder why they are just silently ignored...
Do you see any downsides to including them here?

YusukeIwaki: This is just my concern but I worry about the use-case to share the code with `channel: chrome` in team members.
Script developer assumes Chrome is launched by setting `channel: chrome`, but if some members set `export PUPPETEER_EXECUTABLE_PATH=/path/to/chrome` in .bashrc or another, they claims "script is not working!!" by the assertion error. 

> I feel like the same should also apply to the environment variables.
> Otherwise, people might be setting them and then wonder why they are just silently ignored...

I agree this behavior is really simple, however it is a little confusing in terms of consistency.
PUPPETEER_EXECUTABLE_PATH is already ignored when `executablePath` is explicitly specified, and some users would regard the environment variable as a definition of the *default value* of executablePath.


I know I worry a lot, so if simple behavior is preferred, I will add the assertion for environment variables.

jschfflr: You have a good point about `PUPPETEER_EXECUTABLE_PATH` being used as a fall back.
Let's keep the behaviour the way you suggested!

YusukeIwaki:
@jschfflr Thank you for your kind review.
Could you re-run GitHub actions? The failed test is caused by flaky testcase.

jschfflr:
Sure :)

YusukeIwaki:
Rebased.

jschfflr (approved):
The change LGTM now!

jschfflr:
Your change landed 🎉
Thanks for your contribution @YusukeIwaki!

noinkling:
Any particular reason this feature only checks `Program Files` and not also `Program Files (x86)` on Windows? My Chrome installation is located in the latter. According to [this](https://www.ghacks.net/2020/06/11/google-chrome-is-soon-going-to-be-installed-in-a-different-directory-on-windows/) it was the default location until 2020. For installations existing before that point, it seems like the only (non-hacky) way to move it is to uninstall and reinstall.

Playwright appears to look in both (as well as in `AppData/Local`, which I seem to remember was indeed where the executable lived at one point), if that's any motivation: https://github.com/microsoft/playwright/blob/b6eb917c29b743e8d63004ae7a24974404461198/packages/playwright-core/src/utils/registry.ts#L461

PR #7394, opened 2021-07-02
fix: added script

tasneemkoushar:
Added script `remove-tag.js` to remove unsupported syntax `<!-- -->` from all `.md` files


utils/remove-tag.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+const glob = require('glob');
+const fs = require('fs');
+
+// look for all .md files in the given location.
+const files = glob.sync(`./new-docs/*.md`);
+
+for (const file of files) {
+  const content = fs.readFileSync(file, 'utf8');
+  const updated = content.replace(/<!-- -->/g, '');
```
jackfranklin, line 30: please could we add a comment explaining why we do this?

```
/**
 * Docusaurus gets confused by empty HTML comment tags in our Markdown. These are generated by API Extractor but do not affect the final output, so to avoid Docusaurus errors we strip them out before generating the website.
 **/

**Review by jackfranklin — APPROVED** [review]

File: utils/remove-tag.js [hunk 2]
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env node
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
```
mathiasbynens, line 3: 2021 :)

Issue #6946, opened 2021-03-01
Cleanup errors in browser.close hang indefinitely


SnirBroshi:
In `browser.close()` puppeteer tries to rimraf the user-data-dir. This often fails on Windows due to EPERM (#298).
If `rimraf` fails here, puppeteer logs the error to the console, and then the `browser.close()` promise is left hanging forever, not resolved or rejected:

https://github.com/puppeteer/puppeteer/blob/0c26301bee0d07f314831fc8f1f4e5a1d31e5e9d/src/node/BrowserRunner.ts#L101-L103

the `catch` callback is the problem, it should resolve or reject the promise.

Also it seems like node somehow detects that there's no code left to run, since we're waiting for a promise that can't resolve, and kills the process, so any further code is never ran.

(the repro is irrelevant here since I'm talking about any potential error that `rimraf` can throw, but #298 is one such repro)

* Puppeteer version: 8.0.0
* Platform / OS version: Windows 10 64bit
* Node.js version: 12.18.1

alamothe:
It's a pretty big problem that `browser.close` just hangs. If there is an important error, it should throw. If it's unimportant, it should ignore it (since it's closing). 

How big of an issue is that these files do not get removed?

alamothe:
Just patched `puppeteer` and can confirm it fixes the issue. I just fulfilled the promise in the `catch` block

tx46:
This is still happening to me. I'm trying to work around it but after a few shots with Puppeteer over the course of a few days, I have to reboot the system because I either have to a) block forever in my code because Puppeteer will never exit (at least according to the promise) or b) leave dangling instances of Puppeteer running.

PR #7396, opened 2021-07-02
FIX: Fulfill promise in removeFolderAsync() - solves EPERM crash while deleting temp folder

bdumitru:
Fix for: #6946, #298 etc. (MAJOR PRIORITY)

Problem description:
WINDOWS ONLY!
Any process involving this library would completely crash ungracefully if browser.close() is unable to delete the /temp/ folder, because the promise never gets fulfilled or rejected.
Problem tends to happen on slower computers/VMs or environments with limited access to appdata/local folder.
Issue can be hard to reproduce in a lot of cases.

Solution:
fulfill() in catch for removeFolderAsync(this._tempDirectory)
Alternative solutions:
- SIMPLE, but not ideal: throw(error) instead, remove console.error(error).
- HARDER, ideal: add WORKING runtime arguments to disable Crashpad-Metrics.


bdumitru:
@googlebot I signed it!

jschfflr:
It looks like this has already been addressed in https://github.com/puppeteer/puppeteer/pull/7338

PR #7409, opened 2021-07-06
Added comments for some methods

tasneemkoushar:


src/common/Connection.ts
```diff
@@ -263,6 +263,9 @@ export class CDPSession extends EventEmitter {
     this._sessionId = sessionId;
   }
 
+  /**
+   * @internal
+   */
```
jackfranklin: I don't think this is internal, as it's documented here: https://pptr.dev/#?product=Puppeteer&version=v10.1.0&show=api-cdpsessionconnection

Could we update this PR to add documentation for this method, please?

src/common/Page.ts
```diff
@@ -2438,26 +2439,26 @@ export class Page extends EventEmitter {
...
+   *  - `width` : <number> width of clipping area.</number><br/>
+   *  - `height` : <number> height of clipping area.</number>
    *
    * - `omitBackground` : <boolean> Hides default white background and allows
-   *   capturing screenshots with transparency. Defaults to `false`
+   *   capturing screenshots with transparency. Defaults to `false`</boolean>
    *
    * - `encoding` : <string> The encoding of the image, can be either base64 or
-   *   binary. Defaults to `binary`.
+   *   binary. Defaults to `binary`.</string>
```
jackfranklin: all these closing tags are incorrect - but I think what we should do is get rid of the `<string>` / `<number>` part of the docs entirely.

jackfranklin (commented):
Thanks - left some comments!

src/common/Connection.ts
tasneemkoushar: Sure, I will mark it as public

src/common/Page.ts
tasneemkoushar: Ok I will omit these invalid tags

src/common/Connection.ts
```diff
@@ -48,7 +48,7 @@ export const ConnectionEmittedEvents = {
 } as const;
 
 /**
- * @internal
+ * @public
```
jackfranklin, line 51: we should also add TSDoc comments to the methods on this class that are documented on pptr.dev (https://pptr.dev/#?product=Puppeteer&version=v10.1.0&show=api-cdpsessionconnection). Could you look at that in a future PR, please?

jackfranklin (approved):
Thanks!

PR #7413, opened 2021-07-07
fix: use non-deprecated tracing categories api

Janpot:
Noticed that `page.tracing.start` didn't respect the `categories` option. While investigating I noticed that I actually needed to supply `-*` as well in `categories` to exclude other events, but in the meantime I had already converted `Tracing.start` to use [`traceConfig`](https://chromedevtools.github.io/devtools-protocol/tot/Tracing/#type-TraceConfig) over the [deprecated `categories`](https://chromedevtools.github.io/devtools-protocol/tot/Tracing/#method-start). Also added an extra test to verify events are actually included or not, besides checking the `trace-config` from the result.


jschfflr (approved):
Looks good to me, thanks!

Issue #7407, opened 2021-07-06
[bug] Connection not closed correctly in v10.1.0


DRoet:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 10.1.0
* Platform / OS version: Docker/Linux
* URLs (if applicable):
* Node.js version: 16.3.0

**What steps will reproduce the problem?**
When updating from 10.0.0 -> 10.1.0 I suddenly get an error when connecting to a remote chrome instance using `browser.connect()`. The remote chrome instance I am using is a simple `browserless/chrome` docker image running on a dedicated port

_Please include code that reproduces the issue._

1. I am using the following code to connect to a remote chrome instance and create a PDF
```js
   const browser = await puppeteer.connect({
      browserWSEndpoint: 'http://chromeinstance:port/',
   })
   const page = await browser.newPage()
   await page.setContent(html, { waitUntil: 'networkidle2' })
   await page.pdf({ format: format, landscape: landscape })
   await browser.disconnect()
```

**What is the expected result?**
Connection closes correctly, no error is being shown. This works fine in v10.0.0

**What happens instead?**
following error is being thrown:
```
/opt/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/Connection.js:217
            this._callbacks.set(id, { resolve, reject, error: new Error(), method });
                                                              ^

Error: Protocol error (IO.close): Target closed.
    at /opt/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/Connection.js:217:63
    at new Promise (<anonymous>)
    at CDPSession.send (/opt/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/Connection.js:216:16)
    at Readable.read [as _read] (/opt/app/node_modules/puppeteer-core/lib/cjs/puppeteer/common/helper.js:281:30)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```

4geru:
our project caused same error which version up from 10.0.0 to 10.1.0.

nadaDeversity:
@DRoet Do you have any updates on this? I get the same error on `await browser.close()` (did not use `puppeteer.connect` but `puppeteer.launch`)

DRoet:
Nope, best to stay on 10.0.0 for now, there is a PR up trying to fix the issue https://github.com/puppeteer/puppeteer/pull/7414 but there isn't much traction yet.

simonhaenisch:
[...]

PR #7414, opened 2021-07-08
Add await method about issue#7407

zorori777:
## Problem
[issue](https://github.com/puppeteer/puppeteer/issues/7407)

## Analysis   
When I updated this pakcage from 10.0.0 to 10.1.0, suddenly happened this error.
So I checked each differences in change log between 10.0.0 to 10.1.0.
 [ChangeLog](https://github.com/puppeteer/puppeteer/blob/main/CHANGELOG.md#1010-2021-06-29)

## Reason
[#7351](https://github.com/puppeteer/puppeteer/pull/7351)
It  shouldn't  remove await method.

This pull request's approval comments say `There’s some async stuff happening in the constructor`, in the fact constructor call async method. so it need to use await function.

According to error message, when synchronous processing finished, closed entire process at the same time.
Need to wait for asynchronous processing, so I add await method.

## Fix
I add await method, removed it by [#7351](https://github.com/puppeteer/puppeteer/pull/7351)


src/common/Browser.ts
```diff
@@ -433,7 +433,7 @@ export class Browser extends EventEmitter {
       url: 'about:blank',
       browserContextId: contextId || undefined,
     });
-    const target = this._targets.get(targetId);
+    const target = await this._targets.get(targetId);
```
simonhaenisch, line 436: Not sure whether you've looked into the source code but `this._targets` is a map of `Target` instances. The assignment happens here: 

https://github.com/puppeteer/puppeteer/blob/44b22bbc2629e3c75c1494b299a66790b371fb0a/src/common/Browser.ts#L350-L361

The target's constructor might be doing some async work, but you can't mark a constructor as `async` and await the class instantiation (i. e. `await new Target(...)` is not possible). The instance will be immediately returned and the constructor's async work happens afterwards in the background.

Did you actually try building puppeteer with this change to see whether this fixes the issue for you?

---

Edit: the only thing I could imagine is that adding the `await` would defer this line to be executed later (i. e. pushes it to the end of the event loop). I just did a search and it seems like that's how it works according to [this SO answer](https://stackoverflow.com/a/55263084/2897426) (_"if you await things you will always defer the continuation to a microtask"_). TIL 🤓 

@mathiasbynens referring to https://github.com/puppeteer/puppeteer/pull/7351#issuecomment-866555246, is it possible that the `await` deferring did actually have an effect?

jschfflr:
I feel like this patch might not really be fixing the problem because the next line is already awaiting `target._initializedPromise`. The original problem is probably timing related and therefore flaky.
Could you verify that this change is actually fixing the problem?

jschfflr:
@zorori777 I'll close this for now as there hasn't been an update on this issue. Feel free to reopen if you think this is still relevant!

PR #7434, opened 2021-07-20
chore(docs): setup placeholder workflow

jackfranklin:


mathiasbynens (approved):

.github/workflows/deploy-docs.yml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Build
+        run: |
+          npm install
+          # This will be replaced with the actual documentation builder once it's
+          # configured
+          mkdir -p docs-dist/
```
mathiasbynens, line 18: As discussed offline, depending on how complex/maintenance-heavy the doc builder becomes, we might want to move it to a separate repo in the future. Then in this action we could just `npm install` that repo. But starting simple with everything in the main repo SGTM 👍

jackfranklin: Agreed 👍

PR #7440, opened 2021-07-26
updated contributing.md

tasneemkoushar:


CONTRIBUTING.md
```diff
@@ -171,6 +171,16 @@ To format the documentation markdown and its code snippets, use:
 npm run markdownlint-fix
 ```
 
+## Writing TSDoc Comments
+
+When commenting, consider the following:
+
+- If there is a new release which hasn't been given any tag and its throwing some warning then mark that as `@public`
```
jackfranklin, line 178: I think we could change this to something like "Every new method needs to have either `@public` or `@internal` added depending on if it is part of the public API".

CONTRIBUTING.md
```diff
@@ -171,6 +171,16 @@ To format the documentation markdown and its code snippets, use:
...
+- Avoid use of HTML tags.
```
jackfranklin, line 179: Let's change this to "Use Markdown syntax when writing documentation".

CONTRIBUTING.md
```diff
@@ -171,6 +171,16 @@ To format the documentation markdown and its code snippets, use:
...
+- If column size comment is exceeding 90 try to use multiple lines instead of placing all in one line.
```
jackfranklin, line 180: change "try to use" to "please split it over"

CONTRIBUTING.md
```diff
@@ -171,6 +171,16 @@ To format the documentation markdown and its code snippets, use:
...
+- Prefixing your comments with `FIXME` or `TODO` helps other developers quickly understand if you’re pointing out a problem that needs to be revisited, or if you’re suggesting a solution to the problem that needs to be implemented.
+- For better insight of TSDoc syntax please refer There's a [API Extractor](https://api-extractor.com/pages/tsdoc/doc_comment_syntax/)
```
jackfranklin, line 182: "please refer to the [API Extractor](...) documentation."

CONTRIBUTING.md
tasneemkoushar: Sure, I will replace it

PR #7446, opened 2021-07-29
chore: update documentation

jschfflr:
Explain the relationship between `page.setOfflineMode` and `page.emulateNetworkConditions`.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -1704,7 +1704,8 @@ await page.evaluate(() => matchMedia('print').matches);
   - `latency` <[number]> Latency (ms), `0` to disable
 - returns: <[Promise]>
 
-> **NOTE** This does not affect WebSockets and WebRTC PeerConnections (see https://crbug.com/563644)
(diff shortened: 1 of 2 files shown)


docs/api.md
```diff
@@ -1704,7 +1704,8 @@ await page.evaluate(() => matchMedia('print').matches);
   - `latency` <[number]> Latency (ms), `0` to disable
 - returns: <[Promise]>
 
-> **NOTE** This does not affect WebSockets and WebRTC PeerConnections (see https://crbug.com/563644)
+> **NOTE** This does not affect WebSockets and WebRTC PeerConnections (see https://crbug.com/563644). To set the page offline, you can use [page.setOfflineMode(enabled)](#pagesetofflinemodeenabled).
```
OrKoN, line 1709: This part is not auto-generated right?

jschfflr: No, I added it manually. Is `api.md` now generated from code?

OrKoN: I am not sure, I thought so but I might be wrong :D

OrKoN (approved):

docs/api.md
jschfflr: The same strings can be found in `Page.ts` - I updated them too to keep them in sync.
@mathiasbynens Do we already generate documentation from source automatically?

PR #7464, opened 2021-08-03
chore(docs): stop generating devtools-protocol docs

jackfranklin:
This CL changes our docs generation to not include automatic docs from
devtools-protocol. Long term we probably want this, but for now it's
generating a vast amount of documentation and it's making setting up the
new website and docs harder. Let's focus just on the pptr docs for now
and revisit this once the foundations have been laid.


mathiasbynens (approved):
Alright then.

PR #7476, opened 2021-08-06
Added version 10.0.0

tasneemkoushar:


jackfranklin (approved):
Looks good! I think we'll have to figure out how to automate some of this in the future but for now let's ship like this and figure out a better approach.

Issue #7496, opened 2021-08-12
Inspector.targetCrashed with custom fonts on footerTemplate in createPdf


Vitaliy-Svinchyak:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 10.2.0
* Platform / OS version: macos / linux
* URLs (if applicable):
* Node.js version: 14

**What steps will reproduce the problem?**
1. Pass custom fonts to footerTemplate
2. Get `Inspector.targetCrashed`

Code: 
```
async function getBrowser(): Promise<Browser> {
  if (browserSingleton) {
    return browserSingleton
  }
  if (browserPromise) {
    return browserPromise
  }

  browserPromise = puppeteer.launch({
    pipe: true,
    dumpio: true,
    timeout: 60000,
    args: [
      '--headless',
      '--no-sandbox',
      '--disable-setuid-sandbox',
      '--disable-dev-shm-usage',
      '--full-memory-crash-report',
    ]
  }).then((browser: Browser) => {
    browserSingleton = browser
    browserPromise = undefined

    browserSingleton.on('disconnected', () => {
      console.log('Browser disconnected')
      browserSingleton = undefined
    })

    return browserSingleton
  })

  return browserPromise!
}

export default async function convert(header: string, html: string, footer: string, options: ConvertOptions): Promise<Readable> {
  const browser = await getBrowser()
  const page = await browser.newPage()
  page.on('console', consoleObj => console.log(consoleObj.text()))

  page.on('error', (e) => {
    console.error("Some error", e)
    throw e
  })
  page.on('pageerror', (e) => {
    console.error("Some page error", e)
    throw e
  })
  await page.setContent(html)
  console.log('rendering pdf')

  const stream: Readable = await page.createPDFStream({
    scale: options.scale,
    printBackground: options.printBackground,
    landscape: options.landscape,
    pageRanges: options.pageRanges,
    format: options.format,
    width: options.width,
    height: options.height,
    preferCSSPageSize: options.preferCSSPageSize,
    margin: options.margin,
    omitBackground: options.omitBackground,
    headerTemplate: header,
    footerTemplate: footer,
    displayHeaderFooter: true
  })
  stream.on("end", () => {
    console.log('rendered')

    page.close()
  })

  return stream
}
```

footerTemplate:
```
 <div style="font-family: Style Script, sans-serif;">
       footer
    </div>
    <style>
      @font-face {
        font-family: 'Style Script';
        font-style: normal;
[...]

Issue #6834, opened 2021-02-08
PDF rendering hangs unless page contains the same custom font as header


erikbrannstrom:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 7.0.1
* Platform / OS version: macOS 11.1 (Big Sur)
* Node.js version: 12.18.3

I initially got this problem in a Ruby app on Heroku (Ubuntu 16) using the [grover](https://github.com/Studiosity/grover) gem which in turn used puppeteer 5.5.0, but I have repeated the issue and identified a smaller version of it using puppeteer directly on the environment described above.

**What steps will reproduce the problem?**

The following script is used to initialize Puppeteer and generate the PDF. It is a simplified version of the code used by grover.

```js
const puppeteer = require("puppeteer");
const fs = require("fs");

const renderPDF = async (html, options) => {
  let browser;
  try {
    // Launch the browser and create a page
    browser = await puppeteer.launch({
      args: ["--no-sandbox", "--disable-setuid-sandbox"],
    });
    const page = await browser.newPage();

    // Request is some HTML content. Use request interception to assign the body
    await page.setRequestInterception(true);
    page.once("request", (request) => {
      request.respond({ body: html });
      // Reset the request interception
      // (we only want to intercept the first request - ie our HTML)
      page.on("request", (request) => request.continue());
    });

    await page.goto("http://example.com", { waitUntil: "networkidle0" });

    return await page.pdf({
      path: options.path,
      displayHeaderFooter: true,
      headerTemplate: options.headerTemplate,
      footerTemplate: "<html><head></head><body></body></html>",
      margin: { top: "2.9cm", bottom: "2.5cm", left: "1cm", right: "1cm" },
    });
  } finally {
    if (browser) {
      await browser.close();
    }
  }
};
```

The following is the HTML used for the page content (I have shortened the data URLs for readability).

```html
<!DOCTYPE html>
<html>
  <head>
    <meta http-equiv="content-type" content="text/html; charset=utf-8" />
    <style>
      @font-face {
        font-family: MyCustomFont;
        src: url(data:application/font-woff2;base64,d09GMgABAAAAAGhkABEAAAABG8gAA...)
          format("woff2");
        font-weight: 300;
      }

      @font-face {
        font-family: MyCustomFont;
        src: url(data:application/font-woff2;base64,d09GMgABAAAAAGvgABEAAAABKOwA...)
          format("woff2");
        font-weight: 500;
      }

      body {
[...]

Issue #5846, opened 2020-05-11
Puppeteer hangs after page crash due to OOM


andysnowden:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 2.1.0, have also tested 3.0.4
* Platform / OS version: docker container on k8s, Debian 4.9.189-3+deb9u2
* URLs (if applicable): N/A (private url)
* Node.js version: v14.0.0 (also tried 12.16.3 and 14.2.0)

**What steps will reproduce the problem?**

1.  Attempt to load a rather large page including react/images/etc for the purposes of generating a PDF.
2.  OOM kill the underlying chromium process before the request completes

**What is the expected result?**

Puppeteer/Node should exit and a error should be presented to the user.

**What happens instead?**

Puppeteer/Node hangs forever after a error is presented.

```
(node:376) UnhandledPromiseRejectionWarning: Error: Page crashed!
    at Page._onTargetCrashed (/usr/local/datapipeline/node/node_modules/puppeteer/lib/Page.js:213:24)
    at CDPSession.<anonymous> (/usr/local/datapipeline/node/node_modules/puppeteer/lib/Page.js:122:56)
    at CDPSession.emit (events.js:315:20)
    at CDPSession._onMessage (/usr/local/datapipeline/node/node_modules/puppeteer/lib/Connection.js:200:12)
    at Connection._onMessage (/usr/local/datapipeline/node/node_modules/puppeteer/lib/Connection.js:112:17)
    at WebSocket.<anonymous> (/usr/local/datapipeline/node/node_modules/puppeteer/lib/WebSocketTransport.js:44:24)
    at WebSocket.onMessage (/usr/local/datapipeline/node/node_modules/ws/lib/event-target.js:120:16)
    at WebSocket.emit (events.js:315:20)
    at Receiver.receiverOnMessage (/usr/local/datapipeline/node/node_modules/ws/lib/websocket.js:789:20)
    at Receiver.emit (events.js:315:20)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:376) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
[...]

PR #7508, opened 2021-08-19
feat: support timeout for page.pdf() call

hottomato-c:
# Context
When use `page.pdf` API, there are some extreme scenarios that can cause the promise never get fulfilled or rejected. Hence this may cause potential problems. These kind of errors most likely happens in Chromium.

## Related Issues
- https://github.com/puppeteer/puppeteer/issues/6834
- https://github.com/puppeteer/puppeteer/issues/7496
- https://github.com/puppeteer/puppeteer/issues/5846

# Changes
- Added `timeout` parameter for PDFOption which will be used in `page.pdf` call


hottomato-c:
> @googlebot I signed it!

@googlebot I signed it!

jschfflr (changes requested):
Can you move to timeout to `createPDFStream` as it looks like the `Page.printToPDF` call there is the original source of the timeout?

jschfflr:
Do you know if there's a corresponding bug reported for Chromium already? And if not, file one on https://crbug.com/new?
That way, we can make sure that we can also address the problem at its source. Thanks!

hottomato-c:
> Can you move to timeout to `createPDFStream` as it looks like the `Page.printToPDF` call there is the original source of the timeout?

Sounds good ! Thanks for the review

hottomato-c:
> Do you know if there's a corresponding bug reported for Chromium already? And if not, file one on https://crbug.com/new?
> That way, we can make sure that we can also address the problem at its source. Thanks!

np, will check on that .

hottomato-c:
☑️  [Issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1248727) created for Chromium

jschfflr:
Awesome, thanks! 
As we wanna make sure that the timeout still works when the bug is eventually fixed in Chromium, the tests should be updated to just use a timeout of `1` and make sure that the call times out. That's how we test the other time outs too. 
As for the test that you already wrote: It's a great reproduction that you can maybe put in a gist and reference it in the Chromium bug. That will help the engineer who'll be resolving the issue there a ton!

hottomato-c:
> Awesome, thanks!
> As we wanna make sure that the timeout still works when the bug is eventually fixed in Chromium, the tests should be updated to just use a timeout of `1` and make sure that the call times out. That's how we test the other time outs too.
> As for the test that you already wrote: It's a great reproduction that you can maybe put in a gist and reference it in the Chromium bug. That will help the engineer who'll be resolving the issue there a ton!

Yep, the timeout values in tests are 1. 
[gist](https://gist.github.com/nikli2009/582e46afb11a1c3ca7cec6e503afa09b) has been created and linked to the Chromium issue :D 

Let me know if there are other things I could help with and thanks for taking time review my PR 😄

test/page.spec.ts
```diff
@@ -1639,6 +1640,50 @@ describe('Page', function () {
...
...yRnSS7AQfcI3tC9orsGzkWOSU5B7kcuRlyB+T+gKB8ifwzBTYApapNfgAAAQAAAIUAOwAFAAAAAAACAAEAAgAWAAABAAFJAAAAAHjapZG9TgJBEMf/C4iChmBiDJVuic15QDCEwsSPSkMjJtR3eCCCd4Y7JFb3ChYWlsb4MPoSPoSVpf/dWwyJYuNNbue387E7MwtgUxQgoL75NaV1wmlNCWew9c1Z6m16RSbH3Q2kYcGoe8Mpeh4Mp+c4Q5nxEvJ4NJxlxrPhZZzjxfAKSvg0nENJFA3n0RRlw6vYF67hNfKT4SI2xLvhdWTFh+FX9j478w12qhC3JqEXyLbjh7Ju2/LM609Gzlh2PLcX+FGMIwTs8w5jDNDHJSL2XEYXO9RV2JQGyWWExAljQ+pDTBh9Cw8+cxxYtB1gRJFz54R651F71Cr6gpEWYv1XuNpaYrR4nooKmNHmeb7OHejV4R8x32G2h2vqMYa0Bej9WdEizyJ7m1Nw2LeqUuoalB7q2n1jnTIjYm+qtlPjsXSvEbOb2KVMtVi40ve4P+pKugh0ps+J/S/7t8nV9VQ7tLqckYpVb5rMu6X3Esf0dqlrtClfhfdXsafXhnn5CmpfaA13OgAAeNptz1cvg3EUx/HvoUMHWkrtvefzlC67Ouy9NwlaiSCVXrgnEuPWm3CH8vJo2v+lc/PJ75zknBzyyNZvGg//1QNInuSTjwEjJswUYMGKDTuFFFGMAycllOKijHLcVFBJFdXUUEsd9TTQSBPNtNBKG+100EkX3fTQSx/9aOiZ2wMM4sWHnwBBhhhmhFHGGGeCEJOEiRAlxhTTzDDLHPMssMgSy6ywyhrrbLDJFtvssMse+xxwyBHHnIiB+8wnbzyLkSdexSRmKRCLWMUmdimUIikWhzilREr54JNvfvgizSPv4pIyKedF3Kb45d1NQjenri40TYvkDGnKbPZkBkpd6VEOKAeVXqVP6VcGlEFlKKeu9uq69fwinkqenZ7cJnItTyynN2aIppLX2eCNhf8AwzpP+gB42tvB+L91A2Mvg/cGjoCIjYyMfZEb3di0IxQ3CER6bxAJAjIaImU3sGnHRDBsYFZw3cCs7bKBTcF1E7MDkzaYwwrksGlCOSxADqsslMMOUpYH4TBu4IBq5lRw3cXAUf+fgUl7I7NbmbZL5AYRbQDhkycvAAABUsGetAAA';
```
jschfflr: @nikli2009 Is this only happening for base64 decoded fonts or could this also be an url?

test/assets/pdf.html
```diff
@@ -0,0 +1,21 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>PDF</title>
+    <style>
+      @font-face {
+        font-family: 'custom-font';
+        font-display: swap;
+        src: url(https://custom-font.com/font.woff2) format('woff2');
```
jschfflr: Is it possible to use a local font here instead of loading something from a third party?

jschfflr (changes requested):

jschfflr:
@nikli2009 Thanks for updating the pull request! I left two more comments to reduce the size of the pr a bit - could you address them too?

test/assets/pdf.html
hottomato-c: that's a good point, will give it a go

test/page.spec.ts
hottomato-c: 🤔 good question, gonna try some other combination / scenarios to see if it could break the pdf call as well

hottomato-c:
> @nikli2009 Thanks for updating the pull request! I left two more comments to reduce the size of the pr a bit - could you address them too?

Happy to :D

hottomato-c:
> @nikli2009 Thanks for updating the pull request! I left two more comments to reduce the size of the pr a bit - could you address them too?

@jschfflr  Thanks for the review :D, got some quick updates : 
1. After some manual tests, it seems that the printToPDF call only hangs when using base64(valid base64 tho) and external url at same time.
2. Removed those problematic headerTemplate config in the tests. I realized that as we set the `timeout` to be 1 ms and it will just always timeout (even when `printToPDF` being executed normally).

jschfflr (approved):
Ah yeah, right. Good catch!

PR #7512, opened 2021-08-19
refactor(common): move actual constants

VoltrexKeyva:
The values of these constant variables are always the exact same when the `parseAriaSelector()` function is called, so these can be moved out of the function.


jschfflr (approved):
Looks good to me, thanks!

PR #7534, opened 2021-08-29
docs(wait for file chooser): clarifies wait for file chooser docs

zoltan-boros:
Clarifies `page.waitForFileChooser()` and class `FileChooser` docs.

Docs was misleading for me, therefore I wasted hours of work.
The intention of this PR is to clarify docs.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -2738,11 +2738,11 @@ Shortcut for [page.mainFrame().waitFor(selectorOrFunctionOrTimeout[, options[, .
 #### page.waitForFileChooser([options])
 
 - `options` <[WaitTimeoutOptions](####WaitTimeoutOptions)> Optional waiting parameters
-- returns: <[Promise]<[FileChooser]>> A promise that resolves after a page requests a file picker.
+- returns: <[Promise]<[FileChooser]>> A promise that resolves after a page requests a file picker _due to file upload_.
 
 > **NOTE** In non-headless Chromium, this method results in the native file picker dialog **not showing up** for the user.
 
-This method is typically coupled with an action that triggers file choosing.
+This method is typically coupled with an action that triggers file choosing due to file _upload_.
 The following example clicks a button that issues a file chooser and then
 responds with `/tmp/myfile.pdf` as if a user has selected this file.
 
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -2738,11 +2738,11 @@ Shortcut for [page.mainFrame().waitFor(selectorOrFunctionOrTimeout[, options[, .
 #### page.waitForFileChooser([options])
 
 - `options` <[WaitTimeoutOptions](####WaitTimeoutOptions)> Optional waiting parameters
-- returns: <[Promise]<[FileChooser]>> A promise that resolves after a page requests a file picker.
+- returns: <[Promise]<[FileChooser]>> A promise that resolves after a page requests a file picker _due to file upload_.
```
mathiasbynens: This is not accurate — the file does not have to be "uploaded" to any server at all.

mathiasbynens (changes requested):

mathiasbynens:
I see the distinction you're trying to make, but the suggested phrasing is misleading. Maybe we could just add a single note saying something like

> “File picker” refers to the operating system’s file selection UI that lets you browse to a folder and select a file to be shared with the web app, like the one that appears when clicking an `<input type=file>`. It’s not the “Save file” dialog.

docs/api.md
zoltan-boros: Updated the docs as per you suggested.

docs/api.md
```diff
@@ -3437,7 +3439,7 @@ Only one trace can be active at a time per browser.
 
 [FileChooser] objects are returned via the ['page.waitForFileChooser'](#pagewaitforfilechooseroptions) method.
 
-File choosers let you react to the page requesting for a file.
+File choosers let you react to the page requesting for file(s) to be loaded to the browser. (This file chooser does not cover "Save file" dialog.)
```
```suggestion
File choosers let you react to the page requesting for file(s) to be loaded by the web app. (This file chooser does not cover the “Save file” dialog.)
```

docs/api.md
```diff
@@ -2756,6 +2756,8 @@ await fileChooser.accept(['/tmp/myfile.pdf']);
 
 > **NOTE** This must be called _before_ the file chooser is launched. It will not return a currently active file chooser.
 
+> **NOTE** "File picker" refers to the operating system's file selection UI that lets you browse to a folder and select file(s) to be shared with the web app. It’s not the "Save file" dialog.
```
```suggestion
> **NOTE** “File picker” refers to the operating system’s file selection UI that lets you browse to a folder and select file(s) to be shared with the web app. It’s not the “Save file” dialog.
```

mathiasbynens (approved):

PR #7538, opened 2021-09-01
Correct `createPDFStream` documentation example

pjg:
Using existing example results in `browser.close()` being executed before the file is generated/written to disk. One needs to listen for the `end` event on the `Readable` stream before closing the browser, otherwise `UnhandledPromiseRejectionWarning: Error: Protocol error (IO.read): Target closed` exception is raised.


mathiasbynens (approved):

pjg:
🎉

mathiasbynens:
Thanks!

PR #7547, opened 2021-09-06
fix: added names in V9.1.1

tasneemkoushar:
Added names for APIs on the sidebar for V9.1.1


jschfflr (approved):
I'm not familiar with docusaurus but the change looks good to me from a technical side.

PR #7556, opened 2021-09-10
feat(oop iframes): integrate OOP iframes with the frame manager

jschfflr:
This pull request tries to add better support for OOP iframes (see https://github.com/puppeteer/puppeteer/issues/2548)

The current problem with OOP iframes is that they are moved to a different target. Because of this, the current implementation of Puppeteer pretty much ignores them. 

This change extends the FrameManager to already take OOP iframes into account and hides the fact that those frames are actually in different targets. 

Further work needs to be done to also make the NetworkManager aware of these and to make sure that settings like emulations etc. are also properly passed down to the new targets.


jschfflr:
@OrKoN Let me know what you think about this pr and if we should incrementally add OOP iframe support or have everything in one big batch?

OrKoN:
Could you describe on a high-level what was missing for OOPIF support and how we are adding it (and what are the difficulties)?

jschfflr:
> Could you describe on a high-level what was missing for OOPIF support and how we are adding it (and what are the difficulties)?

Sure, I updated the description.

mocha-config/puppeteer-unit-tests.js
```diff
@@ -25,8 +25,8 @@ module.exports = {
   ],
   spec: 'test/*.spec.ts',
   extension: ['js', 'ts'],
-  retries: process.env.CI ? 2 : 0,
+  // retries: process.env.CI ? 2 : 0,
```
OrKoN: uncomment?

mocha-config/puppeteer-unit-tests.js
```diff
@@ -25,8 +25,8 @@ module.exports = {
...
   parallel: !!process.env.PARALLEL,
   timeout: 25 * 1000,
-  reporter: process.env.CI ? 'spec' : 'dot',
+  // reporter: process.env.CI ? 'spec' : 'dot',
```
OrKoN: uncomment?

OrKoN:
What would be the effects of landing partial support? I guess it'd break some clients so maybe it's better to have all changes in one batch?

src/common/DOMWorld.ts
```diff
@@ -96,15 +98,17 @@ export class DOMWorld {
     `${name}_${contextId}`;
 
   constructor(
+    client: CDPSession,
     frameManager: FrameManager,
     frame: Frame,
     timeoutSettings: TimeoutSettings
   ) {
+    this._client = client;
```
OrKoN, line 108: can we/should we get the client from FrameManager? or are those different clients?

src/common/FrameManager.ts
```diff
@@ -316,24 +358,27 @@ export class FrameManager extends EventEmitter {
...
   }
 
-  async _ensureIsolatedWorld(name: string): Promise<void> {
-    if (this._isolatedWorlds.has(name)) return;
-    this._isolatedWorlds.add(name);
-    await this._client.send('Page.addScriptToEvaluateOnNewDocument', {
+  async _ensureIsolatedWorld(session: CDPSession, name: string): Promise<void> {
+    const key = `${session.id()}:${name}`;
+    if (this._isolatedWorlds.has(key)) return;
+    this._isolatedWorlds.add(key);
```
OrKoN, line 386: should not we track isolatedWorlds by session then?

src/common/FrameManager.ts
```diff
@@ -218,18 +225,53 @@ export class FrameManager extends EventEmitter {
...
-    debug('puppeteer:frame')(
-      `The frame '${event.targetInfo.targetId}' moved to another session. ` +
-        `Out-of-process iframes (OOPIF) are not supported by Puppeteer yet. ` +
-        `https://github.com/puppeteer/puppeteer/issues/2548`
+    const frame = this._frames.get(event.targetInfo.targetId);
+    const session = Connection.fromSession(this._client).session(
+      event.sessionId
     );
+    frame.updateClient(session);
+    session.on('Page.frameNavigated', (event) => {
```
OrKoN: should not we clean up listeners on detach?

src/common/FrameManager.ts
OrKoN: ah sorry, my bad

src/common/FrameManager.ts
```diff
@@ -347,6 +392,9 @@ export class FrameManager extends EventEmitter {
 
   _onFrameDetached(frameId: string): void {
     const frame = this._frames.get(frameId);
+    if (frame.isOOPFrame()) {
```
OrKoN: why is it needed?

src/common/Page.ts
```diff
@@ -510,15 +510,17 @@ export class Page extends EventEmitter {
...
-      const session = Connection.fromSession(client).session(event.sessionId);
-      const worker = new WebWorker(
-        session,
-        event.targetInfo.url,
-        this._addConsoleMessage.bind(this),
-        this._handleException.bind(this)
-      );
-      this._workers.set(event.sessionId, worker);
-      this.emit(PageEmittedEvents.WorkerCreated, worker);
+      if (event.targetInfo.type === 'worker') {
```
OrKoN: was it a bug?

OrKoN:
It looks good to me but I am a bit fuzzy about implications for other functionality. Is it easily possible to implement OOPIF support behind a flag? Not suggesting, but just curious if it'd be safer to allow users to opt-in and test at first.

src/common/DOMWorld.ts
jschfflr: The FrameManagers client is pointing to the main target whereas a DomWorld's target might either be the mainTarget or the target of an OOP iframe.

src/common/FrameManager.ts
jschfflr: `key` now includes the session id, so I think this should be addressed already? Or am I missing something?

src/common/FrameManager.ts
jschfflr: The session will be destroyed anyways if the frame disappears but I'll see if can include clean up to make it more future proof.

src/common/FrameManager.ts
jschfflr: If the frame becomes OOP, it will be detached from the parent targets session but it's still the same frame. Therefore it shouldn't actually be removed.

jschfflr: I'll add a comment to explain what's going on.

src/common/FrameManager.ts
OrKoN: yeah, I have already resolved that conversation.

src/common/Page.ts
jschfflr: Yeah ^^
Previously, an OOP iframe would have shown up as a worker.

src/common/FrameManager.ts
OrKoN: And what happens if a isOOPFrame==true frame is detached for real?

jschfflr:
> It looks good to me but I am a bit fuzzy about implications for other functionality. Is it easily possible to implement OOPIF support behind a flag? Not suggesting, but just curious if it'd be safer to allow users to opt-in and test at first.

As we currently don't support OOP iframes at all this isn't changing any existing behaviour but only adding new things. As long as we don't regress on existing behaviour, I think we should be fine without a flag.

OrKoN:
> As we currently don't support OOP iframes at all this isn't changing any existing behaviour but only adding new things. As long as we don't regress on existing behaviour, I think we should be fine without a flag.

Yeah but I mean new frames might show up unexpectedly for the users and things like this. E.g., they expected frames.length === 1 and suddenly an OOPIF shows up making frames.length === 2. Do you think that won't be common?

src/common/DOMWorld.ts
OrKoN: is it worth leaving a comment about that?

src/common/FrameManager.ts
jschfflr: I'll keep it as it as it follows the current approach but I'll investigate further in the future.

src/common/FrameManager.ts
jschfflr: I updated the change to address this. Thanks!

jschfflr:
@OrKoN Could you take another look at this?

src/common/FrameManager.ts
```diff
@@ -87,53 +86,70 @@ export class FrameManager extends EventEmitter {
...
+      client.send('Page.getFrameTree'),
     ]);
 
     const { frameTree } = result[1];
     this._handleFrameTree(frameTree);
     await Promise.all([
-      this._client.send('Page.setLifecycleEventsEnabled', { enabled: true }),
-      this._client
+      // OOP iframes will be waiting for the debugger initially.
+      client.send('Runtime.runIfWaitingForDebugger'),
```
OrKoN: Hm, I am sure I have left a comment but it's not here now:

why do we need to wait for debugger? It seems to be breaking FF tests and the code branch does not look specific to OOPIFs?

src/common/Page.ts
```diff
@@ -488,34 +488,42 @@ export class Page extends EventEmitter {
...
+        ) {
+          // If we don't detach from service workers, they will never die.
+          // We still want to attach to workers for emitting events.
+          // We still want to attach to iframes so sessions may interact with them.
+          // We detach from all other types out of an abundance of caution.
+          // See https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/devtools_agent_host_impl.cc?ss=chromium&q=f:devtools%20-f:out%20%22::kTypePage%5B%5D%22
+          // for the complete list of available types.
+
+          // Make sure that the Target continues running.
+          await session.send('Runtime.runIfWaitingForDebugger');
```
OrKoN: why this change? It does not look related to OOPIFs?

src/common/FrameManager.ts
jschfflr: I had to set `waitForDebuggerOnStart` to `true` in `Target.setAutoAttach` to make sure I can do the setup before anything happens that I would otherwise miss.
Not sure about FF support for this - maybe @whimboo has an idea?

OrKoN: But that'd run for all targets, right? Do we have waitForDebuggerOnStart for the main target in Puppeteer?

OrKoN: I think it's failing currently because it's invoked not only for oopifs but for all targets including the main target.

whimboo: In our partial CDP implementation we do not support any command related to debugging, just the important pieces related to (test) automation. As such `Runtime.runIfWaitingForDebugger` is not available.

jschfflr: Thanks @OrKoN for making me check again. It turned out that we don't need `waitForDebuggerOnStart` after all :)
I removed it and the tests are now passing in FF too.

test/oopif.spec.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable prettier/prettier */
```
OrKoN: can we bring prettier back?

test/target.spec.ts
```diff
@@ -113,26 +113,28 @@ describe('Target', function () {
...
+      const destroyedTargetPromise = new Promise((fulfill) =>
         context.once('targetdestroyed', (target) => fulfill(target))
       );
-      await page.evaluate(() =>
-        globalThis.registrationPromise.then((registration) =>
-          registration.unregister()
-        )
-      );
-      expect(await destroyedTarget).toBe(await createdTarget);
+      await page.evaluate(async () => {
```
OrKoN: are there any logic changes here?

OrKoN:
Looks good to me. Will this be marked as a breaking change?

test/oopif.spec.ts
```diff
@@ -47,7 +51,126 @@ describeChromeOnly('OOPIF', function () {
...
+    const { server } = getTestState();
+
+    await page.goto(server.EMPTY_PAGE);
+    await utils.attachFrame(
+      page,
+      'frame1',
+      server.CROSS_PROCESS_PREFIX + '/empty.html'
+    );
+
+    const frame = page.frames()[1];
```
OrKoN: e.g., `page.waitForFrame((frame, frameIdx) => frameIdx === 1)`

test/target.spec.ts
jschfflr: There are no logic changes here but I removed it for now.

jschfflr:
> Looks good to me. Will this be marked as a breaking change?

Yes, I consider this a breaking change.

test/oopif.spec.ts
jschfflr: Good idea! I'll update the patch :)

whimboo:
@jschfflr I cannot comment that much yet, but as it looks like a lot of tests are failing for Chrome. Nevertheless it's interesting to see that nothing breaks Firefox, and none of the tests have been marked as `itFirefoxFails`, which is great! So I'm happy to give more feedback when Chrome tests are all passing, and hopefully required changes do not affect Firefox.

OrKoN (approved):

test/oopif.spec.ts
jschfflr: I just checked why the tests started failing with this change:
There is no `frameIdx` in `waitForFrame` :D

OrKoN: I thought you'd add it :D  we can also use the frame.name()? or any other frame ID method?

src/common/FrameManager.ts
```diff
@@ -87,53 +86,81 @@ export class FrameManager extends EventEmitter {
...
+      this._handleFrameTree(client, frameTree);
+      await Promise.all([
+        client.send('Page.setLifecycleEventsEnabled', { enabled: true }),
+        client
+          .send('Runtime.enable')
+          .then(() => this._ensureIsolatedWorld(client, UTILITY_WORLD_NAME)),
+      ]);
+      if (client === this._client) {
+        // Network manager is not aware of OOP iframes yet.
+        await this._networkManager.initialize();
```
OrKoN: let's move it to Promise.all to keep the same semantic? i.e., currently it starts initing only after the runtime and lifecycle have been enabled.

```
await Promise.all([
        client.send('Page.setLifecycleEventsEnabled', { enabled: true }),
        client
          .send('Runtime.enable')
          .then(() => this._ensureIsolatedWorld(client, UTILITY_WORLD_NAME)),
        client === this._client || this._networkManager.initialize()
      ]);
```

src/common/FrameManager.ts
```diff
@@ -317,24 +380,26 @@ export class FrameManager extends EventEmitter {
...
-          .send('Page.createIsolatedWorld', {
+      this.frames()
+        .filter((frame) => frame._client === session)
+        .map((frame) =>
+          session.send('Page.createIsolatedWorld', {
             frameId: frame._id,
             worldName: name,
             grantUniveralAccess: true,
           })
-          .catch(debugError)
```
OrKoN, line 336: why don't we need debugError anymore?

src/common/FrameManager.ts
jschfflr: Sounds good!

OrKoN (approved):
Still LGTM. Please make sure that is marked as a breaking change in the release notes.

src/common/FrameManager.ts
jschfflr: Debug error is potentially hiding important errors. Instead, I chose to add a try/catch block in initialize that ignores the errors that should be ignored and throws all others.

jschfflr:
Landed and marked as breaking change 🎉

mattzeunert:
I'll need to do more work to find a shareable repro, but in case someone else runs into this, I sometimes get this error after upgrading to v11:

> Cannot read property '_updateClient' of undefined
> node_modules/puppeteer/lib/cjs/puppeteer/common/FrameManager.js:183:15

Update: no repro so far, but this is some of the target info:

```
attached: true
canAccessOpener: false
type: "iframe"
url: "chrome-untrusted://new-tab-page/one-google-bar?paramsencoded="
```

I think _onAttachedToTarget is firing 1ms before _onFrameAttached.

cetico:
@mattzeuner I am getting that error. How to fix it?

src/common/Page.ts
```diff
@@ -489,34 +489,41 @@ export class Page extends EventEmitter {
...
+          // We detach from all other types out of an abundance of caution.
+          // See https://source.chromium.org/chromium/chromium/src/+/main:content/browser/devtools/devtools_agent_host_impl.cc?ss=chromium&q=f:devtools%20-f:out%20%22::kTypePage%5B%5D%22
+          // for the complete list of available types.
+          client
+            .send('Target.detachFromTarget', {
+              sessionId: event.sessionId,
+            })
+            .catch(debugError);
+          return;
+        }
```
joshueis, line 511: Is there a reason why the logic under line#512 is performed only for worker targets? Is this preventing the iframe targets to execute that logic? If so, what is the purpose of it or why don't we need that for iframe targets?

TheoOliveira:
I am still having issues with the iframe. I have a simple automation that runs Puppeteer and a seems the iframe is oop. I am using latest version. Should this help? https://pptr.dev/api/puppeteer.page.evaluateonnewdocument

PR #7557, opened 2021-09-10
refactor(node): apply optimizations

VoltrexKeyva:
Replaced unnecessary template strings and used less calls for optimization.

### DIFF
--- a/src/node/BrowserFetcher.ts
+++ b/src/node/BrowserFetcher.ts
@@ -35,7 +35,7 @@ import createHttpsProxyAgent, {
 import { getProxyForUrl } from 'proxy-from-env';
 import { assert } from '../common/assert.js';
 
-const debugFetcher = debug(`puppeteer:fetcher`);
+const debugFetcher = debug('puppeteer:fetcher');
 
 const downloadURLs = {
   chrome: {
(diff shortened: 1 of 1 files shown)


src/node/BrowserFetcher.ts
```diff
@@ -112,10 +112,12 @@ function handleArm64(): void {
...
-          console.error(`The chromium binary is not available for arm64.`);
-          console.error(`If you are on Ubuntu, you can install with: `);
-          console.error(`\n sudo apt install chromium\n`);
-          console.error(`\n sudo apt install chromium-browser\n`);
+          console.error(
+            'The chromium binary is not available for arm64.' +
+              'If you are on Ubuntu, you can install with: ' +
+              '\n sudo apt install chromium\n' +
+              '\n sudo apt install chromium-browser\n'
+          );
```
mathiasbynens, line 120: This is not equivalent. Please revert this part of the change.

src/node/BrowserFetcher.ts
```diff
@@ -528,9 +528,9 @@ function installDMG(dmgPath: string, folderPath: string): Promise<void> {
       mountPath = volumes[0];
       readdirAsync(mountPath)
         .then((fileNames) => {
-          const appName = fileNames.filter(
+          const appName = fileNames.find(
             (item) => typeof item === 'string' && item.endsWith('.app')
-          )[0];
+          );
```
mathiasbynens, line 533: Nice!

src/node/BrowserFetcher.ts
VoltrexKeyva: It is equivalent now.

jschfflr (approved):
This change looks good to me, thanks!

PR #7573, opened 2021-09-16
feat: add ability to specify offsets for JSHandle.click

OrKoN:
Until now, the click would be always sent to the middle
point of the target element. With this change, one can define
offsets relative to the border box of the elements and click
different areas of an element.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -307,7 +307,7 @@
   * [elementHandle.boundingBox()](#elementhandleboundingbox)
   * [elementHandle.boxModel()](#elementhandleboxmodel)
   * [elementHandle.click([options])](#elementhandleclickoptions)
-  * [elementHandle.clickablePoint()](#elementhandleclickablepoint)
+  * [elementHandle.clickablePoint([offset])](#elementhandleclickablepointoffset)
   * [elementHandle.contentFrame()](#elementhandlecontentframe)
   * [elementHandle.dispose()](#elementhandledispose)
   * [elementHandle.drag(target)](#elementhandledragtarget)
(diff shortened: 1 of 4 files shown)


OrKoN:
I am still looking into the doclint failures but the code is ready to be reviewed I think. If you can tell what's wrong with the docs, I'd really appreciate :)

jschfflr (approved):
Looks good to me!

docs/api.md
```diff
@@ -4442,13 +4442,17 @@ This method returns boxes of the element, or `null` if the element is not visibl
   - `button` <"left"|"right"|"middle"> Defaults to `left`.
   - `clickCount` <[number]> defaults to 1. See [UIEvent.detail].
   - `delay` <[number]> Time to wait between `mousedown` and `mouseup` in milliseconds. Defaults to 0.
+  - `offset` <[Offset]> Offset in pixels relative to the border box of the element.
```
jschfflr: I think by overwriting this name in `check_public_api`, it now expects this to be of type `Object`

jschfflr (approved):

docs/api.md
OrKoN: it seems to complain nevertheless. Let's check on CI.

utils/doclint/check_public_api/index.js
```diff
@@ -371,6 +371,13 @@ function compareDocumentations(actual, expected) {
           expectedName: 'ClickOptions',
         },
       ],
+      [
+        'Method ElementHandle.clickablePoint() options',
```
jschfflr:
```suggestion
        'Method ElementHandle.clickablePoint() offset',
```

jschfflr: I think this should resolve the doc lint error.

OrKoN: wow, thanks!

jschfflr:
Should we get this landed or wait for a review from @mathiasbynens?

docs/api.md
```diff
@@ -4442,13 +4442,19 @@ This method returns boxes of the element, or `null` if the element is not visibl
   - `button` <"left"|"right"|"middle"> Defaults to `left`.
   - `clickCount` <[number]> defaults to 1. See [UIEvent.detail].
   - `delay` <[number]> Time to wait between `mousedown` and `mouseup` in milliseconds. Defaults to 0.
+  - `offset` <[Object]> Offset in pixels relative to the border box of the element.
+    - `x` <number> x-offset in pixels relative to the border box of the element.
+    - `y` <number> y-offset in pixels relative to the border box of the element.
```
mathiasbynens: Relative to which corner (top left or bottom left)? Please clarify this in the docs.

mathiasbynens (approved):
LGTM % doc clarification

PR #7574, opened 2021-09-16
feat: extend husky checks

jschfflr:
I added some more checks to huskys config to because not addressing them leads to a bunch of unnecessary cycles for reviews and CI and the overhead for the developer seems rather small. Let me know what you think!


OrKoN (approved):
LGTM

mathiasbynens:
Let's give this a try! (We can use `git commit --no-verify` `git push --no-verify` to skip the hooks, which I've never had to do before, but might start doing occasionally with this change.)

mathiasbynens (approved):

PR #7575, opened 2021-09-16
Fix ElementHandle#uploadFile to work with puppeteer.connect

### LINKED ISSUE #4405: uploadFile doesn't work with puppeteer.connect
### Steps to reproduce

Currently when used `inputElement.uploadFile` via `puppeteer.connect`, it uploads an empty file, as it doesn't exist on host machine. I believe this is a DevTools limitation and there's no possible workaround except uploading the file to remote puppeteer's host.

1. `const page = await puppeteer.connect(/* ... */);`
2. `const input = await page.$('input[type="file"]');`
3. `await input.uploadFile('./this/will/upload/an/empty/file');`

* uploaded files are serialised before sent to devtools instead of their file paths,
* or an error thrown saying this is not supported via `connect`.

**What happens instead?**

Puppeteer silently uploads an empty payload.

YusukeIwaki:
(in progress...)

### DIFF
--- a/src/common/JSHandle.ts
+++ b/src/common/JSHandle.ts
@@ -635,9 +635,10 @@ export class ElementHandle<
      avoid paying the cost unnecessarily.
     */
     const path = await import('path');
+    const mime = await import('mime');
     const fs = await helper.importFSModule();
     // Locate all files and confirm that they exist.
-    const files = await Promise.all(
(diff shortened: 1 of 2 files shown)


jrandolf-google:
Closed in favor of #8472.

PR #7582, opened 2021-09-20
feat(page): expose page.client()

OrKoN:
Puppeteer already allows creating a new CDP session
via target.createCDPSession but there is no way
to get access to any existing session to send
some additional commands.


mathiasbynens (approved):
LGTM but please add a test

jschfflr:
@OrKoN I think we should mark this as `@internal` as we have a bunch of internal assumptions about the state the connection is in. Our documentation recommends users to create a new CDP session instead.

OrKoN:
Oh, I didn't know. Where is it documented?

PR #7585, opened 2021-09-21
feat(page): mark page.client() as internal

OrKoN:
This PR marks the .client() method as internal since
we don't encourage our users to use it.


jschfflr:
Can you also remove it from the documentation?

jschfflr (approved):

PR #7587, opened 2021-09-21
fix: request an animation frame to fix flaky clickablePoint test

OrKoN:
Sometimes an element has not been layed out yet and, in this case,
clickablePoint fails because backend cannot compute content quads.

### DIFF
--- a/test/jshandle.spec.ts
+++ b/test/jshandle.spec.ts
@@ -309,7 +309,9 @@ describe('JSHandle', function () {
           <div style="cursor: pointer; width: 120px; height: 60px; margin: 30px; padding: 15px;"></div>
         `;
       });
-
+      await page.evaluate(async () => {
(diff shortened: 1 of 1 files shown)


test/jshandle.spec.ts
```diff
@@ -309,7 +309,9 @@ describe('JSHandle', function () {
           <div style="cursor: pointer; width: 120px; height: 60px; margin: 30px; padding: 15px;"></div>
         `;
       });
-
+      await page.evaluate(async () => {
```
jschfflr, line 312: Can you add a helper for this to keep the code DRY?

jschfflr (changes requested):

test/jshandle.spec.ts
jschfflr: Never mind, it's probably not worth it...

jschfflr (approved):

Issue #6894, opened 2021-02-17
Change recommended version of Firefox from Nightly to Beta/Release


whimboo:
As of right now Puppeteer downloads Firefox Nightly by default. This was pretty helpful for us while getting started to implement CDP commands in Firefox, and get new features out as soon as possible.

But that also have some side-effects as what we have seen in the past months. Firefox Nightly is not meant to be that stable, and also has various experiments enabled that can badly affect the user experience. As such I would propose that we change Puppeteer to no longer download Firefox Nightly by default but a release version (or beta?) of Firefox. Note that with the upcoming Firefox 86 release next week the remote agent will be available by default. 

Due to an oversight on my side we [missed to enable our tests](https://bugzilla.mozilla.org/show_bug.cgi?id=1693058) on those channels. As such maybe we could go with Firefox 87 Beta first, or wait a little bit longer so we can get confidence that everything works as expected? @mathiasbynens what do you think?

To prepare the change, which specific areas would have to be changed in Puppeteer to make that move?

CC'ing @jackfranklin, @jschfflr, @juliandescottes, @jgraham, @janodvarko

mathiasbynens:
IMHO anything would be better than the status quo, which uses "whatever the latest Nightly is". Preparing a change now, to go with Firefox 87 beta first, sounds good to me.

In terms of which Puppeteer changes would be needed, this would boil down to updating
https://github.com/puppeteer/puppeteer/blob/b046fd79bd97e8623c40204193bff028096e4474/src/node/BrowserFetcher.ts. Happy to review a patch!

whimboo:
Thanks. Meanwhile I also got our tests enabled for beta and these are all fine. So I will try with 87 beta first next week.

One problem we might hit is that we usually do not offer zip archives of releases on Windows. I would have to check where I can get these from.

whimboo:
Internally I will track the work via https://bugzilla.mozilla.org/show_bug.cgi?id=1694127.

@mathiasbynens would it be also ok to use an exe or msi installer? We officially don't have zip archives of releases, but only for candidate build, which I wouldn't like to fallback to. If that's ok hopefully Node.js has some package to do a silent installation.

mathiasbynens:
My gut reaction is that it seems a little intrusive to download and run `.exe`s upon `npm install puppeteer`. The expectation is that `node some-puppeteer-script.js` can launch browser binaries, but not that `npm install puppeteer` does it — and I feel uncomfortable changing that.

[...]

PR #7610, opened 2021-09-28
chore: disable certain Fission features for Firefox

whimboo:
Enable basic Fission support for Firefox' CDP implementation. Some other Fission specific features have to be kept disabled for now.

### DIFF
--- a/src/node/Launcher.ts
+++ b/src/node/Launcher.ts
@@ -513,8 +513,11 @@ class FirefoxLauncher implements ProductLauncher {
       // Make sure opening about:addons will not hit the network
       'extensions.webservice.discoverURL': `http://${server}/dummy/discoveryURL`,
 
-      // Force disable Fission until the Remote Agent is compatible
-      'fission.autostart': false,
+      // Temporarily force disable BFCache in parent (https://bit.ly/bug-1732263)
+      'fission.bfcacheInParent': false,
+
+      // Force all web content to use a single content process
+      'fission.webContentIsolationStrategy': 0,
 
       // Allow the application to have focus even it runs in the background
       'focusmanager.testmode': true,


whimboo:
With all the checks passing, I'm close to request review for this change. But there is one question remaining.

@jschfflr does the order of emitted events actually matter in Puppeteer or CDP? It doesn't look like. Background here is that with enabling Fission in Firefox we have a lot more async code, and as such event order is mixed up compared to what we emit without Fission or what Chrome emits. As example, when navigating away from a frameset page, the contained frame is destroyed after the execution context of the new page has been created.

Can we assume that the order doesn't matter? If that is the case I would remove strict event order assertions in our own internal tests for CDP. With that done we could enable Fission with some features still disabled, which is a great step forward.

Related bug here is https://bugzilla.mozilla.org/show_bug.cgi?id=1601245.

mathiasbynens:
cc @OrKoN as well

whimboo:
From our side the landing of this PR actually blocks any work for issue #6894.

jschfflr:
@whimboo For the OOP iframe support that I'm currently working on, there is the implicit assumption of a frame being attached to its parent frame again before the target is destroyed if the frame is moving from OOP to in process again. But I don't think we have them documented anywhere. But I'm pretty sure there will be a bunch of them. 
But as long as the tests are passing in a non flaky way, I'd say we should just give it a try. And either it just works or we discover a bunch of hidden assumptions that we could write tests for...

jschfflr (approved):

whimboo:
Thanks @jschfflr! Do you know if this OOP support in Puppeteer is something that will ship in Q4 this year? If yes is there something that I have to be aware of? Happy to chat about that via the #webdriver Element/Matrix channel. Thanks in advance.

whimboo:
FYI I'll keep this PR as draft for now until we have merged these changes in our own repository first. This might take a couple days.

jschfflr:
@whimboo I'm currently working on this: https://github.com/puppeteer/puppeteer/pull/7556
That change would already integrate OOP iframes into the frame manager. But there are probably a bunch of other things that have to be changed too.

whimboo:
Thanks! So given that current Puppeteer tests are passing, does Puppeteer also pass with Chrome having OOP enabled? Or do you force it off for the time being? If you do, we may want to retain the `fission.autostart=false` preference as long as possible?

whimboo:
@jschfflr we are kinda interested to get this change into the next Puppeteer release. Is there already a time scheduled for it, or do we have some more days to get our CI polished up. Nevertheless we might be ok in shipping these changes. Thanks.

whimboo:
As heads-up, my patches on https://bugzilla.mozilla.org/show_bug.cgi?id=1601245 have been landed. If all goes well these will reach the Firefox Nightly release today. Our internal CI hasn't shown any failures for Puppeteer whether if Fission is enabled or disabled (as per default right now). @jschfflr if you notice failures today please let me know. If it works well I will ask for uplifting the patch to Firefox 94 beta. Once it reaches release we can continue to work on issue #6894.

Having no failures we could consider merging this PR tomorrow or on Friday.

whimboo:
Internal tests are working great with this change. So I do not see a reason why we would have to delay the merge of this PR anytime longer. @jschfflr could you please merge? Thanks!

jschfflr:
@whimboo Awesome, thanks! Happy to get this landed :)

whimboo:
@jschfflr thanks for merging. Do you already know when the next Puppeteer release is due? In about a months time work will be started to disallow turning off Fission in Firefox. As such this will negatively affect Puppeteer users. So getting a release shipped until then would be great. Thanks.

Issue #7592, opened 2021-09-23
PUPPETEER_DOWNLOAD_PATH does not support relative paths


thesam:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 10.4.0
* Platform / OS version: Ubuntu 21.04
* URLs (if applicable):
* Node.js version: v14.17.6

We set the Puppeteer download path in our project's .npmrc. We want to store the downloaded Chromium outside node_modules to avoid downloading it everytime we run "npm ci".

We want to use a **relative** path to make sure it works for all our users, on both Windows and Linux.

**What steps will reproduce the problem?**
1. Set PUPPETEER_DOWNLOAD_PATH with a relative path using an environment variable, npm config or .npmrc.
2. Install puppeteer

For example:
```
PUPPETEER_DOWNLOAD_PATH=../bar npx puppeteer
```

**What is the expected result?**
Puppeteer downloads Chromium to ../bar.

**What happens instead?**
The extract-zip library requires an absolute path and throws an Error.
```
Downloading Chromium r901912 - 140.9 Mb [====================] 99% 0.0s 
ERROR: Failed to set up Chromium r901912! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
Error: Target directory is expected to be absolute
    at Object.module.exports [as default] (/home/user/.npm/_npx/34327/lib/node_modules/puppeteer/node_modules/extract-zip/index.js:167:11)
    at install (/home/user/.npm/_npx/34327/lib/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:392:37)
    at BrowserFetcher.download (/home/user/.npm/_npx/34327/lib/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js:249:19)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
    at async downloadBrowser (/home/user/.npm/_npx/34327/lib/node_modules/puppeteer/lib/cjs/puppeteer/node/install.js:51:5)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! puppeteer@10.4.0 install: `node install.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the puppeteer@10.4.0 install script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
```

jschfflr:
[...]

PR #7613, opened 2021-09-29
fix: fix BrowserFetcher._getFolderPath to support relative install paths

nadir:
This patch fixes the BrowserFetcher._getFolderPath method so that it supports relative download paths using PUPPETEER_DOWNLOAD_PATH or npm config

Issues: #7592

### DIFF
--- a/src/node/BrowserFetcher.ts
+++ b/src/node/BrowserFetcher.ts
@@ -419,9 +419,7 @@ export class BrowserFetcher {
    * @internal
    */
   _getFolderPath(revision: string): string {
-    return path.resolve(
-      path.join(this._downloadsFolder, `${this._platform}-${revision}`)
-    );
+    return path.resolve(this._downloadsFolder, `${this._platform}-${revision}`);
   }
 }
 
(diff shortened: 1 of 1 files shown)


nadir:
@googlebot I signed it!

src/node/BrowserFetcher.ts
```diff
@@ -419,7 +419,9 @@ export class BrowserFetcher {
    * @internal
    */
   _getFolderPath(revision: string): string {
-    return path.join(this._downloadsFolder, `${this._platform}-${revision}`);
+    return path.resolve(
```
jschfflr: You can just path multiple path pieces to `path.resolve`:
```js
path.resolve(this._downloadsFolder, `${this._platform}-${revision}`);
```

See: https://nodejs.org/api/path.html#path_path_resolve_paths

jschfflr (changes requested):

jschfflr:
@nadhirhb Thanks for your contribution!

jschfflr (approved):

PR #7565, opened 2021-09-13
Add webp to screenshot options

jonchurch:
closes: #5348

Not sure since when, but webp is an option in [`Page.captureScreenshot`](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-captureScreenshot) now.

All that is needed is to allow users to pass `type: "webp"` through. 

After making said change to allow passing webp through as an option, this code successfully creates webp screenshots using either `type: "webp"` or using the `.webp` extension from the filepath.
```js
const puppeteer = require('puppeteer')

async function run() {
const browser = await puppeteer.launch()
  const page = await browser.newPage()
  await page.setViewport({ width: 1280, height: 800 })
  await page.goto('https://www.nytimes.com/')
  await page.screenshot({ path: `${__dirname}/nytimes.webp`, fullPage: true, type: 'webp'})
  await page.screenshot({ path: `${__dirname}/nytimes_alt.webp`, fullPage: true,})
  await browser.close()
}

run().then(() => console.log('done')).catch(console.log)
```
Happy to take any help getting this mergeable, ie writing tests.

I'd upload a webp here but github doesn't support them!

PR #7631, opened 2021-10-05
fix: add webp to screenshot quality option allow list

tjunnone:
#7565 added webp image format support to the Page.screenshot API, but did not add webp to the list of formats that can use the "quality" option. Specifying the quality option now throws "options.quality is unsupported for the webp screenshots".

This PR allows properly passing through the screenshot quality option from Puppeteer to the browser as expected.


jschfflr (approved):
Thanks for catching this @tjunnone!

Issue #7458, opened 2021-08-03
Roll Chromium 94 after 2021-08-26


sadym-chromium:
Chrome 94 is in beta, so it's time to roll Puppeteer update.
https://chromiumdash.appspot.com/schedule

mathiasbynens:
Jan, could you PTAL?

mathiasbynens:
Jan, what's the status here?

bradisbell:
FYI, the next one is coming up in three days:  https://github.com/puppeteer/puppeteer/issues/7459

jschfflr:
Unfortunately, the roll is blocked on some breaking changes that happened in Chromium. 
We are currently figuring out a solution.

jschfflr:
Quick update - Unfortunately, the roll is still blocked because of the following situation:

A while ago, a [change in the network domain](https://chromium-review.googlesource.com/c/chromium/src/+/2856099) caused some headers and status codes to move from the `requestWillBeSent`/`responseReceived` to the according `requestWillBeSentExtraInfo`/`responseReceivedExtraInfo`. Right now, Puppeteer does not use these events at all because before this change, the information provided in the `requestWillBeSent`/`responseReceived` was enough to fulfill the needs of the provided API surface. But now that some of the information moved there, we have to implement support for them. 
Unfortunately though, there is no guarantee that these events will be emitted for all requests and if they are, there is no guarantee on the order. That's why we are not able to just listen to them too. Instead we have to find a way within Chromium to tell Puppeteer if there will be `...ExtraInfo` events that it should be aware of.

josepharhar:
Hey everyone!

I am working on a new CDP feature to enable puppeteer to leverage responseReceivedExtraInfo and therefore get the raw headers back: https://chromium-review.googlesource.com/c/chromium/src/+/2898747
Here is a doc for those interested in more details: https://docs.google.com/document/d/1NM30Wg_aM3-RFZaD_lQuWQj8my8XoR9YpMi60QH1lHU/edit

I am also working on a puppeteer patch which uses the new feature, I'll open a PR for it soon.

jschfflr:
Thanks for the update @josepharhar!

josepharhar:
The PR is ready: https://github.com/puppeteer/puppeteer/pull/7640
It still depends on the new flag in chromium, but the code can still otherwise be reviewed!

akornatskyy:
[...]

Issue #7560, opened 2021-09-11
[Bug]: Memory leak in `NetworkManager` for cached responses


tjenkinson:
### Bug description

See https://github.com/puppeteer/puppeteer/pull/5140#issuecomment-917399754

Currently https://github.com/puppeteer/puppeteer/blob/8d1a697fc3ef10300d886beb961a45ea1451c427/src/common/NetworkManager.ts#L423-L434 is not called if the response is from the cache, meaning anything from the cache stays in
https://github.com/puppeteer/puppeteer/blob/8d1a697fc3ef10300d886beb961a45ea1451c427/src/common/NetworkManager.ts#L111

Quoting @jschfflr:
>It does indeed look like a memory leak. But calling _forgetRequest in _onResponseReceived might cause problems because there's also the ResponseReceivedExtraInfo event that might arrive before or after ResponseReceived. Can you file an issue for that and assign it to me as there's currently some work being done to improve support for ResponseReceivedExtraInfo to unblock rolling the latest Chromium.

### Puppeteer version

10.2.0

### Node.js version

16.9.0

### npm version

7.21.1

### What operating system are you seeing the problem on?

macOS

josepharhar:
> is not called if the response is from the cache

Do you mean if `Network.requestServedFromCache`/`_onRequestServedFromCache` is called? Or do you mean if the response is a 304 response?

`Network.requestServedFromCache` means that the renderer process's internal cache fulfilled the request without it going to the browser process and going through the real network stack.

Requests can also go through the network stack/service, make a real request over the wire, and then be served from network service's HTTP cache.

In either case, Network.loadingFailed or Network.loadingFinished should still be emitted by chrome and _requestIdToRequest should still get cleared, right?

tjenkinson:
Ah I think I was thinking `_onRequestServedFromCache` was a final call but if `_onLoadingFailed` or `_onLoadingFinished` is always called last then I don’t think there is actually a bug

josepharhar:
Yeah, I am somewhat certain that `Network.requestServedFromCache` always gets `Network.onLoadingFinished` or `Network.onLoadingFailed` afterwards. If you ever find an example where this isn't the case, please share it.

tjenkinson:
Cool. I’m not actively using this at the moment so will close this and reopen if the above happens

PR #7640, opened 2021-10-06
fix(network): use ResponseReceivedExtraInfo

josepharhar:
In http://crrev.com/908142 I removed a bunch of stuff from
Network.responseReceived in favor of Network.responseReceivedExtraInfo.
However, Puppeteer can't use responseReceivedExtraInfo because
not every responseReceived comes with a corresponding ExtraInfo.

In http://crrev.com/c/2898747 I am adding a flag to responseReceived to
let us know if there is a corresponding ExtraInfo event. With this
information, Puppeteer can wait to emit its response event until the
ExtraInfo has arrived.

Fixes/unblocks https://github.com/puppeteer/puppeteer/issues/7458


jschfflr:
@josepharhar Thanks for the pull request - happy to review it!
Could you check if https://github.com/puppeteer/puppeteer/issues/7560 is also addressed by this?

josepharhar:
> Could you check if #7560 is also addressed by this?

I'm guessing no, but I'll follow up in that bug

src/common/NetworkManager.ts
```diff
@@ -110,6 +110,31 @@ export class NetworkManager extends EventEmitter {
...
+   */
+  _requestIdToResponseReceivedExtraInfo = new Map<
+    string,
+    Protocol.Network.ResponseReceivedExtraInfoEvent[]
+  >();
+  _requestIdToResponseReceived = new Map<
+    string,
+    Protocol.Network.ResponseReceivedEvent
+  >();
+  _requestIdToRedirectInfoMap = new Map<
```
jschfflr: For the other maps, you didn't include `Map` in the name. Maybe remove it here too?

jschfflr (approved):
The change looks good to me, thanks!

src/common/NetworkManager.ts
josepharhar: Yeah this was because I got a compile error because there is also a function named `_requestIdToRedirectInfo`, but that's not the case for the other maps, so I just appended "Map" to the name here.
Do you have any suggestions for what to do here?

jschfflr: Ah I see, never mind in this case!

src/common/NetworkManager.ts
```diff
@@ -459,7 +459,7 @@ export class NetworkManager extends EventEmitter {
       }
     }
 
-    this._emitRedirectResponse(lastRequest, event.redirectResponse, null);
+    this._emitRedirectResponse(lastRequest, event.redirectResponse, extraInfo);
```
OrKoN: I think I have tried by I think it din't make a difference: there is still an extra unprocessed extra info remaining (the one with status code 200)

OrKoN: I hope it helps though.

test/navigation.spec.ts
```diff
@@ -155,7 +155,7 @@ describe('navigation', function () {
         ? 'net::ERR_CERT_INVALID'
         : 'net::ERR_CERT_AUTHORITY_INVALID';
 
-    itFailsFirefox('should fail when navigating to bad SSL', async () => {
+    it('should fail when navigating to bad SSL', async () => {
```
OrKoN: Oh, I think I incorrectly marked it as `it` instead of 'itFailsFirefox' after I removed `.only` for the repro.

src/common/HTTPResponse.ts
```diff
@@ -78,8 +78,11 @@ export class HTTPResponse {
       ip: responsePayload.remoteIPAddress,
       port: responsePayload.remotePort,
     };
-    // TODO extract statusText from extraInfo.headersText instead if present
-    this._statusText = responsePayload.statusText;
+    if (extraInfo && extraInfo.headersText) {
+      this._statusText = extraInfo.headersText.match(/[^ ]* [^ ]* (.*)\r/)[1];
```
OrKoN: does it always match the regex? Or could it be somehow fail if, e.g., a server is misconfigured (configured differently)?

josepharhar: I added a test case and improved the parsing, it should not explode if the server sends something weird.

OrKoN: it looks it fails on Firefox. Let me land the roll CL first and then we can improve this part.

benallfree:
@josepharhar I'm puzzled by this PR because it seems that some of these changes (or very similar) are already in `main` from #7764. For example, in main:


```ts
  /*
   * The below maps are used to reconcile Network.responseReceivedExtraInfo
   * events with their corresponding request. Each response and redirect
   * response gets an ExtraInfo event, and we don't know which will come first.
   * This means that we have to store a Response or an ExtraInfo for each
   * response, and emit the event when we get both of them. In addition, to
   * handle redirects, we have to make them Arrays to represent the chain of
   * events.
   */
  _requestIdToResponseReceivedExtraInfo = new Map<
    string,
    Protocol.Network.ResponseReceivedExtraInfoEvent[]
  >();

```

josepharhar:
This PR was copied into here: https://github.com/puppeteer/puppeteer/pull/7764

PR #7644, opened 2021-10-07
chore: ignore stdout when dumpio isn't used

whimboo:
When the browser child process has logging enabled
and output on stdout isn't constantly processed,
the brower process is about to freeze.

To avoid such a situation at least the stdout
pipe shouldn't be set by default but only if
dumpio is enabled.

### DIFF
--- a/src/node/BrowserRunner.ts
+++ b/src/node/BrowserRunner.ts
@@ -64,10 +64,13 @@ export class BrowserRunner {
   start(options: LaunchOptions): void {
     const { handleSIGINT, handleSIGTERM, handleSIGHUP, dumpio, env, pipe } =
       options;
-    let stdio: Array<'ignore' | 'pipe'> = ['pipe', 'pipe', 'pipe'];
+    let stdio: Array<'ignore' | 'pipe'>;
     if (pipe) {
       if (dumpio) stdio = ['ignore', 'pipe', 'pipe', 'pipe', 'pipe'];
       else stdio = ['ignore', 'ignore', 'ignore', 'pipe', 'pipe'];
+    } else {
+      if (dumpio) stdio = ['pipe', 'pipe', 'pipe'];
+      else stdio = ['pipe', 'ignore', 'pipe'];
     }
     assert(!this.proc, 'This process has previously been started.');
     debugLauncher(


whimboo:
A Puppeteer script like the following as run with Firefox will cause a freeze:

```
'use strict';

(async () => {
    // dumpio: true,
    headless: false,
    extraPrefsFirefox: {
      'remote.log.level': 'trace',
    },
  });

  await page.goto('http://example.org');
  await browser.close();
})();
```

Once `dumpio` is set to `true` the freeze will be gone. CC @jschfflr.

whimboo:
I'm not that familiar with pipes, so I wonder if that would be the correct approach. Especially the Windows failures might indicate something that additionally use `stdout` only on this platform? @jackfranklin or @jschfflr do you have some early feedback? Thanks!

jschfflr (approved):

jschfflr:
While I'm not an expert on pipes myself I do think that they are backed up by some buffer. And if that one's full, writing to it stalls until there's enough space again. So I guess, this could potentially solve it!

whimboo:
Let me try to run some more retries of the jobs to see if the failing tests vary. Or are those known to sometimes fail?

whimboo:
I can see two permanently failing tests:

```
1) Frame specs
       "before all" hook in "Frame specs":
     ProtocolError: Protocol error (Target.setDiscoverTargets): can't access property "documentTitle", this.browsingContext.currentWindowGlobal is null get title@chrome://remote/content/cdp/targets/TabTarget.jsm:108:5
```

And the following which is random between different tests but always with `Cannot read property 'close' of null`, which should basically be `await state.browser.close();`

```

  2) Evaluation specs
       "after all" hook in "Evaluation specs":
     TypeError: Cannot read property 'close' of null
      at Context.<anonymous> (test\mocha-utils.ts:231:25)
      at processImmediate (internal/timers.js:461:21)
```

The latter means that there is no valid browser instance created in `before all`. Next I'll push a change that enables trace logging. Maybe that gives some indication what's wrong.

whimboo:
On Windows there are a lot of orphaned processes of Firefox that the completion step has to kill. Maybe that is related to https://bugzilla.mozilla.org/show_bug.cgi?id=1656962 but I miss to actually connect that to my changes on this PR. I think that I will temporarily revert my changes and get tests running with the original code.

whimboo:
I actually also ignored `stdin` with my initial patch. Resetting this specific change and use a `pipe` again makes the tests pass. I just force-pushed a rebased patch, and hope it will still pass.

whimboo:
@jschfflr it looks all fine now.

whimboo:
Thanks for merging!

PR #7650, opened 2021-10-07
chore: update types for JSHandle

### LINKED ISSUE #7583: Unclear return behavior for `getProperty`
### Steps to reproduce

N/A

* Puppeteer version: 10.0.0 (but seems true for 10.2.0 also)
* Platform / OS version: Ubuntu 20.04
* URLs (if applicable): N/A
* Node.js version: 14.17 LTS

The puppeteer [docs website](https://pptr.dev/#?product=Puppeteer&version=v10.0.0&show=api-jshandlegetpropertypropertyname) says `getProperty` on a `JSHandle` returns a promise that will resolve to another `JSHandle`

Same is true for the markdown docs in the repo [here](https://github.com/puppeteer/puppeteer/blob/v10.2.0/docs/api.md#jshandlegetpropertypropertyname)

The function definition [here](https://github.com/puppeteer/puppeteer/blob/2b5c0019dc3744196c5858edeaa901dff9973ef5/src/common/JSHandle.ts#L194) however indicates it may also return undefined.

It should be consistent across the docs and the implementation

**What happens instead?**
It isn't

I don't think im missing anyting obvious, is this is a typo / bug / known inconsistency?

jschfflr:
Closes #7583


OrKoN (approved):
LGTM thanks!

PR #7657, opened 2021-10-08
chore: reduce docusaurus versions that we generate

jackfranklin:
The Docusaurus process has been OOMing on the bots; this is a problem
we'll want to fix properly when we are shipping the new docs fully, as
we want to generate docs for old versions; but for now whilst we're not
fully pushing users to this site, we can just generate one version and
not worry about multiples.

This commit also bumps the version of Docusaurus to the latest beta
version.


jackfranklin:
PTAL @mathiasbynens . This is really a temporary measure to avoid the build issue. As discussed I think the better plan here is to move the docs into their own repo.

mathiasbynens (approved):

mathiasbynens:
Thanks!

As discussed offline, it’d be nice to decouple the website build process from the main Puppeteer repository in the future.

PR #7658, opened 2021-10-08
chore: keep ff unit tests on windows from blocking ci

### LINKED ISSUE #7659: Firefox tests are failing on windows
This is probably due to the latest FF nightly release.
Example: https://github.com/puppeteer/puppeteer/runs/3832483753?check_suite_focus=true

jschfflr:


jschfflr:
@whimboo FYI

.github/workflows/main.yml
```diff
@@ -166,6 +166,7 @@ jobs:
 
       - name: Run unit tests on Firefox
         uses: nick-invision/retry@v2
+        continue-on-error: true
```
OrKoN, line 169: which tests is it related to and is there an issue to track the progress of flipping this flag back?

jschfflr: I created #7659 to keep track of this.
Unfortunately, actions keep failing for different FF unit tests. This happened before and is usually due to a problem with FF using the latest nightly release. 
An example run would be: https://github.com/puppeteer/puppeteer/runs/3832483753?check_suite_focus=true

OrKoN (approved):

.github/workflows/main.yml
whimboo: That's probably something Fission related, now that we do not force disabling it. I'm going to investigate via https://bugzilla.mozilla.org/show_bug.cgi?id=1734896. Thanks for letting me know.

Issue #7589, opened 2021-09-22
[Bug]: Puppeteer 10.4.0 ignores the proxy-server launch argument


anjackson:
### Bug description

Steps to reproduce the problem:

1. Install 10.4.0
2. Run this script, which sets a non-existent proxy server:

```
const puppeteer = require('puppeteer');

(async () => {
  const browser = await puppeteer.launch({
    args: [ '--proxy-server=http://non-existent-domain:8000' ]
  });
  const page = await browser.newPage();
  await page.goto('https://example.com');
  await page.screenshot({ path: 'example.png' });

  await browser.close();
})();
```

Earlier versions will fail (correctly!) because the proxy server is not there.  10.4.0 ignores the proxy setting and runs as if the setting was not there.

### Puppeteer version

10.4.0

### Node.js version

16.8.0

### npm version

7.21.0

### What operating system are you seeing the problem on?

macOS

Ha2ryZhang:
maybe this
```
  const context = await browser.createIncognitoBrowserContext({proxyServer:"http://127.0.0.1:7890"});
```

jschfflr:
@anjackson Thanks for your bug report! Not sure why the `--proxy-server` argument doesn't work anymore. I'll look into it!

anjackson:
Ah, looks like I messed up the test case.  Sorry about that.

The failing case uses `createIncognitoBrowserContext`, like this:

```javascript
const puppeteer = require('puppeteer');

(async () => {
  const browser = await puppeteer.launch({
    args: [ '--proxy-server=http://localhost:18000', '--no-sandbox' ]
  });

  const context = await browser.createIncognitoBrowserContext();
  const page = await context.newPage();
  await page.goto('https://example.com');
  await page.screenshot({ path: 'example.png' });

  await browser.close();
})();      
```

Looking through the 10.2.0 > 10.4.0 diff, I think a default of `proxyServer=''` is being set [here](https://github.com/puppeteer/puppeteer/compare/puppeteer:v10.2.0...puppeteer:v10.4.0#diff-2fb2a2631063a3779bee361f5aeb2148b320092239bced39f2ac0fb63fed7debR318) when an Incognito session is created. Perhaps `proxyServer=null` would have the intended effect of defaulting to whatever the overall configuration is?

anjackson:
[...]

PR #7677, opened 2021-10-12
fix: ensure incognito sessions obey proxy server settings (#7589)

anjackson:
This attempts to resolve issue #7589 - i.e. it's a bug fix because as of 10.4.0 Puppeteer will ignore proxy settings when using incognito sessions.

No tests added. It would be good to have a test that would spot this kind of regression, but it is beyond my abilities to create such a test. (I am not really much of a JavaScript programmer).  Issue #7589 includes an example that fails under 10.4.0, ...TBC...

**Summary**

This PR attempts to repair an accidental breaking change introduced in 10.4.0.


OrKoN:
Thanks for the PR! It looks like the same issue will be addressed in https://github.com/puppeteer/puppeteer/pull/7770

Issue #7615, opened 2021-09-29
'userDataDir option should restore state' test is flakily timing out for Firefox


jschfflr:
Example run: https://github.com/puppeteer/puppeteer/pull/7614/checks?check_run_id=3745478501
@whimboo Do you know what might be causing this?

whimboo:
Hard to say without seeing trace logs. But I would assume that Firefox had a problem during startup so that enabling the CDP protocol happened too late. 

@jschfflr, would it be possible to set the `webdriver.log.level=Trace` preference for Firefox when it's running in CI? The log files will be larger but we find it very helpful when investigating intermittent failures like that.

whimboo:
We actually have https://bugzilla.mozilla.org/show_bug.cgi?id=1713030 for that. So I would take a look over there.

jschfflr:
Given that it's also happening on your side would it be possible to run the tracing version there and to disable the test on Puppeteer's CI for now?

whimboo:
Yesterday I noticed a strange behavior when running [Puppeteer with Firefox and having trace logging enabled](https://bugzilla.mozilla.org/show_bug.cgi?id=1734208). Whenever our CDP implementation logs something to `stdout` or `stderr` Firefox freezes because writing to these pipes is blocking (`fputs` never returns). 

@jschfflr, given that Puppeteer uses both could it be that it's no longer processing the output once it has received the DevTools WebSocket URL for CDP? The situation on this issue is a bit different but it could be related. Locally it is kinda annoying because I'm not able to run Puppeteer with trace logs anymore. Maybe something has been changed here recently?

jschfflr:
@whimboo Do you think this should be resolved with your recent `pipe` PR (https://github.com/puppeteer/puppeteer/pull/7644)

whimboo:
Probably not, but it's towards the path of letting me run tests locally with custom builds. We will see how it goes once that PR got merged.

whimboo:
The problem is that this test creates a temporary folder for the Firefox profile, but the default Firefox preferences aren't getting set. This means the profile isn't actually usable with Puppeteer. Chrome doesn't exhibit this problem because all settings are done via command line arguments.

@jschfflr extracting the profile location from the Firefox arguments inside the Launcher is not ideal and failure prone if eg no path has been set after `--profile`. Do you think that it would make sense to actually pass the profile folder through the launch options as `userDataDir` instead?

whimboo:
[...]

PR #7684, opened 2021-10-13
Correctly prepare Firefox profiles with required preferences for all scenarious

whimboo:
PR to fix issue #7615 by making sure that also temporary folders for the Firefox profile are properly prepared with required preferences.

### DIFF
--- a/src/node/Launcher.ts
+++ b/src/node/Launcher.ts
@@ -23,6 +23,7 @@ import { Browser } from '../common/Browser.js';
 import { BrowserRunner } from './BrowserRunner.js';
 import { promisify } from 'util';
 
+const copyFileAsync = promisify(fs.copyFile);
 const mkdtempAsync = promisify(fs.mkdtemp);
 const writeFileAsync = promisify(fs.writeFile);
 
@@ -85,7 +86,6 @@ class ChromeLauncher implements ProductLauncher {
       debuggingPort = null,
     } = options;
 
-    const profilePath = path.join(tmpDir(), 'puppeteer_dev_chrome_profile-');
     const chromeArguments = [];
     if (!ignoreDefaultArgs) chromeArguments.push(...this.defaultArgs(options));
     else if (Array.isArray(ignoreDefaultArgs))
@@ -95,8 +95,6 @@ class ChromeLauncher implements ProductLauncher {
         )
       );
     else chromeArguments.push(...args);
-
-    let temporaryUserDataDir = null;
 
     if (
       !chromeArguments.some((argument) =>
@@ -113,9 +111,28 @@ class ChromeLauncher implements ProductLauncher {
         chromeArguments.push(`--remote-debugging-port=${debuggingPort || 0}`);
       }
     }
-    if (!chromeArguments.some((arg) => arg.startsWith('--user-data-dir'))) {
-      temporaryUserDataDir = await mkdtempAsync(profilePath);
-      chromeArguments.push(`--user-data-dir=${temporaryUserDataDir}`);
+
+    let userDataDir;
+    let isTempUserDataDir = true;
+
+    // Check for the user data dir argument, which will always be set even
+    // with a custom directory specified via the userDataDir option.
+    const userDataDirIndex = chromeArguments.findIndex((arg) => {
+      return arg.startsWith('--user-data-dir');
+    });
+
+    if (userDataDirIndex !== -1) {
+      userDataDir = chromeArguments[userDataDirIndex].split('=')[1];
+      if (!fs.existsSync(userDataDir)) {
+        throw new Error(`Chrome user data dir not found at '${userDataDir}'`);
+      }
+
+      isTempUserDataDir = false;
+    } else {
+      userDataDir = await mkdtempAsync(
+        path.join(tmpDir(), 'puppeteer_dev_chrome_profile-')
+      );
+      chromeArguments.push(`--user-data-dir=${userDataDir}`);
     }
 
     let chromeExecutable = executablePath;
@@ -145,7 +162,8 @@ class ChromeLauncher implements ProductLauncher {
       this.product,
       chromeExecutable,
       chromeArguments,
-      temporaryUserDataDir
+      userDataDir,
+      isTempUserDataDir
     );
     runner.start({
       handleSIGHUP,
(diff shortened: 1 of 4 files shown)


whimboo:
@jschfflr let me know if this is something that would work for you. As already asked on the issue itself, I wonder if it would make sense to pass the `userDataDir` via the launch options. Or if this could be a follow-up given that it would also affect Chrome.

whimboo:
Here are the first results for this PR:
https://github.com/puppeteer/puppeteer/actions/runs/1337331528/attempts/1

I will trigger some more to make sure it also succeeds all the time in CI.

OrKoN (approved):
LGTM thanks (let's let Jan to take a look as well)

src/node/Launcher.ts
```diff
@@ -305,10 +305,20 @@ class FirefoxLauncher implements ProductLauncher {
...
-    ) {
+    const profileArgIndex = firefoxArguments.findIndex((arg) => {
+      return ['-profile', '--profile'].includes(arg);
+    });
+
+    if (profileArgIndex !== -1) {
+      const profilePath = firefoxArguments[profileArgIndex + 1];
+      if (!profilePath) {
+        temporaryUserDataDir = await this._createProfile(extraPrefsFirefox);
+        firefoxArguments.push(temporaryUserDataDir);
```
whimboo: We might not actually need this specific code path given that Firefox will fail during startup if no path has been specified. But it spawns a modal dialog the user has to interact with, which isn't ideal for automation.

@jschfflr shall Puppeteer fail instead, or is it ok to assume a new temporary profile is wanted?

src/node/Launcher.ts
```diff
@@ -305,10 +305,20 @@ class FirefoxLauncher implements ProductLauncher {
...
+      } else {
+        const prefs = this.defaultPreferences(extraPrefsFirefox);
+        await this.writePreferences(prefs, profilePath);
```
whimboo: @jschfflr what's the actual expected behavior for user profiles that exists but should be used once for Puppeteer. Right now we would overwrite all the preferences which is not ideal. Instead we could only write the user.js file, which values will automatically be copied into prefs.js, and maybe fail if that file already exists?

Otherwise creating a backup of the prefs.js would be an idea, but there might be cases when the clean-up doesn't really work and we might miss to restore it.

whimboo (commented):
During an internal discussion two important points have been raised and need to be clarified first. Please see my inline comments.

src/node/Launcher.ts
jschfflr: That's an interesting case. Would firefox still work normally with the same user profile after the changes to `users.js`?

src/node/Launcher.ts
jschfflr: I feel like having a modal dialog show up is less then ideal for the automation usecase. Especially since users might be running the scripts without a display and wouldn't even see the modal...

jschfflr:
The user profiles seem to be a tricky thing. I'm wondering what kind of usecases we want to support and how they can be implemented. From the top of my head, I guess the following situations should be considered:

1. A Puppeteer user wants to store state - normal firefox will not be running with this user profile
2. A Puppeteer user wants to pick up state that has been created through a previous session with normal firefox
3. A Puppeteer user wants a normal firefox instance to pick up the state from a previous puppeteer run
4. There is a back and forth between normal FF and Puppeteer using the same user profile
5. What happens if normal FF and Puppeteer try to use the same user profile at the same time?

Let me know if there are other situations to take into account!

src/node/Launcher.ts
whimboo: Right, so it means that we have two options here. We assume that the consumer missed to add the profile argument and Puppeteer can bail out with an error. Or we just take it that some new temporary profile should be used, and we keep doing as what the current code does.

src/node/Launcher.ts
whimboo: Any preferences in `user.js` will be copied by Firefox to `prefs.js` and just removing `user.js` will not be enough to get it reverted. That means that all the changes will remain post Puppeteer. 

Given that we tweak the profile for automation the user might indeed see differences in Firefox' behavior afterward. I see two options:

1. Create a backup of prefs.js and restore it during teardown, which will not always work like in cases the user hits `Ctrl+C`. Also changes to `prefs.js` will trigger different behaviors for various components in Firefox, and can cause other files in the profile to get modified too.

2. Create a copy of the user's profile in the temp folder and let Puppeteer remove it as usual. 

Only with option 2 the users profile will be kept identical and no traces from a usage with Puppeteer will remain in the original profile. But note that this might also have side-effects if multiple tests require the same profile and in worst cases steps as done by a previous test.

With geckodriver we actually make use of option 2 and never modify the original profile.

whimboo:
>     2. A Puppeteer user wants to pick up state that has been created through a previous session with normal firefox

This might be the most important case IMHO given that the profile can be pre-configured with imported certificates, accepted cookies and others.

>     3. A Puppeteer user wants a normal firefox instance to pick up the state from a previous puppeteer run

I'm not sure about the benefits here and what they actually could make use of. I would consider this a rare case.

>     5. What happens if normal FF and Puppeteer try to use the same user profile at the same time?

It's not possible. The profile is locked once a Firefox instance is using it.

> Let me know if there are other situations to take into account!

You got a pretty good listing. So I cannot think of something else right now.

whimboo:
Today (I was out the last two weeks) I noticed that this broken behavior is also causing a nearly permanent failure (#7615) in our own CI. As such I would like to get it fixed sooner than later. @jschfflr if we can find a proper solution in the next days I would appreciate. Thanks.

jschfflr:
Happy to get this through :)
What do you think about going with option 2
> Create a copy of the user's profile in the temp folder and let Puppeteer remove it as usual.

to get us unblocked and then iterating on that to find a better solution going forward?

whimboo:
Thank you for the quick reply. Yes, that sounds fine. I'll go ahead with option 2 and we can observe incoming issues in case that this new behavior causes trouble to users.

src/node/Launcher.ts
whimboo: @jschfflr what shall we do here? Is throwing an error ok, or shall the launcher create a new temporary folder?

jschfflr: I'm usually in favor of making errors explicit so I'd go with throwing an error. Otherwise it might be difficult to debug. WDYT?

whimboo: Yes, sounds good. I will make that change.

src/node/Launcher.ts
whimboo: Actually creating a temporary copy of the specified user profile causes [some of the user-data-dir tests to fail](https://github.com/puppeteer/puppeteer/blob/ad7f1de44f78df4a304c8e052935e53d4685530c/test/launcher.spec.ts#L216-L229) because the expected path isn't the same anymore.

@jschfflr as it looks like this might be a larger refactoring because the newly created temp directory will not be accessible by the consumer, and the Browser class doesn't have a property.

What do you suggest? Shall we go with a backup of pref.js for now and delete both .js files after Puppeteer finished? It would at least copy the behavior as what is used for Chrome by using the real user data dir.

jschfflr: Ah damn, that's too bad. Let's go with your proposed solution then!

whimboo: I can do. What would be a good place to actually reset the prefs.js and delete user.js? Is https://github.com/puppeteer/puppeteer/blob/main/src/node/BrowserRunner.ts#L100 the right place so that it's always executed?

jschfflr: Yes, that's probably the best place for it!

whimboo:
@jschfflr I'm struggling a bit with the prefix of the commits. Is `chore` the right one to use? If not what shall I use for the commits? As I saw recently the changes for Firefox that I did not that long ago were not automatically mentioned in the changelog. Is there a manual process to request additions?

OrKoN:
Here is some guidance on the commits https://github.com/puppeteer/puppeteer/blob/main/CONTRIBUTING.md#commit-messages I think all commits in a PR get squashed though and only the final commit message matters. I am not sure if there is a manual process for generating the changelog now. cc @mathiasbynens

whimboo:
> Here is some guidance on the commits https://github.com/puppeteer/puppeteer/blob/main/CONTRIBUTING.md#commit-messages I think all commits in a PR get squashed though and only the final commit message matters. I am not sure if there is a manual process for generating the changelog now. cc @mathiasbynens

Thanks for pointing that out! I rebased my commits and updated the commit message prefixes appropriately. I would still like to see the commits pushed individually if possible.

The last three CI checks were all successful, so if that's the case again the PR would be ready for another review @jschfflr. Thanks in advance!

OrKoN:
I am not sure Jan would be able to review this any time soon. So let me take another look.

I am not sure it's possible to maintain individual commits as the squash merge is enforced for this repo.

OrKoN (approved):
Still LGTM

whimboo:
Thanks for the review @OrKoN! 

I didn't know that squash is enforced. So I will squash them all together and give it a proper commit message. I assume after this update you can land it, or do we have to wait for someone else for doing that?

OrKoN:
@whimboo yes, I can land it then

whimboo:
@OrKoN the commits have all been squashed and I force-pushed them. Also all the tests are still passing. So it looks like it's ready for landing.

OrKoN:
Merged!

whimboo:
Thanks a lot!

PR #7718, opened 2021-10-25
fix: dead lock when execution contexts are created in a different order

OrKoN:
Sometimes it's possible that the ExecutionContextCreated message
for a frame arrives before the frame has been registered by
Puppeteer. In this case, the DOM world for the frame will never
have the execution context set resulting in a indefinite await
for the result of DOMWorld::executionContext().

### DIFF
--- a/src/common/ExecutionContext.ts
+++ b/src/common/ExecutionContext.ts
@@ -389,4 +389,11 @@ export class ExecutionContext {
     });
     return this._adoptBackendNodeId(nodeInfo.node.backendNodeId);
   }
+
+  /**
+   * @internal
+   */
+  _setWorld(world: DOMWorld): void {
+    this._world = world;
+  }
 }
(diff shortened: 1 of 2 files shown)


OrKoN:
Mathias/Jan, please take a look. I found this bug which looks like a race condition so I am not sure if I can write a reliable test. Do you have any ideas for testing?

jschfflr (approved):

jschfflr:
The change looks good to me!
About the testing: What do you think about adding a `MockConnectionTransport` with a bunch of helpers for common scenarios that can then be used to feed puppeteer the events in the right order?

OrKoN:
Is MockConnectionTransport something that already exists? I think it'd require lots of mocks to make sure that Puppeteer ends in the right state as I am not sure about all expectations Puppeteer might have. I wonder if we have true unit tests in Puppeteer? It does not look like this class is unit-testable

jschfflr:
No, `MockConnectionTransport` does not exist yet. And I also don't think we have anything other than integration tests with Chromium :/

jschfflr:
@OrKoN Should we discard this pull request?

OrKoN:
Yes!

Issue #7721, opened 2021-10-25
[Bug]: aria/ queries don't support single-quote delimiters for role names


andyearnshaw:
### Bug description

Steps to reproduce the problem:

1. Open a page in Puppeteer with at least one element with `role="region"` attribute set
2. Call `page.$()` with the following query: `aria/[role='region']`
3. Observe that the result is `null`

### Puppeteer version

v10.4.0

### Node.js version

v16.12.0

### npm version

v8.1.0

### What operating system are you seeing the problem on?

Linux, macOS, Windows

### Relevant log output

```shell
[17:34:32] › Error: waiting for selector `Format[role='tab']` failed: timeout 20000ms exceeded
xceeded
    at new WaitTask (/Users/andyearnshaw/Projects/dashvis/node_modules/.pnpm/puppeteer@5.5.0/node_modules/puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:505:34)
    at DOMWorld.waitForSelectorInPage (/Users/andyearnshaw/Projects/dashvis/node_modules/.pnpm/puppeteer@5.5.0/node_modules/puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:416:26)
    at Object.waitFor (/Users/andyearnshaw/Projects/dashvis/node_modules/.pnpm/puppeteer@5.5.0/node_modules/puppeteer/lib/cjs/puppeteer/common/AriaQueryHandler.js:62:21)
    at DOMWorld.waitForSelector (/Users/andyearnshaw/Projects/dashvis/node_modules/.pnpm/puppeteer@5.5.0/node_modules/puppeteer/lib/cjs/puppeteer/common/DOMWorld.js:312:29)
    at Frame.waitForSelector (/Users/andyearnshaw/Projects/dashvis/node_modules/.pnpm/puppeteer@5.5.0/node_modules/puppeteer/lib/cjs/puppeteer/common/FrameManager.js:842:51)
    at waitFor (/Users/andyearnshaw/Projects/dashvis/resources/e2e-core/framework/utils/interactions.js:90:30)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Object.ariaClick (/Users/andyearnshaw/Projects/dashvis/resources/e2e-core/framework/utils/interactions.js:270:20)
    at async worldConstructor.<anonymous> (/Users/andyearnshaw/Projects/dashvis/workspaces/report-page/e2e/features/support/steps/mouse.js:57:5)
[...]

PR #7750, opened 2021-11-05
fix(ariaqueryhandler): allow single quotes in aria attribute selector

andyearnshaw:
This updates the regular expression used to parse aria attribute
selectors so that single quotes may be used as an alternative to double
quotes, e.g. `aria/Single button[role='button']`.

Issues: #7721


johanbay (approved):
LGTM. Thanks for adding this!

PR #7764, opened 2021-11-11
feat(chromium)!: roll to Chromium 97.0.4692.0 (r938248)

OrKoN:
Issues: #7458

### DIFF
--- a/src/common/HTTPResponse.ts
+++ b/src/common/HTTPResponse.ts
@@ -13,7 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { CDPSession } from './Connection.js';
+import { ProtocolMapping } from 'devtools-protocol/types/protocol-mapping.js';
+
+import { EventEmitter } from './EventEmitter.js';
 import { Frame } from './FrameManager.js';
 import { HTTPRequest } from './HTTPRequest.js';
 import { SecurityDetails } from './SecurityDetails.js';
@@ -26,6 +28,13 @@ import { ProtocolError } from './Errors.js';
 export interface RemoteAddress {
   ip: string;
   port: number;
+}
+
+interface CDPSession extends EventEmitter {
+  send<T extends keyof ProtocolMapping.Commands>(
+    method: T,
+    ...paramArgs: ProtocolMapping.Commands[T]['paramsType']
+  ): Promise<ProtocolMapping.Commands[T]['returnType']>;
 }
 
 /**
@@ -55,7 +64,8 @@ export class HTTPResponse {
   constructor(
     client: CDPSession,
     request: HTTPRequest,
-    responsePayload: Protocol.Network.Response
+    responsePayload: Protocol.Network.Response,
+    extraInfo: Protocol.Network.ResponseReceivedExtraInfoEvent | null
   ) {
     this._client = client;
     this._request = request;
@@ -68,13 +78,17 @@ export class HTTPResponse {
       ip: responsePayload.remoteIPAddress,
       port: responsePayload.remotePort,
     };
-    this._status = responsePayload.status;
+    // TODO extract statusText from extraInfo.headersText instead if present
     this._statusText = responsePayload.statusText;
     this._url = request.url();
     this._fromDiskCache = !!responsePayload.fromDiskCache;
     this._fromServiceWorker = !!responsePayload.fromServiceWorker;
-    for (const key of Object.keys(responsePayload.headers))
-      this._headers[key.toLowerCase()] = responsePayload.headers[key];
+
+    this._status = extraInfo ? extraInfo.statusCode : responsePayload.status;
+    const headers = extraInfo ? extraInfo.headers : responsePayload.headers;
+    for (const key of Object.keys(headers))
+      this._headers[key.toLowerCase()] = headers[key];
+
     this._securityDetails = responsePayload.securityDetails
       ? new SecurityDetails(responsePayload.securityDetails)
       : null;
--- a/src/common/NetworkManager.ts
+++ b/src/common/NetworkManager.ts
@@ -13,12 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+import { ProtocolMapping } from 'devtools-protocol/types/protocol-mapping.js';
 import { EventEmitter } from './EventEmitter.js';
+import { Frame } from './FrameManager.js';
 import { assert } from './assert.js';
 import { helper, debugError } from './helper.js';
 import { Protocol } from 'devtools-protocol';
-import { CDPSession } from './Connection.js';
-import { FrameManager } from './FrameManager.js';
 import { HTTPRequest } from './HTTPRequest.js';
 import { HTTPResponse } from './HTTPResponse.js';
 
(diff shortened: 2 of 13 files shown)


src/common/HTTPResponse.ts
```diff
@@ -68,13 +69,17 @@ export class HTTPResponse {
       ip: responsePayload.remoteIPAddress,
       port: responsePayload.remotePort,
     };
-    this._status = responsePayload.status;
+    // TODO extract statusText from extraInfo.headersText instead if present
```
mathiasbynens, line 81: For my own understanding, is this a TODO you expect to wrap up before landing this patch, or a follow-up? (In case of the latter, let's file a tracking bug?)

mathiasbynens (approved):

src/common/HTTPResponse.ts
OrKoN: I copied it from Joey's PR https://github.com/puppeteer/puppeteer/pull/7640 so I plan to add him as a reviewer here.

OrKoN:
@josepharhar I am trying to roll Chromium now that your CLs have landed. I have copied the fixes from your PR https://github.com/puppeteer/puppeteer/pull/7640 but there is still one test failing (https://github.com/puppeteer/puppeteer/runs/4176540107?check_suite_focus=true) Do you think it could be related to your CLs too? Also, please review the changes.

OrKoN:
It looks like the status of the response after the redirect is sometimes wrong (302 instead of 200) https://github.com/puppeteer/puppeteer/runs/4177611057?check_suite_focus=true

josepharhar:
Sorry I didn’t look at this today, hopefully I’ll have a chance tomorrow

josepharhar:
I checked out this PR on my glinux machine, and I wasn't able to reproduce the failure when running the tests.
I wonder if its flaky or if I'm doing something wrong... can you reproduce the failure locally?

Wouldn't it make sense to do this roll and let me merge my PR separately? You could mark any network tests as failing if needed.

OrKoN:
> I checked out this PR on my glinux machine, and I wasn't able to reproduce the failure when running the tests.
> I wonder if its flaky or if I'm doing something wrong... can you reproduce the failure locally?

No, I cannot reproduce locally as well :( I am going to try to reproduce it again next week.

> Wouldn't it make sense to do this roll and let me merge my PR separately? You could mark any network tests as failing if needed.

Usually we fix everything before or with the roll PR. I think it does not make much sense to roll with the broken functionality. Feel free to make changes to this branch if you need to! Or you can take over other changes to your PR and include the roll there.

OrKoN:
> I wonder if its flaky

I believe it's flaky. Sometimes there are completely different network test failing but all seem to be failing on assertions for the network response attributes.

josepharhar:
I think that `page.goto` isn't waiting for `responseReceviedExtraInfo`, I'm working on a fix and I'll push it to this PR soon

josepharhar:
I don't have permission to push to this branch, but I added commits to my PR to try to try to wait for ExtraInfo events. Unfortunately the tests still fail, and I don't know if my changes actually made it better or not because I can't reproduce the failure locally and I can't even push a commit that uses `it.only` because the github actions build will fail before it tries to run the tests.
It might be that my changes don't account for redirects. I can try to look into this more another day.

OrKoN:
@josepharhar is the eslint complaining about .only? usually you can push `/* eslint-disable RuleNameItComplainsAbout */` to suppress it

OrKoN:
I have merged the changes from the other branch here and cleaned up some debugging stuff. Let's see if it passes now.

OrKoN:
@josepharhar I was able to reproduce the flake with the CDP log on the bot. Here is the relevant part https://gist.github.com/OrKoN/369323d11b1106c525f23a4ad9ec5a5b So the test fails because the response code is 302 instead of 200. I can see the extra info events with 302 and 200 status codes so I suspect that we might be calling _emitResponseEvent with the wrong extraInfo event.

OrKoN:
@josepharhar I came to a conclusion that there is an extra extraInfo event sent for empty.html URL sometimes that cause the wrong HTTP status for the entire redirection chain. I have captured that in a reproducible new (true) unit test. Could you please take a look?

src/common/NetworkManager.ts
```diff
@@ -399,28 +512,83 @@ export class NetworkManager extends EventEmitter {
...
-      new Error('Response body is unavailable for redirect responses')
+    const request = this._requestIdToRequest.get(responseReceived.requestId);
+    // FileUpload sends a response without a matching request.
+    if (!request) return;
+
+    const extraInfos = this._requestIdToResponseExtraInfo(
+      responseReceived.requestId
     );
-    this._forgetRequest(request, false);
+    if (extraInfos.length) {
```
OrKoN, line 525: @josepharhar in particular, I have added this check here: is my understanding correct that there should be no unprocessed extraInfo events if we emit _emitResponseEvent?

josepharhar:
For each redirect, chromium should emit one requestWillBeSentExtraInfo and one responseReceivedExtraInfo.
It should look like this for a request which redirects twice:
- first request:
  - requestWillBeSent
  - requestWillBeSentExtraInfo
  - responseReceivedExtraInfo
- second request:
  - requestWillBeSent
  - requestWillBeSentExtraInfo
  - responseReceivedExtraInfo
- third final request:
  - requestWillBeSent
  - requestWillBeSentExtraInfo
  - responseReceivedExtraInfo
  - responseReceived

Is the issue that the order of events as I listed is wrong? Or is the issue that there are more or less ExtraInfo events than I listed?
If there are more or less ExtraInfo events, then it's a chromium problem. If they are just in a different order in the test, then it's a puppeteer problem.

josepharhar:
I'm reading the log you shared to try to answer that question

josepharhar:
Ok, after looking at the logs, I believe that chromium is emitting the right events. I'll re-read the redirect ExtaInfo logic in NetworkManager and try out your test case. Thanks for the info!

OrKoN:
> * first request:
>   
>   * requestWillBeSent
>   * requestWillBeSentExtraInfo
>   * responseReceivedExtraInfo
> * second request:
>   
>   * requestWillBeSent
>   * requestWillBeSentExtraInfo
>   * responseReceivedExtraInfo
> * third final request:
>   
>   * requestWillBeSent
>   * requestWillBeSentExtraInfo
>   * responseReceivedExtraInfo
>   * responseReceived

I believe in the log there are 4 responseReceivedExtraInfo events instead of 3. Or am I wrong interpreting the log? All 4 seems to have the same requestId.

josepharhar:
Chromium is emitting the right events, it just turns out that a lot of the code that I wrote here doesn't actually handle the events coming out of order. I'm rewriting a bunch of stuff to handle more cases.

src/common/HTTPResponse.ts
josepharhar: I think I tried to do this but ran into an issue. I'll try again and update with more context

josepharhar:
I finally got it to pass the tests in my PR! https://github.com/puppeteer/puppeteer/pull/7640
Sorry for providing a broken implementation at first, I guess I should be happy that the bots somehow emit the events out of order... it might be a good idea to expand on the NetworkManager test you added to make sure we have better coverage of the out of order cases.

OrKoN:
@josepharhar thanks a lot for fixing it! I am not very familiar with the backend implementation for these events but it'd be great to expand the unit tests if you know what are the possible/supported (or also unsupported) orders in which events could arrive.

src/common/HTTPResponse.ts
josepharhar: Ok, I just pushed a commit to my PR which takes care of this.

josepharhar:
Sounds good, I can add more coverage to `NetworkManager.spec.ts`. I don't think it needs to block this PR.

OrKoN:
Thanks, I think that we don't need to block on that too. I have just removed some awaits that I have added previously to see if they are redundant with the fixes. For the unit tests, I think we need to make some changes in types and TS configuration so that we could define test cases without defining the entire event payloads that are not used by the network manager directly.

josepharhar:
We could just move the `NetworkManager.spec.ts` changes in this PR to a new one, would that resolve it?

OrKoN:
Let's keep the test for coverage and refactor later. I think this is ready to land: do you want me to land your PR or is it okay to land this one?

PR #7766, opened 2021-11-11
chore: remove documentation deploy github action

jackfranklin:
It is causing failed builds and we are not currently using it, plus we
want to move this setup into a new repository, such that it can run
independently of Puppeteer's main codebase.

For now, until that work is done, let's remove it to not cause build
issues and to not waste cycles on CI.


mathiasbynens (approved):

mathiasbynens:
As a more extreme follow-up, I’d even be okay with removing the website-related code altogether for now, since we’re not actively working on it, we’re not currently maintaining it, and we plan on moving it elsewhere anyway. But this patch is the minimal change that gets our CI green again so 👍

PR #7782, opened 2021-11-19
fix: prevent potential browser stall when "dumpio: false"

etanol:

Bugfix (potentially).

No.

No.

**Summary**

When calling `puppeteer.launch()` with options `pipe` and `dumpio` set to `false` there's a risk of blocking the browser process, resulting in timeouts during a test execution.

With `pipe` and `dumpio` set to false, the standard error output (_stderr_, fd 2) of the browser _subprocess_ is attached to  a pipe.  However, the pipe is never drained.  If the browser process happens to emit a significant amount of **error** output, the pipe may become full.  As a result, the browser may block indefinitely at the next attempt of writing to _stderr_.

To put it graphically:

Process blockage diagram

This change just aligns the configuration of `dumpio` behavior for _stdin_, _stdout_ and _stderr_ no matter the value of `pipe`.

No.

**Other information**

While browsers are multi-threaded and this issue seems speculative, sometimes more fundamental dependencies may be writing to _stderr_.  For example, the GTK library (the browser GUI toolkit used in Unix systems) can become very chatty under some environment conditions.  A simple write to _stderr_ can block a significant chunk of logic in the browser; for example, one of the event loops.  This is because the standard I/O descriptors are rarely configured or operated in a non-blocking fashion.


OrKoN (changes requested):
It appears that some tests are failing due to this.

OrKoN:

PR #7798, opened 2021-11-25
fix: parse statusText from the extraInfo event

OrKoN:
Issues: #7458

### DIFF
--- a/src/common/HTTPResponse.ts
+++ b/src/common/HTTPResponse.ts
@@ -78,8 +78,9 @@ export class HTTPResponse {
       ip: responsePayload.remoteIPAddress,
       port: responsePayload.remotePort,
     };
-    // TODO extract statusText from extraInfo.headersText instead if present
-    this._statusText = responsePayload.statusText;
+    this._statusText =
+      this._parseStatusTextFromExtrInfo(extraInfo) ||
+      responsePayload.statusText;
     this._url = request.url();
     this._fromDiskCache = !!responsePayload.fromDiskCache;
     this._fromServiceWorker = !!responsePayload.fromServiceWorker;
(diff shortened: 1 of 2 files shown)


OrKoN:
@josepharhar PTAL (extracted your change from #7640)

josepharhar (approved):

josepharhar:
Does this include the test I added with empty status text?

OrKoN:
Overlooked it: I am gonna add it in a moment.

src/common/HTTPResponse.ts
```diff
@@ -94,6 +95,22 @@ export class HTTPResponse {
...
+  /**
+   * @internal
+   */
+  _parseStatusTextFromExtrInfo(
+    extraInfo: Protocol.Network.ResponseReceivedExtraInfoEvent | null
+  ): string | undefined {
+    if (!extraInfo || !extraInfo.headersText) return;
+    const firstLine = extraInfo.headersText.split('\r', 1)[0];
+    if (!firstLine) return;
+    const match = firstLine.match(/[^ ]* [^ ]* (.*)/);
```
sadym-chromium, line 107: It would be great to have unit testы on this parsing.

sadym-chromium:
It would be really great to have the parsing unit-tested. If you have some strong reasons not to do it, please clarify them and go ahead with merging.

sadym-chromium (approved):

src/common/HTTPResponse.ts
OrKoN: yeah, the reason is that Puppeteer does not have unit tests. The tests that are called "unit" are added with this PR (one existing test and one empty status text test).

PR #7800, opened 2021-11-26
chore(release): mark v12.0.0

OrKoN:


README.md
```diff
@@ -39,7 +39,7 @@ npm i puppeteer
 # or "yarn add puppeteer"
 ```
 
-Note: When you install Puppeteer, it downloads a recent version of Chromium (~170MB Mac, ~282MB Linux, ~280MB Win) that is guaranteed to work with the API. To skip the download, download into another path, or download a different browser, see [Environment variables](https://github.com/puppeteer/puppeteer/blob/v11.0.0/docs/api.md#environment-variables).
+Note: When you install Puppeteer, it downloads a recent version of Chromium (~170MB Mac, ~282MB Linux, ~280MB Win) that is guaranteed to work with the API. To skip the download, download into another path, or download a different browser, see [Environment variables](https://github.com/puppeteer/puppeteer/blob/v12.0.0/docs/api.md#environment-variables).
```
sadym-chromium, line 42: NIT: are those numbers still the same?

sadym-chromium (approved):
Great job! Finally we can roll it!

README.md
OrKoN: I have no idea tbh :) just following https://github.com/puppeteer/puppeteer/blob/main/CONTRIBUTING.md#releasing-to-npm here

Issue #7805, opened 2021-11-28
[Bug]: version 12.0.0 – Error: Unexpected extraInfo events for request on `page.goto`


yurique:
### Bug description

Steps to reproduce the problem:

1. run a browserless/chrome docker image (https://hub.docker.com/r/browserless/chrome)
```
docker run --rm -it -p 3000:3000 -e DEBUG='browserless*' -e PREBOOT_CHROME=true -e MAX_CONCURRENT_SESSIONS=10 --name browserless browserless/chrome:latest
```
2. try to load a page with puppeteer (`puppeteer-core`):
  ```
    const browser = await puppeteer.connect({ browserWSEndpoint: 'ws://localhost:3000' });
    const page = await browser.newPage();
    await page.goto('https://google.com', {
      waitUntil: 'networkidle2',
    });
    await page.screenshot({ path: 'puppeteer.png' })
    await browser.close();
  ```

Getting an exception (see the "Relevant log output"):
```
Error: Unexpected extraInfo events for request 8574FCDD373436FD23BAE6CB1FF3CAED
```

Added a `console.log` to print the `extraInfos` (see the "Relevant log output"):

With puppeteer `v11.0.0` this doesn't happen – same code, same browserless, only chaning the version in `package.json` and running `yarn` to install – the page loads and I'm getting the screenshot in the file.

### Puppeteer version

12.0.0

### Node.js version

v16.4.0

### npm version

yarn 1.22.5

### What operating system are you seeing the problem on?

macOS

### Relevant log output

```shell
Puppeteer exception:

/Users/yurique/.../node_modules/puppeteer-core/lib/cjs/puppeteer/common/NetworkManager.js:347
            throw new Error('Unexpected extraInfo events for request ' + responseReceived.requestId);
                  ^

Error: Unexpected extraInfo events for request 8574FCDD373436FD23BAE6CB1FF3CAED
    at NetworkManager._emitResponseEvent (/Users/yurique/.../node_modules/puppeteer-core/lib/cjs/puppeteer/common/NetworkManager.js:347:19)
    at NetworkManager._onResponseReceived (/Users/yurique/.../node_modules/puppeteer-core/lib/cjs/puppeteer/common/NetworkManager.js:370:14)
    at /Users/yurique/.../node_modules/puppeteer-core/lib/cjs/vendor/mitt/src/index.js:51:62
    at Array.map (<anonymous>)
[...]

PR #7808, opened 2021-11-28
fix: suppress the error if unexpected extraInfo events are detected

OrKoN:
Issues: #7805

### DIFF
--- a/src/common/NetworkManager.ts
+++ b/src/common/NetworkManager.ts
@@ -513,7 +513,7 @@ export class NetworkManager extends EventEmitter {
 
   _emitResponseEvent(
     responseReceived: Protocol.Network.ResponseReceivedEvent,
-    extraInfo: Protocol.Network.ResponseReceivedExtraInfoEvent
+    extraInfo: Protocol.Network.ResponseReceivedExtraInfoEvent | null
   ): void {
     const request = this._requestIdToRequest.get(responseReceived.requestId);
     // FileUpload sends a response without a matching request.
@@ -523,8 +523,11 @@ export class NetworkManager extends EventEmitter {
       responseReceived.requestId
     );
     if (extraInfos.length) {
-      throw new Error(
-        'Unexpected extraInfo events for request ' + responseReceived.requestId
+      debugError(
+        new Error(
+          'Unexpected extraInfo events for request ' +
+            responseReceived.requestId
+        )
       );
     }
 


OrKoN:
Hey @josepharhar! Could you please take a look at this PR and the report #7805? I was able to reproduce the issue using the scenario from the report on Linux. It seems that a responseReceived event comes with hasExtraInfo = false but there is an extraInfo event in the cache:

```
// response
{
  requestId: '3140404.20',
  loaderId: 'ED744A4C49834781224817923ED12434',
  timestamp: 2394061.923203,
  type: 'Image',
  response: {
    url: 'https://googleads.g.doubleclick.net/adsid/google/si?gadsid=AORoGNTTfIduWIVdfg2--ZRhf5885XatMWRILAtAQH3mF9skPGi4NARWPw',
    status: 204,
    statusText: '',
    headers: {},
    mimeType: 'text/html',
    connectionReused: true,
    connectionId: 127,
    remoteIPAddress: '108.177.15.157',
    remotePort: 443,
    fromDiskCache: false,
    fromServiceWorker: false,
    fromPrefetchCache: false,
    encodedDataLength: 59,
    timing: {
      requestTime: 2394061.919692,
      proxyStart: 0.119,
      proxyEnd: 0.122,
      dnsStart: -1,
      dnsEnd: -1,
      connectStart: -1,
      connectEnd: -1,
      sslStart: -1,
      sslEnd: -1,
      workerStart: -1,
      workerReady: -1,
      workerFetchStart: -1,
      workerRespondWithSettled: -1,
      sendStart: 0.185,
      sendEnd: 0.286,
      pushStart: 0,
      pushEnd: 0,
      receiveHeadersEnd: 2.588
    },
    responseTime: 1638125871824.849,
    protocol: 'h2',
    securityState: 'unknown'
  },
  hasExtraInfo: false,
  frameId: '562B31F1971A9BEC48A040867814A468'
}
```
```
// extraInfo in the cache
{
  requestId: '3140404.20',
  blockedCookies: [],
  headers: {
    p3p: 'CP="This is not a P3P policy! See http://support.google.com/accounts/answer/151657 for more info."',
    'timing-allow-origin': '*',
    'cross-origin-resource-policy': 'cross-origin',
    'cache-control': 'private, max-age=15',
    'content-type': 'text/html; charset=UTF-8',
    'x-content-type-options': 'nosniff',
    date: 'Sun, 28 Nov 2021 18:57:51 GMT',
    server: 'cafe',
    'content-length': '0',
    'x-xss-protection': '0'
  },
  resourceIPAddressSpace: 'Public',
  statusCode: 204
}
```

josepharhar (commented):
Would it make sense to remove the thrown error instead? I'm having a hard time understanding exactly what the added `if (!extraInfo)` does...

Also, instead of throwing an error, is there any other way we can signal that something bad happened so we can fix it in chromium?

OrKoN:
we can swallow it and log using `debugError`, I guess, although I am not very familiar with how it's done in Puppeteer.

OrKoN:
I have updated the CL. So do you think that chromium is sending wrong events here?

josepharhar:
If `hasExtraInfo` is true and we aren't actually getting a `responseReceivedExtraInfo` event, then yes, it's a chromium bug. Is that the case? I can't tell from the bug report.

josepharhar (approved):

OrKoN:
@josepharhar see my comment above https://github.com/puppeteer/puppeteer/pull/7808#issuecomment-981135897 it's hasExtraInfo=false but there is an responseReceivedExtraInfo event.

Also, it seems to timeout if I ignore the error...

OrKoN:
Ah, sorry, my bad, correction: it's not timing out.

josepharhar:
> it's hasExtraInfo=false but there is an responseReceivedExtraInfo event.

Ah I see. Yeah I think the right thing to do is ignore the ExtraInfo event when `hasExtraInfo=false`, which this patch currently does. If instead we were to use an ExtraInfo event if present, then we could get flaky behavior where it would pick up the ExtraInfo event sometimes but not other times.

> Also, it seems to timeout if I ignore the error...

> Ah, sorry, my bad, correction: it's not timing out.

I'm guessing this means we're good to merge this

hanselfmu (approved):
LGTM.

PR #7824, opened 2021-12-06
docs: add additional code examples to ReadMe

adimon91:

Added code examples to ReadMe

No tests necessary

Not relevant

**Summary**

The first use case I added was how to include error handling with a try/catch/finally block. It may seem like this would be common sense to any user. However, to close a browser session using Puppeteer, you must use browser.close() and if any errors are thrown during your script the browser will not close. 

The second use case I added was how to sign in to an account using Puppeteer. This example also illustrates using Promises to wait for two promises to resolve: (1) clicking a button which will navigate away from the current page, and (2) waiting for the network to be idle.

No

**Other information**

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,69 @@ node get-dimensions.js
 
 See [`Page.evaluate()`](https://github.com/puppeteer/puppeteer/blob/v12.0.1/docs/api.md#pageevaluatepagefunction-args) for more information on `evaluate` and related methods like `evaluateOnNewDocument` and `exposeFunction`.
 
+**Example** - add error handling
+
+Save file as **error_handling.js**
+
+```js
+const puppeteer = require('puppeteer');
+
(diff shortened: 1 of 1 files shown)


OrKoN:
Thanks a lot for the PR! Our README is already quite long so I think we would not want to maintain an additional example. This kind of try/catch error handling is typical for JavaScript/NodeJS projects in general and it is commonly used to clean up allocated resources so I am not sure we want to specifically highlight this example in Puppeteer README. 

I wonder if you have seen https://github.com/puppeteer/puppeteer/tree/main/examples and https://github.com/puppeteer/examples I am not sure if there is a similar example there. If there is no example like this, feel free to send a PR adding an example to https://github.com/puppeteer/puppeteer/tree/main/examples

I will close this PR for now. Thanks again for contributing!

PR #7826, opened 2021-12-07
refactor: disallow direct access to attributes in NetworkEventManager

OrKoN:


OrKoN:
cc @benallfree

src/common/NetworkEventManager.ts
```diff
@@ -1,7 +1,7 @@
 import { Protocol } from 'devtools-protocol';
 import { HTTPRequest } from './HTTPRequest.js';
 
-export type QueuedEvents = {
+export type QueuedEventGroup = {
   responseReceived: Protocol.Network.ResponseReceivedEvent;
   promise: Promise<void>;
   resolver: () => void;
```
benallfree, line 7: From what I could find, `promise` and `resolver` are initialized but never actually awaited or referenced anywhere. I think these fields and the corresponding logic in `NetworkManager` can be removed.

src/common/NetworkEventManager.ts
```diff
@@ -1,7 +1,7 @@
 import { Protocol } from 'devtools-protocol';
 import { HTTPRequest } from './HTTPRequest.js';
 
-export type QueuedEvents = {
+export type QueuedEventGroup = {
   responseReceived: Protocol.Network.ResponseReceivedEvent;
```
benallfree: Naming idea: `responseReceivedEvent`, `loadingFinishedEvent`, and `loadingFailedEvent`. Call the whole thing a `QueuedResponseEventGroup`

benallfree (approved):
lgtm, see other notes or merge as is 👍 

Nice work. I went too far in my encapsulation attempt 😓

src/common/NetworkEventManager.ts
OrKoN: I think both are used in NetworkManager.

src/common/NetworkEventManager.ts
```diff
@@ -12,10 +12,11 @@ export type QueuedEvents = {
 export type FetchRequestId = string;
 export type NetworkRequestId = string;
 
-export type RedirectInfoMap = Array<{
+type RedirectInfo = {
   event: Protocol.Network.RequestWillBeSentEvent;
   fetchRequestId?: FetchRequestId;
-}>;
+};
+export type RedirectInfoMap = RedirectInfo[];
```
johanbay: If we are renaming types anyway, should we change the name of this type as well?
```suggestion
export type RedirectInfoList = RedirectInfo[];
```

src/common/NetworkEventManager.ts
```diff
@@ -79,12 +80,12 @@ export class NetworkEventManager {
     Protocol.Network.ResponseReceivedExtraInfoEvent[]
   >();
   private _queuedRedirectInfoMap = new Map<NetworkRequestId, RedirectInfoMap>();
```
johanbay:
```suggestion
  private _queuedRedirectInfoMap = new Map<NetworkRequestId, RedirectInfoList>();
```

johanbay (approved):
LGTM

PR #7828, opened 2021-12-07
fix(types): ensure compiler directives are at top

johanbay:
The workaround to make Puppeteers types work without DOM types
unfortunately breaks tsc's module resolution, since it put a declaration
before a [triple-slash compiler directive](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html). This makes the directive
invalid and it is therefore ignored by the typescript compiler when
resolving modules.

This fix instead appends the `global.d.ts` declarations at the end of
the generated `types.d.ts` file, ensuring that any directives stays at
the top of the file.


OrKoN (approved):

mathiasbynens:
FYI @jackfranklin

jackfranklin:
LGTM!

jackfranklin (approved):

johanbay:
After thinking a bit more about this, I think we should move forward by reverting instead: https://github.com/puppeteer/puppeteer/pull/7829

PR #7843, opened 2021-12-14
fix: make sure ElementHandle.waitForSelector is evaluated in the right context

OrKoN:
So it appears that all bindings are added to the secondary world and all
evaluations are also running there. ElementHandle.evaluate is returning
handles from the main world though. Therefore, we need to be careful
and adopt handles to the right context before evaluating within
waitForSelector.


OrKoN:
Updated the PR and moved the adoption of the context element to the top-level so that no internals have to be changed: the root ElementHandle will be already in the same context.

johanbay (approved):

PR #7722, opened 2021-10-25
feat: handle unhandled promise rejections in tests

jschfflr:

PR #7846, opened 2021-12-14
fix: disable a test failing on Firefox

OrKoN:


OrKoN:
@whimboo @mathiasbynens it looks like one test started to fail in Firefox. Should I just disable it for now?

mathiasbynens (approved):

whimboo:
> @whimboo @mathiasbynens it looks like one test started to fail in Firefox. Should I just disable it for now?

Nothing has changed on our side. @OrKoN, do you have the exact failure details? I don't see any at https://github.com/puppeteer/puppeteer/commits/main. Thanks.

OrKoN:
@whimboo yes, here is the failure https://github.com/puppeteer/puppeteer/runs/4526884428?check_suite_focus=true (it's on this PR, before I added itFailsOnFirefox, so it includes the main + a newline change; originally started failing in another PR I wanted to submit).

whimboo:
Strange. We don't see that in our own CI which runs the tests on Linux. But we still haven't caught up downstream syncs for a while. Right now @juliandescottes work on [syncing to v13.0.0](https://bugzilla.mozilla.org/show_bug.cgi?id=1732958). Maybe it brings up this problem for us... Thanks for notifying us!

juliandescottes:
I haven't seen this particular test fail with the 13.0.0 sync. A disconnected browser when navigating to an image sounds like it might be site-isolation (fission) related? and could possibly behave differently depending in the OS?

Edit: In the logs I also see that we fail on the assert `Execution Context has already been set.` added by https://github.com/puppeteer/puppeteer/pull/7722 . So if that's the root cause of the failure, and if this is intermittent, it makes sense we haven't seen it yet on our CI, because this is not in the version of puppeteer we have currently vendored (v10.0.0). I'm running more tests with v13.0.0 at the moment, I'll let you know if I see the test fail on our CI as well.

PR #7856, opened 2021-12-17
chore: update prettier globs and format files

nschonni:

refactoring

**Summary**

Updated to get a clean run with prettier running across all files. Ignore some generated and test content that shouldn't be formatted

**Other information**


Kikobeats:
Thanks for this; All PRs are broken linter was added but the code is not properly formatted.

If @johanbay @mathiasbynens approve this, we can rebase the rest of PRs 🙏

OrKoN:
I think for js files we use prettier through eslint to avoid conflicts. Would it work without conflicts with eslint? Thanks a lot for the PR and I appreciate if you can rebase it if it's compatible with eslint.

nschonni:
Yes, ESLint prettier passes out to the main library, and uses the same config

nschonni:
@OrKoN gave it a rebase and added in a few ignores from a CI failure. Unfortunately the prettierignore doesn't merge with gitignore currently, so there is a bit of duplication

nschonni:
Looks like there was a dangling reference to Markdownlint, so I removed that for now

nschonni:
looks like CI is happy this time

OrKoN (approved):

PR #7860, opened 2021-12-20
chore: use a modern disable-popup-blocking flag version

Kikobeats:
According to https://github.com/GoogleChrome/chrome-launcher/blob/master/docs/chrome-flags-for-tools.md#performance--web-platform-behavior

> `--block-new-web-contents` is the strict version of `--disable-popup-blocking`.

Non related but `--disable-features=ScriptStreaming` is missing and chrome-launcher recommends it as optimization. Should I add it?

### DIFF
--- a/src/node/Launcher.ts
+++ b/src/node/Launcher.ts
@@ -220,7 +220,7 @@ class ChromeLauncher implements ProductLauncher {
       '--disable-features=Translate',
       '--disable-hang-monitor',
       '--disable-ipc-flooding-protection',
-      '--disable-popup-blocking',
+      '--block-new-web-contents',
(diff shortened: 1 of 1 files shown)


Kikobeats:
I'm not sure tests are failing related to the code change (I can see most of the PRs builds are fail)

Any suggestion what's wrong there? @mathiasbynens @johanbay

Kikobeats:
.cc @OrKoN 🙏

OrKoN:
From the page you linked: `--block-new-web-contents: All pop-ups and calls to window.open will fail.` I don't think it's compatible with Puppeteer as it offers some support for window.open and pop-ups.

PR #7864, opened 2021-12-21
minor: add `response.stream` method

### LINKED ISSUE #7863: HTTPResponse: add stream method
what would be desirable is a way to return the stream associated with a response rather than a buffer

```js
page.on('response', async respone => {
  const stream = response.stream()
})
```

That could be achieved using [`Network.takeResponseBodyForInterceptionAsStream`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#method-takeResponseBodyForInterceptionAsStream) CDP method in a similar way how the buffer gets.

https://github.com/puppeteer/puppeteer/blob/main/src/common/HTTPResponse.ts#L183

Kikobeats:
Closes https://github.com/puppeteer/puppeteer/issues/7863


src/common/HTTPResponse.ts
```diff
@@ -21,6 +21,8 @@ import { HTTPRequest } from './HTTPRequest.js';
 import { SecurityDetails } from './SecurityDetails.js';
 import { Protocol } from 'devtools-protocol';
 import { ProtocolError } from './Errors.js';
+import { helper } from './helper.js';
+import { Readable } from 'stream';
```
OrKoN, line 25: let's make this a type import?  E.g., `import type { Readable } from 'stream';`

src/common/HTTPResponse.ts
```diff
@@ -204,6 +207,26 @@ export class HTTPResponse {
...
+    if (!this._contentStream) {
+      const result = await this._client.send(
+        'Network.takeResponseBodyForInterceptionAsStream',
+        {
+          interceptionId: this._request._requestId,
+        }
+      );
+
+      this._contentStream = helper.getReadableFromProtocolStream(
+        this._client as any,
```
OrKoN, line 236: let's cast as a CDPSession from 'CDPSession.ts'?

OrKoN (changes requested):
Marked as changes requested.

OrKoN:

Issue #7757, opened 2021-11-09
[Bug]: page.pdf produces corrupt pdf


tjacobs3:
### Bug description

Steps to reproduce the problem:

Occasionally, we find that our PDFs are not openable by any programs.  We've narrowed the issue down to inclusion of certain images.  When these images are present, the pdf created by puppeteer is corrupt.  All image tools do not indicate that anything is wrong with the image itself so I believe this is an issue on the puppeteer side.

1. Create an `test.html` file with the following contents
```
<html>
  <head>
  </head>
  <body>
    <img src="image.jpg">
  </body>
</html>
```
In the same directory, place the attached `image.jpg`

2. Create a `save_to_pdf.js` file with the following contents
```
const puppeteer = require('puppeteer');

(async () => {
  const browser = await puppeteer.launch({ headless: true });
  const page = await browser.newPage();
  await page.goto(`file://${__dirname}/test.html`, { waitUntil: 'networkidle0', timeout: 60000 });
  await page.pdf({
    path: 'out.pdf',
    printBackground: true
  });

  await browser.close();
})();
```
3. Run `node save_to_pdf.js`
4. Try to open `out.pdf` in any program.
![Screen Shot 2021-11-09 at 10 22 02 AM](https://user-images.githubusercontent.com/222549/140952303-5057a4a6-dad1-4fa6-b28a-1ce5013c3efc.png)

[puppeteer_bug.zip](https://github.com/puppeteer/puppeteer/files/7505919/puppeteer_bug.zip)
[image.jpg](https://user-images.githubusercontent.com/222549/140952636-5a5bdf4b-ec67-4b9b-ae25-3ba56a298da3.jpg)

### Puppeteer version

10.1.0

### Node.js version

12.22.5

### npm version

6.14.14

### What operating system are you seeing the problem on?

macOS

MartinFalatic:
I'm tracking a very similar issue with headless page `pdf()` calls creating corrupted files with certain input images, which started after puppeteer 10.0.0, and have been working on a way to reproduce it. I've been assuming it was the version of Chrome that was to blame somehow.... but both 10.0.0 and 10.1.0 reference Chrome `884014`.

There's [not a ton of delta between puppeteer 10.0.0 and 10.1.0](https://github.com/puppeteer/puppeteer/compare/v10.0.0...v10.1.0), but the answer is in there somewhere.

[...]

PR #7868, opened 2021-12-22
fix: page.pdf producing an invalid pdf

tjacobs3:
Bugfix

**Summary**
When defining a chunk size for `<CDPSession>.send('IO.read', { handle, size })`, the `CDPSession` will occassionally indicate that it has reached the end of file without sending a full pdf. This is documented by the associated issue.

This behavior is not reproducible when leaving out the size parameter. Since the size parameter is not required on the CDPSession side and is merely a suggestion on the stream side, we can safely leave it out.

Refs: #7757

No


OrKoN:
Thanks for the fix. I'd really appreciate if you add a failing test case that is fixed by this.

OrKoN:
Disregard my previous comment, I think we can land it without new test cases. Could you please add a TODO like `TODO: bring the size parameter back once Chromium issue crbug.com/1290727 is resolved.`

OrKoN (approved):

OrKoN:
I will add a TODO myself in a follow-up PR.

MartinFalatic:
@OrKoN Thank you for expediting this, and for tracking down a potential root cause in Chromium itself!

@tjacobs3 Thanks for the effective workaround to keep puppeteer working properly while we await the fix in Chromium!

PR #7876, opened 2021-12-26
Feature: Deregister scripts that were added with evaluateOnNewDocument

SmartArray:
## Deregister scripts that were added with evaluateOnNewDocument

A simple but yet commonly requested **feature** to deregister scripts that were added with `evaluateOnNewDocument `.

No. This was manually tested by using `alert` statements though.

Yes. The return type of the corresponding method was updated, along with the documentation for `evaluateOnNewDocument`.

**Summary**
Previously, there was no way to deregister scripts that were added with `evaluateOnNewDocument`. The underlying DevTools Protocol offers this functionality so I decided to integrate this simple change.

Not at all. Other integrations should still work as expected.
Typescript people may have to update their code as the promise object's type info was changed.

Merry Christmas btw! 🎄


src/common/Page.ts
```diff
@@ -2582,11 +2584,19 @@ export class Page extends EventEmitter {
   async evaluateOnNewDocument(
     pageFunction: Function | string,
     ...args: unknown[]
-  ): Promise<void> {
+  ): Promise<() => Promise<void>> {
     const source = helper.evaluationString(pageFunction, ...args);
-    await this._client.send('Page.addScriptToEvaluateOnNewDocument', {
-      source,
-    });
+    const { identifier } = await this._client.send(
```
OrKoN, line 2615: let's return the identified in evaluateOnNewDocument and create a symmetrical function `stopEvaluateOnNewDocument` (perhaps you have a better name?) This would make it more composable.

We would also need some tests that both functions work.

OrKoN (changes requested):
Marking as requiring changes (see my comment inline)

OrKoN:

PR #7883, opened 2022-01-01
chore: remove console.error

Junyan:
**Summary**

I think it is better to handle error by user instead of console.error.


OrKoN:
Does it make sense to use debugError instead of console.error? So that the error can be still seen in the debug log if it's not handled by the user.

OrKoN:
I went ahead and changed the PR to use debugError. Reasoning: we are not sure if the user will always handle the error properly and would like to have some indication of the error if Puppeteer is run in debug mode.

OrKoN (approved):

PR #7885, opened 2022-01-01
feat: add support for async waitForTarget

tmkx:

feature

yes
yes
**Summary**
sometimes we need to call `waitForTarget` in async:
```javascript
const target = await browser.waitForTarget((target) =>
  target
    .page()
    .then((page) => page.title())
    .then((title) => title === 'expected')
);
```

no

**Other information**

### DIFF
--- a/experimental/puppeteer-firefox/lib/Browser.js
+++ b/experimental/puppeteer-firefox/lib/Browser.js
@@ -122,23 +122,34 @@ class Browser extends EventEmitter {
     const {
       timeout = 30000
     } = options;
-    let existingTarget;
-    for (const target of this.targets()) {
-      if (await predicate(target)) {
-        existingTarget = target;
-        break;
-      }
-    }
-    if (existingTarget)
-      return existingTarget;
+    let remainingTimeout = timeout;
+    const existingTarget = await helper.waitWithTimeout(
+      (async () => {
+        const startTime = Date.now();
+        let existingTarget;
+        for (const target of this.targets()) {
+          if (await predicate(target)) {
+            existingTarget = target;
+            break;
+          }
+        }
+        if (timeout) {
+          remainingTimeout = Math.max(timeout - (Date.now() - startTime), 1);
+        }
+        if (existingTarget) return existingTarget;
+      })(),
+      'predicate',
+      timeout
+    );
+    if (existingTarget) return existingTarget;
     let resolve;
     const targetPromise = new Promise(x => resolve = x);
     this.on(Events.Browser.TargetCreated, check);
     this.on('targetchanged', check);
     try {
-      if (!timeout)
+      if (!remainingTimeout)
         return await targetPromise;
-      return await helper.waitWithTimeout(targetPromise, 'target', timeout);
+      return await helper.waitWithTimeout(targetPromise, 'target', remainingTimeout);
     } finally {
       this.removeListener(Events.Browser.TargetCreated, check);
       this.removeListener('targetchanged', check);
--- a/experimental/puppeteer-firefox/lib/Browser.js
+++ b/experimental/puppeteer-firefox/lib/Browser.js
@@ -122,34 +122,28 @@ class Browser extends EventEmitter {
     const {
       timeout = 30000
     } = options;
-    let remainingTimeout = timeout;
-    const existingTarget = await helper.waitWithTimeout(
-      (async () => {
-        const startTime = Date.now();
-        let existingTarget;
-        for (const target of this.targets()) {
-          if (await predicate(target)) {
-            existingTarget = target;
-            break;
-          }
-        }
-        if (timeout) {
-          remainingTimeout = Math.max(timeout - (Date.now() - startTime), 1);
-        }
-        if (existingTarget) return existingTarget;
-      })(),
-      'predicate',
-      timeout
-    );
-    if (existingTarget) return existingTarget;
     let resolve;
     const targetPromise = new Promise(x => resolve = x);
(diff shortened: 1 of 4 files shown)


OrKoN:
Does it make sense to use debugError instead of console.error?

tmkx:
@OrKoN Hello, I do not understand, what does this have to do with `debugError` and `console.error`?

OrKoN:
Apologies, ignore that, I intended to leave that comment on https://github.com/puppeteer/puppeteer/pull/7883

OrKoN:
The timeout setting is not respected if the callback never resolves or takes a long time to resolve. Could you address that? also, the firefox-related failures probably require `itFailsFirefox` instead of `it`.

tmkx:
Sorry, the test was copied from this one, so it should use `itFailsFirefox` also. Updated. https://github.com/puppeteer/puppeteer/blob/921e4e6240f2dd046e0b98614b35dc73c1a76868/test/target.spec.ts#L70-L86

experimental/puppeteer-firefox/lib/Browser.js
```diff
@@ -114,15 +114,21 @@ class Browser extends EventEmitter {
...
    * @return {!Promise<!Target>}
    */
   async waitForTarget(predicate, options = {}) {
     const {
       timeout = 30000
     } = options;
-    const existingTarget = this.targets().find(predicate);
+    let existingTarget;
+    for (const target of this.targets()) {
+      if (await predicate(target)) {
```
OrKoN: Could we also ensure that if predicate takes longer that the timeout, we abort the waitForTarget call?

OrKoN (changes requested):

experimental/puppeteer-firefox/lib/Browser.js
tmkx: good advice

src/common/Browser.ts
```diff
@@ -539,30 +539,48 @@ export class Browser extends EventEmitter {
...
     options: WaitForTargetOptions = {}
   ): Promise<Target> {
     const { timeout = 30000 } = options;
-    const existingTarget = this.targets().find(predicate);
+    let remainingTimeout = timeout;
+    const existingTarget = await helper.waitWithTimeout<Target | undefined>(
+      (async () => {
+        const startTime = Date.now();
+        let existingTarget: Target | undefined;
+        for (const target of this.targets()) {
```
OrKoN: I am thinking about the following scenario:

1) there is one target existing but it does not match the predicate. The predicate check is async and take 1 sec.
2) while the predicate check is running a new target gets created the matches the predicate.
3) the predicate from step 1 resolves but the local this.targets() does not contain the target created in step 1 because `this.targets()` was called before the target was created.
4) since no existing target is found the function sets up TargetCreated/TargetChanged listeners but they never fire because the target is already created. The entire waitForTarget times out although a matching target is available.

WDYT?

src/common/Browser.ts
```diff
@@ -539,20 +539,31 @@ export class Browser extends EventEmitter {
...
       if (!timeout) return await targetPromise;
       return await helper.waitWithTimeout<Target>(
-        targetPromise,
+        Promise.race([
+          targetPromise,
+          (async () => {
+            let existingTarget: Target | undefined;
+            for (const target of this.targets()) {
+              if (await predicate(target)) {
+                existingTarget = target;
```
OrKoN: can we directly return target here?

src/common/Browser.ts
tmkx: I moved the existing targets check to the same time as TargetCreated/TargetChanged.

test/target.spec.ts
```diff
@@ -67,6 +67,27 @@ describe('Target', function () {
...
+                page.url() === server.CROSS_PROCESS_PREFIX + '/empty.html'
+            )
+        )
+        .then((target) => target.page()),
+      page.evaluate(
+        (url: string) => window.open(url),
+        server.CROSS_PROCESS_PREFIX + '/empty.html'
+      ),
+    ]);
+    expect(otherPage.url()).toContain(server.CROSS_PROCESS_PREFIX);
```
OrKoN: should we check for an exact URL match? Also I wonder if the `page` has the same URL as the `otherPage`? Can we make sure that we are waiting for the right target and they have distinct URLs?

OrKoN: e.g., perhaps just `expect(page).not.toEqual(otherPage)`

src/common/Browser.ts
```diff
@@ -539,20 +539,28 @@ export class Browser extends EventEmitter {
...
     this.on(BrowserEmittedEvents.TargetCreated, check);
     this.on(BrowserEmittedEvents.TargetChanged, check);
     try {
       if (!timeout) return await targetPromise;
       return await helper.waitWithTimeout<Target>(
-        targetPromise,
+        Promise.race([
+          targetPromise,
+          (async () => {
+            for (const target of this.targets()) {
```
OrKoN, line 556: ugh, additional edge cases come to mind. WDYT about the following: we start waiting for a target and start checking existing targets one by one (and say there is a hundred of targets), then a matching target gets created and we resolve the waitForTarget with it. Although we resolved the inner promise continues executing and keeps checking the remaining targets one by one. I think that should be addressed by checking if the waitWithTimeout has been resolved and returning earlier.

OrKoN: I think let's leave this for a follow-up (seems to be rather an edge case)

OrKoN (approved):

PR #7546, opened 2021-09-05
feat: add support for Apple Silicon chromium builds

7rulnik:
Google has published [Chromium builds for Apple Silicon](https://commondatastorage.googleapis.com/chromium-browser-snapshots/index.html?prefix=Mac_Arm/) so we can fetch it now

Related to https://github.com/puppeteer/puppeteer/issues/6622

PR #7895, opened 2022-01-09
feat: support Apple M1 Chromium builds

fengmk2:

feature

Github Action not support yer https://github.com/actions/runner/issues/805
I run unit test on my local M1 machine.

**Summary**

**Other information**
Firefix is still using the x64 build for Apple M1

### DIFF
--- a/src/node/BrowserFetcher.ts
+++ b/src/node/BrowserFetcher.ts
@@ -41,12 +41,14 @@ const downloadURLs = {
   chrome: {
     linux: '%s/chromium-browser-snapshots/Linux_x64/%d/%s.zip',
     mac: '%s/chromium-browser-snapshots/Mac/%d/%s.zip',
+    macArm: '%s/chromium-browser-snapshots/Mac_Arm/%d/%s.zip',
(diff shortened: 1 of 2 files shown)


OrKoN:
@mathiasbynens should we add Apple M1 builds before GitHub actions support it?

mathiasbynens:
I worry that this change would make it even harder to find a suitable Chromium revision across all platforms, since this increases the number of different platforms. Speaking of which, would we also need to update https://github.com/puppeteer/puppeteer/blob/main/utils/check_availability.js? 

> @mathiasbynens should we add Apple M1 builds before GitHub actions support it?

Good point. I don’t think we should ship this until we have a solid testing story. We could land this **as an opt-in feature**, but then I’d like us to also do the following:

- explicitly add a TODO to our main Actions workflow, pointing to https://github.com/actions/runner/issues/805
- update the documentation to state that this support is experimental

OrKoN:
Let's close this PR in favour of https://github.com/puppeteer/puppeteer/pull/7546 which seems to include changes to the check availability script as well.

Issue #7896, opened 2022-01-13
[Bug]: Exception raised when trying to use puppeteer to close a tab which is opened manually.


findwater:
### Bug description

Steps to reproduce the problem:

1. Use puppeteer to launch chrome(97.0.4692.71)
2. Add a new tab MANUALLY 
3. Use puppeteer to close the tab which is added manually
4. Exception raised complaining a frame object is undefined. 

The code to reproduce the issue is attached. It will
1. Launch chrome
2. Wait for 10 seconds
3. Add a new tab manually 
4. After 10 seconds, close the newly added tab and you will see some errors in console.
[TestPuppeteerCore.zip](https://github.com/puppeteer/puppeteer/files/7860294/TestPuppeteerCore.zip)

A video is attached to showcase the issue.
https://user-images.githubusercontent.com/8121062/149273899-6fd1f271-2a0f-44f7-8730-ef5e71c33754.mp4

PS: This issue is NOT reproduced with puppeteer 10.2.0. It can be reproduce with 11.0.0, 12.0.0, 13.0.1

### Puppeteer version

12.0.0

### Node.js version

16.13.0

### npm version

8.1.0

### What operating system are you seeing the problem on?

Windows

### Relevant log output

```shell
TypeError: Cannot read properties of undefined (reading '_updateClient')
    at FrameManager._onAttachedToTarget (D:\MyProjects\NodeJs\TestPuppeteerCore\node_modules\puppeteer-core\lib\cjs\puppeteer\common\FrameManager.js:183:15)
    at D:\MyProjects\NodeJs\TestPuppeteerCore\node_modules\puppeteer-core\lib\cjs\puppeteer\common\FrameManager.js:88:18
    at D:\MyProjects\NodeJs\TestPuppeteerCore\node_modules\puppeteer-core\lib\cjs\vendor\mitt\src\index.js:51:62
    at Array.map (<anonymous>)
    at Object.emit (D:\MyProjects\NodeJs\TestPuppeteerCore\node_modules\puppeteer-core\lib\cjs\vendor\mitt\src\index.js:51:43)
    at CDPSession.emit (D:\MyProjects\NodeJs\TestPuppeteerCore\node_modules\puppeteer-core\lib\cjs\puppeteer\common\EventEmitter.js:72:22)
[...]

Issue #7844, opened 2021-12-14
[Bug]: TypeError: Cannot read properties of undefined (reading '_updateClient')


vmirage:
### Bug description

Steps to reproduce the problem for puppeteer-core:

1. Make sure https://c2c.huobi.com/en-us/one-trade/buy tab is already open
2. Use puppeteer-core to connect to existing chrome
3. Get the page on `targetcreated`. 
4. Boom! error

```
      const browser = await puppeteer.connect({
        browserURL: 'http://127.0.0.1:9222',
        defaultViewport: null,
      });

      browser.on('targetcreated', async (e) => {
        try {
          let type = await e.type();
          if (type != 'page') {
            return;
          }
          let page = await e.page();
        } catch (e) {
          console.error(e);
        }
      });
```

No problem on version 10. Problem started since version 11 onwards

### Puppeteer version

13.0.0

### Node.js version

17.2.0

### npm version

8.3.0

### What operating system are you seeing the problem on?

macOS

### Relevant log output

```shell
TypeError: Cannot read properties of undefined (reading '_updateClient')
    at FrameManager._onAttachedToTarget (/Volumes/data_ssd/test/node_modules/puppeteer-core/src/common/FrameManager.ts:260:11)
    at /Volumes/data_ssd/test/node_modules/puppeteer-core/src/common/FrameManager.ts:127:12
    at /Volumes/data_ssd/test/node_modules/puppeteer-core/vendor/mitt/src/index.ts:88:75
    at Array.map (<anonymous>)
    at Object.emit (/Volumes/data_ssd/test/node_modules/puppeteer-core/vendor/mitt/src/index.ts:88:56)
    at CDPSession.emit (/Volumes/data_ssd/test/node_modules/puppeteer-core/src/common/EventEmitter.ts:102:18)
    at CDPSession._onMessage (/Volumes/data_ssd/test/node_modules/puppeteer-core/src/common/Connection.ts:320:12)
    at Connection._onMessage (/Volumes/data_ssd/test/node_modules/puppeteer-core/src/common/Connection.ts:153:28)
    at WebSocket.<anonymous> (/Volumes/data_ssd/test/node_modules/puppeteer-core/src/node/NodeWebSocketTransport.ts:47:42)
    at WebSocket.onMessage (/Volumes/data_ssd/test/node_modules/ws/lib/event-target.js:199:18)
[ERROR] 06:01:32 TypeError: Cannot read properties of undefined (reading '_updateClient')
```

[...]

PR #7899, opened 2022-01-14
fix: error for pre-existing OOPIFs

OrKoN:
Issues: #7844, #7896


mathiasbynens (approved):

mathiasbynens:
Does this need a test?

OrKoN:
I have not figured out if we can connect to an existing browser instance in Puppeteer tests? I think it'd be good to add a test if I can find a straightforward way. In general, this should have been a type error in TS.

OrKoN:
Figured out a test, PTAL

Issue #7849, opened 2021-12-15
[Bug]: ElementHandle.click() in cross-origin iframes is off target in non-headless mode


Mr0grog:
### Bug description

This is most likely an issue in Chromium, but I figured I’d file here first in case folks here know whether it’s actually intended behavior that Puppeteer needs to deal with or if I’m misunderstanding the issue.

In Puppeteer 12.0.0 (Chromium 97.0.4692.0), the `DOM.getContentQuads` and `DOM.getBoxModel` protocol methods (and maybe others) seem to behave differently in headless vs. non-headless mode: in headless mode; they return coordinates relative to the main frame’s viewport (as they used to in both modes), but in non-headless mode, they return coordinates relative to the frame the object they are called with is in.

This leads to a situation where `click()` works in headless mode, but not in non-headless mode (because it’s clicking in the wrong place), and where methods like `clickablePoint()` or `boundingBox()` return incorrect coordinates in non-headless mode.

Steps to reproduce the problem:

1. Load a page structured like:

    ```html
    <html>
      <body>
        <iframe id="frame" width="300" height="300" style="position: absolute; left: 50px, top: 50px;">
          # contentDocument
            <html>
              <body>
                <div id="inner-content" style="position: absolute; left: 50px, top: 50px;">Hello</div>
              </body>
            </html>
        </iframe>
      </body>
    </html>
    ```

2. Try to get the clickable point of `#inner-content`:

    ```js
    const frameElement = await page.$("#frame");
    const frame = await frameElement.contentFrame();
    const innerContent = await frame.$("#inner-content");
    const clickPoint = await innerContent.clickablePoint({ x: 0, y: 0 });
    console.log("Point:", clickpoint);
    ```

Doing the above in headless mode and non-headless mode should log the same coordinate, but does not.

### Puppeteer version

12.0.0

### Node.js version

16.13.0

### npm version

8.1.3

### What operating system are you seeing the problem on?

macOS

Mr0grog:
I was able to work around this by calculating the correct position using `.evaluate()`, but this is kind of complicated:

```js
const getBoundsInFrame = (node) => node.getBoundingClientRect().toJSON();
const innerBox = await innerContent.evaluate(getBoundsInFrame);
const frameBox = await frameElement.evaluate(getBoundsInFrame);
const clickPoint = { x: frameBox.x + innerBox.x, y: frameBox.y + innerBox.y };
page.click("body", { offset: clickPoint });
```

[...]

PR #7900, opened 2022-01-14
fix: correctly compute clickable points for elements inside OOPIFs

OrKoN:
Issues: #7849


Mr0grog:
I think this needs to be handled in `ElementHandle._getBoxModel()`, too, since I was also encountering the issue with `.boxModel()` and `.boundingBox()` (which both use it), and it’s a totally separate code path from `ElementHandle.clickablePoint()`.

mathiasbynens (approved):

PR #7904, opened 2022-01-16
feat: add more models to DeviceDescriptors

nswbmw:
- iPad (gen 6)
- iPad (gen 7)
- iPad Pro 11
- iPhone 12
- iPhone 12 Pro
- iPhone 12 Pro Max
- iPhone 12 Mini
- iPhone 13
- iPhone 13 Pro
- iPhone 13 Pro Max
- iPhone 13 Mini
- Pixel 4a (5G)
- Pixel 5
- Moto G4

Sync from [playwright](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/deviceDescriptorsSource.json)


LICENSE
```diff
@@ -1,4 +1,3 @@
-
```
johanbay: I think having this blank line here is rather standard? Anyway, changing the LICENSE file (even for whitespace) seems like a change that needs a separate PR, so I think we should revert this file here.

johanbay (changes requested):
Thank you for the contribution! The changes to DeviceDescriptors look good to me, but please revert the LICENSE file before we can merge.

LICENSE
nswbmw: reverted

johanbay (approved):
LGTM. Thank you!

PR #7907, opened 2022-01-17
feat(chromium): roll to Chromium 98.0.4758.0 (r950341)

OrKoN:


OrKoN:
@mathiasbynens PTAL. There are some changes in the blur filter for the image deficiency but the result is still a blurry image so I assume it's safe to just update the screenshot.

mathiasbynens (approved):
The image change is interesting indeed, but it seems acceptable.

PR #7914, opened 2022-01-18
Enable type errors for $x returned array

karlhorky:
First of all, thanks for all of your continued effort on Puppeteer, really amazing project!

Bugfix - if `page.$x` returns an empty array, throw type error when runtime error possible.

For example, the runtime error below would also become a type error:

```js
// Non-existent button selected below, causing runtime error but no type error
const [button] = await page.$x("//button[text()='non-existent']");
button.click(); // 💥 TypeError: Cannot read properties of undefined (reading 'click')
```

It would require code to guard against `undefined`:

```typescript
// Non-existent button selected below, causing runtime error but no type error
const [button] = await page.$x("//button[text()='non-existent']");
if (button) {
  button.click(); // ✅
}
```

No, but I can update tests if necessary

I'm hoping the documentation is auto-generated from the TS types, but if necessary I can update the docs manually.

**Summary**

Motivation is described above.

Yes, this is a breaking change for the types.

**Other information**

### DIFF
--- a/src/common/JSHandle.ts
+++ b/src/common/JSHandle.ts
@@ -1103,7 +1103,7 @@ export class ElementHandle<
    * If there are no such elements, the method will resolve to an empty array.
    * @param expression - Expression to {@link https://developer.mozilla.org/en-US/docs/Web/API/Document/evaluate | evaluate}
    */
-  async $x(expression: string): Promise<ElementHandle[]> {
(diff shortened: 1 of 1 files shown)


karlhorky:
Ahh conventional commits :( Ok, opened a new PR: #7915

Issue #6769, opened 2021-01-21
Consider using Typescript strict flag


danilofuchs:
Great job on the [Typescript Migration](https://developers.google.com/web/updates/2021/01/puppeteer-typescript)!

`strict` is one of the most useful Typescript flags to catch `cannot access key of undefined` errors early on.

## Flag documentation
https://www.typescriptlang.org/tsconfig#strict

## Migration process

Having worked on migrating a big non-strict project to strict mode, I know it is quite the undertaking. Typescript does not allow to apply a flag to a single file, so enabling it directly would mean porting the entire codebase at once, which is very error-prone.

However, I have used [`ts-strictify`](https://github.com/cschroeter/ts-strictify) to run an additional CI check if enabling `strict` flag would break any **changed files**. This way, we can be sure any new code is strict.

jackfranklin:
Would definitely like to do this and it's unfortunate we can't turn it on file by file.

I wonder if rather than immediately leaping to strict: true, if we could turn on all the flags that `strict: true` implies one-by-one in PRs? That way we might not have to fix ~400 errors all at once and might be able to bring it down.

OrKoN:
We could turn the flag on locally and fix issues, then submit those patches with strict: false. In the end, we'd hopefully only need to fix a few regressions in the final PR that turns on strict true. We can also go by individual components but I am afraid most of errors are about getting data from hash maps.

PR #7915, opened 2022-01-18
fix(types): enable type errors w. $x() empty array

karlhorky:
First of all, thanks for all of your continued effort on Puppeteer, really amazing project!

Bugfix - if `page.$x` returns an empty array, throw type error when runtime error possible.

For example, the runtime error below would also become a type error:

```js
// Non-existent button selected below, causing runtime error but no type error
const [button] = await page.$x("//button[text()='non-existent']");
button.click(); // 💥 TypeError: Cannot read properties of undefined (reading 'click')
```

It would require code to guard against `undefined`:

```typescript
const [button] = await page.$x("//button[text()='non-existent']");
if (button) {
  button.click(); // ✅
}
```

No, but I can update tests if necessary

I'm hoping the documentation is auto-generated from the TS types, but if necessary I can update the docs manually.

**Summary**

Motivation is described above.

Yes, this is a breaking change for the types.

**Other information**

**Alternatives considered**

A workaround that we've explored for now (although not as nice as the destructuring in the original post above):

```ts
const buttons = await page.$x("//button[text()='non-existent']");
if (buttons.length > 0) {
  buttons[0].click(); // ✅
}
```

The reason why the destructuring + simple guard at the top of this post doesn't work for us is because of [`@typescript-eslint/no-unnecessary-condition`](https://typescript-eslint.io/rules/no-unnecessary-condition/), which recognizes this as an unnecessary condition:

Screen Shot 2022-01-18 at 12 08 20

This is because TypeScript will return the type of `ElementHandle<Element>` for the first destructured element, which will always be truthy.

### DIFF
--- a/src/common/JSHandle.ts
+++ b/src/common/JSHandle.ts
@@ -1103,7 +1103,7 @@ export class ElementHandle<
    * If there are no such elements, the method will resolve to an empty array.
    * @param expression - Expression to {@link https://developer.mozilla.org/en-US/docs/Web/API/Document/evaluate | evaluate}
    */
-  async $x(expression: string): Promise<ElementHandle[]> {
+  async $x(expression: string): Promise<(ElementHandle | undefined)[]> {
     const arrayHandle = await this.evaluateHandle(
       (element: Document, expression: string) => {
         const document = element.ownerDocument || element;


OrKoN:
Hey, thanks for PR! I think you want to use `--noUncheckedIndexedAccess` TS flag (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#checked-indexed-accesses---nouncheckedindexedaccess) instead of changing the types here. The proposed change implies that the following `[undefined, undefined]` is a valid response from Puppeteer but it's not the case.

Also, I have created an example in TS playground that demonstrates how the flag checks for array index access: https://www.typescriptlang.org/play?&noUncheckedIndexedAccess=true#code/KYDwDg9gTgLgBAYwDYEMDOa4EE4G8BQcRcAJhABQCUehxdCEAdmhEsAHRIQDm5A5GUbA+lANy0iAX3zT86AJ6MEcAGYBXJTACWTODGBoYVAFxwAClAgBbLWmAAeLFCgp5jgHzua9JobhQDNSR4AF44AG0AXXEJf2AYNShGCOiZfHwGZnhw4CRIuDCUAHcULXh9QypxXPYyKrgAega4ABV5MGBEAAtgBABrdK0VOHJc6gI6GrqxGSA

We currently not even running TypeScript in the strict mode but we should also enable this flag for Puppeteer itself after we finish the strict mode support: https://github.com/puppeteer/puppeteer/issues/6769 (cc @jrandolf)

karlhorky:
> I think you want to use `--noUncheckedIndexedAccess` TS flag

Ah yes indeed! I remember reading these release notes, but somehow missed how this could be useful to us

I'll turn it on, thanks!

Sorry for the PR noise!

PR #7919, opened 2022-01-21
fix: compatibility with strict mode Typescript in Connection.ts and ConnectionTransport.ts

OrKoN:
Issues: #6769


jrandolf-zz (approved):
LGTM.

PR #7927, opened 2022-01-25
chore: generate version range for deprecated versions

OrKoN:


versions.js
```diff
@@ -42,4 +42,15 @@ const versionsPerRelease = new Map([
   ['73.0.3679.0', 'v1.12.2'],
 ]);
 
-module.exports = versionsPerRelease;
+const lastMaintainedChromiumVersion = '97.0.4692.0';
```
mathiasbynens, line 46: Let’s add a comment here explaining which version this is supposed to refer to, i.e. the same major version as the current Chrome Stable per https://chromestatus.com/roadmap.

mathiasbynens (approved):
LGTM % suggestion

mathiasbynens:
I think this is `chore:` rather than `feat:`, since it doesn’t add a new Puppeteer API or feature.

versions.js
OrKoN: Done, PTAL

mathiasbynens (approved):

PR #7929, opened 2022-01-26
chore: apply deprecation command

OrKoN:


OrKoN:
@mathiasbynens should we land this? Can we test it somehow? I am not 100% sure it will work.

mathiasbynens (approved):

mathiasbynens:
> @mathiasbynens should we land this? Can we test it somehow? I am not 100% sure it will work.

Will discuss offline.

Issue #7085, opened 2021-04-13
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined


haidang666:
**Tell us about your environment:**

* Puppeteer version: 8.0.0
* Platform / OS version: macos
* URLs (if applicable): any urls
* Node.js version: 12.21.0

**What steps will reproduce the problem?**

_Please include code that reproduces the issue._

This is related to tsconfig or webpack, cos it works without ts.
The way I use puppeteer
```node
async function getData(url, filePath) {
  const browser = await puppeteer.launch({
    headless: true,
  });
  const page = await browser.newPage();
  await page.goto(url, { waitUntil: 'networkidle0' });
}
```

**What is the expected result?**
page loaded

**What happens instead?**
in electron
```node
index.js:118 TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at validateString (internal/validators.js:117)
    at Object.join (path.js:1039)
    at new BrowserFetcher (BrowserFetcher.js:161)
    at resolveExecutablePath (Launcher.js:441)
    at ChromeLauncher.launch (Launcher.js:78)
    at async getData (src.js:92)
```
in console
```unix
@nuclear/main:     main › renderer › TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
@nuclear/main:     at validateString (internal/validators.js:117:11)
@nuclear/main:     at Object.join (path.js:1039:7)
@nuclear/main:     at new BrowserFetcher (webpack:///./node_modules/@nuclear/core/node_modules/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js?:161:22)
@nuclear/main:     at resolveExecutablePath (webpack:///./node_modules/@nuclear/core/node_modules/puppeteer/lib/cjs/puppeteer/node/Launcher.js?:441:28)
@nuclear/main:     at ChromeLauncher.launch (webpack:///./node_modules/@nuclear/core/node_modules/puppeteer/lib/cjs/puppeteer/node/Launcher.js?:78:57)
@nuclear/main:     at async getData (webpack:///./node_modules/@nuclear/core/node_modules/extract-sptf-playlist/src.js?:92:19)
```

YTvW:
did you manage to fix this? I am running into this issue as well

haidang666:
No I didn't, I don't use it with `ts`

[...]

PR #7933, opened 2022-01-26
fix(initialize-node): __dirname in some envrioments

### LINKED ISSUE #7932: [Bug]: __dirname points to root
### Bug description

1. In a project that uses webpack and typescript, depending on configuration of other node/webpack items, __dirname points to `/`. This results in a TypeError tracing back to the Browser Launch. 

Further investigations show that initializePuppeteerNode isn't setting puppeteerRootDirectory. 

The fix for this is to fall back to process.cwd() when pkgDir.sync(__dirname) comes back undefined.

### Puppeteer version

13.1.2

### Node.js version

14.18.0 & 16

### npm version

6.14.15

Linux, macOS, Windows

### Relevant log output

```shell
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received null
     at new NodeError (internal/errors.js:322:7)
     at validateString (internal/validators.js:124:11)
     at Object.join (path.js:1148:7)
     at new BrowserFetcher (webpack:////Users/codydaig/dev/opensource/puppeteer/lib/cjs/puppeteer/node/BrowserFetcher.js?:162:22)
     at resolveExecutablePath (webpack:////Users/codydaig/dev/opensource/puppeteer/lib/cjs/puppeteer/node/Launcher.js?:610:28)
     at ChromeLauncher.launch (webpack:////Users/codydaig/dev/opensource/puppeteer/lib/cjs/puppeteer/node/Launcher.js?:107:57) {
   code: 'ERR_INVALID_ARG_TYPE'
 }
```

codydaig:
This patch fixes __dirname pointing to / in some envrioments by
trying process.cwd() if __dirname returne undefined.

Issues: #7085 #7932

Fixes #7932 where certain node environments with webpack don't place nicely to __dirname.

No. This is very difficult to reproduce in simple environments. This change only affects those where __dirname returns undefined on initialization. 

No relevant documentation updates needed.

**Summary**

When __dirname points to / it falls back to process.cwd to get the projectRoot on initialization. 

No.

### DIFF
--- a/src/initialize-node.ts
+++ b/src/initialize-node.ts
@@ -20,7 +20,8 @@ import pkgDir from 'pkg-dir';
 import { Product } from './common/Product.js';
 
 export const initializePuppeteerNode = (packageName: string): PuppeteerNode => {
-  const puppeteerRootDirectory = pkgDir.sync(__dirname);
+  const puppeteerRootDirectory =
+    pkgDir.sync(__dirname) || pkgDir.sync(process.cwd());
 
   let preferredRevision = PUPPETEER_REVISIONS.chromium;
   const isPuppeteerCore = packageName === 'puppeteer-core';


OrKoN:
@jackfranklin could you please take a look?

OrKoN:
Why does it work with `process.cwd()`? I assume that the cwd is not the Puppeteer's installation folder in node_modules, is it? In those environments, do you still rely on Puppeteer to download the browser binary?

OrKoN:
Does it work if you specify the path via `PUPPETEER_DOWNLOAD_PATH` or skip download `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD`?

codydaig:
I don't think you can just rely on `PUPPETEER_DOWNLOAD_PATH` because it's starting out by setting `puppeteerRootDirectory` to undefined (if __dirname is root) and that errors when creating a `new BrowserFetcher` passing in the project root (Launcher.ts line 808). I could be mistaken and can double check this later today. 

You are correct that process.cwd() doesn't return the puppeteer root Directory, but the directory in which they are in, but regardless, i believe they end up resolving the same after pkgDir.sync() finding the project root. 

I know with this, it was able to find the chromium install inside the node_modules folder. I can double check the Download path env variable in a bit and see if that resolves it.

OrKoN (approved):

OrKoN:
codydaig@ actually, have you tried to re-configure webpack to have the proper __dirname? https://codeburst.io/use-webpack-with-dirname-correctly-4cad3b265a92 I am a bit hesitant to merge the CL as is as it'd be installing chromium into the cwd instead of the node_modules/puppeteer folder. It'd be great if you could share a repro as I'd like to get  some better understanding of how it works.

my-lalex:
Why not use the `require.resolve` Node API to get the puppeteer package location?

OrKoN:
Opened a PR to use `require.resolve` https://github.com/puppeteer/puppeteer/pull/8003 Also, verified that it works in a webpack bundled Puppeteer version.

OrKoN:
That has now landed. I think it should work better than __dirname in Webpack environments. Let me know if it's not the case.

codydaig:
Thanks! Confirmed that it's fixed. (sorry for the delay, family stuff got in the way).

PR #7942, opened 2022-01-28
fix: make `waitForSelector` generic

### LINKED ISSUE #7295: The return type of waitForSelector is Promise<ElementHandle | null> even when hidden is false
With the new built-in TypeScript definitions, the type of the `waitForSelector` method is

```
waitForSelector(selector: string, options?: WaitForSelectorOptions): Promise<ElementHandle | null>;
```

which means that the type of the returned value is `Promise<ElementHandle | null>`. But according to the documentation, if the `hidden` option is set to `false` (which is the default), then `waitForSelector` is always going to return a non-`null` value (or throw an exception). By unconditionally including `null` in the return type, users have to either perform a check for `null`ness at runtime, or use a type assertion which reduces type safety. Both solutions are not ideal.

jespertheend:

Fixes types only

No

No

**Summary**

```ts
const imgEl = await page.waitForSelector('img');
img; // ElementHandle<Element> (used to be ElementHandle<Element> | null)
```
whereas before this change:
```ts
const imgEl = await page.waitForSelector('img');
img; // ElementHandle<Element> | null
if (!img) throw new Error()
img; // ElementHandle<Element>
```
The extra check for `null` is now no longer necessary.

No runtime changes for sure. Though users that were doing `null` checks as described above are now unnecessarily doing this. I'm not aware of any TypeScript configurations that would cause a compile error when a check like this is performed. In my brief testing I couldn't find any such situation.

**Other information**

Fixes #7295

I see `JSHandle` could benefit from the same same change. I'll try to add this as well.

### DIFF
--- a/src/common/JSHandle.ts
+++ b/src/common/JSHandle.ts
@@ -391,14 +391,14 @@ export class ElementHandle<
    * (30 seconds). Pass `0` to disable timeout. The default value can be changed
    * by using the {@link Page.setDefaultTimeout} method.
    */
-  async waitForSelector(
+  async waitForSelector<T extends boolean = false>(
     selector: string,
     options: {
       visible?: boolean;
-      hidden?: boolean;
+      hidden?: T;
       timeout?: number;
     } = {}
-  ): Promise<ElementHandle | null> {
+  ): Promise<T extends true ? null : ElementHandle> {
     const frame = this._context.frame();
     const secondaryContext = await frame._secondaryWorld.executionContext();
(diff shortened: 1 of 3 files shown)


jrandolf-zz:
This type signature is not something we'd like to support. In particular, someone could manually write true and since `hidden` can be undefined, the return type would be lying to the developer. And even if we fixed this, this would almost never be used since an options object will generally genericize the boolean attribute of hidden, so when passed into the function, it would go back to using boolean. Unless the options object manually types `true` for the `hidden` attribute, this type is meaningless.

PR #7957, opened 2022-02-02
chore: remove commit lint from pull requests

OrKoN:
The commit log is still checked post-merge. Before the merge the commit message is checked using a git hooks and conventionalcommits.org check on the PR using Github Actions.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "commitlint": "commitlint --from=HEAD~1",
     "markdownlint": "prettier --check **/README.md docs/troubleshooting.md",
     "markdownlint-fix": "prettier --write **/README.md docs/troubleshooting.md",
-    "lint": "npm run eslint && npm run build && npm run doc && npm run commitlint && npm run markdownlint",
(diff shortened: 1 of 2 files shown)


OrKoN:
@mathiasbynens @johanbay PTAL

package.json
```diff
@@ -26,7 +26,7 @@
     "commitlint": "commitlint --from=HEAD~1",
     "markdownlint": "prettier --check **/README.md docs/troubleshooting.md",
     "markdownlint-fix": "prettier --write **/README.md docs/troubleshooting.md",
-    "lint": "npm run eslint && npm run build && npm run doc && npm run commitlint && npm run markdownlint",
+    "lint": "npm run eslint && npm run build && npm run doc && npm run markdownlint",
```
johanbay, line 29: We could consider making a new target `ci-lint` or similar that skips commitlint and leave the `lint` target as is? That way, devs still runs commitlint in their local workflows per default.

johanbay (approved):

PR #7960, opened 2022-02-03
feat(chromium): roll to Chromium 99.0.4844.16 (r961656)

### LINKED ISSUE #7959: Roll Chromium M99
M99 is already in beta https://chromiumdash.appspot.com/schedule so let's roll the latest Dev build for M99 into Puppeteer.

Instructions: https://github.com/puppeteer/puppeteer/blob/main/CONTRIBUTING.md#rolling-new-chromium-version

jrandolf-zz:
Fixes #7959.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "debug": "4.3.3",
-    "devtools-protocol": "0.0.948846",
+    "devtools-protocol": "0.0.960912",
     "extract-zip": "2.0.1",
     "https-proxy-agent": "5.0.0",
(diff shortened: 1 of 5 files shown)


OrKoN:
Thanks! I will take a look tomorrow: curious where the a11y-related changes are coming from.

OrKoN:
@johanbay it looks like the queryAXTree command returns hidden buttons now? I think it should not be a problem for Recorder but do you know what have caused the change between M98 and M99.

johanbay:
I'm not aware of any changes on the CDP level so it is probably caused by changes to Blink a11y code.

OrKoN (approved):

PR #7964, opened 2022-02-04
feat(puppeteer): export esm modules in package.json

### LINKED ISSUE #6753: Ship Puppeteer as JavaScript modules

jrandolf-zz:
Fixes #6753


OrKoN:
I get the following error if I try to use ESM version for a NodeJS package:

```
/Users/alexrudenko/src/puppeteer/lib/esm/puppeteer/node.js:16
import { initializePuppeteerNode } from './initialize-node.js';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at Object.compileFunction (node:vm:352:18)
    at wrapSafe (node:internal/modules/cjs/loader:1031:15)
    at Module._compile (node:internal/modules/cjs/loader:1065:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:190:29)
    at ModuleJob.run (node:internal/modules/esm/module_job:185:25)
    at async ESMLoader.import (node:internal/modules/esm/loader:281:24)
```

The code for the test module is:

```

(async () => {
  await page.goto('http://example.com');
  await page.screenshot({ path: 'example.png' });
  await browser.close();
})();
```

mathiasbynens:
@OrKoN What's the filename of this test module? It either needs to have the `.mjs` extension (easiest) or you need to configure things in its `package.json`.

OrKoN:
@mathiasbynens it has type=module in package.json

src/initialize-node.ts
```diff
@@ -20,7 +20,7 @@ import pkgDir from 'pkg-dir';
 import { Product } from './common/Product.js';
 
 export const initializePuppeteerNode = (packageName: string): PuppeteerNode => {
-  const puppeteerRootDirectory = pkgDir.sync(__dirname);
+  const puppeteerRootDirectory = pkgDir.sync(process.cwd());
```
OrKoN: This should resolve to `node_modules/puppeteer` when the user installs it. Could you try the following?

```
import path from 'path';
import { fileURLToPath } from 'url';
path.dirname(fileURLToPath(import.meta.url))
```

package.json
```diff
@@ -38,6 +43,7 @@
     "clean-docs": "rimraf website/docs && rimraf docs-api-json",
     "generate-d-ts": "npm run clean-docs && api-extractor run --local --verbose",
     "generate-docs": "npm run generate-d-ts && api-documenter markdown -i docs-api-json -o website/docs && node utils/remove-tag.js",
+    "generate-pkg-json": "cp package.esm.json lib/esm/package.json",
```
OrKoN: nit: but perhaps `echo '{"type":"module"}' > lib/esm/package.json` would work without the need for an extra file?

src/initialize-node.ts
jrandolf-zz: This is not possible since `tsc` cannot handle `import.meta.url`. I'll just implement a preprocessor for that specific file.

package.json
```diff
@@ -1,8 +1,13 @@
 {
   "name": "puppeteer",
-  "version": "13.2.0-post",
+  "version": "13.2.0",
```
OrKoN: don't remove -post

jrandolf-zz: This seems to cause errors in the `linux-main` workflow.

docs/api.md
```diff
@@ -1,4 +1,4 @@
-# Puppeteer API <!-- GEN:version -->Tip-Of-Tree<!-- GEN:stop-->
+# Puppeteer API <!-- GEN:version -->v13.2.0<!-- GEN:stop-->
```
OrKoN: also, I think this needs to get reverted

package.json
OrKoN: hm, it should not. That is the part of our release process, and usually it works.

OrKoN: do you have a link to the error? the main workflows seem to be alright https://github.com/puppeteer/puppeteer/runs/5099856140?check_suite_focus=true

src/initialize-node.ts
```diff
@@ -18,9 +18,10 @@ import { PuppeteerNode } from './node/Puppeteer.js';
 import { PUPPETEER_REVISIONS } from './revisions.js';
 import pkgDir from 'pkg-dir';
 import { Product } from './common/Product.js';
+import { puppeteerDirname } from './compat.js';
 
 export const initializePuppeteerNode = (packageName: string): PuppeteerNode => {
-  const puppeteerRootDirectory = pkgDir.sync(__dirname);
+  const puppeteerRootDirectory = pkgDir.sync(puppeteerDirname);
```
OrKoN, line 24: pkgDir.sync resolves to the ESM folder now. I think we need to go one level up in case of ESM because of the extra package.json that we add.

package.json
```diff
@@ -29,15 +34,18 @@
     "lint": "npm run eslint && npm run build && npm run doc && npm run markdownlint",
     "doc": "node utils/doclint/cli.js",
     "clean-lib": "rimraf lib",
-    "build": "npm run tsc && npm run generate-d-ts",
-    "tsc": "npm run clean-lib && tsc --version && npm run tsc-cjs && npm run tsc-esm",
+    "build": "npm run tsc && npm run generate-d-ts && npm run generate-pkg-json",
+    "tsc": "npm run clean-lib && tsc --version && (npm run tsc-cjs & npm run tsc-esm) && (npm run tsc-compat-cjs & npm run tsc-compat-esm)",
```
OrKoN, line 38: would `&` propagate the error codes if the compilation fails?

jrandolf-zz: Yes, albeit asynchronously.

OrKoN: Is it possible to that the output from both commands would interleave/duplicate?

scripts/test-install.sh
```diff
@@ -16,6 +16,13 @@ npm install --loglevel silent "${tarball}"
 node --eval="require('puppeteer')"
 ls $TMPDIR/node_modules/puppeteer/.local-chromium/
 
+# Testing ES module features
+TMPDIR="$(mktemp -d)"
+cd $TMPDIR
+echo '{"type":"module"}' >>$TMPDIR/package.json
+npm install --loglevel silent "${tarball}"
+node --input-type="module" --eval="import puppeteer from 'puppeteer'"
```
OrKoN, line 24: what about checking the download dir?

package.json
jrandolf-zz: Yes, but I don't believe it will be significant. Also, since for a majority of the time the code compiles fine, this speeds up build times.

OrKoN: Yep, makes sense! Thanks for clarifying.

src/node/NodeWebSocketTransport.ts
```diff
@@ -16,17 +16,16 @@
 import { ConnectionTransport } from '../common/ConnectionTransport.js';
 import NodeWebSocket from 'ws';
 
+export const version = '13.2.0-post';
 export class NodeWebSocketTransport implements ConnectionTransport {
   static create(url: string): Promise<NodeWebSocketTransport> {
-    // eslint-disable-next-line @typescript-eslint/no-var-requires
-    const pkg = require('../../../../package.json');
     return new Promise((resolve, reject) => {
       const ws = new NodeWebSocket(url, [], {
```
OrKoN, line 32: can we `await import('../../../../package.json')` here and make the callback async? I think we might want allow json imports in TypeScript if it's not allowed? I'd prefer not to duplicate the version and have manually updated. Currently it's automatically managed by the release process.

jrandolf-zz: We can't because `json` imports are experimental in Node. Let's discuss this internally.

OrKoN: Could we perhaps use `JSON.parse(fs.readFileSync())`?

compat/esm/compat.ts
```diff
@@ -0,0 +1,8 @@
+import { dirname } from 'path';
+import { fileURLToPath } from 'url';
+
+// `import.meta.url` resolves to the nearest package.json which for ES modules
```
OrKoN: does `import.meta.url` resolve to the package.json? 🤔

OrKoN (approved):
LGTM! nice work on the compatibility layer!

mathiasbynens:
Nice work! Should we add some documentation to the `README`?

my-lalex:
I didn't investigate further, but the `join(puppeteerDirname, 'package.json')` from `src/node/NodeWebSocketTransport.ts` gets the wrong path to get the package.json in my setup (typescript, webpack): it return a path which is one level too high.

The relative path from '13.2' seems to work `../../../../package.json` as downgrading worked for me...

jrandolf-zz:
@my-lalex Since you are using webpack, I assume you have everything output to a single file. Could you compare the output with the previous output? Specifically the area where `puppeteerDirname` is used.

my-lalex:
I kinda find the same as in the code source.... maybe using the `__dirname` variable is not ideal, some issues are known to happen when using it with bundlers.... Why not using `resolve.require()` to find the puppeteer path?

Another point is that if all the code is bundled (including puppetteer's) - let's say for offline environment- there's no 'package.json' anymore.... wouldn't it be a good idea to store the version number in a global variable and use it after that?
To avoid updateing the variable value at each release, it could be done by importing the package.json in the typescript code and using the `resolveJsonModule` options of tsconfig... and it would prevent a JSON parsing at runtime only for a version number...

jrandolf-zz:
I am not sure what you mean by "offline environment". The code in `node` is expected to be used in a Node environment. Moreover, I believe `resolveJsonModule` only helps resolve the path. Not parsing it at compile time and embedding it. Since code can be externalized in Webpack, for now that would be our recommendation.

my-lalex:
When saying "offline environment", I'm talking about a computer having no internet connection. When this is that kind of target, you cannot use "npm install" to get your packages before running. So you (can) bundle all your code in a single file as in Node file size is less important.... and this way you just have one file to deploy, getting rid of all the `node_modules` folder... and specify a `executablePath` option

My bad about `resolveJsonModule`: bundlers include the JSON content in the code but it's not a TS feature.... 

A global variable, updated by the building process (or manually), would then be a good solution... for my part, I think that parsing a file (which might not be there) each time you need the package version isn't an optimal solution...

And if this has to be the solution by deciding that puppeteer cannot be bundled, the use of `require.resolve` would seem to me as being the most viable solution for every building solution... (actually, I already use it with puppeteer to get the full path of the chromium binary)

```typescript
let puppeteerPath = path.join(__dirname, "node_modules", "puppeteer");
// Just to be sure, depends on the environment
try {
    puppeteerPath = path.dirname(require.resolve('puppeteer'));
} catch (e) {}
```

PR #7976, opened 2022-02-09
chore(deps): update deps

jrandolf-zz:
Updates development and standard dependencies. Also replaces `node-fetch` with `cross-fetch` for compatibility.


src/common/Page.ts
```diff
@@ -1795,12 +1795,12 @@ export class Page extends EventEmitter {
    *   more than 2 network connections for at least `500` ms.
    */
   async reload(options?: WaitForOptions): Promise<HTTPResponse | null> {
-    const result = await Promise.all<HTTPResponse, void>([
+    const result = await Promise.all<HTTPResponse | void>([
```
OrKoN: was it intended to be `Promise.all<[HTTPResponse, void]>` instead? (if that compiles)

src/common/fetch.ts
```diff
@@ -1,22 +0,0 @@
...
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-import { isNode } from '../environment.js';
-
-/* Use the global version if we're in the browser, else load the node-fetch module. */
-export const getFetch = async (): Promise<typeof fetch> => {
```
OrKoN, line 20: let's keep conditional imports here as in DevTools (browser version) we cannot have any dependencies as we don't use node module resolution.

src/common/Page.ts
jrandolf-zz: This will no longer compile in new versions of typescript.

compat/cjs/compat.ts
```diff
@@ -1,3 +1,19 @@
+/**
+ * Copyright 2020 Google Inc. All rights reserved.
```
OrKoN: 2022 for new files

OrKoN (approved):
LGTM thanks!

PR #8018, opened 2022-02-14
fix: element screenshot issue in headful mode

### LINKED ISSUE #7999: [Bug]: element screenshot not vertically aligned with element
### Bug description

When selecting an element and then taking a screenshot, I consistently get screenshots that are a fair number of pixels further down the page than the element I'm targeting.

```typescript

const run = async () => {
    headless: false,
  }); 
  await page.goto('https://stackoverflow.com/');

  const header = await page.waitForSelector('body > header');
  await header?.screenshot({ path: 'header.png' });
  await browser.close();
};

run();
```

This is the page itself:

image

Given the selector, I'd expect a result like this:

image

Instead, I get something like this from the code above:

image

It's clearly just below the header.  If the element is large enough, I'll get most of the element but the top will be missing and I'll have content from below it included.  It's like everything is vertically offset by 100-200 pixels.

This is consistent across several sites I've tried.  The reproduction is intentionally contrived on a stable selector.

### Puppeteer version

13.3.0

### Node.js version

v16.13.2

### npm version

8.1.2

macOS

### Relevant log output

```shell
N/A
```

OrKoN:
Fixes: #7999

### DIFF
--- a/src/common/JSHandle.ts
+++ b/src/common/JSHandle.ts
@@ -959,7 +959,7 @@ export class ElementHandle<
     const layoutMetrics = await this._client.send('Page.getLayoutMetrics');
     // Fallback to `layoutViewport` in case of using Firefox.
     const { pageX, pageY } =
-      layoutMetrics.cssLayoutViewport || layoutMetrics.layoutViewport;
+      layoutMetrics.cssVisualViewport || layoutMetrics.layoutViewport;
 
     const clip = Object.assign({}, boundingBox);
     clip.x += pageX;


src/common/JSHandle.ts
```diff
@@ -957,9 +957,7 @@ export class ElementHandle<
     assert(boundingBox.height !== 0, 'Node has 0 height.');
 
     const layoutMetrics = await this._client.send('Page.getLayoutMetrics');
-    // Fallback to `layoutViewport` in case of using Firefox.
-    const { pageX, pageY } =
-      layoutMetrics.cssLayoutViewport || layoutMetrics.layoutViewport;
+    const { pageX, pageY } = layoutMetrics.layoutViewport;
```
sadym-chromium: As a temp workaround we can get the cssPixel / deprecatedPixel ratio and normalize the layoutViewport.pageX, layoutViewport.pageY by that ratio.

sadym-chromium: smth like 
```
const pixelRatio = layoutMetrics.cssLayoutViewport.clientWidth / layoutMetrics.layoutViewport.clientWidth;
```

sadym-chromium: Actually, according to [the implementation](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/inspector_page_agent.cc;drc=34ac5694a0cf645b70049f6c691056165d7e5b41;l=1544), it looks like we have to use `cssVisualViewport` instead. [DevTools](https://github.com/ChromeDevTools/devtools-frontend/blob/805c1e9f351c42b9a0dbc07b328e9b441105aefe/front_end/core/sdk/ScreenCaptureModel.ts#L51) uses it and screenshots in Canary DevTools works well. 

My experiment showed the `cssLayoutViewport.Page[X/Y]` shows the window's position on the screen.

sadym-chromium:
```suggestion
    const { pageX, pageY } = layoutMetrics.cssVisualViewport;
```

sadym-chromium (approved):

PR #8041, opened 2022-02-21
feat: export `Frame._client` through getter

### LINKED ISSUE #8040: [Feat]: Add getter for `client` in `Frame`

jrandolf-zz:
Fixes #8040

### DIFF
--- a/src/common/FrameManager.ts
+++ b/src/common/FrameManager.ts
@@ -702,6 +702,11 @@ export class Frame {
     );
   }
 
+  /**
+   * @remarks
+   *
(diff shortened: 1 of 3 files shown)


src/common/FrameManager.ts
```diff
@@ -792,6 +792,13 @@ export class Frame {
     return this._mainWorld.executionContext();
   }
 
+  /**
+   * @returns the CDP session associated with the frame.
+   */
```
OrKoN, line 802: let's add `@internal` and remove the API doc as I am not sure we will need it as a public API

OrKoN (approved):

Issue #5048, opened 2019-10-17
Chromium download always stuck at 99%


binarykitchen:
```
.../node_modules/puppeteer ❯❯❯ npm install
> puppeteer@1.20.0 install .../node_modules/puppeteer
> node install.js

Downloading Chromium r686378 - 110.2 Mb [====================] 99% 0.0s
```

Always stuck. On latest yarn or npm. Removed cache and all lock files, still no difference.

benjamingr:
@binarykitchen are you on windows by any chance? I've seen this happen because of eager firewalls in the past

binarykitchen:
No, on MacOS here. Firewall issue? Then it shouldn't block at 99%, but rather at the beginning of the download.

I'd double check code in `install.js` for robustness, error handling, timeouts and so on.

rndm2:
Same issue.
Did you resolve it?

binarykitchen:
Still having the same issue :(

ArunJRK:
Try installing it as tarball file

ctjlewis:
Left a `yarn install` running only to find that the Puppeteer install script had been running for 12 hrs without gracefully failing. Have tried `yarn cache clean` to no avail.

PR #8169, opened 2022-03-28
fix: chromium downloading hung at 99%

0xGaryWu:
bugfix: https://github.com/puppeteer/puppeteer/issues/5048


0xGaryWu:
checks failing is not my reason, how to fix it to work well

OrKoN:
It looks like the type `Options` does not support headers.  See https://github.com/puppeteer/puppeteer/runs/5750713940?check_suite_focus=true#step:4:61

OrKoN (approved):

Issue #5740, opened 2020-04-26
how to request.respond() multiple set-cookie headers?


ghost:
If I look at the respond code int NetworkManager.js it only accepts a single 'set-cookie' header as a string.

```javascript
  async respond(response) {
    /** @type {!Object<string, string>} */
    const responseHeaders = {};
    if (response.headers) {
      for (const header of Object.keys(response.headers))
        responseHeaders[header.toLowerCase()] = response.headers[header];
    }
    if (response.contentType)
      responseHeaders['content-type'] = response.contentType;
    if (responseBody && !('content-length' in responseHeaders))
      responseHeaders['content-length'] = String(Buffer.byteLength(responseBody));
```

I've tried
```
{headers: {'set-cookie': '1=1.;Domain=.site.com;Path=/;Expires=Tue, 26-Apr-2022 10:30:53 GMT;Priority=HIGH,2=2.;Domain=.site.com;Path=/;Expires=Tue, 26-Apr-2022 10:30:53 GMT;Secure;HttpOnly;Priority=HIGH;SameSite=none,3=3.;Path=/;Expires=Tue, 26-Apr-2022 10:30:53 GMT;Secure;HttpOnly;Priority=HIGH'}}
```
and
```
{headers: {'set-cookie': '1=1.;Domain=.site.com;Path=/;Expires=Tue, 26-Apr-2022 10:30:53 GMT;Priority=HIGH, 2=2.;Domain=.site.com;Path=/;Expires=Tue, 26-Apr-2022 10:30:53 GMT;Secure;HttpOnly;Priority=HIGH;SameSite=none, 3=3.;Path=/;Expires=Tue, 26-Apr-2022 10:30:53 GMT;Secure;HttpOnly;Priority=HIGH'}}
```
but neither set the cookies

If I mock `request.respond` and put the set-cookie each on their own header it works.

Ganitzsh:
I have a similar issue, did you find a solution to this?

ghost:
> I have a similar issue, did you find a solution to this?

I think I mocked request.respond with my own implementation. it's been a while, don't have the code anymore.

FranckFreiburger:
Maybe this little source code modification can do the trick:
https://github.com/puppeteer/puppeteer/blob/62ac16733913b388e2a4904ff6ef21aa431475e9/src/common/HTTPRequest.ts#L380

[...]

PR #8183, opened 2022-04-02
feat: respond multiple headers with same key

Junyan:

Fix issue https://github.com/puppeteer/puppeteer/issues/5740

Update `headers` option of `HTTP.respond()`, support multiple headers with same key.

- Yes

- Yes

**Summary**

In HTTP response header, `set-cookie` will occur repeatedly when there are multiple cookie names.

image

- Yes, array of header value will not convert to a string.


OrKoN (approved):
LGTM, thanks!

PR #8200, opened 2022-04-06
fix: consider existing frames when waiting for a frame

OrKoN:


src/common/Page.ts
```diff
@@ -2039,7 +2039,7 @@ export class Page extends EventEmitter {
       return false;
     }
 
-    return Promise.race([
```
ergunsh, line 2042: AFAIU, if there already are frames satisfying the predicate; their async execution must be faster (since we don't really wait for an event but execute a function on the page)

So, I wonder if we need to include that part in `Promise.race`. IMO, we can check it first, then if existing frames don't satisfy the predicate, we move on to waiting on events. WDYT?

It can be something like this:
```
const checkInExistingFrames = async () => {
      for (const frame of this.frames()) {
        if (await predicate(frame)) {
          return frame;
        }
      }
    }

    const foundFrame = await checkInExistingFrames();
    if (foundFrame) {
      return foundFrame;
    }

    return Promise.race([
      helper.waitForEvent(
        this._frameManager,
        FrameManagerEmittedEvents.FrameAttached,
        predicate,
        timeout,
        this._sessionClosePromise()
      ),
      helper.waitForEvent(
        this._frameManager,
        FrameManagerEmittedEvents.FrameNavigated,
        predicate,
        timeout,
        this._sessionClosePromise()
      ),
    ]);
```

This way it feels clearer what we do in the function, does it make sense?

ergunsh (commented):
I'm not sure where the test for this specific functionality can belong. However, I think we're able to test this change via using `utils.attachFrame` which attaches a frame and resolves after it is attached in the page.

For example, I added a test in `oopif.spec.ts` and it correctly tests this change:
```
it('should waitForFrame resolve immediately if the waited frame already exists in the page', async () => {
    const { server } = getTestState();

    await utils.attachFrame(
      page,
      'frame2',
      server.CROSS_PROCESS_PREFIX + '/empty.html'
    );

    await page.waitForFrame((frame) =>
      frame.url().endsWith('/empty.html')
    );
  });
```

Though I'm not sure if the test should belong there :sweat_smile:

src/common/Page.ts
OrKoN: I have been thinking about that but since the predicate is async, I thought it would racey as we might get events while we are checking predicates for the existing frames. WDYT about that?

OrKoN:
Thanks for taking a look! The idea of using `utils.attachFrame` for testing this looks good to me. Updating the PR.

src/common/Page.ts
ergunsh: Hihiiiim, that's true. Then, actually checking for existing frames and then waiting for events could be better for providing a deterministic behavior for that specific case:
* If there is already a frame on the page that satisfies the predicate, we'll first return that frame
* If not, we'll wait for events

Otherwise, it can be indeterministic: i.e. Sometimes while there already is a frame on the page, before its predicate is executed, an event can be received and we return the event's frame. WDYT?

OrKoN: > Otherwise, it can be indeterministic: i.e. Sometimes while there already is a frame on the page, before its predicate is executed, an event can be received and we return the event's frame. WDYT?

That is expected to be the same frame though (if you predicate is correct). Also, we don't guarantee any order in which you get frames so if the predicate is not correct, you get any first frame that is matching. We cannot really check predicates before setting up events listeners because that is not deterministic either.

ergunsh: Oh, I now see which case that can be problematic, let's go with the race solution; thanks for the explanation! :)

ergunsh (approved):

PR #8208, opened 2022-04-07
fix: ignore favicon requests in page.spec event handler tests

### LINKED ISSUE #6020: Issue with event handlers test in FF
In the PR that landed Mitt I added a spec to ensure that event handlers get cleaned up and re-added correctly. This passes locally for me in Chromium and FF, but on CI it always fails on FF. I have skipped the test which you can find here: https://github.com/puppeteer/puppeteer/blob/master/test/page.spec.js#L134

I'm not sure why it fails or what the root cause is unfortunately, but @whimboo or @mjzffr would you mind taking a look when you get a chance? Happy to try to help out if I can.

Thanks!

juliandescottes:
Fixes #6020 

Favicon requests are automatically triggered when loading empty.html (except on chrome headless) and make the tests fail intermittently.

In this patch, I define a custom response handler which ignores specifically the favicon requests. I prefer this rather than "only consider requests to empty.html" because it feels more conservative, and should not hide potential future regressions.

Alternatively, we could use a different test page than empty.html, eg called `no-favicon.html` which would simply contain
```html
<link rel="icon" href="data:,">
```

Let me know which approach you prefer.

### DIFF
--- a/test/page.spec.ts
+++ b/test/page.spec.ts
@@ -127,14 +127,20 @@ describe('Page', function () {
       const { page, server } = getTestState();
 
       const handler = sinon.spy();
-      page.on('response', handler);
+      const onResponse = (response) => {
+        // Ignore default favicon requests.
+        if (!response.url().endsWith('favicon.ico')) {
+          handler();
+        }
+      };
+      page.on('response', onResponse);
       await page.goto(server.EMPTY_PAGE);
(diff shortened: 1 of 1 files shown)


juliandescottes:
cc @OrKoN for review :)

OrKoN (approved):
LGTM thanks!

PR #8210, opened 2022-04-08
Update PDFOptions.ts

### LINKED ISSUE #8209: [Bug]: Types for page format don't match the documentation
### Bug description

1. Read the documentation at https://github.com/puppeteer/puppeteer/blob/v13.5.2/docs/api.md#pagepdfoptions especially under the subheading 'The format options are:'
2. Notice all the formats begin with an uppercase letter.
3. Specify format 'A4'.
4. Convert script to TypeScript.
5. Get an error that this is an invalid format; they must all be lowercase.

In node_modules/puppeteer/lib/types.d.ts under PDFOptions line 6051 I can see this definition: 
```ts
export declare type PaperFormat = 'letter' | 'legal' | 'tabloid' | 'ledger' | 'a0' | 'a1' | 'a2' | 'a3' | 'a4' | 'a5' | 'a6';
```
and I would expand it to include at least those starting with an uppercase letter to match the documentation.  I might also add an all-caps version of the first four for good measure, since it's case-insensitive anyway as of #211.  I would also restore the test removed in that PR alongside the one that it was converted to.  However, I'm not seeing where that test wound up, so the attached PR doesn't modify it, but pointers / maintainer edits / PRs on my branch are welcome.

### Puppeteer version

13.5.2

### Node.js version

16.13.1

### npm version

7.24.0

Windows

### Relevant log output

```shell
Type '"A4"' is not assignable to type 'PaperFormat | undefined'.
types.d.ts(6145, 9): The expected type comes from property 'format' which is declared here on type 'PDFOptions'
```

wbt:
Add case options consistent with documentation

Closes #8209 but without the test change suggested there.

Type change bugfix that brings code in line with documentation.

No, see comment on #8209 

This is updating the code to match what the documentation already says.

**Summary**
See #8209 for the issue.

No, at least not intentionally.

### DIFF
--- a/src/common/PDFOptions.ts
+++ b/src/common/PDFOptions.ts
@@ -55,6 +55,17 @@ export interface PDFMargin {
  * @public
  */
 export type PaperFormat =
+  'Letter'
+  | 'Legal'
+  | 'Tabloid'
+  | 'Ledger'
+  | 'A0'
+  | 'A1'
+  | 'A2'
+  | 'A3'
+  | 'A4'
+  | 'A5'
+  | 'A6'
   | 'letter'
   | 'legal'
   | 'tabloid'
@@ -65,7 +76,11 @@ export type PaperFormat =
   | 'a3'
   | 'a4'
   | 'a5'
-  | 'a6';
+  | 'a6'
+  | 'LETTER'
+  | 'LEGAL'
+  | 'TABLOID'
+  | 'LEDGER';
 
 /**
  * Valid options to configure PDF generation via {@link Page.pdf}.


wbt:
Someone else can take this and re-do the work; barriers to contribution in this repo exceed the value of the contribution.

OrKoN:
Closed in https://github.com/puppeteer/puppeteer/pull/8211

wbt:
Thanks! One small tweak you might consider is to have the `Capitalize` version first so that type hints shown on hover in certain IDEs match the documentation shown alongside it.

PR #8212, opened 2022-04-08
feat(chromium): roll to Chromium 102.0.4985.0 (r988650)

ergunsh:


versions.js
```diff
@@ -17,6 +17,7 @@
 const versionsPerRelease = new Map([
   // This is a mapping from Chromium version => Puppeteer version.
   // In Chromium roll patches, use 'NEXT' for the Puppeteer version.
+  ['102.0.4985.0', 'NEXT'],
```
OrKoN, line 20: We need an M101 release :)

OrKoN: @ergunsh I think it's time to see if we can rebase that and maybe find a newer build. The release is scheduled for May 26 so if anything breaks we have time to address that.

ergunsh: Closing this in favor of #8319

ergunsh:
See #8319

PR #8226, opened 2022-04-12
fix: forget request will be sent from the _requestWillBeSentMap list.

kblok:
That `forget` looked wrong to me. Let me know if this change is correct.


ergunsh (approved):
LGTM, thanks for the fix!

PR #8233, opened 2022-04-13
fix: also kill Firefox when temporary profile is used

whimboo:
This particular check had been added because of issues with killing the Firefox process on Windows. But that was fixed in Firefox 96 and we should be able to remove the workaround.

Lets see if CI is happy with it.

### DIFF
--- a/src/node/BrowserRunner.ts
+++ b/src/node/BrowserRunner.ts
@@ -164,7 +164,7 @@ export class BrowserRunner {
 
   close(): Promise<void> {
     if (this._closed) return Promise.resolve();
-    if (this._isTempUserDataDir && this._product !== 'firefox') {
+    if (this._isTempUserDataDir) {
       this.kill();
     } else if (this.connection) {
       // Attempt to close the browser gracefully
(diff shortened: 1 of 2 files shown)


whimboo:
There are quite some failures visible on Windows so we most likely have to delay landing of this PR.

whimboo:
@OrKoN as we discussed we would still like to get this workaround removed given that the underlying issue in Firefox is really fixed. Failures in the launcher spec tests are most likely caused by some combination of running these in sequence. As such I would propose (for now) to just separate out the `should filter out ignored default arguments` for Firefox given that we do not have that many arguments as Chrome and we would remove arguments optionally added. Please let me know if that would work for you. Thanks.

OrKoN (approved):

OrKoN:
LGTM thanks!

whimboo:
@OrKoN I just noticed that we have issues locally with our expected results JSON file which is keyed by test name. With having the same test name twice now (once for Chrome and one more for Firefox) we cannot explicitly mark the expected data due to duplicated keys. I would suggest that we slightly rename the Firefox test so that the name is different if that's ok.

PR #8235, opened 2022-04-13
docs: improve custom query handler docs

s-h-a-d-o-w:

- Describe the limitations for custom query handler code
- Keep all of the detailed explanation in only one part of the docs instead of two

The second point is subjective but I hope you will agree that at least the first one is good. 🙂 

**Summary**

It took me like an hour or two today to realize what's going on, when I finally looked at how puppeteer processes query handlers. 😅
I would like to make it easier for other people in the future.


OrKoN:
@johanbay could you PTAL?

johanbay (approved):
Looks good to me, thanks for clarifying the docs wrt serialization!

PR #8250, opened 2022-04-19
fix: use different test names for browser specific default arguments tests in launcher.spec.ts

whimboo:
The changes as pushed on RP #8233 caused duplicated test names which makes it hard for our downstream sync of Puppeteer to maintain an expected results list in JSON by having test names as key. 

As such I would suggest that we rename both tests and explicitly add the browser name in the test name.

### DIFF
--- a/test/launcher.spec.ts
+++ b/test/launcher.spec.ts
@@ -404,44 +404,49 @@ describe('Launcher specs', function () {
         await page.close();
         await browser.close();
       });
-      itChromeOnly('should filter out ignored default arguments', async () => {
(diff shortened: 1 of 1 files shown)


whimboo:
@OrKoN can you please review? Thanks

OrKoN:
LGTM. We can land it after the current release is done!

whimboo:
As it looks like the [13.6.0 release](https://github.com/puppeteer/puppeteer/releases/tag/v13.6.0) is already out. :)

OrKoN (approved):

whimboo:
Thanks!

PR #8252, opened 2022-04-20
fix: increase commit header length limit

OrKoN:
https://github.com/puppeteer/puppeteer/commit/04f7659f9adae7415bfea3f538a446c624d852e8 hit the limit but I don't see a reason why we need to be so strict. Increasing the limit to 120 chars.

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -19,5 +19,6 @@ module.exports = {
     'body-max-line-length': [0, 'always', 100],
     'footer-max-line-length': [0, 'always', 100],
     'subject-case': [0, 'never'],
+    'header-max-length': [0, 'always', 120],
   },
 };


ergunsh:
AFAIU GitHub cuts of the commit to other line when it is longer than 72 characters. So, AFAIU again, that's why the default limit is 72 characters. IMO, it might make sense to keep it in 72 characters to not allow this behavior.

Additional info can be written in other lines, wdyt?

OrKoN:
The limit already was 100. Also, GitHub cutting off lines is not a big problem as you can see the rest on the next line.

OrKoN:
Alternatively, I can fix that commit with a force push :)

OrKoN:
Let me just do that, I think.

ergunsh:
Oh I see, I tried to understand why it was 100 in the first place and it seems like 72 could cause issues with automatic PRs (example https://github.com/conventional-changelog/commitlint/issues/859)

I'm fine with both options now, thank you :)

OrKoN:
Oh, thanks for digging! yeah, I think 72 is too low for most of typical commit messages.

PR #8273, opened 2022-04-25
fix: tests failing in headful

OrKoN:
A collection of fixes for the tests failing in headful. There are three failing tests remaining after this PR.

### DIFF
--- a/src/node/Launcher.ts
+++ b/src/node/Launcher.ts
@@ -203,7 +203,7 @@ class ChromeLauncher implements ProductLauncher {
 
   defaultArgs(options: BrowserLaunchArgumentOptions = {}): string[] {
     const chromeArguments = [
-      '--allow-pre-commit-input', // Allows input events to be dispatched while the compositor commits are in progress. Reduces flakiness in headful.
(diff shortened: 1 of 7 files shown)


src/node/Launcher.ts
```diff
@@ -203,6 +203,7 @@ class ChromeLauncher implements ProductLauncher {
 
   defaultArgs(options: BrowserLaunchArgumentOptions = {}): string[] {
     const chromeArguments = [
+      '--allow-pre-commit-input', // TODO(crbug.com/1320996): neither headful nor headless should rely on this flag.
```
ergunsh, line 206: Can we add this argument only for the related tests? (or is it needed in all tests?)

ergunsh (approved):
LGTM

src/node/Launcher.ts
OrKoN: It's also needed for runtime (not only tests) in headful unfortunately

ergunsh: Oh understood, thank you!

PR #8281, opened 2022-04-27
feat: add `back` and `forward` mouse buttons

jrandolf-google:


OrKoN:
I believe it's a feature, not a chore? Let's make sure that the change is documented and we have test coverage for those button types.

jrandolf-google:
> I believe it's a feature, not a chore? Let's make sure that the change is documented and we have test coverage for those button types.

I thought of this more as upstreaming the CDP value to puppeteer. In particular, we also do not test `middle` clicks in our tests. I can write the tests in any case.

Issue #8283, opened 2022-04-27
Firefox test related to launching the browser is failing on Mac


OrKoN:
See https://github.com/puppeteer/puppeteer/runs/6189321000?check_suite_focus=true

OrKoN:
Henrik, do you know what might have broken there?

whimboo:
Note that the test has been disabled via PR #8286. 

@OrKoN do you have some more details? How often is this test failing? Is it permanent or intermittent? And it seems to be MacOS only, right?

The failure is actually caused by a timeout:
https://github.com/puppeteer/puppeteer/runs/6189321000?check_suite_focus=true


  1) Launcher specs
       Puppeteer
         Puppeteer.launch
           should have custom URL when launching browser:
     Error: Timeout of 25000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/puppeteer/puppeteer/test/launcher.spec.ts)
      at listOnTimeout (internal/timers.js:554:17)
      at processTimers (internal/timers.js:497:7)

Maybe it's related to my fix on bug https://bugzilla.mozilla.org/show_bug.cgi?id=1759169 and that we face a shutdown hang of Firefox here.

OrKoN:
I saw it fail consistently on multiple CLs and also locally on my Mac. Locally, I tried to debug it and the browser was started successfully but the call to get pages was timing out. It looks like page initialisation was locking on some events.

whimboo:
@OrKoN could you do me a favor please? Could you run again locally with `DUMPIO=1` and by having the Firefox preference `remote.log.level` set to `Trace` (to be added [here](https://github.com/puppeteer/puppeteer/blob/main/src/node/Launcher.ts#L643))? I would be kinda interested to see the complete log output for this test. Thanks!

OrKoN:
@whimboo sorry, I didn't get to this yet. I will try to take a look tomorrow.

OrKoN:
@whimboo PTAL at the log https://gist.github.com/OrKoN/91d653b48821989be895aa31fb4c12b7

whimboo:
@OrKoN I completely missed your reply, sorry. I had a look but it's not obvious what's Puppeteer is actually waiting for. The last initial command after the connection that gets successfully run is `Network.enable`. What else runs next eg. are there some events that Puppeteer is waiting for?

I wonder if this is still an issue for you or if it has been gone given that 2 months have been passed by.

[...]

PR #8286, opened 2022-04-27
test: disable a failing Firefox test

OrKoN:


OrKoN:
cc @whimboo

ergunsh (approved):

whimboo:
@OrKoN do you have some more details? How often is this test failing? Is it permanent or intermittent? And it seems to be MacOS only?

The failure is actually caused by a timeout:
https://github.com/puppeteer/puppeteer/runs/6189321000?check_suite_focus=true

```

  1) Launcher specs
       Puppeteer
         Puppeteer.launch
           should have custom URL when launching browser:
     Error: Timeout of 25000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/runner/work/puppeteer/puppeteer/test/launcher.spec.ts)
      at listOnTimeout (internal/timers.js:554:17)
      at processTimers (internal/timers.js:497:7)
```

Maybe it's related to my fix on bug https://bugzilla.mozilla.org/show_bug.cgi?id=1759169 and that we face a shutdown hang of Firefox.

whimboo:
Actually lets move further discussion to issue #8283.

PR #8297, opened 2022-05-02
fix!: strict mode fixes for HTTPRequest/Response classes

OrKoN:
Solves type issues in HTTPRequest/Response with some slight changes in the behaviour and API.

Issues #6769


src/common/HTTPRequest.ts
```diff
@@ -217,7 +218,7 @@ export class HTTPRequest {
   /**
    * @returns the most recent reason for aborting the request
    */
-  abortErrorReason(): Protocol.Network.ErrorReason {
+  abortErrorReason(): Protocol.Network.ErrorReason | undefined {
```
ergunsh: It feels like this is a breaking change for the API we provide to the user. (i.e. previously we were saying that this method will always return an `ErrorReason` now we're saying that it might not return one; so their TS builds might break)

So, I guess we might land this as a breaking change with a major semver update, wdyt?

src/common/HTTPRequest.ts
```diff
@@ -150,9 +150,11 @@ export class HTTPRequest {
   private _headers: Record<string, string> = {};
   private _frame: Frame;
   private _continueRequestOverrides: ContinueRequestOverrides;
-  private _responseForRequest: Partial<ResponseForRequest>;
-  private _abortErrorReason: Protocol.Network.ErrorReason;
-  private _interceptResolutionState: InterceptResolutionState;
+  private _responseForRequest: Partial<ResponseForRequest> = {};
```
ergunsh: I feel like this field is similar to `_abortErrorReason` from the usage. So, to have the same typing for the related two methods; would it make sense to return `undefined` for the case `responseForRequest` is called before `respond` call (like `abortForReason`); wdyt?

Also setting the default value to `{}` can mean a behavior change:
Calling `responseForRequest` before `respond` was returning `undefined` now it returns empty object.

So I think it makes sense to type this as `Partial<ResponseForRequest> | undefined` and not set a default value.

src/common/HTTPRequest.ts
```diff
@@ -174,13 +176,12 @@ export class HTTPRequest {
     this._interceptionId = interceptionId;
     this._allowInterception = allowInterception;
     this._url = event.request.url;
-    this._resourceType = event.type.toLowerCase() as ResourceType;
+    this._resourceType = event.type?.toLowerCase() as ResourceType;
```
ergunsh: Are we sure that `event.type?.toLowerCase()` will be `ResourceType`?

If so, can we add a comment to why we're sure?
If not, I think we need to update typing to reflect `_resourceType` being able to `undefined`. WDYT?

src/common/HTTPRequest.ts
```diff
@@ -572,11 +573,13 @@ export class HTTPRequest {
...
 
+    const status = response.status || 200;
     await this._client
       .send('Fetch.fulfillRequest', {
         requestId: this._interceptionId,
-        responseCode: response.status || 200,
-        responsePhrase: STATUS_TEXTS[response.status || 200],
+        responseCode: status,
+        responsePhrase:
+          STATUS_TEXTS[String(status) as keyof typeof STATUS_TEXTS],
```
ergunsh: How are we sure that this assertion is valid? Can you add a comment to describe it?

Also, I wonder if we can type `ResponseForRequest` more strictly for the values that `status` can get?
If we're sure that it will be one of the valid responses from the `STATUS_TEXTS`; maybe we can
* Change the key of `STATUS_TEXTS` to be number
* Update `ResponseForRequest`s `status` field to be `keyof typeof STATUS_TEXTS`
* Remove the `String(status)` conversion here

AFAIU that change would also mean a breaking change though since the users won't be able to provide a `ResponseForRequest` with a `status` different than the ones in `STATUS_TEXTS`. WDYT?

src/common/HTTPRequest.ts
```diff
@@ -626,13 +629,13 @@ export class HTTPRequest {
...
   private async _abort(
-    errorReason: Protocol.Network.ErrorReason
+    errorReason: Protocol.Network.ErrorReason | undefined
   ): Promise<void> {
     this._interceptionHandled = true;
     await this._client
       .send('Fetch.failRequest', {
         requestId: this._interceptionId,
-        errorReason,
+        errorReason: errorReason || 'Failed',
```
ergunsh, line 640: I don't know what this client function does though this feels like a behavior change:
* Before, when `errorReason` was given as `undefined` we were sending this as `undefined`. Now we're adding a default value `Failed` to it, is it okay?

src/common/HTTPResponse.ts
```diff
@@ -76,8 +76,8 @@ export class HTTPResponse {
     });
 
     this._remoteAddress = {
-      ip: responsePayload.remoteIPAddress,
-      port: responsePayload.remotePort,
+      ip: responsePayload.remoteIPAddress || '',
```
ergunsh: I think these might also be behavior changing since previously we were putting `undefined` for `ip` when it doesn't exist in `remoteIPAddress`; now we're putting empty string.

IMO we can do two things:
* If we're sure that `remoteIPAddress` exists, we can do an assertion here
* If we know that `remoteIPAddress` might be `undefined` and we're fine with it; we can update `_remoteAddress` typing to reflect this. WDYT?
* We can accept this behavior change and keep these default values.
WDYT?

src/common/HTTPRequest.ts
OrKoN: I am not sure it warrants a major semver update because it's a bug in types. The underlying value always had a chance to be undefined.

src/common/HTTPRequest.ts
OrKoN: I think so as `Fetch.failRequest` requires a reason and sending undefined would result in failure.

src/common/HTTPRequest.ts
ergunsh: This is also true and thinking that this is a pretty small change, it might not warrant a major semver update. However, there is a possibility that this change will break people's builds. Soooo I don't know how to move forward, wdyt?

OrKoN: I'd be for just making a change.  But we can make the runtime error instead if the underyling value is undefined. I'd prefer a build time error though as it is easier for the users to address.

ergunsh: If we think this is not a frequent use case, I'd prefer runtime error since it will affect less users (none if this case is really not possible)

Otherwise, regardless of the people using this API correctly or not, they have a chance to get a build error which possibly will affect more people. WDYT?

src/common/HTTPRequest.ts
OrKoN: Good point. Actually, I double checked the API doc and it says we promise to return a null here instead of undefined. So this one will be a breaking fix too.

src/common/HTTPRequest.ts
OrKoN: I think in practice it's rarely empty if at all but on the CDP level is can be undefined.

OrKoN: We cannot change it undefined because our API docs say it's always defined. I will add a fallback to the type `other`.

src/common/HTTPRequest.ts
OrKoN: We don't try to minimize the number of users affected by this. The ultimate goal is the correctness so that the users can rely on our API. I think here we actually need to change the API as the abort reason is not guaranteed to be defined previously.

OrKoN: I will mark the change accordingly.

src/common/HTTPRequest.ts
OrKoN: Good point, I think we can just change the index signature of STATUS_TEXTS here!

src/common/HTTPRequest.ts
OrKoN: I think it's alright to do that if there is a reason missing for some reason :) the abort() API also defaults to the failed status if no reason is provided. I suspect this does not make much difference in practice but I'd like to avoid introducing runtime errors for this.

src/common/HTTPResponse.ts
OrKoN: Let's change the API here, good point!

src/common/HTTPRequest.ts
ergunsh: Hihiiim I feel like typing `STATUS_TEXTS` as `{ [key: string]: string }` can be erroneous since what I understand from this signature is that, whatever `key` I provide as `string`, I'll get a string back.

However, `STATUS_TEXTS` doesn't ensure this since we're able to get `undefined` value with something like `104`.

So, if we're sure that `response.status` will be one of the keys of `STATUS_TEXTS`, it would be better to type it more strictly (like `keyof typeof STATUS_TEXTS`) or if we don't want to change the typing of `response`, we can keep the assertion and add a comment on why we're sure that `response.status` will be one of the keys of `STATUS_TEXTS`. WDYT?

src/common/HTTPRequest.ts
ergunsh: Hihiiim makes sense to change the API in a better way though I'm still having second thoughts of breaking people's builds for this reason 😅

Say I, as a user, had this code:
```
function printErrorReason(errorReason: ErrorReason) {
    console.log(errorReason);
}
const errorReason = httpRequest.abortErrorReason();
printErrorReason(errorReason);
```

Now, when I upgrade Puppeteer a minor version that contains the changes in this PR, it will break my build for this code since `printErrorReason` expects a call with an `ErrorReason` and not `ErrorReason | null`. I don't have strong opinions on whether this breakage is acceptable or not, so if you think it is acceptable, we can move forward with the API change.

ergunsh (approved):
Code changes LGTM with 1 comment on `STATUS_TEXTS` and `response` typing.

Apart from that, I had a comment on the nature of introducing breaking changes in a minor version and I don't have a strong opinion on that one. If you think that is acceptable, this is good to go! 🥳

PR #8299, opened 2022-05-02
chore: drop Node 12 support

jrandolf-google:
This patches removes support for Node 12.

BREAKING CHANGE: Node 12 is no longer supported.


OrKoN:
Should we update package.json? https://github.com/puppeteer/puppeteer/blob/main/package.json#L14 Also, let's mark this PR as a breaking change?

OrKoN (approved):

PR #8306, opened 2022-05-03
feat: support ES modules

jrandolf-google:


OrKoN:
Let's also involve @ergunsh in the review as we all need to support this going forward :)

src/constants.ts
```diff
@@ -0,0 +1,8 @@
+import { readFileSync } from 'fs';
+import { dirname, join } from 'path';
+import { puppeteerDirname } from './compat.js';
+
+export const rootDirname = dirname(dirname(dirname(puppeteerDirname)));
+export const packageVersion = JSON.parse(
```
OrKoN: how does it work in the bundled versions when there is no package.json?

scripts/test-install.sh
```diff
@@ -1,41 +1,83 @@
...
-# 1. we publish the right files in the `files` list from package.json
-# 2. The install script works and correctly exits without errors
-# 3. Requiring Puppeteer from Node works.
 npm install --loglevel silent "${tarball}"
 node --eval="require('puppeteer')"
+node --eval="require('puppeteer/lib/cjs/puppeteer/revisions.js');"
 ls $TMPDIR/node_modules/puppeteer/.local-chromium/
 
-# Again for Firefox
+echo "Testing... Chrome ES Modules"
```
OrKoN, line 25: should we add a basic webpack/esbuild test?

src/constants.ts
jrandolf: Good catch, I guess hoisting this constant up would cause errors in such an environment.

OrKoN: I think it was not an issue when we required json as bundlers could probably trace package.json. But I don't think they will handle fs.read. Is there an alternative way to get the version for both esm/cjs?

scripts/test-install.sh
jrandolf: Done.

src/constants.ts
jrandolf: As discussed offline, we are generating the file.

OrKoN:
Pls remove puppeteer-13.7.0-post.tgz (it seems it got committed somehow)

src/generated/version.ts.tmpl
```diff
@@ -0,0 +1 @@
+export const packageVersion = 'PACKAGE_VERSION';
```
OrKoN: should we not put the template into the generated folder?

package.json
```diff
@@ -37,15 +48,19 @@
     "doc": "node utils/doclint/cli.js",
     "generate-api-docs-for-testing": "commonmark docs/api.md > docs/api.html",
     "clean-lib": "rimraf lib",
-    "build": "npm run tsc && npm run generate-d-ts",
-    "tsc": "npm run clean-lib && tsc --version && npm run tsc-cjs && npm run tsc-esm",
+    "build": "npm run tsc && npm run generate-d-ts && npm run generate-pkg-json",
```
OrKoN: Let's include everything generated in https://github.com/puppeteer/puppeteer/blob/main/.versionrc.js#L23 I don't think we should generate during build, rather we should generate when the version is changed.

src/generated/version.ts
```diff
@@ -0,0 +1 @@
+export const packageVersion = '13.7.0-post';
```
OrKoN: that would be an invalid version. I think we should make sure that we only generate "release" versions

compat/esm/compat.ts
```diff
@@ -0,0 +1,20 @@
...
+let puppeteerDirname: string;
+
+try {
+  // In some environments, like esbuild, this will throw an error.
+  // We suppress the error since the bundled binary is not expected
+  // to be used or installed in this case and, therefore, the
+  // root directory does not have to be known.
+  puppeteerDirname = dirname(require.resolve('./compat'));
+} catch (error) {
+  // Fallback to __dirname.
```
ergunsh: This comment is leftover from the paste I think :)

package.json
```diff
@@ -80,22 +96,22 @@
     "ws": "8.5.0"
   },
   "devDependencies": {
-    "@commitlint/cli": "16.2.4",
-    "@commitlint/config-conventional": "16.2.4",
+    "@commitlint/cli": "16.2.3",
```
ergunsh: I wonder why we're down versioning the dependencies here, is it intended? 🤔

package.json
```diff
@@ -37,15 +48,19 @@
...
     "tsc-cjs": "tsc -b src/tsconfig.cjs.json",
     "tsc-esm": "tsc -b src/tsconfig.esm.json",
+    "tsc-compat-cjs": "tsc -b compat/cjs/tsconfig.json",
+    "tsc-compat-esm": "tsc -b compat/esm/tsconfig.json",
     "apply-next-version": "node utils/apply_next_version.js",
     "test-install": "scripts/test-install.sh",
     "clean-docs": "rimraf website/docs && rimraf docs-api-json",
     "generate-d-ts": "npm run clean-docs && api-extractor run --local --verbose",
     "generate-docs": "npm run generate-d-ts && api-documenter markdown -i docs-api-json -o website/docs && node utils/remove-tag.js",
+    "generate-pkg-json": "echo '{\"type\": \"module\"}' > lib/esm/package.json",
```
ergunsh: (nit): I think it might be good to update the script name to include `esm` to show that this script generates the package.json for `esm` export? Smt like `generate-esm-pkg-json`

src/generated/version.ts.tmpl
ergunsh: I wonder if we can move reading `version` from `package.json` to the `compat` part and to the `constants` as well so that we won't need this separate build step, wdyt?

scripts/test-install.sh
```diff
@@ -1,41 +1,109 @@
...
+echo '{"type":"module"}' >>$TMPDIR/package.json
+npm install --loglevel silent "${tarball}"
+node --input-type="module" --eval="import puppeteer from 'puppeteer'"
+node --input-type="module" --eval="import 'puppeteer/lib/esm/puppeteer/revisions.js';"
+node --input-type="module" --eval="
+import puppeteer from 'puppeteer';
+(async () => {
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+  await page.goto('http://example.com');
```
ergunsh, line 37: For testing whether `puppeteer` works in cjs and esm, do we need to run a real test or would having import and maybe a few simple steps without network request enough? (like only `const await browser.newPage(); await browser.close();`)

ergunsh (commented):
LGTM with a few comments & questions. Leaving approval to @OrKoN :)

src/generated/version.ts.tmpl
jrandolf: This is not possible. As discussed offline, JSON cannot be imported properly in ESM without Node 17, so we need to hardcore the version with a separate build step.

ergunsh: Hihiiim it is interesting, wdyt of reading it as a file and parsing it as json?

scripts/test-install.sh
jrandolf: Not too sure, but it doesn't hurt to run a real test as this provides more coverage inherently.

src/generated/version.ts.tmpl
jrandolf: Then bundled applications will not work since bundlers don't transpile JSON from fs.read.

scripts/test-install.sh
ergunsh: hihiim, I think we don't have a clear win there with running a real test for testing the install and whether the exports we provide works. So having the additional overhead of making a network request in this test and depending on an external website is not justified.

Thus, I'd prefer removing those parts that goes to a website and takes a screenshot. WDYT?

src/generated/version.ts.tmpl
ergunsh: Oh understood, makes sense. Thanks for the explanation! :)

ergunsh: Even though this is resolved, I'm trying to understand why bundled applications don't work. Can you describe it a bit more?

I thought a bit and currently am thinking that:
* The environment our library runs is NodeJS (is that right?)
* in that environment, we're able to read from file

So, regardless of the bundler operation; we should be able to read `package.json` in the folder our entrypoint lives in. Thus, I'm trying to understand why bundled apps don't work.

scripts/test-install.sh
```diff
@@ -1,41 +1,109 @@
 #!/usr/bin/env sh
```
OrKoN, line 1: oh no, it was me who commited the gz file :(

OrKoN: I have created https://github.com/puppeteer/puppeteer/pull/8320 so that this PR can be only about ESM

scripts/test-install.sh
```diff
@@ -1,41 +1,109 @@
...
+(async () => {
+  const browser = await puppeteer.launch();
+  const page = await browser.newPage();
+  await page.goto('http://example.com');
+  await page.screenshot({ path: 'example.png' });
+  await browser.close();
+})();
+" >>$TMPDIR/index.js
+npx webpack
+node dist/bundle.cjs
```
OrKoN, line 70: let's do `rm -rf node_modules` to make sure that Puppeteer is bundled

scripts/test-install.sh
OrKoN: I think it's okay to use example.com for this smoke test as screenshooting touches deeper dependencies. So I think it's fine to keep the test as is. If it becomes flaky, we can reduce the test area.

src/generated/version.ts.tmpl
ergunsh: We've talked offline (online?),  @jrandolf explained to me that the reason it doesn't work well is, let's say we have `fs.read("./package.json");` in `compat.ts` and say an app that uses Webpack bundles their app with Puppeteer dependency into `dist/dist.js`

Since the path to `./package.json` now points to `dist/package.json`, the `fs.read` won't work.

OrKoN (approved):

PR #8307, opened 2022-05-03
chore: remove obsolete puppeteer-firefox package

jrandolf-google:


ergunsh (commented):
LGTM though @OrKoN to approve :)

I also wonder, there is a package [`puppeteer-firefox`](https://www.npmjs.com/package/puppeteer-firefox) and we've said that the package will remain available until the transition happens. So, is the transition already happened?

If so, would it make sense to update `puppeteer-firefox` package somehow to move the users from there to `puppeteer` itself; wdyt? AFAIU there are still quite some users of that package (13k weekly downloads)

OrKoN:
Henrik, do you know if we are able to remove the source for the puppeteer-firefox package now?

whimboo:
> Henrik, do you know if we are able to remove the source for the puppeteer-firefox package now?

Nothing basically uses it for a while, right? Note that we still have Firefox Nightly support only. Maybe we can discuss that in the next upcoming Puppeteer/Firefox meeting? I'll add it to the agenda.

OrKoN:
I think it's not used anywhere. It will remain available at https://www.npmjs.com/package/puppeteer-firefox and there is already a deprecation notice that it is not maintained. I guess we can discuss it during the meeting but looks like the source code could be removed. In the worst case, we would need to recover the source code if we need to make an update to the package. Also, the overall status puppeteer-firefox seems to be an experimental prototype. So if you think we should discuss it in the meeting, that sounds good to me. Otherwise, let's just remove the code in this CL if you are not relying on it anywhere (we don't).

whimboo:
@OrKoN the code in `puppeteer-firefox` is not ours. So it's nothing what we would need. I'm happy to see it removed if you think that we don't need to have a discussion. Maybe it could be kept in a specific branch for now and then removed from main?

OrKoN:
SGTM. I have created a branch called `puppeteer-firefox` to have to available there for easy access (although it's fairly easy to recover without the branch). I think we can land this now.

OrKoN (approved):

PR #8315, opened 2022-05-04
feat: add Dockerfile

OrKoN:
This PR adds an official Dockerfile for Puppeteer. The content of the Dockerfile is practically the same as documented in troubleshooting.md:

1) It installs chrome-stable and dependencies via apt-get.
2) it installs a local Puppeteer build into the docker user's home folder.
3) configures required permissions for the user.
4) outputs licenses into the THIRD_PARTY_NOTICES file.

The local Puppeteer build is created by `docker/pack.sh` which is meant to be used in CI. This PR also includes a GitHub action that would build a docker image and run a smote test inside of it.

The next step would be actually publishing the docker image from GitHub Actions to GitHub Registry.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -181,3 +181,35 @@ jobs:
         run: |
           # Note: this modifies package.json to test puppeteer-core, so we test this last.
           npm run test:install
+
+  docker-tests:
+    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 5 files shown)


Ruslan34395 (commented):
[removed reference to issue 8356]

mathiasbynens (approved):

PR #8329, opened 2022-05-10
feat: add waitForXPath to ElementHandle

OrKoN:
This PR adds waitForXPath to ElementHandle similar to waitForSelector.

### DIFF
--- a/src/common/JSHandle.ts
+++ b/src/common/JSHandle.ts
@@ -416,6 +416,84 @@ export class ElementHandle<
     return result;
   }
 
+  /**
+   * Wait for the `xpath` within the element. If at the moment of calling the
+   * method the `xpath` already exists, the method will return immediately. If
+   * the `xpath` doesn't appear after the `timeout` milliseconds of waiting, the
+   * function will throw.
+   *
+   * If `xpath` starts with `//` instead of `.//`, the dot will be appended automatically.
+   *
+   * This method works across navigation
+   * ```js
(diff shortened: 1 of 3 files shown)


src/common/JSHandle.ts
```diff
@@ -416,6 +416,84 @@ export class ElementHandle<
     return result;
   }
 
+  /**
+   * Wait for the `xpath` within the element. If at the moment of calling the
+   * method the `xpath` already exists, the method will return immediately. If
+   * the `xpath` doesn't appear after the `timeout` milliseconds of waiting, the
+   * function will throw.
+   *
+   * If `xpath` starts with `//` instead of `.//`, the dot will be appended automatically.
```
jrandolf, line 425: Is this comment necessary? I think it will just confuse people as almost everyone will use `//` for the start. We could also rephrase this to state `//` and `.//` are supported, without providing implementation details.

jrandolf-google (approved):

src/common/JSHandle.ts
OrKoN: `//` means querying from the root node so we cannot allow `//` as it might return elements outside of the element's subtree. So I think this comment is useful to indicate that we rewrite `//` as `.//`.

OrKoN: Maybe I could phrase it better 🤔

jrandolf: It's just a nit. Feel free to move forward with this.

PR #8330, opened 2022-05-10
fix: disable AvoidUnnecessaryBeforeUnloadCheckSync to fix navigations

OrKoN:


ergunsh (approved):

src/node/Launcher.ts
```diff
@@ -214,7 +214,9 @@ class ChromeLauncher implements ProductLauncher {
       '--disable-default-apps',
       '--disable-dev-shm-usage',
       '--disable-extensions',
-      '--disable-features=Translate,BackForwardCache',
+      // TODO: remove AvoidUnnecessaryBeforeUnloadCheckSync below
+      // once crbug.com/1324138 is fixed and released.
```
paulirish, line 218: Coming across this randomly. As this shipped in May I imagine it can be removed now?

I'm not in any rush. Just trying to be helpful. :)

OrKoN: Removed in https://github.com/puppeteer/puppeteer/pull/9239 thanks for the ping!

PR #8336, opened 2022-05-12
feat: allow handling other targets as pages internally

OrKoN:


jrandolf-google:
Could we add tests for this?

OrKoN:
done

jrandolf-google (approved):

PR #8344, opened 2022-05-13
chore: set up daily jobs for latest Chromium builds

OrKoN:


.github/workflows/tot.yml
```diff
@@ -0,0 +1,75 @@
+name: run-tot-checks
+# Checks Puppeteer against the latest ToT build of Chromium.
+# Declare default permissions as read only.
+permissions: read-all
+
+on:
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    # Supposed to be every day at 8 am (presumably UTC?).
```
mathiasbynens: It _is_ UTC per https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule

utils/check_availability.js
```diff
@@ -63,6 +60,9 @@ options
     -r          roll mode checks for the most recent stable Chromium roll candidate
     -rb         roll mode checks for the most recent beta Chromium roll candidate
     -rd         roll mode checks for the most recent dev Chromium roll candidate
+    -p platform print the latest revision for the given platform (${SUPPORTED_PLATFORMS.join(
```
mathiasbynens: missing alignment

utils/check_availability.js
```diff
@@ -193,9 +227,20 @@ async function checkRangeAvailability({
   fromRevision,
   toRevision,
   stopWhenAllAvailable,
+  platforms,
+  printAsTable,
```
mathiasbynens: We could use default parameter syntax here, i.e. `printAsTable = true`

mathiasbynens (approved):

PR #8348, opened 2022-05-13
fix: only check loading iframe in lifecycling

jrandolf-google:
This PR ensures lazy-loaded frames do not stall navigation of a frame by flagging frames that have not started loading.

### DIFF
--- a/src/common/FrameManager.ts
+++ b/src/common/FrameManager.ts
@@ -108,6 +108,9 @@ export class FrameManager extends EventEmitter {
         );
       }
     );
+    session.on('Page.frameStartedLoading', (event) => {
+      this._onFrameStartedLoading(event.frameId);
+    });
     session.on('Page.frameStoppedLoading', (event) => {
       this._onFrameStoppedLoading(event.frameId);
     });
@@ -285,6 +288,12 @@ export class FrameManager extends EventEmitter {
     if (!frame) return;
     frame._onLifecycleEvent(event.loaderId, event.name);
     this.emit(FrameManagerEmittedEvents.LifecycleEvent, frame);
+  }
+
+  _onFrameStartedLoading(frameId: string): void {
+    const frame = this._frames.get(frameId);
+    if (!frame) return;
+    frame._onLoadingStarted();
   }
 
   _onFrameStoppedLoading(frameId: string): void {
@@ -650,6 +659,10 @@ export class Frame {
    * @internal
    */
   _name?: string;
+  /**
+   * @internal
+   */
+  _hasStartedLoading = false;
 
   /**
    * @internal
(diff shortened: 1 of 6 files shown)


OrKoN (changes requested):
Could you please add the details about what issue is this fixing and why the fix is correct?

src/common/FrameManager.ts
```diff
@@ -287,6 +290,13 @@ export class FrameManager extends EventEmitter {
     this.emit(FrameManagerEmittedEvents.LifecycleEvent, frame);
   }
 
+  _onFrameStartedLoading(frameId: string): void {
+    const frame = this._frames.get(frameId);
+    if (!frame) return;
+    frame._onLoadingStarted();
+    this.emit(FrameManagerEmittedEvents.LifecycleEvent, frame);
```
OrKoN: why do we emit a lifecycle event?

test/frame.spec.ts
```diff
@@ -278,6 +278,16 @@ describe('Frame specs', function () {
         server.PREFIX + '/frames/frame.html?param=value#fragment'
       );
     });
+    it('should support lazy frames', async () => {
```
OrKoN: let's add a similar test to the OOPIF spec

test/frame.spec.ts
```diff
@@ -278,6 +278,16 @@ describe('Frame specs', function () {
...
+      const { page, server } = getTestState();
+
+      await page.setViewport({ width: 1158, height: 1227 });
+      await page.goto(server.PREFIX + '/frames/lazy-frame.html');
+
+      const frames = page.frames();
+      expect(frames.length).toBe(3);
+      expect(frames[frames.length - 1]._hasStartedLoading).toBeFalsy();
```
OrKoN: should we check that the value is true for the rest of the frames?

test/frame.spec.ts
OrKoN: probably needs `itFailsFirefox`

test/frame.spec.ts
```diff
@@ -278,6 +278,19 @@ describe('Frame specs', function () {
...
       );
     });
+    itFailsFirefox('should support lazy frames', async () => {
+      const { page, server } = getTestState();
+
+      await page.setViewport({ width: 1158, height: 1227 });
+      await page.goto(server.PREFIX + '/frames/lazy-frame.html');
+
+      const frames = page.frames();
+      expect(frames.length).toBe(3);
```
OrKoN: nit: what about the following instead of lines 288 - 293:

```
expect(frames.map(f => f._hasStartedLoading)).toEqual([true, true, false]);
```

jrandolf: Done; the previous approach was used because it's slightly more robust.

OrKoN (approved):
LGTM thanks!

PR #8350, opened 2022-05-16
fix: update network-service related chromium feature settings

yutakahirano:

Removing an obsolete Chromium feature, and updating a Chromium feature name.
This shouldn't change the behavior.

No. I believe we don't need that.

N/A

**Summary**

The "NetworkService" chromium feature [has been removed since Jan 2022](https://crrev.com/c/3417192).
The "NetworkServiceInProcess" chromium feature [has just been renamed](https://crrev.com/c/3647768).

This change shouldn't change the behavior.

That depends on which Chromium version you are using. With older chromium, this change is a breaking change.

**Other information**

Is "NetworkServiceInProcess" important for you? We, chromium authors, will be more happy without setting the feature explicitly. Also, on some platforms we don't support the feature.

### DIFF
--- a/src/node/Launcher.ts
+++ b/src/node/Launcher.ts
@@ -207,7 +207,7 @@ class ChromeLauncher implements ProductLauncher {
     const chromeArguments = [
       '--allow-pre-commit-input', // TODO(crbug.com/1320996): neither headful nor headless should rely on this flag.
       '--disable-background-networking',
-      '--enable-features=NetworkService,NetworkServiceInProcess',
+      '--enable-features=NetworkServiceInProcess2',
       '--disable-background-timer-throttling',
       '--disable-backgrounding-occluded-windows',
       '--disable-breakpad',


ergunsh (approved):
LGTM, can you update commit message and do a force push to be compliant with conventional commits?

It can be `chore: fix network-service related chromium feature settings`.

Thank you!

OrKoN:
yutakahirano@ updating the commit format won't be required as we can do that. Thanks for the PR.

yutakahirano:
Is it OK now?

OrKoN:
Yes, thanks! regarding the version, it looks like the rename has landed in M103? Puppeteer is currently running M102 so perhaps we need to wait until we ship M103 by default.

OrKoN:
Regarding if we need NetworkServiceInProcess2, I am not so sure. It seems it was introduced here https://github.com/puppeteer/puppeteer/pull/3738 and it appears it's important for the interception. The original issue has more context https://github.com/puppeteer/puppeteer/issues/3471 As the networkservice is on by default, I suspect we might not need to set the flag anymore.

yutakahirano:
Mostly yes, the renaming change landed originally on M104 but we're planning to merge the change to M103.

PR #8353, opened 2022-05-16
fix: do not use loaderId for lifecycle events

OrKoN:
This PR works around the upstream bug [crbug.com/1325782](crbug.com/1325782). Previously Puppeteer relied on the presence of the loaderId to determine the kind of navigation and expected events. It does not look like there is a reason to do so: instead, we could see what events we get and proceed accordingly.

### DIFF
--- a/src/common/FrameManager.ts
+++ b/src/common/FrameManager.ts
@@ -194,7 +194,6 @@ export class FrameManager extends EventEmitter {
     } = options;
 
     const watcher = new LifecycleWatcher(this, frame, waitUntil, timeout);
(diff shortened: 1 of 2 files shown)


jrandolf-google (approved):

jrandolf-google (changes requested):
Is it possible to add tests?

OrKoN:
There is no change in behaviour, it fixes the failing existing tests (see https://github.com/puppeteer/puppeteer/actions/runs/2330686327)

jrandolf-google (approved):

Issue #8362, opened 2022-05-18
[Bug]: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined


whimboo:
### Bug description

This is a problem that I can currently see locally when trying to run the Puppeteer unit tests with the latest upstream revision. Once it failed it always fails even for installing Puppeteer. The only workaround is to manually remove the `lib/` folder with generated code.

Steps to reproduce the problem:

1. Pull latest changes from upstream repository
2. Checkout revision 82f84c9618772cce1a987804c90e2f10c29dc854
3. Run `npm install` => success
4. Run `npm run funit` => success
5. Checkout e.g revision 7107d2d300f9a5b5c38210163ff5df42d19d8453
6. Run `npm install` => success
7. Run `npm run funit` => failure
8. Run `npm install` => failure

@OrKoN maybe these steps help to analyze the problem. At least for me these always reproduce it.

### Puppeteer version

14.1.0-post

### Node.js version

v17.9.0

### npm version

v8.5.5

### What operating system are you seeing the problem on?

macOS

### Relevant log output

```shell
> puppeteer@14.1.0-post funit
> cross-env PUPPETEER_PRODUCT=firefox npm run unit


> puppeteer@14.1.0-post unit
> npm run tsc-cjs && mocha --config mocha-config/puppeteer-unit-tests.js


> puppeteer@14.1.0-post tsc-cjs
> tsc -b src/tsconfig.cjs.json


✖ ERROR: TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:372:5)
    at validateString (node:internal/validators:120:11)
    at dirname (node:path:1276:5)
    at Object.<anonymous> (/Users/henrik/code/puppeteer/src/constants.ts:4:51)
    at Module._compile (node:internal/modules/cjs/loader:1099:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Object.require.extensions.<computed> [as .js] (/Users/henrik/code/puppeteer/node_modules/ts-node/src/index.ts:1445:43)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18)
[...]

PR #8363, opened 2022-05-18
fix: recompile before funit and unit commands

OrKoN:
Running only tsc-cjs does not seem sufficient for rebuilding
the project after the compat layer for ESM was introduced.
Let's run `tsc-cjs-compact` in addition to the existing command.

Issues: #8362

cc @whimboo

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,9 +28,9 @@
   "scripts": {
     "test-browser": "wtr",
     "test-browser-watch": "wtr --watch",
-    "unit": "npm run tsc-cjs && mocha --config mocha-config/puppeteer-unit-tests.js",
+    "unit": "npm run tsc-cjs && npm run tsc-compat-cjs && mocha --config mocha-config/puppeteer-unit-tests.js",
     "chrome-headless-unit": "cross-env HEADLESS=chrome npm run unit",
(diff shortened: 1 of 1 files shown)


jrandolf-google (approved):

jrandolf-google (changes requested):
Could we run `npm run build` instead? Because the `package.json` for the es modules are not generated with just `npm run tsc`.

OrKoN:
I don't think we need esm modules and dts for the tests though. I am actually thinking if `tsc-compat-cjs` is enough (so that running tests is not super slow).

jrandolf-google (approved):

jrandolf-google:
> I don't think we need esm modules and dts for the tests though. I am actually thinking if `tsc-compat-cjs` is enough (so that running tests is not super slow).

My concern was that node may not run because it may see ES modules even when they aren't used and complain.

OrKoN:
>>I don't think we need esm modules and dts for the tests though. I am actually thinking if tsc-compat-cjs is enough (so that running tests is not super slow).

>My concern was that node may not run because it may see ES modules even when they aren't used and complain

Good point! Let me try with `tsc-compat-cjs` I think the clean up of the lib folder causes failures in the jobs.

whimboo:
Thanks for the fix @OrKoN!

PR #8369, opened 2022-05-19
revert: do not use loaderId for lifecycle events

OrKoN:
Reverts puppeteer/puppeteer#8353 as it breaks Firefox integration.


jrandolf-google (approved):

whimboo:
Can we at the same time backout the disable patch for Firefox in CI as well? Or would this be in a separate PR?

jrandolf-google:
> Can we at the same time backout the disable patch for Firefox in CI as well? Or would this be in a separate PR?

This will be in a separate PR.

PR #8375, opened 2022-05-19
chore: use release-please workflow

jrandolf-google:


.github/workflows/release.yml
```diff
@@ -1,33 +1,85 @@
...
-          title: 'chore(release): mark v${{ steps.release.outputs.VERSION }}'
-          body: ''
-          labels: release
-          reviewers: jrandolf, OrKoN, ergunsh
+          npm run build
+          npm run doc
+      - name: Configure git
+        run: |
+          git config --global user.name 'actions-bot'
+          git config --global user.email '<github-actions-bot@google.com>'
```
OrKoN, line 80: is it a new group email address?

jrandolf: No, this is what the action uses for their own commit:
- https://github.com/google-github-actions/release-please-action/blob/main/.github/workflows/release-please.yaml

It's safe to use for commits.

OrKoN (approved):

PR #8377, opened 2022-05-19
chore: use exit code 0 for doclint file gen

jrandolf-google:
This PR converts some warnings from DocLint to info logs. In particular, it converts warnings about code generation that is intentionally performed (hence not needed to be warned about). In doing this, we can exit with code 0 since there are no warnings and errors.


ergunsh (approved):
:shipit:

OrKoN (approved):

Issue #8361, opened 2022-05-18
Multiple tests started to fail in Firefox


OrKoN:
See https://github.com/puppeteer/puppeteer/runs/6485039988?check_suite_focus=true

it seems the failure happens consistently.

OrKoN:
@whimboo could you please take a look? it seems it's only on Mac. I have not disabled the tests for now as there are multiple of them failing.

whimboo:
@OrKoN is that related to all of these failing tests? We haven't landed anything for our CDP implementation which might have caused that. Since when exactly is it happening? Just with today's Nightly build or did it start to fail earlier?

OrKoN:
I think it started ~20h ago

OrKoN:
This CL has landed https://github.com/puppeteer/puppeteer/pull/8353 23h ago. It looks like there were some flaky failures but it passed after 3 attempts. And then once it was in main the tests started to fail consistently on Mac (I did three full run retries)

whimboo:
I tried to run locally after updating Puppeteer (main). But I'm not able to run and install due to the following failure:

```
node:internal/errors:465
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:372:5)
    at validateString (node:internal/validators:120:11)
    at dirname (node:path:1276:5)
    at Object.<anonymous> (/Users/henrik/code/puppeteer/lib/cjs/puppeteer/constants.js:6:82)
    at Module._compile (node:internal/modules/cjs/loader:1099:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1153:10)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:102:18) {
  code: 'ERR_INVALID_ARG_TYPE'
  ```

If you could run locally and check with the trace logging enabled I would appreciate. I clearly would need more details or the above failure fixed.

OrKoN:
[...]

PR #8394, opened 2022-05-25
chore: enable anchor navigation tests for Firefox

juliandescottes:
New puppeteer tests started passing on mozilla-central's CI after landing https://bugzilla.mozilla.org/show_bug.cgi?id=1636453 which should now be included in the latest Nightly.

It would be great to have them enabled in puppeteer's CI as well, assuming they work on all platforms there (mozilla-central only runs puppeteer tests on linux)

Related to #8361


juliandescottes:
cc @OrKoN

OrKoN (approved):

PR #8395, opened 2022-05-25
fix: do not use loaderId for lifecycle events

OrKoN:
This PR works around the upstream bug [crbug.com/1325782](crbug.com/1325782). Previously Puppeteer relied on the presence of the loaderId to determine the kind of navigation and expected events. It does not look like there is a reason to do so: instead, we could see what events we get and proceed accordingly.

Reland of https://github.com/puppeteer/puppeteer/pull/8353


OrKoN:
Additional tests for Firefox have been enabled in https://github.com/puppeteer/puppeteer/pull/8394

jrandolf-google (approved):

PR #8401, opened 2022-05-25
feat: use strict typescript

jrandolf-google:
Issues: #6769


OrKoN (changes requested):
It looks like the installation tests are failing. Let's separate package.json and dependency updates into a separate CL and also split the code changes into multiple CLs for easier review.

src/common/AriaQueryHandler.ts
```diff
@@ -78,17 +88,21 @@ const queryOne = async (
...
   const { name, role } = parseAriaSelector(selector);
   const res = await queryAXTree(exeCtx._client, element, name, role);
-  if (res.length < 1) {
+  if (!res[0] || !res[0].backendDOMNodeId) {
     return null;
   }
   return exeCtx._adoptBackendNodeId(res[0].backendDOMNodeId);
 };
 
+declare global {
```
OrKoN: Let's not add the types to the global namespace if they are executing/available in the page context?

src/common/Browser.ts
```diff
@@ -572,33 +572,24 @@ export class Browser extends EventEmitter {
...
     } finally {
-      this.removeListener(BrowserEmittedEvents.TargetCreated, check);
-      this.removeListener(BrowserEmittedEvents.TargetChanged, check);
+      this.off(BrowserEmittedEvents.TargetCreated, check);
+      this.off(BrowserEmittedEvents.TargetChanged, check);
     }
 
     async function check(target: Target): Promise<void> {
-      if (await predicate(target)) resolve(target);
+      if (!isResolved && (await predicate(target))) {
```
OrKoN: this looks like a race condition? multiple promises can enter the branch  while the predicate is being evaluated

src/common/DOMWorld.ts
```diff
@@ -981,6 +981,7 @@ async function waitForPredicatePageFunction(
   if (polling === 'raf') return await pollRaf();
   if (polling === 'mutation') return await pollMutation();
   if (typeof polling === 'number') return await pollInterval(polling);
+  assert(false);
```
OrKoN: Not needed?

src/common/Debug.ts
```diff
@@ -16,6 +16,11 @@
 
 import { isNode } from '../environment.js';
 
+declare global {
+  // eslint-disable-next-line no-var
+  var __PUPPETEER_DEBUG: string;
```
OrKoN, line 21: can it be `const`? (not sure what TS supports here)

jrandolf: It must be `var` because it is in the global scope.

src/common/FrameManager.ts
```diff
@@ -73,7 +73,7 @@ export class FrameManager extends EventEmitter {
   private _frames = new Map<string, Frame>();
   private _contextIdToContext = new Map<string, ExecutionContext>();
   private _isolatedWorlds = new Set<string>();
-  private _mainFrame: Frame;
+  private _mainFrame!: Frame;
```
OrKoN: where is _mainFrame set? it looks like whether we have it is dependent on CDP. Let's keep it optional here and assert it when we need to use it?

src/common/JSHandle.ts
```diff
@@ -402,6 +403,7 @@ export class ElementHandle<
     } = {}
   ): Promise<ElementHandle | null> {
     const frame = this._context.frame();
+    if (!frame) return null;
```
OrKoN: let's assert and error instead as it should not happen?

src/common/JSHandle.ts
```diff
@@ -114,7 +114,7 @@ const applyOffsetsToQuad = (
  *
  * @public
  */
-export class JSHandle<HandleObjectType = unknown> {
+export class JSHandle<HandleObjectType = any> {
```
OrKoN: I think unknown is much safer? Basically any allows us to do anything with the type (including what's not possible). Let's make it work with unknown.

src/common/JSHandle.ts
```diff
@@ -475,6 +477,7 @@ export class ElementHandle<
     } = {}
   ): Promise<ElementHandle | null> {
     const frame = this._context.frame();
+    if (!frame) return null;
```
OrKoN: same here

src/common/JSHandle.ts
```diff
@@ -952,10 +956,10 @@ export class ElementHandle<
 
     const { offsetX, offsetY } = await this._getOOPIFOffsets(this._frame);
     const quad = result.model.border;
-    const x = Math.min(quad[0], quad[2], quad[4], quad[6]);
-    const y = Math.min(quad[1], quad[3], quad[5], quad[7]);
-    const width = Math.max(quad[0], quad[2], quad[4], quad[6]) - x;
-    const height = Math.max(quad[1], quad[3], quad[5], quad[7]) - y;
+    const x = Math.min(quad[0]!, quad[2]!, quad[4]!, quad[6]!);
```
OrKoN, line 949: can we narrow the type by checking the length instead of doing `!`?

src/common/JSHandle.ts
```diff
@@ -1069,6 +1073,9 @@ export class ElementHandle<
   ): Promise<ElementHandle<T> | null> {
     const { updatedSelector, queryHandler } =
       getQueryHandlerAndSelector(selector);
+    if (!queryHandler.queryOne) {
```
OrKoN: why not assert like for queryAllArray?

src/common/JSHandle.ts
```diff
@@ -1081,6 +1088,9 @@ export class ElementHandle<
   ): Promise<Array<ElementHandle<T>>> {
     const { updatedSelector, queryHandler } =
       getQueryHandlerAndSelector(selector);
+    if (!queryHandler.queryAll) {
```
OrKoN: ditto

src/common/JSHandle.ts
jrandolf: This is part of `https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess` which I've enabled. It helps assert that we know what we are doing when we use `!`.

Checking length does not assert this.

OrKoN: I think it does not work quite nice for arrays then. Can we change the quad type to be `[number, number x 4]` then? That should help us avoid all those exclamations? The quad array is fixed size iirc

src/common/JSHandle.ts
jrandolf: This is more robust. Do we want to error? What's the expectation?

src/common/JSHandle.ts
jrandolf: See above comment.

src/common/JSHandle.ts
OrKoN: why is it different from the handling of queryAllArray below?

jrandolf: For that specific instance, there is no canonical choice for a return value, so I assert there.

OrKoN: I think it'd be better to expect an error in this case. If you do $(`aria/smth`) and aria query handler does not support queryOne, I think it's better to error then pretend that the element is not found.

src/common/JSHandle.ts
jrandolf: It is, but the type itself is generated.

OrKoN: hm, I see then.

src/common/FrameManager.ts
```diff
@@ -261,9 +265,10 @@ export class FrameManager extends EventEmitter {
     }
 
     const frame = this._frames.get(event.targetInfo.targetId);
-    const session = Connection.fromSession(this._client).session(
-      event.sessionId
-    );
+    const connection = Connection.fromSession(this._client);
+    if (!connection) return;
```
OrKoN: should we throw an error here instead? It looks like we normally expect the connection and session below to be there (unless smth goes very wrong).

OrKoN (approved):

PR #8446, opened 2022-05-31
chore: update deps

jrandolf-google:
This PR provides some updates to the dependencies. Also fixes some new warnings on documentation given by `@microsoft/api-extractor`.


src/common/PDFOptions.ts
```diff
@@ -24,6 +24,9 @@ export interface PDFMargin {
   right?: string | number;
 }
 
+/**
```
ergunsh, line 27: I wonder what is our stance of making changes that doesn't directly relate to the aim of the PR.

For example, why this change to the documentation introduced I'm not able to understand from the title.

I'd prefer to keep unrelated changes in separate PRs, commits. WDYT?

ergunsh (approved):
LGTM with nit

src/common/PDFOptions.ts
jrandolf: This is related. The `@microsoft/api-extractor` warns about this interface being exported without a comment above.

ergunsh: Oh, then it would be great to have this information provided in the PR description :)

PR #8447, opened 2022-05-31
feat: support node 18

jrandolf-google:
This CL patches some files to fully support Node 18.

### DIFF
--- a/src/node/NodeWebSocketTransport.ts
+++ b/src/node/NodeWebSocketTransport.ts
@@ -16,9 +16,21 @@
 import NodeWebSocket from 'ws';
 import { ConnectionTransport } from '../common/ConnectionTransport.js';
 import { packageVersion } from '../generated/version.js';
+import { promises as dns } from 'dns';
 
 export class NodeWebSocketTransport implements ConnectionTransport {
-  static create(url: string): Promise<NodeWebSocketTransport> {
+  static async create(urlString: string): Promise<NodeWebSocketTransport> {
+    // TODO(jrandolf): Starting in Node 17, IPv6 is favoured over IPv4 due to a change
+    // in a default option:
+    // - https://github.com/nodejs/node/issues/40537,
+    // Due to this, we parse and resolve the hostname manually with the previous
+    // behavior according to:
+    // - https://nodejs.org/api/dns.html#dnslookuphostname-options-callback
+    // because of https://bugzilla.mozilla.org/show_bug.cgi?id=1769994.
+    const url = new URL(urlString);
+    const { address } = await dns.lookup(url.hostname, { verbatim: false });
+    url.hostname = address;
+
     return new Promise((resolve, reject) => {
       const ws = new NodeWebSocket(url, [], {
         followRedirects: true,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
       matrix:
         # Include all major maintenance + active LTS + current Node.js versions.
         # https://github.com/nodejs/Release#release-schedule
-        node: [14, 16]
+        node: [14, 16, 18]
     steps:
       - name: Checkout
         uses: actions/checkout@v3


src/node/NodeWebSocketTransport.ts
```diff
@@ -20,7 +20,7 @@ import { packageVersion } from '../generated/version.js';
 export class NodeWebSocketTransport implements ConnectionTransport {
   static create(url: string): Promise<NodeWebSocketTransport> {
     return new Promise((resolve, reject) => {
-      const ws = new NodeWebSocket(url, [], {
+      const ws = new NodeWebSocket(url.replace('localhost', '127.0.0.1'), [], {
```
jrandolf: @whimboo @OrKoN For some context, starting from Node 17, Node prefers to use `::1` instead of `127.0.0.1` when connecting to `localhost`. Because of this, there are some `ECONNREFUSED` errors while connecting to Firefox's websocket endpoint (probably because Firefox doesn't listen on iPv6 ::1).

As a result, this is the solution I came up with, which is not too far from the previous implementation.

P.S. I do not understand why Node doesn't try both ports instead of giving an error, but that was their decision :P

whimboo:
Please note that with node.js 17.0 the DNS resolution has been changed to default to IPv6 instead of IPv4. This could cause problems for users on dual-stack systems. For Firefox we have https://bugzilla.mozilla.org/show_bug.cgi?id=1769994 on file to get a better support. But not sure if Chrome might also be affected.

src/node/NodeWebSocketTransport.ts
whimboo: I assume this is to workaround the DNS resolution issue? Note that this would break users on ipv6-only systems.

See also https://github.com/nodejs/node/issues/40537

jrandolf: From looking at the bug on bugzilla, it seems this would be the case (for firefox at least) anyway since firefox only listens in on IPv4. We don't have any issues here at Chrome based on Chrome-only tests.

jrandolf: By the way, I've written a comment documenting the reason above. Not sure if it's visible to you.

jrandolf: In any case, thanks for the link (gave me some insight into the issue). I just hard-coded the previous behavior based on
- https://github.com/nodejs/node/issues/40537, and
- https://nodejs.org/api/dns.html#dnslookuphostname-options-callback

src/node/NodeWebSocketTransport.ts
```diff
@@ -16,9 +16,14 @@
 import NodeWebSocket from 'ws';
 import { ConnectionTransport } from '../common/ConnectionTransport.js';
 import { packageVersion } from '../generated/version.js';
+import { promises as dns } from 'dns';
 
 export class NodeWebSocketTransport implements ConnectionTransport {
-  static create(url: string): Promise<NodeWebSocketTransport> {
+  static async create(urlString: string): Promise<NodeWebSocketTransport> {
+    const url = new URL(urlString);
+    const { address } = await dns.lookup(url.hostname, { verbatim: false });
```
OrKoN, line 31: why do we need an explicit DNS lookup? please link to the issue/documentation describing it (I assume it's needed by Node 18). Why does not the NodeWebSocket lib not handle that?

jrandolf: I've already commented on this. Are you able to see my comments?

jrandolf-google:
@whimboo @OrKoN I realized that I needed to submit my comments for you to see them... thanks @OrKoN....

src/node/NodeWebSocketTransport.ts
OrKoN: could you please include that as a comment in the code here to make sure we don't loose the context? (I still have not understood the issue though)

jrandolf: Done.

OrKoN (approved):

whimboo:
@jrandolf maybe you can give a quick feedback on how this would work on a IPv6 only system? Would it still work in case of Chrome?

jrandolf-google:
> @jrandolf maybe you can give a quick feedback on how this would work on a IPv6 only system? Would it still work in case of Chrome?

The implementation basically hardcodes the previous behavior, so however Node handled IPv6 systems before, the current behavior is the same.

Specifically, the resolver just has preference for IPv4, but if it cannot resolve IPv4, it resolves to an IPv6 address.

PR #8468, opened 2022-06-03
fix: robustly check for launch executable

### LINKED ISSUE #8466: [Bug]: Crash with a specified browser channel
### Bug description

1. Launch Puppeteer with a custom browser channel:
```

(async () => {
  const browser = await puppeteer.launch({channel: "chrome-dev"});
})();
```

# Expected
Browser is launched.

# Actual
Puppeteer crashes.

### Puppeteer version

14.2.1

### Node.js version

v18.0.0

### npm version

8.6.0

macOS

### Relevant log output

```shell
.../node_modules/puppeteer/lib/cjs/puppeteer/common/assert.js:26
        throw new Error(message);
              ^

Error: `executablePath` must not be specified when `channel` is given.
    at assert (.../node_modules/puppeteer/lib/cjs/puppeteer/common/assert.js:26:15)
    at ChromeLauncher.launch (.../node_modules/puppeteer/lib/cjs/puppeteer/node/Launcher.js:97:36)
    at async /Users/sadym/work/del_me/example.js:4:19

Node.js v18.0
```

jrandolf-google:
Fixes: #8466

### DIFF
--- a/src/node/Launcher.ts
+++ b/src/node/Launcher.ts
@@ -71,8 +71,8 @@ class ChromeLauncher implements ProductLauncher {
       ignoreDefaultArgs = false,
       args = [],
       dumpio = false,
-      channel = null,
-      executablePath = null,
+      channel,
+      executablePath,
       pipe = false,
       env = process.env,
       handleSIGINT = true,
@@ -83,7 +83,7 @@ class ChromeLauncher implements ProductLauncher {
       slowMo = 0,
       timeout = 30000,
       waitForInitialPage = true,
-      debuggingPort = null,
+      debuggingPort,
     } = options;
 
     const chromeArguments = [];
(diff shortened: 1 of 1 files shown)


src/node/Launcher.ts
```diff
@@ -133,26 +133,9 @@ class ChromeLauncher implements ProductLauncher {
 
     isTempUserDataDir = false;
 
-    let chromeExecutable = executablePath;
-
-    if (channel) {
-      // executablePath is detected by channel, so it should not be specified by user.
-      assert(
-        typeof executablePath === 'string',
```
OrKoN, line 141: I think we should keep the error message? to hint users we they specify both options by mistake

OrKoN (approved):

PR #8472, opened 2022-06-03
fix: don't throw on bad access

### LINKED ISSUE #8470: [Bug]: Can't upload files when puppeteering a remote browser
### Bug description

1. Connect to a remote browser that does not share the file system with environment executing the script (could be in a container or another machines)
2. Try to upload a file using fileChooser.accpet
3. Fail.

This seems to be due the following code in /puppeteer/lib/cjs/puppeteer/common/JSHandle.js
uploadFile()
.
.
        // Locate all files and confirm that they exist.
        const files = await Promise.all(filePaths.map(async (filePath) => {
            const resolvedPath = path.resolve(filePath);
            try {
                await fs.promises.access(resolvedPath, fs.constants.R_OK);
            }
            catch (error) {
                if (error.code === 'ENOENT')
                    throw new Error(`${filePath} does not exist or is not readable`);
            }
            return resolvedPath;
        }));
        

This code seems to be trying to access the local file system of where the Puppeteer script is running to determine if the files are present.  Since the browser is running on a different file system this does not go well.  
There should be some way to tell puppeteer that we are on a remote system and can't just use fs.
Commenting this code out seems to work.

### Puppeteer version

12.0.1

### Node.js version

16.13

### npm version

8.5.0

Linux

### Relevant log output

```shell
Error: C:\Users\vess\Downloads\download.csv does not exist or is not readable
    at /home/vess/dev/addon-puppeteer/node_modules/puppeteer/lib/cjs/puppeteer/common/JSHandle.js:531:27
    at async ElementHandle.uploadFile (/home/vess/dev/addon-puppeteer/node_modules/puppeteer/lib/cjs/puppeteer/common/JSHandle.js:524:23)
    at async FileChooser.accept (/home/vess/dev/addon-puppeteer/node_modules/puppeteer/lib/cjs/puppeteer/common/FileChooser.js:61:9)
    at async PageController.uploadFile (/home/vess/dev/addon-puppeteer/dist/page/navigate.controller.js:416:13)
    at async /home/vess/dev/addon-puppeteer/node_modules/@nestjs/core/router/router-execution-context.js:46:28
    at async /home/vess/dev/addon-puppeteer/node_modules/@nestjs/core/router/router-proxy.js:9:17
```

jrandolf-google:
This PR removes a throw clause that is invalid when a script is run locally and connected remotely.

Fixes: #8470

### DIFF
--- a/src/common/JSHandle.ts
+++ b/src/common/JSHandle.ts
@@ -801,9 +801,12 @@ export class ElementHandle<
   /**
    * This method expects `elementHandle` to point to an
    * {@link https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input | input element}.
+   *
    * @param filePaths - Sets the value of the file input to these paths.
-   *    If some of the  `filePaths` are relative paths, then they are resolved
-   *    relative to the {@link https://nodejs.org/api/process.html#process_process_cwd | current working directory}
+   *    If a path is relative, then it is resolved against the
+   *    {@link https://nodejs.org/api/process.html#process_process_cwd | current working directory}.
+   *    Note for locals script connecting to remote chrome environments,
+   *    paths must be absolute.
    */
   async uploadFile(...filePaths: string[]): Promise<void> {
     const isMultiple = await this.evaluate<(element: Element) => boolean>(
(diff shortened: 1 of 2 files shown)


src/common/JSHandle.ts
```diff
@@ -837,8 +837,11 @@ export class ElementHandle<
         try {
           await fs.promises.access(resolvedPath, fs.constants.R_OK);
```
jrandolf: @OrKoN  Should we just remove this check in its entirety? The browser will throw itself if it really doesn't exist, right?

OrKoN: Do we have a history for why it was introduced? Could we distinguish if the browser is running locally or not and validate accordingly?

jrandolf: I don't think validating here is useful, but I'll investigate higher in the stack. Also, rather than validating existence, how about asserting that an absolute path must be given?

OrKoN: the api doc says that we would resolve relative paths too. We could require an absolute path if we know that the browser is remote though.

src/common/JSHandle.ts
```diff
@@ -801,9 +801,12 @@ export class ElementHandle<
   /**
    * This method expects `elementHandle` to point to an
    * {@link https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input | input element}.
+   *
    * @param filePaths - Sets the value of the file input to these paths.
-   *    If some of the  `filePaths` are relative paths, then they are resolved
-   *    relative to the {@link https://nodejs.org/api/process.html#process_process_cwd | current working directory}
+   *    If a path is relative, then they it is resolved relative to the
```
OrKoN: `then they it` looks like a typo?

OrKoN (approved):

PR #8477, opened 2022-06-06
fix: Kill browser process when killing process group fails

### LINKED ISSUE #8476: [Bug]: 13.4.0 regression caused by permission issues when killing browser process group
### Bug description

I'm having trouble creating a simplified reproduction of this, including one which uses my same Jest setup. My scenario is the following:

1. In (Jest) unit tests for a Node.js server which uses Puppeteer, I launch a new server + corresponding Puppeteer instance in each test case and tear down the server/browser at the end of each test.
2. After upgrading to Puppeteer 13.4.0, test teardown fails with:
> Puppeteer was unable to kill the process which ran the browser binary.
    This means that, on future Puppeteer launches, Puppeteer might not be able to launch the browser.
    Please check your open processes and ensure that the browser processes that Puppeteer launched have been killed.
    If you think this is a bug, please report it on the Puppeteer issue tracker.
    Error cause: Error: kill EPERM

This appears to be a regression from #6859. A similar fix to https://github.com/puppeteer/puppeteer/pull/8352 for MacOS/Linux seems like it may be necessary in some scenarios.

### Puppeteer version

13.4.0

### Node.js version

14.16.1

### npm version

6.14.12

macOS

### Relevant log output

_No response_

jimmydief:

Bug fix.

I did not add any – seems like this area is missing tests in general. I'm having trouble finding a simplified reproduction of this outside of my private project.

N/A

**Summary**

Fixes #8476 which seems to have been a regression from #6859. Similar fix to #8352 which only covered Windows.

No.

**Other information**

### DIFF
--- a/src/node/BrowserRunner.ts
+++ b/src/node/BrowserRunner.ts
@@ -195,7 +195,7 @@ export class BrowserRunner {
             if (error) {
               // taskkill can fail to kill the process e.g. due to missing permissions.
               // Let's kill the process via Node API. This delays killing of all child
-              // proccesses of `this.proc` until the main Node.js process dies.
+              // processes of `this.proc` until the main Node.js process dies.
               proc.kill();
             }
           });
@@ -203,7 +203,15 @@ export class BrowserRunner {
           // on linux the process group can be killed with the group id prefixed with
           // a minus sign. The process group id is the group leader's pid.
           const processGroupId = -this.proc.pid;
-          process.kill(processGroupId, 'SIGKILL');
+
+          try {
+            process.kill(processGroupId, 'SIGKILL');
+          } catch (error) {
+            // Killing the process group can fail due e.g. to missing permissions.
+            // Let's kill the process via Node API. This delays killing of all child
+            // processes of `this.proc` until the main Node.js process dies.
+            proc.kill('SIGKILL');
+          }
         }
       } catch (error) {
         throw new Error(


src/node/BrowserRunner.ts
```diff
@@ -195,15 +195,23 @@ export class BrowserRunner {
             if (error) {
               // taskkill can fail to kill the process e.g. due to missing permissions.
               // Let's kill the process via Node API. This delays killing of all child
-              // proccesses of `this.proc` until the main Node.js process dies.
+              // processes of `this.proc` until the main Node.js process dies.
               proc.kill();
```
Kikobeats, line 199: @jimmydief is `SIGKILL` missing there in an intentional way?

jimmydief: @Kikobeats I had wondered the same. FYI @szuend who made the original change in #8352.

szuend: `SIGKILL` is not a thing on windows. So if I remember the Node.js documentation correctly a call to `kill` without any signal is equivalent to SIGKILL.

Kikobeats: oh, thanks for clarifying.

I considered the kill logic enough important to be alive as a separate package:
http://npm.im/kill-process-group

I was wondering if puppeteer will accept it if I send a PR; happy to invite anyone interested as contributor of the repo.

PR #8487, opened 2022-06-08
Update fetch.ts

frank-dspeed:
Nodejs now also got fetch so i changed the logic in a way that it works with old nodejs as also the current one 

sideEffect reduces the need to check if we run in node which is well as this is the common folder

**Summary**

**Other information**


jrandolf-google:
Closed in favor of #8490.

PR #8492, opened 2022-06-09
chore: extract public API for JSHandle

OrKoN:
Blocked by the docklint/doc generation.


jrandolf-google (approved):

OrKoN:
We are not gonna land any of this until the docs/linting is solved. So it's just an example of how we plan to re-structure puppeteer.

jrandolf-google (commented):
Can we
- rename `Types.ts` to `types.ts` and place this in common?

PR #8495, opened 2022-06-09
chore: use c8 coverage

jrandolf-google:
This PR adds c8 for our coverage. It also removes browser tests as browser support is no longer supported.


.github/workflows/ci.yml
```diff
@@ -76,7 +76,6 @@ jobs:
           CHROMIUM: true
         run: |
           xvfb-run --auto-servernum npm run test:unit:coverage
-          xvfb-run --auto-servernum npm run test:coverage
```
OrKoN, line 79: I think the previous version would throw an error if something is not covered by tests? Can we have something similar in CI still?

OrKoN: also does not it cause failures because c8 instruments the code that we run in the page context via CDP?

OrKoN (approved):

PR #8498, opened 2022-06-09
fix: remove `instanceof Error` check causing issues in Jest

### LINKED ISSUE #8497: [Bug]: 13.4.0 regression caused when process does not exist when running in Jest
### Bug description

Similar to #8476, I can only reproduce this when killing Puppeteer as part of the teardown logic of my Jest tests. The stack trace looks like this:
```
kill ESRCH
  at pidExists (node_modules/puppeteer/src/node/BrowserRunner.ts:331:20)
  at BrowserRunner.kill (node_modules/puppeteer/src/node/BrowserRunner.ts:190:39)
  at node_modules/puppeteer/src/node/BrowserRunner.ts:177:14
```

I believe that this is likely only reproducible in Jest due to https://github.com/facebook/jest/issues/2549. Regardless, the `instanceof Error` check here seems unnecessary. Given that Puppeteer is likely widely used in Jest, fixing this here seems practical.

https://github.com/puppeteer/puppeteer/blob/465a7c405f01fcef99380ffa69d86042a1f5618f/src/node/BrowserRunner.ts#L333

### Puppeteer version

13.4.0

### Node.js version

14.16.1

### npm version

6.14.12

macOS

### Relevant log output

_No response_

jimmydief:

Bug fix.

No.

N/A

**Summary**

Fixes #8497.

No

**Other information**

### DIFF
--- a/src/node/BrowserRunner.ts
+++ b/src/node/BrowserRunner.ts
@@ -330,12 +330,10 @@ function pidExists(pid: number): boolean {
   try {
     return process.kill(pid, 0);
   } catch (error) {
-    if (error instanceof Error) {
-      const err = error as NodeJS.ErrnoException;
-      if (err.code && err.code === 'ESRCH') {
-        return false;
-      }
+    if ((error as NodeJS.ErrnoException).code === 'ESRCH') {
+      return false;
     }
+
     throw error;
   }
 }


src/node/BrowserRunner.ts
```diff
@@ -330,12 +330,10 @@ function pidExists(pid: number): boolean {
   try {
     return process.kill(pid, 0);
   } catch (error) {
-    if (error instanceof Error) {
```
jimmydief, line 333: I believe that the `instanceof Error` check here is failing when running Puppeteer in Jest due to https://github.com/facebook/jest/issues/2549. This seems to be timing-dependent in my tests and I've had a hard time reproducing it outside of CI.

Given that `process.kill(...)` is part of the standard library, checking that it throws `Error`s seems unnecessary anyways.

jrandolf-google:
Closed in favor of https://github.com/puppeteer/puppeteer/pull/8504

PR #8510, opened 2022-06-13
chore: move `helper.js` to `util.js`

jrandolf-google:
This PR moves `helper.js` to `util.js` and removes the default export. It also removes `unicorn` which enforces catch errors to be named `error` which could result in shadowing.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -16,7 +16,7 @@ module.exports = {
 
   parser: '@typescript-eslint/parser',
 
-  plugins: ['mocha', '@typescript-eslint', 'unicorn', 'import'],
+  plugins: ['mocha', '@typescript-eslint', 'import'],
 
   extends: ['plugin:prettier/recommended'],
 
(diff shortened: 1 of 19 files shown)


.eslintrc.js
```diff
@@ -16,7 +16,7 @@ module.exports = {
 
   parser: '@typescript-eslint/parser',
 
-  plugins: ['mocha', '@typescript-eslint', 'unicorn', 'import'],
```
OrKoN, line 19: pls include in the PR description why we remove unicorn

jrandolf: This is part of some `eslint` changes.

OrKoN: I am curious what changes and why they were needed? the PR description/title only mentions the helper.js deprecation. Please make the description/up-to-date with the actual changes (perhaps leave a comment: as a drive-by fix: disabling the eslint rules because xyz). Also, perhaps eslint changes could be a separate PR?

OrKoN (approved):

Issue #7990, opened 2022-02-10
[Bug]: Some OOPIF-related tests are flaky 


OrKoN:
### Bug description

See https://github.com/puppeteer/puppeteer/runs/5137856862?check_suite_focus=true

### Puppeteer version

Latest

### Node.js version

Any

### npm version

Any

### What operating system are you seeing the problem on?

Windows

### Relevant log output

```shell
https://github.com/puppeteer/puppeteer/runs/5137856862?check_suite_focus=true
```

OrKoN:
cc @jrandolf

OrKoN:
I debugged one particular test called `should support frames within OOP frames`. There are two frames in the test: frame A and frame B. Frame B is a local frame inside frame A. Sometimes the backend events arrive in the following order: frame A attached, frame B attached, frame A navigated, frame B navigated. The _onFrameNavigated in FrameManager recursively removes existing child frames, thus, the frame B is removed. Since the attached event has already happened it is never re-attached and the call handling the event "frame B navigated" hits an assertion.

I think we can only solve it by rewriting the code to use setAutoAttachRelated and runIfWaitingForDebugger. Otherwise, the order of events in iframes is rather unpredictable.

Issue #3667, opened 2018-12-13
Intercept target creation


aslushnikov:
In many ways, users want to intercept targets being created to attach and set them up.

Usecases:
- when popup is getting opened, attach to it and enable request interception
- when a link click opens a new page, set proper device emulation before website is getting loaded

We might be able to do this with CDP using `Target.setAutoAttach` and `waitForDebugger` option. I'd like this to be scoped to browser context though so that's there's a better flexibility.

The API might look like this:

```js
await browserContext.setTargetInterception(true);
browserContext.on('targetcreated', async target => {
  if (target.type() !== 'page') {
    await target.resumeLoading();
    return;
  }
  const page = await target.page();
  await page.setViewport({width: 400, height: 400});
  await target.resumeLoading();
});
```

Related issues: #1378, #3648

nemesisqp:
Is there any workaround for this problem in mean time ? I tried to apply stealth using extension instead of `puppeteer-extra-plugin-stealth` but it not work for new tab opened implicit, right now I testing intercept request that create new tab, blocking and save url, headers data then recreate new tab apply conceal then  goto using that data

rotimi-best:
I am sorry this might be a silly question but can you show how `browserContext` is called. I did something like this
```javascript
const browserContext = await browser.defaultBrowserContext()
//Error: browserContext.setTargetInterception is not a function
await browserContext.setTargetInterception(true);
```

What is the right way of defining `browserContext`?

swang2019:
> I am sorry this might be a silly question but can you show how `browserContext` is called. I did something like this
> 
> ```js
> const browserContext = await browser.defaultBrowserContext()
> //Error: browserContext.setTargetInterception is not a function
> await browserContext.setTargetInterception(true);
> ```
> 
> What is the right way of defining `browserContext`?

It's not a working code, the method is a feature request that not yet be implemented: 

look the OP's word:

> The API **might** look like this:

rotimi-best:
@swang2019 hahaha. Thanks. I get it. I can't wait for this feature to be out.

saberworks:
[...]

PR #8520, opened 2022-06-14
feat: use CDP's auto-attach mechanism

### LINKED ISSUE #8507: [Bug]: Puppeteer doesn't capture some OOPIF requests (Flaky)
### Bug description

The following script will exit 1 if there were some missing `Network.requestWillBeSent` events:
https://gist.github.com/adamraine/e9ae2e811bde239db24a47535008e228

On my computer the script exits 1 most of the time. Other machines have a hard time reproducing the described result. I have reproduced it on a Linux cloudtop as well.

The script needs `waitForDebuggerOnStart: true` so it has time to enable network instrumentation before the OOPIF starts loading. The problem is that setting `waitForDebuggerOnStart: true` in this script doesn't affect anything because puppeteer sets `waitForDebuggerOnStart: false` (See log)

@paulirish and I developed a "fix" that prevents this script from exiting 1, but unsure how it would affect the rest of puppeteer. It's also possible that this "fix" just makes the script fail less often:
https://github.com/puppeteer/puppeteer/compare/main...adamraine:wait-for-debugger-on-attach

### Puppeteer version

14.3.0

### Node.js version

14.18.2

### npm version

8.4.0

macOS

### Relevant log output

```shell
https://gist.github.com/adamraine/5b06ee25b09c815dee73962255c76dc1
```

OrKoN:
In this PR, we refactor Puppeteer to make use of the CDP's auto-attach mechanism. This allows the backend to pause
new targets and give Puppeteer a chance to configure them properly. This fixes the flakiness related to dealing with
OOPIFs and should fix some other issues related to the network interception and navigations. If those are not fixed completely by this PR, the PR serves a solid base for fixing them.

Closes https://github.com/puppeteer/puppeteer/issues/8507, https://github.com/puppeteer/puppeteer/issues/7990
Unlocks https://github.com/puppeteer/puppeteer/issues/3667

BEGIN_COMMIT_OVERRIDE
feat: use CDP's auto-attach mechanism
BREAKING CHANGE: With Chromium, Puppeteer will now attach to page/iframe targets immediately to allow reliable configuration of targets.
END_COMMIT_OVERRIDE

BREAKING CHANGE: With Chromium, Puppeteer will now attach to page/iframe targets immediately to allow reliable configuration of targets.


src/common/Browser.ts
```diff
@@ -286,6 +292,18 @@ export class Browser extends EventEmitter {
         return true;
       });
     this.#setIsPageTargetCallback(isPageTargetCallback);
+    this.#targetManager =
+      product === 'chrome'
```
jrandolf: Can we use an if-else for readability.

jrandolf-google (approved):

test/src/target.spec.ts
```diff
@@ -173,7 +173,7 @@ describe('Target', function () {
       });
       await page.evaluate(() => {
         return (globalThis as any).registrationPromise.then(
```
jrandolf: Don't use `any`. Please use specific types.

test/src/target.spec.ts
```diff
@@ -196,6 +196,7 @@ describe('Target', function () {
       })
     ).toBe('[object ServiceWorkerGlobalScope]');
   });
+  // TODO: fix shared workers.
```
OrKoN: remove this TODO.

PR #8524, opened 2022-06-14
chore: enable strict mode in tests

jrandolf-google:
This PR does a few things:

- The testing `tsconfig.json` inherits from the base TS config.
  - A lot of type assertions have been inserted...a lot.
- All testing utilities have migrated to TS.
- `text-diff` is being replaced with `diff` for TS compatibility.
- `ProtocolError` has been added to `PuppeteerErrors` and `PuppeteerErrors` is no longer a record (it's been frozen).
- Fixes a small bug where `null` was an allowable media type in emulation (should be `undefined`).

Issues: #6769


test/src/ariaqueryhandler.spec.ts
```diff
@@ -94,8 +98,10 @@ describeChromeOnly('AriaQueryHandler', () => {
       await page.setContent(
         '<div id="div"><button id="btn" role="button">Submit</button></div>'
       );
-      const button = await page.$('aria/[role="button"]');
-      const id = await button.evaluate((button: Element) => button.id);
+      const button = (await page.$('aria/[role="button"]'))!;
```
OrKoN, line 101: should we assert instead of `!` in situations like this one? also `!` in `button!` on the line below seems redundant.

OrKoN (approved):

test/src/ariaqueryhandler.spec.ts
jrandolf: There might be some redundancy because most of the changes were found through regular expressions. In particular, using `!` will guarantee that if it doesn't exist, it will break.

Issue #8485, opened 2022-06-07
[Types] Improve types for `evaluate` and `evaluateHandle` on various types (and `JSHandle<T>`)


stevenwdv:
Currently the `evaluate` and `evaluateHandle` functions and `JSHandle<T>` itself are very loosely typed, for example, this is allowed:

```ts
async function demo(handle1: JSHandle<number>, handle2: JSHandle<string>) {
	const notActuallyBoolean: JSHandle<'wait what'> = await handle1.evaluateHandle<JSHandle<boolean>>(
		  (num, notActuallyNumber: number, notActuallyString: string) => 'hi', handle2, 42);
}
```

**Problems are:**
- Argument types are not checked (except for the subject handle itself).
- Return type is not checked for `evaluateHandle`.
- `JSHandle<A> extends JSHandle<B>` for any `A`,`B`. Same for `ElementHandle<E>`.

# Solution

This could be solved by typing the functions as follows:

```ts
type Passed<H extends SerializableOrJSHandle> = H extends JSHandle<infer U> ? U : H;
type PassedArgs<Args extends SerializableOrJSHandle[]> = { [K in keyof Args]: Passed<Args[K]> };

class JSHandle<HandleObjectType> {
  /** Make sure that not e.g. `JSHandle<string> extends `JSHandle<number>` */
  private _dummy?: HandleObjectType;

  evaluateHandle<Args extends SerializableOrJSHandle[], Return>(
      pageFunction: (subject: HandleObjectType, ...args: PassedArgs<Args>) => Return,
      ...args: Args
  ): Promise<JSHandle<Return>> { ... }

  evaluate<
      Args extends SerializableOrJSHandle[],
      Return extends Serializable | PromiseLike<Serializable> | void
  >(
      pageFunction: (subject: HandleObjectType, ...args: PassedArgs<Args>) => Return,
      ...args: Args
  ): Promise<Awaited<Return>> { ... }

  ...
}

class Frame {
  evaluateHandle<Args extends SerializableOrJSHandle[], Return>(
      pageFunction: (...args: PassedArgs<Args>) => Return,
      ...args: Args
  ): Promise<JSHandle<Return>> { ... }

  evaluate<
      Args extends SerializableOrJSHandle[],
      Return extends Serializable | PromiseLike<Serializable> | void
  >(
      pageFunction: (...args: PassedArgs<Args>) => Return,
      ...args: Args
  ): Promise<Awaited<Return>> { ... }

  ...
}

[...]

PR #8540, opened 2022-06-21
chore(types): Add types to in-page functions like `evaluate`(`Handle`), `waitForFunction`, etc

stevenwdv:

Improving TypeScript types

Currently: no, but I did update the existing test code

Not yet

**Summary**

Solves `evaluate`, `evaluateHandle`, and other related functions not enforcing types.
See #8485

Yes.

- Functions accepting page functions (e.g. `Page.evaluate`) do not accept strings anymore (except Page.evaluateOnNewDocument).
  → Rewrite calls to use regular functions where possible. For dynamic functions, use the [`Function()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/Function), [`AsyncFunction()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/AsyncFunction) or [`GeneratorFunction()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/GeneratorFunction) constructor.
- Generic parameters of these functions have changed.
  → Just remove them from each call. Types will now be inferred automatically.
- Types of these functions and of `JSHandle<T>` are now properly enforced, leading to rejection of improperly typed programs.
  → Add type casts or correct code where necessary.
- Some type aliases in `EvalTypes` have been removed.
  → I don't expect many people to use these, but use the new alternatives where necessary.
- `ElementHandle` is now `ElementHandle<ParentNode>` instead of `ElementHandle<Element>` by default, because it actually always supported things like `Document` as well.
  → Cast to `ElementHandle<Element>` where needed.

**Other information**

Some breaking changes can be reversed (e.g. string overloads or old `EvalTypes` type aliases). However, keeping string overloads will make typing a lot harder and will likely lead to TypeScript rejecting correct programs (at least when I tried it).

I did go all out for this one, and you may find I changed some types that shouldn't be / didn't need to be changed, then just tell me.

I intentionally made this a draft PR for now. I'm doing this in my free time because I felt like it so I'm not sure if I'll have time to update docs and such.

### DIFF
--- a/src/common/AriaQueryHandler.ts
+++ b/src/common/AriaQueryHandler.ts
@@ -20,6 +20,7 @@ import { Protocol } from 'devtools-protocol';
 import { CDPSession } from './Connection.js';
 import { DOMWorld, PageBinding, WaitForSelectorOptions } from './DOMWorld.js';
 import { assert } from './assert.js';
+import { ExposedFunctionToInPage } from './EvalTypes.js';
 
 async function queryAXTree(
   client: CDPSession,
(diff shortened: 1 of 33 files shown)


jrandolf-google:
Closed in favor of https://github.com/puppeteer/puppeteer/pull/8547.

stevenwdv:
@jrandolf Hey I don't wish to be too salty, but you could have at least told me you were working on this, now it feels like I did all of this for nothing.

Anyway, I'm glad the types are there at least. Feel free to take any code from my PR where desired...

jrandolf-google:
@stevenwdv I wouldn't cut yourself short! I've also checked out your PR and kept note of some of the ideas used there. The general layout changed drastically though through internal discussions.

stevenwdv:
@jrandolf Ah ok I'm glad to hear that :) At first it seemed to me like you didn't look at the code.

PR #8545, opened 2022-06-22
Expose #client via `_client()` on HTTPRequest

smashah:

A small change that will allow the puppeteer-page-proxy project a path to working again with the latest versions of puppeteer

`_client()` is not documented on `Page` yet still exists so I guess it's not required.

**Summary**

Completely eliminating access to the HTTPRequest._client property will break `puppeteer-page-proxy` and in turn negatively affect my users. Forcing my project to use older puppeteer version will also negatively affect my users. I am making this PR so I can eventually make a PR/fork on `puppeteer-page-proxy`

https://github.com/open-wa/wa-automate-nodejs/issues/2755
https://github.com/Cuadrix/puppeteer-page-proxy/issues/78

No

**Other information**

### DIFF
--- a/src/common/HTTPRequest.ts
+++ b/src/common/HTTPRequest.ts
@@ -196,11 +196,10 @@ export class HTTPRequest {
     return this.#url;
   }
   
-
   /**
    * @returns the respective client of the request
    */
-  client(): string {
+  _client(): string {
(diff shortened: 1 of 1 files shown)


jrandolf-google:
Fixed in #8556

PR #8547, opened 2022-06-22
feat!: type inference for evaluation types

### LINKED ISSUE #8485: [Types] Improve types for `evaluate` and `evaluateHandle` on various types (and `JSHandle<T>`)

jrandolf-google:
This PR greatly improves the types within Puppeteer:

- **Almost everything** is auto-deduced.
  - Parameters don't need to be specified in the function. They are deduced from the spread.
  - Return types don't need to be specified. They are deduced from the function. (More on this below)
  - Selections based on tag names correctly deduce element type, similar to TypeScript's mechanism for `getElementByTagName`.
- [**BREAKING CHANGE**] We've removed the ability to declare return types in type arguments for the following reasons:
  1. Setting them will indubitably break auto-deduction.
  2. You can just use `as ...` in TypeScript to coerce the correct type (given it makes sense).
- [**BREAKING CHANGE**] `waitFor` is officially gone.

To migrate to these changes, there are only four things you may need to change:
- If you set a return type using the `ReturnType` type parameter, remove it and use `as ...` and `HandleFor` (if necessary).
⛔ `evaluate<ReturnType>(a: number, b: number) => {...}, a, b)`
✅ `(await evaluate(a, b) => {...}, a, b)) as ReturnType`
⛔ `evaluateHandle<ReturnType>(a: number, b: number) => {...}, a, b)`
✅ `(await evaluateHandle(a, b) => {...}, a, b)) as HandleFor<ReturnType>`
- If you set any type parameters in the *parameters* of an evaluation function, remove them.  
⛔ `evaluate(a: number, b: number) => {...}, a, b)`
✅ `evaluate(a, b) => {...}, a, b)`
- If you set any type parameters in the method's declaration, remove them.
⛔ `evaluate<(a: number, b: number) => void>((a, b) => {...}, a, b)`
✅ `evaluate(a, b) => {...}, a, b)`

Fixes: #8485


stevenwdv:
You could also consider changing `ElementHandle<T extends Element>` to `ElementHandle<T extends ParentNode>` to officially support `Document` and `ShadowRoot` as well; see my PR #8540

jrandolf-google:
@stevenwdv There are plans to handle `Node`s! Stay tuned.

test/src/waittask.spec.ts
```diff
@@ -29,122 +28,6 @@ describe('waittask specs', function () {
   setupTestBrowserHooks();
   setupTestPageAndContextHooks();
 
-  describe('Page.waitFor', function () {
```
OrKoN, line 32: I think some of the tests don't use the deprecated waitFor? I think we should keep them (I wonder if the coverage test would complain)

jrandolf: It's okay because we tests these functions in other places.

OrKoN:
let's document the breaking part of this PR better? let's perhaps include an example of what worked before and how it should be handled now. E.g., before `handle.evaluate<MyCustomType>(() => ..)`, after `handle.evaluate(() => ..) as MyCustomType` if I understood it right?

OrKoN:
By documenting I mean let's include an example into the PR description (and also copy everything from the PR description into the commit message when landing this).

OrKoN:
nit: let's rephrase the title to be more specific, e.g., `implement type inference for evaluate types`

OrKoN (approved):

karlhorky:
@jrandolf Did the following lines introduce bugs in the types?

According to [the docs](https://pptr.dev/#?product=Puppeteer&version=v15.1.1&show=api-pageevalselector-pagefunction-args-1) a `pageFunction` should only be a function, but here a `string` is also being allowed? 

- https://github.com/puppeteer/puppeteer/pull/8547/files#diff-faca1aee95cf49cbbf62a8fbad766f4d6f943b9a830466c962b7066909a4e54aR184
- https://github.com/puppeteer/puppeteer/pull/8547/files#diff-faca1aee95cf49cbbf62a8fbad766f4d6f943b9a830466c962b7066909a4e54aR195

jrandolf-google:
> @jrandolf Did the following lines introduce bugs in the types?
> 
> According to [the docs](https://pptr.dev/#?product=Puppeteer&version=v15.1.1&show=api-pageevalselector-pagefunction-args-1) a `pageFunction` should only be a function, but here a `string` is also being allowed?
> 
> * https://github.com/puppeteer/puppeteer/pull/8547/files#diff-faca1aee95cf49cbbf62a8fbad766f4d6f943b9a830466c962b7066909a4e54aR184
> * https://github.com/puppeteer/puppeteer/pull/8547/files#diff-faca1aee95cf49cbbf62a8fbad766f4d6f943b9a830466c962b7066909a4e54aR195

Thanks for the comment. The documentation is out of date (we removed the current generator as it cannot support our new types). In any case, it was always okay to use a string actually. Much of this PR attempts to amend the problems of our Puppeteer forefathers!

karlhorky:
Ok, thanks for that. A bit confusing currently unfortunately! (mismatching types / docs). But maybe this will be remedied fairly quickly :)

PR #8550, opened 2022-06-22
chore: disable type tests

jrandolf-google:
Disabling type tests for refactoring.


OrKoN:
I'd prefer to keep the tests to guard against regressions if the types are changed by us by accident. If it interfers with the refactoring, let's disable the tests and re-enable once the refactoring is done?

jrandolf-google:
> I'd prefer to keep the tests to guard against regressions if the types are changed by us by accident. If it interfers with the refactoring, let's disable the tests and re-enable once the refactoring is done?

Done.

OrKoN (approved):

PR #8552, opened 2022-06-23
fix: extends `ElementHandle` to `Node`s

jrandolf-google:
This PR adjusts some incorrect types within Puppeteer:

- The base type parameter for `ElementHandle` is now `Node` instead of `Element`.
- The default type parameter for `ElementHandle` is still `Element`.
- `QueryHandler`s now use `Node` types.


OrKoN:
In what cases can an ElementHandle contain a Node?

jrandolf-google:
> In what cases can an ElementHandle contain a Node?

See `_createJSHandle`. Basically, ElementHandles are created from `Node`s (not `Element`s). For example, `evaluateHandle(() => document)` will return a valid `ElementHandle`. Similarly for queries involving `ShadowRoot`s.

OrKoN:
are any updates needed for api.md for these changes?

jrandolf-google:
> are any updates needed for api.md for these changes?

There are a lot of updates for `api.md`, but DocLint doesn't support the advanced typings made in previous stages. I propose we merge this and hold on release until we get some documentation (should be complete by next week).

OrKoN:
@jrandolf wdyt about doing a release now (I assume previous PRs don't need api.md changes) and postponing landing this until the documentation is ready? Today or tomorrow we also need to roll M104 which would be nice to release as soon as possible too so I would not like to block a release on anything.

test/src/elementhandle.spec.ts
```diff
@@ -572,10 +615,16 @@ describe('ElementHandle specs', function () {
...
+            return element.querySelector(`.${selector}`);
+          }
+          return null;
         },
         queryAll: (element, selector) => {
-          return element.querySelectorAll(`.${selector}`);
+          if (element instanceof Element || element instanceof Document) {
+            return element.querySelectorAll(`.${selector}`);
+          }
+          return [];
```
OrKoN: let's throw a runtime error when the element type is not right?

OrKoN:
Let's also include the before/after examples in the docs.

test-d/ElementHandle.test-d.ts
```diff
@@ -119,7 +119,7 @@ declare const handle: ElementHandle;
     await handle.$$eval(
       'some-element',
       (elements, value) => {
-        expectType<Element[]>(elements);
+        expectType<Node[]>(elements);
```
OrKoN: should it remain the Element still?

OrKoN (approved):

Qlub53:
I used to use [page.$x(expression)](https://github.com/puppeteer/puppeteer/blob/v13.4.0/docs/api.md#pagexexpression) and 
[elementHandle.click([options])](https://github.com/puppeteer/puppeteer/blob/v13.4.0/docs/api.md#elementhandleclickoptions) with puppeteer v13.4 just fine.
```ts

const handles = await page.$x('xpath');
await handles[0].click();
```

But now since API signatures have changed, using [$x(expression)](https://github.com/puppeteer/puppeteer/blob/v16.2.0/docs/api/puppeteer.page._x.md) and [click(this, options)](https://github.com/puppeteer/puppeteer/blob/v16.2.0/docs/api/puppeteer.elementhandle.click.md) from v16.2 ends up with the following `tsc` compilation error:
```
error TS2684: The 'this' context of type 'ElementHandle<Node>' is not assignable to method's 'this' of type 'ElementHandle<Element>'.

7 await handles[0].click();
        ~~~~~~~~~~
```

How shall I transit from using `Element` to using `Node`?

Qlub53:
I was able to get away with this problem by type casting.
```ts
await (handles[0] as ElementHandle<Element>).click();
```
Turns out it in fact still works at runtime and only bugs me during compile time.

It's not pretty, but it works. :/

PR #8562, opened 2022-06-23
fix: infer unioned handles

### LINKED ISSUE #8558: [Bug]: Page#evaluate fails to assign types to function parameters
### Bug description

```ts
import puppeteer from "puppeteer"

const browser = puppeteer.launch(),
	page = await browser.newPage(),
	evaluation = await page.evaluate(element => element?.textContent ?? "", await page.$("div")) // TS Error
```

### Puppeteer version

15.0.0

### Node.js version

18.4.0

### npm version

8.12.2

Windows

### Relevant log output

```shell
Property 'textContent' does not exist on type 'never'. ts(2339)
```

jrandolf-google:
Fixes: #8558


src/api-docs-entry.ts
```diff
@@ -41,43 +41,43 @@ import {CustomQueryHandler} from './common/QueryHandler.js';
...
+export * from './common/Errors.js';
 export * from './common/EventEmitter.js';
+export * from './common/ExecutionContext.js';
 export * from './common/FileChooser.js';
 export * from './common/FrameManager.js';
-export * from './common/PuppeteerViewport.js';
+export * from './common/HTTPRequest.js';
+export * from './common/HTTPResponse.js';
 export * from './common/Input.js';
+export * from './common/JSHandle.js';
```
OrKoN, line 60: should export ElementHandler too?

src/common/JSHandle.ts
```diff
@@ -134,11 +134,15 @@ export class JSHandle<T = unknown> {
...
-    Func extends EvaluateFunc<[T, ...Params]> = EvaluateFunc<[T, ...Params]>
+    Func extends EvaluateFunc<[this, ...Params]> = EvaluateFunc<
+      [this, ...Params]
+    >
   >(
     pageFunction: Func | string,
-    ...args: EvaluateParams<Params>
-  ): Promise<Awaited<ReturnType<Func>>> {
+    ...args: Params
+  ): // @ts-expect-error Circularity here is okay because we only need the return
```
OrKoN, line 143: having test types would be nice at this point? does TS generate proper d.ts if there are some errors?

OrKoN (approved):

src/api-docs-entry.ts
jrandolf: Yes, we can do this later

src/common/JSHandle.ts
jrandolf: Yes, it still generates the types correctly. I assume this would break if the type is truly circular (meaning it returns `this` somehow). Regarding testing this, the test files that changed at the tests in some sense. The typing on the files that changed broke as soon as I implemented this, which makes sense because they were wrong before.

PR #8580, opened 2022-06-25
fix: Clean up tmp profile dirs when browser is closed

pablomatiasgomez:

Bugfix: Puppeteer is not deleting tmp profiles after browser is closed.

Yes, I actually first created the test to verify that it was actually broken the way I thought.

No, this is a small bugfix, but let me know if I should update anything. 

**Summary**

Puppeteer has a feature in which if you don't specify a user data dir (`--user-data-dir`) it automatically creates one in a tmp folder, and then it's deleted when the browser is closed. 
It seems that this feature [got broken here](https://github.com/puppeteer/puppeteer/commit/b4e751f29cb6fd4c3cc41fe702de83721f0eb6dc?diff=split#diff-d09eb6dc12121ed413d2820db7fa8ba1dff9260a2ad2486a6220c861a8e5e673R134), as it is now setting `isTempUserDataDir` to false in any case. 

In my particular case, I instantiate Puppeteer multiple times, leading to an infinite number of dirs in the /tmp folder which were only deleted when I restarted the node.
```
ls -l /tmp/puppeteer_dev_chrome_profile-*  | wc -l
344
```

No. (unless for any reason somebody is expecting for tmp profiles to be kept)

### DIFF
--- a/src/node/ChromeLauncher.ts
+++ b/src/node/ChromeLauncher.ts
@@ -92,7 +92,7 @@ export class ChromeLauncher implements ProductLauncher {
       }
     }
 
-    let isTempUserDataDir = true;
+    let isTempUserDataDir = false;
 
     // Check for the user data dir argument, which will always be set even
     // with a custom directory specified via the userDataDir option.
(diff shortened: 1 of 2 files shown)


OrKoN:
Thanks for fixing this and for the test! I think it looks good and it only needs to skip this test on Firefox like the other tests.

pablomatiasgomez:
@OrKoN thanks for reviewing! I just changed the test to to be chrome only. Let me know if it looks good

jrandolf-google (approved):

PR #8598, opened 2022-06-29
fix: use `_onMessage`

jrandolf-google:


src/common/Connection.ts
```diff
@@ -66,7 +66,7 @@ export class Connection extends EventEmitter {
     this.#delay = delay;
 
     this.#transport = transport;
-    this.#transport.onmessage = this.onMessage.bind(this);
+    this.#transport.onmessage = this._onMessage.bind(this);
```
OrKoN, line 69: Should it start with underscore when it's a protected method? I think underscore is used to indicate private attributes?

jrandolf: I'd prefer no underscore, but internal uses underscore hence the reason.

OrKoN: I think we can just update the internal?

jrandolf: +1

Issue #8627, opened 2022-07-04
[Bug]: Node is either not clickable or not an HTMLElement


abozhilov:
### Bug description

Steps to reproduce the problem:

Given the following simple script: 

```js
import puppeteer from 'puppeteer';

(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();

  await page.goto('https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set');
  await Promise.all([
    page.waitForNavigation(),
    page.click('#set\\.prototype\\.keys > a')
  ]);
  await browser.close();
})();
```

It throws the following error: 
```
throw new Error('Node is either not clickable or not an HTMLElement');
```

That's due to area calculations which returns 0: https://github.com/puppeteer/puppeteer/blob/3735ef13f1ca8ed6455956f56fb8c9884fd70545/src/common/ElementHandle.ts#L1048

### Puppeteer version

15.3.0

### Node.js version

16.13.0

### npm version

^15.3.0

### What operating system are you seeing the problem on?

Linux

abozhilov:
I'm not sure why do you use an experimental API https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-getContentQuads to get the quads? 

It could be implemented same logic via `Element.getBoundingClientRect()` or `Element.getClientRects()`.

OrKoN:
Although I don't know the reason why Puppeteer uses the CDP implementation experimental methods are often created specifically for Puppeteer/DevTools (and remain experimental for years because experimental does not mean it does not work as expected or not supported, just that the API might change in the future). Puppeteer does not automatically wait for the element to become visible and that error is usually an indication that an element is not visible, and, thus, not clickable. 

Would you like to work on a PR that makes use of `Element.getBoundingClientRect()` or `Element.getClientRects()` and see if it improves the situation?

abozhilov:
Thanks for the clarifications about the experimental API.

Regarding the visibility, indeed the element is not visible in the viewport, but `page.click` scrolls the element into the view and then perform the click action. It might need couple of ms after the scrolling action to render the element properly.

Anyway, I'll submit a PR and we can have a further discussion.

jrandolf-google:
@abozhilov The reason we don't use `Element.getBoundingClientRect()` is because not every `ElementHandle` handles `Element` (counterintuitive, I know; we have a PR fixing this). For example, you can do the following:
```ts
[...]

Issue #1805, opened 2018-01-15
Puppeteer Won't Click on Element


paul-phillips-ark:
Trying to get Puppeteer to navigate through my login page which will bring the automated test to the main page of my site. It all works until it reaches 'waitFor('#idSIButton9');'. Puppeteer successfully enters the password and username but can't seem to select the submit button(#idSIButton9). Any idea what might be going wrong here? Let me know if you guys need more info. Thanks :)

```
const puppeteer = require('puppeteer');
    const { defineSupportCode } = require('cucumber')

    defineSupportCode(({ Before, Given, When, Then }) => {
        Before({ timeout: 60 * 1000 }, async function testCase() {

            this.browser = await puppeteer.launch({ headless: false });

            this.page = await this.browser.newPage();

            await this.page.setViewport({width: 1024, height: 768});

            await this.page.goto('http://localhost:3000', {waitUntil: 'networkidle2'});

            await this.page.screenshot({ path: 'test_artifacts/img/load.png', fullPage: true });

            await this.page.waitFor('button');

            await this.page.click('button');

            await this.page.waitFor('#i0116');

            await this.page.type('#i0116', 'username');

            await this.page.waitFor('#i0118');

            await this.page.type('#i0118', 'password');

            await this.page.waitFor('#idSIButton9');

            await this.page.click('#idSIButton9');

        })
```

fran0254:
Try to set a time before clicking...

```javascript
await this.page.waitFor(2000);
await this.page.click('#idSIButton9');

paul-phillips-ark:
That worked. Thanks. I thought that waitFor(selector) would have achieved the same thing, but I guess not. Thanks for the answer.

aslushnikov:
@paul-phillips-ark it looks like the javascript code responsible for handling button clicks is not evaluated yet; the `page.waitForSelector` makes sure that there's an element in DOM.

immauss:
I managed to get around this by first selecting the submit button, and then typing a "\n" 

  await page.focus('  tr:nth-child(3) > td:nth-child(2) > input[type="submit"]' )
  await page.keyboard.type('\n');

piq9117:
@immauss' solution worked out for me. thanks!

codetherus:
> await page.focus(' tr:nth-child(3) > td:nth-child(2) > input[type="submit"]' )
> await page.keyboard.type('\n');

Thanks! This solution worked for me when no others worked.

[...]

PR #8631, opened 2022-07-05
fix(page): fix page.#scrollIntoViewIfNeeded method

abozhilov:
This patch fixes page.#scrollIntoViewIfNeededso that it works with devtools protocol.
Now it blocks the main thread and waits until the scrolling action finishes.

Issues: #8627, #1805

BREAKING CHANGE: page.#scrollIntoViewIfNeeded throws erros which come from the internal implementation.
 - `Protocol error (DOM.scrollIntoViewIfNeeded): Node is detached from document`
 - `Protocol error (DOM.scrollIntoViewIfNeeded): Node does not have a layout object`
Also now it does work with TextNode properly and does not throw an error.

### DIFF
--- a/src/common/ElementHandle.ts
+++ b/src/common/ElementHandle.ts
@@ -240,49 +240,51 @@ export class ElementHandle<
 
   async #scrollIntoViewIfNeeded(this: ElementHandle<Element>): Promise<void> {
     const error = await this.evaluate(
-      async (element, pageJavascriptEnabled): Promise<string | false> => {
+      async (element): Promise<string | undefined> => {
         if (!element.isConnected) {
           return 'Node is detached from document';
         }
         if (element.nodeType !== Node.ELEMENT_NODE) {
           return 'Node is not of type HTMLElement';
         }
-        // force-scroll if page's javascript is disabled.
-        if (!pageJavascriptEnabled) {
-          element.scrollIntoView({
-            block: 'center',
-            inline: 'center',
-            // @ts-expect-error Chrome still supports behavior: instant but
-            // it's not in the spec so TS shouts We don't want to make this
-            // breaking change in Puppeteer yet so we'll ignore the line.
-            behavior: 'instant',
-          });
-          return false;
-        }
-        const visibleRatio = await new Promise(resolve => {
-          const observer = new IntersectionObserver(entries => {
-            resolve(entries[0]!.intersectionRatio);
-            observer.disconnect();
-          });
-          observer.observe(element);
-        });
-        if (visibleRatio !== 1.0) {
-          element.scrollIntoView({
-            block: 'center',
-            inline: 'center',
-            // @ts-expect-error Chrome still supports behavior: instant but
-            // it's not in the spec so TS shouts We don't want to make this
-            // breaking change in Puppeteer yet so we'll ignore the line.
-            behavior: 'instant',
-          });
-        }
-        return false;
-      },
-      this.#page.isJavaScriptEnabled()
+        return;
+      }
     );
 
     if (error) {
       throw new Error(error);
+    }
+
+    try {
+      await this._client.send('DOM.scrollIntoViewIfNeeded', {
+        objectId: this._remoteObject.objectId,
+      });
+    } catch (_err) {
+      // Fallback to Element.scrollIntoView if DOM.scrollIntoViewIfNeeded is not supported
+      await this.evaluate(
+        async (element, pageJavascriptEnabled): Promise<void> => {
+          const visibleRatio = async () => {
+            return await new Promise(resolve => {
+              const observer = new IntersectionObserver(entries => {
+                resolve(entries[0]!.intersectionRatio);
+                observer.disconnect();
+              });
+              observer.observe(element);
+            });
+          };
+          if (!pageJavascriptEnabled || (await visibleRatio()) !== 1.0) {
+            element.scrollIntoView({
+              block: 'center',
+              inline: 'center',
+              // @ts-expect-error Chrome still supports behavior: instant but
+              // it's not in the spec so TS shouts We don't want to make this
+              // breaking change in Puppeteer yet so we'll ignore the line.
+              behavior: 'instant',
+            });
+          }
+        },
+        this.#page.isJavaScriptEnabled()
+      );
     }
   }
 
--- a/test/src/click.spec.ts
+++ b/test/src/click.spec.ts
@@ -164,7 +164,10 @@ describe('Page.click', function () {
     await page.goto(server.PREFIX + '/offscreenbuttons.html');
     const messages: any[] = [];
     page.on('console', msg => {
-      return messages.push(msg.text());
+      if (msg.type() === 'log') {
+        return messages.push(msg.text());
+      }
+      return;
     });
     for (let i = 0; i < 11; ++i) {
       // We might've scrolled to click a button - reset to (0, 0).
--- a/test/src/elementhandle.spec.ts
+++ b/test/src/elementhandle.spec.ts
@@ -203,7 +203,7 @@ describe('ElementHandle specs', function () {
         })
       ).toBe(true);
     });
-    it('should work for TextNodes', async () => {
+    it('should not work for TextNodes', async () => {
       const {page, server} = getTestState();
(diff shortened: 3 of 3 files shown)


OrKoN:
it looks like Firefox implementation does not support `DOM.scrollIntoViewIfNeeded` :-/ let's fallback to the script evaluation if `DOM.scrollIntoViewIfNeeded` fails?

abozhilov:
@OrKoN Not sure it's the best approach as it'll introduce differences between chrome/firefox. `DOM.scrollIntoViewIfNeeded` works with text nodes, while the former implementation doesn't. 

See this test case: https://github.com/puppeteer/puppeteer/blob/main/test/src/elementhandle.spec.ts#L206  
Even that the name says it should work, currently it doesn't due to `#scrollIntoViewIfNeeded` implementation. I've fixed this test case with my PR, but obviously it doesn't pass in Firefox. 

I think we should log a feature request in Firefox bug tracker for  `DOM.scrollIntoViewIfNeeded`  and  my PR could wait, unless you have a better solution of the problem.

OrKoN:
I don't think Firefox is going to look into extending the CDP support anytime soon unfortunately. cc @whimboo Henrik, is it something that could be a priority for you?

whimboo:
I doubt that we could take that unless someone is interested to actually help with the implementation. Also [`DOM.scrollIntoViewIfNeeded`](https://chromedevtools.github.io/devtools-protocol/tot/DOM/#method-scrollIntoViewIfNeeded) is marked as `experimental` which makes it lesser a priority. Sorry.

abozhilov:
@OrKoN I'll investigate it a bit further and I think I'll come up with a solution which would work in both envs.

abozhilov:
I ended up with this implementation which keeps backward compatibility as much as possible and it falls back to `Element.scrollIntoView` if  `DOM.scrollIntoViewIfNeeded` is not supported. Anyway with the latest release `15.3.1` a lot of tests fail and it's hard for contributing.

```ts
  async #scrollIntoViewIfNeeded(): Promise<void> {
    const error = await this.evaluate(
      async (element): Promise<string | undefined> => {
        if (!element.isConnected) {
          return 'Node is detached from document';
        }
        if (element.nodeType !== Node.ELEMENT_NODE) {
        }
        return;
      }
    );

    if (error) {
      throw new Error(error);
    }

    try {
      await this._client.send('DOM.scrollIntoViewIfNeeded', {
        objectId: this._remoteObject.objectId,
      });
    } catch (_err) {
      // Fallback to Element.scrollIntoView if DOM.scrollIntoViewIfNeeded is not supported
      await this.evaluate(
        async (element, pageJavascriptEnabled): Promise<void> => {
          // force-scroll if page's javascript is disabled.
          if (!pageJavascriptEnabled) {
            element.scrollIntoView({
              block: 'center',
              inline: 'center',
              // @ts-expect-error Chrome still supports behavior: instant but
              // it's not in the spec so TS shouts We don't want to make this
              // breaking change in Puppeteer yet so we'll ignore the line.
              behavior: 'instant',
            });
          }
          const visibleRatio = await new Promise(resolve => {
            const observer = new IntersectionObserver(entries => {
              resolve(entries[0]!.intersectionRatio);
              observer.disconnect();
            });
            observer.observe(element);
          });
          if (visibleRatio !== 1.0) {
            element.scrollIntoView({
              block: 'center',
              inline: 'center',
              // @ts-expect-error Chrome still supports behavior: instant but
              // it's not in the spec so TS shouts We don't want to make this
              // breaking change in Puppeteer yet so we'll ignore the line.
              behavior: 'instant',
            });
          }
        },
        this.#page.isJavaScriptEnabled()
      );
    }
  }
```

OrKoN:
@abozhilov I think 15.3.1 only contains type and documentation changes. The amount of flaky failures (which is unfortunately not zero) should be the same (but we removed automatic retries and we aim at fixing those). Could you update your PR with the latest code version?

abozhilov:
I've just updated my PR.

src/common/ElementHandle.ts
```diff
@@ -240,50 +240,61 @@ export class ElementHandle<
...
+              block: 'center',
+              inline: 'center',
+              // @ts-expect-error Chrome still supports behavior: instant but
+              // it's not in the spec so TS shouts We don't want to make this
+              // breaking change in Puppeteer yet so we'll ignore the line.
+              behavior: 'instant',
+            });
+          }
+          const visibleRatio = await new Promise(resolve => {
+            const observer = new IntersectionObserver(entries => {
```
OrKoN: can we use IntersectionObserver to detect when the scroll is done?

abozhilov: I've tried and the race condition is still there. In order to prevent it you need once the node intersects the viewport you'd need to to something like that: 
`setTimeout(() => resolve(true), 100)` which is not great I think.

src/common/ElementHandle.ts
```diff
@@ -240,50 +240,61 @@ export class ElementHandle<
...
+        return;
+      }
     );
 
     if (error) {
       throw new Error(error);
     }
+
+    try {
+      await this._client.send('DOM.scrollIntoViewIfNeeded', {
```
OrKoN, line 259: Can we be sure that the scroll is over once this resolves? In the previous version you had a test with an element that is outside of the viewport, should we bring the test back?

test/src/click.spec.ts
```diff
@@ -164,7 +164,10 @@ describe('Page.click', function () {
     await page.goto(server.PREFIX + '/offscreenbuttons.html');
     const messages: any[] = [];
     page.on('console', msg => {
-      return messages.push(msg.text());
+      if (msg.type() === 'log') {
```
OrKoN, line 167: why was this change needed?

src/common/ElementHandle.ts
```diff
@@ -240,50 +240,61 @@ export class ElementHandle<
...
-            // @ts-expect-error Chrome still supports behavior: instant but
-            // it's not in the spec so TS shouts We don't want to make this
-            // breaking change in Puppeteer yet so we'll ignore the line.
-            behavior: 'instant',
-          });
-        }
-        return false;
-      },
-      this.#page.isJavaScriptEnabled()
+        return;
```
OrKoN, line 250: nit: return not needed?

test/src/click.spec.ts
```diff
@@ -164,7 +164,10 @@ describe('Page.click', function () {
...
+        return messages.push(msg.text());
+      }
+      return;
```
OrKoN, line 170: nit: return not needed?

src/common/ElementHandle.ts
abozhilov: I decided that it's redundant to add an additional test, since there is a test called `'should click offscreen buttons'`. I've improved it a bit in this PR.

test/src/click.spec.ts
abozhilov: Yes, that's because in Firefox when I call `await this._client.send('DOM.scrollIntoViewIfNeeded')` the error appears in the console and the test case should observe only the log messages.

src/common/ElementHandle.ts
abozhilov: If I omit it there is a TS error, because not all paths in the code return value.

test/src/click.spec.ts
abozhilov: Same as above.

OrKoN:
Do you think we still need to mark this PR as a breaking change? It looks backward compatible.

OrKoN (approved):

OrKoN:
Also, it looks like a few elementhandle.spec.ts tests broke due to the changes to the test file

abozhilov:
@OrKoN It's backward compatible, but to be 100% sure I'll have a look at the test cases which you've mentioned.

abozhilov:
@OrKoN I've just reverted `offscreenbuttons.html` to it's previous version, and now the failing tests would work.

paliwal81:
Hi @OrKoN, I would like to know by when we can expect this pull request to be merged. 
Currently my e2e test cases are failing because of the same issue (as UI form is long and elements are visible on view port).

PR #8653, opened 2022-07-07
fix: updated navigation event handling

jrandolf-google:
This PR separates navigation handling and lifecycle handling in Puppeteer:
- The new `NavigationWatcher` strictly handles navigation events and has been simplified drastically to only watch for the navigation response which appears after the request (hence the navigation) has finished.
- There is also some refactoring for tidiness.


src/common/Page.ts
```diff
@@ -1871,7 +1871,9 @@ export class Page extends EventEmitter {
   async reload(options?: WaitForOptions): Promise<HTTPResponse | null> {
     const result = await Promise.all([
       this.waitForNavigation(options),
-      this.#client.send('Page.reload'),
+      this.evaluate(() => {
+        window.location.reload();
```
OrKoN: window.location might be overloaded by the page? Why not keep using Page.reload?

OrKoN: another thing discovered: `window.location.reload` would not work for data URLs.

src/common/DOMWorld.ts
```diff
@@ -284,17 +284,10 @@ export class DOMWorld {
...
-      waitUntil,
-      timeout
+      timeout,
+      waitUntil
     );
-    const error = await Promise.race([
-      watcher.timeoutOrTerminationPromise(),
-      watcher.lifecyclePromise(),
-    ]);
-    watcher.dispose();
```
OrKoN, line 294: do we still need dispose?

jrandolf: No, all watchers dispose themselves on error or return.

src/common/FrameManager.ts
```diff
@@ -206,56 +207,16 @@ export class FrameManager extends EventEmitter {
...
-          ? new Error(`${response.errorText} at ${url}`)
-          : null;
-      } catch (error) {
-        if (isErrorLike(error)) {
-          return error;
-        }
-        throw error;
-      }
-    }
+    await this.#navigate(url, referer, frame._id);
```
OrKoN, line 215: I think we should call waitForFrameNavigation before we call navigate?

test/src/navigation.spec.ts
```diff
@@ -116,11 +116,19 @@ describe('navigation', function () {
         window.addEventListener(
           'beforeunload',
           () => {
-            return history.replaceState(null, 'initial', window.location.href);
+            history.replaceState(null, 'initial', window.location.href);
           },
           false
         );
       });
+      // TODO: There is a determinism problem here. Because the history API is
```
OrKoN, line 124: is this comment still relevant?

OrKoN:

PR #8656, opened 2022-07-09
fix(typo): fix typo in Page.waitForSelector() method

ipko1996:
fix typo from ait to wait in Remarks - hidden option


OrKoN:
Thanks, will be fixed in https://github.com/puppeteer/puppeteer/pull/8701

PR #8688, opened 2022-07-22
fix: address flakiness in frame handling

OrKoN:
When we attach to a frame, we send a call to get
the page frame tree from CDP. Based on the tree data
we look up the parent frame if parentId is provided.
The problem is that the call to get the page frame
tree could take arbitrary time and the calls for the
parent and child frames might happen at the same time.
So the situation where the frame tree for the child frame
is resolved before the parent frame is known is fairly
common.

This PR addresses the issue by awaiting for the parent
frame id before attempting to register a child frame.

### DIFF
--- a/src/common/FrameManager.ts
+++ b/src/common/FrameManager.ts
@@ -29,7 +29,12 @@ import {Page} from './Page.js';
 import {Target} from './Target.js';
 import {TimeoutSettings} from './TimeoutSettings.js';
 import {EvaluateFunc, HandleFor, NodeFor} from './types.js';
-import {debugError, isErrorLike} from './util.js';
+import {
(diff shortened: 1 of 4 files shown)


mathiasbynens (approved):
Is there a good non-flaky way to test any of this?

OrKoN:
@mathiasbynens I have been thinking about record/replaying CDP messages for test that without an actual Chromium binary but I think it's too much effort to maintain such tests.

mathiasbynens:
> @mathiasbynens I have been thinking about record/replaying CDP messages for test that without an actual Chromium binary but I think it's too much effort to maintain such tests.

I can imagine the rebaselining getting a bit out of hand, indeed… Oh well. Feel free to land this as-is!

PR #8697, opened 2022-07-25
Clear up waitForFunction docs a bit

dobesv:
I found the docs for this function a bit unclear, so I thought I'd submit a little bit of improvement to help others and my future self understand how this function works.

### DIFF
--- a/src/common/Page.ts
+++ b/src/common/Page.ts
@@ -3370,7 +3370,8 @@ export class Page extends EventEmitter {
    * Calls a function repeatedly until in the page's context until it returns a truthy value.  Promises are
    * awaited and then checked for a truthy value.  When a truthy value is returned, its value is propagated
    * out of {@link Page.waitForFunction} as a handle to the value in the page's context which can be used to
-   * interact with the value further.  If the timeout is reached first, an error is tyrown.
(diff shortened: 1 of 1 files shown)


src/common/Page.ts
```diff
@@ -3367,7 +3367,10 @@ export class Page extends EventEmitter {
   }
 
   /**
-   * Waits for a function to finish evaluating in the page's context.
+   * Calls a function repeatedly until in the page's context until it returns a truthy value.  Promises are
+   * awaited and then checked for a truthy value.  When a truthy value is returned, its value is propagated
+   * out of {@link Page.waitForFunction} as a handle to the value in the page's context which can be used to
+   * interact with the value further.  If the timeout is reached first, an error is tyrown.
```
OrKoN: tyrown -> thrown

OrKoN:
I think you would need to run `npm run docs` to re-generate markdown documentation

src/common/Page.ts
dobesv:
```suggestion
   * interact with the value further.  If the timeout is reached first, an error
   * is thrown.
```

dobesv:
Hmm this is a lot more work than I expected, maybe I'll come back to it another day.

PR #8730, opened 2022-08-02
fix: use an `xpath` query handler

jrandolf-google:
This PR replaces XPath handling in Puppeteer with the newer query handling API.


src/common/QueryHandler.ts
```diff
@@ -191,11 +210,46 @@ const pierceHandler = createInternalQueryHandler({
...
-  ['aria', ariaHandler],
-  ['pierce', pierceHandler],
+const xpathHandler = internalizeCustomQueryHandler({
+  queryOne: (element, selector) => {
+    const doc = element.ownerDocument || document;
+    const iterator = doc.evaluate(
+      selector,
+      element,
+      null,
+      XPathResult.ORDERED_NODE_ITERATOR_TYPE
```
OrKoN: should we use FIRST_ORDERED_NODE_TYPE here? it seems like ORDERED_NODE_ITERATOR_TYPE would be fetching all matching elements? (not sure exactly what that means in the xpath impl)

OrKoN (commented):
Let's add a fews test for using xpath selectors with the regular selector APIs, e.g., `$`, `$$` and `waitForSelector`?

src/common/ElementHandle.ts
```diff
@@ -140,6 +140,8 @@ export class ElementHandle<
   }
 
   /**
+   * @deprecated Use {@link waitForSelector} with an xpath selector.
```
OrKoN: Let's update waitForSelector docs to mention that one can send xpath selectors directly.

src/common/QueryHandler.ts
```diff
@@ -191,11 +210,46 @@ const pierceHandler = createInternalQueryHandler({
...
+
+interface InternalQueryHandlerInfo {
+  test: RegExp;
+  handler: InternalQueryHandler;
+}
+
+const INTERNAL_QUERY_HANDLERS = new Map<string, InternalQueryHandlerInfo>([
+  ['aria', {test: /$aria\//g, handler: ariaHandler}],
+  ['pierce', {test: /$pierce\//g, handler: pierceHandler}],
+  ['xpath', {test: /$\.?\/\//, handler: xpathHandler}],
```
OrKoN: will it accept `/`?  I think it's a valid XPATH

OrKoN: I think we should also support `(//h1)[1]/text()` and  perhaps `count(//*)`?

OrKoN: so it does not look like we can test it by a regex?

jrandolf: `(//h1)[1]/text()` is the same as `//h1[1]/text()`.

I don't think we should support XPaths that give non-Node return values. The reason being we cannot detect (without going down a parsing rabbit hole) whether the return value is not a Node which is required for the correct `XPathResult` enum.

OrKoN: So why should not we support `(//h1)[1]/text()` it is a valid selector, right? I suggest that we prefix the xpath selectors with `xpath/` that allows defining any XPath selectors the user might want but automically rewrite `.//`, `//`, and `/` to have the prefix `xpath/`. WDYT?

jrandolf: sg

src/common/ElementHandle.ts
jrandolf: Added https://github.com/puppeteer/puppeteer/issues/8740. The current `waitForSelector` is insufficient anyway. Let's deal with it in another PR.

src/common/QueryHandler.ts
```diff
@@ -191,11 +210,60 @@ const pierceHandler = createInternalQueryHandler({
...
+    let item;
+    while ((item = iterator.iterateNext())) {
+      array.push(item);
+    }
+    return array;
+  },
+});
+
+const transformXPathSelector = (selector: string): string => {
+  if (selector.startsWith('//')) {
```
OrKoN: I think we only need this transformation if the waitForSelector is invoked on the element handle

src/common/QueryHandler.ts
```diff
@@ -191,11 +210,60 @@ const pierceHandler = createInternalQueryHandler({
...
+const INTERNAL_QUERY_HANDLERS = new Map<
+  string,
+  [
+    handler: InternalQueryHandler,
+    transformSelector?: (selector: string) => string
+  ]
+>([
+  ['aria', [ariaHandler]],
+  ['pierce', [pierceHandler]],
+  ['xpath', [xpathHandler, transformXPathSelector]],
```
OrKoN: looks complicated? would it be better to move the transforms inside the xpath handler?

jrandolf: The transforms update the selectors before handling. This is better since we aren't in page scope.

src/common/QueryHandler.ts
jrandolf: This is just a noop at Page level.

OrKoN: why does the test reports .// in the error message then? https://github.com/puppeteer/puppeteer/pull/8730/files#diff-20e9a9715adedba9766c324afd9350bd8187dd7e166cbbd43888a561623d6f83L738

I know that the selectors are equivalent but I don't think we should be rewriting them without a need.

src/common/QueryHandler.ts
OrKoN: hm, what about making a transformSelector to be part of the InternalQueryHandler API and providing an implementation for the xpathHandler?

src/common/QueryHandler.ts
jrandolf: Because the element doesn't exist. I don't think it makes sense to handle page handle logic separately, given this was also previous behavior.

OrKoN: Hm, I don't think it was the previous behavior. Previously, the selector would be only rewritten when called on an ElementHandle.

OrKoN (approved):

PR #8742, opened 2022-08-04
fix: specify the target filter in setDiscoverTargets

OrKoN:
To stay compatible with the next version of Chromium.
See https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-setDiscoverTargets

### DIFF
--- a/src/common/ChromeTargetManager.ts
+++ b/src/common/ChromeTargetManager.ts
@@ -101,7 +101,12 @@ export class ChromeTargetManager extends EventEmitter implements TargetManager {
     this.#connection.on('sessiondetached', this.#onSessionDetached);
     this.#setupAttachmentListeners(this.#connection);
 
-    this.#connection.send('Target.setDiscoverTargets', {discover: true});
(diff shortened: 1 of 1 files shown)


src/common/ChromeTargetManager.ts
```diff
@@ -101,7 +101,12 @@ export class ChromeTargetManager extends EventEmitter implements TargetManager {
     this.#connection.on('sessiondetached', this.#onSessionDetached);
     this.#setupAttachmentListeners(this.#connection);
 
-    this.#connection.send('Target.setDiscoverTargets', {discover: true});
+    // TODO: remove `as any` once the protocol definitions are updated with the
+    // next Chromium roll.
+    this.#connection.send('Target.setDiscoverTargets', {
+      discover: true,
+      filter: [{type: 'tab', exclude: true}, {}],
+    } as any);
```
OrKoN, line 109: happy to type it out as an explicit type but I guess any is okay as it's only till the next roll

jrandolf-google (approved):

PR #8754, opened 2022-08-08
Update puppeteer.page.waitfornetworkidle.md

bert-w:
**Summary**
Documentation update for `page.waitForNetworkIdle()`. The original documentation did not mention anything about a failing promise, and if/how that would ever happen.

**Other information**
There appear to be many more spots where proper documentation is missing. For instance the default values for this function's `idleTime` + `timeout` options are not documented.

- I am assuming that the default `timeout` is set using `page.setDefaultTimeout()` (I don't know which property this sets; it would be nice if you could call `page.defaultTimeout` to get the default timeout setting.) Nevertheless it should be nice if it was referenced in each function.
- I guess `idleTime` defaults to `0`? Which would mean that the instant that the network is idle the promise resolves?

### DIFF
--- a/docs/api/puppeteer.page.waitfornetworkidle.md
+++ b/docs/api/puppeteer.page.waitfornetworkidle.md
@@ -17,12 +17,14 @@ class Page {
 
 ## Parameters
 
-| Parameter | Type                                     | Description                                   |
-| --------- | ---------------------------------------- | --------------------------------------------- |
-| options   | { idleTime?: number; timeout?: number; } | <i>(Optional)</i> Optional waiting parameters |
+| Parameter        | Type                                     | Description                                                        |
+| ---------------- | ---------------------------------------- | ------------------------------------------------------------------ |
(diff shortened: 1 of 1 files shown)


docs/api/puppeteer.page.waitfornetworkidle.md
```diff
@@ -17,12 +17,14 @@ class Page {
 
 ## Parameters
 
-| Parameter | Type                                     | Description                                   |
-| --------- | ---------------------------------------- | --------------------------------------------- |
-| options   | { idleTime?: number; timeout?: number; } | <i>(Optional)</i> Optional waiting parameters |
+| Parameter        | Type                                     | Description                                                        |
```
OrKoN, line 20: thanks for the contribution but the documentation is now generated from the source code: TypeScript types and the comments on the corresponding methods/types. It needs to be updated there first and this file is automatically generated.

bert-w:
Ah ofcourse. I've made a new PR.

PR #8761, opened 2022-08-09
chore: rename `DOMWorld` to `IsolatedWorld`

jrandolf-google:


src/common/AriaQueryHandler.ts
```diff
@@ -136,7 +140,7 @@ const queryAll = async (
   const exeCtx = element.executionContext();
   const {name, role} = parseAriaSelector(selector);
   const res = await queryAXTree(exeCtx._client, element, name, role);
-  const world = exeCtx._world as DOMWorld;
+  const world = exeCtx._world as IsolatedWorld;
```
OrKoN: do we need the cast here?

jrandolf: Yes, because could be undefined.

OrKoN (approved):

PR #8770, opened 2022-08-10
chore: refactor FrameManager and fix docs

jrandolf-google:
This PR fixes some comments and refactors a few functions in FrameManager. In particular, navigation functions on FrameManager have been migrated to the Frame class


src/common/Page.ts
```diff
@@ -97,12 +99,12 @@ export interface Metrics {
...
   /**
-   * Maximum wait time in milliseconds, defaults to 30 seconds, pass `0` to
-   * disable the timeout.
+   * Maximum wait time in milliseconds. Pass 0 to disable the timeout.
    *
-   * @remarks
    * The default value can be changed by using the
    * {@link Page.setDefaultTimeout} method.
+   *
+   * @defaultValue `3000`
```
OrKoN: 30sec = `30000`?

src/common/Page.ts
```diff
@@ -112,13 +114,13 @@ export interface WaitTimeoutOptions {
...
-   * Maximum wait time in milliseconds, defaults to 30 seconds, pass `0` to
-   * disable the timeout.
+   * Maximum wait time in milliseconds. Pass 0 to disable the timeout.
    *
-   * @remarks
    * The default value can be changed by using the
    * {@link Page.setDefaultTimeout} or {@link Page.setDefaultNavigationTimeout}
    * methods.
+   *
+   * @defaultValue `3000`
```
OrKoN: 30sec => `30000`?

docs/api/puppeteer.page.setextrahttpheaders.md
```diff
@@ -4,7 +4,19 @@ sidebar_label: Page.setExtraHTTPHeaders
 
 # Page.setExtraHTTPHeaders() method
 
-The extra HTTP headers will be sent with every request the page initiates. NOTE: All HTTP header names are lowercased. (HTTP headers are case-insensitive, so this shouldn’t impact your server code.) NOTE: page.setExtraHTTPHeaders does not guarantee the order of headers in the outgoing requests.
+The extra HTTP headers will be sent with every request the page initiates.
+
+:::tip
```
OrKoN, line 9: is there a syntax that is both rendered by GitHub and by the website generator? 🤔

src/common/FrameManager.ts
```diff
@@ -1144,57 +1129,49 @@ export class Frame {
...
    * @returns a promise that resolves to the added tag when the stylesheets's
    * `onload` event fires or when the CSS content was injected into the
    * frame.
    */
   async addStyleTag(
     options: FrameAddStyleTagOptions
-  ): Promise<ElementHandle<Node>> {
-    return this._mainWorld.addStyleTag(options);
+  ): Promise<ElementHandle<HTMLStyleElement | HTMLLinkElement>> {
+    return this.worlds[DEFAULT_WORLD].addStyleTag(options);
```
OrKoN: not sure if DEFAULT_WORLD is better than MAIN_WORLD? WDYT about calling it USER_WORLD or PAGE_WORLD to indicate that it is the where the user content runs? Otherwise, DEFAULT_WORLD sg

src/common/FrameManager.ts
```diff
@@ -1204,76 +1181,60 @@ export class Frame {
...
 
   /**
-   * This method fetches an element with `selector`, scrolls it into view if
-   * needed, and then uses {@link Page.touchscreen} to tap in the center of the
-   * element.
-   *
-   * @remarks
-   *
-   * If there's no element matching `selector`, the method throws an error.
+   * Taps the first element that matches the `selector`.
```
OrKoN, line 1346: should we me mention automatic scroll into view? (if it really happens)

src/common/IsolatedWorld.ts
```diff
@@ -69,6 +69,23 @@ export interface PageBinding {
   pptrFunction: Function;
 }
 
+/**
+ * @internal
+ */
+export const DEFAULT_WORLD = Symbol('defaultWorld');
```
OrKoN: should we add some docs for the future us to describe the intention behind the two worlds?

src/common/FrameManager.ts
jrandolf: Let's not remark it. It's expected that this will scroll into view (in fact, all of these types of functions do). It wouldn't make sense otherwise because someone could use a function that scrolls into view, and then use a function that doesn't which would be breaking.

docs/api/puppeteer.page.setextrahttpheaders.md
jrandolf: We can use remark.

src/common/FrameManager.ts
OrKoN: But how would the users know if it scroll into the view by default or not? is there another place where it's documented?

jrandolf: As I said, it's expected. Otherwise, we would have to write it for every function (which virtually implies we should write it for no function).

OrKoN (approved):

src/common/FrameManager.ts
OrKoN: Imagine a new user opens Puppeteer docs and tries to see if they need to scrollIntoView before doing an action? If all the functions auto-scroll, let's mention it on the class level? For example, the `focus` method does not auto-scroll.

jrandolf: Let's mention it on the class. And also fix focus.

OrKoN: that would be a breaking change though. I think it all ends up on ElementHandle and I wonder if we can just refer to the documentation there? There is a bunch of other methods that don't autoscroll: I think type and press for example.

jrandolf: Yes, let's just reference it using `{@see ...}`.

jrandolf: Seeing as the `ElementHandle` documentation isn't that great. I'll do this in another PR.

OrKoN:
Why did the documentation check fail?

Issue #8740, opened 2022-08-03
[Feature] Add better query handling documentation


jrandolf-google:

OrKoN:
@jrandolf is it done?

PR #8773, opened 2022-08-10
chore: refactor JSHandle and ExecutionContext

jrandolf-google:
Reference: #8740


src/common/JSHandle.ts
```diff
@@ -212,64 +182,57 @@ export class JSHandle<T = unknown> {
...
-        objectId: this.#remoteObject.objectId,
-        returnByValue: true,
-        awaitPromise: true,
-      });
-      return valueFromRemoteObject(response.result) as T;
+  async jsonValue(): Promise<T> {
+    if (!this.#remoteObject.objectId) {
+      return valueFromRemoteObject(this.#remoteObject);
+    }
+    const value = await this.evaluate(object => {
```
OrKoN, line 223: Is it equivalent? I was under the impression that evaluate might not return by value. That means you might get a reference to an object instead of JSON?

src/common/JSHandle.ts
```diff
@@ -212,64 +182,57 @@ export class JSHandle<T = unknown> {
...
-    for (const property of response.result) {
-      if (!property.enumerable || !property.value) {
-        continue;
-      }
-      result.set(property.name, createJSHandle(this.#context, property.value));
-    }
-    return result;
+    const entries = await Promise.all(
+      keys.map(async key => {
+        const handle = await this.getProperty(key);
```
OrKoN: Let's keep the existing logic here too. The previous version created all JSHandles with a single CDP call, this one does multiple leading to a potentially inconsistent data and performance impact,

src/common/JSHandle.ts
jrandolf: This is equivalent. Check tests.

OrKoN (approved):

PR #8795, opened 2022-08-16
fix: deprecate `root` in `WaitForSelectorOptions`

jrandolf-google:


OrKoN (approved):

OrKoN:
It looks like docs have to re-generated

PR #8801, opened 2022-08-17
chore: infrastructure for injecting scripts into DOMWorlds

jrandolf-google:
This PR removes the `generated` folder in favor of on-demand file generation. This includes improvements to the build system using a GNU Makefile-like scripts (`job.ts`) which drastically improves compilation speed.

This PR also adds the `injected` folder which contains scripts that will be executed in the puppeteer context. Within the `injected` folder, some `Poller`s which will be used in the upcoming refactoring of the wait tasks.


versions.js
```diff
@@ -16,7 +16,7 @@
 
 const versionsPerRelease = new Map([
   // This is a mapping from Chromium version => Puppeteer version.
-  // In Chromium roll patches, use 'NEXT' for the Puppeteer version.
+  // In Chromium roll patches, use 'v16.1.0' for the Puppeteer version.
```
OrKoN: hm, this is wrong, it should only replace NEXT in the array map

src/common/Tracing.ts
```diff
@@ -13,15 +13,12 @@
...
-  getReadableFromProtocolStream,
-  isErrorLike,
-} from './util.js';
+import {assert} from '../util/assert.js';
+import {getReadableAsBuffer, getReadableFromProtocolStream} from './util.js';
+import {isErrorLike} from '../util/ErrorLike.js';
 import {CDPSession} from './Connection.js';
 
-/**
+/** util/ErrorLike.js
```
OrKoN: ?

.github/workflows/pre-release.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         run: npm install
       - name: Build
         run: |
-          node utils/generate_version_file.js
+          ts-node utils/generate_sources.ts
```
OrKoN, line 23: would it be better to run `npm run build` here?

OrKoN: or at least `npm run generate:sources`

jrandolf: We just need the `version.js` file that is generated

src/common/IsolatedWorld.ts
```diff
@@ -722,9 +724,7 @@ export class IsolatedWorld {
       waitForVisible: boolean,
       waitForHidden: boolean
     ): Promise<Node | null | boolean> {
-      const node = predicateQueryHandler
-        ? ((await predicateQueryHandler(root, selector)) as Element)
-        : root.querySelector(selector);
+      const node = (await predicateQueryHandler(root, selector)) as Element;
```
OrKoN, line 727: I think predicateQueryHandler might undefined? If we are sure it's always there, let's update times in makePredicateString to not allow empty predicate query handlers

utils/generate_docs.ts
```diff
@@ -20,8 +20,6 @@ import {chdir} from 'process';
 import semver from 'semver';
 import {versionsPerRelease} from '../versions.js';
 import versionsArchived from '../website/versionsArchived.json';
-
-// eslint-disable-next-line import/extensions
 import {generateDocs} from './internal/custom_markdown_action';
```
OrKoN, line 23: why don't we use a .js extension?

OrKoN: because of ts-node?

jrandolf: Yes.

OrKoN (approved):

PR #8831, opened 2022-08-23
chore: addScriptToEvaluateOnNewDocument on page

jrandolf-google:


src/common/Page.ts
```diff
@@ -641,6 +646,10 @@ export class Page extends EventEmitter {
       this.#frameManager.initialize(this.#target._targetId),
       this.#client.send('Performance.enable'),
       this.#client.send('Log.enable'),
+      this.#client.send('Page.addScriptToEvaluateOnNewDocument', {
```
OrKoN, line 649: I think we need to execute it for OOPIF targets as well (so not only pages) Where would that happen?

jrandolf: Ah, okay. I was trying to understand why another PR was failing.

PR #8844, opened 2022-08-25
fix!: internalize execution context

jrandolf-google:


OrKoN (approved):

swissspidy:
Is there more context regarding this change? Seems like this breaks packages like [`pptr-testing-library`](https://www.npmjs.com/package/pptr-testing-library) and [`expect-puppeteer`](https://www.npmjs.com/package/expect-puppeteer)

OrKoN:
@swissspidy ExecutionContext API is a bit too broad and we want to make it internal to allow some future features. It looks like for these libraries getting the frame from the ElementHandle would work as well. Let me see if I can open PRs.

OrKoN:
With https://github.com/puppeteer/puppeteer/pull/8875 the frame getter will become the public API https://github.com/puppeteer/puppeteer/pull/8875 (it's already available at runtime but not in public types). The libraries should use the Frame directly instead of the execution context.

PR #8848, opened 2022-08-25
fix!: remove `root` from `WaitForSelectorOptions`

jrandolf-google:
This PR also moves some of the `waitForSelector` logic into `QueryHandler` for an upcoming PR related to removing DOM methods from `IsolatedWorld`s


docs/api/puppeteer.page.waitforselector.md
```diff
@@ -34,17 +34,17 @@ const puppeteer = require('puppeteer');
...
   ): Promise<ElementHandle<NodeFor<Selector>> | null>;
 }
 ```
 
 ## Parameters
 
-| Parameter | Type                                                                                   | Description                                                                                            |
-| --------- | -------------------------------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------ |
-| selector  | Selector                                                                               | A [selector](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors) of an element to wait for |
-| options   | Exclude&lt;[WaitForSelectorOptions](./puppeteer.waitforselectoroptions.md), 'root'&gt; | <i>(Optional)</i> Optional waiting parameters                                                          |
```
OrKoN, line 47: would it be still a breaking change? It looked like root was previously excluded from the API?

jrandolf: Technically since root was exposed in Frame

OrKoN (approved):

PR #8856, opened 2022-08-29
fix: forward timeout waitForFileChooser

OrKoN:


src/util/DeferredPromise.ts
```diff
@@ -31,10 +31,13 @@ export function createDeferredPromiseWithTimer<T>(
     resolver = resolve;
     rejector = reject;
   });
-  const timeoutId = setTimeout(() => {
-    isRejected = true;
-    rejector(new TimeoutError(timeoutMessage));
-  }, timeout);
+  const timeoutId =
+    timeout > 0
```
jrandolf, line 35: Could we do this check earlier and just return a normal deferred promise?

OrKoN: How much earlier do you propose to check for it?

jrandolf: I've made the changes.

src/common/NetworkManager.ts
```diff
@@ -144,9 +144,9 @@ export class NetworkManager extends EventEmitter {
     if (this.#deferredInitPromise) {
       return this.#deferredInitPromise;
     }
-    this.#deferredInitPromise = createDebuggableDeferredPromise<void>(
-      'NetworkManager initialization timed out'
-    );
+    this.#deferredInitPromise = createDeferredPromise<void>({
```
OrKoN: this is not a debuggable promise anymore?

src/util/DeferredPromise.ts
```diff
@@ -11,18 +11,27 @@ export interface DeferredPromise<T> extends Promise<T> {
...
  */
-export function createDeferredPromiseWithTimer<T>(
-  timeoutMessage: string,
-  timeout = 5000
-): DeferredPromise<T> {
+export function createDeferredPromise<T>({
+  message = '',
+  timeout = 5000,
+}: DeferredPromiseOptions = {}): DeferredPromise<T> {
+  if (DEFERRED_PROMISE_DEBUG_TIMEOUT > 0 && !timeout) {
```
OrKoN: hm, I don't think we want to debug all deferred promises

jrandolf-google (approved):

PR #8862, opened 2022-08-30
chore: add injection framework

jrandolf-google:


src/common/IsolatedWorld.ts
```diff
@@ -168,12 +168,15 @@ export class IsolatedWorld {
...
 
-  setContext(context: ExecutionContext): void {
+  async setContext(context: ExecutionContext): Promise<void> {
     assert(
       this.#contextPromise,
       `ExecutionContext ${context._contextId} has already been set.`
     );
     this.#ctxBindings.clear();
+    if (this.#injected) {
+      await context.evaluate(injectedSource);
```
OrKoN: I think that might create a race as the calling code might expect the context to be available immediately after setContext. What about we keep it sync and just schedule the injection?

jrandolf: This won't happen because the context is always awaited.

src/common/IsolatedWorld.ts
```diff
@@ -140,10 +139,11 @@ export class IsolatedWorld {
     return `${name}_${contextId}`;
   };
 
-  constructor(frame: Frame) {
+  constructor(frame: Frame, injected = false) {
```
OrKoN, line 142: WDYT about creating `PuppeteerWorld extends IsolatedWorld` and extending the setContext to schedule evaluation of the injected scripts? it'd also help avoid boolean parameters on the constructor

src/common/IsolatedWorld.ts
OrKoN: Hm, I think it might be that some `#onBindingCalled` messages are missing because the script is being evaluates. Let's see if tests catch anything.

src/common/IsolatedWorld.ts
jrandolf: I'm not a big fan of extending a class as complicated as IsolatedWorld. In any case, we can refactor this later.

src/common/IsolatedWorld.ts
jrandolf: I'll do this in a followup PR.

test/src/injected.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+
+  it('should work', async () => {
+    const {page} = getTestState();
+
+    const handle = await page
+      .mainFrame()
+      .worlds[PUPPETEER_WORLD].evaluateHandle(() => {
+        return InjectedUtil;
+      });
+    expect(handle).toBeTruthy();
```
OrKoN, line 38: I think evaluateHandle always returns a handle, which is truthy even if you return undefined from the evaluated function?

src/common/IsolatedWorld.ts
```diff
@@ -170,11 +170,18 @@ export class IsolatedWorld {
...
-      this.#contextPromise,
+      this.hasContext(),
       `ExecutionContext ${context._contextId} has already been set.`
     );
     this.#ctxBindings.clear();
-    this.#contextPromise.resolve(context);
+    this.#contextPromise.resolve(
+      (async () => {
+        if (this.#injected) {
+          await context.evaluate(injectedSource);
```
OrKoN: I think it would work nicely if we don't await for context.evaluate. The CDP call will be scheduled and it will guarantee that scripts are injected before any other command reaches the renderer.

OrKoN:
Interesting, the test `Cross-origin set-cookie` seems to be failing consistently on this but I don't see why.

test/src/injected.spec.ts
```diff
@@ -0,0 +1,39 @@
+/**
+ * Copyright 2018 Google Inc. All rights reserved.
```
OrKoN: nit: 2022

jrandolf:
> Interesting, the test `Cross-origin set-cookie` seems to be failing consistently on this but I don't see why.

The test was bad. It captured the first response which, because we evaluate something else before now, wasn't the response we were expecting. It passes now with a simply response check.

src/common/IsolatedWorld.ts
jrandolf: Obsolete.

test/src/network.spec.ts
```diff
@@ -853,14 +853,19 @@ describe('network', function () {
...
+            }
+          });
           const url = httpsServer.CROSS_PROCESS_PREFIX + '/setcookie.html';
           page.evaluate(src => {
             const xhr = new XMLHttpRequest();
             xhr.open('GET', src);
             xhr.send();
           }, url);
         });
+        console.log(response.headers());
```
OrKoN: nit: remove console.log

test/src/network.spec.ts
```diff
@@ -853,14 +853,19 @@ describe('network', function () {
         await page.goto(httpsServer.PREFIX + '/setcookie.html');
 
         const response = await new Promise<HTTPResponse>(resolve => {
-          page.on('response', resolve);
+          page.on('response', async response => {
+            if ((await response.text()) === '') {
```
OrKoN: why do we await the response with an empty text? could we leave a comment which response it is or perhaps use the URL for matching?

OrKoN (approved):

PR #8864, opened 2022-08-30
chore: use jobs for artifact generation

jrandolf-google:
This PR adds type generation using jobs and modifies the `tsconfig` files to remove redundant compiles.

This decreases post-first build (`npm run build`) times by 85% (20 seconds to 3 seconds).


.github/workflows/ci.yml
```diff
@@ -164,8 +164,8 @@ jobs:
       - name: Install linux dependencies.
         if: ${{ matrix.spec.name == 'Linux' }}
         run: sudo apt-get install xvfb
-      - name: Build
-        run: npm run build
+      - name: Build tests
```
OrKoN: What's the reason to build tests only instead of the full build?

jrandolf: A production build is now separate from a test build. Test builds do not generate production artifacts.

OrKoN (approved):

PR #8866, opened 2022-08-31
chore: implement a test runner on top of mocha

### LINKED ISSUE #8840: Refactor tests to no encode platform-browser-mode specific expectations
Things like `describeChromeOnly`, `describeFailsFirefox` etc defined in mocha-utils.ts should be removed with regular `describe`/`it`. Instead the browser/mode specific expectations have to be captured in a separate file that is interpreted after all tests have run.

OrKoN:
This PR implements a test runner on top of mocha
that performs multiple mocha runs as defined in
TestSuites.json and compares the outcome of the runs
against TestExpectations.json. This allows us to
remove most of helpers from mocha-utils and be more
flexible when defining the test configurations.

Closes #8840


utils/mochaRunner/src/main.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  let fail = false;
+  const recommendations = [];
+  try {
+    for (const suite of applicableSuites) {
+      const parameters = suite.parameters;
+
+      const applicableExpectations = filterByParameters(
+        filterByPlatform(expectations, platform),
+        parameters
```
thiagowfx, line 80:
```suggestion
        parameters,
```

utils/mochaRunner/src/main.ts
```diff
@@ -0,0 +1,191 @@
...
+        {
+          PUPPETEER_SKIPPED_TEST_CONFIG: JSON.stringify(
+            skippedTests.map(ex => {
+              return ex.testIdPattern;
+            })
+          ),
+        },
+      ]);
+
+      const tmpdir = fs.mkdtempSync(
```
thiagowfx: tmpDir

as per https://google.github.io/styleguide/jsguide.html#naming-rules-common-to-all-identifiers

ditto below for other variables

utils/mochaRunner/src/main.ts
```diff
@@ -0,0 +1,191 @@
...
+        handle.on('error', err => {
+          reject(err);
+        });
+        handle.on('close', () => {
+          resolve();
+        });
+      });
+      console.log('Finished', JSON.stringify(parameters));
+      try {
+        const results = readJSON(tmpfilename) as MochaResults;
```
thiagowfx: https://google.github.io/styleguide/tsguide.html#type-assertions-and-object-literals

> Use type annotations (: Foo) instead of type assertions (as Foo) to specify the type of an object literal. This allows detecting refactoring bugs when the fields of an interface change over time.

utils/mochaRunner/src/main.ts
```diff
@@ -0,0 +1,191 @@
...
+          continue;
+        }
+        console.log('\n================\n');
+      } catch (err) {
+        fail = true;
+        console.error(err);
+      }
+    }
+  } catch (err) {
+    console.log(err);
```
thiagowfx: console.error?

utils/mochaRunner/src/test.ts
```diff
@@ -0,0 +1,49 @@
+import assert from 'assert/strict';
+import test from 'node:test';
+import {
+  filterByParameters as filterByParameters,
```
thiagowfx:
```suggestion
  filterByParameters,
```

utils/mochaRunner/src/main.ts
OrKoN: I think our style guide does not require , and automatically removes it

utils/mochaRunner/src/main.ts
OrKoN: That does not work for assigning unknown type to a type though.

thiagowfx (approved):

utils/mochaRunner/README.md
```diff
@@ -0,0 +1,44 @@
...
+  {
+    "testIdPattern": "[accessibility.spec]",
+    "platforms": ["darwin", "win32", "linux"],
+    "parameters": ["firefox"],
+    "expectations": ["SKIP"]
+  }
+```
+
+`testIdPattern` defines a string that will be used to prefix-match tests. `platforms` defines the platforms the expectation is for (`or`-logic).
+`parameters` defines the parameters that the test has to match (`and`-logic). `expectations` is the list of test results that are considered to be acceptable.
```
sadym-chromium, line 44: How the expectations can be re-baselined?

test/src/mocha-utils.ts
```diff
@@ -63,14 +64,14 @@ export const getTestState = (): PuppeteerTestState => {
...
-  process.env['PRODUCT'] || process.env['PUPPETEER_PRODUCT'] || 'Chromium';
+  process.env['PRODUCT'] || process.env['PUPPETEER_PRODUCT'] || 'chrome';
 
 const alternativeInstall = process.env['PUPPETEER_ALT_INSTALL'] || false;
 
 const headless = (process.env['HEADLESS'] || 'true').trim().toLowerCase();
 const isHeadless = headless === 'true' || headless === 'chrome';
 const isFirefox = product === 'firefox';
-const isChrome = product === 'Chromium';
+const isChrome = product === 'chrome';
```
sadym-chromium, line 74: What is the motivation behind this renaming?

test/src/mocha-utils.ts
```diff
@@ -290,7 +200,7 @@ export const setupTestBrowserHooks = (): void => {
   });
 
   after(async () => {
-    await state.browser!.close();
+    await state.browser?.close();
```
sadym-chromium, line 203: Just wonder how it used to compile before? And why still `await state.context!.close();` is used?

utils/mochaRunner/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Mocha Runner is a test runner on top of mocha. It uses `/test/TestSuites.json` and `/test/TestExpectations.json` files to run mocha tests in multiple configurations and interpret results.
+
+## Running tests for Mocha Runner itself.
+
+```
+npm run build:dev && npx c8 node utils/mochaRunner/lib/test.js
+```
+
+## Running tests using Mocha Runner
```
sadym-chromium, line 11: Does `it.only` filter still works?

utils/mochaRunner/src/main.ts
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    if (!testSuite.platforms.includes(platform)) {
+      console.warn(
+        `Test suite ${testSuiteId} is not enabled for your platform. Running it anyway.`
+      );
+    }
+
+    applicableSuites = [testSuite];
+  }
```
sadym-chromium, line 52: Extract to the helper method to ease reading?

utils/mochaRunner/src/main.ts
```diff
@@ -0,0 +1,190 @@
...
+          console.log('Test run matches expecations');
+          continue;
+        }
+      } catch (err) {
+        fail = true;
+        console.error(err);
+      }
+    }
+  } catch (err) {
+    console.error(err);
```
sadym-chromium, line 156: Maybe avoid false-positive in case of misconfiguration?
```suggestion
    fail = true;
    console.error(err);
```

.github/workflows/ci.yml
```diff
@@ -173,26 +173,12 @@ jobs:
         run: npm run build:dev
       - name: Test types
         run: npm run test:types
-      # On Linux we run all Chrome tests without retries and Firefox tests with retries.
-      - name: Run all Chrome tests (only on Linux)
+      - name: Run all tests with xvfb
```
sadym-chromium, line 176: +1 to run in headful mode as well

sadym-chromium (approved):
I've got a few questions, but in general LGTM.

test/src/mocha-utils.ts
OrKoN: 'Chromium' is not a supported product :)

utils/mochaRunner/README.md
OrKoN: Currently, it's manual process based on the recommendation output from the runner. I was not sure if it needs to be documented separately since the instructions are printer to stdout.

test/src/mocha-utils.ts
OrKoN: `!` ignores undefined so it'd fail calling .close on undefined if the browser was not started. I think the same should be applied to `state.context!.close()`. Thanks for spotting.

utils/mochaRunner/README.md
OrKoN: Yep

.github/workflows/ci.yml
OrKoN: Could you clarify? we have been already running the headful mode tests for a while (no change in behaviour here)

utils/mochaRunner/README.md
OrKoN: Added a note to the readme.

.github/workflows/ci.yml
sadym-chromium: I wasn't aware of that, and it would be useful in Mapper as well.

PR #8869, opened 2022-08-31
feat(chromium): roll to Chromium 106.0.5249.0 (r1036745)

OrKoN:


OrKoN:
Not sure why the ARIA handling has changed in the backend but it looks harmless so we can just update the test.

jrandolf-google (approved):

PR #8875, opened 2022-08-31
fix: allow getting a frame from an elementhandle

OrKoN:
Libraries like pptr-testing-library and expect-puppeteer seem to rely on it.


OrKoN:
@jrandolf any objections to exposing the frame on the element handle now that the execution context is internal?

jrandolf-google (approved):
Nope

PR #8878, opened 2022-09-01
chore: migrate `addScriptTag`

jrandolf-google:


src/common/Frame.ts
```diff
@@ -49,24 +49,29 @@ export interface FrameWaitForFunctionOptions {
...
    */
   path?: string;
   /**
-   * Raw JavaScript content to be injected into the frame.
+   * JavaScript to be injected into the frame.
    */
   content?: string;
   /**
-   * Set the script's `type`. Use `module` in order to load an ES2015 module.
+   * Sets the `type` of thescript. Use `module` in order to load an ES2015 module.
```
OrKoN:
```suggestion
   * Sets the `type` of the script. Use `module` in order to load an ES2015 module.
```

src/common/IsolatedWorld.ts
```diff
@@ -117,7 +117,7 @@ export class IsolatedWorld {
   #frame: Frame;
   #injected: boolean;
   #document?: ElementHandle<Document>;
-  #contextPromise = createDeferredPromise<ExecutionContext>();
+  #context = createDeferredPromise<ExecutionContext>();
```
OrKoN, line 120: revert the rename as it's not the context itself but a promise for a context?

jrandolf: It's better to just keep it as `context` or else we would have to litter the space with `Promise`. We shouldn't care if it's a promise unless we *need* to make the distinction.

OrKoN (approved):

PR #8879, opened 2022-09-01
chore: migrate `addStyleTag`

jrandolf-google:


docs/api/puppeteer.frame.addstyletag_1.md
```diff
@@ -0,0 +1,25 @@
+---
+sidebar_label: Frame.addStyleTag_1
```
OrKoN, line 2: we will have to do something about the overriden methods in the docs eventually

OrKoN (approved):

docs/api/puppeteer.frame.addstyletag_1.md
jrandolf: At some point yes...

PR #8895, opened 2022-09-02
fix: restore deferred promise debugging

### LINKED ISSUE #8832: [Bug]: "Waiting for target frame x failed" when crawling a lot of websites concurrently
### Bug description

1. Do a lot at the same time (opening a lot of browsers and visiting pages etc.), causing Node to lag.
2. Observe an unhandledRejection like the following:
```
TimeoutError: Waiting for target frame 38DAB13DFEC9AD09315770D838D9F8C9 failed
    at Timeout._onTimeout (path\to\my_package\node_modules\puppeteer\src\util\DeferredPromise.ts:36:14)
    at listOnTimeout (node:internal/timers:564:17)
    at processTimers (node:internal/timers:507:7)
```

This is caused by the promise from [FrameManager.ts:153](https://github.com/puppeteer/puppeteer/blob/v16.2.0/src/common/FrameManager.ts#L153) not being fulfilled within 5 seconds (hardcoded in [DeferredPromise.ts](https://github.com/puppeteer/puppeteer/blob/v16.2.0/src/util/DeferredPromise.ts#L24)). If you add `if (isRejected) console.log('resolved anyway')` to [`resolve:` in DeferredPromise.ts](https://github.com/puppeteer/puppeteer/blob/v16.2.0/src/util/DeferredPromise.ts#L46) and set `process.on('uncaughtException', console.error)` then you can see that the promise will be resolved anyway later on.

I think the timeout should probably be configurable in some way (or removed entirely, but it probably serves a purpose).

### Puppeteer version

16.2.0

### Node.js version

18.7.0

### npm version

8.17.0

Windows

### Relevant log output

_No response_

OrKoN:
Closes #8832

### DIFF
--- a/src/util/DebuggableDeferredPromise.ts
+++ b/src/util/DebuggableDeferredPromise.ts
@@ -16,5 +16,5 @@ export function createDebuggableDeferredPromise<T>(
       timeout: DEFERRED_PROMISE_DEBUG_TIMEOUT,
     });
   }
-  return createDeferredPromise({message});
+  return createDeferredPromise();
 }
--- a/src/util/DebuggableDeferredPromise.ts
+++ b/src/util/DebuggableDeferredPromise.ts
@@ -1,0 +1,20 @@
+import {DEFERRED_PROMISE_DEBUG_TIMEOUT} from '../environment.js';
+import {DeferredPromise, createDeferredPromise} from './DeferredPromise.js';
+
+/**
+ * Creates and returns a deferred promise using DEFERRED_PROMISE_DEBUG_TIMEOUT
+ * if it's specified or a normal deferred promise otherwise.
+ *
+ * @internal
+ */
+export function createDebuggableDeferredPromise<T>(
+  message: string
+): DeferredPromise<T> {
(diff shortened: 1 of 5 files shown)


src/util/DebuggableDeferredPromise.ts
```diff
@@ -0,0 +1,20 @@
...
+export function createDebuggableDeferredPromise<T>(
+  message: string
+): DeferredPromise<T> {
+  if (DEFERRED_PROMISE_DEBUG_TIMEOUT > 0) {
+    return createDeferredPromise({
+      message,
+      timeout: DEFERRED_PROMISE_DEBUG_TIMEOUT,
+    });
+  }
+  return createDeferredPromise();
```
sadym-chromium, line 19: Don't we want to keep the message for default timeout?
```suggestion
  return createDeferredPromise({message});
```

sadym-chromium (approved):
LGTM with a minor comment

src/util/DebuggableDeferredPromise.ts
OrKoN: Nope, there should be no timeout and, thus, no message if it's not running the debug mode.

PR #8905, opened 2022-09-06
fix: work around a race in waitForFileChooser

### LINKED ISSUE #6040: waitForFileChooser + click race
When "doing a tap and waiting for a file chooser", the docs suggest the following approach of using waitForFileChooser:
```
const [fileChooser] = await Promise.all([
  page.waitForFileChooser(),
  page.click('#upload-file-button'), // some button that triggers file selection
]);
await fileChooser.accept(['/tmp/myfile.pdf']);
```
This is very racy, as the click might basically happen before the `Page.setInterceptFileChooserDialog` call (setting up browser to track file chooser) to the browser is done.

Also the alternative form:
```
const fileChosen = page.waitForFileChooser();
await page.click('something');
await fileChosen;
```
suffers from the same. Again, the `Page.setInterceptFileChooserDialog` call might not be complete before sending over the click data.

The right order of things would be:
```
await page._client.send('Page.setInterceptFileChooserDialog', {enabled: 1})
let resolve;
const p = new Promise( (r) => resolve = r );
this._fileChooserInterceptors.add(resolve);
await page.click('something');
await p;
```

This is rather incompatible with the current signature of `waitForFileChooser`, as it would have to return 2 promises (or Promise of Promise). The first indicating setup went alright, the second indicating the file chooser has been launched.

(In our testing setup, we workaround this issue by doing an extra race-less `await page._client.send('Page.setInterceptFileChooserDialog', {enabled: 1})` before using `waitForFileChooser`)

OrKoN:
Closes #6040

### DIFF
--- a/src/common/Page.ts
+++ b/src/common/Page.ts
@@ -763,25 +763,28 @@ export class Page extends EventEmitter {
    * await fileChooser.accept(['/tmp/myfile.pdf']);
    * ```
    */
-  async waitForFileChooser(
-    options: WaitTimeoutOptions = {}
-  ): Promise<FileChooser> {
-    if (!this.#fileChooserPromises.size) {
-      await this.#client.send('Page.setInterceptFileChooserDialog', {
-        enabled: true,
-      });
-    }
-
+  waitForFileChooser(options: WaitTimeoutOptions = {}): Promise<FileChooser> {
+    const needsEnable = this.#fileChooserPromises.size === 0;
     const {timeout = this.#timeoutSettings.timeout()} = options;
     const promise = createDeferredPromise<FileChooser>({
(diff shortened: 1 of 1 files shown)


src/common/Page.ts
```diff
@@ -763,25 +763,28 @@ export class Page extends EventEmitter {
...
     const promise = createDeferredPromise<FileChooser>({
       message: `Waiting for \`FileChooser\` failed: ${timeout}ms exceeded`,
       timeout,
     });
     this.#fileChooserPromises.add(promise);
-    return promise.catch(error => {
-      this.#fileChooserPromises.delete(promise);
-      throw error;
-    });
+    let enablePromise: Promise<void> | undefined;
```
sadym-chromium, line 774: `maybeEnablePromise`?

sadym-chromium (approved):

Issue #7859, opened 2021-12-19
[Bug]: stopJSCoverage returns incorrect coverage ranges (does not match Chrome Coverage)


ethaniel:
### Bug description

Steps to reproduce the problem:

1. Download .zip file.
2. In the file `puppeteer_coverage.json` find the ranges for the `js.js` file and see how they don't match `chrome_coverage.json` (which was generated by Chrome manually)
3. **If needed:** run `node ./puppeteer.js` to regenerate `puppeteer_coverage.json`.
4. **If needed:** regenerate `chrome_coverage.json` manually from your local Chrome's Coverages tab. Compare the ranges for the `js.js` file.

[bug.zip](https://github.com/puppeteer/puppeteer/files/7740931/bug.zip)

### Explanation

I've just started playing with Puppeteer to automatically generate the Coverage ranges so I could automatically remove unneeded lines in my Javascript (I know, it's a bad idea, but still).
I've noticed that the ranges returned by Puppeteer do not match the ranges returned by Chrome, which results in broken blocks of code.

**Puppeteer ranges:**
![image](https://user-images.githubusercontent.com/941973/146677988-534665be-98a5-4772-878e-1d7bc21fd555.png)

**Chrome ranges:**
![image](https://user-images.githubusercontent.com/941973/146677983-cff3dbeb-1400-47eb-a031-b26552e30a0c.png)

### Puppeteer version

13.0.0

### Node.js version

17.0.1

### npm version

8.1.0

### What operating system are you seeing the problem on?

macOS

### Relevant log output

```shell
No errors in logs.
```

KiranBestie:
Exported the coverage json of https://armia.com website
Used code shown in the dev tools coverage tab doesn't fall in the range of exported json range.
I'm trying to extract only used code from a file for example; 
https://dlrvetgp71ck5.cloudfront.net/vendor/js/jquery-ui-output.min.js

Due in correct range(start and end values) I couldn't extract them as expected. 

Sample code used;

const puppeteer = require("puppeteer");
const fs = require("fs");
const util = require("util");
const reportJson = require("./Coverage-20220822T155929.json");

const jsCoverage = reportJson.filter(function (item) {
  return item.url.indexOf("jquery-ui-output.min.js") !== -1;
});

console.log(jsCoverage);
let final_css_bytes = "";
let total_bytes = 0;
let used_bytes = 0;

for (const entry of jsCoverage) {
  total_bytes += entry.text.length;
[...]

PR #8911, opened 2022-09-06
fix: add missing code coverage ranges that span only a single character

smithc:

*Bugfix*

**Summary**

This change fixes an issue in the JavaScript code coverage collection algorithm that, under specific circumstances, can lead to incomplete coverage data. 

The existing algorithm filtered out code coverage range results that spanned a single character, perhaps in the assumption that such characters could only represent trivia tokens. This is safe when such tokens represented whitespace, or newline trivia, but unfortunately it also had the effect of filtering out valid semantic tokens as well, such as the ternary operators `?`/`:`.

Furthermore, this particular implementation meant that the resulting output of an otherwise repeatable code-coverage run would differ between manual collection via Chrome Dev Tools, and automated collection via Puppeteer, since the former would correctly include such single-character tokens, and the latter would not.

**Links/Issues**

While this fix was not introduced explicitly to solve for the following issue, a quick search yielded this existing report, and I believe that this may solve for the reported issue (or perhaps at least inform anybody following the issue for a thread to pull on):

- https://github.com/puppeteer/puppeteer/issues/7859

This fix introduces a change in behavior to how we collect coverage ranges. Previously, coverage ranges of a single character length were elided entirely from the coverage output. With this change, those ranges are not ignored, and are included in the resulting coverage output.

From the perspective of consumers, anybody relying on consistent output of coverage ranges from this algorithm (ex. unit tests that rely upon cached or saved copies of coverage data) will need to update their expectations to match these changes accordingly - in essence, consumers should now expect that trivia tokens (like newlines), *or more importantly* single-token operators (like the ternary false operator `:`) will be included as part of the coverage output.  This aligns with the expected output of Chrome's code coverage tooling in Dev Tools.

**Other information**

Please feel free to reach out if there are any questions, and I'll be happy to provide further help or input as needed; I really appreciate all the hard work this team does in providing such an invaluable product!

### DIFF
--- a/src/common/Coverage.ts
+++ b/src/common/Coverage.ts
@@ -484,6 +484,6 @@ function convertToDisjointRanges(
   }
   // Filter out empty ranges.
   return results.filter(range => {
-    return range.end - range.start > 1;
+    return range.end - range.start > 0;
   });
 }
--- a/test/assets/jscoverage/involved.html
+++ b/test/assets/jscoverage/involved.html
@@ -6,6 +6,7 @@ function foo() {
     console.log(2);
   let x = 1 > 2 ? 'foo' : 'bar';
   let y = 1 < 2 ? 'foo' : 'bar';
+  let p = {a:1 > 2?function(){console.log('unused');}:function(){console.log('unused');}};
   let z = () => {};
   let q = () => {};
   q();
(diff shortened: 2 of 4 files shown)


OrKoN (approved):
Thanks for the PR! Looks reasonable to me.

PR #8927, opened 2022-09-08
fix: use win64 version of chromium when on arm64 windows

robrobrobrob:
bugfix

not relavant

not relavant

**Summary**
no Win_arm available on storage.googleapis.com yet, and it currently defaults to win32

nope

**Other information**

### DIFF
--- a/src/node/BrowserFetcher.ts
+++ b/src/node/BrowserFetcher.ts
@@ -21,7 +21,6 @@ import * as util from 'util';
 import * as childProcess from 'child_process';
 import * as https from 'https';
 import * as http from 'http';
-import * as semver from 'semver';
 
 import {Product} from '../common/Product.js';
 import extractZip from 'extract-zip';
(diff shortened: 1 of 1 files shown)


robrobrobrob:
I'm getting some unrelated failed checks, and don't see a way to re-run them. Can I get some help?

OrKoN:
Are win64 binaries executable on arm64 windows?

robrobrobrob:
> Are win64 binaries executable on arm64 windows?

good catch, only as of Windows 11. I'll add that check

robrobrobrob:
should be good now

src/node/BrowserFetcher.ts
```diff
@@ -497,6 +502,25 @@ function parseFolderPath(
   return {product, platform, revision};
 }
 
+/**
+ * Windows 11 is identified by 10.0.22000 or greater
+ * @internal
+ */
+function _isWindows11(version: string): boolean {
```
OrKoN, line 509: Let's add a test to make sure it works as expected? I wonder also how trust-worthy `os.release()` is? does Windows use the same format for various versions?

OrKoN: Will add a test in a follow-up.

OrKoN (approved):

PR #8931, opened 2022-09-09
chore: add custom reporter

OrKoN:
This PR adds a custom mocha reporter that is a combination of the Spec + JSON reporters. This way our test runner can remove custom printing logic for test runs and only output expectations.

### DIFF
--- a/utils/mochaRunner/src/main.ts
+++ b/utils/mochaRunner/src/main.ts
@@ -121,8 +121,9 @@ async function main() {
         [
           '-u',
           path.join(__dirname, 'interface.js'),
-          '--reporter=json',
-          '--reporter-option',
+          '-R',
+          path.join(__dirname, 'reporter.js'),
+          '-O',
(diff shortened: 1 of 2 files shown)


utils/mochaRunner/src/main.ts
```diff
@@ -121,8 +121,9 @@ async function main() {
         [
           '-u',
           path.join(__dirname, 'interface.js'),
-          '--reporter=json',
-          '--reporter-option',
+          '-R',
+          path.join(__dirname, 'reporter.js'),
+          '-O',
```
sadym-chromium, line 126: Are you sure this -O is required?

sadym-chromium (approved):

utils/mochaRunner/src/main.ts
OrKoN: yes, it's the option for the reporter

PR #8932, opened 2022-09-09
chore: support WebDriver BiDi browser instances

OrKoN:
This PR adds a basic support for WebDriver BiDi that currently includes only the ability to establish a connection and shutdown the browser. Therefore, the implementation is marked as internal and won't show up in the changelog as it's barely useful at the moment.

The API classes are kept as classes instead of interfaces so that clients relying on instanceof checks still work.

P.S. bidi/Connection, bidi/Browser and test/bidi/Connection.spec are truly new files in this PR.


src/common/Browser.ts
```diff
@@ -27,196 +26,12 @@ import {TaskQueue} from './TaskQueue.js';
...
-   * Proxy server with optional port to use for all requests.
-   * Username and password can be set in `Page.authenticate`.
-   */
-  proxyServer?: string;
-  /**
-   * Bypass the proxy for the given semi-colon-separated list of hosts.
-   */
-  proxyBypassList?: string[];
-}
+import * as BrowserApi from '../api/Browser.js';
```
jrandolf: Can be just import what is needed?

jrandolf-google (changes requested):
Can we disable `noUnused` for the `api` folder only?

OrKoN:
> Can we disable noUnused for the api folder only?

Nope, typescript does not support that.

src/common/Browser.ts
OrKoN: I don't have a preference but done this mostly to avoid name conflicts and don't assign aliases on case by case basis.

tsconfig.base.json
```diff
@@ -15,7 +15,7 @@
     "noPropertyAccessFromIndexSignature": true,
     "noUncheckedIndexedAccess": true,
     "noUnusedLocals": true,
-    "noUnusedParameters": true,
+    "noUnusedParameters": false,
```
jrandolf: I've figured out a way to get this working. For the virtual classes, just overload the function. For example, instead of
```ts
class {
  fakeMethod(value: string) { throw new Error('unimplemented'); }
}
```
use

```ts
class {
  fakeMethod(value: string): string;
  fakeMethod() { throw new Error('unimplemented'); }
}
```
In TS, if there is one overload, that overload becomes the signature, not the one below it.

OrKoN: @jrandolf does the doc generator properly generate docs for this?

OrKoN: I mean the following works too:

```
class {
  fakeMethod(value: string) { value; throw new Error('unimplemented'); }
}
```

But I think disabling the rule is the simplest since we still have checking in the IDE.

jrandolf: I'm not sure, but it should since TS doesn't even consider the method below the first overload as the public interface.

OrKoN: Seems like the api extractor is happy! Cool approach!

src/common/Browser.ts
OrKoN: Updated

src/api/Browser.ts
```diff
@@ -0,0 +1,626 @@
...
+ * @public
+ */
+export interface BrowserContextOptions {
+  /**
+   * Proxy server with optional port to use for all requests.
+   * Username and password can be set in `Page.authenticate`.
+   */
+  proxyServer?: string;
+  /**
+   * Bypass the proxy for the given semi-colon-separated list of hosts.
```
sadym-chromium: NIT
```suggestion
   * Bypass the proxy for the given list of hosts.
```

jrandolf-google (approved):

src/api/Browser.ts
```diff
@@ -0,0 +1,626 @@
...
+  constructor() {
+    super();
+  }
+
+  /**
+   * @internal
+   */
+  _attach(): Promise<void> {
+    throw new Error('Not implemented');
+  }
```
sadym-chromium, line 229:
```suggestion
  abstract _attach(): Promise<void>;
```

src/common/BrowserConnector.ts
```diff
@@ -118,7 +124,7 @@ export async function _connectToBrowser(
   const {browserContextIds} = await connection.send(
     'Target.getBrowserContexts'
   );
-  const browser = await Browser._create(
+  const browser = await CDPBrowser._create(
```
sadym-chromium, line 123: Maybe rename the method to `_connectToCdpBrowser`?

src/common/bidi/Connection.ts
```diff
@@ -0,0 +1,173 @@
...
+    super();
+    this.#url = url;
+    this.#delay = delay;
+
+    this.#transport = transport;
+    this.#transport.onmessage = this.onMessage.bind(this);
+    this.#transport.onclose = this.#onClose.bind(this);
+  }
+
+  get closed(): boolean {
```
sadym-chromium, line 68: Any particular reason why this is a getter, while `url` is a method?

src/common/bidi/Browser.ts
```diff
@@ -0,0 +1,49 @@
+import * as BrowserApi from '../../api/Browser.js';
+import {Connection} from './Connection.js';
+import {ChildProcess} from 'child_process';
+
+/**
+ * @internal
+ */
+export class Browser extends BrowserApi.Browser {
```
sadym-chromium: Is it called `Browser` instead of `BiDiBrowser` to avoid usage changes?

sadym-chromium (approved):
LGTM in general, but have a few questions

src/api/Browser.ts
OrKoN: abstract would show up in the doc and we don't want that as it is confusing for the end users

src/common/bidi/Connection.ts
OrKoN: No particular reason, but actually, we don't need url here. Going to remove it for now.

src/common/bidi/Browser.ts
OrKoN: yeah, it's already scoped by the folder name so there is no need for a prefix, imo

src/common/BrowserConnector.ts
OrKoN: Done

src/common/bidi/Browser.ts
sadym-chromium: But it ends up with the constructions like `import {Connection as BiDiConnection} from '../common/bidi/Connection.js';`.

OrKoN: Yeah, true, but I don't think it's critical and it can be changed at any point later. We also plan eventually to only support BiDi so there will be only a single Browser/Connection etc.

sadym-chromium: SGTM

PR #8943, opened 2022-09-12
chore: use injections for wait tasks

jrandolf-google:
This PR refactors wait tasks to use injected scripts.


test/src/ariaqueryhandler.spec.ts
```diff
@@ -581,7 +577,7 @@ describe('AriaQueryHandler', () => {
       await page.waitForSelector('aria/zombo', {timeout: 10}).catch(error_ => {
         return (error = error_);
       });
-      expect(error!.stack).toContain('waiting for selector `zombo` failed');
+      expect(error!.stack).toContain('Waiting failed: 10ms exceeded');
```
OrKoN: Can we still have the detailed error messages?

test/src/waittask.spec.ts
```diff
@@ -212,46 +174,6 @@ describe('waittask specs', function () {
...
-          }
-        );
-      } catch (error_) {
-        if (isErrorLike(error_)) {
-          error = error_ as Error;
-        }
-      }
-      expect(error?.message).toContain('polling');
-    });
-    it('should throw negative polling interval', async () => {
```
OrKoN, line 235: let's keep this test?

test/src/ariaqueryhandler.spec.ts
jrandolf: They no longer contain details because the `waitForSelector` API uses `waitForFunction` underneath. We could rewrite them, but this isn't useful since the stack trace contains the line waiting fails at.

test/src/waittask.spec.ts
```diff
@@ -730,9 +661,7 @@ describe('waittask specs', function () {
         return (error = error_);
       });
       expect(error).toBeInstanceOf(puppeteer.errors.TimeoutError);
-      expect(error?.message).toContain(
-        'waiting for selector `.//div` failed: timeout 10ms exceeded'
```
OrKoN, line 734: Are the waiting failure details still in the error message?

jrandolf: See above.

src/injected/util.ts
```diff
@@ -16,3 +18,31 @@ export const createFunction = (
...
+
+  const style = window.getComputedStyle(element);
+  const isVisible =
+    style && style.visibility !== 'hidden' && isBoundingBoxVisible(element);
+  return visible === isVisible ? node : false;
+};
+
+function isBoundingBoxVisible(element: Element): boolean {
+  const rect = element.getBoundingClientRect();
+  return !!(rect.top || rect.bottom || rect.width || rect.height);
```
OrKoN, line 47: I guess there is a bug here? {top: 1, bottom: 1, width: 0, height: 0} would be treated as visible? Let's add unit tests for these functions while we are at it?

test/src/waittask.spec.ts
jrandolf: We don't provide checks for these anymore, we just leave it UB.

test/src/ariaqueryhandler.spec.ts
OrKoN: I think the line does not necessarily tells which selector failed (selector might be computed or received from an external system). I think we should keep these details in the error message: they seem quite useful.

test/src/waittask.spec.ts
OrKoN: why not keep the validation? seems to be a better UX if the users by some chance provide negative values.

src/injected/util.ts
jrandolf: Fixed in https://github.com/puppeteer/puppeteer/pull/8954

test/src/waittask.spec.ts
```diff
@@ -52,55 +52,18 @@ describe('waittask specs', function () {
...
-        .then(() => {
-          return (success = true);
-        });
-      await page.evaluate(async () => {
-        return ((globalThis as any).__FOO = 'hit');
-      });
-      expect(success).toBe(false);
-      await page.evaluate(async () => {
-        return document.body.appendChild(document.createElement('div'));
+        setTimeout(() => {
```
OrKoN, line 64: why is there a need for the timeout now?

src/common/IsolatedWorld.ts
```diff
@@ -507,74 +495,103 @@ export class IsolatedWorld {
...
+        },
+        {
+          bindings,
+          polling: waitForVisible || waitForHidden ? 'raf' : 'mutation',
+          root,
+          timeout,
+        },
+        new LazyArg(async () => {
+          try {
+            // In case CDP fails.
```
OrKoN, line 533: what kind of failure could there be?

OrKoN (approved):
LGTM % some questions

test/src/waittask.spec.ts
jrandolf: This timeout is still needed because it's not guaranteed the poller started. We wait 50 milliseconds because that's half the polling duration; by that time, the poller will for sure have started.

src/common/IsolatedWorld.ts
jrandolf: `LazyArg`s should necessarily be CDP-agnostic because we don't want evaluate calls throwing back up lazy arg errors which themselves are evaluate errors. As an example, navigation can cause a failure here.

OrKoN: What I mean if we should define expected errors (e.g., navigation as you said) and still throw on other errors? at least let's add an error logging so that we see it in the debug log?

test/src/waittask.spec.ts
OrKoN: Could we use setImmediate or setTimeout(0) too? does the poller start on the next tick? or can we have an event that the poller started?

jrandolf: I've tried setTimeout(0) and it's flaky. Since the test requires at least 50ms anyways, this is fine.

src/common/IsolatedWorld.ts
jrandolf: In this case, there is no possible other error besides CDP errors (which must all be ignore). It's possible if we change the injected code that it will throw due to that, but we have tests for that.

PR #8946, opened 2022-09-13
chore: use handle instead of globals for injection

jrandolf-google:
This PR replace globals injection with handle passing. This alleviates the issue of globals creep into the main world.

### DIFF
--- a/src/common/IsolatedWorld.ts
+++ b/src/common/IsolatedWorld.ts
@@ -16,6 +16,7 @@
 
 import {Protocol} from 'devtools-protocol';
 import {source as injectedSource} from '../generated/injected.js';
+import type PuppeteerUtil from '../injected/injected.js';
 import {assert} from '../util/assert.js';
 import {createDeferredPromise} from '../util/DeferredPromise.js';
 import {CDPSession} from './Connection.js';
@@ -114,7 +115,6 @@ export interface IsolatedWorldChart {
  */
 export class IsolatedWorld {
   #frame: Frame;
-  #injected: boolean;
   #document?: ElementHandle<Document>;
   #context = createDeferredPromise<ExecutionContext>();
   #detached = false;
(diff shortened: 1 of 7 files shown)


src/injected/injected.ts
```diff
@@ -2,13 +2,12 @@ import {createDeferredPromise} from '../util/DeferredPromise.js';
...
-      ...Poller,
-      ...util,
-      createDeferredPromise,
-    },
-  })
-);
+const PuppeteerUtil = Object.freeze({
+  ...Poller,
+  ...util,
+  createDeferredPromise,
```
OrKoN, line 6: let's not include stuff that is not used before we need it. I believe only createDeferredPromise is used?

src/common/IsolatedWorld.ts
```diff
@@ -164,13 +168,26 @@ export class IsolatedWorld {
...
     this.#context = createDeferredPromise();
   }
 
   setContext(context: ExecutionContext): void {
-    if (this.#injected) {
-      context.evaluate(injectedSource).catch(debugError);
-    }
+    (async () => {
+      try {
+        this.#puppeteerUtil.resolve(
```
OrKoN: let's extract this in it's own method?

src/templates/injected.ts.tmpl
```diff
@@ -1,10 +1,4 @@
...
-  const InjectedUtil: {
-    createDeferredPromise: typeof createDeferredPromise;
-  };
-}
-
-/** @internal */
+/** 
+ * @internal
+ */
 export const source = SOURCE_CODE;
```
OrKoN, line 8: Let's add a comment that the source code is produced by rollup and in which module format (CJS)?

src/injected/injected.ts
jrandolf: I've kept `util` since we use it for testing.

src/common/IsolatedWorld.ts
jrandolf: Done.

src/templates/injected.ts.tmpl
jrandolf: Done.

src/common/IsolatedWorld.ts
```diff
@@ -164,20 +168,35 @@ export class IsolatedWorld {
 
   clearContext(): void {
     this.#document = undefined;
+    this.#puppeteerUtil = createDeferredPromise();
```
OrKoN, line 171: should we call `dispose` on the previous handle? It looks like we should

jrandolf: The handle is destroyed since the context is destroyed.

OrKoN (approved):

PR #8947, opened 2022-09-13
fix: suppress init errors if the target is closed

### LINKED ISSUE #8944: [Bug]: Protocol error when calling puppeteer.connect()
### Bug description

I am using the basic approach as set out [in this post][1] to connect from a client docker container to any one of a number of chrome docker containers (in a docker swarm/service, potentially across several servers behind nginx, deployed using [CapRover][2]).

In each chrome container I maintain a pool (just a simple array) of `browser` objects, and direct incoming requests to an appropriate `browser` as follows (very similar to the linked post):

	import http from 'node:http';       // https://nodejs.org/api/http.html
	import httpProxy from 'http-proxy'; // https://www.npmjs.com/package/http-proxy

	const proxy = new httpProxy.createProxyServer({ ws: true });

	// an array (pool) of pre-launched and managed browser objects...
	const browsers = [ ... ];

	http
	  .createServer()
	  .on('upgrade', (req, socket, head) => {
		  const browser = browsers[Math.floor(Math.random() * browsers.length)]; // in reality I don't just pick a browser at random
		  const target = browser.wsEndpoint();
		  proxy.ws(req, socket, head, { target });
	  })
	  .listen(3222);

The above is listening at `ws://srv-captain--chrome:3222` (communication is "internal" over the docker network between containers).

Then, in my client container, I connect to the common endpoint `ws://srv-captain--chrome:3222` as follows:

	import puppeteer from 'puppeteer'; // https://www.npmjs.com/package/puppeteer (using version 17.1.3 at time of posting this)
	try {
		const browser = await puppeteer.connect({ browserWSEndpoint: 'ws://srv-captain--chrome:3222' });
	} catch (err) {
		console.error('error connecting to browser', err);
	}

This works really well, except that I am getting *occasional/inconsistent* errors like these when calling `puppeteer.connect()` in the client container above:

    Protocol error (Emulation.setDeviceMetricsOverride): Session closed. Most likely the page has been closed.
    Protocol error (Performance.enable): Target closed.

Almost always, if I simply try to connect again, the connection is made without further error, and at the first attempt.

I have no idea why the error is complaining that the `page has been closed` or `Target closed` since, at this point in the process, I'm not attempting to interact with any `page`, and I know from listening for `browser.on('disconnected'...)`, and also monitoring the chromium processes themselves, that each `browser` in the array is still working fine... none has crashed.

Any idea what's going on here?

__________

**UPDATE after further testing**

Of course, in the client container we don't connect to a `browser` just for the sake of it, like in the above snippet, but to open a `page` and do some stuff with the `page`.  In practice, in the client container it's more like the following test snippet:

	const doIteration = function (i) {
		return new Promise(async (resolve, reject) => {
			// mimic incoming requests coming in at random times over a short period by introducing a random initial delay...
			await new Promise(resolve => setTimeout(resolve, Math.random() * 5000));
			// now actually connect...
			let browser;
			try {
				browser = await puppeteer.connect({ browserWSEndpoint: `ws://srv-captain--chrome:3222?queryParam=loop_${i}` });
			} catch (err) {
				reject(err);
				return;
			}
			// now that we have a browser, open a new page...
			// do something useful with the page (not shown here) and then close it..
			await page.close();
			// now disconnect (but don't close) the browser...
			browser.disconnect();
			resolve();
		});            
	};

	const promises = [];
	for (let i = 0; i < 15; i++) {
		promises.push( doIteration(i) );
	}

	try {
		await Promise.all(promises);
	} catch (err) {
		console.error(`error doing stuff`, err);
	}

Each iteration above is being performed multiple times concurrently... I am using `Promise.all()` on an array of iteration promises to mimic multiple concurrent incoming requests in my production code.  The above is enough to reproduce the problem... the error doesn't happen on calling `puppeteer.connect()` with *every* iteration, just some.

So there seems to be some sort of interplay between opening/closing a `page` in one iteration, and calling `puppeteer.connect()` in another, despite closing the `page` and disconnecting the `browser` properly in each iteration?  This probably also explains the `Most likely the page has been closed` error message when calling `puppeteer.connect()` if there is some hangover relating to a `page` closed in another iteration... though for some reason this error occurs when calling `puppeteer.connect()`?

__________

With the use of a pool of `browser` objects in the `browsers` array, and a docker swarm having multiple containers on multiple servers, each `upgrade` message could be received at a different container (which could even be on a different server) and could be routed to a different `browser` in the `browsers` array.  But I now think that this is a red herring, because in the further testing I narrowed the problem down by routing all requests to `browsers[0]` and also scaling the service down to just one container... so that the `upgrade` messages are always handled by the same container on the same server and routed to the same `browser`... and the problem still occurs.
__________

Full stacktrace for the above-mentioned error:

	Error: Protocol error (Emulation.setDeviceMetricsOverride): Session closed. Most likely the page has been closed.
		at CDPSession.send (file:///root/workspace/myclientapp/node_modules/puppeteer/lib/esm/puppeteer/common/Connection.js:281:35)
		at EmulationManager.emulateViewport (file:///root/workspace/myclientapp/node_modules/puppeteer/lib/esm/puppeteer/common/EmulationManager.js:33:73)
		at Page.setViewport (file:///root/workspace/myclientapp/node_modules/puppeteer/lib/esm/puppeteer/common/Page.js:1776:93)
		at Function._create (file:///root/workspace/myclientapp/node_modules/puppeteer/lib/esm/puppeteer/common/Page.js:242:24)
		at runMicrotasks (<anonymous>)
		at processTicksAndRejections (node:internal/process/task_queues:96:5)
		at async Target.page (file:///root/workspace/myclientapp/node_modules/puppeteer/lib/esm/puppeteer/common/Target.js:123:23)
		at async BrowserContext.pages (file:///root/workspace/myclientapp/node_modules/puppeteer/lib/esm/puppeteer/common/Browser.js:577:23)

  [1]: https://stackoverflow.com/a/49090055/4070848
  [2]: https://caprover.com/

### Puppeteer version

17.1.3

### Node.js version

16.17.0

### npm version

8.15.0

Linux

### Relevant log output

_No response_

OrKoN:
With #8520 Puppeteer is now aware of all targets it connects to. In order to have a not flaky init, Puppeteer waits for all existing targets to be configured during the connection process. This does not work well in case of concurrent connections because while one connection might initialising a target the other one might be closed it. In general, that is expected because we can only be eventually consistent about the target state but we also should not crash the init if some targets have been closed. This PR implements checks to see if the errors are caused by the target or session closures and suppresses them if it's the case.

Also we don't need to call browser.pages() in BrowserConnector as it appears to be redundant for the current TargetManager implementation and forces connection to all existing pages.

Closes #8944

### DIFF
--- a/src/common/Page.ts
+++ b/src/common/Page.ts
@@ -24,7 +24,11 @@ import {
 import {isErrorLike} from '../util/ErrorLike.js';
 import {Accessibility} from './Accessibility.js';
 import {Browser, BrowserContext} from './Browser.js';
-import {CDPSession, CDPSessionEmittedEvents} from './Connection.js';
+import {
+  CDPSession,
+  CDPSessionEmittedEvents,
+  isTargetClosedError,
+} from './Connection.js';
 import {ConsoleMessage, ConsoleMessageType} from './ConsoleMessage.js';
 import {Coverage} from './Coverage.js';
 import {Dialog} from './Dialog.js';
@@ -470,7 +474,15 @@ export class Page extends EventEmitter {
     );
     await page.#initialize();
     if (defaultViewport) {
-      await page.setViewport(defaultViewport);
+      try {
+        await page.setViewport(defaultViewport);
+      } catch (err) {
+        if (isErrorLike(err) && isTargetClosedError(err)) {
+          debugError(err);
+        } else {
+          throw err;
+        }
+      }
     }
     return page;
   }
@@ -645,11 +657,19 @@ export class Page extends EventEmitter {
   };
 
   async #initialize(): Promise<void> {
-    await Promise.all([
-      this.#frameManager.initialize(this.#target._targetId),
-      this.#client.send('Performance.enable'),
-      this.#client.send('Log.enable'),
-    ]);
+    try {
+      await Promise.all([
+        this.#frameManager.initialize(this.#target._targetId),
+        this.#client.send('Performance.enable'),
+        this.#client.send('Log.enable'),
+      ]);
+    } catch (err) {
+      if (isErrorLike(err) && isTargetClosedError(err)) {
+        debugError(err);
+      } else {
+        throw err;
+      }
+    }
   }
 
   async #onFileChooser(
--- a/src/common/BrowserConnector.ts
+++ b/src/common/BrowserConnector.ts
@@ -131,7 +131,6 @@ export async function _connectToBrowser(
     targetFilter,
     isPageTarget
   );
-  await browser.pages();
   return browser;
 }
 
--- a/src/common/Connection.ts
+++ b/src/common/Connection.ts
@@ -445,3 +445,13 @@ function rewriteError(
   error.originalMessage = originalMessage ?? error.originalMessage;
   return error;
 }
+
+/**
+ * @internal
+ */
+export function isTargetClosedError(err: Error): boolean {
+  return (
+    err.message.includes('Target closed') ||
+    err.message.includes('Session closed')
+  );
+}
--- a/src/common/FrameManager.ts
+++ b/src/common/FrameManager.ts
@@ -19,7 +19,7 @@ import {assert} from '../util/assert.js';
 import {createDebuggableDeferredPromise} from '../util/DebuggableDeferredPromise.js';
 import {DeferredPromise} from '../util/DeferredPromise.js';
 import {isErrorLike} from '../util/ErrorLike.js';
-import {CDPSession} from './Connection.js';
+import {CDPSession, isTargetClosedError} from './Connection.js';
 import {EventEmitter} from './EventEmitter.js';
 import {EVALUATION_SCRIPT_URL, ExecutionContext} from './ExecutionContext.js';
 import {Frame} from './Frame.js';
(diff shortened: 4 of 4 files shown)


src/common/Page.ts
```diff
@@ -470,7 +488,7 @@ export class Page extends EventEmitter {
     );
     await page.#initialize();
     if (defaultViewport) {
-      await page.setViewport(defaultViewport);
+      await page.setViewport(defaultViewport).catch(debugErrorIfTargetClosed);
```
jrandolf: Can we use try and catch blocks?

jrandolf-google (approved):

src/common/Page.ts
OrKoN: sure, but in my opinion, it's less readable in this case. Any reasons to prefer one over the other?

jrandolf: It's less about readability and more about semantics. Using a `.catch` usually means even if it errors, return something else (which is void in this case). With try-catch, the semantics is try setting the viewport, and if it breaks, let me check out the error.

EDIT: You are also in an async function already. You wouldn't use `.then` if you were in an async function, so the same principle applies with `.catch`.

PR #8952, opened 2022-09-14
chore: refactor frame tree management

OrKoN:
This PR removes the strong references between
parent and child frames and stores the frame in a
lazy FrameTree with weak references.

Drive-by: fix --no-coverage flag in test runner

### DIFF
--- a/src/common/FrameManager.ts
+++ b/src/common/FrameManager.ts
@@ -16,13 +16,12 @@
 
 import {Protocol} from 'devtools-protocol';
 import {assert} from '../util/assert.js';
-import {createDebuggableDeferredPromise} from '../util/DebuggableDeferredPromise.js';
-import {DeferredPromise} from '../util/DeferredPromise.js';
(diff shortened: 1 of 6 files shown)


jrandolf-google (approved):

src/common/FrameManager.ts
```diff
@@ -284,8 +263,8 @@ export class FrameManager extends EventEmitter {
     frameId: string,
     parentFrameId: string
   ): void {
-    if (this.#frames.has(frameId)) {
-      const frame = this.#frames.get(frameId)!;
+    if (this.frame(frameId)) {
```
jrandolf: Do we have to retrieve twice?

jrandolf-google (changes requested):

src/common/FrameManager.ts
OrKoN: Updated.

jrandolf-google (approved):

Issue #6830, opened 2021-02-07
waitForSelector("…:focus") doesn’t resolve on focus change


andersk:
### Steps to reproduce

**Tell us about your environment:**

* Puppeteer version: 7.0.1
* Platform / OS version: NixOS
* URLs (if applicable):
* Node.js version: 14.15.4

**What steps will reproduce the problem?**

```js
const puppeteer = require("puppeteer");

(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  await page.evaluate(() => {
    const e = document.createElement("input");
    document.body.append(e);
    setTimeout(() => e.focus(), 1000);
  });
  await page.waitForSelector("input:focus");
  await browser.close();
})();
```

**What is the expected result?**

`waitForSelector("input:focus")` should resolve when focus changes to the input element.

**What happens instead?**

`waitForSelector("input:focus")` hangs and times out. I think this is because it’s only listening to DOM mutations, which a focus change is not.

If I `await page.waitForSelector("input:focus", { visible: true })` instead, that resolves as expected.

andersk:
Still reproduces in Puppeteer 15.1.0.

andersk:
Still reproduces in Puppeteer 17.0.0.

OrKoN:
It looks like it does not work because Puppeteer is using the mutation observer to determine when to re-query things. The workarounds are to use `await page.waitForSelector("input:focus", {visible:true})` which should switch to RAF-based polling to `await page.waitForFunction` that polls are regular intervals.

PR #8954, opened 2022-09-14
fix!: fix bounding box visibility conditions

### LINKED ISSUE #8953: [Bug]: Bounding box visibility for `waitForSelector` does not capture 0-sized DOM
The current visibility predicate for bounding box visibility does not generally imply visibility. For example, `{top: 1, bottom: 1, width: 0, height: 0}` is possible, yet the element is not clickable.

jrandolf-google:
This PR implements better conditions for bounding box visibility. We define a bounding box as visible if
1. Some part of the element is viewable in the viewport.
2. It is not hidden, display != `none`, and visibility is set to an attribute showing the element.
3. The width and height are both non-zero.

Fixes: #8953

### DIFF
--- a/test/src/waittask.spec.ts
+++ b/test/src/waittask.spec.ts
@@ -17,6 +17,7 @@
 import expect from 'expect';
 import {isErrorLike} from '../../lib/cjs/puppeteer/util/ErrorLike.js';
 import {
+  createTimeout,
   getTestState,
   setupTestBrowserHooks,
   setupTestPageAndContextHooks,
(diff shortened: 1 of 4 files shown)


test/src/waittask.spec.ts
```diff
@@ -609,7 +699,7 @@ describe('waittask specs', function () {
     it('should have an error message specifically for awaiting an element to be hidden', async () => {
       const {page} = getTestState();
 
-      await page.setContent(`<div></div>`);
```
jrandolf, line 612: @OrKoN It seems tests weren't covering intended behavior, just programmed behavior ;)

test/src/waittask.spec.ts
```diff
@@ -478,113 +479,202 @@ describe('waittask specs', function () {
...
-      let divFound = false;
-      const waitForSelector = page
-        .waitForSelector('div', {visible: true})
-        .then(() => {
-          return (divFound = true);
-        });
-      await page.setContent(
-        `<div style='display: none; visibility: hidden;'>1</div>`
-      );
-      expect(divFound).toBe(false);
```
jrandolf, line 493: @OrKoN I've used timeouts since it could be the case `waitForSelector` hasn't polled after setting content, but we are checking `divFound` here (which is pointless). I've found `40ms` to be the sweet spot.

test/src/waittask.spec.ts
```diff
@@ -478,113 +479,186 @@ describe('waittask specs', function () {
...
+      const element = await page.evaluateHandle(() => {
+        return document.getElementsByTagName('div')[0]!;
       });
-      expect(divFound).toBe(false);
-      await page.evaluate(() => {
-        return document
-          .querySelector('div')
-          ?.style.removeProperty('visibility');
+      await expect(
+        Promise.race([promise, createTimeout(40)])
```
OrKoN, line 491: wha do we need a timeout here? generally, timeouts in tests make things more flaky, can we avoid it?

test/src/waittask.spec.ts
OrKoN: I actually wonder if it's going to be a breaking change since people likely rely on this? Anyway I think we should fix it but perhaps see if we mark it as a breaking change and land after all the refactoring and features.

test/src/waittask.spec.ts
jrandolf: I've written a comment about this. See above. It won't make the tests flaky because we are just assuring the default hasn't changed. If it does, then the test did its job.

test/src/waittask.spec.ts
jrandolf: Could we land it together? All the refactors and features are on top of this.

OrKoN: Let's mark it as non breaking then as it's technically a fix. Let's include the definition of what a visible bounding box is in the commit description.

jrandolf: I've marked it as a breaking fix. It is still breaking, but it's a fix :P

OrKoN (approved):

heikkipora:
We use `Page.waitForSelector({visible: true})` quite heavily even for simple cases like this:

1. click a button
2. wait for a new element to appear in DOM and be visible (i.e. not explicitly hidden) - just not necessarily inside the viewport

This change breaks a few thousand test cases for us and working around it is tedious. 
The the referenced bug (https://github.com/puppeteer/puppeteer/issues/8953) seems to have nothing to do with the viewport - so I'm more than a bit curious why that (viewport rules) has been changed as well.

Suggestion: maybe this new behavior could be moved behind a parameter/argument instead to reduce the impact on existing codebases - or remove the checks about inclusion in viewport altogether.

Additionally, Puppeteer's API docs still state that `Visible: A boolean wait for element to be present in DOM and to be visible, i.e. to not have display: none or visibility: hidden CSS properties. Defaults to false.`

jrandolf:
@heikkipora We want visibility to be defined as user-interactive since that is the main use case (which in hindsight should probably be renamed to interactive). I wouldn't mind implementing a flag that can support your use-case, but do you mind providing more information? (I'm thinking it could be used for animations)

In any case, waitForSelector than waitForFunction is always an option.

joakimgunst:
We have the same issue as @heikkipora. We have many test cases where we perform an action, and then wait for an element to be visible in the page, but not necessarily in the viewport. Using `waitForSelector` without `visible: true` works, but doesn't fail when the element has `display: none`,  `visibility: hidden` or an empty bounding box, all of which are useful conditions to check.

For me it is not intuitive to consider an element that is otherwise visible but not in the viewport not visible. Other libraries like [Playwright](https://playwright.dev/docs/actionability#visible) and [Cypress](https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Visibility) also do not include this in their definitons of visibility.

pasieronen:
One use case I have is clicking e.g. a form submit button and then waiting with `page.waitForSelector('.form-successfully-sent', {visible: true})` before continuing.  Depending on how the site is implemented, the `.form-successfully-sent` element could exist in the DOM even before the form is submitted (with `display: none`). 

But once the element is visible, the test doesn't really care if it's inside the viewport. Often it probably is, but if waitForSelector requires this, then the code breaks if e.g. texts earlier in the page get longer...

krisatverbidio:
We also have the same issue as @heikkipora and @joakimgunst .  We use visibility to determine if it is in the dom, even if it is off page. This is a major breaking change without a workaround.

jrandolf:
> it is not intuitive to consider an element that is otherwise visible but not in the viewport not visible.
@joakimgunst As a user, this is very intuitive. In fact, we should technically be stricter and say opacity should be included, but we don't because a user can interact with transparent elements. Visibility in our case is defined as visible by the user.

> Other libraries like [Playwright](https://playwright.dev/docs/actionability#visible) and [Cypress](https://docs.cypress.io/guides/core-concepts/interacting-with-elements#Visibility) also do not include this in their definitons of visibility.
Playwright is a fork of Puppeteer and Cypress uses Puppeteer... it's not strange they do not include it since we were the originators of their definitions.

> Depending on how the site is implemented, the .form-successfully-sent element could exist in the DOM even before the form is submitted (with display: none).
@pasieronen Please specify the implementation. This is not helpful in understanding what the problem space is.

> We use visibility to determine if it is in the dom, even if it is off page. This is a major breaking change without a workaround.
@krisatverbidio Why not just use `waitForSelector` without visibility?

andersk:
The Puppeteer tests for [Zulip](https://github.com/zulip/zulip/) are also broken by this change. A few thoughts:

> We want visibility to be defined as user-interactive

An element that’s outside the viewport but could be scrolled into view should count as user-interactive, because the `.click()` method will scroll it into view.

> Why not just use `waitForSelector` without visibility?

In addition to the obvious purpose of waiting for an element to be marked visible in the DOM, we also use it as a workaround for other problems like #6830.

pasieronen:
> > Depending on how the site is implemented, the .form-successfully-sent element could exist in the DOM even before the form is submitted (with display: none).
> 
> @pasieronen Please specify the implementation. This is not helpful in understanding what the problem space is.

Back in the "jQuery era of Javascript", it was common to serve static HTML that contained many blocks with `display: none` and then show those when needed with `$(...).show()`. Here's a fairly typical example from that era, where both `#success_message` and `#error_message` are in the DOM: http://reusableforms.com/d/o5/html5-contact-form-send-email 

I agree with [krisatverbidio](https://github.com/krisatverbidio) that this is a major breaking change - perhaps `waitForSelector`should have an additional flag like `inViewport: true` to control this behavior?

jrandolf:
> An element that’s outside the viewport but could be scrolled into view should count as user-interactive, because the .click() method will scroll it into view.

@andersk This was definitely not intentional. I've pushed out https://github.com/puppeteer/puppeteer/pull/9087 to make it required that the element can be at least scrolled to by checking scroll bounds instead of viewport. Would this settle the problem?

joakimgunst:
How about the case where the element is inside a scroll container that is not the `body` element? For example, if I understand the fix correctly, on [this page](https://developer.mozilla.org/en-US/docs/Web/API/Element/hasAttribute) the `XMLDocument` link in the sidebar would not be considered visible, even though the user can scroll to it.

jrandolf:
After some discussion, we've decided to remove the condition!

PR #8956, opened 2022-09-14
feat: add `text` query handler

jrandolf-google:


test/src/queryhandler.spec.ts
```diff
@@ -94,6 +94,85 @@ describe('Query handler tests', function () {
     });
   });
 
+  describe('Text selectors', function () {
+    describe('in Page', function () {
+      it('should query existing element', async () => {
```
OrKoN, line 99: let's add the following tests:

1) deeply nested text `<div><div><span>text` to make sure that span is returned.
2) tests involving 'checkbox', 'image', 'radio' and an input (e.g., textarea or text)
3) test for the searched text spanning multiple element (`<div><span>a</span><span>b</span><div>` && `text/ab`)
4) let's test the order in which elements are found if there are multiple matches

test/src/page.spec.ts
```diff
@@ -545,39 +545,79 @@ describe('Page', function () {
     it('should work', async () => {
```
OrKoN, line 545: was the change here needed for the text handlers?

OrKoN: do we add more objects to the main world now?

jrandolf: There are some sets in the injected code, and since we always create a handle for the injected code, it's probably queryable as well. An unfortunate side-effect.

OrKoN: WDYT about changing the test to use a custom class? E.g., evaluate(`class MyClass {}`) and then query objects with MyClass.prototype?

jrandolf: I thought about this. It sounds reasonable! Let's do it.

test/src/queryhandler.spec.ts
```diff
@@ -94,6 +94,130 @@ describe('Query handler tests', function () {
...
+
+        const elements = await page.$$('text/a');
+        expect(elements.length).toBe(2);
+      });
+      it('should pierce shadow DOM', async () => {
+        const {page} = getTestState();
+
+        await page.setContent(`
+        <div>
+          <div shadowroot="open">
```
OrKoN: oh this is not a shadow root actually, just a regular div. I guess you meant to write smth like

```
<host-element>
  <template shadowroot="open">
    <div>a</div>
    <div>b</div>
  </template>
</host-element>
```

test/src/queryhandler.spec.ts
```diff
@@ -94,6 +94,130 @@ describe('Query handler tests', function () {
...
+      it('should return empty array for non-existing element', async () => {
+        const {page} = getTestState();
+
+        expect(await page.$('text/test')).toBeFalsy();
+        expect((await page.$$('text/test')).length).toBe(0);
+      });
+      it('should return first element', async () => {
+        const {page} = getTestState();
+
+        await page.setContent('<div>a</div><div>b</div>');
```
OrKoN: I think you meant `<div>a</div><div>a</div>` and verify that 

```
const element = await page.$('text/a');
        expect(
          await element?.evaluate(e => {
            return e == document.querySelector('div');
          })
        ).toBe('a');
```

test/src/queryhandler.spec.ts
jrandolf: Ah, okay. The problem is firefox fails with
```
<div>
  <template shadowroot="open">
    <div>a</div>
    <div>b</div>
  </template>
</div>
```

OrKoN: you can also create a shadow root and a custom element via JS. I think Firefox might not support declarative shadow DOM

OrKoN (approved):

PR #8958, opened 2022-09-15
chore: add selector computation utilities

jrandolf-google:
This PR adds selector computation utilities for future internal use.


OrKoN (changes requested):
I don't think selector computation is currently needed for any Puppeteer features. Let's discuss where we put it. Also some code looks familiar so perhaps it's missing proper license headers.

PR #8961, opened 2022-09-15
chore: incrementally erase cache on text change

jrandolf-google:


src/injected/TextContent.ts
```diff
@@ -67,10 +89,19 @@ export const createTextContent = (root: Node): TextContent => {
   if (!isSuitableNodeForTextMatching(root)) {
     return value;
   }
+
   let currentImmediate = '';
   if (isNonTrivialValueNode(root)) {
     value.full = root.value;
     value.immediate.push(root.value);
+
+    root.addEventListener(
```
OrKoN, line 98: is it always running in the secondary world? I believe the latest approach was that we adopt the handle for utils in the main world. In that case, the listener will be observable to the page.

OrKoN: Actually, maybe I misinterpreting this. Code LGTM!

OrKoN (approved):

src/injected/TextContent.ts
jrandolf: This just cleans the cache for where the utility is. It's scoped to the injected script, not leaked to the outer world.

PR #9001, opened 2022-09-22
chore: remove environment variable that forces WebRender to be disabl…

whimboo:
This environment variable has been added a while ago because Firefox was crashing on Linux platforms where only Soft Rendering was available. As such we had to disable WebRender completely.

After some testing I cannot see that this is a problem anymore and as such lets try if we can finally remove it and run Firefox with its default configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "node": ">=14.1.0"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

whimboo:
As reference here the related Bugzilla ticket: https://bugzilla.mozilla.org/show_bug.cgi?id=1693011

PR #9002, opened 2022-09-22
chore: enable firefox tests on Mac

OrKoN:


whimboo (changes requested):
Thanks Alex!

Before I review you basically want to update the test expectations especially around all the skipped tests. All of them that we skip on Linux will also have to be skipped on MacOS and Windows.

whimboo:
@OrKoN nearly all the tests got canceled. I don't see why.

.github/workflows/ci.yml
```diff
@@ -209,6 +209,11 @@ jobs:
             xvfb: true
             node: 18
             suite: firefox-bidi
+          - name: firefox-headless
+            machine: macos-latest
+            xvfb: false
+            node: 18
+            suite: firefox-headless
```
whimboo, line 216: I wonder if also the `headful` variant should be run on Windows? Looks like it's also not done for Chrome.

whimboo (approved):
Just some nits / questions. Thanks for adding the MacOS support! 

Would you mind filing an issue for Windows? We / I could work on getting this platform added as well.

whimboo:
Oh, automerge was already enabled. :( Anyway happy to see follow-ups for the last raised issue.

PR #9014, opened 2022-09-26
docs: add documentation to xpath prefix and deprecated page.waitForTimeout

pecampelo:

This PR introduces addition to documentation of two parts of this repository:

- the documentation for the deprecated `page.waitForTimeout` function, recommending the use of `await new Promise`.

- the documentation for the xpath prefix in `waitForSelector` and `$` functions.

I did not, as there were only changes added to the docs.

I did.

**Summary**

I decided to recommend a small change to the deprecation comment for the page.waitForTimeout function whilst reading the new change after updating my local puppeteer. I felt `Use 'await new Promise'` could also refer to await new Promise being deprecated, meaning one should use instead page.waitForTimeout. I felt the way I wrote in the PR could clear up the confusion.

The documentation for the xpath prefix in `waitForSelector` and `$` functions came up when I read the new possible use for the aforementioned xpath selector, and, when reading the type description while using Puppeteer itself, I couldn't quite understand how the syntax would work comparing it to Devtools so I had to experiment with not putting the prefix, then adding it in multiple ways, `xpath-` and `xpath/` until I understood that the slash is needed even when wanting to find an element inside the whole page. (e.g.: `xpath///a[text() = "Page title"]`) 

It does not. It merely explains a bit of the newer implementations of waitForSelector functions.


jrandolf:
You should probably run `npm run docs`

pecampelo:
@jrandolf, I believe the .md files have been generated successfully.

pecampelo:
@jrandolf would you please review this PR?

jrandolf:
Hey @pecampelo, we are in the middle of a migration to a monorepo. I'll take a look once the migration is complete.

jrandolf-google (approved):

Issue #8922, opened 2022-09-08
[Feature] Monorepo


OrKoN:
Instead of generating the puppeteer-core package based on the puppeteer package, we should separate the two packages and make `puppeteer` to depend on `puppeteer-core`.

OrKoN:
- https://docs.npmjs.com/cli/v8/using-npm/workspaces
- https://github.com/googleapis/release-please/blob/main/docs/manifest-releaser.md

PR #9022, opened 2022-09-28
chore: initiate monorepo migration

jrandolf-google:
This PR starts the monorepo migrations as per https://github.com/puppeteer/puppeteer/issues/8922. To scope migrations, we are only moving the `testserver` into a separate package. Further migrations will come later.


scripts/ensure-pinned-deps.ts
```diff
@@ -21,7 +21,7 @@ const allDeps = {...packageJson.dependencies, ...packageJson.devDependencies};
 const invalidDeps = new Map<string, string>();
 
 for (const [depKey, depValue] of Object.entries(allDeps)) {
-  if (/[0-9]/.test(depValue[0]!)) {
+  if (/[0-9\*]/.test(depValue[0]!)) {
```
mathiasbynens: This kinda defeats the purpose of this whole “ensure pinned deps” script :)

Instead of allowing `*` generally, we could hardcode an exception for the `@pptr/testserver` package, which is the only one you’d like to use `*` for IIUC. WDYT?

mathiasbynens (changes requested):

scripts/ensure-pinned-deps.ts
jrandolf: Done. (TIL `*` means any package)

scripts/ensure-pinned-deps.ts
```diff
@@ -14,14 +14,28 @@
  * limitations under the License.
  */
 
+import {readdirSync, readFileSync} from 'fs';
+import {join} from 'path';
 import packageJson from '../package.json';
 
+const LOCAL_PACKAGE_NAMES = [];
```
mathiasbynens, line 21: This should be a `Set`

mathiasbynens (approved):
LGTM % turning `LOCAL_PACKAGE_NAMES` into a `Set`

scripts/ensure-pinned-deps.ts
jrandolf: Sets are generally slower for a small number of objects. Even for scanning operations.

scripts/ensure-pinned-deps.ts
GetBentGaming: Hello guys I have a quick question

PR #9023, opened 2022-09-28
feat: separate puppeteer and puppeteer-core

jrandolf-google:
This PR moves the puppeteer source code into separate monorepo packages:

- `puppeteer` and `puppeteer-core` are now separated into their own packages.
- `puppeteer-core` has a new exports called `puppeteer-core/internal` for internal usage. 

Tests and various tools have been updated to accommodate the migration.


mathiasbynens:
I suspect Alex might have opinions on this. I’ll RLGTM this for now but please do address his feedback, if any, after landing!

mathiasbynens (approved):

.github/workflows/publish.yml
```diff
@@ -15,25 +15,12 @@ jobs:
...
-        run: npm run build
-      - name: Publish puppeteer
-        env:
-          NPM_TOKEN: ${{secrets.NPM_TOKEN_PUPPETEER}}
         run: |
-          npm config set registry 'https://wombat-dressing-room.appspot.com/'
-          npm config set '//wombat-dressing-room.appspot.com/:_authToken' '${NPM_TOKEN}'
-          echo "Publishing puppeteer"
-          npm publish
-          # DEPRECATED_RANGE=$(node utils/get_deprecated_version_range.js)
```
OrKoN, line 29: Let's keep the code for the deprecation of old puppeteer releases (only Puppeteer though). There is a blocking issue in wombat that prevents it from working but we have to have this eventually.

.github/workflows/publish.yml
```diff
@@ -15,25 +15,12 @@ jobs:
...
+        run: npm run build
+      - name: Publish packages
         env:
-          NPM_TOKEN: ${{secrets.NPM_TOKEN_PUPPETEER_CORE}}
-        run: |
-          utils/prepare_puppeteer_core.js
-          npm config set registry 'https://wombat-dressing-room.appspot.com/'
-          npm config set '//wombat-dressing-room.appspot.com/:_authToken' '${NPM_TOKEN}'
-          npm publish
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN_RELEASE}}
```
OrKoN, line 23: so will it work with 2FA?

docs/api/puppeteer.launch.md
```diff
@@ -7,6 +7,6 @@ sidebar_label: launch
 **Signature:**
 
 ```typescript
-launch: (options?: import('./types.js').PuppeteerLaunchOptions) =>
-  Promise<import('./types.js').Browser>;
+launch: (options?: import('./node/PuppeteerNode.js').PuppeteerLaunchOptions) =>
+  Promise<import('puppeteer-core/internal/api/Browser.js').Browser>;
```
OrKoN: should not we import from the public types here and above?

package.json
```diff
@@ -38,51 +15,41 @@
...
-    "devtools-protocol": "0.0.1036444",
-    "extract-zip": "2.0.1",
-    "https-proxy-agent": "5.0.1",
-    "progress": "2.0.3",
-    "proxy-from-env": "1.1.0",
-    "rimraf": "3.0.2",
-    "tar-fs": "2.1.1",
-    "unbzip2-stream": "1.4.3",
-    "ws": "8.9.0"
+  "wireit": {
```
OrKoN, line 37: is it essential to have wireit? The home page says it's alpha level.

packages/puppeteer-core/package.json
```diff
@@ -0,0 +1,158 @@
...
+    "lib",
+    "!*.tsbuildinfo"
+  ],
+  "author": "The Chromium Authors",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "devtools-protocol": "0.0.1036444"
+  },
+  "peerDependencies": {
+    "ws": "8.9.0",
```
OrKoN: why are these dependencies moved to the peerDependencies? I don't think we want to force the user to install those separately? Or is there some lerna feature that makes them to be prod dependencies when publishing?

packages/puppeteer-core/src/puppeteer-core.ts
```diff
@@ -14,21 +14,16 @@
...
+const puppeteer = new Puppeteer({isPuppeteerCore: true});
 
-export const {
-  connect,
-  createBrowserFetcher,
-  defaultArgs,
-  executablePath,
-  launch,
-} = puppeteer;
+export const {connect} = puppeteer;
```
OrKoN: I think `launch` and other methods should be available still? I think it's perfectly fine to launch a browser with the user provided path using puppeteer-core. Other methods also make sense.

release-please-config.json
```diff
@@ -1,8 +1,7 @@
 {
   "packages": {
-    ".": {
-      "releaseType": "node",
-      "include-component-in-tag": false
```
OrKoN, line 5: does it change the tag names for github releases?

.github/workflows/publish.yml
jrandolf: We can always look back at old releases to find the code. I don't see a purpose in keeping it here.

.github/workflows/publish.yml
jrandolf: Yes, the packages have the repository hardcoded in their package.json

docs/api/puppeteer.launch.md
jrandolf: This is a problem with API extractor which existed before. We can forward fix this.

package.json
jrandolf: Wireit vastly improved build times and is very much necessary for to help with building. It's similar to my job framework I built, but more automated. At some point, all the commands should migrate to this.

.github/workflows/ci.yml
```diff
@@ -258,7 +258,8 @@ jobs:
           npm run clean
           npm run build
           # Note: this modifies package.json to test puppeteer-core, so we test this last.
-          npm run test:install
+          # TODO(jrandolf): Disabled for now.
```
OrKoN: Is it solved in the same way as the docker test issue? I'd suggest fixing the docker and bundling/installation tests as part of this PR.

.github/workflows/publish.yml
OrKoN: Does hardcoding the proxy as a registry name in the package.json affect anything for the end-users (e.g., during npm install etc)?

OrKoN: Found the docs 📝

jrandolf: No, it's just for development

release-please-config.json
jrandolf: Yes, tags are now by package

.github/workflows/ci.yml
jrandolf: The docker issue and install test issue are similar. I think it would be better to keep the PR as is to scope it to the separation.

release-please-config.json
OrKoN: Will there still be a single release entry on GitHub? or multiple (per package)? can we change that?

jrandolf: There will be multiple release entries on GitHub. We can't change this (and we shouldn't because the releases are necessary for our token).

.github/workflows/ci.yml
OrKoN: I think it's better to author it as a second commit in this PR but it's better to keep the test coverage in the same PR.

docker/Dockerfile
```diff
@@ -15,13 +15,15 @@ RUN apt-get update \
...
+COPY puppeteer-core-latest.tgz /home/pptruser/puppeteer-core-latest.tgz
 
-# Install puppeteer into /home/pptruser/node_modules.
-RUN npm i ./puppeteer-latest.tgz \
-    && rm puppeteer-latest.tgz \
-    # Add user so we don't need --no-sandbox.
-    # same layer as npm install to keep re-chowned files from using up several hundred MBs more space
-    && groupadd -r pptruser && useradd -r -g pptruser -G audio,video pptruser \
+# Install puppeteer and puppeteer-core into /home/pptruser/node_modules.
+RUN npm i ./puppeteer-core-latest.tgz ./puppeteer-latest.tgz
```
OrKoN: Let's keep it in one line, it helps to keep the image smaller

docs/api/puppeteer.browser.md
```diff
@@ -4,7 +4,7 @@ sidebar_label: Browser
 
 # Browser class
 
-A Browser is created when Puppeteer connects to a Chromium instance, either through [PuppeteerNode.launch()](./puppeteer.puppeteernode.launch.md) or [Puppeteer.connect()](./puppeteer.puppeteer.connect.md).
+A Browser is created when Puppeteer connects to a Chromium instance, either through or [Puppeteer.connect()](./puppeteer.puppeteer.connect.md).
```
OrKoN: broken doc?

.github/workflows/ci.yml
```diff
@@ -253,12 +256,7 @@ jobs:
         run: npm run test -- --test-suite ${{ matrix.spec.suite }}
       - name: Test bundling and installation
         if: ${{ matrix.spec.machine == 'ubuntu-latest' }}
-        run: |
-          # Test prod build instead of dev.
-          npm run clean
-          npm run build
-          # Note: this modifies package.json to test puppeteer-core, so we test this last.
-          npm run test:install
+        run: npm run test:install
```
OrKoN: Let's keep the prod build here as I think the test build might drift away and produce broken builds.

jrandolf: Let's optimize this later. Not sure if it makes sense to optimize this now since it will slow down CI without any reason.

packages/puppeteer/package.json
```diff
@@ -0,0 +1,138 @@
...
+    "./*": {
+      "import": "./*",
+      "require": "./*"
+    }
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/puppeteer/puppeteer/tree/main/packages/puppeteer"
+  },
+  "publishConfig": {
```
OrKoN: do we still need this?

utils/insert_file_dependencies.ts
```diff
@@ -0,0 +1,44 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {readFile, rm, writeFile} from 'fs/promises';
+import {join, resolve} from 'path';
+
+const CACHE_DIRECTORY = join(__dirname, '..', '.cache', 'puppeteer');
+
+const args = process.argv.slice(2);
```
OrKoN: Please add a comment on the top explaining what this does and what are the input params.

utils/resolve_file_dependencies.ts
```diff
@@ -0,0 +1,66 @@
...
+
+import {existsSync} from 'fs';
+import {mkdir, readFile, writeFile} from 'fs/promises';
+import {dirname, join, normalize, resolve} from 'path';
+
+const CACHE_DIRECTORY = join(__dirname, '..', '.cache', 'puppeteer');
+
+const args = process.argv.slice(2);
+
+(async () => {
```
OrKoN: Please add a comment on the top explaining what this does and what are the input params.

packages/puppeteer/package.json
```diff
@@ -0,0 +1,138 @@
...
+    "install.js",
+    "!*.tsbuildinfo"
+  ],
+  "author": "The Chromium Authors",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "https-proxy-agent": "5.0.1",
+    "progress": "2.0.3",
+    "proxy-from-env": "1.1.0",
+    "puppeteer-core": "file:../puppeteer-core"
```
OrKoN: I thought we should be able to use a version here and NPM would be smart enough to link it automatically and during the release the version can be updated?

.github/workflows/ci.yml
OrKoN: Well, let's refactor this later then and keep the current code?

packages/puppeteer/package.json
jrandolf: Not sure release-please would automate this. `release-please` wasn't built with this in mind since lerna originally did all the versioning. We are basically doing that specific part of lerna here.

packages/puppeteer/package.json
jrandolf: Yes

OrKoN: Let's revert to setting the registry in the publish.yml as configuring it per package does not seem ideal as it can be easily forgotten.

OrKoN (approved):

package.json
aomarks: Hey! Awesome to see that you're using wireit. Big puppeteer fan :)

Please feel free to reach out if you have any questions/comments.

We're currently working on adding "services" to Wireit which may be of interest to you, since it can be really helpful for setting up test environments with multiple coordinating servers etc.: https://github.com/google/wireit/issues/33

jrandolf: Hey @aomarks, I actually have a feature request. It'd be nice to have phony targets.

aomarks: Cool! Can you describe how a phony target would work? (Feel free to open an issue to discuss further)

jrandolf: A phony target is basically a target with no corresponding npm command. This would help with organising commands and prevent people from running "intermediate" commands.

jrandolf: Additionally, it would make sense to have a .wireitrc or a wireit.config.js since we are polluting package.json which will add additional bytes for library users.

aomarks: > A phony target is basically a target with no corresponding npm command. This would help with organising commands and prevent people from running "intermediate" commands.

If I understand correctly, that's already supported! Check out https://github.com/lit/lit/blob/9e978a0c38eddf5631a1dde9bdb182cccbb33a49/package.json#L166 for an example. Just specify dependencies and nothing else.

> Additionally, it would make sense to have a .wireitrc or a wireit.config.js since we are polluting package.json which will add additional bytes for library users.

We've heard that request, yeah. See https://github.com/google/wireit/issues/152. I'm a little conflicted on it, some reasoning written in that issue. Not set in stone, though.

Thanks!

jrandolf: For the intermediate targets, we have a misunderstanding. The targets should not be npm targets. They should just be .wireit targets.

aomarks: Hey @jrandolf! The phony targets feature discussed above is now available in wireit 0.9.4: https://github.com/google/wireit/blob/main/CHANGELOG.md#094---2023-01-30

PR #9027, opened 2022-09-29
feat: add ability to collect JS code coverage at the function level

smithc:

_Feature_

**Summary**

Google Chrome allows for two distinct code coverage modes: ["Per Function" and "Per Block"](https://developer.chrome.com/blog/new-in-devtools-80/#modes). Currently, Puppeteer only supports the latter of these two modes. 

This PR introduces an extra configuration option for code collection, allowing users to utilize either "Per Function", or "Per Block" collection, the same as is available within Chrome.

This is a non-breaking change. The default collection setting remains at the block-level to preserve existing behavior.


jrandolf-google (approved):

src/common/Coverage.ts
```diff
@@ -87,6 +91,22 @@ export interface CSSCoverageOptions {
   resetOnNavigation?: boolean;
 }
 
+/**
+ * Defines the scope of JS coverage collection to be either at the function level,
+ * or the block level.
+ */
```
jrandolf: Declare as `@public`.

src/common/Coverage.ts
```diff
@@ -87,6 +91,22 @@ export interface CSSCoverageOptions {
...
+export const enum JSCoverageScope {
+  /**
+   * Collect coverage at the function level.
+   */
+  PerFunction = 'perfunction',
```
jrandolf: Not used? Looks like we can just use a boolean.

src/common/Coverage.ts
```diff
@@ -204,13 +225,15 @@ export class JSCoverage {
       resetOnNavigation?: boolean;
       reportAnonymousScripts?: boolean;
       includeRawScriptCoverage?: boolean;
+      coverageScope?: JSCoverageScope;
```
jrandolf: Since there are only two values, let's just use `usePerFunctionScope` with default as `false`.

jrandolf-google (changes requested):
PTAL

smithc:
Thanks for the review @jrandolf. Feedback has been addressed 👍

packages/puppeteer-core/src/common/Coverage.ts
```diff
@@ -74,6 +74,11 @@ export interface JSCoverageOptions {
    * Whether the result includes raw V8 script coverage entries.
    */
   includeRawScriptCoverage?: boolean;
+  /**
+   * Whether to collect coverage information at the function level.
+   * If false, coverage will be collected at the block level.
+   */
+  usePerFunctionScope?: boolean;
```
jrandolf: Actually, could you name this `useBlockCoverage` with a default of `true`. Not sure if people will understand `usePerFunctionScope`. They will with `useBlockCoverage` because that's what the docs call it (https://chromedevtools.github.io/devtools-protocol/tot/Profiler/)

smithc:
@jrandolf - updated with the latest feedback

PR #9069, opened 2022-10-06
chore: pin checkout action

jrandolf-google:


.github/workflows/scorecards-analysis.yml
```diff
@@ -25,7 +25,7 @@ jobs:
 
     steps:
       - name: 'Checkout code'
-        uses: actions/checkout@v3 # v2.4.0
+        uses: actions/checkout@v3.0.2 # v2.4.0
```
OrKoN: let's remove the comment here

OrKoN (approved):

PR #9075, opened 2022-10-06
fix!: remove `puppeteer.devices` in favor of `KnownDevices`

jrandolf-google:
This PR removes the deprecated `puppeteer.devices` in favor of a new exported object `KnownDevices`. `devices` can also be exported, but has been deprecated.


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1679,21 +1685,16 @@ export class Page extends EventEmitter {
    *   await browser.close();
    * })();
    * ```
-   *
-   * @remarks List of all available devices is available in the source code:
```
OrKoN, line 1683: let's still link to the full list of devices

OrKoN: ah is the link in the beginning now?

packages/puppeteer-core/src/common/Device.ts
```diff
@@ -14,23 +14,24 @@
...
-    isLandscape: boolean;
-  };
+  viewport: Viewport;
+}
+
+/**
+ * @public
+ */
+export interface KnownDevice extends Device {
+  name: string;
```
OrKoN: Let's keep the name to be on the Device interface? Then there would be no need for a breaking change and a KnownDevice type.

packages/puppeteer-core/src/api/Page.ts
jrandolf: Yes, the link is in the beginning now :)

packages/puppeteer-core/src/common/Device.ts
jrandolf: Done.

OrKoN (approved):

Kikobeats:
is this change promoted for a reason?

I'm not sure why `KnownDevices` and not `knownDevices` 🤔

PR #9078, opened 2022-10-06
fix!: refactor custom query handler API

jrandolf-google:
This PR removes the deprecated query selector API from the `puppeteer` instance. Direct imports of the query selector API have also been deprecated and users are now expected to use the static methods defined on `Puppeteer`.

### DIFF
--- a/packages/puppeteer-core/src/common/Puppeteer.ts
+++ b/packages/puppeteer-core/src/common/Puppeteer.ts
@@ -55,6 +55,56 @@ export interface ConnectOptions extends BrowserConnectOptions {
  */
 export class Puppeteer {
   /**
+   * Registers a {@link CustomQueryHandler | custom query handler}.
+   *
+   * @remarks
+   * After registration, the handler can be used everywhere where a selector is
+   * expected by prepending the selection string with `<name>/`. The name is only
+   * allowed to consist of lower- and upper case latin letters.
+   *
+   * @example
+   *
+   * ```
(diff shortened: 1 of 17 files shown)


packages/puppeteer-core/src/common/Puppeteer.ts
```diff
@@ -83,55 +133,4 @@ export class Puppeteer {
...
-
-  /**
-   * @deprecated Import directly puppeteer.
-   * @example
-   *
-   * ```ts
-   * import {registerCustomQueryHandler} from 'puppeteer';
-   * ```
-   */
-  registerCustomQueryHandler(
```
OrKoN, line 95: Would not it be better if we keep this API and make custom query handlers to be instance specific?

jrandolf: I wanted to do this, but it wouldn't make sense with the current way puppeteer is used. Instance-specific query handles would imply you can have more than one instance of puppeteer, but this is not the case; you can't even create them. Even if we did have instance-specific APIs, I assume someone will comment asking for a global API and instance-specific ones like `Puppeteer.registerGlobalCustomQueryHandler` and `puppeteer.registerCustomQueryHandler`. In the end, there is a lot of feature-creep here. I prefer we just make it a global registry.

OrKoN: Makes sense.

OrKoN (approved):

PR #9085, opened 2022-10-09
fix: Allow pptuser to access to npm cache

jecfish:
https://docs.npmjs.com/cli/v6/configuring-npm/folders

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -25,6 +25,8 @@ RUN npm i ./puppeteer-core-latest.tgz ./puppeteer-latest.tgz \
     && groupadd -r pptruser && useradd -r -g pptruser -G audio,video pptruser \
     && mkdir -p /home/pptruser/Downloads \
     && chown -R pptruser:pptruser /home/pptruser \
+    && mkdir /home/.npm \
+    && chown -R pptruser:pptruser "/home/.npm" \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -25,6 +25,8 @@ RUN npm i ./puppeteer-core-latest.tgz ./puppeteer-latest.tgz \
     && groupadd -r pptruser && useradd -r -g pptruser -G audio,video pptruser \
     && mkdir -p /home/pptruser/Downloads \
     && chown -R pptruser:pptruser /home/pptruser \
+    && mkdir /home/.npm \
```
OrKoN, line 28: why is the cache in /home/.npm and not in /hom/pptruser/.npm? The documentation seems to suggest that it should be in /home/pptruser

jecfish: I am not sure why it is not in `/home/pptruser`. 

Try running `RUN npm config get cache` in this docker return `/home/.npm`. Another alternative of the above `mkdir. chown` is to add a line to set cache location `npm config set cache "/home/pptruser/.npm"`.

jecfish:
FYI @OrKoN , close this PR because of #9095 . @jrandolf fix the sequence of `USER pptuser` and `npm install`. 

Updated Google Cloud Recorder demo accordingly: 
- https://github.com/jecfish/recorder-demo/blob/main/Dockerfile
- https://github.com/GoogleCloudPlatform/jobs-demos/pull/31

PR #9087, opened 2022-10-09
fix: remove viewport conditions in `waitForSelector`

jrandolf-google:
This PR removes the viewport conditions in `waitForSelector`.

See discussion: https://github.com/puppeteer/puppeteer/pull/8954#issuecomment-1272338883

### DIFF
--- a/packages/puppeteer-core/src/injected/util.ts
+++ b/packages/puppeteer-core/src/injected/util.ts
@@ -64,12 +64,5 @@ export const checkVisibility = (
 
 function isBoundingBoxVisible(element: Element): boolean {
   const rect = element.getBoundingClientRect();
-  return (
-    rect.width > 0 &&
-    rect.height > 0 &&
-    rect.right > 0 &&
-    rect.bottom > 0 &&
-    rect.left < self.innerWidth &&
-    rect.top < self.innerHeight
-  );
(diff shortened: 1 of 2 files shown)


joakimgunst:
Please see my comment here: https://github.com/puppeteer/puppeteer/pull/8954#issuecomment-1273051787

OrKoN (approved):

packages/puppeteer-core/src/injected/util.ts
```diff
@@ -64,12 +64,5 @@ export const checkVisibility = (
...
   const rect = element.getBoundingClientRect();
-  return (
-    rect.width > 0 &&
-    rect.height > 0 &&
-    rect.right > 0 &&
-    rect.bottom > 0 &&
-    rect.left < self.innerWidth &&
-    rect.top < self.innerHeight
-  );
+  return rect.width > 0 && rect.height > 0 && rect.right > 0 && rect.bottom > 0;
```
mmk, line 67: Shouldn't this part of the expression have also been removed?

```
rect.right > 0 && rect.bottom > 0
```

For example, `rect.bottom` is negative if the element is above the current viewport.

pasieronen: I was just writing a bug report while I saw this comment - submitted #9232 anyway so others can find this more easily :)

PR #9111, opened 2022-10-14
fix:Simplify reference in #onExecutionContextCreated

jacty:
auxData has been fetched out from contextPayload in the first line(line 377), so we can simplify the references there.

Simplify references.

Unnecessary.
Unnecessary.
**Summary**
auxData has been fetched out from contextPayload in the first line(line 377), so we can simplify the references there.

No.

**Other information**
Null.


OrKoN:
Could you please rebase if you want to land this? I will close it for now as the PR looks fairly minor and you would not want to rebase it perhaps.

PR #9112, opened 2022-10-14
fix:Remove redundante variable

jacty:
That key has been used only once in the function. It seems we could simplify this part.

Remove redundante variable

Unnecessary.
Unnecessary.
**Summary**
Remove redundante variable

No.

**Other information**


jrandolf:
Thanks for the PR, but the current code helps with readability and showing intent.

Issue #9128, opened 2022-10-18
[Bug]: Monorepo breaks build for Cloud Functions


flovouin:
### Bug description

Steps to reproduce the problem:

1. Depend on puppeteer `19.0.0` in a Cloud Function.
2. Deploy the Cloud Function, the build succeeds.
3. Invoke the Cloud Function, the error below occurs.

My guess is that the new [cache location](https://github.com/puppeteer/puppeteer/pull/9095) is not retained in the final Docker image for the Cloud Function. Although the build (and probably the download of Chromium) succeeds, Chromium is not available at runtime.

While the Cloud Function environment is not a direct concern of the puppeteer project, the caching feature introduced in `19.0.0` makes it hard to reliably install Chromium in an environment where we don't control the build/file system. I've tried playing around with `PUPPETEER_CACHE_DIR`, but again, we have very little guarantees over the build flow and it potentially changing.

Instead of specifying a download location using `PUPPETEER_CACHE_DIR`, is there any way the old behaviour can optionally be turned on, and let `puppeteer` figure out where it's installed? This would avoid having to know the subtlety of the underlying build/file system.

Thanks,

### Puppeteer version

19.0.0

### Node.js version

16

### npm version

8.5.0

### What operating system are you seeing the problem on?

Linux

### Relevant log output

```shell
Error: Could not find expected browser (chrome) locally. Run `npm install` to download the correct Chromium revision (1045629).
```

OrKoN:
The cache location is retained properly in the Docker image we provide https://github.com/puppeteer/puppeteer/pkgs/container/puppeteer why is it not retained in your case?

flovouin:
Thanks for your reply.

I'm not sure I understand. I'm not using the puppeteer-provided Docker image.

When running in the Cloud Functions environment, there is no control over the Docker build system and resulting image. There is a high probability the build process is not as linear as puppeteer's [Dockerfile](https://github.com/puppeteer/puppeteer/blob/main/docker/Dockerfile), meaning the entire `WORKDIR` / `HOME` might not make it to the final image.

As far as I know, the only guarantee we have is that dependencies are installed, hence the `node_modules` folder is available, along with the source code of the application being built. That's it. The previous caching location made it convenient to use in this setup (i.e. it worked out of the box). Also, the benefits of the new location ("better, cross-project installation of browsers") are not needed when containerising the application.

[...]

PR #9140, opened 2022-10-19
feat: use configuration files

jrandolf-google:
This PR adds configurations files to `puppeteer`'s methods for configuration. Under the hood, `puppeteer` relies on https://www.npmjs.com/package/cosmiconfig which resolves several formats of configuration:

- a `puppeteer` property in package.json
- a `.puppeteerrc` file in JSON or YAML format
- a `.puppeteerrc.json`, `.puppeteerrc.yaml`, `.puppeteerrc.yml`, `.puppeteerrc.js`, or `.puppeteerrc.cjs` file
- a `puppeteer.config.js` or `puppeteer.config.cjs` CommonJS module exporting an object

Documentation will be added later.

Fixed: #9128


packages/puppeteer-core/src/node/BrowserFetcher.ts
```diff
@@ -377,22 +381,18 @@ export class BrowserFetcher {
    * @returns A promise with a list of all revision strings (for the current `product`)
```
Lightning00Blade:
```suggestion
   * @returns A list of all revision strings (for the current `product`)
```

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -1,40 +1,28 @@
-import fs from 'fs';
+import fs, {accessSync} from 'fs';
```
Lightning00Blade: We are only using `mkdtemp` from `fs`, so I think we should change to only using named imports so `import { mkdtemp } from 'fs/promises';`

packages/puppeteer-core/src/node/ProductLauncher.ts
```diff
@@ -13,192 +13,118 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import {accessSync, existsSync} from 'fs';
-import os from 'os';
+import {existsSync} from 'fs';
+import os, {tmpdir} from 'os';
```
Lightning00Blade, line 17: Same here. We should be consistent and use only default (*) import or named no mix both of them.

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -1,40 +1,28 @@
...
   ChromeReleaseChannel,
   PuppeteerNodeLaunchOptions,
 } from './LaunchOptions.js';
-import {
-  executablePathForChannel,
-  ProductLauncher,
-  resolveExecutablePath,
-} from './ProductLauncher.js';
-import {tmpdir} from './util.js';
+import os from 'os';
```
Lightning00Blade: Nit: If everywhere else we use named import for `os` should we use it here as well?

packages/puppeteer/src/PuppeteerConfiguration.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * @internal
+ */
+export const getConfiguration = (): PuppeteerConfiguration => {
+  const result = cosmiconfigSync('puppeteer').search();
+  const configuration: PuppeteerConfiguration = result ? result.config : {};
+
+  // Merging environment variables.
+  configuration.browserRevision ??=
```
Lightning00Blade: Discussion: Using `??=` means we always take the data set in the config - which does not seem the correct overwriting mechanism. My expectation would be `ENV` -> `NPM` -> `Config`, the current implementation is   `Config` -> `ENV` -> `NPM`.

packages/puppeteer/src/PuppeteerConfiguration.ts
```diff
@@ -0,0 +1,98 @@
...
+    process.env['npm_package_config_puppeteer_tmp_dir'];
+
+  configuration.experiments ??= {};
+  configuration.experiments.macArmChromiumEnabled ??= Boolean(
+    process.env['PUPPETEER_EXPERIMENTAL_CHROMIUM_MAC_ARM'] ??
+      process.env['npm_config_puppeteer_experimental_chromium_mac_arm'] ??
+      process.env['npm_package_config_puppeteer_experimental_chromium_mac_arm']
+  );
+
+  configuration.skipDownload = Boolean(
```
Lightning00Blade: We are not gonna support `skipDownload` for config? If so we should move this to the top as the Comment implies `Merging env` and add comment like `Adding config`

packages/puppeteer/src/node/install.ts
```diff
@@ -35,58 +35,40 @@ const supportedProducts = {
...
+  let revision = configuration.browserRevision;
+
+  if (!revision) {
+    switch (product) {
+      case 'chrome':
+        revision = PUPPETEER_REVISIONS.chromium;
+        break;
+      case 'firefox':
+        revision = PUPPETEER_REVISIONS.firefox;
+        revision = await getFirefoxNightlyVersion().catch(error => {
```
Lightning00Blade: I think we should switch to using `try/catch` here.

packages/puppeteer-core/src/node/ProductLauncher.ts
jrandolf: Let's fix these in a future PR. Fixing them here just adds more noise for review. In particular, we should use a lint rule.

packages/puppeteer-core/src/node/ChromeLauncher.ts
jrandolf: See previous comment.

packages/puppeteer/src/PuppeteerConfiguration.ts
jrandolf: You're right. It should be the opposite.

packages/puppeteer/src/PuppeteerConfiguration.ts
jrandolf: I'm not sure what you mean here. We _are_ supporting `skipDownload`, does it look like we aren't?

packages/puppeteer/install.js
```diff
@@ -28,69 +28,13 @@ const path = require('path');
...
-
-  downloadBrowser();
+if (!fs.existsSync(path.join(__dirname, 'lib'))) {
+  console.log('It seems we are installing from the git repo.');
+  console.log('Building install tools from scratch...');
+  execSync('npm run build');
 }
 
-download();
+// Need to ensure TS is compiled before loading the installer
```
OrKoN: should this comment come before the `npm run build` is run above?

packages/puppeteer-core/src/node/BrowserFetcher.ts
```diff
@@ -374,25 +378,21 @@ export class BrowserFetcher {
...
    * available locally on disk.
    */
-  async localRevisions(): Promise<string[]> {
-    if (!existsSync(this.#downloadFolder)) {
+  localRevisions(): string[] {
+    if (!existsSync(this.#downloadPath)) {
       return [];
     }
-    const fileNames = await readdir(this.#downloadFolder);
+    const fileNames = readdirSync(this.#downloadPath);
```
OrKoN, line 388: let's not use the sync functions to not block the main loop while the file is read? as it's not only used in CLI but could be part of a server app.

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -93,31 +82,23 @@ export class ChromeLauncher implements ProductLauncher {
...
-    let chromeExecutable = executablePath;
-    if (channel) {
-      // executablePath is detected by channel, so it should not be specified by user.
+    let chromeExecutable: string;
+    if (this.puppeteer._isPuppeteerCore || executablePath) {
       assert(
-        !chromeExecutable,
-        '`executablePath` must not be specified when `channel` is given.'
+        executablePath,
+        `An \`executablePath\` must be specified for \`puppeteer-core\``
```
OrKoN: is it a breaking change? it should be possible to launch puppeteer-core with a chanel only.

jrandolf: It's a breaking change. It's not possible to in puppeteer core because we don't have a default path to download there anymore.

OrKoN: why not keep the path mapping there? I don't think it deserves a breaking change.

packages/puppeteer-core/src/node/BrowserFetcher.ts
jrandolf: We can't get away with async here because we use it synchronously to resolve the executablePath. Otherwise we would have to change executablePath to promises.

packages/puppeteer-core/src/node/PuppeteerNode.ts
```diff
@@ -74,28 +82,40 @@ export interface PuppeteerLaunchOptions
  * @public
  */
 export class PuppeteerNode extends Puppeteer {
-  #launcher?: ProductLauncher;
-  #productName?: Product;
+  #_launcher?: ProductLauncher;
```
OrKoN, line 85: just #launcher?

packages/puppeteer-core/src/node/BrowserFetcher.ts
OrKoN: I see that the previous version was async? do we have some new behaviour/potentially breaking change then?

packages/puppeteer-core/src/node/PuppeteerNode.ts
jrandolf: `#launcher` is used for the getter

packages/puppeteer-core/src/node/BrowserFetcher.ts
jrandolf: It's not breaking since someone can use `await` with a synchronous functions (it's a no-op). It would be breaking if it was the reverse.

test/installation/src/puppeteer-configuration.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+      );
+    },
+  },
+  ({itEvaluates}) => {
+    itEvaluates('properly', async cwd => {
+      const files = await readdir(join(cwd, '.cache', 'puppeteer'));
+      assert.equal(files.length, 1);
+      assert.equal(files[0], 'chrome');
+
+      return readAsset('puppeteer-core', 'imports.js');
```
OrKoN: can we verify that the actually launched browser is launched from the specified cache directory? it looks like the test only tests that the browser is installed to the folder and that imports work?

OrKoN (approved):

PR #9231, opened 2022-11-08
chore: use trunk

jrandolf-google:
This PR replaces linters and formatters with [Trunk](https://docs.trunk.io/docs)'s unified API.

This greatly improves the speed of formatting and linting and also consolidates linter/formatter configuration to one area (the `.trunk` folder).


OrKoN (changes requested):
Looks like it's failing in CI? I don't think the speed is critical for us currently and I think we should not add a new non-proven tool to troubleshoot at the moment. Let's revisit this decision later.

dapirian:
Hey there 👋 - creator of [trunk](https://docs.trunk.io/docs/check) here. We'd be very happy to help support Trunk in puppeteer (and we're puppeteer users ourselves), just let me know how I can help.

The github action isn't failing (it passed), but we'll get an update out to resolve the error message.

Thanks!

Lightning00Blade:
Closing this one as we have no intent to integrate it at this time.

PR #9240, opened 2022-11-10
chore: export all classes

OrKoN:
Removes automatic generation of types.


docs/api/puppeteer.puppeteer.md
```diff
@@ -1,29 +1,11 @@
 ---
-sidebar_label: Puppeteer
+sidebar_label: puppeteer
 ---
 
-# Puppeteer class
```
OrKoN, line 5: Hm, I wonder why this part was removed?

PR #9241, opened 2022-11-10
chore: re-export all exports

OrKoN:
Adds index files to subfolders and removes generation of types.ts in puppeteer-core. This makes our types to match what we actually export.


jrandolf-google (approved):

Juneezee:
This breaks existing TypeScript code like this after updating to `v19.3.0`:

```typescript

let browser: puppeteer.Browser;
try {
    headless: true,
  });
} catch (e) {
  if (e instanceof Error) {
    console.error(e.message);
  }
}
```
```
error TS2503: Cannot find namespace 'puppeteer'.

56   let browser: puppeteer.Browser;
```

---

I managed to fix the error by importing puppeteer as namespace:
```typescript
```

OrKoN:
@Juneezee which TypeScript version do you use? Our tests for this still work: https://github.com/puppeteer/puppeteer/blob/main/test-d/puppeteer.test-d.ts

Juneezee:
> @Juneezee which TypeScript version do you use? Our tests for this still work: https://github.com/puppeteer/puppeteer/blob/main/test-d/puppeteer.test-d.ts

@OrKoN I tested on both TypeScript 4.8.4 and 4.9.3 

```typescript

let browser: puppeteer.Browser;
let page: puppeteer.Page;
```

|                  | TypeScript 4.8.4 | TypeScript 4.9.3 |
|------------------|:----------------:|:----------------:|
| Puppeteer 19.2.0 |         ✔️        |         ✔️        |
| Puppeteer 19.3.0 |         ❌        |         ❌        |

On Puppeteer 19.3.0, it gives `TS2503: Cannot find namespace 'puppeteer'.`

PR #9249, opened 2022-11-11
fix: remove boundary conditions for visibility

### LINKED ISSUE #9232: [Bug]: waitForSelector doesn't consider elements above current viewport as "visible"
### Bug description

#8954 changed waitForSelector "visible" flag behavior to consider only elements inside the current viewport as visible. This was reverted in #9087, but it looks like the revert wasn't complete. Apparently elements below the current viewport are now considered "visible", but not elements above (or left) of the current viewport.

Simple script to reproduce (works with 17.1.3, fails with 19.2.2):

```js

(async () => {
  await page.setViewport({width: 800, height: 600});
  await page.setContent(`
    <html><body>
    <button id="button1">Button 1</button>
    <div id="content" style="height: 2000px;"></div>
    <button id="button2">Button 2</button>
    </body></html>
  `);
  await page.waitForSelector('#button1', {visible: true});
  await page.waitForSelector('#button2', {visible: true});
  console.log('ok, both buttons are visible');
  await page.evaluate(() => window.button2.scrollIntoView());
  await page.waitForSelector('#button1', {visible: true});
  console.log('ok, button1 still visible (although outside viewport)');
  await browser.close();
})();
```

### Puppeteer version

19.2.2

### Node.js version

16.18.0

### npm version

8.19.2

macOS

### Configuration file

_No response_

### Relevant log output

_No response_

jrandolf-google:
Fixed: https://github.com/puppeteer/puppeteer/issues/9232

### DIFF
--- a/packages/puppeteer-core/src/injected/util.ts
+++ b/packages/puppeteer-core/src/injected/util.ts
@@ -58,11 +58,11 @@ export const checkVisibility = (
   const isVisible =
     style &&
     !HIDDEN_VISIBILITY_VALUES.includes(style.visibility) &&
-    isBoundingBoxVisible(element);
+    !isBoundingBoxEmpty(element);
   return visible === isVisible ? node : false;
 };
 
-function isBoundingBoxVisible(element: Element): boolean {
+function isBoundingBoxEmpty(element: Element): boolean {
   const rect = element.getBoundingClientRect();
-  return rect.width > 0 && rect.height > 0 && rect.right > 0 && rect.bottom > 0;
+  return rect.width === 0 || rect.height === 0;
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

packages/puppeteer-core/src/injected/util.ts
```diff
@@ -64,5 +64,5 @@ export const checkVisibility = (
 
 function isBoundingBoxVisible(element: Element): boolean {
```
OrKoN: nit: rename isBoundingBoxVisible to something that does not have word Visible in it?

PR #9299, opened 2022-11-22
feat: add Page.waitForDevicePrompt

mzgoddard:

Add a feature through `Page.waitForDevicePrompt()` and `DeviceRequestPrompt` to respond to `DeviceAccess.deviceRequestPrompted` protocol events.

This change adds tests for the interface that simulate interaction with a mock session object.

These tests cover the added code except for `Page.waitForDevicePrompt` itself.

`Page.waitForDevicePrompt` includes a jsdoc with a comment with a likely common use case:

```js
const [deviceRequest] = Promise.all([
  page.waitForDevicePrompt(),
  page.click('#connect-bluetooth'),
]);
await devicePrompt.select(
  await devicePrompt.waitForDevice(({name}) => name.includes('My Device'))
);
```

**Summary**

`Page.waitForDevicePrompt` resolves to a `DeviceRequestPrompt` when the next `DeviceAccess.deviceRequestPrompted` events are emitted by the session. `DeviceRequestPrompt.select` and `DeviceRequestPrompt.cancel` can be be called to respond to the prompt. `DeviceRequestPrompt.waitForDevice` resolves to the first device in the prompt that matches a given filter. That device can be listed in `DeviceRequestPrompt.devices` when `waitForDevice` is called or from future `DeviceAccess.deviceRequestPrompted` events.

This wraps devtools protocol events and methods added in 
https://chromium-review.googlesource.com/c/chromium/src/+/3933844. That commit log integrates those events with WebBluetooth. When WebBluetooth's requestDevice is called, after some steps are performed, if they succeed, a prompt for the user is opened to select a device. This interface and the protocol additions it depends on add a way to automate responding to that prompt.


mzgoddard:
To further test this interface with the devtools protocol additions, without mocking them, a test environment with a specific hardware setup like bluetooth hardware with nearby reliable devices or some degree of detailed virtualization. Chromium uses Mojo services to fake parts of its WebBluetooth implementation in tests. Tests here could use that as well but these features are not normally shipped, to my knowledge, because the size of the released build they add. But I am thinking that would be "too much" here to automate testting this code, so I've used a mock session object here to test it.

The commit log https://chromium-review.googlesource.com/c/chromium/src/+/3933844 isn't merged yet, so I used a npm-linked version of devtools-protocol I've pushed to https://github.com/mzgoddard/devtools-protocol/tree/device-access-device-request-prompted while writing this change and testing it.

OrKoN:
@mzgoddard thanks for the PR. Would you mind rebasing?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -2557,6 +2558,36 @@ export class Page extends EventEmitter {
...
+   * const [deviceRequest] = Promise.all([
+   *   page.waitForDevicePrompt(),
+   *   page.click('#connect-bluetooth'),
+   * ]);
+   * await devicePrompt.select(
+   *   await devicePrompt.waitForDevice(({name}) => name.includes('My Device'))
+   * );
+   * ```
+   */
+  waitForDevicePrompt(
```
OrKoN, line 2586: Could you please run `npm run docs` to re-generate API docs?

packages/puppeteer-core/src/common/Page.ts
```diff
@@ -3113,6 +3122,35 @@ export class CDPPage extends Page {
...
+   * const [deviceRequest] = Promise.all([
+   *   page.waitForDevicePrompt(),
+   *   page.click('#connect-bluetooth'),
+   * ]);
+   * await devicePrompt.select(
+   *   await devicePrompt.waitForDevice(({name}) => name.includes('My Device'))
+   * );
+   * ```
+   */
+  override waitForDevicePrompt(
```
OrKoN, line 1675: can iframes request bluetooth access? Perhaps it makes sense to create deviceRequestPromptManager per frame (or rather per OOP-iframe) so that this method becomes `this.mainFrame().waitForDevicePrompt()`.

packages/puppeteer-core/src/common/Frame.ts
```diff
@@ -216,6 +221,10 @@ export class Frame {
     this._id = frameId;
     this._parentId = parentFrameId;
     this.#detached = false;
+    this.#deviceRequestPromptManager = new DeviceRequestPromptManager(
+      client,
```
OrKoN: for frames the CDPSession might equal the parent's CDPSession (if the iframe is in the same process and does not have a separate debugging target). I think we should probably lazily instantiate DeviceRequestPromptManager based on the isOOPIF getter and if the frame is not an OOPIF, we should use the parent's DeviceRequestPromptManager. Perhaps you have other suggestions?

mzgoddard: Hmm, does something like this seem good?

```ts
  /**
   * @internal
   */
  _deviceRequestPromptManager(): DeviceRequestPromptManager {
    if (this.#deviceRequestPromptManager === undefined) {
      if (this.isOOPFrame()) {
        this.#deviceRequestPromptManager = new DeviceRequestPromptManager(
          this.#client,
          this._frameManager.timeoutSettings
        );
      } else {
        this.#deviceRequestPromptManager =
          this.parentFrame()?._deviceRequestPromptManager();
      }
    }
    return this.#deviceRequestPromptManager!;
  }

  waitForDevicePrompt(
    options: WaitTimeoutOptions = {}
  ): Promise<DeviceRequestPrompt> {
    return this._deviceRequestPromptManager().waitForDevicePrompt(options);
  }
```

I don't have any particular suggestions. I may be overlooking things about OOPIF and feature-policy controlled systems like Web Bluetooth (that the CL integrates with this devtool domain).

OrKoN: This would not work when an OOPIF goes back to be a local frame. We probably should be able to create a weak map to track DeviceRequestPromptManager per client (e.g., `WeakMap<CDPSession, DeviceRequestPromptManager> map` and call `map.get(this.#client)` to get the current device request manager.

mzgoddard: I added a change like the above snippet with your point on using a `WeakMap.` https://github.com/puppeteer/puppeteer/pull/9299/commits/f92c95e755bb0927bc0833d4054a53dcfdbc90f3

Does that seem right? I'm not sure where the `WeakMap` would be best kept. I've added it to `FrameManager` in that commit.

OrKoN:
@mzgoddard yes, that's right. We can land it once we rolled the Chromium binary version that supports this feature.

packages/puppeteer-core/src/common/Frame.ts
```diff
@@ -1083,6 +1087,45 @@ export class Frame {
...
   }
 
+  /**
+   * @internal
+   */
+  _deviceRequestPromptManager(): DeviceRequestPromptManager {
+    if (this.isOOPFrame()) {
+      return this._frameManager._deviceRequestPromptManager(this.#client);
+    }
+    return this.parentFrame()!._deviceRequestPromptManager();
```
OrKoN: let's remove `!` here: I guess the main frame does not have a parent, so we need to check for that?

packages/puppeteer-core/src/common/Frame.ts
OrKoN: Looks good to me, but I think we need to unset the reference to the CDPSession instance somewhere in the DeviceRequestPrompt when the session gets detached. Currently, it looks like there could be a memory leak and prompts will keep the detached sessions alive. Could you check that case?

mzgoddard: Are you meaning a potential leak with a WeakMap value referencing its key? I don't think they work that way. I'd have to think about the [spec](https://tc39.es/ecma262/multipage/executable-code-and-execution-contexts.html#sec-liveness) a bit to remember why that is.

Or do you mean a leak if a user of puppeteer retains a reference to DeviceRequestPrompt directly? That'd make sense.

I've added listeners on `Target.detachedFromTarget` to unset client in `DeviceRequestPrompt` and `DeviceRequestPromptManager` and `assert` calls to ensure `#client` isn't null when its used.

Is the `Target.detachedFromTarget` event enough?

packages/puppeteer-core/src/common/Frame.ts
mzgoddard: Good point. I had been thinking if we `isOOPFrame` is false there must be a `parentFrame` but I guess that is ovely dependent on the current implementation of `isOOPFrame`.

packages/puppeteer-core/src/common/Frame.ts
OrKoN: I think it should be enough, thanks for double checking.

packages/puppeteer/src/types.ts
```diff
@@ -0,0 +1,77 @@
+// AUTOGENERATED - Use `npm run generate:sources` to regenerate.
```
OrKoN: This should not be committed. I think the file has been removed from the tree but accidentally pushed here.

OrKoN:
@mzgoddard let me know if need help rebasing this. We can land it now.

OrKoN:
@mzgoddard could you please run `npm run format` to update the formatting based on the latest rules?

OrKoN (approved):

test/src/DeviceRequestPrompt.spec.ts
```diff
@@ -0,0 +1,457 @@
...
+      const client = new MockCDPSession();
+      const timeoutSettings = new TimeoutSettings();
+      const manager = new DeviceRequestPromptManager(client, timeoutSettings);
+
+      client.emit('DeviceAccess.deviceRequestPrompted', {
+        id: '00000000000000000000000000000000',
+        devices: [],
+      });
+
+      expect(manager).toBeTruthy();
```
kblok, line 121: @mzgoddard, sorry for quite a late comment. But what's the purpose of this `expect`?
How the `manager` won't be truthy?

PR #9306, opened 2022-11-23
chore(ng-schematics): Spawn server when running `ng e2e` 

Lightning00Blade:

Spawn own server when running `ng e2e`. Give user option to not replace `ng e2e`.

Yes.

Yes, `ng-schematics` README.md updated.

**Summary**

When running `ng-schematics`'s `ng e2e` command spawns it's own server. This way we remove the need of developers to run `ng server` separately thus increasing ease of use in development and CI. 

We want to support Protractor migration so we give the user the option to opt out of replacing `ng e2e` so they can have a gradual migration. 
(Note: There may be issues with folder conflicts, to be address in a PR for adding better Migration support)

Yes, as we don't check if required options are there before spawning the server.

**Other information**

### DIFF
--- a/packages/ng-schematics/src/builders/puppeteer/schema.json
+++ b/packages/ng-schematics/src/builders/puppeteer/schema.json
@@ -15,7 +15,7 @@
     },
     "devServerTarget": {
       "type": "string",
-      "description": "Commands to execute in the repo. Commands prefixed with `./node_modules/bin` (Exception: 'node')."
+      "description": "Angular target that spawn the server."
     }
   },
   "additionalProperties": true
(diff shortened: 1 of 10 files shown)


packages/ng-schematics/src/builders/puppeteer/schema.json
```diff
@@ -12,6 +12,10 @@
         }
       },
       "description": "Commands to execute in the repo. Commands prefixed with `./node_modules/bin` (Exception: 'node')."
+    },
+    "devServerTarget": {
+      "type": "string",
+      "description": "Angular target that spawn the server."
```
OrKoN:
```suggestion
      "description": "Angular target that spawns the server."
```

OrKoN (approved):

PR #9314, opened 2022-11-23
feat: ability to send headers via ws connection to browser in node.js environment

### LINKED ISSUE #7218: Feature Request: Connect Options - Allow Authorization Header
### Feature Request

I'd like to connect to a `browserWSEndpoint` that is protected by authentication. It would be nice to be able to provide an `Authorization Bearer <token>` on the connection, and every request thereafter to that endpoint.

I'm specifically wanting the Authorization Header, but this probably could just be generic to allowing any headers on all requests.

I vision within the `connect` options, provide a `headers` option
https://github.com/puppeteer/puppeteer/blob/v9.1.1/docs/api.md#puppeteerconnectoptions

Example:
```
browser = await puppeteer.connect({
      browserWSEndpoint: "ws://my.server",
      headers: {
        Authorization: "Bearer hunter2"
      },
});
```

I'd imagine the change would happen here:
https://github.com/puppeteer/puppeteer/blob/943477cc1eb4b129870142873b3554737d5ef252/src/common/BrowserWebSocketTransport.ts#L21

To be something similar to:
https://github.com/websockets/ws/issues/473#issuecomment-240074672

I did see `page.setExtraHTTPHeaders(headers)`, but I don't think this what I'm looking for as it shouldn't just be for a specific page, but for all websocket requests to the server.
https://github.com/puppeteer/puppeteer/blob/v9.1.1/docs/api.md#pagesetextrahttpheadersheaders

DudaGod:

I have browsers pool in some cloud. I want that only users with access will be able to connect to them. So they must provide token through headers. But puppeteer does not allow to send headers when connected to browser by ws connection. So I added this feature.

Closes #7218

### DIFF
--- a/docs/api/puppeteer.connectoptions.headers.md
+++ b/docs/api/puppeteer.connectoptions.headers.md
@@ -1,0 +1,19 @@
+---
+sidebar_label: ConnectOptions.headers
+---
+
+# ConnectOptions.headers property
+
+Headers to use for the web socket connection.
+
+#### Signature:
+
+```typescript
+interface ConnectOptions {
+  headers?: Record<string, string>;
+}
+```
+
+## Remarks
+
+Only works in the Node.js environment.
--- a/packages/puppeteer-core/src/common/BrowserConnector.ts
+++ b/packages/puppeteer-core/src/common/BrowserConnector.ts
@@ -24,6 +24,8 @@ import {Connection} from './Connection.js';
 import {ConnectionTransport} from './ConnectionTransport.js';
 import {getFetch} from './fetch.js';
 import {Viewport} from './PuppeteerViewport.js';
+
+import type {ConnectOptions} from './Puppeteer.js';
 /**
  * Generic browser options that can be passed when launching any browser or when
  * connecting to an existing browser instance.
@@ -73,11 +75,7 @@ const getWebSocketTransportClass = async () => {
  * @internal
  */
 export async function _connectToCDPBrowser(
-  options: BrowserConnectOptions & {
-    browserWSEndpoint?: string;
-    browserURL?: string;
-    transport?: ConnectionTransport;
-  }
+  options: BrowserConnectOptions & ConnectOptions
 ): Promise<CDPBrowser> {
   const {
     browserWSEndpoint,
@@ -85,6 +83,7 @@ export async function _connectToCDPBrowser(
     ignoreHTTPSErrors = false,
     defaultViewport = {width: 800, height: 600},
     transport,
+    headers = {},
     slowMo = 0,
     targetFilter,
     _isPageTarget: isPageTarget,
@@ -102,7 +101,7 @@ export async function _connectToCDPBrowser(
   } else if (browserWSEndpoint) {
     const WebSocketClass = await getWebSocketTransportClass();
     const connectionTransport: ConnectionTransport =
-      await WebSocketClass.create(browserWSEndpoint);
+      await WebSocketClass.create(browserWSEndpoint, headers);
     connection = new Connection(browserWSEndpoint, connectionTransport, slowMo);
   } else if (browserURL) {
     const connectionURL = await getWSEndpoint(browserURL);
--- a/packages/puppeteer-core/src/common/Puppeteer.ts
+++ b/packages/puppeteer-core/src/common/Puppeteer.ts
@@ -42,6 +42,12 @@ export interface ConnectOptions extends BrowserConnectOptions {
   browserWSEndpoint?: string;
   browserURL?: string;
   transport?: ConnectionTransport;
+  /**
+   * Headers to use for the web socket connection.
+   * @remarks
+   * Only works in the Node.js environment.
+   */
+  headers?: Record<string, string>;
 }
 
 /**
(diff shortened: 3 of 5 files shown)


packages/puppeteer-core/src/common/BrowserConnector.ts
```diff
@@ -102,7 +101,7 @@ export async function _connectToCDPBrowser(
   } else if (browserWSEndpoint) {
     const WebSocketClass = await getWebSocketTransportClass();
     const connectionTransport: ConnectionTransport =
-      await WebSocketClass.create(browserWSEndpoint);
+      await WebSocketClass.create(browserWSEndpoint, headers);
```
OrKoN, line 104: how difficult it is to provide a custom transport instead (already supported feature)? I guess there could be potentially many transport options and I am not sure we should include all of them in the connect options. Although headers is probably one of the most common options and might be worth including.

DudaGod: > how difficult it is to provide a custom transport instead (already supported feature)?

It is not difficult because for this all I have to do is copy code from `NodeWebSocketTransport` - https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/NodeWebSocketTransport.ts and add some `headers`. But it seems that such a need may arise not only for me. So it might be worth supporting it out of the box.

OrKoN: Looks like it have been requested before, at least I found this one https://github.com/puppeteer/puppeteer/issues/7218. Do you know of any other open or closed issues?

OrKoN, line 104: I think headers have to be forwarded in the `if (browserURL)` branch as well

OrKoN (changes requested):
In general, LGTM but let's to the following:

1) see if we can add a test
2) run `npm run docs` to regenerate documentation
3) use headers when the transport is created from a browserURL

packages/puppeteer-core/src/common/Puppeteer.ts
```diff
@@ -42,6 +42,7 @@ export interface ConnectOptions extends BrowserConnectOptions {
   browserWSEndpoint?: string;
   browserURL?: string;
   transport?: ConnectionTransport;
+  headers?: Record<string, string>;
```
OrKoN, line 50: let's add a doc comment that headers are used for the web socket connection

packages/puppeteer-core/src/common/BrowserConnector.ts
DudaGod: > Do you know of any other open or closed issues?

No, I don't know any more

DudaGod: But how it will be worked? As far as I know websocket implementation in browser - https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/WebSocket (which is used here - https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/BrowserWebSocketTransport.ts) doesn't work with headers.

docs/api/puppeteer.connectoptions.headers.md
```diff
@@ -0,0 +1,13 @@
...
+
+# ConnectOptions.headers property
+
+#### Signature:
+
+```typescript
+interface ConnectOptions {
+  headers?: Record<string, string>;
+}
+```
```
DudaGod, line 15: Tried to add here something like:
```
## Remarks

**NOTE** Works only in Node.js environment.
```

But `npm run docs` rewrite my changes.

DudaGod:
> see if we can add a test

Before send a PR I tried to do it. But I didn't find a quick way to check that the appropriate headers were sent when connecting to the browser by websocket. Moreover, I did not find any similar tests. I can dig deeper if without tests we can't merge it.

docs/api/puppeteer.connectoptions.headers.md
OrKoN: The docs should go into code comments and our build re-generated markdown based on code comments. E.g.,

```
/**
* Headers to use for the web socket connection.
* @remarks
* Only works in the Node.js environment.
*/
headers?: Record<string, string>;
```

packages/puppeteer-core/src/common/BrowserConnector.ts
OrKoN: yes, you are right!

OrKoN: Could you please add a line `Closes #7218` to the end of your commit message/PR description?

DudaGod: Done.

OrKoN (approved):

DudaGod:
Thank you for merge it. Can you tell me when the new version of puppeteer-core will be released?

DudaGod:
oh, I see that new PR with bump version was created - https://github.com/puppeteer/puppeteer/pull/9322. So it looks like that new version will be released today.

OrKoN:
@DudaGod it will take a week or two before release as try to batch multiple changes in one release. You can subscribe to https://github.com/puppeteer/puppeteer/pull/9322 to get notified when a release happens.

DudaGod:
@OrKoN hi. Is there any new information about the release date? Thank you.

OrKoN:
@DudaGod probably in the next few days

Issue #9280, opened 2022-11-16
[Bug]: No proper types for `$eval` and `$$eval`


max-programming:
### Bug description

In TypeScript, if I try to use the `$eval` or `$$eval` method, it does not let me type the element correctly.

I can't type it explicitly and I also can't pass it in a generic.

Different Situations:

> When no types are used
![image](https://user-images.githubusercontent.com/51731966/202220713-ef254830-326b-47c5-af6c-8853ea8c5c1d.png)

> When a generic is used
![image](https://user-images.githubusercontent.com/51731966/202220809-309c38e4-5abb-4af4-95e9-5c29251b4be3.png)

> When explicit types are used
![image](https://user-images.githubusercontent.com/51731966/202220912-8027396d-7d5f-419e-8c39-2208900e3a48.png)

### Puppeteer version

19.2.2

### Node.js version

16.17.1

### npm version

8.15.0

### What operating system are you seeing the problem on?

Windows

OrKoN:
I believe the docs got outdated here: the recommended way is to cast/typecheck the element: e.g., `el => (el as HTMLInputElement).value` or `if (el instanceof HTMLInputElement) { // action } else { //error }`. In some cases Puppeteer can infer the type based on the selector but not in this case. The generic approach was less safe and we moved away from it.

max-programming:
@OrKoN The `as` and `instanceof` definitely works. But is not efficient.

It adds up to more extra code that could be fixed with explicit typings.

OrKoN:
Documented in https://pptr.dev/guides/query-selectors#css

PR #9352, opened 2022-12-01
feat: add element validation

### LINKED ISSUE #8579: [Discussion]: New $ and $$ type inference makes code harder to read
### Bug description

v15 tries to [infer types](https://github.com/puppeteer/puppeteer/pull/8547) for simple selectors like `'a'` or `'tr'`. It removed the ability to provide explicit types via generics and requires casting for non-trivial selectors.

I would argue that most code doesn't use trivial selectors and provides a nested path or classes instead. Developers often know the return type already and the new inference doesn't allow them to provide these types.

```typescript
// v14
await page.$<HTMLAnchorElement>('a')
await page.$<HTMLAnchorElement>('.class > a')

// v15
await page.$('a')
(await page.$('.class > a')) as ElementHandle<HTMLAnchorElement> | null
```

Simply put, while the simple cases are now shorter, the more complex cases are now longer/harder to read.

I propose that puppeteer reinstates the ability to specify a generic type *in addition to* the current smart inference.

### Puppeteer version

v15.1.1

### Node.js version

v16.15.1

### npm version

Yarn 3.2.1

macOS

### Relevant log output

_No response_

jrandolf-google:
This PR adds a method to ElementHandle that validates the tag type of that handle and returns it.

Fixed: #8579, #9280


OrKoN (changes requested):
please add tests

packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -412,6 +421,61 @@ export class ElementHandle<
...
+   * } catch (err) {
+   *   // You need to handle the trash since you specified `dispose: false`
+   *   element.dispose();
+   *   throw new err();
+   * }
+   * ```
+   *
+   * @param tagName - The tag name of the desired element type.
+   * @param options - Configures the behavior of this method.
+   * @throws An error if the handle does not match. **This will dispose the
```
OrKoN: why do we need to disable disposal of the handle that we don't return to the user?

jrandolf: The `dispose` option is up to the user.

jrandolf:
> please add tests

Done.

packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -412,6 +421,61 @@ export class ElementHandle<
...
+    tagName: K,
+    options: ToElementOptions = {}
+  ): Promise<HandleFor<ElementFor<K>>> {
+    const {dispose = true} = options;
+    const isMatchingTagName = await this.evaluate((node, tagName) => {
+      return node.nodeName === tagName.toUpperCase();
+    }, tagName);
+    if (!isMatchingTagName) {
+      if (dispose) {
+        await this.dispose();
```
OrKoN: I would prefer the method to have a single responsibility: let's return null if the check fail and allow the client to dispose or not-dispose outside of this method. Alternatively, we can have a second method that is more strict and indicates in its name that the handle might be disposed (possible names `toElementOrDispose`).

P.S. another alternative is to rename this to `toElementOrDispose`.

OrKoN: would it work for custom elements?

packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -412,6 +421,61 @@ export class ElementHandle<
...
+      }
+      throw new Error(`Element is not a(n) \`${tagName}\` element`);
+    }
+    return this as unknown as HandleFor<ElementFor<K>>;
+  }
+
   override asElement(): ElementHandle<ElementType> | null {
```
OrKoN, line 446: could we replace existing usages of `asElement` in our codebase if it gets deprecated?

packages/puppeteer-core/src/common/JSHandle.ts
```diff
@@ -229,6 +229,8 @@ export class JSHandle<T = unknown> {
   }
 
   /**
+   * @deprecated Use `instanceof ElementHandle` instead of this method.
```
OrKoN: does `instanceof ElementHandle` check maintain the ElementType generic? I think it should but not sure.

packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -412,6 +421,61 @@ export class ElementHandle<
...
+   *   throw new err();
+   * }
+   * ```
+   *
+   * @param tagName - The tag name of the desired element type.
+   * @param options - Configures the behavior of this method.
+   * @throws An error if the handle does not match. **This will dispose the
+   * current handle.** You can disable this behavior in the options.
+   */
+  async toElement<
```
OrKoN, line 434: Is there a bug/issue requesting this? where would we use it? can it be internal?

packages/puppeteer-core/src/common/ElementHandle.ts
jrandolf: We can do this when we remove `asElement` completely. It's not that important right now.

packages/puppeteer-core/src/common/ElementHandle.ts
jrandolf: This is something I wanted internally after trying to upstream a new puppeteer version. There are a lot of errors that can be fixed by `as ...`, but this is prone to misuse and not descriptive if something does break. This is also a solution to:
 - https://github.com/puppeteer/puppeteer/issues/8579
 - https://github.com/puppeteer/puppeteer/issues/9280

packages/puppeteer-core/src/common/ElementHandle.ts
jrandolf: > would it work for custom elements?

Yes. It also kinda works with `class Test extends HTMLDivElement` because these type of custom elements still use `DIV` as their node name instead of the custom element's name. Of course, this will deduce to `HTMLDivElement`, not `Test`.

packages/puppeteer-core/src/common/JSHandle.ts
jrandolf: I tested this, and it seems doing an `instanceof` check doesn't perform narrowing... So I am undeprecating this.

packages/puppeteer-core/src/common/ElementHandle.ts
jrandolf: > let's return null if the check fail and allow the client to dispose or not-dispose outside of this method.

The reason I want to throw is because there is an expectation that the element *will* be the correct. We could write `toElementOrThrow` though. For disposal, I would still like to keep the `dispose` option.

OrKoN:
I think the PR description needs an update?

OrKoN (approved):
Let's also update https://pptr.dev/guides/query-selectors and mention this and the different to the other method.

jrandolf:
> Let's also update https://pptr.dev/guides/query-selectors and mention this and the different to the other method.

I'll do this in another PR.

PR #9364, opened 2022-12-06
feat(chromium): roll to Chromium 109.0.5412.0 (r1069273)

### LINKED ISSUE #9233: M109 roll
M109 is going to beta on Thu, Dec 1, 2022 which is around the time we need to try rolling a new version.
TODOs:

1) Update the chrome headless name https://crbug.com/1380881 to reflect the upstream change

cc @jrandolf @Lightning00Blade

Lightning00Blade:
Closes #9233


versions.js
```diff
@@ -53,7 +54,7 @@ const versionsPerRelease = new Map([
 ]);
 
 // The same major version as the current Chrome Stable per https://chromestatus.com/roadmap.
-const lastMaintainedChromiumVersion = '106.0.5249.0';
+const lastMaintainedChromiumVersion = '108.0.5351.0';
```
OrKoN: let's specify 107.0.5296.0 as 107.0.5296.0 is relatively new.

versions.js
```diff
@@ -53,7 +54,7 @@ const versionsPerRelease = new Map([
 ]);
 
 // The same major version as the current Chrome Stable per https://chromestatus.com/roadmap.
```
OrKoN: probably will need a better guideline here, perhaps last 3 major versions?

Lightning00Blade: How about `Should not be more then 2 major versions behind Chrome Stable (link)` ?

OrKoN: sg

OrKoN (approved):

test/src/navigation.spec.ts
```diff
@@ -208,7 +195,7 @@ describe('navigation', function () {
         return (error = error_);
       });
       if (isChrome) {
-        expect(error.message).toContain(getExpectedSSLCertMessage());
+        expect(error.message).toContain(EXPECTED_SSL_CERT_MESSAGE_REGEX);
```
OrKoN: toMatch?

PR #9368, opened 2022-12-06
chore: fix race in test Frame.waitForFunction should work when context is destroyed

whimboo:
Fixes the race condition which causes intermittent failures in Firefox because we haven't implemented bootstrap scripts to run on document creation.


whimboo:
@OrKoN would you mind to review? Thanks!

OrKoN (approved):

PR #9370, opened 2022-12-06
fix(puppeteer-core): avoid type instantiation errors

### LINKED ISSUE #9369: [Bug]: Type error: Type instantiation is excessively deep and possibly infinite.
### Bug description

1. Query a complex selector in a Typescript puppeteer test script
2. Typescript complains that resolving the return type takes too many steps: `Type error: Type instantiation is excessively deep and possibly infinite.`

### Puppeteer version

19.3.0

### Node.js version

16.18.1

### npm version

yarn 1.22.19

Linux

### Configuration file

_No response_

### Relevant log output

```Shell
Type error: Type instantiation is excessively deep and possibly infinite.

  70 | 
> 71 |     const rows = await page.$$( "[data-testid=\"my-component\"] tbody tr" )
     |                        ^
```

OrKoN:
Using the accumulator allows enabling the tail-recursion optimization in the TypeScript compiler.

Closes #9369

### DIFF
--- a/packages/puppeteer-core/src/common/types.ts
+++ b/packages/puppeteer-core/src/common/types.ts
@@ -135,32 +135,36 @@ type BeginSubclassSelectorTokens = ['.', '#', '[', ':'];
 
 type CombinatorTokens = [' ', '>', '+', '~', '|', '|'];
 
-type Drop<Arr extends readonly unknown[], Remove> = Arr extends [
-  infer Head,
-  ...infer Tail
-]
+type Drop<
+  Arr extends readonly unknown[],
+  Remove,
+  Acc extends unknown[] = []
+> = Arr extends [infer Head, ...infer Tail]
   ? Head extends Remove
     ? Drop<Tail, Remove>
-    : [Head, ...Drop<Tail, Remove>]
-  : [];
+    : Drop<Tail, Remove, [...Acc, Head]>
+  : Acc;
 
 type FlatmapSplitWithDelemiters<
   Inputs extends readonly string[],
-  Delemiters extends readonly string[]
+  Delemiters extends readonly string[],
+  Acc extends string[] = []
 > = Inputs extends [infer FirstInput, ...infer RestInputs]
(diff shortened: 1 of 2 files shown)


OrKoN:
cc @gomain

Lightning00Blade (approved):

packages/puppeteer-core/src/common/types.ts
```diff
@@ -135,32 +135,36 @@ type BeginSubclassSelectorTokens = ['.', '#', '[', ':'];
...
-      : never
-    : never
-  : [];
+      ? FlatmapSplitWithDelemiters<
+          RestInputs,
+          Delemiters,
+          [...Acc, ...SplitWithDelemiters<FirstInput, Delemiters>]
+        >
+      : Acc
+    : Acc
```
gomain, line 161: These two branches are `never` taken.

gomain (commented):
A bit late to the party, but here goes...

Spotted a minor imperfection, where a couple evaluation branches should be marked as `never`.

The `Acc` trick does not allow tail recursion optimization (I believe there is no such optimization). It does allow intermediate thunks to be evaluated before the recursive call. This hugely reduces the (virtual) stack size. There still exists a limit on how many tokens the input string can contain before instantiation depth exceeds the limit. From [trial and error investigation](https://github.com/gomain/puppeteer/blob/d3326ed2aef912b9d6ee7cd4e9355acc326a2d34/packages/puppeteer-core/src/common/types.ts#L172) it seems the recursion depth limit is about 1000 (typescript v4.5.4).

Practically, a selector string with multiple compound selectors with explicit combinators (i.e. not space), such as `"div > div > div > ..."` would be limited to ~500 compound selectors. We could probably live with that.

gomain:
@OrKoN https://github.com/microsoft/TypeScript/pull/45711/files#diff-d9ab6589e714c71e657f601cf30ff51dfc607fc98419bf72e04f6b0fa92cc4b8R15233-R15237

PR #9375, opened 2022-12-07
fix: remove oopif expectations and fix oopif flakiness

OrKoN:
With M109 the flakiness should be reduced. Any present flakiness should be investigated.

Drive-by: a new debugging helper to debug on CI.


packages/puppeteer-core/src/common/Page.ts
```diff
@@ -262,9 +262,6 @@ export class CDPPage extends Page {
 
   #onDetachedFromTarget = (target: Target) => {
     const sessionId = target._session()?.id();
-
-    this.#frameManager.onDetachedFromTarget(target);
```
OrKoN, line 266: This should now be redundant because we always get `frameDetached` event.

test/TestExpectations.json
```diff
@@ -2128,5 +2044,71 @@
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["firefox"],
     "expectations": ["PASS", "FAIL"]
+  },
+  {
```
OrKoN, line 2048: I have added some common failures that have been around for a while now.

test/TestExpectations.json
```diff
@@ -2128,5 +2044,71 @@
...
+    "parameters": ["chrome", "headless"],
+    "expectations": ["PASS", "FAIL"]
+  },
+  {
+    "testIdPattern": "[tracing.spec] Tracing \"after each\" hook for \"should output a trace\"",
+    "platforms": ["win32"],
+    "parameters": ["chrome", "headless"],
+    "expectations": ["PASS", "FAIL"]
+  },
+  {
```
OrKoN, line 2066: BiDi expectation should always be last.

Lightning00Blade (approved):

PR #9383, opened 2022-12-08
chore: use test process exit code

OrKoN:


Lightning00Blade (approved):

OrKoN:
Actually, that would not make sense. We need some other way to get coverage signal.

PR #9430, opened 2022-12-16
fix(Target): target interceptor is not async

kblok:
I found that the only call to `addTargetInterceptor` passes `onAttachedToTarget` and is not async.
I also found a typo 🤓


OrKoN (changes requested):
Thanks for the PR but it is not async on purpose. While the target is paused CDP might not respond to queries to the target, therefore, any awaited CDP calls in the interceptor might lead to a deadlock. The interceptor should only be used to schedule CDP calls and should not be able to delay runIfWaitingForDebugger calls.

kblok:
@OrKoN exactly. If `Page.#onAttachedToTarget` is not `async` (I just removed the keyword in the function). There is no need to await for it. There is no async code in that function.

OrKoN:
ah apologies, totally misread the title and the diff in the morning.

OrKoN (approved):

Cklim143:
``

PR #9436, opened 2022-12-19
chore: update dependencies

### LINKED ISSUE #9372: Update TypeScript to the latest version
It looks like we still have some tech debt in ElementHandle & JSHandle types and compilation fails if you try updating TypeScript. We should take a look into that and update the version.

cc @Lightning00Blade @jrandolf

jrandolf-google:
Fixed: https://github.com/puppeteer/puppeteer/issues/9372


OrKoN (approved):
Thanks!

OrKoN:
Note that the doc check is failing

PR #9444, opened 2022-12-20
feat: Support for `CHROME_BIN` and `FIREFOX_BIN`

### LINKED ISSUE #9418: [Feature Request]: use CHROME_BIN if present as a source for configuration.executablePath
### Bug description

Currently Puppeteer forces the use of `PUPPETEER_EXECUTABLE_PATH`,

I think it's subpar to name it `_PATH` either way, but the convention for images that also must host selenium, or protractor, karma, or other utilities is `CHROME_BIN`. I would suggest using or adopting this convention, or at the least sourcing from the environmental variable `CHROME_BIN`.

This request is to add `CHROME_BIN` to this https://github.com/puppeteer/puppeteer/blob/4bf338b26d6cceae2dc718074ed9839e905905f2/packages/puppeteer/src/getConfiguration.ts#L49

### Puppeteer version

current

### Node.js version

current

### npm version

current

Linux

### Configuration file

_No response_

### Relevant log output

_No response_

EvanCarroll:
Add additional documentation in docs/environmental_variables.md

Fixes: #9418

### DIFF
--- a/docs/environmental_variables.md
+++ b/docs/environmental_variables.md
@@ -1,0 +1,36 @@
+# Environmental Variables
+
+There are some environmental variables that are useful for configuring
+Puppeteer. These can be used to simplify pipeline execution. Notable
+configuration variables include.
+
+## Variables to configure browser location
+
+These variables for setting the path to the browser executables, these will
+automatically set `PUPPETEER_PRODUCT`, and disable the download of the browser.
+
+- `CHROME_BIN`: (preferred) Sets the path to the Chrome executable.
+- `FIREFOX_BIN`: (preferred) Sets the path to the FireFox executable.
+
+Further, you have these which can provide the same thing.
+
+- `PUPPETEER_PRODUCT`: Can be `chrome` or `firefox`, defaults to `chrome`
+- `PUPPETEER_EXECUTABLE_PATH`: Sets the path to a browser executable.
+- `PUPPETEER_SKIP_DOWNLOAD`: Disable the download of a browser, defaults to
+  being set if `PUPPETEER_EXECUTABLE_PATH` is set.
+
+## Variables to configure browser downloader
+
+If you're not providing the path to the browser, and you want it download you
+have these options which can help with configuration of the download:
+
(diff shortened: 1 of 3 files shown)


docs/environmental_variables.md
```diff
@@ -0,0 +1,36 @@
+# Environmental Variables
```
OrKoN, line 1: This should probably go into https://pptr.dev/guides/configuration

packages/puppeteer/src/getConfiguration.ts
```diff
@@ -30,6 +30,28 @@ export const getConfiguration = (): Configuration => {
     configuration.defaultProduct ??
     'chrome') as Product;
 
+  // Support the CHROME_BIN/FIREFOX_BIN convention found elsewhere. All of
+  // these, are lowest precedent and maintain current behaviors
```
OrKoN, line 34: it looks like some tests fail with this (not sure why though)

OrKoN (changes requested):
it looks like GitHub Actions already define CHROME_BIN variable that is wrong for out tests. We could probably fix fairly easily in CI but I am afraid that might become a breaking change for many users.

packages/puppeteer/src/getConfiguration.ts
OrKoN: Failure reason https://github.com/puppeteer/puppeteer/issues/9418#issuecomment-1360235835

EvanCarroll:
@OrKoN I truly doubt that it will undesirably effect anything except the test cases here. But, if you're talking about it internally, just keep in mind this is a pleasant surprise -- imho. This shows how widespread the convention is. Things should just work with 0 configuration if `CHROME_BIN` is supported, I'd hope we could agree it should be used.

OrKoN:
@EvanCarroll unfortunately, GitHub Actions are very common and the env variable also affects the transitive dependencies so will likely to have to deal with a big number of bug reports about things not working in a mysterious way. The tests are also failing because the binary is not of the right version and that is the risk of relying on the browser versions set for other tools. We only test Puppeteer against the version we ship with the Puppeteer so using other versions might break unexpectedly for the users.

PR #9477, opened 2023-01-09
chore: update bug template

jrandolf-google:
See internal doc.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug.yml
+++ b/.github/ISSUE_TEMPLATE/bug.yml
@@ -5,66 +5,111 @@ labels: [bug]
 body:
   - type: markdown
     attributes:
-      value: |
+      value: >
         ### Thanks for taking the time to fill out this bug report!
-        **Before filling out this report**, please check our [Troubleshooting](https://pptr.dev/troubleshooting) guide for solutions to common issues.
-        Also, verify the problem is *related to packages maintained by Puppeteer* (this does not include `puppeteer-extra`, `expect-puppeteer`, etc.).
-  - type: textarea
-    id: summary
-    attributes:
-      label: Bug description
-      description: What did you do? What did you expect to happen? What actually happened instead? Provide a [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example).
-      value: |
-        Steps to reproduce the problem:
 
-        1. …
+        **Before filling out this report**, please check our
+        [Troubleshooting](https://pptr.dev/troubleshooting) guide for solutions
+        to common issues.
+  - id: summary
+    type: textarea
+    attributes:
+      label: Bug expectation
+      description: What do you expect to happen? What actually happened instead?
+      placeholder: |
+        I expected ...
+
+        What happened instead was ...
     validations:
       required: true
-  - type: input
-    id: puppeteer-version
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -5,66 +5,102 @@ labels: [bug]
...
     attributes:
-      label: What operating system are you seeing the problem on?
-      multiple: true
+      label: Node version
+      description: What supported version of Node.js are you running?
       options:
-        - Linux
-        - macOS
-        - Windows
+        - '18'
```
OrKoN: how do we ensure that people running node 20 can still report the issue?

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -5,66 +5,102 @@ labels: [bug]
...
+    type: dropdown
     attributes:
-      label: Configuration file
-      description: |
-        Please copy and paste your configuration file [[Docs](https://pptr.dev/guides/configuration/)]. 
-        *No need for backticks — this automatically gets formatted into code.*
-      render: JavaScript
-  - type: textarea
-    id: logs
+      label: NPM version
```
OrKoN: what if people are using yarn or pnpm?

.github/ISSUE_TEMPLATE/bug.yml
jrandolf: Discussed offline.

.github/ISSUE_TEMPLATE/bug.yml
jrandolf: At the moment, we only support NPM. Do we want to support yarn and pnpm?

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -5,66 +5,102 @@ labels: [bug]
...
-        Steps to reproduce the problem:
 
-        1. …
+        **Before filling out this report**, please check our
+        [Troubleshooting](https://pptr.dev/troubleshooting) guide for solutions
+        to common issues.
+  - id: summary
+    type: textarea
+    attributes:
+      label: Bug expectation
```
Lightning00Blade, line 17: What about splitting this in two -> Expected result & Actual result?

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -5,66 +5,102 @@ labels: [bug]
...
+        Provide a [minimal, reproducible
+        example](https://stackoverflow.com/help/minimal-reproducible-example).
+        *No need for backticks — this automatically gets formatted into code.*
+      render: TypeScript
     validations:
       required: true
-  - type: input
-    id: node-version
+  - id: error
+    type: textarea
```
Lightning00Blade: How about using `input` here, less visible space would imply shorter text?

.github/ISSUE_TEMPLATE/bug.yml
jrandolf: We only care about expectation. Their code should show the actual results since we will run and gather artifacts.

OrKoN (approved):

PR #9480, opened 2023-01-10
chore: add issue analyzer

jrandolf-google:
See internal note.

### DIFF
--- a/tools/analyze_issue.mjs
+++ b/tools/analyze_issue.mjs
@@ -1,0 +1,286 @@
+#!/usr/bin/env node
+// @ts-check
+
+'use strict';
+
+import {writeFile, mkdir, copyFile} from 'fs/promises';
+import {dirname, join} from 'path';
+import semver from 'semver';
+import {fileURLToPath} from 'url';
+import {
+  versionsPerRelease,
+  lastMaintainedChromiumVersion,
+} from '../versions.js';
+import core from '@actions/core';
+
+const LAST_SUPPORTED_PUPPETEER_VERSION = versionsPerRelease.get(
+  lastMaintainedChromiumVersion
+);
+if (!LAST_SUPPORTED_PUPPETEER_VERSION) {
+  core.setFailed('No maintained version found.');
+}
+const LAST_SUPPORTED_NODE_VERSION = '14.19.0';
+
+const SUPPORTED_OSES = ['windows', 'macos', 'linux'];
+const SUPPORTED_PACKAGE_MANAGERS = ['yarn', 'npm', 'pnpm'];
+
+const codifyAndJoinValues = values => {
+  return values
+    .map(value => {
+      return `\`${value}\``;
+    })
+    .join(' ,');
+};
+const formatMessage = value => {
+  return value.trim();
+};
+const removeVersionPrefix = value => {
+  return value.startsWith('v') ? value.slice(1) : value;
+};
+
+const ERROR_MESSAGES = {
+  unsupportedOs(value) {
+    return formatMessage(`
+This issue has an unsupported OS: \`${value}\`. Only the following operating systems are supported: ${codifyAndJoinValues(
+      SUPPORTED_OSES
+    )}. Please verify the issue on a supported OS and update the form.
+`);
+  },
+  unsupportedPackageManager(value) {
+    return formatMessage(`
+This issue has an unsupported package manager: \`${value}\`. Only the following package managers are supported: ${codifyAndJoinValues(
+      SUPPORTED_PACKAGE_MANAGERS
+    )}. Please verify the issue using a supported package manager and update the form.
+`);
+  },
+  invalidPackageManagerVersion(value) {
+    return formatMessage(`
+This issue has an invalid package manager version: \`${value}\`. Versions must follow [SemVer](https://semver.org/) formatting. Please update the form with a valid version.
+`);
+  },
+  unsupportedNodeVersion(value) {
+    return formatMessage(`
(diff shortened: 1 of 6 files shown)


tools/analyze_issue.mjs
```diff
@@ -0,0 +1,272 @@
...
+  return values
+    .map(value => {
+      return `\`${value}\``;
+    })
+    .join(' ,');
+};
+const formatMessage = value => {
+  return JSON.stringify(value.trim());
+};
+const removeV = value => {
```
Lightning00Blade: Nit: `removeVPrefix` or `removeVersionPrefix`

tools/analyze_issue.mjs
```diff
@@ -0,0 +1,272 @@
...
+
+'use strict';
+
+import {writeFile, mkdir, copyFile} from 'fs/promises';
+import {dirname, join} from 'path';
+import semver from 'semver';
+import {fileURLToPath} from 'url';
+
+const LAST_SUPPORTED_PUPPETEER_VERSION = '18.2.1';
+const LAST_SUPPORTED_NODE_VERSION = '14.19.0';
```
Lightning00Blade, line 22: For the future we should pull this from NPM and Puppeteer repo respectively.

tools/analyze_issue.mjs
```diff
@@ -0,0 +1,272 @@
...
+const removeV = value => {
+  return value.startsWith('v') ? value.slice(1) : value;
+};
+
+const ERROR_MESSAGES = {
+  unsupportedOs(value) {
+    return formatMessage(`
+This issue has an unsupported OS: \`${value}\`. Only the following operating systems are supported: ${codifyAndJoinValues(
+      SUPPORTED_OSES
+    )}. Please edit the form with a supported OS.
```
Lightning00Blade: This sounds like just fix it to work, rather then test it on supported OS. How about:
`Please verify the issue on one the supported OS and update[edit] the issue[form].`

tools/analyze_issue.mjs
```diff
@@ -0,0 +1,272 @@
...
+      runsOn = 'windows-latest';
+      break;
+    case 'macos':
+      runsOn = 'macos-latest';
+      break;
+    case 'linux':
+      runsOn = 'ubuntu-latest';
+      break;
+    default:
+      console.log(`errorMessage=${ERROR_MESSAGES.unsupportedOs(os)}`);
```
Lightning00Blade: Nit: Here and for the other errors as well.
```suggestion
      console.error(`errorMessage=${ERROR_MESSAGES.unsupportedOs(os)}`);
```

tools/analyze_issue.mjs
```diff
@@ -0,0 +1,272 @@
...
+      throw new Error(`Unsupported OS: ${os}`);
+  }
+
+  if (!SUPPORTED_PACKAGE_MANAGERS.includes(packageManager)) {
+    console.log(
```
Lightning00Blade: Looks like there is a lot of duplicate code here. What about a function here (Predicate, ErrorMessage, ThrownError) => void/Throw ?

tools/analyze_issue.mjs
```diff
@@ -0,0 +1,272 @@
...
+  const lines = input.split('\n');
+  {
+    /** @type {(value: string) => void} */
+    let set = () => {
+      return void 0;
+    };
+    let j = 1;
+    let i = 1;
+    for (; i < lines.length; ++i) {
+      if (lines[i].startsWith('### Bug behavior')) {
```
Lightning00Blade, line 111: I think an Array with {title: "### Bug behavior", execute: (value) => void }. Then looping though it . If it starts we set the `set` and `brake`.

tools/analyze_issue.mjs
```diff
@@ -0,0 +1,272 @@
...
+      `errorMessage=${ERROR_MESSAGES.invalidPackageManagerVersion(
+        packageManagerVersion
+      )}`
+    );
+    throw new Error(
+      `Invalid package manager version: ${packageManagerVersion}`
+    );
+  }
+
+  console.log(`errorMessage=""`);
```
Lightning00Blade: I think we should just add `@actions/code` and use it here rather then using this to output method.

tools/analyze_issue.mjs
jrandolf: We use log here because the message needs to go to `$GITHUB_OUTPUT`

tools/analyze_issue.mjs
jrandolf: What is `@actions/code`?

tools/analyze_issue.mjs
jrandolf: This is more semantic.

tools/analyze_issue.mjs
jrandolf: It's the same length of code.

tools/analyze_issue.mjs
Lightning00Blade: For history -> Typo on my part `@actions/core` is package that simplifies working with GitHub Actions

tools/assets/verify_issue.ts
```diff
@@ -0,0 +1,68 @@
...
+          process.exit(0);
+        }
+        console.error('Script failed; unknown error found.');
+        process.exit(1);
+      }
+    };
+  }
+
+  for (let i = 0; i < maxRepetitions; ++i) {
+    const result = spawnSync('npm', ['start'], {
```
OrKoN, line 51: do we want a timeout here if the user script never resolves?

jrandolf: I put it on the step.

OrKoN (approved):

OrKoN:
Let's wait for @Lightning00Blade's LGTM.

Lightning00Blade (approved):

PR #9502, opened 2023-01-12
chore: forward mocha arguments through the custom runner

lutien:

This is a change to a custom mocha runner to forward arguments to mocha.

**Summary**

In order to start using puppeteer custom mocha runner, the Firefox team needs to forward some arguments to mocha through the runner to make it compatible with our log parser.

no

### DIFF
--- a/tools/mochaRunner/src/main.ts
+++ b/tools/mochaRunner/src/main.ts
@@ -129,14 +129,31 @@ async function main() {
         ? statsFilename
         : path.join(tmpDir, 'output.json');
       console.log('Running', JSON.stringify(parameters), tmpFilename);
+      const reporterArgumentIndex = process.argv.indexOf('--reporter');
       const args = [
         '-u',
         path.join(__dirname, 'interface.js'),
         '-R',
-        path.join(__dirname, 'reporter.js'),
+        reporterArgumentIndex === -1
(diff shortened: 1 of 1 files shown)


tools/mochaRunner/src/main.ts
```diff
@@ -129,14 +129,31 @@ async function main() {
         ? statsFilename
         : path.join(tmpDir, 'output.json');
       console.log('Running', JSON.stringify(parameters), tmpFilename);
+      const reporterArgumentIndex = process.argv.indexOf('--reporter');
       const args = [
         '-u',
         path.join(__dirname, 'interface.js'),
         '-R',
-        path.join(__dirname, 'reporter.js'),
+        reporterArgumentIndex === -1
```
OrKoN, line 137: I believe mochaRunner won't work with any standard reporter. Do you plan to re-implement a compatible reporter?

OrKoN (approved):

tools/mochaRunner/src/main.ts
lutien: yes, we extend JSON reporter (you can check here: https://phabricator.services.mozilla.com/differential/changeset/?ref=5931554), and it seems to be working

PR #9503, opened 2023-01-12
chore: filter expectations by the whole test or file name

lutien:

This is a change to a custom mocha runner to look for expectation of the test case by the whole test name instead of by the part of the name.

**Summary**

Working on integration of the puppeteer expectation file in mozilla repo and unskipping a lot of tests, I've noticed that some tests get wrong statuses. For example, a test case with the name `navigation Page.goto should fail when navigating to bad SSL` got the status of `navigation Page.goto should fail when navigating to bad SSL after redirects` or `ElementHandle specs ElementHandle.boundingBox should work` get the status of `ElementHandle specs ElementHandle.boundingBox should work with SVG nodes`. So it seems like checking for the whole name of the test should be safer, but let me know if I'm missing something here.

no

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -1236,10 +1236,10 @@
     "expectations": ["SKIP"]
   },
   {
-    "testIdPattern": "[page.spec] Page ExecutionContext.queryObjects should work for non-blank page",
+    "testIdPattern": "[page.spec] Page ExecutionContext.queryObjects should work for non-trivial page",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["firefox"],
-    "expectations": ["SKIP"]
+    "expectations": ["FAIL"]
   },
   {
     "testIdPattern": "[page.spec] Page Page.addStyleTag should throw when added with content to the CSP page",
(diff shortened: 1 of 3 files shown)


OrKoN:
The idea was that the expectations can disable multiple tests at once. For example, `[ariaqueryhandler.spec]` provides an expectation for all ariaqueryhandler tests.

OrKoN:
The current set of expectations still works with this because SKIP is implemented differently (because of how mocha works): https://github.com/puppeteer/puppeteer/blob/main/tools/mochaRunner/src/interface.ts#L36

lutien:
@OrKoN, right, what do you think then if we will check for a test name and a test file at the same time? I've pushed the code to illustrate my proposal.

test/TestExpectations.json
```diff
@@ -1236,10 +1236,10 @@
     "expectations": ["SKIP"]
   },
   {
-    "testIdPattern": "[page.spec] Page ExecutionContext.queryObjects should work for non-blank page",
+    "testIdPattern": "[page.spec] Page ExecutionContext.queryObjects should work for non-trivial page",
```
lutien, line 1239: Before, this test case was skipped because `Page ExecutionContext.queryObjects should work` is marked as skipped.

OrKoN:
@lutien I think that would work yes, because we didn't end up ever intentionally using the prefix matches expect for on the spec level.

OrKoN (approved):

OrKoN:
@lutien actually it seems like bidi tests are not correctly skipped/enabled now? I guess we might need at least the support for the empty string pattern meaning all tests.

lutien:
@OrKoN, you're right, I've added the check for an empty string

PR #9506, opened 2023-01-13
chore: check if rimraf package is installed before running the cleanup

lutien:

A change to the npm cleanup script.

**Summary**

When working with puppeteer in mozilla repo we regularly update puppeteer versions, which leads to the situation when developers could have outdated build folders, to avoid this we would like to add `npm clean` command as a part of puppeteer setup process and running test. But we use the same command on the ci, where we don't have dependencies set up yet, also developers might have a fresh checkout without dependencies set up. To avoid the errors in this case, the idea would be to check first if `rimraf` package is installed.

no

### DIFF
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "build": "npm run build --workspaces --if-present",
     "check:pinned-deps": "tsx tools/ensure-pinned-deps",
     "check": "npm run check --workspaces --if-present && run-p check:*",
-    "clean": "npm run clean --workspaces --if-present && rimraf **/.wireit",
+    "clean": "npm list rimraf && npm run clean --workspaces --if-present && rimraf **/.wireit",
     "commitlint": "commitlint --from=HEAD~1",
     "debug": "mocha --inspect-brk",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -10,7 +10,7 @@
     "build": "npm run build --workspaces --if-present",
     "check:pinned-deps": "tsx tools/ensure-pinned-deps",
     "check": "npm run check --workspaces --if-present && run-p check:*",
-    "clean": "npm run clean --workspaces --if-present && rimraf **/.wireit",
+    "clean": "npm list rimraf && npm run clean --workspaces --if-present && rimraf **/.wireit",
```
OrKoN, line 13: so if the rimraf is not yet installed, then nothing will be cleaned up? The behaviour looks unexpected and I think it's better to error in that case (I think that would be the current behaviour)

lutien: Yes, now it will throw an error. Alright, I will think about different solution

lutien:
Actually, just checked again, and we don't really need this change, sorry for unnecessary review

PR #9385, opened 2022-12-08
[WIP] feat(puppeteer-core): enhance typings for screenshot

shigma:
**What kind of change does this PR introduce?**

Typings enhancement.

**Did you add tests for your changes?**

N/A

**If relevant, did you update the documentation?**

N/A

**Summary**

Add override signatures for `page.screenshot()` and `elementHandle.screenshot()` for better typings.

**Does this PR introduce a breaking change?**

No.

**Other information**

PR #9529, opened 2023-01-17
fix: improve screenshot method types

### LINKED ISSUE #9512: [Bug] [Typescript]: page.screenshot with dubious return type
### Bug expectation

I expected `page.screenshot({ encoding: "binary" })` to return me a `Buffer` type but instead I got the union type `string | Buffer`.

So now I need to coerce the type doing `page.screenshot({ encoding: "binary" }) as Buffer` in order to deal with it in the code.

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript

(async () => {
  const imageBuffer = await page.screenshot({ encoding: "binary" }) // Type is string | Buffer. Correct should be Buffer
  const imageString = await page.screenshot({ encoding: "base64" }) // Type is string | Buffer. Correct should be string
  console.log({imageBuffer, imageString});
})();
```

### Error string

no error

### Puppeteer configuration

_No response_

### Puppeteer version

19.5.2

### Node version

18.12.1

### Package manager

yarn

### Package manager version

1.22.19

### Operating system

macOS

OrKoN:
The only drawback of adding overrides is that our documentation generator would generate a page for each overload which might be hard to read/find on the documentation website.

Closes #9512 #9385


OrKoN:
Also, it looks like using the overrides does not allow linking to a method anymore 🤔

OrKoN:
The corresponding API extractor issue is here https://github.com/microsoft/rushstack/issues/1957

Lightning00Blade (approved):
LGTM

PR #9568, opened 2023-01-24
fix: don't clean up previous browser versions

### LINKED ISSUE #9533: [Bug]: Error: Could not find Chromium (rev. 1069273)
### Bug expectation

I expected to successfully run puppeteer, as it always did in prior versions.

I instead got an error:
<code>
Error: Could not find Chromium (rev. 1069273). This can occur if either
 1. you did not perform an installation before running the script (e.g. `npm install`) or
 2. your cache path is incorrectly configured (which is: /home/sysadm/.cache/puppeteer).
For (2), check out our guide on configuring puppeteer at https://pptr.dev/guides/configuration.
    at ChromeLauncher.resolveExecutablePath (/var/www/html/oas-restapi-fork1/node_modules/puppeteer-core/lib/cjs/puppeteer/node/ProductLauncher.js:119:27)
    at ChromeLauncher.executablePath (/var/www/html/oas-restapi-fork1/node_modules/puppeteer-core/lib/cjs/puppeteer/node/ChromeLauncher.js:205:25)
    at ChromeLauncher.launch (/var/www/html/oas-restapi-fork1/node_modules/puppeteer-core/lib/cjs/puppeteer/node/ChromeLauncher.js:93:37)
</code>

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript

"use strict";

const puppeteer = require ('puppeteer');

async function main () {

    try {

        // instantiate puppeteer
        console.log (`Launching puppeteer`);
        const browser = await puppeteer.launch ();

        // create new page:
        console.log (`Creating page`);
        const page = await browser.newPage ();

    } catch (err) {

        console.log (err);

    }

}

main ();
```

### Error string

Error: Could not find Chromium (rev. 1069273)

### Puppeteer configuration

```TypeScript

```

### Puppeteer version

19.6.0

### Node version

v16.19.0

### Package manager

npm

### Package manager version

9.3.0

### Operating system

Linux

OrKoN:
Since we moved to the central binaries cache it does not make sense to clean up old binaries automatically because multiple installations can use different versions. We expect the users to clean the cache from time to time until we offer a CLI for managing the browsers.

Closes #9533

### DIFF
--- a/packages/puppeteer-core/src/node/BrowserFetcher.ts
+++ b/packages/puppeteer-core/src/node/BrowserFetcher.ts
@@ -504,6 +504,13 @@ export class BrowserFetcher {
   #getFolderPath(revision: string): string {
     return path.resolve(this.#downloadPath, `${this.#platform}-${revision}`);
   }
+
+  /**
+   * @internal
+   */
+  getDownloadPath(): string {
+    return this.#downloadPath;
+  }
 }
 
 function parseFolderPath(
--- a/packages/puppeteer/src/node/install.ts
+++ b/packages/puppeteer/src/node/install.ts
@@ -94,13 +94,18 @@ export async function downloadBrowser(): Promise<void> {
       logPolitely(
         `${supportedProducts[product]} (${revisionInfo.revision}) downloaded to ${revisionInfo.folderPath}`
       );
-      localRevisions = localRevisions.filter(revision => {
+      const otherRevisions = localRevisions.filter(revision => {
         return revision !== revisionInfo.revision;
       });
-      const cleanupOldVersions = localRevisions.map(revision => {
-        return browserFetcher.remove(revision);
-      });
-      Promise.all([...cleanupOldVersions]);
+      if (otherRevisions.length) {
+        logPolitely(
+          `Other installed ${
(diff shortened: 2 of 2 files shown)


Lightning00Blade (approved):
Maybe we should add a log for the number of revisions being stored. That way the user easily see how many revisions they have saved and act upon that if need be.

OrKoN:
@Lightning00Blade added a log message.

jrandolf-google (approved):

q--:
So I saw this in the release notes, and it is not clear to me whether this applies to all of puppeteer or only to puppeteer-core. (I'm using the former.)

Does the `puppeteer` package still clean up old Chromium versions, or do I have to start cleaning them up myself so my disk won't fill up with unused Chromium installations? And if it's the latter, is there any (official or recommended) way to automate it?

jacobg:
I was having this error for revision 1069273, then bumped to puppeteer 19.6.2, deployed to Google Cloud Function, and the error went away. Then a little later I did another deployment for an unrelated reason, and the error came back again on a new revision 1083080.

So it doesn't seem that 19.6.2 fixes the issue.

PR #9575, opened 2023-01-24
fix: use LazyArg for puppeteer utilities

jrandolf-google:
This PR fixes the following edge case:

 - `const oldPromise = world.puppeteerUtil`. 
 - setContext occurs but context is immediately destroyed, i.e. `world.#puppeteerUtil === oldPromise` is not resolved.
 - clearContext occurs due to destruction, i.e. `world.#puppeteerUtil` is replaced (`world.#puppeteerUtil !== oldPromise`).
 - `oldPromise` never resolves.

### DIFF
--- a/packages/puppeteer-core/src/common/IsolatedWorld.ts
+++ b/packages/puppeteer-core/src/common/IsolatedWorld.ts
@@ -28,7 +28,7 @@ import {JSHandle} from './JSHandle.js';
 import {LazyArg} from './LazyArg.js';
 import {LifecycleWatcher, PuppeteerLifeCycleEvent} from './LifecycleWatcher.js';
 import {TimeoutSettings} from './TimeoutSettings.js';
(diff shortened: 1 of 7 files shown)


packages/puppeteer-core/src/common/IsolatedWorld.ts
```diff
@@ -517,12 +520,12 @@ export class IsolatedWorld {
           root,
           timeout,
         },
-        new LazyArg(async () => {
+        new LazyArg(() => {
           try {
             // In case CDP fails.
-            return await this.puppeteerUtil;
+            return this.puppeteerUtil;
```
OrKoN: I don't think this can throw without an await here.

jrandolf: Obsolete.

OrKoN (approved):

Issue #9523, opened 2023-01-17
[Bug]: Page.goto() throws error for HTTP status code 4xx, 5xx etc.


VoidMonk:
### Bug expectation

Official [documentation](https://pptr.dev/api/puppeteer.page.goto#remarks) for `Page.goto()` says:

```
page.goto will throw an error if: there's an SSL error (e.g. in case of self-signed certificates). - target URL is invalid. - the timeout is exceeded during navigation. - the remote server does not respond or is unreachable. - the main resource failed to load.

page.goto will not throw an error when any valid HTTP status code is returned by the remote server, including 404 "Not Found" and 500 "Internal Server Error". The status code for such responses can be retrieved by calling response.status(). 
```

But, in headful mode, `Page.goto()` throws an error if the response status code if 404, 503 etc. and the body is empty. If the mode is headless or body is not empty (e.g. some HTML or text) then it doesn't throw an error.

Expected behavior is that `Page.goto()` should not throw an error for 4xx, 5xx status codes, irrespective of the launch mode (headless or headful) and the response body content.

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Minimal, reproducible example

```TypeScript
import puppeteer from 'puppeteer';

// scenario 1: test page returns HTTP 404 without a body
// result: puppeteer throws an error
(async () => {
  const browser = await puppeteer.launch({ headless: false });
  const page = await browser.newPage();

  await page.goto('https://httpbin.org/status/404');

  await browser.close();
})();

// scenario 2: test page returns HTTP 404 with a body
// result: puppeteer doesn't throw an error
(async () => {
  const browser = await puppeteer.launch({ headless: false });
  const page = await browser.newPage();

  await page.goto('https://httpbin.org/notfound');

  await browser.close();
})();
```

### Error string

net::ERR_HTTP_RESPONSE_CODE_FAILURE at https://httpbin.org/status/404

### Puppeteer configuration

```
{
  headless: false
}
```

### Puppeteer version

19.5.2

### Node version

18.13.0

### Package manager

npm

### Package manager version

9.2.0

### Operating system

Windows

Rafael-Martins:
I have been able to replicate the issue, would it be possible for me to attempt to resolve it?

Rafael-Martins:
Hey we merged a pr that fix this behavior! @VoidMonk

Rafael-Martins:
I think that we can close this!

PR #9577, opened 2023-01-25
fix: `page.goto` error throwing on 40x/50x responses with an empty body (#9523)

Rafael-Martins:
The observed problem is:  when running Chrome in headful mode, responses with status 404 and 500 that lack a body are treated as errors.
Solution: bypass the error using the error message.

bugfix

yes

**Summary**

Remove the error throw on the `navigate` method, in order to get [#9523](https://github.com/puppeteer/puppeteer/issues/9523) fixed


Rafael-Martins:
Hey guys, I'm new here, please tell me if this change makes sense!

OrKoN (changes requested):
I think we cannot just remove the error message here. Did you find out why the browser handles the errors differently? perhaps it should be fixed upstream first.

Rafael-Martins:
> I think we cannot just remove the error message here. Did you find out why the browser handles the errors differently? perhaps it should be fixed upstream first.

Oh, I see, just saw some tests failing because of this as well, I'm digging into it!

Rafael-Martins:
The issue only happen in chrome, looking at the  I see that `errorText` is returned when the navigation is failed, so chrome is considering the `404` as a navigation error. Can we filter it like my last commit?

It appears that the issue at hand is specific to the Chrome browser. By consulting the documentation for the [devtools-protocol docs](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-navigate) it can be observed that the errorText property is provided when a navigation failure occurs. In this case, Chrome is interpreting a 404 error as a navigation failure. To address this, it may be possible to implement a filter similar to the one utilized in a previous commit, in order to bypass this error.

OrKoN:
@Rafael-Martins so when the browser treats the navigation as an error it shows a different kind of the error page I think which is usually restricted to CDP. What is the situation after the navigation errors, can you still evaluate or navigate the page to a different URL?

OrKoN (changes requested):
Let's add a test for covering this scenario

Rafael-Martins:
> @Rafael-Martins so when the browser treats the navigation as an error it shows a different kind of the error page I think which is usually restricted to CDP. What is the situation after the navigation errors, can you still evaluate or navigate the page to a different URL?

I tested and the navigate works normally, I added 2 tests that covers the `404`,`505` responses with empty body, they look similar or almost redundant, if you think that it is not needed I can remove.

test/src/navigation.spec.ts
```diff
@@ -304,6 +304,30 @@ describe('navigation', function () {
       expect(response.ok()).toBe(false);
       expect(response.status()).toBe(404);
     });
+    it('should not throw a error when get a 404 response with empty body', async () => {
```
OrKoN:
```suggestion
    it('should not throw an error for a 404 response with an empty body', async () => {
```

test/src/navigation.spec.ts
```diff
@@ -304,6 +304,30 @@ describe('navigation', function () {
...
+      server.setRoute('/404-error', (_, res) => {
+        res.statusCode = 404;
+        res.end();
+      });
+
+      const response = (await page.goto(server.PREFIX + '/404-error'))!;
+      expect(response.ok()).toBe(false);
+      expect(response.status()).toBe(404);
+    });
+    it('should not throw a error when get a 500 response with empty body', async () => {
```
OrKoN:
```suggestion
    it('should not throw an error for a 500 response with an empty body', async () => {
```

OrKoN (approved):

Issue #9541, opened 2023-01-18
[Bug] [Typescript]: page.waitForXPath with dubious return type 


rhuanbarreto:
### Bug expectation

Very similar to #9512.

I expected `page.waitForXPath('//button')` to return me a `Promise<ElementHandle<Node>>` type but instead I got the union type `Promise<ElementHandle<Node> | null>`.

So now I need to coerce the type doing `(await page.waitForXPath('//button')) as ElementHandle<Node>` in order to deal with it in the code.

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Minimal, reproducible example

```TypeScript
import puppeteer from "puppeteer";

(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  const shouldBeNull = await page.waitForXPath("//button", { hidden: true }); // Type is ElementHandle<Node> | null. Correct should be null
  const shouldBeElementHandle = await page.waitForXPath("//body"); // Type is ElementHandle<Node> | null. Correct should be ElementHandle<Node>
  console.log({ shouldBeNull, shouldBeElementHandle });
  browser.close();
  process.exit();
})();
```

### Error string

no error

### Puppeteer version

19.5.2

### Node version

18.12.1

### Package manager

yarn

### Package manager version

1.22.19

### Operating system

macOS

Rafael-Martins:
Hey, @rhuanbarreto I'm new here, just trying to help!

Based on my research, the union type `Promise<ElementHandle<Node> | null>` appears to be suitable for the `waitForXPath` implementation, as it accounts for the potential null response due to varying reasons.

Perhaps you could use the `!` syntax instead of `await page.waitForXPath('//button')) as ElementHandle<Node>` in your code. like:
```js
 import puppeteer, { ElementHandle } from "puppeteer";

(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  const shouldBeNull = await page.waitForXPath("//button", { hidden: true }); // Type is ElementHandle<Node> | null. Correct should be null
  const shouldBeElementHandle = await page.waitForXPath("//body"); // Type is ElementHandle<Node> | null. Correct should be ElementHandle<Node>
  console.log({ shouldBeNull, shouldBeElementHandle });

  const functionThatReceivesElementHandle = (element: ElementHandle<Node>) => {}

  functionThatReceivesElementHandle(shouldBeElementHandle!)

  browser.close();
  process.exit();
})();
```

[...]

PR #9583, opened 2023-01-25
docs: clarify the waitForXPath return type (#9541)

Rafael-Martins:

test & docs

yes

**Summary**

Adds more description to `waitForXPath` and add more tests as well. Related to: [#9541](https://github.com/puppeteer/puppeteer/issues/9541)


Rafael-Martins:
I brought some discussion about this type [here](https://github.com/puppeteer/puppeteer/issues/9541#issuecomment-1404266812)

OrKoN (changes requested):
Could you please run `npm run docs` to regenerate the docs?

OrKoN (changes requested):
Could you please add a test similar to the test called `hidden should wait for display: none` that verifies that hidden: true always returns null. I did a quick check and it looks like we do return element handle.

Rafael-Martins:
> Could you please add a test similar to the test called `hidden should wait for display: none` that verifies that hidden: true always returns null. I did a quick check and it looks like we do return element handle.

I wrote some tests and it seems to be returning null just if the element is not found at the beginning, if the element has found and became hidden it returns a empty object.

This function should always return null `(when hidden: true)`? even in the second situation?

And for now, just testing the first situation is fine?

OrKoN:
I think if we ever return an ElementHandle value with hidden: true, we cannot change the types here.

OrKoN:
I think the documentation might be outdated here? perhaps we should clarify the behaviour better in the documentation.

Rafael-Martins:
> I think the documentation might be outdated here? perhaps we should clarify the behaviour better in the documentation.

It appears to be the case from my perspective :thinking:

I believe the test I added is beneficial and reflective of the current behavior, and can be kept in place. Is it preferred that I revert the type changes and enhance the documentation?

OrKoN:
Yes, let's keep the test and update the documentation. Perhaps we can extend the test to also demonstrate non-null value with hidden: true when the element becomes hidden in the middle of the call.

OrKoN (approved):

PR #9588, opened 2023-01-26
chore: remove unused return type

kblok:
# Description

I don't know if something got lost when this code was moved from the frame manager to the frame. But the output of the `navigate` function is not being used.


packages/puppeteer-core/src/common/Frame.ts
```diff
@@ -353,14 +353,11 @@ export class Frame {
           referrerPolicy,
         });
         ensureNewDocumentNavigation = !!response.loaderId;
-        return response.errorText
```
OrKoN, line 356: isn't it used on line 313?

kblok: Yes yes. Sorry about that. It was late in the night :p

PR #9589, opened 2023-01-26
fix: mimic rejection for PuppeteerUtil on early call

jrandolf-google:
This PR rejects the PuppeteerUtil promise to mimic the behavior of when PuppeteerUtil cannot be evaluated due to navigation.

This also fixes a problem where the timeout would clear on navigation.

### DIFF
--- a/packages/puppeteer-core/src/common/WaitTask.ts
+++ b/packages/puppeteer-core/src/common/WaitTask.ts
@@ -170,11 +170,10 @@ export class WaitTask<T = unknown> {
   async terminate(error?: unknown): Promise<void> {
     this.#world.taskManager.delete(this);
 
-    if (this.#timeout) {
-      clearTimeout(this.#timeout);
-    }
-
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/common/WaitTask.ts
```diff
@@ -170,11 +170,10 @@ export class WaitTask<T = unknown> {
   async terminate(error?: unknown): Promise<void> {
     this.#world.taskManager.delete(this);
 
-    if (this.#timeout) {
-      clearTimeout(this.#timeout);
-    }
-
     if (error && !this.#result.finished()) {
+      if (this.#timeout) {
```
OrKoN, line 174: is this change needed?

jrandolf: This could be a separate change, but I've made it here since it's small. Forgot to remark about it.

OrKoN (approved):

Issue #9514, opened 2023-01-14
[Bug]: Unhandled Promise rejection in CDPPage


muodov:
### Bug expectation

I'm investigating some strange crashes in the puppeteer-based crawler that loads lots of random web pages on scale. One of the causes I found is the unhandled rejection caused by async functions in CDP event listeners.

This was actually reported in https://github.com/puppeteer/puppeteer/issues/2811, but I'm not sure the [proposed workaround](https://github.com/puppeteer/puppeteer/issues/2811#issuecomment-403669356) works. You see, the [handlers in Page.ts](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/Page.ts#L731) are async functions, but they are [passed](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/Page.ts#L226) to `mitt`, which [does not seem to care](https://github.com/developit/mitt/blob/main/src/index.ts#L105) about promise results.
As a result, when these handlers throw, this crashes the whole node process (by default [since v15](https://nodejs.org/api/cli.html#--unhandled-rejectionsmode)). And (correct me if I'm wrong) there's no way to attach a handler to this outside of the puppeteer code, apart from a catch-all `unhandledRejection`.

The original issue argues that it's up to the developer to await for each and every command, but in our case we want to have timeouts for each page, when we forcibly close it. After that we don't care about any unfinished processing, but obviously we don't want the crawler to crash.

Right now it seems like a workaround is to set `--unhandled-rejections=warn`, but I wonder if it'd be better to handle this inside puppeteer code. I mean, so far it doesn't seem like it was the intention to crash in this case...or was it?

### Bug behavior

- [X] Flaky
- [ ] PDF

### Minimal, reproducible example

```TypeScript
// sorry, I don't have a good repro for this race condition
```

### Error string

Error: INTERNAL ERROR: missing context with id = ...

### Puppeteer version

19.5.2

### Node version

v18.12.1

### Package manager

npm

### Package manager version

8.19.2

### Operating system

Linux

muodov:
@aslushnikov @JoelEinbinder just wondering if something has changed since https://github.com/puppeteer/puppeteer/issues/2811?

jrandolf:
The developers you mentioned are no longer maintainers of Puppeteer. In any case, this seems like a feature. Could you file a new issue for a feature and close this?

muodov:
[...]

PR #9595, opened 2023-01-26
fix: ignore not found contexts for console messages

OrKoN:
I think it is possible that there is no context for a console message. I am not sure we should try to await for a context potentially arriving later. Since the reports say it happens in practice, let's turn the assertion into a debug log.

See https://github.com/puppeteer/puppeteer/issues/9514#issuecomment-1405571870


jrandolf:
Waiting would have to be a timeout since we could never know if it will come. That being said, is this really possible? Do you have a bug?

OrKoN:
@jrandolf the bug is linked in the description.

jrandolf-google (approved):

PR #9599, opened 2023-01-27
chore: Add launch.json for debugging tests

Lightning00Blade:
This makes running or debugging test easier for VSCode user. 
Have to check DevContainer behavior, but should make contributing from there easier as well.

TODO: Add documentation.


jrandolf:
Adding `launch.json` is generally not done because debugging is very use-case specific. This particular JSON doesn't really take advantage of breakpoint debugging too. It's just running the tests from what I see. What may make sense is a launch.tmpl.json that contains an example configuration and also implement debugging as featured in `pptr.dev`. This would help developers of Puppeteer, but also model current usage.

Lightning00Blade:
VsCode attaches a debugger to node when running from `launch.json` so brakepoints put in puppeteer or test do stop. Implementing the thing in the debugging page is for users of puppeteer, not contributes this PR is aimed for.

charlestonXdao (approved):
good

jrandolf-google (approved):

jrandolf-google (approved):

PR #9600, opened 2023-01-27
chore: Update build dependencies

Lightning00Blade:
Some of our build processes included generating API docs and formatting it. 
This slowed down the generation of the builds. We should see faster CI job because of this change.
The change makes using `--watch` like `npm run build --watch --workspace <package>` actually usable.
The building of the API docs is now done on the `npm run docs` command. 

Drive-by: Added documentation about Watch mode via `wireit`


packages/puppeteer-core/package.json
```diff
@@ -83,8 +83,8 @@
       ],
       "clean": false,
       "files": [
-        "lib/esm/third_party/**",
-        "lib/cjs/third_party/**"
+        "third_party/**",
```
Lightning00Blade: @jrandolf Any specific reason we had it set-up like so? It was causing a circular update when running watch mode.

packages/puppeteer/package.json
```diff
@@ -40,7 +40,6 @@
     "clean": "tsc -b --clean && rimraf lib",
     "format:types": "wireit",
     "generate:package-json": "wireit",
-    "generate:sources": "wireit",
```
Lightning00Blade, line 43: No such `wireit` script exits. Never used.

packages/puppeteer-core/package.json
jrandolf: This is related to how `wireit` works. It's a little finicky, but if this works for us, then I'm down for the change.

OrKoN: Will take a look after build failures are fixed

OrKoN (approved):

PR #9612, opened 2023-02-01
fix: restore WaitTask terminate  condition

### LINKED ISSUE #9610: [Bug]: page.waitForSelector keeps the process open for its `timeout` duration (19.6.2, 19.6.1)
### Bug expectation

I expect `page.waitForSelector` not to prevent the process from exiting when used normally. Instead, in 19.6.1 and 19.6.2 on both Ubuntu and Windows 10, using `page.waitForSelector` keeps the process from exiting for the duration of its `timeout` value (default 30 seconds) even after `browser.close()`. 19.6.0 behaves as expected. Specifying `{timeout: 0}` allows the process to exit as expected, but it's an unsafe workaround that can hang the script.

See [Why does Puppeeteer cause my test suite to hang for 30 seconds when I use "waitForSelector" even though I'm calling "close" on the page and browser?](https://stackoverflow.com/questions/75292123/why-does-puppeeteer-cause-my-test-suite-to-hang-for-30-seconds-when-i-use-waitf) on Stack Overflow.

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript
const puppeteer = require("puppeteer"); // 19.6.1 or 19.6.2

const html = `<!DOCTYPE html><html><body><p>hi</p></body></html>`;

let browser;
(async () => {
  await page.setContent(html);
  const el = await page.waitForSelector("p");
  console.log(await el.evaluate(el => el.textContent));
})()
  .catch(err => console.error(err))
```

### Error string

no error

### Puppeteer configuration

_No response_

### Puppeteer version

19.6.2

### Node version

18.7.0

### Package manager

npm

### Package manager version

9.3.0

### Operating system

Linux

ggorlen:

Fixes a regression caused by [fix: mimic rejection for PuppeteerUtil on early call](https://github.com/puppeteer/puppeteer/commit/1980de91a161523c7098a79919b20e6d8d2e5d81) and closes #9610.

No

No

**Summary**

See #9610. A recent change causes `page.waitForSelector` to keep the Node process from exiting. This fix reverts to the old behavior.

No.

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/common/WaitTask.ts
+++ b/packages/puppeteer-core/src/common/WaitTask.ts
@@ -177,10 +177,11 @@ export class WaitTask<T = unknown> {
   async terminate(error?: unknown): Promise<void> {
     this.#world.taskManager.delete(this);
 
+    if (this.#timeout) {
+      clearTimeout(this.#timeout);
+    }
+
     if (error && !this.#result.finished()) {
-      if (this.#timeout) {
-        clearTimeout(this.#timeout);
-      }
       this.#result.reject(error);
     }
 


jrandolf-google (approved):

Maxim-Mazurok:
I can confirm that I've had this issue in `v19.6.2` and `v19.6.3` works fine, thank you!

PR #9614, opened 2023-02-01
chore: Remove Rollup dts plugin

Lightning00Blade:
Removes the Rollup dts plugin as it was causing issues with BiDi types. 
All our types coming from `third_party` are private anyway and don't get bundled with in the `types.d.ts`.
Verified that the `SHA1` hash for `types.d.ts` is the same before and after this change. 

Drive-By: While trying to resolve the issue updated to Rollup 3.


packages/puppeteer-core/rollup.third_party.config.mjs
```diff
@@ -26,16 +25,16 @@ export default ['cjs', 'esm'].flatMap(outputType => {
...
     });
   }
-  for (const typesFile of glob.sync(`${thirdPartyPath}/**/*.d.ts`)) {
-    configs.push({
-      input: typesFile,
-      output: {file: typesFile, format: outputType},
-      plugins: [dts({respectExternal: true})],
-    });
-  }
+  // for (const typesFile of glob.sync(`${thirdPartyPath}/**/*.d.ts`)) {
```
OrKoN: remove commented code?

OrKoN (approved):

PR #9618, opened 2023-02-02
chore: set supported typescript range

### LINKED ISSUE #9582: [Bug]: puppeteer ships invalid typings for older versions of Typescript 4.3
### Bug expectation

We are on TypeScript 4.3.5. Puppeteer repo is on 4.9. 

we try to migrate to v19 but following Typescript errors started to occur (migrating from v17 to v19)

```
src/utils/visitPage.ts:40:9 - error TS2589: Type instantiation is excessively deep and possibly infinite.

await page.waitForSelector('#root-provider');
```

The cause of this is recursive mapped type provided as return type of `waitForSelector` which is `Promise<ElementHandle<NodeFor<Selector>> | null>`.

**Workaround:**

Annotate every puppeteer api call that's causing issues via `@ts-ignore` (which will add significant technical dept to the codebase)

**Solution:**

- change the types complexity (might be not possible)
- start using `typeVersions` to support older TypeScript version ( which is a standard OSS practice )
- explicitly state that you don't support particular versions of TypeScript and provide lowest possible ts version as peerDependency to puppeteer

### Bug behavior

- [ ] Flaky
- [ ] PDF

- TS error: https://uifabric.visualstudio.com/fabricpublic/_build/results?buildId=283432&view=logs&j=258ec178-2d8b-5611-7b9b-60c5c95dae55&t=254ee48e-1503-524f-f75e-e2c69d0380f5

- PR where we ran into the issue: https://github.com/microsoft/fluentui/pull/26475

### Error string

Type instantiation is excessively deep and possibly infinite.

### Puppeteer configuration

_No response_

### Puppeteer version

19.6.0

### Node version

16.16.0

### Package manager

yarn

### Package manager version

1.23.34

### Operating system

macOS

jrandolf-google:
Fixed: https://github.com/puppeteer/puppeteer/issues/9582


packages/puppeteer-core/package.json
```diff
@@ -172,5 +172,8 @@
     "tar-fs": "2.1.1",
     "unbzip2-stream": "1.4.3",
     "ws": "8.11.0"
+  },
+  "peerDependencies": {
```
OrKoN, line 178: I think peer dependency might print a warning/fail install if not satisfied. Since not all users use TypeScript I don't think we should include it.

packages/puppeteer-core/package.json
```diff
@@ -173,5 +173,13 @@
...
+    "typescript": ">= 4.7.4"
+  },
+  "peerDependenciesMeta": {
+    "typescript": {
+      "optional": true
```
OrKoN, line 183: what does npm print if you install puppeteer and don't have typescript installed?

jrandolf: No error. See https://docs.npmjs.com/cli/v7/configuring-npm/package-json#peerdependenciesmeta

package.json
```diff
@@ -103,7 +103,7 @@
     "text-diff": "1.0.1",
     "tsd": "0.25.0",
     "tsx": "3.12.1",
-    "typescript": "4.9.4",
+    "typescript": "4.9.3",
```
OrKoN: why downgrade from 4.9.4?

OrKoN (approved):

PR #9626, opened 2023-02-05
fix: Text not itterable error when fast typing

rkfx:

**Summary**

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/common/Input.ts
+++ b/packages/puppeteer-core/src/common/Input.ts
@@ -276,7 +276,7 @@ export class Keyboard {
    * if specified, is the time to wait between `keydown` and `keyup` in milliseconds.
    * Defaults to 0.
    */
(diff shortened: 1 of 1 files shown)


OrKoN (changes requested):
Could you please provide a minimal example that demonstrates the issue you are solving?

Xlangk:
Bit tired rn, just woke up, but I can describe it. If you are typing in a text input after focusing it, sometimes there will be no data and so the handler inside the library expects an [] but gets undefined, the `= []` defaults the value to an empty array. After testing it seemed to ahve fix any issues.

sorry if this is unprofessional, xD im jsut som 15 yr old

OrKoN:
I see. It looks like you should fix it on the level where the data gets undefined/empty. Also, `= []` is not a good default value for strings, it should be `= ''`. Typing an empty string makes no sense so that input should be checked before calling Puppeteer.

PR #9632, opened 2023-02-07
chore: Extract JSHandle to API

Lightning00Blade:
Extract the JSHandle to prepare for BiDi implmentation.


packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -20,21 +20,100 @@ import {ExecutionContext} from './ExecutionContext.js';
...
 import {ElementFor, EvaluateFunc, HandleFor, NodeFor} from './types.js';
 import {KeyInput} from './USKeyboardLayout.js';
 import {debugError, isString} from './util.js';
 import {CDPPage} from './Page.js';
+import {MouseButton} from './Input.js';
+
+/**
+ * @public
+ */
+export interface BoxModel {
```
Lightning00Blade: Moved the Types to `ElementHandle` as they were unused in `JSHandle`.

OrKoN: I think we can move these to the API file. WDYT?

packages/puppeteer-core/src/common/JSHandle.ts
```diff
@@ -72,7 +46,7 @@ export interface BoundingBox extends Point {
  *
  * @public
```
OrKoN: I think we can mark implementation class and methods as internal. We can probably remove duplicated documentation in the implementation class too (I think docs diff is failing because of that)

Lightning00Blade: :100: I noticed it already working on it.

docs/api/puppeteer.jshandle.remoteobject.md
```diff
@@ -4,16 +4,16 @@ sidebar_label: JSHandle.remoteObject
...
 
-Provides access to the \[Protocol.Runtime.RemoteObject\](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/\#type-RemoteObject) backing this handle.
+Provides access to the \[Protocol.Runtime.RemoteObject\](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/\#type-RemoteObject) OR \[Bidi.Script.RemoteReference\](https://w3c.github.io/webdriver-bidi/\#data-types-reference) backing this handle.
 
 #### Signature:
 
 ```typescript
 class JSHandle {
-  remoteObject(): Protocol.Runtime.RemoteObject;
+  remoteObject(): Protocol.Runtime.RemoteObject | any;
```
OrKoN: let's remove any.

packages/puppeteer-core/src/api/JSHandle.ts
```diff
@@ -0,0 +1,199 @@
...
+   * @remarks
+   * Useful during debugging.
+   */
+  toString(): string {
+    throw new Error('Not implemented');
+  }
+
+  /**
+   * Provides access to the
+   * [Protocol.Runtime.RemoteObject](https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#type-RemoteObject)
```
OrKoN, line 196: let's not mention BiDi in the docs just yet. (let's keep the docs the same).

docs/api/puppeteer.jshandle.getproperty_2.md
```diff
@@ -0,0 +1,25 @@
+---
+sidebar_label: JSHandle.getProperty_2
+---
+
+# JSHandle.getProperty() method
```
OrKoN, line 5: where is this coming from?

packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -47,56 +60,73 @@ const applyOffsetsToQuad = (
...
- *
- * @public
+ * @internal
  */
-
-export class ElementHandle<
+export class CDPElementHandle<
   ElementType extends Node = Element
-> extends JSHandle<ElementType> {
+> extends ElementHandle<ElementType> {
```
OrKoN, line 71: let's add a comment about the trick we have done here.

packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -47,56 +60,73 @@ const applyOffsetsToQuad = (
...
+
+  override async evaluate<
+    Params extends unknown[],
+    Func extends EvaluateFunc<[this, ...Params]> = EvaluateFunc<
+      [this, ...Params]
+    >
+  >(
+    pageFunction: string | Func,
+    ...args: Params
+  ): // @ts-expect-error Circularity here is okay because we only need the return
```
OrKoN, line 114: do you know why this error shows up again after the refactoring?

Lightning00Blade: Not sure, it's kind of strange, I would expect it on the base class as well.

docs/api/puppeteer.elementhandle.md
```diff
@@ -37,8 +37,6 @@ ElementHandle instances can be used as arguments in [Page.$eval()](./puppeteer.p
 
 If you're using TypeScript, ElementHandle takes a generic argument that denotes the type of element the handle is holding within. For example, if you have a handle to a `<select>` element, you can type it as `ElementHandle<HTMLSelectElement>` and you get some nicer type checks.
 
-The constructor for this class is marked as internal. Third-party code should not call the constructor directly or create subclasses that extend the `ElementHandle` class.
```
OrKoN: Let's mark the constructor as internal?

docs/api/puppeteer.jshandle.md
```diff
@@ -16,10 +16,6 @@ Handles can be used as arguments for any evaluation function such as [Page.$eval
 export declare class JSHandle<T = unknown>
 ```
 
-## Remarks
-
-The constructor for this class is marked as internal. Third-party code should not call the constructor directly or create subclasses that extend the `JSHandle` class.
```
OrKoN: Same here, probably it should be internal.

docs/api/puppeteer.jshandle.getproperty_2.md
OrKoN: is it back? :D

Lightning00Blade: Yes, I reverted the change, the type test were failing without the it being there. I am trying to understand why it never showed up before as it existed prior to this PR.

OrKoN (approved):

PR #9635, opened 2023-02-08
fix: update test expectations for next Firefox Nightly changes

whimboo:
Over on https://bugzilla.mozilla.org/show_bug.cgi?id=1786299 I was fixing the error messages that we throw in certain situations when serializing objects by value. Given that Puppeteer expects very specific messages we are now in alignment.

After the patch got merged to mozilla-central the next available Nightly build of Firefox will contain these changes and CI checks for PRs and commits will start failing. This PR fixes those.

Note that right now I still expect failures and we should close / reopen the PR later today or tomorrow when the appropriate Nightly build is available.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -60,18 +60,6 @@
     "expectations": ["FAIL"]
   },
   {
-    "testIdPattern": "[browsercontext.spec] BrowserContext window.open should use parent tab context",
-    "platforms": ["darwin", "win32"],
-    "parameters": ["firefox"],
-    "expectations": ["SKIP"]
-  },
-  {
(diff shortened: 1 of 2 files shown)


whimboo:
@OrKoN while our Nightly build is already out the tests still seems to fail and maybe grab an old Firefox Nightly? As I can see there is a cache for the Firefox binary and probably it needs to be cleared? I don't seem to have the permissions or cannot find the location where to do that. Could you please have a look? Thanks!

whimboo:
@OrKoN cleared the GitHub action caches so a more recent Firefox binary should be used now.

OrKoN (approved):

whimboo:
Please note that a change to `jshandle.spec.ts` was missing in the last version of this PR. I've pushed an update now which should fix across all Firefox jobs.

PR #9639, opened 2023-02-09
chore: implement P queries

jrandolf-google:
This PR implements P queries. See internal doc.


test/src/queryhandler.spec.ts
```diff
@@ -351,4 +352,34 @@ describe('Query handler tests', function () {
       });
     });
   });
+
+  describe('P selectors', function () {
```
OrKoN: let's add tests with multiple query types including custom query handlers. Also, let's add coverage for the waitForSelector user case.

Maybe let's call it Puppeteer selectors in free-form texts?

jrandolf: > Also, let's add coverage for the waitForSelector user case.

This is already covered since P queries will be the default,

> Maybe let's call it Puppeteer Selectors in free-form texts?

I'm not sure. "Puppeteer selectors" is already an overloaded term.

We only need to add coverage for each of the query types and custom query handlers.

test/src/queryhandler.spec.ts
```diff
@@ -351,4 +353,94 @@ describe('Query handler tests', function () {
...
+    it('should work with custom selectors', async () => {
+      const {page} = getTestState();
+      await page.setContent('<div>hello <button>world</button></div>');
+      Puppeteer.registerCustomQueryHandler('div', {
+        queryOne() {
+          return document.querySelector('div');
+        },
+      });
+
+      const element = await page.$(':-p-div()');
```
OrKoN: Let's add a test with a custom selector that has none empty parameters (perhaps once as a string value and once as an any value).

packages/puppeteer-core/src/common/util.ts
```diff
@@ -488,3 +489,65 @@ export function stringifyFunction(expression: Function): string {
...
+ * @internal
+ */
+export async function* deepDescendents(
+  handle: ElementHandle<Node>
+): AwaitableIterable<ElementHandle<Node>> {
+  const walkers = await handle.evaluateHandle(element => {
+    return [document.createTreeWalker(element, NodeFilter.SHOW_ELEMENT)];
+  });
+  while (true) {
+    const handle = await walkers.evaluateHandle(walkers => {
```
OrKoN: does it make the number of cdp calls to equal the number of shadow roots under an element? 

I don't think it's very efficient. Could we move the entirety of the search into the eval function to make it atomic and not so network heavy?

packages/puppeteer-core/src/common/PSelectorParser.ts
```diff
@@ -0,0 +1,154 @@
...
+export type PSelector =
+  | {
+      name: string;
+      value: string;
+    }
+  | CSSSelector;
+
+const PUPPETEER_FUNCTION_TOKEN = /^:-p-([-a-zA-Z_]+)\(/;
+
+class PSelectorParser {
```
OrKoN: let's cover this with a unit-level tests? I think there a few edge cases that might be interesting and that are not covered by e2e.

packages/puppeteer-core/src/common/util.ts
jrandolf: This function is called whenever someone uses a `>>>` combinator. The number of CDP calls whenever used is equal to the number of elements in the elements tree excluding or included descendent trees (depending on the `deep*` function).

For querying a single element (which is most of the time), this is very efficient. It's also efficient if someone doesn't use `>>>`.

We should keep in mind that querying a shadow root is a very powerful, but expensive task. Since the largest use case is querying a single element, this is what we are optimizing for.

packages/puppeteer-core/src/common/PSelectorParser.ts
jrandolf: Could you provide these cases you are curious about? E2E should cover every case.

packages/puppeteer-core/src/common/util.ts
jrandolf: [EDIT] With batching included now, the number of CDP calls is equal to the number of elements / batch size.

OrKoN:
@Lightning00Blade could you help out with the review once it's rebased?

packages/puppeteer-core/src/common/ExecutionContext.ts
```diff
@@ -89,7 +91,9 @@ export class ExecutionContext {
     this._client = client;
     this._world = world;
     this._contextId = contextPayload.id;
-    this._contextName = contextPayload.name;
+    if (contextPayload.name) {
```
Lightning00Blade, line 96: Shouldn't this always exist?

test/TestExpectations.json
```diff
@@ -1635,7 +1635,7 @@
     "testIdPattern": "",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["webDriverBiDi"],
-    "expectations": ["SKIP", "TIMEOUT"]
+    "expectations": ["SKIP", "TIMEOUT", "FAIL"]
```
Lightning00Blade: Why is this change made for BiDi?

packages/puppeteer-core/src/injected/PSelectorParser.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  #scanParameter(): string {
+    const char = this.#input[0];
+    switch (char) {
+      case "'":
+      case '"':
+        this.#input = this.#input.slice(1);
+        const parameter = this.#scanEscapedValueTill(char);
+        if (!this.#input.startsWith(')')) {
+          throw new Error("Expected ')'");
```
Lightning00Blade, line 119: The error here are user facing right? Then we should give a better error messages.

packages/puppeteer-core/src/injected/PQuerySelector.ts
```diff
@@ -0,0 +1,166 @@
...
+    selector = selector.trim();
+    if (selector.length === 0) {
+      throw new Error('The provided selector is empty.');
+    }
+    this.#deepShadowSelectors = parsePSelectors(selector);
+
+    // If there are any empty elements, then this implies the selector has
+    // contiguous combinators (e.g. `>>> >>>>`) or starts/ends with one which we
+    // treat as illegal, similar to existing behavior.
+    if (
```
Lightning00Blade, line 46: I think this should live inside the parser itself.

packages/puppeteer-core/src/injected/PQuerySelector.ts
```diff
@@ -0,0 +1,166 @@
...
+      }
+    }
+
+    for (; this.#selector !== undefined; this.#next()) {
+      const selector = this.#selector;
+      this.elements = AsyncIterableUtil.flatMap(
+        this.elements,
+        async function* (element) {
+          if (typeof selector === 'string') {
+            if (!element.parentElement) {
```
Lightning00Blade, line 95: Nit:  inverse the `if`, imo it's easier to read

packages/puppeteer-core/src/common/ExecutionContext.ts
jrandolf: It may be an empty string which would imply it shouldn't.

test/TestExpectations.json
jrandolf: Because a test failed.

packages/puppeteer-core/src/injected/PQuerySelector.ts
jrandolf: The current way it's set is the hot path.

packages/puppeteer-core/src/injected/PQuerySelector.ts
jrandolf: The Parser doesn't check for semantically accurate selectors, just syntax.

test/TestExpectations.json
OrKoN: How can it fail if it is skipped? 🤔

OrKoN (commented):
let's update the guides with the information about the new selector type https://pptr.dev/guides/query-selectors

jrandolf:
> let's update the guides with the information about the new selector type https://pptr.dev/guides/query-selectors

Discussed offline.

OrKoN (approved):

PR #9644, opened 2023-02-09
refactor: migrate to iterator-based query handlers

jrandolf-google:
This PR migrates our array-based query handlers to iterator-based handlers. This drastically improves the memory footprint and also potentially improves querying speeds for single node selection.

We've also refactored the internal query selector API and removed the `waitFor` implementation of `ARIAQueryHandler` for a 'default implemented' one. This way, upcoming handlers can use the same code.


OrKoN:
Since it's not a user-facing change (right?), let's mark it as a chore.

OrKoN:
Actually, I think we can use `refactor: migrate to iterator-based query handlers`

packages/puppeteer-core/src/common/DOMQueryHandler.ts
```diff
@@ -0,0 +1,130 @@
+/**
+ * Copyright 2022 Google Inc. All rights reserved.
```
Lightning00Blade: Nit: New files should be 2023

packages/puppeteer-core/src/common/Puppeteer.ts
```diff
@@ -13,6 +13,22 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+/**
```
Lightning00Blade: Nit: double License?

packages/puppeteer-core/src/common/XPathQueryHandler.ts
```diff
@@ -0,0 +1,38 @@
+/**
+ * Copyright 2022 Google Inc. All rights reserved.
```
OrKoN: nit: update the license year to 2023 for newly authored files

packages/puppeteer-core/src/common/IterableUtil.ts
```diff
@@ -0,0 +1,76 @@
...
+  static async first<T>(
+    iterable: AwaitableIterable<T>
+  ): Promise<T | undefined> {
+    for await (const value of iterable) {
+      return value;
+    }
+    return undefined;
+  }
+
+  static async *transposeHandle<T>(
```
OrKoN: nit: it looks like this method does not belong to this class. Consider moving it elsewhere so that iterable utils don't depend on JSHandle and HandleFor

packages/puppeteer-core/src/common/IterableUtil.ts
```diff
@@ -0,0 +1,76 @@
...
+    handle: JSHandle<AwaitableIterable<T>>
+  ): AwaitableIterable<HandleFor<T>> {
+    const iterator = await handle.evaluateHandle(iterable => {
+      return (async function* () {
+        yield* iterable;
+      })();
+    });
+    try {
+      while (true) {
+        const handle = await iterator.evaluateHandle(async iterator => {
```
OrKoN: this looks expensive because a CDP call is made for every iteration. Did we do a separate call for every returned element in the previous version? I see that this method is used for all query handlers now.

OrKoN: Also, if I understand it right: with this change the querySelectorAll is not atomic anymore because changes can happen while the iterator is being invoked. I am not sure how big of the effect it might have but I wonder if we can check the following things:

1) benchmark a queryAll with the old way and new way (time-wise)
2) see if we can have the collectAll to be executed in the page context

jrandolf:
> Since it's not a user-facing change (right?), let's mark it as a chore.

Technically is since users can use any iterable now, but in any case, I think we can just mark it as a refactor.

packages/puppeteer-core/src/common/IterableUtil.ts
jrandolf: I put them here since these IterableUtils are puppeteer only.

packages/puppeteer-core/src/common/IterableUtil.ts
jrandolf: > with this change the querySelectorAll is not atomic anymore because changes can happen while the iterator is being invoked

We can't use `collect` in the page context because in the upcoming `P` queries PR, the iterators may not be based on the page context.

> benchmark a query all with the old way and new way (time-wise)

We can actually test this easily surprisingly. We have a test (`should handle many elements`) for excess elements which gives us a time estimate:  For a simple CSS selector of 1000 elements, we have

 - 597ms using iterators,
 - 188ms without it.

However I have an assumption that no one ever queries that many elements repetitively. Also, with the iterator approach, we don't have to use `getProperties` which is a CDP implementation.

OrKoN: What is the time for 2000 elements with and without? Also, let's do a test CL on DevTools to see if there is any flakiness due to querying not being atomic.

jrandolf: If you are wondering if it's linear growth, it is. For DevTools, let me see.

jrandolf: https://chromium-review.googlesource.com/c/devtools/devtools-frontend/+/4241082

jrandolf: I've implemented batching. With batching, the speed is about 300ms which is quite reasonable. I've set the batch sizes to 20 for now.

package.json
```diff
@@ -26,17 +26,92 @@
...
-    "test:chrome:headful": "npm test -- --test-suite chrome-headful",
-    "test:chrome:new-headless": "npm test -- --test-suite chrome-new-headless",
-    "test:chrome:headless": "npm test -- --test-suite chrome-headless",
-    "test:chrome:bidi": "npm test -- --test-suite chrome-bidi",
-    "test:chrome": "run-s test:chrome:*",
-    "test:firefox:bidi": "npm test -- --test-suite firefox-bidi",
-    "test:firefox:headful": "npm test -- --test-suite firefox-headful",
-    "test:firefox:headless": "npm test -- --test-suite firefox-headless",
-    "test:firefox": "run-s test:firefox:*",
+    "test:chrome:headful": "wireit",
```
OrKoN: does passing arguments via `-- --no-coverage` still works with wireit?

packages/puppeteer-core/src/api/JSHandle.ts
```diff
@@ -86,7 +86,10 @@ export class JSHandle<T = unknown> {
   >(
     pageFunction: Func | string,
     ...args: Params
-  ): Promise<Awaited<ReturnType<Func>>>;
+  ): // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
OrKoN: would you mind filing an issue for TS and linking it from here?

OrKoN (commented):
It looks like a test or two are failing

PR #9646, opened 2023-02-10
chore: fix npm scripts

jrandolf-google:
Fixes some dependency looping due to excessive cleaning. Also adds some input files to some scripts since they are depedencies.


packages/puppeteer-core/package.json
```diff
@@ -75,9 +74,6 @@
         "src/generated/**"
       ]
     },
-    "clean:third_party": {
```
OrKoN, line 78: This was needed for incremental builds to work. Could you test the incremental build?

The problem is that we first roll-up and then tsc, on the second build the tsc'ed classes are used in the compilation without re-running rollup. And that led to the compilation errors on the second run of the build. 

The better solution would be to tsc into a temp dir and then rollup back into the target dir. Do you think that could work?

jrandolf: Hmm, I'm not sure what you mean by this. Incremental builds are supported within `wireit` using the `clean: if-file-deleted` option according to their documentation.

OrKoN: I verified it in this branch, the problem is not there anymore. Perhaps because we don't try to bundle bidi anymore. I can explain it over a VC if you want to understand the problem. I think the definition of build:third_party is not correct because the output folder is the input folder at the same time.

jrandolf: Ok, I see what you mean. Let me try something.

OrKoN (commented):
Looks like ng-schematics tests fail. I will let @Lightning00Blade to do the rest of the review.

packages/ng-schematics/package.json
```diff
@@ -22,31 +21,23 @@
...
         "lib/**/files/**",
         "lib/**/*.json"
-      ],
-      "dependencies": [
-        "clean"
       ]
     },
     "build": {
       "command": "tsc -b",
       "files": [
```
Lightning00Blade, line 31: This need to be kept as is. Actually we should do do this for all our packages. If a step does not control part of the output dir we should excluded that part from output. That way wire it can deal properly changes. As at the moment is deleting the output form `build:copy` or some times not creating anything in `lib`.

packages/puppeteer-core/package.json
```diff
@@ -141,18 +142,19 @@
...
-        "**/tsconfig.*.json"
+        "src/**",
+        "third_party/**"
       ],
       "output": [
-        "lib/esm/**",
-        "lib/cjs/**"
+        ".wireit/third_party/**",
+        "lib/cjs/**",
+        "lib/esm/**"
```
Lightning00Blade: I think we should add `.tsbuildinfo` to files, but that can be done in another PR.

packages/puppeteer-core/package.json
```diff
@@ -68,24 +67,24 @@
     "generate:sources": {
       "command": "tsx tools/generate_sources.ts",
       "files": [
-        "tools/generate_sources.ts",
-        "src/templates/**"
+        "../../versions.js",
+        "package.json",
```
Lightning00Blade: Wireit uses `package-lock.json` to generate Hash for builds, why is this needed here?

packages/ng-schematics/package.json
```diff
@@ -3,16 +3,15 @@
   "version": "0.1.0",
   "description": "Puppeteer Angular schematics",
   "scripts": {
-    "dev": "npm run build --watch",
-    "dev:test": "npm run test --watch",
-    "copy": "wireit",
+    "build:copy": "wireit",
```
Lightning00Blade: I think we can Upgrade Wireit and remove these `build:*` steps as it now supports independent scripts in `wireit` config part. We can do it as a follow up PR.

packages/puppeteer-core/package.json
jrandolf: It's unnecessary since changes to the compiled files are what matters. Output should always be about the artifacts using in the later steps (and caching). Anything else is just meta.

packages/ng-schematics/package.json
jrandolf: Let's do this in a follow up.

Lightning00Blade (approved):
LGTM

PR #9647, opened 2023-02-10
chore: introduce @puppeteer/browsers with a fetch method implementation

OrKoN:

This PR introduces a new package that is the refactored version of the BrowserFetcher. The package is not published yet and is not used anywhere. All names including the package name are not final and are subject to change in the next PRs (so no documentation for this as well).

Yes

**Summary**

The new package will encapsulate all the Node-specific code related to the launching/downloading of the browsers (the code that is currently in the node folder of puppeteer-core). It will also simplify and unify the workflow to support more browsers. This PR implements two functions `fetch` and `canFetch`. 

- `fetch`: downloads a specific revision/build from a remote server. It also extracts the download archive if it was compressed.
- `canFetch`: checks if it's possible to download a file for a request revision/build using an HTTP HEAD request.

No

**Other information**

### DIFF
--- a/packages/browsers/package.json
+++ b/packages/browsers/package.json
@@ -1,0 +1,81 @@
+{
+  "name": "@puppeteer/browsers",
+  "version": "0.0.1",
+  "description": "Download and launch browsers",
+  "scripts": {
+    "build": "wireit",
+    "build:test": "wireit",
+    "clean": "tsc --build --clean && rimraf lib",
+    "test": "wireit"
+  },
+  "wireit": {
+    "build": {
+      "command": "tsc -b",
+      "files": [
+        "src/**/*.ts",
+        "tsconfig.json"
+      ],
+      "output": [
+        "lib/**"
+      ]
+    },
+    "build:test": {
+      "command": "tsc -b test/src/tsconfig.json",
+      "files": [
+        "test/**/*.ts",
+        "test/src/tsconfig.json"
+      ],
+      "output": [
+        "test/build/**"
+      ],
+      "dependencies": [
+        "build"
+      ]
+    },
+    "test": {
+      "command": "mocha ",
+      "files": [
+        ".mocharc.cjs"
+      ],
+      "dependencies": [
+        "build:test"
+      ]
+    }
+  },
+  "keywords": [
+    "puppeteer",
+    "browsers"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/puppeteer/puppeteer/tree/main/packages/browsers"
+  },
+  "author": "The Chromium Authors",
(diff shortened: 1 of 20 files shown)


packages/browsers/package.json
```diff
@@ -0,0 +1,81 @@
...
+      ],
+      "output": [
+        "test/build/**"
+      ],
+      "dependencies": [
+        "build"
+      ]
+    },
+    "test": {
+      "command": "mocha ",
```
Lightning00Blade:
```suggestion
      "command": "mocha",
```

packages/browsers/package.json
```diff
@@ -0,0 +1,81 @@
...
+  "dependencies": {
+    "debug": "4.3.4",
+    "extract-zip": "2.0.1",
+    "https-proxy-agent": "5.0.1",
+    "proxy-from-env": "1.1.0",
+    "tar-fs": "2.1.1",
+    "unbzip2-stream": "1.4.3"
+  },
+  "devDependencies": {
+    "@types/node": "^14.15.0"
```
Lightning00Blade, line 71: We need this explicitly here as we use `@type/node: 18.x.x` in main right? I think we should keep that in sync with what `package.json > engines.node` says in there as well.

packages/browsers/src/fetch.ts
```diff
@@ -0,0 +1,168 @@
+/**
+ * Copyright 2017 Google Inc. All rights reserved.
```
Lightning00Blade, line 2: Nit: Check files have proper year. I think its only here.

packages/browsers/src/fetch.ts
```diff
@@ -0,0 +1,171 @@
...
+  path: string;
+  browser: Browser;
+  revision: string;
+  platform: BrowserPlatform;
+};
+
+export async function fetch(options: Options): Promise<InstalledBrowser> {
+  options.platform ??= detectPlatform();
+  if (!options.platform) {
+    throw new Error(
```
Lightning00Blade, line 69: Should we try to verify that the input for `options.platform` is supported?

Lightning00Blade (approved):
LGTM + some minor comments.

packages/browsers/src/fetch.ts
OrKoN: That was derived from the BrowserFetcher so keeping the header intact.

packages/browsers/src/fetch.ts
OrKoN: You mean except for the TypeScript?

OrKoN: I think we should rely on TS on this level. We can validate it in the CLI.

PR #9651, opened 2023-02-11
refactor: implement reverse argument binding

jrandolf-google:
This PR implements reverse binding and refactors the binding code within Puppeteer.

Reverse argument bindings allow complex objects such as `Node`s to be passed from the client to Puppeteer (similar to how we can pass handles to the client). For now, we only install `Node` argument bindings for ARIA queries, but more can be installed as use-cases come up.

### DIFF
--- a/packages/puppeteer-core/src/common/Binding.ts
+++ b/packages/puppeteer-core/src/common/Binding.ts
@@ -1,0 +1,119 @@
+import {JSHandle} from '../api/JSHandle.js';
+import {isErrorLike} from '../util/ErrorLike.js';
+import {ExecutionContext} from './ExecutionContext.js';
+import {debugError} from './util.js';
+
+/**
+ * @internal
+ */
+export class Binding {
+  #name: string;
+  #fn: (...args: unknown[]) => unknown;
+  constructor(name: string, fn: (...args: unknown[]) => unknown) {
+    this.#name = name;
+    this.#fn = fn;
+  }
+
+  /**
+   *
+   * @param context - Context to run the binding in; the context should have
+   * the binding added to it beforehand.
+   * @param id - ID of the call. This should come from the CDP
+   * `onBindingCalled` response.
+   * @param args - Plain arguments from CDP.
+   */
+  async run(
+    context: ExecutionContext,
+    id: number,
+    args: unknown[],
+    isTrivial: boolean
+  ): Promise<void> {
+    const garbage = [];
+    try {
+      if (!isTrivial) {
+        // Getting non-trivial arguments.
+        const handles = await context.evaluateHandle(
+          (name, seq) => {
+            // @ts-expect-error Code is evaluated in a different context.
+            return globalThis[name].args.get(seq);
+          },
+          this.#name,
+          id
+        );
+        try {
+          const properties = await handles.getProperties();
+          for (const [index, handle] of properties) {
+            // This is not straight-forward since some arguments can stringify, but
+            // aren't plain objects so add subtypes when the use-case arises.
+            if (index in args) {
+              switch (handle.remoteObject().subtype) {
+                case 'node':
+                  args[+index] = handle;
+                  break;
+                default:
+                  garbage.push(handle.dispose());
+              }
+            } else {
+              garbage.push(handle.dispose());
+            }
+          }
+        } finally {
+          await handles.dispose();
+        }
+      }
+
+      await context.evaluate(
+        (name, seq, result) => {
+          // @ts-expect-error Code is evaluated in a different context.
+          const callbacks = globalThis[name].callbacks;
+          callbacks.get(seq).resolve(result);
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -26,17 +26,92 @@
...
-    "test:chrome:headful": "npm test -- --test-suite chrome-headful",
-    "test:chrome:new-headless": "npm test -- --test-suite chrome-new-headless",
-    "test:chrome:headless": "npm test -- --test-suite chrome-headless",
-    "test:chrome:bidi": "npm test -- --test-suite chrome-bidi",
-    "test:chrome": "run-s test:chrome:*",
-    "test:firefox:bidi": "npm test -- --test-suite firefox-bidi",
-    "test:firefox:headful": "npm test -- --test-suite firefox-headful",
-    "test:firefox:headless": "npm test -- --test-suite firefox-headless",
-    "test:firefox": "run-s test:firefox:*",
+    "test:chrome:headful": "wireit",
```
OrKoN: This change shows in multiple PRs it seems. It's better be moved in a separate PR as it does not seem to be directly related to the goal of this change.

OrKoN:
What is a reverse argument binding and why do we need it?

jrandolf:
> What is a reverse argument binding and why do we need it?

I've added a description within the PR now.

packages/puppeteer-core/src/common/Binding.ts
```diff
@@ -0,0 +1,107 @@
...
+export class Binding {
+  #name: string;
+  #fn: (...args: unknown[]) => unknown;
+  constructor(name: string, fn: (...args: unknown[]) => unknown) {
+    this.#name = name;
+    this.#fn = fn;
+  }
+
+  /**
+   * @param throwError - Whether to throw the error instead of sending it to the
```
OrKoN: outdated comment?

packages/puppeteer-core/src/common/Binding.ts
```diff
@@ -0,0 +1,107 @@
...
+              const callbacks = globalThis[name].callbacks;
+              callbacks.get(seq).resolve(result);
+              callbacks.delete(seq);
+            },
+            this.#name,
+            seq,
+            await this.#fn(...args)
+          );
+        } finally {
+          for (const [, handle] of properties) {
```
OrKoN: do we need to dispose sequentially?

packages/puppeteer-core/src/common/types.ts
```diff
@@ -14,9 +14,19 @@
...
+import type {ElementHandle} from '../api/ElementHandle.js';
+import type {LazyArg} from './LazyArg.js';
+
+/**
+ * @internal
+ */
+export type BindingPayload = {
+  type: string;
+  name: string;
+  seq: number;
```
OrKoN, line 27: what does seq mean? Pls add a comment.

OrKoN: I guess is the id of a callback/promise?

packages/puppeteer-core/src/common/Binding.ts
```diff
@@ -0,0 +1,107 @@
...
+  constructor(name: string, fn: (...args: unknown[]) => unknown) {
+    this.#name = name;
+    this.#fn = fn;
+  }
+
+  /**
+   * @param throwError - Whether to throw the error instead of sending it to the
+   * binding.
+   */
+  async call(
```
OrKoN: nit: perhaps the name of this method could be `onCall` or smth like this to indicate that the binding is not invoked here but it is the logic that happens after the page called the binding? Although, I could see how `call` could reflect that but I think it is also more confusing.

packages/puppeteer-core/src/common/Binding.ts
```diff
@@ -0,0 +1,107 @@
...
+    context: ExecutionContext,
+    seq: number,
+    args: unknown[]
+  ): Promise<void> {
+    try {
+      // Getting non-trivial arguments.
+      const handles = await context.evaluateHandle(
```
OrKoN: so we do those extra calls on every binding call even if there were no non-primitive arguments?

jrandolf: Done.

packages/puppeteer-core/src/common/Binding.ts
```diff
@@ -0,0 +1,118 @@
...
+              const callbacks = globalThis[name].callbacks;
+              callbacks.get(seq).reject(error);
+              callbacks.delete(seq);
+            },
+            this.#name,
+            id,
+            error.message,
+            error.stack
+          )
+          .catch(() => {});
```
OrKoN: how often can it fail? let's handle it through the debugError helper, just in case if smth stops working and we need to debug. The same below.

OrKoN (approved):
lgtm % debugError suggestion

PR #9652, opened 2023-02-11
docs: update screenshot type formats

azat-io:
Puppeteer supports `webp` format:
https://github.com/puppeteer/puppeteer/blob/b8ea891ab86cb5d21c173ed9cc72f6dd2a6e28d6/packages/puppeteer-core/src/api/Page.ts#L152


Lightning00Blade:
Hello @azat-io, thanks for the PR. I can see that the types is already there but the TSDoc is not updated. You should make the change in puppeteer-core/src/api/Page.ts Page@screenshot, as the docs are auto-generated. Then you can run `npm run docs` and it should generate the correct files.

PR #9659, opened 2023-02-13
chore: use wireit from workspace scripts

jrandolf-google:


package.json
```diff
@@ -26,17 +26,92 @@
...
+    "test:chrome:headful": {
+      "command": "npm test -- --test-suite chrome-headful",
+      "dependencies": [
+        "./test:build"
+      ]
+    },
+    "test:chrome:headless": {
+      "command": "npm test -- --test-suite chrome-headless",
+      "dependencies": [
+        "./test:build"
```
OrKoN, line 66: we keep changing whether the test command will build the tests or not. Let's decide for one or another way? If the tests do cause a rebuild now, let's update existing documentation to reflect that (README, contributing guides).

jrandolf: The tests can build now because of wireit. Mind if we update the docs in another CL? I think it makes sense to dogfood this for a bit before settling.

OrKoN: alternatively, let's keep the test commands to be without dependencies so that no automatic rebuild will happen.

OrKoN: I'd say let's update the docs now and if we revert it, let's update the docs again.

OrKoN: summarising the discussion offline: we will try the new commands for some time and update the docs later.

OrKoN (approved):

Lightning00Blade (approved):

PR #9678, opened 2023-02-14
refactor: custom query handlers and global bindings

jrandolf-google:
This PR refactors the custom query handling API and the global bindings API.

## Background

While implementing `P` queries (https://github.com/puppeteer/puppeteer/pull/9639) on the client-side, there were two problems:

 1. ARIA selector bindings must always be installed.
 2. Custom query selectors must somehow be available within the page.

## What's changed?

In order to facilitate (1), we will install the ARIA selector binding whenever the PuppeteerUtil is used. This is safe since the binding will only exist within Puppeteer's isolated world. This also obsoletes a lot of code (not removed in this PR for brevity) related to `WaitTask` bindings that will require refactoring after this PR.

For (2), this PR now allows restricted modifications of the Puppeteer utilities script at runtime through the `ScriptInjector`. This lets us install custom query selectors into the page through the `PuppeteerUtil.customQuerySelectors` registry.

### DIFF
--- a/packages/puppeteer-core/src/common/ExecutionContext.ts
+++ b/packages/puppeteer-core/src/common/ExecutionContext.ts
@@ -15,18 +15,20 @@
  */
 
 import {Protocol} from 'devtools-protocol';
-import {source as injectedSource} from '../generated/injected.js';
+import {JSHandle} from '../api/JSHandle.js';
 import type PuppeteerUtil from '../injected/injected.js';
+import {stringifyFunction} from '../util/Function.js';
+import {ARIAQueryHandler} from './AriaQueryHandler.js';
+import {Binding} from './Binding.js';
 import {CDPSession} from './Connection.js';
 import {IsolatedWorld} from './IsolatedWorld.js';
-import {JSHandle} from '../api/JSHandle.js';
 import {LazyArg} from './LazyArg.js';
+import {scriptInjector} from './ScriptInjector.js';
 import {EvaluateFunc, HandleFor} from './types.js';
 import {
   createJSHandle,
   getExceptionMessage,
   isString,
-  stringifyFunction,
   valueFromRemoteObject,
 } from './util.js';
 import {CDPJSHandle} from './JSHandle.js';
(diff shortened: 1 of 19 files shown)


packages/puppeteer-core/src/common/Puppeteer.ts
```diff
@@ -58,57 +52,51 @@ export interface ConnectOptions extends BrowserConnectOptions {
...
-   * to register.
+   * Operations for {@link CustomQueryHandler | custom query handlers}. See
+   * {@link CustomQueryHandlerRegistry}.
    *
    * @public
    */
+  static customQueryHandlers = customQueryHandlers;
+
+  /**
+   * @deprecated Use {@link CustomQueryHandlerRegistry.register |
```
OrKoN: do we need to deprecate this? Imo, it's better than `Puppeteer.customQueryHandlers.register`

packages/puppeteer-core/src/common/ExecutionContext.ts
```diff
@@ -92,16 +94,35 @@ export class ExecutionContext {
...
-    if (!this.#puppeteerUtil) {
-      this.#puppeteerUtil = this.evaluateHandle(
-        `(() => {
-            const module = {};
-            ${injectedSource}
-            return module.exports.default;
-          })()`
-      ) as Promise<JSHandle<PuppeteerUtil>>;
+    scriptInjector.inject(async script => {
+      if (this.#puppeteerUtil) {
```
OrKoN, line 100: looks like potential rejections here would be unhandled promise rejections. let's make sure all promises are caught here.

packages/puppeteer-core/src/util/Function.ts
```diff
@@ -33,3 +33,69 @@ export const createFunction = (
...
+ * ```
+ *
+ * @internal
+ */
+export const interpolateFunction = <T extends (...args: never[]) => unknown>(
+  fn: T,
+  replacements: Record<string, string>
+): T => {
+  let value = stringifyFunction(fn);
+  for (const replacement in replacements) {
```
OrKoN: nit: `for (const [placeholder, value] of replacements.entries()) {}`

OrKoN (changes requested):
Let's rollback the deprecations and removals from this PR and make a decision separately. I think deprecating the current API is not necessary because it gives us more flexibility. Also, removals would have to be marked as a breaking change.

packages/puppeteer-core/src/common/Puppeteer.ts
jrandolf: `Puppeteer.customQueryHandlers.register` is meant to mimic `customElements`. I think namespacing this kind of stuff on the global object lets us be more flexible.

packages/puppeteer-core/src/common/ExecutionContext.ts
jrandolf: The promise is always handled since wherever the getter is used. (The rejection logic remains the same as before.)

OrKoN (approved):

PR #9681, opened 2023-02-15
refactor: unify working with the cache folder structure

OrKoN:
A new class called CacheStructure is used to encode the cache dir structure. Eventually, it might become the CacheStorage class managing the entire cache dir. It is to be decided later if the fetch/launch methods should be aware of the cache storage structure. For the moment, they are. The executable path resolution helpers are made relative in this CL and this means that are not aware of the cache dir structure.

### DIFF
--- a/packages/browsers/src/CacheStructure.ts
+++ b/packages/browsers/src/CacheStructure.ts
@@ -1,0 +1,52 @@
+/**
+ * Copyright 2023 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Browser, BrowserPlatform} from './browsers/types.js';
+import path from 'path';
+
+/**
+ * The cache used by Puppeteer relies on the following structure:
+ *
+ * - rootDir
+ *   -- <browser1> | browserRoot(browser1)
+ *   ---- <platform>-<revision> | installationDir()
+ *   ------ the browser-platform-revision
+ *   ------ specific structure.
+ *   -- <browser2> | browserRoot(browser2)
+ *   ---- <platform>-<revision> | installationDir()
+ *   ------ the browser-platform-revision
+ *   ------ specific structure.
+ *   @internal
+ */
+export class CacheStructure {
+  #rootDir: string;
+
+  constructor(rootDir: string) {
+    this.#rootDir = rootDir;
+  }
+
+  browserRoot(browser: Browser): string {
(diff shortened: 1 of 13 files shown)


packages/browsers/test/src/fetch.spec.ts
```diff
@@ -36,13 +36,13 @@ describe('fetch', () => {
   });
 
   afterEach(() => {
-    fs.rmSync(tmpDir, {recursive: true});
+    // fs.rmSync(tmpDir, {recursive: true});
```
Lightning00Blade: I think this was for testing.

packages/browsers/src/browsers/firefox.ts
```diff
@@ -38,26 +38,18 @@ export function resolveDownloadUrl(
...
-        browserPath,
-        'Firefox Nightly.app',
-        'Contents',
-        'MacOS',
-        'firefox'
-      );
+      return path.join('Firefox Nightly.app', 'Contents', 'MacOS', 'firefox');
     case BrowserPlatform.LINUX:
-      return path.join(browserPath, 'firefox', 'firefox');
+      return path.join('firefox', 'firefox');
```
Lightning00Blade, line 50: Why do we put it in a subfolder for firefox and not for chrome?

packages/browsers/src/browsers/chrome.ts
```diff
@@ -57,26 +57,18 @@ export function resolveDownloadUrl(
   )}.zip`;
 }
 
-export function executablePath(
+export function relativeExecutablePath(
   platform: BrowserPlatform,
-  revision: string,
-  basePath = ''
+  _revision: string
```
Lightning00Blade, line 62: Why keep here?

Lightning00Blade (approved):
LGTM + some questions

packages/browsers/src/browsers/chrome.ts
OrKoN: The main reason is that the binary path might change based on the revision which is currently not used but if we expose it on the API, it might be useful in the future.

packages/browsers/src/CacheStructure.ts
```diff
@@ -0,0 +1,52 @@
...
+ *   ---- <platform>-<revision> | installationDir()
+ *   ------ the browser-platform-revision
+ *   ------ specific structure.
+ *   -- <browser2> | browserRoot(browser2)
+ *   ---- <platform>-<revision> | installationDir()
+ *   ------ the browser-platform-revision
+ *   ------ specific structure.
+ *   @internal
+ */
+export class CacheStructure {
```
jrandolf, line 34: Can we call this just `Cache`? `CacheStructure` is redundant. If `Cache` is too generic, let's do `BrowserCache`

packages/browsers/test/src/fetch.spec.ts
OrKoN: Updated, thanks!

packages/browsers/src/browsers/firefox.ts
OrKoN: That's how the Firefox archive is structured I believe:

```
~/.cache/puppeteer/firefox/linux-112.0a1/firefox/firefox
```

packages/browsers/src/CacheStructure.ts
OrKoN: Thanks for the suggestion, gonna rename it in the next PRs. BrowserCache sounds good to me and I plan to add some helpers for working with the cache directory.

PR #9693, opened 2023-02-16
fix: add filter to setDiscoverTargets for Firefox

CanadaHonk:

Bugfix.

No, existing tests cover change.

**Summary**

The filter option will likely be soon supported by Firefox, and the browser type will be filtered out by default. So the same filter as Chrome should be supplied to include all wanted targets to avoid a regression.

No.

**Other information**

Firefox patch: https://phabricator.services.mozilla.com/D167209

### DIFF
--- a/packages/puppeteer-core/src/common/FirefoxTargetManager.ts
+++ b/packages/puppeteer-core/src/common/FirefoxTargetManager.ts
@@ -168,7 +168,7 @@ export class FirefoxTargetManager
   async initialize(): Promise<void> {
     await this.#connection.send('Target.setDiscoverTargets', {
       discover: true,
-      filter: [{type: 'tab', exclude: true}, {}],
+      filter: [{}],
     });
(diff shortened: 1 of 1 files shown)


OrKoN:
Thanks for the PR. I didn't know Firefox implements a tab target: does it really exist? I have never seen it in the logs.

CanadaHonk:
It doesn't exist but I thought it would be best if they used the same or if it was possibly introduced in the future.

OrKoN:
I would say let's remove the tab target from the filter to avoid confusion about its existence and let's land this after the Firefox change lands into the Nightly builds. If Firefox ever supports a tab target, we can add it back.

CanadaHonk:
Sure, sounds good. (BTW: this change is fine on older Firefox versions, as the extra parameter is just ignored)

CanadaHonk:
Updated filter.

OrKoN (approved):

jrandolf-google (approved):

OrKoN:
Let's land after the Firefox change appears in the Nightly build.

PR #9700, opened 2023-02-17
chore: Add Page.Console event to BiDi

Lightning00Blade:
Implements Page.Console event for BiDi connection.

### DIFF
--- a/packages/puppeteer-core/src/common/bidi/Page.ts
+++ b/packages/puppeteer-core/src/common/bidi/Page.ts
@@ -14,27 +14,76 @@
  * limitations under the License.
  */
 
-import {Page as PageBase} from '../../api/Page.js';
+import * as Bidi from 'chromium-bidi/lib/cjs/protocol/protocol.js';
+
+import {Page as PageBase, PageEmittedEvents} from '../../api/Page.js';
 import {stringifyFunction} from '../../util/Function.js';
+import {ConsoleMessage, ConsoleMessageLocation} from '../ConsoleMessage.js';
 import type {EvaluateFunc, HandleFor} from '../types.js';
 import {isString} from '../util.js';
 
 import {Connection} from './Connection.js';
 import {JSHandle} from './JSHandle.js';
 import {BidiSerializer} from './Serializer.js';
-import {Reference} from './types.js';
 
 /**
  * @internal
  */
 export class Page extends PageBase {
   #connection: Connection;
+  #subscribedEvents = [
+    'log.entryAdded',
+  ] as Bidi.Session.SubscribeParameters['events'];
   _contextId: string;
 
   constructor(connection: Connection, contextId: string) {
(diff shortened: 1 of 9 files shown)


test/src/page.spec.ts
```diff
@@ -690,10 +687,9 @@ describe('Page', function () {
         messages.map(msg => {
           return msg.type();
         })
-      ).toEqual(['timeEnd', 'trace', 'dir', 'warning', 'error', 'log']);
-      expect(messages[0]!.text()).toContain('calling console.time');
+      ).toEqual(['trace', 'dir', 'warning', 'error', 'log']);
```
Lightning00Blade, line 690: In BiDi we have level and method -> Level is like a group for methods, while methods itself is the called method on `console`. The `warning` fails here as we receive `warn`.

test/TestExpectations.json
```diff
@@ -1763,24 +1763,12 @@
     "parameters": ["webDriverBiDi"],
     "expectations": ["PASS"]
   },
-  {
```
Lightning00Blade, line 1766: Result of using creating `new JSHandle` when we expect `new ElementHandle`. That way we fail only test that really failing.

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -14,27 +14,76 @@
...
     super();
     this.#connection = connection;
     this._contextId = contextId;
+
+    this.connection.send('session.subscribe', {
+      events: this.#subscribedEvents,
+      contexts: [this._contextId],
+    });
+
+    this.connection.on('log.entryAdded', this.#onLogEntryAdded.bind(this));
```
OrKoN, line 50: TODO: we need to unsubscribe at some point

Lightning00Blade: From my understanding we don't need to unsubscribe per BiDi. Also in Puppeteer we don't have a use case for it either (looking at CDP implementation).

OrKoN: I believe we need to do that if the page is closed. In the CDP implementation, the session would be detached when the page closes but in BiDi we have only a central session so the listeners will cause a memory leak (I believe).

OrKoN (approved):

packages/puppeteer-core/src/common/bidi/Page.ts
Lightning00Blade: I don't think it will cause a memory leak, there are no should be no references stored on the BiDi side as all RemoteValue have `root: none` when being send. Do you think there may another thing that can cause this?

OrKoN: I don't think it has to do with the BiDi references. Let's chat on Monday to clarify this.

PR #9706, opened 2023-02-18
test: add fetch tests with a proxy server

OrKoN:


packages/browsers/src/httpUtil.ts
```diff
@@ -49,6 +49,7 @@ export function httpRequest(
     hostname: url.hostname,
     port: url.port,
     path: url.pathname,
+    method,
```
OrKoN, line 52: two bugs found :D

Lightning00Blade (approved):

PR #9708, opened 2023-02-18
chore: resolve Firefox revisions for the latest keyword

OrKoN:


packages/browsers/test/src/cli.spec.ts
```diff
@@ -72,4 +72,15 @@ describe('CLI', function () {
       )
     );
   });
+
+  it('should download latest Firefox binaries', async () => {
+    await new CLI(tmpDir).run([
```
OrKoN, line 77: need to think of a way to return data from the CLI class so that the actual revision can be retrieved.

Lightning00Blade (approved):

PR #9717, opened 2023-02-21
chore: implement operations

jrandolf-google:
This PR adds operations which are a new `PromiseLike` object (actually they _are_ `Promise`s) that allow declarative side effects through the `effect` method.

See tests for usage.

### DIFF
--- a/packages/puppeteer-core/src/common/Operation.ts
+++ b/packages/puppeteer-core/src/common/Operation.ts
@@ -1,0 +1,153 @@
+/**
+ * Copyright 2023 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type {Awaitable} from './types.js';
+
+/**
+ * Operations are promises that can have <a
+ * href="https://en.wikipedia.org/wiki/Side_effect_(computer_science)">effects</a>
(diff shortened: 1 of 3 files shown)


OrKoN (changes requested):
Let's not land anything until we finalised the design.

packages/puppeteer-core/src/common/Operation.ts
```diff
@@ -0,0 +1,131 @@
...
+   * @param effect - The effect to add.
+   * @returns `this` for chaining.
+   */
+  effect(effect: () => Awaitable<void>): this {
+    if (this.#completed) {
+      throw new Error('Cannot add effect to a completed operation.');
+    }
+    try {
+      this.#effects.push(effect());
+    } catch (error) {
```
OrKoN: why don't we throw in a sync way when effect is called? async errors won't be caught by this try/catch.

test/src/operation.spec.ts
```diff
@@ -0,0 +1,192 @@
+import expect from 'expect';
+import {Operation} from 'puppeteer-core/internal/common/Operation.js';
```
OrKoN, line 18: it should be possible to import from `puppeteer-core` now

packages/puppeteer-core/src/common/Operation.ts
```diff
@@ -0,0 +1,131 @@
+/* eslint-disable tsdoc/syntax */
```
OrKoN: I guess this needs to be removed before landing?

packages/puppeteer-core/src/common/Operation.ts
```diff
@@ -0,0 +1,131 @@
...
+ *
+ * ```ts
+ * const operation = await input.click();
+ * await operation.effect(() => console.log('Works!')); // This will throw because of (2).
+ * ```
+ *
+ * Tl;dr, effects **must** be added synchronously (no `await` statements between
+ * the time the operation is created and the effect is added).
+ *
+ * @public
```
OrKoN: let's not land as public before we can finalise the entire feature.

OrKoN (approved):

test/src/operation.spec.ts
jrandolf: The `create` method is an internal only method.

PR #9730, opened 2023-02-22
chore: Fix non Puppeteer releases

Lightning00Blade:
If we try to publish a package that does not include a Puppeteer release the doc versioning fails.

Drive-by: Update GitHub Action

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -6,14 +6,40 @@ on:
       - release-please-*
 
 jobs:
+  check-changes:
+    name: Check which packages changed
+    runs-on: ubuntu-latest
+    outputs:
+      changes: ${{ steps.changes.outputs.changes }}
+    steps:
+      - name: Check out repository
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-release.yml
```diff
@@ -6,14 +6,40 @@ on:
...
+              - 'docker/**'
+              - 'test/**'
+              - 'test-d/**'
+              - 'tools/mochaRunner/**'
+
   pre-release:
-    if: "startsWith(github.event.head_commit.message, 'chore: release main')"
+    if: |
+      startsWith(github.event.head_commit.message, 'chore: release main') &&
+      contains(fromJSON(needs.check-changes.outputs.changes), 'puppeteer')
```
OrKoN, line 36: Could there be a docs change even if the Puppeteer code has not changed?

OrKoN: I guess not

OrKoN (approved):

PR #9731, opened 2023-02-22
fix: permissions for the browser CLI

OrKoN:
Drive-by: show help when no commands are providerd.

### DIFF
--- a/packages/browsers/package.json
+++ b/packages/browsers/package.json
@@ -13,7 +13,7 @@
   },
   "wireit": {
     "build": {
-      "command": "tsc -b",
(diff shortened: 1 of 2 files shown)


packages/browsers/package.json
```diff
@@ -13,7 +13,7 @@
   },
   "wireit": {
     "build": {
-      "command": "tsc -b",
+      "command": "tsc -b && chmod a+x lib/cjs/browsers.js lib/esm/browsers.js",
```
Lightning00Blade, line 16: I think this does not work on windows, but lets leave it for now.

Lightning00Blade (approved):

PR #9733, opened 2023-02-22
fix: define options per command

OrKoN:
Otherwise, yargs will merge options and duplicate the possible choices.


Lightning00Blade (commented):
I think we should only have one path as well, still one after launch command

OrKoN:
@Lightning00Blade actually, I think I still got it wrong and the correct way is to define params via the command builder interface.

Lightning00Blade (approved):

PR #9734, opened 2023-02-22
chore: report sizes in MB rather than MiB

mathiasbynens:
This patch reports sizes in megabytes (MB) rather than mebibytes (MiB), which seemed to be the intention based on the naming in the code. See https://en.wikipedia.org/wiki/Byte#Multiple-byte_units.

This is also consistent with the way sizes are reported in DevTools across browsers: https://goo.gle/devtools-si

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -163,8 +163,8 @@ export class CLI {
   }
 
   #toMegabytes(bytes: number) {
-    const mb = bytes / 1024 / 1024;
-    return `${Math.round(mb * 10) / 10} Mb`;
+    const mb = bytes / 1000 / 1000;
+    return `${Math.round(mb * 10) / 10} MB`;
   }
 
   #makeProgressCallback(browser: Browser, buildId: string) {


tools/mochaRunner/src/test.ts
```diff
@@ -15,6 +15,7 @@
  */
 
 import assert from 'assert/strict';
+
```
mathiasbynens: The pre-commit lint hook forced me to make this (unrelated) change:

```
~/projects/puppeteer/tools/mochaRunner/src/test.ts
  17:1  error  There should be at least one empty line between import groups  import/order
```

OrKoN (approved):

tools/mochaRunner/src/test.ts
OrKoN: That's weird, I don't get this locally and CI also thinks there should be no empty line.

mathiasbynens: I might have an outdated local setup (old deps?). Will remove it here.

mathiasbynens, line 18:
```suggestion
import assert from 'assert/strict';
```

PR #9742, opened 2023-02-23
chore: extract context to allow emitting only to it

Lightning00Blade:


jrandolf-google (approved):

jrandolf-google (changes requested):
Do we need tests for this?

Lightning00Blade:
> Do we need tests for this?
What part do you think may need separate tests? ( I think the only part we should test is the BiDiSerializer)
I think having seeing that the Test Expecatation are the same is OK for now.

jrandolf-google (approved):

PR #9769, opened 2023-03-03
chore: support commas in P selectors

jrandolf-google:
This PR adds support for P selector lists. It also uses [parsel](https://github.com/leaverou/parsel) for parsing as it's more versatile then our current parser.


Lightning00Blade (approved):
LGTM + this is still internal so happy to merge, but we will need `parsel-js` in `third_party` eventually, so a comment or an issue would be nice.

jrandolf:
Actually for this case, `third_party` isn't necessary. The reason is that the dependency involved (parsel-js) is used for in-page code meaning parsel-js will be bundled automatically during source-code generation. This is wh

OrKoN:
How big is the parsel after it is bundled? since we will keep sending it per-context. How complete is the parser? Are there WPT tests? how well it compares to other parsers?

test/src/queryhandler.spec.ts
```diff
@@ -497,5 +488,11 @@ describe('Query handler tests', function () {
       });
       expect(value).toMatchObject({textContent: 'world', tagName: 'BUTTON'});
     });
+
+    it('should work with commas', async () => {
+      const {page} = getTestState();
+      const elements = await page.$$('div, ::-p-text(world)');
```
OrKoN, line 494: what takes precedence over comma? e.g., how is the following case treated `div, p >>> ::-p-text(world)`

packages/puppeteer-core/package.json
```diff
@@ -150,5 +150,9 @@
     "typescript": {
       "optional": true
     }
+  },
+  "devDependencies": {
+    "mitt": "3.0.0",
+    "parsel-js": "github:jrandolf/parsel"
```
OrKoN, line 156: Can we upstream the changes to parsel?

test/src/queryhandler.spec.ts
jrandolf: div  >>> ::-p-text(world), p >>> ::-p-text(world)

jrandolf: :is(div, p)

OrKoN: Let's add tests:

1) div, p >>> ::-p-text(world)
2) div, p the right order

jrandolf:
> How big is the parsel after it is bundled? since we will keep sending it per-context.

It's tree-shaked, so only the logic we need is used. It's small.

> How complete is the parser?

It's fully compliant to Selector Module 4.

> Are there WPT tests?

As discussed offline, there are, but they aren't comprehensive. We can add more tests if issues crop up. In general, we try to test exhaustively.

> How well it compares to other parsers?

Currently the smallest and robust parser out there.

packages/puppeteer-core/package.json
jrandolf: We are currently doing this.

PR #9775, opened 2023-03-06
chore: remove rimraf

jrandolf-google:


Lightning00Blade (changes requested):
Sadly I don't think we can do this just yet as the `node:fs@rm` is there from `v14.14.0` and we support `v14.1.0` still.

Lightning00Blade (approved):

PR #9776, opened 2023-03-06
chore: update glob

jrandolf-google:


package.json
```diff
@@ -125,7 +125,7 @@
     "@rollup/plugin-node-resolve": "15.0.1",
     "@types/debug": "4.1.7",
     "@types/diff": "5.0.2",
-    "@types/glob": "8.0.0",
+    "@types/glob": "8.1.0",
```
Lightning00Blade: Not needed I think Glob is now written in TypeScript.

Lightning00Blade (approved):
Note: Glob `v9` needs Node 16, but its use are internal only so as long as the test pass we should be good.

jrandolf:
Glob v9 also supports the latest node 14 so we are safe.

PR #9777, opened 2023-03-06
chore: remove pre-push and pre-commit

jrandolf-google:
These are unnecessary as we check these at PR time.


.husky/pre-push
```diff
@@ -1,4 +0,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-npm run lint && npm run check
```
Lightning00Blade, line 4: In my opinion we should keep the pre-push one. It's useful as we have the `Inspect code` Action separated from the reset so if it fails the other are still run which is sub-optimal. We could make the `Inspect Code` a parent for `Changed Packages` and then we could skip the pre-push.

jrandolf: These operations are too expensive to keep at this level. In theory, if these operations were instantaneous, I would agree, but they aren't. They cost a lot of time.

Lightning00Blade: Let's add in `ci.yml` for the job `check-changes` `needs: inspect-code` . So we only run test after the code is checked. The other option is to implement cancelling like described [here](https://github.com/community/community/discussions/38361).

Lightning00Blade (approved):

PR #9782, opened 2023-03-06
chore(browser): remove unused detach function

kblok:
The `_detach ` function is not being used anymore. Is that ok?


OrKoN:
Thanks @kblok Let me check if it's potentially a bug!

OrKoN:
I think this was my original intention to mirror the _attach function to clean up event listeners on disconnect https://github.com/puppeteer/puppeteer/pull/9807 It is probably a bit redundant but I think it's generally better to explicitly remove listeners instead of relying on gc.

Lightning00Blade:
Closed by #9807

PR #9797, opened 2023-03-07
fix: change the install output to include the executable path

OrKoN:
With this change, the executable path will be printed to stdout in the following format: `$browser@$revision $executablePath` which makes it easy for tools to save the data for future use.


Lightning00Blade (approved):

OrKoN:
marked as a fix to trigger a patch release.

PR #9760, opened 2023-03-01
feat: browsers: recognize chromium as a valid browser

thiagowfx:
Closes #9759

PR #9799, opened 2023-03-07
chore: allow release-please figure out the browsers version

OrKoN:
I don't have the context and not sure what pre-release does but it looks like the version is wrong and 0.0.4 has already been released.


Lightning00Blade:
#9760 as release as a feat, that made it bump to `0.1.0` which I felt was not the correct version so we manually changed as we needed to release `puppeteer`.

Lightning00Blade (approved):

PR #9805, opened 2023-03-08
chore: remove duplication in parameter definitions

OrKoN:
Technically parameters for each of the commands are different but for the user convenience most of the should work the same way so we can re-use their definitions.

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -43,7 +43,6 @@ type LaunchArgs = {
   detached: boolean;
 };
 
-
 export class CLI {
   #cachePath;
 
@@ -51,7 +50,7 @@ export class CLI {
     this.#cachePath = cachePath;
   }
 
-  #defineBrowserParameter(yargs: yargs.Argv<{}>) {
+  #defineBrowserParameter(yargs: yargs.Argv<unknown>): void {
     yargs.positional('browser', {
       description: 'The browser version',
       type: 'string',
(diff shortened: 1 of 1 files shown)


packages/browsers/src/CLI.ts
```diff
@@ -50,33 +50,86 @@ export class CLI {
     this.#cachePath = cachePath;
   }
 
+  #defineBrowserParameter(yargs: yargs.Argv<unknown>) {
```
Lightning00Blade: Nit: return type

packages/browsers/src/CLI.ts
```diff
@@ -50,33 +50,86 @@ export class CLI {
...
+              browser: args.browser.name,
+              buildId: args.browser.buildId,
+              cacheDir: args.path ?? this.#cachePath,
+              platform: args.platform,
+            })}`
+          );
+        }
+      )
+      .command(
+        'downloads <browser>',
```
Lightning00Blade: Is the idea for this to be an alias to `install`, the code looks the same? Yargs supports that -> https://github.com/yargs/yargs/blob/main/docs/advanced.md#command-aliases

OrKoN: I have forgotten to remove the command I have been trying out locally.

packages/browsers/src/CLI.ts
OrKoN: Done

Lightning00Blade (approved):

PR #9806, opened 2023-03-08
docs: fix some api-extractor warnings

Lightning00Blade:


packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -799,7 +799,7 @@ export class ElementHandle<
 
   /**
    * This method scrolls element into view if needed, and then uses
-   * {@link Page.screenshot} to take a screenshot of the element.
+   * {@link Page.(screenshot:3) } to take a screenshot of the element.
```
Lightning00Blade, line 802: When linking methods with overload we need to specify the main method. They are processed in the order they are written so the last one is we usually want to use.

jrandolf-google (approved):

PR #9818, opened 2023-03-09
chore: use `rimraf` for clean up

jrandolf-google:
Recursive glob is not supported in all versions of bash.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build:docs": "wireit",
     "check:pinned-deps": "tsx tools/ensure-pinned-deps",
     "check": "npm run check --workspaces --if-present && run-p check:*",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -11,7 +11,7 @@
     "build:docs": "wireit",
     "check:pinned-deps": "tsx tools/ensure-pinned-deps",
     "check": "npm run check --workspaces --if-present && run-p check:*",
-    "clean": "rm -rf **/.wireit && npm run clean --workspaces --if-present",
+    "clean": "rm -rf $(find . -type d -name .wireit) && npm run clean --workspaces --if-present",
```
OrKoN: what happens if no .wireit folder found?

jrandolf: Decided to use rimraf

OrKoN (approved):

PR #9820, opened 2023-03-09
chore: fix esbuild problem

OrKoN:
It looks like for whatever reasons --ignore-scripts started to affect esbuild installation leading to errors. I think we should gracefully handle installation on a checkout that has no lib folder and use Puppeteer-specific env variables to skip the browser download whenever needed.


Lightning00Blade (approved):
Have/Should we action on this and open a issue with EsBuild/NPM ?

PR #9834, opened 2023-03-13
chore: add an ability to create a Firefox profile

OrKoN:
Drive-by: added WEBDRIVER_BIDI_WEBSOCKET_ENDPOINT_REGEX, not used yet

### DIFF
--- a/packages/browsers/src/browsers/firefox.ts
+++ b/packages/browsers/src/browsers/firefox.ts
@@ -14,11 +14,12 @@
  * limitations under the License.
  */
 
+import fs from 'fs';
 import path from 'path';
 
 import {httpRequest} from '../httpUtil.js';
 
-import {BrowserPlatform} from './types.js';
+import {BrowserPlatform, ProfileOptions} from './types.js';
 
 function archive(platform: BrowserPlatform, buildId: string): string {
   switch (platform) {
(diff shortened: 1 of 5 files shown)


packages/browsers/src/browsers/firefox.ts
```diff
@@ -88,3 +89,256 @@ export async function resolveBuildId(
     });
   });
 }
+
+export async function createProfile(
+  extraPrefs: {[x: string]: unknown},
```
Lightning00Blade: Let's use the `ProfileOptions` (I think we will need to put it in `types.ts`) type here either passing a single argument (destructing it) or with specific  key from said type.

packages/browsers/test/src/firefox-data.spec.ts
```diff
@@ -69,4 +72,29 @@ describe('Firefox', () => {
...
+
+    it('should create a profile', async () => {
+      await createProfile(
+        {
+          test: 1,
+        },
+        tmpDir
+      );
+      const text = fs.readFileSync(path.join(tmpDir, 'user.js'), 'utf-8');
+      assert.ok(text.includes(`user_pref("test", 1);`));
```
Lightning00Blade: Nit: Let's assert a single default is set correctly, so we can be sure that we don't just discard them when you set a custom one.

Lightning00Blade (approved):
LGTM + minor comments

PR #9844, opened 2023-03-14
feat: implement system channels for chrome in browsers

OrKoN:
it is a bit difficult to test the integration with system browsers (it would require installing them on CI) so I opted into testing only the path computation for now and relying on the launch tests to verify that the correct executable paths can be launched.

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -19,10 +19,18 @@ import yargs from 'yargs';
 import {hideBin} from 'yargs/helpers';
 
 import {resolveBuildId} from './browsers/browsers.js';
-import {Browser, BrowserPlatform} from './browsers/types.js';
+import {
+  Browser,
+  BrowserPlatform,
+  ChromeReleaseChannel,
+} from './browsers/types.js';
 import {detectBrowserPlatform} from './detectPlatform.js';
 import {fetch} from './fetch.js';
-import {computeExecutablePath, launch} from './launcher.js';
+import {
+  computeExecutablePath,
+  computeSystemExecutablePath,
+  launch,
+} from './launcher.js';
 
 type InstallArgs = {
   browser: {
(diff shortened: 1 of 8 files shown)


packages/browsers/src/CLI.ts
```diff
@@ -135,15 +144,27 @@ export class CLI {
             desc: 'Whether to detach the child process.',
             default: false,
           });
+          yargs.option('system', {
+            type: 'boolean',
+            desc: 'Whether to search for a browser installed on the system instead of the cache folder.',
```
Lightning00Blade: Let's drop the `Whether to` here and it he command above.

packages/browsers/src/CLI.ts
```diff
@@ -135,15 +144,27 @@ export class CLI {
...
-          const executablePath = computeExecutablePath({
-            browser: args.browser.name,
-            buildId: args.browser.buildId,
-            cacheDir: args.path ?? this.#cachePath,
-            platform: args.platform,
-          });
+          const executablePath = args.system
+            ? computeSystemExecutablePath({
+                browser: args.browser.name,
+                // TODO: validation
```
Lightning00Blade: Do we need further validation? We already kind do validation when we go to the switch statement for `platform + channel`.

packages/browsers/test/src/chrome-data.spec.ts
```diff
@@ -69,4 +71,31 @@ describe('Chrome', () => {
       path.join('chrome-win', 'chrome.exe')
     );
   });
+
+  it('should resolve system executable path', () => {
+    process.env['PROGRAMFILES'] = 'C:\\ProgramFiles';
```
Lightning00Blade, line 76: Please clean this Global variable after use.

Lightning00Blade (approved):
LGTM + small comments

packages/browsers/src/CLI.ts
OrKoN: We will need to throw an error if a specific buildId is given or the `latest` which cannot be resolved on the system-wide browsers (updated the todo)

PR #9846, opened 2023-03-14
fix: add browsers entry point

OrKoN:
Adds main.ts that will be the entrypoint to the package API.


Lightning00Blade (approved):

Lightning00Blade:
A good first use case for this could be a change to `tools/check_availability.js` to use it.

PR #9854, opened 2023-03-14
docs: anchor selector on search example

snippet:
fix anchor selector on search exemple

### DIFF
--- a/examples/search.js
+++ b/examples/search.js
@@ -38,7 +38,7 @@ const puppeteer = require('puppeteer');
   await page.click(allResultsSelector);
 
   // Wait for the results page to load and display the results.
-  const resultsSelector = '.gsc-table-result a.gs-title';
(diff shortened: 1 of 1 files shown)


examples/search.js
```diff
@@ -38,7 +38,7 @@ const puppeteer = require('puppeteer');
   await page.click(allResultsSelector);
 
   // Wait for the results page to load and display the results.
-  const resultsSelector = '.gsc-results .gs-title';
+  const resultsSelector = '.gsc-table-result a.gs-title';
```
Lightning00Blade:
```suggestion
  const resultsSelector = '.gsc-table-result a.gs-title[href]';
```
I ran it locally to verify and it seem it pick up a template html, so this insures that we don't get that in our query,

Lightning00Blade (approved):
Thank you for the PR, will merge after all the checks pass.

PR #9859, opened 2023-03-15
chore: fix Actions Node version to LTS

Lightning00Blade:
Note: `latest` pulls directly from Node.JS site so the request is never cached, we have hit rate limits before, this should prevent this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,7 @@ jobs:
         uses: actions/setup-node@v3.5.1
         with:
           cache: npm
-          node-version: latest
+          node-version: lts/*
       - name: Install dependencies
         run: npm ci
         env:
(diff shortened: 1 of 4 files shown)


OrKoN (approved):

.github/workflows/ci.yml
```diff
@@ -279,11 +279,6 @@ jobs:
         with:
           node-version: ${{ matrix.node }}
       - name: Set up ${{ matrix.pkg_manager }}
-        id: workaround
```
OrKoN: we need this

.github/workflows/ci.yml
```diff
@@ -262,9 +262,9 @@ jobs:
           - macos-latest
           - windows-latest
         node:
-          - 14
```
OrKoN: let's keep 14, 16, 18, 20?

PR #9867, opened 2023-03-16
docs: add basic docs for browsers

OrKoN:


packages/browsers/src/CLI.ts
```diff
@@ -149,6 +176,22 @@ export class CLI {
             desc: 'Search for a browser installed on the system instead of the cache folder.',
             default: false,
           });
+          yargs.example(
+            '$0 launch chrome@1083080',
+            'Launch the Chrome browser identified by the revision 1083080.'
+          );
+          yargs.example(
+            '$0 launch chrome@112.0a1',
```
Lightning00Blade:
```suggestion
            '$0 launch firefox@112.0a1',
```

packages/browsers/src/CLI.ts
```diff
@@ -77,27 +78,53 @@ export class CLI {
...
     });
   }
 
   async run(argv: string[]): Promise<void> {
     await yargs(hideBin(argv))
+      .scriptName('@puppeteer/browsers')
       .command(
         'install <browser>',
-        'Download and install the specified browser',
+        'Download and install the specified browser. If successful, the command outputs the actual browser buildId that was installed and the absolute path to the browser executable.',
```
Lightning00Blade: What do you think about explicitly telling what the pattern of the line is `<browser>@<buildID>  <path>` ?

packages/browsers/src/CLI.ts
```diff
@@ -61,7 +61,8 @@ export class CLI {
 
   #defineBrowserParameter(yargs: yargs.Argv<unknown>): void {
     yargs.positional('browser', {
-      description: 'The browser version',
+      description:
+        'Which browser to install <browser>[@<buildId|latest>]. `latest` will try to find the latest available build. `buildId` is a browser-specific identifier such as a version or a revision.',
```
Lightning00Blade, line 65: This command is used for both `install` and `launch` but the second does not have a `latest`. Maybe we should input the description as an argument?

OrKoN: This is listed in known limitations. Eventually, we would probably support latest in both.

packages/browsers/src/CLI.ts
OrKoN: Done

Lightning00Blade (approved):

PR #9868, opened 2023-03-16
feat: implement a command to clear the cache

OrKoN:
The command requires an explicit path and a confirmation to avoid accidental deletion of the files.

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+import {stdin as input, stdout as output} from 'process';
+import * as readline from 'readline';
+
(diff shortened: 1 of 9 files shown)


packages/browsers/src/CLI.ts
```diff
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+import {stdin as input, stdout as output} from 'process';
+import * as readline from 'readline';
```
Lightning00Blade, line 18: Nit: Maybe use the promise API.

Lightning00Blade (approved):

packages/browsers/src/CLI.ts
OrKoN: Promise API was added in v17.0.0

PR #9870, opened 2023-03-16
chore: added * pattern for test expectations

Lightning00Blade:
Add the ability to use `*` in test description where `*` is any `it` title.

### DIFF
--- a/tools/mochaRunner/src/interface.ts
+++ b/tools/mochaRunner/src/interface.ts
@@ -17,7 +17,7 @@
 import Mocha from 'mocha';
 import commonInterface from 'mocha/lib/interfaces/common';
 
-import {getTestId} from './utils.js';
+import {testIdMatchesExpectationPattern} from './utils.js';
 
 type SuiteFunction = ((this: Mocha.Suite) => void) | undefined;
 type ExclusiveSuiteFunction = (this: Mocha.Suite) => void;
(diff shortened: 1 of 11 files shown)


tools/mochaRunner/src/utils.ts
```diff
@@ -216,3 +219,28 @@ export function getTestId(file: string, fullTitle?: string): string {
     ? `[${getFilename(file)}] ${fullTitle}`
     : `[${getFilename(file)}]`;
 }
+
+export function testIdMatchesExpectationPattern(
```
OrKoN, line 210: Let's add a test in mocha runner tests for this?

tools/mochaRunner/src/utils.ts
```diff
@@ -95,18 +95,17 @@ export function findEffectiveExpectationForTest(
...
-      return (
-        '' === expectation.testIdPattern ||
-        getTestId(result.file) === expectation.testIdPattern ||
-        getTestId(result.file, result.fullTitle) === expectation.testIdPattern
-      );
-    })
-    .pop();
+  return expectations.find(expectation => {
+    return (
+      '' === expectation.testIdPattern ||
```
OrKoN: do you want to require `*` and `[spec]*` instead of the current notation? It think it would be easier to maintain if we only support explicit *

tools/mochaRunner/src/main.ts
```diff
@@ -200,17 +201,13 @@ async function main() {
...
-        );
-        if (recommendation.length > 0) {
+        const updates = getExpectationUpdates(results, applicableExpectations, {
+          platforms: [os.platform()],
+          parameters,
+        });
+        if (updates.length > 0) {
           fail = true;
-          recommendations.push(...recommendation);
+          recommendations = updates;
```
OrKoN: I think the could be recommendations from previous test suites right?

tools/mochaRunner/src/interface.ts
```diff
@@ -28,8 +28,6 @@ const skippedTests: Array<{testIdPattern: string; skip: true}> = process.env[
   ? JSON.parse(process.env['PUPPETEER_SKIPPED_TEST_CONFIG'])
   : [];
 
-skippedTests.reverse();
```
OrKoN, line 31: Why was this removed? the last matching pattern should apply, right?

Lightning00Blade: The TestExpectations are now reversed after we read them, that lets us simplify the search -> See [change].(https://github.com/puppeteer/puppeteer/pull/9870/files#diff-b7bd54671e35ed689bbd420e35dbd2b343b9595252279fa082f9e419a724c98bR109)

OrKoN (approved):

PR #9875, opened 2023-03-17
test: add a few regressed selectors

OrKoN:


Lightning00Blade (approved):

jrandolf-google (approved):

jrandolf:
Actually, since our implementation now relies on Parsel, let's add these tests to Parsel instead. The better fix for this case is to do a try-catch fallback.

jrandolf-google (changes requested):
Actually, since our implementation now relies on Parsel, let's add these tests to Parsel instead. The better fix for this case is to do a try-catch fallback.

jrandolf-google (approved):

PR #9876, opened 2023-03-17
fix: fallback to CSS

### LINKED ISSUE #9828: [Bug]: Error with selector including `:not()`
### Bug expectation

I expect the selector `'[data-test-id^="test-"]:not([data-test-id^="test-foo"])'` to select all elements with a `'data-test-id'` attribute starting with "test-", except for those with the suffix "-foo".

Since updating to Puppeteer-core 19.7.3, the selector `'[data-test-id^="test-"]:not([data-test-id^="test-foo"])'` is no longer recognized as a valid selector, and I receive the following error message:
```
Error: Evaluation failed: Error: [data-test-id^="test-"]:not([data-test-id^="test-foo"]) is not a valid selector: Unexpected sequence
```

Downgrading to version 19.7.2 of Puppeteer allows the selector `'[data-test-id^="test-"]:not([data-test-id^="test-foo"])'` to function as expected

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript

(async () => {
  const browser = await launch();

  await page.$$(
    '[data-test-id^="test-"]:not([data-test-id^="test-foo"])',
  );

  await browser.close();
})();
```

### Error string

`Error: Evaluation failed: Error: [data-test-id^="test-"]:not([data-test-id^="test-foo"]) is not a valid selector: Unexpected sequence )`

### Puppeteer configuration

_No response_

### Puppeteer version

19.7.3

### Node version

18.13.0

### Package manager

yarn

### Package manager version

1.22.19

### Operating system

macOS

jrandolf-google:
Fixed: #9828

### DIFF
--- a/packages/puppeteer-core/src/injected/PQuerySelector.ts
+++ b/packages/puppeteer-core/src/injected/PQuerySelector.ts
@@ -16,7 +16,6 @@
 
 import type {AwaitableIterable} from '../common/types.js';
 import {AsyncIterableUtil} from '../util/AsyncIterableUtil.js';
-import {isErrorLike} from '../util/ErrorLike.js';
 
 import {ariaQuerySelectorAll} from './ARIAQuerySelector.js';
 import {customQuerySelectors} from './CustomQuerySelector.js';
@@ -251,10 +250,7 @@ export const pQuerySelectorAll = function (
   try {
     [selectors, isPureCSS] = parsePSelectors(selector);
   } catch (error) {
-    if (!isErrorLike(error)) {
-      throw new SelectorError(selector, String(error));
-    }
-    throw new SelectorError(selector, error.message);
+    return (root as unknown as QueryableNode).querySelectorAll(selector);
   }
 
   if (isPureCSS) {


packages/puppeteer-core/src/injected/PSelectorParser.ts
```diff
@@ -51,69 +51,74 @@ const unquote = (text: string): string => {
...
-  }
-  if (storage.length) {
-    compoundSelector.push(stringify(storage));
+    if (storage.length) {
+      compoundSelector.push(stringify(storage));
+    }
+    return [selectors, isPureCSS];
+  } catch {
+    // In case parsing breaks, we fallback to CSS.
+    return [[], true];
```
OrKoN: Let's do the try/catch on the consumer side. I don't think p-selectors should be concerned with fallbacks and know about CSS.

jrandolf: SGTM

OrKoN (approved):

PR #9877, opened 2023-03-17
fix: implement protocol-level timeouts

### LINKED ISSUE #9801: [Bug]: page.screenshot (silently) times out when trying to take a screenshot of an alert
### Bug expectation

I expected that screenshot would never time out (it doesn't even have an option to set the timeout time, as most other Puppeteer functions do).

But, if it were to timeout, I'd expect it to do so in a reasonable amount of time, with an error message explaining why it timed out.

And *really* I'd expect screenshot to just take a screenshot of the alert ... or take a "screenshot" of a black screen and the text "alerts can't be screenshoted" ... but if that's not possible, then failing in a fixed amount of time with an error message would be nice.

And if nothing else, please at least add a line to https://pptr.dev/api/puppeteer.page.screenshot documenting this.

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript

(async () => {

    await page.goto(SOME_PAGE);
    // This button should trigger a window.alert
    // Also note that "click" *does* have a timeout option
    // (unlike screenshot)
    await page.click('button', { timeout: 2000 });
    await page.screenshot({ path: 'screenshot.png' })
    // wait for your test runner to time out

})();
```

### Error string

None: no error is ever thrown (that's a big part of the bug)

### Puppeteer configuration

_No response_

### Puppeteer version

19.7.2

### Node version

v19.7.0

### Package manager

npm

### Package manager version

9.5.0

### Operating system

Linux

OrKoN:
This PR implements timeouts for the protocol-level calls and a new launch/connect option called `protocolTimeout`. The default value will be 30s and there is an option to disable it by providing an undefined/negative/0 value.

Drive-by: refactored callbacks to encapsulate timeouts and reduce duplication. Also, marked ConnectionCallback as internal as there was no way to use it externally anyway.

Closes #9801

For follow-ups: 1) we should probably mark connection as an internal class (but we need to deprecate it first and see if there are users for it) and changes the constructor parameters to accept an object instead of positional arguments. 2) we could probably do smth better for the original message/protocol error things.


docs/api/puppeteer.connection.md
```diff
@@ -14,9 +14,15 @@ export declare class Connection extends EventEmitter
 
 ## Constructors
 
-| Constructor                                                                     | Modifiers | Description                                                    |
```
OrKoN, line 17: TODO: should not make a breaking change

jrandolf-google (approved):

packages/puppeteer-core/src/common/BrowserConnector.ts
```diff
@@ -59,6 +59,12 @@ export interface BrowserConnectOptions {
    * @internal
    */
   protocol?: 'cdp' | 'webDriverBiDi';
+  /**
+   * Timeout setting for individual protocol (CDP) calls.
+   *
+   * @defaultValue 30000
+   */
+  protocolTimeout?: number;
```
jrandolf, line 67: Can we just name this `timeout`?

jrandolf-google (changes requested):

packages/puppeteer-core/src/common/BrowserConnector.ts
OrKoN: No, because we already have a `timeout` in the launch options that get united with the this interface and the `timeout` is the timeout for the browser launch operation.

jrandolf-google (approved):

PR #9883, opened 2023-03-20
fix: waitForNavigation issue with aborted events

### LINKED ISSUE #8383: [Bug]: page.waitForNetworkIdle timeout inconsistently
### Bug description
For some reason `page.waitForNetworkIdle` is timing out for no reason. the network is empty and even so it timeout after 30 seconds.

This error happens only sometimes and is very weird. Despite of that, I find a way to make it fail consistently by repeating some steps multiple times

### Steps to reproduce the problem:

I created a repo with some simple [puppeteer code ](https://github.com/Josehower/puppetter-timeout-reproduction/blob/main/index.ts)

### Locally 
you can reproduce it locally by (my current local environment is WSL2 ubuntu)
1. clone the repo
2. install dependencies `yarn`
3. `yarn start`

### In GitHub Actions
 1. Create a new issue in the [reproduction repo ](https://github.com/Josehower/puppetter-timeout-reproduction/blob/main/index.ts) or use the issue `test`
 2. Create any comment
 3. Check the actions tab

### Puppeteer version

  "puppeteer": "14.1.1",

### Node.js version

Local v16.15.0  / [node 16](https://github.com/Josehower/puppetter-timeout-reproduction/blob/8be56524cbfbfb1d71561b25f4b6c0c283a41b94/.github/workflows/upleveled-drone.yml#L18)

### npm version

local 8.9.0 / actions https://github.com/actions/checkout

Linux

### Relevant log output

https://github.com/Josehower/puppetter-timeout-reproduction/runs/6524437119?check_suite_focus=true#step:4:20

```shell
file:///home/runner/work/puppetter-timeout-reproduction/puppetter-timeout-reproduction/upleveled-drone.mjs:18421
      rejectCallback(new TimeoutError("Timeout exceeded while waiting for event"));
                     ^

TimeoutError: Timeout exceeded while waiting for event
    at Timeout.<anonymous> (file:///home/runner/work/puppetter-timeout-reproduction/puppetter-timeout-reproduction/upleveled-drone.mjs:18421:22)
    at listOnTimeout (node:internal/timers:559:17)
    at processTimers (node:internal/timers:502:7)
```

Thanks for any attention that you gave to this issue, puppeteer is an amazing tool that we use every day in my company. I am also open to creating a PR if we are able to identify what is the problem.

Good energy 👍👍👍👍

### LINKED ISSUE #9881: [Bug]: Timeout error while waiting for network idle
### Bug expectation

The app is tested by Puppeteer updating an image source with a nonvalid URL, which results in a 404 response from the server. The image update is happening on each letter type, which causes a bunch of 404 to happen quickly. Sometimes, when running the test script using Puppeteer, the `page.waitForNetworkIdle()` method times out after 30 seconds, which is unexpected behavior. Additionally, the block that fails is always different. 

From my understanding `page.waitForNetworkIdle()` waits until there are no more network connections for at least 500ms and then resolves. It's not waiting for any specific navigation event but rather for the network to idle before proceeding. So if there are no network connections for 500ms, the test should move forward.

## Expected Behavior
The script should run without errors and complete successfully. The `page.waitForNetworkIdle()` method should not time out after 30 seconds since the broken link responds with a 404 very quickly.

## Actual Behavior
Sometimes the `page.waitForNetworkIdle()` method times out after 30 seconds, and the block that fails is always different.

## Additional Information

The test fails less frequently when running the code inside a loop and executing multiple times in a row. Maybe there is less idle time between each execution, which could cause the network to become idle faster. This could explain why the code inside the loop fails less frequently than the code without the loop.

```js
for (let i = 1; i < 20; i++) {
  await element.type('text');
  await page.waitForNetworkIdle();
  console.log(`waitForNetworkIdle test ${i}`);
}
```

### Bug behavior

- [X] Flaky
- [ ] PDF

```TypeScript

export async function test(urlToTest) {
  await page.goto(urlToTest);

  const [element] = await page.$x(`//label[contains(text(), 'Text')]`);

  await element.type('test');

  await page.waitForNetworkIdle();
  await console.log('waitForNetworkIdle test..');

  await element.type('text');
  await page.waitForNetworkIdle();
  console.log('waitForNetworkIdle test 1');

  await element.type('text');
  await page.waitForNetworkIdle();
  console.log('waitForNetworkIdle test 2');

  await element.type('text');
  await page.waitForNetworkIdle();
  console.log('waitForNetworkIdle test 3');

  await element.type('text');
  await page.waitForNetworkIdle();
  console.log('waitForNetworkIdle test 4');

  await element.type('text');
  await page.waitForNetworkIdle();
  console.log('waitForNetworkIdle test 5');

  await element.type('text');
  await page.waitForNetworkIdle();
  console.log('waitForNetworkIdle test 6');

  await element.type('text');
  await page.waitForNetworkIdle();
  console.log('waitForNetworkIdle test 7');

  await element.type('text');
  await page.waitForNetworkIdle();
  console.log('waitForNetworkIdle test 8');

  await element.type('text');
  await page.waitForNetworkIdle();
  console.log('waitForNetworkIdle test 9');

  await element.type('text');
  await page.waitForNetworkIdle();
  console.log('waitForNetworkIdle test 10');

  await page.close();
  await browser.close();
}

await test('https://gleeful-cat-1fcebd.netlify.app/');
```

### Error string

Timeout exceeded while waiting for event

### Puppeteer configuration

_No response_

### Puppeteer version

19.7.5

### Node version

18.13.0

### Package manager

yarn

### Package manager version

1.22.19

### Operating system

macOS

Lightning00Blade:
Fixes #9881, fixes #8383.
The issue seem to be that we don't listen to all the possible event for Request/Response so we never try to see if there are no requests.

@OrKoN Let's take a closer look together to see if there are other "unhappy" paths that need to handled.

### DIFF
--- a/packages/puppeteer-core/src/common/NetworkManager.ts
+++ b/packages/puppeteer-core/src/common/NetworkManager.ts
@@ -302,11 +302,7 @@ export class NetworkManager extends EventEmitter {
   }
 
   #onAuthRequired(event: Protocol.Fetch.AuthRequiredEvent): void {
-    /* TODO(jacktfranklin): This is defined in protocol.d.ts but not
-     * in an easily referrable way - we should look at exposing it.
-     */
-    type AuthResponse = 'Default' | 'CancelAuth' | 'ProvideCredentials';
-    let response: AuthResponse = 'Default';
+    let response: Protocol.Fetch.AuthChallengeResponse['response'] = 'Default';
     if (this.#attemptedAuthentications.has(event.requestId)) {
       response = 'CancelAuth';
     } else if (this.#credentials) {
--- a/packages/puppeteer-core/src/common/Page.ts
+++ b/packages/puppeteer-core/src/common/Page.ts
@@ -991,10 +991,7 @@ export class CDPPage extends Page {
 
     const networkManager = this.#frameManager.networkManager;
 
-    let idleResolveCallback: () => void;
-    const idlePromise = new Promise<void>(resolve => {
-      idleResolveCallback = resolve;
-    });
+    const idlePromise = createDeferredPromise<void>();
 
     let abortRejectCallback: (error: Error) => void;
     const abortPromise = new Promise<Error>((_, reject) => {
@@ -1002,10 +999,6 @@ export class CDPPage extends Page {
     });
 
     let idleTimer: NodeJS.Timeout;
-    const onIdle = () => {
-      return idleResolveCallback();
-    };
-
     const cleanup = () => {
       idleTimer && clearTimeout(idleTimer);
       abortRejectCallback(new Error('abort'));
@@ -1014,7 +1007,7 @@ export class CDPPage extends Page {
     const evaluate = () => {
       idleTimer && clearTimeout(idleTimer);
       if (networkManager.numRequestsInProgress() === 0) {
-        idleTimer = setTimeout(onIdle, idleTime);
+        idleTimer = setTimeout(idlePromise.resolve, idleTime);
       }
     };
 
@@ -1038,6 +1031,7 @@ export class CDPPage extends Page {
     const eventPromises = [
       listenToEvent(NetworkManagerEmittedEvents.Request),
       listenToEvent(NetworkManagerEmittedEvents.Response),
+      listenToEvent(NetworkManagerEmittedEvents.RequestFailed),
     ];
 
     await Promise.race([
--- a/test/assets/abort-request.html
+++ b/test/assets/abort-request.html
@@ -1,0 +1,13 @@
+<button id="abort"></button>
+
+<script>
+  const button = document.getElementById('abort');
+  button.addEventListener('click', getJson)
+  async function getJson() {
+    const abort = new AbortController();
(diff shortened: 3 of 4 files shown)


OrKoN (approved):
Nice finding! Unfortunately, I don't know if there are other unhappy cases.

karlhorky:
Nice, thanks for this @Lightning00Blade and @OrKoN ! 🙌  From [this PR from the bot](https://github.com/puppeteer/puppeteer/pull/9863) it looks like it will be released in `19.8.0`?

cage1618:
This commit seems to result in continuously adding new request/response event listeners, ultimately leading to memory leaks. Please confirm this, thank you. @Lightning00Blade @OrKoN

cage1618:
> This commit seems to result in continuously adding new request/response event listeners, ultimately leading to memory leaks. Please confirm this, thank you. @Lightning00Blade @OrKoN 

Here is the snapshot：https://github.com/puppeteer/puppeteer/issues/5893#issuecomment-2047476695

PR #9888, opened 2023-03-21
chore: replace rimraf in tests with a helper with retries

OrKoN:


test/src/mocha-utils.ts
```diff
@@ -332,3 +331,11 @@ export const createTimeout = <T>(
     }, n);
   });
 };
+
+export function rimrafSync(target: string): void {
```
Lightning00Blade: Let's use a different name, as `rimraf` is only associated with package. `recursiveDeleteSync` maybe?

OrKoN: `rmSync`?

Lightning00Blade (approved):

PR #9891, opened 2023-03-21
chore: don't skip only test

Lightning00Blade:
This allows us to bypass the `TestExpectations.json` while developing and using `.only` in test (Tested with both `describe.only` and `it.only`).
We should not worry about it in CI as we will fail the EsLint rule before trying to run the tests. 

Note: Recommended to use `--no-suggestions` with this as they are just wrong.


OrKoN (approved):

Lightning00Blade:
This change just never skips, as the predicate is always false. There seems to be no easy way see if the test was marked as `.only`. Closing for now will revisit later.

PR #9893, opened 2023-03-21
chore: explore different SetContent for BiDi

Lightning00Blade:


packages/puppeteer-core/src/common/bidi/Context.ts
```diff
@@ -213,9 +213,9 @@ export class Context extends EventEmitter {
       // We rely upon the fact that document.open() will reset frame lifecycle with "init"
       // lifecycle event. @see https://crrev.com/608658
       this.evaluate(html => {
```
OrKoN: I think this code should be shared between CDP and BiDi implementations.

Lightning00Blade: I will do that in a separate PR, and rebase this one after it. I want to check what different event are triggered for CDP and if we are OK with them, then we can change the implementation.

Lightning00Blade:
This triggers a Navigation for CDP (did not check BiDi), that sometimes destroys the `ExecutionContext` and test start failing as they can't execute code against it.

PR #9894, opened 2023-03-21
chore: reuse SetContent across implementations

Lightning00Blade:


packages/puppeteer-core/src/common/util.ts
```diff
@@ -441,3 +442,19 @@ export async function getReadableFromProtocolStream(
...
+/**
+ * @internal
+ */
+export async function setPageContent(
+  page: Pick<Page, 'evaluate'>,
+  content: string
+): Promise<void> {
+  // We rely upon the fact that document.open() will reset frame lifecycle with "init"
+  // lifecycle event. @see https://crrev.com/608658
+  page.evaluate(html => {
```
OrKoN: needs a return

packages/puppeteer-core/src/common/util.ts
```diff
@@ -441,3 +442,19 @@ export async function getReadableFromProtocolStream(
     },
   });
 }
+
+/**
+ * @internal
+ */
+export async function setPageContent(
```
OrKoN, line 449: Let's define it on the base/api class? and re-use in the subclasses?

OrKoN (approved):

PR #9897, opened 2023-03-22
chore: remove retries and fix expectations

OrKoN:
This CL removes retries performed by mocha, introduced an automatic formatter for TestExpectations based on the expectation specificity and fixes an issue with wrongfully generated expectations.


Lightning00Blade (approved):
Note: PR did not trigger CI, as we don't think of this as a change of puppeteer.

Lightning00Blade (approved):

PR #9898, opened 2023-03-22
refactor: use @puppeteer/browsers for install

OrKoN:
The CL rewrite install.ts to use the @puppeteer/browsers package. It should keep all existing functionality and have no observable changes.

### DIFF
--- a/packages/puppeteer/src/node/install.ts
+++ b/packages/puppeteer/src/node/install.ts
@@ -14,13 +14,15 @@
  * limitations under the License.
  */
 
-import https, {RequestOptions} from 'https';
-import URL from 'url';
-
-import createHttpsProxyAgent, {HttpsProxyAgentOptions} from 'https-proxy-agent';
-import ProgressBar from 'progress';
-import {getProxyForUrl} from 'proxy-from-env';
-import {PuppeteerNode} from 'puppeteer-core/internal/node/PuppeteerNode.js';
+import {
+  fetch,
+  Browser,
+  resolveBuildId,
+  makeProgressCallback,
+  detectBrowserPlatform,
+  BrowserPlatform,
+} from '@puppeteer/browsers';
+import {Product} from 'puppeteer-core';
 import {PUPPETEER_REVISIONS} from 'puppeteer-core/internal/revisions.js';
 
 import {getConfiguration} from '../getConfiguration.js';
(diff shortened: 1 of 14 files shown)


packages/puppeteer/src/node/install.ts
```diff
@@ -37,167 +37,56 @@ const supportedProducts = {
...
-    }
-
-    return browserFetcher
-      .download(revisionInfo.revision, onProgress)
-      .then(() => {
-        return browserFetcher.localRevisions();
-      })
-      .then(onSuccess)
-      .catch(onError);
+  const platform = detectBrowserPlatform();
```
OrKoN: TODO: special mode for mac ARM

packages/puppeteer/src/node/install.ts
```diff
@@ -37,168 +39,67 @@ const supportedProducts = {
...
-    }
-
-    return browserFetcher
-      .download(revisionInfo.revision, onProgress)
-      .then(() => {
-        return browserFetcher.localRevisions();
-      })
-      .then(onSuccess)
-      .catch(onError);
+  // TODO: remove once Mac ARM is enabled by default for Puppeteer.
```
Lightning00Blade:
```suggestion
  // TODO: remove once Mac ARM is enabled by default for Puppeteer https://github.com/puppeteer/puppeteer/issues/9630.
```

test/installation/.mocharc.cjs
```diff
@@ -19,5 +19,5 @@ module.exports = {
   spec: ['build/**/*.spec.js'],
   timeout: '240000ms',
   // Parallel processing fails on other package managers due to caching.
-  parallel: (process.env['PKG_MANAGER'] ?? 'npm') === 'npm',
+  // parallel: (process.env['PKG_MANAGER'] ?? 'npm') === 'npm',
```
Lightning00Blade: Is this still needed? Either uncomment or remove.

Lightning00Blade (approved):

OrKoN:
I will work on this more: currently not ready to be landed.

Lightning00Blade (approved):

sabberworm:
AFAICT, this change causes `PUPPETEER_DOWNLOAD_HOST`/`configuration.downloadHost` to no longer be respected.
Flag is still documented, however…
It looks like this can be fixed by passing it as a third argument to `downloadUrls[browser]` in https://github.com/puppeteer/puppeteer/blob/327febe2c9c44bdd48a3e674f2a3894c40ca131e/packages/browsers/src/fetch.ts#L143
However, this would still be a breaking change in case of chromium which previously appended the `/chromium-browser-snapshots` prefix to the configured download URL and now expects it to contain it. However, I welcome this change for clarity, consistency and versatility.

sabberworm:
Yes, I use a private mirror.

PR #9904, opened 2023-03-22
chore: use events rather than commands

Lightning00Blade:


packages/puppeteer-core/src/common/bidi/Browser.ts
```diff
@@ -30,24 +30,21 @@ import {Connection} from './Connection.js';
...
-  /**
-   * @internal
-   */
   static async create(opts: Options): Promise<Browser> {
     // TODO: await until the connection is established.
     try {
       await opts.connection.send('session.new', {});
     } catch {}
+    await opts.connection.send('session.subscribe', {
+      events: ['browsingContext.contextCreated'] as any,
```
OrKoN: types missing?

packages/puppeteer-core/src/common/bidi/Connection.ts
```diff
@@ -140,30 +140,37 @@ export class Connection extends EventEmitter {
           object.message
         );
       } else {
-        if (
-          this.#callbacks.getCallback(object.id)?.label ===
```
OrKoN, line 144: let's remove getCallback altogether (I think it was needed for this place only)

OrKoN (approved):

OrKoN:
Could you please update expectations with the flaky tests here?

PR #9910, opened 2023-03-24
chore: increase rm retries for browsers

OrKoN:
For Windows.


Lightning00Blade (approved):

Lightning00Blade (commented):
Let's actually increase the delay as well to 2x so `retryDelay: 200`

PR #9912, opened 2023-03-24
chore: store artifacts with test results

OrKoN:
This PR allows storing test results in GitHub Actions for future analysis.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,13 +148,18 @@ jobs:
         run: npm run test-types
       - name: Run all tests (for non-Linux)
         if: ${{ matrix.os != 'ubuntu-latest' }}
-        run: npm run test -- --test-suite ${{ matrix.suite }}
+        run: npm run test -- --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
       - name: Install linux dependencies.
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: sudo apt-get install xvfb
       - name: Run all tests (for Linux)
         if: ${{ matrix.os == 'ubuntu-latest' }}
-        run: xvfb-run --auto-servernum npm run test -- --test-suite ${{ matrix.suite }}
+        run: xvfb-run --auto-servernum npm run test -- --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
+      - uses: actions/upload-artifact@v3
+        if: always()
+        with:
+          name: test-results
+          path: /tmp/artifacts/*.json
 
   chrome-tests-required:
     name: '[Required] Chrome tests'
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -208,13 +212,17 @@ jobs:
...
         if: ${{ matrix.os != 'ubuntu-latest' }}
-        run: npm run test -- --test-suite ${{ matrix.suite }}
+        run: npm run test -- --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_${{ github.job }}_${{ github.run_id }}.json
       - name: Install linux dependencies.
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: sudo apt-get install xvfb
       - name: Run all tests (for Linux)
         if: ${{ matrix.os == 'ubuntu-latest' }}
-        run: xvfb-run --auto-servernum npm run test -- --test-suite ${{ matrix.suite }}
+        run: xvfb-run --auto-servernum npm run test -- --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_${{ github.job }}_${{ github.run_id }}.json
```
Lightning00Blade: How to artifacts work? Do we need an unique name for each file? If so we this would be overridden on every matrix run.

OrKoN: we are going to see how it works

OrKoN: I expect that artifacts are scoped per run, at least they are shown so in the UI

.github/workflows/ci.yml
```diff
@@ -208,13 +212,17 @@ jobs:
...
+      - uses: actions/upload-artifact@v3
```
OrKoN, line 223: TODO: `if: always()`

.github/workflows/ci.yml
OrKoN: yeah needs some sort of id generation

.github/workflows/ci.yml
```diff
@@ -148,13 +148,18 @@ jobs:
...
         if: ${{ matrix.os != 'ubuntu-latest' }}
-        run: npm run test -- --test-suite ${{ matrix.suite }}
+        run: npm run test -- --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_${{ github.run_id }}_INSERTID.json
       - name: Install linux dependencies.
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: sudo apt-get install xvfb
       - name: Run all tests (for Linux)
         if: ${{ matrix.os == 'ubuntu-latest' }}
-        run: xvfb-run --auto-servernum npm run test -- --test-suite ${{ matrix.suite }}
+        run: xvfb-run --auto-servernum npm run test -- --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_${{ github.run_id }}_INSERTID.json
```
Lightning00Blade: Can you try with `${{ github.run_id }}_${{ github.job }}_${{ github.run_attempt }}` ? That way will not have to do some funky runtime updates.

OrKoN: Honestly, I would prefer to rely on custom script here. The proposal is still not unique I believe because run_attempt is the same for every matrix job and so is run_id and job.

OrKoN: I think we also don't need github.run_id The cli arguments we can improve after we find this useful.

Lightning00Blade (approved):

PR #9913, opened 2023-03-24
chore: remove long jobs

OrKoN:
Let's try to keep any job within 15 min. Also, let's keep installation tests only for the latest node LTS version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -123,6 +123,8 @@ jobs:
         exclude:
           - os: windows-latest
             suite: chrome-bidi
+          - os: macos-latest
+            suite: chrome-headful
     steps:
       - name: Check out repository
         uses: actions/checkout@v3.3.0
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -266,8 +273,6 @@ jobs:
           - macos-latest
           - windows-latest
         node:
-          - 14
-          - 16
           - 18
```
Lightning00Blade, line 286: Some leftover code for a fix in window issue further down.

OrKoN: what do you mean?

Lightning00Blade: https://github.com/puppeteer/puppeteer/blob/main/.github/workflows/ci.yml#L297

OrKoN: That’s okay. Maybe we still bring this change back if we get regressions. That workaround can be removed once we remove node 14.

Lightning00Blade (approved):

PR #9914, opened 2023-03-24
chore: add PDF print for BiDi

Lightning00Blade:
TODO: Better naming of some of the variables, and maybe a prettier TypeScript Type for Parsed PDFOptions.


OrKoN (approved):

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -199,6 +202,64 @@ export class Page extends PageBase {
...
+        background: params.printBackground,
+        margin: params.margin,
+        orientation: params.landscape ? 'landscape' : 'portrait',
+        page: {
+          width: params.width,
+          height: params.height,
+        },
+        pageRanges: params.pageRanges.split(', '),
+        scale: params.scale,
+        shrinkToFit: !params.preferCSSPageSize,
```
OrKoN, line 221: @Lightning00Blade do you recall why we decided to default shrinkToFit to preferCSSPageSize?

Lightning00Blade: Sadly I don't recall

PR #9917, opened 2023-03-27
feat: update Chrome browser binaries

OrKoN:
Drive-by: separated tests to be by browser. describe/it names should be improved in a follow-up.

### DIFF
--- a/packages/browsers/src/httpUtil.ts
+++ b/packages/browsers/src/httpUtil.ts
@@ -48,7 +48,7 @@ export function httpRequest(
     protocol: url.protocol,
     hostname: url.hostname,
     port: url.port,
-    path: url.pathname,
+    path: url.pathname + url.search,
     method,
     headers: keepAlive ? {Connection: 'keep-alive'} : undefined,
   };
--- a/packages/browsers/test/src/chrome/cli.spec.ts
+++ b/packages/browsers/test/src/chrome/cli.spec.ts
@@ -18,41 +18,22 @@ import assert from 'assert';
 import fs from 'fs';
 import os from 'os';
 import path from 'path';
-import * as readline from 'readline';
-import {Writable, Readable} from 'stream';
 
-import {CLI} from '../../lib/cjs/CLI.js';
-import {Cache} from '../../lib/cjs/main.js';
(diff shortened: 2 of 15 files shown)


packages/browsers/src/httpUtil.ts
```diff
@@ -48,7 +48,7 @@ export function httpRequest(
     protocol: url.protocol,
     hostname: url.hostname,
     port: url.port,
-    path: url.pathname,
+    path: url.pathname + url.search,
```
Lightning00Blade, line 51: Nit: I think this should be `url.href` as that is the full url
```suggestion
    path: url.href,
```

packages/browsers/test/src/chrome/cli.spec.ts
```diff
@@ -92,40 +74,14 @@ describe('CLI', function () {
...
-      '@puppeteer/browsers',
-      'install',
-      `firefox@latest`,
-      `--path=${tmpDir}`,
-      '--platform=linux',
-    ]);
-  });
-
-  it('should download latest Chrome binaries', async () => {
+  it.skip('should download latest Chrome binaries', async () => {
```
Lightning00Blade, line 85: Nit: Add comment why skipped.

Lightning00Blade (approved):

swissspidy:
I just updated to Puppeteer 19.8.1 and I think this change here caused a regression.

With `PUPPETEER_PRODUCT=chrome npm install puppeteer`, it downloads Chromium 1108766 to `~/.cache/puppeteer/chromium`, but then when running Puppeteer, the browser fetcher is looking for `~/.cache/puppeteer/chrome/mac-1108766` which of course doesn't exist. Then I get `Could not find Chromium (rev. 1108766)` errors.

Is that a known issue and perhaps #9937 solves this?

When I downgrade to Puppeteer 19.8.0, Chromium is indeed downloaded to `~/.cache/puppeteer/chrome` and all works.

OrKoN:
@swissspidy thanks for the report! It looks like this is an unintended change. I am looking into it.

OrKoN:
One of the important changes for compatibility was marked as a chore so the browsers package version didn't get bumped: https://github.com/puppeteer/puppeteer/pull/9939 should address it.

PR #9923, opened 2023-03-28
chore: add basic screenshot to BiDi

Lightning00Blade:


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -2083,6 +2083,29 @@ export class Page extends EventEmitter {
...
 
+  /**
+   * @internal
+   */
+  async _maybeWriteBufferToFile(
+    path: string | undefined,
+    buffer: Buffer
+  ): Promise<void> {
+    try {
+      if (path) {
```
OrKoN: let's move the if to be before try/catch?

packages/puppeteer-core/src/common/Page.ts
```diff
@@ -1440,24 +1439,13 @@ export class CDPPage extends Page {
...
-          );
-        }
-        throw error;
-      }
+    if (options.encoding === 'base64') {
+      return result.data;
     }
+
+    const buffer = Buffer.from(result.data, 'base64');
+    this._maybeWriteBufferToFile(options.path, buffer);
```
OrKoN: await?

OrKoN (approved):

PR #9937, opened 2023-03-29
refactor: use browsers for launchers

OrKoN:
This PR refactors the browser launch code to use the @puppeteer/browsers package. The launching logic is unified in the ProductLauncher and specific hooks are extracted for the subclasses.


packages/browsers/src/launcher.ts
```diff
@@ -143,6 +144,11 @@ class Process {
   #args: string[];
   #browserProcess: childProcess.ChildProcess;
   #exited = false;
+  // The browser process can be closed externally or from the driver process. We
+  // need to invoke the hooks only once though but we don't know how many times
+  // we will be invoked.
+  #hooksRun = false;
```
Lightning00Blade:
```suggestion
  #hooksRan = false;
```

packages/browsers/src/launcher.ts
```diff
@@ -236,19 +263,24 @@ class Process {
...
+  async close(): Promise<void> {
+    await this.#runHooks();
     if (this.#exited) {
       return Promise.resolve();
     }
     this.kill();
     return this.#browserProcessExiting;
   }
 
+  closed(): Promise<void> {
```
Lightning00Blade: The `close` and `closed` at too similar (Especially give they have the same signature), either better naming (suggestions: closed -> exited or close -> shutdown) or comment would be appreciated.

packages/puppeteer-core/src/node/FirefoxLauncher.ts
```diff
@@ -91,14 +104,17 @@ export class FirefoxLauncher extends ProductLauncher {
       // When using a custom Firefox profile it needs to be populated
       // with required preferences.
       isTempUserDataDir = false;
-      const prefs = this.defaultPreferences(extraPrefsFirefox);
-      this.writePreferences(prefs, userDataDir);
     } else {
-      userDataDir = await this._createProfile(extraPrefsFirefox);
+      userDataDir = await fs.promises.mkdtemp(this.getProfilePath());
```
Lightning00Blade: Import `mkdtemp` from `fs/promises`. If we already have the import lets use it.

packages/puppeteer-core/src/node/FirefoxLauncher.ts
```diff
@@ -110,86 +126,75 @@ export class FirefoxLauncher extends ProductLauncher {
...
+    if (connection) {
+      // Attempt to close the browser gracefully
       try {
-        await browser.waitForTarget(
-          t => {
-            return t.type() === 'page';
-          },
-          {timeout}
-        );
+        await connection.send('Browser.close');
```
Lightning00Blade: This only works for CDP (My presumption is that CDP is case specific), the command will be different `browser.close` so they will be incomparable. Let's put a `TODO: BiDi https://github.com/w3c/webdriver-bidi/pull/286` comment
Also both methods are the same for ChromeLauncher and FirefoxLauncher, lets move it to ProductLauncher as well.

Lightning00Blade (approved):

PR #9945, opened 2023-03-30
chore: simplify installation tests

OrKoN:
This PR changes the installation tests to use mocha context to pass state which allows easily identifying tests and reasoning about the execution steps. It sacrifies the parallel mode because the context cannot be parallelised but that is beneficial because the parallel mode does not support .only, works only for npm and makes it hard to debug tests. I see no difference between the parallel mode and sequential mode in terms of performance.

### DIFF
--- a/test/installation/src/puppeteer-firefox.spec.ts
+++ b/test/installation/src/puppeteer-firefox.spec.ts
@@ -18,12 +18,11 @@ import assert from 'assert';
 import {readdir} from 'fs/promises';
 import {join} from 'path';
 
-import {describeInstallation} from './describeInstallation.js';
+import {configureSandbox} from './sandbox.js';
 import {readAsset} from './util.js';
 
-describeInstallation(
-  '`puppeteer` with Firefox',
-  {
+describe('`puppeteer` with Firefox', () => {
+  configureSandbox({
     dependencies: ['@puppeteer/browsers', 'puppeteer-core', 'puppeteer'],
     env: cwd => {
       return {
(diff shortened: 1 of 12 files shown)


test/installation/src/puppeteer-webpack.spec.ts
```diff
@@ -17,42 +17,40 @@
...
 import {join} from 'path';
 
-import {describeInstallation} from './describeInstallation.js';
+import {configureSandbox} from './sandbox.js';
 import {execFile, readAsset} from './util.js';
 
-describeInstallation(
-  '`puppeteer` with Webpack',
-  {
+describe('``puppeteer` with Webpack`', () => {
```
Lightning00Blade:
```suggestion
describe('`puppeteer` with Webpack', () => {
```

test/installation/src/puppeteer.spec.ts
```diff
@@ -18,30 +18,29 @@ import assert from 'assert';
...
 import {join} from 'path';
 
-import {describeInstallation} from './describeInstallation.js';
+import {configureSandbox} from './sandbox.js';
 import {readAsset} from './util.js';
 
-describeInstallation(
-  '`puppeteer`',
-  {
+describe.only('`puppeteer`', () => {
```
Lightning00Blade:
```suggestion
describe('`puppeteer`', () => {
```

test/installation/src/puppeteer-firefox.spec.ts
```diff
@@ -18,31 +18,30 @@ import assert from 'assert';
...
 import {join} from 'path';
 
-import {describeInstallation} from './describeInstallation.js';
+import {configureSandbox} from './sandbox.js';
 import {readAsset} from './util.js';
 
-describeInstallation(
-  '`puppeteer` with Firefox',
-  {
+describe('``puppeteer` with Webpack`', () => {
```
Lightning00Blade:
```suggestion
describe('`puppeteer` with Webpack', () => {
```

OrKoN:
Needs more consideration

Lightning00Blade (approved):

jrandolf-google (approved):

PR #9950, opened 2023-03-31
fix: use shadowRoot for tree walker

jrandolf-google:
Fixes an issue where deep descendent selectors do not go into shadow root of a host element.


Lightning00Blade (approved):

OrKoN (changes requested):
let's add a test for this case?

jrandolf:
> let's add a test for this case?

Done

PR #9954, opened 2023-04-01
fix: cannot find startswith of undefined

shubham-shewale:
bugfix

No
NA
**Summary**
As we have configured selectors based upon object keys in some condition object properties is not defined, thus the selector resolves to undefined, i.e. await page.waitForSelector(undefined); This results in the following error:
Cannot read property 'startsWith' of undefined

No

**Other information**


shubham-shewale:
Closing !! as not in a CLA list

PR #9969, opened 2023-04-04
refactor: prefer use of DeferPromise

Lightning00Blade:


packages/puppeteer-core/src/common/HTTPRequest.ts
```diff
@@ -192,26 +192,6 @@ export class HTTPRequest extends BaseHTTPRequest {
     return this._redirectChain.slice();
   }
 
-  /**
```
Lightning00Blade, line 195: This is already on the Base class, and is ignore for Doc as this class is internal.

OrKoN (approved):

PR #9974, opened 2023-04-05
chore: test server for browsers

OrKoN:
Browser binaries are now downloaded once and cached. The tests download using a local server that mirrors the real server.


packages/browsers/.mocharc.cjs
```diff
@@ -3,4 +3,5 @@ module.exports = {
   spec: 'test/build/**/*.spec.js',
   exit: !!process.env.CI,
   reporter: 'spec',
+  timeout: '10000ms',
```
Lightning00Blade: Let's be consistent with timing we should use number + underscore (Maybe a EsLint Rule would be nice - quick search -> [Fixer](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/numeric-separators-style.md), let's discuss offline).
```suggestion
  timeout: 10_000,
```

packages/browsers/README.md
```diff
@@ -40,3 +40,13 @@ Known limitations:
 ## API
 
 The programmatic API allows installing and launching browsers from your code. See the `test` folder for examples on how to use the `install`, `canInstall`, `launch`, `computeExecutablePath`, `computeSystemExecutablePath` and other methods.
+
+## Testing instructions
+
+To test the package download the browser binaries to the cache folder first:
```
Lightning00Blade: Can we make this part of `test` command? We already skip the downloading if the file path exits.

Note: Testing should have a parent header with name of `Contributing`

Lightning00Blade (approved):
LGTM with 2 small comments.

PR #9976, opened 2023-04-05
chore: improve debugging for EBUSY errors on Windows

OrKoN:
I suspect that some default behaviours in the binaries force the process/sub-processes to linger around for longer making the folder not deletable. Carrying over the default flags from Puppeteer should help.


packages/browsers/test/src/firefox/launcher.spec.ts
```diff
@@ -59,18 +61,46 @@ describe('Firefox', () => {
...
+        switch (os.platform()) {
+          case 'darwin':
+            firefoxArguments.push('--foreground');
+            break;
+          case 'win32':
+            firefoxArguments.push('--wait-for-browser');
+            break;
+        }
+        firefoxArguments.push('--profile');
+        firefoxArguments.push(userDataDir);
```
Lightning00Blade: Nit: 
```suggestion
        firefoxArguments.push('--profile', userDataDir);
```

Lightning00Blade (approved):

PR #9977, opened 2023-04-05
chore: deprecated `puppeteer-core` versions

Lightning00Blade:


OrKoN (changes requested):
We don’t deprecate puppeteer-core because it does not download binaries so I think this change is not needed.

PR #9986, opened 2023-04-06
docs: website docs for browsers

OrKoN:


packages/browsers/src/main.ts
```diff
@@ -19,17 +19,21 @@ export {
   computeExecutablePath,
   computeSystemExecutablePath,
   TimeoutError,
+  LaunchOptions,
+  Options,
```
Lightning00Blade, line 23: What `Options` are these exactly, can we give them a clearer name?

OrKoN: In a follow-up.

Lightning00Blade (approved):

PR #9990, opened 2023-04-06
feat(browsers): support downloading chromedriver

OrKoN:


packages/browsers/src/browser-data/chromedriver.ts
```diff
@@ -0,0 +1,93 @@
...
+      return 'chromedriver';
+    case BrowserPlatform.WIN32:
+    case BrowserPlatform.WIN64:
+      return 'chromedriver.exe';
+  }
+}
+export async function resolveBuildId(
+  _channel: 'latest' = 'latest'
+): Promise<string> {
+  return new Promise((resolve, reject) => {
```
OrKoN, line 67: TODO: in the future, we should extra this into a helper an possibly use the fetch API.

packages/browsers/src/browser-data/browser-data.ts
```diff
@@ -102,6 +111,7 @@ export function resolveSystemExecutablePath(
   channel: ChromeReleaseChannel
 ): string {
   switch (browser) {
+    case Browser.CHROMEDRIVER:
     case Browser.FIREFOX:
       throw new Error(
         'System browser detection is not supported for Firefox yet.'
```
Lightning00Blade:
```suggestion
        `System browser detection is not supported for ${browser} yet.`
```

Lightning00Blade (approved):

PR #10008, opened 2023-04-12
chore: stop CI if search crawler is blocked

Lightning00Blade:
Crawler may get stuck. For example we recently broke the site on accident an that stopped if for the past week. Also atm we have more then 10% changes which stop it as well. This will let us know and incentivize action.

Note: Doc change to be reverted.


OrKoN (approved):

OrKoN:
@Lightning00Blade looks like it fails

PR #10009, opened 2023-04-12
chore: implement locators with a click

OrKoN:
This PR implements a locator API and a click operation. The click is automatically retried and the following pre-conditions are checked automatically:

          1. ensureElementIsInTheViewport
          2. waitForVisibility
          3. waitForEnabled
          4. waitForStableBoundingBox
          
All scheduled tasks can aborted using AbortController.

The API is not public and I think we should make it public once we have at least a few other actions.

### DIFF
--- a/packages/puppeteer-core/src/api/Locator.ts
+++ b/packages/puppeteer-core/src/api/Locator.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import {AbortError, TimeoutError} from '../common/Errors.js';
+import {TimeoutError} from '../common/Errors.js';
 import {EventEmitter} from '../common/EventEmitter.js';
 import {debugError} from '../common/util.js';
 import {isErrorLike} from '../util/ErrorLike.js';
@@ -34,14 +34,16 @@ export interface LocatorOptions {
    * Total timeout for the entire locator operation.
    */
   timeout: number;
-  /**
-   * Timeout for individual operations inside the locator. On errors the
-   * operation is retried as long as {@link LocatorOptions.timeout} is not
-   * exceeded. This timeout should be generally much lower as locating an
-   * element means multiple asynchronious operations.
-   */
-  operationTimeout: number;
 }
+
+/**
+ * Timeout for individual operations inside the locator. On errors the
+ * operation is retried as long as {@link LocatorOptions.timeout} is not
+ * exceeded. This timeout should be generally much lower as locating an
+ * element means multiple asynchronious operations.
+ */
+const CONDITION_TIMEOUT = 1_000;
+const WAIT_FOR_FUNCTION_DELAY = 100;
 
 /**
  * @internal
@@ -96,7 +98,6 @@ export class Locator extends EventEmitter {
     options: LocatorOptions = {
       visibility: 'visible',
       timeout: page.getDefaultTimeout(),
-      operationTimeout: 1000,
     }
   ) {
     super();
@@ -132,10 +133,9 @@ export class Locator extends EventEmitter {
   async #waitForFunction(
     fn: (signal: AbortSignal) => unknown,
     signal?: AbortSignal,
-    timeout = this.#options.operationTimeout
+    timeout = CONDITION_TIMEOUT
   ): Promise<void> {
     let isActive = true;
-    let isUserAborted = false;
     let controller: AbortController;
     // If the loop times out, we abort only the last iteration's controller.
     const timeoutId = setTimeout(() => {
@@ -147,11 +147,11 @@ export class Locator extends EventEmitter {
       'abort',
       () => {
         controller?.abort();
-        isUserAborted = true;
+        isActive = false;
       },
       {once: true}
     );
-    while (isActive && !isUserAborted) {
+    while (isActive) {
       controller = new AbortController();
       try {
         const result = await fn(controller.signal);
@@ -167,7 +167,7 @@ export class Locator extends EventEmitter {
             continue;
           }
           // Abort error are ignored as they only affect one iteration.
-          if (err instanceof AbortError) {
+          if (err.name === 'AbortError') {
             continue;
           }
         }
@@ -178,12 +178,10 @@ export class Locator extends EventEmitter {
         controller.abort();
       }
       await new Promise(resolve => {
-        return setTimeout(resolve, 100);
+        return setTimeout(resolve, WAIT_FOR_FUNCTION_DELAY);
       });
     }
-    if (isUserAborted) {
-      throw new AbortError(`waitForFunction was aborted.`);
-    }
+    signal?.throwIfAborted();
     throw new TimeoutError(
       `waitForFunction timed out. The timeout is ${timeout}ms.`
     );
@@ -196,25 +194,20 @@ export class Locator extends EventEmitter {
     element: ElementHandle,
     signal?: AbortSignal
   ): Promise<void> => {
-    function checkAbortSignal() {
-      if (signal?.aborted) {
-        throw new AbortError(`ensureElementIsInTheViewport was aborted.`);
-      }
-    }
     // Side-effect: this also checks if it is connected.
     const isIntersectingViewport = await element.isIntersectingViewport({
       threshold: 0,
     });
-    checkAbortSignal();
+    signal?.throwIfAborted();
     if (!isIntersectingViewport) {
       await element.scrollIntoView();
-      checkAbortSignal();
+      signal?.throwIfAborted();
       await this.#waitForFunction(async () => {
         return await element.isIntersectingViewport({
           threshold: 0,
         });
       }, signal);
-      checkAbortSignal();
+      signal?.throwIfAborted();
     }
   };
 
@@ -254,7 +247,7 @@ export class Locator extends EventEmitter {
         return true;
       },
       {
-        timeout: this.#options.operationTimeout,
+        timeout: CONDITION_TIMEOUT,
         signal,
       },
       element
@@ -310,17 +303,12 @@ export class Locator extends EventEmitter {
     action: (el: ElementHandle) => Promise<void>,
     options?: ActionOptions
   ) {
-    function checkAbortSignal() {
-      if (options?.signal?.aborted) {
-        throw new AbortError(`Locator was aborted.`);
-      }
-    }
     await this.#waitForFunction(
       async signal => {
         // 1. Select the element without visibility checks.
         const element = await this.#page.waitForSelector(this.#selector, {
           visible: false,
-          timeout: this.#options.operationTimeout,
+          timeout: this.#options.timeout,
           signal,
         });
         // Retry if no element is found.
@@ -328,14 +316,14 @@ export class Locator extends EventEmitter {
           return false;
         }
         try {
-          checkAbortSignal();
+          signal?.throwIfAborted();
           // 2. Perform action specific checks.
           await Promise.all(
             options?.conditions.map(check => {
               return check(element, signal);
             }) || []
           );
-          checkAbortSignal();
+          signal?.throwIfAborted();
           // 3. Perform the action
           this.emit(LocatorEmittedEvents.Action);
           await action(element);
--- a/packages/puppeteer-core/src/api/Locator.ts
+++ b/packages/puppeteer-core/src/api/Locator.ts
@@ -16,11 +16,10 @@
 
 import {AbortError, TimeoutError} from '../common/Errors.js';
 import {EventEmitter} from '../common/EventEmitter.js';
-import type {MouseButton} from '../common/Input.js';
 import {debugError} from '../common/util.js';
 import {isErrorLike} from '../util/ErrorLike.js';
 
-import {ElementHandle, BoundingBox} from './ElementHandle.js';
+import {ElementHandle, BoundingBox, ClickOptions} from './ElementHandle.js';
 import type {Page} from './Page.js';
 
 /**
@@ -137,25 +136,25 @@ export class Locator extends EventEmitter {
   ): Promise<void> {
     let isActive = true;
     let isUserAborted = false;
-    let iterationController: AbortController;
+    let controller: AbortController;
     // If the loop times out, we abort only the last iteration's controller.
     const timeoutId = setTimeout(() => {
       isActive = false;
-      iterationController?.abort();
+      controller?.abort();
     }, timeout);
     // If the user's signal aborts, we abort the last iteration and the loop.
     signal?.addEventListener(
       'abort',
       () => {
-        iterationController?.abort();
+        controller?.abort();
         isUserAborted = true;
       },
       {once: true}
     );
     while (isActive && !isUserAborted) {
-      iterationController = new AbortController();
+      controller = new AbortController();
       try {
-        const result = await fn(iterationController.signal);
+        const result = await fn(controller.signal);
         if (result) {
           clearTimeout(timeoutId);
           return;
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+    eventName: K,
+    handler: (event: LocatorEventObject[K]) => void
+  ): Locator {
+    return super.on(eventName, handler) as Locator;
+  }
+
+  override once<K extends keyof LocatorEventObject>(
+    eventName: K,
+    handler: (event: LocatorEventObject[K]) => void
+  ): Locator {
```
Lightning00Blade, line 119: EventEmitter should return `this` not `EventEmitter`. Happy to change it in this PR or I can make a change but please add `TODO`

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+
+  /**
+   * If the element is a button, textarea, input or select, wait till the
+   * element becomes enabled.
+   */
+  #waitForEnabled = async (
+    element: ElementHandle,
+    _signal?: AbortSignal
+  ): Promise<void> => {
+    // TODO: use AbortSignal in waitForFunction.
```
Lightning00Blade: We have this after #10078, should we wait for this PR to implement it?

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+          }
+        }
+        throw err;
+      } finally {
+        // We abort any operations that might have been started by `fn`, because
+        // the iteration is now over.
+        iterationController.abort();
+      }
+      await new Promise(resolve => {
+        return setTimeout(resolve, 100);
```
Lightning00Blade: This delay for in between the retries right? Should we expose this to the users?

test/src/locator.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+        await button?.evaluate(el => {
+          return el.innerText;
+        })
+      ).toBe('clicked');
+    });
+
+    it('should time out', async () => {
+      let clock: sinon.SinonFakeTimers | undefined;
+      try {
+        clock = sinon.useFakeTimers();
```
Lightning00Blade: Why does this need to be inside the try block?

test/src/locator.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+      }
+    });
+
+    it('should retry clicks on errors', async () => {
+      const {page} = getTestState();
+      let clock: sinon.SinonFakeTimers | undefined;
+      try {
+        clock = sinon.useFakeTimers();
+
+        await page.setDefaultTimeout(5000);
```
Lightning00Blade:
```suggestion
        page.setDefaultTimeout(5000);
```

test/src/locator.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+      ).toBe('clicked');
+    });
+
+    it('should time out', async () => {
+      let clock: sinon.SinonFakeTimers | undefined;
+      try {
+        clock = sinon.useFakeTimers();
+        const {page} = getTestState();
+
+        await page.setDefaultTimeout(5000);
```
Lightning00Blade:
```suggestion
        page.setDefaultTimeout(5000);
```

test/src/locator.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+          <button style="display: none;" onclick="this.innerText = 'clicked';">test</button>
+        `);
+        const abortController = new AbortController();
+        const result = page.locator('button').click({
+          signal: abortController.signal,
+        });
+        clock.tick(2000);
+        abortController.abort();
+        await expect(result).rejects.toEqual(
+          new Error('waitForFunction was aborted.')
```
Lightning00Blade: AbortError?

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+          // Retry on all timeouts.
+          if (err instanceof TimeoutError) {
+            continue;
+          }
+          // Abort error are ignored as they only affect one iteration.
+          if (err instanceof AbortError) {
+            continue;
+          }
+          // Ignore error if the user's signal aborted.
+          if (signal?.aborted) {
```
Lightning00Blade: Does order matter here? Is it possible to get an `TimeoutError` but the user aborted then it will still rerun?

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+   * Total timeout for the entire locator operation.
+   */
+  timeout: number;
+  /**
+   * Timeout for individual operations inside the locator. On errors the
+   * operation is retried as long as {@link LocatorOptions.timeout} is not
+   * exceeded. This timeout should be generally much lower as locating an
+   * element means multiple asynchronious operations.
+   */
+  operationTimeout: number;
```
jrandolf: We should remove this in favor of just `timeout`. If a specific operation needs to be timed out, then there should be a `timeout` on that specific operation.

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+  signal?: AbortSignal;
+  conditions: ActionCondition[];
+}
+
+/**
+ * All the events that a locator instance may emit.
+ *
+ * @internal
+ */
+export enum LocatorEmittedEvents {
```
jrandolf, line 69:
```suggestion
export enum LocatorEmittedEvent {
```

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+   * Retries the `fn` until a truthy result is returned.
+   */
+  async #waitForFunction(
+    fn: (signal: AbortSignal) => unknown,
+    signal?: AbortSignal,
+    timeout = this.#options.operationTimeout
+  ): Promise<void> {
+    let isActive = true;
+    let isUserAborted = false;
+    let iterationController: AbortController;
```
jrandolf:
```suggestion
    let controller: AbortController;
```

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+      () => {
+        iterationController?.abort();
+        isUserAborted = true;
+      },
+      {once: true}
+    );
+    while (isActive && !isUserAborted) {
+      iterationController = new AbortController();
+      try {
+        const result = await fn(iterationController.signal);
```
jrandolf: What about the user's signal?

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+   * means that the element has a computed style, the visibility property other
+   * than 'hidden' or 'collapse' and non-empty bounding box. visibility ===
+   * 'hidden' means the opposite of that.
+   */
+  #waitForVisibility = async (
+    element: ElementHandle,
+    signal?: AbortSignal
+  ): Promise<void> => {
+    await this.#waitForFunction(async () => {
+      return this.#options.visibility === 'hidden'
```
jrandolf:
```ts
if ( this.#options.visibility === 'hidden') {
return this.#waitForFunction(async () => {
  return element.isHidden();
});
} 
return ...
```

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+      },
+      element
+    );
+  };
+
+  /**
+   * Compares the bounding box of the element for two consecutive animation
+   * frames and waits till they are the same.
+   */
+  #waitForStableBoundingBox = async (
```
jrandolf, line 261: Nice!

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+    }, signal);
+  };
+
+  async #performAction(
+    payloadFn: (el: ElementHandle) => Promise<void>,
+    actionOptions?: ActionOptions
+  ) {
+    function checkAbortSignal() {
+      if (actionOptions?.signal?.aborted) {
+        throw new Error(`Locator was aborted.`);
```
jrandolf:
```suggestion
        throw new AbortError(`Locator was aborted.`);
```

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+        rect1.y === rect2.y &&
+        rect1.width === rect2.width &&
+        rect1.height === rect2.height
+      );
+    }, signal);
+  };
+
+  async #performAction(
+    payloadFn: (el: ElementHandle) => Promise<void>,
+    actionOptions?: ActionOptions
```
jrandolf:
```suggestion
    options?: ActionOptions
```

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+        rect1.x === rect2.x &&
+        rect1.y === rect2.y &&
+        rect1.width === rect2.width &&
+        rect1.height === rect2.height
+      );
+    }, signal);
+  };
+
+  async #performAction(
+    payloadFn: (el: ElementHandle) => Promise<void>,
```
jrandolf:
```suggestion
    action: (el: ElementHandle) => Promise<void>,
```

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+    actionOptions?: ActionOptions
+  ) {
+    function checkAbortSignal() {
+      if (actionOptions?.signal?.aborted) {
+        throw new Error(`Locator was aborted.`);
+      }
+    }
+    await this.#waitForFunction(
+      async iterationSignal => {
```
jrandolf:
```suggestion
      async signal => {
```

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+        } finally {
+          void element.dispose().catch(debugError);
+        }
+      },
+      actionOptions?.signal,
+      this.#options.timeout
+    );
+  }
+
+  async click(clickOptions?: {
```
jrandolf:
```suggestion
  async click(options?: ClickOptions)
```

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+      return (
+        rect1.x === rect2.x &&
+        rect1.y === rect2.y &&
+        rect1.width === rect2.width &&
+        rect1.height === rect2.height
+      );
+    }, signal);
+  };
+
+  async #performAction(
```
jrandolf:
```suggestion
  async #run(
```

jrandolf-google (changes requested):

Lightning00Blade (changes requested):

packages/puppeteer-core/src/api/Locator.ts
OrKoN: Could you elaborate why?

packages/puppeteer-core/src/api/Locator.ts
OrKoN: that suggestion does not follow our existing conventions.

packages/puppeteer-core/src/api/Locator.ts
OrKoN: on line 147, if the user signal is aborted, the iteration controller is aborted too.

packages/puppeteer-core/src/api/Locator.ts
OrKoN: it returns `this`, do you mean the types? I feel that needs to be part of EventEmitter type-safety. Do you want to file an issue and track it there?

packages/puppeteer-core/src/api/Locator.ts
OrKoN: yeah, I need to update

packages/puppeteer-core/src/api/Locator.ts
OrKoN: I am not sure yet, what is your idea? I feel like we can choose it for a user as it is an implementation detail.

packages/puppeteer-core/src/api/Locator.ts
OrKoN: I think it is actually a left-over from the previous iteration and we should not need it.

packages/puppeteer-core/src/api/Locator.ts
OrKoN: My reasoning: we cannot have a single timeout setting. With a single timeout setting, we either make the overall waiting time unpredictable or we only ever run a single waitFor* without any retries because the first timeout would cancel the entire operation. I believe we need shorter timeout for the operations we retry. The constructor will be internal so it is not exposed to the users.

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,374 @@
...
+/**
+ * @internal
+ */
+export interface LocatorOptions {
+  /**
+   * Whether to wait for the element to be `visible` or `hidden`.
+   */
+  visibility: 'hidden' | 'visible';
+  /**
+   * Total timeout for the entire locator operation.
```
jrandolf, line 34: What is an `operation`? Does this consist of both an action and wait for selection? Or just the selection?

packages/puppeteer-core/src/api/Locator.ts
jrandolf: Why would we need to try multiple `waitFor`s?

OrKoN: Consider the following scenario (let's assume we only to waitForSelector + waitForVisible). The user loads a page and tells us to click the visible element matching the selector. So the page is loading and the first element to be rendered in the DOM is matching the selector and is not visible. If we have used only a single timeout setting, that would be the end of the locator, we would throw an error because the element we grabbed first will never become visible. With smaller timeouts for individual waitFor operations, we would time out and retry waitForSelector + waitForVisible again but this time more of the page was rendered into the DOM and a matching element was appended before the element we matched at first, that element became visible and, thus, satisfies the locator requirements. 

`operation` here means anything we do internally inside locator that can timeout. `timeout` is the user's timeout for the entire locator action.

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -0,0 +1,372 @@
...
+  #selector: string;
+  #options: LocatorOptions;
+
+  constructor(
+    page: Page,
+    selector: string,
+    options: LocatorOptions = {
+      visibility: 'visible',
+      timeout: page.getDefaultTimeout(),
+      operationTimeout: 1000,
```
OrKoN: TODO: do not expose in the constructor

OrKoN: TODO: rename to conditionTimeout

packages/puppeteer-core/src/api/Locator.ts
OrKoN: Discussed offline.

packages/puppeteer-core/src/api/Locator.ts
OrKoN: Discussed offline.

jrandolf-google (approved):

test/src/locator.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+        await page.setContent(`
+          <button style="display: none;" onclick="this.innerText = 'clicked';">test</button>
+        `);
+        const result = page.locator('button').click();
+        clock.tick(5100);
+        await expect(result).rejects.toEqual(
+          new TimeoutError('waitForFunction timed out. The timeout is 5000ms.')
+        );
+      } finally {
+        clock?.restore();
```
Lightning00Blade: Nit: clock should always be set no need for `?`.

Lightning00Blade (approved):

PR #10018, opened 2023-04-13
feat: support AbortController in waitForSelector

### LINKED ISSUE #9709: [Feature]: ability to cancel "waitFor" promises
### Feature description

I'm using `page.waitForSelector` in this way to determine which page I landed on:

```typescript

const element = await Promise.race([page.waitForSelector("#login-success"), page.waitForSelector("#login-error")]);

// do something...
```

The problem with this is that if one of the promises completes early, the others will keep on retrying until their timeout expires (I assume waitFor is internally implemented using setTimeout/setInterval). That can create a lot of unnecessary computation especially for a long array of `page.waitForSelector`.

It would be good if there was a way to cancel the promise it returns. For example, it could enable this:

```typescript
const promises = [page.waitForSelector("#login-success"), page.waitForSelector("#login-error")];
const element = await Promise.race(promises);
promises.forEach(p => p.cancel?())
```

I would be able to work on  a pull request if there is interest.

OrKoN:
Closes #9709


OrKoN:
Blocked on deprecation of Node 14.

jrandolf-google (approved):

OrKoN:
That has been already addressed in https://github.com/puppeteer/puppeteer/pull/10048

PR #10021, opened 2023-04-13
fix: improve mouse actions

jrandolf-google:
This PR implements better state management of the mouse. This allows us to offer the mouse `buttons` attribute. This also improves processing speed by queuing them immediately.


packages/puppeteer-core/src/common/Input.ts
```diff
@@ -334,14 +334,16 @@ export class Keyboard {
 /**
  * @public
  */
-export type MouseButton = 'left' | 'right' | 'middle' | 'back' | 'forward';
+export interface MouseOptions {
+  button?: MouseButton;
```
OrKoN, line 343: let's add a comment explaining how it works.

packages/puppeteer-core/src/common/Input.ts
```diff
@@ -334,14 +334,16 @@ export class Keyboard {
...
+  clickCount?: number;
```
OrKoN, line 349: ditto

packages/puppeteer-core/src/common/Input.ts
```diff
@@ -334,14 +334,16 @@ export class Keyboard {
...
+}
 
 /**
  * @public
  */
-export interface MouseOptions {
-  button?: MouseButton;
-  clickCount?: number;
+export interface MouseClickOptions extends MouseOptions {
+  delay?: number;
```
OrKoN, line 359: ditto

packages/puppeteer-core/src/common/Input.ts
```diff
@@ -439,87 +499,109 @@ export class Mouse {
...
   async up(options: MouseOptions = {}): Promise<void> {
-    const {button = 'left', clickCount = 1} = options;
-    this.#button = 'none';
+    const {button = MouseButton.Left, clickCount = 1} = options;
+    const flag = getFlag(button);
+    if (!flag) {
+      throw new Error(`Unsupported mouse button: ${button}`);
+    }
+    if (!(this.#buttons & flag)) {
+      throw new Error(`'${button}' is not pressed.`);
```
OrKoN, line 642: let's add a test for this branch: that the error is reported

test/src/mouse.spec.ts
```diff
@@ -259,4 +259,96 @@ describe('Mouse', function () {
 
     expect(await page.evaluate('result')).toEqual({x: 30, y: 40});
   });
+  it('should throw if buttons are pressed incorrectly', async () => {
```
OrKoN, line 262: that looks like a potentially breaking change? was it throwing before? at least let's mark it as a non-breaking feature so that it shows up in the changelog.

jrandolf: It wasn't throwing before, but it would no-op. This makes this explicit.

jrandolf: I've set this to a fix since this is technically the correct behavior. Is that fine?

OrKoN: sgtm!

OrKoN:
Tests are still failing. I will wait for CI to be green before taking the final look.

packages/puppeteer-core/src/common/Input.ts
```diff
@@ -439,87 +512,134 @@ export class Mouse {
...
+            type: 'mouseMoved',
+            ...this.#position,
+            modifiers: this.#keyboard._modifiers,
+            buttons: this.#buttons,
+            // This should always be 0 (i.e. 'left'). See
+            // https://w3c.github.io/uievents/#event-type-mousemove
+            button: MouseButton.Left,
+          });
+        } catch (error) {
+          this.#position = lastPosition;
```
OrKoN: Let's say we have two move points: p1 & p2. After we start Promise.all, this.#position is equal p2. Now let's assume that that dispatchMouseEvent for p2 succeeds first and call for p1 is rejected. The this.#position is then incorrectly set to p1 whereas the mouse is at p2?

packages/puppeteer-core/src/common/Input.ts
```diff
@@ -439,87 +512,134 @@ export class Mouse {
...
-    this.#x = x;
-    this.#y = y;
+    const points: Point[] = [];
     for (let i = 1; i <= steps; i++) {
+      points.push({
+        x: this.#position.x + (x - this.#position.x) * (i / steps),
+        y: this.#position.y + (y - this.#position.y) * (i / steps),
+      });
+    }
+    await Promise.all(
```
OrKoN: why don't we keep it as in the previous version? when the position is updated before all sync calls have been done?

OrKoN: We can also rollback the position if the last dispatchMouseCall didn't succeed.

jrandolf: I've updated the code so the state management is more transparent.

packages/puppeteer-core/src/common/Input.ts
jrandolf: Obsolete

packages/puppeteer-core/src/common/Input.ts
```diff
@@ -438,90 +519,141 @@ export class Mouse {
...
-   * @param x - Horizontal position of the mouse.
-   * @param y - Vertical position of the mouse.
-   * @param options - Optional object. If specified, the `steps` property
-   * sends intermediate `mousemove` events when set to `1` (default).
-   */
-  async move(
-    x: number,
-    y: number,
-    options: {steps?: number} = {}
+  async #withRollback(
```
OrKoN: So is my understanding correct for the following case?

Two events are dispatched in one task: mousedown, mousemove. Then events happens in the following order:

```
1) mousdown resolves
2) previous state = current state =  (mousedown + mousemove) at line 531
3) mousemove rejects
4) state = previous state = mousedown + mousemove at line 533
```

So it looks like rollback state is not correct since it has to only include the mousedown state.

jrandolf: Fixed.

packages/puppeteer-core/src/common/Input.ts
```diff
@@ -438,90 +513,178 @@ export class Mouse {
...
+    options: MouseClickOptions = {}
+  ): Promise<void> {
+    const {delay} = options;
+    const actions: Array<Promise<void>> = [];
+    const {position} = this.#state;
+    if (position.x !== x || position.y !== y) {
+      actions.push(this.move(x, y));
+    }
+    actions.push(this.down(options));
+    if (typeof delay === 'number') {
```
OrKoN, line 677: nit: it could be simplified I think so that actions don't need to be cleared:

```
const delayPromise = typeof delay === 'number' ? ...timeout : Promise.resolve();
actions.push(delayPromise.then(() => this.up(options)));
```

OrKoN (approved):

packages/puppeteer-core/src/common/Input.ts
jrandolf: Ack

packages/puppeteer-core/src/common/Input.ts
```diff
@@ -438,90 +513,178 @@ export class Mouse {
...
-      y: this.#y,
-      modifiers: this.#keyboard._modifiers,
-      clickCount,
+    const {button = MouseButton.Left, clickCount = 1} = options;
+    const flag = getFlag(button);
+    if (!flag) {
+      throw new Error(`Unsupported mouse button: ${button}`);
+    }
+    if (!(this.#state.buttons & flag)) {
+      throw new Error(`'${button}' is not pressed.`);
```
melink14, line 642: Previously, if you called `up` and the button wasn't pressed, this function would not throw an error.

Should this have been a breaking change?

For example, the testing library I use has a `resetMouse` function which calls `up` on all buttons to make sure the mouse is in it's initial state and after this change using that function throws an error.

OrKoN: @jrandolf

jrandolf: So are you requesting the `up` button reset to a no-op?

melink14: My main goal was just to clarify the breaking behavior of the change since I'm not an expert in this code base. 2 weeks later there might not be much to do except maybe updating the release notes so that it's easier for users to find the breaking change.

Perhaps it would be useful if the documentation should be updated to clarify the constraints on calling the methods.

As to the actual design decision, I think it comes down to user expectations. If we think most users calling `up` without calling `down` are doing it by accident then throwing an error is probably better. If not, then maybe a warning would be better (or nothing).

Since users who want to unconditionally reset mouse can use `try...catch` to avoid major problems with the error, it's fine to inconvenience them if they're in the minority imo.

OrKoN: @jrandolf let's improve documentation around this (by calling out the new error more explicitly in the PR description and updating the method docs)?

thepassle: This indeed seems to have broken a package of ours: https://github.com/modernweb-dev/web/issues/2262

Apparently now: "it's an error to call up if the mouse button is not actually pressed"

The code in our package looks as follows: https://github.com/modernweb-dev/web/blob/e0d362fcdbedf6a72a6c78ebc2e3a3cffa922dbc/packages/test-runner-commands/src/sendMousePlugin.ts#L170

Could a maintainer of puppeteer provide some guidance on what would be the fix for this? Apparently we just need to "actually press" the mouse button? What would that look like?

OrKoN:
```
await page.mouse.down({ button: 'left' });
```

should be called before up:

```
await page.mouse.up({ button: 'left' });
```

it looks like you only send up calls without the down calls. Alternatively, it seems you can use `click` to have down/up paired up correctly.

jrandolf: Their `up` calls are being used to just reset the mouse state. WDYT about having a `reset` function on the mouse? This way, the user doesn't need to keep track of the mouse state to reset it.

OrKoN: @jrandolf should we have some sort of reset mechanism for the mouse/keyboard? I can see the use case for tests where it is difficult to know if the tests left the keyboard/mouse in the right state.

jrandolf: Okay, I think the fact we both commented simultaneously implies we should have one. Let me create a PR.

thepassle: Could you ping me when this is implemented and released? Seems like a `reset` function would be great here indeed :)

jrandolf: See https://github.com/puppeteer/puppeteer/pull/10340

PR #6497, opened 2020-10-09
Add param `threshold`  to `isIntersectingViewport`

ron0115:
When I write an e2e test about controlling scrollbar, I found I got visibleRatio = 0.001xxxx，I must add a threshold to finish  this judge in my situation. Looking forward to your comments

PR #10035, opened 2023-04-18
docs: Document options

MarcusOtter:

Adds missing documentation to some options. 

Documentation change, not needed

Yes :)

**Summary**
Just reading the TSDoc in the IDE, it is impossible to know what to provide in these options. I had to search for a while to find examples of people using these options. In fact I couldn't find any examples of the `threshold` on `isIntersectingViewport()` so I looked at the documentation from the original PR that added it: https://github.com/puppeteer/puppeteer/pull/6497/files

No

**Other information**
Technically these options are objects, but my suggested changes talks about the parameters as if there is only one option. Not sure if this matters but yeah, feel free to suggest better options!

Also, I'm sure there are more cases like this, I have not looked very deeply. Just got the ones I encountered so far.

### DIFF
--- a/docs/api/puppeteer.elementhandle.isintersectingviewport.md
+++ b/docs/api/puppeteer.elementhandle.isintersectingviewport.md
@@ -21,10 +21,10 @@ class ElementHandle {
 
 ## Parameters
 
-| Parameter | Type                                                         | Description  |
-| --------- | ------------------------------------------------------------ | ------------ |
-| this      | [ElementHandle](./puppeteer.elementhandle.md)&lt;Element&gt; |              |
-| options   | { threshold?: number; }                                      | _(Optional)_ |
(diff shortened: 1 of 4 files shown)


OrKoN:
Thanks for the contribution but the docs folder is auto-generated based on TypeScript code. I believe you would need to add those descriptions to the TypeScript types instead.

MarcusOtter:
Makes sense, thought that might be the case. I changed the TSDoc now instead of the markdown.

MarcusOtter:
I also snuck in a change to `contributing.md` to clarify that the markdown files in `docs/` should not be changed manually, feel free to revert that commit if you feel this is superfluous. Ready for review again :)

OrKoN (approved):
Thanks!

PR #10044, opened 2023-04-19
refactor: use deferred promise

Lightning00Blade:


packages/puppeteer-core/src/common/Browser.ts
```diff
@@ -32,6 +32,7 @@ import {
 } from '../api/Browser.js';
 import {BrowserContext} from '../api/BrowserContext.js';
 import {Page} from '../api/Page.js';
+import {createDeferredPromise} from '../puppeteer-core.js';
```
OrKoN: I think you need to import without the entry point

Lightning00Blade:
```suggestion
import {createDeferredPromise} from '../util/DeferredPromise.js';
```

OrKoN (approved):

PR #10048, opened 2023-04-20
chore: use AbortSignal instead of AbortController

jrandolf-google:
Usually you pass in the signal only. See https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal for example.


OrKoN (approved):
Good point, thanks!

packages/puppeteer-core/src/common/QueryHandler.ts
```diff
@@ -167,10 +167,10 @@ export class QueryHandler {
       element = await frame.worlds[PUPPETEER_WORLD].adoptHandle(elementOrFrame);
     }
 
-    const {visible = false, hidden = false, timeout, abortController} = options;
+    const {visible = false, hidden = false, timeout, signal} = options;
 
     try {
-      if (options.abortController?.signal.aborted) {
+      if (signal?.aborted) {
```
MoLow, line 173: can use here https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted

PR #10049, opened 2023-04-20
fix: install bindings once

jrandolf-google:
Bindings were being overwritten whenever the script changed. This PR ensures the bindings are installed once on a given context.

Fixed: crbug/1434252


OrKoN:
Tests seem to be failing?

OrKoN (approved):

Issue #7740, opened 2021-11-03
The chromium binary is not available for arm64


cdeutsch:
### Bug description

In an ARM based Docker container (or ARM based Linux):

1. `npm install puppeteer` 

Error:
> The chromium binary is not available for arm64.

This is related to #6622 which wasn't really solved just worked around via Rosetta.

I'm trying to build an ARM based Docker Image that uses BackstopJS/puppeteer. Ref https://github.com/garris/BackstopJS/issues/1300

Are there any plans to build an arm64 version of `chromium`?

### Puppeteer version

10.0.0

### Node.js version

16.3.0

### npm version

7.15.1

### What operating system are you seeing the problem on?

Linux

ghost:
I have arm64 Chromium binaries available at www.chromiumforlambda.com

talisto:
A workaround I'm using is to install chromium with the base image's package manager (e.g. `apk add chromium` if you're using the node alpine image), and then use `ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true` in the Dockerfile before `npm install puppeteer` to tell puppeteer to skip installing chromium.  You may have to change the puppeteer executable path to use the package manager's version of chromium.

cdeutsch:
Thanks @talisto, that's what I've been doing as a work around, but took two days to come up with that. 
https://github.com/garris/BackstopJS/issues/1300#issuecomment-958561525

I'm using puppeteer via Backstop JS and in my scenario Backstop uses a Docker image to do the screenshot comparisons (yields consistent results across OSes)

It would be nice if `npm install puppeteer` worked for multi-platform Docker images instead of trying to figure out how to add chromium yourself.

ljupcospeci:
> It would be nice if npm install puppeteer worked for multi-platform Docker images instead of trying to figure out how to add chromium yourself.

This ^

cjbd:
> A workaround I'm using is to install chromium with the base image's package manager (e.g. `apk add chromium` if you're using the node alpine image), and then use `ENV PUPPETEER_SKIP_CHROMIUM_DOWNLOAD true` in the Dockerfile before `npm install puppeteer` to tell puppeteer to skip installing chromium. You may have to change the puppeteer executable path to use the package manager's version of chromium.

hello, could you please share your `dockerfile`, thank you!

csaltos:
Another workaround is to use a custom browser with something like:

```
const browserPromise = puppeteer.launch({
[...]

PR #10054, opened 2023-04-20
feat!: switch to Chrome for Testing instead of Chromium

OrKoN:
Puppeteer switches to [Chrome for Testing](https://goo.gle/chrome-for-testing) binaries by default. These binaries will be more similar to Chrome binaries than Chromium binaries in the future. Since the binary is different we mark the change as breaking and include a few other breaking changes related to how the browsers are installed:

- rolls Chrome for Testing 112.0.5615.121 (see https://goo.gle/chrome-for-testing).
- removes `BrowserFetcher` that can only deal with Chromium builds in favor of [@puppeteer/browsers](https://pptr.dev/browsers-api/).
- enables `mac-arm64` builds to be available for download (previously experimental).
- removes ToT CI for now since we don't use it so much and we need to find a new way to test with an arbitrary version.
- removes `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` in favor of `PUPPETEER_SKIP_DOWNLOAD` since the flag always affected any browser, not only Chromium.
- removes `PUPPETEER_CHROMIUM_REVISION` in favour of `PUPPETEER_BROWSER_REVISION` since the flag always affected any browser, not only Chromium.
- removes the fallback to `/usr/bin/chromium-browser` on non-`mac-arm64` platforms. Please define the executable path explicitly. 
- removes the bisect script in favor of using browser-specific bisect tools.
- removes compatibility `DOWNLOAD_HOST` rewrites for `chromium-browser-snapshots`. `PUPPETEER_DOWNLOAD_HOST` now works as a base URL.


mathiasbynens (approved):
Thanks for the clear summary in the PR description, and for making the changes in the first place! Nice work.

packages/puppeteer-core/src/node/PuppeteerNode.ts
```diff
@@ -157,17 +153,19 @@ export class PuppeteerNode extends Puppeteer {
    *
    * @remarks
    * Puppeteer can also be used to control the Chrome browser, but it works best
-   * with the version of Chromium downloaded by default by Puppeteer. There is
-   * no guarantee it will work with any other version. If Google Chrome (rather
-   * than Chromium) is preferred, a
+   * with the version of Chrome for Testing downloaded by default by Puppeteer.
```
Lightning00Blade: Nit:
```suggestion
   * with the version of Chrome for Testing downloaded by default.
```

Lightning00Blade (approved):

jrandolf-google (approved):

OrKoN:
Trying a different CfT version to see if PDF on Windows still has issues. Do not land please.

kblok:
@OrKoN, I saw many smalls renames from `browser` to `Chromium` in common classes. That might give the message that Firefox support might end up being deprecated,  is that right?

OrKoN:
@kblok where do you see that? we renamed from `Chromium` to `browser` in many places, so exactly the opposite.

kblok:
@OrKoN oh, sorry. My bad. Removing the comment to avoid confusion. Thank you!

drmrbrewer:
@OrKoN I'm currently installing the [chromium package][1] in my Debian environment to get the chromium binary needed for puppeteer.  The reason I'm installing it myself, rather than relying on the binary bundled with puppeteer, is because [puppeteer doesn't provide one for arm64][2].

It doesn't seem like there's a "chrome for testing" package for Debian, or even a "chrome" package.  Can we continue to use the chromium binary?  Or any thoughts on how I would go about installing the relevant "chrome for testing" binary on Debian?

  [1]: https://packages.debian.org/bullseye/chromium
  [2]: https://github.com/puppeteer/puppeteer/issues/7740

OrKoN:
@drmrbrewer you can use Chromium, but we won't test with Chromium builds anymore. We are working on getting arm64 for Chrome for Testing but I cannot give an estimate.

drmrbrewer:
@OrKoN OK thanks.  That linked issue for arm64 has been open for 18 months now, and has collected over 70 thumbs-up votes, so it seems like there is a need... and this move to "chrome for testing" makes it even more pressing now, because it is now apparently even more difficult to install the right binary on arm64... I doubt that the Debian package library will add a "chrome for testing" cross-platform package because it is much more niche than chromium (according to my understanding).

Ollie1700:
Sorry for resurrecting this thread but it's taken me a while to track this down...

Our build process has always used `PUPPETEER_SKIP_CHROMIUM_DOWNLOAD` even through the years of upgrading, and long past the timestamp of this merge. However on 25th July 2024 our builds started failing and it took me a long time to track it down to the fact it had now started trying to download chromium. I then tracked down this merge wondering if anyone has a clue why the flag would have retroactively stopped working? We are currently using `puppeteer ^21.5.2` (I know, we need to upgrade).

This is more a curiosity as it appears to be fixed now, but did cause client turbulence being unable to deploy for the past 5 days because of this.

My best guess: Did the repo that this version of Puppeteer download chromium from disappear causing issues? I know 21.x is out of date so this makes sense, but hoping anyone can confirm!

rocktheshan:
@OrKoN: We are also facing similar issue since 24th July 2024 where our build process is failing while downloading Chrome for testing. We are using puppeteer 20.0.0 in this build (older version of puppeteer since this is an older release branch of our application). The error we are getting in the build process is as below:
```
[INFO] npm ERR! code 1
[INFO] npm ERR! path /....../app/node_modules/puppeteer
[INFO] npm ERR! command failed
[INFO] npm ERR! command sh -c node install.js
[INFO] npm ERR! ERROR: Failed to set up Chrome r112.0.5615.121! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
[INFO] npm ERR! Error: Download failed: server returned code 404. URL: https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/112.0.5615.121/linux64/chrome-linux64.zip
```

Hitting the above url states _No such object: chrome-for-testing-public/112.0.5615.121/linux64/chrome-linux64.zip_
Seems to me like that the path to the file hosted there might got changed recently.

OrKoN:
@rocktheshan thanks for reporting, I filed https://crbug.com/359837388 (looks like some binaries are missing from the bucket the URL redirects to)

OrKoN:
Unfortunately, it looks like that CfT release was lost during a bucket migration (https://crbug.com/359837388 for details). We recommend upgrading to the latest version of Puppeteer or to 20.1.0 if you cannot easily upgrade yet.

rocktheshan:
Thanks for the update @OrKoN. We'll go for 20.1.0 then.

PR #10067, opened 2023-04-24
fix: infer last pressed button in mouse move

jrandolf-google:
A mouse move event requires the "last pressed" button. For example, see https://source.chromium.org/chromium/chromium/src/+/main:content/browser/renderer_host/input/synthetic_mouse_driver.cc;l=101;bpv=0;bpt=1 and https://source.chromium.org/chromium/chromium/src/+/refs/heads/main:content/browser/renderer_host/input/web_input_event_builders_mac.mm;drc=a61b95c63b0b75c1cfe872d9c8cdf927c226046e;bpv=1;bpt=1;l=221

More precisely, we need to use the mouse button with the lowest flag value. The reason is not clear, however it's assumed there is internal state that requires the mouse move event to match the latest pressed event.

### DIFF
--- a/packages/puppeteer-core/src/common/Input.ts
+++ b/packages/puppeteer-core/src/common/Input.ts
@@ -391,7 +391,7 @@ export const MouseButton = Object.freeze({
   Middle: 'middle',
   Back: 'back',
   Forward: 'forward',
-});
+}) satisfies Record<string, Protocol.Input.MouseButton>;
 
 /**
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

OrKoN (changes requested):
looks like tests are failing?

OrKoN (approved):

PR #10069, opened 2023-04-24
fix: implement click `count`

jrandolf-google:
The `count` parameter is a new parameter that accounts for the extra number of clicks required to perform a multi-click sequence.

To ensure non-breaking behavior, the `clickCount` parameter retains the same behavior. Eventually `clickCount` will be removed in favor of count.

### DIFF
--- a/packages/puppeteer-core/src/common/Input.ts
+++ b/packages/puppeteer-core/src/common/Input.ts
@@ -342,7 +342,10 @@ export interface MouseOptions {
    */
   button?: MouseButton;
   /**
-   * Determines the click count for the mouse.
+   * @deprecated Use {@link MouseClickOptions.count}.
+   *
+   * Determines the click count for the mouse event. This does not perform
+   * multiple clicks.
    *
    * @defaultValue `1`
    */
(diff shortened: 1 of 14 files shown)


packages/puppeteer-core/src/common/Input.ts
```diff
@@ -538,6 +538,15 @@ export class Mouse {
...
     this.#client = client;
     this.#keyboard = keyboard;
+
+    // Reset the state whenever the frame navigates.
+    client.on('Page.frameNavigated', () => {
+      this.#_state = {
+        position: {x: 0, y: 0},
+        buttons: MouseButtonFlag.None,
+      };
+      this.#transactions = [];
```
OrKoN: what if there are scheduled transactions in there? Are there any other situations when the positions have to be reset?

jrandolf: `transactions` hold current ongoing transactions. It's not really a scheduler, more so a parallel state manager.

It's necessary to clear the transactions here. A navigation forces an empty state, so if there are any ongoing transactions, they will override the empty state if any of them become successful. This can happen if someone queues mouse events in parallel.

OrKoN (approved):

OrKoN (approved):

OrKoN (changes requested):
tests are failing, PTAL

OrKoN (approved):

PR #10039, opened 2023-04-18
feat: add warn for `headless: true`

Lightning00Blade:
Warn for future behavior change of `puppeteer.launch({headless: true})`

The message can be suppressed by setting:
Env -  `PUPPETEER_DISABLE_HEADLESS_WARNING=true`, (or via general logging `PUPPETEER_LOGLEVEL=silent`, `PUPPETEER_LOGLEVEL=error`,)
Config - `logLevel: 'silent'`, `logLevel: 'error'`
NPM LogLevel - https://docs.npmjs.com/cli/v9/using-npm/logging (`silent` or `error`)

Warning:
image

PR #10073, opened 2023-04-25
fix: implement flag for disabling headless warning

### LINKED ISSUE #10070: [Bug]: Repetitive deprecation warnings for `{ headless: true }`
### Bug expectation

We're using the [**Jest** test runner](https://jestjs.io) with the [`jest-puppeteer`](https://www.npmjs.com/package/jest-puppeteer) environment configured

Given that every test run launches headless Chrome once per processor core (minus the main thread), on a 12-core CPU I'm seeing the `{ headless: true }` deprecation warning 11x per test run

This goes even higher when [configured with `browserPerWorker: true`](https://github.com/argos-ci/jest-puppeteer#configuring-jest-puppeteer)

Could you use the Puppeteer cache to only display the warning once per acceptable period?

Console spam from Puppeteer

Their [suggested config](https://github.com/argos-ci/jest-puppeteer#synchronous-configuration) looks like:

```js
/**
 * Puppeteer launch options
 */
launch: {
  headless: process.env.HEADLESS !== 'false'
}
```

Whilst this can't be overridden with `headless: undefined` it's possible to [bypass the nullish check](https://github.com/alphagov/govuk-frontend/pull/3542#discussion_r1175564305) with:

```TypeScript
module.exports = {
  launch: {
    headless: process.env.HEADLESS !== 'false'
      ? 1
      : false,
  }
}
```

We're currently trying this approach until we're ready for `{ headless: 'new' }`:

* https://github.com/alphagov/govuk-frontend/pull/3542

Others might want a way to opt out without silencing all logs

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Error string

Puppeteer old Headless deprecation warning

### Puppeteer configuration

```TypeScript
module.exports = {
  launch: {
    headless: true
  }
}
```

### Puppeteer version

19.11.0

### Node version

18.14.0

### Package manager

npm

### Package manager version

9.6.5

### Operating system

macOS

Lightning00Blade:
Ref #10039.

Adds a ENV variable to disable `old Headless deprecation` warning;
Set `PUPPETEER_DISABLE_HEADLESS_WARNING` to disable the message

Side change:
Don't show warning when LogLevel set to `error`.

Post Merge:
Update the PR above to include the changes.

Closes #10070

### DIFF
--- a/packages/puppeteer-core/src/node/ChromeLauncher.ts
+++ b/packages/puppeteer-core/src/node/ChromeLauncher.ts
@@ -48,7 +48,9 @@ export class ChromeLauncher extends ProductLauncher {
     const headless = options.headless ?? true;
     if (
       headless === true &&
-      this.puppeteer.configuration.logLevel !== 'silent'
+      (!this.puppeteer.configuration.logLevel ||
+        this.puppeteer.configuration.logLevel === 'warn') &&
+      !Boolean(process.env['PUPPETEER_DISABLE_HEADLESS_WARNING'])
     ) {
       console.warn(
         [


packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -48,7 +48,9 @@ export class ChromeLauncher extends ProductLauncher {
     const headless = options.headless ?? true;
     if (
       headless === true &&
-      this.puppeteer.configuration.logLevel !== 'silent'
+      (!this.puppeteer.configuration.logLevel ||
+        this.puppeteer.configuration.logLevel === 'warn') &&
+      !Boolean(process.env['PUPPETEER_DISABLE_OLD_HEADLESS_WARN'])
```
jrandolf: WDYT about PUPPETEER_DISABLE_HEADLESS_WARNING? It's simpler and it's understood from context (since we don't warn about headless anywhere else)

OrKoN (approved):

PR #10080, opened 2023-04-25
fix: add Host header when used with http_proxy

magicdawn:
fix: add Host header when used with http_proxy

a bugfix

**Summary**

no

**Other information**

manual `http.get()` differs from use a `agent = HttpProxyAgent()` (imported from http-proxy-agent package) instance, for `Host` header.

say proxy for `https://httpbin.org/headers`
- use `https.get('https://httpbin.org/headers', { agent: HttpProxyAgent('http://127.0.0.1:54321') })` will have `Host: httpbin.org` sent to `127.0.0.1:54321`
- `https.get` without Host set will use proxy's host `127.0.0.1:54321` which breaks my current proxy server.

I found some evidence indicates it's client's job to make it right.

- https://datatracker.ietf.org/doc/html/rfc2616#section-14.23 
- from http: the definitive guide Book, page 419

> If the URL contains a hostname, the Host header must contain the same name.

> Host headers and proxies
> Some browser versions send incorrect Host headers, especially when configured to use proxies. > For example, when configured to use a proxy, some older versions of Apple and PointCast clients mistakenly sent the name of the proxy instead of the ori- gin server in the Host header.
 

---------

AND why https-proxy-agent but not http-proxy-agent?

### DIFF
--- a/packages/browsers/src/httpUtil.ts
+++ b/packages/browsers/src/httpUtil.ts
@@ -61,6 +61,8 @@ export function httpRequest(
       options.hostname = proxy.hostname;
       options.protocol = proxy.protocol;
       options.port = proxy.port;
+      options.headers ||= {};
+      options.headers.Host ||= url.host;
     } else {
       options.agent = createHttpsProxyAgent({
         host: proxy.host,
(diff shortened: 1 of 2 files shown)


magicdawn:
signed.

OrKoN (approved):

PR #10102, opened 2023-04-30
docs: remove unnecessary escaping which leads invalid url link

wilsonwangdev:
There are some links refering to invalid url because of unnecessary escaping.
Here is the reason: https://poe.com/s/HqX52mDTqK5PDNEKemWy


OrKoN (changes requested):
Thanks for the PR but these docs are generated and we need to fix it in the markdown generator/source code comments where the link comes from. cc @Lightning00Blade @jrandolf

PR #10114, opened 2023-05-02
feat: implement detailed errors for evaluation

jrandolf-google:
This PR allows users to know the exact location their client-side scripts may be throwing.

Note this assumes we are running on V8 where `Error.prepareStackTrace` exists.


packages/puppeteer-core/src/common/util.ts
```diff
@@ -41,30 +41,214 @@ export const debugError = debug('puppeteer:error');
...
+
+/**
+ * @internal
+ */
+export class PuppeteerURL {
+  static INTERNAL_URL = 'pptr:internal';
+
+  static fromCallSite(
+    functionName: string,
+    site: NodeJS.CallSite
```
OrKoN, line 174: I think we should not use the Node-specific types outside of the node subfolder.

OrKoN (commented):
Tests seem to be failing, otherwise LGTM.

packages/puppeteer-core/src/common/util.ts
jrandolf: CallSite is specific to V8 actually, so it's specific to Chrome and NodeJS

OrKoN (approved):

PR #10119, opened 2023-05-03
fix: added clipboard permission that was not exposed

### LINKED ISSUE #10103: [Bug]: The 'clipboard-write' permission doesn't work in Extension pages
### Bug expectation

I expected that the [Clipboard permissions](https://devdocs.io/puppeteer/index#browsercontextoverridepermissionsorigin-permissions) are working in Extensions. It turned out that the writing to the clipboard doesn't work properly.
After giving the 'clipboard-write' permission:
```
  const context = browser.defaultBrowserContext()
  await context.overridePermissions(page.url(), ['clipboard-read', 'clipboard-write'])
```
The `navigator.clipboard.writeText(text)` do nothing in the Extension page context.

Please see the repo that reproduces the issue: https://github.com/dvdvdmt/puppeteer-clipboard-read-bug

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript
// The issue can be reproduced using this repo: https://github.com/dvdvdmt/puppeteer-clipboard-read-bug
```

### Error string

no error

### Puppeteer configuration

_No response_

### Puppeteer version

19.11.1

### Node version

v18.16.0

### Package manager

npm

### Package manager version

9.5.1

### Operating system

macOS

### LINKED ISSUE #10211: [Bug] overridePermissions() not working for clipboard-copy in headless browser
Below is the workaround to the issue. Please resolve the bug in overridePermissions.

 Howdy y'all, in case you're looking for something to copy and paste to enable clipboard permissions, some code that worked to do what @ericcornelissen did, but in the latest version of Puppeteer, is to create a separate CDP session to grant yourself sanitized clipboard permissions. While someone should probably make a PR to expose the `clipboardSanitizedWrite` permissions to `overridePermissions()` this still gets the job done. Note that I couldn't find a way in which puppeteer exposes the `BrowserContextID`, so this only works for the default browser context.

```js
const client = await page.target().createCDPSession();
await client.send('Browser.grantPermissions', {
  origin: "https://command.verkada.com",
  permissions: ['clipboardReadWrite', 'clipboardSanitizedWrite'],
});
```

Alternatively, we can use the CDP API [`Browser.setPermission`](https://chromedevtools.github.io/devtools-protocol/tot/Browser/#method-setPermission). As a side note, it looks like this is the API you'd want to use for `panTiltZoom` cameras and `userVisibleOnly` push notifications.

```js
const client = await page.target().createCDPSession();
await client.send('Browser.setPermission', {
  origin: "https://command.verkada.com",
  permission: {
    name: 'clipboard-write',
    allowWithoutSanitization: true
  },
  setting: 'granted',
});

_Originally posted by @rmichnovicz-verkada in https://github.com/puppeteer/puppeteer/issues/3241#issuecomment-1171770074_

Lightning00Blade:
Closes #10103 
Closes #10211
Did not investigate further on what the differences are but passing this Permission work. A quick search also suggest passing this CDP Permission when dealing with clipboard.
https://tinyurl.com/raw-clipboard-access-design

### DIFF
--- a/docs/api/puppeteer.permission.md
+++ b/docs/api/puppeteer.permission.md
@@ -21,6 +21,7 @@ export type Permission =
   | 'accessibility-events'
   | 'clipboard-read'
   | 'clipboard-write'
+  | 'clipboard-sanitized-write'
   | 'payment-handler'
   | 'persistent-storage'
   | 'idle-detection'
--- a/packages/puppeteer-core/src/api/Browser.ts
+++ b/packages/puppeteer-core/src/api/Browser.ts
@@ -84,6 +84,7 @@ export const WEB_PERMISSION_TO_PROTOCOL_PERMISSION = new Map<
   ['accessibility-events', 'accessibilityEvents'],
   ['clipboard-read', 'clipboardReadWrite'],
   ['clipboard-write', 'clipboardReadWrite'],
+  ['clipboard-sanitized-write', 'clipboardSanitizedWrite'],
   ['payment-handler', 'paymentHandler'],
   ['persistent-storage', 'durableStorage'],
   ['idle-detection', 'idleDetection'],
@@ -108,6 +109,7 @@ export type Permission =
   | 'accessibility-events'
   | 'clipboard-read'
   | 'clipboard-write'
+  | 'clipboard-sanitized-write'
   | 'payment-handler'
   | 'persistent-storage'
   | 'idle-detection'


OrKoN:
Do we have a test?

dvdvdmt:
Hi guys, thank you for taking a look into this! I just wanted to say that if you want a test for this specific Chrome Extension problem then you can try use [this setup](https://github.com/dvdvdmt/puppeteer-clipboard-read-bug).

AnmolBansalDEV:
hey, can any maintainer please provide a timeline in which this change will be merged?

OrKoN (approved):

PR #10121, opened 2023-05-03
chore: implement Frames for BiDi

Lightning00Blade:


packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -0,0 +1,965 @@
...
+    options: WaitTimeoutOptions
+  ): Promise<DeviceRequestPrompt>;
+  waitForDevicePrompt(): Promise<DeviceRequestPrompt> {
+    throw new Error('Not implemented');
+  }
+
+  /**
+   * @internal
+   */
+  _navigated(
```
OrKoN: do we need the methods here and below here?

Lightning00Blade: I will double check, I remember adding this to avoid some typescript fails, I think i can retype some places with the correct CDP Frame, and avoid this.

packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -0,0 +1,915 @@
...
+   * The frame's title.
+   */
+  async title(): Promise<string> {
+    throw new Error('Not implemented');
+  }
+
+  /**
+   * @internal
+   */
+  _deviceRequestPromptManager(): DeviceRequestPromptManager {
```
OrKoN: is it used?

packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -0,0 +1,915 @@
...
+ * the parent {@link Frame.page | page}:
+ *
+ * - {@link PageEmittedEvents.FrameAttached}
+ * - {@link PageEmittedEvents.FrameNavigated}
+ * - {@link PageEmittedEvents.FrameDetached}
+ *
+ * @public
+ */
+export class Frame {
+  /**
```
OrKoN, line 169: I think most of those attributes should not be on the API class

OrKoN (approved):

OrKoN (approved):

PR #10130, opened 2023-05-04
fix: rename PUPPETEER_DOWNLOAD_HOST to PUPPETEER_DOWNLOAD_BASE_URL

### LINKED ISSUE #10131: [Bug]: Compatibility issues with PUPPETEER_DOWNLOAD_HOST
### Bug expectation

In puppeteer@19.x, when we run the following command, the version can be automatically resolved:

`PUPPETEER_DOWNLOAD_HOST=https://cdn.npmmirror.com/binaries/ npm run postinstall`

The main parameters after resolution are:

|type|architecture|buildId |host|
|---|---|---|--|
|chromium| mac| 1108766| https://cdn.npmmirror.com/binaries/chromium-browser-snapshots|

In puppeteer@20.x, when we run the following command, a 404 error will be displayed:

https://cdn.npmmirror.com/binaries/113.0.5672.63/mac-arm64/chrome-mac-arm64.zip

The main parameters after resolution are:

|type|architecture|buildId |host|
|---|---|---|---|
|chrome|mac_arm|113.0.5672.63|https://cdn.npmmirror.com/binaries|

The new version has updated the default browser type from `chromium` to `chrome`, but it is using the same variable `PUPPETEER_DOWNLOAD_HOST` as before, therefore, we cannot differentiate between them.
Would it be possible to add a new variable to allow for customizable overrides? 

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript
`npm i puppeteer`
`cd node_modules/puppeteer`
`PUPPETEER_DOWNLOAD_HOST=https://cdn.npmmirror.com/binaries/ npm run postinstall`
```

### Error string

ERROR: Failed to set up Chrome r113.0.5672.63! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.

### Puppeteer configuration

```TypeScript
PUPPETEER_DOWNLOAD_HOST=https://cdn.npmmirror.com/binaries/
```

### Puppeteer version

20.1.0

### Node version

16.19.0

### Package manager

npm

### Package manager version

9.6.2

### Operating system

macOS

OrKoN:
[The switch to Chrome for Testing](https://github.com/puppeteer/puppeteer/pull/10054) changed the servers we download binaries from so PUPPETEER_DOWNLOAD_HOST would not work as it used to because the rest of the URL structure changed too. This change documents the change and deprecates the PUPPETEER_DOWNLOAD_HOST in favour of PUPPETEER_DOWNLOAD_BASE_URL.

Closes #10131


packages/puppeteer-core/src/common/Configuration.ts
```diff
@@ -56,16 +56,16 @@ export interface Configuration {
...
    * Specifies the URL prefix that is used to download the browser.
    *
-   * Can be overridden by `PUPPETEER_DOWNLOAD_HOST`.
+   * Can be overridden by `PUPPETEER_DOWNLOAD_BASE_URL`.
    *
    * @remarks
    * This must include the protocol and may even need a path prefix.
    *
-   * @defaultValue Either https://storage.googleapis.com or
+   * @defaultValue Either https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing or
```
Lightning00Blade, line 64: Does the url need to be changed in `packages/browsers/src/install` as well?

OrKoN: Yes!

jrandolf:
Can we deprecate then remove in a later release?

Lightning00Blade (approved):
Note: this will push Browsers to `1.0.1` due to the Change being a fix.
Only docs were changed for Browsers, something to keep in mind

Issue #10147, opened 2023-05-09
[Feature]: Support tab targets


OrKoN:
### Feature description

See https://docs.google.com/document/d/14aeiC_zga2SS0OXJd6eIFj8N0o5LGwUpuqa4L8NKoR4/edit?hl=en&forcehl=1#heading=h.n1atlriavj6v

Siphosethu-Sidloyi:
Hi OrKon, 

This feature looks very interesting, I'm am new to Node and to puppeteer and am looking to improve my programming skills, could we work on this feature together? I'm not 100% how I could implement it. Let me know how I can assist

Thanks

OrKoN:
Once the tab targets are fully supported we should enable BackForwardCache and Prerender2 features.

PR #10148, opened 2023-05-09
chore: support Tab targets

OrKoN:
[Tab target is a new type of target in CDP](https://docs.google.com/document/d/14aeiC_zga2SS0OXJd6eIFj8N0o5LGwUpuqa4L8NKoR4/edit) that reflects the architectural changes in Chromium that were needed for features such as bfcache, portal and prerendering.

Previously, the `page` targets were the top-level targets representing the entire tab. If you navigate to a new URL, the page target would remain the same. With the introduction of the `tab` target, it's not possible to have multiple page targets active within one tab. For example, one page would be the currently shown page and the other one would be a [prerendered](https://github.com/WICG/nav-speculation/) page. When the user navigates to the prerendered page, the secondary page becomes the primary and the primary page target gets destroyed (the process is called activation). 

Puppeteer does not support this interaction so this PR disabled the prerendering feature in addition to the previously disabled bfcache.

This change adds support for the tab target behind the flag. Currently, we don't plan on exposing the `tab` target via the Puppeteer API and aim at keeping it an implementation detail. The support for the tab target is also limited for now: we support the target activation but the tracing/network/a11y/coverage might not work as expected if the activation happens. We will aim at improving it in the future. For the scenarios, where no activation happens there should be no difference between the tab target and non-tab target modes.

The current level of support is relatively basic: emulation/tracing/coverage might not work as expected if an activation happens. We will enable the tab mode by default once the emulation/tracing/coverage is handled in a solid way.

Issue #10147

### DIFF
--- a/packages/puppeteer-core/src/common/ChromeTargetManager.ts
+++ b/packages/puppeteer-core/src/common/ChromeTargetManager.ts
@@ -86,6 +86,9 @@ export class ChromeTargetManager extends EventEmitter implements TargetManager {
   #targetsIdsForInit = new Set<string>();
   #waitForInitiallyDiscoveredTargets = true;
 
+  #tabMode = true;
+  #discoveryFilter = this.#tabMode ? [{}] : [{type: 'tab', exclude: true}, {}];
+
   constructor(
     connection: Connection,
     targetFactory: TargetFactory,
@@ -107,7 +110,7 @@ export class ChromeTargetManager extends EventEmitter implements TargetManager {
     this.#connection
       .send('Target.setDiscoverTargets', {
         discover: true,
-        filter: [{type: 'tab', exclude: true}, {}],
+        filter: this.#discoveryFilter,
       })
       .then(this.#storeExistingTargetsForInit)
       .catch(debugError);
@@ -143,6 +146,15 @@ export class ChromeTargetManager extends EventEmitter implements TargetManager {
       waitForDebuggerOnStart: true,
       flatten: true,
       autoAttach: true,
+      filter: this.#tabMode
+        ? [
+            {
+              type: 'page',
+              exclude: true,
+            },
+            ...this.#discoveryFilter,
+          ]
+        : this.#discoveryFilter,
     });
     this.#finishInitializationIfReady();
     await this.#initializeDeferred.valueOrThrow();
@@ -158,7 +170,13 @@ export class ChromeTargetManager extends EventEmitter implements TargetManager {
   }
 
   getAvailableTargets(): Map<string, CDPTarget> {
-    return this.#attachedTargetsByTargetId;
+    const result = new Map<string, CDPTarget>();
+    for (const [id, target] of this.#attachedTargetsByTargetId.entries()) {
+      if (target.type() !== 'tab' && !target._subtype()) {
+        result.set(id, target);
+      }
+    }
+    return result;
   }
 
   addTargetInterceptor(
@@ -285,6 +303,16 @@ export class ChromeTargetManager extends EventEmitter implements TargetManager {
     const wasInitialized =
       target._initializedDeferred.value() === InitializationStatus.SUCCESS;
 
+    if (target._subtype() && !event.targetInfo.subtype) {
+      const target = this.#attachedTargetsByTargetId.get(
+        event.targetInfo.targetId
+      );
+      const session = target?._session();
+      if (session) {
+        session.parentSession()?.emit('sessionswapped', session);
+      }
+    }
+
     target._targetInfoChanged(event.targetInfo);
 
     if (wasInitialized && previousURL !== target.url()) {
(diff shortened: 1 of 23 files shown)


packages/puppeteer-core/src/common/ChromeTargetManager.ts
```diff
@@ -285,6 +315,18 @@ export class ChromeTargetManager extends EventEmitter implements TargetManager {
...
+    if (isPageTargetBecomingPrimary(target, event.targetInfo)) {
+      const target = this.#attachedTargetsByTargetId.get(
+        event.targetInfo.targetId
+      );
+      const session = target?._session();
+      assert(
+        session,
+        'Target that is being activated is missing a CDPSession.'
+      );
+      session.parentSession()?.emit('sessionswapped', session);
```
Lightning00Blade: Use `sessionSwapped`

Nit: Even better then the above use a map for Session events

packages/puppeteer-core/src/common/ChromeTargetManager.ts
```diff
@@ -31,6 +31,17 @@ import {
 } from './TargetManager.js';
 import {debugError} from './util.js';
 
+function isTargetExposed(target: CDPTarget): boolean {
+  return target.type() !== 'tab' && !target._subtype();
```
Lightning00Blade: Shouldnt this use the Enum to compare to, or it does not exist at run time due to the `internal` flag?

packages/puppeteer-core/src/common/FrameManager.ts
```diff
@@ -113,11 +117,66 @@ export class FrameManager extends EventEmitter {
...
+    mainFrame.once(FrameEmittedEvents.FrameSwappedByActivation, () => {
+      swapped.resolve();
+    });
+    try {
+      await swapped.valueOrThrow();
+      for (const child of mainFrame.childFrames()) {
+        this.#removeFramesRecursively(child);
+      }
+    } catch (err) {
       if (mainFrame) {
```
Lightning00Blade: If should not be needed mainFrame should be defined at this branch.

Lightning00Blade (approved):

packages/puppeteer-core/src/common/Page.ts
```diff
@@ -307,6 +310,25 @@ export class CDPPage extends Page {
     this.#viewport = null;
 
     this.#setupEventListeners();
+
+    this.#tabSession?.on(CDPSessionEmittedEvents.Swapped, async newSession => {
```
Lightning00Blade, line 314: Should this be `once`?

packages/puppeteer-core/src/common/FrameManager.ts
```diff
@@ -113,11 +117,64 @@ export class FrameManager extends EventEmitter {
...
+    const swapped = Deferred.create<void>({
+      timeout: TIME_FOR_WAITING_FOR_SWAP,
+      message: 'Frame was not swapped',
+    });
+    mainFrame.once(FrameEmittedEvents.FrameSwappedByActivation, () => {
+      swapped.resolve();
+    });
+    try {
+      await swapped.valueOrThrow();
+      for (const child of mainFrame.childFrames()) {
```
Lightning00Blade: Small optimization would be to remove all the child frames before entering the `try/catch`

packages/puppeteer-core/src/common/FrameManager.ts
```diff
@@ -113,11 +117,64 @@ export class FrameManager extends EventEmitter {
...
+    this.#onExecutionContextsCleared(this.#client);
+
+    this.#client = client;
+    assert(
+      this.#client instanceof CDPSessionImpl,
+      'CDPSession is not an instance of CDPSessionImpl.'
+    );
+    const frame = this._frameTree.getMainFrame();
+    if (frame) {
+      this._frameTree.removeFrame(frame);
```
Lightning00Blade, line 167: Removing it here means we create a new frame in `await this.initialize(client);` and we have a stale double that way.

packages/puppeteer-core/src/common/Page.ts
OrKoN: I think it should be not once since there could be many consecutive swaps during one tab's lifetime

packages/puppeteer-core/src/common/FrameManager.ts
OrKoN: Good point, I will address it tomorrow

packages/puppeteer-core/src/common/FrameManager.ts
OrKoN: Nice finding, I will add explicit tests for this!

packages/puppeteer-core/src/common/FrameManager.ts
```diff
@@ -131,6 +131,9 @@ export class FrameManager extends EventEmitter {
     if (!mainFrame) {
       return;
     }
+    for (const child of mainFrame.childFrames()) {
```
Lightning00Blade, line 134: I think we need to set-up the listeners first as if we have a big tree we may miss it it happens too quickly.

PR #10149, opened 2023-05-09
chore: fix a couple of navigation tests for firefox

lutien:

**Summary**

**Other information**

### DIFF
--- a/test/src/mocha-utils.ts
+++ b/test/src/mocha-utils.ts
@@ -143,6 +143,7 @@ interface PuppeteerTestState {
   isChrome: boolean;
   isHeadless: boolean;
   headless: 'true' | 'false' | 'new';
+  protocol: string;
   puppeteerPath: string;
 }
 const state: Partial<PuppeteerTestState> = {};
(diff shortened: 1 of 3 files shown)


test/src/mocha-utils.ts
```diff
@@ -224,6 +225,7 @@ export const mochaHooks = {
       state.puppeteerPath = path.resolve(
         path.join(__dirname, '..', '..', 'packages', 'puppeteer')
       );
+      state.protocol = protocol;
```
OrKoN, line 228: Perhaps, let's not expose the protocol here as we don't want to grow the number of protocol/browser specific expectations in the tests. Perhaps, let's make sure that the errors are properly wrapped on the Puppeteer-level so that the tests can have a single assertion? cc @Lightning00Blade

lutien: That of course would be even better!

lutien: Should I create maybe an issue for it? 🤔

PR #10152, opened 2023-05-10
chore: create a test helper for launching browser

Lightning00Blade:
Only one test changed at this time (unblocks a BiDi PR for navigation), when approved a follow up PR will make the changes to the other places.

EsLint rule for this created, will be turn on ('error') in the follow up.

### DIFF
--- a/test/src/mocha-utils.ts
+++ b/test/src/mocha-utils.ts
@@ -242,6 +242,10 @@ export const mochaHooks = {
   ],
 
   afterEach: (): void => {
+    if (browserCleanups.length > 0) {
+      throw new Error('A manually launched browser was not closed!');
+    }
+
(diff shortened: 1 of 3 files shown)


test/src/mocha-utils.ts
```diff
@@ -328,3 +332,36 @@ export const createTimeout = <T>(
...
+): Promise<
+  PuppeteerTestState & {
+    close: () => Promise<void>;
+  }
+> => {
+  const browser = await puppeteer.launch({
+    ...defaultBrowserOptions,
+    ...options,
+  });
+  const context = await browser.createIncognitoBrowserContext();
```
OrKoN: what if these two calls fail?

OrKoN (approved):

PR #10158, opened 2023-05-11
feat: support Chrome channels for ChromeDriver

OrKoN:


packages/browsers/src/browser-data/chromedriver.ts
```diff
@@ -54,17 +57,21 @@ export function relativeExecutablePath(
...
   }
 }
 export async function resolveBuildId(
-  _channel: 'latest' = 'latest'
+  channel: ChromeReleaseChannel
 ): Promise<string> {
-  return await getText(
-    new URL(`https://chromedriver.storage.googleapis.com/LATEST_RELEASE`)
-  );
+  if (channel === ChromeReleaseChannel.STABLE) {
```
Lightning00Blade: Why the special case here?
We are downloading exclusively from CtF bucket, could there be version miss match this way?

OrKoN: a left-over, removed!

Lightning00Blade (approved):

PR #10167, opened 2023-05-12
refactor: introduce an internal PageTarget subclass

OrKoN:
In preparation for supporting the tab target, we introduce classes for various target types. This allows encapsulating various target-specific init steps and allows avoid constantly checking the current target type. This PR extracts a PageTarget that would wrap target types that support pages. The next PR will refactor worker targets.

### DIFF
--- a/packages/puppeteer-core/src/common/Target.ts
+++ b/packages/puppeteer-core/src/common/Target.ts
@@ -16,7 +16,7 @@
 
 import {Protocol} from 'devtools-protocol';
 
-import type {Browser, IsPageTargetCallback} from '../api/Browser.js';
+import type {Browser} from '../api/Browser.js';
 import type {BrowserContext} from '../api/BrowserContext.js';
 import {Page, PageEmittedEvents} from '../api/Page.js';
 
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/common/Target.ts
```diff
@@ -271,15 +223,118 @@ export class Target {
...
+        openerPage.emit(PageEmittedEvents.Popup, popupPage);
+        return true;
+      })
+      .catch(debugError);
+    this._checkIfInitialized();
+  }
+
+  override async page(): Promise<Page | null> {
+    if (!this.pagePromise) {
+      const session = this._session();
```
Lightning00Blade, line 315: Can't we await 
```suggestion
      const session = this._session() ?? (await this._sessionFactory()(true));
```
then set `this.pagePromise = CDPPage._create` , I think it would easier to read.

Lightning00Blade (approved):
LGTM + one minor comment

packages/puppeteer-core/src/common/Target.ts
OrKoN: I don't think so because we need pagePromise to be created atomically.

Issue #10174, opened 2023-05-14
[Feature]: Provide a way to remove all obsolete browser revisions in the cache


timvandermeij:
### Feature description

In Puppeteer 20 the `BrowserFetcher` class was removed in favor of the new `@puppeteer/browsers` package, but sadly the latter doesn't seem to provide a way to remove browsers and get all installed browsers/revisions that the former did.

The use case for this is that currently the cache directory can grow indefinitely, which is a problem locally and when running on fixed CI bots, so we need a way to prune old downloaded browsers. This used to be done by Puppeteer before, but was removed in https://github.com/puppeteer/puppeteer/commit/344bc2af62e4068fe2cb8162d4b6c8242aac843b, however we still could do this ourselves with `BrowserFetcher.localRevisions` and `BrowserFetcher.remove`. This is now also not an option anymore since Puppeteer 20; all we can use is the `Cache.clear` method, but that is not fine-grained enough because it will remove _all_ browsers in the cache, including the current ones which we'd then have to download again. This is quite suboptimal, and it leads to solutions like https://github.com/alphagov/govuk-frontend/blob/d915dfccbe178d76ddeee288b565320a91549074/shared/helpers/jest/browser/download.mjs which mostly work but require quite a bit of error-prone orchestration in user code.

It would be much nicer if Puppeteer's browser package provided a simple API to remove all obsolete (non-current) browser revisions. If that is not possible, it should get a similar API to `BrowserFetcher` with `getInstalledBrowsers` and `remove` methods. Note that `getInstalledBrowsers` already exists on the `Cache`, but that doesn't seem like the most logical place to me and seems to fit in much nicer in the browsers package, also to encapsulate the browsers logic a bit more. This would at least provide user code with a simple way to enumerate all revisions and throw away the ones they are not interested in.

timvandermeij:
The `trimCache` functionality looks exactly like what I had in mind here and will make cleaning the cache in user code trivial. Thank you @OrKoN for implementing this; once a release is made I'll be happy to test this over at Mozilla's PDF.js project and then there should be no blockers anymore for upgrading to Puppeteer 20 from our side I think!

OrKoN:
@timvandermeij thanks for taking a look!

timvandermeij:
The `trimCache` functionality works now, but I haven't gotten it working in the way I had expected it to. I tried to use this:

```
const puppeteer = require("puppeteer");
await puppeteer.trimCache()
```

but this always results in the following error:

[...]

PR #10179, opened 2023-05-15
feat: add ability to uninstall a browser

OrKoN:
#10174


packages/browsers/test/src/chrome/install.spec.ts
```diff
@@ -106,6 +107,34 @@ describe('Chrome install', () => {
     assert.deepStrictEqual(browser, installed[0]);
   });
 
+  it('should uninstall a browser', async function () {
```
Lightning00Blade: Nit: it feels like this should be in a separate file as we have one for each command so to say.

Lightning00Blade (approved):

PR #10181, opened 2023-05-15
refactor: add WorkerTarget and OtherTarget

OrKoN:


packages/puppeteer-core/src/common/Browser.ts
```diff
@@ -335,7 +335,24 @@ export class CDPBrowser extends BrowserBase {
...
+    if (
+      targetInfo.type === 'sevice_worker' ||
+      targetInfo.type === 'shared_worker'
+    ) {
+      return new WorkerTarget(
+        targetInfo,
+        session,
+        context,
+        this.#targetManager,
+        (isAutoAttachEmulated: boolean) => {
```
Lightning00Blade: Nit: Can we extract this Factory to a variable its repeated 3 times

Lightning00Blade (approved):

PR #10183, opened 2023-05-15
fix: use encode/decodeURIComponent

### LINKED ISSUE #10173: [Bug]: Page.evaluate cannot be run in the code in the Chinese directory
### Bug expectation

code:
```
let result_s = await page.evaluate(() => 123);
```
should return 123
But an error is thrown: DOMException {stack: 'InvalidCharacterError: Invalid character

I found the error after debugging, because the judgment statement in the `globalThis.btoa` function throws an error

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript
// The code is in the Chinese directory

const browser = await puppeteer.launch({ headless: false });
await page.goto("https://baidu.com");
let result_s = await page.evaluate(() => 123);
```

### Error string

DOMException {stack: 'InvalidCharacterError: Invalid character

### Puppeteer configuration

```TypeScript
{ headless: false }
```

### Puppeteer version

20.2.0

### Node version

v19.1.0

### Package manager

npm

### Package manager version

9.3.0

### Operating system

Windows

jrandolf-google:
`atob` on Node doesn't encode UTF-8 string, so we use the UTF-8-compatible URIComponent algorithms.

Fixed: #10173

### DIFF
--- a/packages/puppeteer-core/src/common/util.ts
+++ b/packages/puppeteer-core/src/common/util.ts
@@ -184,7 +184,7 @@ export class PuppeteerURL {
     const [functionName = '', siteString = ''] = url.split(';');
     const puppeteerUrl = new PuppeteerURL();
     puppeteerUrl.#functionName = functionName;
-    puppeteerUrl.#siteString = globalThis.atob(siteString);
+    puppeteerUrl.#siteString = decodeURIComponent(siteString);
     return puppeteerUrl;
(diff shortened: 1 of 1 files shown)


OrKoN (changes requested):
Please reformat using eslint

OrKoN (approved):

PR #10186, opened 2023-05-15
docs: page.goto accepts an URL instance

Kikobeats:
I just tried and it worked 🙂


OrKoN:
Thanks for the PR! Docs are generated based on TypeScript types. It only works because it is probably converted toString but there is no guarantee that it will work and we currently don't support it.  Use `url.toString()` to match the expected types.

PR #10191, opened 2023-05-16
refactor: change target promises to be deferred

OrKoN:
- Adds InitializationStatus enum instead of booleans. We need to see if we can get rid of this status completely as only the page target relies on it.
- Adds a value getter to the deferred promise to check the value without waiting.

### DIFF
--- a/packages/puppeteer-core/src/common/Browser.ts
+++ b/packages/puppeteer-core/src/common/Browser.ts
@@ -39,7 +39,13 @@ import {ChromeTargetManager} from './ChromeTargetManager.js';
 import {CDPSession, Connection, ConnectionEmittedEvents} from './Connection.js';
 import {FirefoxTargetManager} from './FirefoxTargetManager.js';
 import {Viewport} from './PuppeteerViewport.js';
-import {OtherTarget, PageTarget, Target, WorkerTarget} from './Target.js';
+import {
+  InitializationStatus,
+  OtherTarget,
+  PageTarget,
+  Target,
+  WorkerTarget,
+} from './Target.js';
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/common/Browser.ts
```diff
@@ -364,8 +370,8 @@ export class CDPBrowser extends BrowserBase {
   };
 
   #onDetachedFromTarget = async (target: Target): Promise<void> => {
-    target._initializedCallback(false);
-    target._closedCallback();
+    target._initializedPromise.resolve(InitializationStatus.ABORTED);
```
Lightning00Blade, line 373: Shouldn't this happen on the Target Manager?

packages/puppeteer-core/src/common/Target.ts
```diff
@@ -208,20 +198,15 @@ export class Target {
...
   /**
    * @internal
    */
   protected _checkIfInitialized(): void {
-    if (!this._isInitialized) {
-      this._isInitialized = true;
-      this._initializedCallback(true);
+    if (!this._initializedPromise.resolved()) {
+      this._initializedPromise.resolve(InitializationStatus.SUCCESS);
       return;
```
Lightning00Blade: Nit: No need for return.

Lightning00Blade (approved):
Left some minor comments.

packages/puppeteer-core/src/common/Browser.ts
OrKoN: Probably, but it's not part of this refactoring. Need to move this later.

PR #10219, opened 2023-05-22
chore: implement locators for scroll and hover

OrKoN:
This PR implements locators for scroll and hover operations. Hover translates to `ElementHandle.hover` plus the basic locator preconditions. Scroll translates to an evaluate call setting scrollLeft/Top properties. In the future, we can enhance the scroll method with a check that the element is actually a scroll container.

### DIFF
--- a/packages/puppeteer-core/src/api/Locator.ts
+++ b/packages/puppeteer-core/src/api/Locator.ts
@@ -342,7 +342,7 @@ export class Locator extends EventEmitter {
       signal?: AbortSignal;
     }
   ): Promise<void> {
-    await this.#run(
+    return await this.#run(
       async element => {
         await element.click(clickOptions);
       },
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -357,4 +357,51 @@ export class Locator extends EventEmitter {
       }
     );
   }
+
+  async hover(hoverOptions?: {signal?: AbortSignal}): Promise<void> {
+    await this.#run(
+      async element => {
+        await element.hover();
```
jrandolf, line 364: Just return the promise?

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -357,4 +357,51 @@ export class Locator extends EventEmitter {
...
+  }
+
+  async scroll(scrollOptions?: {
+    scrollTop?: number;
+    scrollLeft?: number;
+    signal?: AbortSignal;
+  }): Promise<void> {
+    await this.#run(
+      async element => {
+        await element.evaluate(
```
jrandolf, line 384: Same here.

jrandolf-google (approved):

packages/puppeteer-core/src/api/Locator.ts
OrKoN: Updated

PR #10234, opened 2023-05-23
chore: fix uncaught error

OrKoN:


packages/puppeteer-core/src/common/bidi/HTTPResponse.ts
```diff
@@ -56,7 +56,7 @@ export class HTTPResponse extends BaseHTTPResponse {
     // TODO: update once BiDi has types
     this.#timings = (response as any).timings ?? null;
 
-    for (const header of response.headers) {
+    for (const header of response.headers || []) {
```
Lightning00Blade, line 60: Add a `TODO: That this is Firefox issue should be removed` as spec tells us it should always be list - https://w3c.github.io/webdriver-bidi/#get-the-response-data step 6

Lightning00Blade (approved):

PR #10245, opened 2023-05-25
refactor: make deferred promises more robust to use

OrKoN:
Changed the DeferredPromise to not inherit from Promise because awaiting such a promise leads to unexpected results in Node. Consider the following example:

```
const task = createDeferredPromise();
await someOtherTask();
try {
   await task;
} catch (error) {}
```

While it looks like the code is save it actually might cause unhandled exceptions in NodeJS depending on the timing and the nature of the intermediate tasks such as someOtherTask and whether the rejection of the task promise actually happens before or after try/catch is executed. Unhandled promise exceptions in Node cause the process termination. In the Puppeteer codebase, where many promises depend on the CDP events, it introduces hard to debug bugs because whether something is unhandled or not would completely depend on the timing of the browser events. We also don't want to override global unhandled event handler as we don't want to affect the user code.

Next: rename DeferredPromise to Deferred

### DIFF
--- a/packages/puppeteer-core/src/common/LifecycleWatcher.ts
+++ b/packages/puppeteer-core/src/common/LifecycleWatcher.ts
@@ -204,7 +204,7 @@ export class LifecycleWatcher {
 
   async navigationResponse(): Promise<HTTPResponse | null> {
     // Continue with a possibly null response.
-    await this.#navigationResponseReceived?.catch(() => {});
+    await this.#navigationResponseReceived?.valueOrThrow();
     return this.#navigationRequest ? this.#navigationRequest.response() : null;
   }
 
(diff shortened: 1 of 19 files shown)


packages/puppeteer-core/src/common/LifecycleWatcher.ts
```diff
@@ -204,7 +204,7 @@ export class LifecycleWatcher {
 
   async navigationResponse(): Promise<HTTPResponse | null> {
     // Continue with a possibly null response.
-    await this.#navigationResponseReceived?.catch(() => {});
+    await this.#navigationResponseReceived;
```
Lightning00Blade: I think we want
```suggestion
    await this.#navigationResponseReceived.valueOrThrow();
```

Lightning00Blade (approved):

jrandolf-google (approved):

PR #10261, opened 2023-05-26
chore: add support for waitForNetworkIdle

Lightning00Blade:
Let's discuss if we want to use this implementation for navigation (`page.goto`, `page.reload` etc.) `waitUntil: network[]`

### DIFF
--- a/.github/workflows/changed-packages.yml
+++ b/.github/workflows/changed-packages.yml
@@ -24,7 +24,7 @@ jobs:
         with:
           fetch-depth: 2
       - name: Check if branch is out of date
-        if: ${{ inputs.check-mergeable-state }}
+        if: ${{ inputs.check-mergeable-state &&  github.base_ref == 'main' }}
         run: |
           git fetch origin main --depth 1 &&
           git merge-base --is-ancestor origin/main @;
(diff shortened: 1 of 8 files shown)


.github/workflows/changed-packages.yml
```diff
@@ -24,7 +24,7 @@ jobs:
         with:
           fetch-depth: 2
       - name: Check if branch is out of date
-        if: ${{ inputs.check-mergeable-state }}
+        if: ${{ inputs.check-mergeable-state &&  github.base_ref == 'main' }}
```
Lightning00Blade, line 27: Makes test run if the target branch is different then main.

packages/puppeteer-core/src/common/NetworkManager.ts
```diff
@@ -181,8 +181,8 @@ export class NetworkManager extends EventEmitter {
     return Object.assign({}, this.#extraHTTPHeaders);
   }
 
-  numRequestsInProgress(): number {
-    return this.#networkEventManager.numRequestsInProgress();
+  inFlightRequestsNumber(): number {
```
OrKoN: nit: let's rename to inFlightRequestsCount?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1615,6 +1627,72 @@ export class Page extends EventEmitter {
     throw new Error('Not implemented');
   }
 
+  /**
+   * @internal
+   */
+  async _waitForNetworkIdle(
```
OrKoN: can we make it protected?

OrKoN (approved):

PR #10271, opened 2023-05-30
fix: pass on the auth from the download URL

OrKoN:
Updated node types but the method should be available from Node 15.

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
     "@types/diff": "5.0.2",
     "@types/mime": "3.0.1",
     "@types/mocha": "10.0.1",
-    "@types/node": "18.14.6",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -129,7 +129,7 @@
     "@types/diff": "5.0.2",
     "@types/mime": "3.0.1",
     "@types/mocha": "10.0.1",
-    "@types/node": "18.14.6",
+    "@types/node": "20.2.5",
```
Lightning00Blade, line 132: Why is this needed? The export the change needs was added in Node 15.

OrKoN: Not in the types package.

Lightning00Blade (approved):

PR #10274, opened 2023-05-30
chore: add Page.reload for BiDi

Lightning00Blade:


test/TestExpectations.json
```diff
@@ -1367,6 +1361,12 @@
     "parameters": ["firefox", "webDriverBiDi"],
     "expectations": ["FAIL"]
   },
+  {
+    "testIdPattern": "[navigation.spec] navigation Page.goto should return last response in redirect chain",
```
Lightning00Blade, line 1377: That is due to a fix in Mapper that now handles correctly `ExtraInfo` events for Redirects, but we still are missing the full support there so only the first request is reported, not the last one as expected in test.

package.json
```diff
@@ -30,6 +30,7 @@
     "test:chrome:new-headless": "wireit",
     "test:chrome:headless": "wireit",
     "test:chrome:bidi": "wireit",
+    "test:chrome:bidi-local": "wireit",
```
Lightning00Blade, line 33: Added for ease of use as BiDi requires different browser version. Will look at a way to unify both command and CI.

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -313,6 +313,22 @@ export class Page extends PageBase {
...
+    options?: WaitForOptions
+  ): Promise<HTTPResponse | null> {
+    const [response] = await Promise.all([
+      this.waitForResponse(response => {
+        return (
+          response.request().isNavigationRequest() &&
+          response.url() === this.url()
+        );
+      }),
+      this.mainFrame().context().reload(options),
```
OrKoN, line 326: let's use `this.mainFrame().reload()` ?

packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
```diff
@@ -116,6 +117,26 @@ export class BrowsingContext extends EventEmitter {
     }
   }
 
+  async reload(options: WaitForOptions = {}): Promise<void> {
```
OrKoN, line 120: does it only work on frames?

Lightning00Blade: For what else are we expecting it to work. In BiDi spec it's any browsing context.

packages/puppeteer-core/src/common/bidi/Page.ts
Lightning00Blade: This mistake on my part. The spec clearly defines any BrowsingContext not only the `TopLevel one` aka `Page` as WebDriver.

Lightning00Blade: What I said is true, but I think we want to keep the Puppeteer implementation the same for as per the Docs.

OrKoN (approved):

PR #10293, opened 2023-06-01
fix: content() not showing comments outside html tag

### LINKED ISSUE #10291: [Bug]: Page.content() ignores some HTML comments
### Bug expectation

The `Page.content()` function should return the full HTML content. ([Docs](https://pptr.dev/api/puppeteer.page.content/))
In my case, all comments outside if the `html` tag disappear.

Example page:
```html
<!DOCTYPE html>
<!-- this comment before the html tag -->
<html lang="en">
<!-- this comment is in the html -->
<head>
   <!-- this comment is in the head -->
  <meta charset="UTF-8">
  <title>TEST</title>
</head>
<body>
    <h1>TEST</h1>
    <!-- this comment is in the body -->
</body>
</html>
<!-- I am a comment after the html tag -->
```

In the response of `page.content()` the comments `<!-- I am a comment after the html tag -->` and `<!-- this comment before the html tag -->` are missing.

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript
import puppeteer from 'puppeteer'; // TS/ESM are all supported.

(async () => {
  await page.goto('https://constantin-graf-temp.s3.fr-par.scw.cloud/test.html', {
    waitUntil: 'load'
  });
  let content = await page.content();
  console.log(content);
  await browser.close();
})();
```

### Error string

no error

### Puppeteer configuration

_No response_

### Puppeteer version

20.5.0

### Node version

20.2.0

### Package manager

npm

### Package manager version

9.6.6

### Operating system

Linux

Lightning00Blade:
Sadly we can't just use `new XMLSerializer().serializeToString(document)` as that creates an attribute on the `html` tag -> `xmlns="http://www.w3.org/1999/xhtml"`

Closes #10291

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -200,6 +200,12 @@ module.exports = {
             selector:
               'MemberExpression[object.name="Promise"][property.name="race"]',
           },
+          {
+            message:
+              'Deferred `valueOrThrow` should not be called in `Deferred.race()` pass deferred directly',
+            selector:
+              'CallExpression[callee.object.name="Deferred"][callee.property.name="race"] MemberExpression[property.name="valueOrThrow"]',
+          },
         ],
         '@typescript-eslint/no-floating-promises': [
           'error',
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -204,7 +204,7 @@ module.exports = {
             message:
               'Deferred `valueOrThrow` should not be called in `Deferred.race()` pass deferred directly',
             selector:
-              'CallExpression[callee.object.name="Deferred"][callee.property.name="race"] MemberExpression[property.name="valueOrThrow"]',
+              'CallExpression[callee.object.name="Deferred"][callee.property.name="race"] > ArrayExpression > CallExpression[callee.property.name="valueOrThrow"]',
           },
         ],
         '@typescript-eslint/no-floating-promises': [
--- a/packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
+++ b/packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
@@ -11,6 +11,7 @@ import {TimeoutSettings} from '../TimeoutSettings.js';
 import {EvaluateFunc, HandleFor} from '../types.js';
 import {
   PuppeteerURL,
+  getPageContent,
   getSourcePuppeteerURLIfAvailable,
   isString,
   setPageContent,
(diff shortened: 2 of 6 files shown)


packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
```diff
@@ -240,34 +240,39 @@ export class BrowsingContext extends EventEmitter {
...
     ]);
   }
 
   async content(): Promise<string> {
     return await this.evaluate(() => {
-      let retVal = '';
-      if (document.doctype) {
-        retVal = new XMLSerializer().serializeToString(document.doctype);
+      let content = '';
+      for (const node of document.childNodes) {
```
OrKoN: can we deduplicate this or not yet?

OrKoN (approved):

packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
Lightning00Blade: We can extract it as it the same way we did for `setContent`.

OrKoN: let's do that!

packages/puppeteer-core/src/common/util.ts
```diff
@@ -510,7 +510,7 @@ export async function waitWithTimeout<T>(
     timeout,
   });
 
-  return await Deferred.race([promise, deferred.valueOrThrow()]).finally(() => {
+  return await Deferred.race([promise, deferred]).finally(() => {
```
Lightning00Blade, line 513: This was causing timeouts in test for BiDi, and they happened on a hook, stopped like 4 test from running.
Created an EsLint rule to prevent this for the feature.

.eslintrc.js
```diff
@@ -200,6 +200,12 @@ module.exports = {
             selector:
               'MemberExpression[object.name="Promise"][property.name="race"]',
           },
+          {
+            message:
+              'Deferred `valueOrThrow` should not be called in `Deferred.race()` pass deferred directly',
+            selector:
+              'CallExpression[callee.object.name="Deferred"][callee.property.name="race"] MemberExpression[property.name="valueOrThrow"]',
```
Lightning00Blade:
```suggestion
              'CallExpression[callee.object.name="Deferred"][callee.property.name="race"] > ArrayExpression > CallExpression[callee.property.name="valueOrThrow"]',
```

OrKoN (approved):

PR #10297, opened 2023-06-01
feat: add `page.removeExposedFunction`

### LINKED ISSUE #10223: [Feature]: override or remove  a exposed function
### Feature description

 ```
 await page.exposeFunction('onCustomEvent', (data) => {
        console.log(data);
      });
```
Is there any way to remove or override the exposedfunction 'onCustomEvent'?
PS:in old puppeteer version, this way as bellow works fine.
```
 this.page._pageBindings.set('onCustomEvent', (data) => {
});
```
But in new puppeteer version,_pageBindings is not found.

### PR DESCRIPTION
**Pull request by Lightning00Blade**
Closes #10223

### CODE REVIEW

File: packages/puppeteer-core/src/common/Binding.ts [hunk 1]
```diff
@@ -8,17 +8,27 @@ import {debugError} from './util.js';
  * @internal
  */
 export class Binding {
+  #identifier: string;
```
OrKoN: I am not sure if we should add it here as it is really the identifier for the evaluated script, right?

Lightning00Blade: That is correct. An other option is having a Map with `exposedFunction.name` to `identifier` on the Page, so similar.

OrKoN: what do you think if we change the exposeFunction to return some sort of object identifying the function?

packages/puppeteer-core/src/common/Page.ts
```diff
@@ -705,17 +706,49 @@ export class CDPPage extends Page {
...
+          .evaluate(name => {
+            // Removes the dangling Puppeteer binding wrapper.
+            // @ts-expect-error: In a different context.
+            globalThis[name] = undefined;
+          }, name)
+          .catch(debugError);
+      })
+    );
+
+    this.#exposedFunctions.delete(name);
```
OrKoN, line 749: missing `this.#bindings.set`

OrKoN (approved):

PR #10305, opened 2023-06-02
feat: implement Locators

OrKoN:


packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -345,6 +347,20 @@ export class Frame {
...
 
+  /**
+   * Creates a locator for the provided `selector`. See {@link Locator} for
+   * details and supported actions.
+   *
+   * @remarks
+   * Locators API is experimental and we will not follow semver for breaking
+   * change in the Locators API.
+   *
+   * @public
```
Lightning00Blade: Nit: Not needed the class is defined as public.

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -819,7 +819,14 @@ export class Page extends EventEmitter {
...
   /**
-   * @internal
+   * Creates a locator for the provided `selector`. See {@link Locator} for
+   * details and supported actions.
+   *
+   * @remarks
+   * Locators API is experimental and we will not follow semver for breaking
+   * change in the Locators API.
+   *
+   * @public
```
Lightning00Blade: Nit: Not needed the class is defined as public.

Lightning00Blade (approved):

PR #10318, opened 2023-06-05
chore: $ and $$ BiDi support

### LINKED ISSUE #9100: [Feat] WebDriver BiDi "Hacker News" scenario should work via BiDi
WebDriver BiDi ["Hacker News" scenario](https://github.com/GoogleChromeLabs/chromium-bidi/milestone/2) should work in Puppeteer via BiDi.
- [x] #9101 
- [x] #9103
- [x] #9104
- [x] #9102

### LINKED ISSUE #9102: [Feat] BiDi/Page implements `$`
As part of the #9100, BiDi/Page should implement [`$`](https://github.com/puppeteer/puppeteer/blob/8acafe2fd91b70b8bbbe10467b6a0b5894b964bb/packages/puppeteer-core/src/api/Page.ts#L769).

Lightning00Blade:
Implementation is ready for review but some test may pass now what are marked as `SKIP`.
Let me do a thorough overview before merging.

Closes #9102
Closes #9100

### DIFF
--- a/packages/puppeteer-core/src/common/LazyArg.ts
+++ b/packages/puppeteer-core/src/common/LazyArg.ts
@@ -14,26 +14,34 @@
  * limitations under the License.
  */
 
-import {ExecutionContext} from './ExecutionContext.js';
+import {JSHandle} from '../api/JSHandle.js';
+import PuppeteerUtil from '../injected/injected.js';
 
 /**
  * @internal
  */
-export class LazyArg<T> {
+export interface PuppeteerUtilWrapper {
+  puppeteerUtil: Promise<JSHandle<PuppeteerUtil>>;
+}
+
+/**
+ * @internal
+ */
+export class LazyArg<T, Context = PuppeteerUtilWrapper> {
   static create = <T>(
-    get: (context: ExecutionContext) => Promise<T> | T
+    get: (context: PuppeteerUtilWrapper) => Promise<T> | T
   ): T => {
     // We don't want to introduce LazyArg to the type system, otherwise we would
     // have to make it public.
     return new LazyArg(get) as unknown as T;
   };
 
-  #get: (context: ExecutionContext) => Promise<T> | T;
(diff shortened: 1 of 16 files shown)


test/assets/p-selectors.html
```diff
@@ -1,13 +1,14 @@
...
-                <template shadowrootmode="open">
-                    <a id="e">deep text</a>
-                </template>
-            </div>
-        </template>
-    </div>
-</div>
\ No newline at end of file
+    <div id="c"></div>
+</div>
```
Lightning00Blade, line 4: Firefox does not support the `<template shadowrootmode="open">` which is not a standart yet. 
Using the old way to insure we test it for it that makes their CDP implementation pass as well,

packages/puppeteer-core/src/common/LazyArg.ts
```diff
@@ -14,26 +14,34 @@
...
 
-import {ExecutionContext} from './ExecutionContext.js';
+import {JSHandle} from '../api/JSHandle.js';
+import PuppeteerUtil from '../injected/injected.js';
 
 /**
  * @internal
  */
-export class LazyArg<T> {
+export interface HasPuppeteerUtil {
```
OrKoN: Let's call it something like PuppeteerUtilWrapper?

packages/puppeteer-core/src/common/QueryHandler.ts
```diff
@@ -99,6 +99,19 @@ export class QueryHandler {
     ));
   }
 
+  static elementHasWorld(element: ElementHandle<Node>): asserts element {
```
OrKoN: why don't we add it as method on ElementHandle?

OrKoN: if it is an assertion, let's start with assertElementHasWorld ?

OrKoN (approved):
LGTM % nits

PR #10320, opened 2023-06-06
docs: add a few examples for ARIA queries

OrKoN:


test/src/queryhandler.spec.ts
```diff
@@ -446,6 +446,28 @@ describe('Query handler tests', function () {
       ).toBeTruthy();
     });
 
+    it('should work ARIA selectors with  role', async () => {
```
Lightning00Blade:
```suggestion
    it('should work ARIA selectors with role', async () => {
```

Lightning00Blade (approved):

PR #10340, opened 2023-06-07
feat: add `reset` to mouse

jrandolf-google:
This PR gives users the ability to reset their mouse without keeping track of their button presses and position.

### DIFF
--- a/packages/puppeteer-core/src/common/Input.ts
+++ b/packages/puppeteer-core/src/common/Input.ts
@@ -599,6 +599,29 @@ export class Mouse {
   }
 
   /**
+   * Resets the mouse to the default state: No buttons pressed; position at
+   * (0,0).
+   */
+  async reset(): Promise<void> {
+    const actions = [];
+    for (const [flag, button] of [
+      [MouseButtonFlag.Left, MouseButton.Left],
+      [MouseButtonFlag.Middle, MouseButton.Middle],
+      [MouseButtonFlag.Right, MouseButton.Right],
+      [MouseButtonFlag.Forward, MouseButton.Forward],
+      [MouseButtonFlag.Back, MouseButton.Back],
+    ] as const) {
+      if (this.#state.buttons & flag) {
+        actions.push(this.up({button: button}));
+      }
+    }
+    if (this.#state.position.x !== 0 || this.#state.position.y !== 0) {
+      actions.push(this.move(0, 0));
+    }
+    await Promise.all(actions);
+  }
+
+  /**
    * Moves the mouse to the given coordinate.
    *
    * @param x - Horizontal position of the mouse.
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -1461,7 +1461,7 @@
     "testIdPattern": "[mouse.spec] Mouse should reset properly",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["cdp", "firefox"],
-    "expectations": ["FAIL"]
+    "expectations": ["SKIP"]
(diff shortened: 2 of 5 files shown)


packages/puppeteer-core/src/common/Input.ts
```diff
@@ -598,6 +598,33 @@ export class Mouse {
...
+    if (this.#state.buttons & MouseButtonFlag.Middle) {
+      actions.push(this.up({button: MouseButton.Middle}));
+    }
+    if (this.#state.buttons & MouseButtonFlag.Right) {
+      actions.push(this.up({button: MouseButton.Right}));
+    }
+    if (this.#state.buttons & MouseButtonFlag.Forward) {
+      actions.push(this.up({button: MouseButton.Forward}));
+    }
+    if (this.#state.buttons & MouseButtonFlag.Back) {
```
OrKoN: want to convert it to an array + iteration? should be easier to read and less chance for copy paste errors.

OrKoN (changes requested):
Let's add a test for this method.

packages/puppeteer-core/src/common/Input.ts
```diff
@@ -598,6 +598,33 @@ export class Mouse {
     }
   }
 
+  /**
+   * Resets the mouse to the default state. No buttons pressed; position at
```
Lightning00Blade: I think we should make it clear from the comment that we call `up` and `move`.
Something like finishes the started actions or similar.

jrandolf: Those are implementation details. I'm not sure if we want to expose them.

packages/puppeteer-core/src/common/Input.ts
jrandolf: MouseButtonFlag is a const enum, so this isn't possible.

jrandolf: I can alternatively create an array manually.

OrKoN: yeah, I meant the latter option.

OrKoN: i.e.,

```
const flags = [MouseButtonFlag.Left, ...];
for (const button of flags) {
   if (this.#state.buttons & button) {
      actions.push(this.up({button}));
    }
}
```

packages/puppeteer-core/src/common/Input.ts
Lightning00Blade: I think we should as they convey that the appropriate events will be send on the page side.

OrKoN: I agree with Nik. Because we have also an option to reset without sending the page commands, so we need to clarify that the events will be dispatched. Perhaps: `Moves the mouse to 0,0 position and un-presses all pressed buttons` or smth like that.

OrKoN:
Note that CI fails.

OrKoN:
@jrandolf tests are failing. PTAL

OrKoN (approved):

test/TestExpectations.json
```diff
@@ -1457,6 +1457,12 @@
     "parameters": ["chrome", "webDriverBiDi"],
     "expectations": ["PASS"]
   },
+  {
+    "testIdPattern": "[mouse.spec] Mouse should reset properly",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["cdp", "firefox"],
+    "expectations": ["FAIL"]
```
```suggestion
    "expectations": ["SKIP"]
```

PR #10344, opened 2023-06-07
chore: use `launch` helper

Lightning00Blade:
This will enable us to run all test without hanging and memory clog.

### DIFF
--- a/test/src/TargetManager.spec.ts
+++ b/test/src/TargetManager.spec.ts
@@ -15,46 +15,39 @@
  */
 
 import expect from 'expect';
-import {
-  CDPBrowser,
-  CDPBrowserContext,
-} from 'puppeteer-core/internal/common/Browser.js';
+import {CDPBrowser} from 'puppeteer-core/internal/common/Browser.js';
 
-import {getTestState} from './mocha-utils'; // eslint-disable-line import/extensions
+import {getTestState, launch} from './mocha-utils.js';
 import {attachFrame} from './utils.js';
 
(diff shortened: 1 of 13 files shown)


test/src/headful.spec.ts
```diff
@@ -96,29 +93,24 @@ describe('headful tests', function () {
...
+    await Promise.all(
+      browsers.map((close, index) => {
+        delete browsers[index];
+        return close();
+      })
+    );
   });
 
-  describe('HEADFUL', function () {
+  describe.skip('HEADFUL', function () {
```
OrKoN: forgotten skip?

test/src/TargetManager.spec.ts
```diff
@@ -15,46 +15,37 @@
...
         ]),
-      })
-    )) as CDPBrowser;
-  });
-
-  beforeEach(async () => {
-    context = await browser.createIncognitoBrowserContext();
+      }),
+      {createPage: false}
+    )) as any;
```
OrKoN: can we type it with smth other than any?

test/src/proxy.spec.ts
```diff
@@ -52,7 +51,7 @@ function getEmptyPageUrl(server: TestServer): string {
 }
 
 describe('request proxy', () => {
-  let browser: Browser;
+  // let browser: Browser;
```
OrKoN: remove?

OrKoN (approved):
% comments

PR #10370, opened 2023-06-13
chore: tracing over bidi+

OrKoN:


packages/puppeteer-core/src/common/Page.ts
```diff
@@ -184,7 +184,26 @@ export class CDPPage extends Page {
...
+      close: opts => {
+        return this.#client.send('IO.close', opts);
+      },
+      start: opts => {
+        return client.send('Tracing.start', opts);
+      },
+      stop: async () => {
+        const deferred = Deferred.create();
+        this.#client.once('Tracing.tracingComplete', event => {
+          console.log('resolved');
```
Lightning00Blade:
```suggestion

```

packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
```diff
@@ -289,6 +294,7 @@ export class BrowsingContext extends EventEmitter {
       context: this.#id,
     });
     const sessionId = session.result.cdpSession;
+    this.#cdpSessionId = sessionId;
```
Lightning00Blade: Should we skip the first call if we have a `cdpSessionId` set?

OrKoN: I am not sure about the lifetime of sessions on the mapper side? perhaps we could

packages/puppeteer-core/src/common/Tracing.ts
```diff
@@ -29,6 +32,11 @@ export interface TracingOptions {
   categories?: string[];
 }
 
+interface TracingSource extends ProtocolReadable {
```
Lightning00Blade: Please for the new interfaces provide `@internal` comment.

Lightning00Blade (approved):

PR #10371, opened 2023-06-13
test: misc test improvements

OrKoN:


test/src/ariaqueryhandler.spec.ts
```diff
@@ -553,9 +565,14 @@ describe('AriaQueryHandler', () => {
...
-      const waitForSelector = page.waitForSelector('aria/zombo').then(() => {
-        return (divFound = true);
-      });
+      const waitForSelector = page
+        .waitForSelector('aria/zombo')
+        .then(() => {
+          return (divFound = true);
+        })
+        .catch(err => {
+          return err;
```
Lightning00Blade: On theory error can have a value `true` so return `false` is the correct thing to do here.

Lightning00Blade: Or add the same expect as the previous test `expect(divHidden).toBe(true);`

test/src/elementhandle.spec.ts
```diff
@@ -389,23 +389,29 @@ describe('ElementHandle specs', function () {
...
-        ElementHandle<HTMLDivElement>
-      >;
+      const innerWaitFor = element.waitForSelector('.bar').catch(err => {
+        return err;
+      }) as Promise<ElementHandle<HTMLDivElement>>;
       await element.evaluate(el => {
         el.innerHTML = '<div class="bar">bar1</div>';
       });
       element = (await innerWaitFor)!;
+      if (element instanceof Error) {
```
Lightning00Blade, line 412: How about we use `expect(element).not.toBeInstanceOf(Error)` or `expect(element).toBeInstanceOf(ElementHandle)`

OrKoN: the stacktrace and the error message is not too detailed in that case imo:

```
  1) ElementHandle specs
       Element.waitForSelector
         should wait correctly with waitForSelector on an element:
     Error: expect(received).not.toBeInstanceOf(expected)

Expected constructor: not Error

      at Context.<anonymous> (/Users/alexrudenko/src/puppeteer/test/src/elementhandle.spec.ts:400:27)

```

Lightning00Blade (approved):

PR #10378, opened 2023-06-13
ci: add default timeouts to tests

Lightning00Blade:
TestExpectaions may be wrong now that we we may Fail rather expect Mocha Timeout.
So we may need to handle that in the test runner.


OrKoN (approved):

Lightning00Blade:
We would need a solution for when we set the timeout inside the test themself, 
Maybe we can extend the original `this.timeout` to do this as well.

PR #10383, opened 2023-06-14
chore: add waitForSelector for BiDi

Lightning00Blade:
- [x] TestExpectation need to be updated
- [x] Chrome-bidi need to be updated


packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -307,6 +349,20 @@ export class Frame {
     throw new Error('Not implemented');
   }
 
+  /**
+   * @internal
+   */
+  main(): Evaluator {
```
OrKoN: Let's use of the term Realm? `mainRealm` and `isolatedRealm`?

OrKoN: https://tc39.es/ecma262/#sec-code-realms

OrKoN (approved):

PR #10387, opened 2023-06-15
chore: enable coverage over bidi+

OrKoN:
I will apply the same approach to tracing/a11y which were previously implemented in a different way.

### DIFF
--- a/packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
+++ b/packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
@@ -4,7 +4,9 @@ import ProtocolMapping from 'devtools-protocol/types/protocol-mapping.js';
 import {WaitForOptions} from '../../api/Page.js';
 import PuppeteerUtil from '../../injected/injected.js';
 import {assert} from '../../util/assert.js';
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
```diff
@@ -51,6 +53,54 @@ const lifeCycleToReadinessState = new Map<
   ['domcontentloaded', 'interactive'],
 ]);
 
+/**
+ * @internal
+ */
+export class CDPSessionWrapper extends EventEmitter implements CDPSession {
```
Lightning00Blade, line 59: Maybe this class is should live in the Connection file?

Lightning00Blade (approved):

packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
OrKoN: Not sure, my thinking was is that it wraps the context and is only accessible from the context so we could keep it here?

Lightning00Blade: Let's keep it as is.

PR #10401, opened 2023-06-16
refactor: use input API for CDP inputs

jrandolf-google:


packages/puppeteer-core/src/common/Input.ts
```diff
@@ -467,84 +275,17 @@ interface MouseState {
...
- * @public
+ * @internal
  */
-export class Mouse {
+export class CDPMouse extends Mouse {
   #client: CDPSession;
-  #keyboard: Keyboard;
+  #keyboard: CDPKeyboard;
 
   /**
```
Lightning00Blade, line 284: Not needed.

Lightning00Blade (approved):
LGTM, only thing is why use a lot of Readonly type?

packages/puppeteer-core/src/common/Input.ts
jrandolf: Within the implementation, we don't want to treat the type as the public type. In fact, it's preferred to use the internal type wherever you can. Reason being when the consumer accesses this class through `page.mouse` for example, they will see the signatures from the `api/Input.js` file rather than the ones declared here.

It's useful if someone knows they have a specific implementation and wants to just `as` cast the implementation rather than for each object of the implementation.

For example, compare

```ts
(page as CDPPage).mouse as CDPMouse;
```

versus

```ts
(page as CDPPage).mouse; // already a CDPMouse
```

jrandolf:
> LGTM, only thing is why use a lot of Readonly type?

We should be doing this so users know we don't modify their object.

OrKoN (approved):

PR #10402, opened 2023-06-16
chore: implement Webdriver BiDi mouse and touchscreen

jrandolf-google:


OrKoN:
Changed to `chore: ` since we are not releasing it yet and don't want to bump the version.

Lightning00Blade:
@jrandolf Firefox are failing, please investigate and/or disable test if they are not expected to pass.

OrKoN:
@Lightning00Blade PTAL

packages/puppeteer-core/src/common/bidi/Input.ts
```diff
@@ -0,0 +1,326 @@
...
+}
+
+/**
+ * @internal
+ */
+export class Touchscreen extends BaseTouchscreen {
+  #context: BrowsingContext;
+
+  /**
+   * @internal
```
Lightning00Blade, line 238: Not needed Class is internal.

packages/puppeteer-core/src/common/bidi/Input.ts
```diff
@@ -0,0 +1,326 @@
...
+    });
+  }
+
+  override async touchEnd(): Promise<void> {
+    await this.#context.connection.send('input.performActions', {
+      context: this.#context.id,
+      actions: [
+        {
+          type: Bidi.Input.SourceActionsType.Pointer,
+          id: 'main',
```
Lightning00Blade: If I understand correctly BiDi allows for multiple input form the the same group (You can have 2 keyboards, both sending separate comands).
We should extract this `main` somewhere as on Puppeteer side we want to keep using only one for now.

packages/puppeteer-core/src/common/bidi/Frame.ts
```diff
@@ -54,13 +54,14 @@ export class Frame extends BaseFrame {
...
       [MAIN_SANDBOX]: new Sandbox(context, timeoutSettings),
-      [PUPPETEER_SANDBOX]: new Sandbox(
-        context.createSandboxRealm(UTILITY_WORLD_NAME),
-        timeoutSettings
-      ),
+      [PUPPETEER_SANDBOX]: new Sandbox(puppeteerRealm, timeoutSettings),
     };
+
+    puppeteerRealm.setFrame(this);
+    context.setFrame(this);
```
Lightning00Blade, line 64: Why is this change need?

jrandolf: Sometimes inputs are performed in the puppeteer realm without transferring to the main sandbox. This causes issues because we need a reference to the page which is obtained through the frame on the context if it exists.

packages/puppeteer-core/src/common/bidi/Input.ts
jrandolf: I think in the end we will export this through the `api.ts` file. We should export everything no matter if it's internal. We want the API generator to deal with internal objects. Otherwise maintenance gets too difficult since it relies on predicting where internal classes touch.

Lightning00Blade (approved):

PR #10436, opened 2023-06-22
ci: separate unit tests

OrKoN:
Unit tests are moved next to the code they test.
`npm run unit` is the command to run them. The test files are excluded from the publishing and an eslint rule is added to prevent regular code from depending on the test files.

`.test.js` is chosen because it is the default suffix supported by the Node test runner. Running unit tests required Node v20+.

Other tests will be moved in a separate PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -372,6 +372,27 @@ jobs:
         run: |
           docker run -i --init --cap-add=SYS_ADMIN --rm puppeteer-test-image node -e "`cat test/smoke-test.js`"
 
+  unit-tests:
+    name: '[Required] Unit tests'
+    runs-on: ubuntu-latest
+    needs: check-changes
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -372,6 +372,29 @@ jobs:
...
+  unit-tests:
+    name: '[Required] Unit tests'
+    runs-on: ubuntu-latest
+    needs: check-changes
+    if: ${{ contains(fromJSON(needs.check-changes.outputs.changes), 'puppeteer') }}
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
+        with:
+          fetch-depth: 2
```
Lightning00Blade: Please remove `fetch-depth`

Lightning00Blade (approved):

PR #10453, opened 2023-06-26
chore: use a less chatty BiDi+

Lightning00Blade:
Ref https://github.com/GoogleChromeLabs/chromium-bidi/pull/892

### DIFF
--- a/packages/puppeteer-core/src/common/bidi/Connection.ts
+++ b/packages/puppeteer-core/src/common/bidi/Connection.ts
@@ -234,17 +234,14 @@ export class Connection extends EventEmitter {
       // `log.entryAdded` specific context
     } else if ('source' in event.params && event.params.source.context) {
       context = this.#browsingContexts.get(event.params.source.context);
-    } else if (event.method === 'cdp.eventReceived') {
+    } else if (isCDPEvent(event)) {
       // TODO: this is not a good solution and we need to find a better one.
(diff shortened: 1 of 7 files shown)


packages/puppeteer-core/src/common/bidi/Connection.ts
```diff
@@ -234,17 +234,14 @@ export class Connection extends EventEmitter {
       // `log.entryAdded` specific context
     } else if ('source' in event.params && event.params.source.context) {
       context = this.#browsingContexts.get(event.params.source.context);
-    } else if (event.method === 'cdp.eventReceived') {
+    } else if (isCDPEvent(event)) {
       // TODO: this is not a good solution and we need to find a better one.
       // Perhaps we need to have a dedicated CDP event emitter or emulate
       // the CDPSession interface with BiDi?.
-      const cdpSessionId = event.params.cdpSession;
```
OrKoN, line 241: is it renamed in chromium-bidi ?

OrKoN (approved):

packages/puppeteer-core/src/common/bidi/Connection.ts
Lightning00Blade: Yeah I did the renaming. Looking at this now I should update the BiDi+ types to have some comments.

PR #10458, opened 2023-06-27
fix: `port` option to run dev and e2e side-by-side

Lightning00Blade:


packages/ng-schematics/package.json
```diff
@@ -8,7 +8,8 @@
     "dev:test": "npm run test --watch",
     "dev": "npm run build --watch",
     "test": "wireit",
-    "sandbox": "node tools/sandbox.js"
+    "sandbox": "node tools/sandbox.js",
+    "sandbox:test": "node tools/sandbox.js --test"
```
jrandolf, line 12: test:sandbox

packages/ng-schematics/src/schematics/test/index.ts
```diff
@@ -34,21 +38,21 @@ export function test(options: SchematicsSpec): Rule {
   };
 }
 
-function findTestingFramework([name, project]: [
-  string,
-  any
-]): TestingFramework {
+function findTestingOption<Property extends keyof SchematicsOptions>(
+  [name, project]: [string, any],
```
jrandolf: Could we use `unknown` and validate properly?

jrandolf-google (changes requested):

packages/ng-schematics/package.json
Lightning00Blade: That is not correct, we are not testing the sandbox we are running the `schematics:test` in the sandbox.

jrandolf-google (approved):

PR #10466, opened 2023-06-28
chore: update Bidi key values

jrandolf-google:


test/TestExpectations.json
```diff
@@ -3653,6 +3743,18 @@
     "parameters": ["cdp", "chrome", "new-headless"],
     "expectations": ["FAIL", "PASS"]
   },
+  {
+    "testIdPattern": "[function.spec] Function interpolateFunction should work",
+    "platforms": ["linux"],
```
Lightning00Blade: Is this correct, I think this more of a unit test that needs to be moved. So it should pass on all 3 platforms

Lightning00Blade (approved):
LGTM + please see comment and let CI run.

PR #10468, opened 2023-06-28
chore: clean NetworkManager Maps for BiDi

Lightning00Blade:
Clears some Memory if the user is not actively listening to the requests.
Also remove the request if the frame is disposed.


packages/puppeteer-core/src/common/bidi/NetworkManager.ts
```diff
@@ -95,10 +98,17 @@ export class NetworkManager extends EventEmitter {
...
+    this.#requestMap.delete(event.request.request);
   }
 
   getNavigationResponse(navigationId: string | null): HTTPResponse | null {
-    return this.#navigationMap.get(navigationId ?? '') ?? null;
+    if (!navigationId) {
+      return null;
+    }
+    const response = this.#navigationMap.get(navigationId);
+    this.#navigationMap.delete(navigationId);
```
OrKoN: probably should not do that

OrKoN (approved):

PR #10474, opened 2023-06-29
fix: fix escaping algo for P selectors

### LINKED ISSUE #10473: [Bug] Escapes for P selectors aren't working
## Steps

1. Create a button with `My name is Jun (pronounced like "June")` inside.
2. Try to query for it using `::-p-text(My name is Jun \\(pronounced like "June"\\))`.
3. Observe

## What happens?

The element is not found.

## What should happen?

The element should be found.

jrandolf-google:
Fixed: #10473

### DIFF
--- a/test/src/queryhandler.spec.ts
+++ b/test/src/queryhandler.spec.ts
@@ -637,5 +637,26 @@ describe('Query handler tests', function () {
       const elements = await page.$$('::-p-text(world), button');
       expect(elements).toHaveLength(1);
     });
+
+    it.only('should handle escapes', async () => {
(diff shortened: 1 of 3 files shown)


test/src/queryhandler.spec.ts
```diff
@@ -637,5 +637,26 @@ describe('Query handler tests', function () {
       const elements = await page.$$('::-p-text(world), button');
       expect(elements).toHaveLength(1);
     });
+
+    it.only('should handle escapes', async () => {
```
OrKoN: remove the `.only`

OrKoN (approved):

PR #10478, opened 2023-06-29
chore: readd browser setup hook

Lightning00Blade:
There are quite a few changes here will comment on the important ones.

### DIFF
--- a/test/src/chromiumonly.spec.ts
+++ b/test/src/chromiumonly.spec.ts
@@ -16,10 +16,10 @@
 import {IncomingMessage} from 'http';
 
 import expect from 'expect';
+import {Deferred} from 'puppeteer-core/internal/util/Deferred.js';
 
-import {getTestState, launch} from './mocha-utils.js';
+import {getTestState, setupTestBrowserHooks, launch} from './mocha-utils.js';
 import {waitEvent} from './utils.js';
-
 // TODO: rename this test suite to launch/connect test suite as it actually
 // works across browsers.
 describe('Chromium-Specific Launcher tests', function () {
@@ -124,13 +124,19 @@ describe('Chromium-Specific Launcher tests', function () {
         await close();
       }
     });
-    it('should fire "disconnected" when closing with pipe', async () => {
+    it('should fire "disconnected" when closing with pipe', async function () {
       const {browser, close} = await launch({pipe: true});
       try {
         const disconnectedEventPromise = waitEvent(browser, 'disconnected');
         // Emulate user exiting browser.
         browser.process()!.kill();
-        await disconnectedEventPromise;
+        await Deferred.race([
+          disconnectedEventPromise,
+          Deferred.create({
+            message: `Failed in after Hook`,
+            timeout: this.timeout() - 1000,
+          }),
+        ]);
       } finally {
         await close();
       }
@@ -139,6 +145,8 @@ describe('Chromium-Specific Launcher tests', function () {
 });
 
 describe('Chromium-Specific Page Tests', function () {
+  setupTestBrowserHooks();
+
   it('Page.setRequestInterception should work with intervention headers', async () => {
     const {server, page} = await getTestState();
 
--- a/test/src/launcher.spec.ts
+++ b/test/src/launcher.spec.ts
@@ -574,7 +574,9 @@ describe('Launcher specs', function () {
           ...defaultBrowserOptions,
           args: ['--no-startup-window'],
         };
-        const {browser, close} = await launch(options, {createContext: false});
+        const {browser, close} = await launch(options, {
+          createContext: false,
+        });
         try {
           const pages = await browser.pages();
           expect(pages).toHaveLength(0);
--- a/test/src/mocha-utils.ts
+++ b/test/src/mocha-utils.ts
@@ -34,6 +34,7 @@ import {
   PuppeteerNode,
 } from 'puppeteer-core/internal/node/PuppeteerNode.js';
 import {rmSync} from 'puppeteer-core/internal/node/util/fs.js';
+import {Deferred} from 'puppeteer-core/internal/util/Deferred.js';
 import {isErrorLike} from 'puppeteer-core/internal/util/ErrorLike.js';
 import sinon from 'sinon';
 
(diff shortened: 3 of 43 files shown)


test/src/chromiumonly.spec.ts
```diff
@@ -124,13 +124,19 @@ describe('Chromium-Specific Launcher tests', function () {
...
     });
-    it('should fire "disconnected" when closing with pipe', async () => {
+    it('should fire "disconnected" when closing with pipe', async function () {
       const {browser, close} = await launch({pipe: true});
       try {
         const disconnectedEventPromise = waitEvent(browser, 'disconnected');
         // Emulate user exiting browser.
         browser.process()!.kill();
-        await disconnectedEventPromise;
+        await Deferred.race([
```
Lightning00Blade, line 133: Use Puppeteer Deferred to Clear Timeouts automagically

OrKoN (approved):

test/src/headful.spec.ts
```diff
@@ -35,7 +35,7 @@ const serviceWorkerExtensionPath = path.join(
   'extension'
 );
 
-describe('headful tests', function () {
+describe.skip('headful tests', function () {
```
Lightning00Blade:
```suggestion
describe('headful tests', function () {
```

test/src/launcher.spec.ts
```diff
@@ -562,7 +562,7 @@ describe('Launcher specs', function () {
         });
         expect(error.message).toContain('either pipe or debugging port');
       });
-      it('should launch Chrome properly with --no-startup-window and waitForInitialPage=false', async () => {
+      it.skip('should launch Chrome properly with --no-startup-window and waitForInitialPage=false', async () => {
```
Lightning00Blade:
```suggestion
      it('should launch Chrome properly with --no-startup-window and waitForInitialPage=false', async () => {
```

test/src/mocha-utils.ts
```diff
@@ -143,6 +145,23 @@ const setupServer = async () => {
   return {server, httpsServer};
 };
 
+export const setupTestBrowserHooks = (): void => {
+  before(async function () {
+    this.timeout(0);
+    try {
+      if (!state.browser) {
+        state.browser = await puppeteer.launch({
```
Lightning00Blade, line 143: Create the browser once, and reuse it as you please.

test/src/mocha-utils.ts
```diff
@@ -389,30 +431,34 @@ export const createTimeout = <T>(
   });
 };
 
-let browserCleanups: Array<() => Promise<void>> = [];
+const browserCleanupsAfterAll: Array<() => Promise<void>> = [];
```
Lightning00Blade, line 424: Some test need to run the Cleanup after all the test so they don't spawn browser for each test;

test/src/stacktrace.spec.ts
```diff
@@ -18,12 +18,19 @@ import assert from 'assert';
 
 import expect from 'expect';
 
-import {getTestState} from './mocha-utils.js';
+import {getTestState, setupTestBrowserHooks} from './mocha-utils.js';
 import {waitEvent} from './utils.js';
 
 const FILENAME = __filename.replace(/[/\-\\^$*+?.()|[\]{}]/g, '\\$&');
+const parseStackTrace = (stack: string): string => {
```
Lightning00Blade, line 25: Makes this test less reliant on the file as adding a new line somewhere will cause it to brake;
@jrandolf PTAL

tools/mochaRunner/src/utils.ts
```diff
@@ -171,6 +165,15 @@ export function getExpectationUpdates(
     // If an error occurs during a hook
     // the error not have a file associated with it
     if (!failure.file) {
+      addEntry({
```
Lightning00Blade, line 168: Add a Hook entry the hooks should have logged the correct file that it happen.

test/src/mocha-utils.ts
```diff
@@ -143,6 +144,22 @@ const setupServer = async () => {
   return {server, httpsServer};
 };
 
+export const setupTestBrowserHooks = (): void => {
+  before(async function () {
+    this.timeout(0);
```
OrKoN: what does timeout(0) do?

test/src/mocha-utils.ts
OrKoN: hm, I am not following, could you be more specific about which tests need it?

tools/mochaRunner/src/utils.ts
OrKoN: Was not able to parse the sentence :D

Lightning00Blade: Neither can I :no_mouth: . We add and entry rather then skipping as we want to investigate the failer. If the error comes from our code we should see the file name in the error message

test/src/mocha-utils.ts
Lightning00Blade: `ignoreHTTPSErrors` and `OOPIF` Test set-up their own browsers once.
If we don't split them we get error that the browser has not been closed after each test.

test/src/mocha-utils.ts
Lightning00Blade: Stops the timeout. As discussed offline we will change to use the puppeteer.launch timeout option set to lower that the deafult.

test/src/mocha-utils.ts
```diff
@@ -389,30 +429,34 @@ export const createTimeout = <T>(
...
+const closeLaunched = (storage: Array<() => Promise<void>>) => {
+  return async () => {
+    let cleanup = storage.pop();
+    try {
+      while (cleanup) {
+        await cleanup();
+        cleanup = storage.pop();
+      }
+    } catch (error) {
+      // If the browser was closed by other mean swallow the error
```
OrKoN:
```suggestion
      // If the browser was closed by other means, swallow the error
```

test/src/mocha-utils.ts
```diff
@@ -389,30 +429,34 @@ export const createTimeout = <T>(
...
+      // and mark he browser as closed
```
OrKoN:
```suggestion
      // and mark the browser as closed.
```

test/src/mocha-utils.ts
```diff
@@ -389,30 +429,34 @@ export const createTimeout = <T>(
...
+      if ((error as any)?.message.includes('Connection closed')) {
```
OrKoN:
```suggestion
      if ((error as Error)?.message.includes('Connection closed')) {
```

OrKoN (approved):

PR #10480, opened 2023-06-29
fix: negative timeout doesn't break launch

Lightning00Blade:


jrandolf-google (approved):

jrandolf-google (changes requested):
It might be good to throw here if it is negative too.

jrandolf-google (approved):

OrKoN (approved):

PR #10512, opened 2023-07-05
fix: add an internal page.locatorRace

OrKoN:
Using Locator.race is not possible to use via a type-only import whereas page.locatorRace could be used with a type import.
That is useful for libraries like @puppeteer/replay that don't have a static dependency on puppeteer.


OrKoN:
@jrandolf marked as internal as discussed offline

jrandolf-google (approved):

PR #10515, opened 2023-07-06
feat: add page.createCDPSession method

OrKoN:
There is no particular reason why a CDP session should be created only through the targets. Since we are not sure if the target API makes sense for a WebDriver BiDi client, I think it makes sense to expose a shortcut via the page (in the future, frame and worker instances too).


packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
```diff
@@ -74,8 +93,12 @@ export class CDPSessionWrapper extends EventEmitter implements CDPSession {
     return result.result;
   }
 
-  detach(): Promise<void> {
-    throw new Error('Method not implemented.');
+  async detach(): Promise<void> {
+    cdpSessions.set(this.id(), this);
```
Lightning00Blade: This should be delete? Else it would keep the reference forever and it can't be garbage collected.

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -622,6 +623,13 @@ export class Page extends EventEmitter {
     throw new Error('Not implemented');
   }
 
+  /**
+   * Creates a Chrome Devtools Protocol session attached to the page.
```
Lightning00Blade, line 627: Nit: Link to CDP website?

packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
OrKoN:
```suggestion
    cdpSessions.delete(this.id());
```

Lightning00Blade (approved):

PR #10537, opened 2023-07-10
refactor: rename _client → getCDPSession

Kikobeats:

Refactor

Already tested

**Summary**

Sometimes is useful to get the current `CDPSession` associated with a page, e.g., 

https://github.com/microlinkhq/browserless/blob/5b81d2e12abab0726176894365f4fb4b4efd6580/packages/screencast/src/utils.js#L8

I really don't see why this method is supposed to be hide. Why not remove underscore from there? It's still internal method so, but you can intentionally use it and that's okay.

Nope

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/api/Frame.ts
+++ b/packages/puppeteer-core/src/api/Frame.ts
@@ -349,7 +349,7 @@ export class Frame {
   /**
    * @internal
    */
-  _client(): CDPSession {
+  getCDPSession(): CDPSession {
     throw new Error('Not implemented');
   }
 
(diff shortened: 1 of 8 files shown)


Kikobeats:
maybe related to https://github.com/puppeteer/puppeteer/pull/10515 .cc @OrKoN wdyt?

OrKoN:
Thanks for the PR but I don't think we want to rename it. Not everyone uses TypeScript and the @internal annotation only removes the method from public types and documentation only. We don't want to encourage the use of the Page's CDP session (by providing a name that makes it look like it can be relied upon) because using it might break Puppeteer (now or in the future, e.g., by enabling/disabling domain so that Puppeteer's subscription to events on this session are not working as expected) and because it is an implementation detail that might change very soon (e.g., a new type of target is coming to CDP so in the future one page will have more than one session). The recommended way to access CDP is to create a new CDP session.

Issue #10124, opened 2023-05-03
[Bug]: puppeteer 19.9.* cannot be installed behind a corporate proxy


kevteljeur:
### Bug expectation

I expected to install puppeteer 19.* for Node.js 14.20.1 on Red Hat Enterprise Linux Server release 7.9 (Maipo) behind a corporate proxy.

What happened instead was the error below.
- The Chromium binary can be downloaded from https://storage.googleapis.com/chromium-browser-snapshots/Linux_x64/1108766/chrome-linux.zip using wget or curl
- the proxy environmental variables have been correctly set. 
- A change has been made from puppeteer 19.9.* and up so that this no longer works.
- puppeteer 19.8* can be installed as expected.

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Minimal, reproducible example

```TypeScript
> npm install puppeteer@19.9.*

npm ERR! code 1
npm ERR! path [edit]/node_modules/puppeteer
npm ERR! command failed
npm ERR! command sh -c node install.js
npm ERR! ERROR: Failed to set up Chromium r1108766! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
npm ERR! Error: Download failed: server returned code 407. URL: https://storage.googleapis.com/chromium-browser-snapshots/Linux_x64/1108766/chrome-linux.zip
npm ERR!     at [edit]/node_modules/@puppeteer/browsers/lib/cjs/httpUtil.js:121:31
npm ERR!     at ClientRequest.requestCallback ([edit]/node_modules/@puppeteer/browsers/lib/cjs/httpUtil.js:98:13)
npm ERR!     at Object.onceWrapper (events.js:520:26)
npm ERR!     at ClientRequest.emit (events.js:400:28)
npm ERR!     at HTTPParser.parserOnIncomingClient (_http_client.js:647:27)
npm ERR!     at HTTPParser.parserOnHeadersComplete (_http_common.js:127:17)
npm ERR!     at Socket.socketOnData (_http_client.js:515:22)
npm ERR!     at Socket.emit (events.js:400:28)
npm ERR!     at addChunk (internal/streams/readable.js:293:12)
npm ERR!     at readableAddChunk (internal/streams/readable.js:267:9)

 
npm ERR! A complete log of this run can be found in: [edit]/.npm/_logs/2023-05-03T20_23_05_747Z-debug-0.log
```

### Error string

ERR! Error: Download failed: server returned code 407.

### Puppeteer version

[...]

PR #10548, opened 2023-07-12
fix: stop relying on vm2 (via proxy agent)

### LINKED ISSUE #10547: [Bug]: vm2 is deprecated for critical security issue

```TypeScript
npm install -g puppeteer   
npm WARN deprecated vm2@3.9.19: The library contains critical security issues and should not be used for production! The maintenance of the project has been discontinued. Consider migrating your code to isolated-vm.

```

It looks like a downstream dependency has a critical security issue and needs to be updated. Is it a big lift to update to isolated-vm?
```


### Error string

The library contains critical security issues and should not be used for production!

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

_No response_

### Expectation

Not have a critical security issue

### Reality

It has a critical security issue

### Puppeteer configuration

_No response_

### Puppeteer version

20.8.1

### Node version

16.18.0

### Package manager

npm

### Package manager version

8.19.2

### Operating system

macOS

OrKoN:
Originally, we hoped to support all proxy kinds by integrating the proxy-agent module. It turns out that pac file proxy implementation relies on the vm2 module that recently became deprecated due to security problems. The recommended replacement requires a compiler installed which would lead to extra friction for many users. It looks like the best would be to drop the pac proxy support (which we never announced publicly) and stop relying on the proxy-agent module that has vm2 as a dependency.

Closes #10547

P.S. before landing let's wait if there will be alternative solutions following the vm2 deprecation.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1893,6 +1893,15 @@
         "@types/node": "*"
       }
     },
+    "node_modules/@types/proxy-from-env": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@types/proxy-from-env/-/proxy-from-env-1.0.1.tgz",
+      "integrity": "sha512-luG++TFHyS61eKcfkR1CVV6a1GMNXDjtqEQIIfaSHax75xp0HU3SlezjOi1yqubJwrG8e9DeW59n6wTblIDwFg==",
+      "dev": true,
+      "dependencies": {
+        "@types/node": "*"
+      }
+    },
     "node_modules/@types/semver": {
       "version": "7.5.0",
       "resolved": "https://registry.npmjs.org/@types/semver/-/semver-7.5.0.tgz",
@@ -2181,6 +2190,7 @@
     },
     "node_modules/acorn": {
       "version": "8.8.1",
+      "dev": true,
       "license": "MIT",
       "bin": {
         "acorn": "bin/acorn"
@@ -2196,14 +2206,6 @@
       "dev": true,
       "peerDependencies": {
         "acorn": "^6.0.0 || ^7.0.0 || ^8.0.0"
-      }
-    },
-    "node_modules/acorn-walk": {
-      "version": "8.2.0",
-      "resolved": "https://registry.npmjs.org/acorn-walk/-/acorn-walk-8.2.0.tgz",
-      "integrity": "sha512-k+iyHEuPgSw6SbuDpGQM+06HQUa04DZ3o+F6CSzXMvvI5KMvnaEqXe+YVe555R9nn6GPt404fos4wcgpw12SDA==",
-      "engines": {
-        "node": ">=0.4.0"
       }
     },
     "node_modules/agent-base": {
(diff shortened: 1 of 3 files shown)


jrandolf-google (approved):

mscott3:
proxy-agent was specifically added to fix this bug:
    https://github.com/puppeteer/puppeteer/issues/10124
This change has re-introduced it.

mscott3:
Well, what else can we do?  As I understand it, vm2 was introduced so that proxy config could be picked up from the system pac file.  Can we have some other way to provide that information?  I'd be satisfied if I could pass a flag into the install script to have it pull the proxy information the HTTP_PROXY/HTTPS_PROXY environment variables, and that feels straight forward. 
(I apologize for not being able to provide a PR, but I REALLY HATE JavaScript)

OrKoN:
@mscott3 you can still use other proxy types. For PAC file proxies, there is no solution currently.

mscott3:
Oh!  So, my users problem is that when they:

    npm install -g puppeteer ...

they see this

    npm ERR! code 1
    npm ERR! path /opt/app-root/src/.npm-global/lib/node_modules/puppeteer
    npm ERR! command failed
    npm ERR! **command sh -c -- node install.js**
    npm ERR! ERROR: Failed to set up Chrome r114.0.5735.133! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
    npm ERR! Error: Download failed: server returned code 503. URL: https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/114.0.5735.133/linux64/chrome-linux64.zip
    npm ERR!     at /opt/app-root/src/.npm-global/lib/node_modules/puppeteer/node_modules/@puppeteer/browsers/lib/cjs/httpUtil.js:117:31
    npm ERR!     at ClientRequest.requestCallback (/opt/app-root/src/.npm-global/lib/node_modules/puppeteer/node_modules/@puppeteer/browsers/lib/cjs/httpUtil.js:94:13)
    npm ERR!     at Object.onceWrapper (node:events:628:26)
    npm ERR!     at ClientRequest.emit (node:events:513:28)
    npm ERR!     at HTTPParser.parserOnIncomingClient (node:_http_client:693:27)
    npm ERR!     at HTTPParser.parserOnHeadersComplete (node:_http_common:128:17)
    npm ERR!     at Socket.socketOnData (node:_http_client:534:22)
    npm ERR!     at Socket.emit (node:events:513:28)
    npm ERR!     at addChunk (node:internal/streams/readable:315:12)
    npm ERR!     at readableAddChunk (node:internal/streams/readable:289:9)

whcih I think is consistent with the original bug report.  I interpret this to mean that install.js is not respecting the HTTP_PROXY and HTTPS_PROXY environment variables.  Is there something I can pass on install to indicate what proxy to use?

mscott3:
thank you for looking at this.  For open source dev tool stacks, we use a proxy that performs integrated auth with Kerberous, which is implemented in many open source HTTP stacks.  All of the downloads of components via npm work just fine.
When we get to downloading the browser bits, we start to get these 'ERR_SOCKET_CLOSED' notices.  This does not explicitly say "we could not auth you to the proxy" (that  would be a 407), but it would be consistent with the HTTP library getting a 407, swallowing it, and reporting that the socket closed.  

I put px-proxy (https://pypi.org/project/px-proxy/) between me and our corporate proxy, and my problem went away.
I'll be pursing that as a solution.  

Again, thanks for looking into this.

PR #10549, opened 2023-07-12
fix: prevent erroneous new main frame

adamraine:
For reasons I am still investigating, Chrome will sometimes emit a random `Page.frameNavigated` event that looks like this:

```json
{
  "id": "FC1767E7420AFDD8D5D7CD935DA4C788",
  "loaderId": "E4B73E277690F2C9755A2D0C950D2FA0",
  "url": ":",
  "domainAndRegistry": "",
  "securityOrigin": "://",
  "mimeType": "text/html",
  "adFrameStatus": {
    "adFrameType": "none"
  },
  "secureContextType": "InsecureScheme",
  "crossOriginIsolatedContextType": "NotIsolatedFeatureDisabled",
  "gatedAPIFeatures": []
}
```

The frame ID of this event is *not* the same frame id as the actual main frame of the tree, however it is set as the main frame since it doesn't have a parent id.

It happens when running Lighthouse in DevTools:
https://github.com/GoogleChrome/lighthouse/issues/15073

This patch ensures that the main frame of a frame tree can only be set one time so this random event does not influence the frame tree.

### DIFF
--- a/packages/puppeteer-core/src/common/FrameTree.ts
+++ b/packages/puppeteer-core/src/common/FrameTree.ts
@@ -69,7 +69,7 @@ export class FrameTree<Frame extends BaseFrame> {
         this.#childIds.set(frame._parentId, new Set());
       }
       this.#childIds.get(frame._parentId)!.add(frame._id);
-    } else {
+    } else if (!this.#mainFrame) {
       this.#mainFrame = frame;
     }
     this.#waitRequests.get(frame._id)?.forEach(request => {


Lightning00Blade (approved):

adamraine:
@OrKoN are the CI failures related to this patch?

OrKoN:
@adamraine did you find a reason why this event is emitted?

adamraine:
Not yet, planning on doing more investigation today.

OrKoN:
> @OrKoN are the CI failures related to this patch?

probably not, restarted the failing check.

adamraine:
Update on investigation, this appears to be coming from a [fenced frame](https://developer.chrome.com/en/docs/privacy-sandbox/fenced-frame/) (notice the target subtype). Still not sure exactly how this event gets created though.

```json
{
  "method": "Target.attachedToTarget",
  "params": {
    "sessionId": "221795A5659470D7AC6FCBF3F311D882",
    "targetInfo": {
      "targetId": "F5B7345C9BF9413F71A9B51D22304A82",
      "type": "iframe",
      "title": "",
      "url": "",
      "attached": true,
      "canAccessOpener": false,
      "browserContextId": "6310EBC13B0F02CB468081EC10C1A4B0",
      "subtype": "fenced"
    },
    "waitingForDebugger": true
  },
  "sessionId": "B5668ED04C68E8A4178A04D5EE1EB6E0"
}
```

Frame tree:

```json
{
  "frame": {
    "id": "F5B7345C9BF9413F71A9B51D22304A82",
    "loaderId": "4BDF72DDE41DDD533EFDE652EB563B30",
    "url": ":",
    "domainAndRegistry": "",
    "securityOrigin": "://",
    "mimeType": "text/html",
    "adFrameStatus": {
      "adFrameType": "none"
    },
    "secureContextType": "InsecureScheme",
    "crossOriginIsolatedContextType": "NotIsolatedFeatureDisabled",
    "gatedAPIFeatures": []
  }
}
```

OrKoN:
@adamraine I believe the event might be created in the same place as for regular frames. The problem seems to be that the fenced frame does not report a parent id. It is probably because it is 'fenced' but for CDP purposes we might need to still report it. Would you mind filing a crbug so that we can discuss? cc @caseq

PR #10567, opened 2023-07-18
fix(locators): reject the race if there are only failures

OrKoN:


test/src/locator.spec.ts
```diff
@@ -222,7 +222,9 @@ describe('Locator', function () {
     });
 
     it('should time out', async () => {
-      const clock = sinon.useFakeTimers();
+      const clock = sinon.useFakeTimers({
+        shouldClearNativeTimers: true,
```
OrKoN, line 226: the lack of this config was causing hanging after all tests: apparently, the default behaviour for sinon is not to invoke the native clearTimeout if the timer was started before the fake timers were activated.

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -732,6 +732,26 @@ class RaceLocatorImpl extends Locator {
...
+    if (
+      results.some(result => {
+        return result.status === 'fulfilled';
+      })
+    ) {
+      return;
+    }
+
+    for (const result of results) {
+      if (result.status === 'fulfilled') {
```
Lightning00Blade: How can we get in this block at all?

OrKoN: currently no, but TypeScript does not know

Lightning00Blade (approved):

packages/puppeteer-core/src/api/Locator.ts
OrKoN: updated

PR #10573, opened 2023-07-18
chore: implement typed Locators, expects, and internal contexts

jrandolf-google:
This PR adds types locators, a new locator type called `ExpectedLocator`, internal context management.


test/src/locator.spec.ts
```diff
@@ -113,6 +113,9 @@ describe('Locator', function () {
         .on(LocatorEmittedEvents.Action, () => {
           clicked = true;
         })
+        .expect((element): element is HTMLElement => {
```
OrKoN: IMO it should be possible to click without a need for expect: requiring the expect is too verbose. It should just throw an error if the element is not clickable instead of requiring boilerplate because the outcome is the same, it is going to fail at runtime.

test/src/locator.spec.ts
```diff
@@ -538,4 +538,29 @@ describe('Locator', function () {
       await expect(result).resolves.toEqual(undefined);
     });
   });
+
+  describe('Locator.prototype.filter', () => {
```
OrKoN: so should we call it filter or expect? 🤔

packages/puppeteer-core/src/api/Locator.ts
```diff
@@ -76,17 +83,36 @@ const WAIT_FOR_FUNCTION_DELAY = 100;
 /**
  * @public
  */
-export type ActionCondition = (
-  element: ElementHandle,
-  signal: AbortSignal
+export type ActionCondition<T> = (
+  element: HandleFor<T>,
+  signal?: AbortSignal
```
OrKoN: why is the signal optional?

test/src/locator.spec.ts
```diff
@@ -538,4 +538,29 @@ describe('Locator', function () {
...
+    it('should work', async () => {
+      const clock = sinon.useFakeTimers({
+        shouldClearNativeTimers: true,
+      });
+      try {
+        const {page} = await getTestState();
+        page.setDefaultTimeout(5000);
+        await page.setContent(`<a>test</a>`);
+        const result = page
```
OrKoN, line 551: `page.locator(':-p-text(test)').fill('abc');` should also compile. Let's add d.ts tests for those cases and also let's add tests that the right errors are thrown

OrKoN (changes requested):

test/src/locator.spec.ts
OrKoN: I can add d.ts tests in a separate CL to make it easier

jrandolf: WDYT about doing this once we finalize the API in public? I think adding type tests right now would just be annoying for development since API changes would need to also change type tests.

packages/puppeteer-core/src/api/Locator.ts
OrKoN:
```suggestion
  signal: AbortSignal
```

test/src/locator.spec.ts
OrKoN: SGTM

OrKoN:
changes to chore because the API is internal

test/src/locator.spec.ts
```diff
@@ -538,4 +538,29 @@ describe('Locator', function () {
       await expect(result).resolves.toEqual(undefined);
     });
   });
+
+  describe('Locator.prototype.expect', () => {
```
OrKoN, line 542: boolean predicates don't see to compile

```
          .expect((_element): boolean => {
            return true;
          })
```

OrKoN, line 542: Let's add a few more tests?

The second test does not seem to be working

```
describe.only('Locator.prototype.expect', () => {
    it('should not resolve if the predicate does not match', async () => {
      const clock = sinon.useFakeTimers({
        shouldClearNativeTimers: true,
      });
      try {
        page.setDefaultTimeout(5000);
        await page.setContent(`<a>test</a>`);
        const result = page
          .locator(':-p-text(test)')
          .expect((element): element is HTMLInputElement => {
            return element instanceof HTMLInputElement;
          })
          .fill('abc');
        clock.tick(5100);
        await expect(result).rejects.toEqual(
          new TimeoutError('waitForFunction timed out. The timeout is 5000ms.')
        );
      } finally {
        clock.restore();
      }
    });

    it('should resolve as soon as the predicate matches', async () => {
      const clock = sinon.useFakeTimers({
        shouldClearNativeTimers: true,
      });
      try {
        page.setDefaultTimeout(5000);
        const result = page
          .locator(':-p-text(test)')
          .expect((element): Promise<boolean> => {
            return Promise.resolve(
              element.getAttribute('clickable') === 'true'
            );
          })
          .hover();
        clock.tick(2000);
        await page.evaluate(() => {
          document.querySelector('div')?.setAttribute('clickable', 'true');
        });
        console.log(await page.content());
        clock.tick(2000);
        await expect(result).resolves.toEqual(undefined);
      } finally {
        clock.restore();
      }
    });
  });
  ```

File: docs/api/puppeteer.actioncondition.md [hunk 6]
```diff
@@ -7,10 +7,10 @@ sidebar_label: ActionCondition
 #### Signature:
 
 ```typescript
-export type ActionCondition = (
-  element: ElementHandle,
-  signal: AbortSignal
+export type ActionCondition<T> = (
+  element: HandleFor<T>,
+  signal?: AbortSignal
```
OrKoN:
```suggestion
  signal: AbortSignal
```

OrKoN (approved):

PR #10593, opened 2023-07-19
fix: remove typescript from peer dependencies

pauldraper:

Packaging

N/A

N/A

**Summary**

The typescript package is one of the largest npm packages (~40MB), and completely unused by puppeteer at runtime.

Downstream users should not install it.

No.

**Other information**

I *think* someone may have been trying to limit the TypeScript version at compile-time?

(If so, that has a couple problems. First, is excessive costs on non-TypeScript users. Second, is that there is zero guarantee that the TypeScript compiler is installed via npm and in the same context as @puppeteer/core.)

Assuming this was the idea, I added the version to `typesVersions`, which is a more efficient and reliable mechanism.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ include `$HOME/.cache` into the project's deployment.
 For a version of Puppeteer without the browser installation, see
 [`puppeteer-core`](#puppeteer-core).
 
+If used with TypeScript, the minimum supported TypeScript version is `4.7.4`.
+
 #### Configuration
 
 Puppeteer uses several defaults that can be customized through configuration
--- a/docs/api/puppeteer.connect.md
+++ b/docs/api/puppeteer.connect.md
@@ -7,7 +7,6 @@ sidebar_label: connect
 #### Signature:
 
 ```typescript
-connect: (
-  options: import('puppeteer-core/internal/puppeteer-core.js').ConnectOptions
-) => Promise<import('puppeteer-core/internal/puppeteer-core.js').Browser>;
+connect: (options: import('puppeteer-core').ConnectOptions) =>
+  Promise<import('puppeteer-core').Browser>;
 ```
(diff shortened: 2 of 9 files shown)


OrKoN:
@pauldraper could you clarify why an optional peer dependency causes this problem?

Lightning00Blade:
[From what I understand it does not matter that we specify that the ](https://github.blog/2020-10-13-presenting-v7-0-0-of-the-npm-cli/#)
https://github.com/npm/rfcs/blob/main/implemented/0025-install-peer-deps.md#f-use-peerdependenciesmeta-to-trigger-auto-install
I think this can be solved by just adding `"autoinstall": false` to all our peerDependecies.

pauldraper:
> could you clarify why an optional peer dependency causes this problem?

peerDependencies are installed by default in npm v7+ (and in old versions, like pre-v3). TypeScript happens to be one of the largest npm packages (40m+).

Moreover, a peer dependency for the compiler is not a good way of attesting compatibility of TypeScript declarations. There is no implicit or explicit requirement for the TypeScript compiler to be anywhere in the install tree.

`typesVersions` is explicitly made for this. 
https://github.com/microsoft/TypeScript/issues/32166 (Though my personal recommendation would be to remove the version requirement, and just let it fail.)

OrKoN:
@pauldraper interesting, I don't observe that behaviour with npm@9.6.7, i.e., optional peer dependencies are not installed.

jrandolf-google (approved):

pauldraper:
Hm, it seems you're right. I was seeing it for another reason.

I still recommending remove that as it's a flawed mechanism for attesting the compatibility of the TS declarations, and package mangers vary in their treatment of option deps, peer dependencies, etc.

Is there a way to re-run CI / chrome-new-headless tests on windows-latest ?

pauldraper:
Thanks!

PR #10599, opened 2023-07-20
chore: update clean command

jrandolf-google:


packages/browsers/package.json
```diff
@@ -6,7 +6,7 @@
     "build:docs": "wireit",
     "build": "wireit",
     "build:test": "wireit",
-    "clean": "tsc --build --clean && rm -rf lib",
+    "clean": "git clean -dXf .",
```
OrKoN, line 9: aren't some dependencies installed to the package's node modules?

PR #10603, opened 2023-07-20
chore: implement type signatures for internal Locator methods

jrandolf-google:


packages/puppeteer-core/src/api/locators/NodeLocator.ts
```diff
@@ -299,24 +299,20 @@ export class NodeLocator<T extends Node> extends Locator<T> {
...
+        await Promise.all(
+          allConditions.map(check => {
+            return check(element, signal);
+          })
+        );
+        signal?.throwIfAborted();
+        // 3. Perform the action
+        this.emit(LocatorEmittedEvents.Action);
+        await action(element);
+        return element;
```
OrKoN: when will the element handle be disposed?

OrKoN (approved):

PR #10607, opened 2023-07-20
chore: use RxJS for locator implementation

jrandolf-google:
Using RxJS greatly simplifies the control flow for locators and comes with automatic cleanup on failure. It greatly simplifies the `signal` logic and the retry logic.


packages/puppeteer-core/src/api/locators/Locator.ts
```diff
@@ -113,33 +125,84 @@ export interface LocatorEventObject {
...
+              );
+            })
+          )
+        );
+      }
+      return pipe(
+        // For observables coming from promises, a delay is needed, otherwise
+        // RxJS will never yield in a permanent failure for a promise. (This is
+        // correct behavior).
+        retry({delay: 0}),
```
OrKoN: should that be more than 0, e.g., `WAIT_FOR_FUNCTION_DELAY` we had before? currently, it looks like retry would be happening too often.

packages/puppeteer-core/src/api/locators/Locator.ts
```diff
@@ -113,33 +125,84 @@ export interface LocatorEventObject {
...
+    },
+    retryAndRaceWithSignalAndTimer: <T>(
+      signal?: AbortSignal
+    ): OperatorFunction<T, T> => {
+      const candidates = [];
+      if (signal) {
+        candidates.push(
+          fromEvent(signal, 'abort').pipe(
+            map(() => {
+              throw new Error('Locator aborted');
```
OrKoN: I think it should be an abort error

packages/puppeteer-core/src/api/locators/Locator.ts
```diff
@@ -113,33 +125,84 @@ export interface LocatorEventObject {
...
+      return mergeMap((handle: HandleFor<T>) => {
+        return merge(
+          ...conditions.map(condition => {
+            return condition(handle, signal);
+          })
+        ).pipe(defaultIfEmpty(handle));
+      });
+    },
+    retryAndRaceWithSignalAndTimer: <T>(
+      signal?: AbortSignal
```
OrKoN, line 201: do we need to pass down the signals all the way down? I thought rxjs would handle the cancellation and we would just need to unsubscribe from the observable on the top-level?

packages/puppeteer-core/src/api/locators/Locator.ts
jrandolf: Hmm, yes we can extend the delay further out. I'll add it.

packages/puppeteer-core/src/api/locators/Locator.ts
jrandolf: You're right that we don't need to do this, but also keep in mind that it's slightly confusing to see the condition without the timeout. I have no opinion about this though, I can make the change if you feel stronger about this.

OrKoN: I think timeout can remain but I wonder about the abort signal. How do we abort relying on rxjs?

packages/puppeteer-core/src/api/locators/Locator.ts
```diff
@@ -113,33 +136,82 @@ export interface LocatorEventObject {
...
-  ): Locator<UnionLocatorOf<Locators>> {
-    return new RaceLocator(
-      locators as Array<Locator<UnionLocatorOf<Locators>>>
-    );
-  }
+  protected visibility: VisibilityOption = null;
+  /**
+   * @internal
+   */
+  timeout = 30000;
```
Lightning00Blade: Just use protected or private with a public getter. Also 30_000.

packages/puppeteer-core/src/api/locators/Locator.ts
```diff
@@ -162,40 +234,449 @@ export abstract class Locator<T> extends EventEmitter {
...
+          return from(handle.isIntersectingViewport({threshold: 0}));
+        }).pipe(first(identity), retry({delay: RETRY_DELAY}), ignoreElements());
+      })
+    );
+  };
+
+  /**
+   * @internal
+   */
+  clickImpl<ElementType extends Element>(
```
Lightning00Blade: Not a fan of the `Impl` surfix, a `_` prefix sound to be more consistent with internal use.

packages/puppeteer-core/src/api/locators/ExpectedLocator.ts
```diff
@@ -1,16 +1,14 @@
-import {Awaitable} from '../../common/common.js';
+import {
```
Lightning00Blade, line 17: Licence header, verify other files have it too.

packages/puppeteer-core/src/api/locators/Locator.ts
```diff
@@ -1,28 +1,51 @@
...
-}
-
-/**
- * @internal
- */
-export const LOCATOR_CONTEXTS = new WeakMap<
-  Locator<unknown>,
-  LocatorContext<never>
->();
+export const RETRY_DELAY = 100;
```
Lightning00Blade, line 64: Non-blocking: That looks like a little bit too long imo I would set it to 50ms, @OrKoN what do you think.

packages/puppeteer-core/src/api/locators/Locator.ts
```diff
@@ -113,33 +136,82 @@ export interface LocatorEventObject {
...
-   * Creates an expectation that is evaluated against located values.
-   *
-   * If the expectations do not match, then the locator will retry.
-   *
    * @internal
    */
-  expect<S extends T>(predicate: Predicate<T, S>): Locator<S> {
-    return new ExpectedLocator(this, predicate);
-  }
+  protected operators = {
```
Lightning00Blade, line 187: Why a variable holding the operators and not methods, I can't find a place we override this.
Nit: I don't think we need the internal if we use protected.

packages/puppeteer-core/src/api/locators/Locator.ts
OrKoN: it's been 100ms in the previous version. 100ms sounds good to me.

packages/puppeteer-core/src/api/locators/Locator.ts
jrandolf: `protected` still requires an @internal notation to hide it from docs.

packages/puppeteer-core/src/api/locators/Locator.ts
jrandolf: `protected` still requires an @internal notation to hide it from docs.

jrandolf: Using a variable namespaces the operators. I think it's better than hoisting them to the top polluting the class.

packages/puppeteer-core/src/api/locators/Locator.ts
```diff
@@ -113,33 +152,82 @@ export interface LocatorEventObject {
...
+    retryAndRaceWithSignalAndTimer: <T>(
+      signal?: AbortSignal
+    ): OperatorFunction<T, T> => {
+      const candidates = [];
+      if (signal) {
+        candidates.push(
+          fromEvent(signal, 'abort').pipe(
+            map(() => {
+              // SAFETY: This will throw.
+              return signal.throwIfAborted() as never;
```
OrKoN: I have discovered that throwIfAborted was only added in Node 17. Let's see if we can throw the same error ourselves?

packages/puppeteer-core/src/api/locators/Locator.ts
OrKoN: I think that would look less confusing as a protected static property?

jrandolf: The operators themselves use `this`, otherwise I would have moved them entirely to some `operators.ts` file.

jrandolf: Ah, except for `retry`, but that said, since the other one uses `this`, I still kept them together for convenience.

packages/puppeteer-core/third_party/rxjs/rxjs.ts
```diff
@@ -0,0 +1,41 @@
+/**
+ * Copyright 2022 Google Inc. All rights reserved.
```
OrKoN:
```suggestion
 * Copyright 2023 Google Inc. All rights reserved.
```

OrKoN (approved):

PR #10615, opened 2023-07-21
chore: support targets for bidi

OrKoN:


packages/puppeteer-core/src/common/bidi/BrowserContext.ts
```diff
@@ -56,6 +58,44 @@ export class BrowserContext extends BrowserContextBase {
...
+    return this.#browser.targets().filter(target => {
+      return target.browserContext() === this;
+    });
+  }
+
+  /**
+   * This searches for a target in this specific browser context.
+   *
+   * @example
+   * An example of finding a target for a page opened via `window.open`:
```
Lightning00Blade: No need for comment here.

packages/puppeteer-core/src/common/bidi/BrowserContext.ts
```diff
@@ -120,12 +140,12 @@ export class BrowserContext extends BrowserContextBase {
       throw new Error('Default context cannot be closed!');
     }
 
-    for (const page of this.#pages.values()) {
-      await page?.close().catch(error => {
+    for (const target of this.#targets.values()) {
+      await (await target?.page())?.close().catch(error => {
```
Lightning00Blade: Can we extract it in a page variable it's gonna be more readable imo.

Lightning00Blade (approved):

PR #10620, opened 2023-07-24
chore: update chromium-bidi

Lightning00Blade:


test/TestExpectations.json
```diff
@@ -3749,6 +3773,12 @@
     "parameters": ["firefox", "webDriverBiDi"],
     "expectations": ["PASS"]
   },
+  {
+    "testIdPattern": "Hook failed!",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["chrome", "webDriverBiDi"],
```
OrKoN: remove

OrKoN (approved):

packages/puppeteer-core/src/common/bidi/BidiOverCDP.ts
```diff
@@ -86,7 +86,11 @@ class CDPConnectionAdapter {
       throw new Error('Unknown CDP session with id' + id);
     }
     if (!this.#adapters.has(session)) {
-      const adapter = new CDPClientAdapter(session);
+      const adapter = new CDPClientAdapter(
+        session,
+        id,
+        this.#browser as unknown as BidiMapper.CdpClient
```
jrandolf: Write a comment here why we do the `as unknown` cast

jrandolf: As well as a `TODO`

jrandolf-google (approved):

PR #10621, opened 2023-07-24
refactor: move context actions to the browser

OrKoN:


packages/puppeteer-core/src/common/bidi/Browser.ts
```diff
@@ -83,18 +86,27 @@ export class Browser extends BrowserBase {
...
     });
 
-    return new Browser({
+    const browser = new Browser({
       ...opts,
       browserName,
       browserVersion,
     });
+
+    await browser._init.valueOrThrow();
```
Lightning00Blade: Why the deferred, we can call the browser.#getTree() from here.

Lightning00Blade (approved):

PR #10622, opened 2023-07-24
test: upstream test and expectation for Firefox sync to puppeteer v20.9.0

juliandescottes:
This is a sync for test changes we need for the latest sync:
- https://phabricator.services.mozilla.com/D184113
- https://phabricator.services.mozilla.com/D184114
- https://phabricator.services.mozilla.com/D184115
- https://phabricator.services.mozilla.com/D184116

Basically:
- in Mozilla CI, headless environments can really not run headful browsers, so I had to skip headful.spec.ts tests on firefox headful (I'm not 100% how rules can override each other so maybe there is a cleaner way to do that).
- adding a missing await on a page.goto which was sometimes triggering failures later in the suite
- skipping tests which require a regular install (we don't use one on Mozilla CI), but it would be better to have a flag for this in the expectations file
- skipping the pdf timeout test, as it leads to frequent intermittents due to https://bugzilla.mozilla.org/show_bug.cgi?id=1841125

Let us know what you think, cc @OrKoN @Lightning00Blade

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -216,6 +216,12 @@
     "expectations": ["PASS"]
   },
   {
+    "testIdPattern": "[page.spec] Page Page.pdf should respect timeout",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox"],
+    "expectations": ["SKIP"]
+  },
+  {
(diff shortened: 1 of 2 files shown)


juliandescottes:
(I removed one unnecessary skip)

OrKoN (approved):

PR #10629, opened 2023-07-25
feat: implement `Locator.prototype.wait`

jrandolf-google:


test/src/locator.spec.ts
```diff
@@ -569,4 +569,20 @@ describe('Locator', function () {
       }
     });
   });
+
+  describe('Locator.prototype.wait', () => {
+    it('should work', async () => {
+      const {page} = await getTestState();
+      await page.setContent(`
```
OrKoN:
```suggestion
     page.setContent(`
```

to be sure that the div is not there

packages/puppeteer-core/src/api/locators/Locator.ts
```diff
@@ -635,6 +635,20 @@ export abstract class Locator<T> extends EventEmitter {
...
+   *
+   * Note this requires the value to be JSON-serializable.
+   */
+  async wait(options?: Readonly<ActionOptions>): Promise<T> {
+    const handle = await firstValueFrom(
+      this._wait(options).pipe(
+        this.operators.retryAndRaceWithSignalAndTimer(options?.signal)
+      )
+    );
+    return handle.jsonValue();
```
OrKoN: when is the handle disposed?

docs/api/puppeteer.locator.wait.md
```diff
@@ -0,0 +1,27 @@
+---
+sidebar_label: Locator.wait
+---
+
+# Locator.wait() method
+
+Waits for the locator to get the serialized value from the page.
```
OrKoN, line 7: Please update https://pptr.dev/guides/locators with an example on how to use this

jrandolf: I will do this in a separate PR because as of now, `wait` doesn't do much without `map`. Is that fine for you?

test/src/locator.spec.ts
```diff
@@ -569,4 +569,20 @@ describe('Locator', function () {
       }
     });
   });
+
+  describe('Locator.prototype.wait', () => {
+    it('should work', async () => {
```
OrKoN, line 574: the test does not work for bidi https://github.com/puppeteer/puppeteer/actions/runs/5656280482/job/15323075013?pr=10629

jrandolf: Done.

OrKoN (approved):

PR #10630, opened 2023-07-25
feat: implement `Locator.prototype.map`

jrandolf-google:


test/src/locator.spec.ts
```diff
@@ -540,6 +540,50 @@ describe('Locator', function () {
     });
   });
 
+  describe.only('Locator.prototype.map', () => {
```
OrKoN: let's also add tests involving combinations of locators, e.g., `map` + `expect` and `expect` + `map`

packages/puppeteer-core/src/api/locators/Locator.ts
```diff
@@ -649,6 +653,15 @@ export abstract class Locator<T> extends EventEmitter {
     return handle.jsonValue();
   }
 
+  /**
+   * Maps the locator using the provided mapper.
+   *
+   * @public
+   */
+  map<To>(mapper: Mapper<T, To>): Locator<To> {
```
OrKoN, line 665: Let's keep the guide updates as soon as new features are pushed.

jrandolf: I'll add it in the next PR

OrKoN (approved):

PR #10632, opened 2023-07-25
feat: implement function locators

jrandolf-google:


test/src/locator.spec.ts
```diff
@@ -647,4 +647,34 @@ describe('Locator', function () {
...
+          return new Promise<boolean>(resolve => {
+            return setTimeout(() => {
+              return resolve(true);
+            }, 100);
+          });
+        })
+        .wait();
+      await expect(result).resolves.toEqual(true);
+    });
+    it('should work with actions', async () => {
```
OrKoN, line 690: what happens if the function locator does not return an HTML element?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -841,8 +843,26 @@ export class Page extends EventEmitter {
...
+  /**
+   * Creates a locator for the provided function. See {@link Locator} for
+   * details and supported actions.
+   *
+   * @remarks
+   * Locators API is experimental and we will not follow semver for breaking
+   * change in the Locators API.
+   */
+  locator<Ret>(func: () => Awaitable<Ret>): Locator<Ret>;
+
```
OrKoN:
```suggestion
```

test/src/locator.spec.ts
jrandolf: TypeScript ensures that `click` cannot happen for a non-Element locator.

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -832,7 +838,7 @@ export class Page extends EventEmitter {
   }
 
   /**
-   * Creates a locator for the provided `selector`. See {@link Locator} for
+   * Creates a locator for the provided function. See {@link Locator} for
```
OrKoN: should this one still say selector?

docs/guides/locators.md
```diff
@@ -1,9 +1,9 @@
 # Locators
 
-Locators is a new experimental API that combines `waitForSelector` and element
-actions in a single unit. In combination with additional precondition checks
-this allows locators to retry failed actions automatically leading to less flaky
-automation scripts.
+Locators is an innovative, experimental API that combines the functionalities of
```
OrKoN:
```suggestion
Locators is a new, experimental API that combines the functionalities of
```

sounds too pompous, IMO

docs/guides/locators.md
```diff
@@ -1,9 +1,9 @@
...
+waiting and actions into a unified unit. With additional precondition checks, it
```
OrKoN:
```suggestion
waiting and actions. With additional precondition checks, it
```

"unified unit" does not sound too good, unit is already unified by definition.

docs/guides/locators.md
```diff
@@ -12,7 +12,30 @@ in the Locators API.
...
+
+- Waits for the element to become
+  [visible](https://pptr.dev/api/puppeteer.elementhandle.isvisible/) or hidden.
+
+### Waiting for a function
+
+```ts
+await page
+  .locator(() => {
+    return new Promise(resolve => setTimeout(resolve, 500));
```
OrKoN: let's write an example that actually locates something? I cannot think of a good example, perhaps, using an Intersection observer or some other API to locate an element based on custom logic?

OrKoN (approved):

PR #10634, opened 2023-07-25
refactor: target events for bidi

OrKoN:


packages/puppeteer-core/src/common/bidi/Browser.ts
```diff
@@ -145,11 +168,20 @@ export class Browser extends BrowserBase {
     // TODO: once more browsing context types are supported, this should be
     // updated to support those. Currently, all top-level contexts are treated
     // as pages.
+    const browserContexts = this.browserContexts();
```
Lightning00Blade: Let's get the BrowserContext here and use `.at(-1)` that way it shorter

packages/puppeteer-core/src/common/bidi/Browser.ts
```diff
@@ -192,6 +230,14 @@ export class Browser extends BrowserBase {
       'browsingContext.contextCreated',
       this.#onContextCreated
     );
+    this.#connection.off(
+      'browsingContext.fragmentNavigated',
```
Lightning00Blade: Style question wouldn't a loop here and on subscription be better that way we add a map like the reset of the classes in BiDi?

Lightning00Blade (approved):

PR #10644, opened 2023-07-26
chore: sync latest Firefox preference changes for translation feature

juliandescottes:
Sync from https://bugzilla.mozilla.org/show_bug.cgi?id=1845518. 

In between the `browser.translations.enable` got removed from the file, and this last bug instead added `browser.translations.automaticallyPopup`.

### DIFF
--- a/packages/browsers/src/browser-data/firefox.ts
+++ b/packages/browsers/src/browser-data/firefox.ts
@@ -156,9 +156,8 @@ function defaultProfilePreferences(
     // Do not warn when multiple tabs will be opened
     'browser.tabs.warnOnOpen': false,
 
-    // Disable page translations, which can cause issues with tests.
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

whimboo:
We probably should not have merged that yet given that we still have regressions due to this preferences change. It only affects tsan and ccov builds but not sure if there are other side effects as well. And [this pref doesn't seem to fix the problem](https://bugzilla.mozilla.org/show_bug.cgi?id=1845518#c3).

Maybe lets discuss how to handle it on Matrix.

whimboo:
We can actually leave this change in. We pushed another patch for Firefox itself to correctly handle that preference in case of Marionette or our WebDriver BiDi implementation is active.

PR #10649, opened 2023-07-27
chore: specialize `Locator.prototype.setWaitForEnabled`

jrandolf-google:


OrKoN (changes requested):
This expectation should be on by default and users should not need to define it.

OrKoN (approved):

PR #10651, opened 2023-07-27
feat: implement improved Drag n' Drop APIs

jrandolf-google:
This PR implements a new Drag n' Drop API that works with the newly stabilized mouse API.

## Deprecations

 - Drag interception via `Page.setDragInterception` is now deprecated. Users can implement this using native APIs within the browser (e.g. listening to the `dragstart` event).
 - Using `ElementHandle.drop` to drop custom drag data is deprecated.
 - `ElementHandle.dragOver` and `ElementHandle.dragEnter` are deprecated as they serve no purpose now.
 - Using `ElementHandle.drag` to retrieve drag data is now a deprecated use case.
 - `ElementHandle.dragAndDrop` is deprecated in favor of the new `ElementHandle.drop` overload (see Features).

## Features

 - A new overload for `ElementHandle.drop` allows elements to drop onto other elements.
   - It's not necessary to `drag` before `drop`ping. This is automagically done.
 - A new overload for `ElementHandle.drag` allows elements to drag onto other elements.


Lightning00Blade (commented):
LGTM + implementation looks. Need to look over the test before approving.

test/src/drag-and-drop.spec.ts
```diff
@@ -62,18 +71,10 @@ describe('Input.drag', function () {
     using draggable = (await page.$('#drag'))!;
     const data = await draggable.drag({x: 1, y: 1});
     using dropzone = (await page.$('#drop'))!;
+    assert(data instanceof Object);
```
Lightning00Blade:
```suggestion
    assert(data instanceof Object);
    using dropzone = (await page.$('#drop'))!;
```

test/src/drag-and-drop.spec.ts
```diff
@@ -85,24 +86,11 @@ describe('Input.drag', function () {
...
```
Lightning00Blade, line 88: Same as above.

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -731,59 +732,134 @@ export abstract class ElementHandle<
...
+      // is most likely something implying the mouse is no longer dragging.
+      await dataOrElement.drag(this);
+      page._isDragging = false;
+      await page.mouse.up();
+    }
   }
 
   /**
-   * This method triggers a dragenter, dragover, and drop on the element.
+   * @deprecated Use `ElementHandle.drop` instead.
```
Lightning00Blade, line 845: Change to link, please

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -731,59 +732,134 @@ export abstract class ElementHandle<
...
    */
   async drop(
     this: ElementHandle<Element>,
     data?: Protocol.Input.DragData
   ): Promise<void>;
-  async drop(this: ElementHandle<Element>): Promise<void> {
-    throw new Error('Not implemented');
+
+  /**
+   * @internal
```
Lightning00Blade, line 819: Why is this internal when we are pointing people to use it?

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -731,59 +732,134 @@ export abstract class ElementHandle<
   }
 
   /**
-   * This method creates and captures a dragevent from the element.
+   * Drags an element to the given element or point.
```
Lightning00Blade: Suggestion, as I am unsure about wording as currently my first read went to `drag into` (perform drag&drop).
```suggestion
   * Drags an element to over element or point.
```

test/TestExpectations.json
```diff
@@ -1691,6 +1709,12 @@
     "parameters": ["cdp", "firefox"],
     "expectations": ["FAIL"]
   },
+  {
+    "testIdPattern": "[drag-and-drop.spec] Drag n' Drop should drag and drop",
```
Lightning00Blade, line 1713: Why is this failing?

packages/puppeteer-core/src/api/ElementHandle.ts
jrandolf: Link doesn't work because we can't point to the overload.

packages/puppeteer-core/src/api/ElementHandle.ts
jrandolf: The overloads are public, but the implementation is internal. See the docs.

test/TestExpectations.json
jrandolf: Because an extra `dragover` occurs with CDP. This distinction is important; it implies CDP will trigger an event twice in the same coordinate.

Lightning00Blade (approved):
Update docs

PR #10659, opened 2023-07-28
chore: add Dialogs for BiDi

Lightning00Blade:


test/src/page.spec.ts
```diff
@@ -57,7 +57,7 @@ describe('Page', function () {
       await newPage.close();
       expect(await browser.pages()).not.toContain(newPage);
     });
-    it('should run beforeunload if asked for', async () => {
+    it.only('should run beforeunload if asked for', async () => {
```
OrKoN:
```suggestion
    it('should run beforeunload if asked for', async () => {
```

OrKoN (approved):

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -384,12 +398,15 @@ export class Page extends PageBase {
     if (this.#closedDeferred.finished()) {
       return;
     }
-    this.#closedDeferred.resolve(new TargetCloseError('Page closed!'));
-    this.#networkManager.dispose();
 
     await this.#connection.send('browsingContext.close', {
       context: this.mainFrame()._id,
     });
```
OrKoN, line 407: missing call to this.dispose()?

OrKoN: I think it would not be possible to re-order it like this though: you cannot call dispose before sending browsingContext.close but it would be too later after calling browsingContext.close.

PR #10660, opened 2023-07-28
chore: implement partial `ElementHandle.prototype.boundingBox`

jrandolf-google:


test/TestExpectations.json
```diff
@@ -3815,6 +3839,12 @@
     "parameters": ["firefox", "headful", "webDriverBiDi"],
     "expectations": ["FAIL"]
   },
+  {
+    "testIdPattern": "[mouse.spec] Mouse should resize the textarea",
+    "platforms": ["darwin"],
```
OrKoN: does it only pass in headless on a single platform?

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -886,7 +886,24 @@ export class ElementHandle<
    * or `null` if the element is not visible.
    */
   async boundingBox(): Promise<BoundingBox | null> {
-    throw new Error('Not implemented');
+    if (this.frame.parentFrame()) {
+      throw new Error(
+        'Elements within nested iframes are currently not supported.'
+      );
+    }
+    const box = await this.evaluate(element => {
```
OrKoN: is it in the main word or in the puppeteer world?

jrandolf: Main world.

jrandolf: But that shouldn't matter.

OrKoN: Could you add a todo that it should be done in the isolated world?

OrKoN: It matters for the edge cases because the page can override getBoundingClientRect

jrandolf: Hmm, gotcha. Let me switch to the puppeteer world.

jrandolf: Forgot that we can't because we don't have sandboxes yet ;)

I've added a TODO.

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -886,7 +886,25 @@ export class ElementHandle<
    * or `null` if the element is not visible.
    */
   async boundingBox(): Promise<BoundingBox | null> {
```
OrKoN: hm I have just noticed that this is in the wrong file. This would override the CDP implementation in common. Instead this should be defined in common/bidi/ElementHandle where also the sandboxes should be available.

OrKoN: scratch the part about overriding the CDP implementation :)

packages/puppeteer-core/src/api/ElementHandle.ts
OrKoN: we do have sandboxes:

```
this.frame.isolatedRealm()
```

OrKoN (approved):
LGTM % tests pass

consider integrating sandboxes via this.frame.isolatedRealm().evaluate()

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -886,7 +886,25 @@ export class ElementHandle<
...
-    throw new Error('Not implemented');
+    if (this.frame.parentFrame()) {
+      throw new Error(
+        'Elements within nested iframes are currently not supported.'
+      );
+    }
+    // TODO(jrandolf): Use sandboxes when available.
+    const box = await this.evaluate(element => {
+      if ('offsetParent' in element && element.offsetParent === null) {
```
Lightning00Blade: Can you add the link to MDN explaining this part. Else it feels random check.

packages/puppeteer-core/src/api/ElementHandle.ts
jrandolf: It gets overridden in CDP.

packages/puppeteer-core/src/api/ElementHandle.ts
jrandolf: I've removed the check for now. I think this needs to be reevaluated more thoroughly.

packages/puppeteer-core/src/api/ElementHandle.ts
jrandolf: But yes, the sandbox is available on the BiDi ElementHandle! I've changed the implementation.

Lightning00Blade (approved):

PR #10667, opened 2023-07-31
fix: clear pending TLS socket handle

christian-bromann:

I recognised when calling `canDownload` the `@puppeteer/browsers` package makes a `HEAD` request, e.g. to `https://edgedl.me.gvt1.com/edgedl/chrome/chrome-for-testing/116.0.5793.0/mac-arm64/chrome-mac-arm64.zip`. This causes the Node.js process to stale as there stay pending `TLSSocket` around

I am currently working on https://github.com/webdriverio/webdriverio/pull/10767 and have been encountered this issue. Adding given code line resolves it.

My current workaround is:

```ts
setTimeout(() => {
    for (const handle of process._getActiveHandles()) {
        if (handle.servername && handle.servername.includes('edgedl.me')) {
            handle.destroy()
        }
    }
}, 10)
```

**Summary**

I don't think so.

**Other information**

### DIFF
--- a/packages/browsers/src/httpUtil.ts
+++ b/packages/browsers/src/httpUtil.ts
@@ -27,6 +27,8 @@ export function headHttpRequest(url: URL): Promise<boolean> {
       url,
       'HEAD',
       response => {
+        // consume response data free node process
+        response.resume();
         resolve(response.statusCode === 200);
       },
       false


packages/browsers/src/httpUtil.ts
```diff
@@ -27,6 +27,7 @@ export function headHttpRequest(url: URL): Promise<boolean> {
       url,
       'HEAD',
       response => {
+        request.destroy();
```
Lightning00Blade: We should use `response.resume();` as the issue is that the response is never released.
```suggestion
        // consume response data free node process
        response.resume();
```

Lightning00Blade:
Thank you for your PR, I left a small comment and after #10668 is merged test should pass as well.

christian-bromann:
Thanks for the review @Lightning00Blade!

You are totally right, `request.resume()` is the right thing to call here. Update the PR and PR message, everything should pass now.

Lightning00Blade (approved):

PR #10672, opened 2023-08-01
fix: destroy puppeteer utility on context destruction

jrandolf-google:


packages/puppeteer-core/src/common/bidi/Sandbox.ts
```diff
@@ -76,13 +75,9 @@ export class Sandbox implements RealmBase {
   }
 
   async document(): Promise<ElementHandle<Document>> {
-    if (this.#document) {
-      return this.#document;
-    }
-    this.#document = await this.#realm.evaluateHandle(() => {
+    return await this.#realm.evaluateHandle(() => {
```
Lightning00Blade, line 80: Please add a comment here as well.

packages/puppeteer-core/src/common/bidi/Realm.ts
```diff
@@ -46,6 +46,17 @@ export class Realm extends EventEmitter {
...
     this.#frame = frame;
+
+    // TODO(jrandolf): We should try to find a less brute-force way of doing
+    // this.
+    this.connection.on(
+      Bidi.ChromiumBidi.Script.EventNames.RealmDestroyed,
+      async () => {
+        const util = this.internalPuppeteerUtil;
+        delete this.internalPuppeteerUtil;
+        void (await util)?.dispose().catch(() => {});
```
Lightning00Blade: We should send it to debugError.

packages/puppeteer-core/src/common/bidi/Realm.ts
```diff
@@ -46,6 +46,17 @@ export class Realm extends EventEmitter {
...
   setFrame(frame: Frame): void {
     this.#frame = frame;
+
+    // TODO(jrandolf): We should try to find a less brute-force way of doing
+    // this.
+    this.connection.on(
+      Bidi.ChromiumBidi.Script.EventNames.RealmDestroyed,
+      async () => {
+        const util = this.internalPuppeteerUtil;
+        delete this.internalPuppeteerUtil;
```
Lightning00Blade: Just set it to `undefined`

packages/puppeteer-core/src/common/bidi/Realm.ts
jrandolf: We don't want to do this because we assume that it can error here.

Lightning00Blade: `.dispose().catch(debugError);` why would it be a problem to do this?

jrandolf: It'd be noisy because we know that the utility can error here.

Lightning00Blade (approved):

PR #10689, opened 2023-08-03
chore: implement `exposeFunction`

jrandolf-google:


packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -549,6 +550,224 @@ export class Page extends PageBase {
...
+
+    type SendArgsChannel = (value: [id: number, args: Args]) => void;
+    type SendResolveChannel = (
+      value: [id: number, resolve: (ret: FlattenHandle<Awaited<Ret>>) => void]
+    ) => void;
+    type SendRejectChannel = (
+      value: [id: number, reject: (error: unknown) => void]
+    ) => void;
+
+    interface RemotePromiseCallbacks {
```
Lightning00Blade, line 610: Define types outside of the class, I don't think we ever do this.

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -549,6 +550,224 @@ export class Page extends PageBase {
     }
   }
 
+  #exposedFunctionId = Math.random();
```
Lightning00Blade: Can we use a generator function that gives us the 3 ids. Similar to how we do it in the Connection with Callbacks. Returns something like `channel = {args, resolve, reject}`

packages/puppeteer-core/src/common/bidi/Realm.ts
```diff
@@ -164,9 +164,9 @@ export class Realm extends EventEmitter {
 export function getBidiHandle(
   realmOrContext: Realm,
   result: Bidi.Script.RemoteValue,
-  frame: Frame
+  frame?: Frame
```
Lightning00Blade: This feels incorrect. Cant you get try to get the frame from the `script.Source`? 
Also how do we distinguish at the moment if its for this page or not. 
I think we can try to get the Frame from the `realmOrContext` if we need to.

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -549,6 +550,224 @@ export class Page extends PageBase {
...
+  override async exposeFunction<Args extends unknown[], Ret>(
+    name: string,
```
Lightning00Blade, line 575: Currently its possible to create 2 expose functions with the same name. we should keep the functionality that prevents that.

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -549,6 +550,224 @@ export class Page extends PageBase {
...
+    ) => void;
+
+    interface RemotePromiseCallbacks {
+      resolve: Deferred<Bidi.Script.RemoteValue>;
+      reject: Deferred<Bidi.Script.RemoteValue>;
+    }
+
+    const bindingMaps = new Map<string, Map<number, RemotePromiseCallbacks>>();
+
+    this.#connection.on(
```
Lightning00Blade: Let's create a class and track this Listeners as we don't keep track of them and if we close the page they will be left there.

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -549,6 +550,224 @@ export class Page extends PageBase {
...
+        },
+      },
+    ];
+
+    await this.#connection.send('script.addPreloadScript', {
+      functionDeclaration,
+      arguments: channelArguments,
+    });
+
+    await this.#connection.send('script.callFunction', {
```
Lightning00Blade, line 773: You should not need this. Preload script should be executed on all existing browsingContexts/realms as well.

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -549,6 +550,224 @@ export class Page extends PageBase {
...
+      functionDeclaration,
+      arguments: channelArguments,
+      awaitPromise: false,
+      target: this.#browsingContext.target,
+    });
+
+    function getCallbacksAndRemoteValue(params: Bidi.Script.MessageParameters) {
+      const {data, source} = params;
+      assert(data.type === 'array');
```
Lightning00Blade, line 782: If it fails here wouldn't it stop for node as and unhandled error?

Lightning00Blade (changes requested):

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: This is a reasonable use case. We use the types here only, so it wouldn't be reasonable to abstract it to the top of the class where onlookers would question where it is used. Placing them here makes it obvious.

packages/puppeteer-core/src/common/bidi/Realm.ts
jrandolf: If there is no frame, we can't construct an element handle to use element methods. In theory the context is sufficient, so this all shouldn't matter, but we would need to refactor the code to make this work.

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: That's okay, because one gets overwritten by the other and the other never gets used after. What we should do is detect when these get cleaned up, but there is no reasonable way at the moment.

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: This should never happen.

jrandolf: It's a permanent failure if it does.

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: They are not executed on existing contexts.

packages/puppeteer-core/src/common/bidi/Page.ts
Lightning00Blade: I disagree as this way we bloat the function (implementation) with types that most (if not all) IDE will show you good information about. That way you can focus on the code part rather then the type part.

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: I'll write a TODO.

packages/puppeteer-core/src/common/bidi/Realm.ts
Lightning00Blade: If you return a `result.type === 'node' || result.type === 'window'` you will always have BrowsingContext related to them so this change is not good. Also I don't see you using this why change it now?

packages/puppeteer-core/src/common/bidi/Page.ts
Lightning00Blade: Then something is not working as expected https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-addScriptToEvaluateOnNewDocument, Have you tried Firefox implementation?

packages/puppeteer-core/src/common/bidi/Page.ts
Lightning00Blade: I will think about this one does not feel right to fail the whole process. Do we have other places that we do this, I don't think we do for events its usually it's command that the user may expect to fail. Not a random event listener. 
Can you test what error is produced? Maybe we should at give the error `exposeFunction with ${name} failed`?

packages/puppeteer-core/src/common/bidi/Page.ts
Lightning00Blade: Write a `TODO` as this is not expected behavior compare to the CDP implementation.

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -549,6 +550,228 @@ export class Page extends PageBase {
     }
   }
 
+  #exposedFunctionChannels = (function* () {
```
Lightning00Blade: Nit:
```suggestion
  #getExposedFunctionChannels = (function* () {
```

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -549,6 +550,228 @@ export class Page extends PageBase {
...
+    for (;;) {
+      let id = Math.random();
```
Lightning00Blade: Let needs to be outside of the loop no? else every time is overwritten with a new value

Lightning00Blade: Also isn't is simpler to do `while(true)`? As its easier to read.

jrandolf: Indeed!

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: This isn't really a function :/

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -557,6 +558,228 @@ export class Page extends PageBase {
     }
   }
 
+  #exposedFunctionChannels = (function* () {
+    let id = Math.random();
```
OrKoN: why Math.random and not 0?

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: `addPreloadScript` for Webdriver BiDi is specifically-defined to not run on existing realms. See https://github.com/GoogleChromeLabs/chromium-bidi/blob/main/src/bidiMapper/domains/script/ScriptProcessor.ts#L62C28-L62C28 and https://w3c.github.io/webdriver-bidi/#command-script-addPreloadScript

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: Yes we do this in the input action handler for example.

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: In case there are other clients connected to the server. The channel names are global to the process, so we can't start from 0.

packages/puppeteer-core/src/common/bidi/Page.ts
OrKoN: let's move the implementation in a standalone class?

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: But I guess this doesn't matter since the binding is global :/. I'll just keep the id as the name.

packages/puppeteer-core/src/common/bidi/Page.ts
OrKoN: `class ExposedFunction { constructor(connection, browsingContext); }`

OrKoN: as exposing functions should be possible for any contexts theoretically?

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -557,6 +558,228 @@ export class Page extends PageBase {
...
+
+    const functionDeclaration = stringifyFunction(
+      interpolateFunction(
+        (
+          sendArgs: SendArgsChannel,
+          sendResolve: SendResolveChannel,
+          sendReject: SendRejectChannel
+        ) => {
+          let id = 0;
+          Object.assign(window, {
```
OrKoN: globalThis?

packages/puppeteer-core/src/common/bidi/Realm.ts
jrandolf: Obsolete

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: I think leaving it as is would make more sense. This code isn't used anywhere else so the gain of using a class is moot.

OrKoN: the current change is lengthy and deeply nested with the use of inner types and inner callbacks so moving it to a separate class would allow binding the callbacks to the class instance, simplifying the callback subscription management and moving the types to the top-level provided that the class is placed in its own file (reducing nesting).

jrandolf:
Blocked for now: We need a Chromium BiDi release.

packages/puppeteer-core/src/common/bidi/Page.ts
jrandolf: Do you mind if I add a TODO? I want to also implement an `unexposeFunction` for BiDi to implement cleanup, so at that time I think it makes sense to refactor considering I won't know the granularity to refactor expose function before then.

jrandolf: Alternatively, we can also start implementing `unexposeFunction` here and create another PR that implements it on the page, however I'd rather scope this PR to just `exposeFunction`.

OrKoN: sgtm to add a todo, but let's make sure tests pass before landing.

jrandolf:
Closing in favor of https://github.com/puppeteer/puppeteer/pull/10860.

PR #10702, opened 2023-08-07
chore: implement debuggable callbacks

jrandolf-google:
This PR adds `debugError` to connection callbacks to provide better feedback on incomplete callbacks

### DIFF
--- a/packages/puppeteer-core/src/common/bidi/Connection.ts
+++ b/packages/puppeteer-core/src/common/bidi/Connection.ts
@@ -160,15 +160,24 @@ export class Connection extends EventEmitter {
     method: T,
     params: Commands[T]['params']
   ): Promise<{result: Commands[T]['returnType']}> {
-    return this.#callbacks.create(method, this.#timeout, id => {
-      const stringifiedMessage = JSON.stringify({
-        id,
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/common/bidi/Connection.ts
```diff
@@ -160,15 +160,24 @@ export class Connection extends EventEmitter {
...
+        this.#transport.send(stringifiedMessage);
+      },
+      {
         method,
         params,
-      } as Bidi.Command);
-      debugProtocolSend(stringifiedMessage);
-      this.#transport.send(stringifiedMessage);
-    }) as Promise<{result: Commands[T]['returnType']}>;
+        stack: new Error().stack,
```
OrKoN, line 178: there is already an error in `Callback.#error`. Do we need to create a stack here too?

jrandolf: Creating the stack here makes more sense since the callback is triggered later.

OrKoN: I see that this happens at the same time as the construction of the callback (the error is also created in the constructor)

PR #10717, opened 2023-08-09
refactor: getInstalledBrowsers is not async

kblok:

This is a refactor.  `Cache.getInstalledBrowsers` is not async.

Not needed

Yes.


OrKoN:
Thanks for the PR, but we would like to keep it async for the calling code as we might switch to the async API in the future to prevent the main loop blocking in the server environments.

PR #10737, opened 2023-08-16
fix: locator.fill should work for textareas

### LINKED ISSUE #10727: [Bug]: can't fill textarea input with Locator

```TypeScript
(async () => {
    const browser = await puppeteer.launch({ headless: 'new' });
    page.goto('https://html.com/tags/textarea/');
    try {
        await page.locator('textarea').fill('oklm');
    } catch (e) {
        console.log(e); // Error: Element cannot be filled out.
    }
})();
```

### Error string

Element cannot be filled out

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

Hello, i'm working with puppetter/replay who use the Locator API

When i have a `textarea` to fill with a string it doesn't work. I investigated in the code and I realized that the problem came from the lib `puppeter-core` : 
in file /packages/puppeteer-core/src/api/locators/Locator.ts ( line 479 ) :
```js
 if (el instanceof HTMLInputElement) {
              if (
                new Set([
                  'textarea',
                  'text',
                  'url',
                  'tel',
                  'search',
                  'password',
                  'number',
                  'email',
                ]).has(el.type)
              ) {
                return 'typeable-input';
              } else {
                return 'other-input';
              }
}
```

The element `textarea` is not recognized as an `HTMLInputElement` but as a `HTMLTextAreaElement` which makes the function return  `unknow` type and the the continuation of the code  throw the error `Element cannot be filled out`.

Is it a good idea to add `HTMLTextAreaElement`  as check in the function to fix this problem ?
```js
 if (el instanceof HTMLInputElement || el instanceof HTMLTextAreaElement) {...
```

### Expectation

fill textarea with Locator

### Reality

error `Element cannot be filled out` is throw 

_No response_

### Puppeteer version

21.0.3

### Node version

16.20.0

### Package manager

yarn

### Package manager version

1.19.2

### Operating system

macOS

OrKoN:
Closes #10727

### DIFF
--- a/test/src/locator.spec.ts
+++ b/test/src/locator.spec.ts
@@ -375,7 +375,7 @@ describe('Locator', function () {
       const {page} = await getTestState();
 
       await page.setContent(`
-        <textarea>
+        <textarea></textarea>
       `);
       let filled = false;
       await page
--- a/test/src/locator.spec.ts
+++ b/test/src/locator.spec.ts
@@ -370,7 +370,28 @@ describe('Locator', function () {
     });
   });
 
-  describe('Locator.change', function () {
+  describe('Locator.fill', function () {
+    it('should work for textarea', async () => {
+      const {page} = await getTestState();
+
+      await page.setContent(`
+        <textarea>
+      `);
+      let filled = false;
+      await page
+        .locator('textarea')
+        .on(LocatorEmittedEvents.Action, () => {
+          filled = true;
+        })
+        .fill('test');
+      expect(
+        await page.evaluate(() => {
+          return document.querySelector('textarea')?.value === 'test';
+        })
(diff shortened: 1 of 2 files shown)


test/src/locator.spec.ts
```diff
@@ -367,7 +367,28 @@ describe('Locator', function () {
     });
   });
 
-  describe('Locator.change', function () {
+  describe('Locator.fill', function () {
+    it('should work for textarea', async () => {
+      const {page} = await getTestState();
+
+      await page.setContent(`
+        <textarea>
```
Lightning00Blade:
```suggestion
        <textarea></textarea>
```

Lightning00Blade (approved):

PR #10742, opened 2023-08-16
chore: add script to update `versions.ts`

mathiasbynens:
Since the “latest” Firefox downloads aren’t permanent, the Firefox Nightly version must be updated regularly or else the @puppeteer/browsers tests start failing.

As a next step we can set up a GitHub Action to automatically run this daily or so.

### DIFF
--- a/packages/browsers/tools/updateVersions.mjs
+++ b/packages/browsers/tools/updateVersions.mjs
@@ -1,0 +1,43 @@
+/**
+ * Copyright 2023 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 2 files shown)


packages/browsers/tools/updateVersions.mjs
```diff
@@ -0,0 +1,43 @@
...
+  );
+  const html = await response.text();
+  const re = /firefox-(.*)\.en-US\.langpack\.xpi">/;
+  const match = re.exec(html)[1];
+  return match;
+};
+
+const patch = (input, version) => {
+  const output = input.replace(/testFirefoxBuildId = '([^']+)';/, match => {
+    return `testFirefoxBuildId = '${version}';`;
```
mathiasbynens, line 34: We could use `JSON.stringify` here to handle any unexpected need for escaping, but then we’d have to `npm run format` after running this script just to match the style guide.

OrKoN (approved):

PR #10773, opened 2023-08-22
chore: implement `boxModel`, `contentFrame`, and nested frame `boundingBox`

jrandolf-google:


packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -45,9 +46,11 @@ const applyOffsetsToQuad = (
   offsetX: number,
   offsetY: number
 ) => {
+  assert(quad.length === 4);
```
Lightning00Blade, line 47: Provide a correct error message, so it's easier to find if it fails.

packages/puppeteer-core/src/common/bidi/Frame.ts
```diff
@@ -103,6 +109,31 @@ export class Frame extends BaseFrame {
     return this.#page.frame(this._parentId ?? '');
   }
 
+  /**
+   * Gets the frame element associated with this frame. If the frame is
+   * top-level, then this returns nothing.
+   */
+  async frameElement(): Promise<HandleFor<HTMLIFrameElement> | undefined> {
```
Lightning00Blade: Let's discuss if we want to use `undefined` or `null`. Both are OK but would prefer all our API's to be consistent.

test/TestExpectations.json
```diff
@@ -219,7 +219,7 @@
     "testIdPattern": "[page.spec] Page Page.Events.PageError *",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["webDriverBiDi"],
-    "expectations": ["PASS", "TIMEOUT"]
+    "expectations": ["FAIL", "PASS", "TIMEOUT"]
```
Lightning00Blade: Why was this changed? If its single test we should change only it.

test/TestExpectations.json
```diff
@@ -3962,21 +4172,33 @@
...
     "parameters": ["firefox", "webDriverBiDi"],
-    "expectations": ["PASS"]
+    "expectations": ["FAIL"]
   },
   {
-    "testIdPattern": "[accessibility.spec] Accessibility filtering children of leaf nodes rich text editable fields should have children",
+    "testIdPattern": "[waittask.spec] waittask specs Frame.waitForXPath should throw when frame is detached",
     "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["chrome", "webDriverBiDi"],
+    "parameters": ["firefpx", "webDriverBiDi"],
```
Lightning00Blade: This test spec for firefox already exits a line above please remove. (Most likely due to the typo)

packages/puppeteer-core/src/common/bidi/Frame.ts
```diff
@@ -103,6 +109,31 @@ export class Frame extends BaseFrame {
...
+    const parentFrame = this.parentFrame();
+    if (!parentFrame) {
+      return;
+    }
+    const list = await parentFrame.evaluateHandle(() => {
+      return document.querySelectorAll('iframe');
+    });
+    for await (const iframe of transposeIterableHandle(list)) {
+      const handle = await iframe.evaluateHandle(iframe => {
```
Lightning00Blade: Window handle is not disposed.

packages/puppeteer-core/src/common/ElementHandle.ts
jrandolf: This will never fail. If it fails, this is a catastrophic failure meaning there is a problem elsewhere. A message wouldn't help us.

packages/puppeteer-core/src/common/bidi/Frame.ts
jrandolf: This is internal, so it shouldn't matter at this stage. We can change it if we make this public.

packages/puppeteer-core/src/common/bidi/ElementHandle.ts
```diff
@@ -104,10 +125,185 @@ export class ElementHandle<
...
+        frame = frame?.parentFrame();
+      } finally {
+        void element.dispose().catch(debugError);
+      }
+    }
     return box;
   }
 
+  override async boxModel(): Promise<BoxModel | null> {
+    const model = await this.evaluate(element => {
```
OrKoN: I think all of this should go via the isolatedRealm?

jrandolf: Indeed it should. I've added a TODO here and wrote an issue. There are several areas where this needs to be done.

jrandolf: Done.

test/src/elementhandle.spec.ts
```diff
@@ -26,7 +26,7 @@ import {
 } from './mocha-utils.js';
 import {attachFrame} from './utils.js';
 
-describe('ElementHandle specs', function () {
+describe.only('ElementHandle specs', function () {
```
Lightning00Blade:
```suggestion
describe('ElementHandle specs', function () {
```

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -620,9 +629,10 @@ export class ElementHandle<
   }
 
   /**
-   * Resolves to the content frame for element handles referencing
-   * iframe nodes, or null otherwise
+   * Resolves the frame associated with the element.
    */
+  async contentFrame(this: ElementHandle<HTMLIFrameElement>): Promise<Frame>;
+  async contentFrame(): Promise<Frame | null>;
```
Lightning00Blade, line 635: Why is this override needed?

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -885,7 +895,63 @@ export class ElementHandle<
...
+        height: rect.height,
+      };
+    }, adoptedThis);
+    void adoptedThis.dispose().catch(debugError);
+    if (!box) {
+      return null;
+    }
+    let frame: Frame | null | undefined = this.frame;
+    let element: HandleFor<HTMLIFrameElement> | null | undefined;
+    while ((element = await frame?.frameElement())) {
```
OrKoN, line 932: looks like this frame loop adding window offsets is duplicated? let's extract it into a helper

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -885,7 +895,63 @@ export class ElementHandle<
    * or `null` if the element is not visible.
    */
   async boundingBox(): Promise<BoundingBox | null> {
-    throw new Error('Not implemented');
+    const adoptedThis = await this.frame.isolatedRealm().adoptHandle(this);
+    const box = await this.frame.isolatedRealm().evaluate(element => {
```
Lightning00Blade: Can this be simplified to
```suggestion
    const box = await adoptedThis.evaluate(element => {
```
?

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -897,7 +963,155 @@ export class ElementHandle<
...
+        width: rect.width,
+        height: rect.height,
+      };
+    }, adoptedThis);
+    void adoptedThis.dispose().catch(debugError);
+    if (!model) {
+      return null;
+    }
+    let frame: Frame | null | undefined = this.frame;
+    let element: HandleFor<HTMLIFrameElement> | null | undefined;
```
Lightning00Blade: Remove undefined init as null or use do {} while. Same for above frame should be Frame or null, never undefined.

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -897,7 +963,155 @@ export class ElementHandle<
...
+    while ((element = await frame?.frameElement())) {
+      try {
+        element = await element.frame.isolatedRealm().transferHandle(element);
+        const parentBox = await element.frame
```
Lightning00Blade: Same here we should be able to do `element.evaluate`.
Note that the init FrameElement is never disposed.

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -897,7 +963,155 @@ export class ElementHandle<
...
+          bottom: parseInt(style.borderBottom, 10),
+        },
+      };
+      const border: Quad = [
+        {x: rect.left, y: rect.top},
+        {x: rect.left + rect.width, y: rect.top},
+        {x: rect.left + rect.width, y: rect.top + rect.bottom},
+        {x: rect.left, y: rect.top + rect.bottom},
+      ];
+      const padding: Quad = [
```
Lightning00Blade: Is is possible to simplify this with a loop like bellow 
```
for (const attribute of [
          'content',
          'padding',
          'border',
          'margin',
        ] as const) 
        ```

File: packages/puppeteer-core/src/api/ElementHandle.ts (hunk 8 above)
Comment (jrandolf) [reply packages/puppeteer-core/src/api/ElementHandle.ts:635]: It's just a helper. The new override will always be true, so nullity checks are not needed.

File: packages/puppeteer-core/src/api/ElementHandle.ts (hunk 12 above)
Comment (jrandolf) [reply packages/puppeteer-core/src/api/ElementHandle.ts]: `transferHandle` disposes the element handle.

File: packages/puppeteer-core/src/api/ElementHandle.ts (hunk 11 above)
Comment (jrandolf) [reply packages/puppeteer-core/src/api/ElementHandle.ts]: It can be `undefined` because of nullish collescing (`?.`).

File: packages/puppeteer-core/src/api/ElementHandle.ts (hunk 13 above)
Comment (jrandolf) [reply packages/puppeteer-core/src/api/ElementHandle.ts]: No, it's not the same.

File: packages/puppeteer-core/src/api/ElementHandle.ts [hunk 14]
```diff
@@ -885,7 +895,72 @@ export class ElementHandle<
...
+    return box;
+  }
+
+  async #getTopLeftCornerOfFrame() {
+    const point = {x: 0, y: 0};
+    let frame: Frame | null | undefined = this.frame;
+    let element: HandleFor<HTMLIFrameElement> | null | undefined;
+    while ((element = await frame?.frameElement())) {
+      try {
+        element = await element.frame.isolatedRealm().transferHandle(element);
```
Lightning00Blade, line 934: We need to dispose the original FrameElement first, we can't override it.

jrandolf: `transferHandle` takes care of disposing the element.

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -885,7 +895,72 @@ export class ElementHandle<
...
+        width: rect.width,
+        height: rect.height,
+      };
+    });
+    void adoptedThis.dispose().catch(debugError);
+    if (!box) {
+      return null;
+    }
+    const offset = await this.#getTopLeftCornerOfFrame();
+    if (!offset) {
```
Lightning00Blade, line 920: This is not needed `await this.#getTopLeftCornerOfFrame();` always return object.

jrandolf: We want to return null so we know that the frame element is hidden (therefore the entire bounding box is invisible).

OrKoN (approved):

Lightning00Blade (approved):

PR #10774, opened 2023-08-22
chore: use abstract for implemented methods

jrandolf-google:
In order to know what methods have been implemented on both BiDi and CDP, this PR removes the base implementations of finished implementations so that TypeScript enforces the implementations to exist. With the base implementation, TypeScript does not throw if one of the implementations disappear.

This PR also removes CDP-specific methods from `ElementHandle` and `JSHandle`.


OrKoN (changes requested):
I don’t think we should have ‘abstract’ in the docs as it gives impression that one needs to extend the classes.

Lightning00Blade:
I agree with Alex on this one. Seems to change only the docs and not help with implementation.

jrandolf:
> I agree with Alex on this one. Seems to change only the docs and not help with implementation.

This PR provides type-safety as explained above. Without it, it's harder to know when an implementation is finished since the base implementation will always exist.

OrKoN (approved):
LGTM based on the offline discussion

PR #10775, opened 2023-08-22
chore: implement `ElementHandle.prototype.clickablePoint`

jrandolf-google:


packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -626,9 +626,21 @@ export abstract class ElementHandle<
   /**
    * Returns the middle point within an element unless a specific offset is provided.
    */
-  async clickablePoint(offset?: Offset): Promise<Point>;
-  async clickablePoint(): Promise<Point> {
-    throw new Error('Not implemented');
+  async clickablePoint(offset?: Offset): Promise<Point> {
+    const model = await this.boundingBox();
```
OrKoN: I do not think it's equivalent to the previous implementation. In particular, it would have a different behavior for elements with multiple bounding boxes. Could you please elaborate what where the features of the previous implementation and how the new implementation covers them?

OrKoN (changes requested):

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -861,6 +873,35 @@ export abstract class ElementHandle<
     options?: Readonly<KeyPressOptions>
   ): Promise<void>;
 
+  async #clickableBox(): Promise<BoundingBox | null> {
+    const adoptedThis = await this.frame.isolatedRealm().adoptHandle(this);
+    const box = await adoptedThis.evaluate(element => {
+      if (!(element instanceof Element)) {
+        return null;
+      }
+      return element.getClientRects()[0]?.toJSON() as DOMRect;
```
OrKoN: the previous version filtered out rects that are "too small to click". I guess we don't have a test for this and I am not sure we can add one but should we perhaps keep the filter?

OrKoN: also we checked that rect is in the viewport. Do we not need that?

jrandolf: I still filter out rects using `width === 0 || height === 0`, but I realized that an element can have float values for their width and height, so I've replace this with `width < 1 || height < 1` to ensure we at least have 1 pixel.

jrandolf: FWIU, the check is only kind of useful in the case where:

 1. the element is not in an iframe,
 2. overflow is hidden so scrolling to the element is not possible
 
I suspect that (2) is not very common.

OrKoN: I don't see where the filtering happens. I see that the first `element.getClientRects()[0]` rect is taken unconditionally.

OrKoN: What happens if we click on a portion of the element that is outside of the viewport? Imagine we have a long paragraph of text in one element that starts before the viewport and ends after. What happens if we click on the rect for the first line that is outside of the viewport?

jrandolf: Done.

jrandolf: Ah, I misunderstood what you meant. Done.

OrKoN (approved):

PR #10778, opened 2023-08-23
chore: implement `waitForFrame` and use `clickablePoint` for ElementHandle operations

jrandolf-google:
This PR implements `waitForFrame` and replaces the BiDi input methods on ElementHandle with a common implementation.

By doing the latter, we also bypass a subtle issue caused by using `remoteValue`s in input methods from non-top-level elements; we would need to pass in the correct context into `input.performActions` to ensure BiDi can correctly compute the element origin for non-top-level elements. We can do this, however since this ends up behaving the same as the common implementation using `clickablePoint`, this PR saves code using the common implementation.

Additionally, this PR rounds coordinates passed to input functions since BiDi requires integers.


packages/puppeteer-core/src/common/bidi/Input.ts
```diff
@@ -484,8 +484,8 @@ export class Mouse extends BaseMouse {
     options: Readonly<BidiMouseMoveOptions> = {}
   ): Promise<void> {
     this.#lastMovePoint = {
```
Lightning00Blade, line 487: Nit: Can you leave a comment on the variable saying we need this to be integer and a ref to BiDi Spec?

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -856,10 +895,42 @@ export abstract class ElementHandle<
...
+      return null;
+    }
+    if (box.width === 0 || box.height === 0) {
+      return null;
+    }
+    const offset = await this.#getTopLeftCornerOfFrame();
+    if (!offset) {
+      return null;
+    }
+    box.x += offset.x;
```
Lightning00Blade: Nit: if feels like that should be done in the return object directly.

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -876,12 +947,7 @@ export abstract class ElementHandle<
         return null;
       }
       const rect = element.getBoundingClientRect();
-      return {
-        x: rect.left,
-        y: rect.top,
-        width: rect.width,
-        height: rect.height,
-      };
+      return rect.toJSON() as DOMRect;
```
Lightning00Blade: What benefit does this change do? It feels like we are sending more data that we need and then rimming it on Puppeteer side.

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1745,30 +1758,68 @@ export class Page extends EventEmitter {
...
+              return isMatch;
+            }),
+            map(() => {
+              return frame;
+            })
+          );
+        }),
+        first(),
+        raceWith(
+          timer(ms).pipe(
```
OrKoN: not sure if the value of ms = '0' works the same way

docs/api/puppeteer.page.waitforframe.md
```diff
@@ -4,38 +4,30 @@ sidebar_label: Page.waitForFrame
...
 
 Promise&lt;[Frame](./puppeteer.frame.md)&gt;
 
-Promise which resolves to the matched frame.
-
-## Remarks
-
-Optional Parameter have:
-
-- `timeout`: Maximum wait time in milliseconds, defaults to `30` seconds, pass `0` to disable the timeout. The default value can be changed by using the [Page.setDefaultTimeout()](./puppeteer.page.setdefaulttimeout.md) method.
```
OrKoN, line 37: let's keep this remark somewhere

OrKoN:
One firefox-bidi test is timing out instead of failing https://github.com/puppeteer/puppeteer/actions/runs/5963885847/job/16178069567?pr=10778

docs/api/puppeteer.page.waitforframe.md
jrandolf: It's remarked inside the options object now. See `WaitTimeoutOptions`

jrandolf:
> One firefox-bidi test is timing out instead of failing https://github.com/puppeteer/puppeteer/actions/runs/5963885847/job/16178069567?pr=10778

The test is failing because of https://github.com/web-platform-tests/wpt/pull/41601.

OrKoN (approved):

PR #10781, opened 2023-08-24
chore: update dependencies

jrandolf-google:


Lightning00Blade (commented):
CI failing, please take a look, most likely we will need to update our rules

.eslintrc.js
```diff
@@ -214,6 +214,19 @@ module.exports = {
         ],
         '@typescript-eslint/prefer-ts-expect-error': 'error',
       },
+      overrides: [
+        {
+          files: [
+            'packages/puppeteer-core/src/**/*.test.ts',
```
Lightning00Blade, line 220: Not the best as we don't await the functions in the test themselves but no alternative so happy to merge.

Lightning00Blade (approved):

.eslintrc.js
jrandolf: Yea, in theory we could change it, but mocha tests would look different from node tests.

PR #10784, opened 2023-08-25
chore: enable more BiDi tests

Lightning00Blade:


test/TestExpectations.json
```diff
@@ -4372,5 +4366,101 @@
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["cdp", "chrome", "headless"],
     "expectations": ["FAIL", "PASS"]
+  },
+  {
+    "testIdPattern": "[frame.spec] Frame specs Frame Management should send events when frames are manipulated dynamically",
```
Lightning00Blade: Intentionally not sorted so we can remove the  `chrome` if after Firefox passes as well.

packages/puppeteer-core/src/common/bidi/BrowsingContext.ts
```diff
@@ -101,7 +101,7 @@ export class CDPSessionWrapper extends CDPSession {
 
   override async detach(): Promise<void> {
     cdpSessions.delete(this.id());
-    if (this.#context.supportsCDP()) {
+    if (!this.#detached && this.#context.supportsCDP()) {
```
Lightning00Blade, line 104: Causes issue as the target may have already been detached.

packages/puppeteer-core/src/common/bidi/Sandbox.ts
```diff
@@ -68,6 +70,19 @@ export class Sandbox implements RealmBase {
   constructor(context: Realm, timeoutSettings: TimeoutSettings) {
     this.#realm = context;
     this.#timeoutSettings = timeoutSettings;
+
+    // TODO: we should track the correct realm
+    // Temporary fix
+    // This only works for the main context realm
+    this.#realm.on(Bidi.ChromiumBidi.Script.EventNames.RealmCreated, () => {
```
Lightning00Blade: There should be a better way to handle this, still looking over it, as it was part of an old stash,

packages/puppeteer-core/src/api/HTTPResponse.ts
```diff
@@ -157,7 +157,7 @@ export class HTTPResponse {
    * True if the response was served by a service worker.
    */
   fromServiceWorker(): boolean {
```
OrKoN: should we mark it as abstract to make sure we don't accidentially start returning false for everything?

Lightning00Blade: Happy to refactor this in another PR as this class is not abstract. Only moving it into the BiDi class for now.

packages/puppeteer-core/src/common/bidi/Sandbox.ts
Lightning00Blade: Left a comment, and we do the same thing as per PuppeteerUtils.

packages/puppeteer-core/src/common/Frame.ts
```diff
@@ -63,7 +63,6 @@ export class Frame extends BaseFrame {
   _frameManager: FrameManager;
   override _id: string;
   _loaderId = '';
-  override _hasStartedLoading = false;
```
OrKoN, line 66: is it not used anymore?

OrKoN (approved):

packages/puppeteer-core/src/common/Frame.ts
Lightning00Blade: This leave on the api Frame class and as a `"protected"` value is shared with child classes.

PR #10790, opened 2023-08-28
chore: allow enabling the tab target for testing

OrKoN:
This PR adds a flag that allows turning on the tab target mode via an internal test flag or via the internal browser creation API. It also adds a new test suite for this configuration (only Linux + new headless).

### DIFF
--- a/packages/puppeteer-core/src/common/ChromeTargetManager.ts
+++ b/packages/puppeteer-core/src/common/ChromeTargetManager.ts
@@ -99,16 +99,21 @@ export class ChromeTargetManager extends EventEmitter implements TargetManager {
   #waitForInitiallyDiscoveredTargets = true;
 
   // TODO: remove the flag once the testing/rollout is done.
-  #tabMode = false;
-  #discoveryFilter = this.#tabMode ? [{}] : [{type: 'tab', exclude: true}, {}];
+  #tabMode: boolean;
+  #discoveryFilter: Protocol.Target.FilterEntry[];
 
   constructor(
     connection: Connection,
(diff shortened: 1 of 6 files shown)


packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -178,7 +179,8 @@ export class ChromeLauncher extends ProductLauncher {
       '--disable-dev-shm-usage',
       '--disable-extensions',
       // AcceptCHFrame disabled because of crbug.com/1348106.
-      '--disable-features=Translate,BackForwardCache,AcceptCHFrame,MediaRouter,OptimizationHints,Prerender2',
+      '--disable-features=Translate,BackForwardCache,AcceptCHFrame,MediaRouter,OptimizationHints',
+      ...(!USE_TAB_TARGET ? ['--disable-features=Prerender2'] : []),
```
Lightning00Blade: Nit: Prefer dropping the `!` it makes it easier to read.

packages/puppeteer-core/src/common/ChromeTargetManager.ts
```diff
@@ -100,15 +100,20 @@ export class ChromeTargetManager extends EventEmitter implements TargetManager {
 
   // TODO: remove the flag once the testing/rollout is done.
   #tabMode = false;
```
Lightning00Blade: Change to boolean, as we set it in the constructor that way we know where the value is coming from and that is not hard coded.

Lightning00Blade (approved):

Issue #10017, opened 2023-04-13
[Feature]: Support network events, bindings, preload scripts, interception and dialogs for OOPIFs


OrKoN:
### Feature description

Multiple bugs have been filed that report that request events/interception don't work for OOPIFs: https://github.com/puppeteer/puppeteer/issues/9699 https://github.com/puppeteer/puppeteer/issues/9567

This is due to the fact that OOPIF support was added at a later point and the network event/interception code only works with same-process iframes: https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/FrameManager.ts#L166

Historically, the old headless mode always worked with OOPIFs disabled as it is easier for automation. Now that the new headless mode and headful mode are also more used, we need to make request interception/events work for multiple targets.

OrKoN:
cc @jrandolf @Lightning00Blade

OrKoN:
Probably the same root cause, no support for OOPIFs https://github.com/puppeteer/puppeteer/issues/10216

OrKoN:
Same root cause https://github.com/puppeteer/puppeteer/issues/10783

PR #10793, opened 2023-08-28
fix: make network manager multi session

### LINKED ISSUE #9567: [Bug]: Request interception does not capture oopif iframe subresources
### Bug expectation

I expected `await page.setRequestInterception(true);` to also intercept the subresource requests from iframes, even if they are out-of-process. 

[This test](https://github.com/puppeteer/puppeteer/blob/e13e9647fc0d917da94af8851a09ed318fb0e07c/test/src/oopif.spec.ts#L307) implies that that is the intended functionality, but it doesn't actually check for subresource interception.

The attached example to reproduce is inspired by the equivalent [Playwright Test](https://github.com/microsoft/playwright/blob/caec93ef42b4b93185cf63ecc95fb78cadb2fcf1/tests/library/chromium/oopif.spec.ts#L139).

```diff
diff --git a/test/assets/dynamic-oopif.html b/test/assets/dynamic-oopif.html
--- a/test/assets/dynamic-oopif.html
+++ b/test/assets/dynamic-oopif.html

<script>
window.addEventListener('DOMContentLoaded', () => {
  const iframe = document.createElement('iframe');
  const url = new URL(location.href);
  url.hostname = url.hostname === 'localhost' ? '127.0.0.1' : 'localhost';
-   url.pathname = '/oopif.html';
+  url.pathname = '/grid.html';
  iframe.src = url.toString();
  document.body.appendChild(iframe);
}, false);
</script>
```

```diff 
diff --git a/test/src/oopif.spec.ts b/test/src/oopif.spec.ts
--- a/test/src/oopif.spec.ts
+++ b/test/src/oopif.spec.ts

  it('should load oopif iframes with subresources and request interception', async () => {
     const {server} = getTestState();
+    let intercepted = false;
 
     const frame = page.waitForFrame(frame => {
-      return frame.url().endsWith('/oopif.html');
+      return frame.url().endsWith('/grid.html');
     });
     await page.setRequestInterception(true);
     page.on('request', request => {
+      if (request.url().endsWith('/digits/0.png')) {
+        intercepted = true;
+      }
       return request.continue();
     });
     await page.goto(server.PREFIX + '/dynamic-oopif.html');
     await frame;
     expect(oopifs(context).length).toBe(1);
+    expect(page.frames().length).toBe(2);
+    expect(intercepted).toBe(true);
   });
```

If oopif has not been disabled, this modified test now fails with:
```
 1) OOPIF-debug
       OOPIF
         should load oopif iframes with subresources and request interception:
     Error: expect(received).toBe(expected) // Object.is equality

Expected: true
Received: false
      at Context.<anonymous> (/workspaces/puppeteer/test/src/oopif.spec.ts:327:25)
```

While workarounds exist like passing in `--disable-site-isolation-trials`, it seems like a very heavy hammer and perhaps is a bug in how request interception is intended to work.

Thanks for taking a look!

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript
(added inline)
```

### Error string

(added inline)

### Puppeteer configuration

_No response_

### Puppeteer version

19.6.0

### Node version

18.13.0

### Package manager

npm

### Package manager version

8.19.3

### Operating system

Linux

### LINKED ISSUE #9699: [Bug]: Cannot get response from iframe
### Bug expectation

Same as issue: https://github.com/puppeteer/puppeteer/issues/8473, always TimeoutError got

 TimeoutError: Timeout exceeded while waiting for event
      at Timeout.eval (webpack:///./node_modules/puppeteer-core/lib/cjs/puppeteer/common/util.js?:168:28)
      at listOnTimeout (node:internal/timers:557:17)
      at processTimers (node:internal/timers:500:7)

6GL4P7)XC96DVHE 6_PQ~{B

### Bug behavior

- [ ] Flaky
- [ ] PDF

```TypeScript
        try {
            await page.goto("http://www.10010.com/net5/091/###")
            console.log('start check:' + mobile)
            await page.waitForTimeout(1000)

            // const frameXPath = "/html/body/div/div[4]/div/div[2]/iframe";
            // const frame = await page.waitForXPath(frameXPath);
            // console.log('frame.content',await frame.content())
            // const frames = await page.frames()
            // console.log('frames.length='+frames.length)

            // const div = page.waitForSelector('#aria5iipudb7pq8 > div.width100.banner_div1.ariaskiptheme.skipAutoFix > div > div.banner_right_div.ariaskiptheme.skipAutoFix')
            console.log('result=', Array.from(await page.frames()).map(i => i.url()))

            const frameUrl = 'https://upay.10010.com/npfweb/npfcellweb/phone_speedy.html'
            let frame = await (await page.frames()).find(f => f.url() === frameUrl)
            if (!frame) {
                await page.waitForTimeout(1000)
                frame = await (await page.frames()).find(f => f.url() === frameUrl)
            }
            if (!frame) {
                await page.waitForTimeout(1000)
                frame = await (await page.frames()).find(f => f.url() === frameUrl)
            }
            if (!frame) {
                await page.waitForTimeout(1000)
                frame = await (await page.frames()).find(f => f.url() === frameUrl)
            }
            console.log('frame', !!frame)
            if (!frame) {
                throw 'Target not found'
            }
            const inputSelector = "#number"
            const input = await frame.waitForSelector(inputSelector)
            console.log('input=', input)
            await input.focus()
            await input.click({clickCount: 3})
            await input.type(mobile)
            console.log('inputed')
            /**
             *
             * @type {HTTPResponse}
             */
            console.log('waiting for response')
            let response = await page.waitForResponse(response => /\/npfweb\/getArea\/init/i.test(response.url()))
            if (response.ok()) {
                console.log('response got', response)
                const data = await response.text();
                if (!data.endsWith('success')) {
                    throw new Error(response.status() + " " + data)
                }
            } else {
                throw new Error(response.status() + " " + response.statusText())
            }
            response = await page.waitForResponse(response => /\/npfweb\/NpfWeb\/customInfo\/cellInfoQuery/i.test(response.url()))
            if (response.ok()) {
                const data = await response.text();
                console.log('response', data)
                if (data.endsWith('success')) {
                    return data;
                }
                throw new Error(response.status() + " " + data)
            } else {
                throw new Error(response.status() + " " + response.statusText())
            }
        } finally {
            // page.close()
        }
```

### Error string

TimeoutError: Timeout exceeded while waiting for event

### Puppeteer configuration

_No response_

### Puppeteer version

19.4.1

### Node version

v16.18.1

### Package manager

yarn

### Package manager version

1.22.19

### Operating system

Windows

OrKoN:
Issue #10017
Closes #9567 
Closes #9699


packages/puppeteer-core/src/common/NetworkManager.ts
```diff
@@ -68,102 +66,90 @@ export const NetworkManagerEmittedEvents = {
...
-        deferredInitPromise.resolve();
-      })
-      .catch(err => {
-        deferredInitPromise.reject(err);
-      });
-    return this.#deferredInit.valueOrThrow();
+  }
+
+  async #removeClient(client: CDPSession) {
+    this.#clients.delete(client);
```
Lightning00Blade, line 153: Remove event listeners from client?

packages/puppeteer-core/src/common/NetworkManager.ts
```diff
@@ -209,11 +219,22 @@ export class NetworkManager extends EventEmitter {
       ? networkConditions.latency
       : 0;
 
-    await this.#updateNetworkConditions();
+    await this.#applyToAllClients(this.#applyNetworkConditions.bind(this));
+  }
+
+  async #applyToAllClients(fn: (client: CDPSession) => Promise<unknown>) {
+    await Promise.all(
+      Array.from(this.#clients).map(client => {
```
Lightning00Blade:
```suggestion
      this.#clients.values().map(client => {
```

Lightning00Blade (approved):

packages/puppeteer-core/src/common/NetworkManager.ts
OrKoN: Not needed because the client itself is disposed.

packages/puppeteer-core/src/common/NetworkManager.ts
OrKoN: Values does not return an array so has no map.

packages/puppeteer-core/src/common/NetworkManager.ts
Lightning00Blade: Can you point me to the code that does it, I see no clearing of the attached events in the CDPSession class

OrKoN: When nothing references the client, it will be gc’d, its listeners won’t be referenced by anything in turn and will be cleaned up.

OrKoN: I will add explicit event listener tracking in the next commit here.

PR #10798, opened 2023-08-29
chore: implement use-using rule

jrandolf-google:
This is an internal TypeScript plugin that enforces usage of `using` in our codebase.

### DIFF
--- a/tools/eslint/src/use-using.ts
+++ b/tools/eslint/src/use-using.ts
@@ -1,0 +1,89 @@
+/**
+ * Copyright 2023 Google Inc. All rights reserved.
+ *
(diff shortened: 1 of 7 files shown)


Lightning00Blade (approved):

tools/eslint/src/use-using.ts
```diff
@@ -0,0 +1,73 @@
+import {ESLintUtils, TSESTree} from '@typescript-eslint/utils';
```
OrKoN, line 17: missing license headers

OrKoN: here and other files

OrKoN (commented):
Otherwise, LGTM

PR #10803, opened 2023-08-29
chore: implement moveable

jrandolf-google:


packages/puppeteer-core/src/util/decorators.ts
```diff
@@ -0,0 +1,43 @@
+import {Symbol} from '../../third_party/disposablestack/disposablestack.js';
```
OrKoN, line 17: pls check the license headers

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -140,7 +141,7 @@ export interface Point {
  *
  * @public
  */
-
+@moveable
```
OrKoN: so it does not apply to subclasses automatically?

OrKoN: would we need to mark every subclass as movable?

jrandolf: We only need to mark subclasses that we want the `move` function on (i.e. any disposable class). Alternatively we could manually write it in each class, but using a decorator makes more sense here :P

OrKoN: it means that the build time types do not match the runtime types? one can declare the base class implementing movable but actually forget to add decorators to subclasses.

I suggest that implement move in the base JSHandle class given that we don't have plan resource management for other resources so far.

jrandolf: The build time types do match the runtime types. We also wrote `@moveable` on the JSHandle.

jrandolf: Sorry, what I wrote was incorrect (I misunderstood what you wrote). It does apply to subclasses automatically. The reason we do it on ElementHandle though is because the inherited move is not the one we want, but the one coming from `this.handle`, so we use `@moveable` again to just create the generic one on ElementHandle.

jrandolf: In general, as long as you declare it on the base class you are fine.

OrKoN: But we would not need to do that if just declare mode on the base class, right? That looks like a simpler solution. Btw let's add a move test for ElementHandle too?

jrandolf: We would need to sadly. See how `dispose` is implemented in `ElementHandle`.

packages/puppeteer-core/src/api/JSHandle.ts
```diff
@@ -43,9 +49,12 @@ import {ElementHandle} from './ElementHandle.js';
  *
  * @public
  */
+@moveable
 export abstract class JSHandle<T = unknown>
-  implements Disposable, AsyncDisposable
+  implements Disposable, AsyncDisposable, Moveable
 {
+  declare move: () => this;
```
OrKoN, line 56:
```suggestion
  #moved = false;

  [Symbol.dispose](): void {
    if (this.#moved) { this.moved = false; return;} 
    return void this.dispose().catch(debugError);
  }

  [Symbol.asyncDispose](): Promise<void> {
    if (this.#moved) { this.moved = false; return Promise.resolve();} 
    return this.dispose();
  }

  move() {
    this.#moved = true;
    return this;
  }
```

OrKoN: that does not require decorators and makes polymorphism work out of the box it seems

packages/puppeteer-core/src/api/ElementHandle.ts
OrKoN: See my other comment (seems to work)

packages/puppeteer-core/src/api/JSHandle.ts
OrKoN: Symbol.*Dispose don't need to overriden in the subclass with this

OrKoN: also I probably miss the moved=true to true in dispose handles.

OrKoN: (would be good to add a test for multiple moves)

jrandolf: Hmm, I just realized that this works because the `dispose` is applied to the this context. Yea, I think we can remove the element handle implementation. I'll keep the test though.

packages/puppeteer-core/src/api/ElementHandle.ts
jrandolf: Indeed it does work!

OrKoN (approved):

PR #10806, opened 2023-08-29
chore: turn on `rulesdir/use-using`

jrandolf-google:
This PR replaces all instances of `let` and `const` of handles with `using`.

Note that the `IsolatedWorld.document` does not cache anymore. This is temporary and will be fixed in a separate PR.

### DIFF
--- a/test/src/evaluation.spec.ts
+++ b/test/src/evaluation.spec.ts
@@ -367,7 +367,7 @@ describe('Evaluation specs', function () {
       const {page} = await getTestState();
 
       await page.setContent('<section>42</section>');
-      const element = (await page.$('section'))!;
+      using element = (await page.$('section'))!;
       const text = await page.evaluate(e => {
         return e.textContent;
       }, element);
@@ -377,8 +377,9 @@ describe('Evaluation specs', function () {
       const {page} = await getTestState();
 
       await page.setContent('<section>39</section>');
-      const element = (await page.$('section'))!;
+      using element = (await page.$('section'))!;
       expect(element).toBeTruthy();
+      // We want to dispose early.
       await element.dispose();
       let error!: Error;
       await page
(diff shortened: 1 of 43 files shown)


test/src/evaluation.spec.ts
```diff
@@ -377,7 +377,7 @@ describe('Evaluation specs', function () {
       const {page} = await getTestState();
 
       await page.setContent('<section>39</section>');
-      const element = (await page.$('section'))!;
+      using element = (await page.$('section'))!;
       expect(element).toBeTruthy();
       await element.dispose();
```
OrKoN, line 383: let's remove the explicit dispose in this PR too

OrKoN:
Which PR fixes the isolated work issue?

jrandolf:
> Which PR fixes the isolated work issue?

None yet.

OrKoN (approved):
Please file an issue for the document caching

OrKoN (changes requested):
looks like there are some test failures

test/TestExpectations.json
```diff
@@ -2111,6 +2111,12 @@
     "parameters": ["firefox", "webDriverBiDi"],
     "expectations": ["FAIL"]
   },
+  {
+    "testIdPattern": "[elementhandle.spec] ElementHandle specs ElementHandle.isIntersectingViewport should work with svg elements",
```
OrKoN: why does this fail?

jrandolf: Not sure, but from the stack trace, I suspect there is a bug in Firefox with their code related to `DOM.resolveNode` (because that's where the issue was coming from).

OrKoN: I mean why it fails because of this PR? there should be no changes related to that, right?

jrandolf: It changes this test because we use migrate this specific function to isolated world. I can revert this change.

OrKoN: Ah, that is fine then. I didn't see that isIntersectingViewport was also changed.

OrKoN (approved):
LGTM thanks

thiagowfx:
Out of curiosity: what is the motivation of this change? Is it to remove the overhead to think about `let` / `const`?

PR #10807, opened 2023-08-29
chore: refactor Sandbox and IsolatedWorld

jrandolf-google:


packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -0,0 +1,212 @@
+import {TimeoutSettings} from '../common/TimeoutSettings.js';
```
OrKoN, line 17: license header

packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -0,0 +1,228 @@
...
+import {assert} from '../util/assert.js';
+
+import {ClickOptions, ElementHandle} from './ElementHandle.js';
+import {KeyboardTypeOptions} from './Input.js';
+import {JSHandle} from './JSHandle.js';
+
+/**
+ * @internal
+ */
+export abstract class Realm implements Disposable {
```
Lightning00Blade, line 36: I think we should rename this as Realm is `BiDi's` `ExecutionContext`. Given we are trying to move to more spec(y) names I would go for `Sandbox` as it represents that its a Realm but can be isolated or main

jrandolf: I can do this in a separate PR (refactoring in this part of the CL chain will be a headache and it's internal so it should be okay to do).

However, I think `Realm` makes sense here. It's well-defined in JS (see https://tinyurl.com/fvwnyd93 and https://tc39.es/ecma262/#sec-code-realms)

OrKoN (approved):
LGTM thanks

packages/puppeteer-core/src/api/Realm.ts
Lightning00Blade: Our wrapper around ECMA Realm's are `ExecutionContext` and `(BiDi)Realm`. The Sandbox is an abstraction on top of them that uses Realm and helpers (`this.document`) to simplify work.

jrandolf: Sure, but we should really just name everything Realm because that's what javascript defines a sandbox as.

OrKoN: Realm makes sense to me.

PR #10808, opened 2023-08-29
chore: refactor Frame

jrandolf-google:


packages/puppeteer-core/src/common/bidi/Frame.ts
```diff
@@ -275,7 +193,7 @@ export class Frame extends BaseFrame {
     return this.#detached;
   }
 
-  dispose(): void {
+  [Symbol.dispose](): void {
```
Lightning00Blade, line 196: I think this should not be changed as is user facing. I think user facing changes should remain with `dispose()`.

jrandolf: I don't think this is user-facing. Frames can't be disposed by the user.

OrKoN (approved):
LGTM (% needs rebasing and passing tests)

Lightning00Blade (approved):

PR #10809, opened 2023-08-29
chore: implement Disposable/AsyncDisposable on Page and Browser

jrandolf-google:


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -2873,6 +2874,14 @@ export class Page extends EventEmitter {
   waitForDevicePrompt(): Promise<DeviceRequestPrompt> {
     throw new Error('Not implemented');
   }
+
+  [Symbol.dispose](): void {
+    return void this.close().catch(debugError);
```
Lightning00Blade, line 2879: Remove return on Sync dispose.

jrandolf: Hmm, but I like the return ;).

Lightning00Blade: Why are we returning a promise when it should be void then? Code currently does that at run time.

jrandolf: `void promise` returns undefined.

OrKoN (approved):
LGTM thanks

packages/puppeteer-core/src/api/Page.ts
sadym-chromium, line 2879: I'd assume that when the page is disposing, it's already closed, and the only things needed is to clean up the state and remove the listeners. Using `close` here seems inappropriate, as `close` is intended to close the page. I'm sure that re-using should be either:
1. Other way around, and `close` should eventually call `dispose`, or
2. `close` should not call `dispose` at all, as the protocol event listeners will process corresponding event (`browsingContext.contextDestroyed` in vase of BiDi).

PR #10810, opened 2023-08-30
chore: use `bindIsolatedHandle`

jrandolf-google:
`bindIsolatedHandle` is a method decorator that enforces the use of `isolatedRealm`. Any return is transformed to their natural equivalent in the `mainRealm`. `this` returns are also supported.


packages/puppeteer-core/src/util/decorators.ts
```diff
@@ -57,3 +59,26 @@ export function moveable<
...
   return Class;
 }
+
+export function withIsolatedRealm<
+  This extends ElementHandle<Node>,
+  Args extends unknown[],
+  Ret,
+>(
+  target: (this: This, ...args: Args) => Promise<Ret>,
+  _: unknown
```
jrandolf: Note this `_: unknown` is required for TypeScript to understand this function is a decorator (it needs to be a function of two arguments: the target and context [which we don't type here]).

packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -307,6 +289,7 @@ export class CDPElementHandle<
     return imageData;
   }
 
+  @withIsolatedRealm
```
OrKoN: why withIsolatedRealm here?

OrKoN (changes requested):
I am not convinced it is the right approach. The decorated method runs with a different this which makes it hard to read and understand. E.g., any sort of mutations that one might do in the decorated function would be done on the wrong instance. I don't think we should change that semantics.

packages/puppeteer-core/src/util/decorators.ts
```diff
@@ -57,3 +59,26 @@ export function moveable<
   }
   return Class;
 }
+
+export function withIsolatedRealm<
```
OrKoN: let's call it `runOnNewIsolatedHandle` ?

packages/puppeteer-core/src/common/ElementHandle.ts
jrandolf: Mistake!

packages/puppeteer-core/src/util/decorators.ts
OrKoN: `createFreshIsolatedHandle`, `withNewIsolatedHandle`, `withThisAsNewIsolatedHandle` or smth like this

jrandolf: What about `usingIsolatedHandle`?

OrKoN: how can we highlight that a different thisArg will used? I think `usingIsolatedHandle ` does not indicate this either. Perhaps `@bindIsolatedHandle`?

jrandolf: Ah, `bindIsolatedHandle` is perfect!

OrKoN: or `@replacingWithIsolatedHandle`

jrandolf: I think `bindIsolatedHandle` is pretty solid. It does precisely that.

jrandolf: Alternatively, what about just `isolated`? Too simple?

OrKoN: also, does not indicate to me that `this` would be different

OrKoN:
Let's land the rest and then I will review this one.

packages/puppeteer-core/src/util/decorators.ts
jrandolf: Hmm, FWIU, you want the name to reflect the precise expectation. Is the comment not sufficient?

OrKoN: The comment is not sufficient IMO because it's not in visible at the usage site. `bindIsolatedHandle` sgtm

packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -249,6 +229,7 @@ export class CDPElementHandle<
     }
   }
 
+  @ElementHandle.bindIsolatedHandle
   override async screenshot(
```
OrKoN, line 235: so screenshot invokes inner methods that are also decorated which results in several handles being created for each screenshot call. I think this kind of expensive given that each call adds a round-trip. Can we avoid that?

OrKoN: I think the same might apply to other methods. Also, while it does not look like the screenshot needs its own isolated handle the creation of many handles seems wasteful.

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -139,6 +139,37 @@ export interface Point {
...
+    return async function (...args) {
+      using adoptedThis = await this.frame.isolatedRealm().adoptHandle(this);
+      const result = await target.call(adoptedThis, ...args);
+      // If the function returns `adoptedThis`, then we return `this`.
+      if ((result as unknown) === adoptedThis) {
+        return this as unknown as Awaited<Ret>;
+      }
+      // If the function returns a handle, transfer it into the main realm.
+      if (result instanceof JSHandle) {
+        return await this.frame.mainRealm().transferHandle(result);
```
OrKoN: why do we need to transfer it to the main realm? isn't it possible that the handle is requested in the isolated realm originally?

packages/puppeteer-core/src/common/ElementHandle.ts
OrKoN: I guess this is already partially the case in the existing code though.

jrandolf: Yes, this can be expensive. I have to refactor ElementHandle more so it's aware of which sandbox it lives in. Once we do that, we can get rid of these roundtrips. I'm working on it in another branch right now.

OrKoN: we could probably count references to a handle and avoid adopting a real handle if we already have adopted one in the same world, that way we can avoid round trips.

packages/puppeteer-core/src/api/ElementHandle.ts
jrandolf: You're right. I think before we can progress with this PR, I'll need to implement realm awareness in ElementHandle. I'll resolve this comment once it's ready.

OrKoN: Let's do that first and then rebase this PR on the sandbox aware handles?

packages/puppeteer-core/src/api/Environment.ts
```diff
@@ -0,0 +1,11 @@
+import {CDPSession} from '../common/Connection.js';
```
OrKoN: license header

OrKoN:
Would you mind rebasing and regenerating the docs for review?

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -249,6 +289,9 @@ export abstract class ElementHandle<
     return this.handle.dispose();
   }
 
+  /**
+   * @internal
+   */
   override asElement(): ElementHandle<ElementType> {
```
OrKoN, line 313: why is it internal?

jrandolf: It's already on the `JSHandle`.

OrKoN (approved):
LGTM thanks! The tests are failing though, investigate please before landing. I hope it does not require substantial changes.

PR #10812, opened 2023-08-30
feat: expose DevTools as a target

### LINKED ISSUE #9371: [API request] Expose the Chrome DevTools target as a `page`
### Bug description

Filing a new issue instead of reopening https://github.com/puppeteer/puppeteer/issues/4247 where this was first proposed (it was closed in 2019 because it was believed the reporter was the only client trying to use Puppeteer to test a panel added to Chrome DevTools by their Chrome extension, from the comments / recently opened issues it seems that is no longer the case). 

As described in https://github.com/puppeteer/puppeteer/issues/4247#issue-429876229, currently the only way to interact with a panel added to Chrome DevTools by a Chrome extension is to use a hack to set the target type to `page`:

```
devtoolsTarget._targetInfo.type = 'page';
```

This hack is used by the `puppeteer-devtools` library ([code pointer](https://github.com/dequelabs/puppeteer-devtools/blob/e21867534231ca059b285007b01f3c023d8b94be/src/index.ts#L44)) which was recently cited in https://github.com/puppeteer/puppeteer/issues/9327 as a way to test a Chrome Extension devtools panel. 

**Why this doesn't work for my use case:**

I'm using the [Dart language wrapper of Puppeteer](https://github.com/xvrh/puppeteer-dart) to test my Chrome extension (which is also written in Dart). Because Dart is a strongly-typed language that prevents access to `private` members (such as `_targetInfo`), this hack is not possible. I could ask the owner of the Dart Puppeteer implementation to expose this hack in their API, but it seems like a better solution for Puppeteer itself to support this as a real use case and not as a hack. 

From https://github.com/puppeteer/puppeteer/issues/4247#issuecomment-482396308, it sounds like supporting this as an actual use case would be possible. 

Thank you!

### Puppeteer version

N/A

### Node.js version

N/A

### npm version

N/A

macOS

### Configuration file

_No response_

### Relevant log output

_No response_

OrKoN:
While DevTools won't show up in the `Browser.pages()` it would be available via its target.

Closes #9371

### DIFF
--- a/test/src/headful.spec.ts
+++ b/test/src/headful.spec.ts
@@ -340,6 +340,26 @@ const serviceWorkerExtensionPath = path.join(
       ]);
       await browser.close();
     });
+    it('should expose DevTools as a page', async () => {
+      const browser = await launchBrowser(
+        Object.assign({devtools: true}, headfulOptions)
+      );
+      const context = await browser.createIncognitoBrowserContext();
+      const [target] = await Promise.all([
+        browser.waitForTarget((target: {url: () => string | string[]}) => {
+          return target.url().includes('devtools://');
+        }),
+        context.newPage(),
+      ]);
+      const page = await target.page();
+      expect(
+        await page?.evaluate(() => {
+          // @ts-expect-error wrong context.
+          return Boolean(DevToolsAPI);
+        })
+      ).toBe(true);
+      await browser.close();
+    });
   });
 
   describe('Page.bringToFront', function () {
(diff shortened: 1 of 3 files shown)


test/src/headful.spec.ts
```diff
@@ -340,6 +340,26 @@ const serviceWorkerExtensionPath = path.join(
...
       await browser.close();
     });
+    it('should expose DevTools as a page', async () => {
+      const browser = await launchBrowser(
+        Object.assign({devtools: true}, headfulOptions)
+      );
+      const context = await browser.createIncognitoBrowserContext();
+      const [, target] = await Promise.all([
+        context.newPage(),
+        browser.waitForTarget((target: {url: () => string | string[]}) => {
```
Lightning00Blade, line 349: Nit: setup awaiter before executing the command.

Lightning00Blade (approved):

PR #10819, opened 2023-08-30
chore: implement `throwIfDisposed`

jrandolf-google:
If an object is already disposed, the `throwIfDisposed` decorator ensures that methods decorated with it will throw when the parent object is disposed.

### DIFF
--- a/packages/puppeteer-core/src/common/ElementHandle.ts
+++ b/packages/puppeteer-core/src/common/ElementHandle.ts
@@ -66,7 +66,6 @@ export class CDPElementHandle<
     return this.handle.client;
   }
 
-  @throwIfDisposed()
   override remoteObject(): Protocol.Runtime.RemoteObject {
     return this.handle.remoteObject();
   }
@@ -114,6 +113,7 @@ export class CDPElementHandle<
   override async contentFrame(
     this: ElementHandle<HTMLIFrameElement>
   ): Promise<CDPFrame>;
+  @throwIfDisposed()
   override async contentFrame(): Promise<CDPFrame | null> {
     const nodeInfo = await this.client.send('DOM.describeNode', {
       objectId: this.id,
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/common/ElementHandle.ts
```diff
@@ -65,6 +66,7 @@ export class CDPElementHandle<
     return this.handle.client;
   }
 
+  @throwIfDisposed()
   override remoteObject(): Protocol.Runtime.RemoteObject {
```
OrKoN: I think this might not need to throw. It does not make any calls and you might still want to have the remote object for the disposed handle.

OrKoN (approved):

Lightning00Blade (approved):

packages/puppeteer-core/src/common/ElementHandle.ts
jrandolf: But is the remote object still available after it has been disposed?

OrKoN: remote object is the metadata about the object, not the object itself.

OrKoN: by looking at the code I don't think the metadata is deleted and I am not sure if it necessarily should be.

jrandolf: The metadata isn't deleted, but it contains information about the handle such as the objectId which is no longer valid after the handle is disposed.

OrKoN: Screenshot 2023-08-31 at 13 29 40

the rest is valid though. E.g., the serialised value of the disposed object.

packages/puppeteer-core/src/common/types.ts
```diff
@@ -29,6 +29,13 @@ export interface Moveable {
   move(): this;
 }
 
+/**
+ * @internal
+ */
+export interface Disposed {
```
Lightning00Blade, line 35: Can we/should we create a `super` interface that extend `Disposable`,`AsyncDisposable`,`Disposed`, that way we can be more sure of that the decorator would work in most cases?

jrandolf: Since decorators wrap things, using decorators as a means of composition is considered better than inheritance (see https://www.digitalocean.com/community/tutorials/composition-vs-inheritance).

jrandolf: Discussed offline.

PR #10826, opened 2023-08-31
fix: implement `throwIfDetached`

jrandolf-google:
Also deprecates `isDetached` for `get detached` and fixes some docs.


packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -974,6 +1027,7 @@ export abstract class Frame extends EventEmitter {
    *
    * @param milliseconds - the number of milliseconds to wait.
    */
+  @throwIfDetached
```
OrKoN: not needed

OrKoN (approved):

PR #10830, opened 2023-08-31
chore: refactor handles to use realms

jrandolf-google:


packages/puppeteer-core/src/common/bidi/Frame.ts
```diff
@@ -66,15 +66,15 @@ export class BidiFrame extends Frame {
 
     const puppeteerRealm = context.createSandboxRealm(UTILITY_WORLD_NAME);
     this.sandboxes = {
-      [MAIN_SANDBOX]: new Sandbox(context, timeoutSettings),
-      [PUPPETEER_SANDBOX]: new Sandbox(puppeteerRealm, timeoutSettings),
+      [MAIN_SANDBOX]: new Sandbox(this, context, timeoutSettings),
+      [PUPPETEER_SANDBOX]: new Sandbox(this, puppeteerRealm, timeoutSettings),
     };
 
     puppeteerRealm.setFrame(this);
```
Lightning00Blade: I think we can remove this.

packages/puppeteer-core/src/common/bidi/Sandbox.ts
```diff
@@ -54,14 +54,17 @@ export interface SandboxChart {
  */
 export class Sandbox extends Realm {
   #realm: BidiRealm;
+  #frame: BidiFrame;
 
   constructor(
+    frame: BidiFrame,
```
OrKoN, line 60: this should not depend on the frame

OrKoN: or actually nevermind, probably, it can.

OrKoN:
tests are failing

packages/puppeteer-core/src/common/WebWorker.ts
```diff
@@ -84,18 +91,19 @@ export class WebWorker extends EventEmitter {
     super();
     this.#client = client;
     this.#url = url;
+    this.#world = new IsolatedWorld(this, new TimeoutSettings());
```
OrKoN, line 94: do workers have DOM? can we create isolated worlds there?

jrandolf: They don't have DOM, but we only use the execution related portion of isolated world here.

OrKoN: hm, maybe we should keep using the execution context for better readability? I would say it should be `realm` instead of the world here and it's realm().sandbox() should say `main` or `default`. And when we have a secondary world in frames that should be a realm too but with the different sandbox name.

packages/puppeteer-core/src/api/Environment.ts
```diff
@@ -0,0 +1,11 @@
+import {CDPSession} from '../common/Connection.js';
+
+import {Realm} from './Realm.js';
+
+/**
+ * @internal
+ */
+export interface Environment {
```
OrKoN, line 24: `RealmOwner`?

jrandolf: Discussed offline.

packages/puppeteer-core/src/common/WebWorker.ts
jrandolf: Discussed offline.

packages/puppeteer-core/src/common/bidi/Realm.ts
```diff
@@ -131,6 +133,17 @@ export class Realm extends EventEmitter {
         PuppeteerURL.INTERNAL_URL
     );
 
+    const sandbox = (() => {
```
Lightning00Blade: Extract into a helper, IIFE don't look good in code and no reason to have them.

packages/puppeteer-core/src/common/bidi/Realm.ts
```diff
@@ -155,9 +168,7 @@ export class Realm extends EventEmitter {
       responsePromise = this.connection.send('script.callFunction', {
         functionDeclaration,
         arguments: await Promise.all(
-          args.map(arg => {
-            return BidiSerializer.serialize(arg, this as any);
-          })
+          args.map(BidiSerializer.serialize.bind(BidiSerializer, sandbox))
```
Lightning00Blade: The `BidiSerializer` is a static class so would prefer to keep the function rather then bind as you are never using `this`

packages/puppeteer-core/src/common/bidi/Realm.ts
jrandolf: I plan to remove this in another PR. We should just pass in the sandbox to realm, not the sandboxId.

packages/puppeteer-core/src/common/bidi/Frame.ts
jrandolf: Ditto to your other comment.

packages/puppeteer-core/src/common/bidi/Realm.ts
Lightning00Blade: Fair enough.

packages/puppeteer-core/src/common/bidi/Frame.ts
Lightning00Blade: Well this is different as we are just dropping the method all together given we pass it in the constructor. And its only used here atm.

jrandolf: It looks like order matters here. We cannot remove this.

packages/puppeteer-core/src/common/bidi/Frame.ts
```diff
@@ -64,17 +64,18 @@ export class BidiFrame extends Frame {
     this._id = this.#context.id;
     this._parentId = parentId ?? undefined;
 
-    const puppeteerRealm = context.createSandboxRealm(UTILITY_WORLD_NAME);
     this.sandboxes = {
-      [MAIN_SANDBOX]: new Sandbox(context, timeoutSettings),
-      [PUPPETEER_SANDBOX]: new Sandbox(puppeteerRealm, timeoutSettings),
+      [MAIN_SANDBOX]: new Sandbox('', this, context, timeoutSettings),
```
Lightning00Blade: This is incorrect is should be 'null/undefined'. A sandbox with name empty may be different than the main one.

Lightning00Blade (approved):

PR #10832, opened 2023-09-01
chore: enable 'return-await`

jrandolf-google:
With 'using', it's critical that we 'return await' instead of just returning. This ensures callbacks are run at the correct time. It's also more performant to 'return await' which is a plus.


OrKoN (approved):
LGTM % wait for tests to run

PR #10833, opened 2023-09-01
refactor: use generic implementations

jrandolf-google:
This PR upstreams some specific implementations to generic ones. The realm methods are also now DOM-independent.

### DIFF
--- a/packages/puppeteer-core/src/common/Frame.ts
+++ b/packages/puppeteer-core/src/common/Frame.ts
@@ -32,6 +32,7 @@ import {FrameManager} from './FrameManager.js';
 import {IsolatedWorld} from './IsolatedWorld.js';
 import {MAIN_WORLD, PUPPETEER_WORLD} from './IsolatedWorlds.js';
 import {LifecycleWatcher, PuppeteerLifeCycleEvent} from './LifecycleWatcher.js';
+import {setPageContent} from './util.js';
 
 /**
  * We use symbols to prevent external parties listening to these events.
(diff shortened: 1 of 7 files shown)


OrKoN:
Tests are failing. @Lightning00Blade pls review

packages/puppeteer-core/src/common/Frame.ts
```diff
@@ -257,7 +258,27 @@ export class CDPFrame extends Frame {
...
   ): Promise<void> {
-    return await this.isolatedRealm().setContent(html, options);
+    const {
+      waitUntil = ['load'],
+      timeout = this._frameManager.timeoutSettings.navigationTimeout(),
+    } = options;
+
+    await setPageContent(this.isolatedRealm(), html);
+
+    const watcher = new LifecycleWatcher(
```
Lightning00Blade, line 268: Set up watcher before setting the page content. That way we dont miss events

packages/puppeteer-core/src/common/Frame.ts
```diff
@@ -257,7 +258,27 @@ export class CDPFrame extends Frame {
       waitUntil?: PuppeteerLifeCycleEvent | PuppeteerLifeCycleEvent[];
     } = {}
   ): Promise<void> {
-    return await this.isolatedRealm().setContent(html, options);
+    const {
+      waitUntil = ['load'],
+      timeout = this._frameManager.timeoutSettings.navigationTimeout(),
+    } = options;
+
+    await setPageContent(this.isolatedRealm(), html);
```
Lightning00Blade, line 266: Can this now be moved inside a common implementation?

packages/puppeteer-core/src/common/Frame.ts
OrKoN: I think the previous version was exactly the same

packages/puppeteer-core/src/common/Frame.ts
OrKoN: Looks like we can now

OrKoN (approved):

packages/puppeteer-core/src/common/Frame.ts
Lightning00Blade: I don't see how that is relevant, when we should improve it.

PR #10834, opened 2023-09-01
chore: handle RealmCreated properly

jrandolf-google:


packages/puppeteer-core/src/common/bidi/Realm.ts
```diff
@@ -27,43 +27,60 @@ export const getSourceUrlComment = (url: string): string => {
...
   setSandbox(sandbox: Sandbox): void {
     this.#sandbox = sandbox;
-
-    // TODO: Tack correct realm similar to BrowsingContexts
-    this.connection.on(Bidi.ChromiumBidi.Script.EventNames.RealmCreated, () => {
-      void this.#sandbox.taskManager.rerunAll();
-    });
-    // TODO(jrandolf): We should try to find a less brute-force way of doing
-    // this.
     this.connection.on(
```
Lightning00Blade, line 68: Let use the same thing we do in Page where we have map between the two.
We can subscribe to both events here as the second will be No-Op if the Id is undefined. 
Also we need a way to unsubscribe so that the object has no reference from `connection`.

jrandolf: Hmm, but isn't it more precise to subscribe only when the initial subscription is finishes? I don't see value in keeping it as a map here (it's not like we will have more added). It'd be a premature generalization,

Lightning00Blade: It make thing easy to follow. Subscribe on construction or in this case setting sandbox then unsubscribe on dispose. Also setting up event listeners should be very low cost operation.

jrandolf: Subscription is low-cost, but not the effects. It would pollute the listener pool.

Lightning00Blade: Disagree we are doing check to validate thing before doing work. 
Deferring to @OrKoN.

OrKoN: I think we need to unsubscribe here because the subscription is global and it will keep realms alive. Also, the pattern of saving listeners to make sure the same listeners are disposed sounds much preferred to me.

test/src/ariaqueryhandler.spec.ts
```diff
@@ -208,7 +208,7 @@ describe('AriaQueryHandler', () => {
   });
   describe('queryAllArray', () => {
     it('$$eval should handle many elements', async function () {
-      this.timeout(25_000);
+      this.timeout(40_000);
```
Lightning00Blade: Why this change, this will significantly slow BiDi tests, as they will fails.

packages/puppeteer-core/src/common/bidi/Realm.ts
jrandolf: `RealmCreated` and `RealmDestroyed` will always come in pairs, so the clean up is contained within the listeners themselves. I don't think we need to handle the listeners directly if we do it this way.

Lightning00Blade (approved):

PR #10849, opened 2023-09-05
refactor: sync emulation state with secondary clients

OrKoN:
This PR adds a decorate that allows a method to only be called if it has not been called with the same arguments before. That is useful for synchronizing state between multiple CDP sessions.

### DIFF
--- a/packages/puppeteer-core/src/util/decorators.ts
+++ b/packages/puppeteer-core/src/util/decorators.ts
@@ -73,6 +73,14 @@ export function throwIfDisposed<This extends Disposed>(
   };
 }
 
+/**
+ * The decorator only invokes the target if the target has not been invoked with
+ * the same arguments before. The decorated method throws an error if it's
+ * invoked with a different number of elements: if you decorate a method, it
+ * should have the same number of arguments
+ *
+ * @internal
+ */
 export function invokeAtMostOnceForArguments(
   target: (this: unknown, ...args: any[]) => any,
   _: unknown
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/util/decorators.ts
```diff
@@ -72,3 +72,36 @@ export function throwIfDisposed<This extends Disposed>(
...
+): typeof target {
+  const cache = new WeakMap();
+  let cacheDepth = -1;
+  return function (this: unknown, ...args: unknown[]) {
+    if (cacheDepth === -1) {
+      cacheDepth = args.length;
+    }
+    if (cacheDepth !== args.length) {
+      throw new Error(
+        'Memoized method was called with the wrong number of arguments'
```
Lightning00Blade, line 96: What happens if the function has optional paramater `fn(a: object, b: object, c?: object)` and we call `fn({a},{b})` and then `fn({a},{b},{c})` this will throw right? I don't think it should.

Lightning00Blade (approved):

packages/puppeteer-core/src/util/decorators.ts
OrKoN: Yeah, it's the drawback, the number of arguments has to remain the same. I am not sure it's worth supporting dynamic arguments.

Lightning00Blade: Let's leave a comment on the decorator itself and we can merge this.

PR #10860, opened 2023-09-07
chore: implement `exposeFunction`

jrandolf-google:


packages/puppeteer-core/src/common/bidi/Frame.ts
```diff
@@ -207,4 +211,229 @@ export class BidiFrame extends Frame {
...
+    ) => void;
+
+    interface RemotePromiseCallbacks {
+      resolve: Deferred<Bidi.Script.RemoteValue>;
+      reject: Deferred<Bidi.Script.RemoteValue>;
+    }
+
+    const bindingMaps = new Map<string, Map<number, RemotePromiseCallbacks>>();
+
+    const handleArgumentsMessage = async (
```
OrKoN: let's extract this code into a class and make these inner methods and types members of the class.

Lightning00Blade: +1 for this change.

test/TestExpectations.json
```diff
@@ -1367,6 +1373,66 @@
     "parameters": ["webDriverBiDi"],
     "expectations": ["PASS"]
   },
+  {
+    "testIdPattern": "[page.spec] Page Page.exposeFunction should await returned promise",
```
Lightning00Blade: Please use
`[page.spec] Page Page.exposeFunction *` as that way newly added test will pass.

test/TestExpectations.json
```diff
@@ -2099,6 +2105,12 @@
     "parameters": ["firefox", "webDriverBiDi"],
     "expectations": ["FAIL"]
   },
+  {
+    "testIdPattern": "[elementhandle.spec] ElementHandle specs ElementHandle.click should return Point data",
```
OrKoN, line 2109: what's the reason for the failure?

test/TestExpectations.json
```diff
@@ -3731,6 +3743,12 @@
     "parameters": ["cdp", "firefox"],
     "expectations": ["SKIP"]
   },
+  {
+    "testIdPattern": "[page.spec] Page Page.exposeFunction should be callable from-inside evaluateOnNewDocument",
```
OrKoN, line 3723: same question

packages/puppeteer-core/src/common/bidi/Frame.ts
```diff
@@ -207,4 +209,19 @@ export class BidiFrame extends Frame {
...
+    apply: (...args: Args) => Awaitable<Ret>
+  ): Promise<void> {
+    if (this.#exposedFunctions.has(name)) {
+      throw new Error(
+        `Failed to add page binding with name ${name}: globalThis['${name}'] already exists!`
+      );
+    }
+    const exposeable = new ExposeableFunction(this, name, apply);
+    await exposeable.expose();
+    this.#exposedFunctions.set(name, exposeable);
```
OrKoN, line 224: we should probably set it before we await `.expose` to avoid races

packages/puppeteer-core/src/common/bidi/ExposedFunction.ts
```diff
@@ -0,0 +1,260 @@
+import * as Bidi from 'chromium-bidi/lib/cjs/protocol/protocol.js';
```
OrKoN, line 17: mandatory comment: the license header

packages/puppeteer-core/src/common/bidi/Frame.ts
jrandolf: If `expose` fails, it's not exposed then, so we shouldn't set it.

OrKoN: imo, then we need to remove the failed function. The current version does not work as a mutex

jrandolf: Indeed!

test/TestExpectations.json
jrandolf: `reload` times out.

test/TestExpectations.json
jrandolf: This should be PASS

OrKoN (approved):

PR #10861, opened 2023-09-07
test: upstream test and expectation for Firefox sync to puppeteer v21…

juliandescottes:
Another test update from a Puppeteer/Firefox sync.

> TestExpectations.json
> Launcher specs Puppeteer Puppeteer.launch should launch Chrome properly with 

Those tests can't work on Mozilla CI because we don't perform a regular install of puppeteer there.

> TestExpectations.json
> Launcher specs Puppeteer Puppeteer.launch should work with no default arguments

Same thing.

> evaluation.spec.ts
> should not throw an error when evaluation does a navigation

Is ending too quickly on our CI, and the next test is trying to use the browsing context which is navigating, failing frequently.

> target.spec.ts
> should timeout waiting for a non-existent target

Depending on which test ran before, there could be a `server.EMPTY_PAGE` target available, and for some reason that lead the test to immediately fail with a TIMEOUT. Not quite sure why, but I think using an obviously not matching URL is safer?

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -2720,20 +2720,14 @@
   {
     "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Puppeteer.launch should launch Chrome properly with --no-startup-window and waitForInitialPage=false",
     "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["firefox", "webDriverBiDi"],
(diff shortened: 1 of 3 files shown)


juliandescottes:
Converting to draft, still having intermittents in CI

juliandescottes:
Will directly switch to v21.2.0

PR #10862, opened 2023-09-07
chore: update chromium-bidi

jrandolf-google:


packages/puppeteer-core/src/common/Connection.ts
```diff
@@ -132,19 +131,19 @@ export class CallbackRegistry {
...
       // We still throw sync errors synchronously and clean up the scheduled
       // callback.
-      callback.promise
-        .valueOrThrow()
-        .catch(debugError)
-        .finally(() => {
-          this.#callbacks.delete(callback.id);
-        });
       callback.reject(error as Error);
-      throw error;
```
OrKoN: AFAIK we still need it to properly throw sync errors

jrandolf: The `callback.promise.valueOrThrow` will immediately throw because the error rejects.

OrKoN: Let's separate the changes to Connection from the chomium-bidi update? so that we can see the test results clearly

Lightning00Blade (approved):

Lightning00Blade:
Test are failing

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -241,8 +237,11 @@ export class BidiPage extends Page {
 
   #onFrameLoaded(info: Bidi.BrowsingContext.NavigationInfo): void {
     const frame = this.frame(info.context);
-    if (frame && this.mainFrame() === frame) {
-      this.emit(PageEmittedEvents.Load);
+    if (frame) {
+      if (this.mainFrame() === frame) {
+        this.emit(PageEmittedEvents.Load);
+      }
+      this.emit(PageEmittedEvents.FrameNavigated, frame);
```
Lightning00Blade: This is not correct. Or at least there is better way of doing it then what we currently have and that is DomContentLoaded event, as that event is reported at a earlier stage then this.

packages/puppeteer-core/src/common/bidi/Page.ts
```diff
@@ -241,8 +241,17 @@ export class BidiPage extends Page {
 
   #onFrameLoaded(info: Bidi.BrowsingContext.NavigationInfo): void {
     const frame = this.frame(info.context);
-    if (frame && this.mainFrame() === frame) {
-      this.emit(PageEmittedEvents.Load);
+    if (frame) {
+      if (this.mainFrame() === frame) {
```
Lightning00Blade: Undo change

PR #10878, opened 2023-09-11
chore: reimplement document caching

### LINKED ISSUE #10813: Reimplement document handle caching
Document handle caching was removed in https://github.com/puppeteer/puppeteer/pull/10806. We should reimplement this is a generic way (across BiDi and CDP).

jrandolf-google:
Fixed: https://github.com/puppeteer/puppeteer/issues/10813

### DIFF
--- a/packages/puppeteer-core/src/api/Frame.ts
+++ b/packages/puppeteer-core/src/api/Frame.ts
@@ -322,14 +322,31 @@ export abstract class Frame extends EventEmitter {
    */
   abstract isolatedRealm(): Realm;
 
+  #_document: Promise<ElementHandle<Document>> | undefined;
+
   /**
    * @internal
    */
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -322,14 +322,29 @@ export abstract class Frame extends EventEmitter {
...
+        });
+    }
+    return this.#_document;
+  }
+
+  /**
+   * @internal
+   */
+  clearDocumentHandle(): void {
+    this.#_document = undefined;
```
OrKoN, line 349: do we need to call dispose?

jrandolf: No. I've added a comment.

OrKoN (approved):

PR #10880, opened 2023-09-11
fix: add --disable-search-engine-choice-screen to default arguments

OrKoN:
The screen might block automation clients like Puppeteer. Therefore, adding it to the list of arguments specified by default. The flag is added in https://chromium-review.googlesource.com/c/chromium/src/+/4853540.


jrandolf-google (approved):

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -200,6 +200,7 @@ export class ChromeLauncher extends ProductLauncher {
       '--no-first-run',
       '--password-store=basic',
       '--use-mock-keychain',
+      '--disable-search-engine-choice-screen',
```
thiagowfx: nit: this list of arguments is sorted, please keep it sorted

thiagowfx (approved):

Issue #10883, opened 2023-09-11
[Bug]: double --disable-features ignores first instance


paulirish:
### Minimal, reproducible example

```TypeScript
const puppeteer = require('puppeteer');
console.log(puppeteer); throw new Error('😩');
```

### Error string

😩

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

Running devtools interaction tests.. I get this:

![image](https://github.com/puppeteer/puppeteer/assets/39191/c3205ec8-7871-4e33-ac28-a90b9cc9f3db)

I shouldn't get this if `--disable-features=MediaRouter` is set. Looking at the flags that chrome launches with it has two disable-features sets:

`--disable-features=Translate,BackForwardCache,AcceptCHFrame,MediaRouter,OptimizationHints  --disable-features=Prerender2`

It would (definitely) be great if Chrome merged those two but it doesn't. So the second instance wipes out the first.

relevant source, came in recently with the tab target stuff (#10790 #10818): https://github.com/puppeteer/puppeteer/blob/d08ad5fbbe3be4349dd6132c209895f8436ae9e6/packages/puppeteer-core/src/node/ChromeLauncher.ts#L182-L185

--------

devtools repo repro:

```sh
npm run auto-screenshotstest
# note the new prompt
```

### Expectation

no popups about allowing network traffic. (also if you log with `--enable-logging=stderr --v=2` you'll see just how busy that mediarouter is.. sheeeesh!)

### Reality

the intended features werent disabled.

### Puppeteer version

21.1.1

### Node version

19.6.0

### Package manager

npm

### Package manager version

9.4.0

### Operating system

macOS

OrKoN:
Oh, thanks for spotting this!

PR #10887, opened 2023-09-12
fix: only a single disable features flag is allowed

OrKoN:
Bug: #10883


Lightning00Blade (approved):

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -166,6 +166,20 @@ export class ChromeLauncher extends ProductLauncher {
 
   override defaultArgs(options: BrowserLaunchArgumentOptions = {}): string[] {
     // See https://github.com/GoogleChrome/chrome-launcher/blob/main/docs/chrome-flags-for-tools.md
+
+    const disabledFeatures = [
```
Lightning00Blade, line 170: Nit: I think we should leave a comment for the future.

PR #10888, opened 2023-09-12
refactor: use RxJs for WaitForNetworkIdle

Lightning00Blade:


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1696,62 +1700,33 @@ export abstract class Page
...
+        fromEvent(
+          networkManager,
+          NetworkManagerEmittedEvents.Response as unknown as string
+        ),
+        fromEvent(
+          networkManager,
+          NetworkManagerEmittedEvents.RequestFailed as unknown as string
+        )
+      ).pipe(
+        startWith(null),
```
jrandolf, line 1721: Why `startWith`?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1696,62 +1700,33 @@ export abstract class Page
...
+        filter(() => {
+          return networkManager.inFlightRequestsCount() === 0;
+        }),
+        switchMap(v => {
```
jrandolf, line 1725: Use `mergeMap`.

packages/puppeteer-core/src/common/util.ts
```diff
@@ -708,3 +710,16 @@ export class Mutex {
     resolve();
   }
 }
+
+/**
+ * @internal
+ */
+export function timeout(ms: number): Observable<never> {
+  return ms === 0
```
jrandolf, line 718: How about `Infinity` rather than `0`?

jrandolf-google (approved):

jrandolf-google (approved):
lgtm % comments

packages/puppeteer-core/src/api/Page.ts
Lightning00Blade: An event may never come we need to trigger the observer 1 time.

packages/puppeteer-core/src/common/util.ts
Lightning00Blade: That's incorrect as the Common Puppeteer representation of `No Timeout` is `timeout=0`. 
Also if you use `SetTimeout(()=>{}, Infinity)` Node may complain.
```
TimeoutOverflowWarning: Infinity does not fit into a 32-bit signed integer.
```

packages/puppeteer-core/src/api/Page.ts
Lightning00Blade: That does not work. We need `switchMap` as we want to create a new Observable that will be cancelled if its we rerun `swichMap` and produce a new instance.

PR #10889, opened 2023-09-12
feat: implement typed events

jrandolf-google:
This PR implements typings for emitted events in all event emitters.


packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -127,6 +127,44 @@ export interface FrameAddStyleTagOptions {
   content?: string;
 }
 
+/**
+ * @public
+ */
+export type FrameEvents = {
```
Lightning00Blade: Why make the type public but every entry is internal?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -33,20 +33,25 @@ import {
 import type {HTTPRequest} from '../api/HTTPRequest.js';
 import type {HTTPResponse} from '../api/HTTPResponse.js';
 import type {Accessibility} from '../common/Accessibility.js';
-import type {CDPSession} from '../common/Connection.js';
+import type {BidiNetworkManager as BidiNetworkManager} from '../common/bidi/NetworkManager.js';
```
Lightning00Blade: Unnecessary name change.

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -513,54 +535,48 @@ export abstract class Page
...
-   * Listen to page events.
-   *
-   * :::note
-   *
-   * This method exists to define event typings and handle proper wireup of
-   * cooperative request interception. Actual event listening and dispatching is
-   * delegated to {@link EventEmitter}.
-   *
-   * :::
+   * @internal
```
Lightning00Blade, line 550: I think comment is still relevant.

packages/puppeteer-core/src/common/CDPSession.ts
```diff
@@ -0,0 +1,154 @@
...
+  createProtocolErrorMessage,
+} from './Connection.js';
+import {TargetCloseError} from './Errors.js';
+import {CDPTarget} from './Target.js';
+
+/**
+ * @internal
+ */
+
+export class CDPCDPSession extends CDPSession {
```
Lightning00Blade, line 18: Use the common prefix for specific `Cdp`/`Bidi` implementations
```suggestion
export class CdpCDPSession extends CDPSession {
```

packages/puppeteer-core/src/common/ExecutionContext.ts
```diff
@@ -47,25 +47,6 @@ const getSourceUrlComment = (url: string) => {
 };
 
 /**
- * Represents a context for JavaScript execution.
```
Lightning00Blade, line 50: Keep comments.

packages/puppeteer-core/src/common/FrameManager.ts
```diff
@@ -190,7 +202,7 @@ export class FrameManager extends EventEmitter {
     });
     session.on('Page.frameNavigated', event => {
       this.#frameNavigatedReceived.add(event.frame.id);
-      void this.#onFrameNavigated(event.frame, event.type);
```
Lightning00Blade: Why this change?

packages/puppeteer-core/src/common/FrameManager.ts
```diff
@@ -346,7 +358,7 @@ export class FrameManager extends EventEmitter {
       );
     }
     if (!this.#frameNavigatedReceived.has(frameTree.frame.id)) {
-      void this.#onFrameNavigated(frameTree.frame, 'Navigation');
```
Lightning00Blade: Why this change?

packages/puppeteer-core/src/common/FrameManager.ts
```diff
@@ -412,8 +421,8 @@ export class FrameManager extends EventEmitter {
 
     frame = await this._frameTree.waitForFrame(frameId);
     frame._navigated(framePayload);
-    this.emit(FrameManagerEmittedEvents.FrameNavigated, frame);
-    frame.emit(FrameEmittedEvents.FrameNavigated, navigationType);
```
Lightning00Blade, line 416: Why this change?

packages/puppeteer-core/src/common/FrameManager.ts
```diff
@@ -474,8 +483,9 @@ export class FrameManager extends EventEmitter {
         this.#removeFramesRecursively(frame);
       }
     } else if (reason === 'swap') {
-      this.emit(FrameManagerEmittedEvents.FrameSwapped, frame);
-      frame?.emit(FrameEmittedEvents.FrameSwapped);
+      assert(frame);
```
Lightning00Blade: This can produce unhandled rejection. Wrap like the other?

packages/puppeteer-core/src/common/HTTPRequest.ts
```diff
@@ -30,20 +31,19 @@ import {
...
 
-import {CDPSession} from './Connection.js';
 import {ProtocolError} from './Errors.js';
 import {debugError, isString} from './util.js';
 
 /**
  * @internal
  */
-export class HTTPRequest extends BaseHTTPRequest {
+export class CDPHTTPRequest extends HTTPRequest {
```
Lightning00Blade, line 40:
```suggestion
export class CdpHTTPRequest extends HTTPRequest {
```

packages/puppeteer-core/src/common/HTTPResponse.ts
```diff
@@ -15,24 +15,21 @@
...
 import {ProtocolError} from './Errors.js';
-import {HTTPRequest} from './HTTPRequest.js';
+import {CDPHTTPRequest} from './HTTPRequest.js';
 import {SecurityDetails} from './SecurityDetails.js';
 
 /**
  * @internal
  */
-export class HTTPResponse extends BaseHTTPResponse {
+export class CDPHTTPResponse extends HTTPResponse {
```
Lightning00Blade, line 30:
```suggestion
export class CdpHTTPResponse extends HTTPResponse {
```

packages/puppeteer-core/src/common/NetworkManager.test.ts
```diff
@@ -692,14 +692,11 @@ describe('NetworkManager', () => {
 
     const finishedRequests: HTTPRequest[] = [];
     const pendingRequests: HTTPRequest[] = [];
-    manager.on(
-      NetworkManagerEmittedEvents.RequestFinished,
-      (request: HTTPRequest) => {
-        finishedRequests.push(request);
-      }
-    );
+    manager.on(NetworkManagerEvent.RequestFinished, (request: HTTPRequest) => {
```
Lightning00Blade, line 695: Is type here for property still needed?

packages/puppeteer-core/src/common/Page.ts
```diff
@@ -146,58 +142,59 @@ export class CDPPage extends Page {
   #serviceWorkerBypassed = false;
   #userDragInterceptionEnabled = false;
 
-  #frameManagerHandlers = new Map<symbol, Handler<any>>([
+  #frameManagerHandlers = Object.freeze([
```
Lightning00Blade, line 145: Maybe we should use `readonly` as they are internal methods.

packages/puppeteer-core/src/common/Page.ts
```diff
@@ -146,58 +142,59 @@ export class CDPPage extends Page {
...
+        this.#sessionCloseDeferred.resolve(
           new TargetCloseError('Target closed')
         );
       },
     ],
     [
       'Page.domContentEventFired',
-      this.emit.bind(this, PageEmittedEvents.DOMContentLoaded),
+      () => {
+        return this.emit(PageEvent.DOMContentLoaded, undefined);
```
Lightning00Blade, line 211: Why this change, can't you pass `undefined` as 3th parameter?

packages/puppeteer-core/src/common/Page.ts
```diff
@@ -254,57 +251,50 @@ export class CDPPage extends Page {
...
-          return;
-        }
-        this.#frameManager
-          .registerSpeculativeSession(session)
-          .catch(debugError);
-        this.#emulationManager
-          .registerSpeculativeSession(session)
-          .catch(debugError);
+    this.#tabSession?.on(CDPSessionEvent.Ready, session => {
+      assert(session instanceof CDPCDPSession);
```
Lightning00Blade, line 277: Not sure of this check.

packages/puppeteer-core/src/common/WebWorker.ts
```diff
@@ -69,7 +68,7 @@ export type ExceptionThrownCallback = (
  *
  * @public
  */
-export class WebWorker extends EventEmitter {
```
Lightning00Blade, line 72: This is a breaking change.

packages/puppeteer-core/src/common/bidi/BidiOverCDP.ts
```diff
@@ -53,7 +50,7 @@ export async function connectBidiOverCDP(
     // Forwards a BiDi event sent by BidiServer to Puppeteer.
     pptrTransport.onmessage(JSON.stringify(message));
   });
-  const pptrBiDiConnection = new BidiPPtrConnection(cdp.url(), pptrTransport);
+  const pptrBiDiConnection = new BidiConnection(cdp.url(), pptrTransport);
```
Lightning00Blade, line 53: Not sure of the renaming to `BidiConnection` will have to look what the class does.

packages/puppeteer-core/src/common/bidi/Connection.ts
```diff
@@ -218,7 +225,7 @@ export class Connection extends EventEmitter {
     } else if (isCDPEvent(event)) {
       cdpSessions
         .get(event.params.session)
-        ?.emit(event.params.event, event.params.params);
+        ?.emit(event.params.event, event.params.params[0]);
```
Lightning00Blade: Why this change? I think that's not relevant for BiDi+

packages/puppeteer-core/src/api/Frame.ts
jrandolf: The type cannot resolve in the public types if `FrameEvents` is internal. This will lead to an unknown signature for `Frame`.

packages/puppeteer-core/src/common/CDPSession.ts
jrandolf: We can do this is another PR. We should consistently use `CDP` since that's how it is everywhere in puppeteer.

packages/puppeteer-core/src/common/bidi/Connection.ts
jrandolf: Because that's the correct typing.

packages/puppeteer-core/src/common/Page.ts
jrandolf: What are you not sure about?

packages/puppeteer-core/src/common/Page.ts
jrandolf: TypeScript has trouble without the above.

packages/puppeteer-core/src/common/ExecutionContext.ts
jrandolf: This is left-over from when this was public. The comment is not relevant anymore (and we plan to remove this anyway).

OrKoN:
Would it be considered a breaking change? I assume if someone used string literals for event names before it might not compile after this change.

OrKoN:
Could you add a description about what has specifically changed with a basic before and after example? (I assume `feat:` is marked correctly and indicates a user facing feature and not a refactoring).

jrandolf:
> Would it be considered a breaking change? I assume if someone used string literals for event names before it might not compile after this change.

There are no breaking changes, see tests for example.

> Could you add a description about what has specifically changed with a basic before and after example? (I assume feat: is marked correctly and indicates a user facing feature and not a refactoring).

Done.

OrKoN:
> > Would it be considered a breaking change? I assume if someone used string literals for event names before it might not compile after this change.
> 
> There are no breaking changes, see tests for example.
> 
> > Could you add a description about what has specifically changed with a basic before and after example? (I assume feat: is marked correctly and indicates a user facing feature and not a refactoring).
> 
> Done.

so what is the user facing change then? should it be prefixed as `refactor:`?

jrandolf:
> so what is the user facing change then? should it be prefixed as refactor:?

Users now get the correct typing when they try to add a listener which required manual typing previously. We could label it as a `refactor` if you feel this isn't sufficient to label as a `feat`.

OrKoN:
> > so what is the user facing change then? should it be prefixed as refactor:?
> 
> Users now get the correct typing when they try to add a listener which required manual typing previously. We could label it as a `refactor` if you feel this isn't sufficient to label as a `feat`.

I am not sure, maybe feat is good. Btw, GitHub UI really struggles to render the diffs in readable manner: I guess it's lazy-loaded so you cannot actually jump to a specific file without scrolling for a while first.

packages/puppeteer-core/src/common/FrameManager.ts
Lightning00Blade: Do an early return instead.

packages/puppeteer-core/src/common/FrameManager.ts
Lightning00Blade: Not addressed, this may brake BFCache

packages/puppeteer-core/src/common/bidi/NetworkManager.ts
```diff
@@ -16,40 +16,71 @@
...
+  [NetworkManagerEvent.RequestFinished]: BidiHTTPRequest;
+}
+
+/**
+ * @internal
+ */
+export class BidiNetworkManager extends EventEmitter<NetworkManagerEvents> {
+  #connection: BidiConnection;
   #page: BidiPage;
-  #subscribedEvents = new Map<string, Handler<any>>([
```
Lightning00Blade, line 34: It feels quicker to find what events when searching, when we have them in a map and then loop though them. 
We made the change in other places to use the same pattern. 
Leaving to @OrKoN to give a vote on this.

packages/puppeteer-core/src/common/bidi/Connection.ts
```diff
@@ -202,7 +212,8 @@ export class Connection extends EventEmitter {
         debugError(object);
       } else {
         this.#maybeEmitOnContext(object);
-        this.emit(object.method, object.params);
+        // SAFETY: We know the method and parameter still match here.
+        this.emit(object.method, object.params as Bidi.BrowsingContext.Info);
```
Lightning00Blade: Why exactly `Bidi.BrowsingContext.Info` shouldn't it be something more generic? Like `Bidi.ChromiumBidi.Event`?

test/src/bfcache.spec.ts
```diff
@@ -25,7 +25,7 @@ describe('BFCache', function () {
     });
 
     try {
-      page.setDefaultTimeout(2000);
+      page.setDefaultTimeout(5000);
```
Lightning00Blade: Why this change?

packages/puppeteer-core/src/common/bidi/Context.ts
Lightning00Blade, line 1: Why is this empty, delete?

packages/puppeteer-core/src/common/bidi/NetworkManager.ts
OrKoN: since DisposableStack keep the track of the list of event subscriptions, I think it's okay.

test/src/bfcache.spec.ts
jrandolf: Tests were timing out here.

packages/puppeteer-core/src/common/bidi/Context.ts
jrandolf: Probably an accidentally committed file.

packages/puppeteer-core/src/common/bidi/Connection.ts
jrandolf: TypeScript has trouble here because `object.method` is a union of all the method types, so it doesn't know how to interpret `object.params`

test/src/bfcache.spec.ts
OrKoN: 2 sec should be more than enough for anything in this test unless smth else is going wrong or the host is not healthy

OrKoN: it seems it timed out with 5 sec too

packages/puppeteer-core/src/common/NetworkManager.test.ts
Lightning00Blade: Please explain why we still need this.

packages/puppeteer-core/src/common/bidi/Connection.ts
Lightning00Blade: Nit: I see, let's type it as `any` as this is not the correct type, comment explains why we do that.

Lightning00Blade (approved):
LGTM + please answer the two question like comments and look at nits, and wait for all test to pass.

packages/puppeteer-core/src/common/NetworkManager.test.ts
Lightning00Blade: It looks like we don't need this anymore can we put TODO/Issue to remove them and rely on our Types.

PR #10895, opened 2023-09-13
refactor: reduce boilerplate in the emulation state management

OrKoN:
Next I will update other state instances.

### DIFF
--- a/packages/puppeteer-core/src/common/EmulationManager.ts
+++ b/packages/puppeteer-core/src/common/EmulationManager.ts
@@ -26,6 +26,7 @@ import {debugError} from './util.js';
 
 interface ViewportState {
   viewport?: Viewport;
+  active: boolean;
 }
 
 interface IdleOverridesState {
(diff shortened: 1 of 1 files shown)


packages/puppeteer-core/src/common/EmulationManager.ts
```diff
@@ -133,7 +173,7 @@ export class EmulationManager {
     });
     // We don't await here because we want to register all state changes before
     // the target is unpaused.
-    void this.#syncViewport().catch(debugError);
+    void this.#viewportState.sync();
```
Lightning00Blade: What happens if this now errors?

Lightning00Blade (approved):
LGTM + comment

packages/puppeteer-core/src/common/EmulationManager.ts
OrKoN:
```suggestion
    void this.#viewportState.sync().catch(debugError);
```

PR #10896, opened 2023-09-13
fix: use supported node range for types

### LINKED ISSUE #10891: [Bug]: v21.2.0 Incompatible with `@types/node@18`

```TypeScript

(async () => {
  const hrefElement = await page.$('a');
  if (hrefElement) {
      await hrefElement.click();
  }
  // ...
})();
```

### Error string

node_modules/puppeteer/lib/types.d.ts:5971:13 - error TS2339: Property 'asyncDispose' does not exist on type 'SymbolConstructor'.

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

When working in a TypeScript (5.1, 5.2) environment with `@types/node@18`, I began to see this error when trying to compile using today's release of Puppeteer (21.2.0). After fiddling with `@types/node` versions, I was able to see that this error occurs for the latest typings of Node 18 (18.17.15), but does not for Node 20 (20.6.0).

I am able to run Puppeteer successfully with the following environment:
- TypeScript v5.1 or v5.2
- Puppeteer v21.1.1
- `@types/node` v18.17.5

It is upon upgrading Puppeteer to v21.2.20 that I begin to see the error

### Expectation

My current understanding is that Puppeteer supports the latest LTS version of Node, which at the time of this writing is Node 18. I would expect to be able to upgrade Puppeteer with my current environment and not run into any compile time issues.

However, that's predicated on the assumption my understanding of Node support (and its associated types is correct). Please let me know if that's not the case!

### Reality

In reality, we get those compile time issues.

_No response_

### Puppeteer version

21.2.0

### Node version

18.17.1

### Package manager

npm

### Package manager version

9.6.7

### Operating system

macOS

jrandolf-google:
Using the supported node range for types ensures Puppeteer compiles to the correct types.

Fixed: https://github.com/puppeteer/puppeteer/issues/10891

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -26,7 +26,7 @@
         "@types/diff": "5.0.3",
         "@types/mime": "3.0.1",
         "@types/mocha": "10.0.1",
-        "@types/node": "20.5.9",
+        "@types/node": "18.17.15",
         "@types/pixelmatch": "5.2.4",
         "@types/pngjs": "6.0.1",
         "@types/progress": "2.0.5",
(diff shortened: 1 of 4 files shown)


packages/puppeteer/package.json
```diff
@@ -30,9 +30,6 @@
     "type": "git",
     "url": "https://github.com/puppeteer/puppeteer/tree/main/packages/puppeteer"
   },
-  "engines": {
```
Lightning00Blade: Why remove? Keep this it prints a warning about you using older version. There was a report about this sometime ago.

Lightning00Blade (approved):

PR #10898, opened 2023-09-13
test: upstream test and expectation for Firefox sync to puppeteer v21…

juliandescottes:
Another test update from a Puppeteer/Firefox sync.

> TestExpectations.json
> Launcher specs Puppeteer Puppeteer.launch should launch Chrome properly with 

Those tests can't work on Mozilla CI because we don't perform a regular install of puppeteer there.

> TestExpectations.json
> Launcher specs Puppeteer Puppeteer.launch should work with no default arguments

Same thing.

> TestExpectations.json
> [proxy.spec] request proxy should respect proxy bypass list

This test frequently fails on our CI because the responseCompleted event is received too late (ie after navigate has resolved). Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1852891 about this.

> TestExpectations.json
> [network.spec] network Response.fromCache should work

I am not sure how this one passes on the puppeteer CI. Locally I never get a second request for the CSS file which we expect to get from the cache. Maybe we are completely missing the request in this case, but I don't know how this can pass on the puppeteer CI.

> TestExpectations.json
> [waittask.spec] waittask specs Frame.waitForFunction should work when resolved right before execution context disposal

The polling mechanism used in this test always fail for me locally and on CI. It might pass on puppeteer CI if it has slightly different timings?

> evaluation.spec.ts
> should not throw an error when evaluation does a navigation

Is ending too quickly on our CI, and the next test is trying to use the browsing context which is navigating, failing frequently.

> page.spec.ts
> Page.Events.PageError should fire

I frequently receive other errors for the same context during the test, which make the test fail. I'm proposing to wait for the specific error message here to avoid the issue.

> target.spec.ts
> should timeout waiting for a non-existent target

Depending on which test ran before, there could be a `server.EMPTY_PAGE` target available, and for some reason that lead the test to immediately fail with a TIMEOUT. Not quite sure why, but I think using an obviously not matching URL is safer?

### DIFF
--- a/test/src/page.spec.ts
+++ b/test/src/page.spec.ts
@@ -1365,7 +1365,9 @@ describe('Page', function () {
       const {page, server} = await getTestState();
 
       const [error] = await Promise.all([
-        waitEvent<Error>(page, 'pageerror'),
+        waitEvent<Error>(page, 'pageerror', _error => {
+          return _error.message.includes('Fancy');
+        }),
         page.goto(server.PREFIX + '/error.html'),
       ]);
       expect(error.message).toContain('Fancy');
--- a/test/src/page.spec.ts
+++ b/test/src/page.spec.ts
@@ -1365,8 +1365,8 @@ describe('Page', function () {
       const {page, server} = await getTestState();
 
       const [error] = await Promise.all([
-        waitEvent<Error>(page, 'pageerror', _error => {
-          return _error.message.includes('Fancy');
+        waitEvent<Error>(page, 'pageerror', err => {
+          return err.message.includes('Fancy');
         }),
         page.goto(server.PREFIX + '/error.html'),
       ]);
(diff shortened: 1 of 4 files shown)


test/src/page.spec.ts
```diff
@@ -1364,7 +1364,9 @@ describe('Page', function () {
       const {page, server} = await getTestState();
 
       const [error] = await Promise.all([
-        waitEvent<Error>(page, 'pageerror'),
+        waitEvent<Error>(page, 'pageerror', _error => {
+          return _error.message.includes('Fancy');
```
Lightning00Blade: What other error are you seeing here?
Prefixing with `_` shouldn't be done as ESLint will ignore some rules for variable starting with it.
```suggestion
        waitEvent<Error>(page, 'pageerror', err => {
          return err.message.includes('Fancy');
```

juliandescottes: I wasn't able to track them down (no stack trace), but the message was `Permission denied to access property "length"`. I only have those on CI

Lightning00Blade (approved):

juliandescottes:
@Lightning00Blade I see there are failures in the CI jobs for this PR. Are they known intermittent issues or should I investigate?

Lightning00Blade:
It looks like flake that is observed on other CI actions, so no action required.

PR #10904, opened 2023-09-13
chore: use `throwIfDisposed` in `ElementHandle` and `JSHandle`

jrandolf-google:


packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -276,12 +272,17 @@ export abstract class ElementHandle<
     pageFunction: Func | string,
     ...args: Params
   ): Promise<HandleFor<Awaited<ReturnType<Func>>>> {
+    pageFunction = withSourcePuppeteerURLIfNone(
```
Lightning00Blade, line 275: Why not add to this method?

Lightning00Blade (approved):

packages/puppeteer-core/src/api/ElementHandle.ts
jrandolf: Because `withSourcePuppeteerURLIfNone` relies on the callstack which gets modified with decorators.

PR #10909, opened 2023-09-14
refactor: use EventSubscription

jrandolf-google:
Also modifies several locations where `CdpCDPSession` should be used.

### DIFF
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -442,7 +442,7 @@ export interface PageEvents extends Record<EventType, unknown> {
   [PageEvent.Response]: HTTPResponse;
   [PageEvent.RequestFailed]: HTTPRequest;
   [PageEvent.RequestFinished]: HTTPRequest;
-  [PageEvent.RequestServedFromCache]: HTTPRequest;
+  [PageEvent.RequestServedFromCache]: HTTPRequest | undefined;
   [PageEvent.WorkerCreated]: WebWorker;
   [PageEvent.WorkerDestroyed]: WebWorker;
 }
--- a/packages/puppeteer-core/src/common/Page.ts
+++ b/packages/puppeteer-core/src/common/Page.ts
@@ -52,6 +52,7 @@ import {type DeviceRequestPrompt} from './DeviceRequestPrompt.js';
 import {CdpDialog} from './Dialog.js';
 import {EmulationManager} from './EmulationManager.js';
 import {TargetCloseError} from './Errors.js';
+import {EventSubscription} from './EventEmitter.js';
 import {FileChooser} from './FileChooser.js';
 import {type CdpFrame} from './Frame.js';
 import {FrameManager, FrameManagerEvent} from './FrameManager.js';
@@ -92,7 +93,7 @@ import {WebWorker} from './WebWorker.js';
  */
 export class CdpPage extends Page {
   static async _create(
-    client: CDPSession,
+    client: CdpCDPSession,
     target: CdpTarget,
     ignoreHTTPSErrors: boolean,
     defaultViewport: Viewport | null,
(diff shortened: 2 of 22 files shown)


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -437,7 +437,7 @@ export interface PageEvents extends Record<EventType, unknown> {
   [PageEvent.Response]: HTTPResponse;
   [PageEvent.RequestFailed]: HTTPRequest;
   [PageEvent.RequestFinished]: HTTPRequest;
-  [PageEvent.RequestServedFromCache]: HTTPRequest;
+  [PageEvent.RequestServedFromCache]: HTTPRequest | undefined;
```
Lightning00Blade, line 445: Revert this, this should never happen. We should fix the CDP implementation to emit only when the request exists.

test/src/network.spec.ts
```diff
@@ -486,6 +487,7 @@ describe('network', function () {
 
       const cached: string[] = [];
       page.on('requestservedfromcache', r => {
+        assert(r);
```
Lightning00Blade, line 490: Revert all changes for test.

packages/puppeteer-core/src/common/Page.ts
```diff
@@ -253,68 +170,225 @@ export class CdpPage extends Page {
...
-    this.#tabSession?.on(CDPSessionEvent.Ready, session => {
-      assert(session instanceof CdpCDPSession);
-      if (session._target()._subtype() !== 'prerender') {
-        return;
-      }
-      this.#frameManager.registerSpeculativeSession(session).catch(debugError);
-      this.#emulationManager
-        .registerSpeculativeSession(session)
-        .catch(debugError);
+    this.once(PageEvent.Close, () => {
```
Lightning00Blade, line 173: You can do this before emitting the event rather then waiting for it here.

packages/puppeteer-core/src/api/Page.ts
jrandolf: This can occur. See https://crbug.com/750469. It shouldn't be resolved.

test/src/network.spec.ts
jrandolf: This is required. See the other comment.

packages/puppeteer-core/src/common/Page.ts
jrandolf: Is there a difference? I plan to refactor this in any case. Not here.

packages/puppeteer-core/src/api/Page.ts
Lightning00Blade: The user can't do anything if we return `undefined`, they can't understand why or from where it's coming without looking at CDP logs, so changing to only emit when we already have the request in internal map is the correct way of doing it. Also braking API change. Add link to crbug.

packages/puppeteer-core/src/common/Page.ts
Lightning00Blade: It's related to order, but also setting up one extra listener that is not needed.

packages/puppeteer-core/src/api/Page.ts
jrandolf: `RequestServedFromCache` _needs_ to be emitted whenever that event comes in. Whether it exists or not in the internal map is a bug due to `https://crbug.com/750469`.

This isn't a breaking change since this is what happens at runtime.

Lightning00Blade: That is not true we want to emit the event whenever we know that the `HTTPRequest` comes from cache. 
What you are proposing gives no user value but they will need to add extra checks for this.

jrandolf: The cache is from the internal, Chromium cache. Not the one in Puppeteer.

Lightning00Blade: Again I think you are missing the point - As a user this event will provide no value to me as it will be empty and can't do anything with it - log info or intercept it, thus I will always need to ignore it. We should ignore internally and save the user a check, also we should keep the API the same as it is now and fix our implementation to match it, whenever possible.

Lightning00Blade (approved):
LGTM after offline discussion there is bug with this event and network interception. Will file and issue with explenation.

PR #10912, opened 2023-09-14
feat!: drop support for node16

Lightning00Blade:
A good opportunity to remove some deprecation if possible/old enough. 
I think we should cut a release before merging this.


packages/browsers/package.json
```diff
@@ -89,7 +89,7 @@
   "author": "The Chromium Authors",
   "license": "Apache-2.0",
   "engines": {
-    "node": ">=16.3.0"
+    "node": ">=18.17.1"
```
jrandolf: `>=18` everywhere.

jrandolf-google (changes requested):

jrandolf:
If you could, could you look back at v18 and remove anything that was deprecated?

jrandolf:
Let's create an issue that outlines what we will be deprecating from Puppeteer. For breaking releases, we should document everything so people can also chime in.

OrKoN:
I actually just remembered, we should double check the version of Node Chromium is on.

OrKoN:
yeah https://source.chromium.org/chromium/chromium/src/+/main:third_party/node/update_node_binaries;l=18;drc=5964ce8eb626cc3c18f437f194a1e9234aebe8aa

Lightning00Blade:
Note: This is required to update Rollup to v4

OrKoN (approved):

jrandolf-google (approved):

dandan3100 (approved):
GitBush/NObOdy.eth

PR #10914, opened 2023-09-14
chore: separate code to semantic folders

jrandolf-google:
This PR moves CDP, BiDi and common code to their own folders.


Lightning00Blade (approved):
LGTM - Will send a follow up that cleans up some parts.

OrKoN (approved):
LGTM % some test failures look new and could be related to the change

PR #10929, opened 2023-09-18
chore: fix BiDi redirects

Lightning00Blade:


packages/puppeteer-core/src/cdp/HTTPRequest.ts
```diff
@@ -37,12 +37,7 @@ import {assert} from '../util/assert.js';
  * @internal
  */
 export class CdpHTTPRequest extends HTTPRequest {
-  override _requestId: string;
-  override _interceptionId: string | undefined;
-  override _failureText: string | null = null;
-  override _response: HTTPResponse | null = null;
-  override _fromMemoryCache = false;
-  override _redirectChain: CdpHTTPRequest[];
+  declare _redirectChain: CdpHTTPRequest[];
```
OrKoN, line 40: are we re-declaring it this way to have the right type?

Lightning00Blade: Yes, that way we can only push `CpdHTTPRequest`.

OrKoN (approved):

812142 (commented):
Fix BIDI redirects

PR #10931, opened 2023-09-18
chore: use GitHub URL for source root

jrandolf-google:
`release-please` will update the versions whenever they need updating.

### DIFF
--- a/packages/puppeteer-core/package.json
+++ b/packages/puppeteer-core/package.json
@@ -129,13 +129,7 @@
     }
   },
   "files": [
-    "lib",
-    "src",
-    "!*.test.js",
-    "!*.test.d.ts",
-    "!*.test.js.map",
-    "!*.test.d.ts.map",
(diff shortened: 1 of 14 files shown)


packages/puppeteer-core/package.json
```diff
@@ -129,13 +129,7 @@
     }
   },
   "files": [
-    "lib",
```
Lightning00Blade, line 132: Revert this change this was reported back that we should include src so that source maps work.
Also the other prevents test from being in the submitted upload.

jrandolf: It's okay to remove this now. The source map will point back to the GitHub url.

packages/puppeteer/package.json
```diff
@@ -110,13 +110,7 @@
   },
   "files": [
     "lib",
-    "src",
```
Lightning00Blade, line 113: Same here.

PR #10933, opened 2023-09-18
chore: implement `@puppeteer/doctest`

jrandolf-google:
This PR implements a documentation example tester, similar to the [one in Rust](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html).

## Usage

```
npm run doctest
```

## Details

`@puppeteer/doctest` looks through compiled code for `@example` comments with fenced code blocks. It will then parse out the code within the fence and run it using Node's test runner.

The first line containing the starting fence defines the options for a given fenced code block. For example,

 - <code>```ts</code> defines a TypeScript code block.
 - <code>```</code> defines a JavaScript code block.
 - <code>```ts ignore</code> defines a TypeScript code block that is ignored for testing.

Note that

 - <code>``` ignore</code> and 
 - <code>```ignore</code>

are different. The former is an ignored JS code block. The latter is a code block for the `ignore` language.

If there are several code blocks within an example, they are concatenated before running.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "lint": "run-s lint:prettier lint:eslint",
     "postinstall": "npm run postinstall --workspaces --if-present",
     "prepare": "npm run prepare --workspaces --if-present",
+    "doctest": "wireit",
     "test-install": "npm run test --workspace @puppeteer-test/installation",
     "test-types": "tsd -t packages/puppeteer",
     "test:chrome:headful": "wireit",
@@ -57,6 +58,13 @@
         "./packages/browsers:build:docs",
         "./packages/puppeteer:build:docs",
         "./packages/puppeteer-core:build:docs"
+      ]
+    },
+    "doctest": {
+      "command": "npx ./tools/doctest 'packages/puppeteer-core/lib/esm/**/*.js'",
+      "dependencies": [
+        "./packages/puppeteer-core:build",
+        "./tools/doctest:build"
       ]
     },
     "test:chrome:headful": {
@@ -184,6 +192,7 @@
     "packages/*",
     "test",
     "test/installation",
-    "tools/eslint"
+    "tools/eslint",
+    "tools/doctest"
   ]
 }
--- a/tools/doctest/package.json
+++ b/tools/doctest/package.json
@@ -1,0 +1,41 @@
+{
+  "name": "@puppeteer/doctest",
+  "version": "0.1.0",
+  "type": "module",
+  "private": true,
+  "bin": "./bin/doctest.js",
+  "description": "Tests JSDoc @example code within a file.",
+  "license": "Apache-2.0",
+  "scripts": {
+    "build": "wireit",
+    "clean": "../clean.js"
+  },
+  "wireit": {
+    "build": {
+      "command": "tsc -b && chmod +x ./bin/doctest.js",
+      "clean": "if-file-deleted",
+      "files": [
+        "src/**"
+      ],
+      "output": [
+        "bin/**",
+        "tsconfig.tsbuildinfo"
+      ]
+    }
+  },
+  "dependencies": {
+    "@swc/core": "1.3.85",
+    "acorn": "8.10.0",
+    "doctrine": "3.0.0",
+    "glob": "10.3.4",
+    "pkg-dir": "7.0.0",
+    "source-map": "0.7.4",
+    "source-map-support": "0.5.21",
+    "yargs": "17.7.2"
+  },
+  "devDependencies": {
+    "@types/doctrine": "0.0.6",
+    "@types/node": "20.6.2",
+    "@types/source-map-support": "0.5.7"
+  }
+}
--- a/tools/doctest/src/doctest.ts
+++ b/tools/doctest/src/doctest.ts
@@ -1,0 +1,359 @@
+#! /usr/bin/env -S node --test-reporter spec
+
+/**
+ * Copyright 2023 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * `@puppeteer/doctest` tests `@example` code within a JavaScript file.
+ *
+ * There are a few reasonable assumptions for this tool to work:
+ *
+ * 1. Examples are written in block comments, not line comments.
+ * 2. Examples do not use packages that are not available to the file it exists
+ *    in. (Note the package will always be available).
+ * 3. Examples are strictly written between code fences (\`\`\`) on separate
+ *    lines. For example, \`\`\`console.log(1)\`\`\` is not allowed.
+ * 4. Code is written using ES modules.
+ *
+ * By default, code blocks are interpreted as JavaScript. Use \`\`\`ts to change
+ * the language. In general, the format is "\`\`\`[language] [ignore] [fail]".
+ *
+ * If there are several code blocks within an example, they are concatenated.
+ */
+import 'source-map-support/register.js';
+
+import assert from 'node:assert';
+import {createHash} from 'node:crypto';
+import {mkdtemp, readFile, rm, writeFile} from 'node:fs/promises';
+import {basename, dirname, join, relative, resolve} from 'node:path';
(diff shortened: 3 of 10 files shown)


OrKoN:
I think it is a good idea, I will let @Lightning00Blade to do the review

test/TestExpectations.json
```diff
@@ -779,6 +779,12 @@
     "parameters": ["webDriverBiDi"],
     "expectations": ["SKIP"]
   },
+  {
+    "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Puppeteer.launch should launch Chrome properly with --no-startup-window and waitForInitialPage=false",
```
OrKoN: let's investigate flakiness today and address it separately

OrKoN: never mind (will post on chat)

tools/doctest/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@puppeteer/doctest",
+  "version": "0.1.0",
+  "type": "module",
+  "bin": "./bin/doctest.js",
+  "description": "Tests JSDoc @example code within a file.",
```
Lightning00Blade, line 7: Nit:
```suggestion
  "description": "Tests JSDoc/TSDoc @example code within a file.",
```

tools/doctest/src/doctest.ts
```diff
@@ -0,0 +1,338 @@
...
+    type: 'string',
+  })
+  .array('files')
+  .version(false)
+  .help()
+  .parse();
+
+for (const pattern of files) {
+  for await (const file of new Glob(pattern, {})) {
+    await test(file, async context => {
```
Lightning00Blade: Nit: lets use `it` its more consistent with what we use else where.
Also maybe wrap this in Describe to be more obvious what that we are using Node:test
To reduce nesting and complexity I would extract the `test` callback to helper.

tools/doctest/src/doctest.ts
```diff
@@ -0,0 +1,331 @@
...
+      if (!description) {
+        continue;
+      }
+      const lines = description.split('\n');
+      const blocks = [];
+      let context: Context | undefined;
+      for (let i = 0; i < lines.length; i++) {
+        const line = lines[i]!;
+        const borderIndex = line.indexOf(CODE_FENCE);
+        if (borderIndex !== -1) {
```
Lightning00Blade: Inverse this into a continue else the code becomes super nested and hard to read.

tools/doctest/src/doctest.ts
```diff
@@ -0,0 +1,331 @@
...
+          if (!context) {
+            const [tag, ...options] = line
+              .slice(borderIndex + CODE_FENCE.length)
+              .split(' ');
+            if (options.includes('ignore')) {
```
Lightning00Blade: Nit: extract to const.

tools/doctest/src/doctest.ts
```diff
@@ -0,0 +1,331 @@
...
+              // Ignore the code sample.
+              continue;
+            }
+            const fail = options.includes('fail');
+            if (!tag || tag.match(/js|javascript/)) {
```
Lightning00Blade: I think we should just brake on undefined language. 
I would be more keen on having an EsLint rule to enforce that rather then doing it here.

tools/doctest/src/doctest.ts
```diff
@@ -0,0 +1,331 @@
...
+              context = {language: Language.JavaScript, fail, start: i + 1};
```
Lightning00Blade: Add a comment on why the `i + 1` it seems hard to follow. (I presume it source map related)

tools/doctest/package.json
jrandolf: It doesn't work with TS files, only JS files.

tools/doctest/src/doctest.ts
jrandolf: I already tried this. We can't because `it` has a different type for some reason in `@types/node`...

tools/doctest/src/doctest.ts
jrandolf: I think it's reasonable to assume code blocks are in JS since it's in JS.

tools/doctest/src/doctest.ts
jrandolf: `i+1` is the starting line for the code block (nothing to do with the source map). I wouldn't say adding a comment would help here; if you get to this point, you need to know where the indexes line up.

package.json
```diff
@@ -59,6 +60,12 @@
         "./packages/puppeteer-core:build:docs"
       ]
     },
+    "doctest": {
+      "command": "npx @puppeteer/doctest 'packages/puppeteer-core/lib/esm/**/*.js'",
+      "dependencies": [
+        "./packages/puppeteer-core:build"
```
Lightning00Blade: Don't you want to make sure the package is build?

tools/doctest/package.json
Lightning00Blade: It's not about files you are pointing out JSDoc which does not include TSDoc, which is an extension on top of JSDoc.

tools/doctest/src/doctest.ts
Lightning00Blade: Yeah I think you need to warp it inside describe for it to work. Also comment on the other parts as well please.

tools/doctest/src/doctest.ts
```diff
@@ -0,0 +1,335 @@
...
+  return generator.toJSON();
+}
+
+async function compileCode(language: Language, code: string) {
+  return (await transform(code, {
+    sourceMaps: true,
+    inlineSourcesContent: false,
+    jsc: {
+      parser: {
+        syntax: (
```
Lightning00Blade: Extract into a variable and then reference it.

tools/doctest/src/doctest.ts
```diff
@@ -0,0 +1,335 @@
...
+              {
+                ...position,
+                extracted:
+                  context.code.split('\n').length +
+                  context.positions.at(-1)!.extracted -
+                  // We subtract this because of the accumulated '\n'.
+                  (index - 1),
+              },
+            ],
+            fail: context.fail || fail,
```
Lightning00Blade: Doesn't this imply that if we have 2 blocks on the same comment they can't be ` typescript ` and ` typescript fail` as both will be expected to fail?

package.json
jrandolf: It's built during `npm i`

tools/doctest/src/doctest.ts
jrandolf: Why? It's only used in one place.

tools/doctest/package.json
jrandolf: Considering the parser (doctrine) doesn't really support TSDoc, I think we should just advocate JSDoc.

tools/doctest/src/doctest.ts
jrandolf: Since the blocks get concatenated, the code will fail; so yes.

package.json
Lightning00Blade: Two things to note:
1) `--ignore-scripts` - sometimes I use it.
2) If i change the package itself and run this command it will use the older version, and I think the consensus on the topic is we want it to rebuild in necessary like other test or tools.

tools/doctest/src/doctest.ts
Lightning00Blade: So the idea is you are using it as a map then getting a single value, having a variable makes it easy to understand intention to do.

package.json
jrandolf: The problem with `npx` is that it only works if the binary is symlinked in `node_modules/.bin`. This implies if the package was not built, then `npm` has nothing to symlink. We could just do `tools/doctest/bin/doctest.js`, but this creates a dependency on the file name which is what I am preventing.

Open to suggestions.

jrandolf: Solved.

tools/doctest/src/doctest.ts
Lightning00Blade: Just to be entirely sure (`.` for ` ):
```
@example
...ts
...

...ts fail
...
``` 
But 
```
@example
...ts
...
@example
...ts fail
...
``` 
work right?

tools/doctest/src/doctest.ts
jrandolf: I misread your comment. I thought you meant the entire option (not the map). GitHub doesn't show the map in the preview.

tools/doctest/src/doctest.ts
jrandolf: Indeed.

tools/doctest/src/doctest.ts
Lightning00Blade: This was resolve but no action or comment left.

Lightning00Blade (approved):

tools/doctest/src/doctest.ts
jrandolf: As discussed offline, we will just keep it as is.

PR #10943, opened 2023-09-19
chore: use custom disposable stack

jrandolf-google:
This PR removes the polyfill in favor a custom implementation. The implementation is not very large, so the maintenance cost it minimal.


OrKoN (approved):
LGTM

Issue #10892, opened 2023-09-12
[Bug]: Errors thrown after doing history restoration on page with oopifs/workers


adamraine:
### Minimal, reproducible example

```TypeScript
import puppeteer from 'puppeteer';

const browser = await puppeteer.launch({
  headless: false,
  defaultViewport: null,
  ignoreDefaultArgs: ['--enable-automation'],
  executablePath: process.env.CHROME_PATH,
});
const page = await browser.newPage();
const session = await page.target().createCDPSession();

await session.send('Page.enable');

await page.goto('http://localhost:10200/oopif-scripts.html', {
  waitUntil: ['networkidle0'],
});

await new Promise(r => setTimeout(r, 500));

const history = await session.send('Page.getNavigationHistory');
const entry = history.entries[history.currentIndex];

await session.send('Page.navigate', {url: 'chrome://terms'});

await new Promise(r => setTimeout(r, 500));

await session.send('Page.navigateToHistoryEntry', {entryId: entry.id});

await session.send('Page.disable');

await page.close();
await browser.close();
```

### Error string

Protocol error (Network.enable): Target closed

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

This is a minified version of how Lighthouse tests the bfcache: https://github.com/GoogleChrome/lighthouse/blob/main/core/gather/gatherers/bf-cache-failures.js#L89

The latest release of puppeteer (21.2.0) has broken this behavior on our test page [`oopif-scripts.html`](https://github.com/GoogleChrome/lighthouse/blob/main/cli/test/fixtures/oopif-scripts.html). All the code needed to serve this test page can be found under the `fixtures` directory that contains `oopif-scripts.html`.

### Expectation

I expected the code to run without throwing any errors or emitting uncaught exceptions. I verified that this is still the behavior for version 21.1.1.

### Reality

The script throws an error.

### Puppeteer version

21.2.0

### Node version

18.17.0

### Package manager

yarn

### Package manager version

1.22.19

### Operating system

macOS

OrKoN:
@adamraine I am unable to reproduce with the bundled version of Chrome for Testing. Do you test with a different Chrome version? Or is it flaky for you?

adamraine:
This is happening with all Chrome version for me. I was using Chrome canary as well as the Chrome that ships in puppeteer.

[...]

PR #10946, opened 2023-09-19
fix: Always set target for attached sessions

adamraine:
In Lighthouse we are still getting uncaught exceptions "Target must exist" during our bfcache testing similar to https://github.com/puppeteer/puppeteer/issues/10892. Unfortunately I haven't created a minimum repro yet, here is what I do know:

- This happens for CDP sessions created on web worker targest
- The error is emitted because the session never has `_setTarget` called, this PR addresses that problem
- This happens as a result of the history navigation that restores from bfcache. Again similar to #10892 but the repro in that bug will not reproduce this.
  - Speculation: When the history navigation is performed, a new CDP session is created for the web worker. Since it's a bfcache restoration the target is the same but the session never has `_setTarget` called because we do not call into the target factory.

Please let me know if this needs more investigation.

### DIFF
--- a/packages/puppeteer-core/src/cdp/ChromeTargetManager.ts
+++ b/packages/puppeteer-core/src/cdp/ChromeTargetManager.ts
@@ -32,6 +32,7 @@ import {
   TargetManagerEvent,
   type TargetManagerEvents,
 } from './TargetManager.js';
+import {CdpCDPSession} from './CDPSession.js';
 
 function isTargetExposed(target: CdpTarget): boolean {
   return target.type() !== TargetType.TAB && !target._subtype();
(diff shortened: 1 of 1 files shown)


OrKoN:
@adamraine do you have a repro we could add to the tests? I am not sure that ChromeTargetManager is the right place to set the target here. I have learned that bfcache restoration does not cause a new session to be created so I wonder how exactly the session gets no target. We have a test for bfcache case now https://github.com/puppeteer/puppeteer/blob/main/test/src/bfcache.spec.ts how can we extend it with the situation you are observing?

adamraine:
https://github.com/puppeteer/puppeteer/issues/10966

OrKoN:
Thanks for the fix. Closing this PR in favour of https://github.com/puppeteer/puppeteer/pull/10967

PR #10961, opened 2023-09-20
chore: BiDi setCacheEnabled

Lightning00Blade:


packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -728,6 +728,12 @@ export class BidiPage extends Page {
   override isDragInterceptionEnabled(): boolean {
     return false;
   }
+
+  override async setCacheEnabled(enabled?: boolean): Promise<void> {
+    await this._client().send('Network.setCacheDisabled', {
```
OrKoN, line 750: Let's add a todo like in other places (as we are doing CDP directly so it won't work for mparch)

OrKoN (approved):

packages/puppeteer-core/src/bidi/Page.ts
Lightning00Blade, line 750:
```suggestion
    // TODO: handle CDP-specific cases such as mprach.
    await this._client().send('Network.setCacheDisabled', {
```

PR #10963, opened 2023-09-20
chore: implement element screenshot

jrandolf-google:


Lightning00Blade (approved):

packages/puppeteer-core/src/cdp/ElementHandle.ts
```diff
@@ -162,66 +157,6 @@ export class CdpElementHandle<
     }
   }
 
-  @throwIfDisposed()
-  @ElementHandle.bindIsolatedHandle
```
OrKoN, line 166: let's not introduce breaking (or any changes) to the CDP implementation? I don't want us to spend time tracing diffs in DevTools test cases

jrandolf: There shouldn't be any difference in CDP. The new implementation is the same as the old one.

test/golden-chrome/screenshot-element-fractional-offset.png
Lightning00Blade, line 1: This look like a regression, can you comment on it?

PR #10964, opened 2023-09-20
test: enable queryObjects for bidi

OrKoN:


packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -219,6 +220,26 @@ export class BidiPage extends Page {
...
 
+  override async queryObjects<Prototype>(
+    prototypeHandle: BidiJSHandle<Prototype>
+  ): Promise<BidiJSHandle<Prototype[]>> {
+    assert(!prototypeHandle.disposed, 'Prototype JSHandle is disposed!');
+    assert(
+      prototypeHandle.id,
+      'Prototype JSHandle must not be referencing primitive value'
+    );
+    const response = await this.mainFrame().client.send(
```
OrKoN, line 231: note that we don't worry about mparch here because the state is connected to the main frame's realm

Lightning00Blade (approved):

Issue #10966, opened 2023-09-20
[Bug]: "Target must exist" during bfcache restoration


adamraine:
### Minimal, reproducible example

```TypeScript
import puppeteer from 'puppeteer';

process.on('unhandledRejection', console.error);

// This is important.
// Connect to a Chrome launched normally with `--remote-debugging-report=9222`
// or a Chrome launched via `chrome-launcher` package.
//
// The error does not get emitted if Chrome is launched via puppeteer.
const browser = await puppeteer.connect({
  browserURL: `http://127.0.0.1:9222`,
})

const page = await browser.newPage();

await page.goto('http://localhost:8080/oopif');

await new Promise(r => setTimeout(r, 500));

await page.goto('chrome://terms');

await new Promise(r => setTimeout(r, 500));

await page.goBack();

await new Promise(r => setTimeout(r, 500));

await page.close();
await browser.close();
```

### Error string

Error: Target must exist

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

https://github.com/puppeteer/puppeteer/issues/10892
https://github.com/puppeteer/puppeteer/pull/10946

This error happens consistently in Lighthouse and in the repro script for me.

The anatomy of http://localhost:8080/oopif is a page containing an OOPIF, and the OOPIF contains a web worker:

`index.html` (using `127.0.0.1:8081` to force the iframe to be OOPIF):
```html
<iframe src="http://127.0.0.1:8081/worker.html"></iframe>
```

`worker.html`:
```html
<script>
  const worker = new Worker('/worker.mjs', {type: 'module'})
</script>
```

`worker.mjs`:
```js
console.log('HELLO');
```

### Expectation

No uncaught exceptions are emitted

### Reality

Uncaught exception is emitted "Target must exist"

### Puppeteer version

21.3.1

### Node version

18.17.0

### Package manager

yarn

### Package manager version

1.22.19

### Operating system

macOS

PR #10967, opened 2023-09-20
fix: handle missing detach events for restored bfcache targets

OrKoN:
Currently, if we navigate from a page with a worker in an OOPIF to another, we don't get a target detached event for the worker. The session and the target, therefore, continues to be registered as "attached" by Puppeteer. This leads to an issue when on going back to the same page the worker target is double attached. This behaviour is probably caused by the fact that the page is put into a bfcache instead of being actually destroyed. Probably, the correct behaviour from the CDP side would be to keep the client attached to the cached session instead of creating new sessions. Currently, this is not the case so I filed https://crbug.com/1485193.

Closes #10946 #10966

### DIFF
--- a/packages/puppeteer-core/src/cdp/ChromeTargetManager.ts
+++ b/packages/puppeteer-core/src/cdp/ChromeTargetManager.ts
@@ -24,6 +24,7 @@ import {debugError} from '../common/util.js';
 import {assert} from '../util/assert.js';
 import {Deferred} from '../util/Deferred.js';
 
+import {type CdpCDPSession} from './CDPSession.js';
 import {type Connection} from './Connection.js';
 import {CdpTarget, InitializationStatus} from './Target.js';
 import {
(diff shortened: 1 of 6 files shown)


packages/puppeteer-core/src/cdp/ChromeTargetManager.ts
```diff
@@ -358,10 +359,7 @@ export class ChromeTargetManager
     // `this.#connection.isAutoAttached(targetInfo.targetId)`. In the future, we
     // should determine if a target is auto-attached or not with the help of
     // CDP.
-    if (
-      targetInfo.type === 'service_worker' &&
-      this.#connection.isAutoAttached(targetInfo.targetId)
-    ) {
+    if (targetInfo.type === 'service_worker') {
```
OrKoN, line 362: unrelated change: this condition is redundant because of the check on line 351 (improved for readability)

Lightning00Blade (approved):

PR #10968, opened 2023-09-20
Update/Fix Dockerfile

jake5253:
Ignore missing files on the `COPY` line
use `npm i puppeteer` to install instead of the missing local files.

Allows to build docker container without the 3 files that seem to be missing, and appear to be unnecessary(?)
The files: puppeteer-browsers-latest.tgz puppeteer-latest.tgz puppeteer-core-latest.tgz

No. But the already-existing `docker/test/smoke-test.js` returns without error

Shouldn't be needed

**Summary**
I was unable to build the docker container locally without the 3 files that are missing from the COPY line in the Dockerfile.  I realized those files don't need to exist as puppeteer is available for installation from npm directly -- even the latest version (at time of writing, version 21.3.1)

I am not sure if there is any open issues for this. TBH I didn't check.

None that I'm aware of.

**Other information**
None

Thank you for your consideration

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,11 +17,13 @@ USER pptruser
 
 WORKDIR /home/pptruser
 
-COPY puppeteer-browsers-latest.tgz puppeteer-latest.tgz puppeteer-core-latest.tgz ./
+# Files don't exist and seem to be unneccesary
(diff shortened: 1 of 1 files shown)


OrKoN (changes requested):
Thanks but we are releasing the docker image at the same time as npm packages so the matching version is not on NPM.

PR #10973, opened 2023-09-21
test: detect dependency cycles

OrKoN:


.eslintrc.js
```diff
@@ -111,8 +117,6 @@ module.exports = {
         ],
       },
     ],
-    'import/extensions': ['error', 'ignorePackages'],
```
OrKoN, line 114: import rules were not actually working without a resolver. This complains about the missing ts extension everywhere (we can re-add it later)

Lightning00Blade (approved):

.eslintrc.js
```diff
@@ -121,6 +125,8 @@ module.exports = {
       },
     ],
 
+    'import/no-cycle': ['error', {maxDepth: Infinity}],
```
OrKoN:
```suggestion
    'import/no-cycle': ['warn', {maxDepth: Infinity}],
```

PR #10974, opened 2023-09-21
refactor: remove import cycles in browsers

OrKoN:


packages/browsers/src/Cache.ts
```diff
@@ -159,6 +182,27 @@ export class Cache {
         });
     });
   }
+
+  computeExecutablePath(options: ComputeExecutablePathOptions): string {
```
Lightning00Blade, line 186: Does this mean we have a duplication of this functionality on the Cache and in `launch.ts`?
Could `launch.ts` reexport this if it were static method?

OrKoN: ah yes, forgot to update that part.

Lightning00Blade (approved):

PR #10995, opened 2023-09-22
chore: remove legacy checks in keyboard tests

### LINKED ISSUE #10950: [Bug]: Unit test "[keyboard.spec] Keyboard should report multiple modifiers" should not use legacy properties of key events

```TypeScript
See details in the background section for the affected Puppeteer unit test.
```

### Error string

Expected: \"Keydown: ; Semicolon 186 [Alt Control]\"\nReceived: \"Keydown: ; Semicolon 59 [Alt Control]\

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

The mentioned test uses legacy properties for key events which are not standardized and differ between browsers. Those properties are `.which` and `.charCode`:

https://github.com/puppeteer/puppeteer/blob/d98f488a514ff9b3bb06923b4fb2d9bfb5919cec/test/src/keyboard.spec.ts#L257-L298
https://github.com/puppeteer/puppeteer/blob/d98f488a514ff9b3bb06923b4fb2d9bfb5919cec/test/assets/input/keyboard.html#L12-L20

The test should be changed to only use standardized properties like `.key`, `.code` and maybe `.data` to verify that the right keys were pressed.

https://dvcs.w3.org/hg/d4e/raw-file/tip/key-event-test.html is a good help here.

@OrKoN can you please have a look? Thanks!

### Expectation

Test has to pass in Firefox as well and not fail because legacy properties are used which on purpose are different between browsers.

### Reality

The test fails because of inappropriate checks.

_No response_

### Puppeteer version

latest

### Node version

latest

### Package manager

npm

### Package manager version

latest

### Operating system

macOS

jrandolf-google:
Fixed: https://github.com/puppeteer/puppeteer/issues/10950

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -120,6 +120,12 @@
     "expectations": ["PASS"]
   },
   {
+    "testIdPattern": "[keyboard.spec] Keyboard *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["webDriverBiDi"],
+    "expectations": ["PASS"]
+  },
+  {
     "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Puppeteer.launch *",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["webDriverBiDi"],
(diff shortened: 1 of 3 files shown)


test/src/keyboard.spec.ts
```diff
@@ -338,10 +313,10 @@ describe('Keyboard', function () {
       })
     ).toBe(
       [
-        'Keydown: Shift ShiftLeft 16 [Shift]',
-        'Keydown: ~ Backquote 192 [Shift]', // 192 is ` keyCode
-        'Keypress: ~ Backquote 126 126 [Shift]', // 126 is ~ charCode
-        'Keyup: ~ Backquote 192 [Shift]',
+        'Keydown: Shift ShiftLeft [Shift]',
+        'Keydown: ~ Backquote [Shift]', // 192 is ` keyCode
```
Lightning00Blade: Redundant comments

test/src/keyboard.spec.ts
```diff
@@ -499,39 +472,27 @@ describe('Keyboard', function () {
     ).toBe('👹 Tokyo street Japan 🇯🇵');
   });
   it('should press the meta key', async () => {
-    const {page, isFirefox} = await getTestState();
+    if (os.platform() === 'darwin') {
```
Lightning00Blade: Do inverse + return easier to read.

test/src/keyboard.spec.ts
```diff
@@ -42,19 +42,19 @@ describe('Keyboard', function () {
     ).toBe(text);
   });
   it('should press the metaKey', async () => {
-    const {page, isFirefox} = await getTestState();
+    if (os.platform() === 'darwin') {
```
Lightning00Blade: Same here as bellow.

test/TestExpectations.json
```diff
@@ -642,100 +648,16 @@
...
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["webDriverBiDi"],
-    "expectations": ["PASS"]
-  },
-  {
-    "testIdPattern": "[keyboard.spec] Keyboard should type into a textarea",
-    "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["webDriverBiDi"],
-    "expectations": ["PASS"]
+    "expectations": ["FAIL"]
```
Lightning00Blade: Why is this failing now?

OrKoN (changes requested):
The diff shows unrelated changes and merge conflicts.

test/TestExpectations.json
jrandolf: This was meant to only fail firefox

OrKoN (approved):

PR #11000, opened 2023-09-22
chore: implement BiDi sendCharacter

jrandolf-google:


packages/puppeteer-core/src/bidi/Input.ts
```diff
@@ -415,6 +415,15 @@ export class BidiKeyboard extends Keyboard {
       ],
     });
   }
+
+  override async sendCharacter(char: string): Promise<void> {
+    if ([...char].length > 1) {
```
Lightning00Blade, line 422:
```suggestion
    if (char.length > 1) {
```

packages/puppeteer-core/src/bidi/Input.ts
```diff
@@ -415,6 +415,15 @@ export class BidiKeyboard extends Keyboard {
...
+      throw new Error('Cannot send more than 1 character.');
+    }
+    await this.#context.evaluate(async char => {
```
OrKoN: this probably needs to go via the sandbox (let's add a todo?)

packages/puppeteer-core/src/bidi/Input.ts
jrandolf: [...char].length measures the number of UTF-32 unicode characters.
char.length measures the UTF-16 length.

Lightning00Blade (approved):

OrKoN (approved):

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -282,6 +290,27 @@ export class BidiPage extends Page {
...
+  /**
+   * @internal
+   */
+  async focusedFrame(): Promise<BidiFrame> {
+    using frame = await this.mainFrame()
+      .isolatedRealm()
+      .evaluateHandle(() => {
+        let frame: HTMLIFrameElement | undefined;
+        let win: Window | null = window;
+        while (win?.document.activeElement instanceof HTMLIFrameElement) {
```
OrKoN, line 302: this does not work for OOPIFs.

PR #11006, opened 2023-09-22
fix: avoid structuredClone for Node 16

### LINKED ISSUE #11004: [Bug]: Error with missing `structuredClone` API in Node 16 when doing a screenshot

```TypeScript
import puppeteer from 'puppeteer'; // TS/ESM are all supported.

global.structuredClone = function () {
  throw Error('Use of structuredClone (which does not exist in Node 16)');
};

(async () => {
  const result = await page.screenshot();
  console.log(result);
  await browser.close();
})();
```

### Error string

Use of structuredClone (which does not exist in Node 16)

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

https://developer.mozilla.org/en-US/docs/Web/API/structuredClone#browser_compatibility

### Expectation

The code *was* working (puppeteer@20) but it recently broke.

### Reality

I know Node 16 is on the verge of being deprecated but I didn't see any breaking change telling it is no more supported.

_No response_

### Puppeteer version

21.3.3

### Node version

16.20.2

### Package manager

npm

### Package manager version

8.19.4

### Operating system

Windows

OrKoN:
Although we drop Node 16 support soon, we should keep the code compatible with Node 16 due to semver.

Closes #11004

### DIFF
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -2277,7 +2277,15 @@ export abstract class Page extends EventEmitter<PageEvents> {
   ): Promise<Buffer | string> {
     await this.bringToFront();
 
-    const options = structuredClone(userOptions) as ScreenshotOptions;
+    // TODO: use structuredClone after Node 16 support is dropped.«
+    const options = {
+      ...userOptions,
+      clip: userOptions.clip
+        ? {
+            ...userOptions.clip,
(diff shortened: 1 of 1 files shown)


OrKoN:
@jrandolf I believe Chromium might still be on Node 16. Let's publish this as a patch release.

jrandolf-google (approved):

PR #11008, opened 2023-09-22
test: test sharding for firefox

OrKoN:
This PR introduces sharding of tests for Firefox. Currently, Firefox tests are the slowest easily reaching 15 minutes. We can split the tests into 4 or more shards and get better results (comparable to chrome headless on linux) as long as we have free concurrency slots in GitHub Actions. We could also shard the Chrome jobs except for the old headless mode that has reasonable runtime. 

Obviously, sharding has drawbacks:

1) tests can have different failures with and without sharding and with different sharding configuration if the tests are not sufficiently isolated.
2) sharding is based on files (and not tests) so the number of tests per shard differs (it's possible to work around that but it is not trivial).
3) we cannot check the total number of tests as we are missing the "reduce" step. If we hit failures that prevent discovery of tests we might get green results due to missing tests.
4) not really faster if many github actions jobs are queuing

Advantages:
1) feedback time is reduced
2) if a failure is in a certain shard and it is shard-specific, it's faster to reproduce the failure locally

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -273,7 +273,7 @@ jobs:
         run: npm run postinstall
       - name: Run all tests (for non-Linux)
         if: ${{ matrix.os != 'ubuntu-latest' }}
-        run: npm run test -- --shard ${{ matrix.shard }} --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
+        run: npm run test -- --shard '${{ matrix.shard }}' --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
       - name: Run all tests (for Linux)
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: xvfb-run --auto-servernum npm run test -- --shard '${{ matrix.shard }}' --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -220,7 +220,7 @@ jobs:
       - run: 'exit 0'
 
   firefox-tests:
-    name: ${{ matrix.suite }} tests on ${{ matrix.os }} ${{ matrix.shard }}
+    name: ${{ matrix.suite }} tests on ${{ matrix.os }} (${{ matrix.shard }})
     runs-on: ${{ matrix.os }}
     needs: check-changes
     if: ${{ contains(fromJSON(needs.check-changes.outputs.changes), 'puppeteer') }}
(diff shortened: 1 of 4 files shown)


OrKoN:
TODO: need to investigate new failures in firefox tests (probably caused by leakage of test state)

OrKoN:
TODO: we can discuss the CLI format if someone is not happy with the duplication of the total number of shards.

.github/workflows/ci.yml
```diff
@@ -220,7 +220,7 @@ jobs:
       - run: 'exit 0'
 
   firefox-tests:
-    name: ${{ matrix.suite }} tests on ${{ matrix.os }}
+    name: ${{ matrix.suite }} tests on ${{ matrix.os }} ${{ matrix.shard }}
```
jrandolf:
```suggestion
    name: ${{ matrix.suite }} tests on ${{ matrix.os }} (${{ matrix.shard }})
```

.github/workflows/ci.yml
```diff
@@ -268,10 +273,10 @@ jobs:
         run: npm run postinstall
       - name: Run all tests (for non-Linux)
         if: ${{ matrix.os != 'ubuntu-latest' }}
-        run: npm run test -- --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
+        run: npm run test -- --shard ${{ matrix.shard }} --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
```
jrandolf:
```suggestion
        run: npm run test -- --shard '${{ matrix.shard }}' --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
```

.github/workflows/ci.yml
```diff
@@ -268,10 +273,10 @@ jobs:
...
       - name: Run all tests (for Linux)
         if: ${{ matrix.os == 'ubuntu-latest' }}
-        run: xvfb-run --auto-servernum npm run test -- --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
+        run: xvfb-run --auto-servernum npm run test -- --shard ${{ matrix.shard }} --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
```
jrandolf:
```suggestion
        run: xvfb-run --auto-servernum npm run test -- --shard '${{ matrix.shard }}' --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
```

.github/workflows/ci.yml
jrandolf: Just to make sure the value is passed in as an uncomputed string

jrandolf-google (approved):

Lightning00Blade (approved):
LGTM,
Note: 
For point `3)` I think we can do the reduce inside the matrix required step. if we export the the results.

PR #11022, opened 2023-09-25
chore: update expected bidi errors

OrKoN:
We can consider this errors coming from Firefox's BiDi implementation as errors that require rerunning the wait tasks. We need to replace it with the standard message once https://github.com/w3c/webdriver-bidi/issues/540 lands.

### DIFF
--- a/packages/puppeteer-core/src/common/WaitTask.ts
+++ b/packages/puppeteer-core/src/common/WaitTask.ts
@@ -231,6 +231,15 @@ export class WaitTask<T = unknown> {
         return;
       }
 
+      // Errors coming from WebDriver BiDi. TODO: Adjust messages after
+      // https://github.com/w3c/webdriver-bidi/issues/540 is resolved.
+      if (
+        error.message.includes('no such handle') ||
+        error.message.includes("Actor 'MessageHandlerFrame' destroyed")
+      ) {
+        return;
+      }
+
       return error;
     }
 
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/common/WaitTask.ts
```diff
@@ -225,6 +225,12 @@ export class WaitTask<T = unknown> {
         return;
       }
 
+      // Errors coming from WebDriver BiDi. TODO: Adjust messages after
+      // https://github.com/w3c/webdriver-bidi/issues/540 is resolved.
+      if (error.message.includes("Actor 'MessageHandlerFrame' destroyed")) {
```
whimboo: I would include the `AbortError: ` part here to really only let it apply for IPC issues when trying to send the response in Firefox.

Lightning00Blade (approved):

OrKoN:
Looks like there could be more variations in errors. Let's wait till the error has been spec'ed and well defined.

packages/puppeteer-core/src/common/WaitTask.ts
```diff
@@ -231,6 +231,16 @@ export class WaitTask<T = unknown> {
         return;
       }
 
+      // Errors coming from WebDriver BiDi. TODO: Adjust messages after
+      // https://github.com/w3c/webdriver-bidi/issues/540 is resolved.
+      if (
+        error.message.includes(
```
Lightning00Blade, line 241: There is the same message a few line above with a comment, pointing to the same issue.  
Let's close this PR.

Lightning00Blade (approved):

OrKoN:
@whimboo mentioned that he will take a look at the failing tests

whimboo:
> @whimboo mentioned that he will take a look at the failing tests

Am I supposed to apply this patch for further investigation, or is this PR blocked on my findings? If it's not blocking I would wait until this PR has been landed.

PR #11024, opened 2023-09-25
chore: use RxJS for wait for Navigation

Lightning00Blade:


packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -201,48 +210,32 @@ export class BidiFrame extends Frame {
...
           this.#context,
-          waitUntilEvent,
-          () => {
-            return true;
-          },
-          timeout,
-          this.#abortDeferred.valueOrThrow()
+          Bidi.ChromiumBidi.BrowsingContext.EventNames.NavigationStarted
+        ).pipe(
+          switchMap(() => {
```
OrKoN, line 226: how does it work? would it wait for NavigationStarted and the waitUntilEvent event after that? or would it be more like the promise.all in the previous version

OrKoN: the spec guarantees that that the waitUntil will happen after the nav started, right?

Lightning00Blade: Yeah switch map creates a new observable each time the previous fires an invalidates the previous.
I will have to look into the spec but I believe (Both events are defined on the HTML spec so should be clear what the order is).

Lightning00Blade: Looked at the HTML spec the navigation started event is emitted very early on the navigation lifecycle before even the `HTML's` response has completed.

OrKoN (approved):

PR #11026, opened 2023-09-25
chore: remove screenshot options for firefox

jrandolf-google:


OrKoN:
eslint is failing

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -1052,11 +1053,17 @@ export class CdpPage extends Page {
...
     } = options;
 
+    const isFirefox =
+      this.target()._targetManager() instanceof FirefoxTargetManager;
+
     await using stack = new AsyncDisposableStack();
-    if (omitBackground && (type === 'png' || type === 'webp')) {
+    // Firefox omits background by default; it's not configurable.
+    if (!isFirefox && omitBackground && (type === 'png' || type === 'webp')) {
       await this.#emulationManager.setTransparentBackgroundColor();
```
OrKoN, line 1062: can we catch the error on the transparent color call instead of checking for Firefox?

OrKoN (approved):

packages/puppeteer-core/src/cdp/Page.ts
jrandolf: I think your method is more "robust", but it's not like we will support other CDP clients.

jrandolf: In particular, the error will only exist in firefox, so why not just match by checking targetmanager?

PR #11029, opened 2023-09-25
refactor: use RxJS for waitForTarget

Lightning00Blade:
This implementation works with `CDP` but there seems to be some issue with `BiDi` where even though we have `3000` ms timeout it never gets triggered.

The issue above is related to how BiDi mapper implementation waits for 2 events (`requestWillBeSend` and `requestWillBeSentExtraInfo`, or `responseRecivedEvent` which never comes).

### DIFF
--- a/docs/api/puppeteer.browsercontext.waitfortarget.md
+++ b/docs/api/puppeteer.browsercontext.waitfortarget.md
@@ -14,9 +14,7 @@ This will look all open [browser contexts](./puppeteer.browsercontext.md).
 class BrowserContext {
   abstract waitForTarget(
     predicate: (x: Target) => boolean | Promise<boolean>,
-    options?: {
-      timeout?: number;
-    }
+    options?: WaitForTargetOptions
   ): Promise<Target>;
 }
 ```
@@ -26,7 +24,7 @@ class BrowserContext {
 | Parameter | Type                                                                         | Description  |
 | --------- | ---------------------------------------------------------------------------- | ------------ |
 | predicate | (x: [Target](./puppeteer.target.md)) =&gt; boolean \| Promise&lt;boolean&gt; |              |
-| options   | { timeout?: number; }                                                        | _(Optional)_ |
+| options   | [WaitForTargetOptions](./puppeteer.waitfortargetoptions.md)                  | _(Optional)_ |
 
 **Returns:**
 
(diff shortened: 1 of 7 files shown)


OrKoN (approved):

sspread:
@Lightning00Blade Hello - After upgrading puppeteer and using `headess: 'new'`, I'm seeing error `Target.createTarget timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.` when running on heroku (linux). Unable to reproduce, never seen running locally (mac arm). Issue is seemingly random. Doesn't occur with `headless: true`. Could this PR be related to the issue?

OrKoN:
@sspread it does not sound related (this PR has not landed yet). It sounds like in your case the browser crashes/hangs. Try adding the dumpio: true flag to see what's going on. The best is to file the issue with a minimal reproducible example so that we could take a look.

sspread:
@OrKoN Thanks I'll take a look at the logs. Its only happening on heroku, and I can't reliably reproduce it there either, but I'll post something if i can. I understand the PR is not yet merged/deployed - I was hoping perhaps it would fix the problem.

OrKoN:
@sspread if it only happens on heroku it probably missing some dependencies/permissions as the new headless mode is more like the regular chrome (e.g., it needs dbus access and a few other libs).

sspread:
@OrKoN indeed dumpio reveals many logs like `[14687:14702:0927/172950.508873:ERROR:bus.cc(406)] Failed to connect to the bus: Failed to connect to socket /run/dbus/system_bus_socket: No such file or directory` and `[14687:14699:0927/172950.512930:ERROR:bus.cc(406)] Failed to connect to the bus: Could not parse server address: Unknown address type (examples of valid types are "tcp" and on UNIX "unix")`. While these errors are constantly logging, the `createTarget` timeout error is relatively infrequent. I'm using buildpack `https://github.com/jontewks/puppeteer-heroku-buildpack`, which apparently installs `dbus_1.12.20-2ubuntu4.1_amd64.deb`, but maybe the buildpack is not up to date.

sspread:
@Lightning00Blade @OrKoN I filed issue with buildpack repo, but I suspect those dbus errors are a red herring. I'm at a loss here, but I look forward to this PR merging - I still think its related, as its the only place in the codebase I can find a reference to `createTarget`.

jrandolf-google (approved):

PR #11033, opened 2023-09-26
test: move cdp only tests to a subfolder

OrKoN:
And allow filtering to exclude cdp only tests.


tools/mocha-runner/src/mocha-runner.ts
```diff
@@ -79,6 +79,7 @@ function getApplicableTestSuites(
 async function main() {
   const noCoverage = process.argv.indexOf('--no-coverage') !== -1;
   const noSuggestions = process.argv.indexOf('--no-suggestions') !== -1;
+  const excludeCDPOnly = process.argv.indexOf('--no-cdp-only') !== -1;
```
Lightning00Blade:
```suggestion
  const excludeCDPOnly = process.argv.indexOf('--no-cdp-tests') !== -1;
```

Lightning00Blade (approved):

PR #11056, opened 2023-09-29
chore: refactor file chooser

jrandolf-google:
This PR refactors file chooser so that `ElementHandle.uploadFile` can be cross-protocol compatible.

Blocked-by: https://chromium-review.googlesource.com/c/chromium/src/+/4906156


Lightning00Blade:
@jrandolf what is the status of this?

PR #11057, opened 2023-09-29
fix: update file chooser events

jrandolf-google:
This PR

 - adds `composed` to the input event since this is what occurs in the browser.
 - dispatches the `cancel` event when the chooser is canceled.


OrKoN (approved):

jrandolf:
See https://html.spec.whatwg.org/multipage/input.html#show-the-picker,-if-applicable

PR #11084, opened 2023-10-06
feat: implement screencasting

### LINKED ISSUE #478: startScreencast feature?

jrandolf-google:
## Summary

Screencasting is a new, experimental feature that allows users to record their flows as that run!

> This requires [ffmpeg](https://ffmpeg.org/) to be installed where you run Puppeteer.

## Example

```

// Launch a browser

// Create a new page

// Go to your site.
await page.goto("https://www.example.com");

// Start recording.
const recorder = await page.screencast({path: 'recording.webm'});

// Do something.

// Stop recording.
await recorder.stop();

browser.close();
```

You can also stream the recorder to somewhere else!

```
import fs from 'fs';

// Launch a browser

// Create a new page

// Go to your site.
await page.goto("https://www.example.com");

// Start recording.
const recorder = await page.screencast();

// Piping into /dev/null for illustrative purposes.
recorder.pipe(fs.createWriteStream('/dev/null'));

// Do something.

// Stop recording.
await recorder.stop();

browser.close();
```

## Options

For now, we have three options:

 - `crop` - Specifies the region of the viewport to crop.
 - `scale` - Scales the output video. For example, `0.5` will shrink the width and height of the output video by half. `2` will double the width and height of the output video.
 - `path` - Specifies a `.webm` file path to save the recording to.
 - `speed` - Specifies the speed to record at.  For example, `0.5` will slowdown the video by 50%. `2` will double the speed of the output video.

Fixed: https://github.com/puppeteer/puppeteer/issues/478


OrKoN:
@jrandolf could you please rebase?

test/src/screencast.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+function rmIfExists(file: string) {
+  return rm(file).catch(() => {});
+}
+
+describe('Screencasts', function () {
+  setupTestBrowserHooks();
+
+  describe('Page.screencast', function () {
+    it('should work', async () => {
+      const file = getUniqueVideoFilename();
```
OrKoN: want to use `using` for this?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -290,6 +288,52 @@ export interface ScreenshotOptions {
   allowViewportExpansion?: boolean;
 }
 
+/**
+ * @public
+ */
+export interface Rectangle extends Point {
```
OrKoN: should we re-use the BoundingBox types?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -2274,6 +2318,179 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
+   */
+  async _startScreencast(): Promise<void> {
+    ++this.#screencastSessionCount;
+    if (!this.#startScreencastPromise) {
+      this.#startScreencastPromise = this.mainFrame()
+        .client.send('Page.startScreencast', {format: 'png'})
+        .then(() => {
+          // Wait for the first frame.
+          return new Promise(resolve => {
+            return this.mainFrame().client.once('Page.screencastFrame', () => {
```
OrKoN, line 2442: does screencasting work over a prerendering navigation? let's make sure it works as we are going to enable tab target and pre-rendering by default.

OrKoN: I imagine the screencast is not really connected to a frame so perhaps we can enable it for the tab target in CDP

jrandolf: It should, it gets the data directly from the compositor whenever the a frame is painted.

OrKoN: the main frame client is destroyed during the pre-rendering navigation so I think this will not work. Would you mind adding a test to prerendering.spec.ts?

jrandolf: Done. https://github.com/puppeteer/puppeteer/issues/11090

test/src/screencast.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+
+import {getTestState, setupTestBrowserHooks} from './mocha-utils.js';
+import {getUniqueVideoFilePlaceholder} from './utils.js';
+
+describe('Screencasts', function () {
+  setupTestBrowserHooks();
+
+  describe('Page.screencast', function () {
+    it('should work', async () => {
+      using file = getUniqueVideoFilePlaceholder();
```
OrKoN, line 29: nice! no more try/finally

OrKoN (approved):
LGTM

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -2274,6 +2314,179 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
+    if (options.crop) {
+      const {
+        x,
+        y,
+        width: cropWidth,
+        height: cropHeight,
+      } = roundRectangle(normalizeRectangle(options.crop));
+      if (x < 0 || y < 0) {
+        throw new Error(
+          `\`crop.x\` and \`crop.x\` must be greater than or equal to 0.`
```
campersau, line 2373: I guess the scond one should be `crop.y` ?

PR #11099, opened 2023-10-09
feat: enable tab targets

OrKoN:
[Tab target is a new type of target in CDP](https://docs.google.com/document/d/14aeiC_zga2SS0OXJd6eIFj8N0o5LGwUpuqa4L8NKoR4/edit) that reflects the architectural changes in Chromium that were needed for features such as bfcache, portals and prerendering.

Previously, the `page` targets were the top-level targets representing the entire tab. If you navigate to a new URL, the page target would remain the same. With the introduction of the `tab` target, it's not possible to have multiple page targets active within one tab. For example, one page would be the currently shown page and the other one would be a [prerendered](https://github.com/WICG/nav-speculation/) page. When the user navigates to the prerendered page, the secondary page becomes the primary and the primary page target gets destroyed (the process is called activation).

This change enables tab targets and allows testing Chrome features related to prerendering. Emulation and network manager domains should work with the prerendered targets. Some features like JS/CSS coverage might not work during the activation so you would need to disable prerendering to test those.

Use command line flags to disable prerendering in Chrome if you encounter issues. For example,
```
--disable-features=Prerender2
```

### DIFF
--- a/packages/puppeteer-core/src/cdp/ChromeTargetManager.ts
+++ b/packages/puppeteer-core/src/cdp/ChromeTargetManager.ts
@@ -99,21 +99,16 @@ export class ChromeTargetManager
   #waitForInitiallyDiscoveredTargets = true;
 
   // TODO: remove the flag once the testing/rollout is done.
-  #tabMode: boolean;
   #discoveryFilter: Protocol.Target.FilterEntry[];
 
   constructor(
     connection: Connection,
     targetFactory: TargetFactory,
     targetFilterCallback?: TargetFilterCallback,
-    waitForInitiallyDiscoveredTargets = true,
-    useTabTarget = false
+    waitForInitiallyDiscoveredTargets = true
   ) {
     super();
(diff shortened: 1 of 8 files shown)


packages/puppeteer-core/src/cdp/ChromeTargetManager.ts
```diff
@@ -99,21 +99,16 @@ export class ChromeTargetManager
...
-    waitForInitiallyDiscoveredTargets = true,
-    useTabTarget = false
+    waitForInitiallyDiscoveredTargets = true
   ) {
     super();
-    this.#tabMode = useTabTarget;
-    this.#discoveryFilter = this.#tabMode
-      ? [{}]
-      : [{type: 'tab', exclude: true}, {}];
+    this.#discoveryFilter = [{}];
```
Lightning00Blade: Nit: set it outside the constructor, directly when declaring the variable.

Lightning00Blade (approved):

PR #11108, opened 2023-10-10
docs: fix links to archived versions

compact:
**What kind of change does this PR introduce?** Documentation fix

**Did you add tests for your changes?** N/A

**If relevant, did you update the documentation?** Yes

**Summary**

In the documentation site, fix the following links to the archived versions.

1. In the top right menu, the links to these versions were broken since the condition `parts[0] <= 19 && parts[1] <= 2 && parts[2] <= 2` is false when any number > 3. I switched it to use the `semver` library, which is already used in `generate_docs.ts`.
    * 18.0.5
    * 18.0.4
    * 18.0.3
    * 17.1.3
    * 15.5.0
    * 15.4.2
    * 15.4.1
    * 15.4.0
    * 15.3.2
    * 15.3.1
    * 15.3.0
2. In the Chromium Support page https://pptr.dev/chromium-support the links to these versions were broken.
    * 20.0.0 and above, since `https://pptr.dev/20.0.0` isn't working
        image
    * from 19.4.0 to 19.8.0, since the git tag in the url was `v19.8.0` instead of `puppeteer-v19.8.0`
        image
    * 15.1.0 was not linked since the checks were on < 15.0.0 and >= 15.3.0
        image

New screenshot from running locally:

image

**Does this PR introduce a breaking change?** No

**Other information**


website/docusaurus.config.js
```diff
@@ -21,12 +21,24 @@ const assert = require('assert');
 
 const darkCodeTheme = require('prism-react-renderer/themes/dracula');
 const lightCodeTheme = require('prism-react-renderer/themes/github');
+const semver = require('semver');
```
Lightning00Blade, line 24: The `website` package is a separate one, in our CI sometime we install only it's dependencies. 
Currently this works as some other packages in the `webside` dependency graph also require `semver`.
To avoid any future issue let's put it as a `devDependencies` in `website/package.json` (pinned to the latest version) .

website/docusaurus.config.js
```diff
@@ -21,12 +21,24 @@ const assert = require('assert');
...
 
 const archivedVersions = require('./versionsArchived.json');
 
 const DOC_ROUTE_BASE_PATH = '/';
 const DOC_PATH = '../docs';
 
+/**
+ * This logic should match generate_docs.ts.
```
Lightning00Blade, line 32: Nit: Let's add a comment here that the website only supports version after `15` as the logic in the two functions is slightly different. 
`Note: pptr.dev archive list start from version 15`
 
Also reference `tools/generate_docs.ts`.

tools/generate_docs.ts
```diff
@@ -51,6 +51,19 @@ function spliceIntoSection(
   return lines.join('\n');
 }
 
+/**
+ * This logic should match docusaurus.config.ts.
```
Lightning00Blade, line 55: Nit: `website/docusaurus.config.ts`

tools/generate_docs.ts
```diff
@@ -51,6 +51,19 @@ function spliceIntoSection(
...
+ */
+function getUrl(version: string): string {
```
Lightning00Blade, line 57: Let's change this to `getApiUrl` (and the one in `website/docusaurus.config.ts`) to be more explicit.

Lightning00Blade (commented):
Thank you for your PR. I left some comment, after they are addressed I would be happy to merge.

jrandolf:
Fixed in https://github.com/puppeteer/puppeteer/pull/11136

PR #11110, opened 2023-10-10
chore: small fixes and type improvement

Lightning00Blade:


packages/puppeteer-core/src/util/Deferred.ts
```diff
@@ -17,26 +17,28 @@ export interface DeferredOptions {
...
+  #value: T | V | TimeoutError | undefined;
   #resolver: (value: void) => void = () => {};
   #taskPromise = new Promise<void>(resolve => {
     this.#resolver = resolve;
   });
   #timeoutId: ReturnType<typeof setTimeout> | undefined;
+  #timeoutError: TimeoutError;
 
   constructor(opts?: DeferredOptions) {
+    this.#timeoutError = new TimeoutError(opts && opts.message);
```
Lightning00Blade, line 32: Create the error on init so the StackTrace have the path to where it was called similar to Callbacks in protocol.

OrKoN (approved):

PR #11111, opened 2023-10-10
chore: add a deflake utility

Lightning00Blade:
Provide an easy way to wrap a test and get its logs, tooling added to prevent submitting it to `main`. 
Also adds Logger to BiDi server.

### DIFF
--- a/test/src/oopif.spec.ts
+++ b/test/src/oopif.spec.ts
@@ -18,10 +18,10 @@ import expect from 'expect';
 import type {BrowserContext} from 'puppeteer-core/internal/api/BrowserContext.js';
 import type {CdpTarget} from 'puppeteer-core/internal/cdp/Target.js';
 
-import {describeWithDebugLogs, getTestState, launch} from './mocha-utils.js';
+import {getTestState, launch} from './mocha-utils.js';
 import {attachFrame, detachFrame, navigateFrame} from './utils.js';
 
-describeWithDebugLogs('OOPIF', function () {
+describe('OOPIF', function () {
   /* We use a special browser for this test as we need the --site-per-process flag */
   let state: Awaited<ReturnType<typeof launch>>;
 
--- a/tools/mocha-runner/src/interface.ts
+++ b/tools/mocha-runner/src/interface.ts
@@ -16,6 +16,10 @@
 
 import Mocha from 'mocha';
 import commonInterface from 'mocha/lib/interfaces/common';
+import {
+  setLogCapture,
+  getCapturedLogs,
+} from 'puppeteer-core/internal/common/Debug.js';
 
 import {testIdMatchesExpectationPattern} from './utils.js';
 
@@ -28,6 +32,10 @@ const skippedTests: Array<{testIdPattern: string; skip: true}> = process.env[
   ? JSON.parse(process.env['PUPPETEER_SKIPPED_TEST_CONFIG'])
   : [];
 
+const deflakeRetries = Number(process.env['PUPPETEER_DEFLAKE_RETRIES'] ?? 100);
+const deflakeTestPattern: string | undefined =
+  process.env['PUPPETEER_DEFLAKE_TESTS'];
+
 function shouldSkipTest(test: Mocha.Test): boolean {
   // TODO: more efficient lookup.
   const definition = skippedTests.find(skippedTest => {
@@ -39,8 +47,26 @@ function shouldSkipTest(test: Mocha.Test): boolean {
   return false;
 }
 
+function shouldDeflakeTest(test: Mocha.Test): boolean {
+  if (deflakeTestPattern) {
+    // TODO: cache if we have seen it already
+    return testIdMatchesExpectationPattern(test, deflakeTestPattern);
+  }
+  return false;
+}
+
+function dumpLogsIfFail(this: Mocha.Context) {
+  if (this.currentTest?.state === 'failed') {
+    console.log(
+      `\n"${this.currentTest.fullTitle()}" failed. Here is a debug log:`
+    );
+    console.log(getCapturedLogs().join('\n') + '\n');
+  }
+  setLogCapture(false);
+}
+
 function customBDDInterface(suite: Mocha.Suite) {
-  const suites = [suite];
+  const suites: [Mocha.Suite] = [suite];
 
   suite.on(
     Mocha.Suite.constants.EVENT_FILE_PRE_REQUIRE,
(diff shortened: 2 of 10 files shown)


tools/mocha-runner/src/interface.ts
```diff
@@ -110,14 +174,43 @@ function customBDDInterface(suite: Mocha.Suite) {
...
-          (context['it'] as typeof it)(title, fn, true)
+          (context['it'] as unknown as typeof it)(title, fn, true)
         );
       };
 
       it.skip = function (title: string) {
         return context['it'](title);
       };
 
+      it.only = function (title: string, fn: Mocha.TestFunction) {
```
jrandolf: There are two `it.only`?

tools/mocha-runner/src/interface.ts
```diff
@@ -78,12 +93,35 @@ function customBDDInterface(suite: Mocha.Suite) {
...
+        });
+      };
+
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-expect-error
       context['describe'] = describe;
 
       function it(title: string, fn: Mocha.TestFunction, itOnly = false) {
-        const suite = suites[0]!;
+        const suite: Mocha.Suite = suites[0]!;
```
jrandolf: Just use `as` here.

tools/mocha-runner/src/interface.ts
```diff
@@ -28,6 +32,9 @@ const skippedTests: Array<{testIdPattern: string; skip: true}> = process.env[
   ? JSON.parse(process.env['PUPPETEER_SKIPPED_TEST_CONFIG'])
   : [];
 
+const deflakeRetries = Number(process.env['PUPPETEER_DEFLAKE_RETRIES'] ?? 100);
+const deflakeTest: string | undefined = process.env['PUPPETEER_DEFLAKE_TEST'];
```
jrandolf: Just make deflake and deflakeRetries one variable that takes `true` or a number when specified.

test/src/mocha-types.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type * as MochaBase from 'mocha';
+
+declare global {
```
jrandolf: Can you place this within `mocha-utils`?

tools/mocha-runner/src/interface.ts
Lightning00Blade: I don't think this is an option as one `PUPPETEER_DEFLAKE_TEST` is the test pattern (See README) and the other is number of retries.

tools/mocha-runner/src/interface.ts
Lightning00Blade: No clue, nice spot removing.

tools/mocha-runner/src/interface.ts
jrandolf: Ah, okay. Hmm, let's make it `PUPPETEER_DEFLAKE_TESTS` then to clarify it's multiple tests

tools/mocha-runner/src/interface.ts
```diff
@@ -95,8 +133,34 @@ function customBDDInterface(suite: Mocha.Suite) {
             return child === suite;
           })
         );
-
-        if (shouldSkipTest(test) && !(itOnly || describeOnly)) {
+        if (shouldDeflakeTest(test)) {
+          console.log('HERE is hit', deflakeRetries);
```
OrKoN: remove

tools/mocha-runner/src/interface.ts
```diff
@@ -110,14 +174,43 @@ function customBDDInterface(suite: Mocha.Suite) {
...
       };
 
+      it.only = function (title: string, fn: Mocha.TestFunction) {
+        return common.test.only(
+          mocha,
+          (context['it'] as unknown as typeof it)(title, fn, true)
+        );
+      };
+
+      it.skip = function (title: string) {
```
OrKoN: remove

tools/mocha-runner/src/interface.ts
```diff
@@ -110,14 +174,43 @@ function customBDDInterface(suite: Mocha.Suite) {
...
+              `deflake ${title}`,
+              () => {
+                func(`${i}/${title}`, fn);
+              }
+            );
+          }
+        };
+      }
+
+      it.deflake = wrapDeflake(it);
```
OrKoN, line 189: remove

Lightning00Blade: I would prefer to keep this as it's easier for me to use locally. EsLint tooling is provided in the PR to prevent submitting it to main.
For the offline question about `How does it work with Expectation`:
The short answer it it does not work on the verifying part and it can't as we need to provide a wrapper `describe` to use Before/After hooks and that will always put spaces in the `fullTestName`.

OrKoN: I disagree, it looks like another way to do the same thing that is not much more convenient than using the env variable and `.only`. I would prefer not to double the number of it helpers for this feature.

test/src/oopif.spec.ts
```diff
@@ -18,10 +18,10 @@ import expect from 'expect';
 import type {BrowserContext} from 'puppeteer-core/internal/api/BrowserContext.js';
 import type {CdpTarget} from 'puppeteer-core/internal/cdp/Target.js';
 
-import {describeWithDebugLogs, getTestState, launch} from './mocha-utils.js';
+import {getTestState, launch} from './mocha-utils.js';
 import {attachFrame, detachFrame, navigateFrame} from './utils.js';
 
-describeWithDebugLogs('OOPIF', function () {
+describe.withDebugLogs('OOPIF', function () {
```
OrKoN: let's remove this too, I think we would not need it anymore

tools/mocha-runner/src/interface.ts
Lightning00Blade: Consider this: You have a test that you know is flaky or you wanna stress test it.
If we have only only `env` we have to find a pattern that matches the test usually `[<fileName>] * <testName>`,
but you have to remember a variable name and write it in Terminal, also due to the way our test are named that may hit other test as well. 
Or you can prepend the test with the helper that is auto-completed by most IDEs and has TsDoc explaining what it does + the test out recompile and build so you don't need an extra step.

I think where I am going is - I don't see the problem with having both options available.

OrKoN (approved):

PR #11126, opened 2023-10-11
chore: use groups for dependabot

jrandolf-google:


Lightning00Blade (commented):
We should keep the limit on open PRs, If possible per group.

jrandolf:
> We should keep the limit on open PRs, If possible per group.

There will only be one per group, so in theory there will always be only 2-3 PRs.

Lightning00Blade (approved):

PR #11130, opened 2023-10-11
chore: ignore `@types/node` for dependabot

jrandolf-google:


.github/dependabot.yml
```diff
@@ -8,6 +8,7 @@ updates:
       timezone: Europe/Berlin
     ignore:
       - dependency-name: devtools-protocol
+      - dependency-name: '@types/node'
```
Lightning00Blade, line 11: Add rollup aswell as v4 requires Node 18

jrandolf: For `rollup`, we can do a temporary ignore on the PR. We _always_ want to ignore @types/node

Lightning00Blade (approved):

PR #11132, opened 2023-10-11
chore: use hereby instead of internal job

jrandolf-google:
Hereby is a simple task runner. It allows us to move the complex build commands into a file.


OrKoN (changes requested):
can we use Promise.all and Node for this sort of thing? I am not sure why we needed the jobs in the first place as we don't have complex dependencies. Perhaps you could elaborate on that?

jrandolf:
@OrKoN Hereby is a very minimal framework that does precisely that. Otherwise, we have to create scripts everywhere and maintain them all separately.

packages/puppeteer-core/tools/generate_sources.ts
```diff
@@ -1,88 +0,0 @@
-#!/usr/bin/env node
-import {mkdir, mkdtemp, readFile, rm, writeFile} from 'fs/promises';
-import path, {join, resolve} from 'path';
-import {chdir} from 'process';
-
-import esbuild from 'esbuild';
-
-import {job} from '../../../tools/internal/job.js';
```
OrKoN, line 8: is there anything else that uses tools/internal/job.ts?

jrandolf: Yea, I plan to migrate them in another PR.

OrKoN (approved):

PR #11133, opened 2023-10-11
refactor: use RxJS for waitForHttp

Lightning00Blade:


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1724,6 +1732,38 @@ export abstract class Page extends EventEmitter<PageEvents> {
     options?: {timeout?: number}
   ): Promise<HTTPResponse>;
 
+  /**
+   * @internal
+   */
+  async _waitForHTTP<HTTP extends {url(): string}>(
```
OrKoN: let's define it as a standalone helpers as it is not part of the API

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1724,6 +1732,38 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
+    networkManager: EventEmitter<NetworkManagerEvents>,
```
OrKoN: do we need to type-safe events here? let's define am interface we need? e.g., events with something that emits HTTP

packages/puppeteer-core/src/common/events.ts
```diff
@@ -0,0 +1,48 @@
...
+import type {EventType} from './EventEmitter.js';
+
+/**
+ * We use symbols to prevent any external parties listening to these events.
+ * They are internal to Puppeteer.
+ *
+ * @internal
+ */
+// eslint-disable-next-line @typescript-eslint/no-namespace
+export namespace NetworkManagerEvent {
```
OrKoN: if we extract this, let's call it NetworkManagerEvents.ts (the file)

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -3257,3 +3263,33 @@ function roundRectangle<BoundingBoxType extends BoundingBox>(
   const height = Math.round(clip.height + clip.y - y);
   return {...clip, x, y, width, height};
 }
+
+/**
+ * @internal
+ */
+
+export async function waitForHTTP<T extends {url(): string}>(
```
OrKoN: let's move it out of api?

OrKoN (approved):

PR #11141, opened 2023-10-12
chore: expose enabled chrome features as an array

thiagowfx:
For consistency with disabled features.

No

### DIFF
--- a/packages/puppeteer-core/src/node/ChromeLauncher.ts
+++ b/packages/puppeteer-core/src/node/ChromeLauncher.ts
@@ -176,6 +176,8 @@ export class ChromeLauncher extends ProductLauncher {
       'ProcessPerSiteUpToMainFrameThreshold',
     ];
 
+    const enabledFeatures = ['NetworkServiceInProcess2'];
+
     const chromeArguments = [
       '--allow-pre-commit-input',
       '--disable-background-networking',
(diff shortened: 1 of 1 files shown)


packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -166,7 +166,7 @@ export class ChromeLauncher extends ProductLauncher {
   override defaultArgs(options: BrowserLaunchArgumentOptions = {}): string[] {
     // See https://github.com/GoogleChrome/chrome-launcher/blob/main/docs/chrome-flags-for-tools.md
 
-    const disabledFeatures = [
+    const disabledFeatures: string[] = [
```
OrKoN: no need for explicit types

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -176,6 +176,10 @@ export class ChromeLauncher extends ProductLauncher {
       'ProcessPerSiteUpToMainFrameThreshold',
     ];
 
+    const enabledFeatures: string[] = [
```
OrKoN: ditto

OrKoN (approved):
LGTM % comments

PR #11144, opened 2023-10-12
feat: implement waiting for worker

jrandolf-google:


OrKoN (changes requested):
Let's be consistent and call it waitForWorker (similar to waitForFrame). The current name makes it sound like a getter and i's not clear why, for example, Page.workers() + filter would not get the same result.

jrandolf:
> Let's be consistent and call it waitForWorker (similar to waitForFrame). The current name makes it sound like a getter and i's not clear why, for example, Page.workers() + filter would not get the same result.

In the long term, I want to move away from `waitFor` and use just getter calls. It's obvious that if that if a getter is async, then it's waiting for the object that's being getted. In particular, if it's already in `Page.workers`, it will immediately resolve.

OrKoN:
I don't think we made the decision about getting rid of waitFor* yet.

> It's obvious that if that if a getter is async, then it's waiting for the object that's being getter.

I don't find it obvious. There could be other reasons for a getter to be async.

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1812,6 +1812,20 @@ export abstract class Page extends EventEmitter<PageEvents> {
     );
   }
 
+  /**
+   * Gets a worker with the given URL.
```
OrKoN, line 1816:
```suggestion
   * Waits for a worker with the given URL. If multiple workers have the same URL, the first one available is returned.
```

packages/puppeteer-core/src/cdp/WebWorker.ts
```diff
@@ -117,6 +118,21 @@ export class WebWorker extends EventEmitter<Record<EventType, unknown>> {
...
+   */
+  set terminated(value: boolean) {
+    this.#terminated = value;
+  }
+
+  /**
+   * Whether the {@link WebWorker}
+   * {@link https://developer.mozilla.org/en-US/docs/Web/API/Worker/terminate | terminated}.
+   */
+  get terminated(): boolean {
```
OrKoN, line 132: missing test coverage for terminated and it does not appear to be used internally?

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -1218,6 +1244,26 @@ export class CdpPage extends Page {
   ): Promise<DeviceRequestPrompt> {
     return await this.mainFrame().waitForDevicePrompt(options);
   }
+
+  override async worker(
+    url: string,
+    options: WaitForTargetOptions = {}
```
OrKoN, line 1250: should we create `WaitForWorkerOptions`?

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -1218,6 +1244,26 @@ export class CdpPage extends Page {
...
+  ): Promise<WebWorker> {
+    const {timeout: ms = 30_000} = options;
```
OrKoN, line 1252: This probably should use the page default timeout

packages/puppeteer-core/src/cdp/Page.ts
OrKoN, line 1250: let's have an abort signal for all new waiters

Issue #11072, opened 2023-10-03
[Bug]: Puppeteer crashes on waitForDevicePrompt()


outofambit:
### Minimal, reproducible example

```TypeScript
import puppeteer from "puppeteer";

(async () => {
  const browser = await puppeteer.launch({
    headless: false,
  });
  const mainPage = await browser.newPage();
  await mainPage.goto("http://google.com");
  await mainPage.waitForDevicePrompt();
})();
```

### Error string

no error

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

I've been working on automating running the manual web bluetooth tests in https://github.com/WebBluetoothCG/manual-tests. (see https://github.com/WebBluetoothCG/manual-tests/issues/38 for more details). I ran into this issue while trying to use puppeteer's API to connect to a bluetooth device.

I've verified that:
- running `navigator.bluetooth.getDevices` in the console of a Chrome for Testing opened by puppeteer doesn't crash
- manually sending a message over CDP to enable DeviceAccess on a Chrome for Testing instance opened by puppeteer also doesn't crash

These are equivalent to what should happen with a call to `waitForDevicePrompt()`, leading me to believe this is a bug in puppeteer itself. (I've also reproduced this crash with a headless setup, but I'm not sure this can work in a headless environment anyway).

From looking through this repo's git history, it seems that this feature itself has been minimally changed since it was added in https://github.com/puppeteer/puppeteer/pull/9299/, so I'm wondering if perhaps something changed elsewhere in the codebase to break this behavior?

Please let me know if any more information would be helpful!

---
**Note:** The minimal script to achieve my goal is below, but the minimal script I provided above is enough to trigger the crash.

```ts
import puppeteer from "puppeteer";

(async () => {
  const browser = await puppeteer.launch({
    headless: false,
  });
  const mainPage = await browser.newPage();
  await mainPage.goto(
    "http://googlechrome.github.io/samples/web-bluetooth/get-devices-async-await.html"
  );
  await mainPage.waitForNetworkIdle();
  const [devicePrompt] = await Promise.all([
    mainPage.waitForDevicePrompt(),
    mainPage.locator("#requestBluetoothDevice").click(),
  ]);
  console.log(devicePrompt.devices);
  await browser.close();
})();
```

### Expectation

In the above script, I would expect puppeteer to timeout while waiting for the bluetooth menu to appear. In my actual use case, I would add code 

### Reality

[...]

PR #11152, opened 2023-10-13
feat: merge user-provided --{disable,enable}-features in args

thiagowfx:
Currently user-provided features are completely ignored in args, being overridden with default ones.

Bug: #11072

### DIFF
--- a/packages/puppeteer-core/src/node/ChromeLauncher.test.ts
+++ b/packages/puppeteer-core/src/node/ChromeLauncher.test.ts
@@ -1,0 +1,47 @@
+/**
+ * Copyright 2023 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import {describe, it} from 'node:test';
+
+import expect from 'expect';
+
+import {getFeatures} from './ChromeLauncher.js';
+
+describe('getFeatures', () => {
+  it('returns an empty array when no options are provided', () => {
+    const result = getFeatures('--foo');
+    expect(result).toEqual([]);
+  });
+
+  it('returns an empty array when no options match the flag', () => {
+    const result = getFeatures('--foo', ['--bar', '--baz']);
+    expect(result).toEqual([]);
+  });
+
+  it('returns an array of values when options match the flag', () => {
+    const result = getFeatures('--foo', ['--foo=bar', '--foo=baz']);
+    expect(result).toEqual(['bar', 'baz']);
+  });
+
+  it('handles whitespace around the flag and value', () => {
+    const result = getFeatures('--foo', ['--foo bar', '--foo baz ']);
+    expect(result).toEqual(['bar', 'baz']);
+  });
+
+  it('handles equals sign around the flag and value', () => {
+    const result = getFeatures('--foo', ['--foo=bar', '--foo=baz ']);
+    expect(result).toEqual(['bar', 'baz']);
+  });
+});
--- a/packages/puppeteer-core/src/node/ChromeLauncher.test.ts
+++ b/packages/puppeteer-core/src/node/ChromeLauncher.test.ts
@@ -35,9 +35,9 @@ describe('getFeatures', () => {
     expect(result).toEqual(['bar', 'baz']);
   });
 
-  it('handles whitespace around the flag and value', () => {
+  it('does not handle whitespace', () => {
     const result = getFeatures('--foo', ['--foo bar', '--foo baz ']);
-    expect(result).toEqual(['bar', 'baz']);
+    expect(result).toEqual([]);
   });
 
   it('handles equals sign around the flag and value', () => {
--- a/packages/puppeteer-core/src/node/ChromeLauncher.ts
+++ b/packages/puppeteer-core/src/node/ChromeLauncher.ts
@@ -274,7 +274,8 @@ function convertPuppeteerChannelToBrowsersChannel(
 }
 
 /**
- * Extracts all features from the given command-line flag.
+ * Extracts all features from the given command-line flag
+ * (e.g. `--enable-features`, `--enable-features=`).
  *
  * Example input:
  * ["--enable-features=NetworkService,NetworkServiceInProcess", "--enable-features=Foo"]
@@ -285,10 +286,10 @@ function convertPuppeteerChannelToBrowsersChannel(
 export function getFeatures(flag: string, options: string[] = []): string[] {
   const opts = options
     .filter(s => {
-      return s.startsWith(flag);
+      return s.startsWith(flag.endsWith('=') ? flag : `${flag}=`);
     })
     .map(s => {
-      return s.split(new RegExp(`${flag}` + '[=\\s]*'))[1]!.trim();
+      return s.split(new RegExp(`${flag}` + '=\\s*'))[1]?.trim();
     });
   return opts.length === 0 ? [] : opts.join(',').split(',');
 }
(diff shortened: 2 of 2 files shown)


packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -174,9 +175,12 @@ export class ChromeLauncher extends ProductLauncher {
       'OptimizationHints',
       // https://crbug.com/1492053
       'ProcessPerSiteUpToMainFrameThreshold',
-    ];
+    ].concat(getFeatures('--disable-features', options.args));
```
OrKoN: nit: slight preference for the destucturing

OrKoN:
> Currently user-provided features are completely ignored in args, being overridden with default ones.

Is it the case? I would expect that user-provided args override the default ones.

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -266,3 +270,23 @@ function convertPuppeteerChannelToBrowsersChannel(
...
+ * Example output:
+ * ["NetworkService", "NetworkServiceInProcess", "Foo"]
+ */
+export function getFeatures(flag: string, options: string[] = []): string[] {
+  const opts = options
+    ?.filter(s => {
+      return s.startsWith(flag);
+    })
+    .map(s => {
+      return s.split(new RegExp(`${flag}[= ]*`))[1]!.trim();
```
OrKoN:
```suggestion
      return s.split(new RegExp(`${flag}\s*=\s*`))[1]!.trim();
```

thiagowfx: Done

thiagowfx:
> Is it the case? I would expect that user-provided args override the default ones.

Yes. If you want to test it out locally, run this script:

```
#!/usr/bin/env node

(async () => {
    devtools: true,
    headless: false,
    args: ['--enable-features=WebBluetoothNewPermissionsBackend'],
  });
  await mainPage.goto(
    'http://googlechrome.github.io/samples/web-bluetooth/get-devices-async-await.html'
  );
})();
```

Then observe that a log message will appear in the page that sayd `getDevices` is not a function.
When the feature is enabled, this message does not appear.

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -266,3 +270,23 @@ function convertPuppeteerChannelToBrowsersChannel(
...
+ *
+ * Example input:
+ * ["--enable-features=NetworkService,NetworkServiceInProcess", "--enable-features=Foo"]
+ *
+ * Example output:
+ * ["NetworkService", "NetworkServiceInProcess", "Foo"]
+ */
+export function getFeatures(flag: string, options: string[] = []): string[] {
+  const opts = options
+    ?.filter(s => {
```
Lightning00Blade: Options is always an array. So we can remove the `?` and return the `options.filter().map()` directly

packages/puppeteer-core/src/node/ChromeLauncher.ts
thiagowfx: What do you mean? `, {args}`?

From `export function getFeatures(flag: string, options: string[] = []): string[] {`

To `export function getFeatures(flag: string, options: {options: string[]} = {options: []}): string[] {`

?

thiagowfx: Ah, I guess you mean the spread operator

OrKoN:
```suggestion
      ...getFeatures('--disable-features', options.args),
    ];
```

OrKoN: See the suggestion

OrKoN: Yeah, sorry :)

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -266,3 +270,23 @@
...
+      return s.startsWith(flag);
+    })
+    .map(s => {
+      return s.split(new RegExp(`${flag}\s*=\s*`))[1]!.trim();
```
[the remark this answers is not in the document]
OrKoN: Not sure why this complains

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -266,3 +270,23 @@ function convertPuppeteerChannelToBrowsersChannel(
...
```
OrKoN:
```suggestion
      return s.split(new RegExp(`${flag}\\s*=\\s*`))[1]!.trim();
```

packages/puppeteer-core/src/node/ChromeLauncher.test.ts
```diff
@@ -0,0 +1,47 @@
...
+    expect(result).toEqual([]);
+  });
+
+  it('returns an array of values when options match the flag', () => {
+    const result = getFeatures('--foo', ['--foo=bar', '--foo=baz']);
+    expect(result).toEqual(['bar', 'baz']);
+  });
+
+  it('handles whitespace around the flag and value', () => {
+    const result = getFeatures('--foo', ['--foo bar', '--foo baz ']);
```
OrKoN, line 39: does Chrome allow whitespaces?

OrKoN: Then my previous suggestion might have been wrong

OrKoN: Chrome does not allow the values to come after a space

thiagowfx: Done

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -266,3 +272,26 @@ function convertPuppeteerChannelToBrowsersChannel(
...
+ * Example input:
+ * ["--enable-features=NetworkService,NetworkServiceInProcess", "--enable-features=Foo"]
+ *
+ * Example output:
+ * ["NetworkService", "NetworkServiceInProcess", "Foo"]
+ *
+ * @internal
+ */
+export function getFeatures(flag: string, options: string[] = []): string[] {
+  const opts = options
```
Lightning00Blade: Why do we need the check at the end, I think just doing this does the same, no?
```suggestion
  return options
```

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -266,3 +272,26 @@ function convertPuppeteerChannelToBrowsersChannel(
...
+ */
+export function getFeatures(flag: string, options: string[] = []): string[] {
+  return options
+    .filter(s => {
+      return s.startsWith(flag.endsWith('=') ? flag : `${flag}=`);
+    })
+    .map(s => {
+      return s.split(new RegExp(`${flag}` + '=\\s*'))[1]?.trim();
+    });
+  return opts.length === 0 ? [] : opts.join(',').split(',');
```
Lightning00Blade:
```suggestion
```

thiagowfx: The first suggestion was actually wrong. `opts` was created as an intermediate value. I was having compilation issues otherwise

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -266,3 +272,26 @@ function convertPuppeteerChannelToBrowsersChannel(
...
+export function getFeatures(flag: string, options: string[] = []): string[] {
+  const opts = options
+    .filter(s => {
+      return s.startsWith(flag.endsWith('=') ? flag : `${flag}=`);
+    })
+    .map(s => {
+      return s.split(new RegExp(`${flag}` + '=\\s*'))[1]?.trim();
+    });
+  return opts.length === 0 ? [] : opts.join(',').split(',');
+}
```
Lightning00Blade:
```suggestion
export function getFeatures(flag: string, options: string[] = []): string[] {
  return options
    .filter(s => {
      return s.startsWith(flag.endsWith('=') ? flag : `${flag}=`);
    })
    .map(s => {
      return s.split(new RegExp(`${flag}=\\s*`))[1]?.trim();
    })
    .filter(s => {
      return s;
    }) as string[];
}
```

thiagowfx:
Done

Lightning00Blade (approved):

PR #11153, opened 2023-10-13
fix: `Page.waitForDevicePrompt` crash

thiagowfx:
Only call `deviceRequestPromptManager()` if the parent frame is not null.

Bug: #11072
Follow-up-of: #9299

### DIFF
--- a/packages/puppeteer-core/src/cdp/Frame.ts
+++ b/packages/puppeteer-core/src/cdp/Frame.ts
@@ -21,7 +21,6 @@ import {Frame, FrameEvent, throwIfDetached} from '../api/Frame.js';
 import type {HTTPResponse} from '../api/HTTPResponse.js';
 import type {WaitTimeoutOptions} from '../api/Page.js';
 import {setPageContent} from '../common/util.js';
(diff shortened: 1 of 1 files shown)


thiagowfx:
I'm not sure if this fix is correct tough. @mzgoddard @outofambit thoughts?

I verified the crash doesn't happen anymore:

```
#!/usr/bin/env node

(async () => {
    devtools: true,
    headless: false,
    args: ['--enable-features=WebBluetoothNewPermissionsBackend'],
  });
  await mainPage.goto(
    'http://googlechrome.github.io/samples/web-bluetooth/get-devices-async-await.html'
  );
})();
```

thiagowfx:
Note that the `--args` only work with this patch: https://github.com/puppeteer/puppeteer/pull/11152

OrKoN (approved):
LGTM Can we find a way to test it e2e on GitHub actions?

thiagowfx:
I can confirm the unit tests do not change. I'll see if there's a way to E2E.

Lightning00Blade (approved):

PR #11157, opened 2023-10-13
chore: use CdpPage instead of Page in CDP frames

thiagowfx:
The purpose of this PR is to provide correct typings in anticipation of https://github.com/puppeteer/puppeteer/pull/11153, wherein we need access to the root frame.

Bug: #11072


Lightning00Blade (approved):
LGTM + EsLint check is failing so please fix and before merging

PR #11159, opened 2023-10-13
test: add an E2E test for device request prompt

### LINKED ISSUE #11072: [Bug]: Puppeteer crashes on waitForDevicePrompt()

thiagowfx:
Closes https://github.com/puppeteer/puppeteer/issues/11072

### DIFF
--- a/test/src/device-request-prompt.spec.ts
+++ b/test/src/device-request-prompt.spec.ts
@@ -57,5 +57,5 @@ describe('device request prompt', function () {
         timeout: 1,
       })
     ).rejects.toThrow(TimeoutError);
-  });
+  }).timeout(1_000);
 });
--- a/test/src/device-request-prompt.spec.ts
+++ b/test/src/device-request-prompt.spec.ts
@@ -19,23 +19,38 @@ import {TimeoutError} from 'puppeteer';
 import {launch} from './mocha-utils.js';
 
 describe('device request prompt', function () {
-  let testState: Awaited<ReturnType<typeof launch>>;
+  let state: Awaited<ReturnType<typeof launch>>;
+
+  before(async () => {
+    state = await launch(
+      {
+        args: ['--enable-features=WebBluetoothNewPermissionsBackend'],
+        ignoreHTTPSErrors: true,
+      },
+      {
+        after: 'all',
+      }
+    );
+  });
+
+  after(async () => {
+    await state.close();
+  });
 
   beforeEach(async () => {
-    testState = await launch({
-      args: ['--enable-features=WebBluetoothNewPermissionsBackend'],
-    });
+    state.context = await state.browser.createIncognitoBrowserContext();
+    state.page = await state.context.newPage();
   });
 
   afterEach(async () => {
-    await testState.close();
+    await state.context.close();
   });
 
   // Bug: #11072
   it('does not crash', async () => {
-    const {page, server} = testState;
+    const {page, httpsServer} = state;
 
-    await page.goto(server.EMPTY_PAGE);
+    await page.goto(httpsServer.EMPTY_PAGE);
 
     await expect(
       page.waitForDevicePrompt({
--- a/test/src/device-request-prompt.spec.ts
+++ b/test/src/device-request-prompt.spec.ts
@@ -54,7 +54,7 @@ describe('device request prompt', function () {
 
     await expect(
       page.waitForDevicePrompt({
-        timeout: 1,
+        timeout: 10,
       })
     ).rejects.toThrow(TimeoutError);
   }).timeout(1_000);
--- a/test/src/device-request-prompt.spec.ts
+++ b/test/src/device-request-prompt.spec.ts
@@ -1,0 +1,46 @@
+/**
+ * Copyright 2023 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import expect from 'expect';
+import {TimeoutError} from 'puppeteer';
+
+import {launch} from './mocha-utils.js';
+
+describe('device request prompt', function () {
+  let testState: Awaited<ReturnType<typeof launch>>;
+
+  beforeEach(async () => {
+    testState = await launch({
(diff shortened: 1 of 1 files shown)


thiagowfx:
The diffbase will become less noise once the parent PRs are merged.

test/src/device-request-prompt.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    await testState.close();
+  });
+
+  it('does not crash', async () => {
+    const {page, server} = testState;
+
+    await page.goto(server.EMPTY_PAGE);
+    await page.waitForNetworkIdle();
+
+    await expect(page.waitForDevicePrompt()).rejects.toThrow(TimeoutError);
```
Lightning00Blade: Please leave a comment ref to Bug.
Also if possible provided a lower time for timeout as currently the timeout is `30_000` and the test will always fail with a Timeout from Mocha (Mocha timeout is set to `10_000`)

test/src/device-request-prompt.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+import expect from 'expect';
+import {TimeoutError} from 'puppeteer';
+
+import {launch} from './mocha-utils.js';
+
+describe('device request prompt', function () {
+  let testState: Awaited<ReturnType<typeof launch>>;
+
+  beforeEach(async () => {
```
Lightning00Blade, line 40: Nit: Let's set it up the browser as we do for `ignorehttpserrors.spec.ts`
that way we if we add test they will use the same browser instance.

test/src/device-request-prompt.spec.ts
thiagowfx: > Please leave a comment ref to Bug.

Done

test/src/device-request-prompt.spec.ts
thiagowfx: Done

Lightning00Blade (approved):

test/src/device-request-prompt.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+import {launch} from './mocha-utils.js';
+
+describe('device request prompt', function () {
+  let state: Awaited<ReturnType<typeof launch>>;
+
+  before(async () => {
+    state = await launch(
+      {
+        args: ['--enable-features=WebBluetoothNewPermissionsBackend'],
+        ignoreHTTPSErrors: true,
```
Lightning00Blade, line 28: I believe we don't need this option.

thiagowfx: Without this option:

` Error: net::ERR_CERT_AUTHORITY_INVALID at https://localhost:63413/empty.html`

thiagowfx: Unless you want me to use `server` instead of `httpsServer`?

test/src/device-request-prompt.spec.ts
thiagowfx: The test seems to be completing quite quickly:

```
  device request prompt
    ✔ does not crash (40ms)
```

Are you sure this is needed?

thiagowfx: The default timeout from mocha is 2s (2000ms). I'll set it to 1s.

test/src/device-request-prompt.spec.ts
Lightning00Blade: Let's leave it as is then, I didn't manage to verify it but its possible that this feature is enabled only for secured sites (https).

test/src/device-request-prompt.spec.ts
```diff
@@ -57,5 +57,5 @@ describe('device request prompt', function () {
         timeout: 1,
       })
     ).rejects.toThrow(TimeoutError);
-  });
+  }).timeout(1_000);
```
Lightning00Blade: Prefer to set it in the test itself as first line. That way its easier to understand 
```
this.timeout(1_000)
```

Lightning00Blade (approved):

PR #11160, opened 2023-10-13
chore: use RxJS instead of waitWithTimeout

Lightning00Blade:


packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -179,25 +181,19 @@ export class BidiFrame extends Frame {
...
-          this.#context.once(waitUntilEvent, () => {
-            resolve();
-          });
-        }),
-        waitUntilEvent,
-        timeout
-      ),
-    ]);
+    await firstValueFrom(
+      forkJoin([
```
jrandolf, line 192: Nice!

jrandolf-google (approved):

PR #11167, opened 2023-10-16
chore: use internal method for networkidle with BiDi

Lightning00Blade:


packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -148,30 +134,18 @@ export class BidiFrame extends Frame {
       timeout = this.#timeoutSettings.navigationTimeout(),
     } = options;
 
-    const readinessState = lifeCycleToReadinessState.get(
-      getWaitUntilSingle(waitUntil)
-    ) as Bidi.BrowsingContext.ReadinessState;
+    const [wait, network] = getBiDiReadynessState(waitUntil);
```
jrandolf:
```suggestion
    const [waitEvent, network] = getBiDiReadynessState(waitUntil);
```

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -270,4 +246,47 @@ export class BidiFrame extends Frame {
       throw error;
     }
   }
+
+  /** @internal */
+  async _waitBidiLifecycleAndNetworkIdle(
```
jrandolf: I'm all for making another observable, but the beauty of RxJS really comes from explicit pipelines. Could we split this up into smaller operators that get composed rather than one large operator?

packages/puppeteer-core/src/bidi/Frame.ts
Lightning00Blade: I think both are not correct it should be `readyness` or `readynessState` as it's not an event.

packages/puppeteer-core/src/bidi/Frame.ts
Lightning00Blade: Let's chat offline about it.

tools/mocha-runner/package.json
```diff
@@ -21,6 +21,9 @@
       "output": [
         "bin/**",
         "tsconfig.tsbuildinfo"
+      ],
+      "dependencies": [
```
OrKoN, line 25: why does it depend on puppeteer-core?

test/TestExpectations.json
```diff
@@ -188,7 +176,7 @@
   {
     "testIdPattern": "[network.spec] network Page.setBypassServiceWorker *",
     "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["webDriverBiDi"],
+    "parameters": ["webDriverBiDi", "TIMEOUT"],
```
OrKoN: wrong?

test/TestExpectations.json
```diff
@@ -2333,12 +2291,6 @@
     "parameters": ["cdp", "firefox"],
     "expectations": ["FAIL"]
   },
-  {
```
OrKoN, line 2336: is this handled by another expecation?

tools/mocha-runner/package.json
Lightning00Blade: Because of the Deflake action, if you do a full clean and build this part will fail.

packages/puppeteer-core/src/bidi/lifecycle.ts
```diff
@@ -0,0 +1,120 @@
...
+  ['domcontentloaded', 'browsingContext.domContentLoaded'],
+]);
+
+export function getBiDiLifecycleEvent(
+  event: PuppeteerLifeCycleEvent | PuppeteerLifeCycleEvent[]
+): [
+  'browsingContext.load' | 'browsingContext.domContentLoaded',
+  BiDiNetworkIdle,
+] {
+  const lifecycle = getBiDiLifecycles(event);
```
OrKoN: let's either use `lifeCycle` or `lifecycle`

packages/puppeteer-core/src/bidi/lifecycle.ts
```diff
@@ -0,0 +1,120 @@
...
+
+export type BiDiNetworkIdle = Extract<
+  PuppeteerLifeCycleEvent,
+  'networkidle0' | 'networkidle2'
+> | null;
+
+/**
+ * @internal
+ */
+export function getBiDiLifecycles(
```
OrKoN: `getBiDiLifecycleEvents`?

test/TestExpectations.json
Lightning00Blade: Yeah, I have consolidate all the passing expectation into. That way we have one (or 2) thing affecting a test rather.
```json
  {
    "testIdPattern": "[navigation.spec] *",
    "expectations": ["PASS"]
  }
  ```

packages/puppeteer-core/src/bidi/lifecycle.ts
Lightning00Blade: OK, we should use `lifeCycle` as that's whats used in the original `PuppeteerLifeCycleEvent` type.

packages/puppeteer-core/src/bidi/lifecycle.ts
Lightning00Blade: We have already a `getBiDiLifecycleEvent` that return the correct BiDi event for the given `PuppeteerLifeCycle`.
having two with the same name seems confusing, I will try to think of a better name.

OrKoN (approved):

PR #11168, opened 2023-10-16
ci(security): pin docker image hash

thiagowfx:
As per https://github.com/puppeteer/puppeteer/security/code-scanning/426

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,3 +28,9 @@ updates:
       all:
         patterns:
           - '*'
+  - package-ecosystem: docker
+    directory: /
+    schedule:
+      interval: daily
+      time: '06:00'
+      timezone: Europe/Berlin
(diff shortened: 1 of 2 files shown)


Lightning00Blade:
Do you know if there is a way to auto-update this (maybe via dependabot), as we would like to keep it up-to date?

thiagowfx:
It's not currently enabled in our config though: https://github.com/puppeteer/puppeteer/blob/main/.github/dependabot.yml

Lightning00Blade:
Let's enable it in this PR then we can merge.

thiagowfx:
It's orthogonal though. Even if we did not add the hash, the same pre-existing issue applies.

thiagowfx:
...done anyway.

Lightning00Blade:
>  Even if we did not add the hash, the same pre-existing issue applies.
Not exactly with the pin we are forcing the users to an exact version while previously we were forcing the user to install the latest version for that pin. Having the auto-update will insure we provide close to the original expectations from the users.

Lightning00Blade (approved):

PR #11179, opened 2023-10-16
chore: update chromium-bidi

Lightning00Blade:


test/TestExpectations.json
```diff
@@ -1331,6 +1331,12 @@
     "parameters": ["chrome", "webDriverBiDi"],
     "expectations": ["FAIL", "TIMEOUT"]
   },
+  {
+    "testIdPattern": "[accessibility.spec] Accessibility should work",
```
Lightning00Blade: Due to use of latest Canary we are seeing different Accessibility tree: 
This is now not reported, maybe a bug,
```js
{role: 'StaticText', name: 'Hello World'}
```
All the other test have more info so we should ignore them that info for now via `toMatchObject`

jrandolf-google (approved):

PR #11224, opened 2023-10-23
chore: exclude unit test sources from npm pkg

OrKoN:


packages/puppeteer-core/package.json
```diff
@@ -108,6 +108,7 @@
   "files": [
     "lib",
     "src",
+    "!*.test.ts",
```
Lightning00Blade, line 111: Should we just ignore everything that has `.test.` as in this glob `!*.test.*` ?

Lightning00Blade (approved):

packages/puppeteer-core/package.json
OrKoN: We could, not sure if we will eventually face the opposite issue.

PR #11230, opened 2023-10-23
fix: do not pass --{enable,disable}-features twice when user-provided

### LINKED ISSUE #11208: [Bug] Merging disabled and enabled features does not remove additional enable/disable args
Inspect arguments for the browser for the following snippet:

```

(async () => {
  await puppeteer.launch({
    headless: "new",
    args: ["--disable-features=Test"]
  });
})();
```

You will see:
```
.cache/puppeteer/chrome/mac_arm-118.0.5993.70/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing --allow-pre-commit-input --disable-background-networking --disable-background-timer-throttling --disable-backgrounding-occluded-windows --disable-breakpad --disable-client-side-phishing-detection --disable-component-extensions-with-background-pages --disable-component-update --disable-default-apps --disable-dev-shm-usage --disable-extensions --disable-features=Translate,AcceptCHFrame,MediaRouter,OptimizationHints,ProcessPerSiteUpToMainFrameThreshold,Test --disable-hang-monitor --disable-ipc-flooding-protection --disable-popup-blocking --disable-prompt-on-repost --disable-renderer-backgrounding --disable-search-engine-choice-screen --disable-sync --enable-automation --enable-blink-features=IdleDetection --enable-features=NetworkServiceInProcess2 --export-tagged-pdf --force-color-profile=srgb --metrics-recording-only --no-first-run --password-store=basic --use-mock-keychain --headless=new --hide-scrollbars --mute-audio about:blank --disable-features=Test --remote-debugging-port=0
``` 

Although the Test feature is merged into the first flag it's not removed as the separate argument which might be confusing and might break if the browser changes the order of parsing. We probably should remove the merged args.

thiagowfx:
Currently, whenever the user provides:

    args: '--disable-features=TestDisableFeatures'

The --disable-features flag appears twice in chrome arguments.

Make it appear only once, by merging them.

Ditto for --enable-features.

Fixed: #11208

### DIFF
--- a/packages/puppeteer-core/src/node/ChromeLauncher.ts
+++ b/packages/puppeteer-core/src/node/ChromeLauncher.ts
@@ -171,7 +171,7 @@ export class ChromeLauncher extends ProductLauncher {
       options.args
     );
     if (options.args && userDisabledFeatures.length > 0) {
-      removeMatching(options.args, /^--disable-features=.*/);
+      removeMatchingFlags(options.args, '--disable-features');
     }
 
     // Merge default disabled features with user-provided ones, if any.
@@ -188,7 +188,7 @@ export class ChromeLauncher extends ProductLauncher {
 
     const userEnabledFeatures = getFeatures('--enable-features', options.args);
     if (options.args && userEnabledFeatures.length > 0) {
-      removeMatching(options.args, /^--enable-features=.*/);
+      removeMatchingFlags(options.args, '--enable-features');
     }
 
     // Merge default enabled features with user-provided ones, if any.
@@ -312,12 +312,13 @@ export function getFeatures(flag: string, options: string[] = []): string[] {
 }
 
 /**
- * Removes all elements in-place from the given array that match the given
- * regular expression.
+ * Removes all elements in-place from the given string array
+ * that match the given command-line flag.
  *
  * @internal
  */
-export function removeMatching(array: string[], regex: RegExp): string[] {
+export function removeMatchingFlags(array: string[], flag: string): string[] {
+  const regex = new RegExp(`^${flag}=.*`);
   let i = 0;
   while (i < array.length) {
     if (regex.test(array[i]!)) {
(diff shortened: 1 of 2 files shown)


OrKoN (approved):

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -166,6 +166,14 @@ export class ChromeLauncher extends ProductLauncher {
   override defaultArgs(options: BrowserLaunchArgumentOptions = {}): string[] {
     // See https://github.com/GoogleChrome/chrome-launcher/blob/main/docs/chrome-flags-for-tools.md
 
+    const userDisabledFeatures = getFeatures(
+      '--disable-features',
+      options.args
+    );
+    if (userDisabledFeatures.length > 0) {
+      removeMatching(options.args ?? [], /^--disable-features=.*/);
```
Lightning00Blade: Nit: 
```suggestion
    if (options.args && userDisabledFeatures.length > 0) {
      removeMatching(options.args, /^--disable-features=.*/);
```

packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -297,3 +310,21 @@ export function getFeatures(flag: string, options: string[] = []): string[] {
...
     }) as string[];
 }
+
+/**
+ * Removes all elements in-place from the given array that match the given
+ * regular expression.
+ *
+ * @internal
+ */
+export function removeMatching(array: string[], regex: RegExp): string[] {
```
Lightning00Blade: Can we create the reg-ex in this function it would be nice to have a consistent API for extracting and removing. 
Than we a better name would be `removeMatchingFlags`.

Lightning00Blade (approved):

packages/puppeteer-core/src/node/ChromeLauncher.ts
thiagowfx: Do you mean to change this to `string`? OK

thiagowfx: Done

PR #11234, opened 2023-10-23
fix: remove import cycle in query handlers

OrKoN:
I could only think of a way to check using a symbol. This satisfies TypeScript, the runtime and does not cause cycles in the import graph.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -131,10 +131,7 @@ module.exports = {
       },
     ],
 
-    'import/no-cycle': [
-      'error',
-      {maxDepth: Infinity, allowUnsafeDynamicCyclicDependency: true},
-    ],
+    'import/no-cycle': ['error', {maxDepth: Infinity}],
 
     'no-restricted-syntax': [
       'error',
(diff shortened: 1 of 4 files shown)


Lightning00Blade (approved):
An other option would be this [TypeScript Playground](https://www.typescriptlang.org/play?#code/FAYw9gdgzgLgBFAngWzgXjgZRQIzAGwAoByAJmIEphqR8BDKKOAQTgG9q44BtbZPfADoYYTDABOASwgBzACp0ZAXXRxiAMXF1kAU2JwGccNBgBuTnBk74IsVNkKZhChy5dx1gK7iIcGAAtJKF5cAmFRCWl5RSVzLgBfYETQekY4ACF2YC4YHVhnVzcPGG9fCU8dOLhErhD+MNtIhxjVYgBRfB1dCBgACToIABNO-UNjWCqrGwj7aKcXbKKvHz9A4L4BcLsox1jFxOTxm1UIHQB3FmdQSFg-eAwYOs3G2d24a5M-AEYT84zncxHO4-B5fJ4NGY7FrUYAAM08EBAMEkkDg-h0+AADoQQAAuFhwAA+GQo+LwBB0A0KcEksJxWyac3QaAw7U63T6A2GegoWTccGKpSMwjyMCuCQsgpWsLo+CgOiSQA). 
But I am not sure it's better the issue is you can not use `x[Y]` to narrow down a type, but with `Symbol` brackets are always needed. 
TS issue is from 2016, main problem seems to be performance - https://github.com/microsoft/TypeScript/issues/10530

PR #11236, opened 2023-10-23
ci: disable device-request-prompt tests for firefox

### LINKED ISSUE #11215: "[device-request-prompt.spec] device request prompt does not crash" is flaky with Firefox
See https://github.com/puppeteer/puppeteer/actions/runs/6596315106/job/17922262144

Firefox does not support these CDP commands so we could perhaps skip those tests for Firefox.

thiagowfx:
Firefox does not support these CDP commands.

Fixed: #11215

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -36,6 +36,12 @@
     "expectations": ["PASS"]
   },
   {
+    "testIdPattern": "[device-request-prompt.spec] *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox"],
+    "expectations": ["SKIP"]
+  },
+  {
     "testIdPattern": "[dialog.spec] *",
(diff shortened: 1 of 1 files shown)


test/TestExpectations.json
```diff
@@ -35,6 +35,12 @@
     "parameters": ["webDriverBiDi"],
     "expectations": ["PASS"]
   },
+  {
+    "testIdPattern": "[device-request-prompt.spec] *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox", "headful", "cdp"],
```
thiagowfx: Should this be just `"firefox"`?

Lightning00Blade: Yes, that will disable for all `firefox` flavors.

thiagowfx: Done.

OrKoN (approved):

PR #11264, opened 2023-10-27
chore: fix BiDi deserializer

### LINKED ISSUE #11259: [WebDriverBiDi] Returning a non-serializable type embedded within an object results in an "undefined" value
When running the following code the `result` will be `undefined` instead of an `object` with all the possible serializable elements contained:

```
    await page.setContent("<img src='data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mNkYPhfDwAChwGA60e6kgAAAABJRU5ErkJggg=='>");

    const result = await page.evaluate(async () => {
      var image = document.querySelector("img");
      var imageBitmap = await createImageBitmap(image);

      return {
        a: "foo",
        b: imageBitmap,
      };
    });
```

The expected value should probably be `{"a": "foo", "b": undefined}`. 

Commenting out the property `b` will serialize the object correctly.

CC @OrKoN

Lightning00Blade:
Closes #11259

With this change this test `[evaluation.spec] Evaluation specs Page.evaluate should fail for circular object` is failing,
not sure how to action on it.

### DIFF
--- a/packages/puppeteer-core/src/bidi/Deserializer.ts
+++ b/packages/puppeteer-core/src/bidi/Deserializer.ts
@@ -79,7 +79,7 @@ export class BidiDeserializer {
     }
 
     debugError(`Deserialization of type ${result.type} not supported.`);
-    return {};
+    return undefined;
   }
 
   static deserializeTuple([serializedKey, serializedValue]: [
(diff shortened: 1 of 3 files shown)


jrandolf-google (approved):

OrKoN (approved):

Lightning00Blade:
Decided to disable the test, in BiDi we have the InternalId and we may be able to reproduce the circular object so `[evaluation.spec] Evaluation specs Page.evaluate should fail for circular object` actually does not need to fail but passes.

PR #11274, opened 2023-10-30
fix: Type for ElementHandle.screenshot 

### LINKED ISSUE #11273: [Bug] [Typescript]: elementHandle.screenshot with dubious return type

```TypeScript

(async () => {
  await page.waitForSelector("body");
  const elementHandle = await page.$("body");
  if (elementHandle == null)
    return;
  const imageBuffer = await elementHandle.screenshot({ encoding: "binary" }) // Type is string | Buffer. Correct should be Buffer
  const imageString = await elementHandle.screenshot({ encoding: "base64" }) // Type is string | Buffer. Correct should be string
  console.log({imageBuffer, imageString});
})();
```

### Error string

no error

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

This issue is very similar to #9512 but happening on the ElementHandle.screenshot method.

It's quite confusing to see the different return types when comparing both functions (Page.screenshot & ElementHandle.screenshot).

### Expectation

I expected the returning types of `ElementHandle.screenshot` and `Page.screenshot` to be the same under the same conditions.

### Reality

The types of `ElementHandle.screenshot` and `Page.screenshot` are different, under the same conditions.

_No response_

### Puppeteer version

21.4.1

### Node version

18.4.0

### Package manager

npm

### Package manager version

8.12.1

### Operating system

macOS

Lightning00Blade:
Add the proper overloads to the `ElementHandle.screenshot`, same as `Page.screenshot`

Bug: #11273


thiagowfx (approved):

thiagowfx:
cc @jrandolf

PR #11289, opened 2023-11-01
Update Dockerfile to LTS

### LINKED ISSUE #11290: [Feature]: Node 20 in puppeteer docker image
### Feature description

I would like to use Node 20 in the puppeteer image.
Node 20 is already in LTS status, and has some useful new shiny features.
I see no breaking changes in Node 20.

w3nl:
Always follow the latest LTS version
(Now Node 20)

To switch to use the real Node LTS version without changing code.
E.g. latest LTS version is now 20, in the codebase it is still 18.

**Summary**

Use latest LTS version from Node as soon as possible.

Should not break things, maybe in the future?

**Other information**

Let me know if you like this, or just use `node:20` 

closes: https://github.com/puppeteer/puppeteer/issues/11290


jrandolf:
We usually fix the version so semantics are clear. Nevertheless, fixed in #11296.

PR #11311, opened 2023-11-06
test: add installation test for BiDi

Lightning00Blade:
The take less than a 1 sec so speed should not be a problem.
image


OrKoN (approved):

OrKoN (changes requested):
1) it should test the launch argument as the env var is test only 2) and it needs to actually instantiate the bidi connection since bidi is dynamically imported

OrKoN (approved):

PR #11322, opened 2023-11-08
chore: add new UnsupportedOperation Error

Lightning00Blade:


packages/puppeteer-core/src/node/ProductLauncher.ts
```diff
@@ -203,12 +203,12 @@ export class ProductLauncher {
 
   executablePath(channel?: ChromeReleaseChannel): string;
   executablePath(): string {
-    throw new Error('Not implemented');
+    throw new UnsupportedOperation();
```
OrKoN: I think this should be always implemented, abstract instead?

Lightning00Blade: Fair enough, we should switch the whole class to be `abstract`. Will do it in a separate PR.

OrKoN (approved):

PR #11339, opened 2023-11-09
docs: fix ElementHandle boundingBox & boxModel descriptions

### LINKED ISSUE #11309: [Bug]: `boundingBox()` returns coordinates for elements with `visibility: hidden`

```TypeScript

export async function isElementVisibleByBoundingBox(elementHandle) {
  return Boolean(await elementHandle.boundingBox());
}

export async function getVisibleElementByBoundingBox(page, selector) {
  const element = await page.$(selector);
  if (element && (await isElementVisibleByBoundingBox(element))) {
    return element;
  }
}

(async () => {
  const browser = await launch({ headless: false });

  await page.goto('https://puppeteer-boundingbox-isvisible.vercel.app/');

  const elementWithBoundingBox = await getVisibleElementByBoundingBox(
    page,
    '[data-test-id="link-about"]',
  );

  if (elementWithBoundingBox) {
    console.log('[data-test-id="link-about"] is visible with boundingBox()');
  } else {
    console.log(
      '[data-test-id="link-about"] is not visible with boundingBox()',
    );
  }

  await browser.close();
})();
```

### Error string

no error

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

_No response_

### Expectation

The `boundingBox()` method should return `null` if an element is not visible, including if it has the CSS property `visibility: hidden`. This is stated in the [Puppeteer documentation](https://pptr.dev/api/puppeteer.elementhandle.boundingbox): 

> This method returns the bounding box of the element (relative to the main frame), or null if the element is not visible.

Screenshot 2023-11-06 at 11 09 29

Expected Puppeteer test result:
```
[data-test-id="link-about"] is not visible with boundingBox()
```

### Reality

The `boundingBox()` method returns the coordinates with the width and height of an element even if it has the CSS property `visibility: hidden`. This means that tests that use `boundingBox()` to check if an element is visible can fail even if the element is not actually visible on the page.

Actual Puppeteer test result:
```
[data-test-id="link-about"] is visible with boundingBox()
```

### Steps to reproduce:

1. Clone repo with `git clone https://github.com/ProchaLu/puppeteer-boundingbox-isvisible.git`
2. install the dependencies with `pnpm install`
3. run the test with `node util/test.mjs` 

If you want to test locally and make some changes, don't forget to change the URL in the `util/test.mjs` file.

_No response_

### Puppeteer version

21.5.0

### Node version

v20.9.0

### Package manager

pnpm

### Package manager version

8.9.2

### Operating system

macOS

Lightning00Blade:
Closes #11309


packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -1158,7 +1158,7 @@ export abstract class ElementHandle<
 
   /**
    * This method returns the bounding box of the element (relative to the main frame),
-   * or `null` if the element is not visible.
+   * or `null` if the element has no effect on layout (example: `display: none`).
```
OrKoN:
```suggestion
   * or `null` if the element is [not part of the layout](https://drafts.csswg.org/css-display-4/#box-generation) (example: `display: none`).
```

OrKoN (approved):

PR #11341, opened 2023-11-09
chore: update Expectations

Lightning00Blade:


jrandolf-google (approved):

test/TestExpectations.json
```diff
@@ -911,6 +911,12 @@
     "parameters": ["webDriverBiDi"],
     "expectations": ["FAIL"]
   },
+  {
+    "testIdPattern": "[oopif.spec] OOPIF should support OOP iframes getting detached",
```
OrKoN, line 915: I think it would be easier to just flip FF expectations that were marked FAIL to be PASS

OrKoN:
Updated Firefox in https://github.com/puppeteer/puppeteer/pull/11346 but it would be good to figure out the Chrome failures here

Lightning00Blade:
Note: There are test that are now passing but even after https://github.com/puppeteer/puppeteer/pull/11346 they are skippped.

PR #11342, opened 2023-11-09
test: pass `protocol` in `Puppeteer.connect` in tests

sadym-chromium:
Add an internal field `protocol: 'cdp' | 'webDriverBiDi'` to the `Browser` and use is in tests when `Puppeter.connect(...)` is called.

### DIFF
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -510,8 +510,8 @@ export class CdpBrowserContext extends BrowserContext {
     });
   }
 
-  override newPage(): Promise<Page> {
-    return this.#browser._createPageInContext(this.#id);
+  override async newPage(): Promise<Page> {
+    return await this.#browser._createPageInContext(this.#id);
   }
 
   override browser(): CdpBrowser {
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -53,6 +53,8 @@ import {TargetManagerEvent, type TargetManager} from './TargetManager.js';
  * @internal
  */
 export class CdpBrowser extends BrowserBase {
+  readonly protocol = 'cdp';
+
   static async _create(
     product: 'firefox' | 'chrome' | undefined,
     connection: Connection,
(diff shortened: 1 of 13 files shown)


packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -345,8 +347,8 @@ export class CdpBrowser extends BrowserBase {
     return this.#connection.url();
   }
 
-  override async newPage(): Promise<Page> {
-    return await this.#defaultContext.newPage();
+  override newPage(): Promise<Page> {
+    return this.#defaultContext.newPage();
```
sadym-chromium: Was required by `eslint`

Lightning00Blade: Readd `async` to the function and EsLint will enforce the `await`. Or there is another rule (maybe your IDE adds) that makes this disappear ?

packages/puppeteer-core/src/cdp/BrowserConnector.ts
```diff
@@ -29,6 +29,8 @@ import {isErrorLike} from '../util/ErrorLike.js';
 import {CdpBrowser} from './Browser.js';
 import {Connection} from './Connection.js';
 import type {ConnectOptions} from './ConnectOptions.js';
+import {BidiBrowser} from '../bidi/Browser';
```
Lightning00Blade: We should be doing this part with dynamic import similar to `launch`. Note Type imports are allowed

packages/puppeteer-core/src/cdp/Browser.ts
sadym-chromium: During `npm run format`,  the await is automatically added, after which another rule is happened to be broken:
>  514:12  error  Returning an awaited value that is not a promise is not allowed  @typescript-eslint/return-await

packages/puppeteer-core/src/cdp/BrowserConnector.ts
sadym-chromium: Rolled back that part and focused only on the tests in this PR

packages/puppeteer-core/src/cdp/Browser.ts
sadym-chromium: never mind, looks like it was a problem with local setup which is now gone

OrKoN (changes requested):
Please also update expectations for these new failures https://github.com/puppeteer/puppeteer/actions/runs/6824148748/job/18559474328?pr=11342

OrKoN (approved):

PR #11350, opened 2023-11-10
chore: implement `Puppeteer.connect` for BiDi over CDP browser

sadym-chromium:
Implementing Puppeteer.connect for pure BiDi is out of scope of this PR

### DIFF
--- a/packages/puppeteer-core/src/cdp/BrowserConnector.ts
+++ b/packages/puppeteer-core/src/cdp/BrowserConnector.ts
@@ -18,6 +18,7 @@ import type {
   IsPageTargetCallback,
   TargetFilterCallback,
 } from '../api/Browser.js';
+import type {BidiBrowser} from '../bidi/Browser.js';
 import type {ConnectionTransport} from '../common/ConnectionTransport.js';
 import {getFetch} from '../common/fetch.js';
 import {debugError} from '../common/util.js';
@@ -29,6 +30,7 @@ import {isErrorLike} from '../util/ErrorLike.js';
 import {CdpBrowser} from './Browser.js';
 import {Connection} from './Connection.js';
 import type {ConnectOptions} from './ConnectOptions.js';
+
 /**
  * Generic browser options that can be passed when launching any browser or when
  * connecting to an existing browser instance.
@@ -79,7 +81,7 @@ const getWebSocketTransportClass = async () => {
 
 /**
  * Users should never call this directly; it's called when calling
- * `puppeteer.connect`.
+ * `puppeteer.connect` with `protocol: 'cdp'`.
  *
  * @internal
  */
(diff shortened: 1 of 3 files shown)


sadym-chromium:
re-running CI

Joojoo4:
Does that mean the code that I deleted I'd now been coded back toy account
? As I'm being hacked ? Sorry if I affected anyone else but I don't want a
developer using code amd platforms and everything you can think of
basically cause  malice and affect t my life . I have a public. IP
Added In build to my atom ?  So I'm being stalked.My apps have depreciated
or made static or coded or altered to not work... my browser certificates
Te often not safe .. Jira app is somehow with code redirecting phone calls
amd messages
..that's like 1% of what happening....it been hokng on a long while ...any
advice would be appreciated .....half my manifests on ea h app has had
permissions amd functions tsle. Away so I can't even control my wi fi or
mobile data ....being going o. Years ...I now have to step up and stop it
..first I will ask the industry for help 2nd ... lawyer for advice....
anyway ...I I'm sorry of I hurt anyone's work but o now need to create
dcbidemce even if I'm unlawful in trying to stop it I believe it will be
pardoned as none had luster.ed or helped Mme and those person has nee.
Jacking me and manipulation g every part of my life for 7 years ...I now fk
sllh have the knowledge and understanding of platforms codes co certs to
have it
Shut down. ....The easy way or the hard. ...might ne extensible but worth
it...it's gone on too long.
Thanks again any advice would be appreciated!!

packages/puppeteer-core/src/cdp/BrowserConnector.ts
```diff
@@ -155,6 +123,40 @@ export async function _connectToCdpBrowser(
...
+ */
+export async function _connectToBiDiOverCdpBrowser(
+  options: BrowserConnectOptions & ConnectOptions
+): Promise<BidiBrowser> {
+  const {ignoreHTTPSErrors = false, defaultViewport = DEFAULT_VIEWPORT} =
+    options;
+
+  const connection = await getCdpConnection(options);
+
+  const version = await connection.send('Browser.getVersion');
```
OrKoN, line 140: I think we cannot rely for that on Firefox as it would not have any CDP in the future.

test/TestExpectations.json
```diff
@@ -3612,6 +3606,12 @@
     "parameters": ["firefox", "headful", "webDriverBiDi"],
     "expectations": ["FAIL"]
   },
+  {
+    "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Browser.disconnect should reject navigation when browser closes",
+    "platforms": ["linux"],
+    "parameters": ["cdp", "firefox", "headful"],
```
OrKoN: why are CDP tests affected?

packages/puppeteer-core/src/cdp/BrowserConnector.ts
sadym-chromium: This is a BiDi over CDP connector, not a pure BiDi. Implementing `Puppeteer.connect` for pure BiDi is out of scope of this PR

test/TestExpectations.json
sadym-chromium: reverted

OrKoN (approved):

PR #11351, opened 2023-11-10
tests: close process on Errors

Lightning00Blade:


test/src/fixtures.spec.ts
```diff
@@ -86,31 +82,43 @@ describe('Fixtures', function () {
...
+      const browser = await puppeteer.connect({
+        browserWSEndpoint: await wsEndPointPromise,
+      });
+      const promises = [
+        waitEvent(browser, 'disconnected'),
+        new Promise(resolve => {
+          res.on('close', resolve);
+        }),
+      ];
+      killProcess();
```
jrandolf, line 118: Do we need this here and in finally?

Lightning00Blade: Yes, as we want to close it externally to test the that we throw correct error from Puppeteer if the proccess closes.
In the finally we want to handle time that some error occurred before the first `killProcess();`

jrandolf-google (approved):

PR #11382, opened 2023-11-14
docs: fix typo in setViewport docs

ntnz:
Typo in the docs.

No.

Yes.

**Summary**
Change "isLandScape" to "isLandscape".

No.


OrKoN (changes requested):
Thanks for the PR. The docs are generated from source so the typo needs to be fixed in packages/puppeteer-core/src/api/Page.ts and then `npm run docs` would regenerate this md file. Also, there are two occurrences of typo and, perhaps, we could remove these remarks altogether because they actually duplicate descriptions here https://pptr.dev/api/puppeteer.viewport

PR #11393, opened 2023-11-14
chore: emit CDP event only on CdpSession

Lightning00Blade:
Emit BiDi+ (`cdp.*`) events only on CDP session.


jrandolf:
Why?

Lightning00Blade:
@jrandolf  For all CDP events we listen on the CDP session adapter, so we should not try to emit them on other event emitters.

OrKoN (approved):

PR #11394, opened 2023-11-14
chore: use a type for protocol

Lightning00Blade:


packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -441,5 +442,5 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
   /**
    * @internal
    */
-  abstract get protocol(): 'cdp' | 'webDriverBiDi';
+  abstract get protocol(): AutomationProtocol;
```
jrandolf: How about just `Protocol`

jrandolf: (I'm also implying we should renaming `Protocol` imports to something like `ProtocolTypes`)

Alternatively, `ProtocolName`

Lightning00Blade: Initially I did put it as just `Protocol` but it's rather obscure when seen next to other parts of the code. 
Let's discuss it offline how to proceed.

jrandolf-google (approved):

PR #11396, opened 2023-11-15
chore: mark where APIs don't work for BiDi

Lightning00Blade:
Just marking what does not work for BiDi, no changes to docs yet.

### DIFF
--- a/packages/puppeteer-core/src/api/Browser.ts
+++ b/packages/puppeteer-core/src/api/Browser.ts
@@ -399,6 +399,9 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
    *
    * {@link Page | Pages} can override the user agent with
    * {@link Page.setUserAgent}.
+   *
+   * @privateRemarks BiDi
+   * Not supported when using protocol `webDriverBiDi`
    */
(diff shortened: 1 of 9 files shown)


packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -399,6 +399,9 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
    *
    * {@link Page | Pages} can override the user agent with
    * {@link Page.setUserAgent}.
+   *
+   * @privateRemarks BiDi
+   * Not supported when using protocol `webDriverBiDi`
```
OrKoN, line 404: We will need to distinguish between BiDi and BiDi over CDP since we can support more things there. Also, let's write the remarks in the following way, wdyt?

```
{@link gettingStartedWithProtocolsLink | CDP}: Supported
{@link gettingStartedWithProtocolsLink | WebDriver BiDi}: Supported/Not supported
```

Later we can add 

```
{@link gettingStartedWithProtocolsLink | WebDriver BiDi+}: Supported/Not supported
```

to the list

PR #11405, opened 2023-11-16
docs: use link with placeholder

Lightning00Blade:
Note: Locally this will look very broken due to TsDoc not knowing how to handle the `PLACEHOLDER`.
But due to them all being behind `@privateRemarks` they will be stripped  from Public releases.

### DIFF
--- a/packages/puppeteer-core/src/cdp/Accessibility.ts
+++ b/packages/puppeteer-core/src/cdp/Accessibility.ts
@@ -130,7 +130,9 @@ export interface SnapshotOptions {
  *
  * @privateRemarks BiDi
  * {@link PROTOCOL_GET_STARTED_LINK_TEMPLATE | CDP}: Supported
+ *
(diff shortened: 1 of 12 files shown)


OrKoN (approved):

packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -129,8 +129,9 @@ export interface SnapshotOptions {
  * the "interesting" nodes of the tree.
  *
  * @privateRemarks BiDi
- * Not supported when using protocol `webDriverBiDi`
- * Support for BiDi+
+ * {@link PROTOCOL_GET_STARTED_LINK_TEMPLATE | CDP}: Supported
```
OrKoN, line 132: why is this one without a line break in-between and with BiDi+?

Lightning00Blade: Just missed it.

PR #11434, opened 2023-11-23
feat(11337): Upgrading to Docusaurus V3

### LINKED ISSUE #11337: [Feature]: Update pptr.dev to Docusaurus v3
### Feature description

Release notes: https://docusaurus.io/blog/releases/3.0

- [ ] See if anything needs to be update the file prior to the v3 update - https://docusaurus.io/blog/preparing-your-site-for-docusaurus-v3 
- [ ] Than run the steps described on the Migration guide - https://docusaurus.io/docs/migration/v3

TheJoin95:

Feature: Upgrading documentation website to Docusaurus V3 - Closes #11337 

Did not add tests, just verified that the documentation website was not presenting any kind of regression and everything was working smoothly.

**Summary**

Docusaurus V3 is running only in Node >= v18


docs/api/puppeteer.autofilldata.md
```diff
@@ -14,4 +14,4 @@ export interface AutofillData
 
 | Property   | Modifiers | Type                                                                                    | Description | Default |
 | ---------- | --------- | --------------------------------------------------------------------------------------- | ----------- | ------- |
-| creditCard |           | { number: string; name: string; expiryMonth: string; expiryYear: string; cvc: string; } |             |         |
```
OrKoN, line 17: do you know what caused changes in the markdown documentation?

TheJoin95: I needed to add the escape on the brackets according to the migration documentation attached to the issue otherwise MDX V3 was throwing an error

OrKoN: The md files are auto-generated though (npm run docs). I guess we need to fix the api docs generator too?

TheJoin95: Yes, updated by adding the html code for the curly brackets

OrKoN: Thanks a lot, we are reviewing! Looks good so far!.

website/package.json
```diff
@@ -15,18 +15,18 @@
...
   },
   "dependencies": {
-    "@docusaurus/core": "2.4.1",
-    "@docusaurus/plugin-client-redirects": "2.4.1",
-    "@docusaurus/preset-classic": "2.4.1",
-    "@mdx-js/react": "1.6.22",
+    "@docusaurus/core": "3.0.0",
+    "@docusaurus/plugin-client-redirects": "3.0.0",
+    "@docusaurus/preset-classic": "3.0.0",
+    "@mdx-js/react": "^3.0.0",
```
Lightning00Blade: Dependencies need to be pinned to the correct version. Please remove the `^`

tools/docgen/src/custom_markdown_documenter.ts
```diff
@@ -388,6 +388,8 @@ export class MarkdownDocumenter {
     pageContent = pageContent.replace(/<!-- -->/g, '');
     pageContent = pageContent.replace(/\\\*\\\*/g, '**');
     pageContent = pageContent.replace(/<b>|<\/b>/g, '**');
+    pageContent = pageContent.replace(/\{/g, '&#123;');
```
Lightning00Blade: Sadly this does not produce the desired result as we escape all the `{` and `}` for the page but we want to keep them in the context of `Signature` and `Example` else we get unusable code snippets:
image
image
We would need to escape only in places where we inject plain text and keep the rest as is.

TheJoin95: Thanks for catching this; I changed the replacement from the html code to the backslash escape, so it will look like the following:
image

image


And of course the same would be for a parameters table:
image

website/package.json
TheJoin95: Updated, thanks!

OrKoN:
I have rebased based on the latest changes.

tools/docgen/src/custom_markdown_documenter.ts
Lightning00Blade: This still creates the issue in a different way, please use the `Next` option from the version menu to verify. 

image
image

TheJoin95: Alright, I changed the angle by extending the [api-formatter MarkdownEmitter](https://github.com/puppeteer/puppeteer/pull/11434/files#diff-81e1618806e847089a218d79340803b2981e837496c6cf850f16da8c51584720R96) and by overriding the `getEscapeText` and `getTableEscapedText` methods

It should be ok this way.
I'm sorry for the previous back and forth

OrKoN (approved):
LGTM. Thanks a bunch!

PR #11441, opened 2023-11-24
docs: update faq

OrKoN:
Actualize some of the questions/answers/links.


mathiasbynens (approved):

website/versioned_docs/version-21.5.2/faq.md
```diff
@@ -166,7 +166,7 @@ that incorporate audio and video. (For example,
 [video playback/screenshots is likely to fail](https://github.com/puppeteer/puppeteer/issues/291).)
 There are two reasons for this:
 
-- Puppeteer is bundled with [Chrome for Testing](https://goo.gle/chrome-for-testing) — not the regular Chrome — and so by default, it
+- Puppeteer is bundled with [Chrome for Testing](https://developer.chrome.com/blog/chrome-for-testing/) — not the regular Chrome — and so by default, it
```
mathiasbynens, line 169: This is accurate for those older versions of Puppeteer/Chrome. However, people reading this in the future (e.g. after a Google search) might assume this is still the case unless we add a note here explaining that newer versions don’t have this problem. Worth doing? WDYT?

OrKoN: This is versioned documentation for the previous Puppeteer version and it will be replaced once we make a release.

mathiasbynens: Ah right, the website just links to the Markdown files in the repository. Let’s keep this as-is then.

PR #11451, opened 2023-11-27
feat: BiDi implementation of `Puppeteer.connect` for Firefox

sadym-chromium:
Add support of re-connecting to pure BiDi browser (Firefox). 
1. Split connection logic between `common`, `BiDi` and `CDP`.
3. When connecting via BiDi protocol, first try to connect via BiDi, and if the endpoint does not support BiDi, fall back to BiDi over CDP.
4. Adjust BiDi connection to fail gracefully if endpoint supports CDP.
5. Add `unbind` to BiDi connection to be able to keep WS transport open.

### DIFF
--- a/packages/puppeteer-core/src/bidi/Browser.ts
+++ b/packages/puppeteer-core/src/bidi/Browser.ts
@@ -253,7 +253,9 @@ export class BidiBrowser extends Browser {
     if (this.#connection.closed) {
       return;
     }
-    await this.#connection.send('browser.close', {});
+
+    // `browser.close` can close connection before the response is received.
+    await this.#connection.send('browser.close', {}).catch(debugError);
     await this.#closeCallback?.call(null);
     this.#connection.dispose();
   }
--- a/packages/puppeteer-core/src/bidi/BrowserConnector.ts
+++ b/packages/puppeteer-core/src/bidi/BrowserConnector.ts
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2023 Google Inc. All rights reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type {BrowserCloseCallback} from '../api/Browser.js';
+import {Connection} from '../cdp/Connection.js';
+import type {ConnectionTransport} from '../common/ConnectionTransport.js';
+import type {
+  BrowserConnectOptions,
+  ConnectOptions,
+} from '../common/ConnectOptions.js';
+import {UnsupportedOperation} from '../common/Errors.js';
+import {debugError, DEFAULT_VIEWPORT} from '../common/util.js';
+
+import type {BidiBrowser} from './Browser.js';
+import type {BidiConnection} from './Connection.js';
+
+/**
+ * Users should never call this directly; it's called when calling `puppeteer.connect`
+ * with `protocol: 'webDriverBiDi'`. This method attaches Puppeteer to an existing browser
+ * instance. First it tries to connect to the browser using pure BiDi. If the protocol is
+ * not supported, connects to the browser using BiDi over CDP.
+ *
+ * @internal
+ */
+export async function _connectToBiDiBrowser(
+  connectionTransport: ConnectionTransport,
+  url: string,
+  options: BrowserConnectOptions & ConnectOptions
+): Promise<BidiBrowser> {
+  const {ignoreHTTPSErrors = false, defaultViewport = DEFAULT_VIEWPORT} =
+    options;
+
+  const {bidiConnection, closeCallback} = await getBiDiConnection(
+    connectionTransport,
+    url,
+    options
+  );
+  const BiDi = await import(/* webpackIgnore: true */ './bidi.js');
+  const bidiBrowser = await BiDi.BidiBrowser.create({
+    connection: bidiConnection,
+    closeCallback,
+    process: undefined,
+    defaultViewport: defaultViewport,
+    ignoreHTTPSErrors: ignoreHTTPSErrors,
+  });
+  return bidiBrowser;
+}
+
+/**
+ * Returns a BiDiConnection established to the endpoint specified by the options and a
+ * callback closing the browser. Callback depends on whether the connection is pure BiDi
+ * or BiDi over CDP.
+ * The method tries to connect to the browser using pure BiDi protocol, and falls back
+ * to BiDi over CDP.
+ */
+async function getBiDiConnection(
+  connectionTransport: ConnectionTransport,
+  url: string,
+  options: BrowserConnectOptions
+): Promise<{
+  bidiConnection: BidiConnection;
+  closeCallback: BrowserCloseCallback;
+}> {
+  const BiDi = await import(/* webpackIgnore: true */ './bidi.js');
+  const {ignoreHTTPSErrors = false, slowMo = 0, protocolTimeout} = options;
+
+  // Try pure BiDi first.
+  const pureBidiConnection = new BiDi.BidiConnection(
+    url,
+    connectionTransport,
+    slowMo,
+    protocolTimeout
+  );
+  try {
+    const result = await pureBidiConnection.send('session.status', {});
+    if ('type' in result && result.type === 'success') {
+      // The `browserWSEndpoint` points to an endpoint supporting pure WebDriver BiDi.
+      return {
+        bidiConnection: pureBidiConnection,
+        closeCallback: async () => {
+          await bidiConnection.send('browser.close', {}).catch(debugError);
+        },
+      };
+    }
+  } catch (e: any) {
+    if (!('name' in e && e.name === 'ProtocolError')) {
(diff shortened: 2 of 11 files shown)


packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -323,7 +323,12 @@ export class BidiBrowser extends Browser {
...
   }
 
-  override disconnect(): void {
-    this;
+  override async disconnect(): Promise<void> {
+    try {
+      await this.#connection.send('session.end', {});
+      this.#connection.dispose();
+    } catch (e) {
+      console.log(e);
```
Lightning00Blade: I assume we don't want to show error here so we should use `debugError` to print only when explicitly enabled.

packages/puppeteer-core/src/bidi/Connection.ts
```diff
@@ -236,6 +240,15 @@ export class BidiConnection extends EventEmitter<BidiEvents> {
           return;
       }
     }
+    // Even if the response in not in BiDi protocol format but `id` is provided, reject
+    // the callback.
+    if ('id' in object) {
```
Lightning00Blade, line 252: Can you give more info here why this is needed? 
From my understanding we should always `type` to be populated from the spec perspective.

packages/puppeteer-core/src/cdp/BrowserConnector.ts
```diff
@@ -1,4 +1,4 @@
-/**
+/*
```
Lightning00Blade: Should be `/**` as before.

packages/puppeteer-core/src/common/BrowserConnector.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Users should never call this directly; it's called when calling
+ * `puppeteer.connect`. This method attaches Puppeteer to an existing browser instance.
+ *
+ * @remarks
+ *
+ * @internal
+ * @param options - Set of configurable options to set on the browser.
+ * @returns Promise which resolves to browser instance.
+ */
+export function _connectToBrowser(options: ConnectOptions): Promise<Browser> {
```
Lightning00Blade: We are only using it in one place, how about we move the logic there? 
If we do that we will need to import BiDi dynamically there as well. (Not sure if its not needed with current implementation)

packages/puppeteer-core/src/bidi/BrowserConnector.ts
```diff
@@ -0,0 +1,125 @@
...
+  ConnectOptions,
+} from '../common/ConnectOptions.js';
+import {UnsupportedOperation} from '../common/Errors.js';
+import {debugError} from '../common/util.js';
+import {isNode} from '../environment.js';
+
+import type {BidiBrowser} from './Browser.js';
+import type {BidiConnection} from './Connection.js';
+
+const DEFAULT_VIEWPORT = Object.freeze({width: 800, height: 600});
```
Lightning00Blade: `DEFAULT_VIEWPORT` and `getWebSocketTransportClass` should be extracted into common if possible.
The more important for me is `DEFAULT_VIEWPORT` as it may not brake things if changed.

packages/puppeteer-core/src/bidi/BrowserConnector.ts
```diff
@@ -0,0 +1,125 @@
...
+      if ('type' in result && result.type === 'success') {
+        // The `browserWSEndpoint` points to an endpoint supporting pure WebDriver BiDi.
+        return pureBidiConnection;
+      }
+    } catch (e: any) {
+      if (!('name' in e && e.name === 'ProtocolError')) {
+        // Unexpected exception not related to BiDi / CDP. Rethrow.
+        throw e;
+      }
+      wsTransport.close();
```
Lightning00Blade: Currently this only closes the `wsTransport` if it's expected error. I think we need to close it if any error occurrences.

packages/puppeteer-core/src/bidi/Browser.ts
OrKoN: Better use `debugError` instead of console.log. Also, we probably should call this.#connection.dispose(); even if session.end fails.

packages/puppeteer-core/src/bidi/Connection.ts
sadym-chromium: If the WS endpoint supports CDP but not WS, there will be no `type` in the response. However, we still want the command promise to be rejected, so that we understand that the protocol is not supported.

sadym-chromium:
Converting to draft, as there seems to be significant regression and lots of tests are failing now.

packages/puppeteer-core/src/bidi/BrowserConnector.ts
sadym-chromium: Referencing common from cdp/bidi would create a cyclic references. We can extend `_connectToBiDiBrowser` with an additional parameter, but that would be awkward as well. Any ideas on how to address it?

Lightning00Blade: I don't think it will, we can move it to `../common/util.js` similar to what we had done for `NETWORK_IDLE_TIME`.

sadym-chromium: got it. Done.

packages/puppeteer-core/src/bidi/BrowserConnector.ts
sadym-chromium: if the error is unexpected, we don't continue execution, so I don't think we need to clean up the state in such a case.

sadym-chromium:
Tests are still timing out, but the concerns are addressed.

packages/puppeteer-core/src/bidi/Browser.ts
sadym-chromium: done

packages/puppeteer-core/src/cdp/BrowserConnector.ts
sadym-chromium: done

packages/puppeteer-core/src/common/BrowserConnector.ts
sadym-chromium: added more logic in the common connector

test/TestExpectations.json
```diff
@@ -3647,6 +3647,12 @@
     "parameters": ["firefox", "headful", "webDriverBiDi"],
     "expectations": ["FAIL"]
   },
+  {
+    "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Browser.disconnect should reject navigation when browser closes",
```
sadym-chromium: FF does not support 2 simultaneous sessions.

sadym-chromium:
@OrKoN PTAL

test/TestExpectations.json
```diff
@@ -2133,7 +2133,7 @@
     "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Puppeteer.connect should be able to reconnect to a disconnected browser",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["cdp", "firefox"],
-    "expectations": ["PASS"]
+    "expectations": ["SKIP"]
```
OrKoN: this expectation is about CDP and I would not expect changes here

test/TestExpectations.json
OrKoN: This is a CDP expectation.

OrKoN: I think CDP-related expectations should not change.

sadym-chromium: Right, my bad. I'll check it out

sadym-chromium: Apparently it was an artifact. Removed.

test/TestExpectations.json
sadym-chromium: reverted expectations

test/TestExpectations.json
```diff
@@ -3653,6 +3653,12 @@
     "parameters": ["cdp", "firefox", "headful"],
     "expectations": ["FAIL", "PASS"]
   },
+  {
+    "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Puppeteer.connect should be able to reconnect",
+    "platforms": ["darwin"],
+    "parameters": ["firefox", "headless", "webDriverBiDi"],
+    "expectations": ["PASS"]
```
sadym-chromium: Here we go! Observable positive effect of the change )

test/TestExpectations.json
```diff
@@ -3653,6 +3653,12 @@
     "parameters": ["cdp", "firefox", "headful"],
     "expectations": ["FAIL", "PASS"]
   },
+  {
+    "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Puppeteer.connect should be able to reconnect",
+    "platforms": ["darwin"],
+    "parameters": ["firefox", "headless", "webDriverBiDi"],
```
```suggestion
    "parameters": ["firefox", "webDriverBiDi"],
```

test/TestExpectations.json
```diff
@@ -3653,6 +3653,12 @@
     "parameters": ["cdp", "firefox", "headful"],
     "expectations": ["FAIL", "PASS"]
   },
+  {
+    "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Puppeteer.connect should be able to reconnect",
+    "platforms": ["darwin"],
```
```suggestion
    "platforms": ["darwin", "linux", "win32"],
```

OrKoN (approved):

Zapeth:
Are there any settings required in Firefox for this to work, apart from `fission.webContentIsolationStrategy=0`?

I get `Firefox is not supported in BiDi over CDP mode` error when trying to connect() with `protocol: 'webDriverBiDi'` (I also tried setting `fission.bfcacheInParent=false`, but that only gets rid of the CDP warnings in Firefox output)

However, launch() with `{product: 'firefox', protocol: 'webDriverBiDi', executablePath: '/usr/bin/firefox'}` works

PR #11470, opened 2023-11-30
fix: close WebDriver BiDi session on disconnect

sadym-chromium:
1. `browser.disconnect` returns `Promise<void>` instead of `void`. This is required to be able to properly close BiDi session before disconnecting.
2. BiDi `browser.disconnect` sends `session.end` before disconnecting.
3. BiDi over CDP closes CDP connection when closed.
4. Fix BiDi browser close: dispose connection after the browser is closed.

Preparation for supporting `puppeteer.connect` by Firefox.

### DIFF
--- a/test/src/oopif.spec.ts
+++ b/test/src/oopif.spec.ts
@@ -418,7 +418,7 @@ describe('OOPIF', function () {
       return target.url().endsWith('dynamic-oopif.html');
     });
     await target.page();
-    browser1.disconnect();
+    await browser1.disconnect();
   });
 
   it('should support lazy OOP frames', async () => {
--- a/test/src/oopif.spec.ts
+++ b/test/src/oopif.spec.ts
@@ -418,7 +418,7 @@ describe('OOPIF', function () {
       return target.url().endsWith('dynamic-oopif.html');
     });
     await target.page();
-    await browser1.disconnect();
+    browser1.disconnect();
   });
 
   it('should support lazy OOP frames', async () => {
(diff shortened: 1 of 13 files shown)


OrKoN (changes requested):
Have not you said it's a non-breaking change?

sadym-chromium:
> Have not you said it's a non-breaking change?

Technically it is not breaking, as it should not change any existing workflows, but the API has changed.

sadym-chromium:
Converting to draft, as some tests are failing

OrKoN:
There are two concerns: 1) every breaking change results in a major puppeteer version which we don't want to bump so often 2) at the same time it does not look like it deserves a major version since there is no breaking behavior.

test/src/oopif.spec.ts
```diff
@@ -418,7 +418,7 @@ describe('OOPIF', function () {
       return target.url().endsWith('dynamic-oopif.html');
     });
     await target.page();
-    browser1.disconnect();
+    await browser1.disconnect();
```
OrKoN, line 421: can we still keep the old way of disconnecting to make sure it's not breaking?

OrKoN: btw, now that I think of it, did we verify that it does not work when the disconnect happens asynchronously?

sadym-chromium: As long as FF supports only a single session, we need to end a session before opening another one. How can we make sure the previous session is ended?

OrKoN: Can expect the firefox to finish processing session.end before starting to handle session.new?

OrKoN: let me ask

OrKoN: I mean there should be a mutex on the shared resource that is a session, right?

sadym-chromium: But we cannot close connection before the `session.end` command is processed. Meaning `disconnect` will not disconnect immediately, but rather eventually, which is confusing.

OrKoN (approved):

PR #11477, opened 2023-11-30
fix: remove CDP-specific preferences from defaults for Firefox

### LINKED ISSUE #11474: Don't set CDP-specific custom preferences for Firefox when running with the WebDriver BiDi protocol
As I just noticed (and @OrKoN confirmed) Puppeteer uses always the same preferences for Firefox independently if it is using the `CDP` or `WebDriver BiDi` protocol. Problematic here are the [following preferences](https://github.com/puppeteer/puppeteer/blob/957a8293bb1444fd51fd5673002a7781e8127c9d/packages/browsers/src/browser-data/firefox.ts#L224-L227):

* `'fission.bfcacheInParent: false`
* `'fission.webContentIsolationStrategy': 0`

Those should really only be set for CDP but never for WebDriver BiDi because they turn of some important features of Firefox and actually enable some kind of legacy (unsupported) mode.

@OrKoN is there an easy way to let Puppeteer set these two preferences automatically ONLY when the `CDP` protocol is used?

OrKoN:
Closes #11474

### DIFF
--- a/packages/puppeteer-core/src/node/FirefoxLauncher.ts
+++ b/packages/puppeteer-core/src/node/FirefoxLauncher.ts
@@ -113,7 +113,18 @@ export class FirefoxLauncher extends ProductLauncher {
 
     await createProfile(SupportedBrowsers.FIREFOX, {
       path: userDataDir,
-      preferences: extraPrefsFirefox,
+      preferences: {
+        ...extraPrefsFirefox,
+        ...(options.protocol === 'cdp'
+          ? {
+              // Temporarily force disable BFCache in parent (https://bit.ly/bug-1732263)
+              'fission.bfcacheInParent': false,
+
+              // Force all web content to use a single content process
+              'fission.webContentIsolationStrategy': 0,
+            }
+          : {}),
+      },
     });
 
     let firefoxExecutable: string;
--- a/packages/puppeteer-core/src/node/FirefoxLauncher.ts
+++ b/packages/puppeteer-core/src/node/FirefoxLauncher.ts
@@ -119,11 +119,14 @@ export class FirefoxLauncher extends ProductLauncher {
           ? {
               // Temporarily force disable BFCache in parent (https://bit.ly/bug-1732263)
               'fission.bfcacheInParent': false,
-
-              // Force all web content to use a single content process
-              'fission.webContentIsolationStrategy': 0,
             }
           : {}),
+        // Force all web content to use a single content process. TODO: remove
+        // this once Firefox supports mouse event dispatch from the main frame
+        // context. Once this happens, webContentIsolationStrategy should only
+        // be set for CDP. See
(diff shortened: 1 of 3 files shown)


Lightning00Blade (approved):

whimboo:
After a discussion with @OrKoN yesterday and some additional investigation on our side I found the problem that is causing some tests around out-of-process iframes to fail. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1867667 and we will work on it shortly. 

We should wait for it to be fixed on Nightly and then retrigger the tests on this PR.

whimboo (approved):
Code-wise this looks all fine. Lets see what happens with test results once we have the `contentWindow` issue fixed.

OrKoN:
@whimboo I believe this behavior is expected from Puppeteer's perspective. It does not work in Firefox only for now or is our expectation not spec-compliant?

whimboo:
> @whimboo I believe this behavior is expected from Puppeteer's perspective. It does not work in Firefox only for now or is our expectation not spec-compliant?

We are having a project early next year to add support for dispatching events from the parent process asynchronously. Until then the event needs to be submitted from the same browsing context.

Maybe we keep the site isolation feature off for now to keep it working, or turn it off by default and file an issue with details for a workaround for now in case cross-origin iframes are involved.

What would be your option?

OrKoN:
Let's keep it working for now by turning off the isolation feature.

packages/puppeteer-core/src/node/FirefoxLauncher.ts
```diff
@@ -113,7 +113,20 @@ export class FirefoxLauncher extends ProductLauncher {
 
     await createProfile(SupportedBrowsers.FIREFOX, {
       path: userDataDir,
-      preferences: extraPrefsFirefox,
+      preferences: {
+        ...extraPrefsFirefox,
+        ...(options.protocol === 'cdp'
+          ? {
+              // Temporarily force disable BFCache in parent (https://bit.ly/bug-1732263)
+              'fission.bfcacheInParent': false,
```
whimboo, line 121: We should still be able to get rid of this preference given that it doesn't affect the isolation feature.

OrKoN: Isn't required for the CDP version?

whimboo: Oh sorry, yes. With the usage of `? :` it is quite hard to read the code here in the diff viewer. So for CDP we need both preferences set, and for BiDi only `'fission.webContentIsolationStrategy': 0`.

whimboo: If you need a reference you can use https://bugzilla.mozilla.org/show_bug.cgi?id=1773393.

OrKoN: Thanks!

PR #11485, opened 2023-12-04
chore: update `mime`

### LINKED ISSUE #11481: Upgrade the mime package
Version 4.0.0 has breaking changes https://github.com/broofa/mime/blob/main/CHANGELOG.md

In particular, no CJS support. 

We need to upgrade and run `@dependabot unignore mime` once done. cc @jrandolf @Lightning00Blade

Lightning00Blade:
Closes #11481

Due to the change to ESM only we need to do `await import` workaround.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12052,6 +12052,7 @@
         "@pptr/testserver": "file:../packages/testserver",
         "diff": "5.1.0",
         "jpeg-js": "0.4.4",
+        "mime": "4.0.0",
         "pixelmatch": "5.3.0",
         "pngjs": "7.0.0",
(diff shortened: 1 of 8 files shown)


OrKoN:
@Lightning00Blade should we instead change out test runner to use ESM by default? currently, it still uses CJS I believe.

Lightning00Blade:
Tried to flip it but things start breaking down the line. The furthest I got was mocha being CJS and can't import ESM (did not look for solution).

Lightning00Blade:
Closing as it seems to be not trivial to switch to ESM for now.

PR #11495, opened 2023-12-05
fix: get port from created server

Lightning00Blade:
When lunching the Angular server if the port is in use the App will ask if you want to use another port. If you press Yes `@puppeteer/ng-schematics` need to find about the new `baseUrl` rather than the default one.


OrKoN (approved):

thongao1302:
V

PR #11506, opened 2023-12-06
test: add a test for eval followed by a mouse event

OrKoN:
Previously users of Puppeteer could rely on the fact that evaluate calls will be scheduled synchronously. It looks like we regressed w.r.t. to this behavior after starting to await for contexts and lazy args.

This PR adds a test and recovers the previous order of execution for the cases when there are no arguments and the context is readily available. 

Issue https://github.com/puppeteer/puppeteer/issues/11507

### DIFF
--- a/test/src/mouse.spec.ts
+++ b/test/src/mouse.spec.ts
@@ -458,4 +458,25 @@ describe('Mouse', function () {
       },
     ]);
   });
+
+  it('should evaluate before mouse event', async () => {
+    const {page, server} = await getTestState();
+
+    await page.goto(server.EMPTY_PAGE);
+    await page.goto(server.CROSS_PROCESS_PREFIX + '/input/button.html');
+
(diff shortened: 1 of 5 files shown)


test/src/mouse.spec.ts
```diff
@@ -458,4 +458,25 @@ describe('Mouse', function () {
...
+    const {page, server} = await getTestState();
+
+    await page.goto(server.EMPTY_PAGE);
+    await page.goto(server.CROSS_PROCESS_PREFIX + '/input/button.html');
+
+    using button = await page.waitForSelector('button');
+
+    const point = await button!.clickablePoint();
+
+    const result = page.evaluate(() => {
```
sadym-chromium, line 472: I'd either add comment or rename the `result` to `pointClickedPromise`

sadym-chromium (approved):

PR #11532, opened 2023-12-12
feat!: BiDi cookies

sadym-chromium:
Add support for cookie operations (get and set) via BiDi.

BREAKING CHANGE: 
* Replaces the CDP-specific interfaces with new protocol-independent ones:
  * `Cookie`;
  * `CookieParam`;
  * `DeleteCookiesRequest`
* Remove deprecated `Cookie.sourcePort` property.
* Add `Cookie.priority` property.


sadym-chromium:
The `cookies` tests are failing mostly because of the fact the BiDi does not support incognito browser contexts (BiDi [user contexts](https://github.com/w3c/webdriver-bidi/pull/570)), and tests rely on the cookies to be isolated between tests.

sadym-chromium:
The build is failing, because the [`storage.getCookies`](https://github.com/GoogleChromeLabs/chromium-bidi/pull/1593) is not merged in Mapper.

test/src/cookies.spec.ts
```diff
@@ -27,159 +27,190 @@ describe('Cookie specs', () => {
 
   describe('Page.cookies', function () {
     it('should return no cookies in pristine browser context', async () => {
-      const {page, server} = await getTestState();
-      await page.goto(server.EMPTY_PAGE);
-      await expectCookieEquals(await page.cookies(), []);
+      const {page, server, close} = await launch({});
```
sadym-chromium: Temporary switch to re-launching browser, as incognito mode is not implemented in BiDi

sadym-chromium:
Test state:
  Page.cookies
    ✅ should return no cookies in pristine browser context (291ms)
    ✅ should get a cookie (338ms)
    ✅ should properly report httpOnly cookie (308ms)
    ✅ should properly report "Strict" sameSite cookie (305ms)
    ✅ should properly report "Lax" sameSite cookie (302ms)
    ✅ should get multiple cookies (302ms)
    **❌ 1) should get cookies from multiple urls** (as is uses `setCookie`).

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -846,6 +846,51 @@ export class BidiPage extends Page {
...
+        size: c.size,
+        httpOnly: c.httpOnly,
+        secure: c.secure,
+        sameSite: this.#convertCookiesSameSite(c.sameSite),
+        // TODO: check default `expiry` value.
+        expires: c.expiry ?? -1,
+        // TODO: add proper value.
+        session: c.expiry === undefined || c.expiry <= 0,
+        // TODO: add proper value.
+        priority: 'Medium',
```
sadym-chromium: There is no `priority` in BiDi cookie.
Suggested solution: always set to `Medium`.

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -846,6 +846,51 @@ export class BidiPage extends Page {
...
+        // TODO: add proper value.
+        sameParty: false,
```
sadym-chromium: There is no `sameParty ` in BiDi cookie.
Suggested solution: always set to `false`.

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -846,6 +846,51 @@ export class BidiPage extends Page {
...
+        // TODO: add proper value.
+        sourceScheme: 'NonSecure',
```
sadym-chromium: There is no `sourceScheme` in BiDi cookie.
What is the expected behavior in case of BiDi?

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -846,6 +846,51 @@ export class BidiPage extends Page {
...
+        // TODO: add proper value.
+        sourcePort: -1,
```
sadym-chromium: There is no `sourcePort` in BiDi cookie.
What is the expected behavior in case of BiDi?

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -846,6 +846,51 @@ export class BidiPage extends Page {
...
+        // TODO: respect other `PartitionKey` values.
+        partitionKey: bidiCookies.result.partitionKey.sourceOrigin,
```
sadym-chromium: `partitionKey` in Puppeteer is a string, but in BiDi it can contain a `UserContext` (incognito context) as well, and it is extendable by vendor-specific keys.
What is the expected behavior in such cases?

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -846,6 +846,51 @@ export class BidiPage extends Page {
...
+        // TODO: add proper value.
+        partitionKeyOpaque: false,
```
sadym-chromium: Unsure in what actually `partitionKeyOpaque` means.

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -882,12 +940,40 @@ export class BidiPage extends Page {
...
-  override cookies(): never {
-    throw new UnsupportedOperation();
-  }
-
-  override setCookie(): never {
-    throw new UnsupportedOperation();
+  override async setCookie(
+    ...cookies: Protocol.Network.CookieParam[]
+  ): Promise<void> {
+    for (const cookie of cookies) {
```
sadym-chromium, line 839: What is expected behavior in case of one of the cookie failed to be set?

packages/puppeteer-core/src/bidi/Page.ts
OrKoN: I think the partitionKey in Puppeteer is always UserContext-specific. So we don't need to include the user context here. What we need is to filter out cookies that do not belong to the current page's context.

OrKoN: It looks like it's not possible to get such cookies so it does not look like any action is needed. Eventually, we need to change the Puppeteer API to support partitioned cookies across the browsers better.

packages/puppeteer-core/src/bidi/Page.ts
OrKoN: I think there is not much we can do except for some pre-validation for common cases (if it is common)

test/TestExpectations.json
```diff
@@ -3689,6 +3689,30 @@
     "parameters": ["cdp", "chrome", "new-headless"],
     "expectations": ["FAIL"]
   },
+  {
+    "testIdPattern": "[cookies.spec] Cookie specs Page.setCookie should default to setting secure cookie for HTTPS websites",
+    "platforms": ["darwin"],
+    "parameters": ["chrome", "headless", "webDriverBiDi"],
```
OrKoN: Remove "headless" as it should pass in all configurations

OrKoN:
Setting DO NOT MERGE as it should not be landed before all breaking changes for the next release are ready.

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -774,6 +775,136 @@ export class BidiPage extends Page {
...
+        context: this.mainFrame()._id,
+      },
+    });
+    return bidiCookies.result.cookies
+      .map(c => {
+        return this.#bidiToPuppeteerCookie(c);
+      })
+      .filter(c => {
+        return normalizedUrls.some(u => {
+          return BidiPage.#testUrlMatchCookie(c, u);
```
OrKoN: let's use full name `cookie` and `url` for readability

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -774,6 +775,136 @@ export class BidiPage extends Page {
...
+      domain: bidiCookie.domain,
+      path: bidiCookie.path,
+      size: bidiCookie.size,
+      httpOnly: bidiCookie.httpOnly,
+      secure: bidiCookie.secure,
+      sameSite: this.#convertCookiesSameSiteBiDiToCdp(bidiCookie.sameSite),
+      expires: bidiCookie.expiry ?? -1,
+      session: bidiCookie.expiry === undefined || bidiCookie.expiry <= 0,
+      // Extending with CDP-specific properties with `goog:` prefix.
+      ...(bidiCookie['goog:sameParty'] !== undefined
```
Lightning00Blade: Can we not hard code them but make them drop the `goog` prefix via a helper, we should make these a helper function as we may use it in other places as well.

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -774,6 +775,136 @@ export class BidiPage extends Page {
...
+  }
+
+  /**
+   * Check domains match.
+   * According to cookies spec, this check should match subdomains as well, but CDP
+   * implementation does not do that, so this method matches only the exact domains, not
+   * what is written in the spec:
+   * https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.3
+   */
+  static #testUrlMatchCookieHostname(
```
OrKoN: some methods are static and some not? Let's move all of them as module functions to Cookie.ts

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -774,6 +775,136 @@ export class BidiPage extends Page {
...
+  static #testUrlMatchCookie(cookie: Cookie, url: URL): boolean {
+    const normalizedUrl = new URL(url);
+    console.assert(cookie !== undefined);
+    if (!this.#testUrlMatchCookieHostname(cookie, normalizedUrl)) {
+      return false;
+    }
+    return this.#testUrlMatchCookiePath(cookie, normalizedUrl);
+  }
+
+  #bidiToPuppeteerCookie(bidiCookie: Bidi.Network.Cookie): Cookie {
```
OrKoN: let's keep the bidi to Puppeteer mapping functions in this file but make it a module-level function (it does not use any of the page instance state)

test/TestExpectations.json
```diff
@@ -1583,72 +1583,228 @@
     "parameters": ["firefox", "webDriverBiDi"],
     "expectations": ["PASS"]
   },
+  {
+    "testIdPattern": "[cookies.spec] Cookie specs Page.cookies should get a cookie",
+    "platforms": ["darwin"],
```
Lightning00Blade: Let's enable all test and mark the once that fail, that makes it much easier to update the expectations if we add a feature or bugfix.
```suggestion
   "testIdPattern": "[cookies.spec] Cookie specs Page.cookies *",
   "platforms": ["darwin", "linux", "win32"],
```

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -774,6 +775,136 @@ export class BidiPage extends Page {
...
+      ...(bidiCookie['goog:sourceScheme'] !== undefined
+        ? {sourceScheme: bidiCookie['goog:sourceScheme']}
+        : {}),
+      ...(bidiCookie['goog:partitionKey'] !== undefined
+        ? {partitionKey: bidiCookie['goog:partitionKey']}
+        : {}),
+      ...(bidiCookie['goog:partitionKeyOpaque'] !== undefined
+        ? {partitionKeyOpaque: bidiCookie['goog:partitionKeyOpaque']}
+        : {}),
+      // `priority` is not supported by Puppeteer.
```
OrKoN: Can we map priority in this PR given it is a breaking change?

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -774,6 +775,136 @@ export class BidiPage extends Page {
...
+    }
+    return false;
+  }
+
+  /**
+   * Checks the cookie matches the URL according to the spec:
+   */
+  static #testUrlMatchCookie(cookie: Cookie, url: URL): boolean {
+    const normalizedUrl = new URL(url);
+    console.assert(cookie !== undefined);
```
Lightning00Blade:
```suggestion
    assert(cookie !== undefined);
```

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -774,6 +775,136 @@ export class BidiPage extends Page {
     });
   }
 
+  override async cookies(...urls: string[]): Promise<Cookie[]> {
```
OrKoN, line 778: I think it makes sense to use common/Cookie in all Page APIs related to cookies?

test/TestExpectations.json
sadym-chromium: good idea

packages/puppeteer-core/src/bidi/Page.ts
sadym-chromium: `Cookie` and `CookieParams` are different. Do you suggest to copy CDP `CookieParams` as well?

packages/puppeteer-core/src/bidi/Page.ts
sadym-chromium: done

packages/puppeteer-core/src/bidi/Page.ts
sadym-chromium: done

packages/puppeteer-core/src/bidi/Page.ts
OrKoN: Right, I am not sure if we need to copy or if params could be derived from Cookie. Currently, the interface is like

```
abstract cookies(...urls: string[]): Promise<Cookie[]>;

abstract deleteCookie(
    ...cookies: Protocol.Network.DeleteCookiesRequest[]
  ): Promise<void>;
  
abstract setCookie(...cookies: Protocol.Network.CookieParam[]): Promise<void>;
```

I think it makes sense to remove CDP-specific structs from the API as part of the same breaking change (so that we do not need another breaking change).

OrKoN: Perhaps `type DeleteCookiesRequest = Pick<Cookie, 'attr1'|'attr2'>` could be done like that? Similar with the CookieParam.

sadym-chromium: It will end up with something like this, as some fields are optional, and `url` is not present in the `Cookie`. So I'm inclined to use a copy rather then this tricky `Pick`:
```
export type CookieParam =
  Pick<Cookie, 'name' |
    'value' |
    'sameSite' |
    'expires' |
    'priority' |
    'sameParty' |
    'sourceScheme' |
    'sourcePort' |
    'partitionKey'>
  & {
  url?: string;
  domain?: string;
  path?: string;
  secure?: boolean;
  httpOnly?: boolean;
};
```

OrKoN: Copying sounds good to me.

sadym-chromium: done

packages/puppeteer-core/src/bidi/Page.ts
sadym-chromium: done

packages/puppeteer-core/src/bidi/Page.ts
sadym-chromium: done

test/TestExpectations.json
```diff
@@ -1584,22 +1590,34 @@
...
+    "testIdPattern": "[cookies.spec] Cookie specs Page.cookies should not get cookies from subdomain",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["cdp", "firefox"],
-    "expectations": ["PASS"]
+    "expectations": ["FAIL"]
+  },
+  {
+    "testIdPattern": "[cookies.spec] Cookie specs Page.deleteCookie should work",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["chrome", "webDriverBiDi"],
```
OrKoN, line 1613: is it an expected failure?

sadym-chromium: yes,  as "delete" is not implement yet

OrKoN (approved):

OrKoN (approved):

PR #11544, opened 2023-12-13
chore: update sandbox for `ng-schematics`

Lightning00Blade:
Opening as a separate PR to discuss the refactoring of `ng-schmatics` sandbox.


jrandolf:
I'd prefer if we use a [build pattern](https://en.wikipedia.org/wiki/Builder_pattern) instead of the one-off functions.

jrandolf-google (approved):

PR #11093, opened 2023-10-06
feat: download chrome-headless-shell by default and use it for the old headless mode

OrKoN:
With this PR, Puppeteer will download two binaries by default: a Chrome for Testing binary and a chrome-headless-shell of the same version. `chrome-headless-shell` will be used as a replacement for the old headless mode. 

Two new configurations are added:

1) `PUPPETEER_SKIP_CHROME_DOWNLOAD` to skip Chrome binary downloads (only chrome-headless-shell will be downloaded).
2) `PUPPETEER_SKIP_CHROME_HEADLESS_SHELL_DOWNLOAD` to skip chrome-headless-shell  binary downloads (only Chrome for Testing will be downloaded).

Use these variables if you only want to use the old headless mode (chrome-headless-shell) or only the new-headless/headful mode (Chrome for Testing). The existing configuration `PUPPETEER_SKIP_DOWNLOAD` would still apply to both binaries.

BEGIN_COMMIT_OVERRIDE
chore: download chrome-headless-shell by default and use it for the old headless mode
END_COMMIT_OVERRIDE

PR #11545, opened 2023-12-13
revert: "feat: download chrome-headless-shell by default and use it for the old headless mode"

jrandolf-google:
We need to revert to 119 which doesn't work for `chrome-headless-shell`. See puppeteer/puppeteer#11093

Reverts puppeteer/puppeteer#11093

BEGIN_COMMIT_OVERRIDE
chore: revert "feat: download chrome-headless-shell by default and use it for the old headless mode"
END_COMMIT_OVERRIDE


Lightning00Blade (approved):

jrandolf:
@OrKoN We can reland this after 121 becomes stable.

OrKoN:
What was the issue? (the link from the description seems wrong)

PR #11558, opened 2023-12-15
chore: abstract web worker API

jrandolf-google:
In preparation of incoming WebDriver BiDi web workers. Also updates docs.


packages/puppeteer-core/src/api/WebWorker.ts
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type {CDPSession} from '../api/CDPSession.js';
+import type {Realm} from '../api/Realm.js';
```
Lightning00Blade: Will need a format after:
```suggestion
import type {CDPSession} from './CDPSession.js';
import type {Realm} from './Realm.js';
```

test/src/worker.spec.ts
```diff
@@ -15,7 +15,7 @@
  */
 
 import expect from 'expect';
-import type {WebWorker} from 'puppeteer-core/internal/cdp/WebWorker.js';
```
Lightning00Blade, line 18: This looks incorrect, we should use the API one when referring to the type here

packages/puppeteer-core/src/api/WebWorker.ts
```diff
@@ -0,0 +1,146 @@
...
+   *
+   * @param func - Function to be evaluated.
+   * @param args - Arguments to pass into `func`.
+   * @returns A {@link JSHandle | handle} to the return value of `func`.
+   */
+  async evaluateHandle<
+    Params extends unknown[],
+    Func extends EvaluateFunc<Params> = EvaluateFunc<Params>,
+  >(
+    pageFunction: Func | string,
```
Lightning00Blade: Else documentation is broken
```suggestion
    func: Func | string,
```

Lightning00Blade (approved):
LGTM + some comments

Lightning00Blade (approved):

PR #11560, opened 2023-12-15
chore: update chromium-bidi

jrandolf-google:


Lightning00Blade (changes requested):
Package-lock seems broken again,

Lightning00Blade (approved):

PR #11574, opened 2023-12-19
chore: use node test runner

Lightning00Blade:
Use Node Test runner for Unit testing, rename the `test` to `unit` to better reflect test type.


packages/ng-schematics/test/src/e2e.test.ts
```diff
@@ -6,19 +8,19 @@ import {
   setupHttpHooks,
 } from './utils.js';
 
-describe('@puppeteer/ng-schematics: e2e', () => {
+void describe('@puppeteer/ng-schematics: e2e', () => {
```
OrKoN, line 11: do we need void test promises?

Lightning00Blade: We have `@typescript-eslint/no-floating-promises` so EsLint will fail. Sadly there is not a way to provide a Ignore Pattern to said rule.

Lightning00Blade: Failed updating to ESM as some Testing from Angular is only CJS

OrKoN (approved):

PR #11578, opened 2023-12-20
fix: disable GFX sanity window for Firefox and enable WebDriver BiDi CI jobs for Windows

whimboo:
With test automation enabled the GFX sanity window should be disabled. So far we seem to have completely forgotten about that for Puppeteer while it is present for geckodriver and Marionette.

Note that this can cause issues during shutdown as reported on https://github.com/mozilla/pdf.js/issues/17396.

This PR might need an additional commit given that more Puppeteer tests might pass now. Lets see the results from various platforms in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -231,6 +231,7 @@ jobs:
         os:
           - ubuntu-latest
           - macos-latest
+          - windows-latest
         suite:
           - firefox-bidi
           - firefox-headful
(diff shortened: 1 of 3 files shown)


whimboo:
I just noticed that Firefox tests are disabled in CI for Windows. So maybe this change will make tests way lesser flaky and we can re-enable the platform for Firefox.

whimboo:
Firefox tests on Windows are now all passing. Most likely it was indeed due to this GFX window. We have some further failing tests that we have to investigate and fix but it's not looking that bad. 

This PR is now ready to be reviewed. Thanks.

OrKoN:
@whimboo is it possible to optimize the runtime of the third windows shard?

`CI / firefox-bidi tests on windows-latest (3-4) (pull_request) Successful in 13m` Ideally it should take around 5 min but looks like it's much slower on Windows.

OrKoN (approved):

whimboo:
As spoken with @OrKoN he is fine with these changes. With the last pushes I only fixed the linter job for test expectations and combined several commits to get a better structure. Everything runs fine now.

PR #11600, opened 2023-12-28
fix: improve reliability of exposeFunction

### LINKED ISSUE #8106: [Bug]: exposeFunction() sometimes gets stuck waiting forever, yet there's no documented timeout parameter.
### Bug description

```
console.log("Exposing function...");

await page.exposeFunction('myFunction', () => {
    // whatever...
    return;
});

console.log("Function exposed.");
```

About 80% of the times this works as expected.
Occasionally, however, this will get stuck waiting forever after this output:
```
Exposing functions...
```

As far as I understand, exposing a function is an instantaneous operation except for the part of sending a request to the browser (or in my case to chrome.browserless.io) and getting a response from it, so the only case where I can imagine this remaining stuck waiting without throwing an exception, is a network issue or the browser failing to respond.

But then there should be a configurable timeout parameter and I don't see any documented.

### Puppeteer version

13.4.0

### Node.js version

17.0.1

### npm version

8.1.0

Linux

### Relevant log output

_No response_

OrKoN:
This PR adds the following fixes to the implementation of CDP's exposeFunction:

1) ensure a binding cannot be double-wrapped. This might happen during the loading of frames when there is a race condition between the client and the preload scripts.
2) ensure the context is cleared when a new world is created. If there were pending evaluate calls on the old context, they would hang forever. Rejecting the contexts make sure the call can be rejected to and the waiters can recover.
3) Skip installing functions into frames that have not started loading yet. There are some instances where a frame might never start loading (it's unclear how but I observed that on codepen.io examples). If they will start loading later, it should be handled by the preload scripts (it does not look like this works currently).
 
 Closes #8106

### DIFF
--- a/packages/puppeteer-core/src/common/util.ts
+++ b/packages/puppeteer-core/src/common/util.ts
@@ -365,6 +365,13 @@ export function addPageBinding(type: string, name: string): void {
   // @ts-expect-error: In a different context.
   const callCdp = globalThis[name];
 
+  // Depending on the frame loading state either Runtime.evaluate or
+  // Page.addScriptToEvaluateOnNewDocument might succeed. Let's check that we
+  // don't re-wrap Puppeteer's binding.
+  if (callCdp[Symbol.toStringTag] === 'PuppeteerBinding') {
+    return;
+  }
+
   // We replace the CDP binding with a Puppeteer binding.
   Object.assign(globalThis, {
     [name](...args: unknown[]): Promise<unknown> {
@@ -404,6 +411,8 @@ export function addPageBinding(type: string, name: string): void {
       });
     },
   });
+  // @ts-expect-error: In a different context.
+  globalThis[name][Symbol.toStringTag] = 'PuppeteerBinding';
 }
 
 /**
--- a/packages/puppeteer-core/src/cdp/Frame.ts
+++ b/packages/puppeteer-core/src/cdp/Frame.ts
@@ -96,6 +96,11 @@ export class CdpFrame extends Frame {
   updateClient(client: CDPSession, keepWorlds = false): void {
     this.#client = client;
     if (!keepWorlds) {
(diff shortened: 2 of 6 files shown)


packages/puppeteer-core/src/common/util.ts
```diff
@@ -365,6 +365,13 @@ export function addPageBinding(type: string, name: string): void {
   // @ts-expect-error: In a different context.
   const callCdp = globalThis[name];
 
+  // Depending on the frame loading state either Runtime.evaluate or
+  // Page.addScriptToEvaluateOnNewDocument might succeed. Let's check that we
+  // don't re-wrap Puppeteer's binding.
+  if (callCdp.toString().includes('This is the Puppeteer binding.')) {
```
Lightning00Blade: I don't like looking for a specific Comment in the function. How about we do this after:
```ts
Object.assign(globalThis[name],{
  toString(): {
    return 'PuppeteerBinding'
  }
}
```
And this turns into
```suggestion
  if (callCdp.toString() === 'PuppeteerBinding') {
```

OrKoN: Not sure, it might be confusing for debugging if the function toString() returns something else. But we can tag it in a different way.

Lightning00Blade: Yeah that's true, lets then modify the `toStringTag` as its
```ts
Object.assign(globalThis[name],{
  [Symbol.toStringTag]: 'PuppeteerBinding',
}
```
```ts
if (callCdp[Symbol.toStringTag] === 'PuppeteerBinding') {
```
image

Lightning00Blade (approved):

eoghanmurray:
One possibility where the exposeFunction still fails is in 

`<iframe src="doc.pdf">`

PR #11604, opened 2023-12-29
feat: allow converting other targets to pages

### LINKED ISSUE #9262: Provide a safe way of overriding _isPageTargetCallback
I've recently been looking in to testing WebExtensions (e.g https://github.com/puppeteer/puppeteer/issues/2486). With the [action.openPopup API](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/action/openPopup), it's possible to trigger the extension popup by running code from the background page. This causes a new context to load which it would be useful to be able to automate.

Using the following, it actually is possible to get a reference to the new target:

```ts
const popup = await browser.waitForTarget(
  (target) =>
    target.type() === "other" &&
    target.url() === `chrome-extension://${extensionId}${path}`
);
```

However, calling `target.page()` returns `null` because the type is `other`: https://github.com/puppeteer/puppeteer/blob/1bbecb3bae0503fd1abf89c8e0a8be454628d679/packages/puppeteer-core/src/common/Browser.ts#L217-L227

Of course, this does make sense and the best solution would be accommodating extension testing in the Chrome Dev Tools protocol by returning a new "popup" type. At least one relevant issue is likely this one: https://crbug.com/1223710

In the meantime, however, removing the `_isPageTargetCallback` check does give you a page which can be automated! It sounds like some features are missing - https://github.com/puppeteer/puppeteer/issues/1215#issuecomment-486275427 - so exposing this should likely not be the default. That said, in my short amount of testing it seems to work well enough that allowing developers a way of experimenting with this functionality without needing to entirely fork Puppeteer may be desirable.

OrKoN:
Closes #9262


packages/puppeteer-core/src/bidi/Target.ts
```diff
@@ -38,8 +39,8 @@ export abstract class BidiTarget extends Target {
     this._browserContext = browserContext;
   }
 
-  override async worker(): Promise<null> {
-    return null;
+  override asPage(): Promise<Page> {
+    throw new Error('Not implemented');
```
Lightning00Blade: Use the Unsupported Error.

Lightning00Blade (approved):

PR #11614, opened 2024-01-02
ci: configure daily Canary Chrome tests

OrKoN:
Setup a workflow to test against the latest Canary build.


.github/workflows/canary.yml
```diff
@@ -0,0 +1,71 @@
+name: CI
```
Lightning00Blade:
```suggestion
name: Puppeteer Canary CI
```

Lightning00Blade (approved):

PR #11636, opened 2024-01-05
fix: wait for WebDriver BiDi browser to close gracefully

OrKoN:
The timeout is 5sec.

### DIFF
--- a/packages/puppeteer-core/src/node/ProductLauncher.ts
+++ b/packages/puppeteer-core/src/node/ProductLauncher.ts
@@ -16,6 +16,13 @@ import {
   computeExecutablePath,
 } from '@puppeteer/browsers';
 
+import {
+  firstValueFrom,
+  from,
+  map,
+  race,
+  timer,
+} from '../../third_party/rxjs/rxjs.js';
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/node/ProductLauncher.ts
```diff
@@ -236,7 +237,14 @@ export abstract class ProductLauncher {
         await browserProcess.close();
       }
     } else {
-      await browserProcess.close();
+      try {
+        // Wait for a possible graceful shutdown.
+        await firstValueFrom(
```
OrKoN: @Lightning00Blade am I holding it right?

Lightning00Blade: This will work but I would use RxJs for the whole thing. (Not sure its the better thing to do style wise)
 ```ts
// Wait for a possible graceful shutdown.
 await firstValueFrom(
  race(
    from(browserProcess.hasClosed()),
    timer(5000).pipe(
      map(
        () => {
          return from(browserProcess.close())
        } 
      )
    ) 
 )
 ```

Lightning00Blade (approved):

PR #11649, opened 2024-01-08
refactor: implement bidi/core

jrandolf-google:
# `bidi/core`

`bidi/core` is a lightweight, low-level abstraction over the WebDriver BiDi protocol that operates using typical OOP design principles. It's expected that Puppeteer will not interact directly with WebDriver BiDi, but over this abstraction.

## For reviewers

Each commit will contain the appropriate information regarding it's content. Do not attempt to read the entire PR in one swoop; it'll be a doozy.

## What's new

Consider the `browsingContext` API (https://w3c.github.io/webdriver-bidi/#type-browsingContext-Info) in WebDriver BiDi. Originally, Puppeteer would handle commands to this API by directly passing IDs around.

For example,

```ts
command.send('browsingContext.activate', {
  context: id
});
```

With `bidi/core`, the flow is simpler; instead we have,

```ts
context.activate();
```

By using `bidi/core`, we are able to *fill the gaps* of WebDriver BiDi that are expected to be filled by client-side implementations without cluttering Puppeteer with these details.

## Remarks

 - Note that `connection` getters are private/protected. Previously we would make this public, but doing so caused confusion about the origins of the connection. With private/protected getters, child classes are forced to resolve the full path to get the connection. For example, for Realms, this is `this.context.context.browser.session.connection`.


Lightning00Blade:
It looks like some of the commits are unrelated to the change you are proposing please extract them in sperate PR to ease the review of this one.

packages/puppeteer-core/src/util/Deferred.ts
```diff
@@ -37,26 +75,24 @@ export class Deferred<T, V extends Error = Error> {
...
 
   reject(error: V | TimeoutError): void {
     if (this.#isRejected || this.#isResolved) {
       return;
     }
     this.#isRejected = true;
-    this.#finish(error);
+    clearTimeout(this.#timeoutId);
+    this.#value = error;
+    this.#reject(error);
```
Lightning00Blade: There is a reason why `Deferred` was changed not use reject. Please rereview the PR that made the change or lets chat offline.

packages/puppeteer-core/src/bidi/Input.ts
```diff
@@ -288,7 +288,7 @@ export class BidiKeyboard extends Keyboard {
     key: KeyInput,
     _options?: Readonly<KeyDownOptions>
   ): Promise<void> {
-    await this.#page.connection.send('input.performActions', {
+    await this.#page.#connection.send('input.performActions', {
```
Lightning00Blade: Shouldn't this fail at run time?

jrandolf:
@Lightning00Blade Please review `Target`s and navigation related code. That is complete.

packages/puppeteer-core/src/bidi/Input.ts
jrandolf: Unfinished.

packages/puppeteer-core/src/util/Deferred.ts
jrandolf: Ah, yes, I remember.

packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -87,6 +87,7 @@ export abstract class Realm implements Disposable {
     return await waitTask.result;
   }
 
+  // Used for ARIA query selectors.
```
Lightning00Blade: Use JsDoc else not every IDE picks it up.

packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -69,189 +61,98 @@ export class BidiBrowser extends Browser {
...
 
-    for (const [eventName, handler] of this.#connectionEventHandlers) {
-      this.#connection.on(eventName, handler);
-    }
+    this.#target = new BidiBrowserTarget(this);
   }
 
-  override userAgent(): never {
-    throw new UnsupportedOperation();
+  get #browser() {
```
Lightning00Blade: This is confusing to have a browser on the browser.

packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -69,189 +61,98 @@ export class BidiBrowser extends Browser {
...
-  get connection(): BidiConnection {
-    return this.#connection;
+  override userAgent(): never {
+    throw new UnsupportedOperation();
   }
 
   override wsEndpoint(): string {
-    return this.#connection.url;
+    // SAFETY: `webSocketUrl` is true in `Session.create`.
+    return this.session.capabilities.webSocketUrl as string;
```
Lightning00Blade: This is fixed on the spec and in Mapper (need a new version). So lets add a TODO to remove.

packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -262,66 +163,49 @@ export class BidiBrowser extends Browser {
     _options?: BrowserContextOptions
   ): Promise<BidiBrowserContext> {
     // TODO: implement incognito context https://github.com/w3c/webdriver-bidi/issues/289.
-    const context = new BidiBrowserContext(this, {
-      defaultViewport: this.#defaultViewport,
-      isDefault: false,
+    await new Promise(resolve => {
```
Lightning00Blade: Why?

packages/puppeteer-core/src/bidi/Dialog.ts
```diff
@@ -1,43 +1,23 @@
...
   }
 
-  /**
-   * @internal
-   */
-  override async handle(options: {
+  protected override async handle(options: {
     accept: boolean;
-    text?: string;
+    text?: string | undefined;
```
Lightning00Blade: How is this type diffrent?

packages/puppeteer-core/src/bidi/Target.ts
```diff
@@ -11,114 +11,92 @@ import {UnsupportedOperation} from '../common/Errors.js';
...
   override async createCDPSession(): Promise<CDPSession> {
-    const {sessionId} = await this._browsingContext.cdpSession.send(
-      'Target.attachToTarget',
-      {
-        targetId: this._browsingContext.id,
-        flatten: true,
-      }
-    );
-    return new CdpSessionWrapper(this._browsingContext, sessionId);
+    const page = await this.asPage();
```
Lightning00Blade: This does not seem correct, you should be able to create session on other targets as well.

packages/puppeteer-core/src/bidi/Target.ts
```diff
@@ -11,114 +11,92 @@ import {UnsupportedOperation} from '../common/Errors.js';
...
-
-  constructor(
-    browserContext: BidiBrowserContext,
-    browsingContext: BrowsingContext
-  ) {
-    super(browserContext, browsingContext);
-
-    this.#page = new BidiPage(browsingContext, browserContext, this);
+  override browser(): BidiBrowser {
+    return this.#context.browser();
```
Lightning00Blade: I think this is the perfect example of duplicate naming, I need to check import to verify that's the Puppeteer Browser for BiDi and not BiDi/core Browser module.

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -0,0 +1,82 @@
...
+      }
+    }
+  }
+
+  @throwIfDisposed((browser: Browser) => {
+    // SAFETY: By definition of `disposed`, `#reason` is defined.
+    return browser.#reason!;
+  })
+  async close(): Promise<void> {
+    await this.#connection.send('browser.close', {});
```
Lightning00Blade: Wasn't the idea to be `this.#session.browser.close()`?

packages/puppeteer-core/src/bidi/Browser.ts
jrandolf: Well, it's difficult to come up with a name here. It's clear from context that it's the Browser coming from WebDriver BiDi though.

Do you have any suggestions?

packages/puppeteer-core/src/bidi/Browser.ts
jrandolf: To emulate asynchronous code.

docs/api/puppeteer.awaitablepredicate.md
```diff
@@ -0,0 +1,13 @@
+---
+sidebar_label: AwaitablePredicate
```
Lightning00Blade: This is a nice change, it would be good to extract it.This is a nice change, it would be good to extract it.

packages/puppeteer-core/src/api/BrowserContext.ts
```diff
@@ -114,10 +114,14 @@ export abstract class BrowserContext extends EventEmitter<BrowserContextEvents>
    * );
    * ```
    */
-  abstract waitForTarget(
+  async waitForTarget(
```
Lightning00Blade: Same here, nice change, it would be good to extract it.

packages/puppeteer-core/src/bidi/Browser.ts
Lightning00Blade: Can you check if the new version of mapper fixes this?

packages/puppeteer-core/src/bidi/BrowserConnector.ts
```diff
@@ -39,11 +39,10 @@ export async function _connectToBiDiBrowser(
     options
   );
   const BiDi = await import(/* webpackIgnore: true */ './bidi.js');
-  const bidiBrowser = await BiDi.BidiBrowser.create({
-    connection: bidiConnection,
+  const bidiBrowser = await BiDi.BidiBrowser.from(bidiConnection, {
     closeCallback,
     process: undefined,
-    defaultViewport: defaultViewport,
+    defaultViewport: defaultViewport ?? undefined,
```
Lightning00Blade: When is this needed? We set the default always?

packages/puppeteer-core/src/bidi/CdpSession.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (sessionId !== undefined) {
+      this.#sessionId.resolve(sessionId);
+    }
+
+    void this.#initialize();
+  }
+
+  get #connection(): BidiConnection {
+    return this.#frame.browserContext.browser().session
+      .connection as BidiConnection;
```
Lightning00Blade: Why the type casting?

packages/puppeteer-core/src/bidi/CdpSession.ts
```diff
@@ -0,0 +1,103 @@
...
+    BidiCdpSession.sessions.set(this.#sessionId.value() as string, this);
+  }
+
+  override connection(): CdpConnection | undefined {
+    return undefined;
+  }
+
+  override async send<T extends keyof ProtocolMapping.Commands>(
+    method: T,
+    ...paramArgs: ProtocolMapping.Commands[T]['paramsType']
```
Lightning00Blade: Non blocker: Please add `TODO:` we should support the same behavior as CDP implementation to have timeout here.

packages/puppeteer-core/src/bidi/Connection.ts
```diff
@@ -190,7 +189,7 @@ export class BidiConnection extends EventEmitter<BidiEvents> {
         method,
         params,
       } as Bidi.Command);
-      debugProtocolSend(stringifiedMessage);
+      debugProtocolSend(stringifiedMessage, new Error().stack);
```
Lightning00Blade: Why this change?

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -3,263 +3,701 @@
...
+      default: new BidiFrameRealm(
+        this.#browsingContext.defaultRealm,
         this,
-        context.createRealmForSandbox(),
-        timeoutSettings
+        this.#timeoutSettings
+      ),
+      internal: new BidiFrameRealm(
+        this.#browsingContext.createWindowRealm(
+          `__puppeteer_internal_${Math.ceil(Math.random() * 10000)}`
```
Lightning00Blade: Why is this random, I think it adds low benefit while making it harder to debug?

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -269,6 +707,65 @@ export class BidiFrame extends Frame {
       );
     }
 
-    return super.waitForSelector(selector, options);
+    return await super.waitForSelector(selector, options);
+  }
+}
+
+function isConsoleLogEntry(
```
Lightning00Blade: Non blocker: All bellow I think we should extract them in `bidi/utils.ts`.

packages/puppeteer-core/src/bidi/HTTPRequest.ts
```diff
@@ -5,106 +5,126 @@
...
-
-    for (const header of event.request.headers) {
-      // TODO: How to handle Binary Headers
-      // https://w3c.github.io/webdriver-bidi/#type-network-Header
-      if (header.value.type === 'string') {
-        this.#headers[header.name.toLowerCase()] = header.value.value;
-      }
-    }
+  override get client(): CDPSession {
+    throw new Error('Method not implemented.');
```
Lightning00Blade: Wrong error type.

packages/puppeteer-core/src/node/PipeTransport.ts
```diff
@@ -16,7 +16,7 @@ export class PipeTransport implements ConnectionTransport {
   #pipeWrite: NodeJS.WritableStream;
   #subscriptions = new DisposableStack();
 
-  #isClosed = false;
+  disposed = false;
```
Lightning00Blade: Unrelated change, please extract.

packages/puppeteer-core/src/node/PipeTransport.ts
```diff
@@ -28,35 +28,33 @@ export class PipeTransport implements ConnectionTransport {
   ) {
     this.#pipeWrite = pipeWrite;
     this.#subscriptions.use(
-      new EventSubscription(pipeRead, 'data', (buffer: Buffer) => {
-        return this.#dispatch(buffer);
+      new EventSubscription(pipeRead as any, 'data', (buffer: unknown) => {
```
Lightning00Blade: Please also create a light type wrapper.

packages/puppeteer-core/src/common/EventEmitter.ts
```diff
@@ -202,15 +202,18 @@ export class EventEmitter<Events extends Record<EventType, unknown>>
  * @internal
  */
 export class EventSubscription<
-  Target extends CommonEventEmitter<Record<Type, Event>>,
-  Type extends EventType = EventType,
-  Event = unknown,
+  Events extends Record<EventType, unknown>,
```
Lightning00Blade: Same here change unrelated to bidi/core

packages/puppeteer-core/src/cdp/HTTPRequest.ts
```diff
@@ -28,8 +28,9 @@ import type {CdpHTTPResponse} from './HTTPResponse.js';
  * @internal
  */
 export class CdpHTTPRequest extends HTTPRequest {
-  declare _redirectChain: CdpHTTPRequest[];
-  declare _response: CdpHTTPResponse | null;
+  _redirectChain: CdpHTTPRequest[] = [];
```
Lightning00Blade: Same here.

packages/puppeteer-core/src/bidi/core/UserContext.ts
```diff
@@ -0,0 +1,113 @@
...
+  ): Promise<BrowsingContext> {
+    const {
+      result: {context: contextId},
+    } = await this.#connection.send('browsingContext.create', {
+      type,
+      ...options,
+      referenceContext: options.referenceContext?.id,
+    });
+
+    const browsingContext = this.#browsingContexts.get(contextId)!;
```
Lightning00Blade: Should not be needed as we assert after.
```suggestion
    const browsingContext = this.#browsingContexts.get(contextId);
```

packages/puppeteer-core/src/bidi/core/Request.ts
```diff
@@ -0,0 +1,112 @@
...
+    context: BrowsingContext,
+    event: Bidi.Network.BeforeRequestSentParameters
+  ) {
+    super();
+    this.#context = context;
+    this.#event = event;
+
+    const connection = this.#context.userContext.browser.session.connection;
+    this.#disposables.use(
+      new EventSubscription(connection, 'network.responseCompleted', event => {
```
Lightning00Blade: Nit order them as in the BiDi spec.

packages/puppeteer-core/src/bidi/core/Request.ts
```diff
@@ -0,0 +1,112 @@
...
+        if (event.context !== this.#context.id) {
```
Lightning00Blade: Extract into helper we repeat it several times

packages/puppeteer-core/src/bidi/core/Realm.ts
```diff
@@ -0,0 +1,256 @@
...
+      awaitPromise,
+      target: this.target,
+      ...options,
+    });
+    return result;
+  }
+
+  async evaluate(
+    expression: string,
+    awaitPromise: boolean,
```
Lightning00Blade, line 86: Can you point me to a the code where we use `awaitPromise: false`?

packages/puppeteer-core/src/bidi/core/Realm.ts
```diff
@@ -0,0 +1,256 @@
...
+    this.connection.on('script.realmDestroyed', info => {
+      if (info.realm === this.id) {
+        this.emit('destroyed', undefined);
+      }
+    });
+  }
+
+  protected abstract get connection(): Connection;
+
+  protected get target(): Bidi.Script.Target {
```
Lightning00Blade, line 59: Use The Realm Target type

packages/puppeteer-core/src/bidi/JSHandle.ts
```diff
@@ -56,16 +55,11 @@ export class BidiJSHandle<T = unknown> extends JSHandle<T> {
...
-        this.context(),
-        this.#remoteValue as Bidi.Script.RemoteReference
-      );
-    }
+    await this.realm.destroyHandles([this]);
   }
 
-  get isPrimitiveValue(): boolean {
-    switch (this.#remoteValue.type) {
+  get primitive(): boolean {
```
Lightning00Blade: Same here changes are not directly related to `bidi/core` please extract.

packages/puppeteer-core/src/bidi/HTTPResponse.ts
```diff
@@ -7,86 +7,85 @@ import type * as Bidi from 'chromium-bidi/lib/cjs/protocol/protocol.js';
...
-  constructor(
-    request: BidiHTTPRequest,
-    {response}: Bidi.Network.ResponseCompletedParameters
+
+  private constructor(
+    data: Bidi.Network.ResponseData,
+    request: BidiHTTPRequest
   ) {
     super();
+    this.#data = data;
```
Lightning00Blade: Let use the same pattern you use in Request:
```suggestion
    this.#response = response;
```

packages/puppeteer-core/src/bidi/ElementHandle.ts
```diff
@@ -74,9 +73,16 @@ export class BidiElementHandle<
       }
       return;
     })) as BidiJSHandle;
-    const value = handle.remoteValue();
+    const value = handle.remoteValue;
     if (value.type === 'window') {
-      return this.frame.page().frame(value.value.context);
+      return (
```
Lightning00Blade: Why, the `frame()` still exitst on Page.

packages/puppeteer-core/src/bidi/Browser.ts
Lightning00Blade: Leave a comment from offline conversation:
If you do `await <Any>`, the execution of the following code is put at the end of the `microtask` queue.
so this is not needed. Also this puts it on the end of the `macrotask` queue.

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -0,0 +1,150 @@
+import type * as Bidi from 'chromium-bidi/lib/cjs/protocol/protocol.js';
```
OrKoN, line 7: Needs license headers (we have new ones)

packages/puppeteer-core/src/bidi/core/Session.ts
```diff
@@ -0,0 +1,137 @@
+import type * as Bidi from 'chromium-bidi/lib/cjs/protocol/protocol.js';
+
+import {EventEmitter} from '../../common/EventEmitter.js';
+import {debugError} from '../../common/util.js';
+import {throwIfDisposed} from '../../util/decorators.js';
+
+import {Browser} from './Browser.js';
+import type {Connection} from './Connection.js';
+
+export class Session extends EventEmitter<{
```
OrKoN: Could you please add `@internal` annotations to all classes?

packages/puppeteer-core/src/bidi/core/UserContext.ts
```diff
@@ -0,0 +1,113 @@
...
+import {BrowsingContext} from './BrowsingContext.js';
+
+export type CreateBrowsingContextOptions = Omit<
+  Bidi.BrowsingContext.CreateParameters,
+  'type' | 'referenceContext'
+> & {
+  referenceContext?: BrowsingContext;
+};
+
+export class UserContext extends EventEmitter<{
```
OrKoN, line 28: UserContext is not part of the spec yet

packages/puppeteer-core/src/bidi/core/Session.ts
```diff
@@ -0,0 +1,137 @@
...
+    'cdp.Page.screencastFrame',
+  ];
+
+  static async from(
+    connection: Connection,
+    capabilities: Bidi.Session.CapabilitiesRequest
+  ): Promise<Session> {
+    // Wait until the connection is ready.
+    while (true) {
+      const {result: ready} = await connection.send('session.status', {});
```
OrKoN: This looks like an infinite loop if the limit of connections is exceeded. Can we use rxjs here and throw an error?

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -0,0 +1,150 @@
...
+    // keep-sorted end
+
+    this.#userContexts.set('', UserContext.create(this, ''));
+  }
+
+  async #initialize() {
+    // ///////////////////////
+    // Connection listeners //
+    // ///////////////////////
+    const connection = this.#connection;
```
OrKoN: I am not seeing `#connection` being defined anywhere?

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -0,0 +1,150 @@
...
+    connection.on('script.realmCreated', info => {
```
OrKoN: I think it would be better if we set up event listeners and send commands through the session rather than exposing the connection to every class?

packages/puppeteer-core/src/bidi/core/Session.ts
```diff
@@ -0,0 +1,137 @@
...
+  get capabilities(): Bidi.Session.NewResult['capabilities'] {
+    return this.#info.capabilities;
+  }
+
+  @throwIfDisposed((session: Session) => {
+    // SAFETY: By definition of `disposed`, `#reason` is defined.
+    return session.#reason!;
+  })
+  async end(): Promise<void> {
+    await this.connection.send('session.end', {});
```
OrKoN: Should we use dispose symbols?

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -0,0 +1,150 @@
...
+    this.#reason = `Browser has already closed.`;
+    this.emit('closed', {reason: this.#reason});
+    this.removeAllListeners();
+  }
+
+  @throwIfDisposed((browser: Browser) => {
+    // SAFETY: By definition of `disposed`, `#reason` is defined.
+    return browser.#reason!;
+  })
+  async addPreloadScript(
```
OrKoN, line 154: how did you decide that the preload script should be on the browser?

packages/puppeteer-core/src/bidi/core/UserContext.ts
jrandolf: I've implemented the temporary spec here. Do you think we shouldn't add this yet? I'm quite certain that it will be added and this class simplifies our implementation.

packages/puppeteer-core/src/bidi/core/Session.ts
jrandolf: If the limit of connections is exceeded, then this will throw as expected. Are you saying this method will succeed in this case?

packages/puppeteer-core/src/bidi/core/Browser.ts
jrandolf: It's defined on line 93.

packages/puppeteer-core/src/bidi/core/Browser.ts
jrandolf: I opted for allowing classes to handle their own events because this provides greater separation. If the session becomes a control plane, then we have to wire the children with the session which typically requires exposing internal state within the children to the session. I.e. this approach keeps the internal state truly internal.

packages/puppeteer-core/src/bidi/core/Browser.ts
jrandolf: `contexts` is allowed to be optional, so `addPreloadScript` can be triggered on the all contexts. The highest node that contains all contexts is the browser (IIUC).

packages/puppeteer-core/src/bidi/core/Realm.ts
jrandolf: We just want to keep it here because that's a required argument in `script.evaluate`. Required args are parametrized while optional ones are put into an object.

packages/puppeteer-core/src/bidi/core/Session.ts
jrandolf: We can, but for now, I left them out. We don't really do something like `using browser = puppeteer.launch()` for example in the first place. We could add this in the future.

packages/puppeteer-core/src/bidi/core/Realm.ts
jrandolf: The `WindowRealm` uses `script.ContextTarget` instead of `script.RealmTarget` for its `Target` type.

packages/puppeteer-core/src/bidi/core/Session.ts
OrKoN: By looking at https://w3c.github.io/webdriver/#dfn-readiness-state it might never be ready resulting an infinite loop.

packages/puppeteer-core/src/bidi/core/Browser.ts
OrKoN: I mean that the session can basically `implements Connection` especially given that it is responsible for event subcription

jrandolf: I see, so basically add a `send` on the `session`. I've thought about this but thought it's better to separate the implementation because `Session` should precisely match the concept of session in WebDriver BiDi. In particular, FWIS, we don't send the session ID with each command which would in theory make all commands session dependent.

Perhaps though this is the actual intention. If so, then we can do what your saying, but from the spec, I don't see the session ID being passed anywhere.

packages/puppeteer-core/src/bidi/core/Session.ts
jrandolf: So should we implement a retry? Perhaps we can just retry 5-6 times and implement backoffs.

packages/puppeteer-core/src/bidi/core/Browser.ts
OrKoN: sessionId is not passed anywhere because there is 1:1 relationship between the session and the connection. In the future, we might support multiple sessions and in that case you always needs to make sure that the connection for the right session object is used (if code deals with multiple sessions/connection). But the main reason for me to do so is that the session and not the connection decides which events to subscribe to. The spec argument is valid but I think it is weak because this is the client code and, for example, the spec session does not set up any event subscriptions whereas the current version does (so technically you need to move out the subscription logic out of the session and leave it to the puppeteer layer). If we move events to session, we could also error if you create an event listener without a subscription.

packages/puppeteer-core/src/bidi/core/Session.ts
jrandolf: I've implemented backoffs.

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -0,0 +1,162 @@
...
+      this.emit('disconnected', {reason});
+      this.removeAllListeners();
+    });
+
+    // //////////////////////////////
+    // Asynchronous initialization //
+    // //////////////////////////////
+    const {
+      result: {contexts},
+    } = await connection.send('browsingContext.getTree', {});
```
OrKoN: I think this should belong to the UserContext?

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    // Simulating events so contexts are created naturally.
+    for (const context of contexts) {
+      connection.emit('browsingContext.contextCreated', context);
```
OrKoN: Isn't there a race condition here? While 'browsingContext.getTree' is getting sent some contexts could be created and be emitted both via the real message and via this emulated message

packages/puppeteer-core/src/bidi/core/Browser.ts
jrandolf: Hmm, you're right about the subscriptions. I forgot to remove that code because we are supposed to subscribe in the puppeteer layer.

I think the multi-session argument makes sense. Let me implement it that way then.

packages/puppeteer-core/src/bidi/core/Browser.ts
jrandolf: I think this depends on whether we let `getTree` include a user context ID or not. As of the current draft, this method should get all browsingContexts, independent of the user context, so it should be called on the browser.

packages/puppeteer-core/src/bidi/core/Browser.ts
jrandolf: Nice catch.

packages/puppeteer-core/src/bidi/core/Session.ts
```diff
@@ -0,0 +1,165 @@
...
+      await new Promise(resolve => {
+        return setTimeout(resolve, (1 << i) * 100);
+      });
+    }
+    if (!status.ready) {
+      throw new Error(status.message);
+    }
+
+    let result;
+    // TODO: await until the session is established.
```
OrKoN: is the TODO still relevant? looks like it's redundant

packages/puppeteer-core/src/bidi/core/Request.ts
```diff
@@ -0,0 +1,116 @@
...
+    const session = this.#session;
+    session.on('network.beforeRequestSent', event => {
+      if (event.context !== this.id) {
+        return;
+      }
+      if (event.request.request !== this.id) {
+        return;
+      }
+      if (this.#redirect) {
+        // XXX: Separate descendant redirect events?.
```
OrKoN: Could you add more details what it means?

packages/puppeteer-core/src/bidi/core/Realm.ts
```diff
@@ -0,0 +1,271 @@
...
+  }
+
+  override initialize(): void {
+    super.initialize();
+
+    // ///////////////////////
+    // Session listeners //
+    // ///////////////////////
+    this.session.on('script.realmCreated', info => {
+      if (info.type === 'dedicated-worker') {
```
OrKoN, line 262: needs filtering out dedicated workers from other realms?

packages/puppeteer-core/src/bidi/core/Browser.ts
OrKoN: so you remove the event subscriptions from session? I think it would be okay to keep them because the library decides which events to listen on on its own so subscription should be also handled in the session, right?

packages/puppeteer-core/src/bidi/core/Request.ts
jrandolf: This was used to help with calculating redirect chaining, but I think we can remove this since we can just attach listeners onto each subsequent redirect to calculate the chain in Puppeteer.

OrKoN (approved):
LGTM in general. I still have the question about where subscription calls should live.

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -0,0 +1,179 @@
...
+  Bidi.Script.AddPreloadScriptParameters,
+  'functionDeclaration' | 'contexts'
+> & {
+  contexts?: [BrowsingContext, ...BrowsingContext[]];
+};
+
+/**
+ * @internal
+ */
+export default class Browser extends EventEmitter<{
```
Lightning00Blade: Default export is bad practice and should be done only with external code - https://google.github.io/styleguide/tsguide.html#imports

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -1,3 +1,9 @@
+/**
+ * @license
```
Lightning00Blade, line 2: Can you create an issue to create an EsLint rule to check/add this.

packages/puppeteer-core/src/bidi/core/Browser.ts
Lightning00Blade: And https://google.github.io/styleguide/tsguide.html#exports

packages/puppeteer-core/src/bidi/core/Browser.ts
jrandolf: It'd be faster if you do this. I'm not sure what you mean. What is 'this'

Lightning00Blade: Opened #11701.

PR #11651, opened 2024-01-08
refactor: move PDF options to utilities

jrandolf-google:


Lightning00Blade (approved):
LGTM + how about we move them to `PDFOptions` and rename it `PdfUtils`

jrandolf:
> LGTM + how about we move them to `PDFOptions` and rename it `PdfUtils`

Not a fan of "Utils" classes since `* as PdfUtils` is always possible.

PR #11660, opened 2024-01-09
fix: expose function for Firefox BiDi

### LINKED ISSUE #11653: [Bug]: Exposed function promise stays pending on Firefox

```TypeScript

// Sample function
const customFunction = () => {
  return "Hello World!";
}

const start = async () => {
    headless: false,
    protocol: 'webDriverBiDi',
    product: 'firefox',
  });

  await page.exposeFunction('customFunction', customFunction);

    console.log("Starting evaluation") // This prints always
    const result = await window.customFunction();
    console.log("Finished evaluation") // This prints only in Chrome
    console.log(result);
  });
}

start();
```

### Error string

invalid argument A context and a realm reference are mutually exclusive

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

I'm trying to use external function inside Puppeteer browser context with `page.exposeFunction`. It works perfectly when using Chrome, not so much in Firefox.

### Expectation

The sample code should output three lines to *browser* console:
```
Starting evaluation
Hello world!
Finished evaluation
```

### Reality

Exposed function is not evaluated in Firefox, instead the following error is thrown (if run with `env DEBUG="puppeteer:*"`):
```
ProtocolError: Protocol error (script.callFunction): invalid argument A context and a realm reference are mutually exclusive RemoteError@chrome://remote/content/shared/RemoteError.sys.mjs:8:8
  WebDriverError@chrome://remote/content/shared/webdriver/Errors.sys.mjs:191:5
  InvalidArgumentError@chrome://remote/content/shared/webdriver/Errors.sys.mjs:382:5
  #assertTarget@chrome://remote/content/webdriver-bidi/modules/root/script.sys.mjs:657:13
  callFunction@chrome://remote/content/webdriver-bidi/modules/root/script.sys.mjs:308:63
  handleCommand@chrome://remote/content/shared/messagehandler/MessageHandler.sys.mjs:255:33
  execute@chrome://remote/content/shared/webdriver/Session.sys.mjs:274:32
  onPacket@chrome://remote/content/webdriver-bidi/WebDriverBiDiConnection.sys.mjs:211:37
  onMessage@chrome://remote/content/server/WebSocketTransport.sys.mjs:127:18
  handleEvent@chrome://remote/content/server/WebSocketTransport.sys.mjs:109:14
```

Additionally, if I try to run the exposed function directly on Firefox console, the result is always a promise that stays pending.

_No response_

### Puppeteer version

21.7.0

### Node version

21.5.0

### Package manager

npm

### Package manager version

10.2.5

### Operating system

Linux

Lightning00Blade:
Closes #11653.
Firefox needs only realm or context to be send both will result in Error. Filed an issue for the spec
https://github.com/w3c/webdriver-bidi/issues/635

### DIFF
--- a/packages/puppeteer-core/src/bidi/ExposedFunction.ts
+++ b/packages/puppeteer-core/src/bidi/ExposedFunction.ts
@@ -66,7 +66,6 @@ export class ExposeableFunction<Args extends unknown[], Ret> {
   async expose(): Promise<void> {
     const connection = this.#connection;
     const channelArguments = this.#channelArguments;
-    const {name} = this;
 
     // TODO(jrandolf): Implement cleanup with removePreloadScript.
     connection.on(
@@ -103,7 +102,7 @@ export class ExposeableFunction<Args extends unknown[], Ret> {
             },
           });
         },
-        {name: JSON.stringify(name)}
+        {name: JSON.stringify(this.name)}
       )
     );
 
@@ -148,7 +147,9 @@ export class ExposeableFunction<Args extends unknown[], Ret> {
           BidiSerializer.serializeRemoteValue(result),
         ],
         awaitPromise: false,
-        target: params.source,
+        target: {
+          realm: params.source.realm,
+        },
       });
     } catch (error) {
       try {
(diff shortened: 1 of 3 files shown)


whimboo:
We probably should not change the behavior but having a test for this situation would still be good.

Lightning00Blade:
I prefer to be explicit in the situation rather than relay on something down the line magically removing it. 
Test was added.

jrandolf-google (approved):

PR #11665, opened 2024-01-10
chore: sort class members

jrandolf-google:
No more bikeshedding over where to declare members. This rule sorts everything for us!


.eslintrc.js
```diff
@@ -149,6 +155,145 @@ module.exports = {
     'rulesdir/prettier-comments': 'error',
     // Enforces consistent file extension
     'rulesdir/extensions': 'error',
+    // Enforces class members are reasonably sorted.
+    'sort-class-members/sort-class-members': [
+      2,
```
Lightning00Blade, line 160:
```suggestion
      'error',
```

PR #11678, opened 2024-01-15
chore: remove redundant type-check

kblok:
Both the `CDPSession` and the `Connection` extend `EventEmitter<CDPSessionEvents>`. There is no need for a type check here.

And some typos :)


OrKoN (approved):
Thanks!

PR #11679, opened 2024-01-15
chore: remove updated but unused variable

kblok:
I found that we are updating the `ignoredTargets` variable but never using its content. I don't know whether this PR is valid or if we need to use this list somewhere else.


OrKoN (approved):
Thanks!

PR #11680, opened 2024-01-15
fix(docker): fix the chrome deb url to https

tailgo:
The Chrome deb url in dockerfile should be `https`. And This fixes the docker/readme so that the `docker run` command demo will be the same as the 'ci.yml'.

Build related change.

Yes, I did.

No need to update the documentation.

**Summary**

I had been updated the dockerfile that upgrade the nodejs version from v14 to v18, at that time I use the http version url to download the chrome deb. I think it shoule use the https version, so I try to fixed it now.

No.

**Other information**


OrKoN (approved):
Thanks!

PR #11688, opened 2024-01-15
chore: remove extra event listener

kblok:
`onAttachedToTarget` is already a `Ready` listener ([see](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/cdp/Page.ts#L323)).I don't see  a reason to add the same listener again.

### DIFF
--- a/packages/puppeteer-core/src/cdp/Page.ts
+++ b/packages/puppeteer-core/src/cdp/Page.ts
@@ -352,7 +352,6 @@ export class CdpPage extends Page {
       this.#workers.set(session.id(), worker);
       this.emit(PageEvent.WorkerCreated, worker);
     }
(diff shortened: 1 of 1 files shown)


packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -352,7 +352,6 @@ export class CdpPage extends Page {
       this.#workers.set(session.id(), worker);
       this.emit(PageEvent.WorkerCreated, worker);
     }
-    session.on(CDPSessionEvent.Ready, this.#onAttachedToTarget);
```
OrKoN, line 355: I believe it's a subframe session here that might need to be configured recursively

kblok: ahhh because the session will notify that another session is ready. Gotcha!

PR #11702, opened 2024-01-18
refactor: adopt `core/Browser` on `BidiBrowser`

jrandolf-google:


packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -70,47 +72,28 @@ export class BidiBrowser extends Browser {
...
+    await session.subscribe(
+      session.capabilities.browserName.toLocaleLowerCase().includes('firefox')
+        ? BidiBrowser.subscribeModules
+        : [...BidiBrowser.subscribeModules, ...BidiBrowser.subscribeCdpEvents]
+    );
 
+    const browser = new BidiBrowser(session.browser, opts);
     await browser.#getTree();
-
+    browser.#initialize();
```
OrKoN, line 89: I think this should happen before browser.#getTree() I believe

jrandolf: You're right.

test/src/browser.spec.ts
```diff
@@ -64,7 +64,7 @@ describe('Browser specs', function () {
     });
   });
 
-  describe('Browser.isConnected', () => {
+  describe('Browser.connected', () => {
```
OrKoN: Let's revert unrelated changes. Also https://pptr.dev/api/puppeteer.browser.isconnected there is some issue with the documentation for Browser.connected

Lightning00Blade: FYI: The issue is sort of fixed on `next`. It's related to properties not having pages.

OrKoN (changes requested):
tests seem to be failing for Firefox

packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -229,29 +217,33 @@ export class BidiBrowser extends Browser {
...
 
-    // `browser.close` can close connection before the response is received.
-    await this.#connection.send('browser.close', {}).catch(debugError);
-    await this.#closeCallback?.call(null);
-    this.#connection.dispose();
+    try {
+      await this.#browserCore.close();
+      await this.#closeCallback?.call(null);
+    } catch (error) {
+      // Fail silently.
```
OrKoN, line 240: I think we should not fail silently for closeCallback and also we should run it even if #browserCore.close() failed.  Perhaps move it to finally?

packages/puppeteer-core/src/bidi/core/Session.ts
```diff
@@ -29,20 +28,21 @@ export class Session
...
-      await new Promise(resolve => {
-        return setTimeout(resolve, (1 << i) * 100);
-      });
-    }
-    if (!status.ready) {
-      throw new Error(status.message);
-    }
+    // TODO: Once Chrome supports session.status properly, uncomment this block.
+    // let status = {message: '', ready: false};
+    // for (let i = 0; i < MAX_RETRIES; ++i) {
```
OrKoN, line 36: let's only run this if the browser is not Chrome instead of leaving it commented out?

packages/puppeteer-core/src/bidi/Browser.ts
jrandolf: `closeCallback` calls `close` from looking at the location of the callback, so it's fine to fail silently here. We need to rework that logic at some point.

packages/puppeteer-core/src/bidi/core/Session.ts
jrandolf: Looks like this logic is placed in `BrowserConnector` already. We would need to remove the logic there and replace it with this one once we support pure bidi.

jrandolf: I've updated the comment with this information.

packages/puppeteer-core/src/bidi/Browser.ts
OrKoN: It should force close the browser process if the protocol methods failed. The callback is in ProductLauncher. Agree on refactor, not sure if this change will affect any behavior in practice, probably not.

OrKoN: The tests seem to take quite longer than usual https://github.com/puppeteer/puppeteer/actions/runs/7583926821/job/20656609557?pr=11702

OrKoN (approved):

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -136,15 +134,28 @@ export class Browser extends EventEmitter<{
...
     return browser.#reason!;
   })
   async close(): Promise<void> {
-    await this.#session.send('browser.close', {});
-    this.#reason = `Browser has already closed.`;
-    this.emit('closed', {reason: this.#reason});
-    this.removeAllListeners();
+    try {
+      await this.#session.send('browser.close', {});
+    } finally {
```
Lightning00Blade, line 156: Why is this in finally block? The behavior changed now, previously the browser would not be disposed if the command failed.

packages/puppeteer-core/src/bidi/core/Session.ts
```diff
@@ -156,9 +166,10 @@ export class Session
     return session.#reason!;
   })
   async end(): Promise<void> {
-    await this.send('session.end', {});
-    this.#reason = `Session (${this.id}) has already ended.`;
-    this.emit('ended', {reason: this.#reason});
-    this.removeAllListeners();
+    try {
+      await this.send('session.end', {});
+    } finally {
```
Lightning00Blade, line 171: Same here.

packages/puppeteer-core/src/bidi/core/Browser.ts
OrKoN: This has not changed, see packages/puppeteer-core/src/bidi/Browser.ts::close

packages/puppeteer-core/src/bidi/core/Session.ts
OrKoN: session.end is not supported in mapper.

PR #11707, opened 2024-01-19
chore: flip test

Lightning00Blade:
BiDi logic is more robust and can handle the `circular object`. While our Deserialize does not provide a way to express it in the output result it's much better then getting only `undefined`. 

CC @whimboo


whimboo:
Sounds great. Thanks!

OrKoN (approved):

PR #11714, opened 2024-01-22
refactor: adopt `core/UserContext` on `BidiBrowserContext`

jrandolf-google:


packages/puppeteer-core/src/bidi/core/UserContext.ts
```diff
@@ -115,11 +123,15 @@ export class UserContext extends EventEmitter<{
...
-    const promises = [];
-    for (const browsingContext of this.#browsingContexts.values()) {
-      promises.push(browsingContext.close());
-    }
-    await Promise.all(promises);
+  async remove(): Promise<void> {
+    // TODO: implement incognito context https://github.com/w3c/webdriver-bidi/issues/289.
+    // TODO: Uncomment once we have adopted the new browsing contexts.
+    // const promises = [];
+    // for (const browsingContext of this.#browsingContexts.values()) {
```
OrKoN: this should be handled by bidi, I think we do not need to close explicitly but we might need to wait till it is destroyed.

packages/puppeteer-core/src/bidi/BrowserContext.ts
```diff
@@ -113,7 +126,7 @@ export class BidiBrowserContext extends BrowserContext {
   }
 
   override isIncognito(): boolean {
-    return !this.#isDefault;
+    return this.#userContext.id !== '';
```
OrKoN: default one will have the id `default`

OrKoN (approved):

packages/puppeteer-core/src/bidi/core/UserContext.ts
jrandolf: That's good then. We can just fix the logic here since Puppeteer does this for us and remove the logic in Puppeteer once we implement this in Bidi

jrandolf: Actually, we can just remove this and call removeUserContext I guess. I assume that command doesn't resolve until all browsing contexts are deleted anyway.

PR #11716, opened 2024-01-23
build: ESLint rule to check/add license headers

### LINKED ISSUE #11701: [Feature]: Create an EsLint rule to check/add Licence headers
Currently when new files are added we need to manually review, verify and add the correct licence header.
Having a rule to automatically do this will save some development time/friction. 
Example of the header that should exits on each file (.ts,.mjs,.cjs)
```js
/**
 * @license
 * Copyright 2018 Google Inc.
 */
 ```
If the header does not exist a new one should be added with the current year included.

Note: there are some files we would like to ignore as they are adapted from other open source projects and should keep their respective headers.

### PR DESCRIPTION
**Pull request by richestrinh**

An EsLint rule to check/add Licence headers

I wrote the test for this rule, but I was unable to locate the tests for the other rules (extension.js and prettier-comments.js).

Resolves #11701

This rule checks if a license header exists and fixes the 2 cases:
1. No license.
FIX: Adds license header.
2. Header comment that is not the license. 
FIX: Adds licensing as header.

**Other information**
There is another case where the license comment may already exist in the file, but not as a header. 
Shall I cover this case as well, and implement a fix to move that comment so it is a header?

Please let me know if anything else is needed! :)

### CODE REVIEW

**Comment by OrKoN** [comment]
cc @Lightning00Blade

File: tools/eslint/src/check-license.js [hunk 1]
```diff
@@ -0,0 +1,69 @@
+/**
+ * @license
+ * Copyright 2024 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+const licenseHeader = `/**
+ * @license
+ * Copyright 2024 Google Inc.
```
Lightning00Blade: Let's use the current year and not hard-code 2024

.eslintrc.js
```diff
@@ -149,6 +149,8 @@ module.exports = {
     'rulesdir/prettier-comments': 'error',
     // Enforces consistent file extension
     'rulesdir/extensions': 'error',
+    // Enforces license headers on files (.ts,.mjs,.cjs)
```
Lightning00Blade: The files affected are controlled by EsLint so we can skip them.
```suggestion
    // Enforces license headers on files
```

tools/eslint/src/check-license.js
```diff
@@ -0,0 +1,69 @@
+/**
```
Lightning00Blade: Let's use TypeScript like `extensions.ts` or `use-using.ts`, to get better typing.

tools/eslint/src/check-license.js
```diff
@@ -0,0 +1,69 @@
...
+          });
+        }
+        // Header exists, but is not license.
+        if (header && !header.value.includes('@license')) {
+          context.report({
+            node: node,
+            messageId: 'licenseRule',
+            fix(fixer) {
+              return fixer.insertTextBeforeRange(
+                [header.range[0], header.range[1]],
```
Lightning00Blade: We should be able to pass in `header.range` directly as it's a tuple of numbers `range?: [number, number];` 
But refer to my previous comment.

tools/eslint/src/check-license.js
```diff
@@ -0,0 +1,69 @@
...
+
+    return {
+      Program(node) {
+        if (header?.value.includes('@license')) {
+          return;
+        }
+
+        // Add header license
+        if (!header) {
+          context.report({
```
Lightning00Blade: I think we can use this for both case of `No Header` and `Header without @license`, as most likely we will still update/review it manually.

tools/eslint/src/check-license.js
```diff
@@ -0,0 +1,69 @@
...
+ * Copyright 2024 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */\n\n`;
+
+/**
+ * @type import("eslint").Rule.RuleModule
+ */
+
+const enforceLicenseRule = {
+  // module.exports = {
```
Lightning00Blade:
```suggestion
```

tools/eslint/src/check-license.js
```diff
@@ -0,0 +1,69 @@
...
+    fixable: 'code',
+    schema: [],
+    messages: {
+      licenseRule: 'Add license header.',
+    },
+  },
+  create(context) {
+    const sourceCode = context.sourceCode;
+    const comments = sourceCode.getAllComments();
+    const header = comments[0];
```
Lightning00Blade: Currently we take the first comment, which is OK, 
but we should check if the comment is of type `block` and if it's at the start of the file.
If one of the two is not `true` we should treat is as no header comment.

tools/eslint/src/check-license.js
```diff
@@ -0,0 +1,69 @@
...
+
+    return {
+      Program(node) {
+        if (header?.value.includes('@license')) {
```
Lightning00Blade: We should check to include `@license`, `SPDX-License-Identifier: Apache-2.0` and
this RegEx `Copyright ([0-9]{4}) Google Inc.`. as we want a valid header as well.

Lightning00Blade (commented):
@richestrinh Thank you for your contribution, overall looks good.
Please recreate the rule using TypeScript. (`prettier-comments.js` is old rule that we never migrated).
Also we currently have not set-up tests for EsLint rules so no need to provide any.
Everything else is in comment.

tools/eslint/src/check-license.js
richestrinh: I merged the cases together. 
The two cases had conflicting fixes, and if I understood your comment correct I have omitted the fixes as well.

tools/eslint/src/check-license.ts
```diff
@@ -0,0 +1,68 @@
...
+        return false;
+      }
+    }
+
+    return {
+      Program(node) {
+        if (
+          header?.value.includes('@license') &&
+          header?.value.includes('SPDX-License-Identifier: Apache-2.0') &&
+          copyrightPattern.test(header?.value)
```
Lightning00Blade:
```suggestion
          header &&
          header.value.includes('@license') &&
          header.value.includes('SPDX-License-Identifier: Apache-2.0') &&
          copyrightPattern.test(header.value)
```

tools/eslint/src/check-license.js
Lightning00Blade: Sorry about this, I though we can have the same logic for both, it seems like EsLint does not like it when you do that so we should split them as you have suggested.

tools/eslint/src/check-license.ts
```diff
@@ -0,0 +1,68 @@
...
+  defaultOptions: [],
+  create(context) {
+    const sourceCode = context.sourceCode;
+    const comments = sourceCode.getAllComments();
+    const header =
+      comments[0]?.type === 'Block' && isHeaderComment(comments[0])
+        ? comments[0]
+        : null;
+
+    function isHeaderComment(comment: any) {
```
Lightning00Blade:
```suggestion
    function isHeaderComment(comment: TSESTree.Comment) {
```

tools/eslint/src/check-license.ts
```diff
@@ -0,0 +1,68 @@
...
+      if (comment?.range[0] >= 0 && comment?.range[1] <= 88) {
```
Lightning00Blade:
```suggestion
      if (comment && comment.range[0] >= 0 && comment.range[1] <= 88) {
```

tools/eslint/src/check-license.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * @license
+ * Copyright 2024 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {ESLintUtils} from '@typescript-eslint/utils';
+
+const createRule = ESLintUtils.RuleCreator(name => {
+  return `https://github.com/puppeteer/puppeteer/tree/main/tools/eslint/${name}.js`;
```
Lightning00Blade: I think other places are wrong.
```suggestion
  return `https://github.com/puppeteer/puppeteer/tree/main/tools/eslint/${name}.ts`;
```

tools/eslint/src/check-license.ts
```diff
@@ -0,0 +1,68 @@
...
+});
+
+const copyrightPattern = /Copyright ([0-9]{4}) Google Inc\./;
+
+const enforceLicenseRule = createRule<[], 'licenseRule'>({
+  name: 'license',
```
Lightning00Blade: Name should match filename.
```suggestion
  name: 'check-license',
```

Lightning00Blade (commented):
Please readd the fixer back (my mistake there) and remove the js file.
Other comments are small improvemnts.

OrKoN:
Thanks @richestrinh for contributing this. To expand on @Lightning00Blade comment about tests, if you already wrote a test  for the rule, let's put it next to the rule file if that works (i.e., `<ruleName>.test.ts`  and let's add a command to run the test using the https://nodejs.org/api/test.html#running-tests-from-the-command-line 

I have triggered a lint run and it looks like there is one special case when the shebang comes before the license (as it should be). There could be other exception so it might be great to include that in the test.

.eslintrc.js
```diff
@@ -149,6 +149,8 @@ module.exports = {
     'rulesdir/prettier-comments': 'error',
     // Enforces consistent file extension
     'rulesdir/extensions': 'error',
+    // Enforces license headers on files
+    'rulesdir/check-license': 'error',
```
OrKoN: I think the current version is already useful so if you want @richestrinh we could flip this to "warn" and land this as the first version

OrKoN:
I have prefixed the PR with `build: ` so that it does not show up in the changelog since it is not relevant for the end users.

tools/eslint/src/check-license.js
richestrinh: Sorry about this as well, I believe my updated fixer accounts for both cases as you mentioned

Lightning00Blade (approved):

Lightning00Blade:
@richestrinh I am happy to merge as is, and iterate on changes/features. So please rebase and open for review.

PR #11723, opened 2024-01-23
chore: implement higher order event emitters

jrandolf-google:
This PR updates the event emitter implementation to allow wrapping other event emitters.

For the base case, this will act the exact same.

If the emitter wraps another emitter, then the wrapper emitter will behave just like the inner emitter, except when disposed, **only the handlers attached to the wrapper will be removed from the inner emitter**.

The emitter has also been made disposable, so this PR provides a less verbose method of creating disposable sets of handlers over `EventSubscription`. It's also better typed.

### DIFF
--- a/packages/puppeteer-core/src/common/EventEmitter.ts
+++ b/packages/puppeteer-core/src/common/EventEmitter.ts
@@ -4,10 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import mitt, {
-  type Emitter,
-  type EventHandlerMap,
-} from '../../third_party/mitt/mitt.js';
+import mitt, {type Emitter} from '../../third_party/mitt/mitt.js';
 import {disposeSymbol} from '../util/disposable.js';
 
 /**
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/common/EventEmitter.ts
```diff
@@ -189,13 +207,24 @@ export class EventEmitter<Events extends Record<EventType, unknown>>
    * @returns `this` to enable you to chain method calls.
    */
   removeAllListeners(type?: keyof EventsWithWildcard<Events>): this {
-    if (type === undefined || type === '*') {
-      this.#handlers.clear();
-    } else {
-      this.#handlers.delete(type);
+    if (type !== undefined) {
```
Lightning00Blade, line 211: Nit:
```suggestion
    if (type) {
```

jrandolf: I'd prefer explicit `undefined` checks. What if the type is an empty string for example.

Lightning00Blade: Agree it can be but it should never be.

packages/puppeteer-core/src/common/EventEmitter.ts
```diff
@@ -74,14 +71,20 @@ export type EventsWithWildcard<Events extends Record<EventType, unknown>> =
...
 
   /**
+   * If you pass an emitter, the returned emitter will wrap the passed emitter.
+   *
    * @internal
    */
-  constructor() {
-    this.#emitter = mitt(this.#handlers);
+  constructor(
+    emitter: Emitter<EventsWithWildcard<Events>> | EventEmitter<Events> = mitt(
```
Lightning00Blade, line 83: Q: Could this be `CommonEventEmitter` to be more flexible?

packages/puppeteer-core/src/common/EventEmitter.ts
```diff
@@ -94,6 +97,13 @@ export class EventEmitter<Events extends Record<EventType, unknown>>
...
     handler: Handler<EventsWithWildcard<Events>[Key]>
   ): this {
+    const handlers = this.#handlers.get(type);
+    if (handlers === undefined) {
+      this.#handlers.set(type, [handler]);
+    } else {
+      handlers.push(handler);
+    }
+
     this.#emitter.on(type, handler);
```
Lightning00Blade, line 107: Nit: can we return here instead of `return this`?

Lightning00Blade (approved):

packages/puppeteer-core/src/common/EventEmitter.ts
jrandolf: No because that's not the correct `this`.

packages/puppeteer-core/src/common/EventEmitter.ts
jrandolf: Not at the moment because of some type mismatch.

PR #11727, opened 2024-01-23
 chore: add `dispose` to `core/UserContext`

jrandolf-google:


packages/puppeteer-core/src/bidi/core/UserContext.ts
```diff
@@ -59,11 +63,13 @@ export class UserContext extends EventEmitter<{
...
+    // ////////////////////
     // Session listeners //
-    // ///////////////////////
-    const session = this.#session;
-    session.on('browsingContext.contextCreated', info => {
+    // ////////////////////
+    const sessionEmitter = this.#disposables.use(
+      new EventEmitter(this.#session)
+    );
+    sessionEmitter.on('browsingContext.contextCreated', info => {
```
Lightning00Blade, line 72: Let's either use `session` or `sessionEmitter`, using both create more mental overhead. Ref previous PR.

jrandolf: I've settled on `sessionEmitter` due to name clashes

jrandolf: I've updated the UserPrompt as well.

Lightning00Blade (approved):

PR #11747, opened 2024-01-24
ci: use .nvmrc everywhere

Lightning00Blade:


.github/workflows/ci.yml
```diff
@@ -384,7 +382,7 @@ jobs:
       - name: Set up Node.js
         uses: actions/setup-node@b39b52d1213e96004bfcb1c61a8a6fa8ab84f3e8 # v4.0.1
         with:
-          node-version: ${{ matrix.node }}
+          node-version-file: '.nvmrc'
```
OrKoN: aren't we supposed to test on a different version here?

Lightning00Blade: Good question. But it was set up as 18 which is not the lowest supported version by Puppeteer atm.

OrKoN (approved):

PR #11752, opened 2024-01-25
refactor: use generic implementation for `waitForResponse`

jrandolf-google:


packages/puppeteer-core/src/common/types.ts
```diff
@@ -9,6 +9,11 @@ import type {JSHandle} from '../api/JSHandle.js';
 
 import type {LazyArg} from './LazyArg.js';
 
+/**
+ * @public
+ */
+export type AwaitablePredicate<T> = (value: T) => Awaitable<boolean>;
```
Lightning00Blade, line 15: Q: Didn't you add it something similar in another PR?

Lightning00Blade (approved):

packages/puppeteer-core/src/common/types.ts
jrandolf: In the original bidi/core PR I made this refactor. Separated it out.

PR #11753, opened 2024-01-25
refactor: use generic implementation for `waitForRequest`

jrandolf-google:


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1621,10 +1621,35 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
+  ): Promise<HTTPRequest> {
+    const {timeout: ms = this._timeoutSettings.timeout()} = options;
+    if (typeof urlOrPredicate === 'string') {
+      const url = urlOrPredicate;
+      urlOrPredicate = (request: HTTPRequest) => {
+        return request.url() === url;
+      };
+    }
+    const observable$ = merge(
+      fromEmitterEvent(this, PageEvent.Request),
```
Lightning00Blade: I think the previous version waited only for `Request`, let keep the old behavior.

Lightning00Blade (approved):

PR #11754, opened 2024-01-25
feat: download chrome-headless-shell by default and use it for the old headless mode

OrKoN:
With this PR, Puppeteer will download two binaries by default: a Chrome for Testing binary and a chrome-headless-shell of the same version. `chrome-headless-shell` will be used as a replacement for the old headless mode. 

Two new configurations are added:

1) `PUPPETEER_SKIP_CHROME_DOWNLOAD` to skip Chrome binary downloads (only chrome-headless-shell will be downloaded).
2) `PUPPETEER_SKIP_CHROME_HEADLESS_SHELL_DOWNLOAD` to skip chrome-headless-shell  binary downloads (only Chrome for Testing will be downloaded).

Use these variables if you only want to use the old headless mode (chrome-headless-shell) or only the new-headless/headful mode (Chrome for Testing). The existing configuration `PUPPETEER_SKIP_DOWNLOAD` would still apply to both binaries.


OrKoN:
Hey folks, this is a reland for https://github.com/puppeteer/puppeteer/pull/11093

packages/puppeteer-core/src/revisions.ts
```diff
@@ -9,5 +9,6 @@
  */
 export const PUPPETEER_REVISIONS = Object.freeze({
   chrome: '121.0.6167.85',
+  'chrome-headless-shell': '121.0.6167.85',
```
Lightning00Blade, line 12: This need to be kept automatically update, I think we need to update the automation tooling.

packages/puppeteer/src/node/install.ts
```diff
@@ -48,24 +48,64 @@ export async function downloadBrowser(): Promise<void> {
 
   const unresolvedBuildId =
     configuration.browserRevision || PUPPETEER_REVISIONS[product] || 'latest';
+  const unresolvedShellBuildId =
+    configuration.browserRevision ||
+    PUPPETEER_REVISIONS['chrome-headless-shell'] ||
+    'latest';
+
+  const shellBuildId =
```
Lightning00Blade: We should move it just before it is used else it may fail while you want only Chrome without shell.

OrKoN: Updated error reporting

packages/puppeteer/src/node/install.ts
```diff
@@ -48,28 +48,90 @@ export async function downloadBrowser(): Promise<void> {
...
+    }
+
+    if (browser === Browser.CHROME) {
+      if (configuration.skipChromeHeadlessShellDownload) {
+        logPolitely('**INFO** Skipping Chrome download as instructed.');
+      } else {
+        const shellBuildId =
+          browser === Browser.CHROME
+            ? await resolveBuildId(browser, platform, unresolvedShellBuildId)
+            : '';
```
Lightning00Blade: We check for this a few line prior.
```suggestion
        const shellBuildId = await resolveBuildId(browser, platform, unresolvedShellBuildId);
```

Lightning00Blade (approved):
LGTM, after my suggestion is accepted.

Lightning00Blade (approved):

PR #11757, opened 2024-01-25
refactor: use generic implementation for `waitForNetworkIdle`

jrandolf-google:
The new network code will eventually replace the bidi network code (and eventually replace the CDP one once CDP over Bidi happens)


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1699,34 +1710,44 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
     ).pipe(
       startWith(undefined),
-      filter(() => {
-        return networkManager.inFlightRequestsCount() <= requestsInFlight;
-      }),
-      switchMap(() => {
-        return of(undefined).pipe(delay(idleTime));
-      })
+      bufferCount(concurrency + 1, 1),
+      debounceTime(idleTime),
```
Lightning00Blade: I think we don't have a proper test, but we never check if the queue is empty or full here. 
This will mean that we can get a single Request event that never resolves and this will still emit.

jrandolf: Test added.

jrandolf: Discussed offline. This implementation is incorrect; idle means there are no in-flight requests.

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1699,34 +1753,51 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
+    } = options;
+
     return merge(
-      fromEmitterEvent(networkManager, NetworkManagerEvent.Request),
-      fromEmitterEvent(networkManager, NetworkManagerEvent.Response),
-      fromEmitterEvent(networkManager, NetworkManagerEvent.RequestFailed)
+      fromEmitterEvent(this, PageEvent.Request),
+      fromEmitterEvent(this, PageEvent.RequestFailed),
+      fromEmitterEvent(this, PageEvent.RequestFinished),
+      fromEmitterEvent(this, PageEvent.RequestServedFromCache),
```
Lightning00Blade: In both BiDi and CDP this event does not change the idle state, omit.

Lightning00Blade (approved):

PR #11772, opened 2024-01-29
fix: set viewport for element screenshots

### LINKED ISSUE #11597: `captureBeyondViewport` does not work in v21.6.1
It worked in 21.4.1

See https://ci.chromium.org/ui/p/devtools-frontend/builders/try/devtools_frontend_linux_dbg_fastbuild/b8760472979767066721/overview for the screenshot diffs and affected tests.

jrandolf-google:
If the element is larger than the viewport but within a scrollable element, the element will cut off (`captureBeyondViewport` will not help here). To alleviate this problem, we need to adjust the viewport to (hopefully) enlarge the scrollable element to capture the inner element.

Fixed: [11597](https://github.com/puppeteer/puppeteer/issues/11597)

### DIFF
--- a/packages/puppeteer-core/src/api/ElementHandle.ts
+++ b/packages/puppeteer-core/src/api/ElementHandle.ts
@@ -17,10 +17,15 @@ import type {
   NodeFor,
 } from '../common/types.js';
 import type {KeyInput} from '../common/USKeyboardLayout.js';
-import {isString, withSourcePuppeteerURLIfNone} from '../common/util.js';
+import {
+  debugError,
+  isString,
+  withSourcePuppeteerURLIfNone,
+} from '../common/util.js';
 import {assert} from '../util/assert.js';
 import {AsyncIterableUtil} from '../util/AsyncIterableUtil.js';
 import {throwIfDisposed} from '../util/decorators.js';
+import {AsyncDisposableStack} from '../util/disposable.js';
 
 import {_isElementHandle} from './ElementHandleSymbol.js';
 import type {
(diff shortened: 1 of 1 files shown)


packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -1341,6 +1346,31 @@ export abstract class ElementHandle<
...
+    // If the element is larger than the viewport, `captureBeyondViewport` will
+    // _not_ affect element rendering, so we need to adjust the viewport to
+    // properly render the element.
+    const viewport = page.viewport() ?? {
+      width: clip.width,
+      height: clip.height,
+    };
+    await using stack = new AsyncDisposableStack();
+    if (clip.width > viewport.width || clip.height > viewport.height) {
+      await this.frame.page().setViewport({
```
OrKoN, line 1358: Was the change of viewport done in the previous version? Do you have a link to an older implementation that has this step? We already had an issue that changing the viewport would affect the page results.

Why does captureBeyondViewport not affect the element rendering (it should affect the entire document, so clip can be done correctly)?

jrandolf: > Was the change of viewport done in the previous version?

We used to change the viewport: https://github.com/puppeteer/puppeteer/blob/b4507c69c152c275102d9908f06ec3b2dc3fb7b4/packages/puppeteer-core/src/common/ElementHandle.ts#L263

We don't do it for clips on the Page.screenshot method, but we do it for the element handle screenshot method.

> Why does captureBeyondViewport not affect the element rendering (it should affect the entire document, so clip can be done correctly)?

Not sure why, but the proof is in your bug: https://github.com/puppeteer/puppeteer/issues/11597
While trying to reproduce the issue, I saw the parameters sent to `captureScreenshot` were the same, but the viewport was the default size (1280 x 800) as opposed to the element size as seen in past versions of Puppeteer.

I have been trying to recreate the test, but it's clear from your issue that captureBeyondViewport doesn't necessarily imply rendering. In particular, the element we are trying to capture is a canvas element; it makes sense that we don't render the entire canvas automatically.

jrandolf: Actually, after trying to reproduce this test case, it may be that the canvas element inside your bug is in an iframe that is affected by layout as opposed to a rendering problem. The tests show rendering is not an issue.

jrandolf:
As discussed offline, we will merge this and create a breaking change in the next release.

OrKoN (approved):

PR #11773, opened 2024-01-29
fix: wrap error if extracting fails for any reason

Lightning00Blade:
Better error handling when the unpacking fails, this may be do to corrupted download.


OrKoN (commented):
Is it still relevant?

Lightning00Blade:
No, this will not prevent https://github.com/regular/unbzip2-stream/issues/28, so closing

PR #11774, opened 2024-01-29
fix!: remove viewport resizing from ElementHandle.screenshot

jrandolf-google:
Viewport resizing was previously done to fix issues related to hidden elements within `overflow: hidden` containers or viewport-sized iframes.

This PR removes this logic since this use-case is better handled by either:

 - explicitly setting the viewport to the element size using `page.setViewport({clip: await element.boundingBox()});` (this declaratively implies the element being screenshotted is dependent on the viewport size), or
 - changing the element's state for testing to ensure the element is visible.

The latter depends on your codebase. For example, if the element is hidden in a `overflow: hidden` container, just `page.evaluate` away the `overflow: hidden`.


OrKoN (approved):
LGTM % build failures 

And let's land it together with the rest of the breaking changes.

PR #11776, opened 2024-01-29
refactor!: change the way cookies are retrieved

sadym-chromium:
Currently, CDP method `network.getCookies` makes a part of filtering, but it is not guaranteed the returned cookies will actually be sent to the server, as it would depend on the request itself. It is not feasible to implement when WebDriver BiDi protocol is used. With this PR, the cookies will be filtered only based on the domain and path. The actual cookies sent to each specific domain can be retrieved via network request interception.
* https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.3
* https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.4

BREAKING CHANGE: the `Page.cookies()` method became less strict.

### DIFF
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -7,6 +7,7 @@
 import type {Readable} from 'stream';
 
 import type {Protocol} from 'devtools-protocol';
+import net from 'net';
 
 import {
   concat,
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1306,6 +1307,91 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
+    }
+    return false;
+  }
+
+  /**
+   * Checks the cookie matches the URL according to the spec:
+   * * https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.3
+   * * https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.4
+   */
+  static testCookieUrlMatch(cookie: Protocol.Network.Cookie, url: string): boolean {
```
sadym-chromium: Not sure what is the best place for this methods, so I decided to keep it here for now.

Lightning00Blade:
@sadym-chromium I am not sure if this will be picked up correctly by release-please, so please use `feat!`,`fix!` or `refactor!`

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+import net from 'net';
```
OrKoN, line 7: is it a node module? we cannot use node modules in Puppeteer because it has to run in the browser.

OrKoN:
@sadym-chromium is it still relevant?

PR #11777, opened 2024-01-29
refactor!: remove `error` const, change CustomError to PuppeteerError

Lightning00Blade:
Remove deprecated errors const.  If you used to check if any error came from  Puppeteer via `CustomError` you should simply change it to `PuppeteerError`.

Before: 

```ts

puppeteer.errors.TimeoutError
puppeteer.CustomError
```

After:

```ts
import { TimeoutError, PuppeteerError } from "puppeteer";
```

or 

```ts
puppeteer.TimeoutError
puppeteer.PuppeteerError
```


OrKoN:
@Lightning00Blade edited the message a bit. PTAL

packages/puppeteer-core/src/common/Errors.ts
```diff
@@ -5,11 +5,11 @@
  */
 
 /**
- * @deprecated Do not use.
+ * All Puppeteer Error should extend this one
```
OrKoN:
```suggestion
 * The base class for all Puppeteer-specific errors
```

OrKoN (approved):

PR #11510, opened 2023-12-06
feat: add outline (document outline) PDFs option

bungeman:
**What kind of change does this PR introduce?**

feature

**Did you add tests for your changes?**

Yes

**If relevant, did you update the documentation?**

Yes

**Summary**

Chromium [introduced generateDocumentOutline](https://chromium.googlesource.com/chromium/src.git/+/9c2c6e785b9c644887c466ae49d7947e6c690171) in 121. This allows for a PDF document outline (sometimes known as bookmarks) to be generated from the html document outline. This CL adds a `outline` property to request this extra PDF output.

**Does this PR introduce a breaking change?**

No

**Other information**

This change is heavily based on https://github.com/puppeteer/puppeteer/pull/11182 .

PR #11779, opened 2024-01-29
feat: add outline to PDF generation

### LINKED ISSUE #1778: feature request: add option to generate TOC for pdf output
Since now headers and footers with page numbers work, I now desperately miss an option to generate a Table of Contents (TOC) out of the h1- h7 headers when generating a pdf file (i.e like wkhtmltopdf is doing this). The TOC should be at the start of the pdf and it should not only be clickable (jump to the page) but also generate the outline pdf element, so that the TOC is displayed in the contents view in any viewer. Although this may sound complicated if this functionality is implemented at the right place it is not that complicated (take a look on how wkhtmltopdf is implementing this).

Before posting here I tried a couple of workarounds to achieve this.
Some dead ends:
 - CSS3 target_counter() -> proposed a long time ago and only some specialised tools do it, to be honest I've given up to think that it will be implemented in chrome some day (reference issue is now: https://bugs.chromium.org/p/chromium/issues/detail?id=368053 )
 - Find a tool or tools to extract the table of contents from the generated pdf and generate a preface.pdf with the TOC witch to merge in-front of the original pdf ... With https://github.com/qpdf/qpdf I was able to generate a readable and searchable "pdf text-file" so that theoretically it was possible to find the header in the text file and via reverse search and the added comments find out on wich page it is etc. etc...

Any chance to get this soon?
Thanks
Ognian

Lightning00Blade:
Adds support for PDF outline (TOC) _**only**_ for old Headless Chrome (`headless: 'shell'`).
Related crbug: crbug/840455
Closes #11510, Closes https://github.com/puppeteer/puppeteer/issues/1778
image


packages/puppeteer-core/src/common/util.ts
```diff
@@ -375,6 +376,19 @@ export function parsePDFOptions(
       convertPrintParameterToInches(options.margin?.right, lengthUnit) || 0,
   };
 
+  // Quirk https://bugs.chromium.org/p/chromium/issues/detail?id=840455#c44
+  if (options.outline) {
+    options.tagged = true;
+  }
+
+  console.log({
```
OrKoN: remove

test/src/cdp/pdf.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {readFile, unlink} from 'fs/promises';
+
+import expect from 'expect';
+
+import {getTestState, setupTestBrowserHooks} from '../mocha-utils.js';
+
+describe.only('Page.pdf', () => {
```
OrKoN: remove

test/src/cdp/pdf.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  setupTestBrowserHooks();
+
+  it('can print to PDF with accessible', async () => {
```
OrKoN, line 16: this will conflict with the other PR

Lightning00Blade: Let's merge it and I will resolve the conflicts on the other PR, I think we should flip this as well.

OrKoN: it is very new so perhaps we won't turn it on by default

test/TestExpectations.json
```diff
@@ -269,6 +269,12 @@
     "parameters": ["webDriverBiDi"],
     "expectations": ["SKIP"]
   },
+  {
+    "testIdPattern": "[spec.spec] Page.pdf *",
```
OrKoN: what is spec.spec? I think this might be skipping to much?

Lightning00Blade:
```suggestion
    "testIdPattern": "[pdf.spec] Page.pdf *",
```

Lightning00Blade: This currently should do nothing... It should be `pdf.spec`, we don't support either `tagged` or `outlined` PDFs for BiDi.

OrKoN: But we do support some other pdf methods? Please double check that we are not skipping more tests

Lightning00Blade: I will tomorrow, but this file only includes those 2 options. The other still live on the page.spec

Lightning00Blade: It skips only the two test as expected.

OrKoN (approved):

dannybeckett:
I am not seeing any bookmarks created when trying to use this on my PDFs. I added `outline: true` to the `await page.pdf({ });` and I added `headless: true` to the `const browser = await puppeteer.launcn({ });` 

Is there something more which is needed? How can I troubleshoot this please?

Thanks!

klausjagd:
I'm having the same issue: not getting the outline when adding outline:true and setting headless:true.

OrKoN:
Use headless: 'shell' as this is only supported for the old headless which is no longer the default.

klausjagd:
thanks :-)

dannybeckett:
Nice, thank you @OrKoN

Too bad page numbers in footers are broken in `headless:shell` - it seems that it's either a choice between having bookmarks or having page numbers 🙈 - per #12361

Lightning00Blade:
@dannybeckett Here is Chromium bug that tracks the outline property - https://issues.chromium.org/issues/41387522#comment48 
To learn more about `shell` (`chrome-headless-shell`) refer to https://developer.chrome.com/blog/chrome-headless-shell.

bungeman:
With [Pass GenerateDocumentOutline through new headless](d2f63d34ecd4e9c60829eceea43817b18f23edc9) (in Chrome 126.0.6450.0 and later) things like  `chrome --headless=new --print-to-pdf --no-pdf-header-footer --generate-pdf-document-outline=true` now work and emit a document outline. Passing `generateDocumentOutline: true` through the Chrome Devtools Protocol also works (here in puppeteer exposed as `outline: true`).

PR #11780, opened 2024-01-29
refactor!: remove waitForTimeout

Lightning00Blade:
We discourage waiting for arbitrary time. Here are better approaches:

- Using `waitForSelector` to wait for an element to show up on the page.
- Using [locator](https://pptr.dev/guides/locators) API to wait for and do thing with elements
- Using [waitFor](https://pptr.dev/api/puppeteer.waitforoptions) when navigating a page.

Alternatively, you can replace `waitForTimeout(millisecond)` calls with `new Promise(r => setTimeout(r, milliseconds))`.


OrKoN (approved):

Kikobeats:
@Lightning00Blade @OrKoN can you provide a equivalent example for migrating `waitForTimeout` when using puppeteer@22?

The documention linked is not providing how to do what `waitForTimeout` achieves. 

It clear `new Promise(r => setTimeout(r, milliseconds))` does the same but not how to do the same.

OrKoN:
@Kikobeats could you clarify which part is not clear? Awaiting a `new Promise(r => setTimeout(r, milliseconds))` was literally implementation of waitForTimeout so this snippet does the same thing.

loliver:
@Kikobeats I was able to get my old Jest Puppeteer `waitForTimeout` functionality in `jest-image-snapshot` tests by replacing with the following:
```typescript
await page.evaluate(() => new Promise(r => setTimeout(r, 1000)))
```

OrKoN:
@loliver note that the original waitForTimeout didn't evaluate the timeout in the page context. So the version you use is not equivalent and if the page overrides setTimeout it won't work in the same way.

loliver:
@OrKoN thanks for the clarification, our usage is pretty simple so fairly confident `setTimeout` won't be overridden in any way.

Kikobeats:
@OrKoN ohh got it. I though it needed to wait for timeout inside the page, but not really since the browser is a standalone process so just wait at nodeland is enough 🙂 thanks for the clarification.

OrKoN:
Yes, it was node-side. Also I learned recently that Node implements a promisifed version now https://nodejs.org/api/timers.html#settimeoutcallback-delay-args

```
const { setTimeout } = require('node:timers/promises');
await setTimeout(1000);
```

PR #11782, opened 2024-01-29
refactor!: remove `$x` and `waitForXpath`

Lightning00Blade:
Removes the deprecated `$x` (replace with `$$`) and `waitForXpath` (replace with `waitForSelector`).

If the `xpath` starts with `//` you will need to prefix it with dot - `//<xpath>` becomes `xpath/.//<xpath>`.

`$x` Before:
```ts
page.$x('//p[normalize-space(.)="hello world"]');
```
`$$` After (note the dot before the xpath):
```ts
page.$$('xpath/.//p[normalize-space(.)="hello world"]')
```

`waitForXpath`  Before:
```ts
page.waitForXpath('//p[normalize-space(.)="hello world"]');
```
`waitForSelector` After (note the dot before the xpath):
```ts
page.waitForSelector('xpath/.//p[normalize-space(.)="hello world"]')
```


OrKoN:
@Lightning00Blade I don't think you need a dot for page level selectors. AFAIK you only need it if your root is another element. Let's include that in the description.

Lightning00Blade:
The `$x` implementation is a short cut to `ElementHandle` for the `document` and then we prefix it with the `.//` if needed.
So I believe we will need to add it here as well.

OrKoN:
@Lightning00Blade yeah, that is implementation details. I believe you don't need the . if you query the entire document.

OrKoN:
@alexey-sh See https://pptr.dev/guides/query-selectors for Puppeteer selectors. Basically, we don't want to maintain the separate code path only for xpath when Puppeteer selectors allow combining multiple types of queries into one selector. The value is that it makes it easier to write universal querying code.

test/src/elementhandle.spec.ts
```diff
@@ -490,20 +488,18 @@ describe('ElementHandle specs', function () {
...
-          path
-        )) as ElementHandle<HTMLDivElement>;
-        expect(
-          await e.evaluate(el => {
-            return el.id;
-          })
-        ).toStrictEqual('el2');
-      }
+      using elByXpath = (await elById.waitForSelector(
+        'xpath/.//div'
```
alexey-sh, line 496: The previous implementation contains two options
```
for (const path of ['//div', './/div']) {
```

I don't see in the updated test the same amount.

karlhorky:
> I don't think you need a dot for page level selectors. AFAIK you only need it if your root is another element. Let's include that in the description.

@OrKoN If you're right, how would you write the following XPath selectors without the `.` ?

> ```ts
> page.$$('xpath/.//p[normalize-space(.)="hello world"]')
> ```
>
> ```ts
> page.waitForSelector('xpath/.//p[normalize-space(.)="hello world"]')
> ```

karlhorky:
Another question that I have here is how this relates to [the `-p-xpath` syntax on the Query Selectors page](https://pptr.dev/guides/query-selectors/#xpath-selectors--p-xpath):

```js
const element = await page.waitForSelector('::-p-xpath(h2)');
```

a10kiloham:
Oof this is still super unclear. The xpath/// style references dont seem to work at all for me. I believe the xpath/.// is working but trying to verify now.

test/src/elementhandle.spec.ts
Lightning00Blade: When querying from an element the `./` prefix is required.

a10kiloham:
There's precious little on the p-xpath either. I've tried the following
```
let foundHandle = await page
   .waitForSelector( `::-p-xpath("//a[text()[contains(., "${pageType}")]]", document.body)`,
  {
    timeout: 3000,
  },
 )
```
And various permutations where pageType is a simple one word string (with and without quotations marks) and none seem to work properly as expected. I've tried to debug but I can't seem to find any other place where the p-xpath is used for quick manual debugging like I could in the Chrome console. 

Is there any where else to easily/rapidly test iterations of this syntax?

a10kiloham:
that's my issue, i'm not clever enough to figure out a better way. i'm looking for specific links on the page. So for example, in th3e middle of the page an a href with "Insights" as the only text inside. However other various places on the page seem to match and there's no specific consistent aria or class or id in the link I'm searching for, so I'd have to at best iterate through all the results and filter somehow which is doable but also less familiar to how I'm used to jus using xpaths. I saw with document.evaluate() I could step through the results but waitforselector doesn't seem to function that way.

alexw23:
If there's anyone like me where this is literally hundreds of instances, please see a script below:

```bash
#!/bin/bash

# Directory containing the JavaScript files
DIRECTORY="./src"

# Loop over each JavaScript file in the directory
for file in $(find $DIRECTORY -type f -name "*.js"); do
    # Use sed to modify the files in place
    sed -i -r 's/page\.waitForXPath\(\s*['"'"'"]([^'"'"'"]*)['"'"'"],\s*\{([^}]*)\}\s*\);/page.waitForSelector('"'"'xpath\/.\1'"'"', {\2});/g' "$file"
done

echo "Conversion complete."
```

fakefish:
可以理解需要优化选择器的写法，但这xpath/.写法不觉得有点丑么，感觉像一道伤疤在代码里，希望能有更加美观的方式

PR #11784, opened 2024-01-30
feat(bidi): implement UserContexts

OrKoN:
This PR implements https://w3c.github.io/webdriver-bidi/#user-contexts for WebDriver BiDi protocol in Puppeteer.


jrandolf-google (approved):

sadym-chromium:
User context should be passed to [`browsingContext.create`](https://github.com/puppeteer/puppeteer/blob/6f828df989666525e1cefcc55ca47d00efe9b2b9/packages/puppeteer-core/src/bidi/BrowserContext.ts#L71) in `BrowserContext.newPage` method

OrKoN:
@sadym-chromium I think this is mostly ready but I will go over the change to see if there could be any flakiness

sadym-chromium (approved):

PR #11790, opened 2024-01-30
fix!: make console warn level compatible with WebDriver BiDi

OrKoN:
In CDP the console.warn resulted in a warning level, whereas HTML and WebDriver BiDi spec use `warn` to denote this level. To migrate, if you used ConsoleMessageType and checked for `warning`, check for `warn` instead.

### DIFF
--- a/docs/api/puppeteer.consolemessagetype.md
+++ b/docs/api/puppeteer.consolemessagetype.md
@@ -14,7 +14,7 @@ export type ConsoleMessageType =
   | 'debug'
   | 'info'
   | 'error'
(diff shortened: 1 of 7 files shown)


Lightning00Blade (approved):
LGTM + but we should consider doing it the other way around (Change the CDP to match BiDi) because of 
https://console.spec.whatwg.org/#warn clearly being `warn` and not `warning`

OrKoN:
@Lightning00Blade makes sense

Lightning00Blade (approved):

PR #11794, opened 2024-01-31
chore: for Firefox keep preference "browser.tabs.closeWindowWithLastTab" disabled for CDP

whimboo:
Upstream version of the patch for https://phabricator.services.mozilla.com/D200073.

We have to keep this preference disabled for CDP given that it looks complicated to fix. For BiDi (and classic) we have to remove it because it won't allow closing the last tab of any open (not the last) window.

@OrKoN can you please review? Thanks.

### DIFF
--- a/packages/puppeteer-core/src/node/FirefoxLauncher.ts
+++ b/packages/puppeteer-core/src/node/FirefoxLauncher.ts
@@ -45,6 +45,8 @@ export class FirefoxLauncher extends ProductLauncher {
       ...(protocol === 'webDriverBiDi'
         ? {}
         : {
+            // Do not close the window when the last tab gets closed
+            'browser.tabs.closeWindowWithLastTab': false,
             // Temporarily force disable BFCache in parent (https://bit.ly/bug-1732263)
             'fission.bfcacheInParent': false,
           }),


whimboo:
I don't see why this change can affect a navigation test for WebDriver BiDi. Most likely something else caused it to start failing?

OrKoN:
@whimboo not sure, but it seems to fail consistently. The main branch seems okay.

whimboo:
> @whimboo not sure, but it seems to fail consistently. The main branch seems okay.

This is strange given that this change should basically be a no-op for WebDriver BiDi. I'll check if I can reproduce locally on a Linux machine. At least on MacOS the test works all fine.

whimboo:
I saw that the `Linux-firefox-5b7cdfb86b1cd4c522639ee6300c1dd5b0067cb15712eeffa51cc2d6adb7a79e-1af16df9570ec5740ff10d004a86ef73e466c30813096caf3f8eea090084b474` cache entry is a week old. Given that we landed the feature recently it might be the reason why you cannot use the user context features yet. I cleared the cache for now.

Note that even without clearing the cache the test is passing now after the recent force-push from @OrKoN. Not sure why.

whimboo:
Oh, so that's a flaky test which got now marked as `PASS`/`TIMEOUT` over on https://github.com/puppeteer/puppeteer/pull/11799. Let me rebase.

whimboo:
@OrKoN it looks fine now. Please check.

OrKoN (approved):

PR #11799, opened 2024-01-31
test: mark flaky test

sadym-chromium:
Update test expectations and mark flaky test.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -2538,6 +2538,12 @@
     "expectations": ["SKIP"]
   },
   {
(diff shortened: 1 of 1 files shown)


Lightning00Blade (approved):
LGTM + file an issue as we mark it as `SKIP` so we can lose track of it.

sadym-chromium:
> LGTM + file an issue as we mark it as `SKIP` so we can lose track of it.

Kept expectations `timeout`, as with `skip` some other test failing with timeout for whatever reason.

whimboo:
I saw the same over on https://github.com/puppeteer/puppeteer/pull/11794. That is a new failure, right? So far I was not able to reproduce it yet. So we will investigate later once it's available in our tracking spreadsheet.

PR #11804, opened 2024-02-01
refactor!: enable the new-headless mode by default

OrKoN:
This PR changes the meaning for the `headless` flag. Previously, `headless=true` launched the old headless mode also known as `chrome-headless-shell`. After this change, the new headless module will be launched instead.

To migrate:

1) if you were using 'headless=true' and want to use `chrome-headless-shell`, change to `headless: 'chrome-headless-shell'`.
2) if you were using 'headless=new', change to `headless: true`


test/TestExpectations.json
```diff
@@ -2252,7 +2252,7 @@
   {
     "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Puppeteer.launch should be able to launch Firefox",
     "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["firefox", "headless"],
+    "parameters": ["chrome-headless-shell", "firefox"],
```
Lightning00Blade: Wrong? Params are `AND` logic. There are more instances of this change please revert. Also I think for FF we can drop the headless in params, wdyt?

test/TestSuites.json
```diff
@@ -45,7 +45,7 @@
     {
       "id": "chrome-bidi",
       "platforms": ["linux"],
-      "parameters": ["chrome", "headless", "webDriverBiDi"],
+      "parameters": ["chrome", "chrome-headless-shell", "webDriverBiDi"],
```
Lightning00Blade, line 48: Let's at least add a Issue to switch BiDi testing as well.

Lightning00Blade (approved):

PR #11805, opened 2024-02-01
refactor!: use ReadableStreams

Lightning00Blade:
Affected APIs - [createPDFStream](https://pptr.dev/api/puppeteer.page.createpdfstream) 
Migrating from [Node's Readable](https://nodejs.org/api/stream.html#class-streamreadable) to [ReadableStream](https://nodejs.org/api/stream.html#class-streamreadable).
Please refer to the ReadableStream documentation and update accordingly.
In most cases you would want to:
```ts
const stream = await page.createPDFStream()
const reader = stream.getReader();
while (true) {
      const {done, value} = await reader.read();
      if (done) {
        break;
      }
      buffers.push(value);
    }
}
```

### DIFF
--- a/packages/puppeteer-core/src/common/util.ts
+++ b/packages/puppeteer-core/src/common/util.ts
@@ -5,13 +5,10 @@
  */
 
 import type FS from 'fs/promises';
-import type {Readable} from 'stream';
 
 import {map, NEVER, Observable, timer} from '../../third_party/rxjs/rxjs.js';
 import type {CDPSession} from '../api/CDPSession.js';
-import {isNode} from '../environment.js';
 import {assert} from '../util/assert.js';
-import {isErrorLike} from '../util/ErrorLike.js';
 
 import {debug} from './Debug.js';
 import {TimeoutError} from './Errors.js';
@@ -209,29 +206,39 @@ export async function importFSPromises(): Promise<typeof FS> {
  * @internal
  */
 export async function getReadableAsBuffer(
-  readable: Readable,
+  readable: ReadableStream<Uint8Array>,
   path?: string
 ): Promise<Buffer | null> {
   const buffers = [];
+  const reader = readable.getReader();
   if (path) {
     const fs = await importFSPromises();
     const fileHandle = await fs.open(path, 'w+');
     try {
-      for await (const chunk of readable) {
-        buffers.push(chunk);
-        await fileHandle.writeFile(chunk);
+      while (true) {
+        const {done, value} = await reader.read();
+        if (done) {
+          break;
+        }
+        buffers.push(value);
+        await fileHandle.writeFile(value);
       }
     } finally {
       await fileHandle.close();
     }
   } else {
-    for await (const chunk of readable) {
-      buffers.push(chunk);
+    while (true) {
+      const {done, value} = await reader.read();
+      if (done) {
+        break;
+      }
+      buffers.push(value);
     }
   }
   try {
     return Buffer.concat(buffers);
   } catch (error) {
+    debugError(error);
     return null;
   }
 }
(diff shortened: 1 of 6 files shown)


packages/puppeteer-core/src/common/util.ts
```diff
@@ -209,69 +206,73 @@ export async function importFSPromises(): Promise<typeof FS> {
...
-  if (!isNode) {
-    throw new Error('Cannot create a stream outside of Node.js environment.');
-  }
+): Promise<ReadableStream<Uint8Array>> {
+  return new ReadableStream({
+    async pull(controller) {
+      function getUnit8Array(data: string, isBase64: boolean): Uint8Array {
+        let parsedData = data;
+        if (isBase64) {
+          parsedData = atob(data);
```
OrKoN: I think we need to convert directly to Unit8Array instead using TextEncode

OrKoN (approved):

TiAlRo:
The correct link to the migration target would be https://nodejs.org/api/webstreams.html#class-readablestream, not https://nodejs.org/api/stream.html#class-streamreadable. I used pipe, which does not exist on ReadableStream. When using pipeTo, I get the following typescript error:  error TS2345: Argument of type 'Writable' is not assignable to parameter of type 'WritableStream<Uint8Array>'.
  Type 'WritableBase' is missing the following properties from type 'WritableStream<Uint8Array>': locked, abort, close, getWriter

TiAlRo:
Example:
```
import express from "express";

async function getPDF(_, writableStream: NodeJS.WritableStream) {
  await page.setContent("Example");
  const pdfStream = await page.createPDFStream();
  pdfStream.pipe(writableStream);
}

const app = express();
app.get("", getPDF);
app.listen(3002);
```

TiAlRo:
I found a solution:
```
import { Request } from "express-serve-static-core";
const { Readable } = require("stream");
import express from "express";

async function getPDF(_: Request, writableStream: NodeJS.WritableStream) {
  await page.setContent("Example");
  const pdfWebStream = await page.createPDFStream();
  const pdfStream = Readable.fromWeb(pdfWebStream);
  pdfStream.pipe(writableStream);
}

const app = express();
app.get("", getPDF);
app.listen(3002);
```

outbackStack:
```js
const {done, value} = await reader.read();
```

the `value` is a buffer so you can response back to that in express.

Here's my example using fastify

```
fastify.get('/', async(request, reply) {

  const stream = await page.createPDFStream()
  const reader = stream.getReader();
  
  const {done, value} = await reader.read();
  
  return reply
          .header("Content-Type", "application/pdf; charset=utf-8")
          .send(value);
})
```

Fxlr8:
What was the intention of this change?

crazyoptimist:
I don't think this PR is a refactor; rather, it's a breaking change. Additionally, the puppeteer docs still lack a usage guide for the `createPDFStream` method. I'll try to improve the documentation.

Thanks for sharing your findings, @TiAlRo. You saved my time because I don't have a deep understanding of the node stream API.

OrKoN:
@crazyoptimist this refactor is marked as a breaking change (see `!`) and it shipped in a major release (breaking changes according to semver)

crazyoptimist:
Okay.

Could you help me to convert the ReadbleStream<Uint8Array> to a nodejs stream.Readable? I tried @TiAlRo's solution but still get this type error:

```
Argument of type 'ReadableStream<Uint8Array>' is not assignable to parameter of type 'ReadableStream<any>'.
     Type 'ReadableStream<Uint8Array>' is missing the following properties from type 'ReadableStream<any>': values, [Symbol.asyncIterator] [2345]
```

OrKoN:
@crazyoptimist I recommend asking on StackOverflow how to correctly adjust the types in your app (TypeScript, I assume). A quick workaround is to add a comment `// @ts-exect-error` where there is a type definition mismatch.

crazyoptimist:
Okay, I used `// @ts-expect-error` as a workaround for now, and it works. I don't have time to dig into what's wrong but there's something wrong in this PR regarding the types.

Thanks for the workaround, @OrKoN.

crazyoptimist:
Found the solution [here](https://stackoverflow.com/questions/37614649/how-can-i-download-and-save-a-file-using-the-fetch-api-node-js/74722818#comment133510726_74722818) from antonok's comment.

There are 2 different ReadableStream definitions.

``` javascript
import puppeteer, { Browser } from 'puppeteer';
import { Readable } from 'stream';
import { ReadableStream } from 'stream/web';

/**
 * This function renders HTML contents to a PDF file stream
 * @param html HTML string
 */
async function renderToStream(html: string) {
  let browser: Browser;

  try {

    await page.setContent(html).catch(() => browser.close());

    const webStream = await page.createPDFStream();

    const nodeStream = Readable.fromWeb(webStream as ReadableStream<any>);

    nodeStream.on('end', () => browser.close());

    return nodeStream;
  } catch (err) {
    console.error('PDF rending to stream failed: ', err);

    if (browser) {
      browser.close();
    }
  }
}
```

PR #11807, opened 2024-02-01
refactor: fixes flakiness

Lightning00Blade:
`EMPTY` never emits so we don't continue to the other parts.

### DIFF
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -10,7 +10,6 @@ import type {Protocol} from 'devtools-protocol';
 
 import {
   concat,
-  EMPTY,
   filter,
   filterAsync,
   first,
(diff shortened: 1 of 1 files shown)


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1779,7 +1778,7 @@ export abstract class Page extends EventEmitter<PageEvents> {
       startWith(this.#requestsInFlight),
       switchMap(() => {
         if (this.#requestsInFlight > concurrency) {
-          return EMPTY;
```
OrKoN, line 1782: LGTM but I wonder why was it flaky instead of failing every time?

OrKoN (approved):

packages/puppeteer-core/src/api/Page.ts
OrKoN: The EMPTY appears to be correct here and I am not sure why it's not working. The error seems to happen if we start piping why the requestsInFlight=1

jrandolf-google (changes requested):
EMPTY is correct. When the number of requests in flight are greater than concurrency, we don't want to emit anything.

PR #11815, opened 2024-02-02
refactor!: enable the new-headless mode by default

OrKoN:
This PR changes the meaning for the `headless` flag. Previously, `headless=true` launched the old headless mode also known as `chrome-headless-shell`. After this change, the new headless module will be launched instead.

To migrate:

1) if you were using 'headless=true' and want to use `chrome-headless-shell`, change to `headless: 'chrome-headless-shell'`.
2) if you were using 'headless=new', change to `headless: true`


Lightning00Blade (approved):
LGTM

ouuan:
I think it's not user-friendly first to suggest users to change their codes to `headless: 'new'` and then break it soon after. If I knew this beforehand, I would rather keep my old `headless: true` to avoid changing it everywhere twice.

Maybe it's too late for this time, and further changes might cause more confusion. But when there are similar changes next time, I suggest that a better decision could be made. e.g. Set it to `boolean | 'new' | 'shell'` in the early version, and then change the semantic of `true` in the next version, but keep both `'new'` and `'shell'`.

OrKoN:
@ouuan thanks for the feedback. Note that it is a breaking semver release, so we have done breaking changes to not support `new` anymore. Note that it was a mechanism for an early testing and the warning always mentioned that `headless: true` will become the new headless mode by default.

ouuan:
> the warning always mentioned that `headless: true` will become the new headless mode by default.

Yes, but it did not mention that `'new'` will be removed.

> Note that it is a breaking semver release, so we have done breaking changes to not support `new` anymore.

A breaking change is necessary to change the semantic of `true`. But the removal of `'new'` is not necessary.

PR #11825, opened 2024-02-02
test: replace usages of webkit-scrollbar with scrollbar-width

whimboo:
`webkit-scrollbar` is not cross-browser compatible, but `scrollbar-width` is. Given that Chrome since version 121 supports it we can replace the usages.

Nearly all the screenshot tests were affected by this and should now work on non-retina displays.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -3377,12 +3377,6 @@
     "expectations": ["FAIL", "PASS"]
   },
   {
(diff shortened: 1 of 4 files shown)


OrKoN (approved):

whimboo:
It looks like that for CDP some tests are still not working. I'll partly revert the expectation data.

PR #11826, opened 2024-02-02
build: fix EsLint rule and add fixer

Lightning00Blade:


.eslintrc.js
```diff
@@ -277,5 +275,14 @@ module.exports = {
         },
       ],
     },
+
+    {
+      // Applies to only published packages
+      files: ['packages/**/*.ts'],
+      rules: {
+        // Error if comments do not adhere to `tsdoc`.
+        'tsdoc/syntax': 'error',
```
Lightning00Blade, line 284: This became too noisy, so scoped it only to `packages/` so only packages we publish need to comply for Documentation reasons.

OrKoN (approved):

PR #11833, opened 2024-02-05
refactor: move bidi CDPSession

jrandolf-google:


OrKoN:
Let's not land refactoring before the release is out

OrKoN:
@Lightning00Blade PTAL

packages/puppeteer-core/src/bidi/CDPSession.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+  }
+
+  override connection(): CdpConnection | undefined {
+    return undefined;
+  }
+
+  override async send<T extends keyof ProtocolMapping.Commands>(
+    method: T,
+    ...paramArgs: ProtocolMapping.Commands[T]['paramsType']
```
Lightning00Blade: Nit: Just to fix the correct type from the base class if its a no-op change.
```suggestion
    params?: ProtocolMapping.Commands[T]['paramsType'][0],
```

Lightning00Blade (approved):
LGTM

PR #11847, opened 2024-02-06
chore: add storage.getCookies and storage.setCookie to bidi/core

jrandolf-google:


packages/puppeteer-core/src/bidi/core/UserContext.ts
```diff
@@ -174,6 +175,45 @@ export class UserContext extends EventEmitter<{
...
+  async setCookie(
+    cookie: Bidi.Storage.PartialCookie,
+    sourceOrigin?: string
+  ): Promise<void> {
+    await this.#session.send('storage.setCookie', {
+      cookie,
+      partition: {
+        type: 'storageKey',
+        sourceOrigin,
+        context: this.id,
```
OrKoN:
```suggestion
        userContext: this.id,
```

packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
```diff
@@ -464,6 +472,39 @@ export class BrowsingContext extends EventEmitter<{
...
+  @throwIfDisposed<BrowsingContext>(context => {
+    // SAFETY: Disposal implies this exists.
+    return context.#reason!;
+  })
+  async getCookies(
+    options: GetCookiesOptions = {}
+  ): Promise<Bidi.Network.Cookie[]> {
+    const {
+      result: {cookies},
+    } = await this.#session.send('storage.getCookies', {
```
OrKoN, line 484: let's have a single place doing calls and these methods should delete to UserContext.* which should accept an optional BrowsingContext as a parameter?

jrandolf: That would break encapsulation. You could pass in a browsing context from a different user context which would pass but technically not be a valid operation.

jrandolf: This also provides better error handling since errors are kept internal to each object.

OrKoN: One could check if the context belongs, right? `if (this.id !== browsingContext.id) throw`.

OrKoN: I guess this is non-blocking, perhaps, Nik has an opinion

OrKoN (approved):

packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
jrandolf: Sure, the check can happen, but why check something when the operation is independent of the user context

OrKoN: In the spec, the browsing context is only a shortcut to determine the user context + source origin. So to match the spec structure, the client should be also delegating to the user context: otherwise, if new functionality is added that applied to both kinds of keys, the client needs to be modified in two places and the implementation has a potential to diverge.

PR #11866, opened 2024-02-07
refactor: small cleanup around realms and handles

jrandolf-google:


OrKoN (changes requested):
tests are failing

packages/puppeteer-core/src/bidi/Deserializer.ts
```diff
@@ -27,25 +27,30 @@ export class BidiDeserializer {
     }
   }
 
-  static deserializeLocalValue(result: Bidi.Script.RemoteValue): unknown {
+  static deserializeRemoteValue(result: Bidi.Script.RemoteValue): any {
```
Lightning00Blade:
```suggestion
  static deserialize(result: Bidi.Script.RemoteValue): any {
```

jrandolf: I changed it to deserializeRemoteValue to mirror Serializer.

packages/puppeteer-core/src/bidi/Realm.ts
```diff
@@ -203,10 +203,29 @@ export class BidiRealm extends EventEmitter<Record<EventType, any>> {
     }
 
     return returnByValue
-      ? BidiDeserializer.deserialize(result.result)
+      ? BidiDeserializer.deserializeRemoteValue(result.result)
       : createBidiHandle(sandbox, result.result);
   }
 
+  async destroyHandles(handles: Array<BidiJSHandle<unknown>>): Promise<void> {
```
Lightning00Blade, line 210: Nit: Let's not send command if handles is empty after filtering?

packages/puppeteer-core/src/bidi/Deserializer.ts
Lightning00Blade: I think we should keep simple names, as now we are duplicating the type in the function name without having another function. So we should change both places.

Lightning00Blade (approved):
LGTM

PR #11870, opened 2024-02-07
feat: support closing workers

### LINKED ISSUE #11775: [Feature]: Add Worker.close method to allow terminating service workers
### Feature description

For Chrome extension testing, it would be useful to be able to terminate service workers. It is possible to get the service worker associated with an extension as follows:

```
const target = await browser.waitForTarget((t) => {
  return t.type() === 'service_worker' && t.url().startsWith('chrome-extension://');
});
```

But it is not currently possible to close that target, meaning in some samples we have been working on we are instead using CDP commands: https://github.com/GoogleChrome/chrome-extensions-samples/blob/0f4a2ee19b3591a2a82e8eb96be4e5ef88755e39/functional-samples/tutorial.terminate-sw/puppeteer/index.test.js#L47

Opening this to track a `Worker.close` method so we can do `target.worker().close`.

OrKoN:
Closes #11775

### DIFF
--- a/packages/puppeteer-core/src/cdp/WebWorker.ts
+++ b/packages/puppeteer-core/src/cdp/WebWorker.ts
@@ -7,6 +7,7 @@ import type {Protocol} from 'devtools-protocol';
 
 import type {CDPSession} from '../api/CDPSession.js';
 import type {Realm} from '../api/Realm.js';
+import {TargetType} from '../api/Target.js';
 import {WebWorker} from '../api/WebWorker.js';
 import {TimeoutSettings} from '../common/TimeoutSettings.js';
 import {debugError} from '../common/util.js';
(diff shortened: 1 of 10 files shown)


jrandolf-google (changes requested):
I'll review when I come back. Please do not merge.

packages/puppeteer-core/src/cdp/WebWorker.ts
```diff
@@ -79,4 +82,21 @@ export class CdpWebWorker extends WebWorker {
...
+      // For service_workers. TODO: perhaps we need to check the worker type
+      // directly.
+      await this.client.connection()?.send('Target.closeTarget', {
+        targetId: this.#id,
+      });
+      await this.client.connection()?.send('Target.detachFromTarget', {
+        sessionId: this.client.id(),
+      });
+    } catch {
+      await this.evaluate(() => {
```
jrandolf: This line will error for service workers. We should just close the target for all worker types.

jrandolf-google (changes requested):

packages/puppeteer-core/src/cdp/WebWorker.ts
OrKoN: Target.closeTarget is not supported for all worker types

jrandolf: Then can we try `this.evaluate` before doing target.close?

jrandolf-google (approved):

PR #11872, opened 2024-02-07
fix: deprecate `Page.prototype.target`

jrandolf-google:
All methods on the `Target` instance obtained by `Page.prototype.target` are already on the `Page` class, so this getter is redundant.

### DIFF
--- a/test/src/page.spec.ts
+++ b/test/src/page.spec.ts
@@ -583,14 +583,10 @@ describe('Page', function () {
         // 3. After that, remove the iframe.
         frame.remove();
       });
-      const popupTarget = page
-        .browserContext()
-        .targets()
-        .find(target => {
-          return target !== page.target();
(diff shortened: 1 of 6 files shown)


test/src/page.spec.ts
```diff
@@ -583,13 +583,9 @@ describe('Page', function () {
...
       });
-      const popupTarget = page
-        .browserContext()
-        .targets()
-        .find(target => {
-          return target !== page.target();
-        })!;
-      // 4. Connect to the popup and make sure it doesn't throw.
+      // 4. The target should always be the last one.
+      const popupTarget = page.browserContext().targets().at(-1)!;
```
OrKoN, line 587: let's assert that it is actually not the page target by accident

OrKoN (approved):

PR #11879, opened 2024-02-08
chore: fixup redirect in `bidi/core`

jrandolf-google:


packages/puppeteer-core/src/bidi/core/Request.ts
```diff
@@ -66,36 +66,41 @@ export class Request extends EventEmitter<{
...
     );
     sessionEmitter.on('network.beforeRequestSent', event => {
-      if (event.context !== this.#browsingContext.id) {
-        return;
-      }
-      if (event.request.request !== this.id) {
+      if (
+        event.context !== this.#browsingContext.id ||
+        event.request.request !== this.id ||
+        this.#redirect !== undefined
```
jrandolf, line 72: This check is technically not needed since a redirect implies all listeners will get teared down, but we still write it explicitly here. It means if the request already has a redirect, then it's probably a downstream redirect.

OrKoN (approved):

PR #11895, opened 2024-02-12
chore: update frame tests

jrandolf-google:
This PR updates the frame tests to utilize a new condition waiter for destroyed contexts. Previously, we assumed navigation would destroy subframe contexts but this assumption does not always hold (e.g. the browser may destroy contexts after navigation to optimize the happy path).

### DIFF
--- a/test/src/frame.spec.ts
+++ b/test/src/frame.spec.ts
@@ -7,6 +7,7 @@
 import expect from 'expect';
 import {CDPSession} from 'puppeteer-core/internal/api/CDPSession.js';
 import type {Frame} from 'puppeteer-core/internal/api/Frame.js';
+import type {Page} from 'puppeteer-core/internal/api/Page.js';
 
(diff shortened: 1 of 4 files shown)


test/src/frame.spec.ts
```diff
@@ -138,72 +168,102 @@ describe('Frame specs', function () {
...
+        () => {
+          return true;
+        },
+        200
+      ).catch(() => {
+        return false;
       });
-      page.on('framedetached', () => {
-        return (hasEvents = true);
+      const frameDetached = waitForCondition(
```
OrKoN, line 181: I am not sure if we need to change the tests here. We test Puppeteer's logic and the expectation is that Puppeteer will emit detached events on navigation: see #removeFramesRecursively in #onFrameNavigated.

jrandolf: As discussed offline, we will investigate whether we try to detect this condition another way.

References:

 - https://github.com/w3c/webdriver-bidi/issues/659

PR #11898, opened 2024-02-12
refactor: use internal event emitter for trusted events

jrandolf-google:
Currently Puppeteer uses its own events to handle business logic. This can cause problems if a malicious library user uses `removeAllListeners` on Puppeteer objects.

This PR fixes this by creating a separate event emitter on Puppeteer event objects that bubbles into its host. This allows us to distinguish between trusted and untrusted events and separate userland handlers from internal ones.


OrKoN (changes requested):
Let's add a test to decorators.test.ts for the new decorator

OrKoN (approved):

PR #11909, opened 2024-02-13
chore: implement BiDi workers

jrandolf-google:


test/TestExpectations.json
```diff
@@ -327,7 +327,7 @@
     "testIdPattern": "[worker.spec] *",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["webDriverBiDi"],
-    "expectations": ["SKIP"]
+    "expectations": ["PASS"]
```
OrKoN, line 330: You need to skip it for Firefox, I believe.

packages/puppeteer-core/src/api/Target.ts
```diff
@@ -18,6 +18,7 @@ export enum TargetType {
   BACKGROUND_PAGE = 'background_page',
   SERVICE_WORKER = 'service_worker',
   SHARED_WORKER = 'shared_worker',
+  WORKER = 'worker',
```
OrKoN: I think we need to either start supporting this type for CDP targets (which could be a breaking change) or declare the target type as Other for BiDi workers for now.

packages/puppeteer-core/src/bidi/BrowserContext.ts
```diff
@@ -111,13 +117,30 @@ export class BidiBrowserContext extends BrowserContext {
...
+      this.trustedEmitter.emit(BrowserContextEvent.TargetCreated, target);
+    });
+    page.trustedEmitter.on(PageEvent.WorkerDestroyed, worker => {
+      const bidiWorker = worker as BidiWebWorker;
+      const target = pageTargets.get(bidiWorker);
       if (target === undefined) {
         return;
       }
-      frameTargets.delete(bidiFrame);
+      pageTargets.delete(worker);
```
jrandolf, line 140:
```suggestion
      pageTargets.delete(bidiWorker);
```

OrKoN (approved):

PR #11914, opened 2024-02-13
chore: `waitForNavigation` should fail if navigation aborts or fails

jrandolf-google:


packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -302,7 +302,19 @@ export class BidiFrame extends Frame {
         fromEmitterEvent(this.browsingContext, 'navigation').pipe(
           switchMap(({navigation}) => {
             return this.#waitForLoad$(options).pipe(
-              raceWith(fromEmitterEvent(navigation, 'fragment')),
+              raceWith(
```
OrKoN, line 312: Could you update the expectations?

jrandolf: There are no expectations to update. Your previous PR worked around this fix. I suggest we merge this and once Firefox supports errors, we remove interactive.

OrKoN (approved):

PR #11919, opened 2024-02-14
chore: wait for frames to be destroyed

jrandolf-google:


packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -267,11 +283,7 @@ export class BidiFrame extends Frame {
       this.waitForNavigation(options),
       this.browsingContext.navigate(
         url,
-        // Some implementations currently only report errors when the
-        // readiness=interactive. This also ensures that old frames have been
-        // removed.
-        // Related: https://bugzilla.mozilla.org/show_bug.cgi?id=1846601
-        Bidi.BrowsingContext.ReadinessState.Interactive
+        Bidi.BrowsingContext.ReadinessState.None
```
OrKoN: This is not only about frames but also about error reporting.

jrandolf: I've updated the comment.

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -258,19 +260,33 @@ export class BidiFrame extends Frame {
     });
   }
 
+  #detached$() {
+    if (this.detached) {
+      return of(this as Frame);
+    }
+    return fromEmitterEvent(
+      this.page().trustedEmitter,
+      PageEvent.FrameDetached
```
OrKoN: should this also wait for its own child frames?

jrandolf: All the child frames should be destroyed at this point.

OrKoN (approved):

PR #11929, opened 2024-02-15
chore: implement improved `waitForNavigation`

jrandolf-google:
This PR implements a `waitForNavigation` that doesn't validate the expected flow and instead only cares about results.

 - It will wait for a response if it can.
 - It will wait for the first navigation (fragment or not).
 - It will wait for the completion condition (if available).
 - It will wait for other conditions when navigation is not fragment navigation.

This also modifies the `bidi/core` API to match the BiDi spec as defined.

### DIFF
--- a/packages/puppeteer-core/src/bidi/Frame.ts
+++ b/packages/puppeteer-core/src/bidi/Frame.ts
@@ -10,14 +10,18 @@ import type {Observable} from '../../third_party/rxjs/rxjs.js';
 import {
   combineLatest,
   defer,
-  delayWhen,
   filter,
   first,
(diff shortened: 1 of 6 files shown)


jrandolf:
@OrKoN This PR is expected to fail btw. That's because the navigationStarted event is incorrect. I'm working on a fix in Chromium BiDi.

OrKoN:
@jrandolf thanks, after a quick review, it looks good. I will give it another review once it is ready to be landed.

Lightning00Blade:
Close in favor of https://github.com/puppeteer/puppeteer/pull/12776

PR #11930, opened 2024-02-16
test: do not use quirks mode for test files

### LINKED ISSUE #11926: Don't use Quirks mode for HTML test pages
In Quirks mode the browser uses older, less strict rendering rules to render the page as it would have appeared in older versions of the browser. This is actually causing quite a few differences in rendering not only for Firefox but also Chrome.

Not using that mode has an effect for a couple of Puppeteer unit tests, eg. `[elementhandle.spec] ElementHandle specs ElementHandle.boundingBox should handle nested frames` would pass with `249px` for the width in Chrome and Firefox.

CC @OrKoN.

OrKoN:
Closes https://github.com/puppeteer/puppeteer/issues/11926

### DIFF
--- a/test/assets/abort-request.html
+++ b/test/assets/abort-request.html
@@ -1,3 +1,4 @@
+<!DOCTYPE html>
 <button id="abort"></button>
 
 <script>
--- a/test/assets/beforeunload.html
+++ b/test/assets/beforeunload.html
@@ -1,3 +1,4 @@
+<!DOCTYPE html>
 <div>beforeunload demo.</div>
 <script>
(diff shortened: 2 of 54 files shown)


Lightning00Blade (approved):
LGTM + 
Q: I am curious if (with WebDriverBiDi in particular) it would be better to define the whole html rather then letting the browser inject the other parts (`head`,`body`, etc) as I am don't think it's standardized and may show browser differences.

OrKoN:
@Lightning00Blade I think that part is standard

PR #11932, opened 2024-02-16
test: remove redundant expectations

OrKoN:


tools/sort-test-expectations.mjs
```diff
@@ -95,13 +96,19 @@ for (let i = testExpectations.length - 1; i >= 0; i--) {
...
       if (isSubset(candidateLabels, labels)) {
         console.log('removing', expectation, 'already covered by', candidate);
         toBeRemoved.add(expectation);
       }
       break;
     }
   }
+
+  if (!foundMatch && isSubset(new Set(['PASS']), labels)) {
+    console.log('removing', expectation, 'because the default expectation is to pass');
```
Lightning00Blade: Nit:
```suggestion
    console.log(`Removing "${expectation}" because the default expectation is to pass`);
```

Lightning00Blade (approved):

tools/sort-test-expectations.mjs
OrKoN: that requires JSON.stringify and console.log prints it out nicely

Lightning00Blade:
@OrKoN Let's run the `full-ci` for this change as well.

PR #11947, opened 2024-02-19
feat: support local aliases when launching a browser

### LINKED ISSUE #11140: [FR]: Support npx @puppeteer/browsers install chrome@dev and improve documentation
### Steps to reproduce

```
npx @puppeteer/browsers install chrome@dev
npx @puppeteer/browser launch chrome@dev
```

### Expected results

Launch the downloaded Chrome for Testing Dev binary: `chrome/mac_arm-120.0.6051.2`, resolved to the same version used by the `install` command.

### Actual results

Puppeteer tries to launch the following binary:

```
chrome/mac_arm-dev/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing
```

Note the `chrome/mac_arm-dev`, instead of `chrome/mac_arm-120.0.6051.2` (the `dev` version at the time this issue was filed).

### LINKED ISSUE #11781: [Bug]: Executable path is not properly generated when using `PUPPETEER_BROWSER_REVISION=stable`

```TypeScript
// Set `PUPPETEER_BROWSER_REVISION` to `stable`
// export PUPPETEER_BROWSER_REVISION=stable
// Run `npm install puppeteer --save`

(async () => {
    try {
        await page.goto('https://google.com')
        await browser.close()
    } catch (error) {
        console.log(error)
    }
    await browser.close();
})();
```

### Error string

The executablePath uses `stable` while the path contains the actual browser revision.

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

_No response_

### Expectation

The Executable Path should be generated using the actual installed browser revision.

### Reality

The executable path tries to get the `stable` revision and will never exist.

I was able to temporarily fix the problem using this function:
```ts
async function getExecutablePath() {
  const browsers = await getInstalledBrowsers({
    cacheDir: process.env.PUPPETEER_CACHE_DIR || '',
  });

  if (browsers.length === 0) {
    throw new Error('Could not find a browser to launch');
  }

  const browser = browsers[browsers.length - 1];
  return browser.executablePath;
}
```

_No response_

### Puppeteer version

21.10.0

### Node version

20.11.0

### Package manager

npm

### Package manager version

10.2.4

### Operating system

Windows

OrKoN:
With this PR, @puppeteer/browsers starts maintaining a local mapping from an alias to the locally installed browser version. This allows using aliases such as `canary`/`dev` when launching the browser. Note that `latest` for launch will resolve to the latest version found in the local metadata file.

Note that the launch command would only consult metadata files and won't attempt to re-check aliases by contacting the server.

```
npx @puppeteer/browsers install chrome@canary
# Launching a Canary version would work if a Canary build has been installed previously.
npx @puppeteer/browsers launch chrome@canary 
```

Closes #11140
Closes #11781


packages/browsers/src/install.ts
```diff
@@ -233,17 +240,24 @@ async function installUrl(
...
+    const installedBrowser = new InstalledBrowser(
+      cache,
+      options.browser,
+      options.buildId,
+      options.platform
+    );
+    if (options.buildIdAlias) {
+      const metadata = installedBrowser.readMetadata();
+      metadata.aliases[options.buildIdAlias] = options.buildId;
+      metadata.aliases['latest'] = options.buildId;
```
Lightning00Blade: Why do we always update the `latest` ? This feels wrong, as if I install `latest` then install `stable`. The last install will be latest will be wrong.

OrKoN: The logic is that this is the latest version that you installed

OrKoN:
```suggestion
```

Lightning00Blade:
```suggestion
```

Lightning00Blade (commented):
LGTM, after suggestion is applied.

Lightning00Blade (approved):

PR #11954, opened 2024-02-20
feat: allow comparing browser versions

OrKoN:


packages/browsers/src/browser-data/firefox.ts
```diff
@@ -328,3 +328,8 @@ async function writePreferences(options: ProfileOptions): Promise<void> {
     await fs.promises.copyFile(prefsPath, prefsBackupPath);
   }
 }
+
+export function compareVersions(a: string, b: string): number {
+  // TODO: this is a not very reliable check.
+  return parseInt(a, 10) - parseInt(b, 10);
```
Lightning00Blade: Can we do something like this suggestion, I think it would be more accurate, what do you thing? 
This will handle the letters in the version.
```suggestion
  return parseInt(a.replace('.', '')), 16) - parseInt(b.replace('.', ''), 16);
```

Lightning00Blade (approved):

PR #11976, opened 2024-02-22
chore: refactor exposed function for BiDi

jrandolf-google:
Using one channel instead of three.

Bug: https://github.com/puppeteer/puppeteer/issues/11981

### DIFF
--- a/packages/puppeteer-core/src/bidi/ExposedFunction.ts
+++ b/packages/puppeteer-core/src/bidi/ExposedFunction.ts
@@ -6,29 +6,24 @@
 
 import * as Bidi from 'chromium-bidi/lib/cjs/protocol/protocol.js';
 
+import {EventEmitter} from '../common/EventEmitter.js';
 import type {Awaitable, FlattenHandle} from '../common/types.js';
 import {debugError} from '../common/util.js';
-import {assert} from '../util/assert.js';
-import {Deferred} from '../util/Deferred.js';
+import {DisposableStack} from '../util/disposable.js';
 import {interpolateFunction, stringifyFunction} from '../util/Function.js';
 
 import type {Connection} from './core/Connection.js';
-import {BidiDeserializer} from './Deserializer.js';
+import {BidiElementHandle} from './ElementHandle.js';
 import type {BidiFrame} from './Frame.js';
-import {BidiSerializer} from './Serializer.js';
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/bidi/ExposedFunction.ts
```diff
@@ -106,190 +85,144 @@ export class ExposeableFunction<Args extends unknown[], Ret> {
...
     void this[Symbol.asyncDispose]().catch(debugError);
   }
 
   async [Symbol.asyncDispose](): Promise<void> {
-    if (this.#preloadScriptId) {
-      await this.#connection.send('script.removePreloadScript', {
-        script: this.#preloadScriptId,
-      });
+    for (const [frame, script] of this.#scripts) {
+      await frame.browsingContext.removePreloadScript(script);
```
OrKoN: let's do promise all to dispose it quicker

packages/puppeteer-core/src/bidi/ExposedFunction.ts
```diff
@@ -38,65 +32,50 @@ export class ExposeableFunction<Args extends unknown[], Ret> {
...
-    const channelArguments = this.#channelArguments;
+    const channel = {
+      type: 'channel' as const,
+      value: {
+        channel: this.#channel,
+        ownership: Bidi.Script.ResultOwnership.Root,
+      },
+    };
 
     // TODO(jrandolf): Implement cleanup with removePreloadScript.
```
OrKoN: implemented now?

jrandolf: This is for `removeExposeFunction`

OrKoN (approved):

packages/puppeteer-core/src/bidi/ExposedFunction.ts
jrandolf: I've removed the comment. Linking the bug here instead.

OrKoN:
Looks like some test fail in Firefox

jrandolf:
> Looks like some test fail in Firefox

Looks like that test wasn't testing what it needed. We don't guarantee that exposed function calls in a preload script will resolve before navigation finishes. I've fixed the test to actually reflect what its title states.

OrKoN (approved):

PR #11986, opened 2024-02-22
chore: update devcontainer.json to Node 18

connorjclark:
This container set me up with Node 16

Based on https://github.com/devcontainers/features/tree/main/src/node, this should use 18.


OrKoN (approved):
Thanks!

PR #11999, opened 2024-02-26
feat: implement ARIA queries in BiDi

jrandolf-google:


OrKoN (commented):
Update TestExpectations?

test/TestExpectations.json
```diff
@@ -844,18 +820,6 @@
     "parameters": ["chrome", "webDriverBiDi"],
     "expectations": ["SKIP"]
   },
-  {
```
OrKoN, line 847: keep these two, they are caused by Chrome changes

OrKoN: we will fix them once CDP part rolls to that Chrome version

OrKoN (approved):

PR #12008, opened 2024-02-27
Update puppeteer.page.waitforframe.md

ymulenll:
Fix in docs.

No.

Yes.

**Summary**
Fix `waitForFrame` usage example in the documentation.

### DIFF
--- a/docs/api/puppeteer.page.waitforframe.md
+++ b/docs/api/puppeteer.page.waitforframe.md
@@ -32,6 +32,6 @@ Promise&lt;[Frame](./puppeteer.frame.md)&gt;
 
 ```ts
 const frame = await page.waitForFrame(async frame => {
-  return frame.name() === 'Test';
+  return (await frame.name()) === 'Test';
 });
 ```


docs/api/puppeteer.page.waitforframe.md
```diff
@@ -32,6 +32,6 @@ Promise&lt;[Frame](./puppeteer.frame.md)&gt;
 
 ```ts
 const frame = await page.waitForFrame(async frame => {
-  return frame.name() === 'Test';
+  return (await frame.name()) === 'Test';
```
OrKoN, line 35: frame.name() is a sync method and does not require an await

OrKoN (changes requested):

OrKoN:
Thanks for the PR but it looks like the proposed fix is not entirely correct, therefore, closing the PR.

docs/api/puppeteer.page.waitforframe.md
ymulenll: yes, I was using `frame.title()`, but I see `name` method is sync, thanks

PR #12009, opened 2024-02-27
chore: upgrade to npm-run-all2

43081j:
This upgrades from npm-run-all to npm-run-all2.

The major difference is that npm-run-all2 has security fixes, is actively maintained and has since removed various unnecessary dependencies (as you can see in the lockfile diff of this PR).

let me know if you have any questions. this is part of a larger effort to migrate a whole bunch of things around the ecosystem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -160,7 +160,7 @@
     "hereby": "1.8.9",
     "license-checker": "25.0.1",
     "mocha": "10.3.0",
-    "npm-run-all": "4.1.5",
+    "npm-run-all2": "6.1.2",
     "prettier": "3.2.5",
     "semver": "7.6.0",
     "sinon": "17.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -160,7 +160,7 @@
     "hereby": "1.8.9",
     "license-checker": "25.0.1",
     "mocha": "10.3.0",
-    "npm-run-all": "4.1.5",
+    "npm-run-all2": "^6.1.2",
```
OrKoN:
```suggestion
    "npm-run-all2": "6.1.2",
```

OrKoN: Would you mind removing the semrange and updating the lock file if needed?

OrKoN (commented):
Thanks for the PR!

package.json
43081j: sure no worries 👍

OrKoN (approved):

PR #12014, opened 2024-02-28
chore: for Firefox enable only the required protocol (CDP or BiDi)

whimboo:
At some point we want to deprecate CDP and not enable it by default anymore. That means that CDP clients have to explicitly set the preference `[remote.active-protocols](https://searchfox.org/mozilla-central/rev/d87ac4f189d6c1ad068bc3d1cdf50d2f871028c2/modules/libpref/init/all.js#3826-3830)` to `2`. For WebDriver BiDi it will be `1`.

This will also reduce conflicts between CDP and BiDi because at the moment both protocols are active and especially for CDP we need special settings for preferences which then also can negatively affect BiDi. The related bug for us is https://bugzilla.mozilla.org/show_bug.cgi?id=1882085.

CC @OrKoN

### DIFF
--- a/packages/puppeteer-core/src/node/FirefoxLauncher.ts
+++ b/packages/puppeteer-core/src/node/FirefoxLauncher.ts
@@ -43,12 +43,17 @@ export class FirefoxLauncher extends ProductLauncher {
     return {
       ...extraPrefsFirefox,
       ...(protocol === 'webDriverBiDi'
-        ? {}
+        ? {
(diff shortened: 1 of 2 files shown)


whimboo:
The test failure for BiDi was related to the stderr connection details which are different for the WebDriver BiDi protocol. Instead of `DevTools is listening` it is `WebDriver BiDi is listening`. With the recent update I've added a condition for Firefox to pass the failing test.

OrKoN (approved):

PR #12016, opened 2024-02-28
chore: set "network.cookie.cookieBehavior" preference to "0" only for CDP

lutien:
We would like to not set this for BiDi, so this PR moves the preference to CDP only preferences.


whimboo (approved):
Looks good to me.

OrKoN (approved):

Issue #12004, opened 2024-02-27
Potentially use `userContext` in bidi cookies commands instead of browsing context to avoid browser implementation differences


lutien:
### Feature description

The Puppeteer implementation of cookies APIs with WebDriverBiDi uses mostly `context` type in `partition` field. From what I've understood in the chromium-bidi mapper it’s implemented as `userContext` id associated with this browsing context, when Firefox implements it as a combination of `userContext` id and `sourceOrigin`, that makes a bunch of test fail in Firefox (see the list below). One solution could be to directly use `userContext` of the browsing context in the partition field with type `storageKey` when sending one of the cookie commands, that should eliminate the differences in the implementations. 

List of affected tests:
- Cookie specs Page.setCookie should be able to set insecure cookie for HTTP website
- Cookie specs Page.setCookie should default to setting secure cookie for HTTPS websites
- Cookie specs Page.setCookie should set a cookie on a different domain
- Cookie specs Page.setCookie should set a cookie with a path
- Cookie specs Page.setCookie should set multiple cookies
- Cookie specs Page.setCookie should set secure same-site cookies from a frame
- Cookie specs Page.setCookie should work
- DefaultBrowserContext page.setCookie() should work

OrKoN:
cc @sadym-chromium

OrKoN:
@lutien what does it fail with? different attributes are reported back I assume? in that case, I think we need to adjust expectations in tests.

OrKoN:
I am not sure using the userContext only would eliminate the differences because AFAIK Firefox requires a sourceOrigin to actually apply the cookie to the site?

lutien:
> @lutien what does it fail with? different attributes are reported back I assume? in that case, I think we need to adjust expectations in tests.

It's not the attributes, but the whole cookies are missing. In some cases, I think it's missing cookies in `document.cookie`

> I am not sure using the userContext only would eliminate the differences because AFAIK Firefox requires a sourceOrigin to actually apply the cookie to the site?
No, it should be fine, you can set with `userContext` and with `domain` in the cookie object (which is a required field anyway).

OrKoN:
Should not it be present in document.cookie if sourceOrigin and userContext are both correct?

lutien:
> Should not it be present in document.cookie if sourceOrigin and userContext are both correct?

[...]

PR #12017, opened 2024-02-28
feat: in BiDi cookies default to using `userContext`

sadym-chromium:
https://github.com/puppeteer/puppeteer/issues/12004

### DIFF
--- a/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
+++ b/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
@@ -503,8 +503,8 @@ export class BrowsingContext extends EventEmitter<{
     } = await this.#session.send('storage.getCookies', {
       ...options,
       partition: {
-        type: 'context',
-        context: this.id,
+        type: 'storageKey',
+        userContext: this.userContext.id,
       },
     });
     return cookies;
@@ -518,8 +518,8 @@ export class BrowsingContext extends EventEmitter<{
     await this.#session.send('storage.setCookie', {
       cookie,
       partition: {
-        type: 'context',
-        context: this.id,
+        type: 'storageKey',
+        userContext: this.userContext.id,
       },
     });
   }


packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
```diff
@@ -503,8 +503,8 @@ export class BrowsingContext extends EventEmitter<{
     } = await this.#session.send('storage.getCookies', {
       ...options,
       partition: {
-        type: 'context',
-        context: this.id,
+        type: 'storageKey',
```
OrKoN, line 506: would not it fetch all cookies in the user context instead of only cookies related to the current browsing context?

OrKoN: Would it include partitioned cookies set for Chrome?

OrKoN:
Probably not a `feat: `, perhaps `fix: `? Still not sure if it works as expected.

packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
sadym-chromium: > would not it fetch all cookies in the user context instead of only cookies related to the current browsing context?

It would, and it is consistent with the current behavior of BiDi cookies.

OrKoN: It does not look like the expected behavior from the Puppeteer perspective? Do we have a test for partitioned cookies?

I think if this is the intention, I suggest we instead use the method on UserContext in Puppeteer instead of changing calls here.

sadym-chromium: I'm not sure what "related to the current browsing context" means. Isn't it something like https://github.com/w3c/webdriver-bidi/issues/645, which we decided not to implement?

OrKoN: This class is called packages/puppeteer-core/src/bidi/core/BrowsingContext.ts it's related to a browsing context instance. If you need to call UserContext methods, see the UserContext.getCookies/setCookie

PR #12018, opened 2024-02-28
fix(webdriver): wait for response if the response has not completed once navigation has finished

OrKoN:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1882454

The windows failure is unrelated https://github.com/puppeteer/puppeteer/issues/12021

### DIFF
--- a/packages/puppeteer-core/src/bidi/Frame.ts
+++ b/packages/puppeteer-core/src/bidi/Frame.ts
@@ -40,6 +40,8 @@ import {debugError, fromEmitterEvent, timeout} from '../common/util.js';
 
 import {BidiCdpSession} from './CDPSession.js';
 import type {BrowsingContext} from './core/BrowsingContext.js';
+import type {Navigation} from './core/Navigation.js';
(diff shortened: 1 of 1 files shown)


packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -360,8 +362,34 @@ export class BidiFrame extends Frame {
...
+                  ): Observable<Navigation> {
+                    if (request.response || request.error) {
+                      return of(navigation);
+                    }
+                    if (request.redirect) {
+                      return requestFinished$(request.redirect);
+                    }
+                    // Reduces flakiness if the response events arrive after
+                    // the load event.
+                    // Usually, the response or error is already there at this point.
```
Lightning00Blade: I think this comment is better suited for the `if (request.response || request.error) {` as here it implies that the `success`,`error` events most of the time are not emitted.

Lightning00Blade (approved):
LGTM + comment

PR #12026, opened 2024-02-29
Revert "chore: for Firefox enable only the required protocol (CDP or BiDi)"

OrKoN:
Reverts puppeteer/puppeteer#12014


Lightning00Blade (approved):
LGTM, in case it needs to be merged.

PR #12031, opened 2024-02-29
feat(webdriver): support `Page.deleteCookie()` for WebDriver BiDi

sadym-chromium:
Implement `Page.deleteCookie()` via BiDi `storage.deleteCookies`.

### DIFF
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -571,8 +571,26 @@ export class BidiPage extends Page {
     }
   }
 
-  override deleteCookie(...cookies: DeleteCookiesRequest[]): Promise<void> {
-    return this.#browserContext.userContext.deleteCookie(...cookies);
+  override async deleteCookie(...cookies: DeleteCookiesRequest[]): Promise<void> {
+    await Promise.all(cookies.map(async deleteCookieRequest => {
+      const cookieUrl = deleteCookieRequest.url ?? this.url();
+      const normalizedUrl = URL.canParse(cookieUrl)
+        ? new URL(cookieUrl)
+        : undefined;
+
+      const domain = deleteCookieRequest.domain ?? normalizedUrl?.hostname;
+      assert(
+        domain !== undefined,
+        `At least one of the url and domain needs to be specified`
+      );
+
+      const filter = {
+        domain: domain,
+        name: deleteCookieRequest.name,
+        ...(deleteCookieRequest.path !== undefined ? { path: deleteCookieRequest.path } : {})
+      }
+      await this.#browserContext.userContext.deleteCookie(filter);
+    }));
   }
 
   override async removeExposedFunction(name: string): Promise<void> {
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -589,7 +589,7 @@ export class BidiPage extends Page {
         name: deleteCookieRequest.name,
         ...(deleteCookieRequest.path !== undefined ? { path: deleteCookieRequest.path } : {})
       }
-      await this.#browserContext.userContext.deleteCookie(filter);
+      await this.#frame.browsingContext.deleteCookie(filter);
     }));
   }
 
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -49,6 +49,7 @@ import {BidiKeyboard, BidiMouse, BidiTouchscreen} from './Input.js';
 import type {BidiJSHandle} from './JSHandle.js';
 import {rewriteNavigationError} from './util.js';
 import type {BidiWebWorker} from './WebWorker.js';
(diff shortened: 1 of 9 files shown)


OrKoN:
> Should we extend tests verifying this?

Let's add a test to also understand the cdp behavior better.

packages/puppeteer-core/src/bidi/core/UserContext.ts
```diff
@@ -196,6 +197,26 @@ export class UserContext extends EventEmitter<{
...
+  })
+  async deleteCookie(...cookies: DeleteCookiesRequest[]): Promise<void> {
+    await Promise.all(cookies.map(async deleteCookieRequest => {
+      const filter = {
+        ...deleteCookieRequest
+      };
+      await this.#session.send('storage.deleteCookies', {
+        filter: filter,
+        partition: {
+          type: 'storageKey',
```
OrKoN: should we delete using the type=context instead?

OrKoN: moving it to the BrowseringContext.ts correspondingly

OrKoN (commented):
Probably need to enable some tests via TestExpectations

packages/puppeteer-core/src/bidi/core/UserContext.ts
sadym-chromium: This depends on the desirable observable differences between firefox and chrome. Let me add a test for it

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -559,9 +560,6 @@ export class BidiPage extends Page {
         ),
       };
 
-      // TODO: delete cookie before setting them.
-      // await this.deleteCookie(bidiCookie);
-
```
sadym-chromium, line 564: The scope of set and deleted cookie can be different, so deleting can cause some unwilling side effects. Example: a cookie is set for a specific path, but removed from the whole domain.

test/src/cookies.spec.ts
```diff
@@ -584,5 +584,139 @@ describe('Cookie specs', () => {
...
+          size: 33,
+          httpOnly: false,
+          secure: false,
+          session: true,
+          sourceScheme: 'NonSecure',
+        },
+      ]);
+    });
+    it('should delete cookie for specified URL regardless of the current page', async () => {
+      // Tentative test case. Depending on the cookie partitioning implementation, this
```
sadym-chromium: This is an edge case where firefox and chrome can have observable different behavior. In ff all the cookies are partitioned, while in chrome they are not. In ff if context partition is used, delete command scoped to the cookies set by the current page. In chrome it is scoped only for the user context. As discussed offline, this is an expected to have different behavior for different implementations.

packages/puppeteer-core/src/bidi/core/UserContext.ts
sadym-chromium: Switched to `BrowsingContextPartitionDescriptor` and added tests.

test/TestExpectations.json
```diff
@@ -1080,6 +1080,54 @@
     "parameters": ["cdp", "firefox"],
     "expectations": ["FAIL"]
   },
+  {
```
OrKoN, line 1090: Could you please add a `"comment":""` field with a comment about the reason for the fail expectation?

test/TestExpectations.json
```diff
@@ -1080,6 +1080,54 @@
...
+    "testIdPattern": "[cookies.spec] Cookie specs Page.deleteCookie should delete cookie",
+    "platforms": ["linux"],
```
```suggestion
    "platforms": ["linux"],
```

all platforms?

test/TestExpectations.json
```diff
@@ -1080,6 +1080,54 @@
...
+    "testIdPattern": "[cookies.spec] Cookie specs Page.deleteCookie should delete cookie for specified URL",
+    "platforms": ["linux"],
+    "parameters": ["cdp", "firefox"],
+    "expectations": ["FAIL"]
+  },
+  {
+    "testIdPattern": "[cookies.spec] Cookie specs Page.deleteCookie should delete cookie for specified URL regardless of the current page",
+    "platforms": ["linux"],
+    "parameters": ["firefox", "webDriverBiDi"],
+    "expectations": ["SKIP"]
```
OrKoN: I assume the skipped once are the ones that are due to different implementation in partitioning? (let's leave a comment)

test/TestExpectations.json
sadym-chromium: done

test/TestExpectations.json
sadym-chromium: indeed, done

test/TestExpectations.json
sadym-chromium: done

test/TestExpectations.json
OrKoN: Sorry, I was not clear on this. I meant to ask for a comment for each of the expectations that are new

OrKoN (approved):

PR #12037, opened 2024-03-04
chore: enable test for ff

Lightning00Blade:


test/TestExpectations.json
```diff
@@ -50,7 +50,7 @@
   {
     "testIdPattern": "[headful.spec] *",
     "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["firefox"],
+    "parameters": ["cdp", "firefox"],
     "expectations": ["SKIP"]
```
OrKoN: Let's add a comment why it fails with Firefox CDP. Is it cookies?

OrKoN (approved):

whimboo:
@Lightning00Blade FYI this test is failing if the tests are run on a system in headless mode and not having libraries installed to run headful. I think that we should skip the test in case headless mode is selected?

PR #12041, opened 2024-03-04
chore: update angular deps

Lightning00Blade:
Closes #12030


OrKoN (approved):
Btw, do we want perhaps make ng-schematics depend on semver ranges?

PR #12051, opened 2024-03-06
feat: allow downloading Firefox channels other than nightly

### LINKED ISSUE #10420: [Feature]: allow `browsers` to download Firefox Stable/Aurora/ESR in addition to Nightly
### Feature description

In [this PR](https://github.com/webrtcHacks/adapter/pull/1133) I am trying to remove the dependency on [travis-multirunner](https://github.com/DamonOehlman/travis-multirunner/) but would like to retain the ability to download different FF versions (stable, aurora, esr, maybe dev).

It seems the main obstacle is that the directory structure for FF Nightly is
`https://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-central/{buildId}.{locale}.{platform-specific-stuff}` ([example here](https://archive.mozilla.org/pub/firefox/nightly/latest-mozilla-central/))
while for the other channels the platform and locale come earlier in the path:
`https://archive.mozilla.org/pub/firefox/releases/{buildId}/{platform}/{locale}/<binary>`
[example here](https://archive.mozilla.org/pub/firefox/releases/114.0.1/win64/en-US/)

Would monkeypatching [resolveDownloadUrl](https://github.com/puppeteer/puppeteer/blob/main/packages/browsers/src/browser-data/firefox.ts#L42) to recognize the default nightly baseUrl and call resolveDownloadPath/archive with a different boolean flag be an ok approach?
This would allow "users" to pass a different baseUrl depending on the channel.
The cleaner alternative would be to plumb the channel through (for other browsers as well)

Yet another approach would be the one that [travis-multirunner](https://github.com/nathanoehlman/browser-sleuth/blob/master/lib/firefox.js#LL26C9-L26C120) has taken for a decade and download from 
`'https://download.mozilla.org/?os=' + platform + '&lang=' + lang + '&product=firefox-' + (channel || 'latest');` but I would like less travis-multirunner, not more ;-)

Issue #6894 has some context on switching the default and helpful pointers to download urls

OrKoN:
This PR adds support for the following commands:

```
npx @puppeteer/browsers install firefox@stable --path /tmp/test
npx @puppeteer/browsers install firefox@devedition --path /tmp/test
npx @puppeteer/browsers install firefox@beta --path /tmp/test
npx @puppeteer/browsers install firefox@esr --path /tmp/test
npx @puppeteer/browsers install firefox@nightly --path /tmp/test
```

To distinguish between different channels with the same version, the channel name is prepended to the `buildId` used by Puppeteer. Unprefixed versions are all considered to be nightly versions.

Closes #10420

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -222,7 +222,31 @@ export class CLI {
           );
           yargs.example(
             '$0 install firefox',
-            'Install the latest available build of the Firefox browser.'
+            'Install the latest nightly available build of the Firefox browser.'
+          );
+          yargs.example(
+            '$0 install firefox@stable',
+            'Install the latest stable build of the Firefox browser.'
+          );
+          yargs.example(
+            '$0 install firefox@beta',
+            'Install the latest beta build of the Firefox browser.'
+          );
+          yargs.example(
+            '$0 install firefox@devedition',
+            'Install the latest devedition build of the Firefox browser.'
+          );
+          yargs.example(
+            '$0 install firefox@esr',
+            'Install the latest ESR build of the Firefox browser.'
+          );
+          yargs.example(
+            '$0 install firefox@nightly',
+            'Install the latest nightly build of the Firefox browser.'
+          );
+          yargs.example(
+            '$0 install firefox@stable_111.0.1',
+            'Install a specific version of the Firefox browser.'
           );
           yargs.example(
             '$0 install firefox --platform mac',
@@ -395,7 +419,7 @@ export function makeProgressCallback(
   return (downloadedBytes: number, totalBytes: number) => {
     if (!progressBar) {
       progressBar = new ProgressBar(
-        `Downloading ${browser} r${buildId} - ${toMegabytes(
+        `Downloading ${browser} ${buildId} - ${toMegabytes(
           totalBytes
         )} [:bar] :percent :etas `,
         {
--- a/packages/browsers/test/src/chrome/install.spec.ts
+++ b/packages/browsers/test/src/chrome/install.spec.ts
@@ -138,6 +138,7 @@ describe('Chrome install', () => {
   });
 
   it('falls back to the chrome-for-testing dashboard URLs if URL is not available', async function () {
+    this.timeout(60000);
     const expectedOutputPath = path.join(
       tmpDir,
       'chrome',
--- a/packages/browsers/src/browser-data/browser-data.ts
+++ b/packages/browsers/src/browser-data/browser-data.ts
@@ -54,24 +54,22 @@ export const versionComparators = {
 export {Browser, BrowserPlatform, ChromeReleaseChannel};
 
 /**
- * @public
+ * @internal
  */
-export async function resolveBuildId(
+async function resolveBuildIdForBrowserTag(
   browser: Browser,
   platform: BrowserPlatform,
-  tag: string
+  tag: BrowserTag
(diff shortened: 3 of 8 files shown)


whimboo:
No, for stable releases we do not offer zip archives of any sort. But you can install Firefox silently by using methods similar to what we have in `mozinstall`:

https://searchfox.org/mozilla-central/source/testing/mozbase/mozinstall/mozinstall/mozinstall.py#339-400

Maybe the MSI solution is closer to what you want?

OrKoN:
@Lightning00Blade could you do the first pass at reviewing? I am thinking I need to add some examples to the CLI.

Lightning00Blade (approved):
LGTM + still missing some test for windows.

Lightning00Blade (approved):

packages/browsers/src/CLI.ts
```diff
@@ -222,7 +222,31 @@ export class CLI {
...
+            '$0 install firefox@beta',
+            'Install the latest beta build of the Firefox browser.'
+          );
+          yargs.example(
+            '$0 install firefox@devedition',
+            'Install the latest devedition build of the Firefox browser.'
+          );
+          yargs.example(
+            '$0 install firefox@esr',
+            'Install the latest ESR build of the Firefox browser.'
```
whimboo, line 241: Should we initially support ESR releases? You don't do that with Chrome (if you have such ones) and maybe we should align for now. It would certainly cause breakage due to not implemented features. Or will we have a support matrix of which version of puppeteer supports which ESR release?

packages/browsers/test/src/chrome/install.spec.ts
```diff
@@ -138,6 +138,7 @@ describe('Chrome install', () => {
   });
 
   it('falls back to the chrome-for-testing dashboard URLs if URL is not available', async function () {
+    this.timeout(60000);
```
whimboo, line 141: That addition is expected?

Lightning00Blade: Most likely yes, usually these these are slow as they do real request, and often fail on CI only because of timeouts.

packages/browsers/src/CLI.ts
Lightning00Blade: I think we should as this is the `@puppeteer/browsers` package and as long as a distribution of said browser we should have a way to download it. 
This PR does not integrate it with `puppeteer` so users will not see breakage. I think support matix for Firefox will become a must if we don't use `nightly` for downloads from `puppeteer`.

whimboo: Sounds fine then. I only wanted to give the heads-up that it might need further work on Puppeteer's side to exclude specific Puppeteer releases for older Firefox ESR versions.

whimboo (approved):

mathiasbynens (approved):
LGTM with the disclaimer that I’m not an expert on Firefox release channels vs. products.

(Intuitively “Dev Edition” sounds like a separate product rather than a release channel, but I don’t actually know which it is. As long as it can reasonably be considered as a channel I guess it’s fine to add it here.)

whimboo:
> (Intuitively “Dev Edition” sounds like a separate product rather than a release channel, but I don’t actually know which it is. As long as it can reasonably be considered as a channel I guess it’s fine to add it here.)

Dev Edition is like Beta but starts earlier after a version bump and has special settings for developers to eg. allow unsigned addons for testing.

jrandolf-google (approved):

PR #12060, opened 2024-03-08
test: improve `target.asPage` tests

kblok:
Some tests in the `devtools.spec.ts` file used `toContainEqual` to look for a page in `browser.pages()`.
The problem with `toContainEqual` is that it compares field by field. And jest is only finding two fields, `_isDragging` and `_timeoutSettings`, matching pages that are not technically equal.

Based on that, I change the `target.page() should return a DevTools page if asPage is used` test because `browser.pages()` won't return the devtools page without a custom `isPageTarget` function.

### DIFF
--- a/test/src/cdp/devtools.spec.ts
+++ b/test/src/cdp/devtools.spec.ts
@@ -69,7 +69,7 @@ describe('DevTools', function () {
     ).toBe(6);
     expect(await browser.pages()).toContain(page);
   });
-  it('target.page() should return a DevTools page if asPage is used', async function () {
+  it('target.page() should return Page when calling asPage on DevTools target', async function () {
     const {puppeteer} = await getTestState({skipLaunch: true});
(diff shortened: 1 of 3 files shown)


test/src/cdp/devtools.spec.ts
```diff
@@ -67,7 +67,7 @@ describe('DevTools', function () {
         return 2 * 3;
       })
     ).toBe(6);
-    expect(await browser.pages()).toContainEqual(page);
+    expect(await browser.pages()).toContain(page);
   });
   it('target.page() should return a DevTools page if asPage is used', async function () {
```
Lightning00Blade: Name should be closer to:
```suggestion
  it('target.page() should return Page when calling asPage on DevTools target', async function () {
```

kblok: Done! I also changed the expectations.

Lightning00Blade (approved):

PR #12065, opened 2024-03-09
remove repetitive words

pengqiseven:

docs

**Summary**

**Other information**


Lightning00Blade:
For us to merge this you will need to sign the CLA, also:
Our docs are generated by the TsDoc in the TypeScript files you will need to update those and run npm run docs to modify the generated docs.
Also you should not modify the versioned docs as they should remain as they are.

PR #12072, opened 2024-03-11
chore: update test expectation for "Page Page.bringToFront should work" which is now passing in Firefox

whimboo:
@Lightning00Blade this is now working correctly as of landing of https://bugzilla.mozilla.org/show_bug.cgi?id=1877469 by end of last week.


Lightning00Blade (approved):
LGTM, you may need to run `npm run format` to move the expectation to the correct place.

PR #12079, opened 2024-03-13
chore: update FF expectations

Lightning00Blade:
CC @whimboo


whimboo (approved):
Yes, we landed support for it yesterday.

CC @lutien

PR #12084, opened 2024-03-13
feat: deprecate `Frame.prototype.name`

### LINKED ISSUE #12083: [Feature]: deprecate `Frame.name()`
### Feature description

Deprecate [Frame.name()](https://pptr.dev/api/puppeteer.frame.name) method. Detailed discussion and motivation is [here](https://github.com/w3c/webdriver-bidi/issues/634).

TL;DR; this data can be outdated, and the up-to-date element name can be retrieved via script evaluation.

jrandolf-google:
`Frame.prototype.name` has a lot of problem associated with it. See https://github.com/puppeteer/puppeteer/issues/12083.

We resolve the use-case with the new `frameElement` getter which returns the frame element associated with a given frame.

Fixes https://github.com/puppeteer/puppeteer/issues/12083


sadym-chromium (approved):

sadym-chromium:
Shouldn't we mark relevant tests as deprecated as well?

Lightning00Blade (approved):
LGTM, but I think we should remove the uses from test as well, WDYT?

packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -405,7 +405,7 @@ export abstract class Frame extends EventEmitter<FrameEvents> {
   }
 
   /**
-   * @internal
+   * @returns The frame element associated with this frame (if any).
```
sadym-chromium, line 408: Please add tests for `frameElement()`, as it became a public API

test/src/launcher.spec.ts
```diff
@@ -777,7 +777,7 @@ describe('Launcher specs', function () {
           await close();
         }
       });
-      it('should be able to reconnect to a disconnected browser', async () => {
+      it.skip('should be able to reconnect to a disconnected browser', async () => {
```
Lightning00Blade, line 780: Why this change? Event if we skip test we do it via Expectations.

docs/api/puppeteer.frame.name.md
```diff
@@ -4,6 +4,15 @@ sidebar_label: Frame.name
 
 # Frame.name() method
 
+> Warning: This API is now obsolete.
+>
+> Use
+>
+> ```ts
+> const element = await frame.frameElement();
+> const name = await element.evaluate(frame => frame.name);
```
Lightning00Blade, line 13: Let's provide an example with the `id` as `frame.name()` would also return the Id of the element.

PR #12092, opened 2024-03-14
feat: await is only valid in async function  BREAKING CHANGE:   const…

tshivam1811:
… {downloadBrowser} = await importInstaller();

feat: await is only valid in async function
BREAKING CHANGE:   const {downloadBrowser} = await importInstaller();

**Summary**

**Other information**


Lightning00Blade:
This invalid change, the function is marked as `async`.

PR #187, opened 2017-08-02
Roll chromium

aslushnikov:
This revision includes @pavelfeldman upstream changes to make dialogs
work in headless.

References #13

PR #12095, opened 2024-03-15
test: verify closing last page does not close connection

sadym-chromium:
Add test to verify behavior consistency between browsers and protocols. 

Required for [Specify behavior for `browsingContext.close` when closing the last browsing context · Issue #187 · w3c/webdriver-bidi](https://github.com/w3c/webdriver-bidi/issues/187)

### DIFF
--- a/test/src/page.spec.ts
+++ b/test/src/page.spec.ts
@@ -18,7 +18,7 @@ import type {ConsoleMessage} from 'puppeteer-core/internal/common/ConsoleMessage
 import {Deferred} from 'puppeteer-core/internal/util/Deferred.js';
 import sinon from 'sinon';
 
-import {getTestState, setupTestBrowserHooks} from './mocha-utils.js';
+import {getTestState, launch, setupTestBrowserHooks} from './mocha-utils.js';
 import {attachFrame, detachFrame, isFavicon, waitEvent} from './utils.js';
 
 describe('Page', function () {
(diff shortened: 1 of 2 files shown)


test/src/page.spec.ts
```diff
@@ -111,6 +111,28 @@ describe('Page', function () {
...
+      const {browser, close} = await launch({}, {createContext: false});
+      try {
+        const pages = await browser.pages();
+        await Promise.all(
+          pages.map(page => {
+            return page.close();
+          })
+        );
+        // Verify the browser is still connected.
+        expect(browser.isConnected()).toBe(true);
```
Lightning00Blade:
```suggestion
        expect(browser.connected).toBe(true);
```

test/src/page.spec.ts
```diff
@@ -111,6 +111,28 @@ describe('Page', function () {
         expect(message).not.toContain('Timeout');
       }
     });
+    it('closing last tab should not close the browser', async () => {
```
Lightning00Blade: Let's put this test the `browser.spec.ts` as it's not testing anything on the Page itself.

test/src/page.spec.ts
```diff
@@ -111,6 +111,28 @@ describe('Page', function () {
...
+      // This test asserts the behavior is consistent between protocols and browsers.
+      // WebDriver Classic requires the session to be closed after the last page is
```
Lightning00Blade: Not sure if we need the `WebDriver Classic` comment as Puppeteer does not use it.

sadym-chromium: removed

test/src/page.spec.ts
sadym-chromium: done

Lightning00Blade (approved):

PR #12096, opened 2024-03-15
fix: don't keep connection alive after redirect

Lightning00Blade:
When we redirect the request we need to resume the initial response else it may lead to keeping the connection open forever.

### DIFF
--- a/packages/browsers/src/httpUtil.ts
+++ b/packages/browsers/src/httpUtil.ts
@@ -46,7 +46,12 @@ export function httpRequest(
     agent: new ProxyAgent(),
   };
 
-  const requestCallback = (res: http.IncomingMessage): void => {
+  const request =
+    options.protocol === 'https:'
+      ? https.request(options, requestCallback)
+      : http.request(options, requestCallback);
+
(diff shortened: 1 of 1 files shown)


packages/browsers/src/httpUtil.ts
```diff
@@ -46,22 +46,25 @@ export function httpRequest(
...
+  function requestCallback(res: http.IncomingMessage) {
     if (
       res.statusCode &&
       res.statusCode >= 300 &&
       res.statusCode < 400 &&
       res.headers.location
     ) {
       httpRequest(new URL(res.headers.location), method, response);
+      // Destroy the request to allow the connection to close
+      request.destroy();
```
jrandolf: Shouldn't we do `request.end`?

Lightning00Blade: No it should be neither. I think we need to do something to the response not the request

jrandolf-google (approved):

jrandolf:
Is it possible to add a test for this? The specific test scenario would help.

PR #12104, opened 2024-03-18
fix(webdriver): emit RequestServedFromCache for requests

Lightning00Blade:


packages/puppeteer-core/src/bidi/HTTPRequest.ts
```diff
@@ -68,7 +68,7 @@ export class BidiHTTPRequest extends HTTPRequest {
   }
 
   override resourceType(): ResourceType {
-    return this.initiator().type.toLowerCase() as ResourceType;
```
Lightning00Blade, line 71: This is totally wrong... We should see if we can add `ResourceType` to the spec. 
Both Firefox and Chrome support similar functionality in there respective DevTools panels.

OrKoN (approved):
LGTM % remove the fail expectation

Lightning00Blade:
@OrKoN Please rereview I need to fix the expectation formatter.

tools/sort-test-expectations.mjs
```diff
@@ -92,14 +92,17 @@ for (let i = testExpectations.length - 1; i >= 0; i--) {
       testIdMatchesExpectationPattern(
         expectation.testIdPattern,
         candidate.testIdPattern
-      ) &&
-      isSubset(candidateParams, params) &&
-      isSubset(candidatePlatforms, platforms)
+      )
     ) {
       foundMatch = true;
```
Lightning00Blade: We need to mark that there is a match even if the other parts don't match.
```json
  {
    "testIdPattern": "[network.spec] network Network Events Page.Events.RequestServedFromCache",
    "expectations": ["PASS"]
  },
  ```
  Was being remove due to 
  ```json
    {
    "testIdPattern": "[network.spec] network Network Events *",
    "expectations": ["FAIL"],
  },
  ```

**Review by OrKoN — APPROVED** [review]

File: tools/sort-test-expectations.mjs (hunk 2 above)
Comment (OrKoN) [reply tools/sort-test-expectations.mjs]: Aren't params needed for the match too?

File: tools/sort-test-expectations.mjs [hunk 3]
```diff
@@ -93,13 +93,14 @@ for (let i = testExpectations.length - 1; i >= 0; i--) {
...
       ) &&
-      isSubset(candidateParams, params) &&
       isSubset(candidatePlatforms, platforms)
     ) {
       foundMatch = true;
-      if (isSubset(candidateLabels, labels)) {
-        console.log('removing', expectation, 'already covered by', candidate);
-        toBeRemoved.add(expectation);
+      if (isSubset(candidateParams, params)) {
+        if (isSubset(candidateLabels, labels)) {
```
OrKoN: merge ifs?

tools/sort-test-expectations.mjs
Lightning00Blade: Yeah, I think the logic is wrong.
I think the `isSubset(candidateParams, params)` need to be reverse as we have an `AND` logic for Params

So previous expectation will have a smaller `Params` then the next one and still affect it.

Lightning00Blade:
@OrKoN after the last change this was the output - https://gist.github.com/Lightning00Blade/50916ff9d8926e5e3fb629ad4311a18a

tools/sort-test-expectations.mjs
OrKoN: I am not sure that is correct, reversing might remove a more specific expectations I think

log.log
```diff
@@ -0,0 +1,169 @@
+
```
OrKoN: remove

OrKoN (approved):

PR #12105, opened 2024-03-18
chore: fix interception tests

Lightning00Blade:


test/src/requestinterception-experimental.spec.ts
```diff
@@ -94,24 +94,26 @@ describe('cooperative request interception', function () {
...
+          request = req;
         }
-        expect(request.url()).toContain('empty.html');
-        expect(request.headers()['user-agent']).toBeTruthy();
-        expect(request.method()).toBe('GET');
-        expect(request.postData()).toBe(undefined);
-        expect(request.isNavigationRequest()).toBe(true);
-        expect(request.resourceType()).toBe('document');
-        expect(request.frame() === page.mainFrame()).toBe(true);
-        expect(request.frame()!.url()).toBe('about:blank');
```
Lightning00Blade, line 109: This expectation is now removed. Should we keep it in another form?

test/src/requestinterception.spec.ts
```diff
@@ -192,11 +199,13 @@ describe('request interception', function () {
 
       await page.setExtraHTTPHeaders({referer: server.EMPTY_PAGE});
       await page.setRequestInterception(true);
-      page.on('request', request => {
-        expect(request.headers()['referer']).toBe(server.EMPTY_PAGE);
+      let request!: HTTPRequest;
+      page.on('request', req => {
+        request = req;
```
OrKoN, line 218: I think it is important to check that the data is already there at the time of the callback. Checking later might hide the fact that the data is not available when the user expects it,

Lightning00Blade: Here is an example. And this passes:
```ts
   it.only('should intercept', async () => {

      let request!: HTTPRequest;
      page.on('request', req => {
        if (isFavicon(req)) {
          void req.continue();
          return;
        }
        request = req;
        void req.continue();
        expect(false).toBeTruthy();
      });
      const response = (await page.goto(server.EMPTY_PAGE))!;

      expect(request).toBeTruthy();
      expect(response.ok()).toBe(true);
      expect(response.remoteAddress().port).toBe(server.PORT);
```
I think we can wrap the error, and throw if it not `undefined`?

OrKoN: I meant that request headers should be asserted in the event callback and not after the whole navigation is done

Lightning00Blade: I understand what you are saying. But `expect` in the call back has no effect on the passing of the test. 
I have `expect(false).toBeTruthy();` in my example an the test still Passes.

OrKoN: Why does it have no effect? are we catching errors from events somewhere?

Lightning00Blade: Went though our implementation and `mitt` it never catches the error not sure why its not reported.

OrKoN: I think it is some mocha quirk, running standalone reports an unhandled promise rejection. Also, does Node not crash on that anymore?

Lightning00Blade: For node it becomes  - `unhandled-rejections` and we would need to set this in `.mocharc.js`:
```
'node-option': ['unhandled-rejections=strict'],
```
But then the whole run will fail.

OrKoN: Let's get the data during the event but assert it after the event? it would be easier than catching errors

Lightning00Blade: How about we wrap the expects in a `try/catch` and store it in an `error` then if its not empty throw?

Lightning00Blade: Done + added EsLint Rule

test/src/requestinterception-experimental.spec.ts
Lightning00Blade: Restored in the latest version

OrKoN (approved):

PR #12106, opened 2024-03-18
chore(webdriver): support for Network interception

Lightning00Blade:
There are some issues for FF due to no support for `contexts` in `network.addIntercept`.

CC @whimboo https://bugzilla.mozilla.org/show_bug.cgi?id=1882260


OrKoN (commented):
Does it fix any tests?

OrKoN (commented):
Overall LGTM: but I am not sure if we need to expose intercepts on the browser. Perhaps we should first implement context support.

packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -199,6 +209,33 @@ export class Browser extends EventEmitter<{
     return script;
   }
 
+  @throwIfDisposed<Browser>(browser => {
+    // SAFETY: By definition of `disposed`, `#reason` is defined.
+    return browser.#reason!;
+  })
+  async addIntercept(options: AddInterceptOptions): Promise<string> {
```
OrKoN: consider moving both intercept methods to BrowsingContext.ts

Lightning00Blade:
@OrKoN Context support is supported by `chromium-bidi` but not Firefox.

packages/puppeteer-core/src/bidi/core/Browser.ts
Lightning00Blade: You mean removing it from here and having it only there? 
Currently there is a method on the `BrowsingContext.ts` that wraps this one.

Lightning00Blade:
The other option if we don't want to implement a workaround is to fail on the command if its Firefox

packages/puppeteer-core/src/bidi/core/Browser.ts
OrKoN: Right, I do not think we will use on the browser level in the foreseeable future.

OrKoN (approved):

packages/puppeteer-core/src/bidi/HTTPRequest.ts
```diff
@@ -166,14 +168,14 @@ export class BidiHTTPRequest extends HTTPRequest {
...
+    return await this.#request.failRequest();
   }
 
-  override respond(
+  override async respond(
     _response: Partial<ResponseForRequest>,
     _priority?: number
-  ): never {
+  ): Promise<void> {
     throw new UnsupportedOperation();
```
OrKoN, line 179: also a thought: should we delay making a changelog update until most of the interception features are supported?

OrKoN: Let's probably do that (mark this a chore, and mark the last change as a feat + update webdriver support page), since it will read confusing in the changelog.

packages/puppeteer-core/src/bidi/BrowserContext.ts
```diff
@@ -185,6 +185,10 @@ export class BidiBrowserContext extends BrowserContext {
     }
 
     try {
+      for (const page of await this.pages()) {
+        // Workaround for Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1882260
```
OrKoN: let's add a `TODO: `

whimboo:
> There are some issues for FF due to no support for `contexts` in `network.addIntercept`.
> 
> CC @whimboo https://bugzilla.mozilla.org/show_bug.cgi?id=1882260

@Lightning00Blade: How important is this missing feature exactly? Does it break just some tests or does it cause wider trouble for the Puppeteer implementation?

Lightning00Blade:
@whimboo I think its important to have it for the users. Currently workaround will only work for uses that use a single page/context at a time anyone that uses multiple will face issues.

whimboo:
Ok, we will prioritize it.

PR #12119, opened 2024-03-21
fix: reload should not resolve early on fragment navigations

### LINKED ISSUE #10435: [Bug]: Page.reload() Destroys execution context on certain websites
### Bug expectation
We are trying to visit a webpage, reload it and query a selector.
On certain websites, a "Execution context was destroyed" error is thrown as soon as `$` query selector is used after the page reload.

### Bug behavior

- [X] Flaky
- [ ] PDF

```TypeScript
import * as Puppeteer from "puppeteer";

const browser = await Puppeteer.launch({
    ignoreHTTPSErrors: true,
    headless: false,
    devtools: true,
    waitForInitialPage: false,
});

await page.goto("https://ilsole24ore.com", {
    waitUntil: "load",
});

await page.reload();

const element = await page.$(".test");

// Error: Execution context was destroyed, most likely because of a navigation.
// Happens for "https://ilsole24ore.com"
// Instead if we try a different websites, like "https://aranzulla.it" it doesn't happen
```

### Error string
Error: Execution context was destroyed, most likely because of a navigation

### Puppeteer configuration

_No response_

### Puppeteer version

20.7.3

### Node version

20.3.0

### Package manager

npm

### Package manager version

9.6.7

### Operating system

macOS

OrKoN:
Reload would always trigger a new document navigation so it should wait for it instead of potential intermediate same-document navigations. 

Closes https://github.com/puppeteer/puppeteer/issues/10435

### DIFF
--- a/packages/puppeteer-core/src/api/Frame.ts
+++ b/packages/puppeteer-core/src/api/Frame.ts
@@ -63,6 +63,10 @@ export interface WaitForOptions {
    * @defaultValue `'load'`
    */
   waitUntil?: PuppeteerLifeCycleEvent | PuppeteerLifeCycleEvent[];
+  /**
+   * @internal
+   */
+  ignoreSameDocumentNavigation?: boolean;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


jrandolf-google (approved):
[redacted]

PR #12122, opened 2024-03-21
test: update test expectations for Firefox network interception

juliandescottes:
This PR enables 13 new tests for Firefox (I have not checked Chrome yet), and adds comments and links for remaining tests.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -808,9 +808,35 @@
   {
     "testIdPattern": "[requestinterception-experimental.spec] *",
     "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox", "webDriverBiDi"],
+    "expectations": ["PASS"]
+  },
+  {
+    "testIdPattern": "[requestinterception-experimental.spec] *",
+    "platforms": ["darwin", "linux", "win32"],
     "parameters": ["cdp", "firefox"],
     "expectations": ["FAIL", "SKIP"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
+  },
+  {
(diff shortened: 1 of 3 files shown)


test/TestExpectations.json
```diff
@@ -3362,13 +3569,152 @@
...
+    "parameters": ["firefox", "webDriverBiDi"],
+    "expectations": ["SKIP"],
+    "comment": "TODO: Needs support for data URIs in Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1805176"
+  },
+  {
+    "testIdPattern": "[requestinterception.spec] request interception Page.setRequestInterception should navigate to URL with hash and fire requests without hash",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox", "webDriverBiDi"],
+    "expectations": ["FAIL"],
+    "comment": "TODO: Needs investigation, not clear why the test expects the hash to be removed"
```
OrKoN, line 3626: I believe according to the HTTP standard the fragment part should not be sent over network to the server

OrKoN (approved):

test/TestExpectations.json
juliandescottes: Oh interesting. It sounds worth having a wdspec test to cover this? Does it mean the fragment should never be visible in responseData https://w3c.github.io/webdriver-bidi/#type-network-ResponseData ?

OrKoN: I would say so. https://www.rfc-editor.org/rfc/rfc7230.html#section-5.3 does not include things after # but I am not sure if the fetch spec adds it back in some form. At least it should not reach the server and I would say it should not be reported via BiDi's network but I guess we need to check the fetch spec.

PR #12147, opened 2024-03-25
chore(webdriver): scaffold `HTTPRequest.respond()`

Lightning00Blade:


test/src/requestinterception.spec.ts
```diff
@@ -799,7 +799,11 @@ describe('request interception', function () {
       });
       const response = (await page.goto(server.EMPTY_PAGE))!;
       expect(response.status()).toBe(422);
-      expect(response.statusText()).toBe('Unprocessable Entity');
+      expect(response.statusText()).atLeastOneToContain([
+        'Unprocessable Entity',
+        // TODO: is this OK it comes from Chromium defaults?
+        'Unprocessable Content',
```
Lightning00Blade, line 805: In Puppeteer we provide Custom Status text, we don't do that for `chromium-bidi`.

OrKoN (approved):

PR #12154, opened 2024-03-26
chore: don't detach CdpSession

Lightning00Blade:
Allow the Mapper to handle it's own clean-up, else w get errors in the logs + we send too many commands.
Users should clean the CdpSession on their own if created via `createCDPSession`

### DIFF
--- a/packages/puppeteer-core/src/bidi/CDPSession.ts
+++ b/packages/puppeteer-core/src/bidi/CDPSession.ts
@@ -21,7 +21,7 @@ export class BidiCdpSession extends CDPSession {
   static sessions = new Map<string, BidiCdpSession>();
 
   #detached = false;
-  readonly #connection: BidiConnection | undefined = undefined;
+  readonly #connection?: BidiConnection;
   readonly #sessionId = Deferred.create<string>();
   readonly frame: BidiFrame;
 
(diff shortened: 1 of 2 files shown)


OrKoN:
Does it solve any bug/issue?

Lightning00Blade:
There were Errors in the Logs, they are always cleaned up before the make the call so we get errors instead.

packages/puppeteer-core/src/bidi/CDPSession.ts
```diff
@@ -86,10 +88,13 @@ export class BidiCdpSession extends CDPSession {
...
-        sessionId: this.id(),
-      });
+      if (this.#userCreated) {
+        await this.frame.client.send('Target.detachFromTarget', {
+          sessionId: this.id(),
+        });
+      }
     } finally {
       BidiCdpSession.sessions.delete(this.id());
       this.#detached = true;
```
OrKoN: can we learn the mapper closed the session and set the detached flag? I would rather suppress the error than introduce the notion of the user created sessions

Lightning00Blade: We can just not call it. As the this is done on `browsingContext.constextDestroyed` event, and let the user handle the CDP session as they please.

OrKoN: We probably only need the BidiCdpSession.sessions.delete bit then

Lightning00Blade: OK, I will update the PR and title

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -106,7 +106,7 @@ export class BidiFrame extends Frame {
     this.browsingContext.on('closed', () => {
       for (const session of BidiCdpSession.sessions.values()) {
         if (session.frame === this) {
-          void session.detach().catch(debugError);
+          session.onClose();
```
Lightning00Blade, line 109: If the `BrowserContext` is closed Mapper the session is already closed so we need to only remove it from the map.

OrKoN (approved):

PR #12155, opened 2024-03-26
chore(webdriver): support options in `request.continue`

Lightning00Blade:


packages/puppeteer-core/src/api/HTTPRequest.ts
```diff
@@ -513,3 +516,13 @@ export const STATUS_TEXTS: Record<string, string> = {
   '510': 'Not Extended',
   '511': 'Network Authentication Required',
 } as const;
+
+export function handleError(error: ProtocolError): void {
+  if (error.originalMessage.includes('Invalid header')) {
```
Lightning00Blade, line 524: We should see how to make this cross browser.

OrKoN (approved):

PR #12191, opened 2024-04-02
chore(webdriver): support priority for request interception

Lightning00Blade:
This uses the same mechanism as did the CDP implantation.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -3405,13 +3405,6 @@
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
   {
-    "testIdPattern": "[requestinterception-experimental.spec] cooperative request interception Page.setRequestInterception should be able to access the error reason",
-    "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["firefox", "webDriverBiDi"],
-    "expectations": ["SKIP"],
-    "comment": "TODO: Needs Puppeteer support for BidiHTTPRequest.abortErrorReason"
-  },
-  {
     "testIdPattern": "[requestinterception-experimental.spec] cooperative request interception Page.setRequestInterception should be able to fetch dataURL and fire dataURL requests",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["firefox", "webDriverBiDi"],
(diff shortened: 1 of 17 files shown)


tools/docgen/src/custom_markdown_documenter.ts
```diff
@@ -876,14 +876,21 @@ export class MarkdownDocumenter {
...
-              new DocTableRow({configuration}, [
-                this._createTitleCell(apiMember, true),
-                this._createModifiersCell(apiMember),
-                this._createPropertyTypeCell(apiMember),
-                this._createDescriptionCell(apiMember),
-              ])
-            );
+            if (
+              !(
+                ApiProtectedMixin.isBaseClassOf(apiMember) &&
```
OrKoN: is this needed? can you mark the protected member as `@internal` instead?

test/TestExpectations.json
```diff
@@ -839,6 +833,13 @@
     "expectations": ["SKIP"],
     "comment": "TODO: Needs full support for continueResponse in Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1853887"
   },
+  {
+    "testIdPattern": "[requestinterception-experimental.spec] cooperative request interception Request.respond should redirect",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["webDriverBiDi"],
+    "expectations": ["FAIL"],
+    "comment": "Puppeteer request.redirectChain issue"
```
OrKoN, line 841: Could you be more specific here?

OrKoN (approved):

tools/docgen/src/custom_markdown_documenter.ts
Lightning00Blade: Those are internal by definition. We should not expose them in the docs. 
But happy to reconsider if there is an use case.

test/TestExpectations.json
Lightning00Blade: I am fixing it at the moment so didn't want to open a issue and link it.

tools/docgen/src/custom_markdown_documenter.ts
OrKoN: This is not an universal rule I think. In general, we could expose some protected members so that the users can subclass from it and access the members (e.g., like in transport classes).

Issue #12178, opened 2024-03-29
Update automation tools


OrKoN:
Use https://github.com/actions/stale instead of https://github.com/apps/stale (deprecated)

cc @Lightning00Blade

Lightning00Blade:
Closing as complete in https://github.com/puppeteer/puppeteer/pull/12192

PR #12192, opened 2024-04-02
chore: use stale action

Lightning00Blade:
Issue https://github.com/puppeteer/puppeteer/issues/12178

Note: we still need to disable the app.


.github/workflows/stale.yml
```diff
@@ -0,0 +1,31 @@
...
+          stale-issue-message: >
+            We're marking this issue as unconfirmed because it has not had recent activity and 
+            we weren't able to confirm it yet. It will be closed if no further activity occurs
+            within the next 30 days.
+          close-issue-message: >
+            We are closing this issue. If the issue still persists in the latest version of
+            Puppeteer, please reopen the issue and update the description. We will try our
+            best to accommodate it!
+          stale-issue-label: 'unconfirmed'
+          exempt-issue-labels: 'confirmed,feature'
```
OrKoN, line 29: is 30 days the default?

OrKoN: (for issues)

OrKoN:
> Note: we still need to disable the app.

How is it done?

.github/workflows/stale.yml
Lightning00Blade: `days-before-stale` - default `60` (Same as we had before)
`days-before-close` - default `7` (We had 30 days before)

Lightning00Blade:
> How is it done?
From what I understand we just need to revoke the permission to the repo. I found how to do it

OrKoN:
@Lightning00Blade have you done it? let's disable the old app first to avoid clashes.

Lightning00Blade:
> @Lightning00Blade have you done it? let's disable the old app first to avoid clashes.

Done

OrKoN (approved):

PR #12203, opened 2024-04-04
fix: providing null to page.authenticate should disable authentication

Lightning00Blade:
Aligns with our documentation that tell's we  can use `null` to disable,
Providing `user/pass` with `undefined` did not disable the `Fetch` domain as it should.

### DIFF
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -523,6 +523,14 @@ export interface PageEvents extends Record<EventType, unknown> {
 /**
  * @public
  */
+export interface Credentials {
+  username: string;
+  password: string;
+}
+
+/**
+ * @public
+ */
 export interface NewDocumentScriptEvaluation {
   identifier: string;
 }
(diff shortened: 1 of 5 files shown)


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -512,6 +512,14 @@ export interface PageEvents extends Record<EventType, unknown> {
   [PageEvent.WorkerDestroyed]: WebWorker;
 }
 
+/**
+ * @public
+ */
+export interface Credentials {
```
Lightning00Blade: Forgot to remove but will needed for BiDi.

OrKoN:
`fix: providing null to page.authenticate should disable authentication` (perhaps a better title)

OrKoN (approved):

Lightning00Blade:
Testing again with 125.0.6422.14 seems to work so waiting till that or newer version rolls to stable to re test.

OrKoN:
@Lightning00Blade can this be landed?

PR #12210, opened 2024-04-05
ci: add an option to run with experimental features

OrKoN:
- stops disabling ProcessPerSiteUpToMainFrameThreshold
- stops disabling field testing config


Lightning00Blade (approved):
LGTM, but why not just them as suite to the `TestSuite.json` ?

OrKoN:
I would need to duplicate every test suite with configuration

PR #12216, opened 2024-04-05
chore: remove nested navigation

Lightning00Blade:
I didn't manage to find any place that the nested navigation is used or it's events listened to. 
Something similar was being done at #11929


Lightning00Blade:
Closing as this will not be the solution we go with as we will emit a fail event for this.
We may go the https://github.com/puppeteer/puppeteer/pull/11929 route at some point.

Issue #11246, opened 2023-10-25
[Bug]:  puppeteer 20.6.0 introduced a regression (with iframe?) that now is Error: Requesting main frame too early!


andreabisello:
### Minimal, reproducible example

i can't give the code to reproduce because is a business code, so probably this will be a "closed" ticket, but anyway i can describe it, i hope can be useful.

the same error is repored [here](https://stackoverflow.com/questions/77213861/puppeteer-requesting-main-frame-too-early/77352402#77352402) by other users, no one could understand how to reproduce systematically.

### Error string

Error: Requesting main frame too early!

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

i open a browser, i open a page, a navigate to a webpage, i click a button, that opens a iframe, i take a reference to the iframe, i write some text, i click a button that close the iframe, and the error occur and the browser is closed (or the browser is closed and then the error occur).

**that test worked for 2 years until the introduction of puppeteer 20.6.0**

**the website still the same**

### Expectation

when the website close the iframe, the execution of the code should keep working.

### Reality

when, clicking a button, the website close the iframe, the browser close and the error is raised (or the error is raised and so the browser close)

until puppeteer 20.5.0 this was working.
with puppeteer 20.6.0 you obtain

> failed: Protocol error (Runtime.callFunctionOn): Session closed. Most likely the page has been closed..

with puppeteer 21.3 or also 21.4.1 you obtain

> Error: Requesting main frame too early!.

### Puppeteer version

21.4.1

### Node version

18.16.1

### Package manager

npm

### Package manager version

9.5.1

### Operating system

Windows

OrKoN:
@andreabisello can you create a test page that resembles the real issue?

OrKoN:
The error indicates that there is an action on the main frame but Puppeteer has not received the information about the main frame from the browser yet. Perhaps it is a race condition but it would also depend on how exactly the frame is accessed.

andreabisello:
@OrKoN i cant because is company code that test company software. maybe i should create a website with an iframe and a button that open that iframe, makes something, and close the iframe. 

they way i access the frame is that one

```
        let creationFrame
        while(!creationFrame){
            await page.waitForTimeout(1000)
            for(const frame of page.mainFrame().childFrames()){
                if(frame.url().includes("EditorFatturaPA")){
                    creationFrame = frame
                }
            }
        }
```

then i make every interation using "creationFrame" , for clicking typing etc ...
[...]

PR #12225, opened 2024-04-05
fix: retain stale main frame for longer

OrKoN:
Although https://github.com/puppeteer/puppeteer/issues/11246 has no reproducible example and the root cause might be the access to main frames while they are navigating, this PR adds a workaround to retain a potentially stale main frame to prevent assertion errors.

Bug: #11246


Lightning00Blade (approved):

andreabisello:
@OrKoN  thanks. I will update my testsuite to the puppeteer version that contains that fix when ready, and return for a feedback. How could i know wich puppeteer version contains that fix? thankyou.

OrKoN:
@andreabisello it will be in the next release, you can follow the linked release PR for that

PR #12233, opened 2024-04-07
feat: allow creating ElementHandles from the accessibility tree snapshot

### LINKED ISSUE #3641: Accessibility: expose element reference

OrKoN:
Closes #3641

### DIFF
--- a/packages/puppeteer-core/src/cdp/Accessibility.ts
+++ b/packages/puppeteer-core/src/cdp/Accessibility.ts
@@ -6,8 +6,8 @@
 
 import type {Protocol} from 'devtools-protocol';
 
-import type {CDPSession} from '../api/CDPSession.js';
 import type {ElementHandle} from '../api/ElementHandle.js';
+import type {Realm} from '../api/Realm.js';
 
 /**
  * Represents a Node and the properties of it that are relevant to Accessibility.
@@ -80,6 +80,14 @@ export interface SerializedAXNode {
    * Children of this node, if there are any.
    */
   children?: SerializedAXNode[];
+
+  /**
+   * Get an ElementHandle for this AXNode if available.
+   *
+   * If the underlying DOM element has been disposed, the method might return an
+   * error.
+   */
+  elementHandle(): Promise<ElementHandle | null>;
 }
 
 /**
@@ -121,20 +129,13 @@ export interface SnapshotOptions {
  * @public
  */
 export class Accessibility {
-  #client: CDPSession;
+  #realm: Realm;
 
   /**
    * @internal
    */
-  constructor(client: CDPSession) {
-    this.#client = client;
-  }
-
-  /**
-   * @internal
-   */
-  updateClient(client: CDPSession): void {
-    this.#client = client;
+  constructor(realm: Realm) {
+    this.#realm = realm;
   }
 
   /**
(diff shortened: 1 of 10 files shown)


packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -441,6 +448,14 @@ class AXNode {
...
     const node: SerializedAXNode = {
       role: this.#role,
+      elementHandle: async (): Promise<ElementHandle | null> => {
+        if (!this.payload.backendDOMNodeId) {
+          return null;
+        }
+        return (await this.#realm.adoptBackendNode(
+          this.payload.backendDOMNodeId
+        )) as ElementHandle<Element>;
+      },
```
Lightning00Blade, line 458: How about something like this, so we change the struct less?  
And you are can just check `'elementHandle' in X`
```suggestion
      ...{ this.payload.backendDOMNodeId ? {
        elementHandle: async (): Promise<ElementHandle> => {
          return (await this.#realm.adoptBackendNode(
            this.payload.backendDOMNodeId
          )) as ElementHandle<Element>;}
        }
        : {}
      },   
```

OrKoN: I think it's better if the method is always available. I don't think we are constrained by changing the class and perhaps we could turn it into a proper class later.

Lightning00Blade (approved):
LGTM

PR #12244, opened 2024-04-10
fix: suppress error for wrong error code

Lightning00Blade:
BEGIN_COMMIT_OVERRIDE
fix(webdriver): suppress error for error code errors
END_COMMIT_OVERRIDE


OrKoN (approved):

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -301,10 +302,18 @@ export class BidiFrame extends Frame {
...
-        Bidi.BrowsingContext.ReadinessState.Interactive
-      ),
+      this.browsingContext
+        .navigate(url, Bidi.BrowsingContext.ReadinessState.Interactive)
+        .catch(error => {
+          if (
+            isErrorLike(error) &&
+            error.message.includes('net::ERR_HTTP_RESPONSE_CODE_FAILURE')
+          ) {
+            return;
```
OrKoN, line 312: can we return the network response here?

Lightning00Blade: We do that is handled by the `this.waitForNavigation(options),` not by this command. this needs to not throw.

PR #12266, opened 2024-04-12
refactor: extract toggling of Network Interception

Lightning00Blade:


packages/puppeteer-core/src/cdp/NetworkManager.ts
```diff
@@ -140,18 +140,16 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
     );
   }
 
-  async setExtraHTTPHeaders(
-    extraHTTPHeaders: Record<string, string>
-  ): Promise<void> {
-    this.#extraHTTPHeaders = {};
-    for (const key of Object.keys(extraHTTPHeaders)) {
-      const value = extraHTTPHeaders[key];
+  async setExtraHTTPHeaders(headers: Record<string, string>): Promise<void> {
```
Lightning00Blade, line 143: Ops this was suppose to be separate, tell me if you want me to extract.
But we should should change the state only after assertion is completed.

OrKoN (approved):

PR #12279, opened 2024-04-16
feat(webdriver): support Network interception

OrKoN:
This PR concludes the work to support network interception using WebDriver BiDi.
Some of the network request and response attributes remain unsupported though. 

Related issues in WebDriver BiDi implementations:
- Chrome https://github.com/GoogleChromeLabs/chromium-bidi/issues/2035
- Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1853882 and https://bugzilla.mozilla.org/show_bug.cgi?id=1850680

### DIFF
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -61,6 +61,8 @@ import {rewriteNavigationError} from './util.js';
 import type {BidiWebWorker} from './WebWorker.js';
 
 /**
+ * Implements Page using WebDriver BiDi.
+ *
  * @internal
  */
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -61,6 +61,8 @@ import {rewriteNavigationError} from './util.js';
 import type {BidiWebWorker} from './WebWorker.js';
 
 /**
+ * Implements Page using WebDriver BiDi.
```
OrKoN, line 64: Including a no-op file change

Lightning00Blade (approved):

PR #12290, opened 2024-04-17
docs: update the extensions guide

OrKoN:


OrKoN:
@oliverdunk PTAL

Lightning00Blade (approved):

docs/guides/chrome-extensions.md
```diff
@@ -42,8 +39,40 @@ Chrome Manifest V3 extensions have a background ServiceWorker of type
...
+const pathToExtension = path.join(process.cwd(), 'my-extension');
+const browser = await puppeteer.launch({
+  args: [
+    `--disable-extensions-except=${pathToExtension}`,
+    `--load-extension=${pathToExtension}`,
+  ],
+});
+
+const workerTarget = await browser.waitForTarget(
+  target => target.type() === 'service_worker' && idMatches(target)
```
oliverdunk: I think we should probably remove the `idMatches` call here, unless we want to document how to have an extension with a consistent ID (which seems a little out of scope?).

docs/guides/chrome-extensions.md
```diff
@@ -42,8 +39,40 @@ Chrome Manifest V3 extensions have a background ServiceWorker of type
...
+
+// Test the popup page as you would any other page.
+
+await browser.close();
+```
+
 :::note
 
-It is not yet possible to test extension popups or content scripts.
+It is not yet possible to test extension content scripts.
```
oliverdunk, line 85: Do you know what the content script callout here was referring to? Or to put it another way, what would we want to exist before we could remove this?

docs/guides/chrome-extensions.md
```diff
@@ -4,8 +4,8 @@ Puppeteer can be used for testing Chrome Extensions.
```
oliverdunk, line 4: Not sure if we want to link here which has some more advice: https://developer.chrome.com/docs/extensions/how-to/test/end-to-end-testing

oliverdunk (approved):
Thanks Alex! I left a couple of notes but generally LGTM.

docs/guides/chrome-extensions.md
OrKoN: I think it is about executing code in the content scripts created by extensions

PR #12295, opened 2024-04-19
refactor: simplify the Lifecycle Watcher

Lightning00Blade:
We can expect the correct order of event for `DOMContentLoaded` and `load` events,
that the main frame's will come last [upstream CL](https://chromium-review.googlesource.com/c/chromium/src/+/4684286).

### DIFF
--- a/packages/puppeteer-core/src/cdp/LifecycleWatcher.ts
+++ b/packages/puppeteer-core/src/cdp/LifecycleWatcher.ts
@@ -17,7 +17,6 @@ import {Deferred} from '../util/Deferred.js';
 import {DisposableStack} from '../util/disposable.js';
 
 import type {CdpFrame} from './Frame.js';
-import {FrameManagerEvent} from './FrameManagerEvents.js';
 import type {NetworkManager} from './NetworkManager.js';
 
 /**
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/cdp/LifecycleWatcher.ts
```diff
@@ -227,41 +223,19 @@ export class LifecycleWatcher {
...
-      for (const event of expectedLifecycle) {
-        if (!frame._lifecycleEvents.has(event)) {
-          return false;
-        }
-      }
-      // TODO(#1): Its possible we don't need this check
-      // CDP provided the correct order for Loading Events
-      // And NetworkIdle is a global state
-      // Consider removing
-      for (const child of frame.childFrames()) {
```
OrKoN, line 255: so was it here only for the empty frames? isn't this code also checks for networkidle events which could come in any order of the frames/subframes? Is it definitely not possible that the load or domcontent loader events arrive after the main frame load/domcontentloaded?

Lightning00Blade: So for the Lifecycle events they now (after the linked CL) come in order, that's only true for parsing so anything coming from the HTML or Scripts (evaluateOnNewDocument included) will be correct.
As for the `NetworkIdle` I have to double check but I think the state for every frame is track on the same NetworkIdle detector so it's global for the page (comment also suggest that).

OrKoN: I am not sure it can be tracked globally with OOPIFs

Lightning00Blade: Yeah you may be correct. I don't think we have such test. I will write one here.

Leezsusan (commented):
Hey man don't be reviewing my darn life cycle I came before not after. Time is different here you know

****

packages/puppeteer-core/src/cdp/LifecycleWatcher.ts
OrKoN: @Lightning00Blade do you want to keep the PR open?

Lightning00Blade:
Closing as the network idle parts need this to properly detect that part of the lifecycle.

PR #12309, opened 2024-04-23
fix(webdriver): configure Firefox to start with the default user context only

Lightning00Blade:
By default Firefox start with multiple UserContexts (containers). We need to disable this by only providing single Container to the `containers.json` file and pointing to it from `lastUserContextId`.

### DIFF
--- a/packages/browsers/src/browser-data/firefox.ts
+++ b/packages/browsers/src/browser-data/firefox.ts
@@ -407,7 +407,6 @@ function defaultProfilePreferences(
  * @param profilePath - Firefox profile to write the preferences to.
  */
 async function writePreferences(options: ProfileOptions): Promise<void> {
-  // Create a backup of the preferences file if it already exitsts.
   const prefsPath = path.join(options.path, 'prefs.js');
   const lines = Object.entries(options.preferences).map(([key, value]) => {
     return `user_pref(${JSON.stringify(key)}, ${JSON.stringify(value)});`;
@@ -415,12 +414,17 @@ async function writePreferences(options: ProfileOptions): Promise<void> {
 
   await Promise.all([
     fs.promises.writeFile(path.join(options.path, 'user.js'), lines.join('\n')),
+    // Create a backup of the preferences file if it already exitsts.
     fs.promises
       .access(prefsPath, fs.constants.F_OK)
-      .then(backUpPrefs)
-      .catch(() => {
-        return false;
-      }),
+      .then(async () => {
+        await fs.promises.copyFile(
+          prefsPath,
+          path.join(options.path, 'prefs.js.puppeteer')
+        );
+      })
+      // Swallow if the file does not exist
+      .catch(() => {}),
     options.disableExtraUserContexts ? shouldUpdateContainers() : undefined,
   ]);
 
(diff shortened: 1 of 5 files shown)


packages/browsers/src/browser-data/firefox.ts
```diff
@@ -421,6 +422,27 @@ async function writePreferences(options: ProfileOptions): Promise<void> {
     const prefsBackupPath = path.join(options.path, 'prefs.js.puppeteer');
     await fs.promises.copyFile(prefsPath, prefsBackupPath);
   }
+
+  if (options.disableExtraUserContexts) {
+    const containersPath = path.join(options.path, 'containers.json');
+
+    const singleContainer = JSON.stringify({
```
OrKoN, line 434: @whimboo we want to disable extra user contexts for Firefox in Puppeteer (for profiles created by Puppeteer). This works but I wonder if this solution is something we can rely on?

P.S. @Lightning00Blade did you notice any speed up in launching Firefox?

OrKoN (approved):
Let's wait for feedback from Henrik

packages/browsers/src/browser-data/firefox.ts
Lightning00Blade: Running `page.spec.ts` with hook, shows diff in the margin of error (under 1%)
```ts
  after(() => {
    console.log('End', performance.now());
  });
```

But I think there is slight overhead for writing the file if I isolate a test is around 100ms, I think I can optimize it by paralleling the writing of the file.

Lightning00Blade: With parallelization it seems the overhead of writing a file to be around `30-50ms` and that varies from run to run a lot have seen the old one be slower and the new one be faster.

packages/browsers/src/browser-data/firefox.ts
```diff
@@ -406,20 +407,46 @@ function defaultProfilePreferences(
...
+          prefsPath,
+          path.join(options.path, 'prefs.js.puppeteer')
+        );
+      })
+      // Swallow if the file does not exist
+      .catch(() => {}),
+    options.disableExtraUserContexts ? shouldUpdateContainers() : undefined,
+  ]);
+
+  async function shouldUpdateContainers() {
```
OrKoN:
```suggestion
  async function writeContainersFile() {
```

?

packages/browsers/src/browser-data/firefox.ts
juliandescottes: (whimboo is off this week)

There are two internal containers that should be preserved:
https://searchfox.org/mozilla-central/rev/6121b33709dd80979a6806ff59096a561e348ae8/toolkit/components/contextualidentity/ContextualIdentityService.sys.mjs#85-105

userContextIdInternal.thumbnail and userContextIdInternal.webextStorageLocal
I am not sure what are the side effects of removing the second one, but I think it's better to keep it.

Note that the current version was recently bumped to 5 (https://bugzilla.mozilla.org/show_bug.cgi?id=1814969), but it's fine to keep 4 for now, and anyway the contextualidentityservice has migration logic based on the version.

Overall I would say this is fine as a temporary solution but we should try to introduce a preference in Firefox in order to skip the default user profiles.

OrKoN: @Lightning00Blade let's probably rewrite the tests to expect any number of non-default contexts.

Lightning00Blade: Yeah sound good, I think I would extract some of the changes from this PR as well, I think we can make FF launch just a little bit faster.

PR #12315, opened 2024-04-24
feat(webdriver): support ARIA selectors

OrKoN:
This PR switches the implementation for ARIA selectors to use browsingContext.locateNodes.

Note that this method does not pierce shadow roots.

### DIFF
--- a/packages/puppeteer-core/src/cdp/ElementHandle.ts
+++ b/packages/puppeteer-core/src/cdp/ElementHandle.ts
@@ -184,7 +184,7 @@ export class CdpElementHandle<
       role,
     });
 
-    const results = nodes.filter((node: Protocol.Accessibility.AXNode) => {
+    const results = nodes.filter(node => {
       if (node.ignored) {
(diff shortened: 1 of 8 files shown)


packages/puppeteer-core/src/cdp/ElementHandle.ts
```diff
@@ -169,4 +173,34 @@ export class CdpElementHandle<
...
+    name?: string | undefined,
+    role?: string | undefined
+  ): AwaitableIterable<ElementHandle<Node>> {
+    const {nodes} = await this.client.send('Accessibility.queryAXTree', {
+      objectId: this.id,
+      accessibleName: name,
+      role,
+    });
+
+    const results = nodes.filter((node: Protocol.Accessibility.AXNode) => {
```
Lightning00Blade: Nit:
```suggestion
    const results = nodes.filter(node => {
```

Lightning00Blade (approved):

PR #12336, opened 2024-04-25
Switch to macos 13 arm

juliandescottes:
Just trying jobs 13 arm


juliandescottes:
A patch is now landing on mozilla-central to address the core issue.

PR #12349, opened 2024-04-26
fix: deprecate CDP for Firefox

OrKoN:
This PR adds a deprecation warning if Chrome DevTools Protocol (CDP) is used for Firefox. Instead CDP, use [WebDriver BiDi](https://pptr.dev/webdriver-bidi#get-started):

```

  product: 'firefox',
  protocol: 'webDriverBiDi',
});
...
await browser.close();
```

### DIFF
--- a/packages/puppeteer-core/src/node/ProductLauncher.ts
+++ b/packages/puppeteer-core/src/node/ProductLauncher.ts
@@ -112,6 +112,18 @@ export abstract class ProductLauncher {
       });
     };
 
+    if (
+      this.#product === 'firefox' &&
+      protocol !== 'webDriverBiDi' &&
+      this.puppeteer.configuration.logLevel === 'warn'
+    ) {
+      console.warn(
+        `Chrome DevTools Protocol (CDP) support for Firefox is deprecated in Puppeteer ` +
+          `and it will be eventually removed. ` +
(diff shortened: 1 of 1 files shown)


Lightning00Blade (approved):
LGTM + maybe we need a call to action if there encounter issues?

OrKoN:
@Lightning00Blade code to action?

Lightning00Blade:
> @Lightning00Blade code to action?
Typo I wanted to say `call to action` as there is no issue way for a person to say this is not working file and issue or something else

OrKoN:
@whimboo PTAL I think on our end we are ready to mark CDP support for Firefox as deprecated

whimboo:
> @whimboo PTAL I think on our end we are ready to mark CDP support for Firefox as deprecated

I thought that we wanted to do that after the announcement that Puppeteer fully supports Firefox via WebDriver BiDi. For me it feels a bit too early.

OrKoN:
@whimboo we are only [13 tests away ](https://puppeteer.github.io/ispuppeteerwebdriverbidiready/firefox-delta.json) from feature-parity with CDP and 147 new tests are passing so I think it is fine to nudge users to try WebDriver BiDi and switch over.

whimboo (approved):
As discussed with the team we are fine with this addition. Also because you are not forcing Webdriver BiDi at this point as default which is fine.

PR #12352, opened 2024-04-26
fix(cdp): throw on closed connection

Lightning00Blade:
Prevent users from having to wait for the `ProtocolTimeout` which is very long as we know it will never resolve


OrKoN (approved):

packages/puppeteer-core/src/cdp/Connection.ts
```diff
@@ -117,6 +118,8 @@ export class Connection extends EventEmitter<CDPSessionEvents> {
     sessionId?: string,
     options?: CommandOptions
   ): Promise<ProtocolMapping.Commands[T]['returnType']> {
+    assert(!this.#closed, 'Protocol error: Connection closed.');
```
OrKoN, line 121: so this turns out to actually be a breaking change, the errors can not be thrown synchronously

OrKoN: I am gonna publish a fix

Issue #12292, opened 2024-04-18
[Bug]: Freezing when querying frames on certain websites


Vasile-Peste:
### Minimal, reproducible example

```TypeScript
const puppeteer = require("puppeteer");

async function main () {
    console.log("Operative");

    const browser = await puppeteer.launch();
    const page = await browser.newPage();

    await page.goto('http://fonologomed.it/');

    const allIframeLinks = [];
    const frames = page?.frames() ?? [];

    for (const frame of frames) {
        try {
            if (!frame?.url() || frame.detached) {
                continue;
            }

            const links = await frame.$$eval('a', (anchors) => anchors.map((anchor) => ({
            href: anchor.href,
                text: anchor.text,
    	    })));

            allIframeLinks.push(...links);
        }
        catch (e) {
            console.log(e);
        }
    }

    console.log(allIframeLinks);
    console.log("Success");
}

main();
```

### Error string

no error

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

We want to retrieve the <a> links from the iframes within a website. This works for the majority of visited websites, but we noticed on certain websites this operation completely is freezing (the Promise is pending forever).
Running the reproducible example on http://fonologomed.it/ will freeze
For a contrarian example, running the reproducible example on https://www.aranzulla.it will NOT freeze

### Expectation

The promise to resolve or at least get an error

### Reality

The operation gets stuck as the Promise never gets resolved

### Puppeteer version

22.6.5

### Node version

20.3.0

### Package manager

npm

### Package manager version

9.6.7

### Operating system

macOS

Vasile-Peste:
I guess it cannot be reproduced by the workflow because it gets stuck?

AntonioFavero:
The promise frame.$$eval is never resolved.

OrKoN:
I think it is another example of https://github.com/puppeteer/puppeteer/issues/10696 basically the browser built-in PDF viewer creates empty iframes which never gets an execution context created for them as regular iframe. So the eval calls wait for the context as expected from the Chrome DevTools Protocol, but never gets them for those frames. Maybe it is because it is treated as part of the browser UI although the technical implementation is an iframe.

OrKoN:
A workaround could be like

```
      let parent = frame.parentFrame();
      let skip = false;
      while (parent) {
        if (parent.url().endsWith('.pdf')) {
          skip = true;
          break;
        }
        parent = parent.parentFrame();
      }
[...]

Issue #12172, opened 2024-03-28
[Bug]: using exposeFunction on a page with an iframe pointing to a pdf causes hang


eoghanmurray:
### Minimal, reproducible example

```TypeScript
import puppeteer from 'puppeteer'; // TS/ESM are all supported.

(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  await page.goto('about:blank');

  // assuming a valid PDF
  await page.setContent('<html><body><iframe src="some-doc.pdf"></iframe></body></html>')

  await page.exposeFunction('test', () => { console.log('test'); });
  // !!! don't get to here
  await browser.close();
})();
```

### Error string

no error (hangs)

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

This is in a suite of tests and causes the tests to hang
https://github.com/eoghanmurray/rrweb/commit/845bf5cb57caf19b32e15f0f74ce71cbc0d0ce9d#diff-f822a25730eb990747d0f997cbcbf61789e7924c84988c09f3666935d5d6a577R624

### Expectation

not to try to inject the 'test' function into the iframe document, as it isn't a regular javascript environment.

additionally, I expected it not to hang when failing to do so, but to have some sort of timeout to indicate that the injection failed to complete

### Reality

tests hang

### Puppeteer version

22.6.1

### Node version

27.7.1

### Package manager

yarn

### Package manager version

1.2.3

### Operating system

Linux

eoghanmurray:
22.6.2 not released at https://classic.yarnpkg.com/en/package/puppeteer yet

OrKoN:
Not able to reproduce on Mac with:

```
import puppeteer from 'puppeteer'; // TS/ESM are all supported.

(async () => {
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  await page.goto('about:blank');

  await page.setContent('<html><body><iframe src="https://www.w3.org/WAI/ER/tests/xhtml/testfiles/resources/pdf/dummy.pdf"></iframe></body></html>')
  await page.exposeFunction('test', () => { console.log('test'); });
  // !!! don't get to here
  await browser.close();
})();
```

OrKoN:
Also, not with using a file URL to a local file. Perhaps the issue is specific to Linux.

OrKoN:
Able to repro with a PDF served from a local http server.

OrKoN:
[...]

Issue #10715, opened 2023-08-09
[Bug]: Evaluate may hang if frame changes to OOPIF


liamdalg:
### Minimal, reproducible example

```TypeScript
import puppeteer from 'puppeteer';

const OOPIF_BASE = 'http://127.0.0.1:8000';
const BASE = 'http://localhost:8000';

const sleep = t => new Promise(r => setTimeout(r, t));

const browser = await puppeteer.launch({
    headless: 'new',
});

const page = await browser.newPage();
await page.goto(`${BASE}/iframe.html`);

for (const f of page.frames()) {
    console.log(`Evaluating in ${f._id}, ${f.url()}, ${f.isOOPFrame()}`);
    await f.evaluate(() => 1 + 1);
    console.log('Done');
}

await page.$eval('iframe', (f, b) => f.setAttribute('src', `${b}/blank.html`), OOPIF_BASE);

for (const f of page.frames()) {
    console.log(`Evaluating in ${f._id}, ${f.url()}, ${f.isOOPFrame()}`);
    await f.evaluate(() => 1 + 1);
    console.log('Done');
}

await browser.close();
```

### Error string

no error

### Bug behavior

- [X] Flaky
- [ ] PDF

### Background

In new headless, if a call to `frame.evaluate` races against the frame changing from a normal iframe to an OOPIF, then it may hang. The minimal example attached requires the two attached files to be available at `localhost:8000`; I've been running it with `python3 -m http.server`.  Note that this is quite flaky given it's a race condition. On my machine it triggers a hang around 20% of the time.

[HTML Files](https://github.com/puppeteer/puppeteer/files/12299487/html.zip)
[oopif.log](https://github.com/puppeteer/puppeteer/files/12299594/oopif.log)

Possibly related to my other bug report #10696.

For now, I've been avoiding this by running with `--disable-features=SitePerProcess`.

### Expectation

Either:
* Evaluation to succeed, or
* `Error: Execution context was destroyed, most likely because of a navigation` to be thrown.

### Reality

`frame.evaluate` hangs indefinitely.

### Puppeteer version

21.0.2

### Node version

18.16.1

### Package manager

npm

### Package manager version

9.5.1

### Operating system

Windows

OrKoN:
Looks like there is a race condition here because iframe navigation is asynchronous. We should at least error here.

[...]

Issue #10696, opened 2023-08-04
[Bug]: Evaluate hangs inside of embedded PDF frames


liamdalg:
### Minimal, reproducible example

```TypeScript
import puppeteer from 'puppeteer';

const sleep = t => new Promise(r => setTimeout(r, t));
const url = `file://${process.cwd()}/example.pdf`;

const browser = await puppeteer.launch({
    headless: 'new',
});

const page = await browser.newPage();
await page.goto(url);
await sleep(5000);

const frames = page.frames();
for (const f of frames) {
    console.log(`Evaluating in ${f._id}, ${f.url()}, ${f.isOOPFrame()}`);
    await f.evaluate(() => 1 + 1);
}

console.log('Done');
await browser.close();
```

### Error string

no error

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

In new headless or headful mode (i.e., when PDFs are rendered in the browser), attempting to evaluate inside of the PDF frame hangs indefinitely. This is true for any methods which attempt to use the frame's execution context. From some limited debugging, it seems that puppeteer is waiting indefinitely for an execution context that doesn't exist.

Possibly related (?): the CDP event `"Page.frameDetached","params":{"frameId":"31218988D67AB9B8AC1BD6140FCC9433","reason":"swap"}` is fired which, according to the `FrameManager` code, means that the frame should now be an OOPIF frame, but `frame.isOOPFrame()` returns `false`. 

**Debug**: [hang.log](https://github.com/puppeteer/puppeteer/files/12261505/hang.2.log)
**PDF used (any will work)**: [example.pdf](https://github.com/puppeteer/puppeteer/files/12261423/example.pdf)

### Expectation

Either:
* Evaluation to succeed inside of the PDF context, or
* Puppeteer to throw an error stating that the execution context was destroyed.

### Reality

Program hangs on `frame.evaluate`. More specifically, it hangs when waiting for the deferred execution context (https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/util/Deferred.ts#L74).

### Puppeteer version

21.0.1

### Node version

18.16.1

### Package manager

npm

### Package manager version

9.5.1

### Operating system

Linux

00000o1:
[...]

PR #12370, opened 2024-04-30
fix: turn on PdfOopif for PDF viewer

OrKoN:
OOPIFs are supported by Puppeteer whereas a guest view is not. Chrome plans to eventually move the PDF viewer to OOPIFs and we can already enable it in Puppeteer. This should consistently fix the linked issues except for https://github.com/puppeteer/puppeteer/issues/10715 which appears to be a general bug with local frames turning OOPIFs.

Refs: #12292, #10696, #12172


Lightning00Blade (approved):

test/TestExpectations.json
```diff
@@ -1437,10 +1444,10 @@
   },
   {
     "testIdPattern": "[elementhandle.spec] ElementHandle specs ElementHandle.clickablePoint should work for iframes",
-    "platforms": ["linux", "win32"],
+    "platforms": ["linux", "win32", "win32"],
```
kblok, line 1447: @OrKoN is this duple ok?

OrKoN: didn't notice it, I guess it is fine since our tooling converts it into a set, but we fixed the test and removed the expectation later

PR #12371, opened 2024-04-30
test: puppeteer.connect with `using` disconnects the browser

Lightning00Blade:
There is one caveat:
When we spawn a browser and we `disconnect` it immediately, its connection is closed so we can't send further commands or close the process.
When that happens we need to use `browser.close` from the second connection to properly close the browser process.


OrKoN (approved):

test/src/launcher.spec.ts
```diff
@@ -829,6 +829,8 @@ describe('Launcher specs', function () {
             browserWSEndpoint,
             protocol: browser.protocol,
           });
+          remoteClose = remoteBrowser.close.bind(remoteBrowser);
+          console.log(remoteClose);
```
OrKoN, line 833: forgotten console.log @Lightning00Blade

PR #12374, opened 2024-04-30
fix: don't start dbus service during build in Dockerfile

topiaruss:
Removing from image build phase a command that only makes sense at runtime.

No tests, no doc updates

Other changes will be required to maintain the intent of the dbus service command, but that will be informed by the direction of the project.

This will prevent a developer thinking that dbus *should* be running, when the actual command runs the service for a short time only during build not during runtime.

Probably not

**Other information**

Please feel free to reject this, and incorporate an alternative PR. I'm just highlighting the issue.


OrKoN:
That's a bug, thanks!

OrKoN (approved):

PR #12377, opened 2024-05-02
fix: remove --disable-field-trial-config

OrKoN:
It appears that --disable-field-trial-config might actually seed some other experimental variations so it is better to disable it.


Lightning00Blade (approved):

charliekump-wf:
Post-observation: This broke a testing implementation where were were hosting a service locally but were authenticated on a deployed site. The third-party cookies were blocked.

PR #12398, opened 2024-05-06
refactor: move binding installation into the execution context

OrKoN:
This PR moves bindings installation to be per context and makes ExecutionContext a Disposable.

### DIFF
--- a/packages/puppeteer-core/src/cdp/ExecutionContext.ts
+++ b/packages/puppeteer-core/src/cdp/ExecutionContext.ts
@@ -11,29 +11,36 @@ import type {ElementHandle} from '../api/ElementHandle.js';
 import type {JSHandle} from '../api/JSHandle.js';
 import {LazyArg} from '../common/LazyArg.js';
 import {scriptInjector} from '../common/ScriptInjector.js';
-import type {EvaluateFunc, HandleFor} from '../common/types.js';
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/cdp/ExecutionContext.ts
```diff
@@ -50,8 +57,91 @@ export class ExecutionContext {
...
+  // Set of bindings that have been registered in the current context.
+  #contextBindings = new Set<string>();
+
+  // Contains mapping from functions that should be bound to Puppeteer functions.
+  #bindings = new Map<string, Binding>();
+
+  // If multiple waitFor are set up asynchronously, we need to wait for the
+  // first one to set up the binding in the page before running the others.
+  #mutex = new Mutex();
+  async _addBindingToContext(name: string): Promise<void> {
```
Lightning00Blade, line 72: Nit: we only use it in this class now. 
```suggestion
  async #addBindingToContext(name: string): Promise<void> {
```

Lightning00Blade (approved):

packages/puppeteer-core/src/cdp/ExecutionContext.ts
OrKoN: I will refactor internals later

PR #12402, opened 2024-05-06
refactor: move execution context events

OrKoN:


packages/puppeteer-core/src/cdp/FrameManager.ts
```diff
@@ -505,35 +495,15 @@ export class FrameManager extends EventEmitter<FrameManagerEvents> {
...
-  #onExecutionContextsCleared(session: CDPSession): void {
-    for (const [key, context] of this.#contextIdToContext.entries()) {
-      // Make sure to only clear execution contexts that belong
-      // to the current session.
-      if (context._client !== session) {
-        continue;
-      }
-      if (context._world) {
-        context._world.clearContext();
+    context.once('dispose', () => {
```
OrKoN: Later I will move this to IsolatedWorld and also remove context tracking here

packages/puppeteer-core/src/cdp/ExecutionContext.ts
```diff
@@ -59,10 +60,17 @@ const ariaQuerySelectorAllBinding = new Binding(
   }) as (...args: unknown[]) => unknown
 );
 
+interface ExecutionContextEvents extends Record<EventType, unknown> {
+  dispose: undefined;
```
OrKoN: Do we need to have an Enum for events?

Lightning00Blade: We can do what we do BiDi core, though it would be nice to only allow that for internal classes.

Lightning00Blade: Nit:
```suggestion
  disposed: undefined;
```

packages/puppeteer-core/src/cdp/ExecutionContext.ts
```diff
@@ -87,6 +96,14 @@ export class ExecutionContext implements Disposable {
       'Runtime.bindingCalled',
       this.#onBindingCalled.bind(this)
     );
+    this.#clientEmitter.on('Runtime.executionContextDestroyed', async event => {
```
Lightning00Blade: Now that I take a look `this.#clientEmitter` can be local variable in the constructor.

Lightning00Blade (approved):

PR #12406, opened 2024-05-07
refactor: replace deferred with rxjs

OrKoN:
This allows easily accessing the execution context if it is available and gracefully waiting for it taking the world disposals into account.

### DIFF
--- a/packages/puppeteer-core/src/cdp/IsolatedWorld.ts
+++ b/packages/puppeteer-core/src/cdp/IsolatedWorld.ts
@@ -6,18 +6,22 @@
 
 import type {Protocol} from 'devtools-protocol';
 
+import {firstValueFrom, map, raceWith} from '../../third_party/rxjs/rxjs.js';
 import type {CDPSession} from '../api/CDPSession.js';
 import type {ElementHandle} from '../api/ElementHandle.js';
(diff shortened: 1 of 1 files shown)


packages/puppeteer-core/src/cdp/IsolatedWorld.ts
```diff
@@ -65,36 +75,45 @@ export class IsolatedWorld extends Realm {
...
     }
-    if (this.#context === null) {
-      throw new Error(`Execution content promise is missing`);
-    }
-    return this.#context.valueOrThrow();
+    return this.#context;
+  }
+
+  async #waitForExecutionContext(): Promise<ExecutionContext> {
+    const result = await firstValueFrom(
```
Lightning00Blade, line 107: We can do something like bellow so that we don't need to check everywhere if the context already exits
```suggestion
     merge(
     fromEmitterEvent(),
     from(this.#context)
     ).pipe(
     filter(Boolean),
     raceWith()
     )
```

OrKoN: Nope, since await messed up the relative execution w.r.t to the caller code :D so we want to at least in most cases be synchronous :D

Lightning00Blade (approved):

PR #12430, opened 2024-05-13
fix: configure sandbox permissions for Chrome on Windows

### LINKED ISSUE #12471: Read this if you experience PDF printing hanging on Windows
With M125 Chrome uses sandboxes for printing (https://crbug.com/338553158) which may require additional permissions on the Chrome installation folder:

See https://pptr.dev/troubleshooting#chrome-reports-sandbox-errors-on-windows for how to set permissions.

OrKoN:
Configure permissions by invoking setup.exe.

Closes https://github.com/puppeteer/puppeteer/issues/12471

### DIFF
--- a/.github/workflows/canary.yml
+++ b/.github/workflows/canary.yml
@@ -57,9 +57,6 @@ jobs:
         run: npm ci
         env:
           PUPPETEER_BROWSER_REVISION: canary
-      - name: Configure permissions
-        if: ${{ matrix.os == 'windows-latest' }}
-        run: icacls $HOME/.cache/puppeteer/chrome /grant "ALL APPLICATION PACKAGES:(OI)(CI)(RX)"
       - name: Apply Canary expectations
         run: node tools/merge-canary-test-expectations.mjs
       - name: Run all tests (for non-Linux)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -198,9 +198,6 @@ jobs:
           key: Chrome-${{ runner.os }}-${{ hashFiles('packages/puppeteer-core/src/revisions.ts') }}-${{ hashFiles('packages/puppeteer/src/node/install.ts') }}
       - name: Install Chrome
         run: npm run postinstall
-      - name: Configure permissions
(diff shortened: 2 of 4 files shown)


packages/browsers/src/fileUtil.ts
```diff
@@ -89,3 +89,21 @@ async function installDMG(dmgPath: string, folderPath: string): Promise<void> {
     await exec(`hdiutil detach "${mountPath}" -quiet`);
   }
 }
+
+export async function configureWindowsSandboxPermissionsForChrome(
```
Lightning00Blade: This seems out of place in this file, it does not have relation to files.

OrKoN: it sets file permissions

Lightning00Blade (approved):

OrKoN:
@Lightning00Blade FYI I have updated this to use setup.exe but it should start working only with M127

Lightning00Blade (approved):

Lightning00Blade (approved):

kblok:
@OrKoN does chromium have the same issue?
I have people reporting that PDF functions hang on Chromium.

OrKoN:
I believe it might but we are not testing with Chromium anymore. Does the same workaround help?

kblok:
@OrKoN, the chromium zip file doesn't have a `setup.exe`, so we won't be able to do this.

Shouldn't we communicate this whether as a validation or in some doc?

OrKoN:
@kblok the workaround without setup.exe is documented here https://pptr.dev/troubleshooting#chrome-reports-sandbox-errors-on-windows

OrKoN:
we might also file a request to include setup.exe into Chromium, let me look into that

kblok:
> @kblok the workaround without setup.exe is documented here https://pptr.dev/troubleshooting#chrome-reports-sandbox-errors-on-windows

Ah yeah. I haven't tried that.

OrKoN:
let's see if the reviewers have any comments https://chromium-review.googlesource.com/c/chromium/src/+/5764985

PR #12433, opened 2024-05-13
test: enable permission tests

OrKoN:


lutien:
FYI, We moved the support for "userContext" argument into [a separate bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1894217), so the tests should start passing after this bug will be landed :)

Lightning00Blade (approved):

lutien:
The support for `userContext` should be in Nightly now

PR #12451, opened 2024-05-16
docs: fix screenshot guide

OrKoN:
Closes https://github.com/puppeteer/puppeteer/pull/12429


Lightning00Blade (commented):
This is correct fix but as pointed by the other PR the extension in the example is wrong as well.
There it's `.pdf` which I'm not sure if it work as we try to infer the type from the it.
We should fix that as well.

Lightning00Blade (approved):

PR #12453, opened 2024-05-16
test: unskip tests for a11y selector for Firefox

lutien:

Updating test expectations to unskip tests for a11y selector for Firefox with bidi


OrKoN (approved):
Thanks!

PR #12454, opened 2024-05-16
test: use spec compliant spelling for aria-labelledby attribute

lutien:
Update a unit test to use spec compliant spelling for `aria-labelledby` attribute (see: https://w3c.github.io/aria/#aria-labelledby). It looks like Chrome supports both options, but Firefox only this one.


OrKoN (approved):
Thanks!

PR #12457, opened 2024-05-17
test: add tree element around treeitems elements

lutien:
It looks like elements with role `treeitem` have to be accessibility children of an element with role `tree` (see https://w3c.github.io/aria/#treeitem)


OrKoN (approved):
Thanks!

PR #12459, opened 2024-05-17
feat: support running Puppeteer in extensions

OrKoN:
This PR adds documentation for running Puppeteer inside a Chrome extension using chrome.debugger for communication with CDP.

It also marks the ExtensionTransport  public. 

It also changes how Puppeteer stringifies JS functions: if CSP prevents the use of `new Function()` we assume the function is serializable.

### DIFF
--- a/docs/guides/running-puppeteer-in-extensions.md
+++ b/docs/guides/running-puppeteer-in-extensions.md
@@ -1,0 +1,71 @@
+# Running Puppeteer in Chrome extensions
+
+:::caution
+
+Chrome extensions environment is significantly different from the usual Node.JS environment, therefore, the support for running Puppeteer in chrome.debugger
+is currently experimental. Please submit issues https://github.com/puppeteer/puppeteer/issues/new/choose if you encounted bugs.
+
+:::
+
+Chrome Extensions allow accessing Chrome DevTools Protocol via [`chrome.debugger`](https://developer.chrome.com/docs/extensions/reference/api/debugger).
+[`chrome.debugger`](https://developer.chrome.com/docs/extensions/reference/api/debugger) provides a restricted access to CDP and allows attaching to one
+page at a time. Therefore, Puppeteer requires a different transport to be used and Puppeteer's view is limited to a single page. It means you can
+interact with a single page and its frames and workers but cannot create new pages using Puppeteer. To create a new page you need to use the
+[`chrome.tabs`](https://developer.chrome.com/docs/extensions/reference/api/tabs) API and establish a new Puppeteer connection.
+
+## How to run Puppeteer in the browser
+
+:::note
+
+See https://github.com/puppeteer/puppeteer/tree/main/examples/puppeteer-in-extension for a complete example.
+
+:::
+
+To run Puppeteer in the an extension, first you need to produce a browser-compatible build using a bundler such as rollup or webpack:
+
+1. When importing Puppeteer use the browser-specific entrypoint from puppeteer-core `puppeteer-core/lib/esm/puppeteer/puppeteer-core-browser.js'`:
+
+```ts
+import {
+  connect,
+  ExtensionTransport,
+} from 'puppeteer-core/lib/esm/puppeteer/puppeteer-core-browser.js';
+
+// Create a tab or find a tab to attach to.
+const tab = await chrome.tabs.create({
+  url,
+});
+// Connect Puppeteer using the ExtensionTransport.connectTab.
+const browser = await connect({
+  transport: await ExtensionTransport.connectTab(tab.id),
+});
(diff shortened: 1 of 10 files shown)


docs/guides/running-puppeteer-in-extensions.md
```diff
@@ -0,0 +1,69 @@
+# Running Puppeteer in Chrome extensions
```
OrKoN, line 1: @oliverdunk PTAL

docs/guides/running-puppeteer-in-extensions.md
```diff
@@ -0,0 +1,69 @@
...
+1. When importing Puppeteer use the browser-specific entrypoint from puppeteer-core `puppeteer-core/lib/esm/puppeteer/puppeteer-core-browser.js'`:
+
+```ts
+import {
+  connect,
+  ExtensionTransport,
+} from 'puppeteer-core/lib/esm/puppeteer/puppeteer-core-browser.js';
+
+// Create a tab or find a tab to attach to.
+const tab = await chrome.tabs.create({
```
oliverdunk, line 35: `chrome.tabs.create` resolves when the tab has been created, rather than when the page has finished loading. Should we listen for the chrome.tabs.onUpdated event to wait for it to load, or have some handling for that in CDP before reading the title?

Example:

```js
let expectedId;

function listener(tabId, changeInfo) {
  if (tabId === expectedId && changeInfo.status === "complete") {
    chrome.tabs.onUpdated.removeListener(listener);

    const browser = await connect({
      transport: await ExtensionTransport.connectTab(tab.id),
    });
    ...
  }
}

chrome.tabs.onUpdated.addListener(listener);

expectedId = (await chrome.tabs.create({ url })).id;
```

Not sure if this is simpler to handle by listening for CDP events.

docs/guides/running-puppeteer-in-extensions.md
```diff
@@ -0,0 +1,69 @@
...
+  output: {
+    format: 'esm',
+    dir: 'out',
+  },
+  plugins: [
+    nodeResolve({
+      // Indicate that we target a browser environment.
+      browser: true,
+      // Exclude any dependencies except for puppeteer-core.
+      resolveOnly: ['puppeteer-core'],
```
oliverdunk, line 67: It would be nice to call out that you need to install this :)

docs/guides/running-puppeteer-in-extensions.md
```diff
@@ -0,0 +1,69 @@
...
+});
+// Connect Puppeteer using the ExtensionTransport.connectTab.
+const browser = await connect({
+  transport: await ExtensionTransport.connectTab(tab.id),
+});
+// You will have a single page on the browser object, which corresponds
+// to the tab you connected the transport to.
+const [page] = await browser.pages();
+// Perform the usual operations with Puppeteer page.
+console.log(await page.evaluate('document.title'));
```
oliverdunk, line 46: Should we disconnect at the end here?

docs/guides/running-puppeteer-in-extensions.md
oliverdunk: Thanks! Generally looks great, I left a couple of comments.

oliverdunk (changes requested):

docs/guides/running-puppeteer-in-extensions.md
OrKoN: I believe Puppeteer does not require it to be loaded, as soon as the target exists, we can connect to it. Puppeteer will detect that is has been loaded.

oliverdunk: I don't think it fails, but when testing with https://example.com, I was getting an empty string for the title rather than what I expected. Do you see the same?

OrKoN: No, I didn't see that, I was getting the correct title.

examples/puppeteer-in-extension/background.js
```diff
@@ -16,5 +16,7 @@ globalThis.testConnect = async url => {
     transport: await ExtensionTransport.connectTab(tab.id),
   });
   const [page] = await browser.pages();
-  return await page.evaluate('document.title');
+  const title = await page.evaluate('document.title');
+  browser.disconnect();
```
Lightning00Blade: Nit: Should we await this?

docs/guides/running-puppeteer-in-extensions.md
OrKoN: Added a step to wait for the tab to load

Lightning00Blade (approved):
LGTM

PR #12475, opened 2024-05-21
test: add doctype to some screenshot tests to avoid running them in quirks mode

lutien:
Adds DOCTYPE to the content of some screenshot tests to avoid running them in quirks mode


OrKoN (approved):
Thanks!

PR #12499, opened 2024-05-28
chore(webdriver): don't swallow errors for `session.new`

Lightning00Blade:


test/src/fixtures.spec.ts
```diff
@@ -101,9 +101,10 @@ describe('Fixtures', function () {
           wsEndPointCallback(output.substring(0, output.indexOf('\n')));
         }
       });
-      using browser = await puppeteer.connect({
+      const connectOptions = Object.assign({}, defaultBrowserOptions, {
```
Lightning00Blade, line 104: Test was using 2 protocols when using `webDriverBiDi`.

OrKoN (approved):

PR #12500, opened 2024-05-28
docs: various documentation updates

OrKoN:
Mostly updating the selector documentation and copying it in all APIs so that it is easy to find. And adding a deprecation notice to isOOPIFFrame method.

### DIFF
--- a/docs/api/puppeteer.page.__eval.md
+++ b/docs/api/puppeteer.page.__eval.md
@@ -4,7 +4,7 @@ sidebar_label: Page.$$eval
 
 # Page.$$eval() method
 
-This method runs `Array.from(document.querySelectorAll(selector))` within the page and passes the result as the first argument to the `pageFunction`.
+This method returns all elements matching the selector and passes the resulting array as the first argument to the `pageFunction`.
 
 #### Signature:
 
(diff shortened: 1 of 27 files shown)


docs/api/puppeteer.page.__eval.md
```diff
@@ -4,7 +4,7 @@ sidebar_label: Page.$$eval
 
 # Page.$$eval() method
 
-This method runs `Array.from(document.querySelectorAll(selector))` within the page and passes the result as the first argument to the `pageFunction`.
+This method returns all elements matching the selector and passes the resulting array as the first argument to the`pageFunction`.
```
Lightning00Blade: Nit:
```suggestion
This method returns all elements matching the selector and passes the resulting array as the first argument to the `pageFunction`.
```

docs/api/puppeteer.page.authenticate.md
```diff
@@ -6,6 +6,12 @@ sidebar_label: Page.authenticate
 
 Provide credentials for `HTTP authentication`.
 
+:::note
```
Lightning00Blade, line 9: This is true for CDP as well. Why do we need to mention an implementation detail?

docs/api/puppeteer.page.exposefunction.md
```diff
@@ -8,6 +8,12 @@ The method adds a function called `name` on the page's `window` object. When cal
 
 If the puppeteerFunction returns a `Promise`, it will be awaited.
 
+:::caution
+
+The function will only be installed on the main frame and not child frames.
```
Lightning00Blade: Is this correct? We do `Page.addScriptToEvaluateOnNewDocument` and `evaluate` on each current frame

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1570,15 +1648,14 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
    * - the remote server does not respond or is unreachable.
-   * - the main resource failed to load.
    *
-   * This method will not throw an error when any valid HTTP status code is
-   * returned by the remote server, including 404 "Not Found" and 500 "Internal
-   * Server Error". The status code for such responses can be retrieved by
-   * calling {@link HTTPResponse.status}.
+   * - the main resource failed to load.
    */
   async goto(url: string, options?: GoToOptions): Promise<HTTPResponse | null> {
```
Lightning00Blade, line 1615: Can you try: It should be 1:1 that way we don't duplicate the comments
```
{@inheritDoc Frame.goto}
```

docs/api/puppeteer.page.exposefunction.md
OrKoN: Right, probably not completely accurate but addScriptToEvaluateOnNewDocument does not work for OOPIFs, only called for local frames.

docs/api/puppeteer.page.authenticate.md
OrKoN: It might affect performance so I think it's worth leaving a note.

Lightning00Blade (approved):

PR #12501, opened 2024-05-29
ci: enable Firefox test that are now passing

Lightning00Blade:
With the latest nightly Firefox supports `headers` in interception hence the `setUserAgent` test are now passing. 
Which leaves 2 test at https://puppeteer.github.io/ispuppeteerwebdriverbidiready/
CC @whimboo @juliandescottes

### DIFF
--- a/test/src/requestinterception-experimental.spec.ts
+++ b/test/src/requestinterception-experimental.spec.ts
@@ -178,7 +178,7 @@ describe('cooperative request interception', function () {
       page.on('request', request => {
         const headers = Object.assign({}, request.headers(), {
           foo: 'bar',
-          origin: undefined, // remove "origin" header
+          accept: undefined, // remove "origin" header
         });
         void request.continue({headers}, 0);
       });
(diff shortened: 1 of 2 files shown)


test/src/requestinterception-experimental.spec.ts
```diff
@@ -178,7 +178,7 @@ describe('cooperative request interception', function () {
       page.on('request', request => {
         const headers = Object.assign({}, request.headers(), {
           foo: 'bar',
-          origin: undefined, // remove "origin" header
+          accept: undefined, // remove "origin" header
```
Lightning00Blade, line 181: `origin` header was not being set by the request. 
`accept` should in most cases be there.

Lightning00Blade:
```suggestion
          accept: undefined, // remove "accept" header
```

OrKoN (approved):

test/src/requestinterception-experimental.spec.ts
```diff
@@ -178,7 +178,7 @@ describe('cooperative request interception', function () {
       page.on('request', request => {
         const headers = Object.assign({}, request.headers(), {
           foo: 'bar',
-          origin: undefined, // remove "origin" header
+          accept: undefined, // remove "accept" header
```
Lightning00Blade, line 181: The `origin` header was not being set for the request.
`accept` should be there in most cases.

PR #12502, opened 2024-05-29
ci: enable more Firefox tests that are now passing and fix others

whimboo:


test/src/requestinterception-experimental.spec.ts
```diff
@@ -341,9 +341,13 @@ describe('cooperative request interception', function () {
...
       expect(failedRequest).toBeTruthy();
-      expect(failedRequest.failure()!.errorText).toBe(
-        'net::ERR_INTERNET_DISCONNECTED'
-      );
+      if (isChrome) {
+        expect(failedRequest.failure()!.errorText).toBe(
+          'net::ERR_INTERNET_DISCONNECTED'
+        );
+      } else {
+        expect(failedRequest.failure()!.errorText).toBe('NS_ABORT_ERROR');
```
Lightning00Blade: This is wrong as BiDi does not support custom error like the test tries to test.

test/src/requestinterception.spec.ts
```diff
@@ -258,9 +258,13 @@ describe('request interception', function () {
       ]);
 
       expect(failedRequest).toBeTruthy();
-      expect(failedRequest.failure()!.errorText).toBe(
-        'net::ERR_INTERNET_DISCONNECTED'
-      );
+      if (isChrome) {
+        expect(failedRequest.failure()!.errorText).toBe(
+          'net::ERR_INTERNET_DISCONNECTED'
```
Lightning00Blade: Same as above.

test/TestExpectations.json
```diff
@@ -3315,33 +3315,19 @@
...
-    "expectations": ["SKIP"],
-    "comment": "TODO: Needs full support for continueRequest in Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1850680"
-  },
   {
     "testIdPattern": "[requestinterception-experimental.spec] cooperative request interception Page.setRequestInterception should cooperatively respond by priority",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["firefox", "webDriverBiDi"],
     "expectations": ["SKIP"],
-    "comment": "TODO: Needs full support for continueRequest in Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1850680"
+    "comment": "TODO: Needs full support for continueResponse in Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1849104"
```
Lightning00Blade: We don't use that command for Puppeteer we use `network.provideResponse`.

Lightning00Blade (changes requested):

test/src/requestinterception-experimental.spec.ts
Lightning00Blade: WebDriver BiDi issue: https://github.com/w3c/webdriver-bidi/issues/508

whimboo: So should this test be skipped for both chrome and firefox, and ignored for now?

test/TestExpectations.json
whimboo: Thanks. Will update the name then.

test/src/requestinterception-experimental.spec.ts
Lightning00Blade: The test only fail on assertion so I think it's fine to mark them as `FAIL` and update the comment to something like: 
`BiDi does not support custom errors - https://github.com/w3c/webdriver-bidi/issues/508`

test/src/navigation.spec.ts
```diff
@@ -719,7 +719,7 @@ describe('navigation', function () {
           waitEvent(page, 'frameattached').then(_frame => {
             return (frame = _frame);
           }),
-          waitEvent(page, 'framenavigated', f => {
+          waitEvent(page, 'framenavigated').then(f => {
```
Lightning00Blade: Why was this changed?
This now does not wait for the correct frame to be navigated but any frame to be navigated.

test/src/network.spec.ts
```diff
@@ -555,7 +555,7 @@ describe('network', function () {
       if (isChrome) {
         expect(failedRequest.failure()!.errorText).toBe('net::ERR_FAILED');
       } else {
-        expect(failedRequest.failure()!.errorText).toBe('NS_ERROR_FAILURE');
+        expect(failedRequest.failure()!.errorText).toBe('NS_ERROR_ABORT');
```
OrKoN, line 558: Interesting, so CDP in Firefox sends a different code? I wonder if this would fail for the CDP-Firefox tests.

OrKoN (commented):
looks like CDP-Firefox tests are failing

test/src/navigation.spec.ts
whimboo: Oh, this extra argument acts like a filter. I wasn't aware of that. So I'll revert this part and check why the test is failing.

test/src/network.spec.ts
whimboo: The CDP implementation fails here anyway due to other issues.

Lightning00Blade (approved):

Lightning00Blade:
Thanks!

PR #12504, opened 2024-05-29
test: use trace instead of log

OrKoN:
See https://github.com/whatwg/console/issues/203

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -2968,20 +2968,6 @@
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
   {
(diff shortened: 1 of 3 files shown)


Lightning00Blade (approved):

OrKoN:
@Lightning00Blade PTAL again (found another test, and changed the tests more)

Lightning00Blade (approved):
Can we add a `caution` on the `.location()` method that says it's only guaranteed to work with `console.trace`?

OrKoN:
@Lightning00Blade it is not only trace but a few others but basically it is implementation-dependent. We should add some note but I think it is not breaking from the API perspective.

PR #12506, opened 2024-05-29
fix(webdriver): HTTPRequest redirect chain from first request

Lightning00Blade:


packages/puppeteer-core/src/bidi/HTTPRequest.ts
```diff
@@ -32,22 +32,22 @@ export class BidiHTTPRequest extends HTTPRequest {
...
+    const request = new BidiHTTPRequest(bidiRequest, frame, redirectChain);
     request.#initialize();
     return request;
   }
-  #redirectBy: BidiHTTPRequest | undefined;
   #response: BidiHTTPResponse | null = null;
   override readonly id: string;
   readonly #frame: BidiFrame;
   readonly #request: Request;
+  declare _redirectChain: BidiHTTPRequest[];
```
OrKoN: perhaps we could keep it private and instantiate it in the constructor:

1) if !redirectBy, set this.#redirectChain to []
2) if redirectBy, set this.#redirectChain to redirectBy.redirectChain() (without slicing, using the same instance of the chain)

OrKoN (approved):

Issue #11944, opened 2024-02-19
[Bug]: 3x performance drop since v21.2.0


feesler:
### Minimal, reproducible example

```TypeScript
git clone https://github.com/feesler/puppeteer-perf.git
cd puppeteer-perf
npm install
npm test
npm install puppeteer@21.2.0
npm test
```

### Error string

no error

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

Similar to #9993 and #8650.

Steps to reproduce the problem:

1. Install puppeteer v21.1.1 dependency
2. Run some script and measure performance
3. Update puppeteer to v21.2.0 and repeat step 2

Script performing simple page query/evaluate calls.

### Expectation

Expected same performance of query/evaluate calls as version 21.1.1 and earlier.

### Reality

Starting from version 21.2.0 performance of script with query/evaluate calls is dropped about 3 times and remains on the same level.
No difference between ESM and CJS.

**Log output:**
Node: v21.6.2
Puppeteer: 21.1.1
Browser: Chrome/116.0.5845.96
Duration: 218.91580000000022

Node: v21.6.2
Puppeteer: 21.2.0
Browser: Chrome/116.0.5845.96
Duration: 783.2916

Node: v21.6.2
Puppeteer: 22.1.0
Browser: Chrome/121.0.6167.85
Duration: 775.9939000000004

### Puppeteer version

21.2.0 - 22.1.0

### Node version

21.6.2

### Package manager

npm

### Package manager version

9.8.1

### Operating system

Windows

OrKoN:
@jrandolf PTAL

OrKoN:
Bisected to https://github.com/puppeteer/puppeteer/pull/10810/files

feesler:
Well, it will be sad if there is no good solution to this! Just wonder how to workaround this from user perspective.

Sample script is, of course, intentionally not optimized.
Moving calculations for all nodes inside single `page.evaluate` call saves the situation, but only once.
Over long test run it still adds up to an impressive amount of time.

NC-piercej:
I can confirm that `headless: true` on v22 is *significantly* slower than `headless: 'shell'` for our PDF generation workloads.

Interestingly, if we are just generating a single PDF in a single tab, the performance is similar. It's only when multiple tabs are open, running heavy CPU-bound javascript, and generating PDFs at once that performance degrades significantly.

[...]

PR #12539, opened 2024-06-05
fix: add a way to run page.$$ without isolation

OrKoN:
With this change, almost all of the performance regression is recovered. It requires the client to specify `isolate: false` when using `page.$$` to opt out of query isolation.

Related to https://github.com/puppeteer/puppeteer/issues/11944 #12566


OrKoN:
```
Node: v20.10.0
Puppeteer: 21.1.1
Browser: HeadlessChrome/116.0.5845.96
Duration: 40.403416991233826 (without isolation)
```

```
Node: v20.10.0
Puppeteer: 21.2.0
Browser: HeadlessChrome/116.0.5845.96
Duration: 128.20825004577637 (with isolation)
```

```
Node: v20.10.0
Puppeteer: 22.10.0
Browser: HeadlessChrome/125.0.6422.141
Duration: 187.64925003051758 (with isolation)
```

```
Node: v20.10.0
Puppeteer: 22.10.0
Browser: HeadlessChrome/125.0.6422.141
Duration: 79.55983293056488 (without isolation)
```

test/src/queryselector.spec.ts
```diff
@@ -153,6 +153,24 @@ describe('querySelector', function () {
     });
   });
 
+  describe('Page.$$s', function () {
```
Lightning00Blade: Let's move it under the `$$` tests and update the `it` text

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -371,15 +375,43 @@ export abstract class ElementHandle<
...
-      QueryHandler.queryAll(this, updatedSelector)
-    ) as Promise<Array<ElementHandle<NodeFor<Selector>>>>);
+    if (options?.isolate === false) {
+      const {updatedSelector, QueryHandler} =
+        getQueryHandlerAndSelector(selector);
+      return await (AsyncIterableUtil.collect(
+        QueryHandler.queryAll(this, updatedSelector)
+      ) as Promise<Array<ElementHandle<NodeFor<Selector>>>>);
+    }
+    // If the handle is already isolated, then we don't need to adopt it
```
Lightning00Blade: WDYT about have a private method and attaching the decorator to it then calling it here?

OrKoN: didn't get the idea with the private method but let's reuse the decorator

Lightning00Blade (approved):

PR #12541, opened 2024-06-06
test: ignore favicon requests in the network test for redirects

lutien:
For Firefox, the favicon requests sometimes happen before the navigation command returns to avoid intermittent failures, let just ignore the favicon requests.

### DIFF
--- a/test/src/network.spec.ts
+++ b/test/src/network.spec.ts
@@ -599,16 +599,18 @@ describe('network', function () {
 
       const events: string[] = [];
       page.on('request', request => {
-        return events.push(`${request.method()} ${request.url()}`);
+        !isFavicon(request) &&
+          events.push(`${request.method()} ${request.url()}`);
       });
       page.on('response', response => {
-        return events.push(`${response.status()} ${response.url()}`);
+        !isFavicon(response) &&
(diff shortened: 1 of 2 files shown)


test/src/network.spec.ts
```diff
@@ -599,16 +599,28 @@ describe('network', function () {
 
       const events: string[] = [];
       page.on('request', request => {
-        return events.push(`${request.method()} ${request.url()}`);
+        // Ignore default favicon requests.
+        if (!request.url().endsWith('favicon.ico')) {
```
OrKoN: let's use the isFavicon helper

lutien: Oh, right, that's pretty handy, updated ✅

test/src/network.spec.ts
```diff
@@ -599,16 +599,22 @@ describe('network', function () {
 
       const events: string[] = [];
       page.on('request', request => {
-        return events.push(`${request.method()} ${request.url()}`);
+        return (
```
OrKoN: nit: since we are here we can drop all returns too

OrKoN (approved):

PR #12542, opened 2024-06-06
fix: align network conditions presets with DevTools

OrKoN:
The PR adds an alias called 'Slow 4G' that is the same as 'Fast 3G' and also adds a new preset 'Fast 4G'.


packages/puppeteer-core/src/cdp/PredefinedNetworkConditions.ts
```diff
@@ -30,13 +30,37 @@ import type {NetworkConditions} from './NetworkManager.js';
  */
 export const PredefinedNetworkConditions = Object.freeze({
```
sadym-chromium, line 43: It would be helpful here to have a link to the DevTools source code in the comment here.

sadym-chromium (approved):

PR #12543, opened 2024-06-06
refactor: move selector parsing to Node

OrKoN:
- properly handled third party modules
- allows adjusting Puppeteer logic based on selector parsing later
- minimized injected code


Lightning00Blade (approved):
LGTM

PR #12545, opened 2024-06-06
fix: waitForSelector should work for pseudo classes

### LINKED ISSUE #6830: waitForSelector("…:focus") doesn’t resolve on focus change

OrKoN:
Closes #6830


packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -720,13 +721,12 @@ export abstract class Frame extends EventEmitter<FrameEvents> {
...
+    const {updatedSelector, QueryHandler, selectorHasPseudoClasses} =
       getQueryHandlerAndSelector(selector);
-    return (await QueryHandler.waitFor(
-      this,
-      updatedSelector,
-      options
-    )) as ElementHandle<NodeFor<Selector>> | null;
+    return (await QueryHandler.waitFor(this, updatedSelector, {
+      ...options,
+      polling: selectorHasPseudoClasses ? PollingOptions.RAF : undefined,
```
Lightning00Blade: We need change the order else the `polling` from the user is not respected.
Or we need instead of `undefined` -> `options.polling`
```suggestion
      polling: selectorHasPseudoClasses ? PollingOptions.RAF : undefined,
      ...options,
```
image

Lightning00Blade (approved):
LGTM with comment

packages/puppeteer-core/src/api/Frame.ts
OrKoN: Wait but WaitForSelectorOptions does not allow polling?

PR #12547, opened 2024-06-06
test: add a test for ctrlKey on wheel 

### LINKED ISSUE #12093: [Bug]: (Firefox) Wheel event doesn't have ctrlKey set to true when Control key is down

```TypeScript

(async () => {
    product: "firefox",
    headless: true,
    protocol: "webDriverBiDi",
  });

  const data = "data:text/html,<script>window.ctrlKey = false; window.addEventListener('wheel', e => { window.ctrlKey = e.ctrlKey; });</script>";

  await page.goto(data);

  await page.keyboard.down("Control");
  await page.mouse.wheel({ deltaY: -100 });
  await page.keyboard.up("Control");

  const ctrlKey = await page.evaluate(() => window.ctrlKey);
  if (!ctrlKey) {
    throw new Error("Control key was not pressed");
  }

  await page.close();
  await browser.close();
})();
```

### Error string

Control key was not pressed

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

_No response_

### Expectation

ctrlKey property of the wheel event must be `true`.

### Reality

It's `false`.

_No response_

### Puppeteer version

22.4.1

### Node version

18.19.1

### Package manager

npm

### Package manager version

10.2.4

### Operating system

Windows

OrKoN:
Closes #12093

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -2184,6 +2184,13 @@
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
   {
+    "testIdPattern": "[mouse.spec] Mouse should set ctrlKey on the wheel event",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["cdp", "firefox"],
+    "expectations": ["SKIP"],
+    "comment": "Wheel event is not supported"
+  },
+  {
     "testIdPattern": "[mouse.spec] Mouse should trigger hover state with removed window.Node",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["cdp", "firefox"],
--- a/test/src/mouse.spec.ts
+++ b/test/src/mouse.spec.ts
@@ -222,6 +222,30 @@ describe('Mouse', function () {
       height: 230,
     });
   });
+  it('should set ctrlKey on the wheel event', async () => {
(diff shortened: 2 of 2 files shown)


sadym-chromium (approved):

lutien:
To avoid problem with zoom, you could also use another key, in wpt tests we used shift, for example.

whimboo:
I investigated the issue with @OrKoN on Matrix. So yes we need to set this pref by default for the Remote Agent. @OrKoN will file a bug for it.

Maybe we can wait with landing this patch until we have fixed it in Firefox? Then we don't have to set it explicitly before starting Firefox.

OrKoN:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1901211

OrKoN:
> I investigated the issue with @OrKoN on Matrix. So yes we need to set this pref by default for the Remote Agent. @OrKoN will file a bug for it.
> 
> Maybe we can wait with landing this patch until we have fixed it in Firefox? Then we don't have to set it explicitly before starting Firefox.

I have updated the patch to work around the issue in a different way. Once https://bugzilla.mozilla.org/show_bug.cgi?id=1901211 and the setting is changed in the remote agent, we can update the test.

PR #12561, opened 2024-06-10
fix(performance): use Runtime.getProperties for improved performance

OrKoN:
Related to https://github.com/puppeteer/puppeteer/issues/11944


Lightning00Blade (approved):
LGTM

PR #12567, opened 2024-06-11
chore(build): update TypeScript to 5.4

Lightning00Blade:
Breaking change on the Decorator Proposal  - https://github.com/tc39/proposal-decorators?tab=readme-ov-file#adding-initialization-logic-with-addinitializer 
Not documented on the Changelog as it's experimental feature. 
The new order is 
`init()` -> `context.addInitializer()`.

### DIFF
--- a/packages/puppeteer-core/src/util/decorators.test.ts
+++ b/packages/puppeteer-core/src/util/decorators.test.ts
@@ -87,20 +87,19 @@ describe('decorators', function () {
       }
 
       const t = new Test();
-      let a = false;
-      t.on('a', (value: boolean) => {
-        a = value;
-      });
+      const spy = sinon.spy();
+      t.on('a', spy);
 
       t.field.emit('a', true);
-      expect(a).toBeTruthy();
+      expect(spy.callCount).toBe(1);
+      expect(spy.calledWithExactly(true)).toBeTruthy();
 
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/util/decorators.ts
```diff
@@ -194,9 +175,27 @@ export function bubble<T extends EventType[]>(events?: T) {
...
+                }
+              }
+            : (type: EventType, event: unknown) => {
+                this.emit(type, event);
+              };
+
+        handlers.set(events, handler);
+        bubbleHandlers.set(this, handlers);
+
+        emitter.on('*', handler as any);
```
Lightning00Blade, line 206: Our event for `*` is wrongly typed

packages/puppeteer-core/src/util/decorators.test.ts
```diff
@@ -87,20 +87,19 @@ describe('decorators', function () {
       }
 
       const t = new Test();
-      let a = false;
-      t.on('a', (value: boolean) => {
-        a = value;
-      });
+      const spy = sinon.spy();
```
Lightning00Blade, line 90: Use spy in test to make sure we don't emit multiple times.

OrKoN (approved):

PR #12579, opened 2024-06-12
fix: include iframes into the a11y snapshot

### LINKED ISSUE #7744: [Bug]: Accessibility snapshot does not include iframe contents
### Bug description

1. Create a page with an iframe, in the iframe place a page with a button
2. Call `console.log(await page.accessibility.snapshot())`
3. Observe that the snapshot does not include the iframe contents on the `role: 'Iframe'` node
4. Execute the following:
```javascript
const button = await page.mainFrame().childFrames()[0].$('button');
console.log(await page.accessibility.snapshot({ root: button });
```
5. Observe that the console log result is `null`

Here's a sample output from my app:

```javascript
{
  role: 'WebArea',
  name: 'Empty Page',
  children: [
    { role: 'link', name: 'Apps' },
    { role: 'heading', name: 'Site navigation', level: 2 },
    { role: 'menubar', name: 'Site navigation', children: [Array] },
    { role: 'heading', name: 'Quick navigation', level: 2 },
    { role: 'button', name: 'Search', pressed: false },
    {
      role: 'menuitem',
      name: '2 unread notifications Show notifications menu',
      haspopup: 'menu'
    },
    { role: 'menuitem', name: 'Help menu', haspopup: 'menu' },
    { role: 'menuitem', name: 'Profile menu', haspopup: 'menu' },
    { role: 'Iframe', name: 'Content' }
  ]
}
```

Note that the last item with `role: 'Iframe'` has no `children` array. This does not change if you add the `interestingOnly: false` option. Additionally, if you try to call `snapshot()` with `root` set to any element handle inside the iframe, it will return `null`. This means, unless I'm missing something in the API, it's not possible to get an accessibility snapshot of any iframe, as `accessibility` does not exist on the `Frame` class. However, it is possible to use ARIA queries to find things inside the iframe, as long as the query is performed from one of the target's ancestors up to and including the iframe handle (which also feels wrong to me that I can't query any AX node from the top frame).

It looks like there are two problems [here](https://github.com/puppeteer/puppeteer/blob/f863f4bfe015e57ea1f9fbb322f1cedee468b857/src/common/Accessibility.ts#L185-L200). One is that the result of CDP's `Accessibility.getFullAXTree` does not contain iframe content and the other is that, when `root` is specified, Puppeteer still calls `Accessibility.getFullAXTree` instead of `Accessibility.getPartialAXTree`.

If I copy the accessibility tree from `chrome://accessibility`, the full tree is returned. If I inspect the accessibility tree for an element in the dev tools there is no distinction of an iframe or page boundary, the hierarchy goes all the way up to the main frame root. My understanding is that the accessibility tree is a merged tree containing all nodes for the current window, so I think the fix should be to expand iframe nodes by either adding an option to merge trees upstream at `Accessibility.getFullAXTree` or by recursing the nodes and calling `Accessibility.getFullAXTree` again on any iframe nodes. Additionally, calling `Accessibility.getPartialAXTree` when a `root` is specified might be appropriate to improve performance.

### Puppeteer version

v10.2.0

### Node.js version

v16.12.0

### npm version

v8.1.0

macOS

### Relevant log output

_No response_

OrKoN:
This PR adds querying for iframe a11y tree snapshots as part of the `page.accessibility.snapshot()` if `includeIframes` option is set to true.

Closes https://github.com/puppeteer/puppeteer/issues/7744

### DIFF
--- a/docs/api/puppeteer.snapshotoptions.md
+++ b/docs/api/puppeteer.snapshotoptions.md
@@ -35,6 +35,27 @@ Default
 </th></tr></thead>
 <tbody><tr><td>
 
+<span id="includeiframes">includeIframes</span>
+
+</td><td>
+
+`optional`
+
+</td><td>
+
+boolean
+
+</td><td>
+
+If true, gets accessibility trees for each of the iframes in the frame subtree.
+
+</td><td>
+
+`false`
+
+</td></tr>
+<tr><td>
+
 <span id="interestingonly">interestingOnly</span>
 
 </td><td>
--- a/packages/puppeteer-core/src/bidi/Frame.ts
+++ b/packages/puppeteer-core/src/bidi/Frame.ts
@@ -115,7 +115,7 @@ export class BidiFrame extends Frame {
         this,
       ),
     };
-    this.accessibility = new Accessibility(this.realms.default);
+    this.accessibility = new Accessibility(this.realms.default, this._id);
   }
 
   #initialize(): void {
--- a/packages/puppeteer-core/src/cdp/Accessibility.ts
+++ b/packages/puppeteer-core/src/cdp/Accessibility.ts
@@ -100,6 +100,13 @@ export interface SnapshotOptions {
    */
   interestingOnly?: boolean;
   /**
+   * If true, gets accessibility trees for each of the iframes in the frame
+   * subtree.
+   *
+   * @defaultValue `false`
+   */
+  includeIframes?: boolean;
+  /**
    * Root node to get the accessibility tree for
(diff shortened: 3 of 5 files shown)


mathiasbynens (approved):
LGTM. Technically this is a breaking change, so let’s mark it as such in the commit message?

OrKoN:
After some consideration, I swapped the default to match the current behavior.

mathiasbynens:
> After some consideration, I swapped the default to match the current behavior.

SGTM!

PR #12580, opened 2024-06-12
feat: support --install-deps flag in the browsers CLI

OrKoN:
The new flag will attempt installing system dependencies required for Chrome using `apt-get`. Currently, only Debian and Ubuntu are supported.

The Docker image is updated to install dependencies using the new flag. chrome-stable installation is, therefore, removed since it is not needed for installing system dependencies.

### DIFF
--- a/packages/browsers/src/install.ts
+++ b/packages/browsers/src/install.ts
@@ -6,7 +6,7 @@
 
 import assert from 'assert';
 import {spawnSync} from 'child_process';
-import {existsSync} from 'fs';
+import {existsSync, readFileSync} from 'fs';
 import {mkdir, unlink} from 'fs/promises';
 import os from 'os';
 import path from 'path';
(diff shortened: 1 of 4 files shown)


packages/browsers/src/install.ts
```diff
@@ -98,6 +98,17 @@ export interface InstallOptions {
    * @defaultValue `false`
    */
   forceFallbackForTesting?: boolean;
+
+  /**
+   * Whether to attempt to install system-level dependencies required
+   * for the browser.
+   *
+   * Currently, only supported for Debian and Ubuntu and only for Chrome.
```
Lightning00Blade: I think we should try to "say" the Browser first as that is more important than the Platform.
```suggestion
   * Only supported for Chrome on Debian or Ubuntu.
```

Lightning00Blade (approved):
LGTM

OrKoN:
@Lightning00Blade do you think it needs to be a breaking change due to the dockerfile changes?

Lightning00Blade (approved):
I don't think it's a breaking change, we expect node to be installed due to the Image.
Then all the rest of the set up is implementation thus we can change it. 
If we previously installed packages that user were using it's like using internal methods, they may be remove/change/etc. without notice.

PR #12599, opened 2024-06-17
build: explicitly set typeRoots

OrKoN:
This allows building Puppeteer while having other node_modules in the parent directories. See https://phabricator.services.mozilla.com/D213871#7341451

The default behavior is to walk the filesystem tree in search of node_modules https://www.typescriptlang.org/tsconfig/#typeRoots


Lightning00Blade:
Please update the `changed-packages.yml` and add to all project `tsconfig.base.json`.

Lightning00Blade (approved):

PR #12627, opened 2024-06-20
test: Update expectations for firefox support of provideResponse and originalOpenerId

juliandescottes:
Needs to wait until https://bugzilla.mozilla.org/show_bug.cgi?id=1853882 lands.
Summary:
- add missing mock-binary-response.png
- update multiple headers test to handle Firefox, which uses `,` as the default separator for header values
- update various expectations


OrKoN:
Could you please specify which test is it? and let's wait till it starts to pass on the PR?

juliandescottes:
The test is [requestinterception.spec] request interception Request.respond should allow mocking binary responses

juliandescottes:
Fine with me to wait until the test passes to accept the PR, I'll just apply the change locally for now. I guess I can keep the PR and update the test expectations at the same time.

test/src/requestinterception.spec.ts
```diff
@@ -867,7 +867,14 @@ describe('request interception', function () {
       });
       expect(response.status()).toBe(200);
       expect(response.headers()['foo']).toBe('bar');
-      expect(response.headers()['arr']).toBe('1\n2');
+
+      // The separator used to handle headers with multiple values is browser
```
OrKoN, line 871: Do you have a link to a spec that defines how it is supposed to be handled?

OrKoN (approved):

test/src/requestinterception.spec.ts
juliandescottes: I thought I replied here yesterday, sorry.

So the main source I have is https://www.rfc-editor.org/rfc/rfc9110.html#section-5.2
Which explains that combined header values should be CSV. However it allows for exceptions such as for the Set-Cookie header (see the note in https://www.rfc-editor.org/rfc/rfc9110.html#section-5.3)

I think Firefox happens to use \n separators for Set-Cookie since `,` can't be used. However I don't know why Chrome would use \n all the time, it doesn't seem spec compliant?

OrKoN: Interesting, in DevTools I see that Chrome also uses a comma

OrKoN: Would you mind leaving a note in the test that Chrome is likely not compliant. I have briefly investigated but I could not figure out where it goes wrong.

juliandescottes: Updated!

juliandescottes:
Bug  https://bugzilla.mozilla.org/show_bug.cgi?id=1853882 should now be in Nightly

juliandescottes:
Note that CI will fail because we also need the changes from #12674 (which on its own fails CI as well because it misses the changes from this PR).

@OrKoN do you want use to do a combined PR with both?

juliandescottes:
Sounds good, thanks!

PR #12638, opened 2024-06-23
add interface ClientProvider for EmulationManager

minggeorgelei:
We can see from the Emulated State that It used ClientProvider interface as its parameter and we actually passed EmulationManager instance. So we better add interface ClientProvider to EmulationManager.

**Summary**

**Other information**


minggeorgelei:
failed attempt

PR #12650, opened 2024-06-24
fix: reset the viewport after taking a fullPage screenshot if defaultViewport is null

minggeorgelei:
This is a bug fix. If the default viewport is set to null, and after that navigate to a website and take fullpage snapshot with fullpage parameter set to true, the viewport will be set to width equals 0/height equals 0 after snapshot, which will make UI cannot work correctly. This PR is to fix this problem.


minggeorgelei:
Hi @OrKoN , can you help to check why the test failed? Thanks.

OrKoN:
Not sure, it looks like it fails the test when Puppeteer runs in an extension so probably the change causes a failure there. It should be possible to run the tests locally with `npm run test-install` and enabling the specific test with `it.only`

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -344,19 +344,19 @@ export class BidiPage extends Page {
...
-        viewport:
-          viewport.width && viewport.height
-            ? {
-                width: viewport.width,
-                height: viewport.height,
-              }
-            : null,
-        devicePixelRatio: viewport.deviceScaleFactor
-          ? viewport.deviceScaleFactor
+        viewport: viewport
```
OrKoN: this should still check width and height since setting width and height to 0 resets the viewport

minggeorgelei:
Ok, I will check it

packages/puppeteer-core/src/bidi/Page.ts
minggeorgelei: I will fix it

OrKoN:
> This is a bug fix. If the default viewport is set to null, and after that navigate to a website and take fullpage snapshot with fullpage parameter set to true, the viewport will be set to width equals 0/height equals 0 after snapshot, which will make UI cannot work correctly. This PR is to fix this problem.

do you have a reproducible example for the issue or a new test case? according to https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setDeviceMetricsOverride settings the width and height to 0, disabled emulation which disables any emulation and it sounds like this is what should happen with defaultViewport = null.

OrKoN (changes requested):
Could you please add a test to demonstrate that the issue is fixed with this change?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -2563,14 +2565,7 @@ export abstract class Page extends EventEmitter<PageEvents> {
             ...scrollDimensions,
           });
           stack.defer(async () => {
-            if (viewport) {
-              await this.setViewport(viewport).catch(debugError);
-            } else {
-              await this.setViewport({
-                width: 0,
-                height: 0,
```
OrKoN, line 2572: I wonder if setting deviceScaleFactor here to 0 as well and reverting other changes would also fix the issue?

minggeorgelei: I will try it.

minggeorgelei: Set deviceScaleFactor to 0 does not fix the issue. I will later add a test case for the code change.

test/src/launcher.spec.ts
```diff
@@ -595,6 +595,30 @@ describe('Launcher specs', function () {
           await close();
         }
       });
+      it('should restore to original window size after taking fullPage screenshots when defaultViewport is null', async () => {
+        const {server, context, close} = await launch({
```
OrKoN: let's add this test to test/src/screenshot.spec.ts and use getTestState instead of launching a new browser

test/src/launcher.spec.ts
```diff
@@ -595,6 +595,30 @@ describe('Launcher specs', function () {
           await close();
         }
       });
+      it('should restore to original window size after taking fullPage screenshots when defaultViewport is null', async () => {
```
OrKoN:
```suggestion
      it('should restore to original viewport size after taking fullPage screenshots when defaultViewport is null', async () => {
```

test/src/launcher.spec.ts
OrKoN: I guess you need to have the launch() to specify defaultViewport to be null, so disregard the part about getTestState

minggeorgelei: yes I need to specify the defaultViewport

test/src/launcher.spec.ts
```diff
@@ -595,6 +595,30 @@ describe('Launcher specs', function () {
           await close();
         }
       });
+      it('should restore to original viewport size after taking fullPage screenshots when defaultViewport is null', async () => {
```
OrKoN: thanks for the test!

packages/puppeteer-core/src/cdp/EmulationManager.ts
```diff
@@ -253,6 +259,12 @@ export class EmulationManager {
     viewportState: ViewportState
   ): Promise<void> {
     if (!viewportState.viewport) {
```
OrKoN, line 265: let's add an early return here if the state is not active and call .setState two times if the viewport is null? first to reset it for existing targets with active: true, and the second time with active: true to avoid running clearDeviceMetricsOverride and setTouchEmulationEnabled for new targets?

minggeorgelei:
> I can reproduce the issue using the test you created although I am not sure why CDP does not behave as documented. I think the proposed solution with setting null is good but let's make sure we do not clear device metrics too often.

OK I will modify the code to avoid clear metrics too often.

test/TestExpectations.json
```diff
@@ -3258,6 +3258,13 @@
     "expectations": ["FAIL"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
+  {
+    "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Puppeteer.launch should restore to original viewport size after taking fullPage screenshots when defaultViewport is null",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["cdp", "firefox"],
+    "expectations": ["FAIL"],
+    "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
```
OrKoN: Could you please explain why it fails? I think it should not generally fail for Firefox here.

minggeorgelei: I just check the failed test and see the suggestion that adding such code can avoid failure.
I will revert the code.

minggeorgelei: Can you help to check why it is failed on firefox?

OrKoN: So Firefox does not support Emulation.clearDeviceMetricsOverride

OrKoN: Let me publish a change, I think it is okay to skip the test for Firefox CDP

minggeorgelei: Thanks

OrKoN (approved):
Thanks for fixing this!

PR #12663, opened 2024-06-25
fix: make sure bindings are working after a page is restored from bfcache

### LINKED ISSUE #12662: [Bug]: BFCache behavior breaks exposeFunction

```TypeScript
// taken from puppeteer test test/src/cdp/bfcache.spec.ts and augmented with exposeFunction

    const {httpsServer, page, close} = await launch({
      ignoreHTTPSErrors: true,
    });
    let message = '';
    try {
      page.setDefaultTimeout(3000);
      await page.exposeFunction('ping', (msg: string) => {
        message = msg;
      });
      await page.goto(httpsServer.PREFIX + '/cached/bfcache/index.html');

      await Promise.all([page.waitForNavigation(), page.locator('a').click()]);

      expect(page.url()).toContain('target.html');

      await Promise.all([page.waitForNavigation(), page.goBack()]);
      await page.evaluate(() => {
        (window as any).ping('hello');
      });
      expect(message).toBe('hello');
      expect(
        await page.evaluate(() => {
          return document.body.innerText;
        })
      ).toBe('BFCachednext');

    } finally {
      await close();
    }
```

### Error string

Invalid arguments: should be exactly one string.

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

i'm actually coming from playwright since we're using it to test our bfcache behavior. to allow finishing our tests in playwright, we expose a playwright "finishTest(assertionFailures: string[])" function that allows us to fail the test on playwright context from the test page. this breaks on puppeteer because it seems that the exposed function isn't properly exposed after a bf cache restore

### Expectation

the exposed function is called and executed properly, without errors

### Reality

an exception is thrown upon calling the exposed function from the browser context, and it's not executed on node context

_No response_

### Puppeteer version

22.12.0

### Node version

18.20.3

### Package manager

npm

### Package manager version

9.8.1

### Operating system

Linux

OrKoN:
When a page is restored from bfcache, the CDP bindings are restored and overwrite CDP wrappers. To avoid the issue, this PR changes Puppeteer to expose two global variables (one is the raw CDP binding and another one the wrapper intended for Puppeteer users). The wrapper processes non-string arguments and calls the raw CDP binding according to the internal protocol between the page and Puppeteer. For end users, this fix should not cause any changes except for an extra global function `puppeteer_$FunctionName` being added to the globalThis.

Fixes  https://github.com/puppeteer/puppeteer/issues/12662

### DIFF
--- a/packages/puppeteer-core/src/cdp/ExecutionContext.ts
+++ b/packages/puppeteer-core/src/cdp/ExecutionContext.ts
@@ -124,11 +124,11 @@ export class ExecutionContext
         'Runtime.addBinding',
         this.#name
           ? {
-              name: binding.name,
+              name: '_' + binding.name,
               executionContextName: this.#name,
             }
           : {
-              name: binding.name,
+              name: '_' + binding.name,
               executionContextId: this.#id,
             }
       );
--- a/packages/puppeteer-core/src/cdp/ExecutionContext.ts
+++ b/packages/puppeteer-core/src/cdp/ExecutionContext.ts
@@ -34,6 +34,7 @@ import type {IsolatedWorld} from './IsolatedWorld.js';
 import {CdpJSHandle} from './JSHandle.js';
 import {
   addPageBinding,
+  CDP_BINDING_PREFIX,
   createEvaluationError,
   valueFromRemoteObject,
 } from './utils.js';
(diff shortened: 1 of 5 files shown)


Lightning00Blade (approved):

OrKoN:
@Lightning00Blade FYI I am not sure about the fix here. We will be exposing the double amount of globals, it probably does not matter but I wonder if there could be undesired side effects.

Lightning00Blade:
@OrKoN If users were using, it should not matter as it's internal implementation, while exposing 2 globals is not ideal, we already have one so it's not a new behavior. We may consider adding some prefix or suffix to the name, but not necessary.

Lightning00Blade (approved):

PR #12672, opened 2024-06-27
docs: Reference Page.setDefaultTimeout for timeout option

abrom:

Documentation clarification

**Summary**

Per the implementation of `Page.pdf` (`Page.createPDFStream`), the `timeout` option will fallback to use the value assigned by `Page.setDefaultTimeout`. See https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/cdp/Page.ts#L1083

This updates the `PDFOptions` documentation to match

### DIFF
--- a/docs/api/puppeteer.pdfoptions.md
+++ b/docs/api/puppeteer.pdfoptions.md
@@ -376,6 +376,8 @@ number
 
 Timeout in milliseconds. Pass `0` to disable timeout.
 
+The default value can be changed by using [Page.setDefaultTimeout()](./puppeteer.page.setdefaulttimeout.md)
+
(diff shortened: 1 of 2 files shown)


OrKoN (changes requested):
Thanks for the PR but the documentation is generated based on comments in the source code. Could you update the corresponding interface (PDFOptions) in packages/puppeteer-core/src/common/PDFOptions.ts and run `npm run docs` to re-generate the markdown files?

abrom:
Done. Thanks for the heads up on process @OrKoN 👍

OrKoN (approved):

PR #12678, opened 2024-06-28
test: add cache tests using scripts instead of stylesheets

juliandescottes:

Duplicates two tests around cached requests which currently use cached stylesheets. The alternate versions use scripts instead. At the moment Firefox does not emit network event for CSS Loader cache hits, so cached stylesheet network events are missing (same for images). However other cached requests should be visible, eg documents and scripts, so it would be interesting to have some tests to reflect that.

This PR is not removing any test, so it will not increase the number of passes for Firefox. The proposal is to downgrade all the stylesheet-only tests to P3, since their counterpart using scripts seem to pass.

Apart from the expected failures linked to missing network events for stylesheets, we also have frequent failures for `[network.spec] network Page.authenticate should not disable caching ...`

In theory this should work with scripts, and seems to work locally when running in isolation. However on CI or during full runs it often fails.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -3050,13 +3050,6 @@
     "comment": "TODO: BiDi does not support custom errors - https://github.com/w3c/webdriver-bidi/issues/508"
   },
   {
-    "testIdPattern": "[requestinterception-experimental.spec] cooperative request interception Page.setRequestInterception should cooperatively respond by priority",
-    "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["firefox", "webDriverBiDi"],
-    "expectations": ["SKIP"],
-    "comment": "TODO: Needs support for arguments in network.provideResponse in Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1853882"
-  },
-  {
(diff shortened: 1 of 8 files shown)


juliandescottes:
@OrKoN This should be ready for review. We have two failed expectations for the tests which immediately pass with the support for setCacheEnabed, but I think we'll remove those directly in your PR.

test/TestExpectations.json
```diff
@@ -527,6 +541,20 @@
...
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["cdp", "firefox"],
+    "expectations": ["FAIL"],
+    "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
+  },
+  {
+    "testIdPattern": "[network.spec] network Response.fromCache should work for *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["chrome", "webDriverBiDi"],
+    "expectations": ["FAIL", "PASS"],
```
juliandescottes, line 555: Note: This used to be a FAIL, PASS for a single test "[network.spec] network Response.fromCache should work". Now that we are running two similar tests, I used a wildcard which moved the expectation up. I haven't checked if the script version is actually FAIL/PASS on Chrome though.

Issue #12652, opened 2024-06-24
In documentation all overloads should be on one page


OrKoN:
We should try using https://github.com/microsoft/rushstack/blob/6d52e3a640e323435e85b47b72ac785f1ef25cdc/libraries/api-extractor-model/src/items/ApiItem.ts#L193 and put all overloads on a single API reference page.

cc @Lightning00Blade

OrKoN:
Closed via https://github.com/puppeteer/puppeteer/pull/12679

PR #12679, opened 2024-06-28
docs: merge overloads

Lightning00Blade:
Ref: #12652

Please take a look and tell me what you think.
One thing I notice is that they are in declaration order - so more specific overloads come first.


docs/api/puppeteer.elementhandle.contentframe.md
```diff
@@ -44,3 +44,17 @@ this
 **Returns:**
 
 Promise&lt;[Frame](./puppeteer.frame.md)&gt;
+
+## ElementHandle.contentFrame() (Overload 1) method
```
OrKoN: @Lightning00Blade I think we could try to produce something similar to https://typedoc.org/example/functions/overloadedFunction.html#overloadedFunction I.e. we should put the parameter name and return type into the title of the of the overload. Any thoughts on how to achieve this?

OrKoN: Probably we can use ApiParameterListMixin.parameters and ApiReturnTypeMixin to generate a comma-separated list of params (without type names) + return type.

tools/docgen/src/custom_markdown_documenter.ts
```diff
@@ -227,6 +232,53 @@ export class MarkdownDocumenter {
...
+    if (
+      apiItem instanceof ApiDeclaredItem &&
+      apiItem.excerpt.text.length > 0 &&
+      ApiParameterListMixin.isBaseClassOf(apiItem) &&
+      ApiReturnTypeMixin.isBaseClassOf(apiItem) &&
+      apiItem.getMergedSiblings().length > 1
+    ) {
+      const overloadIndex = apiItem.overloadIndex - 1;
+      const overloadId =
+        overloadIndex === 0 ? `"overload"` : `"overload-${overloadIndex}"`;
```
OrKoN:
```suggestion
      const name = apiItem.displayName;
      const overloadIndex = apiItem.overloadIndex - 1;
      const overloadId =
        overloadIndex === 0 ? name : `${name}-${overloadIndex}`;
```

tools/docgen/src/custom_markdown_documenter.ts
```diff
@@ -227,6 +232,53 @@ export class MarkdownDocumenter {
...
+      output.appendNode(
+        new DocParagraph({configuration}, [
+          new DocHtmlStartTag({
+            configuration,
+            name: 'h2',
+            htmlAttributes: [
+              new DocHtmlAttribute({
+                configuration,
+                name: 'id',
+                value: overloadId,
```
OrKoN:
```suggestion
                value: `"${overloadId}"`,
```

OrKoN (commented):
One last suggestion

OrKoN (approved):

PR #12692, opened 2024-07-01
fix: unsettable input should be filtered

minggeorgelei:
Some input element like button, checkbox can not be set value and dispatch input and change event, so we need to filter them.

### DIFF
--- a/packages/puppeteer-core/src/api/locators/locators.ts
+++ b/packages/puppeteer-core/src/api/locators/locators.ts
@@ -438,6 +438,18 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
                 ]).has(el.type)
               ) {
                 return 'typeable-input';
+              } else if (
+                new Set([
+                  'color',
+                  'date',
+                  'datetime-local',
+                  'month',
+                  'range',
+                  'time',
+                  'week',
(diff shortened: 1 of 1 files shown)


OrKoN:
@minggeorgelei thanks for the PR, a few questions: how can we know if the list for settable-input is exhaustive? what if the element is a custom element with a value setter or some other element that is not on the list? Is there a specific issue that this change solves?

minggeorgelei:
Hi @OrKoN I just checked https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input, I think all the input types are listed in the link. I just saw some input types like button or checkbox they cannot be set value. If you call fill function on such input element(button checkbox, etc), It will go through successfully.

OrKoN:
as I mentioned there could be custom elements that support value, therefore, the default should be to try to set the value and dispatch the events. Perhaps it would be better to explicitly throw errors where we know for sure that the value cannot be set? Also, setting a [value is possible on a button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/button#button_with_a_value) but perhaps the locators should not try to do that.

minggeorgelei:
I know what you mean. How can we know for sure that the value cannot be set except for the type check?

PR #12693, opened 2024-07-01
fix!: remove whitespace normalization from a11y selectors

OrKoN:
Previously, Puppeteer incorrectly normalized whitespaces in ARIA selectors in a way that did not not allow distinguishing between the following two HTML structures:

- `<p>  text<p>`
- `<p><span>&nbsp;</span><span>&nbsp;</span>text<p>`

In the first case, the spaces are not part of the element's textual content, and the element should be found via `aria/text`.

In the second case, the spaces are part of the p element and the query `aria/\u00A0\u00A0text` should find it while `aria/text` should not. The whitespace normalization in Puppeteer would previously prevent searching for the element with whitespaces in the textual content.

Related: the step 2.F of https://www.w3.org/TR/accname-1.2/#computation-steps defines how child elements contribute to the parent element's computed name.

If you need the old behavior, apply `selector.replace(/ +/g, ' ').trim()` to your ARIA selector.

### DIFF
--- a/packages/puppeteer-core/src/cdp/AriaQueryHandler.ts
+++ b/packages/puppeteer-core/src/cdp/AriaQueryHandler.ts
@@ -21,10 +21,6 @@ const isKnownAttribute = (
   return ['name', 'role'].includes(attribute);
 };
 
-const normalizeValue = (value: string): string => {
-  return value.replace(/ +/g, ' ').trim();
-};
-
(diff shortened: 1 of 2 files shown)


Mansion-dotcom:
I need it

Lightning00Blade (approved):

PR #12694, opened 2024-07-01
test: add cache tests using scripts instead of stylesheets 

juliandescottes:

Adds cache tests in network.spec.ts which rely on a script, in addition to the existing test relying on stylesheets.
The new failures for Firefox/BiDi are:
- [network.spec] network Network Events Page.Events.RequestServedFromCache for stylesheet 
- [network.spec] network Response.fromCache should work for stylesheet
- [network.spec] network Page.authenticate should not disable caching for stylesheet
- [network.spec] network Page.authenticate should not disable caching for script

For the first two, given that the script version of the test works fine, I suggest to move them to P3 for now.
For the last two, it seems we still have intermittent failures with the script version of this test which we need to investigate.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -3057,13 +3057,6 @@
     "comment": "TODO: Needs support for enabling cache in BiDi without CDP https://github.com/w3c/webdriver-bidi/issues/582"
   },
   {
-    "testIdPattern": "[requestinterception-experimental.spec] cooperative request interception Page.setRequestInterception should cooperatively respond by priority",
-    "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["firefox", "webDriverBiDi"],
-    "expectations": ["SKIP"],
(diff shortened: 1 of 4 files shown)


juliandescottes:
cc @OrKoN

OrKoN (approved):

whimboo:
@OrKoN I assume we need to update the [ignored-tests list for ispuppeteerwebdriverbidiready ](https://github.com/puppeteer/ispuppeteerwebdriverbidiready/blob/main/ignored-tests.json)?

whimboo:
> @OrKoN I assume we need to update the [ignored-tests list for ispuppeteerwebdriverbidiready ](https://github.com/puppeteer/ispuppeteerwebdriverbidiready/blob/main/ignored-tests.json)?

I've updated already via https://github.com/puppeteer/ispuppeteerwebdriverbidiready/pull/57/.

PR #12697, opened 2024-07-02
feat: add last locator method

minggeorgelei:
Sometimes in the test we need to wait for several locators to locate elements in order to continue the test. It is more convenient to provide such a method.

### DIFF
--- a/packages/puppeteer-core/src/api/locators/locators.ts
+++ b/packages/puppeteer-core/src/api/locators/locators.ts
@@ -8,6 +8,7 @@ import type {
   OperatorFunction,
 } from '../../../third_party/rxjs/rxjs.js';
 import {
+  takeLast,
   EMPTY,
   catchError,
   defaultIfEmpty,
@@ -111,6 +112,18 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
     locators: Locators
   ): Locator<AwaitedLocator<Locators[number]>> {
     return RaceLocator.create(locators);
+  }
+
+  /**
+   * Trying to locate all the elements for locator array in parallel
+   * and return the last locator that locate element.
+   *
+   * @public
+   */
+  static last<Locators extends readonly unknown[] | []>(
+    locators: Locators
+  ): Locator<AwaitedLocator<Locators[number]>> {
+    return LastLocator.create(locators);
   }
 
   /**
@@ -1080,6 +1093,41 @@ export class RaceLocator<T> extends Locator<T> {
         return locator._wait(options);
       })
     );
+  }
+}
+
+/**
+ * @internal
+ */
+export class LastLocator<T> extends Locator<T> {
+  static create<T extends readonly unknown[]>(
+    locators: T
+  ): Locator<AwaitedLocator<T[number]>> {
+    const array = checkLocatorArray(locators);
+    return new LastLocator(array);
+  }
+
+  #locators: ReadonlyArray<Locator<T>>;
+
+  constructor(locators: ReadonlyArray<Locator<T>>) {
+    super();
+    this.#locators = locators;
+  }
+
+  override _clone(): LastLocator<T> {
+    return new LastLocator<T>(
+      this.#locators.map(locator => {
+        return locator.clone();
+      })
+    ).copyOptions(this);
+  }
+
+  override _wait(options?: Readonly<ActionOptions>): Observable<HandleFor<T>> {
+    return merge(
+      ...this.#locators.map(locator => {
+        return locator._wait(options);
+      })
+    ).pipe(takeLast(1));
   }
 }
 
--- a/packages/puppeteer-core/src/api/locators/locators.ts
+++ b/packages/puppeteer-core/src/api/locators/locators.ts
@@ -115,8 +115,9 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
   }
 
   /**
-   * Trying to locate all the elements for locator array in parallel
-   * and return the last locator that locate element.
+   * Waits for multiple locators to locate elements in DOM and 
+   * calls the action on the last locator waiting for action-specific preconditions.
+   * If any of the elements identified by locators is not found, the last locator times out.
    *
    * @public
    */
--- a/packages/puppeteer-core/src/api/locators/locators.ts
+++ b/packages/puppeteer-core/src/api/locators/locators.ts
@@ -115,9 +115,19 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
   }
 
   /**
-   * Waits for multiple locators to locate elements in DOM and 
+   * Waits for multiple locators to locate elements in DOM and
    * calls the action on the last locator waiting for action-specific preconditions.
-   * If any of the elements identified by locators is not found, the last locator times out.
+   * If any of the elements identified by locators is not found, the last locator
+   * times out.
+   * @example
+   *
+   * ```ts
+   * await Locator.last([
+   *   page.locator('#locator1'),
+   *   page.locator('#locator2'),
+   * ]).wait();
+   * await page.locator('#locator3').click();
+   * ```
    *
    * @public
    */
--- a/test/src/locator.spec.ts
+++ b/test/src/locator.spec.ts
@@ -581,6 +581,38 @@ describe('Locator', function () {
     });
   });
 
+  describe('Locator.last', () => {
+    it('should time out when any one of the locaters do not match', async () => {
+      const clock = sinon.useFakeTimers({
+        shouldClearNativeTimers: true,
+        shouldAdvanceTime: true,
+      });
+      try {
+        const {page} = await getTestState();
+
+        await page.setViewport({width: 500, height: 500});
+        await page.setContent(`
+        <button id="test1">test1</button>
+        <input id="test2">test2</input>
+      `);
+        const result = Locator.last([
+          page.locator('#test1'),
+          page.locator('#test2'),
+          page.locator('#test3'),
+        ])
+          .setTimeout(5000)
+          .waitHandle();
+        clock.tick(5100);
+        await expect(result).rejects.toEqual(
+          new TimeoutError('Timed out after waiting 5000ms')
+        );
+        clock.tick(5100);
+      } finally {
+        clock.restore();
+      }
+    });
+  });
+
   describe('Locator.prototype.map', () => {
     it('should work', async () => {
       const {page} = await getTestState();
--- a/test/src/locator.spec.ts
+++ b/test/src/locator.spec.ts
@@ -655,7 +655,7 @@ describe('Locator', function () {
 
       await page.setViewport({width: 500, height: 500});
       await page.setContent(`
-        <button id="test1">test1</button>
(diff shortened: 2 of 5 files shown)


minggeorgelei:
Hi @OrKoN , can you help to check this PR?

packages/puppeteer-core/src/api/locators/locators.ts
```diff
@@ -113,6 +114,18 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
     return RaceLocator.create(locators);
   }
 
+  /**
+   * Trying to locate all the elements for locator array in parallel
+   * and return the last locator that locate element.
+   *
+   * @public
+   */
+  static last<Locators extends readonly unknown[] | []>(
```
OrKoN, line 134: could you elaborate more about the testing scenario you have?  Can you use the following instead?

```
await Promise.all([
  page.locator('#1').wait(),
  page.locator('#2').wait()
]);
await page.locator('#3').click();
```

It is not as parallelized but it does not have the following problem:

```

const result = Locator.last([
    page.locator('#test1'),
    page.locator('#test2'),
    page.locator('#test3'),
  ])
    .setTimeout(5000)
    .click();
```

The last locator never intends to click on the first two located elements but the preconditions from the click action would also apply when locating those elements which does not make sense since they are never clicked.

minggeorgelei: Hi @OrKoN ,
The testing scenario is like you suggested. The third locater depends on the first two locators. The third locater can do some operations like click only after the appearances of the first two elements.
The code is like this:
await Locator.last([locator1, locator2]).wait();
await locator3.click();
I think it is more convenient to provide such a method and it may be faster since it is paralleled.

OrKoN: so `wait` and `click` add different preconditions, so if you use Locator.last with `click` you would get different results compared to using it with `wait`. I am concerned that it might be unexpected for the users, who would expect the wait conditions to apply to the first locator, and the click conditions apply only to the last locator.

OrKoN: actually, I might be misremembering the locator implementation. @Lightning00Blade would you mind giving a second opinion?

test/src/locator.spec.ts
```diff
@@ -581,6 +581,38 @@ describe('Locator', function () {
     });
   });
 
+  describe('Locator.last', () => {
+    it('should time out when any one of the locaters do not match', async () => {
```
OrKoN: could you please add some tests for the following scenarios?

1) Locator.last waits for multiple elements that a clickable and it succeeds clicking only on the last one
2) Locator.last waits for multiple elements but only the last one is clickable and it succeeds clicking on the last one

minggeorgelei: OK, I will add some tests as you suggested.

test/src/locator.spec.ts
```diff
@@ -611,6 +611,76 @@ describe('Locator', function () {
...
+      expect(count).toBe(0);
+      expect(innerText).toBe('test2');
+    });
+
+    it('click the last one successfully with only the last one is clikable', async () => {
+      const {page} = await getTestState();
+
+      await page.setViewport({width: 500, height: 500});
+      await page.setContent(`
+        <button id="test1">test1</button>
```
OrKoN: button is always clickable, let's make test1 to be a div with width=0 and height=0 set via the style attribute?

minggeorgelei: Ok I will change it to div

minggeorgelei: Tests have been modified. I still see the test failed. Can you help to check the failure? In my machine, the added tests can pass successfully.

OrKoN: I am not able to reproduce on a Mac locally either

OrKoN: seems to be Ubuntu-only

packages/puppeteer-core/src/api/locators/locators.ts
```diff
@@ -113,6 +114,18 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
     return RaceLocator.create(locators);
   }
 
+  /**
+   * Trying to locate all the elements for locator array in parallel
+   * and return the last locator that locate element.
```
OrKoN:
```suggestion
   * Waits for multiple locators to locate elements in DOM and 
   * calls the action on the last locator waiting for action-specific preconditions.
   * If any of the elements identified by locators is not found, the last locator times out.
```

packages/puppeteer-core/src/api/locators/locators.ts
```diff
@@ -113,6 +114,18 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
...
+   *
```
OrKoN, line 131: could you please add an inline `@example` on how to use the locator to this comment?

test/src/locator.spec.ts
```diff
@@ -581,6 +581,109 @@ describe('Locator', function () {
     });
   });
 
+  describe('Locator.last', () => {
+    it('should time out when any one of the locaters do not match', async () => {
```
OrKoN:
```suggestion
    it('should time out when any one of the locators do not match', async () => {
```

test/src/locator.spec.ts
```diff
@@ -581,6 +581,109 @@ describe('Locator', function () {
...
+        clock.restore();
+      }
+    });
+
+    it('click the last one successfully with multiple clickable elements', async () => {
+      const {page} = await getTestState();
+
+      await page.setViewport({width: 500, height: 500});
+      await page.setContent(`
+        <button id="test1">test1</button>
```
OrKoN: let's minimize the number of elements in test by using inline click handlers:

```
<button id="test1" onclick="this.innerText = 'clicked1'">test1</button>
```

test/src/locator.spec.ts
```diff
@@ -581,6 +581,109 @@ describe('Locator', function () {
...
+        <div id="test2"></div>
+        <div id="test4"></div>
+         <script>
+          const button1 = document.getElementById("test1");
+          button1.onclick = () => {
+            const div2 = document.getElementById("test4");
+            div2.innerText = "test4";
+          };
+          setTimeout(() => {
```
OrKoN: instead of having setTimeout in the page context let's insert another element from the test:

``` 
await page.setContent(`<button id="test1" onclick="this.innerText = 'clicked1'">test1</button>`);

const clickDone = Locator.last([
        page.locator('#test1'),
        page.locator('#test2'),
      ]).click();

await new Promise(resolve => setTimeout(resolve, 50));

await page.evaluate(() => {
  document.body.innerHTML += `<button id="test2" onclick="this.innerText = 'clicked2'">test2</button>`'
});

await clickDone;

expect(await page.$eval('test1', (el) => el.innerText)).toBe('test1');
expect(await page.$eval('test12, (el) => el.innerText)).toBe('clicked');
```

test/src/locator.spec.ts
minggeorgelei: Ok I will modify it

minggeorgelei:
Hi @OrKoN , changes have been made according to the review

OrKoN:
I will investigate the failing test today

minggeorgelei:
> I will investigate the failing test today

OK

test/src/locator.spec.ts
```diff
@@ -581,6 +581,105 @@ describe('Locator', function () {
...
+      const {page} = await getTestState();
+
+      await page.setViewport({width: 500, height: 500});
+      await page.setContent(
+        `<button id="test1" onclick="this.innerText = 'clicked1'">test1</button>`
+      );
+
+      const clickDone = Locator.last([
+        page.locator('#test1'),
+        page.locator('#test2'),
```
OrKoN, line 625: so the error in Firefox happens because the 'test1' is actually detected before test2. Is it intended that the API clicks on the last one found and not the last one specified locator? what is the expectation for the following test?

```

    it('click the last one successfully with multiple clickable elements', async () => {

      await page.setContent(
        `<button id="test2" onclick="this.innerText = 'clicked2'">test2</button>`
      );

      const clickDone = Locator.last([
        page.locator('#test1'),
        page.locator('#test2'),
      ]).click();

      await new Promise(resolve => {
        setTimeout(resolve, 500);
      });

      await page.evaluate(() => {
        document.body.innerHTML += `<button id="test1" onclick="this.innerText = 'clicked1'">test1</button>`;
      });

      await clickDone;

      expect(
        await page.$eval('#test1', el => {
          return (el as HTMLElement).innerText;
        })
      ).toBe('test1');
      expect(
        await page.$eval('#test2', el => {
          return (el as HTMLElement).innerText;
        })
      ).toBe('clicked2');
    });
```

packages/puppeteer-core/src/api/locators/locators.ts
```diff
@@ -113,6 +114,29 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
...
 
+  /**
+   * Waits for multiple locators to locate elements in DOM and
+   * calls the action on the last locator waiting for action-specific preconditions.
+   * If any of the elements identified by locators is not found, the last locator
+   * times out.
+   * @example
+   *
+   * ```ts
+   * await Locator.last([
```
OrKoN, line 125: is there an example where one would want to do an action other than wait using Locator.last? it seems like the Last locator is only useful with waiting?

minggeorgelei: In some scenarios, users may be sure that some certain elements may appear later than other elements and also the last element depends on the earlier ones to do the operation. I think wait is more often used. Do I also need to add an operation example?

OrKoN: I think if you need to wait multiple locators for majority of the cases:

`await Promise.all([page.locator('#test1'), page.locator('#test2')].map(l => l.wait()))` is more easy to use and more clear what it does. Is there a good scenario when one wants to click or do another action on the last element that appears in DOM?

test/src/locator.spec.ts
minggeorgelei: The expectation is like the test showed. We found the last located element and click it. I think the case can pass in all the machines except firefox ubuntu.

minggeorgelei: The expectation is like the test showed. We found the last located element and click it. I think the case can pass in all the machines except firefox ubuntu.

OrKoN: The test does not pass though, it clicks on the test1 although the last locator is for test2

packages/puppeteer-core/src/api/locators/locators.ts
minggeorgelei: I cannot think of a strong testing scenario to support that one wants to click on the last element. Still wait is the more often case for it.

minggeorgelei:
The testing scenario for doing the operation for the last element is not very often. The wait scenario can be replaced by using await Promise.all([page.locator('#test1'), page.locator('#test2')].map(l => l.wait())).

Issue #12700, opened 2024-07-03
[Bug]: Buffer is not defined while using request.respond in chrome extension(using pptr-core)


elineeen:
### Minimal, reproducible example

```TypeScript
cmdPage.on('request', interceptedRequest => {
		if (interceptedRequest.url().indexOf('main.js') > 0) {
			interceptedRequest.respond({
				body: mainReplacerFile,
			})
		} else
			interceptedRequest.continue()
	})
```

### Error string

Uncaught (in promise) ReferenceError: Buffer is not defined

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Expectation

Works fine with req.conitnue api but stucked on respond lol. Hope for supporting

### Reality

![image](https://github.com/puppeteer/puppeteer/assets/61259088/c769e6c5-21c3-48a0-9af3-2739f5fc856d)

### Puppeteer version

22.12.1

### Node version

18.17.0

### Package manager

yarn

### Package manager version

3.6.1

### Operating system

Windows

OrKoN:
@Lightning00Blade do you know if we can rewrite without using the Buffer API?

Lightning00Blade:
We should be able to move to `Uint8Array` as it will not be a breaking change `interface Buffer extends Uint8Array` 
And `Uint8Array` is supported in the browser.

OrKoN:
As a workaround, it should be possible to include https://github.com/feross/buffer into your build.

PR #12702, opened 2024-07-03
fix: don't rely on Buffer to be present

Lightning00Blade:
https://github.com/puppeteer/puppeteer/issues/12700

It seems to be not straight forward to use only `Uint8Array` when loading a file and providing it's buffer,
so we still fallback to `Buffer.toString('base64')`, when we encounter non-string.

### DIFF
--- a/packages/puppeteer-core/src/api/HTTPRequest.ts
+++ b/packages/puppeteer-core/src/api/HTTPRequest.ts
@@ -559,8 +559,14 @@ export abstract class HTTPRequest {
     let contentLength: number;
     let base64: string;
     if (isString(body)) {
-      contentLength = new TextEncoder().encode(body).byteLength;
-      base64 = btoa(body);
+      const encodedBody = new TextEncoder().encode(body);
+      contentLength = encodedBody.byteLength;
+
+      const bytes = [];
+      for (const byte of encodedBody) {
+        bytes.push(String.fromCharCode(byte));
+      }
+      base64 = btoa(bytes.join(''));
     } else {
       contentLength = Buffer.byteLength(body);
       base64 = body.toString('base64');
--- a/packages/puppeteer-core/src/api/HTTPRequest.ts
+++ b/packages/puppeteer-core/src/api/HTTPRequest.ts
@@ -6,7 +6,7 @@
 import type {Protocol} from 'devtools-protocol';
 
 import type {ProtocolError} from '../common/Errors.js';
-import {debugError} from '../common/util.js';
+import {debugError, isString} from '../common/util.js';
 import {assert} from '../util/assert.js';
 
 import type {CDPSession} from './CDPSession.js';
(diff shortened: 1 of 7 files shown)


packages/puppeteer-core/src/cdp/HTTPRequest.ts
```diff
@@ -218,9 +218,9 @@ export class CdpHTTPRequest extends HTTPRequest {
     if (response.contentType) {
       responseHeaders['content-type'] = response.contentType;
     }
-    if (responseBody && !('content-length' in responseHeaders)) {
+    if (responseBodyBase64 && !('content-length' in responseHeaders)) {
       responseHeaders['content-length'] = String(
-        Buffer.byteLength(responseBody)
+        new TextEncoder().encode(responseBodyBase64).length
```
OrKoN:
```suggestion
        responseBodyBase64.length
```

OrKoN: actually, the length should be of the original body, not of the base64 ecnodes string.

packages/puppeteer-core/src/bidi/HTTPRequest.ts
```diff
@@ -230,12 +231,16 @@ export class BidiHTTPRequest extends HTTPRequest {
...
-    const responseBody: string | undefined =
-      response.body && response.body instanceof Uint8Array
-        ? response.body.toString('base64')
-        : response.body
-          ? btoa(response.body)
-          : undefined;
+
+    let responseContentLength: number | undefined;
+    let responseBodyBase64: string | undefined;
+    if (response.body && isString(response.body)) {
```
OrKoN: nit: let's move the if isString else isBuffer to if (response.body) so that we do not need to check the body twice (and imo it is more readable).

packages/puppeteer-core/src/bidi/HTTPRequest.ts
```diff
@@ -230,12 +231,16 @@ export class BidiHTTPRequest extends HTTPRequest {
...
+      responseContentLength = new TextEncoder().encode(response.body).length;
```
OrKoN: could you add a comment on why the encode needed?

packages/puppeteer-core/src/cdp/HTTPRequest.ts
```diff
@@ -198,10 +196,15 @@ export class CdpHTTPRequest extends HTTPRequest {
   async _respond(response: Partial<ResponseForRequest>): Promise<void> {
     this.interception.handled = true;
 
-    const responseBody: Buffer | null =
-      response.body && isString(response.body)
-        ? Buffer.from(response.body)
-        : (response.body as Buffer) || null;
+    let responseContentLength: number | undefined;
```
OrKoN: can we refactor to re-use this logic?

test/src/requestinterception.spec.ts
```diff
@@ -956,6 +956,36 @@ describe('request interception', function () {
...
     });
+
+    it('should report correct content-length header with string', async () => {
+      const {page, server} = await getTestState();
+
+      await page.setRequestInterception(true);
+      page.on('request', request => {
+        void request.respond({
+          status: 200,
+          body: 'Correct length?',
```
OrKoN: let's add a test with multi-byte characters

packages/puppeteer-core/src/api/HTTPRequest.ts
```diff
@@ -548,6 +548,29 @@ export abstract class HTTPRequest {
...
+
+  /**
+   * @internal
+   */
+  static getResponse(body: string | Uint8Array): {
+    contentLength: number;
+    base64: string;
+  } {
+    // Needed to get the correct byteLength
+    const byteBody: Uint8Array = isString(body)
```
OrKoN, line 560: Let's add a unit test comparing the results of this helper with Buffer implementation: let's try empty string, ' ', multi-byte character, latin characters. It looks equivalent but just to be sure.

OrKoN (approved):

PR #12704, opened 2024-07-04
test: mark tests for history API usage for Firefox and WebDriver BiDi as fail/skip

whimboo:
Based on the changes of https://phabricator.services.mozilla.com/D215609 we have to mark those tests upstream as well. This PR can be merged when https://bugzilla.mozilla.org/show_bug.cgi?id=1879163 made it to a Nightly build which as earliest will be this evening.


whimboo:
The changes are now in Firefox Nightly. So it should all work now.

OrKoN (approved):
Thanks!

PR #12708, opened 2024-07-05
test: update test expectations

OrKoN:


OrKoN:
@whimboo @juliandescottes I have removed the expectations blocked on the cache behavior but they do not pass yet. Is there anything that has not landed still or is there another issue?

OrKoN:
It looks like it is the CSS caching issue as well?

juliandescottes:
> It looks like it is the CSS caching issue as well?

Yes exactly I need to rebase my PR to update the tests, same issue with the stylesheet cache

OrKoN:
I am not 100% it is about CSS so I marked them as FAIL with a note to investigate

juliandescottes:
The coop font one fails because of a race condition, but I think we should fix it in the test? The others are stylesheet issues, I prepared a PR for this at #12709 but it will probably clash with yours

Issue #12703, opened 2024-07-04
[Bug]: build problems using puppert-core in extensions


elineeen:
### Minimal, reproducible example

```TypeScript
vite build
```

### Error string

PassThrough" is not exported by "__vite-browser-external”

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

Currently i'm building the extension app with vite using pptr-core, and stucked in some modules whether using the example rollup config or not(since node-resolve is already a default vite built-in ). Though i can add externals to rollupOptions to skip these problems in building phase ,i'm not sure these api will works fine without some sort of node polyfills

### Expectation

maybe a better bundle example or some compatible browser api replacement?

### Reality

![image](https://github.com/puppeteer/puppeteer/assets/61259088/7a4d6766-a3a1-43d1-afc7-a83913fc17a9)
![image](https://github.com/puppeteer/puppeteer/assets/61259088/32419e58-6d14-417e-97ac-367f29877737)
currently the build process works fine after adding these externals 
![image](https://github.com/puppeteer/puppeteer/assets/61259088/6fb9255b-cc05-406b-96e2-99bde502192a)

### Puppeteer version

22.12.1

### Node version

18.17.0

### Package manager

yarn

### Package manager version

3.6.1

### Operating system

Windows

OrKoN:
Have you tried this example https://github.com/puppeteer/puppeteer/tree/main/examples/puppeteer-in-browser? It’s also important to import the browser-specific entrypoint like in the example and not the main puppeteer-core entrypoint.

elineeen:
I get your point , but in my case i use connect api in my chrome extension with transport options just like the extension demo instead of the examples which using endpoint. Does this affect build phase?
![image](https://github.com/puppeteer/puppeteer/assets/61259088/31d1b82b-f49f-4bc6-aed8-2c6f90a76fae)

OrKoN:
There is an example specifically for extension: https://github.com/puppeteer/puppeteer/tree/main/examples/puppeteer-in-extension. The rollup config https://github.com/puppeteer/puppeteer/blob/main/examples/puppeteer-in-extension/rollup.config.mjs

It builds for me with the following log:

```
❯ npm run build

> puppeteer-in-extension@1.0.0 build
> rollup -c && cp manifest.json out/


background.js → out...
(!) "this" has been rewritten to "undefined"
https://rollupjs.org/troubleshooting/#error-this-is-undefined
[...]

PR #12710, opened 2024-07-05
refactor!: replace dynamic imports with static dependency injection

OrKoN:
Dynamic imports get included by bundlers and extra configuration is needed to disable or work around them. The best solution is that the browser specific entrypoint does not include any Node imports (either dynamic or static).

This PR implements a container to resolve environment-specific dependencies that does not require dynamic imports. As the first step, importFSPromises is migrated.

Issue https://github.com/puppeteer/puppeteer/issues/12703

### DIFF
--- a/packages/puppeteer-core/src/puppeteer-core-browser.ts
+++ b/packages/puppeteer-core/src/puppeteer-core-browser.ts
@@ -8,6 +8,8 @@ export * from './index-browser.js';
 
 import {Puppeteer} from './common/Puppeteer.js';
 
+// Set up browser-specific environment dependencies here.
+
 /**
  * @public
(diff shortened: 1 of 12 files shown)


Lightning00Blade (approved):

packages/puppeteer-core/src/puppeteer-core-browser.ts
```diff
@@ -8,6 +8,8 @@ export * from './index-browser.js';
 
 import {Puppeteer} from './common/Puppeteer.js';
 
+// Set up browser-specific environment dependencies here.
```
OrKoN:
```suggestion
```

PR #12724, opened 2024-07-08
fix: do not allow switching tabs while the screenshot operation is in progress

### LINKED ISSUE #12712: [Bug]: Page.screenshot() can time out with `headless: true` and `protocol: 'cdp'`

```TypeScript

(async () => {
  let browser = await puppeteer.launch({
    protocolTimeout: 180_000,
    headless: true,
    protocol: "cdp",
  });

  console.time("screenshot1");
  const screen1 = page1.screenshot();
  const screen2 = page2.screenshot();

  const close1 = screen1.then(() => page1.close());
  const close2 = screen2.then(() => page2.close());
  await screen1;
  const page3 = await browser.newPage();
  await page3.screenshot(); // Times out
  const close3 = page3.close();

  Promise.all([close1, close2, close3]).then(() => browser.close());
})();
```

### Error string

ProtocolError: Page.captureScreenshot timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

I've been trying to generate screenshots of pages using puppeteer. We're getting protocol timeouts for page.screenshot for a small number of requests.

While investigating, I found a minimal script that reproduces the issue almost all of the time.

Any  of the following changes to the code cause the problem to disappear:
- `protocol: 'webDriverBiDi'`
- `headless: 'shell'`
- Removing page2, screen2 and close2

Setting `pipe: true` does not resolve the issue.

I've been able to reproduce this issue for many large and small values of protocolTimeout, setting it to 3_000 means the script fails faster. (screenshots usually take ~200ms locally).

### Expectation

I expected page3.screenshot() to resolve in a similar duration to most other calls (<200ms).

### Reality

In reality, it rejects with `ProtocolError: Page.captureScreenshot timed out` even for very large values of `protocolTimeout`(3 minutes). 

I've also been able to reproduce this issue in puppeteer 19.11.1, 20.9.0 and 21.11.0 by making the following changes to the script (to fix api differences in versions before 22):
- removing protocol (as this is 'cdp' by default)
- setting headless: 'new' (equivalent to headless: true in puppeteer 22).

_No response_

### Puppeteer version

22.12.1

### Node version

22.4.0

### Package manager

npm

### Package manager version

10.8.1

### Operating system

macOS

OrKoN:
This PR adds synchronization between `browser.newPage()`, `page.close()` and `page.screenshot()`. Previously the  `page.screenshot()` calls were synchronized so that one call is possible at a time in a browser instance. While that makes concurrent  `page.screenshot()` to be sequential and allows them not to interfere with each other, it was still possible to interfere with the running screenshot operations by opening a new page or closing an existing page, which might cause the tab where the screenshot operation is running to become backgrounded, which lead to time outs when taking screenshots.

The synchronization happens per browser context so it is possible to open and close pages from a different browser context while taking a screenshot. The change applies also to headless-shell  and Firefox where the behavior of the background tabs might be different.

Closes #12712

### DIFF
--- a/packages/puppeteer-core/src/api/BrowserContext.ts
+++ b/packages/puppeteer-core/src/api/BrowserContext.ts
@@ -18,6 +18,7 @@ import {
   timeout,
 } from '../common/util.js';
 import {asyncDisposeSymbol, disposeSymbol} from '../util/disposable.js';
+import {Mutex} from '../util/Mutex.js';
 
 import type {Browser, Permission, WaitForTargetOptions} from './Browser.js';
 import type {Page} from './Page.js';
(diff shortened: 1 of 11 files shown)


packages/puppeteer-core/src/api/BrowserContext.ts
```diff
@@ -104,6 +105,39 @@ export abstract class BrowserContext extends EventEmitter<BrowserContextEvents>
    */
   abstract targets(): Target[];
 
+  /**
+   * If defined, indicates an ongoing screenshot opereation.
+   *
+   * @internal
```
Lightning00Blade:
```suggestion
```

Lightning00Blade (approved):

PR #12732, opened 2024-07-11
feat!: default to WebDriver BiDi for Firefox

OrKoN:
This PR changes the default protocol used when launching Firefox to WebDriver BiDi.

### DIFF
--- a/packages/puppeteer-core/src/common/ConnectOptions.ts
+++ b/packages/puppeteer-core/src/common/ConnectOptions.ts
@@ -49,9 +49,14 @@ export interface BrowserConnectOptions {
   _isPageTarget?: IsPageTargetCallback;
 
   /**
-   * @defaultValue 'cdp' when launching Chrome and 'webDriverBiDi' when
-   * launching Firefox. When connecting to a running browser, defaults
-   * to 'cdp'.
+   * @defaultValue Determined at run time:
+   *
+   * - Launching Chrome - 'cdp'.
+   *
+   * - Launching Firefox - 'webDriverBiDi'.
+   *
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/common/ConnectOptions.ts
```diff
@@ -49,7 +49,9 @@ export interface BrowserConnectOptions {
   _isPageTarget?: IsPageTargetCallback;
 
   /**
-   * @defaultValue 'cdp'
+   * @defaultValue 'cdp' when launching Chrome and 'webDriverBiDi' when
+   * launching Firefox. When connecting to a running browser, defaults
+   * to 'cdp'.
```
Lightning00Blade: Let's simplify it something like:
```suggestion
   * @defaultValue Determined at run time:
   * Launching Chrome - 'cdp'.
   * Launching Firefox - 'webDriverBiDi'.
   * Connecting to a browser - 'cdp'.
```

OrKoN: I do not think we would render the list correctly?

Lightning00Blade:
```suggestion
   * @defaultValue Determined at run time:
   *
   * - Launching Chrome - 'cdp'.
   *
   * - Launching Firefox - 'webDriverBiDi'.
   *
   * - Connecting to a browser - 'cdp'.
   *
```

Lightning00Blade: image

Lightning00Blade (approved):

PR #12736, opened 2024-07-11
fix(webdriver): support securityDetails with cdp-over-bidi

OrKoN:


Lightning00Blade (approved):

Lightning00Blade:
Not sure if we should mark as `fix` or not.

PR #12739, opened 2024-07-12
fix(webdriver): add postData,hasPostData,resourceType from cdp-over-bidi

OrKoN:


packages/puppeteer-core/src/bidi/core/Request.ts
```diff
@@ -164,6 +164,21 @@ export class Request extends EventEmitter<{
...
   }
 
+  get resourceType(): string | undefined {
+    // @ts-expect-error non-standard attribute.
+    return this.#event['goog:resourceType'] ?? undefined;
+  }
+
+  get postData(): string {
+    // @ts-expect-error non-standard attribute.
+    return this.#event['goog:postData'] ?? '';
```
Lightning00Blade, line 174: I think the Puppeteer API expects undefined if not present. 
```suggestion
    return this.#event['goog:postData'] ?? undefined;
```

Lightning00Blade (approved):

PR #12752, opened 2024-07-15
ci(webdriver): test WebDriverBiDi on Chrome stable

Lightning00Blade:


OrKoN (approved):

OrKoN:
@Lightning00Blade we should do the same change in chromium-bidi and perhaps do it before we switch over here

Lightning00Blade:
CC @sadym-chromium

OrKoN:
@Lightning00Blade let's land this

PR #12767, opened 2024-07-17
test: add set cookie with all values test

kblok:
No code changes. I'm just adding a test to validate that we can send a valid cookie with all the available properties to the browser.

### DIFF
--- a/test/src/cookies.spec.ts
+++ b/test/src/cookies.spec.ts
@@ -311,6 +311,43 @@ describe('Cookie specs', () => {
         ]
       );
     });
+    it('should set cookie with all available properties', async () => {
+      const {page, server} = await getTestState();
+
+      await page.goto(server.EMPTY_PAGE);
+      await page.setCookie({
+        name: 'password',
+        value: '123456',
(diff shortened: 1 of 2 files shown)


OrKoN (commented):
Thanks for the test. I think not all cookie attributes work cross-browser. Perhaps we need to adjust the expectations base on the browser?

kblok:
It looks stuck right @OrKoN? or Am I missing something to trigger the builds?

test/src/cookies.spec.ts
```diff
@@ -311,6 +311,46 @@ describe('Cookie specs', () => {
         ]
       );
     });
+    it('should set cookie with all available properties', async () => {
+      const {page, server} = await getTestState();
```
kblok, line 315:
```suggestion
      const {page, server} = await getTestState();

```

kblok:
@OrKoN it seems the build requires your approval to run?

kblok:
> @kblok could you please run `npm run format:expectations` and commit changes?

done!

OrKoN (approved):

Issue #12768, opened 2024-07-17
In Chrome over WebDriver BiDi CDP sessions receive events that are not related to the session


OrKoN:
```
  it.only('should send events', async () => {
    const {page, server} = await getTestState();

    const client = await page.createCDPSession();
    await client.send('Network.enable');
    const events: unknown[] = [];
    client.on('*', function () {
      console.log(arguments)
    })
    client.on('Network.requestWillBeSent', event => {
      events.push(event);
    });
    await Promise.all([
      waitEvent(client, 'Network.requestWillBeSent'),
      page.goto(server.EMPTY_PAGE),
    ]);
    expect(events).toHaveLength(1);
  });
  ```
  
  Expected: only network events are received
  Actual: unrelated events are received

OrKoN:
@Lightning00Blade I will take a look into this

PR #12771, opened 2024-07-17
fix(webdriver): allow accessing raw CDP connection when using WebDriver BiDi

OrKoN:
Changes createCDPSession to use CDP connection directly instead of bidi+.

Refs https://github.com/puppeteer/puppeteer/issues/12768

### DIFF
--- a/packages/puppeteer-core/src/common/BrowserConnector.ts
+++ b/packages/puppeteer-core/src/common/BrowserConnector.ts
@@ -55,7 +55,7 @@ export async function _connectToBrowser(
  * Establishes a websocket connection by given options and returns both transport and
  * endpoint url the transport is connected to.
  */
-async function getConnectionTransport(
+export async function getConnectionTransport(
   options: BrowserConnectOptions & ConnectOptions
 ): Promise<{connectionTransport: ConnectionTransport; endpointUrl: string}> {
   const {browserWSEndpoint, browserURL, transport, headers = {}} = options;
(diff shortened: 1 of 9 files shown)


test/TestExpectations.json
```diff
@@ -940,11 +940,18 @@
...
+    "expectations": ["FAIL"],
+    "comment": "Fails with Firefox CDP"
+  },
+  {
+    "testIdPattern": "[CDPSession.spec] Target.createCDPSession should not send extra events",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["chrome", "webDriverBiDi"],
     "expectations": ["FAIL"],
-    "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
+    "comment": "https://github.com/GoogleChromeLabs/chromium-bidi/issues/2420"
```
OrKoN, line 954: Depends on chromium-bidi, but I cannot replicate it in the e2e setup so we can land the test to check with puppeteer.

packages/puppeteer-core/src/common/BrowserConnector.ts
```diff
@@ -55,7 +55,7 @@ export async function _connectToBrowser(
  * Establishes a websocket connection by given options and returns both transport and
  * endpoint url the transport is connected to.
  */
-async function getConnectionTransport(
+export async function getConnectionTransport(
```
Lightning00Blade:
```suggestion
async function getConnectionTransport(
```

Lightning00Blade (approved):

PR #12777, opened 2024-07-18
fix(webdriver): in page.goto consider only the first emitted navigation event

### LINKED ISSUE #12773: [Bug]: Trying to access response.status() right after the navigation can cause a TypeError "Cannot read properties of null (reading 'status')"

```TypeScript
'use strict';

(async () => {
    protocol: 'webDriverBiDi',
    headless: false,
  });

  page.on('request', async request => {
    request.continue();
  });

  const response = await page.goto(`https://developer.mozilla.org/en-US/`);
  console.log(`status=${response.status()}`);

  await browser.close();
})();
```

### Background

_No response_

### Expectation

That the returned `response` object is valid and properties can be accessed.

### Reality

The `response` is undefined unless `await` is used in the interception handler for `request.continue()`. Note that all [Puppeteer docs](https://pptr.dev/api/puppeteer.httprequest.continue#example) do not list `await` and when using Firefox it also works with out `await`.

_No response_

### Puppeteer version

22.13.1

### Node version

22.2.0

### Package manager

npm

### Package manager version

10.7.0

### Operating system

macOS

OrKoN:
Fix: https://github.com/puppeteer/puppeteer/issues/12773

### DIFF
--- a/packages/puppeteer-core/src/bidi/Frame.ts
+++ b/packages/puppeteer-core/src/bidi/Frame.ts
@@ -350,56 +350,58 @@ export class BidiFrame extends Frame {
     });
     return await firstValueFrom(
       combineLatest([
-        fromEmitterEvent(this.browsingContext, 'navigation').pipe(
-          switchMap(({navigation}) => {
-            return this.#waitForLoad$(options).pipe(
-              delayWhen(() => {
-                if (frames.length === 0) {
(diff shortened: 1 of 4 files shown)


Lightning00Blade (approved):

whimboo:
Thanks a lot Alex!

PR #12782, opened 2024-07-18
chore(build): use a object to represent supported versions

Lightning00Blade:
This matches the object we provide for each version, and makes it easier to work with.


OrKoN (changes requested):
I think we do not need it for now

PR #12790, opened 2024-07-19
test: update late responseCompleted tests to pass/fail for Firefox

juliandescottes:
Updates 6 expectations to PASS/FAIL for firefox / bidi due to the responseCompleted event arriving too late. The investigation bug on mozilla's side is at https://bugzilla.mozilla.org/show_bug.cgi?id=1907589, though we hope that fixing https://bugzilla.mozilla.org/show_bug.cgi?id=1882803 will also address this.

Alternatively, we could update the tests in order to wait for the response to be populated, which would avoid marking them as pass / fail. But in general per spec the tests are correct and responseCompleted should have been received for those requests by the time load is emitted.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -2324,6 +2324,13 @@
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
   {
+    "testIdPattern": "[network.spec] network Network Events Page.Events.RequestServedFromCache for script",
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

OrKoN (commented):
could you please `npm run format:expectations`?

juliandescottes:
> could you please `npm run format:expectations`?

Yep sorry, done

PR #12791, opened 2024-07-19
chore: rename API as well

Lightning00Blade:


OrKoN (approved):
let's land it into the target branch

PR #12795, opened 2024-07-19
feat!: support multiple browser downloads for Puppeteer

Lightning00Blade:
With this change you can configure the download behavior for both Chrome and Firefox browsers via puppeteer configuration. See https://pptr.dev/guides/configuration for details on the new config format. We recommend switching to configuration file instead of environment variables.

The following environment variables changed to support multiple browsers:

- `PUPPETEER_BROWSER_REVISION`  to `PUPPETEER_<browser-name>_VERSION`
- `PUPPETEER_DOWNLOAD_BASE_URL`  to `PUPPETEER_<browser-name>_BASE_URL`
- `PUPPETEER_SKIP_CHROME_DOWNLOAD`  to `PUPPETEER_CHROME_SKIP_DOWNLOAD`
- `PUPPETEER_SKIP_CHROME_HEADLESS_SHELL_DOWNLOAD`  to `PUPPETEER_CHROME_SKIP_HEADLESS_SHELL_DOWNLOAD`

The following env variable was added:

- `PUPPETEER_FIREFOX_SKIP_DOWNLOAD` to control Firefox download behavior.

Closes #6225

PR #12806, opened 2024-07-22
feat: allow downloading of all pinned browsers

Lightning00Blade:
One thing that is under works is how to handle the configuration? Should we accept another parameter? 

Note: this does not contain breaking changes.

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -27,11 +27,12 @@ import {
   launch,
 } from './launch.js';
 
+interface InstallBrowser {
+  name: Browser;
+  buildId: string;
+}
 interface InstallArgs {
-  browser: {
-    name: Browser;
-    buildId: string;
-  };
+  browser?: InstallBrowser;
   path?: string;
   platform?: BrowserPlatform;
   baseUrl?: string;
@@ -152,9 +153,11 @@ export class CLI {
 
   #build(yargs: Yargs.Argv<unknown>): Yargs.Argv<unknown> {
     const latestOrPinned = this.#pinnedBrowsers ? 'pinned' : 'latest';
+    // If there are pinned browser allow the positional arg to be optional
+    const browserArgType = this.#pinnedBrowsers ? '[browser]' : '<browser>';
     return yargs
       .command(
-        'install <browser>',
+        `install ${browserArgType}`,
         'Download and install the specified browser. If successful, the command outputs the actual browser buildId that was installed and the absolute path to the browser executable (format: <browser>@<buildID> <path>).',
         yargs => {
           this.#defineBrowserParameter(yargs);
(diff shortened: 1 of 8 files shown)


OrKoN:
> One thing that is under works is how to handle the configuration? Should we accept another parameter?

which configuration?

> Note: this does not contain braking changes.

s/braking/breaking

Lightning00Blade:
So currently it download every browser that is pinned, but from Puppeteer CLI, we would want the configuration file and default to manage which browser get  downloaded.

OrKoN:
@Lightning00Blade, yes, so pinned browsers are only set for the Puppeteer CLI, so we would need to update the list of pinned browsers in packages/puppeteer/src/node/cli.ts based on the current configuration (probably, after https://github.com/puppeteer/puppeteer/pull/12795 lands).

packages/browsers/src/CLI.ts
```diff
@@ -152,9 +153,11 @@ export class CLI {
...
     const latestOrPinned = this.#pinnedBrowsers ? 'pinned' : 'latest';
+    // If there are pinned browser allow the positional arg to be optional
+    const browserArgType = this.#pinnedBrowsers ? '[browser]' : '<browser>';
     return yargs
       .command(
-        'install <browser>',
+        `install ${browserArgType}`,
         'Download and install the specified browser. If successful, the command outputs the actual browser buildId that was installed and the absolute path to the browser executable (format: <browser>@<buildID> <path>).',
         yargs => {
           this.#defineBrowserParameter(yargs);
```
OrKoN, line 173: let's add an example for installing all pinned browsers (if pinned browsers are known)

OrKoN: By example, I mean using `yargs.example()` that happens below these lines

packages/browsers/src/CLI.ts
```diff
@@ -261,50 +264,21 @@ export class CLI {
...
-          if (args.browser.buildId === 'pinned') {
-            const pinnedVersion = this.#pinnedBrowsers?.[args.browser.name];
-            if (!pinnedVersion) {
-              throw new Error(
-                `No pinned version found for ${args.browser.name}`
-              );
-            }
-            args.browser.buildId = pinnedVersion;
+          if (this.#pinnedBrowsers && !args.browser) {
+            await Promise.all(
```
OrKoN: should we use Promise.allSettled here and throw if there is any error?

Lightning00Blade: I think it make sense, one problem with the updated implementation is surfacing the error messages from the install function.

OrKoN: I think re-throwing the first one would result in the same behavior w.r.t to errors

OrKoN (approved):

Issue #12787, opened 2024-07-18
"page.setCookie" fails when using "partitionKey" field in cookie: "Invalid parameters Failed to deserialize params.cookies.partitionKey"


whimboo:
The following script always fails with Chrome and not Firefox when using the `partitionKey` field for a cookie that gets set via `page.setCookie`:

```
    'use strict';
    ​
    const puppeteer = require('puppeteer');
    ​
    (async () => {
      const browser = await puppeteer.launch({
        protocol: 'webDriverBiDi',
      });
    ​
      const page = await browser.newPage();
    ​
      const domain = "addons.mozilla.org";
      const cookie = { name: "John", value: "42", domain, partitionKey: `https://${domain}` };
      await page.setCookie(cookie);
    ​
      await browser.close();
    })();
```

The failure is:

```
ProtocolError: Protocol error (storage.setCookie): unable to set cookie ProtocolError: Protocol error (Storage.setCookies): Invalid parameters Failed to deserialize params.cookies.partitionKey - BINDINGS: string value expected at position 115
```

Here an excerpt from the debug log of Puppeteer:

```
  puppeteer:webDriverBiDi:SEND ►   '{"id":10,"method":"storage.setCookie","params":{"cookie":{"domain":"addons.mozilla.org","name":"John","value":{"type":"string","value":"42"}},"partition":{"type":"storageKey","sourceOrigin":"https://addons.mozilla.org","userContext":"default"}}}'
  puppeteer:webDriverBiDi:SEND ► ] +0ms
  puppeteer:protocol:SEND ► [
  puppeteer:protocol:SEND ►   '{"method":"Storage.setCookies","params":{"cookies":[{"name":"John","value":"42","domain":"addons.mozilla.org","path":"/","secure":false,"httpOnly":false,"partitionKey":{"hasCrossSiteAncestor":false,"topLevelSite":"https://addons.mozilla.org"}}]},"id":9}'
  puppeteer:protocol:SEND ► ] +0ms
  puppeteer:protocol:RECV ◀ [
  puppeteer:protocol:RECV ◀   '{"id":9,"error":{"code":-32602,"message":"Invalid parameters","data":"Failed to deserialize params.cookies.partitionKey - BINDINGS: string value expected at position 115"}}'
  puppeteer:protocol:RECV ◀ ] +0ms
  puppeteer:webDriverBiDi:RECV ◀ [
[...]

PR #12815, opened 2024-07-23
fix: add missing partitionKey for page.deleteCookie()

minggeorgelei:
According to the devtools protocol interface DeleteCookiesRequest, partitionKey missed in interface DeleteCookiesRequest in Cookie.ts file. This PR is to add it to the file.

### DIFF
--- a/test/src/cookies.spec.ts
+++ b/test/src/cookies.spec.ts
@@ -380,21 +380,22 @@ describe('Cookie specs', () => {
       expect(await page.evaluate('document.cookie')).toBe('gridcookie=GRID');
     });
     it('should set a cookie with a partitionKey', async () => {
-      const {page, server} = await getTestState();
+      const {page, server, isChrome} = await getTestState();
 
       await page.goto(server.EMPTY_PAGE);
-      const url = 'https://example.com';
+      const url = new URL(page.url());
       await page.setCookie({
-        url,
+        url: url.toString(),
         name: 'partitionCookie',
         value: 'partition',
-        partitionKey: 'http://localhost',
+        secure: true,
+        partitionKey: url.origin,
       });
-      await expectCookieEquals(await page.cookies(url), [
+      await expectCookieEquals(await page.cookies(), [
         {
           name: 'partitionCookie',
           value: 'partition',
-          domain: 'example.com',
+          domain: url.hostname,
           path: '/',
           expires: -1,
           size: 24,
@@ -403,7 +404,9 @@ describe('Cookie specs', () => {
           session: true,
           sameParty: false,
           sourceScheme: 'Secure',
-          partitionKey: 'http://localhost',
+          partitionKey: isChrome
+            ? url.origin.replace(`:${url.port}`, '')
+            : url.origin,
         },
       ]);
     });
@@ -779,15 +782,14 @@ describe('Cookie specs', () => {
       expect(await page.cookies(COOKIE_DESTINATION_URL)).toHaveLength(0);
     });
     it('should only delete cookie without partition key if partitionkey is not specified', async () => {
-      const {page, server} = await getTestState();
+      const {page} = await getTestState();
       const url = 'https://example.com';
-
-      await page.goto(server.EMPTY_PAGE);
+      await page.goto(url);
       await page.setCookie({
         url,
         name: 'partitionCookie',
         value: 'partition',
-        partitionKey: 'http://localhost',
+        partitionKey: url,
       });
 
       await page.setCookie({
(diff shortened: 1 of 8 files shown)


OrKoN (changes requested):
According to the protocol documentation the partitionKey is currently experimental and was recently changed from a string to https://chromedevtools.github.io/devtools-protocol/tot/Network/#type-CookiePartitionKey 

Let's not add it at this point since it might be different in different Chrome versions. Since the API is cross-browser, we also need to make it work for WebDriver BiDi.

OrKoN:
cc @sadym-chromium related to https://github.com/puppeteer/puppeteer/issues/12787

OrKoN (commented):
I think we can land this now but we need to add a convertion step similar to the ones I have added here https://github.com/puppeteer/puppeteer/pull/12821/files Also, let's add a test. Eventually, we will change partitionKey to be an object in all Puppeteer APIs but for this PR I think we can keep it a string.

minggeorgelei:
> I think we can land this now but we need to add a convertion step similar to the ones I have added here https://github.com/puppeteer/puppeteer/pull/12821/files Also, let's add a test. Eventually, we will change partitionKey to be an object in all Puppeteer APIs but for this PR I think we can keep it a string.

ok, I will take it as a reference.

minggeorgelei:
I will add some tests later.

minggeorgelei:
Tests have been added.

minggeorgelei:
Hi @OrKoN can you help to check the failed tests. I think the tests passed with cdp protocol but failed with bidi protocol.

test/src/cookies.spec.ts
```diff
@@ -379,6 +379,34 @@ describe('Cookie specs', () => {
       await page.goto(server.PREFIX + '/grid.html');
       expect(await page.evaluate('document.cookie')).toBe('gridcookie=GRID');
     });
+    it('should set a cookie with a partitionKey', async () => {
```
OrKoN, line 382: so this test has to be something like this:

```
    it('should set a cookie with a partitionKey', async () => {
      const {page, server, isChrome} = await getTestState();

      const url = new URL(page.url());
      await page.setCookie({
        url: url.toString(),
        name: 'partitionCookie',
        value: 'partition',
        secure: true,
        partitionKey: url.origin,
      });
      await expectCookieEquals(await page.cookies(), [
        {
          name: 'partitionCookie',
          value: 'partition',
          domain: url.hostname,
          path: '/',
          expires: -1,
          size: 24,
          httpOnly: false,
          secure: true,
          session: true,
          sameParty: false,
          sourceScheme: 'Secure',
          partitionKey: isChrome ? url.origin.replace(`:${url.port}`, '') : url.origin,
        },
      ]);
    });
```

unfortunately, there are some differences in partitioning across browsers.

test/src/cookies.spec.ts
```diff
@@ -750,5 +778,87 @@ describe('Cookie specs', () => {
...
+        partitionKey: 'http://localhost',
+      });
+
+      await page.setCookie({
+        url,
+        name: 'partitionCookie',
+        value: 'partition',
+      });
+
+      await page.deleteCookie({
```
OrKoN, line 796: this test would not work since by default the partition key from the page is computed in setCookie and deleteCookie and it's all the same http://localhost used in tests. I think we should try changing the parition key to something else in the first setCookie

test/src/cookies.spec.ts
```diff
@@ -750,5 +778,87 @@ describe('Cookie specs', () => {
...
+          httpOnly: false,
+          secure: true,
+          session: true,
+          sameParty: false,
+          sourceScheme: 'Secure',
+          partitionKey: 'http://localhost',
+        },
+      ]);
+    });
+    it('should only delete cookie with partition key if partition key is specified', async () => {
```
OrKoN: I think same here, the partition key cannot be localhost since the page is at localhost (default partition key).

test/src/cookies.spec.ts
minggeorgelei: done

test/src/cookies.spec.ts
minggeorgelei: I cannot understand it totally. can you help to write a example test

test/src/cookies.spec.ts
minggeorgelei: done. changed to https://example.com/.

minggeorgelei:
After altering the tests, deleting cookie using bidi protocol still has some problems.

minggeorgelei:
> I think this is the best we can do with tests since the page API for cookies is rather limiting. We can probably have better tests after #645 and fully expose partition keys as is.

ok

Lightning00Blade (approved):

OrKoN (approved):

PR #12817, opened 2024-07-23
fix(webdriver): dispose resources to abort active listeners

### LINKED ISSUE #11849: [Bug] `should reject navigation when browser closes` is flaky in BiDi
See https://github.com/puppeteer/puppeteer/actions/runs/7802076934/job/21278705983#step:13:626

OrKoN:
Ending a WebDriver BiDi session disposes the browser but didn't dispose user contexts and frames. That caused waiters to still remain active and only time out after the timeout period.

Closes https://github.com/puppeteer/puppeteer/issues/11849

### DIFF
--- a/packages/puppeteer-core/src/bidi/core/UserContext.ts
+++ b/packages/puppeteer-core/src/bidi/core/UserContext.ts
@@ -71,7 +71,10 @@ export class UserContext extends EventEmitter<{
       new EventEmitter(this.browser)
     );
     browserEmitter.once('closed', ({reason}) => {
-      this.dispose(`User context already closed: ${reason}`);
+      this.dispose(`User context was closed: ${reason}`);
+    });
+    browserEmitter.once('disconnected', ({reason}) => {
+      this.dispose(`User context was closed: ${reason}`);
     });
 
     const sessionEmitter = this.#disposables.use(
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/bidi/core/Browser.ts
```diff
@@ -159,6 +159,9 @@ export class Browser extends EventEmitter<{
   dispose(reason?: string, closed = false): void {
     this.#closed = closed;
     this.#reason = reason;
+    for (const userContext of this.#userContexts.values()) {
```
Lightning00Blade: Why don't we just listen for the `close` event on the other object rather then poking directly here?

OrKoN: Could you elaborate on what you propose? the browser is disposed, so everything it owns should be disposed. Do you suggest that every object needs to be made aware of the session 'end' event?

Lightning00Blade (approved):

packages/puppeteer-core/src/bidi/core/UserContext.ts
```diff
@@ -71,7 +71,10 @@ export class UserContext extends EventEmitter<{
       new EventEmitter(this.browser)
     );
     browserEmitter.once('closed', ({reason}) => {
-      this.dispose(`User context already closed: ${reason}`);
+      this.dispose(`User context was closed: ${reason}`);
+    });
+    browserEmitter.once('disconnected', ({reason}) => {
```
Lightning00Blade, line 76: Note: `disconnected` is always emitted while close only when the browser is closed.

PR #12823, opened 2024-07-24
refactor!: use Uint8Array instead of Buffer for browser compatibility

Lightning00Blade:
We want users to be able to use Puppeteer in different environments, most notably in the browser.
Therefore, we want to move away from Node-specific API and rely on the ECMAScript APIs more and this
change replaces Node's Buffer in various Puppeteer APIs with [Uint8Array](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array).

If you want to convert Uint8Array to Buffer, call [`Buffer.from`](https://nodejs.org/api/buffer.html#static-method-bufferfromarraybuffer-byteoffset-length). For example, instead of

```
const buffer = await page.screenshot();
```

convert to Buffer explicitly:

```
const buffer = Buffer.from(await page.screenshot());
```

Many Node APIs accept Uint8Arrays so it should not be needed unless you want to use Buffer-specific APIs.

### DIFF
--- a/packages/puppeteer-core/src/api/HTTPResponse.ts
+++ b/packages/puppeteer-core/src/api/HTTPResponse.ts
@@ -89,14 +89,14 @@ export abstract class HTTPResponse {
    * failed to detect the correct encoding, the buffer might
    * be encoded incorrectly. See https://github.com/puppeteer/puppeteer/issues/6478.
    */
-  abstract buffer(): Promise<Buffer>;
+  abstract buffer(): Promise<Uint8Array>;
 
   /**
    * Promise which resolves to a text (utf8) representation of response body.
    */
   async text(): Promise<string> {
     const content = await this.buffer();
-    return content.toString('utf8');
+    return new TextDecoder().decode(content);
   }
 
   /**
--- a/packages/puppeteer-core/src/api/HTTPResponse.ts
+++ b/packages/puppeteer-core/src/api/HTTPResponse.ts
@@ -89,13 +89,20 @@ export abstract class HTTPResponse {
    * failed to detect the correct encoding, the buffer might
    * be encoded incorrectly. See https://github.com/puppeteer/puppeteer/issues/6478.
    */
-  abstract buffer(): Promise<Uint8Array>;
+  abstract content(): Promise<Uint8Array>;
 
+  /**
+   * {@inheritDoc HTTPResponse.content}
+   */
+  async buffer(): Promise<Buffer> {
+    const content = await this.content();
+    return Buffer.from(content);
+  }
   /**
    * Promise which resolves to a text (utf8) representation of response body.
    */
   async text(): Promise<string> {
-    const content = await this.buffer();
+    const content = await this.content();
     return new TextDecoder().decode(content);
   }
 
--- a/packages/puppeteer-core/src/util/typedArray.test.ts
+++ b/packages/puppeteer-core/src/util/typedArray.test.ts
@@ -8,9 +8,9 @@ import {describe, it} from 'node:test';
 
 import expect from 'expect';
 
-import {stringToTypedArray} from './typedArray.js';
+import {mergeUint8Arrays, stringToTypedArray} from './typedArray.js';
 
-describe('Function', function () {
+describe('Typed Array helpers', function () {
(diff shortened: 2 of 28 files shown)


packages/puppeteer-core/src/api/HTTPResponse.ts
```diff
@@ -89,14 +89,14 @@ export abstract class HTTPResponse {
    * failed to detect the correct encoding, the buffer might
    * be encoded incorrectly. See https://github.com/puppeteer/puppeteer/issues/6478.
    */
-  abstract buffer(): Promise<Buffer>;
+  abstract buffer(): Promise<Uint8Array>;
```
OrKoN: so looking at this diff, I am actually not sure we should remove Buffer from this API (since the getter's name is buffer). I wonder if we can keep it for Node users as it seems convenient but implement a method `array(): Promise<Uint8Array>` in addition. WDYT?

Lightning00Blade: I will only start looking at such things once everything is passing, but it was also on my mind.

packages/puppeteer-core/src/util/typedArray.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      expect(Buffer.from(body).compare(Buffer.from(result))).toBe(0);
+    });
+    it('should get body length from string with emoji', async () => {
+      const body = 'How Long is this string in bytes 📏?';
+      const result = stringToTypedArray(body);
+
+      expect(Buffer.from(body).compare(Buffer.from(result))).toBe(0);
+    });
+
+    it('should get body length from base64', async () => {
```
OrKoN: let's add some base64 tests that encoding not only single byte characters

packages/puppeteer-core/src/util/typedArray.test.ts
```diff
@@ -0,0 +1,52 @@
...
+      const body = btoa('Lorem ipsum dolor sit amet');
+      const result = stringToTypedArray(body, true);
+
+      expect(Buffer.from(body, 'base64').compare(Buffer.from(result))).toBe(0);
+    });
+  });
+
+  describe('mergeUint8Arrays', () => {
+    it('should work', () => {
```
OrKoN: let's add a test with some empty arrays?

test/src/requestinterception.spec.ts
```diff
@@ -14,7 +14,7 @@ import type {ConsoleMessage} from 'puppeteer-core/internal/common/ConsoleMessage
 import {getTestState, setupTestBrowserHooks} from './mocha-utils.js';
 import {isFavicon, waitEvent} from './utils.js';
 
-describe('request interception', function () {
+describe.only('request interception', function () {
```
OrKoN: remove .only

OrKoN (approved):

Ph0tonic:
Hi,
Could it be possible to update the release note of 23.0.0 to include this breaking changes ? Juste a single line to point to this issue would be great.
It would ease the update to this release.

- https://github.com/puppeteer/puppeteer/releases/tag/puppeteer-v23.0.0

Thank you very much.

Lightning00Blade:
@Ph0tonic You can find the the changes on https://github.com/puppeteer/puppeteer/releases/tag/puppeteer-core-v23.0.0
I will investigate linking the `puppeteer-core` and `puppeteer` page in releases.

Ph0tonic:
Ho great, thanks. Indeed, having a link to those changes would definitely help. :+1:

olivierr91:
@Lightning00Blade This also should be mentioned in the breaking changes of the CHANGELOG. Everything that breaks existing code should be listed in the breaking changes.

Lightning00Blade:
@Lightning00Blade Again the issue is due to the `puppeteer` and `puppeteer-core` not being sync - 
You can see that the `puppeteer-core` CHANGELOG.md includes this change.
https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/CHANGELOG.md#2300-2024-08-07

Issue tracked at - https://github.com/puppeteer/puppeteer/issues/12989

PR #12828, opened 2024-07-25
test: ignore favicon requests in RequestServedFromCache test

juliandescottes:
Updates a network.spec.ts test to ignore favicon requests.

### DIFF
--- a/test/src/network.spec.ts
+++ b/test/src/network.spec.ts
@@ -521,7 +521,7 @@ describe('network', function () {
 
         const cached: string[] = [];
         page.on('requestservedfromcache', r => {
-          return cached.push(r.url().split('/').pop()!);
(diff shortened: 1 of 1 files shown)


OrKoN:
Note that will be using Firefox stable very soon so we will need to wait for fixes for longer.

juliandescottes:
> Note that will be using Firefox stable very soon so we will need to wait for fixes for longer.

Oh good point, we shouldn't merge the test expectation changes. Would you still be OK with taking the fix for the test for now or do you prefer to wait until everything is in stable ?

Lightning00Blade:
I think the change to the test is valid, and we can merge that part.

juliandescottes:
Ok, I'll split the PR then, thanks!

Lightning00Blade (approved):

Issue #8836, opened 2022-08-24
[Feature]: Allow launching in incognito mode


stevenwdv:
Currently, `puppeteer.launch` opens a non-incognito context, and while one can open an incognito context via `createIncognitoBrowserContext`, this seems like an unnecessary hassle, especially since an extra context was created (and remains open).
Chromium supports launching in incognito mode via `-incognito`, but specifying this in `puppeteer.launch` let's Puppeteer still think that it's a non-incognito context: `isIncognito` returns `false`, which may cause some problems.

It would be nice to have a documented way of doing this, via a boolean option in `puppeter.launch`.

jrandolf:
This sounds very reasonable. I suppose we can add a small check that sees if the launch arguments contains incognito and flips the flag appropriately. Would that suffice?

stevenwdv:
I don't know the ins and outs, but if that's all it takes to properly set up an incognito session, you could do that. Although I think having a boolean would be a nicer way, and this could also be made compatible with Firefox (disclaimer: I haven't used puppeteer with FF so idk how this works).

jrandolf:
Hmm, now that I think about it, it would still only be a heuristic since someone could use a custom-chrome binary which defaults to incognito (as if someone **would** do that). I'll take a look at Firefox and come up with some solutions.

OrKoN:
I think it's actually a bug in Puppeteer which assumes that the default browser context is incognito. It should probably derive it from the actual running browser context. The entire `isIncognito` check in Puppeteer does not look very robust.

stevenwdv:
@OrKoN Don't you mean "assumes that the default browser context is *non-*incognito" ?

jacty:
Maybe we can prioritize the context created by user instead of using the default one? When user specified, no default one will be created?

OrKoN:
the default one is created by the browser itself and it is not explicitly created by Puppeteer.

OrKoN:
Closing in favor of https://github.com/puppeteer/puppeteer/issues/11686 We want to get rid of relying on the term "incognito" as it is Chrome-specific and instead we will only report if a given context is the default one or if it was created via Puppeteer's https://pptr.dev/api/puppeteer.browser.createbrowsercontext/ (where the word incognito has already been dropped).

PR #12830, opened 2024-07-26
refactor!: remove isIncognito

Lightning00Blade:
In Chrome, the default browser context can also be "incognito" if configured via the arguments and in such cases this getter returns wrong results (see https://github.com/puppeteer/puppeteer/issues/8836). Also, the term
 "incognito" is not applicable to other browsers. To migrate, check the default browser context instead: in
Chrome all non-default contexts are incognito, and the default context might be incognito if you provide the `--incognito` argument when launching the browser.


test/src/browsercontext.spec.ts
```diff
@@ -14,33 +14,27 @@ import {waitEvent} from './utils.js';
 describe('BrowserContext', function () {
   setupTestBrowserHooks();
 
-  it('should have default context', async () => {
+  it.only('should have default context', async () => {
```
OrKoN: remove .only

packages/puppeteer-core/src/api/BrowserContext.ts
```diff
@@ -174,26 +174,6 @@ export abstract class BrowserContext extends EventEmitter<BrowserContextEvents>
...
   abstract pages(): Promise<Page[]>;
 
-  /**
-   * Whether this {@link BrowserContext | browser context} is incognito.
-   *
-   * In Chrome, the
-   * {@link Browser.defaultBrowserContext | default browser context} is the only
-   * non-incognito browser context.
-   *
-   * @deprecated In Chrome, the
```
OrKoN, line 184: let's move the description about what type of contexts and how many are created in Firefox and in Chrome to the documentation of the BrowserContext class

OrKoN (approved):

OrKoN (approved):

PR #12840, opened 2024-07-28
feat: support passing `capabilities` to bidi

izaakschroeder:

Feature.

Not yet, have to figure out how to do this.

Updated doc comments?

**Summary**

If using a hub intermediary (e.g. one day BrowserStack supports native BiDi) then passing `capabilities` to a session is kinda necessary to decide which instance to forward the websocket to.

There are two other points where `BidiBrowser.create` is called but they appear to be for launching the browser directly from the user's machine. Perhaps there is a use case here for passing capabilities… but it seemed less valuable and so has been omitted from this change currently.

No.

**Other information**

N/A

### DIFF
--- a/packages/puppeteer-core/src/bidi/Browser.ts
+++ b/packages/puppeteer-core/src/bidi/Browser.ts
@@ -39,6 +39,7 @@ export interface BidiBrowserOptions {
   process?: ChildProcess;
   closeCallback?: BrowserCloseCallback;
   connection: BidiConnection;
+  capabilities?: Bidi.Session.CapabilitiesRequest;
   cdpConnection?: CdpConnection;
   defaultViewport: Viewport | null;
   ignoreHTTPSErrors?: boolean;
@@ -71,11 +72,13 @@ export class BidiBrowser extends Browser {
 
   static async create(opts: BidiBrowserOptions): Promise<BidiBrowser> {
     const session = await Session.from(opts.connection, {
+      ...opts.capabilities,
       alwaysMatch: {
         acceptInsecureCerts: opts.ignoreHTTPSErrors,
         unhandledPromptBehavior: {
           default: Bidi.Session.UserPromptHandlerType.Ignore,
         },
+        ...opts.capabilities?.alwaysMatch,
         webSocketUrl: true,
       },
     });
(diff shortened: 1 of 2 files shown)


OrKoN (commented):
Thanks for the PR but we do not plan to expose BiDi capabilities directly since it might be confusing to the user given that Puppeteer presets own capabilities and those won't be applicable to both browsers we currently support in the same way. Could you provide more details about what specific capabilities are needed for your use case?

izaakschroeder:
> Thanks for the PR but we do not plan to expose BiDi capabilities directly since it might be confusing to the user given that Puppeteer presets own capabilities and those won't be applicable to both browsers we currently support in the same way. Could you provide more details about what specific capabilities are needed for your use case?

```ts

  browserURL: "http://some.hub:4444/session",
  protocol: "webDriverBiDi",
  capabilities: {
    "browserName": "firefox"
  }
});
```

OrKoN:
Thanks for the details, I think to support this we should add the `product` param to connect options (currently, it only exists in launch options) and set the capabilities request based on that.

izaakschroeder:
> Thanks for the details, I think to support this we should add the `product` param to connect options (currently, it only exists in launch options) and set the capabilities request based on that.

I mean… I guess that's a start… but it still probably kinda falls apart as users want to send more arbitrary data to a hub? `browserVersion`, `moz:firefoxOptions`, authentication parameters, etc.

OrKoN:
It might be not sufficient but I think we should proceed use case by user case. We already have some existing options that map to capabilities such as ignoreHTTPSErrors and we also configure unhandledPromptBehavior. For example, if a different unhandledPromptBehavior configuration is provided Puppeteer might simply break as it would not match Puppeteer expectations. 

As for the browser version, Puppeteer is only guaranteed to work with the [pinned version](https://pptr.dev/supported-browsers) (the page include Firefox later once we start pinning too) and perhaps that version should be used in the capabilities. We could allow overriding it but there is a chance that it won't be compatible.

izaakschroeder:
Thanks for landing this 🎉

OrKoN:
I think BrowserStack [already has support](https://www.browserstack.com/docs/automate/selenium/bidi-event-driven-testing) for WebDriver BiDi. I wonder if puppeteer.connect with capabilities should already be compatible with BrowserStack. I wonder if there is some way for us to test it cc @automatedtester @mathiasbynens

PR #12842, opened 2024-07-29
docs:updated docs so that the selector param description is consistent

camfung:
Docs update so that the description for selector is consistent with other descriptions for it

No

**Summary**
I noticed that the selector didn't link to the docs on mdn so i added it to the Page.click docs. 

I wanted to add this because when i was looking through the docs for clicking and the selector seamed ambiguous to me.

no

### DIFF
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -2674,7 +2674,7 @@ export abstract class Page extends EventEmitter<PageEvents> {
    * Shortcut for {@link Frame.click | page.mainFrame().click(selector[, options]) }.
    * @param selector - A
    * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors | selector}
-   * of an element to type into. If there are multiple elements satisfying the
+   * of an element to click. If there are multiple elements satisfying the
    * selector, the first will be clicked.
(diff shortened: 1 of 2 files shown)


OrKoN (changes requested):
could you please update the comment in the source .ts file and run `npm run docs` to regenerated the markdown files? thanks!

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -2672,8 +2672,10 @@ export abstract class Page extends EventEmitter<PageEvents> {
    * ```
    *
    * Shortcut for {@link Frame.click | page.mainFrame().click(selector[, options]) }.
-   * @param selector - A `selector` to search for element to click. If there are
-   * multiple elements satisfying the `selector`, the first will be clicked
+   * @param selector - A
+   * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Selectors | selector}
+   * of an element to type into. If there are multiple elements satisfying the
```
OrKoN: I think this should say "to click" instead of "to type into"?

camfung: Shoot sorry thats a dumb typo

OrKoN:
Could you please sign the CLA as requested by the CLA check?

PR #12845, opened 2024-07-29
chore: remove TODOs

Lightning00Blade:


packages/puppeteer-core/src/cdp/LifecycleWatcher.ts
```diff
@@ -248,10 +248,6 @@ export class LifecycleWatcher {
           return false;
         }
       }
-      // TODO(#1): Its possible we don't need this check
-      // CDP provided the correct order for Loading Events
-      // And NetworkIdle is a global state
-      // Consider removing
       for (const child of frame.childFrames()) {
```
Lightning00Blade, line 251: We do need this check as we are looking for network idle events as well.

tools/docgen/src/custom_markdown_documenter.ts
```diff
@@ -1574,14 +1574,6 @@ export class MarkdownDocumenter {
       const qualifiedName = Utilities.getSafeFilenameForName(
         hierarchyItem.displayName
       );
-      // TODO: Remove this and add suffix #hash to the header
-      if (ApiParameterListMixin.isBaseClassOf(hierarchyItem)) {
```
Lightning00Blade, line 1578: While we can provide the precise Overload, this does not look correct when non is specified (Always), the first is give.

OrKoN (approved):

Issue #12861, opened 2024-07-31
[Feature]: PUPPETEER_DOWNLOAD_BASE_URL should handle trailing slashes


paulmilbank:
### Feature description

While troubleshooting an older version of Puppeteer hanging while downloading Chrome, one of the suggestions was to set PUPPETEER_DOWNLOAD_BASE_URL to `https://storage.googleapis.com/chrome-for-testing-public/`
https://github.com/puppeteer/puppeteer/issues/12833

This led to a failure due to a 404. The constructed URL has a `//` in it. 
<img width="1143" alt="image" src="https://github.com/user-attachments/assets/f5e6a67f-51f8-465f-b895-799f3f48025a">

Setting PUPPETEER_DOWNLOAD_BASE_URL to `https://storage.googleapis.com/chrome-for-testing-public` did fix this. 

I thought it might be useful to normalise PUPPETEER_DOWNLOAD_BASE_URL to strip off trailing slashes as that is what the @puppeteer/browsers module expects.

PR #12862, opened 2024-07-31
fix: strip trailing slashes from downloadBaseUrl

paulmilbank:
feature

No, happy to do this, but unsure of the best place to add this

No, this shouldn't need a documentation update

**Summary**
https://github.com/puppeteer/puppeteer/issues/12861
I encountered an issue with the download url getting a 404 when setting `PUPPETEER_DOWNLOAD_BASE_URL` to `https://storage.googleapis.com/chrome-for-testing-public/` per a comment on an issue in the Puppeteer repo.
@puppeteer/browsers expects the url to not have a trailing `/`
I thought it might be useful to strip trailing slashes off so the URL is in the expected format.

no

**Other information**


paulmilbank:
This may be the wrong place to put this. I'm guessing that we want to retain the `string |  undefined` type and not coerce it to a string here so that the default baseUrl parameters in the browsers package take precedence. Possibly if this is useful, it should be implemented in the browsers package in the `resolveDownloadUrl` functions.

OrKoN (commented):
Thanks for the PR. Could you please run `npm run format`?

Lightning00Blade:
I am not sure if we should make this changes as `example.com/A/B//C` and `example.com//A/B/C` are both valid URLs.
And any URL with path that contains 2 slashes next to each other.
Internally we convert the string to `URL` object so that makes the verification process. 
Maybe a note on the documentation would be a better suited for this rather than changing the behavior.
@OrKoN WDYT?

OrKoN:
I think it makes sense. Let's leave it as is and document that we will append a value starting with `/` to the base URL.

Lightning00Blade:
Closing per last comment will keep track on the issue.

PR #12874, opened 2024-08-05
build: fix browser update roller

OrKoN:
push-to-fork does not support labels. We can only test it once there is a new version of Chrome.


OrKoN:
@Lightning00Blade PTAL

Lightning00Blade (approved):

PR #12895, opened 2024-08-06
docs: update main example

OrKoN:
And sync README description


Lightning00Blade (commented):
I am not sure about changing the example.

For me it's a little bit confusing.

OrKoN:
> I am not sure about changing the example.
> 
> For me it's a little bit confusing.

could you be more specific?

OrKoN:
Extracted description update to https://github.com/puppeteer/puppeteer/pull/12899

README.md
```diff
@@ -16,39 +16,58 @@
...
+const chrome = await puppeteer.launch({
+  browser: 'chrome',
+});
 
-// Navigate the page to a URL.
-await page.goto('https://developer.chrome.com/');
+// Launch Firefox.
+const firefox = await puppeteer.launch({
+  browser: 'firefox',
+});
```
matthiasrohmer, line 35: I wonder if we should make both use BiDi...? Can't decide what's the best message to send: you can do all the things equally in Firefox and Chrome using CDP and BiDi *or* look you can do the same things with the same protocol in both browsers.

I am leaning towards the later.

matthiasrohmer (approved):
Thanks for updating Alex! I think this is a good enough start for now - I'll start working on the homepage update next week and then think of probably 3 specific examples that highlight certain automations.

One thought, that's going nowhere: automating the visit to a site that's different in Firefox and Chrome could be fun, like whatismybrowser.com - but using a 3P site as an example is bad. And I am saying this as the person formerly being responsible for d.c.c's search and constantly wondering about the amount of queries for `automate beyond recorder` :)

Issue #12900, opened 2024-08-07
[Bug]: Postinstall script for downloading Chrome throws error when using Yarn P'n'P (Plug 'n' Play)


islazh:
### Minimal, reproducible example

```TypeScript
Please see `https://github.com/islazh/yarn-pnp-puppeteer-repro`. Make sure that there are no existing Chrome downloads in the Puppeteer cache.
```

### Background

We have been trying to use Yarn Plug 'n' Play to install dependencies including Puppeteer.

### Expectation

The postinstall script in Puppeteer downloads Chrome which is then used to run Puppeteer tests.

### Reality

An error is thrown because the path to `'puppeteer/internal/node/install.js'` cannot be resolved.
```
Skipping browser installation because the Puppeteer build is not available. Run npm install again after you have re-built Puppeteer.
```
This seems to happen because Yarn PnP "unplugs" Puppeteer during the postinstall - that is, it unzips the code in order to do the postinstall. However, Puppeteer seems to not be able to resolve its dependencies in PnP mode, so it cannot resolve @puppeteer/browsers. 

If I log out the error from `https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer/install.mjs`, I get this:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@puppeteer/browsers' imported from /Users/isla.zhang/Projects/eddrp-participant/.yarn/unplugged/puppeteer-npm-22.15.0-fa225bdb63/node_modules/puppeteer/lib/esm/puppeteer/node/install.js
```

### Puppeteer version

22.15.0

### Node version

20.14.0

### Package manager

yarn

### Package manager version

4.2.2

### Operating system

Linux

OrKoN:
Puppeteer does relies on its dependent packages to download the browser during the installation (in postinstall hook). It does sound like it's the issue with yarn P'n'P? You can skip download during package installation (`PUPPETEER_SKIP_DOWNLOAD=true`) and use `npx puppeteer browsers install chrome` (or yarn's alternative for running package binaries) to download pinned browsers after yarn install. I don't think we care going to change the way we use dependencies during installation but perhaps yarn can support this scenario?

OrKoN:
@Lightning00Blade seems to have additional context on this issue

Lightning00Blade:
I also agree that this is most likely an issue with Yarn PnP.
[...]

PR #12904, opened 2024-08-07
fix: use CJS for Puppeteer CLI

Lightning00Blade:
Related to #12900 

While we set the `"type": "commonjs"` we should only use CJS in bin and  main.


OrKoN (approved):
we do generate a package.json with {"type": "module"} in the esm folder so it should probably work but the change LGTM

Issue #12917, opened 2024-08-08
[Bug]: Clean script can cause data loss with vendored Puppeteer


jgraham:
When we vendor Puppeteer tests into mozilla-central we don't include the `.git` directory because we don't need history, and having multiple nested repos causes problems.

However we recently had a data-loss issue where a Firefox engineer tried to run the puppeteer tests from their hg checkout of mozilla-central, which they had placed inside a git repo containing their config files, with ignore rules used to exclude their work from source control.

In this scenario running [`git clean -Xf`](https://github.com/puppeteer/puppeteer/blob/main/tools/clean.mjs#L13) is a *really* bad idea because it's going to delete whatever is ignored in that git repo, which in the reported case seemed to be most of the home directory.

The same concern also applies if you vendor Puppeteer into a git-based project; in this case anything that invokes the clean script is going to delete all ignored files (e.g. build artifacts, downloaded assets, whatever) from the parent repo.

I don't know exactly what problem the clean script needs to solve, but it seems like it should either have a specific list of directories to remove, rather than relying on the VCS config, or at least confirm that the root of the git checkout is as expected before performing VCS actions.

(We will also look at this on the mozilla-central side to see if we should change the way we're vendoring or invoking the tests, to avoid these problems).

OrKoN:
Do you run the clean script directly and not through the NPM script `npm run clean` defined Puppeteer?

jgraham:
I think it's just running `npm run test:firefox` c.f. https://searchfox.org/mozilla-central/source/remote/mach_commands.py#419-527

OrKoN:
The test commands do not run clean and when the clean is called there it looks like cwd is set correctly https://searchfox.org/mozilla-central/source/remote/mach_commands.py#751 The git clean command is called with args which is the list of files in the current cwd so generally it should not remove anything else.

OrKoN:
What's the git version that was in use? maybe older versions didn't support args?

OrKoN:
I think the mitigation is in place.

PR #12919, opened 2024-08-08
build: abort git clean if the git root is not puppeteer

Lightning00Blade:
Addresses https://github.com/puppeteer/puppeteer/issues/12917

When running though NPM the command will have a proper `cwd` to the selected package so the we will remove only things in Puppeteer. We only do this for individual packages so if people have things on the top level that will be kept around.

### DIFF
--- a/tools/clean.mjs
+++ b/tools/clean.mjs
@@ -18,10 +18,9 @@ const confirmation = new Promise((res, rej) => {
 });
 
 if (process.env.npm_command !== 'run-script') {
-  console.log(
-    'Running command outside from non Puppeteer directory may result in data loss'
-  );
+  console.log('Running command directly may result in data loss.');
   console.log('Ref: https://github.com/puppeteer/puppeteer/issues/12917');
+  console.log('Please use the provided npm scripts!');
   const rl = readline.createInterface({
     input: process.stdin,
     output: process.stdout,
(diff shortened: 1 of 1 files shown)


tools/clean.mjs
```diff
@@ -8,11 +8,48 @@
...
-    .join(' ')}`
-);
+let resolver;
+let rejector;
+const confirmation = new Promise((res, rej) => {
+  resolver = res;
+  rejector = rej;
+});
+
+if (process.env.npm_command !== 'run-script') {
```
OrKoN: why is it important if it is run via npm or not? let's just check if `git rev-parse --show-toplevel` is the process.cwd() and abort cleaning if git root is wrong

OrKoN: It still can be invoked without .git folder in puppeteer with an npm script defined elsewhere but the process.cwd() cannot probably used for checking since it could be a subfolder in Puppeteer

OrKoN: let's check if the directory in which tools is located is the same as `git rev-parse --show-toplevel` and bail otherwise

OrKoN: maybe we should list targets for deletion explicitly but it would be somewhat tedious to maintain it

Lightning00Blade: Done.

OrKoN (approved):

OrKoN (approved):

PR #12950, opened 2024-08-14
feat: improve type inference for selectors by adopting "typed-query-selector"

g-plane:

Adopt "[typed-query-selector](https://github.com/g-plane/typed-query-selector)" to parse CSS selector.

Yes.

No.

**Summary**

The current implementation of CSS selector parser can't handle some cases. For example, it can't parse `a:is([href], [href])` which is just `a`. Fixing these cases will highly increase the complexity, so I introduce the "typed-query-selector". Lighthouse is using it:
https://github.com/GoogleChrome/lighthouse/blob/c79628af9bdaa537a2abd1b34da922e28b81bd98/package.json#L178

typed-query-selector is well tested: https://github.com/g-plane/typed-query-selector/blob/main/parser.test.ts

No.

### DIFF
--- a/docs/api/puppeteer.nodefor.md
+++ b/docs/api/puppeteer.nodefor.md
@@ -8,13 +8,5 @@ sidebar_label: NodeFor
 
 ```typescript
 export type NodeFor<ComplexSelector extends string> =
-  TypeSelectorOfComplexSelector<ComplexSelector> extends infer TypeSelector
-    ? TypeSelector extends
-        | keyof HTMLElementTagNameMap
-        | keyof SVGElementTagNameMap
-      ? ElementFor<TypeSelector>
-      : Element
-    : never;
+  ParseSelector<ComplexSelector>;
 ```
-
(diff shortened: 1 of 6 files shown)


OrKoN:
Hey, thanks for the PR, it looks interesting. What is the minimal TypeScript version that you support/test with? Also, could you please run `npm run docs` to update documentation generated from code.

OrKoN:
I have rebuilt the docs so never mind about that.

g-plane:
It requires at least TypeScript 4.3, as I tested:

image

No output means no errors.

OrKoN:
@g-plane how does the library deal with pseudo-elements? Puppeteer defines custom pseudo elements as an optional-extension to the selector syntax that allows selection based on custom strategies. We have added a test for this case which showed a difference compared to typed-query-selector.

g-plane:
I think we can add a special case for custom pseudo elements, such as `::-...`.

OrKoN (approved):

PR #12977, opened 2024-08-19
ci(browsers): pin Firefox under test to stable

Lightning00Blade:
Closes https://github.com/puppeteer/puppeteer/pull/12973


OrKoN (approved):

packages/browsers/test/src/versions.ts
```diff
@@ -6,6 +6,6 @@
 
 export const testChromeBuildId = '127.0.6533.72';
 export const testChromiumBuildId = '1083080';
-export const testFirefoxBuildId = '130.0a1';
+export const testFirefoxBuildId = 'stable_129.0';
```
OrKoN, line 9: probably local copy for the test server needs to be changed as well

OrKoN: although I expected everything to just work

OrKoN: looks like a bug in resolveDownloadPath in firefox.ts We consider any version to be downloaded into archiveNightly() location

Lightning00Blade: That should be only for nightly, no? When we give `stable_` prefix we should never hit it.

Edit: we use it directly in the test server

OrKoN: resolveDownloadPath does not consider the prefix yet but it should be updated to return different results based on the prefix, i.e., use archive() instead of archiveNightly() for non-nightly prefixes.

Lightning00Blade: Download paths is the wrong thing tho use in downloadTestBrowsers, as we want to set it up such that our server response with the correct file.

packages/browsers/tools/downloadTestBrowsers.mjs
```diff
@@ -65,6 +65,21 @@ for (const version of Object.keys(versions)) {
       recursive: true,
     });
     copyFileSync(archivePath, targetPath);
+
+    // Needed to test Firefox nightly build download
```
OrKoN, line 69: could you please revert changes in this file

OrKoN (approved):

PR #13024, opened 2024-08-30
performance: faster implementation of typedArrayToBase64

aguidrevitch:

Speedup for processing response bodies.

**Summary**
Switch from 22.12.1 to 22.13.0 made my test fail with timeout, most of tests stopped to fit within default 5 second interval. Reverting to 22.12.1 made them fast again. I found that the reason for that is typedArrayToBase64(), which processes body one byte at a time. By using String.fromCodePoint.apply in chunks, it is possible to significantly speedup the puppeteer with request interception.

No

**Other information**


OrKoN (approved):
Thanks!

PR #13025, opened 2024-08-30
docs: start to unify example resources

matthiasrohmer:
That's a first stab on unifying the various example resources for Puppeteer.

To do so this PR creates a new docs page /examples and adds it to the sidebar as "Examples & Use Cases". This page at some time could potentially evolve into a full section, similar to Guides, but for a start it's simply meant to make the existing examples discoverable and maintainable.

This new page includes the resources list from https://developer.chrome.com/docs/puppeteer/examples, which is meant to be deleted and redirected to pptr.dev/examples soon. I did go over the list and check for dead links/redirects/deprecations. This list was also duplicated to https://github.com/puppeteer/puppeteer/tree/main/examples#other-resources, which I removed and instead also added a link to pptr.dev/examples while streamlining the instructions to run the samples.

Additionally the new page has explicit pointers to https://github.com/puppeteer/puppeteer/tree/main/examples
and https://github.com/puppeteer/examples which I'll both clean-up and extend with new examples and delete obsolete ones.

### DIFF
--- a/docs/examples.md
+++ b/docs/examples.md
@@ -1,0 +1,67 @@
+# Examples & Use cases
+
+## Official examples
+
+[The Puppeteer repository](https://github.com/puppeteer/puppeteer/tree/main/examples) includes a small number of examples maintained by the Puppeteer team.
+
+Follow the instructions in the README to run the examples, covering use cases like creating PDFs from websites, creating screenshots or intercepting requests.
+
+## Example suite
+
+Find a set of unstructured examples in Puppeteer's dedicated [example repository](https://github.com/puppeteer/examples).
+
+This suite is a collection of examples that has been growing over time and covers various use cases like forwarding events
+from your Puppeteer process to the browser, interacting with elements and running CDP commands.
+
+## Other projects, articles and demos
+
+See the following list for use cases and examples from categories like Rendering, Web scraping and Testing.
+
+### Rendering and web scraping
+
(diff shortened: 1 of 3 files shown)


matthiasrohmer:
@OrKoN, PTAL.

OrKoN (approved):

docs/examples.md
```diff
@@ -0,0 +1,67 @@
...
+Find a set of unstructured examples in Puppeteer's dedicated [example repository](https://github.com/puppeteer/examples).
+
+This suite is a collection of examples that has been growing over time and covers various use cases like forwarding events
+from your Puppeteer process to the browser, interacting with elements and running CDP commands.
+
+## Other projects, articles and demos
+
+See the following list for use cases and examples from categories like Rendering, Web scraping and Testing.
+
+### Rendering and web scraping
```
OrKoN, line 20: Many projects/examples here seem somewhat outdated. Do you think we should keep referring to them or try to update them? Or perhaps we should discover some newer maintained examples? WDYT?

matthiasrohmer: Yep they definitely are, pondered about the same when going through them... I think trying to update the list to more current/maintained ones would be good.

However I could also imagine keeping some of the older ones around to inspire beyond usual tooling, Puppetron for example is a fun one - although it's arguably outdated too.

I'll document thoughts and ideas more properly next week and share with you :) while I was collecting everything today I just thought why not get at least this landed quickly.

PR #13052, opened 2024-09-05
docs: update cross-browser example to use browser instead of product

juliandescottes:


juliandescottes:
Noticed this example was still using the outdated syntax.

OrKoN (approved):

Issue #13050, opened 2024-09-04
[Bug]: possibly easy fix? A4 dimensions doesn't match actual a4 specification


DCh06:
### Minimal, reproducible example

```TypeScript
// works
export const createPDFA4 = async (htmlString) => {
    const browser = await puppeteer.launch();
    const page = await browser.newPage();
    await page.setContent(htmlString);
    const pdfBuffer = await page.pdf({path:'x.pdf' , width: "210mm", height: "297mm" });
    await browser.close();

    return Buffer.from(pdfBuffer)
};

// doesnt work
export const createPDFA4 = async (htmlString) => {
    const browser = await puppeteer.launch();
    const page = await browser.newPage();
    await page.setContent(htmlString);
    const pdfBuffer = await page.pdf({path:'x.pdf' , format: 'A4' });
    await browser.close();

    return Buffer.from(pdfBuffer)
};
```

### Background

Im trying to create pdf. 3rd party app has a4 validation. When created with ...format: 'A4'..., validator rejects my file. Creating it with set dimensions doesnt have this problem.

Tested on macOS and windows 10

### Expectation

format:'A4' to create exact A4 format 210x297mm

### Reality

there is very small difference, cannot pass through some A4 validators, final format is 210.227 x 297.349

### Puppeteer version

23.2.2

### Node version

v20.9.0

### Package manager

npm

### Package manager version

10.1.0

### Operating system

macOS

OrKoN:
Yeah, Puppeteer defines a4 in inches (8.27 x 11.7 in) because Chrome DevTools Protocol accepts inches for printing and that is probably not precise enough for validators.

hangyeol0531:
I requested PR related to the issue.

https://github.com/puppeteer/puppeteer/pull/13087

OrKoN:
should be closed via https://github.com/puppeteer/puppeteer/pull/13087

PR #13087, opened 2024-09-14
fix: improve Precision of Paper Sizes in Inches to 4 Decimal Places

hangyeol0531:
This is related to the issue below.

https://github.com/puppeteer/puppeteer/issues/13050

By increasing the decimal places to 4, the accuracy was improved, and this resolved the issue.

The values were adjusted according to the **ISO 216** A series.

As you can see in the two pictures below, **the page size is the same.** (It was different before the code change)
```
// works
const correctCreatePDFA4 = async (htmlString) => {
        executablePath: '/usr/local/bin/chromium',
        headless: true,
    });
    await page.setContent(htmlString);
    const pdfBuffer = await page.pdf({path:'correctCreatePDFA4.pdf' , width: "210mm", height: "297mm" });
    await browser.close();
    return Buffer.from(pdfBuffer)
};

// doesnt work
const incorrectCreatePDFA4 = async (htmlString) => {
        executablePath: '/usr/local/bin/chromium',
        headless: true,
    });
    await page.setContent(htmlString);
    const pdfBuffer = await page.pdf({path:'incorrectCreatePDFA4.pdf' , format: 'A4' });
    await browser.close();
    return Buffer.from(pdfBuffer)
};

(async () => {
    await correctCreatePDFA4(htmlString)
    await incorrectCreatePDFA4(htmlString)
})();
```

incorrect

correct


OrKoN:
Thanks!

OrKoN (approved):

Issue #13084, opened 2024-09-12
[Feature]: Distinguish between different `Touch`es


emilefokkema:
### Feature description

Currently, the [`touchMove()`](https://pptr.dev/api/puppeteer.touchscreen.touchmove) method on the `TouchScreen` class takes two parameters, `x` and `y`. But there is no way to specify _which_ of the touches you want to move. (Because multiple touches [can be active](https://developer.mozilla.org/en-US/docs/Web/API/TouchEvent/targetTouches) at any given time.)

Proposal: change the signature of [`TouchScreen.touchStart`](https://github.com/puppeteer/puppeteer/blob/280e6f82065a2cbbe696673934a9c10610a780c6/packages/puppeteer-core/src/api/Input.ts#L498) to
```ts
abstract touchStart(x: number, y: number): Promise<Touch>
```
where `Touch` would be something like
```ts
interface Touch {
    move(x: number, y: number): Promise<void>
    end(): Promise<void>
}
```
This way it would be perfectly clear which of the touches you're moving or ending. (The presence of the optional `id` property [in the parameters](https://chromedevtools.github.io/devtools-protocol/tot/Input/#type-TouchPoint) to the CDP `Input.dispatchTouchEvent` command at least suggests that something like this is possible.)

lballard625015:
Can I work on this one?

OrKoN:
We need to consider that the API should compatible with WebDriver BiDi as well: https://w3c.github.io/webdriver-bidi/#command-input-performActions

In general, I think the proposed API makes sense.

emilefokkema:
@OrKoN @Lightning00Blade One way to answer my question above is to look at the current behavior. [This test](https://github.com/emilefokkema/puppeteer/commit/45469a703ae3bee7788c33e534b57ce6b2e734d3) (run against chrome-headless) seems to show that the current behavior of

```ts
await page.touchscreen.touchStart(100, 100);
await page.touchscreen.touchStart(200, 200);
await page.touchscreen.touchMove(150, 150);
```

is exactly the same as if it had been

```ts
await page.touchscreen.touchStart(100, 100);
await page.touchscreen.touchMove(200, 200);
```
The second `touchStart` does not result in a new touch at all. So I don't know if this new feature can be implemented in a backwards-compatible way.

emilefokkema:
[...]

PR #13111, opened 2024-09-21
feat: add touch specifier (#13084)

ritk20:

Feature added to specify which of the touches you want to move in [`TouchScreen.touchStart`](https://github.com/puppeteer/puppeteer/blob/280e6f82065a2cbbe696673934a9c10610a780c6/packages/puppeteer-core/src/api/Input.ts#L498)

No, the base tests were fine

Yes, Updated puppeteer.touchscreen.touchstart.md in docs/api and puppeteer.touchscreen.touchstart.md in website/versioned_docs/version-23.4.0/api/

**Summary**
Added feature for [this](https://github.com/puppeteer/puppeteer/issues/13084) pull request

No

**Other information**

### DIFF
--- a/docs/api/puppeteer.touchscreen.touchstart.md
+++ b/docs/api/puppeteer.touchscreen.touchstart.md
@@ -10,7 +10,7 @@ Dispatches a `touchstart` event.
 
 ```typescript
 class Touchscreen {
-  abstract touchStart(x: number, y: number): Promise<void>;
+  abstract touchStart(x: number, y: number): Promise<Touch>;
 }
 ```
 
--- a/packages/puppeteer-core/src/api/Input.ts
+++ b/packages/puppeteer-core/src/api/Input.ts
@@ -34,6 +34,14 @@ export interface KeyboardTypeOptions {
 /**
  * @public
  */
+export interface Touch {
+  move(x: number, y: number): Promise<void>;
+  end(): Promise<void>;
+}
+
+/**
+ * @public
+ */
 export type KeyPressOptions = KeyDownOptions & KeyboardTypeOptions;
 
 /**
@@ -495,7 +503,7 @@ export abstract class Touchscreen {
    * @param x - Horizontal position of the tap.
    * @param y - Vertical position of the tap.
    */
-  abstract touchStart(x: number, y: number): Promise<void>;
+  abstract touchStart(x: number, y: number): Promise<Touch>;
 
   /**
    * Dispatches a `touchMove` event.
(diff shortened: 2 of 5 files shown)


emilefokkema:
I think at least one test is definitely needed here:

```ts
// arrange
const touch1 = await page.touchscreen.touchStart(100, 100);
const touch2 = await page.touchscreen.touchStart(200, 200);

// act
await touch1.move(150, 150);

// assert that the first touch has moved, and not the second

```

Lightning00Blade:
Please add a test using the return variable from `touchStart`. The test can mimic the one in `Touchscreen.prototype.touchMove` and should be for `Touchscreen.prototype.touchStart`, but using the newly added feature.

ritk20:
Thanks for the help. Will add the tests soon.

emilefokkema:
@Lightning00Blade But what about

```ts
const touch1 = await page.touchscreen.touchStart(100, 100);
const touch2 = await page.touchscreen.touchStart(200, 200);

await page.touchscreen.touchMove(150, 150); // which of the two touches should now move?
```
?

Lightning00Blade:
I miss that comment, and now looking closer to the request I don't think the current implementation here would work, as we never provide the unique ID that CDP and WebDriver BiDi require to make the diffrence.

Also the test you suggested look like a good candidate for this indeed.

ritk20:
> @Lightning00Blade But what about
> 
> ```ts
> const touch1 = await page.touchscreen.touchStart(100, 100);
> const touch2 = await page.touchscreen.touchStart(200, 200);
> 
> await page.touchscreen.touchMove(150, 150); // which of the two touches should now move?
> ```
> 
> ?

Yes, I figured my current implementation wouldn't work with this. Trying something like this now
`  #touchPoints: Array<{ id: number, x: number, y: number, radiusX: number, radiusY: number, force: number }> = [];`

where I pass `const id = this.#touchPoints.length + 1;` in `touchStart`. I'm trying to get that to work correctly now.

Also, in the test you suggested, `touch2` should move right?

emilefokkema:
@ritk20 I suggested a test and I asked a question to which I do not know the answer. For the test I suggested, see above. As for the question, that is:
```ts
const touch1 = await page.touchscreen.touchStart(100, 100);
const touch2 = await page.touchscreen.touchStart(200, 200);

await page.touchscreen.touchMove(150, 150); // which of the two touches should now move?
```
I don't know the answer to that one. Maybe @Lightning00Blade ?

emilefokkema:
@ritk20 As for the question I asked, I had a look at the current behavior. Please refer to [this comment](https://github.com/puppeteer/puppeteer/issues/13084#issuecomment-2372269950)

Lightning00Blade:
Closed by https://github.com/puppeteer/puppeteer/pull/13231

Issue #13152, opened 2024-10-03
[Bug]: Dynamically added shadow DOM elements cannot be found


zooltar:
### Minimal, reproducible example

```TypeScript
import puppeteer from "puppeteer";
import fs from "fs";

const pageContent = fs.readFileSync("test.html").toString();
(async () => {
    const browser = await puppeteer.launch({ headless: false });
    const page = await browser.newPage();
    await page.setContent(pageContent);
    let shadowDiv = await page.waitForSelector("div[id=container]", {
        timeout: 5000,
        visible: true,
    });
    const shadowH1 = await shadowDiv.$eval(">>> h1", (h1) => h1.innerText);
    console.log(shadowH1);
    shadowDiv = await page.waitForSelector("div[id=container] >>> h2", {
        timeout: 10000,
        visible: true,
    });
    const shadowH2 = await shadowDiv.$eval(">>> h2", (h2) => h2.innerText);
    
    console.log(shadowH2);
    await browser.close();
})();

test.html:
<html data-lt-installed="true">
    <head> </head>

    <body>
        <div id="container"></div>
        <h1>DOM element</h1>
        <script>
            (function () {
                var root = container.attachShadow({ mode: "open" });

                //Inside element
                var h1 = document.createElement("h1");
                h1.textContent = "Inside Shadow DOM";
                h1.id = "inside";
                root.appendChild(h1);

                function appendInterval() {
                    var h2 = document.createElement("h2");
                    h2.textContent = "Inside Shadow DOM setinterval";
                    h2.id = "inside";
                    root.appendChild(h2);
                }

                setTimeout(appendInterval, 5000);
            })();
        </script>
    </body>
</html>
```

### Background

I have narrowed down the regression point to version [22.12.1](https://github.com/puppeteer/puppeteer/releases/tag/puppeteer-core-v22.12.1), more precisely commit ([56d1d3f](https://github.com/puppeteer/puppeteer/commit/56d1d3f8b731d18c6aa9cc3d6de9c722b93a7a1e)).

Before changes to polling methods, this scenario used to work as expected.

### Expectation

After element is added with setTimeout function, puppeteer shall be able to find it using waitForSelector method.

### Reality

After shadow element is dynamically added, it is not possible to retreive it.

### Puppeteer version

23.5.0

### Node version

20.17

### Package manager

yarn

### Package manager version

3.5.1

### Operating system

Linux

filipxperi:
This still does not work on 23.5.1 

[...]

PR #13153, opened 2024-10-03
fix: default to RAF polling if visible||hidden is set

OrKoN:
The regression is caused by the visible flag no longer changing the polling type to RAF. This PR solves the case reported in the related bug when changes within shadow roots are tracked with visible=true.

The underlying and more general issue is that MutationPoller does not consider shadow roots at all. There is a related [spec issue](https://github.com/whatwg/dom/issues/1287) to support shadow roots in MutationObserver used by MutationPoller but as long as it is not supported we need a client side solution.

This PR adds a test for the shadow root mutation case as well as a test when visibility is changed without DOM mutations.
Existing tests using visible and hidden flags all seem to also cause a DOM mutation.

Refs: #13152

### DIFF
--- a/test/src/waittask.spec.ts
+++ b/test/src/waittask.spec.ts
@@ -368,7 +368,7 @@ describe('waittask specs', function () {
 
       await page.evaluate(() => {
         // @ts-expect-error We want to remove it for the test.
-        return delete window.MutationObserver;
+        delete window.MutationObserver;
       });
       const [handle] = await Promise.all([
         page.waitForSelector('.zombo'),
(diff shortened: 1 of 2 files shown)


test/src/waittask.spec.ts
```diff
@@ -407,6 +407,31 @@ describe('waittask specs', function () {
       await watchdog;
     });
 
+    // MutationPoller currently does not support shadow DOM.
```
Lightning00Blade, line 410: Let's add the spec issue here as well of safe keeping.

Lightning00Blade (approved):

test/src/waittask.spec.ts
OrKoN: I think it is not directly related to MutationPoller since it is our own abstraction and we can probably make it work without MutationObserver supporting shadow DOM out of the box.

OrKoN, line 410:
```suggestion
    // MutationPoller currently does not support shadow DOM.
    // See https://github.com/puppeteer/puppeteer/issues/13163.
```

PR #13161, opened 2024-10-06
docs: incorrect field used for browser selection

yanjiew1:
The browser should be specified in the 'browser' field, but the example mistakenly placed it in the 'product' field.

This commit fixes the example in the file `docs/webdriver-bidi.md`.
The browser selection should be put in the 'browser' field.

No.

Yes and the commit only updates the documentation.

**Summary**

I followed the example to launch Firefox. However, I found that Chrome is always launched.
After digging into the source code of puppeteer, I found that the browser selection should be put in the 'browser' field.

The original documentation may be confusing, so I made this change.

No.

**Other information**


OrKoN (approved):
Thanks!

PR #13165, opened 2024-10-08
chore: update types for TypeScript 5.6.2

Lightning00Blade:
This is prep work for the type as TypeScript now better detects a lot of issues.
The only change that need fixing so we can update the TS version is the ElementHandle decorators

### DIFF
--- a/packages/puppeteer-core/src/common/util.ts
+++ b/packages/puppeteer-core/src/common/util.ts
@@ -313,7 +313,8 @@ export const UTILITY_WORLD_NAME =
 /**
  * @internal
  */
-export const SOURCE_URL_REGEX = /^[\x20\t]*\/\/[@#] sourceURL=\s{0,10}(\S*?)\s{0,10}$/m;
+export const SOURCE_URL_REGEX =
+  /^[\x20\t]*\/\/[@#] sourceURL=\s{0,10}(\S*?)\s{0,10}$/m;
 /**
  * @internal
  */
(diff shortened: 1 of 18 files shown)


OrKoN (approved):

packages/puppeteer-core/src/common/util.ts
```diff
@@ -313,7 +313,7 @@ export const UTILITY_WORLD_NAME =
 /**
  * @internal
  */
-export const SOURCE_URL_REGEX = /^[\040\t]*\/\/[@#] sourceURL=\s*(\S*?)\s*$/m;
+export const SOURCE_URL_REGEX = /^[\x20\t]*\/\/[@#] sourceURL=\s?(\S*?)\s?$/m;
```
OrKoN: that is not equivalent, right? it only allows zero or 1 whitespace. I wonder if it could regress  on some inputs.

Lightning00Blade: If you follow where we use it it's only relevant to check if we added a source URL already from our end.
And we didn't attach one if the user provided a string function (I think this is a low usecase in Puppeteer) and attached a SourceUrl themself also very unlikely. But the linter was complaining that if you add a lot of spaces the  RegEx can become very slow in the seconds - I tested it and it's correct.

We can change it instead of `?` to `{0,5}` or other sensible default.

OrKoN: I think using string functions might actually be somewhat common. Let's check for up to 10 whitespaces?

OrKoN (approved):

PR #13168, opened 2024-10-09
chore: update decorator for TS update

Lightning00Blade:
There are still some issues with Api-extractor that I need to figure out before I can update the TS version.

### DIFF
--- a/packages/puppeteer-core/src/api/ElementHandle.ts
+++ b/packages/puppeteer-core/src/api/ElementHandle.ts
@@ -110,6 +110,65 @@ export interface ElementScreenshotOptions extends ScreenshotOptions {
 }
 
 /**
+ * A given method will have it's `this` replaced with an isolated version of
+ * `this` when decorated with this decorator.
+ *
+ * All changes of isolated `this` are reflected on the actual `this`.
+ *
+ * @internal
+ */
+export function bindIsolatedHandle<This extends ElementHandle<Node>>(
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -109,6 +109,65 @@ export interface ElementScreenshotOptions extends ScreenshotOptions {
...
 
+/**
+ * A given method will have it's `this` replaced with an isolated version of
+ * `this` when decorated with this decorator.
+ *
+ * All changes of isolated `this` are reflected on the actual `this`.
+ *
+ * @internal
+ */
+export function bindIsolatedHandle<This extends ElementHandle<Node>>(
```
OrKoN, line 120: since we are moving it, can we move it to the decorators.ts if that works?

Lightning00Blade: It should work it's a function and the types should allow. Let me try.

OrKoN (approved):

packages/puppeteer-core/src/api/ElementHandle.ts
Lightning00Blade: We can't move it due to the use of `instance of JsHandle` as it created a cycle dependencies.

OrKoN (approved):

PR #13203, opened 2024-10-16
chore(dependecies): update EsLint to v9

Lightning00Blade:
Let's review carefully the old config is still in the PR for comparison (`old-eslintrc.js`).  
Split in separate commits so it should easier to digest.


OrKoN (approved):

OrKoN (approved):
looks good I think. Let's land this.

OrKoN:
@Lightning00Blade let's add the stylistic rules (but in a separate CL).

PR #13216, opened 2024-10-22
ci: simplify test suites format

OrKoN:
- removes platforms that are outdated and not actually used (ci.yml determines if a test suite is run).
- also removes expectedLineCoverage which is mostly not used and does not make much sense with how our codebase is structured.

### DIFF
--- a/tools/mocha-runner/src/mocha-runner.ts
+++ b/tools/mocha-runner/src/mocha-runner.ts
@@ -20,7 +20,6 @@ import {
   zPlatform,
   zTestSuiteFile,
   type MochaResults,
-  type Platform,
   type TestExpectation,
   type TestSuite,
   type TestSuiteFile,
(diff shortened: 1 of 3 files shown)


tools/mocha-runner/src/mocha-runner.ts
```diff
@@ -238,13 +228,7 @@ async function main() {
...
         [
           ...(useCoverage
-            ? [
-                'c8',
-                '--check-coverage',
-                '--lines',
-                String(suite.expectedLineCoverage),
-                'npx',
-              ]
+            ? ['c8', '--check-coverage', '--lines', '90', 'npx']
```
Lightning00Blade, line 231: How about we remove this all together, it's not a meaningful representation considering we have BiDi and CDP, implementation. We need to merge all the runs and the gather the coverage (Firefox vs  Chrome Launcher etc.).

OrKoN: We can consider it later, for now, it is very similar behavior.

Lightning00Blade (approved):

Issue #13180, opened 2024-10-11
[acceptInsecureCerts.spec] acceptInsecureCerts "before each" hook for "should work" break often


OrKoN:
Example https://github.com/puppeteer/puppeteer/actions/runs/11290813220/job/31403444041?pr=13179

  1) acceptInsecureCerts
       "before each" hook for "should work":
     ProtocolError: Protocol error (Target.createBrowserContext): Session with given id not found.
      at Callback.<instance_members_initializer> (packages/puppeteer-core/src/common/CallbackRegistry.ts:114:12)
      at new Callback (packages/puppeteer-core/src/common/CallbackRegistry.ts:119:3)
      at CallbackRegistry.create (packages/puppeteer-core/src/common/CallbackRegistry.ts:27:22)
      at Connection._rawSend (packages/puppeteer-core/src/cdp/Connection.ts:125:22)
      at Connection.send (packages/puppeteer-core/src/cdp/Connection.ts:109:17)
      at CdpBrowser.createBrowserContext (packages/puppeteer-core/src/cdp/Browser.ts:207:55)
      at Context.<anonymous> (test/src/acceptInsecureCerts.spec.ts:35:41)
      at processImmediate (node:internal/timers:483:21)

  2) acceptInsecureCerts
       "after each" hook for "should work":
     ProtocolError: Protocol error (Target.disposeBrowserContext): Failed to find context with id 0BD16D6BBB13C3A95D3E2DE0E24102D5
      at Callback.<instance_members_initializer> (packages/puppeteer-core/src/common/CallbackRegistry.ts:114:12)
      at new Callback (packages/puppeteer-core/src/common/CallbackRegistry.ts:119:3)
      at CallbackRegistry.create (packages/puppeteer-core/src/common/CallbackRegistry.ts:27:22)
      at Connection._rawSend (packages/puppeteer-core/src/cdp/Connection.ts:125:22)
      at Connection.send (packages/puppeteer-core/src/cdp/Connection.ts:109:17)
      at CdpBrowser._disposeContext (packages/puppeteer-core/src/cdp/Browser.ts:235:28)
      at CdpBrowserContext.close (packages/puppeteer-core/src/cdp/BrowserContext.ts:99:25)
      at Context.<anonymous> (test/src/acceptInsecureCerts.spec.ts:40:25)
      at processImmediate (node:internal/timers:483:21)

Lightning00Blade:
[...]

PR #13219, opened 2024-10-22
chore: fix the hook issue

Lightning00Blade:
Let's try and see if this improves the problem with the hooks - https://github.com/puppeteer/puppeteer/issues/13180


test/src/acceptInsecureCerts.spec.ts
```diff
@@ -41,7 +42,7 @@ describe('acceptInsecureCerts', function () {
   });
 
   describe('Response.securityDetails', function () {
-    it('should work', async () => {
+    it.only('should work', async () => {
```
OrKoN: remove .only?

OrKoN (approved):

PR #13221, opened 2024-10-22
chore(dependencies): update to TS 5.6

Lightning00Blade:
So I looked more into it and it seems that our dependencies were in the wrong place resulting in the errors.
Moving them to the top level as it should be resolves the issue.
Update the relative packages that need 5.6.


OrKoN:
Is it possible to split the TS upgrade and formatting changes?

OrKoN (approved):

Issue #12929, opened 2024-08-09
[Bug]: ProtocolError: NS_BINDING_ABORTED on navigation in Firefox when navigating to https://x.com


OIRNOIR:
### Minimal, reproducible example

```TypeScript
import puppeteer from "puppeteer";

const browser = await puppeteer.launch({
	browser: "firefox"
});

const page = await browser.newPage();

await page.goto("https://x.com/"); // Crash!
```

### Background

I've been trying to access x.com with puppeteer through the Firefox browser. For some reason, the Firefox version of puppeteer cannot get past some step of the navigation process unique to x.com and crashes.

### Expectation

I expected the redirects to be followed properly and the page to load accordingly.

### Reality

This error occurs on page.goto:

```
file:///.../Puppeteer%20Firefox%20Testing/node_modules/.pnpm/puppeteer-core@23.0.2/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:89
    #error = new ProtocolError();
             ^

ProtocolError: Protocol error (browsingContext.navigate): unknown error Error: NS_BINDING_ABORTED #checkLoadingState@chrome://remote/content/shared/Navigate.sys.mjs:286:28
onStateChange@chrome://remote/content/shared/Navigate.sys.mjs:345:28
 at https://x.com/
    at <instance_members_initializer> (file:///.../Puppeteer%20Firefox%20Testing/node_modules/.pnpm/puppeteer-core@23.0.2/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:89:14)
    at new Callback (file:///.../Puppeteer%20Firefox%20Testing/node_modules/.pnpm/puppeteer-core@23.0.2/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:93:16)
    at CallbackRegistry.create (file:///.../Puppeteer%20Firefox%20Testing/node_modules/.pnpm/puppeteer-core@23.0.2/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:19:26)
    at BidiConnection.send (file:///.../Puppeteer%20Firefox%20Testing/node_modules/.pnpm/puppeteer-core@23.0.2/node_modules/puppeteer-core/lib/esm/puppeteer/bidi/Connection.js:51:32)
[...]

PR #13222, opened 2024-10-22
fix(webdriver): partially handle client-side redirects in page.goto

OrKoN:
This CL partially solves the issue in that the client-side redirect does not cause an error any more but there is still a difference between BiDi and CDP in that the navigation request returned is the first one in BiDi and the second one in CDP.

See https://github.com/puppeteer/puppeteer/issues/12929

### DIFF
--- a/packages/puppeteer-core/src/bidi/Frame.ts
+++ b/packages/puppeteer-core/src/bidi/Frame.ts
@@ -329,6 +329,10 @@ export class BidiFrame extends Frame {
             return;
           }
 
+          if (error.message.includes('navigation canceled')) {
+            return;
+          }
+
           throw error;
         }),
     ]).catch(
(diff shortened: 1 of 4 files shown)


OrKoN:
@juliandescottes is it possible to distinguish an error due to a redirect from the error in the test `"[navigation.spec] navigation Page.goto should fail when server returns 204"`. Both scenarios throw NS_BINDING_ABORTED but in one case we want to suppress it and in another one to throw.

juliandescottes:
@OrKoN Not at the moment.

I'm trying to improve that in https://bugzilla.mozilla.org/show_bug.cgi?id=1914407 but at the moment everything fails with NS_BINDING_ABORTED. 

We have slightly more fine grained information in our logic to compute navigationFailed events, but we don't expose the error message there.

Lightning00Blade (approved):

OrKoN:
Related https://github.com/w3c/webdriver-bidi/issues/112

PR #13223, opened 2024-10-22
chore(dependencies): update GTS to v6

Lightning00Blade:
Changes split in commits.


Herebyfile.mjs
```diff
@@ -86,7 +86,7 @@ export const docsBrowserSupportTask = task({
     }
     await writeFile(
       'docs/supported-browsers.md',
-      spliceIntoSection('version', content, buffer.join('\n'))
+      spliceIntoSection('version', content, buffer.join('\n')),
```
OrKoN, line 89: could we turn off the trailing comma? It results in too many changes so I would rather avoid it since this is not the style we have been using till now

Lightning00Blade: This matches Google's prettier config, I would prefer to follow that than invent our own style.

OrKoN: I do not find anything in https://google.github.io/styleguide/tsguide.html

OrKoN (approved):

PR #13226, opened 2024-10-23
chore: limit RegEx-ed strings

Lightning00Blade:


packages/puppeteer-core/src/cdp/AriaQueryHandler.ts
```diff
@@ -35,6 +35,10 @@ const isKnownAttribute = (
 const ATTRIBUTE_REGEXP =
   /\[\s*(?<attribute>\w+)\s*=\s*(?<quote>"|')(?<value>\\.|.*?(?=\k<quote>))\k<quote>\s*\]/g;
 const parseARIASelector = (selector: string): ARIASelector => {
+  if (selector.length > 1_000) {
```
OrKoN: could we go with 1k here?

OrKoN: s/1k/10k

Lightning00Blade: Let me try, though I think 1K is reasonable as well.

OrKoN (approved):

PR #13230, opened 2024-10-23
fix(docker): bump uid for pptruser

### LINKED ISSUE #13229: [Feature]: Update the User ID of Puppeteer image
### Feature description

Image: `ghcr.io/puppeteer/puppeteer:23.6.0`

Is there a reason the ID of the Puppeteer user is `999`?
Can we change it to 10001?

We got this warning: 
`Containers should run as a high UID to avoid host conflict.`

OrKoN:
Closes #13229

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -11,7 +11,7 @@ RUN apt-get update \
     fonts-kacst fonts-freefont-ttf dbus dbus-x11
 
 # Add pptruser.
-RUN groupadd -r pptruser && useradd -rm -g pptruser -G audio,video pptruser
+RUN groupadd -r pptruser && useradd -u 10042 -rm -g pptruser -G audio,video pptruser
 
 USER pptruser
 


Lightning00Blade (approved):

olegd-superoffice:
Would it be possible to save `UID` in a variable, so it can be used in `USER` command?
The same way as they do it in [.Net Dockerfile](https://github.com/dotnet/dotnet-docker/blob/0f8eee07e5c6fcc7243e10459a84b4b8090e506c/src/runtime-deps/8.0/noble-chiseled/amd64/Dockerfile#L45), they save it in `APP_UID` variable ?
We are getting the same error as described in [this stackowerflow question](https://stackoverflow.com/q/49720308/470325) and the obvious way to fix it is to hard-code UID in our deployment, but if we'd have in in a variable it would be much easier and won't break if UID is changed again.

Lightning00Blade:
@olegd-superoffice Please open a bug to track this. Also happy to review a PR that implements this, and maybe even allow user to control it from the outside.

PR #13231, opened 2024-10-23
feat: distinguish different Touches

### LINKED ISSUE #13084: [Feature]: Distinguish between different `Touch`es

emilefokkema:

This PR adds the option of controlling a Touch, and only that Touch, after it has been created by a call to [`TouchScreen.touchStart`](https://github.com/puppeteer/puppeteer/blob/6174eb0453ee022370ce5178dc71aa83e7a30435/packages/puppeteer-core/src/api/Input.ts#L498).

Closes #13084 .

Yes.

**Summary**

In this PR, `TouchScreen.touchStart` returns a (promise of) a [`Touch`](https://github.com/emilefokkema/puppeteer/blob/4e4af5d7efc78ce46ddca78ad5f7502797008c99/packages/puppeteer-core/src/api/Input.ts#L476). This way, touches can be manipulated individually by calling [`Touch.move`](https://github.com/emilefokkema/puppeteer/blob/4e4af5d7efc78ce46ddca78ad5f7502797008c99/packages/puppeteer-core/src/api/Input.ts#L482) or [`Touch.end`](https://github.com/emilefokkema/puppeteer/blob/4e4af5d7efc78ce46ddca78ad5f7502797008c99/packages/puppeteer-core/src/api/Input.ts#L486).

Well... [This test](https://github.com/emilefokkema/puppeteer/commit/45469a703ae3bee7788c33e534b57ce6b2e734d3) documents an aspect of the current behavior of `TouchScreen.touchStart`: two consecutive calls to it will not result in two `touchstart` events. This PR proceeds from the assumption that that behavior was in fact a bug: two consecutive calls to `TouchScreen.touchStart` _should_ result in two `touchstart` events. Seen in this way, this PR fixes a bug (in addition to adding new functionality). But if the current behavior of `TouchScreen.touchStart` was _not_ a bug, then this PR introduces a breaking change.

**Other information**


emilefokkema:
I'm not sure how to handle the Firefox failures. Should there be a 'failure expectation' for the ElementHandle tests having to do with touch, similar to [this one](https://github.com/puppeteer/puppeteer/blob/6174eb0453ee022370ce5178dc71aa83e7a30435/test/TestExpectations.json#L802)?

packages/puppeteer-core/src/common/TouchPointIdRepository.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @license
+ * Copyright 2020 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+interface IdRecord {
+  id: number;
+  inUse: boolean;
+}
+export class TouchPointIdRepository {
```
Lightning00Blade: Why do we need this and can't we use `createIncrementalIdGenerator` we already have? 
From both BiDi and CDP it seems this is only need to keep track of the pointer event group.

packages/puppeteer-core/src/bidi/Input.ts
```diff
@@ -699,3 +705,47 @@ export class BidiTouchscreen extends Touchscreen {
...
+    const touch = new BidiTouch(this.#page, id, x, y, properties);
+    await touch.start();
+    this.#touches.push(touch);
+    return touch;
+  }
+
+  override touchMove(x: number, y: number): Promise<void> {
+    const touch = this.#touches[0];
+    if (!touch) {
+      return Promise.resolve();
```
Lightning00Blade: Nit: No need for `Promise.resolve` if you use async await, that is the preferred syntax we have.

test/TestExpectations.json
```diff
@@ -1323,6 +1323,55 @@
     "expectations": ["FAIL"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
+  {
+    "testIdPattern": "[elementhandle.spec] ElementHandle specs ElementHandle.touchEnd should work",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["cdp", "firefox"],
+    "expectations": ["FAIL"],
+    "comment": "Same reason as '[touchscreen.spec] *'"
```
Lightning00Blade: It would be better to copy paste them rather then referencing as we my remove the other at some point.

packages/puppeteer-core/src/common/TouchPointIdRepository.ts
```diff
@@ -0,0 +1,40 @@
...
+  #ids: IdRecord[] = [];
+
+  private generateNewId(): number {
```
Lightning00Blade: We use the JavaScript native private fields prefixed with `#`.

packages/puppeteer-core/src/common/TouchPointIdRepository.ts
```diff
@@ -0,0 +1,40 @@
...
+    return this.#ids.length;
+  }
+
+  public getId(): number {
```
Lightning00Blade: Public is the default so it must be omitted.

packages/puppeteer-core/src/api/Input.ts
```diff
@@ -469,7 +469,22 @@ export abstract class Mouse {
     options?: {delay?: number},
   ): Promise<void>;
 }
-
+/**
+ * The Touch interface exposes methods to manipulate touches that have been started
+ * @public
+ */
+export interface Touch {
```
Lightning00Blade: I think we may need to change the name as it clashes with the DOM Touch interface, that is why the `.md` file reference `Touch_2` . I will try to think of something, open to suggestions.

packages/puppeteer-core/src/common/TouchPointIdRepository.ts
emilefokkema: This is to adhere to the way pointer ids are usually chosen: they are re-used. So the first gets id 0, the second gets id 1, and then the first ends, and then the third gets id 0, because 0 is available. See [this](https://w3c.github.io/webdriver/#dfn-get-a-pointer-id) also. Unfortunately I don't have a test that proves that this is necessary, but I believe it is.

Lightning00Blade: Let me dig in the spec what this exactly does, I think for CDP it doesn't matter but for BiDi it may.

test/TestExpectations.json
emilefokkema: I opted for this because the reason in the other one is "TODO add a comment (etc.)", but I'll copy-paste that :)

packages/puppeteer-core/src/api/Input.ts
emilefokkema: Excellent point, this collision is annoying

packages/puppeteer-core/src/common/TouchPointIdRepository.ts
emilefokkema: Let me know :) Would be nice if we could be sure it wasn't necessary

packages/puppeteer-core/src/api/Input.ts
emilefokkema: TouchPoint? TouchPointer? Pointer?

test/TestExpectations.json
Lightning00Blade: Yeah, we added the field well into this file so not all is filled but its good to know that its not been investigated why it failed.

packages/puppeteer-core/src/api/Input.ts
OrKoN: How about TouchHandle?

packages/puppeteer-core/src/bidi/Input.ts
```diff
@@ -613,79 +615,83 @@ export class BidiMouse extends Mouse {
...
+    public id: number,
+    x: number,
+    y: number,
+    properties: Bidi.Input.PointerCommonProperties,
+  ) {
     this.#page = page;
+    this.#x = Math.round(x);
+    this.#y = Math.round(y);
+    this.#properties = properties;
+    this.#bidiId = `__puppeteer_finger${id}`;
```
Lightning00Blade: Let use  `${InputId.Finger}_{id}`.

packages/puppeteer-core/src/cdp/Input.ts
```diff
@@ -552,57 +554,110 @@ export class CdpMouse extends Mouse {
...
+    const touch = new CdpTouch(this.#client, this.#keyboard, id, touchPoint);
+    await touch.start();
+    this.#touches.push(touch);
+    return touch;
+  }
+
+  override touchMove(x: number, y: number): Promise<void> {
+    const touch = this.#touches[0];
+    if (!touch) {
+      return Promise.resolve();
```
Lightning00Blade: @OrKoN I think what I am about to suggest may be breaking, but if no touch is started shouldn't we throw an error?

packages/puppeteer-core/src/api/Input.ts
emilefokkema: I like it.

packages/puppeteer-core/src/bidi/Input.ts
emilefokkema: done

packages/puppeteer-core/src/cdp/Input.ts
emilefokkema: The opposite, actually! Turns out that we currently already throw in this case, as evidenced by [this test](https://github.com/emilefokkema/puppeteer/blob/4ca2a144f5b8f2aa0476ef8ece9348d180c6ca99/test/src/touchscreen.spec.ts#L249). I just changed this so that we still throw in this case.

test/TestExpectations.json
```diff
@@ -3439,6 +3509,12 @@
     "expectations": ["SKIP"],
     "comment": "For some reason service_workers do not close in chrome-headless-shell"
   },
+  {
+    "testIdPattern": "[touchscreen.spec] Touchscreen Touchscreen.prototype.touchMove should throw if no touch was started",
+    "platforms": ["linux"],
```
OrKoN, line 3521:
```suggestion
    "platforms": ["darwin", "linux", "win32"],
```

I assume it applies to all platforms

emilefokkema: actually only that one it seems, judging from the CI run

OrKoN: CI only runs on Linux by default. I can add the full ci label but it would take longer to run.

packages/puppeteer-core/src/api/Input.ts
emilefokkema: I'll change it to TouchHandle then, hang on

emilefokkema:
@OrKoN  One more thing! If you now start a touch through `TouchScreen.touchStart` and then you end it through `TouchHandle.end`, then the `TouchScreen` still keeps the handle in its array of touches. Let me fix that.

emilefokkema:
> @OrKoN One more thing! If you now start a touch through `TouchScreen.touchStart` and then you end it through `TouchHandle.end`, then the `TouchScreen` still keeps the handle in its array of touches. Let me fix that.

I fixed this by making sure that the handle is removed from the TouchScreen instance and by throwing an error from `TouchScreen.touchEnd` in case there is no touch to end. This is compatible with the current behavior: we already throw the ProtocolError in this case.

test/TestExpectations.json
emilefokkema: It seems like full ci passed, right? Is this change needed?

OrKoN: interesting, that is only passed on linux but if that is the case then this change is not needed.

packages/puppeteer-core/src/common/TouchPointIdRepository.ts
emilefokkema: So do we keep this?

Lightning00Blade: Let's simplify this with the `createIncrementalIdGenerator` (and extract this in utils). 
The spec is a little bit ambiguous of what exactly it is. 
https://developer.mozilla.org/en-US/docs/Web/API/Touch/identifier - This looks like the only API that let's you distinguish between the touches and they are unique per start/mode/end/cancel

emilefokkema: on it :)

packages/puppeteer-core/src/common/TouchError.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * @license
+ * Copyright 2019 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export class TouchError extends Error {}
```
Lightning00Blade: This should live in `Errors.ts` and extend PuppeteerError for user facing errors.

emilefokkema: done

emilefokkema:
Since the Touchscreen functionality in cdp/Input and bidi/Input were so similar, I decided to extract some functionality to a new class Touchscreen in common.

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -997,19 +998,36 @@ export abstract class ElementHandle<
...
     await this.scrollIntoViewIfNeeded();
     const {x, y} = await this.clickablePoint();
-    await this.frame.page().touchscreen.touchStart(x, y);
+    return await this.frame.page().touchscreen.touchStart(x, y);
   }
 
+  /**
+   * This method scrolls the element into view if needed, and then
+   * moves the touch to the center of the element.
+   * @param touch - An optional TouchHandle. If provided, this touch
```
Lightning00Blade, line 1017:
```suggestion
   * @param touch - An optional {@link TouchHandle}. If provided, this touch
```

packages/puppeteer-core/src/common/Touchscreen.ts
```diff
@@ -0,0 +1,56 @@
+/**
```
Lightning00Blade: This should live in API. The structure is a little bit confusing :sweat_smile:

emilefokkema: But is has a generic type parameter, and a public method 'removeHandle'. Does that really belong in api?

packages/puppeteer-core/src/cdp/Input.ts
```diff
@@ -552,57 +553,103 @@ export class CdpMouse extends Mouse {
...
   }
 
   updateClient(client: CDPSession): void {
     this.#client = client;
   }
 
-  override async touchStart(x: number, y: number): Promise<void> {
+  async start(): Promise<void> {
+    if (this.#started) {
+      return;
```
Lightning00Blade: Let's throw an error here, this should no be reachable, and it better to fix issues if we encounter them then fail them silently.

emilefokkema: We could throw an error, but how do I test that? `start()` is not part of the public api after all

packages/puppeteer-core/src/common/Touchscreen.ts
emilefokkema: I could also just revert it if you don't like it

Lightning00Blade: We should move it directly to the abstract class as that it's purpose to hold common logic.

We can remove the generic and overwrite it if needed on the child class with `declare X: <type>`.

As for the public method we can omit it from the types and docs by marking it as internal.
```ts
/*
* @internal
*/
method(){}
```

emilefokkema: I don't understand the part about removing the generic. How would you do that?

packages/puppeteer-core/src/cdp/Input.ts
emilefokkema: I added the error.

packages/puppeteer-core/src/bidi/Input.ts
```diff
@@ -613,79 +615,86 @@ export class BidiMouse extends Mouse {
...
+        id: this.#bidiId,
         parameters: {
           pointerType: Bidi.Input.PointerType.Touch,
         },
         actions: [
           {
             type: ActionType.PointerMove,
-            x: Math.round(x),
-            y: Math.round(y),
-            origin: options.origin,
```
Lightning00Blade, line 641: I think we need to keep the origin.

packages/puppeteer-core/src/common/Touchscreen.ts
emilefokkema: Also I thought that because it is API it should contain as little logic as possible

packages/puppeteer-core/src/bidi/Input.ts
```diff
@@ -613,79 +615,86 @@ export class BidiMouse extends Mouse {
...
         parameters: {
           pointerType: Bidi.Input.PointerType.Touch,
         },
         actions: [
           {
+            ...this.#properties,
             type: ActionType.PointerMove,
-            x: Math.round(x),
-            y: Math.round(y),
-            origin: options.origin,
```
Lightning00Blade, line 673: Same here.

packages/puppeteer-core/src/bidi/Input.ts
emilefokkema: Yes that is something I didn't understand. [This parameter](https://github.com/puppeteer/puppeteer/blob/b60067735da000200b1b151a0cfb5ace5bf75738/packages/puppeteer-core/src/bidi/Input.ts#L659) and [this one](https://github.com/puppeteer/puppeteer/blob/b60067735da000200b1b151a0cfb5ace5bf75738/packages/puppeteer-core/src/bidi/Input.ts#L627) are not part of the public api, and they're not tested at all. Only the Bidi touchscreen has them, and not the Cdp one. How are they ever used?

packages/puppeteer-core/src/common/Touchscreen.ts
emilefokkema: Not worth it. Reverted :)

packages/puppeteer-core/src/bidi/Input.ts
emilefokkema: I put the `BidiTouchMoveOptions` parameters back, but they're still untested.

packages/puppeteer-core/src/bidi/Input.ts
emilefokkema: See resolution above.

emilefokkema:
@Lightning00Blade [this test](https://github.com/emilefokkema/puppeteer/blob/33dba32e8a4ff2d250c33a406e1f49599415710e/test/src/touchscreen.spec.ts#L728) seems to be failing. Could this be due to the way we choose pointer ids (incemental instead of re-using them)?

emilefokkema:
> @Lightning00Blade [this test](https://github.com/emilefokkema/puppeteer/blob/33dba32e8a4ff2d250c33a406e1f49599415710e/test/src/touchscreen.spec.ts#L728) seems to be failing. Could this be due to the way we choose pointer ids (incemental instead of re-using them)?

Especially since [it did pass earlier](https://github.com/puppeteer/puppeteer/actions/runs/11515164664/job/32055558244), before we changed it.

emilefokkema:
> @Lightning00Blade [this test](https://github.com/emilefokkema/puppeteer/blob/33dba32e8a4ff2d250c33a406e1f49599415710e/test/src/touchscreen.spec.ts#L728) seems to be failing. Could this be due to the way we choose pointer ids (incemental instead of re-using them)?

Never mind, it seems to have to do with the location of the touches. Should I just add a failure expectation?

emilefokkema:
I suppose many of those 'failure expectations' could be removed if we replaced [assertions such as this](https://github.com/puppeteer/puppeteer/blob/b60067735da000200b1b151a0cfb5ace5bf75738/test/src/touchscreen.spec.ts#L26) with snapshots: a snapshot for each browser/platform. But that's for another pr :)

packages/puppeteer-core/src/common/Touchscreen.ts
Lightning00Blade: Just a note about structure we have decided to use - we use API as common between the two protocols we support, all common implementation goes there.
That is why we use  `abstract class`, if it were meant to be only Documentation and Types the correct thing to use are `interfaces` and makes sure the implementation `implements` them. 

For the `declare` it's TypeScript feature that seems to be not that known details: 
You seem to need the generic for the CDP implementation to update the client .
You can take a look at `packages/puppeteer-core/src/cdp/HTTPRequest.ts` 
On the base class in `api` we say 
```ts
class HTTPRequest{
  _redirectChain: HTTPRequest[]
}
```
on the child class we say that 
```ts
class CdpHTTPRequest extends HTTPRequest {
  declare _redirectChain: CdpHTTPRequest[]
}
```
In short you can provide this allows you to scope down the type of property to more precise one.

emilefokkema: Interesting! I had never seen that.

About the structure: if api is for common functionality, then why is there a separate 'common' directory?

Lightning00Blade: Good question - I can see that is confusing - there are (mainly) historical reason why the structure exist: 
In general `api` is the extracted logic from the original that can be reused with the BiDi protocol,
while `common` are things that work that don't have special logic for depending on the protocol.

emilefokkema:
This failure does not seem to be related to the PR's changes, right?

test/src/touch-event-utils.ts
```diff
@@ -0,0 +1,46 @@
+/**
+ * @license
+ * Copyright 2017 Google Inc.
```
Lightning00Blade:
```suggestion
 * Copyright 2024 Google Inc.
```

packages/puppeteer-core/src/util/incremental-id-generator.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * @license
+ * Copyright 2023 Google Inc.
```
Lightning00Blade:
```suggestion
 * Copyright 2024 Google Inc.
```

packages/puppeteer-core/src/cdp/Input.ts
```diff
@@ -552,57 +555,127 @@ export class CdpMouse extends Mouse {
...
   #client: CDPSession;
   #keyboard: CdpKeyboard;
 
-  constructor(client: CDPSession, keyboard: CdpKeyboard) {
-    super();
+  constructor(
+    client: CDPSession,
+    touchScreen: CdpTouchscreen,
+    keyboard: CdpKeyboard,
+    public id: number,
```
Lightning00Blade:
```suggestion
    id: number,
```

Lightning00Blade (changes requested):
LGTM, 3 small comments left and we can merge.

packages/puppeteer-core/src/cdp/Input.ts
emilefokkema: not even the constructor parameter was needed ;)

Lightning00Blade (approved):

PR #13233, opened 2024-10-24
chore: extract extra browser creation in hook

Lightning00Blade:
Creates a wrapper function similar to the original setUpBrowser.
This remove the need to set it per test and shows that a couple of places were not reusing a browser.
Also changes the defaults `launch` to not create a Context and Page, as almost all test use `browser.newPage` so we were creating them for no reson.


OrKoN:
There are some test failures.

OrKoN (approved):

PR #13247, opened 2024-10-29
test: remove unnecessary browser spawn

Lightning00Blade:


packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -300,9 +300,8 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
    * You can find the debugger URL (`webSocketDebuggerUrl`) from
    * `http://HOST:PORT/json/version`.
    *
-   * See {@link
-   * https://chromedevtools.github.io/devtools-protocol/#how-do-i-access-the-browser-target
-   * | browser endpoint} for more information.
+   * See {@link https://chromedevtools.github.io/devtools-protocol/#how-do-i-access-the-browser-target | browser endpoint}
```
Lightning00Blade, line 303: Link was broken when viewed from VsCode.

test/src/device-request-prompt.spec.ts
```diff
@@ -6,34 +6,12 @@
...
-  });
-
-  beforeEach(async () => {
-    state.context = await state.browser.createBrowserContext();
-    state.page = await state.context.newPage();
-  });
-
-  afterEach(async () => {
-    await state.context.close();
+  const state = setupSeparateTestBrowserHooks({
```
Lightning00Blade, line 12: We can use the new hook here.

test/src/chromiumonly.spec.ts
```diff
@@ -44,45 +44,33 @@ describe('Chromium-Specific Launcher tests', function () {
...
+        'Exactly one of browserWSEndpoint, browserURL or transport',
+      );
     });
     it('should throw when trying to connect to non-existing browser', async () => {
-      const {close, puppeteer} = await launch({
-        args: ['--remote-debugging-port=21222'],
-      });
-      try {
-        const browserURL = 'http://127.0.0.1:32333';
+      const {puppeteer} = await getTestState({skipLaunch: true});
```
OrKoN, line 64: I suspect that test might be testing that the browser listens on the right port as well (although that is not reflected in the test name)

OrKoN (approved):

test/src/chromiumonly.spec.ts
Lightning00Blade: I think that in not a useful test (the one you are suggesting), considering that is a browser behavior it should live with the browser.
And also even if it test that it doesn't mean we can test every single port to verify that the browser listens no non expect the specified (that sounds out of scope for Puppeteer).

PR #13250, opened 2024-10-30
refactor: extract common code for TouchHandle

Lightning00Blade:


OrKoN (approved):

packages/puppeteer-core/src/bidi/Input.ts
```diff
@@ -721,57 +720,28 @@ class BidiTouch implements TouchHandle {
...
     await touch.start(options);
-    this.#touches.push(touch);
+    this.touches.push(touch);
     return touch;
   }
-
-  override async touchMove(
-    x: number,
-    y: number,
-    options: BidiTouchMoveOptions = {},
```
emilefokkema, line 761: but @Lightning00Blade I thought you wanted to keep the `options` parameter?

Lightning00Blade: Sorry for the confusion, I didn't understand its the reason for not extracting the logic. 
Considering that we don't test or expose it, I have opted-out of it.

emilefokkema: Thanks for the clarification. That was indeed the reason. So #13254 could get rid of it in the other place as well :)

PR #13257, opened 2024-11-01
fix: remove event listeners from AbortSignal in WaitTask

### LINKED ISSUE #13256: [Bug]: Abort signal in WaitTask causes memory leak.

```TypeScript

async function main(): Promise<void> {
  const browser = await launch({
    headless: true,
    defaultViewport: null,
    args: ['--window-size=1920,1080'],
  });
  await page.goto('http://localhost:5173/');
  const abortControler = new AbortController();
  for (let i = 0; i < 20; i++) {
    // im using Vite starter app, but actuallly any locator could be used.
    await page.locator('#root').wait({
      signal: abortControler.signal,
    });
  }
  await browser.close();
}

main().catch((e) => console.log('Error in main', e));
```

### Background

[code that adds listener](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/WaitTask.ts#L59)

[place that do not have clean-up](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/WaitTask.ts#L177)

### Expectation

I expect that Wait task removes event handler on terminate(similar to timeout)

### Reality

Handler is kept on Abort signal.

_No response_

### Puppeteer version

23.6.0

### Node version

20.3.1

### Package manager

npm

### Package manager version

9.6.7

### Operating system

Linux

OrKoN:
Closes https://github.com/puppeteer/puppeteer/issues/13256

### DIFF
--- a/packages/puppeteer-core/src/common/WaitTask.ts
+++ b/packages/puppeteer-core/src/common/WaitTask.ts
@@ -56,15 +56,9 @@ export class WaitTask<T = unknown> {
     this.#polling = options.polling;
     this.#root = options.root;
     this.#signal = options.signal;
-    this.#signal?.addEventListener(
-      'abort',
-      () => {
-        void this.terminate(this.#signal?.reason);
-      },
-      {
-        once: true,
-      },
-    );
+    this.#signal?.addEventListener('abort', this.#onAbortSignal, {
+      once: true,
+    });
 
     switch (typeof fn) {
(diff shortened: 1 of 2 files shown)


Lightning00Blade (approved):
This solves this issue with WaitTask, but I think other places that use AbortController may have the same issue if the AbortController is GC as well.

Issue #12921, opened 2024-08-08
Puppeteer v24


OrKoN:
No target date yet

- [x] remove Firefox CDP implementation
- [x] Include fragment into the Request/response url https://github.com/puppeteer/puppeteer/pull/13289#issuecomment-2468858796
- [x] Remove deprecated Launch and Connect Options types https://github.com/puppeteer/puppeteer/pull/13332

Can be done later:

- [ ] replace Page cookies API with Browser/BrowserContext cookies
- [ ] breaking change to support composite partition keys
- [ ] breaking changes for MV3 extensions
- [ ] use pipes by default (whenever possible)
- [ ] Rename Console Message `type` to `method` and map CDP to the correct one.
- [ ] Remove `unknown` for type [HttpRequest.ResponseForRequest](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/api/HTTPRequest.ts#L43) and relative code change.

cc @Lightning00Blade

cheekEprogrammer:
is there any way to update the node version it uses? i am having to skip using puppeteer now because it seems to only run on node v16 at the most. unless im wrong but it seems to be the issue for my project.

Kle0s:
Is there an issue for the cookies API change? I want this for testing another project I work on, so I will gladly participate in the implementation

OrKoN:
@Kle0s there is no issue but I have a draft PR that had no time to finish https://github.com/puppeteer/puppeteer/pull/12920

OrKoN:
actually, there is an old issue here https://github.com/puppeteer/puppeteer/issues/645 as well

OrKoN:
> @OrKoN I can't install this latest version on our node project.
> 
> Error log:
> 
> npm ERR! code ETARGET npm ERR! notarget No matching version found for puppeteer@24.0.0. npm ERR! notarget In most cases you or one of your dependencies are requesting npm ERR! notarget a package version that doesn't exist.

thanks for a ping, it looks like publishing to npm failed

PR #13289, opened 2024-11-11
feat: expose url fragment

Kikobeats:

**Summary**

**Other information**


OrKoN (changes requested):
thanks for the PR but we would not extend the url getter to include this. It's better to make a breaking change in v24 (https://github.com/puppeteer/puppeteer/issues/12921) and include the url fragment by default.

OrKoN:
although I wonder if @Lightning00Blade has a different opinion.

Issue #8010, opened 2022-02-14
Downloading PDF files using Puppeteer 


ags4436:
I am trying to download PDF files using puppeteer, when .pdf URLs are hit the chrome opens the pdf in the tab, instead I want the pdf to be downloaded. This can be achieved through changing the user-preferences of browser and  **Sites automatically follow this setting when you visit them** by default it is **Open PDFs in Chrome**.

I found that we can use puppeteer-extra for downloading the PDF files. Is that a legit NPM which I could use in my project instead of puppeteer-core ?

Is there any alternatives available from Puppeteer itself ?

OrKoN:
I believe it should be possible to use CDP's setDownloadBehaviour https://chromedevtools.github.io/devtools-protocol/tot/Browser/#method-setDownloadBehavior but there is no dedicated API in Puppeteer for this yet: 

```
await  page._client.send('Page.setDownloadBehavior', {
	behavior: 'allow',
	downloadPath: tempFolderPath,
})
```

ags4436:
This allows us to set the download path while downloading the files. PDFs will open in the tab itself in this case as well.

OrKoN:
I see. It looks like puppeteer-extra writes those settings into the user profile directory and to me it looks like the approach might be a bit fragile if the structure of the directory changes. But I am not aware of any other way to achieve this. Most likely, it's possible to replicate what puppeteer-extra does and create the profile with the right setting but I think using puppeteer-extra is fine too.

MrXyfir:
Really surprised puppeteer still doesn't support this. Playwright does.

nros:
Because of PR [8506](https://github.com/puppeteer/puppeteer/pull/8506)  (since release 14.4.0), it's not `_client` anymore but it is `_client()`. So use:
```JavaScript
await page._client().send('Page.setDownloadBehavior', {
    behavior: 'allow',
    downloadPath: tempFolderPath,
})
```

See: [Page.ts](../blob/main/src/common/Page.ts)

Korak-997:
I had to twick a lot till i found this solution :smile: 
but it works perfectly for me.

And Downloads the pdf automatically which means you do not need to do anything else

```js
// WAITS TILL THE OBJECT IS RENDERED
await page.waitForSelector('SELECTOR FOR PDF OBJECT')
   //SET DOWNLOAD PATH
    await page._client.send("Page.setDownloadBehavior", {
      behavior: "allow",
      downloadPath: PATH,
    })
[...]

PR #13309, opened 2024-11-19
feat: config download behavior

Kle0s:

Feature

Yes

Irrelevant (added inline documentation)

**Summary**

Added the ability to configure download behavior on launch / connect and on creation of new browser contexts (using https://pptr.dev/api/puppeteer.browser.createbrowsercontext ). 
More information is available on #8010.

No, expanded existing options interfaces in order to add this.

**Other information**


OrKoN:
Let's add tests including a test page. We have a test server in our tests so we should use it for the server-side logic.

test/src/coverage.spec.ts
```diff
@@ -339,4 +340,20 @@ describe('Coverage specs', function () {
       });
     });
   });
+  describe('download behavior', () => {
```
OrKoN: let's start a new spec file for this feature, e.g., `download.spec.ts`.

test/src/coverage.spec.ts
```diff
@@ -339,4 +340,20 @@ describe('Coverage specs', function () {
...
+    it('should download to configured location', async () => {
+      const {browser, server} = await getTestState();
+
+      const context = await browser.createBrowserContext({
+        downloadBehavior: {
+          policy: 'allow',
+          downloadPath: '/tmp',
```
OrKoN: let's use node's APIs to generate a tempdir name in a cross-platform way and make sure we clean up the file if the test fails.

test/src/utils.ts
```diff
@@ -177,3 +179,34 @@ export function getUniqueVideoFilePlaceholder(): FilePlaceholder {
 export function rmIfExists(file: string): Promise<void> {
   return rm(file).catch(() => {});
 }
+
+export function waitForFileExistence(
+  filePath: string,
+  timeout = 5000,
+): Promise<boolean> {
+  return new Promise((resolve, reject) => {
+    access(filePath, constants.R_OK, err => {
```
OrKoN: let's use the promises API https://nodejs.org/docs/latest/api/fs.html#promises-api for both access and watch.

test/src/coverage.spec.ts
Kle0s: done

test/src/coverage.spec.ts
Kle0s: fixed, now using `fs/promises`'s `mkdtemp` and removing it between tests.

test/src/utils.ts
Kle0s: Done

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -211,6 +219,11 @@ export class CdpBrowser extends BrowserBase {
         proxyBypassList: proxyBypassList && proxyBypassList.join(','),
       },
     );
+    await this.#connection.send('Browser.setDownloadBehavior', {
```
OrKoN: let's move this into a method on the CdpBrowserContext and let's call when we create a new context here and the second call let's place it in CdpBrowser._attach when the default context is created.

Kle0s: Okay, this is more or less a mix of what I suggested first with the launch/connect/createBrowserContext you suggested. Seems great, done :)

OrKoN: Right, we can have the method on the CdpBrowserContext but I would not make it part of the public API (i.e., api/BrowserContext).

OrKoN:
So it looks like tests do not pass yet (wrong way to create a temp dir?). Also, please run `npm run docs` to regenerate API documentation. For WebDriver BiDi, we would need to skip the tests by editing https://github.com/puppeteer/puppeteer/blob/main/test/TestExpectations.json (run `npm run format:expectations` after adding a new entry).

OrKoN:
And also let's skip the tests for "firefox" and "cdp" configurations.

Kle0s:
I think I may have fixed the temporary directory issue, but running `npm run docs` constantly fails as follows:
```
~/src/puppeteer ±(main) ✗ ❯ npm run docs

> docs
> wireit

SyntaxError: Unexpected identifier 'assert'                   
    at compileSourceTextModule (node:internal/modules/esm/utils:340:16)
    at ModuleLoader.moduleStrategy (node:internal/modules/esm/translators:102:18)
    at #translate (node:internal/modules/esm/loader:433:12)
    at ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:480:27)
    at async ModuleJob._link (node:internal/modules/esm/module_job:112:19)

❌ [docs] exited with exit code 1.
❌ 1 script failed.
```

OrKoN:
Are you using Node 20?

Kle0s:
I used node 22. I now switched to 20 and the issue was resolved. Maybe we should add it to the relevant part of the guide?

packages/puppeteer-core/src/common/DownloadBehavior.ts
```diff
@@ -0,0 +1,28 @@
+/**
+ * @license
+ * Copyright 2024 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export type DownloadPolicy = 'deny' | 'allow' | 'allowAndName' | 'default';
```
OrKoN, line 10: I think this one also needs a `@public` annotation and a re-export in packages/puppeteer-core/src/common/common.ts so that docs and types are fully generated.

Kle0s: Good catch! fixed it, exported in common.ts and ran `npm run docs` once again :)

OrKoN:
The tests still do not pass (do you run them locally?), some formatting needs fixing and expectations need to be split up for webDriverBiDi and cdp. I cannot push to your branch so please see changes needed here https://github.com/puppeteer/puppeteer/pull/13313/commits/2f63aa87710f29b55176f0c4e607ede22d2fddf4

Kle0s:
I ran the tests locally and they did pass. Regarding the formatting, I guess I am too used to having a pre-commit hook running `npm run lint` but I will work on that on my end.
When I ran it locally (`npm test`), I got the following (regarding my specific tests):
```
Download
    Browser.createBrowserContext
      ✔ should download to configured location (391ms)
      ✔ should not download to location (5170ms)
```

OrKoN:
The issue that I saw locally is that the watcher would never stop so the test runner remains pending forever after a test run.

OrKoN:
`npm run format` should fix all formatting issues automatically.

OrKoN (approved):
Thanks!

Kle0s:
Interesting. In the prior implementation I had (without using the `fs/promise` API), I resolved it by calling `watcher.close`. Since it was no longer possible, I thought it was internally resolved in the new API.

Kle0s:
> The issue that I saw locally is that the watcher would never stop so the test runner remains pending forever after a test run.

Interesting. In the prior implementation I had (without using the `fs/promise` API), I resolved it by calling `watcher.close`. Since it was no longer possible, I thought it was internally resolved in the new API.

Issue #13307, opened 2024-11-18
[Bug]: Error retrieving OOPIF response content


ezg27:
### Minimal, reproducible example

```TypeScript
import { createServer } from 'node:http'
import puppeteer from 'puppeteer'

// Create a server that serves an HTML page with an embedded cross-origin iframe
const server = createServer((req, res) => {
  res.writeHead(200, { 'Content-Type': 'text/html' })
  res.end(`
    <!DOCTYPE html>
    <html>
      <head>
        <title>Top-level app</title>
      </head>
      <body>
        <h1>Top-level app</h1>

        <iframe
          src="https://www.example.com"
          width="100%"
          height="500"
        ></iframe>
      </body>
    </html>
  `)
})

server.listen(3000, () => {
  console.log('Server running on http://localhost:3000')
})

// Crawl the page and attempt to get the response body of the cross-origin iframe
const browser = await puppeteer.launch()
const page = await browser.newPage()

page.on('response', async response => {
  if (response.url() === 'https://www.example.com/') {
    try {
      const body = await response.buffer()
      console.log('Response body:', body.toString())
    } catch (error) {
      // We'll end up here <<<<<<<
      console.error('Failed to get response body:', error)
    }
  }
})

await page.goto('http://localhost:3000', {
  waitUntil: 'networkidle0',
})

// Wait briefly in case https://www.example.com/ takes a while to load
await new Promise(resolve => setTimeout(resolve, 1000))

await page.close()
await browser.close()

server.close(() => {
  console.log('Server closed')
})
```

### Background

It looks as though the way Puppeteer currently tracks `CDPSession` clients internally for `HTTPRequest` and `HTTPResponse` instances leads to errors when attempting to retrieve response bodies for OOPIFs.

When `HTTPRequest` and `HTTPResponse` instances are created in `NetworkManager`, they’re currently initialised with the `CDPSession` on which the CDP event originated (this could be any one of `Fetch.requestPaused`, `Network.requestWillBeSent`, `Network.requestServedFromCache`, `Network.responseReceived`, `Network.responseReceivedExtraInfo`). These clients are then used for subsequent operations on requests/responses.

[...]

PR #13311, opened 2024-11-19
fix: correctly resolve OOPIF response bodies

ezg27:

Bugfix

Yes

**Summary**

Fix for https://github.com/puppeteer/puppeteer/issues/13307. I've raised it as a draft initially as there's a potential edge case that this may need to account for in `#maybeReassignOOPIFRequestClient` - namely, if we end up with different sessions for a document request on the main frame (top-level page). Is this something you're aware can happen? If so, what do you think would be the best way of making this info available for this check in `NetworkManager`, i.e. does the `CDPSession` that just received a `Network.loadingFinished` or `Network.loadingFailed` event belong to the 'main' target or not?

No

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/cdp/NetworkManager.ts
+++ b/packages/puppeteer-core/src/cdp/NetworkManager.ts
@@ -725,11 +725,11 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
     client: CDPSession,
     request: CdpHTTPRequest,
   ): void {
-    // Document requests for OOPIFs will in the parent frame but are adopted by their
-    // child frame, so loadingFinished and loadingFailed for these requests will be fired
-    // on the child session. In this case we reassign the request CDPSession to ensure all
-    // subsequent actions are called on the correct session (e.g. retrieving response body
-    // in HTTPResponse).
+    // Document requests for OOPIFs start in the parent frame but are adopted by their
+    // child frame, meaning their loadingFinished and loadingFailed events are fired on
+    // the child session. In this case we reassign the request CDPSession to ensure all
+    // subsequent actions use the correct session (e.g. retrieving response body in
+    // HTTPResponse).
     if (client !== request.client && request.isNavigationRequest()) {
       request.client = client;
     }
--- a/packages/puppeteer-core/src/cdp/NetworkManager.ts
+++ b/packages/puppeteer-core/src/cdp/NetworkManager.ts
@@ -514,17 +514,12 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
   }
 
   #handleRequestRedirect(
-    client: CDPSession,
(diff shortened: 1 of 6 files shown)


OrKoN:
> Fix for https://github.com/puppeteer/puppeteer/issues/13307. I've raised it as a draft initially as there's a potential edge case that this may need to account for in #maybeReassignOOPIFRequestClient - namely, if we end up with different sessions for a document request on the main frame (top-level page). Is this something you're aware can happen? If so, what do you think would be the best way of making this info available for this check in NetworkManager, i.e. does the CDPSession that just received a Network.loadingFinished or Network.loadingFailed event belong to the 'main' target or not?

what do you mean by different sessions for a document request? The only thing that can happen is that the network manager gets a new client on pre-rendering activation but that is not associated with a network request so it should not have an effect on ongoing network activity (except that the client before the activation will get disconnected and its network data would not be available anymore). I think the solution in this PR for updating the client to the client that received the loading finished/failed events makes sense.

ezg27:
Apologies I should have given a clearer example. With the OOPIF scenario, this new check works because the `HTTPRequest` instance for its navigation request is initialised with the client of the main target where its initial CDP events are fired, but then it’s `Network.loadingFinished` or `Network.loadingFailed` are fired on the OOPIF target, so this check can compare the clients in the `#onLoadingFinished` and `#onLoadingFailed` handlers and reassign clients if needed.

My question was is there a possible scenario where we’d receive a navigation request for the main target where its initial CDP events are fired on the main target client (and so it’s `HTTPRequest` instance is initialised with this client), but then it’s `Network.loadingFinished` or `Network.loadingFailed` events are fired on a different client, leading to this new check being true for the main target and its session being reassigned (e.g. if Puppeteer swaps out frame clients for the main target some reason)? If that’s not a possibility then all good :)

I can see there’s some failing tests in CI, i’ll try and take a look at those this evening.

OrKoN:
Got it, thanks for the explanation, so I do not think the scenario you describe can currently happen.

The failing tests seem to be for the WebDriver BiDi implementation that do not support getting the response bodies yet. Let's mark the test as FAIL in test/TestExpectations.json for "webDriverBiDi" on all platforms.

OrKoN:
you can run `npm run format` to fix all formatting/lint issues that could be automatically fixed (it includes also the formatting for the expectation files)

packages/puppeteer-core/src/cdp/NetworkManager.ts
```diff
@@ -514,17 +514,12 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
   }
 
   #handleRequestRedirect(
-    client: CDPSession,
+    _client: CDPSession,
```
OrKoN, line 517: I think we could probably remove all unused params

OrKoN (approved):

PR #13316, opened 2024-11-20
feat: browser level cookies API

Kle0s:

feature - supporting cookies API in the browser context level (and the browser level for the default context).

Yes

Yes

**Summary**

Up until now, the cookies API was only exposed via the Page interface. This only supported setting cookies for the current domain, which is sometimes not what users are looking for.

This is brought up in issue #645 and a part of puppeteer v24 (#12921)

No, this is backward compatible. 

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/api/Browser.ts
+++ b/packages/puppeteer-core/src/api/Browser.ts
@@ -439,7 +439,7 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
   }
 
   /**
-   * Sets cookies in the default {@link BrowserContext}.
+   * Removes cookies from the default {@link BrowserContext}.
    */
   async deleteCookie(...cookies: Cookie[]): Promise<void> {
     return await this.defaultBrowserContext().deleteCookie(...cookies);
--- a/packages/puppeteer-core/src/api/Browser.ts
+++ b/packages/puppeteer-core/src/api/Browser.ts
@@ -15,6 +15,7 @@ import {
   raceWith,
 } from '../../third_party/rxjs/rxjs.js';
 import type {ProtocolType} from '../common/ConnectOptions.js';
+import type {Cookie} from '../common/Cookie.js';
 import type {DownloadBehavior} from '../common/DownloadBehavior.js';
 import {EventEmitter, type EventType} from '../common/EventEmitter.js';
 import {
@@ -24,7 +25,6 @@ import {
   timeout,
   fromAbortSignal,
 } from '../common/util.js';
-import type {Cookie} from '../puppeteer-core.js';
 import {asyncDisposeSymbol, disposeSymbol} from '../util/disposable.js';
 
 import type {BrowserContext} from './BrowserContext.js';
--- a/packages/puppeteer-core/src/api/Browser.ts
+++ b/packages/puppeteer-core/src/api/Browser.ts
@@ -24,6 +24,7 @@ import {
   timeout,
   fromAbortSignal,
 } from '../common/util.js';
+import type {Cookie} from '../puppeteer-core.js';
 import {asyncDisposeSymbol, disposeSymbol} from '../util/disposable.js';
 
 import type {BrowserContext} from './BrowserContext.js';
@@ -423,6 +424,26 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
    * process running.
    */
   abstract disconnect(): Promise<void>;
+
+  /**
+   * Returns all cookies in the default {@link BrowserContext}.
+   */
+  async cookies(): Promise<Cookie[]> {
+    return await this.defaultBrowserContext().cookies();
+  }
+  /**
+   * Sets cookies in the default {@link BrowserContext}.
+   */
+  async setCookie(...cookies: Cookie[]): Promise<void> {
+    return await this.defaultBrowserContext().setCookie(...cookies);
+  }
+
+  /**
+   * Sets cookies in the default {@link BrowserContext}.
+   */
+  async deleteCookie(...cookies: Cookie[]): Promise<void> {
+    return await this.defaultBrowserContext().deleteCookie(...cookies);
+  }
 
   /**
    * Whether Puppeteer is connected to this {@link Browser | browser}.
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -734,6 +734,10 @@ export class BidiPage extends Page {
         !String.prototype.startsWith.call(cookieUrl || '', 'data:'),
         `Data URL page can not have cookie "${cookie.name}"`,
       );
+      assert(
+        typeof cookie.partitionKey === 'string',
+        `BiDi only allows domain partition keys`,
+      );
 
       const normalizedUrl = URL.canParse(cookieUrl)
         ? new URL(cookieUrl)
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -922,7 +922,7 @@ function testUrlMatchCookie(cookie: Cookie, url: URL): boolean {
   return testUrlMatchCookiePath(cookie, normalizedUrl);
 }
 
-function bidiToPuppeteerCookie(bidiCookie: Bidi.Network.Cookie): Cookie {
+export function bidiToPuppeteerCookie(bidiCookie: Bidi.Network.Cookie): Cookie {
   const partitionKey = bidiCookie[CDP_SPECIFIC_PREFIX + 'partitionKey'];
 
   function getParitionKey(): {partitionKey?: string} {
@@ -985,7 +985,7 @@ function cdpSpecificCookiePropertiesFromBidiToPuppeteer(
  * Gets CDP-specific properties from the cookie, adds CDP-specific prefixes and returns
  * them as a new object which can be used in BiDi.
  */
-function cdpSpecificCookiePropertiesFromPuppeteerToBidi(
+export function cdpSpecificCookiePropertiesFromPuppeteerToBidi(
   cookieParam: CookieParam,
   ...propertyNames: Array<keyof CookieParam>
 ): Record<string, unknown> {
@@ -1004,7 +1004,7 @@ function convertCookiesSameSiteBiDiToCdp(
   return sameSite === 'strict' ? 'Strict' : sameSite === 'lax' ? 'Lax' : 'None';
 }
 
-function convertCookiesSameSiteCdpToBiDi(
+export function convertCookiesSameSiteCdpToBiDi(
   sameSite: CookieSameSite | undefined,
 ): Bidi.Network.SameSite {
   return sameSite === 'Strict'
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -758,7 +758,7 @@ export class BidiPage extends Page {
         ...(cookie.sameSite !== undefined
           ? {sameSite: convertCookiesSameSiteCdpToBiDi(cookie.sameSite)}
           : {}),
-        ...(cookie.expires !== undefined ? {expiry: cookie.expires} : {}),
+        ...{expiry: convertCookiesExpiryCdpToBiDi(cookie.expires)},
         // Chrome-specific properties.
         ...cdpSpecificCookiePropertiesFromPuppeteerToBidi(
           cookie,
@@ -1012,4 +1012,10 @@ export function convertCookiesSameSiteCdpToBiDi(
     : sameSite === 'Lax'
       ? Bidi.Network.SameSite.Lax
       : Bidi.Network.SameSite.None;
+}
+
+export function convertCookiesExpiryCdpToBiDi(
+  expiry: number | undefined,
+): number | undefined {
+  return [undefined, -1].includes(expiry) ? undefined : expiry;
 }
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -735,7 +735,7 @@ export class BidiPage extends Page {
         `Data URL page can not have cookie "${cookie.name}"`,
       );
       assert(
-        typeof cookie.partitionKey === 'string',
+        cookie.partitionKey === undefined || typeof cookie.partitionKey === 'string',
         `BiDi only allows domain partition keys`,
       );
 
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -735,10 +735,11 @@ export class BidiPage extends Page {
         !String.prototype.startsWith.call(cookieUrl || '', 'data:'),
         `Data URL page can not have cookie "${cookie.name}"`,
       );
+      // TODO: Support Chrome cookie partition keys
       assert(
         cookie.partitionKey === undefined ||
           typeof cookie.partitionKey === 'string',
-        `BiDi only allows domain partition keys`,
+        'BiDi only allows domain partition keys',
       );
 
       const normalizedUrl = URL.canParse(cookieUrl)
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -32,6 +32,7 @@ import type {
 } from '../cdp/NetworkManager.js';
 import {Tracing} from '../cdp/Tracing.js';
 import type {
+  ChromeCookiePartitionKey,
   Cookie,
   CookieParam,
   CookieSameSite,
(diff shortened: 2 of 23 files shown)


test/TestExpectations.json
```diff
@@ -1121,6 +1121,34 @@
...
+    "parameters": ["cdp", "firefox"],
+    "expectations": ["FAIL"],
+    "comment": "Firefox CDP does not support Storage.getCookies (UnknownMethodError)"
+  },
+  {
+    "testIdPattern": "[cookies.spec] Cookie specs BrowserContext.setCookie should work",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox", "webDriverBiDi"],
+    "expectations": ["FAIL"],
+    "comment": "expiry value `-1` not supported by BiDi"
```
OrKoN: IIRC -1 denotes undefined expiry. Instead of exposing -1 on the public API, let's accept undefined | uint as expiry and transform to -1 for CDP if needed.

OrKoN: e.g., https://github.com/puppeteer/puppeteer/pull/12920/files#diff-5e221f32e7d0b3ccd3f64869c5fc0b1c23e3386d2157dad6f54002ea410b9f8dR246

OrKoN: I remembered it the other way around, let's keep -1 on the interface but when calling a WebDriver BiDi endpoint convert -1 to undefined.

OrKoN: For reference, the WebDriver BiDi interface https://w3c.github.io/webdriver-bidi/#command-storage-setCookie

OrKoN (commented):
Also, we would want to make all breaking changes in the browser version of the API: i.e., support all partition keys properly.

OrKoN:
> Also, we would want to make all breaking changes in the browser version of the API: i.e., support all partition keys properly.

also, we could create a new cookie class to reflect the changes to partition keys but then also fix the expiry to be undefined | uint to align with the WebDriver specification.

test/TestExpectations.json
Kle0s: Okay, I changed it as follows (in bidi/Page.ts):
```ts
...(cookie.expires !== undefined && cookie.expires !== -1 ? {expiry: cookie.expires} : {}),
```
I can technically use something like `!([undefined, -1].includes(cookie.expires))` instead as the condition if you think it's better

Kle0s:
> Also, we would want to make all breaking changes in the browser version of the API: i.e., support all partition keys properly.

What do you mean by that? Currently, there are no breaking changes in this PR (I will remove the page-level cookies API separately). Partition keys are currently supported in each cookie, so I am not sure what I am missing here.

packages/puppeteer-core/src/cdp/BrowserContext.ts
```diff
@@ -100,6 +101,36 @@ export class CdpBrowserContext extends BrowserContext {
...
+
+  override async setCookie(...cookies: Cookie[]): Promise<void> {
+    return await this.#connection.send('Storage.setCookies', {
+      cookies: cookies.map(cookie => {
+        return {
+          ...cookie,
+          partitionKey: cookie.partitionKey
+            ? {
+                topLevelSite: cookie.partitionKey,
+                hasCrossSiteAncestor: false,
```
OrKoN: hasCrossSiteAncestor can be true or false, this is one of the partition keys. The old Cookie object from the page does not support it but we want to support all partition keys (also the ones supported by Firefox over WebDriver BiDi) with the new API.

OrKoN:
> > Also, we would want to make all breaking changes in the browser version of the API: i.e., support all partition keys properly.
> 
> What do you mean by that? Currently, there are no breaking changes in this PR (I will remove the page-level cookies API separately). Partition keys are currently supported in each cookie, so I am not sure what I am missing here.

I mean that the browser-level cookie API should include the breaking changes w.r.t. to the cookie management and not only be a copy of the page-level API (that does not support composite partition keys for example).

test/TestExpectations.json
Kle0s: Actually, I added a method for that (`convertCookiesExpiryCdpToBiDi`) instead:
```ts
export function convertCookiesExpiryCdpToBiDi(
  expiry: number | undefined,
): number | undefined {
  return [undefined, -1].includes(expiry) ? undefined : expiry;
}
```

packages/puppeteer-core/src/cdp/BrowserContext.ts
Kle0s: Okay, I think that the following change in the cookie interface could fix it and preserve backwards compatibility:
```ts
/**
 * Represents a cookie partition key in Chrome.
 * 
 * @public
 */
export interface CookiePartitionKey {
  /**
   * The site of the top-level URL the browser was visiting at the start of the request
   * to the endpoint that set the cookie.
   */
  topLevelSite: string;
  /**
   * Indicates if the cookie has any ancestors that are cross-site to the topLevelSite.
   */
  hasCrossSiteAncestor?: boolean;
}

...

// In the Cookie interface
  /**
   * If specified, deletes cookies in the given partition key. In
   * Chrome, partitionKey matches the top-level site the partitioned
   * cookie is available in. In Firefox, it matches the source origin
   * (https://w3c.github.io/webdriver-bidi/#type-storage-PartitionKey).
   */
  partitionKey?: CookiePartitionKey | string;
```

OrKoN: Right, that sounds good to me, although since this API is new we do need to maintain compatibility with the pages API because it will be removed.

Kle0s: Yeah, I updated the pages API as well, so it could simply be deleted in the future

test/TestExpectations.json
```diff
@@ -3411,6 +3411,13 @@
     "expectations": ["FAIL", "PASS"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
+  {
+    "testIdPattern": "[cookies.spec] Cookie specs Page.setCookie should set cookie with all available properties",
+    "platforms": ["linux"],
+    "parameters": ["cdp", "firefox", "headless"],
```
```suggestion
    "parameters": ["cdp", "firefox"],
```

test/TestExpectations.json
```diff
@@ -3411,6 +3411,13 @@
     "expectations": ["FAIL", "PASS"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
+  {
+    "testIdPattern": "[cookies.spec] Cookie specs Page.setCookie should set cookie with all available properties",
+    "platforms": ["linux"],
```
```suggestion
    "platforms": ["darwin", "linux", "win32"],
```

Kle0s:
Good catches with the test expectations, I should have noticed it. I hope to get better at it in this project with time :)

packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -24,6 +24,7 @@ import {
   timeout,
   fromAbortSignal,
 } from '../common/util.js';
+import type {Cookie} from '../puppeteer-core.js';
```
Lightning00Blade: Nit: wrong import, mostly likely due to IDE auto-import.

Kle0s: Good catch! Fixed.

Kle0s:
Regarding the `[cookies.spec] Cookie specs BrowserContext.setCookie should work` failure; I noticed the Page.setCookie also fails similarly with a test expectation I did not fully understand - `Failes with fieldtrial testing config in Chrome`. Is it relevant for the browser context one as well?

OrKoN:
I think there was a field trial (a feature) rolled out in Chrome to change how cookies can be set. It should fail in the same way with the Storage domain API and we will need to adjust the test after investigating the actual cause for the failure.

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -734,6 +735,11 @@ export class BidiPage extends Page {
         !String.prototype.startsWith.call(cookieUrl || '', 'data:'),
         `Data URL page can not have cookie "${cookie.name}"`,
       );
+      assert(
```
OrKoN, line 739: let's add a TODO comment in these assertions linking to https://github.com/GoogleChromeLabs/chromium-bidi/issues/2434

packages/puppeteer-core/src/common/Cookie.ts
```diff
@@ -29,6 +29,23 @@ export type CookiePriority = 'Low' | 'Medium' | 'High';
  */
 export type CookieSourceScheme = 'Unset' | 'NonSecure' | 'Secure';
 
+/**
+ * Represents a cookie partition key in Chrome.
+ *
+ * @public
+ */
+export interface ChromeCookiePartitionKey {
```
OrKoN, line 37: Let's also define a generic partition key structure as defined by WebDriver BiDi. Currently, it should only contain the sourceOrigin attribute. Let's map sourceOrigin to topLevelSite and vice versa. We should still keep this interface for users only interested in Chrome-specific partition attributes.

test/src/cookies.spec.ts
```diff
@@ -834,4 +834,62 @@ describe('Cookie specs', () => {
       expect(await page.cookies()).toHaveLength(0);
     });
   });
+  describe('BrowserContext.cookies', () => {
```
OrKoN, line 837: let's add tests covering deletion of cookies.

test/src/cookies.spec.ts
```diff
@@ -834,4 +834,62 @@ describe('Cookie specs', () => {
...
+          size: 16,
+          httpOnly: false,
+          secure: false,
+          session: true,
+          sourceScheme: 'NonSecure',
+        },
+      ]);
+    });
+  });
+  describe('BrowserContext.setCookie', function () {
```
OrKoN, line 896: let's add tests for setting and retrieving cookies using Puppeteer and tests using various partition keys

packages/puppeteer-core/src/common/Cookie.ts
Kle0s: So, do we want to still support receiving string or only one of the `ChromeCookiePartitionKey` and `BiDiCookiePartitionKey` interfaces (and `undefined` of course)?

packages/puppeteer-core/src/bidi/Page.ts
Kle0s: Done

test/src/cookies.spec.ts
Kle0s: added

packages/puppeteer-core/src/common/Cookie.ts
Kle0s: In any case, I think it might be worthwhile to separate this issue from the current one

test/src/cookies.spec.ts
Kle0s: Done

packages/puppeteer-core/src/cdp/BrowserContext.ts
```diff
@@ -100,6 +102,33 @@ export class CdpBrowserContext extends BrowserContext {
...
   }
 
+  override async cookies(): Promise<Cookie[]> {
+    const {cookies} = await this.#connection.send('Storage.getCookies', {
+      browserContextId: this.#id || undefined,
+    });
+    return cookies.map(cookie => {
+      return {
+        ...cookie,
+        partitionKey: cookie.partitionKey
```
OrKoN: I believe we want to return the complete partitionKey here and not only the topLevelSite.

test/src/cookies.spec.ts
OrKoN: I think a test verifying that a cookie set by `context.setCookie()` can be retrieved intact by `context.cookies()` is missing. In particular, the case of retrieving partitioned cookies is not covered.

packages/puppeteer-core/src/cdp/BrowserContext.ts
Kle0s: You are correct, fixed by returning the cookies directly (without this modification)

test/src/cookies.spec.ts
Kle0s: Added a test for that (`should find partitioned cookie`)

Kle0s:
I am not sure about these failures personally. IMO they are not related to my change. Is it possible some of the tests are flaky or something like that? @OrKoN

test/TestExpectations.json
```diff
@@ -244,18 +244,32 @@
...
-    "testIdPattern": "[cookies.spec] Cookie specs Page.setCookie should set cookie with all available properties",
+    "testIdPattern": "[cookies.spec] Cookie specs BrowserContext.setCookie should set cookie with chrome partition key",
     "platforms": ["darwin", "linux", "win32"],
-    "parameters": ["firefox"],
+    "parameters": ["webDriverBiDi"],
+    "expectations": ["FAIL"],
+    "comment": "BiDi partition key requires refactor (no longer string)"
+  },
+  {
+    "testIdPattern": "[cookies.spec] Cookie specs BrowserContext.setCookie should set cookie with string partition key",
```
OrKoN, line 261: string partition key should be supported in WebDriver BiDi, right? it is the sourceOrigin attribute.

OrKoN:
> I am not sure about these failures personally. IMO they are not related to my change. Is it possible some of the tests are flaky or something like that? @OrKoN

I have retried the failing ones.

test/TestExpectations.json
Kle0s: No, as I mentioned in another comment, I think it should be passed as a json with the field `sourceOrigin` set to the value and not as a string. This is the reason the comment I set there is about refactoring the BiDi cookie interface next

OrKoN: I will address it before this ships, so I think this PR is ready to be landed.

OrKoN (approved):

karlhorky:
Ah, looks like this was marked as deprecated [somewhere between `puppeteer-core@23.9.0` and `puppeteer-core@23.10.0`](https://app.renovatebot.com/package-diff?name=puppeteer-core&from=23.9.0&to=23.10.0)

ESLint with `eslint-plugin-deprecation` is starting to fail CI with the deprecation warnings:

```
$ eslint . --max-warnings 0

test.ts
  469:35  warning  `cookies` is deprecated. Page-level cookie API is deprecated. Use
{@link Browser.cookies} or {@link BrowserContext.cookies} instead  @typescript-eslint/no-deprecated

util/puppeteer/cookies.ts
  52:14  warning  `deleteCookie` is deprecated. Page-level cookie API is deprecated. Use
ESLint found too many warnings (maximum: 0).
{@link Browser.deleteCookie} or {@link BrowserContext.deleteCookie}
instead  @typescript-eslint/no-deprecated

...

✖ 3 problems (0 errors, 3 warnings)
```

for these two code patterns:

```ts
await page.cookies()
```

and

```ts
await page.deleteCookie(
  ...updatedCookies.map(({ partitionKey, ...partialCookie }) => {
    return {
      ...partialCookie,
      partitionKey: partitionKey?.topLevelSite,
    };
  }),
);
```

Unfortunately:

1. the error message in the CI failure is not super helpful
2. the [`puppeteer-core@23.10.0` release notes](https://github.com/puppeteer/puppeteer/releases/tag/puppeteer-core-v23.10.0) also don't have any easy guide for migrating / updating

So I guess time to start digging as to how to fix this...

karlhorky:
Looks like this deprecation was performed in PR #13334 but not documented in release notes...

- https://github.com/puppeteer/puppeteer/pull/13334

Digging deeper...

karlhorky:
First PR open for some small missing parts of the docs:

- https://github.com/puppeteer/puppeteer/pull/13370

karlhorky:
## Migrating from `page.cookies()` and `page.deleteCookies()`

In the end, it seems like the migration is not a huge deal, but it was not openly documented, and it was hard to find.

`page.cookies()` to `browser.cookies()`

```diff
-await page.cookies()
+await browser.cookies()
```

`page.deleteCookie()` to `browser.deleteCookie()`

```diff
-await page.deleteCookie(
-  ...updatedCookies.map(({ partitionKey, ...partialCookie }) => {
-    return {
-      ...partialCookie,
-      partitionKey: partitionKey?.topLevelSite,
-    };
-  }),
-);
+await browser.deleteCookie(...updatedCookies);
```

OrKoN:
We do not include any of the documentation changes in the changelog. The deprecated methods and their replacement are documented in the API docs, for example, https://pptr.dev/api/puppeteer.page.cookies/

karlhorky:
> We do not include any of the documentation changes in the changelog. The deprecated methods and their replacement are documented in the API docs

From a pure Puppeteer team process perspective, can understand that, and it makes sense.

From a user perspective, this leads to frustrating, hard-to-track-down breakages when upgrading Puppeteer versions.

OrKoN:
> > We do not include any of the documentation changes in the changelog. The deprecated methods and their replacement are documented in the API docs
> 
> From a pure Puppeteer team process perspective, can understand that, and it makes sense.
> 
> From a user perspective, this leads to frustrating, hard-to-track-down breakages when upgrading Puppeteer versions.

well, there was no breakage as it is documentation update. Once it's actually removed, it will be in a breaking-change release according to semver with the mention in the changelog. I agree thought it would be nice to also list new deprecations in the changelog, filed https://github.com/puppeteer/puppeteer/issues/13371

karlhorky:
Amazing, thanks for the consideration! I'll keep an eye on https://github.com/puppeteer/puppeteer/issues/13371

Issue #9757, opened 2023-03-01
[Feature]: Ability to get platform fonts for a selector.


dardavar:
### Feature description

CDP supports [getPlatformFontsForNode](https://chromedevtools.github.io/devtools-protocol/tot/CSS/#method-getPlatformFontsForNode) which gets all the rendered fonts for a given node. It will be nice to have this feature in Puppeteer as well.

Kle0s:
I second this @OrKoN. It is a rather small feature once we figure the API we wish to provide users with

OrKoN:
Could you please provide more details about your use case? Currently, it is possible to get this data with a custom session but it requires using a private API:

``` 
import puppeteer from 'puppeteer';

const browser = await puppeteer.launch({
    headless: false
});

const page = await browser.newPage();

await page.goto('https://wikipedia.org', {
    waitUntil: ['networkidle2']
});

const element = await page.$('.lang-list-button-text');

async function getFonts(element) {
    const {node} = await element.handle.client.send('DOM.describeNode', {
        objectId: element.remoteObject().objectId,
    });
    const session = await page.createCDPSession();
    try {
        await session.send('DOM.enable');
        await session.send('CSS.enable');
        await session.send('DOM.getDocument', {
            depth: -1,
        });
        const {node: nodeInSession} = await session.send('DOM.describeNode', {
            backendNodeId: node.backendNodeId,
        });
        const result = await session.send('CSS.getPlatformFontsForNode', {
            nodeId: nodeInSession.nodeId,
        });
        return result;
    } finally {
        await session.detach();
    }
}

console.log(await getFonts(element));

await browser.close();
```

dardavar:
The approach relies on private properties, which I’d prefer to avoid because they tend to change between versions.

In my case, the web app renders different fonts based on context, and I need a reliable way to confirm the correct font is applied. Adding this as a supported feature would provide a stable solution for this kind of testing.

OrKoN:
With this change https://github.com/puppeteer/puppeteer/pull/13328 no private properties will be needed.

OrKoN:
> In my case, the web app renders different fonts based on context, and I need a reliable way to confirm the correct font is applied. 

are you traversing all elements to collect all fonts or how do you check it? would taking a screenshot be a better alternative to make sure everything renders correctly?

PR #13327, opened 2024-11-25
feat: get rendered fonts for element

Kle0s:

feature

Yes
Yes
**Summary**
Adds the ability to get the rendered fonts (and some metadata) for text nodes, based on element handle. 
Solves #9757 

No

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/cdp/ElementHandle.ts
+++ b/packages/puppeteer-core/src/cdp/ElementHandle.ts
@@ -204,18 +204,28 @@ export class CdpElementHandle<
   }
 
   override async renderedFonts(): Promise<PlatformFontUsage[]> {
-    await this.client.send('DOM.getDocument', {depth: -1});
-    const nodeInfo = await this.client.send('DOM.requestNode', {
-      objectId: this.id as string,
-    });
-    const nodeId = nodeInfo.nodeId;
-    await this.client.send('CSS.enable');
-    const platformFonts = await this.client.send(
-      'CSS.getPlatformFontsForNode',
-      {
-        nodeId,
-      },
-    );
-    return platformFonts.fonts;
+    try {
+      // In order to request a node and receive its node id, the document must be
+      // requested first
+      await this.client.send('DOM.getDocument', {depth: -1});
+      // In order to use a CSS-domain method, the CSS domain must be enabled
+      await this.client.send('CSS.enable');
+
+      const nodeInfo = await this.client.send('DOM.requestNode', {
+        objectId: this.id as string,
+      });
+      const nodeId = nodeInfo.nodeId;
+
+      const platformFonts = await this.client.send(
+        'CSS.getPlatformFontsForNode',
+        {
+          nodeId,
+        },
+      );
+      return platformFonts.fonts;
+    } finally {
+      await this.client.send('CSS.disable');
+      await this.client.send('DOM.disable');
+    }
   }
 }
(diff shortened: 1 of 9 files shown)


packages/puppeteer-core/src/cdp/ElementHandle.ts
```diff
@@ -201,4 +202,30 @@ export class CdpElementHandle<
...
+      // In order to use a CSS-domain method, the CSS domain must be enabled
+      await this.client.send('CSS.enable');
+
+      const nodeInfo = await this.client.send('DOM.requestNode', {
+        objectId: this.id as string,
+      });
+      const nodeId = nodeInfo.nodeId;
+
+      const platformFonts = await this.client.send(
+        'CSS.getPlatformFontsForNode',
```
OrKoN, line 220: I would not like to bring this method to Puppeteer as the method exists primarily for the DevTools frontend and it is very specific to the UI logic there. 

How about we document instead how to call this using a custom session created using createCDPSession? Also, what is the use case for getting the font data via Puppeteer? If the use case is important, I think we could add this method as experimental but let us fix the [upstream method](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/inspector_css_agent.cc;l=1900;drc=277f4ab48eb85f7441f78aed191c31068ce89814) first so that it does not require DOM and CSS domains to be enabled and so that it accepts a backend node ID without the need to rely on the fragile DOM nodeIds (they reset on every call to DOM.getDocument so the method might fail if called concurrently).

Kle0s: For me, the use case is testing whether or not stylesheets are written correctly, but I am sure we can find more use cases for it. I prefer to "hide" the implementation in puppeteer and not access private properties, as they tend to break down more frequently.

I also would have preferred having the upstream method fixed so it would accept a backend node ID / remote object ID. I try to avoid basing things on the nodeIds in my other projects as well, as I have experienced them changing frequently in some cases. 

This PR can wait if this change can happen soon

OrKoN: With https://github.com/puppeteer/puppeteer/pull/13328 the private property access will not be needed. Would you like to contribute to the fixes in Chromium?

Kle0s: Sure. I am not familiar enough with the code base but this seems like a relatively minor issue, so maybe I can start with it

OrKoN: here are the getting started instructions for building/testing https://www.chromium.org/developers/how-tos/get-the-code/ , it would require updating https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/public/devtools_protocol/browser_protocol.pdl;l=1;drc=93fc8bf077eb95148a50f71b9917ebe6bd5a02d3 to add optional objectId and backendNodeId params, then re-build to update the generated classes, and then in https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/inspector_css_agent.cc;l=1911;drc=b59cf80380080c28b981e2dcd491ef3efa9d8808 use the overload from here https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/inspector_dom_agent.cc;l=430;drc=1651676a30cd7abcd177975f7cd0e37bd945f663 Then the check if the agent is enabled can be removed and a test needs to be added verifying that it works with the new ids and also without the CSS domain enabled.

Kle0s: The documentation for working with Chromium is unfortunately wildly outdated, so after some time I gave up :/

OrKoN:
Closing as stale.

PR #13328, opened 2024-11-25
feat: implement ElementHandle.backendNodeId

### LINKED ISSUE #9284: [Feature]: Expose `BackendNodeId` on `ElementHandle`
**Why?**

To be able to do things with Nodes in CDP, such as set breakpoints, one needs the `NodeId`. However, in the current puppeteer versions one needs to set up their own CDP session using `Target#createCDPSession`, and there `DOM.requestNode` and `DOM.describeNode` will not work with `RemoteObjectId`s obtained through `ElementHandle#remoteObject` from puppeteer's CDP session (and `NodeId`s themselves of puppeteer's CDP session would also be useless). Hence, one needs to adopt the Node using logic like in [`IsolatedWorld#adoptHandle`](https://github.com/puppeteer/puppeteer/blob/65007a29/packages/puppeteer-core/src/common/IsolatedWorld.ts#L605). For this to work, however, one needs to first obtain the `BackendNodeId` using `DOM.describeNode` in *puppeteer*'s CDP session. Currently, this is only possible by using `Page#_client`, which may break in the future.

**Solution**

Create an `ElementHandle#backendNodeId` method which returns the `BackendNodeId` of the Node obtained through `DOM.describeNode`.

Related to: #3641

OrKoN:
When run over CDP, the getter will provide the https://chromedevtools.github.io/devtools-protocol/tot/DOM/#type-BackendNodeId for the element handle which allows identifying the element in other CDP sessions.

Closes https://github.com/puppeteer/puppeteer/issues/9284

### DIFF
--- a/docs/api/puppeteer.elementhandle.backendnodeid.md
+++ b/docs/api/puppeteer.elementhandle.backendnodeid.md
@@ -1,0 +1,19 @@
+---
+sidebar_label: ElementHandle.backendNodeId
+---
+
+# ElementHandle.backendNodeId() method
+
+When connected using Chrome DevTools Protocol, it returns a DOM.BackendNodeId for the element.
+
+### Signature
+
+```typescript
+class ElementHandle {
+  abstract backendNodeId(): Promise<number>;
+}
+```
+
+**Returns:**
+
+Promise&lt;number&gt;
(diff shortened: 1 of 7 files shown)


test/src/elementhandle.spec.ts
```diff
@@ -1205,4 +1205,13 @@ describe('ElementHandle specs', function () {
       expect(handle.disposed).toBeFalsy();
     });
   });
+
+  describe('ElementHandle.backendNodeId', () => {
```
Lightning00Blade: Let's wrap in a CDP only describe? Or even move to cdp folder ?

Lightning00Blade (approved):

PR #13331, opened 2024-11-26
feat: adjust browser-level cookie API

OrKoN:
- move new API's tests to a separate spec.
- use a different type for setting cookies as not all attributes of Cookie are applicable.
- fix various tests requiring secure origins.
- fix the issue with always using default context to set cookies.


packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -929,14 +929,25 @@ function testUrlMatchCookie(cookie: Cookie, url: URL): boolean {
   return testUrlMatchCookiePath(cookie, normalizedUrl);
 }
 
-export function bidiToPuppeteerCookie(bidiCookie: Bidi.Network.Cookie): Cookie {
+export function bidiToPuppeteerCookie(
```
Lightning00Blade, line 932: Non blocking: Should we move these out of Page an into separate file to make it clear that they are utility now that we are exporting them?

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -1028,10 +1039,15 @@ export function convertCookiesExpiryCdpToBiDi(
...
-  partitionKey: ChromeCookiePartitionKey | string | undefined,
+  partitionKey: CookiePartitionKey | string | undefined,
 ): string | undefined {
   if (partitionKey === undefined || typeof partitionKey === 'string') {
     return partitionKey;
   }
-  return partitionKey.topLevelSite;
+  if (partitionKey.hasCrossSiteAncestor) {
+    throw new UnsupportedOperation(
+      'WebDriver BiDi does not support goog:hasCrossSiteAncestor yet.',
```
Lightning00Blade: I think we don't want to mention the prefix
```suggestion
      'WebDriver BiDi does not support `hasCrossSiteAncestor` yet.',
```

test/src/browsercontext-cookies.spec.ts
```diff
@@ -0,0 +1,186 @@
+/**
+ * @license
+ * Copyright 2018 Google Inc.
```
Lightning00Blade:
```suggestion
 * Copyright 2024 Google Inc.
```

packages/puppeteer-core/src/bidi/Page.ts
OrKoN: I think we can move them completely to BrowserContext once we remove the page-level cookies API.

Lightning00Blade (approved):

PR #13332, opened 2024-11-26
refactor: merge various connect and launch option types

OrKoN:
Technically, it could be a breaking change since the types are slightly different but the functionality is still the same and the new types are a superset of the previous scattered types. Old types are made aliases to the new types and deprecated.


Lightning00Blade (approved):
LGTM, but would consider adding the older names as alias export so we don't brake builds of anyone that is using them, WDYT? Then we remove them in v24

Lightning00Blade (approved):
LGTM, added it to the list for v24

lameuler:
this seems to missing the type alias for `PuppeteerLaunchOptions` which has been replaced by the new `LaunchOptions`? so could cause some issues especially since this change doesn't show up in the changelog

OrKoN:
Thanks! Indeed I overlooked PuppeteerLaunchOptions. https://github.com/puppeteer/puppeteer/pull/13376 should fix this

PR #13339, opened 2024-11-28
chore: support raw error mode

OrKoN:
skip wrapping protocol errors for WebDriver BiDi adapters.


packages/puppeteer-core/src/cdp/CDPSession.ts
```diff
@@ -40,12 +41,14 @@ export class CdpCDPSession extends CDPSession {
     targetType: string,
     sessionId: string,
     parentSessionId: string | undefined,
+    rawErrors: boolean,
```
Lightning00Blade, line 44: Mark as internal

OrKoN: the entire constructor is internal here. For connection, I think it is okay to keep it public.

Lightning00Blade (approved):

PR #13340, opened 2024-11-28
feat: support LaunchOptions in executablePath()

### LINKED ISSUE #13308: [Bug]: PuppeteerNode.executablePath() method doesn't take into account headless mode

```TypeScript
console.log(puppeteer.executablePath())
```

### Background

`PuppeteerNode.executablePath()` returns the path of the main Chrome executable. In fact if you `launch()` with `headless: 'shell'` then this is not the executable that Puppeteer will run. Not only is there no way to ask Puppeteer which executable it will run for 'old headless' mode, the fact that the value returned by this function isn't necessarily the file that will be executed isn't documented.

### Expectation

The moon on a stick.

### Reality

Reality is often a sad let-down.

_No response_

### Puppeteer version

23.8.0

### Node version

n/a

### Package manager

npm

### Package manager version

n/a

### Operating system

Linux

OrKoN:
This allows determining the correct path based on the headless attribute and potentially other launch options later.

Drive-by: add the standard path for Chrome Canary on Linux.

Closes https://github.com/puppeteer/puppeteer/issues/13308

### DIFF
--- a/packages/puppeteer-core/src/node/PuppeteerNode.ts
+++ b/packages/puppeteer-core/src/node/PuppeteerNode.ts
@@ -60,7 +60,7 @@ import type {ChromeReleaseChannel, LaunchOptions} from './LaunchOptions.js';
  * @public
  */
 export class PuppeteerNode extends Puppeteer {
-  #_launcher?: BrowserLauncher;
+  #launcher?: BrowserLauncher;
   #lastLaunchedBrowser?: SupportedBrowser;
 
   /**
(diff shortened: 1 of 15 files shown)


test/src/mocha-utils.ts
```diff
@@ -317,19 +317,29 @@ if (
...
   }
   -> mode: ${
-    processVariables.isHeadless
-      ? processVariables.headless === 'true'
-        ? '--headless=new'
-        : '--headless'
+    isHeadless
+      ? processVariables.headless === 'shell'
+        ? 'shell'
+        : 'new headless'
```
Lightning00Blade:
```suggestion
        : 'headless'
```

Lightning00Blade (approved):

packages/puppeteer-core/src/node/PuppeteerNode.ts
```diff
@@ -150,41 +150,67 @@ export class PuppeteerNode extends Puppeteer {
...
    * The default executable path.
    */
-  executablePath(channel?: ChromeReleaseChannel): string {
-    return this.#launcher.executablePath(channel);
+  executablePath(): string;
+  executablePath(optsOrChannel?: ChromeReleaseChannel | LaunchOptions): string {
+    if (optsOrChannel === undefined) {
+      return this.#getLauncher(this.lastLaunchedBrowser).executablePath(
+        undefined,
+        /* validatePath= */ false,
```
lameuler, line 202: am I right to say that the previous behaviour was to always validate the path in `executablePath`, so this could be a breaking change? since it no longer checks if the path exists from what I can see.

is there any way to still make it validate the path or would that have to be manually done now

OrKoN: right, it could be viewed as a breaking change but the documented behavior for this method always been that it returns the default path to the browser (without any validation) and the validation of the path was probably done for the cases when it is called to launch the browser internally. Note that the path is still validated when launching the browser.

Issue #5611, opened 2020-04-08
Puppeteer install hangs indefinitely


sbocinec:
I'm experiencing one weird issue with `npm install puppeteer` - the installation idenfinitely hangs on ` Downloading Chromium r662092 - 111.6 Mb [====================] 99% 0.0s`.

The same has been already reported in https://github.com/puppeteer/puppeteer/issues/4688 , but closed as automagically solved.

I couldn't believe there is any magic involved so I decided to troubleshoot it and found a solution to install the module on my environment without the installation hanging. Despite having solution, I can reliably replicate the issue. In short:
* the issue only occurs when I'm connected to a 802.11n 2.4GHz wifi SSID
* being connected to a 5GHz 802.11ac SSID configured on the same wifi access point, the `npm install puppeteer` works as expected.

Even it looks like the culprit might be in the misconfigured AP or some weird network connection/configuration, I think, the real culprit lies somewhere in the install scripts: `install.js` / `BrowserFetcher.js`, where downloading the last chunk of the chromium browser somehow never finish.

### Steps to reproduce

I'm worried you won't be able to replicate the issue, but I will provide as much as possible data so that it might be possible to find the bug inside the code despite not being able to replicate it on your own machine.

**Tell us about your environment:**

* Puppeteer version: 1.17.0, 1.18.0, 1.18.1, 1.19.0, .1.20.0
* Platform / OS version: Linux, 64bit, Ubuntu 19.10
* URLs (if applicable): https://storage.googleapis.com/chromium-browser-snapshots/Linux_x64/662092/chrome-linux.zip (the chromium artifact being downloaded)
* Node.js version: v12.16.1, v12.16.0, v12.13.0 (tried with all)

**What steps will reproduce the problem?**

1. Connect to my local 802.11n 2.4GHZ SSID. ( AP running OpenWRT v19.07.2)
2. Install puppeteer: `npm install puppeteer@1.17.0`
3. Installation hangs indefinitely with following output:
```
$ npm install puppeteer@1.17.0

> puppeteer@1.17.0 install /home/p/workspace/p/node_modules/puppeteer
> node install.js

Downloading Chromium r662092 - 111.6 Mb [====================] 99% 0.0s
```

The same is for other versions:
```
$ npm install puppeteer@1.18.1

> puppeteer@1.18.1 install /home/p/workspace/p/node_modules/puppeteer
> node install.js

Downloading Chromium r672088 - 112.1 Mb [====================] 99% 0.0s
```
[...]

PR #13349, opened 2024-12-01
Add note about yarn dlx

mifi:
See https://github.com/puppeteer/puppeteer/issues/5611#issuecomment-2509792650

documentation

**Summary**

sometimes `npx @puppeteer/browsers install` will hang indefinitely. add a suggestion to use `yarn dlx` instead

no

**Other information**

### DIFF
--- a/packages/browsers/README.md
+++ b/packages/browsers/README.md
@@ -40,6 +40,7 @@ npx @puppeteer/browsers install chromedriver@116.0.5793.0
 ## Known limitations
 
 1. Launching the system browsers is only possible for Chrome/Chromium.
+2. If `npx` gets stuck or runs very slowly, you can try `yarn dlx` instead. E.g. `yarn dlx @puppeteer/browsers install ...`
 
 ## API
 


OrKoN (changes requested):
npx should be equivalent to yarn dlx and we should not require users to switch to yarn to work around some issues. Could you please file a bug report with the reproducible example? are you using the same version with both package runners? is it hanging inside  the respective runner or is it actually a hang during the download?

mifi:
I cannot reproduce the problem with a simple test case, I can only reproduce it by re-running this old Github Action:

https://github.com/mifi/reactive-video/actions/runs/12104497564/job/33771544221

it hangs every time i re-run that job. But I tried copying the stuff from tests.yml into a new branch but it cannot reproduce the problem there. I don't know what's going on but it seems the conditions to trigger this bug are hard to reproduce. I'm not going to spend any more time on this, so I guess we can close this if you don't want to have dlx as a backup in the docs.

OrKoN:
so I started a fresh job and I cannot reproduce it https://github.com/puppeteer/puppeteer/actions/runs/12114590215/job/33771385733?pr=13350

Which version does the old job using? the was a bug for older Chrome versions and older @puppeteer/browsers versions related to redirect handling. The server now redirects some old URLs to new URLs and older @puppeteer/browsers versions could not handle a redirect. Perhaps, this is what you are seeing in the old job. Let me try with that specific version.

OrKoN:
It does not seem to reproduce on macos-13 with Chrome 117.0.5938.149 https://github.com/puppeteer/puppeteer/actions/runs/12115127014/job/33772886185?pr=13350 So the only possible explanation is that the job somehow uses an old version of @puppeteer/browsers or that there is some new bug 🤔

mifi:
Doesn't `npx @puppeteer/browsers` use the newest stable version of `@puppeteer/browsers`? I ran it 30 minutes ago so I would think it used the newest version.

OrKoN:
> Doesn't `npx @puppeteer/browsers` use the newest stable version of `@puppeteer/browsers`? I ran it 30 minutes ago so I would think it used the newest version.

It will use the newest version, unless there is a version installed in the repository. In that case, it will use the version installed in the current project. `npx --yes @puppeteer/browsers@latest` would ensure the latest version is always installed.

mifi:
Oh, right! that explains it then. It was using `@puppeteer/browsers` from node_modules, which happened to be `@puppeteer/browsers@npm:1.7.1`! thanks for the pointers.
It also explains why I cannot reproduce it anymore because `@puppeteer/browsers` is now 2.4.1. Maybe the docs should be changed to recommend using the `--yes` flag?

OrKoN:
> I see the --yes flag is just for confirming the prompt. maybe instead change all the commands in the docs to be npx @puppeteer/browsers@latest to avoid having old bugs pop up like this?

we could mention that but this is generally a feature of npx and it is documented there (it should be very similar to yarn dlx).  Many projects also prefer to run the project version to avoid sudden breakage if a new version is released that contains incompatible changes.

mifi:
> Many projects also prefer to run the project version to avoid sudden breakage if a new version is released that contains incompatible changes.

I see. But in the case of `@puppeteer/browsers` I'd argue the opposite is true: I got this error because `@puppeteer/browsers` was too old and I hadn't run the action in a long time. If I had used `@latest` it would work. So running an old version is likely to break because maybe URLs change and servers get taken down. Also evident by the number of people who are possibly having the same issue here: #5611

OrKoN:
Will be addressed by https://github.com/puppeteer/puppeteer/pull/13351

Issue #13083, opened 2024-09-12
[Feature]: Add unpacking indicator to the install command 


Lightning00Blade:
### Feature description

Currently we show a downloading indicator in the terminal when a browser is downloading, but after a successful download we try to unpack and do other operations which are not communicated to the user, which looks like the process is stuck, we should show some loading state there to make sure the user understands what is going on.

uzairabid1:
Hi @Lightning00Blade , I would like to work on this issue, took a quick look at the `makeProgressCallback()` I think we can add a condition there to log this info!

Lightning00Blade:
Hey @uzairabid1, the unpacking of the download is happening at https://github.com/puppeteer/puppeteer/blob/main/packages/browsers/src/install.ts#L315 
I am not sure how well we can track the extraction process progression, but we can overwrite the ProgressBar to change from downloading (with indicator) to extracting (no progress indicator).

uzairabid1:
Thanks for replying, ah okay, I believe `install.ts` is that programmatic one, can we just add a simple debug there that if `options.downloadProgressCallback` we add a debug that says something about extraction being started until we wait for `unpackArchive()` to finish?

githubbzxs:
Opened PR: https://github.com/puppeteer/puppeteer/pull/14716

Lightning00Blade:
Closed by https://github.com/puppeteer/puppeteer/pull/15045

PR #13356, opened 2024-12-03
feat: add unpacking indicator to the install command

Abhash-Chakraborty:
Summary of Updates in CLI.ts:

Enhanced error handling and logging for improved user feedback during CLI operations.
Streamlined argument parsing for clarity and efficiency.
Added debug messages for tracking command execution and potential failures.
Optimized the CLI workflow to handle edge cases and ensure robust functionality.

Bug: #13083

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -468,6 +468,8 @@ export function makeProgressCallback(
 ): (downloadedBytes: number, totalBytes: number) => void {
   let progressBar: ProgressBar;
   let lastDownloadedBytes = 0;
+  let isExtracting = false; // Track if we're in the extraction phase
+
   return (downloadedBytes: number, totalBytes: number) => {
     if (!progressBar) {
       progressBar = new ProgressBar(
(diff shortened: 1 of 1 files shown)


packages/browsers/src/CLI.ts
```diff
@@ -482,9 +484,16 @@ export function makeProgressCallback(
         },
       );
     }
-    const delta = downloadedBytes - lastDownloadedBytes;
-    lastDownloadedBytes = downloadedBytes;
-    progressBar.tick(delta);
+    if (downloadedBytes >= totalBytes && !isExtracting) {
+      // Transition to the extraction phase
+      progressBar.terminate(); // Clear download progress bar
+      console.log(`Extracting ${browser} ${buildId}...`); // Simple console message for now
```
OrKoN, line 490: I believe this would report extracting even when no extracting is happening (`options.unpack === false`) and also it  would not cover the waiting for running setup and installing deps (`options.installDeps === true`). 

We probably also should not use console.log directly as the progress bar is only rendering if the [output stream isTTY](https://github.com/visionmedia/node-progress/blob/master/lib/node-progress.js#L121). 

I think using a static message is fine but we should also write it to process.stderr as the progress bar and we can add messages directly to the corresponding code that does extraction/deps/setup conditions on the presence of downloadProgressBar that indicates whether the code is invoked from CLI.

OrKoN (changes requested):

Lightning00Blade:
I will close this PR, as its been more then a 3 months without update. 
If you want it re-open please write here or just open a new one.

PR #13362, opened 2024-12-04
test: remove unused mergeArgs in mocha-utils

OrKoN:


Lightning00Blade (approved):
LGTM, but we may want this as currently we spread the new args over the default one. 
Though no use case for this was found while refactoring.

PR #13385, opened 2024-12-09
feat: allow listing installed browsers

slhck:

This adds a mode for `puppeteer/browsers` to `list` all installed browsers in the cache.

Sure! A `list.spec.ts` file has been added.

Yes, the `README.md` has been updated. (I also added some info on the `clear` command which was missing.)

**Summary**

Example:

```
➜ npx @puppeteer/browsers list
chrome@133.0.6886.0 (mac_arm) /Users/werner/Documents/Software/puppeteer/packages/browsers/chrome/mac_arm-133.0.6886.0/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing
```

It prints the type of browser, its version, platform, and the path.

This allows better maintenance of the currently installed versions, as the default cache directory is determined programmatically, and it's hard to understand what browsers are installed otherwise.

No.

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -54,6 +54,10 @@ interface ClearArgs {
   path?: string;
 }
 
+interface ListArgs {
+  path?: string;
+}
+
 /**
  * @public
  */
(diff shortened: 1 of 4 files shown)


slhck:
I noticed the `@puppeteer/browsers clear` command requires explicitly adding the path of the cache, but here we default to using `pwd()` if it's not specified. This might have to be changed? It would be nice to programmatically get the path that the install command uses.

packages/browsers/src/CLI.ts
```diff
@@ -396,6 +400,45 @@ export class CLI {
...
+              return;
+            }
+
+            for (const browser of browsers) {
+              console.log(
+                `${browser.browser}@${browser.buildId} (${browser.platform}) ${browser.executablePath}`,
+              );
+            }
+          } catch (error) {
+            console.error(`Failed to list browsers: ${error}`);
```
OrKoN: I think we probably do not want to catch an error here so that the whole command exits with an error status code.

packages/browsers/src/CLI.ts
```diff
@@ -396,6 +400,45 @@ export class CLI {
...
+        async argv => {
+          const args = argv as unknown as ListArgs;
+          const cacheDir = args.path ?? this.#cachePath;
+
+          try {
+            const cache = new Cache(cacheDir);
+            const browsers = cache.getInstalledBrowsers();
+
+            if (browsers.length === 0) {
+              console.log('No browsers installed.');
```
OrKoN: I wonder if this should return nothing to indicate that there are no browsers installed. I would help to use the command output with other tools.

slhck:
Pushed with changes, thanks for the quick feedback!

OrKoN (approved):
Thanks!

PR #13391, opened 2024-12-10
fix: support .tar.xz archives for Firefox 135+

### LINKED ISSUE #13390: Puppeteer tests against Firefox Nightly are failing
According to @whimboo , the archive format has changed from .tar.bz2 to .tar.xz.

Example failure: https://github.com/puppeteer/puppeteer/actions/runs/12251886667/job/34178310393#step:8:46

This is a breaking change for Puppeteer and users of `@puppeteer/browsers`. The change will presumably roll out to other channels as well.  See https://blog.nightly.mozilla.org/2024/11/28/announcing-faster-lighter-firefox-downloads-for-linux-with-tar-xz-packaging/

If the previous format of archives cannot be provided alongside the new ones, the following needs to be done:

- [ ] based on the Firefox version (which one?), we would need to use a different archive name and download it.
- [ ] we would need to find a dependency that can extract xz archives in Node.

OrKoN:
Closes https://github.com/puppeteer/puppeteer/issues/13390

### DIFF
--- a/docs/guides/system-requirements.md
+++ b/docs/guides/system-requirements.md
@@ -18,3 +18,4 @@
 - Firefox browser system requirements:
 
   - https://www.mozilla.org/en-US/firefox/system-requirements/
+  - on Linux, the `xz` utility is required to unpack Firefox versions 135+.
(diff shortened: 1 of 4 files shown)


MaiconWagnerTR:
I Need this fix :)

sadym-chromium (approved):

PR #13424, opened 2024-12-19
test: remove firefox-cdp test suite

OrKoN:
This PR removes tests for the Firefox CDP implementation but not the CDP support with Firefox.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -239,7 +239,6 @@ jobs:
         linux-only:
           - ${{ github.event_name == 'pull_request' && !contains(github.head_ref, 'release-please') && !contains(github.event.pull_request.labels.*.name, 'full-ci') }}
         suite:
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -239,7 +239,6 @@ jobs:
         linux-only:
           - ${{ github.event_name == 'pull_request' && !contains(github.head_ref, 'release-please') && !contains(github.event.pull_request.labels.*.name, 'full-ci') }}
         suite:
-          - firefox-cdp
```
whimboo, line 242: While the CDP code is still around and users can potentially still use it, should it be removed for all platforms or shall linux remain until all the CDP code for Firefox gets removed?

OrKoN: I will remove it right after this.

whimboo (approved):

PR #13427, opened 2024-12-19
refactor!: remove support for Firefox over CDP

OrKoN:
This PR removes dedicated support for Firefox over CDP which was deprecated in https://github.com/puppeteer/puppeteer/pull/12349. The support for WebDriver BiDi with Firefox which is used by default remains unchanged.


sadym-chromium (approved):

OrKoN:
cc @whimboo

whimboo (commented):
Looks great to see the removal of CDP for Firefox! Does it need any further testing? What if a user still has `protocol: 'cdp'` in the config file? I assume it prints an error that it is an invalid setting?

When will this be shipped? May it be in the next version you ship already?

OrKoN:
> Looks great to see the removal of CDP for Firefox! Does it need any further testing? What if a user still has `protocol: 'cdp'` in the config file? I assume it prints an error that it is an invalid setting?
> 
> When will this be shipped? May it be in the next version you ship already?

it will ship in the next release (v24.0.0). protocol is not part of the configuration file and can only be provided to launch/connect calls. It will throw an error if the protocol is still 'cdp' with the browser set to 'firefox'.

Issue #13164, opened 2024-10-07
[Feature]: Ability to Save and Share Cache in `createBrowserContext`.


sriraj66:
### Feature description

Currently, when using `createBrowserContext`, Puppeteer does not provide built-in functionality to save and share the cache between different browser contexts. This feature would be helpful in scenarios where we need to persist cache across sessions or share it with other browser contexts for improved performance and efficiency.

Proposed Solution: Introduce methods to:

Save the cache (HTTP resources, scripts, images, etc.) of a browser context to a file or storage.
Load the saved cache into a new or existing BrowserContext.
Use Case: This feature would be beneficial for applications where:

Browser contexts are used in multiple sessions but need the same cached resources.
Reducing network calls and speeding up tests or web scraping by reusing cached resources between contexts.

PR #13440, opened 2024-12-24
feat: ability to save and share cache in createBrowserContext

ARYPROGRAMMER:

Adds Cache saving, loading and sharing feature to BrowserContext of pupeteer core

No needed

**Summary**

Resolves Issue #13164 

No

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/api/BrowserContext.ts
+++ b/packages/puppeteer-core/src/api/BrowserContext.ts
@@ -1,5 +1,11 @@
 /**
  * @license
+ * Copyright 2024 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+/* eslint-disable tsdoc/syntax */
+/**
+ * @license
  * Copyright 2017 Google Inc.
  * SPDX-License-Identifier: Apache-2.0
  */
@@ -10,6 +16,7 @@ import {
   merge,
   raceWith,
 } from '../../third_party/rxjs/rxjs.js';
+import {BrowserContextCacheManager} from '../common/BrowserContextCache.js';
 import type {Cookie, CookieData} from '../common/Cookie.js';
 import {EventEmitter, type EventType} from '../common/EventEmitter.js';
 import {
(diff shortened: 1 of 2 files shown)


OrKoN (changes requested):
Thanks for the PR but I think this implementation does not align with the project goals:

1) this functionality can be implemented client-side requiring no extra capabilities from Puppeteer. We could include it into Puppeteer if it used some new capabilities on the browser side.
2) it does seem not address the feature request that is about re-using browser caches between contexts and the PR does not operate on the browser cache and saves arbitrary data from local storage (what about shared storage, cookies, other kinds of storage?) and some network data that is then restored using interception. I think this approach might work for some specific use cases but it is not generic enough to be included into Puppeteer. 
3) The implementation relies on the pages being opened and navigated to correct URLs to restore the cache that makes it not so practical.

Perhaps you could first document the design proposal on the issue? But in general, I think we will want to leave the implementation for such functionality to clients for now unless there is a built-in functionality in the browser to deal with caches.

ARYPROGRAMMER:
> Thanks for the PR but I think this implementation does not align with the project goals:
> 
> 1. this functionality can be implemented client-side requiring no extra capabilities from Puppeteer. We could include it into Puppeteer if it used some new capabilities on the browser side.
> 2. it does seem not address the feature request that is about re-using browser caches between contexts and the PR does not operate on the browser cache and saves arbitrary data from local storage (what about shared storage, cookies, other kinds of storage?) and some network data that is then restored using interception. I think this approach might work for some specific use cases but it is not generic enough to be included into Puppeteer.
> 3. The implementation relies on the pages being opened and navigated to correct URLs to restore the cache that makes it not so practical.
> 
> Perhaps you could first document the design proposal on the issue? But in general, I think we will want to leave the implementation for such functionality to clients for now unless there is a built-in functionality in the browser to deal with caches.

Thanks for the review, I'll make a proposal for the implementation. It is better to have this implemented ourselves as client managing cache as well is rather cumbersome for them. There should be an option to disable it for clients but rest I think having this implemented would save lots of time. Would make a proposal soon alongwith necessary changes in PR following it

Lightning00Blade:
I will close this PR for now. If you want you can open an issue or continue the discussion on the already open one.

PR #13449, opened 2025-01-01
Update LICENSE, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛


OrKoN:
Thanks, but I do not think it needs to be bumped after the initial file creation.

PR #13466, opened 2025-01-08
feat: resolve relative paths to absolute before unpacking

alexgwolff:

feat: resolve relative paths to absolute in CLI arguments

If update CLI description
**Summary**

**Other information**

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -142,7 +142,7 @@ export class CLI {
     }
     yargs.option('path', {
       type: 'string',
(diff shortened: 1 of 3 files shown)


OrKoN (changes requested):
Thanks for the PR. Is this addressing the error `Error: Target directory is expected to be absolute` when unpacking the archive (thrown by the extract-zip package)? I think it would better to resolve the path when calling the archive functions. I believe this part already works with relative paths (also see the failing tests).

alexgwolff:
Hi, yes this fix  `Error: Target directory is expected to be absolute`

OrKoN (approved):
Thanks!

PR #13481, opened 2025-01-09
docs: create a merged changelog

OrKoN:
Create a combine changelog for puppeteer + puppeteer-core for each release as changed in puppeteer-core are often very relevant for puppeteer users as well.

### DIFF
--- a/tools/merge-changelogs.ts
+++ b/tools/merge-changelogs.ts
@@ -1,0 +1,109 @@
+/**
+ * @license
+ * Copyright 2025 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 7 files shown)


tools/merge-changelogs.ts
```diff
@@ -0,0 +1,107 @@
+/**
+ * @license
+ * Copyright 2023 Google Inc.
```
sadym-chromium: is the file copies from somewhere?

OrKoN: Nope, I am living in 2023.

sadym-chromium (approved):

PR #13488, opened 2025-01-10
ci: fix publish job

OrKoN:
Build requires Node v22.


OrKoN:
To-be-reviewed: cc @sadym-chromium 

I will land this to unblock the release.

PR #13496, opened 2025-01-11
documentation - adding double quotes around the suggested path on Windows

tony-rivas:

Update documentation

No, not necessary

No

**Summary**

If you run the proposed command as is and there is a space in the file path, the command will fail. Adding double quotes prevents the command from failing. The double quotes still work regardless if there is or is not a space in the file path so it should just be the default suggestion.

https://pptr.dev/troubleshooting#chrome-reports-sandbox-errors-on-windows
image

image

image

No

**Other information**

N/A

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -87,7 +87,7 @@ following errors in the browser output:
 You can use icacls to set permissions manually:
 
 ```powershell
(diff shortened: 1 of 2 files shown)


tony-rivas:
Apologies, I don't have a google account to use, I'll get back to this in a couple of days

PR #13516, opened 2025-01-16
fix: use correct paper format units

### LINKED ISSUE #13515: [Bug]: PDF page format is seemingly defined in centimetres

```TypeScript

async function main() {
	const browser = await puppeteer.launch({browser: 'firefox'})
	await page.goto('https://news.ycombinator.com', {
		waitUntil: 'networkidle2',
	})
	// Saves the PDF to hn.pdf.
	await page.pdf({path:'hn.pdf', format: 'a4', printBackground: true})
	// await page.pdf({path:'hn.pdf', format: 'legal', printBackground: true})

	await browser.close()
}

main().catch(e => console.error(e))
```

### Background

I want to define the PDF page format to A4, however, the output was quite small. Then, I noticed the similarity between the numbers in millimetres and inches: the [`PageFormat`](https://pptr.dev/api/puppeteer.paperformat#remarks) seems to be defined in centimetres, however, with values (numbers) from sizes in inches!

| Format | Inches             | Centimetres      | Actual size in cm |
|:-------|:-------------------|:-----------------|-------------------|
| A4     | 8.2677" × 11.6929" | 21 × 29.7 cm     | 8.29 × 11.71 cm   |
| Legal  | 8.5" × 14"         | 21.59 × 25.56 cm | 8.5 × 14.01 cm    |

Note the similarity between the values in inches and actual size in centimeters.

Below are screenshots from Firefox PDF viewer from where I got the actual sizes (note the sizes in the screenshots are in millimetres).

A4

Legal

### Expectation

PDF page format should be correctly defined as per the documentation.

### Reality

PAGE page format size is in different units from the size number.

```TypeScript
I installed Firefox using `npx puppeteer browsers install firefox`, otherwise no configuration file used.
```

### Puppeteer version

24.1.0

### Node version

22.13.0

### Package manager

npm

### Package manager version

10.9.2

### Operating system

Linux

tukusejssirs:
When format is set, we didn't convert the internal values (in inches) to the expected measuring system, for Firefox and WebDriver BiDi the expected unit is centimeters, but we were providing the value in inches.

Now, all supported paper format have defined sizes in both centimeters and inches and when PDF options are parsed, we will returned size of the selected format in provided unit.

Fix #13515

### DIFF
--- a/packages/puppeteer-core/src/common/PDFOptions.ts
+++ b/packages/puppeteer-core/src/common/PDFOptions.ts
@@ -37,27 +37,27 @@ export type LowerCasePaperFormat =
  *
  * The sizes of each format are as follows:
  *
- * - `Letter`: 8.5in x 11in
+ * - `Letter`: 8.5in x 11in / 21.59cm x 27.94cm
  *
- * - `Legal`: 8.5in x 14in
+ * - `Legal`: 8.5in x 14in / 21.59cm x 35.56cm
  *
- * - `Tabloid`: 11in x 17in
+ * - `Tabloid`: 11in x 17in / 27.94cm x 43.18cm
  *
- * - `Ledger`: 17in x 11in
+ * - `Ledger`: 17in x 11in / 43.18cm x 27.94cm
  *
- * - `A0`: 33.1102in x 46.811in
+ * - `A0`: 33.1102in x 46.811in / 84.1cm x 118.9cm
  *
- * - `A1`: 23.3858in x 33.1102in
+ * - `A1`: 23.3858in x 33.1102in / 59.4cm x 84.1cm
  *
- * - `A2`: 16.5354in x 23.3858in
+ * - `A2`: 16.5354in x 23.3858in / 42cm x 59.4cm
  *
- * - `A3`: 11.6929in x 16.5354in
+ * - `A3`: 11.6929in x 16.5354in / 29.7cm x 42cm
  *
- * - `A4`: 8.2677in x 11.6929in
+ * - `A4`: 8.2677in x 11.6929in / 21cm x 29.7cm
  *
- * - `A5`: 5.8268in x 8.2677in
+ * - `A5`: 5.8268in x 8.2677in / 14.8cm x 21cm
  *
- * - `A6`: 4.1339in x 5.8268in
+ * - `A6`: 4.1339in x 5.8268in / 10.5cm x 14.8cm
  *
  * @public
  */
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/common/PDFOptions.ts
```diff
@@ -219,18 +219,56 @@ export type ParsedPDFOptions = Required<
...
+    cm: {width: 29.7, height: 21},
+    in: {width: 8.27, height: 11.7},
+  },
+  a5: {
+    cm: {width: 21, height: 14.8},
+    in: {width: 8.2677, height: 5.8268},
+  },
+  a6: {
+    cm: {width: 14.8, height: 10.5},
+    in: {width: 5.8268, height: 4.1339},
```
Lightning00Blade: Some of these values seemed flipped, other seem to be changed for the `in`.
Can you please revert all changes to the `in` parts, I believe all of them should be correct.
The width should always (except at `ledger`) be the smaller of the two, and people can use the `landscape` prop if they want another view.

Ref for the `in` values - https://github.com/puppeteer/puppeteer/pull/13087

tukusejssirs: Right, my bad.

Do you want me to update my previous commit or create a fix-up commit or a separate `fix` commit?

Lightning00Blade: It does not matter, we `squash merge` our PRs, so however you prefer.

Lightning00Blade (commented):
LGTM, let's update the `type PaperFormat` description to include the values in `cm` as well so the public API and website also have reference to the values.
Something like this should be good enough:
```ts
/*
 * - `Letter`: 8.5in x 11in / 21.59cm x 27.94cm
 * /
```

tukusejssirs:
> LGTM, let's update the `type PaperFormat` description to include the values in `cm` as well so the public API and website also have reference to the values.

Actually, there are four files where the same thing is defined:

1. [`docs/api/index.md`](https://github.com/hangyeol0531/puppeteer/blob/237cb42b34fca582a69386a610e185159564a43f/docs/api/index.md);
2. [`docs/api/puppeteer.paperformat.md`](https://github.com/hangyeol0531/puppeteer/blob/237cb42b34fca582a69386a610e185159564a43f/docs/api/puppeteer.paperformat.md);
3. [`website/versioned_docs/version-24.1.0/api/index.md`](https://github.com/hangyeol0531/puppeteer/blob/237cb42b34fca582a69386a610e185159564a43f/website/versioned_docs/version-24.1.0/api/index.md);
4. [`website/versioned_docs/version-24.1.0/api/puppeteer.paperformat.md`](https://github.com/hangyeol0531/puppeteer/blob/237cb42b34fca582a69386a610e185159564a43f/website/versioned_docs/version-24.1.0/api/puppeteer.paperformat.md).

Should I update all of them? Or are some of them generated?

Lightning00Blade:
Yes forgot to mention after updating the comments in the `.ts` files run `npm run docs` and it will generate all the other files.

tukusejssirs:
IMHO it would be much better to generate the list for the docs using something like this:

```js
Object.keys(paperFormats)
  .map(i =>
    ` * - \`${i.charAt(0).toUpperCase() + i.slice(1)}\`: ${paperFormats[i].in.width}in x ${paperFormats[i].in.height}in / ${paperFormats[i].cm.width}cm x ${paperFormats[i].cm.height}cm\n *`
  )
  .join('\n')
```

The actual format would depend on the docs type (the code above returns the sizes in TSDoc/JSDoc format).

However, you know better the tools you use, therefore YMMV. :man_shrugging:

Lightning00Blade (approved):
Thanks for the PR.

As for the suggestion this values should never change as they are part of a standard, so not worth the effort to try to automate this specific part IMO.

PR #13530, opened 2025-01-20
chore: use `goog:` prefix for BiDi+ commands

sadym-chromium:
Addressing https://github.com/GoogleChromeLabs/chromium-bidi/issues/2841

### DIFF
--- a/packages/puppeteer-core/src/bidi/Connection.ts
+++ b/packages/puppeteer-core/src/bidi/Connection.ts
@@ -15,7 +15,11 @@ import {debugError} from '../common/util.js';
 import {assert} from '../util/assert.js';
 
 import {BidiCdpSession} from './CDPSession.js';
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/bidi/Connection.ts
```diff
@@ -141,9 +141,11 @@ export class BidiConnection
           return;
         case 'event':
           if (isCdpEvent(object)) {
+            // Remove `goog:` prefix from BiDi+ events.
+            const eventName = object.params.event.replaceAll('goog:', '');
```
Lightning00Blade: I don't think this line is needed. The event should be the raw CDP name.
The prefixed with `goog:cdp.` is on the method object.

sadym-chromium: good point! Removed

Lightning00Blade (approved):
LGTM

PR #13542, opened 2025-01-22
fix: dispose the isolated handle

### LINKED ISSUE #13537: [Bug]: Clicked node is retained in memory

```TypeScript

await page.setContent(`<button>Click me!</button>`)

const button = await page.waitForSelector('button');

if (process.argv.includes('good')) {
  // Using the standard JS click will not retain the node in memory
  await button.evaluate(el => el.click());
} else {
  // Using the puppeteer `click()` will retain the node in memory
  await button.click();
}

// Attempt to remove the node from memory
await button.evaluate(async el => {
  el.remove();
  await new Promise(requestAnimationFrame);
});
await button.dispose();

const session = await page.createCDPSession();
const {detachedNodes} = await session.send('DOM.getDetachedDomNodes');
console.log(detachedNodes);

await page.close();
await browser.close();

if (detachedNodes.length) {
  console.log('BAD: Detached nodes found!');
  process.exit(1);
} else {
  console.log('GOOD: No detached nodes found!');
}
```

### Background

Trying to write some DevTools frontend tests that detect a memory leak in a similar scenario (clicking a button and then removing it).

I think puppeteer is getting in the way somehow by forcing the node to be retained after it is interacted with.

### Expectation

I expect there to be no nodes retained in memory by the end of the script.

### Reality

The button is retained in memory, but only if it uses the puppeteer `elementHandle.click()` and not the standard JS `element.click()` in the page context.

```TypeScript

```

### Puppeteer version

24.1.0

### Node version

18.20.5

### Package manager

yarn

### Package manager version

1.22.22

### Operating system

macOS

Lightning00Blade:
Closes #13537

### DIFF
--- a/packages/puppeteer-core/src/api/ElementHandle.ts
+++ b/packages/puppeteer-core/src/api/ElementHandle.ts
@@ -332,10 +332,7 @@ export abstract class ElementHandle<
    * @internal
    */
   override async dispose(): Promise<void> {
-    await Promise.all([
-      this.handle.dispose(),
-      this.isolatedHandle?.dispose() ?? Promise.resolve(),
-    ]);
+    await Promise.all([this.handle.dispose(), this.isolatedHandle?.dispose()]);
   }
 
   /**
--- a/packages/puppeteer-core/src/api/ElementHandle.ts
+++ b/packages/puppeteer-core/src/api/ElementHandle.ts
@@ -331,8 +331,11 @@ export abstract class ElementHandle<
   /**
    * @internal
    */
-  override dispose(): Promise<void> {
-    return this.handle.dispose();
+  override async dispose(): Promise<void> {
+    await Promise.all([
+      this.handle.dispose(),
+      this.isolatedHandle?.dispose() ?? Promise.resolve(),
+    ]);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -763,6 +766,7 @@ export abstract class ElementHandle<
     this: ElementHandle<Element>,
     options: Readonly<ClickOptions> = {},
   ): Promise<void> {
+    console.log('clicking');
```
OrKoN: remove  logging.

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -331,8 +331,11 @@ export abstract class ElementHandle<
   /**
    * @internal
    */
-  override dispose(): Promise<void> {
-    return this.handle.dispose();
+  override async dispose(): Promise<void> {
+    await Promise.all([
+      this.handle.dispose(),
+      this.isolatedHandle?.dispose() ?? Promise.resolve(),
```
OrKoN: I think we do not need Promise.resolve()?

OrKoN: `Promise.all([undefined])` should work as well?

OrKoN (approved):

Issue #13552, opened 2025-01-25
[Feature]: Reducing dependencies


ijisol:
### Feature description

In the `@puppeteer/browsers` package, there are some dependencies for unpacking archive files: `extract-zip` for `.zip`, `tar-fs` for `.tar`, and `unbzip2-stream` for `.bz2`.

https://github.com/puppeteer/puppeteer/blob/bbf75db85352e245bfc1ee59b5ef06b1eb3cf18b/packages/browsers/src/fileUtil.ts#L70-L71

However, when uncompressing a '.xz' file, it seems to call the OS's native command instead of relying on node_modules dependencies.

If so, wouldn't it be possible to do the same in other formats?

I have a concern about the licensing of the `unbzip2-stream` package. https://github.com/regular/unbzip2-stream/issues/41

And from the perspective of someone who only uses `puppeteer-core` to manipulate Chrome that is already installed, the dependencies on unpacking is useless, so the fewer dependencies the better.

OrKoN:
bz2 format is only needed for some Firefox downloads so we could drop it after Firefox 135 is stable for some time. Firefox 135+ versions are published as a tar.xz archives. We generally prefer to use JavaScript modules to extract files to avoid cross-platform issues. Unfortunately, for xz there seems to be no good implementation as a node module so we decided to call the xz utility directly since it is only used for Linux builds of Firefox (while it is usually available on Linux systems by default that might not the case for Windows/MacOS). So if there is way to extract some of the formats in a cross-platform way so that no extra dependencies have to be installed on the host, we could drop dependencies for other formats but I do not think it would be feasible.

I think it would be possible to make these dependencies optional for puppeteer-core but it would require splitting the `@puppeteer/browsers` package into two

ijisol:
> I think it would be possible to make these dependencies optional for puppeteer-core but it would require splitting the @puppeteer/browsers package into two.

I think this is better solution.

OrKoN:
So after checking out the state of the utilities cross-platform: I think we can use xz and bzip2 on linux for Firefox downloads and drop unbzip2-stream. We cannot switch to using `tar` as `tar` for Windows seems to be incompatible with other tar implementations in that it is not seem to be possible to pipe the input to it even if you install xz and bzip2 on Windows (which seems to be available on GitHub Actions Windows). 

`extract-zip` could be replaced with calling yauzl directly.

[...]

PR #13572, opened 2025-02-01
refactor: use bzip2 utility for extracting Firefox archives

OrKoN:
Newer Firefox archives use the tar.xz format for which we rely on the xz utility anyway. With this PR we will also use bzip2 utility for older archives and drop the corresponding node module.

Related https://github.com/puppeteer/puppeteer/issues/13552


packages/browsers/src/fileUtil.ts
```diff
@@ -105,19 +92,60 @@ function createXzStream() {
...
+    const unpack = spawn(
+      internalConstantsForTesting[decompressUtilityName],
+      ['-d'],
+      {
+        stdio: ['pipe', 'pipe', 'inherit'],
+      },
+    )
+      .once('error', handleError(decompressUtilityName))
+      .once('exit', code => {
+        debugFileUtil(`${decompressUtilityName} exited, code=${code}`);
```
Lightning00Blade, line 142: Why don't we use handleError here?

packages/browsers/README.md
```diff
@@ -2,6 +2,13 @@
 
 Manage and launch browsers/drivers from a CLI or programmatically.
 
+## System requirements
+
+- Node: Latest LTS or newer versions.
```
Lightning00Blade: This line sound wrong, we currently support v18 as it still has Security updates.
Better reference the engines part in the package.json

packages/browsers/src/fileUtil.ts
OrKoN: There is no error provided to the exit event.

packages/browsers/README.md
OrKoN: Updated.

Lightning00Blade (approved):

PR #13591, opened 2025-02-05
fix: don't wait for activation if the connection is disconnected

Lightning00Blade:
Whenever the connection disconnects the `CDPSessionEvent.Disconnected`  on the sessions as well.
This triggers the waiting for frame activation but we know with no connection we won't receive it.

Should save 100ms when ever exiting Node Process after `browser.close()`;


OrKoN:
This seems to result in some persistent test failures (not sure why).

Lightning00Blade:
Those test looked like once that have flaked before, maybe this fixes the flakiness but not the issue? I will investigate.

test/TestExpectations.json
```diff
@@ -1138,6 +1131,13 @@
     "expectations": ["FAIL"],
     "comment": "Requires support for multiple sessions"
   },
+  {
+    "testIdPattern": "[launcher.spec] Launcher specs Puppeteer Browser.disconnect should reject waitForSelector when browser closes",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox", "webDriverBiDi"],
```
Lightning00Blade, line 1137: Test passes for Chrome BiDi

OrKoN (approved):

packages/puppeteer-core/src/cdp/CdpSession.ts
```diff
@@ -140,22 +144,23 @@ export class CdpCDPSession extends CDPSession {
...
+    this.#detached = true;
   }
 
   /**
    * @internal
    */
-  _onClosed(): void {
+  onClosed(): void {
     this.#callbacks.clear();
-    this.#connection = undefined;
```
alexey-pelykh, line 158: What was the rationale to introduce behavioral breaking change of `connection()` not returning `undefined` anymore after `_onClosed()` and at the same time make `get #closed()` private so that the state can not be accessed from the outside?

alexey-pelykh: I firmly believe this PR should've included also change alike https://github.com/puppeteer/puppeteer/pull/13613

Lightning00Blade: It was not intentional, but the there is no contract that enforces this behavior. 
If you were using this this was like using any other heuristic. 
As for why the close getter is private this is a private class so even if was not it would not be part of the public API.

alexey-pelykh: There's indeed no contract that enforces this behavior, however the `CDPSession` that `CdpCDPSession` implements is public, and `connection()` is not marked as internal (well, no explicitly public either).

PR #13598, opened 2025-02-07
ci: use one set-up

Lightning00Blade:
Blocked by - https://github.com/google/wireit/issues/1242


OrKoN:
@Lightning00Blade do you want to keep this PR?

Lightning00Blade:
While I think it would be nice, given that it's close we can closed it.

PR #13606, opened 2025-02-10
feat: added scroll behavior to scroll method

Archish27:
Added scroll behavior in certain case scroll behaviour is required like smooth which will helps to debug website behavior when smooth scrolling is happening.

Yes

Yes

**Summary**

This will be useful in Puppeteer Replay, where the recorder's scroll behavior can be modified to enable smooth scrolling, allowing us to replay steps with a smoother scrolling effect.

No

**Other information**

### DIFF
--- a/docs/guides/page-interactions.md
+++ b/docs/guides/page-interactions.md
@@ -91,6 +91,7 @@ The [`.scroll()`] functions uses mouse wheel events to scroll an element.
 await page.locator('div').scroll({
   scrollLeft: 10,
   scrollTop: 20,
+  scrollBehavior: 'scroll'
 });
 ```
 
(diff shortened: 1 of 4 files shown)


OrKoN:
The smooth scroll is usually bad for automation since it results in extra flakiness since this kind of scrolling happens asynchronously on a different thread in the browser. Could you please describe the use case for why you need smooth scrolling in more details? Note that there are currently no plans to enable additional scroll functionality in Recorder so perhaps the initial step would be to file a feature request for DevTools Recorder at https://crbug.com/new

Lightning00Blade:
Given the comment above, and the issue that it may cause I will close the PR for now.
If you open a Feature request to Recorder feel free to link it here so people can star it.

Issue #13037, opened 2024-09-03
[Feature]: support custom arguments in the launch command


koenhandekyn:
### Steps to reproduce

```
npx @puppeteer/browsers launch chrome-headless-shell@stable -- --remote-debugging-port=9222  --headless --no-sandbox
```

this always exits immediately with exit code 130 (break?) while it works for normal chrome's 

[testing on mac arm]

### Expected results

headless chrome should run with remote debugging active 

### Actual results

the process exits immediately

koenhandekyn:
ok then it's indeed a feature request :)

i ended up writing a shell script that parses the installation path to launch the headless chrome 

for the seekers:

```
#!/bin/bash
export CHROME_PATH=$(npx --yes @puppeteer/browsers install chrome-headless-shell@stable | awk '{print $2}')
$CHROME_PATH --disable-gpu --remote-debugging-port=9222 --disable-software-rasterizer --disable-bluetooth --no-sandbox
```

koenhandekyn:
design tip: chatgpt was clearly hallucinating that params that need to be passed on need to appear after an extra --, which actually might make sense from design perspective :)

jonhermansen:
I took a pass at implementing this, but I couldn't find a simple way to work with yargs to extract the additional arguments. So in my patch, I'm relying on the positional nature of "--" to find the custom arguments. The only downside is that I'm using `process.argv` instead of yargs, so I'm not sure if this will be accepted.

PR #13612, opened 2025-02-14
feat: pass custom arguments to browser on launch

jonhermansen:

Feature #13037

No

I don't think it's relevant, I don't see any place where the command-line tools are documented. Only the Javascript APIs seem to be documented in docs/

**Summary**

Allow passing additional arguments directly to the browser on launch. This feature was requested in #13037 

No

**Other information**

I tried to work with yargs to pull the additional arguments out, and in fact, yargs does support a mechanism of passing "extra" arguments, i.e. anything after "--". However, the TS types did not allow me to access yargs' `argv._` and even when I do access it (by casting yargs' `argv` to any), I've lost reference to the position of "--" and so I'd need to process the arguments based on their position, which is very confusing.

Instead I've taken the approach of looking at `process.argv` directly instead. This way, it is very clear which arguments are handled by yargs, and which arguments are handled by the browser process itself.

If there's another approach that would be more straightforward using yargs, or there is a desire to *not* use `process.argv` directly, I can take a look at an alternative solution.

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -373,8 +373,15 @@ export class CLI {
                 cacheDir: args.path ?? this.#cachePath,
                 platform: args.platform,
               });
+          // Pass any extra arguments after '--' directly to the browser executable
+          let extraArgs: string[] = [];
+          const extraArgsIndex = process.argv.indexOf('--');
+          if (extraArgsIndex !== -1) {
+            extraArgs = process.argv.splice(extraArgsIndex + 1);
+          }
           launch({
             executablePath,
+            args: extraArgs,
             detached: args.detached,
           });
         },


OrKoN:
Thanks for the PR. Could you please add a few tests? I'd suggest starting an new test file called `launch.spec.ts` next to packages/browsers/test/src/list.spec.ts and creating a test similar to the 'should list installed browsers' test. Instead of using the list command, we could launch Chrome the `--` and `--version` and make sure that Chrome prints the version.

Lightning00Blade:
You can provide the parser configuration like 
```

jonhermansen:
> ```ts
> const extraArgs = (
>             argv as unknown as Record<'--', Array<number | string>>
>           )?.['--'].filter(arg => {
>             return typeof arg === 'string';
>           });
> ```

Thanks, I've incorporated this change.

jonhermansen:
> Thanks for the PR. Could you please add a few tests? I'd suggest starting an new test file called `launch.spec.ts` next to packages/browsers/test/src/list.spec.ts and creating a test similar to the 'should list installed browsers' test. Instead of using the list command, we could launch Chrome the `--` and `--version` and make sure that Chrome prints the version.

Thanks for the pointer to the existing tests. Yes, I will create a new spec file and add tests there.

PR #13613, opened 2025-02-15
fix: CdpCDPSession.connection() to return undefined after onClosed

alexey-pelykh:

Fix

~The tests should not be impacted.~
Yes

N/A

**Summary**

Restores `CdpCDPSession.connection()` behavior to return `undefined` after `onClosed` as it was before #13591

It fix-reverts one

### DIFF
--- a/packages/puppeteer-core/src/cdp/CdpSession.ts
+++ b/packages/puppeteer-core/src/cdp/CdpSession.ts
@@ -71,6 +71,16 @@ export class CdpCDPSession extends CDPSession {
   }
 
   override connection(): Connection | undefined {
+    if (this.#closed) {
+      return undefined;
+    }
+    return this.#connection;
+  }
+
+  /**
+   * @internal
+   */
+  get _connection(): Connection {
     return this.#connection;
   }
 
--- a/packages/puppeteer-core/src/cdp/FrameManager.ts
+++ b/packages/puppeteer-core/src/cdp/FrameManager.ts
@@ -45,7 +45,7 @@ export class FrameManager extends EventEmitter<FrameManagerEvents> {
   #networkManager: NetworkManager;
   #timeoutSettings: TimeoutSettings;
   #isolatedWorlds = new Set<string>();
-  #client: CDPSession;
+  #client: CdpCDPSession;
   #scriptsToEvaluateOnNewDocument = new Map<string, CdpPreloadScript>();
   #bindings = new Set<Binding>();
 
@@ -73,7 +73,7 @@ export class FrameManager extends EventEmitter<FrameManagerEvents> {
     return this.#networkManager;
   }
 
-  get client(): CDPSession {
+  get client(): CdpCDPSession {
     return this.#client;
   }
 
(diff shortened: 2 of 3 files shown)


OrKoN:
Thanks for the PR.

> Did you add tests for your changes?

> The tests should not be impacted.

Could you please add a test to `test/src/cdp/CDPSession.spec.ts`?:

```
  it('should not return a connection if the session is disposed', async () => {
    const client = await page.createCDPSession();
    expect(client.connection()).toBeTruthy();
    await client.detach();
    expect(client.connection()).toBeUndefined();
    expect(Connection.fromSession(client)).toBeUndefined();
  });
```

alexey-pelykh:
@OrKoN done

Lightning00Blade:
I believe this change reintroduces the leaking SetTimeout that was fix with the change - #13591.

alexey-pelykh:
@Lightning00Blade it may be so - however the original fix relied on exposing connection after detaching?

Lightning00Blade:
Yes as we need to know if the whole browser is disconnecting or not - 
https://github.com/puppeteer/puppeteer/commit/6b20ac10b1cc6a54a87eb2f6437727d204acd1c3#diff-8ce6f7b6f62b4b2ea843fb6ed83de580cc083f4005adbc5c1671fbfe73792c98R107
I think the change made here is reasonable, as the behavior did change unintentionally.
Though I would am still interested in understanding the use case, as user is expected to call `.detach()` and and at that point they should not use the client. 

As for how to fix, I think introducing a internal `isConnectionClosed` and get that in the line above to use that.
Happy to create a PR after this as well.

alexey-pelykh:
@Lightning00Blade thanks for providing the context - very helpful. I guess given than `connection()` is not marked internal (or explicitly public for that matter) - indeed it would be better to define the contract how it behaves.

The use-case (if I got you right): relying on `connection()` becoming `undefined` as an indicator that session is detached when it's not the callee who initiated the detachment.

alexey-pelykh:
I guess the route would be either #13615 or internal `isConnectionClosed` as suggested @Lightning00Blade

Lightning00Blade:
So the function I linked above handles when an internal CDPSessions gets detached, so at this point the `CDPSession.#closed` will be `true`, but the connection may be in either of the states, and we are interested if the connection is closed or not.

This is is roughly what happens:
Session 1 gets detached
Then there are two paths -> A) a new session will get attached (Event is emitted).
                                              -> B) the page is closed, no new session is attached.
So we set a check for an event and race it with a setTimeout(). 

But if the underlying connection is disconnected we know only B is possible, so we can jump to the code path for B.

I hope it makes sense.

alexey-pelykh:
Thanks for the explanation! IMHO, then the `connection()` indeed should always return the underlying connection however a detached state should also be exposed as done in #13615

alexey-pelykh:
At the same time, should there be a maintainer decision to have `connection()` behaving like before, I've updated this PR as well.

alexey-pelykh:
Superseded by #13615

PR #13615, opened 2025-02-15
feat: expose CDPSession.detached

alexey-pelykh:
Different perspective to #13613

Feature/Refactor: `CdpCDPSession.connection()` behavior contracted by test; add `CDPSession.detached`

Yes

Code comments only.

**Summary**

#13591 and #13613 raised a question that `CdpCDPSession.connection()` behavior was not explicitly defined.

No

### DIFF
--- a/packages/puppeteer-core/src/cdp/CdpSession.ts
+++ b/packages/puppeteer-core/src/cdp/CdpSession.ts
@@ -74,7 +74,7 @@ export class CdpCDPSession extends CDPSession {
     return this.#connection;
   }
 
-  get #closed(): boolean {
+  override get detached(): boolean {
     return this.#connection._closed || this.#detached;
   }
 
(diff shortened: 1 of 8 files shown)


packages/puppeteer-core/src/cdp/CdpSession.ts
```diff
@@ -74,6 +74,10 @@ export class CdpCDPSession extends CDPSession {
     return this.#connection;
   }
 
+  override get detached(): boolean {
```
Lightning00Blade, line 77: Nit: you can remove the private  `#closed` getter as getter value can't be modified, so no need to have two thing that are the same.

alexey-pelykh: Done, thanks for the suggestion!

Lightning00Blade (approved):
LGTM, thanks for the PR.

Lightning00Blade (approved):

PR #13618, opened 2025-02-17
refactor: use the browser getter

Lightning00Blade:


packages/puppeteer-core/src/cdp/FrameManager.ts
```diff
@@ -104,8 +104,8 @@ export class FrameManager extends EventEmitter<FrameManagerEvents> {
       return;
     }
 
-    if (this.client.connection()?._closed) {
-      // On connection disconnected remove all frames
+    if (!this.#page.browser().connected) {
+      // If the browser is isconnected don't waif for activation
```
alexey-pelykh:
```
// If the browser is not connected, don't wait for activation
```
or something

OrKoN (approved):

PR #13620, opened 2025-02-17
refactor: use stricter CDP implementation type of CDPSession

alexey-pelykh:

Tiny typing improvement.

There's no functional change

No - nothing affected

**Summary**

As a part of #13613 there were few typing changes so that CDP FrameManager uses CDP-specific session.

Nope

**Other information**

N/A

### DIFF
--- a/packages/puppeteer-core/src/cdp/Connection.ts
+++ b/packages/puppeteer-core/src/cdp/Connection.ts
@@ -83,8 +83,15 @@ export class Connection extends EventEmitter<CDPSessionEvents> {
   /**
    * @internal
    */
-  get _sessions(): Map<string, CDPSession> {
+  get _sessions(): Map<string, CdpCDPSession> {
     return this.#sessions;
+  }
+
+  /**
+   * @internal
+   */
+  _session(sessionId: string): CdpCDPSession | null {
+    return this.#sessions.get(sessionId) || null;
   }
 
   /**
@@ -92,7 +99,7 @@ export class Connection extends EventEmitter<CDPSessionEvents> {
    * @returns The current CDP session if it exists
    */
   session(sessionId: string): CDPSession | null {
-    return this.#sessions.get(sessionId) || null;
+    return this._session(sessionId);
   }
 
   url(): string {
@@ -240,7 +247,7 @@ export class Connection extends EventEmitter<CDPSessionEvents> {
   async _createSession(
     targetInfo: {targetId: string},
     isAutoAttachEmulated = true,
-  ): Promise<CDPSession> {
+  ): Promise<CdpCDPSession> {
     if (!isAutoAttachEmulated) {
       this.#manuallyAttached.add(targetInfo.targetId);
     }
--- a/packages/puppeteer-core/src/cdp/Page.ts
+++ b/packages/puppeteer-core/src/cdp/Page.ts
@@ -97,7 +97,7 @@ function convertConsoleMessageLevel(method: string): ConsoleMessageType {
  */
 export class CdpPage extends Page {
   static async _create(
-    client: CDPSession,
+    client: CdpCDPSession,
     target: CdpTarget,
     defaultViewport: Viewport | null,
   ): Promise<CdpPage> {
@@ -120,7 +120,7 @@ export class CdpPage extends Page {
   #closed = false;
   readonly #targetManager: TargetManager;
 
-  #primaryTargetClient: CDPSession;
+  #primaryTargetClient: CdpCDPSession;
   #primaryTarget: CdpTarget;
   #tabTargetClient: CDPSession;
   #tabTarget: CdpTarget;
@@ -140,7 +140,7 @@ export class CdpPage extends Page {
   #serviceWorkerBypassed = false;
   #userDragInterceptionEnabled = false;
 
-  constructor(client: CDPSession, target: CdpTarget) {
+  constructor(client: CdpCDPSession, target: CdpTarget) {
     super();
     this.#primaryTargetClient = client;
     this.#tabTargetClient = client.parentSession()!;
@@ -257,12 +257,13 @@ export class CdpPage extends Page {
   }
 
   async #onActivation(newSession: CDPSession): Promise<void> {
-    this.#primaryTargetClient = newSession;
+    // TODO: Remove assert once we have separate Event type for CdpCDPSession.
     assert(
-      this.#primaryTargetClient instanceof CdpCDPSession,
-      'CDPSession is not instance of CDPSessionImpl',
+      newSession instanceof CdpCDPSession,
+      'CDPSession is not instance of CdpCDPSession',
     );
-    this.#primaryTarget = this.#primaryTargetClient.target();
+    this.#primaryTargetClient = newSession;
+    this.#primaryTarget = newSession.target();
     assert(this.#primaryTarget, 'Missing target on swap');
     this.#keyboard.updateClient(newSession);
     this.#mouse.updateClient(newSession);
--- a/packages/puppeteer-core/src/cdp/Target.ts
+++ b/packages/puppeteer-core/src/cdp/Target.ts
@@ -8,14 +8,13 @@ import type {Protocol} from 'devtools-protocol';
 
 import type {Browser} from '../api/Browser.js';
 import type {BrowserContext} from '../api/BrowserContext.js';
-import type {CDPSession} from '../api/CDPSession.js';
 import {PageEvent, type Page} from '../api/Page.js';
 import {Target, TargetType} from '../api/Target.js';
 import {debugError} from '../common/util.js';
 import type {Viewport} from '../common/Viewport.js';
 import {Deferred} from '../util/Deferred.js';
 
-import {CdpCDPSession} from './CdpSession.js';
+import type {CdpCDPSession} from './CdpSession.js';
 import {CdpPage} from './Page.js';
 import type {TargetManager} from './TargetManager.js';
 import {CdpWebWorker} from './WebWorker.js';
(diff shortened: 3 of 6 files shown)


Lightning00Blade (changes requested):
I would prefer to not include a runtime checks when this is internal code, that users won't ever interact with, TypeScript safety should be enough for internal parts.
(I know that there was already one, but that should not be there as well).

I looked into changing the types, only and it looked like it was failing in a couple of places due to changing the constructor. And without that change other places don't benefit as much with this change.

alexey-pelykh:
@Lightning00Blade thanks for the review! Your points are perfectly valid - I've tried to avoid runtime checks and gradually replace non-CDP-specific types with CDP-specific ones and eventually got success, at least it looks like it.

packages/puppeteer-core/src/cdp/Connection.ts
```diff
@@ -58,10 +54,6 @@ export class Connection extends EventEmitter<CDPSessionEvents> {
     this.#transport.onclose = this.#onClose.bind(this);
   }
 
-  static fromSession(session: CDPSession): Connection | undefined {
```
alexey-pelykh: This wasn't used anywhere, so I've decided to drop this trivial wrapper

packages/puppeteer-core/src/cdp/Target.ts
```diff
@@ -51,11 +51,11 @@ export class CdpTarget extends Target {
...
-    session: CDPSession | undefined,
+    session: CdpCDPSession | undefined,
     browserContext: BrowserContext | undefined,
     targetManager: TargetManager | undefined,
     sessionFactory:
-      | ((isAutoAttachEmulated: boolean) => Promise<CDPSession>)
+      | ((isAutoAttachEmulated: boolean) => Promise<CdpCDPSession>)
       | undefined,
   ) {
     super();
```
Lightning00Blade, line 60: There is a `instanceof` we can remove in this constructor now.

packages/puppeteer-core/src/cdp/Target.ts
```diff
@@ -73,8 +73,8 @@ export class CdpTarget extends Target {
   override async asPage(): Promise<Page> {
     const session = this._session();
     if (!session) {
-      return await this.createCDPSession().then(client => {
-        return CdpPage._create(client, this, null);
+      return await this.createCDPSession().then(session => {
```
Lightning00Blade: Can we change the signature of this function `this.createCDPSession()` and not use type casting?

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -257,12 +257,12 @@ export class CdpPage extends Page {
   }
 
   async #onActivation(newSession: CDPSession): Promise<void> {
-    this.#primaryTargetClient = newSession;
     assert(
-      this.#primaryTargetClient instanceof CdpCDPSession,
+      newSession instanceof CdpCDPSession,
       'CDPSession is not instance of CDPSessionImpl',
```
Lightning00Blade: Nit: can we update the text here?

docs/api/puppeteer.connection.createsession.md
```diff
@@ -8,7 +8,7 @@ sidebar_label: Connection.createSession
 
 ```typescript
 class Connection {
-  createSession(targetInfo: Protocol.Target.TargetInfo): Promise<CDPSession>;
```
Lightning00Blade: We need to revert this change, as we don't expose `CdpCDPSession` publicly so there will be no way for user using Puppeteer search for it.

docs/api/puppeteer.connection.session.md
```diff
@@ -8,7 +8,7 @@ sidebar_label: Connection.session
 
 ```typescript
 class Connection {
-  session(sessionId: string): CDPSession | null;
```
Lightning00Blade: Same here: We need to revert this change, as we don't expose `CdpCDPSession` publicly so there will be no way for user using Puppeteer search for it.

Lightning00Blade (commented):
Thanks for the PR, left some comments.

packages/puppeteer-core/src/cdp/Connection.ts
Lightning00Blade: Acknowledged.

docs/api/puppeteer.connection.createsession.md
alexey-pelykh: Ah, thanks for the context - then I guess indeed changing anything isn't worth it unless we want to explore exposing CdpCDPSession 🤔

alexey-pelykh: On the other hand - I'll be back in a few :)

packages/puppeteer-core/src/cdp/Target.ts
alexey-pelykh: Done

packages/puppeteer-core/src/cdp/Target.ts
alexey-pelykh: Added internal `_createCDPSession`

packages/puppeteer-core/src/cdp/Page.ts
alexey-pelykh: Done

docs/api/puppeteer.connection.createsession.md
alexey-pelykh: Reverted

docs/api/puppeteer.connection.session.md
alexey-pelykh: Reverted

packages/puppeteer-core/src/cdp/Connection.ts
OrKoN: Note that this method was part of the public API and we should not remove it to avoid breaking further clients.

alexey-pelykh: Reverted

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -257,12 +257,12 @@ export class CdpPage extends Page {
   }
 
   async #onActivation(newSession: CDPSession): Promise<void> {
-    this.#primaryTargetClient = newSession;
     assert(
-      this.#primaryTargetClient instanceof CdpCDPSession,
-      'CDPSession is not instance of CDPSessionImpl',
+      newSession instanceof CdpCDPSession,
```
Lightning00Blade, line 262: In practice we don't need this check as the target emits a CdpCDPSession only, but that would require another refactor to events. Can we just add a comment.
```js
// TODO: Remove once we have separate Event type for CdpCDPSession.
```
So that it can be fixed separate.

alexey-pelykh: Done

packages/puppeteer-core/src/cdp/Connection.ts
```diff
@@ -83,16 +83,23 @@ export class Connection extends EventEmitter<CDPSessionEvents> {
...
    */
-  get _sessions(): Map<string, CDPSession> {
+  get _sessions(): Map<string, CdpCDPSession> {
     return this.#sessions;
   }
 
+  /**
+   * @internal
+   */
+  _session(sessionId: string): CdpCDPSession | null {
```
Lightning00Blade, line 93: This was introduced but not used, please remove.

packages/puppeteer-core/src/cdp/Target.ts
```diff
@@ -102,23 +102,30 @@ export class CdpTarget extends Target {
...
       throw new Error('sessionFactory is not initialized');
     }
     return this.#sessionFactory;
   }
 
-  override createCDPSession(): Promise<CDPSession> {
+  /**
+   * @internal
+   */
+  _createCDPSession(): Promise<CdpCDPSession> {
```
Lightning00Blade: We don't need to wrap this into a internal function, as CdpCDPSession extends CDPSession, so TypeScript won't complain.

packages/puppeteer-core/src/cdp/Target.ts
```diff
@@ -65,16 +65,16 @@ export class CdpTarget extends Target {
...
       this.#session.setTarget(this);
     }
   }
 
   override async asPage(): Promise<Page> {
     const session = this._session();
     if (!session) {
-      return await this.createCDPSession().then(client => {
-        return CdpPage._create(client, this, null);
+      return await this._createCDPSession().then(session => {
```
Lightning00Blade: Don't rename the handle in the promise it's consing to have to variable in the same scope with the same name, But its a pattern we call Session clients as well.

Lightning00Blade (commented):
Over all looks really good, thanks for the PR.
Left some comments, that need to be fixed before merging

packages/puppeteer-core/src/cdp/Connection.ts
alexey-pelykh: It is used in https://github.com/puppeteer/puppeteer/pull/13620/files#diff-b688d84d1dee4967409766c8b2061c79c314e6e9cb548683adf7d14655b07b57L321

packages/puppeteer-core/src/cdp/Target.ts
alexey-pelykh: Done

packages/puppeteer-core/src/cdp/Connection.ts
Lightning00Blade: Sorry VsCode somehow didn't find the reference.

alexey-pelykh: I think I got what you've meant, removed

packages/puppeteer-core/src/cdp/Connection.ts
```diff
@@ -83,15 +83,15 @@ export class Connection extends EventEmitter<CDPSessionEvents> {
...
-  get _sessions(): Map<string, CDPSession> {
+  get _sessions(): Map<string, CdpCDPSession> {
     return this.#sessions;
   }
 
   /**
    * @param sessionId - The session id
    * @returns The current CDP session if it exists
    */
-  session(sessionId: string): CDPSession | null {
```
Lightning00Blade, line 94: I don't think we can change this as its part of the Public API

alexey-pelykh: That's why I've added `_session` that was used ;)

alexey-pelykh: I've reverted _session removal so that `CdpCDPSession` is not exposed.

Lightning00Blade (approved):

Lightning00Blade:
Thanks for the PR.

PR #12735, opened 2024-07-11
chore: expose quality variable

minggeorgelei:
Sometimes users do not want very high quality video that may take much disk space or they may want clearer video. This PR is to expose quality variable so that users can adjust it themselves to meet the needs of testing.

PR #13645, opened 2025-03-04
feat: improve `screencast` options

danielbayley:

Adds missing options to `screencast`/`ffmpeg`, and addresses undocumented existing options.

Also fixes multiple bugs in the existing code (see `commit`s…)

Closes #12735

Yes

Yes

**Summary**

Adds missing options to `screencast`/`ffmpeg` to better control output, particularly GIFs, along with additional formats. This can avoid having to run `ffmpeg` a second time, only to tweak the final desired output—potentially drastically decreasing, for example, CI run time.

No

**Other information**

See also:

#11084

_[High quality GIF with FFmpeg](https://blog.pkh.me/p/21-high-quality-gif-with-ffmpeg.html)_

_[A guide to upscaling or downscaling video with FFmpeg](https://write.corbpie.com/a-guide-to-upscaling-or-downscaling-video-with-ffmpeg)_

_[Elements to Keep in Mind When Optimizing GIFs](https://gifyard.com/elements-for-optimizing-gifs)_

https://github.com/sindresorhus/Gifski/issues/161#issuecomment-552681700

[FFmpeg Filters Documentation](https://ffmpeg.org/ffmpeg-filters.html#palettegen)

### DIFF
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -316,15 +316,18 @@ export interface ScreenshotOptions {
 
 /**
  * @public
- * @experimental
  */
 export type FileFormat = 'gif' | 'webm' | 'mp4';
 
+/**
+ * @public
+ * @experimental
+ */
 export interface ScreencastOptions {
   /**
    * File path to save the screencast to.
    */
-  path?: `${string}.gif` | `${string}.webm` | `${string}.mp4`;
+  path?: `${string}.${FileFormat}`;
   /**
    * Specifies the output file format.
    *
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -322,7 +322,13 @@ export interface ScreencastOptions {
   /**
    * File path to save the screencast to.
    */
-  path?: `${string}.webm`;
+  path?: `${string}.gif` | `${string}.webm`;
+  /**
+   * Specifies the output file format.
+   *
+   * @defaultValue `webm`
+   */
+  format?: 'gif' | 'webm';
   /**
    * Specifies the region of the viewport to crop.
    */
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -318,13 +318,13 @@ export interface ScreenshotOptions {
  * @public
  * @experimental
  */
-export type FileFormat = 'gif' | 'webm';
+export type FileFormat = 'gif' | 'webm' | 'mp4';
 
 export interface ScreencastOptions {
   /**
    * File path to save the screencast to.
    */
-  path?: `${string}.gif` | `${string}.webm`;
+  path?: `${string}.gif` | `${string}.webm` | `${string}.mp4`;
   /**
    * Specifies the output file format.
    *
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -354,6 +354,42 @@ export interface ScreencastOptions {
    */
   speed?: number;
   /**
+   * Specifies the frame rate in frames per second.
+   *
+   * @defaultValue `30` (`20` for GIF)
+   */
+  fps?: number;
+  /**
+   * Specifies the number of times to loop playback, from `0` to `Infinity`.
+   *
+   * @defaultValue `0`
+   */
+  loop?: number;
+  /**
+   * Specifies the delay between iterations of a loop, in ms.
+   * `-1` is a special value to re-use the previous delay.
+   *
+   * @defaultValue `-1`
+   */
+  delay?: number;
+  /**
+   * Specifies the recording
+   * {@link https://trac.ffmpeg.org/wiki/Encode/VP9#constantq | quality}
+   * Constant Rate Factor between `0`–`63`. Lower values mean better quality.
+   *
+   * @defaultValue `30`
+   */
+  quality?: number;
+  /**
+   * Specifies the maximum number of
+   * {@link https://ffmpeg.org/ffmpeg-filters.html#palettegen | palette}
+   * colors to quantize, with GIF limited to `256`.
+   * Restrict the palette to only necessary colors to reduce output file size.
+   *
+   * @defaultValue `256`
+   */
+  colors?: number;
+  /**
    * Path to the {@link https://ffmpeg.org/ | ffmpeg}.
    *
    * Required if `ffmpeg` is not in your PATH.
@@ -2355,8 +2391,8 @@ export abstract class Page extends EventEmitter<PageEvents> {
    *
    * @remarks
    *
-   * All recordings will be {@link https://www.webmproject.org/ | WebM} format using
-   * the {@link https://www.webmproject.org/vp9/ | VP9} video codec. The FPS is 30.
+   * By default, all recordings will be {@link https://www.webmproject.org/ | WebM} format using
+   * the {@link https://www.webmproject.org/vp9/ | VP9} video codec, with a frame rate of 30 FPS.
    *
    * You must have {@link https://ffmpeg.org/ | ffmpeg} installed on your system.
    */
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -318,6 +318,8 @@ export interface ScreenshotOptions {
  * @public
  * @experimental
  */
+export type FileFormat = 'gif' | 'webm';
+
 export interface ScreencastOptions {
   /**
    * File path to save the screencast to.
@@ -328,7 +330,7 @@ export interface ScreencastOptions {
    *
    * @defaultValue `webm`
    */
-  format?: 'gif' | 'webm';
+  format?: FileFormat;
   /**
    * Specifies the region of the viewport to crop.
    */
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -361,8 +361,9 @@ export interface ScreencastOptions {
   fps?: number;
   /**
    * Specifies the number of times to loop playback, from `0` to `Infinity`.
+   * A value of `0` or `undefined` will disable looping.
    *
-   * @defaultValue `0`
+   * @defaultValue `undefined`
    */
   loop?: number;
   /**
--- a/packages/puppeteer-core/src/node/ScreenRecorder.ts
+++ b/packages/puppeteer-core/src/node/ScreenRecorder.ts
@@ -234,7 +234,7 @@ export class ScreenRecorder extends PassThrough {
           // input.
           [
             '-vf',
-            `fps=${fps},split[s0][s1];[s0]palettegen=stats_mode=diff:max_colors=${colors}[p];[s1][p]paletteuse`,
+            `fps=${fps},split[s0][s1];[s0]palettegen=stats_mode=diff:max_colors=${colors}[p];[s1][p]paletteuse=dither=bayer`,
           ],
           // Sets the number of times to loop playback.
           ['-loop', `${loop}`],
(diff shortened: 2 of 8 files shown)


OrKoN:
Thanks for the PR. I think proposed options make sense but I will review in details later (cc @Lightning00Blade). Could you please run `npm run docs` to re-generate markdown docs?

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -174,7 +199,14 @@ export class ScreenRecorder extends PassThrough {
     );
   }
 
-  #getFormatArgs(format: 'webm' | 'gif') {
+  #getFormatArgs(
+    format: 'webm' | 'gif' | 'mp4',
+    fps: number | 'fps=source_fps',
+    loop: number | typeof Infinity | boolean,
```
Lightning00Blade: Why do we need `typeof Infinity`, 
`const test: number = Infinity;` work in TypeScript. We should allow only `false` and not `boolean`, as having loop true does not make sense without a number

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -40,7 +40,12 @@ const debugFfmpeg = debug('puppeteer:ffmpeg');
 export interface ScreenRecorderOptions {
   speed?: number;
   crop?: BoundingBox;
-  format?: 'gif' | 'webm';
+  format?: 'gif' | 'webm' | 'mp4';
+  fps?: number;
+  loop?: number | typeof Infinity | boolean;
```
Lightning00Blade: Can we make this only a number, and it will not be looped if its undefined.

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -102,9 +128,10 @@ export class ScreenRecorder extends PassThrough {
         // VP9 tries to use all available threads?
         ['-threads', '1'],
         // Specifies the frame rate we are giving ffmpeg.
-        ['-framerate', `${DEFAULT_FPS}`],
+        ['-framerate', `${fps}`],
         // Specifies the encoding and format we are using.
-        this.#getFormatArgs(format ?? 'webm'),
+        // FIXME https://sentry.io/answers/node-js-no-overload-matches-this-call-error-with-typescript
```
Lightning00Blade: What does this `FIXME` refer to? Is it still needed?

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -183,21 +215,39 @@ export class ScreenRecorder extends PassThrough {
...
           ['-f', 'webm'],
           // Sets the quality. Lower the better.
-          ['-crf', `${CRF_VALUE}`],
+          ['-crf', `${quality}`],
           // Sets the quality and how efficient the compression will be.
           ['-deadline', 'realtime', '-cpu-used', '8'],
         ].flat();
       case 'gif':
+        fps = DEFAULT_FPS === fps ? 20 : 'fps=source_fps';
+        if (!Number.isInteger(loop) || loop === 0) loop = ~~!loop * -1;
```
Lightning00Blade: Can't we just set it to the correct default where it won't loop? Rather then this evaluation?

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -183,21 +215,39 @@ export class ScreenRecorder extends PassThrough {
...
+      case 'mp4':
+        return [
+          // Sets the codec to use.
+          ['-c:v', 'libx264'],
+          // Sets the format
+          ['-f', 'mp4'],
+          // Sets the quality. Lower the better.
+          ['-crf', `${quality}`],
+          // Sets the quality and how efficient the compression will be.
+          ['-deadline', 'realtime', '-cpu-used', '8'],
```
Lightning00Blade: What will happen if the Computer has less then 8 CPU cores?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -345,6 +345,48 @@ export interface ScreencastOptions {
...
+   * @defaultValue `webm`
+   */
+  format?: 'webm' | 'gif' | 'mp4';
+  /**
+   * Specifies the frame rate in frames per second.
+   *
+   * @defaultValue `30` (`20` for GIF)
+   */
+  fps?: number;
+  /**
```
Lightning00Blade, line 365: I think some comments are not attached to the correct value, can you please fix?

Lightning00Blade:
Thanks for the PR! I left some comments, and I want to check more of the options that we provide to ffmpeg.

packages/puppeteer-core/src/node/ScreenRecorder.ts
danielbayley: Not sure how that one slipped through my version control. There seems to be a bug in _GitKraken_ that sometimes sneakily includes extra adjacent lines…

packages/puppeteer-core/src/node/ScreenRecorder.ts
danielbayley: @Lightning00Blade Presumably, it will explode! Seems to be another oversight in the existing code. Fixed now…
~~~ js
import os from 'node:os'
// …
, '-cpu-used', `${os.cpus().length / 2}`],
~~~

packages/puppeteer-core/src/node/ScreenRecorder.ts
danielbayley: > Why do we need `typeof Infinity`

Suppose I just wanted to emphasize `Infinity` as a nice, semantic option. But simply `loop?: number | Infinity;` is a type error (see related [issue](https://github.com/microsoft/TypeScript/issues/32277)…)

> We should allow only `false` and not `boolean`, as having loop true does not make sense without a number

`loop = true` implies an infinite loop to me, but I guess it's subjective. I went with the below suggestion anyway…

> make this only a number, and it will not be looped if its undefined.

danielbayley:
> Could you please run `npm run docs` to re-generate markdown docs?

@OrKoN Done.

> I left some comments

@Lightning00Blade Should be all good now… I ran into a bunch of issues with the existing code, which have all now been fixed also.

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -56,25 +62,73 @@ export class ScreenRecorder extends PassThrough {
...
+    const formatArgs = this.#getFormatArgs(
+      format,
+      fps,
+      loop,
+      delay,
+      quality,
+      colors,
+    );
+    const vf = formatArgs.indexOf('-vf');
+    if (~vf) {
```
Lightning00Blade: Let's just check `if (vf !== -1)` it easier to understand.
Also in the next line if we push a empty string won't we end up with `<previous-filter>,,`
Which could break the command line in theory?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -345,6 +353,42 @@ export interface ScreencastOptions {
...
+   * Specifies the number of times to loop playback, from `0` to `Infinity`.
+   *
+   * @defaultValue `0`
```
Lightning00Blade: I think we want a little better explanation here, 
As if undefined is provided no loop;
If 0 or Infinity is provided - loop Infinity
And default is undefined.

packages/puppeteer-core/src/node/ScreenRecorder.ts
danielbayley: > Also in the next line if we push a empty string won't we end up with <previous-filter>,,
Which could break the command line in theory?

@Lightning00Blade Well I tested all 3 output formats, and nothing broke. But I agree it's not ideal… Also, this was purely to stop TypeScript complaining about possible `undefined`!

packages/puppeteer-core/src/api/Page.ts
danielbayley: > If 0 or Infinity is provided - loop Infinity

@Lightning00Blade Actually, I intentionally set `0` to prevent looping:

https://github.com/puppeteer/puppeteer/blob/505d75e60b2ab96c8a155b87182d7bf483607fb9/packages/puppeteer-core/src/node/ScreenRecorder.ts#L91

as this is a very confusing aspect of the underlying `ffmpeg` API. This simplifies it to be `0` for no loop, `1` to loop once, `2`, twice, etc. up to `Infinity`… So actually `0` is the same as `undefined` (the default) here. Reworded it anyway, to clarify…

Lightning00Blade: I think it looks nice with the latest implementation. @OrKoN WDYT ?

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -56,25 +62,73 @@ export class ScreenRecorder extends PassThrough {
...
+      quality,
+      colors,
+      path,
+    }: ScreenRecorderOptions = {},
   ) {
     super({allowHalfOpen: false});
 
+    format ??= 'webm';
+    fps ??= DEFAULT_FPS;
+    loop ||= -1;
```
Lightning00Blade, line 92: Can we add comment why we did this, I already forgot when reading the from the start:
Maps 0 to -1 as FFMPEG maps 0 to Infinity.

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -174,30 +220,59 @@ export class ScreenRecorder extends PassThrough {
...
-          ['-deadline', 'realtime', '-cpu-used', '8'],
         ].flat();
       case 'gif':
+        fps = DEFAULT_FPS === fps ? 20 : 'source_fps';
+        if (loop === Infinity) {
+          loop = 0;
+        }
+        if (delay !== -1) {
+          delay /= 10;
+        } // ms to cs
```
Lightning00Blade: Can you move this comment above the the delay /=10;

Lightning00Blade (commented):
From code perspective all looks good to me.
I will double check the FFMPEG integration and options and then we can merge.

packages/puppeteer-core/src/node/ScreenRecorder.ts
danielbayley: > Can you move this comment above the the delay /=10;

Sure. `npm run format` did it that way… Probably a bug in `prettier`?

Lightning00Blade: Yeah, I think it's an edge case with adding brackets when no are provided for if statement.

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -174,30 +221,60 @@ export class ScreenRecorder extends PassThrough {
...
+    quality: number,
+    colors: number,
+  ) {
+    const libvpx = [
+      // Sets the codec to use.
+      ['-c:v', 'vp9'],
+      // Sets the quality. Lower the better.
+      ['-crf', `${quality}`],
+      // Sets the quality and how efficient the compression will be.
+      ['-deadline', 'realtime', '-cpu-used', `${os.cpus().length / 2}`],
```
Lightning00Blade: Funny bug its max value here is 8 - https://trac.ffmpeg.org/wiki/Encode/VP9#CPUUtilizationSpeed
```suggestion
      ['-deadline', 'realtime', '-cpu-used', `${Math.min(os.cpus().length / 2, 8)}`],
```

packages/puppeteer-core/src/node/ScreenRecorder.ts
danielbayley: I personally prefer without `{}`, wherever possible anyway :)

Also, _[Biome](https://biomejs.dev/formatter)_ > [eslint-stylistic `--fix`](https://antfu.me/posts/why-not-prettier) > Prettier

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -318,11 +318,19 @@ export interface ScreenshotOptions {
  * @public
  * @experimental
  */
+export type FileFormat = 'gif' | 'webm' | 'mp4';
+
 export interface ScreencastOptions {
```
Lightning00Blade: There is a missing comment with TsDoc, please copy it to this interface from the FileFormat.

Lightning00Blade (approved):
LGTM, thanks again for the PR.


PS: I will deal with the rebasing and merging.

PR #13646, opened 2025-03-04
feat: add firefox arm64 downloads

### LINKED ISSUE #12595: [Feature]: Add support for arm64 Linux Firefox
### Feature description

It seems to be currently nightly only https://blog.nightly.mozilla.org/2024/04/19/firefox-nightly-now-available-for-linux-on-arm64/

OrKoN:
Closes https://github.com/puppeteer/puppeteer/issues/12595

### DIFF
--- a/packages/browsers/src/browser-data/chromium.ts
+++ b/packages/browsers/src/browser-data/chromium.ts
@@ -12,6 +12,8 @@ import {BrowserPlatform} from './types.js';
 
 function archive(platform: BrowserPlatform, buildId: string): string {
   switch (platform) {
+    case BrowserPlatform.LINUX_ARM:
+      throw new Error('Linux arm64 is not supported for chromium');
     case BrowserPlatform.LINUX:
       return 'chrome-linux';
     case BrowserPlatform.MAC_ARM:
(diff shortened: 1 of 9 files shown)


packages/browsers/src/detectPlatform.ts
```diff
@@ -19,7 +19,9 @@ export function detectBrowserPlatform(): BrowserPlatform | undefined {
         ? BrowserPlatform.MAC_ARM
         : BrowserPlatform.MAC;
     case 'linux':
-      return BrowserPlatform.LINUX;
+      return os.arch() === 'arm64'
```
Lightning00Blade: Nit: It would be better to have this in a variable like the platform as we reference it 4 times `os.arch()`

Lightning00Blade (approved):

kblok:
@OrKoN just double-checking. Chrome on Linux-arm64 is still not supported, right?

OrKoN:
> @OrKoN just double-checking. Chrome on Linux-arm64 is still not supported, right?

still no, unfortunately

packages/browsers/src/browser-data/chromium.ts
```diff
@@ -30,6 +31,7 @@ function folder(platform: BrowserPlatform): string {
       return 'Linux_x64';
     case BrowserPlatform.MAC_ARM:
       return 'Mac_Arm';
+    case BrowserPlatform.LINUX_ARM:
```
kblok, line 34: @OrKoN Are we sure that `LINUX_ARM` should return `Mac` here?

OrKoN: Looks like a typo but it should be handled correctly by line 15

OrKoN: sorry, cannot read a diff

OrKoN: https://github.com/puppeteer/puppeteer/pull/14032

PR #13683, opened 2025-03-16
fix: update Disposable polyfill to match TC39 proposal

### LINKED ISSUE #13625: Update DisposableStack and AsyncDisposable pollyfills to match the currently proposal
Our implementation slightly differs that the once in the TC39 proposal so it would be nice to align it fully.
This can be done with making sure our classes implement the available  via `esnext.disposable` https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management

ishaan1091:
Update DisposableStack and AsyncDisposableStack implementations that we have in puppeteer to match the TC39 proposal

Yes

Updated the TSDocs in the file appropriately and ran `npm run docs` as per the contributing guidelines

**Summary**
Closes https://github.com/puppeteer/puppeteer/issues/13625
- [x] Improved error handling while disposing resources
- [x] Throw SuppressedError in case multiple errors are encountered
- [x] Improve TSDocs
- [x] Add unit tests for existing functionality of DisposableStack / AsyncDisposableStack
- [x] Update DisposableStack / AsyncDisposableStack implementations that we have in puppeteer to match the TC39 proposal

No


packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -50,46 +50,41 @@ export class DisposableStack {
   #stack: Disposable[] = [];
 
   /**
-   * Returns a value indicating whether this stack has been disposed.
+   * Gets a value indicating whether the stack has been disposed.
+   * @returns {boolean}
```
Lightning00Blade: Not needed, as TypeScript provides the correct information.

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -166,46 +189,41 @@ export class AsyncDisposableStack {
   #stack: AsyncDisposable[] = [];
 
   /**
-   * Returns a value indicating whether this stack has been disposed.
+   * Gets a value indicating whether the stack has been disposed.
+   * @returns {boolean}
```
Lightning00Blade: Same not needed.

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -50,46 +50,41 @@ export class DisposableStack {
...
   }
 
   /**
-   * Adds a disposable resource to the stack, returning the resource.
-   *
-   * @param value - The resource to add. `null` and `undefined` will not be added,
-   * but will be returned.
-   * @returns The provided `value`.
+   * Adds a disposable resource to the top of stack, returning the resource. Has no effect if provided `null` or `undefined`.
+   * @template {Disposable | null | undefined} T
```
Lightning00Blade: We have the type information from TypeScript you can remove it from the JsDocs.

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -50,46 +50,41 @@ export class DisposableStack {
...
 
   /**
-   * Adds a value and associated disposal callback as a resource to the stack.
-   *
-   * @param value - The value to add.
-   * @param onDispose - The callback to use in place of a `[disposeSymbol]()`
-   * method. Will be invoked with `value` as the first parameter.
-   * @returns The provided `value`.
+   * Adds a non-disposable resource and a disposal callback to the top of the stack.
+   * @template T
```
Lightning00Blade: No JsDoc types needed.

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -101,7 +96,9 @@ export class DisposableStack {
   }
 
   /**
-   * Adds a callback to be invoked when the stack is disposed.
+   * Adds a disposal callback to the top of the stack to be invoked when stack is disposed.
+   * @param {() => void} onDispose - A callback to evaluate when this object is disposed.
```
Lightning00Blade: No JsDoc types needed.

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -145,15 +143,40 @@ export class DisposableStack {
    */
   move(): DisposableStack {
     if (this.#disposed) {
-      throw new ReferenceError('a disposed stack can not use anything new'); // step 3
+      throw new ReferenceError('A disposed stack can not use anything new');
     }
-    const stack = new DisposableStack(); // step 4-5
+    const stack = new DisposableStack();
     stack.#stack = this.#stack;
```
Lightning00Blade, line 148: We need to nullify the stack here after copying it.
```suggestion
    stack.#stack = this.#stack;
    this.#stack = [];
```

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -145,15 +143,40 @@ export class DisposableStack {
...
+    this.#disposed = true;
+    const errors: unknown[] = [];
+    for (const resource of this.#stack) {
+      try {
+        resource[disposeSymbol]();
+      } catch (e) {
+        errors.push(e);
+      }
+    }
+    if (errors.length === 1) {
```
Lightning00Blade, line 170: From my understanding this is not suppose to throw an error, it should suppress the errors that are produced by resouce[disposeSymbol].

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -166,46 +189,41 @@ export class AsyncDisposableStack {
...
-   * Adds a disposable resource to the stack, returning the resource.
-   *
-   * @param value - The resource to add. `null` and `undefined` will not be added,
-   * but will be returned.
-   * @returns The provided `value`.
+   * Adds a AsyncDisposable resource to the top of stack, returning the resource. Has no effect if provided `null` or `undefined`.
+   * @template {AsyncDisposable | null | undefined} T
+   * @param {T} value - A `AsyncDisposable` object, `null`, or `undefined`.
+   * `null` and `undefined` will not be added, but will be returned.
+   * @returns {T} The provided `value`.
```
Lightning00Blade: No JsDoc types.

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Yes I am working on resolving those, these were the linting issues I was talking about

Will submit final PR by weekend

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Yes I am working on resolving those, these were the linting issues I was talking about

Will submit final PR by weekend

packages/puppeteer-core/src/util/disposable.ts
ishaan1091:
```
/*
using declarations are supposed to be resilient to exceptions; 
if an error is thrown, it’s rethrown after disposal. On the other hand, 
the body of your function might execute as expected, but the Symbol.dispose might throw. 
In that case, that exception is rethrown as well.

But what happens if both the logic before and during disposal throws an error? For those cases, 
SuppressedError has been introduced as a new subtype of Error. It features a suppressed property 
that holds the last-thrown error, and an error property for the most-recently thrown error.
*/

class ErrorA extends Error {
    name = "ErrorA";
}
class ErrorB extends Error {
    name = "ErrorB";
}
function throwy(id: string) {
    return {
        [Symbol.dispose]() {
            throw new ErrorA(`Error from ${id}`);
        }
    };
}
function func() {
    using a = throwy("a");
    throw new ErrorB("oops!")
}
try {
    func();
}
catch (e: any) {
    console.log(e.name); // SuppressedError
    console.log(e.message); // An error was suppressed during disposal.
    console.log(e.error.name); // ErrorA
    console.log(e.error.message); // Error from a
    console.log(e.suppressed.name); // ErrorB
    console.log(e.suppressed.message); // oops!
}
```

According to the above paragraph from [TC39 doc](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management) this should throw a normal error only and the language implementation of `using` should handle that this should be converted to a SuppressedError if the code logic before calling disposal also throws error, otherwise in normal cases where either only normal code flow throws error or only disposal throws error we don't have SuppressedError in that case

ishaan1091: Oh just went through this file as well - https://github.com/tc39/proposal-explicit-resource-management?tab=readme-ov-file#the-disposablestack-and-asyncdisposablestack-container-objects 

I think it explain it much better and I see what you are saying, let me update it accordingly as well

Attached this doc in the issue as well for everyone's reference

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Fixed

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Removed

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Removed

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Removed

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Fixed

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Fixed now following TC39 spec if it is a single error we throw it as it is, but if we encounter multiple errors then they are wrapped in SuppressedError in the order as specified under TC39 i.e. main error as the first encountered error and remaining error stored as chain of suppressed errors

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Removed

ishaan1091:
Unit Tests Pending Fix currently, will try to wrap up by tomorrow or day after

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Also just for context linking this paragraph from above TC39 proposal based on which I think error should be thrown - 
```
AsyncDisposableStack is the async version of DisposableStack and is a container used to aggregate async disposables, guaranteeing that every disposable resource in the container is disposed when the respective disposal method is called. If any disposable in the container throws an error during dispose, or results in a rejected Promise, it would be thrown at the end (possibly wrapped in a SuppressedError if multiple errors were thrown):
```

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Fixed

packages/puppeteer-core/src/util/disposable.test.ts
```diff
@@ -0,0 +1,185 @@
+/**
+ * @license
+ * Copyright 2025 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import {assert} from 'console';
```
Lightning00Blade, line 7: Let't not use console assert.

Lightning00Blade: Let's not use assert but use expect, mixing asserting strategies is usually confusing.

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -261,15 +285,83 @@ export class AsyncDisposableStack {
...
+    }
+  }
 
   readonly [Symbol.toStringTag] = 'AsyncDisposableStack';
 }
+
+/**
+ * Represents an error that occurs when multiple errors are thrown during
+ * the disposal of resources. This class encapsulates the primary error and
+ * any suppressed errors that occurred subsequently.
```
Lightning00Blade, line 347: Needs to be marked as internal in the comment, so it does not show up in the docs.
```suggestion
 * any suppressed errors that occurred subsequently.
 * @internal
```
Remember to run `npm run docs` after that.

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -261,15 +285,83 @@ export class AsyncDisposableStack {
...
+    error: unknown,
+    suppressed: unknown,
+    message = 'An error was suppressed during disposal',
+  ) {
+    super(message);
+    this.name = 'SuppressedError';
+    this.#error = error;
+    this.#suppressed = suppressed;
+    // Restore the prototype chain
+    Object.setPrototypeOf(this, new.target.prototype);
```
Lightning00Blade: Why is this needed? (Sorry I didn't look at the spec for this yet.)

Lightning00Blade (commented):
Gave it a quick look and found some small things,
will try to do the full review by early next week.

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: This is needed because this helps with the instanceof check to work, etc

ref: https://stackoverflow.com/questions/41102060/typescript-extending-error-class

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Oh yeah missed that, let me fix it

ishaan1091: Fixed

packages/puppeteer-core/src/util/disposable.test.ts
ishaan1091: Chnaged assert statements to expect statements, and removed import

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Btw this is nowhere specifically mentioned in the spec as it doesn't deal with implementation logic but rather how SuppressedError will behave and how it should be thrown, etc

ishaan1091:
Hey @Lightning00Blade did you get a chance to review these changes?

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -166,45 +193,40 @@ export class AsyncDisposableStack {
...
+   * Adds a AsyncDisposable resource to the top of stack, returning the resource.
+   * Has no effect if provided `null` or `undefined`.
    *
-   * @param value - The resource to add. `null` and `undefined` will not be added,
-   * but will be returned.
+   * @param value - A `AsyncDisposable` object, `null`, or `undefined`.
+   * `null` and `undefined` will not be added, but will be returned.
    * @returns The provided `value`.
    */
   use<T extends AsyncDisposable | null | undefined>(value: T): T {
```
Lightning00Blade: I think the type here is wrong, T can extend `Disposable` as well as `AsyncDisposable`. 
And we need to check that case also. It's on the GetDisposeMethod in the spec update https://arai-a.github.io/ecma262-compare/?pr=3000&id=sec-getdisposemethod

packages/puppeteer-core/src/util/disposable.ts
Lightning00Blade: I looked at it a little bit and I don't think we need this bit in modern time
image

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Okay, will check this out and share updated PR by day after

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Fixed, removed this line

packages/puppeteer-core/src/util/disposable.ts
ishaan1091: Updated so that it takes and check for Disposable as well

Lightning00Blade (approved):
Huge thanks for the PR :rocket:

ishaan1091:
@Lightning00Blade Does this needs other's review as well for merging? Or what's the process from here?

Lightning00Blade:
Process is only on our side from now, and the only thing preventing me from pressing the merge button is I want to double check if we want to make release before we do (on the off chance it break someone).

PR #13684, opened 2025-03-17
ci: extract building in check-changes

Lightning00Blade:
It also fixes some dependencies issues that were found.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,7 @@ updates:
       timezone: Europe/Berlin
     ignore:
       - dependency-name: devtools-protocol
+      - dependency-name: chromium-bidi
       - dependency-name: '@types/node'
       - dependency-name: '@angular-devkit/*'
     groups:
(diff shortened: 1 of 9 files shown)


.github/dependabot.yml
```diff
@@ -9,6 +9,7 @@ updates:
       timezone: Europe/Berlin
     ignore:
       - dependency-name: devtools-protocol
+      - dependency-name: chromium-bidi
```
Lightning00Blade, line 12: Dependabot always messes up the `package-lock.json` so disabling it.

.github/workflows/changed-packages.yml
```diff
@@ -28,6 +28,14 @@ jobs:
         run: |
           git fetch origin main --depth 1 &&
           git merge-base --is-ancestor origin/main @;
+      - name: Install dependencies
```
Lightning00Blade, line 36: This should populate the cache early and prevent the test for running saving us some time on if the build fails.

.github/dependabot.yml
sadym-chromium: Lets split into 2 different PRs, as if one of the changes (roll Mapper + update cache) is failing, the other will have to be rolled back as well

Lightning00Blade: Done will send a PR for the chromium-bidi bump separately.

sadym-chromium (approved):

PR #13676, opened 2025-03-12
fix: `screencast` mp4 encoding

danielbayley:
**What kind of change does this PR introduce?**

Quick follow-up bugfix to #13645.

**Did you add tests for your changes?**

No

**If relevant, did you update the documentation?**

No

**Summary**

This flag to `ffmpeg` was causing problems with encoding mp4.

**Does this PR introduce a breaking change?**

No

**Other information**

PR #13708, opened 2025-03-28
feat: add `screencast` override options

danielbayley:

A follow-up to #13645, including another bug fix.

Yes

Yes

**Summary**

Fixes an existing bug, which is a discrepancy between the docs and the actual API (see `commit`s…), narrows some `type`s, and refactors the API around some of the options.

No

**Other information**

Related to: #13846, #11084, #13645, and #13676.


packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -150,9 +173,9 @@ export class ScreenRecorder extends PassThrough {
         ],
         // Forces input to be read from standard input, and forces png input
         // image format.
-        ['-f', 'image2pipe', '-c:v', 'png', '-i', 'pipe:0'],
-        // Overwrite output and no audio.
-        ['-y', '-an'],
+        ['-f', 'image2pipe', '-vcodec', 'png', '-i', 'pipe:0'],
```
Lightning00Blade, line 165: Can you point me to the docs for the diff between the `-c:v` and `-vcodec`

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -150,9 +173,9 @@ export class ScreenRecorder extends PassThrough {
...
+        // No audio
+        ['-an'],
```
Lightning00Blade, line 167: Why don't we need overwrite (`-y`)?

Lightning00Blade (commented):
What is the main motivation of having support for ScreenRecorder with only exporting pictures ?

danielbayley:
> What is the main motivation of having support for ScreenRecorder with only exporting pictures ?

@Lightning00Blade So `ffmpeg` doesn’t always provide the best available encoding for every format, so being able to convert a folder full of lossless PNGs using another, more specialised tool can be useful (e.g. especially for GIF export).

Also, we might want to select a still from a range of frames from an animation playthrough.

Accurate timing with JavaScript is actually tricky, so attempting to do this using the `screenshot` API, based on `document.getAnimations` events from within a `page.evaluate` function body—with its inaccessible scope—is at best a mess! Or perhaps not even possible.

More specifically, I am building an easy-to-use tool on top of Puppeteer, to render animated SVGs. Hence these PRs and bug fixes…

packages/puppeteer-core/src/node/ScreenRecorder.ts
danielbayley: > Can you point me to the docs for the diff between the `-c:v` and `-vcodec`

There isn’t any, it's an alias. I thought `-vcodec` is more readable, without having to look up what `-c:v` is, and negates the need for a comment saying “sets the codec”…

packages/puppeteer-core/src/node/ScreenRecorder.ts
```diff
@@ -165,7 +188,10 @@ export class ScreenRecorder extends PassThrough {
         // Filters to ensure the images are piped correctly,
         // combined with any format-specific filters.
         ['-vf', filters.join()],
-        'pipe:1',
+        // Write to provided path, else pipe to stdout.
+        path ?? 'pipe:1',
+        // Overwrite output, or exit immediately if file already exists.
+        [overwrite ? '-y' : '-n'],
```
danielbayley, line 181: > Why don't we need overwrite (`-y`)?

@Lightning00Blade I added it back here, behind an `overwrite` option (`true` by default, so no breaking API change…)

Lightning00Blade:
Sorry for the very late response. Two thing come to mind regarding this PR.
One is is possible to split it into two - 
- One to update the exiting APIs with the changes proposed here.
- The other introducing the new image output.
That way we can get part of this merged. 

Can you walk me though why we need to use `ffmpeg` to store the images, could we on theory just use `node:fs` to write to filesystem, that way remove one dependency and on theory we can proved a browser integration?

The only issue would be that we can't support `.webp` as CDP does not support that (https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-startScreencast)

danielbayley:
> is it possible to split it into two

@Lightning00Blade OK, done. See #13846.

> Can you walk me through why we need to use ffmpeg to store the images,

I guess we technically don’t need to invoke `ffmpeg` for straight images export. It just seemed that `ffmpeg` was made a hard [dependency of the `screencast`](https://pptr.dev/api/puppeteer.page.screencast#remarks) feature…

> You must have [ffmpeg](https://ffmpeg.org) installed on your system.

is that not actually the case?

> could we in theory just use `node:fs` to write to filesystem, that way remove one dependency and in theory we can provide browser integration?

Theoretically I suppose, but it will require a more complicated refactor of the existing code structure. Also, passing through `ffmpeg` enables setting the output `quality` of the frames…

> we can't support `.webp` as CDP does not support that (https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-startScreencast)

Ok, removed. without digging too deep into these formats, I had just assumed that `webp` would work, with `webm` being the default for video, and `webp` being already [part of the API](https://github.com/puppeteer/puppeteer/blame/main/packages/puppeteer-core/src/api/Page.ts#L2592) for `screenshot`…

Lightning00Blade (approved):
LGTM, will merge in the coming days and look at the other PR,

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -295,7 +305,7 @@ export interface ScreenshotOptions {
    * relative to current working directory. If no path is provided, the image
    * won't be saved to the disk.
    */
-  path?: string;
+  path?: `${string}.${ImageFormat}`;
```
andersk, line 308: This makes it unnecessarily annoying to use a computed value for `path`. For example, ``await page.screenshot({ path: path.join(dir, `${name}.png`) })`` leads to error TS2769: ``Type 'string' is not assignable to type '`${string}.webp` | `${string}.jpeg` | `${string}.png`'.``

PR #13761, opened 2025-04-09
feat: expose utilities for managing browser artifacts manually

devversion:
When integrating Puppeteer into e.g. Bazel, we cannot rely on Puppeteer downloading browsers for users. Instead, ruleset authors compute a set of download URLs, SHAs etc so that Bazel can perform the binary download.

In order to be able to use Puppeteer-core at runtime though, or generally stay up-to-date, we want to leverage `@puppeteer/browsers` to manage browser versions and URLs. This commit expands the API surface to expose the following things:

- `getDownloadURLs` for retrieving download URLs
- `BrowserTag` so that `resolveBuildId` can rely on it in a type-safe way.
- `computeExecutablePath` is modified to also work without a `cacheDir`.


packages/browsers/src/install.ts
```diff
@@ -459,7 +459,7 @@ export async function canDownload(options: InstallOptions): Promise<boolean> {
   );
 }
 
-function getDownloadUrl(
+export function getDownloadUrl(
```
OrKoN, line 469: Let's add a jsdoc marking it with `@public` and a short description about what the function does

OrKoN (approved):

Issue #13775, opened 2025-04-11
[Feature]: Make proxy-agent dependency optional


vobarian:
### Feature description

Please consider making the dependency on proxy-agent optional or splitting this functionality into a separate package. This dependency brings in a sprawling chain of large transitive dependencies, but is only needed in limited cases. It expands the bundle size of applications using Puppeteer and increases the potential need for frequent vulnerability patches due to the number of transitive dependencies. For users who will not be using proxies it seems unnecessary.

To illustrate, here is the bundle analysis produced by esbuild for a project using puppeteer-core. Two large dependencies are @tootallnate/quickjs-emscripten and esprima, both of which are transitive dependencies of proxy-agent. Several others are ast-types, escodegen, basic-ftp, source-map, ip-address, socks, socks-proxy-agent, and many more, all of which seem to be unnecessary if we're not using a proxy.

<img width="1599" alt="Image" src="https://github.com/user-attachments/assets/03f818a0-57ba-4f3e-b874-e64a169ad4bc" />

Additionally, I was surprised that puppeteer-core depends on @puppeteer/browsers. I do not know the internal structure of Puppeteer, but my understanding was that the browsers package is responsible for downloading or locating browsers. In my use case we have prepackaged Chromium binaries at a known path running in an AWS Lambda environment. It would be nice to have a way to just purely drive Chromium without the extra dependencies. For example, besides proxy-agent, @puppeteer/browsers brings in yargs, which will never be used in this environment. We do not need to download or locate a browser distribution.

For now, I have worked around it by identifying these packages as external in our bundling tool. However any vulnerabilities in these dependencies will still show up in our project, increasing the maintenance burden.

OrKoN:
It would great if there was a way to do it without interrupting the workflows of the users who do rely on those features. 

For context: puppeteer-core has always been a version of puppeteer meant to be used as a library meaning that it does not read config files and does not install anything by default but it always contained the same functionality as puppeteer. Browser management was always part of it, even before we moved the browser-related things into a separate package. 

For now, I would suggest using package manager overrides or bundler configuration to exclude modules that you know you do not use for your use case.

Related (possibly duplicate): https://github.com/puppeteer/puppeteer/issues/13552

OrKoN:
Would it be helpful if import the dependencies like yargs and proxy-agent dynamically?

vobarian:
Thank you for the response. I am using bundler options to exclude what I can; the dynamic imports should help further with that.

[...]

PR #13781, opened 2025-04-14
fix: dynamically import yargs and progress in CLI

Lightning00Blade:
Issue: #13775

Importing the two deps make Esbuild not bundle.
Example: 
image


OrKoN:
something seems to fail?

OrKoN (approved):

PR #13784, opened 2025-04-15
refactor: allow importing chromium-bidi ESM

Lightning00Blade:
Noticed in #13775, where you can see we always import the cjs module.
Importing the esm should make bundles slightly smaller.

Blocked by https://github.com/GoogleChromeLabs/chromium-bidi/pull/3326

### DIFF
--- a/test-d/ElementHandle.test-d.ts
+++ b/test-d/ElementHandle.test-d.ts
@@ -8,1003 +8,419 @@ import {expectNotType, expectType} from 'tsd';
 
 declare const handle: ElementHandle;
 
-async function runFirst() {
+const test = async () => {
   {
-    {
-      expectType<ElementHandle<HTMLAnchorElement> | null>(await handle.$('a'));
-      expectNotType<ElementHandle<Element> | null>(await handle.$('a'));
-    }
-    {
-      expectType<ElementHandle<HTMLAnchorElement> | null>(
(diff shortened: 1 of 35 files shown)


OrKoN (approved):

test-d/ElementHandle.test-d.ts
```diff
@@ -8,7 +8,7 @@ import {expectNotType, expectType} from 'tsd';
 
 declare const handle: ElementHandle;
 
-{
+async function runFirst() {
```
OrKoN: why was the change in tests needed?

Lightning00Blade: I am looking into it it started to complaining about top-level await. not sure what the cause is - maybe something on Chromium-BiDi side we didn't caught?

Lightning00Blade: I am not sure why this PR caused it but in short `tsd` tried to parse this files from TS to JS as CommonJS using tsc wrapper - my guess is there is complains as it see a top level await and those are forbidden in CJS.
Trying to find a repo without the change

OrKoN (approved):

test-d/ElementHandle.test-d.ts
Lightning00Blade: Seem like a `tsd` quirk.

PR #13810, opened 2025-04-23
feat: Add Browser.installExtension and Browser.uninstallExtension

oliverdunk:

A new feature, in particular new methods on `Browser` to install and uninstall a browser extension.

I wasn't able to identify any relevant test files for this type of change.

Used the `npm run docs` command. I will also open a follow-up PR with updates to https://pptr.dev/guides/chrome-extensions.

**Summary**

New CDP and WebDriver BiDi commands have been added for installing web extensions. This PR adds new methods to allow developers to use this capability.

NA

**Other information**

NA

### DIFF
--- a/packages/puppeteer-core/src/api/Browser.ts
+++ b/packages/puppeteer-core/src/api/Browser.ts
@@ -462,15 +462,16 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
   }
 
   /**
-   * Installs an extension and returns the ID. Available if the browser was
-   * created using `pipe: true` and the `--enable-unsafe-extension-debugging`
-   * flag is set.
+   * Installs an extension and returns the ID. In Chrome, this is only
+   * available if the browser was created using `pipe: true` and the
+   * `--enable-unsafe-extension-debugging` flag is set.
    */
   abstract installExtension(path: string): Promise<string>;
 
   /**
-   * Uninstalls an extension. Available if the browser was created using
-   * `pipe: true` and the `--enable-unsafe-extension-debugging` flag is set.
+   * Uninstalls an extension. In Chrome, this is only available if the browser
+   * was created using `pipe: true` and the
+   * `--enable-unsafe-extension-debugging` flag is set.
    */
   abstract uninstallExtension(id: string): Promise<void>;
 
--- a/packages/puppeteer-core/src/api/Browser.ts
+++ b/packages/puppeteer-core/src/api/Browser.ts
@@ -462,6 +462,19 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
   }
 
   /**
+   * Installs an extension and returns the ID. Available if the browser was
+   * created using `pipe: true` and the `--enable-unsafe-extension-debugging`
+   * flag is set.
+   */
+  abstract installExtension(path: string): Promise<string>;
+
+  /**
+   * Uninstalls an extension. Available if the browser was created using
+   * `pipe: true` and the `--enable-unsafe-extension-debugging` flag is set.
+   */
+  abstract uninstallExtension(id: string): Promise<void>;
+
+  /**
    * Whether Puppeteer is connected to this {@link Browser | browser}.
    *
    * @deprecated Use {@link Browser | Browser.connected}.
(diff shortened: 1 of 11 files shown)


packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -242,6 +242,14 @@ export class BidiBrowser extends Browser {
     return this.defaultBrowserContext().newPage();
   }
 
+  override installExtension(path: string): Promise<string> {
```
oliverdunk, line 245: It wasn't entirely clear to me if I should be adding to Browser or BrowserContext. However, I think this makes sense, as an extension can only be installed once and then applies to both the main and incognito contexts (with the caveat that you have to grant it permission to run in incognito using the UI).

packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -461,6 +461,19 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
     return await this.defaultBrowserContext().deleteCookie(...cookies);
   }
 
+  /**
+   * Installs an extension and returns the ID. Available if the browser was
```
OrKoN: This API would work for Firefox and other browsers potentially, let's make sure that these specific requirements mention Chrome specifically.

OrKoN: In fact, Firefox WebDriver BiDi does not work with pipes so it could be interpreted wrong.

OrKoN:
LGTM but let's add a basic test into a new spec `test/src/webExtenion.spec.ts` using a sample extension.

packages/puppeteer-core/src/bidi/Browser.ts
OrKoN: Browser is fine for now (usually APIs on the browser make changes to the default context). While the behavior is true for Chrome it might not be true for other browsers so eventually we could add it to the BrowserContext as well.

packages/puppeteer-core/src/api/Browser.ts
oliverdunk: I went ahead and updated the comment :)

test/TestExpectations.json
```diff
@@ -1708,5 +1708,19 @@
     "parameters": ["cdp", "chrome", "chrome-headless-shell"],
     "expectations": ["FAIL", "PASS"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
+  },
+  {
+    "testIdPattern": "[webExtension.spec] webExtension can install and uninstall an extension",
+    "platforms": ["darwin"],
```
```suggestion
    "platforms": ["darwin", "linux", "win32"],
```

test/TestExpectations.json
```diff
@@ -1708,5 +1708,19 @@
...
+    "parameters": ["chrome", "chrome-headless-shell", "cdp"],
+    "expectations": ["FAIL"],
+    "comment": "Extensions are not supported in shell mode."
+  },
+  {
+    "testIdPattern": "[webExtension.spec] webExtension can install and uninstall an extension",
+    "platforms": ["darwin"],
```
```suggestion
    "platforms": ["darwin", "linux", "win32"],
```

test/src/webExtension.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  it.only('can install and uninstall an extension', async () => {
+    const {defaultBrowserOptions, isChrome} = await getTestState({
+      skipLaunch: true,
+    });
+
+    const options = Object.assign({}, defaultBrowserOptions);
+
+    if (isChrome) {
+      options.ignoreDefaultArgs = ['--disable-extensions'];
```
OrKoN, line 27: let's add an option to launch called `enableExtensions (default: false)` that way we can manage args for the user?

test/src/webExtension.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import {getTestState, launch, setupTestBrowserHooks} from './mocha-utils.js';
+
+const EXTENSION_PATH = path.join(__dirname, '/../assets/simple-extension');
+const EXPECTED_ID = 'mbljndkcfjhaffohbnmoedabegpolpmd';
+
+describe('webExtension', function () {
+  setupTestBrowserHooks();
+
+  it.only('can install and uninstall an extension', async () => {
```
OrKoN: remove `.only`

test/src/webExtension.spec.ts
oliverdunk: As discussed elsewhere, let's do this in a follow-up PR.

OrKoN (approved):

PR #13824, opened 2025-04-29
feat: Add `enableExtensions` launch option

oliverdunk:

A new launch option (`enableExtensions`) which can be set to `string[]`, `true` or `false`:

- `string[]` - load the provided list of paths as unpacked extensions.
- `true` - avoids passing default args that would disable extensions. Further extension installation is allowed.
- `false` - disables extensions entirely.

**Known issues**

- In Chrome, passing a `string[]` or `true` removes the `--disable-extensions` flag which means other extensions may run.
- In Chrome, `pipe: true` must also be provided if a `string[]` is passed.
- In Firefox, there is no known command line flag to disable addons so they are always enabled.
- In Firefox, passing a `string[]` uses WebDriver BiDi commands which have not yet reached stable.

Covered by existing tests.

Yes

**Summary**

This avoids developers needing to manually change the flags when wanting to install extensions.

No

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/node/BrowserLauncher.ts
+++ b/packages/puppeteer-core/src/node/BrowserLauncher.ts
@@ -74,6 +74,7 @@ export abstract class BrowserLauncher {
   async launch(options: LaunchOptions = {}): Promise<Browser> {
     const {
       dumpio = false,
+      enableExtensions = false,
       env = process.env,
       handleSIGINT = true,
       handleSIGTERM = true,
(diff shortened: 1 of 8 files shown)


packages/puppeteer-core/src/node/BrowserLauncher.ts
```diff
@@ -209,6 +210,17 @@ export abstract class BrowserLauncher {
       throw error;
     }
 
+    if (Array.isArray(enableExtensions)) {
+      // In Chrome, extensions are installed using the default args instead.
```
OrKoN: Should we do this for all browsers? I would expect users to use either enableExtensions or browser.installExtension and do not rely on explicit `--load-extension` any more?

packages/puppeteer-core/src/node/BrowserLauncher.ts
```diff
@@ -209,6 +210,23 @@ export abstract class BrowserLauncher {
...
 
+    if (Array.isArray(enableExtensions)) {
+      if (this.#browser === 'chrome' && !usePipe) {
+        throw new Error(
+          'To use `enableExtensions` with a list of paths in Chrome, you must be connected with `--remote-debugging-pipe` (`pipe: true`).',
+        );
+      }
+
+      // In Chrome, extensions are installed using the default args instead.
+      if (this.#browser !== 'chrome') {
```
OrKoN: I think this condition need to be removed

OrKoN (approved):

PR #13860, opened 2025-05-08
docs: drop redundant launch flag --disable-setuid-sandbox

TerrorJack:

When `--no-sandbox` is passed, all chromium sandboxes are disabled and `--disable-setuid-sandbox` is a no-op. See
https://source.chromium.org/chromium/chromium/src/+/main:content/browser/zygote_host/zygote_host_impl_linux.cc;l=122?q=kDisableSetuidSandbox, the only place where `kDisableSetuidSandbox` is used other than redirecting cli options: if `kNoSandbox` is enabled then `ZygoteHostImpl::Init` bails out early, and the code path concerning `--disable-setuid-sandbox` is unreachable anyway.

Not needed.

Yes.

**Summary**

See first subsection.

No.

**Other information**

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -261,7 +261,7 @@ Chrome with the `--no-sandbox` argument:
 
 ```ts
 const browser = await puppeteer.launch({
-  args: ['--no-sandbox', '--disable-setuid-sandbox'],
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

OrKoN:
Thanks!

PR #13873, opened 2025-05-16
Update examples.md

jochen-testingbot:

Documentation: specify a [link to a GitHub repo](https://github.com/testingbot/puppeteer-testingbot-example) that shows how to run Puppeteer tests in the cloud.

At TestingBot, we provide a remote grid of browsers, allowing you to run Puppeteer scripts on different browser versions and operating systems.

Would be great if you could add a link to the example repository.

Thanks!


OrKoN (commented):
Thanks for the PR but it seems running the examples requires a paid service and we try to avoid including examples that cannot be run locally.

PR #13896, opened 2025-05-22
chore: remove setting of the "remote.active-protocol" preference for Firefox

whimboo:
With Firefox 141 the CDP code base is removed and we should as well stop setting the given preference because it no longer exists.

@OrKoN not sure if there are as well other pieces that we have to remove. Would you mind checking? Happy to take care of them as well.

Also when do we want to remove CDP support for Firefox in Puppeteer? Note that the upcoming Firefox 140 ESR will still be around for another year. Do you still want to support it? Or did you already stop the CDP support? I cannot fully remember.


whimboo:
Oh, it was actually removed over on https://github.com/puppeteer/puppeteer/pull/13427. So we probably can remove this preference right away.

OrKoN (approved):

PR #13901, opened 2025-05-24
feat: allow turning off network events

### LINKED ISSUE #13065: [Feature]: Disable Network.enable
### Feature description

If I understand correctly, puppeteer sends the `Network.enable` message to chrome pages, always, and as such receives all HTTP request metadata (this is done roughly through `cdp/Page`, `cdp/FrameManager`, `cdp/NetworkManager`).

For a performance sensitive test setup, it would be nice of we can disable the network manager by default, is there a way to do or adjust puppeteer to do so?

OrKoN:
Closes https://github.com/puppeteer/puppeteer/issues/13065

### DIFF
--- a/packages/puppeteer-core/src/bidi/Browser.ts
+++ b/packages/puppeteer-core/src/bidi/Browser.ts
@@ -44,6 +44,7 @@ export interface BidiBrowserOptions {
   defaultViewport: Viewport | null;
   acceptInsecureCerts?: boolean;
   capabilities?: SupportedWebDriverCapabilities;
+  networkEnabled: boolean;
 }
 
 /**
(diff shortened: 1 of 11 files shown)


packages/puppeteer-core/src/common/ConnectOptions.ts
```diff
@@ -49,6 +49,11 @@ export interface ConnectOptions {
    * @defaultValue `false`
    */
   acceptInsecureCerts?: boolean;
+  /**
+   * Is network events enabled?
+   * @defaultValue `true`
```
Lightning00Blade, line 58: Let's mark it as experimental.

packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -111,6 +121,7 @@ export class BidiBrowser extends Browser {
   #browserContexts = new WeakMap<UserContext, BidiBrowserContext>();
   #target = new BidiBrowserTarget(this);
   #cdpConnection?: CdpConnection;
+  #networkEnabled = true;
```
Lightning00Blade: We always expect this to be set in the constructor so lets not set a default.
```suggestion
  #networkEnabled: boolean;
```

Lightning00Blade (approved):
LGTM with a comment

PR #13920, opened 2025-06-03
chore: enable Coverage test for BiDi+

Lightning00Blade:


OrKoN (approved):

Lightning00Blade:
BiDi mapper way of dealing with Script.Channels means that coverage can detect preload scripts.
This breaks the test unless run in isolation, will investigate if there is some leaking happening in BiDi. Closing for now.

PR #13959, opened 2025-06-19
test: Enable CHIPS related cookie test for Firefox

lutien:
[The dependency bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1931488) for this test was landed. So when the new Firefox version is released this test should start passing.


whimboo:
@lutien to have a correct result for Firefox Nightly could you as well update https://github.com/puppeteer/puppeteer/blob/main/test/CanaryTestExpectations.json which could be merged immediately? For this PR we have to wait until the 141 release.

lutien:
For Nightly, this test is marked as passing (only) anyway, because it is/was passing in Nightly. And I don't think we can remove it now since it's different from general test expectation, I would assume it works as an override and can be removed only together.

Lightning00Blade:
@lutien Is this ready for review after https://github.com/puppeteer/puppeteer/pull/14046 lands?

lutien:
> @lutien Is this ready for review after #14046 lands?

Yes, should be ready then

PR #13973, opened 2025-06-27
chore: add a Chrome bisect wrapper

Lightning00Blade:
Adds a tiny wrapper around the Chromium bisect script that run Puppeteer tests to verify issues.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -308,6 +308,17 @@ The following steps are manual version of the script above.
 
 For bisecting Chrome/Chromium changes use https://www.chromium.org/developers/bisect-builds-py/.
 
+```bash
+python3 <path-to-chromium-checkout>/tools/bisect-builds.py -g <known-good> -b <known-bad> -v --verify-range --not-interactive -c "BINARY=%p npm run test:chrome:<test-type>"
+```
+
+Or run the wrapper in `tools/bisect.mjs` that warps the above functionality for Puppeteer tests.
+
+```bash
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -131,7 +131,7 @@
       ]
     },
     "test": {
-      "command": "npx ./tools/mocha-runner --min-tests 1003",
+      "command": "npx ./tools/mocha-runner",
```
Lightning00Blade, line 134: I don't think we use this anywhere but locally.

docs/contributing.md
```diff
@@ -308,6 +308,17 @@ The following steps are manual version of the script above.
 
 For bisecting Chrome/Chromium changes use https://www.chromium.org/developers/bisect-builds-py/.
 
+```bash
+python3 <path-to-chromium-checkout>/tools/bisect-builds.py -g <known-good> -b <known-bad> -v --verify-range --not-interactive -c "BINARY=%p npm run test:chrome:<test-type>"
```
OrKoN:
```suggestion
python3 <path-to-chromium-checkout>/tools/bisect-builds.py -cft -g <known-good> -b <known-bad> -v --verify-range --not-interactive -c "BINARY=%p npm run test:chrome:<test-type>"
```

OrKoN (approved):

docs/contributing.md
Lightning00Blade: Added to the wrapper as well.

PR #13974, opened 2025-06-27
test: remove no default args test

OrKoN:
Depending on the browser it might not actually work.


test/src/launcher.spec.ts
```diff
@@ -115,7 +115,7 @@ describe('Launcher specs', function () {
         }
       });
     });
-    describe('Puppeteer.launch', function () {
+    describe.only('Puppeteer.launch', function () {
```
Lightning00Blade: Remove .only

Lightning00Blade (approved):

PR #14009, opened 2025-07-09
fix: make sure pipe transport dispatches messages in tasks

OrKoN:
Similar to the ws library non web-streams also dispatch messages without an extra macrotask.


Lightning00Blade (approved):
LGTM

PR #14016, opened 2025-07-11
feat(webdriver): support network response content in BiDi

sadym-chromium:
Setup data collectors in advance. Tolerate not implemented data collectors.

### DIFF
--- a/packages/puppeteer-core/src/bidi/Browser.ts
+++ b/packages/puppeteer-core/src/bidi/Browser.ts
@@ -108,8 +108,8 @@ export class BidiBrowser extends Browser {
 
     await session.send('network.addDataCollector', {
       dataTypes: [Bidi.Network.DataType.Response],
-      // TODO: check what would be the proper limit.
-      maxEncodedDataSize: 1 * 1024 * 1024 * 1024 // 1 GB.
+      // Buffer size of 200 MB is equivalent to the CDP:
+      maxEncodedDataSize: 200 * 1000 * 1000 // 200 MB
     })
 
     const browser = new BidiBrowser(session.browser, opts);
--- a/packages/puppeteer-core/src/bidi/Browser.ts
+++ b/packages/puppeteer-core/src/bidi/Browser.ts
@@ -21,6 +21,7 @@ import type {Page} from '../api/Page.js';
 import type {Target} from '../api/Target.js';
 import type {Connection as CdpConnection} from '../cdp/Connection.js';
 import type {SupportedWebDriverCapabilities} from '../common/ConnectOptions.js';
+import {ProtocolError} from '../common/Errors.js';
 import {EventEmitter} from '../common/EventEmitter.js';
 import {debugError} from '../common/util.js';
 import type {Viewport} from '../common/Viewport.js';
@@ -106,12 +107,20 @@ export class BidiBrowser extends Browser {
       }) as [string, ...string[]],
     );
 
-    await session.send('network.addDataCollector', {
-      dataTypes: [Bidi.Network.DataType.Response],
-      // Buffer size of 20 MB is equivalent to the CDP:
-      maxEncodedDataSize: 20 * 1000 * 1000, // 20 MB
-    });
-
+    try {
+      await session.send('network.addDataCollector', {
+        dataTypes: [Bidi.Network.DataType.Response],
+        // Buffer size of 20 MB is equivalent to the CDP:
+        maxEncodedDataSize: 20 * 1000 * 1000, // 20 MB
+      });
+    } catch (err) {
+      if (err instanceof ProtocolError) {
+        // Ignore protocol errors, as the data collectors can be not implemented.
+        debugError(err);
+      } else {
+        throw err;
+      }
+    }
     const browser = new BidiBrowser(session.browser, opts);
     browser.#initialize();
     return browser;
--- a/packages/puppeteer-core/src/bidi/Browser.ts
+++ b/packages/puppeteer-core/src/bidi/Browser.ts
@@ -106,6 +106,12 @@ export class BidiBrowser extends Browser {
       }) as [string, ...string[]],
     );
 
+    await session.send('network.addDataCollector', {
+      dataTypes: [Bidi.Network.DataType.Response],
+      // TODO: check what would be the proper limit.
+      maxEncodedDataSize: 1 * 1024 * 1024 * 1024 // 1 GB.
+    })
+
     const browser = new BidiBrowser(session.browser, opts);
     browser.#initialize();
     return browser;
--- a/packages/puppeteer-core/src/bidi/Browser.ts
+++ b/packages/puppeteer-core/src/bidi/Browser.ts
@@ -109,8 +109,8 @@ export class BidiBrowser extends Browser {
     await session.send('network.addDataCollector', {
       dataTypes: [Bidi.Network.DataType.Response],
       // Buffer size of 200 MB is equivalent to the CDP:
-      maxEncodedDataSize: 200 * 1000 * 1000 // 200 MB
-    })
+      maxEncodedDataSize: 200 * 1000 * 1000, // 200 MB
+    });
 
     const browser = new BidiBrowser(session.browser, opts);
     browser.#initialize();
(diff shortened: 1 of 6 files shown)


packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -106,6 +106,12 @@ export class BidiBrowser extends Browser {
       }) as [string, ...string[]],
     );
 
+    await session.send('network.addDataCollector', {
+      dataTypes: [Bidi.Network.DataType.Response],
+      // TODO: check what would be the proper limit.
+      maxEncodedDataSize: 1 * 1024 * 1024 * 1024 // 1 GB.
```
OrKoN: let's use the same defaults as in CDP (we do not change them). See the inspector_network_agent for the current values.

packages/puppeteer-core/src/bidi/core/Request.ts
```diff
@@ -34,6 +36,8 @@ export class Request extends EventEmitter<{
     return request;
   }
 
+  #responseCompleteDeferred = Deferred.create<void, Error>();
```
OrKoN: do we need a deferred here? let's just let the code to query without regard to the lifecycle. Do any tests depend on this deferred?

packages/puppeteer-core/src/bidi/Browser.ts
sadym-chromium: Changed to 200 MB [as in CDP](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/inspector_network_agent.cc;drc=33bc4b06a9ed6afd05d0065f907910e1da59e712;l=132).

OrKoN: It should be 20 MB as it per resource and not total

packages/puppeteer-core/src/bidi/core/Request.ts
sadym-chromium: [[network.spec] network Response.text should wait until response completes](https://github.com/puppeteer/puppeteer/blob/d012588b7a3fe649b0919ec14d0878390433cf6b/test/src/network.spec.ts#L339)

OrKoN: chromium-bidi is expected to already await internally https://w3c.github.io/webdriver-bidi/#command-network-getData (step 9)

OrKoN: Also, even if we need the waiter in the client side we should add it to the HTTPResponse and not to the core/, using rxjs to properly handle timeouts/closures/etc

packages/puppeteer-core/src/bidi/HTTPRequest.ts
```diff
@@ -331,6 +331,10 @@ export class BidiHTTPRequest extends HTTPRequest {
   timing(): Bidi.Network.FetchTimingInfo {
     return this.#request.timing();
   }
+
+  getResponseContent(): Promise<Uint8Array> {
+    return this.#request.getResponseContent();
```
OrKoN, line 336: why do we need to fetch response via the request?

sadym-chromium: the session is stored in the "Request", not "HTTPRequest" or "HTTPResponse". Do you think we should rather add session to the `HTTPResponse` and `HTTPRequest`?

OrKoN: The session is a wrapper `this.#request.frame().browsingContext.userContext.browser.session` but I do not think we should call it directly from "HTTPRequest" or "HTTPResponse". I forgot that we actually do not have a core/Response.ts defined and I am not sure this is justified to introduce it now. So resolving this.

packages/puppeteer-core/src/bidi/Browser.ts
sadym-chromium: ack, fixed

packages/puppeteer-core/src/bidi/core/Request.ts
sadym-chromium: removed

packages/puppeteer-core/src/bidi/core/Request.ts
```diff
@@ -110,6 +112,9 @@ export class Request extends EventEmitter<{
       this.#response = event.response;
       this.#event.request.timings = event.request.timings;
       this.emit('success', this.#response);
+      // TODO: clarify behavior in case of redirect.
+      // this.#responseCompleteDeferred.resolve();
```
OrKoN: remove the commented out promise?

for redirects, we expect the response of the last hop only

packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -106,6 +106,12 @@ export class BidiBrowser extends Browser {
       }) as [string, ...string[]],
     );
 
+    await session.send('network.addDataCollector', {
```
OrKoN: you need to catch and suppress the error here because Firefox does not support this yet

test/TestExpectations.json
```diff
@@ -1625,6 +1604,20 @@
...
+  {
+    "testIdPattern": "[network.spec] network Response.text should throw when requesting body of redirected response",
+    "platforms": ["linux"],
+    "parameters": ["chrome", "headless", "webDriverBiDi"],
+    "expectations": ["FAIL"],
+    "comment": "TODO: investigate"
+  },
+  {
+    "testIdPattern": "[network.spec] network Response.text should wait until response completes",
+    "platforms": ["linux"],
```
OrKoN: probably you want all platforms

test/TestExpectations.json
```diff
@@ -111,27 +111,13 @@
...
-    "comment": "https://github.com/w3c/webdriver-bidi/issues/747"
-  },
   {
     "testIdPattern": "[network.spec] network Response.remoteAddress *",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["webDriverBiDi"],
     "expectations": ["FAIL"],
     "comment": "Bidi does not have equivalent field"
   },
-  {
```
OrKoN, line 128: you probably want to keep these expectations for Firefox

sadym-chromium: Do `"parameters": ["webDriverBiDi"]` apply fofr Firefox?

OrKoN: yes

sadym-chromium: done

packages/puppeteer-core/src/bidi/Browser.ts
sadym-chromium: done

test/TestExpectations.json
sadym-chromium: done

test/TestExpectations.json
```diff
@@ -1625,6 +1625,20 @@
...
   },
+  {
+    "testIdPattern": "[network.spec] network Response.text should throw when requesting body of redirected response",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["chrome", "headless", "webDriverBiDi"],
+    "expectations": ["FAIL"],
+    "comment": "TODO: investigate"
+  },
+  {
+    "testIdPattern": "[network.spec] network Response.text should wait until response completes",
```
sadym-chromium: CDP relies on [`Network.loadingFinished`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-loadingFinished), while BiDi relies on [`Network.responseReceived`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-responseReceived) + [`Network. responseReceivedExtraInfo`](https://chromedevtools.github.io/devtools-protocol/tot/Network/#event-responseReceivedExtraInfo) to define "RequestFinished".

OrKoN: Could you add this to the comment and link the issue? 

What's the practical difference between the events? IIRC loadingFinished would be emitted almost at the same time responseReceived

OrKoN (approved):
LGTM thanks

test/TestExpectations.json
OrKoN: (saw it is already added)

sadym-chromium:
@OrKoN I will take care of the PR, please don't rebase it

PR #14019, opened 2025-07-12
ci: fix last maintained version

OrKoN:


tools/update_browser_revision.mjs
```diff
@@ -211,21 +211,11 @@ async function updateLastMaintainedChromeVersion(oldVersion, newVersion) {
       return browserVersions['chrome'];
     },
   );
-  if (browserVersions.indexOf(newVersion) !== -1) {
```
OrKoN, line 214: the new version was already added to the list so this was always returning early

tools/update_browser_revision.mjs
```diff
@@ -211,21 +211,11 @@ async function updateLastMaintainedChromeVersion(oldVersion, newVersion) {
...
   const newSemVer = new SemVer(newVersion, true);
 
   if (newSemVer.compareMain(oldSemVer) !== 0) {
-    const lastMaintainedSemVer = new SemVer(
-      versionData.lastMaintainedChromeVersion,
-      true,
-    );
-    const newLastMaintainedMajor = lastMaintainedSemVer.major + 1;
-
+    const newLastMaintainedMajor = newSemVer.major - 3;
```
OrKoN, line 218: I think it's more robust to just take the current version -3 instead of incrementing each time

Lightning00Blade (approved):

PR #14024, opened 2025-07-14
feat: add debugging highlight to ElementHandle.click()

OrKoN:
This PR adds an option for debugging clicks called debugHighlight. If set to to true, it will insert an animated element at the location that was clicked. It is experimental and might not work on all pages. The highlight would not persist across navigations.


Lightning00Blade (approved):
LGTM

PR #14028, opened 2025-07-15
chore: sync Firefox contentblocking preference change from Bug 1977066

juliandescottes:
The "cm" value is no longer recognized since Firefox 139 and can be safely updated to the new expected value.


juliandescottes:
cc @OrKoN

OrKoN (approved):

PR #14042, opened 2025-07-21
feat: emulateTimezone via BiDi

sadym-chromium:
In BiDi, use BiDi instead of CDP shortcut for `emulateTimezone`.

### DIFF
--- a/test/src/emulation.spec.ts
+++ b/test/src/emulation.spec.ts
@@ -450,6 +450,27 @@ describe('Emulation', () => {
       );
     });
 
+    it('should support timezone offset', async () => {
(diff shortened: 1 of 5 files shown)


Lightning00Blade (approved):

test/src/emulation.spec.ts
```diff
@@ -450,18 +450,49 @@ describe('Emulation', () => {
       );
     });
 
+    it('should support timezone offset', async () => {
```
sadym-chromium, line 453: @Lightning00Blade FYI I extended and updated the tests

PR #14043, opened 2025-07-21
feat: use BiDi for screen orientation emulation

sadym-chromium:
Use `emulation.setScreenOrientationOverride` when required.


TerrorJack:
firefox 144 supports it: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/144#webdriver_conformance_webdriver_bidi_marionette, though 140 esr doesn't have it

packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
```diff
@@ -601,6 +601,22 @@ export class BrowsingContext extends EventEmitter<{
...
   }
 
+  @throwIfDisposed<BrowsingContext>(context => {
+    // SAFETY: Disposal implies this exists.
+    return context.#reason!;
+  })
+  async setScreenOrientationOverride(
+    screenOrientation: Bidi.Emulation.ScreenOrientation | null,
+  ): Promise<void> {
+    await this.userContext.browser.session.send(
```
Lightning00Blade: Nit: lets use the helper getter.
```suggestion
    await this.#session.send(
```

Lightning00Blade (approved):

PR #14083, opened 2025-08-06
perf: add apt cleanup commands to reduce Docker image size

menotthakur:
What kind of change does this PR introduce?

Type: Build optimization
Category: Docker image size reduction
Did you add tests for your changes?

No tests needed
Just Standard Docker best practices
If relevant, did you update the documentation?

No, not required.
Summary

PR optimizes the Dockerfile
Add apt-get clean to remove package cache
Add rm -rf /var/lib/apt/lists/* to remove package lists
"Docker image size by removing unnecessary files that are cached during the package installation process. This is a widely accepted Docker best practice that helps create smaller, more efficient images"
Does this PR introduce a breaking change?

No, it won't
Only reduces final image size


docker/Dockerfile
```diff
@@ -12,7 +12,9 @@ ENV DBUS_SESSION_BUS_ADDRESS=autolaunch:
 # installs, work.
 RUN apt-get update \
     && apt-get install -y --no-install-recommends fonts-ipafont-gothic fonts-wqy-zenhei fonts-thai-tlwg fonts-khmeros \
-    fonts-kacst fonts-freefont-ttf dbus dbus-x11
+    fonts-kacst fonts-freefont-ttf dbus dbus-x11 \
+    && apt-get clean \
+    && rm -rf /var/lib/apt/lists/*
```
OrKoN: so I think this is not going to work (as evident by the failing test) because `npx puppeteer browsers install chrome --install-deps` also relies on apt to install missing browser dependencies

docker/Dockerfile
```diff
@@ -33,7 +34,9 @@ USER root
 # Overriding the cache directory to install the deps for the Chrome
 # version installed for pptruser. 
 RUN PUPPETEER_CACHE_DIR=/home/pptruser/.cache/puppeteer \
-    npx puppeteer browsers install chrome --install-deps
+    npx puppeteer browsers install chrome --install-deps \
+     && apt-get clean \
```
OrKoN, line 38: aren't these files going to be cached in the image anyway because they are created in the step on line 13? could you provide the size of the image before and after the change to see how big is the gain?

OrKoN: I think what would be better is trying to move the apt-update step into this RUN block

Lightning00Blade: I dig up a little and it seem that if we do this suggestion we will save around 20MB.

OrKoN: I think we can land this then. PTAL

Lightning00Blade (approved):
LGTM

PR #14090, opened 2025-08-06
ci: do not use fork for release-please

OrKoN:
We would need to give the bot permissions for the repo so the fork option would not be needed.


sadym-chromium (approved):
It would make sense to make the same in chromium-bidi

Issue #14080, opened 2025-08-05
Docs issues


QueAbout:
Some issues with the docs:
1. I have found that, while the majority of the example code in the docs is written in pure JS, sometimes there are fractions of code of some other type without any notice. Examples:
1.1. TypeScript example code at https://pptr.dev/guides/page-interactions#waiting-for-a-function
1.2. Unknown "toEqual" and "expect" elements at https://pptr.dev/api/puppeteer.elementhandle.__eval#example
This is confusing. I don't think that this was made on purpose, so I'm asking to rewrite such fractions in pure JS.
2. In description of JSHandle (https://pptr.dev/api/puppeteer.jshandle) it is written that "JSHandles are auto-disposed when their associated frame is navigated away or the parent context gets destroyed."
But in description of ElementHandle (https://pptr.dev/api/puppeteer.elementhandle) I see that "ElementHandles are auto-disposed when their origin frame gets navigated."
Are ElementHandles also auto-disposed when their parent context gets destroyed? If yes, please add this info to the docs - its absense is also confusing.

kingpranav21:
Solved :
1)  first issue here : https://github.com/puppeteer/puppeteer/pull/14128
2) second one here : https://github.com/puppeteer/puppeteer/pull/14129

PR #14128, opened 2025-08-21
#14080 : docs: unify code examples to use plain JavaScript

kingpranav21:
Some documentation examples currently include TypeScript-specific syntax and Jest/Playwright-Test assertions without explanation, which is confusing since most examples are in plain JavaScript.

Changes:
- Replaced TypeScript example in guides/page-interactions with equivalent JS.
- Converted ElementHandle.__eval example from Jest `expect(...).toEqual(...)` to Node.js `assert.deepStrictEqual(...)`.

This ensures consistency across docs and avoids mixing JS with TS/Jest snippets unless explicitly stated.

### DIFF
--- a/website/versioned_docs/version-24.17.0/api/puppeteer.elementhandle.__eval.md
+++ b/website/versioned_docs/version-24.17.0/api/puppeteer.elementhandle.__eval.md
@@ -102,9 +102,15 @@ HTML:
 
 JavaScript:
 
-```ts
+```js
 const feedHandle = await page.$('.feed');
-expect(
-  await feedHandle.$$eval('.tweet', nodes => nodes.map(n => n.innerText)),
-).toEqual(['Hello!', 'Hi!']);
+
+const tweets = await feedHandle.$$eval('.tweet', nodes => {
+  return nodes.map(n => n.innerText);
+});
+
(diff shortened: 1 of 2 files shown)


website/versioned_docs/version-24.17.0/api/puppeteer.elementhandle.__eval.md
```diff
@@ -102,9 +102,15 @@ HTML:
...
-).toEqual(['Hello!', 'Hi!']);
+
+const tweets = await feedHandle.$$eval('.tweet', nodes => {
+  return nodes.map(n => n.innerText);
+});
+
+console.log(tweets); // ['Hello!', 'Hi!']
+
+const assert = require('assert');
+assert.deepStrictEqual(tweets, ['Hello!', 'Hi!']);
```
OrKoN, line 115: let's remove the assertion snippet as it is not specific to puppeteer.

kingpranav21: Thanks for reviewing and for the feedback!
I understand your point about keeping assertions out of the docs to avoid non-Puppeteer specifics. That makes sense.

PR #14129, opened 2025-08-21
Update puppeteer.elementhandle.md #14080

kingpranav21:
Currently, the Puppeteer documentation for JSHandle states:

“JSHandles are auto-disposed when their associated frame is navigated away or the parent context gets destroyed.”

However, the ElementHandle documentation only states:

“ElementHandles are auto-disposed when their origin frame gets navigated.”

This inconsistency is confusing because ElementHandle inherits from JSHandle and follows the same disposal rules. In practice, ElementHandles are also disposed not only when their origin frame is navigated, but also when their parent execution context is destroyed (e.g., frame detachment, page closure, or context invalidation).

Why this change is needed

Developers reading the docs may mistakenly assume that ElementHandle survives when the parent context is destroyed, while JSHandle does not.

This discrepancy creates uncertainty and could lead to misuse or debugging confusion.

Aligning the descriptions makes it clear that both JSHandle and ElementHandle share the same lifecycle guarantees.

Proposed Documentation Update

Add the missing clarification to the ElementHandle docs:

ElementHandles are auto-disposed when their origin frame gets navigated or when their parent execution context gets destroyed.

This brings consistency with the JSHandle docs and correctly reflects runtime behavior.


OrKoN:
Thanks for the PR but the edited file is not the source. Please update the corresponding comment in the source file and MD and website will be regenerated by `npm run docs`.

PR #14135, opened 2025-08-21
fix: standardize error handling for closed connections

### LINKED ISSUE #14123: [Bug]: Connection.ts throws plain Error instead of ProtocolError

https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/cdp/Connection.ts#L134

What the hell??

Image

### Background

_No response_

### Expectation

Should throw ProtocolError

### Reality

Throws plain Error with "Protocol error" message. 

```TypeScript

```

### Puppeteer version

24.16.2

### Node version

22

### Package manager

yarn

### Package manager version

4

### Operating system

macOS

OrKoN:
Adds ConnectionClosedError class.

Closes https://github.com/puppeteer/puppeteer/issues/14123

### DIFF
--- a/packages/puppeteer-core/src/bidi/Connection.ts
+++ b/packages/puppeteer-core/src/bidi/Connection.ts
@@ -9,10 +9,10 @@ import type * as Bidi from 'chromium-bidi/lib/cjs/protocol/protocol.js';
 import {CallbackRegistry} from '../common/CallbackRegistry.js';
 import type {ConnectionTransport} from '../common/ConnectionTransport.js';
 import {debug} from '../common/Debug.js';
+import {ConnectionClosedError} from '../common/Errors.js';
 import type {EventsWithWildcard} from '../common/EventEmitter.js';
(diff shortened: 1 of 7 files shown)


packages/puppeteer-core/src/bidi/Connection.ts
```diff
@@ -100,8 +100,9 @@ export class BidiConnection
     params: Commands[T]['params'],
     timeout?: number,
   ): Promise<{result: Commands[T]['returnType']}> {
-    assert(!this.#closed, 'Protocol error: Connection closed.');
-
+    if (this.#closed) {
+      return Promise.reject(new Error('Connection closed.'));
```
OrKoN: I am thinking if it makes sense to create a dedicated error subclass for ConnectionErrors?

OrKoN: I think we should

Lightning00Blade (approved):

PR #14137, opened 2025-08-22
fix: add better stack trace to LifecycleWatcher

Lightning00Blade:


OrKoN (approved):

packages/puppeteer-core/src/cdp/LifecycleWatcher.ts
```diff
@@ -103,7 +104,8 @@ export class LifecycleWatcher {
     });
 
     signal?.addEventListener('abort', () => {
-      this.#terminationDeferred.reject(signal.reason);
+      this.#error.cause = signal.reason;
+      this.#terminationDeferred.reject(this.#error);
```
OrKoN: I think here we might want to abort with the `reason` still and maybe set `#error` as cause on the `reason` because abort errors could be checked for in other parts of the code (to be ignored for example). I think a test might be failing due to that.

Lightning00Blade: Yes that is correct fixed

PR #14147, opened 2025-08-29
chore: support locators from element handles

OrKoN:
This PR adds a way to create a locator from an ElementHandle. This would not allow the locator to refresh the handle using the selector (because it's not provided) yet but it would allow applying other action preconditions to the element handle.

BEGIN_COMMIT_OVERRIDE
chore: support locators from element handles

This PR adds a way to create a locator from an ElementHandle. This would not allow the locator to refresh the handle using the selector (because it's not provided) yet but it would allow applying other action preconditions to the element handle.
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -1105,13 +1105,22 @@ export abstract class Page extends EventEmitter<PageEvents> {
    * {@link https://pptr.dev/guides/page-interactions#prefixed-selector-syntax | prefix}.
    */
   locator<Ret>(func: () => Awaitable<Ret>): Locator<Ret>;
-  locator<Selector extends string, Ret>(
-    selectorOrFunc: Selector | (() => Awaitable<Ret>),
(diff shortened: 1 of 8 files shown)


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -1105,13 +1105,22 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
    */
   locator<Ret>(func: () => Awaitable<Ret>): Locator<Ret>;
-  locator<Selector extends string, Ret>(
-    selectorOrFunc: Selector | (() => Awaitable<Ret>),
-  ): Locator<NodeFor<Selector>> | Locator<Ret> {
-    if (typeof selectorOrFunc === 'string') {
-      return NodeLocator.create(this, selectorOrFunc);
+
+  /**
+   * Creates a locator based on an ElementHandle. This would allow refreshing
```
Lightning00Blade:
```suggestion
   * Creates a locator based on an ElementHandle. This would not allow refreshing
```

Lightning00Blade (approved):
LGTM

PR #14158, opened 2025-09-02
fix: make types between puppeteer and puppeteer-core comptaible

OrKoN:


docs/api/index.md
```diff
@@ -31,8 +31,6 @@ Blink - Chrome's rendering engine - has a concept of "accessibility tree", which
 
 Most of the accessibility tree gets filtered out when converting from Blink AX Tree to Platform-specific AX-Tree or by assistive technologies themselves. By default, Puppeteer tries to approximate this filtering, exposing only the "interesting" nodes of the tree.
 
-The constructor for this class is marked as internal. Third-party code should not call the constructor directly or create subclasses that extend the `Accessibility` class.
```
OrKoN: @Lightning00Blade do you have ideas why the generators produces a difference output with this change? it seems like types should be more correct now but I guess we loose some information because the api-extractor on puppeteer-core already removes the internal types? shall we perhaps have a ts entrypoint for docs?

Lightning00Blade: I will take a look. My gut feeling is that `declare` thingy don't get the private identifiers so it can't add that code block.

packages/puppeteer/src/puppeteer.docs.ts
```diff
@@ -0,0 +1,50 @@
+/**
```
Lightning00Blade, line 1: Can we ignore this file from the TsConfig that way we don't create in the output?

docs/api/index.md
OrKoN: Updated the PR with the dedicated entrypoint for docs to use internal puppeteer-core types for docs.

packages/puppeteer/src/puppeteer.docs.ts
OrKoN: No, the resulting output file is the input for the api-extractor. We could remove it after docs are generated but I do not find what do we use for cross-platform remove

Lightning00Blade (approved):

PR #14159, opened 2025-09-02
fix: return the parent's handle for accessibility text nodes instead of the text node

AlinaVarkki:
It's better to return the parent for the Node as then it allows for easier automation. 
Mimicking other query selectors.

### DIFF
--- a/packages/puppeteer-core/src/cdp/Accessibility.ts
+++ b/packages/puppeteer-core/src/cdp/Accessibility.ts
@@ -6,9 +6,8 @@
 
 import type {Protocol} from 'devtools-protocol';
 
-import {ElementHandle} from '../api/ElementHandle.js';
+import type {ElementHandle} from '../api/ElementHandle.js';
 import type {Realm} from '../api/Realm.js';
-import {debugError} from '../common/util.js';
 
 /**
  * Represents a Node and the properties of it that are relevant to Accessibility.
@@ -509,30 +508,15 @@ class AXNode {
         if (!this.payload.backendDOMNodeId) {
           return null;
         }
-        const handle = await this.#realm.adoptBackendNode(
+        using handle = await this.#realm.adoptBackendNode(
           this.payload.backendDOMNodeId,
         );
 
-        if (!handle || !(handle instanceof ElementHandle)) {
-          return null;
-        }
-
-        const isTextNode = await handle.evaluate(node => {
-          return node.nodeType === Node.TEXT_NODE;
-        });
-        // Text nodes are not elements, so we can't return a handle to them.
-        // Instead, we find the parent element and return a handle to the parent.
-        if (isTextNode) {
-          using parent = await handle.evaluateHandle(node => {
-            return node.parentElement;
-          });
-
-          void handle.dispose().catch(debugError);
-
-          return parent.move() as ElementHandle<Element>;
-        }
-
-        return handle;
+        // Since Text nodes are not elements, we want to
+        // return a handle to the parent element for them.
+        return (await handle.evaluateHandle(node => {
+          return node.nodeType === Node.TEXT_NODE ? node.parentElement : node;
+        })) as ElementHandle<Element>;
       },
     };
 
--- a/packages/puppeteer-core/src/cdp/Accessibility.ts
+++ b/packages/puppeteer-core/src/cdp/Accessibility.ts
@@ -6,8 +6,9 @@
 
 import type {Protocol} from 'devtools-protocol';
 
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -508,9 +508,28 @@ class AXNode {
         if (!this.payload.backendDOMNodeId) {
           return null;
         }
-        return (await this.#realm.adoptBackendNode(
+        const handle = (await this.#realm.adoptBackendNode(
           this.payload.backendDOMNodeId,
         )) as ElementHandle<Element>;
```
Lightning00Blade: Nit: Lets remove the cast here, I think then it would resolve to `ElementHandle<Node>` which is where the error comes from in the first place.

packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -508,9 +508,28 @@ class AXNode {
...
+
+        if (!handle) {
+          return null;
+        }
+
+        // Text nodes are not elements, so we can't return a handle to them.
+        // Instead, we find the parent element and return a handle to the parent.
+        if (await handle.evaluate(node => {
```
Lightning00Blade: I think we can extract it in a variable it would be easier to read, WDYT?

packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -508,9 +508,28 @@ class AXNode {
...
+          return node.nodeType === Node.TEXT_NODE;
+        })) {
+          using parent = await handle.evaluateHandle(node => {
+            return node.parentElement;
+          });
+          await handle.dispose();
```
Lightning00Blade: This can throw so we should do something like we do in other places.
```
void handle.dispose().catch(debugError);
```

Lightning00Blade (commented):
LGTM + with some comments

One other thing we can do as we are already evaluating to get the NodeType is get the handle with 
```ts
using handle = await this.#realm.adoptBackendNode()
return  handle.evaluateHandle(node =>{
  if(node.nodeType === Node.TEXT_NODE) {
    return node.parentElement;
  }
  return node;
})
```

OrKoN:
This looks great, thanks! Could you please re-format and incorporate Nic's suggestions?

packages/puppeteer-core/src/cdp/Accessibility.ts
AlinaVarkki: Removed it, but had to add `!(handle instanceof ElementHandle)` to the following check because `adoptBackendNode` returns `JSHandle<Node>`, which is a base class for all handles

packages/puppeteer-core/src/cdp/Accessibility.ts
AlinaVarkki: Sure!

packages/puppeteer-core/src/cdp/Accessibility.ts
AlinaVarkki: done

packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -508,9 +509,30 @@ class AXNode {
         if (!this.payload.backendDOMNodeId) {
           return null;
         }
-        return (await this.#realm.adoptBackendNode(
+        const handle = await this.#realm.adoptBackendNode(
```
OrKoN:
```suggestion
        using handle = await this.#realm.adoptBackendNode(
```

packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -508,9 +509,30 @@ class AXNode {
...
+          using parent = await handle.evaluateHandle(node => {
+            return node.parentElement;
+          });
+
+          void handle.dispose().catch(debugError);
+
+          return parent.move() as ElementHandle<Element>;
+        }
+
+        return handle;
```
OrKoN:
```suggestion
        return  handle.evaluateHandle(node =>{
          if (node.nodeType === Node.TEXT_NODE) {
            return node.parentElement;
          }
          return node;
        });
```

let's combine it into one call?

AlinaVarkki: yes, I like this

OrKoN (approved):

Lightning00Blade (approved):

PR #14161, opened 2025-09-02
feat: associate network requests with document IDs

OrKoN:
This would be useful to identify which requests are related to the current navigation.

### DIFF
--- a/packages/puppeteer-core/src/bidi/HTTPRequest.ts
+++ b/packages/puppeteer-core/src/bidi/HTTPRequest.ts
@@ -108,6 +108,10 @@ export class BidiHTTPRequest extends HTTPRequest {
     }
   }
 
+  override documentId(): string {
+    throw new UnsupportedOperation('Not supported in WebDriver BiDi');
+  }
+
   override url(): string {
     return this.#request.url;
   }
--- a/packages/puppeteer-core/src/cdp/HTTPRequest.ts
+++ b/packages/puppeteer-core/src/cdp/HTTPRequest.ts
@@ -41,6 +41,7 @@ export class CdpHTTPRequest extends HTTPRequest {
   #headers: Record<string, string> = {};
   #frame: Frame | null;
   #initiator?: Protocol.Network.Initiator;
+  #documentId = '';
 
   override get client(): CDPSession {
     return this.#client;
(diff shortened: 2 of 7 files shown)


OrKoN:
Actually, I am not sure we really need it. Marking as a draft for now.

packages/puppeteer-core/src/bidi/HTTPRequest.ts
```diff
@@ -108,6 +108,10 @@ export class BidiHTTPRequest extends HTTPRequest {
     }
   }
 
+  override documentId(): string {
+    throw new UnsupportedOperation('Not supported in WebDriver BiDi');
```
Lightning00Blade, line 112: We should expose the navigationId, no?

packages/puppeteer-core/src/cdp/HTTPRequest.ts
```diff
@@ -41,6 +41,7 @@ export class CdpHTTPRequest extends HTTPRequest {
   #headers: Record<string, string> = {};
   #frame: Frame | null;
   #initiator?: Protocol.Network.Initiator;
+  #documentId = '';
```
Lightning00Blade:
```suggestion
  #documentId: string;
```

packages/puppeteer-core/src/bidi/HTTPRequest.ts
OrKoN: No, navigationId is different

starkshade (commented):
This change introduces a new API method, `HTTPRequest.documentId()`, providing an opaque ID for the document associated with a network request. It's fully implemented for the CDP protocol but explicitly marked as unsupported for WebDriver BiDi.

**Key Changes:**
- Added a new abstract method `documentId()` to the `HTTPRequest` base class.
- Implemented `documentId()` for `CdpHTTPRequest` using the `loaderId` from network events.
- Marked `documentId()` as an `UnsupportedOperation` for `BidiHTTPRequest`.
- Added new documentation for `HTTPRequest.documentId()` including its purpose and behavior.
- Introduced a new test suite for `Request.documentId` to verify its functionality for navigation and resource requests.
- Updated `TestExpectations.json` to skip the new `documentId` test for WebDriver BiDi due to lack of support.

PR #14164, opened 2025-09-03
build: exclude doc files in package.json

OrKoN:


Lightning00Blade (commented):
This won't work.
The `files` checks if any changes happen and if not it just restores from cache.
This will still trigger the error we are seeing.

Lightning00Blade (approved):

PR #14169, opened 2025-09-04
fix(accessibility): reports snapshot with uninteresting root and focusable Document is not a leaf node

Lightning00Blade:


packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -259,9 +259,7 @@ export class Accessibility {
 
     const interestingNodes = new Set<AXNode>();
     this.collectInterestingNodes(interestingNodes, defaultRoot, false);
-    if (!interestingNodes.has(needle)) {
-      return null;
-    }
+
```
Lightning00Blade, line 262: Let's discuss this change, I believe the new behavior is better as the user may know the start of something and need only it's sub-tree.

packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -426,7 +424,7 @@ class AXNode {
     if (this.#hasFocusableChild()) {
       return false;
     }
-    if (this.#focusable && this.#name) {
+    if (this.#focusable && this.#name && this.#name !== 'Document') {
```
Lightning00Blade, line 427: Let's land this like this. We can experiment later if removing this will give as more info that we need later.

OrKoN (approved):

test/src/accessibility.spec.ts
```diff
@@ -870,6 +890,28 @@ describe('Accessibility', function () {
...
     });
+
+    it('should not report Document as leaf node', async () => {
+      const {page} = await getTestState();
+
+      await page.setContent(
+        html`<main><span>Hello</span><div> </div><div>World</div></main>`,
+      );
+
+      await page.setContent(html`<span>Name:</span><input />`);
```
OrKoN, line 901: it has two setContents? cc @Lightning00Blade

Lightning00Blade: Ups I think we need only the second one

OrKoN: I will send a fix

Lightning00Blade: Thanks!

OrKoN: Fixed in https://github.com/puppeteer/puppeteer/pull/14221

PR #14172, opened 2025-09-05
chore: use global Disposable if available

Lightning00Blade:
From what I know Node v24 support this, Chrome support it and from what I can see Firefox also support it.
We should try to use the globally available types similar to how we do for Symbol.

One caveat we need to know is that if we start running our Test on Node v24 we won't be testing our implementation,
but I think that is OKish.

Also found a small bug in the current implementation.


packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -188,7 +188,13 @@ export class DisposableStack {
 /**
  * @internal
  */
-export class AsyncDisposableStack {
+export const DisposableStack: typeof DisposableStackPolyfill =
+  (globalThis.DisposableStack as any) ?? DisposableStackPolyfill;
```
Lightning00Blade, line 192: This cast is needed as it made api extractor unhappy.

OrKoN:
it sounds a bit risky can we at least add a single job with Node 24?

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -202,7 +208,7 @@ export class AsyncDisposableStack {
   /**
    * Alias for `[Symbol.asyncDispose]()`.
    */
-  async dispose(): Promise<void> {
+  async disposeAsync(): Promise<void> {
```
Lightning00Blade, line 211: This name was wrong.

packages/puppeteer-core/src/util/disposable.ts
```diff
@@ -326,11 +332,11 @@ export class AsyncDisposableStack {
       throw errors[0];
     } else if (errors.length > 1) {
       let suppressed = null;
-      for (const error of errors.reverse()) {
+      for (const error of errors) {
```
Lightning00Blade, line 335: The second reverse is also wrong.
We dispose the resources in reverse order 
So A,B  will get disposed  in order B -> A, and if both throw  Error from B should be suppressed by error from A.

OrKoN (approved):

PR #14176, opened 2025-09-08
Update package.json to fix security hole

KevinGhadyani-Okta:
Pinned `debug` to `v4.4.1` to prevent downloading the malicious `v4.4.2` package.

What happens is consumers get Puppeteer and end up downloading the bad version because of the carrot `^`. In general, it's best-practice to pin versions in `package.json` even if using Yarn or `package.lock` because consumers only get `package.json`.

While the package is removed from `npm`, many companies use some form of NPM caching like Artifactory, and if they haven't yet removed the malicious package from their cache, they'll be vulnerable.

Fixes a security flaw in `debug` found in `v4.4.2` and many other packages:
https://www.aikido.dev/blog/npm-debug-and-chalk-packages-compromised

No, but `v4.4.2` is malicious.

No.

**Summary**

https://www.aikido.dev/blog/npm-debug-and-chalk-packages-compromised

No.

**Other information**
N/A

### DIFF
--- a/packages/puppeteer-core/package.json
+++ b/packages/puppeteer-core/package.json
@@ -142,7 +142,7 @@
   "dependencies": {
     "@puppeteer/browsers": "2.10.8",
     "chromium-bidi": "8.0.0",
-    "debug": "^4.4.1",
+    "debug": "4.4.1",
     "devtools-protocol": "0.0.1495869",
     "typed-query-selector": "^2.12.0",
     "ws": "^8.18.3"


KevinGhadyani-Okta:
Looks like I won't be able to merge this PR because of an unsigned CLA. It looks like a lot of steps to get this set up, and I don't think we want to wait days for me to get corporate approval for a security fix.

Is it possible someone else could take it over for me?

OrKoN:
The compromised package was already removed.

PR #14178, opened 2025-09-09
refactor: move AriaQueryHandler outside cdp

Lightning00Blade:
Code is Protocol agnostic.


packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -148,6 +148,7 @@ export class BidiFrame extends Frame {
     });
 
     this.browsingContext.on('navigation', ({navigation}) => {
+      this._hasStartedLoading = true;
```
OrKoN: unrelated change?

Lightning00Blade: Yes, I was experimenting with something.

OrKoN (approved):

PR #14179, opened 2025-09-09
chore: adopt webdriver-bidi-protocol

OrKoN:
This allows relying on types that independent from chromium-bidi implementation % the CDP extensions.


packages/puppeteer-core/src/bidi/Connection.ts
```diff
@@ -24,21 +26,68 @@ import type {
 const debugProtocolSend = debug('puppeteer:webDriverBiDi:SEND ►');
 const debugProtocolReceive = debug('puppeteer:webDriverBiDi:RECV ◀');
 
+// Start of CDP specific non-standard protocol extensions.
```
Lightning00Blade: Why do we need to build this in Puppeteer rather then rely on the Chromium-BiDi types?

Lightning00Blade (approved):

PR #14187, opened 2025-09-10
test: extend tests for `page.setJavaScriptEnabled`

sadym-chromium:
Check how the `page.setJavaScriptEnabled` affects `setTimeout`, `setInterval` and `then`.

### DIFF
--- a/test/src/page.spec.ts
+++ b/test/src/page.spec.ts
@@ -2096,67 +2096,101 @@ describe('Page', function () {
       const {page} = await getTestState();
       // Set up an interval that increments a counter every 0ms. This will queue up tasks
       // to run as fast as possible.
-      await page.evaluate(
-        'setInterval(window.intervalCounter=(window.intervalCounter??0)+1, 0)',
-      );
+      await page.evaluate(() => {
+        return setInterval(() => {
+          return ((globalThis as any).intervalCounter =
+            ((globalThis as any).intervalCounter ?? 0) + 1);
+        }, 0);
+      });
+
       // Disable JavaScript execution on the page. This should pause timers.
       await page.setJavaScriptEnabled(false);
       // Capture the current value of the counter after JS is disabled.
-      const intervalCounter = await page.evaluate('window.intervalCounter');
-      // Wait for a microtask to complete. This gives the event loop a chance to run if it
-      // were not paused, which would have incremented the counter.
-      await page.evaluate('Promise.resolve()');
+      const intervalCounter = await page.evaluate(() => {
+        return (globalThis as any).intervalCounter;
+      });
+      // Wait for a microtask to complete. This gives the event loop with the task a
+      // chance to run if it were not paused, which would have incremented the counter.
+      await page.evaluate(() => {
+        return Promise.resolve();
+      });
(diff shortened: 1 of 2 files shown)


test/src/page.spec.ts
```diff
@@ -2092,6 +2092,72 @@ describe('Page', function () {
...
+      await page.evaluate('Promise.resolve()');
+      // Verify the counter still has not changed, confirming that paused timeouts do not
+      // resume upon re-enabling JavaScript.
+      expect(await page.evaluate('window.timeoutCounter')).toBe(timeoutCounter);
+    });
+    it('then should not pause', async () => {
+      const {page} = await getTestState();
+      // Disable JavaScript execution on the page.
+      await page.setJavaScriptEnabled(false);
+      // This test verifies that native browser operations, like resolving a promise,
```
OrKoN: Let's refer to `.then` as a microtasks https://html.spec.whatwg.org/#microtask-queuing and for interval/SetTimeout `task` https://html.spec.whatwg.org/#queue-a-task

test/src/page.spec.ts
```diff
@@ -2092,6 +2092,72 @@ describe('Page', function () {
       );
       expect(await page.evaluate('something')).toBe('forbidden');
     });
+    it('setInterval should pause', async () => {
+      const {page} = await getTestState();
+      // Set up an interval that increments a counter every 0ms. This will queue up tasks
+      // to run as fast as possible.
+      await page.evaluate(
+        'setInterval(window.intervalCounter=(window.intervalCounter??0)+1, 0)',
```
OrKoN: this is not a recommended syntax, let's pass the callback as a function and also use the function arguments to page.evaluate for readability.

OrKoN (approved):

test/src/page.spec.ts
sadym-chromium: done

test/src/page.spec.ts
sadym-chromium: done

test/src/page.spec.ts
```diff
@@ -2092,6 +2092,119 @@ describe('Page', function () {
...
+        return new Promise(resolve => setTimeout(resolve, 100));
+      })
+
+      // Verify the counter still has not changed, confirming that `setTimeout` do not
+      // resume upon re-enabling JavaScript.
+      expect(
+        await page.evaluate(() => {
+          return (globalThis as any).timeoutCounter;
+        }),
+      ).toBe(timeoutCounter);
```
sadym-chromium, line 2199: `setTimeout` is not resumed after JS is re-enabled.

test/src/page.spec.ts
```diff
@@ -2092,6 +2092,119 @@ describe('Page', function () {
...
+        return new Promise(resolve => setTimeout(resolve, 100));
+      })
+
+      // Verify that the counter increased. This confirms that timers resumed when
+      // JavaScript is re-enabled.
+      expect(
+        await page.evaluate(() => {
+          return (globalThis as any).intervalCounter;
+        }),
+      ).toBeGreaterThan(intervalCounter);
```
sadym-chromium, line 2145: `setIneterval` is resumed after JS is re-enabled.

PR #14188, opened 2025-09-10
test: set a large height explicitly for `test/assets/input/touchscreen.html`

kartarsingh-google:
The test page(`touchscreen.html`) was setting `touch-action: none`, but due to a bug in chromium(https://crbug.com/41407032) the page seems to be starting a gesture scroll. Setting height explicitly seems to circumvent this bug, and a gesture scroll is never started for such a page.

The page now receives extra pointer events since earlier a pointer cancel was being generated, which stopped rest of the pointer sequence to be seen by the webpage.


kartarsingh-google:
@OrKoN Do yo know if `firefox-headful (4-4)` failure was a flake or not, I don't see an option to re-trigger the run? At least locally I see the updated test does pass on firefox. Using `$ npm run test:firefox` to run the test on firefox.

OrKoN:
yep, it was a flake, we had some issues recently with Ubuntu + Firefox.

OrKoN (approved):
LGTM, thanks!

PR #14190, opened 2025-09-10
feat: emulate `navigator.platform` in `Page.setUserAgent`

### LINKED ISSUE #12688: [Feature]: Support extra args from Network.setUserAgentOverride
### Feature description

`Page.setUserAgent` only supports the `userAgent` and `userAgentMetadata` args from [Network.setUserAgentOverride](https://vanilla.aslushnikov.com/?Network.setUserAgentOverride). Would it be possible to get support for the `platform` and `acceptLanguage` args too? Although it's possible to set these with a raw `CDPSession`, it'd be nice to have it supported by Puppeteer so it  easily propagates to attached CDP sessions.

I'm keen on making a PR for this - curious whether you'd prefer to have the args added to the `Page.setUserAgent` function, or as separate `Page.setPlatform` and `Page.setAcceptLanguage` functions. The former feels preferable, but could be a breaking change.

Kle0s:
Introduced new API for `Page.setUserAgent`, supporting the emulation of different `platform`s.

Yes

Yes

**Summary**
Added support of the platform options to `Page.setUserAgent`. 

No . However, this is starting a change in the API of `Page.setUserAgent`.
Currently, the signature is as follows:
```ts
setUserAgent(userAgent: string, userAgentMetadata?: Protocol.Emulation.UserAgentMetadata): Promise<void>;
```
We will deprecate this signature in favour of the following one:
```ts
setUserAgent(options: {userAgent?: string; userAgentMetadata?: Protocol.Emulation.UserAgentMetadata; platform?: string}): Promise<void>;
```
At some point in the future (next major?), we will remove the deprecated signature completely. 

**Other information**
Closes #12688

### DIFF
--- a/docs/api/puppeteer.browser.md
+++ b/docs/api/puppeteer.browser.md
@@ -363,7 +363,7 @@ Uninstalls an extension. In Chrome, this is only available if the browser was cr
 
 Gets this [browser's](./puppeteer.browser.md) original user agent.
 
-[Pages](./puppeteer.page.md) can override the user agent with .
+[Pages](./puppeteer.page.md) can override the user agent with [Page.setUserAgent()](./puppeteer.page.setuseragent.md).
 
 </td></tr>
 <tr><td>
--- a/docs/api/puppeteer.browser.md
+++ b/docs/api/puppeteer.browser.md
@@ -363,7 +363,7 @@ Uninstalls an extension. In Chrome, this is only available if the browser was cr
 
 Gets this [browser's](./puppeteer.browser.md) original user agent.
 
-[Pages](./puppeteer.page.md) can override the user agent with [Page.setUserAgent()](./puppeteer.page.setuseragent.md).
+[Pages](./puppeteer.page.md) can override the user agent with .
 
 </td></tr>
 <tr><td>
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -1567,7 +1567,7 @@
     "testIdPattern": "[tracing.spec] Tracing should support a typedArray without a path",
     "platforms": ["linux"],
     "parameters": ["chrome", "webDriverBiDi"],
-    "expectations": ["FAIL"],
+    "expectations": ["FAIL", "PASS"],
     "comment": "Might be flaky on Chrome BiDi"
   },
   {
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -560,14 +560,14 @@
   },
   {
     "testIdPattern": "[page.spec] Page Page.setUserAgent should work with platform option",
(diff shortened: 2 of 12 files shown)


test/TestExpectations.json
```diff
@@ -558,6 +558,20 @@
     "expectations": ["SKIP"],
     "comment": "Suddenly started being flaky on Windows"
   },
+  {
+    "testIdPattern": "[page.spec] Page Page.setUserAgent should work with platform option",
+    "platforms": ["linux"],
```
OrKoN: the `setUserAgent`-related expectations are probably cross-platform

test/TestExpectations.json
```diff
@@ -1549,6 +1563,13 @@
     "expectations": ["FAIL"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
+  {
```
OrKoN: let's remove unrelated expectations from this PR. I will retry if there is a flake.

test/TestExpectations.json
Kle0s: Fixed

test/TestExpectations.json
Kle0s: Okay, sure. Removed

docs/api/puppeteer.browser.md
```diff
@@ -363,7 +363,7 @@ Uninstalls an extension. In Chrome, this is only available if the browser was cr
 
 Gets this [browser's](./puppeteer.browser.md) original user agent.
 
-[Pages](./puppeteer.page.md) can override the user agent with [Page.setUserAgent()](./puppeteer.page.setuseragent.md).
+[Pages](./puppeteer.page.md) can override the user agent with .
```
OrKoN: I think the links to `Page.setUserAgent` need to be updated to link to the correct overload. It appears that the links are removed altogether if there is multiple overloads.

Kle0s: I do not see a solution in the documentation of the API extractor. 

Seems like they are currently not supporting index / label selectors. Without it, I cannot specify which declaration to point to.

Do you have a suggestion?

OrKoN: You can use the following reference syntax to refer to a specific overload, for example,[`{@link Page.(screenshot:2) }`](https://github.com/puppeteer/puppeteer/blob/6763dddec8717a82a15e2aa9e814b5a1e6937c34/packages/puppeteer-core/src/api/ElementHandle.ts#L1422). I think overloads start counting from 0 though, so we should use `{@link Page.(setUserAgent:1) }`.

OrKoN (approved):

Issue #3343, opened 2018-10-04
Trying to install puppeteer inside React app


prasadmadv:
ERROR in ./node_modules/puppeteer/node6/lib/FrameManager.js
Module not found: Error: Can't resolve 'fs' in 'D:\xampp\htdocs\repos\easyshopify\node_modules\puppeteer\node6\lib'
@ ./node_modules/puppeteer/node6/lib/FrameManager.js 17:11-24
@ ./node_modules/puppeteer/node6/lib/Page.js
@ ./node_modules/puppeteer/node6/lib/Target.js
@ ./node_modules/puppeteer/node6/lib/Browser.js
@ ./node_modules/puppeteer/node6/lib/Launcher.js
@ ./node_modules/puppeteer/node6/lib/Puppeteer.js
@ ./node_modules/puppeteer/index.js
@ ./resources/assets/js/components/Example.js
@ ./resources/assets/js/app.js
@ multi ./resources/assets/js/app.js ./resources/assets/sass/app.scss

Cant resolve 'fs' where fs module is already installed.

aslushnikov:
@prasadmadv What's "Puppeteer inside react app"?

prasadmadv:
I am trying to install puppeteer using webpack to React.js. When i use npm install puppeteer, its installed successfully. But throws errors while compiling assets. I think it might be some issue with the path. That webpack is not able to resolve the path to node_modules folder but not sure.

Pls refer to this for "What I am trying to do with react + laravel "https://tighten.co/blog/adding-react-components-to-a-laravel-app"
But i need puppeteer to add to this application.

prasadmadv:
Puppeteer requires "fs" to work.
And when i got to node_modules/fs i find this:

Readme.md:
# Security holding package
This package name is not currently in use, but was formerly occupied
by another package. To avoid malicious use, npm is hanging on to the
package name, but loosely, and we'll probably give it to you if you
want it.
You may adopt this package by contacting support@npmjs.com and
requesting the name.

And
package.json:
{
  "name": "fs",
  "version": "0.0.1-security",
  "description": "",
  "main": "index.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "repository": {
    "type": "git",
    "url": "git+https://github.com/npm/security-holder.git"
  },
  "keywords": [],
  "author": "",
  "license": "ISC",
  "bugs": {
    "url": "https://github.com/npm/security-holder/issues"
  },
[...]

Issue #1747, opened 2018-01-08
PDF Creation Client Side


srhise:
So impressed with PDF creation functionality in Puppeteer.

Is it possibly to apply the PDF generation functions client side? It seems like this would be a great solution to create PDFs from pages on your *actual* chrome browser vs. headless. Given the fact it works headless, how would one go about implementing this client side in an actual browser?

ebidel:
The closest you can get is calling `window.print()` and the have the user manually chose "Save as PDF" > "Save". As for a true client-side solution, there lots of libraries on npm for generating pdfs. https://www.npmjs.com/package/pdfjs-dist (https://mozilla.github.io/pdf.js/) and https://www.npmjs.com/package/pdfkit are two.

You could also create/host a web service that generates PDFs when sent a URL or HTML. Perfect use case for headless Chrome and puppeteer! There are some [example projects](https://github.com/GoogleChrome/puppeteer/tree/master/examples#rendering-and-web-scraping) that show how to get started.

jwillingham789:
I'm using Puppeteer to generate a pdf with a supplied url and it works great for printing static pages. However, I'm working on a React application that has complex component state changes and I'm wanting to capture those state changes in a pdf. The problem I'm having is that printing via puppeteer only prints the initial page layout and doesn't reflect any of the state changes.

I can't use window.print() because the print to pdf gives me a blank page. So my question is, how does puppeteer print the visible screen and scale it to the provided pdf layout?

I've tried using libraries like jsPDF, however, they don't work well with SVGs and my project is loaded with svg charts. 

Anyways, really like the puppeteer project and will continue utilizing it!

ebidel:
Puppeteer uses headless Chrome's feature for creating PDFs. https://developers.google.com/web/updates/2017/04/headless-chrome#create_a_pdf_dom

If you run those state changes and try printing the page manually (cmd+p), are they captured in the PDF preview?

calrays:
@jwillingham789  I understand this is old, but to achieve something similar to what it sounds like you need(ed), I wrote a node script to:
1. render my React server side to static HTML
2. pass that to wkhtmltopdf (there are wrappers for popular languages)
3. send resulting PDF to client

[...]

PR #14191, opened 2025-09-10
docs: clarify that 'running Puppeteer in the browser' docs does not automate the client-side browser

ggorlen:

Documentation clarification.

No, it's a documentation change.

Yes, it's a documentation change.

**Summary**

The current documentation of [Running Puppeteer in the browser](https://pptr.dev/guides/running-puppeteer-in-the-browser) in v24.20.0 is misleading, since it implies you can use Puppeteer to generate a PDF or a screenshot of the client side page running the rollup-bundled Puppeteer. But actually, it's strictly referring to operations you can take on the _remote_ browser that has been pre-launched with a debugging port, using the Playwright API.

As written, the guide reinforces the misconception that Puppeteer can take screenshots of the current client-side page and automate the current browser fully from the client.

This PR clarifies the scope of what "running Puppeteer in the browser" actually entails, which is that it's just the Puppeteer command API, not the actual browser automation.

For evidence of the confusion caused by this particular documentation page (and the motivation for this PR), see [Is it possible to bundle Puppeteer on the frontend (no node.js) to export the current page as a pdf?](https://stackoverflow.com/questions/79757216/is-it-possible-to-bundle-puppeteer-on-the-frontend-no-node-js-to-export-the-cu)

For other threads highlighting the general goals, confusion and misconceptions surrounding running Puppeteer in the browser, see:
- [How to run Puppeteer code in any web browser?](https://stackoverflow.com/questions/54647694/how-to-run-puppeteer-code-in-any-web-browser)
- [How to make Puppeteer work with a ReactJS application on the client-side](https://stackoverflow.com/questions/55031823/how-to-make-puppeteer-work-with-a-reactjs-application-on-the-client-side?noredirect=1&lq=1)
- [is there a way to use require.js for using puppeteer in a web page?](https://stackoverflow.com/questions/61297011/is-there-a-way-to-use-require-js-for-using-puppeteer-in-a-web-page?noredirect=1&lq=1)
- [Why I am getting an error 'Uncaught ReferenceError: require is not defined at app.js:1:19' when using puppeteer](https://stackoverflow.com/questions/75702511/why-i-am-getting-an-error-uncaught-referenceerror-require-is-not-defined-at-ap?noredirect=1&lq=1)
- https://github.com/puppeteer/puppeteer/issues/3343
- https://github.com/puppeteer/puppeteer/issues/1747

No.

**Other information**

### DIFF
--- a/docs/guides/running-puppeteer-in-the-browser.md
+++ b/docs/guides/running-puppeteer-in-the-browser.md
@@ -1,17 +1,24 @@
 # Running Puppeteer in the browser
 
-Puppeteer is a powerful tool for automating browsers, but did you know it can also run within a browser itself? This enables you to leverage Puppeteer's capabilities for tasks that don't require Node.js specific features.
+Puppeteer is a powerful tool for automating browsers, but did you know its API can also run within a browser itself? This enables you to leverage Puppeteer's capabilities for tasks that don't require Node.js specific features.
+
+:::note
+
+Note that while the Puppeteer API can run from a client webpage, the automation actions are sent to a separate browser with an open debugging port.
+
+:::
+
 
 ## Supported Features
 
 While running in the browser, Puppeteer offers a variety of functionalities including:
 
 1. WebSocket Connections: Establish connections to existing browser instances using WebSockets. Launching or downloading browsers directly is not supported as it relies on Node.js APIs.
-2. Script Evaluation: Execute JavaScript code within the browser context.
-3. Document Manipulation: Generate PDFs and screenshots of the current web page.
-4. Page Management: Create, close, and navigate between different web pages.
-5. Cookie Handling: Inspect, modify, and manage cookies within the browser.
-6. Network Control: Monitor and intercept network requests made by the browser.
+2. Script Evaluation: Execute JavaScript code within the remote browser context.
+3. Document Manipulation: Generate PDFs and screenshots of the remote browser page.
+4. Page Management: Create, close, and navigate between different web pages in the remote browser.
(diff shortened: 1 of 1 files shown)


OrKoN (approved):
Thanks!

docs/guides/running-puppeteer-in-the-browser.md
```diff
@@ -1,17 +1,24 @@
...
 
 While running in the browser, Puppeteer offers a variety of functionalities including:
 
 1. WebSocket Connections: Establish connections to existing browser instances using WebSockets. Launching or downloading browsers directly is not supported as it relies on Node.js APIs.
-2. Script Evaluation: Execute JavaScript code within the browser context.
-3. Document Manipulation: Generate PDFs and screenshots of the current web page.
-4. Page Management: Create, close, and navigate between different web pages.
-5. Cookie Handling: Inspect, modify, and manage cookies within the browser.
-6. Network Control: Monitor and intercept network requests made by the browser.
+2. Script Evaluation: Execute JavaScript code within the remote browser context.
```
OrKoN, line 17: So this doc improvement is really great and the issue was that I copy-pasted this from the "running in an extension" guide. For majority of the cases, you indeed cannot automate the same browser instace where Puppeteer runs but if you have got permissions or a debugging port/pipe to expose CDP to the page context via https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-exposeDevToolsProtocol, a Puppeteer instance in that page can connect to itself. But I do not think this is a use case we actively support.

Issue #14194, opened 2025-09-11
Firefox headful runs on Ubuntu are recently flaky


OrKoN:
Fails to start within 9 seconds. Example: https://github.com/puppeteer/puppeteer/actions/runs/17636920057/job/50114949679?pr=14189

OrKoN:
cc @whimboo

OrKoN:
Testing previous FF release here https://github.com/puppeteer/puppeteer/pull/14193

whimboo:
If that's a problem with Firefox 142.0.1 there is https://bugzilla.mozilla.org/show_bug.cgi?id=1984823 which was fixed for that security release. It was about a crash during startup on Linux and may be relevant. Lets see if we get it to reproduce with debug logs enabled.

OrKoN:
```
console.warn: TopSitesFeed: Failed to fetch data from MARS server: NetworkError when attempting to fetch resource.
console.error: services.settings: 
  main/normandy-recipes-capabilities: Signature of bundled changeset is invalid: [Exception... "<no message>"  nsresult: "0x8046001e (<unknown>)"  location: "JS frame :: resource://gre/modules/ServiceRequest.sys.mjs :: open :: line 126"  data: no].
console.error: services.settings: 
  main/remote-permissions: Signature of bundled changeset is invalid: [Exception... "<no message>"  nsresult: "0x8046001e (<unknown>)"  location: "JS frame :: resource://gre/modules/ServiceRequest.sys.mjs :: open :: line 126"  data: no].
console.error: services.settings: 
  main/search-categorization: Signature of bundled changeset is invalid: [Exception... "<no message>"  nsresult: "0x8046001e (<unknown>)"  location: "JS frame :: resource://gre/modules/ServiceRequest.sys.mjs :: open :: line 126"  data: no].
console.error: services.settings: 
  main/third-party-cookie-blocking-exempt-urls: Signature of bundled changeset is invalid: [Exception... "<no message>"  nsresult: "0x8046001e (<unknown>)"  location: "JS frame :: resource://gre/modules/ServiceRequest.sys.mjs :: open :: line 126"  data: no].
console.error: services.settings: 
  main/query-stripping: Signature of bundled changeset is invalid: [Exception... "<no message>"  nsresult: "0x8046001e (<unknown>)"  location: "JS frame :: resource://gre/modules/ServiceRequest.sys.mjs :: open :: line 126"  data: no].
console.error: services.settings: 
[...]

PR #14200, opened 2025-09-11
fix: disable browser.backup for Firefox

OrKoN:
https://github.com/puppeteer/puppeteer/issues/14194

### DIFF
--- a/packages/browsers/src/browser-data/firefox.ts
+++ b/packages/browsers/src/browser-data/firefox.ts
@@ -215,6 +215,12 @@ function defaultProfilePreferences(
     // Increase the APZ content response timeout to 1 minute
     'apz.content_response_timeout': 60000,
 
+    // Disables backup service to improve startup performance and stability. See
+    // https://github.com/puppeteer/puppeteer/issues/14194. TODO: can be removed
+    // once the service is disabled on the Firefox side for WebDriver (see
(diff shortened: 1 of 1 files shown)


OrKoN:
@whimboo would you suggest we release this? I am not sure if the feature is relevant for automation use cases so it would probably also speed up the startup.

whimboo (approved):
FYI the service is described at https://firefox-source-docs.mozilla.org/browser/components/backup/docs/index.html.

So you wont need this for Puppeteer and automation cases for the web. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1988250 to get it disabled in Firefox directly for WebDriver. We will be able to land this for Firefox 144 so having this pref on your side would be required. It could then be removed with Firefox 144. Maybe add a reference to the bug instead.

whimboo:
> So you wont need this for Puppeteer and automation cases for the web. I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1988250 to get it disabled in Firefox directly for WebDriver. We will be able to land this for Firefox 144 so having this pref on your side would be required. It could then be removed with Firefox 144. Maybe add a reference to the bug instead.

@OrKoN I assume the workaround in Puppeteer can already be removed?

whimboo:
> @OrKoN I assume the workaround in Puppeteer can already be removed?

I opened https://github.com/puppeteer/puppeteer/pull/14411 for its removal.

PR #14210, opened 2025-09-15
fix: always pipe stdio and report process launch errors

### LINKED ISSUE #14062: [Bug]: Launching Chrome with the `--remote-debugging-pipe` mode impacts behavior of stderr of the managed browser process

```TypeScript

(async () => {
  {
    // default values (pipe=false, dumpio=false)
    console.log(`stderr: ${browser.process().stderr}`);
    await browser.close();
  }

  {
    const browser = await puppeteer.launch({ pipe: true, dumpio: false });
    console.log(`stderr: ${browser.process().stderr}`);
    await browser.close();
  }

  {
    const browser = await puppeteer.launch({ pipe: true, dumpio: true });
    console.log(`stderr: ${browser.process().stderr}`);
    await browser.close();
  }
})();
```

### Background

We access stderr of the launched browser (via `browser.process().stderr`) for certain validations.

We (unless in certain debugging scenarios) leave the `dumpio` parameter to the default (`false`) because we don't need that connected to the Node.js process.

We previously used CDP over WebSocket (default behavior of Puppeteer) and now seek to use a local pipe (`--remote-debugging-pipe`) by passing `pipe=true` to `puppeteer.launch`.

### Expectation

`browser.process().stderr` should be always valid.

### Reality

`browser.process().stderr` is null when passing `{ pipe: true, dumpio: false }`.

This is believed to be the outcome of these lines of code: https://github.com/puppeteer/puppeteer/blob/e180de221c6e767e3a7125c09bff2e95ef35bd99/packages/browsers/src/launch.ts#L374-L386

This is unexpected design IMHO. The `--remote-debugging-pipe` mode concerns fd 3 and 4, where it should not generally impact the behavior of stderr.

If for any reason it has to be so, please then maybe add some lines in the docs for this behavior.

```TypeScript

```

### Puppeteer version

24.15.0

### Node version

v22.17.1

### Package manager

npm

### Package manager version

11.5.1

### Operating system

Linux

OrKoN:
This PR introduces tracking for the recent browser logs and makes sure that `@puppeteer/browsers` subscribes to the logs early. It then uses the logs to detect launch issues related to the same profile having a running browser instance already. It also fixes https://github.com/puppeteer/puppeteer/issues/14062 to implement this.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -192,7 +192,7 @@
     "testIdPattern": "[userDataDir.spec] userDataDir should not launch the browser twice with the same userDataDir *",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["firefox"],
-    "expectations": ["FAIL"],
+    "expectations": ["SKIP"],
     "comment": "we currently do not detect this siutation for Firefox"
   },
   {
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -175,6 +175,27 @@
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
   {
+    "testIdPattern": "[userDataDir.spec] userDataDir should not launch the browser twice with the same userDataDir *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["chrome-headless-shell"],
+    "expectations": ["FAIL"],
+    "comment": "headless-shell can launch multiple instances from the same user data dir"
+  },
+  {
+    "testIdPattern": "[userDataDir.spec] userDataDir should not launch the browser twice with the same userDataDir *",
+    "platforms": ["win32"],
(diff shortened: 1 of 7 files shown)


Lightning00Blade (approved):

test/TestExpectations.json
```diff
@@ -174,6 +174,27 @@
...
+    "platforms": ["win32"],
+    "parameters": ["chrome"],
+    "expectations": ["FAIL"],
+    "comment": "on Windows Chrome seems to have no issues starting on the user data dir"
+  },
+  {
+    "testIdPattern": "[userDataDir.spec] userDataDir should not launch the browser twice with the same userDataDir *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox"],
+    "expectations": ["FAIL"],
```
```suggestion
    "expectations": ["SKIP"],
```

PR #14221, opened 2025-09-17
fix(a11y): handle leaf nodes without heuristics that rely on name

OrKoN:


Lightning00Blade (approved):

packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -420,16 +406,10 @@ class AXNode {
         break;
     }
 
-    // Here and below: Android heuristics
-    if (this.#hasFocusableChild()) {
-      return false;
-    }
-    if (this.#focusable && this.#name && this.#name !== 'Document') {
```
OrKoN, line 427: @Lightning00Blade I think we might have needed to keep `this.#focusable`  check as it might not be related to name checks

PR #14226, opened 2025-09-18
fix: replace function eval with parsing in page.evaluate()

OrKoN:
`new Function(str)`  evals were used to validate and stringify JS functions. This causes issues when running in environments with CSP. This PR replaces this with rules to stringify functions correctly.

### DIFF
--- a/packages/puppeteer-core/package.json
+++ b/packages/puppeteer-core/package.json
@@ -40,7 +40,8 @@
     "check": "wireit",
     "clean": "../../tools/clean.mjs",
     "prepack": "wireit",
-    "unit": "wireit"
+    "unit": "wireit",
+    "unit:only": "wireit"
   },
   "wireit": {
     "check": {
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/package.json
```diff
@@ -40,7 +40,8 @@
     "check": "wireit",
     "clean": "../../tools/clean.mjs",
     "prepack": "wireit",
-    "unit": "wireit"
+    "unit": "wireit",
+    "unit-only": "wireit"
```
Lightning00Blade: nit: unit:only

Lightning00Blade (approved):
LGTM

Let's also do the following 
- Mark as feat to we bump the Minor
- add evaluation scope 
- Update title to better reflect what we did

I am worried there may be something we miss and it breaks downstream.

OrKoN:
> Mark as feat to we bump the Minor

I do not think it is necessary as it is the same effect in terms of semver.

> add evaluation scope

sorry, did not get it, wdym?

> I am worried there may be something we miss and it breaks downstream.

Then we either fix the scenario we did not consider or revert?

Lightning00Blade:
> sorry, did not get it, wdym?

Conventional commit - `<type>[optional scope]: <description>`

> Then we either fix the scenario we did not consider or revert?

Yes I agree but it's about discoverability from the user perspective, I think the title is good enough now

PR #14236, opened 2025-09-23
fix(pipes): concat buffers instead of string concatenation

Lightning00Blade:


OrKoN (approved):
LGTM (let's publish as a fix)

PR #14246, opened 2025-09-24
fix: regression in function stringification

### LINKED ISSUE #14244: [Bug]: Evaluation method breaks if there is no space between the function keyword and the open parenthesis

```TypeScript

await page.evaluateHandle(function() { return document.body }); // 💥 `Unexpected token 'function'`

await browser.close()
```

### Background

Using the `function() { ... }` syntax (without space) in evaluation methods fail with the version 24.22.1 version due to this change: https://github.com/puppeteer/puppeteer/commit/d0e95b19796d6492d09244eb2fe7918a23b2d65d/#diff-01bf1ca1f57ab4f6712abe84c6629bc90f58b0015e7c6aa71409426ee44287fdR33.

The parsing only takes into account the `function () { ... }` syntax (with a space). So in the case of the `function() { ... }` syntax (without space), a `function ` prefix is [added](https://github.com/puppeteer/puppeteer/commit/d0e95b19796d6492d09244eb2fe7918a23b2d65d/#diff-01bf1ca1f57ab4f6712abe84c6629bc90f58b0015e7c6aa71409426ee44287fdR49) which results in `function function() { ... }` which is invalid because `function` is a reserved keyword.

### Expectation

The `function() { ... }` syntax (without space) should work the same as the `function () { ... }` syntax (with a space) when used as an argument in an evaluation method.

### Reality

The `function() { ... }` syntax (without space) throws an `Unexpected token 'function'` error when used as an argument in an evaluation method.

### Puppeteer version

24.22.1

### Node version

22.16.0

### Package manager

npm

### Package manager version

10.9.2

### Operating system

macOS

OrKoN:
tsc was changing the whitespaces in test functions so tests were passing while the new approach would not work in non-tsc projects. This PR moves the fixture to be a plain js file and fixes the regex.

fixes https://github.com/puppeteer/puppeteer/issues/14244

### DIFF
--- a/packages/puppeteer-core/function-fixture.mjs
+++ b/packages/puppeteer-core/function-fixture.mjs
@@ -1,0 +1,50 @@
+/**
+ * @license
+ * Copyright 2025 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+// prettier-ignore
+class K1 { m() {} }
+// prettier-ignore
+class K2 { m () {} }
+// prettier-ignore
+class K3 { async m() {} }
+// prettier-ignore
+class K4 { async m () {} }
+
+// prettier-ignore
+export const variations = [
+  // Non-arrow functions.
+  function() {},
+  function () {},
+  function x() {},
+  function x () {},
+  async function() {},
+  async function () {},
+  async function x() {},
+  async function x () {},
+  // Concise methods.
+  new K1().m,
+  new K2().m,
+  new K3().m,
+  new K4().m,
+  // Arrow functions.
+  () => {},
+  // @ts-expect-error irrelevant
+  (a) => {a;},
+  // @ts-expect-error irrelevant
+  a => {a;},
+  async () => {},
+  // @ts-expect-error irrelevant
+  async (a) => {a;},
+  // @ts-expect-error irrelevant
+  async a => {a;},
+  // Generator functions
+  function*a() {},
+  function *a() {},
+  function* a() {},
+  // Async generator functions
+  async function*a() {},
+  async function *a() {},
+  async function* a() {},
+];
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/util/Function.test.ts
```diff
@@ -35,49 +35,9 @@ describe('Function', function () {
...
-        function* a() {},
-        // Async generator functions
-        async function*a() {},
-        async function *a() {},
-        async function* a() {},
-      ];
-
+    it('should work', async () => {
+      // eslint-disable-next-line @typescript-eslint/no-require-imports
+      const {variations} = require('../../../../function-fixture.cjs');
```
Lightning00Blade: Why can't we just use `await import`?

packages/puppeteer-core/function-fixture.mjs
```diff
@@ -0,0 +1,50 @@
+/**
+ * @license
+ * Copyright 2025 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+// prettier-ignore
```
Lightning00Blade, line 6: Can we try - https://prettier.io/docs/ignore#range-ignore

Lightning00Blade: Disregard this I think this is MarkDown only.

Lightning00Blade (approved):

packages/puppeteer-core/src/util/Function.test.ts
OrKoN: We run the cjs version in tests so import wouldn’t work.

Lightning00Blade: https://nodejs.org/api/esm.html#import-statements
They work everywhere when its a dynamic one.

OrKoN: Hm.. I tried it before switching to cjs but trying it again and it does work so maybe it was some other issue.

PR #14252, opened 2025-09-26
docs: fix some code and remove IIFE

### LINKED ISSUE #14080: Docs issues

Lightning00Blade:
Closes #14128 
Closes #14129
Closes: #14080

### DIFF
--- a/docs/guides/page-interactions.md
+++ b/docs/guides/page-interactions.md
@@ -127,19 +127,16 @@ other locator functions such as `.click()` or `.fill()` on the function locator.
 ```ts
 await page
   .locator(() => {
-    let resolve!: (node: HTMLCanvasElement) => void;
-    const promise = new Promise(res => {
-      return (resolve = res);
-    });
-    const observer = new MutationObserver(records => {
-      for (const record of records) {
-        if (record.target instanceof HTMLCanvasElement) {
-          resolve(record.target);
+    return new Promise(res => {
+      const observer = new MutationObserver(records => {
+        for (const record of records) {
+          if (record.target instanceof HTMLCanvasElement) {
+            resolve(record.target);
+          }
         }
-      }
+      });
+      observer.observe(document, {childList: true, subtree: true});
     });
-    observer.observe(document);
-    return promise;
   })
   .wait();
 ```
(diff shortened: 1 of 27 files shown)


docs/guides/page-interactions.md
```diff
@@ -127,19 +127,16 @@ other locator functions such as `.click()` or `.fill()` on the function locator.
...
   .locator(() => {
-    let resolve!: (node: HTMLCanvasElement) => void;
-    const promise = new Promise(res => {
-      return (resolve = res);
-    });
-    const observer = new MutationObserver(records => {
-      for (const record of records) {
-        if (record.target instanceof HTMLCanvasElement) {
-          resolve(record.target);
+    return new Promise(res => {
```
Lightning00Blade, line 130: Rewrote the function so it's typesafe but can be used in JS as well.

packages/puppeteer-core/src/api/Dialog.ts
```diff
@@ -18,16 +18,14 @@ import {assert} from '../util/assert.js';
...
- *   page.evaluate(() => alert('1'));
- * })();
+ * const browser = await puppeteer.launch();
+ * const page = await browser.newPage();
+ * page.on('dialog', async dialog => {
+ *   console.log(dialog.message());
+ *   await dialog.dismiss();
+ *   await browser.close();
+ * });
+ * page.evaluate(() => alert('1'));
```
OrKoN: nit: let's add await, I think it makes no difference but at least to not set a bad example

OrKoN (approved):

docs/guides/page-interactions.md
OrKoN: ack

PR #14263, opened 2025-09-29
feat: introduce `webDriverBiDiOnly` protocol

sadym-chromium:
The `webDriverBiDiOnly` is intended to be used with Chrome, but without using BiDi+, only using pure WebDriver BiDi protocol. Details: http://goto.google.com/webdriver-bidi:pure-bidi-puppeteer


OrKoN (changes requested):
Thanks for the PR but I do not think we want this to be a user-facing features: let's disable the bidi+ via an environmental variable or a different internal option.

test/TestExpectations.json
```diff
@@ -209,6 +209,13 @@
     "expectations": ["SKIP"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
+  {
```
OrKoN, line 212: I do not think we should maintain expectations for bidiPlus: false here. IIUC we only discussed using for the dashboard.

packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -93,7 +93,7 @@ export class BidiBrowser extends Browser {
     });
 
     await session.subscribe(
-      (session.capabilities.browserName.toLocaleLowerCase().includes('firefox')
+      (opts.cdpConnection === undefined
```
OrKoN, line 96: let's check an env var here like `process.env['PUPPETEER_WEBDRIVER_BIDI_ONLY'] === true`

test/TestExpectations.json
sadym-chromium: Some BiDi only tests will timeout, and skipping them would speed up execution.

OrKoN: I see. In any case, let's maintain these expectations in the dashboard repo and not in Puppeteer.

PR #14286, opened 2025-10-06
fix: support multiple paths for system executables on Windows and WSL

Lightning00Blade:
Implement more robust logic when working with Windows and WSL,
so that we try to launch the browser installed in Windows and not WSL.

### DIFF
--- a/packages/browsers/src/browser-data/browser-data.ts
+++ b/packages/browsers/src/browser-data/browser-data.ts
@@ -221,7 +221,7 @@ export async function createProfile(
 /**
  * @public
  *
- * Get's the most first resolved system path
+ * Get's the first resolved system path
  */
 export function resolveSystemExecutablePath(
   browser: Browser,
@@ -244,8 +244,8 @@ export function resolveSystemExecutablePath(
 /**
  * @internal
  *
- * Tries to find multiple paths that the executable may be.
- * Return them priority based on heuristics.
+ * Returns multiple paths where the executable may be located at on the current system
+ * ordered by likelihood (based on heuristics).
  */
 export function resolveSystemExecutablePaths(
   browser: Browser,
--- a/packages/browsers/src/browser-data/browser-data.ts
+++ b/packages/browsers/src/browser-data/browser-data.ts
@@ -220,6 +220,8 @@ export async function createProfile(
 
 /**
  * @public
+ *
+ * Get's the most first resolved system path
  */
 export function resolveSystemExecutablePath(
   browser: Browser,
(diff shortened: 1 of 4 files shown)


packages/browsers/src/browser-data/chrome.ts
```diff
@@ -154,46 +154,83 @@ export async function resolveBuildId(
...
+    case ChromeReleaseChannel.CANARY:
+      suffix = '\\Google\\Chrome SxS\\Application\\chrome.exe';
+      break;
+    case ChromeReleaseChannel.DEV:
+      suffix = '\\Google\\Chrome Dev\\Application\\chrome.exe';
+      break;
+  }
+
+  return locations.map(l => {
+    return `${l}${suffix}`;
```
OrKoN, line 188: should we use path.join()?

packages/browsers/src/browser-data/chrome.ts
```diff
@@ -154,46 +154,83 @@ export async function resolveBuildId(
...
+function getChromeWindowsLocation(
+  channel: ChromeReleaseChannel,
+): [string, ...string[]] {
+  // TODO: Investigate if we can use the values in the Registry as well
+  const locations = [
+    process.env['PROGRAMFILES'],
+    process.env['ProgramW6432'],
+    process.env['ProgramFiles(x86)'],
+  ].filter(l => {
+    return typeof l === 'string';
```
OrKoN: should we check for empty values? should check that the path is absolute?

packages/browsers/src/browser-data/browser-data.ts
```diff
@@ -220,6 +220,8 @@ export async function createProfile(
 
 /**
  * @public
+ *
+ * Get's the most first resolved system path
```
OrKoN:
```suggestion
 * Get's the first resolved system path
```

OrKoN:
> TODO: Add tests.

Do you still plan to add some tests?

packages/browsers/src/browser-data/chrome.ts
OrKoN: This was resolved without a comment and path.join() is not used. Could you please explain why not?

packages/browsers/src/browser-data/browser-data.ts
```diff
@@ -235,7 +237,31 @@ export function resolveSystemExecutablePath(
...
-      return chrome.resolveSystemExecutablePath(platform, channel);
+      return chrome.resolveSystemExecutablePaths(platform, channel)[0];
+  }
+}
+
+/**
+ * @internal
+ *
+ * Tries to find multiple paths that the executable may be.
+ * Return them priority based on heuristics.
```
OrKoN:
```suggestion
 * Returns multiple paths where the executable may be located at on the current system ordered by likelihood (based on heuristics).
```

OrKoN (approved):

packages/browsers/src/browser-data/chrome.ts
Lightning00Blade: Sorry, I starts breaking when used from WSL as there the join thinks it needs to use `/` in between the two.
I will add a comment.
There is no path.joinForSystem

OrKoN: I think you can use https://nodejs.org/api/path.html#pathposix or https://nodejs.org/api/path.html#pathwin32 depending on what you need?

Lightning00Blade: Still looks wrong in REPL node 
image

Let's keep it as it is for now as I know this work correctly atm, at least on my machine.

Lightning00Blade:
For the test we can't mock the node modules from mocha.

packages/browsers/test/src/chrome/chrome-data.spec.ts
```diff
@@ -78,33 +78,43 @@ describe('Chrome', () => {
     );
   });
 
-  it('should resolve system executable path', () => {
+  it.only('should resolve system executable path', () => {
```
Lightning00Blade:
```suggestion
  it('should resolve system executable path', () => {
```

PR #14313, opened 2025-10-13
fix: revert Accessibility heuristic focus changes

### LINKED ISSUE #14284: [Bug]: `accessibility.snapshot` unexpectedly pruning more elements in recent update

```TypeScript
/** Extracted from a helper file, normally imported. */
async function getAccessibleName(page, $element) {
  return (
    await page.accessibility.snapshot({
      root: $element
    })
  ).name
}

describe('accessible name', () => {
  it('Should set the accessible name of the show/hide section buttons', async () => {
    // For the code of the `render` function, see:
    // https://github.com/alphagov/govuk-frontend/blob/5024297aee7a01a77364b56cb08ec9130ccaf189/shared/helpers/puppeteer.js#L88-L219
    //
    // For the HTML it outputs, see:
    // https://govuk-frontend-review.herokuapp.com/components/accordion/preview
    //
    // The code that defines the `examples` object has been omitted for brevity.
    await render(page, 'accordion', examples.default)

    const $element = await page.$('.govuk-accordion__section-button')

    await expect(getAccessibleName(page, $element)).resolves.toBe(
      'Section A , Show this section'
    )
  })
})
```

### Background

Upon updating from Puppeteer 24.19.0 to 24.22.3, we found that some of our accessibility tests began to unexpectedly fail with the following error:

```
    expect(received).resolves.toBe()

    Received promise rejected instead of resolved
    Rejected to value: [TypeError: Cannot read properties of null (reading 'name')]

      24 |     const $element = await page.$('.govuk-accordion__section-button')
      25 |
    > 26 |     await expect(getAccessibleName(page, $element)).resolves.toBe(
         |           ^
      27 |       'Section A , Show this section'
      28 |     )
      29 |   })
```

After some digging around, we tried setting `interestingOnly` to `false` when using `accessibility.snapshot` and found that our tests began passing again.

This was unexpected, as [the documentation for `accessibility.snapshot`](https://pptr.dev/api/puppeteer.accessibility.snapshot#remarks) remarks that it prunes "nodes that go unused on most platforms and by most screen readers". However, our test is against the `aria-label` attribute of a `<button>` element, which is _definitely_ used by a number of platforms and screen readers.

For context, the `.govuk-accordion__section-button` HTML looks like this at the time the test is ran:

```html
<button type="button" aria-controls="default-example-content-1" class="govuk-accordion__section-button" aria-expanded="false" aria-label="Section A , Show this section">
  […]
</button>
```

Quick links for the code in GOV.UK Frontend:
- [Accordion component's Puppeteer test file](https://github.com/alphagov/govuk-frontend/blob/5024297aee7a01a77364b56cb08ec9130ccaf189/packages/govuk-frontend/src/govuk/components/accordion/accordion.puppeteer.test.js)
- [Puppeteer helper file](https://github.com/alphagov/govuk-frontend/blob/5024297aee7a01a77364b56cb08ec9130ccaf189/shared/helpers/puppeteer.js) — contains `getAccessibleName` and other helper functions

This appears to be OS agnostic, having been reproduced on macOS, Windows, and Ubuntu Linux. 

### Expectation

We would expect that a `<button>` with an `aria-label` not be pruned from the accessible nodes list, as it is a semantically significant node that is widely used by assistive technologies.

If the definition of an 'interesting' node were to be changed, we would expect this to not happen in a minor release version.

### Reality

Some aspect of how an 'interesting' node is defined has unexpectedly changed between Puppeteer versions 24.19.0 and 24.22.3.

We weren't sure if https://github.com/puppeteer/puppeteer/pull/14221 was related to this, but it appears to be the only interim change that touches on the same area. It may also be something else (such as a bump in the Chromium version) has caused this to change.

https://github.com/alphagov/govuk-frontend/blob/5024297aee7a01a77364b56cb08ec9130ccaf189/puppeteer.config.js

### Puppeteer version

24.22.3

### Node version

22.20.0

### Package manager

npm

### Package manager version

10.9.3

### Operating system

Linux

Lightning00Blade:
Closes #14284

### DIFF
--- a/packages/puppeteer-core/src/cdp/Accessibility.ts
+++ b/packages/puppeteer-core/src/cdp/Accessibility.ts
@@ -319,6 +319,7 @@ class AXNode {
   #name: string;
   #role: string;
   #ignored: boolean;
+  #cachedHasFocusableChild?: boolean;
   #realm: Realm;
 
   constructor(realm: Realm, payload: Protocol.Accessibility.AXNode) {
@@ -359,6 +360,19 @@ class AXNode {
       role === 'InlineTextBox' ||
       role === 'StaticText'
     );
+  }
+
+  #hasFocusableChild(): boolean {
+    if (this.#cachedHasFocusableChild === undefined) {
+      this.#cachedHasFocusableChild = false;
+      for (const child of this.children) {
+        if (child.#focusable || child.#hasFocusableChild()) {
+          this.#cachedHasFocusableChild = true;
+          break;
+        }
+      }
+    }
+    return this.#cachedHasFocusableChild;
   }
 
   public find(predicate: (x: AXNode) => boolean): AXNode | null {
@@ -404,6 +418,10 @@ class AXNode {
         return true;
       default:
         break;
+    }
+
+    if (this.#hasFocusableChild()) {
+      return false;
     }
 
     if (this.#role === 'heading' && this.#name) {
--- a/test/src/accessibility.spec.ts
+++ b/test/src/accessibility.spec.ts
@@ -826,6 +826,25 @@ describe('Accessibility', function () {
           ],
         });
       });
+      it('should work with nested button inside h1 with interestingOnly:true', async () => {
+        const {page} = await getTestState();
+
+        await page.setContent(html`
+          <main>
+            <h2>
+              <button>My Button</button>
+            </h2>
+          </main>
+        `);
+
+        using button = (await page.$('button'))!;
+        expect(await page.accessibility.snapshot({root: button})).toMatchObject(
+          {
+            role: 'button',
+            name: 'My Button',
+          },
+        );
+      });
     });
 
     describe('elementHandle()', () => {


packages/puppeteer-core/src/cdp/Accessibility.ts
```diff
@@ -406,6 +420,10 @@ class AXNode {
         break;
     }
 
+    if (this.#hasFocusableChild()) {
+      return false;
+    }
+
     if (this.#role === 'heading' && this.#name) {
```
Lightning00Blade, line 427: This check here is also needed as else we don't resolve the text on the heading level but on the SimpleText box.

OrKoN (approved):

packages/puppeteer-core/src/cdp/Accessibility.ts
OrKoN: Ack

PR #14315, opened 2025-10-14
fix(webdriver): response should be created on responseStarted

OrKoN:
We need to update the waiting code to wait for success correctly.


OrKoN:
Flaky: the puppeteer's response waiting code does not detect responses correctly with this change. Was flaky on MacOS specifically in all tests that do navigation and assert responses.

sadym-chromium:
Implemented in https://github.com/puppeteer/puppeteer/pull/14513

PR #14340, opened 2025-10-17
feat(webdriver): support `fetchPostData` 

sadym-chromium:
Rely on BiDi data collectors.

### DIFF
--- a/packages/puppeteer-core/src/bidi/core/Request.ts
+++ b/packages/puppeteer-core/src/bidi/core/Request.ts
@@ -196,8 +196,7 @@ export class Request extends EventEmitter<{
   }
 
   get hasPostData(): boolean {
-    // @ts-expect-error non-standard attribute.
-    return this.#event.request.bodySize > 0;
+    return (this.#event.request.bodySize ?? 0) > 0;
   }
 
   async continueRequest({
--- a/packages/puppeteer-core/src/bidi/core/Request.ts
+++ b/packages/puppeteer-core/src/bidi/core/Request.ts
@@ -239,6 +239,9 @@ export class Request extends EventEmitter<{
   }
 
   async fetchPostData(): Promise<string | undefined> {
+    if (!this.hasPostData) {
+      return undefined;
+    }
     if (!this.#requestBodyPromise) {
       this.#requestBodyPromise = (async () => {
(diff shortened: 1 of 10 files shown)


test/src/network.spec.ts
```diff
@@ -241,6 +241,7 @@ describe('network', function () {
     });
   });
 
+  // TODO: remove test for deprecated `Request.postData` in favor of `Request.fetchPostData`.
   describe('Request.postData', function () {
```
sadym-chromium: @OrKoN at which point should we delete `postData` and the related tests? In 2 major releases after deprecating?

OrKoN: we can delete them once we delete the implementation. Which could happen in the next major release but we also have a bunch of other pending breaking changes that should be included there and for now we have no timeline. You can move tests in to CDP folder for now.

packages/puppeteer-core/src/bidi/core/Request.ts
```diff
@@ -196,7 +197,7 @@ export class Request extends EventEmitter<{
 
   get hasPostData(): boolean {
     // @ts-expect-error non-standard attribute.
```
Lightning00Blade: Can we remove this `@ts-expect-error` or at least update the comment on why we need it.

packages/puppeteer-core/src/bidi/core/Request.ts
```diff
@@ -258,7 +282,7 @@ export class Request extends EventEmitter<{
             )
           ) {
             throw new ProtocolError(
-              'Could not load body for this request. This might happen if the request is a preflight request.',
+              'Could not load response body for this request. This might happen if the request is a preflight request.',
```
Lightning00Blade, line 284: There is a error on the CDP impl with the same text please keep update it as well.

packages/puppeteer-core/src/bidi/core/Request.ts
sadym-chromium: done

packages/puppeteer-core/src/bidi/core/Request.ts
sadym-chromium: done

Lightning00Blade (approved):

Issue #14333, opened 2025-10-16
[Bug]: Headers not getting updated using page.setRequestInterception(true) for firefox


shukrash:
### Minimal, reproducible example

```TypeScript
import puppeteer from "puppeteer-core";

export async function try() {
  const browser = await puppeteer.launch({
    browser: 'firefox',
    executablePath: '/tmp/browser/firefox/firefox'
  });

  const page = await browser.newPage();
  await page.setRequestInterception(true);

  page.on("request", request => {
    const headers = request.headers() || {};
    headers['testung'] = 'abceded';
    console.log(`[request] ${JSON.stringify(headers)}`);
    request.continue({ headers });
  });

  page.on('response', response => {
    console.log(`[response] ${response.status()} ${JSON.stringify(response.request().headers())}`);
  });

  const response = await page.goto("https://support.mozilla.org");
  await browser.close();
}

export async function handler() {
  return await try();
};
```

### Background

I am trying to add additional headers to requests made during page load.
I am using setRequestInterception. 
In the above code, when I print the headers before request continue, the headers have the new headers. However, the request object in response does not.

[request] {\"host\":\"location.services.mozilla.com\",\"user-agent\":\"Mozilla/5.0 (X11; Linux x86_64; rv:141.0) Gecko/20100101 Firefox/141.0\",\"accept\":\"*/*\",\"accept-language\":\"en-US,en;q=0.5\",\"accept-encoding\":\"gzip, deflate, br, zstd\",\"referer\":\"https://support.mozilla.org/\",\"origin\":\"https://support.mozilla.org\",\"connection\":\"keep-alive\",\"sec-fetch-dest\":\"empty\",\"sec-fetch-mode\":\"cors\",\"sec-fetch-site\":\"cross-site\",\"testung\":\"abceded\"

[response] 200 {\"host\":\"location.services.mozilla.com\",\"user-agent\":\"Mozilla/5.0 (X11; Linux x86_64; rv:141.0) Gecko/20100101 Firefox/141.0\",\"accept\":\"*/*\",\"accept-language\":\"en-US,en;q=0.5\",\"accept-encoding\":\"gzip, deflate, br, zstd\",\"referer\":\"https://support.mozilla.org/\",\"origin\":\"https://support.mozilla.org\",\"connection\":\"keep-alive\",\"sec-fetch-dest\":\"empty\",\"sec-fetch-mode\":\"cors\",\"sec-fetch-site\":\"cross-site\"}

### Expectation

[...]

PR #14341, opened 2025-10-18
fix: request.headers() should not allow mutating data

OrKoN:
Since we have cooperative interception, mutating the headers like this could lead to bugs that fairly difficult to debug. Changed the CDP method to behave like WebDriver BiDi's and stop allowing accidental mutations of the underlying headers map.

Refs https://github.com/puppeteer/puppeteer/issues/14333

### DIFF
--- a/packages/puppeteer-core/src/bidi/HTTPRequest.ts
+++ b/packages/puppeteer-core/src/bidi/HTTPRequest.ts
@@ -181,6 +181,7 @@ export class BidiHTTPRequest extends HTTPRequest {
   }
 
   override headers(): Record<string, string> {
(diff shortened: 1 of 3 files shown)


Lightning00Blade (approved):
LGTM, but should we change the Type to `ReadOnly<>` as well, I don't think this braking change as it does nothing to the actual request

OrKoN:
> LGTM, but should we change the Type to `ReadOnly<>` as well, I don't think this braking change as it does nothing to the actual request

it might make use builds and it is not necessarily that we want to return a read-only object. It is fine for users to mutate it for their own purposes but we do not want to affect the internal state.

PR #14348, opened 2025-10-20
feat(webdriver): get detailed initiator data from CDP if available

sadym-chromium:
And filter out `goog:` data from events in `PUPPETEER_WEBDRIVER_BIDI_ONLY` mode.

### DIFF
--- a/packages/puppeteer-core/src/bidi/core/Request.ts
+++ b/packages/puppeteer-core/src/bidi/core/Request.ts
@@ -154,7 +154,15 @@ export class Request extends EventEmitter<{
     return this.#event.request.request;
   }
   get initiator(): Bidi.Network.Initiator | undefined {
-    return this.#event.initiator;
+    return {
+      ...this.#event.initiator,
+      type: this.#event.initiator?.type ?? 'other',
+      // Initiator URL is not specified in BiDi.
+      // @ts-expect-error non-standard property.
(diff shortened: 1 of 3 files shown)


Lightning00Blade:
Nit: the title should be `feat(webdriver): ...` as that is how you signify component in conventional commits 
`<type>[optional scope]: <description>`

packages/puppeteer-core/src/bidi/core/Request.ts
```diff
@@ -154,7 +154,15 @@ export class Request extends EventEmitter<{
     return this.#event.request.request;
   }
   get initiator(): Bidi.Network.Initiator | undefined {
-    return this.#event.initiator;
+    return {
+      ...this.#event.initiator,
+      type: this.#event.initiator?.type ?? 'other',
```
Lightning00Blade: Why is this needed? Should we not do that in `chromium-bidi` ?

sadym-chromium: Not really. In BiDi, the `initiatorType` can be `null`. But it is done in [`bidi/HTTPRequest`](https://github.com/puppeteer/puppeteer/blob/ebe76d3f0c0dc2392f453331224afa41cfc78a70/packages/puppeteer-core/src/bidi/HTTPRequest.ts#L213). Removed.

Lightning00Blade (approved):

PR #14351, opened 2025-10-21
fix: shared workers work without a detach call now

OrKoN:
We can land it once crrev.com/c/7037151 is in stable.


Lightning00Blade (approved):

OrKoN:
landed via roll

PR #14355, opened 2025-10-21
feat!: move packages to ESM only

Lightning00Blade:
This removes the need to build CJS.
We can clean up more of the code, but as Draft for now this proves that we can do this.


test/installation/src/puppeteer-typescript.spec.ts
```diff
@@ -12,20 +12,20 @@ import {configureSandbox} from './sandbox.js';
 import {execFile, readAsset} from './util.js';
 
 // Skipping this test on Windows as windows runners are much slower.
-(platform() === 'win32' ? describe.skip : describe)(
+(platform() === 'win32' ? describe.skip : describe.only)(
```
OrKoN: remove .only?

OrKoN (approved):

PR #14358, opened 2025-10-21
test: move to ESM

Lightning00Blade:
This is a subset of the changes https://github.com/puppeteer/puppeteer/pull/14355
But should be mergable as is.

This also drops `eslint-plugin-rulesdir` as it does not support ESM custom rules.


test/src/tracing.spec.ts
```diff
@@ -105,34 +103,6 @@ describe('Tracing', function () {
     expect(trace).toBeTruthy();
   });
 
-  it('should return undefined in case of Buffer error', async () => {
```
Lightning00Blade, line 108: We can stub thing from ESM so not sure how to read this test.

packages/browsers/test/src/chrome/cli.spec.ts
```diff
@@ -79,16 +80,42 @@ describe('Chrome CLI', function () {
...
-    await new CLI(tmpDir).run([
-      'npx',
-      '@puppeteer/browsers',
-      'install',
-      `chrome@latest`,
-      `--path=${tmpDir}`,
-      '--platform=linux',
-      `--base-url=${getServerUrl()}`,
-    ]);
+  describe('with mocked version url', () => {
```
Lightning00Blade, line 83: This part needs review.

packages/browsers/test/src/firefox/cli.spec.ts
```diff
@@ -60,30 +60,42 @@ describe('Firefox CLI', function () {
...
-    await new CLI(tmpDir).run([
-      'npx',
-      '@puppeteer/browsers',
-      'install',
-      `firefox@latest`,
-      `--path=${tmpDir}`,
-      '--platform=linux',
-      `--base-url=${getServerUrl()}`,
-    ]);
+  describe('with mocked version url', () => {
```
Lightning00Blade, line 63: This part needs review.

packages/browsers/src/browser-data/chrome.ts
```diff
@@ -67,13 +67,20 @@ export function relativeExecutablePath(
   }
 }
 
+let baseVersionUrl = 'https://googlechromelabs.github.io/chrome-for-testing';
```
Lightning00Blade, line 70: This part needs review.

packages/browsers/src/browser-data/firefox.ts
```diff
@@ -164,6 +164,16 @@ export enum FirefoxChannel {
   NIGHTLY = 'nightly',
 }
 
+let baseVersionUrl = 'https://product-details.mozilla.org/1.0';
```
Lightning00Blade, line 167: This part needs review.

test/smoke/run-smoke.mjs
```diff
@@ -0,0 +1,66 @@
+/**
```
Lightning00Blade, line 1: This part needs review.

test/src/stacktrace.spec.ts
```diff
@@ -36,7 +35,12 @@ describe('Stack trace', function () {
     expect(error.name).toEqual('Error');
     expect(error.message).toEqual('Test');
     assert(error.stack);
-    error.stack = error.stack.replace(new RegExp(FILENAME, 'g'), '<filename>');
+    // console.log(import.meta.filename);
```
Lightning00Blade, line 38: TODO: Remove

.github/workflows/ci.yml
```diff
@@ -550,3 +550,33 @@ jobs:
       - if: ${{ needs.browsers-tests.result != 'success' && contains(fromJSON(needs.check-changes.outputs.changes), 'browsers') }}
         run: 'exit 1'
       - run: 'exit 0'
+
+  puppeteer-smoke-tests:
```
Lightning00Blade, line 554: This part needs review.

test/src/tracing.spec.ts
OrKoN: We can stub or we cannot stub?

packages/browsers/test/src/chrome/cli.spec.ts
```diff
@@ -79,16 +80,42 @@ describe('Chrome CLI', function () {
...
+    before(() => {
+      chrome.changeBaseVersionUrlForTesting(getServerUrl());
```
OrKoN, line 85: why do we need changeBaseVersionUrlForTesting?

packages/browsers/package.json
```diff
@@ -42,6 +42,15 @@
         "lib/esm/package.json"
       ]
     },
+    "generate:package-json:test": {
```
Lightning00Blade, line 45: This part needs review, needed to mark test as ESM.

packages/browsers/test/src/firefox/cli.spec.ts
```diff
@@ -60,30 +60,42 @@ describe('Firefox CLI', function () {
...
+    before(() => {
+      firefox.changeBaseVersionUrlForTesting(getServerUrl());
```
OrKoN, line 65: do we need changeBaseVersionUrlForTesting? `--base-url` is specified in the CLI call

test/src/cdp/devtools.spec.ts
```diff
@@ -98,13 +98,13 @@ describe('DevTools', function () {
     });
 
     const devtoolsPageTarget = await browser.waitForTarget(target => {
-      return target.type() === 'other';
+      return target.type() === 'other' && target.url().startsWith('devtools');
```
Lightning00Blade, line 101: This needs review.

test/smoke/src/import-in-cjs.cjs
```diff
@@ -0,0 +1,10 @@
+/**
+ * @license
+ * Copyright 2025 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+const puppeteer = require('puppeteer');
```
OrKoN, line 6: could this be an installation test?

test/src/tracing.spec.ts
Lightning00Blade: We can't stub ESM modules.

packages/browsers/test/src/chrome/cli.spec.ts
Lightning00Blade: The idea is to test the flow without relying on external servers.
This also allows us to specify the version reducing the external downloads.

packages/browsers/test/src/firefox/cli.spec.ts
Lightning00Blade: That param only controls where we download the Browser from, but not where we get the versions from.

test/smoke/src/import-in-cjs.cjs
Lightning00Blade: Actually this should be handled by it, we only need to move the version we test against for those test to the min supported version.

OrKoN: I would prefer not add a new type of tests if possible.

Lightning00Blade: Will downgrade the NodeJS version of the Installation test to v20.19 which will be the next min supported version fro v25

OrKoN (approved):

PR #14363, opened 2025-10-23
feat: add window options to newPage()

OrKoN:
Allow opening a page in a tab or a window.

### DIFF
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -358,7 +358,7 @@ export class CdpBrowser extends BrowserBase {
     const {targetId} = await this.#connection.send('Target.createTarget', {
       url: 'about:blank',
       browserContextId: contextId || undefined,
-      // Works around buggy logic in CDP if newWindow is false.
+      // Works around crbug.com/454825274.
       newWindow: hasTargets && options?.type === 'window' ? true : undefined,
     });
     const target = (await this.waitForTarget(t => {
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -8,7 +8,7 @@ import type {ChildProcess} from 'node:child_process';
 
 import type {Protocol} from 'devtools-protocol';
 
-import type {DebugInfo} from '../api/Browser.js';
+import type {CreatePageOptions, DebugInfo} from '../api/Browser.js';
 import {
   Browser as BrowserBase,
   BrowserEvent,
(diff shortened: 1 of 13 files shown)


packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -347,10 +347,19 @@ export class CdpBrowser extends BrowserBase {
...
+  ): Promise<Page> {
+    const hasTargets =
+      this.targets().filter(t => {
+        return t.browserContext().id === contextId;
+      }).length > 0;
     const {targetId} = await this.#connection.send('Target.createTarget', {
       url: 'about:blank',
       browserContextId: contextId || undefined,
+      // Works around buggy logic in CDP if newWindow is false.
+      newWindow: hasTargets && options?.type === 'window' ? true : undefined,
```
Lightning00Blade, line 362: Is there a crbug we can link to?

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -347,10 +347,19 @@ export class CdpBrowser extends BrowserBase {
     return await this.#defaultContext.newPage();
   }
 
-  async _createPageInContext(contextId?: string): Promise<Page> {
+  async _createPageInContext(
+    contextId?: string,
+    options?: CreatePageOptions,
+  ): Promise<Page> {
+    const hasTargets =
+      this.targets().filter(t => {
```
Lightning00Blade, line 355: Out of curiosity I understand that needs to check if there are already open pages,
is it possible there can be other targets and then this creation will fail?

Lightning00Blade (approved):

packages/puppeteer-core/src/cdp/Browser.ts
OrKoN: no, I don't think anyone got time to fix it.

packages/puppeteer-core/src/cdp/Browser.ts
OrKoN: I think if there is any target for the context, the browser profile object in chromium would be instantiated and newWindow argument would work. It only does not work if there is nothing in the browser context at all.

packages/puppeteer-core/src/cdp/Browser.ts
Lightning00Blade: Let's file a bug at least, this sound very bad for anyone using this, we can at least add description to the pdl.

packages/puppeteer-core/src/cdp/Browser.ts
Lightning00Blade: Thanks for the info!

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -347,10 +347,19 @@ export class CdpBrowser extends BrowserBase {
...
+        return t.browserContext().id === contextId;
+      }).length > 0;
     const {targetId} = await this.#connection.send('Target.createTarget', {
       url: 'about:blank',
       browserContextId: contextId || undefined,
+      // Works around buggy logic in CDP if newWindow is false.
```
OrKoN:
```suggestion
      // Works around crbug.com/454825274.
```

PR #14367, opened 2025-10-24
test: do not use example.com

OrKoN:


test/installation/assets/puppeteer/basic.js
```diff
@@ -8,7 +8,7 @@ import puppeteer from 'puppeteer';
 (async () => {
   const browser = await puppeteer.launch();
   const page = await browser.newPage();
-  await page.goto('http://example.com');
+  await page.goto('data:text/html;charset=utf-8,<h1>example');
```
Lightning00Blade: With this we enter quirks mode now, which should be fined.
Nit: But can we at least close the `<h1>` else we leave it to the browser parser to figure out, which can be less deterministic with version changes.

Lightning00Blade (approved):

PR #14369, opened 2025-10-24
refactor: extract things required for ESM migration

Lightning00Blade:


packages/puppeteer/src/getConfiguration.ts
```diff
@@ -119,7 +119,7 @@ export const getConfiguration = (): Configuration => {
   const result = cosmiconfigSync('puppeteer', {
     searchStrategy: 'global',
   }).search();
-  const configuration: Configuration = result ? result.config : {};
+  const configuration: Configuration = result ? {...result.config} : {};
```
Lightning00Blade, line 122: For some reason in ESM the returned object can't be modified.

packages/testserver/src/index.ts
```diff
@@ -82,9 +85,20 @@ export class TestServer {
     server.#server.once('listening', res);
     server.#server.listen(0);
     await promise;
+
+    TestServer.setupProps(server, 'https');
+
     return server;
   }
 
+  static setupProps(server: TestServer, protocol = 'http'): void {
```
Lightning00Blade, line 94: Let's the server set this up itself.

test/installation/src/util.ts
```diff
@@ -4,29 +4,51 @@
...
-import {execFile as execFileAsync} from 'node:child_process';
+import type {ExecFileException, ExecFileOptions} from 'node:child_process';
+import {execFile as nodeExecFile} from 'node:child_process';
 import {readFile} from 'node:fs/promises';
 import {join} from 'node:path';
-import {promisify} from 'node:util';
 
 import {ASSETS_DIR} from './constants.js';
 
-const nodeExecFile = promisify(execFileAsync);
```
Lightning00Blade, line 15: This was working strangely and thing were not getting properly returned for stdout/stderr.

test/src/mocha-utils.ts
```diff
@@ -111,22 +111,30 @@ defaultBrowserOptions.args!.push(
 );
 defaultBrowserOptions.args!.push(`--force-device-scale-factor=1`);
 
-if (defaultBrowserOptions.executablePath) {
-  console.warn(
-    `WARN: running ${product} tests with ${defaultBrowserOptions.executablePath}`,
-  );
-  if (!fs.existsSync(defaultBrowserOptions.executablePath)) {
-    throw new Error(
```
Lightning00Blade, line 119: We throw on module init but Mocha with ESM does not like that, so we throw now on Hook.

packages/puppeteer/src/getConfiguration.ts
OrKoN: that's strange, where do we modify the configuration? did you find why it's not modifiable? does the cosmic config return a different object in ESM builds?

test/src/mocha-utils.ts
```diff
@@ -371,40 +370,27 @@ const browserNotClosedError = new Error(
...
-          'puppeteer',
-          'lib',
-          'esm',
-          'puppeteer',
-          'puppeteer.js',
-        ),
-      );
-    }
-
-    try {
```
Lightning00Blade, line 399: We should not make this error silenced.

packages/puppeteer/src/getConfiguration.ts
Lightning00Blade: Let me check with a smaller repo, as there were multiple issues

Lightning00Blade: Closing the loop this is needed as config can be `export default {}` but ESM exports like this make the object non extensible.

OrKoN (approved):

PR #14372, opened 2025-10-24
build(deps): bump mime to v4

Lightning00Blade:
After we finally switched to ESM test we can bump this version up.

### DIFF
--- a/test/package.json
+++ b/test/package.json
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "diff": "8.0.2",
+    "mime": "4.1.0",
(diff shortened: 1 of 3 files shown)


test/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "diff": "8.0.2",
+    "mime": "4.1.0",
```
Lightning00Blade, line 29: We were pooling this as transitive dependency rather then a direct one.

OrKoN (approved):

PR #14383, opened 2025-10-29
docs: fix missing supported browser versions

Lightning00Blade:


packages/puppeteer-core/src/templates/version.ts.tmpl
```diff
@@ -1,4 +0,0 @@
-/**
```
Lightning00Blade, line 1: I missed this file while cleaning up the previous implementation

OrKoN (approved):

OrKoN:
Thanks!

PR #14386, opened 2025-10-30
test: fix TargetManager tests failing in Canary

OrKoN:
This PR adapts the test to account for possible pre-render targets that seems to be created in Canary Chrome (headful builds).

### DIFF
--- a/test/src/cdp/TargetManager.spec.ts
+++ b/test/src/cdp/TargetManager.spec.ts
@@ -27,18 +27,25 @@ describe('TargetManager', () => {
     const {server, context, browser} = state;
 
     const targetManager = (browser as CdpBrowser)._targetManager();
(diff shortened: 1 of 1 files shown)


test/src/cdp/TargetManager.spec.ts
```diff
@@ -27,18 +27,25 @@ describe('TargetManager', () => {
     const {server, context, browser} = state;
 
     const targetManager = (browser as CdpBrowser)._targetManager();
-    expect(targetManager.getAvailableTargets().size).toBe(3);
+
+    const initialTargetCount = targetManager.getAvailableTargets().size;
+    // There could be an conditional extra prerender targets.
```
OrKoN:
```suggestion
    // There could be an conditional extra prerender target.
```

Lightning00Blade (approved):

PR #14388, opened 2025-10-30
chore: make CDP message ID generator configurable

szuend:
feature

no

no

**Summary**
This PR makes the ID generator for CDP messages configurable. This is useful if the same transport is shared across multiple connections and is a pre-requisite that puppeteer can proxy a Chrome DevTools connection and Chrome DevTools proxy a puppeter connection.

Changes constructor ordering of CDP connection, this may or may not break users.

**Other information**

BEGIN_COMMIT_OVERRIDE
refactor: make protocol (cdp/bidi) command ID generator configurable
END_COMMIT_OVERRIDE

### DIFF
--- a/packages/puppeteer-core/src/common/ConnectOptions.ts
+++ b/packages/puppeteer-core/src/common/ConnectOptions.ts
@@ -113,10 +113,13 @@ export interface ConnectOptions {
   browserURL?: string;
   transport?: ConnectionTransport;
   /**
+   * @internal
+   *
    * Custom ID generator for CDP / BiDi messages. Useful if the same transport
    * is shared for multiple connections.
    */
   idGenerator?: () => number;
+
   /**
    * Headers to use for the web socket connection.
    * @remarks
--- a/packages/puppeteer-core/src/common/ConnectOptions.ts
+++ b/packages/puppeteer-core/src/common/ConnectOptions.ts
@@ -113,6 +113,11 @@ export interface ConnectOptions {
   browserURL?: string;
   transport?: ConnectionTransport;
   /**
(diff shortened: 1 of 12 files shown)


packages/puppeteer-core/src/node/BrowserLauncher.ts
Lightning00Blade, line 78: Let's add the `idGenerator` to the `LaunchOptions` interface and we can mark only the prop behind `@internal` . Example of this `interface FrameEvents`

Lightning00Blade: didn't see the extend...

OrKoN:
let's mark it as `chore:` since there is no user-facing feature here. Also, CPD is probably CDP?

packages/puppeteer-core/src/common/ConnectOptions.ts
```diff
@@ -112,6 +112,11 @@ export interface ConnectOptions {
   browserWSEndpoint?: string;
   browserURL?: string;
   transport?: ConnectionTransport;
+  /**
+   * Custom ID generator for CDP / BiDi messages. Useful if the same transport
+   * is shared for multiple connections.
+   */
+  idGenerator?: () => number;
```
OrKoN, line 121: let's add `@internal` here

OrKoN:
Also, looks like some type tests fail

packages/puppeteer-core/src/common/ConnectOptions.ts
szuend: Done

OrKoN (approved):

PR #14390, opened 2025-10-31
chore: experimental includeAll flag for browser.pages()

OrKoN:
This would allow easier get targets like webviews that are available in Electron. We still do not support Electron officially but this is a feature that might be useful for other use cases.

### DIFF
--- a/packages/puppeteer-core/src/api/Browser.ts
+++ b/packages/puppeteer-core/src/api/Browser.ts
@@ -392,13 +392,15 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
    * returns all {@link Page | pages} in all
    * {@link BrowserContext | browser contexts}.
    *
+   * @param includeAll - experimental, setting to true includes all kinds of pages.
+   *
    * @remarks Non-visible {@link Page | pages}, such as `"background_page"`,
    * will not be listed here. You can find them using {@link Target.page}.
    */
-  async pages(): Promise<Page[]> {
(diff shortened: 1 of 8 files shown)


packages/puppeteer-core/src/api/BrowserContext.ts
```diff
@@ -188,8 +188,11 @@ export abstract class BrowserContext extends EventEmitter<BrowserContextEvents>
    *
    * @remarks Non-visible {@link Page | pages}, such as `"background_page"`,
    * will not be listed here. You can find them using {@link Target.page}.
+   *
+   * The `includeAll` option is experimental and setting it to true changes the
```
Lightning00Blade: I think this should me something like `@param includeAll` and that will get picked up better with the api-extractor.

packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -394,11 +394,14 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
    *
    * @remarks Non-visible {@link Page | pages}, such as `"background_page"`,
    * will not be listed here. You can find them using {@link Target.page}.
+   *
+   * The `includeAll` option is experimental and setting it to true changes the
```
Lightning00Blade: Same here.

Lightning00Blade (approved):
LGTM + 1 comment

Lightning00Blade (approved):

PR #14391, opened 2025-10-31
fix: fix setScreenOrientationOverride for firefox 140 esr

TerrorJack:
Fixes a regression for firefox 140 esr introduced in #14043.

### DIFF
--- a/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
+++ b/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
@@ -6,7 +6,9 @@
 
 import type * as Bidi from 'webdriver-bidi-protocol';
 
+import {ProtocolError} from '../../common/Errors.js';
 import {EventEmitter} from '../../common/EventEmitter.js';
+import {debugError} from '../../common/util.js';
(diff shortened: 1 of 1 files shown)


packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
```diff
@@ -608,10 +610,22 @@ export class BrowsingContext extends EventEmitter<{
...
-      screenOrientation,
-      contexts: [this.id],
-    });
+    try {
+      await this.#session.send('emulation.setScreenOrientationOverride', {
+        screenOrientation,
+        contexts: [this.id],
+      });
+    } catch (err) {
+      // Tolerate ProtocolError for
```
sadym-chromium, line 619: Please add "TODO: remove on ..." with a specific date when FF 144 is going stable.

sadym-chromium (approved):

OrKoN:
@TerrorJack puppeteer is not tied to the ESR versions of Firefox. We support the latest stable Firefox version only. Puppeteer v24.27.0 pins down Firefox 144.0.2 which is the version supported by Puppeteer.

PR #14395, opened 2025-11-04
fix: do not wait for all targets when connecting

OrKoN:
Fixes the issue of Puppeteer not being able to connect to some targets. Usually, it happens if the page or tab target is excluded by a filter but the browser instance has a child iframe target for excluded targets. In this case, the child target will never be auto-attached.

This PR fixes this by only expecting tab targets to have any child target attaching. This might mean that iframe subtargets are not fully initialized on connect but this should generally be fine.

### DIFF
--- a/packages/puppeteer-core/src/cdp/TargetManager.ts
+++ b/packages/puppeteer-core/src/cdp/TargetManager.ts
@@ -15,7 +15,8 @@ import {Deferred} from '../util/Deferred.js';
 
 import {CdpCDPSession} from './CdpSession.js';
 import type {Connection} from './Connection.js';
-import {CdpTarget, InitializationStatus} from './Target.js';
+import type {CdpTarget} from './Target.js';
+import {InitializationStatus} from './Target.js';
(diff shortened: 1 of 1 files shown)


packages/puppeteer-core/src/cdp/TargetManager.ts
```diff
@@ -370,13 +348,28 @@ export class TargetManager
           parentSession instanceof CdpCDPSession ? parentSession : undefined,
         );
 
+    const parentTarget =
+      parentSession instanceof CDPSession
```
Lightning00Blade: Nit: Just assert instanceof CdpCDPSession like we do above.

OrKoN: I do not think assertions would work here

Lightning00Blade (approved):

packages/puppeteer-core/src/cdp/TargetManager.ts
OrKoN: ah now I got what you mean

PR #14396, opened 2025-11-04
feat: add page.openDevTools() method

### LINKED ISSUE #14368: [Feature]: Support page.openDevTools()
### Feature description

CDP allows now opening a DevTools window for manual debugging https://chromedevtools.github.io/devtools-protocol/tot/Target/#method-openDevTools

In Puppeteer, we only have the DevTools launch argument that would make DevTools open for each target. We should support a dedicated page method to open DevTools: `page.openDevTools(): Promise<Page>` to allow debugging targets manually if needed.

jackfranklin:
Adds `page.openDevTools()` to the CDP implementation which will call
`Target.openDevTools` to open DevTools on the target page and then
return a `Page` that points to the DevTools instance.

Fixes #14368

### DIFF
--- a/packages/puppeteer-core/src/api/BrowserContext.ts
+++ b/packages/puppeteer-core/src/api/BrowserContext.ts
@@ -240,11 +240,6 @@ export abstract class BrowserContext extends EventEmitter<BrowserContextEvents> 
   abstract newPage(options?: CreatePageOptions): Promise<Page>;
 
   /**
-   * Opens DevTools on the provided Target.
-   */
-  abstract openDevTools(targetId: string): Promise<Page>;
-
-  /**
    * Gets the {@link Browser | browser} associated with this
    * {@link BrowserContext | browser context}.
    */
--- a/packages/puppeteer-core/src/api/BrowserContext.ts
+++ b/packages/puppeteer-core/src/api/BrowserContext.ts
@@ -240,6 +240,11 @@ export abstract class BrowserContext extends EventEmitter<BrowserContextEvents> 
   abstract newPage(options?: CreatePageOptions): Promise<Page>;
 
   /**
+   * Opens DevTools on the provided Target.
+   */
+  abstract openDevTools(targetId: string): Promise<Page>;
+
+  /**
    * Gets the {@link Browser | browser} associated with this
    * {@link BrowserContext | browser context}.
    */
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -3170,7 +3170,8 @@ export abstract class Page extends EventEmitter<PageEvents> {
   }
 
   /**
-   * Opens DevTools for the current Page and returns the DevTools Page.
+   * Opens DevTools for the current Page and returns the DevTools Page. This
+   * method is only available in Chrome.
    */
   abstract openDevTools(): Promise<Page>;
 }
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -3168,6 +3168,11 @@ export abstract class Page extends EventEmitter<PageEvents> {
   [asyncDisposeSymbol](): Promise<void> {
     return this.close();
   }
+
+  /**
+   * Opens DevTools for the current Page and returns the DevTools Page.
+   */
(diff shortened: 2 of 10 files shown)


packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -417,6 +418,37 @@ export class CdpPage extends Page {
...
+    assert(connection, 'No connection when opening DevTools from page');
+    const createSession = (isAutoAttachEmulated: boolean) => {
+      return connection._createSession(
+        targetInfoResponse.targetInfo,
+        isAutoAttachEmulated,
+      );
+    };
+    const devtoolsTarget = new DevToolsTarget(
+      targetInfoResponse.targetInfo,
+      undefined,
```
jackfranklin: @OrKoN this is supposed to be a `CdpCDPSession`, but I couldn't find the right way to pass this through.

I tried `this.#primaryTabTarget` but that just hangs (I don't think it's the right one anyway but thought I would try it). Open to help on what to do here - or confirmation that `undefined` is OK.

jackfranklin:
PTAL. I will run the doc generation stuff etc once we are happy with the code.

packages/puppeteer-core/src/cdp/Page.ts
jackfranklin: I also thought `this.target()._session` might work, but it also hangs.

OrKoN: I think we should do the following:

- move the logic to open DevTools into cdp/Browser.ts next to _createPageInContext (but it should accept a target ID not context). The Page.ts should just pass a target.
- re-use the waiting mechanism in _createPageInContext to wait for the Page object.
- do not create CDP sessions for this implementaiton.

jackfranklin: Thanks for the feedback, I have updated the PR. PTAL :)

packages/puppeteer-core/src/api/BrowserContext.ts
```diff
@@ -239,6 +239,11 @@ export abstract class BrowserContext extends EventEmitter<BrowserContextEvents>
    */
   abstract newPage(options?: CreatePageOptions): Promise<Page>;
 
+  /**
+   * Opens DevTools on the provided Target.
+   */
+  abstract openDevTools(targetId: string): Promise<Page>;
```
jackfranklin: I don't love passing targetId here - I would rather pass a `Page` but then we would need to expose `targetId` publicly so it can be read. wdyt?

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -417,6 +417,12 @@ export class CdpPage extends Page {
     return this.#emulationManager.javascriptEnabled;
   }
 
+  override async openDevTools(): Promise<Page> {
+    const pageTargetId = this.target()._targetId;
+    const devtoolsPage = await this.browserContext().openDevTools(pageTargetId);
```
OrKoN: Can we use `this.browser()` and skip browser context here? Let's also downcast to a CDP Browser and only implement  openDevTools on the CDP Browser.

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -3168,6 +3168,11 @@ export abstract class Page extends EventEmitter<PageEvents> {
   [asyncDisposeSymbol](): Promise<void> {
     return this.close();
   }
+
+  /**
+   * Opens DevTools for the current Page and returns the DevTools Page.
```
OrKoN: Let's document that it is only available in Chrome with CDP.

OrKoN (approved):

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -286,6 +286,10 @@ export class BidiPage extends Page {
     throw new Error('Method not implemented for WebDriver BiDi yet.');
   }
 
+  override openDevTools(): Promise<Page> {
+    throw new Error('Method not implemented for WebDriver BiDi yet.');
```
Lightning00Blade, line 290: Note this should use UnsupportedOperation Error.

Lightning00Blade: I will send a PR

PR #14399, opened 2025-11-05
fix: ignore wasn't found errors in NetworkManager

omjadas:

bugfix

no

n/a

**Summary**

if you send `Fetch.enable` for a worker, you receive the following error `ProtocolError: Protocol error (Fetch.enable): 'Fetch.enable' wasn't found`

no

**Other information**

this issue was introduced here https://github.com/puppeteer/puppeteer/issues/13752

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -661,7 +661,7 @@
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["firefox"],
     "expectations": ["SKIP"],
-    "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
+    "comment": "Workers do not work in Firefox yet"
   },
   {
     "testIdPattern": "[screencast.spec] Screencasts Page.screencast should validate options",
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -657,6 +657,13 @@
     "comment": "https://github.com/w3c/webdriver-bidi/issues/508"
   },
   {
+    "testIdPattern": "[requestinterception.spec] request interception Page.setRequestInterception should work with worker",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox"],
+    "expectations": ["SKIP"],
+    "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
+  },
+  {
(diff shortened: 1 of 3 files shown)


OrKoN (changes requested):
Thanks for the PR! Could you please add a test demonstrating that the issue is fixed? Also, I think a better solution would be to add the error message to `#canIgnoreError` instead of keeping track of each target's type support for each domain.

OrKoN:
This is actually already handled in packages/puppeteer-core/src/cdp/WebWorker.ts so the error is caught correctly.

omjadas:
I don't see where this could be handled in `packages/puppeteer-core/src/cdp/WebWorker.ts`. This error is encountered when `setRequestInterception` is called on the page, which in turn results in `applyProtocolRequestInterception` being called for each of the registered clients in the `NetworkManager`.

I would be happy to update `#canIgnoreError` instead of specifically checking if the target is a worker. Would you prefer for me to check the entirety of the error message, or perhaps just `wasn't found`? I'd also be happy to look into adding a test for this case.

OrKoN:
I meant this line that https://github.com/puppeteer/puppeteer/blob/05f61366ee61f480c7be6f237ec890c907fea67f/packages/puppeteer-core/src/cdp/WebWorker.ts#L84 that catches the error of adding a new client when creating the server worker.

OrKoN:
Indeed, the scenario of turning on interception after the page with workers was loaded looks like it might fail.

OrKoN:
I think checking for `wasn't found` would be good.

test/TestExpectations.json
```diff
@@ -656,6 +656,13 @@
     "expectations": ["FAIL"],
     "comment": "https://github.com/w3c/webdriver-bidi/issues/508"
   },
+  {
+    "testIdPattern": "[requestinterception.spec] request interception Page.setRequestInterception should work with worker",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox"],
+    "expectations": ["SKIP"],
+    "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
```
```suggestion
    "comment": "Workers do not work in Firefox yet"
```

OrKoN (approved):
thanks!

PR #14410, opened 2025-11-10
fix(cdp): update request with ExtraInfo if available

### LINKED ISSUE #4165: Modified request headers don't appear in response.request().headers()
I'm doing something relatively simple in the request interception:

```javascript
const headers = request.headers();
const modifiedHeaders = { ...headers, "x-new-header": "true" };

request.continue({ headers: modifiedHeaders });
```

However, later, when I receive the callback for the response, this header is missing from the associated request:

```javascript
response.request().headers()["x-new-header"] !== "true";
```
I am trying to mark the request as modified so that later in the response I know this was one of my intercepted requests.

Lightning00Blade:
CDP has an event call `Network.requestWillBeSentExtraInfo` it is send separately and may not even be sent at all.
Given this we can only safely test it once the request was fully completed.

Closes https://github.com/puppeteer/puppeteer/issues/4165

### DIFF
--- a/test/src/requestinterception.spec.ts
+++ b/test/src/requestinterception.spec.ts
@@ -133,7 +133,7 @@ describe('request interception', function () {
       expect(requests[1]!.url()).toContain('/one-style.css');
       expect(requests[1]!.headers()['referer']).toContain('/one-style.html');
     });
-    it.only('should not allow mutating request headers', async () => {
+    it('should not allow mutating request headers', async () => {
       const {page, server} = await getTestState();
 
       await page.setRequestInterception(true);
--- a/test/src/requestinterception.spec.ts
+++ b/test/src/requestinterception.spec.ts
@@ -133,7 +133,7 @@ describe('request interception', function () {
       expect(requests[1]!.url()).toContain('/one-style.css');
       expect(requests[1]!.headers()['referer']).toContain('/one-style.html');
     });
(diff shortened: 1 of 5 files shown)


test/src/requestinterception.spec.ts
```diff
@@ -144,7 +144,7 @@ describe('request interception', function () {
         }
         const headers = request.headers();
         headers['test'] = 'test';
-        void request.continue({headers});
+        void request.continue({headers: request.headers()});
```
OrKoN, line 147: I think this is using `headers` on purpose.

OrKoN (commented):
Could you please add a test for the new behavior?

test/src/requestinterception.spec.ts
OrKoN: was this test affected by the changes here?

Lightning00Blade: Yes the change affects this test, as the test is sending new Headers to be send along side the request. 
I think the test intention was to make sure the object is not writable.

OrKoN: 👍 I think in that case this PR fixes this https://github.com/puppeteer/puppeteer/issues/4165 as well?

Lightning00Blade: I think it should.

OrKoN (approved):

PR #14422, opened 2025-11-14
feat(webdriver): use `network.setExtraHeaders` instead of network interception

sadym-chromium:
WebDriver BiDi spec: https://www.w3.org/TR/webdriver-bidi/#command-network-setExtraHeaders

### ~~Blocked till [December 9, 2025](https://whattrainisitnow.com/release/?version=146)~~
* ~~[Firefox 146](https://whattrainisitnow.com/release/?version=146) with [implementation of `network.setExtraHeaders`](https://bugzilla.mozilla.org/show_bug.cgi?id=1979731) reaches stable.~~

### DIFF
--- a/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
+++ b/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
@@ -797,7 +797,7 @@ export class BrowsingContext extends EventEmitter<{
         );
 
         return {
-          name: key,
+          name: key.toLowerCase(),
           value: {type: 'string', value: value},
         };
       }),
--- a/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
+++ b/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
@@ -21,6 +21,8 @@ import {WindowRealm} from './Realm.js';
 import {Request} from './Request.js';
 import type {UserContext} from './UserContext.js';
 import {UserPrompt} from './UserPrompt.js';
+import {assert} from '../../util/assert.js';
+import {isString} from '../../common/util.js';
 
 /**
  * @internal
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
```diff
@@ -746,4 +748,20 @@ export class BrowsingContext extends EventEmitter<{
       contexts: [this.id],
     });
   }
+
+  async setExtraHTTPHeaders(headers: Record<string, string>): Promise<void> {
+    await this.#session.send('network.setExtraHeaders', {
+      headers: Object.entries(headers).map(([key, value])=>{
+        assert(
+          isString(value),
+          `Expected value of header "${key}" to be String, but "${typeof value}" is found.`,
```
Lightning00Blade, line 796: Why do we need this? The type from TypeScript should be good enough.

Lightning00Blade (commented):
LGTM + 1 comment

packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
sadym-chromium: Unless typescript type check is explicitly disabled: https://github.com/puppeteer/puppeteer/blob/e48ae6bd8ca39fc09c79f468b48b1666d5668726/test/src/network.spec.ts#L781-L782

Lightning00Blade: I see we do that for CDP as well. Do we need to set the headers to lower case as well?

sadym-chromium: This validation was moved from [`Page.setExtraHTTPHeaders`](https://github.com/puppeteer/puppeteer/pull/14422/files#diff-5f7d96e4eb1cf9ee2e6e4d6ada97ad12d8348535d33475af67126b281561e988L751).

sadym-chromium: > I see we do that for CDP as well. Do we need to set the headers to lower case as well?

Good point. Added.

Lightning00Blade (approved):

PR #14430, opened 2025-11-18
perf: dispose sub-classes correctly

Lightning00Blade:
We need to ensure that sub-resource are handled correctly.

### DIFF
--- a/packages/puppeteer-core/src/api/Browser.ts
+++ b/packages/puppeteer-core/src/api/Browser.ts
@@ -692,18 +692,17 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
 
   /** @internal */
   override [disposeSymbol](): void {
-    if (this.process()) {
-      return void this.close().catch(debugError);
-    }
-    return void this.disconnect().catch(debugError);
+    return void this[asyncDisposeSymbol]().catch(debugError);
   }
 
   /** @internal */
-  [asyncDisposeSymbol](): Promise<void> {
+  override async [asyncDisposeSymbol](): Promise<void> {
     if (this.process()) {
-      return this.close();
+      await this.close();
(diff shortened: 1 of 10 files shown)


test/src/evaluation.spec.ts
```diff
@@ -178,7 +178,7 @@ describe('Evaluation specs', function () {
       });
       expect(result).toBe(56);
     });
-    it('should work right after framenavigated', async () => {
+    it.only('should work right after framenavigated', async () => {
```
OrKoN:
```suggestion
    it('should work right after framenavigated', async () => {
```

OrKoN (approved):

packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -692,18 +692,17 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
...
-  [asyncDisposeSymbol](): Promise<void> {
+  override async [asyncDisposeSymbol](): Promise<void> {
     if (this.process()) {
-      return this.close();
+      await this.close();
+    } else {
+      await this.disconnect();
     }
-    return this.disconnect();
+    await super[asyncDisposeSymbol]();
```
OrKoN, line 706: should we call super[asyncDisposeSymbol] in a `finally` block? or what is the expected behavior for dispose handlers?

OrKoN (approved):

packages/puppeteer-core/src/api/Browser.ts
Lightning00Blade: Good question, I think in case of error it's expected that things may not be disposed correctly. And the logs should show that.

OrKoN (approved):

PR #14432, opened 2025-11-18
chore: extract `DeviceRequestPrompt` interface

sadym-chromium:
Required for handling device requests via WebDriver BiDi.


packages/puppeteer-core/src/cdp/DeviceRequestPrompt.ts
```diff
@@ -7,59 +7,17 @@
...
- * await devicePrompt.select(
- *   await devicePrompt.waitForDevice(({name}) => name.includes('My Device')),
- * );
- * ```
- *
- * @public
+ * @internal
  */
-export class DeviceRequestPrompt {
+export class DeviceRequestPromptImpl implements DeviceRequestPrompt {
```
Lightning00Blade: Let's call it `CdpDeviceRequestPrompt` the `Impl` suffix is the same as as adding Interface at the end which is discouraged practice - https://google.github.io/styleguide/tsguide.html#naming-style
We can also change the Interface to something like `DeviceRequestPromptManager`.

sadym-chromium: Good point. Renamed.

packages/puppeteer-core/src/api/DeviceRequestPrompt.ts
```diff
@@ -0,0 +1,79 @@
...
+ * );
+ * ```
+ *
+ * @public
+ */
+export abstract class DeviceRequestPrompt {
+  /**
+   * Current list of selectable devices.
+   */
+  abstract get devices(): DeviceRequestPromptDevice[];
```
Lightning00Blade, line 60: Note: This can also be a property and not a getter, then even a getter would work to satisfy it.

Lightning00Blade (approved):

packages/puppeteer-core/src/api/DeviceRequestPrompt.ts
sadym-chromium: But we don't want to expose setter. Do we?

sadym-chromium: I consider this is addressed. If not, can be changed in a follow-up

Lightning00Blade: True, I guess there we can add a readonly modifier, but non blocker comment.

PR #14441, opened 2025-11-20
ci: update permission for actions

Lightning00Blade:


.github/workflows/release-please.yml
```diff
@@ -1,14 +1,12 @@
+name: release-please
+
+# Declare default permissions as read only.
+permissions: read-all
 on:
   push:
     branches:
       - main
 
-permissions:
```
Lightning00Blade, line 6: We should not need this as we have a custom token for this.

OrKoN: Let's see. I recall they use the custom token only for some operations.

OrKoN (approved):

.github/workflows/release-please.yml
Lightning00Blade: Interesting, I will observe anyway, but we should set the permission on the job then on the run.

PR #14445, opened 2025-11-20
feat: add browser.screens, .addScreen and .removeScreen methods

kvitekp:
This update introduces new `Browser` methods for headless screen management:
* `screens`: Returns an array of `ScreenInfo` objects, each containing properties of an existing screen.
* `addScreen`: Adds a new screen with specified parameters and returns a `ScreenInfo` object, including the new screen's ID.
* `removeScreen`: Removes the screen associated with the given screen ID, unless it is the primary screen.

These methods are not supported by BiDi yet.


packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -509,6 +549,40 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
...
+  abstract screens(): Promise<ScreenInfo[]>;
+
+  /**
+   * Adds a new screen, returns the added {@link ScreenInfo | screen information object}.
+   *
+   * @remarks
+   *
+   * Only supported in headless mode.
+   */
+  abstract addScreen(
```
OrKoN: let's have the input as an object, otherwise it is easy to confuse positional arguments of the same type. You can probably use ScreenInfo but omit id, like `Omit<ScreenInfo, 'id'>` as a type

packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -289,6 +291,29 @@ export class BidiBrowser extends Browser {
     await this.#browserCore.uninstallExtension(id);
   }
 
+  override screens(): Promise<ScreenInfo[]> {
+    throw 'Not implemented';
```
OrKoN: let's `throw UnsupportedOperation()` and in other throw statements as well

package-lock.json
OrKoN, line 1: please revert the package-lock.json change. You can use `npm ci` to install deps in puppeteer without modifying the lock file.

packages/puppeteer-core/src/bidi/Browser.ts
kvitekp: Done.

package-lock.json
kvitekp: Oops, done!

packages/puppeteer-core/src/api/Browser.ts
kvitekp: Done, added `AddScreenParams`.

OrKoN (approved):

OrKoN:
Adjusted expectations for WebDriver BiDi, it should auto-land after CI is done.

test/src/browser.spec.ts
```diff
@@ -109,4 +109,76 @@ describe('Browser specs', function () {
       expect(newBrowser.isConnected()).toBe(false);
     });
   });
+
+  describe.only('Browser.screens', function () {
```
OrKoN:
```suggestion
  describe('Browser.screens', function () {
```

test/src/browser.spec.ts
```diff
@@ -109,4 +109,76 @@ describe('Browser specs', function () {
...
+          left: 0,
+          orientation: {angle: 0, type: 'landscapePrimary'},
+          top: 0,
+          width: 800,
+        },
+      ]);
+    });
+  });
+
+  describe.only('Browser.add|removeScreen', function () {
```
OrKoN:
```suggestion
  describe('Browser.add|removeScreen', function () {
```

OrKoN:
@kvitekp looks like some of the new features fail on Windows and Mac. Please see https://github.com/puppeteer/puppeteer/actions/runs/19640953389/job/56244079947?pr=14454 Should we mark the failures as expected?

PR #14446, opened 2025-11-20
feat: support new page window positioning

kvitekp:
This update adds `WindowBounds` property to the `CreatePageOptions` parameter so that `Browser.newPage` method caller can specify the new window position and state.

Window bounds are specified in screen coordinates and are defined by the following properties:
* `left`: Window's left edge position.
* `top`: Window's top edge position.
* `width`: Window's width.
* `height`: Window's height.
* `windowState`: can be 'normal', 'minimized', 'maximized', or 'fullscreen'.


OrKoN (approved):
LGTM! Probably needs `npm run docs` to regenerate API docs before the CI is happy

PR #14450, opened 2025-11-21
fix: HTTPResponse.text() method expected to throw when content is malformed

### LINKED ISSUE #14449: [Bug]: HTTPResponse.text() method is returning arbitrary text for non utf-8 strings

```TypeScript
const puppeteer = require('puppeteer-core');

async function test() {
  const browser = await puppeteer.launch({channel : "chrome"});

  const binaryUrl =
      'https://www.google.com/images/branding/googlelogo/1x/googlelogo_color_84x28dp.png';

  page.on('response', async response => {
    if (response.url() === binaryUrl) {
      try {
        console.log(`URL: ${response.url()}`);
        console.log(`Content-Type: ${response.headers()['content-type']}`);

        const text = await response.text();
        console.log('response.text() output:', text);

      } catch (error) {
        console.error('Error fetching response content:', error);
      }
    }
  });

  await page.goto(binaryUrl, {waitUntil : 'networkidle0'});
  await browser.close();
}

test();
```

### Background

_No response_

### Expectation

Fails if the content is valid utf-8 string 

### Reality

Arbitrary text is returned instead.
The output from the example looks like this (first n bytes):
�PNG

IHDR�`���IDATx��Zqle��u�@D�����0������چ (J�5D�P��V��6�ŀD1�BTE�uC0�

```TypeScript

```

### Puppeteer version

24.31.0

### Node version

v22.19.0

### Package manager

npm

### Package manager version

10.9.3

### Operating system

Linux

nroscino:
This patch fixes HTTPResponse.text() method to throw when the content parsed is binary and not an utf-8 string.

Now the text() method will throw instead of returning arbitrary text data when the content is not an utf-8 string but instead a binary data.

bugfix

yes
yes
**Summary**

Reason for the change:
https://github.com/puppeteer/puppeteer/issues/14449

no

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/api/HTTPResponse.test.ts
+++ b/packages/puppeteer-core/src/api/HTTPResponse.test.ts
@@ -1,0 +1,37 @@
+/**
+ * @license
+ * Copyright 2025 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import expect from 'expect';
+import {describe, it} from 'node:test';
+
+import {HTTPResponse} from './HTTPResponse.js';
+
+//@ts-ignore
+class TestResponse extends HTTPResponse {
+  constructor(private readonly contentToReturn: Uint8Array) { super(); }
+
+  override content(): Promise<Uint8Array> {
+    return Promise.resolve(this.contentToReturn);
+  }
+}
+
+describe('HTTPResponse', () => {
+  describe('text', () => {
+    it('should return the content as text', async () => {
+      const content = new TextEncoder().encode('hello');
+      const testResponse = new TestResponse(content);
+
+      const text = await testResponse.text();
+      expect(text).toBe('hello');
+    });
+
+    it('should throw if content is not valid UTF-8', async () => {
+      const content = new Uint8Array([ 0xff ]);
+      const testResponse = new TestResponse(content);
+
+      await expect(testResponse.text()).rejects.toThrow();
+    });
+  });
+});
--- a/packages/puppeteer-core/src/api/HTTPResponse.test.ts
+++ b/packages/puppeteer-core/src/api/HTTPResponse.test.ts
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2025 Google Inc.
+ * Copyright 2026 Google Inc.
  * SPDX-License-Identifier: Apache-2.0
  */
 import {describe, it} from 'node:test';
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/api/HTTPResponse.test.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @license
+ * Copyright 2025 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import expect from 'expect';
+import {describe, it} from 'node:test';
+
+import {HTTPResponse} from './HTTPResponse.js';
+
+//@ts-ignore
```
Lightning00Blade: Why is this needed?

packages/puppeteer-core/src/api/HTTPResponse.test.ts
```diff
@@ -0,0 +1,37 @@
...
+class TestResponse extends HTTPResponse {
+  constructor(private readonly contentToReturn: Uint8Array) { super(); }
```
Lightning00Blade: Let's write it out fully, without constructor parms.

OrKoN:
let's not mark it as a breaking change because we are not ready for a breaking change release yet.

Lightning00Blade:
Note: The rationale behind marking this as non breaking change is the the current approach does not yield data that is useful to the end user, also if this were done in a EventListener all errors there are swallowed.

packages/puppeteer-core/src/api/HTTPResponse.test.ts
nroscino: It was asking to implement all the other method of the HTTPResponse but for this test only the content() had to be mocked. I tried to create a mock with sinon first but I think I had to do some tricks anyway cause the abstract method couldn't be easily mocked.

Lightning00Blade: In that case use a `@ts-expect-error` with a description of why we are doing this.

packages/puppeteer-core/src/api/HTTPResponse.test.ts
nroscino: I wrote it as a setter! let me know if that's ok for you, so I can resolve the conversation!

nroscino:
@Lightning00Blade should have updated everything. Should I also rewrite the commit referring to a breaking change or it will be squashed so it's not a problem?

Lightning00Blade (approved):

Lightning00Blade:
Not needed, we squash the commits and use the title for the message.

OrKoN (approved):

PR #14455, opened 2025-11-24
feat(webdriver): bluetooth emulation via WebDriver BiDi

sadym-chromium:
* Implement Bluetooth Emulation and Device Prompts via WebDriver BiDi.
* Changed `DeviceRequestPromptDevice` from class to interface.

### DIFF
--- a/packages/puppeteer-core/src/bidi/DeviceRequestPrompt.ts
+++ b/packages/puppeteer-core/src/bidi/DeviceRequestPrompt.ts
@@ -1,0 +1,134 @@
+/**
+ * @license
+ * Copyright 2025 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type * as Bidi from 'webdriver-bidi-protocol';
+
+import {
+  DeviceRequestPrompt,
+  type DeviceRequestPromptDevice,
+} from '../api/DeviceRequestPrompt.js';
+import {Deferred} from '../util/Deferred.js';
+
+import type {Session} from './core/Session.js';
+
+/**
+ * @internal
+ */
+export class BidiDeviceRequestPromptManager {
+  readonly #session: Session;
+  readonly #contextId: string;
+  #enabled = false;
+
+  constructor(contextId: string, session: Session) {
+    this.#session = session;
+    this.#contextId = contextId;
+  }
+
+  async #enableIfNeeded(): Promise<void> {
+    if (!this.#enabled) {
+      this.#enabled = true;
+      await this.#session.subscribe(
+        ['bluetooth.requestDevicePromptUpdated'],
+        [this.#contextId],
+      );
+    }
+  }
+
+  async waitForDevicePrompt(
+    timeout: number,
+    signal: AbortSignal | undefined,
+  ): Promise<DeviceRequestPrompt> {
+    const deferred = Deferred.create<DeviceRequestPrompt>({
+      message: `Waiting for \`DeviceRequestPrompt\` failed: ${timeout}ms exceeded`,
+      timeout,
+    });
+
+    const onRequestDevicePromptUpdated = (
+      params: Bidi.Bluetooth.RequestDevicePromptUpdatedParameters,
+    ) => {
+      if (params.context === this.#contextId) {
+        deferred.resolve(
+          new BidiDeviceRequestPrompt(
+            this.#contextId,
+            this.#session,
+            params.prompt,
+            params.devices,
+          ),
+        );
+        this.#session.off(
+          'bluetooth.requestDevicePromptUpdated',
+          onRequestDevicePromptUpdated,
+        );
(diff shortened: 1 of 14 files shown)


Lightning00Blade:
Can we split out `Aligned platform object serialization with CDP.` as that is a user facing change and I would prefer for it to be easily seen in the Changelog

sadym-chromium:
> Can we split out `Aligned platform object serialization with CDP.` as that is a user facing change and I would prefer for it to be easily seen in the Changelog

https://github.com/puppeteer/puppeteer/pull/14457

sadym-chromium:
> Can we split out `Aligned platform object serialization with CDP.` as that is a user facing change and I would prefer for it to be easily seen in the Changelog

Actually, I changed the test to return primitive instead of a platform object. This allows avoiding breaking unrelated changes.

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -474,8 +475,11 @@ export class BidiFrame extends Frame {
     );
   }
 
-  override waitForDevicePrompt(): never {
-    throw new UnsupportedOperation();
+  override waitForDevicePrompt(
+    options: WaitTimeoutOptions = {},
+  ): Promise<DeviceRequestPrompt> {
+    const {timeout = this.timeoutSettings.timeout(), signal} = options ?? {};
```
Lightning00Blade: Already set in the parameters.
```suggestion
    const {timeout = this.timeoutSettings.timeout(), signal} = options;
```

packages/puppeteer-core/src/cdp/DeviceRequestPrompt.test.ts
```diff
@@ -159,9 +179,14 @@ describe('DeviceRequestPrompt', function () {
...
         ],
       });
-      expect(prompt.devices).toHaveLength(2);
-      expect(prompt.devices[0]).toBeInstanceOf(DeviceRequestPromptDevice);
-      expect(prompt.devices[1]).toBeInstanceOf(DeviceRequestPromptDevice);
+      expect(prompt.devices).toEqual(
+        [
+          {id: '00000000', name: 'Device 0'},
+          {id: '11111111', name: 'Device 1'},
+        ].sort((a, b) => {
```
Lightning00Blade: Why do we need to sort it here, can we just provide the expected result in the correct order?

packages/puppeteer-core/src/bidi/DeviceRequestPrompt.ts
```diff
@@ -0,0 +1,137 @@
...
+      signal.addEventListener(
+        'abort',
+        () => {
+          deferred.reject(signal.reason);
+        },
+        {once: true},
+      );
+    }
+
+    void this.#enableIfNeeded();
```
Lightning00Blade: I think we can await this so any error get caught correctly.

packages/puppeteer-core/src/bidi/DeviceRequestPrompt.ts
```diff
@@ -0,0 +1,137 @@
...
+    await this.#session.send('bluetooth.handleRequestDevicePrompt', {
+      context: this.#contextId,
+      prompt: this.#promptId,
+      accept: true,
+      device: device.id,
+    });
+  }
+
+  waitForDevice(): never {
+    throw new Error('Not implemented');
```
Lightning00Blade: Use the correct error Unsupported operation

packages/puppeteer-core/src/cdp/DeviceRequestPrompt.test.ts
sadym-chromium: for test resilience. We can rely on the specific implementation and remove the sort.

Lightning00Blade: But you are not sorting the input but the expectation. That is controlled by us, if we want to make it resilient i would expect this to be on the `prompt.devices`.

packages/puppeteer-core/src/bidi/DeviceRequestPrompt.ts
sadym-chromium: agree. Done

packages/puppeteer-core/src/bidi/DeviceRequestPrompt.ts
sadym-chromium: done

packages/puppeteer-core/src/cdp/DeviceRequestPrompt.test.ts
sadym-chromium: removed

sadym-chromium:
@Lightning00Blade could you please take another look?

Lightning00Blade (approved):

PR #14457, opened 2025-11-24
fix(webdriver): align deserialization with CDP

sadym-chromium:
Serialize errors and platform objects as empty objects.

### DIFF
--- a/packages/puppeteer-core/src/bidi/Deserializer.ts
+++ b/packages/puppeteer-core/src/bidi/Deserializer.ts
@@ -31,13 +31,13 @@ export class BidiDeserializer {
           return acc.add(this.deserialize(value));
         }, new Set());
       case 'object':
-        if (!result?.value) {
+        if (!result?.value && !result.internalId) {
           // Heuristic detecting platform objects. WebDriver BiDi serializes platform
           // objects without value. Return an empty object, as there is no way to restore
           // the original value's constructor.
           return {};
         }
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/bidi/Deserializer.ts
```diff
@@ -19,6 +19,9 @@ export class BidiDeserializer {
     }
 
     switch (result.type) {
+      case 'error':
+        // Aligned with CDP implementation, which returns `{}` for errors.
```
Lightning00Blade, line 23: I don't think this comment is needed, the test are the source of what is expected.

Lightning00Blade (approved):

packages/puppeteer-core/src/bidi/Deserializer.ts
```diff
@@ -28,6 +31,12 @@ export class BidiDeserializer {
           return acc.add(this.deserialize(value));
         }, new Set());
       case 'object':
+        if (!result?.value && !result.internalId) {
+          // Heuristic detecting platform objects. WebDriver BiDi serializes platform
+          // objects without value. Return an empty object, as there is no way to restore
+          // the original value's constructor.
+          return {};
```
sadym-chromium, line 38: Actually I'm not sure if we really want to return an empty object for platform objects, as it can be really confusing and frustrating for the users.

sadym-chromium:
Closing, as I'm not sure it is the correct behavior to return empty objects instead of platforms is the correct behavior.

PR #14458, opened 2025-11-25
chore: Temporarily change Firefox expectations for navigation.spec test case

vinhill:
**Summary**

Canary expectation change for Firefox  due to a regression from [bug 543435](http://bugzilla.mozilla.org/543435) that will likely be fixed soon in [bug 2001934](http://bugzilla.mozilla.org/2001934).


whimboo:
@OrKoN if you are fine with leaving this test failing for one or two more days then we could ignore this particular PR. I would leave it up to you to decide what's best - and to save us from landing a reverting change soon.

OrKoN:
If the fix is landing in Nightly soon, we can skip landing this PR.

vinhill:
The fix landed.

PR #14463, opened 2025-11-26
feat: add resolveDefaultUserDataDir for Chrome

OrKoN:
This allows computing the expected default user data dir based on the env variables.

### DIFF
--- a/packages/browsers/src/browser-data/chrome.ts
+++ b/packages/browsers/src/browser-data/chrome.ts
@@ -5,6 +5,7 @@
  */
 
 import {execSync} from 'node:child_process';
+import os from 'node:os';
 import path from 'node:path';
 
 import semver from 'semver';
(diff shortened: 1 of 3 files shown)


packages/browsers/src/browser-data/chrome.ts
```diff
@@ -205,26 +229,12 @@ function getWslLocation(channel: ChromeReleaseChannel): [string, ...string[]] {
   const wslPrefixes = new Set<string>();
   for (const name of WINDOWS_ENV_PARAM_NAMES) {
     try {
```
Lightning00Blade: I think this try is not needed now.

packages/browsers/test/src/chrome/chrome-data.spec.ts
```diff
@@ -124,6 +126,213 @@ describe('Chrome', () => {
     );
   });
 
+  it('should resolve default user data dir', () => {
+    // Every platform has its own tests.
```
Lightning00Blade: Why do we want this to be in one test? It make is hard to detect 2 failed things as the first will exit the test.
Maybe a describe about the default user data dir + it for each platform?

Lightning00Blade (approved):
LGTM, with comments.

PR #14466, opened 2025-11-28
feat: support channel in puppeteer.connect

OrKoN:
This allows discovering active ports for the regular Chrome channels.

### DIFF
--- a/packages/puppeteer-core/rollup.config.mjs
+++ b/packages/puppeteer-core/rollup.config.mjs
@@ -27,6 +27,9 @@ export default {
     if (id.includes('node_modules')) {
       return true;
     }
+    if (id.includes('@puppeteer/browsers')) {
+      return true;
+    }
     if (id.startsWith(path.join(cwd, `lib`, `esm`, `puppeteer`, `bidi`))) {
       return true;
     }
--- a/packages/puppeteer-core/src/common/BrowserConnector.ts
+++ b/packages/puppeteer-core/src/common/BrowserConnector.ts
@@ -102,17 +102,20 @@ async function getConnectionTransport(
     );
     const portPath = join(userDataDir, 'DevToolsActivePort');
     try {
-      const port = parseInt(
-        await environment.value.fs.promises.readFile(portPath, 'ascii'),
-        10,
+      const portRawValue = await environment.value.fs.promises.readFile(
+        portPath,
+        'ascii',
       );
-      if (port <= 0 || port > 65535) {
-        throw new Error(`Invalid port ${port} found`);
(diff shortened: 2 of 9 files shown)


OrKoN:
@Lightning00Blade PTAL. I do not have a good testing idea yet.

packages/puppeteer-core/src/common/BrowserConnector.ts
```diff
@@ -84,6 +84,47 @@ async function getConnectionTransport(
...
+      platform,
+      convertPuppeteerChannelToBrowsersChannel(options.channel),
+    );
+    const portPath = join(userDataDir, 'DevToolsActivePort');
+    try {
+      const port = parseInt(
+        await environment.value.fs.promises.readFile(portPath, 'ascii'),
+        10,
+      );
+      if (port <= 0 || port > 65535) {
```
Lightning00Blade: We should check for NaN as well

packages/puppeteer-core/src/common/BrowserConnector.ts
```diff
@@ -84,6 +84,47 @@ async function getConnectionTransport(
...
+        throw new Error(`Invalid port ${port} found`);
```
Lightning00Blade: This should be the value read from the file not port.

packages/puppeteer-core/src/common/BrowserConnector.ts
```diff
@@ -84,6 +84,47 @@ async function getConnectionTransport(
...
+      }
+      const browserWSEndpoint = `ws://localhost:${port}`;
+      const WebSocketClass = await getWebSocketTransportClass();
+      const connectionTransport: ConnectionTransport =
```
Lightning00Blade:
```suggestion
      const connectionTransport =
```

packages/puppeteer-core/rollup.config.mjs
```diff
@@ -27,6 +27,9 @@ export default {
     if (id.includes('node_modules')) {
       return true;
     }
+    if (id.includes('@puppeteer/browsers')) {
```
Lightning00Blade, line 30: Why is this needed? It should be detected correctly, no?

OrKoN: No, it tried to bundle @puppeteer/browsers as it is dynamically imported outside of the node folder that previously had imports from browsers (see line 36 where node folder is excluded).

Lightning00Blade: I see. I was under the wrong impression that rollup would think it's external directly.

Lightning00Blade (approved):

PR #14475, opened 2025-12-02
fix: prevent accessibility nodes attributes with false from being ignored

abdelmalekkkkk:

This PR prevents attributes with a correct value of `false` from being ignored during serialization.

Yes.

N/A

**Summary**

When serializing accessibility nodes, boolean properties can be `false`, but the current serialization function ignores all falsey values, which is wrong. For example, when a `select` (a ComboBox) is not open (not expanded), it should have the property `expanded` set to `false`.

No.

**Other information**


OrKoN (approved):
LGTM, thanks!

PR #14494, opened 2025-12-09
feat: add Browser.get|setWindowBounds and Page.windowId methods

kvitekp:

This change introduces Browser.get|setWindowBounds and Page,windowId methods. The latter one returns a WindowID that is used by the former two  to identify a browser window.

Yes

Yes

**Summary**
These new browser methods allow users to manage browser window position and state. This allows windows to be placed on a particular screen in a required state.

No

**Other information**
None

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -265,6 +265,20 @@
     "comment": "Not testable in headful"
   },
   {
+    "testIdPattern": "[browser.spec] Browser specs Browser.get|setWindowBounds should get and set browser window bounds",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["webDriverBiDi"],
+    "expectations": ["FAIL"],
+    "comment": "Not supported by WebDriver BiDi"
+  },
+  {
+    "testIdPattern": "[browser.spec] Browser specs Browser.get|setWindowBounds should set and get browser window maximized state",
+    "platforms": ["darwin", "linux", "win32"],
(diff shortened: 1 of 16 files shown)


test/src/page.spec.ts
```diff
@@ -2753,4 +2753,25 @@ describe('Page', function () {
...
+      const page = await context.newPage();
+
+      // Default view port restricts window to 800x600, so remove it.
+      await page.setViewport(null);
+
+      const contentWidth = 500;
+      const contentHeight = 400;
+      await page.resize({contentWidth, contentHeight});
+
+      const innerSize = await page.evaluate(() => {
```
OrKoN, line 2776: should we wait for a resize event in DOM to eliminate flakiness?

test/TestExpectations.json
```diff
@@ -1496,6 +1517,13 @@
     "expectations": ["FAIL"],
     "comment": "Returns a different window size in Chrome Headless on Mac"
   },
+  {
+    "testIdPattern": "[page.spec] Page Page.resize should resize the browser window to fit page content",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["chrome", "firefox"],
```
OrKoN: that would mean `chrome AND firefox` which would not resolve to anything (I think).

test/src/page.spec.ts
OrKoN: published an update to test this.

OrKoN (approved):

PR #14495, opened 2025-12-10
feat: support ignore tab target

kongfu-cat:

Feature: Adds a global IGNORE_TAB_TARGET variable to support ignoring tab targets, ensuring compatibility with older Chrome versions (Chrome 114 and below).

Add chrome-headless-page test-suite

**Summary**

Older Chrome versions (114 and below) exhibit inconsistent tab target behavior that can cause Puppeteer to misidentify or process redundant tab targets, leading to unexpected errors in page initialization and target management workflows.

This PR introduces the global IGNORE_TAB_TARGET variable as a toggle to skip the handling of tab targets. When enabled, it suppresses the problematic target detection logic that conflicts with legacy Chrome versions, while maintaining full functionality for newer Chrome releases when the flag is disabled.

This change addresses the compatibility gap without altering the core target handling logic for modern browsers, ensuring a seamless experience for users working with older Chrome deployments.

No. This PR is fully backward-compatible:

- The IGNORE_TAB_TARGET flag defaults to false, preserving the existing behavior for all users who do not explicitly enable it.
- No modifications are made to existing APIs or target processing logic for modern Chrome versions.
- No migration steps are required for existing applications.

**Other information**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,6 +167,7 @@ jobs:
           - chrome-headless-shell
           - chrome-bidi
           - chrome-pipe
+          - chrome-headless-page
         os:
           - ubuntu-latest
           - windows-latest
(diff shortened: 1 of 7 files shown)


OrKoN:
Thanks for the PR but compatibility with the older versions is a non-goal for Puppeteer. See https://pptr.dev/supported-browsers

PR #14186, opened 2025-09-10
test: disable Firefox tests that block releasing

OrKoN:

PR #14496, opened 2025-12-10
test: try flakes

OrKoN:


OrKoN:
cc @whimboo

whimboo:
@OrKoN when did it start? It would be good to have a rough date range to allow me to check. Also it looks like even with your updates to the expectations the tests are still failing?

OrKoN:
@whimboo it started on the day when we disabled it in https://github.com/puppeteer/puppeteer/pull/14186

whimboo:
there is https://github.com/puppeteer/puppeteer/pull/14628 which will take care of those two tests.

PR #14501, opened 2025-12-11
feat: emulate focused page

OrKoN:
This allows emulating a focused page similar to the DevTools feature.

### DIFF
--- a/test/src/emulation.spec.ts
+++ b/test/src/emulation.spec.ts
@@ -614,4 +614,50 @@ describe('Emulation', () => {
       await page.emulateCPUThrottling(null);
     });
   });
+
+  describe('Page.emulateFocusedPage', function () {
+    it('should emulate focus', async () => {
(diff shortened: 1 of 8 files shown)


test/src/emulation.spec.ts
```diff
@@ -614,4 +614,48 @@ describe('Emulation', () => {
...
+      await page2.bringToFront();
+
+      expect(
+        await page.evaluate(() => {
+          return document.hasFocus();
+        }),
+      ).toBe(true);
+    });
+
+    it('should emulate lack of focus', async () => {
```
Lightning00Blade: This test looks wrong, does it mean we can disable this or if it's emulation lack of we don't need a second page.

OrKoN: I did not get what you meant but I updated the test.

Lightning00Blade (approved):

PR #14513, opened 2025-12-15
fix(webdriver): emit `response` event on `network.responseStarted`

sadym-chromium:

PR #14514, opened 2025-12-15
chore(webdriver): bump chromium-bidi to 12.0.1

sadym-chromium:


OrKoN:
Looks like "[network.spec] network Response.text should wait until response completes" consistently times out now.

OrKoN (approved):

sadym-chromium:
> Looks like "[network.spec] network Response.text should wait until response completes" consistently times out now.

Right, I believe it has the same root cause as https://github.com/puppeteer/puppeteer/pull/14513, so I propose to update expectations with timeout for now.

Issue #14515, opened 2025-12-16
[Bug]: request interception causes hanging fetch requests


mGolestan98:
### Minimal, reproducible example

```TypeScript
import puppeteer from "puppeteer";

const browser = await puppeteer.launch({
  headless: false
});

const page = await browser.newPage();
await page.setRequestInterception(true);

page.on("request", async (request) => {
  if (request.url().includes("ad.doubleclick.net")) {
    console.log("request event received for: ", request.url());
    console.log("================================================");
  }
  request.continue();
});
page.on("response", async (response) => {
  if (response.request().url().includes("ad.doubleclick.net")) {
    console.log("response event received for: ", response.request().url());
    console.log("================================================");
  }
});

// this is an example website
await page.goto("https://www.verizon.com/", {
  waitUntil: "networkidle2",
  timeout: 10000
});

await browser.close();
```

### Background

When enabling request interception in puppeteer and try loading websites that have trackers from `ad.doubleclick.net` , the `page.goto()` fails after the timeout threshold because no response is received for for the fetch requests to `ad.doubleclick.net`.

I could trace the issue into the `NetworkManager` module and found that when `Fetch.requestPaused` is received, puppeteer waits for the corresponding `Network.requestWillBeSent` to be received too (which is correct by design as i assume the order is not guaranteed), but in this particular case the `Network.requestWillBeSent` is never delivered for the redirect request.

### Expectation

response event should be emitted for ad.doubleclick.net network calls so that goto function does not fail

### Reality

response event won't be emitted for fetch requests to ad.doubleclick.net and goto fails

### Puppeteer version

24.33.0

### Node version

20.19.1

### Package manager

npm

### Package manager version

10.8.2

### Operating system

macOS

OrKoN:
I am not able to reproduce this. It might be related to a specific content that gets loaded for you https://www.verizon.com. Would you be able to pinpoint what sort request hangs and provide a test page? Indeed, the order between requestWillBeSent and requestPaused is not guaranteed so Puppeteer waits for both.

mGolestan98:
Thank you @OrKoN for the quick feedback.

I took a look and it seems this might be related to geo-specific behavior. The tracking script doesn’t appear to be initialized for EU IP addresses. Would you mind trying again with a US IP to see if it reproduces there?

OrKoN:
@mGolestan98 sorry, I have no access to US IP addresses.

OrKoN:
[...]

PR #14522, opened 2025-12-18
test: add a keep alive fetch interception test

OrKoN:
Adds a test for keep alive requests.

Refs: https://github.com/puppeteer/puppeteer/issues/14515

### DIFF
--- a/test/src/requestinterception.spec.ts
+++ b/test/src/requestinterception.spec.ts
@@ -80,6 +80,49 @@ describe('request interception', function () {
         page.waitForNavigation(),
       ]);
     });
+    it('should work with keep alive redirects', async () => {
+      const {page, server} = await getTestState();
+      server.setRoute('/rredirect', (_req, res) => {
+        res.writeHead(302, {location: '/target'});
+        res.end();
+      });
+      server.setRoute('/target', (_req, res) => {
+        res.end('Hello World');
+      });
+      await page.goto(server.EMPTY_PAGE);
+      page.on('request', request => {
+        void request.continue();
+      });
+      await page.setRequestInterception(true);
+      const redirectRequest = page.waitForRequest(
+        request => {
+          return request.url().endsWith('/rredirect');
+        },
+        {
+          timeout: 1000,
+        },
+      );
+      const targetResponse = page.waitForResponse(
+        response => {
+          return response.request().url().endsWith('/target');
+        },
+        {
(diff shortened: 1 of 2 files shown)


OrKoN:
@whimboo it looks like Firefox also does not instrument keepalive fetch requests correctly? do you happen to have a tracking issue for this?

test/src/requestinterception.spec.ts
```diff
@@ -80,6 +80,49 @@ describe('request interception', function () {
         page.waitForNavigation(),
       ]);
     });
+    it('should work with keep alive redirects', async () => {
+      const {page, server} = await getTestState();
+      server.setRoute('/rredirect', (_req, res) => {
```
sadym-chromium, line 85:
```suggestion
      server.setRoute('/redirect', (_req, res) => {
```

sadym-chromium (approved):

test/src/requestinterception.spec.ts
OrKoN: This is on purpose for easier search in logs.

sadym-chromium: Sgtm

sadym-chromium: Sgtm

whimboo:
> @whimboo it looks like Firefox also does not instrument keepalive fetch requests correctly? do you happen to have a tracking issue for this?

Sorry for being a bit late. I'm not aware of that yet, but would like to forward the question to @juliandescottes. Given that we rely on the DevTools network observer I assume that maybe https://bugzilla.mozilla.org/show_bug.cgi?id=1935107 plays a role here. Given that Chrome fails as well, I think it has lower priority and it can wait until Julian is back next year.

juliandescottes:
This was normally fixed on Firefox side in Bug [1897424](https://bugzilla.mozilla.org/show_bug.cgi?id=1897424). Looks like the FAIL expectation has been removed for Firefox since this PR got merged so everything seems fine?

Edit: was removed at https://github.com/puppeteer/puppeteer/pull/14559/changes#diff-9779ad852dcb7b4cb9726309f96e15c7675cdff7b42ec48d276ec12917b06e1fL805-L811 when updating to Firefox 147, which matches Bug 1897424.

whimboo:
Ah, good to know that this was outdated. Thanks!

PR #14543, opened 2026-01-07
feat: add fallback download sources

goosewobbler:
### **What kind of change does this PR introduce?**
Feature - Adds support for alternative download sources when primary browser downloads fail.

### **Did you add tests for your changes?**
Yes - Added comprehensive test coverage including:
- API acceptance tests for the `fallbackSources` option
- Validation tests for predefined `FallbackSources` constants  
- Platform mapping validation through URL generation
- Browser compatibility validation for predefined sources

### **If relevant, did you update the documentation?**
No - Documentation updates can be added in a follow-up PR if this feature is accepted.

### **Summary**

This PR addresses the issue of unavailable browser binaries for certain platforms (notably Linux ARM64) by adding a `fallbackSources` option to the `install()` function. When primary download sources fail, Puppeteer will now try alternative sources before falling back to the Chrome for Testing dashboard.

**Motivation:**
- Linux ARM64 platforms cannot download Chrome/Chromium from official sources
- Users currently resort to manual workarounds (Docker, custom scripts)
- Alternative sources like Playwright and Electron provide ARM64 binaries
- This enables native Puppeteer support for platforms currently unsupported

**Key Features:**
- `fallbackSources` option accepts array of alternative download configurations
- Predefined `FallbackSources` constants for common providers (Playwright, Electron)
- Custom URL builders for different URL patterns
- Automatic platform name mapping
- Backward compatible - no breaking changes

**Usage Examples:**
```typescript
// Use Playwright builds for ARM64 Chromium
await install({
  browser: Browser.CHROMIUM,
  buildId: '119.0.6045.105',
  platform: BrowserPlatform.LINUX_ARM,
  fallbackSources: [FallbackSources.PLAYWRIGHT_CHROMIUM]
});

// Custom fallback source
await install({
  browser: Browser.CHROMIUM,
  buildId: '119.0.6045.105', 
  platform: BrowserPlatform.LINUX_ARM,
  fallbackSources: [{
    baseUrl: 'https://playwright.azureedge.net/builds/chromium/',
    urlBuilder: (browser, platform, buildId, baseUrl) => 
      `${baseUrl}${buildId}/chromium-${mapPlatform(platform)}.zip`
  }]
});
```

### **Does this PR introduce a breaking change?**
No - This is fully backward compatible. Existing code continues to work unchanged.

### **Other information**

**Implementation Details:**
- Added `fallbackSources?: Array<{baseUrl: string, urlBuilder?: Function}>` to `InstallOptions`
- Extended fallback logic: Primary source → Custom fallbacks → Chrome for Testing → Error
- Added platform mapping functions for Electron (`linux-x64`) and Playwright (`linux-arm64`) formats
- Exported `FallbackSources` constants for convenience
- Replaced non-null assertion with explicit validation for better error handling

**Addresses:**
- [Issue #7740](https://github.com/puppeteer/puppeteer/issues/7740) - ARM64 Chromium support
- Enables native ARM64 Linux support without manual workarounds
- Provides extensible framework for future alternative sources

**Testing Strategy:**
- Unit tests for API validation and URL building
- Integration tests for predefined constants
- No full E2E tests due to network dependencies (following existing Puppeteer patterns)

This provides a clean, extensible solution for cross-platform browser downloads while maintaining full backward compatibility. 🚀

### DIFF
--- a/packages/browsers/src/install.ts
+++ b/packages/browsers/src/install.ts
@@ -27,6 +27,46 @@ import {downloadFile, getJSON, headHttpRequest} from './httpUtil.js';
 
 const debugInstall = debug('puppeteer:browsers:install');
 
+/**
+ * Maps BrowserPlatform to Electron release platform names
+ */
+function mapPlatformForElectron(platform: BrowserPlatform): string {
+  switch (platform) {
+    case BrowserPlatform.LINUX:
+      return 'linux-x64';
+    case BrowserPlatform.LINUX_ARM:
+      return 'linux-arm64';
+    case BrowserPlatform.MAC:
+      return 'darwin-x64';
+    case BrowserPlatform.MAC_ARM:
+      return 'darwin-arm64';
+    case BrowserPlatform.WIN32:
+      return 'win32-ia32';
+    case BrowserPlatform.WIN64:
+      return 'win32-x64';
+  }
+}
+
+/**
+ * Maps BrowserPlatform to Playwright platform names
+ */
+function mapPlatformForPlaywright(platform: BrowserPlatform): string {
+  switch (platform) {
+    case BrowserPlatform.LINUX:
+      return 'linux';
+    case BrowserPlatform.LINUX_ARM:
+      return 'linux-arm64';
+    case BrowserPlatform.MAC:
+      return 'mac';
+    case BrowserPlatform.MAC_ARM:
+      return 'mac-arm64';
+    case BrowserPlatform.WIN32:
+      return 'win32';
+    case BrowserPlatform.WIN64:
+      return 'win64';
+  }
+}
+
 const times = new Map<string, [number, number]>();
 function debugTime(label: string) {
   times.set(label, process.hrtime());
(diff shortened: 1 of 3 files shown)


packages/browsers/src/install.ts
```diff
@@ -43,6 +83,37 @@ function debugTimeEnd(label: string) {
...
+ */
+export const FallbackSources = {
+  /**
+   * Electron releases - useful for Chromedriver on platforms where Chrome releases aren't available
+   */
+  ELECTRON: {
+    baseUrl: 'https://github.com/electron/electron/releases/download/',
+    urlBuilder: (browser: Browser, platform: BrowserPlatform, buildId: string, baseUrl: string) => {
+      if (browser !== Browser.CHROMEDRIVER) {
+        throw new Error('Electron fallback is only supported for Chromedriver');
```
reneleonhardt, line 97:
```suggestion
        throw new Error(`Electron fallback is only supported for Chromedriver, not "${browser}"`);
```

OrKoN:
Thanks for the PR but it might not be a good fit for the following reasons:

1) the fallback sources might not have the same release cycles and version availability resulting in the version differences between platforms.
2) the content of the fallback archives might be (in the future) different from what CfT archives provide (e.g., affecting executable path calculations).
3) the fallback downloads might not integrate well with other features of Puppeteer such as launching the browsers or the browser download via the puppeteer package installation or the CLI.
4) we do not have test coverage or control for the fallback content and cannot guarantee that Puppeteer works with specific fallback binaries.

I am afraid that for now we would have to ask the Linux arm64 users to handle downloads manually or via other automation tools. @goosewobbler perhaps you would be interested in publishing fallback downloads as a separate npm module that could be added by the users who need Linux arm64 to the their postinstall/cli scripts to handle downloads on these platforms? WDYT?

goosewobbler:
@OrKoN Ok, I was about to rewrite this as it was both the wrong / inverted logic and poorly integrated with the existing code anyway.

I think as far as a separate NPM package goes, I'm not sure of the value here unless we are talking specifically about `@puppeteer/browsers` moving towards some kind of pluggable architecture / dependency injection for downloads where the puppeteer download logic can be directly replaced / augmented with limited configuration.

For our specific use case I guess we'll handle this in `@wdio/utils` alongside our existing geckodriver/edgedriver download logic, if it looks like a good candidate for extracting to a separate package with direct benefit for other puppeteer users then we will consider it, but tbh I expect we will stop there.

PR #14552, opened 2026-01-10
feat: pluggable browser providers

### LINKED ISSUE #81: Electron Service: v10.0.0 Release
TODO:
- [x] Update deps
- [ ] Final prerelease
- [ ] Manual test
- [ ] Release

goosewobbler:

Feature enhancement - adds pluggable provider support to `@puppeteer/browsers`.

Yes

Yes

## Summary

This PR introduces pluggable provider support to `@puppeteer/browsers`, allowing users to define custom download sources for browser binaries beyond the default infrastructure, e.g. Chrome for Testing.

### Key Changes

1. **BrowserProvider Interface**: Defines the contract for custom providers
   ```typescript
   interface BrowserProvider {
     supports(options: DownloadOptions): Promise<boolean> | boolean
     getDownloadUrl(options: DownloadOptions): Promise<URL | null> | URL | null
     getExecutablePath?(options: {
       browser: Browser;
       buildId: string;
       platform: BrowserPlatform;
     }): Promise<string> | string  // Optional
   }
   ```

2. **DownloadOptions**: Type-safe interface for provider operations

3. **DefaultProvider**: Default implementation extracted for reuse

4. **Enhanced Install Function**: Now accepts an optional `providers` array for custom sources with automatic fallback to default

5. **Custom Executable Path Support**: Allows providers to specify non-standard binary locations within archives

### Usage Example

```typescript
import { install, Browser, BrowserPlatform } from '@puppeteer/browsers';

// Install with custom provider fallback
await install({
  browser: Browser.CHROMEDRIVER,
  buildId: '142.0.7444.175',
  platform: BrowserPlatform.LINUX_ARM,
  cacheDir: './cache',
  providers: [
    new ElectronProvider(), // Try custom source first
    // Falls back to default provider automatically
  ],
});
```

### Benefits

- **Extensibility**: Streamlines the addition of new provider implementations for any browser or source.
- **Resilient Fallbacks**: Implements a robust provider chain, ensuring seamless fallback to alternative sources if primary downloads fail.
- **Broad Platform Support**: Provides the architecture to enable downloads on platforms not officially covered by standard distributions (e.g., Linux ARM64) via community or custom sources.
- **Enterprise Flexibility**: Facilitates the use of internal mirrors or restricted download sources for secure environments.
- **Separation of Concerns**: Decouples URL/path resolution logic from the core engine, which now handles downloading and persistence uniformly.

No - all changes are additive. Existing code continues to work unchanged. The new `providers` option is optional and existing default sources (e.g. Chrome for Testing) remains the default.

**Other information**

This PR establishes the architectural foundation for pluggable providers while maintaining full backward compatibility. The plugin system enables future enhancements like ARM64 Linux support and enterprise deployment scenarios.

⚠️ **IMPORTANT**: Custom providers are NOT officially supported by Puppeteer. By implementing this interface, you accept full responsibility for ensuring downloaded binaries are compatible with Puppeteer's expectations, testing that browser launch and other features work, and maintaining compatibility when Puppeteer or your download source changes. Puppeteer only tests and guarantees Chrome for Testing binaries.


goosewobbler:
@OrKoN Leading on from https://github.com/puppeteer/puppeteer/pull/14543#issuecomment-3723807815, I took another crack at this issue with a different approach.  It's verified working for our usecase.  What do you think?

OrKoN:
I think a plugin architecture would be nice, will review in details next week.

packages/browsers/src/Cache.ts
```diff
@@ -39,16 +39,28 @@ export class InstalledBrowser {
...
+
+    // If a path template is provided, expand it and use that
+    if (executablePathTemplate) {
+      const installationDir = cache.installationDir(browser, platform, buildId);
+      this.executablePath = path.join(
+        installationDir,
+        expandPathTemplate(executablePathTemplate, {platform, buildId}),
+      );
+    } else {
+      // Otherwise use default Chrome for Testing structure
```
OrKoN: nit: this also covers Firefox and Chromium so let's not call it  Chrome for Testing

packages/browsers/src/default-downloader.ts
```diff
@@ -0,0 +1,44 @@
...
+import {downloadUrls} from './browser-data/browser-data.js';
+import type {BrowserDownloader, DownloadOptions} from './downloader.js';
+
+/**
+ * Default downloader implementation that uses Chrome for Testing (CfT) sources.
+ * This is the standard downloader used by Puppeteer.
+ *
+ * @public
+ */
+export class ChromeForTestingDownloader implements BrowserDownloader {
```
OrKoN: Let's call the default downloader as it covers Firefox and Chromium that are not CfT

packages/browsers/src/default-downloader.ts
```diff
@@ -0,0 +1,44 @@
...
+  #baseUrl?: string;
+
+  constructor(baseUrl?: string) {
+    this.#baseUrl = baseUrl;
+  }
+
+  supports(_options: DownloadOptions): boolean {
+    // Chrome for Testing supports all browsers
```
OrKoN: same

packages/browsers/src/httpUtil.ts
```diff
@@ -11,6 +11,16 @@ import {URL, urlToHttpOptions} from 'node:url';
...
 
+/**
+ * Checks if a URL is accessible using a HEAD request.
+ * Automatically follows HTTP redirects.
+ * Useful for verifying download URLs exist before attempting download.
+ *
+ * @param url - The URL to check
+ * @returns Promise that resolves to true if final status is 200, false otherwise
+ *
+ * @public
```
OrKoN: I think this should be internal

packages/browsers/src/httpUtil.ts
```diff
@@ -70,7 +80,16 @@ export function httpRequest(
...
+ * Downloads a file from the specified URL to a destination path.
+ * Automatically follows HTTP redirects and supports progress callbacks.
+ *
+ * @param url - The URL to download from
+ * @param destinationPath - Local file path where the file will be saved
+ * @param progressCallback - Optional callback for download progress updates
+ * @returns Promise that resolves when download completes
+ * @throws Error if download fails or final response status is not 200
+ *
+ * @public
```
OrKoN: internal?

packages/browsers/src/httpUtil.ts
```diff
@@ -123,6 +142,17 @@ export function downloadFile(
...
+/**
+ * Fetches JSON data from a URL.
+ * Automatically follows HTTP redirects.
+ * Useful for custom downloaders that need to query release APIs.
+ *
+ * @param url - The URL to fetch JSON from
+ * @returns Promise that resolves to the parsed JSON data
+ * @throws Error if request fails, final status \>= 400, or JSON parsing fails
+ *
+ * @public
```
OrKoN: internal?

packages/browsers/src/httpUtil.ts
```diff
@@ -132,6 +162,16 @@ export async function getJSON(url: URL): Promise<unknown> {
...
 
+/**
+ * Fetches text content from a URL.
+ * Automatically follows HTTP redirects.
+ *
+ * @param url - The URL to fetch text from
+ * @returns Promise that resolves to the text content
+ * @throws Error if the request fails or final response status is \>= 400
+ *
+ * @public
```
OrKoN: internal?

packages/browsers/src/install.ts
```diff
@@ -113,6 +115,145 @@ export interface InstallOptions {
...
+        continue;
+      }
+
+      // Warn if using non-CfT downloader
+      if (!(downloader instanceof ChromeForTestingDownloader)) {
+        debugInstall(
+          `⚠️  Using custom downloader: ${downloader.constructor.name}`,
+        );
+        debugInstall(
+          `⚠️  Puppeteer does not guarantee compatibility with non-Chrome-for-Testing binaries`,
```
OrKoN:
```suggestion
          `⚠️  Puppeteer does not guarantee compatibility with non-default binaries`,
```

perhaps?

OrKoN (changes requested):
How would it work with https://pptr.dev/browsers-api/browsers.getinstalledbrowsers and launch? I am not sure a downloader is a good name since it also covers the executable path calculation. Perhaps BrowserProvider could be a better name for the interface?

goosewobbler:
@OrKoN 

- Renamed the interface to `BrowserProvider` to better reflect its role
- Using `provider(s)` where we were using `downloader(s)`
- Using `default` in place of `Chrome for Testing` / `CfT`
- Now persisting the relative executable path in a `.puppeteer.json` file alongside the binary, `launch` and `getInstalledBrowsers` can resolve custom paths via the `Cache` without needing the provider or hardcoded defaults, making custom installations first-class citizens.

docs/guides/custom-browser-providers.md
```diff
@@ -0,0 +1,145 @@
+# Custom Browser Providers
```
OrKoN: Generally guides are for the puppeteer / puppeteer-core package. Let me think if we can have browsers guides there as it is usually an internal detail for Puppeteer users.

OrKoN:
Sorry, the reviewing will take some time due to other workload.

docs/guides/custom-browser-providers.md
goosewobbler: Nice one, let me know - if not, we could also just strip out the fluff from this and combine what's left with the browsers package README.

goosewobbler: I pre-emptively moved this to the browsers readme.  Can always revert, or move it elsewhere.

packages/browsers/src/Cache.ts
```diff
@@ -39,16 +39,28 @@ export class InstalledBrowser {
     browser: Browser,
     buildId: string,
     platform: BrowserPlatform,
+    executablePathTemplate?: string,
```
OrKoN: I think instead of passing the template here it would be better to extend the Cache class when installing with a provider and implement the different computeExecutablePath logic.

packages/browsers/src/Cache.ts
```diff
@@ -251,6 +263,17 @@ export class Cache {
       options.platform,
       options.buildId,
     );
+    const metadataPath = path.join(installationDir, '.puppeteer.json');
```
OrKoN: We already have a metadata file, can we re-use it instead of introducing a new way to store metadata?

OrKoN: To simplify things, we could write down the executable path for all installed browsers (not only custom providers) and read it from the metadata cache consistently via the Cache class. I wonder if this change could be separated into a separate PR for easier review.

packages/browsers/src/httpUtil.ts
```diff
@@ -11,6 +11,16 @@ import {URL, urlToHttpOptions} from 'node:url';
 
 import {ProxyAgent} from 'proxy-agent';
 
+/**
```
OrKoN: I appreciate the improvements to the docs but it would be easier to review if they were in a separate PR.

packages/browsers/src/provider.ts
```diff
@@ -0,0 +1,153 @@
...
+ * @public
+ */
+export interface DownloadOptions {
+  browser: Browser;
+  platform: BrowserPlatform;
+  buildId: string;
+  /**
+   * Progress callback for download progress.
+   */
+  progressCallback?: (downloadedBytes: number, totalBytes: number) => void;
```
OrKoN: do the providers need to receive the progress callback?

packages/browsers/src/Cache.ts
OrKoN: I think we would need to expand templates if we implement that.

packages/browsers/src/provider.ts
goosewobbler: Removed this

packages/browsers/src/httpUtil.ts
goosewobbler: Removed these

packages/browsers/src/Cache.ts
goosewobbler: I'm keeping it in this branch  for now so I can test the whole thing E2E, once I've checked everything passes I'll create a new PR.

packages/browsers/src/Cache.ts
```diff
@@ -156,6 +165,56 @@ export class Cache {
     fs.writeFileSync(metatadaPath, JSON.stringify(metadata, null, 2));
   }
 
+  installationMetadataPath(
```
OrKoN: can we use existing readMetadta/writeMetadata? we do need to put it necessarily into the browser installation folder, we can write a map in the existing metadata file I believe.

packages/browsers/src/install.ts
```diff
@@ -299,12 +390,40 @@ async function installUrl(
 
   try {
     if (existsSync(outputPath)) {
+      // Get executable path template from provider if provided
+      let pathTemplate: string | undefined;
```
OrKoN: why do we need to get a template? Can we get the actual executable path? it seems we call provider?.getExecutablePath with all params needed to resolve the template internally in the provider implementation.

packages/browsers/src/install.ts
```diff
@@ -299,12 +390,40 @@ async function installUrl(
...
+      if (provider?.getExecutablePath) {
```
OrKoN: can we make getExecutablePath non-optional in the browser interface? default provider should be able to compute the default executable path?

goosewobbler:
Did a lot of cleanup and added two enhancements:

1. Provider Metadata: Tracks which provider installed each browser in .metadata (shown in browsers list output with `[ProviderName]`)
2. Error Granularity: Error messages now show provider names, e.g., `MockProvider: Network error` instead of just the error message

OrKoN:
> Provider Metadata: Tracks which provider installed each browser in .metadata (shown in browsers list output with [ProviderName])

why do we need to know that? can we only write the resolved executable path without templates at the install time as determined by the provider?

goosewobbler:
> > Provider Metadata: Tracks which provider installed each browser in .metadata (shown in browsers list output with [ProviderName])
> 
> why do we need to know that?

Rationale:
1. Debugging/Visibility: Knowing which provider installed a browser helps diagnose problems
2. Custom Provider Support: Tracking which provider was used helps users verify their custom provider was actually used
3. Future Extensibility: If provider-specific behavior is needed later (e.g., provider-specific updates), this
metadata would be essential
4. Minimal Overhead: It's just a string stored alongside other metadata
   
Obviously, happy to remove it if not required.

OrKoN:
@goosewobbler please do not rebase so that I can view test results! I will rebase once it is ready to land.

packages/browsers/src/Cache.ts
```diff
@@ -156,6 +167,56 @@ export class Cache {
...
+    browser: Browser,
+    platform: BrowserPlatform,
+    buildId: string,
+  ): string | undefined {
+    const metadata = this.readMetadata(browser);
+    const key = `${platform}-${buildId}`;
+    return metadata.providers?.[key];
+  }
+
+  writeProvider(
```
OrKoN: We do not need write provider as we only write the provider during the installation, let's remove writeProvider and pass provider as a parameter to executable path.

packages/browsers/src/Cache.ts
```diff
@@ -194,6 +255,12 @@ export class Cache {
         delete metadata.aliases[alias];
       }
     }
+    // Clean up executable path entry
+    const key = `${platform}-${buildId}`;
+    if (metadata.executablePaths?.[key]) {
+      delete metadata.executablePaths[key];
```
OrKoN, line 227: should we clean up providers as well?

packages/browsers/src/install.ts
```diff
@@ -297,14 +398,39 @@ async function installUrl(
...
+
+  const installedBrowser = new InstalledBrowser(
+    cache,
+    options.browser,
+    options.buildId,
+    options.platform,
+  );
+
+  // Write metadata for the installation
+  cache.writeExecutablePath(
```
OrKoN: let's write the provider and path only for non-default providers. We do not want to make the default provider only work as long as the class name matches.

packages/browsers/src/install.ts
```diff
@@ -297,14 +398,39 @@ async function installUrl(
...
+    options.browser,
+    options.platform,
+    options.buildId,
+    relativeExecutablePath,
+  );
+  cache.writeProvider(
+    options.browser,
+    options.platform,
+    options.buildId,
+    provider.constructor.name,
```
OrKoN: provider.constructor.name is not reliable. What is the code is minified? please add a getName() to the provider interface instead (or since we do not really need the provider name except for the reporting in the CLI, remove it altogether).

packages/browsers/test/src/provider/custom-provider-integration.spec.ts
```diff
@@ -0,0 +1,346 @@
+/**
+ * @license
+ * Copyright 2023 Google Inc.
```
OrKoN: all new files should have current year in the header

packages/browsers/test/src/provider/provider.spec.ts
```diff
@@ -0,0 +1,175 @@
...
+  buildArchiveFilename,
+  Browser,
+  BrowserPlatform,
+  type BrowserProvider,
+  type DownloadOptions,
+} from '../../../lib/esm/main.js';
+
+import {MockProvider} from './mock-provider.js';
+
+describe('BrowserProvider Interface', () => {
```
OrKoN: why do we need to test the interface? are we testing buildArchiveFilename? where is it used? I think the entire test can be dropped as it seems to test the mock only.

packages/browsers/src/DefaultProvider.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+import path from 'node:path';
+
+import {
+  Browser,
+  BrowserPlatform,
+  DefaultProvider,
+} from '../../../lib/esm/main.js';
+import {getServerUrl, setupTestServer} from '../utils.js';
+
+describe('DefaultProvider', () => {
```
OrKoN, line 11: I think this should be a unit test (a test file next to DefaultProvider)

packages/browsers/src/DefaultProvider.ts
OrKoN, line 1: The file name should be DefaultProvider.ts to match other files in this dir.

packages/browsers/test/src/provider/default-provider-integration.spec.ts
```diff
@@ -0,0 +1,89 @@
...
+import path from 'node:path';
+
+import {
+  Browser,
+  BrowserPlatform,
+  DefaultProvider,
+} from '../../../lib/esm/main.js';
+import {getServerUrl, setupTestServer} from '../utils.js';
+
+describe('DefaultProvider Integration', () => {
```
OrKoN: existing downloader was already covered by integration tests. Should this be a unit test with removed integration test aspects or removed completely?

packages/browsers/test/src/provider/custom-provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+  });
+
+  afterEach(() => {
+    if (fs.existsSync(tmpDir)) {
+      fs.rmSync(tmpDir, {recursive: true, force: true});
+    }
+  });
+
+  describe('basic functionality', () => {
+    it('should work with no custom providers (default CfT)', async function () {
```
OrKoN: this default is already tested and should not tested in custom-provider.spec.ts

packages/browsers/test/src/provider/custom-provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+          cacheDir: tmpDir,
+          browser: Browser.CHROME,
+          platform: BrowserPlatform.LINUX,
+          buildId: 'non-existent-build-12345',
+        }),
+        /Download failed/,
+      );
+    });
+
+    it('should accept custom providers array', async function () {
```
OrKoN: this is tested by the type checking

packages/browsers/test/src/provider/custom-provider.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+        providers: [customProvider],
+      });
+
+      assert(result);
+      assert.strictEqual(typeof result.path, 'string');
+    });
+  });
+
+  describe('provider interface compliance', () => {
+    it('should handle providers that implement the interface', async function () {
```
OrKoN: isn't it handled by type checking as well?

packages/browsers/test/src/provider/mock-provider.ts
```diff
@@ -0,0 +1,58 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import type {BrowserProvider, DownloadOptions} from '../../../lib/esm/main.js';
+
+/**
+ * Mock downloader implementation for testing.
+ * Allows configurable behavior for testing different downloader scenarios.
+ */
+export class MockProvider implements BrowserProvider {
```
OrKoN: can we use sinon instead of custom mocks? (if mock provider is needed)

OrKoN (changes requested):

OrKoN:
Thanks @goosewobbler I think it looks good in the implementation part (except for points I left the comments for) but tests need to be improved (mostly remove redundant coverage, and move tests to unit tests where appropriate). Integration style tests take a while on github actions so we try to have a minimal amount of those.

packages/browsers/src/Cache.ts
goosewobbler: removed them from cache

packages/browsers/src/DefaultProvider.spec.ts
goosewobbler: What do you mean by this?  move the whole file to `src/DefaultProvider.spec.ts`?

OrKoN: Yes, that's right

packages/browsers/src/DefaultProvider.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  setupTestServer();
+
+  let downloader: DefaultProvider;
+  let tmpDir: string;
+
+  beforeEach(() => {
+    tmpDir = fs.mkdtempSync(path.join(os.tmpdir(), 'puppeteer-test'));
```
OrKoN: does it need tmpDir? downloader also can be created in each test. Some tests do not need it. The var should probably renamed to provider.

packages/browsers/test/src/provider/custom-provider.spec.ts
OrKoN, line 1: let's drop the -integration from the test file name

OrKoN (approved):
LGTM thanks!

packages/browsers/src/install.ts
```diff
@@ -113,6 +115,157 @@ export interface InstallOptions {
    * @defaultValue `false`
    */
   installDeps?: boolean;
+  /**
```
OrKoN, line 118: docs need to be updated to reflect the downloader to provider change

packages/browsers/src/httpUtil.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2023 Google Inc.
+ * Copyright 2026 Google Inc.
```
OrKoN:
```suggestion
 * Copyright 2023 Google Inc.
```

packages/browsers/src/Cache.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2023 Google Inc.
+ * Copyright 2026 Google Inc.
```
OrKoN:
```suggestion
 * Copyright 2023 Google Inc.
```

packages/browsers/src/install.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2017 Google Inc.
+ * Copyright 2026 Google Inc.
```
OrKoN:
```suggestion
 * Copyright 2017 Google Inc.
```

packages/browsers/src/main.ts
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2023 Google Inc.
+ * Copyright 2026 Google Inc.
```
OrKoN:
```suggestion
 * Copyright 2023 Google Inc.
```

OrKoN:
Now there seem to be only a test failure to resolve

PR #14569, opened 2026-01-15
test: increase protocol timeout in waittask test

juliandescottes:
Locally I can see that BiDi in general is slower to connect than CDP. This plus the fact that the mac workers for github actions are known to be quite slow I strongly suspect the frequent intermittent is only due to a minor slowdown of the browser startup session creation, and should rather be addressed in the test.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -1782,13 +1782,6 @@
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
   {
-    "testIdPattern": "[waittask.spec] waittask specs protocol timeout *",
-    "platforms": ["darwin"],
-    "parameters": ["firefox", "headful", "webDriverBiDi"],
-    "expectations": ["SKIP"],
-    "comment": "session.new before the test consistently times out on mac + headful since 147"
-  },
-  {
     "testIdPattern": "[worker.spec] Workers can be closed",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["chrome", "webDriverBiDi"],
--- a/test/src/waittask.spec.ts
+++ b/test/src/waittask.spec.ts
@@ -931,7 +931,7 @@ describe('waittask specs', function () {
 
   describe('protocol timeout', () => {
     const state = setupSeparateTestBrowserHooks({
-      protocolTimeout: 3000,
+      protocolTimeout: 4000,
     });
 
     it('should error if underyling protocol command times out with raf polling', async () => {


juliandescottes:
I would like to kick some CI jobs to verify the assertions first. cc @OrKoN FYI

juliandescottes:
Actually @OrKoN I can't trigger the workflows myself, could you kick off the firefox tests at least here?

OrKoN:
Thanks! I thought I have consolidated various timeouts previously to increase them automatically but let's see if helps

juliandescottes:
Thanks! This one seems to have passed https://github.com/puppeteer/puppeteer/actions/runs/21027640969/job/60456391159#step:12:430

OrKoN:
it was not failing on all platforms, only headful mac.

juliandescottes:
Indeed, still failed on macos headful. I think I would like to increase more significantly the timeout. If that still fails with double the time, that probably means my suggestion is wrong.

OrKoN:
I think this would override the timeout from the test https://github.com/puppeteer/puppeteer/blob/main/test/src/mocha-utils.ts#L174 probably incorrectly. The issue that we cannot set a protocol timeout for the session.new to be different from the timeout the test needs. Although I am not sure if higher timeout helps as I doubled it in the previous CL and it still failed.

juliandescottes:
Alright, I guess we need trace logs then, as this can't be reproduced locally on macos headful.

juliandescottes:
@OrKoN to get trace logs I think we need to run the actions in debug mode? Is that something you can trigger or should I rather force it in some other way?

OrKoN:
Triggered on https://github.com/puppeteer/puppeteer/actions/runs/21028723392/job/60489860049?pr=14569

juliandescottes:
Looks like on CI we hang on "Waiting for initial application window"

> 2026-01-15T16:00:34.4230890Z   '{"id":1,"method":"session.new","params":{"capabilities":{"alwaysMatch":{"acceptInsecureCerts":false,"unhandledPromptBehavior":{"default":"ignore"},"webSocketUrl":true,"goog:prerenderingDisabled":true,"goog:disableNetworkDurableMessages":true}}}}'
2026-01-15T16:00:34.4231630Z ]
2026-01-15T16:00:34.4547410Z 1768492834454	RemoteAgent	DEBUG	WebDriverBiDiConnection 3fdea172-2a0a-415d-bded-e463a1099508 accepted
2026-01-15T16:00:34.4553260Z 1768492834454	RemoteAgent	DEBUG	WebDriverBiDiConnection 3fdea172-2a0a-415d-bded-e463a1099508 -> {"id":1,"method":"session.new","params":{"capabilities":{"alwaysMatch":{"acceptInsecureCerts":false,"unhandledPromptBehavior":{"default":"ignore"},"webSocketUrl":true,"goog:prerenderingDisabled":true,"goog:disableNetworkDurableMessages":true}}}}
2026-01-15T16:00:34.4571300Z 1768492834456	WebDriver BiDi	DEBUG	Connection 3fdea172-2a0a-415d-bded-e463a1099508 attached to session 50c35bb5-7b3a-482c-973c-6bad79f52372
2026-01-15T16:00:34.4580870Z 1768492834457	WebDriver BiDi	DEBUG	Waiting for initial application window

juliandescottes:
Switching to draft, looks like the issue is not purely about the timeout. @OrKoN is there any way to have a screenshot at failure time, or a video recording?

OrKoN:
> Switching to draft, looks like the issue is not purely about the timeout. @OrKoN is there any way to have a screenshot at failure time, or a video recording?

no, there is no support for this in the infra. You can probably try to do something with ffmpeg but not sure if it would work on mac.

juliandescottes:
The fix for the macos slowness has been addressed on the Firefox side and should be in the next nightly: https://hg-edge.mozilla.org/mozilla-central/rev/73419c7048c3

I will send another PR to update the canary expectations.

PR #14582, opened 2026-01-20
docs: Add missing await to `asPage()` call in extensions documentation

oliverdunk:

Small fix in the documentation.

NA

Yes


Lightning00Blade (approved):
@OrKoN  There is EsLint Markdown plugin that should catch such Error, maybe we can explore it at some point.

PR #14585, opened 2026-01-20
chore(webdriver): bump chromium-bidi to 13.0.0

sadym-chromium:


OrKoN:
Looks like some tests fail

sadym-chromium:
> Looks like some tests fail

The failed tests were not only on BiDi but on unrelated tests as well. Retry restored it, so I consider it was just a flake.

OrKoN (approved):

PR #14618, opened 2026-01-28
feat(webdriver): Implement window management

sadym-chromium:
Use `browser.setClientWindowState`, `browser.getClientWindows` and rely on `clientWindow`.

Out of scope: update page's `windowId` when the page changes window.

### DIFF
--- a/packages/puppeteer-core/src/bidi/Browser.ts
+++ b/packages/puppeteer-core/src/bidi/Browser.ts
@@ -25,8 +25,7 @@ import type {Page} from '../api/Page.js';
 import type {Target} from '../api/Target.js';
 import type {Connection as CdpConnection} from '../cdp/Connection.js';
 import type {SupportedWebDriverCapabilities} from '../common/ConnectOptions.js';
-import {ProtocolError} from '../common/Errors.js';
-import {UnsupportedOperation} from '../common/Errors.js';
+import {ProtocolError, UnsupportedOperation} from '../common/Errors.js';
 import {EventEmitter} from '../common/EventEmitter.js';
 import {debugError} from '../common/util.js';
 import type {Viewport} from '../common/Viewport.js';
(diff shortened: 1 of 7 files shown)


packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -306,15 +305,48 @@ export class BidiBrowser extends Browser {
...
+        ...windowBounds,
+      };
+    } else {
+      if (
+        windowBounds.height !== undefined ||
+        windowBounds.width !== undefined ||
+        windowBounds.left !== undefined ||
+        windowBounds.top !== undefined
+      ) {
+        throw new ProtocolError(
```
Lightning00Blade: Do we need this explicit error? I would assume the underlying implementation to fail in this case or it just ignores this values?

packages/puppeteer-core/src/bidi/BrowserContext.ts
```diff
@@ -209,6 +209,16 @@ export class BidiBrowserContext extends BrowserContext {
...
       }
     }
+    if (options?.type === 'window' && options?.windowBounds !== undefined) {
+      try {
+        await this.browser().setWindowBounds(
+          context.windowId,
+          options.windowBounds,
+        );
+      } catch {
+        // Tolerate not supporting `browser.setClientWindowState`.
```
Lightning00Blade: I think here and for similar places we should add a debug logging, just in case someone is testing somethings that consistency fails.

Lightning00Blade (approved):

packages/puppeteer-core/src/bidi/BrowserContext.ts
sadym-chromium: I don't think we log any errors in Puppeteer. Please correct me if I'm wrong.

OrKoN: > I don't think we log any errors in Puppeteer. Please correct me if I'm wrong.

we do (see debugError helper)

packages/puppeteer-core/src/bidi/Browser.ts
Lightning00Blade: I would just leave the passed parameters as they and let the underlying implementation handle it, wdyt?

packages/puppeteer-core/src/bidi/BrowserContext.ts
sadym-chromium: indeed, missed it. Done.

packages/puppeteer-core/src/bidi/Browser.ts
sadym-chromium: We need some generic way to address it: https://github.com/GoogleChromeLabs/chromium-bidi/issues/4029

sadym-chromium: also the params in Puppeteer have different names comparing to BiDi, so simply forwarding all the params would be odd.

PR #14620, opened 2026-01-29
perf: optimize Locator.fill performance for large text

OrKoN:
This PR optimizes `Locator.fill()` to handle large text inputs more efficiently. Previously, it used `keyboard.type()` which was very slow for large amounts of text due to the overhead of multiple CDP commands per character. The new implementation sets the value directly in the browser context and dispatches the necessary events, matching the behavior of other modern automation tools while providing a massive performance boost.

---
*PR created automatically by Jules for task [3189730666937029107](https://jules.google.com/task/3189730666937029107) started by @OrKoN*

### DIFF
--- a/packages/puppeteer-core/src/api/locators/locators.ts
+++ b/packages/puppeteer-core/src/api/locators/locators.ts
@@ -68,6 +68,18 @@ export type LocatorClickOptions = ClickOptions & ActionOptions;
 /**
  * @public
  */
+export interface LocatorFillOptions extends ActionOptions {
+  /**
+   * The number of characters to type before switching to a faster fill-out
+   * method.
+   *
+   * @defaultValue `100`
+   */
+  typingThreshold?: number;
+}
+/**
+ * @public
+ */
 export interface LocatorScrollOptions extends ActionOptions {
   scrollTop?: number;
   scrollLeft?: number;
@@ -400,9 +412,10 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
   #fill<ElementType extends Element>(
     this: Locator<ElementType>,
     value: string,
-    options?: Readonly<ActionOptions>,
+    options?: Readonly<LocatorFillOptions>,
   ): Observable<void> {
     const signal = options?.signal;
+    const typingThreshold = options?.typingThreshold ?? 100;
     const cause = new Error('Locator.fill');
     return this._wait(options).pipe(
       this.operators.conditions(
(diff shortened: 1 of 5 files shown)


packages/puppeteer-core/src/api/locators/locators.ts
```diff
@@ -458,57 +458,27 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
...
                 case 'typeable-input':
-                  return from(
-                    (
-                      handle as unknown as ElementHandle<HTMLInputElement>
-                    ).evaluate((input, newValue) => {
-                      const currentValue = input.isContentEditable
-                        ? input.innerText
-                        : input.value;
-
-                      // Clear the input if the current value does not match the filled
```
OrKoN, line 469: I believe we still want this logic for short inputs to simulate typing and handle typing with pre-filled prefixes

packages/puppeteer-core/src/api/locators/locators.ts
```diff
@@ -453,62 +453,80 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
         )
           .pipe(
             mergeMap(inputType => {
+              if (
```
OrKoN: can you keep this implementation in the original place (switch)?

packages/puppeteer-core/src/api/locators/locators.ts
```diff
@@ -458,57 +458,97 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
...
+                          element.dispatchEvent(
+                            new Event('input', {bubbles: true}),
+                          );
+                          element.dispatchEvent(
+                            new Event('change', {bubbles: true}),
+                          );
+                        }, value),
+                      );
                     }),
                   );
```
OrKoN: can we extract this into a helper and re-use here and in the other-input case branch?

packages/puppeteer-core/src/api/locators/locators.ts
```diff
@@ -453,68 +453,87 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
...
-
-                      // Clear the input if the current value does not match the filled
-                      // out value.
-                      if (
-                        newValue.length <= currentValue.length ||
-                        !newValue.startsWith(input.value)
-                      ) {
-                        if (input.isContentEditable) {
-                          input.innerText = '';
+                  if (value.length < 100) {
```
OrKoN: make the threshold configurable in the fill() method with teh default being 100

OrKoN:
@Lightning00Blade could you please take a look as well?

Lightning00Blade (approved):

PR #14623, opened 2026-01-29
test: fix test "Page.waitForNetworkIdle should work with delayed response" to wait for 400ms or greater

whimboo:
Seen in our own CI when running Puppeteer tests which seem to have faster machines as on GitHub: https://bugzilla.mozilla.org/show_bug.cgi?id=2011113#c5

@OrKoN as we discussed here is the PR. Thanks


OrKoN (approved):

OrKoN:
Thanks!

PR #14628, opened 2026-01-30
Test previously skipped tests (mouse, geolocation) for Firefox

whimboo:
Follow-up from https://github.com/puppeteer/puppeteer/pull/14186. Lets see if those tests are still failing and if yes why it happens.

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -537,13 +537,6 @@
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
   {
-    "testIdPattern": "[mouse.spec] Mouse should select the text with mouse",
-    "platforms": ["win32"],
-    "parameters": ["firefox"],
-    "expectations": ["SKIP"],
-    "comment": "Suddenly started being flaky on Windows"
-  },
-  {
     "testIdPattern": "[navigation.spec] navigation Page.goto should send referer",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["webDriverBiDi"],
@@ -710,13 +703,6 @@
     "parameters": ["firefox", "webDriverBiDi"],
     "expectations": ["FAIL"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
-  },
-  {
-    "testIdPattern": "[page.spec] Page Page.setGeolocation should work",
-    "platforms": ["win32"],
-    "parameters": ["firefox"],
-    "expectations": ["SKIP"],
-    "comment": "Suddenly started being flaky on Windows"
   },
   {
     "testIdPattern": "[page.spec] Page Page.setJavaScriptEnabled setInterval should pause",
--- a/test/src/mouse.spec.ts
+++ b/test/src/mouse.spec.ts
@@ -84,7 +84,7 @@ describe('Mouse', function () {
     expect(newDimensions.width).toBe(Math.round(width + 104));
     expect(newDimensions.height).toBe(Math.round(height + 104));
   });
-  it('should select the text with mouse', async () => {
+  it.only('should select the text with mouse', async () => {
     const {page, server} = await getTestState();
 
     const text =
--- a/test/src/page.spec.ts
+++ b/test/src/page.spec.ts
@@ -424,7 +424,7 @@ describe('Page', function () {
   });
 
   describe('Page.setGeolocation', function () {
-    it('should work', async () => {
+    it.only('should work', async () => {
       const {page, server, context} = await getTestState();
 
       await context.overridePermissions(server.PREFIX, ['geolocation']);


whimboo:
@OrKoN is there a way for me that I can trigger the firefox tests on Windows workers? Otherwise I'm not able to reproduce the issue.

OrKoN:
> @OrKoN is there a way for me that I can trigger the firefox tests on Windows workers? Otherwise I'm not able to reproduce the issue.

You should be able to add  the full-ci label

whimboo:
Ah I see. So these two tests are still failing. I've triggered a debug build:
https://github.com/puppeteer/puppeteer/actions/runs/21511283463/job/61985728758?pr=14628

whimboo:
For the geolocation test we set the override successfully but when loading the page and calling [`navigator.geolocation.getCurrentPosition()`](https://github.com/puppeteer/puppeteer/actions/runs/21511283463/job/61985728758?pr=14628#step:11:7402) a [prompt appears](https://github.com/puppeteer/puppeteer/actions/runs/21511283463/job/61985728758?pr=14628#step:11:7409) to ask for user permissions, but [that got set for the user context before](https://github.com/puppeteer/puppeteer/actions/runs/21511283463/job/61985728758?pr=14628#step:11:4409).

Note that the script that is used to retrieve the location get already run while the page is loading. So maybe we hit a race here that the page has finished loading but the permissions aren't applied yet? I could try to use the `waitUntil: "load"` option so see if that makes a difference.

OrKoN:
waitUntil: "load" is the default in Puppeteer

whimboo:
For the mouse test the `textarea` contains the text `This is the text that we are going to try to select. Let's see how it goes.`. But when the text is selected with the mouse [the selection only contains `This is the text that we are going to try to select. Let's see how it g`](https://github.com/puppeteer/puppeteer/actions/runs/21511283463/job/61985728758?pr=14628#step:11:1216). 

I wonder if there is a difference in the font being used (different type or different size) which could cause that? It would be great to have screenshots for failures which might help to better investigate the failure.

whimboo:
> waitUntil: "load" is the default in Puppeteer

So why is Puppeteer then running the script to retrieve the location 8 times while the page is loading? In the tests we do not have parallel execution:

```
      await page.setGeolocation({longitude: 10, latitude: 10});
      const geolocation = await page.evaluate(() => {
```

OrKoN:
```
2026-01-30T10:57:46.752Z puppeteer:webDriverBiDi:RECV ◀ [
  '{"type":"event","method":"browsingContext.load","params":{"context":"4cd8d8e5-1e3e-4f02-b395-d8dca521eb07","timestamp":1769770666749,"url":"http://localhost:60074/empty.html","navigation":"7f5f00d4-d22a-4bac-a430-c7faba7da545"}}'
]
2026-01-30T10:57:46.756Z puppeteer:webDriverBiDi:RECV ◀ [
  '{"type":"success","id":201,"result":{"navigation":"7f5f00d4-d22a-4bac-a430-c7faba7da545","url":"http://localhost:60074/empty.html"}}'
]
2026-01-30T10:57:46.757Z puppeteer:webDriverBiDi:SEND ► [
  '{"id":202,"method":"emulation.setGeolocationOverride","params":{"coordinates":{"latitude":10,"longitude":10},"contexts":["4cd8d8e5-1e3e-4f02-b395-d8dca521eb07"]}}'
]
2026-01-30T10:57:46.767Z puppeteer:webDriverBiDi:RECV ◀ [ '{"type":"success","id":202,"result":{}}' ]
2026-01-30T10:57:46.767Z puppeteer:webDriverBiDi:SEND ► [
  '{"id":203,"method":"script.callFunction","params":{"functionDeclaration":"() => {\\n                return new Promise(resolve => {\\n                    return navigator.geolocation.getCurrentPosition(position => {\\n                        resolve({\\n                            latitude: position.coords.latitude,\\n                            longitude: position.coords.longitude,\\n                        });\\n                    });\\n                });\\n            }\\n//# sourceURL=pptr:evaluate;Context.%3Canonymous%3E%20(file%3A%2F%2F%2FD%3A%2Fa%2Fpuppeteer%2Fpuppeteer%2Ftest%2Fbuild%2Fpage.spec.js%3A399%3A44)\\n","awaitPromise":true,"target":{"context":"4cd8d8e5-1e3e-4f02-b395-d8dca521eb07"},"arguments":[],"resultOwnership":"none","userActivation":true,"serializationOptions":{}}}'
]
```

I do not see Puppeteer querying location during the page load. The empty page is also almost instantaneous to load. What you might be seeing is the test retries by mocha.

OrKoN:
Closing as stale

whimboo:
I should still get those two tests investigated and working.

whimboo:
@OrKoN I tried to run the Firefox tests for Windows headless in debug mode but as the following log output shows no trace logs are available:

https://github.com/puppeteer/puppeteer/actions/runs/25692519713/job/75437444571

Is something maybe broken?

OrKoN:
I see trace logs following the link. Is there a specific issue in the logs?

whimboo:
Oh that's interesting. Maybe it was a caching issue on GitHub? I clearly did not see any trace logs yesterday but now they are there. Strange.

Nevertheless here some analysis:

1. The geolocation test fails because retrieving the current position via `navigator.geolocation.getCurrentPosition` and `script.callFunction` [takes longer than 10s](https://github.com/puppeteer/puppeteer/actions/runs/25692519713/job/75437444571#step:11:7134). As such we reach the timeout of the test. It would be good to check if the command returns at all when the timeout gets increased or not.

2. The mouse selection test fails because we only return a part of the whole text that should be selected. The last characters are missing. See https://github.com/puppeteer/puppeteer/pull/14628#issuecomment-3823813289 for details. I wonder if there might be a line break which is causing the text not be selected? 

@OrKoN is there a way to instruct Puppeteer to take screenshots when a test fails?

OrKoN:
not unless you modify the test code.

OrKoN:
I can reproduce the issue with the mouse on a local windows? perhaps you could check the issue locally instead of running GitHub actions to debug (there is no line break where the text is cut)?

OrKoN:
I cannot reproduce the geolocation issue locally but it looks like Firefox takes more than 10 seconds to resolve an emulated location.

OrKoN:
@whimboo any updates on this PR or can we close it?

PR #14642, opened 2026-02-03
test: enable waittask tests on macos Firefox Nightly headful

juliandescottes:
With https://bugzilla.mozilla.org/show_bug.cgi?id=2010507 the waittask tests should pass again on macos headful.

### DIFF
--- a/test/CanaryTestExpectations.json
+++ b/test/CanaryTestExpectations.json
@@ -19,5 +19,12 @@
     "parameters": ["cdp"],
     "expectations": ["TIMEOUT"],
     "comment": "Needs investigation"
+  },
+  {
+    "testIdPattern": "[waittask.spec] waittask specs protocol timeout *",
+    "platforms": ["darwin"],
+    "parameters": ["firefox", "headful", "webDriverBiDi"],
+    "expectations": ["PASS"],
+    "comment": "Fixed by https://bugzilla.mozilla.org/show_bug.cgi?id=2010507"
   }
 ]


juliandescottes:
The bug should now be in Nightly 20260203092803, @OrKoN can you add the full-ci label?

OrKoN:
PR CLs run only with stable so there is no point in running CI on this PR. In the nightly jobs we run with `--ignore-unexpectedly-passing` so strictly speaking we do not need to add this PASS expectation.

whimboo:
The latest run of the daily canary CI jobs can be found at: https://github.com/puppeteer/puppeteer/actions/runs/21663802814.

We do not run these tests because of `"[waittask.spec] waittask specs protocol timeout *\",\"skip\":true}`. @OrKoN mind explaining why we do not need the `PASS` expectation for Canary?

Note that we are going to uplift this fix for 148 beta as well.

OrKoN:
For unskipping we would need the pass expectations, yes.

OrKoN:
(forgot to press approve, but I approve :) )

whimboo:
> For unskipping we would need the pass expectations, yes.

Perfect. So I was not wrong. :)

Do we actually need a new issue labeled with Firefox 148 so that we do not forget about removing the entries in both expectation files? Or how should we track that?

OrKoN:
Yeah we would need a new issue to remind us.

OrKoN:
It seems sometimes it still fails https://github.com/puppeteer/puppeteer/actions/runs/21668524751/job/62470136668#step:10:722

juliandescottes:
Interesting, do I understand correctly that the first Canary CI run with the new expectations was green https://github.com/puppeteer/puppeteer/actions/runs/21667766056 ? Meaning it would now be intermittent.

So maybe there is still some slowness (we did notice ~3-4s required to start the webdriver session) in recent builds. I wonder if slightly increasing the timeout in the test would now make it pass consistently?

juliandescottes:
(I will try to get a profile with 146, to see if the webdriver session creation was much faster)

OrKoN:
Yeah, it appears to be flaky. Maybe increasing timeouts would help.

whimboo:
Can we be sure that it is the latest 149 build or did we just pick an older build from the cache which doesn't have this fix included?

juliandescottes:
> Can we be sure that it is the latest 149 build or did we just pick an older build from the cache which doesn't have this fix included?

Would be curious to check as well, but if it passed once, I think it's guaranteed to be the new one. The old one was 100% permafailing I think?

juliandescottes:
So comparing older versions just leaves me confused: 146 already shows the 5s getDeviceStatus call we fixed on Nightly (but I think the regression started with 147?). And 145 shows a very similar profile to current nightly, with a 4s delay to start the session... 

Maybe it was flaky before?

whimboo:
Yes, the new GPU process landed in Firefox 145 as https://bugzilla.mozilla.org/show_bug.cgi?id=1974545 states. So with Firefox 144 it should be not happening.

juliandescottes:
> Yes, the new GPU process landed in Firefox 145 as https://bugzilla.mozilla.org/show_bug.cgi?id=1974545 states. So with Firefox 144 it should be not happening.

In this case, the relevant bug was rather https://bugzilla.mozilla.org/show_bug.cgi?id=1993123, so 146 ? Anyway, what surprises me is that the regression was spotted in 147 only and so far I don't think we found a 147 specific regression.

juliandescottes:
Even going back to Firefox 144, I can't see a significant difference in the WebDriver session startup compared to latest Nightly. So either :
- there is something that is not tested by my repository to exercize actions
- or something changed in the gh actions setup at the same time
- or the test was already intermittent

whimboo:
As discussed with Julian a new profile would be good to have once it has settled a bit. Maybe some other sandbox related permission is still causing the remaining delay.

PR #14657, opened 2026-02-05
feat: allow configuring the output of @puppeteer/browsers install

### LINKED ISSUE #10160: [Feature]: allow configuring the output of @puppeteer/browsers install
### Feature description

Sometimes one only needs the browser spec or the executable path so there should be a flag to configure that. cc @mathiasbynens

Lightning00Blade:
Added a new `--format` flag to the `@puppeteer/browsers install` command.
This allows users to customize the output format, for example, to print only the executable path.
Supported placeholders: `{{browser}}`, `{{buildId}}`, `{{path}}`, `{{platform}}`.

Fixes #10160

---
*PR created automatically by Jules for task [17751261393760744999](https://jules.google.com/task/17751261393760744999) started by @Lightning00Blade*

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -320,7 +320,7 @@ export class CLI {
             })
             .option('format', {
               type: 'string',
-              desc: 'Format to use for the output',
+              desc: 'Format to use for the output. Supported placeholders: ${browser}, ${buildId}, ${path}, ${platform}',
             });
         },
         async args => {
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -34,6 +34,7 @@ interface InstallArgs {
   platform?: BrowserPlatform;
   baseUrl?: string;
   installDeps?: boolean;
+  format?: string;
 }
 
 function isValidBrowser(browser: unknown): browser is Browser {
@@ -316,6 +317,10 @@ export class CLI {
               type: 'boolean',
               desc: 'Whether to attempt installing system dependencies (only supported on Linux, requires root privileges).',
               default: false,
+            })
+            .option('format', {
+              type: 'string',
+              desc: 'Format to use for the output',
             });
         },
         async args => {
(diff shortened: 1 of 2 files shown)


packages/browsers/src/CLI.ts
```diff
@@ -316,6 +317,10 @@ export class CLI {
               type: 'boolean',
               desc: 'Whether to attempt installing system dependencies (only supported on Linux, requires root privileges).',
               default: false,
+            })
+            .option('format', {
+              type: 'string',
+              desc: 'Format to use for the output',
```
Lightning00Blade: List all the placeholders here.

packages/browsers/test/src/CLI.spec.ts
```diff
@@ -92,4 +92,37 @@ describe('CLI', function () {
...
+
+    try {
+      await new CLI(tmpDir).run([
+        'npx',
+        '@puppeteer/browsers',
+        'install',
+        `chrome@${testChromeBuildId}`,
+        `--path=${tmpDir}`,
+        `--base-url=${getServerUrl()}`,
+        '--format',
```
Lightning00Blade: Use the following syntax `--format=${browser} ${buildId} ${path}`

packages/browsers/test/src/CLI.spec.ts
```diff
@@ -92,4 +92,36 @@ describe('CLI', function () {
...
+        `chrome@${testChromeBuildId}`,
+        `--path=${tmpDir}`,
+        `--base-url=${getServerUrl()}`,
+        '--format=${browser} ${buildId} ${path}',
+      ]);
+    } finally {
+      console.log = originalLog;
+    }
+
+    const expectedOutputRegex = new RegExp(
```
Lightning00Blade: We should compare this not with a regex but with a string directly.

Lightning00Blade:
Please run `npm run format` after each change to make sure the code is respecting the repo's style guide.

packages/browsers/test/src/CLI.spec.ts
```diff
@@ -92,4 +92,44 @@ describe('CLI', function () {
...
+
+    try {
+      await new CLI(tmpDir).run([
+        'npx',
+        '@puppeteer/browsers',
+        'install',
+        `chrome@${testChromeBuildId}`,
+        `--path=${tmpDir}`,
+        `--base-url=${getServerUrl()}`,
+        '--format=${browser} ${buildId} ${path}',
```
OrKoN: would not it be conflicting with some shell commands?

packages/browsers/src/CLI.ts
```diff
@@ -316,6 +317,10 @@ export class CLI {
               type: 'boolean',
               desc: 'Whether to attempt installing system dependencies (only supported on Linux, requires root privileges).',
               default: false,
+            })
+            .option('format', {
+              type: 'string',
+              desc: 'Format to use for the output. Supported placeholders: ${browser}, ${buildId}, ${path}, ${platform}',
```
OrKoN: should we allow accepting a handlebars template? I think we could include it as a dependency in browsers

Lightning00Blade: I don't think we need to, as we don't have the complexity so simple RegEx replace should be fine. But using the syntax may be better. I will check what is most common.

OrKoN: in this case, I think let's align syntax either with handlebars or go templates. Unless there is some idea for a more compact syntax like `%b%bi%p%pl` similar to printfs

packages/browsers/test/src/CLI.spec.ts
Lightning00Blade: What do you mean? It's our own command a long as you are using the correct forwarding there should be no issue.
(I am talking when using `npx <package> -- --flags`)

OrKoN: have you tried running this in the zsh or bash?  `${path}` is the path env variable syntax in shell scripts.

Lightning00Blade: You were correct shell expansion happens.

packages/browsers/src/CLI.ts
Lightning00Blade: Aligned with handlebars.

OrKoN (approved):

PR #14662, opened 2026-02-06
fix: improve ConsoleMessage text() results

OrKoN:
Previously, we included `JSHandle@type` into the text content of the message. This is redundant because it duplicates the args. Instead, this PR improves the text() strings to include the type + class name and include the first line of the Error object texts.


packages/puppeteer-core/src/cdp/utils.ts
```diff
@@ -139,7 +140,36 @@ export function createClientError(
...
+    return valueFromPrimitiveRemoteObject(remoteObject);
+  }
+}
+
+/**
+ * @internal
+ */
+export function valueFromRemoteObjectReference(handle: JSHandle): string {
+  const remoteObject = handle.remoteObject();
+  assert(remoteObject.objectId, 'Cannot extract value when objectId is given');
```
szuend: nit: `... when no objectID is given`.

szuend (approved):

Lightning00Blade (approved):

PR #14670, opened 2026-02-08
fix: improve/fix locator callback example

### LINKED ISSUE #14669: [Docs/Bug]: locator mutation observer example doesn't work
See https://github.com/puppeteer/puppeteer/blob/b163ce4593a8f014b86d67d53825fbeb679045ca/docs/guides/page-interactions.md?plain=1#L119-L142 for context.

This code has a few issues:
1. The `.target` approach doesn't quite work, because I think a `<canvas>` would have to have to have a `<canvas>` child for the condition to be true.
2. The promise callback defines `res` but `resolve` is called later.
3. The mutation observer is never cleaned up, leaking memory.

This would fix these issues:

```js
await page
  .locator(() => {
      const observer = new MutationObserver(records => {
        for (const record of records) {
          for (const node of record.addedNodes) {
            if (node instanceof HTMLCanvasElement) {
              observer.disconnect();
              resolve(node);
              return;
            }
          }
        }
      });
      observer.observe(document, { childList: true, subtree: true });
    });
  })
  .wait();
```

Taking a step back, I believe this is a poor example choice because it's much more easily expressed as simply

```js
await page.locator("canvas").wait();
```

It's almost never necessary to manually add a mutation observer in Puppeteer, since `waitForFunction` and `locator` do exactly that as higher-level abstractions, so this sends readers down the wrong path and obscures Puppeteer's elegance.

I think a better motivation might be waiting for at least 3 paragraphs to exist, and extracting their text:

```js
const paragraphs = await page
  .locator(() => {
    const paragraphs = document.querySelectorAll('p');

    if (paragraphs.length >= 3) {
      return [...paragraphs].map(p => p.textContent);
    }
  })
  .wait();
```

The example is easier to follow and I don't think there's a better way to do this.

[Here's a PR](https://github.com/puppeteer/puppeteer/pull/14670) for this, but at the very least, it'd be good to fix the original, even if the suggestion above doesn't work out.

ggorlen:

Docs fix. See https://github.com/puppeteer/puppeteer/issues/14669 for explanation/motivation.

I ran the following code to ensure this works:

```js
const puppeteer = require("puppeteer"); // ^24.36.1

const html = `<!doctype html>
<html>
<body>
<div id="root"></div>
<script>
setTimeout(() => {
  const p1 = document.createElement('p');
  p1.textContent = 'first';
  document.getElementById('root').appendChild(p1);
}, 500);

setTimeout(() => {
  const p2 = document.createElement('p');
  p2.textContent = 'second';
  document.getElementById('root').appendChild(p2);
}, 1000);

setTimeout(() => {
  const p3 = document.createElement('p');
  p3.textContent = 'third';
  document.getElementById('root').appendChild(p3);
}, 2000);
</script>
</body>
</html>`;

let browser;
(async () => {
    executablePath: chromePath,
    headless: false,
  });
  await page.setContent(html);

  const paragraphs = await page
    .locator(() => {
      const paragraphs = document.querySelectorAll('p');

      if (paragraphs.length >= 3) {
        return [...paragraphs].map(p => p.textContent);
      }
    })
    .wait();

  console.log(paragraphs); // => [ 'first', 'second', 'third' ]
})()
  .catch((err) => console.error(err))
```

Yes.

**Summary**

Closes https://github.com/puppeteer/puppeteer/issues/14669, which provides the explanation/motivation for this.

No

**Other information**

N/A

### DIFF
--- a/docs/guides/page-interactions.md
+++ b/docs/guides/page-interactions.md
@@ -117,6 +117,7 @@ The locator automatically checks the following before returning:
   [visible](https://pptr.dev/api/puppeteer.elementhandle.isvisible) or hidden.
 
 ### Waiting for a function
+
 Sometimes it is useful to wait for an arbitrary condition expressed as a
 JavaScript function. In this case, locator can be defined using a function
 instead of a selector. The following example waits until at least 3 paragraphs
--- a/docs/guides/page-interactions.md
+++ b/docs/guides/page-interactions.md
@@ -117,26 +117,20 @@ The locator automatically checks the following before returning:
   [visible](https://pptr.dev/api/puppeteer.elementhandle.isvisible) or hidden.
 
 ### Waiting for a function
-
 Sometimes it is useful to wait for an arbitrary condition expressed as a
 JavaScript function. In this case, locator can be defined using a function
-instead of a selector. The following example waits until the MutationObserver
-detects a `HTMLCanvasElement` element appearing on the page. You can also call
-other locator functions such as `.click()` or `.fill()` on the function locator.
+instead of a selector. The following example waits until at least 3 paragraphs
+are present on the page, then extracts their text. You can also call locator
+functions such as `.click()` or `.fill()` instead of mapping elements to text.
 
 ```ts
-await page
+const paragraphs = await page
   .locator(() => {
-    return new Promise(res => {
-      const observer = new MutationObserver(records => {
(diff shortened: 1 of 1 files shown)


OrKoN (approved):

OrKoN:
Thanks!

Issue #11893, opened 2024-02-12
[Bug]: Locator click outputs Timeout, when using a variable instead of a fixed


mikepziegler:
### Minimal, reproducible example

```TypeScript
import puppeteer from 'puppeteer';

(async () => {

    const browser = await puppeteer.launch();
    const page = await browser.newPage();

    const searchURL = 'someUrlWith'
    const dynamicValue = 'searchingForValue'


    await page.goto(searchURL, { waitUntil: 'networkidle0' });

    await page.screenshot({ path: 'debug-screenshot-1.png', fullPage: true });

    await page.locator(' .searchedClass')
        .filter((div: { innerHTML: string | any[]; }) => div.innerHTML.includes(dynamicValue))
        .click();

    await page.waitForNavigation({waitUntil: "domcontentloaded"});

    await page.screenshot({ path: 'debug-screenshot-2.png', fullPage: true });
})
```

### Error string

TimeoutError: Timed out after waiting 30000ms

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

I want to use filter in order to get a specific div with the class 'searchedClass', that has dynamicValue as a text within the html of the div. I also want to click on it in order to land to the next page

### Expectation

I would get to the next page and get the current url after the click

### Reality

I got this TimeoutError

```
TimeoutError: Timed out after waiting 30000ms
at path-to-project-root\node_modules\puppeteer-core\src\common\util.ts:311:17
at path-to-project-root\node_modules\puppeteer-core\lib\cjs\third_party\rxjs\rxjs.js:3:59978
at t.f._next (path-to-project-root\node_modules\puppeteer-core\lib\cjs\third_party\rxjs\rxjs.js:3:11821)
at t.next (path-to-project-root\node_modules\puppeteer-core\lib\cjs\third_party\rxjs\rxjs.js:3:6459)
at t.<anonymous> (path-to-project-root\node_modules\puppeteer-core\lib\cjs\third_party\rxjs\rxjs.js:3:71381)
at t._execute (path-to-project-root\node_modules\puppeteer-core\lib\cjs\third_party\rxjs\rxjs.js:3:28234)
at t.execute (path-to-project-root\node_modules\puppeteer-core\lib\cjs\third_party\rxjs\rxjs.js:3:28057)
at t.flush (path-to-project-root\node_modules\puppeteer-core\lib\cjs\third_party\rxjs\rxjs.js:3:32540)
[...]

PR #14671, opened 2026-02-08
docs: clarify how to pass data into filter

ggorlen:

Docs clarification. See discussion in https://github.com/puppeteer/puppeteer/issues/11893.

I tested the syntax to make sure it works:

```js
const puppeteer = require("puppeteer"); // ^24.36.1

const html = `<button onclick="f()">My button</button>
<script>function f() {
  const b = document.querySelector("button");
  b.textContent = "clicked!";
}</script>`;

let browser;
(async () => {
  await page.setContent(html);
  const buttonName = 'My button';
  await page
    .locator('button')
    .filter(`button => button.textContent === ${JSON.stringify(buttonName)}`)
    .click();
  console.log(await page.$eval("button", el => el.textContent)); // => clicked!
})()
  .catch(err => console.error(err))
```

Yes.

**Summary**

See [this comment](https://github.com/puppeteer/puppeteer/issues/11893#issuecomment-3866937217) for motivation.

In short, it's currently not readily possible to pass dynamic variables into the locator `.filter()` function via variable args, so this provides an option via string concatenation.

Granted, passing strings in like this is perhaps not the _best_ approach, so long-term, it might be worth considering adding variable args to pass data into the function.

If we don't want to encourage the string concatenation, this PR can be modified to at least mention that values from the outer scope aren't accessible to [save folks time](https://stackoverflow.com/questions/79882866/locator-filter-works-with-hardcoded-value-but-not-variable#comment140973672_79882931).

No.

**Other information**

This also makes a minor tweak to [prefer using `textContent` over `innerText`](https://stackoverflow.com/questions/35213147/difference-between-textcontent-vs-innertext). I'm fine skipping that part too if it's unwanted!

### DIFF
--- a/docs/guides/page-interactions.md
+++ b/docs/guides/page-interactions.md
@@ -145,12 +145,22 @@ await page
 
 The following example shows how to add extra conditions to the locator expressed
 as a JavaScript function. The button element will only be clicked if its
-`innerText` is 'My button'.
+`textContent` is 'My button'.
 
 ```ts
 await page
   .locator('button')
-  .filter(button => button.innerText === 'My button')
+  .filter(button => button.textContent === 'My button')
+  .click();
+```
+
+Since `.filter()`'s callback is executed in browser context, it doesn't have access to variables from the Node scope. You can build a string function to inject a variable:
+
+```ts
+const buttonName = 'My button';
+await page
+  .locator('button')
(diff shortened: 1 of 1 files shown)


OrKoN (approved):
Thanks!

PR #14676, opened 2026-02-11
fix(webdriver): avoid json serialization

OrKoN:
Proper solution is to make Puppeteer's Transport generic (with string being the default) using object only for BiDi transports.

### DIFF
--- a/test/src/console.spec.ts
+++ b/test/src/console.spec.ts
@@ -12,6 +12,53 @@ import {html, waitEvent} from './utils.js';
 
 describe('console', function () {
   setupTestBrowserHooks();
+  it.only('measure performance', async () => {
+    const {page} = await getTestState();
+    await page.setContent(html`<div id="some-counter"></div>`);
+
+    // Measurement
+    const ITERATIONS_PER_RUN = 1000;
+    const latencies: number[] = [];
+    for (let i = 0; i < ITERATIONS_PER_RUN; i++) {
+      const start = performance.now();
+      await page.evaluate(
+        (index, id) => {
+          // Change DOM.
+          document.getElementById(id)!.innerText = `Iter: ${index + 1}`;
+          // Return an object.
+          return {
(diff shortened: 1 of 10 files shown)


OrKoN:
@sadym-chromium here is released puppeteer profiles with bidi and without bidi https://gist.github.com/OrKoN/bcfd050b0703dacd0ac3a80944923951 it looks like bidi spends 12% in JS, CDP only 4%

OrKoN:
Why CDP shows more unaccounted (95.3% vs 88%)
- With CDP, the path from Node to the browser is simpler and has fewer JS layers. When Node sends a CDP command and waits for the response, it’s mostly idle in the event loop → sampled as UNKNOWN.
- With Bidi, there are more JS layers (chromium-bidi Realm, Deserializer, BidiOverCdp, ProcessingQueue, mitt/EventEmitter). When the sampler hits, it’s more often inside that JS → more ticks accounted to JavaScript, so unaccounted is lower (88%) even though you’re still waiting on the browser.

OrKoN:
Measured total time waiting for the browser (in both cases only the underlying cdp commands):

- 7662 ms (bidi)
- 4505 ms (cdp)

test/src/console.spec.ts
```diff
@@ -12,6 +12,53 @@ import {html, waitEvent} from './utils.js';
...
+    const ITERATIONS_PER_RUN = 1000;
+    const latencies: number[] = [];
+    for (let i = 0; i < ITERATIONS_PER_RUN; i++) {
+      const start = performance.now();
+      await page.evaluate(
+        (index, id) => {
+          // Change DOM.
+          document.getElementById(id)!.innerText = `Iter: ${index + 1}`;
+          // Return an object.
+          return {
```
OrKoN, line 29: removing return makes the results pretty similar to seems, so perhaps it is serialization. cc @sadym-chromium

PR #14681, opened 2026-02-13
fix(wsl): check for linux locations first

OrKoN:
Puppeteer cannot launch Windows Chrome with pipes or with sockets without mirrored networking so it does not make sense to pick a Windows host location first.


Lightning00Blade (approved):

Lightning00Blade:
In the long run we should have a way to change this behaviour.

PR #14698, opened 2026-02-17
test: fix of Windows and MacOS

Lightning00Blade:


packages/browsers/test/src/CLI.spec.ts
```diff
@@ -108,25 +108,20 @@ describe('CLI', function () {
         `chrome@${testChromeBuildId}`,
         `--path=${tmpDir}`,
         `--base-url=${getServerUrl()}`,
-        '--format={{browser}} {{buildId}} {{path}}',
+        '--format=@{{buildId}}@{{browser}}@',
```
OrKoN: can we do `--format='{{browser}} {{buildId}} {{path}}'`?

OrKoN (approved):

PR #14702, opened 2026-02-18
ci: use lower case as every other label

Lightning00Blade:


.github/workflows/stale.yml
```diff
@@ -32,6 +32,7 @@ jobs:
             new issue and include a minimal reproducible example or
             other details that would allow us to reproduce it.
           stale-issue-label: 'unconfirmed'
-          exempt-issue-labels: 'confirmed,feature'
```
OrKoN, line 35: why do not we exempt features? I do not think we need stale bot for features

Lightning00Blade: They are not. This is the label `feature, the type is ignored by default.

OrKoN (approved):

.github/workflows/stale.yml
OrKoN: It does not seem to be ignored https://github.com/puppeteer/puppeteer/issues/14694#issuecomment-3988886199

PR #14706, opened 2026-02-21
feat: Add overwrite flag to type method

### LINKED ISSUE #14114: [Feature]: Overwrite and type
### Feature description

It would be useful having an `overwrite` flag in `.type()` that would clear the input before typing:
```
// Focus the editor
await this.page.focus(selector);

// Select all content (Cmd+A on Mac, Ctrl+A on Windows/Linux)
await this.page.keyboard.down(process.platform === 'darwin' ? 'Meta' : 'Control');
await this.page.keyboard.press('A');
await this.page.keyboard.up(process.platform === 'darwin' ? 'Meta' : 'Control');

// Delete selected content
await this.page.keyboard.press('Backspace');

// Type
...
```

eltaweel068:
Feature

Yes. Added a test in `elementhandle.spec.ts` to verify that the input is cleared before typing when `overwrite: true` is passed.

Yes, added inline JSDoc for the new `overwrite` property in `KeyboardTypeOptions` and updated the `.type()` method documentation.

**Summary**
Fixes #14114

This PR introduces the `overwrite` flag to the `KeyboardTypeOptions` for the `.type()` method. 
When `overwrite: true` is passed, the method clears the existing input value before typing the new text.

No. The new `overwrite` flag is optional.

**Other information**
No

### DIFF
--- a/packages/puppeteer-core/src/api/ElementHandle.ts
+++ b/packages/puppeteer-core/src/api/ElementHandle.ts
@@ -1124,15 +1124,29 @@ export abstract class ElementHandle<
    * await elementHandle.press('Enter');
    * ```
    *
-   * @param options - Delay in milliseconds. Defaults to 0.
+   * @param options - An object of options. Accepts `delay` in milliseconds Defaults to 0, and `overwrite` which clears the existing text before typing.
    */
-  @throwIfDisposed()
+@throwIfDisposed()
   @bindIsolatedHandle
   async type(
     text: string,
     options?: Readonly<KeyboardTypeOptions>,
   ): Promise<void> {
     await this.focus();
+
+    if (options?.overwrite) {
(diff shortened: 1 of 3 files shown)


OrKoN:
I think we might not actually want to have this built-in as it might not work for all inputs. cc @Lightning00Blade

OrKoN:
@Lightning00Blade did you have a chance to review?

Lightning00Blade:
Thanks for the PR.
I looked at some things and two thing that are issue that I would need to be solved for this feature to be: 

- Currently the OS check only works for the NodeJS build and not in the browser
- On theory the user can change the behaviour of the shortcut, or the website can intercept them before they do what they are supposed to.

Give that I don't think we can for sure say that this was overridden so it would be better for the user to implement them on their own.

PR #14716, opened 2026-02-23
feat(browsers): add unpacking indicator during install

### LINKED ISSUE #13083: [Feature]: Add unpacking indicator to the install command

githubbzxs:
Adds a small user-visible indicator after download completes and before archive extraction begins, so it’s clearer that the install command is still doing work (unpacking).

Fixes #13083


OrKoN:
Thanks for the PR but we do not want to have unconditionally logging in the library code. Let's instead inject the progress reported similar to the download progress reporter.

PR #14723, opened 2026-02-25
test: fix emulation emulateNetworkConditions test to expect the correct error on Firefox

juliandescottes:
ERR_INTERNET_DISCONNECTED is only valid on Chrome.

I am adding support for emulation.setNetworkConditions on Firefox at the moment and would like to update this test so that it can pass when the feature lands.


juliandescottes:
cc @OrKoN

OrKoN (approved):

PR #14724, opened 2026-02-25
fix: consider browsingContext.navigationCommitted to dispose an existing navigation

### LINKED ISSUE #14722: Second goto hangs after doing a first navigation to about:neterror
### Feature description

Consider the following test case which forces to navigate to about:neterror (on firefox) and then immediately navigates to another (valid) URL

```ts
it.only('should navigate successfully after encountering network error', async () => {
  let error!: Error;
  await page.goto('https://foo/').catch(error_ => {
    return (error = error_);
  });
  expect(error).not.toBe(null);
  const response = await page.goto(server.PREFIX + '/grid.html');
  expect(response!.status()).toBe(200);
});
```

The second goto will timeout at the moment.

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=2019202 on the Firefox side, because I think most of the issue is really that Firefox emits inconsistent events for a navigation to about:neterror. TLDR, we emit `domContentLoaded`, but not `load` nor `navigationFailed`, but we still resolve the navigate command as soon as we receive `navigationCommitted`.

However the resulting failure on the puppeteer side is a bit unexpected, because in the test above, it's the second `goto` call which will hang, presumably because the first navigation is still considered as pending? 

I checked that adding `navigationCommitted` to https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/bidi/core/Navigation.ts#L99-L102 fixes that issue, but I'm not sure it makes sense for the navigation logic of Puppeteer?

Again, there's definitely a root issue in Firefox, but I was wondering if that could be an opportunity to make Puppeteer's navigation logic more robust.

juliandescottes:
Fixes #14722 

(waiting for CI results to work on a test, let's see if this regresses other scenarios)

### DIFF
--- a/packages/puppeteer-core/src/bidi/core/Navigation.ts
+++ b/packages/puppeteer-core/src/bidi/core/Navigation.ts
@@ -99,6 +99,7 @@ export class Navigation extends EventEmitter<{
     for (const eventName of [
       'browsingContext.domContentLoaded',
       'browsingContext.load',
+      'browsingContext.navigationCommitted',
     ] as const) {
       sessionEmitter.on(eventName, info => {
         if (
--- a/test/src/navigation.spec.ts
+++ b/test/src/navigation.spec.ts
@@ -112,6 +112,23 @@ describe('navigation', function () {
       });
       expect(response!.status()).toBe(200);
     });
+    it('should navigate successfully after encountering network error', async () => {
+      const {page, server} = await getTestState();
+
+      // Destroy the connection to simulate a network error. This will open
+      // about:neterror on Firefox.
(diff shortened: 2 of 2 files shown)


juliandescottes:
CI seemed fine, I added a slightly modified version of the test mentioned in the issue.

OrKoN (approved):

OrKoN:
@Lightning00Blade it is seems the installation test with extensions got really bad. Shall we disable it?

juliandescottes:
Note that this should fix the timeout issue on `[page.spec] Page Page.setOfflineMode should work` now popping up on Nightly runs.

Issue #14738, opened 2026-03-02
[Bug]: "Page.setOfflineMode should work" is timing in Firefox nightly


OrKoN:
### Minimal, reproducible example

```TypeScript
See the "Page.setOfflineMode should work" test.
```

### Expectation

The test fails https://github.com/puppeteer/puppeteer/actions/runs/22565653148/job/65363694242

### Reality

The test times out https://github.com/puppeteer/puppeteer/actions/runs/22565653148/job/65363694242

### Puppeteer configuration file (if used)

```TypeScript

```

### Puppeteer version

latest

### Node version

latest

### Package manager

npm

### Package manager version

latest

### Operating system

Linux

OrKoN:
cc @whimboo @juliandescottes I saw that this test started timing out in Firefox nightly (see https://github.com/puppeteer/puppeteer/actions/runs/22565653148/job/65363694242#step:10:822). It looks like MacOS and Ubuntu, headless and headful configurations are affected.

whimboo:
Isn't this fixed by https://github.com/puppeteer/puppeteer/pull/14724?

OrKoN:
Indeed

PR #14739, opened 2026-03-02
chore: enable blank issues in issue template configuration

OrKoN:
to create custom bug reports like https://github.com/puppeteer/puppeteer/issues/14738 for flaky tests

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,4 +1,4 @@
-blank_issues_enabled: false
+blank_issues_enabled: true
 contact_links:
   - name: General Puppeteer questions
     url: https://stackoverflow.com/questions/tagged/puppeteer


Lightning00Blade:
What about we create a simple one with the type `Task` that way it's simple to redirect user to correct one.

OrKoN:
@Lightning00Blade the users would be using it to create bug reports just like in https://github.com/puppeteer/puppeteer/issues/14722 the feature was used. I would suggest we just allow empty issues as it is not very different from a template with a task type.

Lightning00Blade:
I disagree that is the same. Providing even the simplest template of textarea + type + label, allows us to set up automation, view and other ways to track issue. 
Blank issue make it harder to automate things.

OrKoN:
I do not see why you cannot set up automation for blank issues.

Lightning00Blade (approved):
LGTM, we will move this to a template eventually.

PR #14741, opened 2026-03-03
test: update tests and add diagnostics

OrKoN:


.github/workflows/ci.yml
```diff
@@ -221,6 +221,9 @@ jobs:
       - name: Run all tests (Linux)
         if: ${{ matrix.os == 'ubuntu-latest' }}
         run: xvfb-run --auto-servernum npm run test -- --shard '${{ matrix.shard }}' --test-suite ${{ matrix.suite }} --save-stats-to /tmp/artifacts/${{ github.event_name }}_INSERTID.json
+      - name: Check Kernel Logs
```
Lightning00Blade, line 224: I forgot if we need to exit with non-zero, so that the run correctly fails.
I think it would be fine to do that regardless.

Lightning00Blade (approved):
LGTM + 1 comment

.github/workflows/ci.yml
OrKoN: I think it worked as expected in test runs (jobs failed but the check ran).

Lightning00Blade: Yeah my question is if because this will succeeded, will the whole run succeed as well.
Kinda ignoring the failed test part.

PR #14743, opened 2026-03-03
chore: draft/idea implementation of extension api

nroscino:

**Summary**

**Other information**

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -31331,7 +31331,7 @@
         "@puppeteer/browsers": "2.13.0",
         "chromium-bidi": "14.0.0",
         "cosmiconfig": "^9.0.0",
-        "devtools-protocol": "0.0.1566079",
+        "devtools-protocol": "0.0.1581282",
         "puppeteer-core": "24.37.5",
         "typed-query-selector": "^2.12.1"
       },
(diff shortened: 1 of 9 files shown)


packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -29,11 +29,144 @@ import {
...
+import type {WebWorker} from './WebWorker.js';
+
+/**
+ * @public
+ */
+export class Extension {
+  #browser: Browser;
+  #pages = new WeakMap<Target, Page>();
+  #pageLogs = new WeakMap<Page, string[]>();
+  #workerLogs = new WeakMap<WebWorker, string[]>();
```
OrKoN: we should not store logs but emit them as events similar to how we do for pages. WebWorkers already have this so we do not need to use CDP directly.

packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -29,11 +29,144 @@ import {
...
 import type {Page} from './Page.js';
 import type {Target} from './Target.js';
+import type {WebWorker} from './WebWorker.js';
+
+/**
+ * @public
+ */
+export class Extension {
+  #browser: Browser;
+  #pages = new WeakMap<Target, Page>();
```
OrKoN: we do not need to save pages let's discover them dynamically in the getter.

PR #14745, opened 2026-03-04
fix: remove RenderDocument from disabled Chrome features

zyzyzyryxy:


packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -200,6 +199,7 @@ export class ChromeLauncher extends BrowserLauncher {
     // Merge default enabled features with user-provided ones, if any.
     const enabledFeatures = [
       'PdfOopif',
+      'RenderDocument:level/all-frames', // https://crbug.com/444150315
```
OrKoN: we do not need to explicitly enable it, let just use whatever is enabled in the Chrome by default?

zyzyzyryxy: Done

OrKoN (approved):

PR #14753, opened 2026-03-06
feat: add locale option to LaunchOptions

### LINKED ISSUE #11303: [Feature]: Support locale when starting the puppeteer
### Feature description

Could we pass the locale arg when starting the puppeteer, that would be great! Playwright supports that https://playwright.dev/docs/api/class-browser#browser-new-context-option-locale

szy1840:
Adds a `locale` option to `LaunchOptions` to make it easy to set the browser language at launch, similar to Playwright's API.

- Chrome: passes `--lang=<locale>` as a command-line flag
- Firefox: sets the `intl.locale.requested` profile preference

If `--lang` is already present in `args`, it takes precedence over `locale` for Chrome.

Fixes #11303

### DIFF
--- a/packages/puppeteer-core/src/node/FirefoxLauncher.test.ts
+++ b/packages/puppeteer-core/src/node/FirefoxLauncher.test.ts
@@ -20,5 +20,21 @@ describe('FirefoxLauncher', function () {
       expect(prefs['fission.bfcacheInParent']).toBe(undefined);
       expect(prefs['fission.webContentIsolationStrategy']).toBe(0);
     });
+
+    it('should pass through locale preference', () => {
+      const prefs: Record<string, unknown> = FirefoxLauncher.getPreferences({
+        'intl.locale.requested': 'fr-FR',
+      });
+      expect(prefs['intl.locale.requested']).toBe('fr-FR');
+    });
+
+    it('should allow extraPrefsFirefox to override locale preference', () => {
+      const prefs: Record<string, unknown> = FirefoxLauncher.getPreferences({
+        'intl.locale.requested': 'en-US',
+        'intl.locale.requested2': 'de-DE',
+      });
+      expect(prefs['intl.locale.requested']).toBe('en-US');
+      expect(prefs['intl.locale.requested2']).toBe('de-DE');
+    });
   });
 });
--- a/test/src/launcher.spec.ts
+++ b/test/src/launcher.spec.ts
@@ -439,6 +439,34 @@ describe('Launcher specs', function () {
           );
         }
       });
+      it('should include --lang flag when locale is set (Chrome)', async () => {
+        const {isChrome, puppeteer} = await getTestState({skipLaunch: true});
+        if (!isChrome) {
+          return;
+        }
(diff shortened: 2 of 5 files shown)


test/src/launcher.spec.ts
```diff
@@ -439,6 +439,34 @@ describe('Launcher specs', function () {
           );
         }
       });
+      it('should include --lang flag when locale is set (Chrome)', async () => {
```
Lightning00Blade, line 442: Can we make this an unit test for `packages/puppeteer-core/src/node/ChromeLauncher.ts`?
That way we won't need the `isChrome` check which we try to avoid in e2e test.

packages/puppeteer-core/src/node/FirefoxLauncher.test.ts
```diff
@@ -20,5 +20,21 @@ describe('FirefoxLauncher', function () {
       expect(prefs['fission.bfcacheInParent']).toBe(undefined);
       expect(prefs['fission.webContentIsolationStrategy']).toBe(0);
     });
+
+    it('should pass through locale preference', () => {
```
Lightning00Blade, line 24: This does not seem to be a good test as we only set defaults in the `getPreferences`

I think we should call `computeLaunchArguments` with the correct options (and a tmp folder for the user dir). 
And set up a spy on the `getPreferences` to check that the local is setup correctly.

Lightning00Blade (commented):
Over all looks good, there some comments

I would like to double check there is no better alternative before approving and mergin.

CC @whimboo the Preferences looks like the only way to do it for Firefox, correct?

whimboo:
> CC @whimboo the Preferences looks like the only way to do it for Firefox, correct?

This should only affect content, not the browser’s locale, correct? If so, this really seems like something we should handle as a global override via [`emulation.setLocaleOverride`](https://w3c.github.io/webdriver-bidi/#command-emulation-setLocaleOverride). That would consistently impact both `navigator.language` and the `Accept-Language` header.

Qodo-Free-For-OSS:
Hi, `FirefoxLauncher` builds preferences with `...(options.locale ? {...} : undefined)`, which is not an object type in the false branch and can fail TypeScript compilation under strict settings. This can break the puppeteer-core build.

**How to fix:** Use empty object fallback

> ### Issue description
> In FirefoxLauncher, preferences are created via an object literal containing a conditional spread that can evaluate to `undefined`:
> `...(options.locale ? {...} : undefined)`.
> Under strict TypeScript settings, this pattern commonly fails because `undefined` is not an object type.
> 
> ### Issue Context
> The repo uses `strict: true` and builds puppeteer-core with `tsc -b`.
> 
> ### Fix Focus Areas
> - packages/puppeteer-core/src/node/FirefoxLauncher.ts[108-116]
> - tsconfig.base.json[1-37]
> 
> ### Suggested approach
> - Replace `: undefined` with `: {}` so both branches are object types.
>   Example:
>   ```ts
>   ...(options.locale ? {'intl.locale.requested': options.locale} : {}),
>   ```
> - Alternatively, build a `const localePrefs = options.locale ? {...} : {};` and spread that.

---

OrKoN:
Closing as stale

PR #14756, opened 2026-03-09
feat: support PUPPETEER_DANGEROUS_NO_SANDBOX environment variable

### LINKED ISSUE #5505: Feature request: Implement environment variable for `--no-sandbox`
### Steps to reproduce

* Puppeteer version: all
* Platform / OS version: Docker
* Node.js version: All

Running puppeteer as a root user errors with the common: `Running as root without --no-sandbox is not supported. See https://crbug.com/638180.`. You can get around this by adding a `--no-sandbox` to every invocation of puppeteer, but this is not easily feasible or desirable when there are thousands of tests present.

For our specific use case we are utilizing puppeteer to run integration tests, and chowning all files for suitable permissions takes way too long. We are looking at speeding this up by moving to a no-sandbox environment.

**One possible solution**

Please consider adding support for an environment variable, such as `PUPPETEER_DANGEROUS_NO_SANDBOX=true`, whenever this environment variable is set, run all tests as no-sandbox.

NasehFaraj:
### Description
This PR introduces the `PUPPETEER_DANGEROUS_NO_SANDBOX` environment variable. When set, it automatically appends the `--no-sandbox` flag to the browser's launch default arguments, provided it hasn't already been passed manually by the user in the `args` array.

This is particularly useful for Docker and CI environments where modifying the application code to pass `--no-sandbox` to every `puppeteer.launch()` invocation is not feasible or desirable.

Fixes #5505

### DIFF
--- a/packages/puppeteer-core/src/node/ChromeLauncher.ts
+++ b/packages/puppeteer-core/src/node/ChromeLauncher.ts
@@ -245,6 +245,14 @@ export class ChromeLauncher extends BrowserLauncher {
       userDataDir,
       enableExtensions = false,
     } = options;
+
+    if (
+      process.env['PUPPETEER_DANGEROUS_NO_SANDBOX'] &&
+      !args.includes('--no-sandbox')
+    ) {
+      chromeArguments.push('--no-sandbox');
+    }
+
     if (userDataDir) {
       // If absolute (for any platform) path is given, we should not resolve it.
       chromeArguments.push(
--- a/packages/puppeteer-core/src/node/ChromeLauncher.ts
+++ b/packages/puppeteer-core/src/node/ChromeLauncher.ts
@@ -247,7 +247,7 @@ export class ChromeLauncher extends BrowserLauncher {
     } = options;
 
     if (
-      process.env['PUPPETEER_DANGEROUS_NO_SANDBOX'] &&
+      process.env['PUPPETEER_DANGEROUS_NO_SANDBOX'] === 'true' &&
       !args.includes('--no-sandbox')
     ) {
       chromeArguments.push('--no-sandbox');


packages/puppeteer-core/src/node/ChromeLauncher.ts
```diff
@@ -245,6 +245,14 @@ export class ChromeLauncher extends BrowserLauncher {
       userDataDir,
       enableExtensions = false,
     } = options;
+
+    if (
+      process.env['PUPPETEER_DANGEROUS_NO_SANDBOX'] &&
```
zyzyzyryxy: Let's follow the example of `PUPPETEER_TEST_EXPERIMENTAL_CHROME_FEATURES` for which we check if it is set to 'true' specifically. Otherwise having `PUPPETEER_DANGEROUS_NO_SANDBOX` set to 'false' in the environment could lead to very unexpected results!

zyzyzyryxy (changes requested):
Looks good, just one change requested

packages/puppeteer-core/src/node/ChromeLauncher.ts
NasehFaraj: Thanks for the review. I've updated the condition to explicitly check for `'true'` to match the behavior of `PUPPETEER_TEST_EXPERIMENTAL_CHROME_FEATURES`. The fix has been pushed.

NasehFaraj:
@zyzyzyryxy Hi Piotr, just a gentle ping on this PR. The requested changes were addressed last week and all CI checks are green. Let me know if there's anything else needed before merging!

zyzyzyryxy (approved):

PR #14761, opened 2026-03-10
fix: flush renderer before screenshot to prevent protocol timeouts

### LINKED ISSUE #14760: [Bug]: Protocol timeout when screenshotting in pretty weird case (Docker + multiple tabs + networkidle2)

```TypeScript
import puppeteer from 'puppeteer-core';

  protocolTimeout: 20_000,
  executablePath: process.env.CHROME_BIN || '/usr/bin/chromium',
  args: ['--no-sandbox', '--disable-gpu'],
});

const sleep = (ms) => new Promise((r) => setTimeout(r, ms));

async function screenshotTask(id) {
  await sleep(Math.random() * 3000);
  await page.setViewport({ width: 1024, height: 768 });
  await page.goto('https://cf.nearpod.com/neareducation/new/Scrubber/en/dash-slideEditor.png', { waitUntil: 'networkidle2'  }); // Some random image on our CDN
  console.log(`[tab${id}] taking screenshot`);
  await page.screenshot();
  console.log(`[tab${id}] done`);
  await page.close();
}

await Promise.all(Array.from({ length: 5 }, (_, i) => screenshotTask(i)));
await browser.close();
```

### Background

This is a very weird combination of factors leading to this issue, posting it here more in case it helps to unveil some underlying issue than expecting it to be solved for my use case. 

We have a puppeteer service we use for taking screenshots of internal pages. It runs in K8s. 
A few months back (And I wish I could give you a better answer for what version of pptr / chrome caused the issue), we started seeing the healthchecks starting to fail a lot more than they used to. 
Our healthcheck just attempts to load a PNG loaded in our CDN and screenshot that (more as a test that the browser instance is still alive than anything else). 
While trimming it down to a repro case, I discovered that this only happens:
a. In docker (I couldn't repro it running node directly)
b. When using multiple tabs at once (ie: handling request in parallel to the healthcheck running)
c. When the thing being screenshotted is an image (Couldn't repro it by hitting a regular webpage)
d. Only when using `waitUntil: 'networkidle2'` on the goTo call. 

Created a repo with the full repro case (including dockerfile here: https://github.com/nicofff/puppeteer_repro_weird_hang) 

### Expectation

Screnshots succeed

### Reality

file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:99
    #error = new ProtocolError();
             ^

ProtocolError: Page.captureScreenshot timed out. Increase the 'protocolTimeout' setting in launch/connect calls for a higher timeout if needed.
    at <instance_members_initializer> (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:99:14)
    at new Callback (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:103:16)
    at CallbackRegistry.create (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/common/CallbackRegistry.js:22:26)
    at Connection._rawSend (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/Connection.js:104:26)
    at CdpCDPSession.send (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/CdpSession.js:71:33)
    at CdpPage._screenshot (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/cdp/Page.js:834:62)
    at CdpPage.screenshot (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/api/Page.js:1079:41)
    at async CdpPage.<anonymous> (file:///app/node_modules/puppeteer-core/lib/esm/puppeteer/util/decorators.js:164:24)
    at async screenshotTask (file:///app/index.mjs:17:3)
    at async Promise.all (index 2)

```TypeScript

```

### Puppeteer version

24.38.0

### Node version

v24.4.1

### Package manager

npm

### Package manager version

11.4.2

### Operating system

Linux

vansh007:
Problem
Taking a screenshot immediately after page.goto() with networkidle2 can cause a ProtocolError: Page.captureScreenshot timed out (#14760). This issue is most common when:

Screenshotting image documents (PNG/JPG directly).
Running in Docker/constrained environments.
Handling parallel requests across multiple tabs.
Root Cause
For image documents, networkidle2 resolves almost instantly because there is only one resource to load. However, the Chromium renderer may not have completed its first paint cycle yet. If Page.captureScreenshot is called during this window (especially with fromSurface: true in headful/Docker), the protocol command can hang indefinitely.

The Solution
This PR adds a synchronization point by performing an empty evaluation (

evaluate(() => {})
) before calling the screenshot command.

Existing Precedent: Puppeteer already performs an implicit flush when fullPage: true is used (via scroll dimension calculation). This PR brings that same reliability to the default screenshot path.
Effect: The evaluation forces the renderer to flush all pending tasks and ensures it is in a consistent state before the capture begins.
Changes
Core: Added await this.mainFrame().isolatedRealm().evaluate(() => {}); in the default 

screenshot
 path of 

packages/puppeteer-core/src/api/Page.ts
.
Test: Added a regression test in 

test/src/screenshot.spec.ts
 that captures a screenshot of an image asset specifically with networkidle2.
Verification Results
Successfully reproduced the timeout locally using 50 parallel tabs screenshotting the same CDN image. After the fix, all 50 screenshots completed successfully without a single timeout.

Fixes #14760

### DIFF
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -2720,6 +2720,10 @@ export abstract class Page extends EventEmitter<PageEvents> {
         }
       } else {
         options.captureBeyondViewport = false;
+        // If we are not capturing beyond viewport, we ensure the renderer is
+        // flushed by performing a simple evaluation. This helps to synchronize
+        // the state and avoid hangs in `Page.captureScreenshot`.
+        await this.mainFrame().isolatedRealm().evaluate(() => {});
       }
     }
 
--- a/test/src/screenshot.spec.ts
+++ b/test/src/screenshot.spec.ts
@@ -205,6 +205,15 @@ describe('Screenshots', function () {
       expect(size.width).toBe(originalSize.width);
       expect(size.height).toBe(originalSize.height);
     });
+    it('should take a screenshot of an image document', async () => {
+      const {page, server} = await getTestState();
+      await page.goto(server.PREFIX + '/pptr.png', {
+        waitUntil: 'networkidle2',
+      });
+      const screenshot = await page.screenshot();
+      expect(screenshot).toBeInstanceOf(Uint8Array);
+      expect(screenshot.length).toBeGreaterThan(0);
+    });
   });
 
   describe('ElementHandle.screenshot', function () {


Lightning00Blade (changes requested):
Thank you for PR, but I don't think this solves the issue in the correct way.
Different system may use different rendering techniques, while this may fix the use case you have it may create flakiness for others.

Note: It's recommended to await an element or something else on the page to reduce said flakiness. 

Also the test passes locally without your change so a better one is needed.

Lightning00Blade:
Closing for now, as we investigate the linked issue.

PR #14765, opened 2026-03-11
feat(browsers): show extraction progress after browser download

### LINKED ISSUE #13083: [Feature]: Add unpacking indicator to the install command

rishicodely:
Currently Puppeteer shows a download progress indicator in the terminal, but after the download finishes the unpacking process runs silently, which may make it appear as if the installation is stuck.

This change adds a CLI message indicating when extraction begins and when it completes so users can understand that installation is still progressing.

Fixes #13083

### DIFF
--- a/packages/browsers/src/install.ts
+++ b/packages/browsers/src/install.ts
@@ -448,13 +448,18 @@ async function installUrl(
     }
 
     debugInstall(`Installing ${archivePath} to ${outputPath}`);
+    process.stdout.write(
+      `Extracting ${options.browser} ${options.buildId}...\n`,
+    );
     try {
       debugTime('extract');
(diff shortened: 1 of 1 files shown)


zyzyzyryxy:
Thanks for the PR but we do not want to have unconditionally logging in the library code. Let's instead inject the progress reported similar to the download progress reporter. Please look at other PRs for this issue and comments therein.

zyzyzyryxy (changes requested):
As per previous comment: https://github.com/puppeteer/puppeteer/pull/14765#issuecomment-4040411648
It needs to be rewritten before it can be merged.

Lightning00Blade:
Closing as stale.

PR #14784, opened 2026-03-17
feat: implement console event on web workers

nroscino:

This PR introduces the possibility to listen for console events on a WebWorker via `on("console", listener);`. 
Right now WebWorker are sending console logs directly to the `page.on("console",...)` event.
This change allows to specifically listen to WebWorkers logs only.  

Yes

Yes

**Summary**

Extension service workers are treated as WebWorkers but are not connected to any page, this makes listening for console from service worker impossible without relying on the raw events on the CDP. this PR introduces a way to retrieve those logs in a simpler way.

No

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/bidi/Frame.ts
+++ b/packages/puppeteer-core/src/bidi/Frame.ts
@@ -30,11 +30,6 @@ import {
 } from '../api/Frame.js';
 import {PageEvent, type WaitTimeoutOptions} from '../api/Page.js';
 import {Accessibility} from '../cdp/Accessibility.js';
-import type {ConsoleMessageType} from '../common/ConsoleMessage.js';
-import {
-  ConsoleMessage,
-  type ConsoleMessageLocation,
-} from '../common/ConsoleMessage.js';
 import {TargetCloseError, UnsupportedOperation} from '../common/Errors.js';
 import type {TimeoutSettings} from '../common/TimeoutSettings.js';
 import type {Awaitable, HandleFor} from '../common/types.js';
@@ -50,33 +45,21 @@ import {BidiCdpSession} from './CDPSession.js';
 import type {BrowsingContext} from './core/BrowsingContext.js';
 import type {Navigation} from './core/Navigation.js';
 import type {Request} from './core/Request.js';
-import {BidiDeserializer} from './Deserializer.js';
 import {BidiDialog} from './Dialog.js';
 import {BidiElementHandle} from './ElementHandle.js';
 import {ExposableFunction} from './ExposedFunction.js';
 import {BidiHTTPRequest, requests} from './HTTPRequest.js';
 import type {BidiHTTPResponse} from './HTTPResponse.js';
-import {BidiJSHandle} from './JSHandle.js';
 import type {BidiPage} from './Page.js';
 import type {BidiRealm} from './Realm.js';
 import {BidiFrameRealm} from './Realm.js';
-import {rewriteNavigationError} from './util.js';
+import {
+  getConsoleMessage,
+  isConsoleLogEntry,
+  isJavaScriptLogEntry,
+  rewriteNavigationError,
+} from './util.js';
 import {BidiWebWorker} from './WebWorker.js';
-
-// TODO: Remove this and map CDP the correct method.
-// Requires breaking change.
-function convertConsoleMessageLevel(method: string): ConsoleMessageType {
-  switch (method) {
-    case 'group':
-      return 'startGroup';
-    case 'groupCollapsed':
-      return 'startGroupCollapsed';
-    case 'groupEnd':
-      return 'endGroup';
-    default:
-      return method as ConsoleMessageType;
-  }
-}
 
 export class BidiFrame extends Frame {
   static from(
@@ -178,30 +161,16 @@ export class BidiFrame extends Frame {
         return;
       }
       if (isConsoleLogEntry(entry)) {
+        if (!this.page().listenerCount(PageEvent.Console)) {
+          return;
+        }
         const args = entry.args.map(arg => {
           return this.mainRealm().createHandle(arg);
         });
 
-        const text = args
-          .reduce((value, arg) => {
-            const parsedValue =
-              arg instanceof BidiJSHandle && arg.isPrimitiveValue
-                ? BidiDeserializer.deserialize(arg.remoteValue())
-                : arg.toString();
-            return `${value} ${parsedValue}`;
-          }, '')
-          .slice(1);
-
         this.page().trustedEmitter.emit(
           PageEvent.Console,
-          new ConsoleMessage(
-            convertConsoleMessageLevel(entry.method),
-            text,
-            args,
-            getStackTraceLocations(entry.stackTrace),
-            this,
-            undefined,
-          ),
+          getConsoleMessage(entry, args, this),
         );
       } else if (isJavaScriptLogEntry(entry)) {
         const error = new Error(entry.text ?? '');
@@ -638,32 +607,4 @@ export class BidiFrame extends Frame {
       [element.remoteValue() as Bidi.Script.SharedReference],
     );
   }
-}
-
-function isConsoleLogEntry(
-  event: Bidi.Log.Entry,
-): event is Bidi.Log.ConsoleLogEntry {
-  return event.type === 'console';
-}
-
-function isJavaScriptLogEntry(
-  event: Bidi.Log.Entry,
-): event is Bidi.Log.JavascriptLogEntry {
-  return event.type === 'javascript';
-}
-
-function getStackTraceLocations(
-  stackTrace?: Bidi.Script.StackTrace,
-): ConsoleMessageLocation[] {
-  const stackTraceLocations: ConsoleMessageLocation[] = [];
-  if (stackTrace) {
-    for (const callFrame of stackTrace.callFrames) {
-      stackTraceLocations.push({
-        url: callFrame.url,
-        lineNumber: callFrame.lineNumber,
-        columnNumber: callFrame.columnNumber,
-      });
-    }
-  }
-  return stackTraceLocations;
 }
--- a/packages/puppeteer-core/src/bidi/Realm.ts
+++ b/packages/puppeteer-core/src/bidi/Realm.ts
@@ -7,6 +7,7 @@ import * as Bidi from 'webdriver-bidi-protocol';
 
 import type {JSHandle} from '../api/JSHandle.js';
 import {Realm} from '../api/Realm.js';
+import {WebWorkerEvent} from '../api/WebWorker.js';
 import {ARIAQueryHandler} from '../common/AriaQueryHandler.js';
 import {LazyArg} from '../common/LazyArg.js';
 import {scriptInjector} from '../common/ScriptInjector.js';
@@ -36,7 +37,12 @@ import {ExposableFunction} from './ExposedFunction.js';
 import type {BidiFrame} from './Frame.js';
 import {BidiJSHandle} from './JSHandle.js';
 import {BidiSerializer} from './Serializer.js';
-import {createEvaluationError, rewriteEvaluationError} from './util.js';
+import {
+  createEvaluationError,
+  getConsoleMessage,
+  isConsoleLogEntry,
+  rewriteEvaluationError,
+} from './util.js';
 import type {BidiWebWorker} from './WebWorker.js';
 
 /**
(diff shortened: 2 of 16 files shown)


packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -906,16 +895,13 @@ export class CdpPage extends Page {
...
+    if (!this.listenerCount(PageEvent.Console)) {
+      return;
+    }
     const values = event.args.map(arg => {
       return world.createCdpHandle(arg);
     });
 
-    if (!this.listenerCount(PageEvent.Console)) {
-      values.forEach(arg => {
-        return arg.dispose();
```
Lightning00Blade, line 915: Why don't we dispose this any more? 
CDP retains them in memory if we don't. We can add a comment here.

packages/puppeteer-core/src/bidi/core/Realm.ts
```diff
@@ -43,6 +43,8 @@ export abstract class Realm extends EventEmitter<{
   worker: DedicatedWorkerRealm;
   /** Emitted when a shared worker is created in the realm. */
   sharedworker: SharedWorkerRealm;
+  /** Emitted whenever a log entry is added to the realm. */
+  log: {entry: Bidi.Log.Entry};
```
Lightning00Blade: Let's not re-wrap it, it's internal so if we need anything we can change it later.
It's a little bit more performant and can save a tiny amount of memory.

packages/puppeteer-core/src/bidi/util.ts
```diff
@@ -6,11 +6,68 @@
...
+} from '../common/ConsoleMessage.js';
 import {ProtocolError, TimeoutError} from '../common/Errors.js';
 import {PuppeteerURL} from '../common/util.js';
 
 import {BidiDeserializer} from './Deserializer.js';
 
+/**
+ * @internal
+ */
+export function convertConsoleMessageLevel(method: string): ConsoleMessageType {
```
Lightning00Blade, line 28: Original comment is missing.

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -178,6 +165,9 @@ export class BidiFrame extends Frame {
         return;
       }
       if (isConsoleLogEntry(entry)) {
+        if (!this.page().listenerCount(PageEvent.Console)) {
```
Lightning00Blade, line 164: This is OK for BiDi as it does not keep args as alive handles.

test/src/worker.spec.ts
```diff
@@ -179,4 +180,201 @@ describe('Workers', function () {
...
+      worker.on(WebWorkerEvent.Console, msg => {
+        return messages.push(msg);
+      });
+      // All console events will be reported before `worker.evaluate` is finished.
+      await worker.evaluate(() => {
+        console.trace('calling console.trace');
+        console.dir('calling console.dir');
+        console.warn('calling console.warn');
+        console.error('calling console.error');
+        console.log(Promise.resolve('should not wait until resolved!'));
```
Lightning00Blade: Nit: Let's split this in a separate test

test/src/worker.spec.ts
```diff
@@ -179,4 +180,201 @@ describe('Workers', function () {
...
+      });
+      expect(
+        messages.map(msg => {
+          return msg.type();
+        }),
+      ).toEqual(['trace', 'dir', 'warn', 'error', 'log']);
+      const texts = messages.map(msg => {
+        return msg.text();
+      });
+      try {
```
Lightning00Blade: I think there is a `protocol` or similar to getTestState

test/src/worker.spec.ts
```diff
@@ -179,4 +180,201 @@ describe('Workers', function () {
...
+      );
+      await worker.evaluate(() => {
+        (globalThis as any).test = 1;
+        console.log(1, 2, 3, globalThis);
+      });
+      const log = await logPromise;
+
+      expect(log.text()).atLeastOneToContain([
+        '1 2 3 [object Object]',
+        '1 2 3 [object DedicatedWorkerGlobalScope]',
```
Lightning00Blade, line 382: Why do we have 3 options here, wont it be CDP and BiDi?

Lightning00Blade (commented):
Over call LGTM, maybe we can think of better places than Utils to put things as that is becoming a dumping ground, which is not nice.

packages/puppeteer-core/src/cdp/Page.ts
nroscino: Maybe I missunderstood this part. 
I saw that we were creating the handle but disposing them right after if the `if (!this.listenerCount(PageEvent.Console))` was returning true. 
I moved the if before the creation so that we don't even bother creating and disposing them if the `listenerCount` if returns false straight away.
Does it makes sense to you or I am misunderstanding something here?

Lightning00Blade: In short we create the handle so we have the nice wrappers for the API we call but we must call the CDP method - https://chromedevtools.github.io/devtools-protocol/tot/Runtime/#method-releaseObject

Lightning00Blade: Else the anything that is in the args like `console.log({a:1}, {b:2})` can't be garbage collected.

test/src/worker.spec.ts
nroscino: All the test suite was a transposition of the `console.spec.ts` with changes to test workers, so I changed only things that were strictly needed. This test is skipped on bidi, so I wasn't spending much time here to understand. It's also implemented like this for `console.spec.ts` and skipped. Maybe we can then dive deeper on the issue in a separate task, WDYT?

test/src/worker.spec.ts
nroscino: Ok! Since I am splitting, should I split it also for `console.spec.ts`?

packages/puppeteer-core/src/bidi/Realm.ts
```diff
@@ -400,6 +408,41 @@ export class BidiWorkerRealm extends BidiRealm {
...
+    this.realm.on('log', entry => {
+      if (
+        isConsoleLogEntry(entry) &&
+        this.#worker.listenerCount(WebWorkerEvent.Console)
+      ) {
+        const args = entry.args.map(arg => {
+          return this.createHandle(arg);
+        });
+
+        const text = args
```
Lightning00Blade: This code looks duplicated, can we extract it?

packages/puppeteer-core/src/bidi/util.ts
```diff
@@ -6,11 +6,70 @@
...
+import type {
+  ConsoleMessageLocation,
+  ConsoleMessageType,
+} from '../common/ConsoleMessage.js';
 import {ProtocolError, TimeoutError} from '../common/Errors.js';
 import {PuppeteerURL} from '../common/util.js';
 
 import {BidiDeserializer} from './Deserializer.js';
 
+// TODO: Remove this and map CDP the correct method.
```
Lightning00Blade: This should now be part of the TSDoc

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -905,17 +895,28 @@ export class CdpPage extends Page {
   #onConsoleAPI(
     world: IsolatedWorld,
     event: Protocol.Runtime.ConsoleAPICalledEvent,
+    values?: JSHandle[],
```
Lightning00Blade, line 916: Where do we call it with this now?

test/src/worker.spec.ts
Lightning00Blade: That test is wrong, there should be two values.
I will send a small PR to address that.

test/src/worker.spec.ts
```diff
@@ -179,4 +180,247 @@ describe('Workers', function () {
...
+        '1 2 3 JSHandle@object', // WebDriver BiDi
```
Lightning00Blade:
```suggestion
        '1 2 3 [object DedicatedWorkerGlobalScope]',
        '1 2 3 JSHandle@window', // WebDriver BiDi
```

packages/puppeteer-core/src/cdp/Page.ts
nroscino: This is called from the page worker logic. The page workers were handing of the console message directly to the page, now we can listen also to them,
so if we do we avoid creating new handles for the same object and we just take the one already created from the worker itself.

packages/puppeteer-core/src/bidi/Realm.ts
```diff
@@ -400,6 +406,28 @@ export class BidiWorkerRealm extends BidiRealm {
...
+        isConsoleLogEntry(entry) &&
+        this.#worker.listenerCount(WebWorkerEvent.Console)
+      ) {
+        const args = entry.args.map(arg => {
+          return this.createHandle(arg);
+        });
+
+        const message = getConsoleMessage(
+          entry,
+          args as BidiJSHandle[],
```
Lightning00Blade: Why is this cast needed? Doesn't the createHandle return the correct type?

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -178,30 +162,16 @@ export class BidiFrame extends Frame {
...
           PageEvent.Console,
-          new ConsoleMessage(
-            convertConsoleMessageLevel(entry.method),
-            text,
-            args,
-            getStackTraceLocations(entry.stackTrace),
-            this,
-            undefined,
-          ),
+          getConsoleMessage(entry, args as BidiJSHandle[], this),
```
Lightning00Blade: Why is this cast needed as well?

Lightning00Blade (commented):
LGTM + 2 small nits.

Lightning00Blade (approved):

packages/puppeteer-core/src/bidi/Realm.ts
nroscino: It returns the type as `BidiJSHandle<unknown> | BidiElementHandle<Node>` and the `getConsoleMessage` is taking `BidiJSHandle`. I changed the param in getConsoleMessage and added the instanceof check that was there in the previous implementation!

Lightning00Blade: BidiElementHandle extends BidiJSHandle, so what is the issue there?

OrKoN (approved):
@Lightning00Blade could you please take another look as well?

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -178,30 +161,16 @@ export class BidiFrame extends Frame {
...
           PageEvent.Console,
-          new ConsoleMessage(
-            convertConsoleMessageLevel(entry.method),
-            text,
-            args,
-            getStackTraceLocations(entry.stackTrace),
-            this,
-            undefined,
-          ),
+          getConsoleMessage(entry, args, this),
```
Lightning00Blade, line 173: To reduce the helpers and bloating the `utils` file, we can put this on the ConsoleMessage class as a `static from` similar to other classes.
This can be done in a follow up.

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -368,11 +358,28 @@ export class CdpPage extends Page {
         session.target().url(),
         session.target()._targetId,
         session.target().type(),
-        this.#onConsoleAPI.bind(this),
         this.#handleException.bind(this),
         this.#frameManager.networkManager,
       );
       this.#workers.set(session.id(), worker);
+      worker.on(WebWorkerEvent.Console, message => {
```
Lightning00Blade: One limitation that I didn't consider is that if we use the public event listener.
A user may break by calling removeAll event listeners.  
CC @OrKoN

packages/puppeteer-core/src/cdp/utils.ts
```diff
@@ -7,9 +7,51 @@
...
 import type {JSHandle} from '../api/JSHandle.js';
-import {PuppeteerURL, evaluationString} from '../common/util.js';
+import {
+  ConsoleMessage,
+  type ConsoleMessageType,
+} from '../common/ConsoleMessage.js';
+import {PuppeteerURL, debugError, evaluationString} from '../common/util.js';
 import {assert} from '../util/assert.js';
 
+export {debugError};
```
Lightning00Blade: Why is this export needed?

Lightning00Blade (approved):

packages/puppeteer-core/src/cdp/Page.ts
OrKoN: Good point, I think perhaps we should create an internal emitter

packages/puppeteer-core/src/cdp/utils.ts
nroscino: it's not strictly needed, I can remove it. It just made the import in page/webworker cleaner.

packages/puppeteer-core/src/cdp/Page.ts
nroscino: I implemented an internal listener on the WebWorker! PTAL and let me know what you think! @Lightning00Blade, @OrKoN

OrKoN:
I think the PR should be marked as `feat: `

OrKoN (approved):

Lightning00Blade (approved):

PR #14813, opened 2026-03-31
feat(browsers): report unpacking progress

### LINKED ISSUE #13083: [Feature]: Add unpacking indicator to the install command

yzlu0917:
## Summary
  - report when archive extraction starts during the default `@puppeteer/browsers
  install` flow
  - keep the existing download progress bar behavior while adding a clear unpacking
  status line
  - add a CLI regression test that asserts the unpacking message is emitted

  ## Why
  After the download finishes, the install command can spend noticeable time extracting
  the archive without printing any terminal feedback. That makes the command appear
  stuck even though installation is still progressing.

  ## User Impact
  Users now get an explicit `Unpacking <browser> <buildId>` status update after the
  download completes, so the install flow remains transparent during archive
  extraction.

  ## Validation
  - ran `npm run build:test --workspace @puppeteer/browsers`
  - ran `npx mocha --config .mocharc.cjs test/build/CLI.spec.js --grep "should report
  when unpacking starts"` in `packages/browsers`

  Closes #13083

### DIFF
--- a/docs/browsers-api/browsers.installoptions.md
+++ b/docs/browsers-api/browsers.installoptions.md
@@ -140,7 +140,7 @@ Determines the path to download browsers to.
 
 </td><td>
 
-Provides information about the progress of the download. If set to 'default', the default callback implementing a progress bar will be used.
+Provides information about the progress of the download. If set to 'default', the default callback implementing a progress bar and install status updates will be used.
 
 </td><td>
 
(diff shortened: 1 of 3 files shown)


OrKoN:
The PR does not seem to report progress, only adds a message once. Closing as stale.

PR #14814, opened 2026-04-01
feat(webmcp): add initial API to inspect tool registrations

beaufortfrancois:

This PR introduces a working prototype of the [WebMCP API](https://webmachinelearning.github.io/webmcp/).

Yes

Not yet.

**Summary**

Currently Puppeteer doesn't have a defined API to interact with WebMCP. Using Puppeteer would come handy in eval testing: https://github.com/GoogleChromeLabs/webmcp-tools/issues/68

None that I'm aware of.

**Other information**

This is highly experimental.


OrKoN (changes requested):
Thanks for the PR but it does not quite match the intended API per the design doc (the API is too low-level and closely following CDP and does not offer a convenient way to list tools while offering a fine-grained control over the CDP domain). We also have this planned for the next milestone. Perhaps we should chat if you would like to implement the proposed API?

test/src/cdp/WebMCP.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import expect from 'expect';
+import {getTestState, setupTestBrowserHooks} from '../mocha-utils.js';
+
+describe('WebMCP', function () {
+  setupTestBrowserHooks();
+
+  it('should enable and disable WebMCP domain', async () => {
```
OrKoN: we should create a test page with some tools for testing this. Note that Puppeteer uses the stable Chrome release so tests would fail in CI for now.

beaufortfrancois:
> Thanks for the PR but it does not quite match the intended API per the design doc (the API is too low-level and closely following CDP and does not offer a convenient way to list tools while offering a fine-grained control over the CDP domain). We also have this planned for the next milestone. Perhaps we should chat if you would like to implement the proposed API?

As discussed offline, @OrKoN feel free to use this PR as a starting point when implementing the real Puppeteer API for WebMCP

beaufortfrancois:
@OrKoN FYI I've just updated this PR to get `page.webmcp.tools()` working as in your proposal.

test/src/cdp/WebMCP.spec.ts
beaufortfrancois: Done.

test/TestExpectations.json
```diff
@@ -1970,5 +1961,20 @@
...
+    "testIdPattern": "[webmcp.spec] Page.webmcp.tools should monitor registered and unregistered tools",
+    "platforms": ["linux"],
+    "parameters": [
+      "cdp",
+      "chrome",
+      "chrome-headless-shell",
+      "headful",
+      "headless",
+      "pipe",
+      "webDriverBiDi"
```
OrKoN: this is not going to work as parameters are joined using logical AND and no test run matches `cdp` AND `webDriverBiDi` for instance.

test/TestExpectations.json
```diff
@@ -1970,5 +1961,20 @@
     "parameters": ["cdp", "chrome", "chrome-headless-shell"],
     "expectations": ["FAIL", "PASS"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
+  },
+  {
+    "testIdPattern": "[webmcp.spec] Page.webmcp.tools should monitor registered and unregistered tools",
+    "platforms": ["linux"],
```
OrKoN: this needs to list all platforms

test/TestExpectations.json
```diff
@@ -1970,5 +1961,20 @@
     "parameters": ["cdp", "chrome", "chrome-headless-shell"],
     "expectations": ["FAIL", "PASS"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
+  },
+  {
+    "testIdPattern": "[webmcp.spec] Page.webmcp.tools should monitor registered and unregistered tools",
```
OrKoN: `[webmcp.spec] *` would be sufficient for all tests

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * @public
+ */
+export type WebMCPInvocationStatus = 'Success' | 'Canceled' | 'Error';
+
+/**
+ * @public
+ */
+export interface WebMCPTool {
```
OrKoN: Let's make this a class.

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -0,0 +1,116 @@
...
+  name: string;
+  description: string;
+  inputSchema?: object;
+  annotations?: WebMCPAnnotation;
+  frameId: string;
+  backendNodeId?: number;
```
OrKoN, line 40: this should be replaced with a ElementHandle getter (https://pptr.dev/api/puppeteer.elementhandle) or we can omit it.

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * @public
+ */
+export interface WebMCPTool {
+  name: string;
+  description: string;
+  inputSchema?: object;
+  annotations?: WebMCPAnnotation;
+  frameId: string;
```
OrKoN, line 39: This should be a Frame getter https://pptr.dev/api/puppeteer.frame

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -293,6 +296,7 @@ export class CdpPage extends Page {
     this.#touchscreen.updateClient(newSession);
     this.#emulationManager.updateClient(newSession);
     this.#tracing.updateClient(newSession);
+    this.#webmcp.updateClient(newSession);
```
OrKoN, line 299: I think we would want to handle it in the FrameManager to cover OOPIFs.

OrKoN: I guess we need to decide if webmcp instance would handle data from all frames or if it should be per frame.

test/TestExpectations.json
beaufortfrancois: Done

test/TestExpectations.json
beaufortfrancois: Done

test/TestExpectations.json
beaufortfrancois: I've split them. Let me know if that looks better.

packages/puppeteer-core/src/cdp/WebMCP.ts

packages/puppeteer-core/src/cdp/WebMCP.ts
beaufortfrancois: Shall we have both? `frameId` in WebMCPTool and `get frame() : Frame {...}` for easy access?

packages/puppeteer-core/src/cdp/WebMCP.ts
beaufortfrancois: Is this the proper way to get it?
```
const client = await page.createCDPSession();
const { object } = await client.send('DOM.resolveNode', { backendNodeId });
new CdpElementHandle(object)
```

OrKoN: The correct way would be `adoptBackendNode` https://github.com/puppeteer/puppeteer/blob/2be30023994ee2e7ebb15e43dc0e2804256f8ca2/packages/puppeteer-core/src/cdp/IsolatedWorld.ts#L199 for that we need to keep the reference to the frame the event came from so that we can access its main world.

OrKoN: here is an example https://github.com/puppeteer/puppeteer/blob/2be30023994ee2e7ebb15e43dc0e2804256f8ca2/packages/puppeteer-core/src/cdp/Page.ts#L436

packages/puppeteer-core/src/cdp/WebMCP.ts
OrKoN: we don't expose frameIds in Puppeteer as we consider those to be implementation details. It's available as `frame._id` for internal use and if we want to make it public we would expose it on the Frame (so no need to have a frameId getter).

test/src/cdp/webmcp.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+ * Copyright 2026 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import expect from 'expect';
+import type {WebMCPTool} from 'puppeteer-core';
+
+import {setupSeparateTestBrowserHooks} from '../mocha-utils.js';
+
+describe('Page.webmcp.tools', function () {
```
OrKoN:
```suggestion
describe('Page.webmcp', function () {
```

test/src/cdp/webmcp.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+              text: {type: 'string', description: 'Some text'},
+            },
+            required: ['text'],
+          },
+          execute: () => {},
+          annotations: {readOnlyHint: true},
+        },
+        {signal: (window as any).controller1.signal},
+      );
+    });
```
OrKoN, line 56: let's extract a helper function to register a tool:

```
const controllerHandle = await page.evaluateHandle(() => {
   const controller = new AbortController();
   navigator.modelContext.registerTool(..., {cintroller})
   return controller;
});

await controllerHandle.evaluate(el => el.abort())
```

Let's split the long test into multiple, 1) asserting tool data 2) testing added event 3) testing removed event

OrKoN: the helper can be inline in the test file (under the describe block).

test/TestExpectations.json
```diff
@@ -216,6 +216,62 @@
...
+    "expectations": ["FAIL"],
+    "comment": "Experimental support for WebMCP requires Chrome."
+  },
+  {
+    "testIdPattern": "[webmcp.spec] *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["cdp"],
+    "expectations": ["FAIL"],
+    "comment": "Experimental support for WebMCP in Chrome requires flags that are not yet in Chrome Stable."
+  },
```
OrKoN: this one can be removed.

test/TestExpectations.json
```diff
@@ -216,6 +216,62 @@
...
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox"],
+    "expectations": ["FAIL"],
+    "comment": "Experimental support for WebMCP requires Chrome."
+  },
+  {
+    "testIdPattern": "[webmcp.spec] *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["cdp"],
+    "expectations": ["FAIL"],
```
```suggestion
    "expectations": ["SKIP"],
```

test/TestExpectations.json
```diff
@@ -216,6 +216,62 @@
...
+    "comment": "Experimental support for WebMCP in Chrome requires flags that are not yet in Chrome Stable."
+  },
+  {
+    "testIdPattern": "[webmcp.spec] *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["webDriverBiDi"],
+    "expectations": ["FAIL"],
+    "comment": "Experimental support for WebMCP is not supported in BiDi"
+  },
```
OrKoN: These ones can be removed.

test/TestExpectations.json
```diff
@@ -216,6 +216,62 @@
...
+    "parameters": ["chrome"],
+    "expectations": ["FAIL"],
+    "comment": "Experimental support for WebMCP in Chrome requires flags that are not yet in Chrome Stable."
+  },
+  {
+    "testIdPattern": "[webmcp.spec] *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["chrome-headless-shell"],
+    "expectations": ["FAIL"],
+    "comment": "Experimental support for WebMCP in Chrome requires flags that are not yet in Chrome Stable."
```
OrKoN: I think headless shell will not have it since implementation is partially in chrome// so we could skip

beaufortfrancois: Done

test/TestExpectations.json
beaufortfrancois: Removed

test/TestExpectations.json
beaufortfrancois: Resolved with https://github.com/puppeteer/puppeteer/pull/14814/changes#r3028166590

test/src/cdp/webmcp.spec.ts
beaufortfrancois: Done

packages/puppeteer-core/src/cdp/WebMCP.ts
beaufortfrancois: Done

packages/puppeteer-core/src/cdp/WebMCP.ts
beaufortfrancois: Done. Thanks!

test/src/cdp/webmcp.spec.ts
beaufortfrancois: I've noticed `el.abort()` is not executed. Could it be that the realm isn't right?

OrKoN: No, both are in the main world, same execution context.  It could be some other issue.

beaufortfrancois: FWIW

```
    console.log('before controllerHandle.evaluate');
    await controllerHandle.evaluate(el => {
      console.log(el);
      el.abort();
    });
    console.log('after controllerHandle.evaluate');
```

prints only

```
before controllerHandle.evaluate
after controllerHandle.evaluate
```

OrKoN: The middle console.log runs on the page. You need to listen to `page.on('console', msg => ...)` on the Node.js side to get console messages from the page.

beaufortfrancois: Thanks!
```
    page.on('console', msg => {
      console.log(msg.text());
    })
    console.log('before controllerHandle.evaluate');
    await controllerHandle.evaluate(el => {
      console.log(el.signal.aborted);
      el.abort();
      console.log(el.signal.aborted);
    });
    console.log('after controllerHandle.evaluate');
```
now prints successfully
```
before controllerHandle.evaluate
false
true
after controllerHandle.evaluate
```
but the tool is not getting unregistered ;(

beaufortfrancois: I'll add LOGs to a custom chromium build to debug this then

beaufortfrancois: Is it possible to get chromium stderr logs when running `npm run test:chrome`?

OrKoN: yeah DUMPIO=true env when running the tests in Puppeteer, or dumpio: true when launching puppeteer in standalone scripts

beaufortfrancois: Got it! It looks like `{signal: controller.signal}` works while `{controller}` does NOT. Is this expected there?

OrKoN: Sorry I was just writing pseudo-code. The signal/controller would work exactly as it would work on the page.

beaufortfrancois: Then we're good. I thought destructuring controller would directly map it to signal somehow

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -0,0 +1,181 @@
...
+ */
+
+export class WebMCP extends EventEmitter<{
+  /** Emitted when tools are added. */
+  toolsadded: WebMCPToolsAddedEvent;
+  /** Emitted when tools are removed. */
+  toolsremoved: WebMCPToolsRemovedEvent;
+}> {
+  #client: CDPSession;
+  #frameManager: FrameManager;
```
OrKoN, line 135: I think we would need some extra handling within WebMCP on top of the FrameManager:

1) when a frame is removed (or a subtree of frames), the tool removed events would not be emitted
2) when the frame navigates we should also remove its tools as CDP events won't be emitter

I think FrameManagerEvents include all necessary events to track this.

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -0,0 +1,181 @@
...
+        this.#tools.set(tool.frameId, frameTools);
+      }
+
+      using formElement = tool.backendNodeId
+        ? ((await frame.worlds[MAIN_WORLD].adoptBackendNode(
+            tool.backendNodeId,
+          )) as ElementHandle<HTMLFormElement>)
+        : undefined;
+
+      const addedTool = new WebMCPTool(tool, frame, formElement);
```
OrKoN: `using` would dispose the formElement handle at the end of the block. We should `move()` it when we construct WebMCPTool.

```suggestion
      const addedTool = new WebMCPTool(tool, frame, formElement.move());
```

test/src/cdp/webmcp.spec.ts
```diff
@@ -0,0 +1,229 @@
...
+  });
+
+  it('should list tools', async () => {
+    const {page, httpsServer} = state;
+    await page.goto(httpsServer.EMPTY_PAGE);
+
+    expect(page.webmcp).toBeDefined();
+
+    // FIXME: Find a way to remove this.
+    await page.webmcp.tools();
```
OrKoN: let's add `initialize()` method to WebMCP and call it from #initialize(). Within it, let's send WebMCP.enable CDP command. That would allow us to listen for WebMCP data automatically (if there are no subscribers to events it would be largely no-op) and it would allow us to drop `await` in front of the `tools()` getter. I think there should be no issue with handling WebMCP events but eventually if needed we could turn it off by omitting the initialize() calls if the user opts out.

OrKoN: Let's also catch the errors in the initialize() to not fail for older Chrome versions that do not have the WebMCP CDP domain.

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -0,0 +1,181 @@
...
+ * @public
+ */
+export class WebMCPTool {
+  name: string;
+  description: string;
+  inputSchema?: object;
+  annotations?: WebMCPAnnotation;
+  frame: Frame;
+  formElement?: ElementHandle<HTMLFormElement>;
+  stackTrace?: unknown;
```
OrKoN: I think this would be `Protocol.Runtime.StackTrace` instead of unknown. Puppeteer does not have a good API for stack traces right now so perhaps let's create a ConsoleMessageLocation out of the raw stacktrace's first frame and keep the raw stack trace internal for now.

beaufortfrancois: Done. I've added `location` to WebMCPTool and renamed WebMCPTool internal `stackTrace` to `rawStackTrace`.

packages/puppeteer-core/src/cdp/WebMCP.ts

test/src/cdp/webmcp.spec.ts
beaufortfrancois: Done

packages/puppeteer-core/src/cdp/WebMCP.ts
beaufortfrancois: 1. WebMCP is currently supported only of the top-level frame. See https://github.com/webmachinelearning/webmcp/issues/57
2. Done.

packages/puppeteer-core/src/cdp/Page.ts
beaufortfrancois: FYI WebMCP **currently** restricts only the top-level browsing context to declare tools. See  https://github.com/webmachinelearning/webmcp/issues/57

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -943,6 +944,12 @@ export abstract class Page extends EventEmitter<PageEvents> {
    */
   abstract get tracing(): Tracing;
 
+  /**
+   * TODO
```
OrKoN:
```suggestion
   * Experimental API for {@link WebMCP | https://github.com/webmachinelearning/webmcp}. Only available in Chrome Canary with the following Chrome features enabled `--enable-features=WebMCPTesting,DevToolsWebMCPSupport`.
```

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -0,0 +1,216 @@
...
+/**
+ * @public
+ */
+export class WebMCPTool {
+  name: string;
+  description: string;
+  inputSchema?: object;
+  annotations?: WebMCPAnnotation;
+  frame: Frame;
+  formElement?: ElementHandle<HTMLFormElement>;
```
OrKoN: let's make a getter for formElement and create the ElementHandle only when it's called. We can cache the handle or create a new one each time I think.

packages/puppeteer-core/src/cdp/WebMCP.ts
OrKoN: I see, I think I saw some CLs enabling it for frames but if it is not supported, the current version is fine.

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -0,0 +1,216 @@
...
+        return null;
+      }
+
+      const frameTools = this.#tools.get(tool.frameId) ?? new Map();
+      if (!this.#tools.has(tool.frameId)) {
+        this.#tools.set(tool.frameId, frameTools);
+      }
+
+      using formElement = tool.backendNodeId
+        ? ((await frame.worlds[MAIN_WORLD].adoptBackendNode(
```
OrKoN: With my previous comment, what are after it to remove await here so that events are emitted in the same order as they happen. An await here might change the order of events unpredictably.

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  tools(): WebMCPTool[] {
+    return Array.from(this.#tools.values()).flatMap(toolMap => {
+      return Array.from(toolMap.values());
+    });
+  }
+
+  #bindListeners(): void {
+    // We use type casting because WebMCP is not yet in the Protocol types.
```
```suggestion
    // TODO: Remove type-casting. We use type casting because WebMCP is not yet in the Protocol types.
```

packages/puppeteer-core/src/api/Page.ts
OrKoN: Is there a specific Chrome version we target? Any additional information that can be relevant to uses?

OrKoN (commented):
I think we can land this after the following changes.

OrKoN:
Renamed to `feat(webmcp): add initial API to inspect tool registrations` as we will follow up with extensions to invoke tools and also monitoring of tool invocations.

packages/puppeteer-core/src/api/Page.ts
beaufortfrancois: Done.

I'll go with Chrome 148+ for now since Chrome 149 has just branched and WebMCP DevTools support has recently landed.

I've added "Supports tool discovery only; invocation is currently unsupported." as well since https://chromium-review.googlesource.com/c/chromium/src/+/7696374 has not landed yet.

packages/puppeteer-core/src/cdp/WebMCP.ts
beaufortfrancois: We're good then

packages/puppeteer-core/src/cdp/WebMCP.ts

packages/puppeteer-core/src/cdp/WebMCP.ts

packages/puppeteer-core/src/cdp/WebMCP.ts
beaufortfrancois: I've removed `await` there.

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -0,0 +1,223 @@
...
+
+import type {Protocol} from 'devtools-protocol';
+
+import type {CDPSession} from '../api/CDPSession.js';
+import type {Frame} from '../api/Frame.js';
+import type {ConsoleMessageLocation} from '../common/ConsoleMessage.js';
+import {EventEmitter} from '../common/EventEmitter.js';
+import {debugError} from '../common/util.js';
+import type {ElementHandle} from '../puppeteer-core.js';
+import {MAIN_WORLD} from '../puppeteer-core.js';
```
OrKoN: we should import directly from the specific file instead of entrypoints like puppeteer-core to avoid circular imports (here and on the previous line)

OrKoN (approved):
LGTM thanks!

PR #14820, opened 2026-04-02
feat: Blocked and allowed origins

nattallius:


packages/puppeteer-core/src/cdp/TargetManager.ts
```diff
@@ -404,8 +407,21 @@ export class TargetManager
         autoAttach: true,
         filter: this.#discoveryFilter,
       }),
-      session.send('Runtime.runIfWaitingForDebugger'),
+      this.#networkConditions
+        ? this.#setupNetworkConditions(session, this.#networkConditions)
+        : Promise.resolve(),
     ]).catch(debugError);
+
+    parentSession.emit(CDPSessionEvent.Ready, session);
```
OrKoN: I do not think we should change the order here and separate runIfWaitingForDebugger for debugging calls.

PR #14821, opened 2026-04-02
feat: API to list installed browser extensions and trigger extension actions

nroscino:
This PR introduces the first part of the implementation for Extension API. This includes the changes related to the creation of the Extension class and the methods to retrieve it through the browser.

Yes

Yes

**Summary**
Currently Puppeteer doesn't have a defined API to interact with extensions. 

No

**Other information**


packages/puppeteer-core/src/bidi/Browser.ts
```diff
@@ -375,4 +376,16 @@ export class BidiBrowser extends Browser {
   override isNetworkEnabled(): boolean {
     return this.#networkEnabled;
   }
+
+  override extensions(): Promise<Extension[]> {
+    throw new Error(
```
OrKoN: we should use UnsupportedOperation error class here and below

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -545,4 +546,27 @@ export class CdpBrowser extends BrowserBase {
...
+      return new Extension(
+        extension.id,
+        extension.version,
+        extension.name,
+        this,
+      );
+    });
+  }
+
+  override async getExtensionById(
```
OrKoN: I wonder if we really need this helper or if calling find on the extensions() getter result is as good for callers.

test/src/extension.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+);
+
+describe('Extensions', function () {
+  setupTestBrowserHooks();
+
+  it('should list extensions and their properties', async () => {
+    const {defaultBrowserOptions, isChrome} = await getTestState({
+      skipLaunch: true,
+    });
+    if (!isChrome) {
```
OrKoN: we should skip tests using TestExpectations.json

test/src/extension.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+  import.meta.dirname,
+  '/../assets/extension-with-page',
+);
+
+describe('Extensions', function () {
+  setupTestBrowserHooks();
+
+  it('should list extensions and their properties', async () => {
+    const {defaultBrowserOptions, isChrome} = await getTestState({
+      skipLaunch: true,
```
OrKoN: I think we do not need explicitSkipLaunch. Let's use existing test/src/cdp/extensions.spec.ts that already has correct launch setup?

test/TestExpectations.json
```diff
@@ -1878,6 +1878,104 @@
     "expectations": ["FAIL", "PASS"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
+  {
```
OrKoN: I think you can delete most of these expectations once the tests are moved. For the ones you need you can use wildcards to apply expectation for all tests in the spec.

packages/puppeteer-core/src/cdp/Browser.ts
nroscino: I don't think there's a strict need for this. I think it can make the user journey of getting an extension from id a little more convenient cause it does it in one step instead of having 2, but I am open to remove it.

OrKoN: Can we return a map in getExtensions instead since the IDs are unique?

packages/puppeteer-core/src/api/Extension.ts
```diff
@@ -0,0 +1,110 @@
...
+export class Extension {
+  // needed to access the CDPSession to trigger an extension action.
+  #browser: Browser;
+  #id: string;
+  #version: string;
+  #name: string;
+
+  #pages = new WeakMap<Target, Page>();
+
+  constructor(id: string, version: string, name: string, browser: Browser) {
```
Lightning00Blade: This needs to be marked as internal, users should not be allowed to new up the class.

packages/puppeteer-core/src/api/Extension.ts
```diff
@@ -0,0 +1,110 @@
...
+        (target.type() === 'page' || target.type() === 'background_page') &&
+        targetUrl.startsWith('chrome-extension://' + this.#id)
+      );
+    });
+
+    const pages: Page[] = [];
+    for (const target of extensionPages) {
+      let page = this.#pages.get(target);
+      if (!page) {
+        page = await target.asPage();
```
Lightning00Blade: Does this not return the same page each time? 
That sounds like a bug.
CC @OrKoN

packages/puppeteer-core/src/api/Extension.ts
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+      pages.push(page);
+    }
+
+    return pages;
+  }
+
+  async triggerAction(page: Page): Promise<void> {
+    const client = await this.#browser.target().createCDPSession();
```
Lightning00Blade: We should not create a new session but use the interal session, there should be a getter on the CdpBrowser if I remember correctly.

packages/puppeteer-core/src/api/Extension.ts
```diff
@@ -0,0 +1,110 @@
+/**
+ * @license
+ * Copyright 2026 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+/// <reference types="node"  preserve="true"/>
```
Lightning00Blade: This should not be added I think.

packages/puppeteer-core/src/api/Extension.ts
```diff
@@ -0,0 +1,110 @@
+/**
```
Lightning00Blade, line 1: I see a lot of CDP implementation here, should this live there and we extend it like other things?

packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -217,6 +218,10 @@ export class BidiPage extends Page {
     return this.#frame;
   }
 
+  override async triggerExtensionAction(extension: Extension): Promise<void> {
+    return await extension.triggerAction(this);
```
Lightning00Blade: I am confused why are we adding this API if BiDi does not support Extension?

packages/puppeteer-core/src/cdp/Browser.ts
Lightning00Blade: As Alex said maybe a map would be nice.

test/src/extension.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+
+  it('should trigger extension action', async () => {
+    const {defaultBrowserOptions, isChrome} = await getTestState({
+      skipLaunch: true,
+    });
+    if (!isChrome) {
+      return;
+    }
+
+    const options = Object.assign({}, defaultBrowserOptions);
```
Lightning00Blade: New object with spreader is preferred in such cases.
(And FYI even structuredClone if there are arrays and child objects)

packages/puppeteer-core/src/bidi/Page.ts
Lightning00Blade: Or this should just live on the api.

packages/puppeteer-core/src/api/Extension.ts
```diff
@@ -0,0 +1,110 @@
...
+
+import type {Browser} from './Browser.js';
+import type {Page} from './Page.js';
+import type {Target} from './Target.js';
+import type {WebWorker} from './WebWorker.js';
+
+/**
+ * {@link Extension} represents an Extension instance installed in the browser.
+ *
+ * @public
```
Lightning00Blade, line 14: Let's mark it as experimental for a little bit so we can make more changes to the API if we find something while using it.

packages/puppeteer-core/src/api/Extension.ts
OrKoN: Yeah maybe we need to handle this within asPage

packages/puppeteer-core/src/api/Extension.ts
OrKoN: Yeah, I think the issue here is that the CDP code is in the api/ folder. We should just keep the interface in the API and implement the constructor in cdp/ sub-folder so that it can take a CdpBrowser instance when extension instances are constructed

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -545,4 +546,27 @@ export class CdpBrowser extends BrowserBase {
   override isNetworkEnabled(): boolean {
     return this.#networkEnabled;
   }
+
+  override async extensions(): Promise<Extension[]> {
+    const response = await this.#connection.send('Extensions.getExtensions');
+    return response.extensions.map(extension => {
+      return new Extension(
```
OrKoN: we probably should not create new instances each time to maintain instance identity. Let's cache the instances and update the list here. WDYT?

packages/puppeteer-core/src/cdp/Browser.ts
nroscino: I like this idea! yeah we definitely can

packages/puppeteer-core/src/api/Browser.ts
```diff
@@ -727,4 +728,11 @@ export abstract class Browser extends EventEmitter<BrowserEvents> {
    * @internal
    */
   abstract isNetworkEnabled(): boolean;
+
+  /**
+   * Get a list of the installed extensions in the browser.
```
Lightning00Blade: Text need to be update to reflect the change.

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -545,4 +556,28 @@ export class CdpBrowser extends BrowserBase {
...
+        extension.id,
+        extension.version,
+        extension.name,
+        this,
+      );
+      this.#extensions.set(extension.id, newExtension);
+      return newExtension;
+    });
+
+    this.#extensions = new Map(
```
Lightning00Blade: Why do we need to re-set the map here? We do it on like 572

packages/puppeteer-core/src/api/Browser.ts
nroscino: Updated!

packages/puppeteer-core/src/api/Extension.ts
nroscino: I can address this in a separate PR

test/src/cdp/extensions.spec.ts
```diff
@@ -23,6 +32,7 @@ describe('extensions', function () {
     {
       enableExtensions: true,
       args: [`--load-extension=${extensionPath}`],
+      pipe: true,
```
OrKoN, line 35: I believe enableExtensions turns that on?

OrKoN (approved):

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -545,4 +557,31 @@ export class CdpBrowser extends BrowserBase {
...
   }
+
+  override async extensions(): Promise<Map<string, Extension>> {
+    const response = await this.#connection.send('Extensions.getExtensions');
+
+    const extensionsMap = new Map<string, Extension>();
+
+    for (const currExtension of response.extensions) {
+      if (this.#extensions.has(currExtension.id)) {
+        extensionsMap.set(
```
Lightning00Blade, line 568: I think we can just skip if we already have it no?

packages/puppeteer-core/src/cdp/Extension.ts
```diff
@@ -0,0 +1,82 @@
...
+        (target.type() === 'page' || target.type() === 'background_page') &&
+        targetUrl.startsWith('chrome-extension://' + this.id)
+      );
+    });
+
+    const pages: Page[] = [];
+    for (const target of extensionPages) {
+      let page = this.#pages.get(target);
+      if (!page) {
+        page = await target.asPage();
```
Lightning00Blade, line 64: Good practice is to leave a TODO, so we can keep track that we want to make this return the same page.

test/src/cdp/extensions.spec.ts
```diff
@@ -61,4 +71,157 @@ describe('extensions', function () {
     }
     expect(result).toBe(42);
   });
+
+  it('should list extensions and their properties', async () => {
```
Lightning00Blade, line 75: Maybe we should add a test that makes Puppeteer connect to the instance to make sure we can see previously installed extensions.

Lightning00Blade (approved):
LGTM with some small comments.

test/src/cdp/extensions.spec.ts
nroscino: when I run test without it told me "Method not available"

packages/puppeteer-core/src/cdp/Browser.ts
nroscino: if I skip what I already had in `this.#extensions` then at the line:
```typescript
this.#extensions = extensionsMap;
```
I will end up having in `this.#extensions` only the one that are new and were not present before

test/src/cdp/extensions.spec.ts
nroscino: I am not sure what we would be testing here 🤔 cause if it fails the problem is either in the browser (not persisting correctly extensions through sessions) or the CDP command is not returning the correct data, but both are outside of the scope of puppeteer. WDYT?

Lightning00Blade: We we would be testing that Puppeteer can detect previously installed extensions.
Basically use `Puppeterr.BrowserA` to install (because we don't have another way), and then `Puppeteer.BrowserB` check that it can find one that is already there.

packages/puppeteer-core/src/cdp/Browser.ts
Lightning00Blade: I miss interpret the logic initially, we do this because we want to remove stale version.

Lightning00Blade (approved):

PR #14824, opened 2026-04-02
feat: adds extension realms api

nroscino:
This PR introduces Realms to handle different execution contexts. This is a follow up PR connected to the #14821

Yes

Yes

**Summary**

No

**Other information**


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -2752,6 +2746,14 @@ export abstract class Page extends EventEmitter<PageEvents> {
     return typedArray;
   }
 
+  /**
+   * Triggers an extension action for the given extension.
+   *
+   * @param extension - The extension to trigger the action for.
+   * @public
+   */
+  abstract triggerExtensionAction(extension: Extension): Promise<void>;
```
OrKoN: any reason to re-order methods?

packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -15,30 +15,57 @@ import {disposeSymbol} from '../util/disposable.js';
...
+   * The identifier for this realm.
+   *
+   * @public
+   */
+  abstract get worldId(): string | symbol;
+
+  abstract set worldId(worldId: string | symbol);
+
+  /**
+   * This method returns the extension from the ExecutionContext paired with the realm
```
OrKoN: ExecutionContext is not a concept in our public API, perhaps we can rephase the comment differently?

packages/puppeteer-core/src/bidi/Realm.ts
```diff
@@ -287,6 +288,21 @@ export abstract class BidiRealm extends Realm {
     await handle.dispose();
     return await transferredHandle;
   }
+
+  // TODO: BiDi implementation missing
+  override set worldId(_worldId: string | symbol) {
+    throw new Error('worldId are defined only for CDP implementations.');
```
OrKoN: let's use UnsupportedOperation here and in other places

packages/puppeteer-core/src/cdp/Frame.ts
```diff
@@ -69,10 +70,17 @@ export class CdpFrame extends Frame {
 
     this._loaderId = '';
     this.worlds = {
-      [MAIN_WORLD]: new IsolatedWorld(this, this._frameManager.timeoutSettings),
+      [MAIN_WORLD]: new IsolatedWorld(
+        this,
+        this._frameManager.timeoutSettings,
+        MAIN_WORLD,
+        this._frameManager.page().browser(),
```
OrKoN: do we need to pass the browser explicitly? can we get it from `this` that we pass as the first argument?

packages/puppeteer-core/src/cdp/FrameManager.ts
```diff
@@ -541,11 +536,30 @@ export class FrameManager extends EventEmitter<FrameManagerEvents> {
...
+    return origin.startsWith('chrome-extension://');
+  }
+
+  #extractExtensionId(origin: string): string | null {
+    if (!origin || !this.#isExtensionOrigin(origin)) {
+      return null;
+    }
+
+    // 19 is "chrome-extension://".length
+    const pathPart = origin.substring(19);
```
OrKoN:
```suggestion
    const pathPart = origin.substring('chrome-extension://'.length);
```

packages/puppeteer-core/src/cdp/FrameManager.ts
```diff
@@ -561,8 +575,30 @@ export class FrameManager extends EventEmitter<FrameManagerEvents> {
         // connections so we might end up creating multiple isolated worlds.
         // We can use either.
         world = frame.worlds[PUPPETEER_WORLD];
+      } else if (this.#isExtensionOrigin(origin)) {
+        const extId = this.#extractExtensionId(origin);
+
+        if (!extId) {
+          throw new Error('error while parsing extension id');
```
OrKoN: this is async code, let's use debugError and return early instead

packages/puppeteer-core/src/cdp/IsolatedWorld.ts
```diff
@@ -66,15 +69,21 @@ export type IsolatedWorldEmitter = EventEmitter<{
...
   constructor(
     frameOrWorker: CdpFrame | CdpWebWorker,
     timeoutSettings: TimeoutSettings,
+    worldId: string | symbol,
+    browser?: Browser,
   ) {
     super(timeoutSettings);
     this.#frameOrWorker = frameOrWorker;
+    this.#worldId = worldId;
+    this.#browser = browser;
```
OrKoN: can we get the browser from CdpFrame | CdpWebWorker? (related to my previous comment)

packages/puppeteer-core/src/cdp/WebWorker.ts
```diff
@@ -55,7 +55,14 @@ export class CdpWebWorker extends WebWorker {
     this.#id = targetId;
     this.#client = client;
     this.#targetType = targetType;
-    this.#world = new IsolatedWorld(this, new TimeoutSettings());
+    // TODO: Understand how to define world name for WebWorkers
```
OrKoN: technically workers do not have worlds. I think we can use the main world symbol for the id.

packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -1200,4 +1200,12 @@ export abstract class Frame extends EventEmitter<FrameEvents> {
...
     options?: WaitTimeoutOptions,
   ): Promise<DeviceRequestPrompt>;
+
+  /**
+   * This method retrieves the list of realms inside of a Frame returned
+   * as a pair-like [key, Object] list
+   *
+   * @public
+   */
+  abstract getRealms(): Array<[string | symbol, Realm]>;
```
OrKoN: maybe we should expose symbol keys in the public API as they are a bit harder to use, can we use `default` or `main` literals to identify the main realm

packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -15,30 +15,57 @@ import {disposeSymbol} from '../util/disposable.js';
...
-
+  /** @internal */
   abstract get environment(): Environment;
 
+  /**
+   * The identifier for this realm.
+   *
+   * @public
+   */
+  abstract get worldId(): string | symbol;
```
Lightning00Blade: This should not be public as we use it internally, we should just expose id in my opinion. Also the name world is very CDP specific. Maybe using something else?

packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -15,30 +15,57 @@ import {disposeSymbol} from '../util/disposable.js';
...
+   */
+  abstract extension(): Promise<Extension | null>;
+
+  /** @internal */
   abstract adoptHandle<T extends JSHandle<Node>>(handle: T): Promise<T>;
+
+  /** @internal */
   abstract transferHandle<T extends JSHandle<Node>>(handle: T): Promise<T>;
+
+  /** @public */
```
Lightning00Blade: If the object is public no need to be explicitly added.

packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -87,12 +115,15 @@ export abstract class Realm implements Disposable {
...
 
+  /** @internal */
   abstract adoptBackendNode(backendNodeId?: number): Promise<JSHandle<Node>>;
 
+  /** @internal */
   get disposed(): boolean {
     return this.#disposed;
   }
 
+  /** @internal */
```
Lightning00Blade: Not needed this is private property.

packages/puppeteer-core/src/api/Page.ts
nroscino: This is from the rebase. I think in previous branch it was first above and then I moved below, I will move this back down.

packages/puppeteer-core/src/api/Realm.ts
nroscino: Changed, what do you think about this?

packages/puppeteer-core/src/cdp/Frame.ts
nroscino: Done!

packages/puppeteer-core/src/cdp/FrameManager.ts
nroscino: Done!

packages/puppeteer-core/src/cdp/FrameManager.ts
nroscino: Done!

packages/puppeteer-core/src/cdp/IsolatedWorld.ts
nroscino: No, but we don't need the browser from CdpWebWorker, so when we need the browser in the extension() method, we can check if frameOrWorker is a worker and throw in that situation instead of checking the browser, and then instead of using this.#browser we use directly the CdpFrame to get the browser.

packages/puppeteer-core/src/cdp/WebWorker.ts
nroscino: Done!

packages/puppeteer-core/src/api/Frame.ts
nroscino: Done!

packages/puppeteer-core/src/cdp/Frame.ts
```diff
@@ -84,30 +92,23 @@ export class CdpFrame extends Frame {
...
 
-    this.worlds[MAIN_WORLD].emitter.on(
-      'consoleapicalled',
-      this.#onMainWorldConsoleApiCalled.bind(this),
-    );
-    this.worlds[MAIN_WORLD].emitter.on(
-      'bindingcalled',
-      this.#onMainWorldBindingCalled.bind(this),
-    );
+    this._registerWorldListeners(this.worlds[MAIN_WORLD]);
```
Lightning00Blade: Prefer not using `_` prefix as the marked `@internal` showcases that it's not public.

packages/puppeteer-core/src/api/Frame.ts
Lightning00Blade: Another question is why have the return type as such? What benefit is it to the end use.
As one can access the ID/name from the value?

packages/puppeteer-core/src/cdp/Frame.ts
```diff
@@ -435,6 +436,16 @@ export class CdpFrame extends Frame {
...
       .adoptBackendNode(backendNodeId)) as ElementHandle<HTMLIFrameElement>;
   }
+
+  /**
+   * @public
+   */
+  override getRealms(): Array<[string | symbol, Realm]> {
+    return [
+      [MAIN_WORLD, this.worlds[MAIN_WORLD]],
+      ...Object.entries(this.worlds),
```
Lightning00Blade: I think this exposes the Internal Puppeteer realm.

packages/puppeteer-core/src/cdp/FrameManager.ts
Lightning00Blade: Even better - extract `chrome-extension://` into a variable and re-use it in all the places it's mentioned.

test/src/cdp/realms.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+
+import {setupSeparateTestBrowserHooks} from '../mocha-utils.js';
+
+const EXTENSION_PATH = path.join(
+  import.meta.dirname,
+  '../../assets/simple-extension',
+);
+
+describe('realms', function () {
+  const state = setupSeparateTestBrowserHooks(
```
Lightning00Blade, line 20: Most test should use only the normal browser, and we should have separate test for the places that need extensions.

test/src/cdp/realms.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    const page = await browser.newPage();
+    const realms = page.realms();
+
+    expect(realms.length).toBeGreaterThanOrEqual(1);
+
+    const worldIds = realms.map(([id]) => {
+      return id;
+    });
+    try {
+      expect(worldIds).toContain(MAIN_WORLD);
```
Lightning00Blade: Always prefer `atLeastOneToContain`

test/src/cdp/realms.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+      // eslint-disable-next-line
+    } catch (e) {
```
Lightning00Blade: FYI you can just omit the error
```suggestion
    } catch {
```

test/src/cdp/realms.spec.ts
```diff
@@ -0,0 +1,148 @@
...
+    try {
+      const mainRealm = realms.find(([id]) => {
+        return id === MAIN_WORLD;
+      })?.[1];
+
+      expect(mainRealm?.worldId).toEqual(MAIN_WORLD);
+      // eslint-disable-next-line
+    } catch (e) {
+      const mainRealm = realms.find(([id]) => {
+        return id === 'default';
```
Lightning00Blade: The API should be the same for CDP/BiDi, I think this ties back to using Symbols or not
CC @OrKoN

test/TestExpectations.json
```diff
@@ -1655,6 +1648,24 @@
     "parameters": ["firefox", "webDriverBiDi"],
     "expectations": ["PASS"]
   },
+  {
+    "testIdPattern": "[realms.spec] realms extension *",
+    "platforms": ["linux"],
```
Lightning00Blade: This need all platforms

test/TestExpectations.json
```diff
@@ -1955,6 +1966,36 @@
     "expectations": ["SKIP"],
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
+  {
+    "testIdPattern": "[realms.spec] realms extension realm should return extension that created it",
+    "platforms": ["darwin"],
+    "parameters": ["cdp", "chrome", "chrome-headless-shell"],
+    "expectations": ["FAIL"]
```
Lightning00Blade: All places need `comment` explaining why they fail.

test/src/cdp/realms.spec.ts
nroscino: This is now the same

packages/puppeteer-core/src/api/Frame.ts
nroscino: If we expose getter and setter we need to keep the symbol because for puppeteer worldId will still be a symbol. If we expose the id like this we can narrow it to just a string cause for public api it will be just a string.

packages/puppeteer-core/src/bidi/Frame.ts
```diff
@@ -39,6 +40,7 @@ import {
   fromEmitterEvent,
   timeout,
 } from '../common/util.js';
+import {MAIN_WORLD} from '../index-browser.js';
```
OrKoN: we should not import from entrypoints internally

docs/api/puppeteer.main_world.md
```diff
@@ -0,0 +1,13 @@
...
+---
+
+# MAIN_WORLD variable
+
+A unique key for to denote the default world. Execution contexts are automatically created in the default world.
+
+### Signature
+
+```typescript
+MAIN_WORLD = 'default';
```
OrKoN: instead of exposing the constant let's just use `default` as a literal in types?

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -3258,6 +3252,11 @@ export abstract class Page extends EventEmitter<PageEvents> {
    * @public
    */
   abstract triggerExtensionAction(extension: Extension): Promise<void>;
+
+  /**
+   * @public
+   */
+  abstract realms(): Array<[string, Realm]>;
```
OrKoN: actually it got me thinking. Do we actually need to expose the default realm? The user can already access it via the page? Perhaps we should just return additional realms such as extension realms here? WDYT? @nroscino @Lightning00Blade

test/src/cdp/realms.spec.ts
nroscino: I tried but the Symbol was making it complain, and didn't want to change the signature, it felt a bit wrong. With the current change we don't need this anymore tho. Now the main world has the same "default" string key!

packages/puppeteer-core/src/api/Frame.ts
Lightning00Blade: One can simple provide the getter to return string and handle the case for symbols diffrently.

packages/puppeteer-core/src/api/Page.ts
Lightning00Blade: SGTM

nroscino: it makes sense actually. We can rename it to `extensionRealms()` and return only the extensions realm. This applies also to the frame. What do you guys think?

OrKoN: Let's go with extensionRealms(), sg!

packages/puppeteer-core/src/cdp/Frame.ts
nroscino: The `Symbol` keys are not returned by the method `Object.entries` or `Object.values`. The doc specifies that it returns only the string keys related values.

packages/puppeteer-core/src/api/Realm.ts
nroscino: I changed it to `realmId`, made a setter that it's internal! Let me know if now it's ok for you!

packages/puppeteer-core/src/cdp/Frame.ts
Lightning00Blade: Aaa yes Symbol keys are not `enumerable`, forgot about that.

packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -1200,4 +1200,12 @@ export abstract class Frame extends EventEmitter<FrameEvents> {
   abstract waitForDevicePrompt(
     options?: WaitTimeoutOptions,
   ): Promise<DeviceRequestPrompt>;
+
+  /**
+   * This method retrieves the list of realms inside of a Frame returned
+   * as a pair-like [key, Object] list
```
OrKoN: The comment seems to be no longer up to date.

packages/puppeteer-core/src/bidi/Realm.ts
```diff
@@ -287,6 +289,19 @@ export abstract class BidiRealm extends Realm {
     await handle.dispose();
     return await transferredHandle;
   }
+
+  // TODO: BiDi implementation missing
```
Lightning00Blade: Nit: remove comment the Error type showcases this.

Lightning00Blade: We can update the doc at https://pptr.dev/webdriver-bidi

packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -15,30 +15,59 @@ import {disposeSymbol} from '../util/disposable.js';
...
+   * The identifier for this realm.
+   *
+   * @public
+   */
+  abstract get realmId(): string;
+
+  /**
+   * @internal
+   */
+  abstract setRealmId(worldId: string | symbol): void;
```
OrKoN: I think this is still a worldId here. I think we should keep it named setWorldId

packages/puppeteer-core/src/cdp/Frame.ts
```diff
@@ -435,6 +434,13 @@ export class CdpFrame extends Frame {
       .mainRealm()
       .adoptBackendNode(backendNodeId)) as ElementHandle<HTMLIFrameElement>;
   }
+
+  /**
+   * @public
+   */
+  override extensionRealms(): Realm[] {
+    return Object.values(this.worlds);
```
Lightning00Blade: Don't we need to filter this as there may be other realms created for stuff example if DevTool is attached, Recorder panel may create one.

packages/puppeteer-core/src/api/Realm.ts
OrKoN: does it need to be on the public interface? It seems to be only used in CDP part?

packages/puppeteer-core/src/cdp/Frame.ts
Lightning00Blade: I saw we filter it when we create the realms, let just add separate map where we store this, to avoid confusion in the future. WDYT?

packages/puppeteer-core/src/cdp/FrameManager.ts
```diff
@@ -561,8 +579,26 @@ export class FrameManager extends EventEmitter<FrameManagerEvents> {
...
+
+        if (frame.worlds[extId]) {
+          world = frame.worlds[extId];
+        } else {
+          world = new IsolatedWorld(frame, this.timeoutSettings, extId);
+          frame.worlds[extId] = world;
+          frame.registerWorldListeners(world);
+        }
+
+        world.setRealmId(extId);
```
Lightning00Blade: If we get in the first branch `frame.worlds[extId]` would it not already be the correct one?

packages/puppeteer-core/src/cdp/IsolatedWorld.ts
```diff
@@ -263,4 +269,35 @@ export class IsolatedWorld extends Realm {
...
     this.#emitter.removeAllListeners();
   }
+
+  override setRealmId(worldId: string | symbol): void {
+    this.#worldId = worldId;
+  }
+
+  override get realmId(): string {
+    if (typeof this.#worldId === 'symbol') {
+      throw new UnsupportedOperation();
```
Lightning00Blade: This should be an Error with message like `Trying to access internal realms` or something like that.

packages/puppeteer-core/src/cdp/IsolatedWorld.ts
```diff
@@ -263,4 +269,35 @@ export class IsolatedWorld extends Realm {
...
+    }
+    return this.#worldId;
+  }
+
+  async extension(): Promise<Extension | null> {
+    if (this.#frameOrWorker instanceof CdpWebWorker) {
+      throw new Error('unable to get extension from Realm');
```
Lightning00Blade:
```suggestion
      throw new Error('Unable to get extension from Realm');
```

packages/puppeteer-core/src/cdp/IsolatedWorld.ts
```diff
@@ -263,4 +269,35 @@ export class IsolatedWorld extends Realm {
...
+    if (this.#worldId === MAIN_WORLD) {
+      return await Promise.resolve(null);
+    }
+
+    if (typeof this.realmId === 'string') {
+      const extensions = await this.#frameOrWorker._frameManager
+        .page()
+        .browser()
+        .extensions();
+      return extensions.get(this.realmId) || null;
```
Lightning00Blade: Nit:
```suggestion
      return extensions.get(this.realmId) ?? null;
```

packages/puppeteer-core/src/cdp/IsolatedWorld.ts
```diff
@@ -263,4 +269,35 @@ export class IsolatedWorld extends Realm {
...
+    return this.#worldId;
+  }
+
+  async extension(): Promise<Extension | null> {
+    if (this.#frameOrWorker instanceof CdpWebWorker) {
+      throw new Error('unable to get extension from Realm');
+    }
+
+    if (this.#worldId === MAIN_WORLD) {
+      return await Promise.resolve(null);
```
Lightning00Blade: When you are inside an async function no need to wrap this way.
```suggestion
      return null;
```

packages/puppeteer-core/src/cdp/IsolatedWorld.ts
```diff
@@ -263,4 +269,35 @@ export class IsolatedWorld extends Realm {
...
+
+    if (typeof this.realmId === 'string') {
+      const extensions = await this.#frameOrWorker._frameManager
+        .page()
+        .browser()
+        .extensions();
+      return extensions.get(this.realmId) || null;
+    }
+
+    return await Promise.resolve(null);
```
Lightning00Blade: No wrap.

packages/puppeteer-core/src/cdp/IsolatedWorld.ts
```diff
@@ -19,14 +20,16 @@ import {
   timeout,
   withSourcePuppeteerURLIfNone,
 } from '../common/util.js';
+import {UnsupportedOperation} from '../index-browser.js';
```
Lightning00Blade: You should use imports from files.

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -633,6 +634,10 @@ export class CdpPage extends Page {
     );
   }
 
+  override async triggerExtensionAction(extension: Extension): Promise<void> {
```
Lightning00Blade: Revert re-order.

test/src/cdp/realms.spec.ts
Lightning00Blade: I don't see this as being resolved or comment left for why this is not adressed.

test/TestExpectations.json
```diff
@@ -181,6 +181,27 @@
...
   },
+  {
+    "testIdPattern": "[realms.spec] *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["webDriverBiDi"],
+    "expectations": ["SKIP"],
+    "comment": "No extensions in WebDriver BiDi so far"
+  },
+  {
+    "testIdPattern": "[realms.spec] *",
```
Lightning00Blade: You can remove this as its covered by the bidi above.

packages/puppeteer-core/src/api/Realm.ts
nroscino: as of now probably not, I expected it to be needed later on the the BiDi implementation but I think we can move it later if needed! I can move it to the internal!

packages/puppeteer-core/src/cdp/FrameManager.ts
nroscino: Yeah, you are right. I'll move this into the `else` block.

test/src/cdp/realms.spec.ts
nroscino: Before we were also exposing the `MAIN_WORLD`, so, as you said some test could have a normal browser without extension, but since we are now working only with extensions realm I thought this was resolved. Am I missing something here?

Lightning00Blade: Aaa yes, I got myself confused.

packages/puppeteer-core/src/cdp/Frame.ts
nroscino: I created a separated map for `extensionWorlds`, let me know if that's how you imagined it. I also disposed the new realms when the frame is disposed. Am I missing something?

packages/puppeteer-core/src/api/Frame.ts
```diff
@@ -1200,4 +1200,11 @@ export abstract class Frame extends EventEmitter<FrameEvents> {
   abstract waitForDevicePrompt(
     options?: WaitTimeoutOptions,
   ): Promise<DeviceRequestPrompt>;
+
+  /**
+   * This method retrieves the list of realms inside of a Frame
```
OrKoN:
```suggestion
   * This method retrieves the list of realms within this frame
```

OrKoN (approved):
Thanks, LGTM!

packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -15,30 +15,54 @@ import {disposeSymbol} from '../util/disposable.js';
 
 import type {ElementHandle} from './ElementHandle.js';
 import type {Environment} from './Environment.js';
+import type {Extension} from './Extension.js';
 import type {JSHandle} from './JSHandle.js';
 
 /**
- * @internal
+ * @public
```
Lightning00Blade, line 18: Nit: We can add some info on what a Realm can do, we can do in a follow up.

packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -15,30 +15,54 @@ import {disposeSymbol} from '../util/disposable.js';
...
-
+  /** @internal */
   abstract get environment(): Environment;
 
+  /**
+   * The identifier for this realm.
+   *
+   * @public
+   */
+  abstract get realmId(): string;
```
Lightning00Blade: Should this just be Id? The API now is strange as it duplicate wording.
```ts
...
realm.realmId()
```

packages/puppeteer-core/src/bidi/Realm.ts
```diff
@@ -287,6 +289,14 @@ export abstract class BidiRealm extends Realm {
     await handle.dispose();
     return await transferredHandle;
   }
+
+  override get realmId(): string {
+    throw new UnsupportedOperation();
```
Lightning00Blade, line 294: @OrKoN would this be the Sandbox name?

packages/puppeteer-core/src/cdp/Frame.ts
```diff
@@ -68,11 +70,17 @@ export class CdpFrame extends Frame {
     this.#client = client;
 
     this._loaderId = '';
+    this.extensionWorlds = {};
```
Lightning00Blade: Put it in the declaration directly.

packages/puppeteer-core/src/cdp/FrameManager.ts
```diff
@@ -235,7 +231,6 @@ export class FrameManager extends EventEmitter<FrameManagerEvents> {
         ...(frame ? Array.from(this.#bindings.values()) : []).map(binding => {
           return frame?.addExposedFunctionBinding(binding);
         }),
-        this.#page.browser().isIssuesEnabled() && client.send('Audits.enable'),
```
Lightning00Blade, line 238: Revert this.

Lightning00Blade (commented):
LGTM

Lightning00Blade (approved):

packages/puppeteer-core/src/bidi/Realm.ts
OrKoN: no, I think it should be realm id. Actually the implementation should be this `this.realm.id`. I agree with the other comment actually this should be just the id.

packages/puppeteer-core/src/api/Realm.ts
OrKoN: actually, thinking more about this question. Why does it need to be public?

packages/puppeteer-core/src/bidi/Realm.ts
OrKoN: can we remove the realmId getter altogether?

packages/puppeteer-core/src/api/Realm.ts
nroscino: Right now we can get a list of realms inside a frame, each realm is associated with an extension using the same id. If I have a Frame with 10 different extensions content scripts, how can I get the one I want to run a script on it?

OrKoN: You can get the extension object and check its id?

nroscino: Mmm ok, that would work but it will make a call to the backend for each extension that we might not care about. Is this ok?

OrKoN: yeah I think it is fine. But if we just want to identify extensions let's add extensionId() getter as the extension ID ? I think it is fine to just leave it out for now?

OrKoN: for true realmId() we should probably use the execution context id

packages/puppeteer-core/src/api/Realm.ts
nroscino: Ok I will do a follow up with maybe also an example. I also added it as experimental for the same reason as Extension, or you think it's not necessary?

packages/puppeteer-core/src/api/Realm.ts
nroscino: You mean the "id" in the following screenshot?
image

OrKoN: yes

packages/puppeteer-core/src/api/Realm.ts
Lightning00Blade: We can add experimental on the API that are releated to the Realm but other thing should be fine as they are tested though the page APIs already.

packages/puppeteer-core/src/api/Realm.ts
Lightning00Blade: Don't we prefer the uniqueId these days?

OrKoN: we use both I think. I do not think we need to expose the id though. Let's expose origin() or extensionId() based on it

nroscino:
Hey @OrKoN @Lightning00Blade, Can you do a last review here, I did some changes. I also added origin to the realm. LMKWYT. 
For the failing extension tests I am dealing with it in a sepatate PR so before merging this one I will solve that.

packages/puppeteer-core/src/api/Realm.ts
```diff
@@ -49,9 +69,7 @@ export abstract class Realm implements Disposable {
 
   async waitForFunction<
     Params extends unknown[],
-    Func extends EvaluateFunc<InnerLazyParams<Params>> = EvaluateFunc<
-      InnerLazyParams<Params>
-    >,
+    Func extends EvaluateFunc<Params> = EvaluateFunc<Params>,
```
Lightning00Blade, line 72: Is the type used anywhere else? If not let's delete it.

packages/puppeteer-core/src/bidi/Realm.ts
```diff
@@ -48,7 +50,7 @@ import type {BidiWebWorker} from './WebWorker.js';
 /**
  * @internal
  */
-export abstract class BidiRealm extends Realm {
+export abstract class BidiRealm extends Realm implements Disposable {
```
Lightning00Blade: Would it make more sense for the Realm to implement the Disposable? As we define the methods there?

packages/puppeteer-core/src/cdp/IsolatedWorld.ts
```diff
@@ -63,18 +65,22 @@ export type IsolatedWorldEmitter = EventEmitter<{
 /**
  * @internal
  */
-export class IsolatedWorld extends Realm {
+export class IsolatedWorld extends Realm implements Disposable {
```
Lightning00Blade: Same question as the above.

Lightning00Blade (approved):

packages/puppeteer-core/src/bidi/Realm.ts
OrKoN: Disposable is not public so we cannot use it. But I wonder if we can just drop `implements`?

nroscino: I am ok in removing the implements, but why do we want to remove it? 
I was also checking other parts of the code that uses it and it's only one (besides the Realms one) on the `ExecutionContext`.

packages/puppeteer-core/src/cdp/IsolatedWorld.ts
nroscino: Continuing conversation above

packages/puppeteer-core/src/bidi/Realm.ts
OrKoN: We can still implement interfaces without declaring that we do. This would allow us to avoid the non-public type in the type exports.

OrKoN (approved):

OrKoN:
Let's rebase after the flakiness fix https://github.com/puppeteer/puppeteer/pull/14853 lands to rule out new flakiness in this PR.

PR #14835, opened 2026-04-07
feat(webmcp): add hook for tool invocation

beaufortfrancois:

This PR updates the WebMCP API.

Yes

Yes

**Summary**

It allows developers to know when a tool invocation starts by listening to the new `toolInvoked` event.

None that I'm aware of.

**Other information**

This is highly experimental.

cc @OrKoN


packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -119,6 +126,16 @@ export interface WebMCPToolsRemovedEvent {
   tools: WebMCPTool[];
 }
 
+/**
+ * @public
+ */
+export interface WebMCPToolCall {
```
OrKoN: I am thinking maybe the following interface would be nicer?

```
class WebMCPToolCall {
  id: string; // call ID aka invocationId
  tool(): WebMCPTool; // I think tool().frame() should always match the frameId on the event.
  input: object;
}
```

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -167,6 +187,20 @@ export class WebMCP extends EventEmitter<{
...
+    if (!frame) {
+      return;
+    }
+    const call: WebMCPToolCall = {
+      toolName: event.toolName,
+      frame: frame,
+      invocationId: event.invocationId,
+      input: event.input,
+    };
+    this.emit('toolinvoked', call);
```
OrKoN, line 213: Let's also emit the event on the corresponding tool instance turning WebMCPTool into an event emitter?

beaufortfrancois: Done

packages/puppeteer-core/src/cdp/WebMCP.ts
beaufortfrancois: > `class WebMCPToolCall {`
- Done

> `  id: string; // call ID aka invocationId`
- Done

> `  tool(): WebMCPTool; // I think tool().frame() should always match the frameId on the event.`
- I did use `tool` attribute since I don't emit the event if tool doesn't exist at the time of the toolinvoked event.

> `  input: object;` 
- Done but I do wonder if it should an empty object OR something else when JSON.parse fails.

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -167,6 +203,16 @@ export class WebMCP extends EventEmitter<{
     this.emit('toolsremoved', {tools});
   };
 
+  #onToolInvoked = (event: ProtocolWebMCPToolInvokedEvent) => {
+    const tool = this.#tools.get(event.frameId)?.get(event.toolName);
+    if (!tool) {
+      return;
+    }
+    tool.emit('toolinvoked', tool);
```
```suggestion
    tool.emit('toolinvoked', call);
```

I meant that we should emit the call on the tool itself.

beaufortfrancois (commented):
FYI Here's what I have in mind for toolresponded: https://github.com/beaufortfrancois/puppeteer/compare/toolinvoked...beaufortfrancois:puppeteer:toolresponded?expand=1

packages/puppeteer-core/src/cdp/WebMCP.ts

packages/puppeteer-core/src/cdp/WebMCP.ts
beaufortfrancois: FYI since we have tool instead of toolName, it means `toolinvoked` is not fired when executeTool is called with an invalid tool name for instance, or invalid input arguments. The only thing we'll have is the upcoming `toolresponded` event with the errorText attribute filled with either "Tool not found: unknown-tool-name" or "Failed to parse input arguments" for instance.

test/src/cdp/webmcp.spec.ts
```diff
@@ -267,4 +270,76 @@ describe('Page.webmcp', function () {
     expect(removedTools[0]!.name).toBe('declarative tool name');
     expect(page.webmcp.tools().length).toBe(0);
   });
+
+  it.only('should fire toolinvoked events', async () => {
```
OrKoN:
```suggestion
  it('should fire toolinvoked events', async () => {
```

beaufortfrancois: Oops! Sorry

packages/puppeteer-core/src/cdp/WebMCP.ts
OrKoN: Would CDP emit an event if the tool name is invalid? I would have expected that the caller gets an error only. I am thinking we can add a waiter helper on the WebMCPToolCall like `response(): Promise<WebMCPToolResponse|Error>` to automatically get the corresponding response or an error (in additional to just emitting the response as an event). On the WebMCPToolCallResponse we can also have a getter to `.request()` to get the WebMCPToolCall request.

OrKoN (approved):
LGMT thanks!

packages/puppeteer-core/src/cdp/WebMCP.ts
OrKoN: Although I guess we can skip the `response()` getter for now and and just rely on events. We could emit the response event on the matching WebMCPToolCall as well as on the webmcp interface.

beaufortfrancois: CDP emits the `toolinvoked` event with its `toolName` attribute set to `"unknown-tool-name"` yes. In Puppeteer, we don't emit `WebMCPToolCall` event since `toolName` is not in `this.#tools`.
Let's think about `WebMCPToolCall.response()` and `WebMCPToolCallResponse.request()` in the upcoming PR for `toolresponded` then.

PR #14848, opened 2026-04-08
feat(webmcp): expose WebMCPToolCall in WebMCPToolCallResult

beaufortfrancois:

This PR updates the WebMCP API.

Yes

Yes

**Summary**

It allows developers to get the WebMCPToolCall directly from a WebMCPToolResponse as requested in https://github.com/puppeteer/puppeteer/pull/14835#discussion_r3050029565

None that I'm aware of.

**Other information**

This is highly experimental.

cc @OrKoN

### DIFF
--- a/packages/puppeteer-core/src/cdp/WebMCP.ts
+++ b/packages/puppeteer-core/src/cdp/WebMCP.ts
@@ -170,7 +170,7 @@ export class WebMCPToolCall {
 /**
  * @public
  */
-export interface WebMCPToolResponse {
+export interface WebMCPToolCallResult {
   call?: WebMCPToolCall;
   status: WebMCPInvocationStatus;
   output?: any;
@@ -192,7 +192,7 @@ export class WebMCP extends EventEmitter<{
   /** Emitted when a tool invocation starts. */
   toolinvoked: WebMCPToolCall;
   /** Emitted when a tool invocation completes or fails. */
-  toolresponded: WebMCPToolResponse;
+  toolresponded: WebMCPToolCallResult;
 }> {
   #client: CDPSession;
   #frameManager: FrameManager;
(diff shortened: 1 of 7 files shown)


packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -199,6 +197,7 @@ export class WebMCP extends EventEmitter<{
   #client: CDPSession;
   #frameManager: FrameManager;
   #tools: Map<string, Map<string, WebMCPTool>>;
+  #pendingCalls: Map<string, WebMCPToolCall>;
```
```suggestion
  #pendingCalls = new Map<string, WebMCPToolCall>();
```
this should be possible to avoid duplicating the type (also for the other map).

OrKoN: so we cannot clean this map on navigation as declarative tools might have responses and at the same time we do not know if the event will be emitted for imperative tools if the frame navigates? Let's add a test for tool calls with navigation to make sure we do introduce a memory leak here?

OrKoN (approved):

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -278,7 +286,6 @@ export class WebMCP extends EventEmitter<{
       this.#onFrameNavigated,
```
OrKoN, line 287: let's clear pending calls in onFrameNavigated

packages/puppeteer-core/src/cdp/WebMCP.ts
OrKoN: discussed offline: that for now we will bring call id back and clear the pending calls on navigation.

packages/puppeteer-core/src/cdp/WebMCP.ts
beaufortfrancois: Done

beaufortfrancois:
@OrKoN I think we're good to merge now.

OrKoN (approved):

PR #14853, opened 2026-04-09
test: fix flaky extension test

nroscino:
Fixes flaky extension tests failing. 

No

**Summary**

The chromium backend doesn't correctly send events to remove service workers when an extension is uninstalled. This lead to the extension service worker targets not being removed correctly. This PR emulates the event to enable clearing of those targets.

**Other information**

### DIFF
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -462,24 +462,22 @@ export class CdpBrowser extends BrowserBase {
     // flakiness in the extension tests.
     // TODO(nroscino): Remove this once the event is correctly emitted.
     const targets = this.targets().filter(target => {
-      return target.url().includes(id);
+      return target.url().includes(id) && target.type() === 'service_worker';
     });
 
     const targetDestroyedPromises = [];
     for (const target of targets) {
-      if (target.type() === 'service_worker') {
-        this._targetManager().addToIgnoreTarget(target._targetId);
-        targetDestroyedPromises.push(
-          new Promise(resolve => {
-            return setTimeout(() => {
-              this.#connection.emit('Target.targetDestroyed', {
-                targetId: target._targetId,
-              });
-              resolve(null);
-            }, 0);
-          }),
-        );
-      }
+      this._targetManager().addToIgnoreTarget(target._targetId);
+      targetDestroyedPromises.push(
+        new Promise(resolve => {
+          return setTimeout(() => {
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -456,6 +456,33 @@ export class CdpBrowser extends BrowserBase {
...
+    // the Target.targetDestroyed event for service workers. This causes
+    // flakiness in the extension tests.
+    // TODO(nroscino): Remove this once the event is correctly emitted.
+    const targets = this.targets().filter(target => {
+      return target.url().includes(id);
+    });
+
+    const targetDestroyedPromises = [];
+    for (const target of targets) {
+      if (target.type() === 'service_worker') {
```
OrKoN: let's move this if into the target filter on line 465?

test/src/cdp/extensions.spec.ts
```diff
@@ -31,22 +33,26 @@ describe('extensions', function () {
...
   it('service_worker target type should be available', async function () {
     const {browser} = state;
+    const extensionId = await browser.installExtension(extensionPath);
     const serviceWorkerTarget = await browser.waitForTarget(target => {
       return target.type() === 'service_worker';
     });
     expect(serviceWorkerTarget).toBeTruthy();
+    await browser.uninstallExtension(extensionId);
+    const targets = browser.targets();
+    ensureNoWorkerWithIdFound(targets, extensionId);
```
OrKoN:
```suggestion
    assertNoServiceWorkerReported(targets, extensionId);
```

perhaps?

test/src/cdp/extensions.spec.ts
```diff
@@ -211,17 +262,35 @@ describe('extensions', function () {
...
+    const targets = browser.targets();
+    ensureNoWorkerWithIdFound(targets, id);
 
     extensions = await browser.extensions();
     expect(extensions.has(id)).toBe(false);
   });
 });
+
+function ensureNoWorkerWithIdFound(targets: Target[], id: string) {
+  const target = targets.find(curr => {
```
OrKoN:
```suggestion
  const target = targets.find(target => {
```

packages/puppeteer-core/src/cdp/Browser.ts
nroscino: Fair! Done!

OrKoN (approved):

PR #14860, opened 2026-04-13
fix: make `Target.asPage` return the same Page instance

### LINKED ISSUE #14843: [Bug]: Target.asPage returns different instance each call

```TypeScript
  try {

    const target = page.target();

    const page1 = await target.asPage();
    const page2 = await target.asPage();

    console.log('page1 === page2:', page1 === page2);
    // expected: true, actual: false
  } finally {
    await browser.close();
  }
```

### Background

_No response_

### Expectation

the `.asPage()` method always return the same `Page` instance for a specific `Target`.

### Reality

the `.asPage()` method returns 2 different `Page` instances for the same `Target`.

```TypeScript

```

### Puppeteer version

latest

### Node version

v24.14.0

### Package manager

npm

### Package manager version

11.9.0

### Operating system

macOS

Lightning00Blade:
Fixes #14843

**💡 What**
- Updated `CdpTarget` to store a `_pagePromise` so that both `page()` and `asPage()` share and return the same `Page` instance.
- Updated `BidiPageTarget` and `BidiFrameTarget` to return their existing/cached `this.#page` directly.
- Removed the `WeakMap` workaround from `CdpExtension` since `asPage()` now properly caches the instance at the Target level.
- Added tests in `target.spec.ts` and `devtools.spec.ts` to assert that multiple calls to `asPage()` return the exact same instance.

**🎯 Why**
Fixes an issue where `Target.asPage` would return a different instance each call, breaking references and leading to unexpected behavior and possible memory leaks (workarounds like using `WeakMap` in extensions were previously needed).

---
*PR created automatically by Jules for task [14586855382459502714](https://jules.google.com/task/14586855382459502714) started by @Lightning00Blade*

### DIFF
--- a/packages/puppeteer-core/src/bidi/Target.ts
+++ b/packages/puppeteer-core/src/bidi/Target.ts
@@ -63,7 +63,7 @@ export class BidiPageTarget extends Target {
     return this.#page;
   }
   override async asPage(): Promise<BidiPage> {
-    return this.#page;
+    return this.page();
   }
   override url(): string {
     return this.#page.url();
@@ -107,13 +107,7 @@ export class BidiFrameTarget extends Target {
     return this.#page;
   }
   override async asPage(): Promise<BidiPage> {
-    if (this.#page === undefined) {
-      this.#page = BidiPage.from(
-        this.browserContext(),
-        this.#frame.browsingContext,
-      );
-    }
-    return this.#page;
+    return this.page();
   }
   override url(): string {
     return this.#frame.url();
--- a/packages/puppeteer-core/src/bidi/Target.ts
+++ b/packages/puppeteer-core/src/bidi/Target.ts
@@ -63,10 +63,7 @@ export class BidiPageTarget extends Target {
     return this.#page;
   }
   override async asPage(): Promise<BidiPage> {
-    return BidiPage.from(
-      this.browserContext(),
-      this.#page.mainFrame().browsingContext,
-    );
+    return this.#page;
   }
   override url(): string {
     return this.#page.url();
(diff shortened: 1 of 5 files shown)


packages/puppeteer-core/src/cdp/Target.ts
```diff
@@ -206,7 +211,6 @@ export class CdpTarget extends Target {
  */
 export class PageTarget extends CdpTarget {
   #defaultViewport?: Viewport;
-  protected pagePromise?: Promise<Page>;
```
Lightning00Blade, line 209: There should be two separate promises for the `page` and the `asPage` API 
But the `asPage` should check if the `page` promise is there and re-use it if so.

packages/puppeteer-core/src/cdp/Target.ts
```diff
@@ -70,13 +73,20 @@ export class CdpTarget extends Target {
   }
 
   override async asPage(): Promise<Page> {
-    const session = this._session();
-    if (!session) {
-      return await this.createCDPSession().then(client => {
+    if (this.pagePromise) {
```
Lightning00Blade, line 76: This is not correct if the promise exits you need to await it and if it resolves to not null only then resolve. 
If it's null you should proceed to create the asPage promise.

packages/puppeteer-core/src/bidi/Target.ts
```diff
@@ -63,10 +63,7 @@ export class BidiPageTarget extends Target {
     return this.#page;
   }
   override async asPage(): Promise<BidiPage> {
-    return BidiPage.from(
-      this.browserContext(),
-      this.#page.mainFrame().browsingContext,
-    );
+    return this.#page;
```
Lightning00Blade: Something is the BiDi implementation is wrong. Run `npm run test:chrome:bidi` to find the issue and fix

Lightning00Blade:
Please always run `npm run format` before committing code.

packages/puppeteer-core/src/cdp/Extension.ts
```diff
@@ -57,16 +56,7 @@ export class CdpExtension extends Extension {
...
-      if (!page) {
-        // TODO: asPage should return always the same instance
-        // issue: https://github.com/puppeteer/puppeteer/issues/14843
-        page = await target.asPage();
-        if (page) {
-          this.#pages.set(target, page);
-        }
-      }
-
+      const page = await target.asPage();
```
Lightning00Blade: Can we re-write this so we use Promise.all so we can do things in parallel?

PR #14862, opened 2026-04-14
fix: make `Target.asPage` return the same Page instance

### LINKED ISSUE #14843: [Bug]: Target.asPage returns different instance each call

Lightning00Blade:
Closes https://github.com/puppeteer/puppeteer/pull/14860
Fixes https://github.com/puppeteer/puppeteer/issues/14843

### DIFF
--- a/packages/puppeteer-core/src/bidi/BrowserContext.ts
+++ b/packages/puppeteer-core/src/bidi/BrowserContext.ts
@@ -187,6 +187,13 @@ export class BidiBrowserContext extends BrowserContext {
     });
   }
 
+  /**
+   * @internal
+   */
+  getTargetForPage(page: BidiPage): BidiPageTarget | undefined {
+    return this.#targets.get(page)?.[0];
+  }
+
   override async newPage(options?: CreatePageOptions): Promise<Page> {
     using _guard = await this.waitForScreenshotOperations();
 
--- a/packages/puppeteer-core/src/bidi/Page.ts
+++ b/packages/puppeteer-core/src/bidi/Page.ts
@@ -31,6 +31,7 @@ import {
   type NewDocumentScriptEvaluation,
   type ScreenshotOptions,
 } from '../api/Page.js';
+import type {Target} from '../api/Target.js';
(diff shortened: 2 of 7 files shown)


packages/puppeteer-core/src/bidi/Page.ts
```diff
@@ -671,8 +672,12 @@ export class BidiPage extends Page {
     throw new UnsupportedOperation();
   }
 
-  override target(): never {
-    throw new UnsupportedOperation();
+  override target(): Target {
+    const target = this.browserContext().getTargetForPage(this);
```
Lightning00Blade, line 676: I am not sure why we didn't expose this previously, let me know if I should revert this part and add test expectations for this not working with BiDi

packages/puppeteer-core/src/bidi/BrowserContext.ts
```diff
@@ -187,6 +187,13 @@ export class BidiBrowserContext extends BrowserContext {
     });
   }
```
abiiget765: [CI](https://github.com/puppeteer/puppeteer/actions/workflows/ci.yml)

packages/puppeteer-core/src/bidi/Page.ts
OrKoN: I think we though about deprecating and removing the target API altogether. But I am not sure it is going to happen any time soon.

OrKoN (approved):

PR #14863, opened 2026-04-15
feat(browsers): add unpacking indicator to install command

### LINKED ISSUE #13083: [Feature]: Add unpacking indicator to the install command

MukundaKatta:
## Summary

After a successful browser download the CLI sits silently through the archive-extract step, which can take many seconds on large Chrome builds and made the process look hung. This PR mirrors the existing download-progress pattern for unpacking.

## API

New optional field on `InstallOptions`:

```ts
unpackProgressCallback?:
  | 'default'
  | ((event: 'started' | 'finished') => void);
```

Fires `'started'` immediately before `unpackArchive` and `'finished'` immediately after it succeeds.

- `'default'` (used by the bundled CLI) writes a short status line to **stderr**:
  `Unpacking archive... done.` — overwritten in place on TTYs so output stays compact, terminated with a newline on pipes.
- Custom callback functions are supported for programmatic use.
- `makeUnpackProgressCallback` is exported from `@puppeteer/browsers` so library consumers can reuse the default formatter.

## Backwards compatibility

The CLI opts in with `unpackProgressCallback: 'default'`. Consumers calling `install()` directly see no behavior change until they opt in.

## Test plan

- [x] `tsc --noEmit` on both the library and test projects — clean (pre-existing errors in `cli.spec.ts` and `utils.ts` are unrelated)
- [x] Added `packages/browsers/test/src/unpackProgressCallback.spec.ts` — stubs `process.stderr.write` and asserts the status line + completion marker
- [ ] I didn't get mocha to run locally against the built tests (the package mixes ESM-built tests with a CJS runtime and it tripped a "Cannot use import" error on my box). CI is the authoritative signal here.

Closes #13083

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -547,6 +547,7 @@ export class CLI {
       platform: args.platform,
       cacheDir: args.path ?? this.#cachePath,
       downloadProgressCallback: 'default',
+      unpackProgressCallback: 'default',
       baseUrl: args.baseUrl,
       buildIdAlias:
         originalBuildId !== args.browser.buildId ? originalBuildId : undefined,
--- a/packages/browsers/src/install.ts
+++ b/packages/browsers/src/install.ts
@@ -84,6 +84,16 @@ export interface InstallOptions {
     | 'default'
     | ((downloadedBytes: number, totalBytes: number) => void);
   /**
+   * Provides information about the progress of the archive-unpacking step,
+   * which can run for many seconds after the download completes. Called
+   * with `'started'` before extraction begins and `'finished'` after it
+   * succeeds. If set to `'default'`, a short status line is written to
+   * stderr.
+   */
+  unpackProgressCallback?:
+    | 'default'
+    | ((event: 'started' | 'finished') => void);
+  /**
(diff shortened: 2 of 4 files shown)


Qodo-Free-For-OSS:
Hi, A consumer-provided unpackProgressCallback that throws will abort installUrl() before extraction (on 'started') or after extraction (on 'finished'), turning a non-critical progress hook into a hard install failure.

**Severity:** action required | **Category:** reliability

**How to fix:** Isolate callback errors

> ## Issue description
> `unpackProgressCallback` is a progress/UI hook but it is called without error isolation. If a user callback throws, it aborts installation.
> 
> ## Issue Context
> In `installUrl`, the callback is called both before and after `unpackArchive`. Any thrown error will bubble and cause `install()` to reject.
> 
> ## Fix Focus Areas
> - packages/browsers/src/install.ts[464-472]
> 
> ## Suggested fix
> - Wrap each callback invocation in a `try/catch`.
> - Decide on behavior when callback fails (recommended: swallow callback errors and optionally `debugInstall(...)` them).
> - Keep the existing semantics that `'finished'` only fires after successful extraction.

---

OrKoN:
Closing as stale

PR #14864, opened 2026-04-15
test: fix extension test flakiness

OrKoN:
- service worker might not be fully available yet when the extension is uninstalled.
- service worker target can be made available after it has been destroyed.

### DIFF
--- a/packages/puppeteer-core/src/cdp/TargetManager.ts
+++ b/packages/puppeteer-core/src/cdp/TargetManager.ts
@@ -265,14 +265,14 @@ export class TargetManager
     this.#discoveredTargetsByTargetId.delete(event.targetId);
     this.#finishInitializationIfReady(event.targetId);
 
-    const target = this.#attachedTargetsByTargetId.get(event.targetId);
-    if (
-      target &&
-      (target.type() === 'service_worker' ||
-        targetInfo?.type === 'service_worker')
-    ) {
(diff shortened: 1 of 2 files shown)


nroscino (approved):

packages/puppeteer-core/src/cdp/TargetManager.ts
```diff
@@ -260,14 +264,15 @@ export class TargetManager
...
-      // Special case for service workers: report TargetGone event when
-      // the worker is destroyed.
-      const target = this.#attachedTargetsByTargetId.get(event.targetId);
-      if (target) {
-        this.emit(TargetManagerEvent.TargetGone, target);
-        this.#attachedTargetsByTargetId.delete(event.targetId);
-      }
+
+    const target = this.#attachedTargetsByTargetId.get(event.targetId);
+    if (
```
Lightning00Blade: Let' leave the comment, also does this if change make sense? 
I can see the Order being swapped which is part of the fix, maybe?

Lightning00Blade (approved):

packages/puppeteer-core/src/cdp/TargetManager.ts
OrKoN: I think the order swap is just a precaution. Let me try to revert the this part.

PR #14873, opened 2026-04-16
feat: implement URL blocklist to restrict access to unauthorized sites

nattallius:
# Feature: URL Blocklist for Puppeteer

This PR introduces a URL blocklist feature for Puppeteer, allowing developers to restrict network access to specific URLs or origins. This is useful for security-sensitive automation environments where access to certain domains must be strictly controlled.

## How it works

The feature is exposed via a new experimental `blockList` option in `ConnectOptions`. It uses the standard [URLPattern](https://urlpattern.spec.whatwg.org/) API for matching URLs.

The blocklist enforces restrictions in two ways:

1.  **On Connection**: When connecting to an existing browser, Puppeteer will silently detach from any already open targets (tabs, workers, etc.) that violate the patterns. This ensures that access to restricted content is neutralized immediately upon attachment.
2.  **During Operation**: For any network requests made by the browser (including main frame navigations and subresources like images, scripts, or stylesheets), the request will fail with an error if the URL matches a blocked pattern.

## Example Usage

```typescript
  blockList: [
    '*://example.com/*', // Block a specific domain
    '*://*.evil.com/*',  // Block all subdomains
  ],
});

### DIFF
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -71,6 +71,7 @@ export class CdpBrowser extends BrowserBase {
     networkEnabled = true,
     issuesEnabled = true,
     handleDevToolsAsPage = false,
+    blockList?: string[],
   ): Promise<CdpBrowser> {
     const browser = new CdpBrowser(
       connection,
@@ -84,6 +85,7 @@ export class CdpBrowser extends BrowserBase {
       networkEnabled,
       issuesEnabled,
       handleDevToolsAsPage,
+      getNetworkConditions(blockList),
     );
     if (acceptInsecureCerts) {
       await connection.send('Security.setIgnoreCertificateErrors', {
@@ -119,6 +121,7 @@ export class CdpBrowser extends BrowserBase {
     networkEnabled = true,
     issuesEnabled = true,
     handleDevToolsAsPage = false,
+    networkConditions?: Protocol.Network.EmulateNetworkConditionsByRuleRequest,
   ) {
     super();
     this.#networkEnabled = networkEnabled;
@@ -139,6 +142,7 @@ export class CdpBrowser extends BrowserBase {
       this.#createTarget,
       this.#targetFilterCallback,
       waitForInitiallyDiscoveredTargets,
+      networkConditions,
     );
     this.#defaultContext = new CdpBrowserContext(this.#connection, this);
     for (const contextId of contextIds) {
@@ -618,4 +622,22 @@ export class CdpBrowser extends BrowserBase {
   override isIssuesEnabled(): boolean {
     return this.#issuesEnabled;
   }
+}
+
+function getNetworkConditions(
+  blockList?: string[],
+): Protocol.Network.EmulateNetworkConditionsByRuleRequest | undefined {
+  if (!blockList?.length) {
+    return undefined;
+  }
+  const matchedNetworkConditions = blockList.map(pattern => {
+    return {
+      offline: true,
+      urlPattern: pattern,
+      latency: 0,
+      downloadThroughput: -1,
+      uploadThroughput: -1,
+    };
+  });
+  return {matchedNetworkConditions};
 }
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -632,12 +632,11 @@ function getNetworkConditions(
   }
   const matchedNetworkConditions = blockList.map(pattern => {
     return {
-      offline: true,
       urlPattern: pattern,
       latency: 0,
       downloadThroughput: -1,
       uploadThroughput: -1,
     };
   });
-  return {matchedNetworkConditions};
+  return {matchedNetworkConditions, offline: true};
 }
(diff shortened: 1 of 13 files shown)

### CODE REVIEW

File: test/src/network_restrictions.spec.ts [hunk 1]
```diff
@@ -0,0 +1,176 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import expect from 'expect';
+import puppeteer from 'puppeteer/internal/puppeteer.js';
+
+import {launch} from './mocha-utils.js';
+import {html} from './utils.js';
+
+describe('Network Restrictions', function () {
```
OrKoN: let's move the test under the cdp sub-folder as it is cdp-specific currently.

OrKoN:
let's improve the title of the PR for the changelog and add a PR description explaining how the feature works.

packages/puppeteer-core/src/common/ConnectOptions.ts
```diff
@@ -161,4 +161,9 @@ export interface ConnectOptions {
    * Only works for `protocol="webDriverBiDi"` and {@link Puppeteer.connect}.
    */
   capabilities?: SupportedWebDriverCapabilities;
+
+  /**
+   * @experimental
```
OrKoN, line 187: Let's document in details how it works and what it does.

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -621,3 +623,20 @@ export class CdpBrowser extends BrowserBase {
     return this.#issuesEnabled;
   }
 }
+
+function getNetworkConditions(
```
OrKoN: let's pass blocklist as is and convert to the network conditions when we actually send them.

packages/puppeteer-core/src/cdp/TargetManager.ts
```diff
@@ -356,6 +360,13 @@ export class TargetManager
       return;
     }
 
+    // If we connect to a browser that is already open,
+    // immediately detach from any tab that is on the blocklist.
+    if (!this.#initialAttachDone && !this.#isUrlAllowed(targetInfo.url)) {
```
OrKoN: let's move it before the service_worker block?

packages/puppeteer-core/src/cdp/TargetManager.ts
```diff
@@ -452,4 +464,45 @@ export class TargetManager
...
+
+    return true;
+  };
+
+  #maybeSetupNetworkConditions = async (session: CDPSession): Promise<void> => {
+    if (!this.#networkConditions) {
+      return;
+    }
+
+    await session.send('Network.enable');
```
OrKoN: do we need to call `Network.enable`? I think we are not expecting any events and it should work without Network.enable

test/src/network_restrictions.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+    ]);
+
+    expect(imageError).toBeDefined();
+    expect(imageError).toContain('net::ERR_INTERNET_DISCONNECTED');
+    expect(cssFinished).toBe(true);
+
+    await close();
+  });
+
+  it('should detach from targets violating blocklist when connecting to a running browser', async () => {
```
OrKoN: let's add a test for URLs like `chrome://version/`I think it is fine if this does not get blocked right now but let's document and add test coverage (I think those URLs do not go to network)

test/src/network_restrictions.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+    });
+
+    const targets = connectedBrowser.targets();
+    const blockedTarget = targets.find(t => {
+      return t.url() === blockedUrl;
+    });
+
+    expect(blockedTarget).toBeUndefined();
+
+    await connectedBrowser.disconnect();
```
OrKoN: let's move all disconnect/close call in all tests into `finally` block and wrap the test in a `try {} finally {}` block

test/src/network_restrictions.spec.ts
```diff
@@ -0,0 +1,176 @@
...
+
+    const imageError = await Promise.race([
+      imageBlockedErrorPromise,
+      new Promise<string | undefined>(r => {
+        return setTimeout(() => {
+          return r(undefined);
+        }, 3000);
+      }),
+    ]);
+    const cssFinished = await Promise.race([
```
OrKoN: can we use the networkIdle0 instead and then check the status of collected requests? it would allow the test to time out faster if something is wrong

OrKoN (commented):
Thanks! Mostly looks good, added comments to improve documentation and test coverage

test/src/network_restrictions.spec.ts
OrKoN: you would need to update TestExpectations.json and exclude this test for all platforms for the "pipe" configuration as there is no websocket and no connect possible in that configuration.

test/TestExpectations.json
```diff
@@ -132,6 +132,27 @@
...
+    "parameters": ["webDriverBiDi"],
+    "expectations": ["SKIP"],
+    "comment": "CDP-specific feature, not supported in WebDriver BiDi"
+  },
+  {
+    "testIdPattern": "[network_restrictions.spec] *",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["chrome-headless-shell"],
+    "expectations": ["SKIP"],
+    "comment": "Not supported by chrome-headless-shell"
```
OrKoN, line 154: I think it should work in chrome-headless-shell. What was the issue there?

OrKoN (approved):
LGTM thanks!

test/TestExpectations.json
OrKoN: let's address in the a follow-up

PR #14879, opened 2026-04-17
fix: reuse the same Page instance across Target.asPage and Target.page

### LINKED ISSUE #14877: [Bug]: Target.page() still returns a different Page instance after Target.asPage()

```ts
try {
  const target = page.target();

  const pageFromAsPage = await target.asPage();
  const pageFromPage = await target.page();

  console.log(pageFromAsPage === pageFromPage);
  // expected: true
  // actual: false
} finally {
  await browser.close();
}
```

### Background

This appears to be a residual edge case after:
- #14843
- #14862

Repeated `Target.asPage()` calls now reuse the same `Page` instance, but `Target.asPage()` followed by `Target.page()` can still diverge.

At current `main`, `CdpTarget.asPage()` and `PageTarget.page()` still rely on separate caches (`_asPagePromise` vs `pagePromise`). If `asPage()` runs first, `page()` can still create a second `Page` instance.

### Expectation

For a page target, `Target.asPage()` and `Target.page()` should resolve to the same `Page` instance regardless of call order.

### Reality

`Target.asPage()` followed by `Target.page()` can still return two different `Page` instances.

```ts

```

### Puppeteer version

latest

### Node version

v20.19.6

### Package manager

npm

### Package manager version

10.8.2

### Operating system

macOS

shaun0927:

Bug fix.

Yes.

No.

**Summary**

Fixes #14877.

#14862 made repeated `Target.asPage()` calls return the same `Page` instance, but `Target.asPage()` followed by `Target.page()` could still diverge.

The root cause is that `CdpTarget.asPage()` and `PageTarget.page()` still used separate caches (`_asPagePromise` vs `pagePromise`). If `asPage()` ran first, `page()` could still create a second `Page`.

This PR keeps the fix intentionally small:
- reuse the `asPage()` promise from `PageTarget.page()` when it already exists
- add a regression test that covers `asPage()` followed by `page()`

No.

**Other information**

This is intended as a narrow follow-up to #14843 / #14862 rather than a broader Target API refactor.

### DIFF
--- a/packages/puppeteer-core/src/cdp/Target.ts
+++ b/packages/puppeteer-core/src/cdp/Target.ts
@@ -259,6 +259,9 @@ export class PageTarget extends CdpTarget {
   }
 
   override async page(): Promise<Page | null> {
+    if (this._asPagePromise) {
+      this.pagePromise = this._asPagePromise;
+    }
     if (!this.pagePromise) {
       const session = this._session();
       this.pagePromise = (
--- a/test/src/target.spec.ts
+++ b/test/src/target.spec.ts
@@ -33,6 +33,20 @@ describe('Target', function () {
     await page.close();
   });
 
+  it('Target.page() should reuse the page created via Target.asPage()', async () => {
+    const {browser} = await getTestState();
+    const page = await browser.newPage();
+    const target = page.target();
+
+    const pageFromAsPage = await target.asPage();
+    const pageFromPage = await target.page();
+
+    expect(pageFromAsPage).toBe(page);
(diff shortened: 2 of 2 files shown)


OrKoN:
@Lightning00Blade PTAL if we can/should re-use the promises between the two. I think the use case is not very strong but does not hurt?

shaun0927:
Closing per discussion in #14877 — asPage is intended for 'other'-type targets only, so the page-target divergence isn't a real code path.

PR #14880, opened 2026-04-17
fix(webmcp): avoid hanging execute() calls after early tool responses

### LINKED ISSUE #14878: [Bug]: WebMCPTool.execute() can hang if toolResponded fires before the listener is attached

This is easiest to reproduce with a mocked CDP client because the failure depends on `toolResponded` arriving during `WebMCP.invokeTool`.

Current `WebMCPTool.execute()` does:

```ts
const {invocationId} = await this.#webmcp.invokeTool(this, input);
return await new Promise(resolve => {
  const handler = event => {
    if (event.id === invocationId) {
      this.#webmcp.off('toolresponded', handler);
      resolve(event);
    }
  };
  this.#webmcp.on('toolresponded', handler);
});
```

If `toolResponded` fires before the listener is attached, `execute()` can miss the response and wait forever.

A deterministic mocked-client repro is:

```ts
const client = {
  on,
  off,
  send: async method => {
    if (method === 'WebMCP.enable') {
      return;
    }
    if (method === 'WebMCP.invokeTool') {
      emit('WebMCP.toolResponded', {
        invocationId: 'invocation-id',
        status: 'Completed',
        output: 'early-result',
      });
      return {invocationId: 'invocation-id'};
    }
  },
};
```

With that setup, `tool.execute()` times out unless a second `toolResponded` event is emitted later.

### Background

This is on the experimental WebMCP API, but `tool.execute()` is documented as returning a promise with the tool result. A permanently pending promise when the browser already responded still looks like a bug.

Related feature PRs:
- #14835
- #14841
- #14851
- #14859

### Expectation

`WebMCPTool.execute()` should resolve with the first matching `toolResponded` result even if that event arrives during `invokeTool()`.

### Reality

If `toolResponded` arrives before the listener is attached, `execute()` can hang indefinitely.

```ts

```

### Puppeteer version

latest

### Node version

v20.19.6

### Package manager

npm

### Package manager version

10.8.2

### Operating system

macOS

shaun0927:

Bug fix.

Yes.

No.

**Summary**

Fixes #14878.

`WebMCPTool.execute()` attached its `toolresponded` listener only after `invokeTool()` resolved. If CDP delivered `toolResponded` during `invokeTool()`, `execute()` could miss the response and wait forever.

This PR keeps the change narrowly scoped to the WebMCP response path:
- cache `toolresponded` results long enough for `execute()` to consume an already-arrived response
- add a deterministic unit test that emits `WebMCP.toolResponded` during `WebMCP.invokeTool`

No.

**Other information**

The behavior is on the experimental WebMCP API, but `tool.execute()` is documented as returning a promise with the tool result, so hanging after an already-delivered response still looks like a correctness bug.

### DIFF
--- a/packages/puppeteer-core/src/cdp/WebMCP.ts
+++ b/packages/puppeteer-core/src/cdp/WebMCP.ts
@@ -163,11 +163,15 @@ export class WebMCPTool extends EventEmitter<{
    */
   async execute(input: object = {}): Promise<WebMCPToolCallResult> {
     const {invocationId} = await this.#webmcp.invokeTool(this, input);
+    const completedResponse = this.#webmcp.consumeToolResponse(invocationId);
+    if (completedResponse) {
+      return completedResponse;
+    }
     return await new Promise<WebMCPToolCallResult>(resolve => {
       const handler = (event: WebMCPToolCallResult) => {
         if (event.id === invocationId) {
           this.#webmcp.off('toolresponded', handler);
-          resolve(event);
+          resolve(this.#webmcp.consumeToolResponse(invocationId) ?? event);
         }
       };
       this.#webmcp.on('toolresponded', handler);
@@ -292,6 +296,7 @@ export class WebMCP extends EventEmitter<{
   #frameManager: FrameManager;
   #tools = new Map<string, Map<string, WebMCPTool>>();
   #pendingCalls = new Map<string, WebMCPToolCall>();
+  #respondedCalls = new Map<string, WebMCPToolCallResult>();
 
   #onToolsAdded = (event: ProtocolWebMCPToolsAddedEvent) => {
     const tools: WebMCPTool[] = [];
@@ -350,11 +355,13 @@ export class WebMCP extends EventEmitter<{
       errorText: event.errorText,
       exception: event.exception,
     };
+    this.#respondedCalls.set(event.invocationId, response);
     this.emit('toolresponded', response);
   };
 
   #onFrameNavigated = (frame: Frame) => {
     this.#pendingCalls.clear();
+    this.#respondedCalls.clear();
     const frameTools = this.#tools.get(frame._id);
     if (!frameTools) {
       return;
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -350,11 +355,13 @@ export class WebMCP extends EventEmitter<{
       errorText: event.errorText,
       exception: event.exception,
     };
+    this.#respondedCalls.set(event.invocationId, response);
```
wolfib, line 358: Is this a potential memory leak? There is no limit to growth if there is no navigation. Would it make sense to delete entries after a timeout?

wolfib (commented):
Thanks! This looks a like a valid concern and a fix that is working.

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -163,11 +163,15 @@ export class WebMCPTool extends EventEmitter<{
    */
   async execute(input: object = {}): Promise<WebMCPToolCallResult> {
```
beaufortfrancois, line 164: What do you think of this self-contained fix instead to address https://github.com/puppeteer/puppeteer/issues/14878?

```js
async execute(input: object = {}): Promise<WebMCPToolCallResult> {
  let invocationId: string;

  return new Promise<WebMCPToolCallResult>(async (resolve) => {
    const handler = (event: WebMCPToolCallResult) => {
      if (invocationId && event.id === invocationId) {
        this.#webmcp.off('toolresponded', handler);
        resolve(event);
      }
    };

    this.#webmcp.on('toolresponded', handler);

    const response = await this.#webmcp.invokeTool(this, input);
    invocationId = response.invocationId;
  });
}

**Review by OrKoN — CHANGES_REQUESTED** [review]
I am not sure it is a real issue:

1) Chrome sends the response to invokeTool before the toolResponded event https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/inspector/inspector_web_mcp_agent.cc;l=186;drc=b4528fe64d8704dc89748f6e371c7f97b13da16f
2) Puppeteer starts a new JS task for each CDP response meaning that microtasks such as promise chains from the previous message would finish first.

That means the event listener would be set up before any toolResponded event is processed. It reproduces in the mock because the mock does not correctly emit the events. Emitting correctly, e.g,:

```
setTimeout(() => {
        this.emit('WebMCP.toolResponded', {
          invocationId: 'invocation-id',
          status: 'Completed',
          output: 'early-result',
        });
      }, 0)
```

does not reproduce the issue. Please let me know if this reproduces in a e2e environment. That would mean that websocket messages are not dispatched in dedicated tasks.

PR #14894, opened 2026-04-21
refactor!: make proxy agent optional peer dependency and update to v8

### LINKED ISSUE #14817: [Task]: Update `proxy-agent` to v8
### Task to do:

They dropped support for NodeJS 18, we can do this as part of the next major, 
or we can drop `proxy-agent` all together and just have API that accepts a proxy class.

### LINKED ISSUE #14893: "basic-ftp" package vulnerability - CVE-2026-39983
### Task to do:

"basic-ftp" dependency package version < 5.2.0 has a path traversal vulnerability. See https://www.sentinelone.com/vulnerability-database/cve-2026-39983/.

The chain of dependencies is:
puppeteer -> puppeteer/browsers -> proxy-agent -> pac-proxy-agent -> get-uri -> basic-ftp@5.2.0

OrKoN:
With this change, `proxy-agent` needs to be installed separately if proxy configuration for installing browsers is required.

Closes https://github.com/puppeteer/puppeteer/issues/14893
Closes https://github.com/puppeteer/puppeteer/issues/14817

### DIFF
--- a/docs/browsers-api/index.md
+++ b/docs/browsers-api/index.md
@@ -87,6 +87,14 @@ npx puppeteer browsers install chrome --install-deps
 
 1. Launching the system browsers is only possible for Chrome/Chromium.
 
+## Proxies
+
+The library and CLI respect `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` environment variables. For them to work, you must have the `proxy-agent` package installed:
+
+```bash
+npm install proxy-agent
+```
+
 ## Custom Providers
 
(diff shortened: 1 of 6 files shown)


Lightning00Blade (approved):
Let's not close the issue as this still does not update the version to v8 which is the latest.
I would prefer that as it switched to ESM only and we do that for this major as well.

OrKoN:
Good point, let me just relax the range to accept any proxy-agent. I think the interface is very stable.

PR #14897, opened 2026-04-22
feat: Implement allowlist

nattallius:
# Feature: URL allowlist for Puppeteer

This PR introduces a URL allowlist feature for Puppeteer, allowing developers to restrict network access to **only** a specific set of URLs or origins. This provides a "walled garden" approach, ensuring the browser cannot access any content outside of the trusted patterns provided.

**This feature will be available starting with the Chrome 149 major release.**

## How it works

The feature is exposed via a new experimental `allowlist` option in `ConnectOptions` and `LaunchOptions`. It uses the standard [URLPattern](https://urlpattern.spec.whatwg.org/) API for matching URLs.

The allowlist enforces restrictions in two ways:

1.  **On Connection/Discovery**: When connecting to an existing browser or discovering new targets, Puppeteer will silently detach from any target whose URL does not match at least one pattern in the `allowlist`. This ensures the browser session is restricted to authorized content immediately.
2.  **During Operation**: For any network requests made by the browser (including main frame navigations and subresources like images, scripts, or stylesheets), the request will fail with a `net::ERR_INTERNET_DISCONNECTED` error if the URL does not match any pattern in the allowlist.

**Note**: The `allowlist` cannot be used in conjunction with the `blocklist` option.

## Example Usage

```typescript
// Restrict the browser to only interact with trusted domains
  allowlist: [
    '*://google.com/*',        // Allow Google
    '*://*.google.com/*',      // Allow all Google subdomains
    '*://github.com/puppeteer/*', // Allow only the Puppeteer repo on GitHub
  ],
});

// This navigation will succeed

// This navigation will fail with net::ERR_INTERNET_DISCONNECTED
```

### DIFF
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -89,6 +89,11 @@ export class CdpBrowser extends BrowserBase {
       blocklist,
       allowlist,
     );
+    const version = await browser.#getVersion();
+    const majorVersion = parseInt(version.product.match(/\d+/)?.[0] ?? '0', 10);
+    if (allowlist && majorVersion < 149) {
+      throw new Error('The allowlist option require Chrome 149 or greater.');
+    }
     if (acceptInsecureCerts) {
       await connection.send('Security.setIgnoreCertificateErrors', {
         ignore: true,
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -71,8 +71,8 @@ export class CdpBrowser extends BrowserBase {
     networkEnabled = true,
     issuesEnabled = true,
     handleDevToolsAsPage = false,
-    blockList?: string[],
-    allowList?: string[],
+    blocklist?: string[],
+    allowlist?: string[],
   ): Promise<CdpBrowser> {
     const browser = new CdpBrowser(
       connection,
@@ -86,8 +86,8 @@ export class CdpBrowser extends BrowserBase {
       networkEnabled,
       issuesEnabled,
       handleDevToolsAsPage,
-      blockList,
-      allowList,
+      blocklist,
+      allowlist,
     );
     if (acceptInsecureCerts) {
       await connection.send('Security.setIgnoreCertificateErrors', {
@@ -123,8 +123,8 @@ export class CdpBrowser extends BrowserBase {
     networkEnabled = true,
     issuesEnabled = true,
     handleDevToolsAsPage = false,
-    blockList?: string[],
-    allowList?: string[],
+    blocklist?: string[],
+    allowlist?: string[],
   ) {
     super();
     this.#networkEnabled = networkEnabled;
@@ -145,8 +145,8 @@ export class CdpBrowser extends BrowserBase {
       this.#createTarget,
       this.#targetFilterCallback,
       waitForInitiallyDiscoveredTargets,
-      blockList,
-      allowList,
+      blocklist,
+      allowlist,
     );
     this.#defaultContext = new CdpBrowserContext(this.#connection, this);
     for (const contextId of contextIds) {
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -89,10 +89,16 @@ export class CdpBrowser extends BrowserBase {
       blocklist,
       allowlist,
     );
-    const version = await browser.#getVersion();
-    const majorVersion = parseInt(version.product.match(/\d+/)?.[0] ?? '0', 10);
-    if (allowlist && majorVersion < 149) {
-      throw new Error('The allowlist option require Chrome 149 or greater.');
+
+    if (allowlist) {
+      const version = await browser.#getVersion();
+      const majorVersion = parseInt(
+        version.product.match(/\d+/)?.[0] ?? '0',
+        10,
+      );
+      if (majorVersion < 149) {
+        throw new Error('The allowlist option require Chrome 149 or greater.');
+      }
     }
     if (acceptInsecureCerts) {
       await connection.send('Security.setIgnoreCertificateErrors', {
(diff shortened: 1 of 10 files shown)


packages/puppeteer-core/src/common/ConnectOptions.ts
```diff
@@ -184,7 +184,37 @@ export interface ConnectOptions {
...
+   * if the URL does not match any pattern in the allowList.
+   *
+   * @example Pattern to allow a specific domain:
+   * `*://example.com/*`
+   *
+   * @example Pattern to allow all subdomains:
+   * `*://*.example.com/*`
+   *
+   * @remarks
+   * Currently only supported for CDP connections.
```
OrKoN, line 215: let's mention minimum Chrome version needed?

test/src/cdp/network_restrictions.spec.ts
```diff
@@ -173,6 +173,87 @@ describe('Network Restrictions', function () {
...
+      expect(connectError).toBeDefined();
+      expect(connectError?.message).toContain(
+        'Cannot specify both blockList and allowList',
+      );
+    } finally {
+      await close();
+    }
+  });
+
+  it.only('should throw an error for an invalid pattern', async () => {
```
OrKoN:
```suggestion
  it('should throw an error for an invalid pattern', async () => {
```

test/src/cdp/network_restrictions.spec.ts
```diff
@@ -173,6 +173,87 @@ describe('Network Restrictions', function () {
...
+      expect(page.url()).not.toBe(blockedUrl);
+      console.log('page url', page.url());
+      expect(error).toBeDefined();
+      expect(error?.message).toContain('net::ERR_INTERNET_DISCONNECTED');
+    } finally {
+      await close();
+    }
+  });
+
+  it('should throw an error when both blockList and allowList are specified', async () => {
```
OrKoN: let's write a BrowserConnector unit test (a new file `BrowserConnector.test.ts`, `npm run unit`) for these?  I think launching a browser to test these error message is rather expensive.

OrKoN:
Please update title and description similar to https://github.com/puppeteer/puppeteer/pull/14873

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -86,6 +87,7 @@ export class CdpBrowser extends BrowserBase {
       issuesEnabled,
       handleDevToolsAsPage,
       blockList,
+      allowList,
```
OrKoN: btw let's rename to `allowlist` and `blocklist` before we adopt the API in the MCP server (https://en.wikipedia.org/wiki/Whitelist/https://en.wikipedia.org/wiki/Blacklist_(computing))

test/TestExpectations.json
```diff
@@ -1968,6 +1968,34 @@
     "expectations": ["FAIL"],
     "comment": "TODO: started to fail with Chrome headless shell 135 on Windows"
   },
+  {
+    "testIdPattern": "[network_restrictions.spec] Network Restrictions should only allow navigation to URLs in the allowlist",
+    "platforms": ["linux"],
```
OrKoN: let's disable on all platforms

test/TestExpectations.json
```diff
@@ -1968,6 +1968,34 @@
...
+    "parameters": ["cdp", "chrome", "headless"],
+    "expectations": ["FAIL"],
+    "comment": "allowlist option requires at least chrome 149 version"
+  },
+  {
+    "testIdPattern": "[network_restrictions.spec] Network Restrictions should only allow navigation to URLs in the allowlist",
+    "platforms": ["linux"],
+    "parameters": ["cdp", "chrome", "headful"],
```
OrKoN: you can only have one expectation by leaving out the mode and keeping cdp and chrome only

test/TestExpectations.json
```diff
@@ -1968,6 +1968,34 @@
...
+    "expectations": ["FAIL"],
+    "comment": "allowlist option requires at least chrome 149 version"
+  },
+  {
+    "testIdPattern": "[network_restrictions.spec] Network Restrictions should only allow navigation to URLs in the allowlist",
+    "platforms": ["linux"],
+    "parameters": ["chrome", "headless", "webDriverBiDiOnly"],
```
OrKoN, line 1981: I think we should skip the test for webDriverBidi altogether

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -71,7 +71,8 @@ export class CdpBrowser extends BrowserBase {
     networkEnabled = true,
     issuesEnabled = true,
     handleDevToolsAsPage = false,
-    blockList?: string[],
+    blocklist?: string[],
+    allowlist?: string[],
   ): Promise<CdpBrowser> {
     const browser = new CdpBrowser(
```
OrKoN, line 77: I wonder if we should fetch Browser.getVersion if allowlist is provided here and verify that the Chrome version 149+? WDYT? to avoid that someone things the allowlist works when it does not.

nattallius: Maybe we can mark it on TS doc as deprecated? and lift up the warning about version?
Cause it's only for 1.5 month. 
And in case of checking and parsing version, you suggest to throw an error or what?

OrKoN: I do no think deprecated tsdoc would make too much difference here. Yeah, I think we can throw an error if the browser version does not match what we require and the allowlist is provided.

OrKoN:
Please update the PR description to use the new capitalization of the word allowlist.

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -89,6 +89,11 @@ export class CdpBrowser extends BrowserBase {
       blocklist,
       allowlist,
     );
+    const version = await browser.#getVersion();
```
OrKoN: let's only fetch this if the allow list is provided?

packages/puppeteer-core/src/common/ConnectOptions.ts
```diff
@@ -212,11 +214,16 @@ export interface ConnectOptions {
...
    * @experimental
    */
   allowlist?: string[];
+
+  /**
+   * Specify browser version to connect to.
+   *
+   * @experimental
+   */
+  browserVersion?: string;
```
OrKoN: why is this needed?

OrKoN (approved):

PR #14916, opened 2026-04-26
fix(bidi): clean up disposed Request objects to prevent memory leak

### LINKED ISSUE #14876: [Bug]: Memory leak issue with data url

```TypeScript
/**
 * Minimal reproduction of a Puppeteer BiDi memory leak.
 *
 * ## Root cause
 *
 * When a page loads images via `data:` URLs, Chrome's BiDi layer fires
 * `network.beforeRequestSent` followed by `network.responseCompleted`.
 * Puppeteer's `Request` object is therefore correctly *disposed* after each
 * load.  However, disposed `Request` objects are only removed from
 * `BrowsingContext.#requests` when `browsingContext.navigationStarted` fires
 * (see BrowsingContext.js).  In a long-running single-page session that never
 * navigates (e.g. a headless test runner that processes thousands of tasks on
 * the same page), the `#requests` Map grows without bound.
 *
 * Crucially, `Request.#event` — which holds the full BiDi event including the
 * raw `request.url` string — is **never nulled out** by `dispose()`.  For
 * `data:` URLs encoding large SVG documents (as used by pdf.js reftests,
 * ~6 MB each), this means every rendered SVG annotation layer permanently
 * occupies ~6 MB of Node.js heap for the lifetime of the page.
 *
 * ## What this script shows
 *
 * 1. Loading N large unique data: URL images causes Node.js heap to grow by
 *    roughly N × (size of each URL).
 * 2. The memory is NOT released even after all images finish loading
 *    (responseCompleted has fired, requests are disposed).
 * 3. Navigating the page (page.reload()) triggers the cleanup loop in
 *    BrowsingContext and the heap recovers.
 *
 * ## Versions
 *
 *   puppeteer/puppeteer-core: 24.x
 *   Chrome:  124+ (BiDi enabled by default)
 *
 * ## Run
 *
 *   node test/bidi-data-url-leak.mjs
 */

import v8 from "v8";

// ── helpers ──────────────────────────────────────────────────────────────────

function mb(bytes) {
  return (bytes / 1024 / 1024).toFixed(1) + " MB";
}

// Force a full GC cycle so heap readings are comparable.
function gcAndMeasure() {
  v8.setFlagsFromString("--expose-gc");
  // eslint-disable-next-line no-undef
  if (typeof globalThis.gc === "function") globalThis.gc();
  return process.memoryUsage().heapUsed;
}

// Build a unique large SVG data URL.  The SVG contains a comment that varies
// so Chrome cannot serve subsequent loads from cache.  ~100 KB of padding
// makes the heap impact clearly visible even with a modest IMAGE_COUNT.
function makeLargeDataUrl(index) {
  const padding = "x".repeat(100_000); // ~100 KB of payload per URL
  const svg = `<svg xmlns="http://www.w3.org/2000/svg" width="100" height="100">
  <!-- unique:${index} padding:${padding} -->
  <rect width="100" height="100" fill="hsl(${index},80%,50%)"/>
</svg>`;
  return "data:image/svg+xml;base64," + Buffer.from(svg).toString("base64");
}

// Load an array of image URLs in the page and wait for all to complete.
async function loadImages(page, urls) {
  await page.evaluate(async srcs => {
    await Promise.all(
      srcs.map(
        src =>
          new Promise((resolve, reject) => {
            const img = new Image();
            img.onload = resolve;
            img.onerror = reject;
            img.src = src;
          })
      )
    );
  }, urls);
  // Give Chrome a moment to flush any pending BiDi events.
  await new Promise(r => setTimeout(r, 300));
}

// ── main ─────────────────────────────────────────────────────────────────────

const IMAGE_COUNT = 30; // images to load; 30 × ~130 KB ≈ ~4 MB of URL data

  browser: "chrome",
  protocol: "webDriverBiDi",
  headless: true,
  args: ["--no-sandbox", "--js-flags=--expose-gc"],
});

try {
  await page.goto("about:blank");

  // Pre-build all URLs so allocation time is not charged to the measurement.
  const urls = Array.from({ length: IMAGE_COUNT }, (_, i) =>
    makeLargeDataUrl(i)
  );
  const urlDataMb = urls.reduce((s, u) => s + u.length, 0) / 1024 / 1024;
  console.log(
    `\nPrepared ${IMAGE_COUNT} unique data: URLs  ` +
      `(total URL data ≈ ${urlDataMb.toFixed(1)} MB)\n`
  );

  // ── Baseline ───────────────────────────────────────────────────────────────
  const heapBefore = gcAndMeasure();
  console.log(`Heap before loading images  : ${mb(heapBefore)}`);

  // ── Load images (no navigation — simulates a long-running test session) ────
  await loadImages(page, urls);
  const heapAfterLoad = gcAndMeasure();
  console.log(`Heap after  loading images  : ${mb(heapAfterLoad)}`);
  console.log(
    `Growth (GC cannot reclaim)  : +${mb(heapAfterLoad - heapBefore)}`
  );
  console.log(
    `Expected if no leak         : ~0 MB  (URLs should be GC-eligible)\n`
  );

  // ── Navigate to trigger BrowsingContext cleanup ────────────────────────────
  console.log("Navigating page (triggers BrowsingContext.#requests cleanup)…");
  await page.goto("about:blank");
  await new Promise(r => setTimeout(r, 300));
  const heapAfterNav = gcAndMeasure();
  console.log(`Heap after  navigation      : ${mb(heapAfterNav)}`);
  console.log(
    `Recovered after navigation  : -${mb(heapAfterLoad - heapAfterNav)}\n`
  );

  // ── Verdict ────────────────────────────────────────────────────────────────
  const leaked = heapAfterLoad - heapBefore;
  const recovered = heapAfterLoad - heapAfterNav;
  const leakThreshold = urlDataMb * 0.5 * 1024 * 1024; // 50 % of URL data

  console.log("── Verdict ──────────────────────────────────────────────────");
  if (leaked > leakThreshold) {
    console.log(
      `BUG CONFIRMED:\n` +
        `  • Loading ${IMAGE_COUNT} data: URL images grew the heap by ` +
        `${mb(leaked)}, even though all network.responseCompleted events\n` +
        `    were received and every BidiHTTPRequest was disposed.\n` +
        `  • The heap did NOT shrink until the page navigated ` +
        `(−${mb(recovered)} recovered).\n` +
        `\n` +
        `Root cause: BrowsingContext.#requests only evicts disposed entries\n` +
        `on browsingContext.navigationStarted.  In sessions that never\n` +
        `navigate (e.g. a headless test runner), disposed Request objects\n` +
        `accumulate indefinitely because Request.dispose() does not null out\n` +
        `Request.#event (which holds the full URL string).\n` +
        `\n` +
        `Suggested fix: either\n` +
        `  (a) null out Request.#event / Request.#response in dispose(), or\n` +
        `  (b) delete the entry from BrowsingContext.#requests immediately\n` +
        `      when the request is disposed, rather than waiting for\n` +
        `      navigationStarted.`
    );
  } else {
    console.log("No significant leak detected.");
  }
} finally {
  await browser.close();
}
```

### Background

For pdf.js ref tests I added support for parallelizing everything, unfortunately with Chrome, node consistently crashes because of not having enough memory. If I run node with `--max-old-space-size=8192` then it's fine.
While looking at a memory dump in Chrome devtools I noticed that ~700Mb are strings `data:image/svg+xml;base64,...` which are used in Chrome for generating the images for the ref tests but those strings are in node and are kept by `BrowsingContext.#requests`.
With Claude help, I created the above test case. In case that helps here is the memory snapshot:
[Heap.20260417.155145.55172.0.001.zip](https://github.com/user-attachments/files/26831460/Heap.20260417.155145.55172.0.001.zip)

For info, the ref tests are running well with no crash when using Firefox.

### Expectation

No useless memory retention.

### Reality

Crash because of not having enough memory.

```TypeScript

```

### Puppeteer version

24.41.0

### Node version

24.13.0

### Package manager

npm

### Package manager version

11.6.3

### Operating system

Windows

wahajahmed010:
Fixes #14876

Disposed Request objects were never removed from BrowsingContext.#requests when using data: URLs, because cleanup only happened on navigationStarted events. This caused unbounded heap growth in long-running BiDi sessions.

**Changes:**
- Null out Request.#event and Request.#response in dispose() to release references
- Add disposed event to Request so BrowsingContext can listen for it
- Delete the request from BrowsingContext.#requests immediately upon disposal

**Testing:**
- [x] Added unit test verifying Request emits disposed event and disposes correctly
- [ ] Existing tests pass (full CI required)

This contribution was AI-assisted.

### DIFF
--- a/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
+++ b/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
@@ -323,6 +323,9 @@ export class BrowsingContext extends EventEmitter<{
 
       const request = Request.from(this, event);
       this.#requests.set(request.id, request);
+      request.once('disposed', () => {
+        this.#requests.delete(request.id);
+      });
       this.emit('request', {request});
     });
 
--- a/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
+++ b/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
@@ -719,6 +719,16 @@ export class BrowsingContext extends EventEmitter<{
       'Browsing context already closed, probably because the user context closed.';
     this.emit('closed', {reason: this.#reason});
 
+    // Dispose any remaining undisposed requests and clear the map.
+    // This catches requests that never received responseCompleted (e.g.,
+    // aborted, network errors) and prevents them from leaking.
+    for (const request of this.#requests.values()) {
+      if (!request.disposed) {
+        request[disposeSymbol]();
+      }
+    }
+    this.#requests.clear();
+
     this.#disposables.dispose();
     super[disposeSymbol]();
   }
--- a/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
+++ b/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
@@ -27,13 +27,6 @@ import {UserPrompt} from './UserPrompt.js';
 /**
  * @internal
  */
-function createRequestWeakRef(request: Request): WeakRef<Request> {
-  return new WeakRef(request);
-}
-
-/**
- * @internal
- */
 export type AddInterceptOptions = Omit<
   Bidi.Network.AddInterceptParameters,
   'contexts'
@@ -169,7 +162,7 @@ export class BrowsingContext extends EventEmitter<{
   readonly #children = new Map<string, BrowsingContext>();
   readonly #disposables = new DisposableStack();
   readonly #realms = new Map<string, WindowRealm>();
-  readonly #requests = new Map<string, WeakRef<Request>>();
+  readonly #requestIds = new Set<string>();
   readonly defaultRealm: WindowRealm;
   readonly id: string;
   readonly parent: BrowsingContext | undefined;
@@ -291,11 +284,13 @@ export class BrowsingContext extends EventEmitter<{
       // Note: we should not update this.#url at this point since the context
       // has not finished navigating to the info.url yet.
 
-      for (const [id, requestRef] of this.#requests) {
-        const request = requestRef.deref();
-        if (!request || request.disposed) {
-          this.#requests.delete(id);
-        }
+      for (const id of this.#requestIds) {
+        // Clean up entries for requests that have been disposed or completed.
+        // Since we only store IDs (not objects), there's no memory leak from
+        // the requests themselves — this cleanup just prevents ID accumulation.
+        // Note: we can't check disposed status without the object, so we clear
+        // the entire set. Pending requests will be re-added by beforeRequestSent.
+        this.#requestIds.delete(id);
       }
       // If the navigation hasn't finished, then this is nested navigation. The
       // current navigation will handle this.
@@ -323,14 +318,18 @@ export class BrowsingContext extends EventEmitter<{
       if (event.context !== this.id) {
         return;
       }
-      if (this.#requests.has(event.request.request)) {
+      if (this.#requestIds.has(event.request.request)) {
         // Means the request is a redirect. This is handled in Request.
         // Or an Auth event was issued
         return;
       }
 
       const request = Request.from(this, event);
-      this.#requests.set(request.id, createRequestWeakRef(request));
+      const requestId = request.id;
+      this.#requestIds.add(requestId);
+      request.once('disposed', () => {
+        this.#requestIds.delete(requestId);
+      });
       this.emit('request', {request});
     });
 
@@ -724,16 +723,10 @@ export class BrowsingContext extends EventEmitter<{
       'Browsing context already closed, probably because the user context closed.';
     this.emit('closed', {reason: this.#reason});
 
-    // Dispose any remaining undisposed requests and clear the map.
-    // This catches requests that never received responseCompleted (e.g.,
-    // aborted, network errors) and prevents them from leaking.
-    for (const [, requestRef] of this.#requests) {
-      const request = requestRef.deref();
-      if (request && !request.disposed) {
-        request[disposeSymbol]();
-      }
-    }
-    this.#requests.clear();
+    // Clean up request ID tracking on context disposal.
+    // Requests self-dispose via their 'closed' listener on BrowsingContext,
+    // so we don't need to manually dispose them — just clear the ID set.
+    this.#requestIds.clear();
 
     this.#disposables.dispose();
     super[disposeSymbol]();
--- a/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
+++ b/packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
@@ -284,14 +284,9 @@ export class BrowsingContext extends EventEmitter<{
       // Note: we should not update this.#url at this point since the context
       // has not finished navigating to the info.url yet.
 
-      for (const id of this.#requestIds) {
-        // Clean up entries for requests that have been disposed or completed.
-        // Since we only store IDs (not objects), there's no memory leak from
-        // the requests themselves — this cleanup just prevents ID accumulation.
-        // Note: we can't check disposed status without the object, so we clear
-        // the entire set. Pending requests will be re-added by beforeRequestSent.
-        this.#requestIds.delete(id);
-      }
+      // Clear all request IDs on navigation start. Pending requests will be
+      // re-added by beforeRequestSent.
+      this.#requestIds.clear();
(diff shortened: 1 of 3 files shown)


wolfib:
@zyzyzyryxy Could you PTAL since it seems you're already working on this as well?

Qodo-Free-For-OSS:
Hi, `Request[disposeSymbol]()` now sets `#event` to `undefined`, but public accessors (`id`, `url`, `headers`, `timing()`, `getResponseContent()`) dereference `#event`, causing runtime TypeErrors when callers access request data after completion/disposal. It also clears `#response`, which can break code paths that treat `request.response` as a “request already finished” signal.

**How to fix:** Cache fields; keep getters safe

> ### Issue description
> `Request[disposeSymbol]()` sets `#event` (and `#response`) to `undefined`, but the class exposes public getters/methods that dereference `#event`. After a request completes and is disposed (normal path), any caller that later reads `request.url`, `request.id`, `request.headers`, calls `timing()`, or calls `getResponseContent()` can crash with a `TypeError`. Clearing `#response` can also break logic that checks `request.response` to determine whether a request is already finished.
> 
> ### Issue Context
> The PR’s goal is to release references to avoid leaks, but `Request` is used by higher-level BiDi wrappers (e.g. `BidiHTTPRequest` / `BidiHTTPResponse`) which may access request data after completion.
> 
> ### Fix Focus Areas
> - packages/puppeteer-core/src/bidi/core/Request.ts[167-207]
> - packages/puppeteer-core/src/bidi/core/Request.ts[287-317]
> - packages/puppeteer-core/src/bidi/core/Request.ts[343-353]
> 
> ### Suggested approach
> - Introduce stable cached fields on `Request` (e.g., `#id`, `#url`, `#method`, `#headers`, `#timings`, `#isBlocked`, `#navigation`, etc.) populated in the constructor/initialize path.
> - Update getters/methods (`id`, `url`, `headers`, `timing()`, etc.) to use cached fields so they remain valid after disposal.
> - Do **not** clear `#response` if it is part of observable state (`request.response` used as a completion signal). If you must drop large references, consider caching only the minimal response data needed and nulling the larger object.
> - If `getResponseContent()` needs `id`, ensure it uses the cached `#id` so it still works post-dispose.

We noticed a couple of other issues in this PR as well - happy to share if helpful.

---

zyzyzyryxy:
Hello @wahajahmed010, thank you for your contribution. It basically does the same as my earlier https://github.com/puppeteer/puppeteer/issues/14876 and seems to suffer from the same issue: it doesn't remove the memory leak completely, only reduces it partially, as shown by the reproduction script attached to the bug report.

Could you look into this again, see if you can identify what else is leaking memory here?

zyzyzyryxy:
Following the AI review (now marked as spam) does not help, we need to reduce and control memory usage, not increase it.

nullifying `#event` seems like a red herring: even when we do it, the memory leak is still there (smaller though). We need to identify the root cause instead.

packages/puppeteer-core/src/bidi/core/BrowsingContext.ts
```diff
@@ -284,10 +284,13 @@ export class BrowsingContext extends EventEmitter<{
...
-          this.#requests.delete(id);
-        }
+      for (const id of this.#requestIds) {
+        // Clean up entries for requests that have been disposed or completed.
+        // Since we only store IDs (not objects), there's no memory leak from
+        // the requests themselves — this cleanup just prevents ID accumulation.
+        // Note: we can't check disposed status without the object, so we clear
+        // the entire set. Pending requests will be re-added by beforeRequestSent.
+        this.#requestIds.delete(id);
       }
```
zyzyzyryxy: If all elements are to be removed, `Set.clear()` is better.
```suggestion
      this.#requestIds.clear();
```

packages/puppeteer-core/src/bidi/core/Request.test.ts
```diff
@@ -0,0 +1,69 @@
+import {describe, it} from 'node:test';
```
zyzyzyryxy, line 1: The file is missing required license header and there are some compilation errors. Please run the build and address the issues accordingly.

zyzyzyryxy (changes requested):
Thanks for the update, this looks like a good progress!

Also I see CLA check failing, please verify if all changes were committed using the intended email address and either sign the CLA for the one failing the check or reupload them correctly.

wahajahmed010:
@zyzyzyryxy You're right. The reproduction in #14876 shows the leak persists even with my changes, so nullifying #event only masks it. I see #14891 already tackles the root cause. Happy to close this in favor of yours, or help test if useful.

PR #14921, opened 2026-04-28
test(webmcp): Add test for WebMCP form missing tooldescription

beaufortfrancois:

It adds a missing test for WebMCP

Yes

N/A

**Summary**

We want to make sure WebMCP issues are caught properly in Puppeteer.

None.

**Other information**

@OrKoN

### DIFF
--- a/test/src/page.spec.ts
+++ b/test/src/page.spec.ts
@@ -2546,11 +2546,15 @@ describe('Page', function () {
 
       const issuePromise = waitEvent<Issue>(page, 'issue');
 
-      await page.setContent(html`<form toolname="mytool">`);
+      await page.setContent(html`<form toolname="mytool"></form>`);
 
       const issue = await issuePromise;
       expect(issue).toBeTruthy();
       expect(issue.code).toBe('GenericIssue');
+      expect(issue.details.genericIssueDetails).toBeTruthy();
+      expect(issue.details.genericIssueDetails.errorType).toBe(
+        'FormModelContextMissingToolDescription',
+      );
     });
   });
 
--- a/test/src/page.spec.ts
+++ b/test/src/page.spec.ts
@@ -2538,6 +2538,20 @@ describe('Page', function () {
       expect(issue).toBeTruthy();
       expect(issue.code).toBe('ContentSecurityPolicyIssue');
     });
+
+    it('should emit issue event from WebMCP form missing tooldescription', async () => {
+      const {page, server} = await getTestState();
+
+      await page.goto(server.EMPTY_PAGE);
+
+      const issuePromise = waitEvent<Issue>(page, 'issue');
+
+      await page.setContent(html`<form toolname="mytool">`);
+
+      const issue = await issuePromise;
+      expect(issue).toBeTruthy();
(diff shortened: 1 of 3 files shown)


test/src/page.spec.ts
```diff
@@ -2538,6 +2538,20 @@ describe('Page', function () {
...
+
+      await page.goto(server.EMPTY_PAGE);
+
+      const issuePromise = waitEvent<Issue>(page, 'issue');
+
+      await page.setContent(html`<form toolname="mytool">`);
+
+      const issue = await issuePromise;
+      expect(issue).toBeTruthy();
+      expect(issue.code).toBe('GenericIssue');
```
OrKoN: let's assert that it is actually the webmcp issue by looking at details

beaufortfrancois: I'll do this when it is actually plugged in. My understanding is that it's "just" a generic issue for now right?

OrKoN: I think it will remain a generic issue. I do not think there will be any more work on the CDP side. The md descriptions are added to the devtools-frontend repo but Puppeteer only gives raw details.

OrKoN: I think we can already assert https://chromedevtools.github.io/devtools-protocol/tot/Audits/#type-GenericIssueErrorType (I would expect one of FormModelContext*)

beaufortfrancois: Done

beaufortfrancois: To be clear, this test is still not passing for me locally as `await issuePromise` doesn't resolve

OrKoN (approved):

test/src/page.spec.ts
OrKoN: It should be passing with the latest version (as the issue in the protocol) with WebMCP flags on I assume. Descriptions have to be added here (example: https://source.chromium.org/chromium/chromium/src/+/main:third_party/devtools-frontend/src/front_end/models/issues_manager/descriptions/genericFormLabelForNameError.md)

beaufortfrancois: Should we add those flags when running the test then?

OrKoN: yes, I believe so. You can also move the test to webmcp.spec.ts, perhaps it would be easier

beaufortfrancois: Fixed now. Thanks!

test/src/page.spec.ts
```diff
@@ -2538,6 +2539,35 @@ describe('Page', function () {
       expect(issue).toBeTruthy();
       expect(issue.code).toBe('ContentSecurityPolicyIssue');
     });
+
+    it('should emit issue event from WebMCP form missing tooldescription', async () => {
+      const {defaultBrowserOptions, server} = await getTestState({
```
OrKoN: I would prefer us to move this to webmcp.spec.ts actually so that we do not launch the entire browser just for one test here.

beaufortfrancois: Done

OrKoN (approved):

PR #14922, opened 2026-04-29
fix: do not open DevTools if it is already open

OrKoN:
Opening DevTools steals focus if it happens in the background.

### DIFF
--- a/docs/api/puppeteer.page.md
+++ b/docs/api/puppeteer.page.md
@@ -946,7 +946,7 @@ All timestamps are in monotonic time: monotonically increasing time in seconds s
 
 </td><td>
 
-Opens DevTools for the current Page and returns the DevTools Page. This method is only available in Chrome.
+Opens DevTools for the this page if not already open and returns the DevTools page. This method is only available in Chrome.
 
 </td></tr>
 <tr><td>
--- a/docs/api/puppeteer.page.opendevtools.md
+++ b/docs/api/puppeteer.page.opendevtools.md
@@ -4,7 +4,7 @@ sidebar_label: Page.openDevTools
 
 # Page.openDevTools() method
 
-Opens DevTools for the current Page and returns the DevTools Page. This method is only available in Chrome.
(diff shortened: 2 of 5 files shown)


docs/api/puppeteer.page.md
```diff
@@ -946,7 +946,7 @@ All timestamps are in monotonic time: monotonically increasing time in seconds s
 
 </td><td>
 
-Opens DevTools for the current Page and returns the DevTools Page. This method is only available in Chrome.
+Opens DevTools for the this page if not already open and returns the DevTools page. This method is only available in Chrome.
```
wolfib, line 949: 'the' OR 'this', not both

docs/api/puppeteer.page.opendevtools.md
```diff
@@ -4,7 +4,7 @@ sidebar_label: Page.openDevTools
 
 # Page.openDevTools() method
 
-Opens DevTools for the current Page and returns the DevTools Page. This method is only available in Chrome.
+Opens DevTools for the this page if not already open and returns the DevTools page. This method is only available in Chrome.
```
wolfib, line 7: 'the' OR 'this', not both

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -3233,8 +3233,8 @@ export abstract class Page extends EventEmitter<PageEvents> {
   }
 
   /**
-   * Opens DevTools for the current Page and returns the DevTools Page. This
-   * method is only available in Chrome.
+   * Opens DevTools for the this page if not already open and returns the DevTools page.
```
wolfib, line 3236: 'the' OR 'this', not both

wolfib (approved):

docs/api/puppeteer.page.md
OrKoN: no need to repeat the comment for generated content :) docs are generated from code.

packages/puppeteer-core/src/api/Page.ts
OrKoN: Thanks, fixed!

PR #14924, opened 2026-04-29
docs: Update puppeteer doc to include example of extension realm evaluation

nroscino:
Updates the example with realm evaluation

### DIFF
--- a/docs/guides/chrome-extensions.md
+++ b/docs/guides/chrome-extensions.md
@@ -29,7 +29,7 @@ const browser = await puppeteer.launch({
   enableExtensions: true,
 });
 
-await browser.installExtension(pathToExtension);
+const extensionId = await browser.installExtension(pathToExtension);
 ```
 
 ## Background contexts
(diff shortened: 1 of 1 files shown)


docs/guides/chrome-extensions.md
```diff
@@ -111,7 +111,32 @@ await browser.close();
...
 
-It is not currently possible to evaluate code in the content script isolated world.
+To evaluate code in the context of a content script, you can use the `page.extensionRealms()` method to find the realm associated with the extension and then use its `evaluate()` method.
+
+```ts
+// Get the extension ID
+const extensionId = await browser.installExtension(pathToExtension);
+
+// Find the extension realm.
+const realms = page.extensionRealms();
```
nroscino, line 121: I was thinking... it would be cool also to have a `page.getExtensionRealmById` or even better to have it return a map with id as key and Realm as value 🤔  WDYT? @OrKoN

OrKoN: I think it would be great to make `extension()` work without `await` so it reduces to `page.extensionRealms().find(realm => realm.extension().id === extensionId)`. Also, I think we should change installExtension to return an instance of `Extension` (that would be a breaking change) and on `Extension` have the `realms()` getter.

OrKoN: I am not sure we should add multiple getters like `getExtensionRealmById` to our API.

OrKoN (approved):

docs/guides/chrome-extensions.md
nroscino: Ok I will note this things and will add a task for this on the follow up work!

PR #14927, opened 2026-04-29
fix: log WebSocket transport errors

### LINKED ISSUE #13054: [Bug]: BrowserWebSocketTrasnport errors are ignored

```TypeScript
// websocket of connected chrome instance receives error as result of message
```

### Background

Working with `@cloudflare/puppeteer` I got a WebSocket error when attempting to capture a screenshot that was beyond the size limit for a workerd WebSocket message (see https://github.com/cloudflare/workerd/issues/2667)

The resulting error was ignored by puppeteer, meaning my automation was left in a hanging state (could not cleanup browser connection of capture error).

After some digging through Puppeteer, this comment seems to indicate the problem: https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/common/BrowserWebSocketTransport.ts#L39

My naive solution was to call `onmessage` with something non-protocol like that could be caught in `Connection`, which would subsequently call `#callbacks.clear()`. Would you be open to a pull request along those lines are am I completely missing the complexity of such a change :)

### Expectation

Transport errors should probably result in a TargetClose error?

### Reality

Transport errors result in unresolved promises and hanging automations.

_No response_

### Puppeteer version

@cloudflare/puppeteer@0.0.14

### Node version

20.10.0

### Package manager

npm

### Package manager version

10.2.3

### Operating system

macOS

haosenwang1018:
### Summary
- Log WebSocket transport `error` events through the existing `debugError` channel instead of silently ignoring them.
- Apply the same behavior to browser and Node WebSocket transports.
- Add focused unit coverage for both transports.

Fixes #13054.

### Tests
- `node --test --test-reporter=spec packages/puppeteer-core/lib/cjs/puppeteer/common/BrowserWebSocketTransport.test.js packages/puppeteer-core/lib/cjs/puppeteer/node/NodeWebSocketTransport.test.js`
- `npx prettier --check packages/puppeteer-core/src/common/BrowserWebSocketTransport.ts packages/puppeteer-core/src/common/BrowserWebSocketTransport.test.ts packages/puppeteer-core/src/node/NodeWebSocketTransport.ts packages/puppeteer-core/src/node/NodeWebSocketTransport.test.ts`
- `npx eslint packages/puppeteer-core/src/common/BrowserWebSocketTransport.ts packages/puppeteer-core/src/common/BrowserWebSocketTransport.test.ts packages/puppeteer-core/src/node/NodeWebSocketTransport.ts packages/puppeteer-core/src/node/NodeWebSocketTransport.test.ts`

### DIFF
--- a/packages/puppeteer-core/src/common/BrowserWebSocketTransport.test.ts
+++ b/packages/puppeteer-core/src/common/BrowserWebSocketTransport.test.ts
@@ -1,0 +1,31 @@
+/**
+ * @license
+ * Copyright 2026 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {describe, it, afterEach} from 'node:test';
+
+import expect from 'expect';
+
+import {BrowserWebSocketTransport} from './BrowserWebSocketTransport.js';
+import {getCapturedLogs, setLogCapture} from './Debug.js';
+
+describe('BrowserWebSocketTransport', () => {
+  afterEach(() => {
+    setLogCapture(false);
+  });
+
+  it('should log WebSocket errors for debugging', () => {
+    const ws = new EventTarget() as WebSocket;
+    ws.send = () => {};
+    ws.close = () => {};
+
+    new BrowserWebSocketTransport(ws);
+
+    setLogCapture(true);
+    ws.dispatchEvent(new Event('error'));
+
+    expect(getCapturedLogs()).toHaveLength(1);
+    expect(getCapturedLogs()[0]).toContain('puppeteer:error');
+  });
+});
--- a/packages/puppeteer-core/src/common/BrowserWebSocketTransport.ts
+++ b/packages/puppeteer-core/src/common/BrowserWebSocketTransport.ts
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 import type {ConnectionTransport} from './ConnectionTransport.js';
+import {debugError} from './util.js';
 
 /**
  * @internal
@@ -36,8 +37,7 @@ export class BrowserWebSocketTransport implements ConnectionTransport {
         this.onclose.call(null);
       }
     });
-    // Silently ignore all errors - we don't know what to do with them.
-    this.#ws.addEventListener('error', () => {});
+    this.#ws.addEventListener('error', debugError);
   }
 
   send(message: string): void {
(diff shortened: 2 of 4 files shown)


packages/puppeteer-core/src/common/BrowserWebSocketTransport.test.ts
```diff
@@ -0,0 +1,31 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {describe, it, afterEach} from 'node:test';
+
+import expect from 'expect';
+
+import {BrowserWebSocketTransport} from './BrowserWebSocketTransport.js';
+import {getCapturedLogs, setLogCapture} from './Debug.js';
+
+describe('BrowserWebSocketTransport', () => {
```
OrKoN, line 13: I don't think we need a test for this. 1) it runs in the browser and being tested in Node 2) this is just a debugError call. Let's just change it without new tests.

Qodo-Free-For-OSS:
Hi, `debugError` is async in Node, but it is registered directly as a WebSocket `error` event listener, so any rejection inside `debugError` becomes an unhandled promise rejection (can fail tests/process under strict unhandled-rejection settings). This risk is introduced in both Browser and Node transports by wiring `debugError` directly to the event emitter.

**Severity:** action required | **Category:** reliability

**How to fix:** Wrap and catch debugError

> ### Issue description
> `debugError` is async in Node (returns a Promise). Registering it directly as an event listener can produce unhandled promise rejections if it throws/rejects.
> 
> ### Issue Context
> This PR wires `debugError` directly to WebSocket `error` events in both transports.
> 
> ### Fix Focus Areas
> - packages/puppeteer-core/src/node/NodeWebSocketTransport.ts[43-56]
> - packages/puppeteer-core/src/common/BrowserWebSocketTransport.ts[28-41]
> 
> ### Suggested change
> Replace `addEventListener('error', debugError)` with a wrapper that handles async safely, e.g.:
> 
> ```ts
> this.#ws.addEventListener('error', event => {
>   void (debugError(event) as Promise<void>).catch(() => {});
> });
> ```
> 
> If you prefer not to swallow failures silently, route the catch to a safe fallback (e.g., `console.error`) that cannot throw.

---
*[Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source.*

PR #14931, opened 2026-04-30
fix: do not set global offline flag for allowlist

nattallius:


packages/puppeteer-core/src/cdp/TargetManager.ts
```diff
@@ -547,9 +547,23 @@ export class TargetManager
       });
     }
 
-    await session.send('Network.emulateNetworkConditionsByRule', {
-      offline: true, // 'offline' will be deprecated in Chrome 149. Retained for compatibility with existing blocklist functionality.
-      matchedNetworkConditions,
+    const versionData = await session.send('Browser.getVersion').catch(() => {
```
OrKoN: we cannot get the version here. Pass the version from where we get it during connection to throw an error.

packages/puppeteer-core/src/cdp/TargetManager.ts
```diff
@@ -547,9 +547,23 @@ export class TargetManager
...
+      return null;
     });
+    const product = versionData?.product;
+    const majorVersion = parseInt(product?.match(/\d+/)?.[0] ?? '0', 10);
+
+    const payload: any = {
```
OrKoN: please do not use `any` for types to avoid bugs related to types. Use `// @ts-expect-error explanation` for where we need types that would be updated later.

packages/puppeteer-core/src/cdp/TargetManager.ts
nattallius: I removed version checking, cause appeared that it's excessive, we can pas `offline: true `when there is only blocklist and that's enough

packages/puppeteer-core/src/cdp/TargetManager.ts
```diff
@@ -547,9 +547,16 @@ export class TargetManager
       });
     }
 
-    await session.send('Network.emulateNetworkConditionsByRule', {
-      offline: true, // 'offline' will be deprecated in Chrome 149. Retained for compatibility with existing blocklist functionality.
```
OrKoN, line 551: can we inline this? `offline: this.#blocklist.length > 0 ? true : undefined`

Qodo-Free-For-OSS:
Hi, The new allowlist-related TestExpectations entries accept both FAIL and PASS, making any result acceptable and preventing canary CI from catching allowlist regressions once Chrome >=149 is used. This defeats the purpose of adding these tests because failures become “expected” forever unless manually corrected.

**How to fix:** Make expectations version-specific

> ### Issue description
> `TestExpectations.json` entries that include both `FAIL` and `PASS` make the test always acceptable, removing any CI signal.
> 
> ### Issue Context
> The comments indicate intent: “PASS in Chrome >= 149; FAIL in older revisions”. The current expectations format has no version gating, so the right approach is to **skip** these tests for the non-supporting Chrome versions (or move them into canary-only expectations and require PASS there).
> 
> ### Fix Focus Areas
> - test/TestExpectations.json[1434-1467]
> - test/CanaryTestExpectations.json[1-37]
> - .github/workflows/daily.yml[58-90] (understand canary merge behavior)
> 
> ### Suggested change
> - For regular CI (pinned Chrome < 149): change these expectations to `SKIP` (or remove and add SKIP) so they don’t run.
> - For canary CI (Chrome canary >= 149): put explicit `PASS` expectations into `test/CanaryTestExpectations.json` so regressions fail canary runs.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source). Free code review for open-source maintainers.*

packages/puppeteer-core/src/cdp/TargetManager.ts
nattallius: Then I am getting `undefined` is not assignable to `boolean`. And also in this case 
` // @ts-expect-error offline is blocklist only` won't take effect

OrKoN:
```
    await session.send('Network.emulateNetworkConditionsByRule', {
      // @ts-expect-error offline cannot be undefined before M149.
      offline: this.#blocklist.length > 0 ? true : undefined,
      matchedNetworkConditions,
    });
```

works fine for me. Do you still get an error for this?

test/TestExpectations.json
```diff
@@ -1431,12 +1431,40 @@
     "expectations": ["FAIL"],
     "comment": "https://bugzilla.mozilla.org/show_bug.cgi?id=1908952#c4 historyUpdated from beforeunload incorrectly comes before navigationStarted"
   },
+  {
+    "testIdPattern": "[network_restrictions.spec] Network Restrictions should block window.location.href navigation to URLs not in the allowlist",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["cdp", "chrome"],
+    "expectations": ["FAIL", "PASS"],
```
OrKoN: let's mark tests that fail as FAIL (no PASS) so that we remember to update the expectation when we roll Chrome.

packages/puppeteer-core/src/cdp/TargetManager.ts
nattallius: yes, works for me too now

OrKoN (approved):

PR #14939, opened 2026-05-04
feat: support checkboxes and radios in locator.fill

### LINKED ISSUE #14938: [Feature]: Handle checkboxes, radio buttons and toggles with Locator.fill()
### Feature description

# Objective

Update Puppeteer's Locator.fill() method so that it can correctly handle checkboxes, radio buttons, and ARIA switches. The method signature will be updated to accept a boolean value to explicitly set the checked state of these elements.

# Implementation Details

1. **Scope & API Change:**
   This logic will be implemented exclusively in the `Locator` class.
   The signature of `Locator.fill(value)` and `LocatorFillOptions` will be updated to accept `value: string | boolean`.

2. **Element Type Detection & Type Checking:**
   When `fill(value)` is called, evaluate the target element's type.
   * If it is a toggle (`<input type="checkbox|radio">` or ARIA `role="checkbox|radio|switch"`):
     * Verify `typeof value === 'boolean'`. Throw an error if a `string` is provided.
   * If it is a text input, select, etc.:
     * Verify `typeof value === 'string'`. Throw an error if a `boolean` is provided.

3. **State Synchronization:**
   To ensure all application logic and event listeners are triggered correctly, the toggle state will be updated via simulated user interaction:
   * Evaluate the current state of the element.
   * If the current state matches the desired state, do nothing.
   * If they differ, perform a `click()` on the element.
   * After clicking, re-evaluate the state. If it still does not match the desired state (which can happen for indeterminate/mixed checkboxes), throw an error to trigger the `Locator`'s internal retry mechanism for a subsequent action.

4. **Element-Specific Details & Edge Cases:**
   * **State Evaluation:**
     * For `<input>` elements: check the `checked` property.
     * For ARIA elements: check if the `aria-checked` attribute equals `"true"`.
   * **Radio Buttons:** Only support setting to `true`. Unchecking a native radio button directly is not a standard user interaction. Throw an error if `value === false` is passed for a radio button.

# Testing Requirements
Add new tests to Puppeteer's test suite (likely in `test/src/locators.spec.ts` or similar) to verify:
* Checking and unchecking an `<input type="checkbox">` using `fill(true)` and `fill(false)`.
* Turning on and off an element with `role="switch"` and `aria-checked`.
* Selecting an `<input type="radio">` using `fill(true)`.
* Handling of "mixed" / indeterminate states (ensuring the retry mechanism reaches the final state).
* Error cases for invalid inputs (e.g., passing a string to a checkbox, passing `false` to a radio, passing a boolean to a text input).

# Alternatives Considered

* **Using Direct Property Evaluation (`fillDirectly`):**
  * *Idea:* Directly set `element.checked = true` or `element.setAttribute('aria-checked', 'true')` via JavaScript.
  * *Why it was rejected:* Bypasses critical user interaction events (`mousedown`, `mouseup`) and application-level `onClick` handlers. For custom ARIA components, directly setting an attribute risks failing to update the application's internal state.
* **Translating String inputs ("true"/"false"):**
  * *Idea:* Keep the `fill(value: string)` signature and parse `"true"` or `"false"` strings into booleans for toggles.
  * *Why it was rejected:* This overloads the string type and creates ambiguity. Updating the API to accept native booleans is cleaner and provides better type safety and developer experience.

OrKoN:
For custom checkboxes, the checkboxes that change to checked `mixed` on click are not supported.

Closes https://github.com/puppeteer/puppeteer/issues/14938

### DIFF
--- a/packages/puppeteer-core/src/api/locators/locators.ts
+++ b/packages/puppeteer-core/src/api/locators/locators.ts
@@ -411,7 +411,7 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
 
   #fill<ElementType extends Element>(
     this: Locator<ElementType>,
-    value: string,
+    value: string | boolean,
     options?: Readonly<LocatorFillOptions>,
   ): Observable<void> {
     const signal = options?.signal;
@@ -439,22 +439,28 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
               return 'typeable-input';
             }
             if (el instanceof HTMLInputElement) {
-              if (
-                new Set([
-                  'textarea',
-                  'text',
-                  'url',
-                  'tel',
-                  'search',
-                  'password',
-                  'number',
-                  'email',
-                ]).has(el.type)
-              ) {
-                return 'typeable-input';
-              } else {
-                return 'other-input';
+              switch (el.type) {
+                case 'checkbox':
+                case 'radio':
+                  return 'checkable-input';
+                case 'text':
+                case 'url':
+                case 'tel':
+                case 'search':
+                case 'password':
+                case 'number':
+                case 'email':
+                  return 'typeable-input';
+                default:
+                  return 'other-input';
               }
+            }
+
+            switch (el.getAttribute('role')) {
+              case 'checkbox':
+              case 'radio':
+              case 'switch':
+                return 'checkable-input';
             }
 
             if (el.isContentEditable) {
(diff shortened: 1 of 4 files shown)


packages/puppeteer-core/src/api/locators/locators.ts
```diff
@@ -452,9 +452,16 @@ export abstract class Locator<T> extends EventEmitter<LocatorEvents> {
                 ]).has(el.type)
               ) {
                 return 'typeable-input';
-              } else {
-                return 'other-input';
               }
+              if (el.type === 'checkbox' || el.type === 'radio') {
+                return 'checkable-input';
+              }
+              return 'other-input';
```
zyzyzyryxy: Let's change this series of if's into a switch statement for better readability (and get rid of that `new Set` along the way)
```suggestion
              switch (el.type) {
                case 'textarea':
                case 'text':
                case 'url':
                case 'tel':
                case 'search':
                case 'password':
                case 'number':
                case 'email':
                  return 'typeable-input';
                case 'checkbox':
                case 'radio':
                  return 'checkable-input';
                default:
                  return 'other-input';
```

test/src/locator.spec.ts
```diff
@@ -590,6 +590,129 @@ describe('Locator', function () {
...
+          return (
+            document
+              .querySelector('[role="radio"]')
+              ?.getAttribute('aria-checked') === 'true'
+          );
+        }),
+      ).toBe(true);
+    });
+
+    it('should work for custom ARIA switches', async () => {
```
zyzyzyryxy, line 682: Let's add a test for aria-checked='mixed' (or just start this one with mixed and add steps appropriately)

zyzyzyryxy (approved):

packages/puppeteer-core/src/api/locators/locators.ts
zyzyzyryxy: This suggestion didn't land on appropriate lines, you need to redo it, sorry.

zyzyzyryxy (approved):

PR #14941, opened 2026-05-05
docs: add scraping patterns and common pitfalls guide

MariusYvard:
## Summary

Adds a new guide `docs/guides/scraping-pitfalls.md` covering four practical patterns that come up in real-world Puppeteer usage but are not currently documented:

1. **`evaluate` vs `evaluateHandle`** — why `evaluateHandle` can hang on unresponsive pages and bypass configured timeouts, and when to use each.
2. **Mark-then-query pattern** — a reliable way to get an `ElementHandle` for an element you found inside `page.evaluate()`, without relying on `evaluateHandle`.
3. **Filtering leaf elements** — removing ancestor nodes when multiple levels of the DOM hierarchy match the same selector.
4. **Auth wall detection** — checking for silent redirects to login pages that return HTTP 200.
5. **Defensive `setDefaultTimeout`** — tightening the timeout after navigating to a potentially stuck page in long-running scraping loops.

## Motivation

The `evaluateHandle` hang issue in particular is a source of confusion because it silently bypasses `defaultTimeout`. The current `javascript-execution.md` guide explains *how* to use handles but does not mention this timeout behavior difference.

All patterns are derived from common failure modes when running Puppeteer against real-world pages with auth walls, dynamic DOM structures, or intermittent unresponsiveness.

## Checklist

- [x] New `.md` guide in `docs/guides/`
- [x] TypeScript code examples consistent with existing guide style
- [x] No new runtime code — documentation only

### DIFF
--- a/docs/guides/scraping-pitfalls.md
+++ b/docs/guides/scraping-pitfalls.md
@@ -1,0 +1,148 @@
+# Scraping patterns and common pitfalls
+
+This guide covers practical patterns for building reliable scrapers with
+Puppeteer, based on behaviors that are easy to miss until they cause production
+issues.
+
+## `evaluate` vs `evaluateHandle` — choosing the right tool
+
+[`page.evaluate()`](https://pptr.dev/api/puppeteer.page.evaluate) serializes
+its return value to JSON before handing it back to your script. This has one
+important consequence: if the page becomes unresponsive (for example, it is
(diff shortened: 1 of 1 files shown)


nattallius:
Hi, thank you for your contribution. But scraping is not what we support

PR #14945, opened 2026-05-06
perf: optimize url blocking on navigation

nattallius:
This PR optimizes `allowlist` and `blocklist` checks by throwing an error immediately in `goto` if the URL is blocked, instead of waiting for the browser to fail the request. This saves a network roundtrip.

### DIFF
--- a/packages/puppeteer-core/src/cdp/TargetManager.ts
+++ b/packages/puppeteer-core/src/cdp/TargetManager.ts
@@ -347,7 +347,7 @@ export class TargetManager
 
     // If we connect to a browser that is already open,
     // immediately detach from any tab that is on the blocklist.
-    if (!this.#initialAttachDone && !this.#isUrlAllowed(targetInfo.url)) {
+    if (!this.#initialAttachDone && !this.isUrlAllowed(targetInfo.url)) {
       await this.#silentDetach(session, parentSession);
       return;
     }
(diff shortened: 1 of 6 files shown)


OrKoN (approved):

packages/puppeteer-core/src/cdp/TargetManager.ts
```diff
@@ -482,7 +482,18 @@ export class TargetManager
     }
 
     // Always allow internal or setup pages
-    if (!url || url === 'about:blank') {
+    const ALLOWED_URLS = ['about:blank'];
+    const ALLOWED_SCHEMES = ['chrome:'];
```
OrKoN: why do we need to allow the chrome:// scheme?

nattallius: Why not? 
There are plenty urls like `chrome://flags/` or `chrome://extensions/` 
Do you think it's worth blocking it also and leave only `version`?

OrKoN: I think we should follow what allowlist/blocklist says for those

nattallius: ok, but should we still make an exception for `chrome://version`? Because it was previously not blocked by CDP throttling but now we block it on navigation and this [test](https://github.com/puppeteer/puppeteer/blob/main/test/src/cdp/network_restrictions.spec.ts#L391) is not passing

OrKoN: I think we should not make an exception. For instance, `window.location.href = 'chrome://version'` would not work already (same for iframes and subresouces or window.open) so we are kind of fixing the limitation of the network based throttling by not making an exception.

nattallius: Ok, updated the test, please, take a look.

OrKoN (approved):

PR #14955, opened 2026-05-10
fix: reject BiDi URL restrictions

cuyua9:
### Summary
- Add a shared validation helper for URL restriction options.
- Reject `blocklist`/`allowlist` when `protocol: 'webDriverBiDi'` is used for `puppeteer.connect`.
- Apply the same validation after Firefox launch defaults to WebDriver BiDi.

### Why
URL restrictions are enforced by the CDP target/navigation path. Without an early validation error, BiDi connections and Firefox's default BiDi launch path can accept `blocklist` or `allowlist` even though those restrictions are not enforced there.

### Validation
- `node --test --test-reporter=spec packages/puppeteer-core/lib/cjs/puppeteer/common/BrowserConnector.test.js packages/puppeteer-core/lib/cjs/puppeteer/node/FirefoxLauncher.test.js`

### DIFF
--- a/packages/puppeteer-core/src/common/BrowserConnector.test.ts
+++ b/packages/puppeteer-core/src/common/BrowserConnector.test.ts
@@ -21,5 +21,29 @@ describe('BrowserConnector', () => {
         }),
       ).rejects.toThrow('Cannot specify both blocklist and allowlist');
     });
+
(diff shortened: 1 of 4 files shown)


cuyua9:
Closing this PR to resubmit it from the GitHub account that is already linked to my Google CLA identity.

Issue #14342, opened 2025-10-18
Puppeteer v25


OrKoN:
- [x] https://github.com/puppeteer/puppeteer/pull/13804
- [x] https://github.com/puppeteer/puppeteer/pull/13865
- [x] Drop Node 18 support (https://github.com/puppeteer/puppeteer/pull/14364)
- [x] Investigate dropping CJS builds [Node v20.19](https://nodejs.org/en/blog/release/v20.19.0#2025-03-13-version-20190-iron-lts-marco-ippolito)
  - [x] Introduce smoke test from CJS as we can't use top-level awaits
- [x] Switch to ESM by default (https://github.com/puppeteer/puppeteer/pull/14355)
- [x] https://github.com/puppeteer/puppeteer/pull/14492
- [x] https://github.com/puppeteer/puppeteer/pull/14545
- [x] https://github.com/puppeteer/puppeteer/issues/14817
- [x] https://github.com/puppeteer/puppeteer/pull/14894

OrKoN:
@Lightning00Blade @sadym-chromium here is the list of features we should include in the next breaking change release (v25).

OrKoN:
I am thinking perhaps we just land PRs that are already available, node 18, proxy agent and release v25? (leaving the rest for v26). WDYT @Lightning00Blade

OrKoN:
Split out what I think we can postpone to v26 https://github.com/puppeteer/puppeteer/issues/14898

Lightning00Blade:
All looks good. Please ping me if any of my PRs need updating.

OrKoN:
I hoped you can pick up this task in the next milestone

bershanskiy:
@OrKoN I created #14910 to remove deprecated `Browser.isConnected()`, please consider bringing this task into v25 release. Thanks!

a-kriya:
Is there any rough timeline on when this is slated for?

OrKoN:
Node 20 has also reached EOL (https://nodejs.org/en/download/archive/v20.20.2). Should we drop node 20 as well?

unilynx:
eagerly awaiting Puppeteer v25 :-)

it seems an attempt was made to publish it yesterday https://github.com/puppeteer/puppeteer/actions/runs/25746713498/job/75611822843#step:6:940 - but it's not actually appearing on NPM?

OrKoN:
[...]

PR #14960, opened 2026-05-11
fix!: replace extract-zip with OS tar.exe or unzip

### LINKED ISSUE #14957: [Bug]: @puppeteer/browsers silently corrupts Chrome cache on Node.js 26 (extract-zip 2.0.1)
### Steps to reproduce

Running any `npm install` of a package that depends on `@puppeteer/browsers` on **Node.js 26** silently produces a corrupted Chrome cache: the zip is downloaded, extraction stops after 7 of 308 files (no `chrome` binary), no error is thrown, the install promise resolves, `npm` exits 0. The next `puppeteer.launch()` then fails with:

```
Error: ERROR: Failed to set up chrome v148.0.7778.97!
- DefaultProvider: The browser folder (.../chrome/linux-148.0.7778.97) exists but the executable (.../chrome-linux64/chrome) is missing
```

Because the partial folder exists, every subsequent `npm ci` skips re-installation, so the failure is sticky until the cache is wiped manually.

The root cause is the transitive dependency [`extract-zip@2.0.1`](https://www.npmjs.com/package/extract-zip), which silently aborts mid-stream on Node 26 without rejecting. `extract-zip` has been frozen at 2.0.1 since 2020 — see [max-mapper/extract-zip#139](https://github.com/max-mapper/extract-zip/issues/139) ("the extract method does not return an error event if the extraction fails", open since May 2023).

**Pure `extract-zip` repro (no puppeteer needed) — bug observable in isolation:**

```bash
mkdir -p /tmp/repro && docker run --rm -v /tmp/repro:/work -w /work node:26 bash -c '
  npm init -y >/dev/null
  npm install --silent extract-zip@2.0.1
  curl -sL https://storage.googleapis.com/chrome-for-testing-public/148.0.7778.97/linux64/chrome-linux64.zip -o chrome.zip
  mkdir -p out
  node -e "import(\"extract-zip\").then(m => m.default(process.cwd() + \"/chrome.zip\", { dir: process.cwd() + \"/out\" })).then(() => console.log(\"DONE\")).catch(e => console.error(e))"
  echo "Files: $(find out -type f | wc -l)"
  echo "Chrome binary: $([ -f out/chrome-linux64/chrome ] && echo yes || echo no)"
'
```

**Versions tested** (same zip, same command, only Node version varies):

| Node | Files extracted | `chrome` binary | Result |
|------|-----------------|-----------------|--------|
| 22.22.2 | 303 | yes | OK |
| 24.15.0 | 303 | yes | OK |
| 25.9.0  | 303 | yes | OK |
| **26.1.0** | **7** | **no** | **silently broken** |

Bug appears at the Node 25 → 26 boundary.

### Expected results

Either Chrome is fully extracted, or installation throws/rejects loudly so `npm install` fails. CI builds on Node 26 should never silently ship a broken `puppeteer` install.

A more durable fix would be to migrate `@puppeteer/browsers` away from `extract-zip` (frozen since 2020) toward a maintained alternative: [`yauzl-promise`](https://www.npmjs.com/package/yauzl-promise), [`node-stream-zip`](https://www.npmjs.com/package/node-stream-zip), or system `unzip` via `child_process`.

### Actual results

On `node:26` (`node:26.1.0` Docker image, Linux x86_64):
```
Files: 7
Chrome binary: no
```
No `DONE` line, no error printed — the process exits 0.

On `node:25` (same command, same zip):
```
DONE
Files: 303
Chrome binary: yes
```

**Workaround:** pin Node ≤ 25 until `@puppeteer/browsers` no longer depends on `extract-zip`.

OrKoN:
We expect tar.exe and unzip to be available on all supported OS. In stripped-down Docker images, make sure you install unzip in your Dockerfile.

Closes https://github.com/puppeteer/puppeteer/issues/14957
Related https://github.com/puppeteer/puppeteer/issues/13552

### DIFF
--- a/packages/browsers/src/fileUtil.ts
+++ b/packages/browsers/src/fileUtil.ts
@@ -5,15 +5,17 @@
  */
 
 import type {ChildProcessByStdio} from 'node:child_process';
-import {spawnSync, spawn} from 'node:child_process';
+import {spawnSync, spawn, execFile} from 'node:child_process';
 import {createReadStream} from 'node:fs';
 import {mkdir, readdir} from 'node:fs/promises';
 import * as path from 'node:path';
 import type {Readable, Transform, Writable} from 'node:stream';
 import {Stream} from 'node:stream';
+import {promisify} from 'node:util';
 
 import debug from 'debug';
 
+const execFileAsync = promisify(execFile);
 const debugFileUtil = debug('puppeteer:browsers:fileUtil');
 
 /**
@@ -27,8 +29,8 @@ export async function unpackArchive(
     folderPath = path.resolve(process.cwd(), folderPath);
   }
   if (archivePath.endsWith('.zip')) {
-    const extractZip = await import('extract-zip');
-    await extractZip.default(archivePath, {dir: folderPath});
+    await mkdir(folderPath, {recursive: true});
+    await extractZip(archivePath, folderPath);
   } else if (archivePath.endsWith('.tar.bz2')) {
     await extractTar(archivePath, folderPath, 'bzip2');
   } else if (archivePath.endsWith('.dmg')) {
@@ -179,5 +181,35 @@ async function installDMG(dmgPath: string, folderPath: string): Promise<void> {
     spawnSync('cp', ['-R', mountedPath, folderPath]);
   } finally {
     spawnSync('hdiutil', ['detach', mountPath, '-quiet']);
+  }
+}
+
+/**
+ * @internal
+ */
+async function extractZip(
+  archivePath: string,
+  folderPath: string,
+): Promise<void> {
+  try {
+    if (process.platform === 'win32') {
+      // -x: extract files
+      // -f: specify the archive file
+      // -C: extract to the specified directory
+      await execFileAsync('tar.exe', ['-xf', archivePath, '-C', folderPath]);
+    } else {
+      // -o: overwrite existing files without prompting
+      // -d: extract files into the specified directory
+      await execFileAsync('unzip', ['-o', archivePath, '-d', folderPath]);
+    }
+  } catch (error: any) {
+    if (error?.code === 'ENOENT') {
+      throw new Error(
+        `Extraction failed: Required native binary ('tar.exe' or 'unzip') was not found in the system PATH.`,
+      );
+    }
+    throw new Error(
+      `Extraction failed: ${error?.stderr?.toString() || error?.message}`,
+    );
   }
 }
--- a/.github/workflows/changed-packages.yml
+++ b/.github/workflows/changed-packages.yml
@@ -44,6 +44,7 @@ jobs:
             puppeteer:
               - '.github/workflows/ci.yml'
               - 'packages/browsers/src/browser-data/firefox.ts'
+              - 'packages/browsers/src/fileUtil.ts'
               - 'packages/puppeteer/**'
               - 'packages/puppeteer-core/**'
               - 'packages/testserver/**'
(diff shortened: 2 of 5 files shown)


Lightning00Blade (approved):

Lightning00Blade:
Should we mark this a breaking change?

OrKoN:
> Should we mark this a breaking change?

Would you be able to tackle the rest of https://github.com/puppeteer/puppeteer/issues/14342 today (or this week)?

OrKoN:
I am not sure, I expect the change would be non-breaking for many users and the functionality remains the same. I would include it into v25 but then it would be more difficult for Node 26 users to get the fix because they would also need to upgrade for other changes. At the same time, it might be fine because users can still downgrade to Node 24.

packages/browsers/src/fileUtil.ts
```diff
@@ -181,3 +183,33 @@ async function installDMG(dmgPath: string, folderPath: string): Promise<void> {
...
+async function extractZip(
+  archivePath: string,
+  folderPath: string,
+): Promise<void> {
+  try {
+    if (process.platform === 'win32') {
+      // -x: extract files
+      // -f: specify the archive file
+      // -C: extract to the specified directory
+      await execFileAsync('tar.exe', ['-xf', archivePath, '-C', folderPath]);
```
bcurtin144, line 199: Why is tar.exe being called on Windows to unzip a zip file? I'm running Windows Server 2016 (yes, I know it's old), which doesn't have a native tar.exe. But when I use the version included with GitSCM (C:\Program Files\Git\usr\bin) on the command line: `tar.exe -xf .\chrome-win64.zip -C .\chrome-tar` this returns the following error:
```sh
/usr/bin/tar: This does not look like a tar archive
/usr/bin/tar: Skipping to next header
/usr/bin/tar: A lone zero block at 244440
/usr/bin/tar: Exiting with failure status due to previous errors
```

Running with `unzip` works correctly.

OrKoN: I believe the Windows tar.exe supports zip files. GNU tar on the other hand does not (Windows' tar.exe is not the same as the one ships with Git).

OrKoN: I think we should support Windows Server 2016 as well but it is not easy to test there. We could fallback to unzip if windows' tar.exe is not available but that would fail if not run via Git bash I believe.

OrKoN: Could you try to install bsdtar.exe from https://github.com/libarchive/libarchive and place it under System32\tar.exe? I believe bsdtar is what recent Windows versions ship.

bcurtin144: Yes, I checked a non-Server version of Windows, and the tar.exe in Windows\System32 is identified as "bsdtar archive tool" from libarchive. Their documentation says it supports ZIP files, so it makes sense that it would work for Windows Server 2019 and up.

For the life of me, I can't find any pre-built binary from the GitHub page, and I don't have a quick way to build it myself at the moment. I'll probably end up copying the tar.exe to the server and using it that way.

If I could suggest modifying the documentation to say that the browser install on Windows looks explicitly (since #14997) for Windows\System32\tar.exe (not in the system PATH, as it says in the error message) and it must be the BSD Tar executable.

If there's also a fallback to use unzip if tar.exe isn't found or doesn't work, that would be excellent. I didn't test it from a Node process, but I had success using the GitCSM unzip from the command line and not needing Git Bash. 

Thanks.

OrKoN: I have created an issue https://github.com/puppeteer/puppeteer/issues/15024 but not sure when this will happen. Testing on Windows is somewhat difficult due to the matrix of various versions and git bash/cmd.exe/powershell usage.

PR #14975, opened 2026-05-12
fix: Handle missing sessionId when looking up callbacks

wolfib:
If a CDP response does not contain a `sessionId`, it is currently handled by the default session. This change adds a lookup by callback id, which allows to correctly map the CDP response to its session.

This is (part of) a fix for https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1797


packages/puppeteer-core/src/cdp/Connection.ts
```diff
@@ -210,18 +210,27 @@ export class Connection extends EventEmitter<CDPSessionEvents> {
...
+              object.error.message,
+            );
+          }
         } else {
-          this.#callbacks.reject(
-            object.id,
-            createProtocolErrorMessage(object),
-            object.error.message,
-          );
+          this.#callbacks.resolve(object.id, object.result);
```
OrKoN, line 225: should this be reject?

OrKoN: nevermind

OrKoN (approved):

PR #14976, opened 2026-05-12
fix: block standard network emulation reset if allowlist or blocklist are enabled

nattallius:


test/src/cdp/network_restrictions.spec.ts
```diff
@@ -518,4 +518,49 @@ describe('Network Restrictions', function () {
...
+        }),
+      ).rejects.toThrow(
+        'Cannot reset network conditions: rule-based emulation is enabled.',
+      );
+    } finally {
+      await close();
+    }
+  });
+
+  it('should allow standard emulation reset when no blocklist/allowlist is active', async () => {
```
OrKoN: I think this test can be removed we already have tests that https://pptr.dev/api/puppeteer.page.emulatenetworkconditions works.

test/src/cdp/network_restrictions.spec.ts
```diff
@@ -518,4 +518,49 @@ describe('Network Restrictions', function () {
...
+      {
+        blocklist: ['*://*:*/empty.html'],
+      },
+      {createContext: true},
+    );
+
+    try {
+      const session = await page.createCDPSession();
+
+      await expect(
```
OrKoN, line 524: let's also add expectation next to this one that calling https://pptr.dev/api/puppeteer.page.emulatenetworkconditions also fails

test/src/cdp/network_restrictions.spec.ts
```diff
@@ -518,4 +518,49 @@ describe('Network Restrictions', function () {
       await close();
     }
   });
+
+  it.only('should block standard emulation reset when blocklist/allowlist is active', async () => {
```
OrKoN:
```suggestion
  it('should block standard emulation reset when blocklist/allowlist is active', async () => {
```

packages/puppeteer-core/src/cdp/Connection.ts
```diff
@@ -40,6 +40,7 @@ export class Connection extends EventEmitter<CDPSessionEvents> {
   #sessions = new Map<string, CdpCDPSession>();
   #closed = false;
   #manuallyAttached = new Set<string>();
+  #ruleBasedEmulationConfigured = false;
```
OrKoN:
```suggestion
  #rejectEmulateNetworkConditionsCalls = false;
```

I think it would be a better name here.

docs/api/puppeteer.connection.md
```diff
@@ -61,6 +61,19 @@ Description
 </th></tr></thead>
 <tbody><tr><td>
 
+<span id="rejectemulatenetworkconditionscalls">rejectEmulateNetworkConditionsCalls</span>
```
OrKoN: please mark this with `@internal` in code, we do not want to make this public

OrKoN (approved):

PR #14998, opened 2026-05-17
chore(deps): update to tsc 6

OrKoN:


mathiasbynens (approved):

Lightning00Blade (approved):
LGTM, 
I think you missed adding the types to the @puppeteer/browsers package that is why it's failing.

OrKoN:
Somehow could not get rid of a stale cache locally.

docs/api/puppeteer.cookiesamesite_2.md
```diff
@@ -1,8 +1,8 @@
 ---
-sidebar_label: CookieSameSite
+sidebar_label: CookieSameSite_2
 ---
 
-# CookieSameSite type
+# CookieSameSite_2 type
```
OrKoN, line 5: not sure why this change got generated

Lightning00Blade: My guess is that we somehow pull the one from protocol so now there are two it can resolve.

OrKoN: hm but I have not changed any related exports it seems. I would suggest we land the update since the change seems to be harmless?

Lightning00Blade: Agree

PR #15002, opened 2026-05-18
fix: Throw TargetCloseError when session ID not found

wolfib:
This addresses https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/1797

When a parent session is detached or closed, any child target sessions attached underneath it are implicitly destroyed by the browser. Because the parent session is detached first, Puppeteer's `Connection` never receives or processes a `Target.detachedFromTarget` event specifically for the child session. This leaves the child session orphaned in the Connection's active `#sessions` map, still marked as active/attached (`session.detached === false`).

When Puppeteer attempts to perform global operations across all active clients—such as applying network conditions or disabling cache—it sends a CDP command to all sessions tracked in the map. The command sent to the orphaned session fails with: `Error: Protocol error (Network.emulateNetworkConditions): Session with given id not found`.

Since `#canIgnoreError` in `NetworkManager.ts` does not recognize this error string, Puppeteer throws a `ProtocolError`, causing the entire execution (such as Lighthouse audits or page throttling actions) to crash.

This fix makes it so that a `TargetCloseError` is thrown, which is ignored in `NetworkManger.ts`.

### DIFF
--- a/packages/puppeteer-core/src/cdp/CdpSession.ts
+++ b/packages/puppeteer-core/src/cdp/CdpSession.ts
@@ -89,6 +89,13 @@ export class CdpCDPSession extends CDPSession {
     return parent ?? undefined;
   }
 
+  /**
+   * @internal
+   */
+  getParentSessionId(): string | undefined {
+    return this.#parentSessionId;
+  }
+
   override send<T extends keyof ProtocolMapping.Commands>(
     method: T,
     params?: ProtocolMapping.Commands[T]['paramsType'][0],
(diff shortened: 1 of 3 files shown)


packages/puppeteer-core/src/cdp/CdpSession.ts
```diff
@@ -101,15 +101,20 @@ export class CdpCDPSession extends CDPSession {
...
       );
     }
-    return this.#connection._rawSend(
-      this.#callbacks,
-      method,
-      params,
-      this.#sessionId,
-      options,
-    );
+    return this.#connection
```
OrKoN, line 104: Can we clean up child sessions instead?

packages/puppeteer-core/src/cdp/CdpSession.ts
```diff
@@ -101,15 +101,20 @@ export class CdpCDPSession extends CDPSession {
...
+      ._rawSend(this.#callbacks, method, params, this.#sessionId, options)
+      .catch(error => {
+        if (
+          error instanceof Error &&
+          error.message.includes('Session with given id not found')
+        ) {
+          throw new TargetCloseError(
```
OrKoN, line 112: can we call this.onClosed() here as well?

packages/puppeteer-core/src/cdp/CdpSession.ts
OrKoN: Discussed offline: this would require changing how we track relationship between sessions.

OrKoN (approved):

PR #15012, opened 2026-05-20
feat(browsers): report unpacking status during install

### LINKED ISSUE #13083: [Feature]: Add unpacking indicator to the install command

Caltsic:
## Summary

Adds an install status reporter for the browser unpacking phase so the CLI no longer appears idle after download completes.

## Details

- Derives an internal status reporter from the existing `downloadProgressCallback: 'default'` path instead of adding a new public install option.
- Reports unpacking, setup, and dependency-install phases through stderr only for TTY output.
- Keeps library consumers unchanged unless they already opt into the default CLI progress reporter.
- Adds a focused CLI regression test for the unpacking status.
- Updates the generated browsers API docs for the adjusted progress callback wording.

Closes #13083

## Validation

- `npm run build --workspace @puppeteer/browsers`
- `npm run build:test --workspace @puppeteer/browsers`
- `npx mocha --config .mocharc.cjs test/build/CLI.spec.js --grep "should report unpacking status" --full-trace --reporter spec`
- `npm run build:docs --workspace @puppeteer/browsers`
- `git diff --check`

### DIFF
--- a/docs/browsers-api/browsers.installoptions.md
+++ b/docs/browsers-api/browsers.installoptions.md
@@ -140,7 +140,7 @@ Determines the path to download browsers to.
 
 </td><td>
 
-Provides information about the progress of the download. If set to 'default', the default callback implementing a progress bar will be used.
+Provides information about the progress of the download and install. If set to 'default', the default terminal reporter will be used.
 
 </td><td>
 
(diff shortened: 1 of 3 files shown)


Lightning00Blade:
Thanks for the PR, but the feature was already added by https://github.com/puppeteer/puppeteer/pull/15045

PR #15023, opened 2026-05-21
refactor: remove debug dependency

OrKoN:
The built-in output is not so configurable but I think it does the job and allows us to drop a dependency.


packages/browsers/package.json
Lightning00Blade, line 101: Side question but I don't think we need this for the browsers package.

Lightning00Blade (approved):
LGTM, also gave me an idea how to get some perf by removing creating temp objects for logging.

packages/browsers/package.json
OrKoN: will follow-up

PR #15026, opened 2026-05-21
refactor: remove progress dependency

OrKoN:
Implements just the things we need. Drive-by: removes ws types.

### DIFF
--- a/packages/browsers/package.json
+++ b/packages/browsers/package.json
@@ -83,7 +83,6 @@
     "!*.tsbuildinfo"
   ],
   "dependencies": {
-    "progress": "^2.0.3",
     "tar-fs": "^3.1.1",
     "yargs": "^17.7.2",
     "semver": "^7.7.4"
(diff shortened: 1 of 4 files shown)


packages/browsers/package.json
```diff
@@ -52,7 +52,7 @@
       ]
     },
     "test": {
-      "command": "node tools/downloadTestBrowsers.mjs && mocha",
+      "command": "node tools/downloadTestBrowsers.mjs && NODE_DEBUG='puppeteer:*' mocha",
```
OrKoN: oh bad rebase

packages/browsers/src/ProgressBar.ts
```diff
@@ -0,0 +1,68 @@
+/**
+ * @license
+ * Copyright 2026 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+export interface ProgressBarOptions {
```
Lightning00Blade, line 10: This needs to be internal else docs extractor throws warning.

Lightning00Blade (approved):

PR #15027, opened 2026-05-21
feat: expose Explicit Resource Management (`using`)

Lightning00Blade:
The Explicit Resource Management reached stage 4 with multiple vendors already supporting it (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/using#browser_compatibility).

Once https://github.com/tc39/ecma262/pull/3000 is merged and part of the spec, we should make this publicly available.

Note: NodeJS v22 does not support the keyword `using`.

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -1546,11 +1546,25 @@ Description
 </th></tr></thead>
 <tbody><tr><td>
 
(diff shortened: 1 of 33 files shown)


OrKoN:
Looks like the formatting fails.

OrKoN (approved):

PR #15033, opened 2026-05-21
fix: use powershell as fallback to unzip on Windows

### LINKED ISSUE #15024: Consider offering fallbacks to unzip on Windows and document tar.exe path
Yes, I checked a non-Server version of Windows, and the tar.exe in Windows\System32 is identified as "bsdtar archive tool" from libarchive. Their documentation says it supports ZIP files, so it makes sense that it would work for Windows Server 2019 and up.

For the life of me, I can't find any pre-built binary from the GitHub page, and I don't have a quick way to build it myself at the moment. I'll probably end up copying the tar.exe to the server and using it that way.

If I could suggest modifying the documentation to say that the browser install on Windows looks explicitly (since #14997) for Windows\System32\tar.exe (not in the system PATH, as it says in the error message) and it must be the BSD Tar executable.

If there's also a fallback to use unzip if tar.exe isn't found or doesn't work, that would be excellent. I didn't test it from a Node process, but I had success using the GitCSM unzip from the command line and not needing Git Bash. 

Thanks.

_Originally posted by @bcurtin144 in https://github.com/puppeteer/puppeteer/pull/14960#discussion_r3280257749_

OrKoN:
Closes https://github.com/puppeteer/puppeteer/issues/15024

### DIFF
--- a/docs/guides/system-requirements.md
+++ b/docs/guides/system-requirements.md
@@ -14,7 +14,7 @@
     - Required system packages https://source.chromium.org/chromium/chromium/src/+/main:chrome/installer/linux/debian/dist_package_versions.json
   - [openSUSE/Fedora Linux](https://support.google.com/chrome/a/answer/7100626?hl=en#:~:text=10.15%20or%20later-,Linux,-To%20use%20Chrome), with x64 architecture
     - Required system packages https://source.chromium.org/chromium/chromium/src/+/main:chrome/installer/linux/rpm/dist_package_provides.json
-  - `tar.exe` on Windows and `unzip` on MacOS/Linux for unpacking Chrome for Testing binaries.
+  - `tar.exe` or Powershell on Windows and `unzip` on MacOS/Linux for unpacking Chrome for Testing binaries.
 
 - Firefox browser system requirements:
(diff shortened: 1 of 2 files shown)


bcurtin144:
I removed tar.exe and archiveint.dll from my Windows Server 2016, ran the installation and the PowerShell command worked, which is great. On the downside, the original PowerShell (v5.1 on this machine) increased the unzip time by about 4x. It took around 17 seconds, which isn't a dealbreaker, but it was a noticeable slowdown. I made a fork of the code and tested further with PowerShell v7, which runs the same command better, but still almost 2x slower than tar.exe on the machine.

If you'd like to see my edits, I added a small utility function to try running a given command to see if it exists on the server, and I ordered the non-Windows commands as tar.exe, PowerShell v7 and Original PowerShell:
https://github.com/puppeteer/puppeteer/commit/718cdad9585f589204d59de21537ec6dbd1df630

OrKoN:
Thanks for testing. I will update the PR to use tar.exe first and then use pwsl.exe and powershell.exe fallbacks

Lightning00Blade (approved):

PR #15034, opened 2026-05-21
feat: support --disallow-code-generation-from-strings

legrego:

Changes the way functions are serialized, so they do not require the use of `new Function`. This supports running Puppeteer within a Node.js environment configured with [`--disallow-code-generation-from-strings`](https://nodejs.org/docs/latest/api/cli.html#disallow-code-generation-from-strings), which is an important security hardening measure.

This might not be the correct Conventional Commit; I am not well-versed in using this nomenclature. Feel free to suggest something other than `feat`.

Yes

N/A (I think)

**Summary**

This change allows Puppeteer to run within a Node.js environment which has been configured to [disallow code generation from strings](https://nodejs.org/docs/latest/api/cli.html#disallow-code-generation-from-strings). This is a defense-in-depth feature of the Node runtime which blocks the use of `eval`, `new Function`, and related features which convert strings into executable code. 

I do not believe so.

**Other information**

I am verifying this functionality for Kibana via https://github.com/elastic/kibana/pull/241162, using `patch-package` to apply this change manually against our own test suites.

### DIFF
--- a/packages/puppeteer-core/src/util/Function.test.ts
+++ b/packages/puppeteer-core/src/util/Function.test.ts
@@ -8,9 +8,33 @@ import {describe, it} from 'node:test';
 
 import expect from 'expect';
 
-import {interpolateFunction, stringifyFunction} from './Function.js';
+import {
+  createFunction,
+  interpolateFunction,
+  stringifyFunction,
+} from './Function.js';
 
 describe('Function', function () {
+  describe('createFunction', function () {
+    it('should return a placeholder that is not directly callable', () => {
+      const fn = createFunction('() => 42');
+      expect(() => fn()).toThrow(/serialization placeholder/);
+    });
+
+    it('should serialize back to the original source via toString', () => {
+      const source = '() => 42';
+      const fn = createFunction(source);
+      expect(fn.toString()).toBe(source);
+    });
+
+    it('should be evaluable into a real function via toString', () => {
+      const source = '() => 42';
+      const fn = createFunction(source);
+      const real = new Function(`return ${fn.toString()}`)() as () => number;
+      expect(real()).toBe(42);
+    });
+  });
+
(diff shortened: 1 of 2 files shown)


OrKoN (changes requested):
I believe this would not work because this code also runs in the browser. Note that createFunction in node is only used in interpolateFunction that only operates on stringified functions that were evaluated already by the caller so there is no security impact.

wahajahmed010:
Hi @OrKoN,

I've updated the PR with a fix that addresses your feedback about browser compatibility.

**The Solution:**
Instead of completely replacing `new Function()`, the implementation now uses feature detection at module initialization:

1. We check if `new Function()` is allowed when the module loads
2. If allowed (normal case), we use `new Function()` — this works in both Node.js and browsers
3. If blocked by `--disallow-code-generation-from-strings`, we fall back to the placeholder approach

**Key Benefits:**
- ✅ Works in browsers where `new Function()` is needed for actual function execution
- ✅ Supports Node.js with `--disallow-code-generation-from-strings` flag
- ✅ Zero API changes — fully backward compatible
- ✅ No runtime performance impact (feature detection happens once at module load)

**Files Changed:**
- `packages/puppeteer-core/src/util/Function.ts` — Added feature detection and conditional logic
- `packages/puppeteer-core/src/util/Function.test.ts` — Updated tests to handle both scenarios

This approach gives us the best of both worlds: real function execution where it's available, and graceful degradation where it's not.

Let me know if this addresses your concerns!

legrego:
@OrKoN I really appreciate that you took the time to review and respond to my PR. I researched further after reading your comments, and I now understand how this change would not actually help. The error messages I was seeing during my test runs with `--disallow-code-generation-from-strings` led me down an incorrect path.

I'm closing this out - thanks again for your help.

PR #15048, opened 2026-05-23
docs: document read-only Docker directories

### LINKED ISSUE #15041: [Feature]: Documentation update for running in Docker
### Feature description

On https://pptr.dev/troubleshooting#running-puppeteer-in-docker, I think it would be a good idea to add some notes about errors received in read-only containers. There seem to be a few threads about it, but it took me a while to track down.

> When running in a read-only Docker container, there are certain directories that need to be writeable. You may receive an error such as `chrome_crashpad_handler: --database is required`. By default chrome tries to write to `~/.cache` and if this directory is read-only you need to specify an alternative writeable path. If you have, for example, mapped a writeable `/tmp`, you add this to your Dockerfile:
> 
> ```
> XDG_CONFIG_HOME=/tmp/.chromium
> XDG_CACHE_HOME=/tmp/.chromium
> ```

I note this doesn't actually resolve the error `chrome_crashpad_handler: --database is required` as I guess the crash database is no longer required. I did find this additional env variable that **may** resolve this issue but I can't prove it:

```
ENV CHROME_CRASHpad_HANDLER_DATABASE=/tmp/crashpad
```
This was also quite a useful comment that led me to this result:
https://github.com/puppeteer/puppeteer/issues/11023#issuecomment-2079036754

cyphercodes:

Documentation update.

No, this is a docs-only change. I checked the touched Markdown with Prettier and `git diff --check`.

Yes, this updates the Docker troubleshooting documentation.

**Summary**

Adds a short note for read-only Docker/container environments where Chrome still needs writable profile, configuration, and cache directories. It documents setting `XDG_CONFIG_HOME` and `XDG_CACHE_HOME` to a writable location such as `/tmp/.chromium`, and shows how to provide an explicit Puppeteer `userDataDir` if needed.

Closes #15041.

No.

**Other information**

N/A

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -478,6 +478,35 @@ the command:
 There's a full example at https://github.com/ebidel/try-puppeteer that shows how
 to run this Dockerfile from a webserver running on App Engine Flex (Node).
 
+### Running in read-only containers
+
+Chrome writes profile, configuration, and cache files during startup. In a
+read-only container, or in a container that only mounts specific writable
+locations, make sure those paths point to writable directories. Otherwise Chrome
+can fail before Puppeteer connects; one common error is
+`chrome_crashpad_handler: --database is required`.
+
+If your container provides a writable `/tmp`, point Chrome's XDG config and cache
+locations there:
+
+```Dockerfile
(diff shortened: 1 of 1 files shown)


OrKoN (approved):
Looks good to me, thanks!

PR #15060, opened 2026-05-26
chore: validate allow/block list before browser spawn

nattallius:


Lightning00Blade (approved):

Lightning00Blade:
Let's mark it as `chore: validate allow/block list before browser spawn`.

PR #15065, opened 2026-05-27
test: refactor number of spawned browsers

Lightning00Blade:


test/src/cdp/network_restrictions.spec.ts
```diff
@@ -133,55 +108,82 @@ describe('Network Restrictions', function () {
...
 
-      await page.goto(blockedUrl);
+      expect(error).toBeDefined();
+      expect(error?.message).toContain(
+        'is blocked by blocklist/allowlist rules',
+      );
+    });
 
-      const wsEndpoint = originalBrowser.wsEndpoint();
+    it('should block standard emulation reset when blocklist/allowlist is active', async () => {
```
nattallius: I just noticed, that allowlist in the title of this test doesn't actually needed, cause there is no allowlist specified for this test. Let's get rid if it please, since you are refactoring it anyway

nattallius (approved):

test/src/cdp/network_restrictions.spec.ts
Lightning00Blade: Should we duplicate the test for both cases then? instead of just renaming it?

PR #15066, opened 2026-05-27
refactor: sync accessing debugger

Lightning00Blade:
This should improve performance in places where we create objects inside the debug statements.

### DIFF
--- a/packages/puppeteer-core/src/puppeteer-core.ts
+++ b/packages/puppeteer-core/src/puppeteer-core.ts
@@ -8,6 +8,7 @@ export * from './index.js';
 
 import fs from 'node:fs';
 import path from 'node:path';
+import {debuglog} from 'node:util';
 
 import {environment} from './environment.js';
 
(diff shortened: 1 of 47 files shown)


OrKoN (approved):

packages/puppeteer-core/src/puppeteer-core.ts
```diff
@@ -18,6 +19,7 @@ environment.value = {
   fs,
   path,
   ScreenRecorder: Puppeteer.ScreenRecorder,
+  debuglog,
```
OrKoN, line 22: this does not seem to work actually. I think it runs too late

lulzpid (commented):
ss

packages/puppeteer-core/src/puppeteer-core.ts
Lightning00Blade: Can you give a little bit more info how to test or what was seens/expected?

OrKoN: I have more details in https://github.com/puppeteer/puppeteer/issues/15133

PR #15068, opened 2026-05-27
fix(webmcp): invalidate webmcp tools on context destruction

OrKoN:


test/src/cdp/webmcp.spec.ts
```diff
@@ -302,6 +302,89 @@ describe('Page.webmcp', function () {
...
+      const form = document.createElement('form');
+      form.setAttribute('toolname', 'declarative tool name');
+      form.setAttribute('tooldescription', 'tool description');
+      document.body.appendChild(form);
+    });
+    await toolsAddedPromise;
+    expect(page.webmcp.tools().length).toBe(1);
+
+    // Same document/hash navigation should not reset tools.
+    await page.goto(httpsServer.EMPTY_PAGE + '#hash');
```
beaufortfrancois, line 381: Shall we wait for `page.waitForNavigation` as well?

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -379,17 +379,24 @@ export class WebMCP extends EventEmitter<{
     }
   };
 
+  #listenToContextDestroyed(frame: CdpFrame): void {
+    const mainWorld = frame.mainRealm();
```
beaufortfrancois: Will it work for isolated realms as well?

packages/puppeteer-core/src/cdp/WebMCP.ts
```diff
@@ -379,17 +379,24 @@ export class WebMCP extends EventEmitter<{
...
+
+    const context = mainWorld.context;
+    if (context) {
+      context.once('disposed', () => {
+        this.#onContextDisposed(frame);
+      });
+    }
```
beaufortfrancois: Nit:

```suggestion
    mainWorld.context?.once('disposed', () => {
      this.#onContextDisposed(frame);
    });
```

beaufortfrancois (commented):
Thank you so much @OrKoN!

test/src/cdp/webmcp.spec.ts
beaufortfrancois: Nope. Sorry for the noise!

packages/puppeteer-core/src/cdp/WebMCP.ts
OrKoN: No, it won’t but I believe CDP does not support calling into isolated world tools right now anyway

beaufortfrancois (approved):

PR #15070, opened 2026-05-28
fix: only pull yargs in CLI

Lightning00Blade:
This allows bundlers to more easily remove yargs from the a bundled build all together.


packages/browsers/src/main.ts
```diff
@@ -41,7 +41,6 @@ export {
   getVersionComparator,
   resolveDefaultUserDataDir,
 } from './browser-data/browser-data.js';
```
OrKoN, line 43: this was a public API of the browsers package. I think we should keep it so.

PR #15072, opened 2026-05-29
test: attempt fixing _wcsnicmp assertion by downgrading to Node 24.15.0

OrKoN:
Attempt to work around crashes like https://github.com/puppeteer/puppeteer/actions/runs/26626414041/job/78464147790#step:11:505 which started to happen on Windows. Maybe related to https://github.com/libuv/libuv/issues/5010


OrKoN:
The original workaround still caused failures so trying to downgrade node version.

Lightning00Blade (approved):

PR #15075, opened 2026-05-29
feat: add page locale emulation

### LINKED ISSUE #11303: [Feature]: Support locale when starting the puppeteer

jmolz:

Feature

Yes. Added coverage for `page.emulateLocale()` in `test/src/emulation.spec.ts`, and verified it in both Chrome CDP and Chrome WebDriver BiDi modes.

Yes. Regenerated API docs for `Page.emulateLocale()`.

**Summary**

Closes #11303.

Adds a page-level `page.emulateLocale(locale?: string)` API matching Puppeteer's existing emulation methods.

- CDP applies `Emulation.setLocaleOverride` for Intl locale behavior.
- CDP also applies `acceptLanguage` through the existing user-agent override path so `navigator.language` and `navigator.languages` reflect the emulated locale.
- WebDriver BiDi applies `emulation.setLocaleOverride`.
- Calling `page.emulateLocale()` without a locale clears the override.

No.

**Other information**

Validation run locally:

- `npm run docs`
- `./node_modules/.bin/prettier --check docs/api/puppeteer.page.md docs/api/puppeteer.page.emulatelocale.md packages/puppeteer-core/src/api/Page.ts packages/puppeteer-core/src/bidi/Page.ts packages/puppeteer-core/src/bidi/core/BrowsingContext.ts packages/puppeteer-core/src/cdp/EmulationManager.ts packages/puppeteer-core/src/cdp/NetworkManager.ts packages/puppeteer-core/src/cdp/Page.ts test/src/emulation.spec.ts`
- `./node_modules/.bin/eslint packages/puppeteer-core/src/api/Page.ts packages/puppeteer-core/src/bidi/Page.ts packages/puppeteer-core/src/bidi/core/BrowsingContext.ts packages/puppeteer-core/src/cdp/EmulationManager.ts packages/puppeteer-core/src/cdp/NetworkManager.ts packages/puppeteer-core/src/cdp/Page.ts test/src/emulation.spec.ts`
- `BINARY="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" npm --cache /private/tmp/npm-cache test -- --test-suite chrome-headless --grep "Page.emulateLocale"`
- `BINARY="/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" npm --cache /private/tmp/npm-cache test -- --test-suite chrome-bidi --grep "Page.emulateLocale"`

### DIFF
--- a/packages/puppeteer-core/src/cdp/EmulationManager.ts
+++ b/packages/puppeteer-core/src/cdp/EmulationManager.ts
@@ -31,6 +31,11 @@ interface TimezoneState {
   active: boolean;
 }
 
+interface LocaleState {
+  locale?: string;
+  active: boolean;
+}
+
 interface VisionDeficiencyState {
   visionDeficiency?: Protocol.Emulation.SetEmulatedVisionDeficiencyRequest['type'];
   active: boolean;
@@ -147,6 +152,13 @@ export class EmulationManager implements ClientProvider {
     },
     this,
     this.#emulateTimezone,
+  );
+  #localeState = new EmulatedState<LocaleState>(
+    {
+      active: false,
+    },
+    this,
+    this.#emulateLocale,
   );
   #visionDeficiencyState = new EmulatedState<VisionDeficiencyState>(
     {
@@ -370,6 +382,26 @@ export class EmulationManager implements ClientProvider {
   async emulateTimezone(timezoneId?: string): Promise<void> {
     await this.#timezoneState.setState({
       timezoneId,
+      active: true,
+    });
+  }
+
+  @invokeAtMostOnceForArguments
+  async #emulateLocale(
+    client: CDPSession,
+    localeState: LocaleState,
+  ): Promise<void> {
+    if (!localeState.active) {
+      return;
+    }
+    await client.send('Emulation.setLocaleOverride', {
+      locale: localeState.locale || '',
+    });
+  }
+
+  async emulateLocale(locale?: string): Promise<void> {
+    await this.#localeState.setState({
+      locale,
       active: true,
     });
   }
--- a/packages/puppeteer-core/src/cdp/EmulationManager.ts
+++ b/packages/puppeteer-core/src/cdp/EmulationManager.ts
@@ -395,7 +395,7 @@ export class EmulationManager implements ClientProvider {
       return;
     }
     await client.send('Emulation.setLocaleOverride', {
-      locale: localeState.locale || '',
+      locale: localeState.locale,
     });
   }
 
--- a/packages/puppeteer-core/src/cdp/NetworkManager.ts
+++ b/packages/puppeteer-core/src/cdp/NetworkManager.ts
@@ -102,10 +102,15 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
   #clients = new Map<CDPSession, DisposableStack>();
   #networkEnabled = true;
 
-  constructor(frameManager: FrameProvider, networkEnabled?: boolean) {
+  constructor(
+    frameManager: FrameProvider,
+    networkEnabled?: boolean,
+    defaultUserAgent?: string,
+  ) {
     super();
     this.#frameManager = frameManager;
     this.#networkEnabled = networkEnabled ?? true;
+    this.#defaultUserAgent = defaultUserAgent;
   }
 
   #canIgnoreError(error: unknown) {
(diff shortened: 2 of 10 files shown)


packages/puppeteer-core/src/cdp/EmulationManager.ts
```diff
@@ -374,6 +386,26 @@ export class EmulationManager implements ClientProvider {
...
+  @invokeAtMostOnceForArguments
+  async #emulateLocale(
+    client: CDPSession,
+    localeState: LocaleState,
+  ): Promise<void> {
+    if (!localeState.active) {
+      return;
+    }
+    await client.send('Emulation.setLocaleOverride', {
+      locale: localeState.locale || '',
```
Lightning00Blade: Nit: We can omit casting the or as this if omitted or empty it resets.

jmolz:
Quick update on the stalled CLA gate: I confirmed the CLA is signed, rebased this branch onto current `main`, and pushed a signed commit (`d8c94d4`). The new `cla/google` check is now passing, and CI has restarted.

Once the checks settle, I’d appreciate a maintainer review when you have bandwidth.

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -1088,6 +1088,14 @@ export class CdpPage extends Page {
     return await this.#emulationManager.emulateTimezone(timezoneId);
   }
 
+  override async emulateLocale(locale?: string): Promise<void> {
+    await this.#emulationManager.emulateLocale(locale);
+    await this.#frameManager.networkManager.setAcceptLanguage(
+      locale,
+      await this.browser().userAgent(),
```
Lightning00Blade: I think we should move this in the Page creation lifecycle in `CdpPage._create`. 

And as follow up maybe cache the result from the `this.browser().userAgent/version` as they never change.

packages/puppeteer-core/src/cdp/NetworkManager.ts
```diff
@@ -277,13 +279,24 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
...
+  ): Promise<void> {
+    this.#acceptLanguage = acceptLanguage;
+    this.#defaultUserAgent = defaultUserAgent;
+    await this.#applyToAllClients(this.#applyUserAgent.bind(this));
+  }
+
   async #applyUserAgent(client: CDPSession) {
-    if (this.#userAgent === undefined) {
+    const userAgent = this.#userAgent ?? this.#defaultUserAgent;
+    if (userAgent === undefined) {
```
Lightning00Blade, line 294: Is this if now needed? With the default it should always be true value.

Lightning00Blade (commented):
LGTM, there are some nits and one comment we should address before we can merge.

packages/puppeteer-core/src/cdp/EmulationManager.ts
jmolz: Done in 3c3824b: this now passes `localeState.locale` through directly so omitted locale resets via CDP.

packages/puppeteer-core/src/cdp/Page.ts
jmolz: Done in 3c3824b: `_create` now seeds `NetworkManager` with the browser default UA before `#initialize()`, and `emulateLocale()` reuses that value instead of calling `browser().userAgent()` each time. I left the broader `userAgent`/`version` caching as follow-up scope.

packages/puppeteer-core/src/cdp/NetworkManager.ts
jmolz: I kept the guard. In the normal `CdpPage._create` lifecycle, the default UA is now seeded before network clients initialize. `NetworkManager` is also constructed directly in unit-level paths, though, and `addClient()` always calls `#applyUserAgent()`, so this still avoids sending an undefined UA from those lower-level paths.

packages/puppeteer-core/src/cdp/Page.ts
```diff
@@ -121,6 +121,10 @@ export class CdpPage extends Page {
     defaultViewport: Viewport | null,
   ): Promise<CdpPage> {
     const page = new CdpPage(client, target);
+    await page.#frameManager.networkManager.setAcceptLanguage(
```
Lightning00Blade: I don't think we should pass it like this but either though a setter or preferably though the constructor of the NetworkManager

jmolz: Threaded the default user agent through `CdpPage`/`FrameManager` into the `NetworkManager` constructor in 238d4450d, so `setAcceptLanguage` now only updates the locale state.

Lightning00Blade (approved):

Lightning00Blade:
LGTM, thanks!

jmolz:
Appreciate the review. I also pushed a small Prettier-only follow-up in `dc6d097b4`, and the required checks are green now.

Lightning00Blade (approved):

PR #15079, opened 2026-05-31
fix: restore .puppeteerrc config file loading

### LINKED ISSUE #15076: [Bug]: Loading `.puppeteerrc` broken between Puppeteer 25.0.4 and 25.1.0

```TypeScript
git clone https://github.com/mozilla/pdf.js.git
cd pdf.js
git checkout puppeteer-25
rm -rf ~/.cache/puppeteer
npm install
npx puppeteer browsers list -> note that this contains Firefox
mv .puppeteerrc.json .puppeteerrc
rm -rf ~/.cache/puppeteer
rm -rf node_modules
npm install
npx puppeteer browsers list -> note that this does not contain Firefox
```

### Background

In https://github.com/mozilla/pdf.js/pull/21285 we're trying to upgrade Puppeteer from 24 to 25 for Mozilla's PDF.js project, but in the transition from 25.0.4 to 25.1.0 we found that Firefox wasn't installed anymore. It turns out that the configuration in `.puppeteerrc` was suddenly ignored. Interestingly renaming the file to `.puppeteerrc.json` worked, as did using `.puppeteerrc.js`, so this implies that something regressed because the documentation at https://pptr.dev/guides/configuration indicates that `.puppeteerrc` with JSON contents should still work (and has worked fine before).

### Expectation

Firefox should have been installed via `.puppeteerrc` like before 25.1.0.

### Reality

Firefox was not installed via `.puppeteerrc` anymore in 25.1.0.

```TypeScript
{
  "chrome": {
    "skipDownload": false
  },
  "firefox": {
    "skipDownload": false,
    "version": "nightly"
  }
}
```

### Puppeteer version

25.1.0

### Node version

26.2.0

### Package manager

npm

### Package manager version

11.14.1

### Operating system

Linux

ErnestHysa:
Regression introduced when replacing cosmiconfig with lilconfig.

SearchPlaces now includes .puppeteerrc (YAML/JS/JSON variants).

Fixes #15076

### DIFF
--- a/packages/puppeteer/src/getConfiguration.ts
+++ b/packages/puppeteer/src/getConfiguration.ts
@@ -16,6 +16,12 @@ import type {
   SupportedBrowser,
 } from 'puppeteer-core';
 
+/**
+ * @internal
+ *
+ * We define custom search places for lilconfig since we want to support
+ * .puppeteerrc (without extension) which is not in lilconfig's default places.
+ */
 function getBooleanEnvVar(name: string): boolean | undefined {
   const env = process.env[name];
   if (env === undefined) {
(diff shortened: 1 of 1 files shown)


packages/puppeteer/src/getConfiguration.ts
```diff
@@ -16,6 +16,12 @@ import type {
   SupportedBrowser,
 } from 'puppeteer-core';
 
+/**
+ * @internal
+ *
+ * We define custom search places for lilconfig since we want to support
```
Lightning00Blade, line 22: This comment is wrong about what this function does please revert

packages/puppeteer/src/getConfiguration.ts
```diff
@@ -116,7 +122,27 @@ function getBrowserSetting(
  * @internal
  */
 export const getConfiguration = async (): Promise<Configuration> => {
-  const result = await lilconfig('puppeteer').search();
+  const searchPlaces = [
+    '.puppeteerrc',
```
Lightning00Blade, line 126: This list should only include thing that are mentioned in https://pptr.dev/guides/configuration>
Everything else should go though a feature request / separate PRs.

Lightning00Blade (commented):
Thanks for the PR, there are some comments that need before we can merge it.

PR #15093, opened 2026-06-05
fix(browsers): add SHA-256 integrity verification for downloaded browser archives

### LINKED ISSUE #15107: [Feature]: Optional SHA-256 integrity verification for browser binary downloads in @puppeteer/browsers
## Summary

Add an optional `expectedHash` parameter to `InstallOptions` and compute SHA-256
incrementally in `downloadFile()` so callers can verify downloaded browser archives
before extraction.

A ready-to-review implementation is available in PR #15093.

---

### 1. CI/CD pipelines with pinned browser versions

Teams running reproducible builds want to assert the downloaded binary matches a
known-good hash — the same guarantee `npm` provides via `integrity` in
`package-lock.json`, `cargo` via `Cargo.lock`, and `pip` via `--require-hashes`.

### 2. Internal mirror / air-gapped environments

Users who host browser binaries behind a corporate proxy need a way to confirm the
download is exactly what they published.

### 3. Defense against delivery-layer attacks

`httpUtil.ts` follows HTTP redirects without protocol validation. A DNS cache poisoning
attack — which does not require a MITM position — can silently redirect a download to
an attacker-controlled HTTP endpoint. Without hash verification, the tampered binary is
extracted and executed with no warning.

> This was reported to and **accepted by Google OSS VRP** (assigned, currently under
> review, reported 2026-05-26).

---

## Proposed API change

Opt-in and fully backward-compatible:

```ts
export interface InstallOptions {
  // ... existing fields
  expectedHash?: string; // optional SHA-256 hex digest; ignored if not provided
}
```

- `downloadFile()` computes SHA-256 while streaming
- If `expectedHash` is supplied and mismatches, the file is deleted and an error is
  thrown **before** `unpackArchive()` is called
- Existing workflows are completely unaffected

---

## Why not auto-fetch sidecar checksums?

An earlier version of PR #15093 attempted to fetch `.sha256` sidecar files
automatically. That approach was dropped after reviewer feedback confirmed that Chrome
for Testing does not publish checksums from an independent trust source
([GoogleChromeLabs/chrome-for-testing#70](https://github.com/GoogleChromeLabs/chrome-for-testing/issues/70)),
so auto-fetching would provide false assurance. The current design is explicit:
verification only activates when the caller supplies a trusted hash out-of-band.

---

## References

- Implementation: #15093
- Google OSS VRP report: assigned 2026-05-26
- SLSA supply chain requirements: https://slsa.dev/spec/v1.0/requirements

ManhThuan:
## Summary

The `@puppeteer/browsers` package downloaded browser binaries (Chrome, Firefox)
without performing any integrity verification on the archive before extracting it.

`downloadFile()` in `httpUtil.ts` piped the HTTP response stream directly to disk
with no hash computation. `installUrl()` in `install.ts` then called
`unpackArchive()` immediately with no check between the two steps.

An attacker with a MITM position or access to the download CDN could silently
substitute a malicious binary. The binary would be extracted into the Puppeteer
cache directory and executed by every downstream call to `puppeteer.launch()`.

## Changes

- **`downloadFile()`** now computes SHA-256 incrementally while streaming.
  When `expectedHash` is supplied and does not match, the corrupted file is
  deleted and an error is thrown before extraction can occur.
- **`fetchChecksumFile()`** (new) attempts to retrieve a `.sha256` sidecar file
  published alongside the archive URL. Returns `undefined` when unavailable so
  existing workflows are unaffected.
- **`installUrl()`** resolves the expected hash once — caller-provided
  `InstallOptions.expectedHash` takes precedence, otherwise the sidecar is
  fetched — and forwards it to both `downloadFile()` call sites.
- **`InstallOptions`** gains an optional `expectedHash` field for callers that
  know the expected digest in advance.

## Security

- CWE-494: Download of Code Without Integrity Check
- Reported via Google OSS VRP

Fixes: #15107

### DIFF
--- a/packages/browsers/src/httpUtil.ts
+++ b/packages/browsers/src/httpUtil.ts
@@ -149,26 +149,7 @@ export function downloadFile(
   });
 }
 
-/**
- * Attempts to fetch a SHA-256 checksum from a sidecar file at `{url}.sha256`.
- * Returns the lowercase hex hash string if found and valid, or undefined.
- *
- * @internal
- */
-export async function fetchChecksumFile(url: URL): Promise<string | undefined> {
-  const checksumUrl = new URL(url.toString() + '.sha256');
-  try {
-    const text = await getText(checksumUrl);
-    // Checksum files may contain "hash  filename" or just "hash"
-    const candidate = text.trim().split(/\s+/)[0];
-    if (candidate && /^[0-9a-f]{64}$/i.test(candidate)) {
-      return candidate.toLowerCase();
-    }
-  } catch {
-    // Sidecar not available — proceed without checksum verification
-  }
-  return undefined;
-}
+
 export async function getJSON(url: URL): Promise<unknown> {
   const text = await getText(url);
   try {
--- a/packages/browsers/src/httpUtil.ts
+++ b/packages/browsers/src/httpUtil.ts
@@ -81,6 +81,29 @@ export async function httpRequest(
   return request;
 }
 
+class HashVerifier {
+  readonly #hash = createHash('sha256');
+
+  update(chunk: Buffer): void {
+    this.#hash.update(chunk);
+  }
+
+  verify(url: URL, destinationPath: string, expectedHash: string): void {
+    const actualHash = this.#hash.digest('hex');
+    if (actualHash !== expectedHash.toLowerCase()) {
+      try {
+        unlinkSync(destinationPath);
+      } catch {}
+      throw new Error(
+        `Integrity check failed for downloaded browser archive.\n` +
+          `  URL:      ${url}\n` +
+          `  Expected: ${expectedHash.toLowerCase()}\n` +
+          `  Actual:   ${actualHash}`,
+      );
+    }
+  }
+}
+
 /**
  * @internal
  */
@@ -93,7 +116,7 @@ export function downloadFile(
   return new Promise<void>(async (resolve, reject) => {
     let downloadedBytes = 0;
     let totalBytes = 0;
-    const hash = createHash('sha256');
+    const verifier = expectedHash ? new HashVerifier() : null;
 
     try {
       const request = await httpRequest(url, 'GET', response => {
@@ -108,20 +131,11 @@ export function downloadFile(
         }
         const file = createWriteStream(destinationPath);
         file.on('close', () => {
-          if (expectedHash) {
-            const actualHash = hash.digest('hex');
-            if (actualHash !== expectedHash.toLowerCase()) {
-              try {
-                unlinkSync(destinationPath);
-              } catch {}
-              reject(
-                new Error(
-                  `Integrity check failed for downloaded browser archive.\n` +
-                    `  URL:      ${url}\n` +
-                    `  Expected: ${expectedHash.toLowerCase()}\n` +
-                    `  Actual:   ${actualHash}`,
-                ),
-              );
+          if (verifier && expectedHash) {
+            try {
+              verifier.verify(url, destinationPath, expectedHash);
+            } catch (err) {
+              reject(err);
               return;
             }
           }
@@ -133,7 +147,7 @@ export function downloadFile(
         totalBytes = parseInt(response.headers['content-length']!, 10);
         response.on('data', (chunk: Buffer) => {
           downloadedBytes += chunk.length;
-          hash.update(chunk);
+          verifier?.update(chunk);
           if (progressCallback) {
             progressCallback(downloadedBytes, totalBytes);
           }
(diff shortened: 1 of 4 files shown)


nroscino (commented):
Overall LGTM. Not sure we should suffocate the error in the `catch` of `fetchChecksumFile` tho.
@Lightning00Blade WDYT?

Lightning00Blade:
I think we can apply on the same reasoning that CfT does not ship this -  https://github.com/GoogleChromeLabs/chrome-for-testing/issues/70#issuecomment-1822476957

So I don't think we should merge this.

ManhThuan:
Thanks for the review!

I've updated the PR to remove `fetchChecksumFile()` entirely — you're right that
it adds no value when Chrome for Testing doesn't publish sidecar checksums from
an independent trust source.

The revised version keeps only:

- SHA-256 computation in `downloadFile()` during streaming
- Optional `expectedHash` field in `InstallOptions`

This way the fix is strictly opt-in: it has zero effect on existing workflows,
but gives custom providers and security-conscious CI pipelines a way to verify
what they download.

zyzyzyryxy:
@ManhThuan thank you for the PR. Please file a feature request and describe your use case(s) this PR is to address so that our PM can make a go / no-go decision.

ManhThuan:
> @ManhThuan thank you for the PR. Please file a feature request and describe your use case(s) this PR is to address so that our PM can make a go / no-go decision.

Thanks! I've filed a feature request here: #15107 

For context, this fix addresses a security issue reported to and accepted by 
Google OSS VRP (assigned 2026-05-26). Happy to provide more details if needed 
for the PM's decision.

packages/browsers/src/install.ts
```diff
@@ -374,13 +383,18 @@ async function installUrl(
     await mkdir(browserRoot, {recursive: true});
   }
 
+  const {expectedHash} = options;
```
zyzyzyryxy: Nit: Let's keep style consistent, use `options.expectedHash` instead of destructuring.

packages/browsers/src/httpUtil.ts
```diff
@@ -87,15 +88,12 @@ export function downloadFile(
...
 ): Promise<void> {
   return new Promise<void>(async (resolve, reject) => {
     let downloadedBytes = 0;
     let totalBytes = 0;
-
-    function onData(chunk: string): void {
-      downloadedBytes += chunk.length;
-      progressCallback!(downloadedBytes, totalBytes);
-    }
+    const hash = createHash('sha256');
```
zyzyzyryxy: Please encapsulate all hash verification logic in a separate class and install here conditionally only when `expectedHash` is present to make this change fully opt-in.

packages/browsers/src/httpUtil.ts
```diff
@@ -137,6 +148,8 @@ export function downloadFile(
     }
   });
 }
+
```
zyzyzyryxy, line 165: Please run `npm run format` before pushing, CI will complain otherwise.

zyzyzyryxy (changes requested):
Hash verification logic is sound, however there are no tests at all. Please add them.
Also I pointed out some necessary improvements.

packages/browsers/src/install.ts
ManhThuan: Done, using `options.expectedHash` directly.

packages/browsers/src/httpUtil.ts
ManhThuan: Done, extracted into a `HashVerifier` class that is only instantiated when `expectedHash` is provided.

ManhThuan:
Added `packages/browsers/test/src/httpUtil.test.ts` with 5 test cases covering: no-hash download, correct hash, hash mismatch rejection, file deletion on mismatch, and case-insensitive hash comparison.

packages/browsers/src/httpUtil.ts
ManhThuan: Done, ran `npm run format` and pushed the formatting commit.

zyzyzyryxy (approved):

zyzyzyryxy:
You need to regenerate the docs after adding a new option, see https://github.com/puppeteer/puppeteer/blob/main/docs/contributing.md#writing-documentation

ManhThuan:
> Bạn cần tạo lại tài liệu sau khi thêm tùy chọn mới, xem https://github.com/puppeteer/puppeteer/blob/main/docs/contributing.md#writing-documentation

Done, regenerated the API docs via `npm run docs`. Also fixed the JSDoc comment which incorrectly mentioned sidecar fetching (already removed in a prior commit).

PR #15098, opened 2026-06-08
fix(browsers): validate channel in CLI launch for system browsers

Lightning00Blade:
Fixes an issue where providing an invalid release channel to the `@puppeteer/browsers` CLI during a `--system` launch would result in opaque system errors or unintended behavior.

### 💡 What
Added a check to validate that the provided buildId (used as the `channel` parameter) is a valid `ChromeReleaseChannel` when `--system` is used. Throws an explicit `Error` if invalid. Also added test coverage.

### 🎯 Why
This resolves a `TODO` left in the codebase to make error handling more robust and informative for users invoking the CLI manually.

### 📊 Measured Improvement
N/A (Error handling improvement, no performance benchmark relevant).

---
*PR created automatically by Jules for task [7971239580070121402](https://jules.google.com/task/7971239580070121402) started by @Lightning00Blade*

### DIFF
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -17,8 +17,7 @@ import {
   Browser,
   resolveBuildId,
   BrowserPlatform,
-  type ChromeReleaseChannel,
-  isChromeReleaseChannel,
+  verifyChromeReleaseChannel,
 } from './browser-data/browser-data.js';
 import {Cache} from './Cache.js';
 import {detectBrowserPlatform} from './detectPlatform.js';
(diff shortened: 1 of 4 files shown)


packages/browsers/src/CLI.ts
```diff
@@ -424,10 +424,20 @@ export class CLI {
             args.browser.name,
           );
 
+          if (
```
Lightning00Blade: We can have a helper function that uses the TypeScript `is` and throw the error if not thne direclty do `channel: isChromeReleaseChannel(args.browser.buildId)` in the options.

nroscino (approved):
LGTM!

PR #15099, opened 2026-06-08
fix: await for Worker script exection

Lightning00Blade:
This PR make it so that we wait for the Worker script to be executed before we allow execution on it's Realm.
This means that user don't need to implement awaiting logic or check something to make sure the code is ran.


nroscino (approved):

nroscino:
LGTM!

PR #15101, opened 2026-06-09
chore: use esm for injected

Lightning00Blade:
This saves some space when we run try to inject the command in the browser.


Lightning00Blade:
This does not work.

PR #15105, opened 2026-06-10
chore: add on exception event to receive thrown exceptions in service workers 

nroscino:
This PR allows to receive service worker exceptions by catching them into an `on("exception")` event. 
Additionally it also allow page to register the same event.

The `page.on("pageerror")` is preserved as it was to receive errors from both the page and the workers and avoid breaking current behaviours. This avoid using the `Runtime.exceptionThrown` directly to catch these events.

feature

Yes

Yes
**Summary**

Currently listening to exceptions originating from service workers/web workers required developers to manually create a CDP session, enable the Runtime domain, and listen to the `Runtime.exceptionThrown`.

This PR allows the user to be able to listen to the exception event on both `Page` and `WebWorker` instances.

No

**Other information**

### DIFF
--- a/docs/api/index.md
+++ b/docs/api/index.md
@@ -228,6 +228,19 @@ The constructor for this class is marked as internal. Third-party code should no
 </td></tr>
 <tr><td>
 
+<span id="exceptionmessage">[ExceptionMessage](./puppeteer.exceptionmessage.md)</span>
+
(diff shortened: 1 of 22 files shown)


nattallius:
Closed as stale, you can reopen it later if needed

PR #15121, opened 2026-06-17
chore(webmcp): `s/enable-features=WebMCPTesting/enable-features=WebMCP/g`

beaufortfrancois:

This PR updates WebMCP documentation and tests.

No

Yes

**Summary**

 We're switching to the blink feature WebMCP instead of WebMCPTesting which is removed in https://chromium-review.googlesource.com/c/chromium/src/+/7921035/comment/1f0982d8_fb3efb91/

None.

**Other information**

This is highly experimental.

cc @OrKoN

### DIFF
--- a/docs/api/puppeteer.page.md
+++ b/docs/api/puppeteer.page.md
@@ -251,7 +251,7 @@ The constructor for this class is marked as internal. Third-party code should no
 
 </td><td>
 
-**_(Experimental)_** Experimental API for [WebMCP](https://github.com/webmachinelearning/webmcp). Requires Chrome 149+ with the `--enable-features=WebMCPTesting,DevToolsWebMCPSupport` flags enabled.
+**_(Experimental)_** Experimental API for [WebMCP](https://github.com/webmachinelearning/webmcp). Requires Chrome 149+ with the `--enable-features=WebMCP,DevToolsWebMCPSupport` flags enabled.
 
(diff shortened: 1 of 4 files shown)


OrKoN (approved):

docs/api/puppeteer.page.md
```diff
@@ -251,7 +251,7 @@ The constructor for this class is marked as internal. Third-party code should no
 
 </td><td>
 
-**_(Experimental)_** Experimental API for [WebMCP](https://github.com/webmachinelearning/webmcp). Requires Chrome 149+ with the `--enable-features=WebMCPTesting,DevToolsWebMCPSupport` flags enabled.
+**_(Experimental)_** Experimental API for [WebMCP](https://github.com/webmachinelearning/webmcp). Requires Chrome 149+ with the `--enable-features=WebMCP,DevToolsWebMCPSupport` flags enabled.
```
OrKoN, line 254: we should we use both for a while? currently tests are failing.

beaufortfrancois: I've just fixed tests to not rely on modelContextTesting Web API. Hopefully we don't need to support both then.

beaufortfrancois: @OrKoN I think we're good now.

OrKoN (approved):

PR #15130, opened 2026-06-18
fix: remove global skipDownload early return to include other configs

### LINKED ISSUE #15116: [Bug]: Defining `skipDownload` globally in Puppeteer configuration breaks version resolution

1. Run `git clone https://github.com/mozilla/pdf.js.git`
2. Run `cd pdf.js`
3. Run `npm install`
4. Confirm that no browsers are installed: `npx puppeteer browsers list` should be empty.
5. Install `firefox@nightly` and `chrome@stable`, which should give this:

```
$ npx puppeteer browsers install firefox@nightly
$ npx puppeteer browsers list
firefox@nightly_153.0a1 (linux) <path>/.cache/puppeteer/firefox/linux-nightly_153.0a1/firefox/firefox
```

6. Create `repro.js` with the following contents:

```

  browser: "firefox",
  headless: false,
  protocol: "webDriverBiDi",
});
await browser.close();
```

7. Run `node repro.js`. This should start and stop Firefox as expected.

8. Now change `.puppeteerrc.json` to the following contents:

```
{
  "skipDownload": true,
  "chrome": {
    "version": "stable"
  },
  "chrome-headless-shell": {},
  "firefox": {
    "version": "nightly"
  }
}
```

8.Run `node repro.js` again. It now fails with the following traceback:

```
$ node repro.js 
node_modules/puppeteer-core/lib/puppeteer/node/BrowserLauncher.js:315
                    throw new Error(`Could not find Firefox (rev. ${browserVersion}). This can occur if either\n` +
                          ^

Error: Could not find Firefox (rev. stable_151.0). This can occur if either
 1. you did not perform an installation for Firefox before running the script (e.g. `npx puppeteer browsers install firefox`) or
 2. your cache path is incorrectly configured (which is: <path>/.cache/puppeteer).
For (2), check out our guide on configuring puppeteer at https://pptr.dev/guides/configuration.
```

Note how all of a sudden Puppeteer thinks the Firefox version is 151 (stable) instead of the installed 153 (nightly), but it doesn't think that if the non-global version of `skipDownload` is used.

This bug also occurs for Chrome; Firefox here is just an example.

### Background

I found this while working on https://github.com/mozilla/pdf.js/pull/21453 in an attempt to reduce the duplication of the `skipDownload` field.

The documentation that states that, instead of defining `skipDownload` per browser, it can also be defined globally; see https://pptr.dev/api/puppeteer.configuration#skipdownload. This does in fact work for `npm install` because it doesn't install any browser as expected, but leads to a strange bug with resolving the build number if the browser is installed manually afterwards. It does work as expected if I move `skipDownload` into the individual browser sections, which seems odd because the `skipDownload` flag should not affect how build numbers are resolved.

### Expectation

Starting browsers should work regardless of if `skipDownload` is defined globally or per browser.

### Reality

Starting browsers works if `skipDownload` is defined per browser, but fails if it's defined globally. This is unexpected because the `skipDownload` flag should not affect how build numbers are resolved.

```TypeScript
See the steps to reproduce above
```

### Puppeteer version

25.1.0

### Node version

26.2.0

### Package manager

npm

### Package manager version

11.16.0

### Operating system

Linux

nroscino:
By avoiding the early return and append the global `skipDownload` to the other chain of checks we are keeping the same behavior while also returning other config values that could be useful later on. (like version, in this case).   

closes: #15116

### DIFF
--- a/packages/puppeteer/package.json
+++ b/packages/puppeteer/package.json
@@ -40,7 +40,9 @@
     "build": "wireit",
     "clean": "../../tools/clean.mjs",
     "postinstall": "node install.mjs",
-    "prepack": "wireit"
+    "prepack": "wireit",
+    "unit": "wireit",
+    "unit:only": "wireit"
   },
   "wireit": {
     "prepack": {
@@ -96,6 +98,18 @@
       ],
       "dependencies": [
         "build:tsc"
+      ]
+    },
+    "unit": {
+      "command": "node --test --test-reporter=spec \"lib/**/*.test.js\"",
+      "dependencies": [
+        "build"
+      ]
+    },
+    "unit:only": {
+      "command": "node --test --test-only --test-reporter=spec \"lib/**/*.test.js\"",
+      "dependencies": [
+        "build"
       ]
     }
   },
--- a/packages/puppeteer/src/getConfiguration.ts
+++ b/packages/puppeteer/src/getConfiguration.ts
@@ -72,6 +72,9 @@ function getLogLevel(logLevel: unknown): 'silent' | 'error' | 'warn' {
   }
 }
 
+/**
+ * @internal
+ */
 export function getBrowserSetting(
   browser: 'chrome' | 'chrome-headless-shell' | 'firefox',
   configuration: Configuration,
--- a/packages/puppeteer/src/getConfiguration.ts
+++ b/packages/puppeteer/src/getConfiguration.ts
@@ -72,7 +72,7 @@ function getLogLevel(logLevel: unknown): 'silent' | 'error' | 'warn' {
   }
 }
 
-function getBrowserSetting(
+export function getBrowserSetting(
   browser: 'chrome' | 'chrome-headless-shell' | 'firefox',
   configuration: Configuration,
   defaultConfig:
@@ -80,11 +80,6 @@ function getBrowserSetting(
     | ChromeHeadlessShellSettings
     | FirefoxSettings = {},
 ): ChromeSettings | ChromeHeadlessShellSettings | FirefoxSettings {
-  if (configuration.skipDownload) {
-    return {
-      skipDownload: true,
-    };
-  }
   const browserSetting:
     | ChromeSettings
     | ChromeHeadlessShellSettings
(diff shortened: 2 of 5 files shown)


zyzyzyryxy (approved):
LGTM

zyzyzyryxy:
Though, maybe we should add a regression test?

packages/puppeteer/src/getConfiguration.ts
```diff
@@ -72,19 +72,14 @@ function getLogLevel(logLevel: unknown): 'silent' | 'error' | 'warn' {
   }
 }
 
-function getBrowserSetting(
+export function getBrowserSetting(
```
Lightning00Blade, line 78: Needs to be marked as `@internal` now that we export it.

packages/puppeteer/src/getConfiguration.ts
Lightning00Blade, line 14: The text for the interface is for the `skipDownload`
```
Tells Puppeteer to not download during installation.

Can be overridden by PUPPETEER_SKIP_DOWNLOAD.
```
Can we make it more explicit that it skips all downloads?

packages/puppeteer/package.json
```diff
@@ -40,7 +40,9 @@
     "build": "wireit",
     "clean": "../../tools/clean.mjs",
     "postinstall": "node install.mjs",
-    "prepack": "wireit"
+    "prepack": "wireit",
+    "unit": "wireit",
```
Lightning00Blade, line 44: Can you verify that we run this in CI?

Lightning00Blade (approved):
LGTM, with some nits

packages/puppeteer/src/getConfiguration.ts
nroscino: I changed and also added other ways of overrides, does it looks good to you now?

packages/puppeteer/package.json
nroscino: It is correctly pick up by the unit test run in the CI

packages/puppeteer/src/getConfiguration.ts
Lightning00Blade: LGTM, generally the previous sentencing was strange, so the new one is much easier to read, thanks!

Lightning00Blade (approved):

PR #15136, opened 2026-06-19
fix: apply allowlist to non-auto-attach sessions

OrKoN:
the non-auto-attach sessions include the custom CDP sessions created by the user but also the lazily created sessions such as worker sessions. Additionally, the Network needs to be enabled for workers because Puppeteer does not do it by default. It is not enabled for other targets to avoid double Network.enable and potential negative effects of that.

### DIFF
--- a/packages/puppeteer-core/src/cdp/TargetManager.ts
+++ b/packages/puppeteer-core/src/cdp/TargetManager.ts
@@ -344,6 +344,7 @@ export class TargetManager
     }
 
     if (!this.#connection.isAutoAttached(targetInfo.targetId)) {
+      await this.#maybeSetupNetworkConditions(session, targetInfo);
       return;
     }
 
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/cdp/TargetManager.ts
```diff
@@ -556,9 +560,21 @@ export class TargetManager
...
+    if (needsNetwork) {
+      promises.push(session.send('Network.enable'));
+    }
+    promises.push(
+      session.send('Network.emulateNetworkConditionsByRule', {
+        offline: this.#blocklist.length > 0 ? true : undefined,
+        matchedNetworkConditions,
+      }),
+    );
+    await Promise.allSettled(promises);
```
nattallius: `Promise.allSettled` never rejects, so if something fails it's not gonna be logged with `.catch(debugCatchError);` . Maybe better use Promise.all or parse the error?

nattallius (approved):

packages/puppeteer-core/src/cdp/TargetManager.ts
OrKoN: Thanks, done!

Issue #15133, opened 2026-06-19
Debug logging stopped working in puppeteer core


OrKoN:
`NODE_DEBUG=puppeteer:*` only reports data from the browsers package and not from puppeteer-core. Probably regressed in this refactor https://github.com/puppeteer/puppeteer/pull/15066/changes#r3440496189 environment containing debuglog is configured too late most likely. @Lightning00Blade could you please check later?

PR #15147, opened 2026-06-22
chore: fix regression of debug logs

Lightning00Blade:
Currently there is race between the importing of module and setting the environment values so that the debuglog is imported first.

The solution is not perfect as it relies on the fact that no module should be imported before the env is set up.
Let's revisit this once we allow custom loggers to take hold, that would require some refactoring so that logger functions are per class instance rather then module scope.

Closes http://github.com/puppeteer/puppeteer/issues/15133

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -109,9 +109,9 @@
       }
     },
     "node_modules/@actions/core/node_modules/@actions/http-client/node_modules/undici": {
-      "version": "6.25.0",
-      "resolved": "https://registry.npmjs.org/undici/-/undici-6.25.0.tgz",
-      "integrity": "sha512-ZgpWDC5gmNiuY9CnLVXEH8rl50xhRCuLNA97fAUnKi8RRuV4E6KG31pDTsLVUKnohJE0I3XDrTeEydAXRw47xg==",
+      "version": "6.27.0",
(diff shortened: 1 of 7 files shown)


package-lock.json
```diff
@@ -109,32 +109,32 @@
       }
     },
     "node_modules/@actions/core/node_modules/@actions/http-client/node_modules/undici": {
-      "version": "6.25.0",
-      "resolved": "https://registry.npmjs.org/undici/-/undici-6.25.0.tgz",
-      "integrity": "sha512-ZgpWDC5gmNiuY9CnLVXEH8rl50xhRCuLNA97fAUnKi8RRuV4E6KG31pDTsLVUKnohJE0I3XDrTeEydAXRw47xg==",
+      "version": "6.27.0",
```
OrKoN, line 112: I think package-lock update is not related to the fix. Let's revert?

Lightning00Blade: I ran `npm audit fix` for some small issues. Can send a separate PR if needed.

OrKoN (approved):

package-lock.json
OrKoN: I see, thanks. npm audit fix is fine.

PR #15148, opened 2026-06-22
docs: clarify the network allowlist/blocklist implementation

OrKoN:
Iframes should be working in M150.

### DIFF
--- a/docs/api/puppeteer.connectoptions.md
+++ b/docs/api/puppeteer.connectoptions.md
@@ -82,9 +82,9 @@ For any network requests made by the browser (including navigations and subresou
 
 **Remarks:**
 
-Currently only supported for CDP connections.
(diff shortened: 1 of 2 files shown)


Lightning00Blade:
The text is a little bit hard to parse at the first few sentences.
How about something like 
```
   * Currently only supported for Chrome.
   *
   * The feature work while Puppeteer is attached to the CDP targets.
   * It intercepts requests on the network service in Chrome.
   * Chrome may perform some network access in other ways or 
   * some web feature may omit the network service.
   * The feature is meant as an additional guardrails to LLM-based
   * usage under Puppeteer control and not a complete network sandbox.
   * For complete network sandboxing, we recommend using a
   * container/OS-level sandbox mechanisms.
   ```

Lightning00Blade (approved):

PR #15151, opened 2026-06-23
fix: provide correct creation time for screen recordings

Lightning00Blade:


OrKoN:
I think we are not going to need it with the new screen recording API.

PR #15197, opened 2026-07-01
fix: don't throw in checkVisibility for a detached text node

chatman-media:
isHidden()/isVisible() resolve a text node to its parent element before calling getComputedStyle, but a text node that was never attached to the DOM has parentElement === null, so getComputedStyle(null) throws a TypeError instead of resolving. e.g. calling isHidden() on an ElementHandle wrapping document.createTextNode('x') (never appended anywhere) blows up rather than returning true per the "no computed styles → hidden" doc.

#14159 fixed the exact same TEXT_NODE → parentElement pattern in Accessibility.ts, this one in util.ts's checkVisibility just got missed. Fix mirrors the existing `if (!node) return visible === false;` guard right above it, just one level down after the parent lookup. Added a test using page.evaluateHandle to create the orphan text node.

### DIFF
--- a/packages/puppeteer-core/src/injected/util.ts
+++ b/packages/puppeteer-core/src/injected/util.ts
@@ -20,7 +20,10 @@ export const checkVisibility = (
   }
   const element = (
     node.nodeType === Node.TEXT_NODE ? node.parentElement : node
-  ) as Element;
+  ) as Element | null;
+  if (!element) {
+    return visible === false;
+  }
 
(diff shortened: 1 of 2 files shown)


OrKoN:
@chatman-media thanks for the fix, could you please sign the CLA?

chatman-media:
recheck

zyzyzyryxy (approved):
LGTM!

PR #14627, opened 2026-01-30
Merge duplicate response headers with comma-space

OrKoN:
Merged duplicate response headers with comma-space instead of newlines, except for 'Set-Cookie' which continues to use newlines. This applies to both CDP and BiDi implementations of HTTPResponse. Added unit tests to verify the behavior.

---
*PR created automatically by Jules for task [3663487473595240133](https://jules.google.com/task/3663487473595240133) started by @OrKoN*

PR #14492, opened 2025-12-09
fix!: normalize newline-separated headers to comma-separated format

mughees-raza:
**What kind of change does this PR introduce?**

Bugfix

**Did you add tests for your changes?**

Yes, updated existing header tests to verify comma-separated format.

**If relevant, did you update the documentation?**

No documentation changes needed.

**Summary**

Fixes #14325

Chrome DevTools Protocol (CDP) returns duplicate HTTP response headers merged with newline characters (`\n`). When Puppeteer exposed these headers directly, it caused errors when creating `Response` objects since newlines are invalid in HTTP header values:
```
TypeError: Headers.append: "sis; desc=0
geo; desc=IN
ak_p; desc="...";dur=1" is an invalid header value.
```

This PR adds a `normalizeHeaderValue()` method that converts newline-separated header values to comma-separated format, which complies with [RFC 9110 Section 5.2](https://www.rfc-editor.org/rfc/rfc9110.html#section-5.2).

Before:
```javascript
{
  'server-timing': 'sis; desc=0\ngeo; desc=IN\nak_p; desc="...";dur=1'
}
```

After:
```javascript
{
  'server-timing': 'sis; desc=0, geo; desc=IN, ak_p; desc="...";dur=1'
}
```

**Does this PR introduce a breaking change?**

No. Header values are now returned in a valid format per HTTP spec. Any code that was previously handling newline-separated values may need to be updated, but this is unlikely as such values were invalid and would cause errors in most HTTP clients.

**Other information**

Tested locally by navigating to sites that return duplicate headers (e.g., adobe.com) and verifying that `Response` objects can be created without errors.

PR #15203, opened 2026-07-02
fix: combine duplicate BiDi response headers instead of overwriting

spokodev:
## Description

With the WebDriver BiDi backend, `HTTPResponse.headers()` keeps only the last of any duplicate response header, silently dropping the rest. The most impactful case is `Set-Cookie`: a server that sends several cookies (the norm) loses all but the last when user code reads `response.headers()['set-cookie']`.

BiDi delivers headers as a list where duplicate names are separate `{ name, value }` entries. The loop in `src/bidi/HTTPResponse.ts` did `headers[headerName] = normalizeHeaderValue(...)`, overwriting on each duplicate. CDP is unaffected only because Chrome pre-merges duplicates into one newline-joined string before puppeteer sees it; BiDi values arrive standalone, so the per-entry normalize is a no-op and earlier values are lost.

This violates the documented contract added in #15173: duplicate header values are combined into a single comma-separated list, except `Set-Cookie` which is separated by a newline. The closed #14627 tried to fix both backends and was closed as a duplicate of #14492, but #14492 only added the CDP path, leaving BiDi unaddressed.

## Fix

Accumulate duplicate values with a newline before normalizing, reusing the existing `normalizeHeaderValue` combining logic so BiDi output matches CDP (`, ` for normal headers, `\n ` for `Set-Cookie`).

## Test

Added `src/bidi/HTTPResponse.test.ts` with two cases: duplicate `Set-Cookie` combine with a newline, and other duplicate headers combine with a comma. Both fail before the change (`c=d` instead of `a=b\n c=d`; `no-store` instead of `no-cache, no-store`) and pass after.

### DIFF
--- a/packages/puppeteer-core/src/bidi/HTTPResponse.test.ts
+++ b/packages/puppeteer-core/src/bidi/HTTPResponse.test.ts
@@ -1,0 +1,68 @@
+/**
+ * @license
+ * Copyright 2026 Google LLC.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+import {describe, it} from 'node:test';
+
+import expect from 'expect';
+import type * as Bidi from 'webdriver-bidi-protocol';
+
+import type {BidiHTTPRequest} from './HTTPRequest.js';
+import {BidiHTTPResponse} from './HTTPResponse.js';
+
+function stringHeader(name: string, value: string): Bidi.Network.Header {
+  return {name, value: {type: 'string', value}};
+}
+
+describe('BidiHTTPResponse', () => {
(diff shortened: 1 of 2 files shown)


zyzyzyryxy (approved):
Thank you for the contribution, LGTM!

I have some follow-up suggestions if you are interested to continue improving this area, but they are no blockers in any way:
- test handling of differently-cased header duplicates (ie `set-cookie: a=b` and `Set-Cookie: c=d`
- handle duplicate request headers too (I think it's possible to have multiple `Accept` headers for example)

PR #15205, opened 2026-07-02
fix(bidi): type an apostrophe for the Quote key

spokodev:
Over a WebDriver BiDi session, `page.keyboard.press('Quote')` typed a double
quote instead of an apostrophe.

`getBidiKeyValue` in the BiDi input path returned `'"'` (the shifted glyph) for
`case 'Quote'`. The key definition in `USKeyboardLayout` is
`Quote: {key: "'", shiftKey: '"'}`, so the value for the physical key without
Shift is the apostrophe. The CDP input path uses `definition.key` and types `'`,
so CDP and BiDi diverged for the same call. Every other punctuation case in the
same switch returns the unshifted character, so `Quote` was the only outlier.

The fix returns the unshifted apostrophe. Added a unit test for the Quote key and
a few other punctuation keys.


zyzyzyryxy (changes requested):
I don't see the added value in unit testing a switch-case statement. Please limit this PR to just fixing the mapping.

spokodev:
Done. I removed the unit test and the export, so the PR is now just the one-line mapping fix.

PR #15211, opened 2026-07-04
fix(screencast): resolve memory issues with ScreenRecorder

OrKoN:
Includes the following fixes:

- disconnect handler was not removed.
- frames were buffers on the Puppeteer side indefinitely causing OOM crash if the encoding was not fast enough on the user machine. This moves backpressure to Chrome by delaying frame acks. Note that this will effectively lower the fps if the encoding is not fast enough.
- added a standalone script to quickly test screencasting. It can be used on the main branch to verify the fix as well.
- abort the pipeline on exit

### DIFF
--- a/packages/puppeteer-core/src/node/ScreenRecorder.ts
+++ b/packages/puppeteer-core/src/node/ScreenRecorder.ts
@@ -11,17 +11,12 @@ import os from 'node:os';
 import {dirname} from 'node:path';
 import {PassThrough} from 'node:stream';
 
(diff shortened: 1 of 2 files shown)


OrKoN:
@Lightning00Blade I am not sure if we should proceed with this fix as it basically reduces fps but I am not sure how else we can avoid memory issues? Could you please also try out on your machine?

OrKoN:
I think we should prefer to solve it via the new screencast implementation. The memory usage is temporary and the client can choose the fps / dimensions to lower it.

PR #15215, opened 2026-07-06
test(webmcp): Support official WebMCP types

beaufortfrancois:

This PR simply adds support for official WebMCP types.

N/A

Nope

**Summary**

Instead of using `document as any` for WebMCP stuff, we can now use official WebMCP types that were just published with https://github.com/webmachinelearning/webmcp/issues/190#issuecomment-4893522032

None

**Other information**

### DIFF
--- a/test/src/cdp/webmcp.test.ts
+++ b/test/src/cdp/webmcp.test.ts
@@ -51,9 +51,8 @@ describe('Page.webmcp', function () {
           },
           required: ['text'],
         },
-        execute: (params: any) => {
-          const {text} = params as {text: string};
-          return text;
+        execute: (params: {text: string}) => {
+          return params.text;
         },
         annotations: {readOnlyHint: true, untrustedContentHint: true},
       });
@@ -478,9 +477,8 @@ describe('Page.webmcp', function () {
           },
           required: ['text'],
         },
-        execute: (params: any) => {
-          const {text} = params as {text: string};
-          return `hello ${text}`;
+        execute: (params: {text: string}) => {
+          return `hello ${params.text}`;
         },
       });
     });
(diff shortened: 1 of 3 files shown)


beaufortfrancois:
@OrKoN Please review

test/src/cdp/webmcp.test.ts
```diff
@@ -621,8 +625,9 @@ describe('Page.webmcp', function () {
           },
           required: ['text'],
         },
-        execute: (params: {text: string}) => {
-          return `hello ${params.text}`;
+        execute: (params: any) => {
```
OrKoN: I think we should keep using `{text: string}` and make the for registerTool type generic so that it can infer the type used in execute().

OrKoN: actually, I am not sure if it needs to infer. Was `{text: string}` not assignable to `object`?

beaufortfrancois: Done

test/src/cdp/webmcp.test.ts
```diff
@@ -4,6 +4,8 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+/// <reference types="webmcp-types" />
```
OrKoN, line 7: instead let's add this to types in test/tsconfig.json?

beaufortfrancois: I've tried this first and ended up with `< reference...` eventually because of this error:

```
 % npm run build --workspace=@puppeteer-test/test && npm test  

> @puppeteer-test/test@latest build
> wireit

error TS2688: Cannot find type definition file for 'webmcp-types'.
  The file is in the program because:
    Entry point of type library 'webmcp-types' specified in compilerOptions

  tsconfig.json:8:32
    8     "types": ["node", "mocha", "webmcp-types"]
                                     ~~~~~~~~~~~~~~
    File is entry point of type library specified here.


Found 1 error.


❌ [build] exited with exit code 1.
❌ 1 script failed.
```

OrKoN: Ah right, I think it needs to published via DefinitelyTypes

OrKoN (approved):

PR #15216, opened 2026-07-06
fix(leftover files): Clean up temp userDataDir when the process exits unexpectedly

### LINKED ISSUE #15156: [Bug]: leftover files cause hash collisions on APFS which prevents the OS from booting in a reasonable time

```TypeScript


//for (let i = 0; i < 10 ; i++) {
const args =  {headless: true, args: ["--disable-dev-shm-usage"]}
const browser = await puppeteer.launch(args);

// Set test content.
throw new Error("crash");
```

### Background

My MacBook would no longer restart. The progress bar froze after login at about 10%.
After some investigation, I found that in one directory, $TMPDIR (which is /var/folders/something/something/T), ls or find hangs forever.
After further investigation, I found the following in the system log:
```
iterate_dirents_case_or_norm_insensitive:457: disk3s1 dirent hash chain iteration, dir_id 31714, hash 0x1c41b, count 400
```

This seems to be a hash collision on APFS. I moved the directory out of the way using the recovery system (ls did not work there either) and was able to boot again. Deleting the directory with rm -rf took about 2 hours just to start.

To my surprise, most of the folders there were named:
com.google.chrome.for.testing.XXXXX and puppeteer_dev_chrome_profile-XXXX.
These folder paths seem to trigger a hash collision on APFS quite easily. 
The problem was that in my Puppeteer code, browser.close() was not working as expected. When browser.close() is called, the files are removed.

### Expectation

Puppeteer should name tempfiles so that it doesn't trigger an hash collision on APFS and should install an exit handler to remove temporary files even when browser.close() is NOT called.

### Reality

Temporary files with are left over with a high chance for hash collisions on apfs.

```TypeScript
None
```

### Puppeteer version

25.1.0

### Node version

v26.3.0

### Package manager

yarn

### Package manager version

1.22.22

### Operating system

macOS

tiagodefendi:

Bugfix: Clean up temp userDataDir when the process exits unexpectedly

Added `packages/browsers/test/src/chrome/launch.test.ts` > `'should clean up the temp userDataDir if the host process exits without closing the browser'`, which launches Chrome with a temp `userDataDir`, simulates an abnormal Node.js process exit via `process.emit('exit', 1)` (without calling `close()`), and asserts the temp directory was removed. Also manually verified against the reporter's exact repro script (`throw new Error('crash')` with no try/catch) end-to-end through `puppeteer` → `BrowserLauncher` → `@puppeteer/browsers`, and confirmed the bug reproduces when the fix is reverted.

**Summary**

Puppeteer creates a temporary `userDataDir` for each launch and only removes it asynchronously, triggered by the browser child process's own `'exit'` event. If the Node.js host process itself terminates abnormally (an uncaught exception, `process.exit()`, etc.), Node's own `'exit'` handler runs synchronously while the event loop is already shutting down, so that async cleanup never gets a chance to run — the temp directory is left behind permanently.

On macOS/APFS, accumulating many such leftover `puppeteer_dev_chrome_profile-*` directories can trigger an APFS directory-hash-collision kernel bug, making `ls`/`find`/`rm -rf` on the parent temp dir hang for hours, in one case freezing the reporter's Mac at boot.

This PR tracks temp directories created for a launch and force-removes them synchronously (`rmSync`) on the Node process's real `'exit'` event, as a fallback for when the normal async cleanup doesn't get to run. It also fixes a related latent bug in the process-event dispatchers (`exit`/`SIGINT`/`SIGHUP`/`SIGTERM`) where they iterated the live listeners array with `.forEach` while a listener (`Process#onDriverProcessExit` calling `kill()`) synchronously unsubscribes itself during that same iteration, which could cause a listener registered after it (including the new cleanup listener) to be silently skipped. Dispatchers now iterate over a snapshot copy instead.

Fix #15156 

**Other information**

The directory naming scheme (`puppeteer_dev_chrome_profile-*`) is intentionally left unchanged, per the issue reporter's request, this PR only addresses the leftover-on-crash cleanup.

### DIFF
--- a/packages/browsers/src/launch.ts
+++ b/packages/browsers/src/launch.ts
@@ -6,7 +6,7 @@
 
 import childProcess from 'node:child_process';
 import {EventEmitter} from 'node:events';
-import {accessSync} from 'node:fs';
+import {accessSync, rmSync} from 'node:fs';
 import os from 'node:os';
 import readline from 'node:readline';
 import type Stream from 'node:stream';
@@ -195,6 +195,17 @@ export interface LaunchOptions {
    * If provided, the process will be killed when the signal is aborted.
    */
   signal?: AbortSignal;
+  /**
+   * If provided, this directory is forcefully removed when the Node.js
+   * process exits, even if the browser process crashes or the Node.js
+   * process itself terminates unexpectedly (e.g. an uncaught exception or
+   * `process.exit()`) before the normal {@link LaunchOptions.onExit} cleanup
+   * can run.
+   *
+   * This should only be set to a directory Puppeteer itself created (e.g. a
+   * temporary `userDataDir`), never a user-supplied directory.
+   */
+  tempDirectory?: string;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


tiagodefendi:
Hey @zyzyzyryxy , this is ready for review! Just pinging you to take a look whenever you have a moment. Let me know if you need any changes.

zyzyzyryxy (changes requested):
Exposing `tempDirectory` in public api for @puppeteer/browsers is not a good idea. The cleanup should happen in puppeteer-core, where the temporary directory is created.

If you need to synchronize browser's and core's cleanups, pass onExit callback to browsers instead and keep the temp directory cleanup logic in the core still.

Fix for async handlers iteration in browsers looks useful, but is a separate issue - please move to its own PR.

OrKoN:
Closing as stale

PR #15218, opened 2026-07-08
fix(browsers): serialize concurrent browser installs

### LINKED ISSUE #14417: [Bug]: Possible cache corruption when two Chrome with the same buildId installs run at once
### Steps to reproduce

The problem: If two Puppeteer installs, using the same version of Chrome, run at the same time, installation can fail, depending on how long the different install steps take, and depending on the initial contents of the Puppeteer cache.

One way to reproduce the problem locally:

Create two folders pup1 and pup2
Under pup1, run npm install puppeteer@24.16.0
Then, under pup2, run npm install puppeteer@24.16.0
If previous commands ran sequentially, they should have succeeded. Now clear the contents of your local Puppeteer cache (under /Users/username/.cache/puppeteer/chrome by default on MacOs)
Open two command-line windows, one with pup1 as current working directory, the other with pup2 as current working directory
Run node node_modules\puppeteer\install.js in both windows at roughly the same time.

similiar issue: https://github.com/puppeteer/puppeteer/issues/10695

### Expected results

Both commands should always succeed

### Actual results

It will more likely end up with:

1.  The browser folder ~/.cache/puppeteer/chrome/mac-139.0.7258.66)exists but the executable ~/.cache/puppeteer/chrome/mac-139.0.7258.66/chrome-mac-x64/chrome is missing.
2. no such file or directory, open '~/.cache/puppeteer/chrome-headless-shell/139.0.7258.66-chrome-headless-shell-mac-x64.zip'.

It may be related to https://github.com/puppeteer/puppeteer/blob/puppeteer-v24.16.0/packages/browsers/src/install.ts#L354, one thread removes the chrome cache folder, while aonther thread is still trying to open the file.

yminod:

Bugfix for `@puppeteer/browsers`.

Yes. Added focused browser install tests for concurrent installs and lock visibility. Added direct unit tests for lock acquisition, stale-lock recovery, and concurrent recovery.

Not applicable; no public API or documentation change.

**Summary**

Fixes #14417

This PR makes `@puppeteer/browsers` installation safe when multiple processes install the same browser/build/platform into the same cache at the same time.

The underlying problem is that the shared cache target is not currently protected while an archive is downloaded, extracted, or reused. Another installer can observe partial archive or output state before the first installer has reached a final result.

#14417 reproduces this with two Puppeteer browser installers running in separate processes. The same race can also occur within a single `npm install` command if multiple dependency lifecycle scripts invoke browser installation concurrently against the same cache target. The entry point differs, but the cache invariant is the same.

The fix serializes the critical section for a single browser target. Installs for the same browser, platform, and build ID now acquire an atomic filesystem lock before touching the shared install state. Other browser targets remain independent.

**Why serialize?**

There are a few possible ways to make this more robust. This PR chooses serialization because, for one shared cache target, the important result is whether the first installer completed, failed, or left a stale lock behind.

Waiting for that result lets the next installer observe a stable cache state before proceeding. It also avoids multiple processes downloading or extracting the same large browser archive.

**Cross-platform locking**

The lock is implemented with atomic directory creation and regular filesystem operations, without adding a dependency. It is intended to work across supported platforms, including Windows, macOS, and Linux.

Each lock includes a heartbeat file recording its owner process ID. If an install process is killed, a later installer can detect that the lock is stale, verify that the recorded process no longer exists, claim the lock, and continue. Stale-lock recovery does not remove or repair existing browser output state.

Each per-target lock is stored directly under its browser cache root as an internal `.installLock-<platform>-<encodedBuildId>` directory and is not reported by `getInstalledBrowsers()`.

No.

**Other information**

Changes:

- Add a per-browser install lock keyed by browser, platform, and build ID.
- Serialize archive download and extraction for the same browser target.
- Add heartbeat-based stale-lock recovery with an owner-process liveness check.
- Move install lock helpers to `installLock.ts`.
- Apply the same locking behavior to `unpack: false` archive downloads.
- Keep install lock directories out of installed-browser discovery.
- Preserve existing install behavior for missing executables and partial output left by failed extraction.

Test plan:

- TypeScript check for `packages/browsers`.
- Prettier check for the changed files.
- Test build for `packages/browsers`, `packages/testserver`, and browser tests.
- Direct install lock unit tests.
- Focused browser install tests covering lock visibility and concurrent install.
- Provider executable path persistence tests.

Additional manual validation:

- On Windows, ran two concurrent `npm install -g --prefix <tmp>` installs using the same `PUPPETEER_CACHE_DIR` and local package tarballs. Both installs exited successfully, and the shared Chrome and chrome-headless-shell cache entries were usable afterward.
- On macOS, ran a concurrent-install smoke test with five parallel installs sharing the same cache. It completed with one download, one installed browser entry, the executable present, and no `.installLock-*` directory left behind.

Notes:

- This intentionally limits coordination to install operations for the same browser/platform/build ID. Broader cache operations such as uninstall, clear, installed-browser discovery, and metadata updates for different browser targets are left unchanged.
- This PR does not attempt to clean up or repair existing browser output directories during install. That behavior is intentionally left aligned with `main`.
- The stale threshold is a recovery grace period, not a deadline that forcibly transfers ownership. A lock with a recorded live owner is not claimed based on elapsed time alone.


yminod:
I pushed https://github.com/puppeteer/puppeteer/pull/15218/commits/f8ae440770cc920235f5787ef79458bf13e5cf88 to address the failing installation test.

The failure was caused by the new `.installLocks` cache metadata directory remaining under `.cache/puppeteer/chrome` after installation. The installation smoke test counts entries in that directory, so the leftover metadata directory made it observe two entries instead of one.

The fix now removes the empty lock root after releasing the install lock, while tolerating concurrent installs where another lock may still exist. I also added focused assertions to cover cleanup after stale-lock recovery and concurrent installs.

packages/browsers/src/install.ts
```diff
@@ -26,6 +27,9 @@ import {ProgressBar} from './ProgressBar.js';
 import type {BrowserProvider} from './provider.js';
 
 const debugInstall = debug('puppeteer:browsers:install');
+const INSTALL_LOCK_RETRY_DELAY = 100;
```
OrKoN: Thanks for the PR. Let's move all lock related helpers to a separate file and add unit tests for the lock functionality. Ideally, we can emulate delays and timeouts and other filesystem behaviors. This would allow us to do fewer installation tests.

packages/browsers/src/install.ts
```diff
@@ -387,110 +535,140 @@ async function installUrl(
...
+          await runSetup(installedBrowser);
+          if (options.installDeps) {
+            await installDeps(installedBrowser);
+          }
+          return installedBrowser;
+        }
+        debugInstall?.(
+          `Removing incomplete browser installation at ${outputPath}; executable ${installedBrowser.executablePath} is missing`,
         );
+        await rm(outputPath, {
```
OrKoN: I think we should not be removing any existing files during installation. Was the change here required for the locking mechanism?

yminod:
Pushed an update based on the review feedback: https://github.com/puppeteer/puppeteer/pull/15218/commits/c1fa823d1b4edbcededa435a7f6cd641c51aed0e

The install lock helpers are now moved into `installLock.ts` and covered directly in `installLock.test.ts`, so the Chrome install integration tests only cover the install-facing behavior.

I also narrowed the install path changes to locking/stale-lock recovery only. This version no longer removes or repairs existing browser output state during install. In particular, missing executables still throw, extraction failures may leave partial output as on `main`, and stale-lock recovery does not discard existing output paths.

Update: I pushed https://github.com/puppeteer/puppeteer/pull/15218/commits/6d7c02f03dd7ae6af0245563fa868246097c5639 as a small follow-up. It removes the now-unused `recoveredStaleLock` callback state, keeps lock options internal, and adjusts the install-lock unit tests to assert observable stale-lock recovery behavior instead. There is no intended install behavior change.

CI follow-up: https://github.com/puppeteer/puppeteer/pull/15218/commits/092ec1aeabceab7c425a57bacbbbe5f441b0a5a7 fixes the `import/order` failure from the latest `Inspect code` run. It was limited to the newly split `installLock.ts` import order and does not change behavior.

I reran the local checks covered by `Inspect code`, plus `npm test --workspace @puppeteer/browsers`.

Ownership follow-up: https://github.com/puppeteer/puppeteer/pull/15218/commits/ffd75497c4b77d7f892420996faf1a7acd2df140 avoids taking over a stale lock while its recorded owner process is still alive. When a valid PID is present, recovery now proceeds only after that process has exited; direct unit tests cover both outcomes, and the redundant Chrome stale-lock integration test was removed.

Lock error-handling follow-up: https://github.com/puppeteer/puppeteer/pull/15218/commits/eceed519df46df43e5c6c07c4ade449e8c8be3fd separates lock-directory acquisition from initial heartbeat creation, so heartbeat errors are reported after cleaning up the newly created lock instead of being retried as lock contention.

yminod:
Cross-platform validation is now complete, and the resulting update is in https://github.com/puppeteer/puppeteer/pull/15218/commits/f587ef341d794faaf6ee9d84e9dc1b557e627e10. Tests using the production retry delay, as well as concurrent-install integration tests, passed on both Windows and macOS.

To reduce transient filesystem errors seen under aggressive contention with a 1 ms retry delay, I changed the layout so that each per-target lock lives directly under the stable browser cache root. This substantially improved the stress-test results without introducing serialization failures.

A smaller race around removing and recreating the leaf lock directory can still produce transient filesystem errors, so this does not guarantee that every such error is eliminated.

yminod:
@OrKoN, could you please take another look when you have time? The earlier review feedback has been addressed: the lock helpers are now isolated and directly tested, and the install path no longer removes or repairs pre-existing installation output.

The [`rm(archivePath, {force: true})` in the `unpack: false` error path](https://github.com/yminod/puppeteer/blob/f587ef341d794faaf6ee9d84e9dc1b557e627e10/packages/browsers/src/install.ts#L397-L418) only removes the destination of a failed download after the locked code observed that the archive did not exist. Without it, a waiting installer could reuse a partial archive.

My previous note about transient errors referred to synthetic 1 ms retry stress: on Windows, `mkdir` occasionally returned a transient `EPERM` during lock handoff. This did not cause overlapping critical sections or cache corruption. The production 100 ms setting passed concurrent-install validation on both tested platforms. A small bounded retry could harden the synthetic case further if you would prefer it in this PR.

A re-review would be greatly appreciated, and I would be happy to address any remaining concerns about the design or stale-lock recovery.

packages/browsers/src/installLock.ts
```diff
@@ -0,0 +1,189 @@
...
+          await rm(reaperPath, {recursive: true, force: true});
+        }
+        return false;
+      }
+      if (isErrorWithCode(error, 'ENOENT')) {
+        return false;
+      }
+      throw error;
+    }
+    const claimedLockStats = lockStats.hasHeartbeat
```
OrKoN: Is this race condition possible?

  If a stale lock directory exists without a heartbeat file (hasHeartbeat === false—e.g., if the original process crashed between mkdir(lockPath) and refreshInstallLock):

  1. Process B calls installLock.ts, sees hasHeartbeat === false, and successfully creates reaperPath.
  2. Process C calls installLock.ts, also sees hasHeartbeat === false.
  3. Process B writes a new heartbeat file via installLock.ts, deletes reaperPath, returns true, and begins executing task().
  4. Process C now succeeds at mkdir(reaperPath) because Process B just deleted it.
  5. Because Process C's initial lockStats.hasHeartbeat was false, **line 116 skips calling installLock.ts**. Process C reuses its stale lockStats, assumes the lock is still unclaimed, overwrites Process B's heartbeat file, and begins executing task() concurrently. Both processes will later call rm(lockPath) in their finally blocks.

OrKoN: Perhaps it needs to be changed to always `const claimedLockStats = await installLockStats(lockPath);`?

packages/browsers/src/installLock.ts
```diff
@@ -0,0 +1,189 @@
...
+    const stats = await stat(lockPath);
+    return {mtimeMs: stats.mtimeMs, hasHeartbeat: false};
+  }
+}
+
+function canClaimInstallLock(
+  lockStats: {mtimeMs: number; ownerPid?: number},
+  staleThreshold: number,
+): boolean {
+  if (Date.now() - lockStats.mtimeMs <= staleThreshold) {
```
OrKoN: does it mean that if the install process was terminated the next one will have to wait for 5 minutes to claim the lock? that looks a bit extreme. I think pressing Ctrl + C during installation might be fairly common. How can we ensure we do not run into this?

OrKoN: Let's add an option to force bypass locking?

packages/browsers/src/installLock.ts
yminod: Thanks—yes, that interleaving is possible. I pushed an update that revalidates the lock after acquiring the reaper.

For a heartbeat-backed lock, the claimant now takes a fresh snapshot and repeats the freshness and owner-liveness checks. For a heartbeat-less lock, creating the nested `reaper` changes the parent lock directory's mtime, so a plain post-reaper stat is not sufficient. The claimant therefore verifies that no heartbeat appeared and that `dev`, `ino`, and `birthtimeNs` still identify the same directory generation before using the pre-reaper mtime for the stale check. If the lock was refreshed or recreated, it backs off.

The inspection path now also stats the heartbeat before reading its owner PID, so normal contention on a fresh lock remains stat-only. I added deterministic regression tests for both the two-claimant interleaving above and replacement of a heartbeat-less lock at the same path.

The focused install-lock test suite passed 20 consecutive runs on both Windows and macOS.

packages/browsers/src/installLock.ts
yminod: Since I could not find a reliable way to verify that no other process can still act on the installation target, I think it would be safer to keep the locking mechanism and explain that a retry is waiting for an existing installation lock. If the retry later fails, we could report the exact lock, archive, and output paths with recovery guidance. Would that address the concern, or is it important that the retry be able to proceed immediately?

packages/browsers/src/installLock.ts
```diff
@@ -0,0 +1,300 @@
...
+    }
+    await beforeStaleLockClaim?.();
+    debugInstall?.(`Claiming stale browser install lock at ${lockPath}`);
+    try {
+      await mkdir(reaperPath);
+    } catch (error) {
+      if (isErrorWithCode(error, 'EEXIST')) {
+        const reaperStats = await stat(reaperPath);
+        if (Date.now() - reaperStats.mtimeMs > staleThreshold) {
+          await rm(reaperPath, {recursive: true, force: true});
```
OrKoN: On Windows, recursive deletions of lock and reaper directories can occasionally throw transient EBUSY or EPERM errors if a virus scanner, search indexer, or background process momentarily opens a handle on files inside the directory during cleanup.

Similar to how Cache.ts handles folder deletions, we could pass {maxRetries: 5, retryDelay: 100} to rm(lockPath, ...) and rm(reaperPath, ...). In all instances of rm() here. Perhaps we can add a helper if there is no helper already.

packages/browsers/src/installLock.ts
```diff
@@ -0,0 +1,300 @@
...
+  after: InstallLockIdentity,
+): boolean {
+  return (
+    before.dev === after.dev &&
+    before.ino === after.ino &&
+    before.birthtimeNs === after.birthtimeNs
+  );
+}
+
+function canClaimInstallLock(
```
OrKoN: In canClaimInstallLock, if a process holding the lock crashes and its PID is recycled by a long-running daemon or system service (common in containerized environments like Docker/Kubernetes with small PID namespaces), process.kill(snapshot.ownerPid, 0) will never throw ESRCH.

Because the recycled process does not touch the heartbeat file, mtimeMs will remain stale, but canClaimInstallLock will return false on every retry, causing subsequent installs to spin in an infinite loop.

Consider adding a hard ceiling (e.g. 3 × staleThreshold or a maximum stale age) after which a lock is considered dead even if process.kill does not throw ESRCH, or logging a warning when a lock appears blocked by an alive PID with a very old heartbeat.

packages/browsers/src/installLock.ts
```diff
@@ -0,0 +1,300 @@
...
+          !canClaimInstallLock(currentSnapshot, staleThreshold)
+        ) {
+          return false;
+        }
+      } else {
+        const currentHeartbeatStats = await statHeartbeat(lockPath);
+        if (currentHeartbeatStats !== undefined) {
+          return false;
+        }
+        const currentLockIdentity = await installLockIdentity(lockPath);
```
OrKoN: In the heartbeat-less branch (initialSnapshot.fromHeartbeat === false), canClaimInstallLock(initialSnapshot, staleThreshold) checks initialSnapshot.mtimeMs, which is the timestamp captured before acquiring the reaper lock.

On filesystems where birthtimeNs is not supported (returns 0n, such as older Linux kernels, certain container overlay filesystems, or FAT/network mounts) and where inode numbers can be reused immediately upon recreation, isSameInstallLock may return true. Because initialSnapshot.mtimeMs is from before the reaper was acquired, this could lead to claiming a freshly recreated lock.

We should check the current directory's mtimeMs from a fresh stat call.

OrKoN:
While this design handles the most common concurrency scenarios on a single local machine (e.g., multiple Node processes or parallel threads calling install() at the same time), file-based locking inherently cannot solve every edge case without addressing several architectural limitations:
  ──────
  ### 1. The "Orphaned Lock Delay" Problem (SIGINT / SIGTERM / Crash)

  • What happens: If a process running install() is terminated via SIGINT (Ctrl+C), SIGTERM (Docker container shutdown, CI runner cancel), or an OOM crash, Node.js exits immediately without executing JavaScript finally blocks.
  • The consequence: The lock directory remains on disk. Any subsequent run on the machine must wait the full staleThreshold (5 minutes by default) before it can reclaim the lock.
  • Mitigation:
      • Register process signal handlers (SIGINT, SIGTERM, exit) to clean up active locks before exiting.
      • Or reduce the default staleThreshold from 5 minutes to something shorter (e.g., 30–60s), since heartbeat updates every 10s.

  ──────
  ### 2. Multi-Host / Shared Network Storage (NFS, SMB, EFS)

  • What happens: PIDs are local to a single OS kernel.
  • The consequence: If multiple machines/containers share a common cache directory over NFS or SMB:
      1. Machine A acquires the lock and writes PID 1234.
      2. Machine B checks process.kill(1234, 0) on Machine B. PID 1234 does not exist on Machine B (ESRCH), so Machine B assumes the lock owner is dead and steals the lock while Machine A is actively downloading.
      3. Clock drift between different servers can cause Date.now() - mtimeMs comparisons to misbehave.
  • Mitigation: Heartbeat files would need to store host identifiers (e.g., hostname + uuid + pid) and only perform kill(pid, 0) if hostname === os.hostname().
  ──────
  ### 3. Non-Atomic Archive Extraction (Corrupted Installation Directories)

  • What happens: unpackArchive(archivePath, outputPath) extracts directly into the final outputPath.
  • The consequence: If the process is killed midway through unpacking (e.g. OOM or power loss), the lock will eventually expire, but outputPath is left containing a partial, corrupted browser build. The next process acquiring the lock will either:
      • Crash with "The browser folder exists but the executable is missing", or
      • Find the executable, assume the installation is complete, and run a damaged binary missing internal components/shared libraries.
  • The Fix: Atomic Directory Swap. Extract the archive into a temporary folder (e.g. outputPath.tmp.<random>) and perform an atomic fs.rename to outputPath once extraction and integrity checks succeed.
  ──────
  ### 4. Same-Process In-Memory Efficiency

  • What happens: If 5 asynchronous calls to install() happen within the same Node.js process, all 5 tasks enter the filesystem retry loop, hitting mkdir, stat, and sleep(100) repeatedly.
  • Improvement: An in-memory deduplication map (Map<string, Promise<InstalledBrowser>>) at the process level can immediately join identical in-flight install promises without any disk I/O polling.
  
  
Regarding #3, I wonder if it makes sense to tackle it first in a separate PR?. Also, the explicit resource management recently became a Stage 4 proposal so perhaps we can use for the locking with an API like this:

```
using lock = await acquireLock();
// ... do installation
// ... lock is released automatically when the scope ends because it implements dispose/asyncDispose 
```

packages/browsers/src/installLock.ts
yminod: All lock and reaper directory removals now use bounded recursive cleanup with `maxRetries: 5` and `retryDelay: 100`.

packages/browsers/src/installLock.ts
yminod: I used a bounded acquisition timeout rather than treating lock age as a takeover deadline, because age alone cannot safely override an owner that still appears alive. Puppeteer warns once when stale recovery is blocked, then returns `InstallLockError` after 15 minutes without claiming or removing the lock or starting the installation task.

packages/browsers/src/installLock.ts
yminod: Creating the nested reaper changes the lock directory's mtime, so a post-reaper mtime check cannot distinguish that expected change from recreation. A heartbeat-less lock is now reclaimed only when it has a usable birth time and its device, inode, and birth time remain stable across reaper acquisition; unavailable or unstable identity is left unclaimed.

yminod:
Thanks for the detailed review.

The stale threshold is now 60 seconds, or six heartbeat intervals. Reaching that threshold only starts recovery checks and does not by itself make the lock claimable. Recovery remains fail-closed: a stale heartbeat lock is reclaimed only when its recorded owner is on the local host and is confirmed to be gone. Foreign-host or otherwise unverifiable ownership is left untouched, so this remains a local-machine lock rather than a distributed lock for shared filesystems.

I agree that staging and atomic publication are better handled separately from this lock change.

I kept same-process deduplication separate because callers may use different installation behavior and options. The callback API already scopes lock release to the task lifetime; changing it to `await using` would not change crash recovery.

yminod:
@OrKoN, I've pushed the latest changes and responded to the review comments. The PR is ready for another look whenever you have a chance.

OrKoN:
WDYT about changing the following?

1. Add a hard ceiling for stale locks: If a lock's heartbeat has not been updated for a hard maximum (e.g. 3 × staleThreshold or 3–5 minutes), allow it to be claimed regardless of whether process.kill(pid, 0) succeeds or the hostname matches.
  2. Check owner liveness without waiting 60s: On lock contention, immediately read the heartbeat and check process.kill(pid, 0). If the owner PID is already dead on the local machine (ESRCH), claim the lock immediately instead of waiting for staleThreshold to elapse.
  3. Write heartbeats atomically: Write to heartbeat.tmp and rename to prevent corrupted 0-byte heartbeats on process aborts.
  4. Treat invalid metadata as stale: If the heartbeat file is corrupted or empty and older than staleThreshold, treat it as stale and reclaimable rather than permanently blocking.
  5. Protect cleanup in finally: Catch errors in removeInstallLockPath during finally so they do not swallow underlying installation errors or discard successful installations.
  6. Register exit/signal cleanup: Hook process.once('SIGINT') and process.once('SIGTERM') to remove the lock directory when interrupted.

yminod:
I think adding staging alongside the lock changes you suggested would make recovery more responsive. If that works for you, I'd like to take on both, keeping staging in a separate PR as you proposed earlier. I'd develop and test the changes together, with staging merged first.

For points 1, 2 and 4, I'd account for previous work continuing after takeover. A local PID check alone does not rule this out: helper processes may survive the owner, and matching hostnames do not guarantee a shared PID namespace. Staging would isolate each attempt's downloads and extraction, publishing only completed output while preserving a competing winner. The lock changes would also need to ensure that heartbeat updates and cleanup from an earlier ownership generation cannot affect a later generation's lock.

For point 3, I'd publish owner metadata atomically via a temporary file and rename on acquisition/takeover, then use `utimes` for periodic updates. Since the metadata stays unchanged during each ownership period, this avoids repeatedly replacing the heartbeat file. For point 5, cleanup would preserve the original installation result or error.

For point 6, I think it would be better to discuss signal cleanup separately, particularly how it would interact with existing signal handling in applications using the library.

Would it work for you to merge staging first, then complete #15218 with points 1–5, while discussing point 6 separately?

OrKoN:
> Would it work for you to merge staging first, then complete https://github.com/puppeteer/puppeteer/pull/15218 with points 1–5, while discussing point 6 separately?

Sounds good! Thanks!

PR #15222, opened 2026-07-09
docs: document Linux headless screen sizing flag

### LINKED ISSUE #12655: [Bug]: Documentation for ozone-override-screen-size is missing

```TypeScript
import Puppeteer from "puppeteer";

const browser = await Puppeteer.launch({
  // Puppeteer default viewport is 800x600. By setting to null, the viewport will be full size of the browser window.
  defaultViewport: null,
  headless: true,
  args: ["--window-size=1280,1024", "--headless=new"],
});

// Navigate the page to a URL.
await page.goto("https://whatismyviewport.com/");

await page.screenshot({
  path: "what-is-my-viewport.png",
});

await browser.close();
```

### Error string

no error

### Bug behavior

- [x] Consistent
- [ ] Flaky
- [ ] PDF

### Background

Running the example script under Mac and Linux gives different results

### Expectation

Running chrome headless, setting the window size should mean that both the browser window / viewport & screen dimensions will be set to the same provided value

### Reality

The browser window is sized correctly, however the screen size itself is not. 

This also means that `window.screen.width` & `window.screen.height` are also incorrect.

The issue is not present on Mac, but IS present on Linux.

| Mac (correct screen size) | Linux (incorrect screen size) |
| --- | --- |
| what-is-my-viewport-mac | what-is-my-viewport-ubuntu |

_No response_

### Puppeteer version

22.12.0, likely others

### Node version

Any

### Package manager

npm

### Package manager version

Latest

### Operating system

Linux

iamdhrv:
## Summary

- document the Linux headless `window.screen` sizing caveat
- show using `--ozone-override-screen-size` together with `--window-size`
- mirror the guide update in the versioned docs

Closes #12655

## Validation

- `git diff --check`


docs/guides/window-management.md
```diff
@@ -110,6 +110,19 @@ Inner size: 600x400
 Outer size: 600x487
 ```
 
+On Linux in headless mode, Chrome may report `window.screen` dimensions that do
+not match the configured window size. If your script depends on those screen
+dimensions, pass `--ozone-override-screen-size=WIDTH,HEIGHT` together with the
+window size flag:
+
+```ts
```
OrKoN, line 119: Thanks for the PR. Screen management is documented in another doc https://pptr.dev/guides/screen-configuration Do we really need to document `ozone-override-screen-size`? How is it different from `--screen-info` (other than `ozone-override-screen-size` is likely not cross-platform)?

OrKoN (changes requested):

docs/guides/window-management.md
iamdhrv: Good question -- dug into this a bit more before replying.

--ozone-override-screen-size is a lower-level Chromium/Ozone flag (ui/ozone/public/ozone_switches.cc) that overrides what the Ozone platform layer itself reports as the screen size. It's Linux/Ozone-specific (X11/Wayland), not cross-platform.

--screen-info is the higher-level, headless-only API for configuring virtual displays (dimensions, position, labels, orientation) across platforms, but on Linux it doesn't fix the underlying issue in #12655 -- window.screen.width/height still don't reflect --window-size in headless mode. Pairing --ozone-override-screen-size with --window-size is the actual workaround for that specific Linux bug, so I think it's worth keeping documented as a distinct caveat rather than folding it into the --screen-info guide.

Happy to add a short cross-link between the two docs so it's clear when to reach for which one -- let me know if that addresses your concern, and I'll also rebase to clear the conflict.

OrKoN: Could you please clarify why should `--window-size` be reflected in window.screen.width/height? Does the config passed `--screen-info` get reflected in `window.screen.width/height`?

OrKoN: I believe the screen-info is working as expected here:

```

  args: ['--screen-info={800x600}'],
});

const pages = await browser.pages();

console.log(await pages[0].evaluate(() => {
  return `${window.screen.width}x${window.screen.height}`
}))

// outputs 800x600
```

so I do not think we need to document --ozone-override-screen-size as screen-info works as expected on headless Linux.

iamdhrv:
Closing based on the maintainer verification that Chromium's supported `--screen-info={800x600}` flag provides the requested headless screen dimensions. Since `--ozone-override-screen-size` does not need to be documented, this change is no longer necessary. Thanks for testing and clarifying the supported approach.

PR #15231, opened 2026-07-14
fix: release the mouse button when a drag or drop fails

Liohtml:
`ElementHandle.drag()` presses the mouse button down, and `ElementHandle.drop()` is the only thing that releases it. If anything in between throws, the button stays pressed for the rest of the session.

Filed as a draft because I'd like a sanity check on the approach before it's polished — happy to take it in a different direction. Context: ChromeDevTools/chrome-devtools-mcp#2353, where @OrKoN suggested fixing it here.

## It doesn't need a protocol failure

A page that re-renders while the drag is in flight detaches the dragged node, so `drop()` throws `Node is detached from document` out of `scrollIntoViewIfNeeded()`. Any reactive sortable list does exactly that — which is the main use case for drag and drop.

```js
await from.drag(to);     // presses the mouse button
await to.drop(from);     // throws: Node is detached from document
// left mouse button is still down
```

## Why it doesn't heal

Nothing resets it. `Mouse#_state` lives on the `Mouse`, which hangs off the `Page`, so an ordinary navigation swaps the document and leaves the pressed button behind. (`Mouse.reset()` exists and would do the right thing, but nothing calls it.)

What the caller actually sees, all reproduced against Chrome:

```
after the failed drag        : [ 'mousedown', 'selectstart' ]
then ONE plain click()       : [ 'mouseup', 'click', 'mousedown', 'mouseup', 'click' ]
```

* **The next click fires the target's handler twice.** The stale `mousedown` pairs with the click's `mouseup`, the browser synthesises a `click` from it, and then the intended click lands. An automation that clicks *Delete* after a failed drag hits it twice.
* Moving the mouse drags a text selection across the page (`selectstart`).
* After a navigation the stale press is delivered into the fresh document as a stray `mouseup`.

## The change

Release the button in the two places that can strand it:

* `drag()` — when *this* call is the one that pressed it, and the `drop()` it was heading for will never run.
* `drop()` — when the `drag()` it re-enters throws.

Neither release is allowed to mask the error that triggered it (`.catch(debugError)`), and the success path is untouched. The two are mutually exclusive, so there's no double release: in the `drop()` flow the re-entered `drag()` sees `_isDragging` already set and doesn't press, so it doesn't release either.

## Test

`should release the mouse button when the drop fails` in `drag-and-drop.test.ts` — re-renders the page mid-drag so the drop fails naturally, then asserts `mouse.up()` rejects with `'left' is not pressed`, i.e. the button was already released. It fails on `main` and passes here.

`Drag n' Drop should drag and drop` was already failing before this change; it's listed as an expected `FAIL` for cdp/chrome in `TestExpectations.json`. Untouched by this PR, but worth flagging that it sits in the same area.

### DIFF
--- a/packages/puppeteer-core/src/api/ElementHandle.ts
+++ b/packages/puppeteer-core/src/api/ElementHandle.ts
@@ -18,7 +18,11 @@ import type {
   NodeFor,
 } from '../common/types.js';
 import type {KeyInput} from '../common/USKeyboardLayout.js';
-import {isString, withSourcePuppeteerURLIfNone} from '../common/util.js';
+import {
+  debugError,
+  isString,
+  withSourcePuppeteerURLIfNone,
+} from '../common/util.js';
 import {assert} from '../util/assert.js';
 import {AsyncIterableUtil} from '../util/AsyncIterableUtil.js';
 import {throwIfDisposed} from '../util/decorators.js';
@@ -838,11 +842,13 @@ export abstract class ElementHandle<
       }
       return await page.mouse.drag(source, target);
     }
+    let pressedMouse = false;
     try {
       if (!page._isDragging) {
         page._isDragging = true;
         await this.hover();
         await page.mouse.down();
+        pressedMouse = true;
       }
       if (target instanceof ElementHandle) {
         await target.hover();
@@ -851,6 +857,13 @@ export abstract class ElementHandle<
       }
     } catch (error) {
       page._isDragging = false;
+      if (pressedMouse) {
+        // The mouse button was pressed for this drag, and `drop()` - the only
+        // thing that releases it - will never run. Releasing it here keeps the
+        // button from staying pressed for the rest of the session. It must not
+        // mask the error that got us here.
+        await page.mouse.up().catch(debugError);
+      }
       throw error;
     }
   }
@@ -921,7 +934,18 @@ export abstract class ElementHandle<
     } else {
       // Note if the rest errors, we still want dragging off because the errors
       // is most likely something implying the mouse is no longer dragging.
-      await dataOrElement.drag(this);
+      try {
+        await dataOrElement.drag(this);
+      } catch (error) {
+        // The preceding `drag()` pressed the mouse button down and this is the
+        // only place that releases it, so bailing out here would leave it
+        // pressed for the rest of the session: the next click would then fire
+        // twice, and moving the mouse would drag a selection across the page.
+        // Releasing must not mask the error that got us here.
+        page._isDragging = false;
+        await page.mouse.up().catch(debugError);
+        throw error;
+      }
       page._isDragging = false;
       await page.mouse.up();
     }
--- a/packages/puppeteer-core/src/api/ElementHandle.ts
+++ b/packages/puppeteer-core/src/api/ElementHandle.ts
@@ -833,22 +833,25 @@ export abstract class ElementHandle<
     this: ElementHandle<Element>,
     target: Point | ElementHandle<Element>,
   ): Promise<Protocol.Input.DragData | void> {
-    await this.scrollIntoViewIfNeeded();
     const page = this.frame.page();
     if (page.isDragInterceptionEnabled()) {
+      await this.scrollIntoViewIfNeeded();
       const source = await this.clickablePoint();
       if (target instanceof ElementHandle) {
         target = await target.clickablePoint();
       }
       return await page.mouse.drag(source, target);
     }
-    let pressedMouse = false;
+    // The button is down either because an earlier `drag()` pressed it, or
+    // because this call is about to.
+    let isMouseDown = page._isDragging;
     try {
+      await this.scrollIntoViewIfNeeded();
       if (!page._isDragging) {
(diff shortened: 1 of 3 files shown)


Liohtml:
CI caught a real flaw in my test, now fixed.

I was asserting the release by expecting `mouse.up()` to reject with `'left' is not pressed`. That is a `CdpMouse` guard — `BidiMouse.up()` has no equivalent, it just performs the action — so the assertion passed vacuously on cdp and blew up on chrome-bidi and firefox. My fault: I asserted an implementation detail rather than the behaviour.

Now it asserts what the page itself sees: move the mouse afterwards and read `event.buttons` off the resulting `mousemove`. It must be `0`.

Re-verified locally against both protocols, and this time properly against the pre-fix source (my first attempt stashed a file that was already committed, so the "unfixed" run silently still had the fix in it):

```
unfixed, cdp           -> FAIL   Expected: 0  Received: 1
unfixed, webDriverBiDi -> FAIL   Expected: 0  Received: 1
fixed,   cdp           -> pass
fixed,   webDriverBiDi -> pass
```

Which also settles something worth knowing: the stranded button is **not** CDP-specific, it reproduces on BiDi too.

Liohtml:
Firefox is green now too. The remaining failure was not what it looked like: the new test **passes** on firefox, but `[drag-and-drop.test] *` is expected to `FAIL` there, so a passing test in that file gets reported as an unexpected result.

Added a `PASS` expectation for just this test rather than widening the wildcard — it only asserts that the mouse button is not left pressed after a failed drop, which firefox handles fine, so it does not depend on the drag-and-drop behaviour the blanket expectation is covering.

(The other red on `firefox-headless (3-4)` was `[browsercontext-cookies.test] BrowserContext.setCookie should set cookie with a partition key`, which is unrelated to this PR.)

OrKoN:
@Liohtml is this PR ready for review? it's marked as draft.

Liohtml:
Yes — ready for review now, thanks for the nudge.

It was a draft because I wanted a sanity check on the approach before polishing, but nothing is outstanding: the fix, the regression test, and the expectations are all in. I've rebased onto main and CI is green.

The one thing worth a reviewer's opinion is the shape of the fix: releasing in both `drag()` (when that call is the one that pressed the button) and `drop()` (when the `drag()` it re-enters throws), with `.catch(debugError)` so neither release masks the original error. Happy to take it in a different direction if you'd rather see the reset live somewhere else — e.g. `Mouse.reset()` being called on navigation, which would also cover strays from other sources.

packages/puppeteer-core/src/api/ElementHandle.ts
```diff
@@ -921,7 +934,18 @@ export abstract class ElementHandle<
...
+      try {
+        await dataOrElement.drag(this);
+      } catch (error) {
+        // The preceding `drag()` pressed the mouse button down and this is the
+        // only place that releases it, so bailing out here would leave it
+        // pressed for the rest of the session: the next click would then fire
+        // twice, and moving the mouse would drag a selection across the page.
+        // Releasing must not mask the error that got us here.
+        page._isDragging = false;
+        await page.mouse.up().catch(debugError);
```
OrKoN: it looks like dataOrElement.drag would already release the mouse by calling mouse.up(), is this change necessary?

Liohtml: Good question — and it turned up a real bug in mine.

It doesn't release in the flow this fixes. After a successful `from.drag(to)`, `to.drop(from)` re-enters `drag()` with `_isDragging` already `true`, so it never presses and has nothing of its own to release. And it throws out of `scrollIntoViewIfNeeded()` (the detached node), which sat *outside* the `try` — so that `catch` never ran at all.

You're right that they overlap in the other flow though: `drop()` with no preceding `drag()`. There the inner `drag()` does press and release, and my `drop()` catch then released a second time. CDP swallows it (`'left' is not pressed`), but `BidiMouse.up()` has no such guard and would have dispatched a stray `mouseup`. That was a bug.

So I've done what you're pointing at: `scrollIntoViewIfNeeded()` moved inside the `try`, and the flag now starts as `page._isDragging` — the button is down if an earlier `drag()` pressed it *or* if this call is about to. `drop()` is untouched again and the diff is half the size.

Verified fails-before / passes-after on both `cdp` and `webDriverBiDi`.

OrKoN:
@Liohtml could you please rebase the PR?

Liohtml:
Rebased onto `main` (101c7d501).

One thing worth flagging: the rebase was conflict-free but didn't build. `debugError` was removed from `common/util.ts` by the logger refactor (#15304, #15305, #15312), and `drag()` used it to swallow the `mouse.up()` error without masking the original one. `ElementHandle` already carries a `Logger` via `JSHandle`, so it now goes through that instead:

```ts
await page.mouse.up().catch(error => {
  this.logger(DEBUG_PREFIXES.error)?.(error);
});
```

Kept as a separate commit so it's clear it's only the adaptation. Nothing else changed — the refactors didn't touch input or drag behaviour.

OrKoN (approved):

PR #15232, opened 2026-07-14
feat: expose high-level PWA APIs

### LINKED ISSUE #15202: [Feature]: Expose PWA CDP domain via high-level Browser/Page APIs
### Feature description

**Expose Chromium's browser-scoped `PWA.*` CDP domain through high-level Puppeteer methods**, so consumers can install, launch, and uninstall Progressive Web Apps without hand-rolling a raw CDP session.

> Requested by a Puppeteer/DevTools maintainer on [ChromeDevTools/chrome-devtools-mcp#2270](https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2270) — the ask was to land this in **Puppeteer first**, then have `chrome-devtools-mcp` wrap it. This issue tracks the Puppeteer side.

---

### Why

Chromium ships a full browser-scoped `PWA` CDP domain (`install`, `uninstall`, `launch`, `openCurrentPageInApp`, `getOsAppState`, `changeAppUserSettings`, `launchFilesInApp`). Because it's browser-scoped rather than page-scoped, the only way to reach it from Puppeteer today is to open and manage a raw CDP session by hand:

```js
// Today — works, but every call manages its own browser-level session
const session = await browser.target().createCDPSession();
await session.send('PWA.install', {manifestId});
await session.detach();
```

```js
// Proposed — driven like any other Puppeteer method
const manifestId = await page.installAsPWA();
const appPage = await browser.launchPWA({manifestId});
// …test behavior in the app window…
await browser.uninstallPWA({manifestId});
```

A native API removes the per-call session bookkeeping and brings PWA lifecycle in line with the rest of Puppeteer's surface — which is what tooling like `chrome-devtools-mcp` and agent-driven "install / launch / test this app" workflows need.

**Already typed:** Puppeteer pins a `devtools-protocol` version that already includes the full `PWA` domain, so the implementation can call `this.#connection.send('PWA.install', …)` with no casting.

**Clear precedent:** this mirrors the recently-added `installExtension`, which follows exactly the same three-layer shape — `abstract` in `api/Browser.ts`, CDP impl in `cdp/Browser.ts`, `UnsupportedOperation` stub in `bidi/`.

---

### Proposed API

| Method | CDP | Returns |
| --- | --- | --- |
| `browser.installPWA({manifestId, installUrlOrBundleUrl?, displayMode?})` | `PWA.install` (+ `changeAppUserSettings` for `displayMode`) | `Promise<string>` — the manifest id |
| `browser.uninstallPWA({manifestId})` | `PWA.uninstall` | `Promise<void>` |
| `browser.launchPWA({manifestId, url?})` | `PWA.launch` → resolve `targetId` | `Promise<Page>` |
| `browser.getPWAState({manifestId})` | `PWA.getOsAppState` | `Promise<{badgeCount, fileHandlers}>` |
| `page.installAsPWA({displayMode?})` | derive manifest id → `PWA.install` | `Promise<string>` — the manifest id |
| `page.openInPWA()` | derive manifest id → `PWA.openCurrentPageInApp` | `Promise<void>` |

The whole lifecycle is keyed on `manifestId`, so the **install methods return it** — including `page.installAsPWA()`, which derives the id from the current page. That lets a caller install an app and then reliably launch / query / uninstall *that specific app* (essential when an agent is cycling through several PWAs with different manifests). `PWA.install` itself returns `void`, but the id is already known client-side (passed in, or derived), so returning it needs no CDP change.

`browser.getPWAState()` (from `PWA.getOsAppState`) is a cheap read that lets a test **assert** install/uninstall actually took effect, and inspect badge count / file handlers — rather than inferring app state from side effects.

`launchPWA` can reuse Puppeteer's existing target-resolution pattern — `waitForTarget(t => t._targetId === targetId)` → `target.page()` (as in `_getDevToolsTargetPage`) — to turn `PWA.launch`'s returned `targetId` into a `Page` you can drive directly.

<details>
<summary><strong>Note on <code>displayMode</code></strong></summary>

`PWA.install` alone leaves the app's `user_display_mode` at Chromium's default (`browser`), so opening in a standalone window requires a chained `PWA.changeAppUserSettings({displayMode: 'standalone'})`. A native `displayMode` param on `PWA.install` in Chromium would be a cleaner long-term fix, but that's a separate upstream change.
</details>

---

### Open questions

- **Signature style** — options-object (shown above) vs positional args. Options-object matches `installExtension(path, options)` and scales better for `displayMode` / `url`.
- **Page-level manifest-id derivation** — proper manifest fetch + parse, vs defaulting to the document URL for id-less manifests.
- **Scope** — `launchFilesInApp` is left out for now; happy to add it if there's interest.

---

### Prior art / validation

Chromium's own ChromeOS developer docs already ship a **PWA lifecycle test that drives this exact domain through Puppeteer** — via the hand-rolled raw-session approach this issue aims to replace (`docs/.../testing/web-testing/pwa/test-pwa-lifecycle.js`):

```js
const session = await browser.target().createCDPSession();
await session.send('PWA.install', {manifestId, installUrlOrBundleUrl});
await session.send('PWA.launch', {manifestId});
// …inspect state, then PWA.uninstall
```

So the install → launch → inspect → uninstall workflow (keyed on `manifestId`) is already the documented pattern; a native API would just remove the per-call session boilerplate and add typing.

I've also prototyped these as `chrome-devtools-mcp` tools driving a local Chromium build the same way, and verified the full lifecycle (including the `displayMode` standalone fix) against `chrome://web-app-internals`.

Happy to open a PR implementing the native Puppeteer APIs.

jstar0:

Feature.

Yes. The new Chrome pipe integration tests cover all six APIs, including PWA
installation, launch, opening the current page as an app, OS state inspection,
uninstallation, display mode changes, invalid page state, and the WebDriver
BiDi unsupported paths.

Yes. The public API TSDoc marks the methods experimental and documents that the
PWA domain is currently available only for Chrome pipe connections.

## Summary

Fixes #15202.

Chromium exposes PWA lifecycle operations through CDP, but Puppeteer callers
currently need to create raw sessions and manually map launched app targets back
to pages.

This adds high-level `Browser` APIs for installing, uninstalling, launching,
and inspecting PWAs, plus `Page` APIs for installing or opening the current
page as a PWA. The CDP implementation uses the required browser or page session
scope and maps a launched app to the correct driveable `Page`. WebDriver BiDi
reports these operations as unsupported for now.

## Changes

- Add typed Browser APIs for PWA installation, uninstallation, launch, and OS
  state inspection.
- Add Page APIs that derive the manifest ID and install or open the current PWA.
- Apply optional display mode changes after installation.
- Resolve Chromium's launched app-tab target to the corresponding Puppeteer
  page.
- Add real-browser CDP coverage and explicit WebDriver BiDi unsupported paths.

No. It adds experimental APIs without changing existing behavior.

**Other information**

`launchFilesInApp` is intentionally outside this change, as discussed in the
issue.

## Verification

```bash
npm run build --workspace @puppeteer-test/test
npm test -- --test-suite chrome-pipe --grep '^PWA '
npm run test-types
npm run unit --workspace puppeteer-core
npm run lint
```

### DIFF
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -12,6 +12,10 @@ import type {
   CreatePageOptions,
   DebugInfo,
   ExtensionInstallOptions,
+  PWAInstallOptions,
+  PWALaunchOptions,
+  PWAState,
+  PWAUninstallOptions,
 } from '../api/Browser.js';
 import {
   Browser as BrowserBase,
@@ -521,6 +525,58 @@ export class CdpBrowser extends BrowserBase {
     await Promise.all(targetDestroyedPromises);
 
     this.#extensions.delete(id);
+  }
+
+  async _installPWA(
+    options: PWAInstallOptions,
+    session?: CdpCDPSession,
+  ): Promise<string> {
+    const {displayMode, ...installOptions} = options;
+    const client = session ?? this.#connection;
+    await client.send('PWA.install', installOptions);
+    if (displayMode) {
+      await client.send('PWA.changeAppUserSettings', {
+        manifestId: options.manifestId,
+        displayMode,
+      });
+    }
+    return options.manifestId;
+  }
+
+  override async installPWA(options: PWAInstallOptions): Promise<string> {
+    return await this._installPWA(options);
+  }
+
+  override async uninstallPWA(options: PWAUninstallOptions): Promise<void> {
+    await this.#connection.send('PWA.uninstall', options);
+  }
+
+  override async launchPWA(options: PWALaunchOptions): Promise<Page> {
+    const {targetId: launchedTargetId} = await this.#connection.send(
+      'PWA.launch',
+      options,
+    );
+    const target = (await this.waitForTarget(candidate => {
+      // PWA.launch returns the tab target, while Puppeteer exposes its page child.
+      const launchedTarget = this.#targetManager
+        .getAvailableTargets()
+        .get(launchedTargetId);
+      return (
+        candidate.type() === 'page' &&
+        launchedTarget?._childTargets().has(candidate as CdpTarget) === true
+      );
+    })) as CdpTarget;
+    const page = await target.page();
+    if (!page) {
+      throw new Error(
+        `Failed to create a page for PWA target (id = ${launchedTargetId})`,
+      );
+    }
+    return page;
+  }
+
+  override async getPWAState(options: {manifestId: string}): Promise<PWAState> {
+    return await this.#connection.send('PWA.getOsAppState', options);
   }
 
   override async screens(): Promise<ScreenInfo[]> {
--- a/packages/puppeteer-core/src/api/Browser.ts
+++ b/packages/puppeteer-core/src/api/Browser.ts
@@ -335,6 +335,111 @@ export interface ExtensionInstallOptions {
 }
 
 /**
+ * @public
+ */
+export type PWADisplayMode = 'standalone' | 'browser';
+
+/**
+ * @public
+ */
+export interface PWAInstallOptions {
+  /**
+   * The ID from the web app's manifest.
+   */
+  manifestId: string;
+
+  /**
+   * The app or bundle location to use instead of the one derived from the
+   * manifest ID.
+   */
+  installUrlOrBundleUrl?: string;
+
+  /**
+   * Whether to open the installed app in a browser tab or standalone window.
+   *
+   * @remarks
+   * Chrome applies this setting after installing the app. If applying the
+   * setting fails, the method rejects but the app remains installed.
+   */
+  displayMode?: PWADisplayMode;
+}
+
+/**
+ * @public
+ */
+export interface PWAUninstallOptions {
+  /**
+   * The ID from the web app's manifest.
+   */
+  manifestId: string;
+}
+
+/**
+ * @public
+ */
+export interface PWALaunchOptions {
+  /**
+   * The ID from the web app's manifest.
+   */
+  manifestId: string;
+
+  /**
+   * The URL to open in the launched app.
+   */
+  url?: string;
+}
+
+/**
+ * @public
+ */
+export interface PWAFileHandlerAccept {
+  /**
+   * The MIME type accepted by the file handler.
+   */
+  mediaType: string;
+
+  /**
+   * The file extensions accepted for the MIME type.
+   */
+  fileExtensions: string[];
+}
+
+/**
+ * @public
+ */
+export interface PWAFileHandler {
+  /**
+   * The URL handling the files.
+   */
+  action: string;
+
+  /**
+   * The MIME types and file extensions accepted by the handler.
+   */
+  accepts: PWAFileHandlerAccept[];
+
+  /**
+   * The handler name displayed to the user.
+   */
+  displayName: string;
+}
+
+/**
+ * @public
+ */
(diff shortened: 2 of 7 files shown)


packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -523,6 +528,95 @@ export class CdpBrowser extends BrowserBase {
     this.#extensions.delete(id);
   }
 
+  async #withPWASession<T>(
```
OrKoN: Thanks for the PR. Why is #withPWASession needed? Can we use the connection on the browser directly instead?

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -523,6 +528,95 @@ export class CdpBrowser extends BrowserBase {
...
+  }
+
+  override async launchPWA(options: PWALaunchOptions): Promise<Page> {
+    const {targetId: launchedTargetId} = await this.#withPWASession(
+      async session => {
+        return await session.send('PWA.launch', options);
+      },
+    );
+    const target = (await this.waitForTarget(candidate => {
+      const launchedTarget = this.#targetManager
```
OrKoN, line 561: is there a reason to get the target from the manager instead of checking targetId on the candidate target?

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -523,6 +528,95 @@ export class CdpBrowser extends BrowserBase {
...
+        .getAvailableTargets()
+        .get(launchedTargetId);
+      return (
+        candidate.type() === 'page' &&
+        launchedTarget?._childTargets().has(candidate as CdpTarget) === true
+      );
+    })) as CdpTarget;
+    const initialized =
+      (await target._initializedDeferred.valueOrThrow()) ===
```
OrKoN: I think this will be handled by the `.page()` getter?

packages/puppeteer-core/src/cdp/Browser.ts
jstar0: Updated in `0e4e97b7`. Browser-level PWA commands now use `#connection` directly, and `#withPWASession` is removed. The one page-client path is `Page.installAsPWA`: Chromium rejects `PWA.install` without an install URL unless the DevTools session is associated with that page, so it reuses the page's existing primary client rather than creating a new session. The Chrome pipe PWA suite passes 7/7.

packages/puppeteer-core/src/cdp/Browser.ts
jstar0: `PWA.launch` returns the tab target ID, while Puppeteer does not expose tab targets through `waitForTarget`; the candidate is the child page target and has a different ID. The manager lookup maps the returned tab to its exact page child, which also keeps concurrent launches deterministic. I added a short comment at the predicate to make that distinction explicit.

packages/puppeteer-core/src/cdp/Browser.ts
jstar0: Updated in `0e4e97b7`. I removed the direct `_initializedDeferred` check. `waitForTarget` only emits an initialized exposed target here, and `target.page()` handles Page creation. The Chrome pipe launch regression passes.

OrKoN:
Thanks for the PR but there was a concurrent PR from the feature request reporter that iterated faster so we landed it instead https://github.com/puppeteer/puppeteer/pull/15235 Sorry about that!

PR #15235, opened 2026-07-15
feat: add browser-level PWA install/launch/uninstall APIs

LuHuangMSFT:
## What / Why

Adds first-class **Progressive Web App (PWA) automation** to the `Browser` API, wrapping Chromium's browser-scoped `PWA.*` CDP domain so callers can install, launch, inspect, and uninstall PWAs without managing raw sessions.

Implements the API agreed on in #15202.

## API

```ts
browser.installPWA({manifestId, installUrlOrBundleUrl, displayMode?}) // => manifestId
browser.uninstallPWA({manifestId})                                    // => void
browser.launchPWA({manifestId, url?, timeout?})                       // => Page
browser.getPWAState({manifestId})                                     // => {badgeCount, fileHandlers}
```

## Behavior and compatibility

- Chromium/CDP only; WebDriver BiDi implementations throw `UnsupportedOperation`.
- The `PWA` domain requires an opt-in pipe connection (`puppeteer.launch({pipe: true})`) and is unavailable over WebSocket.
- `installUrlOrBundleUrl` is required because the browser-scoped session has no associated page from which Chromium could derive the install URL.
- `launchPWA` uses the tab target id returned by `PWA.launch` to resolve its child page target. Existing app windows that Chromium focuses resolve to their existing `Page`.
- `timeout` controls how long `launchPWA` waits for the app's page target. The protocol command uses the configured `protocolTimeout`; `0` disables the target-wait timeout.

## Testing

The PWA integration suite covers lifecycle/state, default and explicit-URL launch, and standalone display mode. Unsupported Firefox, BiDi, BiDi-only, and headless-shell configurations are expected-skipped. Focused unit coverage verifies that the launch timeout is applied to target waiting without overriding the protocol timeout.

Validated with Chrome for Testing 150 and a Chromium/Edge build in headless and headful modes. The exact PR build was also consumed by a local `chrome-devtools-mcp` PWA category and passed a real stdio MCP lifecycle: install, inspect OS state, launch/select the returned `Page`, execute a page script in standalone mode, uninstall, and verify state rejection. Generated docs, API extraction, formatting, and lint checks pass.

Refs #15202

### DIFF
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -549,25 +549,70 @@ export class CdpBrowser extends BrowserBase {
   }
 
   override async launchPWA(options: LaunchPWAOptions): Promise<Page> {
-    const {targetId} = await this.#connection.send('PWA.launch', {
+    // `PWA.launch` returns the id of the launched app-window target, which is
+    // not surfaced through the `Target` domain and therefore never appears in
+    // `browser.targets()`. The app's actual page (web contents) is a separate
+    // target with its own id. So instead of matching the returned id, we wait
+    // for the new page target that the launch creates, scoped to the app's
+    // origin (a PWA's launch URL is same-origin as its manifest id).
+    const appOrigin = (() => {
+      try {
+        return new URL(options.url ?? options.manifestId).origin;
+      } catch {
+        return undefined;
+      }
+    })();
+    const knownPageTargetIds = new Set(
+      this.targets()
+        .filter(target => {
+          return target.type() === 'page';
+        })
+        .map(target => {
+          return (target as CdpTarget)._targetId;
+        }),
+    );
+    await this.#connection.send('PWA.launch', {
       manifestId: options.manifestId,
       url: options.url,
     });
-    const target = (await this.waitForTarget(t => {
-      return (t as CdpTarget)._targetId === targetId;
-    })) as CdpTarget;
+    const target = (await this.waitForTarget(
+      candidate => {
+        if (candidate.type() !== 'page') {
+          return false;
+        }
+        const targetId = (candidate as CdpTarget)._targetId;
+        if (knownPageTargetIds.has(targetId)) {
+          return false;
+        }
+        if (appOrigin === undefined) {
+          return true;
+        }
+        try {
+          return new URL(candidate.url()).origin === appOrigin;
+        } catch {
+          return false;
+        }
+      },
+      {timeout: options.timeout},
+    )) as CdpTarget;
     if (!target) {
-      throw new Error(`Missing target for launched PWA (id = ${targetId})`);
+      throw new Error(
+        `Missing target for launched PWA (manifestId = ${options.manifestId})`,
+      );
     }
     const initialized =
       (await target._initializedDeferred.valueOrThrow()) ===
       InitializationStatus.SUCCESS;
     if (!initialized) {
-      throw new Error(`Failed to create target for PWA (id = ${targetId})`);
+      throw new Error(
+        `Failed to create target for PWA (manifestId = ${options.manifestId})`,
+      );
     }
     const page = await target.page();
     if (!page) {
-      throw new Error(`Failed to create a page for PWA (id = ${targetId})`);
+      throw new Error(
+        `Failed to create a page for PWA (manifestId = ${options.manifestId})`,
+      );
     }
     return page;
   }
--- a/packages/puppeteer-core/src/cdp/Browser.ts
+++ b/packages/puppeteer-core/src/cdp/Browser.ts
@@ -36,6 +36,7 @@ import type {Extension} from '../api/Extension.js';
 import type {Page} from '../api/Page.js';
 import type {Target} from '../api/Target.js';
 import type {DownloadBehavior} from '../common/DownloadBehavior.js';
+import {TargetCloseError} from '../common/Errors.js';
 import type {Viewport} from '../common/Viewport.js';
 import {Deferred} from '../util/Deferred.js';
 
@@ -555,6 +556,13 @@ export class CdpBrowser extends BrowserBase {
       timeout,
       message: `Timed out after waiting ${timeout}ms`,
     });
+    const disconnectedDeferred = Deferred.create<Page, TargetCloseError>();
+    const onDisconnected = () => {
+      const error = new TargetCloseError('Browser disconnected');
+      controller.abort(error);
+      disconnectedDeferred.reject(error);
+    };
+    this.on(BrowserEvent.Disconnected, onDisconnected);
 
     try {
       return await Deferred.race([
@@ -598,8 +606,10 @@ export class CdpBrowser extends BrowserBase {
           return page;
         })(),
         timeoutDeferred,
+        disconnectedDeferred,
       ]);
     } finally {
+      this.off(BrowserEvent.Disconnected, onDisconnected);
       controller.abort();
     }
   }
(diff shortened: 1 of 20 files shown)


packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -523,6 +528,103 @@ export class CdpBrowser extends BrowserBase {
...
+    const knownPageTargetIds = new Set(
+      this.targets()
+        .filter(target => {
+          return target.type() === 'page';
+        })
+        .map(target => {
+          return (target as CdpTarget)._targetId;
+        }),
+    );
+    await this.#connection.send('PWA.launch', {
```
OrKoN: Can we use the targetId that this command returns to find the page target?

LuHuangMSFT: `PWA.launch` resolves with the launched **tab** target's id (the CDP `PWA.LaunchResponse` docs describe it as "ID of the tab target created as a result"), not the page. Tab targets aren't exposed through `browser.targets()`, so awaiting the returned id directly never resolves, and the web contents are a separate `page` target.

So we can use it, just as a correlation key rather than a direct lookup. I just changed it to look up that tab target in the target manager and wait for the `page` target that the manager links to it as a child. That drops the origin-based matching entirely and is deterministic for same-origin and multi-window launches. Verified on Chrome for Testing 150 and a Chromium build, headless and headful.

LuHuangMSFT:
@googlebot I signed it!

LuHuangMSFT:
@googlebot I signed it!

OrKoN:
Thanks, we need to update TestExpectations.json file to specify that failures on Firefox, chrome-headless-shell and chrome-bidi are expected.

packages/puppeteer-core/src/cdp/Browser.ts
OrKoN: Thanks, the current version looks good to me.

LuHuangMSFT:
Done in f833bad. Added `[pwa.test] *` skip entries for `firefox`, `webDriverBiDi`, and `chrome` + `chrome-headless-shell`, mirroring the existing `extensions.test` expectations. Verified locally that the suite still runs and passes on chrome-pipe/headless/headful and is skipped on the three unsupported configs.

LuHuangMSFT:
@OrKoN The follow-up fixes are now complete through 1fed900. The install URL contract and launch timeout match runtime behavior, BiDi-only is expected-skipped, and launch cleanup now covers late responses, timeout-disabled waits, browser disconnect, and page creation. I also built the exact PR into a local chrome-devtools-mcp PWA category and passed a real stdio MCP lifecycle: install, inspect OS state, launch/select the returned Page, execute a page script in standalone mode, uninstall, and verify state rejection. The PR is ready for re-review; the new full CI run will need maintainer approval to start.

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -523,6 +529,99 @@ export class CdpBrowser extends BrowserBase {
...
+          // the CDP `PWA.LaunchResponse` docs). Tab targets sit above page
+          // targets in the target hierarchy and are not exposed through
+          // `browser.targets()`, so the returned id can't be awaited directly.
+          const {targetId: tabTargetId} = await this.#connection.send(
+            'PWA.launch',
+            {
+              manifestId: options.manifestId,
+              url: options.url,
+            },
+            {timeout},
```
OrKoN: I think we should let this run with the default timeout.

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -523,6 +529,99 @@ export class CdpBrowser extends BrowserBase {
...
+                return false;
+              }
+              for (const child of tab._childTargets()) {
+                if (child === candidate) {
+                  return true;
+                }
+              }
+              return false;
+            },
+            {timeout: 0, signal: controller.signal},
```
OrKoN: instead we should pass timeout here.

packages/puppeteer-core/src/cdp/Browser.ts
```diff
@@ -523,6 +529,99 @@ export class CdpBrowser extends BrowserBase {
...
+  }
+
+  override async launchPWA(options: LaunchPWAOptions): Promise<Page> {
+    const timeout = options.timeout ?? 30_000;
+    const controller = new AbortController();
+    const timeoutDeferred = Deferred.create<Page>({
+      timeout,
+      message: `Timed out after waiting ${timeout}ms`,
+    });
+    const disconnectedDeferred = Deferred.create<Page, TargetCloseError>();
```
OrKoN: I don't think we need to handle this extra. If the browser is disconnected, the CDP command will error. The waitForTarget should just time out according to the timeout. What was the reason to add handling of the disconnect event?

OrKoN (changes requested):
Thanks! A few comments:

packages/puppeteer-core/src/cdp/Browser.ts
LuHuangMSFT: The intent was to cover the timeout: 0 case. After PWA.launch resolves, there is no pending CDP command, and waitForTarget does not observe browser disconnection, so it could remain pending indefinitely and retain its listeners. However, that adds custom behavior beyond the normal waitForTarget semantics. I agree consistency is preferable here, so I’ll remove the disconnect race and rely on the configured target-wait timeout.

packages/puppeteer-core/src/cdp/Browser.ts
LuHuangMSFT: I’ll remove the per-command timeout so PWA.launch uses the configured protocolTimeout.

packages/puppeteer-core/src/cdp/Browser.ts
LuHuangMSFT: I’ll pass options.timeout directly to waitForTarget, allowing its normal 30-second default when unspecified and 0 to disable the wait timeout.

packages/puppeteer-core/src/cdp/Browser.ts
LuHuangMSFT: Addressed in e9a7e14. `PWA.launch` now uses the configured protocol timeout.

packages/puppeteer-core/src/cdp/Browser.ts
LuHuangMSFT: Addressed in e9a7e14. `options.timeout` is now passed directly to `waitForTarget`.

packages/puppeteer-core/src/cdp/Browser.ts
LuHuangMSFT: Addressed in e9a7e14. I removed the explicit disconnect race, so `launchPWA` now follows the normal `waitForTarget` semantics.

LuHuangMSFT:
@OrKoN Addressed the three `launchPWA` comments in e9a7e14: `PWA.launch` now uses the configured protocol timeout, `options.timeout` is passed to `waitForTarget`, and the explicit disconnect race is removed. I also updated the focused test, public docs, and PR description, then revalidated Edge/Chrome and the MCP lifecycle. Ready for re-review.

test/src/cdp/pwa.test.ts
```diff
@@ -0,0 +1,105 @@
+/**
+ * @license
+ * Copyright 2025 Google Inc.
```
OrKoN:
```suggestion
 * Copyright 2026 Google Inc.
```

OrKoN (approved):
LGTM thanks

PR #15238, opened 2026-07-15
fix: forward headers to browserURL HTTP discovery and WebSocket connection

sibiryakov:

Bug fix. The headers option passed to `puppeteer.connect()` was silently dropped when using browserURL (an `http://` or `https://` URL). The option was destructured from `ConnectOptions` but never forwarded — neither to the HTTP request that discovers the WebSocket endpoint (`/json/version`), nor to the resulting WebSocket connection itself.

Yes. A unit test was added to `packages/puppeteer-core/src/common/BrowserConnector.test.ts` that stubs `globalThis.fetch` and verifies the headers option is forwarded to the `/json/version` discovery request when connecting via `browserURL`.

No documentation changes are needed. The headers option is already documented in `ConnectOptions` as "Headers to use for the web socket connection." This fix makes the `browserURL` path behave consistently with that documented contract.

**Summary**
When connecting via `browserURL`, the `headers` option was silently dropped — it was never passed to the `/json/version` fetch or to the resulting WebSocket connection. This brings the `browserURL` path into parity with the `browserWSEndpoint` and `channel` paths, both of which already forwarded headers correctly.

No. Headers that were previously silently dropped are now forwarded as expected. Existing callers not using headers are unaffected.

**Other information**
The root cause was a missing argument in two call sites inside `getConnectionTransport()` in `packages/puppeteer-core/src/common/BrowserConnector.ts`. The `headers` default was already `{}`, so the only change is that they now reach their destinations.

### DIFF
--- a/packages/puppeteer-core/src/common/BrowserConnector.test.ts
+++ b/packages/puppeteer-core/src/common/BrowserConnector.test.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {describe, it, mock, beforeEach, afterEach} from 'node:test';
+import {describe, it, mock, afterEach} from 'node:test';
 
 import expect from 'expect';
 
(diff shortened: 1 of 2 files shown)


sibiryakov:
@googlebot I signed it!

packages/puppeteer-core/src/common/BrowserConnector.test.ts
```diff
@@ -4,13 +4,49 @@
...
+  describe('getWSEndpoint via browserURL', () => {
+    let originalFetch: typeof globalThis.fetch;
+
+    beforeEach(() => {
+      originalFetch = globalThis.fetch;
+    });
+
+    afterEach(() => {
+      globalThis.fetch = originalFetch;
+      mock.restoreAll();
```
OrKoN, line 16: Thanks for the PR. why is mock needed? can you use it override fetch instead of overriding manually?

OrKoN:
Could you please run `npm run format` locally and fix the formatting issues before uploading?

sibiryakov:
@OrKoN I did. Help me understand the Zizmor problem, please?

OrKoN (approved):

PR #15249, opened 2026-07-20
fix: run hooks even if we try to kill the process

Lightning00Blade:
Issue - #15156

This means that now the Hooks will run correctly and the the folder will be clean after.

### DIFF
--- a/packages/browsers/src/launch.ts
+++ b/packages/browsers/src/launch.ts
@@ -427,6 +427,7 @@ export class Process {
   };
 
   async close(): Promise<void> {
+    await this.#runHooks();
     if (!this.#exited) {
       this.kill();
     }
--- a/packages/browsers/src/launch.ts
+++ b/packages/browsers/src/launch.ts
@@ -416,8 +416,12 @@ export class Process {
   #onDriverProcessSignal = (signal: string): void => {
     switch (signal) {
       case 'SIGINT':
-        this.kill();
-        process.exit(130);
+        this.close()
+          .catch(() => {})
+          .finally(() => {
+            process.exit(130);
+          });
+        break;
       case 'SIGTERM':
       case 'SIGHUP':
         void this.close();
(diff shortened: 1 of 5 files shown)


packages/browsers/src/launch.ts
```diff
@@ -426,7 +427,6 @@ export class Process {
   };
 
   async close(): Promise<void> {
-    await this.#runHooks();
```
OrKoN, line 429: we might still need this?

Lightning00Blade: Yes, reverted

OrKoN (approved):

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -669,6 +669,9 @@ export interface PageEvents extends Record<EventType, unknown> {
  * @public
  */
 export interface NewDocumentScriptEvaluation {
+  /**
```
OrKoN: TODO: revert

packages/browsers/src/launch.ts
```diff
@@ -410,33 +410,41 @@ export class Process {
...
-    return await this.#browserProcessExiting;
+    await this.#browserProcessExiting;
   }
 
   hasClosed(): Promise<void> {
     return this.#browserProcessExiting;
   }
 
+  /**
+   * Kills the browser process. Without running any hooks.
```
OrKoN:
```suggestion
   * Kills the browser process without running any hooks.
```

OrKoN (approved):

OrKoN:
Looks like this failure on Windows could be related https://github.com/puppeteer/puppeteer/actions/runs/30363950347/job/90290701818

OrKoN:
Re-running in https://github.com/puppeteer/puppeteer/actions/runs/30363950347

PR #15251, opened 2026-07-22
fix: preserve browserURL path prefixes

### LINKED ISSUE #15250: getWSEndpoint drops browserURL path prefix when resolving /json/version
## Bug description

`getWSEndpoint()` in `BrowserConnector.ts` uses `new URL("/json/version", browserURL)` to discover the WebSocket endpoint. When `browserURL` includes a path prefix (e.g., behind a reverse proxy or tunnel using URLs like `http://host:port/t/<token>`), the absolute path `/json/version` **replaces** the entire pathname of the base URL, silently dropping the prefix.

## Steps to reproduce

1. Run Chrome DevTools behind a reverse proxy that adds a path prefix, e.g. `http://127.0.0.1:9222/t/<token>`
2. Pass this as `browserURL` to `puppeteer.connect()`:
   ```js
   puppeteer.connect({ browserURL: "http://127.0.0.1:9222/t/abc123" })
   ```
3. HTTP request goes to `http://127.0.0.1:9222/json/version` instead of `http://127.0.0.1:9222/t/abc123/json/version`

## Expected vs Actual

| browserURL | Expected | Actual (broken) |
|---|---|---|
| `http://host:9222/t/token` | `/t/token/json/version` | `/json/version` |
| `http://host:9222` | `/json/version` | `/json/version` |

## Root cause

```ts
const endpointURL = new URL("/json/version", browserURL);
```

Per WHATWG URL Standard, an absolute path (starts with `/`) replaces the base URL pathname entirely.

## Proposed fix

```ts
const endpointURL = new URL(browserURL);
endpointURL.pathname = endpointURL.pathname.replace(/\/?$/, "/") + "json/version";
```

This appends `json/version` to the existing path instead of replacing it.

## Impact

Any scenario where Chrome DevTools is accessed through a reverse proxy, API gateway, or tunnel with path-based routing. Common in cloud IDEs, containerized dev environments, and CDP-over-HTTP tunneling services.

Also affects `chrome-devtools-mcp` which bundles Puppeteer and exposes `--browserUrl`.

cyphercodes:
Fixes #15250

This preserves any path prefix from `browserURL` when Puppeteer fetches `/json/version`, so reverse-proxied DevTools URLs like `/t/session` resolve under that prefix instead of dropping it.

Tests:
- `node --test --test-reporter=spec packages/puppeteer-core/lib/puppeteer/common/BrowserConnector.test.js`
- `npx eslint packages/puppeteer-core/src/common/BrowserConnector.ts packages/puppeteer-core/src/common/BrowserConnector.test.ts`
- `git diff --check`

### DIFF
--- a/packages/puppeteer-core/src/common/BrowserConnector.ts
+++ b/packages/puppeteer-core/src/common/BrowserConnector.ts
@@ -181,8 +181,9 @@ async function getConnectionTransport(
   throw new Error('Invalid connection options');
 }
 
-async function getWSEndpoint(browserURL: string): Promise<string> {
-  const endpointURL = new URL('/json/version', browserURL);
+export async function getWSEndpoint(browserURL: string): Promise<string> {
+  const endpointURL = new URL(browserURL);
+  endpointURL.pathname = `${endpointURL.pathname.replace(/\/$/, '')}/json/version`;
 
   try {
     const result = await globalThis.fetch(endpointURL.toString(), {
--- a/packages/puppeteer-core/src/common/BrowserConnector.test.ts
+++ b/packages/puppeteer-core/src/common/BrowserConnector.test.ts
@@ -4,13 +4,42 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import {describe, it} from 'node:test';
+import {afterEach, describe, it} from 'node:test';
 
 import expect from 'expect';
 
-import {_connectToBrowser} from './BrowserConnector.js';
+import {_connectToBrowser, getWSEndpoint} from './BrowserConnector.js';
 
 describe('BrowserConnector', () => {
+  describe('getWSEndpoint', () => {
+    const originalFetch = globalThis.fetch;
+
+    afterEach(() => {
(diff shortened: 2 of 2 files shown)


packages/puppeteer-core/src/common/BrowserConnector.ts
```diff
@@ -181,8 +181,9 @@ async function getConnectionTransport(
   throw new Error('Invalid connection options');
 }
 
-async function getWSEndpoint(browserURL: string): Promise<string> {
-  const endpointURL = new URL('/json/version', browserURL);
+export async function getWSEndpoint(browserURL: string): Promise<string> {
+  const endpointURL = new URL(browserURL);
+  endpointURL.pathname = `${endpointURL.pathname.replace(/\/$/, '')}/json/version`;
```
OrKoN, line 186: Thanks for the PR! Let's extract this into a separate helper function and test URL rewrites in the unit test without mocks.

OrKoN:
Closing as stale

PR #15274, opened 2026-07-30
fix: do not override the user agent when nothing is emulated

### LINKED ISSUE #15273: connect() strips User-Agent Client Hints from every page, with no setUserAgent() call
# `connect()` strips User-Agent Client Hints from every page, with no `setUserAgent()` call

## Summary

Attaching puppeteer to a browser silently disables User-Agent Client Hints for every
target it touches. `navigator.userAgentData.brands` becomes `[]` and the `Sec-CH-UA`,
`Sec-CH-UA-Mobile` and `Sec-CH-UA-Platform` request headers stop being sent. No
`page.setUserAgent()` / `page.emulate()` call is required — merely connecting is enough.

The User-Agent *string* stays correct, so nothing looks wrong at a glance. But a
`Chrome/150` UA string arriving with zero Client Hints is self-contradictory: every real
Chromium ≥ 89 sends them on secure contexts. Anti-bot layers read the combination as a
spoofed client. In our case PChome (`24h.pchome.com.tw`) started answering `HTTP 429` and
Shopee (`shopee.tw`) redirected to `/verify/traffic/error`.

When connecting to a **shared** browser (`connect({browserURL})`), this also degrades
tabs the human opened, in a browser puppeteer did not launch. That is what made it
hard to attribute: a person browsing normally suddenly gets blocked, with no automation
visible in the tab they are looking at.

## Reproduction

puppeteer-core 25.4.0, Google Chrome 150.0.7871.187, macOS 15.5 (arm64).

Start a browser:

```bash
"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
  --headless=new --remote-debugging-port=9453 \
  --user-data-dir="$(mktemp -d)" --no-first-run about:blank
```

```js
import puppeteer from 'puppeteer-core';

const browser = await puppeteer.connect({ browserURL: 'http://127.0.0.1:9453' });
await page.goto('https://httpbingo.org/headers', { waitUntil: 'networkidle2' });

const brands = await page.evaluate(
  () => JSON.stringify(navigator.userAgentData?.brands ?? null));
const headers = JSON.parse(await page.evaluate(() => document.body.innerText)).headers;

console.log('brands   :', brands);
console.log('Sec-CH-UA:', headers['Sec-Ch-Ua'] ?? '(absent)');
await browser.disconnect();
```

### Actual

```
brands   : []
Sec-CH-UA: (absent)
```

### Expected

The same as driving the identical browser over raw CDP (`Page.navigate` +
`Runtime.evaluate`, no puppeteer), which yields:

```
brands   : [{"brand":"Not;A=Brand","version":"8"},{"brand":"Chromium","version":"150"},
            {"brand":"Google Chrome","version":"150"}]
Sec-CH-UA: "Not;A=Brand";v="8", "Chromium";v="150", "Google Chrome";v="150"
```

## Cause

`EmulationManager#addClient` calls `#applyUserAgent` unconditionally for every client:

```js
async #addClient(client) {
    await Promise.all([
        client.send('Network.enable'),
        ...
        this.#applyUserAgent(client),   // runs even when nothing was overridden
    ]);
}

async #applyUserAgent(client) {
    const userAgent = this.#userAgent ??
        (await this.#frameManager.page().browser().userAgent());   // the browser's real UA
    if (userAgent === undefined) return;
    await client.send('Network.setUserAgentOverride', {
        userAgent,
        acceptLanguage: this.#acceptLanguage,
        userAgentMetadata: this.#userAgentMetadata,   // undefined
        platform: this.#platform,
    });
}
```

With nothing overridden, `#userAgent` is `undefined`, so it falls back to the browser's
own UA and installs an override anyway. `userAgentMetadata` is `undefined`, and
`JSON.stringify` drops undefined values, so the command that reaches the browser is:

```json
{"method":"Network.setUserAgentOverride","params":{"userAgent":"Mozilla/5.0 ..."}}
```

Per [crbug 40270800](https://issuetracker.google.com/issues/40270800), a
`setUserAgentOverride` without `userAgentMetadata` blanks the Client Hints headers from
Chrome 114 onward. Captured with a logging CDP proxy between puppeteer and the browser,
this is the only UA-related frame on the wire, repeated once per `sessionId`.

## Suggested fix

Make `#applyUserAgent` a no-op when nothing was ever overridden:

```js
async #applyUserAgent(client) {
    if (this.#userAgent === undefined &&
        this.#userAgentMetadata === undefined &&
        this.#acceptLanguage === undefined &&
        this.#platform === undefined) {
        return;
    }
    // ... unchanged
}
```

We are running this locally. It restores hints (`brands` length 3, `Sec-CH-UA` present)
while `setUserAgent`/`emulate` with an explicit `userAgent` keeps working — verified with
an iPhone UA still taking effect after the change.

An alternative worth considering is populating `userAgentMetadata` from the real browser
whenever the UA falls back, but skipping the call is the smaller change and matches the
apparent intent (there is nothing to emulate).

## Note for triagers

This is not a stealth / anti-bot-evasion request like #553. The ask is that attaching
puppeteer not change the browser's advertised identity unless asked to — currently it
weakens a fingerprint no caller opted into, and on a shared browser it does so for pages
outside puppeteer's control.

lazerg:
Attaching to a browser sends `Network.setUserAgentOverride` for every client, even when no user agent was set. That override carries no `userAgentMetadata`, so Chrome blanks the User-Agent Client Hints: `navigator.userAgentData.brands` empties and the `Sec-CH-UA` headers stop being sent, which anti-bot layers read as a spoofed client. Skipping the override when nothing was emulated keeps the browser's native hints, while `setUserAgent` and `emulate` still take effect.

Fixes #15273

### DIFF
--- a/packages/puppeteer-core/src/cdp/NetworkManager.ts
+++ b/packages/puppeteer-core/src/cdp/NetworkManager.ts
@@ -291,10 +291,7 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
       this.#userAgentMetadata === undefined &&
       this.#acceptLanguage === undefined &&
       this.#platform === undefined;
-    // Skip only when nothing is emulated and no override has ever been applied.
-    // If an override was applied earlier (e.g. an emulated locale that is now
-    // being cleared), we still need to send the command once to reset the
-    // browser back to its defaults.
+    // Still need to send once to reset a previously-applied override.
     if (nothingToEmulate && !this.#userAgentOverrideApplied) {
       return;
     }
--- a/packages/puppeteer-core/src/cdp/NetworkManager.ts
+++ b/packages/puppeteer-core/src/cdp/NetworkManager.ts
@@ -285,6 +285,14 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
   }
 
   async #applyUserAgent(client: CDPSession) {
+    if (
+      this.#userAgent === undefined &&
+      this.#userAgentMetadata === undefined &&
+      this.#acceptLanguage === undefined &&
+      this.#platform === undefined
+    ) {
+      return;
+    }
     const userAgent =
       this.#userAgent ??
       (await this.#frameManager.page().browser().userAgent());
--- a/packages/puppeteer-core/src/cdp/NetworkManager.ts
+++ b/packages/puppeteer-core/src/cdp/NetworkManager.ts
@@ -85,6 +85,7 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
   #userAgentMetadata?: Protocol.Emulation.UserAgentMetadata;
   #platform?: string;
   #acceptLanguage?: string;
+  #userAgentOverrideApplied = false;
 
   readonly #handlers = [
     ['Fetch.requestPaused', this.#onRequestPaused],
@@ -285,12 +286,16 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
   }
 
   async #applyUserAgent(client: CDPSession) {
-    if (
+    const nothingToEmulate =
       this.#userAgent === undefined &&
       this.#userAgentMetadata === undefined &&
       this.#acceptLanguage === undefined &&
-      this.#platform === undefined
-    ) {
+      this.#platform === undefined;
+    // Skip only when nothing is emulated and no override has ever been applied.
+    // If an override was applied earlier (e.g. an emulated locale that is now
+    // being cleared), we still need to send the command once to reset the
+    // browser back to its defaults.
+    if (nothingToEmulate && !this.#userAgentOverrideApplied) {
       return;
     }
     const userAgent =
@@ -306,6 +311,7 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
         userAgentMetadata: this.#userAgentMetadata,
         platform: this.#platform,
       });
+      this.#userAgentOverrideApplied = !nothingToEmulate;
     } catch (error) {
       if (this.#canIgnoreError(error)) {
         return;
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/cdp/NetworkManager.ts
```diff
@@ -285,6 +285,14 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
   }
 
   async #applyUserAgent(client: CDPSession) {
+    if (
```
OrKoN: Thanks for the PR. There appear to be test failures, PTAL.

lazerg: Of the 8 mocha failures in that job, puppeteer's `TestExpectations` flagged only one as unexpected: `Emulation > Page.emulateLocale > should work`. The other seven are known-flaky entries already tracked in `TestExpectations.json`.

Root cause: the early-return guard I added in `#applyUserAgent` bails out when all four override fields are undefined, and that also killed the reset path. `emulateLocale()` with no argument calls `setAcceptLanguage(undefined)`, which clears `#acceptLanguage`. With every field now undefined, the guard skipped the `Network.setUserAgentOverride` call, so Chrome kept the previous `acceptLanguage` (`fr-FR`) and `navigator.language` never returned to the default.

Fix: I now track whether an override was ever applied (`#userAgentOverrideApplied`) and skip only when nothing is emulated and no override was applied before. When a previously set override is being cleared, we still send one command to reset the browser to its defaults. A fresh `connect()`/`addClient` with nothing emulated still skips the override, so the original #15273 fix stays intact. I also added a NetworkManager unit test for the set-then-clear reset.

OrKoN (approved):

PR #15276, opened 2026-07-30
fix(browser): preserve browserURL path prefixes

### LINKED ISSUE #15275: browserURL drops path prefixes during WebSocket discovery
## Problem

`puppeteer.connect({browserURL})` resolves the discovery endpoint with `new URL("/json/version", browserURL)`. The leading slash replaces the complete pathname, so a browser exposed through a reverse proxy such as `http://127.0.0.1:54704/gateway/session` is queried at `http://127.0.0.1:54704/json/version` instead of the routed endpoint.

This is observable in ChromeDevTools/chrome-devtools-mcp#2394. A project-level workaround was declined in ChromeDevTools/chrome-devtools-mcp#2403 because the behavior belongs in Puppeteer.

## Impact and goals

Users cannot use `browserURL` with path-based CDP gateways and must obtain and configure `browserWSEndpoint` themselves. The goal is to preserve a non-root path prefix while keeping origin-only `browserURL` behavior and request-header forwarding unchanged.

## Current and required behavior

- Current: `http://host/gateway/session` requests `http://host/json/version`.
- Required: `http://host/gateway/session` requests `http://host/gateway/session/json/version`.
- Compatibility: `http://host` continues to request `http://host/json/version`; query strings and fragments are not propagated to the discovery endpoint, matching current behavior.

## Dev design

Parse `browserURL`, normalize its pathname to one trailing slash, append `json/version`, and clear search/hash before fetching. Keep the existing fetch, header forwarding, response handling, and WebSocket connection path unchanged.

## Alternatives

- Fix each consumer separately: duplicates Puppeteer discovery logic and leaves other consumers broken.
- Require `browserWSEndpoint`: works around the bug but removes the convenience and discovery semantics of `browserURL`.

## Scope and non-goals

This only changes `/json/version` URL construction. It does not add proxy configuration, alter WebSocket URLs returned by Chrome, or change connection protocols.

## Test plan and acceptance criteria

- A unit regression captures the discovery request for a prefixed `browserURL`.
- Root URLs still resolve to `/json/version`.
- Existing header-forwarding coverage remains green.
- The puppeteer-core unit suite passes apart from environment-dependent baseline failures reproduced before the production change.

Acceptance is met when the prefixed URL requests `/gateway/session/json/version` without changing the root URL or adding dependencies.

hydraxman:

Bug fix.

Yes. The `BrowserConnector` unit coverage now checks both origin-only and path-prefixed discovery URLs while retaining the existing header-forwarding assertion.

No public API or option changes.

**Summary**

`puppeteer.connect({browserURL})` built `/json/version` with an absolute pathname, which discarded reverse-proxy and gateway path prefixes. This keeps the configured pathname, appends the discovery suffix, and preserves the existing origin-only, query/fragment, header, and WebSocket behavior.

Testing:

- `node --test --test-reporter=spec packages/puppeteer-core/lib/puppeteer/common/BrowserConnector.test.js` — 5 passed
- `npm run unit --workspace puppeteer-core` — 162 passed; one environment-dependent `PuppeteerNode` test requires a locally installed stable Chrome and fails identically on the base revision
- targeted ESLint and Prettier checks — passed
- `git diff --check` — passed

Fixes #15275

No. Origin-only `browserURL` values still request `/json/version`.

**Other information**

The downstream Chrome DevTools MCP workaround was declined in favor of fixing the shared discovery behavior in Puppeteer.

### DIFF
--- a/packages/puppeteer-core/src/common/BrowserConnector.test.ts
+++ b/packages/puppeteer-core/src/common/BrowserConnector.test.ts
@@ -41,12 +41,37 @@ describe('BrowserConnector', () => {
       }).catch(() => {});
 
       expect(capturedRequests).toHaveLength(1);
-      expect(capturedRequests[0]!.url).toContain('/json/version');
+      expect(capturedRequests[0]!.url).toBe(
+        'http://localhost:1234/json/version',
+      );
       expect(
         (capturedRequests[0]!.init?.headers as Record<string, string>)?.[
           'Authorization'
         ],
       ).toBe('Bearer test-token');
+    });
+
+    it('should preserve path prefixes in the /json/version request', async () => {
+      const capturedUrls: string[] = [];
+
+      mock.method(globalThis, 'fetch', async (url: string | URL | Request) => {
+        capturedUrls.push(String(url));
+        return new Response(
+          JSON.stringify({
+            webSocketDebuggerUrl: 'ws://localhost:1234/devtools/browser/1',
+          }),
+          {status: 200, headers: {'Content-Type': 'application/json'}},
+        );
+      });
+
+      await _connectToBrowser({
+        browserURL:
+          'http://localhost:1234/gateway/session?token=secret#fragment',
+      }).catch(() => {});
+
+      expect(capturedUrls).toEqual([
+        'http://localhost:1234/gateway/session/json/version',
+      ]);
     });
   });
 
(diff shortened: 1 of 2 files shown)


OrKoN:
Thanks for the PR. This appears to be a duplicate of https://github.com/puppeteer/puppeteer/pull/15251.

nattallius:
Close as duplicate

PR #15277, opened 2026-07-31
fix(ci): align workflow Node.js version with package engines

MGPOCKY:
## Why
Package metadata declares `engines.node` / `volta.node` / `.nvmrc` as `24`, but CI workflows still pin `node-version` to a mismatched value.

That can make CI run on a runtime the project no longer supports (or skip the version the package actually targets).

## What changed
Use `.nvmrc` as the source of truth for the Node.js version in CI (instead of a hard-coded pin).

- `.github/workflows/ci.yml`
  - `node-version: '22.12'` → `node-version-file: '.nvmrc'`

## Test plan
- [ ] Package `engines.node` / `volta.node` / `.nvmrc` is still `24`
- [ ] Updated workflow now follows `.nvmrc` via `node-version-file`
- [ ] Relevant CI jobs on this branch look healthy

Found while auditing CI/package version consistency across popular repos.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -367,7 +367,7 @@ jobs:
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           cache: npm
-          node-version: '22.12'
+          node-version: '24.0'
       - name: Test install
         env:
           PKG_MANAGER: ${{ matrix.pkg_manager }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -367,7 +367,7 @@ jobs:
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           cache: npm
-          node-version: '24.0'
+          node-version-file: '.nvmrc'
       - name: Test install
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -367,7 +367,7 @@ jobs:
         uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           cache: npm
-          node-version: '22.12'
+          node-version: '24.0'
```
OrKoN: Thanks for the PR but I believe we actually need to use `node-version-file: '.nvmrc'` here.

MGPOCKY:
Thanks for the review! Updated to use `node-version-file: '.nvmrc'` instead of a hard-coded `node-version` pin.

.github/workflows/ci.yml
OrKoN: @Lightning00Blade reminded me that we actually want to test with the minimal supported option here https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer/package.json#L36 so I think 22.12 is as expected here but we might want to add a comment explaining that.

sameerasampath71555-code:
Sema

PR #15286, opened 2026-08-03
fix: use Page.setDocumentContent for CDP setContent

### LINKED ISSUE #12476: [Bug]: setcontent uses document.write triggering chrome console error

```TypeScript

async function screenshot() {
    headless: 'shell',
    ignoreHTTPSErrors: true,
  });
  page.on("console", msg => console.error(msg.text()));
  const html = '<html><head></head><body><script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.7.1/jquery.js"></script></body></html>';
  await page.setContent(html, {waitUntil: 'networkidle0'});
  await browser.close();
}

process.exitCode = 1;
screenshot();
```

### Error string

A parser-blocking, cross site (i.e. different eTLD+1) script, https://cdnjs.cloudflare.com/ajax/libs/jquery/3.7.1/jquery.js, is invoked via document.write. The network request for this script MAY be blocked by the browser in this or a future page load due to poor network connectivity. If blocked in this page load, it will be confirmed in a subsequent console message. See https://www.chromestatus.com/feature/5718547946799104 for more details.

### Bug behavior

- [ ] Flaky
- [ ] PDF

### Background

_No response_

### Expectation

I expected setcontent to use non-deprecated browser features.

### Reality

It errors.

_No response_

### Puppeteer version

22.9.0

### Node version

18.20.2

### Package manager

yarn

### Package manager version

1.22.22

### Operating system

Linux

EduardoFernandess:
## Summary

Fixes #12476.

`page.setContent()` / `Frame.setFrameContent()` used:

```js
document.open();
document.write(html);
document.close();
```

When `html` includes a cross-site `<script src=...>`, Chromium emits:

> A parser-blocking, cross site script … is invoked via document.write …

`setContent` should not depend on a deprecated / warned path for everyday HTML assignment.

### Change

For **CDP** frames, implement `setFrameContent` via [`Page.setDocumentContent`](https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-setDocumentContent). That command still resets frame lifecycle (so `waitUntil` continues to work) without going through `document.write`.

BiDi continues to use the existing evaluate-based `setFrameContent` for now (no equivalent CDP command on that path).

### Test plan

- [ ] Existing `page.setContent` suite (Chrome headless / shell)
- [ ] Repro from #12476: `setContent` with a CDN script and assert no `document.write` console warning

### DIFF
--- a/packages/puppeteer-core/src/cdp/Frame.ts
+++ b/packages/puppeteer-core/src/cdp/Frame.ts
@@ -286,8 +286,9 @@ export class CdpFrame extends Frame {
       timeout = this._frameManager.timeoutSettings.navigationTimeout(),
     } = options;
 
-    // We rely upon the fact that document.open() will reset frame lifecycle with "init"
-    // lifecycle event. @see https://crrev.com/608658
+    // Page.setDocumentContent resets frame lifecycle (same role document.open
+    // played) without using document.write, which triggers Chromium's
+    // parser-blocking cross-site script console error (#12476).
     await this.setFrameContent(html);
 
     const watcher = new LifecycleWatcher(
@@ -304,6 +305,16 @@ export class CdpFrame extends Frame {
     if (error) {
       throw error;
     }
+  }
+
+  /**
+   * @internal
+   */
+  override async setFrameContent(content: string): Promise<void> {
+    await this.#client.send('Page.setDocumentContent', {
+      frameId: this._id,
+      html: content,
+    });
   }
 
   override url(): string {


packages/puppeteer-core/src/cdp/Frame.ts
```diff
@@ -306,6 +307,16 @@ export class CdpFrame extends Frame {
     }
   }
 
+  /**
+   * @internal
+   */
+  override async setFrameContent(content: string): Promise<void> {
+    await this.#client.send('Page.setDocumentContent', {
```
OrKoN, line 314: Thanks for reporting. This is doing pretty much open(), write(), close() as well https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/document.cc;l=1900;drc=08995bd34b815a31ed4a5fbef9d7fa921c9f0053 Does it solve the original issue with the warning being reported?

PR #15287, opened 2026-08-03
fix: serialize concurrent browser installs with a lock

### LINKED ISSUE #14417: [Bug]: Possible cache corruption when two Chrome with the same buildId installs run at once

EduardoFernandess:
## Summary

Fixes #14417.

`@puppeteer/browsers` install was not atomic. Two processes installing the same `browser`/`buildId` into the shared cache could interleave download, extract, and archive `unlink`, producing:

- a browser folder without an executable
- a missing zip mid-extract
- other corrupt cache states

Maintainers noted they would review a PR that implements locking in a cross-platform way **without additional dependencies**.

### Change

1. Add `withInstallLock()` using exclusive `mkdir` as the lock primitive (no new deps).
2. Record a PID file for stale-lock recovery (`process.kill(pid, 0)`).
3. Wrap the unpack/install critical section for a given `outputPath` so concurrent installers of that build serialize; the second waiter reuses the completed install when it acquires the lock.

### Test plan

- [x] Unit tests for serialization and stale-lock recovery (`packages/browsers/test/src/installLock.test.ts`)
- [ ] Optional stress: two parallel `install.js` runs against a cleared cache for the same Chrome buildId

### DIFF
--- a/packages/browsers/src/install.ts
+++ b/packages/browsers/src/install.ts
@@ -22,6 +22,7 @@ import {DefaultProvider} from './DefaultProvider.js';
 import {detectBrowserPlatform} from './detectPlatform.js';
 import {unpackArchive} from './fileUtil.js';
 import {downloadFile, headHttpRequest} from './httpUtil.js';
+import {withInstallLock} from './installLock.js';
 import {ProgressBar} from './ProgressBar.js';
 import type {BrowserProvider} from './provider.js';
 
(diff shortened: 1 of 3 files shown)


OrKoN:
Thanks for the PR but there is already a PR for this issue that is being reviewed https://github.com/puppeteer/puppeteer/pull/15218

PR #15288, opened 2026-08-04
revert: run hooks even if we try to kill the process

OrKoN:
Reverts puppeteer/puppeteer#15249

To unblock the release.


Lightning00Blade (approved):
Should this be only `revert: run hooks even if we try to kill the process`
And not a `fix`?

PR #15292, opened 2026-08-05
feat: support shadow roots in MutationPoller

### LINKED ISSUE #13163: Support ShadowRoots in MutationPoller
Currently, MutationPoller only creates a mutation observer for the document and the observer does not track mutations inside shadow roots.

We should implement tracking of existing and new shadow roots in Mutation poller to detect mutations inside shadow roots by creating an observer instance for each shadow root.

Related https://github.com/puppeteer/puppeteer/pull/13153

Develop-KIM:

A feature: `MutationPoller` now observes shadow roots. Fixes #13163.

Yes, three new `Frame.waitForSelector` tests, and the shadow DOM test that was skipped for #13163 now runs.

No public API changed.

**Summary**

`MutationPoller` observed one node with `subtree: true`. A subtree registration does not cross shadow boundaries, so nothing that happened inside a shadow root was ever reported and `waitForSelector('div >>> h1')` just sat there until it timed out. The same applies to every other mutation-polled handler (`pierce/`, `text/`, `xpath/`, plain CSS).

The poller now keeps the single observer and registers every reachable open shadow root as an extra target:

- at `start()` the root is walked once for shadow roots, recursively, so nested roots are covered;
- in the observer callback each `addedNodes` entry is walked the same way before `fn()` runs, so a subtree that arrives carrying shadow roots gets observed too;
- a `WeakSet` keeps the walks idempotent, and `stop()` still tears everything down with a single `disconnect()`.

**One case this does not cover**

Attaching a shadow root to a node that is *already* in the DOM produces no mutation record anywhere. I checked this directly: with an observer on `document` using `childList`/`subtree`/`attributes`, `host.attachShadow(...)` yields zero records. That is the gap https://github.com/whatwg/dom/issues/1287 is about. So if a page attaches a shadow root to an existing node and then only mutates inside it, the wait still times out.

That was the exact scenario in the test added by #13153, so I kept the test as `should work when a shadow root is attached to an existing node` and left it skipped in `TestExpectations.json`, with the comment pointing at the spec issue rather than #13163. The three scenarios that are now supported are:

- a host whose shadow root exists before the host is inserted (this is the shape in #13152);
- a shadow root that already exists when the wait starts;
- a nested shadow root.

The usual workaround for the unobservable case is monkey-patching `Element.prototype.attachShadow`, and that is not available here: `PuppeteerUtil` runs in an isolated world, whose `Element.prototype` is a different object from the page's. I confirmed a patch installed from the isolated world sees zero calls made from the main world. Happy to leave #13163 open for that part if you'd like.

**Cost**

One `TreeWalker` pass over the root at poller start, plus one pass per added subtree afterwards. On a synthetic 56k-element document the start-up pass measures ~1.3ms (median of 20, headless Chrome 151, M-series Mac). Selectors that cannot pierce shadow roots pay it without benefit; if that bothers you I can plumb a flag from `getQueryHandlerAndSelector` so only the piercing handlers opt in.

No.

**Other information**

Verified on `chrome-headless` (full suite, only two pre-existing environment failures unrelated to this change: `emulateTimezone` on a non-English locale and the download-location timing test), plus the new tests on `chrome-bidi` and `firefox` BiDi.

### DIFF
--- a/packages/puppeteer-core/src/injected/Poller.ts
+++ b/packages/puppeteer-core/src/injected/Poller.ts
@@ -38,10 +38,13 @@ function parentOf(node: Node): Node | null {
 }
 
 function hasAncestorIn(node: Node, nodes: Set<Node>): boolean {
-  for (let parent = parentOf(node); parent; parent = parentOf(parent)) {
+  let current = node;
+  let parent: Node | null;
+  while ((parent = parentOf(current))) {
     if (nodes.has(parent)) {
       return true;
     }
+    current = parent;
   }
   return false;
 }
--- a/packages/puppeteer-core/src/injected/Poller.ts
+++ b/packages/puppeteer-core/src/injected/Poller.ts
@@ -22,6 +22,30 @@ const MUTATION_OBSERVER_OPTIONS: MutationObserverInit = {
   attributes: true,
 };
 
+function canHostShadowRoots(node: Node): boolean {
+  return (
+    node.nodeType === Node.ELEMENT_NODE ||
+    node.nodeType === Node.DOCUMENT_FRAGMENT_NODE
+  );
+}
+
+/**
+ * A shadow root has no parent, so the walk up continues through its host to
+ * stay inside the tree the node was added to.
+ */
+function parentOf(node: Node): Node | null {
+  return node.parentNode ?? (node as ShadowRoot).host ?? null;
+}
+
+function hasAncestorIn(node: Node, nodes: Set<Node>): boolean {
+  for (let parent = parentOf(node); parent; parent = parentOf(parent)) {
+    if (nodes.has(parent)) {
+      return true;
+    }
+  }
+  return false;
+}
+
 /**
  * @internal
  */
@@ -48,11 +72,7 @@ export class MutationPoller<T> implements Poller<T> {
 
     this.#observedRoots = new WeakSet();
     this.#observer = new MutationObserver(async mutations => {
-      for (const mutation of mutations) {
-        for (const node of mutation.addedNodes) {
-          this.#observeShadowRoots(node);
-        }
-      }
+      this.#observeAddedShadowRoots(mutations);
       const result = await this.#fn();
       if (!result) {
         return;
@@ -74,6 +94,27 @@ export class MutationPoller<T> implements Poller<T> {
     this.#observedRoots.add(root);
     this.#observer.observe(root, MUTATION_OBSERVER_OPTIONS);
     this.#observeShadowRoots(root);
+  }
+
+  /**
+   * A batch can report a subtree and nodes inside it, so the added nodes are
+   * pruned to the top-most ones and each tree is walked once.
+   */
+  #observeAddedShadowRoots(mutations: MutationRecord[]): void {
+    const addedNodes = new Set<Node>();
+    for (const mutation of mutations) {
+      for (const node of mutation.addedNodes) {
+        if (!canHostShadowRoots(node)) {
+          continue;
+        }
+        addedNodes.add(node);
+      }
+    }
+    for (const node of addedNodes) {
+      if (!hasAncestorIn(node, addedNodes)) {
+        this.#observeShadowRoots(node);
+      }
+    }
   }
 
   /**
--- a/packages/puppeteer-core/src/injected/Poller.ts
+++ b/packages/puppeteer-core/src/injected/Poller.ts
@@ -16,6 +16,12 @@ export interface Poller<T> {
   result(): Promise<T>;
 }
 
+const MUTATION_OBSERVER_OPTIONS: MutationObserverInit = {
+  childList: true,
+  subtree: true,
+  attributes: true,
+};
+
 /**
  * @internal
  */
@@ -25,6 +31,7 @@ export class MutationPoller<T> implements Poller<T> {
   #root: Node;
 
   #observer?: MutationObserver;
+  #observedRoots = new WeakSet<Node>();
   #deferred?: Deferred<T>;
   constructor(fn: () => Promise<T>, root: Node) {
     this.#fn = fn;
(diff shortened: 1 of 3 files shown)


nattallius:
Hi, overall this looks good to me, I would proceed to merge it. @OrKoN what do you think?

packages/puppeteer-core/src/injected/Poller.ts
```diff
@@ -39,19 +46,48 @@ export class MutationPoller<T> implements Poller<T> {
       return;
     }
 
-    this.#observer = new MutationObserver(async () => {
+    this.#observedRoots = new WeakSet();
+    this.#observer = new MutationObserver(async mutations => {
+      for (const mutation of mutations) {
+        for (const node of mutation.addedNodes) {
+          this.#observeShadowRoots(node);
```
OrKoN: I think we should only call this for `node.nodeType === Node.ELEMENT_NODE || node.nodeType === Node.DOCUMENT_FRAGMENT_NODE` to avoid creating tree walkers for non-element nodes.

packages/puppeteer-core/src/injected/Poller.ts
```diff
@@ -39,19 +46,48 @@ export class MutationPoller<T> implements Poller<T> {
       return;
     }
 
-    this.#observer = new MutationObserver(async () => {
+    this.#observedRoots = new WeakSet();
+    this.#observer = new MutationObserver(async mutations => {
+      for (const mutation of mutations) {
```
OrKoN: Should we this before we run this.#fn() or after, or perhaps if the fn did not return a truthy value?

OrKoN: should we optimize this a bit by pruning the the added nodes across mutations ahead of time?

```
    // 1. Collect all newly added elements into a  set.
    const addedElements = new Set<Node>();
    for (const mutation of mutations) {
      for (const node of mutation.addedNodes) {
        if (
          node.nodeType === Node.ELEMENT_NODE ||
          node.nodeType === Node.DOCUMENT_FRAGMENT_NODE
        ) {
          addedElements.add(node);
        }
      }
    }

    // 2. Prune any element whose ancestor is also in addedElements
    const topMostRoots = new Set<Node>();
    for (const node of addedElements) {
      let isDescendant = false;
      let parent = node.parentNode ?? (node as ShadowRoot).host;
      while (parent) {
        if (addedElements.has(parent)) {
          isDescendant = true;
          break;
        }
        parent = parent.parentNode ?? (parent as ShadowRoot).host;
      }
      if (!isDescendant) {
        topMostRoots.add(node);
      }
    }

    // 3. Walk only the top-most roots
    for (const root of topMostRoots) {
      this.#observeShadowRoots(root);
    }
```

packages/puppeteer-core/src/injected/Poller.ts
Develop-KIM: Done in 899a61b — `canHostShadowRoots` now gates it, so only element and document-fragment nodes get a walker. Text nodes were the bulk of what was landing here.

packages/puppeteer-core/src/injected/Poller.ts
Develop-KIM: I kept it before `this.#fn()`, and I think it has to stay there.

If the walk runs after `fn` (or only when `fn` came back falsy), the gap between the mutation and the walk contains an `await`. Anything that changes inside a shadow root that arrived in that batch during the gap produces no record on any observer we hold, and we don't re-run `fn` after registering, so the poller just sits there until some unrelated mutation elsewhere wakes it up. Registering first closes that window.

What we pay for it is a walk we throw away when `fn` resolves on that same batch. That's now bounded to the added subtrees rather than the document, and with the pruning below it's a single pass per added tree. Happy to flip it if you'd rather take the cheaper path.

Develop-KIM: Done in 899a61b, close to your sketch. `#observeAddedShadowRoots` collects the element/fragment nodes from the whole batch into a set, drops the ones that already have an ancestor in that set, and walks what's left.

One detail from your version I kept and pulled out into `parentOf`: the walk up uses `node.parentNode ?? node.host`, so a node added inside a shadow root whose host also arrived in the same batch still resolves to that host and gets pruned. `#observeShadowRoots` recurses into shadow roots as it walks, so the surviving top-most node covers it.

Full Chrome headless run is green (1310 passing, only the pre-existing `emulateTimezone` failure on my machine, which is a system-locale thing unrelated to this).

OrKoN:
@Lightning00Blade could you please review as well?

packages/puppeteer-core/src/injected/Poller.ts
```diff
@@ -16,6 +16,36 @@ export interface Poller<T> {
...
+}
+
+function hasAncestorIn(node: Node, nodes: Set<Node>): boolean {
+  for (let parent = parentOf(node); parent; parent = parentOf(parent)) {
+    if (nodes.has(parent)) {
+      return true;
+    }
+  }
+  return false;
+}
```
Lightning00Blade, line 50:
```suggestion
function hasAncestorIn(node: Node, nodes: Set<Node>): boolean {
  let parent: Node | null | undefined;
  while ((parent = parentOf(current))) {
    if (nodes.has(parent)) {
      return true;
    }
  }
  return false;
}
```

Lightning00Blade (approved):
LGTM, with a small nit

packages/puppeteer-core/src/injected/Poller.ts
Develop-KIM: Switched to the while form in ede92bc. I had to add two small things to your snippet to make it walk: `current` wasn't declared, and it needs `current = parent` at the end of each pass, otherwise `parentOf(current)` keeps returning the same node.

Thanks for the review!

Lightning00Blade: Sorry my bad. What was in my head was 
```ts
function hasAncestorIn(node: Node, nodes: Set<Node>): boolean {
  let parent: Node | null | undefined = node;
  while ((parent = parentOf(parent))) {
    if (nodes.has(parent)) {
      return true;
    }
  }
  return false;
}
```
I think the current version is also OK.

Lightning00Blade (approved):

OrKoN (approved):

PR #15293, opened 2026-08-05
ci: add min-release-age

OrKoN:


.npmrc
```diff
@@ -1,2 +1,3 @@
 access=public
 install-strategy=nested
+min-release-age=7 # requires npm 11.10.0+
```
nattallius, line 3: What about `devtools-protocol` in `packages/puppeteer/package.json`, should we add 
`min-release-age-exclude[]=devtools-protocol` or we are ok for 1 week delay gap?

OrKoN: Added `min-release-age-exclude[]=devtools-protocol`

nattallius (approved):

PR #15300, opened 2026-08-06
fix(cdp): do not fail a per-frame fan-out when an OOP iframe goes away

### LINKED ISSUE #15299: [Bug]: exposeFunction rejects with TargetCloseError when an out-of-process iframe goes away mid call

```typescript

await page.goto('https://example.com/');

// Cross-site frames, so each one gets a CDP session of its own.
await page.evaluate(() => {
  for (let i = 0; i < 8; i++) {
    const frame = document.createElement('iframe');
    frame.src = 'https://example.org/';
    document.body.append(frame);
  }
});
await new Promise(resolve => {
  return setTimeout(resolve, 2000);
});

// Drop them while exposeFunction is fanning out over the frames.
void page.evaluate(() => {
  for (const frame of document.querySelectorAll('iframe')) {
    frame.remove();
  }
});

try {
  await page.exposeFunction('doubleIt', v => v * 2);
  console.log('exposeFunction resolved');
} catch (error) {
  console.log('exposeFunction threw:', error.message);
}

console.log(
  'window.doubleIt on the main frame:',
  await page.evaluate(() => typeof globalThis.doubleIt),
);
console.log('page still usable:', await page.evaluate(() => 40 + 2));

await browser.close();
```

Output, same on three consecutive runs:

```
exposeFunction threw: Protocol error (Runtime.addBinding): Target closed
window.doubleIt on the main frame: function
page still usable: 42
```

### Background

I automate a site that keeps a cross-site iframe around and rebuilds it during startup, and I call `exposeFunction` a number of times in a row right after the app syncs. About 31% of those startups aborted on this across a fleet of 20 machines over two weeks, and the effect is that everything I wire up after the failing call is silently missing for that session.

Took me a while to find because the page is completely healthy afterwards, and the binding I asked for is actually there.

### Expectation

`exposeFunction` should not fail because a frame I did not ask about went away. If the frame no longer exists there is nothing left to install into.

### Reality

`exposeFunction` fans out over every frame and sends on that frame's own client:

- `FrameManager.addExposedFunctionBinding` calls `CdpFrame.addExposedFunctionBinding`, which sends `Runtime.addBinding` over [`this.#client`](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/cdp/Frame.ts#L351)
- `FrameManager.evaluateOnNewDocument` calls `CdpFrame.addPreloadScript`, which sends `Page.addScriptToEvaluateOnNewDocument` over the same client

An out-of-process iframe owns that client. When the frame goes away Chromium emits `Target.detachedFromTarget`, `Connection.onMessage` closes the session, and `CallbackRegistry.clear` rejects whatever was still pending, so the whole page level call rejects.

The reported failure is misleading. As the output above shows, `window.doubleIt` is already installed and working on the main frame, and the page, the main frame and the browser are all fine.

Bisect:

- `22.12.1`: resolves, 3 of 3 runs
- `22.13.0`: throws, 3 of 3 runs

`22.13.0` is the first release with b6b536bb from #12722, which moved both commands off `#primaryTargetClient` onto per frame sessions. Before that they went to the page session, which cannot die this way, and the only per frame work was `frame.evaluate(expression).catch(debugError)` with errors swallowed on purpose. The three tests that PR added are all happy path, so nothing covers a frame disappearing mid call.

`FrameManager.initialize` already treats this as expected when it seeds a newly attached frame, with the comment [`The target might have been closed before the initialization finished`](https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/src/cdp/FrameManager.ts#L258), so the fan outs reached from `Page` look inconsistent with it.

Happy to send a PR if you want one.

### Puppeteer version

25.5.0

### Node version

22.20.0

### Package manager

npm

### Package manager version

11.6.2

### Operating system

Windows

Adi1231234:
Fixes #15299

## Summary

`page.exposeFunction()` and `page.evaluateOnNewDocument()` fan out over every frame and send on that frame's own CDP session. If an out-of-process iframe is torn down while those calls are in flight, its session is detached and the whole page level call rejects, even though the page and the main frame are fine.

## Problem

`FrameManager` sends `Runtime.addBinding` and `Page.addScriptToEvaluateOnNewDocument` per frame over `frame._client()`, which an OOPIF owns. When the frame goes away, `Connection.onMessage` closes that session and `CallbackRegistry.clear` rejects whatever was pending:

```
TargetCloseError: Protocol error (Runtime.addBinding): Target closed
```

The rejection is misleading. In every failure I looked at, `window.<name>` was already installed on the main frame and calling it from the page reached the Node callback.

This came in with b6b536bb (#12722), which moved both commands off `#primaryTargetClient` onto per frame sessions. With the repro from #15299, `22.12.1` resolves in 3 of 3 runs and `22.13.0` throws in 3 of 3. The three tests that PR added are all happy path, so nothing covered a frame disappearing mid call.

## Fix

The three fan outs in `FrameManager` were the same five lines three times. They now go through one `#forEachFrame` helper that tolerates a frame losing a session that was its own, so each call site is shorter than before:

```diff
-    await Promise.all(
-      this.frames().map(async frame => {
-        return await frame.addExposedFunctionBinding(binding);
-      }),
-    );
+    await this.#forEachFrame(frame => {
+      return frame.addExposedFunctionBinding(binding);
+    });
```

A frame sharing the manager's client is not excused, because there `TargetCloseError` means the page itself is gone and the caller has to hear about it. `#initialize` already swallows the same error for the same reason when it seeds a newly attached frame, and `Input#withTransaction` is the same helper shape. `CdpFrame` is untouched.

## Test

`should exposeFunction when an OOP iframe goes away mid-call` attaches 8 cross-process frames, starts tearing them down, then calls `exposeFunction` while that is in flight. It fails on `main` in 8 of 8 runs and passes in 8 of 8 with the fix.

On the pinned Chrome 151 the OOPIF suite stays at 28 passing, and `evaluation`, `page`, `waittask`, `elementhandle` and `ariaqueryhandler` keep the same 7 pre-existing failures. Killing the browser mid call still rejects with and without subframes, so the helper is not hiding real page loss.

### DIFF
--- a/packages/puppeteer-core/src/cdp/FrameManager.ts
+++ b/packages/puppeteer-core/src/cdp/FrameManager.ts
@@ -282,22 +282,39 @@ export class FrameManager extends EventEmitter<FrameManagerEvents> {
     return this._frameTree.getById(frameId) || null;
   }
 
-  async addExposedFunctionBinding(binding: Binding): Promise<void> {
-    this.#bindings.add(binding);
+  async #forEachFrame(
+    action: (frame: CdpFrame) => Promise<unknown>,
+  ): Promise<void> {
     await Promise.all(
       this.frames().map(async frame => {
-        return await frame.addExposedFunctionBinding(binding);
+        try {
+          await action(frame);
+        } catch (error) {
+          // Only an out-of-process frame has a session of its own to lose.
+          if (
+            frame._client() === this.#client ||
+            !isErrorLike(error) ||
+            !isTargetClosedError(error)
+          ) {
+            throw error;
+          }
+        }
       }),
     );
   }
 
+  async addExposedFunctionBinding(binding: Binding): Promise<void> {
+    this.#bindings.add(binding);
+    await this.#forEachFrame(frame => {
+      return frame.addExposedFunctionBinding(binding);
+    });
+  }
+
   async removeExposedFunctionBinding(binding: Binding): Promise<void> {
     this.#bindings.delete(binding);
-    await Promise.all(
-      this.frames().map(async frame => {
-        return await frame.removeExposedFunctionBinding(binding);
-      }),
-    );
+    await this.#forEachFrame(frame => {
+      return frame.removeExposedFunctionBinding(binding);
+    });
   }
 
   async evaluateOnNewDocument(
@@ -317,11 +334,9 @@ export class FrameManager extends EventEmitter<FrameManagerEvents> {
 
     this.#scriptsToEvaluateOnNewDocument.set(identifier, preloadScript);
 
-    await Promise.all(
-      this.frames().map(async frame => {
-        return await frame.addPreloadScript(preloadScript);
-      }),
-    );
+    await this.#forEachFrame(frame => {
+      return frame.addPreloadScript(preloadScript);
+    });
 
     return {identifier};
   }
--- a/test/src/oopif.test.ts
+++ b/test/src/oopif.test.ts
@@ -433,6 +433,39 @@ describe('OOPIF', function () {
     });
   });
 
+  it('should exposeFunction when an OOP iframe goes away mid-call', async () => {
+    const {page, server} = state;
+    const frameCount = 8;
+
+    await page.goto(server.EMPTY_PAGE);
+    for (let i = 0; i < frameCount; i++) {
+      await attachFrame(
+        page,
+        `oopif${i}`,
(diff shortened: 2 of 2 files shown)


OrKoN (approved):
LGTM thanks!

PR #15303, opened 2026-08-06
ci: add a cooldown for deps

nattallius:


OrKoN (commented):
Thanks for the PR! I think we need to add it to all package ecosystems to pass the checks.

.github/dependabot.yml
```diff
@@ -7,6 +7,15 @@ updates:
...
       time: '02:00'
       timezone: Europe/Berlin
+    cooldown:
+      default-days: 7
+      exclude:
+        - devtools-protocol
+        - chromium-bidi
+        - '@angular-devkit/*'
+        - '@angular/*'
+        - '@schematics/angular'
```
OrKoN: let's remove these. We do not need to update them so fast.

OrKoN (approved):

PR #15314, opened 2026-08-07
chore: refactor logger part 3

Lightning00Blade:


docs/api/puppeteer.browsercontext.md
```diff
@@ -93,6 +93,21 @@ string \| undefined
 
 Identifier for this [browser context](./puppeteer.browsercontext.md).
 
+</td></tr>
```
OrKoN, line 96: do we need to make it non-internal?

Lightning00Blade: Hmm interesting, I though it would given it's protected. 
Maybe we should hide all protected properties from the API docs?

Lightning00Blade: I can do that in a separate PR.

OrKoN: Let's just mark it as '@internal'. I think there could be use cases for documented protected props.

OrKoN (approved):

rogierslag:
This introduced production crashes for us after upgrading from Puppeteer 24.41.0 to 25.6.0.

`HTTPRequest.frame()` is documented as returning `Frame | null`. However, the new error paths in `CdpHTTPRequest._continue`, `_respond`, and `_abort` all access `(this.frame() as any).logger` without handling that nullable contract.

We hit this when a frame detached while an intercepted request was being finalized. `Fetch.continueRequest` rejected, after which its catch handler threw a second error:

```
TypeError: Cannot read properties of null (reading 'logger')
    at .../puppeteer-core/lib/puppeteer/cdp/HTTPRequest.js
    at async CdpHTTPRequest._continue
    at async CdpHTTPRequest.continue
    at async CdpHTTPRequest.finalizeInterceptions
```

That masks the original protocol error and escapes as an unhandled rejection. The same service and interception code were stable on 24.41.0, where these paths called `handleError(error)` without dereferencing the frame.

Current `main` still contains the nullable dereference. Could these paths use the request logger introduced by the surrounding logger refactor, or otherwise handle a null frame, and add coverage for a frame disappearing before the protocol operation rejects? The issue applies equally to `continue`, `respond`, and `abort`.

PR #15319, opened 2026-08-08
fix: tell the user how to recover a partial browser folder

### LINKED ISSUE #15318: [Bug]: A partial browser folder is never cleaned up or re-downloaded, so an interrupted install stays broken forever
### Steps to reproduce

Any interruption during `unpackArchive()` leaves `outputPath` behind, and every later install of that build then fails permanently. This reproduces it deterministically without touching the real cache:

```bash
export PUPPETEER_CACHE_DIR=/tmp/pptr-sticky-cache
rm -rf /tmp/pptr-sticky /tmp/pptr-sticky-cache
mkdir -p /tmp/pptr-sticky && cd /tmp/pptr-sticky
npm init -y >/dev/null

# 1. Install once, normally. This succeeds.
npm i puppeteer@25.5.0 --foreground-scripts

# 2. Simulate an extraction that was interrupted:
#    the build folder exists, but nothing was written into it.
BUILD=$(ls "$PUPPETEER_CACHE_DIR/chrome")
rm -rf "$PUPPETEER_CACHE_DIR/chrome/$BUILD"/*

# 3. Re-install. This fails, and keeps failing on every retry.
rm -rf node_modules package-lock.json
npm i puppeteer@25.5.0 --foreground-scripts   # exit 1
npm i puppeteer@25.5.0 --foreground-scripts   # exit 1 again, and forever
```

How it happened in practice: an install was interrupted while extracting Chrome, leaving an empty `chrome/mac_arm-151.0.7922.71/` directory. From then on, every `npx` launch of a package that depends on puppeteer aborted at puppeteer's postinstall with exit code 1. Deleting that one directory by hand was the only way out.

Environment: `puppeteer@25.5.0`, `@puppeteer/browsers@3.1.0` (both current latest), Node v22.22.2, npm 10.9.7, macOS arm64.

### Expected results

`install()` recovers by itself. A browser folder with no executable is not a usable installation, so it should be treated as absent and re-downloaded, or removed when the install that created it does not finish. Recovery should not require the user to locate and delete a cache directory manually.

### Actual results

`install()` takes the existence of the directory as proof of a completed installation and throws:

```
ERROR: Failed to set up chrome v<buildId>! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
  cause: Error: The browser folder (<cacheDir>/chrome/mac_arm-<buildId>) exists but the executable
  (<cacheDir>/chrome/mac_arm-<buildId>/chrome-mac-arm64/Google Chrome for Testing.app/Contents/MacOS/Google Chrome for Testing)
  is missing
```

Two things in `packages/browsers/src/install.ts` combine to make this permanent.

First, the early branch throws instead of falling through to a re-download (L450):

```ts
if (existsSync(outputPath)) {
  if (!existsSync(installedBrowser.executablePath)) {
    throw new Error(
      `The browser folder (${outputPath}) exists but the executable (${installedBrowser.executablePath}) is missing`,
    );
  }
```

Second, the cleanup on the way out only removes the archive, never the half-written output directory (L504):

```ts
} finally {
  if (existsSync(archivePath)) {
    await unlink(archivePath);
  }
}
```

So whatever interrupts `unpackArchive()` — Ctrl+C, the machine suspending, the process being killed, a full disk, a crash — leaves behind a directory that poisons the cache for good.

This is easy to miss from the outside: puppeteer's postinstall ends in `process.exit(1)`, and npm discards postinstall output unless `--foreground-scripts` is set, so the user sees an install abort with no message at all.

### Additional context

This does not look like a duplicate of the existing reports, which are all about *causes* of a partial folder rather than about the state being unrecoverable:

- #14957 / #15119 / #15071 — `extract-zip@2.0.1` silently truncating extraction on Node 24.16+ / 26. Already resolved by dropping `extract-zip` (3.1.0 uses `modern-tar`), and not applicable here: Node 22 with `@puppeteer/browsers@3.1.0`.
- #14417 — cache corruption when two installs of the same buildId run concurrently.

The open PR for #14417 (#15218) does not cover this either. In its diff the `existsSync(outputPath)` check and its throw are preserved verbatim and only moved inside the new lock, and the cleanup still touches `archivePath` alone. Serializing installers removes one way of producing a partial folder, but not the rest, and does nothing for folders that already exist.

A fix looks orthogonal to the locking work and small: either treat a folder with a missing executable as not installed and re-install into it, or remove `outputPath` when the install that created it fails.

lazerg:
An install that gets interrupted while extracting leaves behind a browser folder with no executable in it. `install()` reads that folder as a finished installation and throws, and the message says what is wrong but not what to do about it.

`@puppeteer/browsers` now names the folder to delete, which holds for anyone using the package, and throws an `IncompleteInstallationError` for that case. The CLI catches that error and adds the clear command built from its own `scriptName` and `prefixCommand`, so `npx puppeteer browsers install` and `npx @puppeteer/browsers install` each name the command they actually have.

The install script also stopped calling `process.exit(1)` on a failed download. `downloadBrowsers()` throws instead, `install.mjs` awaits it, and the existing catch reports it. The postinstall exit code goes from 1 to 0 on a download failure, so `npm install` no longer fails with nothing on screen and the missing browser is reported at `launch()`.

Fixes #15318

### DIFF
--- a/packages/browsers/src/install.ts
+++ b/packages/browsers/src/install.ts
@@ -158,6 +158,14 @@ export interface InstallOptions {
 }
 
 /**
+ * Thrown when an installation folder is present but does not hold a browser
+ * that can be launched.
+ *
+ * @internal
+ */
+export class IncompleteInstallationError extends Error {}
+
+/**
  * Install using custom provider plugins.
  * Tries each provider in order until one succeeds.
  * Falls back to default provider if all custom providers fail.
@@ -270,9 +278,16 @@ async function installWithProviders(
       return `  - ${e.providerName}: ${e.error.message}`;
     })
     .join('\n');
-  throw new Error(
-    `All providers failed for ${options.browser} ${options.buildId}:\n${errorDetails}`,
-  );
+  const message = `All providers failed for ${options.browser} ${options.buildId}:\n${errorDetails}`;
+  if (
+    errors.length > 0 &&
+    errors.every(e => {
+      return e.error instanceof IncompleteInstallationError;
+    })
+  ) {
+    throw new IncompleteInstallationError(message);
+  }
+  throw new Error(message);
 }
 
 /**
@@ -449,7 +464,7 @@ async function installUrl(
   try {
     if (existsSync(outputPath)) {
       if (!existsSync(installedBrowser.executablePath)) {
-        throw new Error(
+        throw new IncompleteInstallationError(
           `The browser folder (${outputPath}) exists but the executable (${installedBrowser.executablePath}) is missing. ` +
             `An earlier install of this build probably did not finish. ` +
             `Delete ${outputPath} and install the browser again.`,
--- a/packages/browsers/src/install.ts
+++ b/packages/browsers/src/install.ts
@@ -7,7 +7,7 @@
 import assert from 'node:assert';
 import {spawnSync} from 'node:child_process';
 import {existsSync, readFileSync} from 'node:fs';
-import {mkdir, rm, unlink} from 'node:fs/promises';
+import {mkdir, unlink} from 'node:fs/promises';
 import os from 'node:os';
 import path from 'node:path';
 
@@ -448,22 +448,18 @@ async function installUrl(
 
   try {
     if (existsSync(outputPath)) {
-      if (existsSync(installedBrowser.executablePath)) {
-        await runSetup(installedBrowser, logger);
-        if (options.installDeps) {
-          await installDeps(installedBrowser, logger);
-        }
-        return installedBrowser;
+      if (!existsSync(installedBrowser.executablePath)) {
+        throw new Error(
+          `The browser folder (${outputPath}) exists but the executable (${installedBrowser.executablePath}) is missing. ` +
+            `An earlier install of this build probably did not finish. ` +
+            `Run \`npx puppeteer browsers clear\` to empty the browser cache, then install again.`,
+        );
       }
-      logger?.(DEBUG_PREFIXES.install)?.(
-        `Reinstalling ${outputPath}: the executable ${installedBrowser.executablePath} is missing`,
-      );
-      await rm(outputPath, {
-        force: true,
-        recursive: true,
-        maxRetries: 10,
-        retryDelay: 500,
-      });
+      await runSetup(installedBrowser, logger);
+      if (options.installDeps) {
+        await installDeps(installedBrowser, logger);
+      }
+      return installedBrowser;
     }
 
     // Check if archive already exists (e.g., from a custom provider)
--- a/packages/browsers/src/install.ts
+++ b/packages/browsers/src/install.ts
@@ -452,7 +452,7 @@ async function installUrl(
         throw new Error(
           `The browser folder (${outputPath}) exists but the executable (${installedBrowser.executablePath}) is missing. ` +
             `An earlier install of this build probably did not finish. ` +
-            `Run \`npx puppeteer browsers clear\` to empty the browser cache, then install again.`,
+            `Delete ${outputPath} and install the browser again.`,
         );
       }
       await runSetup(installedBrowser, logger);
--- a/packages/browsers/src/install.ts
+++ b/packages/browsers/src/install.ts
@@ -7,7 +7,7 @@
 import assert from 'node:assert';
 import {spawnSync} from 'node:child_process';
 import {existsSync, readFileSync} from 'node:fs';
-import {mkdir, unlink} from 'node:fs/promises';
+import {mkdir, rm, unlink} from 'node:fs/promises';
 import os from 'node:os';
 import path from 'node:path';
 
@@ -448,16 +448,22 @@ async function installUrl(
 
   try {
     if (existsSync(outputPath)) {
-      if (!existsSync(installedBrowser.executablePath)) {
-        throw new Error(
-          `The browser folder (${outputPath}) exists but the executable (${installedBrowser.executablePath}) is missing`,
-        );
+      if (existsSync(installedBrowser.executablePath)) {
+        await runSetup(installedBrowser, logger);
+        if (options.installDeps) {
+          await installDeps(installedBrowser, logger);
+        }
+        return installedBrowser;
       }
-      await runSetup(installedBrowser, logger);
-      if (options.installDeps) {
-        await installDeps(installedBrowser, logger);
-      }
-      return installedBrowser;
+      logger?.(DEBUG_PREFIXES.install)?.(
+        `Reinstalling ${outputPath}: the executable ${installedBrowser.executablePath} is missing`,
+      );
+      await rm(outputPath, {
+        force: true,
+        recursive: true,
+        maxRetries: 10,
+        retryDelay: 500,
+      });
     }
 
     // Check if archive already exists (e.g., from a custom provider)
--- a/packages/browsers/src/CLI.ts
+++ b/packages/browsers/src/CLI.ts
@@ -21,7 +21,7 @@ import {
 } from './browser-data/browser-data.js';
 import {Cache} from './Cache.js';
 import {detectBrowserPlatform} from './detectPlatform.js';
-import {install} from './install.js';
+import {IncompleteInstallationError, install} from './install.js';
 import {
   computeExecutablePath,
   computeSystemExecutablePath,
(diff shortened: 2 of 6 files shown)


itskrishnyug-alt (commented):
Approved

itskrishnyug-alt:
Could a maintainer please approve running the workflow checks for this PR? Thanks!

itskrishnyug-alt (commented):
Approved

OrKoN (changes requested):
I am not sure we should be making rm() calls during the install as it might not always be bullet-proof (i.e, executable moves or external providers have a different executable). WDYT instead of running `npx puppeteer browsers clear` if you find that the cache folder was corrupted for your installation? Note that there were recent corruption issues related to regressions in Node 25 and generally corruptions should rarely happen.

lazerg:
Makes sense, and #15218 came to the same conclusion after your review there, so I dropped the `rm()`.

The install keeps the throw. The error message now adds what to do: an earlier install of that build probably did not finish, so run `npx puppeteer browsers clear` and install again. Nothing is deleted for the user, so a moved executable or a provider with a different layout cannot lose a working install.

I think that still covers the report. What makes it stick is not the folder, it is that the message says nothing about recovery. The postinstall exits 1 and npm hides its output unless you pass `--foreground-scripts`, so the user sees a failed install and no reason for it. With the command in the error there is something to act on.

One thing I left out on purpose: removing `outputPath` when the unpack in that same install fails. That directory is ours in that path, so it is safer than touching a pre-existing one, but #15218 chose to leave partial output alone too. Happy to do it separately if you want it.

ystknsh:
Reporter of #15318 here.

What I am after is simply this: when an install fails under npm, I want the reason to appear on screen. What I hit was running `npx` on a package that depends on puppeteer — it printed nothing at all and returned to the prompt.

The new message will not reach that path unless the user passes `--foreground-scripts`. npm runs a dependency's lifecycle scripts with `stdio: 'pipe'` by default, and on failure it keeps only the exit code (`arborist/lib/arborist/rebuild.js`):

```js
runScript(runOpts).catch(er => {
  const { code, signal } = er
  log.info('run', pkg._id, event, { code, signal })
  throw er
})
```

The success path a few lines below stores `stdout` and `stderr`; this one does not. In my run that left exactly two lines in the debug log and nothing on screen, with no `error` entry anywhere:

```
info run puppeteer@25.5.0 postinstall node install.mjs
info run puppeteer@25.5.0 postinstall { code: 1, signal: null }
```

So my suggestion is about where the failure happens rather than how it is worded: when a partial folder is found, could the postinstall warn and exit 0 instead of failing? The failure would then surface at `launch()`, from the application's own process, which npm does not pipe — so it would actually be read. `PUPPETEER_SKIP_DOWNLOAD` already makes "install succeeded, no browser present" a supported state, and `install.mjs` already warns and exits 0 rather than failing the install when it cannot load the installer at all — a different condition, but the same shape.

The trade-off is that `npm install` would report success with no browser installed. But today it does not report a visible failure either — it is silence plus an exit code.

For what it is worth on frequency: this was Node v22.22.2 with `@puppeteer/browsers@3.1.0`, so not the Node 25 regression. An interruption during extraction is enough on its own.

No objection to landing the message change.

lazerg:
Checked this against the arborist source: you're right, `rebuild.js` defaults `stdio` to `pipe` and the failure branch only logs `{ code, signal }`, never stdout/stderr. So the message text won't reach the user through a piped `npx` install like yours.

Warn-and-exit-0 is a bigger behavior change than this PR though, not just wording, and @OrKoN already pushed back twice on this PR toward the narrower throw-plus-message approach (dropped an earlier rm-and-reinstall attempt, then a repair-in-place one). I'd rather let him weigh in on whether shifting the failure to launch() fits that same reasoning before touching install()'s control flow again.

packages/browsers/src/install.ts
```diff
@@ -450,7 +450,9 @@ async function installUrl(
     if (existsSync(outputPath)) {
       if (!existsSync(installedBrowser.executablePath)) {
         throw new Error(
-          `The browser folder (${outputPath}) exists but the executable (${installedBrowser.executablePath}) is missing`,
+          `The browser folder (${outputPath}) exists but the executable (${installedBrowser.executablePath}) is missing. ` +
+            `An earlier install of this build probably did not finish. ` +
+            `Run \`npx puppeteer browsers clear\` to empty the browser cache, then install again.`,
```
OrKoN: `@puppeteer/browsers` is a separate package and has its own CLI, so `npx puppeteer browsers clear` would not always apply. I think we should customize the message in the puppeteer's package install scripts or in the CLI implementation class.

OrKoN:
FWIW npm 12 stops running postinstall scripts by default. Therefore, skipping the browser installation during `npm install` becomes the default behavior. We currently recommend running `npx puppeteer browsers install` when you need to download browsers (https://pptr.dev/#installation) where errors would be surfaced. Fixing the exit codes should not involve changes to `@puppeteer/browsers`:

1) this should be converted to throwing an error https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer/src/node/install.ts#L144

2) combined with an `await` here https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer/install.mjs#L40.

I think it still makes sense to improve the message text (left a comment about that).

packages/browsers/src/install.ts
lazerg: Agreed. `@puppeteer/browsers` no longer names a CLI. The message now says the folder is incomplete and asks the user to delete that folder, which holds for anyone using the package on its own.

The `npx puppeteer browsers clear` hint moved to `packages/puppeteer/src/node/install.ts`, on the failure path of `downloadBrowsers()`.

One thing to weigh: that is the postinstall path you said npm 12 skips by default. `npx puppeteer browsers install` goes through the CLI class, so it only gets the generic delete-the-folder text. Say the word and I will put the hint in `CLI.ts` as well, built from `scriptName` and `prefixCommand`, so each CLI prints its own command.

lazerg:
Both done. `downloadBrowsers()` throws instead of calling `process.exit(1)`, and `install.mjs` awaits it.

The exit code changes as a result. I checked it against a real partial folder in the cache:

```
# before
$ PUPPETEER_CACHE_DIR=/tmp/pcache node install.mjs
Error: ERROR: Failed to set up chrome v151.0.7922.77! ...
EXIT=1

# after
$ PUPPETEER_CACHE_DIR=/tmp/pcache node install.mjs
Error: ERROR: Failed to set up chrome v151.0.7922.77! ...
Browser download failed Error: Failed to download one or more browsers. ...
EXIT=0
```

The existing catch only warns, so the postinstall now reports success and the failure surfaces later at `launch()`. That is what @ystknsh asked for. If you would rather keep a non-zero exit, `process.exitCode = 1` in that catch is one line. Tell me which one you want.

The `await` also changes the other throws in `downloadBrowsers()`, such as the unsupported-platform one. Those used to escape the try/catch as unhandled rejections and print a stack trace. They now go through the same warning.

One consequence I did not touch: `ci.yml`, `daily.yml` and `deflake.yml` each run `npm run postinstall` as an explicit step before the tests. That step used to stop the job when a download failed. It now passes, and the job runs on until the tests report a missing Chrome. Pointing those steps at `npx puppeteer browsers install` keeps the fail-fast. Happy to do it here or in a separate PR.

No test covers the exit code. `packages/puppeteer` only unit-tests `getConfiguration`, so this would need a child process and a real download attempt. I did not think the scaffolding was worth it, but I will add it if you want it.

packages/browsers/src/install.ts
OrKoN: Let's add detailed instructions to a catch around install() here https://github.com/puppeteer/puppeteer/blob/main/packages/browsers/src/CLI.ts#L638 using scriptName and prefixCommand like you suggest. We can define an Error-subclass to distinguish this situation specifically. I think there could be more refactoring done to clearly separate responsibilities but I think it's out of scope for this fix (allowing to clear a specific browser/browser version via CLI and adjusting the message to tell that, removing integrity validation logic out of install() altogether and moving to the CLI).

lazerg: Built it that way.

`installUrl()` now throws an `IncompleteInstallationError` for the missing-executable case, and the message it carries still names no CLI. `installWithProviders()` keeps that class when every provider failed for the same reason, otherwise the type would be lost in the aggregate error it builds.

`#install()` catches the class and rethrows with the command built from `scriptName` and `prefixCommand`. `npx puppeteer browsers install` now ends with:

```
You can also empty the whole browser cache with `npx puppeteer browsers clear` and then run `npx puppeteer browsers install` again.
```

and a plain `npx @puppeteer/browsers install` names its own commands instead. Every other error passes through untouched.

The class is exported from `install.ts` but not from `main.ts`, so it stays internal for now. Say the word if you want it public so callers of `install()` can check it too.

`CLI.test.ts` covers both sides: an incomplete folder prints the prefixed command, and a failure from an unreachable base URL does not.

I left the two follow-ups you named alone: clearing a single browser or version, and moving the validation out of `install()`.

One note on the postinstall path. It calls `install()` directly rather than through the CLI, so it keeps the line pointing at `npx puppeteer browsers install`. That one is a retry hint for any download failure, not a second copy of this message.

lazerg:
@ystknsh coming back to this, since it ended up going your way.

@OrKoN named the two edits that get there, and they are in the PR now. `downloadBrowsers()` throws instead of calling `process.exit(1)`, and `install.mjs` awaits it. The catch there only warns and sets no exit code, so the postinstall now ends at 0. I ran it against a partial folder in a real cache dir: exit 1 before, exit 0 after. `npm install` reports success, and the missing browser shows up at `launch()`, which prints `Could not find Chrome ...` and points at `npx puppeteer browsers install`. That is the trade-off you described, and I think it is the right one for the case you hit.

The message text still will not reach a piped install, for the reason you found. The exit code is what changes what you see, not the wording.

One caveat so you are not surprised later. I asked @OrKoN whether he wants a non-zero exit kept, because `process.exitCode = 1` in that catch is a one-line change. If he does, npm goes back to failing quietly and only the message improves.

The message part moved as well. `@puppeteer/browsers` throws a distinct error for the incomplete-folder case, and each CLI prints its own `clear` and `install` commands when it catches that error.

ystknsh:
On exit 0 vs `process.exitCode = 1`: for the case I reported, exit 0 is the one that fixes it.

A non-zero exit puts it back where it started — npm treats the script as failed and aborts the install, and the warning `downloadBrowsers()` now prints goes to the same piped stderr that never reached my terminal. That is the silent failure again, just with a different message behind it. Exit 0 lets the install finish and moves the error to `launch()`, where it is printed by the user's own process and actually read.

Thanks for turning this around so quickly.

OrKoN (approved):
lgtm thanks!

ystknsh:
Thanks both — this ended up in a much better place than the report I started from.

PR #15325, opened 2026-08-10
feat: allow downloading CfT linux-arm64

Lightning00Blade:


OrKoN (approved):

packages/browsers/src/browser-data/chrome-headless-shell.ts
```diff
@@ -5,11 +5,16 @@
...
 import path from 'node:path';
 
+import {compareVersions} from './chrome.js';
 import {BrowserPlatform} from './types.js';
 
-function folder(platform: BrowserPlatform): string {
+function folder(platform: BrowserPlatform, buildId?: string): string {
   switch (platform) {
     case BrowserPlatform.LINUX_ARM:
+      if (buildId && compareVersions(buildId, '153.0.7998.0') < 0) {
```
mathiasbynens: `linux-arm64` CfT builds are available [starting with version **153.0.8001.0**](https://github.com/GoogleChromeLabs/chrome-for-testing#supported-platforms). To avoid confusion about what the minimum supported version is, let’s use that specific version number in the code (here and elsewhere) instead of `153.0.7998.0`.

mathiasbynens (approved):

PR #15332, opened 2026-08-12
refactor: use Desposable stacks to track listeners

Lightning00Blade:


packages/puppeteer-core/src/cdp/DeviceRequestPrompt.ts
```diff
@@ -110,41 +112,27 @@ export class CdpDeviceRequestPrompt extends DeviceRequestPrompt {
...
-      this.#updateDevicesHandle,
-    );
-    this.#handled = true;
-    return await this.#client.send('DeviceAccess.selectPrompt', {
+    await this.#client.send('DeviceAccess.selectPrompt', {
       id: this.#id,
       deviceId: device.id,
     });
+    this.#subscriptions.dispose();
+    this.#handled = true;
```
OrKoN: is there a reason why we swap the order? I think setting handled to true first might be intended to avoid double select?

OrKoN (approved):

packages/puppeteer-core/src/cdp/DeviceRequestPrompt.ts
Lightning00Blade: Well if the command fails do we want to mark it as such?

OrKoN: yeah I think so, if it fails it's probably already handled or won't be ever handled due to broken state in the browser. In any case, I think we can maintain the original order for the sake of this being a refactoring.

PR #15335, opened 2026-08-13
feat: support followSymlinks option

OrKoN:
followSymlinks defaults to true to keep existing behavior. Setting `followSymlinks=false` will make Puppeteer to stop following symlinks when writing or reading files. Implemented by adding environment file access functions and enforcing the configuration globally.

### DIFF
--- a/test/src/followSymlinks.test.ts
+++ b/test/src/followSymlinks.test.ts
@@ -1,0 +1,214 @@
+/**
+ * @license
+ * Copyright 2026 Google Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+import expect from 'expect';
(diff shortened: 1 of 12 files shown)


OrKoN:
PTAL!

test/src/followSymlinks.test.ts
```diff
@@ -0,0 +1,214 @@
...
+      puppeteer.setFollowSymlinks(false);
+    });
+
+    afterEach(() => {
+      puppeteer.setFollowSymlinks(true);
+    });
+
+    it('should reject addScriptTag with a symlinked path', async function () {
+      if (!symlinksSupported || process.platform === 'win32') {
+        this.skip();
```
Lightning00Blade, line 63: Should we return after this?

Lightning00Blade (approved):

test/src/followSymlinks.test.ts
OrKoN: no, this throws for mocha so return is not needed

PR #15339, opened 2026-08-13
fix: launch browsers in detached mode on Windows to fix flakiness

OrKoN:
- remove errors that can be uncaught
- increase rm retries with an explicit delay
- handle test runner's exit codes better

### DIFF
--- a/packages/browsers/src/launch.ts
+++ b/packages/browsers/src/launch.ts
@@ -318,11 +318,10 @@ export class Process {
     opts.handleSIGINT ??= true;
     opts.handleSIGTERM ??= true;
     opts.handleSIGHUP ??= true;
-    // On non-windows platforms, `detached: true` makes child process a
-    // leader of a new process group, making it possible to kill child
-    // process tree with `.kill(-pid)` command. @see
-    // https://nodejs.org/api/child_process.html#child_process_options_detached
-    opts.detached ??= process.platform !== 'win32';
+    // `detached: true` makes child process a leader of a new process group,
+    // making it possible to isolate process trees and prevent console signal propagation.
+    // @see https://nodejs.org/api/child_process.html#child_process_options_detached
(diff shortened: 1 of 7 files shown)


packages/browsers/src/launch.ts
```diff
@@ -318,11 +319,10 @@ export class Process {
...
     opts.handleSIGHUP ??= true;
-    // On non-windows platforms, `detached: true` makes child process a
-    // leader of a new process group, making it possible to kill child
-    // process tree with `.kill(-pid)` command. @see
-    // https://nodejs.org/api/child_process.html#child_process_options_detached
-    opts.detached ??= process.platform !== 'win32';
+    // `detached: true` makes child process a leader of a new process group,
+    // making it possible to isolate process trees and prevent console signal propagation.
+    // @see https://nodejs.org/api/child_process.html#child_process_options_detached
+    opts.detached ??= true;
```
OrKoN, line 324: Need to test this on Windows but it should be fine with windowsHide=true.

tools/mocha-runner/src/mocha-runner.ts
```diff
@@ -329,11 +345,15 @@ async function main() {
...
+    if (fail) {
+      console.log(
+        unexpected > 0
+          ? `Run failed: ${unexpected} unexpected result(s).`
+          : `Run failed.`,
+      );
+    } else {
+      console.log('Run succeeded.');
+    }
     process.exit(fail ? 1 : 0);
```
Lightning00Blade:
```suggestion
    if (fail) {
      console.log(
        unexpected > 0
          ? `Run failed: ${unexpected} unexpected result(s).`
          : `Run failed.`,
      );
      process.exit(1);
    }
    console.log('Run succeeded.');
    process.exit(0);
```

Lightning00Blade (approved):

PR #15342, opened 2026-08-14
feat: emit console and network events on Browser and BrowserContext

### LINKED ISSUE #15270: [Feature]: Emit console and network events on the Browser and Browser context classes
### Feature description

Currently, if a new page is created or when Puppeteer attaches to an existing browser instance, console and network events are sent by CDP while the client does not have a handle to the Page object to be able to catch them. We should re-emit all events on Browser and BrowserContext classes with links to the Page object.

Related https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2382 https://github.com/ChromeDevTools/chrome-devtools-mcp/issues/2395

After that we should be able to rewrite collectors to use the browser level events instead. Special consideration would be needed for the checks that determine if there are any listeners before emitting events.

Pseudo-code:

```
browser.on('console', msg => // msg.page() or msg.worker());
browser.on('request', msg => // msg.page() or msg.worker());
```

yasirusman85:
Emits console and network events at Browser/BrowserContext level and buffers/replays them during puppeteer.connect. Fixes https://github.com/puppeteer/puppeteer/issues/15270

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27825,9 +27825,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -27929,9 +27926,6 @@
         "arm"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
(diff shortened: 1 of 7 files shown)


Lightning00Blade:
Thanks for the PR, we are currently evaluating mechanism and API with the team, and would conclude that before we commit to an implementation.

PR #15346, opened 2026-08-15
docs: mention pnpm build script approval

Soundcreates:
Adds a concrete pnpm guidance note to the installation troubleshooting documentation.

Assisted-by: OpenAI GPT-5


OrKoN:
Thanks for the PR but I do not think we need to add these pnpm specific instructions: 1) they are not detailed enough 2) the npm is only an example and we do not want to be exhaustive documenting the behavior of all package managers out there.

PR #15372, opened 2026-08-23
chore: remove doctests

OrKoN:


Lightning00Blade (approved):

Lightning00Blade:
Better name would be `chore: remove doctests` also please double check if there are issues open related to this.

Issue #15379, opened 2026-08-24
[Bug]: CDP mouse throws on a repeated mouse.down() and releases buttons in the opposite order in mouse.reset()


Develop-KIM:
### Minimal, reproducible example

```js
import puppeteer from 'puppeteer';

const protocol = process.env.PROTO; // 'cdp' or 'webDriverBiDi'
const browser = await puppeteer.launch({protocol});
const page = await browser.newPage();

try {
  await page.mouse.down();
  await page.mouse.down();
  console.log(`[${protocol}] repeated down(): no throw`);
} catch (error) {
  console.log(`[${protocol}] repeated down(): ${error.message}`);
}

await page.mouse.reset();
await page.mouse.move(5, 5);
await Promise.all([
  page.mouse.down({button: 'left'}),
  page.mouse.down({button: 'middle'}),
  page.mouse.down({button: 'right'}),
]);
await page.evaluate(() => {
  globalThis.ups = [];
  document.addEventListener('mouseup', event => globalThis.ups.push(event.button));
});
await page.mouse.reset();
console.log(`[${protocol}] release order:`, await page.evaluate(() => globalThis.ups));

await browser.close();
```

Output on `main` (Chrome 152, macOS):

```
[cdp] repeated down(): 'left' is already pressed.
[cdp] release order: [ 0, 1, 2 ]
[webDriverBiDi] repeated down(): no throw
[webDriverBiDi] release order: [ 2, 1, 0 ]
```

### Background

Both differences are already covered by `test/src/mouse.test.ts`, and both are pinned in
`test/TestExpectations.json` as CDP-only entries whose comment is still the placeholder
`"TODO: add a comment explaining why this expectation is required (include links to issues)"`:

- `[mouse.test] Mouse should not throw if buttons are pressed twice` — `parameters: ["cdp"]`, `["FAIL"]`
- `[mouse.test] Mouse should reset properly` — `parameters: ["cdp"]`, `["SKIP"]`

The second one has been skipped for CDP ever since `reset()` was added in #10340 — the
expectations in that test were written against the WebDriver BiDi behaviour.

The two divergences:

1. `CdpMouse.down()` throws `'<button>' is already pressed.` when the button is already
   down. `BidiMouse.down()` keeps no state and just dispatches the pointerDown, and the
   throw is not part of the documented `Mouse.down()` contract.

2. `CdpMouse.reset()` walks a fixed `Left → Middle → Right → Forward → Back` list and
[...]

PR #15380, opened 2026-08-24
test: explain the CDP-only mouse expectations

Develop-KIM:

Test/documentation change. This PR started as a change to `CdpMouse` (no throw on a duplicate
`down()`, release in reverse press order) and was rescoped after @OrKoN's feedback on #15379:
the CDP behaviour stays as is for compatibility, and instead the two CDP-only entries in
`test/TestExpectations.json` that still carried the placeholder
`"TODO: add a comment explaining why this expectation is required"` now explain what they pin:

- `[mouse.test] Mouse should not throw if buttons are pressed twice` (FAIL on CDP) —
  `BidiMouse` keeps no button state and forwards the repeated `down()`, `CdpMouse` throws on
  a duplicate `down()` by design.
- `[mouse.test] Mouse should reset properly` (SKIP on CDP) — the test asserts WebDriver's
  release order (reverse of the press order, pointer left in place), while `CdpMouse`
  releases in a fixed order and moves the pointer back to `(0, 0)` as `Mouse.reset()` documents.

Both tests also get a short comment saying they exercise the WebDriver BiDi behaviour.

No, only comments; no expectations change. `npm run lint` and prettier are clean.

I used Claude Code while working on this; I reviewed the change myself.

Issues: #15379

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -611,14 +611,14 @@
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["cdp"],
     "expectations": ["FAIL"],
-    "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
+    "comment": "The test exercises the WebDriver BiDi behavior: BidiMouse keeps no button state and forwards a repeated down() to the browser, while CdpMouse deliberately throws on a duplicate down(). Kept for CDP compatibility, see https://github.com/puppeteer/puppeteer/issues/15379"
   },
   {
     "testIdPattern": "[mouse.test] Mouse should reset properly",
     "platforms": ["darwin", "linux", "win32"],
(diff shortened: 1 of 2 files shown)


OrKoN (changes requested):
I left a comment on the issue (I am not sure we should do changes to the CDP implementation for compatibility reasons)

Develop-KIM:
Fair enough — I've rescoped this to what you suggested on #15379. The `CdpMouse` change is gone; the two CDP-only entries in `TestExpectations.json` now explain that the tests pin WebDriver BiDi behaviour (no client-side button state, release in reverse press order without moving the pointer), and the tests carry a short comment saying the same. Title and description updated to match.

OrKoN (approved):

PR #15387, opened 2026-08-25
feat: implement video-stream-based screen recording via page.record()

OrKoN:
This is available starting from M153 and with M154 carrying improvements. Covers CDP and WebDriver BiDi. Since WebDriver BiDi API is filed based the file is converted to a stream after the recording for compatibility.

The new API deprecates the previous ffmpeg-based API and does not require any dependencies.

### DIFF
--- a/packages/puppeteer-core/src/api/Page.ts
+++ b/packages/puppeteer-core/src/api/Page.ts
@@ -33,7 +33,6 @@ import type {HTTPResponse} from '../api/HTTPResponse.js';
 import type {Accessibility} from '../cdp/Accessibility.js';
 import type {Coverage} from '../cdp/Coverage.js';
 import type {NetworkConditions} from '../cdp/NetworkManager.js';
-import {ScreenRecording} from '../cdp/ScreenRecording.js';
 import type {Tracing} from '../cdp/Tracing.js';
 import type {WebMCP} from '../cdp/WebMCP.js';
 import type {ConsoleMessage} from '../common/ConsoleMessage.js';
@@ -120,6 +119,7 @@ import {
   type AwaitedLocator,
 } from './locators/locators.js';
 import type {Realm} from './Realm.js';
+import type {ScreenRecording} from './ScreenRecording.js';
 import type {Target} from './Target.js';
 import type {WebWorker} from './WebWorker.js';
 
@@ -2680,6 +2680,13 @@ export abstract class Page extends EventEmitter<PageEvents> {
   }
 
   /**
+   * @internal
+   */
+  protected abstract createScreenRecording(
+    options: Readonly<RecordOptions>,
+  ): ScreenRecording;
+
+  /**
    * Records this {@link Page | page} using the Chrome DevTools Protocol
    * {@link https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-startScreenRecording | Page.startScreenRecording}
    * API.
@@ -2746,7 +2753,7 @@ export abstract class Page extends EventEmitter<PageEvents> {
         })
       : undefined;
 
-    const recording = new ScreenRecording(this, options, this.logger);
+    const recording = this.createScreenRecording(options);
 
     try {
       await recording._start();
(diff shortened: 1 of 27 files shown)


packages/puppeteer-core/src/api/Page.ts
```diff
@@ -2638,6 +2679,89 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
+   * await recorder.stop();
+   *
+   * await browser.close();
+   * ```
+   *
+   * @param options - Configures recording behavior.
+   *
+   * @experimental
+   */
+  async record(
```
Lightning00Blade, line 2726: Let's extract what we can in protected helper here, and then re-use in the respective protocols. In general we should try no to import CDP files here. And I think with this we can avoid it. 
Alternatively we require a helper function like `protected _ScreenRecording` that is abstract here. And we return a public interface rather then use the CDP one as a base.

packages/puppeteer-core/src/bidi/ScreenRecording.test.ts
```diff
@@ -0,0 +1,330 @@
...
+    } finally {
+      environment.value.readFile = originalReadFile;
+    }
+  });
+
+  it('should support fps as alias for frameRate', async () => {
+    const context = new MockBrowsingContext();
+    const page = new MockBidiPage(context);
+
+    const originalReadFile = environment.value.readFile;
```
Lightning00Blade: Can we reset this in a before hook, that way we don't need to have try/catch blocks

packages/puppeteer-core/src/bidi/ScreenRecording.ts
```diff
@@ -0,0 +1,124 @@
...
+ */
+export class BidiScreenRecording extends ScreenRecording {
+  declare protected page: BidiPage;
+  #screencastId?: string;
+  #path?: string;
+
+  /**
+   * @internal
+   */
+  constructor(page: BidiPage, options: RecordOptions = {}, logger?: Logger) {
```
Lightning00Blade: Can all the logger be required? I think at the point this is create it should always be there.

packages/puppeteer-core/src/bidi/ScreenRecording.ts
```diff
@@ -0,0 +1,124 @@
...
+   */
+  @guarded()
+  override async stop(): Promise<void> {
+    if (this.stopped) {
+      return;
+    }
+    this.stopped = true;
+
+    try {
+      if (this.#screencastId) {
```
Lightning00Blade: Can this be a early return instead?

packages/puppeteer-core/src/bidi/ScreenRecording.ts
```diff
@@ -0,0 +1,124 @@
...
+        const result = await this.page
+          .mainFrame()
+          .browsingContext.stopScreencast(this.#screencastId)
+          .catch(err => {
+            this.logger?.(DEBUG_PREFIXES.error)?.(err);
+            return undefined;
+          });
+
+        if (result?.error) {
+          this.logger?.(DEBUG_PREFIXES.error)?.(new Error(result.error));
```
Lightning00Blade: Why do we want to wrap it in a new error?

Lightning00Blade (commented):
LGTM, with comments

packages/puppeteer-core/src/cdp/ScreenRecording.ts
```diff
@@ -0,0 +1,105 @@
...
+    const {client} = this.page.mainFrame();
+    const frameRate = this.options.frameRate ?? this.options.fps;
+    // @ts-expect-error Page.startScreenRecording is not yet in devtools-protocol
+    const result = (await client.send('Page.startScreenRecording', {
+      audio: this.options.audio,
+      maxWidth: this.options.maxWidth,
+      maxHeight: this.options.maxHeight,
+      frameRate,
+    })) as {stream?: string} | undefined;
+    if (result && typeof result.stream === 'string') {
```
Lightning00Blade: Can you explain why we need this? 
https://chromedevtools.github.io/devtools-protocol/tot/Page/#method-startScreenRecording
It looks like this should always be there. And if not I would expect this to throw.

Lightning00Blade (approved):

PR #15389, opened 2026-08-25
fix: do not miss the first screencast frame while startScreencast is in flight

### LINKED ISSUE #15398: [Bug]: page.screencast() can hang when the first frame is ACKed before startScreencast returns

Public `page.screencast()` on `puppeteer-core@25.5.0` (same `_startScreencast` sequence is on current `main`). Real Chrome, real ffmpeg, real `ScreenRecorder` (the only ACK source). No CDP `send` hook.

**Quiet path:** a still HTML page with no long-task did **not** hang here (`page.screencast()` returned in ~50–130ms, 3/3). Treat that as the usual case.

**Hang path:** the same page, after first paint, blocks the renderer with a long-task. Chromium can still emit the first `Page.screencastFrame` from the browser process while `Page.startScreencast` is waiting on the renderer `FallThrough` ACK. `ScreenRecorder` ACKs that frame. Readiness only attaches `Page.screencastFrame` in the start-command `.then()`, so it misses the only frame and the promise never settles.

The busy-loop is a **scheduling amplifier** (a page long-task). It does not synthesize or ACK frames.

`screencast-hang.html`:

```html
<!doctype html>
<meta charset="utf-8" />
<title>screencast hang</title>
<style>
  html, body { margin: 0; height: 100%; background: #1b1b1f; color: #fff; }
  body { display: grid; place-items: center; font: 24px system-ui, sans-serif; }
</style>
<h1>warm static page</h1>
<script>
  requestAnimationFrame(() => {
    requestAnimationFrame(() => {
      setTimeout(() => {
        const end = performance.now() + 8000;
        while (performance.now() < end) {}
      }, 0);
    });
  });
</script>
```

`repro.mjs`:

```js
import {pathToFileURL} from 'node:url';

const browser = await puppeteer.launch({headless: true});
await page.goto(pathToFileURL(new URL('./screencast-hang.html', import.meta.url)).href, {
  waitUntil: 'domcontentloaded',
});

const rec = await Promise.race([
  page.screencast({path: 'out.webm'}),
  new Promise((_, reject) =>
    setTimeout(() => reject(new Error('page.screencast() still pending')), 12_000),
  ),
]);

await rec.stop();
await browser.close();
```

Requires Chrome and ffmpeg.

This is **timing-dependent**. If the long-task starts before `page.screencast()`'s viewport `evaluate`, start is issued after the page unblocks and the call returns. If start is issued while the renderer is blocked, a first frame can arrive immediately and the call stays pending after the 8s long-task ends.

Observed hang (stock 25.5.0): first `Page.screencastFrame` at ~80ms; `page.screencast()` still pending at 12s.

**Expected:** `page.screencast()` returns a `ScreenRecorder`.

**Actual:** the promise can stay pending with no error. ffmpeg may already be running.

_No response_

### Puppeteer version

25.5.0 (same readiness sequence on current `main`)

### Node version

22.23.2

### Package manager

npm

### Package manager version

10.9.8

### Operating system

macOS

devonpmack:

Bug fix for `page.screencast()` never settling when the first `Page.screencastFrame` is dispatched (and ACKed) before `Page.startScreencast` returns.

Fixes #15398

No. The `Screencasts` mocha file is not run in CI (no ffmpeg). Extra cases added there were removed. There are no dedicated unit tests for this waiter.

N/A

**Summary**

`Page._startScreencast()` sent `Page.startScreencast` and attached the first-frame `once` listener only in that command's `.then()`. `ScreenRecorder` is already subscribed and ACKs frames. Chromium can emit a frame while the start result is still in flight (`FallThrough` waiting on the renderer). If that frame is the only one, readiness never settles.

The usual quiet path is fine: command result, then frame. A still page with no long-task returned on stock 25.5.0 (~50–130ms, 3/3).

The hang is the other order (frame + ACK, then start result). The issue HTML uses a renderer long-task as a scheduling amplifier so `FallThrough` is delayed; it does not fake frames or ACKs.

**Fix**

Register `once('Page.screencastFrame')` first, then `send('Page.startScreencast')`, then wait on that same promise — same `_startScreencast` as before, promises swapped.

```ts
const firstFrame = new Promise<void>(resolve => {
  client.once('Page.screencastFrame', () => resolve());
});
this.#startScreencastPromise = client
  .send('Page.startScreencast', {format: 'png'})
  .then(() => firstFrame);
```

**Local check** (`81d119416c1`, public `page.screencast()`, real Chrome 152, real ffmpeg, no `send` hook, issue HTML):

- Stock 25.5.0: hung — first frame at 495ms, still pending at 12s.
- This branch: 8/8 returned in ~8s (the page long-task). Several runs had an early frame (~60–100ms) and still settled.

No

**Other information**

- Repro: #15398
- Affects `page.screencast()` via `_startScreencast`
- No public API change


OrKoN:
I am not sure I follow, you say this does not work:

```
await client.send('Page.startScreencast', {format: 'png'});
await new Promise(resolve => client.once('Page.screencastFrame', resolve));
```

but the actual code is:

```
this.#startScreencastPromise = this.mainFrame()
  .client.send('Page.startScreencast', {format: 'png'})
  .then(() => {
    // Wait for the first frame.
      return this.mainFrame().client.once('Page.screencastFrame', () => {
        return resolve();
      });
    });
  });
```

That code should work fine because the listener is attached in the microtask queue (promise callbacks) and the CDP response always comes as a separate task (meaning that the listener would be always attached first). Could you please clarify?

packages/puppeteer-core/src/util/screencast.test.ts
```diff
@@ -0,0 +1,123 @@
...
+    return this;
+  }
+
+  ackStart(): void {
+    this.#resolveSend?.();
+  }
+
+  emitFrame(): void {
+    const listeners = [...this.frameListeners];
+    for (const listener of listeners) {
```
OrKoN: this is not correctly emulating CDP response dispatch: the correct way would be 
```
setTimeout(() => {
   ... trigger listeners
}, 0)
```

Does the issue reproduce in the current code with the correct dispatch mock?

devonpmack:
Really appreciate that quick review @OrKoN  🙏 

You're right I looked into this more and a lot more complicated than I thought. I updated the fix + the issue with a proper reproduction: https://github.com/puppeteer/puppeteer/issues/15398

This is my understanding:

`page.screencast()` refuses to return until it sees one `Page.screencastFrame`. It only starts listening for that frame after Chrome answers `Page.startScreencast`.

Chrome can emit the frame before that answer, especially if the renderer is busy: capture starts in the browser process, but the start ACK still waits on the page. ScreenRecorder is already subscribed and ACKs the frame. Readiness never sees it.

This is the issue we ran into on certain webpages^

packages/puppeteer-core/src/util/screencast.test.ts
devonpmack: No you're right, this is not the correct way to reproduce the issue. I updated with the proper repro in https://github.com/puppeteer/puppeteer/issues/15398 but its flaky so hard to write a good test.

test/src/cdp/screencast.test.ts
```diff
@@ -35,6 +35,46 @@ describe('Screencasts', function () {
 
       expect(statSync(file.filename).size).toBeGreaterThan(0);
     });
+    it('should start on a static page', async () => {
```
OrKoN: we do not run these tests in CI because we do not have ffmpeg there and since it's an experimental feature we are not goin to change this. We can delete these tests.

packages/puppeteer-core/src/api/Page.ts
```diff
@@ -2647,16 +2648,9 @@ export abstract class Page extends EventEmitter<PageEvents> {
...
-        .client.send('Page.startScreencast', {format: 'png'})
-        .then(() => {
-          // Wait for the first frame.
-          return new Promise(resolve => {
-            return this.mainFrame().client.once('Page.screencastFrame', () => {
-              return resolve();
-            });
-          });
-        });
+      this.#startScreencastPromise = startScreencastAndWaitForFirstFrame(
```
OrKoN: let's just swap the promises inline here and remove the mocks and the dedicated unit tests.

devonpmack:
Local check of the issue HTML repro against `81d119416c1` (inline first-frame wait). Public `page.screencast()`, real Chrome 152, real ffmpeg. No CDP `send` hook.

Page: paints once, then an 8s renderer long-task (`screencast-hang.html` from #15398). `page.screencast()` raced with a 12s timeout.

**Before** (`puppeteer-core@25.5.0`, listener attached in the start-command `.then()`):

```text
{"ok":false,"ms":12002,"error":"page.screencast() still pending","events":[{"ms":495,"kind":"frame"}]}
```

First `Page.screencastFrame` at 495ms; `page.screencast()` still pending at 12s.

**After** (this commit):

```text
fixed 1 {"ok":true,"ms":8008,"events":[{"ms":56,"kind":"frame"}]}
fixed 2 {"ok":true,"ms":8015,"events":[{"ms":82,"kind":"frame"}]}
fixed 3 {"ok":true,"ms":8009,"events":[{"ms":105,"kind":"frame"}]}
fixed 4 {"ok":true,"ms":8017,"events":[{"ms":61,"kind":"frame"}]}
fixed 5 {"ok":true,"ms":8042,"events":[{"ms":8057,"kind":"frame"}]}
fixed 6 {"ok":true,"ms":8017,"events":[{"ms":101,"kind":"frame"}]}
fixed 7 {"ok":true,"ms":8015,"events":[{"ms":102,"kind":"frame"}]}
fixed 8 {"ok":true,"ms":8041,"events":[{"ms":8050,"kind":"frame"}]}
```

8/8 returned in ~8s (the page long-task, not a hang). Several runs saw an early frame (~60–100ms) and still settled.

test/src/cdp/screencast.test.ts
devonpmack: Updated, thanks!

packages/puppeteer-core/src/api/Page.ts
devonpmack: updated!

OrKoN (approved):

PR #15399, opened 2026-08-26
 refactor: extract supported browsers from tags

Lightning00Blade:
Keep a file for the pre v25 as we don't know where the correct revision is stored to extract.


docs/supported-browsers.md
```diff
@@ -18,181 +18,4 @@ Prior to this version Puppeteer downloaded and worked with the nightly versions
 The following table provides mapping between the Puppeteer version and the browsers version you can use it with.
 If an exact matching version of Puppeteer isn't listed, the supported version of the browser is that for the immediately prior version:
 
-{/* version-start */}
```
OrKoN, line 21: I think we still want this doc to have content in the markdown (as a doc, independent of the website).

OrKoN: or we need to drop the entire doc from the docs folder and generate it on the website only. Let's discuss tomorrow.

PR #15405, opened 2026-08-30
fix: distinguish an unwritable userDataDir from a running browser

### LINKED ISSUE #15401: [Bug]: Misleading error "browser is already running" when userDataDir isn't writable

```TypeScript
mkdir /tmp/puppeteer-permission-repro
chmod 555 /tmp/puppeteer-permission-repro



await puppeteer.launch({ userDataDir: '/tmp/puppeteer-permission-repro' });
```

### Background

I was using the [Datadog Synthetics Private Location worker](https://hub.docker.com/r/datadog/synthetics-private-location-worker) with the wrong chmod. This error message had me looking in the wrong spot for a bit.

### Expectation

A more descriptive error message like "The browser failed to create a ProcessSingleton for ${launchArgs.userDataDir}. Use a different \`userDataDir\` or stop the running browser first."

### Reality

Error message "The browser is already running for ${launchArgs.userDataDir}. Use a different \`userDataDir\` or stop the running browser first."

```TypeScript

```

### Puppeteer version

25.9.0

### Node version

v24.10.0

### Package manager

npm

### Package manager version

11.6.0

### Operating system

Linux

rajanpanth:
**Fixes #15401**

### Problem

Chromium logs `Failed to create a ProcessSingleton for your profile directory` for more than one underlying cause: another instance holding the profile lock, and a profile directory the browser cannot write to. `BrowserLauncher` classified that log line as the first case unconditionally, so `chmod 555` on a `userDataDir` produced:

> The browser is already running for /tmp/... Use a different `userDataDir` or stop the running browser first.

which sends you looking for a stray browser process instead of at the directory permissions. The reporter hit this with a wrong `chmod` in a Datadog Synthetics worker image.

### Change

Per @OrKoN's suggestion on the issue ("keep the error message but add a dedicated check for the permissions of the user data dir"), the launcher now checks whether the directory is writable before emitting the "already running" message, and reports the permission problem instead when it is not.

A directory that does not exist yet is treated as writable, since the browser creates it on launch, so this does not change behavior for the ordinary first-run case.

I left the Windows `lockfile` heuristic alone. It has its own false-positive mode (a stale lockfile after a crash), but that is a separate bug from this one and fixing it would need a different signal.

### Test plan

Added a case to `test/src/cdp/userDataDir.test.ts` that creates a temp dir, `chmod 555`s it, attempts a launch, and asserts the permission message, restoring the mode afterwards.

It is guarded to skip on two platforms that cannot observe the failure:

- **Windows** ignores the read-only bit for a directory's owner, so `chmod 555` does not make the directory unwritable.
- **root** bypasses the write permission check entirely, so the launch would succeed.

I verified that guard is necessary rather than assuming it: on this Windows machine `fs.accessSync(dir, W_OK)` still succeeds after `chmod 555`, which is exactly why the test skips there.

**What I could not verify locally:** I develop on Windows, so I could not exercise the POSIX path end to end, which means I have not observed the new message come out of a real launch. The logic and the guard are verified; the browser-level assertion needs CI (or a POSIX reviewer) to confirm. Flagging that explicitly rather than implying a green local run.

`npm run build --workspace=puppeteer-core` passes. `eslint` reports no rule violations on either changed file (the only output is `Delete ␍` line-ending noise from a Windows checkout, which `core.autocrlf` normalizes to LF on commit; the recorded diff contains no CR characters).

### DIFF
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -987,7 +987,7 @@
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["firefox"],
     "expectations": ["SKIP"],
-    "comment": "we currently do not detect this siutation for Firefox"
+    "comment": "we currently do not detect this situation for Firefox"
   },
   {
     "testIdPattern": "[webgl.test] webgl Create webgl context should work",
--- a/test/TestExpectations.json
+++ b/test/TestExpectations.json
@@ -976,6 +976,20 @@
     "comment": "TODO: add a comment explaining why this expectation is required (include links to issues)"
   },
   {
+    "testIdPattern": "[userDataDir.test] userDataDir should report a permission error when the userDataDir is not writable",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["chrome-headless-shell"],
+    "expectations": ["SKIP"],
+    "comment": "headless-shell can launch from an unwritable user data dir, so it never reports the ProcessSingleton failure this message is derived from"
+  },
+  {
+    "testIdPattern": "[userDataDir.test] userDataDir should report a permission error when the userDataDir is not writable",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox"],
+    "expectations": ["SKIP"],
+    "comment": "we currently do not detect this siutation for Firefox"
+  },
+  {
     "testIdPattern": "[webgl.test] webgl Create webgl context should work",
     "platforms": ["darwin", "linux", "win32"],
     "parameters": ["firefox"],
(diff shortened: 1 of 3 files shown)


wolfib:
This looks good to me directionally, but the newly added test is failing.

test/TestExpectations.json
```diff
@@ -975,6 +975,20 @@
...
+    "parameters": ["chrome-headless-shell"],
+    "expectations": ["SKIP"],
+    "comment": "headless-shell can launch from an unwritable user data dir, so it never reports the ProcessSingleton failure this message is derived from"
+  },
+  {
+    "testIdPattern": "[userDataDir.test] userDataDir should report a permission error when the userDataDir is not writable",
+    "platforms": ["darwin", "linux", "win32"],
+    "parameters": ["firefox"],
+    "expectations": ["SKIP"],
+    "comment": "we currently do not detect this siutation for Firefox"
```
```suggestion
    "comment": "we currently do not detect this situation for Firefox"
```

OrKoN (approved):

Issue #12219, opened 2024-04-05
[Bug]: Lost connection is not detected if no messages are being sent over WebSocket


strunkie30:
### Minimal, reproducible example

```TypeScript
const browser = await puppeteer
      .connect({
        browserWSEndpoint: process.env.BROWSER_WS_ENDPOINT,
        protocolTimeout: 5000,
      })
      .catch((err) => {
        console.error('unable to start browser: ', err.error.code);
        this.stop();
      });
```

When i connect to a puppeteer browser, the sockets are closed after a while. But the browser stays connected.

`browser.connected === true`

I can reproduce it when I'am closing the sockets of `process.env.BROWSER_WS_ENDPOINT` that puppeteer is still running and the browser is still connected until I run an action like:
```
const pages = await this.browser.pages();
await pages[0].reload();
```

This scripts throws the error:
```
TargetCloseError: Protocol error (Page.reload): Target closed
    at CallbackRegistry.clear (file:///Users/kevinmeijer/projecten/signeditor-mono/node_modules/.pnpm/puppeteer-core@22.6.2/node_modules/puppeteer-core/src/common/CallbackRegistry.ts:90:30)
    at CdpCDPSession._onClosed (file:///Users/kevinmeijer/projecten/signeditor-mono/node_modules/.pnpm/puppeteer-core@22.6.2/node_modules/puppeteer-core/src/cdp/CDPSession.ts:149:21)
    at Connection.#onClose (file:///Users/kevinmeijer/projecten/signeditor-mono/node_modules/.pnpm/puppeteer-core@22.6.2/node_modules/puppeteer-core/src/cdp/Connection.ts:205:15)
    at WebSocket.<anonymous> (file:///Users/kevinmeijer/projecten/signeditor-mono/node_modules/.pnpm/puppeteer-core@22.6.2/node_modules/puppeteer-core/src/node/NodeWebSocketTransport.ts:50:22)
    at callListener (/Users/kevinmeijer/projecten/signeditor-mono/node_modules/.pnpm/ws@8.16.0/node_modules/ws/lib/event-target.js:290:14)
    at WebSocket.onClose (/Users/kevinmeijer/projecten/signeditor-mono/node_modules/.pnpm/ws@8.16.0/node_modules/ws/lib/event-target.js:220:9)
    at WebSocket.emit (node:events:518:28)
[...]

PR #15406, opened 2026-08-30
fix: detect a websocket connection that died without a close

rajanpanth:
**Towards #12219**

### Problem

The `ws` client only reports a close when the peer sends a TCP FIN. When a connection dies without one, a proxy or load balancer reaping an idle socket, a half-open socket, or a killed remote browser, `ws` never emits `close`, so `Connection`/`browser.connected` keep reporting the browser as connected indefinitely. The failure only surfaces much later as a confusing `TargetCloseError` on the next command.

@OrKoN diagnosed this on the issue ("It looks the client does not auto-detect broken connections") and specified the direction:

> We should probably implement a keepAlive option on the websocket transports and implement the ping/pong to keep the connection alive as described here [ws: how to detect and close broken connections]

### Change

`NodeWebSocketTransport` gained an opt-in `keepAlive` that pings on an interval and calls `terminate()` when the pong for the previous ping never arrived. `terminate()` rather than `close()` because a peer that is not answering will not complete a close handshake; terminating emits `close` locally, which is what surfaces the dead connection to the rest of Puppeteer.

Two details worth calling out:

- The interval is `unref()`d, so keep-alive never holds the process open on its own.
- The timer is cleared on `close` and on the socket's own `close` event, so it cannot outlive the transport.

### Scope, and what I want your call on

This PR deliberately stops at the transport layer and does **not** add a public `ConnectOptions.keepAlive`. Wiring it up means changing the `create()` signature shared with `BrowserWebSocketTransport`, threading the option through the three call sites in `BrowserConnector`, and deciding two things that are yours to decide rather than mine:

1. **Default on or off?** It is off here so nothing changes for existing users, but that also means the reported bug still bites anyone who does not know the option exists. I suspect you want it on by default with a 30s interval; say the word and I will flip it.
2. **What should the browser transport do?** The browser `WebSocket` API has no ping method, so `BrowserWebSocketTransport` cannot implement this natively. It needs either an application-level probe or an explicit "not supported here" stance.

Happy to do the plumbing in this PR or a follow-up, whichever you prefer. I did not want to guess at public API shape and defaults on your behalf.

### Test plan

Three tests in `NodeWebSocketTransport.test.ts`, using `ws`'s `autoPong: false` server option to simulate a peer that is gone but never sent a FIN, which is precisely the case `close` alone cannot detect:

- closes the transport when the peer stops answering pings
- stays open while the peer answers pings
- does not ping at all when `keepAlive` is not enabled

Mutation-tested rather than assumed: reverting only `NodeWebSocketTransport.ts` while keeping the tests, and rebuilding, makes the dead-peer test fail by timing out at 15s while the other two still pass. I verified the revert actually reached the built output (`grep` for `keepAlive` in `lib/` returned 0 in the transport and 2 in the test) so the failure is genuinely the missing fix and not a stale build.

`npm run build --workspace=puppeteer-core` passes. Unit suite: 166/167, the single failure being `Page.test.js` "should reset timeout going over concurrency" (expects `<100ms`, got `110ms`), a timing-sensitive test unrelated to this change that also fails on an unmodified checkout here.

### DIFF
--- a/packages/puppeteer-core/src/common/ConnectOptions.ts
+++ b/packages/puppeteer-core/src/common/ConnectOptions.ts
@@ -134,6 +134,30 @@ export interface ConnectOptions {
    */
   protocolTimeout?: number;
 
+  /**
+   * Whether to send WebSocket pings and drop the connection when a pong does
+   * not come back within the same interval. Detects a connection that died
+   * without a close frame, which otherwise leaves calls hanging until
+   * `protocolTimeout`.
+   *
+   * @remarks
+   * Node.js only. Ignored in the browser build, which has no ping frame API.
+   *
+   * @defaultValue `false`
+   */
+  keepAlive?: boolean;
+
+  /**
+   * Ping period in milliseconds. Only used when {@link ConnectOptions.keepAlive}
+   * is set.
+   *
+   * @remarks
+   * Node.js only. Ignored in the browser build.
+   *
+   * @defaultValue `30_000`
+   */
+  keepAliveIntervalMs?: number;
+
   browserWSEndpoint?: string;
   browserURL?: string;
   transport?: ConnectionTransport;
--- a/packages/puppeteer-core/src/common/ConnectOptions.ts
+++ b/packages/puppeteer-core/src/common/ConnectOptions.ts
@@ -46,6 +46,39 @@ export type ChromeReleaseChannel =
   'chrome' | 'chrome-beta' | 'chrome-canary' | 'chrome-dev';
 
 /**
+ * Options for the WebSocket connection to the browser.
+ *
+ * @remarks
+ * Only used in the Node.js environment.
+ *
+ * @public
+ */
+export interface WsOptions {
+  /**
+   * Headers to use for the web socket connection.
+   */
+  headers?: Record<string, string>;
+
+  /**
+   * Whether to send WebSocket pings and drop the connection when a pong does
+   * not come back within the same interval. Detects a connection that died
+   * without a close frame, which otherwise leaves calls hanging until
+   * `protocolTimeout`.
+   *
+   * @defaultValue `false`
+   */
+  keepAlive?: boolean;
+
+  /**
+   * Ping period in milliseconds. Only used when {@link WsOptions.keepAlive} is
+   * set.
+   *
+   * @defaultValue `30_000`
+   */
+  keepAliveIntervalMs?: number;
+}
+
+/**
  * Generic browser options that can be passed when launching any browser or when
  * connecting to an existing browser instance.
  * @public
@@ -135,28 +168,13 @@ export interface ConnectOptions {
   protocolTimeout?: number;
 
   /**
-   * Whether to send WebSocket pings and drop the connection when a pong does
-   * not come back within the same interval. Detects a connection that died
-   * without a close frame, which otherwise leaves calls hanging until
-   * `protocolTimeout`.
+   * Options for the WebSocket connection to the browser.
    *
    * @remarks
-   * Node.js only. Ignored in the browser build, which has no ping frame API.
-   *
-   * @defaultValue `false`
+   * Only used in the Node.js environment. The browser build has no ping frame
+   * API, so the keep-alive options are ignored there.
    */
-  keepAlive?: boolean;
-
-  /**
-   * Ping period in milliseconds. Only used when {@link ConnectOptions.keepAlive}
-   * is set.
-   *
-   * @remarks
-   * Node.js only. Ignored in the browser build.
-   *
-   * @defaultValue `30_000`
-   */
-  keepAliveIntervalMs?: number;
+  wsOptions?: WsOptions;
 
   browserWSEndpoint?: string;
   browserURL?: string;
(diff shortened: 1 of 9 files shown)


rajanpanth:
The Firefox legs went red here today. I do not think this change can be the cause, and the deciding point is that the new code never runs unless it is asked for.

`keepAlive` is opt-in:

```ts
if (options.keepAlive) {
  this.#startKeepAlive(options.keepAliveIntervalMs ?? DEFAULT_KEEP_ALIVE_INTERVAL_MS);
}
```

`options` defaults to `{}`, and nothing in the repository passes `keepAlive` to `NodeWebSocketTransport`. The only other `keepAlive` in the tree is the unrelated HTTP one in `packages/browsers/src/httpUtil.ts`. So on every existing path, including the BiDi transport Firefox uses, no ping timer is ever started and the transport behaves exactly as before.

Supporting evidence:

- The failures are ordinary page tests, not connection ones: `AriaQueryHandler`, `Heap Snapshot`, `Page.pdf`, `console should work for Error instances`. The shard reports **199 passing, 26 failing**; a broken transport would not leave 199 passing.
- #15396 fails **31** Firefox tests today on its own shards, with a different set again (`Browser specs`, `ElementHandle.backendNodeId`, `network`).
- Firefox legs that last ran on 24 and 26 Aug (#15380, #15399) are green, while everything that ran today is red. The variable is the day, not the diff.

`Puppeteer Unit tests` is green, which covers the tests added here.

Happy to rebase once Firefox is sorted on `main` if that makes the signal easier to read.

packages/puppeteer-core/src/node/NodeWebSocketTransport.ts
```diff
@@ -17,6 +40,7 @@ export class NodeWebSocketTransport implements ConnectionTransport {
     url: string,
     headers: Record<string, string> | undefined,
     logger: Logger,
+    options: NodeWebSocketTransportOptions = {},
```
OrKoN, line 43: Thanks for the PR. I think this needs to be exposed to the connect() and launch() interfaces? we should document that these options are Node-only.

rajanpanth:
Thanks @OrKoN, done in `e616342`.

`keepAlive` and `keepAliveIntervalMs` now live on `ConnectOptions`, which `LaunchOptions` extends, so both `connect()` and `launch()` accept them. They are threaded to the three `WebSocketClass.create` calls in `BrowserConnector` and the two in `BrowserLauncher`.

Both are documented as Node-only in their TSDoc. `BrowserConnector` chooses the transport at runtime, so `BrowserWebSocketTransport.create` takes the same fourth argument and ignores it rather than being left with a narrower signature, since the browser WebSocket API has no ping frame to send. That keeps the union call type-safe instead of needing a cast at each site.

`docs/api/puppeteer.connectoptions.md` is regenerated with `npm run docs`.

On verification: `tsc --noEmit`, ESLint and Prettier are clean, and the `NodeWebSocketTransport` unit tests still pass. `npm run unit` has one failure, `should reset timeout going over concurrency` (expected < 100, received 110), which I confirmed also fails on a clean tree with my changes stashed, so it is a pre-existing timing assertion rather than something this introduces.

Also worth noting the Firefox legs that were red yesterday are green again on this branch, 33 checks passing, consistent with that having been the repo-wide Firefox breakage rather than anything here.

One open question if you would rather decide it: I left the options off the public `puppeteer.connect()` docs page beyond the generated `ConnectOptions` entry. Happy to add a short prose note to the connect guide if you think it warrants one.

packages/puppeteer-core/src/common/ConnectOptions.ts
```diff
@@ -134,6 +134,30 @@ export interface ConnectOptions {
...
+  /**
+   * Ping period in milliseconds. Only used when {@link ConnectOptions.keepAlive}
+   * is set.
+   *
+   * @remarks
+   * Node.js only. Ignored in the browser build.
+   *
+   * @defaultValue `30_000`
+   */
+  keepAliveIntervalMs?: number;
```
OrKoN: let's introduce a new holder object for these: `wsOptions`. We should also move headers in there but keep the top-level headers options for compatibility (adding `@deprecated`).

rajanpanth:
Done in `4abe488`, thanks @OrKoN.

`WsOptions` now holds the websocket-level options, with `headers` moved in alongside the keep-alive ones:

```ts
export interface WsOptions {
  headers?: Record<string, string>;
  keepAlive?: boolean;
  keepAliveIntervalMs?: number;
}
```

The top-level `headers` stays and is marked `@deprecated`, pointing at `WsOptions.headers`. Resolution is `options.wsOptions?.headers ?? options.headers ?? {}`, so `wsOptions` wins when both are set and existing callers are untouched.

`WsOptions` is `@public`, so `LaunchOptions` picks it up through `ConnectOptions` and it reaches both entry points. Docs regenerated, which adds `docs/api/puppeteer.wsoptions.md`.

Also rebased onto `main`, since `WritableDestination` landed in the meantime and collided in the generated `docs/api/index.md`. I resolved that by taking `main`'s copy and re-running `npm run docs` rather than hand-editing generated output, and confirmed both entries are present afterwards.

`tsc --noEmit`, ESLint and Prettier are clean, and the transport tests pass. `npm run unit` still has the two `Page > waitForNetworkIdle` timing failures, which I re-confirmed reproduce on a clean tree with my changes stashed.

OrKoN (approved):

PR #15415, opened 2026-09-03
fix: revert launch browsers in detached mode on Windows to fix flakiness

### LINKED ISSUE #15410: [Bug]: Windows: unconditional `detached: true` makes each chrome-headless-shell.exe open a visible console window (regression in @puppeteer/browsers 3.2.1)
### Steps to reproduce

Windows 11, Node 24.20.0, npm 12.0.2.
npm i puppeteer@25.9.0

```
// repro.mjs - run with: node repro.mjs

const browser = await puppeteer.launch({headless: 'shell'});
await page.setContent('<h1>hello</h1>');
await page.pdf({path: 'out.pdf'});
await browser.close();
```

Then repeat with puppeteer@25.7.0 for comparison.

### Expected results

Nothing visible on screen. chrome-headless-shell runs invisibly, as it did up to and including puppeteer 25.7.0 / @puppeteer/browsers 3.2.0, and as the windowsHide: true at the spawn site is asking for.

### Actual results

From puppeteer 25.8.0 / @puppeteer/browsers 3.2.1 onwards, every launched chrome-headless-shell.exe gets its own visible console window, which stays on screen for the lifetime of the browser. An application that launches several browsers concurrently shows that many console windows (in our case four per generated PDF), which makes headless use unworkable in a desktop or service context.

**Cause**
#15339 ("fix: launch browsers in detached mode on Windows to fix flakiness", released in v25.8.0) changed the default in packages/browsers/src/launch.ts from platform-conditional to unconditional:
-    // On non-windows platforms, `detached: true` makes child process a
-    // leader of a new process group, making it possible to kill child
-    // process tree with `.kill(-pid)` command.
-    opts.detached ??= process.platform !== 'win32';
+    // `detached: true` makes child process a leader of a new process group,
+    // making it possible to isolate process trees and prevent console signal propagation.
+    opts.detached ??= true;
(packages/browsers/src/launch.ts line 332 on main.)

On Windows, detached: true makes libuv add DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP to the CreateProcessW creation flags (libuv src/win/process.c, in the UV_PROCESS_DETACHED branch). That has two consequences:

DETACHED_PROCESS stops the child inheriting the parent's console. chrome-headless-shell.exe is a console-subsystem binary (PE Subsystem = 3, IMAGE_SUBSYSTEM_WINDOWS_CUI), so Windows gives it a freshly allocated console of its own — a visible window.
The windowsHide: true added by the same PR cannot suppress that window. windowsHide maps to CREATE_NO_WINDOW, and the CreateProcess documentation states that flag "is ignored if the application is not a console application, or if it is used with either CREATE_NEW_CONSOLE or DETACHED_PROCESS."
In other words, on Windows the two flags that PR added are mutually exclusive and DETACHED_PROCESS wins.

It may also be worth noting that the stated motivation for detaching does not apply to the Windows code path: kill() in the same file already uses taskkill /pid <pid> /T /F on Windows (line 471 onwards on main) and only uses the process.kill(-pid) process-group trick on POSIX, so Windows never needed a new process group in order to tear down the process tree.

**Possible fixes**
Any of:

Restore the platform-conditional default, opts.detached ??= process.platform !== 'win32', and address the flakiness #15339 targeted another way.
Keep the isolation but without DETACHED_PROCESS on Windows — CREATE_NEW_PROCESS_GROUP alone provides the console-signal isolation and is compatible with CREATE_NO_WINDOW. Node's child_process cannot request those two flags separately, so this may need a Node/libuv change or a documented opt-out.
At a minimum, make it switchable from puppeteer. @puppeteer/browsers already exposes LaunchOptions.detached, but puppeteer.launch() offers no way to reach it, so puppeteer users currently have no workaround other than pinning 25.7.0.

**Versions**
Affected: puppeteer 25.8.0 and 25.9.0 (both pin @puppeteer/browsers 3.2.1 exactly)
Last good: puppeteer 25.7.0 (@puppeteer/browsers 3.2.0)
Node: 24.20.0
Package manager: npm 12.0.2
OS: Windows 11 Pro, 10.0.26200.9168, x64
Launch call: puppeteer.launch({headless: 'shell', args: ['--no-sandbox']})

OrKoN:
Partially reverts puppeteer/puppeteer#15339

Closes https://github.com/puppeteer/puppeteer/issues/15410

### DIFF
--- a/packages/browsers/src/launch.ts
+++ b/packages/browsers/src/launch.ts
@@ -326,10 +326,11 @@ export class Process {
     opts.handleSIGINT ??= true;
     opts.handleSIGTERM ??= true;
     opts.handleSIGHUP ??= true;
-    // `detached: true` makes child process a leader of a new process group,
-    // making it possible to isolate process trees and prevent console signal propagation.
-    // @see https://nodejs.org/api/child_process.html#child_process_options_detached
-    opts.detached ??= true;
+    // On non-windows platforms, `detached: true` makes child process a
+    // leader of a new process group, making it possible to kill child
+    // process tree with `.kill(-pid)` command. @see
+    // https://nodejs.org/api/child_process.html#child_process_options_detached
+    opts.detached ??= process.platform !== 'win32';
     const stdio = this.#configureStdio({
       pipe: opts.pipe,
     });


packages/puppeteer-core/src/node/BrowserLauncher.ts
```diff
@@ -172,11 +172,6 @@ export abstract class BrowserLauncher {
     });
 
     if (!existsSync(launchArgs.executablePath)) {
-      if (launchArgs.isTempUserDataDir) {
```
Lightning00Blade, line 175: Maybe we can keep this?

tools/mocha-runner/src/mocha-runner.ts
```diff
@@ -11,7 +11,6 @@ import {randomUUID} from 'node:crypto';
 import fs, {globSync} from 'node:fs';
```
Lightning00Blade, line 11: This change may be OK as well?

Lightning00Blade (approved):
LGTM, leave the comments to your discretion

packages/puppeteer-core/src/node/BrowserLauncher.ts
OrKoN: Good idea.

Lightning00Blade (approved):

Issue #11641, opened 2024-01-06
[Bug]: neither `requestfinished` nor `requestfailed` event is fired for duplicated requests


maxkoretskyi:
### Minimal, reproducible example

```TypeScript
const browser = await puppeteer.launch({
    headless: 'new',
    executablePath: 'google-chrome',
    args: [
        '--accept-lang=en-GB',
        '--no-sandbox',
        '--disable-setuid-sandbox',
        '--disable-dev-shm-usage',
    ],
});

const page = await browser.newPage();
await page.setRequestInterception(true);

page.on('request', (req) => console.log('onRequestStarted: ' + req.url()), req.continue());
page.on('requestfinished', (req) => console.log('onRequestFinished: ' + req.url()));
page.on('requestfailed', (req) => console.log('onRequestFailed: ' + req.url()));

await page.goto(url, {
    timeout: 60 * 1000,
    waitUntil: ['domcontentloaded'],
});
```

### Error string

no error

### Bug behavior

- [X] Flaky
- [ ] PDF

### Background

I'm tracking the number of requests/responses as part of the [custom networkidle event implementation](https://github.com/puppeteer/puppeteer/issues/1353#issuecomment-648299486). Suprisingly, I don't get any notification/event for a duplicated requests for fonts, hence the function that tracks request keep waiting indefinitely.

I've done some research. Firstly, by simple logging from handlers I identifiied the requests that cause trouble:

```
onRequestStarted: /KFOlCnqEu92Fr1MmEU9fBBc4.3170fd9aa9258fe0.woff2	onRequestFinished: /KFOlCnqEu92Fr1MmEU9fBBc4.3170fd9aa9258fe0.woff2
onRequestStarted: /KFOlCnqEu92Fr1MmEU9fBBc4.3170fd9aa9258fe0.woff2	-
onRequestStarted: /KFOlCnqEu92Fr1MmWUlfBBc4.71b2beb82c4d1389.woff2	onRequestFinished: /KFOlCnqEu92Fr1MmWUlfBBc4.71b2beb82c4d1389.woff2
onRequestStarted: /KFOlCnqEu92Fr1MmWUlfBBc4.71b2beb82c4d1389.woff2	-
onRequestStarted: /KFOmCnqEu92Fr1Mu4mxK.f2894edcf7d09d36.woff2		onRequestFinished: /KFOmCnqEu92Fr1Mu4mxK.f2894edcf7d09d36.woff2
[...]

PR #15416, opened 2026-09-04
fix(network): do not emit duplicate request on Fetch.requestPaused restart

cl-ansy:

Bugfix

Yes. Also updated the "double pause" test case above it that shows the bug. It feeds one `requestWillBeSent` followed by two `Fetch.requestPaused` events for the same network id, and asserted that Puppeteer emits two request events.

Not needed

**Summary**

When Chrome restarts a request it sends a second `Fetch.requestPaused` with a new interception id for the same `Network.requestWillBeSent` (crbug.com/1196004). This is common for fonts declared in stylesheets. Puppeteer emitted a second `request` for it, and since the two shared a network id the first never received `requestfinished` or `requestfailed`.

Repoint the existing request at the new interception id instead. If the user already resolved the interception, replay it with the original continue overrides so the restarted job does not hang.

Issues: #7060, #11641

No

### DIFF
--- a/packages/puppeteer-core/src/cdp/NetworkManager.test.ts
+++ b/packages/puppeteer-core/src/cdp/NetworkManager.test.ts
@@ -576,7 +576,84 @@ describe('NetworkManager', () => {
       networkId: '11ACE9783588040D644B905E8B55285B',
     });
 
-    expect(requests).toHaveLength(2);
+    expect(requests).toHaveLength(1);
+  });
+  it(`should emit requestfinished for a stylesheet font request that Chrome restarts (github.com/puppeteer/puppeteer/issues/11641)`, async () => {
+    const mockCDPSession = new MockCDPSession();
+    const manager = createNetworkManager();
+    await manager.addClient(mockCDPSession);
+    await manager.setRequestInterception(true);
+
+    const finished: HTTPRequest[] = [];
+    manager.on(NetworkManagerEvent.Request, async (request: HTTPRequest) => {
+      await request.continue();
+    });
+    manager.on(NetworkManagerEvent.RequestFinished, (request: HTTPRequest) => {
+      finished.push(request);
+    });
+
+    const fontRequest = {
+      url: 'https://example.com/fonts/font.woff2',
+      method: 'GET',
+      headers: {},
+      initialPriority: 'VeryHigh',
+      referrerPolicy: 'strict-origin-when-cross-origin',
+    } as const;
+    mockCDPSession.emit('Network.requestWillBeSent', {
+      requestId: '3C6E9E2A5D1B4F3F5E8C0B1A2D3E4F50',
+      loaderId: '11ACE9783588040D644B905E8B55285B',
+      documentURL: 'https://example.com/',
+      request: {...fontRequest, mixedContentType: 'none', isSameSite: true},
+      timestamp: 224604.980827,
+      wallTime: 1637955746.786191,
+      initiator: {type: 'parser', url: 'https://example.com/style.css'},
+      redirectHasExtraInfo: false,
+      type: 'Font',
+      frameId: '84AC261A351B86932B775B76D1DD79F8',
+      hasUserGesture: false,
+    });
+    // Chrome restarts the font fetch and pauses it again under a new
+    // interception id (crbug.com/1196004).
+    for (const requestId of ['interception-job-1.0', 'interception-job-2.0']) {
+      mockCDPSession.emit('Fetch.requestPaused', {
+        requestId,
+        request: fontRequest,
+        frameId: '84AC261A351B86932B775B76D1DD79F8',
+        resourceType: 'Font',
+        networkId: '3C6E9E2A5D1B4F3F5E8C0B1A2D3E4F50',
+      });
+    }
+    mockCDPSession.emit('Network.responseReceived', {
+      requestId: '3C6E9E2A5D1B4F3F5E8C0B1A2D3E4F50',
+      loaderId: '11ACE9783588040D644B905E8B55285B',
+      timestamp: 224605.0,
+      type: 'Font',
+      response: {
+        url: fontRequest.url,
+        status: 200,
+        statusText: 'OK',
+        headers: {},
+        mimeType: 'font/woff2',
+        charset: '',
+        connectionReused: false,
+        connectionId: 1,
+        encodedDataLength: 100,
+        securityState: 'secure',
+      },
+      hasExtraInfo: false,
+      frameId: '84AC261A351B86932B775B76D1DD79F8',
+    });
+    mockCDPSession.emit('Network.loadingFinished', {
+      requestId: '3C6E9E2A5D1B4F3F5E8C0B1A2D3E4F50',
+      timestamp: 224605.1,
+      encodedDataLength: 100,
+    });
+
+    expect(
+      finished.map(r => {
+        return r.url();
+      }),
+    ).toEqual([fontRequest.url]);
   });
   it(`should handle Network.responseReceivedExtraInfo event after Network.responseReceived event (github.com/puppeteer/puppeteer/issues/8234)`, async () => {
     const mockCDPSession = new MockCDPSession();
--- a/packages/puppeteer-core/src/cdp/NetworkManager.ts
+++ b/packages/puppeteer-core/src/cdp/NetworkManager.ts
@@ -473,27 +473,54 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
       return;
     }
 
-    const requestWillBeSentEvent = (() => {
-      const requestWillBeSentEvent =
-        this.#networkEventManager.getRequestWillBeSent(networkRequestId);
-
-      // redirect requests have the same `requestId`,
+    const requestWillBeSentEvent =
+      this.#networkEventManager.getRequestWillBeSent(networkRequestId);
+    if (!requestWillBeSentEvent) {
+      // Chrome may restart a request and send a second Fetch.requestPaused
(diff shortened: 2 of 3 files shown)


packages/puppeteer-core/src/cdp/NetworkManager.ts
```diff
@@ -473,27 +473,54 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
...
-
-      // redirect requests have the same `requestId`,
+    const requestWillBeSentEvent =
+      this.#networkEventManager.getRequestWillBeSent(networkRequestId);
+    if (!requestWillBeSentEvent) {
+      // Chrome may restart a request and send a second Fetch.requestPaused
+      // for a request Puppeteer already emitted (crbug.com/1196004). Repoint
+      // that request instead of emitting a duplicate that never finishes.
+      const request = this.#networkEventManager.getRequest(networkRequestId);
       if (
```
OrKoN, line 483: let's add a test for the following scenario: 

1. If a user overrides the URL or method via request.continue({ url: '...' }), Chrome fetches the new URL and emits the second pause with the overridden URL.
2. Because request.url() and request.method() remain the original values, request.url() === event.request.url evaluates to false, causing the restart to be missed and the request to hang.
3. Fix: Check against request.continueRequestOverrides().url ?? request.url() (and same for method).

packages/puppeteer-core/src/cdp/NetworkManager.ts
```diff
@@ -473,27 +473,54 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
...
-        requestWillBeSentEvent &&
-        (requestWillBeSentEvent.request.url !== event.request.url ||
-          requestWillBeSentEvent.request.method !== event.request.method)
+        request &&
+        request.url() === event.request.url &&
+        request.method() === event.request.method
       ) {
-        this.#networkEventManager.forgetRequestWillBeSent(networkRequestId);
+        this.#onRequestRestarted(request, fetchRequestId);
```
OrKoN, line 488: Same-URL Redirects vs Restarts: If an endpoint redirects (302/307) to the exact same URL and method, and Fetch.requestPaused arrives before Network.requestWillBeSent, it could be mistaken for a restart.

Fix: Check !event.redirectedRequestId and !request.response() before treating a request as restarted.

Let's add a test for this scenario.

packages/puppeteer-core/src/cdp/NetworkManager.test.ts
```diff
@@ -576,7 +576,84 @@ describe('NetworkManager', () => {
...
+      hasExtraInfo: false,
+      frameId: '84AC261A351B86932B775B76D1DD79F8',
+    });
+    mockCDPSession.emit('Network.loadingFinished', {
+      requestId: '3C6E9E2A5D1B4F3F5E8C0B1A2D3E4F50',
+      timestamp: 224605.1,
+      encodedDataLength: 100,
+    });
+
+    expect(
```
OrKoN, line 652: The new test verifies finished count, but does not spy on mockCDPSession.send(...) to confirm that Fetch.continueRequest was actually called for the new interception ID ('interception-job-2.0') or that continuation overrides (like headers) were replayed.

OrKoN (changes requested):

packages/puppeteer-core/src/cdp/NetworkManager.ts
cl-ansy: Done. Restart match now uses `continueRequestOverrides().url` ?? `url()`

Added a test that continues with a url override and asserts the replayed `Fetch.continueRequest` targets the new interception id

packages/puppeteer-core/src/cdp/NetworkManager.ts
cl-ansy: Added both guards (`!event.redirectedRequestId` and `!request.response()`) and a test where the redirect's `Fetch.requestPaused` arrives before its `Network.requestWillBeSent`

packages/puppeteer-core/src/cdp/NetworkManager.test.ts
cl-ansy: The restart test now records send calls and asserts `Fetch.continueRequest` was calls for `interception-job-2.0` with the original headers replayed

packages/puppeteer-core/src/cdp/NetworkManager.ts
```diff
@@ -473,27 +473,68 @@ export class NetworkManager extends EventEmitter<NetworkManagerEvents> {
...
-        this.#networkEventManager.getRequestWillBeSent(networkRequestId);
-
-      // redirect requests have the same `requestId`,
-      if (
-        requestWillBeSentEvent &&
-        (requestWillBeSentEvent.request.url !== event.request.url ||
-          requestWillBeSentEvent.request.method !== event.request.method)
-      ) {
-        this.#networkEventManager.forgetRequestWillBeSent(networkRequestId);
+    const requestWillBeSentEvent =
```
OrKoN, line 476: I think the following are still remaining edge cases:

### 1. `#onRequestRestarted` always calls `_continue()`, breaking restarted requests that were aborted or responded to
**File:** `packages/puppeteer-core/src/cdp/NetworkManager.ts` (lines 525–539) & `packages/puppeteer-core/src/api/HTTPRequest.ts` (line 435)

```ts
  #onRequestRestarted(
    request: CdpHTTPRequest,
    fetchRequestId: FetchRequestId,
  ): void {
    request._interceptionId = fetchRequestId;
    if (
      this.#userRequestInterceptionEnabled &&
      request.isInterceptResolutionHandled()
    ) {
      // The user already resolved the interception; the restarted job needs
      // the same resolution or the request hangs.
      void request._continue(request.continueRequestOverrides()).catch(err => {
        this.#logger?.(DEBUG_PREFIXES.error)?.(err);
      });
    }
  }
```

- **Issue**: `request.isInterceptResolutionHandled()` returns `true` not only when the user called `request.continue()`, but also when the user called `request.abort()` or `request.respond()`.
  - If a user blocks fonts (`if (req.resourceType() === 'font') await req.abort()`) or mocks a response (`await req.respond(...)`), `request.response()` is still `null` when the restarted `Fetch.requestPaused` arrives (`_response` is only populated when `Network.responseReceived` arrives from CDP), so `#isRestartedRequest` returns `true`.
  - `#onRequestRestarted` then executes `request._continue(...)`, which **continues the aborted/mocked request to the network** instead of re-aborting or re-fulfilling it.
- **Fix**:
  - Just as the PR moved `this.interception.requestOverrides = overrides` above `if (priority === undefined)` in `HTTPRequest.continue()`, do the same for `this.interception.response = response` in `HTTPRequest.respond()` and `this.interception.abortReason = errorReason` in `HTTPRequest.abort()`, and record the action taken (`Continue`, `Respond`, or `Abort`).
  - In `#onRequestRestarted`, replay `_continue`, `_respond`, or `_abort` matching the user's resolution action.

---

### 2. `#isRestartedRequest` fails on URLs with `#fragment` (`request.url()` vs `event.request.url`)
**File:** `packages/puppeteer-core/src/cdp/NetworkManager.ts` (lines 506–523)

```ts
  #isRestartedRequest(
    request: CdpHTTPRequest,
    event: Protocol.Fetch.RequestPausedEvent,
  ): boolean {
    if (event.redirectedRequestId || request.response()) {
      return false;
    }
    // Chrome pauses the request the user continued to, not the original.
    const overrides = request.continueRequestOverrides();
    return (
      (overrides.url ?? request.url()) === event.request.url &&
      (overrides.method ?? request.method()) === event.request.method
    );
  }
```

- **Issue**: On `CdpHTTPRequest`, `request.url()` includes the URL fragment (`data.request.url + (data.request.urlFragment ?? '')`). However, on CDP's `Protocol.Fetch.RequestPausedEvent`, `event.request.url` is **without fragment** (any hash is in `event.request.urlFragment`).
  - Stylesheets often declare fonts with URL fragments (e.g., `url('font.eot?#iefix')` or `url('font.svg#id')`).
  - For these requests, `request.url()` contains `#...` while `event.request.url` does not, causing `request.url() === event.request.url` to return `false` and the restarted request to hang.
- **Fix**: Include `event.request.urlFragment` when comparing against `request.url()`:
  ```ts
  const eventUrl = event.request.url + (event.request.urlFragment ?? '');
  ```

---

### 3. `#isRestartedRequest` fails if Chrome emits the second pause before `Fetch.continueRequest` overrides take effect
**File:** `packages/puppeteer-core/src/cdp/NetworkManager.ts` (lines 517–522)

```ts
    const overrides = request.continueRequestOverrides();
    return (
      (overrides.url ?? request.url()) === event.request.url &&
      (overrides.method ?? request.method()) === event.request.method
    );
```

- **Issue**: When `request.continue({ url, method })` is called, `overrides.url` is set synchronously in Node.js before the `Fetch.continueRequest` command is processed by Chrome over WebSocket.
  - If Chrome emits the second `Fetch.requestPaused` before receiving `Fetch.continueRequest` (or if Blink restarts the font fetch using the original request URL), `event.request.url` on the second pause will still be the original `request.url()`.
  - Because `overrides.url` is defined, `(overrides.url ?? request.url())` evaluates only against `overrides.url`, returning `false` and causing the restarted request to hang.
- **Fix**: Match if the paused event URL/method matches *either* the original request URL/method *or* the overridden URL/method:
  ```ts
  const eventUrl = event.request.url + (event.request.urlFragment ?? '');
  const urlMatches =
    eventUrl === request.url() ||
    eventUrl === overrides.url ||
    event.request.url === overrides.url;
  const methodMatches =
    event.request.method === request.method() ||
    event.request.method === overrides.method;
  return urlMatches && methodMatches;
  ```

OrKoN (changes requested):

PR #15435, opened 2026-09-08
fix(connect): preserve browserURL path prefixes

dajiaohuang:
**What kind of change does this PR introduce?**  Bug fix.  **Did you add tests for your changes?**  Yes. Added unit coverage for a path-prefixed `browserURL`, including a trailing slash and query string.  **If relevant, did you update the documentation?**  Not applicable.  **Summary**  `getWSEndpoint` now appends `/json/version` to the existing `browserURL` pathname, preserving reverse-proxy and tunnel path prefixes. Addresses #15250.  **Does this PR introduce a breaking change?**  No.  **Other information**  Validated with `npm run build`, `npm run unit --workspace puppeteer-core`, the focused `BrowserConnector` unit test, ESLint, Prettier, expectations lint, and repository lint. The repository check remains unable to run on Windows because its existing dependency-version script invokes the Unix-only `tail` command.

### DIFF
--- a/packages/puppeteer-core/src/common/BrowserConnector.test.ts
+++ b/packages/puppeteer-core/src/common/BrowserConnector.test.ts
@@ -51,6 +51,35 @@ describe('BrowserConnector', () => {
         ],
       ).toBe('Bearer test-token');
     });
+
+    it('should preserve the browserURL path prefix', async () => {
+      const capturedRequests: string[] = [];
+
+      mock.method(
+        globalThis,
+        'fetch',
+        async (url: string | URL | Request) => {
+          capturedRequests.push(String(url));
+          return new Response(
+            JSON.stringify({
+              webSocketDebuggerUrl: 'ws://localhost:1234/devtools/browser/1',
+            }),
+            {status: 200, headers: {'Content-Type': 'application/json'}},
+          );
+        },
+      );
+
+      await _connectToBrowser({
+        browserURL: 'http://localhost:1234/t/session/?token=secret',
+        logger: () => {
+          return undefined;
+        },
+      }).catch(() => {});
+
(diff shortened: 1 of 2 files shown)


packages/puppeteer-core/src/common/BrowserConnector.ts
```diff
@@ -198,7 +198,10 @@ async function getWSEndpoint(
   browserURL: string,
   headers?: Record<string, string>,
 ): Promise<string> {
-  const endpointURL = new URL('/json/version', browserURL);
+  const endpointURL = new URL(browserURL);
+  endpointURL.pathname =
+    endpointURL.pathname.replace(/\/?$/, '') + '/json/version';
+  endpointURL.hash = '';
```
OrKoN: Thanks for the PR! Let's extract this into a separate helper function and test URL rewrites in the unit test without mocks.

dajiaohuang:
Implemented the requested follow-up in commit 0d1a6bf: extracted browserURL-to-/json/version rewriting into exported getWSEndpointURL(), and added a direct unit test covering path prefix, query preservation, and hash removal without fetch/WebSocket mocks. Validation: targeted BrowserConnector tests 6/6, Prettier check for both files, and git diff --check pass.

packages/puppeteer-core/src/common/BrowserConnector.test.ts
```diff
@@ -51,6 +57,35 @@ describe('BrowserConnector', () => {
         ],
       ).toBe('Bearer test-token');
     });
+
+    it('should preserve the browserURL path prefix', async () => {
```
OrKoN, line 61: let's remove this test then!

Issue #15436, opened 2026-09-08
[Bug]: npm run check fails on Windows because the dependency check invokes tail


dajiaohuang:
### Minimal, reproducible example

On Windows PowerShell, from Puppeteer main at `a2086c77718ac817784453cc4f489ffda2fedf12`:

```powershell
npm ci --ignore-scripts
npm run check --workspace puppeteer-core
```

### Background

The contributing guide documents `npm run check` as part of the maintainer workflow. The dependency-version check in `packages/puppeteer-core/tools/ensure-correct-devtools-protocol-package.mts` runs `npm view ... | tail -1`.

### Expectation

The repository check should run on supported Windows development environments.

### Reality

The command exits before validating the dependency because `tail` is not available in the default Windows shell:

```
'tail' is not recognized as an internal or external command,
operable program or batch file.
```

### Puppeteer version

25.10.0

### Node version

24.14.1

### Package manager

npm

### Package manager version

11.11.0

### Operating system

Windows

PR #15438, opened 2026-09-08
chore: avoid Unix-only tail in protocol check

dajiaohuang:
**What kind of change does this PR introduce?**  Bug fix.  **Did you add tests for your changes?**  The existing dependency-version check now runs cross-platform; validation exercises it on Windows with the current pinned revision.  **If relevant, did you update the documentation?**  Not applicable.  **Summary**  Replace the Unix-only `tail` pipeline with npm's JSON output and select the newest matching protocol version in Node. This lets `npm run check --workspace puppeteer-core` run from the supported Windows shell. Addresses #15436.  **Does this PR introduce a breaking change?**  No.  **Other information**  Validated on Windows with `npm run check --workspace puppeteer-core` and Prettier.

### DIFF
--- a/packages/puppeteer-core/tools/ensure-correct-devtools-protocol-package.mts
+++ b/packages/puppeteer-core/tools/ensure-correct-devtools-protocol-package.mts
@@ -59,16 +59,14 @@ async function main() {
 
   console.log(`Revisions for ${chromeVersion}: ${chromeRevision}`);
 
+  const command = `npm view "devtools-protocol@<=0.0.${chromeRevision}" version --json`;
   console.log(
     'Checking npm for devtools-protocol revisions:\n',
-    `'npm view "devtools-protocol@<=0.0.${chromeRevision}" version --json'`,
+    `'${command}'`,
     '\n',
   );
 
-  const output = execSync(
-    `npm view "devtools-protocol@<=0.0.${chromeRevision}" version --json`,
-    {encoding: 'utf8'},
-  );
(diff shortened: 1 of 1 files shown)


packages/puppeteer-core/tools/ensure-correct-devtools-protocol-package.mts
```diff
@@ -59,19 +59,21 @@ async function main() {
...
-    `'${command}'`,
+    `'npm view "devtools-protocol@<=0.0.${chromeRevision}" version --json'`,
     '\n',
   );
 
-  const output = execSync(command, {
-    encoding: 'utf8',
-  });
+  const output = execSync(
+    `npm view "devtools-protocol@<=0.0.${chromeRevision}" version --json`,
```
OrKoN: let's keep the command as a variable instead of duplicating it.

OrKoN (changes requested):

packages/puppeteer-core/tools/ensure-correct-devtools-protocol-package.mts
dajiaohuang: Moved the npm lookup into the command variable and reuse it for both logging and execution in 948f90f. npm run check --workspace puppeteer-core and targeted Prettier validation pass; the existing unit suite remains affected by local 0.0.0.0:8080 permission and timing failures.

OrKoN (approved):

PR #15441, opened 2026-09-09
fix(launcher): clean temporary profiles on process exit

### LINKED ISSUE #15156: [Bug]: leftover files cause hash collisions on APFS which prevents the OS from booting in a reasonable time

dajiaohuang:

Bugfix: remove temporary browser profiles synchronously when the Node.js host process exits unexpectedly.

Yes. Added unit coverage for synchronous cleanup on an exit event and for unregistering the fallback after normal browser cleanup.

No documentation changes are needed.

**Summary**

When the host process terminates due to an uncaught exception or `process.exit()`, the browser child process cannot run Puppeteer's existing asynchronous profile cleanup. This change registers a synchronous `process.exit` fallback in `puppeteer-core` for each temporary profile, unregistering it after the browser process cleanup completes or launch fails. The temporary profile naming remains unchanged.

Related issue: #15156

No.

**Other information**

Validation:
- targeted Prettier and ESLint checks
- targeted `BrowserLauncher.test.js`: 2 passed
- full `npm run unit --workspace puppeteer-core`: 179 passed, 6 environment-sensitive failures (Page timing and NodeWebSocketTransport binding to `0.0.0.0:8080`)

### DIFF
--- a/packages/puppeteer-core/src/node/BrowserLauncher.test.ts
+++ b/packages/puppeteer-core/src/node/BrowserLauncher.test.ts
@@ -16,6 +16,10 @@ import {registerProcessExitCleanup} from './BrowserLauncher.js';
 class ProcessEmitter {
   #listeners = new Set<() => void>();
 
+  get listenerCount(): number {
+    return this.#listeners.size;
+  }
+
   once(_event: 'exit', listener: () => void): void {
     this.#listeners.add(listener);
   }
@@ -66,5 +70,24 @@ describe('registerProcessExitCleanup', () => {
 
     expect(existsSync(userDataDir)).toBe(true);
     rmSync(userDataDir, {recursive: true, force: true});
+  });
+
+  it('uses one process-exit listener for multiple temporary profiles', () => {
+    const firstDir = mkdtempSync(
+      join(tmpdir(), 'puppeteer-process-exit-cleanup-'),
+    );
+    const secondDir = mkdtempSync(
+      join(tmpdir(), 'puppeteer-process-exit-cleanup-'),
+    );
+    const processEmitter = new ProcessEmitter();
+
+    registerProcessExitCleanup(firstDir, logger, processEmitter);
+    registerProcessExitCleanup(secondDir, logger, processEmitter);
+    expect(processEmitter.listenerCount).toBe(1);
+
+    processEmitter.emit('exit');
+
+    expect(existsSync(firstDir)).toBe(false);
+    expect(existsSync(secondDir)).toBe(false);
   });
 });
--- a/packages/puppeteer-core/src/node/BrowserLauncher.test.ts
+++ b/packages/puppeteer-core/src/node/BrowserLauncher.test.ts
@@ -1,0 +1,70 @@
+/**
+ * @license
+ * Copyright 2026 Google Inc.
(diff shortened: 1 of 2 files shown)


nattallius (changes requested):
Thank you for tackling this! The sync fallback on process exit can be the right direction.

A few things worry me before we merge:

- Ordering: Since this is registered before launch(), Node runs rmSync before @puppeteer/browsers calls kill(), so we're trying to delete the folder while Chrome is still running. We need kill() to happen first.

- Listener leak: Adding an exit listener per browser will trigger Node's MaxListenersExceededWarning if someone launches >10 browsers in parallel. We should share a single listener over a Set of dirs.

- Windows locks: On Windows, taskkill is async and Chrome often holds open file handles in the profile for a moment after being killed. A single rmSync will likely fail with EBUSY. Should we add retries (maxRetries), wait for the browser PID to exit first, or use a detached cleanup script?

- Integration test: Could you add an e2e test that spawns a crashing Node process to verify this works end-to-end?

What do you think?

dajiaohuang:
Implemented the requested lifecycle fixes in 06cae2fafbe.\n\n- Cleanup registration now happens after @puppeteer/browsers installs its process-exit dispatcher, so the browser is killed before the synchronous fallback removes the profile.\n- All temporary profiles share one exit listener, avoiding one listener per browser and MaxListenersExceededWarning.\n- Synchronous removal uses bounded retries for transient Windows file locks.\n- Added focused coverage for multi-profile listener sharing; targeted tests, package build, Prettier, ESLint, and git diff --check pass.\n\nI did not add a crash-process e2e test because the lifecycle ordering is covered at the process-emitter boundary and the existing hosted CI exercises the browser process lifecycle.

nattallius (approved):

